PR #34438, opened 2023-10-06
FE: Do not show linked filter options if the parameter has values_source_type of "static-list" or "card"

calherries:
Part 2 out of 3 for #33892, see comment [here](https://github.com/metabase/metabase/issues/33892#issuecomment-1752488489) for the plan.

Part 1 is here and can be merged separately: #34424

This PR prevents users from linking dashboard filter's values to another filter if the filter's values source is not compatible with linked filters.

Linked filters only work if a filter has "From connected fields" selected under "Where values should come from" in the filter settings (this corresponds to `values_source_config != undefined` in the code).

PR #34424, opened 2023-10-06
BE: Remove linked filters from dashboard if values source type is not "From connected fields"

calherries:
BE part of #33892. Read [this comment](https://github.com/metabase/metabase/issues/33892#issuecomment-1752488489) to understand why this change is needed. 

Effectively migrates dashboard parameters on demand as they come out of the database.

In English the migration does the following:

IF the parameter has a "linked filter" (aka "filtering param" in the code) and "Where values should come from" is not "From connected fields" (in the code, values_source_config is not nil):

THEN remove the linked filters from the parameter.

Issue #33892, opened 2023-09-14
Linked filters are not working with "Custom List" or "From another model or question"


Kavinya21:
### Describe the bug

I am trying to use linked filter in dropdown but it not working
- But after saving when filter1 is selected filter2 is not limiting its value based on the filter as it is linked all the values are appearing.

### To Reproduce

- Dropdown filters have been used via filter --> Text or Category --> is --> Dropdown.
- In Selectable value tried both **From another model or question** and **Custom list**.
- In the Linked filter **Limit this filter's choices** have also been selected.

### Information about your Metabase installation

```JSON
Google Chrome Version 116.0.5845.188 
Window 11
Postgres
Metabse Version 0.47.0
```

### Severity

Not able to connect any of teh linked filters

calherries:
Repro'd. 

If the limited filter's values source is either of type "Custom list" or "From another model or question" then the filter values are not limited in the dropdown. It doesn't matter what the type is for the filter doing the limiting.

<img width="214" alt="image" src="https://github.com/metabase/metabase/assets/39073188/3d02f388-1d14-4edb-92f3-3f7f9e3e68b2">

Here's a full demo:
https://www.loom.com/share/f9c221a0486d4638b0b11b35359ebb63?sid=fca14dbe-ce10-41e1-a45f-f0fdd2056df3

calherries:
@cdeweyx Linked filters have never worked with filters that have "Custom List" or "From another model or question" as their values source. What should we do here? 

I see two paths forward:
1. Support linked filters for "Custom List" and/or "From another model or question". The downside is development time vs other features. Linked filters for "Custom List" is relatively easy and "From another model or question" is much more difficult and should probably wait for the MLv2 refactor to complete.
2. Enforce linked filters to never be enabled when the values source is one of these options. We can add a BE migration on previously created filters so their linked filters toggle is disabled with these options, and we can update the UI to prevent it being set on new parameters. This is considerably less work than (1).

Aside from this issue, linked filters only work for GUI questions, not native ones (I can't find an issue for this). That's probably not changing any time soon either, at least until MLv2 is complete. We face the same decision with that issue too. 

Related #16872

calherries:
[From slack](https://metaboat.slack.com/archives/C057T1QTB3L/p1696573883883329?thread_ts=1696518563.953549&cid=C057T1QTB3L)

Path 1 was chosen. 

Then it needs three changes:
[...]

PR #34531, opened 2023-10-10
FE: Disable invalid values source options if parameter has linked filters

calherries:
Epic: https://github.com/metabase/metabase/issues/34604

Part 3 out of 3 for #33892, see comment [here](https://github.com/metabase/metabase/issues/33892#issuecomment-1752488489) for the plan.

Part 1 is here: #34424
Part 2 is here: #34438

This PR prevents users from setting a filter's values source to anything other than "From connected fields" if the filter has any linked filters. It does so by disabling the "Edit" button here and showing an explanation in the tooltip:

image

### DIFF
--- a/frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
+++ b/frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
@@ -92,7 +92,7 @@ const RadioLabel = ({
           isEnabled={!!isEditDisabled}
         >
           {/* This div is needed to make the tooltip work when the button is disabled */}
-          <div>
+          <div data-testid="values-source-settings-edit-btn">
             <Button
               onClick={onEditClick}
               disabled={!!isEditDisabled}
--- a/frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
+++ b/frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
@@ -80,7 +80,7 @@ const RadioLabel = ({
   title,
   isSelected,
   onEditClick,
-  isEditDisabled,
+  isEditDisabled = false,
 }: RadioLabelProps): JSX.Element => {
   return (
     <RadioLabelRoot>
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
```diff
@@ -62,38 +64,68 @@ const ValuesSourceSettings = ({
   );
 };
 
+function isOnlyConnectedFieldSourceAllowed(parameter: Parameter) {
```
ranquild, line 67: The name and the function body is a bit hard to follow. I'd change it to something like:
```
function hasLinkedFilters({ filteringParameters }: Parameter) {
  return filteringParameters != null && filteringParameters.length > 0;
}
```

and later use it like this:
```
isEditDisabled: hasLinkedFilters(parameter)
```

frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
```diff
@@ -62,38 +64,68 @@ const ValuesSourceSettings = ({
...
   isSelected?: boolean;
   onEditClick?: () => void;
+  isEditDisabled?: boolean;
 }
 
 const RadioLabel = ({
   title,
   isSelected,
   onEditClick,
+  isEditDisabled,
```
ranquild, line 83: set default value here `isEditDisabled = false` to avoid using `!!` later

frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.unit.spec.tsx
```diff
@@ -48,4 +48,46 @@ describe("ValuesSourceSettings", () => {
       values: ["A"],
     });
   });
+
+  it("Editing the values source should be disabled when the filter has linked filters", () => {
```
ranquild, line 52: we usually don't sentence-case test names, it should be `editing the values ...`

ranquild (approved):
Please take a look at comments before merging

calherries:
@ranquild implemented your suggestions, thanks

PR #34412, opened 2023-10-05
Add missing models to `copy` command

camsaul:
Fixes #34411

PR #34533, opened 2023-10-10
Manual backport of #34412

camsaul:


enterprise/backend/test/metabase_enterprise/models/entity_id_test.clj
```diff
@@ -77,6 +75,8 @@
 
 (deftest ^:parallel comprehensive-entity-id-test
   (doseq [model (->> (v2.seed-entity-ids/toucan-models)
+                     (remove (fn [model]
+                               (not= (namespace model) "model")))
```
metamben, line 79: In fact, filtering the ones in namespace "model" would be simpler.

enterprise/backend/test/metabase_enterprise/models/entity_id_test.clj
```diff
@@ -86,6 +86,8 @@
 
 (deftest ^:parallel comprehensive-identity-hash-test
   (doseq [model (->> (v2.seed-entity-ids/toucan-models)
+                     (remove (fn [model]
+                               (not= (namespace model) "model")))
```
metamben, line 90: Here too.

src/metabase/cmd/copy.clj
```diff
@@ -345,7 +311,27 @@
...
+  keyword)
+
+(defmethod postgres-id-sequence-name :default
+  [model]
+  (str (name (t2/table-name model)) "_id_seq"))
+
+;;; we changed the table name to `sandboxes` but never updated the underlying ID sequences or constraint names.
+(defmethod postgres-id-sequence-name :model/GroupTableAccessPolicy
+  [_model]
+  "group_table_access_policy_id_seq")
```
metamben, line 334: Does this really have to be extensible? Wouldn't it be simpler to use a `case` form?

metamben (approved):
LGTM

metamben (approved):
LGTM

PR #33690, opened 2023-09-03
Upgrade ace editor

dragonsahead:
This PR will do the following:
1) upgrade ace editor
2) upgrade react-ace (made a fork of the original package, trimmed some useless functionality and upgraded ace to the latest version)
3) shed many imports, I don't think these are needed

With these changes now the application is not only lighter, but loads faster (these changes trim 200kb from the bundle, with a gain of 100ms of the load time)

PR #33102, opened 2023-08-10
Migrate `NativeQueryEditor` to TypeScript

kulyk:
Migrates `NativeQueryEditor` to TypeScript trying to avoid any significant code changes. TypeScript should become an additional safety net while refactoring the editor (it's _very_ hard to unit test because of the ace editor).

* had to extend `ace-builds.d.ts`. Not sure, but seems like we're using a lot of private stuff 😢 
* a lot of changes are null checks for ace editor ref and resizable box ref

PR #34534, opened 2023-10-10
Remove db dependant extensions

dragonsahead:
Solves a lot of frustration with keywords in SQL autocomplete, see https://github.com/metabase/metabase/issues/34601

This PR does 2 things:
1) Wipes many db specific extensions of ace-editor (e.g. https://github.com/metabase/metabase/blob/be4e6cd0463fad83fad568e795061f5198eaaff6/frontend/src/metabase/query_builder/components/NativeQueryEditor/NativeQueryEditor.tsx#L22)

The reason I do that is because we add keywords for postgres, mysql and sqlserver, so we should add that for all DBs or for none of them
2) we were pulling src-min-noconflict AND src-noconflict on ace-builds. As we can't yet transform to src-min-noconflict completely, I'm droping the src-min-noconflict to unify to just 1 package so we don't add it twice

EDIT: this is not yet https://github.com/metabase/metabase/pull/33690, which is a separate project


frontend/src/metabase/query_builder/components/NativeQueryEditor/NativeQueryEditor.tsx
```diff
@@ -12,16 +12,8 @@ import "ace/ace";
 import "ace/ext-language_tools";
 import "ace/ext-searchbox";
 import "ace/mode-sql";
-import "ace/mode-mysql";
-import "ace/mode-pgsql";
-import "ace/mode-sqlserver";
-import "ace/mode-json";
```
calherries, line 18: We still need mode-json, otherwise autocomplete will default to SQL keywords for Mongo and Druid

calherries:
I've brought back json-mode and removed the modes from [getEngineNativeAceMode](https://github.com/metabase/metabase/pull/34534/commits/ac50a1b5dd2d7b9d5cc45c01bce458391c8939c2).

But this is definitely an improvement in the experience.

One cherry-picked example for MySQL:

Before:
image

After:
image

frontend/src/metabase/query_builder/components/NativeQueryEditor/NativeQueryEditor.tsx
```diff
@@ -12,15 +12,9 @@ import "ace/ace";
 import "ace/ext-language_tools";
 import "ace/ext-searchbox";
 import "ace/mode-sql";
-import "ace/mode-mysql";
-import "ace/mode-pgsql";
-import "ace/mode-sqlserver";
```
kamilmielnik, line 17: We should also remove related mocks:
https://github.com/metabase/metabase/blob/cd852d225526178a034cc79dfce38b8995d2e415/frontend/test/__support__/mocks.js#L29-L31

https://github.com/metabase/metabase/blob/cd852d225526178a034cc79dfce38b8995d2e415/frontend/test/__support__/mocks.js#L34-L36

webpack.config.js
```diff
@@ -164,7 +164,7 @@ const config = (module.exports = {
       __support__: TEST_SUPPORT_PATH,
       e2e: E2E_PATH,
       style: SRC_PATH + "/css/core/index",
-      ace: __dirname + "/node_modules/ace-builds/src-min-noconflict",
```
kamilmielnik, line 167: > 2) As we can't yet transform completely to src-min-noconflict completely

Why not?

`src-min-noconflict` has been used [since forever](https://github.com/metabase/metabase/commit/7af54fa27d270230e0268e0a02fc2eb26868d381).
I've been trying to understand why we're using `src-noconflict` and got as far as this: #33102
I don't see a reason why we used `src-noconflict` there.

kamilmielnik (commented):
I don't fully understand the change so I left a question.

----

BTW is there an issue for it that we can link here?

webpack.config.js
calherries: I don't know the answer to this question either. @kulyk can you explain if `src-noconflict` is needed over `src-min-noconflict`?

kulyk: TBH I think I just copied it from [`ExpressionEditorTextfield`](https://github.com/metabase/metabase/blob/a2d3b0052093e5def5a5d66d9d9b2fca55aef05a/frontend/src/metabase/query_builder/components/expressions/ExpressionEditorTextfield/ExpressionEditorTextfield.tsx#L7) to make TypeScript happy, there's no intention in picking `src-noconflict` here

calherries:
> BTW is there an issue for it that we can link here?

I added an issue for this and put it in the description https://github.com/metabase/metabase/issues/34601

webpack.config.js
calherries: Is there any reason to prefer `src-min-noconflict` over `src-noconflict`? It all gets minified in the build anyway

kulyk: Not sure, but if both are eventually minified, `src-noconflict` could be more handy in development if we need to look something up in ace's source code

kamilmielnik: > It all gets minified in the build anyway

This could affect build speed though - worth measuring.

calherries: > leave it for another PR

Done [src-min-noconflict -> src-noconflict](https://github.com/metabase/metabase/pull/34602#top)

kamilmielnik (approved):

webpack.config.js
paoliniluis: If you see what webpack does on yarn dev, we end up pulling src-min-noconflict and also with src-noconflict, 2 deps that are exactly the same, one minified, one not. By using always src-noconflict we ensure we pull the dep just once as webpack will end up minifying it on production builds. My change was only to ensure we pull things just once and reuse it as much as possible

kamilmielnik: > As we can't yet transform completely to src-min-noconflict completely

@paoliniluis Can you please explain why we can't transform completely to src-min-noconflict?

paoliniluis: I have my suspicions that the aliasing in the webpack config doesn’t work, at least when you do yarn , you still see both deps being pulled and separate

PR #34538, opened 2023-10-10
Don't suggest joining something by its implicitly joinable columns

### LINKED ISSUE #34526: [MLv2] Unexpected `suggested-condition` behavior
`suggested-condition` behaves strangely when following these steps:

> **note**
> Please use the `mlv2-fe-joins/feature-branch` branch to reproduce

2. Summarize Count by Orders.USER_ID
3. Save as Q1
4. New > Question > Raw Data > Sample Database > Orders
5. Join Q1
6. Inspect the suggested condition

**Expected behavior**
I don't think there's a meaningful condition to offer TBH

**Actual behavior**
- the suggested LHS column is Orders.USER_ID, but it's not marked as selected in `join-condition-lhs-columns` output
- the suggested RHS column is People.ID even though we're not joining the People table, it seems like `suggested-condition` is doing an implicit join

A bit more context in [internal discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1696946184701849)

metamben:
Fixes #34526

### DIFF
--- a/src/metabase/lib/join.cljc
+++ b/src/metabase/lib/join.cljc
@@ -758,9 +758,10 @@
   "Given something `x` (e.g. a Table metadata) find the PK column."
   [query        :- ::lib.schema/query
    stage-number :- :int
-   x]
+   x
+   options]
   (m/find-first lib.types.isa/primary-key?
-                (lib.metadata.calculation/visible-columns query stage-number x)))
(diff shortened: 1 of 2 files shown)


kulyk (approved):
Works as expected 👍

snoe (approved):
LGTM

PR #34544, opened 2023-10-10
Add dashboard parameter mapping for date range

### LINKED ISSUE #17879: Click Behavior on ranges always parses single-date to destination question's filters
**Describe the bug**
When a question is grouped by ex. month and adding a Click Behavior to another question, then it parses on a single-date instead of the entire month, causing misleading results unless the filter change is noticed.

This is basically the same as #13785, which was fixed in 0.40.0, but only for dashboard filters.
This might not be simple to fix.

**To Reproduce**
1. Simple question > Sample Datasets > Products - save as "Q1"
2. Simple question > Sample Datasets > Reviews - summarize Count by CreatedAt:Month - save as "Q2" and add to dashboard
3. On dashboard, hover "Q2" > Click Behavior > Destination: Question > find "Q1" and connect CreatedAt to CreatedAt.
image
4. Now click one of the points - ex. May 2018
image
5. Now you are taken to the Click Behavior question, but the filter is now May 1st 2018, so it's only returning 1 row instead of the expected 6 rows
image

Tested 0.40.0 thru 0.40.4 and master `552ab7d`

deniskaber:
Closes https://github.com/metabase/metabase/issues/17879

### Description

This improves how we pass data on dashboard chart data point click to target question. In case source question has data binning we change filter date to be a date range.

### How to verify

Check reproduction steps in [the issue](https://github.com/metabase/metabase/issues/17879)

### Demo

https://github.com/metabase/metabase/assets/7983744/5386dfc8-4b83-44cc-9fa5-489e173916a0

### Checklist

### DIFF
--- a/e2e/test/scenarios/dashboard/reproductions/17879-date-parameter-mapping.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/reproductions/17879-date-parameter-mapping.cy.spec.js
@@ -1,0 +1,131 @@
+import {
+  editDashboard,
+  popover,
+  restore,
+  saveDashboard,
+  showDashboardCardActions,
+  visitDashboard,
+} from "e2e/support/helpers";
+import { ORDERS, ORDERS_ID } from "metabase-types/api/mocks/presets";
+
+describe("issue 17879", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsAdmin();
+
+    cy.intercept("POST", "/api/dashboard/*/dashcard/*/card/*/query").as(
+      "dashcardQuery",
+    );
+  });
+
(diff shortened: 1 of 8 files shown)


ranquild (commented):
The fix doesn't seem to work with some temporal units, like 
Screenshot 2023-10-16 at 14 11 58

I think the approach used in this PR can only work with date truncation units https://github.com/metabase/metabase/blob/18c1b1984e4e4ff2c62aa279e15eae96d12be27a/src/metabase/lib/schema/temporal_bucketing.cljc#L26. For extraction units it's unclear. I think we should do the following:
- Add an explicit check for unit which we can convert to date ranges. The current check for  `"default", day"` is incorrect.
- Ask product people about other units and what we should do with them in click actions.

ranquild (approved):

PR #34545, opened 2023-10-10
replace `java-time` namespace with `java-time.api`

### LINKED ISSUE #34506: Replace `java-time` namespace with `java-time.api`
We should remove the uses of `[java-time :as t]` in our codebase. Note that someone has realized they are deprecated and suppressed it, but we should just move over to the `java-time.api` namespace.

```clojure
    #_{:clj-kondo/ignore [:deprecated-namespace]}
    [java-time :as t]
```

```clojure
;; NOTE: This namespace is generated by java-time.dev.gen
(ns ^{:deprecated "1.1.0", :doc "This namespace has been deprecated due to [#91](https://github.com/dm3/clojure.java-time/issues/91).\n     Please migrate to [[java-time.api]]\n\n     This namespace will continue to exist and be updated. For\n     maximum JVM compatibility, please migrate to `java-time.api`,\n     which provides the same interface. `java-time` and `java-time.api`\n     and can be freely intermixed within the same library, so you can\n     safely migrate your own code even if your dependencies use the old namespace.\n     \n     Migration steps:\n\n     1. rename all references from `java-time` to [[java-time.api]].\n        eg., `(:require [java-time :as jt])` => `(:require [java-time.api :as jt])`", :superseded-by "java-time.api"} java-time (:refer-clojure :exclude (zero? range iterate max min contains? format abs)) (:require [java-time core properties temporal amount zone single-field local chrono convert sugar seqs adjuster interval format joda clock pre-java8 mock]))
```

https://github.com/dm3/clojure.java-time/issues/91

johnswanson:
This namespace was deprecated and replaced with `java-time.api`.

Closes https://github.com/metabase/metabase/issues/34506

### Description

The bare `java-time` was deprecated and replaced with `java-time.api` because single-segment namespaces don't work well with GraalVM. We can migrate over to the new namespace with no effects - I have verified that the namespaces are identical except for the deprecation notice.

### Checklist

- [x] Tests have been added/updated to cover changes in this PR (the existing tests should cover this comfortably)

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/audit_app/pages/common.clj
+++ b/enterprise/backend/src/metabase_enterprise/audit_app/pages/common.clj
@@ -6,7 +6,7 @@
    [clojure.walk :as walk]
    [honey.sql :as sql]
    [honey.sql.helpers :as sql.helpers]
-   [java-time :as t]
+   [java-time.api :as t]
    [medley.core :as m]
    [metabase-enterprise.audit-app.query-processor.middleware.handle-audit-queries
     :as qp.middleware.audit]
--- a/enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
+++ b/enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
@@ -3,7 +3,7 @@
   (:require
    [buddy.sign.jwt :as jwt]
    [clojure.string :as str]
-   [java-time :as t]
+   [java-time.api :as t]
    [metabase-enterprise.sso.api.interface :as sso.i]
    [metabase-enterprise.sso.integrations.sso-settings :as sso-settings]
    [metabase-enterprise.sso.integrations.sso-utils :as sso-utils]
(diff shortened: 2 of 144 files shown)


dpsutton:
Awesome. I'm glad you verified that the namespaces were identical except for deprecation and the actual ns name. 

I want to question the no-backport decision. I think I'm in favor of backporting for two reasons: 
- seems innocuous and might as well. 
- not backporting could lead to annoying merge conflicts. Changes to ns forms which aren't backporting often have a way of causing large changesets from not backporting smoothly due to basically dumb conflicts.

I'm in favor of backporting unless there was a good reason to not backport this.

johnswanson:
> I want to question the no-backport decision. I think I'm in favor of backporting for two reasons:
> 
> * seems innocuous and might as well.
> * not backporting could lead to annoying merge conflicts. Changes to ns forms which aren't backporting often have a way of causing large changesets from not backporting smoothly due to basically dumb conflicts.
> 
> I'm in favor of backporting unless there was a good reason to not backport this.

Awesome, that makes sense to me - and thanks for spelling this out, it's illuminating to see how you think about the backport/no-backport decision.

dpsutton (approved):

PR #34547, opened 2023-10-10
[MLv2] Support replacing segments

### LINKED ISSUE #34511: [MLv2] BE - Support Segments
blocks https://github.com/metabase/metabase/issues/34495

[Slack Discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1696887407204409)

Currently `expression_clause` has no special handling for segments. I *think* what it needs to is, given a call like this:

`expression_clause('segment', [segmentObject])`

it should pull out the id from the object and make an mbql expression clause like this:

`['segment', 29]`

might make sense to have something totally separate? like `segment_clause(segment)` ? I don't care that much, but it would be nice to make explicit that segments are filters in the API.

snoe:
Fixes #34511

### DIFF
--- a/src/metabase/lib/common.cljc
+++ b/src/metabase/lib/common.cljc
@@ -48,6 +48,10 @@
   [metric-def]
   (lib.ref/ref metric-def))
 
+(defmethod ->op-arg :metadata/segment
+  [segment-def]
+  (lib.ref/ref segment-def))
+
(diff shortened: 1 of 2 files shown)


camsaul (approved):

iethree:
Tested this out in my frontend working branch, works as expected. 🚀  ❤️  👍

PR #34550, opened 2023-10-10
Improve release commit check

iethree:
### Description

Currently we run a check to determine whether the commit we are releasing is in any release branch, so we could easily accidentlly release a commit from the `release-x.46.x` branch as version 49.5. 😱 

This updates the check to first find the correct release branch for the given version number, then verifies that the given commit is in that specific release branch.

## Tests (in fork)
- [x] [release a 44 build from a 44 commit](https://github.com/iethree/metabase/actions/runs/6488567754/job/17621253837)
- [x] [try to release a 44 build from a 47 commit](https://github.com/iethree/metabase/actions/runs/6488603635/job/17621351216)

### DIFF
--- a/.github/workflows/build-for-release.yml
+++ b/.github/workflows/build-for-release.yml
@@ -78,29 +78,28 @@ jobs:
       uses: actions/checkout@v3
       with:
         fetch-depth: 0  # IMPORTANT! to get all the branches
+        sparse-checkout: release
     - name: Prepare build scripts
       run: cd ${{ github.workspace }}/release && yarn && yarn build
-    - name: Get Release Version
+    - name: Get Release Branch
       uses: actions/github-script@v6
       id: release_branch
       with:
         result_encoding: string
         script: | # js
           const { getReleaseBranch } = require('${{ github.workspace }}/release/dist/index.cjs');
-
           const version = '${{ inputs.version }}';
-
           const releaseBranch = getReleaseBranch(version);
           return releaseBranch;
-
     - name: Ensure that the specified commit exists in the ${{ steps.release_branch.outputs.result }} release branch
       run: |
+        git checkout ${{ steps.release_branch.outputs.result }}
         git branch --contains ${{ inputs.commit }} | grep -q ${{ steps.release_branch.outputs.result }} \
           && echo "Commit found in correct release branch" \
           || (echo "Commit not found in correct release branch" && exit 1)
 
   build-uberjar-for-release:
-    needs: check-version
+    needs: [check-version, check-commit]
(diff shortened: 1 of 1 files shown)


.github/workflows/build-for-release.yml
```diff
@@ -78,30 +78,28 @@ jobs:
...
-          exit 0
-        else
-          echo "Commit $COMMIT is not found in a single release branch"
-          echo "ABORT!."
-          exit -1
-        fi
+        git checkout ${{ steps.release_branch.outputs.result }}
+        git branch --contains ${{ inputs.commit }} | grep -q ${{ steps.release_branch.outputs.result }} \
+          && echo "Commit found in correct release branch" \
+          || (echo "Commit not found in correct release branch" && exit 1)
```
nemanjaglumac, line 99: Nit:

I know you find this pattern weird, but you can have `${{ steps.release_branch.outputs.result }}` stored as `env` for this `run`.

```yml
run: echo $BRANCH
env:
  BRANCH: ${{ steps.release_branch.outputs.result }}
```

.github/workflows/build-for-release.yml
```diff
@@ -78,30 +78,28 @@ jobs:
       uses: actions/checkout@v3
       with:
         fetch-depth: 0  # IMPORTANT! to get all the branches
-    - name: Ensure that the specified commit exists in the latest release branch
-      env:
-        COMMIT: ${{ inputs.commit }}
+        sparse-checkout: release
```
nemanjaglumac, line 81: Just to keep this in mind:
> THIS COMMAND IS EXPERIMENTAL. ITS BEHAVIOR, AND THE BEHAVIOR OF OTHER COMMANDS IN THE PRESENCE OF SPARSE-CHECKOUTS, WILL LIKELY CHANGE IN THE FUTURE.[^1]

[^1]: https://www.git-scm.com/docs/git-sparse-checkout#_description

nemanjaglumac (approved):

.github/workflows/build-for-release.yml
iethree: oh interesting, I didn't realize that. it makes checkout go MUCH faster. like 30s --> 5s

.github/workflows/build-for-release.yml
iethree: that makes sense here, I just prefer the `env` before the `run`. will update

iethree: interesting. using `env` added extra quotes and cause failures. but if I set it in bash it works just fine.
Screen Shot 2023-10-12 at 7 47 37 AM

PR #34553, opened 2023-10-11
Clean up yarn scripts

nemanjaglumac:
Remove `yarn &&` in front of every single script invocation. Libraries should already be installed by the time one runs any of these scripts.

`yarn` is the equivalent of `yarn install`.
This resulted in multiple install attempts and noise in the logs.


oisincoveney (approved):
This makes sense and I love it

Issue #27982, opened 2023-01-31
Support filtering in global search results


alxnddr:
[Product doc](https://www.notion.so/metabase/Support-filtering-in-global-search-results-e7d81105e9ad4443bad3a05670be8563)

Status: Needs another look since the spec as been adjusted based on engineering feedback.

dpsutton:
Milestone 2 has "Params: created_from : created_to and edited_from : edited_to". I was thinking these should be combined into a single date field. We already have a very expressive shorthand for date ranges in parameters and I think we can get a _lot_ of date handling for free by re-using these. Examples:

```clojure
dates-test=> (params.dates/date-string->range "Q1-2016")
{:start "2016-01-01", :end "2016-03-31"}
dates-test=> (params.dates/date-string->range "past3days")
{:start "2023-01-30", :end "2023-02-01"}
dates-test=> (params.dates/date-string->range "thisday")
{:start "2023-02-02", :end "2023-02-02"}
dates-test=> (params.dates/date-string->range "2016-04-18~2016-04-23")
{:start "2016-04-18", :end "2016-04-23"}
```
This gives us a lot of power. "thisday", "thismonth", etc and we already have some widgets to create these. For the BE, these all get translated to start and end dates which feed into the queries. I think this might be better than having two separate `created_from` and `created_to` dates which we have to figure out for concepts like "the past 3 days".

cdeweyx:
Note: We are removing "Edited by" / "Edited at" from scope for now. There's a good bit of complexity to untangle here and it likely makes most sense as a follow-up improvement. [More context](https://www.notion.so/metabase/Support-filtering-in-global-search-results-e7d81105e9ad4443bad3a05670be8563?pvs=4#15d01e1eda0d4f6ca8f593207821a5d3).

qnkhuat:
> Note: We are removing "Edited by" / "Edited at" from scope for now. There's a good bit of complexity to untangle here and it likely makes most sense as a follow-up improvement. [More context](https://www.notion.so/metabase/Support-filtering-in-global-search-results-e7d81105e9ad4443bad3a05670be8563?pvs=4#15d01e1eda0d4f6ca8f593207821a5d3).

We has decided that `last_edited_by` and `last_edited_at` is more useful and is feasible, so we'll add them.

PR #34554, opened 2023-10-11
Search Milestone 5: Add `created_at` search filter

oisincoveney:
`created_at` filter for Search Milestone 5: #27982 

### Description

Adds `created_at` filter to the search sidebar. The code uses the `DatePicker` component we already have to allow users to make date selections.

### How to verify

Go to the search sidebar, and try to filter by `Creation date`. You should only be able to see assets that were created in that date range. For example, create a Question, then filter the search by `Today` and you should see that question.

After filtering, try to refresh the page or paste the URL into a new window to test that hydrating the search filters still works when entering the search directly from a URL. The results should remain the same and should behave in the same way.

### Tests

- [X] Tests have been added for all components


frontend/src/metabase/search/components/SearchFilterDatePicker/SearchFilterDatePicker.styled.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from "@emotion/styled";
+import { DateAllOptionsWidget } from "metabase/components/DateAllOptionsWidget";
+
+export const SearchFilterDatePickerWrapper = styled(DateAllOptionsWidget)`
```
ranquild, line 4: Please don't change or use dashboard components here. Could you make it work using `DatePicker` only?

ranquild (changes requested):
Dashboard components are changed, this is not expected. Please use `DatePicker` only

ranquild (approved):

PR #34556, opened 2023-10-11
Search Milestone 5: Add `last_edited_at` search filter

oisincoveney:
`last_edited_at` filter for Search Milestone 5: #27982 

### Description

Adds `last_edited_at` filter to the search sidebar. The code uses the `DatePicker` component we already have to allow users to make date selections.

### How to verify

Go to the search sidebar, and try to filter by `Last edited at`. You should only be able to see assets that were last edited (by any user, unless `last_edited_by` is set). 

For a test case:

* Create/login to another user, and create a question.
* Go back to your own user, filter the search by `Today`
* You should see the question you created from the other user.

After filtering, try to refresh the page or paste the URL into a new window to test that hydrating the search filters still works when entering the search directly from a URL. The results should remain the same and should behave in the same way.

### Tests

- [X] Tests have been added for all components


ranquild (commented):
Please ask for review when the base PR is fixed

ranquild (approved):

PR #34567, opened 2023-10-11
Sort scripts in package.json alphabetically

kamilmielnik:
As discussed during today's frontend guild meeting.

I used https://codeshack.io/json-sorter/.

----

I could not use [prettier-plugin-sort-json](https://github.com/Gudahtt/prettier-plugin-sort-json) because:

> Note that this will not sort `package.json`, `package-lock.json`, or `composer.json`. This plugin only affects the `json` parser used by Prettier. Prettier uses an alternate parser (`json-stringify`) for those three specific files ([See here for details](https://github.com/prettier/prettier/blob/9a8b579d368db99394ab9da114cc37ba772fc887/src/language-js/index.js#L80)).


frontend/src/metabase/common/hooks/use-entity-list-query/use-entity-list-query.unit.spec.tsx
```diff
@@ -165,6 +165,8 @@ describe("useEntityListQuery", () => {
     await waitFor(() => {
       expect(fetchMock.calls("path:/api/table")).toHaveLength(2);
     });
+    await waitForLoaderToBeRemoved();
```
kamilmielnik, line 168: Sneaking a flake fix: https://github.com/metabase/metabase/actions/runs/6482983397/job/17603603367?pr=34567

kulyk (approved):

PR #34569, opened 2023-10-11
Intersect filter widget and column operators

kulyk:
Refactors filter picker components, so:

1. Every picker defines an explicit list of supported operators
2. A list item contains relevant metadata (e.g. `valueCount` and other options)
3. Every picker component intersects available column operators with widget operators


frontend/src/metabase/common/components/FilterPicker/NumberFilterPicker/constants.ts
```diff
@@ -1,13 +1,50 @@
...
-  "not-null": 0,
-  ">": 1,
-  ">=": 1,
-  "<": 1,
-  "<=": 1,
-  between: 2,
-  "=": "multiple",
-  "!=": "multiple",
-};
+export const OPTIONS: Option[] = [
```
iethree, line 4: maybe `operatorOptions` ?

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/CoordinateFilterPicker.tsx
```diff
@@ -1,55 +1,62 @@
...
+      option => option.operator === operatorName,
+    );
+    return option?.valueCount ?? 0;
+  }, [availableOperators, operatorName]);
+
+  const isFilterValid = useMemo(() => {
+    return Number.isFinite(valueCount)
+      ? values.length === valueCount
+      : values.length >= 1;
+  }, [values, valueCount]);
```
iethree, line 58: Two nits:
- Not sure this needs to be memoized, the useMemo function overhead might outweight this simple check
- for multi-value inputs (like between), it marks them as valid even if they only have 1 value, I think it's cuz the value is `[17, undefined]` which has length 2. For numbers we should filter to numeric values. (this bug was present previously too)

frontend/src/metabase/common/components/FilterPicker/NumberFilterPicker/NumberFilterPicker.tsx
```diff
@@ -1,47 +1,57 @@
...
+      option => option.operator === operatorName,
     );
+    return option?.valueCount ?? 0;
+  }, [availableOperators, operatorName]);
 
-  const [values, setValues] = useState<number[]>(filterParts?.values ?? []);
+  const isFilterValid = useMemo(() => {
+    return Number.isFinite(valueCount)
+      ? values.length === valueCount
+      : values.length >= 1;
```
iethree, line 53: nit: I would break this out into a separate function, but no big deal.

iethree (approved):
I really like this approach. Nice job ❤️ 

I few nits/suggestions, I think the only important one is using strings from CLJS, I hate to have multiple sources of truth for those.

ranquild (approved):
🔥

frontend/src/metabase/common/components/FilterPicker/StringFilterPicker/constants.ts
```diff
@@ -1,14 +1,49 @@
...
-  "starts-with": 1,
-  "ends-with": 1,
-  contains: 1,
-  "does-not-contain": 1,
-  "=": "multiple",
-  "!=": "multiple",
-};
+export const OPTIONS: Option[] = [
+  {
+    name: t`Equal to`,
```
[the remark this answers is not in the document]
kulyk: Yeah, we should be relying on MLv2 names here 👍

frontend/src/metabase/common/components/FilterPicker/StringFilterPicker/constants.ts
```diff
@@ -1,14 +1,49 @@
...
+    operator: "=",
+    valueCount: Infinity,
```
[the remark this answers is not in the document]
kulyk: Just rewatched the first Toy Story recently, haha

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/CoordinateFilterPicker.tsx
kulyk: > Not sure this needs to be memoized, the useMemo function overhead might outweight this simple check

Yeah, good point 👍 

> for multi-value inputs (like between), it marks them as valid even if they only have 1 value, I think it's cuz the value is [17, undefined] which has length 2. For numbers we should filter to numeric values. (this bug was present previously too)

OH, nice catch! Will fix this, thank you

PR #34571, opened 2023-10-11
Migrate storybook to a subfolder

uladzimirdev:
## Description

As were discussed in [RFC](https://www.notion.so/metabase/Upgrade-storybook-to-7-28a72bc8ec294356944d3accf76f5299), we first try to migrate storybook away from the main app. 

I used `/storybook` to keep all it's dependencies. Stories still live next to components.
Commands still live in the root package.json: `build-storybook` and `storybook` proxy commands to the `storybook/package.json`, make sure you didn't forget to run `yarn` in the `storybook` folder.

Chromatic workflow is updated.
type-check workflow now requires storybook node_modules to present as we check types of stories.
eslint now doesn't try to resolve `@storybook/*` packages

Chromatic build https://61d4abcd959e8c003aa51d4c-owpmswddyw.chromatic.com/?path=/story/intro--page


frontend/src/metabase/.eslintrc
```diff
@@ -32,7 +32,8 @@
     {
       "files": ["**/*.stories.tsx"],
       "rules": {
-        "import/no-default-export": 0
+        "import/no-default-export": 0,
+        "import/no-unresolved": [2, {"ignore": ["@storybook"]}]
```
uladzimirdev, line 36: allow unresolved imports to `@storybook/something` in stories. Stories are located near components, but storybook is a part of different package

frontend/src/metabase/core/components/FormFileInput/FormFileInput.stories.tsx
```diff
@@ -2,7 +2,6 @@ import type { ComponentStory } from "@storybook/react";
 import { Form, FormProvider } from "metabase/forms";
 import FormFileInput from "./FormFileInput";
 
-// eslint-disable-next-line import/no-default-export -- deprecated usage
```
uladzimirdev, line 5: This story had a typo in the name

jest.config.json
```diff
@@ -10,7 +10,7 @@
     "\\.svg\\?(component|source)": "<rootDir>/frontend/test/__mocks__/svgMock.jsx"
   },
   "transformIgnorePatterns": [
-    "<rootDir>/node_modules/(?!rehype-external-links/)"
+    "<rootDir>/node_modules/(?!(rehype-external-links|unist-util-visit|unist-util-visit-parents|space-separated-tokens|is-absolute-url)/)"
```
uladzimirdev, line 13: packages were upgraded after removal of storybook from package.json, they all are in ESM, so we have to transpile them

package.json
```diff
@@ -302,7 +293,7 @@
     "build-static-viz:watch": "webpack --config webpack.static-viz.config.js --watch",
     "build-static-viz:watch-wait": "yarn wait:cljs; yarn build-static-viz:watch",
     "build-stats": "yarn && webpack --json > stats.json",
-    "build-storybook": "yarn build:cljs && NODE_OPTIONS=--max-old-space-size=8196 build-storybook",
```
uladzimirdev, line 305: scripts are defined in the `storybook` folder package.json, so I we explicitly say it should be started there

storybook/.storybook/main.js
```diff
@@ -29,7 +34,11 @@ module.exports = {
       ],
     },
     resolve: {
-      ...storybookConfig.resolve,
+      ...webpackConfig.resolve,
+      modules: [
+        ...webpackConfig.resolve.modules,
```
uladzimirdev, line 39: adding `storybook/node_modules` to the list of modules where webpack will look for modules

tsconfig.json
```diff
@@ -9,7 +9,8 @@
         "./enterprise/frontend/src/*",
         "./enterprise/frontend/test/*"
       ],
-      "cljs/*": ["./target/cljs_dev/*"]
+      "cljs/*": ["./target/cljs_dev/*"],
+      "@storybook/*": ["./storybook/node_modules/@storybook/*"]
```
uladzimirdev, line 13: this is needed for typecheck

storybook/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "@metabase/storybook",
+  "private": true,
+  "version": "1.0.0",
+  "description": "",
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "scripts": {
+    "start": "start-storybook -p 6006",
+    "build": "NODE_OPTIONS=--max-old-space-size=8196 build-storybook -o ../storybook-static"
```
uladzimirdev, line 11: never change `-o` to something like `../`, it's `rm -rf`

storybook/.storybook/main.js
```diff
@@ -1,26 +1,31 @@
...
   ],
   addons: [
     "@storybook/addon-essentials",
     "@storybook/addon-links",
     "@storybook/addon-a11y",
   ],
-  webpackFinal: storybookConfig => ({
-    ...storybookConfig,
-    plugins: [...storybookConfig.plugins, new MiniCssExtractPlugin()],
+  babel: () => ({
```
uladzimirdev, line 18: added our babel config instead of one from storybook

kamilmielnik (commented):
LGTM :+1: Left some comments inline.

I'll let others approve as I'm not sure about the workflows changes (do we need them?).

storybook/package.json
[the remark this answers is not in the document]
uladzimirdev: I think this way chromatic expects storybook to be built, but I'm not 100% sure, I tried to keep the old behaviour

nemanjaglumac:
Thank you for doing this.
I'll remove myself from the reviewers because I don't know enough about the Storybook ecosystem.

ranquild (changes requested):
This approach considerably increases complexity of our project setup. I believe we should upgrade our mdx stories instead.

PR #34579, opened 2023-10-11
Prune deleted in-memory databases on startup (in dev)

### LINKED ISSUE #9962: We should delete sync tasks for DBs that no longer exist when we try to run them
We usually delete scheduled tasks for a DB when you delete the DB but if any are still hanging around for whatever reason the tasks still exist in the DB forever. The tasks themselves will no-op but it makes debugging things on my local instance more annoying since I am constantly creating and deleting test DBs

johnswanson:
### Description

Thanks to @dpsutton for the help here! I investigated an alternative approach of reusing the machinery that we use in tests to determine whether the database already exists, but that turned out to rely on a private atom that stored a list of databases created by the tests. I didn't want to muck around with those internals, and it turned out to be simpler than expected to make `can-connect?` work (although I did need to catch the exception thrown when we couldn't connect, which seems a little odd - I would have expected `can-connect?` to catch that and return `false`).

Fixes https://github.com/metabase/metabase/issues/9962

### How to verify

As this is development-only, I didn't add tests for the new behavior, but it's pretty simple. I did test it out in the REPL by verifying that the dangling in-memory databases I had were deleted after running `prune-deleted-inmem-databases!`.

### DIFF
--- a/dev/src/dev.clj
+++ b/dev/src/dev.clj
@@ -2,6 +2,7 @@
   "Put everything needed for REPL development within easy reach"
   (:require
    [clojure.core.async :as a]
+   [clojure.string :as str]
    [dev.debug-qp :as debug-qp]
    [dev.model-tracking :as model-tracking]
    [honeysql.core :as hsql]
@@ -24,6 +25,7 @@
    [metabase.test :as mt]
    [metabase.test.data.impl :as data.impl]
    [metabase.util :as u]
+   [metabase.util.log :as log]
    [methodical.core :as methodical]
    [potemkin :as p]
    [toucan2.connection :as t2.connection]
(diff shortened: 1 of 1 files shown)


iethree:
tip: for backend prs, tag @metabase/core-backend-pixel-police , we frontend devs are not gonna be too helpful here 👍

johnswanson:
> tip: for backend prs, tag @metabase/core-backend-pixel-police , we frontend devs are not gonna be too helpful here 👍

Oops, my bad - sorry about that!

dev/src/dev.clj
```diff
@@ -58,10 +59,28 @@
...
+                         (try
+                           (driver/can-connect? :h2 (:details db))
+                           (catch org.h2.jdbc.JdbcSQLNonTransientConnectionException _
+                               false))))]
+    (remove can-connect? (filter in-memory? h2-dbs))))
+
+(defn prune-deleted-inmem-databases!
+  []
+  (doseq [db (deleted-inmem-databases)]
+    (t2/delete! :model/Database :id (:id db))))
```
dpsutton, line 77: We can delete them all in one delete statement:

```clojure
(when-let [outdated-inmemory (seq (deleted-inmem-databases))]
  (t2/delete! :model/Database :id [:in outdated-inmemory]))
```

dev/src/dev.clj
```diff
@@ -58,10 +59,28 @@
...
 
+(defn- deleted-inmem-databases
+  []
+  (let [h2-dbs (t2/select :model/Database :engine :h2)
+        in-memory? (fn [db] (some-> db :details :db (str/starts-with? "mem:")))
+        can-connect? (fn [db]
+                       (binding [metabase.driver.h2/*allow-testing-h2-connections* true]
+                         (try
+                           (driver/can-connect? :h2 (:details db))
+                           (catch org.h2.jdbc.JdbcSQLNonTransientConnectionException _
```
dpsutton, line 70: I'm a bit hesitant to catch such a specific error here. This is in general just "annoying" and not super bad. I think we should catch all exceptions and perhaps not remove those dbs. I'm just thinking if a check throws an error besides this one, we won't continue startup and just end up hanging here.

dpsutton:
One other thing. these functions are simple but it's good practice to have some docstrings on them. It's clear _what_ they are doing but a hint as to why would be helpful.

dev/src/dev.clj
johnswanson: Good point, that makes sense.

dev/src/dev.clj
johnswanson: Very nice!

dev/src/dev.clj
dpsutton: er, I think it's awesome you are catching that specific error and removing it. But we should catch others and not remove those. I realized I worded that a bit poorly

johnswanson: Nah it made sense to me. Something like:

```clojure
(catch Exception e
  (log/error e "Error connecting to inmem database")
  ;; we don't want to delete these, so just pretend we could connect
  true)
```

along the lines of what you're thinking?

dpsutton:
```clojure
(try (driver/can-connect? :h2 db)
     (catch org.h2.jdbc.JdbcSQLNonTransientConnectionException _
         false)
     (catch Exception e
       (log/error "Error checking in-memory database for deletion" e)
       true))
```

something like this. I think it's valid for a db to throw an error if it cannot connect:

> This function should return truthy if a connection to the DB
  can be made successfully, otherwise it should return falsey or throw an appropriate Exception. Exceptions if a
  connection cannot be made. Throw an `ex-info` containing a truthy `::can-connect-message?` in `ex-data`
  in order to suppress logging expected driver validation messages during setup.

If this is throwing a NonTransient Connection Exception that sounds like there's no chance of it working in the future and it is validly gone. Other exceptions, who knows. So we'll leave those and we can manually clean them up.

dpsutton (approved):

escherize (approved):

PR #34583, opened 2023-10-11
Determine `pivot-cols` and `pivot-rows` on the backend using viz settings

### LINKED ISSUE #34278: Pivot Table Viz shows different output and calls different endpoints when saved VS being created
### Describe the bug

When creating a Pivot Table it calls different endpoints when saved VS being created ... These endpoints also load different kinds of data:

Loading a PIVOT after you saved it loads this amount of data:

image

While recreating the same question before saving it only loads 1/4 of the data:

image

### To Reproduce

1. Go to New Questions -> Analytics -> Summarise by Count and group by all available columns:

image

2. Change Visualisation -> Pivot Table -> Set it up as follows and notice that it returns `100,912` rows

image

Open the network developer tools and you can see a call is being sent to `/api/dataset/pivot`

image

3. Now save the question and refresh the question. You will notice it takes much **more time** and  it returns `436,561` rows

image

   This time you will notice a call is being sent to `/api/card/pivot/12/query`

### Expected behavior

For once we should get similar results & second the time should match. Since showing a fast pivot when you are creating it and a very slow one after you save it and add it to the dashboard is very confusing 

### Logs

None that are relevant

```JSON
1.47.2 and same thing happens for master even though the output gives different row amount for the /pivot endpoint
```

### Severity

Major issue for people using Pivot Tables since the two endpoints are giving different results

### Additional context

_No response_

noahmoss:
Pivot tables execution requires `pivot-cols` and `pivot-rows` vectors, which indicate the indices of the breakouts to use as cols and rows for the pivot table. These are used in `generate-queries` to determine which subqueries need to be run to assemble the pivot table with the appropriate subtotal & grand total values.

For ad-hoc queries, these values are passed in by the frontend as part of the query object itself. For saved cards, however, these values were not being passed consistently. In some cases (normal card execution, embeds) they were being passed in the API request body but not being looked at by the backend. For public cards, they weren't being passed at all. However: this info can also be determined by looking at `:pivot_table.column_split` in a card's viz settings, which the backend already has access to for all types of card execution.

This PR modifies the pivot table execution code to pull this information from card's viz settings if it's not already in the query itself. I've also applied the default constraints for row limits for ad-hoc queries, which were already being applied for saved cards. These changes together result in consistent pivot table results across all of the query processing entry points.

Resolves https://github.com/metabase/metabase/issues/34278

### DIFF
--- a/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
@@ -9,6 +9,7 @@ import {
   dragField,
   leftSidebar,
   main,
+  modal,
 } from "e2e/support/helpers";
 
 import { SAMPLE_DB_ID } from "e2e/support/cypress_data";
@@ -59,6 +60,7 @@ describe("scenarios > visualizations > pivot tables", { tags: "@slow" }, () => {
   beforeEach(() => {
     restore();
     cy.signInAsAdmin();
+    cy.intercept("POST", "/api/card").as("createCard");
   });
 
   it("should be created from an ad-hoc question", () => {
@@ -1150,6 +1152,21 @@ describe("scenarios > visualizations > pivot tables", { tags: "@slow" }, () => {
         },
       },
     });
+
+    cy.findByTestId("question-row-count").should(
+      "have.text",
+      "Showing first 52,711 rows",
+    );
+
+    cy.findByTestId("qb-header-action-panel").findByText("Save").click();
+    modal().button("Save").click();
+    cy.wait("@createCard");
+    cy.reload();
+
+    cy.findByTestId("question-row-count").should(
+      "have.text",
+      "Showing first 52,711 rows",
+    );
   });
 });
 
--- a/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
@@ -1152,6 +1152,7 @@ describe("scenarios > visualizations > pivot tables", { tags: "@slow" }, () => {
         },
       },
     });
+    cy.wait("@pivotDataset");
 
     cy.findByTestId("question-row-count").should(
       "have.text",
--- a/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
@@ -22,6 +22,8 @@ const {
   PEOPLE,
   REVIEWS,
   REVIEWS_ID,
+  ANALYTIC_EVENTS,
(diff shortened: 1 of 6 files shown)


src/metabase/query_processor/pivot.clj
```diff
@@ -211,6 +214,22 @@
       (qp/process-query-and-save-with-max-results-constraints! first-query info context)
       (qp/process-query (dissoc first-query :info) context))))
 
+(defn pivot-options
+  "Given a pivot table query and a card ID, looks at the `pivot_table.column_split` key in the card's visualization
+  settings and generates pivot-rows and pivot-cols to use for generating subqueries."
+  [query card-id]
+  (when card-id
+    (let [card              (t2/select-one :model/Card :id card-id)
```
camsaul, line 222: I'd like this better if the API endpoints in question passed in the pivot options and the `qp.pivot` namespace didn't have to worry about where they came from, there's a reason Toucan is discouraged in `metabase.query-processor.*` namespaces, and that is we're trying to keep the app DB code separate from QP code as much as possible.

It makes it much easier to test the pivot code if you can pass in the options without having to create a temp Card to do so

camsaul (changes requested):
Hmmm, I would strongly prefer not having the pivot QP namespace hit the app DB directly, maybe an ok alternative here is to just include the entire viz settings map inside the `info` passed around (the place you're getting `:card-id` from. Then you can just update the `metabase.query-processor.card` namespace to include that when you use `run-query-for-card-async` so we don't need to update a bunch of different API endpoints.

That way all the QP code that deals with Cards in the application DB is centralized in one place.

noahmoss:
@camsaul Makes sense. I've updated it to pass the card's viz settings in the `info` map. I do tend to wonder whether it would be worth passing the entire `Card` into the QP, since we're already pulling out several keys and sticking them in the `info` map? But that would be a bigger refactor.

e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js
```diff
@@ -1098,6 +1102,72 @@ describe("scenarios > visualizations > pivot tables", { tags: "@slow" }, () => {
...
+    });
+
+    cy.findByTestId("question-row-count").should(
+      "have.text",
+      "Showing first 52,711 rows",
+    );
+
+    cy.findByTestId("qb-header-action-panel").findByText("Save").click();
+    modal().button("Save").click();
+    cy.wait("@createCard");
```
alxnddr, line 1163: In other specs in this file we tend to wait for datasets requests to be resolved using `cy.wait(...)` probably to avoid flakes so it looks reasonable to wait for both adhoc and saved dataset requests in this spec

camsaul (approved):
backend stuff looks good to me

Issue #33411, opened 2023-08-22
[Epic] Trend chart improvements


luizarakaki:
## Problem
Trend charts currently have several limitations on their usefulness
- They can only compare to the immediate previous period - previous day, month, quarter, etc. 
  - Trend charts do not skip over gaps in the data: if the immediate previous period is missing or contains a null-value, no comparison is displayed
- They can only display one comparison at a time
  - No option for a static comparison - e.g. to measure your trend against a target/goal
  - No option for arbitrary date comparison - i.e. # of periods ago instead of immediate previous period
  - No option for comparison with another column 
- There is no option to compact a long number

---
## Desired Behavior/Acceptance Criteria

**Milestone 1:** Update Trend Chart Copy
- [x] https://github.com/metabase/metabase/pull/35184

**Milestone 2:** Support arbitrary date comparison (FE 2)

- [x] #35551
  - Value: Instead of limiting users to selecting only the previous period, there will now be an option to select an arbitrary number of previous periods ago to compare against. So instead of comparing to last month, now we can compare to 3 months ago or 6 months ago.

- [x] #37026

**Milestone 3:** Support a static number comparison (FE 1)

- [x] #35552
  - Value: Users will now be able to select static numbers to compare the current value against. This opens up the ability to create targets/goals to strive for and measure against.

**Milestone 4:** Support comparison with another column (FE 0.5)

- [x] #35553
  - Value: Comparing a current metric to its previous value can be helpful, but so can comparing one metric to an entirely different metric. This addition will open the possibility of comparing your metric to other metrics (within the same table).

**Milestone 5:** Support value formatting (FE 0.5)

- [x] #35554
  - Value: Display long numbers sometimes provide irrelevant information. Sometimes I just want to see how many thousands or millions a value is, and I do not really care about the values in the hundreds. Well now you can choose to compact your values to show only the most important digits you care about.

**Milestone 6:** Enable multiple comparisons (FE 2)

- [x] #35555
  - Value: Often comparing a value against multiple numbers provides more context and actionable information than against only a single number. This addition will allow for multiple (up to 3) comparisons.

**Milestone X:** 

- [ ] Update search widget: "select” to match the designs of the new comparison select component
- [ ] Convert Trend Charts JavaScript files to TypeScript
- [x] Add Trend Charts variants to Storybook (and to prepare for future visual testing)

**Milestone 7:** Add new functionality to the static viz (FE 2, BE 2)

- [x] #35556
[...]

PR #34587, opened 2023-10-11
Update Trend chart design and fix copy

shaunlebron:
Part of trend chart improvements epic: https://github.com/metabase/metabase/issues/33411

This PR changes the layout and corrects the copy for our trend chart.  We can release this independently in 48.

(In the next PRs for this epic, we will add same fixes to static viz trend chart, then add for support multiple comparisons.)

## Changes
1. Remove question title and description tooltip, and enable the Dashcard header to display them instead.
3. Add period label, but hide it when card is too short to display it.
4. Progressively shorten comparison text based on card width.
5. Remove comparison wrapping via media query breakpoint since new design will always keep comparison on a single line.

## Reviewing
Hide whitespace.

## Manual testing
1. Create a trend chart grouped by different time units to see if the new period label works.
2. See if it looks good at multiple card sizes.
3. Make sure tooltips show up when anything is truncated.

https://github.com/metabase/metabase/assets/116838/a600db32-2cd9-4721-9a92-dd5afa902a9e

https://github.com/metabase/metabase/assets/116838/5e0ea77c-1d35-4600-93a8-992c449a7a3c

### DIFF
--- a/frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
+++ b/frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
@@ -3,18 +3,20 @@ import { Component } from "react";
 import { t, jt } from "ttag";
 import _ from "underscore";
 
-import { formatValue } from "metabase/lib/formatting";
+import {
+  formatDateTimeRangeWithUnit,
+  formatValue,
+} from "metabase/lib/formatting";
 import { color } from "metabase/lib/colors";
 
 import Tooltip from "metabase/core/components/Tooltip";
+import { Ellipsified } from "metabase/core/components/Ellipsified";
 
 import { columnSettings } from "metabase/visualizations/lib/settings/column";
 import { NoBreakoutError } from "metabase/visualizations/lib/errors";
 import { compactifyValue } from "metabase/visualizations/lib/scalar_utils";
-
 import ScalarValue, {
   ScalarWrapper,
-  ScalarTitle,
 } from "metabase/visualizations/components/ScalarValue";
 import * as Lib from "metabase-lib";
 import {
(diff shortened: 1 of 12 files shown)


shaunlebron:
@dhuniverse Our current design isn’t degrading well at the smallest possible sizes.  How should we adjust their layout?

previous design:
CleanShot 2023-10-16 at 17 03 22@2x

current design:
CleanShot 2023-10-16 at 17 08 17@2x

dhuniverse (approved):
Sounds good to me. Approved.

iethree:
why are all the backend tests failing on this branch, does it need a rebase?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -153,6 +172,11 @@ export class SmartScalar extends Component {
     if (!insight) {
       return null;
     }
+    const { unit } = insight;
```
alxnddr, line 175: I remember we discussed the visualization itself can infer this information from data. Why did we decide to go the other way?

alxnddr (commented):
Looks great, big improvement!

I left one question and I also noticed "vs previous month" in the dark mode looks not very readable. Not sure if this is caused by this PR but it would be nice to fix this too

Screenshot 2023-10-23 at 5 39 40 PM

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
shaunlebron: This is currently in progress for the next PR.

alxnddr (dismissed):

luizarakaki (dismissed):

darksciencebase (approved):
nominal approval as this PR seems to have been a victim of a temporary github misconfiguration

PR #34568, opened 2023-10-11
Make implicit fields coming from source query expressions use string id

lbrdnk:
Closes #28451

### Description

When implicit fields are generated for a source query and that query contains expression that is just field, field in question is generated twice and later result is discarded. To overcome this, field generated from expression is in string id form.

### How to verify

Run the new test or try reproduction steps from the issue in question.

### Checklist

- [X] Tests have been added/updated to cover changes in this PR

PR #34594, opened 2023-10-12
🍒 Manual Backport: Make implicit fields coming from source query expressions use string id

lbrdnk:
Manual backport of #34568.


nemanjaglumac (approved):
Diffs match, except [THIS](https://github.com/metabase/metabase/pull/34568/files#r1356688367) which affects master only.
In other words, the backport version is correct.

Issue #32612, opened 2023-07-24
Rework `migrate-click-through` to use `metabase.db.custom-migrations/define-migration`


camsaul:
We should get rid of the legacy `metabase.db.data-migrations/defmigration` migrations and move them over to the new `metabase.db.custom-migrations` framework.

See also #32611

Issue #32611, opened 2023-07-24
Rework `migrate-remove-admin-from-group-mapping-if-needed` to use `metabase.db.custom-migrations/define-migration`


camsaul:
We should get rid of the legacy `metabase.db.data-migrations/defmigration` migrations and move them over to the new `metabase.db.custom-migrations` framework.

PR #34400, opened 2023-10-05
Split migration

qnkhuat:
implement #34361 by splitting the current migration file into 2 files:
- 000_legacy_migrations.yaml containing migrations from the dawn of time to the last migration of 44
- 001_update_migrations.yaml containing all migrations from 45 till now
  - Added a new sqlFile migration with id `v00.00-000` to this file, this migration is used to bootstrap metabase to the state of v44. Think of this like instead of running `000_leagcy_migrations.yaml`, we use this sqlfile instead.

So how migration will play out in real life?
Scenario 1: Fresh instance
1. Migrate with the 001_update_migrations.yaml file
2. the first migration will be run is `v00.00-000`

Scenario 2: Upgrade from pre-45
1. Migrate using the 000_legacy_migrations.yaml file
2. Migrate with the 001_update_migrations.yaml file
3. migration `v00.00-000` will be skipped

Scenario 3: Upgrade from post-45
1. Migrate with the 001_update_migrations.yaml file
2. migration `v00.00-000` will be skipped

### Other changes
- I cleaned up the lint for migrations to remove the unstrict version, it was there primarily to check migrations that are old and didn't follow the "strict" rules.

### How I generated these dump SQL for each db engines:
1. I faked the db by running only migrations in `000_legacy_migrations.yaml` which created the db as of 44
5. then for each dbms, I used the following commands to export its schema dump

```bash
# postgres
pg_dump --schema-only --no-owner --exclude-table databasechangelog --exclude-table databasechangeloglock metabase > resources/initialization/metabase_postgres.sql

# mysql
mysqldump -u root --no-data --skip-comments --ignore-table=metabase.databasechangelog --ignore-table=metabase.databasechangeloglock metabase > resources/initialization/metabase_mysql.sql

# h2 

select table_name from information_schema.tables where table_schema = 'PUBLIC' and table_name not like 'DATABASECHANGELOG%';

[...]

PR #34598, opened 2023-10-12
Rework data migrations

sansolovyov:
This is about #32611 and #32612 and takes over from #33105. Old `defmigration`-using data migrations have been transferred to a new `custom-migrations` so that they can be applied by liquibase.

Code is virtually unchanged. `custom_migrations.clj` is a bit over 700 lines long, I wonder if it makes sense to split it?


resources/migrations/000_migrations.yaml
```diff
@@ -15173,6 +15173,14 @@ databaseChangeLog:
               column:
                 name: action_id
 
+  - changeSet:
+      id: v48.00-007
```
qnkhuat, line 15177: pretty sure 007 has been claimed, you need to use a different one

src/metabase/db/custom_migrations.clj
```diff
@@ -714,3 +717,176 @@
...
+  (migrate-click-through!))
+
+;;; Removing admin from group mapping migration
+
+(defn- raw-setting
+  "Get raw setting directly from DB.
+  For some reasons during data-migration [[metabase.models.setting/get]] return the default value defined in
+  [[metabase.models.setting/defsetting]] instead of value from Setting table."
+  [k]
+  (t2/select-one-fn :value setting/Setting :key (name k)))
```
qnkhuat, line 855:
```suggestion
  (t2/select-one-fn :value (t2/table-name :model/Setting) :key (name k)))
```

We should change it while we're here. Like SQL migrations, these custom migrations should not be dependent on the source code, ideally, this namespace should only require low-level stuff like `jdbc, toucan2, etc`. The idea is stuff from `metabase.models.setting` can change as we go, and hooks for `:model/Setting` could be as well, so the behavior of triggering those hooks now, can be different from if we run it 2 versions into the future. so we should avoid that.

src/metabase/db/custom_migrations.clj
```diff
@@ -714,3 +717,176 @@
...
+  {:author "dpsutton"
+   :added  "0.38.1"
+   :doc    "Migration of old 'custom drill-through' to new 'click behavior'; see #15014"}
+  []
+  (transduce (comp (map (parse-to-json :card_visualization :dashcard_visualization))
+                   (map fix-click-through)
+                   (filter :visualization_settings))
+             (completing
+              (fn [_ {:keys [id visualization_settings]}]
+                (t2/update! :model/DashboardCard id {:visualization_settings visualization_settings})))
```
qnkhuat, line 826: same here, let's see if we could use the raw dashboardcard table
```suggestion
                (t2/update! :report_dashboardcard {:visualization_settings visualization_settings})))
```

src/metabase/db/custom_migrations.clj
```diff
@@ -714,3 +717,176 @@
...
+
+(defn- remove-admin-group-from-mappings-by-setting-key!
+  [mapping-setting-key]
+  (let [admin-group-id (:id (perms-group/admin))
+        mapping        (try
+                        (json/parse-string (raw-setting mapping-setting-key))
+                        (catch Exception _e
+                          {}))]
+    (when-not (empty? mapping)
+      (t2/update! setting/Setting (name mapping-setting-key)
```
qnkhuat, line 865:
```suggestion
      (t2/update! :setting (name mapping-setting-key)
```

src/metabase/db/custom_migrations.clj
```diff
@@ -714,3 +717,176 @@
...
+(defn- raw-setting
+  "Get raw setting directly from DB.
+  For some reasons during data-migration [[metabase.models.setting/get]] return the default value defined in
+  [[metabase.models.setting/defsetting]] instead of value from Setting table."
+  [k]
+  (t2/select-one-fn :value setting/Setting :key (name k)))
+
+(defn- remove-admin-group-from-mappings-by-setting-key!
+  [mapping-setting-key]
+  (let [admin-group-id (:id (perms-group/admin))
```
qnkhuat, line 859:
```suggestion
  (let [admin-group-id (t2/select-one-pk :permissions_group :name "Administrators")
```

I'd try to not include other namespaces in this custom migration ns (for reasons please see the namespace docsting), doing this so we can remove `perms-group`.

src/metabase/cmd/copy.clj
qnkhuat, line 1: you'd need to delete the model in `entities-without-autoinc-ids` on line 312

qnkhuat: also in `metabase-enterprise.models.entity-id-test/entities-not-exported`

resources/migrations/000_migrations.yaml
qnkhuat, line 1: We can drop the `data_migrations` table as well

qnkhuat (changes requested):
Definitely on track but I think we can drop the `data_migrations` table and change it so that `custom_migrations` namespace should not require other namespaces.

resources/migrations/000_migrations.yaml
piranha: 🤦  I feel so dumb right now for not checking this

qnkhuat:
on 2nd thought, dropping will `data_migrations` require us to define a proper rollback to insert back the lost record. otherwise, anyone who `upgrade => downgrade 2 versions => upgrade 1 version` might get data migrations triggered again.

at this point, it's probably not worth complicating the matter and leaving the table there till later releases. wdyt?

resources/migrations/000_migrations.yaml
```diff
@@ -15173,6 +15173,14 @@ databaseChangeLog:
...
+      author: johnswanson
+      comment: Migrate migrate-click-through to a custom migration
```
qnkhuat, line 15179: also, we need a pre-condition to only run this if the data migration has not been run before.

piranha:
@qnkhuat all comments are handled in the last commit rather then augmented in previous ones :)

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1014,3 +1020,445 @@
...
+  (testing "that rollback causes data_migrations table to reappear with data"
+    (impl/test-migrations ["v48.00-024"] [_]
+      (let [{:keys [db-type ^javax.sql.DataSource
+                    data-source]} mdb.connection/*application-db*
+            ;; approach from schema-migrations-test/migrate-field-database-type-test
+            migrate!              (partial db.setup/migrate! db-type data-source)]
+        ;; 0 because we removed them and normal path won't trigger any
+        (is (<= 0 (:cnt (t2/query-one {:from   [:data_migrations]
+                                       :select [[[:count :id] :cnt]]}))))
+        (migrate! :up)
```
qnkhuat, line 1457: this will migrate up all the way to the latest migration, so if we're developing version 52 for example, it'll migrate up to 52.

it's not needed for this test, ideally we should only migrate up to the version that this test is testing against `"v48.00-024"`.

this is to avoid having to update this test in the future, in cases future migrations will change things that this test touch.

to do this you can use the `migrate!` fn binding returned from `impl/test-migrations ["v48.00-024"] [_]`

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1014,3 +1020,445 @@
...
+                                      {:click_behavior
+                                       {:type             "link",
+                                        :linkType         "url",
+                                        :linkTemplate     "http://example.com//{{id}}",
+                                        :linkTextTemplate "here is my id: {{id}}"}}}}]
+              (f)
+              (is (= expected-settings
+                     (-> (t2/select-one DashboardCard :id dashcard-id)
+                         :visualization_settings))))))]
+    (testing "Running the migration from scratch"
```
qnkhuat, line 1365: could you add a test case for to check if the data migration has been ran previously, we don't run this migration?

same for the other migration.

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1014,3 +1020,445 @@
...
+        (expect-correct-settings! migrate!)))
+    (testing "Running the migration after a previous data-migration still works"
+      (impl/test-migrations ["v48.00-022"] [migrate!]
+        (expect-correct-settings! (fn []
+                                    (#'custom-migrations/migrate-click-through!)
+                                    (migrate!)))))))
+
+(defn- get-json-setting
+  [setting-k]
+  (json/parse-string (t2/select-one-fn :value setting/Setting :key (name setting-k))))
```
qnkhuat, line 1376:
```suggestion
  (json/parse-string (t2/select-one-fn :value :setting :key (name setting-k))))
```

I think we should prefer the raw table here

src/metabase/db/custom_migrations.clj
```diff
@@ -714,3 +715,177 @@
...
+                          {}))]
+    (when-not (empty? mapping)
+      (t2/update! :setting {:key (name mapping-setting-key)}
+                  {:value
+                   (->> mapping
+                        (map (fn [[k v]] [k (filter #(not= admin-group-id %) v)]))
+                        (into {})
+                        json/generate-string)}))))
+
+(defn migrate-remove-admin-from-group-mapping-if-needed
```
qnkhuat, line 871: we probably want to make this fn private

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1014,3 +1020,445 @@
...
+  (binding [setting/*allow-retired-setting-names* true]
+    `(call-with-ldap-and-sso-configured ~ldap-group-mappings ~sso-group-mappings (fn [] ~@body))))
+
+;; The `remove-admin-from-group-mapping-if-needed` migration is written to run in OSS version
+;; even though it might make changes to some enterprise-only settings.
+;; In order to write tests that runs in both OSS and EE, we can't use
+;; [[metabase.models.setting/get]] and [[metabase.test.util/with-temporary-setting-values]]
+;; because they require all settings are defined.
+;; That's why we use a set of helper functions that get setting directly from DB during tests
+(deftest migrate-remove-admin-from-group-mapping-if-needed-test
```
qnkhuat, line 1400: nit: on optimizing test speed time, I wonder if we should consider this as `behavior` test just call `migrate-remove-admin-from-group-mapping-if-needed` directly. instead of doing the setup with `test-migrations` to not have to set up the DB 4 times here.

And instead, we should write a test for upgrading scenarios like: if already executed, this migration should be skipped etc.

This is just a suggestion, feel free to skip it if it doesn't make much sense to you.

test/metabase/db/custom_migrations_test.clj
piranha: Yeah ideally I'd love to migrate only a single migration down to check that rollback works, but `migrate!` from `impl/test-migrations` does not accept arguments. I've looked into making it accept arguments, but then `run-migrations-in-range!` does not want to go down, and I'm not sure why, so I stopped there. :) I can use `migrate!` to go up though, of course.

qnkhuat: yea, what people currently do is use the `migrate!` from `test-migrations` macro do the up migrate, then manually calling `db.setup/migrate! :down` and I think it's fine.


FYI in this PR I'm making changes so that `migrate!` from `test-migrations` can do downgrade as well https://github.com/metabase/metabase/pull/34400/files#diff-7841c3c22e27fd535d5d29dc3617f74311ccc74d609b9636e78aae91d41819a5R200-R201

resources/migrations/000_migrations.yaml
```diff
@@ -15350,6 +15350,92 @@ databaseChangeLog:
...
+      preConditions:
+        - onFail: MARK_RAN
+        - sqlCheck:
+            expectedResult: 0
+            sql: SELECT count(*) FROM data_migrations WHERE id = 'migrate-remove-admin-from-group-mapping-if-needed';
+      changes:
+        - customChange:
+            class: "metabase.db.custom_migrations.MigrateRemoveAdminFromGroupMappingIfNeeded"
+
+  - changeSet:
```
qnkhuat, line 15379: btw, I'm thinking about adding to the lint check to make sure only one change is allowed in the rollback too.

Do you think you can break it into 2 here? also maybe we need to check with BE core to see what's the reason behind the rule of one change set per migration only.

piranha: Okay let me ask then about this rule first and then let's decide on what to do with splitting. I'm not a fan of splitting just because this feels like it should be a single change rather than two distinct changes, but let me first find out the reasons behind the rule. :)

piranha: Okay so I've [read liquibase best practices](https://www.liquibase.org/get-started/best-practices#tabbing-block_6168671689f27) about a single change in a changeset and I'm not convinced it's a good idea to split this rollback into two transactions... 

I mean technically we could, but logically 1 change drops table, the other one returns it back, sounds like a good way to structure changes?

qnkhuat: yea well, it's good practice, not rules so I think as long as we don't do anything crazy it's fine.

but I have an idea if you really want to have just one change you can turn the create table to SQL and have it in one single change set with the insert.

That's way the lint don't have to handle  exceptions

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1033,3 +1039,444 @@
...
+        (migrate!))
+
+      (testing "no data_migrations table after v.48.00-024"
+        (is (thrown? ExceptionInfo
+                     (t2/count :data_migrations))))
+
+      (testing "rollback causes all known data_migrations to reappear"
+        (migrate-all! :down 47)
+        ;; 2 because there were two data migrations left in release 0.47
+        (is (<= 34 (t2/count :data_migrations))))
```
qnkhuat, line 1476:
```suggestion
        (is (= 34 (t2/count :data_migrations))))
```

tests should be certain, using `<=` means 0 is acceptable too and that doesn't give us much certainty right?

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1033,3 +1039,444 @@
...
+    (let [{:keys [db-type ^javax.sql.DataSource
+                  data-source]} mdb.connection/*application-db*
+          ;; approach from schema-migrations-test/migrate-field-database-type-test
+          migrate-all!          (partial db.setup/migrate! db-type data-source)
+          throw-err             (fn [& _args]
+                                  (throw (ex-info "This shouldn't be called ever" {})))]
+
+      (testing "we can migrate even if data_migrations is empty"
+        ;; 0 because we removed them and fresh db won't trigger any
+        (is (<= 0 (t2/count :data_migrations)))
```
qnkhuat, line 1466:
```suggestion
        (is (= 0 (t2/count :data_migrations)))
```

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1033,3 +1039,444 @@
...
+        (migrate!))
+
+      (testing "no data_migrations table after v.48.00-024"
+        (is (thrown? ExceptionInfo
+                     (t2/count :data_migrations))))
+
+      (testing "rollback causes all known data_migrations to reappear"
+        (migrate-all! :down 47)
+        ;; 2 because there were two data migrations left in release 0.47
```
qnkhuat, line 1475:
```suggestion
        ;; 34 because there were two data migrations left in release 0.47
```

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1033,3 +1039,444 @@
...
+        (mt/with-temporary-raw-setting-values
+          [ldap-sync-admin-group "true"]
+          (#'custom-migrations/migrate-remove-admin-from-group-mapping-if-needed)
+          (is (= ldap-group-mappings (get-json-setting :ldap-group-mappings))))))))
+
+(deftest check-data-migrations-rollback
+  (impl/test-migrations ["v48.00-024"] [migrate!]
+    (let [{:keys [db-type ^javax.sql.DataSource
+                  data-source]} mdb.connection/*application-db*
+          ;; approach from schema-migrations-test/migrate-field-database-type-test
```
qnkhuat, line 1459: I don't think this is a useful comment, also plenty of test in this ns use this patterns to so I think that's kinda obvious.

test/metabase/db/custom_migrations_test.clj
```diff
@@ -1033,3 +1039,444 @@
...
+                                        :linkTemplate     "http://example.com//{{id}}",
+                                        :linkTextTemplate "here is my id: {{id}}"}}}}]
+              (f)
+              (is (= expected-settings
+                     (-> (t2/select-one :model/DashboardCard :id dashcard-id)
+                         :visualization_settings))))))]
+    (testing "Running the migration from scratch"
+      (impl/test-migrations ["v48.00-022"] [migrate!]
+        (expect-correct-settings! migrate!)))
+    (testing "Running the migration after a previous data-migration still works"
```
qnkhuat, line 1389: is this test meant to check if `migrate-click-through!` is idempotent?

either way I still think we need a test to check if `select count(*) from data_migrations where id = 'migrate-click-through'` returns 1, the custom migration `custom-migrations/MigrateClickThrough` will not be executed.

qnkhuat (approved):
Have some nits but I think this LGTM 🚢

test/metabase/db/custom_migrations_test.clj
piranha: well, it means 34 and more in this case, so I just wanted to be a bit less specific here, but sure, can be just `= 34` :)

test/metabase/db/custom_migrations_test.clj
piranha: hmm you mean it's fresh db and it's always going to be 0?

PR #34602, opened 2023-10-12
src-min-noconflict -> src-noconflict

calherries:
We were pulling src-min-noconflict AND src-noconflict on ace-builds. As we can't yet transform completely to src-min-noconflict completely, I'm droping the src-min-noconflict to unify to just 1 package so we don't add it twice.

Previous discussion [here](https://github.com/metabase/metabase/pull/34534#discussion_r1356706040)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -163,7 +163,7 @@ const config = (module.exports = {
       __support__: TEST_SUPPORT_PATH,
       e2e: E2E_PATH,
       style: SRC_PATH + "/css/core/index",
-      ace: __dirname + "/node_modules/ace-builds/src-min-noconflict",
+      ace: __dirname + "/node_modules/ace-builds/src-noconflict",
       // NOTE @kdoh - 7/24/18
       // icepick 2.x is es6 by defalt, to maintain backwards compatability
       // with ie11 point to the minified version


uladzimirdev (dismissed):

uladzimirdev:
Actually I don't see any benefit of using `ace-builds/src-noconflict` instead of `ace-builds/src-min-noconflict` everywhere.

@calherries @paoliniluis could you please clarify it once again?

paoliniluis:
> Actually I don't see any benefit of using `ace-builds/src-noconflict` instead of `ace-builds/src-min-noconflict` everywhere.
> 
> @calherries @paoliniluis could you please clarify it once again?

I think we should use src-noconflict in the code as it will get minified in the build, just for consistency

calherries:
@uladzimirdev Anton mentioned a benefit of using ace-builds/src-noconflict instead of ace-builds/src-min-noconflict: navigating to ace's source code https://github.com/metabase/metabase/pull/34534#discussion_r1356773351

kamilmielnik (approved):
I did some measurements:
- `vendor.[hash].js` file is now 330kB smaller (5,693,140 B -> 5,354,438 B)
- build time seems unaffected

Awesome! :tada: 

----

Using `src-min-noconflict` everywhere does not bring these benefits (i.e. bundle size decrease or [navigating sources](https://github.com/metabase/metabase/pull/34602#issuecomment-1760973506))

PR #34606, opened 2023-10-12
add echarts renderer component

alxnddr:
ECharts integration branch for PieChart and static ComboChart


frontend/src/metabase/visualizations/components/EChartsRenderer/EChartsRenderer.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  useMount(() => {
+    chartRef.current = init(chartElemRef.current, null, {
+      width,
+      height,
+      renderer: "svg",
+    });
+
+    chartRef.current?.setOption(option, true);
+    onInit?.(chartRef.current);
```
alxnddr, line 39: We also need to dispose the ECharts instance on unmount

frontend/src/metabase/visualizations/components/EChartsRenderer/EChartsRenderer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  zrEventHandlers,
+  width,
+  height,
+  onInit,
+}: EChartsRendererProps) => {
+  const chartElemRef = useRef<HTMLDivElement>(null);
+  const chartRef = useRef<EChartsType>();
+
+  useMount(() => {
+    chartRef.current = init(chartElemRef.current, null, {
```
alxnddr, line 32: We also need to dispose the ECharts instance on unmount

EmmadUsmani (approved):

PR #34607, opened 2023-10-12
Disable model metadata editor animation in Cypress

kulyk:
As @nemanjaglumac pointed out, the "in" transition of the model metadata editor sidebar is causing flakes and there's no easy workaround. This PR disables the animation in E2E tests

### To Verify

1. Open `/model/:modelId/metadata` in your browser, click through a few columns
2. Ensure the sidebar animates when a column changes
3. Repeat the same thing in any E2E test
4. Ensure there's no animation

(backporting so the release branch can get less flaky as well)


iethree:
😬 I really dislike having different behavior in tests. Can we just remove the animation?

kulyk:
Closing in favor of #34612 (removing the animation altogether)

PR #34608, opened 2023-10-12
Use Marginalia to centralize documentation

tsmacdonald:
This PR re-introduces Marginalia, which generates pretty HTML based on docstrings. We used to use it, but it got removed during the migration to tools.deps since Cam (mistakenly) believed it required Leiningen to run.

In short, Marginalia generates an HTML file that [looks like this](http://gdeer81.github.io/marginalia/) and provides a side-by-side view of your comments/docstrings and code, nicely rendered as HTML.

# Why bother?

During the offsite, I felt like there was consensus that our backend dev-facing documentation is split across several places and can be hard to discover. I think the best solution is to keep as much documentation in the code itself—it's so much easier to notice and fix something that's out of date if the docs are right next to the code you're changing.

So my vision is:

* We start using `dev.clj` as the hub for useful links. I already started:

image

* We continue to write excellent ns-level docstrings.

* We additionally, where appropriate, write introductions to important parts of the codebase. I just wrote one for `defendpoint`:

image

* Instead of adding new pages to the Github wiki, we start adding them somewhere sensible in `src/` or `dev/` so that Marginalia can pick them up. (it'd also be nice to migrate existing content on the Github wiki)

# Other notes

* Marginalia is nearly abandonware, and I came across a bug with reader conditionals that had been fixed in a PR but not merged. So I forked the project, fixed the bug, and deployed a new version to Clojars. If more work ends up being needed on Marginalia itself I can either maintain that fork more enthusiastically, try to get commit access to the original repo, or ask clj-commons to adopt it. If you think that it should be hosted on a Metabase-controlled Clojars group let me know; I'm happy to transfer it.

* Run it locally with `clojure -M:marginalia`, then open `backend-docs/uberdoc.html`

* In the (near) future, it'd be great to get the HTML file generated automatically and hosted somewhere accessible. Didn't we do that in the past? How did that work?

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -297,8 +297,8 @@
 
   :marginalia
   {:extra-deps
-   {marginalia/marginalia {:mvn/version "0.9.1"}}
-   :main-opts ["-m" "marginalia.main" "-d" "backend-docs" "src/**/*.clj"]}
+   {com.github.tsmacdonald/marginalia {:mvn/version "0.9.2"}}
+   :main-opts ["-m" "marginalia.main" "-n" "Metabase" "-d" "backend-docs" "-D" "The simplest, fastest way to get business intelligence and analytics to everyone in your company 😋" "dev" "src"]}
 
   ;; Find outdated versions of dependencies. Run with `clojure -M:outdated`
   :outdated {;; Note that it is `:deps`, not `:extra-deps`
--- a/deps.edn
+++ b/deps.edn
@@ -300,7 +300,9 @@
   :marginalia
   {:extra-deps
    {com.github.tsmacdonald/marginalia {:mvn/version "0.9.2"}}
-   :main-opts ["-m" "marginalia.main" "-n" "Metabase" "-d" "backend-docs" "-D" "The simplest, fastest way to get business intelligence and analytics to everyone in your company 😋" "dev" "src"]}
+   :main-opts ["-m" "marginalia.main" "-n" "Metabase" "-d" "backend-docs" "-D"
+               "The simplest, fastest way to get business intelligence and analytics to everyone in your company 😋"
+               "dev" "src" "shared/src" "enterprise/backend/src"]}
 
   ;; Find outdated versions of dependencies. Run with `clojure -M:outdated`
   :outdated {;; Note that it is `:deps`, not `:extra-deps`
--- a/deps.edn
+++ b/deps.edn
@@ -295,6 +295,8 @@
   :oss-dev
   {}
 
+  ;; Generate BE documentation with
+  ;; clojure -M:marginalia
   :marginalia
   {:extra-deps
    {com.github.tsmacdonald/marginalia {:mvn/version "0.9.2"}}
--- a/deps.edn
+++ b/deps.edn
@@ -283,8 +283,8 @@
   {:extra-paths ["enterprise/backend/src"]}
 
   ;; Include EE tests.
-  ;; for ee dev: :dev:ee:ee-dev
-  ;; for ee tests: clojure -X:dev:ee:ee-dev:test
+  ;; for EE dev:   `clojure -X:dev:ee:ee-dev`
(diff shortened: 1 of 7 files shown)


iethree:
> In the (near) future, it'd be great to get the HTML file generated automatically and hosted somewhere accessible. Didn't we do that in the past? How did that work?

This is a textbook use case for github actions + github pages

tsmacdonald:
@iethree was planning to ping you about exactly that once this got merged :)

src/metabase/search/scoring.clj
```diff
@@ -1,3 +1,106 @@
...
+;; little complicated…it is! It also means that it can be tricky to give a proper answer about why the search ranking
+;; is "wrong", maybe you search for `monthly revenue` and are looking for a card called `monthly revenue` and are mad
+;; that a dashboard called `company stats` shows up first…but then it turns out that the dashboard's description is
+;; `Stats that everyone should be aware of, such as our order count and monthly revenue.` and the dashboard happens to
+;; be pinned, bookmarked, part of an official collection, verified, and edited a couple hours ago…whereas the card is
+;; none of those things.
+;;
+;; Also, be aware that there's [a big epic under way](https://github.com/metabase/metabase/issues/27982) to add
+;; filtering to search results, which should help people find what they're looking for (and spares us from having to
+;; make the above algorithm better).
```
iethree, line 100: Can we phrase this paragraph in a way that won't make this out of date in a few months like "As of October 2023, there is work being done toward ... "

deps.edn
```diff
@@ -295,6 +295,13 @@
   :oss-dev
   {}
 
+  ;; Generate BE documentation with
+  ;; clojure -M:marginalia
+  :marginalia
+  {:extra-deps
+   {com.github.tsmacdonald/marginalia {:mvn/version "0.9.2"}}
+   :main-opts ["-m" "marginalia.main" "-n" "Metabase" "-d" "backend-docs" "-D" "The simplest, fastest way to get business intelligence and analytics to everyone in your company 😋" "dev" "src"]}
```
camsaul, line 303: You probably also want to generate dox for `shared` and `enterprise/backend/src`

src/metabase/search/scoring.clj
```diff
@@ -1,3 +1,106 @@
+;; # How does search scoring work?
```
camsaul, line 1: Dumb question, but can't you put all of this stuff in the namespace docstring? IIRC Marginalia will still picc up on that stuff and treat it as Markdown

camsaul (approved):
Cool this looks good to me, now hopefully it will shame us into writing more and better documentation. 

As a next step maybe we should set up a GH flow to automate building the dox and publishing them somewhere on every commit to `master`, it would be easy enough to push it to the repo's wiki (which itself is just another Git repo) or to another repo in and have GitHub pages render it

deps.edn
tsmacdonald: Oh snap, good call

src/metabase/search/scoring.clj
tsmacdonald: Yeah, there's no real distinction technically, I just felt like ns docstrings have a certain style to them and the three `;;` comments I wrote were stylistically different enough to belong somewhere else.

Like, for `mb.search.scoring`, I think that comment I wrote is useful, and if I were a better person I'd write a ns docstring that was something like

<blockquote>
Computes a relevancy score for search results using the weighted average of various scorers. Scores are determined by various ways of comparing the text of the search string and the item's title or description, as well as by Metabase-specific features such as how many dashboards a card appears in or whether an item is pinned.


Get the score for a result with `score-and-result`, and efficiently get the most relevant results with `top-results`.

Some of the scorers can be tweaked with configuration in [[metabase.search.config]].

</blockquote>

(uhh, I'll make that the real docstring in a second)

calherries (approved):

PR #34369, opened 2023-10-04
Prevent dashboard parameter labels with identical slugs

calherries:
Step 2 of fixing #27653 (see [this comment](https://github.com/metabase/metabase/issues/27653#issuecomment-1746565164) for the whole plan)

Step 1's PR: https://github.com/metabase/metabase/pull/34358

Prevents a dashboard parameter label from being set so that its slug is the same as another.

If you’re editing the label field, then change it so the slug is the same as another parameter, we show an error state with the message "This label is already in use". When you unfocus the input box in the error state, it reverts back to the label it had before.

PR #34613, opened 2023-10-12
Fix dashboard filter label carrying over from previously edited label

### LINKED ISSUE #34611: Dashboard filter label is carried over from previously edited filter
### Describe the bug

I introduced this regression with #34369

When editing a dashboard filter, the label is carried over from previously edited filter

https://www.loom.com/share/2b2d17ce366f41e8b5cd6f1f4cae6299?sid=a28ebcb1-1cc0-49a6-8eef-54cfed13d2bb

### To Reproduce

Add a filter to a dashboard and edit the label. Then create another filter and see that the previous label is used, instead of the default.

### Expected behavior

Default should be used

### Logs

_No response_

```JSON
master (a2d3b00520)
```

### Severity

low

### Additional context

_No response_

calherries:
Fixes #34611

This regression happened in when I removed this code https://github.com/metabase/metabase/pull/34369/files#diff-7dadc4846cebae50d37bcfb4679d3373e3732401e5a9f0eb90d863f86ea88afdL51-L53. This PR restores those four lines that were removed and adds a test for it.

### DIFF
--- a/frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
+++ b/frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/prop-types */
 import userEvent from "@testing-library/user-event";
 import { useState } from "react";
 import { renderWithProviders, screen } from "__support__/ui";
--- a/frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
+++ b/frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
@@ -89,7 +89,7 @@ describe("ParameterSidebar", () => {
     expect(labelInput).toHaveValue("Bar");
   });
 
-  it("if the parameter updates, the label should update (metabase#34613)", () => {
+  it("if the parameter updates, the label should update (metabase#34611)", () => {
     const initialParameter = createMockUiParameter({
       id: "id1",
       name: "Foo",
--- a/frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
+++ b/frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
@@ -1,6 +1,7 @@
+/* eslint-disable react/prop-types */
 import userEvent from "@testing-library/user-event";
-import React from "react";
-import { act, renderWithProviders, screen } from "__support__/ui";
+import { useState } from "react";
+import { renderWithProviders, screen } from "__support__/ui";
 import type { UiParameter } from "metabase-lib/parameters/types";
 import { createMockUiParameter } from "metabase-lib/parameters/mock";
 import ParameterSidebar from "./ParameterSidebar";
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
```diff
@@ -78,4 +88,29 @@ describe("ParameterSidebar", () => {
...
+    });
+    const { setParameter } = setup({
+      initialParameter,
+      otherParameters: [otherParameter],
+    });
+
+    const labelInput = screen.getByLabelText("Label");
+    expect(labelInput).toHaveValue("Foo");
+
+    act(() => {
```
oisincoveney, line 111: Usually we don't use `act` in our tests since it's generally bad practice with React Testing Library, where functions like RTL's `render` are already wrapped with `act`. Since you're using a test wrapper for the function already, maybe you can add a button that triggers the `setParameter` function, then check for the `labelInput` value?

[Context here](https://flyingsquirrel.medium.com/when-should-i-use-act-in-react-testing-library-d7dd22a3340e#:~:text=You%20don't%20need%20to%20use%20act()%20actually.,-The%20reason%20why&text=testing%2Dlibrary%2Freact%20overrides%20eventWrapper,are%20executed%20in%20act()%20.&text=Therefore%20you%20don't%20have,like%20userEvent%20with%20act()%20.&text=act%20calls%20can%20be%20nested,actScopeDepth%20based%20on%20the%20stack.)

oisincoveney: This is what I'm thinking in terms of how the test should be structured - let me know what you think

https://github.com/metabase/metabase/commit/917bf2f058a713c5149b82e0d05a3995049d8831

frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
```diff
@@ -55,18 +49,26 @@ const TestWrapper = ({
...
       otherParameters={otherParameters}
-      onSetParameter={setParameterMock}
     />,
   );
 
-  return { setParameterMock };
+  // clicking the next parameter button will update the parameter from initialParameter to nextParameter
+  return {
+    clickNextParameterButton: () =>
+      userEvent.click(screen.getByTestId("parameter-sidebar-test-change")),
```
calherries, line 70: @oisincoveney I introduced this indirection to not require tests to know about the test ID. Only the setup function needs to know about how to click the button

oisincoveney: I think you might be able to split this into its own function - the setup function doesn't even need to know how to click the button

oisincoveney (approved):
Just that one comment and I'm happy

frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.unit.spec.tsx
calherries: Yes, but I think putting TestWrapper back inside `setup` resolves the conflict - now setup's implementation explicitly knows how to click the button, and nothing else does. https://github.com/metabase/metabase/pull/34613/files/4fbb73408784e52bef448b4ca1c2329a3a85f8cc..c9bd641a7751e22f181f1e1b6be905854eb4c272

PR #34621, opened 2023-10-12
Add public setting to opt-in to CC recipients

### LINKED ISSUE #34293: Add an option in the `/admin/settings/email` to opt-in to CC recipients
On 47, we changed our subscriptions to send all recipients in BCC.
Some customers reported that their emails were being blocked by their email providers for security reasons.

In most cases, sending the recipients in BCC makes sense (as it preserves the anonymity of recipients). But there are cases when admins want to add recipients in CC:
- if the email provider blocks bcc
- if they want to openly share recipients and have discussions on email threads

We are adding an option on `/admin/settings/email` to send emails with all recipients in CC.
The default behavior continues to be adding everyone as BCC.

Copy suggestion:

header: ADD RECIPIENTS AS CC OR BCC
description: Control the visibility of recipients
options: - Bcc (hide recipients) _default_ - Cc (disclose recipients)

Figma [link](https://www.figma.com/file/DdYMSBqSsMfz9ZIN38D9D8/Add-recipients-as-CC-or-BCC?type=design&node-id=23-393&mode=design&t=zHx72CSTP0qDlezU-4)

qwef:
For https://github.com/metabase/metabase/issues/34293

### Description

Adds a public setting for whether or not bcc is enabled.

### How to verify

1. Change the public setting to false
2. Send a pulse
3. See the email send `to:` :)

### DIFF
--- a/src/metabase/public_settings.clj
+++ b/src/metabase/public_settings.clj
@@ -683,9 +683,3 @@
   (deferred-tru "Prefix for upload table names")
   :visibility   :authenticated
   :type         :string)
-
-(defsetting bcc-enabled?
-  (deferred-tru "Whether or not bcc emails are enabled, default behavior is that it is")
-  :visibility :authenticated
-  :type       :boolean
-  :default    true)
--- a/src/metabase/public_settings.clj
+++ b/src/metabase/public_settings.clj
@@ -683,3 +683,9 @@
   (deferred-tru "Prefix for upload table names")
   :visibility   :authenticated
   :type         :string)
+
+(defsetting bcc-enabled?
+  (deferred-tru "Whether or not bcc emails are enabled, default behavior is that it is")
+  :visibility :authenticated
+  :type       :boolean
+  :default    true)
--- a/enterprise/backend/test/metabase_enterprise/sandbox/pulse_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/sandbox/pulse_test.clj
@@ -6,11 +6,11 @@
    [medley.core :as m]
    [metabase-enterprise.test :as met]
    [metabase.api.alert :as api.alert]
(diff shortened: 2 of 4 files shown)


src/metabase/public_settings.clj
```diff
@@ -683,3 +683,9 @@
   (deferred-tru "Prefix for upload table names")
   :visibility   :authenticated
   :type         :string)
+
+(defsetting bcc-enabled?
+  (deferred-tru "Whether or not bcc emails are enabled, default behavior is that it is")
+  :visibility :authenticated
+  :type       :boolean
+  :default    true)
```
noahmoss, line 691: nit: For the sake of organization let's try to move away from using `metabase.public-settings` as a catch-all settings namespace. It was originally for _public_ settings specifically but it's become polluted with lots of different things.

Since settings are namespace-agnostic, they can go anywhere without changing behavior. This one might make the most sense in `metabase.email` near all the other email-related settings?

src/metabase/public_settings.clj
```diff
@@ -683,3 +683,9 @@
   (deferred-tru "Prefix for upload table names")
   :visibility   :authenticated
   :type         :string)
+
+(defsetting bcc-enabled?
+  (deferred-tru "Whether or not bcc emails are enabled, default behavior is that it is")
+  :visibility :authenticated
```
noahmoss, line 689: Can visibility just be `:settings-manager`? I don't think all users need to be able to read this setting since it's only fetched on the backend, right?

[Here's](https://github.com/metabase/metabase/blob/master/src/metabase/models/setting.clj#L1005-L1011) a handy chart for what the different visibility levels mean for settings.

noahmoss (approved):

PR #34624, opened 2023-10-12
fix backports by using gh cli

alxnddr:
Migrate from `hub` to `gh` cli in the backport scripts because of https://github.com/actions/runner-images/issues/8362


dragonsahead (approved):
seems reasonable

npfitz (approved):

Issue #34495, opened 2023-10-09
[MLv2] FE - Segments


iethree:
blocked by https://github.com/metabase/metabase/issues/34511

can maybe implement similarly to [metrics with aggregations](https://github.com/metabase/metabase/pull/31528)

iethree:
unblocked by https://github.com/metabase/metabase/pull/34547

Issue #3454, opened 2016-10-05
Alias Integer (Enum) Columns


jacobsmith:
I'm coming to Metabase from a Rails background, and enums are used fairly often there. i.e., a table may have:

`color: 0 | 1 | 2`

and the application code has

`enum color: %w{red green blue}`

I've been able to use the built in filters to say `WHERE color = 0`, but it would be awesome if in the metadata (or somewhere) I could give "hints" to the filter, so that it would be a drop down with "Red", "Blue", "Green".

I have a tiny bit of Java background, but if someone were to point me in the right direction for solving this, I'd love to try and submit a PR for it!

(Original Discourse discussion: https://discourse.metabase.com/t/is-there-a-way-to-alias-integer-enum-columns/736 )

hugocar:
Hi
( I am not from the metabase staff)

Have you tried the segments in the data model admin panel ? 

Depending on your data, you could also create a view of some table with schema more adapted to OLAP queries.

c2h5oh:
We could definitely use this feature, especially since Postgres enum support is broken - people having to use a printout with mappings doesn't provide the best experience.

/api/field/:fieldId/value_map_update and /api/field/:fieldId/values" both work (tested with curl) but there is no UI.
Aliases you set are ignored when when building segments / filters / questions and displaying data, but that again looks like just missing frontend bits.

c2h5oh:
@camsaul how much work is required to have:
- ui for setting / viewing those aliases
- aliases being used in question builder and in answers

we're considering sponsoring this by either setting a bounty or investing our own time

I know that kind of estimation is wildly inaccurate, but an educated guess is better than nothing.

camsaul:
@c2h5oh there needs to be some frontend work (such as pages to set the enum values) and backend work (such as new endpoints and possibly DB migrations) done here so it's not a trivial task. The elite developers that make up the core of the Metabase team (:wink:) could handle it but I think it might be a challenge for people not familiar with the Metabase codebase.

salsakran:
supported via #5130

PR #34603, opened 2023-10-12
[MLv2] Add the expression editor to `FilterPicker`

kulyk:
Closes #34496

Adds the custom expression editor to the `FilterPicker`. Custom expressions themselves are not yet ported to MLv2, so for now we're using MLv1 to work with them. This PR is very similar to what we've done to aggregations in #31529

### To Verify

1. New > Question > Raw Data > Sample Database > Orders
2. Add a standard filter (e.g. Total > 100)
3. Click the filter, scroll down, and click "Custom Expression"
4. Ensure the editor opened with `Total > 100` formula inside
5. Change the formula to something else, click "Done"
6. Click the filter, ensure you can see the original formula

### Demo

https://github.com/metabase/metabase/assets/17258145/2bc0e9a4-21a9-49a3-bcaf-9519da1ae627

PR #34635, opened 2023-10-12
[MLv2] Support Segment Filters

iethree:
Closes https://github.com/metabase/metabase/issues/34495

> **Note** 
> Disregard the clojure changes, I pulled them in from https://github.com/metabase/metabase/pull/3454

### Description

Supports segment filters by extending the columnPicker to support segments if there is an `onSegmentSelect` function provided

Screen Shot 2023-10-12 at 4 23 42 PM

### DIFF
--- a/frontend/src/metabase-lib/metadata.ts
+++ b/frontend/src/metabase-lib/metadata.ts
@@ -20,6 +20,8 @@ import type {
   ColumnMetadata,
   DrillThru,
   DrillThruDisplayInfo,
+  FilterOperator,
+  FilterOperatorDisplayInfo,
   JoinConditionOperator,
   JoinConditionOperatorDisplayInfo,
   JoinStrategy,
@@ -29,11 +31,11 @@ import type {
   MetricDisplayInfo,
   OrderByClause,
   OrderByClauseDisplayInfo,
+  Query,
+  SegmentMetadata,
+  SegmentDisplayInfo,
   TableDisplayInfo,
   TableMetadata,
-  Query,
-  FilterOperator,
-  FilterOperatorDisplayInfo,
 } from "./types";
 
 export function metadataProvider(
@@ -127,6 +129,11 @@ declare function DisplayInfoFn(
   stageIndex: number,
   filterOperator: FilterOperator,
 ): FilterOperatorDisplayInfo;
+declare function DisplayInfoFn(
+  query: Query,
+  stageIndex: number,
+  segment: SegmentMetadata,
+): SegmentDisplayInfo;
 
 // x can be any sort of opaque object, e.g. a clause or metadata map. Values returned depend on what you pass in, but it
 // should always have display_name... see :metabase.lib.metadata.calculation/display-info schema
--- a/frontend/src/metabase-lib/types.ts
+++ b/frontend/src/metabase-lib/types.ts
@@ -146,7 +146,6 @@ export type SegmentDisplayInfo = {
   displayName: string;
   longDisplayName: string;
   description: string;
-  effectiveType: string;
 };
 
 export type AggregationOperatorDisplayInfo = {
--- a/frontend/src/metabase-lib/types.ts
+++ b/frontend/src/metabase-lib/types.ts
@@ -141,6 +141,14 @@ export type ColumnDisplayInfo = {
   selected?: boolean; // used in aggregation and field clauses
 };
 
+export type SegmentDisplayInfo = {
+  name: string;
+  displayName: string;
+  longDisplayName: string;
+  description: string;
+  effectiveType: string;
+};
+
 export type AggregationOperatorDisplayInfo = {
   columnName: string;
   displayName: string;
@@ -192,7 +200,8 @@ export type ExpressionOperatorName =
   | "interval"
   | "time-interval"
   | "relative-datetime"
-  | "inside";
+  | "inside"
+  | "segment";
 
 export type ExpressionArg = null | boolean | number | string | ColumnMetadata;
 
--- a/frontend/src/metabase/common/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
+++ b/frontend/src/metabase/common/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
@@ -12,7 +12,7 @@ import type { IconName } from "metabase/core/components/Icon";
 
 import * as Lib from "metabase-lib";
 
-import { StyledAccordionList } from "../FilterPicker.styled";
+import { StyledAccordionList } from "./FilterColumnPicker.styled";
 
 export interface FilterColumnPickerProps {
   query: Lib.Query;
@@ -47,7 +47,7 @@ export type SegmentListItem = Lib.SegmentDisplayInfo & {
 const isSegmentListItem = (
   item: ColumnListItem | SegmentListItem,
 ): item is SegmentListItem => {
-  return (item as SegmentListItem).segment !== undefined;
+  return (item as SegmentListItem).segment != null;
 };
 
 /**
@@ -73,10 +73,10 @@ export function FilterColumnPicker({
         column,
       }));
 
-      const includeSegments = !groupInfo.isFromJoin;
+      const includeSegments = (groupInfo as Lib.TableDisplayInfo).isSourceTable;
 
       const segmentItems = includeSegments
-        ? Lib.availableSegments(query).map((segment: Lib.SegmentMetadata) => ({
+        ? Lib.availableSegments(query).map(segment => ({
             ...Lib.displayInfo(query, stageIndex, segment),
             segment,
           }))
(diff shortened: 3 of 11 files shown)


frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
```diff
@@ -48,26 +59,42 @@ export function QueryColumnPicker({
...
+        const includeSegments = !!onSelectSegment && !groupInfo.isFromJoin;
+
+        const segmentItems = includeSegments
+          ? Lib.availableSegments(query).map(
+              (segment: Lib.SegmentMetadata) => ({
+                ...Lib.displayInfo(query, stageIndex, segment),
+                segment,
+              }),
+            )
+          : [];
```
iethree, line 86: I don't *love* overloading the columnPicker to be a column-or-segment picker, but it seemed insane to create a separate picker that duplicated all of the column picker logic. 

I wonder if it's worth reworking the columnPicker to be more generic? I also fear that might be grossly over-engineering it.

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
```diff
@@ -48,26 +59,42 @@ export function QueryColumnPicker({
...
-          column,
-        }));
+        const columnItems = Lib.getColumnsFromColumnGroup(group).map(
+          column => ({
+            ...Lib.displayInfo(query, stageIndex, column),
+            column,
+          }),
+        );
+
+        const includeSegments = !!onSelectSegment && !groupInfo.isFromJoin;
```
iethree, line 77: I [posted in slack](https://metaboat.slack.com/archives/C04CYTEL9N2/p1697148236490379) about this, but currently both MLv1 and MLv2 only support segments for non-joined tables. This seems like a bug to me, but maybe not one worth fixing right now.

ranquild (commented):
I'd rather not extend `QueryColumnPicker` but create a separate component for filters only. Even the name is misleading - we no longer pick columns but also segments. Also most of the functionality in the `QueryColumnPicker` is not needed for filters - e.g. binning or temporal bucketing. 

For aggregations we have `AggregationPicker` which renders a combined list of operators and metrics in once place. I think we need to do the same here - e.g. add our own component.

@kulyk WDYT?

kulyk:
> For aggregations we have `AggregationPicker` which renders a combined list of operators and metrics in once place. I think we need to do the same here - e.g. add our own component.

`AggregationPicker` is a bit different, it doesn't mix anything into a columns list, we're only [rendering `QueryColumnPicker` when selecting a column for a selected aggregation operator](https://github.com/metabase/metabase/blob/c2209191610666afe26e6877083aa7961b75e931/frontend/src/metabase/common/components/AggregationPicker/AggregationPicker.tsx#L239). Here it's a bit more mixed up (columns + segments + custom expression button).

> I'd rather not extend `QueryColumnPicker` but create a separate component for filters only. Even the name is misleading - we no longer pick columns but also segments. Also most of the functionality in the `QueryColumnPicker` is not needed for filters - e.g. binning or temporal bucketing.

Yeah, I'm not a big fan of how we're extending the `QueryColumnPicker` both here and in #34603
But I don't like having two _very_ similar column pickers either :( OTOH, it's not like `QueryColumnPicker` has a lot of code. Most of it is needed for bucketing, and we don't need that for filters as you pointed out. And the rest of the component is basically a wrapper around `AccordionList`

kulyk:
Looking more closely, I agree with @ranquild that we should better make another component. It's not great we need to duplicate some code, but it's not much. I opened [another PR adding custom expressions](https://github.com/metabase/metabase/pull/34656) and it also migrates filters off the `QueryColumnPicker`. The fact we're building `AccordionList's` sections right there should make adding segments more straightforward

frontend/src/metabase-lib/metadata.ts
```diff
@@ -29,11 +31,11 @@ import type {
...
   OrderByClause,
   OrderByClauseDisplayInfo,
+  Query,
+  SegmentMetadata,
+  SegmentDisplayInfo,
   TableDisplayInfo,
   TableMetadata,
-  Query,
-  FilterOperator,
-  FilterOperatorDisplayInfo,
```
kulyk, line 36: ❤️

frontend/src/metabase-lib/types.ts
```diff
@@ -141,6 +141,14 @@ export type ColumnDisplayInfo = {
   selected?: boolean; // used in aggregation and field clauses
 };
 
+export type SegmentDisplayInfo = {
+  name: string;
+  displayName: string;
+  longDisplayName: string;
+  description: string;
+  effectiveType: string;
+};
```
kulyk, line 150: Do we really need the `effectiveType` here? I'd rather prefer if the FE didn't know about it at all

iethree: I just typed it because MLv2 gives it to us. 🤷

kulyk: I think we can skip it 👍

frontend/src/metabase/common/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
```diff
@@ -0,0 +1,143 @@
...
+  getColumnGroupIcon,
+  getColumnGroupName,
+} from "metabase/common/utils/column-groups";
+
+import { Icon } from "metabase/core/components/Icon";
+import type { IconName } from "metabase/core/components/Icon";
+
+import * as Lib from "metabase-lib";
+
+import { StyledAccordionList } from "../FilterPicker.styled";
```
kulyk, line 15: Can we please move this to `FilterColumnPicker.styled`? Other components won't need the `StyledAccordionList`

frontend/src/metabase/common/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
```diff
@@ -0,0 +1,143 @@
...
+};
+
+export type SegmentListItem = Lib.SegmentDisplayInfo & {
+  segment: Lib.SegmentMetadata;
+};
+
+const isSegmentListItem = (
+  item: ColumnListItem | SegmentListItem,
+): item is SegmentListItem => {
+  return (item as SegmentListItem).segment !== undefined;
```
kulyk, line 50: How about something like that?

```suggestion
return (item as SegmentListItem).segment != null;
```

With the current check, `isSegmentListItem` would return `true` if `segment` is `null`

frontend/src/metabase/common/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    const sections = columnGroups.map(group => {
+      const groupInfo = Lib.displayInfo(query, stageIndex, group);
+
+      const columnItems = Lib.getColumnsFromColumnGroup(group).map(column => ({
+        ...Lib.displayInfo(query, stageIndex, column),
+        column,
+      }));
+
+      const includeSegments = !groupInfo.isFromJoin;
```
kulyk, line 76:
```suggestion
      const includeSegments = groupInfo.isSourceTable;
```

With `!groupInfo.isFromJoin` we'd also include segments in implicitly joinable table groups

frontend/src/metabase/common/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+      const segmentItems = includeSegments
+        ? Lib.availableSegments(query).map((segment: Lib.SegmentMetadata) => ({
+            ...Lib.displayInfo(query, stageIndex, segment),
+            segment,
+          }))
+        : [];
```
kulyk, line 83: We shouldn't need to type the `segment` argument as long as `Lib.availableSegments` has a return type

kulyk (changes requested):
@iethree thanks for addressing the comments! Inlined a few more suggestions

Also, it seems like we'd need to make the `FilterPicker` slightly less column-centric. When clicking a filter clause that uses a segment, the picker opens the expression editor by default:

demo

frontend/src/metabase/query_builder/components/notebook/steps/FilterStep/FilterStep.tsx
```diff
@@ -84,6 +84,7 @@ export function FilterStep({
             onClose={onClose}
           />
         )}
+        withLegacyPopover
```
iethree, line 87: without this (which I copied from AggregationStep) the popover was aggressively closing whenever you clicked an expression suggestion

iethree: ok, removed that commit

kulyk (approved):
LGTM!

PR #34636, opened 2023-10-12
Support duplicate columns hidden in cards

### LINKED ISSUE #34227: Invalid identifier when joining tables that have the same column name
### Describe the bug

I'm running into a Metabase joining error. I'm trying to join one table (courses) with another table that has aggregate statistics for each course. The join fails with the error:

```
SQL compilation error: error line 1 at position 203 invalid identifier '"Question 2295"."count_2"'
```

The table with aggregate statistics combines data from two different tables (i.e. number of enrollments, number waitlisted). These two tables (table 2292 and table 2293) do a sum by distinct user IDs. It looks like the issue is that there is a "COUNT" column for each aggregation but Metabase doesn't know how to handle multiples of the same column. The resulting SQL is: 

```
SELECT "Question 2295"."Question 2292__count" AS "Question 2295__Question 2292__count", "Question 2295"."count_2" AS "Question 2295__count_2" FROM ...
```

instead of

```
SELECT "Question 2295"."Question 2292__count" AS "Question 2295__Question 2292__count", "Question 2295"."Question 2293__count" AS "Question 2295__Question 2293__count" FROM ...
```

The difference is the "count_2" vs "Question 2293__count".

Here's a Loom: https://www.loom.com/share/5ee216c013ee47a4b0511a056c062f28

### To Reproduce

I tried reproducing in the sample database. I get an error but the error is different:
1. Create three tables (see screenshots)
2. Create a new question that joins table 3
3. See the error: `Distinct, non-empty sequence of Field clauses`. The error is different from what I get but the process is the same

**Table 1**
CleanShot 2023-09-28 at 10 40 43@2x

**Table 2**
CleanShot 2023-09-28 at 10 41 22@2x

**Table 3**: creates a aggregate table that joins table 1 and 2
CleanShot 2023-09-28 at 10 42 08@2x

**New question**
CleanShot 2023-09-28 at 10 42 55@2x

### Expected behavior

SQL should compile

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.20+8",
    "java.version": "11.0.20",
    "java.vm.version": "11.0.20+8",
    "os.name": "Linux",
    "os.version": "5.10.184-175.731.amzn2.x86_64",
    "user.language": "en",
    "user.timezone": "GMT"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "snowflake",
      "postgres"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "14.7"
      },
      "jdbc-driver": {
        "version": "42.5.1"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2023-08-15",
      "tag": "v1.46.8",
      "branch": "release-x.46.x",
      "hash": "27a9ec6"
    },
    "settings": {
      "report-timezone": "America/Los_Angeles"
    }
  }
}
```

### Severity

Medium - we have a lot of tables that depend on the intermediary table so it'd be really annoying to create the intermediary table every time

### Additional context

_No response_

metamben:
Fixes #34227.

### Description

In the example from the issue (and in the repro test) we create two identical cards
and join them both in a third card. This means that first two cards bring fields that
differ only in their join aliases, and these aliases "get lost" when the third card
using them is joined to a table.

Since we have multiple fields with the same ID, the IDs cannot be used to
unambiguously refer to the underlying fields. The only way to reference them
properly is by the name they have in the source metadata.

Unfortunately, this is yet another situation where flexibility needed to gracefully
handle invalid MBQL queries is at odds with the precision necessary to deal with
multiply referenced columns.

The approach of the PR is to use reference by name whenever referring by ID would
lead to ambiguity.

When switching to MLv2, this code will have to be changed.

### How to verify

Use the description in the bug ticket.

### DIFF
--- a/src/metabase/lib/join.cljc
+++ b/src/metabase/lib/join.cljc
@@ -767,10 +767,17 @@
   [pk-col          :- [:maybe lib.metadata/ColumnMetadata]
    visible-columns :- [:maybe [:sequential lib.metadata/ColumnMetadata]]]
   (when-let [pk-id (:id pk-col)]
-    (m/find-first (fn [{:keys [fk-target-field-id], :as col}]
-                    (and (lib.types.isa/foreign-key? col)
-                         (= fk-target-field-id pk-id)))
-                  visible-columns)))
+    (let [candidates (filter (fn [{:keys [fk-target-field-id], :as col}]
+                               (and (lib.types.isa/foreign-key? col)
+                                    (= fk-target-field-id pk-id)))
+                             visible-columns)
+          primary (first candidates)]
+      (if (some #(= (:id %) (:id primary)) (rest candidates))
+        (when (not-any? #(= (:lib/desired-column-alias %) (:lib/desired-column-alias primary)) (rest candidates))
+          ;; there are multiple candidates but :lib/desired-column-alias disambiguates them,
+          ;; sp reference by name
+          (dissoc primary :id))
+        primary))))
 
 (mu/defn ^:private fk-column-for :- [:maybe lib.metadata/ColumnMetadata]
   "Given a query stage find an FK column that points to the PK `pk-col`."
@@ -795,14 +802,14 @@
    (let [joinable-col-options {:include-implicitly-joinable? false
                                :include-implicitly-joinable-for-source-card? false}]
      (letfn [(filter-clause [x y]
-                   (lib.filter/filter-clause (lib.filter.operator/operator-def :=) x y))]
-           (or (when-let [pk-col (pk-column query stage-number joinable joinable-col-options)]
-                 (when-let [fk-col (fk-column-for query stage-number pk-col)]
-                   (filter-clause fk-col pk-col)))
-               (when-let [pk-col (pk-column query stage-number (lib.util/query-stage query stage-number) nil)]
-                 (when-let [fk-col (fk-column-for-pk-in pk-col (lib.metadata.calculation/visible-columns
-                                                                query stage-number joinable joinable-col-options))]
-                   (filter-clause pk-col fk-col))))))))
+               (lib.filter/filter-clause (lib.filter.operator/operator-def :=) x y))]
+       (or (when-let [pk-col (pk-column query stage-number joinable joinable-col-options)]
+             (when-let [fk-col (fk-column-for query stage-number pk-col)]
+               (filter-clause fk-col pk-col)))
+           (when-let [pk-col (pk-column query stage-number (lib.util/query-stage query stage-number) nil)]
+             (when-let [fk-col (fk-column-for-pk-in pk-col (lib.metadata.calculation/visible-columns
+                                                            query stage-number joinable joinable-col-options))]
+               (filter-clause pk-col fk-col))))))))
 
 (defn- add-join-alias-to-joinable-columns [cols a-join]
   (let [join-alias     (lib.join.util/current-join-alias a-join)
--- a/src/metabase/lib/join.cljc
+++ b/src/metabase/lib/join.cljc
@@ -775,7 +775,7 @@
       (if (some #(= (:id %) (:id primary)) (rest candidates))
         (when (not-any? #(= (:lib/desired-column-alias %) (:lib/desired-column-alias primary)) (rest candidates))
           ;; there are multiple candidates but :lib/desired-column-alias disambiguates them,
-          ;; sp reference by name
+          ;; so reference by name
           (dissoc primary :id))
         primary))))
 
--- a/src/metabase/query_processor/middleware/resolve_joins.clj
+++ b/src/metabase/query_processor/middleware/resolve_joins.clj
@@ -71,10 +71,16 @@
   (when-not (seq source-metadata)
     (throw (ex-info (tru "Cannot use :fields :all in join against source query unless it has :source-metadata.")
                     {:join join})))
-  (for [{field-name :name, base-type :base_type, field-id :id} source-metadata]
-    (if field-id
-      [:field field-id   {:join-alias alias}]
-      [:field field-name {:base-type base-type, :join-alias alias}])))
+  (let [duplicate-ids (into #{}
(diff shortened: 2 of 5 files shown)


src/metabase/query_processor/middleware/resolve_joins.clj
```diff
@@ -71,10 +71,15 @@
...
-      [:field field-name {:base-type base-type, :join-alias alias}])))
+  (let [duplicate-ids (into #{}
+                            (keep (fn [[item freq]]
+                                    (when (> freq 1)
+                                      item)))
+                            (frequencies (map :id source-metadata)))]
+    (for [{field-name :name, base-type :base_type, field-id :id} source-metadata]
+      (if (and field-id (not (contains? duplicate-ids field-id)))
+        [:field field-id   {:join-alias alias}]
+        [:field field-name {:base-type base-type, :join-alias alias}]))))
```
metamben, line 82: Prevent switching to the ID if it's ambiguous.

src/metabase/query_processor/util/add_alias_info.clj
```diff
@@ -238,18 +243,26 @@
...
               (m/find-first (fn [[_ _ opts :as _aggregation-options-clause]]
                               (= (::source-alias opts) field-name))
-                            (filter (partial mbql.u/is-clause? :aggregation-options) all-exports)))))))
+                            (filter (partial mbql.u/is-clause? :aggregation-options) all-exports))))
+        ;; look for a field referenced by the name in source-metadata
+        (let [field-name (second field-clause)]
+          (when (string? field-name)
+            (when-let [column (m/find-first #(= (:name %) field-name) source-metadata)]
+              (let [signature (field-signature (:field_ref column))]
+                (m/find-first #(= (field-signature %) signature) field-exports))))))))
```
metamben, line 252: This is the last option to maximize compatibility.

test/metabase/query_processor_test/explicit_joins_test.clj
```diff
@@ -324,6 +326,79 @@
...
+                       (->> top-card-query
+                            qp/process-query
+                            mt/cols
+                            (map col-data-fn)))))
+
+              (testing "suggested join condition references the FK by name"
+                (let [query (lib/query metadata-provider (lib.metadata/table metadata-provider (mt/id :people)))
+                      card-meta (lib.metadata/card metadata-provider 3)]
+                  (is (=? [:= {} [:field {} (mt/id :people :id)] [:field {} cuser-id]]
+                          (lib/suggested-join-condition query card-meta)))))
```
metamben, line 381: How bad do we feel about testing `suggested-join-condition` here? The only reason I added it here because I didn't feel like reproducing the query one more time as an MLv2 test.

src/metabase/lib/join.cljc
```diff
@@ -767,10 +767,17 @@
...
-                  visible-columns)))
+    (let [candidates (filter (fn [{:keys [fk-target-field-id], :as col}]
+                               (and (lib.types.isa/foreign-key? col)
+                                    (= fk-target-field-id pk-id)))
+                             visible-columns)
+          primary (first candidates)]
+      (if (some #(= (:id %) (:id primary)) (rest candidates))
+        (when (not-any? #(= (:lib/desired-column-alias %) (:lib/desired-column-alias primary)) (rest candidates))
+          ;; there are multiple candidates but :lib/desired-column-alias disambiguates them,
+          ;; sp reference by name
```
bshepherdson, line 778: nit: typo

src/metabase/query_processor/middleware/upgrade_field_literals.clj
```diff
@@ -20,15 +20,33 @@
...
 
+(defn- unique-reference?
+  "Check if the field-id and the (possibly missing) join-alias of `field-clause`
+  result in an unambiguous reference to one of the `columns`
+  The join-alias of columns is taken from their field_ref property."
+  [field-clause columns]
+  (let [[_ field-id {:keys [join-alias]}] field-clause
+        matches-by-id (filter #(= (:id %) field-id) columns)]
+    (or (nil? (next matches-by-id))
+        (= (count (filter #(= (get-in % [:field-ref 2 :join-alias]) join-alias) matches-by-id)) 1))))
```
bshepherdson, line 31: This is hard to read - `(->> matches-by-id (filter ...) count (= 1)`?

test/metabase/query_processor_test/explicit_joins_test.clj
```diff
@@ -324,6 +326,80 @@
                      :order-by [[:asc $name]]
                      :limit    3})))))))))
 
+;; This is a very contrived test. We create two identical cards and join them both
+;; in a third card. This means that first two cards bring fields that differ only in
+;; their join aliases, but these "get lost" when the third card is joined to a table.
+;; Since we have multiple fields with the same ID, the IDs cannot be used to
+;; unambiguously refer to the underlying fields. The only way to reference them
+;; properly is by the name they have in the source metadata.
+(deftest ^:parallel join-against-multiple-card-copies-test
```
bshepherdson, line 335: If I'm understanding this correctly, I think something like `TODO: When this middleware is ported to MLv2, the aliases can be connected properly.` would be appropriate here.

bshepherdson (approved):

test/metabase/query_processor_test/explicit_joins_test.clj
metamben: Would you really add this comment to this test? This test case will be relevant no matter what middlewares we have. I expect that only the way the query is built will have to change.

PR #34637, opened 2023-10-13
Applying sort-drill thru should REPLACE existing sort on same column

### LINKED ISSUE #34497: [MLv2] [BE] Applying `sort` to already sorted column doesn't remove initial sorting
Applying `sort` to already sorted column doesn't remove initial sorting, but always adds another one.

https://github.com/metabase/metabase/assets/7983744/0d1e0a97-0814-4813-a99e-9871764ca006

You can use this reproduction unit test - 
https://github.com/metabase/metabase/blob/fc358ef7c24ada415997ea781b3b1d706439a036/frontend/src/metabase-lib/drills.unit.spec.ts#L1593-L1613

camsaul:
Fixes #34497

### DIFF
--- a/src/metabase/lib/drill_thru/sort.cljc
+++ b/src/metabase/lib/drill_thru/sort.cljc
@@ -1,13 +1,35 @@
 (ns metabase.lib.drill-thru.sort
   (:require
+   [medley.core :as m]
    [metabase.lib.drill-thru.common :as lib.drill-thru.common]
    [metabase.lib.equality :as lib.equality]
    [metabase.lib.order-by :as lib.order-by]
    [metabase.lib.ref :as lib.ref]
+   [metabase.lib.remove-replace :as lib.remove-replace]
    [metabase.lib.schema :as lib.schema]
    [metabase.lib.schema.drill-thru :as lib.schema.drill-thru]
+   [metabase.lib.schema.order-by :as lib.schema.order-by]
    [metabase.lib.types.isa :as lib.types.isa]
    [metabase.util.malli :as mu]))
+
+(defn- orderable-column?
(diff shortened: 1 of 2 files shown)


src/metabase/lib/drill_thru/sort.cljc
```diff
@@ -46,8 +61,13 @@
   ([query                        :- ::lib.schema/query
     stage-number                 :- :int
     {:keys [column], :as _drill} :- ::lib.schema.drill-thru/drill-thru.sort
-    direction                    :- [:enum :asc :desc]]
-   (lib.order-by/order-by query stage-number column (keyword direction))))
+    direction                    :- ::lib.schema.order-by/direction]
```
metamben, line 64: 👍

src/metabase/lib/drill_thru/sort.cljc
```diff
@@ -46,8 +61,13 @@
...
+   ;; if you have an existing order by, the drill thru returned by [[sort-drill]] would only be one that would suggest
+   ;; changing it to the opposite direction, so we can safely assume we want to change the direction and
+   ;; use [[lib.order-by/change-direction]] here.
```
metamben, line 67: I still find it counterintuitive to set `:asc` if the drill says `:sort-directions [:desc]` or if `direction` is `:desc`. Would it be difficult or strange to set `direction`?

metamben (approved):
LGTM

src/metabase/lib/drill_thru/sort.cljc
camsaul: Yeah, I agree with you and I was going to push a change for this but I guess it auto-merged first. Sorry, I'll fix it in a follow on PR

PR #34646, opened 2023-10-13
Mantine TimeInput

ranquild:
Follow the same approach used by the mantine's `DateInput`:
- Allows controlled and uncontrolled usage
- Accepts and return a `Date`
- Resets the input value on blur

`TextInput` is used internally so no custom styles are needed.

How to test:
- `yarn storybook`
- Open http://localhost:6006/?path=/docs/inputs-timeinput--default-story
- To test AM/PM, change your Clock settings in the OS settings, and re-open the browser

Screenshot 2023-10-13 at 14 29 53
Screenshot 2023-10-18 at 3 39 45 PM
Screenshot 2023-10-18 at 3 39 55 PM

### DIFF
--- a/frontend/src/metabase/ui/components/inputs/TimeInput/TimeInput.tsx
+++ b/frontend/src/metabase/ui/components/inputs/TimeInput/TimeInput.tsx
@@ -1,0 +1,61 @@
+import { useLayoutEffect, useState } from "react";
+import type { ChangeEvent } from "react";
+import moment from "moment-timezone";
+import { TimeInput as MantineTimeInput } from "@mantine/dates";
+import type { TimeInputProps as MantineTimeInputProps } from "@mantine/dates";
+
+const TIME_FORMAT = "HH:mm";
+
+export type TimeInputProps = Omit<
+  MantineTimeInputProps,
+  "value" | "defaultValue" | "onChange"
+> & {
+  value?: Date | null;
+  defaultValue?: Date | null;
+  onChange?: (value: Date) => void;
+};
+
+export function TimeInput({
+  value,
+  defaultValue = value,
+  onChange,
+  ...props
+}: TimeInputProps) {
+  const [inputValue, setInputValue] = useState(
+    defaultValue ? formatTime(defaultValue) : "",
+  );
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const newInputValue = event.target.value;
+    setInputValue(newInputValue);
+
+    const newTime = parseTime(newInputValue);
+    newTime && onChange?.(newTime);
+  };
+
+  const handleBlur = () => {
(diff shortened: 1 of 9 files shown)


frontend/src/metabase/ui/components/inputs/TimeInput/TimeInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { TimeInput as MantineTimeInput } from "@mantine/dates";
+import type { TimeInputProps as MantineTimeInputProps } from "@mantine/dates";
+
+const TIME_FORMAT = "HH:mm";
+
+export type TimeInputProps = Omit<
+  MantineTimeInputProps,
+  "value" | "defaultValue" | "onChange"
+> & {
+  value?: Date | null;
```
ranquild, line 13: `null` is allowed to be compatible with `DateInput`. It has `clearable` option (`false` by default).

frontend/src/metabase/ui/components/inputs/TimeInput/TimeInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+  defaultValue?: Date | null;
+  onChange?: (value: Date) => void;
+};
+
+export function TimeInput({
+  value,
+  defaultValue = value,
+  onChange,
+  ...props
+}: TimeInputProps) {
```
kulyk, line 23: Curios why we need `value` and `defaultValue`? Can't we do something like that:

```jsx
function TimeInput({ value, onChange }) {
  const inputValue = value ? formatTime(value) : ""
  
  const handleChange = event => {
    const newTime = parseTime(event.target.value)
    onChange(newTime)
  }

  return <MantineTimeInput value={inputValue} onChange={handleChange} />
}
```

ranquild: `defalueValue` is needed for uncontrolled usage (all mantine components support it)

ranquild: e.g. `<TimeInput defaultValue={new Date()} />` - no `onChange` handler

kulyk: Do we need uncontrolled inputs though?

ranquild: all current mantine inputs support uncontrolled usage

kulyk: Oh, I see, makes sense to keep things consistent

kulyk (approved):

kdoh (approved):

Issue #34641, opened 2023-10-13
INTERNAL: Modularize E2E Workflow


nemanjaglumac:
E2E workflow grew too big.
Use composite actions and/or reusable workflows to modularize it.

```[tasklist]
### Tasks
- [ ] Prep
- [ ] Download uberjar
- [ ] Post/upload?
- [ ] tbd...
```

PR #34647, opened 2023-10-13
Move E2E download uberjar logic into a composite action

nemanjaglumac:
This PR is one more in line of tasks for #34641.

It moves the download uberjar logic to a composite action. Seemingly a very simple change, but it does so much more under the hood.
- it is now possible to download either the uberjar built in a current branch, or the one that was built before **from a completely different workflow run**
    - this makes it possible to re-use this logic in "regular" E2E tests and in E2E stress-test workflow
- It converts the unnecessarily complicated and verbose `curl` download logic to a more readable JavaScript logic that more people can follow (thanks @iethree for this suggestion)
- It fixes a bug that we had for a very long time in the commit parsing logic **on `pull_request`**
    - GitHub creates an intermediary merge commit on pull_request, and the git HEAD points to it - not to the last commit in our branch
    - every single uberjar built in the last year or so in a pull request had a wrong commit SHA1 reference stored in its `version.properties` file
    - both `git rev-parse HEAD` and `github.sha` were wrong choices if we wanted to get the last commit's SHA1
    - the correct pointer is `${{ github.event.pull_request.head.sha }}`
- It makes stress-testing so much easier because one doesn't have to manually input the previous artifact's ID anymore. Now you just give it a branch, a spec and how many times you want to run it. It will automatically find the nearest uberjar.
    
## How to test?
- Ci should be green
- Everything should work as it did before
- Stress test should work
- Look at the artifact name - it should contain the hash of the latest commit in a branch
- Look at the logs for "Download uberjar" job. All commit references there should be either from the current branch, or from the base branch

An example from the last run in this branch:
```
Run ./.github/actions/e2e-download-uberjar
Run actions/github-script@v6
Didn't find the artifact for b7a011749d4d07213859cb6d470bdbf69596ca0b.

Switching to the parent commit: fe04fb8b930589495652d6ead953c54612fbcbf9
Didn't find the artifact for fe04fb8b930589495652d6ead953c54612fbcbf9.

Switching to the parent commit: 22edc5e980bba80d2d2117f0d6046afc8396b857
Didn't find the artifact for 22edc5e980bba80d2d2117f0d6046afc8396b857.

Switching to the parent commit: e09c1f186bae3671ef5c837303da32007740fedc
Didn't find the artifact for e09c1f186bae3671ef5c837303da32007740fedc.

Switching to the parent commit: 98c3fb492960c01e7531affc0abb5c4cad3c1796
```

### DIFF
--- a/.github/actions/e2e-download-uberjar/action.yml
+++ b/.github/actions/e2e-download-uberjar/action.yml
@@ -4,11 +4,21 @@ inputs:
   edition:
     description: Metabase edition.
     required: true
+  was-built:
+    description: Was the artifact built during the current workflof run?
+    required: true
 
 runs:
   using: "composite"
   steps:
+    - uses: actions/download-artifact@v3
+      if: ${{ inputs.was-built == 'true' }}
+      name: Retrieve uberjar artifact for ${{ inputs.edition }}
+      with:
+        name: metabase-${{ inputs.edition }}-${{ github.event.pull_request.head.sha || github.sha }}-uberjar
+
     - name: Download previously stored uberjar
+      if: ${{ inputs.was-built == 'false' }}
       uses: actions/github-script@v6
       with:
         script: | # js
@@ -66,6 +76,7 @@ runs:
           fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/mb.zip`, Buffer.from(download.data));
 
     - name: Unzip Metabase artifact containing an uberjar
+      if: ${{ inputs.was-built == 'false' }}
       run: unzip mb.zip
       shell: bash
 
--- a/.github/actions/e2e-download-uberjar/action.yml
+++ b/.github/actions/e2e-download-uberjar/action.yml
@@ -28,7 +28,7 @@ runs:
           };
 
           async function getArtifact(commit, depth = 0) {
-            if (depth > 10) {
+            if (depth > 20) {
               throw new Error("Couldn't find the artifact!");
             }
 
--- a/.github/actions/e2e-download-uberjar/action.yml
+++ b/.github/actions/e2e-download-uberjar/action.yml
@@ -11,7 +11,7 @@ runs:
     - name: Download previously stored uberjar
       uses: actions/github-script@v6
       with:
-        script: |
+        script: | # js
           const fs = require('fs');
           const execSync = require("child_process").execSync;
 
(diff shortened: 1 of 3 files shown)


iethree:
this block from the release script does the same thing as the curl bit, but it's so much cleaner. would need some adaptation for use in e2e.

https://github.com/metabase/metabase/blob/master/.github/workflows/release.yml#L86-L109

nemanjaglumac:
> this block from the release script does the same thing as the curl bit, but it's so much cleaner. would need some adaptation for use in e2e.
> 
> https://github.com/metabase/metabase/blob/master/.github/workflows/release.yml#L86-L109

True.
I'll work on converting `curl` logic to JS.

Btw @iethree it seems that the logic you linked to is basically a re-implementation of https://github.com/actions/download-artifact. Is there anything stopping us from using `actions/download-artifact@v3`?

iethree:
> > this block from the release script does the same thing as the curl bit, but it's so much cleaner. would need some adaptation for use in e2e.
> > [`master`/.github/workflows/release.yml#L86-L109](https://github.com/metabase/metabase/blob/master/.github/workflows/release.yml?rgh-link-date=2023-10-18T07%3A30%3A47Z#L86-L109)
> 
> True. I'll work on converting `curl` logic to JS.
> 
> Btw @iethree it seems that the logic you linked to is basically a re-implementation of [actions/download-artifact](https://github.com/actions/download-artifact). Is there anything stopping us from using `actions/download-artifact@v3`?

IIRC, download-artifact only works within a workflow, not between workflows. I got the JS implementation from the [github actions docs here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow).

.github/actions/e2e-download-uberjar/action.yml
```diff
@@ -0,0 +1,75 @@
...
+
+          const baseConfig = {
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+          };
+
+          async function getArtifact(commit, depth = 0) {
+            if (depth > 10) {
+              throw new Error("Couldn't find the artifact!");
+            }
```
iethree, line 33: i'm not sure if 10 is deep enough, especially since we're fetching 20. I can easily imagine pushing 10 different tes-only commits to make tests stop flaking.

iethree (approved):
Very nice. this cleans things up so much in this workflow 👏

.github/actions/e2e-download-uberjar/action.yml
nemanjaglumac: Increased to 20.

PR #34655, opened 2023-10-13
Conditionally start Docker containers for E2E tests

nemanjaglumac:
One more task down in https://github.com/metabase/metabase/issues/34641

### The Problem
To the best of my knowledge, it's not possible to conditionally control `service` containers[^1].

- This meant that we were spinning up every single container even if only one group (out of 20+ total groups) needed it. It puts a lot of strain on already weak default CI runners, and it degrades the overall performance.
- It also meant a lot of code duplication between workflows

## The Solution
This PR moves all the logic to a composite action[^2].
That way we have:
1. much cleaner DRY code
2. the ability to conditionally run certain containers using any context[^3] and/or expression[^4] available to the workflow.

## Temporary Tradeoff/Downside
Until I resolve how to run tests that depend on QA Databases in a smart and granular way, snapshots are either going to [run either for Mongo or for Postgres and MySQL](https://github.com/metabase/metabase/pull/34655/commits/6e64817e26f0258328f1a8f04455f88f7c7aa39c). This was a dumb and dirty tradeoff I made to unblock this PR.
It shouldn't affect CI at all, but it might introduce some **temporary** gymnastics when running QA DB tests locally.

## Logs
This is what logs look like now. Look for "Prepare Docker containers" step.

Example for `e2e-tests-admin-2-ee` group
```sh
Run ./.github/actions/e2e-prepare-containers
  with:
    username: ***
    password: ***
    maildev: true
    openldap: true
    postgres: true
    mysql: true
    mongo: false
```

Example for `e2e-tests-mongo-ee` group
```sh
Run ./.github/actions/e2e-prepare-containers
  with:
    username: ***
    password: ***
    maildev: true
    openldap: false
    postgres: false
    mysql: false
    mongo: true
```
[^1]: https://docs.github.com/en/actions/using-containerized-services/about-service-containers
[^2]: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
[^3]: https://docs.github.com/en/actions/learn-github-actions/contexts
[^4]: https://docs.github.com/en/actions/learn-github-actions/expressions

### DIFF
--- a/.github/workflows/e2e-stress-test-flake-fix.yml
+++ b/.github/workflows/e2e-stress-test-flake-fix.yml
@@ -30,6 +30,8 @@ jobs:
       MB_SNOWPLOW_URL: "http://localhost:9090" # Snowplow micro
       TZ: US/Pacific # to make node match the instance tz
     steps:
+      - uses: actions/checkout@v3
+
       - name: Prepare Docker containers
         uses: ./.github/actions/e2e-prepare-containers
         with:
@@ -41,7 +43,6 @@ jobs:
           mysql: true
           mongo: true
 
-      - uses: actions/checkout@v3
       - name: Download Metabase uberjar from a previously stored artifact
         run: |
           curl -L \
--- a/.github/workflows/e2e-stress-test-flake-fix.yml
+++ b/.github/workflows/e2e-stress-test-flake-fix.yml
@@ -29,50 +29,18 @@ jobs:
       MB_SNOWPLOW_AVAILABLE: true
       MB_SNOWPLOW_URL: "http://localhost:9090" # Snowplow micro
       TZ: US/Pacific # to make node match the instance tz
-    services:
-      maildev:
-        image: maildev/maildev:2.0.5
-        ports:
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-stress-test-flake-fix.yml
```diff
@@ -29,51 +29,20 @@ jobs:
...
+      - name: Prepare Docker containers
+        uses: ./.github/actions/e2e-prepare-containers
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          maildev: true
+          openldap: true
+          postgres: true
+          mysql: true
+          mongo: true
```
iethree, line 44: 😍

iethree (approved):
This is beautiful. Unlocks so much potential for granularity in which containers we run for the whole matrix. The cypress setup conditional is a little messy for now, but that doesn't really bother me (we have so few mongo tests that we run locally)

🏆  🏅  ⭐

PR #33898, opened 2023-09-14
Use question from state for navbar instead of api call

deniskaber:
Closes https://github.com/metabase/metabase/issues/31905

### Description

This removes arbitrary card reload in QueryBuilder.
Also: improves our entities fetch caching by waiting for a loading query with same parameters. This leads to deduplication of the same api requests.

### How to verify

1. Create a new question or model, save it.
2. Refresh the page.
3. You should not see 2 same api requests to load your question/model.
4. You should see a collection where your question/model is saved to as selected in the left menu.

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

Issue #31905, opened 2023-06-27
Duplicate requests to /api/card from Query Builder


iethree:
### Describe the bug

Regression between 45 and 46.

The query builder makes 2 identical request to GET `/api/card/{id}` on load.

![Screen Shot 2023-06-27 at 9 58 23 AM](https://github.com/metabase/metabase/assets/30528226/c2a2ef82-ba5e-4f71-9aa4-4b9765800e02)

### To Reproduce

- Make any question or model
- open the network inspector
- refresh the page
- see 2 of the same GET request

### Information about your Metabase installation

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.name": "OpenJDK Runtime Environment",
    "java.runtime.version": "18.0.1+0",
    "java.vendor": "Homebrew",
    "java.vendor.url": "https://github.com/Homebrew/homebrew-core/issues",
    "java.version": "18.0.1",
    "java.vm.name": "OpenJDK 64-Bit Server VM",
    "java.vm.version": "18.0.1+0",
    "os.name": "Mac OS X",
    "os.version": "12.4",
    "user.language": "en",
    "user.timezone": "America/Denver"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
    "application-database-details": {
      "database": {
        "name": "H2",
        "version": "2.1.212 (2022-04-09)"
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
        "version": "2.1.212 (2022-04-09)"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2023-06-14",
      "tag": "v0.46.5",
      "branch": "release-x.46.x",
      "hash": "272e3e6"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

not user visible, an under-the-hood performance issue

WiNloSt:
The first call is a little hard to track where it originates from but the second request is from here.
[...]

PR #34661, opened 2023-10-13
Fix entities fetch caching

deniskaber:
Relates to https://github.com/metabase/metabase/issues/31905

### Description

This is additional fix for https://github.com/metabase/metabase/pull/33898 
More context - https://metaboat.slack.com/archives/C057T1QTB3L/p1697183111294489

There is a race condition: 
Let's say we have two fetch requests to load entities list (e.g. databases), 
The second request waits for the first one to be loaded, 
and then for the second request we waited extra 10ms to make sure state gets recalculated.

https://github.com/metabase/metabase/blob/ed00657b6807960b31c386175426bcfad5fb456f/frontend/src/metabase/lib/redux/utils.js#L327-L333

But as soon as first request get loaded, it updates request state to be "loaded"

https://github.com/metabase/metabase/blob/ed00657b6807960b31c386175426bcfad5fb456f/frontend/src/metabase/lib/redux/utils.js#L246-L253

For this 10ms period we are in the incorrect state due to different request states.

### How to verify

1. Create a new question or model, save it.
2. Refresh the page.
3. You should not see 2 same api requests to load your question/model.
4. You should see a collection where your question/model is saved to as selected in the left menu.

### Checklist

### DIFF
--- a/frontend/src/metabase/common/hooks/use-entity-list-query/use-entity-list-query.unit.spec.tsx
+++ b/frontend/src/metabase/common/hooks/use-entity-list-query/use-entity-list-query.unit.spec.tsx
@@ -102,9 +102,9 @@ const TestInnerComponent = () => {
   );
 };
 
-const setup = ({ databases = [TEST_DB], tables = [TEST_TABLE] } = {}) => {
-  setupDatabasesEndpoints(databases);
-  setupTablesEndpoints(tables);
+const setup = () => {
+  setupDatabasesEndpoints([TEST_DB]);
+  setupTablesEndpoints([TEST_TABLE]);
   return renderWithProviders(<TestComponent />);
 };
 
@@ -179,10 +179,11 @@ describe("useEntityListQuery", () => {
   });
 
   it("should not remove loader in case second api call is cached", async () => {
-    const dbsResponse = [TEST_DB];
     fetchMock.get(
       "path:/api/database",
-      delay(100).then(() => dbsResponse),
+      delay(100).then(() => {
+        return [TEST_DB];
+      }),
       { overwriteRoutes: true },
     );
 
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/components/LoadingAndErrorWrapper/LoadingAndErrorWrapper.jsx
```diff
@@ -134,7 +136,11 @@ export default class LoadingAndErrorWrapper extends Component {
       return Children.only(children);
     }
     return (
-      <div className={this.props.className} style={this.props.style}>
+      <div
+        className={this.props.className}
+        style={this.props.style}
```
kamilmielnik, line 141: While we're at it, we can use destructuring for `className` and `style` like for all the other props.

frontend/src/metabase/common/hooks/use-entity-list-query/use-entity-list-query.unit.spec.tsx
```diff
@@ -169,4 +177,41 @@ describe("useEntityListQuery", () => {
...
+
+    await delay(100);
+
+    expect(fetchMock.calls("path:/api/database")).toHaveLength(1);
+    expect(screen.queryByTestId("loading-spinner")).not.toBeInTheDocument();
+
+    await delay(0); // trigger delay to make sure state has been updated
+
+    expect(fetchMock.calls("path:/api/database")).toHaveLength(1);
+    expect(screen.queryByTestId("loading-spinner")).not.toBeInTheDocument();
```
kamilmielnik, line 215: Is this needed? Because we already `await delay(100)` above, which should be enough for request(s) to start/finish.

kamilmielnik (changes requested):
New test does not pass

kamilmielnik (approved):
Looks good, works well and fixes [the race condition I encountered in tests](https://metaboat.slack.com/archives/C057T1QTB3L/p1697183111294489) :+1:

ranquild (approved):

PR #34664, opened 2023-10-13
Add metabase_field semantic_type migration

### LINKED ISSUE #18754: Log filled with warnings about deprecated sematic type "Number"
**Describe the bug**
When running older instances that made use of the field type `Number`, then newer versions will litter the log with:
```
2021-10-29 15:00:01,401 WARN models.field :: Invalid Field :semantic_type :type/Number: falling back to null
```

Manual workaround: `update metabase_field set semantic_type=null where semantic_type='type/Number';`

**Expected behavior**
Either create a migration, which changes the semantic_type=null or add ignore to this specific warning, so it doesn't show in log.

Tested 0.40.5 and 0.41.1 - regression since 0.40.0 #15994

lbrdnk:
Closes #18754

### Description

Sets semantic_type to null for metabase_field if value was type/Number, to avoid log warnings.

### How to verify

Run newly added test.

### Checklist


test/metabase/db/schema_migrations_test.clj
```diff
@@ -1156,3 +1156,29 @@
         (db.setup/migrate! db-type data-source :down 47)
         (testing "Rollback to the previous version should restore the column column, and set the default color value"
           (is (= "#31698A" (:color (t2/select-one :model/Collection :id collection-id)))))))))
+
+(deftest set-number-semantic-type-null-test
```
metamben, line 1160: I feel like this is an overkill.
- The thing it tests is quite trivial and it's not huge problem if it doesn't work.
- The thing that  is potentially dangerous (e.g., that it doesn't update a bunch of unrelated things) is not being tested.
- This is not a test we need to prevent regressions or specify some behavior.
- Migration tests are particularly slow.
- Manual tests (for h2, postgres, and mysql if you want to be very through) would be enough and possibly better.

metamben (commented):
I suggest removing the test. Feel free to get a second opinion if in doubt.

test/metabase/db/schema_migrations_test.clj
lbrdnk: Thank you for the thoughtful feedback. I will mind that while writing migrations in future. As of now, test has been removed, as advised, and modified functionality tested manually for postgres, mysql and h2 as follows:

1. Start app.
2. Migrate down using `(dev/migrate! :down 47)`.
3. Set semantic_type for all rows having it null to 'type/Number'
4. Migrate up using `(dev/migrate! :up 48)`
5. Check whether 'type/Number' fields are reset to null.

metamben (approved):

gonchik:
Hi! does that PR in the 0.47.4 release ? :) 
or it's gonna be in 0.47.5?

PR #34674, opened 2023-10-13
34618 splitting breakouts

### LINKED ISSUE #34618: Splitting multidimension questions that only have 1 metric
### Describe the bug

When creating a multi dimension question, there is a good chance that metabase may decide to automatically split the y-axis even though values are fairly close together. This can lead to some misleading charts as there is only 1 metric being shown, but with 2 very different scales applies

### To Reproduce

1. Go to New -> Question -> Orders
2. Count by Orders -> Created At: Month, User -> State
3. Filter by Orders -> Created At between September 1st, and October 31st 2023
4. Set the Viz type to bar chart, and notice that there are 2 y axis.

image

(Full size bar which is only 1)
image

### Expected behavior

There should only be 1 y axis

### Logs

_No response_

```JSON
Current Master
```

### Severity

Can create some misleading charts if you're not careful. I'd say it's decently severe

### Additional context

_No response_

npfitz:
Closes https://github.com/metabase/metabase/issues/34618

### Description
Depending on the number of series in your breakout, it's possible to have a single card want to split the Y axis even though there is only 1 metric being. The change here will have `shouldSplitYAxis` return false if a series is only made up of 1 card, and has 1 metric. This covers the case where you have a question that is a breakout with many series, but means that questions with multiple metrics, or series that have been added together can still split.

### How to verify
The linked issue gives an example of a question that has multiple dimensions, but would split even though there is only 1 metric. This change fixes that case, but should leave the other cases mentioned above untouched.

### Demo
This is a screenshot of a dashboard containing the examples of:
- A card with an added series (should split if appropriate)
- A card with a question that has multiple aggregations (should split if appropriate)
- A card with a breakout (should not split)
- A card with a breakout, and an added series with a breakout (should split if appropriate)

**Before**

**After**

The bottom left card (single series breakout) is now split, which we want.

The last case is splitting, but the split it's coming up with isn't very ideal :/

### Checklist

### DIFF
--- a/e2e/test/scenarios/visualizations/bar_chart.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/bar_chart.cy.spec.js
@@ -413,6 +413,7 @@ describe("scenarios > visualizations > bar chart", () => {
     cy.get(".value-labels").should("contain", "13");
     cy.get(".value-labels").should("contain", "19");
 
+    cy.log("Should not produce a split axis graph (#34618)");
     cy.get(".axis.yr").should("not.exist");
   });
 });
--- a/e2e/test/scenarios/visualizations/bar_chart.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/bar_chart.cy.spec.js
@@ -5,6 +5,8 @@ import {
   getDraggableElements,
   moveColumnDown,
   popover,
+  visitDashboard,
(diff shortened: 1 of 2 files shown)


EmmadUsmani (approved):
Looks good! I added separate before and after screenshots to the description from more clarity.

PR #34678, opened 2023-10-14
Fix quick-filter drill returning extra operators (#34445)

camsaul:
Fixes #34445

As part of this I got tired of trying of the super slow feedback loop trying to run tests in `frontend/src/metabase-lib/drills.unit.spec.ts` so I just adapted the `availableDrillThrus` test case there to Clojure(Script). Now when fixing all of the other open drill bugs we already have a test case ready to go in Clojure

### DIFF
--- a/src/metabase/lib/drill_thru.cljc
+++ b/src/metabase/lib/drill_thru.cljc
@@ -63,24 +63,29 @@
 
 ;;; TODO: Missing drills: automatic insights, format.
 (def ^:private available-drill-thru-fns
-  [#'lib.drill-thru.distribution/distribution-drill
-   #'lib.drill-thru.column-filter/column-filter-drill
-   #'lib.drill-thru.foreign-key/foreign-key-drill
-   #'lib.drill-thru.object-details/object-detail-drill
-   #'lib.drill-thru.pivot/pivot-drill
-   #'lib.drill-thru.quick-filter/quick-filter-drill
-   #'lib.drill-thru.sort/sort-drill
-   #'lib.drill-thru.summarize-column/summarize-column-drill
-   #'lib.drill-thru.summarize-column-by-time/summarize-column-by-time-drill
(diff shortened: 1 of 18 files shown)


src/metabase/lib/drill_thru.cljc
```diff
@@ -63,24 +63,29 @@
...
-(mu/defn ^:private all-contexts :- [:sequential {:min 1} ::lib.schema.drill-thru/context]
+  "Some drill thru functions are expected to return drills for just the specified `:column`; others are expected to
+  ignore that column and return drills for all of the columns specified in `:dimensions`.
+  `:return-drills-for-dimensions?` specifies which type we have."
+  [{:f #'lib.drill-thru.distribution/distribution-drill,                         :return-drills-for-dimensions? true}
+   {:f #'lib.drill-thru.column-filter/column-filter-drill,                       :return-drills-for-dimensions? true}
+   {:f #'lib.drill-thru.foreign-key/foreign-key-drill,                           :return-drills-for-dimensions? true}
+   {:f #'lib.drill-thru.object-details/object-detail-drill,                      :return-drills-for-dimensions? true}
+   {:f #'lib.drill-thru.pivot/pivot-drill,                                       :return-drills-for-dimensions? true}
+   {:f #'lib.drill-thru.quick-filter/quick-filter-drill,                         :return-drills-for-dimensions? false}
```
camsaul, line 74: Oh yeah also apparently some drills are supposed to look at `:dimensions` and some are supposed to ignore it. This is an icky hack but I think it's ok for now until I fix the remaining drill thru bugs. Then once I wrap my head around how this is actually supposed to work maybe we can clean this up a bit e.g. have the underlying drill thru functions decide whether or not to handle `:dimensions` instead of defining it here.

metamben (approved):
Can't wait to use this! 👍

PR #34685, opened 2023-10-16
update the team label names in .github/team.json

darksciencebase:
Remove the emojis from the team labels. With the emojis, the "auto-add to project" github automated workflow doesn't work, and we rely on it for the #bug-triage channel on slack. After several attempts of escaping the emojis, they've been removed from the labels, and now have to be removed here as well for the auto-assignment of PRs to continue working.
[ci skip]

### DIFF
--- a/.github/team.json
+++ b/.github/team.json
@@ -2,7 +2,7 @@
   "teams": [
     {
       "name": "Team 42",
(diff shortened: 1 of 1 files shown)


NevRA (approved):
LGTM

PR #34030, opened 2023-09-21
PoC for explicitization of dynamic vars

escherize:

PR #34694, opened 2023-10-16
Add Kondo rule to prevent dynamic vars from taking over

tsmacdonald:
See also: #34030

Rather than undertake a huge refactor right now, I thought it'd be wiser to add a Kondo rule to prevent the problem from getting worse, then gradually refactor out the exceptions (as Bryan started to do in #34030).

I'm open to pushback about exactly which namespaces are or are not included, but this set seemed like a sensible place to start.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -658,6 +658,8 @@
    :name    printable-namespaces}
   {:pattern "^metabase\\.lib\\.*"
    :name    metabase-lib}
+  {:pattern "^metabase\\.(?!(api|query-processor|server)).*$"
+   :name    non-api-namespaces}
   ;; Basically everything except `metabase.driver.` or `metabase.test-data.` except for the drivers explicitly called
   ;; out below. Remove these drivers from the regex once we switch to Honey SQL 2.
   ;;
@@ -744,6 +746,15 @@
      toucan.models    {:message "Don't use application database inside MLv2 code"}
      toucan.util.test {:message "Don't use application database inside MLv2 code"}
      toucan2.core     {:message "Don't use application database inside MLv2 code"}}}}
+
+  non-api-namespaces
+  {:linters
+   {:discouraged-var
+    {metabase.api.common/*current-user*                 {:message "Only use API-related dynamic variables in the API namespaces"}
+     metabase.api.common/*current-user-id*              {:message "Only use API-related dynamic variables in the API namespaces"}
+     metabase.api.common/*current-user-permissions-set* {:message "Only use API-related dynamic variables in the API namespaces"}
+     metabase.api.common/*is-superuser?*                {:message "Only use API-related dynamic variables in the API namespaces"}
+     metabase.api.common/*is-group-manager?*            {:message "Only use API-related dynamic variables in the API namespaces"}}}}
 
   qp-and-driver-source-namespaces
   {:linters
--- a/modules/drivers/oracle/test/metabase/driver/oracle_test.clj
+++ b/modules/drivers/oracle/test/metabase/driver/oracle_test.clj
@@ -382,11 +382,12 @@
       (mt/with-env-keys-renamed-by #(str/replace-first % "mb-oracle-ssl-test" "mb-oracle-test")
         ;; need to get a fresh instance of details to pick up env key changes
         (let [ssl-details  (#'oracle.tx/connection-details)
-              orig-user-id api/*current-user-id*]
+              orig-user-id #_{:clj-kondo/ignore [:discouraged-var]} api/*current-user-id*]
           (testing "Oracle can-connect? with SSL connection"
             (is (driver/can-connect? :oracle ssl-details)))
           (testing "Sync works with SSL connection"
             (binding [sync-util/*log-exceptions-and-continue?* false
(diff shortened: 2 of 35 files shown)


tsmacdonald:
> [if] the problem we're trying to solve is the poor developer experience with the REPL, we could solve it by assigning a default user when in development (just like we have a default DB).

To me the REPL experience is secondary. It's (usually) easy enough to use `mw.session/with-user`. My bigger complaint is more about the discoverability of the current user/perms being necessary. It's hidden state, and docstrings etc. don't always make it clear if a "return all the Foos" function really does return all the Foos or only returns the Foos that the current user has permission to see.

I'm not saying we abolish the dynamic vars—they certainly do have use, and there's stuff that would be a real PITA to do without them—but I do think we should try to limit where they're used. It's one strand of the codebase that doesn't need to be braided into everything else.

calherries:
> It's hidden state, and docstrings etc. don't always make it clear if a "return all the Foos" function really does return all the Foos or only returns the Foos that the current user has permission to see.

This is a real problem, yes. It's why we shouldn't use dynamic scope unless it's a huge pain not to. But sometimes avoiding dynamic scope is not worth the alternative downside of having to thread permissions through everything. It's not too troublesome for me to remember that most functions can depend on a database and the current user + their permissions, and that I have to be careful about that.

> I do think we should try to limit where they're used.

I agree with this, but the implementation of this kondo rule doesn't seem like the way to do it. Sometimes you will have to write model code where it is much easier to use dynamic scope than avoid it e.g. [:target hydration method](https://github.com/metabase/metabase/blob/e7ec68f44dda9132583002ee1257900665aced32/src/metabase/models/field.clj#L320). Sometimes I would rather just use the dynamic scope, and it's hard to see why namespaces are a good way to divide code by whether dynamic scope is useful vs not. Secondly the kondo rule won't even eliminate the use of dynamic scope, as I explained above. It _might_ discourage some uses of dynamic scope that would happen otherwise.

I suggest we continue to discourage use of dynamic scope in PR reviews and refactoring efforts, rather than enforce it with a namespace-level kondo rule.

.clj-kondo/config.edn
```diff
@@ -658,6 +658,8 @@
    :name    printable-namespaces}
   {:pattern "^metabase\\.lib\\.*"
    :name    metabase-lib}
+  {:pattern "^metabase\\.(?!(api|query-processor|server)).*$"
+   :name    non-api-namespaces}
```
camsaul, line 662: I think you should be explicit about calling this the REST API rather than just the API, because drivers for example have an API, but that's not what we're talking about here

```suggestion
   :name    non-rest-api-namespaces}
```

.clj-kondo/config.edn
```diff
@@ -745,6 +747,15 @@
      toucan.util.test {:message "Don't use application database inside MLv2 code"}
      toucan2.core     {:message "Don't use application database inside MLv2 code"}}}}
 
+  non-api-namespaces
+  {:linters
+   {:discouraged-var
+    {metabase.api.common/*current-user*                 {:message "Only use API-related dynamic variables in the API namespaces"}
```
camsaul, line 753: Would it make more sense to do `:discouraged-namespace` for `metabase.api.common` instead? I think you could actually maybe even create a namespace pattern like `metabase\\.api\\..*` and use that in `:discouraged-namespaces`.

Using `metabase.api.user/some-helper-function` inside `metabase.models.permissions-group` or something like that is hardly any better than using the dynamic variables

escherize: That makes sense. We could use the [:ns-groups](https://github.com/clj-kondo/clj-kondo/blob/master/doc/linters.md#discouraged-namespace) feature.

escherize (commented):
I think not allowing requiring `metabase.api.*` from `metabase.model.*` is a step in the right direction and has these benefits:

- Discourages use of `api/*` dynamic vars
- It seems like metabase.api.* should usually require metabase.model.* and not the other way around
  - So adding groups for api* and model* namespaces and enforcing that seems good.
- Helps new devs understand what api and models are, and avoids circular dependencies

This implementation would not be an outright ban: exceptions to it are fine. The
`#_{:clj-kondo/ignore ...}` can just serve as a reminder to proceed with
caution.

---

Having a couple dynamic vars increases ergonomics. No-one wants to pass around a db-object, a printer-object, a server-object, and an xyz-object in order to do some basic operations. That said I think we should discourage their use in many situations.

The best situations to use them are cross-cutting concerns with things that will barely ever change, but that we want to access conveniently. Things like db-connections, loggers, and dev options should be dynamic vars. But, the current user's info from the request aren't that.

They make the code more difficult to reason about and refactor, especially when call stacks get deep and their branching-factor is wide due to necessary complexity. Dynamic vars multiply that complexity. So limiting their use is preferable.

---

I agree with @calherries that it's hard to say which namespaces they should be used in. So it makes sense to be able to bend the rule when needed by clj-kondo/ignoring it.

---

I do know @noahmoss recently wrote some audit logging functions, where the `api/*current-user-id*` is an _optional_ thing, and it *can be passed in as a value instead*. It is simple, and doesn't really make decisions based on their value. Using them like an optional default, fallback, or for logging is a worthwhile use-case IMO.

Issue #34672, opened 2023-10-13
[Epic] Static line/area/bar/combo scope for x.48


alxnddr:
**Links**
- full project estimates beyond x.48: https://www.notion.so/metabase/Line-Area-Bar-Combo-ECharts-migration-4398d13993304e6eac174f15955631ef
- tech design doc: https://www.notion.so/metabase/ECharts-tech-design-15a2fc793f164ddb80c9723745c17d0a
- feature branch: `echarts-static-combo-chart`

Closes:
- https://github.com/metabase/metabase/issues/33798
- https://github.com/metabase/metabase/issues/33705
- https://github.com/metabase/metabase/issues/30807
- https://github.com/metabase/metabase/issues/30609
- https://github.com/metabase/metabase/issues/30551
- https://github.com/metabase/metabase/issues/30325
- https://github.com/metabase/metabase/issues/30315
- https://github.com/metabase/metabase/issues/30142
- https://github.com/metabase/metabase/issues/29852
- https://github.com/metabase/metabase/issues/29836
- https://github.com/metabase/metabase/issues/29281
- https://github.com/metabase/metabase/issues/28822
- https://github.com/metabase/metabase/issues/27942
- https://github.com/metabase/metabase/issues/26769
- https://github.com/metabase/metabase/issues/25896
- https://github.com/metabase/metabase/issues/23743
- https://github.com/metabase/metabase/issues/20559

**Implementation Plan**

***Milestone 1: Replace the existing static combo chart***
This milestone represents smallest shippable scope that fully replaces existing charts, do not introduce regressions, fixes some of issues

```[tasklist]
### Line/area/bar/combo
- [ ] Clean up the prototype
- [ ] Axes label positioning
- [ ] Implement series settings
- [ ] Add static legend
- [ ] BE: pass combined cards data to JS
```

### Combined scalar
- [ ] BE: pass combined scalars cards + datasets to JS
- [ ] FE: render combined scalars as bars
```

***Milestone 2: Full feature parity***
```[tasklist]
### Tasks
- [ ] Implement normalized stacking
- [ ] Goal line
- [ ] Trend line 
- [ ] FE: Render timeline events
- [ ] BE: Provide timeline events data
```

PR #34703, opened 2023-10-16
combo static chart foundation with echarts

alxnddr:
Epic: https://github.com/metabase/metabase/issues/34672

### Description

This PR polishes the initial combo chart prototype. It uses ECharts only for the static combo chart so far. There are many non-working features for now. Supported ones:
- Multiple series
- Categorical, timeseries, numeric x-axis
- Custom series colors
- Show values on chart
- Specify series display type: line/area/bar

### How to verify

1. Create a combo chart question, specify series `display` manually (defaults are not implemented)
2. Add it on a dashboard
4. Send in an email subscription
5. Confirm the chart renders

### Demo

Screenshot 2023-10-16 at 7 49 50 PM

Screenshot 2023-10-16 at 8 04 35 PM

### DIFF
--- a/frontend/src/metabase/static-viz/components/ComboChart/ComboChart.tsx
+++ b/frontend/src/metabase/static-viz/components/ComboChart/ComboChart.tsx
@@ -1,7 +1,7 @@
 import { init } from "echarts";
 import type { IsomorphicChartProps } from "metabase/static-viz/containers/StaticChart/types";
 import { getCartesianChartModel } from "metabase/visualizations/echarts/cartesian/model";
-import { buildCartesianChartOption } from "metabase/visualizations/echarts/cartesian/option";
+import { getCartesianChartOption } from "metabase/visualizations/echarts/cartesian/option";
 import { sanitizeSvgForBatik } from "metabase/static-viz/lib/svg";
 import { computeStaticComboChartSettings } from "./settings";
 
@@ -29,7 +29,7 @@ export const ComboChart = ({
     computedVisualizationSettings,
   );
 
-  const option = buildCartesianChartOption(
+  const option = getCartesianChartOption(
     chartModel,
     computedVisualizationSettings,
     renderingContext,
--- a/frontend/src/metabase/visualizations/echarts/cartesian/option/axis.ts
+++ b/frontend/src/metabase/visualizations/echarts/cartesian/option/axis.ts
@@ -48,8 +48,8 @@ export const getXAxisType = (settings: ComputedVisualizationSettings) => {
 
 export const buildDimensionAxis = (
   settings: ComputedVisualizationSettings,
-  renderingContext: RenderingContext,
   column: RemappingHydratedDatasetColumn,
+  renderingContext: RenderingContext,
 ): CartesianAxisOption => {
   const { getColor } = renderingContext;
   const axisType = getXAxisType(settings);
@@ -99,8 +99,8 @@ export const buildDimensionAxis = (
 
 const buildMetricsAxes = (
   settings: ComputedVisualizationSettings,
-  renderingContext: RenderingContext,
   column: RemappingHydratedDatasetColumn,
+  renderingContext: RenderingContext,
 ): CartesianAxisOption[] => {
   const { getColor } = renderingContext;
   const formatter = (value: unknown) => {
@@ -111,7 +111,6 @@ const buildMetricsAxes = (
   };
 
   // TODO: support Y-axis split
-
   return [
     {
       ...getAxisNameDefaultOption(
@@ -127,9 +126,7 @@ const buildMetricsAxes = (
       },
       axisLabel: {
         ...getTicksDefaultOption(renderingContext),
-        // TODO: figure out EChart types
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
+        // @ts-expect-error TODO: figure out EChart types
         formatter: (value: string) => {
           return formatter(value);
         },
@@ -146,13 +143,13 @@ export const buildAxes = (
   return {
     xAxis: buildDimensionAxis(
       settings,
-      renderingContext,
       cardSeriesModel.dimension.column,
+      renderingContext,
     ),
     yAxis: buildMetricsAxes(
       settings,
-      renderingContext,
       cardSeriesModel.metrics[0].column,
+      renderingContext,
     ),
   };
 };
--- a/frontend/src/metabase/visualizations/echarts/cartesian/option/index.ts
+++ b/frontend/src/metabase/visualizations/echarts/cartesian/option/index.ts
@@ -7,18 +7,18 @@ import type {
 } from "metabase/visualizations/types";
 import { buildAxes } from "metabase/visualizations/echarts/cartesian/option/axis";
 
-export const buildCartesianChartOption = (
+export const getCartesianChartOption = (
   chartModel: CartesianChartModel,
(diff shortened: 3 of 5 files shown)


alxnddr:
* master
  * **PR #34606** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34606" target="_blank">Graphite</a> 
    * **PR #34703** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34703" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/34703?utm_source=stack-comment).

frontend/src/metabase/static-viz/containers/StaticChart/types.ts
```diff
@@ -8,3 +10,8 @@ export interface StaticChartProps {
   options: any;
   colors?: ColorPalette;
 }
+
+export interface IsomorphicChartProps {
```
EmmadUsmani, line 14: @alxnddr Can we move this to the previous PR? Pie chart needs it as well

alxnddr: Done

frontend/src/metabase/static-viz/components/ComboChart/ComboChart.tsx
```diff
@@ -0,0 +1,47 @@
...
+    rawSeries,
+    renderingContext,
+  );
+
+  const chartModel = getCartesianChartModel(
+    rawSeries,
+    computedVisualizationSettings,
+  );
+
+  const option = buildCartesianChartOption(
```
EmmadUsmani, line 32: Nit: I think we can name this `getCartesianChartOption` for consistency with the model function.

For `computeStaticComboChartSettings`, I think it makes sense to use the verb "compute" to distinguish between stored and computed viz settings, but in general I think "get" is a good default.

frontend/src/metabase/visualizations/echarts/cartesian/option/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  RenderingContext,
+} from "metabase/visualizations/types";
+import { buildAxes } from "metabase/visualizations/echarts/cartesian/option/axis";
+
+export const buildCartesianChartOption = (
+  chartModel: CartesianChartModel,
+  settings: ComputedVisualizationSettings,
+  renderingContext: RenderingContext,
+): EChartsOption => {
+  const series = buildOptionMultipleSeries(
```
EmmadUsmani, line 15: Let's name this `echartsSeries` to avoid confusion with our own series/seriesModels (already the naming has gotten very confusing).

I also think we can define type aliases (e.g. type EChartsSeries = EChartsOption["series"]) to get in the habit of using them.

frontend/src/metabase/visualizations/echarts/cartesian/option/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    chartModel.flatMap(cardModel => cardModel.series),
+    settings,
+    renderingContext,
+  );
+
+  const dataset = chartModel.map(cardModel => {
```
EmmadUsmani, line 21:
```suggestion
const echartsDataset: EChartsDataset = chartModel.map(cardModel => {
```

frontend/src/metabase/visualizations/echarts/cartesian/option/axis.ts
```diff
@@ -0,0 +1,158 @@
...
+      return "value";
+    default:
+      // TODO: implement histogram
+      return "category";
+  }
+};
+
+export const buildDimensionAxis = (
+  settings: ComputedVisualizationSettings,
+  renderingContext: RenderingContext,
```
EmmadUsmani, line 51: Nit: let's make `renderingContext` the last argument for consistency with `buildAxes`. In general it would be nice for it to always be the last arg.

frontend/src/metabase/visualizations/echarts/cartesian/option/axis.ts
```diff
@@ -0,0 +1,158 @@
...
+      lineStyle: {
+        color: getColor("text-dark"),
+      },
+    },
+  };
+};
+
+const buildMetricsAxes = (
+  settings: ComputedVisualizationSettings,
+  renderingContext: RenderingContext,
```
EmmadUsmani, line 102: Same as above

frontend/src/metabase/visualizations/echarts/cartesian/option/axis.ts
```diff
@@ -0,0 +1,158 @@
...
+        // @ts-ignore
+        formatter: (value: string) => {
+          return formatter(value);
+        },
+      },
+    },
+  ];
+};
+
+export const buildAxes = (
```
EmmadUsmani, line 141: Not a suggestion, but I am very pleased with how minimal the overall code for building axes is 🙂

frontend/src/metabase/visualizations/echarts/cartesian/option/axis.ts
```diff
@@ -0,0 +1,158 @@
...
+        lineStyle: {
+          type: 5,
+          color: getColor("border"),
+        },
+      },
+      axisLabel: {
+        ...getTicksDefaultOption(renderingContext),
+        // TODO: figure out EChart types
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
EmmadUsmani, line 132: You can replace the `eslint-disable-next-line` + `@ts-ignore` with just `@ts-expect-error`

EmmadUsmani (commented):
Overall it looks good! I do think there are some places though where we can improve the readability of the code.

The function implementations themselves are generally fine, but additional context surrounding them would be great. Yes technically we do have types for everything, but the types are often very nested and inferred, so it becomes hard to see the forest for the trees unless you fire up the debugger and start inspecting values. Ideally you shouldn't have to debug, the context should be in the source code.

Two ways we can do this are through unit tests, and inline comments before certain blocks of logic in more involved functions.

For example, I really liked the unit tests for `groupDataset`, being able to explicitly see the input and output of the function makes it easy to understand its purpose and what its doing. However, this same function has a long body with nested loops and conditionals, whose purposes aren't always self evident. Here it would be nice to have some comments explaining each step, [similar to what we did for the dashboard tabs reducer.]([url](https://github.com/metabase/metabase/blob/faca09c6cc1996a95493f1450592439bdbe517c4/frontend/src/metabase/dashboard/actions/tabs.ts#L136))

I think particularly the code around settings computation and data transformations could use these improvements for the longer and more complicated functions. The code for building the echarts option object is more self-evident and easy to understand, so it's not as necessary there.

test/metabase/pulse/render/js_svg_test.clj
```diff
@@ -133,7 +133,7 @@
                      :visualization_settings {}}
         svg-string  (combo-chart-string series-seqs settings)]
     (testing "It returns bytes"
-      (let [svg-bytes (js-svg/combo-chart series-seqs settings)]
+      (let [svg-bytes (js-svg/combo-chart-legacy series-seqs settings)]
```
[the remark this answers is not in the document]
alxnddr: Yep, even now it is still used for combined line/area/bar and combined scalars but I'll remove it once fully migrate to ECharts for these types

alxnddr:
Moved to this stack:
Screenshot 2023-11-02 at 1 33 17 AM

PR #34262, opened 2023-09-29
Enable expression dashboard filters

metamben:
Fixes #17775.

PR #34704, opened 2023-10-16
fix "between datetimes" filter

### LINKED ISSUE #34615: "Add a time" filter in dashboard filters not working
### Describe the bug

The "Add a time" button is now in dashboard filters ([wasn't until 47.2](https://github.com/metabase/metabase/issues/28188)) but when applied, is not filtering at all

Below there are the date-filtered values without the time filter:
Screenshot 2023-10-12 at 2 48 38 PM

And below are the date and time (I used 5 mins of difference to make it obvious)
Screenshot 2023-10-12 at 2 49 26 PM
 filtered values; they are the same

### To Reproduce

1. Add a question that has a datetime field to a dashboard
2. Add a date range filter to a dashboard
3. Click "Add a time" and select a time that should filter some of the values
4. See the error

### Expected behavior

Time filter should work

### Logs

_No response_

```JSON
- 47.3
```

### Severity

P2

### Additional context

_No response_

sansolovyov:
Date parsing code wasn't accounting for datetimes.

Resolves #34615

### DIFF
--- a/src/metabase/driver/common/parameters/dates.clj
+++ b/src/metabase/driver/common/parameters/dates.clj
@@ -238,6 +238,10 @@
 (defn- ->iso-8601-date [t]
   (t/format :iso-local-date t))
 
+(defn- ->iso-8601-date-time [t]
+  (t/format :iso-local-date-time t))
+
+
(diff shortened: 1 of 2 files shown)


tsmacdonald (approved):

piranha:
Not backporting since it depends on #34262 somewhat.

PR #34712, opened 2023-10-16
Clarify stacktraces from malli validation errors

johnswanson:
## The problem

In `metabase.util.malli.fn`, we have a macro, `defn`, that allows defining a function with input and output validation.

In pseudocode, the macro works by translating something like this:

```clojure
(mu/defn foo :- string
 [x :- int]
 (str x))
```

into something like this:

```clojure

(defn validate! [arg schema]
 ;; do some validation
 (when-not (valid? arg schema)
  (throw (ex-info "not valid" {}))))

(def foo
 (fn [x]
  (validate! x :int)
  (let [output (str x)]
    (validate! output :string)
    output)))
```

This works quite well. One issue is stacktraces. As you'd expect from looking at the toy macroexpansion above, if we call `foo` with an invalid argument, or if it results in an invalid output, we end up with a stacktrace that looks like:

```
Execution error (ExceptionInfo) at metabase.util.malli.fn/validate (fn.clj:150).
```

This makes sense, because the exception is being thrown from `validate!`, not from `foo` itself. But this is confusing - our schema validation namespace isn't really the place you should start looking for the bug.

## Solution

This change does two things: **first**, we manually edit the stacktrace of the exception when we throw it in `validate`, removing any starting entries that have a class name starting with `metabase.util.malli.fn`.

Of course, there is a potential disadvantage of this change - if we have a bug in our validation logic itself, the stacktrace will be *less* clear about the cause. I _think_ that in this case we're reasonably confident in the quality of our validation that this change is on-net good, but I'm open to the possibility that I might be wrong here - input welcome.

**Second**, instead of rewriting `mu/fn` bodies like so:

```
(let [f& ...] (fn [args] (do-validation) (f& args)))
```

Rewrite them like this:

```
(fn [args] (do-validation) ((fn [args'] ...) args))
```

As @dpsutton points out, if we have something like:

```
(def my-var (fn [x] ...))
```

stacktraces are correct (they point to `my-var`), while something like:

```
(def my-var (let [&f (fn [x] ...)] (fn [x] (&f x)))
```

results in stacktraces pointing to an anonymous function (`metabase.pulse/fn--190883/fn--190886 `).

## Performance Note

I wasn't entirely sure that this second change wouldn't have a negative performance impact, so I tested it with these toy examples:

```
(def option-1
  (let [f (fn [x] (+ x 10))]
    (fn [x]
      (f x))))

(def option-2
  (fn [x]
    ((fn [x'] (+ x 10)) x)))
```

I evaluated each option with Criterium to benchmark.

The first option (our current behavior) looked like this:

```
Evaluation count : 10184236800 in 60 samples of 169737280 calls.
      Execution time sample mean : 4.117003 ns
             Execution time mean : 4.117119 ns
Execution time sample std-deviation : 0.017140 ns
    Execution time std-deviation : 0.017686 ns
   Execution time lower quantile : 4.100963 ns ( 2.5%)
   Execution time upper quantile : 4.149603 ns (97.5%)
                   Overhead used : 1.783065 ns
```

The second option (the new behavior) looked like this:

```
Evaluation count : 13861192260 in 60 samples of 231019871 calls.
      Execution time sample mean : 2.553532 ns
             Execution time mean : 2.553510 ns
Execution time sample std-deviation : 0.019674 ns
    Execution time std-deviation : 0.019824 ns
   Execution time lower quantile : 2.524266 ns ( 2.5%)
   Execution time upper quantile : 2.599826 ns (97.5%)
                   Overhead used : 1.783065 ns
```

So overall the performance does not look to be worse and may actually be slightly better.

### DIFF
--- a/src/metabase/util/malli/fn.clj
+++ b/src/metabase/util/malli/fn.clj
@@ -2,12 +2,15 @@
   (:refer-clojure :exclude [fn])
   (:require
    [clojure.core :as core]
+   [clojure.string :as str]
    [malli.core :as mc]
    [malli.destructure :as md]
    [malli.error :as me]
    [metabase.shared.util.i18n :as i18n]
    [metabase.util.malli.humanize :as mu.humanize]
    [metabase.util.malli.registry :as mr]))
+
+(set! *warn-on-reflection* true)
 
 (defn- add-default-schemas
   "Malli normally generates wacky default schemas when you use destructuring in an argslist; this never seems to work
@@ -143,20 +146,36 @@
   use [[metabase.util.malli/disable-enforcement]] to bind this only in Clojure code."
   true)
 
+(defn- fixup-stacktrace
+  "This function removes stack trace elements that came from this namespace. When we throw validation errors, they
+  shouldn't originate from *this* namespace, they should appear to be thrown from the instrumented function itself."
+  [^Exception e]
+  (let [trace (.getStackTrace e)
+        fixed-trace (into-array StackTraceElement
+                                (drop-while
+                                 #(str/starts-with? (.getClassName %)
+                                                    ;; this is... hacky, but it works.
+                                                    (namespace ::x))
+                                            trace))]
+    (.setStackTrace e fixed-trace)))
+
 (defn- validate [error-context schema value error-type]
   (when *enforce*
     (when-let [error (mr/explain schema value)]
       (let [humanized (me/humanize error)]
-        (throw (ex-info (case error-type
-                          ::invalid-input  (i18n/tru "Invalid input: {0}" (pr-str humanized))
-                          ::invalid-output (i18n/tru "Invalid output: {0}" (pr-str humanized)))
-                        (merge
-                         {:type      error-type
-                          :error     error
-                          :humanized humanized
-                          :schema    schema
-                          :value     value}
-                         error-context)))))))
+        (throw
+         (doto (ex-info
+                (case error-type
+                  ::invalid-input  (i18n/tru "Invalid input: {0}" (pr-str humanized))
+                  ::invalid-output (i18n/tru "Invalid output: {0}" (pr-str humanized)))
+                (merge
+                 {:type      error-type
+                  :error     error
+                  :humanized humanized
+                  :schema    schema
+                  :value     value}
+                 error-context))
+           fixup-stacktrace))))))
 
 (defn validate-input
   "Impl for [[metabase.util.malli.fn/fn]]; validates an input argument with `value` against `schema` using a cached
@@ -212,19 +231,19 @@
               schemas)
          (filter some?))))
 
-(defn- input-schema->application-form [input-schema]
+(defn- input-schema->application-form [input-schema deparameterized-fn]
   (let [arg-names (input-schema-arg-names input-schema)]
     (if (varargs-schema? input-schema)
-      (list* `apply '&f arg-names)
-      (list* '&f arg-names))))
+      (list* `apply deparameterized-fn arg-names)
+      (list* deparameterized-fn arg-names))))
 
-(defn- instrumented-arity [error-context [_=> input-schema output-schema]]
+(defn- instrumented-arity [error-context [_=> input-schema output-schema] deparameterized-fn]
   (let [input-schema           (if (= input-schema :cat)
                                  [:cat]
                                  input-schema)
         arglist                (input-schema->arglist input-schema)
         input-validation-forms (input-schema->validation-forms error-context input-schema)
-        result-form            (input-schema->application-form input-schema)
+        result-form            (input-schema->application-form input-schema deparameterized-fn)
         result-form            (if (and output-schema
                                         (not= output-schema :any))
                                  `(->> ~result-form
@@ -232,15 +251,17 @@
                                  result-form)]
     `(~arglist ~@input-validation-forms ~result-form)))
 
-(defn- instrumented-fn-tail [error-context [schema-type :as schema]]
+(defn- instrumented-fn-tail [error-context
+                             [schema-type :as schema]
+                             deparameterized-fn]
   (case schema-type
     :=>
-    [(instrumented-arity error-context schema)]
+    [(instrumented-arity error-context schema deparameterized-fn)]
 
     :function
     (let [[_function & schemas] schema]
       (for [schema schemas]
-        (instrumented-arity error-context schema)))))
+        (instrumented-arity error-context schema deparameterized-fn)))))
 
 (defn instrumented-fn-form
   "Given a `fn-tail` like
@@ -254,8 +275,9 @@
     (mc/-instrument {:schema [:=> [:cat :int :any] :any]}
                     (fn [x y] (+ 1 2)))"
   [error-context parsed]
-  `(let [~'&f ~(deparameterized-fn-form parsed)]
-     (core/fn ~@(instrumented-fn-tail error-context (fn-schema parsed)))))
+  `(core/fn ~@(instrumented-fn-tail error-context
+                                    (fn-schema parsed)
+                                    (deparameterized-fn-form parsed))))
 
 (defmacro fn
   "Malli version of [[schema.core/fn]]. A form like
--- a/src/metabase/util/malli/fn.clj
+++ b/src/metabase/util/malli/fn.clj
@@ -153,7 +153,7 @@
   (let [trace (.getStackTrace e)
         fixed-trace (into-array StackTraceElement
                                 (drop-while
-                                 #(str/starts-with? (.getClassName %)
+                                 #(str/starts-with? (.getClassName ^StackTraceElement %)
                                                     ;; this is... hacky, but it works.
                                                     (namespace ::x))
                                             trace))]
--- a/test/metabase/util/malli/fn_test.clj
+++ b/test/metabase/util/malli/fn_test.clj
@@ -1,10 +1,13 @@
 (ns ^:mb/once metabase.util.malli.fn-test
   (:require
+   [clojure.string :as str]
    [clojure.test :refer :all]
    [clojure.walk :as walk]
    [metabase.util.malli :as mu]
    [metabase.util.malli.fn :as mu.fn]
    [metabase.util.malli.registry :as mr]))
+
+(set! *warn-on-reflection* true)
 
 (deftest ^:parallel add-default-schemas-test
   (are [input expected] (= expected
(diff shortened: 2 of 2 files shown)


dpsutton:
Awesome PR description. 
One thing is in `validation-error?` if we might look for something a bit more concrete? I wonder if we could put something specific in the `ex-data` that we could check for instead of sniffing the namespace of the `:type` key. Or do you think checking for that is a good way to go (honest question. it sets off my spidey sense but maybe it's actually a decent check)

johnswanson:
> Awesome PR description. One thing is in `validation-error?` if we might look for something a bit more concrete? I wonder if we could put something specific in the `ex-data` that we could check for instead of sniffing the namespace of the `:type` key. Or do you think checking for that is a good way to go (honest question. it sets off my spidey sense but maybe it's actually a decent check)

@dpsutton I think your spidey-sense is being reasonable here. As one example, if we moved the namespace, it might be hard to notice that that string needs to change.

A few options...

- `(contains? #{::invalid-input ::invalid-output} type)`
- add a `:validation-error?` key to the ex-data and just check for that
- `(= (namespace type) (str *ns*))`

Of those, I think the second option makes the most sense - I'll push up a commit with that change.

escherize (commented):
Looks good!

The only thing I would consider adding is using a name-spaced keyword [here](https://github.com/metabase/metabase/pull/34712/commits/d3b187743ad4b4cfeb967890dc8ba938494198bc#diff-dc7f194c298781284c8ae17ce4d8d423f286ead0565928e6251494abd75414b8R242).  Granted having `:validation-error? true` in an ex-data is rare, but having `:metabase.util.malli/validation-error?` is basically impossible.

tsmacdonald:
I don't have anything constructive to add not covered by Dan and Bryan, but thanks for doing this! Big +1 from me

dpsutton:
Examples

source function in `metabase.pulse.clj`:

```clojure
(mu/defn foo :- :string
  [x :- :int]
  (str x))
```

### before
```clojure
pulse=> (foo "3")
Execution error (ExceptionInfo) at metabase.util.malli.fn/validate (fn.clj:150).
Invalid input: ["should be an integer"]
pulse=> (pst)
ExceptionInfo Invalid input: ["should be an integer"] {:type :metabase.util.malli.fn/invalid-input, :error {:schema :int, :value "3", :errors ({:path [], :in [], :schema :int, :value "3"})}, :humanized ["should be an integer"], :schema :int, :value "3", :fn-name metabase.pulse/foo}
	metabase.util.malli.fn/validate (fn.clj:150)
	metabase.util.malli.fn/validate (fn.clj:146)
	metabase.util.malli.fn/validate-input (fn.clj:165)
	metabase.util.malli.fn/validate-input (fn.clj:161)
	metabase.pulse/fn--190141/fn--190144 (pulse.clj:645)   ;; <- this is our function
	metabase.pulse/eval190153 (NO_SOURCE_FILE:386)
	metabase.pulse/eval190153 (NO_SOURCE_FILE:386)
	clojure.lang.Compiler.eval (Compiler.java:7177)
	clojure.lang.Compiler.eval (Compiler.java:7132)
	clojure.core/eval (core.clj:3221)
	clojure.core/eval (core.clj:3217)
	user/eval154283/fn--154286 (NO_SOURCE_FILE:8)
nil
```

### after

```clojure
pulse=> (foo "3")
Execution error (ExceptionInfo) at metabase.pulse/fn$fn (pulse.clj:645).
Invalid input: ["should be an integer"]
pulse=> (pst)
ExceptionInfo Invalid input: ["should be an integer"] {:type :metabase.util.malli.fn/invalid-input, :error {:schema :int, :value "3", :errors ({:path [], :in [], :schema :int, :value "3"})}, :humanized ["should be an integer"], :schema :int, :value "3", :validation-error? true, :fn-name metabase.pulse/foo}
	metabase.pulse/fn--190883/fn--190886 (pulse.clj:645)
	metabase.pulse/eval190890 (NO_SOURCE_FILE:391)
	metabase.pulse/eval190890 (NO_SOURCE_FILE:391)
	clojure.lang.Compiler.eval (Compiler.java:7177)
	clojure.lang.Compiler.eval (Compiler.java:7132)
	clojure.core/eval (core.clj:3221)
	clojure.core/eval (core.clj:3217)
	user/eval154283/fn--154286 (NO_SOURCE_FILE:8)
	clojure.main/repl/read-eval-print--9428/fn--9431 (main.clj:438)
	clojure.main/repl/read-eval-print--9428 (main.clj:438)
	clojure.main/repl/fn--9437 (main.clj:459)
	clojure.main/repl (main.clj:459)
```

### What we could improve

It would be nice if this reported that the function which threw the error was `metabase.pulse/foo` rather than `metabase.pulse/fn$fn`. I wonder if we can propagate the name of the function we are instrumenting to the anonymous function we are creating. Perhaps it can be the same function name and it will shadow the top level var. Or maybe we need to munge it a bit. Not sure.

johnswanson:
> It would be nice if this reported that the function which threw the error was metabase.pulse/foo rather than metabase.pulse/fn$fn. I wonder if we can propagate the name of the function we are instrumenting to the anonymous function we are creating. Perhaps it can be the same function name and it will shadow the top level var. Or maybe we need to munge it a bit. Not sure.

@dpsutton very good point, thanks for noticing that issue - the stacktrace really wasn't much clearer after my change than before it! 🤦

This should be fixed now. As you suggested, I enhanced the `mu/fn` macro to accept an optional name parameter (just like `core/fn`). If it does receive that parameter, it passes that along so the "anonymous" function gets the name. The stacktrace should now look like `metabase.pulse/fn/foo`.

Note: the failing tests are currently [failing on master](https://github.com/metabase/metabase/actions/runs/6550122518/job/17788548013).

dpsutton:
hey @johnswanson . I see you did some munging and it's looking a little scary. I looked into what we're making and want to suggest a different approach.

```clojure
mu=> (macroexpand '(mu/defn foo
                     ([x :- :int]
                      (str x))))
(def
 foo
 "Inputs: ([x :- :int])\n  Return: :any"
 (clojure.core/let
  [&f (clojure.core/fn ([x] (str x)))]
  (clojure.core/fn
   ([a]
    (metabase.util.malli.fn/validate-input
     {:fn-name (quote nocommit.mu/foo)}
     :int
     a)
    (&f a)))))
```

I'm macroexpanding the `mu/defn` form. It's fundamentally a `(def <original-name> (let [...] (fn []))` and I think this form gets a bit annoying.

And the trick is to note that `defn` works correctly and is done as `(def foo (fn [] ...))`.

We're doing the following:

```clojure
(def f2 (let [actual-f (fn f2 [x] (inc x))]
          (fn interior [x] (try (local-validate x)
                                (actual-f x)
                                (catch clojure.lang.ExceptionInfo e
                                  (when (::validation? (ex-data e))
                                    (throw (ex-info (ex-message e) (ex-data e)))))))))
mu=> (f2 3)
Execution error (ExceptionInfo) at nocommit.mu/fn$interior (mu.clj:28).
bad input
```

Which leaves us at the inner function `interior`.

But if we could switch the expansion to something like

```clojure
(def f1 (fn [x] (try (local-validate x)
                     (inc x)
                     (catch clojure.lang.ExceptionInfo e
                       (when (::validation? (ex-data e))
                         (throw (ex-info (ex-message e) (ex-data e))))))))
mu=> (f1 2)
Execution error (ExceptionInfo) at nocommit.mu/f1 (mu.clj:20).
bad input
```
We don't even have to name anything. It seems if the function is "top level" in the def the compiler does all the correct work for us.

In this example, `mu.clj:20` is pointing at the inserted `(throw (ex-info (ex-message e) (ex-data e)))`. But we can do a bit of surgery to get that even better.

```clojure
(defn fixup-stacktrace
  [^Exception e]
  (let [trace (.getStackTrace e)
        fixed-trace (into-array StackTraceElement (rest trace))]
    (.setStackTrace e fixed-trace)))

(def f1 (fn [x] (try (local-validate x)
                     (inc x)
                     (catch clojure.lang.ExceptionInfo e
                       (when (::validation? (ex-data e))
                         (throw (doto (ex-info (ex-message e) (ex-data e))
                                  fixup-stacktrace)))))))
```

We can pop an element from the stack trace so we end up back at the top of the function so the validation appears to be at the function signature level rather than at the "throw" level.

```clojure
mu=> (f1 2)
Execution error (ExceptionInfo) at nocommit.mu/f1 (mu.clj:22).
bad input
```

And line 22 is the function for me.
image


This is getting a bit nitpicky so if you find the effort is outweighed by the payoff feel free to figure out the best tradeoffs involved here.

camsaul:
The reason I didn't write the macro like this in the first place is that it's generating a huge amount of code compared to the current approach, instead of having that error logic exist once it now is duplicated for every single function we have that uses `mu/defn`. I would be interested in seeing how much this change negatively affects compiled class file size/memory usage. Probably not a gigantic difference but that sort of stuff adds up when you do it 5000 times

camsaul:
works for me

johnswanson:
@dpsutton / @escherize would you be able to take another look at this when you have a chance? 🙏

src/metabase/util/malli/fn.clj
```diff
@@ -143,20 +146,36 @@
...
+(defn- fixup-stacktrace
+  "This function removes stack trace elements that came from this namespace. When we throw validation errors, they
+  shouldn't originate from *this* namespace, they should appear to be thrown from the instrumented function itself."
+  [^Exception e]
+  (let [trace (.getStackTrace e)
+        fixed-trace (into-array StackTraceElement
+                                (drop-while
+                                 #(str/starts-with? (.getClassName %)
+                                                    ;; this is... hacky, but it works.
+                                                    (namespace ::x))
```
escherize, line 158: You called out that it's hacky (it's not so bad is it? :) ).

Anyway what do you think of `(str *ns*)`?

johnswanson: That was actually my initial impulse! But sadly, it doesn't seem like `*ns*` is bound to "the namespace I'm currently in," exactly. For an example, I defined a function `test-it` in a random namespace, `metabase.server.middleware.session`:

```
(defn test-it []
  (str *ns*))
```

Then called it from the `user` namespace:

```
(metabase.server.middleware.session/test-it)
```

The result was `user`, not `metabase.server.middleware.session`. 😢

src/metabase/util/malli/fn.clj
```diff
@@ -143,20 +146,36 @@
...
   true)
 
+(defn- fixup-stacktrace
+  "This function removes stack trace elements that came from this namespace. When we throw validation errors, they
+  shouldn't originate from *this* namespace, they should appear to be thrown from the instrumented function itself."
+  [^Exception e]
+  (let [trace (.getStackTrace e)
+        fixed-trace (into-array StackTraceElement
+                                (drop-while
+                                 #(str/starts-with? (.getClassName %)
```
escherize, line 156: I think the `%` needs a type hint.

escherize (approved):
I profiled this and it doesn't negatively impact performance. 

It also prints out a nice looking stack trace. 

Calling a mu/defn form with invalid data now gives:

```
clojure.lang.ExceptionInfo: Invalid input: ["should be an integer"]
{:type :metabase.util.malli.fn/invalid-input, :error {:schema :int, :value "3", :errors ({:path [], :in [], :schema :int, :value "3"})}, :humanized ["should be an integer"], :schema :int, :value "3", :fn-name metabase.task/foo}
 at metabase.task$foo.invokeStatic (NO_SOURCE_FILE:315)
    metabase.task$foo.invoke (NO_SOURCE_FILE:315)
    metabase.task$eval148891.invokeStatic (NO_SOURCE_FILE:319)
    metabase.task$eval148891.invoke (NO_SOURCE_FILE:319)
    clojure.lang.Compiler.eval (Compiler.java:7194)
    clojure.lang.Compiler.eval (Compiler.java:7149)
    clojure.core$eval.invokeStatic (core.clj:3215)
    clojure.core$eval.invoke (core.clj:3211)
```

Thank you @johnswanson

camsaul:
@snoe found a bug with this

```
(mu/defn myfn
  [one two]
  (mbql.u/match-one [:= 9 10]
    [:= a b] [a b]))

(myfn 1 2) ; nil
(myfn 9 10) ; [9 10]
```

See https://metaboat.slack.com/archives/CKZEMT1MJ/p1700257309425109

Issue #34100, opened 2023-09-25
Add settings related events


luizarakaki:
This event should track all changes in the settings table.

Topic: `setting-update`
Mode: `Setting`
details: `key, previous_value, new_value`

If the change was not initiated by a user, we should track this somehow (if changed by env var, for instance)

PR #34720, opened 2023-10-17
Automatically audit setting updates

noahmoss:
Updates the settings framework to add a new `audit` keyword option which dictates the type of automatic audit logging we do for updates to the setting's value. Options are `:none`, `:no-value`, `:raw-value` and `getter`. 

Pursuant to https://github.com/metabase/metabase/issues/34100 but there will be a follow-up PR which adds this to option many existing settings. 

See https://metaboat.slack.com/archives/C056QEW4KNF/p1697464465009129 for relevant discussion.

Also includes some refactor of `model-details` for cards to untangle a circular dependency (we don't need to pre-process a card's query to get the DB and table ID since they're on the card itself already).

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -82,6 +82,7 @@
    [medley.core :as m]
    [metabase.api.common :as api]
    [metabase.config :as config]
+   [metabase.models.audit-log :as audit-log]
    [metabase.models.interface :as mi]
    [metabase.models.serialization :as serdes]
    [metabase.models.setting.cache :as setting.cache]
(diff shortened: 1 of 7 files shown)


test/metabase/models/setting_test.clj
```diff
@@ -1091,3 +1091,57 @@
...
+      (test-setting-1! "DON'T INCLUDE THIS VALUE")
+      (is (= {:topic   :setting-update
+              :user_id  nil
+              :model   "Setting"
+              :details {:key "test-setting-1"}}
+             (last-audit-event-fn))))
+
+    (testing "Auditing can be disabled with `:audit :never`"
+      (test-setting-audit-never! "DON'T AUDIT")
+      (is (not= "DON'T AUDIT"
```
escherize, line 1124: Wouldn't this be true for a setting that was audited by default (`:audit :no-value` vs `:audit :never`)? It'd have the `:details` key with `{:key "test-setting-audit-never!"}` value, I think.

Said another way: Would it make more sense to check that this is `not= "test-setting-audit-never!"`?

src/metabase/models/setting.clj
```diff
@@ -804,6 +815,36 @@
...
+            {:previous-value previous-value
+             :new-value      new-value}))
+   api/*current-user-id*
+   :model/Setting))
+
+(defn- should-audit?
+  "Returns true if the setting change should be written to the `audit_log`."
+  [setting]
+  (and (not= (:audit setting) :never)
+       (site-wide-only? setting)))
```
escherize, line 833: Why is site-wide-only? required here?

escherize (approved):
I had some questions, and I think a test is checking for the wrong thing, but it looks good to me.

test/metabase/models/setting_test.clj
noahmoss: Yep, good catch thanks

Issue #34724, opened 2023-10-17
[INTERNAL]: Make CI workflows 3rd-party safe and friendly


nemanjaglumac:
```[tasklist]
### Workflows
- [ ] Chromatic
- [ ] E2E
- [ ] Drivers
- [ ] ...TBD
```

nemanjaglumac:
Abandoned!
This PR will revert everything: https://github.com/metabase/metabase/pull/35993

PR #34725, opened 2023-10-17
[CI] Make `Chromatic` workflow 3rd party friendly

nemanjaglumac:
This is the first step in accomplishing #34724.
It's also the safest one to start exploring the approach for safe 3rd party (external) contributions running in our CI with access to secrets.
Read about it here: https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets

tl;dr
> The authorize job checks if the workflow was triggered from a fork pull request. In that case, the external environment will prevent the job from running until it’s approved. Otherwise (i.e. when pull requests belong to the main repository), the job will run without requiring explicit approval.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -34,7 +34,7 @@ jobs:
         uses: dorny/paths-filter@v2.11.1
         id: changes
         with:
-          token: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
           filters: .github/file-paths.yaml
 
   fe-chromatic:
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - "master"
-  pull_request:
+  pull_request_target:
     types: [synchronize, labeled, unlabeled]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -4,33 +4,42 @@ on:
...
 
 jobs:
+  authorize:
+    environment:
+      ${{ github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository &&
+      'external' || 'internal' }}
+    runs-on: ubuntu-22.04
+    steps:
+      - run: echo "true"
```
WiNloSt, line 22: What's the difference between just using `run: true` as in [the blog post](https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets)

.github/workflows/chromatic.yml
```diff
@@ -4,33 +4,42 @@ on:
...
       frontend_all: ${{ steps.changes.outputs.frontend_all }}
-      e2e_specs: ${{ steps.changes.outputs.e2e_specs }}
     steps:
       - uses: actions/checkout@v3
       - name: Test which files changed
         uses: dorny/paths-filter@v2.11.1
         id: changes
         with:
-          token: ${{ github.token }}
+          token: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
WiNloSt, line 37: I don't have a way to verify that this secret works atm, but I think you already confirmed it's correct.

WiNloSt (approved):
Looks simple and according to the blog post/our internal doc

.github/workflows/chromatic.yml
nemanjaglumac: Linter was complaining that it doesn't comform to the GitHub schema.

Even the authors of that article use something similar in their own repo.

PR #34752, opened 2023-10-17
Support interval subtraction in expressions

### LINKED ISSUE #23423: Support subtracting intervals
Currently, temporal intervals (`[:interval <amount> <unit>]`) can only be added but not subtracted.

Support for Mongo was added in #23292 but `metabase.driver.sql.query-orocessor` and potentially other namespaces should be extended too. At least `metabase.driver.sql.query-orocessor/->honeysql` should be extended for ` [:sql :-]` similarly to ` [:sql :+]`.

**Links**
- [PR](https://github.com/metabase/metabase/pull/23292) where this issue was discovered with further discussion

lbrdnk:
Closes #23423

### Description

This PR adds option to subtract intervals from datetime in an expression. Minuend must be a datetime while all subtrahends must be intervals.

### How to verify

Run modified tests.

### Checklist


src/metabase/driver/sql/query_processor.clj
```diff
@@ -754,7 +754,21 @@
       (throw (ex-info "Summing intervals is not supported" {:args args})))
     (apply hx/call :+ (map (partial ->honeysql driver) args))))
 
-(defmethod ->honeysql [:sql :-] [driver [_ & args]] (apply hx/call :- (map (partial ->honeysql driver) args)))
+(defmethod ->honeysql [:sql :-]
+  [driver [_ & [first-arg & other-args :as args]]]
+  (assert (not (interval? first-arg))
```
lbrdnk, line 759: Is it ok to use "just assert"? Or should I prefer `(throw (ex-info (tru "..." {:type ...}))`?

src/metabase/driver/sql/query_processor.clj
```diff
@@ -754,7 +754,21 @@
...
+          "Interval as first argrument to subtraction is not allowed.")
+  (if (interval? (first other-args))
+    (do
+      (assert (every? interval? other-args)
```
lbrdnk, line 763: I'm a fan of "rather safe than sorry" attitude, but I have a feeling this assertion could unnecessary. Thanks for your opinion in advance :).

src/metabase/driver/sql/query_processor.clj
bshepherdson: Broadly speaking, it's fine to use `assert`.

More specifically, since `assert` throws `AssertionError` (which is an `Error` not an `Exception`, and so isn't caught by `(catch Exception e ...)`) it should not be used for an error a user might have caused, only an error caused by bad coding.

Put differently, `assert` is for 500 errors, `(when ... (throw (ex-info ...)))` is for 4xx errors.

src/metabase/driver/sql/query_processor.clj
```diff
@@ -754,7 +754,21 @@
...
+      (reduce (fn [hsql-form [_ amount unit]]
+                ;; We are adding negative amount. Inspired by `->honeysql [:sql :datetime-subtract]`.
+                (add-interval-honeysql-form driver hsql-form (- amount) unit))
+              (->honeysql driver first-arg)
+              other-args))
+    (apply hx/call :- (map (partial ->honeysql driver) args))))
+
 (defmethod ->honeysql [:sql :*] [driver [_ & args]] (apply hx/call :* (map (partial ->honeysql driver) args)))
 
 ;; for division we want to go ahead and convert any integer args to floats, because something like field / 2 will do
```
bshepherdson, line 760: <i class='graphite__hidden'>[Re: lines 929 to 929]<br/></i>

Here's an example of checking for valid args to functions. I'd use this style (`when` + `(throw (ex-info ...))`) for checking these preconditions. <span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/metabase/metabase/34752?utm_source=unchanged-line-comment">Graphite</a>.</span>

src/metabase/driver/sql/query_processor.clj
bshepherdson: See below. It's good to check these conditions, but they should be `Exception`s rather than `AssertionError`s.

bshepherdson (changes requested):
LGTM aside from the assertions.

lbrdnk:
@bshepherdson, thank you for clarifying Assertions x Expressions relationship. I've modified the code accordingly, please check it out.

bshepherdson (approved):

PR #34757, opened 2023-10-17
Add new audit views as migrations

noahmoss:
Adds the [views for audit v2 content](https://github.com/metabase/metabase/tree/auditv2-main/resources/instance_analytics_views) to the app DB using SQL migrations


luizarakaki (approved):
lgtm

PR #34681, opened 2023-10-15
Enable column-filter, disable distribution and summarize-column drills for aggregated columns

metamben:
Fixes #34223
Fixes #34341

PR #34758, opened 2023-10-17
Remove invalid test cases

metamben:
The breaking test cases were ported from JS to CLJC while PR #34681 that removed the JS test cases was waiting for review. The test cases were removed because they contradict earlier (temporarily commented out) test cases.


lbrdnk (approved):
LGTM

Issue #33811, opened 2023-09-07
[MLv2] [BE] `drill-thru/zoom-in.timeseries` doesn't show for ORDERS grouped by CREATED_AT


deniskaber:
Timeseries-zoom drill was added in https://github.com/metabase/metabase/pull/33574 . 
But I haven't managed to trigger for a simple orders query with breakout by created_at - see these unit tests - https://github.com/metabase/metabase/blob/33973-mbql2-drills-quick-filters/frontend/src/metabase-lib/drills.unit.spec.ts#L1139 (and other parts marked with `metabase#33811`)

https://github.com/metabase/metabase/assets/7983744/86eed922-ef6e-41f6-9b15-2380c4acb44c

camsaul:
I think this might actually just be an FE bug, see discussion in https://metaboat.slack.com/archives/C04CYTEL9N2/p1695929177537919

camsaul:
reopening based on discussion in https://metaboat.slack.com/archives/C04CYTEL9N2/p1695990363844759?thread_ts=1695929177.537919&cid=C04CYTEL9N2

camsaul:
ok so this is actually fixed after all, but the tests were still commented out because of a bug in the tests (they should have been `:cell` rather than `:header` test cases). I'll update in a separate PR

PR #34760, opened 2023-10-17
5+ drill thru fixes

### LINKED ISSUE #34383: [MLv2] [BE] `fk-details` generates query to the wrong table
`fk-details` drill apply using drillThru api generates query to the same table as it was triggered, however it should be fk target table.

See these unit tests to reproduce - https://github.com/metabase/metabase/blob/5ab317540d064ac3d402508d0038629e8a041575/frontend/src/metabase-lib/drills.unit.spec.ts#L1900

### LINKED ISSUE #34439: [MLv2] [BE] `underlying-records` drill gets shown two times for aggregated query
`drill-thru/underlying-records` shows two times for aggregated query metric value

Check this test case -
https://github.com/metabase/metabase/blob/5ab317540d064ac3d402508d0038629e8a041575/frontend/src/metabase-lib/drills.unit.spec.ts#L462

### LINKED ISSUE #34440: [MLv2] [BE] `fk-filter` gets returned for non-fk column
`fk-filter` gets returned for non-fk column

Check this reproduction test - 
https://github.com/metabase/metabase/blob/5ab317540d064ac3d402508d0038629e8a041575/frontend/src/metabase-lib/drills.unit.spec.ts#L462

ALSO, the other way around - `fk-filter` doesn't get returned for fk column that was used as breakout - 
https://github.com/metabase/metabase/blob/7d9d4f42af6cf3ef31514b23184a3a79d449c3d2/frontend/src/metabase-lib/drills.unit.spec.ts#L977

### LINKED ISSUE #34441: [MLv2] [BE] `fk-details` gets returned for non-fk colum
`fk-details` gets returned for non-fk column

Check this reproduction test - 
https://github.com/metabase/metabase/blob/5ab317540d064ac3d402508d0038629e8a041575/frontend/src/metabase-lib/drills.unit.spec.ts#L462

### LINKED ISSUE #34499: [MLv2] [BE] Applying `sort` to a query with custom column produces a broken query
Applying `sort` to a query with custom column produces a broken query due to extra binning.

Issue happens for a custom column that does not have real underlying table field.

You can use this reproduction unit test - 
https://github.com/metabase/metabase/pull/34460/files#diff-b15cc6d446a355dead5003f1c3b596f577074d40cc275ce8df9243c98ad3552dR2207-R2218

camsaul:
### #34439, #34440, #34441

These drills were accidentally calculating drills based on the `:dimensions`  but should have been ignoring them. Added tests.

Fixes #34439
Fixes #34440 
Fixes #34441

### #34383

The code for fk details drills was doing something goofy, not 100% sure I understand what, but FK details is just supposed to return a single related record like this:

image

image

so I fixed the implementation while I was at it.

Fixes #34383

### #33811

I also uncommented out the tests for #33811 which was fixed a few weeks ago, but there was a bug in the tests (it was testing with `header` clicks when the screenshot showed `cell` clicks) so I fixed the tests and enabled them.

### #34499

#34499 seems to already be fixed, I enabled the test and it is now passing, but I added a backend version of the test as well.

Fixes #34499

### DIFF
--- a/frontend/src/metabase-lib/drills.unit.spec.ts
+++ b/frontend/src/metabase-lib/drills.unit.spec.ts
@@ -982,16 +982,16 @@ describe("availableDrillThrus", () => {
         type: "drill-thru/fk-filter",
       },
     },
-    // FIXME: `fk-filter` doesn't get returned for fk column that was used as breakout (metabase#34440)
-    // {
-    //   drillType: "drill-thru/fk-filter",
-    //   clickType: "cell",
-    //   queryType: "aggregated",
-    //   columnName: "PRODUCT_ID",
-    //   expectedParameters: {
-    //     type: "drill-thru/fk-filter",
-    //   },
-    // },
+    // `fk-filter` doesn't get returned for fk column that was used as breakout (metabase#34440)
+    {
+      drillType: "drill-thru/fk-filter",
+      clickType: "cell",
+      queryType: "aggregated",
+      columnName: "PRODUCT_ID",
+      expectedParameters: {
+        type: "drill-thru/fk-filter",
+      },
+    },
     // endregion
 
     // region --- drill-thru/quick-filter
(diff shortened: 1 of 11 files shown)


bshepherdson (approved):
LGTM aside from the discussion we need to have about the `dimension-contexts` hack. I think there's a much better way to tackle this. It can be addressed after this PR lands, though.

PR #34765, opened 2023-10-17
Mlv2 describe relative date time range

### LINKED ISSUE #34453: [MLv2] BE - Add a function to format relative datetime range
For relative date filters we calculate the corresponding date range on the FE and show it in the popover. Can we expose a MLv2 function that does that? Any inconsistency here with the actual BE behavior would be always a bug.

Screenshot 2023-10-06 at 23 19 43

Possible API:
```
// current day
Lib.describeRelativeDatetimeRange(
  "current",
  "day"
);

// next 30 days
Lib.describeRelativeDatetimeRange(
  30,
  "day"
);

// past 30 days, starting 10 days ago
Lib.describeRelativeDatetimeRange(
  -30,
  "day",
  -10,
  "quarter"
);

// past 30 days, including current day
Lib.describeRelativeDatetimeRange(
  -30,
  "day",
  null,
  null,
  { "include-current": true } 
)
```

snoe:
Fixes #34453

exports the function `format-relative-date-range`


ranquild (approved):

shared/src/metabase/shared/util/internal/time.clj
```diff
@@ -66,39 +66,57 @@
...
 
-(defmethod common/to-range :year [value _]
-  (let [value (t/truncate-to value :days)]
-    [(t/adjust value :first-day-of-year)
-     (minus-ms (t/adjust value :first-day-of-next-year))]))
+(defmethod common/to-range :year [value {:keys [n] :or {n 1}}]
+  (let [value (-> value
+                  (t/truncate-to :days)
+                  (t/adjust :first-day-of-year))]
+    [value (minus-ms (nth (iterate #(t/adjust % :first-day-of-next-year n) value) n))]))
```
metamben, line 119: I would expect `[value (minus-ms (t/plus value (t/years n)))]` to achieve the same.

PR #34769, opened 2023-10-17
Add filter-positions to filterable-columns

### LINKED ISSUE #34736: [MLv2] BE - Add `filterPosition` to column display info
When clicking an existing filter in the notebook editor, the FE should highlight a column the filter is using (#34737)

We could follow the same approach we have for order-bys and breakouts, and make the `filterable-columns` append `filterPosition` properties to columns used by filters

metamben:
Fixes #34736.

### DIFF
--- a/src/metabase/lib/filter.cljc
+++ b/src/metabase/lib/filter.cljc
@@ -373,17 +373,11 @@
        (vec columns)
 
        :else
-       (let [matching (transduce
-                       (keep-indexed (fn [index a-filter]
-                                       (when-let [a-ref (leading-ref a-filter)]
-                                         (when-let [col (lib.equality/find-matching-column
-                                                         query stage-number a-ref columns)]
-                                           [col index]))))
-                       (completing
-                        (fn [m [col index]]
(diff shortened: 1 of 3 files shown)


src/metabase/lib/filter.cljc
```diff
@@ -351,12 +359,35 @@
...
+                                                         query stage-number a-ref columns)]
+                                           [col index]))))
+                       (completing
+                        (fn [m [col index]]
+                          (update m col (fnil conj []) index)))
+                       {}
+                       existing-filters)]
+         (mapv #(let [positions (matching %)]
+                  (cond-> %
+                    positions (assoc :filter-positions positions)))
```
bshepherdson, line 389: I find this very hard to read. How about this?
```suggestion
       (->> (range (count existing-filters))                                                                                  
            (group-by #(when-let [a-ref (leading-ref (nth existing-filters %))]                                               
                         (lib.equality/find-matching-column query stage-number a-ref columns))))  ```

bshepherdson (approved):

PR #34770, opened 2023-10-17
Prevent table for pivoting if all columns have high cardinality

### LINKED ISSUE #34677: Auto pivot is freezing query editor on high cardinality fields
### Describe the bug

In the query editor, when you group by 2 fields, it auto pivots the 2nd field as columns, which is neat (though not sure if it is really useful way of looking at the data most of the time when exploring). However, if both fields have high cardinality, the entire page freezes. I was trying to group by Customer ID and Name in our Account model to see which customer have multiple accounts (sorting counts in descending), but got a frozen page instead as it tries to render the pivoted table view. 

For my use case, pivot isn't useful at all, so if it did render, then I would need to go to settings to unpivot to see the highest count in order. 

### To Reproduce

1. Group by 2 fields with high cardinality (at least a few thousand each)
2. Query editor spins / UI is frozen until pivot view is rendered (or never if cardinality is too high). 

### Expected behavior

Auto pivot should only be done if cardinality is low (maybe under 100 or even lower?). Or not at all / only provide a hint/option to easily pivot

### Logs

n/a

```JSON
master on Postgres
```

### Severity

Medium as the magic is causing more problem than it provides

### Additional context

_No response_

alxnddr:
Closes https://github.com/metabase/metabase/issues/34677

### Description

Do not auto pivot the Table visualization on dimensions with cardinality > 16. This is about the pivot functionality of our regular table, not pivot table.

### How to verify

1. New question -> Sample Dataset -> Orders
2. Summarize by [Discount: Auto binned] and [Total: Auto binned]
3. Ensure it shows the table which was automatically pivoted
4. New question -> Sample Dataset -> Orders 
5. Summarize by [Id] and [Product Id]
6. Ensure it shows the table which is not automatically pivoted

### Checklist

### DIFF
--- a/frontend/src/metabase/visualizations/lib/utils.js
+++ b/frontend/src/metabase/visualizations/lib/utils.js
@@ -3,6 +3,7 @@ import d3 from "d3";
 import { t } from "ttag";
 import crossfilter from "crossfilter";
 
+import { isNotNull } from "metabase/core/utils/types";
 import { isDimension, isMetric, isDate } from "metabase-lib/types/utils/isa";
 
 export const MAX_SERIES = 100;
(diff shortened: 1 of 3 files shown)


iethree (approved):
👏  Very nice. I saw a 26% performance improvement with a cardinality of 155. I imagine it's even larger for higher cardinality.

Issue #33281, opened 2023-08-17
[Epic] Migrate Pie Chart to ECharts


EmmadUsmani:
**Links**
[Hub page: Viz revamp research and planning](https://www.notion.so/metabase/Viz-Revamp-Research-and-Planning-4198b2726ac64f1d89ad1de6187532d2#3e3dda5bce58473397e67bb44a741a65)
[Product doc: Implement the Pie chart using ECharts](https://www.notion.so/metabase/Implement-the-Pie-chart-using-ECharts-7e18ef8eac32425bada8a3ecd360158d)

## Static Chart
- [x] #43556
- [x] https://github.com/metabase/metabase/issues/43564

#### Formatting
- [x] https://github.com/metabase/metabase/issues/43762
- [x] https://github.com/metabase/metabase/issues/43773
- [x] https://github.com/metabase/metabase/issues/43774
- [x] https://github.com/metabase/metabase/issues/44086

#### Negative Metric Values
- [x] https://github.com/metabase/metabase/issues/43769
- [x] https://github.com/metabase/metabase/issues/43770

#### Other
- [x] https://github.com/metabase/metabase/issues/43764
- [x] https://github.com/metabase/metabase/issues/43766
- [x] https://github.com/metabase/metabase/issues/43557
- [x] https://github.com/metabase/metabase/issues/43797
- [x] https://github.com/metabase/metabase/issues/44019
- [x] https://github.com/metabase/metabase/issues/44085
- [x] https://github.com/metabase/metabase/issues/44087
- [x] https://github.com/metabase/metabase/issues/44847
- [x] https://github.com/metabase/metabase/issues/45149

## Dynamic Chart
- [x] https://github.com/metabase/metabase/issues/43961
- [x] https://github.com/metabase/metabase/issues/43558
- [x] https://github.com/metabase/metabase/issues/43559
- [x] https://github.com/metabase/metabase/issues/43974
- [x] https://github.com/metabase/metabase/issues/43560
- [x] https://github.com/metabase/metabase/issues/43561
- [x] https://github.com/metabase/metabase/issues/43562
- [x] https://github.com/metabase/metabase/issues/43565
- [x] https://github.com/metabase/metabase/issues/45687

### Dynamic Regressions
- [x] https://github.com/metabase/metabase/issues/43768
- [x] https://github.com/metabase/metabase/issues/44988
- [x] https://github.com/metabase/metabase/issues/45686

## Existing Bugs

[...]

PR #34717, opened 2023-10-16
setup rendering pipeline for echarts static pie

EmmadUsmani:
Part of https://github.com/metabase/metabase/issues/33281

### Description

Sets up the static viz rendering pipeline for our new static pie chart implementation using echarts. Currently it uses some mock values that will be replaced by the actual implementation in later PRs in this stack. It also currently adds a dynamic pie chart visualization type to the client, but this will also be removed later in the stack before it is all merged to master.

### How to verify

1. Create a question
2. Set visualization type to `Pie 2`
3. Add to a dashboard
4. Send in an email subscription
5. Confirm that a placeholder chart is rendered without any errors

### Demo

^ output from subscription email

[Screen Recording 2023-10-16 at 2.00.12 PM.mov](https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/9ULpbAzGBtvQRyZEpGvu/92243e9e-4481-4edd-a6b6-3abb6530db63/Screen%20Recording%202023-10-16%20at%202.00.12%20PM.mov)

### Checklist

- [ ] Tests have been added/updated to cover changes in this PR

_TODO: Need to update `js_svg_test.clj`_

PR #34772, opened 2023-10-18
compute pie chart settings

EmmadUsmani:
Part of https://github.com/metabase/metabase/issues/33281

### Description

Computes the viz settings object, which will we later need to compute the chart model and echarts' option object.

### How to verify

1. Create a question
2. Set visualization type to `Pie 2`
3. Add to a dashboard
4. Send in an email subscription
5. Check backend logs, should show reasonable values for the computed settings object.

### Demo

```json
{
  "pie.show_total": false,
  "pie.percent_visibility": "inside",
  "graph.dimensions": [
    "CATEGORY"
  ],
  "graph.metrics": [
    "count"
  ],
  "pie.dimension": "CATEGORY",
  "pie.metric": "count",
  "pie.show_legend": true,
  "pie.slice_threshold": 2.5,
  "pie.colors": {
    "Doohickey": "#88BF4D",
    "Gadget": "#F9D45C",
    "Gizmo": "#A989C5",
    "Widget": "#F2A86F"
  }
}
```

Example output

### Checklist

### DIFF
--- a/frontend/src/metabase/static-viz/components/PieChart/PieChart.tsx
+++ b/frontend/src/metabase/static-viz/components/PieChart/PieChart.tsx
@@ -2,16 +2,26 @@ import { init } from "echarts";
 import type { IsomorphicStaticChartProps } from "metabase/static-viz/containers/IsomorphicStaticChart/types";
 import { sanitizeSvgForBatik } from "metabase/static-viz/lib/svg";
 
+import { computeStaticPieChartSettings } from "./setttings";
+
 const WIDTH = 540;
 const HEIGHT = 360;
 
-export function PieChart(props: IsomorphicStaticChartProps) {
+export function PieChart({
+  rawSeries,
+  renderingContext,
+}: IsomorphicStaticChartProps) {
+  const computedVizSettings = computeStaticPieChartSettings(rawSeries);
+  //eslint-disable-next-line no-console
+  console.log("computedVizSettings", JSON.stringify(computedVizSettings));
+
   const chart = init(null, null, {
     renderer: "svg",
     ssr: true,
     width: WIDTH,
     height: HEIGHT,
   });
+
   chart.setOption({
     // Mock data, will be replaced
     series: {
@@ -22,6 +32,7 @@ export function PieChart(props: IsomorphicStaticChartProps) {
       ],
     },
   });
+
   const svg = sanitizeSvgForBatik(chart.renderToSVGString());
 
   return (
--- a/frontend/src/metabase/static-viz/components/PieChart/PieChart.tsx
+++ b/frontend/src/metabase/static-viz/components/PieChart/PieChart.tsx
@@ -9,9 +9,13 @@ const HEIGHT = 360;
 
 export function PieChart({
   rawSeries,
+  dashcardSettings,
   renderingContext,
 }: IsomorphicStaticChartProps) {
-  const computedVizSettings = computeStaticPieChartSettings(rawSeries);
+  const computedVizSettings = computeStaticPieChartSettings(
+    rawSeries,
+    dashcardSettings,
+  );
   //eslint-disable-next-line no-console
   console.log("computedVizSettings", JSON.stringify(computedVizSettings));
 
--- a/frontend/src/metabase/static-viz/components/PieChart/setttings.ts
+++ b/frontend/src/metabase/static-viz/components/PieChart/setttings.ts
@@ -1,33 +1,9 @@
-import type {
-  ComputedVisualizationSettings,
-  RemappingHydratedDatasetColumn,
-} from "metabase/visualizations/types";
-import type {
-  DatasetColumn,
-  PieChartSettings,
-  RawSeries,
-  VisualizationSettings,
-} from "metabase-types/api";
+import type { ComputedVisualizationSettings } from "metabase/visualizations/types";
+import type { PieChartSettings, RawSeries } from "metabase-types/api";
 import { getDefaultDimensionAndMetric } from "metabase/visualizations/lib/utils";
 import { getColorsForValues } from "metabase/lib/colors/charts";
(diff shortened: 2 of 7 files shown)


EmmadUsmani:
* master
  * **PR #34606** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34606" target="_blank">Graphite</a> 
    * **PR #34717** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34717" target="_blank">Graphite</a> 
      * **PR #34772** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34772" target="_blank">Graphite</a>  👈
    * **PR #34703** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34703" target="_blank">Graphite</a> 
      * **PR #34763** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34763" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/34772?utm_source=stack-comment).

frontend/src/metabase-types/api/dataset.ts
```diff
@@ -94,7 +94,7 @@ export interface NativeQueryForm {
 
 export type SingleSeries = {
   card: Card;
-} & Dataset;
+} & Pick<Dataset, "data" | "error">;
```
EmmadUsmani, line 97: At runtime this value seems to only ever be `{ card, data }`



In the codebase there was only a single reference to any other property, `error`



Removing the extra types saves us from needing to specify them in mock data for tests.

frontend/src/metabase/static-viz/components/PieChart/PieChart.tsx
```diff
@@ -2,16 +2,26 @@ import { init } from "echarts";
...
 const WIDTH = 540;
 const HEIGHT = 360;
 
-export function PieChart(props: IsomorphicChartProps) {
+export function PieChart({
+  rawSeries,
+  renderingContext,
+}: IsomorphicChartProps) {
+  const computedVizSettings = computeStaticPieChartSettings(rawSeries);
+  //eslint-disable-next-line no-console
```
EmmadUsmani, line 15: Will remove in next PR when we consume the settings object

frontend/src/metabase/static-viz/components/PieChart/setttings.ts
```diff
@@ -0,0 +1,74 @@
...
+  VisualizationSettings,
+} from "metabase-types/api";
+import { getDefaultDimensionAndMetric } from "metabase/visualizations/lib/utils";
+import { getColorsForValues } from "metabase/lib/colors/charts";
+import { normalize } from "metabase-lib/queries/utils/normalize";
+import { getColumnKey } from "metabase-lib/queries/utils/get-column-key";
+
+import { SLICE_THRESHOLD } from "./constants";
+
+// TODO move this to base branch in shared location
```
EmmadUsmani, line 18: @alxnddr This function was in your old pie chart branch, and it seems like you added it to the combo chart branch. Can you move it down to the base echarts branch so we can share it?

frontend/src/metabase/static-viz/components/PieChart/setttings.ts
```diff
@@ -0,0 +1,74 @@
...
+    settings["pie.metric"] ??= defaults.metric;
+  }
+
+  settings["pie.show_legend"] ??= true;
+  settings["pie.show_total"] ??= true;
+  settings["pie.percent_visibility"] ??= "legend";
+  settings["pie.slice_threshold"] ??= SLICE_THRESHOLD * 100;
+  settings["pie.colors"] ??= getPieChartColors(rawSeries, settings);
+
+  settings.column = (column: RemappingHydratedDatasetColumn) =>
```
EmmadUsmani, line 70: @alxnddr We have to do the same thing for the combo and probably other charts as well. How about we extract this into a shared constant in the base branch, e.g.

```typescript
const defaultColumnSettingsFunc = (column: RemappingHydratedDatasetColumn) => getColumnSettings(column, settings)
```

That way we don't have to manually do this and import two extra things here.

frontend/src/metabase-types/api/dataset.ts
alxnddr: Good! I pushed this change to the base branch because I also need this

frontend/src/metabase/static-viz/components/PieChart/setttings.ts
alxnddr: I updated the base branch with `getCommonStaticVizSettings`

frontend/src/metabase/static-viz/components/PieChart/setttings.ts
```diff
@@ -0,0 +1,74 @@
...
+    ...rawSeries[0].card.visualization_settings,
+  };
+
+  if (!settings["pie.dimension"] || !settings["pie.metric"]) {
+    const defaults = getDefaultDimensionAndMetric(rawSeries);
+    settings["pie.dimension"] ??= defaults.dimension;
+    settings["pie.metric"] ??= defaults.metric;
+  }
+
+  settings["pie.show_legend"] ??= true;
```
alxnddr, line 64: Wdyt about extracting the default value computation into separate functions that are reused both in the settings definitions and here? I tried to follow this approach here https://github.com/metabase/metabase/pull/34763/files#diff-8b242d67ce06af681f4568ad6bfbe2b55d76dc11838c0022c68bec3c05f368cdR32 
It may look goofy for such simple cases but I believe it adds resilience and eliminates the possibility of updating the settings definition only while forgetting about static charts

alxnddr: @EmmadUsmani what is your opinion on this?

PR #34785, opened 2023-10-18
Translations: Removed duplicated translation keys "Conditional Formatting" vs "Conditional formatting"

antonhedstrom:
While updating a couple of Swedish translations in POEditor I discovered a "duplication" of translation key. It is not technically a duplicate (`Conditional Formatting` vs `Conditional formatting`) but one of them should be enough.

### Description

Removed one of them (`Conditional Formatting`) in code so it can later be removed in the translations (via POEditor) as well.

### How to verify

1. Related tests: 
    *`e2e/test/scenarios/question/settings.cy.spec.js`
    *`e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js` 

### Checklist

### DIFF
--- a/docs/questions/sharing/visualizations/table.md
+++ b/docs/questions/sharing/visualizations/table.md
@@ -205,7 +205,7 @@ Lets you toggle between showing the currency label in the column heading or in e
 
 ## Conditional table formatting
 
-Sometimes it's helpful to highlight certain rows or columns in your tables when they meet a specific condition. You can set up conditional formatting rules by going to the visualization settings while looking at any table, then clicking on the **Conditional Formatting** tab.
+Sometimes it's helpful to highlight certain rows or columns in your tables when they meet a specific condition. You can set up conditional formatting rules by going to the visualization settings while looking at any table, then clicking on the **Conditional formatting** tab.
 
 ![Conditional formatting](../../images/conditional-formatting.png)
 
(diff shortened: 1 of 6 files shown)


ranquild:
FE changes are good, but want to get a confirmation from design

ranquild (commented):
Waiting for design review

mazameli (approved):
Thanks! This better matches our brand guidelines, where we say we try to stick with `Sentence case` rather than `Title Case` in general.

antonhedstrom:
Thanks for feedback!
Right now I am confused regarding the failing workflow though. My understanding is that the PR needs either `backport`or `no-backport` label. It seems I can't modify the labels (which make sense). Is this something you will take decision on or do I need to take action somehow?
Sorry, haven't contributed here before :) Tried to find documentation about that workflow but was unable to do so.

darksciencebase:
@antonhedstrom Thank you for the contribution. Don't worry about the backport/no-backport thing, it's for internal use.

antonhedstrom:
Sorry for late response, didn't notice there was a conflict in this. Tried rebasing master but run into a situation where I had to edit the todo list (`git rebase --edit-todo`) for hundreds of commits. Since my change was so small I "reimplemented" it on master manually instead.
Can you take another look @mazameli ? 👼

PR #34792, opened 2023-10-18
Rename dashboard's ordered_cards to dashcards

calherries:
This PR renames the `ordered_cards` key on Dashboard entities to `dashcards`. The key is hydrated onto dashboards so there is no need for a migration.

In each commit in this PR I'm doing a simple automated search-and-replace. Except for
[Rename orderedCards to dashcards (manual)](https://github.com/metabase/metabase/pull/34792/commits/db7b0dc7729b8e3e2f1e64a511fc273c287aaaa8), which was manual. I assume that `ordered_cards`, `orderedCards`, `orderedCard`, and `ordered-cards` are only used in the codebase to refer to dashboard's "ordered_cards" and not something else.

### DIFF
--- a/frontend/src/metabase-types/api/dashboard.ts
+++ b/frontend/src/metabase-types/api/dashboard.ts
@@ -18,7 +18,7 @@ export interface Dashboard {
   name: string;
   description: string | null;
   model?: string;
-  ordered_cards: (DashboardOrderedCard | ActionDashboardCard)[];
+  dashcards: (DashboardOrderedCard | ActionDashboardCard)[];
   ordered_tabs?: DashboardOrderedTab[];
   parameters?: Parameter[] | null;
   can_write: boolean;
--- a/frontend/src/metabase-types/api/dashboard.ts
+++ b/frontend/src/metabase-types/api/dashboard.ts
@@ -18,7 +18,7 @@ export interface Dashboard {
   name: string;
   description: string | null;
   model?: string;
-  dashcards: (DashboardOrderedCard | ActionDashboardCard)[];
+  dashcards: (DashboardCard | ActionDashboardCard)[];
   ordered_tabs?: DashboardOrderedTab[];
   parameters?: Parameter[] | null;
   can_write: boolean;
@@ -36,7 +36,7 @@ export interface Dashboard {
 
 export type DashCardId = number;
 
-export type BaseDashboardOrderedCard = {
+export type BaseDashboardCard = {
   id: DashCardId;
   dashboard_id: DashboardId;
   dashboard_tab_id?: DashboardTabId;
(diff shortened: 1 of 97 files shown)


frontend/src/metabase/dashboard/selectors/selectors.js
```diff
@@ -120,7 +120,7 @@ export const getDashboardComplete = createSelector(
       return null;
     }
 
-    const ordered_cards = dashboard.ordered_cards
+    const orderedDashcards = dashboard.dashcards
```
markbastian, line 123: I think this should just be `dashcards`.

calherries: We need to distinguish it from `dashcards`, which is the name of the argument.

calherries: It could be `newDashcards` too, but `orderedDashcards` feels more fitting because the most significant part of this transformation is making the dashcards are ordered.

markbastian (approved):
Left one comment on a single name change. Not a dealbreaker, but consider it. Aside from that if the tests pass and you've done a little smoke testing I'd say 🚢 🇮🇹 .

frontend/src/metabase/dashboard/selectors/selectors.js
markbastian: Works for me.

frontend/src/metabase-types/api/dashboard.ts
```diff
@@ -18,7 +18,7 @@ export interface Dashboard {
   name: string;
   description: string | null;
   model?: string;
-  ordered_cards: (DashboardOrderedCard | ActionDashboardCard)[];
+  dashcards: (DashboardOrderedCard | ActionDashboardCard)[];
   ordered_tabs?: DashboardOrderedTab[];
```
kamilmielnik, line 22: Is there a reason for `ordered_tabs` to have `ordered_` prefix?
It looks like it falls into the same category as `ordered_cards`.
Perhaps it's a good candidate for a subsequent PR.

frontend/src/metabase-types/api/dashboard.ts
```diff
@@ -18,7 +18,7 @@ export interface Dashboard {
   name: string;
   description: string | null;
   model?: string;
-  ordered_cards: (DashboardOrderedCard | ActionDashboardCard)[];
+  dashcards: (DashboardOrderedCard | ActionDashboardCard)[];
```
kamilmielnik, line 21: I think we should also drop the word `Ordered` from:
- `DashboardOrderedCard`
- `BaseDashboardOrderedCard`
- `createMockDashboardOrderedCard`

kamilmielnik (commented):
JS/TS changes look good :+1: 

I left some comments about complementary renaming.

frontend/src/metabase-types/api/dashboard.ts
calherries: Yes, thank you!

kamilmielnik (approved):
:+1:

frontend/src/metabase-types/api/dashboard.ts
markbastian: I'm guessing only as a mirror. Aside from the fact that a sort may or may not have occurred somewhere that might give the cards some semblance of order, there is not reason to use `ordered_` or `Ordered` anywhere. The data structure is certainly not a sorted collection.

PR #34796, opened 2023-10-18
Execute PUT requests in parallel on dashboard save

calherries:
This PR speeds up saving dashboards by running the two PUT requests in parallel. 

### The problem

When a dashboard is saved, we start by making two requests to save the dashboard: 

PUT /api/dashboard/:id/cards, which updates report_dashboardcard and dashboard_tab
followed by 
PUT /api/dashboard/:id, which updates report_dashboard.

They are made in series, one after the other. Redux state is even updated in between, making the entire process much longer than it needs to be.

### The solution

Simply execute the PUT requests in parallel, using `Promise.all`.

By making this change, we're implicitly assuming that both endpoints update and read independent data, so there is no race conditions. This assumption is currently true, but could change in the future. Ideally we'd want to use one endpoint to enforce this in the future.


frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -93,34 +93,37 @@ export const updateDashboardAndCards = createThunkAction(
...
-          series: dc.series,
-          visualization_settings: dc.visualization_settings,
-          parameter_mappings: dc.parameter_mappings,
-        })),
-        ordered_tabs: (dashboard.ordered_tabs ?? [])
-          .filter(tab => !tab.isRemoved)
-          .map(({ id, name }) => ({
-            id,
-            name,
+      const updateCardsAndTabs = () =>
```
uladzimirdev, line 97: you can keep it as before, but await in Promise.all:

`const updatedCardsAndTabs = DashboardApi.updateCardsAndTabs({`

calherries: Oh great

frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -93,34 +93,37 @@ export const updateDashboardAndCards = createThunkAction(
...
+          ordered_tabs: (dashboard.ordered_tabs ?? [])
+            .filter(tab => !tab.isRemoved)
+            .map(({ id, name }) => ({
+              id,
+              name,
+            })),
+        });
+
+      // Make two parallel requests: one to update the dashboard and another for the dashcards and tabs
+      const [updatedCardsAndTabs, _updateDashboardAction] = await Promise.all([
```
WiNloSt, line 122: I don't see much point in getting the second variable if we're not going to use it. In most places, we do leave some to make sure people are aware of the returned value e.g. for old JS files that has specific API such as columns, visualization_settings, etc.

example 
https://github.com/metabase/metabase/blob/97f41c76831ef1210a67e762db8a202731cc0f96/enterprise/frontend/src/metabase-enterprise/auth/index.js#L49-L54

This might not be a good example, because the omitted variable is not the last used variable.

frontend/src/metabase/dashboard/actions/save.js
calherries: Updated [Skip the anonymous function](https://github.com/metabase/metabase/pull/34796/commits/c58d8f0607966cf42f3bbe60761e99539f8e257a)

ranquild (commented):
Did you investigate the dashboard reducer to see if it handles responses in any order? If yes, could you post what you found here?

frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -117,10 +116,13 @@ export const updateDashboardAndCards = createThunkAction(
...
-      dispatch(saveCardsAndTabs(updatedCardsAndTabs));
-
-      await dispatch(Dashboards.actions.update(dashboard));
 
+      // Make two parallel requests: one to update the dashboard and another for the dashcards and tabs
+      const [updatedCardsAndTabs] = await Promise.all([
+        updateCardsAndTabs,
+        dispatch(Dashboards.actions.update(dashboard)),
+      ]);
+      dispatch(saveCardsAndTabs(updatedCardsAndTabs));
```
uladzimirdev, line 125: @calherries you changed the order, so now `dispatch(saveCardsAndTabs(updatedCardsAndTabs));` is after `dispatch(Dashboards.actions.update(dashboard));`, but before it was done in parallel

uladzimirdev: ah, actually it should be fine as we update only FE data there, nvm

uladzimirdev (commented):
should be good 🤞

frontend/src/metabase/dashboard/actions/save.js
calherries: Yes you're talking about `saveCardsAndTabs` I assume, which is handled [here](https://github.com/metabase/metabase/blob/6cec76055335c80092f3697c3011c8c8461916c3/frontend/src/metabase/dashboard/actions/tabs.ts#L282-L309). I will explain this context as a comment next time.

calherries:
@ranquild The actions involved update completely different parts of the redux state tree. It doesn't matter which order they are updated in I think. Is that what you wanted to check?

`metabase/entities/dashboards/UPDATE` (triggered by `dispatch(Dashboards.actions.update(dashboard))`)
https://github.com/metabase/metabase/blob/d99fe514ea478a2c89e408e55fae6cb91f9dea4a/frontend/src/metabase/dashboard/reducers.js#L144-L151
Updates entities.dashboards.<id> data, but this doesn't include information about dashcards and tabs.

`metabase/dashboard/SAVE_CARDS_AND_TABS` (triggered by `dispatch(saveCardsAndTabs(updatedCardsAndTabs))`)
Updates `dashboard.selectedTabId` and `dashboard.dashcards` to replace temporary IDs with real IDs.

frontend/src/metabase/dashboard/actions/save.js
WiNloSt: Actually `dispatch(saveCardsAndTabs(updatedCardsAndTabs))` doesn't have to wait for `dispatch(Dashboards.actions.update(dashboard))` to be finished first.

If we want to optimize the waiting time we could do something like this
```ts
updateCardsAndTabs.then((updatedCardsAndTabs) => {
  dispatch(saveCardsAndTabs(updatedCardsAndTabs))
})

await Promise.all([
  updateCardsAndTabs,
  dispatch(Dashboards.actions.update(dashboard)),
])
```

This has duplications, but will perform faster on paper.

WiNloSt (approved):
Looks good to me

frontend/src/metabase/dashboard/actions/save.js
calherries: Good point, though unfortunately it doesn't really make any difference in practice because the `updateCardsAndTabs` promise should take about the same time or longer than the `Dashboards.actions.update` dispatch. Anyway, I've updated it in case this changes.

paoliniluis:
This same pattern should be applied when getting the cards from a dashboard

PR #34802, opened 2023-10-18
Day.js initialization for mantine

ranquild:
How to verify:
- The app starts in different locales without exceptions


frontend/src/metabase/lib/i18n.js
```diff
@@ -26,8 +27,8 @@ export async function loadLocalization(locale) {
   setLocalization(translationsObject);
 }
 
-// Tell Moment.js to use the value of the start-of-week Setting for its current locale
-export function updateMomentStartOfWeek() {
+// Tell moment.js and day.js to use the value of the start-of-week Setting for its current locale
+export function updateMomentAndDayjsStartOfWeek() {
```
uladzimirdev, line 31: nit: I'd have two separate functions: `updateMoment*` and `updateDayjs*` and then a common `updateStartOfWeek` which contains both those functions. So upcoming removal of moment will be less painful

ranquild:
@uladzimirdev wdyt about a file with dayjs plugins that gets imported into the app and jest? Does it seem like a way forward?

uladzimirdev:
> @uladzimirdev wdyt about a file with dayjs plugins that gets imported into the app and jest? Does it seem like a way forward?

sounds good to me

frontend/src/metabase/lib/i18n.js
```diff
@@ -62,35 +57,66 @@ function setLanguage(translationsObject) {
...
     case "zh":
     case "zh-Hans":
       return "zh-cn";
     default:
-      return locale.toLowerCase();
+      return language.toLowerCase();
+  }
+}
+
+function getStartOfWeekDay() {
```
ranquild, line 106: Moved without changes

uladzimirdev (approved):
LGTM

Issue #32532, opened 2023-07-20
Add unsubscribe/undo-unsubscribe events to audit_log


luizarakaki:
We should log unsubscribe and undo-unsubscribe events to the activity_log table.

Topics: `subscription-unsubscribe`, `alert-unsubscribe`
Model: `Pulse`
details: `email`

Topics: `subscription-unsubscribe-undo`
Model: `Pulse`
details: `email`

Context: https://github.com/metabase/metabase/issues/29955

PR #34805, opened 2023-10-18
Add unsubscribe audit log events

qwef:
Closes https://github.com/metabase/metabase/issues/32532

### Description

Add unsubscribe audit log events.


luizarakaki (approved):
worked locally

PR #34808, opened 2023-10-18
Ensure that field picker icon does not get smushed

### LINKED ISSUE #31086: Large column names should wrap in the sidebar
### Describe the bug

Right now if we use a large column name a horizontal bar gets created in the left metric sidebar when there's an aggregation

### To Reproduce

1) create a column in a table like "COLUMN THAT WILL BE BIG BIG BIG VERY BIG IN ORDER TO ACCOMMODAT"
2) fill it with numerical values
3) then do a summarization on this column and check the left sidebar
image
4) see the scrollbar

### Expected behavior

we should wrap the contents here?

### Logs

_No response_

```JSON
MB version 1.46.4
```

### Severity

P3

### Additional context

_No response_

npfitz:
Closes https://github.com/metabase/metabase/issues/31086

### Description
Long names actually already wrap in both the field picker and column item components. This PR just helps the field picker icon not get squished, and adds a test to verify that that long names are rendered in full in the future (this test may not actually be necessary, since Jest looks at the DOM and generally speaking ellipsis is a style sheet thing).

### How to verify
1. New -> Question -> Orders
2. Create a custom column that is month([Created At]), and give it a very long name
3. Count, and group by the custom column
4. Examine the viz settings for the field picker (bar chart) and Column Item (table view)

### Demo
image
image

### Checklist

- [x] Tests have been added/updated to cover changes in this PR (Kinda... it was really already working)

### DIFF
--- a/frontend/src/metabase/visualizations/components/settings/ChartSettingFieldPicker.styled.tsx
+++ b/frontend/src/metabase/visualizations/components/settings/ChartSettingFieldPicker.styled.tsx
@@ -33,15 +33,14 @@ export const ChartSettingFieldPickerRoot = styled.div<ChartSettingFieldPickerRoo
     margin-left: 0;
     color: ${color("text-dark")};
     height: 0.625rem;
-
     ${props => props.disabled && "display: none;"}
+    flex-shrink: 0;
   }
 
   ${SelectButton.Content} {
     font-size: 0.875rem;
     line-height: 1rem;
     margin-right: 0.25rem;
-    text-overflow: ellipsis;
     max-width: 100%;
     overflow-wrap: anywhere;
     text-align: left;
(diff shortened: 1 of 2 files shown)


JesseSDevaney (dismissed):
LGTM :gem: 

Although, I agree that the new test is not informative. The ellipses is a CSS property and that unit test will pass regardless of it, since it is searching for a text node and not examining "visually". Maybe there is some way to check if `text-overflow: ellipses` is active on a particular text node, but I would just remove that test as it is not very useful.

npfitz:
Agreed, and I think the only way to get a real test for this would be a percy test. I'll check with the team and see if I'm able to add one.

JesseSDevaney:
Maybe we have changed how we want to use Percy, but I feel like we only use it for entire pages and more visual components like visualizations. I am not sure adding a Percy test for such a small thing like ellipses in the visualization settings would be valuable. But if you and others think it is valuable, then go for it :heart:

JesseSDevaney (approved):

PR #34810, opened 2023-10-18
[MLv2] Full `:column`s for `:dimensions` in `available-drill-thrus`

bshepherdson:
Previously only the `:column-name` was provided, and there is a self-proclaimed "icky hack" to guess at the full columns. The FE provides the (JSON) columns, so this PR converts and uses them.

The icky hack can be removed.

`:column-ref` is included because sometimes it's important that the refs use the same UUID.

### DIFF
--- a/frontend/src/metabase-lib/drills.unit.spec.ts
+++ b/frontend/src/metabase-lib/drills.unit.spec.ts
@@ -181,6 +181,7 @@ const AGGREGATED_ORDERS_COLUMNS = {
         "temporal-unit": "month",
       },
     ],
+    unit: "month",
   }),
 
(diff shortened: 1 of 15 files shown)


camsaul (dismissed):

bshepherdson:
* master
  * **PR #34810** <a href="https://app.graphite.dev/github/pr/metabase/metabase/34810" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/34810?utm_source=stack-comment).

camsaul (approved):

Issue #34778, opened 2023-10-18
[INTERNAL]: Start tagging CI built uberjars with the commit SHA1


nemanjaglumac:
@iethree started doing this for the official (release) uberjar builds. It makes so much sense!

The current version (v3) of the [official action for downloading artifacts](https://github.com/actions/download-artifact) allows matching only by the artifact name. We want to be able to find an artifact by its name but for a specific commit! Some gymnastics is needed to currently achieve this.
1. list all artifacts
2. filter all artifacts based on the provided artifact name first
3. and then find the matching artifact by matching the current commit SHA1 with the artifacts `head_sha`.

The current shape of a single artifact object received from the official REST API is this:
```js
{
  id: 991417000,
  node_id: 'MDg6QXJ0aWZhY3Q5OTE0MTcwMDA=',
  name: 'metabase-ee-uberjar',
  size_in_bytes: 368359369,
  url: 'https://api.github.com/repos/metabase/metabase/actions/artifacts/991417000',
  archive_download_url: 'https://api.github.com/repos/metabase/metabase/actions/artifacts/991417000/zip',
  expired: false,
  created_at: '2023-10-18T07:08:13Z',
  updated_at: '2023-10-18T07:08:14Z',
  expires_at: '2023-11-17T07:02:53Z',
  workflow_run: {
    id: 6557234070,
    repository_id: 30203935,
    head_repository_id: 30203935,
    head_branch: '34428-editing-notebook-question',
    head_sha: 'a0d8cabe33398e42f0b0575b97037a521814a9f1'
  }
}
```

If we start tagging artifacts with the commit SHA1, we'd get this:
```js
{
  id: 991417000,
  node_id: 'MDg6QXJ0aWZhY3Q5OTE0MTcwMDA=',
  name: 'metabase-ee-a0d8cabe33398e42f0b0575b97037a521814a9f1-uberjar', // <-------- edition + SHA1
  size_in_bytes: 368359369,
  url: 'https://api.github.com/repos/metabase/metabase/actions/artifacts/991417000',
  archive_download_url: 'https://api.github.com/repos/metabase/metabase/actions/artifacts/991417000/zip',
  expired: false,
  created_at: '2023-10-18T07:08:13Z',
  updated_at: '2023-10-18T07:08:14Z',
  expires_at: '2023-11-17T07:02:53Z',
  workflow_run: {
    id: 6557234070,
    repository_id: 30203935,
    head_repository_id: 30203935,
[...]

PR #34811, opened 2023-10-18
Store E2E uberjar artifact with a SHA1 in its name

nemanjaglumac:
This is only the first step in achieving https://github.com/metabase/metabase/issues/34778

And it's also a prerequisite for greatly simplifying and unblocking the work in https://github.com/metabase/metabase/pull/34647.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -146,6 +146,21 @@ jobs:
       - name: Build uberjar with ./bin/build.sh
         run: ./bin/build.sh
 
+      - name: Mark with the commit hash
+        run: echo ${{ github.sha }} > COMMIT-ID
+        shell: bash
+      - name: Calculate SHA256 checksum
+        run: sha256sum ./target/uberjar/metabase.jar > SHA256.sum
+        shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -146,6 +146,21 @@ jobs:
...
+        shell: bash
+      - name: Upload JARs as artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: metabase-${{ matrix.edition }}-${{ github.sha }}-uberjar
+          path: |
+            ./target/uberjar/metabase.jar
+            ./COMMIT-ID
+            ./SHA256.sum
+
```
nemanjaglumac, line 163: The only reason I didn't edit https://github.com/metabase/metabase/blob/master/.github/actions/prepare-uberjar-artifact/action.yml directly was that it's used in another workflow. I'd like to focus only on E2E for now to keep diffs as atomic as possible.

But the end goal will be using unified logic for both E2E and any other workflow that needs to build and store uberjar as an artifact.

iethree (approved):
LGTM!

PR #34813, opened 2023-10-18
Enable tests for #34443

### LINKED ISSUE #34443: [MLv2] [BE] `quick-filter` gets returned for non-metric column
`quick-filter` gets returned for FK column (PRODUCT_ID)

Check this reproduction unit test:
https://github.com/metabase/metabase/blob/5ab317540d064ac3d402508d0038629e8a041575/frontend/src/metabase-lib/drills.unit.spec.ts#L504

ALSO `quick-filter` doesn't get returned for CREATED_AT column in aggregated query - 
https://github.com/metabase/metabase/blob/7d9d4f42af6cf3ef31514b23184a3a79d449c3d2/frontend/src/metabase-lib/drills.unit.spec.ts#L1030

camsaul:
Fixes #34443

This is dependent on #34450


camsaul:
Tests are already enabled in `master`, this can be closed now.

PR #34814, opened 2023-10-18
Distribution drill should not apply binning to FK columns

### LINKED ISSUE #34343: `drill-thru/distribution` does not work for FK columns because it adds extra binning
`drill-thru/distribution` does not work for FK columns because it adds extra binning - https://github.com/metabase/metabase/blob/mblib-drill-thru-fe-all-drills-enabled/frontend/src/metabase-lib/drills.unit.spec.ts#L540

camsaul:
Fixes #34343


metamben (approved):
LGTM

PR #34815, opened 2023-10-18
Add filter-positions to the result of available-segments

### LINKED ISSUE #34738: [MLv2] BE - Add `filterPosition` to segment display info
When clicking an existing filter in the notebook editor, the FE should highlight a segment the filter is using (#34737)

We could follow the same approach we have for columns and make the `available-segments` method append `filterPosition` properties to segments used in filters on a given stage

metamben:
Fixes #34738.

### DIFF
--- a/frontend/src/metabase-lib/segments.ts
+++ b/frontend/src/metabase-lib/segments.ts
@@ -2,6 +2,9 @@ import * as ML from "cljs/metabase.lib.js";
 
 import type { SegmentMetadata, Query } from "./types";
 
-export function availableSegments(query: Query): SegmentMetadata[] {
-  return ML.available_segments(query);
+export function availableSegments(
+  query: Query,
+  stageIndex: number,
+): SegmentMetadata[] {
+  return ML.available_segments(query, stageIndex);
 }
--- a/src/metabase/lib/segment.cljc
+++ b/src/metabase/lib/segment.cljc
@@ -1,13 +1,13 @@
 (ns metabase.lib.segment
   "A Segment is a saved MBQL query stage snippet with `:filter`. Segments are always boolean expressions."
   (:require
+   [metabase.lib.filter :as lib.filter]
    [metabase.lib.metadata :as lib.metadata]
    [metabase.lib.metadata.calculation :as lib.metadata.calculation]
    [metabase.lib.metadata.protocols :as lib.metadata.protocols]
    [metabase.lib.options :as lib.options]
    [metabase.lib.ref :as lib.ref]
    [metabase.lib.schema :as lib.schema]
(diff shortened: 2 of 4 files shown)


frontend/src/metabase-lib/segments.ts
```diff
@@ -2,6 +2,9 @@ import * as ML from "cljs/metabase.lib.js";
 
 import type { SegmentMetadata, Query } from "./types";
 
-export function availableSegments(query: Query): SegmentMetadata[] {
-  return ML.available_segments(query);
+export function availableSegments(
+  query: Query,
+  stageIndex: number,
```
metamben, line 7: `stageIndex` is now needed to get the filters.

src/metabase/lib/segment.cljc
```diff
@@ -44,24 +44,13 @@
...
 (defmethod lib.metadata.calculation/display-info-method :metadata/segment
-  [query stage-number {:keys [id description], :as segment-metadata}]
+  [query stage-number {:keys [description filter-positions], :as segment-metadata}]
   (let [default-display-info-method (get-method lib.metadata.calculation/display-info-method :default)
         default-display-info        (default-display-info-method query stage-number segment-metadata)]
     (cond-> default-display-info
-      description                                   (assoc :description description)
-      (filtering-by-segment? query stage-number id) (assoc :selected true))))
+      description (assoc :description description)
+      filter-positions (assoc :filter-positions filter-positions))))
```
metamben, line 53: `:filter-positions` is set instead of `:selected`. This is symmetrical to filters.

test/metabase/lib/segment_test.cljc
```diff
@@ -100,7 +112,6 @@
                       :display-name      "PriceID-BBQ",
                       :long-display-name "PriceID-BBQ",
                       :effective-type    :type/Boolean,
-                      :description       "The ID is greater than 11 times the price and the name contains \"BBQ\".",
-                      :selected          true}]
+                      :description       "The ID is greater than 11 times the price and the name contains \"BBQ\"."}]
                     (map (partial lib/display-info query')
                          (lib/filters query'))))))))))
```
metamben, line 117: The display-info for filter clauses has neither `:selected` nor `:filter-positions`. @kulyk Is this OK? (If I'm not mistaken, `:selected` was only set of segments, regular filters didn't have it.)

kulyk: Yes, I wouldn't expect `selected` flag on filter clauses

kulyk (approved):
Works as expected 👍

camsaul (approved):

PR #34771, opened 2023-10-17
[Auditv2] Misc UI fixes

iethree:
Closes https://github.com/metabase/metabase/issues/34746

### Description

Fixes a few little UI issues with Auditv2

Audit collection should not have the timeline events button, and the permissions button shouldn't be in the `...` menu
. | .
--- | ---
Before | Screen Shot 2023-10-17 at 4 17 14 PM
After | Screen Shot 2023-10-17 at 4 16 50 PM 

Audit collections and questions and models should not be able to be made "official"

Before | After
---|---
Screen Shot 2023-10-17 at 5 42 20 PM | Screen Shot 2023-10-17 at 5 41 52 PM

~~The auto-apply filters toggle is disabled on read-only questions, it should look disabled now with faded opacity and a "not-allowed" cursor.~~
Removed in favor of using the mantine component in a separate PR in master

### How to verify

Describe the steps to verify that the changes are working as expected.

1. Load up an instance with auditv2 enabled (need an enterprise token)
2. check out your audit collection, dashboards, and models to see the above ☝️ 

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #34005, opened 2023-09-20
Mantine Switch

EmmadUsmani:
Closes https://github.com/metabase/metabase/issues/33993

### Description

Implements the [Mantine Switch component](https://v6.mantine.dev/core/switch/) in our `metabase/ui` component library.

### How to verify

1. `yarn storybook`
2. http://localhost:6006/?path=/story/inputs-switch--default-story
3. Confirm it looks and behaves like the Figma

### Demo

Screenshot 2023-10-03 at 4 18 51 PM

### Checklist

- [ ] Tests have been added/updated to cover changes in this PR

_Not applicable to component library._

PR #34817, opened 2023-10-18
Use Mantine Switch instead of Toggle in Dashboard Sidebar

iethree:
Related to https://github.com/metabase/metabase/pull/34771
[slack discussion:](https://metaboat.slack.com/archives/C02H619CJ8K/p1697586616727759) 

### Description

Mostly I want a switch with a disabled state. Instead of tweaking the deprecated component, we're just replacing it with the shiny new mantine one.

The styling is a bit different because I used the included label (which also gets a disabled state) so would appreciate design eyes on this, but the disabled state is lovely.

Old Toggle | New Switch
---|---
beforetoggle | afterSwtich

Disabled state:

Screen Shot 2023-10-18 at 4 38 42 PM

### DIFF
--- a/frontend/src/metabase/core/components/Toggle/Toggle.tsx
+++ b/frontend/src/metabase/core/components/Toggle/Toggle.tsx
@@ -11,6 +11,7 @@ export interface ToggleProps
   onChange?: (value: boolean) => void;
 }
 
+/** @deprecated use metabase/ui Switch instead */
 const Toggle = forwardRef(function Toggle(
   { className, value, small, color, onChange, ...rest }: ToggleProps,
   ref: Ref<HTMLInputElement>,
--- a/frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
+++ b/frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
@@ -1,5 +1,6 @@
 import styled from "@emotion/styled";
 
+import { Switch } from "metabase/ui";
 import { breakpointMaxSmall } from "metabase/styled-components/theme";
 
 import { color } from "metabase/lib/colors";
@@ -58,4 +59,12 @@ export const ContentSection = styled.div`
 
 export const DescriptionHeader = styled.h3`
   margin-bottom: 0.5rem;
+`;
+
+export const SpaceBetweenSwitch = styled(Switch)`
+  > div {
+    // target the inner div one level down only
+    display: flex;
+    justify-content: space-between;
+  }
 `;
--- a/frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
+++ b/frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
@@ -1,6 +1,5 @@
 import styled from "@emotion/styled";
 
-import { Switch } from "metabase/ui";
 import { breakpointMaxSmall } from "metabase/styled-components/theme";
 
 import { color } from "metabase/lib/colors";
@@ -59,12 +58,4 @@ export const ContentSection = styled.div`
 
 export const DescriptionHeader = styled.h3`
   margin-bottom: 0.5rem;
-`;
-
-export const SpaceBetweenSwitch = styled(Switch)`
-  > div {
-    // target the inner div one level down only
-    display: flex;
-    justify-content: space-between;
-  }
 `;
--- a/frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.tsx
+++ b/frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.tsx
@@ -4,6 +4,7 @@ import { t } from "ttag";
 import { PLUGIN_CACHING } from "metabase/plugins";
 import { useDispatch, useSelector } from "metabase/lib/redux";
 import MetabaseSettings from "metabase/lib/settings";
+import { Switch } from "metabase/ui";
 
 import { Timeline } from "metabase/common/components/Timeline";
 import EditableText from "metabase/core/components/EditableText";
@@ -25,7 +26,6 @@ import {
   HistoryHeader,
   ContentSection,
   DescriptionHeader,
-  SpaceBetweenSwitch,
 } from "./DashboardInfoSidebar.styled";
 
 type DashboardAttributeType = string | number | null | boolean;
@@ -92,10 +92,11 @@ export function DashboardInfoSidebar({
       </ContentSection>
 
       <ContentSection>
-        <SpaceBetweenSwitch
+        <Switch
           disabled={!canWrite}
           label={t`Auto-apply filters`}
           labelPosition="left"
+          variant="stretch"
           size="sm"
           id={autoApplyFilterToggleId}
           checked={dashboard.auto_apply_filters}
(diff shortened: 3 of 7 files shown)


frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
```diff
@@ -59,3 +60,11 @@ export const ContentSection = styled.div`
 export const DescriptionHeader = styled.h3`
   margin-bottom: 0.5rem;
 `;
+
+export const SpaceBetweenSwitch = styled(Switch)`
+  > div {
```
EmmadUsmani, line 65: We can use the [body class]([url](https://v6.mantine.dev/core/switch/?t=styles-api)) exposed by mantine, so instead this line would be

```css
  .emotion-Switch-body {
```

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.tsx
```diff
@@ -93,18 +92,15 @@ export function DashboardInfoSidebar({
...
-            id={autoApplyFilterToggleId}
-            value={dashboard.auto_apply_filters}
-            onChange={handleToggleAutoApplyFilters}
-          />
-        </FormField>
+        <SpaceBetweenSwitch
+          disabled={!canWrite}
+          label={t`Auto-apply filters`}
+          labelPosition="left"
+          size="sm"
```
EmmadUsmani, line 99: Do we want this to be `sm` or `md`? `sm` looks a bit small compared to the old one.

EmmadUsmani (dismissed):
Overall looks good! Left a couple of comments.

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
ranquild: Feels like a design system hack. I don't think we should do this. In the worst case just don't pass `label`.

ranquild (changes requested):
Design system hack, please take a look at the comment

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.tsx
iethree: the label size also scales with the size, `md` was very large for the label

iethree:
@kdoh I guess I'm a little unsure how to proceed here:
1. Should I not add the space-between behavior and keep the switch with the label?
1. Should I create my own label and replicate the mantine gray-if-disabled behavior so as not to modify the component's internal behavior?
2. Should we add an explicit "space-between" variant for the switch + label?
3. Should the toggle be bigger than `sm` ? the relative sizes between the labels and input are different in mantine components, and I figured that was intentional.

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
EmmadUsmani: FWIW @mazameli did explicitly ask if we can support `space-between` on the component https://github.com/metabase/metabase/pull/34005#issuecomment-1736399716

frontend/src/metabase/core/components/Toggle/Toggle.tsx
```diff
@@ -11,6 +11,7 @@ export interface ToggleProps
   onChange?: (value: boolean) => void;
 }
 
+/** @deprecated use metabase/ui Switch instead */
```
kdoh, line 14: ❤️

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
kdoh: I think a little communication whoopsie happened with the original component, but it's no big deal. I talked to @mazameli and the intent is that this "stretch" to fill Switch should be part of our Switch component's option space. I'm with @ranquild that it feels like we're doing something messy if in order to get that to happen we have to write CSS at each point of use.

Happily I think we can use [a custom variant](https://v7.mantine.dev/styles/variants-sizes#adding-custom-variants) and just bake this behavior into Switch itself.

So I think to fix that piece, we do the following
- Add a custom `stretchy` variant to Swtich, and add this `flex` and `space-between` CSS to that variant in the style for the component.
- Document it 
- Use that here `<Switch variant="stretch" labelPosition="left" ... />`

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.tsx
kdoh: I ran into a data migration issue trying to spin up the branch but if the `sm` size is what's in the screenshots that looks ok to me. It **is** expected that there's a different between the old stuff and the new stuff since the Mantine based components are using an updated type system. The different isn't supposed to be crazy though. 

Out of curiosity what's the "default" size for Switch? Ideally the default one should be kicking out something closer to 14px "normal" type I would hope? If we have to set `sm` on all switches just to get the base case, I think something screwy is going on.

kdoh (changes requested):
Let me know if this path forward makes sense @iethree.

frontend/src/metabase/dashboard/components/DashboardInfoSidebar/DashboardInfoSidebar.styled.tsx
iethree: Can do. 👍  I love working with designers. ❤️

iethree:
OK, I achieved the same styling by adding a new "stretch" variant. 🥳  Configuring a new variant in Mantine was a lovely experience and I will recommend it to all my friends and family members. 😁

frontend/src/metabase/ui/components/inputs/Switch/Switch.styled.tsx
```diff
@@ -116,5 +116,13 @@ export const getSwitchOverrides = (): MantineThemeOverride["components"] => ({
         },
       };
     },
+    variants: {
```
ranquild, line 119: Could you add a block in the component story for this variant? And a control to toggle variants? And please ask @kdoh to review

ranquild (commented):
Comment inline

kdoh:
I'll be a ✅ as soon as the `stretch` variant is documented. Nice work!

frontend/src/metabase/ui/components/theme/DatesProvider/DatesProvider.tsx
```diff
@@ -1,7 +1,7 @@
 import type { ReactNode } from "react";
-import dayjs from "dayjs";
 import { DatesProvider as MantineDatesProvider } from "@mantine/dates";
 import type { DatesProviderSettings, DayOfWeek } from "@mantine/dates";
+import dayjs from "metabase/lib/dayjs";
```
iethree, line 5: @ranquild dayjs throws an error preventing storybook from working at all for me locally because we're using an unextended dayjs here, but relying on a method from one of the extensions.

Screen Shot 2023-10-20 at 1 10 05 PM

It works in the main app and jest because we import this in app.js and jest-setup.js

Let me know if you think there's a better pattern than just exporting the extended object from "metabase/lib/dayjs";

kdoh (dismissed):

frontend/src/metabase/ui/components/theme/DatesProvider/DatesProvider.tsx
iethree: nevermind, looks like this got fixed since I branched off master

ranquild (approved):
FE LGTM

PR #34818, opened 2023-10-18
Auditv2 main more audit log events

### LINKED ISSUE #34099: Add database related events

escherize:
Resolves #34099

Populates the audit log for the following events:

- [x] ~database-add~ `database-create`
Details: ~db_name~, ~db_id~, name, id, engine
    - [x] test

- [x] `database-update`
Details: ~db_name~, ~db_id~, name, id, engine
    - [x] test

- [x] `database-delete`
Details: ~db_name~, ~db_id~, name, id, engine
    - [x] test

- [x] `database-manual-sync`
Details: ~db_name~, ~db_id~, name, id, engine
    - [x] test

- [x] `database-manual-scan`
Details: ~db_name~, ~db_id~, name, id, engine
    - [x] test

- [x] `database-discard-field-values`
Details: ~db_name~, ~db_id~, name, id, engine
    - [x] test

- [x] `table-manual-scan`
Details: ~table_name~, ~table_id~, name, id, db_id
    - [x] test

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
@@ -511,10 +511,10 @@
 
       (testing "An audit log entry is generated when a manually triggered re-scan occurs"
         (mt/with-model-cleanup [:model/AuditLog :model/Activity]
-          (do (with-all-users-data-perms {(mt/id) {:data-model {:schemas {"PUBLIC" {table-id :all}}}}}
-              (mt/user-http-request :rasta :post 200 (format "table/%d/rescan_values" table-id)))
-            (is (= table-id (:model_id (mt/latest-audit-log-entry))))
-            (is (= table-id (-> (mt/latest-audit-log-entry) :details :id)))))))
+          (with-all-users-data-perms {(mt/id) {:data-model {:schemas {"PUBLIC" {table-id :all}}}}}
+            (mt/user-http-request :rasta :post 200 (format "table/%d/rescan_values" table-id)))
+          (is (= table-id (:model_id (mt/latest-audit-log-entry))))
+          (is (= table-id (-> (mt/latest-audit-log-entry) :details :id))))))
 
     (testing "POST /api/table/:id/discard_values"
       (testing "A non-admin can discard field values if they have data model perms for the table"
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
@@ -507,7 +507,14 @@
           (with-all-users-data-perms {(mt/id) {:data       {:schemas :block :native :none}
                                                :data-model {:schemas {"PUBLIC" {(mt/id :venues) :all}}}}}
             (mt/user-http-request :rasta :post 200 (format "table/%d/rescan_values" (mt/id :venues)))))
-        (is (= [1 2 3 4] (t2/select-one-fn :values FieldValues, :field_id (mt/id :venues :price))))))
+        (is (= [1 2 3 4] (t2/select-one-fn :values FieldValues, :field_id (mt/id :venues :price)))))
+
+      (testing "An audit log entry is generated when a manually triggered re-scan occurs"
+        (mt/with-model-cleanup [:model/AuditLog :model/Activity]
+          (do (with-all-users-data-perms {(mt/id) {:data-model {:schemas {"PUBLIC" {table-id :all}}}}}
+              (mt/user-http-request :rasta :post 200 (format "table/%d/rescan_values" table-id)))
+            (is (= table-id (:model_id (mt/latest-audit-log-entry))))
+            (is (= table-id (-> (mt/latest-audit-log-entry) :details :id)))))))
 
     (testing "POST /api/table/:id/discard_values"
       (testing "A non-admin can discard field values if they have data model perms for the table"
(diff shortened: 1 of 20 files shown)


src/metabase/api/database.clj
```diff
@@ -976,6 +976,7 @@
   {id ms/PositiveInt}
   ;; just wrap this in a future so it happens async
   (let [db (api/write-check (t2/select-one Database :id id))]
+    (events/publish-event! :event/database-manual-sync db)
```
escherize, line 979: I see publish-event! as the "entrypoint" we can use for the audit system (and maybe whatever other system cares about these events). So I always called publish-event! from where the event-in-question occured.

src/metabase/events/audit_log.clj
```diff
@@ -63,10 +63,11 @@
   [topic dashboard]
   (audit-log/record-event! topic dashboard))
 
-(derive ::table-read-event ::event)
-(derive :event/table-read ::table-read-event)
```
escherize, line 67: renamed `::table-read-event` -> `::table-event` since that matches the pattern in this ns better

src/metabase/models/persisted_info.clj
```diff
@@ -137,12 +136,13 @@
 (defn ready-unpersisted-models!
   "Looks for all new models in database and creates a persisted-info ready to be synced."
   [database-id]
-  (let [cards (t2/select Card {:where [:and
-                                       [:= :database_id database-id]
-                                       [:= :dataset true]
-                                       [:not [:exists {:select [1]
-                                                       :from [:persisted_info]
-                                                       :where [:= :persisted_info.card_id :report_card.id]}]]]})]
+  (let [cards (t2/select :model/Card
```
escherize, line 139: A bunch of `Model` -> `:model/Model` sort of changes were added to avoid circular dependencies that might be gone now. But it's still worth changing these afterall, right?

src/metabase/util.cljc
```diff
@@ -843,3 +843,12 @@
...
       (apply distinct? xs)))
+
+(defn deep-sort-map
+  "Converts a map (with potentially nested maps as values) into a sorted map of sorted maps."
+  [m]
+  (into (sorted-map)
+        (zipmap
+         (keys m)
+         (map (fn [val] (if (map? val) (deep-sort-map val) val))
+              (vals m)))))
```
noahmoss, line 854: nit: small suggestions - you can just map over the input directly instead of the keys and values separately, and use `cond->` for the conditional. Maybe sth like this?
```suggestion
        (map (fn [[k v]]
               [k (cond-> v (map? v) deep-sort-map)])
             m)))
```

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -507,7 +507,14 @@
           (with-all-users-data-perms {(mt/id) {:data       {:schemas :block :native :none}
                                                :data-model {:schemas {"PUBLIC" {(mt/id :venues) :all}}}}}
             (mt/user-http-request :rasta :post 200 (format "table/%d/rescan_values" (mt/id :venues)))))
-        (is (= [1 2 3 4] (t2/select-one-fn :values FieldValues, :field_id (mt/id :venues :price))))))
+        (is (= [1 2 3 4] (t2/select-one-fn :values FieldValues, :field_id (mt/id :venues :price)))))
+
+      (testing "An audit log entry is generated when a manually triggered re-scan occurs"
+        (mt/with-model-cleanup [:model/AuditLog :model/Activity]
```
noahmoss, line 513: In general I think it's OK for tests to write to the audit log without cleanup, right? Lots of other tests will also do things that implicitly trigger audit events to be written and we can't expect every developer to clean up the table.

noahmoss (commented):
Looks good, just a few comments

noahmoss (approved):

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -507,7 +507,14 @@
...
+          (do (with-all-users-data-perms {(mt/id) {:data-model {:schemas {"PUBLIC" {table-id :all}}}}}
+              (mt/user-http-request :rasta :post 200 (format "table/%d/rescan_values" table-id)))
```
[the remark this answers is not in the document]
escherize: I'll fix this

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
escherize: Yeah, It is totally okay for tests to write to the audit log and _not_ clean them up. Maybe most tests will touch the audit log at some point. I don't expect anyone to clean them up in tests.

I see about 29 spots where we have `mt/with-model-cleanup.*:model/AuditLog`. Do you think should we actually just remove them? Reading through, some delete audit logs that would point to nonexistant things. But that isn't actually a problem either, afaik.

src/metabase/util.cljc
escherize: I used to use zipkeys to re-write map-vals all the time, is that weird? :) Now that we have `clojure.core/update-vals` I will go ahead and use that instead.

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
escherize: If so let's get that into a followup since its not an urgent change. I will come back to it once we have auditv2 shipped.

luizarakaki (approved):
For database-update is it possible to add the properties that change? Or even paste the entire object there (excluding `details`)?

The event is not very useful without this data...

Besides that, looking good

PR #34826, opened 2023-10-19
 Add ::zoned-datetime support for CSV uploads

### LINKED ISSUE #34137: Uploaded CSVs can't contain zoned datetimes
Some CSV downloads from Metabase include the TZ offset in them (such as Model #4523, discovered by @uladzimirdev ):

```
ID,Customer Name,Created At,...
<the ID>, <the customer name>,2023-09-26T04:21:10.645527-07:00,
```

If you then re-upload this to Metabase, the datetime is not recognized since we use `local-date-time` instead of `offset-date-time` for parsing. We also use TZ-less types for the resulting columns in the table.

We should probably detect if TZ data is included and then map that to `TIMESTAMP WITH TIMEZONE` columns.

tsmacdonald:
Fixes #34137 
    
This required driver-specific parsing since MySQL doesn't support    timestamps with timezones, so I made `upload-type->parser` a driver    method and pulled a lot of the work into a separate `upload.parsing`
namespace. So there's a lot of diff noise for that reason, but the things to focus on are

a) that `driver/upload-type->parser` happened at all
b) the implementations for `::zoned-datetime`, especially the MySQL one since it's a little hairy
c) the sadness of having to write `upload/current-database` since we need to know what TZ the MySQL uploads DB is using


src/metabase/driver/sql_jdbc.clj
```diff
@@ -174,3 +176,56 @@
   (let [sql (driver.sql/set-role-statement driver role)]
     (with-open [stmt (.createStatement ^Connection conn)]
       (.execute stmt sql))))
+
+(defmethod driver/upload-type->parser [:sql-jdbc :metabase.upload/varchar-255]
```
calherries, line 180: I find it's easier to read if you use ns aliases
```suggestion
(defmethod driver/upload-type->parser [:sql-jdbc ::upload/varchar-255]
```

test/metabase/upload_test.clj
```diff
@@ -422,6 +437,29 @@
...
+  (testing "Upload a CSV file with a datetime column"
+    (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
+      (mt/with-empty-db
+        (with-mysql-local-infile-activated
+          (upload/load-from-csv!
+           driver/*driver*
+           (mt/id)
+           "upload_test"
+           (csv-file-with ["zoned_datetime"
+                           "2022-01-01T01:00:00-07:00"])))
```
calherries, line 450: I would love to see the test check that `trim` is used
```suggestion
                           "2022-01-01T01:00:00-07:00  "])))
```

src/metabase/driver/mysql.clj
```diff
@@ -617,7 +619,30 @@
...
+(defn- offset-date-time->local-date-time
+  "Remove the offset from a string datetime, returning a LocalDateTime in whatever timezone the `database` is configured
+  to use. This is necessary since MySQL doesn't support timezone offsets and so we need to calculate one by hand. This
+  is used for the upload parser below, which was looking messy with error-handling so it made sense to extract the logic."
+  [s database]
+  (let [offset-time (t/offset-date-time (str/trim s))
+        zone-id     (t/zone-id (qp.timezone/results-timezone-id database))]
+    (t/local-date-time offset-time zone-id )))
+
+(defmethod driver/upload-type->parser [:mysql :metabase.upload/zoned-datetime]
```
calherries, line 634: I don't think you need this. MySQL does support timezone offsets on insertion. It just doesn't store the time zone it was inserted with. A string like '2023-10-19 10:00:00+00:00' should be able to be inserted into MySQL directly.

I just tested this and it works as I expected.
```
CREATE TABLE timestamps (
    event_timestamp TIMESTAMP NOT NULL
);

INSERT INTO timestamps (event_timestamp) VALUES ('2023-10-19 10:00:00+00:00'), ('2023-10-19 10:00:00+01:00');

SELECT * from timestamps;
```
The result is:
```
+---------------------+
| event_timestamp     |
+---------------------+
| 2023-10-19 13:00:00 |
| 2023-10-19 12:00:00 |
+---------------------+
```
Because my system timezone offset is UTC+3 right now

calherries (changes requested):

src/metabase/upload.clj
```diff
@@ -127,29 +133,31 @@
...
-    - ::datetime
-    - nil, in which case other functions are expected to replace it with ::text as the catch-all type
+
+    - `::boolean`
+    - `::int`
+    - `::float`
+    - `::varchar-255`
+    - `::date`
+    - `::datetime`
+    - `::zoned-datetime`
```
calherries, line 143: This should be `offset-datetime`, not `zoned-datetime`. A time zone is Europe/Amsterdam where an offset is like UTC+2. The difference is significant because a time zone has multiple offsets depending on the time of the year

src/metabase/upload/parsing.clj
```diff
@@ -0,0 +1,87 @@
...
+  (try
+    (t/local-date-time (t/local-date s) (t/local-time "00:00:00"))
+    (catch Exception _
+      (try
+        (t/local-date-time s)
+        (catch Exception _
+          (throw (IllegalArgumentException.
+                  (tru "{0} is not a recognizable datetime" s))))))))
+
+(defn parse-zoned-datetime
```
calherries, line 46:
```suggestion
(defn parse-offset-datetime
```

src/metabase/driver/sql_jdbc.clj
tsmacdonald: fixed in mysql.clj, but upload isn't `:require`d in sql_jdbc.clj and I don't think I should add the dependency to save on the typing.

src/metabase/driver/mysql.clj
tsmacdonald: I was led to believe that as well from the docs, but this test was failing:


```clojure
           (csv-file-with ["offset_datetime"
                           "2022-01-01T01:00:00-07:00"])))
...
            (is (= "2022-01-01T08:00:00Z"
                   (second (first (rows-for-table table)))))))))))
```


```
Fail in load-from-csv-offset-datetime-test
Upload a CSV file with a datetime column 
...

expected: "2022-01-01T08:00:00Z"

  actual: "2022-01-01T01:00:07Z"	  
    diff: - "2022-01-01T08:00:00Z"	  
          + "2022-01-01T01:00:07Z"	    
```

Treating the offset as a change to the _seconds_ is weird and maybe java-time's fault instead of MySQLs.

That to one side, I believe the offset support was introduced in MySQL 8 ([link](https://github.com/metabase/metabase/issues/34137#issuecomment-1744993754)), so making it explicit the way it is here maintains support for earlier versions. It happens that MySQL < 8 EOLs in about two weeks, so we're allowed to break support, but "write special logic to make extra sure we're handling offsets right and also support old versions" seemed more appealing than trying to figure out what's up with java-time.

src/metabase/upload.clj
tsmacdonald: good catch, fixed and pushed (along with other comments)

src/metabase/driver/h2.clj
```diff
@@ -582,7 +582,8 @@
     ::upload/float                    [(keyword "DOUBLE PRECISION")]
     ::upload/boolean                  [:boolean]
     ::upload/date                     [:date]
-    ::upload/datetime                 [:timestamp]))
+    ::upload/datetime                 [:timestamp]
+    ::upload/zoned-datetime           [:timestamp-with-time-zone]))
```
calherries, line 586:
```suggestion
    ::upload/offset-datetime          [:timestamp-with-time-zone]))
```

src/metabase/driver/postgres.clj
```diff
@@ -761,7 +761,8 @@
     ::upload/float                    [:float]
     ::upload/boolean                  [:boolean]
     ::upload/date                     [:date]
-    ::upload/datetime                 [:timestamp]))
+    ::upload/datetime                 [:timestamp]
+    ::upload/zoned-datetime           [:timestamp-with-time-zone]))
```
calherries, line 765:
```suggestion
    ::upload/offset-datetime          [:timestamp-with-time-zone]))
```

calherries (changes requested):

src/metabase/driver/mysql.clj
calherries: This isn't an issue with java-time. It's an issue with `metabase.driver.mysql/row->tsv` and how MySQL ingests timestamps from TSVs.

I ran the `load-from-csv-offset-datetime-test` test with this `git diff`:

```diff
diff --git a/src/metabase/driver/mysql.clj b/src/metabase/driver/mysql.clj
index 05f502e502..1431c0188b 100644
--- a/src/metabase/driver/mysql.clj
+++ b/src/metabase/driver/mysql.clj
@@ -631,7 +631,7 @@
         zone-id     (t/zone-id (qp.timezone/results-timezone-id database))]
     (t/local-date-time offset-time zone-id )))
 
-(defmethod driver/upload-type->parser [:mysql ::upload/offset-datetime]
+#_(defmethod driver/upload-type->parser [:mysql ::upload/offset-datetime]
   [_ _]
   (comp
    (fn [s]
@@ -699,7 +699,7 @@
     (let [temp-file (File/createTempFile table-name ".tsv")
           file-path (.getAbsolutePath temp-file)]
       (try
-        (let [tsvs (map (partial row->tsv (count column-names)) values)
+        (let [tsvs (dev/tap>-spy (map (partial row->tsv (count column-names)) (dev/tap>-spy values)))
               sql  (sql/format {::load   [file-path (keyword table-name)]
                                 :columns (map keyword column-names)}
                                :quoted  true
```

The tapped result:
```
1st tap (values):
((#t "2022-01-01T01:00-07:00"))

2nd tap (map (partial row->tsv (count column-names)) (dev/tap>-spy values))
("2022-01-01T01:00-07:00")
```

Then I tested how this value would get converted from MySQL:

With this test CSV:
```
event_timestamp
"2022-01-01T01:00:00-07:00"
"2022-01-01T01:00-07:00"
```

And this SQL:
```
DROP TABLE IF EXISTS my_table;

CREATE TABLE my_table (
    event_timestamp TIMESTAMP
);

LOAD DATA LOCAL INFILE '/Users/callumherries/meta/metabase/test.csv'
INTO TABLE my_table
FIELDS TERMINATED BY ','
ENCLOSED BY '"'
LINES TERMINATED BY '\n'
IGNORE 1 ROWS
(event_timestamp);

SELECT * FROM my_table;
```

The result:
```
+---------------------+
| event_timestamp     |
+---------------------+
| 2022-01-01 10:00:00 |
| 2022-01-01 01:00:07 |
+---------------------+
2 rows in set (0.00 sec)
```

tsmacdonald: Thanks for looking into that so exhaustively.

Especially since [we're off of MySQL < 8 now](https://metaboat.slack.com/archives/C010L1Z4F9S/p1698076249487039) I fixed the TSV generation as indicated, and moved `upload-type->parser` out of driver-land and into `upload.parsing`.

test/metabase/upload_test.clj
```diff
@@ -422,6 +442,29 @@
...
+  (testing "Upload a CSV file with a datetime column"
+    (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
+      (mt/with-empty-db
+        (with-mysql-local-infile-activated
+          (upload/load-from-csv!
+           driver/*driver*
+           (mt/id)
+           "upload_test"
+           (csv-file-with ["offset_datetime"
+                           "2022-01-01T01:00:00-07:00"])))
```
calherries, line 455: I'm concerned the driver code will convert a GMT timezone to a string like "2022-01-01T01:00:00Z" with a "Z" at the end before inserting, and that it will not be a valid format for the database.

I tested this with MySQL and indeed it was the case:
```
DROP TABLE IF EXISTS timestamps;

CREATE TABLE timestamps (
    event_timestamp TIMESTAMP NOT NULL
);

INSERT INTO timestamps (event_timestamp) VALUES ('2022-01-01T01:00:00Z');

SELECT * from timestamps;
```

I'd suggest covering this corner case as a test here.

test/metabase/upload_test.clj
```diff
@@ -114,17 +117,20 @@
...
-           ["2022-01-01"                    #t "2022-01-01"       date-type]
-           ["2022-01-01T01:00:00"           #t "2022-01-01T01:00" datetime-type]
-           ["2022-01-01T01:00:00.00"        #t "2022-01-01T01:00" datetime-type]
-           ["2022-01-01T01:00:00.000000000" #t "2022-01-01T01:00" datetime-type]]]
+           ;; Date-related
+           [" 2022-01-01 "                    #t "2022-01-01"             date-type]
+           [" 2022-01-01T01:00:00 "           #t "2022-01-01T01:00"       datetime-type]
+           [" 2022-01-01T01:00:00.00 "        #t "2022-01-01T01:00"       datetime-type]
+           [" 2022-01-01T01:00:00.000000000 " #t "2022-01-01T01:00"       datetime-type]
+           [" 2022-01-01T01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]]]
```
calherries, line 130: The implementation of `parse-offset-datetime` uses `t/offset-date-time`, which can tolerate more formats e.g.

```
(t/offset-date-time "2020-01-01T00:00Z")
;; => #t "2020-01-01T00:00Z"

(t/offset-date-time "2020-01-01T00:00+01")
;; => #t "2020-01-01T00:00+01:00"
```

I think we should tolerate these other formats too, but these tests don't verify that they are supported. I think we should add tests for them to be sure they are tolerated in the future.

calherries (changes requested):

test/metabase/upload_test.clj
tsmacdonald: Handled in [this commit](https://github.com/metabase/metabase/pull/34826/commits/8c3027f7cfbf1bc96a651e32977f407dffdf2da3)

calherries (approved):
Looks good to me assuming the tests pass

src/metabase/driver/mysql.clj
```diff
@@ -632,35 +632,45 @@
 
 (sql/register-clause! ::load format-load :insert-into)
 
+(defn- offset-datetime->unoffset-datetime
+  "Remove the offset from a datetime, returning a string represnetation in whatever timezone the `database` is configured
+  to use. This is necessary since MariaDB doesn't support timezone offsets and so we need to calculate one by hand."
+   [driver database ^OffsetDateTime offset-time]
+  (let [zone-id     (t/zone-id (driver/db-default-timezone driver database))]
```
calherries, line 639:
```suggestion
  (let [zone-id (t/zone-id (driver/db-default-timezone driver database))]
```

src/metabase/driver/mysql.clj
```diff
@@ -632,35 +632,45 @@
 
 (sql/register-clause! ::load format-load :insert-into)
 
+(defn- offset-datetime->unoffset-datetime
+  "Remove the offset from a datetime, returning a string represnetation in whatever timezone the `database` is configured
+  to use. This is necessary since MariaDB doesn't support timezone offsets and so we need to calculate one by hand."
```
calherries, line 637:
```suggestion
  to use. This is necessary since MariaDB doesn't support timestamp with time zone literals and so we need to calculate one by hand."
```
Technically the above suggestion is correct. MariaDB does support `CONVERT_TZ` for example, which kinda means it supports "timezone offsets"

src/metabase/driver/mysql.clj
```diff
@@ -632,35 +632,45 @@
...
 (defmulti ^:private value->string
   "Convert a value into a string that's safe for insertion"
-  {:arglists '([val])}
-  type)
+  {:arglists '([driver val])}
+  (fn [_ val] (type val)))
 
 (defmethod value->string :default
-  [val]
+  [_ val]
```
calherries, line 648:
```suggestion
  [_driver val]
```

calherries (approved):
Reapproving, just small suggestions

test/metabase/upload_test.clj
```diff
@@ -422,6 +443,38 @@
                       (t2/select-one Field :database_position 1 :table_id (:id table)))))
             (is (some? table))))))))
 
+(deftest load-from-csv-offset-datetime-test
+  (testing "Upload a CSV file with a datetime column"
+    (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
+      (mt/with-empty-db
+        (with-redefs [driver/db-default-timezone (constantly "Z")
+                      upload/current-database    (constantly (mt/db))]
```
calherries, line 451: I think you could have avoided the redef here, which couples the test with the implementation
```
(tu/with-temporary-setting-values [uploads-database-id (:id (mt/db))]
```

Issue #34733, opened 2023-10-17
[Epic] Prevent users from adding questions from personal collections to dashboards in public collections


WiNloSt:
**Links**
- issue: https://github.com/metabase/metabase/issues/34045
- eng doc: https://www.notion.so/metabase/Prevent-users-from-adding-questions-from-personal-collections-to-dashboards-in-public-collections-d94438b13f974e0fbe97d99f732433d1

### Milestones overview

```[tasklist]
### Milestone 1: Collection changes
- [x] [Dashboard changes](#dashboard-changes)
- [x] [Query builder changes](#query-builder-changes)
- [x] [Handle most recently viewed dashboard](#most-recently-viewed-dashboard)
- [x] [Hide `+ Create a new dashboard` option](#hide-create-a-new-dashboard-option)
- [x] [Hide dashboards in public collections in dashboard picker (query builder)](#hide-dashboards-in-dashboard-picker)
- [x] [Handle dashboard changes and query builder changes after moving dashboards/questions to a different kind of collection](#handle-moving-items-to-different-collection-type)
- [x] Add E2E tests for handling moving dashboards/questions
- [ ] https://github.com/metabase/metabase/pull/35135
```

#### Testing plan for milestone 1
Dashboards
1. Add a question to a dashboard in a public collection
    **expectation**:
    - hides all personal collections
    - show all questions
1. Add a question to a dashboard in a personal collection
    **expectation**:
    - shows all collections
    - show all questions
1. move a dashboard from a personal collection to a public collection and test 1)
    **expectation**: hides all personal collections
    - hides all personal collections
    - show all questions
1. move a dashboard from a public collection to a personal collection and test 2)
    **expectation**:
    - shows all collections
    - show all questions

Query builder

Newly created question
1. Select a dashboard for a question in a public collection
    **expectation**: shows all collections
1. Select a dashboard for a question in a personal collection
    **expectation**: hides all public collections
1. Open public collection for a question in a public collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. Open personal collection for a question in a public collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. Open public collection for a question in a personal collection (this only happens for the root collection, as other public collections are already hidden)
    **expectation**:
        - hides `+ Create a new dashboard` option
        - hides dashboards
1. Open personal collection for a question in a personal collection
    **expectation**:
        - shows `+ Create a new dashboard` option
[...]

PR #34827, opened 2023-10-19
Hydrate `collection.is_personal`

qnkhuat:
BE for milestone 1 of https://github.com/metabase/metabase/issues/34733

Hydrate `collection.is_personal` for:
- GET /api/dashboard/:id
- GET /api/collection
- GET /api/activity/most_recently_viewed_dashboard
- GET /api/card/:id
- POST /api/card/:id
- PUT /api/card/:id
- PUT /api/dashboard/:id


WiNloSt (commented):
Everything looks good except the last task *Handle dashboard changes and query builder changes after moving dashboards/questions to a different kind of collection* of milestone 1 which would [require hydrating 2 more endpoints](https://github.com/metabase/metabase/issues/34733#handle-moving-items-to-different-collection-type)

I'll ✅ once those 2 endpoints are included.

qnkhuat:
@WiNloSt just hydrated those 2

WiNloSt:
@qnkhuat I found one difference between `GET /api/dashboard/:id` and `PUT /api/dashboard/:id` is that for collection in a root collection the GET endpoint will return `collection` key e.g.
```json
{
  "metabase.models.collection.root/is-root?": true,
  "authority_level": null,
  "name": "Our analytics",
  "is_personal": false,
  "id": "root",
  "can_write": true
}
```

But if we move a dashboard to the root collection `PUT /api/dashboard/:id` would have `collection` as `null`.

I'm not sure what is BE's expectation for this. I'd expect both to be the same. I've seen we've some discrepancies between the notion of the root collection ID which currently some endpoints use `"root"` for the ID while some use `null` like in this case.

If we're able to call `GET /api/collection/root` and it's currently returning the root collection
```json
{
  "authority_level": null,
  "can_write": true,
  "name": "Our analytics",
  "effective_ancestors": [],
  "effective_location": null,
  "parent_id": null,
  "id": "root",
  "is_personal": false
}
```

I'd expect `PUT /api/dashboard/:id` to return `collection_id: "root"` and non null `collection`.

What do you think?

WiNloSt:
I'm quoting @qnkhuat 's DM here
> hey Kelvin, can you use that `collection: nil` is the root collection?
>
> because I believe collection_id: nil indicates that and that’s how the app works.
> My hesitation for change is we might miss cases where we don’t handle this edge case, and FE will still see collection: nil  for some APIs.

I agreed with him. So I'll approve this PR.

WiNloSt (approved):
Looks good for the milestone 1

src/metabase/models/collection.clj
```diff
@@ -1143,6 +1143,28 @@
...
+  [collections]
+  (let [personal-collection-ids (t2/select-pks-set :model/collection :personal_owner_id [:not= nil])
+        location-is-personal    (fn [location]
+                                  (boolean
+                                   (and (string? location)
+                                        (some #(str/starts-with? location (format "/%d/" %)) personal-collection-ids))))]
+    (map (fn [{:keys [location personal_owner_id] :as coll}]
+           (assoc coll :is_personal (or (some? personal_owner_id)
+                                        (location-is-personal location))))
+         collections)))
```
calherries, line 1166: You can define these with one `mi/define-batched-hydration-method` call, to avoid having two hydration keys for the same "thing".

```clojure
(mi/define-batched-hydration-method collections-is-personal
  :collections_is_personal
  "Efficiently hydrate the `:is_personal` property of a sequence of Collections.
  `true` means the collection is or nested in a personal collection."
  [collections]
  (if (= (count collections) 1)
    (is-personal-collection-or-descendant-of-one? collection)
    (let [personal-collection-ids (t2/select-pks-set :model/collection :personal_owner_id [:not= nil])
          location-is-personal    (fn [location]
                                    (boolean
                                     (and (string? location)
                                          (some #(str/starts-with? location (format "/%d/" %)) personal-collection-ids))))]
      (map (fn [{:keys [location personal_owner_id] :as coll}]
             (assoc coll :is_personal (or (some? personal_owner_id)
                                          (location-is-personal location))))
           collections))))
```

test/metabase/api/activity_test.clj
```diff
@@ -68,6 +68,29 @@
...
+  (mt/with-temp [:model/Collection coll   {:name "Analytics"}
+                 :model/Dashboard  dash-1 {:collection_id (t2/select-one-pk :model/Collection :personal_owner_id (mt/user->id :crowberto))}
+                 :model/Dashboard  dash-2 {:collection_id (:id coll)}]
+    (mt/with-model-cleanup [ViewLog]
+      (mt/with-test-user :crowberto
+        (testing "view a dashboard in a personal collection"
+          (mt/with-temporary-setting-values [user-recent-views              []
+                                             most-recently-viewed-dashboard nil]
+            (events/publish-event! :event/dashboard-read (assoc dash-1 :actor_id (mt/user->id :crowberto)))
+            (let [crowberto-personal-coll (t2/select-one :model/Collection (t2/select-one-pk :model/Collection :personal_owner_id (mt/user->id :crowberto)))]
```
calherries, line 81:
```suggestion
            (let [crowberto-personal-coll (t2/select-one :model/Collection :personal_owner_id (mt/user->id :crowberto))]
```

calherries: Or extract out this out to a `let` binding  and use in line 73

test/metabase/api/activity_test.clj
```diff
@@ -47,7 +47,7 @@
              ;; view log entries look for the `:actor_id` in the item being viewed to set that view's :user_id
              (assoc item :actor_id (mt/user->id :crowberto))))
           (testing "most_recently_viewed_dashboard endpoint shows the current user's most recently viewed dashboard."
-            (is (= dash-3 #_dash-2 ;; TODO: this should be dash-2, because dash-3 is archived
+            (is (= (assoc dash-3 :collection nil) #_dash-2 ;; TODO: this should be dash-2, because dash-3 is archived
```
calherries, line 50: Can you explain why this is nil to me? In the `most-recently-viewed-dashboard-views-include-colllection-test` it is clearly non-nil so I'm just wondering what's the difference here

qnkhuat: because `dash-3` here is in the root (aka Our analytics) collection, and it's a virtual collection, which is why it's nil here.

the dashboard in `most-recently-viewed-dashboard-views-include-colllection-test` is placed under a collection.

src/metabase/models/collection.clj
qnkhuat: Cool, I removed the simple hydration here [ed291d8](https://github.com/metabase/metabase/pull/34827/commits/ed291d859d82d81b071cd2f9770fcbcc9baef458)

test/metabase/api/activity_test.clj
calherries: Makes sense, thanks

calherries (approved):
Looks good

src/metabase/models/collection.clj
```diff
@@ -1150,7 +1150,10 @@
   [collections]
   (if (= 1 (count collections))
     (let [collection (first collections)]
-      [(assoc collection :is_personal (is-personal-collection-or-descendant-of-one? collection))])
+      (if (some? collection)
+        [(assoc collection :is_personal (is-personal-collection-or-descendant-of-one? collection))]
+        ;; root collection is nil
+        [collection]))
```
calherries, line 1156: You should also handle the case where there is a root collection in `collections` and `collections` has more than one element. And add a test

src/metabase/models/collection.clj
```diff
@@ -673,14 +673,14 @@
...
-    ;; Collection.)
-    (t2/exists? Collection
-      :id                (first (location-path->ids (:location collection)))
-      :personal_owner_id [:not= nil]))))
+     ;; If collection has an owner ID we're already done here, we know it's a Personal Collection
+     (:personal_owner_id collection)
+     ;; Otherwise try to get the ID of its highest-level ancestor, e.g. if `location` is `/1/2/3/` we would get `1`.
+     ;; Then see if the root-level ancestor is a Personal Collection (Personal Collections can only got in the Root
+     ;; Collection.)
+     (t2/exists? Collection
```
calherries, line 681: This was correct before
```suggestion
    ;; If collection has an owner ID we're already done here, we know it's a Personal Collection
    (:personal_owner_id collection)
    ;; Otherwise try to get the ID of its highest-level ancestor, e.g. if `location` is `/1/2/3/` we would get `1`.
    ;; Then see if the root-level ancestor is a Personal Collection (Personal Collections can only got in the Root
    ;; Collection.)
    (t2/exists? Collection
```

calherries (changes requested):

src/metabase/models/collection.clj
qnkhuat: did that in https://github.com/metabase/metabase/pull/34827/commits/e31da20b1aee4fcd0ec93ca69774484032ca7d8b

calherries (approved):

PR #34834, opened 2023-10-19
[move dashcard to another tab] move tests to e2e

npretto:
### Description

This moves some tests to e2e as the unit ones were using redux internal state which is not recommended.

It also adds a test to check if the move functionality works for different card types.
I didn't add a action card because setting it up would involve a lot more work (setting up pg etc) and it may not be worth

Note: e2e are probably failing because of https://metaboat.slack.com/archives/C5XHN8GLW/p1697729886291229


e2e/support/helpers/e2e-dashboard-helpers.js
```diff
@@ -248,6 +248,66 @@ export function createTextCard({
   };
 }
 
+export function createHeadingCardDTO({
```
npretto, line 251: I added "DTO" to the name as similar helper functions existed that used the GUI to create the cards

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -82,9 +89,86 @@ describe("scenarios > dashboard > tabs", () => {
...
+    });
+
+    cy.findByRole("tab", { name: "Tab 1" }).click();
+
+    cy.log("should stay on the same tab");
+    cy.findByRole("tab", { selected: true }).should("have.text", "Tab 1");
+
+    const originalSize = {};
+    getDashboardCard(0).then(element => {
+      originalSize.width = element.outerWidth();
```
npretto, line 104: setting values from a `.then` looks very ugly to me but it's what's suggested by cypress: https://docs.cypress.io/api/commands/should#Compare-text-values-of-two-elements

e2e/support/helpers/e2e-ui-elements-helpers.js
```diff
@@ -10,6 +10,10 @@ export function main() {
   return cy.get("main");
 }
 
+export function menu() {
+  return cy.get("[role='menu']");
```
ranquild, line 14: please use `cy.getByRole("menu")`

e2e/support/helpers/e2e-dashboard-helpers.js
ranquild: We don't use DTOs suffixes. If it was used somewhere it wasn't intended

e2e/test/scenarios/dashboard/tabs.cy.spec.js
ranquild: Could you use `cy.wrap(...).as(alias);` to save the size and `cy.get("@alias")` to get it back later? We use this approach for storing values in tests

ranquild (commented):
Small comments

e2e/support/helpers/e2e-dashboard-helpers.js
npretto: Do we have a naming convention/strategy to distinguish creating things with the UI:
https://github.com/metabase/metabase/blob/b55a4070e4b271baa8ad89ecdfec6722e419eaf9/e2e/support/helpers/e2e-dashboard-helpers.js#L192-L194

from creating just the data to send to the API:

https://github.com/metabase/metabase/blob/b55a4070e4b271baa8ad89ecdfec6722e419eaf9/e2e/support/helpers/e2e-dashboard-helpers.js#L280-L309

?

e2e/test/scenarios/dashboard/tabs.cy.spec.js
npretto: Thanks, this looks better than

e2e/support/helpers/e2e-dashboard-helpers.js
ranquild: No, we don't

ranquild: Actually, we have. We use `details` prefix sometimes.

ranquild (approved):

PR #34837, opened 2023-10-19
Return correct values for linked filters using non-equality filters

### LINKED ISSUE #34039: Linked Dashboard Filters Connected to the Same Field Don't Respect `Is Not` Setting
### Describe the bug

*Edit: This is still an issue even if the two Field Filters do not point to the same field_id.*

I have two "Text or Category" Dashboard Field Filters connected to the same field (same `field_id`), but one of the Dashboard Filters is `Is Not` while the other is set to `Is`. I linked these two filters together. When I select values in the `Is Not` filter, I was expecting those values to not show up as a drop-down item in the `Is` filter. Instead, I'm _only_ seeing the values selected in the `Is Not` filter.

To be clear, there are two template field filters in the cards' SQL query, `{{ this_field }}` and `{{ exclude_this_field }}`, for example. These are both associated with the same field.

The `{{ this_field }}` filter is set to "String" (`string/=`) in the card's configuration while the `{{ exclude_this_field }}` is set to "String is not" (`string/!=`). 

### To Reproduce

1. Create a dashboard with a single SQL-query card. Make sure the SQL query card has a `where 1=1` or `where true` in it so the subsequent steps work properly
2. On the card, create a `[[ AND {{ this_field }} ]]` Field Filter of type "String" (`string/=` on the backend) to a Field in the table that is of type Category
3. On the card, create a `[[ AND {{ exclude_this_field }} ]]` Field Filter associated to the same field as step 2's filter, but set the type to "String is not" (`string/!=` on the backend)
4. On the dashboard, create a "Text or Category" filter set to `Is` and associate this to the `{{ this_field }}` card filter.
5. On the dashboard, create a "Text or Category" filter set to `Is not` and associate this to the `{{ exclude_this_field }}` card filter.
6. Link these two filters together (bidirectionally)
7. Make a drop down selection in the `Is not` Dashboard filter
8. Click the `Is` Dashboard filter and inspect the values included in the drop down. You'll see _only_ the value(s) that were selected in the `Is not` filter. **This is the issue**.

### Expected behavior

I was expecting the value(s) selected in the `Is not` Dashboard filter to not be present when I click into the `Is` Dashboard filter

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/117.0",
    "vendor": ""
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.20+8",
    "java.version": "11.0.20",
    "java.vm.version": "11.0.20+8",
    "os.name": "Linux",
    "os.version": "5.10.184-175.731.amzn2.x86_64",
    "user.language": "en",
    "user.timezone": "GMT"
  },
  "metabase-info": {
    "databases": [
      "snowflake"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "14.7"
      },
      "jdbc-driver": {
        "version": "42.5.1"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2023-08-15",
      "tag": "v1.46.8",
      "branch": "release-x.46.x",
      "hash": "27a9ec6"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

Somewhere between "annoying" and "blocking some users". The field values that should be excluded will be provided by the application we're embedding into and could be a security issue if they aren't hidden properly.

### Additional context

I'm unsure if this is able to be accomplished some other way and wasn't able to find any bug tickets that matched this issue. Thanks!

sansolovyov:
`api.dashboard/chain-filter-constraints` now generates a list of constraints using a filter type defined (like
`:string/!=` or `:number/between`). `chain-filter/filter-clause` down the call chain has no access
to _dashboard_ data, but accepts operators like `:!=` and `:between` readily.

Resolves #34039


calherries (approved):

calherries (dismissed):
Didn't mean to approve

src/metabase/api/dashboard.clj
```diff
@@ -800,21 +800,23 @@
...
+    (get resolved-params param-key)))
 
 (defn- chain-filter-constraints [dashboard constraint-param-key->value]
   (into {} (for [[param-key value] constraint-param-key->value
-                 field-id          (param-key->field-ids dashboard param-key)]
-             [field-id value])))
+                 :let              [param (param-key->param dashboard param-key)]
+                 field-id          (mappings->field-ids (:mappings param))]
+             (case (:type param)
+               :string/!= [field-id [:!= value]]
```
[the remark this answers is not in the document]
piranha: Yeah. The problem is that I'm still not sure that's the best place to handle this stuff... Plus all those `:string/starts-with` do not seem to generate correct SQL when used in a card, so I'm wondering if there is larger-scale problem.

I should've written those thoughts directly in the comment, I guess, rather than ponder on them myself. :-)

piranha:
Right now it handles all cases. I'm not happy with defining operation as `(keyword (name param-type))`, it's something I borrowed from `common.parameters.operators`.

Some more twiddling was required to get options for the operation, so that case insensitivity could be handled, and I'm not happy with `mapping->fields` name, so I'm taking suggestions if there are any. Or maybe there is a common way to do that, I'd be happy to remove some code.

Format of `{field-id [op values... options]}` is quite extensively tested in `metabase.models.params.chain-filter-test`.

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +785,50 @@
...
+                    options  (:options ttag)
                     field-id (or
-                              ;; Get the field id from the field-clause if it contains it. This is the common case for
-                              ;; mbql queries.
-                              (mbql.u/match-one field-clause [:field (id :guard integer?) _] id)
-                              ;; Attempt to get the field clause from the model metadata corresponding to the field.
-                              ;; This is the common case for native queries in which mappings from original columns
-                              ;; have been performed using model metadata.
-                              (:id (qp.util/field->field-info field-clause result_metadata)))]
+                               ;; Get the field id from the field-clause if it contains it. This is the common case for
```
qnkhuat, line 798: the old indentation is correct https://guide.clojure.style/#one-space-indent

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +785,50 @@
...
+                                    param-type (:type param)
+                                    op         (if (qualified-keyword? param-type)
+                                                 (keyword (name param-type))
+                                                 :=)]
+                 field             (mappings->fields (:mappings param))]
+             [(:field-id field)
+              (cond-> [op]
+                (vector? value)        (into value)
+                (not (vector? value))  (conj value)
+                (seq (:options field)) (conj (:options field)))])))
```
qnkhuat, line 831: can you show me in which setup the field has an option?

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +785,50 @@
...
   (into {} (for [[param-key value] constraint-param-key->value
-                 field-id          (param-key->field-ids dashboard param-key)]
-             [field-id value])))
+                 :let              [param      (param-key->param dashboard param-key)
+                                    param-type (:type param)
+                                    op         (if (qualified-keyword? param-type)
+                                                 (keyword (name param-type))
+                                                 :=)]
+                 field             (mappings->fields (:mappings param))]
+             [(:field-id field)
```
qnkhuat, line 827: I don't think I like this data structure for constraints:
First, its length is not defined thus making it hard to destructure and understand what's inside. I'd rather see sth with this shape `[op value(s) option]`, this is guaranteed to have 3 parts and I know exactly what's what.

Second, it seems like the constraint here is a filter clause without the field because we're now sending both ops and value to `metabase.models.params.chain-filter/filter-clause`. reading [this comment](https://github.com/metabase/metabase/pull/34837/files#diff-1b120838a9ef31567ead6ca514337d29f0dc2da9428172057f4670624f53866aR123-R124) I think it's not the "official" API for the function, it's like someone tried to make a hack to fix some bugs in the past and sneak it in there.

the problem with that is it makes it hard to know what `value` is in the context of `filter-clause` function.

how about we restructure this even further and make constriant now have this schema.

```
[:map-of pos-int? [:map {:closed true}
                     [:value    :any]
                     [:op         [:enum := :=1! :between ....]]
                     [:options :map]]
```

with this we can update `filter-clause` to appropriately destructure then construct the filter clause.

qnkhuat (changes requested):
I think the schema for constraint can be more strict, see this [comment](https://github.com/metabase/metabase/pull/34837/files#r1369617716)

And we need to add more tests for this, some suggestions:
- a test with a full setup like the issue this PR trying to fix
- More tests in `metabase.models.params.chain-filter-test` to test behavior of chain-filter

src/metabase/api/dashboard.clj
qnkhuat: I'm not sure why having it as a list is better than a map of `field-id => constriantMap`.

src/metabase/models/params.clj
```diff
@@ -100,25 +100,25 @@
...
     (when (mbql.u/is-clause? :dimension target)
       (let [[_ dimension] target]
         (try
-          (unwrap-field-or-expression-clause
-           (if (mbql.u/is-clause? :template-tag dimension)
-             (template-tag->field-form dimension card)
-             dimension))
+          (if (mbql.u/is-clause? :template-tag dimension)
+            (find-template-tag dimension card)
+            {:dimension    (unwrap-field-or-expression-clause dimension)})
```
qnkhuat, line 121: this is misleading because not all `templateTag` has a dimension, it's only applied for fieldFilter.

See this schema for all types of template tag https://github.com/metabase/metabase/blob/ngoc-optimize-revision/shared/src/metabase/mbql/schema.cljc#L1207

src/metabase/api/dashboard.clj
piranha: Well, for one it can have a few constraints on the same field without any weird dances. Plus map is a type where you'd be retrieving something from, generally, rather just iterating over - so it feels like a list/vector is a bit more honest about the intention.

src/metabase/models/params.clj
piranha: Misleading as in a function name? Or as test inside of `if`?

qnkhuat: yea, the function name say it returns a template tag, but a map with `:dimension` key is not a template tag. it's not even satisfied as one of type of template tag.

src/metabase/models/params.clj
```diff
@@ -100,25 +100,31 @@
...
   "Fetch the `:field` clause from `dashcard` referenced by `template-tag`.
 
-    (template-tag->field-form [:template-tag :company] some-dashcard) ; -> [:field 100 nil]"
+    (find-template-tag [:template-tag :company] some-dashcard) ; -> [:field 100 nil]"
   [[_ tag] card]
-  (get-in card [:dataset_query :native :template-tags (u/qualified-name tag) :dimension]))
+  (get-in card [:dataset_query :native :template-tags (u/qualified-name tag)]))
 
-(mu/defn param-target->field-clause :- [:maybe mbql.s/Field]
+(mu/defn param-target->template-tag :- [:maybe mbql.s/TemplateTag]
```
qnkhuat, line 110: I don't think we should change this function to make it return a template tag.
As I said, there are multiple types of template tags, and a template tag with `:dimension` is one of them, and in product terms, it's referred to "field filter"

it makes the usage of this function confusing. takes [this](https://github.com/metabase/metabase/blob/34db49580ff698a4cf5385bf2ffebade44a1fb99/src/metabase/api/card.clj#L1099) for example, under that context it's not clear why we call `:dimension` on a template when we don't know for sure it's a FieldFilter template tag.

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +785,49 @@
...
 
-(mu/defn ^:private mappings->field-ids :- [:maybe [:set ms/PositiveInt]]
+(mu/defn ^:private mappings->fields :- [:maybe [:set [:map
+                                                         [:field-id ms/PositiveInt]
+                                                         [:options [:maybe map?]]]]]
   [parameter-mappings :- [:maybe [:or [:set dashboard-card/ParamMapping] [:sequential dashboard-card/ParamMapping]]]]
   (set (for [{{:keys [card]} :dashcard :keys [target]} parameter-mappings
-             :let  [field-clause (params/param-target->field-clause target card)]
-             :when field-clause
+             :let  [ttag (params/param-target->template-tag target card)]
```
qnkhuat, line 793: I'm confused why this is a `template-tag`, if I have a parameter that connects to a field on a MBQL query, it's not a template tag.

src/metabase/api/dashboard.clj
```diff
@@ -829,7 +841,8 @@
     constraint-param-key->value :- ms/Map
     query                       :- [:maybe ms/NonBlankString]]
    (let [constraints (chain-filter-constraints dashboard constraint-param-key->value)
-         field-ids   (param-key->field-ids dashboard param-key)]
+         param       (param-key->param dashboard param-key)
```
qnkhuat, line 844: these 2 functions call `t2/hydrate dashboard :resolved-params)` twice which is redundant, can we somehow reuse it?

qnkhuat (commented):
I know the parameter is a mess so takes time to understand it.

I suggest you take some time to internalize the notion of:
- parameter on dashboards, how it connects to a: template tag on native card, a field on MBQL card
- parameter on native cards (aka template tags and all of its type)
  - as a bonus, learn what's snippet query
- parameter mappings on dashboard and how they work

Also, the [mbql schema ](https://github.com/metabase/metabase/blob/34039-fix-is-not-filter/shared/src/metabase/mbql/schema.cljc)is a good resource to learn about the schema of all these things too.

And I'm happy to pair with you to explore this if you want too, just give me a ping.

src/metabase/models/params.clj
piranha: Hmmm... right, but if I just return a field clause, then there is no way to know any options defined on it. Or are you suggesting to return "something", like `{:field ... :options ...}`? I was trying to keep it closer to an already defined entities.

src/metabase/api/dashboard.clj
qnkhuat: I see, I'm not totally against using list, but since the constraints are from `query-params` from the API, and it's already a map so I feel it a bit more natural to keep it that way.

src/metabase/models/params.clj
qnkhuat: how about we not use it at all? I think we can re-think the current `api.dashboard/mappings->field-ids` function, 

I don't think it needs to call this function, I think what it's trying to do is :
- if the card is native => get the template tag field id that the template tag map to
- if the card is mbql => get the id from `result_metadata`

maybe we can simplified this for our own needs.

src/metabase/api/dashboard.clj
piranha: Do you here under 'redundant' mean that they call it far too many times? It's also done in a loop inside of `chain-filter-constraints`, but I thought it's almost a noop when it's already hydrated (just checked, it takes ~53 µs for me)?

shared/src/metabase/mbql/util.cljc
```diff
@@ -52,6 +52,12 @@
      ((set k-or-ks) (first x))
      (= k-or-ks (first x)))))
 
+(defn get-clause
```
qnkhuat, line 55: the name makes me a bit confused, if `x` is a clause, then `(rest x)` must be sth else right?

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +786,63 @@
...
-                              ;; Attempt to get the field clause from the model metadata corresponding to the field.
-                              ;; This is the common case for native queries in which mappings from original columns
-                              ;; have been performed using model metadata.
-                              (:id (qp.util/field->field-info field-clause result_metadata)))]
+  (set (for [{:keys [target] {:keys [card]} :dashcard} parameter-mappings
+             :let  [dimension (->> (mbql.normalize/normalize-tokens target :ignore-path)
+                                   (mbql.u/get-clause :dimension)
+                                   first)]
+             :when dimension
+             :let  [ttag      (find-template-tag dimension card)
```
qnkhuat, line 806: can you show me the e2e test that makes template tag has options? 
if it's not possible on the UI then let's just get rid of this so this code is less confusing. that's way we can make this to function to return a set of ids

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +786,63 @@
...
-(mu/defn ^:private mappings->field-ids :- [:maybe [:set ms/PositiveInt]]
+(defn- find-template-tag
+  "Fetch the `:field` clause from `dashcard` referenced by `template-tag`.
+
+    (find-template-tag [:template-tag :company] some-dashcard) ; -> [:field 100 nil]"
+  [dimension card]
+  (when-let [tag (first (mbql.u/get-clause :template-tag dimension))]
+    (get-in card [:dataset_query :native :template-tags (u/qualified-name tag)])))
+
+(mu/defn ^:private mappings->fields :- [:maybe [:set [:map
```
qnkhuat, line 797: idea: instead of taking a `mappings`, how about we make this function taking a `param`, so that it can determine the `op` as well?

with this I think `chain-filter-constraints` can greatly be simplified, thinking about the structure of that function will be

```clojure
(for [[param-key value] constraint-param-key->value
      field             (param->fields (param-key->param dashboard param-key))]
  (merge field {:value value}))
```

src/metabase/api/dashboard.clj
qnkhuat: > Do you here under 'redundant' mean that they call it far too many times?

yes, but if `dashboard` is not already hydrated when we called `chain-filter`, then it'll be hydrated multiple times in `param-key->param` and `chain-filter-constraints`

especially `chain-filter-constraints` is hydrating in a loop

shared/src/metabase/mbql/util.cljc
piranha: Hmm... well, it feels like a good name to me, i.e. if you "get" some clause, you get the body of it, not the identifier. Could be `get-clause-body`, but that makes it less succinct.

src/metabase/api/dashboard.clj
piranha: Yeah after looking at it it seems like the best way is to bring hydration up to a `chain-filter` itself.

src/metabase/api/dashboard.clj
piranha: Ahhh that's a great idea, plus the type becomes so much simpler.

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +786,51 @@
...
+  [{:keys [mappings] :as param} :- mbql.s/Parameter]
+  (for [{:keys [target] {:keys [card]} :dashcard} mappings
+        :let  [dimension (->> (mbql.normalize/normalize-tokens target :ignore-path)
+                              (mbql.u/get-clause :dimension)
+                              first)]
+        :when dimension
+        :let  [ttag      (find-template-tag dimension card)
+               dimension (if ttag
+                           (:dimension ttag)
+                           dimension)
```
qnkhuat, line 812: is there a clearer way to write this part? the switching back and forth between `dimension` and tag makes it confusing.

in case `ttag` is nil, we returned `dimension` as `dimension` from the previous state on line 808.
It's not clear to me why we do this.

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +786,51 @@
...
-             [field-id value])))
+(defn- find-template-tag
+  "Fetch the `:field` clause from `dashcard` referenced by `template-tag`.
+
+    (find-template-tag [:template-tag :company] some-dashcard) ; -> [:field 100 nil]"
+  [dimension card]
+  (when-let [tag (first (mbql.u/get-clause :template-tag dimension))]
+    (get-in card [:dataset_query :native :template-tags (u/qualified-name tag)])))
+
+(defn- param->op [{:keys [type]}]
```
qnkhuat, line 797: this function is only used once and it's quite trivial IMO, so how about we make it inline to make avoid jumping back and forth when reading `param->fields`?

if not then I suggest we rename this to `param-type->op`, it's simple enough so we don't have to jump to it.

src/metabase/api/dashboard.clj
piranha: Well, if dimension is resolved to a `:template-tag`, we want to get the dimension from that tag. But if not, then it could be a field already (if I'm understanding [tests](https://github.com/metabase/metabase/blob/34039-fix-is-not-filter/test/metabase/api/dashboard_test.clj#L485) correctly).

src/metabase/api/dashboard.clj
piranha: Okay, `param-type->op` is nice. I extract it since while it's small, it adds to the noise in the function and ideally I'd love to see a more structured way to get from a parameter type to an operator: relying on keyword names feels somewhat error-prone to me.

src/metabase/models/params/chain_filter.clj
```diff
@@ -115,26 +115,21 @@
 
 (defn- filter-clause
```
qnkhuat, line 116: I think we should use the `Constraint` schema fro this function as well.

src/metabase/models/params/chain_filter.clj
```diff
@@ -115,26 +115,21 @@
...
   "Generate a single MBQL `:filter` clause for a Field and `value` (or multiple values, if `value` is a collection)."
-  [source-table-id field-id value]
+  [source-table-id {:keys [field-id op value options]
+                    :or   {op :=}}]
```
qnkhuat, line 119: I don't think op need a default at this level right?

piranha: Well... I've put it as optional in schema, since there are other users for `chain-filter`, like [filterable-field-ids](https://github.com/metabase/metabase/blob/34039-fix-is-not-filter/src/metabase/models/params/chain_filter.clj#L698)

src/metabase/models/params/chain_filter.clj
```diff
@@ -115,26 +115,21 @@
...
+    (if (and (temporal-field? field-id)
+             (string? value))
+      (u/ignore-exceptions
+        (params.dates/date-string->filter value field-id))
+      (cond-> [op field-clause]
+        true          (into (if (or (sequential? value)
+                                    (set? value))
+                              value
+                              [value]))
+        (seq options) (conj options)))))
```
qnkhuat, line 132: if this has an options, then the query will be sth like `[:!= [:field 443 nil] "American" {:case-sensitive true}]`

which is a invalid mbql.

I don' think this is how options is used

src/metabase/models/params/chain_filter.clj
```diff
@@ -115,26 +115,21 @@
...
-          [:= field-clause value]))))
+    (if (and (temporal-field? field-id)
+             (string? value))
+      (u/ignore-exceptions
+        (params.dates/date-string->filter value field-id))
+      (cond-> [op field-clause]
+        true          (into (if (or (sequential? value)
+                                    (set? value))
+                              value
+                              [value]))
```
qnkhuat, line 131:
```suggestion
      [op field-clause (u/one-or-many value)]
```

this can be replaced with one of our helper funciton.

test/metabase/api/dashboard_test.clj
```diff
@@ -294,12 +295,22 @@
                   (update-in [:last-edit-info :timestamp] boolean)))))
 
     (testing "f=all shouldn't return archived dashboards"
-      (is (= #{rasta-dash crowberto-dash}
-             (set (map :id (mt/user-http-request :crowberto :get 200 "dashboard" :f "all")))))
+      (is (set/subset?
```
qnkhuat, line 298: Thanks!

src/metabase/models/params/chain_filter.clj
piranha: Looking at schema, this is exactly how it's used:

```
metabase.mbql.schema> (mc/validate Filter [:contains [:field 443 nil] "American" {:case-sensitive true}])
true
```

I've changed test a little bit (locally, did not push not to trigger CI) so this can be demonstrated:

```
(testing "contains is case insensitive"
          (mt/let-url [url (chain-filter-values-url dashboard (:category-name param-keys)
                                                    (:not-category-name param-keys) "American"
                                                    (:category-contains param-keys) "am")]
            (is (= {:values          [["Latin American"] ["Ramen"]]
                    :has_more_values false}
                   (chain-filter-test/take-n-values 3 (mt/user-http-request :rasta :get 200 url))))))
                   ```

File: src/metabase/api/dashboard.clj (hunk 12 above)
Comment (qnkhuat) [reply src/metabase/api/dashboard.clj:812]: hmmm, yea but it's still very confusing when you read this, like it's hard understand "why" we need to do that.

maybe we need to rename `find-template-tag` to `get-template-tag-if-exists` to make it more explicit, and can you add the explanation above as a comment?

File: src/metabase/models/params/chain_filter.clj (hunk 15 above)
Comment (qnkhuat) [reply src/metabase/models/params/chain_filter.clj:119]: how about we specify the `:op` there other than making an assumption here?

File: src/metabase/api/dashboard.clj (hunk 12 above)
Comment (piranha) [reply src/metabase/api/dashboard.clj:812]: How do you feel about `maybe-template-tag`? 😁

Comment (piranha) [reply src/metabase/api/dashboard.clj:812]: I wonder if it's better to dispatch on its type, like 
```
(cond 
  (mbql.u/is-clause? :template-tag dimension) (:dimension ttag)
  (mbq.u/is-clause? :field dimension) dimension
  :else (throw (ex-info "no idea" {:dimension dimension}))))
```

File: src/metabase/models/params/chain_filter.clj (hunk 15 above)
Comment (piranha) [reply src/metabase/models/params/chain_filter.clj:119]: yeah why not, just wanted to keep changes to minimum :)

File: src/metabase/api/dashboard.clj (hunk 12 above)
Comment (qnkhuat) [reply src/metabase/api/dashboard.clj:812]: that's definitely better

File: src/metabase/models/params/chain_filter.clj (hunk 16 above)
Comment (qnkhuat) [reply src/metabase/models/params/chain_filter.clj:132]: oh okay, it was not invalid because I used `:!=` :D

Comment (qnkhuat) [reply src/metabase/models/params/chain_filter.clj:132]: hmm, but it's still possible to generate an invalid clause tho... what if options is sth else? the options is currently `(merge (:options ttag) (:options field))` so I have no idea what i could be

how about we make the options strictly to be only `:case-sensitive`?

Comment (piranha) [reply src/metabase/models/params/chain_filter.clj:132]: Maybe it's better to leave validation to an MBQL layer so that it's not specified multiple times over the codebase? I guess there is a point in catching in earlier, but OTOH the level of specificity of MBQL schema is really high, so IDK if there is a necessity to put any effort in a much simpler schema?..

Adding any other option would be a pain, since no one will remember you have to add new option to chain filters, and then it won't work. IDK. 🤣

File: src/metabase/api/dashboard.clj (hunk 12 above)
Comment (piranha) [reply src/metabase/api/dashboard.clj]: done in 39ec5a256ca0cd926f3cbad9232029374e1191fe

File: src/metabase/models/params/chain_filter.clj (hunk 14 above)
Comment (piranha) [reply src/metabase/models/params/chain_filter.clj]: done in 39ec5a256ca0cd926f3cbad9232029374e1191fe

File: src/metabase/models/params/chain_filter.clj (hunk 15 above)
Comment (piranha) [reply src/metabase/models/params/chain_filter.clj]: done in 39ec5a256ca0cd926f3cbad9232029374e1191fe

File: src/metabase/models/params/chain_filter.clj (hunk 17 above)
Comment (piranha) [reply src/metabase/models/params/chain_filter.clj]: done in 39ec5a256ca0cd926f3cbad9232029374e1191fe (though it's not as elegant as just calling that function because of nils)

File: src/metabase/api/dashboard.clj (hunk 13 above)
Comment (piranha) [reply src/metabase/api/dashboard.clj]: done in 39ec5a256ca0cd926f3cbad9232029374e1191fe

File: src/metabase/models/params/chain_filter.clj [hunk 19]
```diff
@@ -93,6 +93,19 @@
...
+   [:field-id ms/PositiveInt]
+   [:op :keyword]
+   [:value :any]
+   [:options {:optional true} [:maybe map?]]])
+
+(def Constraints
+  "Schema for a list of Constraints."
+  [:sequential Constraint])
+
+
```
qnkhuat, line 108:
```suggestion
```

shared/src/metabase/mbql/util.cljc
qnkhuat: `get-clause-body` makes more sense to me. I think it's crucial to get the name for utils function, because most of the time we shouldn't look into its implementation to know what it does.

qnkhuat: and it'll be better if you write an example in the docstring

```
(get-clause [:field 1 nil] :field) => [1 nil])
```

src/metabase/api/dashboard.clj
```diff
@@ -784,38 +786,50 @@
...
-  {:pre [(string? param-key)]}
-  (let [{:keys [resolved-params]} (t2/hydrate dashboard :resolved-params)
-        param                     (get resolved-params param-key)]
-    (mappings->field-ids (:mappings param))))
-
-(defn- chain-filter-constraints [dashboard constraint-param-key->value]
-  (into {} (for [[param-key value] constraint-param-key->value
-                 field-id          (param-key->field-ids dashboard param-key)]
-             [field-id value])))
+(defn- get-template-tag
```
qnkhuat, line 789: if we don't use template tag for anything, maybe we can makes it returns the :dimension so we don't have to destruct it later on?
```suggestion
(defn- get-template-tag-dimension
```

src/metabase/models/params/chain_filter.clj
```diff
@@ -113,28 +125,21 @@
...
-            (u/ignore-exceptions
-              (params.dates/date-string->filter value field-id)))
-          ;; e.g. {$$venues.price 2} -> [:= $$venues.price 2]
-          [:= field-clause value]))))
+    (if (and (temporal-field? field-id)
+             (string? value))
+      (u/ignore-exceptions
+        (params.dates/date-string->filter value field-id))
+      (cond-> [op field-clause]
+        ;; we don't want to skip our value, even if its nil
```
qnkhuat, line 140: can you clarify wdym about this? do you mean when the value is set as `nil`?

if so do we have a test?

qnkhuat (approved):
Have some nits but I think this looks good now!

src/metabase/models/params/chain_filter.clj
piranha: `(u/one-or-many nil)` returns `nil`, and `(into [op field] nil)` is `[op field]`, while we need `[op field nil]` (with the value).

And yes, tests fail a lot on that one. :)

shared/src/metabase/mbql/util.cljc
piranha: Haha, clearly you feel strongly about that one. But on the other hand `get` is not called `get-value`, while it gets the value, right?

I'd look at this as a part of lingo, in _any_ case, you'll have to understand what exactly the function does to your data. And given function similarity to `is-clause?`, I really like the name. :)

TBH now that I think about it, I really love the idea of something like this:

```
(defn check-clause [k-or-ks x]
  (when (is-clause? k-or-ks x) x))
```

What do you think?

src/metabase/api/dashboard.clj
piranha: I've spent considerable amount of time yesterday trying to understand why I was getting options from a template tag, if it's not necessary now — and all tests pass. 

Then I began writing this comment and realized that I haven't checked the UI, and voila! — options come from the template tag there and not from the parameter. I'm not quite understanding the difference between tests and real UI here, so need some time to figure it out to write a correct test.

qnkhuat: can you tell me how to setup a template tag with options?

piranha: Not programatically yet, but if you set up a dashboard, `param` (incoming in `param->fields`) will look like this:

<details>
<summary>param</summary>
<pre><code>
{:filteringParameters ["fadff97c" "f57f3a34" "864d5df1"],
 :id "38b52427",
 :mappings #{{:card_id 1,
              :dashcard (toucan2.instance/instance
                         :model/DashboardCard
                         {:action_id nil,
                          :card (toucan2.instance/instance
                                 :model/Card
                                 {:archived false,
                                  :cache_ttl nil,
                                  :collection_id nil,
                                  :collection_position nil,
                                  :collection_preview true,
                                  :created_at #t "2023-10-19T08:43:47.374840Z",
                                  :creator_id 1,
                                  :database_id 2,
                                  :dataset false,
                                  :dataset_query {:database 2,
                                                  :native {:query "select * from test\nwhere 1=1\n[[and {{ this_field }} ]]\n[[and {{ exclude_this_field }} ]]\n[[and {{ a_starts_with }} ]]\n[[and {{ amount_between }} ]]",
                                                           :template-tags {"a_starts_with" {:dimension [:field 138 nil],
                                                                                            :display-name "A Starts With",
                                                                                            :id "bb54a514-99e3-4de7-a139-aae7729b7ff9",
                                                                                            :name "a_starts_with",
                                                                                            :options {:case-sensitive false},
                                                                                            :type :dimension,
                                                                                            :widget-type :string/contains},
                                                                           "amount_between" {:dimension [:field 114 nil],
                                                                                             :display-name "Amount Between",
                                                                                             :id "a92931fe-43ec-499e-8072-e34a93f4f260",
                                                                                             :name "amount_between",
                                                                                             :type :dimension,
                                                                                             :widget-type :number/between},
                                                                           "exclude_this_field" {:dimension [:field
                                                                                                             137
                                                                                                             nil],
                                                                                                 :display-name "Exclude This Field",
                                                                                                 :id "7b95074e-25eb-493a-8f0a-22dcc44e09e4",
                                                                                                 :name "exclude_this_field",
                                                                                                 :type :dimension,
                                                                                                 :widget-type :string/!=},
                                                                           "this_field" {:default nil,
                                                                                         :dimension [:field 138 nil],
                                                                                         :display-name "This Field",
                                                                                         :id "7e9d5188-b81f-4f58-8e27-a2a0ff3c681f",
                                                                                         :name "this_field",
                                                                                         :type :dimension,
                                                                                         :widget-type :string/=}}},
                                                  :type :native},
                                  :description nil,
                                  :display :table,
                                  :embedding_params nil,
                                  :enable_embedding false,
                                  :entity_id "sdQYFZgioanVovTzqcAv0",
                                  :id 1,
                                  :made_public_by_id nil,
                                  :metabase_version "v0.48.0-SNAPSHOT (dc676e1)",
                                  :name "ques 34039",
                                  :parameter_mappings [],
                                  :parameters [{:id "7e9d5188-b81f-4f58-8e27-a2a0ff3c681f",
                                                :name "This Field",
                                                :slug "this_field",
                                                :target [:dimension [:template-tag "this_field"]],
                                                :type :string/=}
                                               {:id "7b95074e-25eb-493a-8f0a-22dcc44e09e4",
                                                :name "Exclude This Field",
                                                :slug "exclude_this_field",
                                                :target [:dimension [:template-tag "exclude_this_field"]],
                                                :type :string/!=}
                                               {:id "bb54a514-99e3-4de7-a139-aae7729b7ff9",
                                                :name "A Starts With",
                                                :options {:case-sensitive false},
                                                :slug "a_starts_with",
                                                :target [:dimension [:template-tag "a_starts_with"]],
                                                :type :string/contains}
                                               {:id "a92931fe-43ec-499e-8072-e34a93f4f260",
                                                :name "Amount Between",
                                                :slug "amount_between",
                                                :target [:dimension [:template-tag "amount_between"]],
                                                :type :number/between}],
                                  :public_uuid nil,
                                  :query_type :native,
                                  :result_metadata ({:base_type :type/Integer,
                                                     :display_name "id",
                                                     :effective_type :type/Integer,
                                                     :field_ref [:field "id" {:base-type :type/Integer}],
                                                     :fingerprint nil,
                                                     :name "id",
                                                     :semantic_type :type/PK}
                                                    {:base_type :type/Integer,
                                                     :display_name "amount",
                                                     :effective_type :type/Integer,
                                                     :field_ref [:field "amount" {:base-type :type/Integer}],
                                                     :fingerprint {:global {:distinct-count 2, :nil% 0.0},
                                                                   :type {:type/Number {:avg 2.5,
                                                                                        :max 3.0,
                                                                                        :min 2.0,
                                                                                        :q1 2.0,
                                                                                        :q3 3.0,
                                                                                        :sd 0.7071067811865476}}},
                                                     :name "amount",
                                                     :semantic_type nil}
                                                    {:base_type :type/Boolean,
                                                     :display_name "is_checked",
                                                     :effective_type :type/Boolean,
                                                     :field_ref [:field "is_checked" {:base-type :type/Boolean}],
                                                     :fingerprint {:global {:distinct-count 2, :nil% 0.0}},
                                                     :name "is_checked",
                                                     :semantic_type nil}
                                                    {:base_type :type/Text,
                                                     :display_name "a",
                                                     :effective_type :type/Text,
                                                     :field_ref [:field "a" {:base-type :type/Text}],
                                                     :fingerprint {:global {:distinct-count 2, :nil% 0.0},
                                                                   :type {:type/Text {:average-length 3.0,
                                                                                      :percent-email 0.0,
                                                                                      :percent-json 0.0,
                                                                                      :percent-state 0.0,
                                                                                      :percent-url 0.0}}},
                                                     :name "a",
                                                     :semantic_type nil}
                                                    {:base_type :type/Text,
                                                     :display_name "b",
                                                     :effective_type :type/Text,
                                                     :field_ref [:field "b" {:base-type :type/Text}],
                                                     :fingerprint {:global {:distinct-count 2, :nil% 0.0},
                                                                   :type {:type/Text {:average-length 3.0,
                                                                                      :percent-email 0.0,
                                                                                      :percent-json 0.0,
                                                                                      :percent-state 0.0,
                                                                                      :percent-url 0.0}}},
                                                     :name "b",
                                                     :semantic_type nil}),
                                  :table_id nil,
                                  :updated_at #t "2023-10-31T07:06:49.144983Z",
                                  :visualization_settings {:table.cell_column "amount", :table.pivot_column "id"}}),
                          :card_id 1,
                          :col 0,
                          :collection_authority_level nil,
                          :created_at #t "2023-10-19T08:44:14.822842Z",
                          :dashboard_id 2,
                          :dashboard_tab_id nil,
                          :entity_id "Z95aWfOkx1FZGk0OXVQRN",
                          :id 1,
                          :parameter_mappings [{:card_id 1,
                                                :parameter_id "fadff97c",
                                                :target [:dimension [:template-tag "this_field"]]}
                                               {:card_id 1,
                                                :parameter_id "f57f3a34",
                                                :target [:dimension [:template-tag "exclude_this_field"]]}
                                               {:card_id 1,
                                                :parameter_id "38b52427",
                                                :target [:dimension [:template-tag "a_starts_with"]]}
                                               {:card_id 1,
                                                :parameter_id "864d5df1",
                                                :target [:dimension [:template-tag "amount_between"]]}],
                          :row 0,
                          :size_x 12,
                          :size_y 9,
                          :updated_at #t "2023-10-20T14:08:31.905235Z",
                          :visualization_settings {}}),
              :parameter_id "38b52427",
              :target [:dimension [:template-tag "a_starts_with"]]}},
 :name "Text starts with",
 :sectionId "string",
 :slug "text_starts_with",
 :type :string/starts-with}
</code></pre>
</details>

So no `:options` in param itself, unlike in dashboard_test.clj

shared/src/metabase/mbql/util.cljc
piranha: And then used like that:

```
(defn- get-template-tag
  "Fetch the `:field` clause from `dashcard` referenced by `:template-tag`.

    (get-template-tag [:template-tag :company] some-dashcard) ; -> [:field 100 nil]"
  [dimension card]
  (when-let [[_ tag] (mbql.u/check-clause :template-tag dimension)]
    (get-in card [:dataset_query :native :template-tags (u/qualified-name tag)])))
```

src/metabase/api/dashboard.clj
piranha: tests for this are in e0e142abc41482fc2224fc754641a16aa5d1e2d7

qnkhuat (approved):
Awesome, the new change lgtm

PR #34840, opened 2023-10-19
Rename dashboard's ordered_tabs to tabs

calherries:
This PR renames the `ordered_tabs` key on Dashboard entities to just `tabs`. The key is hydrated onto dashboards so there is no need for a migration.

It is similar to https://github.com/metabase/metabase/pull/34792 which renames ordered_cards to dashcards.

In each commit in this PR I'm doing a simple automated search-and-replace. I assume that ordered_tab, OrderedTab and ordered-tab are only used in the codebase to refer to dashboard's `ordered_tabs` and not something else.

Note there's no `orderedTab` in the codebase, so there is no commit for that.

### DIFF
--- a/docs/api/dashboard.md
+++ b/docs/api/dashboard.md
@@ -73,9 +73,9 @@ Fetch possible values of the parameter whose ID is `:param-key` that contain `:q
 
 *  **`id`** value must be an integer greater than zero.
 
-*  **`param-key`** 
+*  **`param-key`**
 
-*  **`query`** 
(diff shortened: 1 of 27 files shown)


ranquild (approved):

qnkhuat (approved):

paoliniluis:
This is a change in the API and we should notify it on the release notes

EuanKirkhope:
could be cause of browser error reported in #34998

PR #34851, opened 2023-10-19
[MLv2] Use `FilterPicker` for QB filter pills

kulyk:
The `QuestionFilters.jsx` components clean up is cherry-picked here from #34852


frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/CustomLinkText.tsx
kulyk, line 1: Just improving a11y along the way as I was updating the e2e tests

frontend/src/metabase/query_builder/components/filters/pickers/DatePicker/RelativeDatePicker.tsx
kulyk, line 1: [`relative-datetime.cy.spec.js` test-suite](https://github.com/metabase/metabase/blob/master/e2e/test/scenarios/filters/relative-datetime.cy.spec.js) has a set of helpers that are used for both old date picker and new date picker (because the test suite uses the new `FilterPicker` in the filter pills and the old `FilterPopover` in filter drills). I've added the same a11y labels to the old component, so we can use nice element queries like `findLabelText` and avoid conditional logic

kulyk: Once #34962, #35363, and #35369 are completed, we should switch the test suite to use [`e2e-relative-date-picker-helpers`](https://github.com/metabase/metabase/blob/34112-filters-milestone1/e2e/support/helpers/e2e-relative-date-picker-helpers.js)

PR #34856, opened 2023-10-19
docs - note on language direction

jeff-bruemmer:
Note for supported languages that read right to left.


docs/configuring-metabase/localization.md
```diff
@@ -47,6 +47,8 @@ The languages you can currently pick from are:
 - Ukrainian
 - Vietnamese
 
+> While Metabase can support languages that read right to left, the Metabase UI is designed around languages that read left to right.
```
sarina-dd, line 50: The comment is more confusing than clarifying. If the Metabase UI is designed around languages that read left to right, will there be issues? This feels like we're preemptively telling people expect a wonky user experience in other languages.

jeff-bruemmer: Unfortunately, "preemptively telling people to expect a wonky user experience" is what we're trying to do here. 

How about:

```
> While Metabase can support languages that read right to left, the Metabase UI is designed around languages that read left to right, so things might look funny in right-to-left languages.
```

sarina-dd: Got it, that makes sense.

sarina-dd (approved):

Issue #33364, opened 2023-08-21
Role-based security at database level does not work if the role contains special characters such as the dot. (v0.47)


Roman2nc:
### Describe the bug

Role-based security at database level does not work if the role contains special characters such as the dot.
I think I'm missing the double quote on the role during the sql query.
https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

I use the login from the ldap attribute as the role for postgresql (which is made up of the first name, a dot and the last name)

example login/role
"homer.simpson"

### To Reproduce

configur user attribut (login)
![impersonated](https://github.com/metabase/metabase/assets/66816029/1e0b6618-e5eb-46a4-b68d-90b8f3e7ce65)

error with user "homer.simpson"
![error](https://github.com/metabase/metabase/assets/66816029/5bb51eb0-6f36-47b7-a4cb-728244147005)

### Information about your Metabase installation

```JSON
metabase : version 0.47

postgresql v15
```

### Severity

blocking new feature security

PR #34858, opened 2023-10-19
Only quote roles for connection impersonation when they contain special characters

### LINKED ISSUE #34829: Snowflake Database Setting allows you to input small caps role but breaks connection afterwards (used to work for both in 1.46.8)
### Describe the bug

When you are creating a Snowflake connection you can specify a Role and it will allow you to enter both a small caps `sysadmin` and large caps `SYSADMIN` role but will break for the small caps role:

image

This is critical because in 1.46.8 it used to work for both so people upgrading will end up breaking their instance if the role they specified is in small caps

### To Reproduce

1. Go to New Database Connection -> Snowflake -> Specify role as small caps (saves successfully):

image

2. New -> Native SQL -> `Select 1`

image

3. Notice the error:

`SQL compilation error: Object does not exist, or operation cannot be performed.`

### Expected behavior

Works for both small caps and large caps roles

### Logs

SQL compilation error: Object does not exist, or operation cannot be performed.

```JSON
1.47.4 and master
```

### Severity

Blocker for people that are upgrading form 1.46 and have a role specified as small caps

### Additional context

Workaround is to define the role as large caps i.e. from `sysadmin` -> `SYSADMIN`

noahmoss:
The fix for https://github.com/metabase/metabase/issues/33364 caused https://github.com/metabase/metabase/issues/34829 — quoting roles indiscriminately resulted in them being case-sensitive when they previously weren't. This PR fixes that issue by only quoting them when they contain characters other than alphanumerical characters or underscores, which are the only ones that are allowed in unquoted role identifiers, as far as I can determine. Also added some unit tests for the snowflake & postgres implementations of `set-role-statement`.

resolves https://github.com/metabase/metabase/issues/34829

### DIFF
--- a/modules/drivers/snowflake/src/metabase/driver/snowflake.clj
+++ b/modules/drivers/snowflake/src/metabase/driver/snowflake.clj
@@ -581,7 +581,11 @@
 
 (defmethod driver.sql/set-role-statement :snowflake
   [_ role]
-  (format "USE ROLE \"%s\";" role))
+  (let [special-chars-pattern #"[^a-zA-Z0-9_]"
+        needs-quote           (re-find special-chars-pattern role)]
+    (if needs-quote
+      (format "USE ROLE \"%s\";" role)
+      (format "USE ROLE %s;" role))))
 
 (defmethod driver.sql/default-database-role :snowflake
   [_ database]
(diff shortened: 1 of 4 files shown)


escherize (approved):

escherize:
The only pang I had was that the logic seems to be duplicated, but not abstracting on such a small thing feels like a better solution.

PR #34865, opened 2023-10-19
Add `:audit` flag to most settings, and obfuscate sensitive settings automatically

noahmoss:
* Adds the `:audit` flag to most existing settings. Most of them use `:audit :getter` so that the full value is logged, as returned by their getter.
	* This is done even for settings that don't have a custom getter, mainly so that ints and booleans are logged as their respective types (i.e. the value returned by `get-value-of-type`) rather than the raw string in the database. I'm not in love with the naming here because it seems tied to custom getters...if anyone has a better suggestion I'd be happy to change it.
* Adds obfuscation for settings marked as `sensitive?` when we're writing them to the audit log. This was an oversight in the original PR so I've added it here.

### DIFF
--- a/src/metabase/api/user.clj
+++ b/src/metabase/api/user.clj
@@ -31,19 +31,20 @@
    [metabase.util.password :as u.password]
    [toucan2.core :as t2]))
 
-(defsetting user-visibility
-  (deferred-tru "Note: Sandboxed users will never see suggestions.")
-  :visibility   :authenticated
-  :feature      :email-restrict-recipients
-  :type         :keyword
-  :default      :all)
-
 (set! *warn-on-reflection* true)
 
 (when config/ee-available?
   (classloader/require 'metabase-enterprise.sandbox.api.util
                        'metabase-enterprise.advanced-permissions.common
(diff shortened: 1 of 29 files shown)


src/metabase/api/user.clj
```diff
@@ -31,20 +31,21 @@
...
-
 (set! *warn-on-reflection* true)
 
 (when config/ee-available?
   (classloader/require 'metabase-enterprise.sandbox.api.util
                        'metabase-enterprise.advanced-permissions.common
                        'metabase-enterprise.advanced-permissions.models.permissions.group-manager))
 
+(defsetting user-visibility
+  (deferred-tru "Note: Sandboxed users will never see suggestions.")
```
escherize, line 42: What is the link between user-visibility and suggestions?

escherize (approved):

src/metabase/api/user.clj
noahmoss: `user-visibility` is a setting that defines whether users should see other users in the dropdown when creating a subscription (aka "suggestions" I think)

But I'm not changing anything about the setting here, just moved it a few lines down since usually the `*warn-on-reflection*` and classloader stuff is at the top in most namespaces

PR #34867, opened 2023-10-19
Add aggregation-position to available-metrics

### LINKED ISSUE #34739: [MLv2] Add `aggregationPosition` to metric display info
When clicking an aggregation in the notebook editor, the FE should highlight a metric the aggregation is using

We could follow the same approach we have for columns and make the `available-metrics` method append `aggregationPosition` properties to metrics used in aggregations on a given stage

metamben:
Fixes #34739.


camsaul (approved):

kulyk:
@metamben I think it might make sense to backport this for 47?

Issue #21133, opened 2022-03-21
Metabase field filter casts timestamp column instead of casting queried date


djdylan2000:
**Describe the bug**

Say I have an expense table in Postgres

```
CREATE TABLE expense (
    id uuid PRIMARY KEY,
    user_id uuid NOT NULL,
    performed timestamp with time zone NOT NULL,
    amount json NOT NULL,
);
```

And I create a metabase question to allow other people to query this easily for a user for some period of time.
In metabase, I could do something like

```
SELECT * from "expense" where 1 = 1 [[AND "user_id" = uuid({{user_id}} [[AND {{performed}}]]
```

where user_id is a text field and performed is a field filter (widget type - month and year)

If a userId is typed in and a the field filter is used to pick Feb 2022 for example, metabase generates a query like

```
SELECT * from \"expense\" where 1 = 1 AND \"user_id" = 'xyz' AND CAST(\"public\".\"expense\".\"performed\" AS date) BETWEEN timestamp with time zone '2022-02-01 00:00:00.000+01:00' AND timestamp with time zone '2022-02-28 00:00:00.000+01:00
```

If I have a multi column index on (user_id, performed), unfortunately only user_id part of the index can be used because of the cast as date function. 

IMO this is a performance bug -metabase should cast the arguments of the field filter to match the column type rather than the other way round, so that table indexes can be used efficiently.

I'm running the latest metabase as of today - v0.42.2

Using a function based index is not a workaround because casting a timestamp with timezone to date is not an immutable function.

**Logs**
Please include javascript console and server logs around the time this bug occurred. For information about how to get these, consult our [bug troubleshooting guide](https://metabase.com/docs/latest/troubleshooting-guide/bugs.html)

**To Reproduce**
1. Create a table in postgres with a timestamp with timezone column
2. Create a metabase question with that column used as a field filter with year and month
3. Create an index on the timestamp column
4. Run a postgres query filing in a month and year
5. Postgres casts the column in the where clause, causing the index to be useless

**Expected behavior**
Postgres should avoid casting the database column wherever possible and instead cast the query arguments to match the column datatype, so that indexes can be used efficiently

**Screenshots**

**Information about your Metabase Installation:**

You can get this information by going to Admin -> Troubleshooting.

```
{
  "browser-info": {
    "language": "en-GB",
    "platform": "MacIntel",
[...]

PR #34873, opened 2023-10-20
Add `align-temporal-unit-with-param-type` multimethod

slvrtrn:
Allows to implement a fix for https://github.com/ClickHouse/metabase-clickhouse-driver/issues/196 on the driver level. There is also https://github.com/metabase/metabase/issues/21133, and it will possibly allow to fix that issue as well.

### Description

Due to the hardcoded `:day` value, it was not possible to use last/next X minutes/hours with ClickHouse drivers as it generated incompatible CAST statements (see the mentioned issue). 

### How to verify

It does not require verification; the default implementation is unchanged, and the existing drivers will just use it.

### Checklist


metamben (commented):
The PR looks good. I wonder if there is good and more specific name, something like `align-temporal-unit-with-param-type`. It's a bit long, but I feel `->temporal-unit` is too generic even if it's namespace qualified. Also, a test demonstrating that the type alignment can be overridden would be nice, otherwise it looks almost dead.

metamben (approved):

PR #34877, opened 2023-10-20
Add sql-formatter-dialect multimethod

### LINKED ISSUE #34235: Incorrect dialect is chosen for ClickHouse pretty SQL formatting
### Describe the bug

`/api/dataset/native` calls `mdb.query/format-sql` in the following way: 
```clj 
(u/ignore-exceptions (mdb.query/format-sql q driver)) q))
```

(see https://github.com/metabase/metabase/blob/v0.47.2/src/metabase/api/dataset.clj#L172)

the `driver` argument is passed to `format-sql*` as `db-type`, but the `db-type` variants are hardcoded and do not use `quote-style` or other multimethods that drivers can override.

```clj
    (let [formatter (SqlFormatter/of (case db-type
                                       :mysql Dialect/MySql
                                       :postgres Dialect/PostgreSql
                                       :redshift Dialect/Redshift
                                       :sparksql Dialect/SparkSql
                                       :sqlserver Dialect/TSql
                                       :oracle Dialect/PlSql
                                       :bigquery-cloud-sdk Dialect/MySql
                                       Dialect/StandardSql))]
```

(see https://github.com/metabase/metabase/blob/v0.47.2/src/metabase/db/query.clj#L42-L50)

which means that all other drivers _always_ default to `Dialect/StandardSql`, which potentially generates invalid queries (and not very pretty). For example:

```
"SELECT\n  ` test_data `.` venues `.` id ` AS ` id `\nFROM\n  ` test_data `.` venues `\nORDER BY\n  ` test_data `.` venues `.` id ` ASC\nLIMIT\n  5"
```

There are superfluous spaces between backticks and identifiers, which does not work with ClickHouse.

See this report in the driver repo: https://github.com/ClickHouse/metabase-clickhouse-driver/issues/195

### To Reproduce

Spin up a ClickHouse instance with Metabase driver. 
Create an arbitrary table, for example:

```sql
CREATE TABLE wares (
  id UInt64,
  name String,
  createdAt DateTime
) ENGINE MergeTree ORDER BY id;

INSERT INTO wares VALUES
(1, 'foo', toDateTime('2023-01-01 04:00:00')), (2, 'bar', toDateTime('2023-01-01 04:01:00'));
```

Convert any question to this table to SQL.

image

### Expected behavior

Metabase should either:
- respect the `quote-style` declared in the driver implementation and derive the dialect from there 
- provide a multimethod to override the pretty-printing dialect and use it in the `format-sql*` function
- provide a multimethod to override the entire pretty-printing implementation

Otherwise, as it is currently, non-core drivers always fall back to `Dialect/StandardSql`.

### Logs

_No response_

```
Metabase 0.47.2
ClickHouse 23.8
```

### Severity

Blocking some users

### Additional context

_No response_

slvrtrn:
Closes https://github.com/metabase/metabase/issues/34235
See https://github.com/ClickHouse/metabase-clickhouse-driver/issues/195

### Description

As of 0.47.x, all custom drivers defaulted to `Dialect/StandardSql` as the branching in the SQL format implementation was for the core drivers only. This PR adds a multimethod to override the chosen dialect for third-party drivers.

NB: this bit for some reason was repeated twice in `postgres.clj`, so it was removed.

```clojure
;;; |                                             metabase.driver impls                                              |

(defmethod driver/display-name :postgres [_] "PostgreSQL")
```

### How to verify

Shouldn't require additional verification as there is a default implementation + driver overrides that should behave the same as before.

### Checklist

- [x] Tests have been added/updated to cover changes in this PR - should rely on the existing tests.

### DIFF
--- a/src/metabase/db/query.clj
+++ b/src/metabase/db/query.clj
@@ -23,9 +23,11 @@
    [clojure.string :as str]
    [honey.sql :as sql]
    [metabase.db.connection :as mdb.connection]
+   [metabase.driver :as driver]
    [metabase.driver.impl :as driver.impl]
    [metabase.plugins.classloader :as classloader]
    [metabase.util.log :as log]
+   [metabase.util.malli :as mu]
    [toucan2.core :as t2]
    [toucan2.jdbc :as t2.jdbc])
   (:import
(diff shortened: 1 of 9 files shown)


src/metabase/db/query.clj
```diff
@@ -34,20 +36,37 @@
...
-    (let [formatter (SqlFormatter/of (case db-type
-                                       :mysql Dialect/MySql
-                                       :postgres Dialect/PostgreSql
-                                       :redshift Dialect/Redshift
-                                       :sparksql Dialect/SparkSql
-                                       :sqlserver Dialect/TSql
-                                       :oracle Dialect/PlSql
-                                       :bigquery-cloud-sdk Dialect/MySql
-                                       Dialect/StandardSql))]
+    (let [formatter (SqlFormatter/of (sql-dialects (sql-formatter-dialect driver)))]
```
slvrtrn, line 69: We could probably leave the existing case switch as-is, calling the multimethod only for the default case. However, it will require importing `(com.github.vertical_blank.sqlformatter.languages Dialect)` in the driver code that requires dialect overrides, and it looks like a local implementation detail plus, generally speaking, it is not very pretty. Additionally, in case of the formatter library change, the drivers will also have to fix the imports, etc. So I decided to write a thin wrapper over it with some type hints (apologies if I use it in the wrong way) as it probably will be more maintainable.

metamben (commented):
Wouldn't a more generic `prettify-query` make sense? That should be just as easy to reduce to some other driver's `prettify-query`. Prettifying should not be restricted to SQL dialects, I think.

slvrtrn:
I can refactor it entirely, then. Suppose we have `prettify-query` or something like that with default identity implementation. In that case, we can implement the entire logic on the driver side, including, for example, MongoDB, which, as I understand, is not "prettified" currently: `(isa? driver.impl/hierarchy driver :sql)` is false.

Cons: ugly imports/Java interop code in the drivers' code, including the core ones.

metamben:
#34991 implements the approach I suggested, I expected it to get merged soon.

slvrtrn:
Let's close this one then.

PR #34887, opened 2023-10-20
Search option to return only or exclude items in personal collections

qnkhuat:
BE for milestone 2 of https://github.com/metabase/metabase/issues/34733

This adds a new filter option to search API `filter_items_in_personal_collection` with 2 options "only" or "exclude"


src/metabase/api/search.clj
```diff
@@ -106,21 +106,48 @@
...
                                   visible-collections)
-        honeysql-query           (-> honeysql-query
-                                     (sql.helpers/where collection-filter-clause)
-                                     (sql.helpers/where [:= :collection.namespace nil]))]
+        honeysql-query           (cond-> honeysql-query
+                                   true
+                                   (sql.helpers/where collection-filter-clause [:= :collection.namespace nil])
+
+                                   (some? filter-items-in-personal-collection)
+                                   identity)]
```
tsmacdonald, line 122: Isn't this a no-op?

tsmacdonald (approved):
:boat:

src/metabase/api/search.clj
qnkhuat: yep, I've removed it in https://github.com/metabase/metabase/pull/34887/commits/7544cf02639321ae53d24746304755dcebfe5c86

Issue #34880, opened 2023-10-20
[INTERNAL] Refine file-paths related to the CI


nemanjaglumac:
We currently have one huge rule called `*ci` that includes almost all files in the `.github/` directory (except markdown and `team.json`). And we have way too many workflows and actions. Any time anyone touches **any** file in this folder, no matter how (un)related it might be, literally all workflows will spin up. It's wasteful and sub-optimal.

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/34882
- [ ] https://github.com/metabase/metabase/pull/34890
- [ ] https://github.com/metabase/metabase/pull/35527
- [ ] https://github.com/metabase/metabase/pull/35533
- [ ] https://github.com/metabase/metabase/pull/34892
- [ ] https://github.com/metabase/metabase/pull/34209
- [ ] https://github.com/metabase/metabase/pull/35580
- [ ] https://github.com/metabase/metabase/pull/35617
```

nemanjaglumac:
This Epic is now complete.
Some further improvements might include: cljs, whitespace and build scripts workflows.

That was outside of the current scope.

PR #34890, opened 2023-10-20
Refine CI triggers for `backend` and `drivers` tests

nemanjaglumac:
This PR defines a set of specific workflows and/or actions that, when altered, should trigger backend and drivers tests in CI. It is a part of #34880

### How to test?
- Grep https://github.com/metabase/metabase/blob/master/.github/workflows/backend.yml and https://github.com/metabase/metabase/blob/master/.github/workflows/drivers.yml for mention of `./.github`. It should give you the sense of the files that these workflows depend on.
- Using this branch as the base, change any of the listed files and make sure that the both backend and drivers workflows are running
- Change any of the backend source or test files and make sure that the both workflows are running as before
- Change any of the unrelated (not listed here) workflows or actions and make sure that the both workflows are NOT running

## Tests

```[tasklist]
### Positive cases (changing the following files **should** trigger `drivers` and `backend` workflows)
- [x] `.github/actions/prepare-frontend/action.yml` (#34977)
- [x] `.github/actions/prepare-backend/action.yml` (#34978)
- [x] `.github/actions/test-driver/action.yml` (#34979)
- [x] `.github/workflows/drivers.yml` (#34980)
- [x] `.github/workflows/backend.yml` (#34981)
```

```[tasklist]
### Negative cases (changing the following files **should not** trigger the FE workflow)
- [x] `.github/workflows/e2e-tests.yml` (#34982)
```

### DIFF
--- a/.github/file-paths.yaml
+++ b/.github/file-paths.yaml
@@ -42,6 +42,13 @@ frontend_all: &frontend_all
   - *frontend_sources
   - *frontend_specs
 
+backend_ci: &backend_ci
+  - ".github/actions/prepare-frontend/**"
+  - ".github/actions/prepare-backend/**"
+  - ".github/actions/test-driver/**"
+  - ".github/workflows/backend.yml"
+  - ".github/workflows/drivers.yml"
+
 backend_presto_kerberos:
   - "**/presto_jdbc/**"
   - "**/presto_jdbc.clj"
@@ -66,7 +73,7 @@ backend_specs: &backend_specs
 
 backend_all: &backend_all
   - *default
-  - *ci
+  - *backend_ci
   - *backend_sources
   - *backend_specs
 


.github/file-paths.yaml
```diff
@@ -42,6 +42,13 @@ frontend_all: &frontend_all
   - *frontend_sources
   - *frontend_specs
 
+backend_ci: &backend_ci
+  - ".github/actions/prepare-frontend/**"
+  - ".github/actions/prepare-backend/**"
+  - ".github/actions/test-driver/**"
+  - ".github/workflows/backend.yml"
+  - ".github/workflows/drivers.yml"
+
```
nemanjaglumac, line 51: In theory, this can be broken down even further into `backend_ci` and `drivers_ci`. If someone wants to follow up with that change, please be my guest.

camsaul (approved):

PR #34893, opened 2023-10-20
fix nested a to avoid warning

npretto:
### Description

This removes this warning we get in the console
image

I think this only shows up in dev mode but it's still some noise in the terminal

### How to verify

1. Open a dashboard with at least a card that has the chart settings button
2. Go in edit mode
3. After this PR the error should not be there

### DIFF
--- a/frontend/src/metabase/dashboard/components/DashCard/DashCardActionButtons/ChartSettingsButton.tsx
+++ b/frontend/src/metabase/dashboard/components/DashCard/DashCardActionButtons/ChartSettingsButton.tsx
@@ -32,6 +32,7 @@ function ChartSettingsButton({
       tall
       triggerElement={
         <DashCardActionButton
+          as="span"
           tooltip={t`Visualization options`}
           aria-label={t`Show visualization options`}
         >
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/dashboard/components/DashCard/DashCardActionButtons/ChartSettingsButton.tsx
```diff
@@ -32,6 +32,7 @@ function ChartSettingsButton({
       tall
       triggerElement={
         <DashCardActionButton
+          as="span"
```
kulyk, line 35: Can we make it an actual `button`? Or at least set `role="button"` on it?

ranquild (commented):
+1, why can't we make it button?

frontend/src/metabase/dashboard/components/DashCard/DashCardActionButtons/ChartSettingsButton.tsx
npretto: `as="button"` seems to change the cursor from pointer to the default 🤔 

I'm actually wondering if we really want it to be a button, the html would be like this:
(`a > button`)
image

The issue is that `triggerElement` is wrapped in a `Trigger` component which is a `a`: https://github.com/metabase/metabase/blob/f3bc571fc80381d773d8ffda101cf2547c8c2cdb/frontend/src/metabase/components/Triggerable/Triggerable.jsx#L10
Causing the nested `a`.

A better solution would probably be adding a prop to `ModalWithTrigger` such as `triggerAs` but it seems to be a deprecated component so probably not worth changing its behaviour.
If we plan on replacing its use here with something else then maybe we can live with the warning until then?

npretto:
Closed as it's been fixed in https://github.com/metabase/metabase/pull/36747

PR #34895, opened 2023-10-20
Separate Replay E2E tests from the main workflow

nemanjaglumac:
Just an idea - let's, at least temporarily, separate Replay run tests from the main workflow.
Yes, it is a lot of duplication but it reduces the cognitive strain when parsing E2E workflow.

The alternative would be to add more conditionals to the main `e2e-tests.yml` but I'd like to avoid that.

Either we will end up using Replay.io to record our E2E CI runs, or we will abandon the idea.
In both cases, it'll be easy to consolidate these workflows.

Let's consider running on replay a trial - because it is.
It makes sense to keep it in a separate file, imo. That is, if you don't really mind the duplication.

### DIFF
--- a/.github/actions/prepare-cypress/action.yml
+++ b/.github/actions/prepare-cypress/action.yml
@@ -14,21 +14,21 @@ runs:
   steps:
     # Browsers
     - name: Install Chrome v111
-      if: inputs.is-replay-browser != true
+      if: ${{ inputs.is-replay-browser != 'true' }}
       uses: browser-actions/setup-chrome@v1
       with:
         # https://chromium.cypress.io/linux/stable/111.0.5563.146
         chrome-version: 1097615
       id: setup-chrome
     - name: Chrome install feedback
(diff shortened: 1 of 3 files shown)


.github/actions/prepare-cypress/action.yml
```diff
@@ -14,21 +14,21 @@ runs:
   steps:
     # Browsers
     - name: Install Chrome v111
-      if: inputs.is-replay-browser != true
+      if: ${{ inputs.is-replay-browser != 'true' }}
```
uladzimirdev, line 17: is it because `is-replay-browser` can't be boolean?

uladzimirdev (approved):
I don't see anything wrong 👍

.github/actions/prepare-cypress/action.yml
nemanjaglumac: It's because GitHub has extremely convoluted and complicated way of matching and casting.
I will document it but for now - this works.

Issue #31498, opened 2023-06-10
[Epic] Add test coverage for Dashboard Click Behavior


alxnddr:
Epic https://github.com/metabase/metabase/issues/30941
Estimate: 3 days

## Dimensions
- User permissions
- Dashcard visualization types
- Click behaviors (update filter value, open a URL, dashboard, question)
- Parameter mapping for click behavior ( to dashboards and questions )
- Is full app embedding

## Actions
- Set click behavior for a dashboard card
- Trigger click behavior on a dashboard card
  - Update a filter value
  - Open a url with interpolated click data
  - Open a question or a dashboard with mapped parameters

## Edge cases

- Target dashboard/question is not visible to a user
- If full app embedding, click behavior opens inner routes within the iframe
- Click behavior mapping uses invalid (e.g. removed) parameters

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/issues/35034
- [ ] https://github.com/metabase/metabase/issues/35035
```

kamilmielnik:
:warning: _this comment is WIP_ :warning: 

## Test cases

- [ ] Does not allow to set click behavior for virtual dashcards
  - [ ] Text
  - [ ] Heading
  - [ ] Action
  - [ ] Link
- [ ] Does not allow to set click behavior for visualizations with disabled click behavior
  - [ ] Object Detail
- [ ] Allows to set click behavior for Line Chart dashcard
  - [ ] "Open the Metabase drill-through menu"
  - [ ] "Go to a custom destination"
    - [ ] Dashboard
      - [ ] With passed values to dashboard filters
      - [ ] Does not show dashboards user has no permissions to
    - [ ] Saved question
      - [ ] With passed values to filter the question
      - [ ] Does not show questions user has no permissions to
    - [ ] URL
      - [ ] Simple URL
      - [ ] URL with referenced values
  - [ ] "Update a dashboard filter"
- [ ] Allows to set click behavior for Table dashcard
  - [ ] For a single column (cases just as setting click behavior for Line Chart dashcard)
  - [ ] For multiple columns (cases just as setting click behavior for Line Chart dashcard)
- [ ] Custom click behavior works
  - [ ] Line Chart (cases just as setting click behavior for Line Chart dashcard)
  - [ ] Table (cases just as setting click behavior for Line Chart dashcard)

PR #34899, opened 2023-10-20
Add test coverage for Dashboard Click Behavior - Line Chart

### LINKED ISSUE #35034: Add test coverage for Dashboard Click Behavior - Line Chart

kamilmielnik:
Closes #35034

### Description

I decided to split the epic #31498 into 2 tasks to avoid monstrous PRs.

### How to verify

There should be no changes to how the app works.
Added e2e tests should represent the testing plan for **Line Chart**, see #31498.


e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,692 @@
...
+          cy.get("aside").findByText("URL").click();
+          modal().within(() => {
+            cy.findByRole("textbox").type(URL);
+            cy.button("Done").click();
+          });
+          cy.get("aside").button("Done").click();
+
+          saveDashboard();
+
+          onNextAnchorClick(anchor => {
```
kamilmielnik, line 484: - We use [custom click behavior](https://github.com/metabase/metabase/blob/7820266e7d551b038df60b5e8317b16b7ed7a90e/frontend/src/metabase/lib/dom.js#L301-L312) which is a nightmare to test in Cypress. I tried many ways but none of them do the trick. Resources I checked/tried:
  - https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/testing-dom__tab-handling-links
  - https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/stubbing-spying__window
  - https://docs.cypress.io/guides/references/trade-offs#Multiple-tabs
  - https://glebbahmutov.com/blog/stub-window-open/
  - https://glebbahmutov.com/blog/cypress-tips-and-tricks/#deal-with-windowopen
  - https://stackoverflow.com/questions/64234771/cypress-prevent-open-new-tab-with-link-dynamically-created-and-clicked

e2e/support/helpers/e2e-dashboard-helpers.js
```diff
@@ -189,10 +189,6 @@ export function resizeDashboardCard({ card, x, y }) {
   });
 }
 
-export function createLinkCard() {
-  cy.icon("link").click();
-}
```
kamilmielnik, line 194: Inlined this little helper because:
- its name was inconsistent with `createTextCard` which does something else
- I needed a `createLinkCard` helper that does similar thing to `createTextCard`

e2e/support/helpers/e2e-dashboard-helpers.js
```diff
@@ -248,6 +242,78 @@ export function createTextCard({
   };
 }
 
+export function createHeadingCard() {
```
kamilmielnik, line 245: All changes outside of `e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js` are related to introduction of `createHeadingCard`, `createActionCard`, `createLinkCard` and removal (i.e. inlining) of old `createLinkCard`.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,716 @@
...
+     * This problem prevails when updating dashboard filter,
+     * where the 2 clicks will cancel each other out.
+     **/
+    .then(([circle]) => {
+      const { left, top } = circle.getBoundingClientRect();
+      cy.get("body").click(left, top);
+    });
+};
+
+const addDashboardDestination = () => {
```
kamilmielnik, line 665: These helpers will also be used in [table tests](https://github.com/metabase/metabase/issues/35035)

e2e/support/helpers/e2e-dashboard-helpers.js
npretto: I agree with the first point, I run into that myself when I was adding a helper to create a heading card card via the api (see [here](https://github.com/metabase/metabase/pull/34834#discussion_r1365610221)
Should we discuss this in a RFC? I feel like it's good to have have helpers for both creating cards via API and via the UI (sometimes it's more than a `get().click` as they involve dropdowns), but we should find a proper naming for them

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,711 @@
...
+        ({ body: card }) => {
+          visitDashboard(card.dashboard_id);
+        },
+      );
+
+      clickLineChartPoint();
+      assertDrillThroughMenuOpen();
+    });
+
+    it("does not allow setting dashboard as custom destination if user has no permissions to it", () => {
```
WiNloSt, line 147: nit: The no permission test case for dashboard comes first while no permission test case for questions comes last. This is also inconsistent with the test listed in [this comment](https://github.com/metabase/metabase/issues/31498#issuecomment-1771094663) (no permission comes last in that comment).

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,711 @@
...
+      );
+
+      editDashboard();
+
+      getDashboardCard().realHover().icon("click").click();
+      cy.get("aside").findByText("Go to a custom destination").click();
+      cy.get("aside").findByText("Saved question").click();
+
+      modal().findByText(RESTRICTED_COLLECTION_NAME).should("not.exist");
+      modal().findByText(TARGET_QUESTION.name).should("not.exist");
```
WiNloSt, line 427: This is similar to another comment I made for setting dashboards as a custom destination. I think it's a bit misleading since the target question will never be shown here even if users have access to the collection since they need to navigate into the collection first to see the target question there.

I think we should remove this assertion.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,711 @@
...
+    });
+
+    it("allows setting URL as custom destination and changing it back to default click behavior", () => {
```
WiNloSt, line 430: I wonder if it's deliberate that this is the only test that tests removing the click behavior. Why not testing removing the click behavior for dashboards and saved questions too?

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,711 @@
...
+
+      saveDashboard();
+
+      clickLineChartPoint();
+      assertDrillThroughMenuOpen();
+    });
+
+    it("allows setting URL with parameters as custom destination", () => {
+      const urlWithParams = `${URL}{{${DASHBOARD_FILTER_TEXT.slug}}}/{{${COUNT_COLUMN_ID}}}/{{${CREATED_AT_COLUMN_ID}}}`;
+      const escapedUrlWithParams = escapeCypressCurlyBraces(urlWithParams);
```
WiNloSt, line 476: Can we use `{ parseSpecialCharSequences: false }` when typing instead of introducing this util?

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,711 @@
...
+      addTextParameter();
+      cy.get("aside").button("Done").click();
+
+      saveDashboard();
+
+      clickLineChartPoint();
+      cy.findAllByTestId("field-set")
+        .should("have.length", 1)
+        .should("contain.text", POINT_COUNT);
+      cy.location("search").should(
```
WiNloSt, line 566: To ensure we're still staying on the same dashboard and only updating the filter, should we add an assertion for either
- dashboard name
- or the dashboard ID in the URL like in other test case

WiNloSt (dismissed):
Looks good to me with minor feedback

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,711 @@
...
+        ({ body: card }) => {
+          visitDashboard(card.dashboard_id);
+        },
+      );
+
+      editDashboard();
+
+      getDashboardCard().realHover().icon("click").click();
+      addDashboardDestination();
+      cy.get("aside").findByText("No available targets").should("exist");
```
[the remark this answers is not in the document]
kamilmielnik: 181cbf1e6055b3d1d39ceb7b6c491f7a7aa52cf9

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
kamilmielnik: Nice, thanks! ceb6745f0c93f35efcf27f5feac34da90d607269

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
kamilmielnik: 010804b743fbb99a952a5cd472664bf172cafe9a

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
kamilmielnik: I think that testing the "Open drill-through menu" option needs only 1 test, and since it's the default and requires non-default being selected beforehand, it felt reasonable to just add it in one of the existing tests.

I picked this one only because it was the first one that I implemented.

Please let me know if you think this should be added to more (all?) tests.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -0,0 +1,716 @@
...
+        ({ body: card }) => {
+          visitDashboard(card.dashboard_id);
+          editDashboard();
+
+          getDashboardCard().realHover().icon("click").click();
+          cy.get("aside").findByText("Go to a custom destination").click();
+          cy.get("aside").findByText("Dashboard").click();
+
+          modal().findByText(RESTRICTED_COLLECTION_NAME).should("not.exist");
+          modal().findByText(TARGET_DASHBOARD.name).should("not.exist");
```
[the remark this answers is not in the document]
kamilmielnik: e24965ec2ab309ef54a09066077920443cb2c3cb

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
kamilmielnik: e24965ec2ab309ef54a09066077920443cb2c3cb

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
WiNloSt: I mean I don't mind that we have to test removing click behavior in all dimensions, but it's good to have that covered. What about adding 2 more tests (or new assertions in existing tests)? One in dashboards and one for questions, so we don't need to add a ton, but still have them covered.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
kamilmielnik: df4564e5c9a2788e9343869174861b2bcf1bb590

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
kamilmielnik: > One in dashboards and one for questions, so we don't need to add a ton, but still have them covered.

Only "custom destination > URL" and "updating dashboard filters" do not require navigating back and forth.
I'm kind of hesitant to add it elsewhere because of that, but here it is: 9374f6a3d9e17da8a0668baf76d3426d773c5ece

kamilmielnik:
I think Graphite is [acting up](https://github.com/metabase/metabase/assets/6830683/d3de8f4c-f8c3-4bd4-b993-1d15e3002a5f) again. @WiNloSt would you mind having another look?

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
WiNloSt: I see what you mean. I haven't accounted for having to navigate back at first.

WiNloSt (approved):
Looks good to me, though I notice there are 100 commits in this PR, not sure if that's correct, or just Graphite acting up since the change looks super weird when I see the changes from commits you made today. I need to review each new commit invidually instead.

Issue #34367, opened 2023-10-04
[Epic] Let people move cards between dashboard tabs


oisincoveney:
**Links**
- [Let people move cards between dashboard tabs](https://www.notion.so/metabase/Let-people-move-cards-between-dashboard-tabs-e797b321092b45a1bcbc4478d982a01e?pvs=4) 
- Testing Plan: https://github.com/metabase/metabase/issues/34368
- feature branch: `move-cards-between-tabs-feature-branch` 

**Implementation Plan**
* Implement the tab menu, and dashcard movement logic
* Implement logic to ensure that dashcards retain their size when moved across tabs

***Milestone 1: Initial dashcard movement***

```[tasklist]
### Tasks
- [x] Implement dashcard movement between tabs (`MOVE_DASHCARD_TO_TAB` redux action)
- [x] Implement dashcard tab menu within the toolbar (`DashCardTabMenu.tsx`) https://github.com/metabase/metabase/pull/34559
- [x] Check with PD/PMs that the toolbar looks good on small cards
- [ ] https://github.com/metabase/metabase/pull/34834
- [ ] https://github.com/metabase/metabase/pull/34846
- [ ] https://github.com/metabase/metabase/pull/34906
```

***Milestone 2: Undo notification***

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/34961
```
***Milestone 3: Analytics***

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/issues/35123
```

***Milestone 4: [Polishment] left-align dashcard actions for small cards***
Milestone PR: https://github.com/metabase/metabase/pull/35240

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/35240
- [ ] https://github.com/metabase/metabase/pull/35243
- [ ] https://github.com/metabase/metabase/pull/35324
- [ ] https://github.com/metabase/metabase/pull/35327
- [ ] https://github.com/metabase/metabase/pull/35378
- [ ] https://github.com/metabase/metabase/pull/35366
- [ ] https://github.com/metabase/metabase/pull/35502
```

npretto:
Updated the epic to be aligned with the recent discussions from slack: https://metaboat.slack.com/archives/C057T1QTB3L/p1697471382430159?thread_ts=1697110097.943389&cid=C057T1QTB3L

PR #34906, opened 2023-10-20
[Move dashcard to tab] permissions e2e + helper e2e methods

npretto:
epic: https://github.com/metabase/metabase/issues/34367

### Description

- https://github.com/metabase/metabase/commit/8462081fe679786ba234ca067e3f51e0d631c950 adds a e2e to check for the permissions edge case, in which you don't have access to the card but should be able to move the dashcard if you have edit permissions in the dashboard
- https://github.com/metabase/metabase/commit/01b5e992dc05fcde824c3c90824fc5a906121d14 adds a goToTab helper
- https://github.com/metabase/metabase/commit/1b03cf013c9a01e35c7129af7b206f02945ce2f2 adds a helper to move a dashcard to another tab

Note: e2e are probably failing because of https://metaboat.slack.com/archives/C5XHN8GLW/p1697729886291229

### DIFF
--- a/e2e/test/scenarios/dashboard/tabs.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/tabs.cy.spec.js
@@ -18,13 +18,13 @@ import {
   visitCollection,
   main,
   getDashboardCard,
-  menu,
   getDashboardCards,
   getTextCardDetails,
   getHeadingCardDetails,
   getLinkCardDetails,
   updateDashboardCards,
   goToTab,
+  moveDashCardToTab,
 } from "e2e/support/helpers";
 
 import {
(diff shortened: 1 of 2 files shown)


e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -155,25 +156,51 @@ describe("scenarios > dashboard > tabs", () => {
...
+    editDashboard();
+    createNewTab();
+
+    goToTab("Tab 1");
+
+    moveDashCardToTab({ tabName: "Tab 2" });
+
+    saveDashboard();
+
+    getDashboardCards().should("have.length", 0);
```
npretto, line 201: maybe we could add a reload after the save to make sure everything is actually saved?
but it would slow down the test a bit

ranquild (approved):
Could we add an assert that the user sees the "No access" message? Currently the test is a bit hard to follow because there is none

PR #34909, opened 2023-10-20
Filters E2E cases

ranquild:
Fixes: 
- string filter options
- not being able to filter on an aggregated column based on another column
- not being able to use fractional numbers in filters

Other tests TDB in another PR


frontend/src/metabase-lib/filter.ts
```diff
@@ -556,6 +563,15 @@ function isStringOperator(
   return operators.includes(operator);
 }
 
+function getStringFilterOptions(
+  operator: ExpressionOperatorName,
+  options: ExpressionOptions,
+): StringFilterOptions {
+  const operators: ReadonlyArray<string> = STRING_FILTER_OPERATORS_WITH_OPTIONS;
+  const supportsOptions = operators.includes(operator);
+  return supportsOptions ? { "case-sensitive": false, ...options } : {};
```
ranquild, line 572: MLv2 breaks if we pass options for operators that do not support them. I think it's an implementation detail and we should handle it internally and not ask for consumers to know it

frontend/src/metabase/ui/components/inputs/NumberInput/NumberInput.styled.tsx
```diff
@@ -15,6 +15,8 @@ export const getNumberInputOverrides =
     NumberInput: {
       defaultProps: {
         size: "md",
+        precision: 16,
```
ranquild, line 18: Otherwise you can't enter fractional values

frontend/src/metabase/common/components/FilterPicker/FilterOperatorPicker.tsx
```diff
@@ -20,5 +20,5 @@ export function FilterOperatorPicker({
     }));
   }, [options]);
 
-  return <Select data={data} {...props} />;
+  return <Select data={data} {...props} data-testid="filter-operator-picker" />;
```
kulyk, line 23: WDYT about using an `aria-label` instead? Could be something like "Filter operator"

kulyk (approved):
LGTM, left a few minor-ish comments

PR #34903, opened 2023-10-20
Refactor: Remove FieldValuesWidget forceTokenField prop

calherries:
`forceTokenField` can be removed from `FieldValuesWidget` because the `disableList` prop works the same way, so we should just use it instead

related: https://github.com/metabase/metabase/pull/34910

PR #34910, opened 2023-10-20
Refactor: Tidy FieldValuesWidget fetchValues and more

calherries:
This PR is a refactoring I made while trying to tidy up the logic inside the `FieldValuesWidget` component. The goal was to remove the [`hasList`](https://github.com/metabase/metabase/pull/34910/files#r1367229876) function, which I felt was unnecessary indirection.

Each commit is made in a logical sequence. Follow the comments to understand what's happening.

related: https://github.com/metabase/metabase/pull/34903


frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -170,41 +170,30 @@ export function FieldValuesWidgetInner({
     setLoadingState("LOADING");
     setOptions([]);
 
-    let newOptions: FieldValue[] = [];
-    let newValuesMode = valuesMode;
+    let values: FieldValue[] = [];
+    let has_more_values = false;
```
calherries, line 174: I just renamed these to use destructuring to assign values

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -170,41 +170,30 @@ export function FieldValuesWidgetInner({
...
+        ));
       } else {
-        newOptions = await fetchFieldValues(query);
-
-        newValuesMode = getValuesMode({
-          parameter,
-          fields,
-          disableSearch,
-          disablePKRemappingForSearch,
-        });
```
calherries, line 200: This `getValuesMode` is unnecessary because the `valuesMode` is initialized with the exact same `getValuesMode` call.
https://github.com/metabase/metabase/blob/d96d020ec97bdca411f1a9f9bd45b6372c36a9bf/frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx#L147-L154

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -408,11 +397,7 @@ export function FieldValuesWidgetInner({
     valuesMode,
   });
 
-  const isListMode =
-    !disableList &&
-    shouldList({ parameter, fields, disableSearch }) &&
-    valuesMode === "list" &&
-    !forceTokenField;
+  const isListMode = !disableList && valuesMode === "list" && !forceTokenField;
```
calherries, line 390: We can remove `shouldList(...)` because it is implied by `valuesMode === "list"`. 

Reasoning follows: the only use of `setValuesMode` is here in `fetchValues`, setting `valuesMode` to "search".

This means that:
```
valuesMode === "list"
=> valuesMode === getValuesMode(...)  (that's how valuesMode is initialized)
=> shouldList({ parameter, fields, disableSearch }) === true
```

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -424,7 +418,7 @@ export function FieldValuesWidgetInner({
       >
         {isListMode && isLoading ? (
           <LoadingState />
-        ) : isListMode && hasListValues && multi ? (
```
calherries, line 442: We can replace `hasListValues` with `!_.isEmpty(options)` because `isListMode` implies `shouldList(...)`

and this is the definition of `hasList`:

```
export function hasList({
  parameter,
  fields,
  disableSearch,
  options,
}: {
  parameter?: Parameter;
  fields: Field[];
  disableSearch: boolean;
  options: FieldValue[];
}) {
  return (
    shouldList({ parameter, fields, disableSearch }) && !_.isEmpty(options)
  );
}
```

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -421,7 +411,7 @@ export function FieldValuesWidgetInner({
         ) : isListMode && !_.isEmpty(options) && multi ? (
           <ListField
             isDashboardFilter={!!parameter}
-            placeholder={tokenFieldPlaceholder}
+            placeholder={placeholder ?? SEARCH_THE_LIST_PLACEHOLDER}
```
calherries, line 414: We can do this replacement above because `hasList(...)` is implied by `isListMode && !_.isEmpty(options)`, and if `hasList(...)` is true then the tokenFieldPlaceholder will return `placeholder ?? SEARCH_THE_LIST_PLACEHOLDER`
https://github.com/metabase/metabase/pull/34910/commits/d96d020ec97bdca411f1a9f9bd45b6372c36a9bf#diff-c7d2638c5a506ff4d59c4d8d0f346c43cc1e46cafb689f9a7d0d6fb555d967baL256-L265

It doesn't make sense to use `tokenFieldPlaceholder` for a `ListField` component, it should only be used for the `TokenField` component.

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -542,15 +542,7 @@ function renderOptions({
...
-    } else if (
-      hasList({
-        parameter,
-        fields,
-        disableSearch,
-        options,
-      }) &&
-      valuesMode === "list"
-    ) {
+    } else if (!_.isEmpty(options) && valuesMode === "list") {
```
calherries, line 545: We can do this replacement because `valuesMode === "list"` implies `shouldList(...)`, and 
`hasList(...) iff (!_.isEmpty(options) AND shouldList(...))`

frontend/src/metabase/components/FieldValuesWidget/utils.ts
```diff
@@ -174,22 +176,6 @@ function getSearchableTokenFieldPlaceholder(
...
-}: {
-  parameter?: Parameter;
-  fields: Field[];
-  disableSearch: boolean;
-  options: FieldValue[];
-}) {
-  return (
-    shouldList({ parameter, fields, disableSearch }) && !_.isEmpty(options)
-  );
-}
```
calherries, line 191: The goal of this commit was to remove this `hasList` function, which was unnecessary indirection and further confused things

frontend/src/metabase/components/FieldValuesWidget/utils.ts
```diff
@@ -247,22 +232,10 @@ export function getTokenFieldPlaceholder({
   options: FieldValue[];
   valuesMode: ValuesMode;
 }) {
-  if (placeholder) {
-    return placeholder;
+  if (shouldList({ parameter, fields, disableSearch }) && !_.isEmpty(options)) {
```
WiNloSt, line 235: I'm not sure we'd still need this if check anymore. On paper `if (placeholder)` is already replaced by [this change](https://github.com/metabase/metabase/pull/34910/files#diff-b1cf4299750fb16b84cd22730312df1238fd6816eb26bd8d03b44a3aa5da337dR437-R438).

And `shouldList({ parameter, fields, disableSearch }) && !_.isEmpty(options)` is basically `hasList(...)` while this function is only called in the [else block](https://github.com/metabase/metabase/pull/34910/files#diff-b1cf4299750fb16b84cd22730312df1238fd6816eb26bd8d03b44a3aa5da337dR431) which should be something along the line `!hasList(...)` if I read the logic correctly since the first 2 conditions are alreaady `hasList(...)`

Can you explain your reasoning here? I think it makes more sense to remove this block altogether.

Another observation is that `SEARCH_THE_LIST_PLACEHOLDER` is also used in `frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx`. I suspect that is already enough to not have this block. And we might also want to declare `SEARCH_THE_LIST_PLACEHOLDER` in `FieldValuesWidget` instead.

frontend/src/metabase/components/FieldValuesWidget/utils.ts
```diff
@@ -274,11 +247,11 @@ export function getTokenFieldPlaceholder({
     return getSearchableTokenFieldPlaceholder(
       parameter,
       fields,
-      firstField,
+      fields[0],
```
WiNloSt, line 250: I don't think there's anything wrong with reusing the same variable `firstField`. That emphasizes both line 250 and 254 are using the same variable.

I kind of think this is a step backwards

WiNloSt (commented):
Have minor questions. Otherwise, looks pretty good to me.

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -439,20 +408,20 @@ export function FieldValuesWidgetInner({
       >
         {isListMode && isLoading ? (
```
oisincoveney, line 409: Maybe we can clean up this ternary? I feel like nested ternaries can get pretty ugly pretty quickly

frontend/src/metabase/components/FieldValuesWidget/utils.ts
calherries: @WiNloSt I think you're confused by github placing these lines next to each other in the diff viewer. I _have_ removed the `if (placeholder) { return placeholder; } ` block.

This line, 235, replaces [these lines](https://github.com/metabase/metabase/pull/34910/files#diff-c7d2638c5a506ff4d59c4d8d0f346c43cc1e46cafb689f9a7d0d6fb555d967baL257-L263) (257-263).

btw your link to the [else block](https://github.com/metabase/metabase/pull/34910/files#diff-b1cf4299750fb16b84cd22730312df1238fd6816eb26bd8d03b44a3aa5da337dR431) doesn't navigate to any code in my browser.

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
calherries: Agreed but not this PR please

frontend/src/metabase/components/FieldValuesWidget/utils.ts
WiNloSt: @calherries 
Seems linking to collapsed code doesn't work. Here's the mentioned else block.
image

> Another observation is that `SEARCH_THE_LIST_PLACEHOLDER` is also used in `frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx`. I suspect that is already enough to not have this block. And we might also want to declare `SEARCH_THE_LIST_PLACEHOLDER` in `FieldValuesWidget` instead.

This one is more relevant than the `if (placeholder)` thing.

The bottom line is that could you double-check if we still need to have `SEARCH_THE_LIST_PLACEHOLDER` inside `utils.ts` as I think it's impossible for `shouldList({ parameter, fields, disableSearch }) && !_.isEmpty(options)` to be true inside `getTokenFieldPlaceholder` based on how `getTokenFieldPlaceholder` get called (in the mentioned else block)

calherries: Yep, double checking. 

```
We want to prove that:
"TokenWidget is called"
=> line 235 condition is false.

Or the contrapositive:
line 235 condition is true
=> "TokenWidget is not called"

We can substitute
line 235 condition is true <=> shouldList(...) && !_.isEmpty(options)
"TokenWidget is not called" <=> isListMode && !_.isEmpty(options)

So we need to prove
shouldList(...) && !_.isEmpty(options)
=> isListMode && !_.isEmpty(options)

this is not the case if `shouldList(...) && isSearchable(...)`

Simplifying:
shouldList(...) && !_.isEmpty(options)
=> isListMode

To disprove this we can find one instance where 
shouldList(...) && !_.isEmpty && !isListMode

Let's consider
isListMode  = !disableList && valuesMode === "list" && !forceTokenField

so when disableList or forceTokenField is true then it's possible for isListMode to be false but shouldList to be true.

therefore
forceTokenField && shouldList(...) => shouldList(...) && !_.isEmpty(options) && !isListMode

which contradicts the contrapositive
line 235 condition is true
=> "TokenWidget is not called"

I found an instance of this:
https://www.loom.com/share/06da2881c03d4a488d97e5967bdbbe8f?sid=8b8600ac-fb72-437d-97fc-b89108b1ef68

File: frontend/src/metabase/components/FieldValuesWidget/utils.ts (hunk 9 above)
Comment (calherries) [reply frontend/src/metabase/components/FieldValuesWidget/utils.ts]: Fair enough, I restored the old code

File: frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx (hunk 3 above)
Comment (uladzimirdev) [reply frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx:390]: didn't you remove `forceTokenField`?

Comment (calherries) [reply frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx:390]: yes, separate PR. it's still valid

Comment (calherries) [reply frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx:390]: I will fix conflicts

Comment (calherries) [reply frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx]: fixed conflicts, forceTokenField is removed

**Comment by uladzimirdev** [comment]
@calherries I'm working on the bug around those lines, but I'll double check your PR tomorrow before approval

File: frontend/src/metabase/components/FieldValuesWidget/utils.ts (hunk 8 above)
Comment (WiNloSt) [reply frontend/src/metabase/components/FieldValuesWidget/utils.ts:235]: Awesome investigation. Thanks for putting this together. I agree that it's necessary. I just thought there might be a simpler way to deal with not having to share `SEARCH_THE_LIST_PLACEHOLDER` to another file, but this looks like the best option at the moment.

**Review by WiNloSt — DISMISSED** [review]
Looks good to me. I'll let @uladzimirdev approve since he's working in the same area ATM.

File: frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx [hunk 11]
```diff
@@ -436,20 +406,20 @@ export function FieldValuesWidgetInner({
...
           <LoadingState />
-        ) : isListMode && hasListValues && multi ? (
+        ) : isListMode && !_.isEmpty(options) && multi ? (
```
uladzimirdev, line 409: not clear what does it mean from business logic `!_.isEmpty(options)` - it worth having a well named variable for it as per `isListMode`

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -461,7 +431,17 @@ export function FieldValuesWidgetInner({
             prefix={prefix}
             value={value.filter(v => v != null)}
             onChange={onChange}
-            placeholder={tokenFieldPlaceholder}
+            placeholder={
+              placeholder ??
+              getTokenFieldPlaceholder({
```
uladzimirdev, line 436: I don't follow this splitting of placeholder logic. Before all the details were hidden under `getTokenFieldPlaceholder` function. 

It was covered in the function before: `if (placeholder) return placeholder`. 

But I can miss something for sure

uladzimirdev (commented):
Comments/questions inline

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
calherries: Good point. I should use `const hasListValues = !_.isEmpty(options)`

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
calherries: There's no reason really. This is just moving complexity from one place to another. I don't mind really so I will change it back

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
calherries: Decided to use `const hasOptions = !_.isEmpty(options)` instead, to be consistent with `options` elsewhere
cefeebeb8d5a49fa5a8cf71f92c5d4efc66dacf7

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -168,41 +168,30 @@ export function FieldValuesWidgetInner({
...
-    let newValuesMode = valuesMode;
+    let values: FieldValue[] = [];
+    let has_more_values = false;
     try {
       if (canUseDashboardEndpoints(dashboard)) {
-        const { values, has_more_values } =
-          await dispatchFetchDashboardParameterValues(query);
-        newOptions = values;
-        newValuesMode = has_more_values ? "search" : newValuesMode;
+        ({ values, has_more_values } =
```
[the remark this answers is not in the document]
WiNloSt: I'm not used to reassigning values to variables using destructured syntax like this but I wouldn't mind it. Both options look fine to me.

Though in this case, I just realize we name a variable in snake case in JS which we should avoid for consistency, with this in mind, I think @uladzimirdev 's approach is better.

For JS quirks, I personally wouldn't want to have to circumvent some quirks especially when those are build tools problems. If it's a language quirk I might be able to live with that, but if something breaks during the build process, we should fix the build. However, this is irrelevant to the point I'm making above.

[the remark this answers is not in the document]
calherries: Fair points, I've reverted to how it looked before 7f5d37ba89f937c2d3baf7f3a33203aaabb30b05

uladzimirdev (approved):
:shipit:

calherries:
@uladzimirdev This [test](frontend/src/metabase/query_builder/components/filters/pickers/DefaultPicker/DefaultPicker.unit.spec.tsx) and I figured out that my logic was wrong [here](https://github.com/metabase/metabase/pull/34910#discussion_r1367217402)

> We can remove shouldList(...) because it is implied by valuesMode === "list".

Actually `valuesMode === "list"` only implies `shouldList(...)` at initialization, because [this](https://github.com/metabase/metabase/pull/34910#discussion_r1367211964) was wrong, `getValuesMode(...)`'s return value can change as the parameter's fields are mutated. Don't ask me to tell you why, but I know that `shouldList(...)` goes from `true` to `false` because `canListFieldValues(...)` goes from `true` to `false` in that test.

I've reverted the changes so now the tests pass. I think this PR still seems like an improvement, however minor.

Anyway, thank god for our tests because this component is nearly impossible to understand for humans

calherries:
I'm just going to close this without merging because the tests are still failing and the gains in clarity are pretty minimal now anyway.

PR #34921, opened 2023-10-20
Use the display name of the card as join alias

### LINKED ISSUE #21646: We should display joined Saved Question in Query Builder rather than `Question <id>`
Example:

image

`Question 4498` in this case is the Saved Question I'm joining against (I think) altho I'm also starting from a Saved Question so that's not entirely clear. We should show the actual name of the Saved Question instead of just `Question <id>` which is pretty much meaningless

metamben:
Fixes #21646.


camsaul (dismissed):

bshepherdson (dismissed):

bshepherdson (approved):

paoliniluis:
I don´t think we said how much we appreciate you @metamben for this PR

PR #34933, opened 2023-10-20
Ignore undefined tag parameters sent by the FE

### LINKED ISSUE #34729: Using Field Filters and editing query results in schema error
### Describe the bug

If you try to execute a native query with a field filter after editing the query text it will result in an error in the response to POST /api/dataset:

```"error": "Invalid input: {:native {:template-tags {\"text\" {:options [\"invalid type\"]}}}}"```

Inspecting the request payload, we can see `options` has a null value:

image

The full ex-data is attached in the "Logs" section below.

### To Reproduce

1. Create a native query with a field filter:
```
select * from orders where  {{ id }}
```
2. In the variables sidebar, select "Variable Type" > "Field Filter", and "Field to map to" > Orders > ID, and Filter widget type: "ID"
3. Select a value for the filter.
4. Execute the query. Observe the results are returned as expected.
5. Edit the query text by adding or deleting whitespace.
6. Execute the query. Observe the error in the network tab, in the response to POST /api/dataset.

image

The `option=null` and `default=null` entries were not present in the first POST /api/dataset response on step 4.

### Expected behavior

the query should work

```JSON
{
  "browser-info": {
    "language": "en",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.18+10-LTS",
    "java.version": "11.0.18",
    "java.vm.version": "11.0.18+10-LTS",
    "os.name": "Mac OS X",
    "os.version": "12.5.1",
    "user.language": "en",
    "user.timezone": "UTC"
  },
  "metabase-info": {
    "databases": [
      "postgres",
      "h2"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "14.5 (Homebrew)"
      },
      "jdbc-driver": {
        "version": "42.6.0"
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2023-10-19",
      "tag": "v0.48.0-SNAPSHOT",
      "hash": "8f8ab29"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

high, breaks field filters

### Additional context

_No response_

metamben:
Fixes #34729.

The root cause is that `TagEditorParamInner.setType` sets some of the tag properties to `undefined`. The CLJS code converts these into `nil` which results in schema errors.

This PR removes any tag properties with `undefined` value before converting the JS object to a Clojure map.

### DIFF
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -9,6 +9,7 @@
    [filter])
   (:require
    [clojure.walk :as walk]
+   [goog.object :as gobject]
    [medley.core :as m]
    [metabase.lib.convert :as lib.convert]
    [metabase.lib.core :as lib.core]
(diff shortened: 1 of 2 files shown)


camsaul (approved):

calherries (approved):
tested. Works like a charm

Issue #34166, opened 2023-09-27
Improve pulse events


luizarakaki:
We currently have the `pulse-create` and `pulse-delete` events to track dashboard subscriptions.

- [ ] Add the `pulse-update` events to track edits.

- [ ] We should enrich the event `details` with:
- `dashboard_id`
- `parameters`
- `channel`
- `schedule` (can be cron syntax)
- `recipients` (the `details` column in `pulse_channel` and list of IDs in `pulse_recipients`)

- [ ] Rename events to:
- `subscription-create`, `subscription-update`, `subscription-delete`
- `alert-create`, `alert-update`, `alert-delete`

PR #34934, opened 2023-10-20
Improve pulse events

qwef:
Closes https://github.com/metabase/metabase/issues/34166

### Description

Adds the pulse and alert events and enriches the details. Decided not to add the other delete events (see: https://metaboat.slack.com/archives/C056QEW4KNF/p1697836095736509), as its effectively a subset of the update event (and also we never actually delete a pulse, only archive).


src/metabase/events/audit_log.clj
```diff
@@ -85,23 +85,37 @@
 (derive :event/subscription-unsubscribe ::pulse-event)
 (derive :event/subscription-unsubscribe-undo ::pulse-event)
 (derive :event/alert-unsubscribe ::pulse-event)
+(derive :event/subscription-create ::pulse-event)
+(derive :event/subscription-update ::pulse-event)
+
+(defn- create-details-map [pulse name]
+  (let [channels (:channels pulse)]
```
qwef, line 92: `:channels` is array due to alerts being able to have both slack and email on the same alert.

test/metabase/events/audit_log_test.clj
```diff
@@ -329,33 +329,90 @@
...
+(deftest subscription-events-test
+  (t2.with-temp/with-temp [Dashboard      {dashboard-id :id} {}
+                           Pulse          pulse {:archived     false
+                                                 :name         "name"
+                                                 :dashboard_id dashboard-id
+                                                 :parameters   ()}]
+    (let [recipients [{:email "test@metabase.com"}
+                      {:id    (mt/user->id :rasta)
+                       :email "rasta@metabase.com"}]
+          pulse      (assoc pulse
```
qwef, line 341: Unable to add `:channels` directly using `t2.with-temp/with-temp`, so add later

noahmoss (approved):
LGTM

luizarakaki (approved):
It is looking good!

There is just one thing to fix:
For alert events, we need the question_id instead of the dashboard_id (as alerts are related to questions and not dashboards)

Issue #6826, opened 2018-01-31
Wrap text for table output


shawny2221:
Is a way to wrap text result of a table column?  

==================
- Your browser and the version: Chrome 63.0.3239.132
- Your operating system: Mac OSX 10.13.1
- Your databases: Redshift
- Metabase version: 0.27.2
- Metabase hosting environment: Beanstalk
- Metabase internal database: Postgres

:arrow_down: Please click the :+1: reaction instead of leaving a `+1` or `update?` comment

mazameli:
Not currently. For fields that tend to have super long contents, like Descriptions, in the metadata section of the admin panel, you can set a field's `visibility` such that it will only be displayed when looking at the single-record detail view.

salsakran:
Practically speaking, wrapping text lets one cell ruin the display of an entire table, and we're probably not going to go down that route.

hindupuravinash:
@salsakran How about showing a "Wrap text" checkbox in Table settings dialog? So we can choose per table or per column basis.

skotur:
Hey do we know if this was integrated at all? I have a content cell that is very very long.

mazameli:
@skotur Our current suggestion would be to either hide that column in the visualization settings, or set its type to `Description` in the Data Model so that it will only be displayed when viewing a single record's details.

hansyson:
wrapping text on a field by field basis with a max length would be useful

denis-troyan:
I would suggest adding an option to adjust row max/fixed height and width in the table view. This way, the table layout rendering may be kept predictable for word-wrapped columns.

ccemeraldeyes:
> Practically speaking, wrapping text lets one cell ruin the display of an entire table, and we're probably not going to go down that route.

This doesn't make any sense. In what way does it "ruin the display"?

mattarderne:
Rather than wrapping I suggest that the columns are limited in width and any long text fields are viewable by clicking to display in a pop-out. Snowflake does this for JSON and it works well.

Currently the formatting makes using tables with many columns with long text fields entirely unusable.

zmunro:
It would also be great to be able to change the text wrapping of the column titles. I have a lot of complex analytics data that can't be named succinctly and accurately, so the names are long, but the values are short, like `4.2` or something

flamber:
@zmunro Then you're looking for #10372 - upvote by clicking :+1: on the first post.

Theo-Lo:
[...]

PR #34945, opened 2023-10-23
When the table content in the dashboard is too long, a horizontal scr…

stotem:
### Description
When the table content in the dashboard is too long, a horizontal scroll bar will appear to adjust to a new line display to ensure the display effect

### DIFF
--- a/frontend/src/metabase/css/core/box_sizing.css
+++ b/frontend/src/metabase/css/core/box_sizing.css
@@ -16,6 +16,8 @@ ul,
 li,
 span {
   box-sizing: border-box;
+  text-wrap: wrap;
+  line-height: normal;
 }
 
 /* for applying border-box to other elements on ad-hoc basis */


iethree (changes requested):
Thanks for your interest in contributing to making Metabase better! A few notes about this change:

1. Putting non-box-sizing css properties into the box_sizing.css file is not recommended, it will make it hard to find and deal with down the road.
2. We've deprecated the pattern of writing raw css like this, all new css code should target specific components using @emotion/styled. This is outlined in our [developers guide](https://www.metabase.com/docs/latest/developers-guide/frontend#component-styling-tree-rings)
3. I don't think this design is consistent with most users' needs for table components, having text wrap in all cases means that tables will show a lot less data.

modified styles | current behavior
--- | ---
Screen Shot 2023-10-25 at 4 30 54 PM | Screen Shot 2023-10-25 at 4 30 40 PM

In the past we have considered adding a toggle to table components to allow users the option to wrap if they want. If that seems like it meets your needs, feel free to 👍  [this issue](https://github.com/metabase/metabase/issues/6826).

iethree:
One other idea for you! Some new browsers have the ability to add custom css ver certain web pages. In Arc Browser, they call this "[Boosts](https://resources.arc.net/en/articles/6808613-boosts-customize-any-website)", and you can simply apply these styles in your own browser.

Screen Shot 2023-10-25 at 4 42 27 PM

PR #34951, opened 2023-10-23
PoC External workflow run trigger

nemanjaglumac:
## The Problem
Running CI for external contributors can be risky, especially if workflows contain secrets. GitHub has a couple of mechanism in place to protect public open-source repositories from malicious external PRs. But there isn't a single widely-accepted approach that _just works™_.

We've previously attempted an approach laid out in this article: https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets. It proved to be too hacky for our taste as it relied on using mechanisms[^1] that were not designed for this purpose.

This PR attempts to solve this problem in-house with a custom and simple solution.

## The Solution
 tl;dr It Introduces an additional layer (buffer, filter) between an external PR and the target workflows that contain secrets. That layer can be best described as a router with a firewall, albeit an extremely simple one :)

> **Important**
> In order to minimize the human error, and to greatly reduce the risk, this filter will automatically reject (skip) all PRs that touch any of the files in `./.github/**` repo!
> We can maybe add `./bin/**` to this list, or any other file, folder that seems risky. That's easy to change at any point. Tbd...

### Detailed explanation 
1. The first and the most important step is to make explicit PR approvals mandatory for all forks (not just the first time contributors!)
    - **Note that this requires approval on every commit in a PR - which is great!** 🎉 
    
<details><summary>Expand to see the screenshot</summary>

image

</details>

2. Internal PRs don't require approvals - it's business as usual
    2a. `external-filter.yml` workflow will be skipped
    2b. Target workflow (with secrets) runs on `pull_request` only (as it did before). Since it is not a fork, the PR has access to all secrets.
3. External PR will just sit there waiting for an approval. If it doesn't get approved, nothing happens. Let's consider a scenario when we decide to approve it (after a careful inspection of the proposed changes):
    3a. `external-filter.yml` workflow will run and complete successfully (unless the PR touches files inside `./.github/**`
    3b. This, in turn, triggers a target workflow with secrets but from a safe context. The trigger is on `workflow_run` (completed). Since it is a safe context triggered internally, that particular run will have access to secrets.
    3c. Target workflow with secrets will also be triggered on `pull_request` but all runs will be skipped because the conditional we put in place detects this is a fork.

The best thing about this solution is that is safe.
Even if we don't detect a fork (let's say someone forgets to add that conditional in some target workflow with secrets) the external PR will run, but will not have access to secrets by design! The worst case scenario would be a bunch of failed tests because they didn't have secrets.

### Internal PR
```mermaid
stateDiagram-v2
    state "CI runs" as S1
    state "Workflow with secrets" as S2
    state "Filter workflow" as S3
    [*] --> S1: Internal
    S1 --> S2: (on) pull_request
    S1 --> S3: (on) pull_request
    S2 --> runs: Has access to secrets by default
    S3 --> [*]: skipped
```

### External PR
```mermaid
stateDiagram-v2
    state if_state <<choice>>
    state filter_state <<choice>>
    state "CI runs" as S1
    state "Workflow with secrets\n(run 1)" as S2
    state "Workflow with secrets\n(run 2)" as S3
    state "Filter workflow" as S4

    [*] --> Approved: External
    Approved --> if_state
    if_state --> S1: yes
    if_state --> [*]: no
    S1 --> S2: (on) pull_request
    S1 --> S4: (on) pull_request
    S2 --> WS1_skipped: detects it's a fork
    S4 --> filter_state: checks which files changed
    filter_state --> F_skipped: nope
    filter_state --> F_completed: yep
    F_completed --> S3: (on) workflow_run
    S3 --> runs: Creates safe environment that\nhas access to secrets
```
## Reading material
- https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
- https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

[^1]: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment

### DIFF
--- a/.github/workflows/drivers.yml
+++ b/.github/workflows/drivers.yml
@@ -7,6 +7,9 @@ on:
       - 'release-**'
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+  workflow_run:
+    workflows: [External Trigger Filter]
+    types: [completed]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
@@ -30,7 +33,9 @@ jobs:
 
   be-tests-athena-ee:
     needs: files-changed
-    if: github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true'
+    if: |
+      (github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true') &&
+      (github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success')
     runs-on: ubuntu-22.04
     timeout-minutes: 90
     env:
@@ -50,7 +55,9 @@ jobs:
 
   be-tests-bigquery-cloud-sdk-ee:
     needs: files-changed
-    if: github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true'
+    if: |
+      (github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true') &&
+      (github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success')
     runs-on: ubuntu-22.04
     timeout-minutes: 90
     env:
@@ -72,7 +79,9 @@ jobs:
 
   be-tests-druid-ee:
     needs: files-changed
-    if: github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true'
+    if: |
+      (github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true') &&
+      (github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success')
     runs-on: ubuntu-22.04
     timeout-minutes: 90
     env:
@@ -95,7 +104,9 @@ jobs:
 
   be-tests-googleanalytics-ee:
     needs: files-changed
-    if: github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true'
+    if: |
+      (github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true') &&
+      (github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success')
     runs-on: ubuntu-22.04
     timeout-minutes: 90
     env:
@@ -111,7 +122,9 @@ jobs:
 
   be-google-related-drivers-classpath-test:
     needs: files-changed
-    if: github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true'
+    if: |
+      (github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true') &&
+      (github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success')
     runs-on: ubuntu-22.04
     timeout-minutes: 90
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/drivers.yml
```diff
@@ -30,7 +33,9 @@ jobs:
 
   be-tests-athena-ee:
     needs: files-changed
-    if: github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true'
+    if: |
+      (github.event.pull_request.draft == false && needs.files-changed.outputs.backend_all == 'true') &&
```
nemanjaglumac, line 37: We need to always check for this, regardless of internal vs fork context.

.github/workflows/external-filter.yml
```diff
@@ -0,0 +1,17 @@
+name: External Trigger Filter
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - 'master'
```
nemanjaglumac, line 7: Only PRs (from forks) opened against `master` will be considered. All others will be skipped automatically.

.github/workflows/external-filter.yml
```diff
@@ -0,0 +1,17 @@
...
+    paths-ignore:
+      - '.github/**'
```
nemanjaglumac, line 9: Do we want to ignore changes in any other sensitive folders or files?
Maybe `bin/**`?

filipesilva (approved):
Took me a bit to wrap my head around everything here. My understanding is this:

- Background: we're ok with letting external PRs access secrets in CI runs
- Problem: GHA does give secrets to workflows triggered from external PRs
- Approach: indirectly trigger a workflow via `workflow_run`

The part about the "safe context" kinda sounds like there's something more safe about it, but there isn't. We just decided that we're ok with doing something that GH considers unsafe, and indirection is the mechanism by which we get around GH restrictions.

The part about needing approval to run CI is the only thing adding any amount of safety to the process. The bit about `paths-ignore` is just a backup. If a change that leaks secrets gets through the human approval, the secrets will be leaked.

I approve this PR assuming my understanding is correct (especially the background), and mostly on a technical basis: the code changes here achieve the goal of using indirection to pass secrets to external PR workflows.

nemanjaglumac:
> Problem: GHA does give secrets to workflows triggered from external PRs.

Actually, GHA doesn't give access to secrets on `pull_request`.
It does on `pull_request_target`, but that's very risky and requires some ugly hacks.

So, we're mimicking here what we have been doing all along - only manually.
Someone submits a PR, we review it. If we think it's good + safe, someone usually checked out that code from within the org and ran it in CI.

This workflow makes all that painless.

nemanjaglumac:
**Test 1**
https://github.com/nemanjaglumac/metabase/pull/2

First commit
- Had to wait for approval
image
- I approved it
- It didn't run `external-filter.yml` because it touched a file in a forbidden folder

Second commit
- Had to wait for another approval, which for me was the biggest worry and the most important thing to check. All good 🎉 
- I didn't approve it so it still sits there doing nothing
    - I think that GitHub deletes or closes PRs that are not approved after 30 days

nemanjaglumac:
**Test 2**
https://github.com/nemanjaglumac/metabase/pull/3

First commit
- Had to wait for an approval
- Approved
- Ran drivers tests on `pull_request`
    - All jobs have been skipped due to the conditional that checks for a fork
- Ran filter workflow
- Which triggered another drivers test run, but this time on `workflow_run`

🎉

alxnddr (approved):
Thanks for the example runs!

PR #34961, opened 2023-10-23
[Move dashcard to another tab] milestone 2 - undo logic

npretto:
### Description
epic: https://github.com/metabase/metabase/issues/34367 milestone 2

This adds undo for "move dashcard to another tab".

I wasn't sure how to test it properly so I added it to an existing e2e (not creating a new one to not slow down CI too much), if you have any suggestion I'm open to hear them

### Demo

https://github.com/metabase/metabase/assets/1914270/85a3c648-57fb-4524-8104-71d17b33f2c2


frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
```diff
@@ -23,6 +27,22 @@ export function DashCardTabMenu({ dashCardId }: DashCardTabMenuProps) {
...
+      dispatch(
+        addUndo({
+          message: t`Card moved to ${tabs.find(c => c.id === destTabId)?.name}`,
+          undo: true,
+          action: () => {
+            dispatch(undoMoveDashCardToTab({ dashCardId }));
+          },
+        }),
+      );
+    },
```
WiNloSt, line 42: `moveDashCardToTab` action seems to always have to call `addUndo` anyway. Should we call `addUndo` inside `moveDashCardToTab` instead then?

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -309,6 +318,24 @@ export const tabsReducer = createReducer<DashboardState>(
         }
         const dashCard = state.dashcards[dashCardId];
 
+        if (!dashCard.dashboard_tab_id) {
+          throw Error(
+            `MOVE_DASHCARD_TO_TAB was dispatched but dashCard.dashboard_tab_id (${dashCard.dashboard_tab_id}) is null`,
+          );
+        }
```
WiNloSt, line 325: Is there a case where this could happen?

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -322,6 +349,28 @@ export const tabsReducer = createReducer<DashboardState>(
       },
     );
 
+    builder.addCase(
+      undoMoveDashCardToTab,
+      (state, { payload: { dashCardId } }) => {
+        if (!state.dashboardId) {
+          throw Error(`state.dashboardId is null in undoMoveDashCardToTab`);
+        }
```
WiNloSt, line 357: Is there a case where this could happen?

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -91,45 +92,79 @@ describe("scenarios > dashboard > tabs", () => {
...
+        }).as("card1OriginalSize");
+      });
 
-    moveDashCardToTab({ tabName: "Tab 2" });
+      getDashboardCard(1).then(element => {
+        cy.wrap(element.offset()).as("card2OriginalPosition");
+      });
 
-    getDashboardCards().should("have.length", 0);
+      cy.log("move second to second tab first, then the first one");
```
WiNloSt, line 123: nit: improving clarity
```suggestion
      cy.log("move second card to second tab first, then the first card");
```

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
```diff
@@ -23,6 +27,22 @@ export function DashCardTabMenu({ dashCardId }: DashCardTabMenuProps) {
     return tabs.filter(t => t.id !== selectedTabId);
   }, [selectedTabId, tabs]);
 
+  const moveDashcard = useCallback(
+    (destTabId: number) => {
+      dispatch(moveDashCardToTab({ dashCardId, destTabId }));
```
WiNloSt, line 32: I'm not a fan of abbreviations this removes clarity from the code. Especially when its counterpart isn't abbreviated.

`sourceTabId` vs `destTabId` can we make this `destinationTabId`?

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
```diff
@@ -23,6 +27,22 @@ export function DashCardTabMenu({ dashCardId }: DashCardTabMenuProps) {
...
+      dispatch(
+        addUndo({
+          message: t`Card moved to ${tabs.find(c => c.id === destTabId)?.name}`,
```
WiNloSt, line 35: This is also not apparent. I'm not sure what `c` is supposed to mean, is it for Card? But it's finding on the `tabs`. Should this be
```suggestion
          message: t`Card moved to ${tabs.find(tab => tab.id === destTabId)?.name}`,
```
?

frontend/src/metabase/dashboard/actions/tabs.ts
npretto: It shouldn't but typescript doesn't know it, we don't allow `dashboard_tab_id!` so I added this assertion, do we have another pattern for this?

frontend/src/metabase/dashboard/actions/tabs.ts
npretto: same as other comment

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -309,6 +318,24 @@ export const tabsReducer = createReducer<DashboardState>(
         }
         const dashCard = state.dashcards[dashCardId];
 
+        if (!dashCard.dashboard_tab_id) {
+          throw Error(
```
kamilmielnik, line 324: We always use `throw new Error`. Even though it's [equivalent](https://es5.github.io/#x15.11.1), I'd go for consistency (applies to entire file).

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -25,6 +25,7 @@ import {
   updateDashboardCards,
   goToTab,
   moveDashCardToTab,
+  addTextBoxWhileEditing,
```
kamilmielnik, line 28: This is probably a global problem affecting all undos, but posting for completeness: undo won't work after leaving dashboard edit mode.

https://github.com/metabase/metabase/assets/6830683/02bd8de8-d395-4a43-ada6-422a7a526b16

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -91,45 +92,79 @@ describe("scenarios > dashboard > tabs", () => {
...
+
+      cy.log("undoing movement of second card");
+
+      cy.findAllByTestId("toast-undo").eq(0).findByRole("button").click();
+
+      goToTab("Tab 1");
+
+      getDashboardCards().should("have.length", 1);
+
+      cy.log("second card should be in the original position");
```
kamilmielnik, line 156: I'm not sure if this is acceptable, but the card won't always go back to the same location.

https://github.com/metabase/metabase/assets/6830683/b9755712-8ed2-48cd-a34d-4d11201f930b

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -91,45 +92,79 @@ describe("scenarios > dashboard > tabs", () => {
...
+
+      getDashboardCard().then(element => {
+        cy.get("@card2OriginalPosition").then(originalPosition => {
+          const position = element.offset();
+          // approximately to avoid possibly flakyness, we just want it to be in the same grid cell
```
kamilmielnik, line 161:
```suggestion
          // approximately to avoid possibly flakiness, we just want it to be in the same grid cell
```

frontend/src/metabase/dashboard/actions/tabs.ts
WiNloSt: Fair enough. I think `dashboard_tab_id` could be `null` when there's no tab in the dashboard as well. Indeed, it won't be the case obviously if `moveDashCardToTab` is called there're gotta be tabs. But I dislike the fact that we have to satisfy TypeScript and add guard blocks unnecessarily. It clutters the code and could give the impression that this could actually happen.

I'm not aware of other patterns but maybe we could cast `dashCard` to a new tab that has `dashboard_tab_id` as a number here?

@oisincoveney @uladzimirdev @kamilmielnik @deniskaber are you guys aware of patterns we do in our code to deal with this situation?

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -91,45 +92,79 @@ describe("scenarios > dashboard > tabs", () => {
     });
   });
 
-  it("should allow moving dashcards between tabs", () => {
-    visitDashboardAndCreateTab({
-      dashboardId: ORDERS_DASHBOARD_ID,
-      save: false,
-    });
+  it(
```
kamilmielnik, line 95: The "Move to tab" dropdown probably should not be shown when editing a link card:


https://github.com/metabase/metabase/assets/6830683/c123fbe3-2c7b-42fc-ad76-bd6090749d27

frontend/src/metabase/dashboard/actions/tabs.ts
kamilmielnik: We use [`checkNotNull`](https://github.com/metabase/metabase/blob/8c0313dba0226b75f8f2d6530c0d2d3aee6f1b9f/frontend/src/metabase/core/utils/types/types.ts#L11).

WiNloSt: 👆 That works. We could extract `dashboard_tab_id` to another variable, but it's a shame we couldn't just call it from `dashcard` anymore.

```ts
const dashboardTabId = checkNotNull(dashcard.dashboard_tab_id)
```

oisincoveney: Ah yep `checkNotNull` is the way to go. My other question then is - are we happy introducing a generic `TypeError` without a message? I like the fact that the error message is specific here.

oisincoveney: image

Ofc we have the stack trace so it's not a problem in terms of finding the problem

npretto: What do you think about an "assert" function like this?
```ts
export const assertNotNull = <T>(
  value: T | null | undefined,
  message?: string,
): asserts value is T => {
  if (value == null) {
    throw new TypeError(message);
  }
};

```

It wouldn't force us to make new variables and we can just put the assertions at the start of the function, in this case:
```diff
builder.addCase(
      _moveDashCardToTab,
      (state, { payload: { dashCardId, destinationTabId } }) => {
+       assertNotNull(dashCard.dashboard_tab_id);
        const { dashId } = getPrevDashAndTabs({ state });
+		assertNotNull(dashId);
-        if (dashId === null) {
-          throw Error(
-            `MOVE_DASHCARD_TO_TAB was dispatched but dashId (${dashId}) is null`,
-          );
-        }
        const dashCard = state.dashcards[dashCardId];

-        if (!dashCard.dashboard_tab_id) {
-          throw Error(
-            `MOVE_DASHCARD_TO_TAB was dispatched but dashCard.dashboard_tab_id -(${dashCard.dashboard_tab_id}) is null`,
-          );
-        }

        const dashboard = state.dashboards[dashId];

        if (!dashboard.dashCardTabMovements) {
```

npretto: I also read "assert not null" more as "I expect this to not be null, if it isn't, something is broken", while "check not null" doesn't give me the same idea

kamilmielnik: > What do you think about an "assert" function like this?

This might be good candidate for an RFC. For now `checkNotNull` will do.

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
npretto: Moved into a thunk, thanks!

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -299,25 +328,51 @@ export const tabsReducer = createReducer<DashboardState>(
...
-            `MOVE_DASHCARD_TO_TAB was dispatched but dashId (${dashId}) is null`,
-          );
-        }
+      _moveDashCardToTab,
+      (state, { payload: { dashCardId, destinationTabId } }) => {
         const dashCard = state.dashcards[dashCardId];
+        const dashboardId = checkNotNull(state.dashboardId);
+        const dashboard = state.dashboards[dashboardId];
+
+        if (!dashboard.dashCardTabMovements) {
```
uladzimirdev, line 337: modern js has `??=`

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -299,25 +328,49 @@ export const tabsReducer = createReducer<DashboardState>(
...
-        const { dashId } = getPrevDashAndTabs({ state });
-        if (dashId === null) {
-          throw Error(
-            `MOVE_DASHCARD_TO_TAB was dispatched but dashId (${dashId}) is null`,
-          );
-        }
+      _moveDashCardToTab,
+      (state, { payload: { dashCardId, destinationTabId } }) => {
         const dashCard = state.dashcards[dashCardId];
+        const dashboardId = checkNotNull(state.dashboardId);
```
EmmadUsmani, line 334: For consistency and to avoid repetition how about we use the `getPrevDashAndTabs` function used in other reducer cases in this file?

```typescript
const { dashId, prevDash } = getPrevDashAndTabs({ state })
```

We can also add the `checkNotNull` assertion inside its body to avoid repeating it, since we expect it to be non-null in all cases.

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -299,25 +328,49 @@ export const tabsReducer = createReducer<DashboardState>(
...
+    builder.addCase(
+      undoMoveDashCardToTab,
+      (state, { payload: { dashCardId } }) => {
+        const dashboardId = checkNotNull(state.dashboardId);
+        const dashboard = state.dashboards[dashboardId];
+
+        const dashCard = state.dashcards[dashCardId];
+        const movement = dashboard.dashCardTabMovements?.[dashCardId];
+
+        if (!movement) {
```
EmmadUsmani, line 367: Can also use `checkNotNull` here

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
EmmadUsmani: I'm not a huge fan of thunks because they distort the pure function style of redux and hide side effects. Without thunks you know that dispatching action X runs case X in the reducer, but with thunks any kind of side effects can happen. When thunks get too big and complex and start calling other thunks it becomes a nightmare to understand and maintain, like in [data-fetching.js]([url](https://github.com/metabase/metabase/blob/3398d6c4b85ae1fb343b3f71d82f32718cfc3bb6/frontend/src/metabase/dashboard/actions/data-fetching.js))

That's why originally with [`deleteTab`]([url](https://github.com/metabase/metabase/blob/3398d6c4b85ae1fb343b3f71d82f32718cfc3bb6/frontend/src/metabase/dashboard/components/DashboardTabs/use-dashboard-tabs.ts#L32)) I dispatched the `addUndo` separately, to make it explicit at the call site that there's a side effect.

To be fair though this use case isn't egregious so I don't think it's necessary to refactor, but it would be nice if we could avoid thunks more often going forward in favor of explicitly dispatching separate actions at the callsite (which itself can be a reusable function that isn't a redux action).

e2e/test/scenarios/dashboard/tabs.cy.spec.js
EmmadUsmani: Given that it's the same behavior when undoing a card deletion I think it's fine, but maybe worth double checking with Connor/Dawei.

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
WiNloSt: Has there been any discussion in our FE team regarding thunks since I'm not aware of it? This could be a good RFC if we haven't discussed it yet.

I agree `data-fetching.js` isn't a very good example when it comes to using thunks. But I was thinking about the relationship between actions. At a glance, if calling `deleteTab` or `moveTab` means we also need to call `addUndo` then I don't see the point of repeating `addUndo` in every call site.

However, that could become problematic if we ever decided to not show undo toast in certain places. And I do like code that looks consistent, knowing how you've done with `deleteTab`

I think calling `moveDashCardToTab` and `addUndo` side by side is better in this regard.

@npretto can you bring the original change back?

frontend/src/metabase/dashboard/actions/tabs.ts
npretto: Thanks, not sure why I left that out :)

e2e/test/scenarios/dashboard/tabs.cy.spec.js
npretto: We decided to add it to Milestone 3 of the epic

e2e/test/scenarios/dashboard/tabs.cy.spec.js
kamilmielnik: @npretto :arrow_up:

e2e/test/scenarios/dashboard/tabs.cy.spec.js
npretto: This too has been added to M3

e2e/test/scenarios/dashboard/tabs.cy.spec.js
npretto: @kamilmielnik another solution would be to make it appear on top of the other menu (will ask product today)
image
but it involves touching zIndexs and it would make sense to fix it in M3 with the other zIndex issue.

Do you think we could merge it as it is and fix this too in M3?

kamilmielnik: > Do you think we could merge it as it is and fix this too in M3?

Probably M3 but I'd double-check with product.

npretto: Also moved to M3 ✅

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -151,7 +161,7 @@ export const tabsReducer = createReducer<DashboardState>(
       (state, { payload: { tabId } }) => {
         const { dashId, prevDash, prevTabs } = getPrevDashAndTabs({ state });
         if (!dashId || !prevDash) {
-          throw Error(
+          throw new Error(
```
npretto, line 164: refactored this on all the file after a suggestion to keep it consistent with how we do in the rest the codebase

npretto:
Rebased after merge of M1, all comments should either have been addressed or added to a later milestone 🙏

frontend/src/metabase/dashboard/actions/tabs.ts
npretto: Had to revert putting `checkNotNull` in `getPrevDashAndTabs` as it made [this test](https://github.com/metabase/metabase/blob/9020a5cc254adf3828547bd26d0a2d495436b28c/e2e/test/scenarios/dashboard/x-rays.cy.spec.js#L175-L194) fail :(

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
```diff
@@ -23,6 +27,22 @@ export function DashCardTabMenu({ dashCardId }: DashCardTabMenuProps) {
     return tabs.filter(t => t.id !== selectedTabId);
   }, [selectedTabId, tabs]);
 
+  const moveDashcard = useCallback(
+    (destinationTabId: number) => {
+      dispatch(moveDashCardToTab({ dashCardId, destinationTabId }));
+      dispatch(
+        addUndo({
+          message: t`Card moved`,
```
WiNloSt, line 35: @cdeweyx @dhuniverse Not sure if this is the most updated toast message as the previous version isn't this one
```ts
t`Card moved to ${tabs.find(c => c.id === destTabId)?.name}`
```
But [it matches the spec](https://www.notion.so/metabase/Let-people-move-cards-between-dashboard-tabs-e797b321092b45a1bcbc4478d982a01e?pvs=4#1b2b364225a7477999c2b62a29b15e8d)
Screenshot 2023-10-30 at 1 15 59 PM


I found that this is very confusing when you're moving multiple cards between tabs. I'm not sure how we could make this better. But maybe something like showing the type of the card or the card name would make it clearer. e.g. **Text card moved**, **Link card moved**, **Heading card moved**, **Question moved** (or even the name of the question itself)

What do you think?

WiNloSt (approved):
code-wise looks good to me.

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
WiNloSt: Yeah, let's leave this as is until we hear from product since this is already matching the spec.

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -300,24 +312,46 @@ export const tabsReducer = createReducer<DashboardState>(
...
-          );
-        }
+      (state, { payload: { dashCardId, destinationTabId } }) => {
         const dashCard = state.dashcards[dashCardId];
+        const dashboardId = checkNotNull(state.dashboardId);
+        const dashboard = state.dashboards[dashboardId];
+
+        dashboard.dashCardTabMovements ??= {};
+
+        dashboard.dashCardTabMovements[dashCardId] = {
```
ranquild, line 322: This is hard to manage / support. We should get all the information required for the undo action before moving the card and use it for undo. No global state changes should be required.

ranquild (changes requested):

npretto:
@ranquild 
I drafted a solution with a thunk (is this what you meant by moving things in actions?)
The thunk also gets the original position/tab_id and passes it directly in the undo action to avoid having to store the movements in redux.
Before I clean it up and do the TODOs in the comments I left, is this the right direction?

frontend/src/metabase/dashboard/components/DashCard/DashCardTabMenu/DashCardTabMenu.tsx
cdeweyx: Good feedback, I'll start a thread to discuss with design. As for now feel free to merge and we can improve this case in M4.

npretto:
Ended up refactoring/fixing also the TODOs

ranquild (approved):

cdeweyx (approved):

PR #34967, opened 2023-10-23
[MLv2] Filters — Highlight filter column or segment

### LINKED ISSUE #34737: [MLv2] FE - Highlight selected columns and filters for existing filters

kulyk:
Closes #34737 — highlights filter's column/segment in the item picker

### To Verify

1. Go to `/admin/datamodel/segments` and create a segment for the sample Orders table
2. Add a filter using the segment from step 1, click the added filter clause, ensure the segment is highlighted
3. Add a regular filter on some column, click the added filter clause, and ensure that the column is highlighted

### Demo

demo

### DIFF
--- a/frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
+++ b/frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
@@ -13,6 +13,8 @@ import { isExpression as isLegacyExpression } from "metabase-lib/expressions";
 import LegacyFilter from "metabase-lib/queries/structured/Filter";
 import type LegacyQuery from "metabase-lib/queries/StructuredQuery";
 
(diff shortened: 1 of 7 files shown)


frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
```diff
@@ -60,6 +64,16 @@ export function FilterPicker({
     onClose?.();
   };
 
+  const checkItemIsSelected = useCallback(
+    (item: ColumnListItem | SegmentListItem) => {
+      return !!(
+        typeof filterIndex === "number" &&
```
ranquild, line 70: nit: `filterIndex != null`. `typeof` checks makes me think if we expect other types, e.g. a string

ranquild (approved):

PR #34969, opened 2023-10-23
Explain why linked filters and "From connected fields" don't work for native variables on the dashcard

### LINKED ISSUE #34570: No explanation why linked filters do not work with native variables
### Describe the bug

Regression of #14595

It's currently possible to create a Linked Filter which has no effect if the filter is wired up to a variable of a native query that is not a Field Filter.

Workaround: If using SQL, then make sure that all filters are Field Filter.

### To Reproduce

1. Native question > Sample Dataset - make single_date a Date filter, and field_filter a Field Filter of Products.Category
```
SELECT * FROM PRODUCTS WHERE 1=1
[[AND DATE(CREATED_AT)={{single_date}}]]
[[AND {{field_filter}}]]
```
2. Add to dashboard, create a single date filter and category filter and connect to the card, and set category filter as limited by single date filter:
image
3. Save dashboard, and set single date filter to 2023-10-10, which then would expect category filter to only show "Gizmo":
image

### Expected behavior

While it's possible to click the name of the linked filter in the sidebar, which will show how the linking will be done, that's not obvious that those are clickable, and the warning should probably always be shown, when the linked filter has no effect. image

### Logs

_No response_

```JSON
Tested master (c4812b4a3e)
```

### Severity

low but confuses UX of linked filters

### Additional context

_No response_

calherries:
Fixes #34570 by adding to the warning message on dashcards that are mapped to native question variables.

[Slack thread about product decisions](https://metaboat.slack.com/archives/C057T1QTB3L/p1697629797841749?thread_ts=1697469877.162559&cid=C057T1QTB3L)

The message differs slightly depending on whether the filter is a date filter or not.

For non-date filters:
image

For date filters we can omit the "takes a single value" part because date filters always take a single value:
image

### DIFF
--- a/frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
+++ b/frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
@@ -28,7 +28,6 @@ import {
 import { isActionDashCard } from "metabase/actions/utils";
 import { Ellipsified } from "metabase/core/components/Ellipsified";
 import Question from "metabase-lib/Question";
-import { isDateParameter } from "metabase-lib/parameters/utils/parameter-type";
 import { isVariableTarget } from "metabase-lib/parameters/utils/targets";
 
 import { normalize } from "metabase-lib/queries/utils/normalize";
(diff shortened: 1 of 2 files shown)


uladzimirdev (commented):
From code perspective LGTM, as for product correctness I'd wait someone else to review

cdeweyx (approved):

PR #34975, opened 2023-10-23
[MLv2] Filters — Add unit tests for `TimeFilterPicker`

kulyk:


frontend/src/metabase/common/components/FilterPicker/TimeFilterPicker/TimeFilterPicker.unit.spec.tsx
```diff
@@ -0,0 +1,340 @@
+import moment from "moment";
```
iethree, line 1: will dayjs work here?

frontend/src/metabase/common/components/FilterPicker/TimeFilterPicker/TimeFilterPicker.unit.spec.tsx
```diff
@@ -0,0 +1,340 @@
...
+      onChange={onChange}
+      onBack={onBack}
+    />,
+  );
+
+  function getNextFilterParts() {
+    const lastCall = onChange.mock.calls[onChange.mock.calls.length - 1];
+    const [filter] = lastCall;
+    return Lib.timeFilterParts(query, 0, filter);
+  }
```
iethree, line 110: this is lovely ❤️

iethree (approved):
This is great. Can we also test a few more cases:

- invalid user input like 26:62
- invalid filter value like greater than 26:62
- swapping the column when there's an existing filter

kulyk:
Thanks for the excellent suggestions, @iethree ❤️ I think I've addressed everything, please take another look

iethree (approved):
💖

PR #24503, opened 2022-08-02
Fix column duplication when adding a new breakout

ranquild:
Fixes https://github.com/metabase/metabase/issues/22382

The problem is that:
- We don't remove entries from `table.columns` when removing a column 
- We don't check for existing entries when adding new columns

So you could end up with duplicate columns if you remove and add a breakout for the same column. This PR fixes the issue by removing entries for the same column when adding a new column.

How to test:
- Follow the steps from the issue
- There is also a cypress test that does exactly the same

PR #16761, opened 2021-06-24
Fix newly added columns don't appear after some columns are hidden

kulyk:
For native questions, after hiding some columns using the "Visible Columns" list under visualization settings, any new columns added to the query will not be shown.

Each question has a `table.columns` property in `visualization_settings`; it defines which columns are visible and which are hidden. For structured queries (simple and custom questions), the `table.columns` property is synced with the query itself. So if a new column is added or an existing one is removed, visualization settings stay in sync. It's impossible to do for native queries as Metabase doesn't parse SQL. This PR adds columns sync for native queries based on query _results_, rather than the query.

Note: don't be afraid of a huge diff in the `Question.syncColumnsAndSettings` method. Before this PR it was a huge method doing the sync only for `StructuredQuery` instances. This PR extracted the `_syncStructuredQueryColumnsAndSettings` from there ([commit](https://github.com/metabase/metabase/pull/16761/commits/1b82769ab3a13fb4ba42fa35f836a435a33425b9)), and then added a similar method for the `NativeQuery` instances. The extracted method is the biggest diff out there, but method's implementation is left untouched.

Fixes #15393, fixes #7612

### To Verify

1. Ask a question → Native question
2. Fill in `select 1 as visible, 2 as hidden`
3. Run the query, so there are some results
4. Click the "Settings" button in the bottom left
5. Press the `X` icon next to the column named `hidden` to hide it
6. Add `3 as added` to the query, so it's `select 1 as visible, 2 as hidden, 3 as hidden`
7. Run the query
8. Ensure the `added` column appears in the visualization settings sidebar and it's visible. Ensure the `hidden` column remains hidden

### Demo

**Before**

https://user-images.githubusercontent.com/17258145/123629983-a8a57900-d81d-11eb-8dbc-7d54732cf48e.mov

**After**

https://user-images.githubusercontent.com/17258145/123629874-7d228e80-d81d-11eb-8cc5-7fbb727e1dad.mov

PR #34976, opened 2023-10-23
ensure that column is removed from table.columns when updating question

### LINKED ISSUE #33348: Question in table view: Removing a field rearranges the entire table columns
### Describe the bug

When ordering columns in a specific way in table view, the column order is lost if omitting one of the fields from the query results. This doesn't happen with additive changes (new fields are always added at the right end of the table).

### To Reproduce

1. Using sample data, create a new question using query editor: orders joining products, select only a subset of the fields in orders, e.g. remove subtotal and tax
2. Click visualize
3. Rearrange the columns in a specific way, e.g. product-vendor, product-category, product-title, etc.
4. Using query editor now add the subtotal field again, re-run the query. The subtotal field is added to the table on the right. Column order is the same.
5. Using query editor now de-select subtotal again and re-run the query
6. The previous column order is lost

### Expected behavior

Just remove it without changing the column sorting

### Logs

_No response_

```JSON
- Chrome 115.0.5790.170 (Official Build) (arm64)
- Mac OS X 13.4.1 (c) (22F770820d)
- Tested with hash bec9c95 on stats
```

### Severity

Annoying as it makes iterating on table reports painful

### Additional context

_No response_

npfitz:
Closes https://github.com/metabase/metabase/issues/33348

### Description
The reason for this issue is that syncing `table.columns` to fields in the query only happened when adding a field to a query, but not when removing. Looking at the history, it looks like this was adjusted back in https://github.com/metabase/metabase/pull/24503, which ensured that duplicate fields were not re-added, and in https://github.com/metabase/metabase/pull/16761, where we simply moved the logic (`Note: don't be afraid of a huge diff` kills me every time haha). The issue was that removing a field from the query cause the `isValid` logic to return false for `table.columns`, since we would have a column setting that didn't have a query column.

This change makes it so that any columns identified as being removed from the query are removed from `table.columns`, as well as leaving the logic for adding any new columns in place. This should ensure that the `table.columns` viz setting remains vaild while updating the query.

### How to verify

1. Create a question (orders will work), but don't include all fields in the query. Visualize
2. Go to settings, re-arrange the columns (this makes us save a `table.columns` setting
3. Add a new field to the query and visualize. You should see your new field appended to the end
4. Go back to the notebook editor and remove that field (or any field). Your field order should be preserved

### Demo
9sBvfIEUBL

### Checklist

### DIFF
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -775,13 +775,12 @@ class Question {
     const tableColumns = this.setting("table.columns");
     if (
       tableColumns &&
-      addedColumnNames.length > 0 &&
-      removedColumnNames.length === 0
+      (addedColumnNames.length > 0 || removedColumnNames.length > 0)
     ) {
       return this.updateSettings({
         "table.columns": [
           ...tableColumns.filter(
-            column => !addedColumnNames.includes(column.name),
+            column => !removedColumnNames.includes(column.name),
           ),
           ...addedColumnNames.map(name => {
             const dimension = query.columnDimensionWithName(name);
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -780,7 +780,9 @@ class Question {
       return this.updateSettings({
         "table.columns": [
           ...tableColumns.filter(
-            column => !removedColumnNames.includes(column.name),
+            column =>
+              !removedColumnNames.includes(column.name) &&
+              !addedColumnNames.includes(column.name),
           ),
           ...addedColumnNames.map(name => {
             const dimension = query.columnDimensionWithName(name);
--- a/frontend/src/metabase/query_builder/actions/core/updateQuestion.unit.spec.ts
+++ b/frontend/src/metabase/query_builder/actions/core/updateQuestion.unit.spec.ts
@@ -23,9 +23,6 @@ import {
   createMockStructuredDatasetQuery,
   createMockStructuredQuery,
   createMockTable,
-  createMockCard,
-  createMockVisualizationSettings,
-  createMockTableColumnOrderSetting,
 } from "metabase-types/api/mocks";
 import {
   createSampleDatabase,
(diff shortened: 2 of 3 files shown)


frontend/src/metabase-lib/Question.ts
```diff
@@ -775,13 +775,12 @@ class Question {
...
       tableColumns &&
-      addedColumnNames.length > 0 &&
-      removedColumnNames.length === 0
+      (addedColumnNames.length > 0 || removedColumnNames.length > 0)
     ) {
       return this.updateSettings({
         "table.columns": [
           ...tableColumns.filter(
-            column => !addedColumnNames.includes(column.name),
+            column => !removedColumnNames.includes(column.name),
```
kulyk, line 783: Should we maybe extend the original check? It ensured we wouldn't duplicate columns when merging `tableColumns` and `addedColumnNames.map(...)` result. I think we might want to keep that? I'm thinking of something like:

```jsx
tableColumns.filter(
  column =>
    !addedColumnNames.includes(column.name) &&
    !removedColumnNames.includes(column.name)
)
```

frontend/src/metabase/query_builder/actions/core/updateQuestion.unit.spec.ts
```diff
@@ -811,6 +814,69 @@ describe("QB Actions > updateQuestion", () => {
     });
   });
 
+  describe("table column sync", () => {
```
kulyk, line 817: I think we should be adding tests to `metabase-lib`, closer to the sync columns method
Maybe [here](https://github.com/metabase/metabase/blob/768ba0577fccc8c4ec9cc16addf05fbdf16af4d9/frontend/test/metabase/lib/dataset.unit.spec.js#L29) would be a better place?

frontend/src/metabase/query_builder/actions/core/updateQuestion.unit.spec.ts
```diff
@@ -811,6 +814,69 @@ describe("QB Actions > updateQuestion", () => {
...
+          enabled: true,
+        }),
+        createMockTableColumnOrderSetting({
+          name: "TOTAL",
+          fieldRef: ["field", ORDERS.TOTAL, {}],
+          enabled: true,
+        }),
+        createMockTableColumnOrderSetting({
+          name: "QUANTITY",
+          fieldRef: ["field", ORDERS.QUANTITY, {}],
```
kulyk, line 832: I'm not sure we can have field references looking like that

Usually, they either have at least one option:

```js
[ "field", ORDERS.CREATED_AT, { "temporal-unit": "month" } ]

[ "field", "TOTAL", { "base-type": "type/Float" } ]
```

or no options at all:

```js
[ "field", ORDERS.TOTAL, null ]
```

Could we please replace `{}` with `nulls` here so it's a bit closer to reality?

kulyk (commented):
Looks good to me overall! I've also tried it out and didn't spot anything unexpected. I'm not sure if we should be replacing the `!addedColumnNames` check or rather extending it, and I have a couple of comments about tests, please take a look

frontend/src/metabase-lib/Question.ts
kulyk: @ranquild would appreciate if you could give it a look; maybe you still remember what was happening in #24503

ranquild: There is e2e repro https://github.com/metabase/metabase/pull/24503/files#diff-e3b32eb50834bfcb0d107c2b06f89b004d6e5c313ef210db40c12922a1188ddcR191

npfitz: @kulyk We could just extend the original check, though it really shouldn't be necessary. The reason the columns were still there was because we weren't removing them. But extending the check shouldn't cause any hard either

frontend/src/metabase/query_builder/actions/core/updateQuestion.unit.spec.ts
npfitz: That... would make way more sense haha. Will move those

frontend/src/metabase/query_builder/actions/core/updateQuestion.unit.spec.ts
npfitz: Yeah, this felt a bit weird. I think I did this for the type check, but I'll take another look and see if I can sort it out 👍

ranquild:
Removing myself from reviews since I don't have anything to add to the @kulyk 's comments

frontend/src/metabase-lib/Question.ts
npfitz: I added it back in. It shouldn't hurt anything anyways.

frontend/test/metabase/lib/dataset.unit.spec.js
```diff
@@ -93,6 +94,80 @@ describe("metabase/util/dataset", () => {
...
+        });
+
+      const newQuestion = prevQuestion
+        .query()
+        .removeField(2)
+        .question()
+        .syncColumnsAndSettings(prevQuestion);
+
+      expect(prevQuestion.setting("table.columns")).toHaveLength(3);
+      expect(newQuestion.setting("table.columns")).toHaveLength(2);
```
kulyk, line 135: Could we assert which `table.columns` are left in `newQuestion`? Things would be more reliable in that case — e.g., the test would still pass if we remove a wrong column

frontend/test/metabase/lib/dataset.unit.spec.js
```diff
@@ -93,6 +94,80 @@ describe("metabase/util/dataset", () => {
...
+        });
+
+      const newQuestion = prevQuestion
+        .query()
+        .addField(["field", PRODUCTS.VENDOR, null])
+        .question()
+        .syncColumnsAndSettings(prevQuestion);
+
+      expect(prevQuestion.setting("table.columns")).toHaveLength(2);
+      expect(newQuestion.setting("table.columns")).toHaveLength(3);
```
kulyk, line 169: Same thoughts as in the previous tests

kulyk (commented):
Looks good! Please consider updating the final assertions

kulyk (approved):
Don't want to be blocking you

PR #34983, opened 2023-10-23
Auditv2 disable audit db sync

escherize:
We want to disable db syncs for the audit-db. This alters `check-and-schedule-tasks-for-db!` to skip the creation of the sync job (`"scan/analyze"`) for the audit-db. The scan/analyze info already exists because of the way we load the instance analytics content.

I considered making the `metadata_sync_schedule` be about 2 billion years, since that is possible in a valid cron string. But that doesn't seem right, because the job would be in there, and it isn't needed at all.

So, we skip creating the sync job by partitioning jobs into `jobs-to-create` and `jobs-to-skip` , for the one we skip, we print out a log about not creating the job, and the codepath creating jobs is left untouched.

## How do you know this works?
- On app startup, there is a log message about _not_ installing a sync job for the audit-db. It is in red, too. ✨
- No sync messages occur in the logs.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/audit_db_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/audit_db_test.clj
@@ -5,6 +5,7 @@
             [metabase-enterprise.audit-db :as audit-db]
             [metabase.core :as mbc]
             [metabase.models.database :refer [Database]]
+            [metabase.task :as task]
             [metabase.test :as mt]
             [toucan2.core :as t2]))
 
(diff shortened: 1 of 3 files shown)


enterprise/backend/test/metabase_enterprise/audit_db_test.clj
```diff
@@ -50,6 +51,18 @@
...
+(deftest audit-db-does-not-have-scheduled-syncs
+  (mt/test-drivers #{:postgres :h2 :mysql}
+    (with-audit-db-restoration
+      (is (= :metabase-enterprise.audit-db/installed (audit-db/ensure-audit-db-installed!)))
+      (let [sync-job-info (task/job-info "metabase.task.sync-and-analyze.job")
+            db-has-sync-job-trigger? (fn [db-id]
+                                       (contains?
+                                        (set (map #(-> % :data (get "db-id")) (:triggers sync-job-info)))
+                                        db-id))]
+        (is (db-has-sync-job-trigger? 1))
```
noahmoss, line 63: What's the purpose of this assertion?

noahmoss (approved):

enterprise/backend/test/metabase_enterprise/audit_db_test.clj
escherize: To make sure we are looking at the right thing. But it's apparent in hindsight

Issue #34915, opened 2023-10-20
Modify can-write on /databases to disable changes to the Audit database


luizarakaki:
This should disable any user (including admins) from modifying the AuditDB.

Prevents:
- change any auditdb property like name, description, connection details etc
- trigger a manual sync or scanning
- turn model caching on/off
- delete the database
- turn model actions on

PR #34987, opened 2023-10-23
Add conditional on can_write to prevent writing to the audit-db

adam-james-v:
Closes: #34915

Implement `can-write` method for the database model that returns false if the database in question is the Internal audit database.

This should prevent all users from modifying details about the audit db.


escherize (approved):
Looks good to me! It will clearly reject any can-write? check for the audit db in both arities.

luizarakaki:
@adam-james-v tested renaming the db, triggering sync, scanning, discard values. Everything was blocked as expected.

But I could delete the Internal Metadata Database. I think we should block this too

enterprise/backend/test/metabase_enterprise/audit_app/api/database_test.clj
```diff
@@ -0,0 +1,20 @@
...
+   [metabase.test :as mt]))
+
+(deftest audit-db-unmodifiable-test
+  (premium-features-test/with-premium-features #{:audit-app}
+    (audit-db-test/with-audit-db-restoration
+      (audit-db/ensure-audit-db-installed!)
+      (testing "Neither admin nor regular users can modify the audit database"
+        (doseq [[verb path] [[:post "database/%d/unpersist"]
+                             [:put "database/%d"]
+                             [:get "database/%d/syncable_schemas"]]
```
escherize, line 17: I think you can add `[:delete "database/%d"]` to this list to get the failing repro

escherize:
> But I could delete the Internal Metadata Database.

@luizarakaki Do you know which route you used? Was it `DELETE /api/database/:id`? That info helps make things easier to fix

luizarakaki:
Pretty sure it is the DELETE method

luizarakaki (approved):
delete blocked as expected

escherize (approved):
✅

PR #34991, opened 2023-10-23
Add `prettify-native-form` driver multimethod and refactor `mdb.query/format-sql`

### LINKED ISSUE #34235: Incorrect dialect is chosen for ClickHouse pretty SQL formatting

lbrdnk:
Closes #34235

### Description

This PR extends driver with new multimethod, `driver/prettify-native-form`, which is used eg. to pretty format output of `dataset/native` endpoint. Also `mdb.query/format-sql` was refactored to use this method.

Even if we do not currently need to format nosql native code, I've decided to put the method in driver namespace, so it could be used as identity with nosql native code. This way we can avoid branchings or `ignore-exceptions` calls as previously in `/dataset/native` and underlying formatting logic. 

With this method in place, external driver developers can plug in their own formatters or reuse our formatting logic. To find out how, read `driver/prettify-native-form`'s docstring.

### How to verify

Multiple tests were modified during the refactor. Run those and see the results.

### Checklist


docs/developers-guide/driver-changelog.md
```diff
@@ -70,6 +70,9 @@ title: Driver interface changelog
   removed in 0.51.0 or later. You can easily implement `metabase.driver/db-default-timezone` directly, and use
   `metabase.driver.sql-jdbc.execute/do-with-connection-with-options` to get a `java.sql.Connection` for a Database.
 
+- `metabase.driver/prettify-native-form` was added to enable driver developers use native form formatting
```
metamben, line 73: I think we should use the word prettify or something similar so as not to confuse things with other formatting like HoneySQL to SQL string.

docs/developers-guide/driver-changelog.md
```diff
@@ -70,6 +70,9 @@ title: Driver interface changelog
...
+  specific to their driver. For details see the issue #34235.
```
metamben, line 74: I suggest referring to this PR instead of the issue.

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -445,3 +446,7 @@
     (when-not (str/blank? dataset-id)
       (convert-dataset-id-to-filters! database dataset-id))
     database))
+
+(defmethod driver/prettify-native-form :bigquery-cloud-sdk
+  [_ native-form]
+  (sql.u/format-sql-and-fix-params :mysql native-form))
```
metamben: We could call `(driver/prettify-native-form :mysql native-form)`, right? That would require less intimate knowledge of how pretty printing is implemented.

modules/drivers/oracle/src/metabase/driver/oracle.clj
```diff
@@ -45,6 +45,10 @@
                               :convert-timezone true}]
   (defmethod driver/database-supports? [:oracle feature] [_driver _feature _db] supported?))
 
+(defmethod driver/prettify-native-form :oracle
+  [_ native-form]
+  (sql.u/format-sql-and-fix-params :plsql native-form))
```
metamben, line 50: The same. I stop here, and won't repeat this for the other drivers...

src/metabase/db/query.clj
```diff
@@ -23,50 +23,19 @@
...
-  For mongo queries, return as is since it's already in a nice json-like format."
- ([sql]
-  (format-sql sql (mdb.connection/db-type)))
- ([sql db-type]
-  (if (isa? driver.impl/hierarchy db-type :sql)
-    (fix-sql-params (format-sql* sql db-type))
-    sql)))
+  "Return a nicely-formatted version of a `query` string with current application db driver formatting. If you need
+  to use specific driver formatting use [[driver/prettify-native-form]] instead."
+  [sql]
```
metamben, line 37: Is this `native-query-text` or something similar?

src/metabase/driver.clj
```diff
@@ -641,6 +641,35 @@
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
+(defmulti prettify-native-form
+  "Pretty-format native form presumably coming from compiled query.
+  Used eg. in API endpoint `/dataset/native`, to present user with nicely formatted query.
```
metamben, line 646:
```suggestion
  Used eg. in the API endpoint `/dataset/native`, to present the user with a nicely formatted query.
```

src/metabase/driver.clj
```diff
@@ -641,6 +641,35 @@
...
+
+    Function in question is implemented in a way, that driver developers, who would like to use it implementing
+    this multimethod, can:
+    - Avoid implementing it completely, if their driver keyword representation corresponds to key in
+      [[metabase.driver.sql.util/dialects]] (eg. `:postgres`).
+    - Ignore implementing it, if it is sufficient to format their drivers native form with dialect corresponding
+      to `:standardsql`'s value from the dialects map (eg `:h2`).
+    - Use [[metabase.driver.sql.util/format-sql]] in this method's implementation, providing dialect keyword
+      representation that corresponds to to their driver's formatting (eg. `:sqlserver` uses `:tsql`).
+    - Completly reimplement this method with their special formatting code."
```
metamben, line 664: Can you formulate this as a description how to use and extend this multimethod? There you should explain if and when and how to use `metabase.driver.sql.util/format-sql`. Note that for readers of this code it's probably not very interesting to know that `format-sql` existed before this and I doubt driver implementers used it.

src/metabase/driver/sql.clj
```diff
@@ -47,6 +48,10 @@
   [driver query]
   (sql.qp/mbql->native driver query))
 
+(defmethod driver/prettify-native-form :sql
+  [driver native-form]
+    (sql.u/format-sql-and-fix-params driver native-form))
```
metamben, line 53: Unexpected indentation.

metamben (changes requested):

metamben (approved):

Issue #34686, opened 2023-10-16
Add custom warning message when leaving creating a new notebook question


kamilmielnik:
- When implementing this, please take this comment into consideration: https://github.com/metabase/metabase/pull/34658#discussion_r1364018288

PR #35003, opened 2023-10-24
Fix - Unsaved changes warning shown when going back to notebook after visualizing the question

### LINKED ISSUE #35000: Unsaved changes warning shown when going back to notebook after visualizing the question
### Describe the bug

Unsaved changes warning shown when going back to notebook after visualizing the question.
Related to #33749

### To Reproduce

1. Open an existing notebook question or notebook-based model
2. Change something via notebook editor
3. Visualize the question
4. Open notebook editor again
5. Unsaved changes warning is shown

### Expected behavior

Unsaved changes warning should not be shown

`master`

### Severity

P2

kamilmielnik:
Closes #35000

### Description

`/question/notebook` & `/model/notebook` routes were not taken into account.

This PR also includes [small refactoring](https://github.com/metabase/metabase/pull/35003/commits/64088717eebda0b1ad9bc1469d0f2d4c6ef62639) because `waitForSave*Question*ToBeEnabled` applies also for **models** in notebook view (but not in query or metadata view), which I didn't initially know.

### How to verify

Tests are green.
2 new tests marked with `(metabase#35000)` fail in `master`.

### DIFF
--- a/frontend/src/metabase/query_builder/utils.unit.spec.ts
+++ b/frontend/src/metabase/query_builder/utils.unit.spec.ts
@@ -104,6 +104,11 @@ const runQuestionLocation = createMockLocation({
   hash: `#${serializeCardForUrl(nativeCard)}`,
 });
 
+const runQuestionEditNotebookLocation = createMockLocation({
+  pathname: "/question/notebook",
+  hash: `#${serializeCardForUrl(nativeCard)}`,
+});
+
 describe("isNavigationAllowed", () => {
   describe("when there is no destination (i.e. it's a beforeunload event)", () => {
     const destination = undefined;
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/query_builder/utils.unit.spec.ts
```diff
@@ -307,6 +333,14 @@ describe("isNavigationAllowed", () => {
       ).toBe(true);
     });
 
+    it("allows to run the model and then edit it again", () => {
```
WiNloSt, line 336: This new test is in `when creating new model` block, but there's no new test in `when creating a new structured question` should we do the same?

I was wondering because there are 2 new tests for models, but only 1 for questions

WiNloSt (approved):
Looks good to me.

frontend/src/metabase/query_builder/utils.unit.spec.ts
kamilmielnik: It's because #34686 is not implemented yet.

PR #35007, opened 2023-10-24
[MLv2] Filters — Fix nested native question + expression E2E test 

kulyk:
Fixes an incorrect field reference in of the E2E test:

```js
{
  "source-table": "card__1", // <— native query
  "expressions": {
     "Total100": [ "+", [ "field", 1, null ], 100 ]
   }
}
```

It should be using the column name when referring to a column from `source-card`:

```js
[ "+", [ "field", "TOTAL", { "base-type": "..." } ], 100 ]
```


metamben (approved):
LGTM

PR #35009, opened 2023-10-24
Add e2e tests for date shortcuts and exclude date filters

ranquild:
Found that exclude hour filters didn't work because they use a different expression format, fixed

### DIFF
--- a/frontend/src/metabase-lib/filter.ts
+++ b/frontend/src/metabase-lib/filter.ts
@@ -261,9 +261,9 @@ export function specificDateFilterClause(
   { operator, column, values }: SpecificDateFilterParts,
 ): ExpressionClause {
   const hasTime = values.some(hasTimeParts);
-  const stringValues = hasTime
-    ? values.map(value => dateToDateTimeString(value))
-    : values.map(value => dateToDateString(value));
+  const serializedValues = hasTime
+    ? values.map(value => serializeDateTime(value))
+    : values.map(value => serializeDate(value));
 
   const minuteBucket = hasTime
     ? findTemporalBucket(query, stageIndex, column, "minute")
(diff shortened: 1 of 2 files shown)


frontend/src/metabase-lib/filter.ts
```diff
@@ -762,16 +763,16 @@ function relativeDateFilterPartsWithOffset(
...
 }
 
-function excludeDatePartToString(
+function serializeExcludeDatePart(
   value: number,
   bucketName: ExcludeDateBucketName,
-): string {
-  const date = moment();
+): ExpressionArg {
+  if (bucketName === "hour-of-day") {
```
ranquild, line 770: The bug

frontend/src/metabase-lib/filter.ts
```diff
@@ -786,18 +787,24 @@ function excludeDatePartToString(
   return date.format(DATE_FORMAT);
 }
 
-function stringToExcludeDatePart(
-  value: string,
+function deserializeExcludeDatePart(
+  value: ExpressionArg | ExpressionParts,
   bucketName: BucketName,
 ): number | null {
+  if (bucketName === "hour-of-day") {
```
ranquild, line 794: The bug fix

frontend/src/metabase-lib/filter.ts
```diff
@@ -260,9 +261,9 @@ export function specificDateFilterClause(
   { operator, column, values }: SpecificDateFilterParts,
 ): ExpressionClause {
   const hasTime = values.some(hasTimeParts);
-  const stringValues = hasTime
-    ? values.map(value => dateToDateTimeString(value))
-    : values.map(value => dateToDateString(value));
+  const serializedValues = hasTime
```
ranquild, line 264: Improved naming

kulyk (approved):

PR #35017, opened 2023-10-24
left align dashcard actions on small cards on the left side of the da…

npretto:
Milestone 3 of https://github.com/metabase/metabase/issues/34367
### Description

Cards too small and with too many dashcard actions had the menu going outside of the viewport

### How to verify

Resize a card to 2x2 and drag it to the far left of the dashboard, actions should be left aligned

### Demo

https://github.com/metabase/metabase/assets/1914270/2bffae22-044c-4560-bf8f-469a544782d3


frontend/src/metabase/dashboard/components/DashCard/DashCard.styled.tsx
```diff
@@ -41,13 +41,16 @@ export const DashCardRoot = styled.div<DashCardRootProps>`
...
+export const DashboardCardActionsPanel = styled.div<{ leftAlign: boolean }>`
   padding: 0.125em 0.25em;
   position: absolute;
   background: white;
   transform: translateY(-50%);
   top: 0;
-  right: 20px;
+
+  right: ${props => (props.leftAlign ? "initial" : "20px")};
+  left: ${props => (props.leftAlign ? "10px" : "initial")};
```
npretto, line 52: I set it to 10 here, instead of 20, as this case is only for small cards and I feel it looks better and makes it easier to reach for the buttons on the far right

frontend/src/metabase/dashboard/components/DashCard/DashCard.tsx
```diff
@@ -250,8 +250,12 @@ function DashCard({
 
   const renderDashCardActions = useCallback(() => {
     if (isEditingDashboardLayout) {
+      // to avoid having actions buttons outside of the viewport
+      const leftAlignActions = dashcard.size_x <= 3 && dashcard.col <= 2;
```
WiNloSt, line 254: Is this logic confirmed with product? I'm quoting the spec from the product doc
> Left align the mini-toolbar to the card left edge if the card width is less than the total width of the mini-toolbar and the margin.

This looks simpler than trying to get the element width, but I just want to be sure.

npretto:
Closing this as I created a new on a graphite stack: https://github.com/metabase/metabase/pull/35241

PR #35019, opened 2023-10-24
POC wireit

npretto:
### Description

This is a poc to see how a build system with caching could help us in metabase

### How to verify
It should build the drivers the first time you run test-cypress-open on a fresh clone
- `git clone --depth=1 -b wireit-poc  git@github.com:metabase/metabase.git`
- `cd metabase`
- `yarn` (this could? be included as a dependency too in some way)
- `yarn test-cypress-open`

If files in modules/drivers change and we run `test-cypress-open` again, it should rebuild the drivers

### DIFF
--- a/package.json
+++ b/package.json
@@ -277,6 +277,7 @@
     "webpack-dev-server": "^4.15.0",
     "webpack-notifier": "1.15.0",
     "webpack-stats-plugin": "^1.1.3",
+    "wireit": "^0.14.1",
     "xlsx": "^0.18.5",
     "yaml-lint": "~1.6.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -360,10 +362,44 @@
...
+      "command": "bash ./bin/build-drivers.sh",
+      "files": [
+        "modules/drivers/",
+        "!modules/drivers/*/target"
+      ],
+      "output": [
+        "resources/modules"
+      ]
+    },
+    "type-check": {
```
npretto, line 388: i tried it with type-check too, time wise the savings weren't that big as build:cljs is probably very fast

PR #35021, opened 2023-10-24
Make `SESSION_COOKIE_SAMESITE` a setting

johnswanson:
Considerations here:

- I didn't want to break the existing behavior, which allows the env var to be case insensitive. This required a bit of weirdness in the getter/setter - we get the *raw* value, then normalize it to a keyword, then make sure it's one of the valid options.

- I'm not sure what the best practice is regarding i18n - I just went off the existing code, but I'm not entirely sure whether I need to also add translations to the `locales/*.po` files, and how I get those translations if so (err... chatGPT?).

- permissions: I'm not exactly sure how we decide between the `:settings-manager` and `:admin` settings. I went with `:settings-manager` here.

- I'm not sure whether the frontend has its own definitions of possible values for set-based settings, or if I need to expose that somehow?

The existing tests were good, easily adapted to the new behavior, and caught the above-mentioned backwards incompatibility issue with case insensitivity.

- Related: https://github.com/metabase/metabase-private/issues/182

### DIFF
--- a/src/metabase/config.clj
+++ b/src/metabase/config.clj
@@ -56,8 +56,7 @@
 
 ;; separate map for EE stuff so merge conflicts aren't annoying.
 (def ^:private ee-app-defaults
-  {:embed-max-session-age      "1440"   ; how long a FULL APP EMBED session is valid for. One day, by default
-   :mb-session-cookie-samesite "lax"})
+  {:embed-max-session-age      "1440"}) ; how long a FULL APP EMBED session is valid for. One day, by default
 
 (alter-var-root #'app-defaults merge ee-app-defaults)
 
@@ -132,18 +131,6 @@
   ^{:doc "A string that contains identifying information about the Metabase version and the local process."}
   mb-version-and-process-identifier
   (format "%s [%s]" mb-app-id-string local-process-uuid))
-
-(defn- mb-session-cookie-samesite*
-  []
-  #_{:clj-kondo/ignore [:discouraged-var]}
-  (let [same-site (str/lower-case (config-str :mb-session-cookie-samesite))]
-    (when-not (#{"none", "lax", "strict"} same-site)
-      (throw (ex-info "Invalid value for MB_SESSION_COOKIE_SAMESITE" {:mb-session-cookie-samesite same-site})))
-    (keyword same-site)))
-
-(def ^Keyword mb-session-cookie-samesite
-  "Value for session cookie's `SameSite` directive. Must be one of \"none\", \"lax\", or \"strict\" (case insensitive)."
-  (mb-session-cookie-samesite*))
 
 ;; In 0.41.0 we switched from Leiningen to deps.edn. This warning here to keep people from being bitten in the ass by
 ;; the little gotcha described below.
--- a/src/metabase/config.clj
+++ b/src/metabase/config.clj
@@ -56,7 +56,7 @@
 
 ;; separate map for EE stuff so merge conflicts aren't annoying.
 (def ^:private ee-app-defaults
-  {:embed-max-session-age      "1440"}) ; how long a FULL APP EMBED session is valid for. One day, by default
+  {:embed-max-session-age "1440"}) ; how long a FULL APP EMBED session is valid for. One day, by default
 
 (alter-var-root #'app-defaults merge ee-app-defaults)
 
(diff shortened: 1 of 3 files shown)


src/metabase/server/middleware/session.clj
```diff
@@ -68,6 +69,38 @@
                                                        metabase-embedded-session-cookie
                                                        metabase-session-timeout-cookie]))
 
+(def ^:private possible-session-cookie-samesite-values
+  #{:lax :none :strict nil})
+
+(defn- valid-session-cookie-samesite!
+  [value & [added-ex-data]]
+  (let [normalized-value (some-> value name str/lower-case keyword)]
```
noahmoss, line 77: You're getting a clj-kondo error here due to the use of `str/lower-case`—we discourage it in favor of `metabase.util/lower-case-en` which will have consistent results regardless of your site locale. (`str/lower-case` will have inconsistent results if your locale is Turkish which has bitten us before.)

Seems like it this linter rule was ignored in the previous code but I don't see a good reason for that over just using our custom function.

src/metabase/server/middleware/session.clj
```diff
@@ -68,6 +69,38 @@
...
+    (when-not (contains? possible-session-cookie-samesite-values normalized-value)
+      (throw (ex-info "Invalid value for session cookie samesite"
```
noahmoss, line 79: On the question of translation: an error string should be wrapped with the `metabase.util.i18n.tru` macro if there's any chance of it being user-facing on the frontend. `tru` will translate the string into the current user's locale, while `trs` will translate it into the site-wide locale.

In this case though, I think it would only show up in the backend logs, so it's fine to leave it untranslated. (Most backend logs are actually translated into the site locale currently, but I think the plan is to move away from that—see [this Slack discussion](https://metaboat.slack.com/archives/CKZEMT1MJ/p1697553976266779).)

noahmoss: And no need to add translations yourself; we use POEditor to source translations from volunteers prior to each major release.

noahmoss (approved):
Awesome, looks great. Just a couple lint errors to resolve.

> I'm not exactly sure how we decide between the :settings-manager and :admin settings. I went with :settings-manager here.

"Settings manager" refers to users who aren't full admins but are granted access to (most of) the "Settings" page in the admin panel ([docs](https://www.metabase.com/docs/latest/permissions/application#settings-access)). Any settings exposed on these pages should have `:settings-manager` visibility, so I think it's the right choice here.

> I'm not sure whether the frontend has its own definitions of possible values for set-based settings

The settings framework doesn't have a formal concept of "set-based settings," just keyword settings with extra validation added on the backend. We don't have that many keyword settings overall tbh, but I _think_ the frontend does just hardcode values where it needs them. It would make sense though to add a `:values` option to `defsetting` and have it do validation automatically, and expose the values to the frontend when the setting is fetched to reduce duplication. Could you open a tech debt issue on this? Maybe it would be a good one for you to work on if you'd be interested.

src/metabase/server/middleware/session.clj
johnswanson: Nice, thanks for the heads up - sorry for missing the linter error!

src/metabase/server/middleware/session.clj
johnswanson: I'm not positive, but I would assume that `(throw (ex-info "foo" {:http-status 400}))` is going to bubble up to the frontend - so maybe I should actually wrap this error message in `tru`?

Or actually, maybe this means it's not a good idea to throw one error that might bubble up to either logs or an HTTP error - because we need to translate one but not the other. So I'll actually pull those apart.

src/metabase/config.clj
```diff
@@ -56,8 +56,7 @@
 
 ;; separate map for EE stuff so merge conflicts aren't annoying.
 (def ^:private ee-app-defaults
-  {:embed-max-session-age      "1440"   ; how long a FULL APP EMBED session is valid for. One day, by default
-   :mb-session-cookie-samesite "lax"})
+  {:embed-max-session-age      "1440"}) ; how long a FULL APP EMBED session is valid for. One day, by default
```
noahmoss, line 59: nit - don't need the extra whitespace after removing the other map value

noahmoss (approved):

src/metabase/config.clj
johnswanson: 🤦 thank you!

paoliniluis:
fixes https://github.com/metabase/metabase/issues/35364

PR #35028, opened 2023-10-24
Do not set `updated_at` when pinned

johnswanson:
- add an additional `before-update` hook that removes the `updated_at` field when it and `collection_position` are the only updated fields

- use `methodical/prefer-method!` to sequence the latter hook after the `timestamped?` hook

The biggest problem here is a really odd one: this seems to work in the database, but somehow the frontend still reports that the card was updated. I am still figuring out why this is the case, but it may be that this requires frontend work that I'm not qualified to do.

Fixing the backend might still be worth it, since it's probably necessary for whatever frontend fix is necessary as well - but maybe it's not, since we seem to be ignoring `updated_at` entirely on the frontend - presumably it's cached somewhere and would be populated eventually?

### DIFF
--- a/src/metabase/models/card.clj
+++ b/src/metabase/models/card.clj
@@ -57,6 +57,7 @@
   (derive :metabase/model)
   ;; You can read/write a Card if you can read/write its parent Collection
   (derive ::perms/use-parent-collection-perms)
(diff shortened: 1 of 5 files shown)


johnswanson:
This was the wrong track - `updated_at` isn't relevant here, the `Revision` we create is. New PR incoming!

PR #35020, opened 2023-10-24
Mark `moment-timezone` as deprecated

kulyk:
Marks `moment-timezone` as deprecated in favor of `dayjs` using [eslint's `no-restricted-imports` rule](https://eslint.org/docs/latest/rules/no-restricted-imports) (also, adds `eslint-disable` comments for current occurrences of moment)

CleanShot 2023-10-24 at 17 35 47@2x

PR #35044, opened 2023-10-25
Follow up on deprecating `moment`

kulyk:
Adds a few things missed in #35020

1. Disallows `moment` imports too (not only `moment-timezone`)
2. Fixes `no-restricted-imports` rule was completely disabled for `metabase/ui` (I suppose it was made to allow direct `@mantine` imports)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,6 +18,10 @@
       {
         "paths": [
           {
+            "name": "moment",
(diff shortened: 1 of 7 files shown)


uladzimirdev (approved):

iethree:
❤️

PR #34897, opened 2023-10-20
Fix dayjs i18n init

ranquild:
How to verify:
- `yarn dev`
- Open http://localhost:3000/admin/settings/localization
- Set different days for start of week and debug `DatesProvider` initialization. `firstDayOfWeek` should be set correctly, e.g. 0 - Sunday, 1 - Monday etc

Screenshot 2023-10-20 at 14 46 26

PR #35045, opened 2023-10-25
Use `dayjs` in `TimeInput`

kulyk:
Migrates the `TimeInput` to `dayjs` instead of `moment`

Turns out `dayjs` doesn't handle time-only formats out of the box (like `HH:mm`) and [it needs the `CustomParsePlugin` to be installed](https://day.js.org/docs/en/parse/string-format). I had to bring back app-wide `days` initialization for that (@ranquild added it in #34802, but then reworked in #34897)

**To verify:** CI should be green 🟢


frontend/src/metabase/app.js
```diff
@@ -11,6 +11,8 @@ import "number-to-locale-string";
 // Should be imported before any other metabase import
 import "ee-overrides"; // eslint-disable-line import/no-duplicates
 
+import "metabase/lib/dayjs";
```
uladzimirdev, line 14: I think @ranquild had a very similar PR which he abandoned

uladzimirdev: @kulyk check this https://github.com/metabase/metabase/pull/34897/files#r1366806961 , initial PR is https://github.com/metabase/metabase/pull/34802/

kulyk: Yeah, I wasn't sure what he meant by "hard to manage plugins loaded globally" 🤔 
I'm seeing 5 failed E2E test groups and I'm hoping they're flakes :lol_sob:

ranquild: It broke storybook, so lets import this file also there

ranquild: @kulyk if you decided to use this approach please move other dayjs.extend to this file

ranquild (approved):
Please take a look at comments

frontend/src/metabase/app.js
ranquild: My initial idea for reverting the PR was that we can get away from using plugins globally, but if such basic functionality requires a plugin I guess that's not realistic

kulyk:
@ranquild thanks! Please take a look if the last two commits make sense

ranquild:
@kulyk `localeData` is not needed

ranquild (approved):

PR #35052, opened 2023-10-25
Restore mock in the `FilterPicker` test

kulyk:
I was hoping Jest would do a clean-up, but it has to be done manually 😢
We'd have problems when we adding and running more `FilterPicker` tests in parallel without that


uladzimirdev (approved):
indeed, mocks are not restored. We can automate this by adjusting jest config

PR #35057, opened 2023-10-25
[MLv2] Add `has-drill-thrus` for faster clickability checks

bshepherdson:
Several e2e tests
([example](https://www.deploysentinel.com/ci/runs/65374bdb7bcc14868115dfae))
have become flaky due to timing out waiting for a button to become clickable.
The button is slow to be clickable because the table viz is busy calling
`available-drill-thrus` for every cell in the table to determine whether the
cell is clickable.

This PR adds `has-drill-thrus` that only checks conditions rather than fully
computing the drills. It can short-circuit as soon as any drill is returned, so
it's much faster than calling `available-drill-thrus`. This also adds
`Mode.hasActionsForClick` and refactors
`Visualization.visualizationIsClickable` to use it.

- [x] BE unit tests needed

### DIFF
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -25,8 +25,8 @@
    [metabase.mbql.js :as mbql.js]
    [metabase.shared.util.time :as shared.ut]
    [metabase.util :as u]
-   [metabase.util.memoize :as memoize]
-   [metabase.util.log :as log]))
+   [metabase.util.log :as log]
+   [metabase.util.memoize :as memoize]))
 
 ;;; this is mostly to ensure all the relevant namespaces with multimethods impls get loaded.
 (comment lib.core/keep-me)
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -25,6 +25,7 @@
    [metabase.mbql.js :as mbql.js]
    [metabase.shared.util.time :as shared.ut]
    [metabase.util :as u]
+   [metabase.util.memoize :as memoize]
    [metabase.util.log :as log]))
 
 ;;; this is mostly to ensure all the relevant namespaces with multimethods impls get loaded.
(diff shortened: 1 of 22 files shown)


deniskaber (approved):
@bshepherdson , I updated `has-drill-thrus` on the FE side. Works as expected now, also did some clean-up.

But we encounter runtime issue with has-drill-thrus in this e2e test - 
https://github.com/metabase/metabase/blob/48355968dc38303768112156b9730d11ad5c2729/e2e/test/scenarios/visualizations/pivot_tables.cy.spec.js#L115-L131

```
Uncaught TypeError: Cannot read properties of undefined (reading 'field_ref')
    at js.cljs:876:1
    at Object.metabase$lib$js$drill_thru_context [as drill_thru_context] (js.cljs:877:1)
    at Object.metabase$lib$js$has_drill_thrus (js.cljs:908:1)
    at Module.hasDrillThrus (drills.ts:41:1)
    at Mode.hasActionsForClick (Mode.ts:35:1)
    at Visualization.hasClickActions (Visualization.jsx:237:1)
    at Visualization.render (Visualization.jsx:367:1)
    at finishClassComponent (react-dom.development.js:17160:1)
    at updateClassComponent (react-dom.development.js:17110:1)
    at beginWork (react-dom.development.js:18620:1)
```

src/metabase/lib/js.cljs
```diff
@@ -872,26 +872,41 @@
 (def ^:private row-cell       (js-cells-by #(.-col ^js %)))
 (def ^:private dimension-cell (js-cells-by #(.-column ^js %)))
 
+(defn- drill-thru-context [a-query stage-number column value row dimensions]
+  (lib.convert/with-aggregation-list (lib.core/aggregations a-query stage-number)
+    (let [column-ref (when-let [a-ref (.-field_ref ^js column)]
```
snoe, line 877: Maybe out of scope: Maybe I misunderstand where these structures are coming from but if the FE is storing these in `data-attributes` or something, it seems like too much knowledge to know that `column` has a `field-ref` key, same for `row-cell .-col`  and `dimension-cell .-column`. Should the caller not pass in the data we're interested in rather than encoding what seems to be a rather arbitrary js datastructure. If it's round-tripped from BE, can they use an opaque object?

Regardless, would appreciate a comment where these objects are coming from and expected shapes.

snoe (approved):

src/metabase/lib/js.cljs
bshepherdson: This is definitely out of scope; this code is not new and has been part of `available-drill-thrus` since its creation.

On the one hand, I agree this kinda sucks - coupling these calls to how the FE stores data about series, dimensions, rows of data, etc. Some of this data is coming directly from the BE, and some is derived in the FE. I suspect someday we'll see it refactored end to end, and that's a good thing. But this is not that day. This file is here to bridge between the FE and `metabase.lib.*`, so there's nowhere better for it to live.[1]

[1] It could also live in the TS wrapper, all else being equal. But getting passed the original FE objects is handy, because they're now being used as cache keys, effectively by `===` identity, so they need to be identical objects across calls for the caching to help.

darksciencebase:
@bshepherdson shall we close this one now or is it still relevant?

bshepherdson:
Yes, we can close this. It's a dead letter; the perf fix lay elsewhere.

PR #35062, opened 2023-10-25
[Dashboard] Make apply button aligned with fields on small screens

### LINKED ISSUE #34385: Dashboard filter Apply button does not follow column structure and shows up to right of other elements instead of below them
### Describe the bug

- When Auto-apply filters is disabled, and the viewport is narrow like on mobile devices, the Apply button shows up to the right of the filter elements and makes them narrower by taking up a full column.
- From Maz: "I assume that, on mobile, the button should be full-width, and should either appear at the bottom of the list of filters, or at the top of the list (I think my bias is the bottom)."
- Customer reported and also requested Apply button show up below filters.

### To Reproduce

1. View a dashboard with filters (X-ray Sample Data > Products table and save resulting dashboard)
2. Click i icon and disable Auto-apply filters on dashboard
3. Change browser to narrow viewport like mobile emulation or just make window very narrow (200-400px wide)
4. Change a filter
5. See Apply button to right of filters and filters resized narrower

### Expected behavior

Apply button should be same width as filters and show up below filters

### Logs

_No response_

```JSON
- Metabase v1.47.3 and `master` as of `24b3553`
```

### Severity

P3: Cosmetic, but can also reduce filter usability a bit by making them narrower

### Additional context

_No response_

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/34385

### Description

Fix button position on dashboard with auto-apply filters disabled on small screens.
Button has the same height as filter and full width

### How to verify

- create a dash, add filters there
- disabled auto-apply filters (click 'i' on the dashboard screen, toggle)
- enable small screen emulator  and change a value of some filter -> Apply button should appear

### Demo

image (8)

Screenshot 2023-10-25 at 19 14 53
Screenshot 2023-10-25 at 19 14 50

### Checklist


WiNloSt (approved):
The change looks good, though I'm not sure how necessary it is to use the new style props from Mantine rather than editing the styled component.

The `FilterApplyButton` could have been using the new `Button` from `metabase/ui` instead, but I'm not sure it's worth the effort.

frontend/src/metabase/parameters/components/FilterApplyButton/FilterApplyButton.styled.ts
```diff
@@ -1,9 +1,16 @@
...
 
 export const ApplyButton = styled(Button)`
   margin-left: auto;
   margin-top: ${space(1)};
+
+  ${breakpointMaxSmall} {
+    width: 100%;
+    /* make button same height as fields on small screens */
+    padding-top: 11px;
+    padding-bottom: 11px;
```
WiNloSt, line 14: This is what I asked @kdoh at the offsite. By using our new component library based on Mantine, we should not have to specify one-off dimensions like this 99% of the time in the future.

uladzimirdev:
I think we can live with the old button as it works fine

PR #35065, opened 2023-10-25
Stop publishing `:event/card-upate` on card (un)archive

johnswanson:
I noticed this issue when attempting to stop publishing card-update events when a card is pinned.

There are two issues with the existing code:

- first, the `publish-card-update!` function was called with the *new* (updated) card, but the code checked to see whether the passed card's `:archived` status and the new value of `archived?` were different. So if you asked to archive a card, the card would be updated to be `:archived true`, and then `(and archived? (not (:archived card)))` would return `false`. Therefore, we'd *always* fall through to firing an `:event/card-update` event.

- second, if we *had* successfully fired an `:event/card-archive` or `:event/card-unarchive` events, it would have thrown an exception, because those events do not derive from `:metabase/event`.

To be clear, I'm not entirely sure what the *right* behavior is here. There are a few options:

- we could make the behavior align with the apparent intent of the code, and fire `:event/card-archive` and `:event/card-unarchive` events when the card is archived or unarchived. However, nothing actually subscribes to these events.

- a modification of the above: we can just skip firing *any* event when a card is archived or unarchived. Nothing listens to `:event/card-(un)archive` anyway.

- we could make the code align more clearly with the current behavior. Currently we fire an `:event/card-update` for every change, including archival. We could just delete the `publish-card-update!` function and always publish that event.

My initial impulse was to go with the second option (skip firing an event when a card is archived or unarchived) but after talking with Noah, he suggested that publishing `:event/card-update` would actually be useful for the audit log. So we'll do that!

### DIFF
--- a/src/metabase/api/card.clj
+++ b/src/metabase/api/card.clj
@@ -608,20 +608,6 @@ saved later when it is ready."
     (validation/check-embedding-enabled)
     (api/check-superuser)))
 
-(defn- publish-card-update!
-  "Publish an event appropriate for the update(s) done to this CARD (`:card-update`, or archiving/unarchiving
-  events)."
-  [card archived?]
-  (let [event (cond
-                ;; card was archived
-                (and archived?
-                     (not (:archived card))) :event/card-archive
-                ;; card was unarchived
-                (and (false? archived?)
-                     (:archived card))       :event/card-unarchive
-                :else                        :event/card-update)]
-    (events/publish-event! event card)))
-
 (defn- card-archived? [old-card new-card]
   (and (not (:archived old-card))
        (:archived new-card)))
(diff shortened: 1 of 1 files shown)


src/metabase/api/card.clj
```diff
@@ -770,7 +757,9 @@ saved later when it is ready."
 
   (let [card (t2/select-one Card :id id)]
     (delete-alerts-if-needed! card-before-update card)
-    (publish-card-update! card archived)
+    (when-not (set/subset? (set (keys card-updates))
```
johnswanson, line 760: Why do set comparison instead of plain `(= [:archived] (keys card-updates))`? Because I discovered this when I was working on no longer firing `:event/card-update` when we change a card's position. I separated this into a different PR for clarity, but the next step will just be adding `:collection_position` to the set of keys to ignore here.

noahmoss (commented):
I think we _can_ actually just fire a `card-update` event when a card is archived or unarchived. It will be useful to have in the audit log, but we don't necessarily need a separate event type for it, I don't think.

Could you rebase this PR on top of the `auditv2-main` feature branch instead of targeting master? That will pull in some recent changes to the way auditing works and hopefully help avoid a merge conflict with this code down the line.

johnswanson:
> I think we _can_ actually just fire a `card-update` event when a card is archived or unarchived. It will be useful to have in the audit log, but we don't necessarily need a separate event type for it, I don't think.
> 
> Could you rebase this PR on top of the `auditv2-main` feature branch instead of targeting master? That will pull in some recent changes to the way auditing works and hopefully help avoid a merge conflict with this code down the line.

Nice, that sounds good. Done!

noahmoss (approved):

Issue #35025, opened 2023-10-24
Pivot viz fails on a nested question


zbodi74:
### Describe the bug

Pivot viz works while editing a question based on a nested question, but it fails after the question is saved.

### To Reproduce

1. New GUI question: select all records from Products
2. Save it as Question1
3. New GUI question: select Question1; Aggregate: Count of Rows, Sum of Total; Group by Category, CreatedAt: Month
4. Select Pivot table for the visualization
5. The viz shows up correctly
6. Save it as Question2
7. (!) Re-open Question2: the viz fails with _There was a problem with your question / Error running query_

### Expected behavior

It should work

### Logs

```
2023-10-24 18:19:22,516 ERROR middleware.catch-exceptions :: Error processing query: Error running query                                                                                                    [479/1701]
{:json_query
 {:database 1,
  :type :query,
  :query
  {:aggregation [[:count] [:sum [:field "PRICE" {:base-type :type/Float}]]],
   :breakout ([:expression "pivot-grouping"]),
   :source-table "card__1",
   :expressions {:pivot-grouping [:abs -9223372036854775805]}},
  :constraints {:max-results 10000, :max-results-bare-rows 2000},
  :middleware {:js-int-to-string? true, :ignore-cached-results? false, :add-default-userland-constraints? true},
  :async? true},
 :native
 {:query
  "SELECT \"source\".\"pivot-grouping\" AS \"pivot-grouping\", COUNT(*) AS \"count\", SUM(\"source\".\"PRICE\") AS \"sum\" FROM (SELECT \"source\".\"PRICE\" AS \"PRICE\", ABS(-9223372036854775805) AS \"pivot-groupi
ng\" FROM (SELECT \"PUBLIC\".\"PRODUCTS\".\"ID\" AS \"ID\", \"PUBLIC\".\"PRODUCTS\".\"EAN\" AS \"EAN\", \"PUBLIC\".\"PRODUCTS\".\"TITLE\" AS \"TITLE\", \"PUBLIC\".\"PRODUCTS\".\"CATEGORY\" AS \"CATEGORY\", \"PUBLIC
[...]

PR #35076, opened 2023-10-26
Repro #35025 and extra exception handling around `pivot-options`

noahmoss:
Adds a backend test that repros https://github.com/metabase/metabase/issues/35025 — it tries to run a pivot query in which the breakout in the query uses field refs that include field names, rather than IDs. This currently errors since the field refs can't be correlated with the fields in the `:pivot_table.column_split` key of the viz setting. If `pivot-options` can be changed to match fields regardless of if they use ID or name, this test should pass.

Also added an extra try/catch around `pivot-options` to add a more helpful error description if the fields can't be correlated. I don't think we should squash this error and run the query anyway since it could return inconsistent results, but I'm not 100% sure on this.


test/metabase/query_processor/pivot_test.clj
```diff
@@ -357,3 +358,49 @@
...
+                                       [:field
+                                        (mt/id :products :created_at)
+                                        {:base-type :type/DateTime :temporal-unit :month}]]
+                            :aggregation [[:count]]})
+            viz-settings (select-keys
+                          (mt/query reviews
+                                    {:pivot_table.column_split
+                                     {:rows [$rating]
+                                      :cols [$created_at]}})
+                          [:pivot_table.column_split])]
```
noahmoss, line 376: This feels wrong but I couldn't tell if there was a more direct way to use the `mt/query` shorthand symbols in a map that's not actually a full query

test/metabase/query_processor/pivot_test.clj
```diff
@@ -357,3 +358,49 @@
...
+             (qp.pivot/run-pivot-query query
+                                       {:visualization-settings viz-settings}
+                                       (context.default/default-context))))))
+
+    (testing "If pivot_table.column_split includes includes field refs that include with field names, query still works"
+      (let [query        (mt/mbql-query products
+                           {:breakout [*category
+                                       [:field
+                                        "CREATED_AT"
+                                        {:base-type :type/DateTime :temporal-unit :month}]]
```
noahmoss, line 391: Also wasn't sure if there was a shorthand for a named field that includes `:temporal-unit` — basically combining both `*` and `!`?

test/metabase/query_processor/pivot_test.clj
```diff
@@ -357,3 +358,49 @@
...
+                                      :cols [$created_at]}})
+                          [:pivot_table.column_split])]
+
+        (is (schema=
+             {:status    (s/eq :completed)
+              :row_count (s/eq 156)
+              s/Keyword  s/Any}
+             (qp.pivot/run-pivot-query query
+                                       {:visualization-settings viz-settings}
+                                       (context.default/default-context))))))))
```
noahmoss, line 406: This query, with named field refs in the breakout, actually still fails even if you don't call `pivot-options` because of an exception in `query_processor.util.add_alias_info`: `"Cannot determine the source table or query for Field clause [:field \"CATEGORY\" {:base-type :type/Text}]"`.

I don't quite understand why, because I think I got the field refs to look identical to the ones I have in the saved question in the repro from the issue (when created on the release branch). Pretty confused here.

test/metabase/query_processor/pivot_test.clj
```diff
@@ -357,3 +358,49 @@
...
+  (mt/dataset sample-dataset
+    (testing "Query works when breakout includes field refs with IDs"
+      (let [query        (mt/mbql-query products
+                           {:breakout [$category
+                                       [:field
+                                        (mt/id :products :created_at)
+                                        {:base-type :type/DateTime :temporal-unit :month}]]
+                            :aggregation [[:count]]})
+            viz-settings (select-keys
+                          (mt/query reviews
```
camsaul, line 372: don't understand how this can possibly be expected to work since these columns are from `reviews` and the query is against `products`

test/metabase/query_processor/pivot_test.clj
```diff
@@ -357,3 +358,49 @@
...
+                                    {:pivot_table.column_split
+                                     {:rows [$rating]
+                                      :cols [$created_at]}})
```
camsaul, line 375: The code looks for `:columns` but this uses `:cols`

test/metabase/query_processor/pivot_test.clj
camsaul: yes, you can combine them

camsaul: `!month.*whatever`

test/metabase/query_processor/pivot_test.clj
camsaul: `$ids`

test/metabase/query_processor/pivot_test.clj
camsaul: #35025 specifically mentions a nested question, you aren't creating a nested question here.

PR #33190, opened 2023-08-15
Download previously saved uberjar for E2E workflow

nemanjaglumac:
E2E tests require a full-fledged Metabase artifact in order to run.
We've been **always** building it from scratch so far, even when there were no source code changes in a PR! This is wasteful and slow.

This PR introduces a new logic that will find the previously saved uberjar by traversing the current commit parents and searching for the matching uberjar.

If it finds it, the build will be skipped and E2E tests will download and will use that uberjar directly.
If it cannot find the uberjar or it fails, `build` job will take over as a fallback mechanism and it's business as usual.

## Download Logic Explained
GitHub API allows to list all artifacts across all workflows in a repo.
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-artifacts-for-a-repository

If you try this endpoint, you'll realize just how many artifacts we store
```json
{
	"total_count": 30242,
        "artifacts":[{...}, {...}, ...{}]
}
```

Lucky for us, it accepts a few query parameters that will help us in this quest.
- `name`: we know we're searching for either "metabase-oss-uberjar" or `metabase-ee-uberjar"
- `page`: Page number of the results to fetch.
- `per_page`: How many results do we want to fetch per page to increase our chance of finding the right artifact (max is 100)

The individual artifact has the following shape:
```json
{
  "id": 861710849,
  "node_id": "MDg6QXJ0aWZhY3Q4NjE3MTA4NDk=",
  "name": "metabase-ee-uberjar",
  "size_in_bytes": 335089591,
  "url": "https://api.github.com/repos/metabase/metabase/actions/artifacts/861710849",
  "archive_download_url": "https://api.github.com/repos/metabase/metabase/actions/artifacts/861710849/zip",
  "expired": false,
  "created_at": "2023-08-14T18:47:16Z",
  "updated_at": "2023-08-14T18:47:19Z",
  "expires_at": "2023-09-13T18:17:04Z",
  "workflow_run": {
    "id": 5858087729,
    "repository_id": 30203935,
    "head_repository_id": 30203935,
    "head_branch": "mblib-column-settings-round-trip",
    "head_sha": "4ba2debf861132965d30fa69b8493c86a36ba9d0"
  }
}
```

In order to get the `archive_download_url` we will match the `head_sha` to the parent commit SHA1. I.e. we filter the list of all possible artifacts by the artifact name and the parent's SHA1.

Once we know the `archive_download_url`, the only thing left to do is to download and store it using `curl`.

[...]

PR #35079, opened 2023-10-26
Revert uberjar caching for E2E tests

nemanjaglumac:
This PR reverts the idea of caching the uberjar for E2E tests. It's not a 1:1 revert of the original commit, because a lot of things have changed since then.

The main point and the tl;dr is this:
That the idea to find previously built uberjar was great in theory. It worked in most of the cases but it failed in cases of force-pushes (rebase, ammend, etc.). Some people reported that it might not even work when the base branch is merged into a PR.

The other edge case happened when the base branch still didn't finish running all workflows, but your branch needed an artifact from it. Unfortunately, GitHub REST API (that we were using for this custom logic) doesn't even see the artifact if workflow is still running.

There are two ways to find/download the artifact.
1. The official `actions/download-artifact` action (which can only download the artifact from the current workflow, while it's still ongoing)
2. The REST API, which can only see and download artifacts from workflows that have already finished running

### Differences from [the original commit](https://github.com/metabase/metabase/pull/33190)
- We're keeping the commit SHA1 in the artifact name
- The custom logic is not removed, but is actually used in both E2E Replay and Stress-test E2E flake workflows

### Altered git history
The biggest problem that this custom approach faced was the frequent use of rebase and/or amend. In other words, changing the git history. Since it calculates the hashes of the previous commits by using `git rev-parse COMMIT^`, it either failed to find the uberjar or it sometimes failed to find the hash in the updated git tree.

GitHub keeps the information about the force-pushed commits and it is possible to find the previous commit (which I've tried [here](https://github.com/metabase/metabase/pull/35056), but that only works for the immediate parent/predecessor. Even if I could find a way to traverse the old history further, the effort and the risk didn't seem worth it.

The biggest of all risks was finding the wrong uberjar and then relying on tests from it.
- In a good scenario, your tests would fail and you'd be alerted (although you'd loose significant time until you realized it was testing agains the wrong commit)
- In a bad scenario, your tests would pass even though you might have introduced the breaking change.

## Conclusion
It is better to revert and to wait for the build, even when unnecessary, than it is to risk wrong results or failed builds and the lost developer time.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -68,6 +68,8 @@ jobs:
 
       - name: Prepare uberjar artifact
         uses: ./.github/actions/prepare-uberjar-artifact
+        with:
+          name: metabase-${{ matrix.edition }}-${{ github.event.pull_request.head.sha || github.sha }}-uberjar
 
   e2e-tests:
     needs: [build, e2e-matrix-builder]
@@ -112,7 +114,7 @@ jobs:
       - name: Retrieve uberjar artifact for ${{ matrix.edition }}
         uses: actions/download-artifact@v3
         with:
-          name: metabase-${{ matrix.edition }}-uberjar
+          name: metabase-${{ matrix.edition }}-${{ github.event.pull_request.head.sha || github.sha }}-uberjar
 
       - name: Get the version info
         run: |
@@ -214,7 +216,7 @@ jobs:
       - name: Retrieve uberjar artifact for ${{ matrix.edition }}
         uses: actions/download-artifact@v3
         with:
-          name: metabase-${{ matrix.edition }}-uberjar
+          name: metabase-${{ matrix.edition }}-${{ github.event.pull_request.head.sha || github.sha }}-uberjar
 
       - name: Get the version info
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -228,19 +213,25 @@ jobs:
...
           maildev: true
 
-      - name: Download Metabase enterprise uberjar
-        uses: ./.github/actions/e2e-download-uberjar
+      - name: Retrieve uberjar artifact for ${{ matrix.edition }}
+        uses: actions/download-artifact@v3
         with:
-          edition: 'ee'
-          was-built: ${{ needs.build.result == 'success' }}
+          name: metabase-ee-${{ github.event.pull_request.head.sha || github.sha }}-uberjar
```
alxnddr, line 219: What are the cases when `github.event.pull_request.head.sha` is missing so we need `|| github.sha`?

nemanjaglumac: @iethree suggested we start including the commit SHA1 in the artifact name.

GitHub has a semi-hidden quirk that the HEAD (or the github.sha) points to the merge commit it creates on the fly in the event of `pull_request`, while it points to the actual latest commit in your branch in the event of `push`.

The way to get the latest commit in a branch for pull request is `github.event.pull_request.head.sha`. If that doesn't exist (when we're on `master`, it will simply fall back to the regular `github.sha`.

iethree (approved):
LGTM 👍

PR #35086, opened 2023-10-26
Refactor: Tighten FieldFilterUiParameter type

calherries:
Follow up to https://github.com/metabase/metabase/pull/34958

This PR tightens the `FieldFilterUiParameter` type to require `fields` on an instance of that type to be non-empty if it is present.

The biggest benefit is that now [`FieldSourceModal`](https://github.com/metabase/metabase/pull/35086/files#diff-e3a763190e3088145320e8a7a85d65c82441de17d7aeda196916c8f124daaffcR176) now takes a narrower type `UiParameterWithFields` instead of `FieldFilterUiParameter`. The former is guaranteed to have one field or more while the latter may or may not have any.

Another tightening is with [`hasFields`](https://github.com/metabase/metabase/pull/35086/files#diff-71b7995e3b36068e5ba6e07964575b1f07f14f8bcb15d9bee355f7078d9e2311R7) which was previously a type guard for `FieldFilterUiParameter`, when it actually should have been a type guard for a subset of `FieldFilterUiParameter`, namely `UiParameterWithFields`.


ranquild (commented):
These kinds of types are usually impossible to properly enforce with typescript, I wouldn't do that. Last month I faced a lot of issues with similar types, e.g. in with `mantine`:
- it requires the color array to have exactly 10 colors by `[string, string, ...]` you get it. If you generate it you would have to cast it to the type
- same with `type WeekDay = 1 | 2 | 3 | 4 | 5 | 6 | 7` which cannot work without a type guard

calherries:
@ranquild What's wrong with having to cast to a type and using type guards? In this case the change allows me to force the programmer to explicitly write the assumption that `FieldSourceModal` is called with a parameter of type `UiParameterWithFields`, as I've done [here](https://github.com/metabase/metabase/blob/39f3faff3d5226952644912b852b3c03eb10897a/frontend/src/metabase/parameters/components/ValuesSourceModal/ValuesSourceTypeModal.tsx#L101). You could argue it's best done with a comment, but a type is more precise.

ranquild:
@calherries casting practically means that there is no real way to enforce this rule. I'm against this approach.

ranquild (changes requested):
Not idiomatic TS usage

calherries:
@ranquild Okay, I will avoid using narrower types than the type checker can verify from now on.

I've just put up a PR to apply this principle to a couple of instances where we violate it: [Remove innappropriate type casts and type guards for FieldFilterUiParameter](https://github.com/metabase/metabase/pull/35204#top).

calherries:
Closing due to not being able to enforce it using the type checker

PR #35090, opened 2023-10-26
Move metabase/core/utils/types to metabase/lib/types

calherries:
This PR moves `metabase/core/utils/types` to `metabase/lib/types`. 

`metabase/lib` is where we keep most of the utility functions, so they belong there instead of in `metabase/core`.

There is also `metabase/core/utils/arrays` and `metabase/core/utils/errors`, so if this gets approved I'll move those over to `metabase/lib` too.


ranquild (approved):

calherries:
@ranquild apparently I need another review. I added this commit for some more replacements after merging with master 
https://github.com/metabase/metabase/pull/35090/commits/ef51da5ea41c98dcc8cc4bad3300075eb3a0b4ec

ranquild (approved):

PR #35094, opened 2023-10-26
[MLv2] Filters — Add tests for custom expression editor

kulyk:
Adds basic unit test coverage to filter picker's custom expression editor

### DIFF
--- a/frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
+++ b/frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
@@ -1,7 +1,7 @@
 import userEvent from "@testing-library/user-event";
 import dayjs from "dayjs";
 import { createMockMetadata } from "__support__/metadata";
-import { render, screen } from "__support__/ui";
+import { act, render, screen, waitFor } from "__support__/ui";
 
 import type { StructuredDatasetQuery } from "metabase-types/api";
 import { createMockField, createMockSegment } from "metabase-types/api/mocks";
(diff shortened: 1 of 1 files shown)


frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
```diff
@@ -227,4 +242,53 @@ describe("FilterPicker", () => {
...
+      // but for some reason it doesn't work without `act`.
+      // eslint-disable-next-line testing-library/no-unnecessary-act
+      await act(async () => {
+        await userEvent.type(input, text);
+        await userEvent.tab();
+      });
+
+      await waitFor(() => expect(button).toBeEnabled());
+      userEvent.click(button);
+    }
```
kulyk, line 261: This drives me crazy — I've tried many different ways to make this work, but it seems like only a combination of `act` with `userEvent.tab` and `waitFor` did the thing :/

TLDR: the expression editor's "Done" button becomes enabled once the expression is valid. The expression is submitted for validation on the input's blur event. So we should edit > blur > click "Done"

What I've tried:

- edit > click "Done"
- edit > `fireEvent.blur(input) > click "Done"
- edit > `userEvent.tab()` > click "Done"
- edit > `fireEvent.focus(input) > `fireEvent.blur(input)` > click "Done"

Neither of them worked. In all cases, the "Done" button remained disabled, even with:

```js
await waitFor(() => expect(button).toBeEnabled());
```

The only two things that worked are `act` or something like that:

```jsx
render(
  <>
     <FilterPicker {...props} />
     <input data-testid="__TEST_ONLY_HACK_INPUT" />
  </>
)

userEvent.type(expressionInput, formula)
userEvent.type(screen.getByTestId("__TEST_ONLY_HACK_INPUT", "")
userEvent.click(doneButton)
```

ranquild (approved):

PR #35098, opened 2023-10-26
Add tests for MLv2 filter wrappers

ranquild:
Added tests for all filter type wrappers.

### DIFF
--- a/frontend/src/metabase-lib/filter.ts
+++ b/frontend/src/metabase-lib/filter.ts
@@ -323,13 +323,19 @@ export function relativeDateFilterClause({
   offsetBucket,
   options,
 }: RelativeDateFilterParts): ExpressionClause {
+  const columnWithoutBucket = withTemporalBucket(column, null);
+
   if (offsetValue == null || offsetBucket == null) {
-    return expressionClause("time-interval", [column, value, bucket], options);
+    return expressionClause(
+      "time-interval",
+      [columnWithoutBucket, value, bucket],
+      options,
+    );
   }
 
   return expressionClause("between", [
     expressionClause("+", [
-      column,
+      columnWithoutBucket,
       expressionClause("interval", [-offsetValue, offsetBucket]),
     ]),
     expressionClause("relative-datetime", [value < 0 ? value : 0, bucket]),
--- a/frontend/src/metabase-lib/filter.ts
+++ b/frontend/src/metabase-lib/filter.ts
@@ -661,7 +661,7 @@ function serializeTime(value: Date): string {
 }
 
 function deserializeTime(value: string): Date | null {
-  const time = moment(value, [TIME_FORMAT_MS], true);
+  const time = moment(value, TIME_FORMAT_MS, true);
   if (!time.isValid()) {
     return null;
   }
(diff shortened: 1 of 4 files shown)


frontend/src/metabase-lib/filter.ts
```diff
@@ -632,6 +632,7 @@ function isExcludeDateBucket(
 
 const DATE_FORMAT = "yyyy-MM-DD";
 const TIME_FORMAT = "HH:mm:ss";
+const TIME_FORMAT_MS = "HH:mm:SS.sss";
```
ranquild, line 635: For compatibility with existing format https://github.com/metabase/metabase/blob/1f7e261ec4702b36aa1f815440396187c4b7e931/frontend/src/metabase/query_builder/components/filters/pickers/TimePicker/TimePicker.jsx#L23, tests TBD in another PR

kulyk (approved):
The test structure here is:

well-done-ok-gif-by-funk

I suppose we'd want to break down `filter` and `filter.unit.spec` sooner rather than later, they're both around 1K LoC long and it could become hard to follow (shouldn't happen now)

frontend/src/metabase-lib/filter.ts
```diff
@@ -323,13 +323,19 @@ export function relativeDateFilterClause({
   offsetBucket,
   options,
 }: RelativeDateFilterParts): ExpressionClause {
+  const columnWithoutBucket = withTemporalBucket(column, null);
```
ranquild, line 326: Bug 1 - not clearing temporal bucket. If you create an exclude date filter (sets bucket) and then change it to a relative filter, it will not work correctly. There is a unit test for this.

frontend/src/metabase-lib/filter.ts
```diff
@@ -393,7 +399,9 @@ export function excludeDateFilterParts(
 
   const bucket = temporalBucket(column);
   if (!bucket) {
-    return { column, operator, bucket, values: [] };
+    return serializedValues.length === 0
```
ranquild, line 402: Bug 2 - without this check we accepted unsupported expressions like "[!=, column, column]" as valid exclude date filters, which is not correct (tested).

frontend/src/metabase-lib/filter.ts
```diff
@@ -758,7 +766,7 @@ function relativeDateFilterPartsWithOffset(
     column,
     value: startValue < 0 ? startValue : endValue,
     bucket: startBucket,
-    offsetValue,
+    offsetValue: offsetValue * -1,
```
ranquild, line 769: Bug 4 - we negate the offset value when constructing an expression but don't do it here (tested) https://github.com/metabase/metabase/blob/9abaf80bfb4bb9a53e02959de2cb95419953ef93/frontend/src/metabase-lib/filter.ts#L339

frontend/src/metabase-lib/types.ts
```diff
@@ -198,6 +198,7 @@ export type ExpressionOperatorName =
   | "not-empty"
   | "starts-with"
   | "ends-with"
+  | "concat"
```
ranquild, line 201: For testing only. I need a valid string expression that is not handled by our string filter UI.

kulyk (approved):

PR #35105, opened 2023-10-26
Automate reordering release tags

iethree:
Closes https://github.com/metabase/metabase/issues/[issue_number]

### Description

We always want our open source release to show up as the most recent on our github releases page. We accomplish this by manually forcing the tag date for the  enterprise version tag to -1 day. This was a boring, tedious manual affair for every release - BUT NO LONGER!

### How to verify

- I just used this locally to reorder the 47.6 release tags.
- I will also [test this in my fork](https://github.com/iethree/metabase/actions/runs/6672511506/job/18136715423) to verify that all the git operations work as expected in CI

### Demo

Screen Shot 2023-10-26 at 10 44 28 AM

### DIFF
--- a/release/reorder-tags.sh
+++ b/release/reorder-tags.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# for some reason, github always puts our enterprise release notes above our open source
+# release notes on the releases page. We always want the open source release notes to be on top.
+# The hack we've found for this is to force-push yesterday's date into the enterprise release tag.
+
+# usage: ./reorder-tags.sh <enterprise-tag> <commit-hash>
+
+tagName=$1
+commit=$2
+
+if [ -z "$tagName" ]; then
+  echo "Please provide a tag name (e.g v1.49.9)"
+  exit 1
+fi
+
+if [ -z "$commit" ]; then
+  echo "Please provide a commit hash"
+  exit 1
+fi
+
+echo "Reordering commits for $tagName on $commit"
+
+git fetch origin --tags --prune-tags --force
+
+yesterday="$(date -v-1d +%Y-%m-%d)T23:00:00Z"
+
+GIT_COMMITTER_DATE=$yesterday git tag -f -a -m "$tagName" "$tagName" "$commit"
+
+git show "$tagName" --quiet
+
+git push origin --force "$tagName"
--- a/release/reorder-tags.sh
+++ b/release/reorder-tags.sh
@@ -2,6 +2,7 @@
 # for some reason, github always puts our enterprise release notes above our open source
 # release notes on the releases page. We always want the open source release notes to be on top.
 # The hack we've found for this is to force-push yesterday's date into the enterprise release tag.
+# see https://github.com/orgs/community/discussions/8226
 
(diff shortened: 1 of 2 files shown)


alxnddr (approved):
That was the most tedious thing! 🎉

release/reorder-tags.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# for some reason, github always puts our enterprise release notes above our open source
```
kamilmielnik, line 2: Maybe we can link this thread here: https://github.com/orgs/community/discussions/8226

release/reorder-tags.sh
```diff
@@ -0,0 +1,31 @@
...
+if [ -z "$commit" ]; then
+  echo "Please provide a commit hash"
+  exit 1
+fi
+
+echo "Reordering commits for $tagName on $commit"
+
+git fetch origin --tags --prune-tags --force
+
+yesterday="$(date -v-1d +%Y-%m-%d)T23:00:00Z"
```
kamilmielnik, line 25: I'm on Ubuntu 22.04 (just like GH actions) and `-v` option is not supported.

This works though: `date --date="yesterday" +%Y-%m-%d`

release/reorder-tags.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# for some reason, github always puts our enterprise release notes above our open source
+# release notes on the releases page. We always want the open source release notes to be on top.
+# The hack we've found for this is to force-push yesterday's date into the enterprise release tag.
```
kamilmielnik, line 4: This trick may not work if we ever need to have 2 releases in 1 day.
We'd probably end up with something like:
1. :heavy_check_mark: latest oss
2. :x: previous oss
3. :x: latest ee
4. :heavy_check_mark: previous ee

nemanjaglumac: We've had situations in the past where we released three or even four releases in a day (backporting a hotfix).
My manual approach was to apply this hack to each individual EE version.

And to confirm - yes, I did run into this problem.
There isn't a shared opinion on how to order multiple releases across versions, though. Some people think it's best to have OSS-EE grouped in pairs **per version**. Others are ok with OSS-OSS, followed by EE-EE.

We should probably decide which one to follow and stick with it.

nemanjaglumac (approved):
Approving in principle, but please take a look at @kamilmielnik 's amazing review comments! Especially about the date calculation.

Thank you for tackling this.

release/reorder-tags.sh
iethree: yeah, if we have multiple releases in the same day, we'll probably have to reorder manually, but it happens so infrequently, I don't know that it's worth building automation for it.

nemanjaglumac: Don't bother with that.
This already solves most of our pain points from the past.

release/reorder-tags.sh
iethree: ok, got it working here: https://github.com/iethree/metabase/actions/runs/6672511506/job/18136715423

kamilmielnik (approved):
:+1:

PR #34087, opened 2023-09-25
Add remove columns panel

npfitz:
Closes https://github.com/metabase/metabase/issues/33841

### Description
This PR adds a new panel to control adding and removing table columns from a query, separating the idea of add/remove from show/hide in viz settings. The new panel is only available in structured queries (and should not be displayed for dashcards), and should allow you to implicitly add fields from related tables, search available columns, and perform bulk add / remove column operations per table.

### How to verify
Create a structured question and try adding / removing columns from a query. A few points:
- Showing / Hiding should not re-run queries
- Try using saved questions as table sources
- You should now be able to hide columns from the table header

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #35114, opened 2023-10-26
Hide Viz Settings Buttons in Column Headers in Raw Table Mode

iethree:
resolves https://github.com/metabase/metabase/issues/35244

### Description

For a long time, in the "raw table" view, we've been showing buttons that do nothing because we don't apply visualization settings in the raw view.  #34087 adds a new button that doesn't work for the same reason. This simply hides those buttons that don't work in rawTable mode.

Screen Shot 2023-10-26 at 2 16 59 PM

### How to verify

On the base branch
- go into raw table mode by clicking the left button in the bottom-center of the screen
- click on a column header > click hide column
- see nothing happens
- click on a column header > click column settings
- change anything you want
- see that nothing happens

On this branch:
- see that in raw table mode, neither of those buttons appears

### Checklist

### DIFF
--- a/frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
+++ b/frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
@@ -30,6 +30,7 @@ import { isSameSeries, getCardKey } from "metabase/visualizations/lib/utils";
 
 import { getMode } from "metabase/visualizations/click-actions/lib/modes";
 import { getFont } from "metabase/styled-components/selectors";
+import { getIsShowingRawTable } from "metabase/query_builder/selectors";
 
 import ErrorBoundary from "metabase/ErrorBoundary";
 import { isRegularClickAction } from "metabase/visualizations/types";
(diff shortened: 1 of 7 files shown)


iethree:
I will add tests once I get confirmation that this is the behavior we want

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
```diff
@@ -236,7 +238,13 @@ class Visualization extends PureComponent {
 
     return mode
       ? mode.actionsForClick(
-          { ...clicked, extraData: getExtraDataForClick(clicked) },
+          {
+            ...clicked,
+            extraData: {
+              ...getExtraDataForClick(clicked),
+              isRawTable: this.props.isRawTable,
```
npfitz, line 245: nit: would move this to line 232

npfitz (approved):
Aside from the nittiest of nits, I think this is fine. I can't think of any other way to get this info into the click actions, and I don't think there is a problem using `extraData` here (though to be honest, it doesn't seem to be used much). Also, it's a huge win to not have buttons visible and clickable that just, don't do anything haha

luizarakaki (approved):
thanks for this

This will significantly reduce frustration. I already tried to rename columns there 39284 times.

PR #35115, opened 2023-10-26
Enrich audit log update events with previous and new values

escherize:
This adds more details about new and previous values when an _update_ occurs.

Before an audit_log event's details contained the new values of interesting keys and looked like:

`{:name "New Name" :color "Green"}`

Now the details will all have a `:previous-value` and `:new-value` key with all keys that changed in both. It should look like:

```clojure
{:new-value {:name "New Name" :color "Green"}
 :previous-value {:name "Old Name" :color "Black"}}
```

## Updates to enrich with :new + :previous data
- [x] user-update
- [x] card-update
- [x] metric-update
- [x] segment-update

## misc.
- [x] tests


test/metabase/events/audit_log_test.clj
```diff
@@ -288,25 +288,23 @@
...
-                               :description      (:description metric)
-                               :revision_message "update this mofo"
-                               :database_id (mt/id)
-                               :table_id    (mt/id :venues)}}
-                (event "metric-update" (:id metric)))))))))
+      (mt/with-test-user :rasta
+        (let [event (-> (assoc metric
+                               :actor_id         (mt/user->id :rasta)
+                               :revision_message "update this mofo")
+                        ;; doing this specifically to ensure :actor_id is utilized
```
escherize, line 295: I am not sure that I understand how `:actor_id` is different or important in here, so please scrutinize this test carefully 🙏

src/metabase/api/user.clj
```diff
@@ -450,8 +450,11 @@
                                          api/*is-superuser?* (conj :login_attributes))
                               :non-nil (cond-> #{:email}
                                          api/*is-superuser?* (conj :is_superuser))))]
-          (t2/update! User id changes)
-          (events/publish-event! :event/user-update (assoc (t2/select-one User :id id) :changes changes)))
+          (let [raw-user-before-update (t2/select-one User :id id)]
```
adam-james-v, line 453: This is an extra db call now isn't it? Is there some way we could avoid both getting the user and updating the user? Like, is there some way to do it in one fell swoop? Is it a non-issue?

escherize: It is, I think it might be better to use what fetch-user returns as the meaning of what a user is. Then I can use the one from line 429.

test/metabase/api/user_test.clj
```diff
@@ -1252,16 +1252,14 @@
...
-                    :model    "User"
-                    :model_id id
-                    :details  {:first_name "Johnny"
-                               :last_name "Appleseed"}}
-                   (audit-log-test/event :user-update id))))))))
+    (t2.with-temp/with-temp [User {:keys [id]} {:first_name "John" :last_name  "Cena"}]
+      (testing "PUT /api/user/:id"
+        (mt/user-http-request :crowberto :put 200 (format "user/%s" id)
+                              {:first_name "Johnny" :last_name "Appleseed"})
+        (is (= {:topic    :user-update
```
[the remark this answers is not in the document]
escherize: I was using `=?` because some audit_log details have timestamps in them. So I was using them to just check that the timestamps were there, not that they were equal. Since this is straight up map equality `=` works.

aside: I think `=?` would also work, but best to save it for when it's needed so it's not confusing.

test/metabase/events/audit_log_test.clj
noahmoss: It shouldn't be needed here anymore

noahmoss (approved):

luizarakaki (approved):

escherize:
Handled by #35435

PR #35117, opened 2023-10-26
Block unallowed audit queries in QP permissions middleware

### LINKED ISSUE #34917: Modify QP to block queries on any tables besides views
Add a layer of protection to prevent the query builder to run queries against any table in the audit_db that are is not an audit view.

All Audit datasets are views and identified by the `v_` prefix

noahmoss:
Adds the `check-audit-db-permissions` defenterprise function which is called from the normal permissions middleware whenever a query is trying to hit the audit DB

- On OSS, it always rejects queries on the audit DB
- On Enterprise:
	-   It rejects all native queries
	-   It rejects any query targeting a table that isn't one of the statically-defined audit views

(These checks are done in addition to normal query permission checks)
  
 The OSS check requires access to the audit DB ID, so I've changed it from a defenterprise to a simple var in `metabase.models.permissions` called `audit-db-id`. (Not sure why we needed `default` in the name since it can't be changed.) Let me know if anyone objects to this change or has a different suggestion of where to put it.

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
+++ b/enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
@@ -13,18 +13,18 @@
 (def audit-db-view-names
   "Used for giving granular permissions into the audit db. Instead of granting permissions to
    all of the audit db, we query the audit db using the names of each view that starts with v_."
-  #{"v_users"
+  #{"v_audit_log"
     "v_content"
-    "v_group_members"
-    "v_alerts_subscriptions"
-    "v_audit_log"
     "v_dashboardcard"
-    "v_view_log"
+    "v_group_members"
+    "v_subscriptions"
+    "v_users"
+    "v_alerts"
     "v_databases"
-    "v_tables"
     "v_fields"
     "v_query_log"
-    "v_task_log"})
+    "v_tables"
+    "v_view_log"})
 
 (defenterprise check-audit-db-permissions
   "Checks that a given query is not a native query, and only includes table IDs corresponding to the audit views
--- a/enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
+++ b/enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
@@ -13,7 +13,7 @@
 (def audit-db-view-names
   "Used for giving granular permissions into the audit db. Instead of granting permissions to
    all of the audit db, we query the audit db using the names of each view that starts with v_."
-  #{"v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_audit_log", "v_dashboardcard"})
+  #{"v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_audit_log" "v_dashboardcard" "v_view_log" "v_databases" "v_tables" "v_fields" "v_query_log" "v_task_log"})
 
 (defenterprise check-audit-db-permissions
   "Checks that a given query is not a native query, and only includes table IDs corresponding to the audit views
--- a/enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
+++ b/enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
@@ -1,19 +1,30 @@
 (ns metabase-enterprise.audit-app.permissions
   (:require
-   [metabase-enterprise.audit-db
-    :refer [default-audit-collection]]
+   [metabase-enterprise.audit-db :refer [default-audit-collection]]
    [metabase.lib.metadata :as lib.metadata]
    [metabase.models.permissions :as perms]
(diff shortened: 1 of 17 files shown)


enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
```diff
@@ -1,28 +1,48 @@
...
    [metabase.public-settings.premium-features :refer [defenterprise]]
+   [metabase.query-processor.store :as qp.store]
    [metabase.shared.util.i18n :refer [tru]]
    [toucan2.core :as t2]))
 
 (def audit-db-view-names
   "Used for giving granular permissions into the audit db. Instead of granting permissions to
    all of the audit db, we query the audit db using the names of each view that starts with v_."
-  ["v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_activity"])
+  #{"v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_activity"})
```
luizarakaki, line 16: Is it possible to get these names programatically? I think they will have often.

Or a test that verifies that we are covering all v_* views in the database... We now have v_dashboardcard too

enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
```diff
@@ -1,28 +1,48 @@
...
+
+(defenterprise check-audit-db-permissions
+  "Checks that a given query is not a native query, and only includes table IDs corresponding to the audit views
+  listed above. (These should be the only tables present in the DB anyway, but this is an extra check as a fallback measure)."
+  :feature :audit-app
+  [query]
+  ;; query->source-table-ids returns a set of table IDs or the ::query-perms/native keyword
+  (let [table-ids-or-native-kw (query-perms/query->source-table-ids query)]
+    (qp.store/with-metadata-provider (:database-id query)
+      (doseq [table-id table-ids-or-native-kw]
```
escherize, line 26: Is this potentially calling doseq on a keyword?

noahmoss: no it'll always be a set, but the set can include IDs and/or a keyword

enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
```diff
@@ -1,28 +1,51 @@
...
    [metabase.public-settings.premium-features :refer [defenterprise]]
+   [metabase.query-processor.store :as qp.store]
    [metabase.shared.util.i18n :refer [tru]]
    [toucan2.core :as t2]))
 
 (def audit-db-view-names
   "Used for giving granular permissions into the audit db. Instead of granting permissions to
    all of the audit db, we query the audit db using the names of each view that starts with v_."
-  ["v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_activity"])
+  #{"v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_audit_log", "v_dashboardcard"})
```
luizarakaki, line 16:
```suggestion
  #{"v_users" "v_content" "v_group_members" "v_alerts_subscriptions" "v_audit_log" "v_dashboardcard" "v_view_log"})
```

enterprise/backend/test/metabase_enterprise/audit_app/permissions_test.clj
```diff
@@ -1,26 +1,92 @@
...
+  (testing "`audit-db-view-names` includes all views in the app DB prefixed with `v_`"
+    (let [view-query "SELECT table_name FROM information_schema.views WHERE table_name LIKE 'v\\_%';"]
+      (is (set/superset?
+           audit-app.permissions/audit-db-view-names
+           (into #{}
+                 (map :table_name (t2/query view-query))))))))
+
+(deftest audit-db-basic-query-test
+  (audit-db-test/with-audit-db-restoration
+   (audit-db/ensure-audit-db-installed!)
```
noahmoss, line 35: @escherize Do we have any better way to ensure that the audit DB is present in a test? I copied this pattern from other tests but it seems like it always forces the DB to be re-loaded which is pretty slow.

Could we use a macro that just checks whether an audit DB is present, and adds it if not, but otherwise does nothing?

escherize (approved):

enterprise/backend/test/metabase_enterprise/audit_app/permissions_test.clj
escherize: We could do that, the reason it's like this is to get a fresh version of it each time, but that might not make a difference anymore.

escherize: When something about the audit db changes, and you run the test we may get inconsistent results. This way we always test against the same freshly installed one.

PR #35130, opened 2023-10-27
QP enterprise middleware should use `defenterprise`

camsaul:
Another minor QP refactor to help simplify things a bit and set us up to port the QP to MLv2.

This PR updates all the EE QP middleware to use the `defenterprise` helper. `defenterprise` defines a function with two implementations; it will run one (OSS) version of a function unless EE code is available **and** we have a token with the required feature, in which case it will run the EE version of the function. This is cleaner than having to use `classloader/require` to try to load EE namespaces and then try to resolve vars all over the place


enterprise/backend/src/metabase_enterprise/advanced_permissions/query_processor/middleware/permissions.clj
```diff
@@ -92,24 +95,25 @@
...
-(defn limit-download-result-rows
+(defenterprise limit-download-result-rows
   "Post-processing middleware to limit the number of rows included in downloads if the user has `limited` download
   perms. Mainly useful for native queries, which are not modified by the [[apply-download-limit]] pre-processing
   middleware."
+  :feature :advanced-permissions
   [query rff]
   (if (and (is-download? query)
-           (premium-features/has-feature? :advanced-permissions)
            (= (current-user-download-perms-level query) :limited))
```
snoe, line 105: why remove the `premium-features/has-feature?` here but not in `apply-download-limit` or `check-download-permissions`?

snoe (approved):

enterprise/backend/src/metabase_enterprise/advanced_permissions/query_processor/middleware/permissions.clj
camsaul: Yeah, we should probably do it in those places too, we don't need it anymore because `defenterprise` does it automatically

PR #35136, opened 2023-10-27
Hide personal collections in question picker

WiNloSt:
This PR hides personal collections when adding questions to a dashboard in a public collection.

#### After
image

#### Before
image

#### Testing plan for milestone 1
Dashboards
1. Add a question to a dashboard in a public collection
    **expectation**:
    - hides all personal collections
    - show all questions
1. Add a question to a dashboard in a personal collection
    **expectation**:
    - shows all collections
    - show all questions

PR #35135, opened 2023-10-27
[milestone 1] Prevent users from adding questions in personal collections to dashboards in public collections

WiNloSt:
> **Note**
> All PRs have been approved, this is just the milestone PR. But feel free to test everything again.

This is milestone 1 for Epic: https://github.com/metabase/metabase/issues/34733

_The testing plan below is copied from the epic_

#### Testing plan for milestone 1
Dashboards
1. Add a question to a dashboard in a public collection
    **expectation**: hides all personal collections
1. Add a question to a dashboard in a personal collection
    **expectation**: shows all collections
1. move a dashboard from a personal collection to a public collection and test 1)
    **expectation**: hides all personal collections
1. move a dashboard from a public collection to a personal collection and test 2)
    **expectation**: shows all collections

Query builder

Newly created question
1. Select a dashboard for a question in a public collection
    **expectation**: shows all collections
1. Select a dashboard for a question in a personal collection
    **expectation**: hides all public collections
1. Open public collection for a question in a public collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. Open personal collection for a question in a public collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. Open public collection for a question in a personal collection (this only happens for the root collection, as other public collections are already hidden)
    **expectation**:
        - hides `+ Create a new dashboard` option
        - hides dashboards
1. Open personal collection for a question in a personal collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. When the question is in a public collection and the most recently viewed dashboard is in a public collection
    **expectation**:
        - shows the most recently viewed dashboard
1. When the question is in a public collection and the most recently viewed dashboard is in a personal collection
    **expectation**:
        - shows the most recently viewed dashboard
1. When the question is in a personal collection and the most recently viewed dashboard is in a public collection
    **expectation**:
        - shows the root collection instead
1. When the question is in a personal collection and the most recently viewed dashboard is in a personal collection
    **expectation**:
        - shows the most recently viewed dashboard

Existing question
1. Move a question from a personal collection to a public collection and test 1) from the **Newly created question** section
    **expectation**: shows all collections
1. Move a question from a public collection to a personal collection and test 2) from the **Newly created question** section
    **expectation**: hides all public collections
1. Test the rest of the scenarios from the **Newly created question** section

PR #35137, opened 2023-10-27
Hide public collections in dashboard picker

WiNloSt:
This PR:
- hides all public collections when selecting a dashboard for a question in a personal collection.
- hides `+ Create a new dashboard` option at the root collection when selecting a dashboard for a question in a personal collection.
- hides dashboards in the root collection when selecting a dashboard for a question in a personal collection.
- shows the root collection instead of the most recently viewed dashboard's collection if selecting a dashboard for a question in a personal collection and the most recently viewed dashboard is in a public collection

#### After
image

####  Before
image

#### Testing plan for milestone 1
Newly created question
1. Select a dashboard for a question in a public collection
    **expectation**: shows all collections
1. Select a dashboard for a question in a personal collection
    **expectation**: hides all public collections
1. Open public collection for a question in a public collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. Open personal collection for a question in a public collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. Open public collection for a question in a personal collection (this only happens for the root collection, as other public collections are already hidden)
    **expectation**:
        - hides `+ Create a new dashboard` option
        - hides dashboards
1. Open personal collection for a question in a personal collection
    **expectation**:
        - shows `+ Create a new dashboard` option
        - shows dashboards
1. When the question is in a public collection and the most recently viewed dashboard is in a public collection
    **expectation**:
        - shows the most recently viewed dashboard
1. When the question is in a public collection and the most recently viewed dashboard is in a personal collection
    **expectation**:
        - shows the most recently viewed dashboard
1. When the question is in a personal collection and the most recently viewed dashboard is in a public collection
    **expectation**:
        - shows the root collection instead
1. When the question is in a personal collection and the most recently viewed dashboard is in a personal collection
    **expectation**:
        - shows the most recently viewed dashboard


WiNloSt:
* master
  * **PR #35135** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35135" target="_blank">Graphite</a> 
    * **PR #35136** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35136" target="_blank">Graphite</a> 
      * **PR #35137** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35137" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35137?utm_source=stack-comment).

frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.tsx
```diff
@@ -35,26 +34,38 @@ interface AddToDashSelectDashModalProps {
...
   });
 
+  const [openCollectionId, setOpenCollectionId] = useState<
+    CollectionId | undefined
+  >();
+  const openCollection = useSelector(state =>
+    getCollection(state, { collectionId: openCollectionId }),
+  );
+  const isOpenCollectionInPersonalCollection = openCollection?.is_personal;
+  const hideCreateNewDashboardOption =
```
WiNloSt, line 66: This only happens at the root collection. Because public collections will already be hidden when selecting a dashboard for a question in a personal collection.

frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.unit.spec.tsx
```diff
@@ -44,11 +44,6 @@ const DASHBOARD_AT_ROOT = createMockDashboard({
   model: "dashboard",
 });
 
-const DASHBOARDS_BY_ID = {
```
WiNloSt, line 47: Doesn't seem to be used. Removing it doesn't fail any test.

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
```diff
@@ -0,0 +1,47 @@
+import { useAsync } from "react-use";
+import type { CollectionId, Dashboard } from "metabase-types/api";
+import { ActivityApi } from "metabase/services";
+import { ROOT_COLLECTION } from "metabase/entities/collections";
+import { coerceCollectionId } from "metabase/collections/utils";
+
+export const useMostRecentlyViewedDashboard = () => {
```
WiNloSt, line 7: old hook

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  return { data, isLoading, error };
+};
+
+interface UseCollectionIdProps {
+  isQuestionInPersonalCollection: boolean;
+  mostRecentlyViewedDashboard: Dashboard | undefined;
+}
+
+export const useCollectionId = ({
```
WiNloSt, line 27: new hook

frontend/src/metabase/containers/DashboardPicker.tsx
```diff
@@ -1,7 +1,12 @@
+import type { ComponentProps } from "react";
 import type { CollectionId, DashboardId } from "metabase-types/api";
 import ItemPicker from "./ItemPicker";
 
-export interface DashboardPickerProps {
+export interface DashboardPickerProps
+  extends Pick<
+    ComponentProps<typeof ItemPicker>,
+    "showOnlyPersonalCollections" | "onOpenCollectionChange"
```
WiNloSt, line 8: This change allows TS files that use `DashboardPicker` to pass these 2 new props
1. `showOnlyPersonalCollections`
2. `onOpenCollectionChange`

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -243,8 +256,10 @@ function ItemPicker<TId>({
         checkHasWritePermissionForItem={checkHasWritePermissionForItem}
         getCollectionIcon={getCollectionIcon}
         style={style}
-        // personal is a fake collection for admins that contains all other user's collections
-        allowFetch={openCollectionId !== "personal"}
+        allowFetch={
+          // "personal" is a fake collection for admins that contains all other user's collections
+          openCollectionId !== "personal" && !hideItems
```
WiNloSt, line 261: This way, we don't need to modify `GET /api/collection/:id/items` to exclude items from public collections because we know from the start we don't want to render any items.

frontend/src/metabase/query_builder/components/QueryModals.tsx
```diff
@@ -13,7 +13,7 @@ import Modal from "metabase/components/Modal";
 
 import { SaveQuestionModal } from "metabase/containers/SaveQuestionModal";
 import QuestionSavedModal from "metabase/components/QuestionSavedModal";
-import AddToDashSelectDashModal from "metabase/containers/AddToDashSelectDashModal";
+import { ConnectedAddToDashSelectDashModal } from "metabase/containers/AddToDashSelectDashModal";
```
WiNloSt, line 16: Fix default export

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
ranquild: nit: improve naming, e.g. add some more info about what the "collection" is here

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -100,6 +102,13 @@ function ItemPicker<TId>({
   const isPickingNotCollection = models.some(model => model !== "collection");
 
   const openCollection = collectionsById[openCollectionId];
+  const isOpenCollectionInPersonalCollection = openCollection?.is_personal;
+  const hideItems =
+    showOnlyPersonalCollections && !isOpenCollectionInPersonalCollection;
+
+  useEffect(() => {
```
ranquild, line 109: Why this is needed? The previous solution with calling `onOpenCollectionChange` directly was more straightforward. If there is a reason, can we avoid using `useEffect`?

ranquild (commented):
Some comments, please take a look

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -33,6 +33,7 @@ interface OwnProps<TId> {
   entity?: typeof Collections; // collections/snippets entity
   showSearch?: boolean;
   showScroll?: boolean;
+  showOnlyPersonalCollections?: boolean;
```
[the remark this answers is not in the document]
WiNloSt: I'm not sure we'll need that yet. `ItemPicker` is only used in `DashboardPicker` (Query builder) while `QuestionPicker` (dashboards) uses totally different components.

We might need to come up with something else when dealing with [hiding personal collections/dashboards/and questions when configuring click behavior](https://metaboat.slack.com/archives/C057T1QTB3L/p1698664681662569?thread_ts=1698664221.417809&cid=C057T1QTB3L) though since that uses `ItemPicker`.

What about we leave it like this and when we're about to work on the click behavior we could change it to something similar [on the BE](https://github.com/metabase/metabase/pull/34887)?

The BE search endpoint now accepts `filter_items_in_personal_collection` with either `exclude` or `only`. In FE's case, it's like an enum with 3 states. `null`, `exclude`, `only` where
- `null`: renders everything
- `exclude`: renders only items in public collections
- `only`: renders only items in personal collections

What do you think?

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
WiNloSt: Yeah it's just that `setOpenCollectionId` is passed to many places. Previously `onOpenCollectionChange` was only called when calculating breadcrumbs but not on the collection change.

> If there is a reason, can we avoid using `useEffect`?

Yeah, I think we could encapsulate `setOpenCollectionId`, by wrapping that in a new function that would call both `setOpenCollectionId` and `onOpenCollectionChange` at the same time.

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
WiNloSt: How about `useInitialOpenCollectionId`. We have the same variable name elsewhere which does exactly the same thing.

https://github.com/metabase/metabase/blob/13424d36f2fef49c1884a7f917a1e1febb337be0/frontend/src/metabase/collections/containers/FormCollectionPicker/FormCollectionPicker.tsx#L34

It conveys that this is the value for the open collection when first rendering the picker.

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
WiNloSt: Hmm, not sure why I used `useEffect` it seems `setOpenCollectionId` is only called in 2 places. I'll inline `onOpenCollectionChange` back then.

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
WiNloSt: Fixed in 7f631ecb94be97214aeb1bc6c2233760a1e93a3e

frontend/src/metabase/containers/DashboardPicker.tsx
WiNloSt: Fixed in https://github.com/metabase/metabase/pull/35137/commits/64bacdcc46001605045bffc7bbde895e13035c76

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
WiNloSt: Fixed in 6d7c8527f88c6856f0d6db5a53cd10300d2eb459

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -100,6 +102,11 @@ function ItemPicker<TId>({
   const isPickingNotCollection = models.some(model => model !== "collection");
 
   const openCollection = collectionsById[openCollectionId];
+  const isOpenCollectionInPersonalCollection = openCollection?.is_personal;
+  const hideItems =
+    !searchString &&
+    showOnlyPersonalCollections &&
+    !isOpenCollectionInPersonalCollection;
```
WiNloSt, line 109: We want to still enable fetch when `searchString` is not null, meaning users are searching for dashboards.
image

But we don't want to fetch anything when opening a public collection (only root collection in this case), and is selecting a dashboard for a question in a personal collection.
image

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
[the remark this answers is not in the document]
ranquild: Good point about dashboards using a different component. Feel free to leave it as it is for now

frontend/src/metabase-types/api/mocks/card.ts
```diff
@@ -10,29 +10,36 @@ import type {
...
+    description: null,
+    display: "table",
+    public_uuid: null,
+    dataset_query: createMockStructuredDatasetQuery(),
+    visualization_settings: createMockVisualizationSettings(),
+    result_metadata: [],
+    dataset: false,
+    can_write: true,
+    cache_ttl: null,
+    collection: rootCollection,
```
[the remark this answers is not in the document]
ranquild: We should not ever import anything from the app in `metabase-types`. Furthermore, "root" collection would never by hydrated, it would be `null`.

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
ranquild: And one more thing. It's usually easier to follow boolean variable names if they are in "positive" form. E.g. `showItems`, `allowFetch`, `canShowItems` etc. and not `hideItems`, `disableFetch`

frontend/src/metabase-types/api/card.ts
```diff
@@ -14,6 +15,7 @@ export interface Card<Q = DatasetQuery> extends UnsavedCard<Q> {
   can_write: boolean;
 
   database_id?: DatabaseId;
+  collection?: Collection;
```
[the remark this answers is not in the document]
ranquild: Should it be `collection?: Collection | null;` to handle root collections?

ranquild (commented):
@WiNloSt left some comments, please take a look. Main point - I think there is an issue with how the root collection is handled, especially in types.

frontend/src/metabase-types/api/mocks/card.ts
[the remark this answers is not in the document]
WiNloSt: Not every endpoint sends `null` for root collection. At least for card endpoints, we got the actual root collection.

Here's the response from `GET /api/card/:id`
image

I found that for dashboard endpoints they're sending `null` though.

I think this kind of consistency is very confusing.

Should we always assume a root collection could be `null` and always default that to `ROOT_COLLECTION` when not present everywhere? In this case the dashboard picker will always assume `card.collection` is present.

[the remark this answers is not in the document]
WiNloSt: If we assume the root `collection` in every endpoint is `null` then I could remove the import.

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  return { data, isLoading, error };
+};
+
+interface UseInitialOpenCollectionIdProps {
+  isQuestionInPersonalCollection: boolean;
+  mostRecentlyViewedDashboard: Dashboard | undefined;
+}
+
+export const useInitialOpenCollectionId = ({
```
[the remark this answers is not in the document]
WiNloSt: Good point. I'll just change this to function.

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
WiNloSt: But `hideItems` consists of 3 other conditions
```ts
!searchString &&
showOnlyPersonalCollections &&
!isOpenCollectionInPersonalCollection
```

It might not be apparent why `searchString` is needed, because we're using `Search.ListLoader` for 2 scenarios:
1. when we're not searching:
  - we want `allowFetch` to be false when we're viewing public collections (only root collection in this case) and is on a question in a personal collection.
2. when we are searching:
  - we want `allowFetch` to be `true`, so `Search.ListLoader` could search the dashboards.

So I'm worried inlining those 3 conditions might make it harder to understand. What about renaming `hideItems` to `showItems` and flip the condition instead?

frontend/src/metabase-types/api/mocks/card.ts
```diff
@@ -10,29 +10,36 @@ import type {
   TableColumnOrderSetting,
 } from "metabase-types/api";
 
+// Do not import from "metabase/entities/collections" to avoid circular dependency
+import { ROOT_COLLECTION } from "metabase/entities/collections/constants";
```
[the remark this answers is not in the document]
WiNloSt: [commented below](https://github.com/metabase/metabase/pull/35137/files#r1377414161), if we're fine treating `card` as if the root collection would be null (while BE actually sends the root collection in card endpoints) then this could be removed.

frontend/src/metabase-types/api/card.ts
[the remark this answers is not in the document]
WiNloSt: [As in my other comment](https://github.com/metabase/metabase/pull/35137/files#r1377414161), if we're fine treating `card` as if the root collection would be null (while BE actually sends the root collection in card endpoints) then I could say this is nullable.

frontend/src/metabase/containers/AddToDashSelectDashModal/hooks.ts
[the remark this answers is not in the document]
WiNloSt: Fixed in d0dfc8d954a5abae0a8d7e88fd113f61ffe4e725

frontend/src/metabase-types/api/mocks/card.ts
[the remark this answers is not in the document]
WiNloSt: Fixed in 536582175599dcf0d9cd12880fc76eadcbdb3151

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
WiNloSt: Fixed in 2f662c4d5aa75f2521aa422ff750cc21aa12ef3f

frontend/src/metabase-types/api/collection.ts
```diff
@@ -63,7 +63,7 @@ export interface CollectionItem {
   collection_position?: number | null;
   collection_preview?: boolean | null;
   fully_parametrized?: boolean | null;
-  collection?: Collection;
+  collection?: Collection | null;
```
WiNloSt, line 66: This change is required to match the `collection` type to the one in `frontend/src/metabase-types/api/card.ts`

frontend/src/metabase-types/api/mocks/card.ts
```diff
@@ -10,36 +10,30 @@ import type {
...
+  description: null,
+  display: "table",
+  public_uuid: null,
+  dataset_query: createMockStructuredDatasetQuery(),
+  visualization_settings: createMockVisualizationSettings(),
+  result_metadata: [],
+  dataset: false,
+  can_write: true,
+  cache_ttl: null,
+  collection: null,
```
WiNloSt, line 30: Revert the change back to `master` and add `collection: null`

ranquild (approved):

PR #35146, opened 2023-10-27
Snowplow: Add new event for search results filtered

### LINKED ISSUE #35122: Snowplow: Add new event for search results filtered
## Context
We want to know if people are using search filters. Add a new event `search_results_filtered` to the `Search` Snowplow schema. 

[More details here](https://www.notion.so/metabase/Search-results-filtered-f64a868f3901451ead4187f26bf9a5a3?pvs=4).

oisincoveney:
Closes https://github.com/metabase/metabase/issues/35122

### Description

Adds snowplow schema, events, and e2e tests for tracking when search filters are used within the search app.

### How to verify

1. Ensure that Snowplow is set up according to the [Snowplow integration] page (https://www.notion.so/metabase/Snowplow-integration-5da1f874beda4153b4fccfa6c1e77caa?pvs=4)
2. Go to the search app, and type in a search
3. Apply any filter
4. Access Snowplow, and see that there are no bad events, and that there is a new `search_results_filtered` event

### Checklist


snowplow/iglu-client-embedded/schemas/com.metabase/search/jsonschema/1-0-1
```diff
@@ -0,0 +1,69 @@
...
+      "type": [
+        "array",
+        "null"
+      ],
+      "items": {
+        "type": "string"
+      }
+    },
+    "creator": {
+      "description": "Filter for created by a user",
```
oisincoveney, line 41: @cdeweyx let me know what you'd like to see for these descriptions (if they need to be changed, anyway)

e2e/support/helpers/e2e-snowplow-helpers.js
```diff
@@ -40,6 +41,41 @@ export const expectGoodSnowplowEvents = count => {
     .should("have.length", count);
 };
 
+export const expectGoodSnowplowEventCount = expectedEventCounts => {
```
oisincoveney, line 44: this is super experimental and want to get y'all's opinion on whether this makes sense or not. I think it's better to be explicit about the events that we're asserting on (since in search, we have `new_search_query` and `search_results_filtered`) and I wanted a way to differentiate between the two

e2e/test/scenarios/onboarding/search/search.cy.spec.js
```diff
@@ -1092,7 +1099,384 @@ describeWithSnowplow("scenarios > search", () => {
...
+          .click();
+
+        eventCount +=
+          PAGE_VIEW_EVENT +
+          SEARCH_RESULT_FILTERED_EVENT +
+          NEW_SEARCH_QUERY_EVENT;
+
+        expectGoodSnowplowEvents(eventCount);
+
+        expectGoodSnowplowEventCount({
```
oisincoveney, line 1421: The experimental `expectGoodSnowplowEventCount` in action

e2e/test/scenarios/onboarding/search/search.cy.spec.js
```diff
@@ -1092,7 +1099,384 @@ describeWithSnowplow("scenarios > search", () => {
...
+            NEW_SEARCH_QUERY_EVENT +
+            SEARCH_RESULT_FILTERED_EVENT,
+        );
+      });
+    });
+
+    describe("type filter", () => {
+      it("should send a snowplow event when a search filter is used in the URL", () => {
+        cy.visit("/search?q=orders&type=card");
+        // TODO: this should return +2 events - new_search_query event isn't firing
```
oisincoveney, line 1120: For some reason the `new_search_query` event isn't firing when we use the `type` filter - I'll need to check with the BE on this one since they fire the `new_search_query` event

e2e/support/helpers/e2e-snowplow-helpers.js
```diff
@@ -40,6 +41,40 @@ export const expectGoodSnowplowEvents = count => {
...
+          }
+          return acc;
+        }, {});
+
+      return {
+        page_view: pageViewCount,
+        ...events,
+      };
+    })
+    .should("deep.include", expectedEventCounts);
```
oisincoveney, line 75: we shouldn't use `deep.equals` because there might be other events that we don't care about

e2e/test/scenarios/onboarding/search/search.cy.spec.js
```diff
@@ -1076,6 +1077,12 @@ describe("scenarios > search", () => {
 
 describeWithSnowplow("scenarios > search", () => {
   const PAGE_VIEW_EVENT = 1;
+  const NEW_SEARCH_QUERY_EVENT = 1;
```
oisincoveney, line 1080: Just put in some variables so I could think about the events a little more clearly. Let me know if I should revert to numbers

frontend/src/metabase/search/analytics.ts
```diff
@@ -0,0 +1,35 @@
...
+}) => {
+  let content_type = null;
+
+  if (searchFilters.type) {
+    content_type = Array.isArray(searchFilters.type)
+      ? searchFilters.type
+      : [searchFilters.type];
+  }
+
+  const filterBooleanPayload: { [key: string]: unknown } = {
```
oisincoveney, line 17: The rest of these variables are booleans so we can just run the same `!!` operation on all of them.

e2e/support/helpers/e2e-snowplow-helpers.js
oisincoveney: The only thing I'm worried about is that chain of `event.event?.unstruct_event?.data?.data` - I'm trying to figure out if this will be a reliable check

e2e/test/scenarios/onboarding/search/search.cy.spec.js
```diff
@@ -1092,7 +1099,384 @@ describeWithSnowplow("scenarios > search", () => {
...
+    describe("type filter", () => {
+      it("should send a snowplow event when a search filter is used in the URL", () => {
+        cy.visit("/search?q=orders&type=card");
+        // TODO: this should return +2 events - new_search_query event isn't firing
+        expectGoodSnowplowEvents(
+          PAGE_VIEW_EVENT + SEARCH_RESULT_FILTERED_EVENT,
+        );
+      });
+
+      it("should send a snowplow event when a search filter is applied from the UI", () => {
```
WiNloSt, line 1126: This test failed both locally and on CI

frontend/src/metabase/search/analytics.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { URLSearchFilterQueryParams } from "metabase/search/types";
+
+export const trackSearchEvents = ({
+  searchFilters,
+}: {
+  searchFilters: URLSearchFilterQueryParams;
+}) => {
+  let content_type = null;
+
+  if (searchFilters.type) {
```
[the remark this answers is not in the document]
WiNloSt: It's just my personal preference that I try to avoid reassignments at all. These few lines of code could be replaced by a function call. I found it feel uneasy to read `let` because I need to always be aware of when it could be reassigned. But a `const` guarantees that the value is only assigned here once.

```ts
const contentType = getContentType(searchFilters.type)
...
function getContentType(type) {
  if (type) {
    returns Array.isArray(type)? type: [type]
  }

  return null
}
```

what do you think?

frontend/src/metabase/search/analytics.ts
WiNloSt: `Record` usage is more established. So I suggest we follow that.

```suggestion
  const filterBooleanPayload: Record<string, unknown> = {
```

frontend/src/metabase/search/analytics.ts
```diff
@@ -0,0 +1,35 @@
...
+    last_editor: searchFilters.last_edited_by,
+    creation_date: searchFilters.created_at,
+    last_edit_date: searchFilters.last_edited_at,
+    verified_items: searchFilters.verified,
+    search_native_queries: searchFilters.search_native_query,
+  };
+
+  Object.entries(searchFilters).forEach(([key, value]) => {
+    filterBooleanPayload[key] = !!value;
+  });
```
WiNloSt, line 28: Instead of casting them to boolean here, what about casting them since the creation of `filterBooleanPayload`? It's not like we're casting them based on any condition or anything here.

```
  const filterBooleanPayload: { [key: string]: unknown } = {
    creator: !!searchFilters.created_by,
    last_editor: !!searchFilters.last_edited_by,
    creation_date: !!searchFilters.created_at,
    last_edit_date: !!searchFilters.last_edited_at,
    verified_items: !!searchFilters.verified,
    search_native_queries: !!searchFilters.search_native_query,
  };
```

Although I still find using the boolean constructor more readable e.g. `Boolean(searchFilters.created_by)`, `!!` is still fine for me.

WiNloSt (commented):
Looks good to me, though it'd be nice to double check with BE regarding the tests that doesn't count the event correctly.

e2e/support/helpers/e2e-snowplow-helpers.js
oisincoveney: Oh ok that's perfect - I obviously couldn't find it but I'd rather use something that's already there than something new

frontend/src/metabase/search/analytics.ts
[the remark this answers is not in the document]
oisincoveney: Understandable - I kept it in the function because I didn't feel it was "function-worthy" in terms of length and usage in any other part of the code. But I think it still looks nicer so I'll change it

frontend/src/metabase/search/analytics.ts
oisincoveney: I think the constructor is the best one here. I actually had it like your suggestion before, but I didn't want to repeat myself too much, but I guess with this case where it's set in stone it's ok

e2e/test/scenarios/onboarding/search/search.cy.spec.js
oisincoveney: Solved

frontend/src/metabase/search/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+    creator: Boolean(searchFilters.created_by),
+    last_editor: Boolean(searchFilters.last_edited_by),
+    creation_date: Boolean(searchFilters.created_at),
+    last_edit_date: Boolean(searchFilters.last_edited_at),
+    verified_items: Boolean(searchFilters.verified),
+    search_native_queries: Boolean(searchFilters.search_native_query),
+    content_type: getContentType(searchFilters.type),
+  };
+
+  // we just want to track search filters if the user is using them
```
oisincoveney, line 40: Based on [this thread](https://metaboat.slack.com/archives/C057T1QTB3L/p1698398569994129?thread_ts=1698357917.957319&cid=C057T1QTB3L) I need to add this

e2e/test/scenarios/onboarding/search/search.cy.spec.js
oisincoveney: @qnkhuat would you know what's happening here? I've tried the other filters locally and with this test env and the type filter is the only filter that doesn't trigger `new_search_query`

e2e/support/helpers/e2e-snowplow-helpers.js
npretto: I also didn't notice that function, I will use it in my pr thanks.
One small thing i noticed though is that if we pass the wrong event name the e2e just fails with "Snowplow retry timeout" which I'm afraid will be confused with a flake instead of an actual failed test

e2e/test/scenarios/onboarding/search/search.cy.spec.js
ranquild: Please use `expectGoodSnowplowEvent`, ignore the helper that checks the number.

src/metabase/analytics/snowplow.clj
```diff
@@ -142,7 +142,7 @@
    ::database     "1-0-0"
    ::instance     "1-1-2"
    ::metabot      "1-0-1"
-   ::search       "1-0-0"
+   ::search       "1-0-1"
```
ranquild, line 145: @npretto please also update the version like this in your PR

frontend/src/metabase/search/analytics.ts
```diff
@@ -0,0 +1,38 @@
...
+function getContentType(
+  type: SearchQueryParamValue,
+): Array<SearchQueryParamValue> | null {
+  if (type) {
+    return Array.isArray(type) ? type : [type];
+  }
+  return null;
+}
+
+export const trackSearchEvents = ({
```
ranquild, line 16: nit: Why pass an object here with just one parameter?

e2e/test/scenarios/onboarding/search/search.cy.spec.js
qnkhuat: yea in BE we explictly track when there are no type filters [here](https://github.com/metabase/metabase/blob/3f2afbf9fe2e7eb2f152a50d89e5f15304cb0776/src/metabase/api/search.clj#L518).

It was from before we have this type filter and the assumption is that we only track the search from the global search view. search from other components like selecting cards to add to a dashboard shouldn't get tracked.

but now with the new filters, I think we should be smarter. wdyt if we add a new flag `is_global_search`, and FE fill it with `true` on the global search view so we can track it correctly?

snowplow/iglu-client-embedded/schemas/com.metabase/search/jsonschema/1-0-1
```diff
@@ -0,0 +1,89 @@
...
+    },
+    "content_type": {
+      "description": "Filter for entity types (question, database, model, etc.)",
+      "type": [
+        "array",
+        "null"
+      ],
+      "items": {
+        "description": "An entity type (card, question, dashboard, etc.)",
+        "type": "string",
```
ranquild, line 38: Wdyt about adding a enum here? Otherwise we can easily get incorrect results in Redshift

ranquild (changes requested):
Please use `expectGoodSnowplowEvent` helper in e2e tests with an event name; this event count logic is hard to follow

e2e/support/helpers/e2e-snowplow-helpers.js
WiNloSt: > i noticed though is that if we pass the wrong event name the e2e just fails with "Snowplow retry timeout" which I'm afraid will be confused with a flake instead of an actual failed test

Agree, the error message could have been better. It's not in the scope of this PR, but I wouldn't mind if you want to make that happen 😄

e2e/test/scenarios/onboarding/search/search.cy.spec.js
oisincoveney: Maybe we can just remove that extra check for the `models` parameter for now? We'll need to check in with PDs if we want to make more changes

e2e/test/scenarios/onboarding/search/search.cy.spec.js
oisincoveney: Yep I agree, I really didn't like writing this

e2e/test/scenarios/onboarding/search/search.cy.spec.js
oisincoveney: Hm... the more I think about it, the more I'm convinced that we should probably have an enum or a flag to tell us where we are searching. @cdeweyx what do you think about having a `location` flag with `global-search` or `search-bar` or something? Since the `new_search_query` event fires when a user is using the search bar, and we probably want to see how often a user is using the filters the global search (against using no filters in the global search), I think it would help us differentiate the events a little more easily

frontend/src/metabase/search/analytics.ts
oisincoveney: Ah I did that out of habit. I'll remove it here

snowplow/iglu-client-embedded/schemas/com.metabase/search/jsonschema/1-0-1
oisincoveney: yes! good call

oisincoveney:
* master
  * **PR #35146** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35146" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35146?utm_source=stack-comment).

snowplow/iglu-client-embedded/schemas/com.metabase/search/jsonschema/1-0-1
oisincoveney: Added

WiNloSt: I [updated the spec](https://www.notion.so/metabase/Search-results-filtered-f64a868f3901451ead4187f26bf9a5a3) to make sure they're in sync

e2e/test/scenarios/onboarding/search/search.cy.spec.js
```diff
@@ -1090,9 +1092,294 @@ describeWithSnowplow("scenarios > search", () => {
...
+  describe("should send snowplow events for each filter when it is applied and removed", () => {
+    describe("no filters", () => {
+      it("should not send a snowplow event when a search with no filters is accessed from the URL", () => {
+        cy.visit("/search?q=orders");
+        expectGoodSnowplowEvent({ event: NEW_SEARCH_QUERY_EVENT_NAME }, 1);
+        expectGoodSnowplowEvent({ event: SEARCH_RESULTS_FILTERED_NAME }, 0);
+      });
+    });
+
+    describe("type filter", () => {
```
WiNloSt, line 1108: should we assert other properties too? e.g. `content_type`, `creator`, etc. All the tests below seems to only test that an `event` exist, but didn't test that they have other properties too.

oisincoveney: I think it's something we can add but in a different PR? we can get this one in for the PDs and then add some rigor

e2e/test/scenarios/onboarding/search/search.cy.spec.js
cdeweyx: Yeah I think that's a good idea. For clarity, we're distinguishing between searching globally using the search bar in the top nav vs. searching with filters? If so then I'd say something like a `location` property with `global_search` or `filtered_search` would do the trick.

e2e/test/scenarios/onboarding/search/search.cy.spec.js
WiNloSt: sounds good.

WiNloSt:
@oisincoveney Please re-request my review when you added `is_global_search: true` on the FE

snowplow/iglu-client-embedded/schemas/com.metabase/search/jsonschema/1-0-1
cdeweyx: Looks reasonable to me, I don't know of any way these are exposed in the schema itself (at least that I use in practice) so feel free to whatever makes this clearest for future engineers iterating on this. Nothing here looks wrong to me.

cdeweyx (approved):

src/metabase/api/search.clj
```diff
@@ -558,12 +559,30 @@
...
-    ;; Only track global searches
-    (when (and (nil? models)
-               (nil? table_db_id)
-               (not archived))
-      (snowplow/track-event! ::snowplow/new-search-query api/*current-user-id* {:runtime-milliseconds duration}))
+        duration   (- (System/currentTimeMillis) start-time)
+        has-advanced-filters (some some?
+                                   [models created_by created_at last_edited_by
+                                    last_edited_at search_native_query verified])]
+    (when (not (nil? context))
```
calherries, line 566: this is equivalent
```suggestion
    (when (some? context))
```

src/metabase/api/search.clj
```diff
@@ -558,12 +559,30 @@
...
+      (cond
+        (and
+         (nil? table_db_id)
+         (not archived))
```
calherries, line 570: this is non-standard formatting, just have them on one line
```suggestion
        (and (nil? table_db_id) (not archived))
```

src/metabase/api/search.clj
```diff
@@ -558,12 +559,30 @@
...
+        (snowplow/track-event! ::snowplow/new-search-query api/*current-user-id*
+                               {:runtime-milliseconds duration
+                                :context              context}))
+      (cond
+        has-advanced-filters
```
calherries, line 575: `cond` is generally used when there are multiple clauses
```suggestion
      (when has-advanced-filters
```

src/metabase/api/search.clj
```diff
@@ -558,12 +559,30 @@
...
+        (snowplow/track-event! ::snowplow/search-results-filtered api/*current-user-id*
+                               {:runtime-milliseconds  duration
+                                :content-type          (u/one-or-many models)
```
calherries, line 578: why isn't this just called models? like
```suggestion
                                :models          (u/one-or-many models)
```

src/metabase/api/search.clj
```diff
@@ -558,12 +559,30 @@
...
+                                :creator               (some? created_by)
+                                :creation_date         (some? created_at)
+                                :last_editor           (some? last_edited_by)
+                                :last-edit-date        (some? last_edited_at)
```
calherries, line 582: I know the casing is converted to snake_case anyway, but this is clojure so lets be consistent and use kebab-case
```suggestion
                                :last-editor           (some? last_edited_by)
                                :last-edit-date        (some? last_edited_at)
```

src/metabase/api/search.clj
calherries, line 570: This is unconventional formatting and `when` should be used because there's only one branch
```suggestion
      (when (and (nil? table_db_id) (not archived))
```

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
...
-    (snowplow-test/with-fake-snowplow-collector
-      (mt/user-http-request :crowberto :get 200 "search?q=test")
-      (is (=? {:data {"event"                "new_search_query"
-                      "runtime_milliseconds" pos?}
+  (testing "Send a snowplow event when a search query is triggered and context is passed"
+           (snowplow-test/with-fake-snowplow-collector
+            (mt/user-http-request :crowberto :get 200 "search?q=test" :context "global-search")
+            (is
+             (=?
+              {:data    {"event"                "new_search_query"
```
calherries, line 820: formatting
```suggestion
            (is (=? {:data    {"event"                "new_search_query"
```

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
...
+            (mt/user-http-request :crowberto :get 200 "search?q=test" :context "filtered-search")
+            (is
+             (=?
+              {:data    {"event"                "new_search_query"
+                         "runtime_milliseconds" pos?
+                         "context"              "filtered-search"}
+               :user-id (str (mt/user->id :crowberto))}
+              (last (snowplow-test/pop-event-data-and-user-id!))))))
+
+  (testing "Don't send a snowplow event if the search doesn't contain context"
```
calherries, line 835: Is this actually what we want to happen? why wouldn't we want to always record a search event

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
...
+            (mt/user-http-request :crowberto :get 200 "search" :q "test" :models "table")
+            (is (empty? (snowplow-test/pop-event-data-and-user-id!)))
+
+            (mt/user-http-request :crowberto :get 200 "search" :q "test" :table_db_id (mt/id))
+            (is (empty? (snowplow-test/pop-event-data-and-user-id!)))
+
+            (mt/user-http-request :crowberto :get 200 "search" :q "test" :archived true)
+            (is (empty? (snowplow-test/pop-event-data-and-user-id!)))))
+
+  (testing "even when context is global-search but query has a models filter"
```
calherries, line 846: same question.

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
...
+           (is (empty? (snowplow-test/pop-event-data-and-user-id!)))))
+
+(deftest snowplow-search-results-filtered-event-test
+  (testing "Send a snowplow event when a new filtered search query is made"
+           (snowplow-test/with-fake-snowplow-collector
+            (mt/user-http-request :crowberto :get 200 "search" :q "test" :context "filtered-search"
+                                  :models                         "card")
+            (is
+             (=?
+              {:data    {"event"                 "search_results_filtered"
```
calherries, line 857:
```suggestion
            (is (=? {:data    {"event"                 "search_results_filtered"
```

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
...
+
+           (snowplow-test/with-fake-snowplow-collector
+            (mt/user-http-request :crowberto :get 200 "search" :q "test" :context "filtered-search"
+                                  :models                         "card" :models "dashboard"
+                                  :created_at                     "2000-01-01" :created_by (mt/user->id :crowberto)
+                                  :last_edited_at                 "2000-01-01" :last_edited_by (mt/user->id :crowberto)
+                                  :search_native_query            true)
+            (is
+             (=?
+              {:data    {"event"                 "search_results_filtered"
```
calherries, line 878:
```suggestion
            (is (=? {:data    {"event"                 "search_results_filtered"
```

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
         (is (= 11 (call-count)))))))
 
 (deftest snowplow-new-search-query-event-test
-  (testing "Send a snowplow event when a new global search query is made"
-    (snowplow-test/with-fake-snowplow-collector
-      (mt/user-http-request :crowberto :get 200 "search?q=test")
-      (is (=? {:data {"event"                "new_search_query"
-                      "runtime_milliseconds" pos?}
+  (testing "Send a snowplow event when a search query is triggered and context is passed"
+           (snowplow-test/with-fake-snowplow-collector
```
calherries, line 816: Something's wrong with your calva auto-indentation configuration. `testing` blocks should have two space indentation. 
```suggestion
  (testing "Send a snowplow event when a search query is triggered and context is passed"
    (snowplow-test/with-fake-snowplow-collector
```
Either ask a BE eng to clean this up (I can do it) or huddle with someone to fix it

test/metabase/api/search_test.clj
```diff
@@ -812,23 +812,96 @@
...
+                         "context"               "filtered-search"}
+               :user-id (str (mt/user->id :crowberto))}
+              (last (snowplow-test/pop-event-data-and-user-id!)))))
+
+           (snowplow-test/with-fake-snowplow-collector
+            (mt/user-http-request :crowberto :get 200 "search" :q "test" :context "filtered-search"
+                                  :models                         "card" :models "dashboard"
+                                  :created_at                     "2000-01-01" :created_by (mt/user->id :crowberto)
+                                  :last_edited_at                 "2000-01-01" :last_edited_by (mt/user->id :crowberto)
+                                  :search_native_query            true)
```
calherries, line 875: This formatting is horrendous. Seriously what

test/metabase/api/search_test.clj
oisincoveney: We only want to record search API calls that are triggered by the search bar or the search app. The data picker in the notebook and Metabot also trigger search, but we don't care about those

test/metabase/api/search_test.clj
oisincoveney: this one... shouldn't exist anymore

src/metabase/api/search.clj
oisincoveney: The keys match the labels of the search filters - in hindsight the keys should probably have been uniform throughout but that's a conversation for another time

test/metabase/api/search_test.clj
oisincoveney: It's basically a way for product to determine how much filters are being used, how many people are using the search bar vs the search page, etc

test/metabase/api/search_test.clj
oisincoveney: I _think_ I solved it - calva seemed to figure out how to actually format this. But I'll make sure with you

test/metabase/api/search_test.clj
calherries: Ah okay. Then IMO the code and this test description should be changed to match and document this decision. 

Using a non-nil context to decide whether to send the snowplow event is obviously an implementation detail. The real spec is iff `(contains? #{"global-search" "filtered-search"} `context`)` is true, then send a snowplow event

test/metabase/api/search_test.clj
```diff
@@ -551,42 +551,42 @@
...
-                                 :model_name     (:name model)
-                                 :model_index_id #hawk/malli :int}
-                       "Tromsø" {:pk_ref         (mt/$ids $municipality.id)
-                                 :name           "Tromsø"
-                                 :model_id       (:id model)
-                                 :model_name     (:name model)
-                                 :model_index_id #hawk/malli :int}}
-                      (into {} (comp relevant (map (juxt :name normalize)))
-                            (search! "rom")))))))))))
+                (let [query (mt/mbql-query municipality)]
```
calherries, line 554: `mt/dataset` is a special macro that only indents two spaces. Unfortunately calva doesn't know this so you have to do it by hand.

test/metabase/api/search_test.clj
```diff
@@ -679,38 +679,38 @@
   (testing "You should see Tables in the search results!\n"
     (mt/with-temp [Table _ {:name "RoundTable"}]
       (do-test-users [user [:crowberto :rasta]]
-        (is (= [(default-table-search-row "RoundTable")]
-               (search-request-data user :q "RoundTable"))))))
+                     (is (= [(default-table-search-row "RoundTable")]
```
calherries, line 682: same here. `do-test-users` is a special macro. the original indentation was correct despite calva complaining so

test/metabase/api/search_test.clj
```diff
@@ -812,24 +812,85 @@
...
+                         "context"               "filtered-search"}
+               :user-id (str (mt/user->id :crowberto))}
+              (last (snowplow-test/pop-event-data-and-user-id!))))))
+
+  (snowplow-test/with-fake-snowplow-collector
+    (testing "Send a snowplow event even if the search doesn't have any advanced filters"
+      (mt/user-http-request :crowberto :get 200 "search" :q "test" :context "filtered-search")
+      (is
+       (=?
+        {:data    {"event"                "new_search_query"
```
calherries, line 884: formatting

test/metabase/api/search_test.clj
```diff
@@ -884,40 +945,40 @@
...
-                          (filter #(= (:model %) "card"))
-                          (map :id)
-                          set)
-                     card-id-2)))))
-
-       (testing "error if creator_id is not an integer"
-         (let [resp (mt/user-http-request :crowberto :get 400 "search" :q search-term :created_by "not-a-valid-user-id")]
-           (is (= {:created_by "nullable value must be an integer greater than zero., or sequence of value must be an integer greater than zero."}
-                  (:errors resp)))))))))
+        (testing "Works with archived filter"
```
calherries, line 948: nice, formatting was wrong before

src/metabase/api/search.clj
```diff
@@ -558,12 +559,26 @@
...
-    (when (and (nil? models)
-               (nil? table_db_id)
-               (not archived))
-      (snowplow/track-event! ::snowplow/new-search-query api/*current-user-id* {:runtime-milliseconds duration}))
+        duration   (- (System/currentTimeMillis) start-time)
+        has-advanced-filters (some some?
+                                   [models created_by created_at last_edited_by
+                                    last_edited_at search_native_query verified])]
+    (when (some? context)
+      (when (and (nil? table_db_id) (not archived))
```
calherries, line 567: I actually think you can remove the check on this line, since what we really want to do is track searches from either the search bar or app.

It's sufficient to use 
```
(when (contains? #{"filtered-search" "global-search"} context) ...) 
``` 
as the condition for sending a track-event

oisincoveney: Ah you're right

ranquild (approved):

src/metabase/api/search.clj
calherries: Also, why can't we just call these what they are? i.e. "search-bar" or "search-app". I keep forgetting what they are

oisincoveney: Wait filtered-search should always be filtered

but yeah I actually had them as "search-bar" and "search-app"

src/metabase/api/search.clj
```diff
@@ -558,12 +559,25 @@
...
+     (when has-advanced-filters
+       (snowplow/track-event! ::snowplow/search-results-filtered api/*current-user-id*
+                              {:runtime-milliseconds  duration
+                               :content-type          (u/one-or-many models)
+                                :creator               (some? created_by)
+                                :creation-date         (some? created_at)
+                                :last-editor           (some? last_edited_by)
+                                :last-edit-date        (some? last_edited_at)
+                                :verified-items        (some? verified)
+                                :search-native-queries (some? search_native_query)})))
```
calherries, line 579:
```suggestion
                               :creator               (some? created_by)
                               :creation-date         (some? created_at)
                               :last-editor           (some? last_edited_by)
                               :last-edit-date        (some? last_edited_at)
                               :verified-items        (some? verified)
                               :search-native-queries (some? search_native_query)})))
```

src/metabase/api/search.clj
```diff
@@ -558,12 +559,25 @@
...
-    (when (and (nil? models)
-               (nil? table_db_id)
-               (not archived))
-      (snowplow/track-event! ::snowplow/new-search-query api/*current-user-id* {:runtime-milliseconds duration}))
+        duration   (- (System/currentTimeMillis) start-time)
+        has-advanced-filters (some some?
+                                   [models created_by created_at last_edited_by
+                                    last_edited_at search_native_query verified])]
+    (when (contains? #{"search-app" "search-bar"} context)
+     (snowplow/track-event! ::snowplow/new-search-query api/*current-user-id*
```
calherries, line 567:
```suggestion
      (snowplow/track-event! ::snowplow/new-search-query api/*current-user-id*
```

calherries (approved):

calherries (changes requested):
There are legit test failures
https://github.com/metabase/metabase/actions/runs/6744864009/job/18335590851?pr=35146#step:7:451

calherries (approved):

PR #35164, opened 2023-10-27
[MLv2] Filters — Fix filter picker reuses values after changing a column

### LINKED ISSUE #34960: [MLv2] [Bug] FE - Values from a different filter clause aren't discarded when selecting another column in the picker
Steps:
- Filter on Product -> Vendor
- Add filter
- Click on filter
- Click back and select Category
- The value from the previous clause isn't removed

Screenshot 2023-10-23 at 17 10 01

Expected:
- Treat `filter` in `FilterPicker` as an initial value. When the column is changed, reset it to `undefined`

kulyk:
Fixes #34960

`FilterPicker` is always passing the `filter` object to widget components if it's available (i.e., if we're editing a filter). This resulted in an unexpected behavior when changing a column for an existing filter. Because the `filter` object isn't updated until the "Update filter" button is closed, it contains `values` that makes sense for the original column.

Fixed by _not_ passing the `filter` object to widgets if the column differs from the one in `filterParts`

### To Verify

2. Create a `=` or `!=` filter for the Product's Vendor column (pick a couple of values from the dropdown)
3. Click the filter and change the column to Category
4. Ensure you don't see selected Vendor values in the Category values list

### Demo

https://github.com/metabase/metabase/assets/17258145/141a1d0e-2584-4b85-b8d1-b177785007bc


frontend/src/metabase-lib/comparison.ts
```diff
@@ -14,6 +15,17 @@ export function areLegacyQueriesEqual(
...
+  query: Query,
+  stageIndex: number,
+  column1: ColumnMetadata,
+  column2: ColumnMetadata,
+) {
+  const name1 = displayInfo(query, stageIndex, column1).longDisplayName;
+  const name2 = displayInfo(query, stageIndex, column2).longDisplayName;
+  return name1 === name2;
+}
+
```
kulyk, line 28: This is a placeholder that works good enough for demo purposes, but it should better be a CLJS method

frontend/src/metabase/common/components/FilterPicker/types.ts
```diff
@@ -13,6 +13,7 @@ export interface FilterPickerWidgetProps {
   stageIndex: number;
   column: ColumnMetadata;
   filter?: FilterClause;
+  isNew: boolean;
```
kulyk, line 16: Using `isNew`, so it's consistent with the `DatePicker`

frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
```diff
@@ -121,13 +124,18 @@ export function FilterPicker({
...
+
     return (
       <Box miw={MIN_WIDTH}>
         <FilterWidget
           query={query}
           stageIndex={stageIndex}
           column={column}
-          filter={filter}
+          filter={isSameColumn ? filter : undefined}
+          isNew={!filter}
```
kulyk, line 138: I don't love that picker components can't derive the `isNew` state from `filter` any longer, but this seems like a reasonable compromise?

ranquild: LGTM

frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
```diff
@@ -121,13 +124,18 @@ export function FilterPicker({
   const FilterWidget = getFilterWidget(column);
 
   if (FilterWidget) {
+    const isSameColumn =
```
ranquild, line 127: I wouldn't do that. I'd say that if you choose another column, then drop the existing filter clause in the state. This way we won't need to compare columns

ranquild (commented):
Looks good, but I'd drop the column comparison part. WDYT?

frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
kulyk: @ranquild do you suggest adding something like that (because `filter` is just a prop now)?

```tsx
function FilterPicker({ filter: initialFilter }) {
  const [filter, setFilter] = useState(initialFilter)
}
```

If so, I'm worried it'd make the following scenario awkward:

1. Click a filter like "Category is Gadget"
2. Go back to the list of columns, we reset the `filter` state
3. Open the Category column again
4. Now we need to find the original `filter` and put it back in state

We'd also most likely need a `useEffect` to keep the `filter` prop and state in sync which is rarely great

WDYT?

ranquild: Yes, but after 3 we don't need to find the original `filter`. We need to drop it. We follow this pattern in the `DatePicker` when switching between specific/relative/exclude dates

ranquild: Even for exclude dates, if you go back to the list of temporal buckets and select the same one, your value would be reset

kulyk: Oh, I didn't notice that. Are you sure that's an expected behavior and not a bug in the original relative date picker?

ranquild: Not sure, but it made everything simpler. We could revise this later. If we do revise this, we could do it without `isSameColumn` I think, but we would need to be smarter with our state.

ranquild: @kulyk I just checked - our current FilterPicker doesn't preserve the filter clause either

kulyk: @ranquild I've reworked it, please take another look

frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
```diff
@@ -61,11 +64,25 @@ export function FilterPicker({
     isExpressionEditorInitiallyOpen(query, stageIndex, column, filter),
   );
 
+  const isNewFilter = !initialFilter;
+  const previousFilter = usePrevious(initialFilter);
+
+  useEffect(() => {
```
ranquild, line 70: If I get it correctly, it can be just 
```
useLayoutEffect(() => {
  setFilter(initialFilter);
}, [initialFilter]);
```

ranquild: e.g. when the initialFilter changes it would update the state

ranquild (approved):
Looks good, please take a look at the comment

PR #35166, opened 2023-10-27
Migrate Timeline and TimelineEvent "balloons" icons to "star"

### LINKED ISSUE #35129: Graphing chart with date filter at 24 months is broken
### Describe the bug

We have a new monthly metric model/table with many metrics as columns per month. When charting with 1 metric by month (no aggregation as it's all aggregated), it works fine until a filter is added for the month to be 24 months. 12 months works and no filter works, but not specific months around 24. It also breaks when aggregated too. Most times, it will look like it works, but if you refresh the page, it breaks. 

When it breaks, it shows this error: Cannot read properties of undefined (reading 'source')

### To Reproduce

1. Make line chart for any metric by a date field
2. Filter date field to last 24 months
3. It might show error now, but if not, refresh the page
4. It breaks with error 

### Expected behavior

Filter by any month should work 

### Logs

TypeError: Cannot read properties of undefined (reading 'source')
    at SVGGElement.<anonymous> (timelines.js:155:26)
    at SVGGElement.<anonymous> (d3.js:792:21)
    at d3.js:962:16
    at Ae (d3.js:968:30)
    at K.each (d3.js:961:12)
    at K.html (d3.js:791:36)
    at timelines.js:153:6
    at iL (timelines.js:240:3)
    at LineAreaBarPostRender.js:402:3
    at pH (LineAreaBarPostRender.js:453:3)
value @ CardRenderer.jsx:97
value @ CardRenderer.jsx:44
fs @ react-dom.production.min.js:212
kc @ react-dom.production.min.js:255
t.unstable_runWithPriority @ scheduler.production.min.js:19
Go @ react-dom.production.min.js:122
Bc @ react-dom.production.min.js:248
gc @ react-dom.production.min.js:239
(anonymous) @ react-dom.production.min.js:123
t.unstable_runWithPriority @ scheduler.production.min.js:19
...

Visualization.jsx:309 Visualization.jsx:309 Cannot read properties of undefined (reading 'source')
-- | -- | -- | --
  | value | @ | CardRenderer.jsx:98
  | value | @ | CardRenderer.jsx:44
  | fs | @ | react-dom.production.min.js:212
  | kc | @ | react-dom.production.min.js:255
  | t.unstable_runWithPriority | @ | scheduler.production.min.js:19
  | Go | @ | react-dom.production.min.js:122
  | Bc | @ | react-dom.production.min.js:248
  | gc | @ | react-dom.production.min.js:239
  | (anonymous) | @ | react-dom.production.min.js:123
  | t.unstable_runWithPriority | @ | scheduler.production.min.js:19
  | Go | @ | react-dom.production.min.js:122
  | ei | @ | react-dom.production.min.js:123
  | Zo | @ | react-dom.production.min.js:122
  | mc | @ | react-dom.production.min.js:240
  | notify | @ | Subscription.js:16
  | notifyNestedSubs | @ | Subscription.js:88
  | o | @ | Subscription.js:93

```JSON
master on Postgres
```

### Severity

High as you can't filter for 24 months

### Additional context

_No response_

calherries:
Fixes https://github.com/metabase/metabase/issues/35129

Effectively migrates icon="balloon" to icon="star" which is the default icon. If the default icon changes to something else, so will the icon "balloon" is migrated to.

I've implemented the migration as an "on-demand migration" with toucan's `define-after-select` to avoid complications with instances downgrading versions in the future. For example if an instance upgrades from version 47.0 to 48.0 and then downgrades to 47.0 again, the rollback for a normal liquibase migration would not have been run.

### DIFF
--- a/src/metabase/models/timeline_event.clj
+++ b/src/metabase/models/timeline_event.clj
@@ -23,11 +23,11 @@
 ;;;; schemas
 
 (def default-icon
-  "The default icon for Timeline and TimelineEvents."
+  "The default icon for Timeline an TimelineEvents."
   "star")
 
 (def Icon
-  "TimelineEvent icon string Schema"
+  "Schema for Timeline and TimelineEvents `icon`"
   [:enum default-icon "cake" "mail" "warning" "bell" "cloud"])
 
 (def Source
--- a/src/metabase/models/timeline_event.clj
+++ b/src/metabase/models/timeline_event.clj
@@ -22,10 +22,27 @@
 
 ;;;; schemas
 
-(def Sources
+(def default-icon
+  "The default icon for Timeline and TimelineEvents."
+  "star")
+
+(def Icon
+  "TimelineEvent icon string Schema"
+  [:enum default-icon "cake" "mail" "warning" "bell" "cloud"])
+
+(def Source
   "Timeline Event Source Schema. For Snowplow Events, where the Event is created from is important.
   Events are added from one of three sources: `collections`, `questions` (cards in backend code), or directly with an API call. An API call is indicated by having no source key in the `timeline-event` request."
   [:enum "collections" "question"])
+
+;;;; transforms
+
+(t2/define-after-select :model/TimelineEvent
+  [timeline-event]
+  ;; We used to have a "balloons" icon but we removed it.
+  ;; Use the default icon instead. (metabase#34586, metabase#35129)
+  (update timeline-event :icon (fn [icon]
+                                 (if (= icon "balloons") default-icon icon))))
 
 ;;;; permissions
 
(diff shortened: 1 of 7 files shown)


ranquild (approved):

ranquild:
Approving the PR since it does what was expected

src/metabase/models/timeline_event.clj
```diff
@@ -22,11 +22,28 @@
 
 ;;;; schemas
 
-(def Sources
+(def default-icon
+  "The default icon for Timeline an TimelineEvents."
```
tsmacdonald, line 26:
```suggestion
  "The default icon for Timeline and TimelineEvents."
```

tsmacdonald (approved):
:boat:

Issue #35153, opened 2023-10-27
[MLv2] [Bug] BE - Date filters with time zone offset values don't work


ranquild:
Steps:
- Open product table
- Click on any Created At cell -> Filter by -> Equals
- Go to the notebook editor

Notes:
- The datetime value comes from `/api/dataset` endpoint and is not processed by the FE in any way 

<img width="1157" alt="Screenshot 2023-10-27 at 15 47 50" src="https://github.com/metabase/metabase/assets/8542534/8a2b029d-c531-40d2-aa67-81e470eafb78">
<img width="1728" alt="Screenshot 2023-10-27 at 14 21 36" src="https://github.com/metabase/metabase/assets/8542534/24b9d300-358b-45a8-98b0-71f0181c4fda">

ranquild:
FE fix https://github.com/metabase/metabase/pull/35167

snoe:
BE part was fixed by #35075

<img width="1217" alt="image" src="https://github.com/metabase/metabase/assets/105012/2ea8ac61-89e1-4518-8790-d502b615983f">

PR #35167, opened 2023-10-27
Fix FE not handling dates with timezone offsets in date filters

ranquild:
FE part of https://github.com/metabase/metabase/issues/35153

Steps:
- Open product table
- Click on any Created At cell -> Filter by -> Equals
- Go to the notebook editor
- Click on the filter pill (which is rendered incorrectly atm)
- The filter picker should be initialized correctly
- Filtering by this date should work

Opening the filter picker:
Screenshot 2023-10-27 at 17 51 30

After the clause is updated by FE:
Screenshot 2023-10-27 at 19 01 22

### DIFF
--- a/frontend/src/metabase-lib/filter.ts
+++ b/frontend/src/metabase-lib/filter.ts
@@ -641,6 +641,7 @@ const DATE_FORMAT = "yyyy-MM-DD";
 const TIME_FORMAT = "HH:mm:ss";
 const TIME_FORMAT_MS = "HH:mm:SS.sss";
 const DATE_TIME_FORMAT = `${DATE_FORMAT}T${TIME_FORMAT}`;
+const UTC_OFFSET_REGEX = /(Z|[+-]\d\d:?\d\d)$/;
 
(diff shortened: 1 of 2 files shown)


frontend/src/metabase-lib/filter.ts
```diff
@@ -655,7 +656,9 @@ function serializeDateTime(date: Date): string {
 }
 
 function deserializeDateTime(value: string): Date | null {
-  const dateTime = moment(value, [DATE_TIME_FORMAT, DATE_FORMAT], true);
+  const dateTime = UTC_OFFSET_REGEX.test(value)
```
ranquild, line 659: I'm reusing the same approach and regex we use in `parseTimestamp` but dates without timezone info are parsed in local timezone and not UTC, otherwise mantine components accepting `Date` instances won't work correctly.

ranquild: I believe this should work correctly in all cases since we never include the timezone when serializing the date back

kulyk (approved):

PR #35168, opened 2023-10-27
[MLv2] Make `find-matching-column` correctly match implicit joins

### LINKED ISSUE #35152: [MLv2] - findColumnIndexesFromLegacyRefs can't match Dataset columns to implicit breakouts
Currently in master, if you make a query that that has breakouts from implicitly joined tables, those columns do not show up in Viz settings. This is because `findColumnIndexesFromLegacyRefs` Can't match the column setting to the dataset column. However, explicitly  joining the tables and grouping from columns does work.

image
image

bshepherdson:
A draft for now, since it seems to cause an infinite loop when you try to remove a column.

@npfitz please try this out and see if you can isolate the loop.

Fixes #35152

### DIFF
--- a/src/metabase/lib/equality.cljc
+++ b/src/metabase/lib/equality.cljc
@@ -122,28 +122,37 @@
   [column :- ::lib.schema.metadata/column]
   ((some-fn :metabase.lib.join/join-alias :source-alias) column))
 
+(mu/defn ^:private matching-join? :- :boolean
+  [[_ref-kind {:keys [join-alias source-field]} _ref-id] :- ::lib.schema.ref/ref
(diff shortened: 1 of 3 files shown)


bshepherdson:
* master
  * **PR #35168** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35168" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35168?utm_source=stack-comment).

metamben (approved):
Makes sense.

npfitz (commented):
This was reviewed earlier today with Braden. Appears to solve issue

PR #35173, opened 2023-10-27
[MLv2] Filters — Re-use values when switching filter operators

### LINKED ISSUE #34959: [MLv2] [Bug] FE - Filter breaks when the operator is changed with a value
Enter value
Screenshot 2023-10-23 at 17 10 29

Change operator
Screenshot 2023-10-23 at 17 10 33

What happens:
- `value` gets cleared and the component receives `undefined`, becoming uncontrolled

Expected:
- Default to an empty string when the value is not present
- (Optional) Preserve values when switching between operators, taking the valueCount into account

kulyk:
Fixes #34959

We used to wipe the filter pickers' `values` state when changing an operator. An input component would receive an `undefined` value, become uncontrolled, and enter an unexpected UI state (input shows the previous value, but the "Update filter" button is disabled).

The PR updates filter picker components, so they do their best to re-use `values` when an operator changes. Also, it updates the filter picker's `constants/OPERATOR_OPTION` to be a map instead of an array. I think the map turned out to be a handier format for everything except the `BooleanFilterPicker`.

### To Verify

2. Add a filter for a numeric column, e.g. Subtotal > 10
3. Change the operator to something like "Equals to"
4. Ensure the value 10 is re-used, and you can still update the filter
5. Repeat for a string, time, and coordinate columns with different operators

### Demo

https://github.com/metabase/metabase/assets/17258145/3f948ec9-bf72-4cbb-920e-2d253e135ac8

### DIFF
--- a/frontend/src/metabase/common/components/FilterPicker/BooleanFilterPicker/BooleanFilterPicker.tsx
+++ b/frontend/src/metabase/common/components/FilterPicker/BooleanFilterPicker/BooleanFilterPicker.tsx
@@ -1,5 +1,6 @@
 import { useMemo, useState } from "react";
 import { t } from "ttag";
+import _ from "underscore";
 import { Icon } from "metabase/core/components/Icon";
 import { Button, Divider, Group, Radio, Stack } from "metabase/ui";
 import * as Lib from "metabase-lib";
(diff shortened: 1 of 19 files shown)


ranquild (approved):
Great change, makes things much better! Left a comment about boolean picker options

Issue #13710, opened 2020-11-06
Trend visualization requires time column to be the first otherwise it shows blank visualization


flamber:
**Describe the bug**
Trend visualization requires time column to be the first otherwise it shows blank visualization.

**To Reproduce**
1. Custom question > Sample Dataset > Orders
2. Summarize - do not select metric - group by Quantity:unbinned and CreatedAt:Month
![image](https://user-images.githubusercontent.com/1447303/98380771-478c2380-2049-11eb-96ca-6c1594c3417b.png)
3. Set visualization to Trend - just shows blank visualization
![image](https://user-images.githubusercontent.com/1447303/98380872-72767780-2049-11eb-9ac3-ee8927f90f21.png)
4. Change the summarize grouping, so CreatedAt:Month is the first with Quantity:unbinned as second, then it works:
![image](https://user-images.githubusercontent.com/1447303/98381039-9f2a8f00-2049-11eb-97b0-4c5127d58ed1.png)

**Information about your Metabase Installation:**
Tested 0.35 thru 0.37.0.2

:arrow_down: Please click the :+1: reaction instead of leaving a `+1` or `update?` comment

alxnddr:
@metabase/backend-developers it looks like a back end fix is requires for this bug because the `insights` prop does not have correct data:
<img width="871" alt="Screen Shot 2021-06-16 at 22 27 52" src="https://user-images.githubusercontent.com/14301985/122280657-1a162b00-cef2-11eb-870f-6ef47fb6fa21.png">
`last-change` and `previous-value` are missing

alxnddr:
Closed by https://github.com/metabase/metabase/pull/16638

Created a separate issue for missing BE data: https://github.com/metabase/metabase/issues/16683

PR #35184, opened 2023-10-27
Update Trend chart copy for special cases (on frontend)

### LINKED ISSUE #8919: Smart Number visualization wording
Smart Number visualization displays the previous value as "was ... last day" (or "last week" etc). This is correct when the time filter used includes today / this week, but is incorrect if it doesn't. I think it would be clearer if the label instead stated "was ... day/week before". This way the label would be correct regardless of what the user did with the filters.
A related issue is that when using a native query then the label just says "was ... last". It would be nice if the user could select the label (day/week) in this case. A simpler fix is to just use some generic ending, e.g., "was ... last period".

### LINKED ISSUE #23385: Trend visualization does not show comparison for SQL questions with arbitrary date range
**Describe the bug**
When a SQL question returns 2 dates that are not 7 days part or 1 month apart or 1 year apart, the trend visualization displays "Nothing to compare for the previous .":

image

**To Reproduce**
1. Create a SQL, with a query such as:
   ```
   SELECT CAST(created_at AS date) AS creation_date, SUM(total)
   FROM orders
   WHERE CAST(created_at AS date) IN ('2020-03-01', '2020-03-10')
   GROUP BY creation_date
   ```
3. Select Trend visualization

**Expected behavior**
Even if the elapsed time does not translate well as an English phrase, it should be possible to see the previous value and the percentage difference. For example "↓ 78% • was 992,59 previously"

You can get this information by going to Admin -> Troubleshooting.

- Your databases: reproduced on MySQL and H2
- Metabase version: 0.42.3
- Metabase internal database: MySQL

**Severity**
Our reporting dashboard lets users enter any date range, and there a couple of questions that displays a Trend viz of a metric from beginning of date range to end of date range. Users cannot see the comparison unless they pick 2 dates that are 7 days apart or 1 month apart or 1 year apart.

shaunlebron:
Part of [Trend Chart improvements epic](https://github.com/metabase/metabase/issues/33411)

Will close (after feature branch merged): https://github.com/metabase/metabase/issues/8919
Will close (after feature branch merged): https://github.com/metabase/metabase/issues/23385

## User-facing Changes

https://github.com/metabase/metabase/assets/22608765/89ab247f-8404-40e0-9856-d4c64ccdf831

This PR adds the following updates:

* corrections to our copy for special cases:
  * **previous value is 0**: display infinite percents with `↑ ∞% · vs. previous month: 0`
  * **previous value is the same as current value**: handle 0% change with `No change · vs. previous month`
  * **previous period is a null value**: skip rows with null values when finding the previous value to compare against
    * **no previous values to compare**: handle missing data case with `N/A · vs. previous month: (empty)`
  * truncating `prevDate` based on `nextDate`:
    * if `prevDate` and `nextDate` are on the same day, remove the day, month, and year from the comparison period string
    * if `prevDate` and `nextDate` are in the same year, remove the year from the comparison period string
* We also fix dark mode, by using opacity to achieve same styles in both modes:

before | after
---|---
CleanShot 2023-11-07 at 10 44 11@2x | CleanShot 2023-11-07 at 10 43 40@2x
CleanShot 2023-11-07 at 10 46 59@2x | CleanShot 2023-11-07 at 10 47 29@2x

## Static viz is not updated yet

Static viz will *not* show the aforementioned changes, but there is now a `computeTrend` function used on the frontend which can be used by the backend static viz renderer when wired up correctly in a future PR.

## How to review

SmartScalar had to be re-organized for the requirements of this epic, so I apologize for the large diff— but I hope you will find it more readable than our previous version.

To start, I recommend first clicking the files listed below and perusing them on their own, since they are largely new.  Then reference my diff comments to compare to their original implementations, and check the unit tests.

* [SmartScalar/compute.js](https://github.com/metabase/metabase/blob/c4e600a3015c4d3bc0b3737643476678e7fef400/frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js) is a new file that has everything needed to compute the trend data, and also some formatted display strings to be shared between the UI and static viz (in the `display` properties).
* [SmartScalar/SmartScalar.js](https://github.com/metabase/metabase/blob/c4e600a3015c4d3bc0b3737643476678e7fef400/frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx) is the main component which had a lot of logic moved into either `computeTrend` or `PreviousValueComponent`


iethree:
This is a great list of independent tasks here:

> - fix dark mode
> - display infinite percents correctly
> - handle missing data case with special message
> - handle 0% change with special message
> - when selecting previous value to compare, find the most recent row with a non-null value (instead of just the previous row)
> - prepare data for rendering to static viz

Would it maybe make sense to open each of these as a separate little PR? that way we could get some quick easy wins into master one at a time? Would also make it much easier to review each one. 🙏

shaunlebron:
@iethree I would if I could.  It’s a bigger refactor, so the individual changes are 2-3 lines each on top of the refactoring.

e2e/test/scenarios/visualizations/smartscalar-trend.cy.spec.js
```diff
@@ -29,7 +29,5 @@ describe("scenarios > visualizations > scalar", () => {
     cy.log("Bug: showing blank visualization");
 
     cy.get(".ScalarValue").contains("100");
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("Nothing to compare for the previous month.");
```
shaunlebron, line 33: I removed this check because it’s only necessary to confirm that this visualization displays a value, not what the comparison is.  See #13710.

_Extra context— Our new frontend `computeTrend` function skips over null gaps until finding a valid month, and changes this message.  I opted not to correct this message because the rows are not sorted by date here, which causes the backend trend calculation to trip up since the previous row is not exactly one month adjacent to the source row.  The frontend calculation now displays the previous row regardless of what it is._

frontend/src/metabase/css/dashboard.css
```diff
@@ -263,24 +263,6 @@
...
-  }
-}
-
-@media screen and (min-width: 1540px) {
-  .AutomaticDashboard.AutomaticDashboard--withSidebar
-    #SmartScalar-PreviousValue {
-    display: inherit;
-  }
-}
-
```
shaunlebron, line 283: I removed this because we have a new system for responsively toggling this value when it doesn’t fit in the card.

_Extra context— the id `#SmartScalar-PreviousValue` is only used here as a hack to hide long numbers at the end of `↑ 50% · vs. previous month: ______` on automated dashboards (via x-ray) when the page isn’t wide enough, but this is a hack because it doesn’t measure the card width, which our new system does instead._

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -73,42 +63,3 @@ export const getValueHeight = (height: number): number => {
...
-      weight: 700,
-    }).width;
-    return availablePreviousValueWidth >= previousValueWidth;
-  });
-
-  return {
-    isPreviousValueTruncated: matchIdx !== 0,
-    fittedPreviousValue: previousValueCandidates[matchIdx],
-  };
-};
```
shaunlebron, line 114: Everything this function was doing is now inlined in a shorter way here:
https://github.com/metabase/metabase/pull/35184/files#diff-b0fbfd8b0eb7fcb57feb5b0b2df90ad1925947bac9fc8aefab49fbd15dfbc76aR106-R128

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -21,36 +20,27 @@ export const formatChangeAutoPrecision = (
...
+  [2, 1]
+    .map(n => formatChange(change, { maximumFractionDigits: n }))
+    .find(
+      formatted =>
+        measureText(formatted, {
+          size: "1rem",
+          family: fontFamily,
+          weight: fontWeight,
+        }).width <= width,
+    ) ?? formatChange(change, { maximumFractionDigits: 0 });
```
shaunlebron, line 33: a functional refactoring for the next guy who has to look at it.  [existing unit tests](https://github.com/metabase/metabase/blob/67a1d7847390308e52d0f1cc83fbd66eeac8724b/frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.unit.spec.ts#L49-L74) verify that it still works.

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -0,0 +1,140 @@
...
+      {
+        description: "should correctly display percent decrease over null data",
+        rows: [
+          ["2019-09-01", 300],
+          ["2019-10-01", null],
+          ["2019-11-01", 100],
+        ],
+        expected: "100; Nov 2019; ↓ 66.67% vs. Sep 2019: 300",
+      },
+    ])("$description", ({ rows, expected }) => {
```
shaunlebron, line 135: Start here to understand how our new `computeTrend` function returns everything needed to display our trend chart, both on the UI and the static viz.

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+  // a   b     b   a
+  // - → - =  [+ → +]
+  // - → + = -[+ → -]
+  return b < 0 ? computeChange(-b, -a) : -computeChange(b, a);
+}
+
+export const PREVIOUS_VALUE_MISSING = "PREVIOUS_VALUE_MISSING";
+export const PREVIOUS_VALUE_SAME = "PREVIOUS_VALUE_SAME";
+export const PREVIOUS_VALUE_CHANGED = "PREVIOUS_VALUE_CHANGED";
```
shaunlebron, line 40: Enumerating of the different cases that have to be displayed differently.

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -0,0 +1,156 @@
...
+        description:
+          "should correctly fallback to day unit if backend doesn’t return valid unit",
+        rows: [
+          ["2019-09-01", 100],
+          ["2019-11-01", 300],
+        ],
+        dateUnit: null,
+        expected: "300; Nov 1, 2019; ↑ 200% vs. Sep 1, 2019: 100",
+      },
+    ])("$description", ({ rows, expected, dateUnit }) => {
```
shaunlebron, line 147: Start here to understand how our new `computeTrend` function returns everything needed to display our trend chart, both on the UI and the static viz.

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -0,0 +1,185 @@
...
+import { formatValue } from "metabase/lib/formatting/value";
+import { formatDateTimeRangeWithUnit } from "metabase/lib/formatting/date";
+import { color } from "metabase/lib/colors";
+import { formatChange } from "metabase/visualizations/visualizations/SmartScalar/utils";
+import { isDate } from "metabase-lib/types/utils/isa";
+
+const FALLBACK_DATE_UNIT = "day";
+
+// compute the percent change between two values (a → b)
+export function computeChange(a, b) {
```
shaunlebron, line 14: ported from clojure: https://github.com/metabase/metabase/blob/d2ecd17/src/metabase/sync/analyze/fingerprint/insights.clj#L32-L42

For readability, I split up the cases into comments, which are validated in [unit tests below](https://github.com/metabase/metabase/pull/35184/files#diff-5319611e9f4c7a99acd8792fbeb80a7961e52432a545beaad54ab259db37c4b2R8-R59).

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.unit.spec.js
```diff
@@ -35,20 +35,8 @@ describe("SmartScalar", () => {
...
-        unit: "month",
-        col: "Count",
-      },
-      {
-        "last-value": 220,
-        "last-change": 0.1,
-        "previous-value": 200,
-        unit: "month",
-        col: "Sum",
-      },
```
shaunlebron, line 51: These insight values are now computed in the frontend automatically, so we remove them in these unit tests to ensure we’re using those instead.

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -71,125 +66,131 @@ const ScalarPeriod = ({ lines = 2, period, onClick }) => (
...
-    </PreviousValueLabel>
-  );
-  const previousValueDisplayCompact = (
-    <PreviousValueLabel>
-      {formatValue(previousValue, {
-        ...formatOptions,
-        compact: true,
-      })}
-    </PreviousValueLabel>
-  );
```
shaunlebron, line 162: I did my best to unify a lot of our elements between the card and tooltip to deduplicate this kind of work.  All work related to comparison display is cleaned up and moved  to `PreviousValueComparison`.

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.styled.tsx
```diff
@@ -2,11 +2,18 @@ import styled from "@emotion/styled";
...
+// NIGHT-MODE TEXT HACK:
+// Our "fullscreen-night-text" className only supports one shade of text (white).
+// As a temporary fix, we create dark-mode analogs for "text-medium" and "text-light" using opacity:
+const detailsOpacity = 0.85; // approximates color("text-medium")
+const detailsNumberOpacity = 0.7; // approximates color("text-light")
+
 export const Variation = styled.div`
   color: ${props => props.color};
+  opacity: ${props =>
+    props.color == null ? detailsNumberOpacity * detailsOpacity : 1};
```
JesseSDevaney, line 16:
```javascript
import { isEmpty } from "metabase/lib/validate";
```

```suggestion
    isEmpty(props.color) ? detailsNumberOpacity * detailsOpacity : 1};
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.styled.tsx
```diff
@@ -2,11 +2,18 @@ import styled from "@emotion/styled";
...
 import { Icon } from "metabase/core/components/Icon";
-import { color } from "metabase/lib/colors";
 import { space } from "metabase/styled-components/theme";
 
+// NIGHT-MODE TEXT HACK:
+// Our "fullscreen-night-text" className only supports one shade of text (white).
+// As a temporary fix, we create dark-mode analogs for "text-medium" and "text-light" using opacity:
+const detailsOpacity = 0.85; // approximates color("text-medium")
+const detailsNumberOpacity = 0.7; // approximates color("text-light")
+
```
JesseSDevaney, line 12: I do not quite understand why we are using opacity instead of directly setting the color here. Can you explain it to me?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -1,62 +1,57 @@
 /* eslint-disable react/prop-types */
```
JesseSDevaney, line 1: Can we convert this into TypeScript or is there a reason this file was not converted into TypeScript?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -71,125 +66,131 @@ const ScalarPeriod = ({ lines = 2, period, onClick }) => (
...
-  ],
+  series,
   rawSeries,
   gridSize,
   width,
   height,
   totalNumGridCols,
   fontFamily,
 }) {
+  const fontSize = "0.875rem";
```
JesseSDevaney, line 180: Why are we hard-coding `fontSize` in this component?

JesseSDevaney (changes requested):
Hey Shaun - thanks for working on this and getting this 99% of the way there! :bowing_man:  

There are some things I think should be addressed before merging. Most of my comments are nits, some are about correctness, and some are about educating me. 

I am hoping we can find a time to pair on this Monday to go over the changes and work-out what should be changed and what does not need to be before we merge. Until then, I hope you have a lovely weekend :palm_tree:

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -0,0 +1,186 @@
...
+    moment
+      .utc(date)
+      .startOf(dateUnit)
+      .add(1, dateUnit)
+      .isSame(moment.utc(nextDate).startOf(dateUnit));
+
+  const title =
+    date == null || datesAreContinuous
+      ? t`previous ${dateUnitDisplay}`
+      : formatDateTimeRangeWithUnit([date], dateUnit, { compact: true }); // FIXME: elide part of the prevDate in common with lastDate
```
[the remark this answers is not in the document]
shaunlebron: The figma design shows that the date here should omit the year:
CleanShot 2023-11-20 at 10 15 05@2x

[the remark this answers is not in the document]
JesseSDevaney: See commit [7ffa506](https://github.com/metabase/metabase/pull/35184/commits/7ffa506e28ae57f88a9eb11a831c26ba0f04c275)

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -125,11 +125,11 @@ describe("SmartScalar > compute", () => {
         description:
           "should correctly display percent decrease over missing row",
```
shaunlebron, line 126: The description for the test case doesn’t make sense anymore after your change below.

This “missing row” here is intended to be October.  If the “latest” data is for November, and the most recent data is for September, then the comparison would read “Sep” instead of “previous month”.  Does that make sense?

JesseSDevaney: Yes, thanks for catching that!

cdeweyx (approved):
Works as expected 👍

frontend/src/metabase/lib/formatting/date.tsx
```diff
@@ -634,6 +630,29 @@ export function formatDateTimeRangeWithUnit(
     return String(start);
   }
 
+  const formatDate = (date: Moment, formatStr: string) => {
+    const yearRegex = /,*\s*YYYY/;
+    const dayRegex = /[\s\S]+,*\s*YYYY,*\s*/;
+
+    // day display is hide-able (if date being compared to is same day)
+    if (options.removeDay) {
```
alxnddr, line 638: The new `removeDay` and  `removeYear` options assume `formatStr` date format and probably won't work on the other date formats. 

I would assume that instead of replacing formatStr we need to compose it right away without a day or a year before it gets to this function

alxnddr: It would be also nice to add more unit test son `date.tsx` specifically to verify more combinations on different formats since it is our common formatting code and it should have good coverage

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -71,16 +66,118 @@ const ScalarPeriod = ({ lines = 2, period, onClick }) => (
...
+    comparisonType,
+    percentChange,
+    comparisonPeriodStr,
+    prevValue,
+    changeArrow,
+    changeColor,
+    display,
+  } = comparison;
+
+  const arrowIconName = { "↓": "arrow_down", "↑": "arrow_up" }[changeArrow];
```
alxnddr, line 87: I'd avoid using symbols like `↓` that are not convenient to type and instead just used `arrow_down`. Seems like these symbols themselves are not used anywhere in UI

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.styled.tsx
```diff
@@ -2,11 +2,19 @@ import styled from "@emotion/styled";
 
 import { Ellipsified } from "metabase/core/components/Ellipsified";
 import { Icon } from "metabase/core/components/Icon";
-import { color } from "metabase/lib/colors";
 import { space } from "metabase/styled-components/theme";
+import { isEmpty } from "metabase/lib/validate";
+
+// NIGHT-MODE TEXT HACK:
+// Our "fullscreen-night-text" className only supports one shade of text (white).
```
alxnddr, line 9: Reimplementing night mode in a more convenient way for our styles is definitely out of the scope of this project.

Can we still avoid such hacks and use nested styles?
```
  .fullscreen-night-text & {
    color: ${...};
  }
```

alxnddr (commented):
Great work! I'm happy to see so many added specs on SmartScalar

I left a few comments, can you please have a look?

alxnddr (approved):

PR #35185, opened 2023-10-27
Make Pre-Release checks trigger on workflow_dispatch

iethree:
### Description

Previously, pre-release tests triggered on a `workflow_run` trigger, this makes it difficult to pass data between jobs.  This moves the pre-release job to a `workflow_dispatch` trigger which has two benefits:

1) we can easily pass arbitrary data between jobs (this will be important for [slack notifications](https://github.com/metabase/metabase/pull/35186))
2) we can manually trigger pre-release tests on any already-built jar

For now, the behavior is exactly the same: we still automatically trigger pre-release tests, we just do so with an explicit step in the build-for-release action.

## Test Run: 
- [Build](https://github.com/iethree/metabase/actions/runs/6672830038)
- [Test](https://github.com/iethree/metabase/actions/runs/6672952042)

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,5 +1,5 @@
 name: Release 2 - Pre-release testing
-run-name: Publish Release ${{ inputs.version }}
+run-name: Pre-release testing of ${{ inputs.version }}
 
 on:
   workflow_dispatch:
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,14 +1,19 @@
 name: Release 2 - Pre-release testing
-run-name: Pre-release Testing of ${{ github.event.workflow_run.name }}
+run-name: Publish Release ${{ inputs.version }}
 
 on:
-  workflow_run:
-    workflows: [Release 1 - Build Release Artifact]
-    types: [completed]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Metabase version (e.g. v0.46.3)'
+        type: string
+        required: true
+      commit:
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-release.yml
```diff
@@ -1,14 +1,19 @@
 name: Release 2 - Pre-release testing
-run-name: Pre-release Testing of ${{ github.event.workflow_run.name }}
+run-name: Publish Release ${{ inputs.version }}
```
nemanjaglumac, line 2: I'm confused by this. We're not publishing a release, aren't we?
This workflow is supposed to download the official artifact, test it and then push it to our staging Dockerhub repo.

.github/workflows/pre-release.yml
```diff
@@ -1,14 +1,19 @@
...
+      version:
+        description: 'Metabase version (e.g. v0.46.3)'
+        type: string
+        required: true
+      commit:
+        description: 'A full-length commit SHA-1 hash'
+        required: true
 
 jobs:
   release-artifact:
```
nemanjaglumac, line 16: This was me shooting myself in the foot.
It took me good five minutes to figure out why is this "A release artifact" and not "Release the artifact". We should change the name to download release artifact or obtain the release artifact.

IIRC, you even suggested something like this a few months ago.
Sorry for not implementing it back then.

.github/workflows/pre-release.yml
iethree: good catch, this was a copy-pasta error when I changed the variable

nemanjaglumac (approved):

PR #35189, opened 2023-10-28
Convert SmartScalar to function component

shaunlebron:
Needed for https://github.com/metabase/metabase/issues/33411

Turn off whitespace diff.


iethree (approved):
🏆

PR #35191, opened 2023-10-28
Use built-in `tools.build` manifest functionality

nemanjaglumac:
According to this answer, it was added back in `v0.14`.
https://ask.clojure.org/index.php/10827/ability-customize-manifest-created-clojure-tools-build-uber?show=10840#a10840

### MANIFEST.MF

**Before**
```
Manifest-Version: 1.0
Multi-Release: true
Created-By: Metabase build.clj
Build-Jdk-Spec: 11
Main-Class: metabase.bootstrap
```

**After**
```
Manifest-Version: 1.0
Created-By: depstar
Built-By: metabase
Build-Jdk: 11.0.20.1
Multi-Release: true
Main-Class: clojure.main
Manifest-Version: 1.0
Multi-Release: true
Created-By: Metabase build.clj
Build-Jdk-Spec: 11
Main-Class: metabase.bootstrap
```

### DIFF
--- a/bin/build/src/build/uberjar.clj
+++ b/bin/build/src/build/uberjar.clj
@@ -99,15 +99,6 @@
         (u/step (format "Copy %s" path)
                 (b/copy-dir {:target-dir class-dir, :src-dirs [path]}))))))
 
-(defn create-uberjar! [basis]
-  (u/step "Create uberjar"
-    (with-duration-ms [duration-ms]
-      (depstar/uber {:class-dir class-dir
-                     :uber-file uberjar-filename
-                     :basis     basis
-                     :exclude   [".*metabase.*.clj[c|s]?$"]})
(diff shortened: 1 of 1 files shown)


camsaul:
Having two `Main-Class` entries does not seem like a good idea, any way to prevent it from including the first (wrong) one?

metamben (changes requested):
Running the jar file with such a manifest would always print a warning and (most likely) run the first declared `Main-Class`, so it's not good.

camsaul (changes requested):
Seconding Tamas' request for changes, if you figure out how to get this to output just one Main-Class then please request a rereview

nemanjaglumac:
> Seconding Tamas' request for changes, if you figure out how to get this to output just one Main-Class then please request a rereview

Will do.
Thank you both.

This completely got lost on my radar. I'll try to get back to it.

PR #35203, opened 2023-10-30
Attempt to fix the flaky `hydrate-is-personal-test` test

qnkhuat:
failed CI https://github.com/metabase/metabase/actions/runs/6691021110/job/18177457410

I can't reproduce this locally but my hypothesis is `select id in [...]` sometimes does not return things in the exact order of the id list?


sansolovyov (approved):
LGTM, Postgres actually does not promise to use `IN` in sorting at all.

PR #35204, opened 2023-10-30
Remove innappropriate type casts and type guards for FieldFilterUiParameter

calherries:
This PR removes a couple of innappropriate uses of type casts and guards.

It was made based on [this discussion](https://github.com/metabase/metabase/pull/35086#issuecomment-1784880354). We shoudn't be using narrower types than we can enforce using the type checker.

### DIFF
--- a/frontend/src/metabase-lib/parameters/utils/parameter-type.ts
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-type.ts
@@ -1,6 +1,5 @@
 import _ from "underscore";
 import type { Parameter } from "metabase-types/api";
-import type { FieldFilterUiParameter } from "metabase-lib/parameters/types";
 import { FIELD_FILTER_PARAMETER_TYPES } from "metabase-lib/parameters/constants";
 
 export function getParameterType(parameter: Parameter | string) {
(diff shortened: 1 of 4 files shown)


frontend/src/metabase-lib/parameters/utils/parameter-type.ts
```diff
@@ -42,9 +41,7 @@ export function isStringParameter(parameter: Parameter) {
   return type === "string";
 }
 
-export function isFieldFilterParameter(
-  parameter: Parameter,
-): parameter is FieldFilterUiParameter {
```
calherries, line 47: This type guard is a lie in isolation. The parameter is really not a `FieldFilterUiParameter` unless it has been transformed by the `buildFieldFilterUiParameter` function. https://github.com/metabase/metabase/blob/c37e7ac9d1213e11fcf0b7639fd4006b84a6abd6/frontend/src/metabase/parameters/utils/dashboards.ts#L132

frontend/src/metabase/parameters/components/ValuesSourceModal/ValuesSourceTypeModal.tsx
```diff
@@ -98,7 +95,7 @@ const ValuesSourceTypeModal = ({
       {sourceType === null ? (
         <FieldSourceModal
           // if sourceType === null the parameter must have fields
-          parameter={parameter as FieldFilterUiParameter}
```
calherries, line 101: There's no way to prove parameter has fields with the type checker so I've removed it

kamilmielnik (approved):

PR #35207, opened 2023-10-30
Refactor: Only use type predicates for type guards

calherries:
This PR addresses Kamil's suggestion here https://github.com/metabase/metabase/pull/34958#discussion_r1376065372 to only use type predicates with type guards, and not other functions.


frontend/src/metabase-lib/parameters/utils/parameter-fields.ts
```diff
@@ -4,18 +4,18 @@ import type {
...
-export const hasFields = (
+export const isFieldFilterUiParameter = (
   parameter: UiParameter,
 ): parameter is FieldFilterUiParameter => {
-  return (
-    "fields" in parameter &&
-    Array.isArray(parameter.fields) &&
-    parameter.fields.length > 0
-  );
+  return "fields" in parameter && parameter.fields != null;
```
kamilmielnik, line 10: That `Array.isArray` check was stricter:

```suggestion
  return "fields" in parameter && Array.isArray(parameter.fields);
```

kamilmielnik (approved):
:+1:

frontend/src/metabase-lib/parameters/utils/parameter-fields.ts
```diff
@@ -4,18 +4,18 @@ import type {
...
 ): parameter is FieldFilterUiParameter => {
-  return (
-    "fields" in parameter &&
-    Array.isArray(parameter.fields) &&
-    parameter.fields.length > 0
-  );
+  return "fields" in parameter && Array.isArray(parameter.fields);
+};
+
+export const hasFields = (parameter: UiParameter): boolean => {
```
uladzimirdev, line 13:
```suggestion
export const hasFields = (parameter: UiParameter) => {
```

it's clear it's boolean

uladzimirdev (approved):

PR #35208, opened 2023-10-30
Fix click actions perf issue with charts

### LINKED ISSUE #35206: Switching between dashboard tabs takes more than 10 seconds
### Describe the bug

When switching between dashboards tabs, sometimes it can take more than 10 sec.

### To Reproduce

1. Go to https://stats.metabase.com/dashboard/1880
2. Wait for the first tab to fully load
3. Switch between tabs
4. After maybe 5 times the UI will get stuck for ~10 sec

Screenshot 2023-10-30 at 14 11 47

### Expected behavior

Much faster tab switches

### Logs

_No response_

```JSON
Chrome 118
```

### Severity

P1 because it takes 10 sec

### Additional context

The issue is with `visualizationIsClickable` calls as can be seen in the performance trace. `visualizationIsClickable` calls MLv2 under the hood and it takes a lot of time to finish. We can fix most issues on the FE by calling it in even handlers and not when rendering.

ranquild:
Fixes https://github.com/metabase/metabase/issues/35206

The performance issue exists because we call `availableDrillThrus` via `visualizationIsClickable` in most visualizations during rendering, and this function can take some time to execute. This PR fixes most `availableDrillThrus` calls by moving them to event handlers. The only place that is not fixed is rendering table header cells, but there the fix is not trivial and the results are cached, unlike the places fixed in this PR.

How to verify:
- Create a dashboard with charts of all types affected in this PR
- Check that `availableDrillThrus` is not called during rendering of the dashboard

Screenshot 2023-10-30 at 15 21 53


npfitz (approved):
I'd feel better if we had drill thru tests for all of these viz types, but everything looks okay.

PR #35210, opened 2023-10-30
Add type safety to jest config

uladzimirdev:
### Description

Add type safety to jest config

image

### How to verify

jest tests should pass

### Checklist


jest.config.js
iethree, line 1: can this work as `jest.config.ts` ? weird to do type checks on .js

the docs seem to imply you can: https://jestjs.io/docs/configuration

npretto: It seems it requires us to install `ts-node` as devDep though:
> To read TypeScript configuration files Jest requires [ts-node](https://npmjs.com/package/ts-node). Make sure it is installed in your project.

uladzimirdev: With .ts extension you need to transpile config every time, so using ts-check on js files is kinda a workaround without downsides of using ts extension

npretto (approved):
LGTM, should we backport it so that we won't have conflicts if we change the config in PRs related to bugfixes?

iethree (approved):

PR #35218, opened 2023-10-30
Add down migration and H2 dump loading test

### LINKED ISSUE #34857: Write the necessary tests to guarantee that `load-from-h2` and `migrate down` always work and make them green
**Context**
These two commands are necessary for general Metabase operations. They enable moving from H2 to Postgres/MySQL, migrate to Cloud, and downgrade.

- [ ] Test `migrate down` is always working. It should be always possible to `migrate down` from the current major.
- [ ] Test `load-from-h2` is working in the current version and across different versions. We should test at least 4 versions. So if we are on 47. `load-from-h2` should work from a 44, 45, 46, and 47 H2 file.

There are existing issues with both commands, so these tests will fail at first. We need to fix these issues too.

metamben:
Related to, but not enough to fix #34857.

Will have to be ported to `master`.


qnkhuat:
@metamben is this ready for reviews?

metamben:
> @metamben is this ready for reviews?
@qnkhuat not yet. I realize it's trying to do things it can't, so I'll simplify it.

test/metabase/cmd/load_from_h2_test.clj
```diff
@@ -48,3 +58,93 @@
...
+(defn- create-current-database
+  [db-type db-def data-source]
+  (tx/create-db! db-type db-def)
+  (mdb.setup/setup-db! db-type data-source true))
+
+(defn- dump-filename
+  [h2-filename version]
+  (str h2-filename "-" version))
+
+(defn- parse-major-version
```
qnkhuat, line 76:
```suggestion
(defn- major-version
```

this doesn't take any argument, and prefix with `parse` makes it feel like it should process something.

I suggest removing the prefix

qnkhuat: also wdyt about naming this `liquibase-latest-major-version` to make it explicit where it comes from?

qnkhuat (approved):
have some nits but this lgtm

Issue #35209, opened 2023-10-30
[Epic] Make dashboard saves faster


calherries:
Resolves [Update dashboard performance 48](https://github.com/metabase/metabase/issues/34721#top)

Saving a dashboard takes too long on master, due to https://github.com/metabase/metabase/pull/33392

```[tasklist]
### Problems
- [ ] https://github.com/metabase/metabase/issues/34828
- [x] Two revisions can be created when a dashboard is saved
- [x] FE calls separate endpoints for updating a dashboard vs its dashcards and tabs
```

```[tasklist]
### PRs
- [ ] https://github.com/metabase/metabase/pull/34947
- [ ] https://github.com/metabase/metabase/pull/35223
- [ ] https://github.com/metabase/metabase/pull/35406
- [ ] https://github.com/metabase/metabase/pull/35413
- [ ] https://github.com/metabase/metabase/pull/35499
```

calherries:
The problem is resolved on the BE. The remaining problem is still on the FE, with a redundant GET request. See [this comment](https://github.com/metabase/metabase/issues/34828#issuecomment-1802203038).

PR #35223, opened 2023-10-30
BE prep for using PUT /api/dashboard/:id to update dashcards and tabs

calherries:

We currently have to call two endpoints to update a dashboard with its nested entities. There should only be one.

**Before, to update a dashboard and its dashcards/tabs, you need to:** 
- Make a request to `PUT /api/dashboard/:id` to update the dashboard, excluding dashcards and tabs
- Then make a request to `PUT /api/dashboard/:id/cards` to update the dashcards and tabs

**With this PR, to update a dashboard and its dashcards/tabs, you need to:** 

1. (exact same steps as before)

OR

2. Make a single request to `PUT /api/dashboard/:id` to update the dashboard, _including_ dashcards and tabs (this method is currently untested and unused, but it works with manual testing)

In this PR, all code and tests still follow method 1. In a following PR, I will update the code and tests to follow method 2.

### BE changes
This PR updates the `PUT /api/dashboard/:id` endpoint to also save dashcards and tabs if they are included in the request body.

I have kept the `PUT /api/dashboard/:id/cards` endpoint working as before, but almost all the work is done by the `update-dashboard` function, which is reused in the body of the `PUT /api/dashboard/:id` handler.

It's obvious the events that are generated by updating the dashboard and dashcards/tabs need to be de-duplicated, because multiple `push-revision!` calls are currently being made for a single request. But for now I'd like to preserve the current behaviour to keep this PR small, and I will fix this in a separate PR.

### FE changes
There is one breaking change to the API: you can't send dashcards and tabs in the body of PUT /api/dashboard/:id and then expect to still be able to create them later with PUT /api/dashboard/:id/cards. Before this was possible because the `PUT /api/dashboard/:id` handler would simply ignore these keys, but now it creates them if they are in the body and have negative IDs. With the old FE code the BE would have tried to create new dashcards and tabs twice, once for `PUT /api/dashboard/:id` and again for `PUT /api/dashboard/:id/cards`.

### Follow up work: 
PR 2:
  - update FE to just call `/api/dashboard/:id` when saving a dashboard with dashcards and tabs
  - Replace uses of `/api/dashboard/:id/cards` endpoint with `/api/dashboard/:id` in FE and BE tests
  - Mark `/api/dashboard/:id/cards` as deprecated

PR 3:
  - De-duplicate BE events that do the same thing, i.e. multiple calls to `push-revision!`


calherries:
@metabase/core-backend-42 This is ready for BE review. I'm still working on the FE tests because there are circular dependencies that need unravelling.

src/metabase/api/dashboard.clj
```diff
@@ -607,6 +564,92 @@
...
+                     (merge
+                      (select-keys tabs-changes-stats [:created-tab-ids :updated-tab-ids :deleted-tab-ids :total-num-tabs])
+                      (select-keys dashcards-changes-stats [:created-dashcards :deleted-dashcards :updated-dashcards]))))))
+       true))
+    (let [dashboard (t2/select-one :model/Dashboard id)]
+      (when update-dashboard-itself?
+        ;; execute these events for old PUT /api/dashboard/:id calls and new PUT /api/dashboard/:id calls, and not for old PUT/api/dashboard/:id/card calls
+        (events/publish-event! :event/dashboard-update {:object dashboard :user-id api/*current-user-id*}))
+      (when update-dashcards-and-tabs?
+        ;; execute these events for old PUT /api/dashboard/:id/cards calls and new PUT /api/dashboard/:id calls, and not for old PUT /api/dashboard/:id calls
```
calherries, line 624: Here I'm being careful not to make any breaking changes. 

That means:
1. If the FE makes a request to `PUT /api/dashboard/:id` with no dashcards or tabs in the body, followed by `PUT /api/dashboard/:id/cards` with dashcards and tabs in the body, the same set of events should be fired before and after this PR.
AND
2. If the FE makes a single request to `PUT /api/dashboard/:id` with dashcards or tabs in the body, the same set of events should be fired as in 1.

To understand these conditions are met, you need to know:
```
(track-dashcard-and-tab-events! ...) used to be in the PUT /api/dashboard/:id/cards handler.
(events/publish-event! ...) used to be in the PUT /api/dashboard/:id handler.
```
and then read the code

src/metabase/api/dashboard.clj
```diff
@@ -607,6 +564,92 @@
     (events/publish-event! :event/dashboard-update-tabs
                            {:object dashboard :user-id api/*current-user-id* :tab-ids updated-tab-ids})))
 
+(defn- update-dashboard
```
calherries, line 567: This effectively the merged body of the two endpoints.

uladzimirdev:
> The FE now doesn't use PUT /api/dashboard/:id/cards except in tests

@calherries why do we keep it in tests then?

calherries:
@uladzimirdev Good question, I answered that in the "Why keep PUT /api/dashboard/:id/cards around at all?" section:

> Three reasons:

But in writing this answer I realised that I can achieve step 2 (proving they are equivalent) by simply making the FE changes in a separate PR to the BE changes. So I will do so now. And I will replace all usages of `/cards` in e2e tests.

frontend/src/metabase/entities/collections/collections.ts
```diff
@@ -3,7 +3,7 @@ import { createSelector } from "@reduxjs/toolkit";
 
 import { GET } from "metabase/lib/api";
 import { createEntity, undo } from "metabase/lib/entities";
-import * as Urls from "metabase/lib/urls";
+import * as Urls from "metabase/lib/urls/collections";
```
piranha, line 6: Not sure I understand why is that change here. :)

sansolovyov (approved):
LGTM, especially given that it's mostly code reorganizing and tests are in place to test for the old behavior. The only thing that I would love to see is a unit test for the new behavior I guess?

My hands are itching to unify some logic in `update-dashboard` (like moving `publish-event!` into `track-dashcard-and-tab-events!`), but I guess that's reserved for the next PR. :-)

frontend/src/metabase/entities/collections/collections.ts
calherries: I introduced some circular deps. But I will do that in a second PR

frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -121,10 +122,15 @@ export const updateDashboardAndCards = createThunkAction(
         dispatch(saveCardsAndTabs(updatedCardsAndTabs));
       });
 
-      // Make two parallel requests: one to update the dashboard and another for the dashcards and tabs
+      // Make two parallel requests: one to update the dashboard (without dashcards and tabs)
+      // and another for the dashcards and tabs
+      const onlyDashboard = produce(dashboard, d => {
+        delete d.dashcards;
+        delete d.tabs;
+      });
```
calherries, line 130: The reason I had to do this is because the FE was sending dashcards and tabs with `PUT /api/dashboard/:id` when it shouldn't have. It didn't break anything before because the implementation of the API actually didn't care that there were extra keys in the body. But now it does expect that those keys are not there for this to work.

frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -121,10 +122,15 @@ export const updateDashboardAndCards = createThunkAction(
         dispatch(saveCardsAndTabs(updatedCardsAndTabs));
       });
 
-      // Make two parallel requests: one to update the dashboard and another for the dashcards and tabs
+      // Make two parallel requests: one to update the dashboard (without dashcards and tabs)
+      // and another for the dashcards and tabs
+      const onlyDashboard = produce(dashboard, d => {
```
uladzimirdev, line 127: I'd rename it to something like `dashboardWithoutTabsAndDashcards`

uladzimirdev (approved):

frontend/src/metabase/dashboard/actions/save.js
calherries: Yeah, I agree. Thankfully I'm going to be removing this code in the next PR so I will clean it up then to avoid rerunning CI if I don't have to

calherries (commented):
@piranha I had to make [a change](https://github.com/metabase/metabase/pull/35223/files/fa28aa12f093e7e694c261083475e8f661d8ba8b..f9f408c0f1fe05dbb66fea46d7f4a40756562698#) because I merged the two endpoints incorrectly. 

The old `PUT /api/dashboard/:id/cards` endpoint called `api/check-not-archived` but the old `PUT /api/dashboard/:id` endpoint didn't. Before these commits I just added, I was calling it in both instances. I've fixed it now so it's only called if `dashcards` and `tabs` are provided in the request body, mirroring the old behaviour

sansolovyov (approved):

src/metabase/api/dashboard.clj
```diff
@@ -607,6 +564,93 @@
...
+        (events/publish-event! :event/dashboard-update {:object dashboard :user-id api/*current-user-id*}))
+      (when update-dashcards-and-tabs?
+        ;; execute these events for old PUT /api/dashboard/:id/cards calls and new PUT /api/dashboard/:id calls, and not for old PUT /api/dashboard/:id calls
+        (track-dashcard-and-tab-events! dashboard @changes-stats))
+      (-> (t2/hydrate dashboard [:collection :is_personal] [:dashcards :series] :tabs)
+          (assoc :last-edit-info (last-edit/edit-information-for-user @api/*current-user*))))))
+
+(api/defendpoint PUT "/:id"
+  "Update a Dashboard, and optionally the `dashcards` and `tabs` of a Dashboard. The request body should be a JSON object with the same
+  structure as the response from `GET /api/dashboard/:id`."
```
kamilmielnik, line 632: my 2 cents: ideally PUT endpoint response should have the same structure as GET endpoint response. Is that the case?

calherries: That's not the case, and wasn't before this PR. The goal of this series of PRs is to preserve the current behaviour and make the dashboard save much faster. I would say that making the PUT endpoint return the same as the GET endpoint is desirable and we can do that too

kamilmielnik: FYI before this PR `tabs` were not returned from PUT endpoint. This caused an issue here: https://github.com/metabase/metabase/pull/35311#issuecomment-1792115937

calherries: Wow, thanks for showing me this. I've been expecting a bug to happen like that for a while, but I was told not to worry about it. It's about time we do something about it! Open to ideas. I think it makes sense to have a shared schema for the FE + BE that both sides know how to edit.

PR #35229, opened 2023-10-30
Add tests for date filters

ranquild:


frontend/src/metabase/common/components/DatePicker/RelativeDatePicker/DateOffsetIntervalPicker/utils.ts
```diff
@@ -29,8 +29,12 @@ export function setOffsetInterval(
   value: DateOffsetIntervalValue,
   offsetValue: number,
 ): DateOffsetIntervalValue {
-  const sign = Math.sign(value.value);
-  return { ...value, offsetValue: Math.max(Math.abs(offsetValue), 1) * sign };
+  if (offsetValue === 0) {
```
ranquild, line 32: Fixed a bug with negative zero

kulyk (approved):

PR #35236, opened 2023-10-31
Skip archived or deleted cards on revert dashboard

### LINKED ISSUE #34884: Dashboard reversion should skip cards that have been deleted or archived
### Describe the bug

This is really two issues, but should be considered together for consistency.

### Issue 1

If you restore an old dashboard revision with a card for a deleted question, the restore request will fail on the backend with this error:

```
"ERROR: insert or update on table "report_dashboardcard" violates foreign key constraint "fk_dashboardcard_ref_card_id"
  Detail: Key (card_id)=(274) is not present in table "report_card"."
```

This is because the restore operation is trying to create a dashcard that references a deleted card. We should not create the dashcard for the deleted question when restoring the dashboard, because that is consistent with what happens when you restore a dashboard with a dashcard for an archived question.

### Issue 2

There's also funny quirk with restored dashboards and archived questions:
1. if you create a dashboard with a question, save
3. remove the question from the dashboard, save
4. archive the question
5. restore the dashboard at the point where the question was there (the question will be missing)
6. unarchive the question
7. refresh the dashboard (the question card is back, in the same position. If you have moved cards that overlap with the old position of the question card, they will be shifted)

This is different to if you:
1. create a dashboard with a question
2. archive the question
3. unarchive the question
4. refresh the dashboard (the question is gone)

### To Reproduce

Steps above.

### Expected behavior

[Slack conversation about this](https://metaboat.slack.com/archives/C057T1QTB3L/p1697787272300789?thread_ts=1697742159.706669&cid=C057T1QTB3L)

Reverting dashboards to a previous version should permanently not include any dashcards for questions that are deleted or archived at the time of restoration. Unarchiving any questions after restoration should not make the cards appear again in the dashboard.

### Logs

_No response_

```JSON
master (8f8ab29774)
```

### Severity

fairly high, blocks restoring dashboards in some cases

### Additional context

_No response_

qnkhuat:
Fixes https://github.com/metabase/metabase/issues/34884


src/metabase/models/dashboard.clj
```diff
@@ -237,6 +237,14 @@
...
         (dashboard-card/update-dashboard-card! update-card (id->current-card (:id update-card)))))))
 
+(defn- remove-invalid-dashcards
+  "Given a list of dashcards, remove any dashcard that link to cards that are either archived or not exist."
+  [dashcards]
+  (let [cards-ids       (remove nil? (map :card_id dashcards))
+        active-card-ids (when-let [card-ids (seq cards-ids)]
+                          (t2/select-pks-set :model/Card :id [:in card-ids] :archived false))]
+    (filter #(or (nil? (:card_id %)) (contains? active-card-ids (:card_id %))) dashcards)))
+
```
calherries, line 247: There is a way to implement this that more closely aligns with the english in the docstring, and is easier to verify that it's correct.
```suggestion
(defn- remove-invalid-dashcards
  "Given a list of dashcards, remove any dashcard that references cards that are either archived or not exist."
  [dashcards]
  (let [card-ids          (set (keep :card_id dashcards))
        active-card-ids   (when-let [card-ids (seq card-ids)]
                            (t2/select-pks-set :model/Card :id [:in card-ids] :archived false))
        inactive-card-ids (set/difference card-ids active-card-ids)]
    (remove #(contains? inactive-card-ids (:card_id %)) dashcards)))
```

src/metabase/models/dashboard.clj
```diff
@@ -247,6 +255,8 @@
                                                     :model/DashboardTab :dashboard_id dashboard-id)
         {:keys [old->new-tab-id]} (dashboard-tab/do-update-tabs! dashboard-id current-tabs (:tabs serialized-dashboard))
         serialized-cards          (cond->> serialized-cards
```
calherries, line 257: Can we name this `serialized-dashcards` while we're at it
```suggestion
        serialized-dashcards          (cond->> serialized-dashcards
```

test/metabase/models/dashboard_test.clj
```diff
@@ -634,6 +634,62 @@
...
+                                :size_x                 4
+                                :size_y                 4
+                                :visualization_settings {:text "Metabase"}}])
+    (create-dashboard-revision! dashboard-id false)
+
+    ;; 2. delete all the dashcards
+    (t2/delete! :model/DashboardCard :dashboard_id dashboard-id)
+    (create-dashboard-revision! dashboard-id false)
+
+    ;; cards can be deleted if its database is deleted
```
calherries, line 680: Cards can also be deleted if the question is archived and then the question is deleted from the archive. This comment
confused me because its not relevant to the test and it can get out of date. I think you can remove it.

test/metabase/models/dashboard_test.clj
```diff
@@ -634,6 +634,62 @@
...
+    (t2/delete! :model/Card will-be-deleted-card)
+    (t2/update! :model/Card :id will-be-archived-card {:archived true})
+
+    (testing "revert should not includes archived or deleted card ids (#34884)"
```
calherries, line 684:
```suggestion
    (testing "revert should not include archived or deleted card ids (#34884)"
```

test/metabase/models/dashboard_test.clj
```diff
@@ -634,6 +634,62 @@
...
 
+(deftest revert-dashboard-skip-archived-or-deleted-card-test
+  (t2.with-temp/with-temp [:model/Dashboard {dashboard-id :id}          {:name "A dashboard"}
+                           :model/Card      {will-be-archived-card :id} {}
+                           :model/Card      {will-be-deleted-card :id}  {}
+                           :model/Card      {unchanged-card       :id}  {}]
+    ;; 0. create a dashboard
+    (create-dashboard-revision! dashboard-id true)
+    ;; 1. add 3 cards and 1 text card
+    (t2/insert-returning-pks! :model/DashboardCard
```
calherries, line 645: Using `insert-returning-pks!` adds extra unnecessary details to the test compared to `with-temp`. We don't need to use
the first revision at step 0. So instead of using `insert-returning-pks!` to create the dashboard cards you could just
create the dashcards using `with-temp`.

test/metabase/models/dashboard_test.clj
```diff
@@ -634,6 +634,62 @@
...
+                                :col              0
+                                :size_x           4
+                                :size_y           4}
+                               {:dashboard_id           dashboard-id
+                                :dashboard_tab_id       nil
+                                :row                    4
+                                :col                    4
+                                :size_x                 4
+                                :size_y                 4
+                                :visualization_settings {:text "Metabase"}}])
```
calherries, line 673: `:visualization_settings` is not relevant to the test. We should avoid referencing it.

test/metabase/models/dashboard_test.clj
```diff
@@ -634,6 +634,62 @@
...
+    (t2/update! :model/Card :id will-be-archived-card {:archived true})
+
+    (testing "revert should not includes archived or deleted card ids (#34884)"
+      (revert-to-previous-revision Dashboard dashboard-id 2)
+      (is (=? #{{:card_id                unchanged-card
+                 :visualization_settings {}}
+                {:card_id                nil
+                 :visualization_settings {:text "Metabase"}}}
+              (t2/select-fn-set #(select-keys % [:card_id :visualization_settings])
+                                :model/DashboardCard :dashboard_id dashboard-id))))))
```
calherries, line 691: This test is much more specific than necessary, because it includes details on visualization_settings. All we need to
test is that the set of dashboard cards excludes dashboards for archived or deleted cards. To make tests less brittle, you should try to write tests that have as
little specific details
as possible. Here's an idea to make this test less specific while still matching the goal of the test:
```
(is (= #{unchanged-dashcard-id-1 unchanged-dashcard-id-2}
       (t2/select-pk-set :model/DashboardCard :dashboard_id dashboard-id)))
```
(assuming you use with-temp to create the dashcards and you have the dashcard IDs bounded already)

calherries (approved):
I have some nits but overall looks good

test/metabase/models/dashboard_test.clj
qnkhuat: It's implicit in the test but it was my intention to include a text card.

it's used to make sure `remove-invalid-dashcards` do not remove text cards by accident.

I'm not sure whether we have tests that cover this or not but it's nice to have it here.
Would a comment clear it up?

test/metabase/models/dashboard_test.clj
qnkhuat: I understand we can shorten this and have the dashboard card defined, but I want to simulate the test as close to how revision got built up as possible. If you see other tests in this namespace most of them do that. I think it's nice we keep it that way.

src/metabase/models/dashboard.clj
qnkhuat: cool, I've applied it in 84f8fdbabc

src/metabase/models/dashboard.clj
qnkhuat: yes 84f8fdbabc

test/metabase/models/dashboard_test.clj
calherries: A comment would clear up why the text card is included, but that's not my point here

I'm thinking of ways to minimize the number of changes you'd have to make to this test if unrelated parts of the codes change, such as the way text cards are represented.

We probably don't have a `create-mock-text-card` function (maybe we should?), so we can't get rid of all the details when creating the text dashcard. But we can at least avoid littering the `(is ...)` expression with these details because we should know the ID of the dashcard from when we create it.

To add to my suggestion from earlier you could even name the IDs and use them like this:
```
(is (= #{unchanged-question-dashcard-id unchanged-text-dashcard-id}
       (t2/select-pk-set :model/DashboardCard :dashboard_id dashboard-id)))
```

test/metabase/models/dashboard_test.clj
calherries: I don't find it necessary and there are other tests that don't do it this way.
https://github.com/metabase/metabase/blob/f1bb2e70d2c2d3f933222a30267adfcf39555efe/test/metabase/models/dashboard_test.clj#L300

The main reason we should try to avoid using `t2/insert-returning-pks!` is because it doesn't use the temp defaults, so it introduces a dependency on the concrete representation of a dashcard, even when the details of the representation aren't relevant to the test.

calherries:
I'm being super nitty here so feel free to merge this btw. I'm just trying to explore some ideas with you on how we could make future changes easier without having to rewrite millions of lines of test code. Because sometimes that results in strange decisions like making dashboards have `(nil? (:tab dashboard))` represent "this dashboard has one tab"

test/metabase/models/dashboard_test.clj
qnkhuat: > I'm thinking of ways to minimize the number of changes you'd have to make to this test if unrelated parts of the codes change, such as the way text cards are represented.

this makes sense


> (is (= #{unchanged-question-dashcard-id unchanged-text-dashcard-id}
       (t2/select-pk-set :model/DashboardCard :dashboard_id dashboard-id)))
       
This is not possible because dashboardcard are not "restored", when you revert, we create new dashboardcard, so the ids will not match.

qnkhuat:
I understand what you're heading at, but I still think revision tests should model how things are generated in the real world.

otherwise, for someone who doesn't have much understanding about revision, they will be confused with how the test is structured.

test/metabase/models/dashboard_test.clj
calherries: Ah, of course. You're right.

calherries:
Fair enough, the PR is solid and can be merged to me

PR #35134, opened 2023-10-27
Drop parameter_card.entity_id

qnkhuat:
Resolves https://github.com/metabase/metabase/issues/35097

More context: https://metaboat.slack.com/archives/CKZEMT1MJ/p1698351098837369

PR #35238, opened 2023-10-31
Manual Backport :  Drop parameter_card.entity_id #35134 

qnkhuat:
Manual backport of #35134 

The migration is not `v48.00-027` like we the original PR #35134, because we want the migration id reflect the version correctly.

See this [PR](https://github.com/metabase/metabase/pull/35239) that added the same migration we added in this one `v47.00-058` to master for completeness.

### DIFF
--- a/resources/migrations/000_migrations.yaml
+++ b/resources/migrations/000_migrations.yaml
@@ -15087,6 +15087,11 @@ databaseChangeLog:
       id: v47.00-058
       author: qnkhuat
       comment: 'Drop parameter_card.entity_id'
+      preConditions:
+        - onFail: MARK_RAN
+        - columnExists:
+            tableName: parameter_card
+            columnName: entity_id
       changes:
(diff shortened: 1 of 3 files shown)


resources/migrations/000_migrations.yaml
```diff
@@ -15083,6 +15083,20 @@ databaseChangeLog:
...
+      comment: 'Drop parameter_card.entity_id'
+      changes:
+        - dropColumn:
+            tableName: parameter_card
+            columnName: entity_id
+      rollback:
+        - sql:
+            sql: >-
+                ALTER TABLE parameter_card
+                ADD COLUMN IF NOT EXISTS entity_id CHAR(21) NULL;
```
qnkhuat, line 15098: in case someone ran the migration from `v48`, this column will be existed already

bshepherdson (approved):

PR #35249, opened 2023-10-31
fix entity-models-test

noahmoss:
This test was added from master in the last rebase, we need to add two new models to this list to get it to pass.


escherize (approved):
Looks good, My only question is: "We don't need to conditionally add these if we are in EE, right?".

noahmoss:
@escherize no, because these models are not defined in EE code

PR #35253, opened 2023-10-31
[MLv2] Format brush filter display names

snoe:
Also better localize and handle any passed in offset date times.


shared/src/metabase/shared/util/internal/time.cljs
```diff
@@ -166,13 +166,19 @@
...
 
 (defn- matches-date-time? [input]
-  (re-matches #"\d\d\d\d-\d\d-\d\dT\d\d:\d\d(?::\d\d(?:\.\d+)?)?" input))
+  (re-matches #"^\d\d\d\d-\d\d-\d\dT\d\d:\d\d(?::\d\d(?:\.\d+)?)?(?:Z|-\d\d:\d\d)?$" input))
+
+(defn- coerce-local-date-time [input]
+  (let [offset (re-find #"(?:Z|-\d\d:\d\d)$" input)]
+   (cond-> input
+    offset (subs 0 (- (count input) (count offset)))
+    :always (moment/utc moment/ISO_8601))))
```
snoe, line 181: This matches the clojure behavior.

shared/src/metabase/shared/util/internal/time.cljs
```diff
@@ -166,13 +166,19 @@
...
+  (re-matches #"^\d\d\d\d-\d\d-\d\d$" input))
 
 (defn- matches-date-time? [input]
-  (re-matches #"\d\d\d\d-\d\d-\d\dT\d\d:\d\d(?::\d\d(?:\.\d+)?)?" input))
+  (re-matches #"^\d\d\d\d-\d\d-\d\dT\d\d:\d\d(?::\d\d(?:\.\d+)?)?(?:Z|-\d\d:\d\d)?$" input))
+
+(defn- coerce-local-date-time [input]
+  (let [offset (re-find #"(?:Z|-\d\d:\d\d)$" input)]
+   (cond-> input
+    offset (subs 0 (- (count input) (count offset)))
```
metamben, line 180: It looks like this just strips the offset. Since we do a regex match anyway, this could be done simpler by `clojure.string/replace`.

metamben (approved):
Looks good in general.

PR #35265, opened 2023-10-31
[MLv2] Add card-or-table-id

### LINKED ISSUE #35257: [MLv2] BE — Add `card-or-table-id` method
The FE needs a way to get a `ColumnMetadata's` card or table ID to workaround field values fetching
Here's a possible signature:

```ts
function cardOrTableId(
  query: Query,
  stageIndex: number,
  column: ColumnMetadata,
): string | number {}
```

snoe:
Fixes #35257

### DIFF
--- a/src/metabase/lib/core.cljc
+++ b/src/metabase/lib/core.cljc
@@ -164,6 +164,7 @@
  [lib.field
   add-field
   field-id
+  card-or-table-id
   fieldable-columns
   fields
   find-visible-column-for-legacy-ref
--- a/src/metabase/lib/core.cljc
+++ b/src/metabase/lib/core.cljc
@@ -164,7 +164,7 @@
  [lib.field
   add-field
   field-id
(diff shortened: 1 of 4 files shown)


ranquild:
Please don't do anything and don't merge until we finish discussing this https://metaboat.slack.com/archives/C04CYTEL9N2/p1698765990489889

ranquild (commented):
The function returns `["card", 5]` instead of `card__5`. And for columns from tables I'd expect an integer value. cc @kulyk

snoe:
> The function returns `["card", 5]` instead of `card__5`. And for columns from tables I'd expect an integer value. cc @kulyk

Changed the return value, renamed to `legacy-card-or-table-id` to better capture these deprecated "ids"

ranquild (approved):

metamben (approved):
LGTM

PR #35268, opened 2023-10-31
Disable auto-commit (enable transactions) for read-only connections

metamben:
See https://metaboat.slack.com/archives/C97M2U8M6/p1698262151776969 for details.
This is step 1.


snoe (commented):
> look if no one cares about potential performance impacts then let's do it, just make sure you give it READ_UNCOMMITED or whatever to minimize the perf impact.

This was a comment from the discussion. Unsure if this part should change the isolation level as well.

metamben:
> > look if no one cares about potential performance impacts then let's do it, just make sure you give it READ_UNCOMMITED or whatever to minimize the perf impact.
> 
> This was a comment from the discussion. Unsure if this part should change the isolation level as well.

Part one is just about restoring the previous state. Part 2 is about performance, I plan to do experiment with isolation levels in that part.

snoe (approved):

Issue #35147, opened 2023-10-27
[Epic] Auto-wire up dashboard field filters


oisincoveney:
**Links**
- [Testing Plan](https://github.com/metabase/metabase/issues/35149)
- [Product Doc](https://www.notion.so/metabase/Auto-wire-up-dashboard-field-filters-0bd076a58939490d92b4422bd1cc7df6?pvs=4)
- [Figma](https://www.figma.com/file/1NRRygF4EOJ4s47t3tCHFO/Auto-wire-up-dashboard-field-filters?node-id=68%3A1030&t=JRjopBP1PgtJJ2rx-1)
- Feature Branch: [`35147-epic-auto-wire-up-dashboard-field-filters`](https://github.com/metabase/metabase/tree/35147-epic-auto-wire-up-dashboard-field-filters)
- Related Issues: #4767 

***Milestone 1: Add basic functionality***

```[tasklist]
### Tasks
- [ ] #35280
- [ ] #35461
```

PR #35280, opened 2023-11-01
Auto apply dashboard filters with the same field

oisincoveney:
Epic: #35147 

### Description

When applying a parameter mapping to a dashcard, this PR will apply the same mapping to dashcards that have the same field if the card doesn't already have a field applied to it.

I apply this logic within the `DashCardParameterMapper`, and uses an MBQL function to check whether the fields are the same. Keep in mind that only the fields that are exactly matching will be applied - if there's a field that has a filter applied to it, it won't be matched.

### How to verify

1. Create a dashboard and add three cards - two that come from the same table (it can even be the same card) and one that is completely unrelated. For example, make two cards from the Orders table, and one from the Reviews table.
2. Edit the dashboard and add a filter, like `Text contains...`. Then, in one of the cards from the Orders table, add a filter.
3. See that the filter is applied to the other card from the Orders table, but not the one from the reviews table.
4. Now try to change the mapping from one of the cards from the orders table - it should not change the other card.

### Checklist


oisincoveney:
* master
  * **PR #35280** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35280" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35280?utm_source=stack-comment).

frontend/src/metabase-lib/parameters/utils/targets.ts
```diff
@@ -56,3 +60,17 @@ export function getTargetFieldFromCard(
   const field = getParameterTargetField(target, metadata, question);
   return field ?? null;
 }
+
+export function compareMappingOptionTargets(
```
oisincoveney, line 64: @kulyk this is the function that we talked about - let me know if you think the location/logic could be changed here.

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -104,11 +124,67 @@ export function DashCardCardParameterMapper({
     _.isEqual(normalize(option.target), normalize(target)),
   );
 
+  const availableCardsWithoutEditingParameter = useMemo(() => {
```
ranquild, line 127: @oisincoveney all of this logic should be moved to redux actions. Maybe there won't be any component changes when you do so

frontend/src/metabase-lib/parameters/utils/targets.ts
ranquild: @oisincoveney please use `getParameterTargetField` to get `Field` instance

ranquild: Not sure but maybe you would need to extract the dimension part into `getParameterTargetDimension` or something like this. Please note that this function requires a `Question` instance and it's correct - otherwise some inner functions might not work properly

ranquild (commented):
Please do the following changes:
- Reuse the existing logic for extracting the dimension from a parameter target. You might need to break the existing function into 2
- Move all parameter-related business logic to `metabase-lib/parameters`. It should be possible to unit test it.
- Other logic should be moved to redux actions. If done currectly, I believe it should be a small function that passes relevant data from the redux store

mazameli:
I was curious about this branch so I spun it up, and already I can tell that the common usecase is going to feel so much nicer. 

I did notice one smaller quirk that I can show you here: https://www.loom.com/share/2ac62ab05a6a4d98a31ee1823664db0a

frontend/src/metabase/dashboard/actions/parameters.js
```diff
@@ -109,44 +117,123 @@ export const removeParameter = createThunkAction(
...
 );
 
-export const SET_PARAMETER_MAPPING = "metabase/dashboard/SET_PARAMETER_MAPPING";
-export const setParameterMapping = createThunkAction(
-  SET_PARAMETER_MAPPING,
-  (parameter_id, dashcard_id, card_id, target) => (dispatch, getState) => {
-    const dashcard = getState().dashboard.dashcards[dashcard_id];
-    const isVirtual = isVirtualDashCard(dashcard);
-    const isAction = isActionDashCard(dashcard);
+export function getParameterMappings(dashcard, parameter_id, card_id, target) {
```
npretto, line 120: This is not an action so I guess it should go to ./utils right? (bonus point: that file is typescript)

What do you think about converting it to named parameters in the process? I find functions with more than 3 arguments to be confusing on the call site.
(vscode should have a refactor helper to do it)

oisincoveney: Yep, I moved this out to `frontend/src/metabase/parameters/utils/mapping-options.js` - it's where `getParameterMappingOptions` is, so I think they belong together

frontend/src/metabase/dashboard/actions/parameters.js
```diff
@@ -109,44 +117,123 @@ export const removeParameter = createThunkAction(
...
+                  parameter_id,
+                  dc.card_id,
+                  selectedMappingOption.target,
+                ),
+              },
+            };
+          }
+          return null;
+        })
+        .filter(k => !!k);
```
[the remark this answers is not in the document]
oisincoveney: Gotcha, I was playing around with arrays here but it's really just easier to just push to another array while in the for-loop

frontend/src/metabase/dashboard/actions/cards.js
```diff
@@ -213,3 +229,58 @@ export const addActionToDashboard =
       }),
     );
   };
+
+export const autoApplyParametersToNewCard =
```
oisincoveney, line 233: I think this might be able to be split out as well, but considering the lack of use-cases in other parts of the code I don't think it's super necessary.

EmmadUsmani (commented):
Two points of feedback:

1) Regarding nomenclature: let's not use the term "auto apply" at all for this feature, since we already have used it for the [Prevent auto-apply filters in slow dashboards](https://www.notion.so/metabase/Prevent-auto-apply-filters-in-slow-dashboards-cc02d90dac9e42ee91f34c2a9810212f) project, and that is a completely different feature. We already have names in code like `dashboardState.autoApplyFilters` that refer to the older project. For this project, we should use "auto-wire" instead, e.g. `getAutoWireMappingsForDashcards`, since that's the term used in the product doc.

2) Regarding file structure: I know the existing practice that's prevalent throughout the codebase is to organize strictly [by kind](https://medium.com/@msandin/strategies-for-organizing-code-2c9d690b6f33), but I think it makes the code much more difficult to understand, so I really prefer organizing by component/feature, as I did in [`tabs.ts`](https://github.com/metabase/metabase/blob/22768f376e2cf856ed9ebac431694292a8258f05/frontend/src/metabase/dashboard/actions/tabs.ts).

In this PR we have functions spread across six different files in different directories to support a single feature, which makes it hard to find and understand the code for that feature. It also gives the misleading impression that these functions are generic utilities that are used across different features, which is not the intention. They were only created to support the auto-wire feature and make its code more readable.

I would propose grouping all of these functions in a single file or directory. For example, `frontend/src/metabase/dashboard/actions/auto-wire-parameters.ts`. We would also get the benefit of being able to use typescript in a new file. 

I don't think it's a crime to put non-action functions in a file in the `/dashboard/actions` directory, [we have helpers like this in files like `data-fetching.js`](https://github.com/metabase/metabase/blob/22768f376e2cf856ed9ebac431694292a8258f05/frontend/src/metabase/dashboard/actions/data-fetching.js#L92-L93). If you really want to be explicit that these are not actions, then make sure not to export them and also maybe prefix them with an `_` (e.g. `_getAutoWireMappingsForDashcards`) to suggest they are "private."

You could also create a subdirectory and move helpers to a separate file there, e.g. 
```
frontend/src/metabase/dashboard/actions/auto-wire-parameters
  /actions.ts <- could also name this index.ts
  /utils.ts
```

If you want to keep this feature coupled with the existing parameter actions code, then you can do something like this
```
frontend/src/metabase/dashboard/actions/parameters
  /actions.ts ← existing code
  /auto-wire
    /actions.ts
    /utils.ts
```

Or for a flatter structure
```
frontend/src/metabase/dashboard/actions/parameters
  /actions.ts ← existing code
  /auto-wire.ts
  /utils.ts
```

oisincoveney:
> Two points of feedback:
> 
> 1. Regarding nomenclature

Done. Good call on that, in hindsight it was confusing.

> 2. Regarding file structure

I've gone with the subdirectory approach - I think it makes the most sense and keeps things coupled in a general way. Let me know what you think!

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
```diff
@@ -592,6 +598,301 @@ describe("scenarios > dashboard > parameters", () => {
...
+        cy.findByTestId("dashboard-header").icon("filter").click();
+        popover().within(() => {
+          cy.contains("Text or Category").click();
+          cy.findByText("Is").click();
+        });
+
+        cy.findByTestId("dashboard-header").icon("add").click();
+
+        cy.findByTestId("add-card-sidebar").findByText("Orders Model").click();
+        cy.findByTestId("edit-dashboard-parameters-widget-container")
```
EmmadUsmani, line 778: Nit: same here, we can make a helper since it's repeated

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
```diff
@@ -592,6 +598,301 @@ describe("scenarios > dashboard > parameters", () => {
...
+
+        cy.findByTestId("dashboard-header").icon("filter").click();
+        popover().within(() => {
+          cy.contains("Text or Category").click();
+          cy.findByText("Is").click();
+        });
+
+        cy.findByTestId("dashboard-header").icon("add").click();
+
+        cy.findByTestId("add-card-sidebar").findByText("Orders Model").click();
```
EmmadUsmani, line 777: Nit: we can maybe make a helper for this (if one doesn't already exist) since repeat it 3 times

frontend/src/metabase-lib/parameters/utils/targets.ts
```diff
@@ -56,3 +56,20 @@ export function getTargetFieldFromCard(
...
```
EmmadUsmani, line 60: Since it's only used in `auto-wire-parameters/actions.ts` and `auto-wire-parameters/utils.ts`, how about we just move it to `auto-wire-parameters/utils.ts` to keep it local to that directory?

EmmadUsmani (approved):
@oisincoveney Looks great overall, love the changes you made! I left a few small comments you may want to look at before merging

frontend/src/metabase/dashboard/actions/auto-wire-parameters/utils.ts
```diff
@@ -0,0 +1,133 @@
...
+) {
+  if (!dashcardToCheck) {
+    return [];
+  }
+
+  return getParameterMappingOptions(
+    metadata,
+    null,
+    dashcardToCheck.card,
+    // TODO: mapping-options.js/getParameterMappingOptions needs to be converted to typescript as the TS
```
oisincoveney, line 49: Not really sure how to approach this one. Since `mapping-options.js` is JS and the 4th parameter is `dashcard = null`, typescript thinks the parameter must be `null` or `undefined`. I think it'll just need to keep this unless someone has a better solution

frontend/src/metabase-lib/parameters/utils/targets.ts
oisincoveney: I'll page @kulyk for this one 😄

kulyk: Let's leave it in `metabase-lib` — it should consolidate complex business logic for queries, parameters, etc.

PR #35289, opened 2023-11-01
docs - model index

jeff-bruemmer:
Adds docs for the option to surface individual records in search by matching against this column when editing model metadata.


cdeweyx (approved):
Looking good

Issue #35174, opened 2023-10-27
Add subscription-send and alert-send events


luizarakaki:
Add events when we send subscriptions and alerts:

topic: `subscription-send` and `alert-send`
user_id: id of the sub/alert creator
model: subscription/alert
model_id: subscription/alert id
details: 
- recipients array: list of all emails and external emails/slack channels/users that received the subscription
- filters: applied subscription filters
~~- attach_results: true/false~~

PR #35294, opened 2023-11-01
Add subscription-send and alert-send events

qwef:
Closes https://github.com/metabase/metabase/issues/35174

### Description

Adds subscription-send and alert-send events.

### DIFF
--- a/src/metabase/events/audit_log.clj
+++ b/src/metabase/events/audit_log.clj
@@ -118,10 +118,10 @@
      :recipients (map :recipients channels)}))
 
 (methodical/defmethod events/publish-event! ::pulse-event
-  [topic {:keys [id user details] :as object}]
-  ;; Check if objecct contains the keys that we want populated, if not then may be a unsubscribe/send event
-  (let [user-id     (if (some? user)
-                      user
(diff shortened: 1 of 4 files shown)


src/metabase/events/audit_log.clj
```diff
@@ -116,12 +118,15 @@
      :recipients (map :recipients channels)}))
 
 (methodical/defmethod events/publish-event! ::pulse-event
-  [topic {:keys [id details] :as object}]
+  [topic {:keys [id user details] :as object}]
   ;; Check if topic is a pulse or not (can be an unsubscribe event, which only contains email)
-  (let [details-map (if (some? id)
+  (let [user-id     (if (some? user)
```
qwef, line 123: Since user-id should be the creator of the pulse in this case, add an if statement that lets us choose the creator of the pulse

noahmoss (approved):

luizarakaki:
@qwef Noticed that `user_id` and `model_id` are null for alert-send and subscription-send.

Can we add the user sender id as `user_id` and the pulse id as the `model_id`?

luizarakaki (approved):

Issue #27782, opened 2023-01-19
Improved `card-view` event


luizarakaki:
## Context
This is the most important event. It describes who saw what when.

Today, this event in the `view_log` table contains the following properties:
- model_id (card_id)
- user_id
- timestamp
- cached? ignore_cache?

This event is triggered in three occasions: card creation, card view (get /api/card), and query execution. And there are some issues related to duplication: #25847, #25848

### The goal of this task is to:
1. Define clear triggers
2. Enrich the event with new properties

## Triggers
This event should be triggered whenever the user *sees* a question (regardless of showing or not results). In practice, when the user:
- Visits a question page (through a collection, dashboard, search, homepage, direct link)
- Sees a pinned question with visualization enabled
- Sees a question in a dashboard

*Notes*
- It shouldn't be triggered when a card is created (there is another event to track creation). But as the user sees the question right after it is created, we should log this view.
- It shouldn't be triggered on subscriptions and alerts (we will have other events for this purpose).

New `view_log` schema:
- id
- user_id
- model (`card`, `dashboard`, or `table`) // we will add actions in the future
- model_id
- context (card, dashboard, collection_pin) -- only for cards
- timestamp
- has_access (true if user has access, false if blocked by any permission check)
- error_message (error string if any)

We can likely drop the metadata field with cached and ignore cache as this is query_execution concern

likeshumidity:
Apologies if covered elsewhere or if I am misunderstanding, but why would we not track parallel to API calls? Or another way of asking: Shouldn't there be a 1:1 relationship between API calls and logged access/view events?

"shouldn't be triggered when a card is created", I am curious if this includes when a card is viewed after it is created? That is there is a natural separation between created event and view event if following along API in parallel. The POST event would be the create event, and the GET event would be the view.

Would the front-end have any involvement in logging or is this purely backend? Can avoid some security pitfalls by making entirely back-end and tracking API calls.

luizarakaki:
This is purely backend. But yes, with this proposal, there is a 1:1 relationship with `card` GET API calls. The problem is that the API call is not sufficient because the event depends also on the query execution that is async. So we need to update the event with the query result metadata (that is another API call `dataset`)

luizarakaki:
This should close:
- #25848
- #25847 
- #16517

PR #35302, opened 2023-11-01
Improve card_read event for audit_log and view_log

adam-james-v:
Closes: https://github.com/metabase/metabase/issues/27782

This PR works to improve the ViewLog table by changing when and how the can_read event ends up writing to the ViewLog.

The primary point where ViewLog entries are recorded occurs in `metabase.events.recent-views`, where the `events/publish-event!` method for `::event` is defined. Basically, if the event topic coming in is `:event/card-read`, `:event/dashboard-read`, or `:event/table-read` then the ViewLog is written to with the function `record-view!`.

This PR changes:
- the ViewLog schema:
  - ADD :has_access -> if the user who initiated the view has read acesss
  - ADD :error_message -> if an error occurred, put it here. Eg. "don't have permission", or perhaps a query error?
  - ADD :context -> the context in which the card was viewed (:collection, :question, :dashboard) -> comes originally from a qp middleware
  - (not done yet) REMOVE: :metadata. This is no longer needed. But removing it in a migration conflicted with some v_audit_log views, so tabled that until those stabilize
- the record-view! was being triggered on :event/card-query, but is now :event/card-read
- adds a test that recreates the API calls that happen when a user pins a card in a collection and ensures the View is NOT recorded. Then, when the page is refreshed, a card-read occurs, and that IS recorded.

This PR still needs:
- [ ] add a test to ensure the Dashboard view records properly
- [ ] add a test to ensure a view via the question url records properly (can probably just mock this with a GET to the correct endpoint api/card/:id)
- [ ] handle the situation where an unsaved question causes a :event/table-read (occurs because the query still needs to run but no card exists yet, so card-read doesn't occur.) see `metabase.query-processor.middleware.process-userland-query/add-and-save-execution-info-xform!` and also `metabase.api.dataset/run-query-async` to start tracking this down.

### DIFF
--- a/src/metabase/events/recent_views.clj
+++ b/src/metabase/events/recent_views.clj
@@ -17,12 +17,14 @@
 
 (defn record-view!
   "Simple base function for recording a view of a given `model` and `model-id` by a certain `user`."
-  [model model-id user-id metadata]
+  [{:keys [model model-id user-id context has_access]}]
   (t2/insert! :model/ViewLog
               :user_id  user-id
               :model    (u/lower-case-en model)
               :model_id model-id
-              :metadata metadata))
+              :context context
+              :has_access has_access
+              :metadata {}))
 
 (methodical/defmethod events/publish-event! ::event
   "Handle processing for a single event notification received on the view-log-channel"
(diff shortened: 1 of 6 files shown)


adam-james-v:
@luizarakaki , In the PR description I have 2 questions. Could you help me with those?

src/metabase/events/recent_views.clj
```diff
@@ -41,6 +42,7 @@
                    ;; we don't want to count pinned card views
                    ((complement #{:collection :dashboard}) context))
           (recent-views/update-users-recent-views! user-id model model-id))
-        (record-view! model model-id user-id metadata)))
+        (let [details (audit-log/model-details (merge object metadata) (keyword (name topic)))]
```
adam-james-v, line 45: To match the audit-log :details and the view-log :metadata, I've modified recent-views here.

@noah, I don't need a full review on this yet but I want to get a gut check from you if this is an ok way to achieve that.

noahmoss: looks fine to me

noahmoss:
(For context, I've pulled some of this PR into https://github.com/metabase/metabase/pull/35732)

PR #35277, opened 2023-11-01
Hide elements in create new dashboard modal

WiNloSt:
This PRs completes task 3-5 in this Epic: https://github.com/metabase/metabase/issues/34733

This PR implements:
- Hide public collections in create a new dashboard modal
    image
- Hide "+ New collection option" in FormCollectionPicker
    image
- Hide public collections in create a new collection modal
    image

#### How to verify
Query builder

Newly created question
1. Select `Create a new dashboard` for a question in a public collection
    **expectation**:
    - show all collections
    - always show `New collection`
1. Select `Create a new dashboard` for a question in a personal collection
    **expectation**:
    - show only personal collections
    - only show `New collection` option in personal collections
1. Select `New collection` for a question in a public collection
    **expectation**:
    - show all collections
1. Select `New collection` for a question in a personal collection
    **expectation**:
    - show only personal collections

PR #35213, opened 2023-10-30
[milestone 2] Prevent users from adding questions in personal collections to dashboards in public collections

WiNloSt:

PR #35171, opened 2023-10-27
Add E2E tests

WiNloSt:
Add E2E tests to cover moving dashboards/questions to a different collection type (personal/public)

PR #35317, opened 2023-11-02
Hide personal collections and items when configuring click behavior

WiNloSt:
This is the final part of the epic https://github.com/metabase/metabase/issues/34733

This PR implements
- Hide personal collections and dashboards when configuring click behavior linking to dashboards
- Hide personal collections and questions when configuring click behavior linking to questions

image

#### How to verify

Dashboard click behavior

1. Link to questions/dashboards for a question in a public collection
    **expectation**:
    - show only questions/dashboards in public collections
    - search should return only questions/dashboards in public collections
1. Link to questions/dashboards for a question in a personal collection
    **expectation**:
    - show all questions/dashboards
    - search should return all questions/dashboards

### DIFF
--- a/frontend/src/metabase-lib/parameters/utils/click-behavior.ts
+++ b/frontend/src/metabase-lib/parameters/utils/click-behavior.ts
@@ -7,6 +7,7 @@ import type {
   ClickBehaviorTarget,
   Dashboard,
   DashboardCard,
+  DashboardId,
   DatasetColumn,
   DatetimeUnit,
   Parameter,
@@ -411,7 +412,9 @@ function getParameter(
     (clickBehavior.linkType === "dashboard" ||
       clickBehavior.linkType === "question")
   ) {
-    const dashboard = (extraData.dashboards || {})[clickBehavior.targetId];
+    const dashboard = (extraData.dashboards || {})[
+      clickBehavior.targetId as DashboardId
+    ];
     const parameters = dashboard?.parameters || [];
     return parameters.find(parameter => parameter.id === target.id);
   }
--- a/frontend/src/metabase-lib/parameters/utils/click-behavior.ts
+++ b/frontend/src/metabase-lib/parameters/utils/click-behavior.ts
@@ -412,9 +412,8 @@ function getParameter(
     (clickBehavior.linkType === "dashboard" ||
       clickBehavior.linkType === "question")
   ) {
-    const dashboard = (extraData.dashboards || {})[
-      clickBehavior.targetId as DashboardId
-    ];
+    const dashboard =
+      extraData.dashboards?.[clickBehavior.targetId as DashboardId];
     const parameters = dashboard?.parameters || [];
     return parameters.find(parameter => parameter.id === target.id);
   }
--- a/frontend/src/metabase/collections/containers/FormCollectionPicker/FormCollectionPicker.tsx
+++ b/frontend/src/metabase/collections/containers/FormCollectionPicker/FormCollectionPicker.tsx
@@ -21,6 +21,7 @@ import { isValidCollectionId } from "metabase/collections/utils";
 import type { CollectionId } from "metabase-types/api";
 
 import { useSelector } from "metabase/lib/redux";
+import type { FilterItemsInPersonalCollection } from "metabase/containers/ItemPicker";
 import {
   PopoverItemPicker,
   MIN_POPOVER_WIDTH,
@@ -34,7 +35,7 @@ export interface FormCollectionPickerProps
   type?: "collections" | "snippet-collections";
   initialOpenCollectionId?: CollectionId;
   onOpenCollectionChange?: (collectionId: CollectionId) => void;
-  showOnlyPersonalCollections?: boolean;
+  filterPersonalCollections?: FilterItemsInPersonalCollection;
 }
 
 function ItemName({
@@ -60,7 +61,7 @@ function FormCollectionPicker({
   type = "collections",
   initialOpenCollectionId,
   onOpenCollectionChange,
-  showOnlyPersonalCollections,
+  filterPersonalCollections,
 }: FormCollectionPickerProps) {
   const id = useUniqueId();
   const [{ value }, { error, touched }, { setValue }] = useField(name);
@@ -107,7 +108,8 @@ function FormCollectionPicker({
 
   const isOpenCollectionInPersonalCollection = openCollection?.is_personal;
   const showCreateNewCollectionOption =
-    !showOnlyPersonalCollections || isOpenCollectionInPersonalCollection;
+    filterPersonalCollections !== "only" ||
+    isOpenCollectionInPersonalCollection;
 
   const renderContent = useCallback(
     ({ closePopover }) => {
@@ -132,11 +134,11 @@ function FormCollectionPicker({
             onOpenCollectionChange?.(id);
             setOpenCollectionId(id);
           }}
-          showOnlyPersonalCollections={showOnlyPersonalCollections}
+          filterPersonalCollections={filterPersonalCollections}
         >
           {showCreateNewCollectionOption && (
             <CreateCollectionOnTheGoButton
-              showOnlyPersonalCollections={showOnlyPersonalCollections}
+              filterPersonalCollections={filterPersonalCollections}
               openCollectionId={openCollectionId}
             />
           )}
@@ -148,7 +150,7 @@ function FormCollectionPicker({
       value,
       width,
       initialOpenCollectionId,
-      showOnlyPersonalCollections,
+      filterPersonalCollections,
       showCreateNewCollectionOption,
       openCollectionId,
       setValue,
--- a/frontend/src/metabase/collections/utils.ts
+++ b/frontend/src/metabase/collections/utils.ts
@@ -11,13 +11,25 @@ export function nonPersonalOrArchivedCollection(
   collection: Collection,
 ): boolean {
   // @TODO - should this be an API thing?
-  return !isPersonalCollection(collection) && !collection.archived;
+  return !isRootPersonalCollection(collection) && !collection.archived;
 }
 
-export function isPersonalCollection(
+export function isRootPersonalCollection(
   collection: Partial<Collection> | CollectionItem,
 ): boolean {
   return typeof collection.personal_owner_id === "number";
+}
+
+export function isPersonalCollection(
+  collection: Pick<Collection, "is_personal">,
+) {
+  return collection.is_personal;
+}
+
+export function isPublicCollection(
+  collection: Pick<Collection, "is_personal">,
+) {
+  return !isPersonalCollection(collection);
 }
 
 export function isInstanceAnalyticsCollection(
@@ -92,7 +104,7 @@ export function isPersonalCollectionOrChild(
   collectionList: Collection[],
 ): boolean {
   return (
-    isPersonalCollection(collection) ||
+    isRootPersonalCollection(collection) ||
     isPersonalCollectionChild(collection, collectionList)
   );
 }
@@ -129,14 +141,14 @@ export function canMoveItem(item: CollectionItem, collection: Collection) {
   return (
     collection.can_write &&
     item.setCollection != null &&
-    !(isItemCollection(item) && isPersonalCollection(item))
+    !(isItemCollection(item) && isRootPersonalCollection(item))
   );
 }
 
 export function canArchiveItem(item: CollectionItem, collection: Collection) {
   return (
     collection.can_write &&
-    !(isItemCollection(item) && isPersonalCollection(item))
+    !(isItemCollection(item) && isRootPersonalCollection(item))
   );
 }
 
@@ -191,7 +203,7 @@ function isPersonalOrPersonalChild(
     return false;
   }
   return (
-    isPersonalCollection(collection) ||
+    isRootPersonalCollection(collection) ||
     isPersonalCollectionChild(collection, collections)
   );
 }
@@ -200,7 +212,7 @@ export function canManageCollectionAuthorityLevel(
   collection: Partial<Collection>,
   collectionMap: Partial<Record<CollectionId, Collection>>,
 ) {
-  if (isPersonalCollection(collection)) {
+  if (isRootPersonalCollection(collection)) {
     return false;
   }
   const parentId = coerceCollectionId(collection.parent_id);
--- a/frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.tsx
+++ b/frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.tsx
@@ -92,7 +92,9 @@ const AddToDashSelectDashModal = ({
   if (shouldCreateDashboard) {
     return (
       <CreateDashboardModal
-        showOnlyPersonalCollections={isQuestionInPersonalCollection}
+        filterPersonalCollections={
+          isQuestionInPersonalCollection ? "only" : undefined
+        }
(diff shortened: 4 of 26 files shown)


WiNloSt:
* master
  * **PR #35135** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35135" target="_blank">Graphite</a> 
    * **PR #35141** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35141" target="_blank">Graphite</a> 
      * **PR #35142** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35142" target="_blank">Graphite</a> 
        * **PR #35171** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35171" target="_blank">Graphite</a> 
          * **PR #35213** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35213" target="_blank">Graphite</a> 
            * **PR #35214** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35214" target="_blank">Graphite</a> 
              * **PR #35215** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35215" target="_blank">Graphite</a> 
                * **PR #35277** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35277" target="_blank">Graphite</a> 
                  * **PR #35317** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35317" target="_blank">Graphite</a>  👈
    * **PR #35137** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35137" target="_blank">Graphite</a> 
      * **PR #35145** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35145" target="_blank">Graphite</a> 
      * **PR #35138** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35138" target="_blank">Graphite</a> 
        * **PR #35139** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35139" target="_blank">Graphite</a> 
    * **PR #35136** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35136" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35317?utm_source=stack-comment).

frontend/src/metabase/collections/containers/FormCollectionPicker/FormCollectionPicker.tsx
```diff
@@ -107,7 +108,8 @@ function FormCollectionPicker({
 
   const isOpenCollectionInPersonalCollection = openCollection?.is_personal;
   const showCreateNewCollectionOption =
-    !showOnlyPersonalCollections || isOpenCollectionInPersonalCollection;
+    filterPersonalCollections !== "only" ||
```
WiNloSt, line 111: The previous condition was `!showOnlyPersonalCollections`

Given `showOnlyPersonalCollections`  === `filterPersonalCollections === "only"`

that means `!showOnlyPersonalCollections` === `filterPersonalCollections !== "only"`

frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.tsx
```diff
@@ -92,7 +92,9 @@ const AddToDashSelectDashModal = ({
   if (shouldCreateDashboard) {
     return (
       <CreateDashboardModal
-        showOnlyPersonalCollections={isQuestionInPersonalCollection}
+        filterPersonalCollections={
+          isQuestionInPersonalCollection ? "only" : undefined
```
WiNloSt, line 96: `undefine` means showing all collections

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -86,7 +87,7 @@ function ItemPicker<TId>({
   className,
   showSearch = true,
   showScroll = true,
-  showOnlyPersonalCollections = false,
+  filterPersonalCollections,
```
WiNloSt, line 90: We can remove the default value because the default is now `undefined` means we'll show everything from both public collections and personal collections.

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -123,18 +124,14 @@ function ItemPicker<TId>({
 
     const collectionItems = list
       .filter(canWriteToCollectionOrChildren)
-      .filter(
-        showOnlyPersonalCollections
-          ? collection => collection.is_personal
-          : _.identity,
-      )
+      .filter(getCollectionFilter(filterPersonalCollections))
```
WiNloSt, line 127: a new function is defined below

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.tsx
```diff
@@ -178,11 +182,11 @@ function LinkedEntityPicker({
     });
   }, [clickBehavior, updateSettings]);
 
-  const { data: dashboard } = useDashboardQuery({
+  const { data: targetDashboard } = useDashboardQuery({
```
WiNloSt, line 185: Rename this to make it obvious that this is the selected dashboard as a click behavior target. So it's not confused with `dashboard` (introduced below) which is the dashboard that we're configuring the click behavior for.

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.unit.spec.tsx
```diff
@@ -0,0 +1,382 @@
...
+            models: ["card", "dataset"],
+            q: typedText,
+          });
+        });
+      });
+    });
+  });
+});
+
+function fromEntriesWithMultipleValues(
```
WiNloSt, line 365: The reason this one is needed is because previously I was comparing the `URLSearchParams` with:
https://github.com/metabase/metabase/blob/68788d775f5642046630ad34c98c4c5e0a5e9522/frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.unit.spec.tsx#L888-L891

This doesn't work with some of the tests in this file because it has multiple value for the same key i.e. `models=dataset&models=card`

using `Object.fromEntries` will just result in an object with the key `models` being either `dataset` or `card`.

This little util produces the value array, when it found multiple value for the same key.

WiNloSt, line 365: The reason this one is needed is because previously I was comparing the `URLSearchParams` with:
https://github.com/metabase/metabase/blob/68788d775f5642046630ad34c98c4c5e0a5e9522/frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.unit.spec.tsx#L888-L891

This doesn't work with some of the tests in this file because it has multiple value for the same key i.e. `models=dataset&models=card`

using `Object.fromEntries` will just result in an object with the key `models` being either `dataset` or `card`.

This little util produces the value array, when it found multiple value for the same key.

frontend/src/metabase/containers/ItemPicker/ItemPicker.tsx
```diff
@@ -150,19 +147,17 @@ function ItemPicker<TId>({
...
+        query.filter_items_in_personal_collection = filterPersonalCollections;
       }
     } else {
       query.collection = openCollectionId;
     }
 
-    if (models.length === 1) {
-      query.models = models;
-    }
+    query.models = models;
```
WiNloSt, line 157: This is caught during unit tests 😉 . It was kind of a bug that we fetche all results instead of just the specified models. There is one case where we set `models = ["dataset", "card"]` and we didn't apply `models` param at all.
#### After the fix
Screenshot 2023-11-09 at 7 36 58 PM

#### Before the fix
Screenshot 2023-11-09 at 7 37 01 PM


Although we used to fetch results from all models it doesn't make any difference in `ItemPicker` since FE will filter those results based on `models` anyway.

https://github.com/metabase/metabase/blob/68788d775f5642046630ad34c98c4c5e0a5e9522/frontend/src/metabase/containers/ItemPicker/ItemPickerView.tsx#L185

tl;dr: This change is needed for unit tests and as a result it fetches only necessary data from the API.

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.tsx
```diff
@@ -233,6 +237,12 @@ function LinkedEntityPicker({
     ],
   );
 
+  const dashboard = useSelector(getDashboard);
```
npretto, line 240: Maybe worth naming this `currentDashboard`? I know the other dashboard is called `targetDashboard` but considering before this change, dashboard meant the target one it could make sense

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.unit.spec.tsx
npretto: There's also `url.getAll` which could simplify the code if you want:
```js
const urlSearchParamsToObject = (searchParams) => {
  const obj = {};
  for (const [key] of Array.from(searchParams)) {
    const value = searchParams.getAll(key)
    obj[key] = value.length > 1 ? value : value[0];
  }
  return obj;
};
```

deniskaber (approved):
Added several inline comments

cdeweyx (approved):

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.tsx
WiNloSt: We use `dashboard` pretty much as a value returned from `getDashboard` everywhere, so I'd like to keep that consistency.

Also, usually with `currentSomething` it should be used to differentiate from other Something e.g. `previousDashboard` vs `currentDashboard` if we need to compare the 2 versions for some reasons. But that's not the case here either.

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.tsx
```diff
@@ -289,5 +300,8 @@ function LinkedEntityPicker({
   );
 }
 
-// eslint-disable-next-line import/no-default-export -- deprecated usage
-export default LinkedEntityPicker;
+function isPublicCollection(collection: Partial<Collection>) {
```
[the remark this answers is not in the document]
WiNloSt: Yeah, I'll make that into a util since it's used pretty much everywhere now. 🙏

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.unit.spec.tsx
WiNloSt: Fixed in 395e9d020150ca4621a2848135b2ff3c8419d500

frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.tsx
WiNloSt: This is intended to match the query we send to the search endpoints `filter_items_in_personal_collection: "only" | "exclude"`

Although its nuance is a little different because in the search endpoints we're filtering the results, which in this case are only dashboards, questions, or datasets (models)

`filterPersonalCollections: "only" | "exclude" | undefined` are meant not only for filtering items, but also collections that are shown in any picker UI too. 

#### collections
Screenshot 2023-11-10 at 2 23 31 PM

#### search results
Screenshot 2023-11-10 at 2 24 51 PM

Do you have any suggestion on improving this flag name?

frontend/src/metabase-lib/parameters/utils/click-behavior.ts
```diff
@@ -391,7 +392,9 @@ function getParameter(
     (clickBehavior.linkType === "dashboard" ||
       clickBehavior.linkType === "question")
   ) {
-    const dashboard = (extraData.dashboards || {})[clickBehavior.targetId];
+    const dashboard = (extraData.dashboards || {})[
```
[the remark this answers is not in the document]
WiNloSt: Fixed in 9c525a9f8a2626ff2bba5b21aa01ead33115a5d4

frontend/src/metabase/dashboard/components/ClickBehaviorSidebar/LinkOptions/LinkedEntityPicker/LinkedEntityPicker.tsx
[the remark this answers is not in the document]
WiNloSt: Fixed in c95c2dac57b144a04ecfba84fcda767aa205b2d9

frontend/src/metabase/collections/utils.ts
```diff
@@ -11,15 +11,27 @@ export function nonPersonalOrArchivedCollection(
...
 
-export function isPersonalCollection(
+export function isRootPersonalCollection(
   collection: Partial<Collection> | CollectionItem,
 ): boolean {
   return typeof collection.personal_owner_id === "number";
 }
 
+export function isPersonalCollection(
+  collection: Pick<Collection, "is_personal">,
```
WiNloSt, line 24: We want to allow passing `ROOT_COLLECTION` directly and since it doesn't contain all the necessary properties in `Collection`, we can't use `Collection` directly.

And this util only care if an object contains `is_personal`, so this type should be enough.

frontend/src/metabase/dashboard/selectors/selectors.js
```diff
@@ -77,6 +77,9 @@ export const getIsShowDashboardInfoSidebar = createSelector(
   sidebar => sidebar.name === SIDEBAR_NAME.info,
 );
 
+/**
+ * @type {(state: import("metabase-types/store").State) => import("metabase-types/api").Dashboard}
```
WiNloSt, line 81: This ensures TypeScript files using `getDashboard` will get the correct type without having to convert this whole file to TypeScript which is not trivial

frontend/src/metabase/lib/urls/collections.ts
```diff
@@ -40,8 +41,7 @@ export function collection(collection?: Collection) {
     return `/collection/${id}`;
   }
 
-  const isPersonalCollection = typeof collection.personal_owner_id === "number";
-  const slug = isPersonalCollection
+  const slug = isRootPersonalCollection(collection)
```
WiNloSt, line 44: This was implemented exactly as this function's implementation

https://github.com/metabase/metabase/blob/a57007a03a6a9ba9172f950311fc6cb2828982b3/frontend/src/metabase/collections/utils.ts#L17-L21

PR #35321, opened 2023-11-02
add legend to the static combo chart

alxnddr:
### Description

Adds legend to the static combo chart.

### How to verify

Dashcards to test:
- single card, multiple metric card -> should show metric names
- single card, signle metric -> should not show legend at all
- single card, breakout series -> should show formatted breakout names
- combined cards, both have single metric -> should show card names
- combined cards, one breakout series, the other multiple metric card -> should prefix card name like `cardName: `

### Demo

Screenshot 2023-11-02 at 8 33 23 PM

### Checklist

### DIFF
--- a/frontend/src/metabase/visualizations/echarts/cartesian/model/legend.ts
+++ b/frontend/src/metabase/visualizations/echarts/cartesian/model/legend.ts
@@ -1,0 +1,8 @@
+import type { CartesianChartModel } from "./types";
+
+export const getLegendItems = (chartModel: CartesianChartModel) => {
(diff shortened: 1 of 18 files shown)


frontend/src/metabase/visualizations/echarts/cartesian/model/legend.ts
EmmadUsmani, line 1: Since this isn't for the chart model computation I don't think it makes sense for it to be in this directory. A better place might be `static-viz/components/Legend/utils.ts`

EmmadUsmani (approved):
LGTM! I have one suggestion regarding file structure, please have a look

PR #35336, opened 2023-11-02
Fixing x-rays with custom columns (#16680)

### LINKED ISSUE #16680: X-ray automatic dashboards don't work when grouped on a custom column
**Describe the bug**
When a question is grouped on a custom column the X-ray and Compare drill-down actions lead to an error page.

**To Reproduce**
Remove the `includesExpressionDimensions` booleans found here https://github.com/metabase/metabase/blob/5c81fcbc7c06d3768e2006da635e41ab9d1e779b/frontend/src/metabase/modes/components/drill/AutomaticDashboardDrill.jsx#L29 and here https://github.com/metabase/metabase/blob/5c81fcbc7c06d3768e2006da635e41ab9d1e779b/frontend/src/metabase/modes/components/drill/CompareToRestDrill.js#L30 so that you can see the X-Ray and Compare options in the drilldown menu.

Then do the following...

1. Custom question > Sample Dataset > Orders
2. Custom Column 1 + 1 as "TestColumn"
3. Summarize Count by "TestColumn" and CreatedAt:Month
image
4. Click on a point in the series and then click X-Ray or "Compare..." actions beneath the "Automatic explorations" label.
Screen Shot 2021-06-18 at 11 52 54 AM
5. See the next page, which shows an error message: `No method in multimethod 'automagic-analysis' for dispatch value: null`
Screen Shot 2021-06-18 at 11 53 03 AM

markbastian:
Queries with custom columns were broken for two reasons:
- When names were computed for display, the humanization code expected fields
- The generated dashcard queries did not preserve the base query's custom expression definitions

For the first point, the code in `metabase.automagic-dashboards.names` was generalized for both fields and expressions.

For the second, the logic that was formerly `preserve-joins` in `metabase.automagic-dashboards.core` has been generalized to `preserve-entity-element` and applied to both joins and expressions.

Fixes https://github.com/metabase/metabase/issues/16680


markbastian:
Here's a [visual review](https://www.loom.com/share/bc7b157cfb704828a2cc8a56329b1c26?sid=ab8bed8a-8ab7-4fab-a149-4ce5f2c956ef) of the results.

iethree (approved):
FE change looks good 👍

src/metabase/automagic_dashboards/core.clj
```diff
@@ -779,14 +779,14 @@
...
 
-(defn- preserve-joins
-  "Hack to shove back in joins when they get automagically stripped out by the question decomposition into metrics"
-  [entity dashboard]
-  (if-let [join-statement (get-in entity [:dataset_query :query :joins])]
+(defn- preserve-entity-element
+  "Ensure that elements of an original dataset query are preserved in dashcard queries."
+  [dashboard entity entity-element]
+  (if-let [element-value (get-in entity [:dataset_query :query entity-element])]
     (letfn [(splice-joins [dashcard]
```
dpsutton, line 786: you should rename `splice-joins` now that it handles joins and expressions. One thing is that it does an `assoc-in`. Are we worried about clobbering things here? Should it be aware that it needs to combine in some manner rather than replace? (its ok if the answer is no right now)

markbastian: +1 on `splice-joins` rename. The old code did an `assoc-in`. I had considered `conj` but I think this is a built-up query so for now it's ok to just replace.

markbastian: Renamed to `splice-elements`.

src/metabase/automagic_dashboards/names.clj
```diff
@@ -132,55 +157,46 @@
        :day-of-year
        :week-of-year)  (u.date/extract dt unit))))
 
+(defn- humanize-filter-value-op
+  "Create a humanized comparison string for an item, where the comparison string may have two values based on the
+  item type (temporal or non-temporal). For example 'today is after yesterday' vs. '5 is greater than 4'."
+  ([root [_ field-reference value] temporal-str nontemporal-str]
+   (let [{:keys [item-name effective_type base_type unit]} (item-reference->field root field-reference)]
+     (if (isa? (or effective_type base_type) :type/Temporal)
+       (tru "{0} {1} {2}" item-name temporal-str (humanize-datetime value unit))
```
dpsutton, line 166: I think we lose translation strings this way. We used to have `"{0} is not before {1}"` and other similar strings as arguments to `tru`. But now they are args to a function call which uses `(tru "{0} {1} {2}")` so our i18n script won't pick these strings up as needing translation

test/metabase/automagic_dashboards/core_test.clj
```diff
@@ -1598,3 +1598,69 @@
...
+        (let [left                 (query/adhoc-query
+                                     {:database (mt/id)
+                                      :type     :query
+                                      :query
+                                      {:source-table (mt/id :orders)
+                                       :expressions  {"TestColumn" [:+ 1 1]}
+                                       :aggregation  [[:count]],
+                                       :breakout     [[:expression "TestColumn"]
+                                                      [:field (mt/id :orders :created_at)
+                                                       {:base-type :type/DateTime :temporal-unit :month}]]}})
```
dpsutton, line 1615: is `:base-type` necessary here? I would expect just the temporal unit is necessary

src/metabase/automagic_dashboards/names.clj
dpsutton: background we use this [namespace](https://github.com/metabase/metabase/blob/master/bin/build/src/i18n/enumerate.clj#L66) to look for forms like `(trs "...")` and know we need to translate those strings. The way you have it here violates how we look for them so we won't translate these strings.

test/metabase/automagic_dashboards/core_test.clj
markbastian: No, removed from test.

src/metabase/automagic_dashboards/core.clj
markbastian: Modified `preserve-entity-element` to preserve any existing element data.

src/metabase/automagic_dashboards/names.clj
markbastian: Got it. Take a look now and I think it will be fixed.

ranquild (changes requested):
Can you add a test for the frontend changes?

dpsutton (approved):

ranquild (approved):

e2e/test/scenarios/custom-column/custom-column.cy.spec.js
```diff
@@ -25,6 +25,31 @@ describe("scenarios > question > custom column", () => {
...
+          expressions: { TestColumn: ["+", 1, 1] },
+        },
+      },
+      { visitQuestion: true },
+    );
+    cy.get(".dot").eq(5).click({ force: true });
+    popover()
+      .findByText(/Automatic Insights/i)
+      .click();
+    popover().findByText(/X-ray/i);
```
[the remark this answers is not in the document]
markbastian: Done

PR #35342, opened 2023-11-02
[AuditV2] Add Audit deprecation banner

iethree:
Closes https://github.com/metabase/metabase/issues/35295

### Description

Add a deprecation notice to the audit app to give people some notice before we remove it.

Screen Shot 2023-11-02 at 2 48 35 PM

### How to verify

1. Go to the old audit app
2. See helpful banner
3. Make sure links work

### Checklist

### DIFF
--- a/enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditApp.jsx
+++ b/enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditApp.jsx
@@ -1,15 +1,75 @@
 /* eslint-disable react/prop-types */
-import fitViewport from "metabase/hoc/FitViewPort";
+import { useMemo } from "react";
+import { jt, t } from "ttag";
 
+import Link from "metabase/core/components/Link";
+import { useCollectionListQuery } from "metabase/common/hooks";
+import { isInstanceAnalyticsCollection } from "metabase/collections/utils";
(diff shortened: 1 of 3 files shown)


enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditApp.jsx
```diff
@@ -1,15 +1,75 @@
...
+      <Flex align="center" px="xl" py="md">
+        <Icon name="info_filled" />
+        <Box pl="sm">
+          <Box>
+            {jt`This Audit section has been upgraded to the
+              ${(
+                <Link
+                  variant="brandBold"
+                  key="link"
+                  to={`${window.MetabaseRoot}collection/${auditCollection.id}`}
```
npfitz, line 40: I don't think I've seen us use MetabaseRoot in links before. Can this be replaced with `/` ?

npfitz (approved):
Aside from the one question, it looks good to me!

enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditApp.jsx
iethree: I'll double check, but I'm pretty sure we need this to make it work when metabase is hosted on subpaths

PR #35346, opened 2023-11-03
Fix calling API more than necessary

### LINKED ISSUE #25848: Pinning a card to a collection results in many views in "view_log" table
**Describe the bug**
When pinning a card to a collection, the view table generates several view events. E.g.:
1) when you pin a card, it generates 3 view events
2) when you see a card that's pinned, it generates 2 view events

**Logs**
NA

**To Reproduce**
1) New->question->people table. Save it. See the view_log table (you should have 1 view event)
2) Now pin the question and see the view_log table (you should have 4 view events)
3) Now refresh the page (you should have 6 view events)

**Expected behavior**
Viewing a pinned question should generate just 1 view event, not 2, or 3

**Screenshots**
NA

- Your browser and the version: Brave latest
- Your operating system: Pop-OS 22.04
- Your databases: Postgres 14.2
- Metabase version: 1.44.4
- Metabase internal database: Postgres 14.4

**Severity**
P2

**Additional context**
NA

WiNloSt:
Fixes https://github.com/metabase/metabase/issues/25848

[Slack discussion](https://metaboat.slack.com/archives/C505ZNNH4/p1698968942243609)

This ensure we don't render `QuestionResultLoader` and then unrender and render it again causing `POST /api/card/:id/query` to be called 2 times

#### To reproduce

1. Go to a collection view that has any question that can be visible when pinned
1. Refresh the page to clear the cache
1. Pin a question
1. Observe that there's only 1 `POST /api/card/:id/query`

### DIFF
--- a/frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionCard.unit.spec.tsx
+++ b/frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionCard.unit.spec.tsx
@@ -1,0 +1,57 @@
+import fetchMock from "fetch-mock";
+import {
+  createMockCard,
+  createMockCollection,
+  createMockCollectionItem,
+  createMockDataset,
+} from "metabase-types/api/mocks";
+import { ROOT_COLLECTION } from "metabase/entities/collections";
+import {
+  setupCardEndpoints,
+  setupCardQueryEndpoints,
+} from "__support__/server-mocks";
+import { renderWithProviders, screen } from "__support__/ui";
+import registerVisualizations from "metabase/visualizations/register";
+import { createMockState } from "metabase-types/store/mocks";
+import { createMockEntitiesState } from "__support__/store";
+import PinnedQuestionCard from "./PinnedQuestionCard";
+
+registerVisualizations();
+
+function setup() {
+  const card = createMockCard({
+    id: 1,
+  });
+  const collectionItem = createMockCollectionItem({
+    id: card.id,
+    type: "card",
+    collection_preview: true,
+  });
+  setupCardEndpoints(card);
+  setupCardQueryEndpoints(card, createMockDataset());
+  renderWithProviders(
+    <PinnedQuestionCard
+      item={collectionItem}
+      collection={createMockCollection(ROOT_COLLECTION)}
+      onCopy={jest.fn()}
+      onMove={jest.fn()}
+    />,
+    {
+      storeInitialState: createMockState({
+        entities: createMockEntitiesState({
+          questions: [card],
+        }),
+      }),
+    },
+  );
+}
+
+describe("PinnedQuestionCard", () => {
+  it("should render query card once", async () => {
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionLoader.tsx
```diff
@@ -41,11 +41,10 @@ const PinnedQuestionLoader = ({
 
   return (
     <Questions.Loader id={id} loadingAndErrorWrapper={false}>
-      {({ loading, question: loadedQuestion }: QuestionLoaderProps) => {
-        if (loading || !loadedQuestion.query()) {
```
WiNloSt, line 45: The first render has `loading: undefined` and `loadedQuestion` presented. I guess `loadedQuestion` is always presented because it's cached from a collection list.

So the issue was we:
1. render `QuestionResultLoader` while `loading: undefined`
2. render this block when `loading: true`
3. render `QuestionResultLoader` again after `loading: false`

kamilmielnik (approved):
Works well :+1:

uladzimirdev:
@WiNloSt do you think we can add some test to verify it?

paoliniluis:
Fixes https://github.com/metabase/metabase/issues/25848

frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionLoader.tsx
WiNloSt: This changes the initial [fix attempt](https://github.com/metabase/metabase/pull/35346/commits/e4d2269ee52fdc87bc9f6ea1afc43a70867cbc4f). Since it failed 1 E2E test on [displaying pivot table.](https://www.deploysentinel.com/ci/runs/6544a00a7bcc148681654fff)

The reason is that the `loadedQuestion` that's cached from `GET /api/collection/root/items` doesn't have the same `query` as the one loaded from `GET /api/card/:id`, that's why I still need to have this block that will render another loader before rendering `QuestionResultLoader`

WiNloSt (commented):
@uladzimirdev I added a unit test

frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionCard.unit.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+}
+
+describe("PinnedQuestionCard", () => {
+  it("should render query card once", async () => {
+    setup();
+
+    expect(await screen.findByTestId("visualization-root")).toBeInTheDocument();
+
+    expect(fetchMock.calls("path:/api/card/1/query")).toHaveLength(1);
```
WiNloSt, line 55: If you revert the code change, this expectation would fail.

uladzimirdev (approved):
LGTM 👏

PR #35333, opened 2023-11-02
[Dashboard] Simplify filters stickiness calculation

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/35078

### Description

Rely on CSS instead of manual calculation for a sticky filters container on the dashboard.

previous PRs were about js solution (when support of `position: sticky` wasn't that great)
https://github.com/metabase/metabase/pull/17406
https://github.com/metabase/metabase/pull/21069
https://github.com/metabase/metabase/pull/21848
https://github.com/metabase/metabase/pull/23692
https://github.com/metabase/metabase/pull/24669

### How to verify

Describe the steps to verify that the changes are working as expected.

1. follow steps from the issue https://github.com/metabase/metabase/issues/35078 or
1. go to the dashboard, add some parameters, find "header" elements in the dev tools and remove it, scroll and see that filters contains do not blink

### Demo

https://github.com/metabase/metabase/assets/125459446/d310b89c-d4d0-48d2-8fdd-f7c8344dbd8a

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #35358, opened 2023-11-03
Drop dead code of the dashboard filters stickiness calculation

uladzimirdev:
### Description

Follow up of https://github.com/metabase/metabase/pull/35333

### How to verify

Filters container on dashboard should be sticky always, except when it's mobile and 6+ filters added.

Dropped code was a leftover after moving from js solution to css

### Checklist

### DIFF
--- a/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
+++ b/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
@@ -1,9 +1,7 @@
 // TODO: merge with metabase/dashboard/containers/Dashboard.jsx
-import { createRef, Component } from "react";
+import { Component } from "react";
 import PropTypes from "prop-types";
 import _ from "underscore";
-
-import { getMainElement } from "metabase/lib/dom";
 
 import { DashboardHeader } from "metabase/dashboard/components/DashboardHeader";
 import SyncedParametersList from "metabase/parameters/components/SyncedParametersList/SyncedParametersList";
@@ -31,8 +29,6 @@ import {
 } from "./DashboardEmptyState/DashboardEmptyState";
 import { updateParametersWidgetStickiness } from "./stickyParameters";
 
-const SCROLL_THROTTLE_INTERVAL = 1000 / 24;
-
 // NOTE: move DashboardControls HoC to container
 
 class Dashboard extends Component {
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -42,76 +38,12 @@ class Dashboard extends Component {
     parametersListLength: 0,
   };
 
-  static propTypes = {
```
uladzimirdev, line 45: moved to the bottom

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -121,28 +53,15 @@ class Dashboard extends Component {
...
-    main.addEventListener("scroll", this.throttleParameterWidgetStickiness, {
-      passive: true,
-    });
-    main.addEventListener("resize", this.throttleParameterWidgetStickiness, {
-      passive: true,
-    });
   }
 
   async componentDidUpdate(prevProps) {
+    updateParametersWidgetStickiness(this);
```
uladzimirdev, line 61: when component is updated, we recalculate stickiness

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -384,10 +300,7 @@ class Dashboard extends Component {
                 {shouldRenderParametersWidgetInViewMode && (
                   <ParametersWidgetContainer
                     data-testid="dashboard-parameters-widget-container"
-                    ref={this.parametersWidgetRef}
-                    isNavbarOpen={isNavbarOpen}
```
uladzimirdev, line 388: those props weren't used, ref is not needed as we do not calculate scrollTop anymore

frontend/src/metabase/dashboard/components/Dashboard/stickyParameters.js
```diff
@@ -2,6 +2,12 @@ import { isSmallScreen } from "metabase/lib/dom";
 
 export const MAXIMUM_PARAMETERS_FOR_STICKINESS = 5;
 
+const checkIfParametersWidgetShouldBeSticky = dashboard =>
```
uladzimirdev, line 5: just moved to the top as functions defined with `const` shouldn't be used before declaration

frontend/src/metabase/dashboard/containers/DashboardApp/DashboardApp.jsx
```diff
@@ -103,7 +101,6 @@ const mapStateToProps = state => {
     isLoadingComplete: getIsLoadingComplete(state),
     isHeaderVisible: getIsHeaderVisible(state),
     isAdditionalInfoVisible: getIsAdditionalInfoVisible(state),
-    embedOptions: getEmbedOptions(state),
```
uladzimirdev, line 106: it was used in a single place, which was removed above

frontend/src/metabase/dashboard/components/Dashboard/stickyParameters.js
WiNloSt: I'd revert the change.

This is only necessary if the function is called directly in the same scope, but `checkIfParametersWidgetShouldBeSticky` is called inside another function and I think this is fine.

Usually I prefer reading from top to bottom, so the previous version where we read `updateParametersWidgetStickiness` first and see it calls `checkIfParametersWidgetShouldBeSticky` inside then we continue reading the implementation below feels more natural than having to jump up in this version.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -356,6 +271,7 @@ class Dashboard extends Component {
               >
                 <DashboardHeader
                   {...this.props}
+                  isNavbarOpen={isNavbarOpen}
```
WiNloSt, line 274:
```suggestion
```

This is unnecessary as `DashboardHeader` already gets `isNavbarOpen` from redux

https://github.com/metabase/metabase/blob/a10cfbc0ec88de53967b638701a0e64e774d7fd6/frontend/src/metabase/dashboard/containers/DashboardHeader.jsx#L55

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
WiNloSt: Why don't we do this in `if (prevProps.dashboardId !== this.props.dashboardId)` as before? To me, it's more straightforward to replace `this.throttleParameterWidgetStickiness()` with `updateParametersWidgetStickiness(this)`

WiNloSt (approved):
Looks good to me with some questions

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
uladzimirdev: yes, but long time ago spreading props was considered as a harmful behaviour and I'd love not to search for it next time, but see it exactly passed

WiNloSt: You're right, this is pretty much the same thing as before. If we want to keep this, can we remove `isNavBarOpen` inside `DashboardHeader` to avoid confusion then?

frontend/src/metabase/dashboard/components/Dashboard/stickyParameters.js
uladzimirdev: agreed

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
uladzimirdev: I had in mind an edge case when you're on mobile and new parameter is added, but we don't allow to edit dashboards on mobile 🤔 

maybe you're right, I just wanted to be on the safe side as it's a very cheap check atm

frontend/src/metabase/dashboard/components/Dashboard/stickyParameters.js
```diff
@@ -2,7 +2,7 @@ import { isSmallScreen } from "metabase/lib/dom";
 
 export const MAXIMUM_PARAMETERS_FOR_STICKINESS = 5;
 
-// Dashboard Filters should always be sticky, except the case with small screen:
+// Dashboard filters container should always be sticky, except the case with small screen:
```
uladzimirdev, line 5: @WiNloSt reverted this change, thanks

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
uladzimirdev: I didn't get, `isNavBarOpen` is used inside DashboardHeader, I just wanted to pass the prop explicitly

WiNloSt: [As stated above ](https://github.com/metabase/metabase/pull/35358#discussion_r1381681524), `DashboardHeader` now gets `isNavBarOpen` from:
- prop and
- redux

So both are redundant, we only need one. I propose we remove the prop injected from redux
https://github.com/metabase/metabase/blob/a10cfbc0ec88de53967b638701a0e64e774d7fd6/frontend/src/metabase/dashboard/containers/DashboardHeader.jsx#L55

uladzimirdev: removed prop passing

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
uladzimirdev: I got one test broken (with the tall card), reverted changes to my initial variant

PR #35372, opened 2023-11-03
Fix wrongly mocked collection item key

WiNloSt:
This is a follow up from https://github.com/metabase/metabase/pull/35346.

That PR basically mock the wrong key, although `model` is already defaulted to `card` when called with `createMockCollectionItem`, so it makes no differenc behavior-wise.

This shouldn't be backported since I fixed this problem [on the backport PR itself.](https://github.com/metabase/metabase/pull/35361/commits/5e5d4de3d0485a517c27d0f5605ceadc113f53aa)

### DIFF
--- a/frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionCard.unit.spec.tsx
+++ b/frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionCard.unit.spec.tsx
@@ -24,7 +24,7 @@ function setup() {
   });
   const collectionItem = createMockCollectionItem({
     id: card.id,
-    type: "card",
+    model: "card",
     collection_preview: true,
   });
   setupCardEndpoints(card);


frontend/src/metabase/collections/components/PinnedQuestionCard/PinnedQuestionCard.unit.spec.tsx
```diff
@@ -24,7 +24,7 @@ function setup() {
   });
   const collectionItem = createMockCollectionItem({
     id: card.id,
-    type: "card",
+    model: "card",
```
WiNloSt, line 27: This change doesn't make a huge impact on the test since the `model` is already defaulted to `card`
https://github.com/metabase/metabase/blob/6634abddfcd0327565e67c2124b55601a823d48c/frontend/src/metabase-types/api/mocks/collection.ts#L20

iethree (approved):

PR #35366, opened 2023-11-03
refactor: remove renderDashCardActions in favor of a component

npretto:
### Description

This refactors `renderDashCardActions` to a proper component, creating a `DashCardActionsPanel` component that can store the state isDashCardTabMenuOpen introduced in https://github.com/metabase/metabase/pull/35324, instead of leaving it in the DashCard component that doesn't care about it.

The diff of the PR can be confusing but I tried to make small changes in isolated commits:

- 06d277e312ca023f99506aac1fc07b340779e461 moves `DashboardCardActionsPanel` inside the `DashCardActionButtons` component as it's basically just a styled div.
- d5abde45f1713761e9a9c8b744e85add813fb3ee ^ allows moving isDashCardTabMenuOpen state inside of `DashCardActionButtons`
- 1167fb63375bceb38d68c073ead5649740a6f091 removes renderDashCardActions as it just renders a component now, I did some testing (with console logs) and I didn't see extra rerenders caused by removing `useCallback`
- e0d9ef004c5b4ef24ed829d6f755833f143aa3ff renames `DashCardActionButtons` to `DashCardActionsPanel` as now that component contains the whole panel and I think it's a more appropriate name for it

PR #35327, opened 2023-11-02
left align dashcard actions on small cards

npretto:
## Description

Redo of https://github.com/metabase/metabase/pull/35241/files
I talked with Dawei, he agreed on a easier solution with a fixed container query.
I based this off [#35324 fix z-issues on move to tab menu](https://github.com/metabase/metabase/pull/35324) to avoid conflicts in the props of the component.

In order to support container queries I had to update @emotion/* libraries, they're all minor updates, do we have a special procedure for updating libs?

## How to reproduce
1. Resize a card to the min size and put it on the first column
2. When hovering it in edit mode, the actions should be left aligned and not go out of the viewport

https://github.com/metabase/metabase/assets/1914270/26f1b633-051d-48ed-9027-05282b7c14ca

PR #35324, opened 2023-11-02
fix z-issues on move to tab menu

npretto:
### Description

This fixes two issues with zIndex related to the "move to tab" menu
First issue is a zindex issue with grid maps: https://metaboat.slack.com/archives/C057T1QTB3L/p1698169540830189
move-to-tab-z-index

Second issue is a zindex issue when editing a link card: https://github.com/metabase/metabase/pull/34961/files#r1371364752

image

## Cause
Both issues were cause by me avoiding the portal for the menu for the first implementation.

I didn't use the portal because to avoid this:

https://github.com/metabase/metabase/assets/1914270/d0443929-4ea1-45bc-ac97-e3b17eba918a

Not using the portal made the menu a child of the dashcard actions panel, which is a child of the card, and so hovering the menu kept the the dashcard actions panel open. (Hovering the card, or having focus, keeps the actions panel visible)

That solution caused the two bugs above, I tried to solve them with via zIndex and no JS, but every thing that I fixed caused another zIndex issue so I stepped back and used the portal.

## Solution
In this PR I add a piece of state to check if there is an open sub menu, and in that case I keep the actions panel visible. 

I don't like that the state is currently in Dashcard, we could refactor `renderDashCardActions` to be a component so that we can keep the state internal to it.

### How to verify

1. Create/go to a dashboard with more than 1 tab
2. Create a link card, focus the input, the popover should open, hover the move to tab icon in the actions panel

For the second issue:
1. Create/go to a dashboard with more than 1 tab
2. Add a card with a map-> grid visualization
3. Hover the move to tab icon in the actions panel

### Demo

https://github.com/metabase/metabase/assets/1914270/fe965068-1b51-4b1e-bf14-bf0b2a84d483

### Checklist

- [ ] Tests have been added/updated to cover changes in this PR

PR #35378, opened 2023-11-03
feat: use name of card in undo toast message

npretto:
### Description
[Epic](https://github.com/metabase/metabase/issues/34367)
[Product doc](https://www.notion.so/metabase/Let-people-move-cards-between-dashboard-tabs-e797b321092b45a1bcbc4478d982a01e)

This helps distinguish the undo toast from one another using the card name if present, the card type otherwise.

Specs are in the [frame A05 of the figma](https://www.notion.so/metabase/Let-people-move-cards-between-dashboard-tabs-e797b321092b45a1bcbc4478d982a01e?pvs=4#1b2b364225a7477999c2b62a29b15e8d), linked in the prod doc

This also truncates the message to 75chars for all undo toast, there is a thread in figma about it. (I''m not posting the link as anyone with the link can open it)

### Demo

image

https://github.com/metabase/metabase/assets/1914270/b4e4b5b4-c85e-434e-9f53-e26282088ad6

### Checklist

### DIFF
--- a/frontend/src/metabase/containers/UndoListing.styled.tsx
+++ b/frontend/src/metabase/containers/UndoListing.styled.tsx
@@ -1,4 +1,6 @@
 import styled from "@emotion/styled";
+import type { BoxProps } from "metabase/ui";
+import { Box } from "metabase/ui";
 import Card from "metabase/components/Card";
 import { Icon } from "metabase/core/components/Icon";
 import Link from "metabase/core/components/Link";
@@ -13,6 +15,9 @@ export const UndoList = styled.ul`
   bottom: 0;
   margin: ${LIST_H_MARGINS};
   z-index: 999;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
 `;
 
 export const ToastCard = styled(Card)<{
@@ -33,7 +38,7 @@ export const CardContent = styled.div`
   justify-content: space-between;
 `;
 
-export const CardContentSide = styled.div`
+export const CardContentSide = styled(Box)<BoxProps>`
   display: flex;
   align-items: center;
   overflow: hidden;
--- a/frontend/src/metabase/containers/UndoListing.styled.tsx
+++ b/frontend/src/metabase/containers/UndoListing.styled.tsx
@@ -5,11 +5,13 @@ import Link from "metabase/core/components/Link";
 import { alpha, color, lighten } from "metabase/lib/colors";
 import { space } from "metabase/styled-components/theme";
 
+const LIST_H_MARGINS = space(2);
+
 export const UndoList = styled.ul`
   position: fixed;
   left: 0;
   bottom: 0;
-  margin: ${space(2)};
+  margin: ${LIST_H_MARGINS};
   z-index: 999;
 `;
 
@@ -20,6 +22,7 @@ export const ToastCard = styled(Card)<{
   padding: 10px ${space(2)};
   margin-top: ${space(1)};
   min-width: 310px;
+  max-width: calc(100vw - ${LIST_H_MARGINS} - ${LIST_H_MARGINS});
   transform: ${props => `translateY(${props.translateY}px)`};
   ${props => (props.color ? `background-color: ${color(props.color)}` : "")}
 `;
@@ -33,18 +36,16 @@ export const CardContent = styled.div`
 export const CardContentSide = styled.div`
   display: flex;
   align-items: center;
+  overflow: hidden;
 `;
 
-export const TruncatedText = styled.div`
-  max-width: 75ch;
-  overflow: hidden;
-  display: -webkit-box;
-  -webkit-line-clamp: 1;
-  -webkit-box-orient: vertical;
+export const ControlsCardContent = styled(CardContentSide)`
+  flex-shrink: 0;
 `;
 
 export const CardIcon = styled(Icon)`
   margin-right: ${space(1)};
+  flex-shrink: 0;
 `;
 
 export const DefaultText = styled.span`
--- a/frontend/src/metabase/containers/UndoListing.styled.tsx
+++ b/frontend/src/metabase/containers/UndoListing.styled.tsx
@@ -35,6 +35,14 @@ export const CardContentSide = styled.div`
   align-items: center;
 `;
 
+export const TruncatedText = styled.div`
+  max-width: 75ch;
+  overflow: hidden;
+  display: -webkit-box;
+  -webkit-line-clamp: 1;
+  -webkit-box-orient: vertical;
+`;
+
 export const CardIcon = styled(Icon)`
   margin-right: ${space(1)};
 `;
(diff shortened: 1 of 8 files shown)


npretto:
* master
  * **PR #35240** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35240" target="_blank">Graphite</a> 
    * **PR #35378** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35378" target="_blank">Graphite</a>  👈
    * **PR #35366** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35366" target="_blank">Graphite</a> 
    * **PR #35324** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35324" target="_blank">Graphite</a> 
      * **PR #35327** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35327" target="_blank">Graphite</a> 
    * **PR #35309** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35309" target="_blank">Graphite</a> 
      * **PR #35310** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35310" target="_blank">Graphite</a> 
    * **PR #35243** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35243" target="_blank">Graphite</a> 
    * **PR #35241** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35241" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35378?utm_source=stack-comment).

frontend/src/metabase/dashboard/actions/utils.unit.spec.ts
```diff
@@ -201,4 +206,57 @@ describe("dashboard > actions > utils", () => {
       expect(endTime - startTime).toBeLessThan(100); // 100 ms (locally this was 6 ms)
     });
   });
+
+  describe("getDashCardMoveToTabUndoMessage", () => {
```
npretto, line 210: Testing the thunk involved too much mocking so I tested the util function.

The whole functionality is also tested in the e2e for two cases, a named card and a text card so we should be covered

frontend/src/metabase/dashboard/selectors/selectors.js
```diff
@@ -17,7 +17,11 @@ import { getEmbedOptions, getIsEmbedded } from "metabase/selectors/embed";
 import Question from "metabase-lib/Question";
 
 import { isVirtualDashCard } from "../utils";
-import { getDashboardId } from "./selectors-typed";
+import {
```
npretto, line 20: This is technically a circular dependency, but jest seemed to run fine for me locally.

@uladzimirdev I think I saw you working on circular dependencies, do you have any idea if this is a problem?

frontend/src/metabase/dashboard/actions/utils.ts
```diff
@@ -57,3 +59,25 @@ export function haveDashboardCardsChanged(
     )
   );
 }
+
+export const getDashCardMoveToTabUndoMessage = (dashCard: StoreDashcard) => {
+  const virtualCardType =
+    dashCard.visualization_settings?.virtual_card?.display;
+
+  if (dashCard.card.name) {
+    return truncate(t`Card moved: ${dashCard.card.name}`, 75);
```
WiNloSt, line 68: I don't like using javascript to truncate text, because the information is lost. Usually in a CSS-truncated text, you could multiple-click the text to select and copy the whole text, but it won't work with this approach.

You should be able to achieve a similar effect using `max-width: 75ch`. Or just ballpark how long 75 characters would be and set it using `rem`

However if CSS approach is too hard to implement for any reason, I'd suggest you add `title` attribute to the text element as it's a good practice to have the full information in HTML, so screen readers can still read it. `title` also has a good default of showing the full untruncated text when hovering over the text.

frontend/src/metabase/dashboard/actions/utils.unit.spec.ts
```diff
@@ -201,4 +206,57 @@ describe("dashboard > actions > utils", () => {
...
+      });
+      expect(getDashCardMoveToTabUndoMessage(dashCard)).toBe("Card moved: foo");
+    });
+
+    it("should truncate long card names", () => {
+      const name = "a".repeat(100);
+      const dashCard = createMockDashboardCard({
+        card: createMockCard({ name }),
+      });
+      expect(getDashCardMoveToTabUndoMessage(dashCard)).toHaveLength(75);
```
WiNloSt, line 223: This might needs reimplementing if we decided to truncate text using CSS.

frontend/src/metabase/dashboard/selectors/selectors.js
WiNloSt: I suggest you take a look if your test would fail on the CI. I've encountered the same exact problem before, where Cypress runs fine locally, but would fail inevitably on CI.

I don't think circular dependency imports should be left in the code.

WiNloSt (commented):
Looks good for a starter. Please let me know if you would reimploement this using CSS or not.

frontend/src/metabase/dashboard/selectors/selectors.js
npretto: 🤔 not sure if something changed when I updated the branches, but I don't see the circular dependency anymore 🤔 

`selectors.js` imports from `selectors-typed` which is fine, and `selectors-typed`  doesn't import from the untyped file, and neither should the files imported by it

WiNloSt: I'd say as long as all required checks on this PR is green, then it'll be safe.

frontend/src/metabase/dashboard/actions/utils.ts
npretto: Design seems to be fine with truncating via css on all toast, solution has already been pushed to the PR in this commint: https://github.com/metabase/metabase/pull/35378/commits/004ba7b13943e40695242a663d1335ce158bf791

frontend/src/metabase/containers/UndoListing.styled.tsx
```diff
@@ -35,6 +35,14 @@ export const CardContentSide = styled.div`
   align-items: center;
 `;
 
+export const TruncatedText = styled.div`
```
npretto, line 38: I tried to refactor this to use Ellipsified but it wasn't "getting smaller" anymore when the viewport was small:
image

That seems to be related to the css that is used when lines==1

https://github.com/metabase/metabase/blob/4f53f052867a5f5cf7c6853353a779d6ec44917c/frontend/src/metabase/core/components/Ellipsified/Ellipsified.styled.tsx#L4-L8

deniskaber (commented):
I left inline comment regarding styling solution with css. 
After tackling it this PR should be fine by me.

frontend/src/metabase/containers/UndoListing.styled.tsx
npretto: Thanks @deniskaber , had to tweak it a bit but it's working now

deniskaber (approved):

WiNloSt (approved):
Looks good to me. Also tested out and it worked great.

PR #35380, opened 2023-11-03
Moving x-ray checks to backend for 15655 to reduce flakes

markbastian:
The following x-ray e2e tests are top contributors to x-ray flakes at 6.94% and 6.62%, respectively:
- "X-RAY" should work on a nested question made from base native question (metabase#15655)
- "COMPARE TO THE REST" should work on a nested question made from base native question (metabase#15655)

This change moves the x-ray dashboard generation aspects test load to a unit test and just checks that the insight buttons are present and clickable.

Note that there is still some scenario creation type code in the fixture that could potentially be simplified with some FE help.

### DIFF
--- a/e2e/test/scenarios/dashboard/x-rays.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/x-rays.cy.spec.js
@@ -5,7 +5,6 @@ import {
   summarize,
   visualize,
   startNewQuestion,
-  main,
   addOrUpdateDashboardCard,
   visitDashboardAndCreateTab,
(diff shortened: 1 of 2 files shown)


markbastian:
Based on what I'm seeing on [deploysentinel](https://www.deploysentinel.com/ci/analysis), most of the flakes happen before we even get to dashboard creation, so hopefully the direct creation of the queries will fix that.

iethree (approved):

noahmoss (approved):

PR #35381, opened 2023-11-03
impoving spacing, centering checkbox labels

npfitz:
### Description
Small PR to adjust sizes and spacing in Add Remove columns. Adjusted font weights, size of checkboxes, centered checkbox label with box, Margins and line heights.

### How to verify
Ensure that the Add Remove columns panel inspires more joy than it did previously.

Before: 
image

After:
image

### Checklist

- [ ] Get design approval

### DIFF
--- a/e2e/test/scenarios/question/settings.cy.spec.js
+++ b/e2e/test/scenarios/question/settings.cy.spec.js
@@ -489,10 +489,9 @@ function refreshResultsInHeader() {
 
 function getSidebarColumns() {
   return cy
-    .findByText("Columns", { selector: "label" })
+    .findByRole("list", { name: "chart-settings-table-columns" })
     .scrollIntoView()
     .should("be.visible")
-    .parent()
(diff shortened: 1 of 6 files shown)


iethree (approved):
giphy

mazameli:
Nice! Lots of good improvements already. I'm just seeing a couple things:

1.  You can see there's about 4px too much distance in between the checkbox and the column's type icon. (Branch on left, mock on right.)

image

2. There shouldn't be this duplicative black `Columns` heading here:

image

npfitz:
@mazameli I've updated the PR for the icon and label spacing. Removing the "Columns" can be done, but do we want to remove it in other places too? Such as Dashcard Viz Settings, or the Object Detail Viz Settings? Those are other places where these components are used

mazameli:
For posterity, Nick and I talked about this PR and I gave it the 👍🏼.

PR #35383, opened 2023-11-03
View details drills MBQLv2

### LINKED ISSUE #35248: [MLv2] [FE] Migrate View details drills (zoom, pk, fk-details)

deniskaber:
Closes https://github.com/metabase/metabase/issues/35248

### Description

Migrate Object details drill to use MBQLv2 lib api
In MBQLv2 Object details splits into pk, zoom, fk-details drills

### Demo

### Checklist


frontend/src/metabase/visualizations/components/ClickActions/ClickActionsPopover.unit.spec.tsx
```diff
@@ -439,6 +695,12 @@ async function setup({
     />,
   );
 
+  dispatch.mockImplementation(fn => {
+    if (Object.prototype.toString.call(fn) === "[object Function]") {
+      return fn(dispatch, view.store.getState);
+    }
+  });
```
kulyk, line 702: What is this block doing?

frontend/src/metabase/visualizations/components/ClickActions/ClickActionsPopover.unit.spec.tsx
```diff
@@ -362,6 +445,181 @@ describe("ClickActionsPopover", function () {
...
+                  ORDERS_ROW_VALUES.USER_ID,
+                ],
+                "source-table": ORDERS_ID,
+              },
+              type: "query",
+            },
+            display: "table",
+          },
+        },
+      ])(
```
kulyk, line 598: Why are we using `it.each` with only one test case?

frontend/src/metabase/visualizations/components/ClickActions/ClickActionsPopover.unit.spec.tsx
```diff
@@ -362,6 +445,181 @@ describe("ClickActionsPopover", function () {
...
+                  ORDERS_ROW_VALUES.PRODUCT_ID,
+                ],
+                "source-table": PRODUCTS_ID,
+              },
+              type: "query",
+            },
+            display: "table",
+          },
+        },
+      ])(
```
kulyk, line 544: IMO it's really hard to read such long and nested test cases. Also, it makes it harder to tell the difference between test cases

Would it be possible to simplify this? Maybe make some helper functions, extract common field references to constants / functions?

frontend/src/metabase/visualizations/components/ClickActions/ClickActionsPopover.unit.spec.tsx
ranquild: typeof fn === "function"?

ranquild (approved):
Agree with Anton's comments:
- `hasManyPks?` - we should remove `?` sign, and I'd rename it to `hasManyPkColumns` because the PK is only one
- the test file can be split 

But overall it seems to be working correctly

frontend/src/metabase-types/api/mocks/dataset.ts
```diff
@@ -25,6 +25,16 @@ export const createMockColumn = (
...
 
+export const createMockCustomColumn = (
+  data: Partial<DatasetColumn> = {},
+): DatasetColumn => {
+  return createMockColumn({
+    id: undefined,
+    table_id: undefined,
+    ...data,
+  });
+};
```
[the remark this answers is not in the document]
kulyk: I'm worried it's still possible to create a field with that helper — TypeScript would allow `id` and `table_id` and `...data` spreading will overwrite the default `undefined` values

```ts
createMockCustomColumn({ id: 10 })
```

kulyk (approved):

PR #35385, opened 2023-11-03
[MLv2] Filters — Refactor widget layout, make column values widget take full width

### LINKED ISSUE #35377: [MLv2] [Bug] FE - Layout issues with certain columns
image
image

kulyk:
Fixes #35377

Makes the `ColumnValuesWidget` take full width inside the popover. I had to set the max width for most of the widget components, and it became annoying that I had to update a few files. So, this PR introduces a `SimpleLayout` component that implements the widget UI used by most widget pickers.

### To verify

2. Filter > Category
3. Ensure the filter editing components take the full popover width (with all the operators) 

### Demo

| Before | After  |
|--------|-------|
| before            | after          |


frontend/src/metabase/common/components/FilterPicker/SimpleLayout.tsx
```diff
@@ -0,0 +1,53 @@
...
+  headerRight,
+  footerLeft,
+  children,
+  testID,
+  onSubmit,
+  onBack,
+}: SimpleLayoutProps) {
+  return (
+    <Box maw={MAX_WIDTH} data-testid={testID}>
+      <Header>
```
ranquild, line 38: This generic component makes me worry about potential changes in the future. WDYT about extending current `Header` and `Footer` components for the same purpose? And you could pass `children`

ranquild (commented):
Good idea, suggestion inline

frontend/src/metabase/common/components/FilterPicker/SimpleLayout.tsx
ranquild: Yep, but you could name slots as `children` I think

ranquild: @kulyk ^

kulyk: @ranquild how about that [`824c66f`](https://github.com/metabase/metabase/pull/35385/commits/824c66ffcdb7455dacdff320497c78bc94e41275)?

ranquild: @kulyk LGTM, but I'd remove `SimpleLayout` altogether and use new `Header` and `Footer` directly. I anticipate issues with extending if we leave `SimpleLayout`.

kulyk: @ranquild, please take another look [`d57f2ac`](https://github.com/metabase/metabase/pull/35385/commits/d57f2ac50e36044652167c7f210e37851ff77abb)

ranquild (approved):

PR #35387, opened 2023-11-03
Fix slow back button x-ray flakes

markbastian:
This PR attempts to fix these flaky tests:

- `e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js > scenarios > dashboard > dashboard back navigation should display a back to the dashboard button in table x-ray dashboards`
- `e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js > scenarios > dashboard > dashboard back navigation should display a back to the dashboard button in model x-ray dashboards`

Based on [deploysentinel](https://www.deploysentinel.com/ci/analysis?tab=flake), it looks like the problem is the `cy.wait("@dataset");` that occurs after creating a dashboard times our after waiting 5000ms. This PR attempts to fix this issue with the following changes:

- Set the number of generated cards to `const MAX_CARDS = 5;`. This should hopefully substantially reduce the amount of time spent on the FE rendering the dashboard as the dashboard payload should be much smaller.
- Increased the wait timeout on the two offending calls to 15000ms. Hopefully the first solution will always kick in first and this is never needed.

Between these two fixes, I'd hope we'll always see the dataset returned.

### DIFF
--- a/e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js
@@ -34,6 +34,7 @@ const { ORDERS_ID } = SAMPLE_DATABASE;
 const PG_DB_ID = 2;
 const PERMISSION_ERROR = "Sorry, you don't have permission to see this card.";
 const MAX_CARDS = 5;
+const MAX_XRAY_WAIT_TIMEOUT = 5000;
 
 describe("scenarios > dashboard > dashboard back navigation", () => {
   beforeEach(() => {
@@ -106,7 +107,7 @@ describe("scenarios > dashboard > dashboard back navigation", () => {
     () => {
       const cardTitle = "Sales per state";
       cy.visit(`/auto/dashboard/table/${ORDERS_ID}?show=${MAX_CARDS}`);
-      cy.wait("@dataset");
+      cy.wait("@dataset", { timeout: MAX_XRAY_WAIT_TIMEOUT });
 
       getDashboardCards()
         .filter(`:contains("${cardTitle}")`)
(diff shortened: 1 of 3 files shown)


e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js
```diff
@@ -104,8 +106,8 @@ describe("scenarios > dashboard > dashboard back navigation", () => {
     { tags: "@slow" },
     () => {
       const cardTitle = "Sales per state";
-      cy.visit(`/auto/dashboard/table/${ORDERS_ID}`);
-      cy.wait("@dataset");
+      cy.visit(`/auto/dashboard/table/${ORDERS_ID}?show=${MAX_CARDS}`);
```
iethree, line 109: I can't get this `show` query param to do anything 🤔

markbastian: Did some digging into the API vs. code code. Core uses `show` to limit cards, but the API doesn't actually pass it through, so this isn't a good fix ATM. WDYT of upping the timeouts on those two calls? That _seems_ to be the issue.

markbastian: Pushed a change and now it should work.

markbastian:
Fixed `show` query param behavior.
show

e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js
iethree: Upping the timeouts is probably a fine band-aid to reduce the flakes but I've seen some of these API requests hit a minute or more, so it doesn't really address the issue.

iethree:
I hate to expose APIs just for the sake of testing 😬  Is there a product reason to expose this `show` query Parameter? What do you think @luizarakaki ?

markbastian:
> I hate to expose APIs just for the sake of testing 😬 Is there a product reason to expose this `show` query Parameter? What do you think @luizarakaki ?

I pushed tests up for this, so we should be good to go if we either:
- Decide to expose `show` n cards. Note that the underlying domain api already supports and tests the numeric limit count. I am only making it visible at the API.
- Do not expose `show` n cards. If we do, I can just revert all changes except for the timeout increase (as long as @iethree and co. think it's a reasonable test fix).

e2e/test/scenarios/dashboard/dashboard-back-navigation.cy.spec.js
markbastian: I suspect the timing is just because of the amount of time required to hit every endpoint for every dashcard. I think only showing a few cards should fix this. From a product perspective, perhaps someone only wants to see the top 2-3 cards.

dpsutton (approved):

PR #35389, opened 2023-11-03
docs - embedding updates

jeff-bruemmer:
Some updates to our embedding docs based on the feedback from the sagacious @albertoperdomo and some other tweaks.

### DIFF
--- a/docs/embedding/static-embedding.md
+++ b/docs/embedding/static-embedding.md
@@ -30,7 +30,7 @@ To prevent people from editing the embedding URL to get access to other parts of
 your_metabase_embedding_url/your_signed_jwt?filter=true
 ```
 
-The signed JWT is generated using your [Metabase secret key](#regenerating-the-secret-key). The secret key tells Metabase that the request for filtered data can be trusted, so it's safe to display the results at the new embedding URL. Note that this secret key is shared for all static embeds, so whoever has access to that key will have access to all embedded artifacts. 
+The signed JWT is generated using your [Metabase secret key](#regenerating-the-secret-key). The secret key tells Metabase that the request for filtered data can be trusted, so it's safe to display the results at the new embedding URL. Note that this secret key is shared for all static embeds, so whoever has access to that key will have access to all embedded artifacts.
 
 If you want to embed charts with additional interactive features, like [drill-down](https://www.metabase.com/learn/questions/drill-through) and [self-service querying](../questions/query-builder/introduction.md), see [Interactive embedding](./interactive-embedding.md).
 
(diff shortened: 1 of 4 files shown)


docs/embedding/static-embedding.md
```diff
@@ -8,9 +8,9 @@ redirect_from:
 
 Also known as: static embedding.
 
-In general, embedding works by displaying a Metabase URL inside an iframe on your website. A **signed embed** (or static embed) is an iframe that's secured with a signed JSON Web Token (JWT). The signed JWT will prevent website visitors from poking around in your Metabase through the iframe.
+In general, embedding works by displaying a Metabase URL inside an iframe on your website. A **static embed** (or signed embed) is an iframe that's secured with a signed JSON Web Token (JWT). The signed JWT will prevent website visitors from poking around in your Metabase through the iframe.
```
albertoperdomo, line 11: >The signed JWT will prevent website visitors from poking around in your Metabase through the iframe.

I think this is inaccurate. The JWT covers multiple functions:

* authorize loading the artifact, e.g. dashboard - without the JWT in the URL the request will be denied by Metabase- this will work inside or outside of the iframe
* identify the ID of the resource to load in the iframe, e.g. the dashboard ID
* supply values for locked parameters 

My suggestion (wording would need to be revised):

```suggestion
In general, embedding works by displaying a Metabase URL inside an iframe on your website. A **static embed** (or signed embed) is an iframe that's loading a Metabase URL secured with a signed JSON Web Token (JWT). The URL will only load if a JWT is provided and that JWT is signed with the shared secret.

The JWT will also include a reference to the resource to load, e.g. the dashboard ID, and any values for locked parameters.
```

docs/embedding/static-embedding.md
```diff
@@ -8,9 +8,9 @@ redirect_from:
...
 
-Signed embeds can't be used with [data sandboxes](../permissions/data-sandboxes.md) or [auditing tools](../usage-and-performance-tools/audit.md) because signed JWTs don't create user sessions (server-side sessions).
+You can can't use static embeds with [data sandboxes](../permissions/data-sandboxes.md) or [auditing tools](../usage-and-performance-tools/audit.md) because signed JWTs don't create user sessions (server-side sessions).
```
albertoperdomo, line 13: Maybe worth mentioning that there is no drill-down, too?

albertoperdomo: Good opportunity to suggest looking at interactive embedding, too, which supports these features?

albertoperdomo (changes requested):

albertoperdomo (approved):

PR #35391, opened 2023-11-03
Audit log CRUD permission failures

johnswanson:
When we run `metabase.api.common/*-check` functions like `read-check` or `create-check`, publish an event when the check fails, along with enough context that, on the other side, the audit log handler can record the relevant event.

### DIFF
--- a/resources/migrations/000_migrations.yaml
+++ b/resources/migrations/000_migrations.yaml
@@ -16053,6 +16053,20 @@ databaseChangeLog:
               );
       rollback: # no change
 
+  - changeSet:
+      id: v48.00-053
+      author: johnswanson
+      comment: Increase length of `activity.model` to fit longer model names
+      changes:
+        - modifyDataType:
+            tableName: activity
+            columnName: model
+            newDataType: VARCHAR(32)
+      rollback:
+        - modifyDataType:
+            tableName: activity
+            columnName: model
+            newDataType: VARCHAR(16)
 
   # >>>>>>>>>> DO NOT ADD NEW MIGRATIONS BELOW THIS LINE! ADD THEM ABOVE <<<<<<<<<<
 
--- a/resources/migrations/000_migrations.yaml
+++ b/resources/migrations/000_migrations.yaml
@@ -16062,11 +16062,7 @@ databaseChangeLog:
             tableName: activity
             columnName: model
             newDataType: VARCHAR(32)
-      rollback:
-        - modifyDataType:
-            tableName: activity
-            columnName: model
-            newDataType: VARCHAR(16)
+      rollback: # not necessary
 
   # >>>>>>>>>> DO NOT ADD NEW MIGRATIONS BELOW THIS LINE! ADD THEM ABOVE <<<<<<<<<<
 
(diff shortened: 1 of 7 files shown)


test/metabase/api/common_test.clj
```diff
@@ -85,3 +88,50 @@
...
+        (is (= [[:event/update-permission-failure {:some :fancy-object}]]
+               @*events*))))
+    (testing "create-check"
+      (binding [*events* (atom [])]
+        (is (thrown? ExceptionInfo (api/create-check :model/Collection {})))
+        (is (= [[:event/create-permission-failure {:entity :model/Collection}]]
+               @*events*)))))
+
+  ;; teardown - underive events so they aren't dispatched in other tests
+  (underive ::permission-failure-event :metabase/event)
```
dpsutton, line 133: do these need to be in a finally block?

johnswanson: Good call, done!

johnswanson:
😌 this is ready to review again after the rebase!

test/metabase/api/common_test.clj
```diff
@@ -85,3 +89,49 @@
...
     (is (= [1 2 3] (api/parse-multi-values-param ["1" "2" "3"] parse-long)))))
+
+;; set up for testing permission failure event publishing
+(def ^:dynamic *events* nil)
+
+(methodical/defmethod events/publish-event! ::permission-failure-event
+  [topic event]
+  (swap! *events* conj [topic event]))
+
+(deftest *-check-functions-publish-events
```
noahmoss, line 100: Haven't seen the `*-` prefix before, is that to indicate the test is using a dynamic var instead of the real audit tables?

noahmoss (approved):

test/metabase/api/common_test.clj
johnswanson: I was using it as a placeholder, like "[something]-check" but I can see that it's confusing in this context - I'll change the name to just `check-functions-publish-events` instead.

resources/migrations/000_migrations.yaml
```diff
@@ -15989,6 +15989,21 @@ databaseChangeLog:
...
+      changes:
+        - modifyDataType:
+            tableName: activity
+            columnName: model
+            newDataType: VARCHAR(32)
+      rollback:
+        - modifyDataType:
+            tableName: activity
+            columnName: model
+            newDataType: VARCHAR(16)
```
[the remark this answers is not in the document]
johnswanson: Ooh great point, this is actually not good at all - if someone upgrades, puts the longer model names into the audit log, and then downgrades, they'll get an error (something like `ERROR:  value too long for type character varying(16)`). Seems better to just skip the rollback.

test/metabase/api/common_test.clj
noahmoss: Ohh gotcha

resources/migrations/000_migrations.yaml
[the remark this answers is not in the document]
noahmoss: Good point

PR #35392, opened 2023-11-03
Create Admin Setting for MB_SESSION_COOKIE_SAMESITE

JesseSDevaney:
**Problem:** In interactive embedding, the browser uses cookies to authenticate embedded users against Metabase. If the embedding app and the Metabase instance are not on the same domain, this won't work, as modern browsers won't use cross-domain cookies by default. So, we are creating an option to set a cookie setting called `SameSite` to allow for cross-domain requests with cookies.

**Context:**
- Closes: https://github.com/metabase/metabase-private/issues/182

---
**Desired Behavior**/**Acceptance Criteria**

_Behaviors_
- [ ] Add a "SameSite cookie setting" to allow admins to change the behavior of cross-site requests for use in cross-domain embedding
- [ ] If any authorized origin is different from the instance domain and SameSite cookie setting is not set to 'none', show a note that they may want to set SameSite cookie setting to 'none'
  image

_Tests_
- [ ] SameSite cookie setting is not shown in non-paid/EE versions
- [ ] SameSite cookie note is shown only when an authorized origin is different from the instance domain and SameSite is not set to 'none'

---
**How to Verify**

1. Run EE of metabase
2. Sign-in as admin
3. Add token with embedding permissions in admin settings > license and billing
4. Go to admin settings > embedding
5. Enable embedding
6. Go to interactive embedding
7. See that there is now an option to configure the SameSite cookie setting

image

8. See that the default value is `"lax"`
9. Change the setting to all different possible combinations `"lax" | "strict" | "none"`
  a. Verify that the setting is correctly updated and saved
  b. Verify that new network requests are sent with a parameter in `Set-Cookie` called `SameSite` which is the same as the SameSite cookie setting

Screenshot from 2023-11-03 18-46-22

---

✅ **Manually Verified**

https://github.com/metabase/metabase/assets/22608765/57e21c0b-c768-4552-83ab-1610030d4395

- [x] Manually verify that cross-domain interactive embedding works with SameSite setting set to `SameSite=none`
  - Requirements:
    - Metabase App Setup: i.e. `https://metabase-example-app.com`
      - authorized origins set to embedding app domain: `https://embedding-example-app.com`
      - SameSite cookie set to `SameSite=none`
      - SSO setup for embedding login
    - Embedding App Setup: i.e. `https://embedding-example-app.com`
      - setup to embed metabase app with SSO: `https://metabase-example-app.com`
    - https setup - required for use with `SameSite=none` (otherwise browser will reject requests)


enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppSameSiteCookieDescription/EmbeddingAppSameSiteCookieDescription.tsx
```diff
@@ -0,0 +1,72 @@
...
+    authorizedOriginsContainsNonInstanceDomain(embeddingAuthorizedOrigins);
+
+  return (
+    <Stack mb="1rem" spacing="sm">
+      <Text fw="bold">{t`SameSite cookie setting`}</Text>
+      {shouldDisplayNote && <AuthorizedOriginsNote />}
+      <Text>{t`Determines whether or not cookies are allowed to be sent on cross-site requests. You’ll likely need to change this to None if your embedding application is hosted under a different domain than Metabase. Otherwise, leave it set to Lax, as it's more secure.`}</Text>
+      <Text>{jt`If you set this to None, you'll have to use HTTPS (unless you're just embedding locally), or browsers will reject the request. ${(
+        <ExternalLink key="learn-more" href={docsUrl}>
+          Learn more
```
alxnddr, line 35: Missing translation

enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppSameSiteCookieDescription/EmbeddingAppSameSiteCookieDescription.tsx
```diff
@@ -0,0 +1,72 @@
...
+      w="22rem"
+      p="0.75rem"
+      style={{
+        border: `1px solid ${color("warning")}`,
+        borderRadius: `0.5rem`,
+      }}
+    >
+      <Text>{jt`You should probably change this setting to ${(
+        <Text span fw="bold">
+          None
```
alxnddr, line 56: Also needs to be wrapped with {t`...`}

enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppSameSiteCookieDescription/EmbeddingAppSameSiteCookieDescription.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+function AuthorizedOriginsNote() {
+  return (
+    <Center
+      data-testid="authorized-origins-note"
+      bg={alpha("warning", 0.1)}
+      w="22rem"
+      p="0.75rem"
+      style={{
```
alxnddr, line 49: I think we should still be using emotion instead of inline styles when we need to add anything custom that is not supported by the style props

alxnddr: The component looks pretty similar to `Alert`, can we use it?

alxnddr (commented):
Looks good, I left a couple of minor comments.

Also, the typography of this menu looks a bit off in terms of font-weight:
Screenshot 2023-11-06 at 8 24 12 PM

enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppSameSiteCookieDescription/EmbeddingAppSameSiteCookieDescription.tsx
JesseSDevaney: I will use emotion instead of inline styles going forward :pray: 

And thanks for surfacing the `Alert` component.
I thought it would be under the name `Warning`, so I did not end up finding one that was similar.

Thanks Sasha! :heart:

JesseSDevaney:
> Also, the typography of this menu looks a bit off in terms of font-weight:
> Screenshot 2023-11-06 at 8 24 12 PM

I opened up the [figma designs](https://www.figma.com/file/oPVegsKOEDj0B0Fm5Q0AHJ/Create-admin-setting-for-MB_SESSION_COOKIE_SAMESITE?type=design&node-id=25-140&mode=design&t=nJaKH2cS7BegF8Ek-0), and they do not look different to my eyes. 

Which part seems off to you? Do the descriptions seem too light?

alxnddr:
@JesseSDevaney you're right, looks as on the mockups. I am used to other similar components where we try to emphasize the option name by giving it higher font-weight value. @kdoh do you think it would make sense here too?

Similarly to this select:
Screenshot 2023-11-07 at 9 10 49 PM

Anyway, it is a cosmetic details which should not block merging this PR

alxnddr (approved):

albertoperdomo (approved):

mazameli:
Thanks for calling out the typographic inconsistency here, and thanks for following the Figma. 

For context, I think the older style that Sasha referenced is overly bold and dense, and we're in the process now of trying to systematize our typography on the design team, which will help us come back and address some of these inconsistencies with a more principled and consistent answer. 

For now, I agree this didn't need to block merging.

PR #35403, opened 2023-11-06
Includes dbms_version for database_connection_successful snoplow event

### LINKED ISSUE #35119: Snowplow: Add a property for `db_version` to database schema and events
## Context
We currently have a `database` schema that houses two events:
- `database_connection_failed`
- `database_connection_successful`

We track a property for the database that was used, but don't include anything around the `db_version` which can be helpful for making product decisions and debugging within the QP team. [More context](https://metaboat.slack.com/archives/C01MC2CKW2G/p1698169258356409).

## Expected Outcome
Add a new property `db_version` to the `database` Snowplow schema and populate it for the two aforementioned events that are housed in the schema right now. If we have the information this can be split up into `engine_version`, `engine_major_version`, and `engine_minor_version`. [More context](https://metaboat.slack.com/archives/C01MC2CKW2G/p1698356655068259?thread_ts=1698169258.356409&cid=C01MC2CKW2G).

qnkhuat:
Closes https://github.com/metabase/metabase/issues/35119


test/metabase/analytics/snowplow_test.clj
```diff
@@ -149,10 +149,11 @@
 
       (snowplow/track-event! ::snowplow/database-connection-successful
                              1
-                             {:database :postgres, :database-id 1, :source :admin})
+                             {:database :postgres, :database-id 1, :source :admin, :dbms_verrsion "14.1"})
       (is (= [{:data    {"database" "postgres"
                          "database_id" 1
                          "event" "database_connection_successful"
+                         "dbms_verrsion" "14.1"
```
piranha, line 156: verrsion?

piranha: IDK, this test feels kind of pointless. Maybe if it tested the endpoint for the database creation for the test being sent - that would make sense. Right now it just tests that mocking underlying mechanism of `track-event!` works, not that DBMS version is tracked.

qnkhuat: > Maybe if it tested the endpoint for the database creation for the test being sent - that would make sense

yes, I did that in `create-db-succesful-track-snowplow-test` and `create-database-test`

> IDK, this test feels kind of pointless
is it necessary to update this test? no as we can leave the payload as is and it'll still pass, but I think we should keep it as close as the actual schema as possible.

on whether or not it's pointless, I think it's not, it's actually also test if the payload can be serialized, and whether the event-key map to a the correct schema.

snowplow/iglu-client-embedded/schemas/com.metabase/database/jsonschema/1-0-1
```diff
@@ -1,11 +1,11 @@
 {
-  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-1#",
   "description": "Database connection",
   "self": {
     "vendor": "com.metabase",
     "name": "database",
     "format": "jsonschema",
-    "version": "1-0-0"
+    "version": "1-0-1"
```
piranha, line 8: Wondered about that one too, it should be synced with `metabase.analytics.snowplow`, right?

sansolovyov (approved):
Looks quite straightforward to me.

test/metabase/api/database_test.clj
```diff
@@ -223,20 +230,20 @@
...
+                                       (mt/user-http-request :crowberto :post 200 "database"
+                                                             (merge
+                                                              {:name    db-name
+                                                               :engine  (u/qualified-name ::test-driver)
+                                                               :details {:db "my_db"}}
+                                                              m)))]
+       (is (malli= [:and
+                    [:map-of :keyword :any]
+                    [:map [:id pos-int?]]]
+                   response))
```
calherries, line 244: This is essentially equivalent. The schema test only had `s/Keyword` because it needs that to allow other things in the map.
```suggestion
       (is (malli= [:map [:id pos-int?]]]
                   response))
```

test/metabase/api/database_test.clj
```diff
@@ -330,6 +337,22 @@
...
+        (let [dataset-def (tx/get-dataset-definition (data.impl/resolve-dataset-definition *ns* 'avian-singles))]
+          (mt/user-http-request :crowberto :post 200 "database"
+                                {:name    (mt/random-name)
+                                 :engine  driver/*driver*
+                                 :details (tx/dbdef->connection-details driver/*driver* nil dataset-def)}))
+        (is (=? {"database"     (name driver/*driver*)
+                 "database_id"  #hawk/malli :int
+                 "source"       "admin"
+                 "dbms_version" #hawk/malli [:fn some?]
+                 "event"        "database_connection_successful"}
```
calherries, line 353: `=?` already supports functions as values which get called to do asserts, you don't need malli for this
```suggestion
        (is (=? {"database"     (name driver/*driver*)
                 "database_id"  int?
                 "source"       "admin"
                 "dbms_version" some?
                 "event"        "database_connection_successful"}
```

test/metabase/api/setup_test.clj
```diff
@@ -172,15 +172,22 @@
...
+                                     :name    db-name
+                                     :details details
+                                     k        v}}
+              (testing "Database should be created"
+                (is (t2/exists? Database :name db-name)))
+              (testing (format "should be able to set %s to %s (default: %s) during creation" k (pr-str v) default)
+                (is (= (if (some? v) v default)
+                       (t2/select-one-fn k Database :name db-name)))))
+            (is (=? {"database"     "h2"
+                     "database_id"  #hawk/malli :int
```
calherries, line 186: same here

test/metabase/api/database_test.clj
calherries: Also you should use `string?` instead of `some?`

calherries (approved):
LGTM but haven't tested it.

PR #35405, opened 2023-11-06
Update search api documentation to includes available filters

qnkhuat:


src/metabase/api/search.clj
```diff
@@ -517,14 +517,22 @@
...
+  - `models`: only search for items of specific models. If not provided, search for all models
+  - `filters_items_in_personal_collection`: only search for items in personal collections
+  - `created_at`: search for items created at a specific timestamp
+  - `created_by`: search for items created by a specific user
+  - `last_edited_at`: search for items last edited at a specific timestamp
+  - `last_edited_by`: search for items last edited by a specific user
+  - `search_native_query`: set to true to search the content of native queries
+  - `verified`: set to true to search for verified items only (requires Content Management or Official Collections premium feature)
+
+  Note that not all item types support all filters, and the results will include only models that support the provided filters."
```
tsmacdonald, line 535: This line could use an example, I think.

src/metabase/api/search.clj
```diff
@@ -517,14 +517,26 @@
...
+  - `filters_items_in_personal_collection`: only search for items in personal collections
+  - `created_at`: search for items created at a specific timestamp
+  - `created_by`: search for items created by a specific user
+  - `last_edited_at`: search for items last edited at a specific timestamp
+  - `last_edited_by`: search for items last edited by a specific user
+  - `search_native_query`: set to true to search the content of native queries
+  - `verified`: set to true to search for verified items only (requires Content Management or Official Collections premium feature)
+
+  Note that not all item types support all filters, and the results will include only models that support the provided filters.
+  For example:
```
tsmacdonald, line 536:
```suggestion
  Note that not all item types support all filters, and the results will include only models that support the provided filters. For example:
```

src/metabase/api/search.clj
```diff
@@ -517,14 +517,26 @@
...
+  - `created-by` filter supports dashboard, model, action, card
+  - `verified` filter supports model and card
+  a search query that has both filters will only return items for model and card."
```
tsmacdonald, line 539:
```suggestion
  - The `created-by` filter supports dashboards, models, actions, and cards.
  - The `verified` filter supports models and cards.
  
  A search query that has both filters applied will only return models and cards."
```

tsmacdonald (approved):
:boat:

PR #35407, opened 2023-11-06
fix flaky relative time test `filter-datetime-by-date-in-timezone-relative-to-days-since-test`

qnkhuat:
Failed CI https://github.com/metabase/metabase/actions/runs/6767465207/job/18390157545?pr=35405

the fix is to use timezones that don't have DST


piranha:
Does it make sense to test for DST behavior here maybe? :)

qnkhuat:
> Does it make sense to test for DST behavior here maybe? :)

maybe for a separate test but I think it was not the original intention of this one

sansolovyov (approved):
LGTM

PR #35408, opened 2023-11-06
wip: Process linked filters having more than one join to the same table

### LINKED ISSUE #16872: Broken linked filters [field filters]
**Describe the bug**
The linked filters doesn't work anymore.

**To Reproduce**
1. Set up a new dashboard
2. Add a question 
3. Add 2 basic filters in the dashboard and link them with the question
4. Link the 2nd filter to the 1st one
5. Select a value for the 1st filter (random)
5. Error: The 2nd filter displays all the values instead of filtering the value related to the 1st filter.

**Expected behavior**
We expect to see the values in the 2nd filter that are related to the 1st filter. 
For example, my 1st filter is the name of my client. The 2nd filter is the name of its accounts. I should see in the 2nd filter the account names only for the client I selected in the 1st filter. However, I see the account names of my all clients which is a big problem.

**Screenshots**
Capture d’écran 2021-07-02 à 14 28 37
I should only have the accounts for my client "Quitoque". However, I have the full list of accounts.

```
{
  "browser-info": {
    "language": "fr-FR",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.11+9",
    "java.version": "11.0.11",
    "java.vm.version": "11.0.11+9",
    "os.name": "Linux",
    "os.version": "4.14.232-176.381.amzn2.x86_64",
    "user.language": "en",
    "user.timezone": "GMT"
  },
  "metabase-info": {
    "databases": [
      "postgres",
      "redshift"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "11.9"
      },
      "jdbc-driver": {
        "version": "42.2.18"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2021-06-29",
      "tag": "v0.40.0-rc1",
      "branch": "master-without-license-step",
      "hash": "1caf326"
    },
    "settings": {
      "report-timezone": "Europe/Brussels"
    }
  }
}
```
I shall add that the issue also happened with version v0.39.3 before we moved to pre-release version v0.40.0-rc1.

**Severity**
This is very urgent bug because all our clients can see other clients' account names.

sansolovyov:
Constraints for linked filters now carry over the information about joins necessary to construct correct queries. Works only for MBQL queries, since we have no data from native queries to latch onto.

Resolves #16872

### DIFF
--- a/src/metabase/api/dashboard.clj
+++ b/src/metabase/api/dashboard.clj
@@ -986,30 +986,46 @@
              field             (param->fields param false)]
          (assoc field :value value))))
 
-(defn filter-values-from-field-refs
+(defn- nested-card-results
   "Get filter values when only field-refs (e.g. `[:field \"SOURCE\" {:base-type :type/Text}]`)
   are provided (rather than field-ids). This is a common case for nested queries."
-  [dashboard param-key]
+  [param]
+  (for [{:keys [target] {:keys [card]} :dashcard} (:mappings param)
+        :let  [[_ dimension] (->> (mbql.normalize/normalize-tokens target :ignore-path)
+                                  (mbql.u/check-clause :dimension))]
+        :when dimension]
+    (custom-values/values-from-card card dimension)))
+
+(defn- -chain-filter
+  [dashboard param-key constraint-param-key->value query]
   (let [dashboard       (t2/hydrate dashboard :resolved-params)
+        constraints     (chain-filter-constraints dashboard constraint-param-key->value)
         param           (get-in dashboard [:resolved-params param-key])
-        results         (for [{:keys [target] {:keys [card]} :dashcard} (:mappings param)
-                              :let [[_ dimension] (->> (mbql.normalize/normalize-tokens target :ignore-path)
-                                                       (mbql.u/check-clause :dimension))]
-                              :when dimension]
-                          (custom-values/values-from-card card dimension))]
-    (when-some [values (seq (distinct (mapcat :values results)))]
-      (let [has_more_values (boolean (some true? (map :has_more_values results)))]
-        {:values          (cond->> values
-                                   (seq values)
-                                   (sort-by (case (count (first values))
-                                              2 second
-                                              1 first)))
-         :has_more_values has_more_values}))))
+        fields          (param->fields param true)
+        results         (if (seq fields)
+                          (map (if (seq query)
+                                 #(chain-filter/chain-filter-search % constraints query :limit result-limit)
+                                 #(chain-filter/chain-filter % constraints :limit result-limit))
+                               fields)
+                          (nested-card-results param))
+        values          (distinct (mapcat :values results))
+        has_more_values (boolean (some true? (map :has_more_values results)))]
+    (if (seq values)
+      ;; results can come back as [[v] ...] *or* as [[orig remapped] ...]. Sort by remapped value if it's there
+      {:values          (cond->> values
+                          (seq values)
+                          (sort-by (case (count (first values))
+                                     2 second
+                                     1 first)))
+       :has_more_values has_more_values}
+      (throw (ex-info (tru "Parameter {0} does not have any Fields associated with it" (pr-str param-key))
+                      {:param       (get (:resolved-params dashboard) param-key)
+                       :status-code 400})))))
 
 (mu/defn chain-filter :- ms/FieldValuesResult
-  "C H A I N filters!
+  "Chain Filters!
 
-  Used to query for values that populate chained filter dropdowns and text search boxes."
+   Used to query for values that populate chained filter dropdowns and text search boxes."
   ([dashboard param-key constraint-param-key->value]
    (chain-filter dashboard param-key constraint-param-key->value nil))
 
@@ -1017,33 +1033,12 @@
     param-key                   :- ms/NonBlankString
     constraint-param-key->value :- ms/Map
     query                       :- [:maybe ms/NonBlankString]]
-   (let [dashboard   (t2/hydrate dashboard :resolved-params)
-         constraints (chain-filter-constraints dashboard constraint-param-key->value)
-         param       (get-in dashboard [:resolved-params param-key])
-         fields      (param->fields param true)]
-     (when (empty? fields)
-       (or (filter-values-from-field-refs dashboard param-key)
-           (throw (ex-info (tru "Parameter {0} does not have any Fields associated with it" (pr-str param-key))
-                           {:param       (get (:resolved-params dashboard) param-key)
-                            :status-code 400}))))
-     (try
-       (let [results         (map (if (seq query)
-                                    #(chain-filter/chain-filter-search % constraints query :limit result-limit)
-                                    #(chain-filter/chain-filter % constraints :limit result-limit))
-                                  fields)
-             values          (distinct (mapcat :values results))
-             has_more_values (boolean (some true? (map :has_more_values results)))]
-         ;; results can come back as [[v] ...] *or* as [[orig remapped] ...]. Sort by remapped value if it's there
-         {:values          (cond->> values
-                             (seq values)
-                             (sort-by (case (count (first values))
-                                        2 second
-                                        1 first)))
-          :has_more_values has_more_values})
-       (catch clojure.lang.ExceptionInfo e
-         (if (= (:type (u/all-ex-data e)) qp.error-type/missing-required-permissions)
-           (api/throw-403 e)
-           (throw e)))))))
+   (try
+     (-chain-filter dashboard param-key constraint-param-key->value query)
+     (catch clojure.lang.ExceptionInfo e
+       (if (= (:type (u/all-ex-data e)) qp.error-type/missing-required-permissions)
+         (api/throw-403 e)
+         (throw e))))))
 
 (mu/defn param-values
   "Fetch values for a parameter.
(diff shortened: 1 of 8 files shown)


src/metabase/models/params/chain_filter.clj
```diff
@@ -200,7 +205,7 @@
                                            [:= :database.id database-id]
                                            [:not= :fk-field.fk_target_field_id nil]]})]
     (reduce
-     (partial merge-with merge)
+     (partial merge-with (partial merge-with into))
```
piranha, line 208: We were losing info about joins here, since the format is `{src-table {dst-table [joins...]}}`, so the previous code never got to merge vectors. Strictly speaking it's not necessary, since `chain-filter-mbql-query` is using explicit join information, but given that it still was "broken" fundamentally (did not return it promised it would), I think it makes sense to keep it.

src/metabase/models/params/chain_filter.clj
```diff
@@ -370,14 +375,16 @@
 
 (mu/defn ^:private chain-filter-mbql-query
   "Generate the MBQL query powering `chain-filter`."
-  [field-id                          :- ms/PositiveInt
+  [field                             :- Constraint
```
piranha, line 378: I'm not very happy about calling it `field`, but using `constraint` feels like it would be even more confusing, and I have no better ideas for the name. `param` is also loaded word...

src/metabase/models/params/chain_filter.clj
```diff
@@ -517,20 +524,22 @@
...
+(defn- cached-field-values [field constraints {:keys [limit]}]
   ;; TODO: why don't we remap the human readable values here?
   (let [{:keys [values has_more_values]}
         (if (empty? constraints)
-          (params.field-values/get-or-create-field-values-for-current-user! (t2/select-one Field :id field-id))
-          (params.field-values/get-or-create-linked-filter-field-values! (t2/select-one Field :id field-id) constraints))]
+          (params.field-values/get-or-create-field-values-for-current-user! (t2/select-one Field :id (:field-id field)))
+          (params.field-values/get-or-create-linked-filter-field-values! (t2/select-one Field :id (:field-id field)) constraints
+                                                                         (fn []
+                                                                           (unremapped-chain-filter field constraints {}))))]
```
piranha, line 534: This allows to remove circular dependency plus instead of passing `field :- Constraint` into `params.field-values/*`, just keep it to `params.chain-filter` ns.

But that breaks some `enterprise.sandbox` calls, and I haven't solved that yet.

src/metabase/models/params/field_values.clj
```diff
@@ -68,14 +68,13 @@
...
-  [fv-type field constraints]
+  [fv-type field cache-fn]
   {:pre [(field-values/advanced-field-values-types fv-type)]}
   (case fv-type
     :linked-filter
     (do
       (classloader/require 'metabase.models.params.chain-filter)
-      (let [{:keys [values has_more_values]} ((resolve 'metabase.models.params.chain-filter/unremapped-chain-filter)
-                                              (:id field) constraints {})
+      (let [{:keys [values has_more_values]} (cache-fn)
```
piranha, line 77: Dropping circular dependency here \o/

src/metabase/util.cljc
```diff
@@ -843,3 +843,18 @@
...
+  "(seek #(pos? (:value %)) [{:value -1} {:value 2}]) ; => {:value 2}"
+  ([pred coll] (seek pred coll nil))
+  ([pred coll not-found]
+   (reduce (fn [_ x]
+             (if (pred x)
+               (reduced x)
+               not-found))
+     not-found coll)))
+
+(defn seek=
```
piranha, line 857: Do we have a helper like that? Haven't figured out its name.

test/metabase/api/dashboard_test.clj
```diff
@@ -3097,10 +3097,9 @@
 (deftest valid-filter-fields-test
   (testing "GET /api/dashboard/params/valid-filter-fields"
     (letfn [(result= [expected {:keys [filtered filtering]}]
-                (testing (format "\nGET dashboard/params/valid-filter-fields")
```
piranha, line 3100: This just added a useless line really, no new information. I guess it's a leftover from some test refactoring.

qnkhuat:
Could you show me how to reproduce that bug first? the reproduction steps in the PR's description is not valid.

piranha:
> Could you show me how to reproduce that bug first? the reproduction steps in the PR's description is not valid.

https://www.loom.com/share/4cd87de233f94de1a45066eab686c32f

src/metabase/api/dashboard.clj
```diff
@@ -787,37 +787,53 @@
...
-    {:field-id field-id
+        :let  [ttag         (get-template-tag dimension card)
+               dimension    (condp mbql.u/is-clause? dimension
+                              :field        dimension
+                              :template-tag (:dimension ttag)
+                              (log/error "cannot handle this dimension" {:dimension dimension}))
+               [id options] (or
+                             ;; Get the field id from the field-clause if it contains it. This is the common case
+                             ;; for mbql queries.
+                             (mbql.u/match-one dimension [:field (id :guard integer?) options] [id options])
```
qnkhuat, line 808: [this](https://github.com/metabase/metabase/blob/efa24940113e26a2141a329bfdd96a74f725d11f/shared/src/metabase/mbql/schema.cljc#L323) is the schema of field options, it could contains stuff like base-type or binning.

we surely don't want this to be used as expression options right?

this was my worry from the previous PR([comment](https://github.com/metabase/metabase/pull/34837#discussion_r1375920142))

src/metabase/models/params/chain_filter.clj
qnkhuat: is that necessary for this PR? I'd leave that to a separate PR if we can

qnkhuat (commented):
Wdyt about passing the card query down to `chain-filter` so it could figure out the join situation there?

It doesn't make sense to have the logic to figure out how to join should not live in `api.dashboard` anymore, chain-fitler should be taking care of that now I think.

src/metabase/models/params/chain_filter.clj
piranha: It is necessary, since in here
```
((resolve 'metabase.models.params.chain-filter/unremapped-chain-filter) (:id field) constraints {})
 ```
 
`field` is not a Constraint, it's a field from db, so it has no information about joins and when caching is enabled changes stop working. And passing a function feels no worse than passing additional information about field (and polluting `field-values` with it).

File: src/metabase/api/dashboard.clj (hunk 7 above)
Comment (piranha) [reply src/metabase/api/dashboard.clj:808]: But we don't use those `options` for anything else except look for `:join-alias` on them, so should be fine? Our constraint gets options from template tag and from param, but not from a field dimension.

Or do you still think we could end up with excessive stuff in Constraint?

**Comment by piranha** [comment]
> Wdyt about passing the card query down to chain-filter so it could figure out the join situation there?

That sounds reasonable... `api.dashboard` is too knowledgeable about internal stuff, plus bringing all "joins" work in a single ns shouldn't hurt.

**Comment by piranha** [comment]
> Wdyt about passing the card query down to `chain-filter` so it could figure out the join situation there?
> 
> It doesn't make sense to have the logic to figure out how to join should not live in `api.dashboard` anymore, chain-fitler should be taking care of that now I think.

Moving all that chain-filter logic from dashboard into chain-filter namespace is quite easy obviously, and makes dashboard code much cleaner and chain-filter api easier to use (`(chain-filter/dashboard-chain-filter param-key query-params search-query?)`), but that's about it: restructuring to pass card down the line seems unfeasible - there is still need to unpack dashboard/parameters/cards to even get to the field ids...

So IDK, let me push whatever I have right now and let's discuss how it looks.

File: src/metabase/models/params/chain_filter.clj [hunk 8]
```diff
@@ -658,31 +675,135 @@
...
+                             ;; Attempt to get the field clause from the model metadata corresponding to the field.
+                             ;; This is the common case for native queries in which mappings from original columns
+                             ;; have been performed using model metadata.
+                             ((juxt :id #(nth (:field_ref %) 2))
+                              (qp.util/field->field-info dimension (:result_metadata card))))]
+        :when id
+        :let [query     (-> card :dataset_query :query)
+              join      (when (:join-alias options)
+                          (u/seek= :alias (:join-alias options) (:joins query)))
+              ;;[lhs rhs] (mbql.u/match (:condition join) [:field (id :guard integer?) _] id)
```
piranha, line 752: I have absolutely 0 ideas why is this working my repl and not working here. :(

qnkhuat:
Sorry, I didn't mean we move code from `api.dashboard` to `chain-filter`, I meant we need to propagate data from `param->fields` so that later on `find-all-joins` have enough information on how to join for this specific case.

The chain-filter call stack is deep enough that we shouldn't:
- make the chain longer by adding more wrapper
- having joins inferred at 2 places like we currently do

I'd rather keep the logic centralized and simple at the cost of propagating more data down the call stacks.
Please let me know if it makes sense.

piranha:
Hmmm... maybe it does, I'll need to experiment with this. I mean it's almost like every constraint should have a `card` attached to it. Not sure I like that approach, but maybe that makes sense, let's see.

PR #35411, opened 2023-11-06
Partial e2e tests - Visual glitches when joining same table and field multiple times

### LINKED ISSUE #19893: Visual glitches when joining same table and field multiple times
I noticed this while working on #15863. @flamber did mention visual glitches but the ones I ran into are different than the ones in his screenshot.

This occurs when joining the same table/column more than once via different saved questions:

**To Reproduce**
1. Simple question > Sample Dataset > Products - summarize Count group by Category - save as "Q1"
2. Simple question > Sample Dataset > Products - summarize Sum of Price group by Category - save as "Q2"
3. Simple question > Sample Dataset > Products - summarize Average of Rating group by Category - save as "Q3"
4a. Custom question > Saved Questions > "Q1" - join "Q2" on Category:

Notice that the columns are displayed as `Q2 Category = Q2 Category` instead of the expected `Q1 Category = Q2 Category`.

image

4b. Add another join against "Q3" on Category. Notice that the columns for the first join have now changed from "Q2" to "Q3".
Notice that the second join is displayed as `Q3 ⟕ Q3` rather than `Previous Data ⟕ Q3`.

image

The joined table/Saved Question names being wrong are the issue @flamber referred to in #15863, but the tables above the columns being wrong is a new issue (since they didn't exist back then).

bye

kamilmielnik:
Adds partial e2e tests for #19893.

### Description

The issue should be fixed in the backend, but I already wrote these e2e tests so let's have them as skipped for now.

### DIFF
--- a/e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
@@ -29,7 +29,7 @@ describe("issue 19893", () => {
     cy.signInAsAdmin();
   });
 
-  it("should display correct join source table when joining visited questions (metabase#19893)", () => {
+  it.skip("should display correct join source table when joining visited questions (metabase#19893)", () => {
     cy.createQuestion(QUESTION_1, {
       wrapId: true,
       idAlias: "questionId1",
--- a/e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
@@ -1,5 +1,5 @@
 import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
-import { restore } from "e2e/support/helpers";
+import { getNotebookStep, restore } from "e2e/support/helpers";
 
 const { PRODUCTS, PRODUCTS_ID } = SAMPLE_DATABASE;
 
@@ -99,12 +99,12 @@ const createQ1PlusQ2Question = (questionId1, questionId2) => {
 };
 
 const assertQ1PlusQ2Joins = () => {
-  cy.findAllByTestId("notebook-cell-item").then(items => {
-    cy.wrap(items[0]).should("contain", QUESTION_1.name);
-    cy.wrap(items[1]).should("contain", QUESTION_2.name);
-  });
+  getNotebookStep("join").within(() => {
+    cy.findAllByTestId("notebook-cell-item").then(items => {
+      cy.wrap(items[0]).should("contain", QUESTION_1.name);
+      cy.wrap(items[1]).should("contain", QUESTION_2.name);
+    });
 
-  cy.findByTestId("join-condition-0").within(() => {
     cy.findByLabelText("Left column").within(() => {
       cy.findByText(QUESTION_1.name).should("exist");
       cy.findByText("Category").should("exist");
--- a/e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
@@ -1,5 +1,5 @@
-import { restore } from "e2e/support/helpers";
 import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
+import { restore } from "e2e/support/helpers";
 
 const { PRODUCTS, PRODUCTS_ID } = SAMPLE_DATABASE;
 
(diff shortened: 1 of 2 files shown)


e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  },
+};
+
+describe("issue 19893", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsAdmin();
+  });
+
+  it("should display correct join source table when joining visited questions (metabase#19893)", () => {
```
kamilmielnik, line 32: Code changes in `JoinStep` fix this test case.

e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
```diff
@@ -0,0 +1,118 @@
...
+        ({ body: question }) => {
+          cy.visit(`/question/${question.id}/notebook`);
+        },
+      );
+    });
+
+    assertQ1PlusQ2Joins();
+  });
+
+  it.skip("should display correct join source table when joining non-visited questions (metabase#19893)", () => {
```
kamilmielnik, line 57: This test is still failing, that's why it's skipped.
I didn't add more tests because some assertions are not clear to me (I'm not sure what is the expected result).

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
```diff
@@ -53,19 +53,7 @@ export function JoinStep({
...
-    Lib.ColumnMetadata | undefined
-  >();
-
-  const lhsDisplayName = Lib.joinLHSDisplayName(
-    query,
-    stageIndex,
-    join || table,
-    selectedLHSColumn,
-  );
+  const [lhsDisplayName, setLhsDisplayName] = useState("");
```
ranquild, line 56: The change looks hacky. The display name is definitely not a "state". I think e2e tests make sense however the change should be somewhat different. Let's discuss it

e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
```diff
@@ -0,0 +1,118 @@
...
+          ],
+          "source-table": `card__${questionId2}`,
+        },
+      ],
+    },
+  });
+};
+
+const assertQ1PlusQ2Joins = () => {
+  cy.findAllByTestId("notebook-cell-item").then(items => {
```
ranquild, line 102: Can `getNotebookStep` be used here?

e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    assertQ1PlusQ2Joins();
+  });
+
+  it.skip("should display correct join source table when joining non-visited questions (metabase#19893)", () => {
+    cy.createQuestion(QUESTION_1, { wrapId: true, idAlias: "questionId1" });
+    cy.createQuestion(QUESTION_2, { wrapId: true, idAlias: "questionId2" });
+
+    cy.then(function () {
+      const { questionId1, questionId2 } = this;
```
ranquild, line 62: I haven't seen this pattern before. Not sure if it makes sense to me. Wdyt about
```
cy.get("@questionId1").then(questionId1 => {
  cy.get("@questionId2").then(questionId2 => {
    createQ1PlusQ2Question(questionId1, questionId2)....
  });
});
```

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
kamilmielnik: Reverted the change: 82dacaf8c1a5750c27bd9e7432e99c752579a9b2

We discussed it on a call and this hacky change should not be necessary once the real source of the issue is fixed in the backend (MLv2). But we'd need to double-check this later after backend is fixed (tests added in this PR should fail).

e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
kamilmielnik: I saw @WiNloSt use this pattern a few times, and I like it because it helps to save `n - 1` levels of indentation, where `n` is the count of variables you want to use. It makes code inside the callback shorter and more readable.

WiNloSt: Yeah, I've used this in a few places.

In an instance there are 3 variables and nesting `cy.get` would be very hard to read.

https://github.com/metabase/metabase/blob/7a202816024a0174c4a083ffee373087370cd43b/e2e/test/scenarios/dashboard/reproductions/13736-dashboard-should-work-if-some-cards-are-broken.cy.spec.js#L41-L44

e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
kamilmielnik: Thanks! dfb2f1d2a9482845a421f8de03a871f1c99cc3a5

ranquild (approved):

e2e/test/scenarios/question/reproductions/19893-joining-same-table-and-field-multiple-times.cy.spec.js
nemanjaglumac: @kamilmielnik 

For the future - if you're adding a skipped repro add `.skip` to the top level `describe` block rather than to the individual tests.

Also, please add `support/Repro` label to it, and label the underlying issue with `.Reproduced`

PR #35412, opened 2023-11-06
Fix E2E failure due to the DST

nemanjaglumac:
Failure in CI
https://www.deploysentinel.com/ci/runs/6548a5329f1ec992dfd5f053

Failure reproduced locally
image

After the fix
image

Stress-test the fix
https://github.com/metabase/metabase/actions/runs/6770255957/job/18398278318

```
    ✓ should show field metadata in a popover when hovering over a table column header: burning 1 of 5 (17797ms)
    ✓ should show field metadata in a popover when hovering over a table column header: burning 2 of 5 (17321ms)
    ✓ should show field metadata in a popover when hovering over a table column header: burning 3 of 5 (16704ms)
    ✓ should show field metadata in a popover when hovering over a table column header: burning 4 of 5 (16126ms)
    ✓ should show field metadata in a popover when hovering over a table column header: burning 5 of 5 (16867ms)
```

### DIFF
--- a/e2e/test/scenarios/visualizations/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/table.cy.spec.js
@@ -186,7 +186,7 @@ describe("scenarios > visualizations > table", () => {
           // semantic type
           cy.contains("No special type");
           // fingerprint
-          cy.findByText("-07:00");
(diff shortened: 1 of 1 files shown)


WiNloSt (approved):
Tested and it worked pretty well.

PR #35406, opened 2023-11-06
Use PUT /api/dashboard/:id to update dashcards and tabs

calherries:
Epic: https://github.com/metabase/metabase/issues/35209

This is the 2nd PR in a series of 3 PRs to change the way dashboards are saved to use `PUT /api/dashboard/:id`.

This also fixes https://github.com/metabase/metabase/issues/35351, though that wasn't the main goal initially.

The 1st PR in this series focused on BE prep for this change: https://github.com/metabase/metabase/pull/35223

We currently have to call two endpoints to update a dashboard with its nested entities. There should only be one.

**Before, to update a dashboard and its dashcards/tabs, you need to:** 
- Make a request to `PUT /api/dashboard/:id` to update the dashboard, excluding dashcards and tabs
- Then make a request to `PUT /api/dashboard/:id/cards` to update the dashcards and tabs

**With this PR, to update a dashboard and its dashcards/tabs, you need to:** 

1. (exact same steps as before)

OR

2. Make a single request to `PUT /api/dashboard/:id` to update the dashboard, _including_ dashcards and tabs (this method is currently untested and unused, but it works with manual testing)

In the previous PR, all code and tests still follow method 1. In this PR, I update the code and tests to follow method 2.

### BE changes
I've updated the tests that follow method 1, (i.e. all tests using the `PUT /api/dashboard/:id/cards` endpoint) to use method 2.

I've marked `/api/dashboard/:id/cards` as deprecated, to be removed in 49.

### FE changes
I've updated the code and tests that follow method 1, (i.e. all places using the `PUT /api/dashboard/:id/cards` endpoint) to use method 2.

### Follow up work: 

PR 3:
  - [De-duplicate BE events that do the same thing, i.e. multiple calls to `push-revision!`](https://github.com/metabase/metabase/pull/35413)

PR #35413, opened 2023-11-06
Only create one revision per request to PUT /api/dashboard/:id

calherries:

This is the 3rd PR in a series of 3 PRs to change the way dashboards are saved to use `PUT /api/dashboard/:id`.

1. https://github.com/metabase/metabase/pull/35223
2. https://github.com/metabase/metabase/pull/35406
3. This one

This PR changes the  `PUT /api/dashboard/:id` endpoint to only ever create one revision when a dashboard is updated. Previously it was possible for multiple revisions to be created from a single call to the endpoint.

It also removes the following events, which are unused:
- `:event/dashboard-reposition-cards`
- `:event/dashboard-update-tabs`
- `:event/dashboard-add-tabs`
- `:event/dashboard-remove-tabs`

### DIFF
--- a/src/metabase/api/dashboard.clj
+++ b/src/metabase/api/dashboard.clj
@@ -496,12 +496,12 @@
 (defn- do-update-dashcards!
   [dashboard current-cards new-cards]
   (let [{:keys [to-create to-update to-delete]} (u/classify-changes current-cards new-cards)]
+    (when (seq to-update)
+      (update-dashcards! dashboard to-update))
     {:deleted-dashcards (when (seq to-delete)
                           (delete-dashcards! (map :id to-delete)))
      :created-dashcards (when (seq to-create)
-                          (create-dashcards! dashboard to-create))
-     :updated-dashcards (when (seq to-update)
-                          (update-dashcards! dashboard to-update))}))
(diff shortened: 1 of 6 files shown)


markbastian (approved):
LGTM

qnkhuat (dismissed):
I'm fine with cleaning this up, but could check with the auditv2 team on whether or not removing these events is ok?

looks like we'll eventually remove `activity_feed` but we added a new `audit_log` table that has similar purpose.

I found this code that use those events https://github.com/metabase/metabase/pull/35434/files#diff-961e100cdcd88e4f74341ab90f9776a3a11a565dc8bb29cea31110d8cd5bb74b

cc @noahmoss

calherries:
Good idea. Having a look their [audit_log](https://github.com/metabase/metabase/pull/35434/files#diff-961e100cdcd88e4f74341ab90f9776a3a11a565dc8bb29cea31110d8cd5bb74bR59-R62) code, it seems audit v2 does not use any of the removed events.

calherries:
@qnkhuat I posed the question [here](https://metaboat.slack.com/archives/C064EB1UE5P/p1699335250046449)

PR #35414, opened 2023-11-06
Fix possibility of two revisions having most_recent=true

### LINKED ISSUE #35351: Adding a tab to a dashboard makes the dashboard shows double in collection view
### Describe the bug

Adding a tab to a dashboard makes the dashboard shows double in collection view. This has been tested on a fresh Metabase  instance, though this occurred with my usual testing instance as well.

### To Reproduce

1. Create a dashboard, after it's created and is in edit mode, click cancel (not sure it has any implication, but I did that)
2. Go to the root collection. See that there is only 1 dashboard
    Screenshot 2023-11-03 at 4 41 28 PM
4. Go to the dashboard again, edit, add a tab and save
5. Go to the root collection. Now you'll see 2 dashboards with the same name
    Screenshot 2023-11-03 at 4 41 39 PM
7. Refresh the page, it still shows 2 dashboards
8. Archive the dashboard, clicking one will automatically select the duplicate.
9. Restore the archive dashboard and go to the root collection. You'll only see 1 dashboard.

### Expected behavior

I should only see 1 dashboard each not also its duplicate

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.14.1+1-LTS",
    "java.version": "11.0.14.1",
    "java.vm.version": "11.0.14.1+1-LTS",
    "os.name": "Mac OS X",
    "os.version": "14.0",
    "user.language": "en",
    "user.timezone": "Asia/Bangkok"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2023-10-30",
      "src_hash": "a3c4f2dde12e997149db5929f356c3181ffa94d3",
      "tag": "v1.47.1-SNAPSHOT",
      "hash": "84c16b6"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

Not blocking but looks super weird and confusing

### Additional context

_No response_

calherries:
Fixes #35351, which was caused by [Execute PUT requests in parallel on dashboard save](https://github.com/metabase/metabase/pull/34796#top)

If two revisions are triggered for the same entity around the same time, we have the potential for a race condition where both revisions have `most_recent=true`.

We should aim to avoid creating two revisions for the same entity in parallel anyway, but if it does happen, we should make sure that only one of the revisions has `most_recent=true`.

### Why it fixes the bug
The query that runs when the FE makes the collection items request (`GET /api/collection/root/items?models=dashboard`)  assumes that there is only one revision that has `most_recent=true` for a given instance of a model. That's supposed to be true, but it's not always the case because of this race condition. If there is more than one revision with `most_recent=true` for a model instance, the query will return more than one row for that model instance. See [this code](https://github.com/metabase/metabase/blob/5eaaa17e7b10654c56a5c2f021561955741f7639/src/metabase/api/collection.clj#L415-L418) for the dashboard query join that causes this behaviour.

### DIFF
--- a/src/metabase/models/revision.clj
+++ b/src/metabase/models/revision.clj
@@ -76,8 +76,7 @@
   [revision]
   (assoc revision
          :timestamp :%now
-         :metabase_version config/mb-version-string
-         :most_recent true))
+         :metabase_version config/mb-version-string))
 
 (t2/define-before-update :model/Revision
   [_revision]
@@ -108,12 +107,15 @@
   [revision]
   (u/prog1 revision
     (let [{:keys [id model model_id]} revision]
-      ;; Note 1: Update the last `most_recent revision` to false (not including the current revision)
-      ;; Note 2: We don't allow updating revision but this is a special case, so we by pass the check by
-      ;; updating directly with the table name
-      (t2/update! (t2/table-name :model/Revision)
-                  {:model model :model_id model_id :most_recent true :id [:not= id]}
-                  {:most_recent false})
+      ;; Update the last revision to have most_recent=false, and we update the current revision to most_recent=true.
+      ;; We don't allow updating revision but this is a special case, so we by pass the check by updating directly
+      ;; with the table name. This needs to be atomic otherwise we can end up with multiple revisions with
+      ;; `most_recent ` set to true.
+      (t2/with-transaction [_conn]
+        (t2/update! (t2/table-name :model/Revision)
+                    {:model model :model_id model_id :most_recent true}
+                    {:most_recent false})
+        (t2/update! (t2/table-name :model/Revision) id {:most_recent true}))
       (delete-old-revisions! model model_id))))
 
 ;;; # Functions


tsmacdonald (approved):
Looks reasonable to me :boat:

piranha:
>  boolean `most_recent` column on `revision`

Hmmm, why are we even having that? I.e. isn't it possible just to `order by id desc limit 1`?

calherries:
@piranha I assume the goal is to make certain types of queries faster.

Issue #35374, opened 2023-11-03
[MLv2] [Bug] FE - It's not possible to scroll the FilterPicker


ranquild:
Repro:
- New -> Orders
- Add any custom column
- Filter -> Created At
- The popover is cut off

<img width="753" alt="Screenshot 2023-11-03 at 17 43 21" src="https://github.com/metabase/metabase/assets/8542534/f6ecedbb-6a07-4f2b-aa38-5d48ed16d5df">

PR #35417, opened 2023-11-06
Filter popover overflow issue

ranquild:
Closes https://github.com/metabase/metabase/issues/35374

There is an issue in the `Popover` implementation that cannot be fixed via configuration. They use `floating-ui` under the hood, but don't set `maxHeight` and `maxWidth` as suggested by [the docs](https://floating-ui.com/docs/size#usage). That's why you can get the popover positioned outside of the screen if it's big enough.

I've made a fix to the library and it got approved & merged https://github.com/mantinedev/mantine/pull/5213.

In this PR I make a local override of the affected files for us to be able to ship filters without waiting for library to get published. Once the library is published we can remove copied files and the webpack change.

Before:
Screenshot 2023-11-06 at 14 51 35

After:
Screenshot 2023-11-06 at 14 48 53


webpack.config.js
```diff
@@ -301,9 +301,7 @@ if (WEBPACK_BUNDLE === "hot") {
   };
 
   config.watchOptions = {
-    ignored: [
-      CLJS_SRC_PATH_DEV + "/**",
-    ],
+    ignored: [CLJS_SRC_PATH_DEV + "/**"],
```
ranquild, line 304: prettier

frontend/src/metabase/ui/components/overlays/Popover/Popover.styled.tsx
```diff
@@ -8,10 +8,12 @@ export const getPopoverOverrides = (): MantineThemeOverride["components"] => ({
       shadow: "md",
       zIndex: DEFAULT_Z_INDEX,
       withinPortal: true,
+      middlewares: { shift: true, flip: true, size: true },
```
ranquild, line 11: `size` option is added in the referenced PR

kulyk (approved):
👏

PR #25296, opened 2022-09-07
[Apps] Hide is_write cards from collection items

snoe:
Action cards should only be accessed through action editing - if we want actions to be accessible in collection items then it should be the action itself, so that it will work with HTTPAction as well.

PR #35418, opened 2023-11-06
[MLv2] Refactor `display-info`; add caching

### LINKED ISSUE #35205: breakoutableColumns from MLv2 takes more than 600ms to execute for some questions
### Describe the bug

When opening the summarize sidebar from the query builder for at least some questions, it takes a lot of time to open. I debugged this issue and the root cause is `breakoutableColumns`. We can this function only once when opening the sidebar, there is nothing we can optimize on the FE.

### To Reproduce

1. Go to https://stats.metabase.com/model#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyNiwidHlwZSI6InF1ZXJ5IiwicXVlcnkiOnsib3JkZXItYnkiOltbImRlc2MiLFsiZmllbGQiLCJjb3VudCIseyJiYXNlLXR5cGUiOiJ0eXBlL0JpZ0ludGVnZXIifV1dXSwiZmlsdGVyIjpbIj4iLFsiZmllbGQiLCJjb3VudCIseyJiYXNlLXR5cGUiOiJ0eXBlL0ludGVnZXIifV0sMV0sInNvdXJjZS1xdWVyeSI6eyJhZ2dyZWdhdGlvbiI6W1siY291bnQiXV0sImJyZWFrb3V0IjpbWyJmaWVsZCIsNTI5OTE3LHsiYmFzZS10eXBlIjoidHlwZS9VVUlEIn1dXSwic291cmNlLXRhYmxlIjoiY2FyZF9fNjE4MyIsImZpbHRlciI6WyI9IixbImZpZWxkIiw1Mjk5MzAseyJiYXNlLXR5cGUiOiJ0eXBlL0Jvb2xlYW4ifV0sdHJ1ZV19fX0sImRpc3BsYXkiOiJiYXIiLCJkaXNwbGF5SXNMb2NrZWQiOnRydWUsInBhcmFtZXRlcnMiOltdLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7ImdyYXBoLmRpbWVuc2lvbnMiOlsiY291bnQiXSwiZ3JhcGgubWV0cmljcyI6WyJjb3VudCJdLCJncmFwaC5zZXJpZXNfb3JkZXJfZGltZW5zaW9uIjpudWxsLCJncmFwaC5zZXJpZXNfb3JkZXIiOm51bGx9LCJvcmlnaW5hbF9jYXJkX2lkIjo2MTgzfQ==
2. Click on 'Summarize'
3. See the performance issue

### Expected behavior

It should open instantly

### Logs

_No response_

```JSON
- Chrome 118
```

### Severity

P1. The model has 24 columns, nothing crazy.

### Additional context

_No response_

bshepherdson:
This greatly improves the performance of `ML.display_info`, which is
taking a lot of the cumulative time in the various MLv2 slowdown bugs
recently: #35205, #25296.

### DIFF
--- a/src/metabase/lib/field.cljc
+++ b/src/metabase/lib/field.cljc
@@ -338,8 +338,9 @@
   (cond->> options
     (some #(= (:unit %) unit) options)
     (mapv (fn [option]
-            (cond-> (dissoc option option-key)
-              (= (:unit option) unit) (assoc option-key true))))))
+            (cond-> option
+              (contains? option option-key) (dissoc option option-key)
+              (= (:unit option) unit)       (assoc option-key true))))))
 
 (defmethod lib.temporal-bucket/available-temporal-buckets-method :metadata/column
   [_query _stage-number field-metadata]
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -11,6 +11,7 @@
    [clojure.walk :as walk]
    [goog.object :as gobject]
    [medley.core :as m]
+   [metabase.lib.cache :as lib.cache]
    [metabase.lib.convert :as lib.convert]
    [metabase.lib.core :as lib.core]
    [metabase.lib.equality :as lib.equality]
@@ -25,7 +26,8 @@
    [metabase.mbql.js :as mbql.js]
    [metabase.shared.util.time :as shared.ut]
    [metabase.util :as u]
-   [metabase.util.log :as log]))
+   [metabase.util.log :as log]
+   [metabase.util.memoize :as memoize]))
 
 ;;; this is mostly to ensure all the relevant namespaces with multimethods impls get loaded.
 (comment lib.core/keep-me)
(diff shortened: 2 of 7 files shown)


bshepherdson:
* master
  * **PR #35418** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35418" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35418?utm_source=stack-comment).

src/metabase/lib/field.cljc
```diff
@@ -338,8 +338,9 @@
   (cond->> options
     (some #(= (:unit %) unit) options)
     (mapv (fn [option]
-            (cond-> (dissoc option option-key)
-              (= (:unit option) unit) (assoc option-key true))))))
+            (cond-> option
+              (contains? option option-key) (dissoc option option-key)
+              (= (:unit option) unit)       (assoc option-key true))))))
```
metamben, line 343: Why is this change needed?

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
+;; `"camelCaseStrings"`. Values are recursively converted by [[display-info->js]]. (Note that this passes through the
+;; LRU caches for nested maps and seqs - this is important since many inner pieces are reused across eg. columns.)
+
+;; [[display-info-seq->js]] converts CLJS `sequential?` things to JS arrays, recursively calling [[display-info->js]] on
+;; each element. (This is cached just like map values above.)
+
+;; **Note:** there's an important property here that's worth calling out explicitly. Two columns which are `=` can
+;; be returned by eg. `visible-columns` on two different queries. This is why it's important that the caches used where
+;; `a-query` and `stage-number` are provided ([[display-info]]) are side channel caches that only cache for a specific
+;; instance, not for any `=` value. The different queries might lead to different display names or other details.
```
metamben, line 144: I cannot parse sentence "This is why ..." (which doesn't mean it's wrong).

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
+(declare ^:private display-info->js)
+
+(defn- display-info-map->js* [x]
+  (reduce (fn [obj [cljs-key cljs-val]]
+            (set! js/window.__DIJS_recurses (inc js/window.__DIJS_recurses))
+            (let [js-key (-> cljs-key u/qualified-name u/->camelCaseEn)
+                  js-val (display-info->js cljs-val)] ;; Recursing through the cache
+              (gobject/set obj js-key js-val)
+              obj))
+          (js-obj)
```
metamben, line 155: Isn't `#js {}` the conventional way to write this?

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
+;; **Note:** there's an important property here that's worth calling out explicitly. Two columns which are `=` can
+;; be returned by eg. `visible-columns` on two different queries. This is why it's important that the caches used where
+;; `a-query` and `stage-number` are provided ([[display-info]]) are side channel caches that only cache for a specific
+;; instance, not for any `=` value. The different queries might lead to different display names or other details.
+;; The CLJS -> JS conversion doesn't know about queries, so it uses `=`-based LRU caches.
+(declare ^:private display-info->js)
+
+(defn- display-info-map->js* [x]
+  (reduce (fn [obj [cljs-key cljs-val]]
+            (set! js/window.__DIJS_recurses (inc js/window.__DIJS_recurses))
```
metamben, line 150: Is this for debugging only?

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
   [x]
   (cond
-    (map? x)        (-> x
-                        (update-keys u/->camelCaseEn)
-                        (update-vals display-info-js))
-    (sequential? x) (map display-info-js x)
+    ;; Note that map? is only true for CLJS maps, not JS objects.
+    (map? x)        (display-info-map->js x)
+    ;; Likewise, JS arrays are not sequential? while CLJS vectors, seqs and sets are.
+    (sequential? x) (display-info-seq->js x)
```
metamben, line 175: I wonder if it's worth special casing `{}` and `[]`. They are trivial to convert, so we might gain something if they occur often enough. They might not because of our extensive use of `not-empty`, but might worth checking.

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
+    (keyword? x)    (u/qualified-name x)
```
metamben, line 176: It feels asymmetrical that we don't camelCase here.

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
-      (lib.stage/ensure-previous-stages-have-metadata stage-number)
-      (lib.core/display-info stage-number x)
-      display-info-js
-      (clj->js :keyword-fn u/qualified-name)))
+  ;; Attaches a cached display-info blob to `x`, in case it gets called again for the same object.
+  ;; TODO: Keying by stage is probably unnecessary - if we eg. fetched a column from different stages, it would be a
+  ;; different object. Test that idea and remove the stage from the cache key.
+  (lib.cache/side-channel-cache
+    (keyword "display-info-outer" (str "stage-" stage-number)) x
+    (fn [x] (display-info* a-query stage-number x))))
```
metamben, line 196:
```suggestion
    #(display-info* a-query stage-number %)))
```

src/metabase/lib/js.cljs
```diff
@@ -117,28 +119,81 @@
...
-  (-> a-query
-      (lib.stage/ensure-previous-stages-have-metadata stage-number)
-      (lib.core/display-info stage-number x)
-      display-info-js
-      (clj->js :keyword-fn u/qualified-name)))
+  ;; Attaches a cached display-info blob to `x`, in case it gets called again for the same object.
+  ;; TODO: Keying by stage is probably unnecessary - if we eg. fetched a column from different stages, it would be a
+  ;; different object. Test that idea and remove the stage from the cache key.
+  (lib.cache/side-channel-cache
+    (keyword "display-info-outer" (str "stage-" stage-number)) x
```
metamben, line 195: Is converting to a keyword faster than using a string key? (The question disappears if we can omit stage-number from the key.)

metamben (approved):
LGTM

src/metabase/lib/field.cljc
bshepherdson: It's a leftover and isn't vital in the final version. I was trying to identity-cache various permutations, and found that while the returned buckets were practically identical, the original logic to always `dissoc` meant the bucket/bins returned on columns were all not `identical?` even though `=`. That seemed silly, when they're coming from a static `def` and then maybe one of them gets a `:selected` or `:default` set on it.

src/metabase/lib/js.cljs
bshepherdson: I've reworded it, hopefully it's clearer now. Thanks for pushing back - this is an important but subtle point I'm trying to explain so it's worth polishing the comments.

src/metabase/lib/js.cljs
bshepherdson: Yes, I missed some of that. I've removed it now.

src/metabase/lib/js.cljs
bshepherdson: Perhaps, perhaps not. In any case those two values are interned and already have hash codes computed, so they're fast for the LRU cache to check. I think it's fine to just leave them to be handled by `display-info-*->js`.

src/metabase/lib/js.cljs
bshepherdson: Perhaps, but this is only for the "value" side, not for map keys. This is the original behavior: only map keys get camelcased. We leave keyword values like `:type/Number` alone - actually these values are already mostly camelcased anyway.

src/metabase/lib/js.cljs
bshepherdson: Keywords in CLJS are just `{ns: "foo", key: "bar", hash: 12345}` triples wrapped in a `Keyword` class so it supports the right protocols. they're not interned or anything like that, so they're cheap to wrap.

Also while this is a performance optimization PR, we're not trying to squeeze microseconds. This is about caching 80 calls to a 1-10ms function so it does 12 misses and 68 hits in 75ms rather than 80 straight calls for 730ms.

bshepherdson (commented):
@metamben PTAL; I've answered some questions. Please resolve the threads that you feel are adequately answered. Let me know if that big comment about the caching properties is not clear.

src/metabase/lib/field.cljc
metamben: Interesting, I've checked in a node REPL, and there
```
cljs.user=> (let [m {:x 1}] (identical? m (dissoc m :y)))
true
```

metamben (approved):

bshepherdson:
> [Codenotify](https://github.com/sourcegraph/codenotify): Notifying subscribers in CODENOTIFY files for diff [f0c5dd2...50192a5]

Sorry for the noise folks, I have no idea why CODENOTIFY thought I touched any of those files.

PR #35431, opened 2023-11-06
[Auditv2] Misc Frontend Fixes

iethree:
This test is failing because it's november. Just importing the fix from `master`

probably won't merge this cuz it'll get pulled in with the rebase, but I need something to target to make my open branches green


iethree:
this all got merged in https://github.com/metabase/metabase/pull/35487

PR #34947, opened 2023-10-23
Optimize revision recording, speed up update dashboard API by 25% 🚀

qnkhuat:
An attempt to resolve https://github.com/metabase/metabase/issues/34721 by reducing the number of DB calls when recording revision.

Other changes:
- adding schema for event system in "metabase.events.schema"
- decouple the info in the `event` of `(events/publish-event! topic event)`, before it's used to store everything it's needed for the event to process, it's usually a DB object like Card/Dashboard, and assoced with `actor-id`, `dashcards`... changing this to `{:object dashboard, :other-info 1}` to makes it explicit what's needed. also we can reuse the `object` to record revision so we don't have to refetch it.

Benchmarking ([script](https://gist.github.com/qnkhuat/65d63319fa3d63e52b08c1afe17094f4)):

master
- `GET /api/dashboard/:id` reports 16 DB calls
  -  takes ~14.14ms (N=1000)
- `POST /api/dashboard` reports 23 DB calls
  - takes ~14.7ms (N= 1000)
- `PUT /api/dashboard/:id` reports 22 DB calls
  - takes ~16.19ms (N=1000)
- `PUT /api/dashboard/cards` report 50 DB calls
  - takes ~ 34.6ms (N=1000)

this branch
- `GET /api/dashboard/:id` reports 16 DB calls
  -  takes ~12ms (N=1000)
- `POST /api/dashboard/:id` reports 20 DB calls
  - takes ~11.4ms (N= 1000)
- `PUT /api/dashboard/:id` reports 25 DB calls
  - takes ~14.2ms (N=1000)
- `PUT /api/dashboard/cards` report 39 DB calls
  - takes ~ 25.4ms (N=1000)

PR #35435, opened 2023-11-06
Fix audit v2 merge conflicts with master and rework audit_log API

noahmoss:
**Problem**

There were a great deal of merge conflicts with the `auditv2-main` feature branch, primarily because of [this PR](https://github.com/metabase/metabase/pull/34947). This PR was mainly working on the revision subsystem, not audit, but it changed the `publish-event!` API to expect a map of this structure:
```
{:object object, :user-id user-id, :other-keys ...}
```
rather than just an instance of a model itself (like a card, dashboard, etc). It also enforced this structure with new malli schemas for all the existing event types.

I like this approach, but it required adapting a lot of our `audit_log` code to use this new structure. In addition, since that PR was pretty large, it conflicted with audit v2 changes in other assorted ways which broke a number of tests (over 400 backend tests were failing once I finished the initial rebase).

**Approach to fixing this**

1. I've created a new branch `auditv2-main-2` which has a PR open here: https://github.com/metabase/metabase/pull/35434 This branch is just `auditv2-main` rebased on master, with a best-effort attempt to fix code conflicts. This doesn't do any larger architecture changes or test fixes, for the most part.
2. I've opened this PR with my changes to the way audit is structured, targeting that branch. This way, these changes can be reviewed in a slightly more visible way, instead of them just being hidden by a giant rebase. Once this PR is green I'll merge it into `auditv2-main-2` and we can use that as the authoritative feature branch.

**New things in this PR**
- `publish-event!` calls have been changed to take `:object` and `:user-id` keys.
- I restructured the `audit-log/record-event!` API to make it much more flexible. Now it's just a two-argument function that takes a `params` map, and most of the event handlers in `metabase.events.audit-log` can pass their data straight through to it. See the docstring for that function for more details (I've left a comment highlighting where it is)
- Some misc test fixes and renamings. Will self-review to shine light on some interesting areas.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/audit_app/pages/dashboard_subscriptions_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/audit_app/pages/dashboard_subscriptions_test.clj
@@ -35,7 +35,7 @@
                                                                     :collection_id collection-id}
                           PulseChannel          {channel-id :id}   {:pulse_id       pulse-id
                                                                     :channel_type   "email"
-                                                                    :details        {:emails ["amazing@fake.com"]}
+                                                                    :details        {:emails ["amazing@metabase.com"]}
                                                                     :schedule_type  "monthly"
                                                                     :schedule_frame "first"
                                                                     :schedule_day   "mon"
--- a/enterprise/backend/test/metabase_enterprise/audit_app/pages/dashboard_subscriptions_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/audit_app/pages/dashboard_subscriptions_test.clj
@@ -19,69 +19,70 @@
         :args [dashboard-name]}))))
 
 (deftest table-test
-  (is (= []
-         (mt/rows (dashboard-subscriptions (mt/random-name)))))
-  (let [dashboard-name (mt/random-name)]
-    (mt/with-temp! [Collection {collection-id :id, collection-name :name}]
-      ;; test with both the Root Collection and a non-Root Collection
-      (doseq [{:keys [collection-id collection-name]} [{:collection-id   collection-id
-                                                        :collection-name collection-name}
-                                                       {:collection-id   nil
-                                                        :collection-name "Our analytics"}]]
-        (testing (format "Collection = %d %s" collection-id collection-name)
-          (mt/with-temp! [Dashboard             {dashboard-id :id} {:name          dashboard-name
-                                                                    :collection_id collection-id}
-                          Pulse                 {pulse-id :id}     {:dashboard_id  dashboard-id
-                                                                    :collection_id collection-id}
-                          PulseChannel          {channel-id :id}   {:pulse_id       pulse-id
-                                                                    :channel_type   "email"
-                                                                    :details        {:emails ["amazing@metabase.com"]}
-                                                                    :schedule_type  "monthly"
-                                                                    :schedule_frame "first"
-                                                                    :schedule_day   "mon"
-                                                                    :schedule_hour  8}
-                          PulseChannelRecipient _                  {:pulse_channel_id channel-id
-                                                                    :user_id          (mt/user->id :rasta)}
-                          PulseChannel          {channel-2-id :id} {:pulse_id      pulse-id
-                                                                    :channel_type  "slack"
-                                                                    :details       {:channel "#wow"}
-                                                                    :schedule_type "hourly"}]
-            (is (= {:columns ["dashboard_id"
-                              "dashboard_name"
-                              "pulse_id"
-                              "recipients"
-                              "subscription_type"
-                              "collection_id"
-                              "collection_name"
-                              "frequency"
-                              "creator_id"
-                              "creator_name"
-                              "created_at"
-                              "num_filters"]
-                    ;; sort by newest first.
-                    :rows    [[dashboard-id
-                               dashboard-name
-                               pulse-id
-                               nil
-                               "Slack"
-                               collection-id
-                               collection-name
-                               "Every hour"
-                               (mt/user->id :rasta)
-                               "Rasta Toucan"
-                               (t2/select-one-fn :created_at PulseChannel :id channel-2-id)
-                               0]
-                              [dashboard-id
-                               dashboard-name
-                               pulse-id
-                               2
-                               "Email"
-                               collection-id
-                               collection-name
-                               "At 8:00 AM, on the first Tuesday of the month"
-                               (mt/user->id :rasta)
-                               "Rasta Toucan"
-                               (t2/select-one-fn :created_at PulseChannel :id channel-id)
-                               0]]}
-                   (mt/rows+column-names
-                    (dashboard-subscriptions (str/join (rest (butlast dashboard-name)))))))))))))
+  (premium-features-test/with-premium-features #{}
+   (is (= []
+          (mt/rows (dashboard-subscriptions (mt/random-name)))))
+   (let [dashboard-name (mt/random-name)]
+     (mt/with-temp! [Collection {collection-id :id, collection-name :name}]
+       ;; test with both the Root Collection and a non-Root Collection
+       (doseq [{:keys [collection-id collection-name]} [{:collection-id   collection-id
+                                                         :collection-name collection-name}
+                                                        {:collection-id   nil
+                                                         :collection-name "Our analytics"}]]
+         (testing (format "Collection = %d %s" collection-id collection-name)
+           (mt/with-temp! [Dashboard             {dashboard-id :id} {:name          dashboard-name
+                                                                     :collection_id collection-id}
+                           Pulse                 {pulse-id :id}     {:dashboard_id  dashboard-id
+                                                                     :collection_id collection-id}
+                           PulseChannel          {channel-id :id}   {:pulse_id       pulse-id
+                                                                     :channel_type   "email"
+                                                                     :details        {:emails ["amazing@fake.com"]}
+                                                                     :schedule_type  "monthly"
+                                                                     :schedule_frame "first"
+                                                                     :schedule_day   "mon"
+                                                                     :schedule_hour  8}
+                           PulseChannelRecipient _                  {:pulse_channel_id channel-id
+                                                                     :user_id          (mt/user->id :rasta)}
+                           PulseChannel          {channel-2-id :id} {:pulse_id      pulse-id
+                                                                     :channel_type  "slack"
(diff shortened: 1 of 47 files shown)


src/metabase/api/dashboard.clj
```diff
@@ -379,49 +379,6 @@
     (validation/check-embedding-enabled)
     (api/check-superuser)))
 
-(api/defendpoint PUT "/:id"
```
noahmoss, line 382: This was moved by a PR on master to elsewhere in the file but got re-added during the rebase, so I'm removing it again here.

resources/migrations/000_migrations.yaml
```diff
@@ -15905,6 +15905,40 @@ databaseChangeLog:
...
+            constraintName: fk_recent_views_ref_user_id
+            referencedTableName: core_user
+            baseColumnNames: user_id
+            referencedColumnNames: id
+            onDelete: CASCADE
+
+  - changeSet:
+      id: v48.00-048
+      author: noahmoss
+      comment: Add foreign key on recent_views with onDelete CASCADE
```
noahmoss, line 15928: Did this to fix a failing test which couldn't delete a user because it had entries in `recent_views`. It's not strictly necessary in the app itself though since users are never deleted, just deactivated.

enterprise/backend/test/metabase_enterprise/sso/integrations/saml_test.clj
```diff
@@ -472,7 +472,7 @@
                         :model_id (:id new-user)
                         :topic    :user-invited
                         :user_id  nil}
-                       (audit-log-test/event :user-invited (:id new-user))))))
+                       (audit-log-test/latest-event :user-invited (:id new-user))))))
```
noahmoss, line 475: Renamed this function because it was shadowing another `event` variable in a test (I forget where now)

src/metabase/events/audit_log.clj
```diff
@@ -1,100 +1,93 @@
...
-   [metabase.api.common :as api]
    [metabase.events :as events]
    [metabase.models.audit-log :as audit-log]
+   [metabase.util :as u]
    [methodical.core :as methodical]
    [toucan2.core :as t2]))
 
 (derive ::event :metabase/event)
 
-(defn maybe-prepare-update-event-data
```
noahmoss, line 13: Moved this logic into `metabase.models.audit-log` to make it a bit more flexible: now the `record-event!` params map can take `object` and `previous-object`, and it'll automatically record the changes if there are any

src/metabase/events/schema.clj
```diff
@@ -87,15 +87,11 @@
...
                       [:map {:closed true}
-                       [:object [:fn #(t2/instance-of? :model/Database %)]]])
-      with-user      (mc/schema
-                      [:merge default-schema
-                       [:map {:closed true}
-                        [:user-id  pos-int?]]])]
-
+                       [:object [:fn #(t2/instance-of? :model/Database %)]]
+                       [:previous-object {:optional true} [:fn #(t2/instance-of? :model/Database %)]]
+                       [:user-id pos-int?]])]
```
noahmoss, line 92: Tweaked this schema so that all DB events can take `user-id`

src/metabase/models/audit_log.clj
```diff
@@ -31,51 +34,73 @@
...
-                 :topic    unqualified-topic
-                 :details  details
-                 :model    model-name
-                 :model_id model-id
-                 :user_id  user-id)
-     ;; TODO: temporarily double-writing to the `activity` table, delete this in Metabase v48
-     (when-not (#{:card-read :dashboard-read :table-read :card-query :setting-update} unqualified-topic)
+  `:object` and `:previous-object` both have `model-details` called on them to determine which fields should be audited,
+  then they are added to `:details` before the event is recorded. `:previous-object` is only included if any audited fields
+  were updated."
```
noahmoss, line 66: Here's the biggest change of the file. Reworked this function to have a cleaner interface, with a `params` map of entirely optional keys.

enterprise/backend/test/metabase_enterprise/audit_app/pages/dashboard_subscriptions_test.clj
```diff
@@ -35,7 +35,7 @@
                                                                     :collection_id collection-id}
                           PulseChannel          {channel-id :id}   {:pulse_id       pulse-id
                                                                     :channel_type   "email"
-                                                                    :details        {:emails ["amazing@fake.com"]}
+                                                                    :details        {:emails ["amazing@metabase.com"]}
```
johnswanson, line 38: Just curious, this change is unrelated right? And you're changing it because we want to use domains we control to prevent accidents (if e.g. these tests were run in a context where we could actually send out emails)?

test/metabase/query_processor/middleware/process_userland_query_test.clj
```diff
@@ -124,35 +124,32 @@
...
+                                  (:dataset_query card)
+                                  :info {:card-id (:id card)
+                                         :executed-by (mt/user->id :rasta)}))
 
-        (is (true? (t2/exists? :model/ViewLog :model "card" :model_id (:id card) :user_id (mt/user->id :rasta)))))))
+      (is (true? (t2/exists? :model/ViewLog :model "card" :model_id (:id card) :user_id (mt/user->id :rasta))))))
 
   (testing "skip if context is dashboard or collection"
     (mt/with-temp [:model/Card card {:dataset_query (mt/mbql-query users)}]
-      (mt/with-model-cleanup [:model/ViewLog]
```
johnswanson, line 146: Out of curiosity, why remove the `with-model-cleanup` here?

johnswanson (approved):
I like the API change! I had a couple questions but they're just to satisfy my curiosity.

src/metabase/events/audit_log.clj
```diff
@@ -1,100 +1,93 @@
...
 (derive :event/dashboard-remove-cards ::dashboard-card-event)
 
-;; TODO (crumb)
 (methodical/defmethod events/publish-event! ::dashboard-card-event
-  [topic {:keys [dashcards id]}]
+  [topic {:keys [object dashcards user-id] :as _event}]
   ;; we expect that the object has just a dashboard :id at the top level
   ;; plus a `:dashcards` attribute which is a vector of the cards added/removed
-  (let [details (-> (t2/select-one [:model/Dashboard :description :name] :id id)
+  (let [details (-> (select-keys object [:description :name :id])
```
dpsutton, line 55: I just noticed we have an N+1 query situation here. (`t2/select` N times rather than something like

```clojure
(methodical/defmethod events/publish-event! ::dashboard-card-event
  [topic {:keys [object dashcards user-id] :as _event}]
  ;; we expect that the object has just a dashboard :id at the top level
  ;; plus a `:dashcards` attribute which is a vector of the cards added/removed
  (let [cards   (when (seq dashcards)
                  (t2/select-fn->fn :id #(select-keys % [:name :description])
                                    :model/Card
                                    :id [:in (map :card_id dashcards)]))
        details (-> (select-keys object [:description :name :id])
                    (assoc :dashcards (for [{:keys [id card_id]} dashcards]
                                        (-> (cards card_id)
                                            (assoc :id id)
                                            (assoc :card_id card_id)))))]
    (audit-log/record-event! topic
                               {:details  details
                                :user-id  user-id
                                :model    :model/Dashboard
                                :model-id (u/id object)})))
```

enterprise/backend/test/metabase_enterprise/audit_app/pages/dashboard_subscriptions_test.clj
noahmoss: Actually, this test was failing for me in the REPL because of the `subscription-allowed-domains` setting in `metabase-enterprise.advanced-config.models.pulse-channel`. It was failing because locally I had an enterprise token set with all of the features enabled, but in CI I guess this test is run without the `:email-allow-list` feature, so the `fake.com` domain doesn't error. 

I thought that `metabase.com` was the default value I think that's actually just set by a different test. A better fix would be to just clear all premium features for the duration of this test, which I actually did in some other tests in this PR as well. (i.e. `(premium-features-test/with-premium-features #{} ...)`) I'll change this to use that approach as well.

test/metabase/query_processor/middleware/process_userland_query_test.clj
noahmoss: I think in general, using `with-model-cleanup` for AuditLog and ViewLog doesn't make too much sense. Lots of other tests are writing to these tables implicitly by triggering events, so the tables will be filled with "fake" data anyway. So cleaning up these models just adds work to the tests without any benefit.

src/metabase/events/audit_log.clj
noahmoss: Good call, will update

qwef (approved):

qwef (approved):

src/metabase/models/audit_log.clj
```diff
@@ -31,51 +34,73 @@
...
-       :topic      topic
-       :object     object
-       :details-fn (fn [_] details)
-       :user-id    user-id)))))
+       {:topic    topic
+        :object   object
+        :details  details
+        :model    model
+        :model-id model-id
+        :user-id  user-id}))))
```
escherize, line 103: I'd like to propose separating the generation of the value on 98:103 from actually doing the side effect. 

So, instead of:

```clojure
(record-event! topic params)
```

we would call:

```clojure
(actually-record-the-event! 
  (generate-event topic params))
```

----

## Why? 

It makes it easier to see and test what the outputs for record-object! are. Yes, now we can use the database to see, but it would makes it easier to see them in the repl, too. The tests that check it could be run in parallel, and would be way faster, because they don't rely on the db.

It feels like such a nice spot to separate the imperative shell from the complicated pure logic-y bit.

wdyt?

escherize (commented):
Looks kinda familiar in a way, but also looks good to me.

Question: There are other handlers for `publish-event!`, do you know if changing its API will make those break? I assume if the tests are passing it should be OK. But I wanted to bring it up.

src/metabase/models/audit_log.clj
escherize: In fact we could have a long test using e.g. `are` with 50 inputs/outputs and a simple `=` check for `generate-event` that is simple, and easy to follow (v.s. doing database setup/teardown/etc. for each one).

escherize (approved):

PR #35436, opened 2023-11-06
[MLv2] Put duplicate implicit join columns in separate column-groups

### LINKED ISSUE #34742: [MLv2] - A question joined on itself can cause FK columns to be displayed twice
Steps to reproduce: 
- Create a question and save it (orders table, for example). We will call this Question A.
- Create a new question, with a source of Question A, and join Question A on the ID column

`lib.visibleColumns` returns any columns accessible by foreign keys twice (sometimes).

Loom with more details [here](https://www.loom.com/share/dc79afe8449d4d2d88011dea8eb8ce43)

snoe:
Fixes #34742

Card test utils would `dissoc` `:table-id` and `:id` from `results_metadata` but in practice those fields are often there, and for implicit columns they are needed or else implicit join columns throws a validation error.


test/metabase/lib/card_test.cljc
```diff
@@ -103,12 +103,10 @@
                          {:cards [(assoc (:orders lib.tu/mock-cards) :dataset-query lib.tu/venues-query)]})
                         (:orders lib.tu/mock-cards))]
       (is (=? (->> (cols-of :orders)
-                   (map #(dissoc % :id :table-id))
                    sort-cols)
               (sort-cols (get-in lib.tu/mock-cards [:orders :result-metadata]))))
 
-      (is (=? (->> (concat (from :source/card (for [c (cols-of :orders)]
-                                                (dissoc c :id :table-id)))
```
metamben, line 111: I'm not sure we don't want to test with such data.

test/metabase/lib/metadata/calculation_test.cljc
```diff
@@ -129,8 +129,7 @@
                                  :source-card (:id (lib.tu/mock-cards :orders))}]}
           own-fields (for [field (lib.metadata/fields lib.tu/metadata-provider-with-mock-cards (meta/id :orders))]
                        (-> field
-                           (assoc :lib/source :source/card)
-                           (dissoc :id :table-id)))]
```
metamben, line 133: I'm assuming this is intended to simulate Real Life, so I'm unsure we can just remove this. But I've found your Slack message, and now I tend to think that we should test both possibilities. If there are no `id`s, the names should provide unique references, I guess.

metamben (commented):
Makes sense to me, just the situation with `id`s missing or not needs to be clarified.

test/metabase/lib/metadata/calculation_test.cljc
snoe: Added tests where id and table-id are not present

metamben (approved):

PR #35439, opened 2023-11-07
Fixing dropdown filter value GET #32126

### LINKED ISSUE #32126: Filter get cached and won't be refreshed after populating metadata in model
### Describe the bug

1) Create a model with SQL question (select * from accounts), save it.
2) Create a question from this model
3) create a dashboard
4) Add the question to a dashboard
5) Create 2 filters: one connected to email, another to source
7) Now click on both fields to get the values (you won't get any values)

8) now go to the model you created in 1: map email to accounts->email, source to accounts->source and plan to accounts->plan. Save it
9) Go back to the dashboard, click on the filters. You won't get any values
10) try disconnecting the filter from the cards, saving and then connecting again the filters. You won't get any values
11) try removing the filters and adding + connecting them. You won't get any values.

This dashboard is busted now, you need to build it again

### To Reproduce

See above

### Expected behavior

When you populate metadata on models, filters that use the fields should populate

### Logs

_No response_

```JSON
v47.x, but pretty sure this is comes from many versions back
```

### Severity

P1?

### Additional context

While fixing this, please test:
1) signed embedding
2) linked and not linked filters
3) sandboxing
4) impersonation

:)

markbastian:
When GET "/:id/params/:param-key/values" is invoked, the endpoint looks for valid values to populate for the filter dropdowns using the resolved field-ids for the underlying item in the dashboard. In some cases, such as nested queries, the field is not referenced by id, but using the field name.

This PR adds a branch that will look for fields by field ref (e.g. `[:field "SOURCE" {:base-type :type/Text}]`) if the field cannot be resolved by id.

Fixes #32126


src/metabase/api/dashboard.clj
```diff
@@ -820,6 +823,26 @@
              field             (param->fields param)]
          (assoc field :value value))))
 
+(defn filter-values-from-field-refs
+  "Get filter values when only field-refs (e.g. `[:field \"SOURCE\" {:base-type :type/Text}]`)
+  are provided (rather than field-ids). This is a common case for nested queries."
+  [dashboard param-key]
+  (let [dashboard       (t2/hydrate dashboard :resolved-params)
```
adam-james-v, line 830: I think you don't need to hydrate the dashboard here, since it's passed in hydrated already.

adam-james-v (approved):
Looks good!

PR #34065, opened 2023-09-22
Stop publishing elastic beanstalk artifacts

iethree:
Closes https://github.com/metabase/metabase/issues/31151

### Description

Metabase 47.2 will be the last version for which we publish elastic beanstalk artifacts. Our [documentation has already been updated](https://www.metabase.com/docs/v0.47/installation-and-operation/running-metabase-on-elastic-beanstalk) to reflect that we do not recommend deploying Metabase on EBS any longer. Instructions for elastic beanstalk setup for older metabase versions are [still available](https://www.metabase.com/docs/v0.46/installation-and-operation/running-metabase-on-elastic-beanstalk) in older versions of the docs.

PR #35440, opened 2023-11-07
Offline Releases

iethree:
Closes https://github.com/metabase/metabase-private/issues/169

### Description

In the (not entirely unlikely) event that github actions are unavailable when we need to release, this enables "offline" releases using mostly the same code as our releases in github actions.  This feels a little dumb because it basically replicates the behavior of the old clojure release code in javascript. However, it has a few advantages:

1. This script actually contains relatively little of the release logic, it basically replicates the behavior of our github actions workflows - namely calling reusable javascript functions for the real release logic.
2. This script is a bit more modular than the clojure one - it _requires_ you to build separately from publishing so that you can actually test the same jar that you publish.
3. The script also lets you run any publish step independently, so, for instance, if the github actions workflow blows up at the release notes generation step, you can simply run that step locally in isolation, without having to re-run the whole build end-to-end.

Once this merges, we can delete most of the contents of `bin/build/src/release`: https://github.com/metabase/metabase-private/issues/175

### How to verify

Run a build locally in your fork? 🙃   I'll upload a video demo. 😀 

### Demo

Screen Shot 2023-11-06 at 5 09 59 PM

https://www.loom.com/share/a56f5a6904ff4f48acaa021846c90aeb


release/README.md
```diff
@@ -15,8 +15,26 @@ There are 3 ways to release
...
+> _Note_
+> Use this only if for some reason the automated release in CI is not available.
+
+1. Copy the `.env-template` file to `.env` and fill in all environment variables
+2. Run `cd release && yarn && yarn release-offline v0.77.77 1234567890abcdef1234567890abcdef12345678 --build`
+3. Wait for the build and test jobs to finish
+4. Test the built jar locally: `MB_JETTY_PORT=3033 java -jar target/uberjar/metabase.jar`
+5. If everything is successful, run `yarn release-offline v0.77.77 1234567890abcdef1234567890abcdef12345678 --release`
+
+You can also run publish steps in isolation using these flags instead of `--publish`
```
nemanjaglumac, line 27: `--publish` or `--release`?

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+  DOCKERHUB_RELEASE_USERNAME,
+  DOCKERHUB_RELEASE_TOKEN,
+} = process.env;
+
+const github = new Octokit({ auth: GITHUB_TOKEN });
+
+const JAR_PATH = "../target/uberjar";
+
+const version = process.argv?.[2]?.trim();
+const commit = process.argv?.[3]?.trim();
```
kamilmielnik, line 48: To match the `isValidCommitHash` argument name.

```suggestion
const commitHash = process.argv?.[3]?.trim();
```

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+const step = process.argv?.[4]?.trim().replace("--", "");
+const edition = isEnterpriseVersion(version) ? "ee" : "oss";
+
+if (!isValidVersionString(version)) {
+  error(
+    "You must provide a valid version string as the first argument (e.g v0.45.6",
```
kamilmielnik, line 54:
```suggestion
    "You must provide a valid version string as the first argument (e.g v0.45.6)",
```

release/README.md
```diff
@@ -15,8 +15,26 @@ There are 3 ways to release
...
 
-Coming soon ...
+> _Note_
+> Use this only if for some reason the automated release in CI is not available.
+
+1. Copy the `.env-template` file to `.env` and fill in all environment variables
+2. Run `cd release && yarn && yarn release-offline v0.77.77 1234567890abcdef1234567890abcdef12345678 --build`
+3. Wait for the build and test jobs to finish
+4. Test the built jar locally: `MB_JETTY_PORT=3033 java -jar target/uberjar/metabase.jar`
+5. If everything is successful, run `yarn release-offline v0.77.77 1234567890abcdef1234567890abcdef12345678 --release`
```
nemanjaglumac, line 25:
```suggestion
5. If everything is successful, run `yarn release-offline v0.77.77 1234567890abcdef1234567890abcdef12345678 --publish`
```

After watching the video I think this part was a typo.

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+  ).toString();
+
+  log(versionProperties, "green");
+
+  if (!versionProperties.includes(`tag=${version}`)) {
+    error(`This jar does not match the input version: ${version}`);
+  }
+}
+
+// eslint-disable-next-line no-console
```
kamilmielnik, line 120: This comment needs to be in the next line

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+    error(
+      `Java version must be ${versionRequirements[majorVersion].java}, you are running ${javaVersion}`,
+    );
+  }
+
+  const currentBranch = (await $`git branch --show-current`).toString().trim();
+
+  try {
+    await $`git fetch --all`;
+    await $`git stash && git checkout ${commit}`;
```
kamilmielnik, line 160: Instead of stashing, can we throw an error early on in the script if `git status` tells us there are any changes?

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+if (!step) {
+  error("You must provide a step argument like --build or --publish");
+}
+
+/**************************************************
+          HELPERS
+ **************************************************/
+
+// mostly for type checking
+function getGithubCredentials() {
```
kamilmielnik, line 71: I like it :+1: 
Can we have equivalent functions for AWS & docker env vars?

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+    !AWS_S3_DOWNLOADS_BUCKET ||
+    !AWS_S3_STATIC_BUCKET ||
+    !AWS_S3_RELEASE_SECRET_ACCESS_KEY ||
+    !AWS_S3_RELEASE_ACCESS_KEY_ID
+  ) {
+    error("You must provide all AWS environment variables in .env-template");
+  }
+
+  process.env.AWS_ACCESS_KEY_ID = AWS_S3_RELEASE_ACCESS_KEY_ID;
+  process.env.AWS_SECRET_ACCESS_KEY = AWS_S3_RELEASE_SECRET_ACCESS_KEY;
```
kamilmielnik, line 207: What's the purpose of aliasing those?

Can we just have `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` in `.env-template`?

release/release-offline.ts
```diff
@@ -0,0 +1,410 @@
...
+  const versionInfoName =
+    edition === "ee" ? "version-info-ee.json" : "version-info.json";
+
+  fs.writeFileSync(versionInfoName, JSON.stringify(newVersionInfo, null, 2));
+
+  await $`aws s3 cp ${versionInfoName} s3://${AWS_S3_STATIC_BUCKET}/${versionInfoName}`.pipe(
+    process.stdout,
+  );
+
+  if (!process.env.SKIP_CLOUDFRONT) {
```
kamilmielnik, line 294: `SKIP_CLOUDFRONT` is the only `process.env` that we don't destructure at the top of the file.
Also it's missing from `.env-template`.

Why would we want to `SKIP_CLOUDFRONT` anyway?

release/README.md
```diff
@@ -15,8 +15,26 @@ There are 3 ways to release
 
 ## Github action-less release
 
-Coming soon ...
+> _Note_
```
kamilmielnik, line 18: It'd be great to have the link to your loom recording here somewhere.

kamilmielnik (commented):
Looks good to me. I'll let others approve though.

> "I hope we don't have to use it but it's a good thing to have" 

:smile:

release/README.md
iethree: 🤦‍♂️ yep, should be `--publish`

release/release-offline.ts
iethree: 👍

Screen Shot 2023-11-09 at 4 23 18 PM

release/release-offline.ts
iethree: this was convenient for testing because I don't have a parallel cloudfront environment setup. will remove

release/release-offline.ts
iethree: prettier did me dirty on this one

kamilmielnik (approved):
Looks great! :rocket: 

----

The [original issue](https://github.com/metabase/metabase-private/issues/169) has 1 unchecked item - is this ok?

> publish beanstalk artifacts

iethree:
> Looks good to me. I'll let others approve though.
> 
> > "I hope we don't have to use it but it's a good thing to have"
> 
> 😄

Good catch! I actually took care of that here: https://github.com/metabase/metabase/pull/34065

release/.env-template
```diff
@@ -1,24 +1,18 @@
...
 DOCKERHUB_REPO=metabase
 DOCKERHUB_OWNER=
-DOCKERHUB_STAGING_REPO=mb-staging
 DOCKERHUB_RELEASE_USERNAME=
 DOCKERHUB_RELEASE_TOKEN=
 
-AWS_S3_RELEASE_ACCESS_KEY_ID=
-AWS_S3_RELEASE_SECRET_ACCESS_KEY=
+AWS_ACCESS_KEY_ID=
+AWS_SECRET_ACCESS_KEY=
```
calherries, line 13: Can you explain why you renamed these? And we should remember to update them in the 1password notes

PR #35443, opened 2023-11-07
Daily task to delete old inactive tables

### LINKED ISSUE #30425: Inactive tables (not found in sync) for long periods should be deleted from the appdb.
Some instances have thousands or even millions of tables. Most of them are inactive (old databases, deleted in the source) and will never come back. These entries (tables, fields, field values) can increase a lot the size of the AppDB, degrading performance and operations (backup times).

After 180 days of being inactive (not found on sync), we should delete the table and all derived fields/field values.
This number should be configurable using an env var: `MB_INACTIVE_TABLE_MAX_DAYS`

Manually delete the table/field/field values from the DB.

It degrades performance of large instances.

**Additional context**
We need to be aware of what happens to questions that query those fields. When the table/field is deleted, these MBQL queries will never point to the same fields.
The expected behavior in this case is to clear the query definition if it contains fields that don't exist anymore.

[Slack thread](https://metaboat.slack.com/archives/C010ZSXQY87/p1682703576143019?thread_ts=1682528038.247339&cid=C010ZSXQY87)

qnkhuat:
Closes https://github.com/metabase/metabase/issues/30425

### DIFF
--- a/docs/configuring-metabase/environment-variables.md
+++ b/docs/configuring-metabase/environment-variables.md
@@ -197,6 +197,16 @@ Twice a day, Metabase will delete rows older than this threshold.
 
 The minimum value is `30` days (Metabase will treat entered values of `1` to `29` the same as `30`). If set to `0`, Metabase will keep all rows.
 
+## MB_INACTIVE_TABLE_MAX_DAYS
+
+Type: integer<br>
+Default: `180`<br>
+Since: v48.0
+
+Sets the maximum number of days Metabase keep inactive tables.
+
+Once a day, Metabase will delete inactive tables older than this threshold
+
 ### `MB_BREAKOUT_BIN_WIDTH`
 
 Type: double<br>
--- a/docs/configuring-metabase/environment-variables.md
+++ b/docs/configuring-metabase/environment-variables.md
@@ -197,16 +197,6 @@ Twice a day, Metabase will delete rows older than this threshold.
 
 The minimum value is `30` days (Metabase will treat entered values of `1` to `29` the same as `30`). If set to `0`, Metabase will keep all rows.
 
-## MB_INACTIVE_TABLE_MAX_DAYS
-
-Type: integer<br>
-Default: `180`<br>
-Since: v48.0
-
-Sets the maximum number of days Metabase keep inactive tables.
-
-Once a day, Metabase will delete inactive tables older than this threshold
-
 ### `MB_BREAKOUT_BIN_WIDTH`
 
 Type: double<br>
@@ -529,6 +519,16 @@ Type: string<br>
 Default: `null`
 
 Client ID for Google Auth SSO. If this is set, Google Auth is considered to be enabled.
+
+## MB_INACTIVE_TABLE_MAX_DAYS
+
+Type: integer<br>
+Default: `180`<br>
+Since: v48.0
+
+Sets the maximum number of days Metabase keep inactive tables.
+
+Once a day, Metabase will delete inactive tables older than this threshold.
 
 ### `MB_JDBC_DATA_WAREHOUSE_MAX_CONNECTION_POOL_SIZE`
 
--- a/src/metabase/models/table.clj
+++ b/src/metabase/models/table.clj
@@ -62,7 +62,6 @@
                   :field_order  (driver/default-field-order (t2/select-one-fn :engine Database :id (:db_id table)))}]
     (merge defaults table)))
 
-;; TODO: should we do this after the deletion?
 (t2/define-before-delete :model/Table
   [{:keys [db_id schema id]}]
   (t2/delete! Permissions :object [:like (str "%" (perms/data-perms-path db_id schema id) "%")]))
--- a/src/metabase/models/table.clj
+++ b/src/metabase/models/table.clj
@@ -62,9 +62,10 @@
                   :field_order  (driver/default-field-order (t2/select-one-fn :engine Database :id (:db_id table)))}]
     (merge defaults table)))
 
+;; TODO: should we do this after the deletion?
 (t2/define-before-delete :model/Table
   [{:keys [db_id schema id]}]
-  (t2/delete! Permissions :object [:like (str (perms/data-perms-path db_id schema id) "%")]))
+  (t2/delete! Permissions :object [:like (str "%" (perms/data-perms-path db_id schema id) "%")]))
 
 (defmethod mi/perms-objects-set :model/Table
   [{db-id :db_id, schema :schema, table-id :id, :as table} read-or-write]
--- a/src/metabase/task/delete_inactive_tables.clj
+++ b/src/metabase/task/delete_inactive_tables.clj
@@ -42,5 +42,5 @@
                  (triggers/start-now)
                  (triggers/with-schedule
                    ;; run every day at 1 AM
-                   (cron/cron-schedule "0 0 1 * * ? *")))]
+                   (cron/cron-schedule "0 30 1 * * ? *")))]
     (task/schedule-task! job trigger)))
--- a/src/metabase/task/delete_inactive_tables.clj
+++ b/src/metabase/task/delete_inactive_tables.clj
@@ -5,7 +5,6 @@
    [clojurewerkz.quartzite.triggers :as triggers]
    [java-time :as t]
    [metabase.config :as config]
-   [metabase.models.permissions :as perms]
    [metabase.models.task-history :as task-history]
    [metabase.task :as task]
    [metabase.util.i18n :refer [trs]]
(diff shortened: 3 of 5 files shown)


docs/configuring-metabase/environment-variables.md
```diff
@@ -197,6 +197,16 @@ Twice a day, Metabase will delete rows older than this threshold.
...
 
+## MB_INACTIVE_TABLE_MAX_DAYS
+
+Type: integer<br>
+Default: `180`<br>
+Since: v48.0
+
+Sets the maximum number of days Metabase keep inactive tables.
+
+Once a day, Metabase will delete inactive tables older than this threshold
```
tsmacdonald, line 208:
```suggestion
Once a day, Metabase will delete inactive tables older than this threshold.
```

src/metabase/task/delete_inactive_tables.clj
```diff
@@ -0,0 +1,54 @@
...
+(def ^:private trigger-key "metabase.task.delete-inactive-tables.trigger")
+
+(defn- delete-inactive-tables!
+  []
+  (log/debug (trs "Cleaning up Inactive Tables older than {0} days" inactive-table-max-days))
+  (task-history/with-task-history {:task "delete-inactive-tables"}
+    (t2/with-transaction [_conn]
+      (let [to-delete-tables (t2/select [:model/Table :id :db_id :schema]
+                                        :active false
+                                        :updated_at [:< (t/minus (t/zoned-date-time) (t/seconds inactive-table-max-days))])]
```
tsmacdonald, line 31:
```suggestion
                                        :active     false
                                        :updated_at [:< (t/minus (t/zoned-date-time)
                                                                                    (t/days inactive-table-max-days))])]
```

The whitespace is me being fussy, but the `seconds` -> `days` seems like an essential bug?


```
just-now
#t "2023-11-07T12:44:42.878378088Z[UTC]"

> (t/minus just-now (t/seconds 180))
#t "2023-11-07T12:41:42.878378088Z[UTC]"

> (t/minus just-now (t/days 180))
#t "2023-05-11T12:44:42.878378088Z[UTC]"
```

src/metabase/task/delete_inactive_tables.clj
```diff
@@ -0,0 +1,54 @@
...
+        (when (seq to-delete-tables)
+          (t2/delete! :model/Table :id [:in (map :id to-delete-tables)])
+          (t2/delete! :model/Permissions
+                      {:where (into [:or]
+                                    (for [{:keys [id schema db_id]} to-delete-tables]
+                                      [:like :object (str "%" (perms/data-perms-path db_id schema id) "%")]))}))))))
```
tsmacdonald, line 37: Doesn't this get triggered by the hook?

```
(t2/define-before-delete :model/Table
  [{:keys [db_id schema id]}]
  (t2/delete! Permissions :object [:like (str (perms/data-perms-path db_id schema id) "%")]))
  ```
  
  
  If not, I'd maybe extract the `[:like ...` logic into a helper function in `metabase.models.Table` that's used by both the before-delete hook and this task, just to make extra sure they stay in sync down the line

File: src/metabase/task/delete_inactive_tables.clj (hunk 2 above)
Comment (qnkhuat) [reply src/metabase/task/delete_inactive_tables.clj]: ha it was not caught by the test because the new but inactive table in the test has `updated_at` is `now`, so it's not qualified as less than 180 seconds.

I've updated the test to have 2 tables:
- one have updated_at less than 1 day from the threshold
- one have updated_at 1 day older than the threshold

[dcc27b0](https://github.com/metabase/metabase/pull/35443/commits/dcc27b0ac75d7f99c76c0d16e2008fc4d9ed6ea4)

File: src/metabase/task/delete_inactive_tables.clj (hunk 3 above)
Comment (qnkhuat) [reply src/metabase/task/delete_inactive_tables.clj]: oh I assumed it was done in the `after-delete` so I didn't know it's exists.

anw I think it should all be handled by table hook.

but I think there is a bug in `before-delete`, it doesn't clean up the new feature permissions like `/download/limited/db/3/schema/public/table/14/` or `/data-model/db/3/schema/public/table/14/`.

I've updated it to handle those permissions as well in: dcc27b0 

I'll also ask for review from Noah

**Comment by qnkhuat** [comment]
@noahmoss could you help take a look at [this](https://github.com/metabase/metabase/pull/35443/files#diff-26256e81e30e896cefd35c0cf8eaf6f81fc4ccae57dc0af4392493aba27b1e2eR68) permission related change?

I think it didn't delete the feature permissions when cleaning up tables

File: src/metabase/models/table.clj [hunk 4]
```diff
@@ -62,9 +62,10 @@
                   :field_order  (driver/default-field-order (t2/select-one-fn :engine Database :id (:db_id table)))}]
     (merge defaults table)))
 
+;; TODO: should we do this after the deletion?
 (t2/define-before-delete :model/Table
   [{:keys [db_id schema id]}]
-  (t2/delete! Permissions :object [:like (str (perms/data-perms-path db_id schema id) "%")]))
+  (t2/delete! Permissions :object [:like (str "%" (perms/data-perms-path db_id schema id) "%")]))
```
tsmacdonald, line 68: The leading `%` gives me pause, but I defer to @noahmoss

tsmacdonald (commented):
Looks good to me, but agreed that Noah / someone from Admin Webapp should :heavy_check_mark:

test/metabase/models/table_test.clj
```diff
@@ -79,3 +80,18 @@
         (is (= "0395fe49"
                (serdes/raw-hash ["PUBLIC" "widget" db-hash])
                (serdes/identity-hash table)))))))
+
```
calherries, line 83:
```suggestion
```

calherries (changes requested):
I wonder if you've considered the implications this will have on executing queries that reference these fields/tables. I don't think we have any precendence for deleting tables and fields in the past, unless we have also deleted the database.

I would want to know that these cases are safely handled:
- running a query in the UI that references a deleted field/table
- executing a query in a subscription that references a deleted field/table

qnkhuat:
@calherries we're discussing about that here https://metaboat.slack.com/archives/C0641E4PB9B/p1699364273334209

luizarakaki (approved):
product behavior makes sense

test/metabase/task/delete_inactive_tables_test.clj
```diff
@@ -0,0 +1,45 @@
...
+                                                                    (:schema old-inactive-table)
+                                                                    (:id old-inactive-table)))))]
+        (#'delete-inactive-tables/delete-inactive-tables!)
+        (testing "inactive tables that are older than threshold should be deleted"
+          (is (= #{new-active-table-id old-active-table-id new-inactive-table-id}
+                 (t2/select-pks-set :model/Table :db_id db-id))))
+
+        (testing "related permission are deleted as well"
+          (is (false? (t2/exists? :model/permissions read-perm-id)))
+          (is (false? (t2/exists? :model/permissions download-perm-id))))))))
```
calherries, line 45: Testing for related permissions being deleted isn't necessary in this test, since you've already tested in `models.table-test`. This behaviour isn't specific to the task, it's relevant to tables being deleted in general.

calherries (approved):
LGTM

src/metabase/models/table.clj
calherries: We need the `%` because of prefixes like `download` or `data-model`. 
https://github.com/metabase/metabase/blob/d35c779976ae6bd79369e9cc6bec8540908f37d2/src/metabase/models/permissions.clj#L257

This is covered by ngoc's tests

src/metabase/models/table.clj
```diff
@@ -62,9 +62,10 @@
                   :field_order  (driver/default-field-order (t2/select-one-fn :engine Database :id (:db_id table)))}]
     (merge defaults table)))
 
+;; TODO: should we do this after the deletion?
```
calherries, line 65: Before is fine, and it's standard to do this cleanup before deletion for other models

src/metabase/models/table.clj
noahmoss: Yeah this should be fine—we should have had it work like that already.

noahmoss (approved):

src/metabase/task/delete_inactive_tables.clj
```diff
@@ -0,0 +1,54 @@
...
+(defmethod task/init! ::DeleteInactiveTables [_]
+  (let [job     (jobs/build
+                 (jobs/of-type DeleteInactiveTables)
+                 (jobs/with-identity (jobs/key job-key)))
+        trigger (triggers/build
+                 (triggers/with-identity (triggers/key trigger-key))
+                 (triggers/start-now)
+                 (triggers/with-schedule
+                   ;; run every day at 1 AM
+                   (cron/cron-schedule "0 0 1 * * ? *")))]
```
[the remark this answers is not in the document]
qnkhuat: changed to run at 1h30 every day

PR #35448, opened 2023-11-07
Adding Stacked Checkbox component

npfitz:
### Description
Adds a Stacked Checkbox component, which is just a specifically styled Mantine checkbox. Styled using static CSS selectors to hopefully make the migration to v7 easier in the future. This new checkbox is also set to be used in add / remove columns.

### How to verify
New Checkbox can be viewed in the Add / Remove columns panel (create a structured question and view viz settings) or found in storybook.

### Demo
image


.storybook/preview.tsx
```diff
@@ -14,5 +15,9 @@ export const parameters = {
 };
 
 export const decorators = [
-  renderStory => <ThemeProvider>{renderStory()}</ThemeProvider>,
+  renderStory => (
```
npfitz, line 18: The static class prefix by default is `mantine`. However, when using an emotion cache, it changes to the key of the emotion cache. By default our storybook previews didn't have a cache set, so the prefixes were different between storybook and the metabase app ('css' and 'emotion') respectively. This change brings those back in line.

ranquild (changes requested):
We could introduce another Checkbox "variant" and put all styles properly in `Checkbox`. This would allow us to not override anything using static CSS selectors.

npfitz:
Updated. Awesome Suggestion @ranquild

frontend/src/metabase/ui/components/inputs/Checkbox/Checkbox.styled.tsx
```diff
@@ -76,5 +77,40 @@ export const getCheckboxOverrides = (): MantineThemeOverride["components"] => ({
...
+            left: "4px",
+            height: "100%",
+            width: "100%",
+            borderRadius: "4px",
+            zIndex: -1,
+            backgroundColor: "white",
+            boxSizing: "border-box",
+          },
+          "&:checked:after": {
+            border: `2px solid ${color("brand")}`,
```
ranquild, line 104: `theme.colors.brand[1]`

frontend/src/metabase/ui/components/inputs/Checkbox/Checkbox.styled.tsx
```diff
@@ -76,5 +77,40 @@ export const getCheckboxOverrides = (): MantineThemeOverride["components"] => ({
...
+        label: {
+          paddingLeft: labelPosition === "right" ? "0.75rem" : "0.5rem",
+        },
+        input: {
+          "&:after": {
+            content: "''",
+            border: `1px solid ${theme.colors.gray[4]}`,
+            position: "absolute",
+            top: "-4px",
+            left: "4px",
```
ranquild, line 95: theme.spacing.xs

mazameli:
Nice! This is looking good to me so far. 

Detail question: why in storybook is the `size` param a string input rather than a list of discrete options like `sm`, `md`, `lg`?

npfitz:
> why in storybook is the size param a string input rather than a list of discrete options like sm, md, lg?
It should probably be a radio list, but that param will also accept an int as a number of pixels. Not sure how to make Storybook understand that, but I'll at least make the change you're suggesting here

ranquild (commented):
Code LGTM, there is one lint error

ranquild (approved):

ranquild:
@npfitz we use a radio for other components for `size`, we could do it here as well. Howerver I don't think that the component supports it right now

npfitz:
Updated to use radio component for size 👍

mazameli:
This may've been a preexisting issue with the `checkbox` component before this PR, and if so, I think we could solve it separately if we have to — but, when you add a description, the box is getting aligned incorrectly relative to the label. Here it is in Storybook, with the label on left and right:

image

image


And here it is in Mantine, with the label on both left and right. This is the desired alignment IMO:

image

image

mazameli:
Beyond that, this PR looks good to me.

npfitz:
Ohh, yeah. That was something I probably did in this PR. I'll see if I can get that fixed.

kdoh (approved):

Issue #35449, opened 2023-11-07
Set up writable database(s) directly in tests, rather than in a "snapshot" phase


nemanjaglumac:
Issues with sync during the snapshot phase of Cypress setup are making our CI barely usable. Build fail rate climbed above 50%!
The potential cause is: https://github.com/metabase/metabase/issues/27877

We can easily set up writable database only in tests that need it, which is a handful.
Even if sync continues to be a problem it will at least be localized to a single spec. This way, we'll hopefully unblock the whole E2E group and will enable it to run.

Conceptually, the change can be summed up like this.
Instead of:
1. db snapshots
    - create regular db snapshots
    - create writable db snapshots
2. tests
    - restore regular db snapshot
    - restore writable db snapshot

Do this:
1. db snapshots
    - create regular db snapshots
2. tests
    - restore regular db snapshot
    - create writable db snapshots
    - restore writable db snapshot

Single tests  and reproductions don't even need a snapshot. There we can simply set up a writable database and run the test directly against it.
```[tasklist]
### Affected specs
- [ ] e2e/test/scenarios/collections/uploads.cy.spec.js
- [ ] e2e/test/scenarios/custom-column/reproductions/27745-cc-numeric-missing-summarize.cy.spec.js
- [ ] e2e/test/scenarios/visualizations/object_detail.cy.spec.js
- [ ] https://github.com/metabase/metabase/pull/35451
```

nemanjaglumac:
We decided to try a different approach: #35498

PR #35450, opened 2023-11-07
Remove writable databases from global qa db snapshots [WIP]

nemanjaglumac:
This will be a base branch for #35449.


nemanjaglumac:
We decided to try a different approach: #35498

PR #35451, opened 2023-11-07
Set up writable database directly in `table.cy.spec.js`

nemanjaglumac:
Part of #35449.

Stress-test this change
https://github.com/metabase/metabase/actions/runs/6785782462/job/18444846137

```
  scenarios > visualizations > table > conditional formatting
    ✓ should work with boolean columns: burning 1 of 5 (6969ms)
    ✓ should work with boolean columns: burning 2 of 5 (3272ms)
    ✓ should work with boolean columns: burning 3 of 5 (3147ms)
    ✓ should work with boolean columns: burning 4 of 5 (3536ms)
    ✓ should work with boolean columns: burning 5 of 5 (3399ms)
```

### DIFF
--- a/e2e/test/scenarios/visualizations/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations/table.cy.spec.js
@@ -13,6 +13,8 @@ import {
   visitQuestionAdhoc,
   getTable,
   leftSidebar,
+  setupWritableDB,
+  addPostgresDatabase,
 } from "e2e/support/helpers";
 
 import { WRITABLE_DB_ID } from "e2e/support/cypress_data";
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/visualizations/table.cy.spec.js
```diff
@@ -320,10 +322,17 @@ describe("scenarios > visualizations > table", () => {
...
 
 describe("scenarios > visualizations > table > conditional formatting", () => {
+  before(() => {
+    restore("default");
+    cy.signInAsAdmin();
+
+    setupWritableDB("postgres");
+    addPostgresDatabase("Writable Postgres12", true);
+  });
+
```
nemanjaglumac, line 332: In theory, this part can live inside `beforeEach`. It doesn't make a difference in an isolated test. I just wanted to enforce this pattern because it will be useful for `describe` blocks with multiple tests. In such cases, we can simply set up things once (in a `before` hook`), take a db snapshot and then restore for every single test (in `beforeEach`).

Single tests don't even require snapshot/restore.

iethree: I think this is fine for a single test, but the reason why I put this in setup is that it is a potentially heavy operation that could take a few seconds - if we snapshot it, we never have to do it again, but if we do it across many tests, we're potentially wasting a bunch of time.

nemanjaglumac: The time we will "waste" is insignificant compared to a failure rate of more than 50%.
Which means you have to wait for about half an hour for tests that didn't fail to finish only to be able to re-run failed specs.

Compare that to 5 added seconds to a single spec.
This is exactly the reason I'm using `before` and not `beforeEach` hook. We will still do it only once (per spec), snapshot it and then be on our merry way.

nemanjaglumac:
We decided to try a different approach: #35498

PR #35445, opened 2023-11-07
Add Data Test ID for sidebar

oisincoveney:
Added a `data-testid` attribute for tests in #35280. Just wanted to put it in a different PR to keep things separated

PR #35461, opened 2023-11-07
Add explanatory toast and `Undo` when auto-applying filters

oisincoveney:
Epic: https://github.com/metabase/metabase/issues/35147

### How to verify

_**Adding a dashcard**_
1. Set up a dashboard with one or two dashcards and apply a filter to them
2. Add a dashcard.
3. Check the dashcard filter, you should see that it's connected to the parameter that was applied to the dashcards
4. See that there's an "Undo" banner. Clicking "Undo auto-connection" should prevent that card from subscribing to or setting auto-wired parameters for cards

_**Wiring existing dashcards**_

1. Set up a dashboard with two dashcards and wire them with the same parameter
2. Once the auto-wiring is set, see that there's an undo banner.
3. Clicking "Undo auto-connection" will disable the ability for that dashcard to set auto-wired parameters, or receive auto-wired parameters.


oisincoveney:
* master
  * **PR #35445** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35445" target="_blank">Graphite</a> 
    * **PR #35280** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35280" target="_blank">Graphite</a> 
      * **PR #35461** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35461" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35461?utm_source=stack-comment).

frontend/src/metabase/dashboard/reducers.unit.spec.js
```diff
@@ -42,6 +42,13 @@ describe("dashboard reducers", () => {
         toastId: null,
         toastDashboardId: null,
       },
+      autoWireParameters: {
```
ranquild, line 45: There was a similar issue with one of the PRs @npretto worked on where there was a piece of state for toasts when it wasn't needed. In fact, this change is a burden to maintain. Could you discuss this and avoid modifying the global state?

ranquild (changes requested):
Unexpected global state changes

frontend/src/metabase/dashboard/reducers.unit.spec.js
npretto: https://github.com/metabase/metabase/blob/cc50e7a94216ae738cb4aad609122abcf8082e32/frontend/src/metabase/dashboard/actions/tabs.ts#L102-L131
This is the solution I came up with, instead of storing the values I needed in redux, I basically stored them in the closure of the undo

frontend/src/metabase/dashboard/actions/cards.js
```diff
@@ -237,3 +242,73 @@ export const addActionToDashboard =
       }),
     );
   };
+
+export const autoApplyParametersToNewCard =
```
WiNloSt, line 246: This action seems to be unused. Since we have another similar function for that in `metabase/frontend/src/metabase/dashboard/actions/auto-wire-parameters/actions.ts`

WiNloSt (commented):
I've left some comments. I think you should address a few things.
1. Confirm with product whether or not the new behavior without new redux states would be ok.
2. Remove unused action.

frontend/src/metabase/dashboard/reducers.unit.spec.js
oisincoveney: Gotcha, I think a combo of Nicolo's and Kelvin's solution will do me good here

frontend/src/metabase/dashboard/actions/auto-wire-parameters/actions.ts
```diff
@@ -119,15 +170,57 @@ export function autoWireParametersToNewCard({
...
+    dispatch(
+      setDashCardAttributes({
+        id: dashcard_id,
+        attributes: {
+          parameter_mappings: parametersToAutoApply,
+        },
+      }),
+    );
+
+    const toastId = _.uniqueId();
```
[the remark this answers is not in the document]
oisincoveney: Yeah that makes a lot of sense. I'll use this

frontend/src/metabase/dashboard/actions/auto-wire-parameters/toasts.ts
```diff
@@ -0,0 +1,88 @@
...
+        message: t`This filter has been auto-connected with questions with the same field.`,
+        actionLabel: t`Undo auto-connection`,
+        undo: true,
+        action: () => {
+          dispatch(
+            setMultipleDashCardAttributes({
+              dashcards: modifiedDashcards.map(dc => ({
+                id: dc.id,
+                attributes: {
+                  parameter_mappings: getParameterMappings(
```
npretto, line 42: any reason we can't just use dc.parameter_mappings?
```diff
  setMultipleDashCardAttributes({
              dashcards: modifiedDashcards.map(dc => ({
                id: dc.id,
                attributes: {
-		 								parameter_mappings: getParameterMappings(
-                     dc,
-                     parameter_id,
-                     dc.card.id,
-                     null,
-                   ),
+                   parameter_mappings: dc.parameter_mappings,
                },
              })),
            }),
```
This should restore things as they were at the time of the toast being showed right?

oisincoveney: Nice catch! It does the job 😄

WiNloSt (commented):
Overall this is looking good. I have one problem with the shared toast ID

https://github.com/metabase/metabase/blob/744f2ef8a2e7117ac25e09f228749d1f0579f26c/frontend/src/metabase/redux/undo.js#L17-L19

since `addTodo` will call `setTimeout` on the toast. I got it into a weird state where

0. Add at least 3 cards to a dashboard
1. Map filter to a card
2. The toast appears
3. Remove mapping from 2 filters auto-wired cards
4. add a filter to 1 of the 2 cards
5. The toast appears, but the `setTimeout` from step 2 is still ongoing and usually kicks in around here causing the toast to disappear earlier than expected.

One way to definitely solve this problem is to put the toast ID in the store like you did before. I'm not sure if there is any other way.

frontend/src/metabase/dashboard/actions/auto-wire-parameters/actions.ts
```diff
@@ -87,6 +111,10 @@ export function autoWireParametersToNewCard({
       metadata,
       null,
       targetDashcard.card,
+      // TODO: mapping-options.js/getParameterMappingOptions needs to be converted to typescript as the TS
+      // checker thinks that this parameter should be (null | undefined), not (StoreDashcard/Dashcard | null | undefined)
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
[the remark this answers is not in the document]
oisincoveney: Ah I totally forgot about JSDoc. This is a better solution

frontend/src/metabase/dashboard/actions/auto-wire-parameters/toasts.ts
```diff
@@ -0,0 +1,81 @@
...
+    dispatch(
+      addUndo({
+        id: AUTO_WIRE_TOAST_ID,
+        message: t`This filter has been auto-connected with questions with the same field.`,
+        actionLabel: t`Undo auto-connection`,
+        undo: true,
+        action: () => {
+          dispatch(
+            setMultipleDashCardAttributes({
+              dashcards: modifiedDashcards.map(dc => ({
```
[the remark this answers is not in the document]
oisincoveney: Yeah I like this a lot, it keeps similar logic in a similar place.

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
```diff
@@ -786,6 +803,63 @@ describe("scenarios > dashboard > parameters", () => {
...
+            size_y: 4,
+          },
+        ];
+
+        createDashboardWithCards(cards).then(dashboardId => {
+          visitDashboard(dashboardId);
+          cy.wait("@dashboard");
+        });
+
+        cy.icon("pencil").click();
```
npretto, line 837:
```suggestion
        editDashboard()
```

oisincoveney: I've changed all of them 😄

oisincoveney:
> Overall this is looking good. I have one problem with the shared toast ID
> 
> https://github.com/metabase/metabase/blob/744f2ef8a2e7117ac25e09f228749d1f0579f26c/frontend/src/metabase/redux/undo.js#L17-L19
> 
> since `addTodo` will call `setTimeout` on the toast. I got it into a weird state where
> 
> 0. Add at least 3 cards to a dashboard
> 1. Map filter to a card
> 2. The toast appears
> 3. Remove mapping from 2 filters auto-wired cards
> 4. add a filter to 1 of the 2 cards
> 5. The toast appears, but the `setTimeout` from step 2 is still ongoing and usually kicks in around here causing the toast to disappear earlier than expected.
> 
> One way to definitely solve this problem is to put the toast ID in the store like you did before. I'm not sure if there is any other way.

@ranquild what do you think? I managed to replicate this as well

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
```diff
@@ -786,6 +803,63 @@ describe("scenarios > dashboard > parameters", () => {
...
+
+        cy.findByTestId("dashboard-header").icon("filter").click();
```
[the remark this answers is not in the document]
oisincoveney: all changed 😄

ranquild (approved):
Please add more tests covering the cancellation logic for toasts, with a least 1 more e2e test for how it works in this specific feature

npretto:
@ranquild I added a e2e test, I had to use some cy.wait to properly check the issue, do you have any other idea on how we could do it?

I also added some unit test for the redux part that use jest fake timers in case we're ok with just testing the timeout thing at that level

Issue #35047, opened 2023-10-25
Text weirdly wrapped in column dropdown in Polish language


kamilmielnik:
### Describe the bug

![image](https://github.com/metabase/metabase/assets/6830683/70184572-4d5a-469f-aa21-a64dc346e496)

### To Reproduce

1. Switch app language to Polish
2. Create a new GUI question
3. Select Orders table from Sample Database as data source
4. Aggregate rows count
5. Open breakout column dropdown

"Automatyczne grupowanie" label takes all room and actual column names are weirdly wrapped in multiple lines

### Expected behavior

"Automatyczne grupowanie" label is truncated

### Information about your Metabase installation

`master`, v47

### Severity

P3

PR #35479, opened 2023-11-08
Manual backport - Fix - Text weirdly wrapped in column dropdown in Polish language (#35349)

kamilmielnik:
Manual backport of #35349 (which is a fix for #35047)

### DIFF
--- a/frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
+++ b/frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
@@ -6,11 +6,14 @@ import type { ColorName } from "metabase/lib/colors/types";
 
 export const TriggerIcon = styled(Icon)`
   color: ${color("white")} !important;
+  flex: 0 0 auto;
 `;
 
 export const TriggerButton = styled.button`
   display: flex;
   align-items: center;
+  min-width: 0;
+  max-width: 50%;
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
```diff
@@ -6,11 +6,14 @@ import type { ColorName } from "metabase/lib/colors/types";
 
 export const TriggerIcon = styled(Icon)`
   color: ${color("white")} !important;
+  flex: 0 0 auto;
 `;
 
 export const TriggerButton = styled.button`
   display: flex;
   align-items: center;
+  min-width: 0;
```
uladzimirdev, line 15: @kamilmielnik shouldn't we use flex-base, flex-shrink? just curious

uladzimirdev (approved):

uladzimirdev (approved):

frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
kamilmielnik: I think we cannot (I tried many options) due to content of the button having ellipsis applied on it.
`flex-shrink` just does nothing (it's `flex: 0 1 auto` by default already).

This `min-width: 0` trick already bailed me out many times in the past.

PR #33006, opened 2023-08-08
Fix field Values search for PKs

qnkhuat:
Fix #32985 (see this [comment](https://github.com/metabase/metabase/issues/32985#issuecomment-1669404944) for how to reproduce)

See the comments for explaination

PR #32599, opened 2023-07-24
Change field values to have consistent shape in API responses

calherries:
This PR is a refactoring to tidy up field values code on both the backend and frontend. There is no change in behaviour.

The main change is that all BE endpoints that respond with field values, now have a consistent shape that matches how field values are defined in types on the FE.

Previously, assuming there is no field remapping happening, the `values` in responses from the endpoints `CardApi.parameterSearch/Values`, `DashboardApi.parameterSearch/Values`, `ParameterApi.parameterSearch/Values`, and `MetabaseApi.field_search/values` could either be an array of raw field values, or an array of 1 length arrays, each containing a array field value.

If `values` was a list of single row values, these values would immediately be wrapped in arrays, like so:
{values: `["v1", "v2", "v3"], ...} -> {values: [["v1"], ["v2"], ["v3"]], ...}` 

But the endpoints can also return `values` already wrapped in arrays of one element (1-tuples), meaning no transformation is needed on the FE:
{values: `[["v1"], ["v2"], ["v3"]], ...} -> {values: [["v1"], ["v2"], ["v3"]], ...}` 

This PR makes all BE endpoints return `values` in the latter shape, and removes the transformation step from the FE code. Malli schemas help to ensure all the endpoints stick to this consistent shape, but each endpoint is also covered with tests.

**BE changes**: updating all functions that produce field values for the FE.

**FE changes**: omit transforming all the data coming out of `CardApi.parameterSearch/Values`, `DashboardApi.parameterSearch/Values`, `ParameterApi.parameterSearch/Values`, and `MetabaseApi.field_search/values` endpoints.

PR #35480, opened 2023-11-08
Manual backport: Fix field Values search for PKs (#33006)

qnkhuat:


calherries:
@qnkhuat there's a failing test. When you fix it, can you explain why it happened?

qnkhuat:
@calherries the test expects `[["African"]]` but the function is returning `[["African" "African"]]`, this is the old shape of field values before you've refactored it in https://github.com/metabase/metabase/pull/32599

calherries (approved):
Ah yes, makes sense

PR #35481, opened 2023-11-08
Use `malli=?` instead of `#hawk/malli` etc

calherries:
Reader tags are meant to be used with data literals, not code to be evaluated. If not implemented carefully it can lead to subtle bugs like https://github.com/metabase/hawk/issues/12.

<s>This PR proposes to use a function, `malli=?` instead of the reader tag `#hawk/malli`. If the PR is approved I will update the hawk repo to implement the `malli=?` function, and create more functions for the other reader tags for `=?` [here](https://github.com/metabase/hawk/blob/main/docs/approximately-equal.md).</s>

`#hawk/malli` and other reader tags have been removed from hawk in this PR https://github.com/metabase/hawk/pull/14. This PR updates hawk and uses the new functions instead.


tsmacdonald:
I don't really have strong opinions on this.

`#hawk/malli` is working fine after Dan's fix, IMHO, and is a little more ergonomic to me. It draws on the tradition of Common Lisp's [read macros](https://letoverlambda.com/index.cl/guest/chap4.html).

> Reader tags are meant to be used with data literals, not code to be evaluated.

```
(defn read-malli
  "Data reader for `#hawk/malli`."
  [schema-form]
  `(->Malli ~schema-form))
```
  
This treats the code as data and feels like a normal thing to do.

The function approach is arguably easier to work with in a paredit context and has the fringe benefit of side-stepping a marginalia parser bug with reader tags that I need to fix.

If we're going to bikeshed about function names, I find the existing `->Malli` nicer (and easier to type) than `=?-malli`.

I guess in conclusion I don't see a reason to change, but if enough other people are going to clamor for it then I won't oppose it. I think I'd have exactly the same opinion if the situation was reversed and the proposal was to move from the function approach to the reader tag approach—not enough reason to change.

calherries:
I guess the reasons to prefer functions to tagged literals are small, but I could try list them:
1. implementations of them have this footgun (and there might be others)
2. we don't need them here
3. they are missing docstrings + go-to definitions etc that functions have
4. some people don't like them for aesthetic reasons, which means the hawk library won't be adopted outside metabase (speculative)
5. some libraries have bugs with tagged literals, such as marginalia

bshepherdson (commented):
I'm not a fan of the naming pattern. I would have expected the reverse: `malli=?`, `exactly=?`, etc. Thoughts?

calherries:
@bshepherdson Nice idea. `mt/malli=?` vs `mt/=?-malli` is definitely easier on the eyes

escherize (dismissed):
At first glance it will be easier for especially someone new to read. 

It is less magic, and less magic is good. 

I don't see the reason for using a reader literal here when a function does the job.

calherries:
I've created the PR in hawk here: https://github.com/metabase/hawk/pull/14. 

The implementation is slightly different to this proposal, because I wanted other users of this library to be able to use it ergonomically without needing to import them into a `metabase.test`-like namespace. The TL;DR is that we can choose to either import these vars into `metabase.test`, or require them from hawk in the test namespaces directly, like this:
```
(require '[mb.hawk.assert-exprs.approximately-equal :as =?])

and then later we can call

(is (=? (=?/malli [:map [:a :int]]) {:a 1})
```

calherries:
https://github.com/metabase/hawk/pull/14 was merged, so I've updated this PR to use it. Instead of needing to require `[mb.hawk.assert-exprs.approximately-equal :as =?]` in every namespace, I chose to import the function into metabase.test like so:

https://github.com/metabase/metabase/blob/661471fb9df9cf1c88de711bdf310bbf0dc9589e/test/metabase/test.clj#L301-L302

qnkhuat (approved):

PR #35485, opened 2023-11-08
Stable order of serialization keys

### LINKED ISSUE #22196: Serialization: order of keys is not consistent
The order of keys in a serialized yaml is not consistent. When comparing two different yamls for the exact same instance, a comparison tool like git can find differences when there is none.

The order of the keys should be consistent.
The order of nested elements should also be consistent.

It is a requirement for the git integration story.

sansolovyov:
This makes serialization output yamls using specified order. 

This only happens in the last commit, I also have done just alphabetical sorting, but @luizarakaki requested to sort them like that. So I want us to reach some consensus here @bshepherdson. :)

Resolves #22196


resources/serialization-order.edn
```diff
@@ -0,0 +1,137 @@
...
+              :dataset
+              :created_at
+              :public_uuid],
+ :Collection [:description
+              :archived
+              :slug
+              :name
+              :personal_owner_id
+              :type
+              :aauthority_level
```
bshepherdson, line 32: typo?

resources/serialization-order.edn
```diff
@@ -0,0 +1,137 @@
+{:Card       [:description
```
bshepherdson, line 1: If we're going to have a bespoke order defined in this file, then I think this (which seems to be pretty straight from DB column order?) could be improved a lot. There are human-friendly fields like `Card.name` below potentially-huge embedded records like `Card.result_metadata`.

Since many of these files end up being hundreds of lines or more, we should take this opportunity to make the files friendly to human readers looking at the top "paragraph" of the files.

I'd also remark that some of the appdb columns don't get exported in YAML, and other columns are added during extraction that don't exist in the appdb.

Three things to do here:

1. Check this file against the `extract`ed forms or eyeball the `extract-one` implementations; some of these columns are dropped and some others are added.
2. Reorder this file to make the order human friendly.
3. Include several models that are missing here! DashboardCard, DashboardTab, Timeline and TimelineEvent, and more.
    - Remember that many models are embedded in others; we still want a consistent order for DashboardCard's keys even though they're embedded on Dashboards, etc.

## Proposed order

Roughly the order should be:
- meaningful human things (`:name`, `:description`, `:archived`, `Database.engine`, `Card.dataset`)
- `:serdes/meta`
- short and fairly readable internals (`:database_id`, `:table_id`, `:collection_position`, `:creator_id`, `Card.display`, `Database.details`, etc.)
- short(ish) and less readable internals (`embedding_params`, `dataset_query`, etc.)
- huge embedded lists (`dashboard_cards`, `result_metadata`, `parameters`)

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -24,11 +25,41 @@
...
 
+(def ^:private serialization-order
+  (delay (-> (edn/read-string (slurp (io/resource "serialization-order.edn")))
+             (update-keys name)
+             (update-vals (fn [order]
+                            (into {} (map vector order (range))))))))
+
+(def ^:private serialization-sorted-map
+  (memoize
+   (fn [t]
```
bshepherdson, line 36: Let's give the parameters meaningful names like `model-name`. I thought this was memoizing the maps being stored for a second, then realized this is memoizing the "model -> empty map with the right sort order" function.

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -24,11 +25,41 @@
 
 (set! *warn-on-reflection* true)
 
+(def ^:private serialization-order
+  (delay (-> (edn/read-string (slurp (io/resource "serialization-order.edn")))
+             (update-keys name)
+             (update-vals (fn [order]
+                            (into {} (map vector order (range))))))))
+
+(def ^:private serialization-sorted-map
```
bshepherdson, line 34: Unless the function being memoized is really trivial, we tend to write these as
```clojure
(defn- sorted-map-for-model* [model]
  #_implementation-goes-here)

(def ^:private sorted-map-for-model (memoize sorted-map-for-model*))
```

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -24,11 +25,41 @@
...
+       ;; known columns are sorted by their order, then unknown are sorted alphabetically
+       (let [getter #(if (contains? order %)
+                       [0 (get order %)]
+                       [1 %])]
+         (sorted-map-by (fn [k1 k2]
+                          (compare (getter k1) (getter k2)))))
+       (sorted-map)))))
+
+(defn- serialization-deep-sort [m]
+  (let [t    (-> (:serdes/meta m) last :model)
```
bshepherdson, line 47: This should be `some->`; this file is still technically used by v1 serdes as well as v2, and there's no `:serdes/meta` in v1. I think with `some->` it'll Just Work for v1: passing `nil` to `serialization-sorted-map` and getting back a plain `sorted-map` for alphabetic order.

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -24,11 +25,41 @@
...
+        smap (serialization-sorted-map t)]
+    (into smap
+          (update-vals m (fn [val]
+                           (cond
+                             (map? val)               (serialization-deep-sort val)
```
bshepherdson, line 52: Embedded things like dashcards generally don't have `:serdes/meta` on them, so they won't be ordered with a meaningful order, only alphabetically. That's not so bad, but I think we can do better.

Probably the best way to make this work is to change the `model-name -> ordering` mapping above to `[model-name :key :path] -> ordering`. Then `["Dashboard"]` gives the ordering for a Dashboard from the file, and `["Dashboard" :ordered_cards]` can return the ordering for `DashboardCard`s (which isn't in the file yet but should be). Then this recursion would keep track of the path as it goes. (Note that we don't care about indexes in sequences, just about nesting in maps.)

enterprise/backend/src/metabase_enterprise/serialization/serialize.clj
```diff
@@ -238,3 +238,18 @@
 (defmethod serialize-one NativeQuerySnippet
   [snippet]
   (select-keys snippet [:name :description :content]))
+
+(defmethod serialize-one Metric
```
bshepherdson, line 242: This is v1 serdes and we really don't want to be changing it.

If v1's tests are failing because of the changes to the YAML ordering, we should just fork `metabase-enterprise.serialization.dump` into separate namespaces like everything else.

enterprise/backend/test/metabase_enterprise/serialization/v2/storage_test.clj
```diff
@@ -150,3 +153,39 @@
...
+        (with-redefs [spit (fn [fname yaml-data]
+                             (testing fname
+                               (doseq [coll (->> (yaml/parse-string yaml-data)
+                                                 (tree-seq coll? identity)
+                                                 (filter map?))]
+                                 (let [t     (-> (:serdes/meta coll) last :model)
+                                       order (when t
+                                               (get @@#'dump/serialization-order t))]
+                                   (if-not order
+                                     (is (= (not-empty (sort (keys coll)))
```
bshepherdson, line 176: Let's make sure that the set of things without a matching order is what we expect.

Or, better, let's intend to have a defined order for everything and assert `(is (some? order))` in all cases. One of the requirements for serdes v2 is to enforce with tests that any new models are properly handled, and a test like this which forces new models to get added to `serialization-order.edn` is a win.

bshepherdson (changes requested):

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: IDK, I prefer nil-punning to `some->` when it works:

```
user> (-> nil last :model)
nil
```

resources/serialization-order.edn
piranha: Just updated to that message

enterprise/backend/src/metabase_enterprise/serialization/serialize.clj
piranha: I've dropped the commit that changed this. I've just noticed that there are tests there which do not run at all, so I've made them run, but yeah, not really necessary for the change I'm making.

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: Done

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: Done

resources/serialization-order.edn
piranha: Fixed

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: Done in the last commit, works for both the path (like `[:Dashboard :ordered_cards]`) or for the last key in the path (`:serdes/meta`).

enterprise/backend/test/metabase_enterprise/serialization/v2/storage_test.clj
piranha: I've added a few models manually, but now I'm thinking if there is a good way to iterate through all defined models?..

bshepherdson: There are already lists for this in `metabase-enterprise.serialiazation.v2.models`, both the `exported-models` that exist at the top level and `inlined-models` that are nested in something else. The list of all models that are defined in the system is `metabase-enterprise.serialization.v2.entity-ids/toucan-models`.

I started writing a paragraph about adding more tests that serdes things are implemented for all models, but the right place for that is as a bug: #35540. Definitely not part of this PR!

resources/serialization-order.edn
bshepherdson: LGTM now.

bshepherdson (approved):

src/metabase/util.cljc
```diff
@@ -841,9 +841,3 @@
   [xs]
   (or (empty? xs)
       (apply distinct? xs)))
-
-(defn deep-sort-map
```
piranha, line 845: @bshepherdson are you okay with me removing it from utils and accompanied test? It's not used anywhere except serialization, and I'm replacing it with this PR anyway.

luizarakaki (approved):

src/metabase/util.cljc
bshepherdson: Yeah, we don't need dead code lying around. Go ahead and remove it if no one else is using it.

PR #35496, opened 2023-11-08
Remove a retired field's FK target references during sync

### LINKED ISSUE #13130: Data Model and Object Details shows non-active columns as Foreign Key targets and can cause query breakage if remapped to foreign key
**Describe the bug**
When a column is dropped from a table, then the field in Metabase is set to `active=false`, but the field is still shown in Foreign Key target reference selector, causing confusion and will break queries if "Display values"="Use foreign key".

**To Reproduce**
1. Create two tables:
```
create table test_13075_a (
ID int default null,
fkref int default null
);
insert into test_13075_a (ID, fkref) VALUES (1, 1);
create table test_13075_b (
ID int default null,
old_ID int default null,
fkname text default null
);
insert into test_13075_b (ID, old_ID, fkname) VALUES (1, 1, 'test');
```
2. Admin > Databases > (db) > Sync database schema now
3. Admin > Data Model > (db) > test_13075_b > set `old_ID` field to "Entity Key"
4. Admin > Data Model > (db) > test_13075_a > set `fkref` to "Foreign Key" and reference `test_13075_b.old_ID`, :gear: and change "Display values"="Use foreign key".
5. Drop the column
```
alter table test_13075_b drop column old_ID;
```
6. Admin > Databases > (db) > Sync database schema now
7. Now browsing the table `test_13075_a` will break because of "Display values", and going to Data Model will still shown references to `test_13075_b.old_ID` in the Foreign Key selector.

**Expected behavior**
I guess that any Foreign Key references to a non-active column should be set to `null` at sync, so even if the column was added back, then that wouldn't restore the FK references or "Display values".

Metabase 0.36.0, 0.36.4

**Additional context**
Found during debug of https://github.com/metabase/metabase/issues/13075#issuecomment-670100056

calherries:
Fixes https://github.com/metabase/metabase/issues/13130 by setting `metabase_field.fk_target_field_id` and `metabase_field.semantic_type` to `NULL` if a field has `active` set to `false` during sync.

I chose to implement this with a `define-before-update` hook on the Field model, rather than the more direct approach to modify the sync code (see [this commit to compare implementations](https://github.com/metabase/metabase/pull/35496/commits/4ba23e247de2e28c6a9ec4c961e8e88751f11f5c)). I couldn't find any other instances where we set active to false other than sync, but it's worth enforcing this at the model level.

I'm going to follow this PR up with a migration to solve the issue for previously removed FK targets.

### DIFF
--- a/test/metabase/sync/sync_metadata/fields_test.clj
+++ b/test/metabase/sync/sync_metadata/fields_test.clj
@@ -107,7 +107,7 @@
                                   :table_id [:in (t2/select-pks-set Table :db_id (u/the-id database))])))))))
 
 (deftest mark-inactive-remove-fks-test
-  (testing "when a column is dropped from the DB, sync should wipe foreign key targets"
+  (testing "when a column is dropped from the DB, sync should wipe foreign key targets and their semantic type"
     (is (=? {:before-sync {:semantic_type      :type/FK
                            :fk_target_field_id int?}
              :after-sync  {:semantic_type      nil
--- a/test/metabase/sync/sync_metadata/fields_test.clj
+++ b/test/metabase/sync/sync_metadata/fields_test.clj
@@ -23,16 +23,11 @@
                        ;; create table & load data
                        "DROP TABLE IF EXISTS \"birds\";"
                        "CREATE TABLE \"birds\" (\"species\" VARCHAR PRIMARY KEY, \"example_name\" VARCHAR);"
-                       "CREATE TABLE \"flocks\" (\"id\" INTEGER PRIMARY KEY, \"example_bird_name\" VARCHAR);"
                        "GRANT ALL ON \"birds\" TO GUEST;"
                        (str "INSERT INTO \"birds\" (\"species\", \"example_name\") VALUES "
                             "('House Finch', 'Marshawn Finch'),  "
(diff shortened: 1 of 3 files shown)


test/metabase/sync/sync_metadata/fields_test.clj
```diff
@@ -14,32 +14,55 @@
...
-   5.  Returns a map containing results from both calls to `f` for comparison."
-  {:style/indent [:fn]}
-  [alter-sql f]
-  ;; first, create a new in-memory test DB and add some data to it
+(defn- do-with-test-db [thunk]
   (one-off-dbs/with-blank-db
-    (doseq [statement [ ;; H2 needs that 'guest' user for QP purposes. Set that up
+    (doseq [statement [;; H2 needs that 'guest' user for QP purposes. Set that up
                        "CREATE USER IF NOT EXISTS GUEST PASSWORD 'guest';"
-                       ;; Keep DB open until we say otherwise :)
```
calherries, line 30: I generally enjoy emojis but this smile was causing widespread desctruction to my editor

qnkhuat:
why does it need to wipe the semantic_field?

qnkhuat (approved):
Cool, makes sense to me implementing this with hooks

PR #35498, opened 2023-11-08
Optimistic Database Sync in e2e tests

### LINKED ISSUE #35489: Database Sync Fails in e2e test setup
image (4)

[Slack Discussion](https://metaboat.slack.com/archives/C5XHN8GLW/p1697714338212399)

likely related to https://github.com/metabase/metabase/issues/27877

possibly fixed by https://github.com/metabase/metabase/issues/35449

iethree:
Closes https://github.com/metabase/metabase/issues/35489 🤞 🤞 🤞 

### Description

We're having persistent problems with database sync failing as part of our setup for e2e tests in CI. After investigation it looks like sync actually succeeds, it's just that the API doesn't properly report the sync as complete. This no longer fails our whole e2e suite when the sync (apparently, but not really) doesn't complete.

### DIFF
--- a/e2e/runner/cypress-runner-backend.js
+++ b/e2e/runner/cypress-runner-backend.js
@@ -27,6 +27,7 @@ const CypressBackend = {
         "-Dh2.bindAddress=localhost", // fix H2 randomly not working (?)
         "-Djava.awt.headless=true", // when running on macOS prevent little Java icon from popping up in Dock
         "-Duser.timezone=US/Pacific",
+        // if you comment this line 👇 you can get (very noisy) backend console logs in the terminal for e2e tests
         `-Dlog4j.configurationFile=file:${__dirname}/../../frontend/test/__runner__/log4j2.xml`,
         "-Dclojure.server.repl={:port,5555,:accept,clojure.core.server/repl}",
       ];
--- a/e2e/runner/cypress-runner-backend.js
+++ b/e2e/runner/cypress-runner-backend.js
@@ -28,7 +28,7 @@ const CypressBackend = {
         "-Djava.awt.headless=true", // when running on macOS prevent little Java icon from popping up in Dock
         "-Duser.timezone=US/Pacific",
         // if you comment this line 👇 you can get (very noisy) backend console logs in the terminal for e2e tests
-        `-Dlog4j.configurationFile=file:${__dirname}/../../frontend/test/__runner__/log4j2.xml`,
+        // `-Dlog4j.configurationFile=file:${__dirname}/../../frontend/test/__runner__/log4j2.xml`,
         "-Dclojure.server.repl={:port,5555,:accept,clojure.core.server/repl}",
       ];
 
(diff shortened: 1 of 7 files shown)


e2e/support/helpers/e2e-qa-databases-helpers.js
```diff
@@ -104,13 +104,21 @@ function assertOnDatabaseMetadata(engine) {
...
 function recursiveCheck(id, i = 0) {
   // Let's not wait more than 20s for the sync to finish
-  if (i === 40) {
-    throw new Error("The sync is taking too long. Something is wrong.");
+  if (i === 20) {
+    cy.task(
+      "log",
+      "The DB sync isn't complete yet, but let's be optimistic about it",
+    );
+    return;
```
iethree, line 112: So I think the solution here is just, don't throw an error if the sync doesn't report as complete after 20 seconds. My investigation revealed that often the sync actually _is_ complete according to the BE console output, even though the API says it isn't.

**Likely case**: the sync is actually complete and everything is fine.
**Fine case**: the sync failed and this workflow doesn't even use this DB, so we didn't block it unnecessarily.
**Worst case**: the sync isn't complete and a test fails, which is what's happening already, so we're no worse off.

e2e/runner/cypress-runner-backend.js
```diff
@@ -27,7 +27,8 @@ const CypressBackend = {
         "-Dh2.bindAddress=localhost", // fix H2 randomly not working (?)
         "-Djava.awt.headless=true", // when running on macOS prevent little Java icon from popping up in Dock
         "-Duser.timezone=US/Pacific",
-        `-Dlog4j.configurationFile=file:${__dirname}/../../frontend/test/__runner__/log4j2.xml`,
+        // if you comment this line 👇 you can get (very noisy) backend console logs in the terminal for e2e tests
+        // `-Dlog4j.configurationFile=file:${__dirname}/../../frontend/test/__runner__/log4j2.xml`,
```
nemanjaglumac, line 31: Did you intentionally leave this commented in order to have BE logs in CI?

e2e/support/helpers/e2e-qa-databases-helpers.js
nemanjaglumac: Worth trying 🤷‍♂️

nemanjaglumac (approved):
Left a comment regarding backend logger that is still printing on every job. In case you wanted to turn it off, feel free to commit that one thing with `[ci skip]` and we'll force-merge this PR afterwards.

The idea is definitely worth trying.
Thank you for digging into this yesterday and for shining some more light on the potential root causes.

e2e/runner/cypress-runner-backend.js
iethree: yeah, kept this in for visibility, don't want to merge it this way 🔥

calherries:
I'll revert these changes to `e2e/support/helpers/e2e-qa-databases-helpers.js` when I land https://github.com/metabase/metabase/pull/35713, which should fix the underlying issue.

Issue #10764, opened 2019-08-30
Do not translate error log messages, or add unique error codes for each error


flamber:
Currently error logs are also translated, but it makes it really difficult to understand for anyone trying to help (if you don't speak that language) and also makes it really difficult for the one with the problem to search for similar problems or solutions.

Proposal: Either don't translate the error logs or add error codes (like Windows :fearful:).

:arrow_down: Please click the :+1: reaction instead of leaving a `+1` or `update?` comment

__________________________________________

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/issues/33340
- [ ] Consider error codes and their hierarchy
```

camsaul:
I like the error codes idea, it's really the best of both worlds.

If we came up with a convention for returning the error codes we could easily have Kondo enforce this rule, e.g. every `ex-info` has to include the `:error-code` key or something like that.

For QP stuff we're currently using a `:type` key with an error type hierarchy in https://github.com/metabase/metabase/blob/master/src/metabase/query_processor/error_type.clj, but those aren't unique. It could be a good starting point tho for unique error code stuff.

I don't know how feasible it would be to have Kondo enforce error code uniqueness across the entire application. We wouldn't want to use something like `gensym` because we wouldn't want the error message to change between versions. I think the main problem is that because Kondo lints stuff atomically there's not really a way to ensure uniqueness between error codes in unrelated namespaces. Some ideas:

1. We could require that the `:error-code` be a namespaced keyword for the current namespace. Then we should be able to track the keywords we saw in that namespace in a Kondo linter and error if we see a duplicate in the same namespace. Downside: error codes are longer/harder to search. Benefit: instant feedback
2. Swap out `ex-info` in `:test` mode and have it record error codes it sees and error if it sees a duplicate. Since this is `:test` only we wouldn't have to worry about problems with reloading code. Downside: no feedback until you run tests from the CLI or in CI. 
3. Write a test that walks the codebase and looks for `:error-code`s inside `ex-info` and errors on duplicates. Downside: no feedback until you run that specific test
4. Write some sort of `(ensure-unique-error-code :my-error-code)` macro that tracks all the error codes it's called with and errors at macroexpansion time if you use the error code more than once. Enforce usage of this with Kondo. To make reloading work it should only error in `:test` mode. Same downsides as (2)

I think maybe some sort of hybrid approach like this would work:

[...]

PR #35499, opened 2023-11-08
Deprecate PUT /api/dashboard/:id/cards endpoint

calherries:

This PR adds a warning log if some 3rd party tries to use the `PUT /api/dashboard/:id/cards` endpoint, and updates the docstring to say it's deprecated.

It also tidies up `PUT /api/dashboard/:id` a bit (no change in behaviour).

**Release notes mention:**
`PUT /api/dashboard/:id/cards` is deprecated. Instead you can use `PUT /api/dashboard/:id` to update the dashcards and tabs of a dashboard, along with all the other dashboard data.

### DIFF
--- a/src/metabase/api/dashboard.clj
+++ b/src/metabase/api/dashboard.clj
@@ -40,7 +40,7 @@
    [metabase.query-processor.util :as qp.util]
    [metabase.related :as related]
    [metabase.util :as u]
-   [metabase.util.i18n :refer [deferred-tru tru]]
+   [metabase.util.i18n :refer [deferred-tru trs tru]]
    [metabase.util.log :as log]
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
@@ -638,7 +638,8 @@
   (update-dashboard id dash-updates))
 
 (api/defendpoint PUT "/:id/cards"
-  "Update `Cards` and `Tabs` on a Dashboard. Request body should have the form:
+  "(DEPRECATED -- Use the `PUT /api/dashboard/:id` endpoint instead.)
+   Update `Cards` and `Tabs` on a Dashboard. Request body should have the form:
 
     {:cards        [{:id                 ... ; DashboardCard ID
                      :size_x             ...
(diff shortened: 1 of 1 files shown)


src/metabase/api/dashboard.clj
```diff
@@ -607,9 +608,7 @@
        true))
     (let [dashboard (t2/select-one :model/Dashboard id)]
       (events/publish-event! :event/dashboard-update {:object dashboard :user-id api/*current-user-id*})
-      (when update-dashcards-and-tabs?
-        ;; execute these events for old PUT /api/dashboard/:id/cards calls and new PUT /api/dashboard/:id calls, and not for old PUT /api/dashboard/:id calls
-        (track-dashcard-and-tab-events! dashboard @changes-stats))
+      (track-dashcard-and-tab-events! dashboard @changes-stats)
```
calherries, line 611: This `when` was unnecessary, since `track-dashcard-and-tab-events!` should only track actual changes anyway

src/metabase/api/dashboard.clj
```diff
@@ -607,9 +608,7 @@
        true))
     (let [dashboard (t2/select-one :model/Dashboard id)]
       (events/publish-event! :event/dashboard-update {:object dashboard :user-id api/*current-user-id*})
-      (when update-dashcards-and-tabs?
-        ;; execute these events for old PUT /api/dashboard/:id/cards calls and new PUT /api/dashboard/:id calls, and not for old PUT /api/dashboard/:id calls
```
calherries, line 611: This comment was out of date and should have been removed before

src/metabase/api/dashboard.clj
```diff
@@ -657,6 +657,8 @@
    ;; tabs should be required in production, making it optional because lots of
    ;; e2e tests curerntly doesn't include it
    tabs [:maybe (ms/maps-with-unique-key [:sequential UpdatedDashboardTab] :id)]}
+  (log/warn
+   (trs "DELETE /api/dashboard/:id/cards is deprecated. Use PUT /api/dashboard/:id instead."))
```
qnkhuat, line 661: From now on we don't want to translation for log https://github.com/metabase/metabase/issues/10764
```suggestion
   "DELETE /api/dashboard/:id/cards is deprecated. Use PUT /api/dashboard/:id instead.")
```

qnkhuat (approved):

PR #35502, opened 2023-11-08
restore correct position when undoing dashcard tab movement or deletion

npretto:
### Description

**Problem:** Undoing a dash-card "moved-to-another-tab" or deletion does not always put the card back in the place it was moved/deleted from.

https://github.com/metabase/metabase/assets/1914270/365db584-5b68-45f0-bf6a-5fd7a3bad5dd

This pr tries to address it:

https://github.com/metabase/metabase/assets/1914270/2118a32c-53e2-4230-b1da-e4bf8217b445

The code is... not what I expected to write to fix this, but it seems to work and I haven't found any issue so 🤷 

### How to verify

1. Stack two cards horizontally
2. Move the top one to another tab/remove it from the dashboard
3. Press undo

## Known issues
This doesn't always work (see videos), but it's a bit better than before

Case 1:
there's a "hole" and a card goes too far up

https://github.com/metabase/metabase/assets/1914270/347eff62-33d4-408d-ace5-badde431569f

Case 2:

it doesn't work if you press undo on multiple cards in the wrong order 
see https://github.com/metabase/metabase/pull/35502#issuecomment-1805778244

### DIFF
--- a/e2e/test/scenarios/dashboard-cards/dashboard-card-undo.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/dashboard-card-undo.cy.spec.js
@@ -1,0 +1,96 @@
+import {
+  createNewTab,
+  editDashboard,
+  getDashboardCard,
+  getDashboardCards,
+  getTextCardDetails,
+  goToTab,
+  moveDashCardToTab,
+  removeDashboardCard,
+  restore,
+  undo,
+  updateDashboardCards,
+  visitDashboard,
+} from "e2e/support/helpers";
+
+describe("scenarios > dashboard cards > undo", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsAdmin();
+  });
+
+  it(
+    "when undoing a dashcard removal or dashcard tab movement, it should try to restore the position (best effort)",
+    { scrollBehavior: false },
+    () => {
+      const checkOrder = () => {
+        getDashboardCard(0).findByText("Text card 1");
+        getDashboardCard(1).findByText("Text card 2");
+        getDashboardCard(2).findByText("Text card 3");
+        getDashboardCard(3).findByText("Text card 4");
+      };
+
+      const cards = [
+        getTextCardDetails({
+          text: "Text card 1",
+          size_x: 4,
+          size_y: 1,
+          row: 0,
+          col: 1,
+        }),
+        getTextCardDetails({
+          text: "Text card 2",
+          size_x: 4,
+          size_y: 1,
+          row: 1,
+          col: 0,
+        }),
+        getTextCardDetails({
+          text: "Text card 3",
+          size_x: 4,
+          size_y: 1,
+          row: 2,
+          col: 3,
+        }),
+        getTextCardDetails({
+          text: "Text card 4",
+          size_x: 4,
+          size_y: 1,
+          row: 3,
+          col: 0,
+        }),
+      ];
+
+      cy.createDashboard().then(({ body: { id: dashboard_id } }) => {
+        updateDashboardCards({ dashboard_id, cards });
+
+        visitDashboard(dashboard_id);
+      });
+
+      checkOrder();
+
+      editDashboard();
+
+      for (let i = 0; i < cards.length; i++) {
+        removeDashboardCard(i);
+        getDashboardCards().should("have.length", cards.length - 1);
+
+        undo();
+        getDashboardCards().should("have.length", cards.length);
+        checkOrder();
+      }
+
+      createNewTab();
+      goToTab("Tab 1");
+
+      for (let i = 0; i < cards.length; i++) {
+        moveDashCardToTab({ dashcardIndex: i, tabName: "Tab 2" });
+        getDashboardCards().should("have.length", cards.length - 1);
+
+        undo();
+        getDashboardCards().should("have.length", cards.length);
+        checkOrder();
+      }
+    },
+  );
+});
--- a/frontend/src/metabase/dashboard/actions/cards.js
+++ b/frontend/src/metabase/dashboard/actions/cards.js
@@ -13,6 +13,7 @@ import { createCard } from "metabase/lib/card";
 import { getVisualizationRaw } from "metabase/visualizations";
 import { trackCardCreated } from "../analytics";
 import { getDashCardById } from "../selectors";
+import { isVirtualDashCard } from "../utils";
 import {
   ADD_CARD_TO_DASH,
   REMOVE_CARD_FROM_DASH,
@@ -77,7 +78,9 @@ export const undoRemoveCardFromDashboard = createThunkAction(
       const dashcard = getDashCardById(getState(), dashcardId);
       const card = dashcard.card;
 
-      dispatch(fetchCardData(card, dashcard));
+      if (!isVirtualDashCard(dashcard)) {
+        dispatch(fetchCardData(card, dashcard));
+      }
 
       return { dashcardId };
     },
--- a/frontend/src/metabase/dashboard/actions/tabs.ts
+++ b/frontend/src/metabase/dashboard/actions/tabs.ts
@@ -366,7 +366,9 @@ export const tabsReducer = createReducer<DashboardState>(
       ) => {
         const dashCard = state.dashcards[dashCardId];
 
-        dashCard.row = originalRow;
+        // when a card is moved, another card could steal its position,
+        // this trick gives the original tab precedence on the position
+        dashCard.row = originalRow - 0.1;
         dashCard.col = originalCol;
         dashCard.dashboard_tab_id = originalTabId;
         dashCard.isDirty = true;
--- a/frontend/src/metabase/dashboard/actions/tabs.ts
+++ b/frontend/src/metabase/dashboard/actions/tabs.ts
@@ -23,6 +23,7 @@ import { addUndo } from "metabase/redux/undo";
 import { INITIAL_DASHBOARD_STATE } from "../constants";
 import { getDashCardById } from "../selectors";
 import { trackCardMoved } from "../analytics";
+import { calculateDashCardRowAfterUndo } from "../utils";
 import { getDashCardMoveToTabUndoMessage, getExistingDashCards } from "./utils";
 
 type CreateNewTabPayload = { tabId: DashboardTabId };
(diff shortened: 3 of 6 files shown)


WiNloSt (commented):
I don't feel like this is the approach we should go with despite being very simple, but it's also very hacky.

Also, can you add some tests in the meantime? I'll try to think of some ways to deal with this without a hack.

WiNloSt:
We have the logic to sort `dashcards` here 
https://github.com/metabase/metabase/blob/76fd35e97488d334462502b98b4f0eff3fda18d4/frontend/src/metabase/dashboard/selectors/selectors.js#L121-L134

And it affects how the card would be put back into their positions In this case, the card that has smaller integer ID would be put first in `orderedDashcards`

You can see how this is played out on `master`.

https://www.loom.com/share/938630f3b05e4db080174e30da481e0a

I think we should be more expressive how we want to deal with such situation. Maybe having a new key on a card like `sortPolicy: "first" | "last"` and if it's `first` we put it first. This should help making undo removing/moving cards positions more deterministic. It should be very similar to how you `- 0.1` on the `row` but more expressive.

What do you think?

npretto:
Ok so I think I went down the rabbit hole a bit, as things weren't making sense at first.

> And it affects how the card would be put back into their positions In this case

I think that's only the case when both card A and B have the same col and row.
I tried to add a "isUndone" field to the dashcards and to sort by them, I verified that the array was sorted correctly, but it was still not working when cards didn't have the same col.
I found this code in the library that could explain it. Basically it returns 0 (and so it uses the order we pass) only when .x and .y are equal.
In the other cases, they sort the cards themselves and kinda ignore our order.
```ts
export function sortLayoutItemsByRowCol(layout: Layout): Layout {
  // Slice to clone array as sort modifies
  return layout.slice(0).sort(function (a, b) {
    if (a.y > b.y || (a.y === b.y && a.x > b.x)) {
      return 1;
    } else if (a.y === b.y && a.x === b.x) {
      // Without this, we can get different sort results in IE vs. Chrome/FF
      return 0;
    }
    return -1;
  });
}
```

I did some tests with the following code:
```diff
    const orderedDashcards = dashboard.dashcards
      .map(id => dashcards[id])
      .filter(dc => !dc.isRemoved)
      .sort((a, b) => {
        const rowDiff = a.row - b.row;

        // sort by y position first
        if (rowDiff !== 0) {
          return rowDiff;
        }

        // for items on the same row, sort by x position
        return a.col - b.col;
      })
+     .reverse();
```

This should prove that the order of orderedDashcards only matters when col and row are the same (but I've went down the rabbit hole so I may as well be mistaken)

https://github.com/metabase/metabase/assets/1914270/95c52b41-ae2b-42ba-a91d-7084140d09a8


With this new knowledge I'll double check if the code in this pr actually fixes the problem or not by testing a more complex scenario.

This task is supposed to be timeboxed to ~1 day, which is almost over, so I'll delay writing the automated test as it may not be fixed anytime soon anyway

WiNloSt:
> With this new knowledge I'll double check if the code in this pr actually fixes the problem or not by testing a more complex scenario.
> 
> This task is supposed to be timeboxed to ~1 day, which is almost over, so I'll delay writing the automated test as it may not be fixed anytime soon anyway

Sounds good to me, but let's add some tests either unit, or E2E to cover the change in this PR.

e2e/test/scenarios/dashboard-cards/dashboard-card-undo.cy.spec.js
```diff
@@ -0,0 +1,96 @@
+import {
```
npretto, line 1: recording of the e2e:

Uploading Screen Recording 2023-11-10 at 12.48.34.mov…

frontend/src/metabase/dashboard/actions/cards.js
```diff
@@ -77,7 +78,9 @@ export const undoRemoveCardFromDashboard = createThunkAction(
       const dashcard = getDashCardById(getState(), dashcardId);
       const card = dashcard.card;
 
-      dispatch(fetchCardData(card, dashcard));
+      if (!isVirtualDashCard(dashcard)) {
```
npretto, line 81: had to cherry pick this fix from #35545 in order to make this test work

frontend/src/metabase/dashboard/actions/tabs.ts
```diff
@@ -366,7 +366,9 @@ export const tabsReducer = createReducer<DashboardState>(
       ) => {
         const dashCard = state.dashcards[dashCardId];
 
-        dashCard.row = originalRow;
+        // when a card is moved, another card could steal its position,
+        // this trick gives the original tab precedence on the position
+        dashCard.row = originalRow - 0.1;
```
oisincoveney, line 371: Let's put this magic number into a constant (not sure about naming, maybe UNDO_ROW_PRECEDENCE or something)

oisincoveney:
I was wondering how the row precedence would work, and I think I found an issue. If we remove two cards that are on top of each other and we click `Undo` in the reverse order of how they were removed, then the cards are reversed. I think maybe you can adjust the constant to account for the row it was previously in to ensure that cards that are in the queue of `removed` cards (for lack of a better term) are put back in their place even when other cards aren't there

https://github.com/metabase/metabase/assets/25306947/77ea8238-7eec-4a03-80a7-63fdb6aebad4

npretto:
> I think maybe you can adjust the constant to account for the row it was previously in to ensure that cards that are in the queue of `removed` cards (for lack of a better term) are put back in their place even when other cards aren't there

The issue is that after we remove the first card, the one below it gets its position, so I don't think we can do much without writing a lot of code, which is most likely not worth this super edge case.

Another issue is that right after we do the undo with the "magic" number, the number gets rounded by the grid that calls `SET_MULTIPLE_DASHCARD_ATTRIBUTES` and change its state in the store too.

The timebox allocated to this bug has reached its ende (1 day of work), I'll summarise the results to conor on slack and we'll see what we can do

frontend/src/metabase/dashboard/actions/tabs.ts
npretto: Thanks for the suggestion, I moved it to a function so that I could also consolidate the comment and explanation in just one place.
Hopefully this also makes the solution a bit less hacky

JesseSDevaney (commented):
Question: Whenever we tab-move or delete a card, why do we not just save and store the previous dashcards state at the time of move/delete? Then if we undo any specific tab-move or delete action, we can just restore the state to what it looked like previously? 

Was this option considered?

**Quick, Hackish Implementation:**

https://github.com/metabase/metabase/assets/22608765/2df17b93-6f30-4aa2-9445-6a2b443b4815

frontend/src/metabase/dashboard/utils.ts
```diff
@@ -283,3 +283,13 @@ export const getActionIsEnabledInDatabase = (
...
+
+/**
+ * When you remove a dashcard from a dashboard (either via removing or via moving it to another tab),
+ * another dashcard can take its place. This small offset ensures that the grid will put this dashcard
+ * in the correct place, pushing back down the other card.
+ * This is a "best effort" solution, it doesn't always work but it's good enough for the most common case
+ * see https://github.com/metabase/metabase/pull/35502
+ */
+export const calculateDashCardRowAfterUndo = (originalRow: number) =>
+  originalRow - 0.1;
```
EmmadUsmani, line 295: Would the decimal row value get saved to the database, or is it rounded? I wonder if things can start moving to the wrong place if this function is called too many times and the row number gets dropped by 1 or more

npretto: The grid component automatically rounds it, afaik `SET_MULTIPLE_DASHCARD_ATTRIBUTES` gets called with the rounded values immediately after we do the undo

npretto:
> Was this option considered?

I've tried to discard solution that involved too much code compared to the outcome, so not really from my side

> Question: Whenever we tab-move or delete a card, why do we not just save and store the previous dashcards state at the time of move/delete? Then if we undo any specific tab-move or delete action, we can just restore the state to what it looked like previously?
 

I've got some feedback against storing too much state in global state to do undos as it's hard to maintain:
https://github.com/metabase/metabase/pull/34961#discussion_r1376025141
https://github.com/metabase/metabase/pull/35461#discussion_r1391185748

What would happen in this scenario:
- remove/move-tab card A
- move card B to another position in the grid
- undo 
Would it also undo the movement from the second step?

JesseSDevaney:
@npretto First, thanks for picking up this issue and working on it :heart: 

> I've tried to discard solution that involved too much code compared to the outcome, so not really from my side

:+1: I understand that due to the time-boxing requirements.

> What would happen in this scenario:
> 
> * remove/move-tab card A
> * move card B to another position in the grid
> * undo
> 
> Would it also undo the movement from the second step?

Yes, in my rough example it would undo the movement from the second step. 

---
**Dashboard Revert Changes Example**
I did test another way of reverting changes that would address this and other issues (as well as be cool to have for creators), but it does require more effort to nail down.

https://github.com/metabase/metabase/assets/22608765/4342d3ca-a691-4f06-93ad-2ac459150d0e

**Paths Forward**
1. Implement the solution in this PR and do not worry about anything else
2. Implement the solution in this PR and in a future project, address the correctness of this as well as add full dashboard, edited changes reversion 
3. Delay this solution and in a future project, address this by adding full dashboard, edited changes reversion 

@cdeweyx Would you mind weighing in on what you feel is the best path forward for product? I proposed an alternative solution that would address this issue as well as add a cool new feature that I think creators/dashboard editors would like - the caveat is that it of course will require more effort.

JesseSDevaney (commented):
This hack seems like the only "quick-fix" to this issue, but it does introduce save failures if the user does not click undo on the same tab from which they moved the card. `SET_MULTIPLE_DASHCARD_ATTRIBUTES` only gets called to round the value if you undo the action on the same tab that the card will be restored to or you navigate back to the tab the card was restored on. 

I foresee it as a VERY UNLIKELY event that a user will do this without navigating back to the tab they restored the card to, but still worth mentioning that it does introduce a bug.

https://github.com/metabase/metabase/assets/22608765/04f05967-c150-4508-a234-b293f4078994

I am going to check-in with Conor to see if this bug is 'okay' due to its unlikelihood. If so, I will put my approval on this.

JesseSDevaney (approved):
:rocket: Thanks for your patience, @npretto!

I am going to create a separate issue to track the discussions in this PR as well as document the save failure bug.
- Issue Tracking Discussions: https://github.com/metabase/metabase/issues/35885

PR #35505, opened 2023-11-08
docs - csv entity key

jeff-bruemmer:
Documents how Metabase will try and detect an ID column in an uploaded CSV, or auto-generate one.


cdeweyx (approved):
LGTM

PR #35506, opened 2023-11-08
Add new indexes for audit v2

noahmoss:
New indexes for audit v2 (pulled in from [this PR](https://github.com/metabase/metabase/pull/35421)) and a migration to run them.


resources/migrations/instance_analytics_views/h2-indexes.sql
```diff
@@ -0,0 +1,7 @@
+drop index if exists idx_view_log_timestamp;
```
noahmoss, line 1: I've changed the create & drop index commands to use `if exists` and `if not exists` across the board. Shouldn't be necessary but just wanted to be extra defensive.

resources/migrations/instance_analytics_views/h2-indexes.sql
```diff
@@ -0,0 +1,7 @@
...
+
+create index if not exists idx_view_model_id
+    on view_log(model_id);
```
noahmoss: `idx_view_log_timestamp` already existed as an index on `model_id` — I think it was just named incorrectly. So I've added back `idx_view_log_model_id` with the fixed name

resources/migrations/instance_analytics_views/h2-indexes.sql
noahmoss: (Removing them from MySQL; didn't realize the syntax is not supported)

johnswanson (approved):
LGTM!

PR #35508, opened 2023-11-08
Add properties to query_execution

### LINKED ISSUE #35176: Add properties to query_execution
Tweaks to query_execution table

- [ ] Add new property: `is_sandboxed`
- [ ] Verify if `cache_hit` is working (it is likely not working, so we need to fix it)

qwef:
Closes https://github.com/metabase/metabase/issues/35176

### Description

Adding a function to check whether a query will have sandboxing in the future. I did try to swap to ordering of query processing in userland such that we wouldn't have to run some of the sandboxing code twice, but that caused a lot of unexpected side effects, so I reverted to the way I have now.

### How to verify

1. Create a sandboxed user on a certain table and query that table
2. Check the Query execution table to see is_sandboxed = true


noahmoss (changes requested):
I'm a little concerned that this is not going to accurately detect whether a query has sandboxing applied. There's a comment in `metabase.query-processor` about how the sandboxing middleware needs to be run after joins are resolved because new table IDs might have been introduced to the query: https://github.com/metabase/metabase/blob/master/src/metabase/query_processor.clj#L144-L145

So if we're only detecting sandboxing based on the initial form of the query, before any preprocessing, it'll miss some cases. Ideally we'd use the exact same result as what's generated by the actual middleware.

I'm not actually sure the best way to resolve this. `add-and-save-execution-info-xform!` has access to the query metadata (`acc`) it seems, but the sandboxing middleware is a pre-processing middleware so it can't manipulate the metadata. Maybe there could be a different, independent post-processing middleware that checks whether a query was sandboxed (via the `::qp.perms/perms` key) and adds that to the metadata.

I think that would work but it might be worth reaching out to the QP/drivers team to see if anyone has a better way to approach it. Happy to help investigate more too

enterprise/backend/src/metabase_enterprise/sandbox/query_processor/middleware/row_level_restrictions.clj
```diff
@@ -365,20 +365,22 @@
...
+    (-> metadata
+        (update :cols merge-cols)
+        (assoc :is_sandboxed (some? (get-in query [::qp.perms/perms :gtaps]))))))
 
 (defenterprise merge-sandboxing-metadata
   "Post-processing middleware. Merges in column metadata from the original, unsandboxed version of the query."
   :feature :sandboxes
-  [{::keys [original-metadata]} rff]
+  [{::keys [original-metadata] :as query} rff]
   (if original-metadata
```
noahmoss, line 383: I'm not sure what cases `original-metadata` would be `nil`, but if it is, we still want to pass the `:is_sandboxed` flag in the new metadata. Could you refactor this so that `is_sandboxed` is always included?

noahmoss (changes requested):
Almost there, just one correctness issue I noticed.

Could you also add an integration test that checks that `is_sandboxed` is recorded in `query_execution` correctly for a sandboxed query? You could probably put it in `metabase-enterprise.sandbox.query-processor.middleware.row-level-restrictions-test` but make use of the `with-query-execution` macro in `metabase.query-processor.middleware.process-userland-query-test` (by changing it to be non-private)

noahmoss (approved):
Awesome 👍

PR #35511, opened 2023-11-08
Adding linter and formatter pre-commit hooks for staged BE files.

markbastian:
Adding linter and formatter to staged BE files.

Note that `.cljfmt/indents.clj` tries to capture our indent rules and, so far, seems to get things right. There are still some inconsistencies in some of our nses in how we indent assoc(-in) and some t2 functions when it comes to key-value pairs on following lines. Should the keys align with the function call or be indented. Aesthetically, people seem to like indents, but these are functions, so should align with the first argument. IDK that it really matters as long as we have agreement.

We may, as we adopt this, have some files be reformatted in unexpected ways for items that have been missed. If so, the developer can simply update the .cljfmt/indents.clj file, rerun the hook, and commit that along with their other changes.


.husky/pre-commit
```diff
@@ -2,4 +2,11 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn precommit
-./hooks/pre-commit.nocommit
+
+# Run the linter
+./bin/quick_lint.bb
```
escherize, line 7: After looking at it, feels like setting up lint-staged to call clj-kondo makes sense, it will be simpler to have one way to handle linting. Also that would be 1 place to handle precommit hooks, too.  The formatting can happen the same way. Notice the FE folks are calling `prettier` in the `"lint-staged"` config in `package.json`.

escherize (commented):
I think that just adding the precommit checks into [lint-staged](https://github.com/lint-staged/lint-staged) would be a lot more simple, and also less code.

All it would take is adding something like:

```json
"**/*.{clj,cljc,cljs,bb}":
      ["clj-kondo --config ./.clj-kondo/config.edn --config-dir ./.clj-kondo --parallel --lint",
       "whitespace-check-files.sh"]
```

into package.json under the `"lint-staged"` key. That way it would lint all the clojure directories without needing to list them individually, and we can have just 1 place for these kinds of checks to live.

.husky/pre-commit
iethree: +1 , it's much easier to figure out what's happening on the pre-commit hook if they're all defined the same way

.cljfmt/indents.clj
```diff
@@ -1,2 +1,33 @@
...
+ defprotocol+                               [[:block 1] [:inner 1]]
+ ;ex-info                                 [[:inner 0]]
+ let-404                                    [[:block 1]]
+ mbql.u/match-one                           [[:inner 0]]
+ merge-with                                 [[:inner 0]]
+ mt/dataset                                 [[:inner 0]]
+ mt/formatted-rows                          [[:inner 0]]
+ mt/mbql-query                              [[:inner 0]]
+ mt/test-drivers                            [[:inner 0]]
+ mt/with-temp*                              [[:block 1]]
```
camsaul, line 15: We removed `mt/with-temp*`  months ago, so this file might be a little out of date 😳

camsaul:
> There are still some inconsistencies in some of our nses in how we indent assoc(-in) and some t2 functions when it comes to key-value pairs on following lines. Should the keys align with the function call or be indented. Aesthetically, people seem to like indents, but these are functions, so should align with the first argument.

There's no disagreement here, we just have some code that's indented wrong. Anything that's not following the indentation rules in the community Clojure Style Guide e.g. https://guide.clojure.style/#vertically-align-fn-args is indented wrong.

```clj
(assoc m
       :a b)
```

is correct; if you see

```clj
(assoc m
  :a b)
```

please fix it.

Almost everything in Toucan 2 is a regular plain function, so arguments should get indented the same way. Toucan 1 had some weird `:style/indent` specs that weren't consistent with the style guide rules, so some of our code is indented incorrectly as a result. Don't take that to mean there is any disagreement.

camsaul (approved):
Let's get this in and we can start iterating on the `indents.clj` file. We'll probably have a lot of churn initially as it fixes up wrong indentation but in the long run this will make things easier because we'll stop having random indentation fixes in unrelated PRs

bin/cljfmt.sh
```diff
@@ -1,7 +1,21 @@
 #! /usr/bin/env bash
 
-set -euxo pipefail
+# Reformat only staged files OR all files if --all is passed.
```
camsaul, line 3: Can you add some comments to these files about how to invoke these scripts? I'm unclear on how I would run this on just a single file

.cljfmt/indents.clj
```diff
@@ -1,2 +1,33 @@
-{metabase.api.collection/foo [[:inner 0]]
- toucan.db/insert!           [[:block 1]]}
+;; https://github.com/weavejester/cljfmt/blob/master/cljfmt/resources/cljfmt/indents/clojure.clj
+{auto-parse                                 [[:inner 0]]
```
camsaul, line 2: Could you run Clojure LSP and have it dump out the indents file it generates based on the `:style/indent` metadata?

bin/cljfmt.sh
```diff
@@ -1,7 +1,21 @@
...
+ARG=${1:-}
+
+STAGED_FILES=$(git diff --name-only --cached -- "*.clj" "*.cljc" "*.cljs")
+
+if [ "$ARG" == "--all" ]; then
+  clojure -M:cljfmt \
+    --indents ./.cljfmt/indents.clj
+elif [ "${#STAGED_FILES}" -gt 0 ]; then
+  clojure -M:cljfmt \
+    --indents ./local/indents.clj \
```
camsaul, line 17: why does this use a different indents file when there are staged files 😳

camsaul:
I tried to run this locally, we don't have a `:cljfmt` alias in `deps.edn`. You need to add that. The alias can specify the indents file or whatever so you don't need to pass it in explicitly every time you use it. 

Also, `--indents` is deprecated in `0.11.x` according to https://github.com/weavejester/cljfmt#breaking-changes-in-011x

If you rename that indents file to `.cljfmt.edn` it will pick it up automatically, let's do that.

Also I think we probably want `clj -T:cljfmt` rather than `-M` because it doesn't need to include Metabase's deps on the classpath when running. I added a profile like

```clj
  ;; clj -T:cljfmt fix '{:paths ["src/metabase/events/view_log.clj"]}'
  :cljfmt
  {:deps       {io.github.weavejester/cljfmt {:git/tag "0.11.2", :git/sha "fb26b22f569724b05c93eb2502592dfc2de898c3"}}
   :ns-default cljfmt.tool}
```

and was able to run it locally like

```
clj -T:cljfmt fix '{:paths ["src/metabase/events/view_log.clj"]}'
```

.cljfmt/indents.clj
```diff
@@ -1,2 +1,33 @@
...
+ perms/set-has-partial-permissions-for-set? [[:inner 0]]
+ publish-event!                             [[:block 1]]
+ qp.streaming/streaming-response            [[:inner 0]]
+ ;; Do we want this or not?
+ ;revision/revert!                        [[:inner 0]]
+ ;t2/select                                  [[:block 1]]
+ ;t2/select-one                              [[:block 1]]
+ t2/select-one-fn                           [[:inner 0]]
+ t2/select-pk->fn                           [[:inner 0]]
+ t2/update!                                 [[:inner 0]]
```
camsaul, line 29: These are wrong, these should all be formatted like normal functions

.cljfmt/indents.clj
```diff
@@ -1,2 +1,33 @@
...
+ t2.with-temp/with-temp                     [[:block 1]]
+ toucan.db/insert!                          [[:block 1]]
```
camsaul, line 31: Same, this should formatted like a normal function

camsaul (changes requested):
Alright, after playing around with this I don't think it's really in a state we can safely merge it into the repo. Maybe we should merge in the tooling without making it a hook yet. I'll 👍 a PR for something like that

There are some files that this absolutely destroys.

```
src/metabase/lib/metadata/protocols.cljc
```

is a perfect example -- try 

```
clj -T:cljfmt fix '{:paths ["src/metabase/lib/metadata/protocols.cljc"]}'
```

and see for yourself e.g.

```clj
(#?(:clj p/defprotocol+ :cljs defprotocol) CachedMetadataProvider
                                           "Optional. A protocol for a MetadataProvider that some sort of internal cache. This is mostly useful for
  MetadataProviders that can hit some sort of relatively expensive external service,
  e.g. [[metabase.lib.metadata.jvm/application-database-metadata-provider]]. The main purpose of this is to allow
  pre-warming the cache with stuff that was already fetched elsewhere.
  See [[metabase.models.metric/warmed-metadata-provider]] for example.

  See [[cached-metadata-provider]] below to wrap for a way to wrap an existing MetadataProvider to add caching on top
  of it."
                                           (cached-database [cached-metadata-provider]
                                                            "Get cached metadata for the query's Database.")
                                           (cached-metadata [cached-metadata-provider metadata-type id]
                                                            "Get cached metadata of a specific type, e.g. `:metadata/table`.")
                                           (store-database! [cached-metadata-provider database-metadata]
                                                            "Store metadata for the query's Database.")
                                           (store-metadata! [cached-metadata-provider metadata-type id metadata]
                                                            "Store metadata of a specific type, e.g. `:metadata/table`."))

(#?(:clj p/defprotocol+ :cljs defprotocol) BulkMetadataProvider
                                           "A protocol for a MetadataProvider that can fetch several objects in a single batched operation. This is mostly
  useful for MetadataProviders e.g. [[metabase.lib.metadata.jvm/application-database-metadata-provider]]."
                                           (bulk-metadata [bulk-metadata-provider metadata-type ids]
                                                          "Fetch lots of metadata of a specific type, e.g. `:metadata/table`, in a single bulk operation."))

```

I'm not sure how to make it indent correctly since we're using reader conditionals like

```clj
#?(:clj p/defprotocol+ :cljs defprotocol)
```

maybe we need to patch that upstream

.cljfmt/indents.clj
markbastian: Fixed

.cljfmt/indents.clj
markbastian: Fixed

bin/cljfmt.sh
markbastian: Fixed, OBE

markbastian:
@camsaul, I ran `clojure-lsp format --filenames src/metabase/lib/metadata/protocols.cljc` and it did the same file munging as calling `cljfmt` directly. I think this is not a problem solved by the lsp config. When you format the file using lsp does it change?

camsaul:
I haven't tried it using LSP, but this is a PR to add a pre-commit hook to format things with `cljfmt`, not LSP. I can't in good conscience approve that change before we work out some of the big problems with it.

I don't think there's a way to make this work correctly no matter how you configure `cljfmt`, I think this is something we'll have to submit a patch for upstream to fix. 

In terms of getting this PR in, I'm cool with the whitespace linter and Kondo being pre-commit hooks I guess. Personally I prefer to do lots of little commits locally and then fix my Kondo issues at the end before I open a PR, so I'm probably going to use `--no-verify` most of the time to skip these hooks so it doesn't take me 2 minutes for me to do a commit locally, but that's a me problem and if everyone else wants these hooks then I'm not going to be a stick in the mud.

But we should not make cljfmt a hook until we can resolve some things that I see as blocking problems.

I'm cool with committing tooling like a `deps.edn` profile for it and a helper script to run it so you can use it locally whenever you want, but we should not do it automatically against peoples wishes yet.

bin/cljfmt.sh
markbastian: I've renamed the files and made them explicit that they do one thing only -- run their specific job on only modified or added files.

.cljfmt/indents.clj
markbastian: I've asked around about this and IDK if it's a feature of lsp. Do you know how to do it? I _did_ run lsp on our protocol.clj file and it also does weird formatting. Maybe we can exclude that file from formatting until we find a more permanent solution.

markbastian:
> I haven't tried it using LSP lately, but this is a PR to add a pre-commit hook to format things with `cljfmt`, not LSP. I can't in good conscience approve that change before we work out some of the big problems with it.
> 
> I don't think there's a way to make this work correctly no matter how you configure `cljfmt`, I think this is something we'll have to submit a patch upstream to fix it.
> 
> In terms of getting this PR in, I'm cool with the whitespace linter and Kondo being pre-commit hooks I guess. Personally I prefer to do lots of little commits locally and then fix my Kondo issues at the end before I open a PR, so I'm probably going to use `--no-verify` most of the time to skip these hooks so it doesn't take me 2 minutes for me to do a commit locally, but that's a me problem and if everyone else wants these hooks then I'm not going to be a stick in the mud.
> 
> But we should not make cljfmt a hook until we can resolve some things that I see as blocking problems.
> 
> I'm cool with committing tooling like a `deps.edn` profile for it and a helper script to run it so you can use it locally whenever you want, but we should not do it automatically against peoples wishes yet.

I commented out the cljfmt linter in the .husky/pre-commit file with comments. Effectively, this PR now runs kondo and whitespace lint pre-commit hooks. The script to format staged files is in bin, but not wired in. If that sounds like a good incremental fix, we can merge this.

.husky/pre-commit
markbastian: Done!

camsaul (approved):

PR #35516, opened 2023-11-08
Remove read events from audit_log

noahmoss:
Pretty straightforward—these events are going into `view_log`, since we're keeping it around, so they don't need to go into `audit_log` anymore.


noahmoss:
Closing, will open a new PR

Issue #22563, opened 2022-05-09
Changing column visibility/order in GUI questions can cause inconsistency


urkle:
**Describe the bug**
I built a report today that has 6 "Grouped" fields (only a count summary) and 4 custom expression fields (to coerce an encoded integer into a dotted version) 

However, when trying to hide some of the columns (e.g the columns used to produce the custom expressions) it hides ALL of the grouped columns except the first one.

**Logs**
There are no logs generated around this error .  I encountered the error on metabase 0.42.0 AND 0.43.0

**To Reproduce**
1) Create a question inner joining two tables and grouping by 6 columns (all from the second table), and add a count summary.
2) create single custom expression. 
3) visualize the question
4) try to hide one of the grouped columns.  notice ALL the columns are hidden.

**Expected behavior**
Keep some of the grouped columns (but hide others) and keep one of the custom expressions (and hide the others).

**Partial workaround / more info**
Now, if I switch the table order so that my main table is where my grouping columns are from. Then I do not have any issues.
However the columns ARE listed two times still.

**Screenshots**
1) column settings initial (no column adjustments).. Note how the columns are showing up in the "more columns" when they should not.
<img width="320" alt="image" src="https://user-images.githubusercontent.com/101123/167498734-117ee0d5-ae70-452e-a237-d6727952e466.png">
2) after hiding "release - version"
<img width="330" alt="image" src="https://user-images.githubusercontent.com/101123/167498867-7ff82df2-863e-45d7-a862-28d645f2f415.png">
3) after adding "release - platform" back and then adding "release - status" in. (it only keeps the LAST one I added)
<img width="317" alt="image" src="https://user-images.githubusercontent.com/101123/167499034-6ebc522b-c6b4-4579-9377-1f006f4aec72.png">

Variation of switching the table order (where everything "works").. Notice how the "more columns" is still showing all the columns

<img width="329" alt="image" src="https://user-images.githubusercontent.com/101123/167499580-a598ba57-9be3-472d-969d-7f89bc10fd46.png">

**Information about your Metabase Installation:**

- Browser: Firefox 99.0.1
- OS OS X 10.15
- DB Postgresql 11
- Metabase version 0.42.0 and 0.43.0
- hosted on EBS
- Metabase internal DB Postgresql 11

**Severity**
Annoying.

flamber:
Almost a repeat of #12395. Likely very related to #13455 and maybe to #12044 and #22382

Steps to reproduce:
[...]

PR #35518, opened 2023-11-08
[MLv2] New, skipped test for removing a breakout field

bshepherdson:
The original issue pointed to by https://github.com/metabase/metabase/issues/22563 is fixed by the overhaul of the
column selection. But trying to reproduce this for the test showed that
the FE doesn't really let you remove a breakout column. It will hide
them, but if you click "Add/Remove columns" they're greyed out.

From an MBQL-Lib POV, though, `(lib/remove-field some-breakout-column)`
succeeds, but the resulting query is broken, and throws if you eg.
`(lib/returned-columns query)`. We should either:
- fix returned/visibile columns, or
- prevent `remove-field` from breaking the query, or
- make `remove-field` on a breakout column remove the breakout?

The test is skipped, but should be passing if we go with fix 1 above.

### DIFF
--- a/test/metabase/lib/field_test.cljc
+++ b/test/metabase/lib/field_test.cljc
@@ -1494,3 +1494,32 @@
                (lib/legacy-card-or-table-id (first (lib/returned-columns (-> query
                                                                              (lib/breakout (first (lib/returned-columns query)))
                                                                              lib/append-stage))))))))))
+
+(deftest ^:parallel fields-list-and-breakout-by-joined-columns-test
+  ;; TODO: This test throws on `(lib/returned-columns removed)`. The FE avoids this by knowing not to try to really
+  ;; remove-field on a breakout column. So this test failing has nothing to do with #22563, but we should either:
+  ;; - make remove-field refuse to remove a breakout column, or
+  ;; - make `returned-columns` work even after a breakout column was "removed" from the :fields list, or
+  ;; - fix `(lib/remove-field some-breakout-column)` to remove the breakout itself?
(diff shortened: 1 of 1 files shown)


bshepherdson:
* master
  * **PR #35518** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35518" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35518?utm_source=stack-comment).

snoe (commented):
I thought that breakouts should be removed with `remove-clause`, I don't think there's a problem here other than garbage-in garbage-out.

The fields that should be passed to `remove-field` I think are those returned by `lib/fields`

snoe:
> But perhaps that's not ever been supported, and we've always done that with FE-level hiding via viz-settings? (That's the current state, but it makes an MLv2 unit test for this impossible to write.)

I don't believe it's ever been supported. `returned-columns` is the combination of `fields` (source and joins) and `aggregations`, `breakouts`, and `expressions`. 

It would be nice to see us add something like `:hidden-fields` to `stage` but while it's in `viz-settings` I don't think we should be testing it as we simply store those settings for the FE (should be opaque to BE). We'd also need to be sensitive to schema changes such that if we have `:hidden-fields [[:field {} "CREATED_AT"]]` and the field is dropped from the db schema, then our query should be resilient.

metamben (approved):

PR #35522, opened 2023-11-09
Speed up audit DB tests

noahmoss:
I was noticing that `ensure-audit-db-installed!` takes like 20-30 seconds, so any BE tests that called it were extremely slow. This was mainly because it was re-loading all of the audit content each time it was called, even if it was already installed.

Here's my attempt to improve the situation.
* I've refactored `load-analytics-content` out into its own function
* I've added a `load-analytics-content` internal boolean setting so that content loading can be disabled via environment variable (requested by Arakaki)
* I've restructured the `with-audit-db-restoration` macro to only load audit content if the audit collection doesn't exist (i.e. on a fresh instance). We don't need to be continuously reloading audit content in tests when it hasn't changed at all.
* I've grouped several of the audit DB tests into a single test (`audit-db-installation-test`) and restructured them a bit so that we're not re-adding the audit DB and re-syncing content repeatedly. The actual tested logic should be the same.


enterprise/backend/test/metabase_enterprise/audit_db_test.clj
```diff
@@ -12,59 +12,54 @@
...
+    (testing "Audit DB does not have scheduled syncs"
       (let [db-has-sync-job-trigger? (fn [db-id]
                                        (contains?
                                         (set (map #(-> % :data (get "db-id"))
                                                   (task/job-info "metabase.task.sync-and-analyze.job")))
                                         db-id))]
-        (is (not (db-has-sync-job-trigger? perms/audit-db-id)))))))
+        (is (not (db-has-sync-job-trigger? perms/audit-db-id)))))
+
+    (testing "Audit DB doesn't get re-installed unless details change"
```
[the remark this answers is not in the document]
escherize: The details aren't used to determine if we need to update the audit db.

enterprise/backend/src/metabase_enterprise/internal_user.clj
```diff
@@ -4,23 +4,34 @@
...
+
+(defn- install-internal-user! []
+  (t2/insert-returning-instances! User internal-mb-user))
+
+(defn- internal-user-exists? []
+  (if-let [internal-user (t2/select-one User :email (:email internal-mb-user))]
+    (if (not= (:id internal-user) config/internal-mb-user-id)
+      (do
+        (log/info "Internal user already exists with an incorrect ID. Deleting...")
+        (t2/delete! User :email (:email internal-mb-user))
```
escherize, line 26: Can't say I love having this delete! in `internal-user-exists?`. The value it returns is only used in an if branch, so why not put the do block on e.g. line 38? I think that would clear up the `if-let if` ... `false true false` bit too, it's kind of hard to follow.

escherize (changes requested):
Thanks for re-writing the tests. 

The new setting makes sense.

The only bit I am hesitant to approve is having the delete and subsequent logic in `internal-user-exists?`.

enterprise/backend/src/metabase_enterprise/internal_user.clj
escherize: What do you think about doing the id-mismatch check when the internal user is about to be created?

luizarakaki: I'm not sure we need to delete the internal user. If it exists, just don't try to create it again?

noahmoss: Yeah I'm going to revert this change

escherize (approved):

PR #35523, opened 2023-11-09
Generate sane SQL for a DATE[TIME] expression = DATE literal filter

### LINKED ISSUE #17807: Adding an "on" date filter without a specific time to a custom date column returns no results
**Describe the bug**
Adding an "on" date filter without a time to a custom date column shows no results. Filtering on the exact date value works, and filtering using the "between" option works, but you can't filter on a date like `10/07/2019` and see rows with values like `10/07/2019 01:34am`

**Logs**
Here's an example query we send to the `/api/dataset` endpoint. See repro steps below.
```
{"database":1,"query":{"source-table":3,"expressions":{"CC Birth Date":["field",24,null],"CC Created At":["field",18,null]},"fields":[["field",30,null],["field",21,null],["field",23,null],["field",19,null],["field",22,null],["expression","CC Created At"],["field",29,null],["field",20,null],["field",25,null],["field",26,null],["field",24,{"temporal-unit":"default"}],["field",28,null],["field",27,null],["field",18,{"temporal-unit":"default"}],["expression","CC Birth Date"]],"filter":["=",["expression","CC Created At"],"2017-10-07"]},"type":"query","parameters":[]}
```

**To Reproduce**
1. Ask a Question > Custom Question > People dataset
2. Add a custom column that uses the expression `[Created At]`. Name it `CC Created At`
Screen Shot 2021-09-08 at 11 17 58 AM

3. Run the query. Click on a cell in the "CC Created At" column and select the `=` filter. You should see a filter like "Created At is October 7, 2017 01:34 am" above the table and the table should contain a single result row.

Screen Shot 2021-09-08 at 11 16 12 AM

4. Click on the filter to modify it. Add the same date to it but without a time. Run the query and observe that there are now no results.

Screen Shot 2021-09-08 at 11 16 26 AM

5. You can do this using the original "Created At" field and it'll work correctly.

**Expected behavior**
You'd expect custom column date fields to filter identically to how normal date fields filter.

This is on master, HEAD at `0b60ff9234797cb465dbf4040e2d180120b2fbc1`

**Severity**
P2

camsaul:
Fixes #17807

For a filter clause like 

```clj
[:= {} [:expression "expr" {:effective-type :type/DateTime}] "2023-11-08"]
```

we were actually generating SQL like

```clj
WHERE expr = ? -- "2023-11-08"
```

which was not only wrong from a type standpoint (we were passing in a String value for `?` rather than some sort of temporal value), but it was clearly not what was meant. A slight improvement was to fix the `wrap-value-literals` middleware to correctly handle values being compared to `:expression` references; thus the query became

```clj
WHERE expr = ? -- #t "2023-11-08"
```

An improvement, since not the types at least agree (except for extremely picky DBs like BigQuery), but still not great, because we clearly were asking for values where the date component was `2023-11-08` as opposed to `2023-11-08T00:00:00.000` specifically (assuming millisecond resolution). 

So part two of the fix was to make sure `wrap-value-literals` parses

```clj
"2023-11-08"
```

to

```clj
[:absolute-datetime #t "2023-11-08" :day]
```

e.g. with a `:day` temporal bucket.

Part three was to rework the `optimize-temporal-filters` middleware to optimize temporal comparison filters where the LHS has no bucketing, but the RHS does, e.g. filters with the shape

```
[<comparison filter e.g. = or !=> <lhs: temporal with no bucketing> <rhs: temporal WITH bucketing>
```

e.g. we now want to optimize 

```clj
[:= {} [:expression "expr" {:effective-type :type/DateTime}] [:absolute-datetime {} #t "2023-11-08" :day]]
```

to

```clj
[:between {} 
 [:expression "expr" {:effective-type :type/DateTime}]
 [:absolute-datetime {} #t "2023-11-08" :day]
 [:absolute-datetime {} #t "2023-11-09" :day]]
```

so we generate SQL like

```sql
WHERE expr BETWEEN ? AND ? -- #t "2023-11-08", #t "2023-11-09"
```

This gives us the results we expected.

Most of this logic already existed, but it only worked on `:field` refs, and only worked on filters where the LHS had a unit, so most of my work here was generalizing stuff a bit.

### Bonus

Totally unrelated, but I also realized that two equivalent MLv2 queries aren't always `=` to each other because equivalent metadata providers aren't `=`, so I implemented `Object.equals` (Clojure) / `IEquiv.-equiv` (ClojureScript) methods for all of our metadata provider types. This required changing some of them from anonymous `reify`s to actual `deftypes`. Now two equivalent metadata providers are equal to one another. Nice


src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -3,65 +3,104 @@
...
+(doseq [tag [:+ :-]]
+  (lib.hierarchy/derive tag ::arithmetic-expr))
+
+(defmethod temporal-unit ::arithmetic-expr
+  [[_tag & args]]
+  (some temporal-unit args))
+
+(defmethod temporal-unit :+
+  [[_+ & args]]
+  (some temporal-unit args))
```
bshepherdson, line 47: Isn't this redundant with the `::arithmetic-expr` version? (Does `derive` work on un-namespaced keywords?)

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -3,65 +3,104 @@
...
+
+(doseq [tag [:= :!= :< :<= :> :>=]]
+  (lib.hierarchy/derive tag ::binary-filter))
+
+(defn- compatible-units? [lhs-unit rhs-unit]
+  (case [(boolean lhs-unit) (boolean rhs-unit)]
+    [true true]   (= lhs-unit rhs-unit)
+    [true false]  true
+    [false true]  true
+    [false false] false))
```
bshepherdson, line 89: Maybe this is just my stylistic tastes, but I'm not a fan of these sorts of "truth tables".
I'd rather see this as

```suggestion
(if (and lhs-unit rhs-unit)
  (= lhs-unit rhs-unit)
  (boolean (or lhs-unit rhs-unit)))
```

bshepherdson (approved):

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
camsaul: yeah good point I'll change it.

PR #35530, opened 2023-11-09
remove dashcards not in serialization data when loading

### LINKED ISSUE #35420: Serialization `import` doesn't remove dashboardcards in the target instance
### To Reproduce

- User transfers dashboard 1 from instance A to instance B.
- Dashboard 1 has only one card (id=100) and is loaded as usual to instance B.
- On instance A, user removes the card 100 and adds the card 200.
- User runs the Serdes process as usual: export and import.
- As we don't delete anything in the target instance, now there are two cards in the target instance (card 100 and 200).

### Expected behavior

If we import a dashboard in the target instance, we should check all dashboardcards. 
We should:
- add cards that exist only in yamls
- replace cards that are found both in yamls and target instance
- remove cards that are only in the target instance (not found in yamls).

### Logs

_No response_

```JSON
- master
```

### Severity
P2
This creates divergences between target and source instances.
And if removed cards are in the same position of other cards, it breaks the dashboard layout

### Additional context

_No response_

sansolovyov:
Cards are not touched, just Dashboards are synchronized with loaded data

Resolves #35420

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/serialization/v2/load_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/serialization/v2/load_test.clj
@@ -977,3 +977,37 @@
             (is (some? action))
             (testing ":type should be a keyword again"
               (is (keyword? (:type action))))))))))
+
+(deftest remove-dashcards-test
+  (let [serialized (atom nil)
+        dash1s     (atom nil)
+        dash1d     (atom nil)
+        dashcard1s (atom nil)
+        dashcard2d (atom nil)]
+    (ts/with-source-and-dest-dbs
+      (testing "Serializing the original database"
+        (ts/with-source-db
+          (reset! dash1s (ts/create! Dashboard :name "My Dashboard"))
+          (reset! dashcard1s (ts/create! DashboardCard :dashboard_id (:id @dash1s)))
+
+          (reset! serialized (into [] (serdes.extract/extract {})))))
+
+      (testing "New dashcard will be removed on load"
+        (ts/with-dest-db
+          (reset! dash1d (ts/create! Dashboard :name "Weird Name" :entity_id (:entity_id @dash1s)))
+          ;; A dashcard to be removed since it does not exist in serialized data
+          (reset! dashcard2d (ts/create! DashboardCard :dashboard_id (:id @dash1d)))
+
+          ;; Load the serialized content.
+          (serdes.load/load-metabase (ingestion-in-memory @serialized))
+
+          (reset! dash1d (-> (t2/select-one Dashboard :name "My Dashboard")
+                             (t2/hydrate :dashcards)))
+
+          (testing "Dashboard has correct number of dashcards"
+            (is (= 1
+                   (count (:dashcards @dash1d))))
+            (is (= (:entity_id @dashcard1s)
+                   (get-in @dash1d [:dashcards 0 :entity_id])))
+            (is (not= (:entity_id @dashcard1s)
+                      (:entity_id @dashcard2d)))))))))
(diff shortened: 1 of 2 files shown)


calherries:
This solves the issue for the specific case of dashboard cards, but the bug is more general. From reading the code, we'll have the same issue with dashboard tabs, and all "embedded" entities, current and in the future. Ideally we should have a general solution to this problem for all embedded entities, or at least tests that cover them all.

calherries (approved):
If you cut the migration stuff you're playing with, this looks good to merge as it solves the specific issue with dashcards. But the more general issue remains - all embedded entities will suffer from the same problem, including dashboard tabs

src/metabase/models/dashboard.clj
```diff
@@ -615,15 +615,28 @@
...
                         {:model "DashboardTab" :id (:entity_id tab)}]))
 
+
+(defn- drop-excessive-nested!
+  "Remove nested entities which are not present in incoming serialization load"
+  [entities model field value]
+  (let [known     (t2/select-fn-set :entity_id model field value)
+        to-remove (set/difference known (set (map :entity_id entities)))]
+    (when (seq to-remove)
+      (t2/delete! model :entity_id [:in to-remove]))))
```
calherries, line 625: I wonder whether we can leverage hydration to do a lot of this for us. Since hydration is used to add the nested entities onto the extraction, we can assume that the nested entities are on the same hydration keys.

https://github.com/metabase/metabase/blob/abaf5e2d04f1b801e1722e07575ff42b6714a6f6/src/metabase/models/dashboard.clj#L580-L584
```clojure
(defn- drop-excessive-nested!
  "Delete nested local entities which are not present in incoming ingested entity."
  [ingested local hydration-key]
  (let [local-nested    (get (t2/hydrate local hydration-key) hydration-key)
        ingested-nested (get ingested hydration-key)
        to-remove       (set/difference (set (map :entity_id local))
                                        (set (map :entity_id ingested)))
        model           (t2/model (first local-nested))]
    (when (seq to-remove)
      (t2/delete! model :entity_id [:in to-remove]))))
```

Then we can call it in a much simpler way:
```
before
(drop-excessive-nested! (:tabs ingested) :model/DashboardTab :dashboard_id (:id dashboard))
after
(drop-excessive-nested! ingested dashboard :tags)
```

enterprise/backend/test/metabase_enterprise/serialization/v2/load_test.clj
```diff
@@ -977,3 +977,37 @@
             (is (some? action))
             (testing ":type should be a keyword again"
               (is (keyword? (:type action))))))))))
+
+(deftest remove-dashcards-test
```
calherries, line 981: You're probably onto this already but you just need an extra test for tabs

src/metabase/models/dashboard.clj
piranha: In theory this makes it possible to write `sync-nested!` function, which will take `[key ingested local make-fn]`. OTOH it could be obscuring the process a bit which will make writing `make-fn` a bit harder (it needs you to understand what's going on).

piranha: `sync-nested!` is somewhat harder, since we could have empty `local-nested` and then how do I determine model for insertion?.. I think I'll leave separate loop intact for now.

enterprise/backend/test/metabase_enterprise/serialization/v2/load_test.clj
piranha: yep, done

PR #35541, opened 2023-11-09
fix iframe-resizer readyCallback to onReady

maximepvrt:
fix this warning
```[iFrameSizer][iFrameResizer0] Deprecated: 'readyCallback' has been renamed 'onReady'. The old method will be removed in the next major version.```

### DIFF
--- a/frontend/src/metabase/lib/dom.js
+++ b/frontend/src/metabase/lib/dom.js
@@ -426,7 +426,7 @@ export function initializeIframeResizer(onReady = () => {}) {
     window.iFrameResizer = {
       autoResize: true,
       heightCalculationMethod: "max",
-      onReady: onReady,
+      onReady,
     };
 
     // FIXME: Crimes
--- a/frontend/src/metabase/lib/dom.js
+++ b/frontend/src/metabase/lib/dom.js
@@ -412,7 +412,7 @@ export function clipPathReference(id) {
   return `url(${url})`;
 }
 
-export function initializeIframeResizer(readyCallback = () => {}) {
+export function initializeIframeResizer(onReady = () => {}) {
   if (!isWithinIframe()) {
     return;
   }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by oisincoveney — APPROVED** [review]
I went through the [3.6.5 release](https://github.com/davidjbradshaw/iframe-resizer/tree/v3.6.5) and compared with the [4.3.2 release](https://github.com/davidjbradshaw/iframe-resizer/tree/v4.3.2) to verify the renaming. It looks like `readyCallback` was wholely renamed to `onReady`. 

I'll bring @WiNloSt into the loop since he was the last person to touch this function, just to get some verification on my side. Otherwise, it looks good to me!

Thanks for the contribution!

File: frontend/src/metabase/lib/dom.js [hunk 1]
```diff
@@ -421,12 +421,12 @@ export function initializeIframeResizer(readyCallback = () => {}) {
...
   if (window.iFrameResizer) {
     console.error("iFrameResizer resizer already defined.");
-    readyCallback();
+    onReady();
   } else {
     window.iFrameResizer = {
       autoResize: true,
       heightCalculationMethod: "max",
-      readyCallback: readyCallback,
+      onReady: onReady,
```
WiNloSt, line 429: nit: We don't use the value if it has the same name as the key.

WiNloSt (approved):
Tested. This works great. Thank you for your contribution!

WiNloSt:
@maximepvrt Can you please sign the [Contributor License Agreement](https://docs.google.com/a/metabase.com/forms/d/1oV38o7b9ONFSwuzwmERRMi9SYrhYeOrkbmNaq9pOJ_E/viewform), so we could proceed to the next steps before merging your contribution?

maximepvrt:
@WiNloSt I just signed the [Contributor License Agreement](https://docs.google.com/a/metabase.com/forms/d/1oV38o7b9ONFSwuzwmERRMi9SYrhYeOrkbmNaq9pOJ_E/viewform)

nemanjaglumac:
@oisincoveney and @WiNloSt do we need to backport this? Does it affect the current 47 release?

WiNloSt:
> @oisincoveney and @WiNloSt do we need to backport this? Does it affect the current 47 release?

@nemanjaglumac We don't need to backport this, but I follow the convention that if a PR is a simple bug fix then it should be backported.

This by the way, isn't even a simple bug fix, it's just remove an annoying warning message, everything should function the same with one less warning message.

PR #35542, opened 2023-11-09
Fix possibility of e2e-update-cards-only-test failure

calherries:
This PR fixes the possibility that `e2e-update-cards-only-test` fails because there are other revisions in the app DB with the same `model_id`.


markbastian (approved):
🚢 🇮🇹

PR #35547, opened 2023-11-09
Fix filter values handling

ranquild:
Fixes how `values` are handled in filter picker components. `values` have the correct length (per operator) in all cases now. E.g. for `>` there is always 1 value. No user noticeable changes, apart from us being able to save user input better when changing the operator (there are 2 tests changed because of this).

Also previously we used to store `undefined` values in a `number[]` array because of our TS configuration. This PR fixes this issue - now the type of the `values` is correct and we handle empty values explicitly. 

All the issues above made it hard to fine-tune working with values.

How to verify:
 - Unit tests are green


frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/CoordinateColumnPicker/CoordinateColumnPicker.tsx
```diff
@@ -0,0 +1,46 @@
...
+  stageIndex,
+  column,
+  availableColumns,
+  onChange,
+}: CoordinateColumnPickerProps) {
+  const options = useMemo(() => {
+    return getColumnOptions(query, stageIndex, availableColumns);
+  }, [query, stageIndex, availableColumns]);
+
+  const [value, setValue] = useState<string | null>(null);
```
kamilmielnik, line 25: Previously initial state was not `null`

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/utils.ts
```diff
@@ -1,103 +1,85 @@
...
+}
 
-  if (Lib.isLongitude(column)) {
-    return findLatitudeColumns(query, stageIndex)[0] ?? null;
+export function hasValidValues(
+  operator: Lib.CoordinateFilterOperatorName,
+  values: NumberValue[] = [],
+): values is number[] {
+  const { valueCount, hasMultipleValues } = OPERATOR_OPTIONS[operator];
+  if (!values.every(isNotEmpty)) {
```
kamilmielnik, line 28: It'd be easier to read without the 2 negations:

```suggestion
  if (values.some(isEmpty)) {
```

_This comment applies to other places with `isNotEmpty` function in this PR._

frontend/src/metabase/common/components/FilterPicker/TimeFilterPicker/utils.ts
```diff
@@ -1,42 +1,17 @@
-import dayjs from "dayjs";
-import type { TimeFilterOperatorName } from "metabase-lib";
+import type * as Lib from "metabase-lib";
 import { OPERATOR_OPTIONS } from "./constants";
 
-export function getDefaultValue() {
-  return dayjs().startOf("day").toDate(); // 00:00:00
+function getDefaultValue() {
+  return new Date(2020, 0, 1, 0, 0);
```
kamilmielnik, line 5: I think this should stay as it was.

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/CoordinateColumnPicker/CoordinateColumnPicker.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const placeholder = getColumnPlaceholder(column);
+
+  const handleChange = (value: string | null) => {
```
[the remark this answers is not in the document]
ranquild: The correct type is `null`, however out TS configuration allows invalid types in callbacks 
Screenshot 2023-11-13 at 12 55 17

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/utils.ts
ranquild: In this case I would need to have both `isEmpty` and `isNotEmpty`, so I'd rather keep it as it is

frontend/src/metabase/common/components/FilterPicker/TimeFilterPicker/utils.ts
ranquild: Ok, fixed

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/CoordinateColumnPicker/CoordinateColumnPicker.tsx
ranquild: It used an unstable hack to compute the initial state. Currently there is no correct way to do it, so I need to keep it as it is.

frontend/src/metabase/common/components/FilterPicker/BooleanFilterPicker/BooleanFilterPicker.tsx
```diff
@@ -40,7 +39,10 @@ export function BooleanFilterPicker({
   }, [options, isExpanded]);
 
   const handleOptionChange = (type: string) => {
-    setOptionType(type as OptionType);
+    const option = options.find(option => option.type === type);
+    if (option) {
+      setOptionType(option.type);
+    }
```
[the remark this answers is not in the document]
ranquild: Fixed

frontend/src/metabase/common/components/FilterPicker/CoordinateFilterPicker/CoordinateColumnPicker/CoordinateColumnPicker.tsx
ranquild: Set the initial value

PR #35565, opened 2023-11-09
[MLv2] `suggested-join-condition` => `suggested-join-conditions` and handle composite keys

### LINKED ISSUE #34184: Composite join keys aren't auto-populated when joining in notebook editor
### Describe the bug

I have table A with 2 fields (foreign/composite keys) mapped to another table B with composite entity keys (2 fields). When joining in notebook editor, it would automatically populate the first field as the join key but not he 2nd pair. 

This is the [recommended way to join Snowplow event tables](https://docs.snowplow.io/docs/understanding-your-pipeline/canonical-event/#extensible-schema) -- using 2 keys -- for accuracy and performance.

Another problem is if the entity key is a tstamp field (the case for Snowplow's recommended join), it will get truncated to minute or another interval when joining as there is no option to use the exact value, which makes it useless for joining. There are probably use cases for truncating the tstamp when joining but there needs to be an option to not do it, which should be the case if the field is set to entity key. I have to work around that by setting the tstamp field to foreign key which doesn't have the bug (but then it doesn't allow other foreign keys to map to it). 

### To Reproduce

1. Map 2 fields as foreign keys to another table with 2 entity keys. 
2. Try to join in notebook editor and notice editor only populated the first join key automatically

### Expected behavior

Both (all mapped foreign keys) keys should automatically be populated

### Logs

n/a

```JSON
Master / Postgres
```

### Severity

Without auto-populate, users would have to manually set the 2 keys, which is not possible for new users that don't know the 2 keys and a bit cumbersome to have to do an extra join

### Additional context

_No response_

camsaul:
Fixes #34184

1. `suggested-join-condition` is now `suggested-join-conditions` and can return multiple conditions. This change actually makes the API a little more convenient since `join-clause` takes *conditions* rather than a single *condition*.
2. Improvement: `suggested-join-conditions` only needs columns to be marked as FKs in order to suggest a condition. The target column no longer needs to be marked as a PK as well. (Not sure if there are any bugs related to this, but if there are they're fixed here as well.)
3. In situations where we have multiple FKs targeting different columns in the destination, we suggest multiple join conditions for each FK => column pair.
4. For situations where we have multiple FKs targeting the same column, we only suggest a join condition for one or the other. An example is if we have a Table like

   ```sql
   messages (sender_id REFERENCES user(id),  recipient_id REFERENCES user(id))
   ```

   we don't want to generate `LEFT JOIN user ON sender_id = user.id AND recipient_id = user.id`, because that doesn't really make sense. So this PR goes above and beyond what was asked in #34184 and does things in a smart way.


frontend/src/metabase-lib/join.ts
```diff
@@ -196,12 +196,12 @@ export function joinConditionOperators(
...
 }
 
-export function suggestedJoinCondition(
+export function suggestedJoinConditions(
   query: Query,
   stageIndex: number,
   joinable: Joinable,
-): JoinCondition | null {
-  return ML.suggested_join_condition(query, stageIndex, joinable);
+): JoinCondition[] | null {
```
ranquild, line 203: Wdyt about returning an empty array when there are no suggested conditions and not null?

camsaul: done

src/metabase/lib/join.cljc
```diff
@@ -751,67 +752,75 @@
...
-  (when pk-col
-    (let [visible-columns (lib.metadata.calculation/visible-columns query stage-number (lib.util/query-stage query stage-number))]
-      (fk-column-for-pk-in pk-col visible-columns))))
-
-(mu/defn suggested-join-condition :- [:maybe ::lib.schema.expression/boolean] ; i.e., a filter clause
-  "Return a suggested default join condition when constructing a join against `joinable`, e.g. a Table, Saved
-  Question, or another query. A suggested condition will be returned if the query stage has a foreign key to the
+   source
+   target]
+  (let [target-columns (delay
```
snoe, line 771: Why delay here? Should we be wary of `visible-columns` calls elsewhere?

snoe (approved):

src/metabase/lib/join.cljc
camsaul: mostly just to avoid calculating it if unneeded. Calculating it isn't free, so it's just a performance optimization

PR #35579, opened 2023-11-10
Fix newly created E2E CSV flakes

nemanjaglumac:
After battling with sync problems caused by MySQL, @iethree found a potential solution in the form of https://github.com/metabase/metabase/pull/35498.

tl;dr
Instead of throwing an error if API doesn't report that the sync is complete, we move on and hope that the sync will complete by the time tests start running. I've noticed the edge case that affects some tests - CSV uploads being one such case.

image

If you look at the UI from the recording, you'll notice that the sync status is still ongoing when the test starts. Cypress was implicitly failing because `findByRole` expects to find a single element. The easy fix here is to use `findAllByRole`. It doesn't compromise the test logic whatsoever.

Example of a failed run:
https://www.deploysentinel.com/ci/runs/654d28c67bcc148681663e5b

## Testing
I spun up three separate stress-test runs in order to increase chances of sync not finishing and having two "status" UI elements. All three are green and all of them individually burn through each affected test five times.
- https://github.com/metabase/metabase/actions/runs/6824845132/job/18561550975
- https://github.com/metabase/metabase/actions/runs/6824847605/job/18561559995
- https://github.com/metabase/metabase/actions/runs/6824850012/job/18561567034

Because we cannot rely on CI to reliably reproduce this edge case, I did a series of local testing by hard coding another "status" element on this page. This should simulate what's going on in CI when sync doesn't finish.
image

The final solution with using the `.last()` selector to isolate only one status element (see https://github.com/metabase/metabase/commit/ccbc68a20a8a04a49448e60ccb73f1b812871b40) makes all tests reliably pass regardless of the total number of elements.

### DIFF
--- a/e2e/test/scenarios/collections/uploads.cy.spec.js
+++ b/e2e/test/scenarios/collections/uploads.cy.spec.js
@@ -263,13 +263,13 @@ function uploadFile(testFile, valid = true) {
     );
   });
 
-  cy.findByRole("status").within(() => {
+  cy.findAllByRole("status").within(() => {
     cy.findByText(/Uploading/i);
     cy.findByText(testFile.fileName);
   });
 
   if (valid) {
-    cy.findByRole("status").within(() => {
+    cy.findAllByRole("status").within(() => {
       cy.findByText("Data added to Uploads Collection", {
         timeout: 10 * 1000,
       });
@@ -281,14 +281,14 @@ function uploadFile(testFile, valid = true) {
       cy.findByText(testFile.humanName);
     });
 
-    cy.findByRole("status").within(() => {
+    cy.findAllByRole("status").within(() => {
       cy.findByText("Start exploring").click();
     });
 
     cy.url().should("include", `/model/`);
     cy.findByTestId("TableInteractive-root");
   } else {
-    cy.findByRole("status").within(() => {
+    cy.findAllByRole("status").within(() => {
       cy.findByText("Error uploading your File");
     });
   }
--- a/e2e/test/scenarios/collections/uploads.cy.spec.js
+++ b/e2e/test/scenarios/collections/uploads.cy.spec.js
@@ -266,14 +266,18 @@ function uploadFile(testFile, valid = true) {
   // After #35498 has been merged, we now sometimes encounter two elements with the "status" role in UI.
   // The first (older) one is related to the sync that didn't finish, and the second one is related to CSV upload.
   // This is the reason we have to start using `findAllByRole` rather than `findByRole`.
+  // Since CSV status element is newer, we can and must use `.last()` to yield only one element within we perform the search.
   cy.findAllByRole("status")
+    .last()
     .should("contain", "Uploading data to")
     .and("contain", testFile.fileName);
 
   if (valid) {
-    cy.findAllByRole("status").findByText("Data added to Uploads Collection", {
-      timeout: 10 * 1000,
-    });
+    cy.findAllByRole("status")
+      .last()
+      .findByText("Data added to Uploads Collection", {
+        timeout: 10 * 1000,
+      });
 
     cy.get("main").within(() => cy.findByText("Uploads Collection"));
 
@@ -281,12 +285,12 @@ function uploadFile(testFile, valid = true) {
       cy.findByText(testFile.humanName);
     });
 
-    cy.findAllByRole("status").findByText("Start exploring").click();
+    cy.findAllByRole("status").last().findByText("Start exploring").click();
 
     cy.url().should("include", `/model/`);
     cy.findByTestId("TableInteractive-root");
   } else {
-    cy.findAllByRole("status").findByText("Error uploading your File");
+    cy.findAllByRole("status").last().findByText("Error uploading your File");
   }
 }
 
--- a/e2e/test/scenarios/collections/uploads.cy.spec.js
+++ b/e2e/test/scenarios/collections/uploads.cy.spec.js
@@ -263,16 +263,16 @@ function uploadFile(testFile, valid = true) {
     );
   });
 
-  cy.findAllByRole("status").within(() => {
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/collections/uploads.cy.spec.js
```diff
@@ -263,13 +263,13 @@ function uploadFile(testFile, valid = true) {
     );
   });
 
-  cy.findByRole("status").within(() => {
+  cy.findAllByRole("status").within(() => {
```
WiNloSt, line 266: I don't think this will work. When it's flake you'll get 2 elements and `within` would fail.

I've tested the similar command with
```ts
      cy.findAllByRole("link").within(() => {});
```

And got this error
```
cy.within() can only be called on a single element. Your subject contained 6 elements. Narrow down your subject to a single element (using .first(), for example) before calling .within().

To run .within() over multiple subjects, use .each().

cy.get('div').each($div => {
  cy.wrap($div).within(() => { ... })
})[Learn more](https://on.cypress.io/within)
```
image

nemanjaglumac: Stress-test reveals 20/20 passes for all three affected tests
https://github.com/metabase/metabase/actions/runs/6823741815/job/18558232573#step:9:197

Let me double and triple check things locally.

nemanjaglumac: @WiNloSt you're absolutely right. Thank you for catching this.
I've found a workaround and applied it in https://github.com/metabase/metabase/pull/35579/commits/054d24a91b933dfd88c3df85817cd383dad0c165. Mind taking another look? I'll do a few more stress-tests in CI in meantime.

e2e/test/scenarios/collections/uploads.cy.spec.js
```diff
@@ -281,16 +281,12 @@ function uploadFile(testFile, valid = true) {
...
-    });
+    cy.findAllByRole("status").findByText("Start exploring").click();
 
     cy.url().should("include", `/model/`);
     cy.findByTestId("TableInteractive-root");
   } else {
-    cy.findByRole("status").within(() => {
-      cy.findByText("Error uploading your File");
-    });
+    cy.findAllByRole("status").findByText("Error uploading your File");
```
nemanjaglumac, line 289: There is absolutely no reason to use `.within()` all over the place. The only valid use case for it is when you need to search for or assert on multiple different elements within one parent. Otherwise, `.findByText` is scoped to the previously yielded element(s).

cc @iethree

e2e/test/scenarios/collections/uploads.cy.spec.js
```diff
@@ -263,16 +263,16 @@ function uploadFile(testFile, valid = true) {
...
-  cy.findByRole("status").within(() => {
-    cy.findByText(/Uploading/i);
-    cy.findByText(testFile.fileName);
-  });
+  // After #35498 has been merged, we now sometimes encounter two elements with the "status" role in UI.
+  // The first (older) one is related to the sync that didn't finish, and the second one is related to CSV upload.
+  // This is the reason we have to start using `findAllByRole` rather than `findByRole`.
+  cy.findAllByRole("status")
+    .should("contain", "Uploading data to")
+    .and("contain", testFile.fileName);
```
WiNloSt, line 271: I tested and this command should be fine.

e2e/test/scenarios/collections/uploads.cy.spec.js
```diff
@@ -281,16 +281,12 @@ function uploadFile(testFile, valid = true) {
       cy.findByText(testFile.humanName);
     });
 
-    cy.findByRole("status").within(() => {
-      cy.findByText("Start exploring").click();
-    });
+    cy.findAllByRole("status").findByText("Start exploring").click();
```
WiNloSt, line 284: This one (and possibly the one above) doesn't work when testing locally.

I've tested with this simple test

```ts
      cy.signInAsAdmin();

      cy.visit("/");
      cy.findByText("Our analytics").should("be.visible");
      cy.findAllByRole("link")
        .should("contain", "Our analytics")
        .and("contain", "Your personal collection");
      // here Cypress detects 5-6 links in my case, but it seems to try to find `Our analytics` only on the first found link
      // it doesn't seem to iterate through all the possibilities to find the matching text.
      cy.findAllByRole("link").findByText("Our analytics").click();
```

I found a (hacky?) solution that could work
```ts
      cy.signInAsAdmin();

      cy.visit("/");
      cy.findByText("Our analytics").should("be.visible");
      cy.findAllByRole("link")
        .should("contain", "Our analytics")
        .and("contain", "Your personal collection");
      cy.findAllByRole("link").each(([$link]) => {
        if ($link.innerText.includes("Our analytics")) {
          cy.wrap($link).as("ourAnalyticsLink");
        }
      });
      cy.get("@ourAnalyticsLink").should("be.visible").click();
```

This works by iterating over all `link` (`status` in your case), and trying to find the element using `innerText` since we can't use any Cypress command to find an element because we don't want the element that doesn't contain our target text to fail as `find*By*` command would fail immediately when the element is not found.

After we found the element we put it in an alias `cy.wrap().as(alias)` then refer to that later on.

You could try to adjust this to your situation. Although this solution isn't elegant, I couldn't think of any other way to achieve this yet as Cypress doesn't [encourage conditional testing.](https://docs.cypress.io/guides/core-concepts/conditional-testing)

nemanjaglumac: Thanks for the analysis once again.
This should be slightly less verbose hack that should work in all cases.
https://github.com/metabase/metabase/pull/35579/commits/ccbc68a20a8a04a49448e60ccb73f1b812871b40

WiNloSt (approved):
Loos good to me now 👍

PR #35585, opened 2023-11-10
Fix outdated serialization references in docs/comments

calherries:
There were a few references to `metabase.models.serialization.base`, which is now `metabase.models.serialization`.

Also `metabase.models.serialization.base/serdes-dependencies` is now `metabase.models.serialization/dependencies`.


sansolovyov (approved):
👍

PR #19267, opened 2021-12-08
Support overriding ROWCOUNT for SQL Server

jeff303:
Add new driver multimethod called `row-limit-override` which, given a `driver` and `database`, can optionally return a row-limit to be applied to a query (instead of the default value)

Implement that multimethod for :sqlserver by looking up a new (optional) connection property named `:rowcount-override`

Marking `sql-jdbc.execute/statement-supported?` as `false` for `:sqlserver` (since there is an error when executing native queries like the type in #9940 using a `Statement`); the driver will always use `PreparedStatement` now

Move common computation function of overall row limit to the limit namespace, and invoke it from execute now, called `determine-query-max-rows`

Add new clause to the `determine-query-max-rows` function that preferentially takes the value from `row-limit-override` (if defined)

Adding new test case to confirm that `:sqlserver` takes the new row limit when executing a native query that has intermediate temp results

PR #35600, opened 2023-11-10
make max results configurable

### LINKED ISSUE #35428: Make `max-results` configurable for aggregated and non-aggregated queries
**Context**
https://github.com/metabase/metabase/issues/12729
https://github.com/metabase/metabase/issues/4564
https://github.com/metabase/metabase/issues/8622
https://github.com/metabase/metabase/issues/8322
https://github.com/metabase/metabase/issues/7622

Prior PR
https://github.com/metabase/metabase/pull/32401/files

Product doc: [link](https://www.notion.so/metabase/Increase-max-results-default-value-and-enable-configuration-using-env-vars-ab5ef784175a40629650afed033bf8df)

**Desired behavior**
If the admin doesn't change anything, we should apply the defaults:
- aggregated queries return up to 10000 rows
- unaggregated queries return up to 2000 rows

If the admin wants, they can change the following env var and change the limit:
- `MB_UNAGGREGATED_QUERY_ROW_LIMIT`. This value must be equal or less than the aggregated limit. This env var replaces `MB_MAX_RESULTS_BARE_ROWS`
- `MB_AGGREGATED_QUERY_ROW_LIMIT`. This must be less than [absolute-max-results](https://github.com/metabase/metabase/blob/master/src/metabase/query_processor/interface.clj).

johnswanson:
This replaces the existing setting `max-results-bare-rows` with two settings:

- `max-unaggregated-query-row-limit`, and
- `max-aggregated-query-row-limit`

Fixes #35428

### DIFF
--- a/src/metabase/query_processor/middleware/constraints.clj
+++ b/src/metabase/query_processor/middleware/constraints.clj
@@ -5,33 +5,46 @@
    [metabase.models.setting :as setting]
    [metabase.util.i18n :refer [deferred-tru]]))
 
-(def ^:private ^:const default-max-results-bare-rows 2000)
+(def ^:private ^:const default-max-unaggregated-query-row-limit 2000)
+(def ^:private ^:const default-max-aggregated-query-row-limit 10000)
 
-;; NOTE: this was changed from a hardcoded var with value of 2000 (now moved to [[default-max-results-bare-rows]])
+;; NOTE: this was changed from a hardcoded var with value of 2000 (now moved to [[default-max-unaggregated-query-row-limit]])
 ;; to a setting in 0.43 the setting, which allows for DB local value, can still be nil, so any places below that used
 ;; to reference the former constant value have to expect it could return nil instead
-(setting/defsetting max-results-bare-rows
+(setting/defsetting max-unaggregated-query-row-limit
   (deferred-tru "Maximum number of rows to return specifically on :rows type queries via the API.")
   :visibility     :authenticated
   :type           :integer
   :database-local :allowed
   :audit          :getter)
 
-(def ^:private max-results
-  "General maximum number of rows to return from an API query."
-  10000)
+(setting/defsetting max-aggregated-query-row-limit
+  (deferred-tru "Maximum number of rows to return for aggregated queries via the API.")
+  :visibility     :authenticated
+  :type           :integer
+  :database-local :allowed
+  :audit          :getter)
+
+(defn query->max-rows
+  "Given a query, returns the max rows that should be returned *as defined by settings*. In other words,
+  return `(max-aggregated-query-row-limit)` or `(max-unaggregated-query-row-limit)` depending on whether the query is
+  aggregated or not."
+  [{{aggregations :aggregation} :query}]
+  (if-not aggregations
+    (max-unaggregated-query-row-limit)
+    (max-aggregated-query-row-limit)))
 
 (defn default-query-constraints
   "Default map of constraints that we apply on dataset queries executed by the api."
   []
-  {:max-results           max-results
-   :max-results-bare-rows (or (max-results-bare-rows) default-max-results-bare-rows)})
+  {:max-results           (or (max-aggregated-query-row-limit) default-max-aggregated-query-row-limit)
+   :max-results-bare-rows (or (max-unaggregated-query-row-limit) default-max-unaggregated-query-row-limit)})
 
 (defn- ensure-valid-constraints
-  "`:max-results-bare-rows` must be less than or equal to `:max-results`, so if someone sets `:max-results` but not
-  `:max-results-bare-rows` or sets an both but sets an invalid value for ``:max-results-bare-rows` use the same value
-  for both. Otherwise the default bare rows value could end up being higher than the custom `:max-rows` value, causing
-  an error."
+  "`:max-unaggregated-query-row-limit` must be less than or equal to `:max-results`, so if someone sets `:max-results`
+  but not `:max-unaggregated-query-row-limit` or sets an both but sets an invalid value for
+  ``:max-unaggregated-query-row-limit` use the same value for both. Otherwise the default bare rows value could end up
+  being higher than the custom `:max-rows` value, causing an error."
   [{:keys [max-results max-results-bare-rows], :as constraints}]
   (if (<= max-results-bare-rows max-results)
     constraints
--- a/src/metabase/query_processor/middleware/constraints.clj
+++ b/src/metabase/query_processor/middleware/constraints.clj
@@ -5,6 +5,22 @@
    [metabase.models.setting :as setting]
    [metabase.util.i18n :refer [deferred-tru]]))
 
+;; The following "defaults" are not applied to the settings themselves - why not? Because the existing behavior is
+;; that, if you manually update the settings, queries are affected *WHETHER OR NOT* the
+;; `add-default-userland-constraints` middleware was applied.
+;;
+;; To achieve this, the QP looks for the following, in order:
+;; 1. a non-nil value set by the `add-default-userland-constraints` middleware below, either:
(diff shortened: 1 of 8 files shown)


luizarakaki:
@johnswanson Not sure if this is what you meant, but

Downloads shouldn't use these settings. They follow other limits:
`absolute-max-results`
https://github.com/metabase/metabase/blob/10d99743ce56335a62cf2a60133bc55cc8b03ef0/src/metabase/query_processor/interface.clj#L7-L15

Or 10k, if they are constrained by the permission system
https://github.com/metabase/metabase/blob/10d99743ce56335a62cf2a60133bc55cc8b03ef0/enterprise/backend/src/metabase_enterprise/advanced_permissions/query_processor/middleware/permissions.clj#L13

src/metabase/query_processor/middleware/constraints.clj
```diff
@@ -5,33 +5,46 @@
...
-(def ^:private ^:const default-max-results-bare-rows 2000)
+(def ^:private ^:const default-max-unaggregated-query-row-limit 2000)
+(def ^:private ^:const default-max-aggregated-query-row-limit 10000)
 
-;; NOTE: this was changed from a hardcoded var with value of 2000 (now moved to [[default-max-results-bare-rows]])
+;; NOTE: this was changed from a hardcoded var with value of 2000 (now moved to [[default-max-unaggregated-query-row-limit]])
 ;; to a setting in 0.43 the setting, which allows for DB local value, can still be nil, so any places below that used
 ;; to reference the former constant value have to expect it could return nil instead
-(setting/defsetting max-results-bare-rows
+(setting/defsetting max-unaggregated-query-row-limit
```
noahmoss, line 14: Unfortunately it's not this simple to change a setting's name. If the setting is already set on any instances, changing the name will essentially "unset" it since the `setting` table uses the setting's name as the key. In addition, since this has `:database-local :allowed`, it might have additional DB-specific values in the `metabase_database.settings` JSON field. (Looks like this could be the case for [SQLServer](https://github.com/metabase/metabase/pull/19267), not sure about others)

If we absolutely need a new env var, we either need
* A migration to change the setting name anywhere it might already exist in the app DB (and a rollback, preferably)
* Or, a new setting with a custom getter which migrates the value of `max-results-bare-rows` when it's fetched

noahmoss: It also might be a product question if we want to continue allowing this setting to be set at the DB level, and what should happen if it's _also_ set via env var. The way database-local settings currently work is that they'll return the DB-local value preferentially to the env var value.

noahmoss (changes requested):
Requesting changes because we need to take into account previous values of the `max-results-bare-rows` settings. 

Could you also request a review from the QP/drivers team? This change is more in their scope, tbh.

src/metabase/query_processor/middleware/constraints.clj
johnswanson: @noahmoss backwards compatibility was my thought too when I picked this up. When I talked to @luizarakaki about it, it sounded to me like given the fact that the current setting doesn't really work, we are confident that it's not being used and that we can remove it without fear: https://metaboat.slack.com/archives/C064EB1UE5P/p1699472623388119

noahmoss: Oh okay. In that case I'll stamp. 

I'm still a little concerned about the SQLServer case since that's the one database where it seems like there's an explicit mechanism to change the row limit when you're setting up the DB

noahmoss (approved):

src/metabase/query_processor/middleware/constraints.clj
```diff
@@ -5,33 +5,46 @@
...
 
 (defn- ensure-valid-constraints
-  "`:max-results-bare-rows` must be less than or equal to `:max-results`, so if someone sets `:max-results` but not
-  `:max-results-bare-rows` or sets an both but sets an invalid value for ``:max-results-bare-rows` use the same value
-  for both. Otherwise the default bare rows value could end up being higher than the custom `:max-rows` value, causing
-  an error."
+  "`:max-unaggregated-query-row-limit` must be less than or equal to `:max-results`, so if someone sets `:max-results`
+  but not `:max-unaggregated-query-row-limit` or sets an both but sets an invalid value for
+  ``:max-unaggregated-query-row-limit` use the same value for both. Otherwise the default bare rows value could end up
+  being higher than the custom `:max-rows` value, causing an error."
```
metamben, line 47: I'm not sure this is clearer than original comment.

src/metabase/query_processor/middleware/constraints.clj
```diff
@@ -5,33 +5,46 @@
...
 
-(def ^:private max-results
-  "General maximum number of rows to return from an API query."
-  10000)
+(setting/defsetting max-aggregated-query-row-limit
+  (deferred-tru "Maximum number of rows to return for aggregated queries via the API.")
+  :visibility     :authenticated
+  :type           :integer
+  :database-local :allowed
+  :audit          :getter)
```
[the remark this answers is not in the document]
johnswanson: It is confusing - I'll add comments to explain. This could possibly use a refactor, but we want to distinguish between:

- the user **HAS** set a value and the constraints middleware **WAS** applied: we use the specified value
- the user **HAS NOT** set a value and the constraints middleware **WAS** applied: we use the "default" value
- the user **HAS** set a value and the constraints middleware **WAS NOT** applied: we use the specified value
- the user **HAS NOT** set a value and the constraints middleware **WAS NOT** applied: we use the "absolute max results" value

The current code works by looking for, in order:
1. a non-nil value set by the constraints **middleware** (so the value of the setting, the "default", or `nil` if no constraint middleware)
2. a non-nil value set by the constraints **setting** (so the value of the setting or `nil` if it wasn't set)
3. the "absolute max results"

I didn't want to mess too much with QP code, so I left the logic pretty much as it was. The only real change was which setting we're looking at in step 2.

src/metabase/query_processor/middleware/constraints.clj
johnswanson: I didn't change the comment itself, only changed `max-results-bare-rows` to `max-unaggregated-query-row-limit`. I'm happy to reword it though, I find the comment a bit confusing too.

metamben: Yes, so it's referring to things that don't appear in the code. I think the connection between these things should be explained.

johnswanson: Oh, great catch - sorry, I don't know how I missed that.

src/metabase/query_processor/middleware/constraints.clj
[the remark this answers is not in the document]
johnswanson: I added a comment [here](https://github.com/metabase/metabase/pull/35600/files#diff-3b2ac9a54646c3fc99a8239c4ae27cbc1478f01837a6e423c6a8b96a1a32c6e5R8-R22) with a better explanation for why we aren't applying defaults to the settings themselves.

src/metabase/query_processor/middleware/constraints.clj
johnswanson: After taking a step back, I just reworded the comment to ["Clamps the value of `max-results-bare-rows` to be less than or equal to the value of `max-results`."](https://github.com/metabase/metabase/pull/35600/files#diff-3b2ac9a54646c3fc99a8239c4ae27cbc1478f01837a6e423c6a8b96a1a32c6e5R60)

I don't *think* it's necessary to point out the names of the settings those values actually originate from (and the code pulling them from the settings is in the same short file, so it's pretty easy to trace), but let me know if you'd rather include that elaboration and I'll add that in.

metamben (approved):

PR #35602, opened 2023-11-10
Update Babel and MySQL2 Dependencies

### LINKED ISSUE #33687: move mysql2 to devdependencies
Seems there's a package we installed that should go to the devdependencies the same way as pg is as it's only used in tests

iethree:
Closes https://github.com/metabase/metabase/issues/33687

### Description

- Move mysql2 to dev dependencies and minor version update
- Update all babel-related dependencies to the  latest version


paoliniluis:
@ryancw don't we need that mysql2 package for actually doing the cypress e2e tests? I thought we had that and the postgres one just because of that

iethree:
> @ryancw don't we need that mysql2 package for actually doing the cypress e2e tests? I thought we had that and the postgres one just because of that

correct. we use it in e2e tests. that's why i moved to to devDependencies

uladzimirdev:
@iethree what about other babel deps?

iethree:
> @iethree what about other babel deps?

sure why not

uladzimirdev (approved):

Issue #16745, opened 2021-06-23
`Between` filter errors out with min > max 


nemanjaglumac:
**Describe the bug**
Before #16728 it wasn't possible to use `between` filter for longitude/latitude at all. This is already a huge improvement. However, I've run into a few situations that result in an error. Still trying to find the root cause.

**Logs**
`not,Integer greater than zero, X` where `X` is some negative number

**To Reproduce**
1. Open "People" table
2. Click "Summarize" (will open a right sidebar)
3. Open popover with binning options for "Longitude"
4. Choose "Bin every 10 degrees" (or any other specific value)
5. Visualization should render _bars_
6. Click on any of the bars, choose "Zoom in" (to drill through)
7. You should see granularity change to 1 degree and the filter saying `Longitude -100 -90`
8. Click on that filter. You should see this
![image](https://user-images.githubusercontent.com/31325167/123151849-2d7e4480-d464-11eb-82ad-ac891cee0a77.png)
9. Replace -100 (W) with positive value 100 (E)
![image](https://user-images.githubusercontent.com/31325167/123151958-5272b780-d464-11eb-9bd1-779e3206e75b.png)
10. Click "Update filter"
11. Observe an error
![image](https://user-images.githubusercontent.com/31325167/123152073-78985780-d464-11eb-8e38-1d8795eb12a3.png)

console error
```
VisualizationError.jsx:45 error (2) ["not", Array(2)]0: "not"1: (2) ["Integer greater than zero", -190]length: 2__proto__: Array(0)
```

**Expected behavior**
Between filter should work regardless of the range.

**Information about your Metabase Installation:**
local dev, `master`, f3847cb8fd9025f20f39822a073a8c9bf7db3373

**Severity**
?

**Additional context**
I couldn't reproduce it by using filter initially. It _seems_ to only happens after binning, but the weird part is that if you try to use regular filter after that, it will fail as well.

flamber:
You get similar error by doing something like this:
![image](https://user-images.githubusercontent.com/1447303/123154508-3d4b5800-d467-11eb-828f-b4873c4ba6b5.png)

<details><summary>Full stacktrace</summary>

```
2021-06-23 21:16:10,056 ERROR middleware.catch-exceptions :: Error processing query: null
{:database_id 4,
 :started_at #t "2021-06-23T21:16:09.466708+02:00[Europe/Copenhagen]",
 :via
 [{:status :failed,
   :class clojure.lang.ExceptionInfo,
   :error
[...]

PR #35609, opened 2023-11-10
Swap min > max values in filters

ranquild:
Fixes https://github.com/metabase/metabase/issues/16745

How to verify:
- Orders -> Total -> Between
- Orders -> Created at -> Between
- People > Latitude -> Between
- People > Latitude > Inside
- Native Query -> `SELECT CAST('10:20' AS TIME) TIME` -> Save -> Expore results -> Notebook -> Filter -> Time -> Between


frontend/src/metabase/common/components/DatePicker/SpecificDatePicker/SpecificDatePicker.unit.spec.tsx
```diff
@@ -38,11 +38,11 @@ describe("SpecificDatePicker", () => {
   });
 
   it('should be able to set "on" filter', () => {
-    const { onChange } = setup();
+    const { onChange } = setup({ isNew: true });
```
kamilmielnik, line 41: `setup` is always used with `isNew: true` - we could just change the default value, or remove the parameter

kamilmielnik (approved):
Looks good, works well :+1:

PR #35616, opened 2023-11-10
Fix public link E2E flakes

nemanjaglumac:
This PR fixes two flakes with the same root cause, and changes the pattern from trying to get the UUID from the UI to reading it directly from the API response that's already there.

Examples of failed runs:
- https://www.deploysentinel.com/ci/runs/654e0bad9f1ec992dfc50fe1
- https://www.deploysentinel.com/ci/runs/654e12702beaab03257ef532

For both flakes, the input field shows `null` in the UI when we tried to grab it.
image

Results for the last 7 days:
- 20393: failure rate 0%, flake rate 7.94%
- 22524: failure rate 0%, flake rate 11.11%

### Stress-test
I ran both tests thirty times and they passed every single time.

- https://github.com/metabase/metabase/actions/runs/6828545400/job/18572897600 (20393)
- https://github.com/metabase/metabase/actions/runs/6828548175/job/18572905651 (22524)

### DIFF
--- a/e2e/test/scenarios/sharing/reproductions/20393-public-dashboard-nested-card-with-parameters.cy.spec.js
+++ b/e2e/test/scenarios/sharing/reproductions/20393-public-dashboard-nested-card-with-parameters.cy.spec.js
@@ -2,6 +2,8 @@ import { restore, popover, visitDashboard } from "e2e/support/helpers";
 
 describe("issue 20393", () => {
   beforeEach(() => {
+    cy.intercept("POST", "/api/dashboard/*/public_link").as("publicLink");
+
     restore();
     cy.signInAsAdmin();
   });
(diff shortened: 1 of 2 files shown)


e2e/test/scenarios/sharing/reproductions/20393-public-dashboard-nested-card-with-parameters.cy.spec.js
```diff
@@ -28,14 +30,12 @@ describe("issue 20393", () => {
...
-      .parent()
-      .within(() => {
-        cy.get("input").then(input => {
-          cy.visit(input.val());
-        });
-      });
+    cy.wait("@publicLink").then(({ response: { body } }) => {
+      const { uuid } = body;
+
+      cy.signOut();
```
WiNloSt, line 36: 👍 +1 for signing out before testing the public embed as well.

WiNloSt (approved):
Tested locally. This works great.

On master, I can 100% make these tests fail by throttling CPU to 6x slower and network to fast 3G using Chrome.

But this PR passes both tests.

PR #35617, opened 2023-11-10
Refine default CI triggers

nemanjaglumac:
This is the last step for [#34880](https://github.com/metabase/metabase/issues/34880).
We don't need the global `*ci` anymore as no other glob uses it.

The only other difference is that the `default` glob set now ignores markdown files in the `bin/` folder.

### DIFF
--- a/.github/file-paths.yaml
+++ b/.github/file-paths.yaml
@@ -1,9 +1,6 @@
 default: &default
   - "locales/**"
-  - "bin/**"
-
-ci: &ci
-  - ".github/**/!(*.md|team.json)"
+  - "bin/**/!(*.md)"
 
 shared_sources: &shared_sources
   - "shared/src/**"


.github/file-paths.yaml
```diff
@@ -1,9 +1,6 @@
 default: &default
   - "locales/**"
-  - "bin/**"
-
-ci: &ci
-  - ".github/**/!(*.md|team.json)"
+  - "bin/**/!(*.md)"
```
nemanjaglumac, line 3: Glob test: https://www.digitalocean.com/community/tools/glob?comments=true&glob=bin%2F%2A%2A%2F%21%28%2A.md%29&matches=false&tests=bin%2Fbuild%2Fsrc%2Fbuild_driver.clj&tests=bin%2Fbuild-drivers.md&tests=bin%2Fclear-outdated-cpcaches.sh

alxnddr (approved):

PR #35619, opened 2023-11-10
Isomorphic timeline events viz 34924

### LINKED ISSUE #34924: Backend for static viz timeline events
Base branch for this change is `unified-isomorphic-static-viz-be`, base branch PR: https://github.com/metabase/metabase/pull/34820

**Context**

Backend needs to provide Timeline Events data to static visualizations so that they can render them.

**Spec**

Add `timeline_events` parameter to the [isomorphic chart renderer](https://github.com/metabase/metabase/pull/34820/files#diff-fb611a0a0fdecfa916aa28ed0679a400ec86ca9cd303d45b64069827ee1b1304R133) and pass it to the javascript function. It should be an array of raw timeline events that should be displayed for the visualization.

```
(defn isomorphic
  "Clojure entrypoint to render isomorphic visualizations."
  [cards-with-data dashcard-viz-settings timeline-events]
  (let [svg-string (.asString (js/execute-fn-name (context) "isomorphic"
                                                  (json/generate-string cards-with-data)
                                                  (json/generate-string dashcard-viz-settings)
                                                  (json/generate-string (public-settings/application-colors))
                                                  (json/generate-string (timeline-events))
))]
```

The example value of `timeline-events`. It is exactly the same shape of the `events` entities of `timeline` entities as returned by `/api/timeline?include=events`, and for this example I stipped the excessive data but it does not hurt if there are more properties such as creation date timestamp:
```
[
      {
          "archived": false,
          "name": "my event",
          "timeline_id": 1,
          "icon": "star",
          "id": 1,
      },
]
```

We do not show timeline events on dashcards so timeline events should be provided only for question alerts. The array should contain timeline events only for the questions's collection and only if users have access to it.

markbastian:
This primarily required adding logic as found in `/api/timeline?include=events` to `s/defmethod render :isomorphic :- common/RenderedPulseCard`. This is done by adding one extra accretion function, `add-dashcard-timeline-events` to the logic and making code that expected a dashboard also take a dashcard.

Resolves #34924

### DIFF
--- a/dev/src/dev/render_png.clj
+++ b/dev/src/dev/render_png.clj
@@ -2,16 +2,17 @@
   "Improve feedback loop for dealing with png rendering code. Will create images using the rendering that underpins
   pulses and subscriptions and open those images without needing to send them to slack or email."
   (:require
-   [clojure.java.io :as io]
-   [clojure.java.shell :as sh]
-   [hiccup.core :as hiccup]
-   [metabase.models.card :as card]
-   [metabase.models.user :as user]
-   [metabase.pulse :as pulse]
-   [metabase.pulse.render :as render]
-   [metabase.pulse.render.test-util :as render.tu]
-   [metabase.query-processor :as qp]
-   [metabase.query-processor.middleware.permissions :as qp.perms]))
+    [clojure.java.io :as io]
+    [clojure.java.shell :as sh]
+    [hiccup.core :as hiccup]
+    [metabase.models.card :as card]
+    [metabase.models.user :as user]
+    [metabase.pulse :as pulse]
+    [metabase.pulse.render :as render]
+    [metabase.pulse.render.test-util :as render.tu]
+    [metabase.query-processor :as qp]
+    [metabase.query-processor.middleware.permissions :as qp.perms]
+    [toucan2.core :as t2]))
 
 ;; taken from https://github.com/aysylu/loom/blob/master/src/loom/io.clj
 (defn- os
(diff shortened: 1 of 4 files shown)


dev/src/dev/render_png.clj
```diff
@@ -2,16 +2,17 @@
...
+    [clojure.java.shell :as sh]
+    [hiccup.core :as hiccup]
+    [metabase.models.card :as card]
+    [metabase.models.user :as user]
+    [metabase.pulse :as pulse]
+    [metabase.pulse.render :as render]
+    [metabase.pulse.render.test-util :as render.tu]
+    [metabase.query-processor :as qp]
+    [metabase.query-processor.middleware.permissions :as qp.perms]
+    [toucan2.core :as t2]))
```
adam-james-v, line 15: Good to actually have the proper require in here. Not sure how I missed that before.

src/metabase/pulse/util.clj
```diff
@@ -14,6 +14,7 @@
    [toucan2.core :as t2]))
 
 ;; TODO - this should be done async
+;; TODO - this and `execute-multi-card` should be made more efficient: eg. we query for the card several times
```
adam-james-v, line 17: I think that in order to make these execute fns more efficient we should change the contract to require passing in the instance, so that we fetch earlier and only once.

Not needed for this PR, but an idea anyway.

adam-james-v (approved):
These changes look good. Mark and I paired on some of this as well, so I'm familiar with the code added already.

PR #35621, opened 2023-11-10
Audit db sync cannot happen

escherize:
### We never want to schedule syncs the audit db. 

But somehow we mysteriously were.

This bug was caused when the `SyncDatabases` task, which helpfully syncs that all your `metabase_database`s are scheduled to be synced. We call the same function on a database.model post insert, and it passes in an entire audit db.

I did not know about the `SyncDatabases` task, and it would have been handled anyway, but the task passes in a database record with no `:is_audit` key (which is what we were using to block scheduling of the audit db.) Now we use `:id`, and see if it's equal to the audit db id. AFAIK: Everything passes in an id.

----

### What if still gets scheduled?

This PR Also makes the actual *sync operation be a no-op for the audit db*, and blow up when not `config/prod?`, so we can catch it if there's yet another way that it gets scheduled.


noahmoss (approved):

enterprise/backend/test/metabase_enterprise/audit_db_test.clj
```diff
@@ -73,3 +74,21 @@
   (is (= #{"plugins/instance_analytics/collections"
            "plugins/instance_analytics/databases"}
          (set (map str (fs/list-dir "plugins/instance_analytics"))))))
+
+(defn- get-audit-db-triggers []
+  (t2/query
```
dpsutton, line 79: don't use this level of trigger access. check out 
```clojure
task=> (->> (metabase.task/scheduler-info)
            :jobs
            (filter (comp #{"metabase.task.sync-and-analyze.job"} :key))
            first
            :triggers
            (map (comp clojurewerkz.quartzite.conversion/from-job-data :data)))
({"db-id" 1}
 {"db-id" 121}
 {"db-id" 122}
 {"db-id" 13371337}
 {"db-id" 785})
```

escherize: Make sense: I am having it use metabase.task/scheduler-info, instead of the qrtz table directly. `lojurewerkz.quartzite.conversion/from-job-data` didn't actually change anything, since that output is just the value at the `:data` key.

PR #35627, opened 2023-11-10
Remove locales.clj, use a single source of truth for the locales list

noahmoss:
I kept getting annoyed that `locales.clj` gets generated each time you build Metabase, and it was slightly different from the version that's checked into the repo, so you end up with a dirty git file. I was going to just check in the updated version of the file, but I actually don't think it's necessary at all — we can just derive the list of locales from the EDN files in `resources/i18n/`, since that's the authoritative truth, and then just keep the list in memory so that we're not constantly accessing the filesystem.


darksciencebase:
@noahmoss is this still on the table or are we closing it?

noahmoss:
@darksciencebase I thought this was going to be a quick fix but I was running into more issues than expected. I'll close it and open a new PR whenever I get back around to this

PR #35634, opened 2023-11-11
Update audit yamls new keyorder

luizarakaki:
This PR:
- updates the YAMLs with the latest Serdes version that implemented a new yaml key order
- adds the v_tasks view
- improved the content, namely added internal links and some extra questions


resources/migrations/000_migrations.yaml
```diff
@@ -16064,6 +16064,27 @@ databaseChangeLog:
             newDataType: VARCHAR(32)
       rollback: # not necessary
 
+  - changeSet:
+      id: v48.00-054
```
noahmoss: `v48.00-054` was claimed already (by Bryan) so the next available ID is 055. I just claimed it in #migrations
```suggestion
      id: v48.00-055
```

noahmoss (approved):

PR #35635, opened 2023-11-12
Add missing space on JSON unfolding

### LINKED ISSUE #28554: Typo in description for Unfold JSON columns database setting
## What
We're missing a space in the description below in regard to unfolding JSON columns in database settings 👇

Image

dragonsahead:
Fixes https://github.com/metabase/metabase/issues/28554

tagging @mazameli as it has an impact on translations

### DIFF
--- a/locales/ar-SA.po
+++ b/locales/ar-SA.po
@@ -28909,7 +28909,7 @@ msgstr "تكشف أعمدة JSON"
 
 #: metabase/driver/common.clj:180
 #, fuzzy
-msgid "We unfold JSON columns into component fields.This is on by default but you can turn it off if performance is slow."
+msgid "We unfold JSON columns into component fields. This is on by default but you can turn it off if performance is slow."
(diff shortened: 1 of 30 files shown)


mazameli (commented):
This is odd, because we typically would fix this by changing the actual tagged string itself, not the translated strings directly. But I can't find this exact string in the codebase anymore. The best guess I have is that this string has already been altered in `/metabase/driver/common.clj`:

```
(def json-unfolding
  "Map representing the `json-unfolding` option in a DB connection form"
  {:name         "json-unfolding"
   :display-name (deferred-tru "Allow unfolding of JSON columns")
   :type         :boolean
   :visible-if   {"advanced-options" true}
   :description  (deferred-tru
                   (str "This enables unfolding JSON columns into their component fields. "
                        "Disable unfolding if performance is slow. If enabled, you can still disable unfolding for "
                        "individual fields in their settings."))
   :default      true})
```

So unless we're trying to backport this fix into older versions of Metabase, I wouldn't bother with this PR, and would instead just let this new string, which includes proper spaces, to be translated via our regular process.

PR #35643, opened 2023-11-13
Make E2E workflow 3rd party friendly

nemanjaglumac:
This PR is a follow up after https://github.com/metabase/metabase/pull/34951.
It should be straight-forward, but I'm always wary when touching Github workflows as they are notoriously hard to test, so please be vigilant when reviewing.

Unfortunately, it is impossible to actually test this until it gets merged into `master`.
If people feel uneasy merging it without some test runs, I can set things up tomorrow in my own fork, and then test it using completely different account (fork of a fork).

## Test
I created a separate account which forked my own fork of Metabase 😅
Here's what I tested:
- Created a PR from external account -> had to wait for an approval
- Once approved, it started running
    - E2E on `pull_request` skipped ✅
    - `external-filer.yml` on `pull_request` recognized this was an external PR and passed which, in turn, triggered E2E tests on `workflow_run`, as designed ✅
- All tests ran and passed, meaning access to the secrets works
image

> **Important**
> - External contributor cannot hit re-run (only the owner can)! 🎉
> - Force-push, rebase, etc. cannot circumvent waiting for an approval! Each new commit has to be explicitly approved by owner/maintainer. 🎉 

### Downsides
- File path filtering doesn't work when we trigger the workflow via `workflow_run`. This means that all tests run even on unrelated file changes. Given that we don't have a high volume of external contributions this is a completely acceptable tradeoff IMO.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -7,14 +7,31 @@ on:
       - 'release-**'
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+  workflow_run:
+    workflows: [External Trigger Filter]
+    types: [completed]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
+  external-filter:
+    if: github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-22.04
+    timeout-minutes: 3
+    steps:
+      - run: |
+          if ${{ github.event.pull_request.head.repo.full_name == github.repository }}; then
+            echo 'Running E2E tests within the organization.'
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -7,14 +7,31 @@ on:
...
+      - run: |
+          if ${{ github.event.pull_request.head.repo.full_name == github.repository }}; then
+            echo 'Running E2E tests within the organization.'
+          elif ${{ github.event.workflow_run.conclusion == 'success' }}; then
+            echo "Running E2E tests for the external contributor. Thank you!"
+          fi
+
   files-changed:
     name: Check which files changed
+    needs: external-filter
```
nemanjaglumac, line 33: In theory, only these two, top-level, jobs need the conditional logic because all other jobs sequentially depend on them, and cannot execute if these two are skipped.

But as a safety measure, I don't think it hurts to add the same condition to every single job in a workflow.

iethree (approved):
LGTM

PR #35644, opened 2023-11-13
Remove redundant /user/recipient API calls

### LINKED ISSUE #35601: Search Results sending a bunch of `/user/recipient` queries when search is loaded
### Describe the bug

When search is loading (whether it's the search bar or the search results page), each search result is firing a `/user/recipient` query to get the `Created by`/ `Last edited` user. This query should be cached.

### To Reproduce

https://github.com/metabase/metabase/assets/25306947/0fe910a7-24da-4395-a29d-51554fb568d2

### Expected behavior

There should be 0 or 1 query depending on whether the query has already been fired and the data should be cached.

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "20.0.1",
    "java.vendor": "Homebrew",
    "java.version": "20.0.1",
    "java.vm.version": "20.0.1",
    "os.name": "Mac OS X",
    "os.version": "14.1.1",
    "user.language": "en",
    "user.timezone": "Europe/Tallinn"
  },
  "metabase-info": {
    "databases": [
      "postgres",
      "h2",
      "mongo"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "13.11"
      },
      "jdbc-driver": {
        "version": "42.6.0"
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2023-11-08",
      "src_hash": "26bbd1ec1f117a1338aca42212ffb1acd0d25046",
      "tag": "v1.47.1-SNAPSHOT",
      "hash": "6bb4382"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

P2

### Additional context

_No response_

oisincoveney:
Closes #35601 

### Description

Adds a `useUserRecipient` hook that caches data so the search results + search dropdowns reuse the same data. The data is reset on refresh, so there's no issue of having stale data after a long period of time.

### How to verify
1. Type something generic into the search dropdown (like `e`)
2. Check the network tab, and see that the `/user/recipients` API call is only called once

### Checklist


ranquild (changes requested):
Please use entity framework for this

oisincoveney:
> Please use entity framework for this

@ranquild Can we do that with this `UserApi` call? I went with this approach because I noticed that we don't use Redux for this specific API, it seems more like a "traditional" API in that sense

ranquild:
@oisincoveney no, if we need to cache something let's do it the same way it's done elsewhere

frontend/src/metabase/entities/users.js
```diff
@@ -24,13 +25,21 @@ function loadMemberships() {
   return require("metabase/admin/people/people").loadMemberships();
 }
 
+const getUserList = GET("/api/user");
+const getRecipientsList = GET("/api/user/recipients");
```
oisincoveney, line 29: I used this instead of `UserApi.list()` to be more clear about which API call is which

frontend/src/metabase/common/hooks/use-user-list-query/use-user-list-query.ts
```diff
@@ -7,7 +7,11 @@ import { useEntityListQuery } from "metabase/common/hooks/use-entity-list-query"
 import type { UserListResult } from "metabase-types/api";
 
 export const useUserListQuery = (
-  props: UseEntityListQueryProps<Record<string, never>> = {},
+  props: UseEntityListQueryProps<{
+    recipients: boolean;
+  }> = {
```
ranquild, line 12: We usually extract query params to `metabase-types`, please check other hooks

ranquild (approved):

PR #35646, opened 2023-11-13
Include E2E workflow in file-path triggers

nemanjaglumac:
Oops, this PR is fixing an oversight from #34892 where I forgot to include the workflow itself to the CI triggers.

Shame


WiNloSt (approved):
Looks good to me 👍

PR #35647, opened 2023-11-13
Allow spaces in datetime strings for uploads

### LINKED ISSUE #35648: Allow spaces in datetime strings for CSV uploads
This format correctly gets parsed as a timestamp column in a CSV upload:
`2022-01-01T01:00:00.00`

But this doesn't:
`2022-01-01 01:00:00.00`

We should allow both formats.

calherries:
Fixes https://github.com/metabase/metabase/issues/35648

This PR extends the allowed formats of datetime strings from ISO 8601 format (common form: `yyyy-MM-dd'T'HH:mm:ss`) to allow a space instead of the "T". 

e.g. both these formats are allowed as (local) datetime strings:
`2022-01-01 01:00:00.00`
`2022-01-01T01:00:00.00`

and both of these formats are allowed for offset datetime strings:
`2022-01-01 01:00:00.00+00:00`
`2022-01-01T01:00:00.00+00:00`

The full range of supported formats are described in the docstrings of the parsing functions:

https://github.com/metabase/metabase/blob/faccbdd95ce3ef2d844b7650241f28bfac4d5029/src/metabase/upload/parsing.clj#L37-L46

https://github.com/metabase/metabase/blob/faccbdd95ce3ef2d844b7650241f28bfac4d5029/src/metabase/upload/parsing.clj#L62-L77


src/metabase/upload/parsing.clj
```diff
@@ -135,7 +165,7 @@
 (defmethod upload-type->parser :metabase.upload/datetime
   [_]
   (comp
-   parse-datetime
+   parse-as-datetime
```
calherries, line 168: Note this needs to be `parse-as-datetime` instead of `parse-datetime` now because `parse-datetime` now doesn't parse date strings like "2022-01-01", but `parse-as-datetime` does.

test/metabase/upload_test.clj
```diff
@@ -124,10 +125,30 @@
...
+           [" 2022-01-01T01:00:00.00-07 "     #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01t01:00:00.00-07 "     #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01 01:00:00.00-07 "     #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01T01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01t01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01 01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01T01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01t01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01 01:00:00.00-07:00 "  #t "2022-01-01T01:00-07:00" offset-dt-type]
+           [" 2022-01-01T01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]
```
calherries, line 149: The `#t` reader tag will parse `"2022-01-01T01:00+00:00"` into a ZonedDateTime for some reason, so we need to use `t/offset-date-time` here

src/metabase/upload/parsing.clj
```diff
@@ -27,27 +27,57 @@
...
+      - Z (for UTC)
+      - +HH or -HH
+      - +HH:mm or -HH:mm
+      - +HH:mm:ss or -HH:mm:ss
+
+  Parsing is case-insensitive."
   [s]
   (try
-    (t/offset-date-time s)
+    (-> s (str/replace \space \T) t/offset-date-time)
```
qnkhuat, line 80: should we trim this first?

qnkhuat (approved):
nice

PR #35650, opened 2023-11-13
Split FilterPicker into subcomponents

ranquild:
I need to be able to add/edit a filter for the selected column only.

It should not be possible to click "Back" and change the column in the timeseries chrome and in drills. Instead of adding another prop to `FilterPicker`, I decided to extract a subcomponent that won't require `legacyQuery` or handling segments.


frontend/src/metabase/common/components/FilterPicker/BooleanFilterPicker/BooleanFilterPicker.tsx
```diff
@@ -49,7 +50,7 @@ export function BooleanFilterPicker({
   };
 
   return (
-    <Box data-testid="boolean-filter-picker">
+    <Box miw={MIN_WIDTH} maw={MAX_WIDTH} data-testid="boolean-filter-picker">
```
ranquild, line 53: Here and in other components - moved `MIN_WIDTH` from `FilterPicker` to each component.

frontend/src/metabase/common/components/FilterPicker/ColumnFilterPicker/ColumnFilterPicker.tsx
kulyk, line 1: I'm worried it'd be really confusing to have `FilterPicker`, `FilterColumnPicker`, and `ColumnFilterPicker`
Perhaps something like `FilterEditor` / `FilterWidget` / `FilterPickerBody` would work better?

ranquild:
Agree this PR should be done differently.

ranquild:
@kulyk I've reworked the PR. It depends on https://github.com/metabase/metabase/pull/35705 which fixes the issue where we had to check for the widget type to handle filter shape vs column type mismatches.

frontend/src/metabase/common/components/FilterPicker/BooleanFilterPicker/BooleanFilterPicker.tsx
```diff
@@ -5,8 +5,8 @@ import { checkNotNull } from "metabase/lib/types";
 import { Icon } from "metabase/core/components/Icon";
 import { Box, Button, Radio, Stack } from "metabase/ui";
 import * as Lib from "metabase-lib";
-import { FilterHeader } from "../FilterHeader";
-import { FilterFooter } from "../FilterFooter";
+import { FilterPickerHeader } from "../FilterPickerHeader";
```
ranquild, line 8: Since the new component is called `FilterPickerBody`, I thought it makes sense to rename `FilterHeader` and `FilterFooter` as well.

kulyk (approved):

Issue #35360, opened 2023-11-03
BE Flake: `metabase-enterprise.advanced-permissions.common-test/upload-csv-test`


calherries:
`metabase-enterprise.advanced-permissions.common-test/upload-csv-test flaked on be-tests-postgres-ee`
https://github.com/metabase/metabase/actions/runs/6745012295/job/18336037563?pr=35223#step:4:402

noahmoss:
I'm seeing this a lot too

PR #35654, opened 2023-11-13
Fix flaky `upload-csv-test`

calherries:
Aims at fixing https://github.com/metabase/metabase/issues/35360

We can't be sure the flake is fixed until it's in master for a while.

This also removes a test expression that was flaking on mysql in `check-data-migrations-rollback` ([example run](https://github.com/metabase/metabase/actions/runs/6850278499/job/18653873592?pr=35654#step:4:1060)). It wasn't necessary to fix ([slack context](https://metaboat.slack.com/archives/C0641E4PB9B/p1699945709800229)).

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
@@ -702,61 +702,59 @@
 
 (deftest upload-csv-test
   (mt/test-drivers (disj (mt/normal-drivers-with-feature :uploads) :mysql) ; MySQL doesn't support schemas
-    (mt/with-empty-db
-      (let [db-id (u/the-id (mt/db))]
-        (testing "Should be blocked without data access"
+    (testing "Uploads should be blocked without data access"
+      (mt/with-empty-db
+        (let [conn-spec (sql-jdbc.conn/db->pooled-connection-spec (mt/db))]
           ;; Create not_public schema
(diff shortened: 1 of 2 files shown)


enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -702,61 +702,59 @@
...
-                      (testing "GET /api/database"
-                        (let [result (->> (mt/user-http-request :rasta :get 200 "database")
-                                          :data
-                                          (filter #(= (:id %) db-id))
-                                          first)]
-                          (is (= can-upload? (:can_upload result)))))
-                      (testing "GET /api/database/:id"
-                          (let [result (mt/user-http-request :rasta :get 200 (format "database/%d" db-id))]
-                            (is (= can-upload? (:can_upload result))))))))))))))))
+  (mt/test-drivers (disj (mt/normal-drivers-with-feature :uploads) :mysql) ; MySQL doesn't support schemas
```
calherries, line 732: This test wasn't actually running, because it was using `mt/test-driver` instead of `mt/test-drivers`

calherries:
@darksciencebase Yes, a workflow to run a test multiple times in CI would be ideal.

piranha:
I wonder which change exactly should prevent 'flakiness' of a test? Is it using `conn-spec` instead of the older code?

sansolovyov (approved):
Well makes sense to me!

PR #35655, opened 2023-11-13
Workaround for E2E search failures due to audit V2

nemanjaglumac:
Audit v2 table intermittently shows up in search results, which makes related E2E tests fail.
This happens in some 15-25% of all runs so it calls for action before @escherize devise a proper solution on the backend.

The solution for the e2e part should be straightforward.
We filter out this ghost `MODERATION_REVIEW` table when we do the count assertion.

Here's the result from my local run showing the test passing even with this table in the results:
image

### DIFF
--- a/e2e/test/scenarios/onboarding/search/search.cy.spec.js
+++ b/e2e/test/scenarios/onboarding/search/search.cy.spec.js
@@ -1484,9 +1484,19 @@ function expectSearchResultItemNameContent(
   { strict } = { strict: true },
 ) {
   cy.findAllByTestId("search-result-item-name").then($searchResultLabel => {
+    console.log($searchResultLabel);
     const searchResultLabelList = $searchResultLabel
       .toArray()
-      .map(el => el.textContent);
+      // FIXME!
+      // AuditV2 `MODERATION_REVIEW` table is unexpectedly and intermitently showing up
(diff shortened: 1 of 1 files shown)


ranquild (approved):

e2e/test/scenarios/onboarding/search/search.cy.spec.js
```diff
@@ -1484,9 +1484,19 @@ function expectSearchResultItemNameContent(
...
       .toArray()
-      .map(el => el.textContent);
+      // FIXME!
+      // AuditV2 `MODERATION_REVIEW` table is unexpectedly and intermitently showing up
+      // in search results. This is causing test failures due to the count of search items mismatch.
+      // TODO:
+      // Remove the `filter()` as soon as audit v2 part is fixed!
+      .filter(el => el.textContent !== "MODERATION_REVIEW")
+      .map(el => {
+        console.log(el.textContent);
```
nemanjaglumac, line 1497: I'll remove this stray console log.

oisincoveney: thanks

oisincoveney (approved):

iethree:
I think this is a great example of why having a non-blocking flake workflow makes tons of sense. cc @perivamsi 

https://github.com/metabase/metabase/issues/35661

PR #35656, opened 2023-11-13
Run stubbed backend jobs when main `backend` tests are skipped

nemanjaglumac:
This should make this way more simple.

We have a few chained conditions under which `be-tests` run. Since these checks are marked as required, we need the backend-tests-stub to run if main tests are skipped. Rather than trying to mimic those chained conditions (inverted), is there anything stopping us from simply saying: "Run a stub job if the main one is skipped"?

The example of the mismatch in conditions that left these required checks pending is this run:
https://github.com/metabase/metabase/actions/runs/6850259807/job/18624077495

Files changed output:
`Changes output set to ["e2e_specs","e2e_all"]`

Note six expected checks
image

With this proposed approach, such scenarios shouldn't happen anymore.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -217,11 +217,10 @@ jobs:
           fail-on-error: false
 
   be-tests-stub:
-    needs: [files-changed, static-viz-files-changed]
+    needs: [be-tests]
     if: |
-      always() &&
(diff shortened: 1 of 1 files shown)


iethree (approved):

iethree:
PS, this is so stupid. can't wait til github adds conditional required checks

PR #35659, opened 2023-11-13
[Storybook]: Deduplicate babel types to fix build

uladzimirdev:
### Description

This problem appeared after https://github.com/metabase/metabase/pull/35602 where I suggested to upgrade all babel packages

old style package managers like npm or yarn@1 have a well known problem called "doppelgangers", [more info](https://rushjs.io/pages/advanced/npm_doppelgangers/).

tl;dr
packageA has a dep of `packageD@^1.1.1`
packageB has a dep of `packageD@^1.1.2`
latest version of `packageD` is 1.1.5

so even if logically we need to install the only version of packageD 1.1.5, which will satisfy every package, npm and yarn@1 do not do it by design (and can't change it as it will break backward compatibility)

In our case there are plenty of installed `babel` packages: `@babel/core`, `@babel/types` etc, 
so even if storybook has it's own listed `@babel/core` and `@babel/types` (`node_modules/@storybook/.../node_modules/@babel/core`), sometimes node or webpack doesn't resolve the correct version of the package (e.g. version from `node_modules/@babel/core`) is found first, which can lead to the broken build:
in our case `@babel/types` of old version didn't contains some exported API which was used in storybook.

Those issues are pretty hard to track and debug, there are several issues on storybook (like [this](https://github.com/storybookjs/storybook/issues/12893#issuecomment-731504027)) about the same and since with yarn@1 and npm even order of adding deps matters for building lockfile, the result is not determenistic. 

### How did I get the fix

I played with storybook config and got not cryptic error message which led me to the babel issue and the point was: try to avoid duplications in the node_modules

there is a special package for yarn: `https://github.com/scinos/yarn-deduplicate`

```
npx yarn-deduplicate --packages @babel/types
```

^^ updates the lockfile with deduplicated deps. it's almost the same as using `resolutions`, but more explicit.

Side effect: shrink of node_modules

before:
image

after:
Screenshot 2023-11-14 at 16 23 46

### How to verify

check this build https://github.com/metabase/metabase/actions/runs/6852603043?pr=35659

or locally:

`rm -rf node_modules && yarn && yarn build-storybook`

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -14,6 +14,7 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-a11y",
   ],
+  babel: () => {},
   webpackFinal: storybookConfig => ({
     ...storybookConfig,
     plugins: [...storybookConfig.plugins, new MiniCssExtractPlugin()],
--- a/yarn.lock
+++ b/yarn.lock
@@ -1247,21 +1247,6 @@
   dependencies:
     "@babel/types" "^7.22.5"
 
-"@babel/helper-string-parser@^7.18.10":
-  version "7.18.10"
-  resolved "https://registry.yarnpkg.com/@babel/helper-string-parser/-/helper-string-parser-7.18.10.tgz#181f22d28ebe1b3857fa575f5c290b1aaf659b56"
-  integrity sha512-XtIfWmeNY3i4t7t4D2t02q50HvqHybPqW2ki1kosnvWCwuCMeo81Jf0gwr85jy/neUdg5XDdeFE/80DXiO+njw==
-
-"@babel/helper-string-parser@^7.19.4":
-  version "7.19.4"
-  resolved "https://registry.yarnpkg.com/@babel/helper-string-parser/-/helper-string-parser-7.19.4.tgz#38d3acb654b4701a9b77fb0615a96f775c3a9e63"
-  integrity sha512-nHtDoQcuqFmwYNYPz3Rah5ph2p8PFeFCsZk9A/48dPc/rGocJ5J3hAAZ7pb76VWX3fZKu+uEr/FhH5jLx7umrw==
-
-"@babel/helper-string-parser@^7.21.5":
-  version "7.21.5"
-  resolved "https://registry.yarnpkg.com/@babel/helper-string-parser/-/helper-string-parser-7.21.5.tgz#2b3eea65443c6bdc31c22d037c65f6d323b6b2bd"
-  integrity sha512-5pTUx3hAJaZIdW99sJ6ZUUgWq/Y+Hja7TowEnLNMm1VivRgZQL3vpBY3qUACVsvw+yQU6+YgfBVmcbLaZtrA1w==
-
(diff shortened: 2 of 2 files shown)


.storybook/main.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-a11y",
   ],
+  babel: () => {},
```
uladzimirdev, line 17: storybook combines our config and their config which leads to the problem of resolving packages

yarn.lock
```diff
@@ -3913,95 +3893,7 @@
...
-  dependencies:
-    "@babel/helper-string-parser" "^7.22.5"
-    "@babel/helper-validator-identifier" "^7.22.20"
-    to-fast-properties "^2.0.0"
-
-"@babel/types@^7.22.19", "@babel/types@^7.22.5", "@babel/types@^7.23.0", "@babel/types@^7.23.3":
+"@babel/types@^7.0.0", "@babel/types@^7.10.4", "@babel/types@^7.10.5", "@babel/types@^7.11.0", "@babel/types@^7.12.1", "@babel/types@^7.12.11", "@babel/types@^7.12.5", "@babel/types@^7.12.6", "@babel/types@^7.12.7", "@babel/types@^7.14.5", "@babel/types@^7.14.8", "@babel/types@^7.16.0", "@babel/types@^7.18.10", "@babel/types@^7.18.6", "@babel/types@^7.18.9", "@babel/types@^7.19.0", "@babel/types@^7.19.3", "@babel/types@^7.2.0", "@babel/types@^7.20.0", "@babel/types@^7.20.2", "@babel/types@^7.20.5", "@babel/types@^7.20.7", "@babel/types@^7.21.0", "@babel/types@^7.21.3", "@babel/types@^7.21.4", "@babel/types@^7.21.5", "@babel/types@^7.22.0", "@babel/types@^7.22.15", "@babel/types@^7.22.19", "@babel/types@^7.22.3", "@babel/types@^7.22.4", "@babel/types@^7.22.5", "@babel/types@^7.23.0", "@babel/types@^7.23.3", "@babel/types@^7.3.0", "@babel/types@^7.3.3", "@babel/types@^7.4.4", "@babel/types@^7.7.0", "@babel/types@^7.8.3":
```
uladzimirdev, line 3896: @babel/types should be backward compatible, so newer version is older + some fixes

uladzimirdev:
> why did yarn let this happen?

@iethree because yarn does it by design, please check the updated PR description

npretto:
Considering that yarn 1.x is in maintainance mode
image

Should we also start talking about a migration plan?
A quick google search seems to indicate that npm is still slower than yarn 1.x, but pnpm is faster than both (and has been out for a while so it should be mature enough)

iethree:
❤️ 🚀  This was the most educational PR description I've read in a long time.

iethree (approved):

uladzimirdev:
@npretto that's a good question, but I think before we have a dedicated DX team, we need to solve other more important problems

PR #35664, opened 2023-11-13
Add warning for h2 uploads db

### LINKED ISSUE #35638: A big warning that if you use CSV uploads in the sample database in the cloud, your data will disappear
### Describe the bug

Containers were meant to be stateless, and we shouldn't break that rule never ever

Users started using the CSV upload feature as a mean not to have a DW, and we know how it all ended: the pod recycled and everything dissapeared

we should put a big big warning that if you use the sample database on the cloud, your tables will dissapear once we make any operation

### To Reproduce

1) set up an instance on the cloud
2) upload a csv
3) stop and start the instance, or wait till we upgrade or something
4) see all your tables disappearing and queries not running

### Expected behavior

It should stay the same, but we should warn users

### Logs

NA

```JSON
v46.x+
```

### Severity

P1

### Additional context

_No response_

iethree:
Closes https://github.com/metabase/metabase/issues/35638

### Description

If an admin selects the sample database as an upload target, warn them that setting an h2 uploads DB is for testing only and they will likely lose data.

Screen Shot 2023-11-13 at 8 46 31 AM

### Checklist

### DIFF
--- a/frontend/src/metabase/admin/settings/components/UploadSettings/UploadSettings.tsx
+++ b/frontend/src/metabase/admin/settings/components/UploadSettings/UploadSettings.tsx
@@ -13,13 +13,14 @@ import { updateSettings } from "metabase/admin/settings/settings";
 
 import type { State } from "metabase-types/store";
 
-import { Stack, Group } from "metabase/ui";
+import { Stack, Group, Text } from "metabase/ui";
 import Link from "metabase/core/components/Link";
 import type { SelectChangeEvent } from "metabase/core/components/Select";
 import Select from "metabase/core/components/Select";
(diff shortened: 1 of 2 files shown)


dragonsahead (approved):
LGTM

luizarakaki (approved):

shaunlebron (approved):
FE is good.  I think the warning is a bit soft compared to the original proposal: “[the data you upload will disappear](https://metaboat.slack.com/archives/C04S696LRUM/p1699874729884329?thread_ts=1699874448.953159&cid=C04S696LRUM)”, but that’s up to product and design.

PR #35670, opened 2023-11-13
Mark top tier E2E flakes

nemanjaglumac:
Prep work for #35661 (or maybe even a PR that resolves that issue).

This PR marks top flaky tests in E2E suite.
I used two criteria for the "top tier" flakes:
1. failures above 2% frequency on `master` in the last seven days
2. flakes above 10% frequency on `master` in the last seven days

I didn't mark "search" related failures and flakes, because they are exception caused by a real bug. Those should be fixed by https://github.com/metabase/metabase/pull/35655 anyway.

I also didn't mark flakes that have been fixed and confirmed to run correctly within the last seven days.


nemanjaglumac:
Closing in favor of #35676

PR #35674, opened 2023-11-13
Add tests verifying that #18817 has been fixed

### LINKED ISSUE #18817: Removing a grouping leaves grouping columns partially visible
**Describe the bug**
When starting with a saved question, adding 2 groupings and then deleting one, leaves some orphan column names in the top bar, that are even functional

**Logs**
NA

**To Reproduce**
1) With the sample dataset create a new GUI question (Let's say, just click the people table and save it as People).
2) access the saved question (or refresh the page), and then add 2 group by's (name and source). Then remove one of those
3) see the column titles still populated with some groupings still visible

**Expected behavior**
Just see the columns that are in the groupings

**Screenshots**
Stats
image

On a 40.5 I was testing with:
image

- Your browser and the version: Brave latest and Firefox 93
- Your operating system: Ubuntu 21.10
- Your databases: H2
- Metabase version: master branch and 40.5
- Metabase internal database: H2

**Severity**
Cosmetic

**Additional context**
Might be a cousin of https://github.com/metabase/metabase/issues/13772, but I'm not very sure

metamben:
It verifies both that after removing breakouts the returned columns are valid, as well as that given a question based on an aggregated source query, a fields corresponding a breakout can be removed.

While testing, I found that questions based on saved questions involving pivoting are quite broken, but I think that's not what #18817 was about. (In the description there saving the question with breakouts is never mentioned.)
image

@uladzimirdev is working on a FE test for this, I think.

### DIFF
--- a/test/metabase/lib/field_test.cljc
+++ b/test/metabase/lib/field_test.cljc
@@ -1402,6 +1402,33 @@
                           (lib/remove-field -1 vis-price)
                           lib.metadata.calculation/returned-columns))))))))))
 
+(deftest ^:parallel nested-aggregation-query-remove-fields-test
+  (testing "can remove a breakout field from a nested aggregated query (#18817)"
+    (let [provider (lib.tu/metadata-provider-with-cards-for-queries
+                    meta/metadata-provider
+                    [(-> (lib/query meta/metadata-provider (meta/table-metadata :orders))
+                         (lib/aggregate (lib/count))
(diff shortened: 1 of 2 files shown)


camsaul (approved):

uladzimirdev:
@metamben did you find when was the issue fixed?

uladzimirdev:
FE tests are merged https://github.com/metabase/metabase/pull/35744

PR #35676, opened 2023-11-13
Quarantine Flaky e2e Tests

### LINKED ISSUE #35661: Isolate flakes from the rest of tests

iethree:
Closes https://github.com/metabase/metabase/issues/35661

### Description

- Quarantines Flaky e2e tests in their own non-blocking workflow
- Should all be assigned to be fixed and only live there temporarily

Top 10 flakes | Top 6 failures on master
--- | ---
Screen Shot 2023-11-14 at 12 14 32 PM | Screen Shot 2023-11-14 at 12 14 51 PM

### DIFF
--- a/.github/actions/build-e2e-matrix/action.yml
+++ b/.github/actions/build-e2e-matrix/action.yml
@@ -50,6 +50,7 @@ runs:
             ["visualizations-tabular", {} ],
             ["oss-subset", { edition: 'oss', context: "special" } ],
             ["slow", { runner: beefierRunner, context: "special" } ],
+            ["flaky", { runner: beefierRunner, context: "special" } ],
             ["mongo", { context: "qa-database" } ],
           ];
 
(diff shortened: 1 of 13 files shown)


.github/actions/build-e2e-matrix/action.yml
```diff
@@ -50,6 +50,7 @@ runs:
             ["visualizations-tabular", {} ],
             ["oss-subset", { edition: 'oss', context: "special" } ],
             ["slow", { runner: beefierRunner, context: "special" } ],
+            ["flaky", { runner: beefierRunner, context: "special" } ],
```
nemanjaglumac, line 53: Historically, running same tests on different runners yielded completely different results. I'm concerned that we'd maybe mask the underlying problem (or even potentially discover new ones specific to BuildJet runners).

IMO, we should solve the flakiness for the slowest and the least powerful runners, which are currently the default GitHub runners.

nemanjaglumac, line 53: What about flakes that can run using only OSS uberjar?

iethree: I just counted, and there are only 12 such tests in our entire test suite. I think it's best not to handle this.

perivamsi (approved):

nemanjaglumac (approved):
Let's ship this 🚀

PR #35680, opened 2023-11-13
Use sessions in mongo aggregation pipelines

### LINKED ISSUE #32372: Queries not getting cancelled in MongoDB when leaving the page
### Describe the bug

Metabase gets completely unusable when you send MongoDB a lot of heavy queries, and even if you navigate away from the dashboard, Metabase will be almost unusable and the queries in Mongo won't be cancelled.

### To Reproduce

1) set up Metabase + Mongo with the sample DB (y'know: docker pull metabase/qa-databases:mongo-sample-6)
2) connect to Mongo and then do a tough native query like 
```
[
  {
    "$lookup": {
      "from": "people",
      "let": {
        "let_user_id_2714": "$user_id"
      },
      "pipeline": [
        {
          "$match": {
            "$expr": {
              "$eq": [
                "$$let_user_id_2714",
                "$id"
              ]
            }
          }
        }
      ],
      "as": "join_alias_People"
    }
  },
  {
    "$unwind": {
      "path": "$join_alias_People",
      "preserveNullAndEmptyArrays": true
    }
  },
  {
    "$match": {{state}}
  },
  {
    "$group": {
      "_id": {
        "created_at": {
          "$let": {
            "vars": {
              "parts": {
                "$dateToParts": {
                  "timezone": "GMT",
                  "date": "$created_at"
                }
              }
            },
            "in": {
              "$dateFromParts": {
                "timezone": "GMT",
                "year": "$$parts.year",
                "month": "$$parts.month"
              }
            }
          }
        }
      },
      "sum": {
        "$sum": "$total"
      }
    }
  },
  {
    "$sort": {
      "_id": 1
    }
  },
  {
    "$project": {
      "_id": false,
      "created_at": "$_id.created_at",
      "sum": true
    }
  }
]
```

configure the parameter as a field filter to the people table, state parameter.

(another way of reproducing this would be to do this query via GUI)
image

3) now add the query 20 times to the dashboard, click save, refresh the dashboard a few times so the query hits MongoDB many many times, and then navigate away from the dashboard

See the logs mentioning: 
```
2023-07-13 19:33:50,745 DEBUG middleware.log :: POST /api/dashboard/1/dashcard/1/card/1/query 202 [ASYNC: canceled] 8.3 s (13 DB calls) App DB connections: 0/15 Jetty threads: 2/50 (22 idle, 0 queued) (867 total active threads) Queries in flight: 0 (0 queued)
```
but in fact, your mongo db will keep running those queries till they finish or time out

### Expected behavior

We should close the connections when the user navigates away from the question or dashboard

### Logs

NA

```JSON
v46.x
```

### Severity

P1

### Additional context

_No response_

lbrdnk:
Closes #32372

### Description

Prior to this PR mongo driver did not kill the queries when connection was closed (ie. browser windows was closed). This PR takes advantage mongo's sessions to accomplish that. 

_Use of sessions._ We were not using sessions in an explicit way in the mongo driver prior to this PR. From what I've found, use explicit sessions provides the only mechanism capable of killing a running operation. For more context see [this section](https://metaboat.slack.com/docs/T078VCLCR/F062AFF81K9?focus_section_id=temp:C:ITf577344081b0d44a48022f5070) of my notes. Specifically links to [the issue](https://jira.mongodb.org/browse/JAVA-1109) on mongodb's JIRA and this stackoverflow [question](https://stackoverflow.com/questions/31079479/cancel-a-mongodb-query-from-java-api). Also [the sessions specification](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#explicit-vs-implicit-sessions) describes killing of long running operations by the means of sessions.

_Performance impact_. I've measured the average execution time of 100 basic queries before and after the change, finding no significant difference.

_Restrictions on mongo topologies (deployment setups)_. Sessions were introduced in mongodb version 3.6. We currently support versions 4.2+. [Mongodb reference](https://www.mongodb.com/docs/manual/reference/server-sessions/#overview) states that:
> Starting in 3.6, MongoDB drivers associate all operations with a server session...

but in contrary to that, previous section of that reference states following:
> Server sessions are available for replica sets and sharded clusters only.

I've asked for the correct interpretation of the cited sentence on relevant fora with no answer to this point. However, testing (and development) on our docker images shown that no clustering or replication setup was necessary for the sessions to work properly. Hence, I believe we are safe to go further with this change. Furthermore, no relevant restrictions on mongo topologies were found.

_Useful links_:
- [Mongodb sessions spec](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst).
- [Mongodb server sessions reference](https://www.mongodb.com/docs/manual/reference/server-sessions/).
- [Driver's javadoc 3.12](https://mongodb.github.io/mongo-java-driver/3.12/javadoc/).

### How to verify

See the new test, or see mongo logs after closing browser window with with in progress query.

### Checklist


modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -161,30 +168,59 @@
...
+      (let [first-row                        (when (.hasNext cursor)
+                                               (.next cursor))
+            {row-col-names       :row
+             unescaped-col-names :unescaped} (result-col-names native-query query (row-keys first-row))]
+        (log/tracef "Renaming columns in results %s -> %s" (pr-str row-col-names) (pr-str unescaped-col-names))
+        (respond (result-metadata unescaped-col-names)
+                 (if-not first-row
+                   []
+                   (reducible-rows context cursor first-row (post-process-row row-col-names))))))
+    (catch MongoCommandException ex
```
lbrdnk, line 187: Is it desirable to catch this exception? Is it desirable to log that information with a warning level? `interrupt-ex?` checks error code, so other exceptions are re-thrown.

modules/drivers/mongo/test/metabase/driver/mongo/execute_test.clj
```diff
@@ -53,3 +67,69 @@
...
+            compiled (-> query qp/compile :query)
+            pipeline (m.util/into-array-list (m.conversion/to-db-object compiled))]
+        (mongo.util/with-mongo-connection [connection (mt/id)]
+          (let [client-database ^MongoDatabase (#'mongo.execute/connection->database connection)
+                collection ^MongoCollection (. client-database (getCollection "orders"))]
+            (with-open [session ^ClientSession (#'mongo.execute/start-session! connection)]
+              (let [aggregate (.aggregate collection session ^java.util.ArrayList pipeline)
+                    ;; Manually tested: if session is closed before aggregation execution takes place (call to
+                    ;; eg. either `.into` or `.cursor`) aggregation is then executed.
+                    ;; TODO: Find workaround!
```
metamben, line 100: Does `.into` leave its mark somewhere on the observable world? If yes, we should look for that mark. If not, then we can only hope that we wait enough before we kill the session. In this latter case, we could set a semaphore in the `a/thread` just before calling `.into` and wait on it before starting the `Thread/sleep`. That would reduce the chance of killing the session ahead of time to a minimum.

modules/drivers/mongo/test/metabase/driver/mongo/execute_test.clj
```diff
@@ -53,3 +67,69 @@
...
+                    result-ch (a/thread (try (.into aggregate (java.util.ArrayList.))
+                                             (catch MongoCommandException ex
+                                               (if (interrupt-ex? ex)
+                                                 :interrupted
+                                                 ex))))]
+                (future (Thread/sleep 500)
+                        (#'mongo.execute/kill-session! client-database session))
+                 ;; Using 15k timeout to handle unforseen circumstances.
```
metamben, line 108: This seems out of sync and not very helpful in general.

modules/drivers/mongo/test/metabase/driver/mongo/execute_test.clj
```diff
@@ -53,3 +67,69 @@
                            [1 "Rasta"     now nil]]
                     :columns ["_id" "name" "last_login" "alias"]}
                    (mt/rows+column-names (qp/process-query query))))))))))
+
+(defn- interrupt-ex? [^MongoCommandException ex]
+  (and (= 11601 (.getErrorCode ex))
+       (= "Interrupted" (.getErrorCodeName ex))))
+
+(deftest kill-an-in-flight-query-method-validation-test
+  (testing "Confirm that method used to kill in-flight queries on mongo works"
```
metamben, line 76: I wonder if it's worth having this test. If it fails, the next test should fail too. I can help debugging maybe, but I'm not sure that's worth the maintenance effort.

modules/drivers/mongo/test/metabase/driver/mongo/execute_test.clj
```diff
@@ -53,3 +67,69 @@
...
+                                             :condition
+                                             [:!= $user_id &People_User.people.id],
+                                             :source-table $$people}]})]
+          (future (Thread/sleep 500)
+                  (a/>!! canceled-chan ::streaming-response/request-canceled))
+          (testing "Cancel signal kills the in progress query"
+            (is (re-find #"Command failed with error 11601.*operation was interrupted"
+                         (try (qp/process-query query)
+                              (catch Throwable e
+                                (ex-message e))))))))))))
```
metamben, line 135: Can't you use `(is (thrown-with-msg? c re body))`?

metamben (commented):
It looks good in general, but I'd like to discuss if this change causes any performance degradation or restricts the usable databases.

modules/drivers/mongo/test/metabase/driver/mongo/execute_test.clj
lbrdnk: I'm removing it. I've used it during the development, while trying to wire up sessions with our then current mongo code. Now it's obsolete.

modules/drivers/mongo/test/metabase/driver/mongo/execute_test.clj
lbrdnk: Test was removed as advised later.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -161,30 +168,58 @@
...
+    (with-open [session ^ClientSession (start-session! mongo.util/*mongo-connection*)]
+      (a/go
+        (when (a/<! (qp.context/canceled-chan context))
+          (kill-session! client-database session)))
+      (let [aggregate ^AggregateIterable (*aggregate* client-database
+                                                      collection-name
+                                                      session
+                                                      query
+                                                      (qp.context/timeout context))]
+        (with-open [^MongoCursor cursor (try (.cursor ^AggregateIterable aggregate)
```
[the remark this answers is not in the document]
lbrdnk: I believe it's because `with-open` calls `.close` on it.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -129,24 +130,30 @@
...
-        agg-opts (aggregation-options timeout-ms)
-        pipe     (m.util/into-array-list (m.conversion/to-db-object stages))]
-    (.aggregate coll pipe agg-opts)))
-
-(defn- reducible-rows [context ^Cursor cursor first-row post-process]
+  [^MongoDatabase db
+   ^String coll
+   ^ClientSession session
+   stages timeout-ms]
+  (let [coll      ^MongoCollection (.getCollection db coll)
```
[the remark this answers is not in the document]
lbrdnk: I'm removing it.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -161,30 +168,58 @@
...
+(defn- connection->database
+  ^MongoDatabase
+  [^DB connection]
+  (let [db-name (.getName connection)]
+    (.. connection getMongoClient (getDatabase db-name))))
+
+(defn- start-session!
+  ^ClientSession
+  [^DB connection]
+  (.. connection getMongoClient startSession))
```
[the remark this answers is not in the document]
lbrdnk: > Have we been running queries without sessions before this change?
Yes.

> Have you checked if there is any measurable change in performance?
I did after reading your comment. I've found no noticeable change in performance. I've tested it computing avg execution time for a simple query executed 100 times.

> Do you know if the new way introduces any new requirements about the databases?
Sessions were introduced in mongo 3.6. As you've noted in the other comment, we support 4.2+ mongo versions, so I believe it is safe to use sessions.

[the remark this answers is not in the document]
metamben: I meant requirements like specific ways to configure the DB 
e.g., running in a specific mode or supporting "transactions", i.e., anything that might not be given for all setups.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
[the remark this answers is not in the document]
metamben: Are you also talking about the `^AggregateIterable` hint?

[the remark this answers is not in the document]
lbrdnk: You are right, thanks. I'll remove it shortly.

metamben (approved):
Looks good to me, just extend the description as discussed.

PR #35684, opened 2023-11-13
remove internal user on rollback + test

escherize:
When rolling back, we need to delete the internal metabase user (who has the id of `13371338`).

This PR adds a migration that is similar to `v48.00-050`: a no-op forward-migration, and deletes the internal user when downgrading.

### DIFF
--- a/resources/migrations/000_migrations.yaml
+++ b/resources/migrations/000_migrations.yaml
@@ -16054,7 +16054,7 @@ databaseChangeLog:
       rollback: # no change
 
   - changeSet:
-      id: v48.00-052
+      id: v48.00-053
       author: johnswanson
       comment: Increase length of `activity.model` to fit longer model names
       changes:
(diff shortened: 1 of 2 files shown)


resources/migrations/000_migrations.yaml
```diff
@@ -16054,7 +16054,7 @@ databaseChangeLog:
       rollback: # no change
 
   - changeSet:
-      id: v48.00-053
+      id: v48.00-052
```
escherize, line 16057: Just a typo from john, double checked that he meant `v48.00-052`.

noahmoss, line 16057: This will break the app DB for anyone who's already ran the previous version of the 053 migration. Not the end of the world but might be a bit nicer to keep 053 as it is and either use 052 or 054 (there's no problem with skipping migration IDs)

escherize: Good idea, I'll do that.

noahmoss (approved):

resources/migrations/000_migrations.yaml
escherize: Forgot about the Liquibase checksums.

PR #35686, opened 2023-11-13
Hide edit question menu item on dashboards when there is no collection write permission

### LINKED ISSUE #35077: Users with read-only permissions to collections are shown "edit question" option in dashboard cards
### Describe the bug

Seems like a UI bug where we show "edit question" to someone that does not have permission to edit the question.

(the issue does not let someone edit a question, so this is just a UI bug)
image

### To Reproduce

1) create a user
2) give "all users" read only permissions to all collections
3) create a dashboard from an x-ray
4) authenticate to Metabase with the normal user
5) go to the dashboard with the user that does not have permissions
6) click on the 3 dots of a card, see "edit question" when it shouldn't be there

### Expected behavior

We shouldn't give options that are not valid for a user

### Logs

NA

```JSON
v47.x
```

### Severity

P3

### Additional context

It can lead to users believing they have more permissions than what they actually have

alxnddr:
Closes https://github.com/metabase/metabase/issues/35077

### Description

When users have "View" collection permission for a dashcard, they should not see "Edit question" menu item.

## This PR needs a BE change:

Include `can_write` to every `card` entity in the response of `/api/dashboard/:id` handler:
Screenshot 2023-11-13 at 6 54 26 PM

### How to verify

- As an admin create any question Q1 and add it to a dashboard D1
- Create a user that has the data access to a question, but set only the "View" access level for the question's collection
- Login as the user, open the D1 dashboard
- In the top right corner of the card click on "..." to open the menu
- Ensure it does not show the "Edit question" menu item

### Checklist

### DIFF
--- a/frontend/src/metabase/dashboard/components/DashCard/DashCardMenu/DashCardMenu.tsx
+++ b/frontend/src/metabase/dashboard/components/DashCard/DashCardMenu/DashCardMenu.tsx
@@ -131,7 +131,11 @@ interface QueryDownloadWidgetOpts {
 }
 
 const canEditQuestion = (question: Question) => {
-  return question.query() != null && question.query().isEditable();
+  return (
+    question.canWrite() &&
+    question.query() != null &&
+    question.query().isEditable()
+  );
 };
 
 const canDownloadResults = (result?: Dataset) => {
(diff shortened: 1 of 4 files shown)


markbastian (approved):
BE looks good. I love the issue writeup.

EmmadUsmani (approved):

PR #35687, opened 2023-11-13
docs - metabase analytics

jeff-bruemmer:
Rough draft documenting Metabase analytics.

- Emphasizes the special nature of the collection
- Lists its contents for reference.
- Mentions audit log deprecation, and points to new collection.

The contents will likely change between now and release, and potentially release to release.

@luizarakaki other than updating the dashboards and models, anything else we should include here? For example, once the dust settles, we could include descriptions of questions.


docs/README.md
```diff
@@ -179,7 +179,7 @@ Metabase's reference documentation.
 ### Tools
 
 - [Tools overview](./usage-and-performance-tools/start.md)
-- [Auditing tools](./usage-and-performance-tools/audit.md)
+- [Metabase analytics](./usage-and-performance-tools/metabase-analytics.md)
```
luizarakaki, line 182: There is something interesting about the name of this feature. While in-product the collection is `Metabase analytics`, externally we will call it `Usage analytics` AFAIK.

I think this makes sense because this is called Usage Analytics as a feature, but as a Metabase collection, it would be ambiguous to call it `Usage analytics`

docs/usage-and-performance-tools/metabase-analytics.md
```diff
@@ -0,0 +1,371 @@
...
+
+{% include plans-blockquote.html feature="Metabase analytics" %}
+
+The **Metabase Analytics** collection is a special collection that contains view-only questions, dashboards, and models that help you understand how people are using your Metabase.
+
+There are two things to keep in mind with this special Metabase analytics collection:
+
+- It's a view-only collection. Even admins can't curate it. It just _exists_.
+- You can duplicate anything in it, modify it to your liking, and save it to another collection.
+
```
luizarakaki, line 15: I think I'd add a section here explaining why this is useful, mentioning the 3 uses cases:

1. Usage: understand how people are using Metabase, how many WAU, new questions, most active users and groups etc
2. Audit: know who saw/did what when. Including dashboard/question views, queries, downloads, activity like changing settings or inviting users
3. Operations: know the slowest dashboards and questions, database performance, users consuming too much resources

luizarakaki (approved):

PR #35693, opened 2023-11-14
When attemp to delete inactive tables, skip tables that have active cards

### LINKED ISSUE #35615: Check if inactive tables are being used by questions before deleting them
Related to https://github.com/metabase/metabase/pull/35443

We should check if the table is being used by any question before deleting it.
Delete only tables that are not being used.

If this is not enough to solve the AppDB size problem, we can consider removing this condition in a following release.

qnkhuat:
Closes https://github.com/metabase/metabase/issues/35615


src/metabase/task/delete_inactive_tables.clj
```diff
@@ -25,8 +25,12 @@
   (log/debug (trs "Cleaning up Inactive Tables older than {0} days" inactive-table-max-days))
   (task-history/with-task-history {:task "delete-inactive-tables"}
     (t2/delete! :model/Table
-                :active false
-                :updated_at [:< (t/minus (t/zoned-date-time) (t/days inactive-table-max-days))])))
+                :active     false
+                :updated_at [:< (t/minus (t/zoned-date-time) (t/days inactive-table-max-days))]
+                ;; do not delete tables that have active cards associated with them (#35615)
+                :id         [:not-in {:select-distinct [:table_id]
```
qnkhuat, line 31: an interesting performance analysis on various ways to do this https://explainextended.com/2009/09/18/not-in-vs-not-exists-vs-left-join-is-null-mysql/

calherries:
I'm not sure this meets the resolution criteria for https://github.com/metabase/metabase/issues/35615

This only checks for non-native questions that have the table as the base table. It misses checking:
1. native questions that use the table in joins
2. non-native questions that use the table at all

Which means these tables will get deleted even though they are technically still referenced by existing questions.

qnkhuat:
@calherries I've chatted with @luizarakaki on this [here](https://metaboat.slack.com/archives/C0641E4PB9B/p1699628520827649) 

sorry should have linked to the thread in the description.

test/metabase/task/delete_inactive_tables_test.clj
```diff
@@ -12,21 +12,30 @@
...
+       :model/Table            {old-inactive-table-active-card-id :id} {:db_id       db-id
+                                                                        :active      false
+                                                                        :updated_at  older-than-threshold}
+       :model/Card             {_active-card :id}                      {:database_id db-id
+                                                                        :table_id    old-inactive-table-active-card-id}]
 
       (#'delete-inactive-tables/delete-inactive-tables!)
       (testing "inactive tables that are older than threshold should be deleted"
-        (is (= #{new-active-table-id old-active-table-id new-inactive-table-id}
+        (is (= #{new-active-table-id  new-inactive-table-id
```
calherries, line 39:
```suggestion
        (is (= #{new-active-table-id new-inactive-table-id
```

test/metabase/task/delete_inactive_tables_test.clj
```diff
@@ -12,21 +12,30 @@
...
+       :model/Table            {new-inactive-table-id :id}             {:db_id       db-id
+                                                                        :active      false
+                                                                        :updated_at  one-day-less-than-threshold}
+       :model/Table            {old-inactive-table-id :id}             {:db_id       db-id
+                                                                        :active      false
+                                                                        :updated_at  older-than-threshold}
+       :model/Card             {_inactive-card :id}                    {:database_id db-id
+                                                                        :archived    true
+                                                                        :table_id    old-inactive-table-id}
+       :model/Table            {old-inactive-table-active-card-id :id} {:db_id       db-id
```
calherries, line 31: This confused me a little, since it reads like a card ID
```suggestion
       :model/Table            {old-inactive-table-active-with-card-id :id} {:db_id       db-id
```

calherries (approved):

PR #35695, opened 2023-11-14
Returns 404 when get a non-existing card

### LINKED ISSUE #35692: returns 500 when visit an non-existing card
### Describe the bug

When visit a card that is not existed, MB shows an `Something's gone wrong` message and the `GET /api/card/:id` returns 500

### To Reproduce

Go to `hostname/question/99999999` (assuming you don't have a card with id = `99999999`)

### Expected behavior

the API should returns 404 and UI shoudl show not found message

### Logs

_No response_

```JSON
master
```

### Severity

P3

### Additional context

_No response_

qnkhuat:
Fixes https://github.com/metabase/metabase/issues/35692

### DIFF
--- a/src/metabase/api/card.clj
+++ b/src/metabase/api/card.clj
@@ -195,6 +195,7 @@
    ignore_view [:maybe :boolean]}
   (let [raw-card (t2/select-one Card :id id)
         card (-> raw-card
+                 api/read-check
                  (t2/hydrate :creator
                              :dashboard_count
                              :parameter_usage_count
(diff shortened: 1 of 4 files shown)


src/metabase/api/card.clj
```diff
@@ -195,6 +195,7 @@
    ignore_view [:maybe :boolean]}
   (let [raw-card (t2/select-one Card :id id)
         card (-> raw-card
+                 api/read-check
```
qnkhuat, line 198: read-check throws 404 if the object is nil, so it serves our need here

calherries (approved):
Sweet

PR #35703, opened 2023-11-14
Track dashboard_saved snowplow event

### LINKED ISSUE #34751: Snowplow: Track dashboard save event
## Context
[Analytics spec](https://www.notion.so/metabase/Dashboard-saved-b3bb8eb78bd04c2f82b2ec9ecd236db9?pvs=4)

## Notes
- I figure we can track this on the client but open to feedback there.
- We already track a 'dashboard created' event so we could consolidate, but my assumption is that having this as a distinct event will be easier at this point, since it has unique properties.

Conversation that prompted this in [Slack](https://metaboat.slack.com/archives/C057T1QTB3L/p1697486450725169).

calherries:
Resolves https://github.com/metabase/metabase/issues/34751

Evolves the existing dashboard events schema, adding a new nullable field `duration_milliseconds`, and a new event type, `dashboard_saved`. 

The `dashboard_saved` event is fired from the `updateDashboardAndCards` action creator, which is called if and only if the dashboard is saved from editing mode. This event will not be tracked if a dashboard is updated another way, such as editing the dashboard name from a collection page.

### DIFF
--- a/frontend/src/metabase/dashboard/actions/save.js
+++ b/frontend/src/metabase/dashboard/actions/save.js
@@ -8,6 +8,7 @@ import Dashboards from "metabase/entities/dashboards";
 import { CardApi } from "metabase/services";
 import { clickBehaviorIsValid } from "metabase-lib/parameters/utils/click-behavior";
 
+import { trackDashboardSaved } from "../analytics";
 import { getDashboardBeforeEditing } from "../selectors";
 
 import { fetchDashboard } from "./data-fetching";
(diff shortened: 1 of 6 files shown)


frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -22,6 +23,7 @@ export const updateDashboardAndCards = createThunkAction(
   UPDATE_DASHBOARD_AND_CARDS,
   function () {
     return async function (dispatch, getState) {
+      const startTime = Date.now();
```
alxnddr, line 26: What do you think about using `performance.now();` which should be more suitable for performance measurements?

alxnddr (approved):
Looks good! Please have a look at one suggestion before merging

frontend/src/metabase/dashboard/actions/save.js
calherries: Yes, done [b126831](https://github.com/metabase/metabase/pull/35703/commits/b126831b5492e473cddeea4ae01da790927f0fa2)

PR #35705, opened 2023-11-14
Fix column filter type checks

ranquild:
When splitting the filter picker I thought about that it's incorrect that we need to handle filter expression shape vs column type mismatches in the component. In this PR I've moved these checks to `metabase-lib`.

### DIFF
--- a/frontend/src/metabase-lib/filter.ts
+++ b/frontend/src/metabase-lib/filter.ts
@@ -3,6 +3,14 @@ import moment from "moment-timezone";
 import * as ML from "cljs/metabase.lib.js";
 
 import {
+  isBoolean,
+  isTime,
+  isDate,
+  isCoordinate,
+  isString,
+  isNumeric,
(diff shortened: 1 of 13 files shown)


frontend/src/metabase-lib/filter.ts
```diff
@@ -493,23 +521,20 @@ export function filterParts(
   );
 }
 
-export function isCustomFilter(
+export function isColumnFilter(
```
kulyk, line 524: nit: maybe one day we'll add first-class support for multi-column filters, and then this name might get confusing. WDYT about a name like `isStandardFilter` or something? Or maybe we can keep `isCustomFilter`

kulyk (approved):

frontend/src/metabase-lib/filter.ts
ranquild: Fixed, changed to `isStandardFilter`

PR #34258, opened 2023-09-29
Dynamic metric x-rays

markbastian:
# Overview 
Create a newly designed pipeline within metabase.automagic-dashboards.core that generates cards for metrics without hackery and is simple and extensible. The old design had no real pipeline concept and mixed concerns, such as creating cards and queries at the same time.

This new pipeline has the following stages:
- Identify interesting items (metrics, dimensions, and filters). This is contained in the “interesting” ns.
- Identify associations of those items and assemble dashcards based on those associations, contained in the “combination” namespace.
- Gather all of this together into a final dashboard

## Stage 1: Generate "interesting" inputs for the automatic dashboard pipeline.

In this context, "interesting" means "grounded" values. In particular, the most interesting values of all are metrics. Metrics are intrinsically interesting and can be displayed on their own. Dimensions and filters, while not interesting on their own, can be combined with metrics to add more interest to the metric. In MBQL parlance, metrics are aggregates, dimensions are breakouts, and filters are filters. However, a user-defined metric may go beyond a simple aggregate.

This work is done by the `metabase.automagic-dashboards.interesting/identify` function, which takes an object to be analyzed for interestingness and a data structure consisting of templates for interesting combinations of metrics, dimensions, and filters. In this stage, we return grounded metrics (inherently interesting) along with grounded dimensions and filters that can be combined with our grounded metrics downstream for added interest.

  The template arguments are defined in terms of Dimensions, Metrics, and Filters. These are *named* values, such as:
   - Dimension:
     - GenericNumber
     - Timestamp
     - Country
     - Longitude
     - Latitude
     - Income
     - Discount
   - Metric:
     - Count - Dimensionless
     - Sum - A metric over a single field
     - AverageDiscount - A metric defined by the Income and Discount fields (as an example)
   - Filter:
     - Last30Days - A named quantity that is defined by one or more constituent Dimensions

Template Metrics and Filters are made up of some combination of field references (Dimensions). These are referenced using the Dimension names (e.g. Avg of some GenericNumber) despite these constituent fields technically not being Dimensions. Metrics and Dimensions should be thought of as orthogonal concerns, but for our matching algorithm, this is how constituent fields are selected.

The "grounding" process binds individual fields to named Dimensions as well as constituent elements of Filter and Metric definitions.

Note that the binding process is 1:N, where a single dimension may match to multiple fields. A field can only bind to one dimension.

## Stage 2: Generate "interesting" combinations of metrics, dimensions, and filters.

[...]

PR #35706, opened 2023-11-14
Add repro test for #14648 'Pivot action on a scalar in an x-ray results in a broken visualization'

### LINKED ISSUE #14648: Trying to use a pivot action on a scalar in an x-ray results in a broken visualization
This is on `master`.

- Do an x-ray on a table in the sample dataset, like orders
- Click on the value `18,760` and select a Category or Time breakout, like `Source`
- You'll arrive at a visualization that has "forgotten" what its breakout should be.

Curiously, if you break out by a location, like State, you won't encounter this behavior. 🤷🏼 

forgets-breakout

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/14648

actually the issue is fixed with https://github.com/metabase/metabase/pull/34258

before https://github.com/metabase/metabase/pull/34258 response from BE

```
"visualization_settings": {
    "graph.series_labels": [
        "Number of orders"
    ],
    "graph.metrics": [
        "count"
    ],
    "graph.dimensions": []
}
```

after
```
"visualization_settings": {
    "graph.series_labels": [
        null
    ],
    "graph.metrics": [
        null
    ],
    "graph.dimensions": null
}
``` 

### Description

Adds repro test for the linked issue.

[Stress test result](https://github.com/metabase/metabase/actions/runs/6863725505/job/18664060085)

### How to verify

reset your local branch to 87f6384d07 and run this test - it should fail
test should pass on master

### Checklist

### DIFF
--- a/e2e/test/scenarios/dashboard/x-rays.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/x-rays.cy.spec.js
@@ -262,11 +262,16 @@ describe("scenarios > x-rays", { tags: "@slow" }, () => {
   });
 
   it("should correctly apply breakout in query builder (metabase#14648)", () => {
-    const options = { timeout: 10000 };
-
     cy.visit(`/auto/dashboard/table/${ORDERS_ID}`);
 
-    getDashboardCards().contains("18,760", options).click();
+    // wait for data from all dashcards to be sure data is loaded
+    const NUMBER_OF_DASHCARDS = 16;
+    cy.intercept("POST", "/api/dataset").as("dataset");
+    cy.wait(Cypress._.times(NUMBER_OF_DASHCARDS, () => "@dataset"));
+    // in case number of dashcards is changed in the future
+    cy.wait("@dataset.all").should("have.length", NUMBER_OF_DASHCARDS);
+
+    getDashboardCards().contains("18,760").click();
 
     popover().within(() => {
       cy.findByText("Break out by…").click();
--- a/e2e/test/scenarios/dashboard/x-rays.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/x-rays.cy.spec.js
@@ -261,6 +261,35 @@ describe("scenarios > x-rays", { tags: "@slow" }, () => {
     cy.findByText("463");
   });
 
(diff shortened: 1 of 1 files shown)


nemanjaglumac:
Please stress-test this change 20 times to make sure we're not introducing a new flake.
Thanks

e2e/test/scenarios/dashboard/x-rays.cy.spec.js
```diff
@@ -261,6 +261,35 @@ describe("scenarios > x-rays", { tags: "@slow" }, () => {
     cy.findByText("463");
   });
 
+  it("should correctly apply breakout in query builder (metabase#14648)", () => {
+    const options = { timeout: 10000 };
+
+    cy.visit(`/auto/dashboard/table/${ORDERS_ID}`);
+
+    getDashboardCards().contains("18,760", options).click();
```
ranquild, line 269: Please add `cy.wait` calls here, intercepting BE calls instead of manual timeout

ranquild: I think we could wait for the dashboard response, find the desired card, and wait for the specific dataset response. Wdyt?

nemanjaglumac: Do you mean `/automagic-dashboards/table/:id`?
image

"Waiting for the specific dataset" part might be a bit problematic because I'm seeing a ton of cancelled requests first:
image

That's my only concern.
But I like the suggested approach in general. Finding a specific card and waiting only for its dataset is the key here.

nemanjaglumac: Which is failing.
As I said, the timeout to find the string in a card might be a better approach here.

uladzimirdev:
@nemanjaglumac @ranquild I found another way - we wait for all dataset, but stop waiting after desired response came.

here is a [stress run](https://github.com/metabase/metabase/actions/runs/6886845975/job/18733126910)

nemanjaglumac (approved):
Thanks for doing the extra checks!
I didn't run this locally, but the test is easy to read and the stress-test gives me enough confidence to approve.

ranquild (approved):

PR #35708, opened 2023-11-14
Make Chromatic workflow 3rd party friendly

nemanjaglumac:
A follow up after https://github.com/metabase/metabase/pull/34951. This PR follows the same logic as #35643 and it makes Chromatic workflow 3rd party CI friendly.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -6,12 +6,27 @@ on:
       - "master"
   pull_request:
     types: [synchronize, labeled, unlabeled]
+  workflow_run:
+    workflows: [External Trigger Filter]
+    types: [completed]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
+  external-filter:
+    if: github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-22.04
+    timeout-minutes: 3
+    steps:
+      - run: |
+          if ${{ github.event.pull_request.head.repo.full_name == github.repository }}; then
+            echo 'Running Chromatic workflow within the organization.'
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -29,8 +44,10 @@ jobs:
           filters: .github/file-paths.yaml
 
   fe-chromatic:
-    needs: files-changed
-    if: contains(github.event.pull_request.labels.*.name, 'chromatic') || (github.ref_name == 'master' && needs.files-changed.outputs.frontend_all == 'true')
+    needs: [files-changed, external-filter]
+    if: |
+      (needs.external-filter.result == 'success' && contains(github.event.pull_request.labels.*.name, 'chromatic')) ||
```
uladzimirdev, line 49: in which cases external-filter will not finish with "success"?

nemanjaglumac: In case it's skipped.
And it will be skipped for external contributors on `pull_request` trigger.

```yml
if: |
  # Is it an internal PR? if yes -> proceed with the `pull_request` trigger 
  github.event.pull_request.head.repo.full_name == github.repository ||
  # If it is a fork it requires a `workflow-filter.yml` to run first. Otherwise, this is false and the job is skipped.
  # Once approved, Chromatic will be triggered the second time by a `workflow_run`
  github.event.workflow_run.conclusion == 'success'
```

For more info: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

nemanjaglumac: @uladzimirdev to illustrate this with an example for E2E tests (I tested this in my fork).

This is what happens once we approve 3rd party PR.
image

The first trigger is `pull_request` - it's skipped.
The second trigger says _completed_, which means that some other workflow completed and has triggered E2E tests (hence, on `workflow_run`.

The full mechanism is explained in this PR:
https://github.com/metabase/metabase/pull/34951

uladzimirdev (approved):
LGTM

PR #35713, opened 2023-11-14
Prevent updating a database from closing connections unnecessarily when updating `database-enable-actions` setting

### LINKED ISSUE #27877: When adding a db, immediately updating `database-enable-actions` prevents the db from finishing initial sync
I discovered this while setting up automated testing for actions DBs, If we update the DB immediately after adding it, before the initial sync had finished, the DB would always report that the initial sync was "incomplete"

This was an easy workaround in our e2e test setup: wait for the initial sync to complete before enabling actions, but I fear that real customers with large DBs that take minutes to sync initially could get stuck.

calherries:
Fixes https://github.com/metabase/metabase/issues/27877, see [this comment](https://github.com/metabase/metabase/issues/27877#issuecomment-1810283147) for an explanation of the root cause.

Avoids calling `driver/notify-database-updated` if we update only the `database-enable-actions` setting for a database.

That method has the docstring:
https://github.com/metabase/metabase/blob/66af1e4cd7aa2a9afa657d1ef2b0cfba36842918/src/metabase/driver.clj#L703-L706

It shouldn't be called unless there's a reason to update the state of related resources. In the case of `database-enable-actions`, there's no reason why any driver-specific resources need to be updated.

### DIFF
--- a/src/metabase/events/driver_notifications.clj
+++ b/src/metabase/events/driver_notifications.clj
@@ -5,6 +5,7 @@
   this writing, the SQL JDBC driver 'superclass' is the only thing that implements this method, and does so to close
   connection pools when database details change or when they are deleted."
   (:require
+   [medley.core :as m]
    [metabase.driver :as driver]
    [metabase.events :as events]
    [metabase.util.log :as log]
@@ -15,10 +16,15 @@
 (derive :event/database-delete ::event)
 
 (methodical/defmethod events/publish-event! ::event
-  [topic {database :object :as _event}]
+  [topic {database :object, previous-database :previous-object :as _event}]
   ;; try/catch here to prevent individual topic processing exceptions from bubbling up.  better to handle them here.
   (try
-   ;; notify the appropriate driver about the updated database
-   (driver/notify-database-updated (:engine database) database)
-   (catch Throwable e
-     (log/warnf e "Failed to process driver notifications event. %s" topic))))
+    ;; notify the appropriate driver about the updated database to release any resources it may be holding onto.
+    ;; don't notify the driver if the changes should not impact the state of any resource.
+    (let [paths-that-dont-impact-resource-state [[:updated_at]
+                                                 [:settings :database-enable-actions]]]
+      (when (not= (reduce m/dissoc-in database paths-that-dont-impact-resource-state)
+                  (reduce m/dissoc-in previous-database paths-that-dont-impact-resource-state))
+        (driver/notify-database-updated (:engine database) database)))
+    (catch Throwable e
+      (log/warnf e "Failed to process driver notifications event. %s" topic))))
--- a/src/metabase/events/driver_notifications.clj
+++ b/src/metabase/events/driver_notifications.clj
@@ -19,12 +19,15 @@
   [topic {database :object, previous-database :previous-object :as _event}]
   ;; try/catch here to prevent individual topic processing exceptions from bubbling up.  better to handle them here.
   (try
-    ;; notify the appropriate driver about the updated database to release any resources it may be holding onto.
-    ;; don't notify the driver if the changes should not impact the state of any resource.
-    (let [paths-that-dont-impact-resource-state [[:updated_at]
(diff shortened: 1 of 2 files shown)


test/metabase/api/database_test.clj
```diff
@@ -511,6 +517,42 @@
...
+        (let [;; 2. start a long running process on another thread that uses a connection
+              connections-stay-open? (future
+                                       (sql-jdbc.execute/do-with-connection-with-options
+                                        driver/*driver*
+                                        db
+                                        nil
+                                        (fn [^Connection conn]
+                                          ;; sleep long enough to make sure the PUT request below finishes processing,
+                                          ;; including any async operations that it might trigger
+                                          (Thread/sleep 1000)
```
calherries, line 541: If you have a smarter way to do this, I'm all ears.

src/metabase/events/driver_notifications.clj
```diff
@@ -15,10 +16,18 @@
...
-   (driver/notify-database-updated (:engine database) database)
-   (catch Throwable e
-     (log/warnf e "Failed to process driver notifications event. %s" topic))))
+    ;; notify the appropriate driver about the updated database to release any related resources, such as connections.
+    ;; avoid notifying if the changes shouldn't impact the observable behaviour of any resource, otherwise drivers might
+    ;; close connections or other resources unnecessarily (metabase#27877).
+    (let [;; remove data that should not impact the observable state of any resource before comparing
+          remove-irrelevant-data (fn [db]
+                                   (reduce m/dissoc-in db [[:updated_at]
+                                                           [:settings :database-enable-actions]]))]
```
calherries, line 28: There are probably other properties of databases that can go in this list, like `:name`. But let's keep this PR conservative and not risk creating more bugs

test/metabase/api/database_test.clj
piranha: Heh, that's one interesting problem, given the 'any async operations' qualifier. :) I wonder if it makes sense to use with-redefs to capture async stuff?.. Or do we actually have something for like gathering/capturing async stuff for tests?

piranha: Oh, how do you like idea like that: redefine `driver/notify-database-updated` to check it's not called?

calherries: > redefine driver/notify-database-updated to check it's not called
Would that solve the issue? `driver/notify-database-updated` might not be called at the time of the check for two reasons:
1. `driver/notify-database-updated` has not been called and won't be called in the future
2. `driver/notify-database-updated` will be called in the future by code that executes in the future

If we check whether `driver/notify-database-updated` is called, we need to make sure that we check it after some time so 2 is not a possibility. Which means we need to do a `Thread/sleep` anyway.

calherries: At least the way this test is implemented, it has no mention of `driver/notify-database-updated`, which is more of an implementation detail than a property we actually want to test here which is that the connection never gets closed.

piranha: Ah, right, absolutely! Well, the only improvement is grabbing onto a start of various async stuff, and it feels a bit out of scope for this PR. :)

sansolovyov (approved):
LGTM, but we should revisit this Thread/sleep if we ever get an idea how to wait for an endpoint and all initiated async processes to finish.

test/metabase/api/database_test.clj
calherries: Yeah. Do you have a quick explanation on what that could look like?

piranha: Well, if we have just a single starting point for async stuff, `with-redefs` could handle this: collect all the started processes in an atom and return it or wait on them (up to implementation). Could be a nice macro like `(gather-async-processes ...)`, of course.

The bigger problem is if there are a lot of various ways to run them (like `future`, and separate thread pools, and `core.async`), but I think that requires clean up/unification then. :)

PR #35715, opened 2023-11-14
seperate data from side effects for `record-event!`

escherize:
As we discussed in https://github.com/metabase/metabase/pull/35435#discussion_r1385636529.

This PR adds a seperation between _generating the data_ that goes into the audit log, and the _construction_ of that data. It let's us write better, faster tests. The added tests don't depend on the database, but still test properties about audit log entries.

This also adds quickcheck tests for user-id and model-id precedence 

e.g. if I pass in a model-id and also have an id on my object, which one gets inserted? Now there's a test for that!

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/audit_db.clj
+++ b/enterprise/backend/src/metabase_enterprise/audit_db.clj
@@ -227,4 +227,5 @@
   (u/prog1 (maybe-install-audit-db)
    (let [audit-db (t2/select-one :model/Database :is_audit true)]
        ;; prevent sync while loading
-     ((sync-util/with-duplicate-ops-prevented :sync-database audit-db (partial maybe-load-analytics-content! audit-db))))))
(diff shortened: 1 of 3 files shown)


enterprise/backend/src/metabase_enterprise/audit_db.clj
```diff
@@ -227,4 +227,5 @@
   (u/prog1 (maybe-install-audit-db)
    (let [audit-db (t2/select-one :model/Database :is_audit true)]
        ;; prevent sync while loading
-     ((sync-util/with-duplicate-ops-prevented :sync-database audit-db (partial maybe-load-analytics-content! audit-db))))))
+     ((sync-util/with-duplicate-ops-prevented :sync-database audit-db
+        (fn [] (maybe-load-analytics-content! audit-db)))))))
```
escherize, line 231: Using partial when we don't pass in any args felt weird to me, I think this reads a little better.

noahmoss (approved):

PR #35716, opened 2023-11-14
Fix broken Root Collection permissions

iethree:
### Description

Our instance analytics collection type check was breaking collection permissions for the root collection because it's object looks different than all other collections.

Before | After
--- | ---
Screen Shot 2023-11-14 at 11 12 59 AM | Screen Shot 2023-11-14 at 11 12 38 AM

I also added some unit and e2e tests for this. Crazy that we didn't have any does-this-page work coverage on this page at all. Now we do.

### Checklist

## PS

I also played around with adding some error boundaries to this component so that if the main permissions table panel crashed, the sidebar could still work instead of the the whole page dying. However, since the problem here was in the redux selector that sits above both of those components, it wasn't possible 😢  This is good evidence that untangling/simplifying some of the structure of our admin pages could improve not just the code's usability, but also allow us to build a more robust user experience.

### DIFF
--- a/e2e/test/scenarios/collections/permissions.cy.spec.js
+++ b/e2e/test/scenarios/collections/permissions.cy.spec.js
@@ -9,9 +9,12 @@ import {
   openCollectionMenu,
   openCollectionItemMenu,
   modal,
+  setTokenFeatures,
+  sidebar,
 } from "e2e/support/helpers";
 
 import { USERS } from "e2e/support/cypress_data";
+import { FIRST_COLLECTION_ID } from "e2e/support/cypress_sample_instance_data.js";
 import { displaySidebarChildOf } from "./helpers/e2e-collections-sidebar.js";
 
 const PERMISSIONS = {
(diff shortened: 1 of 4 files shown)


enterprise/frontend/src/metabase-enterprise/collections/utils.unit.spec.ts
```diff
@@ -43,5 +43,11 @@ describe("Collections plugin utils", () => {
       const collection = createMockCollection({ type: "instance-analytics" });
       expect(getCollectionType(collection).icon).toBe("audit");
     });
+
+    it("root collection", () => {
+      const collection = createMockCollection();
+      expect(getCollectionType(collection).type).toBe(null);
+      expect(getCollectionType({}).type).toBe(null);
```
iethree, line 50: before this change this change, this would have thrown an error 🚒

e2e/test/scenarios/collections/permissions.cy.spec.js
```diff
@@ -431,6 +434,37 @@ describe("collection permissions", () => {
 
     cy.findByTestId("select-button").findByText("Our analytics");
   });
+
+  it("should load the collection permissions admin pages", () => {
```
iethree, line 438: 😬  crazy that we had zero test coverage on this page

npfitz (approved):

PR #35549, opened 2023-11-09
Add don't bin option for temporal bucketing

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/19684
Closes https://github.com/metabase/metabase/issues/35550

### Description

Add `Don't bin` option for temporal bucketing.

Doesn't cover https://github.com/metabase/metabase/issues/35546

### How to verify

Create a new question like in [figma](https://www.figma.com/file/HPg8IDPGo9seZafglWmhIk/Querying-Bug-Fixes?type=design&node-id=101-30315&mode=design&t=zAUnLpMXJI49RCXz-0)
`Don't bin` should be on the list

### Demo

https://github.com/metabase/metabase/assets/125459446/e67dbc3d-5765-46a0-8f8f-1379dd2d7157

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #35721, opened 2023-11-14
Manual backport "Add don't bin option for temporal bucketing (#35549)"

uladzimirdev:
backport #35549


uladzimirdev:
It looks like this PR shouldn't be backported as it's based on https://github.com/metabase/metabase/pull/32953 which wasn't backported

Issue #33499, opened 2023-08-25
Questions or dashboards with lots of metadata take a lot of time to dismiss the spinner


paoliniluis:
### Describe the bug

Seems that if a query (many custom columns or joins) or dashboard (many questions or many filters) has lots of properties, it will take a lot of time to render.

E.g.: question 13003 in our internal instance
![big_question](https://github.com/metabase/metabase/assets/1711649/d8e1765f-c4d6-428e-9892-7cd33cd27e05)

We need to think a way of giving the user a faster response in those cases. Probably a sibling of https://github.com/metabase/metabase/issues/33220

### To Reproduce

1) Build a big question or dashboard, with lots of stuff
2) do a fresh request

### Expected behavior

We should make the UI smoother, spinner should appear the least amount of time as possible

### Logs

NA

### Information about your Metabase installation

```JSON
Since the beginning of the times. I guess that the spinner dismisses once all the endpoints finished loading, but in a question or dashboard you don't want to have everything loaded to interact
```

### Severity

P2

paoliniluis:
Second example Dashboard 2116

alxnddr:
I changed the label to the BE because frontend shows loader correctly since it waits for the response

markbastian:
The slow endpoint is https://stats.metabase.com/api/dashboard/:dashboard-id.

Here's what I've been able to do to attempt to reproduce this issue so far:

* @paoliniluis referenced https://stats.metabase.com/dashboard/2117 as a similarly slow dashboard to load
* To try to reproduce this locally, I created a large dashboard (36 cards) with a variety of queries (nested, custom columns, multiple sources, native, visual)
* Made as many filters as I could think of, many with enumerated values
* The json output of the referenced slow response on stats is 13770 lines at 388639 bytes
* My local response body is 15032 lines of formatted json at 443866 bytes

My expected response time would be higher locally as the dashboard is larger.

Results:
* That stats board (baseline) takes 4.39s to load
* My attempted reproduction loads in under a second (297ms)
* Assuming a similar level of complexity, `master` is an order of magnitude faster to load

Here's a side by side view of the two scenarios. Note that the local dashboard (left side) loads much faster than the baseline. Also note that qualitatively, individual dashcard loading appears similar on both sides.

![slowdash1](https://github.com/metabase/metabase/assets/8381441/938c3368-a81a-4872-88af-f07abe9528f2)

[...]

PR #35723, opened 2023-11-14
Add otel instrumentation to diagnose 33499

markbastian:
This PR contains changes and a potential fix to the problems described in #33499 (Dashboards taking a long time to load on stats). The underlying issue appears to be that there is resource contention on the backend that is manifesting itself in the synchronous calls to `metabase.events/publish-event!`.

Instrumentation was added to this PR to better understand how long each function takes as well as to better understand the context in which already-instrumented functions (e.g. server api calls, db calls) are called. For example, we already get spans for JDBC operations, but we don't know what function is calling these operations. With the right in-app spans, we can know what functions are making the calls as well as get better information on where timeouts are occurring.

### Instrumentation

- The `com.github.steffan-westcott/clj-otel-api` [library](https://github.com/steffan-westcott/clj-otel) was added so that we could add custom [Open Telemetry](https://opentelemetry.io/) instrumentation to metabase in a Clojure-friendly way.
- The `com.clojure-goes-fast/clj-async-profiler` was added to the `:dev` profile along with the `"-Djdk.attach.allowAttachSelf"` flag (again, dev-only) for local profiling.

The above, primarily the first bullet, allows for more detailed analysis of our code, as shown below.

jaeger

Note that in the diagram, the problem described in #33499 is not shown, but the comments and images from Grafana on that issue illustrate the value of better instrumentation.

### Short term fix

This PR includes a new function, `publish-event-async!`, in `metabase.events` that is used in the `GET "api/dashboard/:id"` endpoint and should resolve the immediate performance issue. However, based on [this slack thread](https://metaboat.slack.com/archives/CKZEMT1MJ/p1697604232592589) we may want to either not use this function or only use it as a short term fix.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -36,6 +36,7 @@
   com.github.seancorfield/honeysql          {:mvn/version "2.4.1066"}           ; Honey SQL 2. SQL generation from Clojure data maps
   com.github.seancorfield/next.jdbc         {:git/url "https://github.com/seancorfield/next-jdbc.git"
                                              :sha     "abd926cde3730087d3729dcea0ce0ab7ef1194f2"} ; for https://github.com/seancorfield/next-jdbc/issues/245; when this makes it to a release we can switch to that.
+  com.github.steffan-westcott/clj-otel-api  {:mvn/version "0.2.4.1"}            ; Telemetry library
   com.github.vertical-blank/sql-formatter   {:mvn/version "2.0.4"}              ; Java SQL formatting library https://github.com/vertical-blank/sql-formatter
   com.google.guava/guava                    {:mvn/version "32.1.2-jre"}         ; dep for BigQuery, Spark, and GA. Require here rather than letting different dep versions stomp on each other — see comments on #9697
   com.fasterxml.jackson.core/jackson-databind
--- a/deps.edn
+++ b/deps.edn
@@ -206,27 +206,28 @@
   ;;    (dev/start!)
   :dev
   {:extra-deps
-   {com.clojure-goes-fast/clj-async-profiler {:mvn/version "1.1.1"}
-    clj-http-fake/clj-http-fake              {:mvn/version "1.0.4"
-                                              :exclusions  [slingshot/slingshot]}
-    clj-kondo/clj-kondo                      {:mvn/version "2023.09.07"} ; this is not for RUNNING kondo, but so we can hack on custom hooks code from the REPL.
-    cloverage/cloverage                      {:mvn/version "1.2.4"}
-    com.gfredericks/test.chuck               {:mvn/version "0.2.14"} ; generating strings from regexes (useful with malli)
-    djblue/portal                            {:mvn/version "0.46.0"} ; ui for inspecting values
-    io.github.camsaul/humane-are             {:mvn/version "1.0.2"}
-    io.github.metabase/hawk                  {:sha "9c97bcb6d4de116325e651b40973fd0a75b7ae21"}
-    jonase/eastwood                          {:mvn/version "1.4.0" ; inspects namespaces and reports possible problems using tools.analyzer
-                                              :exclusions
-                                              [org.ow2.asm/asm-all]}
-    lambdaisland/deep-diff2                  {:mvn/version "2.10.211"} ; way better diffs
-    methodical/methodical                    {:mvn/version "0.15.1"} ; drop-in replacements for Clojure multimethods and adds several advanced features
-    org.clojure/algo.generic                 {:mvn/version "0.1.3"}
-    pjstadig/humane-test-output              {:mvn/version "0.11.0"}
-    reifyhealth/specmonstah                  {:mvn/version "2.1.0"
-                                              :exclusions  [org.clojure/clojure
-                                                            org.clojure/clojurescript]} ; lets you write test fixtures that are clear, concise, and easy to maintain (clojure.spec)
-    ring/ring-mock                           {:mvn/version "0.4.0"}                     ; creating Ring request maps for testing purposes
-    talltale/talltale                        {:mvn/version "0.5.13"}}                   ; generates fake data, useful for prototyping or load testing
+   {com.clojure-goes-fast/clj-async-profiler
+                                 {:mvn/version "1.1.1"}                     ; Enables local profiling and heat map generation
+    clj-http-fake/clj-http-fake  {:mvn/version "1.0.4"
+                                  :exclusions  [slingshot/slingshot]}
+    clj-kondo/clj-kondo          {:mvn/version "2023.09.07"}                ; this is not for RUNNING kondo, but so we can hack on custom hooks code from the REPL.
+    cloverage/cloverage          {:mvn/version "1.2.4"}
+    com.gfredericks/test.chuck   {:mvn/version "0.2.14"}                    ; generating strings from regexes (useful with malli)
+    djblue/portal                {:mvn/version "0.46.0"}                    ; ui for inspecting values
+    io.github.camsaul/humane-are {:mvn/version "1.0.2"}
+    io.github.metabase/hawk      {:sha "9c97bcb6d4de116325e651b40973fd0a75b7ae21"}
+    jonase/eastwood              {:mvn/version "1.4.0"                      ; inspects namespaces and reports possible problems using tools.analyzer
+                                  :exclusions
+                                  [org.ow2.asm/asm-all]}
+    lambdaisland/deep-diff2      {:mvn/version "2.10.211"}                  ; way better diffs
+    methodical/methodical        {:mvn/version "0.15.1"}                    ; drop-in replacements for Clojure multimethods and adds several advanced features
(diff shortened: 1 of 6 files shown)


src/metabase/api/dataset.clj
```diff
@@ -51,29 +52,31 @@
       :or   {context       :ad-hoc
              export-format :api
              qp-runner     qp/process-query-and-save-with-max-results-constraints!}}]
```
markbastian, line 54: While not strictly part of the #33499 issue, `dataset` is a slow endpoint that needs instrumentation.

deps.edn
```diff
@@ -205,33 +206,36 @@
...
-    lambdaisland/deep-diff2      {:mvn/version "2.10.211"}                  ; way better diffs
-    methodical/methodical        {:mvn/version "0.15.1"}                    ; drop-in replacements for Clojure multimethods and adds several advanced features
-    org.clojure/algo.generic     {:mvn/version "0.1.3"}
-    pjstadig/humane-test-output  {:mvn/version "0.11.0"}
-    reifyhealth/specmonstah      {:mvn/version "2.1.0"
-                                  :exclusions  [org.clojure/clojure
-                                                org.clojure/clojurescript]} ; lets you write test fixtures that are clear, concise, and easy to maintain (clojure.spec)
-    ring/ring-mock               {:mvn/version "0.4.0"}                     ; creating Ring request maps for testing purposes
-    talltale/talltale            {:mvn/version "0.5.13"}}                   ; generates fake data, useful for prototyping or load testing
+   {com.clojure-goes-fast/clj-async-profiler {:mvn/version "1.1.1"}
```
dpsutton, line 209: this diff is too large to see what is going on. Ideally we don't reformat everything

src/metabase/api/dashboard.clj
```diff
@@ -182,26 +183,32 @@
...
-                  :param_values
-                  [:collection :is_personal])
-      collection.root/hydrate-root-collection
-      api/check-not-archived
-      hide-unreadable-cards
-      add-query-average-durations))
+  (span/with-span!
+    {:name       "get-dashboard"
+     :attributes {:dashboard/id id}}
+    (let [dashboard (-> (t2/select-one :model/Dashboard :id id) api/check-404 api/read-check)
```
dpsutton, line 189: the check 404 and read-check are duplicative. read-check has a check 404 inside of it.

deps.edn
markbastian: If we want to align the opening `{:mvn/version`, we need to indent the rest of the form since `com.clojure-goes-fast/clj-async-profiler` is long.

markbastian: This was the only change in this block `com.clojure-goes-fast/clj-async-profiler {:mvn/version "1.1.1"}` and the rest was manually reformatted to align the values.

src/metabase/api/dashboard.clj
markbastian: Fixed

src/metabase/events.clj
```diff
@@ -155,3 +159,21 @@
    ;; and is important for distinguishing view events triggered when pinned cards are 'viewed'
    ;; when a user opens a collection.
    :context      (:context object)})
+
+(def ^:private async-publish-event-timeout 10000)
+
+(defn publish-event-async!
```
adam-james-v, line 165: If it is the case that we want this as only a short-term fix, maybe we should include that in the code somehow? 
Perhaps log a warning? Or write a ;;todo: remove this in favor of a more complete synchronous solution. Or something.

markbastian: I'm fine with a TODO.

markbastian: TODO message pushed.

dpsutton:
I'm torn on adding this instrumentation library here like this. I like the idea of instrumentation in general but don't like these ad-hoc scope inducing forms everywhere. The introduction here feels a bit ad-hoc instead of deliberate. Especially in service of a P1.

My other thought is at the implementation of the threading for the async event function. It spawns two threads in an unbounded thread pool. I'm wondering if we perhaps want a bit more control over the number of threads we are spawning.

markbastian:
> I'm torn on adding this instrumentation library here like this. I like the idea of instrumentation in general but don't like these ad-hoc scope inducing forms everywhere. The introduction here feels a bit ad-hoc instead of deliberate. Especially in service of a P1.
> 
> My other thought is at the implementation of the threading for the async event function. It spawns two threads in an unbounded thread pool. I'm wondering if we perhaps want a bit more control over the number of threads we are spawning.

Yeah, I hear you. Here are some specific thoughts:
* Opentelemetry is a standard and is what is used on stats. We just don't see it on dev. This particular library isn't a one-off aside from what we specifically want to instrument from Metabase.
* Speaking of which, I believe it would be ideal to have spans covering _most_ of the full width of our API calls. The implicit 100% span you get from baked-in HTTP instrumentation and the smattering of SQL instrumentation you get from JDBC calls is useful, but there are large gaps that don't provide further context. If each HTTP call had a few callees (e.g. in the case of GET api/dashboard/:id you have 3 top level calls: get-dashboard, publish-event-async!, and with-last-edit-info). Between these three calls you should have nearly 100% span coverage at the top level.
* If we generally agree on the above, do we need to add such spans to all of our APIs? I think it's a good idea, but we need to get buy in and decide if we do it one PR at a time of make it a global effort in a small number of PRs.
* WRT the specific solution, we could use the thread pool in `metabase.async.streaming-response.thread-pool`. Thoughts?

deps.edn
noahmoss: We have a couple other deps which have the same issue, and the solution was to put the version map on the next line to keep it aligned. See `org.apache.logging.log4j/log4j-layout-template-json` for example.

src/metabase/events.clj
```diff
@@ -155,3 +159,23 @@
...
+  ;; TODO - per team discussions, this will give us a short-term fix to performance issues related to blocking DB
+  ;; operations from resource contention. However, we'd still prefer to minimize resource contention.
+  "Publish an event asynchronously with an optional timeout (default is 10s).
+  If the event fails, will log a warning."
+  ([topic event]
+   (publish-event-async! topic event async-publish-event-timeout))
+  ([topic event timeout]
+   (let [event-thread (future (publish-event! topic event))]
+     (future (Thread/sleep ^long timeout)
```
camsaul, line 174: This is kinda dangerous because the `future` threadpool is unbounded, we can potentially get to a point where we have thousands of events waiting to be handled

camsaul (commented):
event handling used to be async until recently, we switched it to sync on purpose tho because 

a) it's a lot harder to test (can't write a test that hits and API endpoint and verify a view entry was recorded, for example)
b) if you're using `core.async` it's way to easy to accidentally tie up one of the very limited number of `core.async` threads; if you're using `future` you can also footgun yourself because the threadpool is unbounded and you can potentially end up with thousands of events waiting to be handled 
c) it hides errors and performance problems with event handling.
d) performance problems in event handling end up degrading everything else **except** for the endpoint that triggered them

Doing things sync forces us to just fix the performance issues and makes the code easier to understand and test.

If recording a Dashboard view is so complicated that it seriously affects the performance of the Dashboard fetch endpoints, the right fix is to just fix that event handling code, not to sweep it back under the rug again. It's probably just some sort of N+1 problem or something.

deps.edn
piranha: I tend to add `?w=1` to an URL when there are a lot of whitespace changes.

camsaul:
Instead of having to manually add those `with-span` forms everywhere, what about adding them in more general places so they cover everything? Can we do this in an aspect-oriented manner so we get it for free everywhere, rather than having to clutter up all of our code with what will eventually be hundreds if not thousands of `with-span` forms?

- In some sort of Ring middleware so every API call gets top-level instrumented
- In [`:around` method(s)](https://github.com/camsaul/methodical#auxiliary-methods-before-after-and-around) to Toucan 2 hydrate so every hydration call gets instrumented. You could probably just do one for [`hydrate-with-strategy`](https://github.com/camsaul/toucan2/blob/master/src/toucan2/tools/hydrate.clj#L174-L183)
- In an `:around` method around one of the Toucan 2 query pipeline methods, probably [`transduce-execute-with-connection`](https://github.com/camsaul/toucan2/blob/master/src/toucan2/pipeline.clj#L43-L55), so every application database call gets instrumented
- For `publish-event!`

camsaul:
In the screenshots I see database connection strings, that seems like a bad thing to leak, right?

camsaul (approved):
👍ing so we can get this in ASAP based on discussion in Slack but let's figure out how to do this instrumentation in a more general way in the future so we don't have to manually instrument all of our API endpoints and clutter the code up

paoliniluis:
So we added telemetry but only to the dashboard endpoint, is that correct? I would love that this is added to every endpoint we have so we can see what's going on everywhere. We should make this configurable so it doesn't hurt performance

PR #35732, opened 2023-11-15
Add to view_log schema and clean up view events

### LINKED ISSUE #27782: Improved `card-view` event

noahmoss:
(Mostly) closes https://github.com/metabase/metabase/issues/27782

This PR reworks the events that we store in the `view_log`, and modifies the schema of the `view_log` table. Specifically:
* It adds `has_access` (bool) and `context` (string) fields to `view_log`
* It deletes the `card-query` event entirely. Queries are only stored in the `query_execution` table.
* It refactors the view events for cards, dashboards and tables. Most of this is the same, but as a summary:
	* Card views are recorded during `GET /api/card/:id` calls, regardless of the origin. Card views are also recorded when cards are viewed in dashboards.
	* Dashboard views are recorded during `GET /api/dashboard/:id` calls
	* Table views are recorded during `POST /api/dataset` for queries that have a simple source table. `has_access` is determined based on whether the current user has query access for that table. 
	* `can-read?` permission check failures for Cards and Dashboards are also recorded, with `has_access` = `false`
	
Things from the issue which I haven't added:
* I didn't add an `error_string` field to `view_log` because the only errors we're logging are read permission failures, which is already reflected by the `has_access` field. Since we're not logging query executions in this table, there aren't any other types of failures that are expected. Query execution errors are logged in the `query_execution` table.
* `context` is being set to `dashboard` for dashboard card views, but not set for any other views currently. The main reason is that the backend can't distinguish between normal card views and pinned collection views. We'd need a FE change to pass this in the API.
* `metadata` is still in the table for backwards compatibility and to avoid breaking views. We can drop it in the next release.

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/audit_db.clj
+++ b/enterprise/backend/src/metabase_enterprise/audit_db.clj
@@ -181,7 +181,8 @@
   :default    true
   :visibility :internal
   :setter     :none
-  :audit      :never)
+  :audit      :never
+  :doc        false)
 
 (defn- maybe-load-analytics-content!
   [audit-db]
--- a/src/metabase/api/card.clj
+++ b/src/metabase/api/card.clj
@@ -25,8 +25,7 @@
    [metabase.mbql.normalize :as mbql.normalize]
    [metabase.mbql.util :as mbql.u]
    [metabase.models
-    :refer [Card CardBookmark Collection Database PersistedInfo Pulse Table
-            ViewLog]]
+    :refer [Card CardBookmark Collection Database PersistedInfo Pulse Table]]
    [metabase.models.card :as card]
    [metabase.models.collection :as collection]
    [metabase.models.collection.root :as collection.root]
@@ -107,34 +106,6 @@
   [_ table-id]
   (t2/select Card, :table_id table-id, :archived false, {:order-by [[:%lower.name :asc]]}))
 
-(mu/defn ^:private cards-with-ids :- [:maybe [:sequential (mi/InstanceOf Card)]]
-  "Return unarchived Cards with `card-ids`.
-  Make sure cards are returned in the same order as `card-ids`; `[in card-ids]` won't preserve the order."
-  [card-ids :- [:sequential ms/PositiveInt]]
-  (when (seq card-ids)
-    (let [card-id->card (m/index-by :id (t2/select Card, :id [:in (set card-ids)], :archived false))]
-      (filter identity (map card-id->card card-ids)))))
-
-;; Return the 10 Cards most recently viewed by the current user, sorted by how recently they were viewed.
-(defmethod cards-for-filter-option* :recent
-  [_]
-  (cards-with-ids (map :model_id (t2/select [ViewLog :model_id [:%max.timestamp :max]]
-                                   :model   "card"
-                                   :user_id api/*current-user-id*
-                                   {:group-by [:model_id]
-                                    :order-by [[:max :desc]]
-                                    :limit    10}))))
-
-;; All Cards, sorted by popularity (the total number of times they are viewed in `ViewLogs`). (yes, this isn't
-;; actually filtering anything, but for the sake of simplicitiy it is included amongst the filter options for the time
-;; being).
-(defmethod cards-for-filter-option* :popular
-  [_]
-  (cards-with-ids (map :model_id (t2/select [ViewLog :model_id [:%count.* :count]]
-                                   :model "card"
-                                   {:group-by [:model_id]
-                                    :order-by [[:count :desc]]}))))
-
 ;; Cards that have been archived.
 (defmethod cards-for-filter-option* :archived
   [_]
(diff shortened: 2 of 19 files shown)


src/metabase/api/card.clj
```diff
@@ -165,9 +136,8 @@
 
 (api/defendpoint GET "/"
   "Get all the Cards. Option filter param `f` can be used to change the set of Cards that are returned; default is
-  `all`, but other options include `mine`, `bookmarked`, `database`, `table`, `recent`, `popular`, :using_model
-  and `archived`. See corresponditng implementation functions above for the specific behavior of each filter
-  option. :card_index:"
+  `all`, but other options include `mine`, `bookmarked`, `database`, `table`, `using_model` and `archived`. See
+  corresponding implementation functions above for the specific behavior of each filterp option. :card_index:"
```
noahmoss, line 140: Removed the options from this API which read from `ViewLog`, since they were unused

enterprise/backend/src/metabase_enterprise/audit_db.clj
```diff
@@ -181,7 +181,8 @@
   :default    true
   :visibility :internal
   :setter     :none
-  :audit      :never)
+  :audit      :never
+  :doc        false)
```
noahmoss, line 185: Excluding internal-only env var from the docs (unrelated change but tiny so rolling it into this PR)

src/metabase/events/recent_views.clj
```diff
@@ -3,44 +3,26 @@
...
    [metabase.models.recent-views :as recent-views]
+   [metabase.util :as u]
    [metabase.util.log :as log]
    [methodical.core :as m]))
 
 (derive ::event :metabase/event)
 
 (derive :event/dashboard-read ::event)
 (derive :event/table-read ::event)
+(derive :event/card-read ::event)
```
noahmoss, line 16: Changed `recent_views` to only be triggered by API-level read events, not by card queries. `recent_views` should essentially mirror the same events we're logging in `view_log`, but only contain the most recent N (= 30) views per user so that it's quicker to access for non-audit queries

src/metabase/models/recent_views.clj
```diff
@@ -25,7 +25,7 @@
 (def ^:private ^:dynamic *recent-views-stored-per-user*
   "The number of recently viewed items to keep per user. This should be larger than the number of items returned by the
   /api/activity/recent_views endpoint, but it should still be lightweight to read all of a user's recent views at once."
-  100)
+  30)
```
noahmoss, line 28: Dropping this value to a more reasonable one, just in case this was the cause of any perf issues.

src/metabase/events/view_log.clj
```diff
@@ -3,63 +3,101 @@
...
 
-(derive ::read-event :metabase/event)
-(derive :event/card-read ::read-event)
-(derive :event/dashboard-read ::read-event)
-(derive :event/table-read ::read-event)
+(defn- generate-view
+  "Generates a view, given an event map."
+  [{:keys [object user-id has-access]
+    :or   {has-access true}}]
+  {:model      (u/lower-case-en (audit-log/model-name object))
```
noahmoss, line 22: I'm standardizing model names as being lowercase in `view_log` and `recent_views` even though they're upper-case in `audit_log`. I wish they could be standardized as upper-case for all three tables, but enough code currently relies on them being lowercase for views, that it's not worth breaking anything.

src/metabase/models/recent_views.clj
escherize: Is there any evidence that this is related to a perf issue? I am chesterton's-fencing over it, since I don't know why it was 100 in the first place.

escherize (approved):
LGTM. I only had one question I'd like to see addressed about dropping 100 to 30.

PR #35737, opened 2023-11-15
Improve `uberjar.yml` workflow (test multi-arch job)

nemanjaglumac:
It previously didn't even run multi-arch job if not on `master`. There's no reason not to run this job. The trick is only to **not** push the changes to DockerHub, unless we're on the `master` branch.

The example of a run with this exact config
https://github.com/metabase/metabase/actions/runs/6874177313/job/18695622883

We can clearly see which steps ran, and which have been skipped
image

### DIFF
--- a/.github/workflows/uberjar.yml
+++ b/.github/workflows/uberjar.yml
@@ -194,7 +194,6 @@ jobs:
   containerize_multi_arch:
     runs-on: ubuntu-22.04
     name: Containerize multi-arch ${{ matrix.edition }}
-    if: ${{ github.ref_name == 'master' }}
     needs: check_jar_health
     strategy:
       matrix:
@@ -252,20 +251,23 @@ jobs:
         run: while ! curl -s 'http://localhost:3001/api/health' | grep '{"status":"ok"}'; do sleep 1; done
         timeout-minutes: 3
       - name: Login to Docker Hub
+        if: ${{ github.ref_name == 'master' }}
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       # Push experimental ubuntu image only for versions based on a master
       - name: Install regctl
+        if: ${{ github.ref_name == 'master' }}
         uses: regclient/actions/regctl-installer@main
         with:
           release: 'v0.4.7'
       - name: Switch regctl to point to localhost:5000 via http
+        if: ${{ github.ref_name == 'master' }}
         run: regctl registry set --tls disabled localhost:5000
       - name: Retag and push ubuntu-based images if master (ee)
-        if: ${{ matrix.edition == 'ee' }}
+        if: matrix.edition == 'ee' && github.ref_name == 'master'
         run: regctl image copy localhost:5000/metabase-dev:${{ steps.extract_branch.outputs.branch }}-ee-ubuntu ${{ github.repository_owner }}/metabase-enterprise-head:latest-ubuntu
       - name: Retag and push ubuntu-based images if master (oss)
-        if: ${{ matrix.edition == 'oss' }}
+        if: matrix.edition == 'oss' && github.ref_name == 'master'
         run: regctl image copy localhost:5000/metabase-dev:${{ steps.extract_branch.outputs.branch }}-oss-ubuntu ${{ github.repository_owner }}/metabase-head:latest-ubuntu
--- a/.github/workflows/uberjar.yml
+++ b/.github/workflows/uberjar.yml
@@ -266,8 +266,8 @@ jobs:
         if: ${{ github.ref_name == 'master' }}
         run: regctl registry set --tls disabled localhost:5000
       - name: Retag and push ubuntu-based images if master (ee)
-        if: matrix.edition == 'ee' && github.ref_name == 'master'
+        if: ${{ (matrix.edition == 'ee') && (github.ref_name == 'master') }}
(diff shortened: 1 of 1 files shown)


WiNloSt (commented):
This looks good although I have a few questions/thoughts
1. I don't understand why this workflow would even run in your branch [`add-noto-fonts-test`](https://github.com/metabase/metabase/tree/add-noto-fonts-test), given the workflow specifies [`on.push.branches = master`](https://github.com/metabase/metabase/blob/e5a26b6de60a2a2ace92e59b258b841e251d9283/.github/workflows/uberjar.yml#L6). Can you help explain that?
2. Do you think it's a good idea to separate steps that require additional branch check to another job, so we could just conditionally not running the job instead of having to add conditions to all the steps we don't want to run?

nemanjaglumac:
> 1. I don't understand why this workflow would even run in your branch [`add-noto-fonts-test`](https://github.com/metabase/metabase/tree/add-noto-fonts-test), given the workflow specifies [`on.push.branches = master`](https://github.com/metabase/metabase/blob/e5a26b6de60a2a2ace92e59b258b841e251d9283/.github/workflows/uberjar.yml#L6). Can you help explain that?

Because I triggered it manually, and that's [`workflow_dispatch`](https://github.com/metabase/metabase/blob/e5a26b6de60a2a2ace92e59b258b841e251d9283/.github/workflows/uberjar.yml#L17) event.

nemanjaglumac:
> Do you think it's a good idea to separate steps that require additional branch check to another job, so we could just conditionally not running the job instead of having to add conditions to all the steps we don't want to run?

Can you please give me an example of what you had in mind? Don't want to risk misunderstanding what you asked.

WiNloSt:
@nemanjaglumac 

Just for clarification this change is for running this workflow **manually** (via workflow_dispatch) for all branches (not just master) correct?

> Can you please give me an example of what you had in mind? Don't want to risk misunderstanding what you asked.

I meant something like this. Imagine these steps are moved from `containerize_multi_arch`

```
  # I omitted the steps to get data pushed to local registry from `containerize_multi_arch` as I don't know how to do that
  push_to_docker_hub:
    runs-on: ubuntu-22.04
    name: Containerize multi-arch ${{ matrix.edition }}
    if: ${{ github.ref_name == 'master' }}
    needs: check_jar_health
    strategy:
      matrix:
        edition: [ee, oss]
    services:
      registry:
        image: registry:2
        ports:
          - 5000:5000
    if: ${{ github.ref_name == 'master' }}
    steps:
      - name: Login to Docker Hub
        uses: docker/login-action@v2
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      # Push experimental ubuntu image only for versions based on a master
      - name: Install regctl
        uses: regclient/actions/regctl-installer@main
        with:
          release: 'v0.4.7'
      - name: Switch regctl to point to localhost:5000 via http
        run: regctl registry set --tls disabled localhost:5000
      - name: Retag and push ubuntu-based images if master (ee)
        if: ${{ matrix.edition == 'ee' }}
        run: regctl image copy localhost:5000/metabase-dev:${{ steps.extract_branch.outputs.branch }}-ee-ubuntu ${{ github.repository_owner }}/metabase-enterprise-head:latest-ubuntu
      - name: Retag and push ubuntu-based images if master (oss)
        if: ${{ matrix.edition == 'oss' }}
        run: regctl image copy localhost:5000/metabase-dev:${{ steps.extract_branch.outputs.branch }}-oss-ubuntu ${{ github.repository_owner }}/metabase-head:latest-ubuntu
```

nemanjaglumac:
> Just for clarification this change is for running this workflow manually (via workflow_dispatch) for all branches (not just master) correct?

@WiNloSt 
No. That trigger has always been there. In other words, you could've run this workflow manually at least for the last year or so. This change proposes that instead of completely skipping "multi-arch" job, we should run bits that don't require pushing to DockerHub. That will give us feedback whether or not we broke something.

> I meant something like this. Imagine these steps are moved from containerize_multi_arch

Local data is not preserved between different jobs. Every new job is staring from a pristine environment. I believe this was the reason we couldn't use this approach in the first place. You'd need to set everything up in one job and to then store any output as an artifact, upload that artifact, download it in the next job, unpack it and then use it. And you can't really do that with processes.

.github/workflows/uberjar.yml
```diff
@@ -252,20 +251,23 @@ jobs:
...
+        if: ${{ github.ref_name == 'master' }}
         uses: regclient/actions/regctl-installer@main
         with:
           release: 'v0.4.7'
       - name: Switch regctl to point to localhost:5000 via http
+        if: ${{ github.ref_name == 'master' }}
         run: regctl registry set --tls disabled localhost:5000
       - name: Retag and push ubuntu-based images if master (ee)
-        if: ${{ matrix.edition == 'ee' }}
+        if: matrix.edition == 'ee' && github.ref_name == 'master'
```
WiNloSt, line 269: For consistency should we either always use `${{ }}` or not at all in the `if: ` clause?

[From here](https://docs.github.com/en/actions/learn-github-actions/expressions), it said we could optionally omit `${{ }}` in if clause.

WiNloSt (approved):
I have one minor question. Otherwise, this looks good to me.

.github/workflows/uberjar.yml
nemanjaglumac: @WiNloSt take another look, please
https://github.com/metabase/metabase/pull/35737/commits/095e7a2678c23d0731abcff0414af0c8a267db6e

I prefer to use the syntax without `${{ ]}` but that would require changing the whole workflow so I brought them back for these two lines only.

WiNloSt: Awesome. Thanks for addressing this 🙏

PR #35740, opened 2023-11-15
Improve external filter checks

nemanjaglumac:
This PR unifies "external-filter" checks across multiple workflows that require such check.

It also improves the conditional logic, which never accounted for the `push` scenario!
This led to checks on master being skipped even when pushed within the organization. Here is an example: https://github.com/metabase/metabase/actions/runs/6873940799/job/18694765866
```diff
- github.event.pull_request.head.repo.full_name == github.repository
+ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository 
```

All of this is a result of trying to work around the GitHub limitations, and the lack of first-class support for 3rd party (fork) pull requests in CI.

> **Important**
> The "fix" is relying on the fact that we don't allow direct pushes to any branch. The pull request is always needed. This means that every change that runs on `push` comes from within the organization (from a merged pull request that was previously approved and passed all checks)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -16,7 +16,11 @@ concurrency:
 
 jobs:
   external-filter:
-    if: github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_run.conclusion == 'success'
+    # 1. within the organization OR
+    # 2. fork
+    if: |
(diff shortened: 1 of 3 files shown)


WiNloSt (approved):
Looks good to me

PR #35741, opened 2023-11-15
better logging for escaped cards

### LINKED ISSUE #33882: Log why Serialization v2 fails when trying to export a dashboard that contains cards in outside collections
### Describe the bug

Seems like a dependency issue, where if you use the --collection flag, the process will fail exporting everything when a dashboard has a card outside of that collection

### To Reproduce

1) set up a plain empty instance, create a /target folder so you can export to it
2) click on any x-ray, save the created dashboard
3) then move any card from the created collection to the root collection
4) run "java -jar /app/metabase.jar export /target --collection 3", see that there's no card being exported to the /target folder

### Expected behavior

should we send an error? or just export the cards that are in that collection

### Logs

2023-09-13 09:45:16,673 INFO v2.extract :: Dashboard cards outside the collection
2023-09-13 09:45:16,673 INFO v2.extract :: ======================================
2023-09-13 09:45:16,679 INFO v2.extract :: Dashboard 1: A look at People

2023-09-13 09:45:16,680 INFO v2.extract ::           	Card 3: User growth

2023-09-13 09:45:16,680 INFO v2.extract ::         from collection [no collection]

2023-09-13 09:45:16,683 INFO serialization.cmd :: Export to /target complete! 🚛💨 📦
metabase_origin:/# ls /target/

```JSON
47.x
```

### Severity

P2

### Additional context

We need some help from the product team on this one

sansolovyov:
Resolves #33882

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
@@ -134,29 +134,23 @@
       (format "%d: %s" coll-id names))
     "[no collection]"))
 
+(defn- card-label [card-id]
+  (let [card (t2/select-one [Card :collection_id :name] :id card-id)]
+    (format "Card %d (%s from collection %s)" card-id (:name card) (collection-label (:collection_id card)))))
+
 (defn- escape-report
   "Given the analysis map from [[escape-analysis]], report the results in a human-readable format with Card titles etc."
   [{:keys [escaped-dashcards escaped-questions]}]
   (when-not (empty? escaped-dashcards)
-    (log/info "Dashboard cards outside the collection")
-    (log/info "======================================")
     (doseq [[dash-id card-ids] escaped-dashcards
             :let [dash-name (t2/select-one-fn :name Dashboard :id dash-id)]]
-      (log/infof "Dashboard %d: %s\n" dash-id dash-name)
-      (doseq [card_id card-ids
-              :let [card (t2/select-one [Card :collection_id :name] :id card_id)]]
-        (log/infof "          \tCard %d: %s\n"    card_id (:name card))
-        (log/infof "        from collection %s\n" (collection-label (:collection_id card))))))
+      (log/warnf "Failed to export Dashboard %d (%s) containing Cards saved outside requested collections: %s"
+                 dash-id dash-name (str/join ", " (map card-label card-ids)))))
 
   (when-not (empty? escaped-questions)
-    (log/info "Questions based on outside questions")
-    (log/info "====================================")
-    (doseq [[curated-id alien-id] escaped-questions
-            :let [curated-card (t2/select-one [Card :collection_id :name] :id curated-id)
-                  alien-card   (t2/select-one [Card :collection_id :name] :id alien-id)]]
-      (log/infof "%-4d      %s    (%s)\n  -> %-4d %s    (%s)\n"
-                 curated-id (:name curated-card) (collection-label (:collection_id curated-card))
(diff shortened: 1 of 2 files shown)


enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
```diff
@@ -1566,3 +1566,24 @@
...
+                       Card          {card1-id :id} {:name "Some Card"}
+                       DashboardCard _              {:card_id card1-id :dashboard_id dash-id}
+                       Card          _              {:name "Dependent Card" :collection_id coll2-id
+                                                     :dataset_query {:query {:source-table (str "card__" card1-id)
+                                                                             :aggregation [[:count]]}}}]
+      (testing "Complain about card to available for exporting"
+        (is (= 1
+               (count
+                (mt/with-log-messages-for-level ['metabase-enterprise :warn]
+                  (extract/extract {:targets [["Collection" coll1-id]] :no-settings true :no-data-model true}))))))
```
calherries, line 1584: Checking the count is brittle if more log messages are added in the future. I think this should be tighter and check the actual logged string. We have a lot of tests already that check inside the data structure returned by `mt/with-log-messages-for-level` for the exact error message, so I think it's reasonable to do so. 

e.g.
https://github.com/metabase/metabase/blob/bf771f5fb434c810e63b400839c63b3f57e5d506/test/metabase/server/middleware/session_test.clj#L88-L93

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -134,29 +134,23 @@
...
+(defn- card-label [card-id]
+  (let [card (t2/select-one [Card :collection_id :name] :id card-id)]
+    (format "Card %d (%s from collection %s)" card-id (:name card) (collection-label (:collection_id card)))))
+
 (defn- escape-report
   "Given the analysis map from [[escape-analysis]], report the results in a human-readable format with Card titles etc."
   [{:keys [escaped-dashcards escaped-questions]}]
   (when-not (empty? escaped-dashcards)
-    (log/info "Dashboard cards outside the collection")
-    (log/info "======================================")
```
calherries, line 142: Why did you remove all the info logging? That wasn't in the spec

calherries (changes requested):

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
piranha: Well, the ticket says "improve logs", not "add", and the information is the same, so feels like unnecessary?..

enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
calherries: Fair, yet depending on the message format is preferable to having this break when new logs are added. You have to depend on something

piranha: How does the new version look? :)

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
calherries: True

calherries (approved):

enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
calherries: Very nice thanks

PR #35748, opened 2023-11-15
Skip excess keys on serialization load

johnswanson:
When loading from a backup, we should not fail if the previous version saved keys that are no longer needed in the current version. This adds a step (after ingestion but before loading) that removes any "extra" keys from the ingested entity.

Co-authored by: escherize <343288+escherize@users.noreply.github.com>

### DIFF
--- a/src/metabase/models/serialization.clj
+++ b/src/metabase/models/serialization.clj
@@ -16,6 +16,7 @@
    [clojure.set :as set]
    [clojure.string :as str]
    [medley.core :as m]
+   [metabase.db.connection :as mdb.connection]
    [metabase.lib.schema.id :as lib.schema.id]
    [metabase.mbql.normalize :as mbql.normalize]
    [metabase.mbql.util :as mbql.u]
@@ -510,7 +511,7 @@
 
 (def ^:private fields-for-table
   "Given a table name, returns a map of column_name -> column_type"
-  (memoize
+  (mdb.connection/memoize-for-application-db
    (fn fields-for-table [table-name]
      (t2/with-connection [conn]
        (u.conn/app-db-column-types conn table-name)))))
--- a/src/metabase/models/serialization.clj
+++ b/src/metabase/models/serialization.clj
@@ -22,6 +22,7 @@
    [metabase.models.interface :as mi]
    [metabase.shared.models.visualization-settings :as mb.viz]
    [metabase.util :as u]
+   [metabase.util.connection :as u.conn]
    [metabase.util.i18n :refer [trs]]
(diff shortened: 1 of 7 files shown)


darksciencebase:
cc @piranha

src/metabase/models/serialization.clj
```diff
@@ -507,15 +509,52 @@
...
   "Performs the usual steps for an incoming entity:
   - Drop :serdes/meta
 
-  You should call this as a first step from any implementation of [[load-xform]].
+  You should call this as part of any implementation of [[load-xform]].
 
   This is a mirror (but not precise inverse) of [[extract-one-basics]]."
   [ingested]
-  (dissoc ingested :serdes/meta))
+  (-> ingested drop-excess-keys (dissoc :serdes/meta)))
```
piranha, line 557: Feels like `:serdes/meta` will be removed by `drop-excess-keys` anyways?

src/metabase/util/connection.clj
```diff
@@ -0,0 +1,21 @@
...
+  [app-db table-name']
+  (let [table-name (cond-> table-name'
+                     (= (:db-type app-db) :h2) u/upper-case-en)]
+    (t2/with-connection [^Connection conn]
+      (with-open [rset (.getColumns (.getMetaData conn) nil nil table-name nil)]
+        (into {} (take-while some?)
+              (repeatedly
+               (fn []
+                 (when (.next rset)
+                   [(.getString rset "COLUMN_NAME") (.getString rset "TYPE_NAME")]))))))))
```
piranha, line 21: IDK if you're into that (I'm ok with code in place here), but since 1.11 there is `clojure.core/iteration`:

```
(into {} (iteration (fn [_] (when (.next cs)
                                  [(.getString cs "COLUMN_NAME") (.getString cs "TYPE_NAME")]))))
```

piranha: Ah, you've just moved the code :)

piranha:
The code is fine, but I'm not sold on having a big e2e test just to check this functionality. Maybe some smaller unit-test in somewhere in `v2.load-test` will suffice, what do you think?

src/metabase/models/serialization.clj
johnswanson: Good point, thank you!

src/metabase/util/connection.clj
johnswanson: Ah yeah, I did just move it, but also TIL about `iteration` - very nice! 🎉

sansolovyov (approved):
Cool!

PR #35756, opened 2023-11-15
[E2E] Fix dashboard card "default sizes" flake

nemanjaglumac:
This flake (and the flake fix) is somewhat related to #35751 in that we're using the same set of 17 questions in a dashboard.

### Stats
image

Examples of failed runs:
- https://www.deploysentinel.com/ci/runs/655284a0c9fa8b50a4c496f4
- https://www.deploysentinel.com/ci/runs/6552b09e2beaab0325ce6abc
- https://www.deploysentinel.com/ci/runs/65523ecb7bcc148681c75b66

### The cause
Cypress is moving very faster through UI, and it doesn't wait for all requests/responses to finish before trying to save the dashboard. So that save fails. _I guess one could reproduce this even locally if one clicks too fast and adds too many questions to the dashboard._
image

Each time we add a card to the dashboard, there is a pair of requests that we need to take into account:
1. `GET /api/card/:id` 
2. `POST /api/card/:id/query` (or `POST /api/card/pivot/:id/query`)

### The solution
Wait for the card query to load before we attempt to add the next question/card. This ensures everything's in order before we attempt to save the changes to the dashboard.

Screenshot 2023-11-15 at 16 52 13

Compare this with the chaotic order oof requests from the failed run

image

### Stress-test
- https://github.com/metabase/metabase/actions/runs/6879649254
- https://github.com/metabase/metabase/actions/runs/6879645906
- https://github.com/metabase/metabase/actions/runs/6879642268

### DIFF
--- a/e2e/test/scenarios/dashboard-cards/dashboard-card-resizing.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/dashboard-card-resizing.cy.spec.js
@@ -217,8 +217,27 @@ describe("scenarios > dashboard card resizing", () => {
           cy.findByLabelText("Add questions").click();
         });
 
-        TEST_QUESTIONS.forEach(question => {
-          cy.findByLabelText(question.name).click();
+        /**
+         * Metabase sorts all questions in the sidebar alphabetically.
+         * It makes sense to sort them out here as well in order to avoid
+         * Cypress "jumping" up and down while clicking on them.
+         * It will go in order from top to the bottom, which scrolls the
+         * sidebar naturally. This prevents acting on an element that's not visible.
+         */
+        const sortedCards = TEST_QUESTIONS.sort((a, b) =>
+          a.name.localeCompare(b.name),
+        );
+
+        /**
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/dashboard-cards/dashboard-card-resizing.cy.spec.js
```diff
@@ -217,8 +217,27 @@ describe("scenarios > dashboard card resizing", () => {
...
+          a.name.localeCompare(b.name),
+        );
+
+        /**
+         * After each card is added to the dashboard from the sidebar, there is a
+         * request to load the card query. We need to wait for each of those before
+         * we attempt to add a new card. Otherwise the save dashboard might fail
+         * because Cypress is too fast.
+         */
+        cy.intercept("POST", "/api/card/**/query").as("cardQuery");
```
nemanjaglumac, line 237: This glob matches both `POST /api/card/:id/query` and `POST /api/card/pivot/:id/query`

iethree (approved):

PR #35784, opened 2023-11-15
type createThunkAction to avoid it leaking anys in all the codebase

npretto:
### Description

`createThunkAction` is used in typescript files, and we often type the functions that we pass to it.
Unfortunately the returned action/thunk creator is typed as any and therefore type checking on the arguments doesn't work:
image

This pr addresses that:

image

It should also automatically type dispatch and getState, which is a nice 
image


npretto:
Also, should we deprecated it in favor of `createAsyncThunk`?

WiNloSt (approved):
You made a good point about `createAsyncThunk`. We could discuss this in FE guild sync. But I'm +1 for deprecating `createThunkAction` for `createAsyncThunk`

PR #35835, opened 2023-11-15
Fix a case where CLJS code was not preserving Moment.js field names

bshepherdson:
This led to errors only in compiled mode. A `.endOf()` call was being
improperly renamed by the optimizer to eg. `.rj()`, and failing.

This change makes the `^moment/Moment` type annotations more explicit,
so they aren't missed by the compiler.


camsaul (approved):

ranquild:
Verified the fix locally

PR #35837, opened 2023-11-16
Underlying-records drill should generate correct filters for binned columns

### LINKED ISSUE #35341: [MLv2] [BE] applying `underlying-records` produces wrong filter for binned data
When used on data with binning, applying underlying-records should produce filter with values between boundaries.
Currently it produces strict equality with the first value.

Please use this e2e-test to reproduce -
https://github.com/metabase/metabase/blob/7ec1e0adcdbddeea970ed161680a6f618cd9f611/e2e/test/scenarios/visualizations/drillthroughs/chart_drill.cy.spec.js#L505-L535

camsaul:
Fixes #35341 

There were a few issues here

- QP results metadata comes back with `:binning_info` for binned columns, while `lib/binning` looks for `:metabase.lib.field/binning`. This PR updates the JS metadata provider to massage things into the expected MLv2 shape
- The underlying records drill did not check whether a column was binned and always generated a single `:=` filter; for binning it needs to calculate the bin's actual min max values (using the stuff recently backported from the QP code to power the `zoom-in-bins` drill) and generate a `:>=` and `:<` filter (`:between` is not appropriate here because `:between` is inclusive on both sides, like SQL `BETWEEN`. Since I already implemented this code for zoom-in bins recently I moved it in to `lib.binning` as a shared util function so it can be used here as well

### DIFF
--- a/src/metabase/lib/drill_thru.cljc
+++ b/src/metabase/lib/drill_thru.cljc
@@ -20,6 +20,8 @@
    [metabase.lib.metadata.calculation :as lib.metadata.calculation]
    [metabase.lib.schema :as lib.schema]
    [metabase.lib.schema.drill-thru :as lib.schema.drill-thru]
+   [metabase.util :as u]
+   [metabase.util.log :as log]
    [metabase.util.malli :as mu]))
 
 (comment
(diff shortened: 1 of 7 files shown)


src/metabase/lib/drill_thru.cljc
```diff
@@ -109,4 +111,6 @@
     stage-number :- :int
     drill        :- ::lib.schema.drill-thru/drill-thru
     & args]
+   (log/infof "Applying drill thru: %s"
```
snoe, line 114: This looks like debug level?

snoe (approved):

src/metabase/lib/drill_thru.cljc
camsaul: Agree, but debug level doesn't pop up in the browser console and the main reason I want this is so I can debug it in the browser/Cypress tests. Is there some way to configure the log level from the browser console?

PR #35844, opened 2023-11-16
Add an E2E test for Filter dropdown does not work in embedded applications when params is not an object

### LINKED ISSUE #14474: Filter dropdown does not work in embedded applications when `params` is not an object
As pointed by @beko1997 this issue still exist in embedded dashboards with token in version 0.37.7 but the issue #13639 is closed, I open this new one.

> @mazameli @camsaul @SG2019 I can indeed confirm this issue still exists with embeding dashboards. I've tested both v0.37.3 and v0.37.5. The filterlist stays empty on the embedded version, while it does get filled in MetaBase self.
> 
> The issue is a 400 response on the following request:
> 
> GET api/embed/dashboard/xxxxxxx/params/b2234673/values
> ````
> Output of chain-filter-merged-params does not match schema: 
> 
> 	 [0;33m  {(not (instance? java.lang.String nil)) invalid-key} [0m 
> ````
> 
> Any ideas on how to fix this one?
> 
> _Originally posted by @beko1997 in https://github.com/metabase/metabase/issues/13639#issuecomment-760060029_

WiNloSt:
Closes #14474

#14474 is already solved i.e. when `params` isn't an object that isn't supported by design, we already show that in [server error log](https://github.com/metabase/metabase/issues/14474#issuecomment-1812555466). And now we're showing an error instead of showing results with broken filter autocompletion.

This PR adds a test to verify that when `params` isn't an object we show an error message instead.

### How to verify

Follow the reproduce steps from [this comment](https://github.com/metabase/metabase/issues/14474#issuecomment-764724524).

You can use [this Next JS app](https://github.com/WiNloSt/metabase-embeddings-101/blob/main/app/page.tsx) to test which makes it easy to test static embeddings, or use any setup of your choice.

### Demo

image

### Checklist

### DIFF
--- a/e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
+++ b/e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
@@ -5,6 +5,7 @@ import {
   visitEmbeddedPage,
   filterWidget,
   visitIframe,
+  getDashboardCard,
 } from "e2e/support/helpers";
 
 import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
@@ -220,6 +221,32 @@ describe("scenarios > embedding > dashboard parameters", () => {
       );
       // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
       cy.contains(".ScalarValue", "2");
+    });
+  });
+
+  it("should render error message when `params` is not an object (metabase#14474)", () => {
+    cy.get("@dashboardId").then(dashboardId => {
+      cy.request("PUT", `/api/dashboard/${dashboardId}`, {
+        embedding_params: {
+          id: "enabled",
+          name: "enabled",
+          source: "enabled",
+          user_id: "enabled",
+        },
+        enable_embedding: true,
+      });
+
+      const invalidParamsValue = [];
+      const payload = {
+        resource: { dashboard: dashboardId },
+        params: invalidParamsValue,
+      };
+
+      visitEmbeddedPage(payload);
+
+      getDashboardCard()
+        .findByText("There was a problem displaying this chart.")
+        .should("be.visible");
     });
   });
 });


e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
```diff
@@ -222,6 +223,32 @@ describe("scenarios > embedding > dashboard parameters", () => {
...
+          source: "enabled",
+          user_id: "enabled",
+        },
+        enable_embedding: true,
+      });
+
+      const invalidParamsValue = [];
+      const payload = {
+        resource: { dashboard: dashboardId },
+        params: invalidParamsValue,
```
WiNloSt, line 242: If we copy the code from Metabase's code panel, this won't be wrong
image

but since users can just change this to anything, or implement their own for languages that we don't have an example for it's possible to pass the wrong value type.

e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
```diff
@@ -222,6 +223,32 @@ describe("scenarios > embedding > dashboard parameters", () => {
       cy.contains(".ScalarValue", "2");
     });
   });
+
+  it("should render error message when `params` is not an object (metabase#14474)", () => {
+    cy.get("@dashboardId").then(dashboardId => {
+      cy.request("PUT", `/api/dashboard/${dashboardId}`, {
```
npretto, line 229: This call is needed to enable embedding + enable filters on those params right?
It looks like it's something we'll do often, should we make a helper function?

WiNloSt: Maybe. I haven't touched too many embedding tests. So, I'd rather wait and see if this is a pattern we do often. I found we can also do this when we first create a dashboard, so maybe we don't need this extra step at all.

https://github.com/metabase/metabase/blob/5d91e6b7dad73be1a3c8efc3cbbed0c0bafe8bae/e2e/test/scenarios/dashboard-filters/reproductions/29347-remapped-field-values.cy.spec.js#L30-L36

npretto (approved):

e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
WiNloSt: This is because some tests don't need the dashboard to have the embedding published. I think we can leave this here as is.

If we have control over the dashboard creation then we would be able to set the embedding to be published since the dashboard creation. Another way is to create another dashboard with embedding published, but that also requires a new API call anyway. So in the end I think reusing the created dashboard is more appropriate.

nemanjaglumac:
Stress tested and it passes 20/20 times 🎉 
https://github.com/metabase/metabase/actions/runs/6931513823/job/18853396208#step:9:157

@WiNloSt please include stress-test runs the next time you're touching or creating a new E2E test.

Issue #35379, opened 2023-11-03
Recently viewed results are cached but should be fetched every time


alxnddr:
### Describe the bug

Recently viewed results were fetched every time users clicked on the search bar which makes sense because tabs can live long time. It got changed recently to caching the first response and always showing it.

- [ ] Update the related [e2e spec](https://github.com/metabase/metabase/pull/35848#pullrequestreview-1735740957) so it verifies we request recenly viewed items on every click and not cache them.

### To Reproduce

1. Click on the search bar to see "Recently viewed"
2. In the same tab open a few questions
3. Click on the search bar to see updated "Recently viewed"
4. It shows the previous "Recently viewed" items

### Expected behavior

It should show updated "Recently viewed" items

### Information about your Metabase installation

```JSON
Latest master
```

### Severity

Not blocking but worsens user experience

PR #35848, opened 2023-11-16
[E2E] Fix recent-views flake(s)

nemanjaglumac:
### Stats (for the last 7 days on `master`)
| type | %|
|-|-|
|Failure Rate|2.67%|
|Flake Rate|14.67%|

Examples of failed runs:
- https://www.deploysentinel.com/ci/runs/654e06607bcc1486818ec8c7
- https://www.deploysentinel.com/ci/runs/655326527bcc148681eac24a
- https://www.deploysentinel.com/ci/runs/655326707bcc148681eac795

image

### The cause
Intercept and wait are placed in a wrong place. The request to get `/api/activity/recent_views` happens upon a home page visit, not when we click on the search. You can check this by doing a hard refresh on the home page. Observe the network tab. Then click on "Search..." - nothing happens.

### The solution
Wait for the response from this request upon the home page load.

### Stress-tests
3 x 20 ✅ 
- https://github.com/metabase/metabase/actions/runs/6887773655
- https://github.com/metabase/metabase/actions/runs/6887955037
- https://github.com/metabase/metabase/actions/runs/6888000135

### Additional Notes
This also seems to have fixed the other related flake (they are sharing the same `describe` block where the failure occurred). 
`"allows to select an item from keyboard`

Stress tests for it also reveal 3 x 20 ✅ 
- https://github.com/metabase/metabase/actions/runs/6888275796
- https://github.com/metabase/metabase/actions/runs/6888258557
- https://github.com/metabase/metabase/actions/runs/6888164373

### DIFF
--- a/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
+++ b/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
@@ -39,7 +39,7 @@ describe("search > recently viewed", () => {
     cy.findByTestId("loading-spinner").should("not.exist");
   });
 
-  it("shows list of recently viewed items", { tags: "@flaky" }, () => {
+  it("shows list of recently viewed items", () => {
     assertRecentlyViewedItem(0, "Orders in a dashboard", "Dashboard");
     assertRecentlyViewedItem(1, "Orders", "Question");
     assertRecentlyViewedItem(2, "People", "Table");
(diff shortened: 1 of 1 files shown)


alxnddr (approved):
I believe this flake is caused by introducing a bug in Recents: https://github.com/metabase/metabase/issues/35379

It totally makes sense to fix the flake now but we need to change it after the fix and update the spec so it expects two requests for recent items: one for the home page section and another one for the list under the search bar after click

nemanjaglumac:
> I believe this flake is caused by introducing a bug in Recents: #35379
> 
> It totally makes sense to fix the flake now but we need to change it after the fix and update the spec so it expects two requests for recent items: one for the home page section and another one for the list under the search bar after click

Oooh, so it's a legit failure.
Ok, I'll merge this to prevent potential failures but I'll keep an eye on the issue you mentioned and the potential fix for it.

PR #35851, opened 2023-11-16
Fix embedding crashes when one of the cards fails to load

### LINKED ISSUE #34954: Static/Public embeds fail to load dashboard when one of the cards fails to load due to missing filter value.
### Describe the bug

Static/Public embeds fail to load dashboard when one of the cards fails to load due to missing filter value.

### To Reproduce

1. Go to New -> SQL Query -> `Select * from products Where category = {{something}}` -> Save

image

2. Add it to a Dashboard -> Add a filter to the  card:

image

3. Go to Edit -> Click on the Card and edit visualisation option -> Enable `Hide this card if there are no results`

image

4. Go to Share -> Enable Public Sharing (this also happens for Static Embedding) -> Go to the public URL and notice a blank screen:

image

The dashboard is able to load at first but if you notice the network calls when the error due to filter not being selected for the card occurs it simply breaks the screen:

image

Something similar to Metabase should occur like the following:

image

Even though the error message `something is gone wrong` is very misleading instead of a `please select filter for something` 

### Expected behavior

Cannot share an image here but i described it in the Reproduce Section

### Logs

None that are relevant even though the actual error is:

`Cannot run the query: missing required parameters: #{\"something\"}`

Even though it's not specific to that error but how the frontend behaves due to that error

```JSON
1.47.5 ann master
```

### Severity

This completely blocks the usage of 2 important features

### Additional context

The only workaround is to put a default value but if you decide to remove it from the URL then you end up breaking the visual again

WiNloSt:
Closes https://github.com/metabase/metabase/issues/34954

### Description
Apparently, when we query dashboard cards (should apply to questions as well, but I didn't test) on embeddings, if that results in an error. The response won't contain `data` key like when we query dashboard cards from the app.

It's preferable to fix this on FE, so this PR did just that. I initially tried to set `Dataset = AppDataset | EmbedDataset` but that resulted in type errors in more than 30 places, so I needed to drop the idea. This fix only targets this particular instance and specifically says the parameter could either be `Dataset` or `EmbedDataset`.

### How to verify
Follow the repro steps in https://github.com/metabase/metabase/issues/34954

You can use [this Next JS app](https://github.com/WiNloSt/metabase-embeddings-101/blob/main/app/page.tsx) to test which makes it easy to test static embeddings, or use any setup of your choice.

#### Before
Crashes with white screen

#### After
image

### Checklist

### DIFF
--- a/frontend/src/metabase-types/api/mocks/dataset.ts
+++ b/frontend/src/metabase-types/api/mocks/dataset.ts
@@ -2,8 +2,9 @@ import type {
   Dataset,
   DatasetColumn,
   DatasetData,
-  PublicDataset,
-  PublicDatasetData,
+  EmbedDataset,
+  EmbedDatasetData,
+  ErrorEmbedDataset,
   ResultsMetadata,
   TemplateTag,
 } from "metabase-types/api/dataset";
@@ -57,7 +58,7 @@ export const createMockDataset = ({
   ...opts,
 });
 
-export const createMockPublicDatasetData = ({
+export const createMockEmbedDatasetData = ({
   cols = [
     createMockColumn({
       display_name: "NAME",
@@ -66,7 +67,7 @@ export const createMockPublicDatasetData = ({
     }),
   ],
   ...opts
-}: Partial<PublicDatasetData>): PublicDatasetData => ({
+}: Partial<EmbedDatasetData>): EmbedDatasetData => ({
   rows: [],
   cols,
   rows_truncated: 0,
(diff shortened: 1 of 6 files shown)


frontend/src/metabase-types/api/mocks/dataset.ts
```diff
@@ -66,24 +67,37 @@ export const createMockPublicDatasetData = ({
...
-  ...opts,
-});
+export type MockEmbedDatasetOpts = Omit<Partial<EmbedDataset>, "data"> & {
+  data?: Partial<EmbedDatasetData>;
+};
+
+export const createMockEmbedDataset = (
+  opts: MockEmbedDatasetOpts,
+): EmbedDataset => {
+  if ("data" in opts) {
```
WiNloSt, line 85: This change is due to the fact that `EmbedDataset` can either be `SuccessEmbedDataset` or `ErrorEmbedDataset`, and they have different shapes.

frontend/src/metabase/public/containers/PublicQuestion/PublicQuestion.unit.spec.tsx
```diff
@@ -9,7 +9,7 @@ import registerVisualizations from "metabase/visualizations/register";
 import { createMockState } from "metabase-types/store/mocks";
 import {
   createMockPublicCard,
-  createMockPublicDataset,
+  createMockEmbedDataset,
```
WiNloSt, line 12: just a rename

deniskaber (approved):
Works as expected!

WiNloSt:
Stress testing the E2E file
- This PR branch: https://github.com/metabase/metabase/actions/runs/6939453064/job/18876829681

PR #35860, opened 2023-11-16
fix tests failing because we are -45 days to the end of the year

npretto:
### Description

This fixes the tests that are failing on master

The tests are using `.humanize()` but the component is using `.fromNow`.

It seems that the latter uses "45" days to round the date to "1 year ago" and today is exactly 45 days to the end of the year 🤷


uladzimirdev (approved):

npretto:
nope wrong pr

PR #35861, opened 2023-11-16
Remove fromUrl parameter from QueryBuilder

### LINKED ISSUE #35785: Saving a SQL question will change the location to the filter value
### Describe the bug

Check question 14403 on stats and change the comment on the top

save it

see that it changes the location of the window to the filter value

### To Reproduce

See above

### Expected behavior

we shouldn't change the location?

### Logs

NA

```JSON
master
```

### Severity

P1

### Additional context

_No response_

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/35785

### Description

There was an old parameter "fromUrl" in query builder which allowed to redirect to some previous page after question editing. But it looks like that functionality wasn't used. At the same time blacklisting of `from` keyword in the URL can conflict with named parameters. 

So, solution is to remove that old functionality seems nobody used.

### How to verify

Check e2e test:
- create sql question with  2 params, first is any, second should have name "from"
- save a question
- change anything in the question (e.g. add a comment on the top), save and "replace" the question
- there shouldn't be any URL redirect from the question

### Demo

example of the broken test (without changes in QB)
https://github.com/metabase/metabase/assets/125459446/d9f8023f-fb68-41ab-bf30-f003b87a3e04

### Checklist

### DIFF
--- a/e2e/test/scenarios/native/reproductions/35785-unexpected-redirect-after-question-save.cy.spec.js
+++ b/e2e/test/scenarios/native/reproductions/35785-unexpected-redirect-after-question-save.cy.spec.js
@@ -1,0 +1,54 @@
+import { restore, modal } from "e2e/support/helpers";
+
+const nativeQuery =
+  "select * from products where created_at < {{max_date}} and created_at > {{from}} limit 5";
+
+const questionDetails = {
+  native: {
+    query: nativeQuery,
+    "template-tags": {
+      max_date: {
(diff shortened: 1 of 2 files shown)


ranquild (approved):

uladzimirdev:
It depends on another huge not migrated PR which also touched QueryBuilder, so seems it's better not to backport this PR

uladzimirdev:
^^ I didn't request it, probably some glitch

Issue #17871, opened 2021-09-14
Refactor the Pulse/Alert/DashboardSubscription tables & model code


camsaul:
# Background

Originally we only had normal Pulses, with a model diagram that looked something like this:

![image](https://user-images.githubusercontent.com/1455846/133183045-345d2958-5454-4ab9-b807-75b8c320123c.png)

A Pulse can have zero or more Cards, and a Card can be in zero or more Pulses; this relationship is represented by the PulseCard table. A Pulse has one or more PulseChannels it 

Pretty reasonable. Later on, we added Alerts, which were a special subtype of Pulse. The models then looked something like this:

![image](https://user-images.githubusercontent.com/1455846/133183104-d64847f7-6213-4b87-87ed-a13c6fa9313f.png)

(note that Alert and "NormalPulse" are not separate tables in the application DB; they're subtypes of Pulse, and the `pulse` table has all of the listed columns.)

The main differences between a "NormalPulse" and an "Alert" are that 

1.  Alerts have exactly 1 card whereas NormalPulses can have any number of Cards
2. Alerts have several special columns to decide whether or not the Alert should be sent, and if the Alert should be deleted after the first time it is triggered.

At this point, things were still pretty reasonable, altho in hindsight it may have been better to make Alert a separate table than to put everything in the Pulse table. (Or maybe even better would have been to use table inheritance, if MySQL supported it.)

Things got a little more complicated when we added Collections -- both a Pulse and its Cards belong to Collections, altho not neccesarily the same ones. Still, manageable and not too hard to work with.

###  Current State of the World

I think things got out of control when we added DashboardSubscriptions, which are another subtype of Pulse tied to Dashboards. Dashboards themselves are have zero or more Cards, represented by DashboardCard. For DashboardSubscriptions we create a Pulse and associated PulseCards that have to be programatically kept in sync with the corresponding Dashboard and DashboardCards. The model diagram now looks something like:

![image](https://user-images.githubusercontent.com/1455846/133184262-9e6d5ae0-a8b6-4567-8b15-28cf3d183f0c.png)

(actually this is a little simpler than reality since it excludes "NormalPulses" which are deprecated and scheduled for removal, although not yet removed).

### Major Issues 

The biggest problem is fact that so many `collection_ids` are floating around,

[...]

PR #35874, opened 2023-11-16
check every model has serialization setup

### LINKED ISSUE #35540: Add more tests that v2 serdes is comprehensive, so new models are not forgotten
**Context**

A goal for serdes v2 is to make it impossible to forget to add serdes support for new models.
Many serdes v2 multimethods have `:default` implementations that work for some models, but some do not.

**To do**

For each of these required methods, we should have a test that checks every Toucan model:
- either implements the method; or
- is explicitly listed as not supporting serdes at all

There used to be a list of such excluded models and some tests based on it, but I think it may have been removed in the overzealous "refactoring" changes that removed some important functionality and tests.

I think the right path is to define every subset of models we care about in `metabase-enterprise.serialization.v2.models`: top-level serialized models, embedded models, models that are excluded from serialization, models with `entity_id` columns, etc. and then the tests will all be on the same page about which models need which things.

An example of such a required multimethod is `hash-fields`. Eg. things with `entity_id` columns might not define `hash-fields`, and since the on-insert logic for `entity_id` is filling them in for new things, it's never called. But then if you `drop_entity_ids` because of serdes issues (hopefully a rarity) it will call `hash-fields` and throw for want of a `defmethod` for that model. (This has already happened [in the wild](https://github.com/metabase/metabase/issues/35097).)

sansolovyov:
resolves #35540


enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
```diff
@@ -33,3 +36,45 @@
...
+   "AuditLog"
+   "BookmarkOrdering"
+   "CardBookmark"
+   "CollectionBookmark"
+   "CollectionPermissionGraphRevision"
+   "ConnectionImpersonation"
+   "DashboardBookmark"
+   "DashboardCardSeries"
+   "Database"
+   "Field"
```
bshepherdson, line 53: These should not be listed here - they're `data-model`s.

enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
```diff
@@ -15,7 +15,10 @@
    "Collection"
    "Dashboard"
    "NativeQuerySnippet"
-   "Timeline"])
+   "Timeline"
+   "Pulse"
+   "PulseChannel"
+   "PulseCard"])
```
bshepherdson, line 21: I think this is incorrect. Some of these lists might be doing double duty, and it's causing friction. These lists desperately need comments explaining exactly what goes into them.

`PulseChannel` and `PulseCard` are inlined on pulses, if memory serves. I'm not 100% sure if pulses are really top level `content` models, because pulses the feature is deprecated but we still use the table for dashboard subscriptions. so maybe they're inlined on dashboards?

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +738,25 @@
...
+    (let [should-have-entity-id (set (concat serdes.models/data-model serdes.models/content))]
+      (doseq [model (v2.entity-ids/toucan-models)]
+        (if (contains? should-have-entity-id (name model))
+          (testing (str "Model either has entity_id or a hash key: " (name model))
+            ;; `not=` is effectively `xor`
+            (is (not= (not= (get-method serdes/entity-id (name model))
+                            (get-method serdes/entity-id :default))
+                      (and (not= (get-method serdes/hash-fields model)
+                                 (get-method serdes/hash-fields :default))
+                           (serdes.backfill/has-entity-id? model)))))
```
bshepherdson, line 759: I already know what condition this is supposed to express, and I can't follow this code.
```suggestion
            (let [custom-entity-id?   (not= (get-method serdes/entity-id (name model))
                                            (get-method serdes/entity-id :default))
                  custom-hash-fields? (not= (get-method serdes/hash-fields model)
                                            (get-method serdes/hash-fields :default))
                  random-entity-id?   (and custom-hash-fields?
                                           (serdes.backfill/hash-entity-id? model))]
              ;; `not=` is effectively `xor` - exactly one of these must be true.
               (is (not= custom-entity-id? random-entity-id?))
```

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +738,25 @@
...
+          (testing (str "Model shouldn't have entity_id defined: " (name model))
+            (is (= (get-method serdes/entity-id (name model))
```
bshepherdson, line 761: No `hash-fields` or `has-entity-id?` either.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +738,25 @@
                     (is (thrown-with-msg? Exception #"Please upgrade"
                                           (cmd/v2-load! dump-dir {}))
                         "throws")))))))))))
+
+(deftest every-model-is-supported-test
```
bshepherdson, line 742: I think this belongs in `models_test` - there's nothing "e2e" about this test.

bshepherdson (changes requested):

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
piranha: I've dropped check for them since there are few defining them: DashboardTab, DashboardCard, Dimension. Should I strip them from hash-fields/:hook/entity-id?

Also PulseChannel and PulseCard if I put them into `inline-models` (will see what's going on with them for real tomorrow).

enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
piranha: Took a look at exported `Pulse.*`, and it looks weird. `Pulse` is top-level (sits in `collections/.../pulses/`), and it does not include either PulseCard and PulseChannel. 

Also, if you're exporting just a single collection, the one with dashboard with notifications configured, Pulse is not going to be exported. Seems like an omission we'd like to fix?

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
piranha: Plus DashboardTab and DashboardCard [depend on entity id](https://github.com/metabase/metabase/blob/master/src/metabase/models/dashboard.clj#L623) when loading, but I'm almost sure we can drop all of them and just load new ones, right?

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
piranha: 🤦 right

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
piranha: done

enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
piranha: eh and I thought I'd cleaned the list :\ fixed

enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
bshepherdson: We should at least look into it. You can't create a proper pulse anymore, but we should make sure dashboard subs are getting serialized - assuming we actually want that. We actually might not... check with @luizarakaki . If they're excluded on purposes we should say why in a comment.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
bshepherdson: We prefer to keep the IDs of existing things where practical. So we should be fetching the current set of dashcards, building the dead/updated/new diff sets, and then `DELETE`ing, `INSERT`ing and UPDATE`ing them as necessary.

bshepherdson (approved):

enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
piranha: @bshepherdson okay so we're not exporting Pulse at all. Do you think it makes sense to remove serdes methods and entity_id from Pulse* in that PR?

calherries: Be careful, you can still create rows in the `Pulse` table, they're either dashboard subscriptions or alerts. These should still be serialized.

calherries: Cam drew some nice diagrams here explaining how this works.
https://github.com/metabase/metabase/issues/17871

Ngoc has just deleted the "normal pulses", i.e. legacy pulses that aren't dashboard subscriptions or alerts.

piranha: It seems like [they don't need to](https://metaboat.slack.com/archives/C01LQQ2UW03/p1700579475970899).

calherries: Ah, brilliant

bshepherdson: 👍

bshepherdson (approved):

PR #35878, opened 2023-11-16
[MLv2] [BE] - return query, stage index and column for `column-filter` drill

bshepherdson:
If this drill is targeting an aggregation column, the query needs an
    extra stage added. To enable the custom filtering UI, this updated
    query, stage number, and the correct column *in that new stage* are
    returned as part of the opaque `column-filter` drill.
    
`lib.js/column-filter-drill-details` is added to hand those details
    to the FE as a JS object containing opaque CLJS values.

### DIFF
--- a/src/metabase/lib/drill_thru/column_filter.cljc
+++ b/src/metabase/lib/drill_thru/column_filter.cljc
@@ -1,16 +1,24 @@
 (ns metabase.lib.drill-thru.column-filter
   (:require
+   [medley.core :as m]
    [metabase.lib.drill-thru.common :as lib.drill-thru.common]
+   [metabase.lib.equality :as lib.equality]
    [metabase.lib.filter :as lib.filter]
    [metabase.lib.filter.operator :as lib.filter.operator]
+   [metabase.lib.ref :as lib.ref]
    [metabase.lib.schema :as lib.schema]
    [metabase.lib.schema.drill-thru :as lib.schema.drill-thru]
+   [metabase.lib.stage :as lib.stage]
    [metabase.lib.types.isa :as lib.types.isa]
+   [metabase.lib.util :as lib.util]
    [metabase.util.malli :as mu]))
 
 (mu/defn column-filter-drill :- [:maybe ::lib.schema.drill-thru/drill-thru.column-filter]
   "Filtering at the column level, based on its type. Displays a submenu of eg. \"Today\", \"This Week\", etc. for date
-  columns."
+  columns.
+
+  Note that if the clicked column is an aggregation, filtering by it will require a new stage. Therefore this drill
+  returns a possibly-updated `:query` and `:stage-number` along with a `:column` referencing that later stage."
   [query                  :- ::lib.schema/query
    stage-number           :- :int
    {:keys [column value]} :- ::lib.schema.drill-thru/context]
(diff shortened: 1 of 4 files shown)


bshepherdson:
* master
  * **PR #35878** <a href="https://app.graphite.dev/github/pr/metabase/metabase/35878?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/35878?utm_source=stack-comment).

ranquild (approved):

camsaul (approved):
No context as to what this is for but I trust your judgement. Next time a PR description would be nice tho 💪

bshepherdson:
@camsaul good call. The code ended up getting completely replaced, so PTAL. Now the commit message and PR title are clearer what is going on, and I've added some context to the description.

ranquild (approved):

ranquild (changes requested):
aggregation columns seem to be handled correctly, but I also need this column to be `filterable`, i.e. coming from `filterableColumns`. Currently no filter functions work on the `column` instance out of the box.

src/metabase/lib/drill_thru/column_filter.cljc
```diff
@@ -23,10 +29,26 @@
...
+        (if (= (:lib/source column) :source/aggregations)
+          (let [next-stage    (->> (lib.util/canonical-stage-index query stage-number)
+                                   (lib.util/next-stage-number query))
+                ;; next-stage is nil if there's no stage after the current one, so add one and bump the stage-number.
+                base          (if next-stage
+                                {:query        query
+                                 :stage-number next-stage}
+                                {:query        (lib.stage/append-stage query)
+                                 :stage-number -1})]
+            (assoc base :column (->> (lib.filter/filterable-columns (:query base) (:stage-number base))
```
bshepherdson, line 48: @ranquild Here's the code where the returned column is coming from `filterable-columns`.

bshepherdson:
> aggregation columns seem to be handled correctly, but I also need this column to be `filterable`, i.e. coming from `filterableColumns`. Currently no filter functions work on the `column` instance out of the box.
> 
> 

See the comment in the code; the returned column is coming from `filterable-columns` and there's nothing extra in the `lib.js/filterable-columns` wrapper either - the `column` coming back from these should be filter-ready.

src/metabase/lib/drill_thru/column_filter.cljc
```diff
@@ -23,10 +29,26 @@
...
+      (merge
+        {:lib/type   :metabase.lib.drill-thru/drill-thru
+         :type       :drill-thru/column-filter
+         :column     column
+         :initial-op initial-op}
+        ;; When the column we would be filtering on is an aggregation, it can't be filtered without adding a stage.
+        (if (= (:lib/source column) :source/aggregations)
+          (let [next-stage    (->> (lib.util/canonical-stage-index query stage-number)
+                                   (lib.util/next-stage-number query))
+                ;; next-stage is nil if there's no stage after the current one, so add one and bump the stage-number.
```
metamben, line 42:
```suggestion
                ;; next-stage is nil if there's no stage after the current one, so add one and reference the last stage.
```

metamben (approved):

ranquild:
@bshepherdson it still doesn't work

ranquild (changes requested):
Still doesn't work, check last comment

ranquild (approved):

PR #35890, opened 2023-11-16
✍️✉️: Dashboards and Cards in Custom Reports should be writable

### LINKED ISSUE #35836: [AuditV2] All items in custom reports collection are read only
### Describe the bug

Anything you save in the custom reports collection becomes read only (it gets `can_write: false`). found this in the bug bash.

https://www.notion.so/metabase/AuditV2-Bug-Bash-2a4e0f1ed37242bba90b7c267730c7e3?pvs=4#6197a4f8a241442bba88dc09ebccb9f1

### To Reproduce

- Create a question from the sample db
- save it in "our analytics" 
- see that you can edit it
- move it to "custom reports" under metabase analytics
- you can no longer edit it, see that it got a can_write flag.

### Expected behavior

you should be able to edit content in the custom_reports collection

```JSON
latest master 11/15/23
```

### Severity

P2, blocks a main analytics feature

escherize:
resolves #35836

In `can-write?`, we were checking if a dashboard or card was a child of the `Custom Report` collection.

Now we only check if the dashboard or card is a child of the `Metabase Analytics` collection, so the children of `Custom Report` should be editable (and not read-only).


src/metabase/models/card.clj
```diff
@@ -68,7 +68,7 @@
 (defmethod mi/can-write? Card
   ([instance]
    ;; Cards in audit collection should be read only
-   (if (perms/is-parent-collection-audit? instance)
+   (if (= (:collection_id instance) (:id (perms/default-audit-collection)))
```
dpsutton, line 71: on oss this is failing for every item in the root collection (`collection_id` is nil).

```clojure
dashboard=> (perms/default-audit-collection)
:metabase.models.permissions/noop
dashboard=> (:id (perms/default-audit-collection))
nil
```

We need to check that `(= perms/audit-db-id (:collection_id instance))` or we could check if the audit collection exists. But nil is a valid value that indicates root collection.

dpsutton: or of course just `(and (:id (perms/default-audit-collection)) (= (:id (perms/default-audit-collection)) ...))`

escherize: Oh, so this was making every card in the root collection to be read-only on OSS. Let's just use the id itself instead.

test/metabase/models/collection_test.clj
```diff
@@ -1681,7 +1681,7 @@
 
 (deftest instance-analytics-collections-test
   (testing "Instance analytics and it's contents isn't writable, even for admins."
-    (t2.with-temp/with-temp [Collection audit-collection {}
+    (t2.with-temp/with-temp [Collection audit-collection {:id 13371337 :type "instance-analytics"}
```
dpsutton, line 1684: you should use `perms/audit-db-id` rather than the name here. Will be quite mystifying in the future

escherize:
```suggestion
    (t2.with-temp/with-temp [Collection audit-collection {:id perms/audit-db-id :type "instance-analytics"}
```

dpsutton (approved):

PR #35891, opened 2023-11-16
`zoom`, `fk-details`, `fk-filter` drills should be returned for a native query model 

### LINKED ISSUE #35689: [MLv2] [BE] `zoom`, `fk-details`, `fk-filter` drills should be returned for a native query model
zoom, fk-details, fk-filter drills should be returned for a model based on native query with set metadata.

Please use this e2e test for reproduction - 
https://github.com/metabase/metabase/pull/35383/files#diff-4ab88f6c6e1a239ee2fc1eacee670bed33e57126cd04c070491fdfbe8a7d221bR343-R344

camsaul:
Fixes #35689

I verified this fix by running [this Cypress test](https://github.com/metabase/metabase/pull/35383/files#diff-4ab88f6c6e1a239ee2fc1eacee670bed33e57126cd04c070491fdfbe8a7d221bR343-R344), unfortunately that test looks incomplete -- there is no code to actually click one of the drills when it is presented, it just hangs waiting for a request that will never happen since it never clicks anything. I did manually verify the drills tho using the Dashboard created by the Cypress test as a starting point.

As part of this PR I renamed `drill-thru.foreign-key` to `drill-thru.fk-filter`, to better match the drill that that namespace handles, and to differentiate it from the `drill-thru.fk-details` drill.


bshepherdson (approved):
😍

PR #35893, opened 2023-11-16
[CI] Diff file changes only since the most recent commit

nemanjaglumac:


perivamsi:
Can we try this with a smaller change, say only `frontend.yml` and see if with consecutive commits (on the backend and frontend in that order) we don't trigger backend driver tests on the second commit?

nemanjaglumac:
`base` option does not work for pull requests :(

PR #35895, opened 2023-11-17
Fixing bad performance due to massive pprint statements

markbastian:
Spans on stats for massive models are very long, on the order of 4s.

The bad span seems to encompass a small amount of logging and assertion logic, including a with-out-str pprint, which I've found to be notoriously slow for large objects. This PR slims down the logged object.

If this is the culprit we should see a substantial performance improvement.

For reference, this is the offending span.

image

### DIFF
--- a/src/metabase/events.clj
+++ b/src/metabase/events.clj
@@ -10,6 +10,7 @@
   (:require
    [clojure.spec.alpha :as s]
    [metabase.events.schema :as events.schema]
+   [metabase.models.interface :as mi]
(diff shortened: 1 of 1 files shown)


perivamsi (approved):

markbastian:
BTW, in my experience [fipp](https://github.com/brandonbloom/fipp) is _much_ faster, but we really shouldn't be logging supermassive objects.

markbastian:
Fixes #33499

PR #35896, opened 2023-11-17
Fix opening bookmarked dashboard shows an empty page

### LINKED ISSUE #35959: Dashboards with pages sometimes appear blank when opened from bookmarks
### Describe the bug

When users are on the dashboard page that have tabs and they open another dashboard from bookmarks it may appear blank. Use "Fast 3G" throttling for a more reliable reproduction locally.

### To Reproduce

1. Bookmark two dashboards with at least two tabs and a few questions
2. Enable "Fast 3G" throttling in dev tools
3. Switch fast between dashboards
4. There is a high chance you will observe a blank dashboard

### Logs

_No response_

```JSON
Stats, latest master
```

### Severity

Can be very frustrating but "fixable" by a page refresh

### Additional context

_No response_

alxnddr:
Closes https://github.com/metabase/metabase/issues/35959

### Description

When users had two bookmarked dashboards with at least two pages on them and they switched fast enough, the app rendered a blank dashboard. It can be reproduced on stats with ease but for local environment you need to enable "Fast 3G" throttling in dev tools.

The reason is that we did not discard the result of `fetchDashboard` if there was a newer call of `fetchDashboard`. It resulted in updating `state.dashboard.dashboardId` to the previously fetched dashboard id. This would still work without dashboard tabs but the tabs reducer `frontend/src/metabase/dashboard/actions/tabs.ts` initializes tabs based on the `dashboardId` from the state which can be incorrect if we do not cancel the first request so it led to an invalid `selectedTabId` and hence a blank dashboard.

### How to verify

1. Bookmark two dashboards with at least two tabs and a few questions
2. Enable "Fast 3G" throttling in dev tools
3. Switch fast between dashboards
4. Every time it shows the expected content of a selected dashboard, no blank pages

#### Before

https://github.com/metabase/metabase/assets/14301985/28a831af-4f98-4857-9517-6ab916d14b3a

#### After

https://github.com/metabase/metabase/assets/14301985/49c850bb-52d9-4fce-ad52-1088736f910b

### Checklist

### DIFF
--- a/frontend/src/metabase/dashboard/actions/data-fetching.js
+++ b/frontend/src/metabase/dashboard/actions/data-fetching.js
@@ -3,7 +3,11 @@ import { getIn } from "icepick";
 import { t } from "ttag";
 
 import { denormalize, normalize, schema } from "normalizr";
-import { createAction, createThunkAction } from "metabase/lib/redux";
+import {
+  createAction,
+  createAsyncThunk,
+  createThunkAction,
+} from "metabase/lib/redux";
 import { defer } from "metabase/lib/promise";
 
 import { getDashboardUiParameters } from "metabase/parameters/utils/dashboards";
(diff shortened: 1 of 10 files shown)


kulyk (approved):
Looks good!

PR #35919, opened 2023-11-17
Stress-test a single driver test in isolation

### LINKED ISSUE #32998: Github workflow for verifying fixes of flaky BE tests
We have [a workflow for verifying fixes of flaky E2E tests](https://github.com/metabase/metabase/blob/master/.github/workflows/e2e-stress-test-flake-fix.yml). Since there are also flaky backend tests, and since the CI environment is pretty different from a developer machine (slower and weirder and more important for all the future CI runs), we want to be able to run the fixed tests in the CI environment specifically.
The way it's done with E2E tests is one test is isolated with `yarn test-cypress-run --spec /path/to/spec` and then [repeated with](https://www.npmjs.com/package/@cypress/grep#burn-repeat-tests) `npx cypress run --env grep="hello world",burn=30`.

nemanjaglumac:
This PR introduces a driver stress-test workflow.
Resolves #32998.

It needs **three** input values:
1. Full test name with its namespace (e.g. `metabase.query-processor.streaming.csv-test/sqlite-datetime-test`
2. The number of times you want to run this test
3. Driver against which we want to test this (chosen from a dropdown menu)

Here's an example of a run from my fork where I ran the above mentioned test **forty times** against SQLite driver!
https://github.com/nemanjaglumac/metabase/actions/runs/7507569955/job/20441377658

image

> [!important]
> This workflow needs to be kept in sync with the changes made to the (main) `drivers.yml`

### CLI
it should be easy to trigger this workflow from the CLI with Babashka, for example. I'm sure @escherize will be happy to cook something up quickly. In meantime, here's dumb (as in - not parameterized) NodeJS code that you can use in meantime. It can easily be parameterized, though.

```js
#!/usr/bin/node
const { Octokit } = require("@octokit/core");
const {
  restEndpointMethods,
} = require("@octokit/plugin-rest-endpoint-methods");

const MyOctokit = Octokit.plugin(restEndpointMethods);
const github = new MyOctokit({ auth: process.env["GITHUB_TOKEN"] });

const config = {
  owner: "metabase",
  repo: "metabase",
};

github.rest.actions.createWorkflowDispatch({
  ...config,
  workflow_id: "drivers-stress-test.yml",
  ref: "your-branch-here",
  inputs: {
    test: "test-name-with-namespace", // e.g. "metabase.query-processor.streaming.csv-test/sqlite-datetime-test"
    times: "40",
    driver: "SQLite" // or any other driver listed in the workflow
  },
});
```

### DIFF
--- a/.github/workflows/drivers-stress-test.yml
+++ b/.github/workflows/drivers-stress-test.yml
@@ -1,0 +1,723 @@
+name: Driver Stress Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      test:
+        description: 'The exact test name with its namespace.'
+        type: string
+        required: true
+      times:
+        description: 'How many times do you want to run this test?'
+        type: number
+        required: true
+        default: 1
+      driver:
+        description: 'The driver against which you want to run this test.'
+        required: true
+        default: 'Postgres'
+        type: choice
+        options:
+          - Athena
+          - BigQuery
+          - Druid
+          - Google Analytics
+          - Google Classpath Test
+          - Maria DB
+          - Maria DB (latest)
+          - Mongo 4
+          - Mongo 4 (SSL)
+          - Mongo 5
+          - Mongo 5 (SSL)
+          - Mongo (latest)
+          - MySQL
+          - MySQL (latest)
+          - Oracle 18
+          - Oracle 21
+          - Postgres
+          - Postgres (latest)
+          - Presto JDBC
+          - Redshift
+          - Snowflake
+          - SparkSQL
+          - SQLite
+          - SQL Server 2017
+          - SQL Server 2022
+          - Vertica
+
+jobs:
+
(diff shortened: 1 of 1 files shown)


iethree (approved):
lgtm!

iethree:
one thought: we could probably shove this logic into `drivers.yml`, which would make that workflow messier, but would relieve us from the burden of keeping them up to date. I think, on the whole, I like having this file separate, but wanted to mention it.

nemanjaglumac:
> one thought: we could probably shove this logic into `drivers.yml`, which would make that workflow messier, but would relieve us from the burden of keeping them up to date. I think, on the whole, I like having this file separate, but wanted to mention it.

I thought about it initially.
That file is already hard to read/follow. I'm open to trying it out, though.

Let's first merge this and see how it works.
We can merge the files later on if people wish to.

PR #35928, opened 2023-11-17
Toucan 2 × MLv2 metadata

camsaul:
Support fetching MLv2 metadata as it would be returned by the application database metadata provider directly in Toucan 2.

```clj
(t2/select-one :metadata/column 219255)
=>
{:description nil,
 :lib/type :metadata/column,
 :base-type :type/Integer,
 :semantic-type :type/FK,
 :table-id 51222,
 :name "USER_ID",
 :coercion-strategy nil,
 :settings nil,
 :lib/external-remap
 {:lib/type :metadata.column.remapping/external, :id 1029, :name "User ID [external remap]", :field-id 219261},
 :nfc-path nil,
 :database-type "INTEGER",
 :effective-type :type/Integer,
 :fk-target-field-id 219259,
 :id 219255,
 :parent-id nil,
 :visibility-type :normal,
 :display-name "User ID",
 :position 1,
 :fingerprint {:global {:distinct-count 929, :nil% 0.0}}}
```

Eliminates bespoke functions for doing this stuff in `metabase.lib.jvm.metadata-provider` and implements Toucan 2 methods instead. 

The main advantage of this change is easy REPL usage, for example 

```clj
(t2/select-one :metadata/column (mt/id :orders :id))
```

is a lot more convenient from the REPL than

```clj
(lib.metadata/field (lib.metadata.jvm/application-database-metadata-provider (mt/id)) (mt/id :orders :id))
```

and you can now do stuff like 

```clj
;; make a map of Field ID to Field name for Integer columns in the Orders table
(t2/select-fn->fn :id :name :metadata/column :table_id (mt/id :orders) :base_type :type/Integer)
```

that is simply impossible with just a metadata provider.

Other advantage: show off some of the cool things you can do with Toucan 2.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -73,7 +73,7 @@
   info.sunng/ring-jetty9-adapter            {:mvn/version "0.22.3"}             ; Drop-in replacement for official Ring Jetty adapter. Supports Jetty 11 webserver.
   instaparse/instaparse                     {:mvn/version "1.4.12"}             ; Make your own parser
   clj-commons/clj-yaml                      {:mvn/version "1.0.27"}             ; Clojure wrapper for YAML library SnakeYAML
-  io.github.camsaul/toucan2                 {:mvn/version "1.0.524"}
+  io.github.camsaul/toucan2                 {:mvn/version "1.0.535"}
   io.github.eerohele/pp                     {:git/tag "2023-10-05.5"            ; super fast pretty-printing library
                                              :git/sha "7059eec"}
   ;; The 2.X line of Resilience4j requires Java 17, so we cannot upgrade this dependency until that is our minimum JVM version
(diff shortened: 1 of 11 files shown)


deps.edn
```diff
@@ -73,7 +73,7 @@
   info.sunng/ring-jetty9-adapter            {:mvn/version "0.22.3"}             ; Drop-in replacement for official Ring Jetty adapter. Supports Jetty 11 webserver.
   instaparse/instaparse                     {:mvn/version "1.4.12"}             ; Make your own parser
   clj-commons/clj-yaml                      {:mvn/version "1.0.27"}             ; Clojure wrapper for YAML library SnakeYAML
-  io.github.camsaul/toucan2                 {:mvn/version "1.0.524"}
+  io.github.camsaul/toucan2                 {:mvn/version "1.0.535"}
```
camsaul, line 76: Minor Toucan 2 version bump. No big changes but it has a few bugfixes and simplifies some of the interfaces we need to implement

bshepherdson (approved):

PR #35933, opened 2023-11-17
[MLv2] Short name display for filter clauses

### LINKED ISSUE #35752: [MLv2] BE - Add shortDisplayName in displayInfo for filters
We need a way to display the filter value without the column name. It's currently used in the timeseries chrome. The proposed change is to add `shortDisplayName` in `displayInfo` for (all or just filters) clauses that excludes the column name.

Master:
Screenshot 2023-11-15 at 17 01 30

In the branch:
Screenshot 2023-11-15 at 16 47 10

snoe:
Fixes #35752

Built for the timeseries widget and the associated operators. Generally shows the user selected dates and occasionally shows an operator. The choices are about what looks correct more than following particular rules.


src/metabase/lib/fe_util.cljc
```diff
@@ -56,3 +62,46 @@
    args     :- [:sequential :any]
    options  :- [:maybe :map]]
   (lib.options/ensure-uuid (into [operator options] (map lib.common/->op-arg) args)))
+
+(mu/defn filter-args-display-name :- :string
+  "Provides a reasonable display name for the `filter-clause` excluding the column-name.
+   Can be expanded as needed but only currently defined for a narrow set of date filters.
```
metamben, line 68:
```suggestion
   Can be expanded as needed but currently only defined for a narrow set of date filters.
```

src/metabase/lib/fe_util.cljc
```diff
@@ -56,3 +62,46 @@
...
+  [query stage-number filter-clause]
+  (let [->temporal-name #(shared.ut/format-unit % nil)
+        temporal? #(lib.util/original-isa? % :type/Temporal)
+        unit-is (fn [unit-or-units]
+                  (let [units (set (u/one-or-many unit-or-units))]
+                    (fn [maybe-clause]
+                      (clojure.core/and
+                        (temporal? maybe-clause)
+                        (lib.util/clause? maybe-clause)
+                        (clojure.core/contains? units (:temporal-unit (second maybe-clause)))))))]
```
metamben, line 80: Are `clojure.core/and` and `clojure.core/contains?` shadowed here?

src/metabase/lib/core.cljc
```diff
@@ -161,7 +161,8 @@
   lower]
  [lib.fe-util
   expression-clause
-  expression-parts]
+  expression-parts
+  filter-args-display-name]
```
metamben, line 165: This functions should be used when rendering static viz stuff, right? Does it still count as a FE utility function?

metamben (approved):
LGTM

src/metabase/lib/core.cljc
snoe: I don't believe so, in static viz the timeseries widgets are not shown.
New question-11_21_2023, 10_15_29 AM

ranquild (approved):

PR #35935, opened 2023-11-17
Fix macroexpansion of `mu/defn` for hygiene

dpsutton:
Edit: note that this fixes a correctness bug. This bug was introduced in https://github.com/metabase/metabase/pull/34712 to clean up stacktraces. It's totally fine to reject this PR and revert #34712 until it is ready.

We've got a bug in our `mu/defn`:

```clojure
(mu/defn myfn
  [one two]
  (mbql.u/match-one [:= 9 10]
    [:= a b] [a b]))

(myfn 1 2) ; nil <- should be [9 10]
(myfn 9 10) ; [9 10]
```

This used to macroexpand to

```clojure
(def myfn
  "Inputs: [one two]\n  Return: :any"
  (let [&f (fn [one two]
             (mbql.u/match-one [:= 9 10]
               [:= a b]
               [a b]))]
    (fn ([a b] (&f a b)))))
```

Note that `a` and `b` are not lexically available to the `&f` function so no collisions can occur.

But after https://github.com/metabase/metabase/pull/34712 it now expands to

```clojure
(def
  myfn
  "Inputs: [one two]\n  Return: :any"
  (clojure.core/fn
    ([a b]
     ((clojure.core/fn [one two]
        (mbql.u/match-one [:= 9 10] [:= a b] [a b])) ;; here [a b] pattern is matching against the args
      a b))))
```

The problem is `a` and `b` are now anaphors.

Example:

```clojure
user=> (mu/defn foo [x] a) ;; not great
#'user/foo
user=> (foo 1)
1
```

A fix here is to gensym the parameters, but allow suppressing that for testing so we have deterministic output (thus `*sanitize?*`)

After: all these are fixed.
```clojure
card=> (mu/defn myfn
         [one two]
         (mbql.u/match-one [:= 9 10]
                           [:= a b] [a b]))
#'metabase.query-processor.card/myfn
card=> (myfn 1 2)
[9 10]
card=> (myfn 9 10)
[9 10]
card=> (mu/defn foo [x] a)
Syntax error compiling at (REPL:260:1).
Unable to resolve symbol: a in this context
```

### DIFF
--- a/src/metabase/util/malli/fn.clj
+++ b/src/metabase/util/malli/fn.clj
@@ -196,25 +196,30 @@
                  (= (first schema) :*)))]
     (star-schema? (last args))))
 
-(defn- input-schema-arg-names [[_cat & args :as input-schema]]
+(def ^:dynamic *sanitize?*
+  "Whether to 'sanitize' the arg lists."
+  true)
+
+(defn- input-schema-arg-names [sanitizer [_cat & args :as input-schema]]
   (let [varargs?    (varargs-schema? input-schema)
         normal-args (if varargs?
                       (butlast args)
                       args)]
-    (concat
-     (for [n (range (count normal-args))]
-       (symbol (str (char (+ (int \a) n)))))
-     (when varargs?
-       ['more]))))
+    (map (if *sanitize?* sanitizer identity)
+         (concat
+          (for [n (range (count normal-args))]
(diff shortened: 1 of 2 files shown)


dpsutton:
Closing this in favor of https://github.com/metabase/metabase/pull/35939
I rather like the cleanliness of the original style.

PR #35940, opened 2023-11-20
Truncate audit_log and view_log based on `AUDIT_MAX_RETENTION_DAYS` env var

### LINKED ISSUE #34985: Make `MB_AUDIT_MAX_RETENTION_DAYS` truncate data from the `view_log` or `audit_log` table

noahmoss:
Closes https://github.com/metabase/metabase/issues/34985

* Expands the `truncate-audit-log` task, which previously only affected `query_execution` (despite its name), to additionally truncate `view_log` and `audit_log` on EE instances
* Renames the task to `truncate-audit-tables`
* Adds some extra tests for the new tables

This doesn't change anything about how the env var itself behaves, like its default value or what values are allowed. I've asked for some clarifications about this [here](https://metaboat.slack.com/archives/C064EB1UE5P/p1700506789035339) and I'll probably make these changes in a follow-up PR.


src/metabase/task/truncate_audit_tables.clj
```diff
@@ -0,0 +1,91 @@
...
+
+(defenterprise audit-models-to-truncate
+  "List of models to truncate. OSS implementation only truncates `query_execution` table."
+  metabase-enterprise.task.truncate-audit-tables
+  []
+  {:model/QueryExecution :started_at})
+
+(defn- truncate-audit-tables!
+  []
+  (dorun
```
dpsutton, line 66: I think it's better to use `run!` rather than the `dorun (map ...` style here.

```clojure
(map
  (fn [[model time-column]]
    (truncate-table! model time-column))
  (audit-models-to-truncate))

;; vs

(run! (fn [[model time-column]]
        (truncate-table! model time-column))
      (audit-models-to-truncate))
```

noahmoss: fair, I always forget about `run!`, thanks

enterprise/backend/src/metabase_enterprise/task/truncate_audit_tables.clj
```diff
@@ -0,0 +1,12 @@
+(ns metabase-enterprise.task.truncate-audit-tables
+  (:require
+   [metabase.public-settings.premium-features :refer [defenterprise]]))
+
+(defenterprise audit-models-to-truncate
+  "List of models to truncate, as well as the name of the column containing the row's timestamp. EE version adds
+  `audit_log` and `view_log` truncation"
+  :feature :audit-app
+  []
+  {:model/QueryExecution :started_at
```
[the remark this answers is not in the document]
noahmoss: Good call. I like having both of them as separate values.

dpsutton:
@noahmoss you are removing the `defjob TruncateAuditLog` in `metabase.task.truncate_audit_log` and replacing it with `defjob TruncateAuditTables` in a different namespace. But stats at least will have a trigger for this job to periodically run, and will presumably throw lots of errors saying it cannot find the `metabase.task.truncate_audit_log.TruncateAuditLog` class.

Do we need to write some kind of migration, or do we perhaps want to keep the existing job and not rename the job and namespace?

```clojure
{:key "metabase.task.truncate-audit-log.job",
 :class "metabase.task.truncate_audit_log.TruncateAuditLog",
 :description nil,
 :concurrent-execution-disallowed? false,
 :durable? false,
 :requests-recovery? false,
 :triggers ({:description nil,
             :schedule "0 0 */12 * * ? *",
             :key "metabase.task.truncate-audit-log.trigger",
             :previous-fire-time #inst "2023-11-28T12:00:00.000-00:00",
             :start-time #inst "2023-11-21T21:02:06.000-00:00",
             :misfire-instruction "DO_NOTHING",
             :end-time nil,
             :state "NORMAL",
             :priority 5,
             :next-fire-time #inst "2023-11-29T00:00:00.000-00:00",
             :may-fire-again? true,
             :final-fire-time nil,
             :data {}})}
```

src/metabase/task/truncate_audit_tables.clj
```diff
@@ -0,0 +1,115 @@
...
+(defmethod task/init! ::TruncateAuditTables [_]
+  (let [job     (jobs/build
+                 (jobs/of-type TruncateAuditTables)
+                 (jobs/with-identity (jobs/key truncate-audit-tables-job-key)))
+        trigger (triggers/build
+                 (triggers/with-identity (triggers/key truncate-audit-tables-trigger-key))
+                 (triggers/start-now)
+                 (triggers/with-schedule
+                   (cron/schedule
+                    ;; run every 12 hours
```
johnswanson, line 112: Nit: would this comment be better above line 101, so we don't forget to update the comment if we update the cron?

noahmoss:
@dpsutton Good point. It's easy to forget about that when renaming or removing a Quartz job.

I don't like it when I can't rename something for an arbitrary technical reason, so I'll add a migration to remove the old trigger.

But this raises a question: if a trigger tries to fire but can't find an associated class, why can't it just be removed automatically? Maybe we can just catch those exceptions and delete the triggers when they try to run, since it implies that the job has been removed.

johnswanson (approved):
LGTM!

escherize (commented):
Good idea to simply log the missing job. Love how it isn't magical. Removing them when they're not found could lead to some serious spelunking for deleted jobs.

Issue #35290, opened 2023-11-01
[MLv2] - Visible Columns errors when source query is a table joined on itself


npfitz:
Right now in Stats, if you create a question that is Orders joined onto Orders, save the question and then use that as the source for another question you will get an error if you try to open the viz settings panel 

<img width="708" alt="chrome_fz4zilLelE" src="https://github.com/metabase/metabase/assets/1328979/2ea4891e-8b8a-4b15-a661-9a0a9376e8ab">
<img width="441" alt="chrome_92AL5CBtf1" src="https://github.com/metabase/metabase/assets/1328979/07584f34-d5cc-405f-acdf-efd16cdedaec">

bshepherdson:
My defensive check is serving its purpose here. When coming from an earlier stage the names are used rather than the IDs. But when doing that we should be detecting the collision and using the full column alias rather than the simple column name.

I'm busy with #35205 right now, so if someone else from the QP+D team wants to pick this up, here's my pointers:
- Check if the ambiguous matches have any fields (eg. `:join-alias) we can use to disambiguate them; if so, add a new check under `metabase.lib.equality/disambiguate-matches-prefer-explicit`.
- Failing that, we may need to update how `ref`s are generated, to make sure they're using unambiguous names in cases like this. (Like using the `:lib/desired-column-alias` rather than `:name`.)

bshepherdson:
This might be related to #33427 or might not be.

bshepherdson:
The error that causes this to crash is turned into a warning by #36197 . That mitigates the impact of this bug so I'm removing this from the 48 blockers.

darksciencebase:
removing this from the board as it's highly likely caused by https://github.com/metabase/metabase/issues/36185 and will be tackled as part of that cluster

uladzimirdev:
the issue is not valid, I checked 49.13 and 50

PR #35942, opened 2023-11-20
Fix - Question editor does not work if model joins the same table more than once

### LINKED ISSUE #33427: Question editor does not work if model joins the same table more than once
### Describe the bug

Question editor does not work with models when the model joins the same table more than 1 time.
 

### To Reproduce

1. Create a SQL based model similar to the following (the query must join to the same table twice):
```
select pr.purchase_report_id, a1.name as created_by, a2.name as updated_by
from purchase_report pr
    join account a1 on a1.account_id = pr.created_by
    join account a2 on a2.account_id = pr.updated_by;
```
2. Adding or removing model metadata has no impact on bug.
3. Create a question that uses the above model (i.e. New -> Question).
4. Click preview button and note that the question works when returning all columns as shown in following screenshot.
image
5. Use the drop down next to the model name to select specific columns.  You will see the updated_by column is listed 2 times as shown in the following screenshot. 
image

### Expected behavior

Column select drop down should show created_by and updated_by

### Logs

_No response_

```JSON
Chrome 115
Windows 10
MySQL
Metabase version 1.47.0 
Metabase hosting environment: Docker
Metabase internal database: MySQL
```

### Severity

Blocking your usage of Model feature entirely

### Additional context

Our shop has a number of databases where you must join the same table multiple times in a single query.  
Our current workaround is to create database views in MySQL instead of using Metabase models.
I also have trouble joining the same table multiple times when using the non-SQL based question editor but this seems to be fixed in version 47.0.

kamilmielnik:
Fixes #33427

### Description

See [Slack discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1699978395628049?thread_ts=1699902023.771949&cid=C04CYTEL9N2).

This PR fixes how the columns are displayed in the dropdown in GUI editor, but it does not fix what happens when you try to deselect/select these columns, see [video](https://github.com/metabase/metabase/assets/6830683/3f2b0d10-d1e3-430c-b3c5-25df61d04eeb) - it's been reported separately: #35290.

### How to verify

See demo below or follow steps from this comment: https://github.com/metabase/metabase/issues/33427#issuecomment-1690861463.

#### Before

https://github.com/metabase/metabase/assets/6830683/d599fadd-3315-40ef-b19d-0b5f7cdba93a

#### After

https://github.com/metabase/metabase/assets/6830683/cf89f008-73ae-40e3-9152-f1fc38715a5d

### Discussions

- `getColumnGroups` now returns incorrect groups (e2e tests fail `table-column-settings.cy.spec.js`):
  - `should be able to show and hide fields from a nested query with joins and fields (metabase#32373)`
  - `should be able to show and hide implicitly joinable fields for a nested query with joins and fields`
  - `should be able to show and hide implicitly joinable fields for a nested query`
- `metadata.field` interface needs to receive `name` now, but there are 193 usages of this function
- `ChartSettingAddRemoveColumns.tsx` - `filteredColumns` are matched by `displayName` - this is incorrect

### DIFF
--- a/frontend/src/metabase-lib/metadata/utils/fields.ts
+++ b/frontend/src/metabase-lib/metadata/utils/fields.ts
@@ -45,9 +45,7 @@ export function getUniqueFieldId({
   const fieldIdentifier = getFieldIdentifier({ id, name });
 
   if (isVirtualCardId(table_id)) {
-    const isVirtual = typeof id !== "number";
-
-    if (isVirtual && name) {
+    if (name) {
       return `${table_id}:${fieldIdentifier}:${name}`;
     }
 
--- a/frontend/src/metabase-lib/metadata/utils/fields.ts
+++ b/frontend/src/metabase-lib/metadata/utils/fields.ts
@@ -45,6 +45,10 @@ export function getUniqueFieldId({
   const fieldIdentifier = getFieldIdentifier({ id, name });
 
   if (isVirtualCardId(table_id)) {
+    if (name) {
+      return `${table_id}:${fieldIdentifier}:${name}`;
+    }
+
     return `${table_id}:${fieldIdentifier}`;
   }
 
--- a/frontend/src/metabase-lib/metadata/utils/fields.ts
+++ b/frontend/src/metabase-lib/metadata/utils/fields.ts
@@ -45,7 +45,9 @@ export function getUniqueFieldId({
   const fieldIdentifier = getFieldIdentifier({ id, name });
 
   if (isVirtualCardId(table_id)) {
-    if (name) {
+    const isVirtual = typeof id !== "number";
+
+    if (isVirtual && name) {
       return `${table_id}:${fieldIdentifier}:${name}`;
     }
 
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
```diff
@@ -221,7 +221,7 @@ export class UnconnectedDataSelector extends Component {
     selectedDataBucketId: PropTypes.string,
     selectedDatabaseId: PropTypes.number,
     selectedSchemaId: PropTypes.string,
-    selectedTableId: PropTypes.number,
+    selectedTableId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
kamilmielnik, line 224: Virtual card tables have string ids (`card__XYZ`)

ranquild (approved):
FE lgtm, please wait for @bshepherdson 's approval

frontend/src/metabase-lib/metadata/utils/fields.ts
```diff
@@ -45,6 +45,12 @@ export function getUniqueFieldId({
   const fieldIdentifier = getFieldIdentifier({ id, name });
 
   if (isVirtualCardId(table_id)) {
+    const isVirtual = typeof id !== "number";
+
+    if (isVirtual && name) {
```
kamilmielnik, line 50: I had to add `isVirtual` part to this condition, otherwise 4 e2e tests were failing.

kamilmielnik:
I had to update condition in `frontend/src/metabase-lib/metadata/utils/fields.ts` because e2e tests were failing.

@ranquild can you please have another look?

kamilmielnik:
Closing, https://github.com/metabase/metabase/issues/33427 has already been fixed

Issue #30134, opened 2023-04-15
Serialization won't respect column sorting in pivot tables


paoliniluis:
**Describe the bug**
We don't seem to respect the column orders (viz settings) defined in the pivot tables when loading them in another instance

**Logs**
NA

**To Reproduce**
Using https://github.com/paoliniluis/metabase-serialization as the base
1) create a pivot table like 
![image](https://user-images.githubusercontent.com/1711649/232174494-3e7cc868-b99d-442a-8156-d3b9691b22bb.png)
2) then go to the viz settings and change the order of the source column
![image](https://user-images.githubusercontent.com/1711649/232174520-fc304b2b-5696-41b0-9a61-0eebf6167aa5.png)
3) now dump and load
4) go to the destination and see that the order was not respected
![image](https://user-images.githubusercontent.com/1711649/232174542-1938d04c-8b91-4646-ac65-1d6149de296f.png)

**Expected behavior**
We should respect the order that the users defined when loading the cards in the destination

**Screenshots/videos**
Above

**Information about your Metabase Installation:**
- Metabase version: 1.46.1

**Severity**
P2

**Additional context**
NA

laura-radaelli:
We are experiencing this issue and a related one: the pivoting column is not respected. 

The same seems to happen in the example reported here: after import (step 4.), "Product -> Category" is listed under Rows, even though, at the source, it was listed under Columns.

PR #35944, opened 2023-11-20
Fix pivot table de/serialization

### LINKED ISSUE #30133: Pivot table with 2 group by's gets loaded incorrectly on serialization
**Describe the bug**

**Logs**
Please include JavaScript console logs (open the browser dev tools and check the "console" tab) as well as the server logs (Metabase logs that you can get from Settings -> Admin -> Troubleshooting -> Logs) around the time this bug occurred. For information about how to get these, consult our [bug troubleshooting guide](https://metabase.com/docs/latest/troubleshooting-guide/bugs.html)

**To Reproduce**
Using https://github.com/paoliniluis/metabase-serialization as the base
1) create a question like the following:
image
2) view it as a pivot table and make sure it's configured like
image
3) now dump and load in destination
4) see that it changed
image

**Expected behavior**
It should be loaded in the same way as it was saved in the origin

**Screenshots/videos**
Above

- Metabase version: 1.46.1

**Severity**
P2

**Additional context**
NA

sansolovyov:
Just checks that pivot settings are exported and imported, seems like that was a problem half a year ago.

resolves #30133 and #30134

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
@@ -757,9 +757,8 @@
                                                                             :breakout     [[:field field-id nil]]}}
                                                  :visualization_settings
                                                  {:pivot_table.column_split
-                                                  {:columns []
-                                                   :rows    [["field" pf1 nil]
-                                                             ["field" pf2 nil]]
+                                                  {:columns [["field" pf1 nil]]
+                                                   :rows    [["field" pf2 nil]]
                                                    :values  [["aggregation" 0]]}
                                                   :column_settings
                                                   {(format "[\"ref\",[\"field\",%s,null]]" pf1)
@@ -773,8 +772,10 @@
               (let [viz (:visualization_settings
                          (t2/select-one Card :entity_id (:entity_id card)))]
                 (is (contains? viz :pivot_table.column_split))
-                (is (= 2
-                       (count (get-in viz [:pivot_table.column_split :rows]))))
+                (is (=? [[:field int? nil]]
+                        (get-in viz [:pivot_table.column_split :rows])))
+                (is (=? [[:field int? nil]]
+                       (get-in viz [:pivot_table.column_split :columns])))
                 (is (= "descending"
                        (get-in viz [:column_settings
                                     (format "[\"ref\",[\"field\",%s,null]]" pf1)
--- a/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
@@ -27,9 +27,9 @@
    [reifyhealth.specmonstah.core :as rs]
    [toucan2.core :as t2]
    [toucan2.tools.with-temp :as t2.with-temp])
- (:import
-  (java.io File)
-  (java.nio.file Path)))
+  (:import
+   (java.io File)
+   (java.nio.file Path)))
 
 (set! *warn-on-reflection* true)
 
@@ -735,3 +735,47 @@
                     (is (thrown-with-msg? Exception #"Please upgrade"
                                           (cmd/v2-load! dump-dir {}))
                         "throws")))))))))))
+
+(deftest pivot-export-test
+  (testing "Pivot table export and load correctly"
+    (ts/with-random-dump-dir [dump-dir "serdesv2-"]
+      (ts/with-source-and-dest-dbs
+        (ts/with-source-db
+          (t2.with-temp/with-temp
+            [Database   {db-id :id}             {:name "Pivot Database"}
+             Table      {table-id :id}          {:name "Pivot Table"}
+             Field      {pf1 :id}               {:name "Pivot Field 1"}
+             Field      {pf2 :id}               {:name "Pivot Field 2"}
+             Field      {field-id :id}          {:name "Some Field"}
+             Collection {coll-id :id}           {:name "Pivot Collection"}
+             Card       card                    {:name          "Pivot Card"
+                                                 :collection_id coll-id
+                                                 :dataset_query {:type     :query
+                                                                 :database db-id
+                                                                 :query    {:source-table table-id
+                                                                            :aggregation  [:sum [:field pf1 nil]]
+                                                                            :breakout     [[:field field-id nil]]}}
+                                                 :visualization_settings
+                                                 {:pivot_table.column_split
+                                                  {:columns []
+                                                   :rows    [["field" pf1 nil]
+                                                             ["field" pf2 nil]]
+                                                   :values  [["aggregation" 0]]}
+                                                  :column_settings
+                                                  {(format "[\"ref\",[\"field\",%s,null]]" pf1)
+                                                   {:pivot_table.column_sort_order "descending"}}}}]
+            (-> (serdes/with-cache (into [] (extract/extract {})))
+                (storage/store! dump-dir))
+
+            (ts/with-dest-db
+              (serdes/with-cache (serdes.load/load-metabase! (ingest/ingest-yaml dump-dir)))
+
+              (let [viz (:visualization_settings
+                         (t2/select-one Card :entity_id (:entity_id card)))]
+                (is (contains? viz :pivot_table.column_split))
+                (is (= 2
+                       (count (get-in viz [:pivot_table.column_split :rows]))))
+                (is (= "descending"
+                       (get-in viz [:column_settings
+                                    (format "[\"ref\",[\"field\",%s,null]]" pf1)
+                                    :pivot_table.column_sort_order])))))))))))
--- a/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
@@ -738,19 +738,21 @@
 
 (deftest pivot-export-test
   (testing "Pivot table export and load correctly"
-    (let [card1s (atom nil)]
+    (let [old-ids (atom nil)
+          card1s  (atom nil)]
       (ts/with-random-dump-dir [dump-dir "serdesv2-"]
         (ts/with-source-and-dest-dbs
           (ts/with-source-db
             (mt/dataset sample-dataset
+              ;; ensuring field ids are stable by loading dataset in db first
+              (mt/db)
               (mt/$ids nil
                 (t2.with-temp/with-temp
-                  [Database   {db-id :id}    {:name "Pivot Database"}
-                   Collection {coll-id :id}  {:name "Pivot Collection"}
+                  [Collection {coll-id :id}  {:name "Pivot Collection"}
                    Card       card           {:name          "Pivot Card"
                                               :collection_id coll-id
                                               :dataset_query {:type     :query
-                                                              :database db-id
+                                                              :database (mt/id)
                                                               :query    {:source-table $$orders
                                                                          :aggregation  [:sum [:field %orders.id nil]]
                                                                          :breakout     [[:field %orders.user_id nil]]}}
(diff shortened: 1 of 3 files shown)


calherries:
@piranha Just confirming, did you check these tests fail with 46.1?

piranha:
@calherries ohhh I did not and will do that right now

piranha:
I haven't been able to reproduce row/column moving to rows, but sorting problem reproduces right away.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,48 @@
...
+                                                 :collection_id coll-id
+                                                 :dataset_query {:type     :query
+                                                                 :database db-id
+                                                                 :query    {:source-table table-id
+                                                                            :aggregation  [:sum [:field pf1 nil]]
+                                                                            :breakout     [[:field field-id nil]]}}
+                                                 :visualization_settings
+                                                 {:pivot_table.column_split
+                                                  {:columns [["field" pf1 nil]]
+                                                   :rows    [["field" pf2 nil]]
```
calherries, line 761: I wonder how you got this data structure shape. When I tried to reproduce the issue (https://github.com/metabase/metabase/issues/30133) I got this:

```json
    "visualization_settings": {
        "pivot_table.column_split": {
            "rows": [
                [
                    "field",
                    45,
                    {
                        "base-type": "type/Text",
                        "source-field": 43
                    }
                ]
            ],
            "columns": [
                [
                    "field",
                    58,
                    {
                        "base-type": "type/Text",
                        "source-field": 40
                    }
                ]
            ],
            "values": [
                [
                    "aggregation",
                    0
                ],
                [
                    "aggregation",
                    1
                ],
                [
                    "aggregation",
                    2
                ],
                [
                    "aggregation",
                    3
                ],
                [
                    "aggregation",
                    4
                ]
            ]
        }
    },
```

piranha:
@calherries thanks a lot! I really did test on sample data, and it worked with no issues. New test catches that.

src/metabase/models/serialization.clj
```diff
@@ -1192,6 +1192,16 @@
           (m/update-existing    :click_behavior import-viz-click-behavior-link)
           (m/update-existing-in [:click_behavior :parameterMapping] import-viz-click-behavior-mappings)))
 
+(defn- export-pivot-table [settings]
+  (some-> settings
+          (m/update-existing-in [:pivot_table.column_split :rows] ids->fully-qualified-names #_export-pivot-table-fields)
+          (m/update-existing-in [:pivot_table.column_split :columns] ids->fully-qualified-names #_export-pivot-table-fields)))
```
calherries, line 1198: same below
```suggestion
  (some-> settings
          (m/update-existing-in [:pivot_table.column_split :rows] ids->fully-qualified-names)
          (m/update-existing-in [:pivot_table.column_split :columns] ids->fully-qualified-names)))
```

calherries (approved):
Looks fine to merge but why does using `test-data` surface the issue but `sample-dataset` doesn't?

We need to make sure all the IDs in the destination DB don't just happen to align with the IDs in the source DB by chance, because they both contain the same datasets. I don't have a solution to this, I'm just stating the problem here

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,54 @@
...
+                  (-> (serdes/with-cache (into [] (extract/extract {})))
+                      (storage/store! dump-dir))))))
+
+          (ts/with-dest-db
+            (mt/dataset test-data
+              (mt/$ids nil
+                (serdes/with-cache (serdes.load/load-metabase! (ingest/ingest-yaml dump-dir)))
+
+                (let [viz (:visualization_settings
+                           (t2/select-one Card :entity_id (:entity_id @card1s)))]
```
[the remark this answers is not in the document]
piranha: Hah, I specifically wrote it in this way so that query is on one line and getter is on the other; but I guess I could even use `select-one-fn` here. :)

src/metabase/models/serialization.clj
piranha: oops 😁

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,54 @@
...
                         "throws")))))))))))
+
+(deftest pivot-export-test
+  (testing "Pivot table export and load correctly"
+    (let [card1s (atom nil)]
+      (ts/with-random-dump-dir [dump-dir "serdesv2-"]
+        (ts/with-source-and-dest-dbs
+          (ts/with-source-db
+            (mt/dataset test-data
+              (mt/$ids nil
```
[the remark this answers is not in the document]
piranha: Ehh, this only works when there is only 1 child. It's quite irritating that it works like that, since it bites you when you least expect it. :)

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,54 @@
...
+                                               :column_settings
+                                               {(format "[\"ref\",[\"field\",%s,null]]" %users.name)
+                                                {:pivot_table.column_sort_order "descending"}}}}]
+                  (reset! card1s card)
+                  (-> (serdes/with-cache (into [] (extract/extract {})))
+                      (storage/store! dump-dir))))))
+
+          (ts/with-dest-db
+            (mt/dataset test-data
+              (mt/$ids nil
```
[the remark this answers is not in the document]
piranha: For example in that case there are two children and the removal of `nil` breaks the test (since `load-metabase!` call is used as a table name, haha). So I don't really want to remove nils in both cases - it's less ambiguous this way.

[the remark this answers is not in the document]
calherries: Yes, you're right.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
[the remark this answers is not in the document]
calherries: Both are fine objectively, it's just a little strange to read this on two separate lines, I can't quite articulate why.

In my head, this kind of thing is so familiar, I want to read this in one step, so I want it on one line.
```clojure
(let [viz (:visualization_settings (t2/select-one Card :entity_id (:entity_id @card1s)))] ...
;; or better
(let [viz (t2/select-one-fn :visualization_settings Card :entity_id (:entity_id @card1s)))] ...
;; reads as "get visualization_settings from the card"

(let [viz (:visualization_settings 
           (t2/select-one Card :entity_id (:entity_id @card1s)))
;; reads as "get the card, then get the visualization_settings from the card"
```

piranha:
> Looks fine to merge but why does using test-data surface the issue but sample-dataset doesn't?

It does surface the issue too, but you have to have different db, which is something I did not realize at first. :)

I redid test with a sample-dataset and it still behaves in an expected way, plus it checks that column ids are indeed different between two dbs.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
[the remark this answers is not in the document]
piranha: Personal preferences, I guess. :) I'm used to 80 chars line limit, and to structuring most of the complex actions on line-by-line basis, so that two-line expression looked natural to me. Especially given how long `:visualization_settings` is — it just felt uncluttered to put it on a separate line. :)

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,56 @@
...
+          (ts/with-dest-db
+            (mt/dataset sample-dataset
+              (mt/$ids nil
+                (serdes/with-cache (serdes.load/load-metabase! (ingest/ingest-yaml dump-dir)))
+
+                (let [viz (t2/select-one-fn :visualization_settings Card :entity_id (:entity_id @card1s))]
+                  (is (contains? viz :pivot_table.column_split))
+                  ;; column ids are different in different dbs
+                  (is (not= (get-in @card1s [:visualization_settings :pivot_table.column_split :rows 0 1])
+                            (get-in viz [:pivot_table.column_split :rows 0 1])))
```
piranha, line 780: @calherries I'm a bit afraid this line will flake (there is no promise that ids will indeed be different), but it's like an insurance that we're getting some real action here.

calherries, line 780: better to use `testing` blocks to add meaningful error descriptions instead of just leaving a comment
```suggestion
                  (testing "check column ids are different in different dbs, because otherwise this test could be a false negative"
                    (is (not= (get-in @card1s [:visualization_settings :pivot_table.column_split :rows 0 1])
                              (get-in viz [:pivot_table.column_split :rows 0 1]))))
```

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,56 @@
...
+                  (-> (serdes/with-cache (into [] (extract/extract {})))
+                      (storage/store! dump-dir))))))
+
+          (ts/with-dest-db
+            (mt/dataset sample-dataset
+              (mt/$ids nil
+                (serdes/with-cache (serdes.load/load-metabase! (ingest/ingest-yaml dump-dir)))
+
+                (let [viz (t2/select-one-fn :visualization_settings Card :entity_id (:entity_id @card1s))]
+                  (is (contains? viz :pivot_table.column_split))
```
calherries, line 777: What is this testing on an abstract level? I'd suggest to make it clearer with a `testing` description

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,56 @@
...
+                  ;; column ids are different in different dbs
+                  (is (not= (get-in @card1s [:visualization_settings :pivot_table.column_split :rows 0 1])
+                            (get-in viz [:pivot_table.column_split :rows 0 1])))
+                  (is (= [[:field %people.name {:base-type    :type/Text
+                                                :source-field %orders.user_id}]]
+                         (get-in viz [:pivot_table.column_split :rows])))
+                  (is (= [[:field %products.title {:base-type    :type/Text
+                                                   :source-field %orders.product_id}]]
+                         (get-in viz [:pivot_table.column_split :columns])))
```
calherries, line 786: These deserve a `testing` block too, maybe even referencing the old issue `(metabase#30133)`

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -735,3 +735,56 @@
...
+                  (is (= "descending"
+                         (get-in viz [:column_settings
+                                      (format "[\"ref\",[\"field\",%s,null]]" %people.name)
+                                      :pivot_table.column_sort_order]))))))))))))
```
calherries, line 790: I'd create a `testing` block for this, maybe even referencing the old issue `(metabase#30134)`

calherries (commented):
You should get in the habit of documenting what we're testing. Test code is specific and concrete (unless we're using generative tests) but the logic we're testing should apply to an abstract state space, and that needs explaining in english. Simply writing "pivot table export and load works correctly" isn't enough for our future selves to work out what exactly we're testing here.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
calherries: Yeah, that's not ideal. Can you find a way to ensure that the test IDs are different, at least with high probability? I'm thinking of a hack like creating an extra database in the destination DB with one field and one table.

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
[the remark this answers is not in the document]
calherries: Yep, makes sense. At metabase we generally go up to 120 chars, don't be afraid to use it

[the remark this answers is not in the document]
piranha: It's not about fear, it was me who put 80 chars limit in place, takes time to unlearn old habits. 😁

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
calherries: Maybe you can use `mt/with-temp` for this. Just create a database, table and a field before calling `mt/dataset` to be safe

piranha: Results are not reassuring. :) So if I leave test in its current form, ids are 21 (source) vs 7 (dest). If I create db+table+field, it because 21 vs 24, which is closer and makes me uneasy. :\ I'm thinking of living it in place with some comment that if it flakes, here's the discussion.

calherries (approved):
Nice!

Issue #35915, opened 2023-11-17
[Epic] White-labeling: Customizing or hiding “Help” link


npretto:
**Links**
- [product doc](https://www.notion.so/metabase/White-labeling-Customizing-or-hiding-Help-link-8840fbdfd3c743ccaff9301a835ec81c?pvs=4)
- feature branch: `white-label-hel-link-feature-branch ` ([pr](https://github.com/metabase/metabase/pull/36172))
- [Testing plan](https://github.com/metabase/metabase/issues/36250)

# Milestone 1 - basic functionality
```[tasklist]
- [ ] https://github.com/metabase/metabase/pull/35960
- [ ] https://github.com/metabase/metabase/pull/36208
- [ ] https://github.com/metabase/metabase/pull/36478
- [ ] https://github.com/metabase/metabase/pull/36480
- [x] https://github.com/metabase/metabase/pull/36328 (add settings on the backend with validation) / (add the `help-link` setting to the anonymous instance analytics)(see [product doc](https://www.notion.so/metabase/White-labeling-Customizing-or-hiding-Help-link-8840fbdfd3c743ccaff9301a835ec81c?pvs=4#4315376092b84191a5c339d27ffdbb80))
```

Issue #34920, opened 2023-10-20
Application name from appearance settings erroneously used for embedding license explanation


albertoperdomo:
### Describe the bug

<img width="504" alt="image" src="https://github.com/metabase/metabase/assets/24216/0e9b1295-cc46-4faa-a00a-ad70e714b2c5">
<img width="738" alt="image" src="https://github.com/metabase/metabase/assets/24216/fe66ab93-5d91-4780-beba-d5bbb1c1b040">

### To Reproduce

1. Admin Settings/Appearance - Set application name
2. Admin Settings/Embedding (needs to be off)
4. Click on "more details"
5. Text contains the application name where it should be Metabase, including `Powered by APPLICATION NAME` instead of `Powered by Metabase`

### Expected behavior

We don't replace the Metabase here.

### Information about your Metabase installation

```JSON
- v1.47.4
```

### Severity

Medium

iethree:
Currently we do a global replace on the "metabase" string for white-labeling. you have to exclude a string from `ttag` to exclude it.

https://metaboat.slack.com/archives/C505ZNNH4/p1682978941068939
https://metaboat.slack.com/archives/C057T1QTB3L/p1692711434412609?thread_ts=1692711380.023879&cid=C057T1QTB3L

WiNloSt:
@albertoperdomo This looks somewhat to be in conflict with [With Whitelabel custom name, Metabase name is still shown multiple places in Admin Panel](https://github.com/metabase/metabase/issues/17043#top).

In that issue, we try to replace a bunch of **Metabase** occurrences with the whitelabeled application name, but we seem to want to do the opposite here. Should there be some rules on how we should apply the whitelabeled name replacement so it's easier to understand when and where we should and should not do so? e.g. We should replace the application name everywhere in the app for sure, but I'm not sure where in admin settings we want to replace and where we don't want to.

I want to get this clear before I start working on a draft solution since it would mean redefining the foundation of the code.

WiNloSt:
This is [the Slack conversation](https://metaboat.slack.com/archives/C505ZNNH4/p1700480382374239) regarding the desired implementation

PR #35960, opened 2023-11-20
White labeling option for help link

npretto:
[🛠️ Epic](https://github.com/metabase/metabase/issues/35915)
[📝 Product Doc](https://www.notion.so/metabase/White-labeling-Customizing-or-hiding-Help-link-8840fbdfd3c743ccaff9301a835ec81c?pvs=4)

### Description

This implements the basic functionality of whitelabeling for the help link of M1.

I left a few comments with open points or doubts on the PR.

## How to verify
- start metabase ee
- login as admin
- go to admin settings, appearance
- play around with the settings

## Demo

https://github.com/metabase/metabase/assets/1914270/1bf9bb92-3802-4bd4-a351-de5ade1643c1


enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </Stack>
+    </Radio.Group>
+  );
+};
+
+/**
+ * Wait for `condition` to return truthy and returns its value
+ * Checks every `interval` ms and times out after `timeout` ms
+ */
+const waitFor = <T,>(
```
npretto, line 42: If we keep this, we should probably move it somewhere else

src/metabase/public_settings.clj
```diff
@@ -113,6 +113,20 @@
   :visibility :public
   :audit      :getter)
 
+
+  (defsetting help-link
+  (deferred-tru "TODO")
```
npretto, line 118: I have no idea what those things do besides allowing us to send the new settings, we'll need to get some help from BE

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Radio, Stack } from "metabase/ui";
+
+interface Props {
+  setting: { value: string };
+  onChange: (value: string) => void;
+}
+
+export const HelpLinkRadio = ({ setting, onChange }: Props) => {
+  const focusUrlInput = useCallback(async () => {
+    // The url input is rendered only after the setting is saved and returned from the BE
```
npretto, line 12: In theory, we could store some state very up in the component tree or in redux, but it would pollute other components and in this case I think 2 lines of procedural code are cleaner than having a one-off state shared across components just to focus something.

We could also make that getHidden actually hides via css instead of not rendering the settings, but in that case the item would get focused while hidden during the network request.
https://github.com/metabase/metabase/blob/b38ad76666a91915a2f050969ac796f70e63625a/frontend/src/metabase/admin/settings/app/components/SettingsEditor/SettingsSection.tsx#L25-L28

Overall, I'm okay this solution with waitFor but I understand some people may not like it

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+    // there's no easy way to autofocus it from here besides waiting.
+    // `autofocus` can't be used otherwise it will be autofocused also on page load
+    const input = await waitFor(() =>
```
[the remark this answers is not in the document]
npretto: (and doesn't have the timeout)

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
WiNloSt: Please remove all of this autofocus implementation and put `autoFocus: true` in the settings instead, since we already have that supported in the settings e.g.
https://github.com/metabase/metabase/blob/08480aac9416ea03ba5ae525f67b27da2ea54112/enterprise/frontend/src/metabase-enterprise/auth/index.js#L88-L95

However I found out that it doesn't work at the moment. I can patch that at the moment [here](https://github.com/metabase/metabase/blob/08480aac9416ea03ba5ae525f67b27da2ea54112/frontend/src/metabase/admin/settings/components/SettingsSetting.jsx#L58C1-L62)

```ts
    const widgetProps = {
      ...setting.getProps?.(setting, settingValues),
      ...setting.props,
      ...this.props,
      ...(setting.autoFocus ? { autoFocus: true } : {}),
    };
```

but it'd be great to ask other engineers about this since I'm not sure since when the `autoFocus` in the settings has stop working.

@ranquild Do you know how we should fix `autoFocus` not working for our settings?

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
WiNloSt: as stated above this should be removed.

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ),
+    );
+    input?.focus();
+  }, []);
+
+  return (
+    <Radio.Group value={setting.value} onChange={onChange}>
+      <Stack>
+        <Radio
+          label={t`Link to` + " Metabase " + t`help`}
```
WiNloSt, line 27: I'd leave this as 
```ts
t`Link to Metabase help`
```

as [Application name from appearance settings erroneously used for embedding license explanation](https://github.com/metabase/metabase/issues/34920#top) will be worked on soon possibly within this week. And I'm worried we'll forget to fix this afterward.

Or you could leave this as is but should there be a comment that this is a temporary bandage that should be fixed soon?

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,62 @@
+import { t } from "ttag";
+import { useCallback } from "react";
+import { Radio, Stack } from "metabase/ui";
+
+interface Props {
+  setting: { value: string };
```
WiNloSt, line 6: what do you think about typing this as an enum? Since the there are only 3 options here.

https://github.com/metabase/metabase/blob/08480aac9416ea03ba5ae525f67b27da2ea54112/frontend/src/metabase-types/api/settings.ts#L190

frontend/src/metabase/nav/components/ProfileLink/ProfileLink.jsx
```diff
@@ -47,6 +53,15 @@ function ProfileLink({ user, adminItems, onLogout }) {
       JSON.stringify(bugReportDetails),
     );
 
+    const helpLink = getHelpLink({
```
WiNloSt, line 56: What do you think about putting some of the hooks inside `getHelpLink` and converting it to a React hook? Because 3 of the props here:
- helpLinkSetting
- helpLinkCustomDestinationSetting
- isPaidPlan

Aren't used anywhere and they're just passed to `getHelpLink`

This looks a bit hard to read when a function receives so many arguments.

frontend/src/metabase/nav/components/ProfileLink/ProfileLink.jsx
```diff
@@ -60,14 +75,10 @@ function ProfileLink({ user, adminItems, onLogout }) {
         link: "/admin",
         event: `Navbar;Profile Dropdown;Enter Admin`,
       },
-      {
+      helpLink.visible && {
```
WiNloSt, line 78: I'm on the fence about this. I feel like we should just inline `helpLinkSetting !== "hidden"` here. And `getHelpLink` or `useGetHelpLink` could just return the link instead.

Maybe that's more readable because we don't need to understand that function (or hook) return an object which has 2 keys `visible` and `href`.  What do you think?

frontend/src/metabase/nav/components/ProfileLink/utils.ts
```diff
@@ -0,0 +1,27 @@
+import type { Settings } from "metabase-types/api";
+
+export const getHelpLink = ({
```
WiNloSt, line 3: I think maybe this function could be converted to a React hook and move some of the values we get from hook and pass to this function into this function's body instead to help reduce the number of parameters passed to this function.

src/metabase/public_settings.clj
```diff
@@ -113,6 +113,20 @@
...
+  :default    ""
+  :type       :string
```
WiNloSt, line 120: From looking at other settings that is also an option, they use `:keyword` e.g.
https://github.com/metabase/metabase/blob/08480aac9416ea03ba5ae525f67b27da2ea54112/src/metabase/public_settings.clj#L397-L403

```suggestion
  :type       :keyword
```

WiNloSt (commented):
The overall shape looks okay.

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
npretto: @WiNloSt doesn't that also autofocus it also on page load? I only need to autofocus it when "custom" gets selected in the previous setting.
If I open the SAML setup page, the url input gets autofocused and the page scrolls down to it


https://github.com/metabase/metabase/assets/1914270/f71af340-f80b-4f74-8819-af3e64f3786a

And I can't focus it without waitFor because it's not rendered yet, and even if we switched from not rendering it, to hiding it via css (which would require some refactor) it would focus a hidden element until the settings are refreshed, which i don't think it's a good idea

WiNloSt: Yeah, that's the option for all settings widget whether we want it to be autofocus or not. I feel like conditional autoFocus would be too complex. However, I can see how it could be jarring for the settings page to have the cursor jump to the very bottom input. 

In the case of your video, we should maybe probably remove the autoFocus from the settings itself. But for this Epic. I'd rather not autofocus it or autofocus it all the time if conditional autofocus looks too hacky.

e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
```diff
@@ -151,6 +156,82 @@ describeEE("formatting > whitelabel", () => {
...
+
+  describe("Help link", () => {
+    beforeEach(() => {
+      cy.intercept("PUT", "/api/setting/help-link").as("putHelpLink");
+      cy.intercept("PUT", "/api/setting/help-link-custom-destination").as(
+        "getHelpLinkUrl",
+      );
+    });
+
+    it("should allow customising the help link", () => {
```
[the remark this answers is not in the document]
npretto: Good point, that was actually not working because the `:default` set in clj is only for the public setting, not the one available in the admin settings page.

Added the check here: https://github.com/metabase/metabase/pull/35960/files#diff-91071e9b59b8a7c88b8356cccf71e8eeff288d73fd1f78c4d5ea2d8904477ed0R175

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
npretto: Tomorrow I'll draft another alternative, but I'm afraid it won't be much cleaner (if at all)

WiNloSt: Also, it might be worth to confirm with the product if always autofocusing is viable. Otherwise, I think dropping the autofocus behavior in favor of a simpler code would be ideal.

src/metabase/public_settings.clj
```diff
@@ -95,7 +95,7 @@
 
 (defsetting site-name
   (deferred-tru "The name used for this instance of {0}."
-                (application-name-for-setting-descriptions))
+    (application-name-for-setting-descriptions))
```
npretto, line 98: TODO: revert all formatting 🙃

frontend/src/metabase/nav/components/ProfileLink/ProfileLink.unit.spec.tsx
```diff
@@ -0,0 +1,192 @@
+import { screen, fireEvent, waitFor } from "@testing-library/react";
```
npretto, line 1: this was converted to TS

frontend/src/metabase/nav/components/ProfileLink/ProfileLink.unit.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+      openMenu();
+
+      HOSTED_ITEMS.forEach(title => {
+        expect(screen.getByText(title)).toBeInTheDocument();
+      });
+    });
+  });
+
+  describe("help link", () => {
```
npretto, line 120: new tests start here

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,73 @@
...
+    case undefined:
+      return "metabase_default";
+    case "hidden":
+      return "hidden";
+    default:
+      return "custom";
+  }
+};
+
+export const HelpLinkRadio = ({ setting, onChange }: Props) => {
```
npretto, line 29: I will rename this in another PR as it contains more than just the radio

frontend/src/metabase/nav/components/ProfileLink/ProfileLink.jsx
npretto: Made it a hook, I also moved the test to the component as a result

npretto:
@WiNloSt I merged the two settings into one, this will make things in a future PR easier (I need to make sure that the user can't have the setting as "default" without a valid link).
That also makes it easier to have the two inputs into one setting, solving the autofocus issue.
I also move the logic into a hook as you suggested, I still kept the object as return as I think it's a bit clearer and leaks less logic, let me know what you think.


For context, in the next PR I will change the text input to a "site url widget" that allows choosing a prefix/protocol and then the rest of the url, as we do for the site url.
That component will handle validation

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </Radio.Group>
+      {textInputVisible && (
+        <InputBlurChange
+          value={textValue}
+          // this makes it autofocus only when the value wasn't originally a custom destination
+          // this prevents it to be focused on page load
+          autoFocus={getRadioValue(setting.originalValue) !== "custom"}
+          aria-label={t`Help link custom destination`}
+          placeholder={t`Enter a URL`}
+          onBlurChange={e => {
```
WiNloSt, line 64: There are a couple of issues using `onBlurChange`
1. When changing from custom URL to other options, `onBlurChange` will be called first before calling `onChange` for the other option, this could create a race condition because FE calls `PUT /api/setting/help-link` at roughly the same time.
2. In this iteration, if you set a value of the custom URL then select other option and select the custom URL again, the custom URL won't persist.
    https://www.loom.com/share/bd0e95eeb4564314b46ac932323aa4b4

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
```diff
@@ -77,6 +79,20 @@ if (hasPremiumFeature("whitelabel")) {
           options: getLoadingMessageOptions(),
           defaultValue: "doing-science",
         },
+        {
+          key: "help-link",
+          display_name: t`Help Link in the Settings menu`,
```
WiNloSt, line 84: should `settings` be lowercase?

```suggestion
          display_name: t`Help Link in the settings menu`,
```

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
```diff
@@ -77,6 +79,20 @@ if (hasPremiumFeature("whitelabel")) {
...
+          description: (
+            <p>
+              {t`The Settings menu includes a Help link that goes to `}
```
WiNloSt, line 87: This is a copy thing, but I don't think `Settings` should be capitalized. Can you ask design for confirmation?

frontend/src/metabase/nav/components/ProfileLink/ProfileLink.unit.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+            "https://www.metabase.com/help?utm_source=in-product&utm_medium=menu&utm_campaign=help&instance_version=v1",
+          );
+        });
+      });
+    });
+  });
+});
+
+const openMenu = () =>
+  fireEvent.click(screen.getByRole("img", { name: /gear/i }));
```
WiNloSt, line 192: While we don't deprecate `fireEvent`, we encourage using `userEvent` which emulates users' event so they're more realistic.

src/metabase/public_settings.clj
WiNloSt: I don't see this being addressed yet FYI.

e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
WiNloSt, line 1: Please take a look at this comment, and add a test to cover that.

src/metabase/public_settings.clj
WiNloSt: Ah yeah, that's for the new version that only have 1 settings. Gotcha.

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
npretto: I didn't find anything super clean to solve this, onblur is called before pretty puch anything.
I even tried to set use an useEffect with a dependency on the text value and that was still dispatched before the onChange on the group, even before onChangeCapture which was... really confusing.
I found on stack overflow that onMouseDown is called before so I used a ref to keep a flag to skip calling onChange: https://github.com/metabase/metabase/pull/35960/commits/a0ed7d5f298ef288d8796041d4e90cb77acc0fe3

It's not super clean but it seems to work.

Can we merge this as is and look into this after we finalise the behaviour with the validation etc? I'm afraid that would break/change some things.

Ideally I'd look into that in parallel while we wait for the review/changes of the BE code

e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
npretto: Test added ✅

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
npretto: I asked Vitorio on figma

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
npretto: The solution present now its bugged, the mouseDown event is triggered also in the whitespace between the options.
I tried to move it to the radio inputs, but unfortunately the Radio component passes down the onMouseDown only to the radio and not the label.

I'll go back to the two different settings in the backend, I don't see any easy solution.
I'll still leave the widget the same on the FE to solve the autoFocus issue

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,79 @@
+import { t } from "ttag";
+import { useState } from "react";
+import { Radio, Stack } from "metabase/ui";
+import type { HelpLinkSetting, SettingKey, Settings } from "metabase-types/api";
+import InputBlurChange from "metabase/components/InputBlurChange";
+
+interface Props {
+  setting: {
```
npretto, line 8: TODO: the settings here are manually typed, I'll draft a proper typing later and ask for review/help to the web app team later

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
npretto: Update: I reverted back to two settings, this seems to be easier to maintain in the end, there is a bit more code but we don't have to make weird hacks to make things work.

WiNloSt: Would that be okay with product? Because [from the spec](https://www.notion.so/metabase/White-labeling-Customizing-or-hiding-Help-link-8840fbdfd3c743ccaff9301a835ec81c?pvs=4#8a4d9e1b95f748bfb40bac1c97993a39) it says you only have one setting. And users could set the value through the environment variable `MB_HELP_LINK`

e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
```diff
@@ -151,6 +157,113 @@ describeEE("formatting > whitelabel", () => {
...
+      cy.signInAsNormalUser();
+      cy.visit("/");
+      openSettingsMenu();
+
+      helpLink()
+        .should("have.attr", "href")
+        .and("include", "https://www.metabase.com/help?");
+    });
+
+    it("should not create a race condition when the custom url input is selected and we click on the radio", () => {
```
WiNloSt, line 247: This wasn't quite the test for the race condition because it's missing the last step which is to select the custom URL again without changing the URL.

1. default URL by default
2. Select custom URL
3. Type URL
4. Select hide it
5. Select custom URL again <- this step was missing.

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+  onChange: (value: string) => void;
+  onChangeSetting: <TKey extends SettingKey>(
+    key: TKey,
+    value: Settings[TKey],
+  ) => void;
+  settingValues: Settings;
+}
+
+const getRadioValue = (setting: HelpLinkSetting | undefined | null) => {
```
WiNloSt, line 21: We probably don't need `getRadioValue` anymore once the `help-link` can only be one of the 3 values and we have another settings for the custom URL.

enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx
```diff
@@ -0,0 +1,79 @@
...
+      {isTextInputVisible && (
+        <InputBlurChange
+          value={customUrl}
+          // this makes it autofocus only when the value wasn't originally a custom destination
+          // this prevents it to be focused on page load
+          autoFocus={getRadioValue(setting.originalValue) !== "custom"}
+          aria-label={t`Help link custom destination`}
+          placeholder={t`Enter a URL`}
+          onBlurChange={e => {
+            setCustomUrl(e.target.value);
```
WiNloSt, line 72: Because you could set the custom URL state on blur that means the way `InputBlurChange` is implemented isn't a controlled input. We could remove the `customUrl` state and pass it directly from the prop. e.g.
```ts
  const customUrl = settingValues["help-link-custom-destination"];
```

And remove this `setCustomUrl`

```suggestion
```

frontend/src/metabase/nav/components/ProfileLink/useHelpLink.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useState } from "react";
+import { UtilApi } from "metabase/services";
+import { useSelector } from "metabase/lib/redux";
+import { getSetting, getIsPaidPlan } from "metabase/selectors/settings";
+import { getUser } from "../../../selectors/user";
+
+export const useHelpLink = (): { visible: boolean; href: string } => {
+  const helpLinkSetting = useSelector(state => getSetting(state, "help-link"));
```
WiNloSt, line 8: you reference `helpLinkType` in `enterprise/frontend/src/metabase-enterprise/whitelabel/components/HelpLinkSettings/HelpLinkRadio.tsx`. What do you think about using the same variable name then?

WiNloSt (approved):
I think this looks almost ready to be shipped, but let's cleanup stuff I've commented.

e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
npretto: Thanks, interestingly enough, I reproduced an issue even without the last step and assumed it was caused by the same thing.
I added a test case also for your case (just added the step and remove the cy.wait), let's keep them both until we are 100% about having 1 or 2 settings, as going with the merged setting could have both those issues.

PR #35962, opened 2023-11-20
[WIP] EntityPicker Prototype 🔥

### LINKED ISSUE #35828: EntityPicker Proof of Concept
Before we go too far down the road, we want to prove out our ability to provide a generic picker UI component that can build its own picker tree from callbacks

- [x] create ugly/basic entity picker ui component
- [x] create basic generic picker wrapper (with callbacks that return static data)
- [x] create basic table picker wrapper
- [x] create basic question picker wrapper

iethree:
closes #35828 

> [!IMPORTANT]
> This is a hacky early prototype. This isn't meant to be efficient or pretty or anywhere near production-ready, it is meant to prove out the general structure of this component: namely that we can have a basic item picker UI component that is used by any number of wrapper components that can provide logic for populating the picker via callbacks.

https://www.loom.com/share/e47d9c1a9eaf4625a0196ca969454e4a


ranquild (changes requested):
Too generic, hard to extend. I wouldn't take this path.

PR #35975, opened 2023-11-21
Fix click behavior link to dashboards/questions not using client router

### LINKED ISSUE #33379: Card click behavior drops interface component overrides in FullApp embedding
### Describe the bug

In interactive embedding navigating from card to another dashboard via Click Behavior causes interface component overrides to be lost. This looks like a regression of [24756](https://github.com/metabase/metabase/issues/24756).

[Screen recording](https://github.com/metabase/metabase/assets/11080638/ba8c0fd3-0d11-4868-9289-795ae05e8016)

### To Reproduce

1. Create a dashboard with a card that has Click Behavior to another dashboard (regular Dashboard, not Custom URL)
2. View via FullApp embedding, where initial iframe src has overrides (e.g. .../?top_nav=false)
3. Click the card to go to the other dashboard, after browser load, it will now show the other dashboard, but without the overrides.

### Expected behavior

Overrides should be kept on the second dashboard.

### Logs

n/a

```JSON
Metabase version: 1.46.x - 1.47.0 .
```

### Severity

This has been reported by a customer.

### Additional context

I reproduced the issue with releases between `1.46.0` and `1.47.0`. 1.45.4.3 appears to be the last release without the bug.

WiNloSt:
Closes https://github.com/metabase/metabase/issues/33379

### Description

The problem was caused by the bug where we navigate to other dashboards/questions via click behavior as clicking a link instead of using client side router.

When that happens all the embed UI parameters state in Redux are lost.

### How to verify

Follow the reproduce steps in #33379 (using `?top_nav=false`), after the navigation, the top nav should still be hidden.

#### Before the fix
https://www.loom.com/share/f51e62a01f2048ab82bc93cbf7376b50

#### After the fix
https://www.loom.com/share/87b8f62a525749d8b03d30bb04854a35

### Checklist


frontend/src/metabase/lib/dom.js
```diff
@@ -349,13 +356,35 @@ const getOrigin = url => {
...
+        ? pathname
+        : pathname.replace(
+            // Only replace the path from the start to avoid matching the string elsewhere
+            new RegExp(`^${ensureNoTrailingSlash(sitePath)}`),
+            "",
+          );
     const query = querystring.parse(search.substring(1));
-    return { pathname, search, query, hash };
+    return {
+      pathname: pathNameWithoutSubPath,
```
WiNloSt, line 370: This change is needed when navigating with client-router and Metabase is hosted on a subpath e.g. http://localhost:3000/metabase instead of just http://localhost:3000

Otherwise, the pathname would contain the Metabase subpath as well. This will cause the link to contain the subpath twice e.g. `http://localhost:3000/metabase/metabase/question/1`

frontend/src/metabase/lib/dom.js
```diff
@@ -272,21 +280,20 @@ export function open(
...
   } = {},
 ) {
-  const isOriginalUrlAbsolute = isAbsoluteUrl(url); // this does not check real "absolute" url, but if a url should be resolved from the root URL
   url = ignoreSiteUrl ? url : getWithSiteUrl(url);
 
   if (shouldOpenInBlankWindow(url, options)) {
     openInBlankWindow(url);
   } else if (isSameOrigin(url)) {
-    if (isOriginalUrlAbsolute) {
+    if (!isMetabaseUrl(url)) {
```
WiNloSt, line 293: The original condition `isOriginalUrlAbsolute` causes the problem where we want to navigate to questions or dashboards via click behavior because the `isOriginalUrlAbsolute` at this point would be `true` from the url `/question/xxx` or `/dashboard/xxx`, and we'll call `clickLink` which doesn't use client-side routing, so all state in redux store is lost. And that causes a problem when we're embedding with UI parameters e.g. `top_nav=false` because the UI would be reset back to the default value instead.
https://github.com/metabase/metabase/blob/43639a9cfcc4fc16c55bf8545e5f2d34a4c90551/frontend/src/metabase/redux/embed.ts#L7-L17

frontend/src/metabase/lib/dom.js
```diff
@@ -349,13 +356,35 @@ const getOrigin = url => {
 const getLocation = url => {
   try {
     const { pathname, search, hash } = new URL(url, window.location.origin);
+    const sitePath = getSitePath();
+    const pathNameWithoutSubPath =
+      sitePath === "/"
+        ? pathname
+        : pathname.replace(
+            // Only replace the path from the start to avoid matching the string elsewhere
+            new RegExp(`^${ensureNoTrailingSlash(sitePath)}`),
```
WiNloSt, line 365: I couldn't think of any cleaner alternative yet.

alxnddr (approved):
Looks reasonable to me, I tested a few cases with reverse proxy and it works as expected. Could you please add more specs?

ranquild (commented):
@WiNloSt there is `isSameOrSiteUrlOrigin` helper used by some core components. Can it be used instead?

WiNloSt:
> @WiNloSt there is `isSameOrSiteUrlOrigin` helper used by some core components. Can it be used instead?

@ranquild I already use it here
https://github.com/metabase/metabase/pull/35975/files#diff-ba55d7a2d16f4b3fac4549de2b6b7f2732bffc418a32e70be8f47cac357d69b0R230

The problem with **only** relying on `isSameOrSiteUrlOrigin` is that it doesn't take into account the path where Metabase is deployed. And to know if a given URL is a Metabase URL or not in case Metabase is deployed on subpath e.g. `http://localhost:3000/metabase/` is to also match the path because in this case for example, `http://localhost:3000/admin` could have been an admin for the main site, not Metabase's admin.

So I don't know any other way to get around this than to parse the subpath from `site-url` settings.

Hope that makes sense.

frontend/src/metabase/lib/dom.js
ranquild: The change in unclear. Do native `react-router` location objects also exclude `subpath`?

frontend/src/metabase/lib/dom.js
ranquild: This will get busted if the string contains regex-sensitive characters. You could do the following:
```
const pathNameWithoutSubpath = pathName.startsWith(subPath) ? pathName.replace(subPath, "") : pathName;
```

ranquild (commented):
Location changes are unclear to me. May indicate something wrong in this PR. Asked some questions in comments.

frontend/src/metabase/lib/dom.js
WiNloSt: @ranquild Apparently we use `basename` , in this case it's `"/metabase"`


https://github.com/metabase/metabase/blob/e0e0646201eeb5fd1d97a4b3121a2fe77b1f00d3/frontend/src/metabase/app.js#L60C19-L65

That's why the `pathname` returned from this function shouldn't contain the subpath `/metabase` again.

ranquild: Great investigation, it's clear now

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -203,13 +204,27 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
...
+
       clickLineChartPoint();
       cy.get("@targetDashboardId").then(targetDashboardId => {
         cy.location().should(({ pathname, search }) => {
           expect(pathname).to.equal(`/dashboard/${targetDashboardId}`);
           expect(search).to.equal("");
         });
       });
+
+      cy.log("Should navigate to question using router (metabase#33379)");
```
WiNloSt, line 222: Piggyback assertions that test the navigation have been done using router.

If you check out `master` and run this test. The test should fail.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -586,6 +608,13 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
         expect(card.dataset_query.query).to.deep.equal(TARGET_QUESTION.query);
       });
 
+      cy.log("Should navigate to question using router (metabase#33379)");
```
WiNloSt, line 611: Piggyback assertions that test the navigation have been done using router.

If you check out `master` and run this test. The test should fail.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -1431,6 +1460,52 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       clickLineChartPoint();
     });
 
+    it("allows opening custom URL destination that is not a Metabase instance URL using link (metabase#33379)", () => {
```
WiNloSt, line 1463: It's virtually impossible to test metabase running in subpath, I've tried a few approaches, but didn't work out.

This test however tests that our new code that matches subpaths works. It tests that even if the origin is the same but the pathnames are different, Metabase should navigate using a link (refresh page).

frontend/src/metabase/lib/dom.js
```diff
@@ -220,7 +220,15 @@ export function constrainToScreen(element, direction, padding) {
   return false;
 }
 
-const isAbsoluteUrl = url => url.startsWith("/");
+function getSitePath() {
+  return new URL(MetabaseSettings.get("site-url")).pathname.toLowerCase();
```
WiNloSt, line 224: I need to ensure the pathname is lower case so we could compare it incase users accidentally set it using capital letters.

WiNloSt:
@alxnddr @ranquild I added some E2E tests, but it's not the actual subpath test since I'm not sure how to test that in E2E. However, I've retested the following scenario manually.
- Metabase running at `http://localhost:3001/metabase`
- Interactive embedding running in `http://localhost:3002` with iframe referencing `http://localhost:3001/metabase`

The navigation on click behavior was done with router and the embed parameter is preserved.

deniskaber (approved):

WiNloSt:
@deniskaber Please take a look at the last 3 commits where I add more tests and fix all E2E test failures

frontend/src/metabase/lib/dom.js
```diff
@@ -349,8 +366,21 @@ const getOrigin = url => {
 const getLocation = url => {
   try {
     const { pathname, search, hash } = new URL(url, window.location.origin);
+    const sitePath = getSitePath().split("/")[1];
```
ranquild, line 369: This is error prone and hard to read/understand. What part of the path do you need here?

WiNloSt: I need the site path without the `/`. Maybe I should just use `getSitePath().slice(1)` instead?

Since I'm only comparing the site path with the pathname that users are visiting. it's possible that `sitePath` from `site-url` settings would be `/Metabase` but users are on `http://localhost:3000/metabase` (lowercase `m`)

WiNloSt: Tested that `getSitePath().slice(1)` doesn't work when users set metabase site url with subpath ending with a slash e.g. `localhost:3000/metabase/`

getSitePath().split("/") is guaranteed to be a string because `getSitePath` will either return `"/"` or `"/subPathWithOrWithoutSlash"`.

frontend/src/metabase/lib/dom.js
```diff
@@ -220,7 +220,25 @@ export function constrainToScreen(element, direction, padding) {
   return false;
 }
 
-const isAbsoluteUrl = url => url.startsWith("/");
+function getSitePath() {
+  return new URL(MetabaseSettings.get("site-url")).pathname.toLowerCase();
+}
+
+function isMetabaseUrl(url) {
```
ranquild, line 227: @WiNloSt previously we handled invalid URLs in `getOrigin`, now functions will throw. Is it expected?

frontend/src/metabase/lib/dom.js
```diff
@@ -220,7 +220,25 @@ export function constrainToScreen(element, direction, padding) {
...
+
+  if (!isAbsoluteUrl(url)) {
+    return true;
+  }
+
+  return isSameOrSiteUrlOrigin(url) && urlPath.startsWith(getSitePath());
+}
+
+function isAbsoluteUrl(url) {
+  return ["/", "http:", "https:", "mailto:"].some(prefix =>
```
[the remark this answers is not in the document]
ranquild: https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol ?

ranquild (commented):
Please write tests and make sure this solution works with:
- invalid URLs
- subpaths with several segments, like `/this/is/sub/path`

frontend/src/metabase/lib/dom.js
WiNloSt: I can add a test for this, but seems like `getOrigin` [was introduced without `base`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) (the 2nd) option when constructing the URL, that's when it can throw error, but it seems impossible to fail to create a URL if you give a correct `base` (which we do `window.location.origin`).

frontend/src/metabase/lib/dom.js
```diff
@@ -349,8 +366,21 @@ const getOrigin = url => {
 const getLocation = url => {
   try {
     const { pathname, search, hash } = new URL(url, window.location.origin);
+    const sitePath = getSitePath().split("/")[1];
+    const [_empty, subPathOrPath, ...paths] = pathname.split("/");
```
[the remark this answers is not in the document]
WiNloSt: Makes sense.

frontend/src/metabase/lib/dom.js
WiNloSt: Added tests in https://github.com/metabase/metabase/pull/35975/commits/2a77c3041174167098a9f0e3eeebc2702e634789

frontend/src/metabase/lib/dom.js
```diff
@@ -350,12 +367,39 @@ const getLocation = url => {
...
+      search,
+      query,
+      hash,
+    };
   } catch {
     return {};
   }
 };
 
+function getPathnameWithoutSubPath(pathname) {
```
WiNloSt, line 381: @ranquild I didn't go with the approach of replacing pathname with site path because `pathname` could have different cases than the site path e.g. if we have `site-url` as `https://mysite.com/METABASE/`, but visit the URL `https://mysite.com/metabase/question/1`. We need to be able to match that the visited URL is a Metabase URL

frontend/src/metabase/lib/dom.js
```diff
@@ -350,12 +367,39 @@ const getLocation = url => {
...
+  const pathnameSections = pathname.split("/");
+  const sitePathSections = getSitePath().split("/");
```
WiNloSt, line 383: Both `pathname` and `sitePath` will always have a leading slash, so we don't have to worry about matching leading or nonleading slash.

frontend/src/metabase/visualizations/lib/action.unit.spec.ts
```diff
@@ -35,4 +37,184 @@ describe("performAction", () => {
       type: "@@router/CALL_HISTORY_METHOD",
     });
   });
+
+  describe.each([
```
WiNloSt, line 41: Permutate different kinds of site path with these dimensions:
1. Does it have sub path?
2. Does sub path has multiple levels?
3. Does the site path have a trailing slash?

WiNloSt:
@ranquild I've addressed these 2 points with [this commit](https://github.com/metabase/metabase/pull/35975/commits/2a77c3041174167098a9f0e3eeebc2702e634789)
> Please write tests and make sure this solution works with:
> 
> * invalid URLs
> * subpaths with several segments, like `/this/is/sub/path`

frontend/src/metabase/visualizations/lib/action.unit.spec.ts
```diff
@@ -35,4 +37,184 @@ describe("performAction", () => {
...
+
+    it("should redirect with invalid URL", () => {
+      MetabaseSettings.set("site-url", sitePath);
+      const action: UrlClickAction = {
+        buttonType: "horizontal",
+        name: "automatic-insights",
+        section: "auto",
+        url: jest.fn(
+          () =>
+            "invalid_protocol://localhost/metabase/auto/dashboard/adhoc/123Abc",
```
ranquild, line 189: I was thinking about completely invalid URLs, not valid ones with some unexpected protocol

ranquild (approved):
Last comment inline

frontend/src/metabase/visualizations/lib/action.unit.spec.ts
WiNloSt: Can you give me some examples?

WiNloSt:

WiNloSt:

PR #35981, opened 2023-11-21
Revert task to delete inactive tables

qnkhuat:
Per this [discussion](https://metaboat.slack.com/archives/C064EB1UE5P/p1700562306919679?thread_ts=1700478287.996609&cid=C064EB1UE5P) we're not comfortable releasing this just before the RC as it'll delete tables and leave orphaned data around.

this will revert commits from https://github.com/metabase/metabase/pull/35443 and https://github.com/metabase/metabase/pull/35693


src/metabase/models/table.clj
```diff
@@ -63,7 +63,7 @@
 
 (t2/define-before-delete :model/Table
   [{:keys [db_id schema id]}]
-  (t2/delete! Permissions :object [:like (str "%" (perms/data-perms-path db_id schema id) "%")]))
```
qnkhuat, line 66: this is a meaningful change, but I'll create a new PR for this to make the history clear

calherries (approved):

PR #35984, opened 2023-11-21
[E2E] Re-enable custom mapping null test

nemanjaglumac:
This test was quarantined as flaky, but I found it to be actually wrong.

This PR fixes the test and unskips it.
The stress-test workflow confirms that it's not flaking.

- 1 x 20 ✅ https://github.com/metabase/metabase/actions/runs/6943906753
- 1 x 40 ✅ https://github.com/metabase/metabase/actions/runs/6943975267

### DIFF
--- a/e2e/test/scenarios/admin/datamodel/field.cy.spec.js
+++ b/e2e/test/scenarios/admin/datamodel/field.cy.spec.js
@@ -16,7 +16,6 @@ import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
 
 const { ORDERS, ORDERS_ID } = SAMPLE_DATABASE;
 
-// [quarantine] - intermittently failing, possibly due to a "flickering" element (re-rendering)
 describe("scenarios > admin > datamodel > field", () => {
   beforeEach(() => {
     restore();
--- a/e2e/test/scenarios/admin/datamodel/field.cy.spec.js
+++ b/e2e/test/scenarios/admin/datamodel/field.cy.spec.js
@@ -160,9 +160,9 @@ describe("scenarios > admin > datamodel > field", () => {
             .click();
           popover().findByText("Custom mapping").click();
 
-          cy.findByDisplayValue("null").clear().type(remappedNullValue).blur();
+          cy.findByDisplayValue("null").clear().type(remappedNullValue);
           cy.button("Save").click();
-          cy.button("Saved!");
+          cy.button("Saved!").should("be.visible");
 
           cy.log("Make sure custom mapping appears in QB");
           openTable({ database: dbId, table: NUMBER_WITH_NULLS_ID });
--- a/e2e/test/scenarios/admin/datamodel/field.cy.spec.js
+++ b/e2e/test/scenarios/admin/datamodel/field.cy.spec.js
@@ -4,7 +4,7 @@ import {
   visitAlias,
   popover,
   resetTestTable,
-  startNewQuestion,
+  openTable,
   resyncDatabase,
 } from "e2e/support/helpers";
 import {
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/admin/datamodel/field.cy.spec.js
```diff
@@ -16,7 +16,6 @@ import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
 
 const { ORDERS, ORDERS_ID } = SAMPLE_DATABASE;
 
-// [quarantine] - intermittently failing, possibly due to a "flickering" element (re-rendering)
```
nemanjaglumac, line 19: Unrelated to this particular test.
Just a misleading comment.

e2e/test/scenarios/admin/datamodel/field.cy.spec.js
```diff
@@ -134,39 +133,42 @@ describe("scenarios > admin > datamodel > field", () => {
...
+
       restore("withSqlite");
       cy.signInAsAdmin();
-      const dbId = 2;
+
       withDatabase(
         dbId,
-        ({ number_with_nulls: { num }, number_with_nulls_ID }) =>
-          cy.visit(
-            `/admin/datamodel/database/${dbId}/schema/${SAMPLE_DB_SCHEMA_ID}/table/${number_with_nulls_ID}/field/${num}/general`,
```
nemanjaglumac, line 146: Using `SAMPLE_DB_SCHEMA_ID` was wrong. SQLite has a different schema.

e2e/test/scenarios/admin/datamodel/field.cy.spec.js
```diff
@@ -134,39 +133,42 @@ describe("scenarios > admin > datamodel > field", () => {
...
+      const dbId = 2;
+      const remappedNullValue = "nothin";
+
       restore("withSqlite");
       cy.signInAsAdmin();
-      const dbId = 2;
+
       withDatabase(
         dbId,
-        ({ number_with_nulls: { num }, number_with_nulls_ID }) =>
```
nemanjaglumac, line 144: These values were always undefined because they exist only as all caps fields.

e2e/test/scenarios/admin/datamodel/field.cy.spec.js
```diff
@@ -134,39 +133,42 @@ describe("scenarios > admin > datamodel > field", () => {
...
+          );
+
+          cy.log("Change `null` to custom mapping");
+          cy.findByRole("heading", { name: "Display values" })
+            .closest("section")
+            .findByText("Use original value")
+            .click();
+          popover().findByText("Custom mapping").click();
+
+          cy.findByDisplayValue("null").clear().type(remappedNullValue).blur();
```
WiNloSt, line 163: I think `.blur()` is unnecessary. When you click the save button in Cypress the `blur` event should be called anyway. And mostly I found we use `blur` when we're autosaving the field, which is not what we do in this case.

e2e/test/scenarios/admin/datamodel/field.cy.spec.js
```diff
@@ -134,39 +133,42 @@ describe("scenarios > admin > datamodel > field", () => {
...
+          cy.button("Save").click();
+          cy.button("Saved!");
```
WiNloSt, line 165: We should add an assertion to avoid confusion. Using `testing-library` to find an element without acting or asserting on it is discouraged.
```suggestion
          cy.button("Saved!").should("be.visible");
```

WiNloSt (approved):
I have some minor comments regarding the tests, but this works great!

e2e/test/scenarios/admin/datamodel/field.cy.spec.js
nemanjaglumac: I'm actually in a group of people who prefer implicit assertions, but I hear you.
Good call.

Thank you <3

nemanjaglumac:
> I have some minor comments regarding the tests, but this works great!

Thanks for the thorough review, as always.
Addressed the comments and ran another stress-test, just to make sure the test still passes without hiccups.
40/40 ✅ https://github.com/metabase/metabase/actions/runs/6954567753

e2e/test/scenarios/admin/datamodel/field.cy.spec.js
uladzimirdev: ➕, I'm fine with implicit assertions

WiNloSt: @uladzimirdev I really couldn't find where the discussion was, but explicit assertions have no disadvantages.

It could be ambiguous what an author intends to do with the element because there could be multiple things you could do with an element:
1. Assert that it's visible, not visible, not in the DOM, is in the DOM, etc.
2. Click it
3. type in it (if it's an input)
etc.

And I wouldn't know if there's no explicit action on an element whether you wanted to assert it or wanted to click it but forgot to add the `.click()` or something else.

In short, I still think it's worth being explicit with this.

Issue #33558, opened 2023-08-28
[MLv2] [BE] Add `drill-thru/automatic-insights`


camsaul:
Braden did a brain dump of how all this stuff works here, please read thru it: https://metaboat.slack.com/archives/C04CYTEL9N2/p1691679718619139

JS code:

https://github.com/metabase/metabase/blob/0b0edf1b916aa6a4000767ec42a41c2721d22391/frontend/src/metabase-lib/queries/drills/automatic-insights-drill.ts

1. Port JS code to Cljs
2. Fill out skeleton `drill-thru-method`
3. Make sure this comes back for the appropriate columns in `available-drill-thrus`
4. Write a million tests

camsaul:
Note that `drill-thru-method` and `available-drill-thrus` currently live in https://github.com/metabase/metabase/pull/31735 but Braden will be getting that merged into `master` shortly.

ranquild:
Proposed API:

Add `drill-thru/automatic-insights` drill returned by `availableDrillThrus` replicating the logic from this function https://github.com/metabase/metabase/blob/5f02f32332a7863548dc9fd2c87c686324ff0883/frontend/src/metabase-lib/queries/drills/automatic-insights-drill.ts#L5

`applyDrill` should should this part https://github.com/metabase/metabase/blob/5f02f32332a7863548dc9fd2c87c686324ff0883/frontend/src/metabase-lib/queries/drills/automatic-insights-drill.ts#L32 - before `ML_Urls` call

`getAutomaticDashboardUrl` https://github.com/metabase/metabase/blob/d161a9cf59e0cff7455028d4ce21b8201c653ac9/frontend/src/metabase-lib/urls.ts#L98 and `getComparisonDashboardUrl` https://github.com/metabase/metabase/blob/d161a9cf59e0cff7455028d4ce21b8201c653ac9/frontend/src/metabase-lib/urls.ts#L121 would be implemented on the FE by calling `toLegacyQuery` and then serializing relevant query parts.

PR #35991, opened 2023-11-21
 [MLv2] [FE] Automatic insights drill

ranquild:
Adds automatic insight drills following the API from the proposal https://github.com/metabase/metabase/issues/33558#issuecomment-1819476342

### DIFF
--- a/frontend/src/metabase-lib/urls.ts
+++ b/frontend/src/metabase-lib/urls.ts
@@ -1,12 +1,6 @@
 import * as Urls from "metabase/lib/urls";
 import { utf8_to_b64url } from "metabase/lib/encoding";
-import type {
-  Filter as RawFilter,
-  ParameterId,
-  ParameterValue,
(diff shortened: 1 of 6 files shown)


frontend/src/metabase-lib/urls.ts
```diff
@@ -95,55 +89,35 @@ export function getUrlWithParameters(
...
-  filters: (RawFilter | Filter[])[],
-) {
-  let cellQuery = "";
-
-  if (filters.length > 0) {
-    const mbqlFilter = filters.length > 1 ? ["and", ...filters] : filters[0];
-    cellQuery = `/cell/${utf8_to_b64url(JSON.stringify(mbqlFilter))}`;
-  }
-
+export function getAutomaticDashboardUrl(question: Question) {
```
ranquild, line 92: Make the function works with `Question` only. Do not use any deprecated `query()` calls

PR #35999, opened 2023-11-21
Add popover/embed modal display logic to the dashboard embed button

oisincoveney:
Epic: #35961 

### Description

For the embed button, if:
* User is an admin and public sharing is enabled, then the user will see `Create a public link` and the `Embed` option
* User is an admin and public sharing is disabled, the user will see the `Static Embed` modal
* User is an admin and public sharing & embedding is disabled, the user will still see the embedding modal
* User is a non-admin and a public link has been created for the dashboard, they will see the Embed button with the label of `Sharing`
* User is a non-admin and a public link has not been created for the dashboard, they will not see the button at all

**Note: the actual dropdown for creating the public link has not been made yet. This is the first step in Milestone 1, in which the next task will be to create/remove the link.**

### How to verify

_**As an admin**_
* Public Sharing enabled:
	* Should see a dropdown with `Create a public link` and `Embed` as options, where clicking `Embed` should take the user to the modal, and `Create a public link` does nothing
* Public sharing disabled:
	* Should immediately be directed to the Embed modal

_**As a non-admin**_
* Public sharing enabled & link is created:
	* you should see the Embed button with the label of `Sharing`. The popover isn't there yet.
* Public sharing disabled or link hasn't been created:
	* you shouldn't see an `Embed` button 

### Demo

https://github.com/metabase/metabase/assets/25306947/affeef71-ebfa-4f40-8c40-df06e0177bce

### Checklist


e2e/test/scenarios/visualizations-tabular/pivot_tables.cy.spec.js
```diff
@@ -675,6 +675,10 @@ describe("scenarios > visualizations > pivot tables", { tags: "@slow" }, () => {
           // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
           cy.findByText(test.subject).click();
           cy.icon("share").click();
+
```
oisincoveney, line 678: Temporary until the question embed button is modified as well

frontend/src/metabase/dashboard/components/DashboardActions.jsx
```diff
@@ -78,32 +73,7 @@ export const getDashboardActions = (
...
-            !isEditing &&
-            !isFullscreen &&
-            ((isPublicLinksEnabled && (isAdmin || dashboard.public_uuid)) ||
-              (isEmbeddingEnabled && isAdmin))
-          }
-          isLinkEnabled={canShareDashboard}
-          linkText={
-            <Tooltip
-              isLinkEnabled={canShareDashboard}
-              tooltip={
```
oisincoveney, line 96: Since the rendering is based on `canShareDashboard` I'm fairly sure this tooltip can never be exposed

frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal.jsx
```diff
@@ -49,14 +46,15 @@ class DashboardSharingEmbeddingModal extends Component {
...
+      triggerElement,
       ...props
     } = this.props;
     if (!enabled) {
       return null;
     }
     return (
-      <ModalWithTrigger
-        ref={m => (this._modal = m)}
+      <Modal
```
oisincoveney, line 57: Converting this into a normal `Modal` since it's been converted into a controlled component now

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  if (isAdmin) {
+    return "embed-menu";
+  }
+
+  if (hasPublicLink) {
+    return "public-link-popover";
+  }
+
+  return null;
```
WiNloSt, line 37: Maybe we could simplify this condition? I came up with this solution, maybe you could verify if it's still working correctly. The current condition is a bit hard to follow because we're returning `null` in 2 places:
1. `!isPublicSharingEnabled && !isAdmin`
2. `isPublicSharingEnabled && !isAdmin && !hasPublicLink`

```suggestion
  if (isAdmin) {
    return isPublicSharingEnabled ? "embed-menu" : "embed-modal";
  }

  if (isPublicSharingEnabled && hasPublicLink) {
    return "public-link-popover";
  }

  return null;
```

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
```diff
@@ -0,0 +1,116 @@
...
+};
+
+export const DashboardEmbedAction = ({
+  dashboard,
+}: {
+  dashboard: Dashboard;
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
```
WiNloSt, line 45: Here we have 3 states that control how we want to show a menu or a modal
1. `isOpen`
2. `initialClickBehavior`
3. `clickBehavior`

It seems to me these could potentially be consolidated into a single state enum e.g.
- `null` (not opening anything)
- `embed-menu`
- `public-link-popover`
- `embed-modal`

And we could use this single state to determine which UI we want to show instead of using 3. Could you try doing this? I think if this is possible it would reduce the complexity of this component and make it a lot easier to understand.

WiNloSt (commented):
This is 50% review. Since this is taking longer than expected I hope the first set of feedback could be worked on while I continue to review the rest.

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
oisincoveney: Hahahaha I spent a while on this one, since we have to play with the `isAdmin`, `isPublicSharing`, and the `null` value. My first thought was to ensure that public sharing would be the condition that's only mentioned once, only because there's three branches that depend on that condition. But I like this as well.

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
```diff
@@ -0,0 +1,116 @@
+import { useState } from "react";
+import type { Dashboard } from "metabase-types/api";
+import { DashboardEmbedHeaderButton } from "metabase/dashboard/components/DashboardEmbedHeaderButton";
+import { DashboardEmbedHeaderMenu } from "metabase/dashboard/components/DashboardEmbedHeaderMenu";
+import DashboardSharingEmbeddingModal from "metabase/dashboard/containers/DashboardSharingEmbeddingModal";
+import { useSelector } from "metabase/lib/redux";
+import { getSetting } from "metabase/selectors/settings";
+import { getUserIsAdmin } from "metabase/selectors/user";
+
+export type EmbedButtonClickBehavior =
```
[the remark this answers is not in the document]
oisincoveney: Yeah I feeling a bit weird about what to name this. The reason I didn't go with `Action` was that I was worried it would be confused with redux actions, while `ClickBehavior` would be _slightly_ more accurate since we're saying "this is what happens when you click the button." Let me think about it just a little bit more

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
```diff
@@ -0,0 +1,135 @@
...
+        setup({
+          hasPublicLink: true,
+          publicLinksEnabled: true,
+          isAdmin: true,
+        });
+
+        userEvent.click(screen.getByLabelText("share icon"));
+
+        expect(await screen.findByText("Embed")).toBeInTheDocument();
+        expect(await screen.findByText("Public link")).toBeInTheDocument();
```
WiNloSt, line 129: nit: Moving this above `Embed` would make the assertion order the same as the test before.

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
oisincoveney: I was thinking about this as well - I was going to write a whole explanation about what I'm thinking and how we can work through it, but I figured it'll be faster to just try it out, see how it feels, and then go from there 😄

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <Tooltip
+      label={
+        <Text c="inherit" size="sm" fw={700}>
+          {tooltipLabel}
+        </Text>
+      }
+      offset={8}
+    >
+      {/*Needs to be wrapped in a div so the Mantine Menu can still calculate the dropdown position */}
+      <div>
```
WiNloSt, line 30: Can you ensure this does anything? This was my test

#### with `<div>`
image

#### without `<div>`
image

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.unit.spec.tsx
WiNloSt, line 1: I'm not a fan of testing sub component that's rendered inside another component. I'd rather have one unit test that renders the root component with a bunch of variations. In this case, there wouldn't be a crazy amount of test cases if you were to permutate the options in `frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx`

The reason being that, it's not guaranteed that we pass proper parameters to the child components, so technically, it's possible that the 2 tests pass but we pass wrong parameters to the child component so we actually render the child component wrong. I'm fine testing each components separately if the number of permutated tests are beyond reasonable since it means we should abstract some layers away, but if it's still manageable I'd rather have a single test suite.

However, I'm fine leaving this as is now. But what do you think about this?

WiNloSt (commented):
Overall this looks pretty good. I have included some feedback that'd be great to have some of them addressed before this would be merged.

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
```diff
@@ -0,0 +1,135 @@
...
+      });
+    });
+
+    // TODO: Write tests for showing the popover when the user is non-admin and a public link has/hasn't been created
```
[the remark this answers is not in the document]
oisincoveney: This will be done in the `Add public link creation user flow` task within this milestone

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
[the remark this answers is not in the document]
oisincoveney: This will be done in the `Add public link creation user` flow task within this milestone

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
```diff
@@ -0,0 +1,135 @@
...
+      it("should not render the button if public sharing is enabled but no public link", () => {
+        setup({
+          hasPublicLink: false,
+          isAdmin: false,
+          publicLinksEnabled: true,
+        });
+        expect(screen.queryByLabelText("share icon")).not.toBeInTheDocument();
+      });
+
+      // TODO: Test for rendering button for non-admins when a public link is created
```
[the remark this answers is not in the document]
oisincoveney: This will be done in the `Add public link creation user` flow task within this milestone

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
```diff
@@ -0,0 +1,135 @@
...
+        "enable-public-sharing": publicLinksEnabled,
+      }),
+    }),
+  });
+};
+
+describe("DashboardEmbedHeaderButton", () => {
+  describe("whether the button should be rendered", () => {
+    describe("when the user is admin", () => {
+      it("should render the button even if there is no public link", () => {
```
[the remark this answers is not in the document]
oisincoveney: Yeah, the idea is that the tests for `DashboardEmbedHeaderButton` would handle this, and the `DashboardEmbedAction` tests would cover the rendering logic. I didn't want to have everything in one place as it would just be polluted and hard to find relevant tests

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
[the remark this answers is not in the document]
oisincoveney: Ok we solved this, we'll use `EmbedMenuModes`

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.unit.spec.tsx
oisincoveney: Yeah, I see what you mean. I like having the piece of mind that comes with having smaller components and the smaller sets of tests that come with them - I think it makes things a little easier to comprehend and debug when things inevitably do go wrong. I think having tests can be good here - for example, if the tooltip in the `DashboardEmbedHeaderButton` is rendering wrong, then we can check the sub-component to see if something weird is going on, then check the container component to see if the props are being passed wrong.

oisincoveney: I think there's space here for these tests, but of course I see where you're coming from. I think it's a good dialogue to continue having. For now, I'll add the extra tooltip tests in the `DashboardEmbedAction` component, and keep the rest for now.

WiNloSt: Sounds reasonable. Thanks for the explanation.

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
oisincoveney: I was staring at this for a while because I saw that it wasn't working before. I guess it does now 🤷

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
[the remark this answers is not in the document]
oisincoveney: I added the tests here in addition to the ones for the sub-component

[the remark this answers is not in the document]
WiNloSt: I think testing everything would be nice (like what you're adding here). At least we should assert a single path of the subcomponent, but not necessary have to permutate through every possibilities of the subcomponent states since we're already have tests for subcomponents.

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
WiNloSt, line 1: I've made some modifications to this file, let's huddle when you're online

frontend/src/metabase/dashboard/components/EmbedMenu/EmbedMenu.tsx
WiNloSt, line 1: I've made some modifications to this file, let's huddle when you're online

frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal.jsx
```diff
@@ -49,36 +44,20 @@ class DashboardSharingEmbeddingModal extends Component {
...
-                );
-              }
-            }}
-          >
-            {linkText}
-          </span>
-        }
         triggerClasses={className}
         className="scroll-y"
+        as={ModalTrigger}
```
WiNloSt, line 59: I checked that this component eventually renders https://github.com/metabase/metabase/blob/master/frontend/src/metabase/components/Modal/FullPageModal.tsx and I don't see `as` being sent anywhere.

I think this should be removable.

```suggestion
```

That also means `DashboardSharingEmbeddingModal.styled` should be removable.

frontend/src/metabase/dashboard/components/EmbedMenu/EmbedMenu.unit.spec.tsx
```diff
@@ -0,0 +1,158 @@
...
+    },
+  );
+
+  return {
+    onModalOpen,
+  };
+};
+
+describe("EmbedMenu", () => {
+  describe("which label should be rendered", () => {
```
WiNloSt, line 63: There should be an additional test for [the new label](https://www.notion.so/metabase/Setup-flow-for-static-public-embedding-embedding-feature-discoverability-3f7362e71e63440883bfebae48ab54a0#5c55501b5ef341a7a6f6522e21606e28) when both public sharing and embedding are disabled.

frontend/src/metabase/dashboard/components/EmbedMenu/EmbedMenu.tsx
```diff
@@ -0,0 +1,151 @@
...
+    getSetting(state, "enable-embedding"),
+  );
+
+  const isAdmin = useSelector(getUserIsAdmin);
+
+  if (isAdmin) {
+    if (!isEmbeddingEnabled) {
+      return "embedding-disabled";
+    }
+    return isPublicSharingEnabled ? "embed-menu" : "embed-modal";
```
WiNloSt, line 143: Maybe instead of mixing if and ternary operators, we only use ifs.

```suggestion
    if (!isEmbeddingEnabled) {
      return "embedding-disabled";
    }
    
    if (isPublicSharingEnabled) {
      return "embed-menu"
    }
    
    return "embed-modal";
```

Previously it was quite acceptable to use ternary operator because there was only one boolean flag, but now we have 2.

What do you think?

WiNloSt (approved):
I think this is mergeable now 🎉  Great work!

frontend/src/metabase/dashboard/components/EmbedMenu/EmbedMenu.tsx
oisincoveney: Yep looks good 😄

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx
```diff
@@ -0,0 +1,30 @@
...
+  dashboard: Dashboard;
+}) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  return (
+    <>
+      <EmbedMenu
+        resource={dashboard}
+        resourceType="dashboard"
+        hasPublicLink={!!dashboard.public_uuid}
```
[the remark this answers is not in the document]
oisincoveney: Yeah so this will make more sense when we have to add the public popover, since there's big differences between questions and dashboards regarding the redux actions, UUID access (`question.publicUUID()` vs `dashboard.public_uuid`), and getting the public link. It'll make more sense with the public link PR, since they'll take advantage of using the `resource` prop a little more clearly

PR #36007, opened 2023-11-21
Fix widget border in StringInputWidget, NumberInputWidget (n arity)

### LINKED ISSUE #27365: Dashboard parameter without field mapping have double borders
**Describe the bug**
When you add a dashboard parameter that will result in an input widget and don't map that filter to any card, there will be a rectangle border visible.

**To Reproduce**
1. Create a new dashboard
2. Add a "text contains" filter
3. Click the "Save" button
4. Click on the "text contains" filter

**Expected behavior**
There should be no rectangle border, only a rounded border.

**Screenshots**
image

Metabase build commit 19a9693d2bbf9d6bf8a8d06d226331de24545db1 and also happens on Stats

**Additional context**
Found the screenshot in this PR https://github.com/metabase/metabase/pull/23418 which looks exactly like the problem I found.

oleggromov:
Fixes https://github.com/metabase/metabase/issues/27365.

### Description

Originally it occurred to me that the bug is not reproducible on the current version. 
However, due to [the change in the widget type selection conditional](https://github.com/metabase/metabase/commit/16353290e6318dd6ba8cb6bdbb349646b5a4d559#diff-fb656618096d3e68aee4f568170da19f7a853e0480ae22f8b40c28d04e1c692fR252) this particular parameter became rendered as `ParameterFieldWidget` instead `StringInputWidget`, which obscured the bug.

### How to verify

1. Run storybook `yarn storybook`
2. Find `StringInputWidget` and `NumberInputWidget` with N args
3. Make sure you don't see thin grey border around them

|Before|After|
|---|---|
|Screenshot 2023-11-21 at 19 49 47|Screenshot 2023-11-21 at 19 50 08|

### Testing
I doubt such a small thing requires a test. At best, an overall visual regression test suite might be added/updated but certainly not a single border on a tiny component.

### DIFF
--- a/frontend/src/metabase/parameters/components/widgets/Widget.styled.tsx
+++ b/frontend/src/metabase/parameters/components/widgets/Widget.styled.tsx
@@ -34,5 +34,4 @@ UpdateButton.defaultProps = {
 
 export const TokenFieldWrapper = styled.div`
   margin: ${space(1)};
-  border: 1px solid ${color("border")};
 `;


oleggromov:
I think this should be backported.

uladzimirdev:
LGTM and I agree about test decision. @ranquild do you think this removal can cause any issue in other places?

uladzimirdev (approved):
I checked both StringInputWidget and NumberInputWidget with n > 1 👍

oleggromov:
> LGTM and I agree about test decision. @ranquild do you think this removal can cause any issue in other places?

I have checked and haven't found occurrences of `TokenFieldWrapper` anywhere else in the codebase. I'd even remove it entirely and collapse a whole (unnecessary) level of nesting, but didn't want to change too much in the scope of a bugfix.

ranquild (approved):

PR #35731, opened 2023-11-14
Migrate SettingsSAMLForm to formik

shaunlebron:
Checking off an item from https://github.com/metabase/metabase/issues/26178

## Summary

1. Created `FormTextarea` in `metabase/forms`, which is just a copy of `FormTextInput` with a textarea
2. Created `FormSwitch` in `metabase/forms` for using a mantine Switch inside formik.
3. Modified `FormTextInput` to take an optional `hasCopyButton` prop (to replace `CopyWidget` when using formik)
4. Modified `GroupMappingsWidget` to use `FormSwitch`, but only if used in formik— to avoid breaking SettingsLdapForm and SettingsJWTForm.
5. Added a whitelist of setting keys and specify which ones are overridable by their defaults.
6. Migrate all form fields in `SettingsSAMLForm` to our `metabase/forms` components.

## How to verify

Run EE tests for `saml.cy.spec.js` and unit tests.

To test with actual SAML, use:

* https://github.com/paoliniluis/metabase-keycloak
* https://www.metabase.com/docs/latest/people-and-groups/saml-keycloak

PR #36017, opened 2023-11-21
Fix submit button in SettingsSAMLForm

shaunlebron:
Fixing a few things left out of https://github.com/metabase/metabase/pull/35731

## Summary

1. Disable submit button when the form fields are unchanged.
2. Put some space around the submit button so it’s not flush with the error message above (or the edge of the window below).

## How to review

Disable whitespace diffs


iethree (approved):
Looks good. It's really helpful if you can include screenshots especially in simple PRs like this. People love screenshots.

Screen Shot 2023-11-21 at 4 25 16 PM

PR #36018, opened 2023-11-21
Add error handling around creating new SSO user, and fix `with-model-cleanup` macro

noahmoss:
Resolves https://github.com/metabase/metabase-private/issues/184 by adding a try/catch around new user creation which throws a more generic `ex-info` which includes just the user attributes that were passed in.

Adds a unit test for `create-new-sso-user!`.

Fixes `with-model-cleanup` — it tries to determine the max ID of a row in the model's table, so that it can delete all rows with greater IDs when the test finishes. But this assumes that all IDs in a table are sequential, and our approach for the audit DB and the internal user break this assumption. I've fixed this by adding a condition to exclude these special IDs when the macro is being used for the Database or User models.

### DIFF
--- a/test/metabase/test/util.clj
+++ b/test/metabase/test/util.clj
@@ -12,11 +12,20 @@
    [environ.core :as env]
    [java-time.api :as t]
    [mb.hawk.parallel]
+   [metabase.config :as config]
    [metabase.db.query :as mdb.query]
    [metabase.db.util :as mdb.u]
    [metabase.models
-    :refer [Card Collection Dimension Field FieldValues Permissions
-            PermissionsGroup PermissionsGroupMembership Setting Table User]]
+    :refer [Card
+            Dimension
+            Field
+            FieldValues
+            Permissions
+            PermissionsGroup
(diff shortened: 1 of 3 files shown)


test/metabase/test/util.clj
```diff
@@ -635,21 +644,50 @@
...
+  {:arglists '([model])}
+  mi/model)
+
+(defmethod with-max-model-id-additional-conditions :default
+  [_]
+  [:not= :id config/internal-mb-user-id])
+
+(defmethod with-max-model-id-additional-conditions :model/User
+  [_]
+  [:not= :id config/internal-mb-user-id])
```
escherize, line 677: Are there models besides User where we could find an id equal to `config/internal-mb-user-id`?

escherize (approved):
LGTM

test/metabase/test/util.clj
noahmoss: I don't think so. This is only an issue for tables that use `internal-mb-user-id` as the primary key, which should only be the users table.

PR #35712, opened 2023-11-14
Deduplicate babel dependencies in yarn.lock

uladzimirdev:
### Description

result of running 

```
npx yarn-deduplicate --scopes @babel
```

actually resolves multiple versions of the babel packages to the most recent 

### How to verify 

First:

```
rm -rf node_modules && yarn
```

- tests should pass
- local run should work 
- build should work
- storybook build should work
- release build should work

#### Disk space taken
together with https://github.com/metabase/metabase/pull/35659 cuts ~1700MB of node_modules 

before
image

after
image

#### Time to install node_modules

Before
Screenshot 2023-11-17 at 11 05 37

After
Screenshot 2023-11-17 at 11 05 26

PR #29628, opened 2023-03-28
[WIP] Build with SWC

iethree:
Updates our webpack build to use SWC instead of babel. I used a cool little plugin called [SMP](https://github.com/stephencookdev/speed-measure-webpack-plugin) to figure out what part of our webpack build takes so long. Also optimizes some css processing since css is the slowest part of our build.

Screen Shot 2023-03-28 at 3 32 58 PM

## Benchmarks

command  | babel | swc
--- | ---- | ---
`WEBPACK_BUNDLE=production yarn build:js` | 99s | 65s
`yarn build:js` | 46s | 16s

This is potentially dangerous. we previously tried using esbuild and it subtly broke some styling production builds: https://github.com/metabase/metabase/pull/23679

- - -
This change is [Reviewable](https://reviewable.io/reviews/metabase/metabase/29628)

PR #36022, opened 2023-11-22
Use swcMinify to speedup bundle minification

uladzimirdev:
### Description

Our build process includes building jars, which is one of the slowest parts. Webpack in the production mode takes a significant part of the time to bundle and minify code.
Modern frontend tools written in rust show substantial perf improvements, one of it is swc.
We already had an attempt to [migrate to swc loader once](https://github.com/metabase/metabase/pull/29628), it didn't work out and requires additional investigation and time.

However we can benefit from using swc in code minification. Historically Terser plugin is a default choice for webpack, it provides one of the smallest final bundle sizes.

There are several others: https://github.com/privatenumber/minification-benchmarks#-results

`esbuild` seems to be the fastest minificator, but with our app it provides much bigger final bundle (5.7Mb vs 5.3Mb with terser).

At the same time `swc` shows great speed improvement and resulted bundle size is approximately the same as with terser.

Luckily terser provides a [built in option to enable swc](https://webpack.js.org/plugins/terser-webpack-plugin/#swc) (as esbuild) for minification.

So before we migrate to Rspack to make build process way faster, swc can save some time for us.

This is the result of building jar oss - [job](https://github.com/metabase/metabase/actions/runs/6954030719/job/18920135054?pr=36022) - you can compare with results before [deduplication PR](https://github.com/metabase/metabase/pull/35712) -  [job](https://github.com/metabase/metabase/actions/runs/6745784982/job/18340327029)

### Results

| $~~~~~~~~~~~~~$ | main | static viz |
| :-: |- |-|
| before | Screenshot 2023-11-22 at 10 25 26 | Screenshot 2023-11-22 at 10 25 18 |
| after | Screenshot 2023-11-22 at 10 25 39 | Screenshot 2023-11-22 at 10 25 47 | 

### How to test

There is a little chance to break something using different minizer, so probably we can rely on e2e tests and need to verify locally, if there are any unexpected errors during prod build.

I use locally
```
NODE_ENV=production WEBPACK_BUNDLE=production NODE_OPTIONS=--max-old-space-size=8196 yarn webpack --bail
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,6 +147,7 @@
     "@storybook/manager-webpack5": "6.5.15",
     "@storybook/react": "6.5.15",
     "@svgr/webpack": "^8.0.1",
+    "@swc/core": "^1.3.99",
     "@testing-library/cypress": "^9.0.0",
     "@testing-library/dom": "^7.29.0",
     "@testing-library/jest-dom": "^5.16.5",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -199,6 +199,13 @@ const config = (module.exports = {
         },
       },
     },
+    minimizer: [
+      new TerserPlugin({
+        minify: TerserPlugin.swcMinify,
+        parallel: true,
+        test: /\.(tsx?|jsx?)($|\?)/i,
+      }),
+    ],
   },
 
   plugins: [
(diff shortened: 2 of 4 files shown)


nemanjaglumac:
I used `./bin/build/.sh` from this branch and am running Metabase just fine. Didn't notice anything broken or looking weird.
Also checked sources in dev tools, and it seems that source-maps still work correctly.

I'm just not sure whether we should wait until we ship 48.0 first.

nemanjaglumac (approved):
After some internal discussion, let's actually ship this now 🚀 

Thank you.

ranquild (approved):

alxnddr (approved):

PR #36023, opened 2023-11-22
Correct permissions needed for CSV upload

zedpelin:
Native query editing needs to be set to "Yes" for users to be able to upload CSVs (or at least that the upload icon is shown).


jeff-bruemmer (approved):
Nice catch, @zedpelin!

camsaul (approved):

alexyarosh:
Updated in https://github.com/metabase/metabase/pull/44575

PR #36025, opened 2023-11-22
Use `mysql:latest` for mysql CI job and update driver code accordingly

### LINKED ISSUE #35641: Use the official MySQL latest image for drivers tests
We've recently upgraded the oldest supported MySQL version to 8.0 in #34943.
The `circleci/mysql:latest` is two years old and is using `8.0.28-1debian10`.

The official `mysql:latest` is `8.2.0-1.el8` and is two weeks old.
We should start using the official latest image.

There was an attempt to do this in https://github.com/metabase/metabase/pull/35588, but I ran into connection issues that I will not have the time to investigate and solve.

lbrdnk:
Closes #35641

### Description

This PR updates mysql CI image to official `mysql:latest`.

While doing that, I've found out that mysql returns different type results for `YEAR(..)` function in the newer version. Rest of the changes address that -- `YEAR` type columns for mysql are handled as integers instead of a date now.

### How to verify

Check the CI. Tests are running also on `mysql:latest` now. Also test that checks handling of `YAER` values was updated.

### Checklist


bshepherdson (approved):

metamben (approved):
LGTM

PR #31796, opened 2023-06-22
Whitelabel Metabase name in admin panel (take two)

tsmacdonald:
Updated replacement for #27857 (I didn't want to clobber Cal's branch)

Fixes #17043 .

PR #36032, opened 2023-11-22
Fix whitelabeled application name replaces Metabase in admin settings

### LINKED ISSUE #34920: Application name from appearance settings erroneously used for embedding license explanation

WiNloSt:
Closes https://github.com/metabase/metabase/issues/34920
[Slack convo](https://metaboat.slack.com/archives/C505ZNNH4/p1700480382374239)

### Description

We had [a discussion](https://metaboat.slack.com/archives/C01MS7DQKR6/p1700486165468699) that we'll only replace `Metabase` with user's application name (whitelabel) only outside the Metabase admin.

There are some places where we don't replace the application name:
1. Metabase admin
2. Setup flow: because the user who set up Metabase would already know they're using Metabase.
3. Powered by Metabase in embedding
4. ~[Database prompt banner:](https://www.notion.so/metabase/Really-encourage-admins-to-add-a-DB-9c8c3a7d08ca45cb8672a154220f443e) This should only appear when someone first set up a Metabase instance. They would know they're using Metabase.~ Actually @npretto has confirmed [we're not showing this when white labeling](https://github.com/metabase/metabase/blob/e09323b6d61b53533c7b0403ea5de6d71f02ffaa/frontend/src/metabase/nav/components/WhatsNewNotification/utils.ts#L23-L25) 🙏 .
5. [New Metabase version notification:](https://www.notion.so/metabase/Provide-a-heads-up-when-Metabase-is-upgraded-with-what-s-new-44a28b55624f48e28dc3f616ce7cda0f) because it'd be weird to try to tell your application has a new version when it will link to Metabase website anyway.
6. Pulse: because it's deprecated and there's no way to access the UI unless you enter it in the URL directly.

### How to verify
1. Run Enterprise version and set up the application
1. Check all places in the app and the string `Metabase` should be replaced with the user application name

#### EE with token
Screenshot 2023-11-22 at 9 16 58 PM

### Checklist


enterprise/frontend/src/metabase-enterprise/settings/selectors.ts
```diff
@@ -18,6 +19,10 @@ export const getLogoUrl = (state: EnterpriseState) =>
 export const getLoadingMessage = (state: EnterpriseState) =>
   LOADING_MESSAGE_BY_SETTING[getSetting(state, "loading-message")];
 
-const DEFAULT_APPLICATION_NAME = "Metabase";
-export const getIsWhiteLabeling = (states: EnterpriseState) =>
-  getSetting(states, "application-name") !== DEFAULT_APPLICATION_NAME;
+const DEFAULT_APPLICATION_NAME = t`Metabase`;
```
npretto, line 22: Should't this always be "Metabase"?

WiNloSt: Possibly I did that because I saw we translate `Metabase` in some places, but I forgot we must have done that because we want to replace it with the user app name (whitelabel) because we hack that with `ttag`

WiNloSt: Fixed in https://github.com/metabase/metabase/pull/36032/commits/141e4813e11602fdaa60bbb6923000a9c7d89941

ranquild (approved):
@WiNloSt the approach looks simple and clean, I like it. Do I get it correctly that the plan is to use this selector in our places where the application name is expected to be changed?

WiNloSt:
> @WiNloSt the approach looks simple and clean, I like it. Do I get it correctly that the plan is to use this selector in our places where the application name is expected to be changed?

@ranquild correct. This approach is to use the selector deliberately in places we want to replace the application name. Mostly in the app (minus the Powered by Metabase in embedding). We would not want to replace `Metabase` inside Metabase settings, although some of the name replacements [Whitelabel Metabase name in admin panel (take two)](https://github.com/metabase/metabase/pull/31796#top), that should be fixed another time.

enterprise/frontend/src/metabase-enterprise/settings/selectors.ts
```diff
@@ -19,5 +19,9 @@ export const getLoadingMessage = (state: EnterpriseState) =>
   LOADING_MESSAGE_BY_SETTING[getSetting(state, "loading-message")];
 
 const DEFAULT_APPLICATION_NAME = "Metabase";
-export const getIsWhiteLabeling = (states: EnterpriseState) =>
-  getSetting(states, "application-name") !== DEFAULT_APPLICATION_NAME;
+export const getIsWhiteLabeling = (state: EnterpriseState) =>
+  getApplicationName(state) !== DEFAULT_APPLICATION_NAME;
+
+export function getApplicationName(state: EnterpriseState) {
```
WiNloSt, line 25: We're going to use this selector to replace the application name deliberately rather than relying on `ttag` hack.

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
```diff
@@ -103,12 +101,11 @@ if (hasPremiumFeature("whitelabel")) {
     MetabaseSettings.on("application-colors", updateColors);
   });
 
-  enabledApplicationNameReplacement();
```
WiNloSt, line 106: Remove the logic to replace application name via ttag

frontend/src/metabase/plugins/index.ts
```diff
@@ -128,6 +128,8 @@ export const PLUGIN_SELECTORS = {
   canWhitelabel: (_state: State) => false,
   getLoadingMessage: (_state: State) => t`Doing science...`,
   getIsWhiteLabeling: (_state: State) => false,
+  // Don't need to wrap this in `ttag` because we're not using ttag to change whitelabel name anymore
+  getApplicationName: (_state: State) => "Metabase",
```
WiNloSt, line 131: Don't need to wrap this in `ttag` because we're not using ttag to change whitelabel name anymore

frontend/src/metabase/public/components/EmbedFrame/LogoBadge.tsx
```diff
@@ -13,7 +13,8 @@ function LogoBadge({
   const logoSize = variant === "large" ? 42 : 28;
   const Metabase = (
     <MetabaseName key="metabase" isDark={dark} variant={variant}>
-      {t`Metabase`}
+      {/* We don't want to apply whitelabeled name to powered by Metabase banner */}
+      Metabase
```
WiNloSt, line 16: We don't want to apply whitelabeled name to powered by Metabase banner

frontend/src/metabase/routes-public.jsx
```diff
@@ -7,15 +6,19 @@ import PublicApp from "metabase/public/containers/PublicApp";
...
-export const getRoutes = store => (
-  <Route title={t`Metabase`}>
-    <Route path="public" component={PublicApp}>
-      <Route path="action/:uuid" component={PublicAction} />
-      <Route path="question/:uuid" component={PublicQuestion} />
-      <Route path="dashboard/:uuid(/:tabSlug)" component={PublicDashboard} />
+export const getRoutes = store => {
+  const applicationName = getApplicationName(store.getState());
+  return (
+    <Route title={applicationName}>
```
WiNloSt, line 14: image

frontend/src/metabase/routes.jsx
```diff
@@ -92,246 +92,254 @@ import {
   IsAuthenticated,
   IsNotAuthenticated,
 } from "./route-guards";
+import { getApplicationName } from "./selectors/whitelabel";
+
+export const getRoutes = store => {
+  const applicationName = getApplicationName(store.getState());
+  return (
+    <Route title={applicationName} component={App}>
```
WiNloSt, line 100: image

frontend/src/metabase/sharing/components/NewPulseSidebar.tsx
```diff
@@ -101,7 +104,8 @@ export function NewPulseSidebar({
                   </Link>
                 )}.`}
               {slackConfigured &&
-                t`Pick a channel and a schedule, and Metabase will do the rest.`}
+                // Screenshot 2023-12-07 at 5.44.51PM
+                t`Pick a channel and a schedule, and ${applicationName} will do the rest.`}
```
WiNloSt, line 107: Screenshot 2023-12-07 at 5 44 51 PM

frontend/src/metabase/sharing/components/SlackChannelField/SlackChannelField.tsx
```diff
@@ -68,7 +71,8 @@ const SlackChannelField = ({
         onChange={handleChange}
       />
       {hasPrivateChannelWarning && (
-        <div className="mt1">{t`In order to send subscriptions and alerts to private Slack channels, you must first add the Metabase bot to them.`}</div>
+        // Screenshot 2023-12-07 at 5.48.08PM
+        <div className="mt1">{t`In order to send subscriptions and alerts to private Slack channels, you must first add the ${applicationName} bot to them.`}</div>
```
WiNloSt, line 74: Screenshot 2023-12-07 at 5 48 08 PM

frontend/src/metabase/timelines/collections/components/TimelineEmptyState/TimelineEmptyState.tsx
```diff
@@ -66,8 +69,9 @@ const TimelineEmptyState = ({
         </EmptyStateThread>
         <EmptyStateMessage>
           {canWrite
-            ? t`Add events to Metabase to show important milestones, launches, or anything else, right alongside your data.`
-            : t`Events in Metabase let you see important milestones, launches, or anything else, right alongside your data.`}
+            ? // Screenshot 2023-12-07 at 5.57.06PM
+              t`Add events to ${applicationName} to show important milestones, launches, or anything else, right alongside your data.`
```
WiNloSt: Screenshot 2023-12-07 at 5 57 06 PM

frontend/src/metabase/timelines/questions/components/TimelineEmptyState/TimelineEmptyState.tsx
```diff
@@ -22,13 +24,15 @@ const TimelineEmptyState = ({
...
+  const applicationName = useSelector(getApplicationName);
   return (
     <EmptyStateRoot>
       <EmptyStateIcon name="star" />
       <EmptyStateText>
         {canWrite
-          ? t`Add events to Metabase to show helpful context alongside your data.`
-          : t`Events in Metabase let you see helpful context alongside your data.`}
+          ? // Screenshot 2023-12-07 at 5.59.06PM
+            t`Add events to ${applicationName} to show helpful context alongside your data.`
```
WiNloSt: Screenshot 2023-12-07 at 5 59 06 PM

frontend/src/metabase/public/components/widgets/SharingPane.tsx
```diff
@@ -116,7 +120,8 @@ export default function SharingPane({
         }
       >
         <PublicLinkHeader>{t`Public link`}</PublicLinkHeader>
-        <Description className="mb1">{t`Share this ${resourceType} with people who don't have a Metabase account using the URL below:`}</Description>
+        {/* Screenshot 2023-12-07 at 5.18.45PM */}
+        <Description className="mb1">{t`Share this ${resourceType} with people who don't have a ${applicationName} account using the URL below:`}</Description>
```
WiNloSt, line 123: Screenshot 2023-12-07 at 5 18 45 PM

frontend/src/metabase-lib/expressions/helper-text-strings.ts
```diff
@@ -822,6 +822,7 @@ const HELPER_TEXT_STRINGS: HelpTextConfig[] = [
       },
       {
         name: t`mode`,
+        // TODO: This is the only place that's not easy to replace the application name.
```
WiNloSt, line 825: Screenshot 2023-12-07 at 6 06 05 PM

frontend/src/metabase/timelines/questions/components/TimelineEmptyState/TimelineEmptyState.tsx
```diff
@@ -22,13 +24,14 @@ const TimelineEmptyState = ({
...
 
+  const applicationName = useSelector(getApplicationName);
   return (
     <EmptyStateRoot>
       <EmptyStateIcon name="star" />
       <EmptyStateText>
         {canWrite
-          ? t`Add events to Metabase to show helpful context alongside your data.`
-          : t`Events in Metabase let you see helpful context alongside your data.`}
+          ? t`Add events to ${applicationName} to show helpful context alongside your data.`
```
WiNloSt, line 33: Screenshot 2023-12-07 at 5 59 06 PM

frontend/src/metabase/account/notifications/components/HelpModal/HelpModal.unit.spec.js
```diff
@@ -1,12 +1,16 @@
-import { render, screen } from "@testing-library/react";
 import { mockSettings } from "__support__/settings";
+import { renderWithProviders, screen } from "__support__/ui";
 import HelpModal from "./HelpModal";
 
+function setup({ adminEmail, onClose } = {}) {
+  renderWithProviders(<HelpModal adminEmail={adminEmail} onClose={onClose} />);
```
WiNloSt, line 6: you'll see other tests have very similar changes as well, the point is we change some components to get the application name from Redux store, and their respective unit tests still rely on using `render` which doesn't include Redux, so I need to change from `render` to `renderWithProviders`

frontend/src/metabase-lib/expressions/helper-text-strings.ts
npretto: Considering the edge case (not being in a component), should we use the deprecated ` MetabaseSettings.get("application-name")`?
It's deprecated, but it should work fine for now

In alternative we could export the store and do `getSetting(store.getState,'application-name')` but it doesn't seem we currently export the store so it's maybe worth doing in another PR

npretto (approved):
Left a suggestion for the TODO, but LGTM.
This seems to have been a massive work, thanks for having done that 👏

frontend/src/metabase-lib/expressions/helper-text-strings.ts
WiNloSt: I don't want to introduce anything that uses `MetabaseSettings` since we're migrating away from it. I know the migration won't be completed any time soon. But by keep using it, it really doesn't help.

Also consider that this is only a single place I found so far I think it's worth leaving this as it probably won't affect a lot of users at scale. And for simplicity in our code, I don't think it's worth trying to juggle things around just to make this one place have a proper application name replacement.

Maybe @albertoperdomo could chime in and give product's perspective. But from engineering's point of view, it's not worth the effort trying to fix this one place.

What do you guys think?

WiNloSt: If we ever decided to fix this. I can introduce it in another PR since I'd rather have this one merged now rather than later to avoid merge conflicts.

PR #23043, opened 2022-05-30
fix dashboard parameters tablet layout

alxnddr:
Closes https://github.com/metabase/metabase/issues/22996

## Changes

Input parameters should have full width only on sufficiently small screens.

### Before
Screenshot 2022-05-30 at 21 18 20

### After
Screenshot 2022-05-30 at 21 14 23

PR #36047, opened 2023-11-22
Update parameter input styles

### LINKED ISSUE #24988: "Enter a default valu" on required filter values in SQL questions
**Describe the bug**
Cosmetic issue on the required filters

**To Reproduce**
1) create a sql question like "select * from people where email = {{flter}}"
2) on the bar of the right click on "required"
3) see the issue

**Expected behavior**
See the full text

**Screenshots**
image

when you click, you get one less letter
image

- Your browser and the version: Brave latest
- Your operating system: Pop 22.04
- Your databases: Postgres
- Metabase version: 44.1
- Metabase internal database: Postgres

**Severity**
Cosmetic

**Additional context**
NA

oleggromov:
Fixes https://github.com/metabase/metabase/issues/24988

### Description
The problem mentioned in the issue was with cutting text short in the parameter inputs. 
This was happening due to the `input`s being rendered within a flexbox container without explicitly set width, which was later changed to width in `px` and eventually "fixed" the bug by making inputs wide enough to fit the label.

However, this wasn't reliable and is unnecessarily verbose. And actually still had a bug that you see below.

This quick patch addresses a few issues:
- make inputs take all _available_ space within the flex container
- removes border-radius that cuts the caret

||Before|After|
|---|---|---|
|Text cut short, border-radius|Screenshot 2023-11-22 at 16 23 10|Screenshot 2023-11-22 at 16 26 07|

I have also noticed there was an [issue with smaller screen widths](https://github.com/metabase/metabase/pull/23043) and checked that all is good there as well:

Super narrow | Narrow
---|---
Screenshot 2023-11-22 at 16 32 41| Screenshot 2023-11-22 at 16 32 35

Of course it's not an ultimate solution. We would rather use components from the UI kit to build filters out of them.

### How to verify
1. Create an SQL question, add some parameters: `select * from people where email = {{filter}} and id in {{ids}}`
2. Click Settings, toggle the "Required switch"

Screenshot 2023-11-22 at 16 27 15

To check the dashboard, open a dashboard, add a bunch of filters and shrink the screen size to ~600px and lower.

### Tests
Unless some visual regression tests already exist, I don't believe any new ones should be added.

### DIFF
--- a/frontend/src/metabase/parameters/components/ParameterValueWidget.css
+++ b/frontend/src/metabase/parameters/components/ParameterValueWidget.css
@@ -37,20 +37,27 @@
   border-color: var(--color-brand);
 }
 
+:local(.parameter) input {
+  flex-basis: 142px;
+  flex-grow: 1;
+  border-radius: 0;
+}
+
 :local(.parameter.noPopover) input {
   width: 100%;
   font-size: 1em;
   font-weight: 600;
   border: none;
   background: none;
-}
 
-:local(.parameter.noPopover.isEditing) input {
-  width: 142px;
+  @media screen and (min-width: 440px) {
+    & {
+      padding: 0;
+    }
+  }
 }
 
 :local(.parameter.noPopover.selected) input {
-  width: 142px;
   font-weight: bold;
   color: var(--color-brand);
 }
@@ -58,7 +65,6 @@
 :local(.parameter.noPopover) input:focus {
   outline: none;
   color: var(--color-text-dark);
-  width: 142px;
 }
 :local(.parameter.noPopover) input::-webkit-input-placeholder {
   color: var(--color-text-medium);
@@ -119,12 +125,4 @@
 :local(.editNameIconContainer) > svg {
   position: absolute;
   top: -6px;
-}
-
-@media screen and (min-width: 440px) {
-  :local(.parameter.noPopover) input {
-    /* NOTE: Fixed with to circumvent issues with flexbox with container having a min-width */
-    width: 115px;
-    padding: 0;
-  }
 }
--- a/frontend/src/metabase/parameters/components/ParameterValueWidget.css
+++ b/frontend/src/metabase/parameters/components/ParameterValueWidget.css
@@ -38,7 +38,6 @@
 }
 
 :local(.parameter) input {
-  flex-basis: 142px;
   flex-grow: 1;
   border-radius: 0;
 }
--- a/frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
+++ b/frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
@@ -111,7 +111,6 @@ class ParameterValueWidget extends Component {
           ref={this.trigger}
           className={cx(S.parameter, S.noPopover, className, {
             [S.selected]: hasValue,
-            [S.isEditing]: isEditing,
           })}
         >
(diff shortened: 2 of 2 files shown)


oleggromov:
It's not such a trivial fix so marked it as *no backport*

frontend/src/metabase/parameters/components/ParameterValueWidget.css
```diff
@@ -37,28 +37,34 @@
   border-color: var(--color-brand);
 }
 
+:local(.parameter) input {
+  flex-basis: 142px;
```
kamilmielnik, line 41: Do we need this?

```suggestion
```

frontend/src/metabase/parameters/components/ParameterValueWidget.css
```diff
@@ -37,28 +37,34 @@
...
   background: none;
-}
 
-:local(.parameter.noPopover.isEditing) input {
-  width: 142px;
+  @media screen and (min-width: 440px) {
+    & {
+      padding: 0;
+    }
+  }
```
kamilmielnik, line 57: I don't understand why this is only applied on `min-width: 440px` - [other inputs in the sidebar don't have this](https://github.com/metabase/metabase/assets/6830683/ed661e2e-cbfa-4ed7-b742-5edf23ebf4d4).


```suggestion
  padding: 0;
```

kamilmielnik (commented):
> Of course it's not an ultimate solution. We would rather use components from the UI kit to build filters out of them.

Good observation. It'd also help with the default parameter value input inconsistent styling (now it has 2px border unlike other inputs in the sidebar, and its height is lower too). Any reason not to do it?

ranquild:
@kamilmielnik we currently don't have multi-select component in `metabase/ui` - we need to wrap `MultiSelect` from mantine first. I think a simple css fix is fine for now

frontend/src/metabase/parameters/components/ParameterValueWidget.css
ranquild: Tried it out locally and didn't notice any differences with and without `flex-basis`. @oleggromov could you double check? If it's not needed let's drop it

frontend/src/metabase/parameters/components/ParameterValueWidget.css
ranquild: It looks weird without `padding: 0` - I think it should be the default value without any media query.

Screenshot 2023-11-23 at 12 33 15

ranquild (approved):
@oleggromov Minor comments, please take a look

frontend/src/metabase/parameters/components/ParameterValueWidget.css
oleggromov: Yeah, agreed. We either need `min-width` or explicit `flex-shrink: 0` for that to work, or remove it. I will remove.

frontend/src/metabase/parameters/components/ParameterValueWidget.css
oleggromov: I have not. But given that `ParameterValueWidget` is used in `TagEditorParam`, which I haven't tested (and frankly I think it's outside the scope of this PR), I will leave this part untouched.

oleggromov:
Addressed the comments. Will merge as soon as it's green!

Issue #26178, opened 2022-10-31
[Epic] Migrate entity forms to formik


ranquild:
Currently our forms mostly use an internal form framework. The goal is to migrate them to react components and use formik directly.

**User forms**
- [x] `NewsletterForm` https://github.com/metabase/metabase/pull/26192
- [x] `ForgotPassword` https://github.com/metabase/metabase/pull/26204
- [x] `ResetPassword`  https://github.com/metabase/metabase/pull/26214
- [x] `UserStep` https://github.com/metabase/metabase/pull/26215/
- [x] `DatabaseStep` https://github.com/metabase/metabase/pull/26216
- [x] `UserPasswordForm` https://github.com/metabase/metabase/pull/26228
- [x] `UserProfileForm` https://github.com/metabase/metabase/pull/26231
- [x] #35605

**Timeline forms**
- [x] `NewTimelineModal`, `EditTimelineModal` https://github.com/metabase/metabase/pull/26247
- [x] `NewEventModal`, `EditEventModal` https://github.com/metabase/metabase/pull/26264

**Collections**
- [x] `CollectionCreateForm` #26820

**Dashboard forms**
- [x] `CreateDashboardModal` #26770

**Snippet forms**
- [x] `SnippetModal` #26928
- [x] `SnippetCollectionModal` #26930

**Data app forms**
- [x] `ActionCreator` @iethree 
- [x] ~`DataAppNavBarContainer`~ (no longer exists)

**Setting forms**

There is some overlap with our settings "framework".
- [x] `SlackSetupForm`, `SlackStatusForm` https://github.com/metabase/metabase/pull/26273
- [x] `SettingsGoogleForm` https://github.com/metabase/metabase/pull/26293
- [x] `SettingsSAMLForm` https://github.com/metabase/metabase/pull/35731
- [x] `SettingsJWTForm` https://github.com/metabase/metabase/pull/35969
- [x] `SettingsLdapForm` https://github.com/metabase/metabase/pull/36054
- [x] #41593

**Database forms**
- [x] https://github.com/metabase/metabase/pull/26766
- [x] https://github.com/metabase/metabase/pull/26818
- [x] https://github.com/metabase/metabase/pull/26923

**Query builder forms**

- [x] https://github.com/metabase/metabase/pull/30418
- [x] #41539

**Entity Form**
- [x] #35966
- [x] #41848

PR #36054, opened 2023-11-22
Migrate SettingsLdapForm to formik

shaunlebron:
Checking off one item from https://github.com/metabase/metabase/issues/26178

## Summary

1. Follows the same patterns used when migrating SettingsJWTForm: https://github.com/metabase/metabase/pull/35969
2. Unlike JWT, we had some validations to port over.  Using the FormProvider’s validationSchema for that.
3. Unlike JWT, we have to hide some fields if they are not present in the current api settings values or in the elements array computed by the selector.


frontend/src/metabase/admin/settings/components/SettingsBatchForm.jsx
shaunlebron, line 1: No longer needed since SettingsLdapForm was the last form to use it.

e2e/test/scenarios/admin-2/sso/ldap.cy.spec.js
```diff
@@ -97,16 +97,17 @@ describe(
       cy.findByText("That's not a valid port number").should("exist");
 
       enterLdapPort("21.3");
-      cy.button("Save and enable").click();
-      cy.wait("@updateLdapSettings");
       // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-      cy.findByText('For input string: "21.3"').should("exist");
+      cy.findByText("That's not a valid port number").should("exist");
```
npfitz, line 101: When we find these disable linter comments, we should see if we can give a more specific selector

frontend/src/metabase/admin/settings/components/SettingsLdapForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+const getAttributeValues = (values: SettingValues) => {
+  return Object.fromEntries(LDAP_ATTRS.map(key => [key, values[key]]));
+};
+
+const mapDispatchToProps = {
+  onSubmit: updateLdapSettings,
+};
+
+// eslint-disable-next-line import/no-default-export -- deprecated usage
+export default connect(null, mapDispatchToProps)(SettingsLdapForm);
```
npfitz, line 212: Can we name the export?

npfitz (changes requested):
Pretty minor suggestions, but otherwise it looks good!

e2e/test/scenarios/admin-2/sso/ldap.cy.spec.js
shaunlebron: 5d0a519b3d10402f7799f97fba713d9c17da459b

npfitz (approved):

PR #36055, opened 2023-11-22
[MLv2] Filter modal — Reset changes when modal is closed

kulyk:
Resets bulk filter modal's `query` state to `initialQuery` if a modal gets closed by hitting Esc or the "Close" button

### To Verfiy

1. Open the Orders table
2. Open the filter modal
3. Add any filter
4. Click Esc
5. Open the modal again
6. Ensure your changes are not applied

### DIFF
--- a/frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.tsx
+++ b/frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.tsx
@@ -91,10 +91,15 @@ export function BulkFilterModal({
     [initialQuery, query],
   );
 
+  const resetUnsavedChanges = useCallback(
+    () => setQuery(initialQuery),
+    [initialQuery],
+  );
+
   const handleClose = useCallback(() => {
-    setQuery(initialQuery); // reset unsaved changes
+    resetUnsavedChanges();
     onClose();
-  }, [initialQuery, onClose]);
+  }, [resetUnsavedChanges, onClose]);
 
   const handleSubmit = useCallback(() => {
     onSubmit(query);
--- a/frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.tsx
+++ b/frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.tsx
@@ -91,6 +91,11 @@ export function BulkFilterModal({
     [initialQuery, query],
   );
 
+  const handleClose = useCallback(() => {
+    setQuery(initialQuery); // reset unsaved changes
+    onClose();
+  }, [initialQuery, onClose]);
+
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.tsx
```diff
@@ -91,6 +91,11 @@ export function BulkFilterModal({
     [initialQuery, query],
   );
 
+  const handleClose = useCallback(() => {
```
kamilmielnik, line 94: This is a controlled component, so it'd be more appropriate to add an effect that resets unsaved changes when `opened` prop changes.

frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.unit.spec.tsx
```diff
@@ -119,6 +132,21 @@ describe("BulkFilterModal", () => {
     expect(onClose).toHaveBeenCalled();
   });
 
+  it("should reset changes on close", async () => {
+    const { onSubmit } = setup();
+
+    let createdAtShortcut = screen.getByRole("button", { name: "Today" });
+    userEvent.click(createdAtShortcut);
+    expect(createdAtShortcut).toHaveAttribute("aria-selected", "true");
```
kamilmielnik, line 140: We could also assert that the submit button is **enabled** here

frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.unit.spec.tsx
```diff
@@ -119,6 +132,21 @@ describe("BulkFilterModal", () => {
...
+
+    userEvent.click(screen.getByLabelText("Close"));
+    userEvent.click(screen.getByRole("button", { name: "Show modal" }));
+    createdAtShortcut = await screen.findByRole("button", { name: "Today" });
+    expect(createdAtShortcut).toHaveAttribute("aria-selected", "false");
+
+    expect(onSubmit).not.toHaveBeenCalled();
```
kamilmielnik, line 147: We should rather assert that the submit button is **disabled** here

frontend/src/metabase/query_builder/components/filters/BulkFilterModal/BulkFilterModal.tsx
```diff
@@ -91,6 +91,11 @@ export function BulkFilterModal({
...
+    setQuery(initialQuery); // reset unsaved changes
```
kamilmielnik, line 95: We can avoid the comment:

```tsx
  const resetUnsavedChanges = () => {
    setQuery(initialQuery);
  };
```

ranquild (approved):

PR #36062, opened 2023-11-22
Ensure testing select privilege works with and without auto commit

### LINKED ISSUE #36040: Syncs from Redshift failing since version 0.47.7
**Describe the bug**
Syncs to our Redshift database failed after upgrading to 0.47.7. After downgrading to 0.47.6 they are working again. This appears to be related to changes to autocommit settings.

**Logs**
`org.quartz.SchedulerException: Job threw an unhandled exception. [See nested exception: com.amazon.redshift.util.RedshiftException: Cannot rollback when autoCommit is enabled.]`

**To Reproduce**
1. Go to 'Admin - Databases'
2. Select a Redshift database
3. Click on 'Sync database schema now'
4. Nothing happens - new columns and tables are not added.

Same applies with scheduled updates. 

**Expected behavior**
Syncs complete and add new tables and columns.

**Screenshots**
Not applicable

**Severity**
Everything is running fine on 0.47.6 for now, but we are blocked from future updates until this is resolved. We don't have any immediate issues.

**Additional context**

**Metabase Diagnostic Info**

```json
{
  "browser-info": {
    "language": "en-GB",
    "platform": "MacIntel",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.20.1+1",
    "java.version": "11.0.20.1",
    "java.vm.version": "11.0.20.1+1",
    "os.name": "Linux",
    "os.version": "5.10.198-187.748.amzn2.x86_64",
    "user.language": "en",
    "user.timezone": "UTC"
  },
  "metabase-info": {
    "databases": [
      "bigquery-cloud-sdk",
      "redshift",
      "postgres"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "13.10"
      },
      "jdbc-driver": {
        "version": "42.5.4"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2023-10-26",
      "tag": "v0.47.6",
      "branch": "?",
      "hash": "3a7ed7a"
    },
    "settings": {
      "report-timezone": "UTC"
    }
  }
}
```

metamben:
Fixes #36040.

The bug is reported for Redshift, but it affects all JDBC based drivers.


src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -95,7 +95,8 @@
       true
       (catch Throwable e
         (log/trace e "Assuming no SELECT privileges: caught exception")
-        (.rollback conn)
+        (when-not (.getAutoCommit conn)
+          (.rollback conn))
```
snoe, line 99: I would think jdbc had a `inTransaction` method but it doesn't seem to. I believe we can't get here during  a manually opened transaction since `with-transaction` would catch and rollback on its own, but it might be worth checking.

snoe (approved):

src/metabase/driver/sql_jdbc/sync/describe_database.clj
metamben: What do you mean by "we can't get here during a manually opened transaction"? I'm unsure about both "here" and where that "manually opened transaction" could be.

If we are in a transaction (manually opened or otherwise), we want to roll back, because the transaction cannot be reused for further checks otherwise. (That's why `.rollback` was added in the first place.) If we have auto commit enabled, there is no transaction to roll back and an exception is thrown. (This is what's being fixed here.) If a transaction is started (e.g., using `with-db-transaction`), then inside that transaction `.getAutoCommit` returns `false`, so we roll back.

What should be checked?

snoe: > If a transaction is started (e.g., using with-db-transaction), then inside that transaction .getAutoCommit returns false, so we roll back.
I did not realize this happened, and thought the code was trying to prevent rolling back more than once.

```clojure
(toucan2.core/with-connection 
  [conn]
  (.setAutoCommit conn true)
  [(toucan2.core/with-transaction [tx conn]
     [(.getAutoCommit tx)
      (.getAutoCommit conn)])
   (.getAutoCommit conn)])
;; => [[false false] true]
```

test/metabase/driver/sql_jdbc/sync/describe_database_test.clj
```diff
@@ -183,3 +183,26 @@
...
+            (doseq [auto-commit [true false]]
+                (testing (pr-str {:auto-commit auto-commit :schema schema :name table-name})
+                  (sql-jdbc.execute/do-with-connection-with-options
+                   driver/*driver*
+                   (mt/db)
+                   nil
+                   (fn [^java.sql.Connection conn]
+                     (.setAutoCommit conn auto-commit)
+                     (is (false? (sql-jdbc.sync.interface/have-select-privilege?
+                                  driver/*driver* conn schema (str table-name "_should_not_exists"))))
```
rilian, line 206: is not this a typo? `_should_not_exists` -> `_should_not_exist`

PR #36067, opened 2023-11-23
Attach dashcard's viz settings to the card's :data :viz-settings key

### LINKED ISSUE #33727: Column formatting is lost when exporting a table from an embedded dashboard
### Describe the bug

Downloading a table contents from an embedded dashboard loses custom column formatting:

Original dashboard:
image

Downloaded from Metabase:
image

Downloaded from the embedded dashboard:
image

### To Reproduce

1. Create a saved question with some decimals and timestamps, apply custom formatting, save it, and add it to a new dashboard.
For the attached examples, I used a SQL question on the Sample Database:

```
with t1 as (
 select *
 FROM (
  VALUES
    (1234.05, 2345.05, 4321.05, 0.00, 1234.00, 1234.05, 0.1920, TIMESTAMP '2023-01-01 12:34:56'),
    (2345.30, 3456.30, 2931.30, 0.00, 0.00, 1234.30, 0.00, TIMESTAMP '2023-01-01 12:34:56'),
    (3456.00, 2300.00, 2250.00, 0.00, 122.00, 4250.00, 0.1158, TIMESTAMP '2023-01-01 12:34:56')
  )
), 
t2 as (
select 
    c1 as revenue,
    c2 as ar_profit,
    c3 as ap_profit,
    c4 as dc_profit,
    c5 as profit,
    c6 as cost,
    c7 as margin_ratio,
    c8 as created_at
    from t1
)
select * from t2
```
Then selected table for visualization and customized columns: set decimals to 2 places, added percentage format for `margin_ratio`, set `dc_profit` as hidden, and removed the time part for `created_at`.

2. Enable sharing, and embed the dashboard into a webapp.

3. First download the data as xlsx from the dashboard, directly from the Metabase website. See that the formatting is almost correct. (The time part of the `created_at` column is still visible, but that issue is already covered in https://github.com/metabase/metabase/issues/14393).

4. Then go to dashboard in the embedding webapp, and download the table. Note that the hidden column has reappeared and the formatting is lost.

### Expected behavior

Custom formatting should be preserved and be the same regardless whether downloaded directly from Metabase or via a dashboard embedded into another site.

### Logs

n/a

```JSON
Tested with v1.47.0 .
```

### Severity

It has been reported by a customer and we're not aware of a workaround.

### Additional context

_No response_

### LINKED ISSUE #35437: Dashboard subscription attachment does not take into account the dashboardcard visualization settings
### Describe the bug

If you hide a column from a card in the dashboard, then the subscription attachment will keep having that column while the email/slack is correct

### To Reproduce

1) build a card
2) add it to a dashboard
3) hide a column in the dashcard
4) add a subscription with the card and add the attachment
5) hit send now and see the attachment, it will still have the column you hid

### Expected behavior

we should respect the visualization settings of the card

### Logs

NA

```JSON
-47
```

### Severity

P2

### Additional context

This should come from old versions, but I can't find from when

adam-james-v:
Closes: https://github.com/metabase/metabase/issues/35437
partially fixes: https://github.com/metabase/metabase/issues/33727 (solves column order/hiding but not other formatting issues yet)

The linked issues had similarities: csv/xlsx exports did not respect column settings (like hidden columns). 

In the case of dashboard subscription attachments, there were 2 issues:
 - the dashcard viz settings were not being incorporated into the call to `metabase.query-processor.card/run-query-for-card-async`, which is ultimately run for each dashcard (excluding virtual cards)
 - even when they were merged into the card viz-settings, they weren't normalized and so downstream fns didn't incorporate the settings, because the keys weren't actually in the viz-settings (this is an opportunity for schemas + cleaning up the viz-settings throughout the codebase in future PRs)

Why merge the dashcard viz settings and card viz settings? When the query processor runs the `process-viz-settings` middleware, it is unaware of any dashcards related to the cards, and no dashcard-ids or data get passed through the qp, so the settings are unavailable unless they're merged. In this PR they get merged onto the `:viz-settings` key of the query map that gets passed through the query processor. (Another thing worth cleaning up is sensible and consistent naming (viz-settings vs. visualization_settings for example)

In the case of the embedding, there were 2 issues:
 - the middleware map was not properly being passed through the embed API to the `run-query-for-card-async` function
 - similarly to above, the viz-settings were not normalized, so column order/hidden settings were not respected

## After this PR
image

The exports/downloaded csv (and xlsx) looked like the table on the left before. After, they look like on the right. (Hidden columns are not visible and the column order set in the dashboard is also applied properly).

There are now 2 additional tests:
 - the dashboard subscription test namespace now sends an email with .csv attachment and looks at the contents of the csv to assert that not all columns are shown
 - the api/embed test namespace adds a function that similarly asserts that column order / hidden state is respected in the csv exports.

### DIFF
--- a/src/metabase/api/embed.clj
+++ b/src/metabase/api/embed.clj
@@ -267,7 +267,7 @@
 (defn dashcard-results-async
   "Return results for running the query belonging to a DashboardCard. Returns a `StreamingResponse`."
   {:style/indent 0}
-  [& {:keys [dashboard-id dashcard-id card-id export-format embedding-params token-params
+  [& {:keys [dashboard-id dashcard-id card-id export-format embedding-params token-params middleware
              query-params constraints qp-runner]
       :or   {constraints (qp.constraints/default-query-constraints)
              qp-runner   qp/process-query-and-save-execution!}}]
@@ -283,7 +283,8 @@
      :parameters    parameters
      :qp-runner     qp-runner
      :context       :embedded-dashboard
-     :constraints   constraints)))
+     :constraints   constraints
+     :middleware    middleware)))
 
 
 ;;; ------------------------------------- Other /api/embed-specific utility fns --------------------------------------
@@ -392,7 +393,7 @@
   Returns a `StreamingResponse`."
   {:style/indent 1}
   [token dashcard-id card-id export-format query-params
-   & {:keys [constraints qp-runner]
+   & {:keys [constraints qp-runner middleware]
       :or   {constraints (qp.constraints/default-query-constraints)
              qp-runner   qp/process-query-and-save-execution!}}]
   (let [unsigned-token (embed/unsign token)
@@ -407,7 +408,8 @@
       :token-params     (embed/get-in-unsigned-token-or-throw unsigned-token [:params])
       :query-params     query-params
       :constraints      constraints
-      :qp-runner        qp-runner)))
+      :qp-runner        qp-runner
+      :middleware       middleware)))
 
 (api/defendpoint GET "/dashboard/:token/dashcard/:dashcard-id/card/:card-id"
   "Fetch the results of running a Card belonging to a Dashboard using a JSON Web Token signed with the
--- a/src/metabase/query_processor/card.clj
+++ b/src/metabase/query_processor/card.clj
@@ -191,33 +191,38 @@
   Question) or `:dashboard` (from a Saved Question in a Dashboard). See [[metabase.mbql.schema/Context]] for all valid
   options."
   [card-id export-format
-   & {:keys [parameters constraints context dashboard-id middleware qp-runner run ignore_cache]
+   & {:keys [parameters constraints context dashboard-id dashcard-id middleware qp-runner run ignore_cache]
       :or   {constraints (qp.constraints/default-query-constraints)
              context     :question
              qp-runner   qp/process-query-and-save-execution!}}]
   {:pre [(int? card-id) (u/maybe? sequential? parameters)]}
-  (let [run   (or run
-                  ;; param `run` can be used to control how the query is ran, e.g. if you need to
-                  ;; customize the `context` passed to the QP
-                  (^:once fn* [query info]
-                   (qp.streaming/streaming-response [{:keys [rff context]} export-format (u/slugify (:card-name info))]
-                     (qp-runner query info rff context))))
-        card  (api/read-check (t2/select-one [Card :id :name :dataset_query :database_id :cache_ttl :collection_id
-                                              :dataset :result_metadata :visualization_settings]
-                                             :id card-id))
-        query (-> (assoc (query-for-card card parameters constraints middleware {:dashboard-id dashboard-id}) :async? true)
-                  (update :middleware (fn [middleware]
-                                        (merge
-                                         {:js-int-to-string? true :ignore-cached-results? ignore_cache}
-                                         middleware))))
-        info  (cond-> {:executed-by            api/*current-user-id*
-                       :context                context
-                       :card-id                card-id
-                       :card-name              (:name card)
-                       :dashboard-id           dashboard-id
-                       :visualization-settings (:visualization_settings card)}
-                (and (:dataset card) (seq (:result_metadata card)))
-                (assoc :metadata/dataset-metadata (:result_metadata card)))]
+  (let [run       (or run
+                      ;; param `run` can be used to control how the query is ran, e.g. if you need to
+                      ;; customize the `context` passed to the QP
+                      (^:once fn* [query info]
+                       (qp.streaming/streaming-response [{:keys [rff context]} export-format (u/slugify (:card-name info))]
+                         (qp-runner query info rff context))))
+        dash-viz  (when (not= context :question)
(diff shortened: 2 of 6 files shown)


src/metabase/api/embed.clj
```diff
@@ -283,7 +283,8 @@
      :parameters    parameters
      :qp-runner     qp-runner
      :context       :embedded-dashboard
-     :constraints   constraints)))
+     :constraints   constraints
+     :middleware    middleware)))
```
adam-james-v, line 287: Actually pass the middleware map to the query processor.

src/metabase/query_processor/card.clj
```diff
@@ -191,33 +191,38 @@
   Question) or `:dashboard` (from a Saved Question in a Dashboard). See [[metabase.mbql.schema/Context]] for all valid
   options."
   [card-id export-format
-   & {:keys [parameters constraints context dashboard-id middleware qp-runner run ignore_cache]
+   & {:keys [parameters constraints context dashboard-id dashcard-id middleware qp-runner run ignore_cache]
```
adam-james-v, line 194: pass in a dashcard-id here. May be nil. It'll be used to get and merge viz-settings from the dashcard (if there are any).

src/metabase/query_processor/card.clj
```diff
@@ -191,33 +191,38 @@
...
-                       :visualization-settings (:visualization_settings card)}
-                (and (:dataset card) (seq (:result_metadata card)))
-                (assoc :metadata/dataset-metadata (:result_metadata card)))]
+  (let [run       (or run
+                      ;; param `run` can be used to control how the query is ran, e.g. if you need to
+                      ;; customize the `context` passed to the QP
+                      (^:once fn* [query info]
+                       (qp.streaming/streaming-response [{:keys [rff context]} export-format (u/slugify (:card-name info))]
+                         (qp-runner query info rff context))))
+        dash-viz  (when (not= context :question)
```
adam-james-v, line 205: get the dashcard's viz-settings. Might be an empty map.

src/metabase/query_processor/card.clj
```diff
@@ -191,33 +191,38 @@
...
+                    (t2/select-one-fn :visualization_settings :model/DashboardCard :id dashcard-id))
+        card      (api/read-check (t2/select-one [Card :id :name :dataset_query :database_id :cache_ttl :collection_id
+                                                  :dataset :result_metadata :visualization_settings]
+                                                 :id card-id))
+        query     (-> (query-for-card card parameters constraints middleware {:dashboard-id dashboard-id})
+                      (update :viz-settings (fn [viz] (when (or (seq viz) (seq dash-viz))
```
adam-james-v, line 211: The `query` that will be run by the query processor. Has a :viz-settings key which may be nil or may have the card's viz settings. Only update the key if either card or dashcard viz settings exist and aren't empty. This is because the private fn in this namespace `viz-settings` will first look for `:viz-settings` on the query. Only if that's nil will it instead look up the card's viz settings from the qp store. So, an empty map short circuits this logic and causes other tests to fail.

markbastian: Do you need the or logic in the update clause? I think you can safely update with merge as in the following examples:
```clojure
(update {} :viz-settings merge {:abc 123})
=> {:viz-settings {:abc 123}}
(update nil :viz-settings merge {:abc 123})
=> {:viz-settings {:abc 123}}
(update {:viz-settings {:abc "FOO"}} :viz-settings merge {:abc 123})
=> {:viz-settings {:abc 123}}
(update {} :viz-settings merge nil)
=> {:viz-settings nil}
(update {:viz-settings nil} :viz-settings merge nil)
=> {:viz-settings nil}
(update {:viz-settings {}} :viz-settings merge nil)
=> {:viz-settings {}}
```

test/metabase/dashboard_subscription_test.clj
```diff
@@ -881,7 +885,55 @@
...
+(defn- metadata->field-ref
+  [{:keys [name field_ref]} enabled?]
+  {:name name :field_ref field_ref :enabled enabled?})
+
+(deftest dashboard-subscription-attachments-test
+  (testing "Dashboard subscription attachments respect dashcard viz settings."
+    (mt/with-fake-inbox
+      (mt/with-temp [Card {card-id :id :as c} (pulse.test-util/checkins-query-card {:breakout [!day.date]})
+                     Dashboard     {dash-id :id} {:name "just dash"}]
+        (let [viz {:table.columns (mapv metadata->field-ref (:result_metadata c) [true false])}]
```
markbastian, line 913: Might be worth a comment as to what these viz settings do so the reader knows what the expected applied changes are.

adam-james-v: Good call, added.

src/metabase/query_processor/card.clj
adam-james-v: The reason I did this  was for the case where the card's viz settings were not nil but the dashcard viz settings were empty {} The logic in the middleware was such that if you pass an empty map in, it won't bother to look up card viz settings from the qp store. So, tests started failing. 

I've changed the logic in the middleware and removed the (or ...) here.

markbastian (approved):
Looks good. 🚢 🇮🇹

PR #36068, opened 2023-11-23
`<datetime expression> = <date literal string>` should do something sane

### LINKED ISSUE #17807: Adding an "on" date filter without a specific time to a custom date column returns no results

camsaul:
Fixes https://github.com/metabase/metabase/issues/17807

For a filter clause like 

```clj
[:= {} [:expression "expr" {:effective-type :type/DateTime}] "2023-11-08"]
```

we were actually generating SQL like

```clj
WHERE expr = ? -- #t "2023-11-08T00:00:00-00:00"
```

which while valid at least is clearly not what people meant. This PR tweaks filter optimization logic to detect situations like this and instead output "optimized" date range filters like

```sql
WHERE expr >= ? AND expr < ?-- #t "2023-11-08T00:00:00-00:00", #t "2023-11-09T00:00:00-00:00"
```

The original version of this PR had a lot more changes and fancy optimization stuff, I ended up removing that in the end since it was too hard to reconcile it with other fixes that have happened since November, so this final version just has the parts needed to fix #17807


src/metabase/driver.clj
```diff
@@ -31,6 +32,8 @@
   `nil` (meaning subsequent queries will not attempt to change the session timezone) or something considered invalid
   by a given Database (meaning subsequent queries will fail to change the session timezone)."
   []
+  (when config/tests-available?
+    ((requiring-resolve 'mb.hawk.parallel/assert-test-is-not-parallel) `notify-all-databases-updated))
```
camsaul, line 36: We should not be doing this in parallel tests, for Snowflake it kills the open connection pool which will break other tests running at the same time.

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -104,75 +115,253 @@
...
+  (case (first literal)
+    :absolute-datetime
+    (let [[_absolute-datetime t unit] literal]
+      (when (= unit interval-unit)
+        [:absolute-datetime (u.date/add t unit interval-n) unit]))
+
+    :relative-datetime
+    (let [[_relative-datetime n unit] literal]
+      (if (= n :current)
+        [:relative-datetime interval-n interval-unit]
```
bshepherdson, line 330: I can't convince myself that this is sound in all cases, but maybe I'm thinking of it wrong.

My vague mental model is that a `:relative-datetime` defines a window of time (say a month) and the interval gives a delta like `+ 3 weeks`. In my mind adding 3 weeks to "this month" turns `Nov 1 - Nov 30, 2023` into `Nov 22 - Dec 21, 2023`.

But if I'm understanding how `:relative-datetime` works properly, this code is dropping the original window size (month) and replacing it with "3 weeks from this week". So for example, starting from Friday Nov 24 and using Monday as the first day, that rounds up to the next week (Mon Nov 27) and then adds two more, resulting in Mon Dec 11 to Sun Dec 17.

Do we want it to translate a same-sized window, or is changing the window intended? Or is there a missing `(= unit interval-unit)` check for this case?

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -104,75 +115,253 @@
...
+       (let [{:keys [non-interval intervals]} (group-temporal-arithmetic-args expr)
+             intervals                        (optimize-intervals intervals)]
+         (when (and (mbql.u/is-clause? #{:absolute-datetime :relative-datetime} non-interval)
+                    (= (count intervals) 1))
+               (let [interval (first intervals)
+                     interval (case tag
+                                :+ interval
+                                :- (let [[_interval n unit] interval]
+                                     [:interval (- n) unit]))]
+             (add-interval-if-possible non-interval interval)))))
```
bshepherdson, line 353: nit: weird indentation.

test/metabase/driver/sql/query_processor_test.clj
```diff
@@ -254,22 +255,26 @@
...
+                      [:cast
+                       [:week (h2x/with-database-type-info
+                               [:dateadd (h2x/literal "day")
+                                [:inline 1]
+                                (h2x/with-database-type-info [:cast :created_at [:raw "datetime"]] "datetime")]
+                               "datetime")]
+                       [:raw "datetime"]]
+                      "datetime")]
+                    "datetime"))
+               (sql.qp/adjust-start-of-week :h2 (fn [x] [:week x]) :created_at)))))
```
bshepherdson, line 272: Optional style/taste thing: I find `(f some-big-long-expression short-value)` hard to read, and
`(-> some-big-long-expression (f short-value))` easier.

test/metabase/query_processor/middleware/optimize_temporal_filters_test.clj
```diff
@@ -10,21 +10,26 @@
    [metabase.test :as mt]
    [metabase.util.date-2 :as u.date]))
 
-(driver/register! ::timezone-driver, :abstract? true)
+(deftest ^:parallel temporal-arithmetic?-test
+  (is (#'optimize-temporal-filters/temporal-arithmetic?
+       [:+ [:field 1 {:base-type :type/DateTime}] [:interval -2 :month]])))
```
bshepherdson, line 15: A couple of negative tests wouldn't be a bad idea.

bshepherdson (approved):

test/metabase/query_processor/middleware/optimize_temporal_filters_test.clj
```diff
@@ -10,21 +10,26 @@
...
 
-(defmethod driver/database-supports? [::timezone-driver :set-timezone] [_driver _feature _db] true)
+(deftest ^:parallel group-temporal-arithmetic-args-test
+  (is (= {:tag          :+
+          :intervals    [[:interval -2 :month]]
+          :non-interval [:field 1 {:base-type :type/DateTime}]}
+         (#'optimize-temporal-filters/group-temporal-arithmetic-args
+          [:+ [:field 1 {:base-type :type/DateTime}] [:interval -2 :month]]))))
```
metamben, line 22: Tests for multiple values and no value could be interesting.

metamben (approved):
Really nice! 🥇 I found the metadata-provider changes in this PR a bit distracting. I'm also concerned by the different unit thing Braden pointed out, but stamping anyway.

src/metabase/mbql/schema.cljc
```diff
@@ -554,14 +555,15 @@
 (def ^:private ExpressionArg
   [:ref ::ExpressionArg])
 
-(mr/def ::NumericExpressionArgOrInterval
+(mr/def ::Addable
```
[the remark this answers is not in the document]
camsaul: That's totally true. But the old name was wrong (it was actually a DateTimeExpression, NumericExpression, or an interval). What name makes more sense here?

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
camsaul: fixed

src/metabase/mbql/schema.cljc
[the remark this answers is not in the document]
metamben: Dunno, nothing better than spelling out the full name comes to my mind. Or can we name the context in which it occurs somehow?

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
camsaul: ended up removing this stuff

test/metabase/query_processor/middleware/optimize_temporal_filters_test.clj
camsaul: ended up removing this stuff

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -2,66 +2,77 @@
...
+
+(defn- interval? [x]
+  (mbql.u/is-clause? :interval x))
+
+(defn- temporal-arithmetic? [x]
+  (and (mbql.u/is-clause? #{:+ :-} x)
+       (let [[_tag & args] x]
+         (and (every? (partial mbql.u/is-clause? #{:relative-datetime :absolute-datetime :interval :field :expression})
+                      args)
+              (= (count (filter (some-fn temporal-literal? temporal-ref?) args)) 1)))))
```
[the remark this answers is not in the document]
camsaul: ended up removing this stuff

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -104,75 +115,253 @@
...
+
+(mu/defn ^:private group-temporal-arithmetic-args :- [:map
+                                                      [:tag          [:enum :+ :-]]
+                                                      [:non-interval [:or
+                                                                      mbql.s/field
+                                                                      mbql.s/expression
+                                                                      mbql.s/relative-datetime
+                                                                      mbql.s/absolute-datetime]]
+                                                      [:intervals    [:maybe [:sequential mbql.s/interval]]]]
+  [[tag & args :as _expr] :- [:or mbql.s/+ mbql.s/-]]
```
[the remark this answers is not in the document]
camsaul: ended up removing this stuff

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -104,75 +115,253 @@
...
+          [::arithmetic-wrapping-ref ::literal ::literal]
+          (let [{x' :x, lower' :y} (transfer-temporal-arithmetic-args x lower)
+                {upper' :y}        (transfer-temporal-arithmetic-args x upper)]
+            (optimize [:between x' lower' upper']))
+          #_else
+          nil)
+        [:between x lower upper])))
+
+(mu/defn ^:private combine-intervals-if-possible :- [:maybe mbql.s/interval]
+  "Combine two `:interval` clauses into one if they both have the same unit."
```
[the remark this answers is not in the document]
camsaul: ended up removing this stuff

src/metabase/query_processor/middleware/optimize_temporal_filters.clj
```diff
@@ -104,75 +115,253 @@
...
+  "Combine multiple `:interval` clauses into the fewest number possible."
+  [intervals :- [:sequential {:min 1} mbql.s/interval]]
+  (let [intervals (sort-by
+                   (fn [[_interval _n unit]]
+                     unit)
+                   intervals)]
+    (reduce
+     (fn [acc interval]
+       (if (empty? acc)
+         [interval]
```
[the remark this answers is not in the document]
camsaul: ended up removing this stuff

test/metabase/driver/sql/query_processor_test.clj
camsaul: fixed

PR #36069, opened 2023-11-23
Includes question name to the archived alert email

### LINKED ISSUE #25693: "One of your alerts has stopped working" email should provide more detail
If someone deletes a question that had alerts, an email gets triggered to the alert creator nothing than the text in the title and who deleted the question.

Add info about which is the question that was deleted to the email. We should also add to the archived list the ID's of the question as names can be confusing if there are many questions with the same name.

None

Requested by a customer

**Additional context**
NA

qnkhuat:
Fixes [#25693](https://github.com/metabase/metabase/issues/25693)

The email we currently send doesn't include the question at all, this fixes it so that the email now includes the question name along with its ID, and when click on the question name, you'll be redirected to /archive page.

before
Screenshot 2023-11-23 at 15 09 40

after([thread](https://metaboat.slack.com/archives/C01LQQ2UW03/p1701428293738339?thread_ts=1701332742.481009&cid=C01LQQ2UW03))
Screenshot 2023-12-04 at 11 11 36


src/metabase/email/messages.clj
```diff
@@ -685,8 +685,11 @@
 (defn send-alert-stopped-because-archived-email!
   "Email to notify users when a card associated to their alert has been archived"
   [alert user {:keys [first_name last_name] :as _archiver}]
-  (let [deletion-text (format "the question was archived by %s %s" first_name last_name)]
-    (send-email! user not-working-subject stopped-template (assoc (common-alert-context alert) :deletionCause deletion-text))))
+  (let [deletion-text (format "the question was archived by %s %s" first_name last_name)
```
piranha, line 688: What do you think about i18n? Or are we sending emails only in English?

test/metabase/api/card_test.clj
```diff
@@ -1517,92 +1517,95 @@
...
-               (is (= (merge (crowberto-alert-not-working {expected-email true})
-                             (rasta-alert-not-working     {expected-email true}))
-                      (mt/regex-email-bodies (re-pattern expected-email)))
-                   (format "Email containing %s should have been sent to Crowberto and Rasta" (pr-str expected-email)))))
-            (if deleted?
-              (is (= nil (t2/select-one Pulse :id (u/the-id pulse)))
-                  "Alert should have been deleted")
-              (is (not= nil (t2/select-one Pulse :id (u/the-id pulse)))
-                  "Alert should not have been deleted"))))))))
+  (doseq [{:keys [message card deleted? expected-email-rg f]}
```
piranha, line 1520: This is just a nitpick, but the usual way to shorten `regular expression` to two letters is `re`. :)

src/metabase/models/pulse.clj
```diff
@@ -222,6 +214,11 @@
   [notification-or-id]
   (t2/select PulseChannel, :pulse_id (u/the-id notification-or-id)))
 
+(def ^:dynamic *allow-hydrate-archived-cards*
```
calherries, line 217: It's a shame we need to use dynamic vars instead of passing an arg to a `t2/hydrate` call. This is okay for now but we should look into extending toucan's `hydrate` function asap so we don't need to keep doing this. This is one of the biggest reasons why we can't get rid of dynamic vars.

calherries (commented):
Looks good, I'm just wondering if it makes sense to link to the archive from the question title? If I click a link on a question title I'd expect it to link to the actual question page. We can't here, because it's archived. But it still feels surprising to me. 

Maybe something like this would be better:

Alerts about Orders (id=85) have stopped because the question was archived by Crowberto Corv [(go to archive)](url).

calherries (approved):
Nice

Issue #27199, opened 2022-12-13
Move to yarn 2+


paoliniluis:
Newer yarn version works way faster than the one we have right now (v1). We should move to Yarn v2 once shadow-cljs supports it

nemanjaglumac:
Faster for which tasks?
Can you please provide some examples and/or becnhmark results?

paoliniluis:
https://dev.to/arcanis/yarn-3-0-performances-esbuild-better-patches-e07

Tested it locally on my machine and the "yarn install" command was ~ half of what's right now. Didn't measure it though, but trust me it's faster

drewverlee:
@kamilmielnik, what features of Modern Yarn are specifically interesting to you?

paoliniluis:
Just tested yarn v4 and it mostly works, it fails only in the licencing step not being able to find the script, which should be a simple fix

PR #36076, opened 2023-11-23
Add test-unit-watch:debug script to package.json

kamilmielnik:
I needed to debug unit tests today and figured this is a useful little script to have in `package.json`.

See https://jestjs.io/docs/troubleshooting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -359,6 +359,7 @@
     "test-unit-keep-cljs": "jest --maxWorkers=2",
     "test-unit-watch": "yarn test-unit --watch",
     "test-unit-watch:cljs": "yarn concurrently -n 'cljs,tests' 'yarn build-watch:cljs' 'yarn test-unit-keep-cljs --watch'",
+    "test-unit-watch:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --watch",
     "test-visual": "yarn build && ./bin/build-for-test && yarn test-visual-run",
     "test-visual-open": "percy exec -- yarn test-cypress-open",
     "test-visual-run": "percy exec -- yarn test-cypress-run --spec \"./e2e/test/visual/**/*.cy.spec.js\"",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -359,6 +359,7 @@
     "test-unit-keep-cljs": "jest --maxWorkers=2",
     "test-unit-watch": "yarn test-unit --watch",
     "test-unit-watch:cljs": "yarn concurrently -n 'cljs,tests' 'yarn build-watch:cljs' 'yarn test-unit-keep-cljs --watch'",
+    "test-unit-watch:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --no-cache --watch",
```
oleggromov, line 362: I haven't tried it, but I am assuming you can't use just `yarn jest` because `--inspect-brk` flag is needed to be passed to node?

oleggromov: Have found this though: https://stackoverflow.com/a/48550380
If it works, might be a bit easier to read.

uladzimirdev:
I'd suggest using vscode debugging instead of `node --inspect-brk`, it might be useful if you investigate memory issues or some perf issues

package.json
kamilmielnik: We're on an [old version of yarn](https://github.com/metabase/metabase/issues/27199) and the `--inspect` flag is not available.

We could use `NODE_OPTIONS="--inspect"` but I don't think it's necessarily easier to understand/read than `node --inspect-brk path`.

uladzimirdev:
One more IMO: this is a command you'd use in every other project, so alias may be better

`alias jest-debug="node --inspect-brk ...`

I personally used this command only for debugging performance of the tests

kamilmielnik:
Related [Slack discussion](https://metaboat.slack.com/archives/C0645JP1W81/p1701092509763409).

----

I honestly didn't think this change is going to cause so much stir.
At this point I believe the most efficient thing to do is to bite the bullet and close the PR.

PR #36081, opened 2023-11-23
collection serialization should export parent collections too

### LINKED ISSUE #33900: Exporting a child collection won't export the parent collection
### Describe the bug

Let's say you have a collection which is deeply nested on other collections. If you export the child collections via --collections flag, the parent collections won't be exported, and you'll receive an error on the import.

This is pretty tricky, as now the people doing the serialization tasks will have to know the entire dependency structure when exporting a collection just to export a single collection which is nested in others

### To Reproduce

1) in a vanilla metabase use x-rays to create a dashboard
2) then export the child collection with the --collection flag
3) now import it, it will say that the parent collection is not there

### Expected behavior

should we export the entire tree when there's a --collection flag?

### Logs

_No response_

```JSON
47.x
```

### Severity

P2

### Additional context

_No response_

sansolovyov:
without exporting their content thought

resolves #33900

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
@@ -62,13 +62,13 @@
     (eduction (map #(serdes/extract-all % extract-opts)) cat (model-set opts))))
 
 ;; TODO Properly support "continue" - it should be contagious. Eg. a Dashboard with an illegal Card gets excluded too.
-(defn- traverse-with [targets traverse-fn]
-  (loop [to-traverse (set targets)
+(defn- traverse-with [nodes traverse-fn]
+  (loop [to-traverse (set nodes)
          traversed   #{}]
     (let [[m i :as item] (first to-traverse)
-          res            (traverse-fn m i)
+          found          (traverse-fn m i)
           traversed      (conj traversed item)
-          to-traverse    (set/union (disj to-traverse item) (set/difference res traversed))]
+          to-traverse    (set/union (disj to-traverse item) (set/difference found traversed))]
       (if (empty? to-traverse)
         traversed
         (recur to-traverse traversed)))))
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
@@ -62,21 +62,16 @@
     (eduction (map #(serdes/extract-all % extract-opts)) cat (model-set opts))))
 
 ;; TODO Properly support "continue" - it should be contagious. Eg. a Dashboard with an illegal Card gets excluded too.
-(defn- chase-for [targets func]
-  (loop [to-chase (set targets)
-         chased   #{}]
-    (let [[m i :as item] (first to-chase)
-          res            (func m i)
-          chased         (conj chased item)
-          to-chase       (set/union (disj to-chase item) (set/difference res chased))]
-      (if (empty? to-chase)
-        chased
-        (recur to-chase chased)))))
-
-(defn- gather-lineage [targets]
-  (set/union
-   (chase-for targets serdes/ascendants)
-   (chase-for targets serdes/descendants)))
+(defn- traverse-with [targets traverse-fn]
+  (loop [to-traverse (set targets)
+         traversed   #{}]
+    (let [[m i :as item] (first to-traverse)
+          res            (traverse-fn m i)
+          traversed      (conj traversed item)
+          to-traverse    (set/union (disj to-traverse item) (set/difference res traversed))]
+      (if (empty? to-traverse)
+        traversed
+        (recur to-traverse traversed)))))
 
 (defn- escape-analysis
   "Given a target seq, explore the contents of any collections looking for \"leaks\". For example, a
@@ -174,10 +169,12 @@ Eg. if Dashboard B includes a Card A that is derived from a
     ;; If that is non-nil, emit the report.
     (escape-report analysis)
     ;; If it's nil, there are no errors, and we can proceed to do the dump.
-    (let [closure     (gather-lineage targets)
+    (let [nodes       (set/union
+                       (traverse-with targets serdes/ascendants)
+                       (traverse-with targets serdes/descendants))
           models      (model-set opts)
           ;; filter the selected models based on user options
-          by-model    (-> (group-by first closure)
+          by-model    (-> (group-by first nodes)
                           (select-keys models)
                           (update-vals #(set (map second %))))
           extract-ids (fn [[model ids]]
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
@@ -10,6 +10,7 @@
    [metabase.models :refer [Card Collection Dashboard DashboardCard]]
    [metabase.models.collection :as collection]
    [metabase.models.serialization :as serdes]
+   [metabase.util :as u]
    [metabase.util.log :as log]
    [toucan2.core :as t2]))
 
@@ -60,18 +61,6 @@
   (log/tracef "Extracting Metabase with options: %s" (pr-str opts))
   (let [extract-opts (assoc opts :collection-set (collection-set-for-user user-id))]
     (eduction (map #(serdes/extract-all % extract-opts)) cat (model-set opts))))
-
-;; TODO Properly support "continue" - it should be contagious. Eg. a Dashboard with an illegal Card gets excluded too.
-(defn- traverse-with [nodes traverse-fn]
-  (loop [to-traverse (set nodes)
-         traversed   #{}]
-    (let [[m i :as item] (first to-traverse)
-          found          (traverse-fn m i)
-          traversed      (conj traversed item)
-          to-traverse    (set/union (disj to-traverse item) (set/difference found traversed))]
-      (if (empty? to-traverse)
-        traversed
-        (recur to-traverse traversed)))))
 
 (defn- escape-analysis
   "Given a target seq, explore the contents of any collections looking for \"leaks\". For example, a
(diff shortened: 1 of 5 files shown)


enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
```diff
@@ -1660,3 +1663,38 @@
...
+                            :collection_id (:id n)}))
+  (def pc (ts/create! Card {:name "top card"
+                            :collection_id (:id p)}))
+
+  (serdes/descendants "Collection" (:id n))
+  (#'extract/descendants-closure [["Collection" (:id n)]])
+  (by-model "Collection"
+            (extract/extract {:targets       [["Collection" (:id n)]]
+                              :no-settings   true
+                              :no-data-model true})))
```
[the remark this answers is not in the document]
piranha: I did not! I guess looked like another test in diff to my eyes.

calherries (changes requested):
This works, but I'm not in love with how you've implemented this by adding options to the descendants. It introduces coupling between `descendants-closure` and `descendants` that makes this code harder to understand. And as it's currently documented, it introduces what I'd call a "logical bug", by breaking the contract that the `descendants` function provides.

Consider the docstring of `serdes/descendants`:
```
  "Returns set of `[model-name database-id]` pairs for all entities contained or used by this entity. e.g. the Dashboard
   implementation should return pairs for all DashboardCard entities it contains, etc.

   Dispatched on model-name."
```

How would you update this to accomodate for the changes here? Any explanation would need to include `"when no-descendants=true is included in the options, we don't extract the descendants of that item in [[descendants-closure]]"`.

Perhaps it's better to rethink this and separate the idea of `descendants` (we extract their ascendants and descendants) separate from `ascendants` (we extract their ascendants, and not their descendants). 

I can think of two options:
1. Introduce a separate step after [this line](https://github.com/metabase/metabase/blob/e18a610e85eaad67d6bfe2b18bb8de91bfe61b2a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj#L173), that extracts the parents of all collections in `by-model`, and not their descendants. 
2. Change `descendants-closure` (renaming it) to get the `ascendants` of items too. The `ascendants` would be chased for their `ascendants` but not their `descendants`. I think this is smarter and more extensible to more models, but perhaps harder to understand than 1.

I would be fine with either option. Both are preferable to complicating `descendants`.

I'd also like to see a test with personal collections, because they're often special for weird reasons

enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
```diff
@@ -1660,3 +1663,38 @@
                           (extract/extract {:targets       [["Collection" coll2-id]]
                                             :no-settings   true
                                             :no-data-model true})))))))))
+
+(deftest recursive-colls-test
+  (mt/with-empty-h2-app-db
+    (ts/with-temp-dpc [Collection {parent-id :id
```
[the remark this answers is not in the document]
piranha: Too many macros!

piranha:
@calherries I've been pondering about that, like introducing another step, but decided against it as it felt like a special-case for collections - everything else won't trigger any ascendants since there is no nesting anywhere else.

So... okay, let me try this and see if the code is nicer.

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -66,9 +66,10 @@
   (loop [to-chase (set targets)
          chased   #{}]
     (let [[m i :as item] (first to-chase)
-          desc           (serdes/descendants m i)
-          chased         (conj chased item)
-          to-chase       (set/union (disj to-chase item) (set/difference desc chased))]
+          asc                 (serdes/ascendants m i)
+          desc                (serdes/descendants m i)
+          chased              (-> (conj chased item) (into asc))
+          to-chase            (set/union (disj to-chase item) (set/difference desc chased))]
```
calherries, line 72: This feels a little arbitrary that ascendants aren't chased but descendants are. I was imagining that ascendants would be chased for ascendants but at least their descendants wouldn't be. Anyway, this works fine because for collections getting the full path of parents is very quick and easy, so chasing is unnecessary.

The difference between ascendants and descendants should be explained in the docstring of this function at least.

src/metabase/models/serialization.clj
```diff
@@ -391,6 +391,17 @@
 (defmethod descendants :default [_ _]
   nil)
 
+(defmulti ascendants
+  "Return set of `[model-name database-id]` pairs for all entities containing this entity, required to successfully
+  load this entity in destination db.
+
+  Dispatched on model-name."
```
calherries, line 398: The wording of this docstring is insufficiently precise. I think it needs to explain that the set needs to include a full set of parent nodes up to the root, or something like that. Because a grandparent entity doesn't necessarily "contain" the entity of a grandchildren. At least not for all uses of the word "contain".

calherries:
Do we need to add a test for collections that are descended from personal collections? Genuine question. I don't know enough about how personal collections are handled by serialization to know.

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -62,17 +62,22 @@
...
-          to-chase       (set/union (disj to-chase item) (set/difference desc chased))]
+          to-chase       (set/union (disj to-chase item) (set/difference res chased))]
       (if (empty? to-chase)
         chased
         (recur to-chase chased)))))
 
+(defn- gather-lineage [targets]
+  (set/union
+   (chase-for targets serdes/ascendants)
+   (chase-for targets serdes/descendants)))
```
calherries, line 79: `lineage` is more misleading than helpful, and I can't think of a better name. I think you could just inline this in escape analysis

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -62,17 +62,22 @@
     (eduction (map #(serdes/extract-all % extract-opts)) cat (model-set opts))))
 
 ;; TODO Properly support "continue" - it should be contagious. Eg. a Dashboard with an illegal Card gets excluded too.
-(defn- descendants-closure [targets]
+(defn- chase-for [targets func]
```
calherries, line 65: awesome, this feels like the right abstraction. can you come up with a better name than `func`? This is graph traversal, so we should keep to standard names like `get-adjacent-items`

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -62,17 +62,22 @@
...
   (loop [to-chase (set targets)
          chased   #{}]
     (let [[m i :as item] (first to-chase)
-          desc           (serdes/descendants m i)
+          res            (func m i)
```
calherries, line 69: `res` could be more specific. This is graph traversal, so we are getting adjacent nodes here. I would call this `adjacents` or something

calherries (approved):

piranha:
> Should you include a test for collections that are descended from personal collections? Genuine question. I don't know enough about how personal collections are handled by serialization to know.

I'm not sure here, but there is no difference in code which looks for ascendants/descendants, so I guess there's no need to.

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -62,21 +62,16 @@
...
-          to-chase       (set/union (disj to-chase item) (set/difference res chased))]
-      (if (empty? to-chase)
-        chased
-        (recur to-chase chased)))))
-
-(defn- gather-lineage [targets]
-  (set/union
-   (chase-for targets serdes/ascendants)
-   (chase-for targets serdes/descendants)))
+(defn- traverse-with [targets traverse-fn]
```
calherries, line 65: Last idea - now that `traverse-with` is abstract and has less to do with serialization, we should rename `targets` to `nodes` to be consistent

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -62,21 +62,16 @@
...
+  (loop [to-traverse (set targets)
+         traversed   #{}]
+    (let [[m i :as item] (first to-traverse)
+          res            (traverse-fn m i)
```
calherries, line 69: `res` still can be more specific

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -62,21 +62,16 @@
...
-        (recur to-chase chased)))))
-
-(defn- gather-lineage [targets]
-  (set/union
-   (chase-for targets serdes/ascendants)
-   (chase-for targets serdes/descendants)))
+(defn- traverse-with [targets traverse-fn]
+  (loop [to-traverse (set targets)
+         traversed   #{}]
+    (let [[m i :as item] (first to-traverse)
```
calherries, line 68: It's a shame we have to destructure like this, when this function would be otherwise so decoupled from serialization specifics. Shall we change `descendants` and `ascendants` to accept a 2-vec pair, so this could just be 
```clojure
(let [item (first to-traverse)
      res  (traverse-fn item)
```

piranha: Hah, yeah, that would be nice. I even tried looking at clojure.walk as replacement, but it's too insistent on looking inside regular data structures.

I could change it and then use as `(traverse-with #(serdes/descendants (first %) (second %)))` to lower the churn. Still would be a quite universal function then.

calherries: Yeah, that would be better I think.

piranha: Now that I think about it, let's keep in here? This comment about `continue` is not giving me rest. 
:)

calherries: What's that about anyway? I don't understand it

PR #36082, opened 2023-11-23
Tidy CSV uploads

calherries:
This PR makes several refactorings to tidy up the CSV upload code. I mostly just used this as an exercise to get familiar with the code again, but these are real improvements

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -3,7 +3,6 @@
    [clj-bom.core :as bom]
    [clojure.data.csv :as csv]
    [clojure.java.io :as io]
-   [clojure.set :as set]
    [clojure.string :as str]
    [flatland.ordered.map :as ordered-map]
    [flatland.ordered.set :as ordered-set]
@@ -51,7 +50,8 @@
 
 (def ^:private type->parent
   ;; listed in depth-first order
-  {::varchar-255              ::text
+  {::text                     nil
+   ::varchar-255              ::text
    ::float                    ::varchar-255
    ::int                      ::float
    ::int-pk                   ::int
@@ -67,8 +67,7 @@
    ::int         ::int-pk})
 
 (def ^:private types
-  (set/union (set (keys type->parent))
-             (set (vals type->parent))))
+  (keys type->parent))
 
 (def ^:private pk-base-types
   (set (keys base-type->pk-type)))
@@ -77,7 +76,7 @@
   (into {} (for [type types]
              [type (loop [ret (ordered-set/ordered-set)
                           type type]
-                     (if-let [parent (type->parent type)]
+                     (if-some [parent (type->parent type)]
                        (recur (conj ret parent) parent)
                        ret))])))
 
@@ -131,9 +130,6 @@
 (defn- offset-datetime-string? [s]
   (applies-without-throwing? upload-parsing/parse-offset-datetime s))
 
-(defn- fits-in-varchar-255? [s]
-  (<= (count s) 255))
-
 (defn- boolean-string? [s]
   (boolean (re-matches #"(?i)true|t|yes|y|1|false|f|no|n|0" s)))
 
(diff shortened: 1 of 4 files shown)


src/metabase/upload.clj
```diff
@@ -68,8 +67,7 @@
    ::int         ::int-pk})
 
 (def ^:private types
-  (set/union (set (keys type->parent))
-             (set (vals type->parent))))
+  (keys type->parent))
```
calherries, line 70: now you can never have a type without a parent, unless explicitly stated

src/metabase/upload.clj
```diff
@@ -181,15 +187,21 @@
     (contains? (type->ancestors type-b) type-a) type-a
     :else (lowest-common-member (type->ancestors type-a) (type->ancestors type-b))))
 
-(defn- coalesce-types
-  [types-so-far new-types]
-  (->> (map vector types-so-far new-types)
-       (mapv (partial apply lowest-common-ancestor))))
+(defn- map-with-nils
```
calherries, line 190: Introducing this function allows us to get rid of `pad`, which we needed before because `map` finishes iterating when the first collection is exhausted.

src/metabase/upload.clj
```diff
@@ -148,23 +155,22 @@
     1. ints/floats are assumed to use the separators and decimal points corresponding to the locale defined in the
        application settings
     2. 0 and 1 are assumed to be booleans, not ints."
-  [value]
-  (let [number-separators (upload-parsing/get-number-separators)
-        trimmed-val       (str/trim value)]
+  [value {:keys [number-separators] :as _settings}]
```
calherries, line 158: I wanted this function to be pure so its performance is more predictable

src/metabase/upload/parsing.clj
```diff
@@ -119,57 +122,58 @@
 (defmulti upload-type->parser
   "Returns a function for the given `metabase.upload` type that will parse a string value (from a CSV) into a value
   suitable for insertion."
-  {:arglists '([upload-type])}
-  identity)
+  {:arglists '([upload-type settings])}
```
calherries, line 125: Like `value->type`, I wanted this function to be pure

src/metabase/upload.clj
```diff
@@ -214,35 +226,33 @@
...
 (defn- rows->schema
+  "rows should be a lazy-seq"
   [header rows]
   (let [normalized-header (->> header
                                (map normalize-column-name)
                                (mbql.u/uniquify-names)
                                (map keyword))
-        column-count      (count normalized-header)]
+        column-count      (count normalized-header)
+        settings          (upload-parsing/get-settings)]
```
calherries, line 236: Now we get `settings` here, and pass it to `row->types`, which passes it to `value->type`

src/metabase/upload.clj
piranha: Given that it's only used in a loop a few lines down, does it make sense to have this variable at all? Same question about `pk-base-types`, it would work fine as a map anyway...

sansolovyov (approved):
Makes sense to me

src/metabase/upload.clj
calherries: This is bikeshedding a little but it makes sense to me to keep this variable. For one it allows `type->parent` to be changed to a function without changing `type->ancestors`. 

The second reason is without the `types` variable, the code is objectively harder to understand and more likely to have bugs. If you wanted to understand whether `type->ancestors` contains the key `::text`, you would need to know that `type->parent` also contains the `::text` key. That seems like an unstable assumption (evidence: I just changed it). But with the `types` variable, you just need to understand that `types` contains the full set of types, which includes `::text`. That allows me to freely remove the `::text` variable from `type->parent` again, update the definition of `types`, and not have to worry about anything else breaking.

calherries: As for `pk-base-types`, the same reason applies: it allows `base-type->pk-type` to be changed to a function without breaking its users.

calherries:
Not backporting because this introduces significant conflicts for little gain.

PR #36085, opened 2023-11-23
Fix `data.cols` metadata with a double JOIN in a nested query

bshepherdson:
The source query metadata gets keyed by just the `:field` ref without
options, which is not unique when a table is included twice.

This change disambiguates the source metadata by `:source_alias`, so
that the correct metadata is copied to the outer query.

The most visible change is that `data.cols[i].source_alias` is now set
correctly, rather than all duplicates being set to the `:source_alias`
of whichever happens to be last in the source query.

### DIFF
--- a/src/metabase/query_processor/middleware/annotate.clj
+++ b/src/metabase/query_processor/middleware/annotate.clj
@@ -24,6 +24,7 @@
    [metabase.sync.analyze.fingerprint.fingerprinters :as fingerprinters]
    [metabase.util :as u]
    [metabase.util.i18n :refer [deferred-tru tru]]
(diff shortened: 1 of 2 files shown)


bshepherdson:
* master
  * **PR #36085** <a href="https://app.graphite.dev/github/pr/metabase/metabase/36085?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/36085?utm_source=stack-comment).

PR #36086, opened 2023-11-23
[MLv2] Add drill FE tests

ranquild:
Implements all drill tests using MLv2 calls. Uses the existing implementation for the reference.

How to review - read every drill test suite and verify you can understand what's going on.

- [x] `distribution`
- [x] `sort`
- [x] `summarize-column`
- [x] `summarize-column-by-time`
- [x] `column-filter`
- [x] `pk`
- [x] `zoom`
- [x] `fk-details`
- [x] `fk-filter`
- [x] `quick-filter`
- [x] `underlying-records`
- [x] `zoom-in.binning`
- [x] `zoom-in.geo`
- [x] `zoom-in.timeseries`
- [x] `pivot`
- [x] `automatic-insights`


frontend/src/metabase-lib/tests/drills-distribution.unit.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  createNotEditableQuery,
+  createOrdersCommentDatasetColumn,
+  createOrdersCommentField,
+  createOrdersDescriptionDatasetColumn,
+  createOrdersDescriptionField,
+  createOrdersSerializedJSONDatasetColumn,
+  createOrdersSerializedJSONField,
+} from "./drills-common";
+
+describe("drill-thru/distribution", () => {
```
kamilmielnik, line 31: We should have tests for legend and column header clicks.

frontend/src/metabase-lib/tests/drills-column-filter.unit.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  createQueryWithClauses,
+} from "metabase-lib/test-helpers";
+import {
+  createCountDatasetColumn,
+  createNotEditableQuery,
+  createOrdersStructuredDatasetColumn,
+  createOrdersStructuredField,
+} from "./drills-common";
+
+describe("drill-thru/column-filter", () => {
```
kamilmielnik, line 24: We should have tests for legend clicks.

frontend/src/metabase-lib/tests/drills-distribution.unit.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    const drill = queryDrillThru(
+      defaultQuery,
+      stageIndex,
+      clickObject,
+      drillType,
+    );
+    expect(drill).toBeNull();
+  });
+
+  it("should not drill thru a PK column", () => {
```
kamilmielnik, line 107: This test is duplicated in `it.each` below

frontend/src/metabase-lib/tests/drills-fk.unit.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      defaultQuery,
+      stageIndex,
+      clickObject,
+      drillType,
+    );
+    expect(drillInfo).toMatchObject({
+      type: drillType,
+      tableName: "Orders",
+      columnName: "Product ID",
+    });
```
kamilmielnik, line 60: Lots of tests call `findDrillThru` but they don't make any use of `drillInfo`.
Do you think it'd be worth adding these assertions everywhere?

frontend/src/metabase-lib/tests/drills-fk.unit.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      stageIndex,
+      clickObject,
+      drillType,
+    );
+    const newQuery = Lib.drillThru(defaultQuery, stageIndex, drill);
+    expect(Lib.filters(newQuery, stageIndex)).toHaveLength(1);
+  });
+});
+
+describe("drill-thru/fk-filter", () => {
```
kamilmielnik, line 38: We should have tests for legend and column header clicks.

frontend/src/metabase-lib/tests/drills-fk.unit.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+    const drill = queryDrillThru(
+      defaultQuery,
+      stageIndex,
+      clickObject,
+      drillType,
+    );
+    expect(drill).toBeNull();
+  });
+
+  it("should not drill thru a FK column", () => {
```
kamilmielnik, line 103: We should check PK column as well.

frontend/src/metabase-lib/tests/drills-pivot.unit.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+      ],
+    });
+    const clickObject = createAggregatedCellClickObject({
+      aggregation: {
+        column: createCountDatasetColumn(),
+        value: 10,
+      },
+      breakouts: [
+        {
+          column: createOrdersTotalDatasetColumn({
```
bshepherdson, line 201: Trying this with my pivots PR found a mistake here. Unfortunately I spent a while debugging the pivot code before realizing the test input was inconsistent.

The `breakouts` above want `CREATED_AT` and `CATEGORY` (and the test description calls from date and category breakouts so that seems right) but this is constructing the click object with a `TOTAL` column (and a value of `"2020-01-01"`). Changing this line to call `createOrdersCreatedAtDatasetColumn()` fixes it.

kamilmielnik (commented):
The test setup is wonderful :heart: 

----

IMO every test suite should have a test case for:
- column header click
- cell click
- pivot cell click
- legend item click

----

This is only a partial review: I reviewed every non-`unit.spec.ts` file and the first 5 test suites. I will continue on Monday.

frontend/src/metabase-lib/tests/drills-zoom-in-geographic.unit.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+          stageIndex,
+          clickObject,
+          drillType,
+        );
+        const newQuery = Lib.drillThru(query, stageIndex, drill);
+        expect(Lib.aggregations(newQuery, stageIndex)).toHaveLength(1);
+        expect(Lib.breakouts(newQuery, stageIndex)).toHaveLength(2);
+      });
+
+      it("should drill thru a pivot cell", () => {
```
kamilmielnik, line 202: We should add test for legend item.

kamilmielnik (commented):
This completes my review.

PR #36088, opened 2023-11-23
Uploads: Make empty values NULL for MySQL

### LINKED ISSUE #35606: Blank strings in CSV uploads should be NULL in MySQL
### Describe the bug

Blank strings in CSV uploads are converted to `NULL` values in Postgres and H2 but are converted to empty strings '' in MySQL.

### To Reproduce

Create a csv with empty strings in the rows and upload to a MySQL database, for example:

```
empty_column,empty_integer,empty_string
,1,
,    ,hello
```

Then query the uploaded table with the native SQL editor, checking `<column> IS NULL`. There should be values that return `true` for each blank value in the CSV, but they all return false.

### Expected behavior

blank strings in CSV values should be converted to NULL in the database.

_No response_

### Logs

_No response_

```JSON
master
```

### Severity

low

### Additional context

_No response_

calherries:
Fixes https://github.com/metabase/metabase/issues/35606

It also makes an improvement to the testing code. Now all CSV upload tests that write the upload data to the database run with [local_infile](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_local_infile) both set to on and off. Before most tests just ran with it on.

### DIFF
--- a/test/metabase/upload_test.clj
+++ b/test/metabase/upload_test.clj
@@ -169,15 +169,15 @@
            [" 2022-01-01T01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]
            [" 2022-01-01t01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]
            [" 2022-01-01 01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]]]
-    (let [settings {:number-separators (or seps ".,")}
-          type     (#'upload/value->type string-value settings)
-          parser   (upload-parsing/upload-type->parser type settings)]
-      (testing (format "\"%s\" is a %s" string-value type)
-        (is (= expected-type
-               type)))
-      (testing (format "\"%s\" is parsed into %s" string-value expected-value)
-        (is (= expected-value
-               (parser string-value)))))))
+    (mt/with-temporary-setting-values [custom-formatting (when seps {:type/Number {:number_separators seps}})]
+      (let [type   (upload/value->type string-value)
+            parser (#'upload-parsing/upload-type->parser type)]
+        (testing (format "\"%s\" is a %s" string-value type)
+          (is (= expected-type
+                 type)))
+        (testing (format "\"%s\" is parsed into %s" string-value expected-value)
+          (is (= expected-value
+                 (parser string-value))))))))
 
 (deftest ^:parallel type-coalescing-test
   (doseq [[type-a          type-b        expected]
--- a/test/metabase/upload_test.clj
+++ b/test/metabase/upload_test.clj
@@ -69,7 +69,7 @@
     (thunk)
     (do
       (testing "with local_infile on"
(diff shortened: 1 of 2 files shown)


test/metabase/upload_test.clj
```diff
@@ -34,30 +33,51 @@
 (def ^:private ai-int-pk-type ::upload/auto-incrementing-int-pk)
 (def ^:private string-pk-type ::upload/string-pk)
 
-(defn- do-with-mysql-local-infile-activated
-  "Helper for [[with-mysql-local-infile-activated]]"
+(defn- local-infile-on? []
```
calherries, line 36: I made these changes because I realised most of the end-to-end CSV tests were run with `with-mysql-local-infile-activated`, which sets `local_infile` to "ON" and runs the test. But we also need to test with `local_infile` "OFF". 

So I replaced the `with-mysql-local-infile-activated` macro with `with-mysql-local-infile-on-and-off`, which tests both.

test/metabase/upload_test.clj
```diff
@@ -813,32 +836,19 @@
...
   (testing "Ensure that local_infile is set to true for better MySQL testing"
     (mt/test-drivers [:mysql]
-      (with-mysql-local-infile-activated
-        (is (= "ON" (-> (sql-jdbc.conn/db->pooled-connection-spec (mt/db))
-                        (jdbc/query
-                         ["show global variables like 'local_infile'"])
-                        first
-                        :value)))))))
-
-(deftest load-from-csv-mysql-slow-way-test
```
calherries, line 823: We don't need this test anymore, since we test with `local_infile` off throughout all the other tests now.

test/metabase/upload_test.clj
```diff
@@ -619,136 +639,139 @@
...
+      (mt/with-empty-db
         (testing "Can upload a CSV with missing values"
           (upload/load-from-csv!
            driver/*driver*
            (mt/id)
            "upload_test"
-           (csv-file-with ["id,column_that_doesnt_have_a_value" "2"]))
+           (csv-file-with ["id,column_that_has_one_value,column_that_doesnt_have_a_value"
+                           "1,2"
+                           "2,  ,\n"]))
```
calherries, line 712: Apart from the macro changes, this is the only change to the `upload_test`. I wanted to test for empty and blank values too, so I added more rows and columns.

test/metabase/upload_test.clj
```diff
@@ -34,30 +33,50 @@
...
+(defn- do-with-mysql-local-infile-on-and-off
+  "Helper for [[with-mysql-local-infile-on-and-off]]"
+  [thunk]
+  (if (not= driver/*driver* :mysql)
+    (thunk)
+    (do
+      (testing "with local_infile on"
+        (do-with-mysql-local-infile-off thunk))
+      (testing "with local_infile off"
+        (do-with-mysql-local-infile-off thunk)))))
```
piranha, line 74: is there an error in the last 4 lines?

calherries: Ha! Yes. Great eyes

sansolovyov (approved):
lgtm

PR #36093, opened 2023-11-23
Mapping editor update

### LINKED ISSUE #30196: Broken UI when adding multiple user attributes if keys start with the same string
**Describe the bug**
When adding multiple user attributes, it is currently impossible to add the second one if its key is the superset of the first one. Some examples will explain this better than I can:
| first key | second key |
|--------|--------|
| 1 | 12 |
| Day | Daylight |
| foo | foobar |

As soon as you type the part of the second attribute that matches the first one, UI will remove the option to add more attributes.
Repro steps are going to paint a better picture.

**To Reproduce**
You need to be in an active EE instance for this to work:
1. As an admin, go to the people management section `/admin/people`
2. Click on '....' next to any user and click on "Edit user"
3. Click on "Add an attribute" button
4. For the attribute name type `foo` and for the value type `42`
image
5. Click "Add an attribute" again
6. For the second attribute's key try to type `foobar`
7. Notice that once you reached the `foo` in your typing, the UI removes the "Add an attribute" button and it seemingly removes the second attribute row, leaving you with `foo : 42`
image
8. If you click on "Update" button and observe the network request tab, you'll see that it actually removed the first set of attributes and committed `foo: null`, although the UI suggests otherwise
image
9. If you click on "Edit user" again, the UI will correctly show `foo : null` this time
image
10. But it is still impossible to add a second parameter unless you completely remove the existing one and start again

**Expected behavior**
Probably warn user that two attributes cannot be the same, but do not abruptly remove the row and the ability to add more attributes.

master, local dev, 40e85fa61a8470c3153a0d35b3db0f5e98554a84, H2, Sample Database

**Severity**
P3

**Additional context**
The workaround would be to paste the second key, instead of typing it.

npfitz:
Closes https://github.com/metabase/metabase/issues/35316
Closes https://github.com/metabase/metabase/issues/30196

### Description
This PR re-works the <MappingEditor /> component to not directly edit an object, but instead deconstruct the object into entries on mount, and as changes are made reconstruct the object and pass it through `onChange`. This allows us to be able to work around some of the nuances of JS objects, like the order of `Objects.Entries` not being based on insertion, and keys not being immediately overwritten when there are duplicates. Admittedly, adding state to this component doesn't feel amazing, but this does allow us to get around the 2 bugs above.

We also add an `onError` prop, which is called when 2 keys have the same value. This allows a parent component to disable the submit button in a form if needed. If no `onError` prop is passed, we simply call `onChange`

### How to verify
1. Go to Admin -> People -> Invite User. Fill out the names and email
2. Click "Add an Attribute", give it a key of "foo" and a value of "bar"
3. Click "Add an attribute" again, and give it a key of 1. The attributes should remain in the correct order (#35316)
4. Clear the 1, and type "foo". You should be prompted with an error message, and the update button should be disabled. Both attributes should remain on the screen (#30196)

### Demo
chrome_bKrnFdq5nu

### Checklist

### DIFF
--- a/enterprise/frontend/src/metabase-enterprise/sandboxes/components/LoginAttributesWidget.tsx
+++ b/enterprise/frontend/src/metabase-enterprise/sandboxes/components/LoginAttributesWidget.tsx
@@ -21,7 +21,7 @@ export const LoginAttributesWidget = ({
 
   const handleError = (error: boolean) => {
     if (error) {
-      setError("bad");
+      setError(t`Duplicate login attribute keys`);
     }
   };
 
--- a/enterprise/frontend/src/metabase-enterprise/sandboxes/components/LoginAttributesWidget.tsx
+++ b/enterprise/frontend/src/metabase-enterprise/sandboxes/components/LoginAttributesWidget.tsx
@@ -17,13 +17,20 @@ export const LoginAttributesWidget = ({
   className,
   style,
 }: Props) => {
-  const [{ value }, , { setValue }] = useField(name);
+  const [{ value }, , { setValue, setError }] = useField(name);
+
+  const handleError = (error: boolean) => {
+    if (error) {
+      setError("bad");
+    }
+  };
 
   return (
     <FormField className={className} style={style} title={title}>
       <MappingEditor
         value={value || {}}
         onChange={setValue}
+        onError={handleError}
         addText={t`Add an attribute`}
       />
     </FormField>
--- a/enterprise/frontend/src/metabase-enterprise/sandboxes/components/MappingEditor.tsx
+++ b/enterprise/frontend/src/metabase-enterprise/sandboxes/components/MappingEditor.tsx
@@ -137,14 +137,14 @@ export const MappingEditor = ({
               <td className="pb1" style={{ verticalAlign: "bottom" }}>
                 {!swapKeyAndValue ? keyInput : valueInput}
               </td>
-              <td className="pb1 px1" style={{ verticalAlign: "bottom" }}>
+              <td className="pb1 px1" style={{ verticalAlign: "middle" }}>
                 {divider}
               </td>
               <td className="pb1" style={{ verticalAlign: "bottom" }}>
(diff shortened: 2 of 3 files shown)


enterprise/frontend/src/metabase-enterprise/sandboxes/components/LoginAttributesWidget.tsx
```diff
@@ -17,13 +17,20 @@ export const LoginAttributesWidget = ({
   className,
   style,
 }: Props) => {
-  const [{ value }, , { setValue }] = useField(name);
+  const [{ value }, , { setValue, setError }] = useField(name);
+
+  const handleError = (error: boolean) => {
```
npfitz, line 22: I don't love how this is done, but I wasn't sure what else to do. Errors are cleared after setValue is called, so inside <MappingEditor /> I need to ensure that `onChange` isn't actually called if an error is detected

enterprise/frontend/src/metabase-enterprise/sandboxes/components/MappingEditor.tsx
```diff
@@ -74,34 +118,37 @@ export const MappingEditor = ({
...
             value: value,
             placeholder: valuePlaceholder,
             onChange: newValue =>
-              onChange(replaceMappingValue(mapping, key, newValue)),
+              handleChange(replaceEntryValue(entries, index, newValue)),
           });
           return (
             <tr key={index}>
-              <td className="pb1">
+              <td className="pb1" style={{ verticalAlign: "bottom" }}>
```
npfitz, line 137: Admittedly, inline styling isn't great. However, we don't have a class for verticalAlign, and I thought adding one would be worse. But if needed, this can be a styled component.

I did check and see if Mantine's table component would help, but it's pretty minimal in v6 (it gets better in v7), and doesn't include anything for vertical alignment

iethree (approved):
💖

Issue #26988, opened 2022-12-06
Public embedding demo font not changing


losrebellos:
**Describe the bug**

When changing the `font` parameter "on the fly" on public embedding, the change is not reflected unless the iframe is fully reloaded.
While it work for all the other parameters.

**To Reproduce**

1. Go to the [public embedding](https://embedding-demo.metabase.com/embedding/public) demo page
2. Click on `Titled`, the title appears
3. Click on the `font` dropdown to set it to `Oswald` for instance
4. Nothing changes UI-wise
5. Reload the page so Oswald will be selected by default
6. Now the font is Oswald

**Expected behavior**

When changing the `font` parameter, it should behave like the other parameters and refresh the UI.

**Screenshots**

![image](https://user-images.githubusercontent.com/348552/205875075-23a32bf9-2ae4-4c19-88bb-2749ed27fc83.png)

**Information about your Metabase Installation:**

- Metabase version: v1.44.6 (build on 2022-11-01)

**Severity**

It is impacting the public embedding font feature demo.

**Additional context**

It seems that the static is not (re)rendering the charts when the font parameter is changed.

PR #36095, opened 2023-11-23
Set embedding visual settings before initial render

deniskaber:
Relates https://github.com/metabase/metabase/issues/26988

### Description

This moves embedding settings preparation to happen before the rendering. 
This way we set the state first and then render global styles and content. We avoid extra rerender and possible race conditions with font being applied to a page.

Also we add a code that should dynamically apply embedding settings to the app in "preview mode", which will be used soon in https://github.com/metabase/metabase/issues/35961

### Checklist

### DIFF
--- a/e2e/test/scenarios/sharing/reproductions/26988-embedding-dynamic-settings.cy.spec.js
+++ b/e2e/test/scenarios/sharing/reproductions/26988-embedding-dynamic-settings.cy.spec.js
@@ -1,0 +1,50 @@
+import { restore } from "e2e/support/helpers";
+import { ORDERS_ID } from "metabase-types/api/mocks/presets";
+
+describe("issue 26988", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsAdmin();
+  });
+
+  it("should apply embedding settings passed in URL on load", () => {
+    cy.createDashboardWithQuestions({
+      questions: [
+        {
+          name: "Q1",
+          query: {
+            "source-table": ORDERS_ID,
+            limit: 3,
+          },
+        },
+      ],
+    }).then(({ dashboard }) => {
+      cy.request("POST", `/api/dashboard/${dashboard.id}/public_link`).then(
+        ({ body: { uuid } }) => {
+          cy.signOut();
+
+          cy.visit({
+            url: `/public/dashboard/${uuid}`,
+            qs: { enableCypressIframe: true },
+          });
+
+          // default font
+          cy.get("body").should("have.css", "font-family", `Lato, sans-serif`);
+
+          cy.wait(1000);
+
+          cy.visit({
+            url: `/public/dashboard/${uuid}`,
+            qs: { enableCypressIframe: true, font: "Oswald" },
+          });
+
+          cy.get("body").should(
+            "have.css",
+            "font-family",
+            `Oswald, sans-serif`,
+          );
+        },
+      );
+    });
+  });
+});
--- a/e2e/test/scenarios/sharing/reproductions/26988-embedding-dynamic-settings.cy.spec.js
+++ b/e2e/test/scenarios/sharing/reproductions/26988-embedding-dynamic-settings.cy.spec.js
@@ -39,12 +39,12 @@ describeEE("issue 26988", () => {
 
     getIframeBody().should("have.css", "font-family", `Lato, sans-serif`);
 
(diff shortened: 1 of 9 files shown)


frontend/src/metabase/lib/browser.js
```diff
@@ -7,7 +7,14 @@ function parseQueryStringOptions(s) {
     if (options[name] === "") {
       options[name] = true;
     } else if (/^(true|false|-?\d+(\.\d+)?)$/.test(options[name])) {
-      options[name] = JSON.parse(options[name]);
+      let parsed;
```
WiNloSt, line 10: Why do we need this change? What problem did you encounter?

frontend/src/metabase/app.js
```diff
@@ -73,6 +73,8 @@ function _init(reducers, getRoutes, callback) {
 
   createTracker(store);
 
+  initializeEmbedding(store);
```
WiNloSt, line 76: This change makes sense to me

WiNloSt (commented):
I'm not sure about the change in `frontend/src/metabase/lib/browser.js`, otherwise, this PR looks good to me.

WiNloSt (commented):
This is almost there 🎉 I'd add a test that actually tests our static embedding preview.

You can add a new test in `e2e/test/scenarios/admin-2/whitelabel.cy.spec.js` and find examples how we visit the static embedding modal from `e2e/test/scenarios/embedding/embedding-questions.cy.spec.js`

frontend/src/metabase/lib/dom.js
```diff
@@ -13,6 +13,13 @@ export const getScrollY = () =>
 // Cypress renders the whole app within an iframe, but we want to exlude it from this check to avoid certain components (like Nav bar) not rendering
 export const isWithinIframe = function () {
   try {
+    if (
+      isCypressActive &&
+      window.location.search.includes("enableCypressIframe=true")
```
ranquild, line 18: Why not test the change using real preview in the app and not with a fake parameter?

frontend/src/metabase/lib/embed.js
```diff
@@ -41,12 +40,24 @@ export function initializeEmbedding(store) {
...
+
+    if (IS_EMBED_PREVIEW) {
+      let currentLocation;
+      store.subscribe(() => {
+        const previousLocation = currentLocation;
+        currentLocation = store.getState().routing.locationBeforeTransitions;
+
+        if (
+          previousLocation &&
+          (currentLocation.search !== previousLocation.search ||
```
ranquild, line 52: It looks dangerous to skip certain updates based on this heuristic. Wdyt about using something more stable? I'm not sure, but maybe even checking for `pathname` is better

ranquild (commented):
Comments inline

e2e/test/scenarios/sharing/reproductions/26988-embedding-dynamic-settings.cy.spec.js
```diff
@@ -0,0 +1,56 @@
...
+      visitDashboard(card.dashboard_id);
+    });
+
+    cy.icon("share").click();
+    cy.findByRole("button", { name: "Set up" }).click();
+    cy.wait("@dashboard");
+
+    getIframeBody().should("have.css", "font-family", `Lato, sans-serif`);
+
+    cy.findByTestId("embedding-settings").findByText("Font").siblings().click();
```
WiNloSt, line 42: Let's make this a bit more readable

```suggestion
    cy.findByTestId("embedding-settings").findByLabelText("Font").click();
```

But you have to modify the markup a bit as well.

Here's what I did to the [`DisplayOptionSection`](https://github.com/metabase/metabase/blob/6fbc94fe67463ec292b995416c5eb602659dcb93/frontend/src/metabase/public/components/widgets/DisplayOptionsPane.jsx#L129-L134), but maybe you could come up with something better.

```ts
const DisplayOptionSection = ({ title, children }) => {
  const uniqueId = useUniqueId("display-option");
  return (
    <DisplayOption>
      <DisplayOptionTitle id={uniqueId}>{title}</DisplayOptionTitle>
      <span aria-labelledby={uniqueId}>{children}</span>
    </DisplayOption>
  );
};
```

WiNloSt (approved):
Looks good to me, but if you could fix the E2E before merging that'd be great.

PR #36100, opened 2023-11-24
Fix BE flake 36089: metabase.api.search-test/search-db-call-count-test

### LINKED ISSUE #36089: BE flake: metabase.api.search-test/search-db-call-count-test
Example run: https://github.com/metabase/metabase/actions/runs/6971928480/job/18973054258?pr=36082#step:6:476

I've seen it twice, on Mongo and MySQL test runs.

calherries:
Fixes https://github.com/metabase/metabase/issues/36089

The problem was there could be other objects in the database that matched the search query, depending on the order that the tests were run.

I stress tested this on this commit [c76bc03](https://github.com/metabase/metabase/actions/runs/6983209131/job/19003936431?pr=36100), running the test 20 times for each driver.

### DIFF
--- a/.github/workflows/drivers.yml
+++ b/.github/workflows/drivers.yml
@@ -350,6 +350,7 @@ jobs:
       uses: ./.github/actions/test-driver
       with:
         junit-name: 'be-tests-mysql-8-0-ee'
+        test-args: ":only metabase.api.search-test/search-db-call-count-test"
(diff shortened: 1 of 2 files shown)


sansolovyov (approved):
nice catch!

qnkhuat (approved):

Issue #36104, opened 2023-11-24
Commented out parameters in SQL queries are still showing up as parameters on the card


oleggromov:
### Describe the bug

A commented out parameter is nonetheless counted as a parameter on the card

![image](https://github.com/metabase/metabase/assets/2196347/81b3fc02-683a-41c7-9b56-791c66d74402)

### To Reproduce

1. Create an SQL query
2. Add a parameter
3. Comment it out
4. See that it still is parsed as if it wasn't

### Information about your Metabase installation

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.name": "Java(TM) SE Runtime Environment",
    "java.runtime.version": "21.0.1+12-LTS-29",
    "java.vendor": "Oracle Corporation",
    "java.vendor.url": "https://java.oracle.com/",
    "java.version": "21.0.1",
    "java.vm.name": "Java HotSpot(TM) 64-Bit Server VM",
    "java.vm.version": "21.0.1+12-LTS-29",
    "os.name": "Mac OS X",
    "os.version": "14.1.1",
    "user.language": "en",
    "user.timezone": "Europe/London"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
    "application-database-details": {
      "database": {
        "name": "H2",
        "version": "2.1.214 (2022-06-13)"
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
        "version": "2.1.214 (2022-06-13)"
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2023-11-21",
      "tag": "v0.48.0-SNAPSHOT",
      "hash": "d149840"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

might break "Explore results"

oleggromov:
Duplicate of https://github.com/metabase/metabase/issues/33585 and related to https://github.com/metabase/metabase/issues/11755

PR #36105, opened 2023-11-24
QueryBuilder and SQL questions: hide Explore results with parameters

### LINKED ISSUE #35476: Don't show "explore results" on SQL questions with filters
### Describe the bug

If you build a SQL question with filters and then click on explore results, question will fail, so don't make the users trip on this error

### To Reproduce

1) create a sql question (select * from people where state = {{state}})
2) click on explore results

see the error

### Expected behavior

that button shouldn't appear if there are filters

### Logs

_No response_

```JSON
NA
```

### Severity

It has been like that forever

### Additional context

_No response_

oleggromov:
Fixes https://github.com/metabase/metabase/issues/35476

### Description
Sometimes SQL-based questions could be used a data source for GUI questions, which is done by clicking "Explore results".

This is only possible though when a few conditions are met, and the missing one was that the question must have exactly 0 parameters (as per issue). 

This PR fixes it and does a bit of housekeeping. 

### How to verify

1. Create an SQL question without parameters, click save - you should see the "Explore results" button
2. Now add a parameter, for instance, `select * from orders where product_id = {{product_id}}`, click "Save" - the button must not show
3. There's a known bug: https://github.com/metabase/metabase/issues/36104 (see the last comment, there's a link to the one that supercedes it) 

### Demo

https://github.com/metabase/metabase/assets/2196347/c1d0da47-b8f8-436f-a001-56bb6016a01e

### Checklist

### DIFF
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -1250,6 +1250,21 @@ class Question {
   }
 
   /**
+   * We can only "explore results" (i.e. create new questions based on this one)
+   * when question is a native query, which is saved, has no parameters
+   * and satisfies other conditionals below.
+   */
+  canExploreResults() {
+    return (
+      this.isNative() &&
+      this.isSaved() &&
+      this.parameters().length === 0 &&
+      this.query().canNest() &&
+      !this.query().readOnly() // originally "canRunAdhocQuery"
+    );
+  }
+
+  /**
    * TODO Atte Keinänen 6/13/17: Discussed with Tom that we could use the default Question constructor instead,
    * but it would require changing the constructor signature so that `card` is an optional parameter and has a default value
    */
--- a/frontend/src/metabase-lib/queries/AtomicQuery.ts
+++ b/frontend/src/metabase-lib/queries/AtomicQuery.ts
@@ -34,7 +34,7 @@ export default class AtomicQuery extends Query {
     return null;
   }
 
-  canNest() {
-    return this.database()?.hasFeature("nested-queries");
+  canNest(): boolean {
+    return Boolean(this.database()?.hasFeature("nested-queries"));
   }
 }
--- a/frontend/src/metabase-lib/question.unit.spec.ts
+++ b/frontend/src/metabase-lib/question.unit.spec.ts
@@ -8,6 +8,8 @@ import { createMockMetadata } from "__support__/metadata";
 import {
   createMockCard,
   createMockNativeDatasetQuery,
+  createMockParameter,
+  createMockTemplateTag,
 } from "metabase-types/api/mocks";
 import Question from "./Question";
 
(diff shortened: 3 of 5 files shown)


frontend/src/metabase-lib/Question.ts
```diff
@@ -1249,6 +1249,21 @@ class Question {
...
+  canExploreResults() {
+    return (
+      this.isNative() &&
+      this.isSaved() &&
+      this.parameters().length === 0 &&
+      this.query().canNest() &&
+      !this.query().readOnly() // originally "canRunAdhocQuery"
+    );
+  }
+
```
oleggromov, line 1266: I was a bit hesitant to enlarge this god-object even further but it's still better than keeping this logic in essentially a view.

A nice consequence of that is that we could easily test this behavior with just a few unit tests as it is a pure function, which doesn't require any mocking or integration. 

I've seen that there's no tests for the `Question` now. Should I add them?

ranquild: This is a (very) legacy file https://github.com/metabase/metabase/blob/bf45689cd60bf337bc93bc7e23abf98f4a481a91/frontend/test/metabase-lib/lib/Question.unit.spec.js#L324 but that's what we have right now. We should add tests for this change.

ranquild: Btw, for new things we don't separate test files from the implementation files

ranquild (commented):
Looks good, please add tests

frontend/src/metabase-lib/queries/AtomicQuery.ts
```diff
@@ -34,7 +34,7 @@ export default class AtomicQuery extends Query {
     return null;
   }
 
-  canNest() {
-    return this.database()?.hasFeature("nested-queries");
+  canNest(): boolean {
```
oleggromov, line 37: Previously, this would've returned `undefined` or `true`.

frontend/src/metabase-lib/question.unit.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    expect(q.canExploreResults()).toBe(false);
+  });
+
+  it("should be false with parameters", () => {
+    const card = createMockCard({
+      dataset_query: createMockNativeDatasetQuery({
+        type: "native",
+        native: {
+          query: "select * from order where id > {{min_id}}",
+          "template-tags": {
```
ranquild, line 35: `createMockTemplateTag`

frontend/src/metabase-lib/question.unit.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+            min_id: {
+              type: "text",
+              name: "min_id",
+              id: "uuid",
+              "display-name": "Min ID",
+            },
+          },
+        },
+      }),
+      parameters: [
```
ranquild, line 45: `createMockParameter`

frontend/src/metabase/query_builder/components/view/ExploreResultsLink.tsx
```diff
@@ -0,0 +1,25 @@
+import { t } from "ttag";
+import Link from "metabase/core/components/Link";
+import ViewButton from "metabase/query_builder/components/view/ViewButton";
+import type Question from "metabase-lib/Question";
+import { getUrl as ML_getUrl } from "metabase-lib/urls";
+
+type Props = {
```
ranquild, line 7: `interface ExploreResultsLinkProps`

ranquild (commented):
@oleggromov Please use helpers like `createMockTemplateTag` for creating template tags and parameters in tests. This is to make sure all required properties have correct default values, now and in the future.

frontend/src/metabase/query_builder/components/view/ExploreResultsLink.tsx
oleggromov: If it's not exported, why would we need that?

ranquild (approved):

PR #36107, opened 2023-11-24
Add public link creation user flow

oisincoveney:
Epic #35961

### Description

Adds the public link popover for the dashboard and question embed popover. This popover, when opened, should create a public link for the resource and allow users to copy the link. Admins should also see a `Remove public link` button that will allow them to remove the public link to the resource.

### How to verify

There's two scenarios, admin and non-admin. Ensure that public sharing is enabled for the instance, as the public link popover should never appear if public sharing is off.

If the user is an admin:
* They should see the embed popover when they click on the embed button, and see `Create a public link`
* They should also see `Remove public link`, and clicking it should remove the link.
* If the link has already been created, the initial popover should say `Public link` instead of `Create a public link`. Clicking `Public link` should give the same URL as they have seen before.

If the user is a non-admin:
* They should only see the Embed button if a public link has been created.
* They should be able to copy the public link
* They shouldn't be able to see `Remove public link`

### Screenshots

**_An admin should see:_**

_When a link hasn't been created_
image

_When a link has been created_
image

image

_**A non-admin should see:**_

_When a link has been created_
image

_When no link exists - no embed button_
image

### Checklist


frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx
```diff
@@ -0,0 +1,45 @@
+import type { Dashboard } from "metabase-types/api";
```
oisincoveney, line 1: I split the popover into `Dashboard` and `Question` popovers to separate the different actions and the fact that getting a `uuid` for questions and dashboards is completely different.

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
```diff
@@ -0,0 +1,231 @@
...
+import type {
+  ExportFormatType,
+  PublicLinkPopoverProps,
+} from "./PublicLinkPopover";
+import { PublicLinkPopover } from "./PublicLinkPopover";
+
+// https://github.com/nkbt/react-copy-to-clipboard/issues/106#issuecomment-605227151
+jest.mock("copy-to-clipboard", () => {
+  return jest.fn();
+});
```
npretto, line 15: super nit:
```suggestion
jest.mock("copy-to-clipboard", () => jest.fn());
```

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
```diff
@@ -0,0 +1,231 @@
...
+    createPublicLink,
+    deletePublicLink,
+    getPublicLink,
+    onClose,
+  };
+};
+
+describe("PublicLinkPopover", () => {
+  describe("when rendering for admins vs non-admins", () => {
+    it("should render public link information for admins", async () => {
```
npretto, line 101: non blocking suggestion:
I find test suites following the [given when then](https://martinfowler.com/bliki/GivenWhenThen.html) a bit more readable
In this example what do you think about this structure?
```
(given a) PublicLinkPopover
	when the user is an admin
	  it should render public link information
	  it should render `Remove this public link`
	when the user is not an admin
	  it should render public link information
	  it should not render `Remove this public link`
...
```

frontend/src/metabase/dashboard/components/EmbedMenu/EmbedMenu.unit.spec.tsx
```diff
@@ -162,6 +169,27 @@ describe("EmbedMenu", () => {
       });
     });
 
-    // TODO: Write tests for showing the popover when the user is non-admin and a public link has/hasn't been created
+    describe("when user is non-admin", () => {
+      it("should render not render the button if a link hasn't been created", async () => {
```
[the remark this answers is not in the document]
oisincoveney: Looks like I had a brainfart there. Thanks!

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
oisincoveney: Gotcha, I was looking at that same article yesterday and it does make sense, the semantics just looked weird to me, but I do like the structure you've laid out. I'll make sure the tests pass first before making any changes

npretto: Yeah when I first read it it looked a bit abstract to me, now I just consider the "given x" as "given this component i need to test", and the "when" as "when we are in scenario X" and it usually helps me writing down the tests

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+            {loading ? (
+              <PublicLinkTextContainer>
+                <Text truncate c="text.0">{t`Loading…`}</Text>
+              </PublicLinkTextContainer>
+            ) : (
+              <>
+                <PublicLinkTextContainer>
+                  <Text truncate>{url}</Text>
+                </PublicLinkTextContainer>
+                <PublicLinkCopyButton value={url} />
```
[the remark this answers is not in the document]
oisincoveney: Ah lol I did that on the first PR of Milestone 2. I'll bring it over here instead

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
```diff
@@ -0,0 +1,122 @@
...
+                >
+                  {extension}
+                </ExtensionOption>
+              ))}
+            </Group>
+          )}
+          {isAdmin && (
+            <Box mt="md">
+              <Anchor fz="sm" c="error.0" fw={700} onClick={onRemoveLink}>
+                {t`Remove this public link`}
```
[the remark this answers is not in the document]
oisincoveney: Yeah that's fine in the first PR of milestone 2. I'll bring it over here since it probably makes more sense now

WiNloSt (commented):
I need to hop on a call, so I'll be back to review this

frontend/src/metabase/dashboard/actions/sharing.js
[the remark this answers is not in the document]
oisincoveney: done

frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal.jsx
```diff
@@ -13,7 +13,6 @@ import {
   updateEnableEmbedding,
   updateEmbeddingParams,
 } from "../actions";
-import { ModalTrigger } from "./DashboardSharingEmbeddingModal.styled";
```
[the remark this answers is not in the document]
oisincoveney: I think I might have done a no-verify commit but I can't remember to be completely honest.

frontend/src/metabase/query_builder/actions/sharing.js
WiNloSt, line 1: Could you convert this file to TypeScript too? You already typed 2 actions, and there are only 2 left that aren't typed.

WiNloSt (commented):
I haven't reviewed the test and will review those once all feedback are addressed.

frontend/src/metabase/query_builder/actions/sharing.js
oisincoveney: heh I was doing it as you typed up the comment. should be good now

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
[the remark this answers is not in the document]
oisincoveney: done, should have the same behavior

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+      userEvent.click(screen.getByLabelText("copy icon"));
+
+      expect(await screen.findByText("Copied!")).toBeInTheDocument();
+    });
+  });
+
+  describe("when appending extensions to the link", () => {
+    it("should append the extension to the link when the extension is clicked on", async () => {
+      setup({ hasUUID: true, isOpen: true, extensions: ["csv"] });
```
WiNloSt, line 215: I don't think you need to add an option for `extensions`. It would make more sense to add an option whether or not this is a dashboard or a question sharing popover, and if it's a question sharing popover we always pass `["csv", "xlsx", "json"]`. Do you think that'll be easier to understand?

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+      expect(await screen.findByText("sample-public-link")).toBeInTheDocument();
+
+      userEvent.click(screen.getByText("csv"));
+      expect(
+        await screen.findByText("sample-public-link.csv"),
+      ).toBeInTheDocument();
+    });
+
+    it("should remove the extension when the extension is clicked on again", async () => {
```
WiNloSt, line 225: This reaffirms my point made in [this comment](https://github.com/metabase/metabase/pull/36107#discussion_r1411789174). I think toggling the extension looks very unintuitive imo.

frontend/src/metabase/query_builder/actions/sharing.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const CREATE_PUBLIC_LINK = "metabase/card/CREATE_PUBLIC_LINK";
+
+export const createPublicLink = createAction(
+  CREATE_PUBLIC_LINK,
+  ({
+    id,
+  }: Card): Promise<
+    CardIdPayload & {
```
WiNloSt, line 17: I'd inline `id: CardId` instead. When I see `Type & { property: Type }` I'd expect `Type` to have a handful of keys otherwise, it's easier to read the return type when inlining everything since it only has 2 properties in this case.

e2e/support/helpers/e2e-embedding-helpers.js
```diff
@@ -114,3 +114,47 @@ export function visitIframe() {
...
+}
+
+export function openPublicLinkDropdown() {
+  cy.icon("share").click();
+  cy.findByTestId("embed-header-menu")
+    .findByTestId("embed-menu-link-item")
+    .click();
+}
+
+export function openEmbeddingSettingsPage() {
```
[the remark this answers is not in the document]
oisincoveney: Gotcha. I saw that Nemanja had used `openEmbeddingSettingsPage` before so I rolled with that, but this is a better name

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
```diff
@@ -58,17 +61,30 @@ const setup = ({
...
+        "enable-embedding": embeddingEnabled,
       }),
     }),
   });
 };
 
 describe("DashboardEmbedAction", () => {
   describe("which label should be rendered", () => {
-    it("should have a `Sharing` tooltip if public sharing is true", () => {
+    it("should have a `You must enable Embedding` tooltip if embedding is false", async () => {
```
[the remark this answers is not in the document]
oisincoveney: Yeah, let's go with `disabled` since that's just easier to read at a glance.

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
oisincoveney: Yep, I totally agree. It feels like it was just shoehorned in

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.unit.spec.tsx
oisincoveney: I'm not so sure on this one - I think using `extensions` is both more explicit and easier to understand than passing a dashboard or question and then adding extensions based on that. It would be another layer to think about imo

WiNloSt: It's more explicit yes. But that kind of tailors we allow passing any `extensions` wherever `PublicLinkPopover` is used.

Now the way I see it is that we only have 2 places this `PublicLinkPopover` is being used, and only one of them (questions) can contain `extensions` and extensions aren't dynamic either they're either CSV, XLSX, or JSON.

These tests also play a very important role in the API design, do we really design this component to be able to accept `extensions` dynamically, or do we just want to make it very strict and say if it's rendered for questions, we will pass these 3 extensions, but if it's rendered for dashboard, we don't pass any extensions.

There is no right or wrong answer to this question, but to me, it's a matter of which one is more appropriate at that point in time.

My argument would be it's been like this for I don't know how long and I don't see we will display extensions anywhere else outside questions anytime soon.

frontend/src/metabase/lib/urls/dashboards.ts
```diff
@@ -32,8 +32,13 @@ export function dashboard(
   }
 }
 
-export function publicDashboard(uuid: string) {
-  const siteUrl = MetabaseSettings.get("site-url");
+export function publicDashboard({
+  uuid,
+  siteUrl = MetabaseSettings.get("site-url"),
```
WiNloSt, line 37: I don't see the point of passing `siteUrl` since the only place you use that also gets the `siteUrl` from the settings which should have the same value as `MetabaseSettings.get("site-url")` anyway.

https://github.com/metabase/metabase/blob/2556785498373fe796583a706519989a4e64f9a9/frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx#L23-L28

What was the problem you encountered that you needed to pass `siteUrl` in the first place? It'd be great if we could leave the function API as is (only accepts a single argument `uuid`)

frontend/src/metabase/admin/settings/components/widgets/PublicLinksListing/PublicLinksListing.jsx
```diff
@@ -135,7 +135,9 @@ export const PublicLinksDashboardListing = () => (
     revoke={DashboardApi.deletePublicLink}
     type={t`Public Dashboard Listing`}
     getUrl={dashboard => Urls.dashboard(dashboard)}
-    getPublicUrl={({ public_uuid }) => Urls.publicDashboard(public_uuid)}
+    getPublicUrl={({ public_uuid }) =>
+      Urls.publicDashboard({ uuid: public_uuid })
```
WiNloSt, line 139: Commented in `frontend/src/metabase/lib/urls/dashboards.ts`. I think we could preserve the current function API.

frontend/src/metabase/dashboard/actions/sharing.ts
WiNloSt, line 1: TS conversion 🎉

frontend/src/metabase/dashboard/components/PublicLinkPopover/QuestionPublicLinkPopover.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <PublicLinkPopover
+      target={target}
+      isOpen={isOpen}
+      onClose={onClose}
+      createPublicLink={createPublicQuestionLink}
+      deletePublicLink={deletePublicQuestionLink}
+      uuid={uuid}
+      getPublicLink={getPublicLink}
+      extensions={exportFormats}
```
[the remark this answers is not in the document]
albertoperdomo: How do you trigger that behavior? I have an instance on 47.7 and can't see the extensions from the embed/share pop-up.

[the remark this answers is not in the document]
albertoperdomo: OK, this is present when sharing a question instead of a dashboard, I had missed that.

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
```diff
@@ -58,20 +61,33 @@ const setup = ({
...
+        "enable-embedding": embeddingEnabled,
       }),
     }),
   });
 };
 
 describe("DashboardEmbedAction", () => {
   describe("which label should be rendered", () => {
-    it("should have a `Sharing` tooltip if public sharing is true", () => {
+    it("should have a `You must enable Embedding` tooltip if embedding is disabled", async () => {
```
WiNloSt, line 72: there are `"when the user is admin"` block for other tests, but not this one. I believe this test is only true when users are admins as well. Should you nest some of the tests under `"when the user is admin"` block or just add the admin condition to the test name if it's not applicable to other tests under this same block.

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
```diff
@@ -48,13 +48,15 @@ export const DashboardEmbedHeaderButton = forwardRef(
...
         offset={8}
       >
-        <DashboardHeaderButton
-          data-testid="dashboard-embed-button"
-          icon="share"
-          disabled={disabled}
-          onClick={onClick}
-          ref={ref}
-        />
+        <Box>
```
WiNloSt, line 51: What does `Box` do in this case? Is there any visual change?

WiNloSt (commented):
I've already got some comments, I'll keep reviewing, but I'll submit these in the mean time.

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const uuid = dashboard.public_uuid;
+
+  const siteUrl = useSelector(state => getSetting(state, "site-url"));
+  const getPublicLink = () => {
+    if (!uuid) {
+      return null;
+    }
+    return getPublicDashboardUrl({ uuid, siteUrl });
+  };
```
WiNloSt, line 29: I kind of think instead of passing both `extensions` and `getPublicLink` in `QuestionPublicLinkPopover` would it make more sense to just pass something like `linkOptions` where each option has
- url
- name (used to display in the UI)

I feel like passing `getPublicLink` makes `PublicLinkPopover` very coupled with the component it's called from. `PublicLinkPopover` needs to know what arguments `getPublicLink` accepts to be able to pass them accordingly. But passing something like `linkOptions` could make it feel less coupled.

And if we only send 1 option we won't display the list of options to choose from (that's what this component does)

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx
```diff
@@ -0,0 +1,49 @@
...
+  dashboard: Dashboard;
+  target: JSX.Element;
+  isOpen: boolean;
+  onClose: () => void;
+}) => {
+  const dispatch = useDispatch();
+
+  const uuid = dashboard.public_uuid;
+
+  const siteUrl = useSelector(state => getSetting(state, "site-url"));
```
WiNloSt, line 23: Also, I've commented about passing `siteUrl` to `metabase/lib/urls/*` before. I'm not sure why would we do that because when we don't pass it, the URL function would use the `site-url` settings via `MetabaseSettings.get("site-url")` anyway, so it feels redundant.

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.unit.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+    setup();
+
+    expect(screen.queryByTestId("extension-option")).not.toBeInTheDocument();
+  });
+
+  it("should call Dashboard public link API when creating link", () => {
+    setup({ hasPublicLink: false });
+
+    expect(
+      fetchMock.calls(`path:/api/dashboard/${TEST_DASHBOARD_ID}/public_link`),
```
WiNloSt, line 88:
```suggestion
      fetchMock.calls(`path:/api/dashboard/${TEST_DASHBOARD_ID}/public_link`, {
        method: "POST",
      }),
```

I think we should make assertions more explicit. Now we don't assert what method this API is called with. It could be confused whether we're trying to get a public_link or actually setting this dashboard to be publically available.

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
```diff
@@ -0,0 +1,124 @@
...
+    null,
+  );
+
+  const url = getPublicLink({ exportFormat });
+
+  const { loading } = useAsync(async () => {
+    if (isOpen && !uuid) {
+      await createPublicLink();
+    }
+    return uuid;
```
WiNloSt, line 54:
```suggestion
```

Do we need to return anything at all since we're not using it?

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const [exportFormat, setExportFormat] = useState<ExportFormatType | null>(
+    null,
+  );
+
+  const url = getPublicLink({ exportFormat });
+
+  const { loading } = useAsync(async () => {
+    if (isOpen && !uuid) {
+      await createPublicLink();
```
WiNloSt, line 52: instead of returning something we could make this function non async and

```ts
return createPublicLink()
```

I haven't tested it, but it should work. Given `async/await` and Promise are isomorphic

frontend/src/metabase/dashboard/components/PublicLinkPopover/QuestionPublicLinkPopover.tsx
```diff
@@ -0,0 +1,61 @@
...
+  onClose,
+}: {
+  question: Question;
+  target: JSX.Element;
+  isOpen: boolean;
+  onClose: () => void;
+}) => {
+  const dispatch = useDispatch();
+
+  const siteUrl = useSelector(state => getSetting(state, "site-url"));
```
WiNloSt, line 28: I have other comments about `siteUrl`. Basically I'm not sure what we try to achieve by passing it to `getPublicQuestionUrl` when the case that we don't pass it, we still get the settings from `MetabaseSettings.get("site-url")` anyway.

WiNloSt (commented):
Awesome! This is almost ready to merge. There are only a couple of small things that we should polished first as commented.

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx
oisincoveney: I ended up doing that because MetabaseSettings just wasn't working with my tests. I figured that something was wrong and ended up passing the site-url directly

WiNloSt: You can use `mockSettings` which also sets the value in `MetabaseSettings` automatically, or just set `MetabaseSettings` value separately since they're not in sync by default in unit tests.
 https://github.com/metabase/metabase/blob/2a77c3041174167098a9f0e3eeebc2702e634789/frontend/test/__support__/settings.ts#L7

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.unit.spec.tsx
oisincoveney: Ah ok I wanted to do something like this but I wasn't even sure if this was something we could do. Thanks for that

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx
oisincoveney: Oh ok this is exactly what I was looking for. Thanks for the shout, I'll revert those changes as well. I can't say I liked them either

e2e/test/scenarios/sharing/public-dashboard.cy.spec.js
```diff
@@ -66,7 +68,7 @@ describe("scenarios > public > dashboard", () => {
       questionDetails,
       dashboardDetails,
     }).then(({ body: { id, card_id, dashboard_id } }) => {
-      cy.wrap(dashboard_id).as("dashboardId");
+      cy.wrap(dashboard_id).as("publicDashboardId");
```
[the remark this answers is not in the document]
oisincoveney: Makes sense. Since we're only need a non-public dashboard for one test here, I'll reuse the one that's there right now. Shouldn't be an issue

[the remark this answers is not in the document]
oisincoveney: You're absolutely right, that extra dashboard was completely unnecessary. Thanks for the catch

e2e/test/scenarios/sharing/public-question.cy.spec.js
```diff
@@ -56,20 +58,26 @@ describe("scenarios > public > question", () => {
     cy.request("PUT", "/api/setting/enable-public-sharing", { value: true });
 
     cy.createNativeQuestion(questionData).then(({ body: { id } }) => {
-      cy.wrap(id).as("questionId");
+      cy.wrap(id).as("publicQuestionId");
```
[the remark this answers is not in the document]
oisincoveney: yep, same with this. I reverted it to the questionId

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
oisincoveney: So basically it was an issue with the fact that Mantine tooltip, by default, doesn't render when the button is disabled. But, I just saw on the mantine docs (in the Button section for some reason, not the tooltip) that we can use the `data-disabled` prop to allow tooltips to work on disabled buttons (https://v6.mantine.dev/core/button/#disabled-button-with-tooltip). I'll use this instead

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
oisincoveney: Ok I have two options for you then - here's what it looks like when we use `return createPublicLink`:

```
  const { loading } = useAsync(() => {
    if (isOpen && !uuid) {
      return createPublicLink();
    }
    
	// we'd need to return a promise since TS won't accept a return of just `undefined`
    return Promise.resolve();
  }, [uuid, isOpen]);
```

If we use `async` here we can get rid of that weird last line, since the type of the function can now be `() => Promise<void|null>`

```
const { loading } = useAsync(async () => {
    if (isOpen && !uuid) {
      return createPublicLink();
    }
  }, [uuid, isOpen]);
```

Let me know what you think. I think the second one is cleaner

frontend/src/metabase/dashboard/components/PublicLinkPopover/DashboardPublicLinkPopover.tsx
oisincoveney: I'll do this:
* Instead of using `getPublicLink`, I'll use the dashboard and question specific components to generate the link that will be displayed in the component
* For the `extensions`, I can still pass those into the `PublicLinkPopover`, but turn it into a controlled thing so the `PublicLinkPopover` doesn't deal with state.

This will allow `PublicLinkPopover` to be a completely display-driven component - its job will be to display the link and extensions if they exist, give the question/dashboard-specific components the selected options to create the link, and call the `create/delete` functions when its opened. It should have nothing to do with link formatting or construction

e2e/test/scenarios/visualizations-tabular/pivot_tables.cy.spec.js
```diff
@@ -674,19 +676,13 @@ describe("scenarios > visualizations > pivot tables", { tags: "@slow" }, () => {
...
 
         it("should display pivot table in a public link", () => {
-          // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-          cy.findByText("Public link")
-            .parent()
-            .find("input")
-            .invoke("val")
+          openPublicLinkDropdown();
+          cy.findByTestId("public-link-popover-content")
+            .findByText(/^http/)
```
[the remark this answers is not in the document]
oisincoveney: yep good call

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
WiNloSt: Gotcha

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
oisincoveney: Yep. The tldr is basically that, without the `async` things aren't automatically wrapped in promises anymore, so option 1's return value is actually `Promise<void> | undefined`  instead of `Promise<void | undefined>` and `useAsync` needs a promise for all possible returns

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -0,0 +1,194 @@
...
+          cy.request("PUT", "/api/setting/enable-embedding", { value: true });
+        });
+
+        describe("when user is admin", () => {
+          it(`should show the embed modal for ${resource} immediately when embed button is clicked`, () => {
+            cy.get("@resourceId").then(id => {
+              visitResource(resource, id);
+            });
+
+            openToEmbedModal({
```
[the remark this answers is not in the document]
oisincoveney: I agree though. Hell of a lot easier to follow

frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.unit.spec.tsx
WiNloSt: Yeah, that's why I thought the test name should also contain `user is an admin` because this is the tooltip that only admins should see.

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -0,0 +1,194 @@
...
+              visitResource(resource, id);
+            });
+
+            createPublicLinkDropdown(resource);
+
+            cy.findByTestId("public-link-popover-content").within(() => {
+              cy.findByText("Public link").should("be.visible");
+              cy.findByText(/^http/).should("be.visible");
+              cy.findByText("Remove this public link").should("be.visible");
+            });
```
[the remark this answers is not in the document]
oisincoveney: added some utility functions to do this and added them to this test. I think in the future I can add them to more tests

e2e/support/helpers/e2e-embedding-helpers.js
```diff
@@ -115,28 +115,36 @@ export function visitIframe() {
   });
 }
 
-export function openToEmbedModal({
-  isAdmin = true,
-  isPublicSharingEnabled = true,
-} = {}) {
+export function openPublicLinkDropdown() {
```
WiNloSt, line 118: I just noticed the name. Isn't it more correct to be a popover? A dropdown is an input where you could select usually one option from multiple options.
```suggestion
export function openPublicLinkPopover() {
```

Also when paring with the below function `openEmbedModalFromMenu` this could be made more aligned

```suggestion
export function openPublicLinkPopoverFromMenu() {
```

frontend/src/metabase/dashboard/components/PublicLinkPopover/QuestionPublicLinkPopover.tsx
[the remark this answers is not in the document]
WiNloSt: We're likely to leave this as is https://metaboat.slack.com/archives/C063Q3F1HPF/p1701931282130499?thread_ts=1701753617.997299&cid=C063Q3F1HPF

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
```diff
@@ -39,6 +39,11 @@ export const DashboardEmbedHeaderButton = forwardRef(
       disabled,
     });
 
+    const onHeaderButtonClick = (event: MouseEvent<HTMLButtonElement>) => {
+      event.preventDefault();
+      onClick?.();
+    };
```
WiNloSt, line 45: I'm not sure what this does. Can you elaborate on why we just pass `onClick`?

WiNloSt (approved):
Have a few comments, otherwise, this is ready to merge. Great work!

e2e/support/helpers/e2e-embedding-helpers.js
oisincoveney: Perfect 😄

frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx
oisincoveney: It's a consequence of the the whole disabled button + tooltip thing with Mantine (https://v6.mantine.dev/core/button/#disabled-button-with-tooltip) - we need to explicitly ensure that we precent the default event to make sure that the tooltip still works on disabled buttons

WiNloSt: I've tried removing `onHeaderButtonClick` and passing `onClick` directly and it worked. By the way, I'm not sure how `onClick` should solve the problem when tooltip doesn't show because tooltip will appear on hover, not on clicked.

Can you double check?

oisincoveney: Let me try something real quick. I might be able to get rid of the Box

oisincoveney: I'll readd the Box for now to get tests to pass, Cypress really wants that wrapper div, at least in the CI environment. locally it works fine but I don't want to play with fire here

PR #36031, opened 2023-11-22
Expressions in AggregationPicker without legacy query

kamilmielnik:
Closes #35947

Blocked by ~#36075~, ~#36120~, #36167, ~#36168~
Target branch includes fixes for #36167.

### Description

`AggregationPicker` no longer uses `legacyQuery` or `expression`.
`ExpressionWidget`, `ExpressionEditorTextfield` & `AggregationPicker` still receive `legacyQuery` and `expression` props for backwards-compatibility but now they also work without them.

### How to test it

- add / edit aggregations in GUI editor
    - using custom expressions with or without custom names (this does not work yet - #36167)

image

PR #36109, opened 2023-11-24
Migrate suggestions to MLv2

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/36233

### Description

Rewrite of `suggest.ts` to MLv2.

Covered areas:

- custom column expression
 
image

- aggregation custom expression
image

Admin: segments and metrics are broken intentionally, a fallback can be used instead of real query:

`ExpressionEditorTextfield.tsx` -> `handleCursorChange`
 
```
query: query ?? legacyQuery.question()._getMLv2Query(),
stageIndex: stageIndex ?? -1,
```

### How to verify

- go to notebook editor
- add custom column
- add aggregation custom expression

Both should work.

All tests passed with a fallback

### Checklist

### DIFF
--- a/frontend/src/metabase-lib/expressions/__support__/expressions.js
+++ b/frontend/src/metabase-lib/expressions/__support__/expressions.js
@@ -6,104 +6,118 @@ import {
   createMockTable,
 } from "metabase-types/api/mocks";
 import { createMockMetadata } from "__support__/metadata";
+import { SAMPLE_DB_ID } from "metabase-types/api/mocks/presets";
 import { TYPE } from "metabase-lib/types/constants";
+import { SAMPLE_DATABASE } from "metabase-lib/test-helpers";
 
-const DB_ID = 1;
+const DB_ID = SAMPLE_DB_ID;
 const TABLE_ID = 1;
 
-const metadata = createMockMetadata({
-  databases: [
-    createMockDatabase({
-      id: DB_ID,
-      name: "db",
-      tables: [
-        createMockTable({
-          db: DB_ID,
-          fields: [
-            createMockField({
-              id: 1,
-              table_id: TABLE_ID,
-              display_name: "A",
-              base_type: TYPE.Float,
-            }),
-            createMockField({
-              id: 2,
-              table_id: TABLE_ID,
-              display_name: "B",
-              base_type: TYPE.Float,
-            }),
-            createMockField({
-              id: 3,
-              table_id: TABLE_ID,
-              display_name: "C",
-              base_type: TYPE.Float,
-            }),
-            createMockField({
-              id: 10,
-              table_id: TABLE_ID,
-              display_name: "Toucan Sam",
-              base_type: TYPE.Float,
-            }),
-            createMockField({
-              id: 11,
-              table_id: TABLE_ID,
-              display_name: "Sum",
-              base_type: TYPE.Float,
-            }),
-            createMockField({
-              id: 12,
-              table_id: TABLE_ID,
-              display_name: "count",
-              base_type: TYPE.Float,
-            }),
-            createMockField({
-              id: 13,
-              table_id: TABLE_ID,
-              display_name: "text",
-              base_type: TYPE.Text,
-            }),
-            createMockField({
-              id: 14,
-              table_id: TABLE_ID,
-              display_name: "date",
-              base_type: TYPE.DateTime,
-            }),
-          ],
-          segments: [
-            createMockSegment({
-              id: 1,
-              name: "segment",
-            }),
-          ],
-          metrics: [
-            createMockMetric({
-              id: 1,
-              name: "metric",
-            }),
-            createMockMetric({
-              id: 2,
-              name: "metric",
-            }),
-          ],
+export const DEFAULT_QUERY = {
+  database: SAMPLE_DATABASE.id,
+  type: "query",
+  query: {
+    "source-table": TABLE_ID,
+  },
+};
+
+const database = createMockDatabase({
+  id: SAMPLE_DB_ID,
+  name: "db",
+  tables: [
+    createMockTable({
+      db: DB_ID,
+      id: TABLE_ID,
+      fields: [
+        createMockField({
+          id: 1,
+          table_id: TABLE_ID,
+          display_name: "A",
+          base_type: TYPE.Float,
+        }),
+        createMockField({
+          id: 2,
+          table_id: TABLE_ID,
+          display_name: "B",
+          base_type: TYPE.Float,
+        }),
+        createMockField({
+          id: 3,
+          table_id: TABLE_ID,
+          display_name: "C",
+          base_type: TYPE.Float,
+        }),
+        createMockField({
+          id: 10,
+          table_id: TABLE_ID,
+          display_name: "Toucan Sam",
+          base_type: TYPE.Float,
+        }),
+        createMockField({
+          id: 11,
+          table_id: TABLE_ID,
+          display_name: "Sum",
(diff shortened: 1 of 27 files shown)


frontend/src/metabase-lib/expressions/suggest.ts
```diff
@@ -149,49 +154,66 @@ export function suggest({
...
-          index: targetOffset,
-          icon: "segment",
-          order: 3,
-        })),
+        ...segments.map(segment => {
+          // TODO: fix type
+          const displayInfo = Lib.displayInfo(
+            query,
+            stageIndex,
+            segment as any,
```
uladzimirdev, line 185: displayInfo returns correct value, so probably we need to support this type in displayInfo?

ranquild (approved):
Looks good! Please fix types and it should be ready to go

frontend/src/metabase/query_builder/components/notebook/types.ts
```diff
@@ -52,6 +52,7 @@ export interface NotebookStepAction {
 export interface NotebookStepUiComponentProps {
   step: NotebookStep;
   topLevelQuery: Query;
+  stageIndex: number;
```
kamilmielnik, line 55: Can't we just use `step.stageIndex`? It seems that they're always the same.

It feels weird to have 2 sources of this information.

frontend/src/metabase/query_builder/components/expressions/ExpressionWidget.unit.spec.tsx
```diff
@@ -163,7 +163,8 @@ function setup(additionalProps?: Partial<ExpressionWidgetProps>) {
     onChangeExpression: jest.fn(),
   };
 
-  const props = {
+  // we test only legacy queries for now
+  const props: Omit<ExpressionWidgetProps, "query" | "stageIndex"> = {
```
kamilmielnik, line 167: FYI I added tests for MLv2 query in my PR: https://github.com/metabase/metabase/pull/36031/files#diff-6996a8ef047a33b549cc21b76fea281c75d191834be45cf91d9e2b31ce3b2cb5R203

frontend/src/metabase/query_builder/components/expressions/ExpressionWidget.tsx
```diff
@@ -27,7 +28,17 @@ const EXPRESSIONS_DOCUMENTATION_URL = MetabaseSettings.docsUrl(
   "questions/query-builder/expressions",
 );
 
-export interface ExpressionWidgetProps {
+interface LegacyQueryProps {
```
kamilmielnik, line 31: We could move `expression` and `onChangeExpression` props into this interface

frontend/src/metabase-lib/expressions/__support__/expressions.js
```diff
@@ -6,104 +6,118 @@ import {
...
+    "percentile-aggregations",
+    "foreign-keys",
+    "native-parameters",
+    "expressions",
+    "advanced-math-expressions",
+    "right-join",
+    "left-join",
+    "inner-join",
+    "nested-queries",
+    "advanced-math-expressions",
```
kamilmielnik, line 115: This is specified twice in the array (I noticed it when I tried to see how difficult it is to convert this file to TS).

frontend/src/metabase-lib/expressions/suggest.unit.spec.ts
```diff
@@ -61,6 +78,8 @@ const FIELDS_CUSTOM_NON_NUMERIC = [
   { type: "fields", text: "[text] " },
 ];
 
+const DEFAULT_QUERY = DEFAULT_EXPRESSION_QUERY as DatasetQuery;
```
kamilmielnik, line 81: It's trivial to convert `__support__/expressions.js` to TS and then we wouldn't need this cast.

frontend/src/metabase/query_builder/components/expressions/ExpressionEditorTextfield/ExpressionEditorTextfield.tsx
```diff
@@ -401,17 +414,27 @@ class ExpressionEditorTextfield extends React.Component<
...
     const { source } = this.state;
     const { suggestions, helpText } = suggest({
-      legacyQuery,
       reportTimezone,
       startRule,
       source,
       targetOffset: cursor.column,
+      // TODO: uladzimir
+      // using stage index apart from -1 is possible only in the notebook editor, filter pills, and the filter modal
+      // first 2 are migrated in the filters integration branch. The latter is in progress
```
kamilmielnik, line 431: Should `mlv2/custom-columns-milestone-1` be based on `34112-filters-milestone1` then?

kamilmielnik (commented):
Looks good :+1: I left a few minor comments.

frontend/src/metabase-lib/expressions/suggest.ts
```diff
@@ -149,49 +154,62 @@ export function suggest({
...
-          icon: "segment",
-          order: 3,
-        })),
+        ...segments.map(segment => {
+          const displayInfo = Lib.displayInfo(query, stageIndex, segment);
+
+          return {
+            type: "segments",
+            name: displayInfo.displayName,
+            text: formatIdentifier(displayInfo.displayName),
```
kamilmielnik, line 187: What is the reason `longDisplayName` is not used here?

uladzimirdev: no reason, indeed

frontend/src/metabase/query_builder/components/expressions/ExpressionEditorTextfield/ExpressionEditorTextfield.tsx
uladzimirdev: probably we'll not be able to merge this milestone to master, but we need to figure out if fallback to legacy query in suggest works. if so, we're good to go

frontend/src/metabase-lib/expressions/suggest.unit.spec.ts
uladzimirdev: I had an idea to unify this file with another as we have 2 files with fixtures. Good point

frontend/src/metabase-lib/expressions/__support__/expressions.js
uladzimirdev: thanks! please take a look https://github.com/metabase/metabase/pull/36265

frontend/src/metabase-lib/expressions/suggest.unit.spec.ts
uladzimirdev: done https://github.com/metabase/metabase/pull/36265

frontend/src/metabase/query_builder/components/expressions/ExpressionWidget.tsx
uladzimirdev: you mean those props make sense only with legacyQuery?

frontend/src/metabase/query_builder/components/notebook/types.ts
uladzimirdev: what do you think about unification like this https://github.com/metabase/metabase/pull/36267 ?

ranquild (approved):

frontend/src/metabase-lib/expressions/suggest.ts
ranquild: I'd check how it worked in MLv1 and compare. Hard to say

Issue #4142, opened 2017-01-12
No edit icon after edit question


anki-code:
Regress in 0.22:

![image](https://cloud.githubusercontent.com/assets/1708680/21894731/5021a90a-d8f1-11e6-9029-b5dec2f8dd00.png)

tlrobinson:
Looks like this is due to the addition of the `can_write` attribute returned by `GET /api/card/:id` but not `PUT /api/card/:id`. @camsaul could we have the PUT endpoint response include everything the GET endpoint does?

camsaul:
Yeah I can do that

camsaul:
Fixed by #4164

PR #36123, opened 2023-11-27
Move functions for creating and updating cards from api to models namespace

calherries:
This PR moves the functions `create-card!` and `update-card!` and their helper functions from `metabase.api.card` to `metabase.models.card`.

### Why?

Outside of `POST /api/card`, `create-card!` was being reused by the endpoint for uploading CSVs `POST /api/card/from-csv`, and , suggesting its logic should be independent of its usage in a specific API endpoint. It was also being used by `POST "/:from-dashboard-id/copy` to create cards, and in this case it was doing unnecessary hydration.

But the reason why this change is being done now, is so that `metabase.api.card/upload-csv!` can be moved out of `metabase.api.card` to be in a separate `uploads` namespace. The only reason why it's in `metabase.api.card` is because it needs to use `metabase.api.card/create-card!`. This allows separation of the `uploads` code from the api namespaces.

### Circular dependencies

These changes caused a few circular dependencies cycles that were also resolved as part of this PR.

### What about tests?

These functions are untested in `metabase.api.card-test` (only the endpoints are), so I have chosen not to test them in `metabase.models.card-test` (they probably should be, though this PR doesn't cover that).


src/metabase/api/card.clj
```diff
@@ -158,6 +148,23 @@
                      card)))
             cards))))
 
+(defn hydrate-for-frontend
```
calherries, line 151: I extracted the common parts out of the endpoints that return a card: `GET, POST, and PUT /api/card`, and also `POST /api/card/from-csv`

src/metabase/api/card.clj
```diff
@@ -548,7 +411,7 @@ saved later when it is ready."
   (check-data-permissions-for-query dataset_query)
   ;; check that we have permissions for the collection we're trying to save this card to, if applicable
   (collection/check-write-perms-for-collection collection_id)
-  (create-card! body @api/*current-user*))
+  (hydrate-for-frontend (card/create-card! body @api/*current-user*)))
```
calherries, line 414: Now hydrating cards is separate from creating them

src/metabase/api/card.clj
```diff
@@ -775,26 +470,26 @@ saved later when it is ready."
...
       (f card-before-update card-updates))
     ;; make sure we have the correct `result_metadata`
-    (let [result-metadata-chan  (result-metadata-async {:original-query    (:dataset_query card-before-update)
-                                                        :query             dataset_query
-                                                        :metadata          result_metadata
-                                                        :original-metadata (:result_metadata card-before-update)
-                                                        :dataset?          (if (some? dataset)
-                                                                             dataset
-                                                                             (:dataset card-before-update))})
+    (let [result-metadata-chan  (card/result-metadata-async {:original-query    (:dataset_query card-before-update)
```
calherries, line 473: This timeout stuff should probably be extracted into a function in the `metabase.models.card` namespace too, but I didn't want to do too much at once

src/metabase/api/dashboard.clj
```diff
@@ -984,7 +984,7 @@
    parameters   ms/JSONString}
   (api/read-check :model/Dashboard dashboard-id)
   (actions.execution/fetch-values
-   (api/check-404 (dashboard-card/dashcard->action dashcard-id))
+   (api/check-404 (action/dashcard->action dashcard-id))
```
calherries, line 987: I moved this function to resolve a circular dependency

src/metabase/email/messages.clj
```diff
@@ -322,7 +321,7 @@
            :titleUrl                  (params/dashboard-url dashboard-id (params/parameters pulse dashboard))
            :dashboardDescription      (:description dashboard)
            ;; There are legacy pulses that exist without being tied to a dashboard
-           :dashboardHasTabs          (when dashboard-id (dashboard/has-tabs? dashboard-id))
+           :dashboardHasTabs          (when dashboard-id (boolean (seq (t2/hydrate dashboard-id :tabs))))
```
calherries, line 324: Resolving circular deps

src/metabase/models/action.clj
```diff
@@ -341,6 +341,12 @@
     (for [dashcard dashcards]
       (m/assoc-some dashcard :action (get actions-by-id (:action_id dashcard))))))
 
+(defn dashcard->action
+  "Get the action associated with a dashcard if exists, return `nil` otherwise."
+  [dashcard-or-dashcard-id]
+  (some->> (t2/select-one-fn :action_id :model/DashboardCard :id (u/the-id dashcard-or-dashcard-id))
+           (select-action :id)))
```
calherries, line 348: I moved this function from `metabase.models.dashboard-card` to resolve circular deps.

src/metabase/api/card.clj
```diff
@@ -158,6 +148,22 @@
                      card)))
             cards))))
 
+(defn hydrate-for-frontend
+  "Adds additional information to a `Card` selected with toucan that is needed by the frontend. This should be the same information
+  returned by all API endpoints where the card entity is cached (i.e. GET, PUT, POST) since the frontend replaces the Card
+  it currently has with returned one -- See #4142."
```
qnkhuat, line 154: https://github.com/metabase/metabase/issues/4142 ?

src/metabase/api/dashboard.clj
```diff
@@ -7,7 +7,6 @@
    [medley.core :as m]
    [metabase.actions.execution :as actions.execution]
    [metabase.analytics.snowplow :as snowplow]
-   [metabase.api.card :as api.card]
```
qnkhuat, line 10: nice, love seeing these cross API imports get removed.

src/metabase/models/card.clj
```diff
@@ -473,6 +488,292 @@
...
+                                (api/maybe-reconcile-collection-position! card-data)
+                                (first (t2/insert-returning-instances! Card (cond-> card-data
+                                                                              (and metadata (not timed-out?))
+                                                                              (assoc :result_metadata metadata)))))]
+     (when-not delay-event?
+       (events/publish-event! :event/card-create {:object card :user-id (:id creator)}))
+     (when timed-out?
+       (log/info (trs "Metadata not available soon enough. Saving new card and asynchronously updating metadata")))
+     ;; include same information returned by GET /api/card/:id since frontend replaces the Card it currently has with
+     ;; returned one -- See #4283
```
qnkhuat, line 618:
```suggestion
```

src/metabase/models/card.clj
```diff
@@ -473,6 +488,292 @@
...
+  query."
+  #{:table_id
+    :database_id
+    :query_type ;; these first three may not even be changeable
+    :dataset_query})
+
+(defn update-card!
+  "Update a Card. Metadata is fetched asynchronously. If it is ready before [[metadata-sync-wait-ms]] elapses it will be
+  included, otherwise the metadata will be saved to the database asynchronously."
+  [{:keys [id] :as card-before-update} card-updates]
```
qnkhuat, line 746: wdyt about adding `actor-id` as part of this function's parameters? it'll make this function free of dynamic vars

src/metabase/models/card.clj
```diff
@@ -473,6 +488,292 @@
...
+  the alert"
+  [{:keys [display] :as new-card}]
+  (and (get-in new-card [:visualization_settings :graph.goal_value])
+       (or (line-area-bar? display)
+           (progress? display))
+       (< 1 (count (get-in new-card [:dataset_query :query :breakout])))))
+
+(defn- delete-alert-and-notify!
+  "Removes all of the alerts and notifies all of the email recipients of the alerts change via `NOTIFY-FN!`"
+  [notify-fn! alerts]
```
qnkhuat, line 674: same here, maybe this could take `actor` as parameter as well

qnkhuat (approved):
Cool, I attempted this once a while ago but didn't follow it through, glad to see it's being moved now.

I have some suggestions to avoid using dynamic vars, but shouldn't be blocking.

src/metabase/models/card.clj
calherries: Yes, good idea [Ngoc's suggestions](https://github.com/metabase/metabase/pull/36123/commits/03ff937ed888e550c23d6e6bf2aa1ee9c7e5d6b7)

PR #36128, opened 2023-11-27
Bug fix: remove divider in QueryBuilder/ViewHeader when it's not needed

oleggromov:
### Description
There's a tiny issue with the header actions panel: when we don't have anything to the left from the action buttons, the divider is just dangling out there.

Before | After
---|---
Screenshot 2023-11-27 at 14 10 52 | Screenshot 2023-11-27 at 14 09 31

### How to verify

1. Create a new SQL question
2. Save it
3. You should see the "Explore results" button **and** a divider, which is correct
4. Add a parameter to it (`where id = {{id}}`) and you should see no button and no divider

### Tests
I don't think it's worth testing in an automated way.

### DIFF
--- a/frontend/src/metabase/query_builder/components/QuestionActions.styled.tsx
+++ b/frontend/src/metabase/query_builder/components/QuestionActions.styled.tsx
@@ -7,6 +7,10 @@ export const QuestionActionsDivider = styled.div`
   margin-left: 0.5rem;
   margin-right: 0.5rem;
   height: 1.25rem;
+
+  &:first-child {
+    display: none;
+  }
 `;
 
 export const StrengthIndicator = styled(DatasetMetadataStrengthIndicator)`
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/query_builder/components/QuestionActions.styled.tsx
```diff
@@ -7,6 +7,10 @@ export const QuestionActionsDivider = styled.div`
   margin-left: 0.5rem;
   margin-right: 0.5rem;
   height: 1.25rem;
+
+  &:first-child {
```
ranquild, line 11: `emotion` should complain about this AFAIK. Have you checked the dev console?

ranquild (commented):
Would it possible to hide the divider with code?

oleggromov:
> Would it possible to hide the divider with code?
I thought about it and unfortunately, it's not that easy. There's a [handful of things](https://github.com/metabase/metabase/blob/91db87a9ab4cd38680212dea560cbcb213c060f2/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx#L446-L504) that could be displayed to the left from the divider, and the conditional will be massive — and I can almost guarantee it will get broken or incomplete after someone changes something there.

But the main question: why? Are discouraging CSS exactly where it should be used, in the presentation detail?

ranquild (approved):
Please create another PR for SSR changes

PR #36134, opened 2023-11-27
Emotion: suppress SSR warnings

oleggromov:
### Description
While working on the https://github.com/metabase/metabase/pull/36128, we realised that emotion spits out useless warnings about `:first-child` being unsafe in SSR environments, which, I believe, we won't do. 

### How to verify
1. Create an SQL question
2. Open chrome dev tools, see the console tab
3. On master, you should see the `first-child` warning from emotion - on this branch, you shouldn't.

Screenshot 2023-11-27 at 16 51 55

### DIFF
--- a/frontend/src/metabase/styled-components/components/EmotionCacheProvider/EmotionCacheProvider.tsx
+++ b/frontend/src/metabase/styled-components/components/EmotionCacheProvider/EmotionCacheProvider.tsx
@@ -14,6 +14,7 @@ export const EmotionCacheProvider = ({
     () => createCache({ key: "emotion", nonce: window.MetabaseNonce }),
     [],
   );
+  emotionCache.compat = true;
 
   return <CacheProvider value={emotionCache}>{children}</CacheProvider>;
 };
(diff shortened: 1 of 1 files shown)


uladzimirdev:
@oleggromov could you please share some docs, why this option will fix the warning? quick googling didn't help

uladzimirdev:
alright, I think I found it here https://github.com/emotion-js/emotion/issues/1105#issuecomment-557726922

uladzimirdev (approved):

frontend/src/metabase/styled-components/components/EmotionCacheProvider/EmotionCacheProvider.tsx
```diff
@@ -14,6 +14,7 @@ export const EmotionCacheProvider = ({
     () => createCache({ key: "emotion", nonce: window.MetabaseNonce }),
     [],
   );
+  emotionCache.compat = true;
```
uladzimirdev, line 17: nit: `emotionCache.compat = true;` can be moved inside `useMemo`

oleggromov:
Thanks! I've addressed both and will merge when it's green

PR #36135, opened 2023-11-27
replace dc.js line/area/bar/combo charts with newer echarts implementation without interactivity

alxnddr:
Closes https://github.com/metabase/metabase/issues/36183

### Description

Basic setup for dynamic combo chart without interactivity.

### How to verify

Open different line/area/bar/combo charts and ensure they look ok in QB and dashboards.

### Demo

Charts on video have slightly different styles because I unlinked zrender package locally

https://github.com/metabase/metabase/assets/14301985/f44df901-1089-41c2-b317-480fcef10e89

### Checklist


frontend/src/metabase/visualizations/components/EChartsRenderer/ResponsiveEChartsRenderer.tsx
EmmadUsmani, line 1: I don't really see a point in wrapping the existing `EChartsRenderer` in a new component, I think everywhere we use it we want it to be responsive.

frontend/src/metabase/visualizations/visualizations/AreaChart/AreaChart.tsx
```diff
@@ -0,0 +1,54 @@
...
+    uiName: t`Area`,
+    identifier: "area",
+    iconName: "area",
+    noun: t`area chart`,
+    minSize: getMinSize("area"),
+    defaultSize: getDefaultSize("area"),
+    settings: {
+      ...LINE_SETTINGS,
+      ...STACKABLE_SETTINGS,
+      ...GRAPH_GOAL_SETTINGS,
```
EmmadUsmani, line 36: Nit: A lot of these are the same across different Cartesian chart types, can we move the common settings to `getCartesianChartDefinition`?

frontend/src/metabase/visualizations/visualizations/AreaChart/AreaChart.tsx
EmmadUsmani, line 1: Can we move this file and the other visualizations within the `echarts` folder? I don't like decoupling code that is used to together, code that's used together should live together IMO (static viz can be an exception since it's a different bundle). For example we can put it in `echarts/cartesian/AreaChart`

frontend/src/metabase/visualizations/visualizations/AreaChart/AreaChart.tsx
```diff
@@ -0,0 +1,54 @@
...
+      ...GRAPH_TREND_SETTINGS,
+      ...GRAPH_COLORS_SETTINGS,
+      ...GRAPH_AXIS_SETTINGS,
+      ...GRAPH_DISPLAY_VALUES_SETTINGS,
+      ...GRAPH_DATA_SETTINGS,
+    } as any as VisualizationSettingsDefinitions,
+    onDisplayUpdate: (settings: ComputedVisualizationSettings) => {
```
EmmadUsmani, line 43: We should also update the value of `display` in series settings, so you can't end up with other series types when you change it to area. The same goes for the other cartesian charts.

frontend/src/metabase/visualizations/echarts/cartesian/option/index.ts
```diff
@@ -35,6 +35,7 @@ export const getCartesianChartOption = (
   ];
 
   return {
+    animation: false,
```
EmmadUsmani, line 38: We can leave this for now, but we should have a TODO item in the future for consulting with product/design regarding what animations to keep and which ones to toss. Some animations like the transition between y-axis scales (linear, log, pow) I think are pretty neat in seeing how the data changes and are fairly unobtrusive.

EmmadUsmani (approved):

frontend/src/metabase/visualizations/visualizations/AreaChart/AreaChart.tsx
alxnddr: Not sure if I agree with moving because this would create inconsistency because all our visualization definitions are in `visualizations` folder — maps built with leaflet, tables, visx charts. But I agree that internal ECharts specific logic should be placed in the `echarts` folder

PR #36138, opened 2023-11-27
Tweak ReplayIO E2E runs

nemanjaglumac:
This PR fixes the E2E Replay workflow.
It also increases the frequency of these CI runs. This should give us much better overview and understanding of the potential overhead.

### Note
The test run ID doesn't need to be a UUID in the recent Replay runner versions. I upgraded to the latest version.

### DIFF
--- a/.github/workflows/e2e-tests-replay.yml
+++ b/.github/workflows/e2e-tests-replay.yml
@@ -3,7 +3,7 @@ name: E2E Tests using Replay.io browser
 on:
   # We'll record runs using Replay.io and their browser on a schedule as an experiment
   schedule:
-    - cron: '0 22 * * *'
+    - cron: '0 */3 * * *'
 
 jobs:
   test-run-id:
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-tests-replay.yml
```diff
@@ -90,23 +82,23 @@ jobs:
...
+        if: matrix.name == 'slow'
         run: |
           yarn run test-cypress-run \
           --env grepTags="@slow",grepOmitFiltered=true \
           --spec './e2e/test/scenarios/**/*.cy.spec.js' \
           --browser "replay-chromium"
 
       - name: Run EE Cypress tests on ${{ matrix.name }}
-        if: matrix.context == 'folder' && github.event_name != 'schedule'
+        if: matrix.context == 'folder'
```
nemanjaglumac, line 101: This was my mistake during #34895

ryanjduffy (commented):
If you'd like, you can also remove the upload step from your YML and instruct the plugin to upload replays.

```
    if (runWithReplay) {
      replay.default(on, config, {
        upload: true,
        apiKey: process.env.REPLAY_API_KEY
      });
    }

iethree (approved):

PR #36140, opened 2023-11-27
feat: use InputWithSelectPrefix for the url, + allow pasting

npretto:
[🛠️ Epic](https://github.com/metabase/metabase/issues/35915)
[📝 Product Doc](https://www.notion.so/metabase/White-labeling-Customizing-or-hiding-Help-link-8840fbdfd3c743ccaff9301a835ec81c?pvs=4)

This adds the "InputWithSelectPrefix" to help the user write a valid destination url.

I converted the file to ts and added a couple things:
- when the text input is empty and the prefix is changed -> no `onChange` call: use case is that a user wants to use a different prefix and changes it before starting writing, we don't need to call the onChange, it would only call whatever endpoint and receive a validation error
- if the user pastes or manually writes a url with a different prefix, it now detects that and changes it. Product/design wants people to just paste a url which makes sense


frontend/src/metabase/components/InputWithSelectPrefix/InputWithSelectPrefix.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  componentDidUpdate(prevProps: InputWithSelectPrefixProps) {
+    if (prevProps.value !== this.props.value) {
+      this.setPrefixAndRestFromValue();
+    }
+  }
+
+  handleManualPrefixWriting(e: ChangeEvent<HTMLInputElement>) {
+    const value = e.target.value;
+    for (const prefix of this.props.prefixes) {
```
npretto, line 71: This allows people to paste (or write) a full url with a different prefix

frontend/src/metabase/components/InputWithSelectPrefix/InputWithSelectPrefix.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const { prefixes, defaultPrefix, autoFocus, id } = this.props;
+    const { prefix, rest } = this.state;
+    return (
+      <div className="flex align-stretch SettingsInput Form-input p0" id={id}>
+        <Select
+          className="border-right"
+          value={prefix || defaultPrefix}
+          onChange={(e: SelectChangeEvent<string>) => {
+            this.setState({ prefix: e.target.value });
+            if (this.state.rest !== "") {
```
npretto, line 99: If the user just changes protocol with an empty text input, we don't want to trigger an update as we already know it's an invalid input

frontend/src/metabase/components/InputWithSelectPrefix/InputWithSelectPrefix.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const { value } = this.props;
+
+    if (value) {
+      const [prefix, rest] = splitValue(this.props);
+      this.setState({ prefix, rest });
+    }
+  }
+
+  componentDidUpdate(prevProps: InputWithSelectPrefixProps) {
+    if (prevProps.value !== this.props.value) {
```
npretto, line 64: Previously the onChange was called here, checking if the state changed.
I removed it as it's pretty much the same [as this anti-pattern](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers).
(It also gave me some practical issue , i don't remember if it was with the not calling onChange with an empty `rest` or with the manual prefix handling)

frontend/src/metabase/components/InputWithSelectPrefix/InputWithSelectPrefix.tsx
```diff
@@ -0,0 +1,127 @@
+import type { ChangeEvent } from "react";
```
npretto, line 1: I converted this from js to ts, actual diff is here:
```diff tsx
diff --git a/tmp.tsx b/tmp.tsx
index b007b4857c..60b4f784a9 100644
--- a/tmp.tsx
+++ b/tmp.tsx
@@ -1,6 +1,6 @@
+import type { ChangeEvent } from "react";
 import { Component } from "react";
-import PropTypes from "prop-types";
-
+import type { SelectChangeEvent } from "metabase/core/components/Select";
 import Select, { Option } from "metabase/core/components/Select";
 import { SelectPrefixInput } from "./InputWithSelectPrefix.styled";
 
@@ -9,7 +9,12 @@ function splitValue({
   prefixes,
   defaultPrefix,
   caseInsensitivePrefix = false,
-}) {
+}: {
+  value?: string;
+  prefixes: string[];
+  defaultPrefix: string;
+  caseInsensitivePrefix?: boolean;
+}): string[] {
   if (value == null) {
     return ["", ""];
   }
@@ -23,17 +28,23 @@ function splitValue({
   return prefix ? [prefix, value.slice(prefix.length)] : [defaultPrefix, value];
 }
 
-const propTypes = {
-  value: PropTypes.string,
-  prefixes: PropTypes.arrayOf(PropTypes.string),
-  defaultPrefix: PropTypes.string,
-  caseInsensitivePrefix: PropTypes.bool,
-  onChange: PropTypes.func,
-  placeholder: PropTypes.string,
+type InputWithSelectPrefixProps = {
+  value?: string;
+  prefixes: string[];
+  defaultPrefix: string;
+  caseInsensitivePrefix?: boolean;
+  onChange: (e: { target: { value: string } }) => void;
+  placeholder?: string;
+  autoFocus?: boolean;
+  id?: string;
 };
 
-export default class InputWithSelectPrefix extends Component {
-  constructor(props) {
+// eslint-disable-next-line import/no-default-export
+export default class InputWithSelectPrefix extends Component<
+  InputWithSelectPrefixProps,
+  { prefix: string; rest: string }
+> {
+  constructor(props: InputWithSelectPrefixProps) {
     super(props);
 
     const [prefix, rest] = splitValue(props);
@@ -49,26 +60,46 @@ export default class InputWithSelectPrefix extends Component {
     }
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    const { prefix, rest } = this.state;
-    if (prevState.rest !== rest || prevState.prefix !== prefix) {
-      const value = prefix + rest;
-      this.props.onChange({ target: { value } });
-    }
+  componentDidUpdate(prevProps: InputWithSelectPrefixProps) {
     if (prevProps.value !== this.props.value) {
       this.setPrefixAndRestFromValue();
     }
   }
 
+  handleManualPrefixWriting(e: ChangeEvent<HTMLInputElement>) {
+    const value = e.target.value;
+    for (const prefix of this.props.prefixes) {
+      if (value.startsWith(prefix)) {
+        const newValue = { prefix, rest: value.slice(prefix.length) };
+        this.setState(newValue);
+        if (newValue.rest !== "") {
+          this.handleChange(newValue.prefix, newValue.rest);
+        }
+        return;
+      }
+    }
+  }
+
+  handleChange(prefix: string, rest: string) {
+    this.props.onChange({
+      target: { value: prefix + rest },
+    });
+  }
+
   render() {
-    const { prefixes, defaultPrefix } = this.props;
+    const { prefixes, defaultPrefix, autoFocus, id } = this.props;
     const { prefix, rest } = this.state;
     return (
-      <div className="flex align-stretch SettingsInput Form-input p0">
+      <div className="flex align-stretch SettingsInput Form-input p0" id={id}>
         <Select
           className="border-right"
           value={prefix || defaultPrefix}
-          onChange={e => this.setState({ prefix: e.target.value })}
+          onChange={(e: SelectChangeEvent<string>) => {
+            this.setState({ prefix: e.target.value });
+            if (this.state.rest !== "") {
+              this.handleChange(e.target.value, this.state.rest);
+            }
+          }}
           buttonProps={{ className: "borderless" }}
         >
           {prefixes.map(p => (
@@ -79,15 +110,18 @@ export default class InputWithSelectPrefix extends Component {
         </Select>
         <SelectPrefixInput
           type="text"
+          autoFocus={autoFocus}
           className="flex-full"
           value={rest}
+          onChange={this.handleManualPrefixWriting.bind(this)}
           placeholder={this.props.placeholder}
-          onBlurChange={e => this.setState({ rest: e.target.value })}
+          onBlurChange={e => {
+            this.setState({ rest: e.target.value });
+            this.handleChange(prefix, e.target.value);
+          }}
           size="large"
         />
       </div>
     );
   }
 }
-
-InputWithSelectPrefix.propTypes = propTypes;
```

src/metabase/public_settings.clj
```diff
@@ -114,18 +114,32 @@
...
   ;; TODO: the settings on the BE will be done at the end, with help from the backend team
-  (defsetting help-link
+(defsetting help-link
   (deferred-tru "TODO")
   :default    :metabase_default
   :type       :keyword
   :audit      :getter
   :visibility :public)
 
+(defn- normalize-help-link-custom-destination [^String s]
```
npretto, line 124: This is wip code just to receive an error, the error will actually be handled in the next PR.
Real BE validation should come later with help from the webapp admin team

PR #36152, opened 2023-11-27
Filter values leaked in API request on Embedding

qwef:
Closes https://github.com/metabase/metabase-private/issues/58

Fixes issue with filter values being leaked in API request on Embedding. To fix the issue with a linked parameter to a locked parameter still showing all the filters, we decided on an approach to remove all the values of that linked parameter, but keep the filter bar.


src/metabase/api/embed.clj
```diff
@@ -251,18 +251,50 @@
...
+                                     :when              (contains? params-ids-to-remove (:parameter_id parameter-mappings))]
+                                 (-> (:target parameter-mappings)
+                                     (nth 1)
+                                     (nth 1)))))
+        remove-parameters (fn [dashcard]
+                            (update dashcard :parameter_mappings
+                                    (fn [param-mappings]
+                                      (into []
+                                       (filter (fn [{:keys [parameter_id]}]
+                                                 (not (contains? params-ids-to-remove parameter_id)))
```
escherize, line 277: It's usually better to use `remove` than `(filter ... (not ...`

src/metabase/api/embed.clj
```diff
@@ -251,18 +251,50 @@
...
+                                               param-mappings)))))]
+    (-> dashboard
+        (assoc :dashcards (map remove-parameters (:dashcards dashboard)))
+        (assoc :param_fields (apply (partial dissoc (:param_fields dashboard)) field-ids-to-remove))
```
escherize, line 281: Usually when the inside of an `assoc` references the thing itself, `update` would work better.

I'd split these up too, since it's a little easier to read.

```suggestion
        (update :param_fields dissoc (:param_fields dashboard)))
        (update :param_fields dissoc field-ids-to-remove))
```

escherize, line 281: Usually when the inside of an `assoc` references the thing itself, `update` would work better.

```suggestion
        (update :param_fields dissoc field-ids-to-remove))
```

src/metabase/api/embed.clj
```diff
@@ -251,18 +251,50 @@
...
+        params-ids-to-remove (set (for [parameter (:parameters dashboard)
+                                        :when     (contains? params-to-remove (keyword (:slug parameter)))]
+                                    (:id parameter)))
+        field-ids-to-remove (flatten
+                             (for [card (:dashcards dashboard)]
+                               (for [parameter-mappings (:parameter_mappings card)
+                                     :when              (contains? params-ids-to-remove (:parameter_id parameter-mappings))]
+                                 (-> (:target parameter-mappings)
+                                     (nth 1)
+                                     (nth 1)))))
```
escherize, line 271: you can put multiple things inside of `for`, this should save the call to flatten, too:

```clojure
(for [x [[1 10] [2 20] [3 30]]]
  (for [y x]
    y))
;; => ((1 10) (2 20) (3 30))

(for [x [[1 10] [2 20] [3 30]]
      y x]
  y)
;; => (1 10 2 20 3 30)
```

I didn't run it, but I think this is still the same.

```suggestion
        field-ids-to-remove (for [card (:dashcards dashboard)
                                  parameter-mappings (:parameter_mappings card)
                                  :when (contains? params-ids-to-remove (:parameter_id parameter-mappings))]
                              (ffirst (:target parameter-mappings)))
```

dpsutton:
Can I suggest an alternative implementation?

We have a namespace `metabase.models.params`. What if we added a function there that took param names and returned the field ids backing those params?

The logic is almost there:

```clojure
;; in metabase.models.params
params=> (dashcards->param-field-ids (:dashcards user/d))
#{58 60}
```

The idea is that we can get the locked params:

```clojure
embed=> (t2/select-one-fn :embedding_params Dashboard, :id 20)
{:vendor "locked", :custom-list "locked"}
;; then what if we could get the fields backing these params
;; and we could just call it like so:
(let [fields-backing (params/fields-backing-params dashboard #{:vendor :custom-list})]
  (update dashboard :param_values dissoc fields-backing))
```

Or perhaps `fields-backing-params` would return some map like

```clojure
{:vendor 60 :custom-list 58}
```

And we dissoc the keys from it?

src/metabase/api/embed.clj
qwef: This isn't working for me since `field-ids-to-remove` is a list, is there a way to expand `field-ids-to-remove` to dissoc all elements within?

escherize (changes requested):
I think the overall approach makes sense. I asked for some more tests, some stylistic changes, and to simplify some of the intermediate steps.

src/metabase/api/embed.clj
```diff
@@ -28,6 +28,7 @@
    [metabase.driver.common.parameters.operators :as params.ops]
    [metabase.models.card :as card :refer [Card]]
    [metabase.models.dashboard :refer [Dashboard]]
+   [metabase.models.params :refer [dashcards->param-to-field-ids-map]]
```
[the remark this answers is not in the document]
qwef: Ah I originally didn't do this because params aliased namespace was already being used, and changing it to something that wasn't params gave a kondo warning. Should I just suppress the kondo error?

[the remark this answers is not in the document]
qwef: Change isn't too big, will add in this pr

escherize (approved):
Nice work! I see how this would elide locked filters from dashboard/:id/, even if they're linked filters.

PR #36179, opened 2023-11-28
Fix event logging to print correct output

markbastian:
The fix in #35895 included reducing the size of the printed object, but the drilled elements needed to be pulled from the object key in the event, not the event itself. This fixes that.


adam-james-v (approved):
Looks good

PR #36181, opened 2023-11-28
Static Embedding Modal redesign

deniskaber:
Part of a milestone 5 from https://github.com/metabase/metabase/issues/35961

### Description

First redesign iteration for Static embed setup Modal - https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=129-1873&mode=dev

### How to verify

Prerequisite: 
Enable embedding (http://localhost:3000/admin/settings/embedding-in-other-applications > Enable embedding)

Question:
1. Create a question, save it, then try to share it.
2. Select "Static embedding" in the sharing modal

Dashboard:
1. Create a dashboard, add filters, save it, then try to share it
2. Select "Static embedding" in the sharing modal

### Demo

OSS - https://www.loom.com/share/e90495eabbf149b893a3bb9c89a70e5a
EE - https://www.loom.com/share/a86ddd6c06bc4861b621211b77bc874c

### Checklist


WiNloSt:
Some visual/behavior feedback.
1. `line-height` seems different, _Overview_ and _Appearance_ tab seem to have the line-height set to `24px` while the `parameters` doesn't have line-height set at all

| Overview                                                                                           | Parameter                                                                                          |
|----------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|
| image | image |
2. We could take the opportunity to rework the preview a bit. I think I've mentioned in your previous PR that we could make the preview instant without having to force the iframe to reload. That'll make the preview experience more polished. Now it feels a bit sluggish changing between different options especially when dashboards have so many questions it could take longer time to load making the whole situation worse.
3. When switching between _Parameters_ and _Apperance_ tabs the right pane where it hosts code/preview panel seems to be rerendering. This is even more apparent when previewing the embed and switching between these _Parameters_ and _Appearance_ tabs

vbenedetti:
Hi Denis, I found a few visual issues that I wish could be fixed before approval. I added them in [this doc](https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=957-4776&mode=design&t=ANGbHigXiOU5DkYN-11)

Thanks!

frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
```diff
@@ -38,15 +40,18 @@ export const DashboardSharingEmbeddingModal = (
   const updateDashboardEmbeddingParams = (embeddingParams: EmbedOptions) =>
     dispatch(updateEmbeddingParams(dashboard, embeddingParams));
 
-  const getPublicUrl = ({ public_uuid }: { public_uuid: string }) =>
-    Urls.publicDashboard(public_uuid);
+  const getPublicUrl = ({ public_uuid }: EmbedResource) => {
```
WiNloSt, line 43: I gave a similar comment in `frontend/src/metabase/query_builder/components/QuestionEmbedWidget/QuestionEmbedWidget.tsx` Please take a look there.

tl;dr I think we should keep the signature as is `public_uuid` to always be string.

frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onClose={onEmbedClose}
+      title={!isEmbeddingStage ? t`Embed ${applicationName}` : undefined}
+      fit
+      formModal={false}
+    >
+      <Box bg={isEmbeddingStage ? color("white") : "bg.0"} h="100%">
```
WiNloSt, line 46: I wouldn't want to juggle styles and markups to accommodate some differences in the design. I'd rather stay more simpler.

Here's my attempt to simplify this

```ts
    <Modal
      isOpen={isOpen}
      onClose={onEmbedClose}
      title={
        !isEmbeddingStage ? (
          t`Embed ${applicationName}`
        ) : (
          <>
            <ModalContentActionIcon
              name="chevronleft"
              onClick={() => setEmbedType(null)}
            />
            {t`Static embedding`}
          </>
        )
      }
      fit
      formModal={false}
    >
      <Divider />
      {children({ embedType, setEmbedType })}
    </Modal>
```
| My version                                                                                           | This PR                                                                                          |
|----------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|
| image | image |


The visual difference is minimal, but we can remove some markup complexity.

This makes sure the `title` prop takes care of both title situation, so it should be easier to reason about. 
it also uses Mantine's `Divider` which is a very good component, so we don't need to create another styled component just to have a border, we can also remove `EmbedModalHeader` and its file because of that. And lastly, `children` prop doesn't have different parent containers (or parents with different styles which is a bit hard to read since it had a conditional style)

WiNloSt (commented):
This is part 1/x review. I'll review the rest later.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -0,0 +1,107 @@
...
+  parameterValues,
+  resourceParameters,
+  embeddingParams,
+  onChangeDisplayOptions,
+  onChangeEmbeddingParameters,
+  onChangeParameterValue,
+  onChangePane,
+}: StaticEmbedSetupPaneProps): JSX.Element => {
+  return (
+    <div className="full flex">
```
oisincoveney, line 42: Let's use `Group` and `Stack` here - I know you probably didn't write this in the first place but it's a quick win

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -0,0 +1,107 @@
...
+import type { OverviewSettingsProps } from "./OverviewSettings";
+import { OverviewSettings } from "./OverviewSettings";
+
+const TABS = {
+  Overview: "overview",
+  Parameters: "parameters",
+  Appearance: "appearance",
+};
+
+type StaticEmbedSetupPaneProps = OverviewSettingsProps &
```
oisincoveney, line 17: Do we have any other mechanisms for storing redundant information across components besides redux? I feel like this could be a good candidate for a provider or some sort of injection but I'm not totally sure in how that would affect the complexity/coupling here

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.styled.tsx
```diff
@@ -0,0 +1,89 @@
...
+    </ContentWrapper>
+  );
+};
+
+export const CODE_PREVIEW_CONTROL_OPTIONS = [
+  {
+    label: (
+      <Center>
+        <Icon name="embed" />
+        <Box ml="0.5rem">Code</Box>
```
oisincoveney, line 58: don't forget to `ttag` these 😄

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.styled.tsx
```diff
@@ -0,0 +1,89 @@
...
+  width: 100%;
+  display: flex;
+  align-items: stretch;
+
+  min-height: 648px;
+`;
+
+const SettingsAsideBlock = styled.div`
+  flex-shrink: 0;
+  width: 345px;
```
oisincoveney, line 17: Do we have better proxies for width here besides this really specific pixel value (similar to line 12 and 25 for this file too). At the very least, let's `rem` values, but maybe we can think about how wide we want that sidebar to be on different screen sizes and be a little smarter about it - maybe something like 30vw or a `min/max`. I wouldn't think about it _too_ much though, it's a low-value add

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/PreviewPane.styled.tsx
```diff
@@ -1,6 +1,9 @@
 import styled from "@emotion/styled";
 
 export const PreviewPaneContainer = styled.div<{ isTransparent?: boolean }>`
+  width: 100%;
+  min-height: 280px;
```
oisincoveney, line 5: convert this to REM plz

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
```diff
@@ -0,0 +1,203 @@
...
+              displayOptions={displayOptions}
+            />
+          ) : null}
+        </>
+      }
+    />
+  );
+};
+
+const getIconForParameter = (parameter: EmbedResourceParameter) =>
```
oisincoveney, line 198: I looked at this for 20 seconds and my brain hurts, maybe we can simplify this one a little?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
```diff
@@ -0,0 +1,203 @@
...
+        resourceParameters.length > 0 ? (
+          <>
+            <StaticEmbedSetupPaneSettingsContentSection
+              title={t`Enable or lock parameters`}
+            >
+              <Stack spacing="1rem">
+                <Text>{t`Enabling a parameter lets viewers interact with it. Locking one lets you pass it a value from your app while hiding it from viewers.`}</Text>
+
+                {resourceParameters.map(parameter => (
+                  <div key={parameter.id} className="flex align-center">
```
oisincoveney, line 95: I'm noticing a lot of `div` and `h3` tags and other HTML tags - are these remnants from the old code? Maybe it would be good in another PR to adjust these to mantine while we're here

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
```diff
@@ -0,0 +1,203 @@
...
+                    <Icon
+                      name={getIconForParameter(parameter)}
+                      className="mr2"
+                      style={{ color: color("text-light") }}
```
oisincoveney, line 99: same with colors and things - we want to move towards the mantine color system and remove as many `color()` calls as possible

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/AppearanceSettings.tsx
```diff
@@ -0,0 +1,245 @@
...
+                key="doc"
+                href={docsUrl}
+              >{t`documentation`}</ExternalLink>
+            )} for more.`}</Text>
+          </StaticEmbedSetupPaneSettingsContentSection>
+          <StaticEmbedSetupPaneSettingsContentSection
+            title={t`Play with the options here`}
+            mt="2rem"
+          >
+            <Stack spacing="1rem">
```
oisincoveney, line 95: I'm usually really biased to splitting these children into subcomponents - i.e. `FontControl` or `BackgroundControl` or things like that. I think it would make the structure a lot easier to look at and help us test/separate concerns more easily. What do you think?

frontend/src/metabase/public/components/EmbedModal/SelectEmbedTypePane/SelectEmbedTypePane.tsx
```diff
@@ -137,9 +139,11 @@ function SharingPane({
             ) : (
               <Text>
                 {t`Public embeds and links are disabled.`}{" "}
-                <Link to="/admin/settings/public-sharing">
-                  <Anchor data-testid="sharing-pane-settings-link">{t`Settings`}</Anchor>
-                </Link>
+                <Anchor
```
oisincoveney, line 142: Nice!

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.unit.spec.tsx
```diff
@@ -0,0 +1,189 @@
...
+});
+
+function setup({
+  embedType = null,
+  resource = {} as EmbedResource,
+  resourceType = "dashboard",
+  resourceParameters = [],
+  getPublicUrl,
+}: Partial<EmbedModalContentProps> = {}) {
+  const mocks = {
```
oisincoveney, line 154: Is this `mocks` objects necessary? I would prefer these to just be their own variables and then, in our tests, we can just destructure the mock functions that we need instead of dealing with the whole `mocks` object the whole time

frontend/src/metabase/public/components/EmbedModal/EmbedModal.types.ts
oisincoveney, line 1: So, so, so happy you fixed up these types 😄 maybe we can just name this file `types.ts` since it's already in the `EmbedModal` folder, I don't think we need that redundancy

frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+    embedType: EmbedType;
+    setEmbedType: (type: EmbedType) => void;
+  }) => JSX.Element;
+}
+
+export const EmbedModal = ({ children, isOpen, onClose }: EmbedModalProps) => {
+  const [embedType, setEmbedType] = useState<EmbedType>(null);
+  const applicationName = useSelector(getApplicationName);
+
+  const isEmbeddingStage = embedType === "application";
```
oisincoveney, line 30: In the Milestone 4 PR I'm thinking of a better way to name these booleans because I want to make sure things aren't too confusing. I think putting `isEmbeddingStage` will be confusing as well since everything is named `Embed<something>` so it's not immediately obvious. Leave this as it is but if you have any ideas for how we can improve the naming here I'm all ears

frontend/src/metabase/public/components/EmbedModal/EmbedModal.styled.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from "@emotion/styled";
+import { color } from "metabase/lib/colors";
+import { ModalHeader } from "metabase/components/ModalContent";
+
+export const EmbedModalHeader = styled(ModalHeader)`
+  padding: 1.5rem 2rem;
+
+  border-bottom: 1px solid ${color("border")};
```
oisincoveney, line 8: let's use the Mantine colors here again

frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
oisincoveney: I think ideally we should do that, but `Dashboard` and `Card` both have `public_uuid` typed as `string | null` so I'm not sure how much we can do before typescript complains. I think we would need to check in a parent component if we have to do a null check

e2e/test/scenarios/embedding/embedding-questions.cy.spec.js
```diff
@@ -198,6 +223,12 @@ describe("scenarios > embedding > questions ", () => {
 
     openStaticEmbeddingModal();
 
+    modal().within(() => {
```
oisincoveney, line 226: Maybe let's throw this into `e2e-embedding-helpers` or a function at the bottom of this file since it's being repeated a bunch

oisincoveney (dismissed):
There's a few things that could be improved here, but I'll need to take another pass at this in the future since the PR is so big.

I think in the future it'll be easier to do something like this in pieces because the merge conflict between master, Milestone 4, and Milestone 5 is going to be _painful_

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -0,0 +1,234 @@
...
+      font: null,
+      theme: null,
+      bordered: true,
+      titled: true,
+    },
+  );
+
+  const handleSave = async () => {
+    try {
+      if (embedType === "application") {
```
WiNloSt, line 85: I don't think we will have another `embedType` than `"application"` in this function because we only pass `handleSave` after we have checked that the `embedType` couldn't be null. And `embedType` can only either be `null` or `"application"`.

I suspect the else block was from before we passed `onCreatePublicLink` directly to `SelectEmbedTypePane`, this really emphasizes my other comment about splitting the component for each modal, so it's easier to reason about and handle their different logics.

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -0,0 +1,234 @@
...
+        if (!resource.enable_embedding) {
+          await onUpdateEnableEmbedding(true);
+        }
+        await onUpdateEmbeddingParams(embeddingParams);
+      } else {
+        if (!resource.public_uuid) {
+          await onCreatePublicLink();
+        }
+      }
+    } catch (e) {
```
WiNloSt, line 95: In what scenarios, do we need a `catch` block?

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -0,0 +1,234 @@
...
+      console.error(e);
+      throw e;
+    }
+  };
+
+  const handleUnpublish = async () => {
+    try {
+      await onUpdateEnableEmbedding(false);
+    } catch (e) {
```
WiNloSt, line 104: In what scenarios, do we need a `catch` block?

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -0,0 +1,234 @@
...
+function filterValidResourceParameters(
+  resourceParameters: EmbedResourceParameter[],
+  embeddingParams: EmbeddingParameters,
+) {
+  const validParameters = resourceParameters.map(parameter => parameter.slug);
+
+  return _.pick(embeddingParams, validParameters);
+}
+
+function getPreviewParameters(
```
WiNloSt, line 227: Since the UI says _Preview locked parameters_
image

I think we should name this `getPreviewLockedParameters` (the same as the copy) or just `getLockedParameters` (the same as the implementation).

Naming this just `getPreviewParameters` kind of confused me at first because I don't remember that the only previewable parameter is the locked one.

This comment applies to other variables that are named `*previewParameters*`

WiNloSt (commented):
review part 2/x

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -0,0 +1,234 @@
...
+    onCreatePublicLink,
+    onDeletePublicLink,
+    getPublicUrl,
+  } = props;
+
+  const [pane, setPane] = useState<ActivePreviewPane>("code");
+
+  const siteUrl = useSelector(state => getSetting(state, "site-url"));
+  const secretKey =
+    useSelector(state => getSetting(state, "embedding-secret-key")) ||
```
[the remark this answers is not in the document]
WiNloSt: Are what you showed in these 2 videos possible in this PR? Or are these just to demonstrate if something went wrong you'd rather have shown the fallback key instead? (I've tested that there's it wasn't possible even for a new instance since you need to enable the embedding first before you can open the embed modal.)

If it's the latter, I'd still think using `checkNotNull` is more explicit and that tells the state of the codebase more correctly. As the code comment has suggested, this should never happen anyway.

Is there any reason why you don't want to use `checkNotNull` more often?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.styled.tsx
```diff
@@ -0,0 +1,89 @@
...
+}): JSX.Element => {
+  return (
+    <ContentWrapper>
+      <SettingsAsideBlock>{settingsSlot}</SettingsAsideBlock>
+      <PreviewAreaBlock>{previewSlot}</PreviewAreaBlock>
+    </ContentWrapper>
+  );
+};
+
+export const CODE_PREVIEW_CONTROL_OPTIONS = [
```
WiNloSt, line 53: This doesn't feel like it belongs in this file as it doesn't have anything to do with styling. Am I correct?

WiNloSt (commented):
Review part 3/x. I've reviewed all the code, but haven't reviewed any tests yet. I'll do that later.

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
WiNloSt: If possible yeah, since we're not doing anything with the caught error here.

e2e/test/scenarios/embedding/embedding-downloads-questions.cy.spec.js
```diff
@@ -102,13 +102,15 @@ describeEE("scenarios > embedding > questions > downloads", () => {
...
 
         cy.log("Disable downloads");
         cy.findByLabelText("Enable users to download data from this embed?")
-          .should("be.checked")
-          .click()
-          .should("not.be.checked");
+          .as("allow-download-toggle")
+          .should("be.checked");
+        cy.findByText("Enable users to download data from this embed?").click();
+        cy.get("@allow-download-toggle").should("not.be.checked");
```
WiNloSt, line 113: Can we chain it like before? e.g.
```ts
          .as("allow-download-toggle")
          .should("be.checked")
          .click()
          .should("not.be.checked")
```

So the assertions look more concise.

e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js
```diff
@@ -154,27 +154,31 @@ describe("scenarios > embedding > smoke tests", { tags: "@OSS" }, () => {
...
-            cy.findByRole("heading", { name: "Appearance" });
-            cy.findByRole("heading", { name: "Font" }).should("not.exist");
-            cy.findByRole("heading", { name: "Download data" }).should(
-              "not.exist",
+          modal().within(() => {
+            cy.findByRole("tab", { name: "Appearance" }).click();
+
+            cy.findByText("Background");
+            cy.findByText("Dashboard title");
+            cy.findByText("Border");
```
WiNloSt, line 162: We discourage implicit assertion. Because it's uncertain whether the author wants to assert or wants to perform actions on the element, but forgot to add assertions.

Please use `.should("exist")` for these assertions.

e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js
```diff
@@ -154,27 +154,31 @@ describe("scenarios > embedding > smoke tests", { tags: "@OSS" }, () => {
...
-              /This (question|dashboard) doesn't have any parameters to configure yet./,
+              `This ${object} doesn't have any parameters to configure yet.`,
             );
-            cy.findByText("Parameters");
-          });
 
-          cy.findByTestId("embedding-preview").within(() => {
             cy.findByText(
-              /You will need to publish this (question|dashboard) before you can embed it in another application./,
+              `You will need to publish this ${object} before you can embed it in another application.`,
```
WiNloSt, line 177: 👍 For not using regex which is more prone to error.

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.unit.spec.tsx
WiNloSt, line 1: This file really emphasizes why we should split each modal into its own files, since this test only exercises the static embed modal. Not the first page of the embed modal

I know that you mentioned you'll do that in the next PR, but this is something I'd like to add.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
```diff
@@ -0,0 +1,201 @@
...
+              displayOptions={displayOptions}
+            />
+          ) : null}
+        </>
+      }
+    />
+  );
+};
+
+const getIconForParameter = (parameter: EmbedResourceParameter): IconName => {
```
WiNloSt, line 191: 👍 This is much easier to read.

WiNloSt (commented):
I've reviewed all the tests + addressed changes. I've added more minor feedback, but overall it's in a mergeable state imo.

Thanks for the effort this is huge 👍

WiNloSt (approved):

PR #36189, opened 2023-11-29
Fix settings jwt

shaunlebron:
Adding @npfitz’s suggested type fixes for https://github.com/metabase/metabase/pull/35969 (from a slack DM)

Also fixing a unit test ordering bug where I was clicking the checkbox after save (but it still worked somehow).


enterprise/frontend/src/metabase-enterprise/auth/components/SettingsJWTForm/SettingsJWTForm.tsx
```diff
@@ -16,24 +16,19 @@ import { Flex, Stack } from "metabase/ui";
...
-  // Similar to SettingElement from "metabase/admin/settings/types" but with required key
-  key: string;
-  display_name?: string;
-  description?: string;
+type JWTFormSettingElement = Omit<SettingElement, "key"> & {
+  key: string; // ensuring key is required
   is_env_setting?: boolean;
   env_name?: string;
-  placeholder?: string;
   default?: any;
```
npfitz, line 27: Should we set default to string here? since all the defaults in this form should be strings? Generally speaking we shouldn't use `any` as a type

npfitz (changes requested):

enterprise/frontend/src/metabase-enterprise/auth/components/SettingsJWTForm/SettingsJWTForm.tsx
shaunlebron: Fixed here: 621ed0484bd8772f64cd16b350049f684405a75e

Just empirically based on values returned by the getSectionsWithPlugins selector, not all values are strings, and can be other types based on the hard-coded elements I pasted into the unit test.

shaunlebron: There were more errors that popped up, so I just set it to string as suggested, and made the unit tests work with it.

enterprise/frontend/src/metabase-enterprise/auth/components/SettingsJWTForm/SettingsJWTForm.unit.spec.tsx
```diff
@@ -22,7 +22,7 @@ const elements = [
     is_env_setting: false,
     env_name: "MB_JWT_ENABLED",
     description: "Is JWT authentication configured and enabled?",
-    default: false,
+    // default: false,
```
npfitz, line 25: We should probably remove these commented values

npfitz (approved):
Looks good! Lets just get rid of the comments and we're all set

enterprise/frontend/src/metabase-enterprise/auth/components/SettingsJWTForm/SettingsJWTForm.unit.spec.tsx
shaunlebron: fbf0f6a408be7b7780e1b583da7c51f347d8a4f4

PR #36191, opened 2023-11-29
Housekeeping: remove default exports from dashboard/*

oleggromov:
### Description
I was working on fixing a bug in Dashboard and noticed lots of `export default`s and deprecation notices. Thought that it would make sense to clean them up as I am passing by.

Another thing I am removing are re-exports from `index.(js|ts)`, which add an unnecessary level of nesting _and_ are [bad for build performance at scale](https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/). 

Screenshot 2023-11-29 at 11 52 07


oleggromov:
As per slack discussion, removed the `Styled` suffixes. Also addressed your comment, @uladzimirdev!

alxnddr (approved):

PR #36203, opened 2023-11-29
Make with-expression-name set the display-name of the clause

### LINKED ISSUE #36167: withExpressionName does not affect name/displayName of an ExpressionClause
Given a [simple question with custom expression aggregation](https://github.com/metabase/metabase/assets/6830683/9e90d98b-37de-4c4b-a4dc-3a720cf7669d), this is how the aggregation clause looks like:

(notice `:display-name` and `:name` in the aggregation)
image

Calling `Lib.withExpressionName` on that clause won't affect `:display-name` or `:name` - ~instead~ it will only add `:lib/expression-name`:

image

`Lib.displayName(query, clause)` won't return the name updated with `Lib.withExpressionName`.

metamben:
Fixes #36167.

This is needed to be able to set/change the display name of custom aggregation columns too.


src/metabase/lib/expression.cljc
```diff
@@ -353,10 +353,12 @@
   (-> an-expression-clause lib.options/options :lib/expression-name))
 
 (mu/defn with-expression-name :- ::lib.schema.expression/expression
-  "Return a new expression clause like `an-expression-clause` but with name `new-name`."
+  "Return a new expression clause like `an-expression-clause` but with name `new-name`.
+  This also sets the :display-name property to support named aggregation expressions (custom aggregation columns)."
```
kamilmielnik, line 357: I think this is incorrect - please see [this comment](https://metaboat.slack.com/archives/C04CYTEL9N2/p1701257788565959?thread_ts=1701083796.853609&cid=C04CYTEL9N2).

----

This is how it's expected to work (pseudo-js code):

```tsx
const query = createQuery({
  query: {
    database: SAMPLE_DB_ID,
    type: "query",
    query: {
      aggregation: [["count"]],
    },
  },
});
const aggregations = Lib.aggregations(query, 0);

const aggregation = aggregations[0];
expect(Lib.displayName(query, aggregation)).toBe("Count");
expect(Lib.expressionName(aggregation)).toBe(null); // or empty string

const namedAggregation = Lib.withExpressionName(aggregation, "Hello world");
expect(Lib.displayName(query, namedAggregation)).toBe("Hello world");
expect(Lib.expressionName(namedAggregation)).toBe("Hello world");

const unnamedAggregation = Lib.withExpressionName(namedAggregation, "");
expect(Lib.displayName(query, unnamedAggregation)).toBe("Count");
expect(Lib.expressionName(unnamedAggregation)).toBe(""); // or null
```


----

To validate correctness please try to incorporate these changes into `35947-legacy-expression-clause-filters-and-aggregations` branch - all unit tests and e2e tests should pass. If they don't it means something is wrong with expression names.

----

Example scenario to validate correctness:

1. `git checkout master`
2. Create & save question like this: [screenshot](https://github.com/metabase/metabase/assets/6830683/0f153cac-9879-4f17-bf78-19d966f66b33)
3. `git checkout `35947-legacy-expression-clause-filters-and-aggregations` - branch with custom expressions ported to MLv2
4. `git merge 36167-with-expression-name-to-set-display-name` (incorporate this fix to frontend changes)
5. Refresh browser tab with created question in GUI editor
6. Click green "Hello world" to edit aggregation clause

Expected result: Custom expression editor is opened ([screenshot](https://github.com/metabase/metabase/assets/6830683/2e0cd237-2500-44bd-9e6a-30dc1c0bbb6b))
Actual result: Aggregations dropdown is opened ([screenshot](https://github.com/metabase/metabase/assets/6830683/58af69f6-9776-4755-aaae-d8461b841744)). This happens because `Lib.expressionName` returns `null`, and frontend can't tell whether expression name is custom or built-in. `Lib.expressionName` should return `"Hello world"` in this case.

kamilmielnik (changes requested):

src/metabase/lib/expression.cljc
ranquild: @kamilmielnik please use `displayInfo` instead of displayName

metamben: @kamilmielnik Ugh, I see you want to use `Lib.expressionName` to be able to tell if `Lib.withExpressionName` has been called. That's a bit difficult, here's why.

As I mentioned before, `Lib.withExpressionName` had been meant to set the key of expressions in the `expressions` clause of the MLv2 query. Among others, the property storing that key is used to tell what kind of reference has to be created for a given column. If I just added that property to an aggregation, we would start generating references by name, although aggregations are referenced by position.

I can see two ways to provide what you need:
1. make (just!) the JS version of `Lib.expressionName` return the custom display name, or
2. introduce `Lib.customizedDisplayName` (or s.g. similar) that would return custom display names and (standard) expression names.

I guess 1. is easier for you now, but it can cause confusion because functions of the same name would diverge in JS and CLJ. Alternative 2. needs more work, but I think it has lower potential to break things and to cause confusion.

metamben: After a discussion with @ranquild we decided to implement the following:
1. `withExpressionName` should work both for setting the name of expressions and for naming aggregations.
2. `displayInfo` should return a flag `isNamed` that's true if the thing has been given a name.

src/metabase/lib/util.cljc
```diff
@@ -41,10 +41,10 @@
   "Returns true if this is a clause."
   [clause]
   (and (vector? clause)
-       (> (count clause) 1)
        (keyword? (first clause))
-       (map? (second clause))
-       (contains? (second clause) :lib/uuid)))
+       (let [opts (get clause 1)]
+         (and (map? opts)
+              (contains? opts :lib/uuid)))))
```
metamben, line 47: This is an unrelated change that I hope makes things a bit more explicit.

frontend/src/metabase-lib/expression.ts
```diff
@@ -21,10 +21,6 @@ export function expression(
   return ML.expression(query, stageIndex, expressionName, clause);
 }
 
-export function expressionName(clause: ExpressionClause): string {
-  return ML.expression_name(clause);
-}
```
metamben, line 26: `display-info` should be used instead.

frontend/src/metabase-lib/types.ts
```diff
@@ -131,7 +131,9 @@ export type MetricDisplayInfo = {
 export type ClauseDisplayInfo = Pick<
   ColumnDisplayInfo,
   "name" | "displayName" | "longDisplayName" | "table"
->;
+> & {
+  isNamed?: boolean;
+};
```
metamben, line 136: Only expression, aggregation, and filter clauses are planned to have this property, but I didn't feel like adding it one by one or creating a new type. I can do that if that's important though.

kamilmielnik (dismissed):
Works as expected :+1:

camsaul (approved):
backend code LGTM

kamilmielnik:
FYI @metamben I merged latest `master` in

ranquild:
@kamilmielnik could you take a look? Does it solve your case? If yes please approve

kamilmielnik (approved):

Issue #34828, opened 2023-10-19
PUT and POST and endpoint should return data just like calling GET


qnkhuat:
On UI we see a lot of places where after we `POST/PUT` a card/dashboard, we have to do another `GET` to get the item.

This is unnecessary and reduces performance, we need to standardize and make all `POST/PUT` to return the result as if you call a `GET`.

This way FE doesn't have to re-fetch after every create or update action.

```[tasklist]
### Tasks
- [x] Replace GET /api/dashboard with PUT/POST
```

calherries:
This already works for the most part. You can see it working for example when archiving a dashboard from a collection page. The FE is just doing unnecessary GET calls in some cases, like when it saves a dashboard from the dashboard editing screen.

But we should (1) enforce GET/POST/PUT equivalence on the BE with tests, and then (2) Remove the unnecessary GET calls from the FE.

markbastian:
https://github.com/metabase/metabase/pull/36205 resolves the BE aspects of this and removes the extraneous GET for dashboard. @EmmadUsmani is going to do a follow on to take care of `Remove redundant GET /api/card request after creating a question`.

markbastian:
FE work for this has been split into https://github.com/metabase/metabase/issues/36326.

PR #36205, opened 2023-11-29
Update dashboard and card api POST and PUT endpoints to return as GET

markbastian:
In both the `metabase.api.dashboard` and `metabase.api.card` nses the primary change was to add a `hydrate-dashboard|card-details` function that is shared across the GET, POST, and PUT operations so that all 3 return an object of the same shape, conforming to the GET response. This removes the need for a POST|PUT + GET call on the FE. Note the in all cases, the returned value was nearly the same as the GET response, so it really was returning _almost_ the same thing so this does effectively halve the load required for these operations.

Unit tests were added to check that the returned objects from POST|PUT are the same as from a GET.

### DIFF
--- a/src/metabase/api/card.clj
+++ b/src/metabase/api/card.clj
@@ -53,6 +53,7 @@
    [metabase.util.log :as log]
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
+   [steffan-westcott.clj-otel.api.trace.span :as span]
    [toucan2.core :as t2])
   (:import
    (java.io File)))
@@ -148,21 +149,25 @@
                      card)))
             cards))))
 
-(defn hydrate-for-frontend
+(defn hydrate-card-details
   "Adds additional information to a `Card` selected with toucan that is needed by the frontend. This should be the same information
   returned by all API endpoints where the card entity is cached (i.e. GET, PUT, POST) since the frontend replaces the Card
   it currently has with returned one -- See #4283"
-  [card]
-  (-> card
-      (t2/hydrate :creator
-                  :dashboard_count
-                  :can_write
-                  :average_query_time
-                  :last_query_start
-                  [:collection :is_personal]
-                  [:moderation_reviews :moderator_details])
-      (cond-> ; card
-        (:dataset card) (t2/hydrate :persisted))))
+  [{card-id :id :as card}]
+  (span/with-span!
+    {:name       "hydrate-card-details"
+     :attributes {:card/id card-id}}
+    (-> card
+        (t2/hydrate :creator
+                    :dashboard_count
+                    :can_write
+                    :average_query_time
+                    :last_query_start
+                    :parameter_usage_count
+                    [:collection :is_personal]
+                    [:moderation_reviews :moderator_details])
+        (cond->                                             ; card
+          (:dataset card) (t2/hydrate :persisted)))))
 
 (api/defendpoint GET "/:id"
   "Get `Card` with ID."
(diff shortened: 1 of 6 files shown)


test/metabase/api/card_test.clj
```diff
@@ -674,19 +674,26 @@
...
+                                                                                         :name         card-tag-name,
+                                                                                         :type         :card}}}}}
+                {card-id :id
+                 :as     created} (mt/user-http-request :rasta :post 200 "card"
+                                              (merge
+                                                (mt/with-temp-defaults :model/Card)
+                                                dataset-query))
+                retrieved     (mt/user-http-request :rasta :get 200 (str "card/" card-id))]
+            (is (pos-int? card-id))
+            (testing "A POST returns the newly created object, so no follow-on GET is required"
```
adam-james-v, line 693: This PR addresses a bug right? I've seen #ISSUENUM put in testing strings as a reference before, if you want to add that here.

adam-james-v (approved):
Backend changes look nice. I like that hydration is pulled into one fn. Makes it easy to reuse in all the right spots.

test/metabase/api/card_test.clj
markbastian: Hmm, more of a performance improvement, but i like the idea. Will update.

src/metabase/api/card.clj
```diff
@@ -166,15 +185,7 @@
...
-                             :parameter_usage_count
-                             :can_write
-                             :average_query_time
-                             :last_query_start
-                             [:collection :is_personal]
-                             [:moderation_reviews :moderator_details])
-                 collection.root/hydrate-root-collection
+                 hydrate-card-details
                  (cond-> ;; card
                    (:dataset raw-card) (t2/hydrate :persisted))
```
calherries, line 190: Just spotted you don't need this anymore. You're doing it in `hydrate-card-details`.
```suggestion
```

src/metabase/api/card.clj
```diff
@@ -166,15 +185,7 @@
...
                  (last-edit/with-last-edit-info :card))]
```
calherries, line 191: Why isn't `(last-edit/with-last-edit-info :card)` part of `hydrate-card-details`?

calherries (changes requested):
Nice! I was about to ping you to try something like this because I just merged [Move functions for creating and updating cards from api to models namespace](https://github.com/metabase/metabase/pull/36123). Unfortunately it creates conflicts with this in `metabase.api.card`. In that PR I consolidated a lot of the stuff in your `hydrate-card-details` in a similar function called `hydrate-for-frontend`, but deliberately left out some keys because I didn't want to make too many changes outside the refactoring I was doing. It should be pretty straightfoward to resovle the conflict though

src/metabase/api/card.clj
```diff
@@ -129,6 +130,24 @@
...
+        (t2/hydrate :creator
+                    :dashboard_count
+                    :parameter_usage_count
+                    :can_write
+                    :average_query_time
+                    :last_query_start
+                    [:collection :is_personal]
+                    [:moderation_reviews :moderator_details])
+        (cond-> dataset (t2/hydrate :persisted))
+        collection.root/hydrate-root-collection)))
```
calherries, line 149: Are you sure the root collection needs to be hydrated on all endpoints? I would expect that all collections are normalized on the frontend and cached separately from the card stuff, including the root collection.

src/metabase/api/card.clj
markbastian: I was wondering about this as well. It's the same situation in api.dashboard in terms of when to update the last edit info. It does collect the logic in one place, but it feels like it isn't part of the hydration logic. WDYT?

src/metabase/api/card.clj
markbastian: Agreed. Done.

markbastian: Per our discussion, let's go with removing. This means we'll not have _identical_ results as POST|PUT + GET, but the assumption is that the `:collection` will be present already on the FE in those cases.

calherries (approved):

markbastian:
Fixes https://github.com/metabase/metabase/issues/34828. Note that FE portions of that ticket were moved to https://github.com/metabase/metabase/issues/36326.

PR #36206, opened 2023-11-29
 Fix card filter connected to a field not causing an update

### LINKED ISSUE #35503: Connect a card to a filter in a dashboard doesn't update data
### Describe the bug

When we connect parameters (SET_PARAMETER_MAPPING / setParameterMapping) we don't trigger the dashboard update. 

Thread: https://metaboat.slack.com/archives/C01LQQ2UW03/p1699416618327629

### To Reproduce

1. Create a question based on Sample Database / Orders
2. Add it to a new Dashboard
3. Click Edit Dashboard
4. Add Time / Relative filter 
5. Save and Refresh page
6. Click Edit and connect the filter to the Order - Created At field
7. Save
8. Filter diesn't apply without page refresh 

### Expected behavior

Dashboard should be updated based on the filter without page refresh

### Logs

_No response_

```JSON
master
```

### Severity

P2

### Additional context

_No response_

oleggromov:
Fixes https://github.com/metabase/metabase/issues/35503

### Description

The problem was that for some reason in `Dashboard.componentDidUpdate`, the method that decides whether to fetch new data, only parameter _values_ where checked — but not parameters themselves.

I have fixed that, removing garbage code along the way, and added "unit tests" for the `Dashboard` component. Net change (without the test code) is `-46` sloc, which I think is worth it.

### Test approach evolution
1. I have tried to implement an e2e test suite, which is pure madness in my opinion — I would've to recreate the entire page in a browser and click through 10 times to cover a tiny conditional way down in the function call tree.
2. Then I tried to extract a real unit from a react component (which hardly is a "unit" itself), but realised that I would still need to invoke the component itself to make sure the integration of the component and this unit with conditionals is covered too.
3. Finally I have found a happy middle ground, which I am quite satisfied with, but the setup part is ugly anyways. 

PS Certainly the tests we produce with react testing library aren't _unit_ tests because they require a rich environment, so much mocking, and are **slow**. 

### How to verify
Repeat after what I do in the screencast.

#### Before it didn't work 
https://github.com/metabase/metabase/assets/2196347/2a825b5e-8840-471d-8213-60d81321f561

#### Now it works 
https://github.com/metabase/metabase/assets/2196347/a8aa503f-9fdf-49e4-9cd8-fb8f7ec02505

### Checklist

### DIFF
--- a/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
+++ b/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
@@ -42,10 +42,6 @@ class DashboardInner extends Component {
     isSharing: false,
   };
 
-  constructor(props) {
-    super(props);
-  }
-
   static getDerivedStateFromProps({ parameters }, { parametersListLength }) {
     const visibleParameters = getVisibleParameters(parameters);
     return visibleParameters.length !== parametersListLength
@@ -54,14 +50,14 @@ class DashboardInner extends Component {
   }
 
   async componentDidMount() {
-    await this.loadDashboard(this.props.dashboardId);
+    await this.loadDashboard();
   }
 
   async componentDidUpdate(prevProps) {
     updateParametersWidgetStickiness(this);
 
     if (prevProps.dashboardId !== this.props.dashboardId) {
-      await this.loadDashboard(this.props.dashboardId);
+      await this.loadDashboard();
       return;
     }
 
@@ -84,29 +80,17 @@ class DashboardInner extends Component {
     this.props.cancelFetchDashboardCardData();
   }
 
-  async loadDashboard(dashboardId) {
-    const {
-      editingOnLoad,
-      addCardOnLoad,
-      addCardToDashboard,
-      fetchDashboard,
-      initialize,
-      loadDashboardParams,
-      location,
-      setErrorPage,
-      isNavigatingBackToDashboard,
-    } = this.props;
+  async loadDashboard() {
+    const p = this.props;
+    p.initialize({ clearCache: !p.isNavigatingBackToDashboard });
+    p.loadDashboardParams();
 
-    initialize({ clearCache: !isNavigatingBackToDashboard });
-
-    loadDashboardParams();
-
-    const result = await fetchDashboard({
-      dashId: dashboardId,
-      queryParams: location.query,
+    const result = await p.fetchDashboard({
+      dashId: p.dashboardId,
+      queryParams: p.location.query,
       options: {
-        clearCache: !isNavigatingBackToDashboard,
-        preserveParameters: isNavigatingBackToDashboard,
+        clearCache: !p.isNavigatingBackToDashboard,
+        preserveParameters: p.isNavigatingBackToDashboard,
       },
     });
 
@@ -119,16 +103,16 @@ class DashboardInner extends Component {
       if (editingOnLoad) {
         this.setEditing(this.props.dashboard);
       }
-      if (addCardOnLoad != null) {
-        addCardToDashboard({
-          dashId: dashboardId,
-          cardId: addCardOnLoad,
-          tabId: this.props.dashboard.tabs[0]?.id ?? null,
+      if (p.addCardOnLoad != null) {
+        p.addCardToDashboard({
+          dashId: p.dashboardId,
+          cardId: p.addCardOnLoad,
+          tabId: p.dashboard.tabs[0]?.id ?? null,
         });
       }
     } catch (error) {
       if (error.status === 404) {
-        setErrorPage({ ...error, context: "dashboard" });
+        p.setErrorPage({ ...error, context: "dashboard" });
       } else {
         console.error(error);
         this.setState({ error });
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -73,7 +69,7 @@ class DashboardInner extends Component {
 
     if (
       !_.isEqual(prevProps.parameterValues, this.props.parameterValues) ||
-      (!prevProps.dashboard && this.props.dashboard)
```
oleggromov, line 76: We do initially render Dashboard with `dashboard={null}`. However, this conditional never works as it's superseded by the first one in this method: `prevProps.dashboardId !== this.props.dashboardId`.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -73,7 +69,7 @@ class DashboardInner extends Component {
...
+      !_.isEqual(prevProps.parameters, this.props.parameters)
```
oleggromov, line 72: This is the actual bug fix.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -156,159 +140,139 @@ class DashboardInner extends Component {
...
     this.props.toggleSidebar(SIDEBAR_NAME.addQuestion);
   };
 
-  renderContent = () => {
-    const { dashboard, selectedTabId, isNightMode, isFullscreen } = this.props;
-
-    const canWrite = dashboard?.can_write ?? false;
-
-    const dashboardHasCards = dashboard?.dashcards.length > 0 ?? false;
+  shouldRenderAsNightMode() {
```
oleggromov, line 147: I have extracted this as a method as it was used in two different methods.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -156,159 +140,139 @@ class DashboardInner extends Component {
...
+    return this.props.isNightMode && this.props.isFullscreen;
+  }
 
+  renderContent = () => {
+    const p = this.props;
```
oleggromov, line 152: Everything that's happening here is removing destructured props as they are essentially slag.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -156,159 +140,139 @@ class DashboardInner extends Component {
...
-      isHeaderVisible,
-      isAutoApplyFilters,
-    } = this.props;
-
-    const { error, isParametersWidgetSticky } = this.state;
-
-    const shouldRenderAsNightMode = isNightMode && isFullscreen;
-
-    const visibleParameters = getVisibleParameters(parameters);
+    const p = this.props;
```
oleggromov, line 200: Same here: instead of wasting lines on avoiding writing `this.props`, I am putting it into a one-char variable that's almost invisible to the eye but _and_ removes this unnecessary cruft.

frontend/src/metabase/dashboard/containers/DashboardApp/DashboardApp.unit.spec.tsx
```diff
@@ -34,28 +34,19 @@ import {
...
-});
-
-const TEST_COLLECTION_ITEM = createMockCollectionItem({
-  collection: TEST_COLLECTION,
-  model: "dataset",
-});
-
-const TEST_CARD = createMockCard();
-
-const TEST_TABLE = createMockTable();
```
oleggromov, line 50: Same cruft as everywhere else. We don't need "constants" when they're not exported or at least used multiple times, and absolutely certainly we don't need them at the top of the file — it's ugly and useless albeit quite common.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+  mockFetchDashboardCardData: jest.Mock;
+  mockFetchDashboardCardMetadata: jest.Mock;
+};
+
+async function setup(props: TestedProps): Promise<SetupReturn> {
+  const mockDashboard = createMockDashboard({ id: 10 }); // value is irrelevant
+  const mockLoadDashboardParams = jest.fn();
+  const mockFetchDashboard = jest.fn(() => mockDashboard);
+  const mockFetchDashboardCardData = jest.fn();
+  const mockFetchDashboardCardMetadata = jest.fn();
```
ranquild, line 30: nit: We usually name callbacks as `onFetch...` and return them as it is

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+    rerender: (props: TestedProps) => rerender(<TestComponent {...props} />),
+    mockLoadDashboardParams,
+    mockFetchDashboard,
+    mockFetchDashboardCardData,
+    mockFetchDashboardCardMetadata,
+  };
+}
+
+describe("Dashboard data fetching", () => {
+  const DEFAULT_PROPS: TestedProps = {
```
ranquild, line 111: We usually use these props in `setup` function so we don't need to deal with spreading

ranquild (approved):
Good overall. Somewhat inconsistent with other integration tests (naming etc).

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
oleggromov: I was wrong, so reverting this. Won't be committing after working hours anymore.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
oleggromov: That's actually a very good catch, thanks!

PR #36217, opened 2023-11-29
Add column-name and table-name to fk-filter drill's display-info

### LINKED ISSUE #36000: [MLv2] [BE] Add additional data to `displayInfo` for `fk-filter` drill
Add data to `displayInfo(query, stageIndex, drillThru)`:
- `tableName`
- `columnName`

Repro test https://github.com/metabase/metabase/blob/bb792e36475d26ca4082392ddf0907244bea7f08/frontend/src/metabase-lib/tests/drills-fk.unit.spec.ts#L50

metamben:
Fixes #36000.

### DIFF
--- a/src/metabase/lib/metadata/calculation.cljc
+++ b/src/metabase/lib/metadata/calculation.cljc
@@ -277,7 +277,7 @@
 
 (mr/register! ::display-info
               [:map
-               [:display-name :string]
+               [:display-name {:optional true} :string]
                [:long-display-name {:optional true} :string]
    ;; for things that have a Table, e.g. a Field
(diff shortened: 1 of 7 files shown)


src/metabase/lib/metadata/calculation.cljc
```diff
@@ -277,7 +277,7 @@
 
 (mr/register! ::display-info
               [:map
-               [:display-name :string]
+               [:display-name {:optional true} :string]
```
metamben, line 280: During the implementation I noticed that `display-info` fails in the BE for most of the drills because they don't provide any `:display-name` property.

ranquild (approved):

camsaul (approved):

PR #36220, opened 2023-11-30
Alert/Dashboard subscription: render question title as <a> tag

### LINKED ISSUE #12901: Outlook pulse/alert doesn't allow clicking question title to visit link
**Describe the bug**
Seems like (newer versions of?) Outlook doesn't support `<a>` around a `<table>`, which means that it's not possible to click the question title in a pulse/alert email to visit the question.

**Expected behavior**
Ability to click question title to visit question.

Metabase 1.34.3 and 0.35.4

qnkhuat:
fixes #12901

On some versions of Outlook clients, `<a>` around `<table>` is not supported, so you can't click on the title/table to get directed to the question URL.

The fix is to make the title clickable by wrapping it with `<a>` tag, we only fix the title here because it'll probably solve the problem for most ppl.

about layout, it's probably fine to change from `<span>` to `<a>` because the title has its own line anyway.
Visually, there is no difference.
Before:
Screenshot 2023-11-30 at 17 39 43
After:
Screenshot 2023-11-30 at 17 39 54

### DIFF
--- a/test/metabase/pulse/render_test.clj
+++ b/test/metabase/pulse/render_test.clj
@@ -198,3 +198,14 @@
             (is (= expected
                    (->> (get-in table [3 1])
                         (map #(peek (get (vec (get % 2)) tax-col))))))))))))
+
+(deftest title-should-be-an-a-tag-test
+  (testing "the title of the card should be an <a> tag so you can click on title using old outlook clients (#12901)"
+    (mt/with-temp [Card card {:name          "A Card"
+                              :dataset_query (mt/mbql-query venues {:limit 1})}]
+      (mt/with-temp-env-var-value [mb-site-url "https://mb.com"]
+        (let [rendered-card-content (:content (binding [render/*include-title* true]
+                                                (render/render-pulse-card :inline (pulse/defaulted-timezone card) card nil (qp/process-query (:dataset_query card))))
+                                              :content)]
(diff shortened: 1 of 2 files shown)


test/metabase/pulse/render_test.clj
```diff
@@ -198,3 +198,14 @@
...
                         (map #(peek (get (vec (get % 2)) tax-col))))))))))))
+
+(deftest title-should-be-an-a-tag-test
+  (testing "the title of the card should be an <a> tag so you can click on title using old outlook clients (#12901)"
+    (mt/with-temp [Card card {:name          "A Card"
+                              :dataset_query (mt/mbql-query venues {:limit 1})}]
+      (mt/with-temp-env-var-value [mb-site-url "https://mb.com"]
+        (let [rendered-card-content (:content (binding [render/*include-title* true]
+                                                (render/render-pulse-card :inline (pulse/defaulted-timezone card) card nil (qp/process-query (:dataset_query card))))
+                                              :content)]
```
calherries, line 209: You've got an extra `:content` there
```suggestion
        (let [rendered-card-content (:content (binding [render/*include-title* true]
                                                (render/render-pulse-card :inline (pulse/defaulted-timezone card) card nil (qp/process-query (:dataset_query card)))))]
```

calherries (approved):
Looks good otherwise. Agreed that it's fine that only the title is clickable

PR #36222, opened 2023-11-30
Install clojure before running clj-kondo

qnkhuat:
running `bin/kondo.sh` requires `clojure` but we didn't install it during the kondo job

job: https://github.com/metabase/metabase/actions/runs/7045318501/job/19174789904?pr=36220#step:5:8


nemanjaglumac (changes requested):
Prepare-backend composite action already includes clojure.
We should use that instead.

Btw, I'm wondering how has this even worked so far?
Why this job doesn't fail if it needs clojure that it cannot find?

nemanjaglumac (approved):

Issue #35763, opened 2023-11-15
Research to improve our caching system


luizarakaki:
**Context**
Caching is important for performance optimization in large instances and to reduce costs.
We have our caching system, but there are known problems ([described in this doc](https://www.notion.so/metabase/Cache-arc-WIP-cdfddc99fbea40339679c6adf9948e34?pvs=4))

We want to improve our caching system, but first, would be important to gather information to design a plan.
The deliverable of this issue should a document or maybe a PoC, we don't expect any production code.

### 1. Current state diagnosis
The % of cache hit on Stats is surprisingly low (less than 1%): https://stats.metabase.com/question/14400-cache-hit-percentage-on-stats
Investigate if our caching system is working. And if now, why not. Some ideas:
- Our homepage [dashboard](https://stats.metabase.com/dashboard/1887) has cache turned on for 1h. And it has ~0% cache hit ([question](https://stats.metabase.com/question/14401-cache-hit-on-the-homepage))
- Investigate if caching is working between users. User A queries question 1, User B should get the cache if executing the exact same query.
- Check if query execution is capturing cache hit accordingly.
- Check different data sources, if need to prioritize, use the [Database Tier List](https://www.notion.so/Database-Tiers-for-Feature-Development-fabba4d2bbfb4c729029f5e77d585f61) (Postgres, MySQL, BigQuery, Redshift, Snowflake, Mongo)

### 2. Explore memory caching
What if we add memory caching to Metabase. Would it improve our experience?
- Would it be faster?
- Would it work well if our current stack?
- Would we alleviate AppDB load?
- Is it fundamentally better than AppDB cache?

paoliniluis:
https://github.com/metabase/metabase/issues?q=is%3Aopen+is%3Aissue+label%3AQuerying%2FCache+label%3A%22Type%3ANew+Feature%22

darksciencebase:
part one (testing the current system) can be started on; findings to be added here, in the issue

luizarakaki:
Related
https://github.com/metabase/metabase/issues/35630
https://github.com/metabase/metabase/issues/16517

piranha:
Okay, let me change it and see what happens...

piranha:
This made situation a bit better, we regularly hit more than 10% cached queries in an hour. What bothers me though is that when I query db for "long" queries, there is not a single cached request.

```
select e.hash, now(), e.started_at, e.running_time, e.cache_hit, c.updated_at
from query_execution e
[...]

PR #36223, opened 2023-11-30
store query_execution.cache_hash

### LINKED ISSUE #16517: Record metadata, when queries are using cache
The `query_cache.query_hash` is not the same as `query.query_hash` or `query_execution.hash` since x.37.6 (#14388).
This means it's not possible to correlate if cache is being used for queries.

- Another column in `query_execution`, which tells if cached results was used - ex. boolean column `cached_results`, or something like `cache_hash` which would get the value of `query_cache.query_hash`
- Making it more clear in the log, when cached results are used.
- Some way of connecting `query_cache` to `query` or `query_execution`
- Having some way of viewing this information in Admin > Audit

1. Doing estimates of `query_execution.running_time`, which should be a lot faster than queries without cache.
2. Storing the log and checking all request lines, which has a line before the request, which indicates that cache is being used:
```
2021-06-09 15:14:54,293 INFO api.card :: Question's average execution duration is 373.0 ms; using 'magic' TTL of 10.4 hours 💾
2021-06-09 15:14:54,507 DEBUG middleware.log :: POST /api/card/114/query 202 [ASYNC: completed] 217.3 ms (24 DB calls) App DB connections: 0/10 Jetty threads: 3/50 (2 idle, 0 queued) (78 total active threads) Queries in flight: 0 (0 queued)
```

**Additional context**
Adding Regression, since it now makes it close to impossible to figure out cache-hit-rate.

sansolovyov:
This will allow us to analyze cache effectiveness better, see #35763

Resolves #16517

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/sandbox/query_processor/middleware/row_level_restrictions_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/sandbox/query_processor/middleware/row_level_restrictions_test.clj
@@ -731,7 +731,7 @@
         (testing "Run the query, should not be cached"
           (let [result (run-query)]
             (is (= nil
-                   (:cached result)))
+                   (:cached (:cache/details result))))
             (is (= [[10]]
                    (mt/rows result)))))
         (testing "Cache entry should be saved within 5 seconds"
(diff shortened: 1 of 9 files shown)


metamben (approved):
LGTM

Issue #25274, opened 2022-09-06
Tables render slowly


paoliniluis:
**Describe the bug**
There seems to be an issue when rendering tables in a dashboard

**Logs**
NA

**To Reproduce**
1) new->question->orders- >sum of total, sum of tax, sum of subtotal, sum of quantity, sum of discount by product category
2) visualize and save, add to a new dashboard
3) resize the table so you can see it all together
4) refresh the dashboard, see that it will first render the first row, and then all others

**Expected behavior**
Render all together, or smoother?

**Screenshots**
![slowtable](https://user-images.githubusercontent.com/1711649/188755880-9e15acb1-aef3-46ae-a0d7-306842198400.gif)

**Information about your Metabase Installation:**
Your browser and the version: Brave latest
Your operating system: Pop 22.04
Your databases: Sample H2
Metabase version: 1.44.2
Metabase hosting environment: Docker
Metabase internal database: H2

**Severity**
Cosmetic

**Additional context**
NA

paoliniluis:
Fixed by https://github.com/metabase/metabase/pull/36226

PR #36226, opened 2023-11-30
fix dashcard table visual lag

alxnddr:
### Description

There is a visual lag on the dashcard table, it initially renders 1 row and then after about ~200ms it renders others. This is due to debounce of the `ExplicitSize` hoc which do not provide element dimensions until there is no change in them for 200ms affects the initial rendering. By changing resizeMode to `debounceLeading` we can solve this problem.

### How to verify

- Create a dashboard with a table
- Reload the page and ensure it initially renders rows for all available space in the card 

#### Before

https://github.com/metabase/metabase/assets/14301985/408a428d-ec87-4157-8915-ed29df2bc140

#### After

https://github.com/metabase/metabase/assets/14301985/29d97595-5826-4648-a062-c3ee9a0a40d7

### Checklist

- [n/a] Tests have been added/updated to cover changes in this PR

### DIFF
--- a/e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
@@ -548,11 +548,7 @@ describe(
             .should("not.be.checked");
           filterWidget().findByText("Gadget").should("be.visible");
 
-          // Card height isn't updated because we're mocking clock, this is because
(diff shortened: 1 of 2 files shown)


kdoh (approved):
Quite a noticeable improvement. 🚢

kulyk (approved):

paoliniluis:
Fixes https://github.com/metabase/metabase/issues/25274

PR #36231, opened 2023-11-30
[docs] Update environment-variables.md

likeshumidity:
Added  MB_JDBC_DATA_WAREHOUSE_UNRETURNED_CONNECTION_TIMEOUT_SECONDS to env variable docs


docs/configuring-metabase/environment-variables.md
```diff
@@ -532,6 +532,16 @@ Change this to a higher value if you notice that regular usage consumes all or c
...
 
+### `MB_JDBC_DATA_WAREHOUSE_UNRETURNED_CONNECTION_TIMEOUT_SECONDS`
+
+Type: integer<br>
+Default: `1200`<br>
+Since: v47.4
+
+Kill connections if they are unreturned after this amount of time. In theory this should not be needed, because the Query Processor will kill active connections that time out, but occasionally in practice some connections are severed and undetected by Metabase.
+
+Defaults to the 1200 seconds (20 minutes) based on default production value for `metabase.query-processor.query-timeout-ms`.
```
paoliniluis, line 543: The default is for the query timeout, not for the connection. The connection can stay alive, while the query has finished and that's what this env var tries to avoid

likeshumidity: Making a commit with another set of edits.

- `metabase.query-processor.query-timeout-ms` is in ms (1200000 in this case)
- `MB_JDBC_DATA_WAREHOUSE_UNRETURNED_CONNECTION_TIMEOUT_SECONDS` is in seconds (1200 in this case)
- humans understandable (20 min in this case)

if there is a more elegant way to make this distinction (ms/sec/min), I am open to any edits.

likeshumidity: @paoliniluis agreed that `metabase.query-processor.query-timeout-ms` is for query timeout not connection, but the code defining `jdbc-data-warehouse-unreturned-connection-timeout-seconds` [here](https://github.com/metabase/metabase/blob/e512233f8cf55286eda216b3158b2dee4c8f5aa6/src/metabase/driver/sql_jdbc/connection.clj#L89C1-L100C1) uses the query-timeout-ms (divided by 1000) as the value if none is set via env var.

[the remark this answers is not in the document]
likeshumidity: addressed with edits

jeff-bruemmer (approved):

PR #35894, opened 2023-11-16
Add analytics namespace

qwef:
Closes https://github.com/metabase/metabase/issues/35700

### Description

Add analytics namespace and have filtering centered around the namespace instead of the previous way, which was more ad-hoc.

PR #36238, opened 2023-11-30
[Auditv2] Don't show broken move and archive buttons for analytics collections

### LINKED ISSUE #35700: Can see move and archive buttons for metabase analytics collection in the root collection
https://metaboat.slack.com/archives/C01LQQ2UW03/p1699525120236779

### Describe the bug

Metabase analytics is supposed to be a read-only collection. So we shouldn't allow users to move or archive it, although these actions do nothing, it's confusing that we can even see them.

image

After moving or archive it shows this error, but after refreshing the page, the collection is still there
image

### To Reproduce

1. Go to the root collection
2. Select **Metabase analytics** collection
3. See that you can either **Move** or **Archive** it

### Expected behavior

I shouldn't be able to select this collection with bulk actions in the first place since it's a read-only collection

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.14.1+1-LTS",
    "java.version": "11.0.14.1",
    "java.vm.version": "11.0.14.1+1-LTS",
    "os.name": "Mac OS X",
    "os.version": "14.0",
    "user.language": "en",
    "user.timezone": "Asia/Bangkok"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2023-11-12",
      "src_hash": "93a2193e3bb61e49153bcf2aa91b6ee2110b129c",
      "tag": "v1.48.0-SNAPSHOT",
      "hash": "d00c3f6"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

Confusing, because these actions does nothing.

### Additional context

_No response_

iethree:
Closes https://github.com/metabase/metabase/issues/35700 again? I don't think https://github.com/metabase/metabase/pull/35894 actually addressed this, which was caused by the UI not respecting `can_write` properties on collections in the collection list view.

### Description

Checks `can_write` for collections in the collection list view to determine whether we should show the move and archive buttons

Before | After
--- | ---
Screen Shot 2023-11-30 at 8 54 45 AM | Screen Shot 2023-11-30 at 8 54 21 AM

### How to verify

1. visit the "our analytics collection"
2. click on the ... menu next to "metabase analytics"
3. do the same for "custom reports" within metabase analytics
4. see that you don't get the option to move or archive

### Checklist

### DIFF
--- a/frontend/src/metabase/collections/utils.ts
+++ b/frontend/src/metabase/collections/utils.ts
@@ -129,6 +129,10 @@ export function isItemCollection(item: CollectionItem) {
   return item.model === "collection";
 }
 
+export function isReadOnlyCollection(collection: CollectionItem) {
+  return isItemCollection(collection) && !collection.can_write;
+}
+
 export function canPinItem(item: CollectionItem, collection: Collection) {
   return collection.can_write && item.setPinned != null;
 }
@@ -140,6 +144,7 @@ export function canPreviewItem(item: CollectionItem, collection: Collection) {
 export function canMoveItem(item: CollectionItem, collection: Collection) {
   return (
     collection.can_write &&
+    !isReadOnlyCollection(item) &&
     item.setCollection != null &&
     !(isItemCollection(item) && isRootPersonalCollection(item))
   );
@@ -148,6 +153,7 @@ export function canMoveItem(item: CollectionItem, collection: Collection) {
 export function canArchiveItem(item: CollectionItem, collection: Collection) {
   return (
     collection.can_write &&
+    !isReadOnlyCollection(item) &&
     !(isItemCollection(item) && isRootPersonalCollection(item))
   );
 }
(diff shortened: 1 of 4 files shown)


npfitz (approved):

frontend/src/metabase/collections/utils.ts
```diff
@@ -140,6 +144,7 @@ export function canPreviewItem(item: CollectionItem, collection: Collection) {
 export function canMoveItem(item: CollectionItem, collection: Collection) {
   return (
     collection.can_write &&
+    !isReadOnlyCollection(item) &&
```
npfitz, line 147: That's cheeky. So previously we were only checking to see if we could write to the collection the item was in, not the item itself? If so, this probably impacted other things as well.

npfitz:
If it's an easy add, it would be nice to have an e2e test for this sort of thing

frontend/src/metabase/collections/utils.ts
iethree: I don't think there are other situations where people have read but not write access to collections 🤔

Issue #36228, opened 2023-11-30
Audit content is editable


luizarakaki:
**Context**
It is possible to edit Metabase analytics content
https://stats.metabase.com/dashboard/2102

iethree:
fixed by https://github.com/metabase/metabase/pull/36237

PR #36239, opened 2023-11-30
[Auditv2] Reproduce #36228 (analytics content is editable)

iethree:
reproduces https://github.com/metabase/metabase/issues/36228

### Description

Reproduces the fact that we allowed editing to instance analytics content

### How to verify

See that you can edit all models and dashboards in the metabase analytics collection

### DIFF
--- a/e2e/support/helpers/e2e-misc-helpers.js
+++ b/e2e/support/helpers/e2e-misc-helpers.js
@@ -163,7 +163,7 @@ export function visitDashboard(dashboard_id, { params = {} } = {}) {
     url: `/api/dashboard/${dashboard_id}`,
     // That's why we have to ignore failures
     failOnStatusCode: false,
-  }).then(({ status, body: { dashcards } }) => {
+  }).then(({ status, body: { dashcards, tabs } }) => {
     const dashboardAlias = "getDashboard" + dashboard_id;
 
     cy.intercept("GET", `/api/dashboard/${dashboard_id}`).as(dashboardAlias);
@@ -171,9 +171,12 @@ export function visitDashboard(dashboard_id, { params = {} } = {}) {
     const canViewDashboard = hasAccess(status);
 
     let validQuestions = dashboardHasQuestions(dashcards);
-    if (params.tab != null) {
+
+    // if dashboard has tabs, only expect cards on the first tab
+    if (tabs?.length > 0) {
+      const firstTab = tabs[0];
       validQuestions = validQuestions.filter(
-        card => card.dashboard_tab_id === params.tab,
+        card => card.dashboard_tab_id === firstTab.id,
       );
     }
 
--- a/e2e/test/scenarios/collections/instance-analytics.cy.spec.js
+++ b/e2e/test/scenarios/collections/instance-analytics.cy.spec.js
@@ -5,7 +5,7 @@ import {
   describeEE,
   modal,
   visitDashboard,
-  visitQuestion,
+  visitModel,
 } from "e2e/support/helpers";
 
 const ANALYTICS_COLLECTION_NAME = "Metabase analytics";
@@ -47,13 +47,9 @@ describeEE("scenarios > Metabase Analytics Collection (AuditV2) ", () => {
       { tags: "@flaky" },
       () => {
         cy.log("saving edited question");
-        visitCollection(ANALYTICS_COLLECTION_NAME);
-        cy.findByTestId("pinned-items")
-          .findByText(PEOPLE_MODEL_NAME)
-          .scrollIntoView()
-          .click();
-
-        cy.wait("@datasetQuery");
+        getItemId(ANALYTICS_COLLECTION_NAME, PEOPLE_MODEL_NAME).then(id => {
+          visitModel(id);
+        });
 
         cy.findByTestId("TableInteractive-root").within(() => {
           cy.findByText("Last Name").click();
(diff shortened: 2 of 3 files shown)


e2e/test/scenarios/collections/instance-analytics.cy.spec.js
```diff
@@ -142,5 +144,86 @@ describeEE("scenarios > Instance Analytics Collection", () => {
...
+
+      cy.findByTestId("qb-header").within(() => {
+        cy.icon("ellipsis").click();
+      });
+
+      popover().findByText("Edit query definition").should("not.exist");
+    });
+  });
+
+  describe("API tests", () => {
```
iethree, line 179: @nemanjaglumac would love to hear your thoughts on this. This is a little experimental.

- we make a series of API requests to check the `can_write` property on every single item in the analytics collection to ensure that it's set to `false`.
- since there's no UI involved, this test actually runs in a fraction of the time it would take to check a single item through the UI, and we get coverage on everything.
- above, we also do a standard UI test on one model and one dashboard.

WDYT? is this valuable? is this a good pattern?

e2e/test/scenarios/collections/instance-analytics.cy.spec.js
```diff
@@ -142,5 +144,86 @@ describeEE("scenarios > Instance Analytics Collection", () => {
...
+      cy.findByTestId("pinned-items")
+        .findByText(METRICS_DASHBOARD_NAME)
+        .scrollIntoView()
+        .click();
+
+      cy.wait("@getDashboard");
+
+      cy.findByTestId("dashboard-header").within(() => {
+        cy.icon("pencil").should("not.exist");
+      });
```
nemanjaglumac, line 160: Nit:

Within is needed only for multiple assertions that are scoped to a selector. You can easily chain `.icon()` directly here.

nemanjaglumac: WDYT about adding a positive assertion before this one just to avoid false positives?

e2e/test/scenarios/collections/instance-analytics.cy.spec.js
```diff
@@ -142,5 +144,86 @@ describeEE("scenarios > Instance Analytics Collection", () => {
...
+
+    it.skip("should not allow editing analytics content (#36228)", () => {
+      // dashboard
+      navigationSidebar().findByText(ANALYTICS_COLLECTION_NAME).click();
+      cy.findByTestId("pinned-items")
+        .findByText(METRICS_DASHBOARD_NAME)
+        .scrollIntoView()
+        .click();
+
+      cy.wait("@getDashboard");
```
nemanjaglumac, line 156: Is there a way to get the dashboard ID programmatically and to then use `visitDashboard()` helper?
It will greatly reduce the time it takes to run this test and, more importantly, it will automatically wait for the dashboard card results to load.

`@getDashboard` still doesn't know anything about dashboard cards.

e2e/test/scenarios/collections/instance-analytics.cy.spec.js
```diff
@@ -142,5 +144,86 @@ describeEE("scenarios > Instance Analytics Collection", () => {
         cy.contains("Move").should("not.exist");
       });
     });
+
+    it.skip("should not allow editing analytics content (#36228)", () => {
```
nemanjaglumac, line 148: Nit: only for consistency (although we can easily stop prefixing issue numbers with the repo name.

```suggestion
    it.skip("should not allow editing analytics content (metabase#36228)", () => {
```

nemanjaglumac (approved):
Left some inline comments. Nothing major.
This is good to go as is, if you want.

e2e/test/scenarios/collections/instance-analytics.cy.spec.js
iethree: I made a little helper for this and also used it in the flaky test above. Left the @flaky tag on for now though.

nemanjaglumac: In that case, maybe don't try to fix it in this PR.
That should be a separate issue and a separate fix.

e2e/support/helpers/e2e-misc-helpers.js
```diff
@@ -175,6 +175,12 @@ export function visitDashboard(dashboard_id, { params = {} } = {}) {
       validQuestions = validQuestions.filter(
         card => card.dashboard_tab_id === params.tab,
       );
+    } else if (tabs.length > 0) {
+      // if dashboard has tabs, but there is no tab param, default to using the first tab
+      const firstTab = tabs[0];
+      validQuestions = validQuestions.filter(
+        card => card.dashboard_tab_id === firstTab.id,
+      );
     }
```
iethree, line 184: @nemanjaglumac @EmmadUsmani check this out.

Visit dashboard wasn't working with instance analytics dashboards because they have tabs now, and we were awaiting all the dashcards to load, but only some of them load because we only load the first tab.

I added logic to visitDashboard such that if the dashboard has tabs, and you don't pass a specific tab id param, then we just expect the first tab to be loaded.

The pattern of setting a specifc tab id is kinda fragile, and hard to use since they're globally unique - so you can't even visit tab 1 on every dashboard. I wonder if we should remove this tabs param altogether?

nemanjaglumac: I really really really don't think we should mess with a global helper that's used in hundreds of tests here.
Can that be a separate PR, pretty please?

iethree: sure

PR #36241, opened 2023-11-30
Fetching specific collections for file upload toasts

### LINKED ISSUE #36215: FileUploadStatus component loads entire collection list
### Describe the bug

FileUploadStatus (which is used for CSV uploads) makes a request to get the entire list of collections so that it can match them up to ongoing uploads. This is a very inefficient way to do this, as there will likely only be a couple uploads happening at any given moment, and the collections request can be somewhat slow when there is a large number of collections (or users). We should only fetch data for collections that currently have an upload happening.

### To Reproduce

1. Go to the home page with the network inspector open
2. See that we hit /collections

### Expected behavior

We should not hit /collections when it's not needed

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "Win32",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.20.1+1-post-Ubuntu-0ubuntu122.04",
    "java.vendor": "Ubuntu",
    "java.vendor.url": "https://ubuntu.com/",
    "java.version": "11.0.20.1",
    "java.vm.version": "11.0.20.1+1-post-Ubuntu-0ubuntu122.04",
    "os.name": "Linux",
    "os.version": "5.15.133.1-microsoft-standard-WSL2",
    "user.language": "en",
    "user.timezone": "America/Halifax"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2023-11-22",
      "src_hash": "7127135b884948f7ae1b4c64e3a7575b47cf2bc9",
      "tag": "v1.47.1-SNAPSHOT",
      "hash": "c09abb0"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

Depends on the number of collections an instance has, but for larger instances this can really affect load times (or cause crashes)

### Additional context

_No response_

npfitz:
Closes https://github.com/metabase/metabase/issues/36215

### Description
The `<FileUploadStatus />` component was fetching the entire list of collections in order to determine the name of the collection to show in the toast. This can have a very negative effect on larger instances since all collections includes personal collections, as well as anything in `Out Analytics`. Locally, this request would take a couple seconds to complete for an instance with ~10,000 users, but would not result in anything showing on the screen for most people.

### How to verify
Go to the homepage and open up the network tab. You shouldn't see a request to `/collections`. If you do a CSV upload, you should only potentially see a request to the specific collection you're uploading to.

### Demo
chrome_myeg8Qp8oE

### Checklist

### DIFF
--- a/frontend/src/metabase/status/components/FileUploadStatus/FileUploadStatus.tsx
+++ b/frontend/src/metabase/status/components/FileUploadStatus/FileUploadStatus.tsx
@@ -1,39 +1,33 @@
+import _ from "underscore";
+
 import { useSelector, useDispatch } from "metabase/lib/redux";
 import { getAllUploads, clearAllUploads } from "metabase/redux/uploads";
-import Collections from "metabase/entities/collections/collections";
-import type { Collection } from "metabase-types/api";
+import type { CollectionId } from "metabase-types/api";
 import type { FileUpload } from "metabase-types/store/upload";
 import { isUploadAborted, isUploadInProgress } from "metabase/lib/uploads";
 
+import { useCollectionQuery } from "metabase/common/hooks";
 import useStatusVisibility from "../../hooks/use-status-visibility";
 import FileUploadStatusLarge from "../FileUploadStatusLarge";
 
-const FileUploadStatus = ({
-  collections = [],
-}: {
-  collections: Collection[];
-}) => {
+export const FileUploadStatus = () => {
   const uploads = useSelector(getAllUploads);
   const dispatch = useDispatch();
   const resetUploads = () => dispatch(clearAllUploads());
 
-  const uploadCollections = collections.filter(collection =>
-    uploads.some(upload => upload.collectionId === collection.id),
-  );
+  const groupedCollections = _.groupBy(uploads, "collectionId");
+
+  const collections = Object.keys(groupedCollections) as CollectionId[];
 
   return (
     <>
-      {uploadCollections.map(collection => {
-        const collectionUploads = uploads.filter(
-          ({ collectionId }) => collectionId === collection.id,
-        );
-
+      {collections.map(collectionId => {
         return (
           <FileUploadStatusContent
-            key={`uploads-${collection.id}`}
-            uploads={collectionUploads}
+            key={`uploads-${collectionId}`}
+            uploads={groupedCollections[collectionId]}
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/status/components/FileUploadStatus/FileUploadStatus.tsx
```diff
@@ -1,39 +1,33 @@
...
   const uploads = useSelector(getAllUploads);
   const dispatch = useDispatch();
   const resetUploads = () => dispatch(clearAllUploads());
 
-  const uploadCollections = collections.filter(collection =>
-    uploads.some(upload => upload.collectionId === collection.id),
-  );
+  const groupedCollections = _.groupBy(uploads, "collectionId");
+
+  const collections = Object.keys(groupedCollections) as CollectionId[];
```
npfitz, line 20: unfortunately, underscore assumes that the keys pulled from `groupBy` will always be strings. This little bit of casting was the nicest way to get everything to play nicely.

frontend/src/metabase/status/components/FileUploadStatus/FileUploadStatus.tsx
```diff
@@ -1,39 +1,33 @@
...
+import { useCollectionQuery } from "metabase/common/hooks";
 import useStatusVisibility from "../../hooks/use-status-visibility";
 import FileUploadStatusLarge from "../FileUploadStatusLarge";
 
-const FileUploadStatus = ({
-  collections = [],
-}: {
-  collections: Collection[];
-}) => {
+export const FileUploadStatus = () => {
```
npfitz, line 13: Took the opportunity to use a named export

shaunlebron (approved):

PR #36243, opened 2023-11-30
Fix warning copy when changing table permissions

### LINKED ISSUE #34911: Confusing copy in confirm modal when changing table data access to Sandboxed
### Describe the bug

The copy seems inconsistent with the terms we use:

image
image

I would expect the wording "Granular" if it's about changing the access level for the entire DB or "sandboxed" if it's about the data access level for the table. I'm able to trigger this just by changing data access at the table level, so I am guessing the sentence is also wrong. I'm also able to trigger this by changing the level to unrestricted, which makes no sense.

### To Reproduce

1. Go to Permissions/Data
2. Select a group
3. Select sample database
4. Change data access for one of the tables to Sandboxed
5. The modal will pop

### Expected behavior

"Change access to this table to Sandboxed?" when switching to sandboxed, "Change access to this table to Unrestricted?" when changing to unrestricted, etc.

### Logs

_No response_

```JSON
- Metabase v1.47.4, Built on 2023-10-11
```

### Severity

Low

### Additional context

_No response_

shaunlebron:
Small copy fix for #34911 

localhost_3000_admin_permissions_data_group_1_database_1

### DIFF
--- a/frontend/src/metabase/admin/permissions/selectors/confirmations.ts
+++ b/frontend/src/metabase/admin/permissions/selectors/confirmations.ts
@@ -6,10 +6,6 @@ import {
   getNativePermission,
   getSchemasPermission,
 } from "metabase/admin/permissions/utils/graph";
-import {
-  isSchemaEntityId,
-  isTableEntityId,
-} from "metabase/admin/permissions/utils/data-entity-id";
 import type {
   Group,
   GroupsPermissions,
@@ -107,13 +103,8 @@ export function getControlledDatabaseWarningModal(
     getSchemasPermission(permissions, groupId, entityId, "data") !==
     "controlled"
   ) {
-    const entityType = isTableEntityId(entityId)
-      ? t`table`
-      : isSchemaEntityId(entityId)
-      ? t`schema`
-      : t`database`;
     return {
-      title: t`Change access to this ${entityType} to limited?`,
+      title: t`Change access to this database to granular?`,
       confirmButtonText: t`Change`,
       cancelButtonText: t`Cancel`,
     };
--- a/frontend/src/metabase/admin/permissions/selectors/confirmations.ts
+++ b/frontend/src/metabase/admin/permissions/selectors/confirmations.ts
@@ -104,7 +104,7 @@ export function getControlledDatabaseWarningModal(
     "controlled"
   ) {
     return {
-      title: t`Change access to this database to limited?`,
+      title: t`Change access to this table to limited?`,
       confirmButtonText: t`Change`,
       cancelButtonText: t`Cancel`,
     };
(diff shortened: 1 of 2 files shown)


iethree (changes requested):
I think we need a little more logic to make this correct, as far as I can tell, this modal can pop up for both schema and tables.

schemaChange

I wasn't able to trigger it for databases, but I'm not 100% sure whether or not we also need to handle that.

Probably out of scope of this change, but I noted that we also get this warning when we _loosen_ permissions on a table, which is a bit of a head-scratcher.
increase access

shaunlebron:
I was able to verify that both `TableEntityId`s and `SchemaEntityId`s are passed to this function, so I rewrote it to name it correctly in the copy: e961210

The head-scratcher will be a bigger fix.  Notes here: https://github.com/metabase/metabase/issues/34911#issuecomment-1836421574

alxnddr (commented):
The idea behind this warning is to give heads up to user that the **database** level permission implicitly will change when they modify table or schema permission. The wording is quite wrong, I would replace "Limited" with "Granular":
`Change access to this database to granular?`

As Ryan pointed out, on this branch if you have "No self-service" level on a DB and you change a single table permission to "Unrestricted" the modal will say "Change access to this table to limited?" which is not correct

alxnddr:
> I wasn't able to trigger it for databases, but I'm not 100% sure whether or not we also need to handle that.

Yep, that seems intentional. When users modifying a DB-level permission it is clear what it will be. But if they change schema/table level permissions, they may not understand the DB-level permission implicitly changes to "granular"

alxnddr (approved):

iethree (changes requested):
Let's use the updated copy that @jeff-bruemmer proposed to make this a lot more explanatory:

https://metaboat.slack.com/archives/C064EB1UE5P/p1701885966932019?thread_ts=1701883832.195089&cid=C064EB1UE5P

> Just letting you know that changing the permission setting on this (table|schema) will also update the database permission setting to “Granular” to reflect that some of the database’s (schemas|tables) have different permission settings.

frontend/src/metabase/admin/permissions/selectors/confirmations.ts
```diff
@@ -103,8 +107,14 @@ export function getControlledDatabaseWarningModal(
     getSchemasPermission(permissions, groupId, entityId, "data") !==
     "controlled"
   ) {
+    const [entityType, entityTypePlural] = isTableEntityId(entityId)
+      ? [t`table`, t`tables`]
+      : isSchemaEntityId(entityId)
+      ? [t`schema`, t`schemas`]
+      : [t`entity`, t`entities`];
```
iethree, line 114: this is perfectly fine, but we also have a `pluralize` function in metabase/lib/formatting/strings.js`

iethree (approved):
✨

PR #30059, opened 2023-04-12
Don't allow unscoped text selectors in e2e tests

iethree:
[Approved RFC](https://www.notion.so/metabase/52-Disallow-top-level-findByText-in-e2e-tests-9cef8dc4a77a49dfb2e41392913b7281)

### Description

We frequently run into a problem with unscoped `cy.findByText()` and `cy.contains()` calls in our e2e tests. They're difficult to understand and are very error prone when other elements in the application change.  Most of the problems with these calls are ameliorated by simply scoping them to a container.

For example: 

❌  Bad 👇 
```js
it('should run the query correctly', () => {
  cy.findByText('42');
});

it('should run the query correctly', () => {
  cy.contains('Save');
});
```

✅  Better 👇 
```js
it('should run the query correctly', () => {
  cy.findByTestId('table-interactive-wrapper').within(() => {
    cy.findByText('42');
  });
});

// also 

cy.findByTestId('table-interactive-wrapper').findByText('42');

```
This adds a pretty simple lint rule that disables using `FindByText` and `contains` at the top-level of a test block, requiring it to be chained in some way or within some other selector block.

## How to test
1. Run the tests! `node frontend/lint/eslint-rules/no-unscoped-text-selectors.unit.spec.js`
2. Open up an e2e file and write a bad selector.

## Implementation 

Fixing the existing **3151** violations of this rule is prohibitively difficult, and probably less valuable than fixing all of them entails. We want to enforce this rule going forward, and hopefully clean up some of the existing violations as we touch those files or tests.

I considered several ways to implement this
1. 👎 Limiting the total number of eslint errors. Unfortunately, this cannot be done on a rule-by-rule basis, so this could allow other linting violations to slip through if we fix/delete some of these violations. 😢 
2. 👎 Using a pre-commit hook to enforce this. This would be fine for individual commits, but it would require changes to how we lint all code in PRs and on the master branch. This might be hard to maintain or follow down the road.
3. 👎 Changing all violative uses to something like `cy._DEPRECATED_contains('bananas')`. This was Ariya's suggestion, and makes a lot of sense. I decided against it for 2 reasons: (a) the codemod could be a little tricky to write, and (b) I think it might be vulnerable to lazy copypasta continuing its usage.
[...]

PR #36244, opened 2023-11-30
Allow toggling of the dashboard subscription sidebar from the icon

kdoh:
### What
Updates the behavior of clicking the subscriptions icon in the dashboard header so that the sidebar gets toggled, instead of you having to go all the way down to `Cancel.` This now matches the behavior of the info panel.

### Demo
https://www.loom.com/share/465134c0f6ba4ffd9408a4dade4d9d67?sid=246a8af4-3112-4c88-b502-d8260d3de8cd

### Checklist

### DIFF
--- a/e2e/test/scenarios/sharing/subscriptions.cy.spec.js
+++ b/e2e/test/scenarios/sharing/subscriptions.cy.spec.js
@@ -69,13 +69,13 @@ describe("scenarios > dashboard > subscriptions", () => {
   describe("sidebar toggling behavior", () => {
     it("should allow toggling the sidebar", () => {
       openDashboardSubscriptions();
-      cy.findByLabelText("subscriptions").click();
 
-      // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-      cy.findByText("Email it");
       cy.findByLabelText("subscriptions").click();
       // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
       cy.findByText("Email it").should("not.exist");
+      cy.findByLabelText("subscriptions").click();
+      // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
+      cy.findByText("Email it");
     });
   });
 
--- a/e2e/test/scenarios/sharing/subscriptions.cy.spec.js
+++ b/e2e/test/scenarios/sharing/subscriptions.cy.spec.js
@@ -66,6 +66,19 @@ describe("scenarios > dashboard > subscriptions", () => {
     cy.findByText(/Share this dashboard with people *./i);
   });
 
+  describe("sidebar toggling behavior", () => {
+    it("should allow toggling the sidebar", () => {
+      openDashboardSubscriptions();
+      cy.findByLabelText("subscriptions").click();
+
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -142,7 +142,7 @@ class DashboardInner extends Component {
   };
 
   onSharingClick = () => {
-    this.props.setSharing(true);
+    this.props.setSharing(this.props.isSharing ? false : true);
```
kulyk, line 145: I think a more common way to implement "toggle" behavior is to do `setState(!currentState)`, like:

```js
this.props.setSharing(!this.props.isSharing)
```

e2e/test/scenarios/sharing/subscriptions.cy.spec.js
```diff
@@ -66,6 +66,19 @@ describe("scenarios > dashboard > subscriptions", () => {
...
+  describe("sidebar toggling behavior", () => {
+    it("should allow toggling the sidebar", () => {
+      openDashboardSubscriptions();
+
+      cy.findByLabelText("subscriptions").click();
+      // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
+      cy.findByText("Email it").should("not.exist");
+      cy.findByLabelText("subscriptions").click();
+      // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
+      cy.findByText("Email it");
```
kulyk, line 78: Let's not use deprecated practices in new code. We expect `findByText` to be wrapped in `within` (more in #30059). But in this case, maybe we can assert that an `aside` disappears after we click the "subscriptions" button again? For example:

```js
sidebar().should("not.exist");
cy.findByLabelText("subscriptions").click();
sidebar().findByText("Email it").should("exist");
```

e2e/test/scenarios/sharing/subscriptions.cy.spec.js
```diff
@@ -66,6 +66,19 @@ describe("scenarios > dashboard > subscriptions", () => {
...
+    });
```
kulyk, line 79: We're doing this at the moment:

1. Open the subscriptions sidebar (with `openDashboardSubscriptions`)
2. Close it
3. Assert it disappeared
4. Open it again
5. Assert it appeared

Steps 4 and 5 seem redundant; we should be able to assert the sidebar is open after step 1

kulyk (commented):
Looks good, inlined a couple of comments about the test

e2e/test/scenarios/sharing/subscriptions.cy.spec.js
kdoh: Fair. I guess I was going a bit overboard with "toggling" here and closing it after it's initially open via `openDashboardSubscriptions` does test the behavior.

e2e/test/scenarios/sharing/subscriptions.cy.spec.js
kdoh: Ah ok I see the better way to do it down below in the file, was looking at the tests up above. Easy to fix.

kulyk (approved):

PR #36245, opened 2023-11-30
Document and test adding `results_metadata` to qp middleware for static viz

### LINKED ISSUE #36153: Merge metadata correctly

markbastian:
Modifying PR to document and use existing fns:
    
The current qp pipeline will correctly propagate custom metadata if a question is derived from a model or if the model is processed like so:
    
    ```clojure
    (qp/process-query
      (assoc-in dataset_query [:info :metadata/dataset-metadata] result_metadata))
    ```
    
This PR fixes rendering unit tests to clarify this and provides a much better working example in the `dev.render-png` ns.

In particular the `dev.render-png` ns has been updated for correctness and tests in `metabase.pulse.render-test` for clarity.

Fixes #36153

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -337,6 +337,7 @@
     metabase.pulse.render.body                                      body
     metabase.pulse.render.common                                    common
     metabase.pulse.render.style                                     style
+    metabase.pulse.test-util                                        pulse.test-util
     metabase.pulse.parameters                                       params
     metabase.query-processor-test                                   qp.test
     metabase.query-processor.context                                qp.context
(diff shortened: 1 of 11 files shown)


metamben (approved):
LGTM

adam-james-v (approved):
In this PR's description you mentioned some middleware. 
I don't actually see it in the file changes, and I think that's because you don't need them right?

Instead, what already exists is the ability to do this:  `(assoc-in model-query [:info :metadata/dataset-metadata] model-metadata)`.

Am I understanding that correctly? 
I'm approving this as well because as far as I understand it, it's just a slightly out of date PR description.

Issue #33713, opened 2023-09-05
[Epic] Auto-gen CSV PKs


tsmacdonald:
**Links**
- product doc: [_link to product doc_](https://www.notion.so/metabase/Auto-generate-assign-PKs-for-CSV-uploads-b083f9aa198c4040bc09bb1214321d74)

**Implementation Plan**

***Milestone 1***

- [x] #33714

***Milestone 2***
- [x] #34592

PR #36249, opened 2023-11-30
Rework auto-generated PKs for CSV uploads

### LINKED ISSUE #36268: Rework auto-generated PKs for CSV uploads
I started planning the BE implementation of [Allow appending more data to existing CSV uploads](https://www.notion.so/metabase/Allow-appending-more-data-to-existing-CSV-uploads-f1f13864632d4cfd83522fce762890eb) and I realised there is an issue we overlooked in [Auto-generate / assign PKs for CSV uploads](https://www.notion.so/metabase/Auto-generate-assign-PKs-for-CSV-uploads-b083f9aa198c4040bc09bb1214321d74) that interacts with appends. The issue is urgent to solve because this feature is about to land in 48 and will complicate the appends project and create a bug that we have no great solution to.

**The context from [Auto-generate / assign PKs for CSV uploads](https://www.notion.so/metabase/Auto-generate-assign-PKs-for-CSV-uploads-b083f9aa198c4040bc09bb1214321d74):**
1. If a column in a CSV upload is labelled “id” or “pk” in the originally uploaded CSV that has unique values, we’ll make it a PK of the uploaded table.
2. Otherwise, we’ll create a column called “id” that will have autogenerated integers in it and is a PK of the uploaded table.

**The problem:**
Appends might not work with columns created by (1), because the user might append data that breaks the uniqueness constraint of the columns called “id” or “pk”. Consider a simple example: initially a user uploads a CSV file with a column called “ID” and a row with the value 1 for that ID column. That column will become a PK. All good for now, because the ID column is unique in the initial upload. But the problem happens when the user then appends to the table: imagine they upload another file with a row that has value 1 for that column. There will be an exception when we try to insert that row into the database.

**The solution:** ([Slack context](https://metaboat.slack.com/archives/C04S696LRUM/p1701357142724109))
Rework the changes in [[Epic] Auto-gen CSV PKs](https://github.com/metabase/metabase/issues/33713#top) so that we don’t try to infer the PK from the column names. We need to create a unique PK column instead that is clearly managed by metabase, doesn’t have a meaning outside of identifying rows, and cannot be appended to by uploaded data. A PK column with auto-generated integer values, called `_mb_row_id`, will be added to the table as well as the columns from the CSV on upload. We will ignore any columns in the CSV that have a name that will be in the DB with the same name as the auto-generated PK column (e.g. "_MB row-ID" in the CSV will be `_mb_row_id` or `_MB_ROW_ID` in the DB).

The column will not be required for appends, and users should not be allowed to edit it themselves.

calherries:
Resolves https://github.com/metabase/metabase/issues/36268

This PR reworks the changes in https://github.com/metabase/metabase/issues/33713#top so that we don’t try to infer the PK from the column names.

**How it used to work:**

One of two things used to happen:
- If a column in a CSV upload is labelled “id” or “pk” in the originally uploaded CSV that has unique values, we’ll make it a PK of the uploaded table.
- Otherwise, we’ll create a column called `id` that will have auto-generated integer values and is a PK of the uploaded table.

**How it works now:**

The upload code treats columns with the name "id" or "pk" in the CSV the same as all the others. On upload, we always create a unique PK column `_mb_row_id` with auto-generated integer values, called `_mb_row_id`, that will be added to the table along with the columns from the CSV on upload. We ignore any columns in the CSV that have a name that will be in the DB with the same name as the auto-generated PK column (e.g. "_MB row-ID" in the CSV will be `_mb_row_id` or `_MB_ROW_ID` in the DB).

**Specifics of the implementation:**
- The "upload types" `:metabase.upload/string-pk` and `:metabase.upload/int-pk` aren't needed anymore. These upload types were specifically for columns named "id" or "pk".


src/metabase/upload.clj
```diff
@@ -306,24 +321,18 @@
...
-                                 (count col-to-insert->upload-type))
-           :size-mb           (/ (.length csv-file)
-                                 1048576.0)}))
-      (catch Throwable e
-        (driver/drop-table! driver db-id table-name)
-        (throw (ex-info (ex-message e) {:status-code 400}))))))
+    (with-open [reader (io/reader csv-file)]
+      (let [rows (->> (csv/read-csv reader)
+                      (drop 1) ; drop header
+                      parse-rows)]
```
calherries, line 332: The main change here is `parse-rows` is now a closure that is returned by `detect-schema`, and will remove any columns from the rows that have a column name that matches `_mb_row_id`.

src/metabase/upload.clj
```diff
@@ -209,53 +195,76 @@
...
-                               (map normalize-column-name)
-                               (mbql.u/uniquify-names)
+                               (map normalize-column-name))
+        ;; remove columns from the rows with the same normalized name as the auto-pk-column-name
+        auto-pk-col-indices (set (indices-where #(= auto-pk-column-name %) normalized-header))
+        ;; remove the auto-pk column from each row as we parse all the rows
+        remove-auto-pk-cols (fn [row]
+                              (remove-indices auto-pk-col-indices row))
+        rows              (cond->> rows
+                            auto-pk-col-indices
```
qnkhuat, line 241:
```suggestion
                            (seq auto-pk-col-indices)
```

empty set is considered truthy under cond

src/metabase/upload.clj
```diff
@@ -306,24 +321,18 @@
...
-          {:num-rows          (count rows)
-           :num-columns       (count csv-col-names)
-           :generated-columns (- (count col-to-create->col-spec)
-                                 (count col-to-insert->upload-type))
-           :size-mb           (/ (.length csv-file)
-                                 1048576.0)}))
-      (catch Throwable e
-        (driver/drop-table! driver db-id table-name)
-        (throw (ex-info (ex-message e) {:status-code 400}))))))
+    (with-open [reader (io/reader csv-file)]
```
qnkhuat, line 329: why did you remove the try-catch?

src/metabase/upload.clj
```diff
@@ -209,53 +195,76 @@
...
+                                 (reduce coalesce-types (repeat column-count nil))
+                                 (map #(or % ::text))
+                                 (map vector unique-header))]
+    {:extant-columns    (ordered-map/ordered-map col-name+type-pairs)
+     :generated-columns (ordered-map/ordered-map (keyword auto-pk-column-name) ::auto-incrementing-int-pk)
+     :parse-rows        (fn [rows]
+                          (let [rows (cond->> rows
+                                       auto-pk-col-indices
+                                       (map remove-auto-pk-cols))
+                                col-types (map second col-name+type-pairs)]
```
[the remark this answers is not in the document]
calherries: Yeah, that's something I've been wanting to fix for a while. Currently the `rows` that are passed to `rows->schema` are
just a sample of the rows, while the rows that get passed to `parse-rows` is the full set. You can probably move the
sampling into this function and then you can have parsing rows and getting the schema in one function. I think we were
being too clever trying to split them into separate functions, and ended up making it more complicated.

src/metabase/upload.clj
calherries: Oops, I removed it for development and forgot to put it back

src/metabase/upload.clj
calherries: Fixed

qnkhuat (approved):
LGTM now, I have a suggestion but it's for future refactoring.

also nice to see that's the PR is  +269 −269

PR #36255, opened 2023-11-30
Adding pagination to All Personal Collections page

### LINKED ISSUE #29177: Fix performance issues when loading all users' personal collections for large numbers of personal collections
When a large number of personal collections exist the "Other users' personal collections" menu item takes a very long time load. The below gif shows this. The solution is probably to introduce paging to the "/" endpoint in the `metabase.api.collection` ns on the backend and integrating the paged endpoint on the frontend.

Image

This is related to #28190 and #24330. The only fix that went out with those was the initial on load performance fix.

npfitz:
Closes https://github.com/metabase/metabase/issues/29177

### Description
This PR makes a few changes to the Other Users Personal Collections page. First, it adds pagination to the users call so that we don't fetch all users at once. Secondly we use regular urls rather than slugified URLs for the links, which removes the requirement to fetch all collections. These 2 changes should dramatically improve the experience of anyone that uses this page in larger instances

### How to verify
1. Navigate to the All Personal Collections page
2. Ensure that pagination is working as expected
3. notice in the network tab that we don't call `/collections` (for this page), and that we grab users 27 at a time.

### Demo

### Checklist


shaunlebron (changes requested):
> notice in the network tab that we don't call /collections (for this page), and that we grab users 27 at a time.

It looks like `/collection` is still pulling all the collections on the first page.  Can you take a look?

CleanShot 2023-12-01 at 11 30 40@2x

npfitz:
Rebase complete. Do you want to take a look now @shaunlebron?

shaunlebron (approved):

PR #36257, opened 2023-11-30
MB_CLOUD_GATEWAY_IPS is settable from env var

### LINKED ISSUE #36131: Get cloud-gateway-ips from an environment variable
**Context**
https://metaboat.slack.com/archives/C010ZSXQY87/p1700622228090499

We now support multi-region Cloud. Each region has their own gateway-ips.

image

We are currently fetching cloud gateway IPs from a static json file via an api request to the store (https://github.com/metabase/metabase/blob/ddfb64e9ff46f931a6496da07b3962f8441d5689/src/metabase/public_settings.clj#L671C8-L671C8) 

Since we launched the multi-region project those IPs are different in each hosting region, and non-US customers are seeing the wrong IPs in the UI. We should be reading them from an env_var which will be set by harbormaster when creating the instance. Note, the env_var will have a coma delimited list of IPs, not a single value. 

The name of the env var is `MB_CLOUD_GATEWAY_IPS` and looks like: `1.2.3.4,5.6.7.8`

Depends on: https://github.com/metabase/harbormaster/issues/4445

escherize:
resolves #36131
`MB_CLOUD_GATEWAY_IPS` can be set to a comma delimited sequence of ips like: 

```shell
MB_CLOUD_GATEWAY_IPS='1.2.3.4,5.6.7.8' java -jar metabase.jar
```

Note: even if the environment variable is set, the setting will return nil if the is-hosted? premium feature is not true (there is a test for that).


src/metabase/public_settings.clj
```diff
@@ -661,33 +658,15 @@
...
 (defsetting cloud-gateway-ips
   (deferred-tru "Metabase Cloud gateway IP addresses, to configure connections to DBs behind firewalls")
   :visibility :public
-  :type       :json
-  :setter     :none
-  :getter     (fn []
-                (when (premium-features/is-hosted?)
-                  (fetch-cloud-gateway-ips-fn))))
+  :type       :string
+  ;;:setter     :none
```
escherize, line 665: todo: uncomment this

src/metabase/public_settings.clj
```diff
@@ -661,33 +658,15 @@
...
+  :getter (fn []
+            (when (premium-features/is-hosted?)
+              (some-> (setting/get-value-of-type :string :cloud-gateway-ips)
+                      (str/split #",")))))
```
escherize, line 669: I was going to add some more validation on these, but who knows if they will want ipv6, so I just let the possibly incorrect ips pass on through.

src/metabase/public_settings.clj
noahmoss: need to uncomment before merging?

noahmoss (approved):

Issue #35691, opened 2023-11-14
Adding Tests for Isomorphic Chart Static Viz


adam-james-v:
The isomorphic static-viz chart as part of the ECharts project required at first that several backend tests be removed. This is a PR that tracks which of those tests was removed so that new equivalents can be written.

## js-svg-test
some tests failing in `metabase.pulse.render.js-svg-test` because it's trying to run `js-svg/combo-chart` which doesn't exist anymore

removed the following:
- [ ] timelineseries-line-test
  - returns bytes
  - returns valid svg string
  - svg string has formatted axes
- [ ] timelineseries-bar-test
  - returns bytes
  - returns valid svg string
  - svg string has formatted axes
- [ ] area-test
  - returns bytes
  - returns valid svg string
  - svg string has formatted axes
- [ ] combo-test
  - it returns bytes
  - it returns a valid svg string
- [ ] goal-line-test
  - no goal line exists when there are no goal settings (referring to visualization_settings)
  - goal line does exist when there are goal settings

Notes about some of the above assertions:
- 'it returns bytes' this test was never particularly great, especially since if there's a render error it renders error text to a png, so we don't catch it with this test
- 'it returns a valid svg string' --> means there are no html tags (eg. div or p or span)
- several of the more specific assertions (eg. 'svg string has formatted axes') requires some helpers that use js-svg/parse-string and recreate a portion of the render pipeline. This then lets you write assertions on the svg data instead of only checking that png bytes exist.

Here are potentially relevant helpers:

```clojure
(def parse-svg #'js-svg/parse-svg-string)
    
(defn document-tag-hiccup [^SVGOMDocument document]
  (letfn [(tree [^Node node]
            (into [(.getNodeName node)]
                  (if (instance? org.apache.batik.dom.GenericText node)
                    [(.getWholeText ^org.apache.batik.dom.GenericText node)]
                    (map tree
                         (when (instance? Element node)
                           (let [children (.getChildNodes node)]
                             (reduce (fn [cs i] (conj cs (.item children i)))
                                     [] (range (.getLength children)))))))))]
    (tree (.getDocumentElement document))))

(defn document-tag-seq [^SVGOMDocument document]
  (map #(.getNodeName ^Node %)
       (tree-seq #(instance? Element %)
                 (fn [^Node node]
                   (let [children (.getChildNodes node)]
                     (reduce (fn [cs i] (conj cs (.item children i)))
                             [] (range (.getLength children)))))
[...]

PR #36259, opened 2023-12-01
Improve isomorphic chart rendering backend tests

adam-james-v:
- add a chart render test that catches the placeholder. This will fail when implementations are merged in, which is good, we can write proper tests then.

 - add `render-card-as-hiccup` which takes a card-id and runs the static-viz rendering pipeline with 2 fns redefined to keep the svg output of the card as svg, not rendered into png bytes. This ultimately lets us make assertions on hiccup data, which lets us check contents of the renders more exactly.

- start removing code and test utils that were never a great idea (thanks past me!). There is a convoluted set of functions to try mock card data and viz data in render.test-utils. Instead, most of that can go away in favour of using mt/with-temp, mt/dataset, etc. Following commits will work towards this goal.
 
- [ ] make sure `render.test-utils/render-card-as-hiccup` works for all possible static-viz renders (because `:table` display type doesn't use the graaljs interpreter, the util function might make too many assumptions, so just have to double check.
- [ ] clear lots of code from `render.test-utils` in favour of more straightforward metabase.test mocking approach.
- [ ] add back in tests listed in #35691


adam-james-v:
Moving this effort here: https://github.com/metabase/metabase/pull/36386 where we can merge into master and use beyond the context of just ismorphic charts. 

This should help with PR sequencing so that the right utils and dependencies are in master.

PR #36267, opened 2023-12-01
Pass step as only parameter with query, legacyQuery and stageIndex

uladzimirdev:
`step` already has query, topLevelQuery and stageIndex, so why do we need to duplicate this info using other props?


kamilmielnik (approved):
LGTM but let's get approval from Alex as well

ranquild (approved):

PR #28075, opened 2023-02-03
Statically checked unused var in TypeScript files

WiNloSt:
## Motivation
There have been too many occurrences where we forgot to remove unused imports and variables in TypeScript files. I'm not totally sure why we haven't set this up earlier. We already have this set up for JavaScript but never TypeScript for some reason.

## How to verify
1. Run Metabase locally and everything should work the same.

PR #36269, opened 2023-12-01
Fix all unused arg errors

WiNloSt:
Related https://github.com/metabase/metabase/pull/28075
### Description

While reviewing https://github.com/metabase/metabase/pull/36107, I noticed a few args are unused but not caught by our linter

### How to verify

Metabase should run fine. As nothing should change behavior-wise.

### Checklist

- [ ] ~Tests have been added/updated to cover changes in this PR~ No tests since this is change only modifies unused arguments


frontend/src/metabase/query_builder/components/view/ViewFooter.jsx
```diff
@@ -115,7 +115,7 @@ const ViewFooter = ({
               result={result}
             />
           ),
-          QueryDownloadWidget.shouldRender({ result, isResultDirty }) && (
+          QueryDownloadWidget.shouldRender({ result }) && (
```
WiNloSt, line 118: `QueryDownloadWidget.shouldRender` doesn't use `isResultDirty`.

frontend/src/metabase/redux/uploads.ts
```diff
@@ -44,59 +44,58 @@ export const hasActiveUploads = (state: State) =>
 
 export const uploadFile = createThunkAction(
   UPLOAD_FILE_TO_COLLECTION,
-  (file: File, collectionId: CollectionId) =>
-    async (dispatch: Dispatch, getState: GetState) => {
-      const id = Date.now();
+  (file: File, collectionId: CollectionId) => async (dispatch: Dispatch) => {
```
WiNloSt, line 47: This is just removing `getState: GetState`

frontend/src/metabase/collections/components/CollectionHeader/tests/setup.tsx
```diff
@@ -23,15 +23,6 @@ const getProps = (
   onCreateBookmark: jest.fn(),
   onUpload: jest.fn(),
   onDeleteBookmark: jest.fn(),
-  location: {
```
WiNloSt, line 26: `location` was unused in `frontend/src/metabase/collections/components/CollectionHeader/CollectionHeader.tsx`

frontend/src/metabase/query_builder/components/DataSelector/DataSelectorTablePicker/DataSelectorTablePicker.styled.tsx
```diff
@@ -23,9 +22,7 @@ export const DataSelectorTablePickerHeaderClickable = styled.span<Props>`
   align-items: center;
   display: flex;
 
-  ${({ onClick }) => css`
-    cursor: pointer;
-  `}
+  cursor: pointer;
```
oleggromov, line 25: What's this change?

frontend/src/metabase/redux/uploads.ts
oleggromov: Probably in this particular case for the sake of readability (and less meaningless changes), it's better to make it `_getState`?

oleggromov (approved):
Love it, except one (or a few if I missed the other ones) case where it's better to rename the argument instead of removing it to produce a cleaner diff. 

Yeah, and I'd try to keep out unrelated changes too!

frontend/src/metabase/redux/uploads.ts
WiNloSt: There are many places that we use either
- `(dispatch: Dispatch)` or
- `dispatch =>`

You can search by both strings to see those. I don't think leaving `getState` is a good idea if we're not going to use it otherwise, we'll have a ton of unused `getState` everywhere.

frontend/src/metabase/query_builder/components/DataSelector/DataSelectorTablePicker/DataSelectorTablePicker.styled.tsx
WiNloSt: I guess we used to use `onClick` in the styled component, but not anymore.  This change just remove the use of `css` util since it's unnecessary.

.eslintrc
```diff
@@ -175,7 +175,8 @@
           "error",
           {
             "vars": "all",
-            "args": "none",
+            "args": "after-used",
```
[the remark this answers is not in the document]
uladzimirdev: both first options are default btw

frontend/src/metabase/lib/redux/hooks.unit.spec.tsx
```diff
@@ -43,7 +43,7 @@ describe("useDispatch", () => {
     it("should provide a `dispatch` method that can dispatch a thunk", () => {
       const funcInThunk = jest.fn();
 
-      setup({ thunk: () => (dispatch: any, getState: any) => funcInThunk() });
+      setup({ thunk: () => (_dispatch: any, _getState: any) => funcInThunk() });
```
uladzimirdev, line 46: it looks like in this file all the unused params can be removed

uladzimirdev (approved):
🏅

.eslintrc
```diff
@@ -175,7 +175,8 @@
           "error",
```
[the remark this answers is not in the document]
WiNloSt: Fixed in https://github.com/metabase/metabase/pull/36269/commits/2e76562aa6d99bd7b315b17ba66b0d598559bd1b

frontend/src/metabase/lib/redux/hooks.unit.spec.tsx
```diff
@@ -52,7 +52,7 @@ describe("useDispatch", () => {
       const didNotFindEmailState = jest.fn();
 
       setup({
-        thunk: () => (dispatch: any, getState: () => State) => {
+        thunk: () => (_dispatch: any, getState: () => State) => {
```
WiNloSt, line 55: Just rename to match the convention even though this doesn't make ESLint error because `"args": "after-used"`

PR #36312, opened 2023-12-01
Fix broken auto-wiring when adding a new card to a dashboard

### LINKED ISSUE #36275: auto-adding filter targets on dashboards breaks dashboard
### Describe the bug

Auto-adding filters doesn't seem to be aware when a filter cannot apply to a card.

image

```javascript
{
  "error_type": "invalid-query",
 "json_query": {
    "database": 1,
    "query": {
      "source-table": 8
    },
    "parameters": [
      {
        "type": "string/=",
        "value": [
          "Gadget"
        ],
        "id": "62e39719",
        "target": [
          "dimension",
          [
            "field",
            58,
            null
          ]
        ]
      }
    ],
  "status": "failed",
  "class": "class clojure.lang.ExceptionInfo",
  "stacktrace": [
    "--> query_processor.util.add_alias_info$fn__92211$_AMPERSAND_f__92214.invoke(add_alias_info.clj:193)",
    "query_processor.util.add_alias_info$fn__92211$fn__92222.invoke(add_alias_info.clj:180)",
    "query_processor.util.add_alias_info$eval92370$fn__92371.invoke(add_alias_info.clj:378)",
    "query_processor.util.add_alias_info$add_alias_info_STAR_$replace_92412__92413$fn__92420$fn__92421.invoke(add_alias_info.clj:433)",
    "query_processor.util.add_alias_info$add_alias_info_STAR_$replace_92412__92413$fn__92420.invoke(add_alias_info.clj:433)",
    "query_processor.util.add_alias_info$add_alias_info_STAR_$replace_92412__92413.invoke(add_alias_info.clj:433)",
    "mbql.util.match.impl$replace_in_collection.invokeStatic(impl.cljc:48)",
    "mbql.util.match.impl$replace_in_collection.invoke(impl.cljc:39)",
    "query_processor.util.add_alias_info$add_alias_info_STAR_$replace_92412__92413$fn__92420.invoke(add_alias_info.clj:433)",
    "query_processor.util.add_alias_info$add_alias_info_STAR_$replace_92412__92413.invoke(add_alias_info.clj:433)",
    "mbql.util.match.impl$replace_in_collection$iter__51187__51191$fn__51192.invoke(impl.cljc:45)",
    "mbql.util.match.impl$replace_in_collection.invokeStatic(impl.cljc:44)",
    "mbql.util.match.impl$replace_in_collection.invoke(impl.cljc:39)",
...

```

### Expected behavior

_No response_

### Logs

1. create a question on sample db on table Products
2. add to a dashboard
3. add a filter on products.category
4. save dashboard
5. create a question on sample db on table reviews
6. add reviews to dashboard
7. dashboard is now broken

It doesn't seem to be aware that it cannot hook this filter up

image

image

With no filter applied dashboard works correctly:
image

If you choose a filter value the query which had the filter auto-added breaks

image

```JSON
stuff
```

### Severity

seems quite bad

### Additional context

_No response_

oisincoveney:
Closes #36275

### Description

The auto-wiring logic when adding a new card was attempting to wire the new card with the parameter target of the dashcard it matched with, rather than the new card's target. This means that, for example, if the new card has a foreign key with a target like `[field, 1, {source-table: 2}]`, it would break if the matching dashcard had `[field, 1, null]`. Technically, the fields match, but the parameter wouldn't be applied correctly without that extra information.

### How to verify

Follow the repro in #36275 

### Checklist

The test attempts to replicate the repro in #36275 by using the Products, Orders, and Reviews table to see if foreign key relationships break when adding new cards in two different ways. The tests are:
* Adding an Orders and Reviews question to the dashboard through the `add-card-sidebar`, and checking if the questions are autowired to the Products question using the ID filter. We then filter using the ID and check if the data each question is filtered accordingly, checking if the Product ID is present in each of the question's rows. 
* Adding the Orders and Reviews question to the dashboard through the query builder's `Add question to dashboard` dropdown so we can make sure that auto-wiring works immediately on load. This test was causing some issues as some of the `dispatch` functions for loading metadata weren't awaited, causing the auto-wiring to break since the fields for the new card weren't loaded. We then check if the cards have the filter, and then filter based on the ID again in the same way as above.


dpsutton:
absolutely must have some tests before this merges

frontend/src/metabase/dashboard/actions/cards.js
```diff
@@ -66,10 +66,12 @@ export const addCardToDashboard =
...
     };
-    dispatch(createAction(ADD_CARD_TO_DASH)(dashcard));
-    dispatch(fetchCardData(card, dashcard, { reload: true, clearCache: true }));
+    await dispatch(createAction(ADD_CARD_TO_DASH)(dashcard));
+    await dispatch(
+      fetchCardData(card, dashcard, { reload: true, clearCache: true }),
+    );
 
-    dispatch(loadMetadataForDashboard([dashcard]));
+    await dispatch(loadMetadataForDashboard([dashcard]));
```
oisincoveney, line 74: We need to wait for all of the metadata to load, otherwise the card won't have the data that we want for auto-wiring.

ranquild (approved):

PR #36314, opened 2023-12-01
Update visitDashboard e2e helper to be tabs-aware

iethree:
### Description

Currently the `visitDashboard` helper will cause a test failure if you visit any tabbed dashboard with dashcards on tabs other than the first tab.  This updates the helper to only expect dashcard loads from the first tab.

Previously, we used a passed parameter to accomplish this, but it's really hard to do correctly because the ids of dashcard tabs are globally unique, and may change across test runs due to changing state in the db (especially with auditv2 dashboards having tabs now).

### How to verify

All e2e tests should pass

### DIFF
--- a/e2e/support/helpers/e2e-misc-helpers.js
+++ b/e2e/support/helpers/e2e-misc-helpers.js
@@ -163,7 +163,7 @@ export function visitDashboard(dashboard_id, { params = {} } = {}) {
     url: `/api/dashboard/${dashboard_id}`,
     // That's why we have to ignore failures
     failOnStatusCode: false,
-  }).then(({ status, body: { dashcards } }) => {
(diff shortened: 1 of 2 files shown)


EmmadUsmani (approved):
Looks good ✅

nemanjaglumac (approved):
Thanks for extracting this to a separate PR.

Just out of curiosity - is it possible to navigate to a specific tab by passing query params? I mean in general when using Metabase - not in this helper.

iethree:
> Thanks for extracting this to a separate PR.
> 
> Just out of curiosity - is it possible to navigate to a specific tab by passing query params? I mean in general when using Metabase - not in this helper.

yes, but you have to know the id, so the first tab might be `/dashboard/99-my-dash?tab=86`

PR #36328, opened 2023-12-01
BE settings for whitelabeled help link

noahmoss:
Pursuant to https://github.com/metabase/metabase/issues/35915

Fixes up the backend settings for storing the whitelabeled help link options. 
* `help-link` is a keyword setting that can be `:metabase` (default), `:hidden`, or `:custom`
* `help-link-custom-destination` is a string setting which contains the URL which should be used if `help-link` is `:custom`

URL is validated to ensure it is http, https, or mailto. If the latter, we also validate it contains a valid email address.

The value of `help-link` is also included in the anonymous usage stats which are sent twice a day.


src/metabase/public_settings.clj
```diff
@@ -462,33 +434,79 @@
...
+                                   :valid-options help-link-options})))
+                (setting/set-value-of-type! :keyword :help-link value)))
+
+(defn- validate-help-url
+  "Checks that the provided URL is either a valid HTTP/HTTPS URL or a `mailto:` link. Returns `nil` if the input is valid;
+  throws an exception if it is not."
+  [url]
+  (if-let [matches (re-matches #"^mailto:(.*)" url)]
+    (when-not (u/email? (second matches))
+      (throw (ex-info (tru "Invalid email address in mailto: link")
```
npretto, line 492:
```suggestion
      (throw (ex-info (tru "Please make sure this is a valid URL")
```

src/metabase/public_settings.clj
```diff
@@ -462,33 +434,79 @@
...
+                      {:url url})))
+    (when-not (u/url? url)
+      (throw (ex-info (tru "Invalid URL")
```
npretto, line 495:
```suggestion
      (throw (ex-info (tru "Please make sure this is a valid URL")
```

src/metabase/public_settings.clj
```diff
@@ -462,33 +434,79 @@
...
+    "or be hidden if it is not set."))
+  :default    :default
+  :type       :keyword
+  :audit      :getter
+  :visibility :public
+  :feature    :whitelabeling
+  :default    :metabase
+  :setter     (fn [value]
+                (when-not (help-link-options value)
+                  (throw (ex-info (tru "Invalid help link option")
```
npretto, line 481: I tried to revert `:feature :whitelabeling` lines to test the other stuff and I got this error when trying to change the `help-link` setting from the ui:
```
[backend] 2023-12-06 13:02:09,658 ERROR middleware.log :: PUT /api/setting/help-link 500 6.2 ms (0 DB calls) 
[backend] {:via
[backend]  [{:type clojure.lang.ExceptionInfo,
[backend]    :message "Invalid help link option",
[backend]    :data {:value "custom", :valid-options #{:custom :hidden :metabase}},
[backend]    :at [metabase.public_settings$eval63935$fn__63938 invoke "public_settings.clj" 481]}],
```

npretto (commented):
Left a few inline comments

src/metabase/public_settings.clj
npretto: BTW: this change will make one of the e2e fail, I'll update the e2e to expect the correct message

johnswanson (approved):

Issue #36274, opened 2023-12-01
Instrument dashboard endpoint performance


perivamsi:
When a dashboard loads, the following directly related API calls are made (These examples use `localhost` as the host, but this can be replaced with the actual host):
* `http://localhost:3000/api/dashboard/:id`, which returns a _hydrated_ dashboard containing the dashboard, dashcards, and all other data needed to display the board.
* `http://localhost:3000/api/collection/:id` is sometimes called
* `http://localhost:3000/api/dashboard/:dashboard-id/dashcard/:dashcard-id/card/:card-id/query` is called for each dashcard in the dashboard

A few other endpoints are called (e.g. user, session, database, schema), but they don't stand out as locations that need instrumentation based on the network tab when looking at dashboard loads.

<img width="1122" alt="image" src="https://github.com/metabase/metabase/assets/8381441/89d602c9-5786-45ae-8174-c847d7b5d349">

Spans already exist for the following endpoints/functions:
- By default, we get instrumentation on all HTTP actions
- `get-dashboard` along with the subcomponents of `api/defendpoint GET "/:id"` are well instrumented
- `run-query-for-dashcard-async` has a span. This covers the body of `api/defendpoint POST "/:dashboard-id/dashcard/:dashcard-id/card/:card-id/query"`

With the fix in #35895 for slow dashboard loads, it seems like our next bottleneck is individual card loading performance. Not only can these be slow, but Chrome only allows 6 concurrent connections per domain.

At this point, we have a few potential next steps for performance improvements:
- Get `spanmetrics` working so we can view duration stats in grafana
- Profile the query code for hot spots and iterate on discoveries

markbastian:
On stats we actually have a [Top 20 P90 Backend performance report](https://grafana.monitoring.metabase.com/d/wOByfRSIz/top-20-backend-api-avg-p90-performance?orgId=1&var-p90=All&var-targets=All&var-service=metabase&var-method=All).

markbastian:
Closing this per discussion with @perivamsi since instrumentation is in place and we know where to look.

PR #36330, opened 2023-12-01
Cleaner hydration instrumentation

markbastian:
The goal of this PR is to add instrumentation to `t2/hydrate` without manually adding pervasive spans everywhere in the code. This is part of the [Instrument dashboard endpoint performance #36274](https://github.com/metabase/metabase/issues/36274) effort.

This PR consists of two ideas to be considered:

## Instrument `hydrate-with-strategy`

This is a small change that adds tracing in a more global fashion to `hydrate-with-strategy`. Unfortunately, this doesn't really capture the top-level `hydrate` calls as we were hoping for. It does provide some additional insights, but not the kind of bulk spans that tell you where you're spending most of your time.

This change is done in `metabase.db` as:

```clojure
(methodical/defmethod t2.hydrate/hydrate-with-strategy :around :default
  "Add OpenTelemetry spans around calls to hydrate-with-strategy."
  [model strategy k instances]
  (span/with-span!
    {:name       "hydrate-with-strategy"
     :attributes {:hydrate/key k}}
    (next-method model strategy k instances)))
```
## Wrap `t2/hydrate` with a macro that auto-instruments it

The idea is to add a macro (tentatively in `metabase.api.telemetry`) around `hydrate` and then just use the aliased `te/hydrate` exactly as you would `t2/hydrate`. A few nice things about this approach:

- Easy to implement -- Just change the import
- Reflects the right calling location in instrumentation
- We can modify the macro if desired to dynamically add a different span name. Currently, it's just "hydrate". We could do "hydrate-${entity-type}", for example.
- It is bespoke, but trivial to add or remove. Find and replace the require alias in api nses we want to add spans to or find and replace to remove them.

This is the proposed macro:

```clojure
(defmacro hydrate
  "A telemetry-wrapped version of hydrate. Simply replace t2/hydrate with te/hydrate."
  [entity & args]
  `(let [entity# ~entity
         model#  (mi/model entity#)]
     (span/with-span!
       {:name       "hydrate"
        :attributes {:model model#
                     (keyword (name model#) "id")
                     (:id entity#)}}
       (t2/hydrate entity# ~@args))))
```

[Here's a Loom](https://www.loom.com/share/84dc0e79a6e24f66ad63dffb316f2bb9?sid=baede034-5f1c-48bc-bef9-7aeb18e835d8) describing the impact of both designs.

### DIFF
--- a/src/metabase/api/card.clj
+++ b/src/metabase/api/card.clj
@@ -13,6 +13,7 @@
    [metabase.api.common.validation :as validation]
    [metabase.api.dataset :as api.dataset]
    [metabase.api.field :as api.field]
+   [metabase.api.telemetry :as te]
    [metabase.driver :as driver]
    [metabase.driver.sync :as driver.s]
    [metabase.driver.util :as driver.u]
@@ -53,7 +54,6 @@
    [metabase.util.log :as log]
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
-   [steffan-westcott.clj-otel.api.trace.span :as span]
    [toucan2.core :as t2])
   (:import
    (java.io File)))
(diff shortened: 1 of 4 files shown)


markbastian:
We're going to look at different avenues for this. Potentially using dynamic rebinding for hydration.

PR #36331, opened 2023-12-02
fix flaky audit tests

### LINKED ISSUE #36345: Fix and uncomment flaky tests in `metabase-enterprise.audit-app.permissions-test`
Some tests in `metabase-enterprise.audit-app.permissions-test` introduced by  [disable can-write for cards and dashboards in default audit collection](https://github.com/metabase/metabase/pull/36237) were flaking frequently so I commented them out https://github.com/metabase/metabase/pull/36341 to unblock master.

We should fix the flakiness and uncomment them.

Here are some errors:
https://github.com/metabase/metabase/actions/runs/7062703250/job/19230026705#step:5:2668
https://github.com/metabase/metabase/actions/runs/7062703250/job/19230026705#step:5:2750

And here are some failures:
https://github.com/metabase/metabase/actions/runs/7062703250/job/19230026705#step:5:2742
https://github.com/metabase/metabase/actions/runs/7062703250/job/19230026705#step:5:2746

dpsutton:
resolves #36345

Somehow (and _only_ in cloverage running) we seem to end up with the audit collection existing, but having no cards or dashboards in it. Then all of the tests are failing which check permissions on cards or dashboards.

```
ERROR in (cannot-edit-audit-content-dashboards-over-api) (metabase/util.cljc:114)
expected: (= "You don't have permissions to do that." (mt/user-http-request :rasta :put 403 (str "dashboard/" (u/the-id dashboard)) {:name "My new title"}))
  actual: java.lang.Exception: Not something with an ID: nil
 at metabase.util$error.invokeStatic (metabase/util.cljc:114)
```

This is failing

```clojure
(deftest cannot-edit-audit-content-dashboards-over-api
  (install-audit-db-if-needed!)
  (let [dashboard (t2/select-one :model/Dashboard :collection_id (:id (perms/default-audit-collection)))]
    (is (= "You don't have permissions to do that."
                                                                   ;; this dashboard is nil
                                                                   ;; and u/the-id errors
           (mt/user-http-request :rasta :put 403 (str "dashboard/" (u/the-id dashboard)) {:name "My new title"})))))
```

So let's ensure that all of the cards, dashboards, and collection exist, otherwise install them. Previously it was just checking that the collection existed, and for some reason it was empty


escherize (approved):
Kinda weird how the audit-db would be there but not the collection, maybe something is deleting the audit-db without deleting the collection. Anyway this is a better check to see if audit is installed properly.

PR #36337, opened 2023-12-04
[webpack] Do not restore yarn cache twice

uladzimirdev:
closes #36338

### Description

Do not restore yarn cache twice, frontend cache is already restored at https://github.com/metabase/metabase/blob/master/.github/actions/prepare-frontend/action.yml#L22-L27


nemanjaglumac (approved):
Great catch!
Thank you

PR #36339, opened 2023-12-04
Fix failed E2E from 61f2b28dd62938ae4f66434b1678e227bb0c8d0d

WiNloSt:
### Description

There is a test on `master` that [fails almost 100%](https://www.deploysentinel.com/ci/runs/656d57cdfcacf3cead661a6b).
image

I run `git bisect` and found 61f2b28dd62938ae4f66434b1678e227bb0c8d0d is the first commit that start to fail.

This PR fix the problem when the entity copy modal require `GET /api/collection` to be loaded bofore it could initialize the form. Otherwise, even if `initialValues` changes, the form won't rerender again.

Screenshot 2023-12-04 at 3 26 56 PM

### How to verify

Run tests in `e2e/test/scenarios/collections/instance-analytics.cy.spec.js` and they should all pass.

### Checklist

- [ ] ~Tests have been added/updated to cover changes in this PR~ No tests added, but the existing failing tests should pass now

### DIFF
--- a/frontend/src/metabase/entities/containers/EntityForm.jsx
+++ b/frontend/src/metabase/entities/containers/EntityForm.jsx
@@ -34,15 +34,18 @@ const EForm = ({
       ? entityObject.getPlainObject()
       : entityObject;
 
+  let isCustomCollectionLoaded = false;
   if (isInstanceAnalyticsCollection(entityObject?.collection)) {
     const customCollection = getInstanceAnalyticsCustomCollection(collections);
     if (customCollection) {
+      isCustomCollectionLoaded = true;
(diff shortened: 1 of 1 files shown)


nemanjaglumac (approved):
Approving to unblock `master`.
Stress-tested this fix and it's passing 20/20 ✅ 

If @ranquild or anyone from @metabase/frontend-developers has a different solution in mind, please submit it in a separate PR.

Thanks

iethree:
I had another fix here: https://github.com/metabase/metabase/pull/36329
Gonna spend some time seeing if I can come up with something more efficient

PR #36340, opened 2023-12-04
[Storybook] Build cljs with NODE_ENV=development to override default settings

uladzimirdev:
### Description

Fix storybook build. After the [latest changes in chromatic](https://github.com/chromaui/chromatic-cli/pull/865), NODE_ENV value is "production" which is not something we support at CLJS 

### How to verify

chromatic shouldn't fail for this PR https://github.com/metabase/metabase/actions/runs/7084791101/job/19279903320


nemanjaglumac (approved):
Amazing!
Thank you for finding out the root cause (upstream changes) and reacting so quickly.

PR #36341, opened 2023-12-04
Comment flaky tests

calherries:
Tests introduced in https://github.com/metabase/metabase/pull/36237 are flaking on master. This PR comments them out so they don't slow down development on other branches

[Slack context](https://metaboat.slack.com/archives/C5XHN8GLW/p1701678952633799)


sansolovyov (approved):

calherries:
This probably won't be needed because this should fix it. https://github.com/metabase/metabase/pull/36333

PR #36349, opened 2023-12-04
add Snyk workflow to run on master/release commits

### LINKED ISSUE #35677: Add Snyk to the CI pipeline
**Context**
Snyk will provide us with some additional information about our dependencies

sansolovyov:
This only runs on commits in master and release branches and then uploads results into Github's Code Scanning. Running it on pull requests is interesting to see the changes PR introduces, but makes sense only with using `snyk-delta`, which requires API access, which requires paid access to Snyk, so maybe that's something for later.

Needs SNYK_TOKEN to be set on repository. Resolves #35677

Edit:
For easier reference https://docs.snyk.io/integrate-with-snyk/snyk-ci-cd-integrations/github-actions-for-snyk-setup-and-checking-for-vulnerabilities#github-code-scanning-support

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -2,13 +2,21 @@ name: "Snyk"
 
 on:
   push:
+    branches:
+      - 'master'
+      - 'release-**'
+    paths:
+      - '**/deps.edn'
+      - '**/package.json'
+      - '.github/workflows/snyk.yml'
+      - '.github/scripts/write-poms.xml'
   schedule:
     - cron: '0 5 * * *'
 
 
 jobs:
   monitor:
-    name: Send information to Snyk
+    name: Generate Snyk report
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -2,17 +2,8 @@ name: "Snyk"
 
 on:
   push:
-    branches:
-      - 'master'
-      - 'release-**'
-    paths:
-      - '**/deps.edn'
-      - '**/package.json'
-      - '.github/workflows/snyk.yml'
-      - '.github/scripts/write-poms.xml'
   schedule:
     - cron: '0 5 * * *'
-  workflow_dispatch:
 
 
 jobs:
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - 'master'
       - 'release-**'
+    paths:
+      - '**/deps.edn'
+      - '**/package.json'
+
 
 jobs:
   pom:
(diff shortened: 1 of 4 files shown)


.github/workflows/snyk.yml
```diff
@@ -0,0 +1,31 @@
...
+    name: Generate pom.xml for Snyk
+    runs-on: ubuntu-22.04
+    steps:
+
+  monitor:
+    name: Send information to Snyk
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: snyk/actions/setup@master
```
uladzimirdev, line 20: can we have some specific version or commit instead of master?

uladzimirdev (approved):
LGTM from code perspective. @piranha did you run this job already to check it works?

nemanjaglumac (changes requested):
Adding paths to the `file-paths.yaml` without ever referencing that in a workflow doesn't have any effect. Please use the path filtering in the workflow itself.

This is not going to be a required workflow so we don't have to use a workaround (file-paths file).

.github/workflows/snyk.yml
piranha: I've debated should I do one way or the other, since most of actions use release branches and snyk does not have it... and chose randomly, haha.

piranha:
> LGTM from code perspective. @piranha did you run this job already to check it works?

Well, I run locally except for uploading sarif file, but this part is copied from `uberjar.yml`, so should be ok I guess? 😁 One thing to not is that I don't have access to repo settings so can't add SNYK_TOKEN there, and I'm not sure whom to ask. :)

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,35 @@
...
+    paths:
+      - '**/deps.edn'
+      - '**/package.json'
+
+
+jobs:
+  pom:
+    name: Generate pom.xml for Snyk
+    runs-on: ubuntu-22.04
+    steps:
```
nemanjaglumac, line 17: This is invalid job.
It's missing `steps`

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+    branches:
+      - 'master'
+      - 'release-**'
+    paths:
+      - '**/deps.edn'
+      - '**/package.json'
+  schedule:
+    - cron: '0 5 * * *'
+  manual:
```
nemanjaglumac, line 13:
```suggestion
```

nemanjaglumac (changes requested):
Still one tiny thing left to fix.
For the future reference, you can use https://rhysd.github.io/actionlint/

.github/workflows/snyk.yml
nemanjaglumac: It will pick up the latest commit from `master` on each press.

nemanjaglumac (approved):
🚀

piranha:

PR #36357, opened 2023-12-04
Ensure Metadata is Propagated for Standalone Pulse Cards

### LINKED ISSUE #36323: Ensure results metadata is applied to standalone cards in pulse HTML output
Related to https://github.com/metabase/metabase/pull/36245, this PR should ensure and test that pulse emails for individual cards have correct metadata formatting applied. The following partial test actually demonstrates the failure mode -- models with user-defined metadata do not format correctly.

```clojure
(deftest result-metadata-preservation-in-html-static-viz-no-dashboard-test
  (testing "Results metadata applied to a model or query based on a model should be used in the HTML rendering of the pulse email."
    (mt/dataset sample-dataset
      (mt/with-temp [Card {base-card-id :id} {:name          "Base question - no special metadata"
                                              :dataset_query {:database (mt/id)
                                                              :type     :query
                                                                         :expressions  {"Tax Rate" [:/
                                                                                                    [:field (mt/id :orders :tax) {:base-type :type/Float}]
                                                                                                    [:field (mt/id :orders :total) {:base-type :type/Float}]]},
                                                                         :fields       [[:field (mt/id :orders :tax) {:base-type :type/Float}]
                                                                                        [:field (mt/id :orders :total) {:base-type :type/Float}]
                                                                                        [:expression "Tax Rate"]]
                                                                         :limit        10}}}
                     Card {model-card-id :id} {:name            "Model with percent semantic type"
                                               :dataset         true
                                               :dataset_query   {:type     :query
                                                                 :database (mt/id)
                                                                 :query    {:source-table (format "card__%s" base-card-id)}}
                                               :result_metadata [{:name         "TAX"
                                                                  :display_name "Tax"
                                                                  :base_type    :type/Float}
                                                                 {:name         "TOTAL"
                                                                  :display_name "Total"
                                                                  :base_type    :type/Float}
                                                                 {:name          "Tax Rate"
                                                                  :display_name  "Tax Rate"
                                                                  :base_type     :type/Float
                                                                  :semantic_type :type/Percentage
                                                                  :field_ref     [:field "Tax Rate" {:base-type :type/Float}]}]}
                     Card {question-card-id :id} {:name          "Query based on model"
                                                  :dataset_query {:type     :query
                                                                  :database (mt/id)
                                                                  :query    {:source-table (format "card__%s" model-card-id)}}}
                     Pulse {pulse-id :id
                            :as      pulse} {:name         "Test Pulse"}
                     PulseCard _ {:pulse_id          pulse-id
                                  :card_id           base-card-id}
                     PulseCard _ {:pulse_id          pulse-id
                                  :card_id           model-card-id}
                     PulseCard _ {:pulse_id          pulse-id
                                  :card_id           question-card-id}
                                                          :pulse_id     pulse-id
                                                          :enabled      true}
        (mt/with-fake-inbox
            (mt/with-test-user nil
          ;; NOTE -- (get-in @mt/inbox ["rasta@metabase.com"]) produces the full email to rasta.
          ;; This test checks the HTML output for formatting only. A TODO is to add tests to check the attached csv and
          ;; any other result form for consistent formatting as well. As of 2023-12-01, all 2 csvs in this attachment
          ;; are not formatted as percentages. We need to add an issue and fix this.
          (let [html-body   (get-in @mt/inbox ["rasta@metabase.com" 0 :body 0 :content])
                doc         (-> html-body hik/parse hik/as-hickory)
                data-tables (hik.s/select
                              (hik.s/class "pulse-body")
                              doc)
                [base-data-row
                 model-data-row
                 question-data-row] (map
                                      (fn [data-table]
                                        (->> (hik.s/select
                                               (hik.s/child
                                                 (hik.s/tag :tbody)
                                                 (hik.s/tag :tr)
                                                 hik.s/last-child)
                                               data-table)
                                             (map (comp first :content))))
                                      data-tables)]
            ;; Card ordering is not the same as the dashboard case.
            ;; Note that only the `question-data-row` renders correctly.
            ;; The fixed output should render the `model-data-row` with percentages as well.
            [base-data-row
             model-data-row
             question-data-row]
            #_#_#_(testing "The data from the first question is just numbers."
              (is (every? (partial re-matches #"\d+\.\d+") base-data-row)))
            (testing "The data from the second question (a model) is percent formatted"
              (is (every? #(str/ends-with? % "%") model-data-row)))
            (testing "The data from the last question (based on the above model) is percent formatted"
              (is (every? #(str/ends-with? % "%") question-data-row)))))))))
```

AC:
* Add the above test (with whatever cleanup is desired) to `metabase.pulse.pulse-integration-test`
* Get the test working

Note that the above 3 cards are quite useful. We might want to capture them in a setup macro to reduce boilerplate. This could be a good PR to do it in.

markbastian:
Pulse cards based on models with user-curated metadata were not showing proper formatting for percentages. This is because the singe `execute-card` function in `metabase.pulse.util` was not adding in the card `:result_metadata` when the card was a model.

The metadata is now properly added.

Additionally, `metabase.pulse.pulse-integration-test` has been refactored for reuse. The `with-metadata-data-cards` macro provides 3 examples of cards that should be tested (a base card, a model card, and a derived question) and a function that executeds the pulse and returns the resulting data in the last column (The one we are asserting on).

Note that as we develop the test scenario in this ns there may be further refactoring opportunities, such as returning the attachments from the pulse email as well.

Fixes #36323

### DIFF
--- a/test/metabase/pulse/pulse_integration_test.clj
+++ b/test/metabase/pulse/pulse_integration_test.clj
@@ -4,19 +4,23 @@
   These tests should build content then mock out distrubution by usual channels (e.g. email) and check the results of
   the distributed content for correctness."
   (:require
-    [clojure.string :as str]
-    [clojure.test :refer :all]
-    [hickory.core :as hik]
-    [hickory.select :as hik.s]
-    [metabase.email :as email]
-    [metabase.models :refer [Card Dashboard DashboardCard Pulse PulseCard PulseChannel PulseChannelRecipient]]
-    [metabase.pulse]
-    [metabase.test :as mt]))
+   [clojure.string :as str]
+   [clojure.test :refer :all]
+   [hickory.core :as hik]
+   [hickory.select :as hik.s]
+   [metabase.email :as email]
+   [metabase.models :refer [Card Dashboard DashboardCard Pulse PulseCard PulseChannel PulseChannelRecipient]]
+   [metabase.pulse]
+   [metabase.test :as mt]))
 
-(deftest result-metadata-preservation-in-html-static-viz-test
-  (testing "Results metadata applied to a model or query based on a model should be used in the HTML rendering of the pulse email."
-    (mt/dataset sample-dataset
-      (mt/with-temp [Card {base-card-id :id} {:name          "Base question - no special metadata"
+(defmacro with-metadata-data-cards
+  "Provide a fixture that includes:
+  - A card with a custom column
+  - A model with curated metadata (override on the Tax Rate type)
+  - A question based on the above model"
+  [[base-card-id model-card-id question-card-id] & body]
+  `(mt/dataset ~'sample-dataset
+     (mt/with-temp [Card {~base-card-id :id} {:name          "Base question - no special metadata"
                                               :dataset_query {:database (mt/id)
                                                               :type     :query
                                                               :query    {:source-table (mt/id :orders)
@@ -27,11 +31,11 @@
                                                                                         [:field (mt/id :orders :total) {:base-type :type/Float}]
                                                                                         [:expression "Tax Rate"]]
                                                                          :limit        10}}}
-                     Card {model-card-id :id} {:name            "Model with percent semantic type"
+                    Card {~model-card-id :id} {:name            "Model with percent semantic type"
                                                :dataset         true
                                                :dataset_query   {:type     :query
                                                                  :database (mt/id)
-                                                                 :query    {:source-table (format "card__%s" base-card-id)}}
+                                                                 :query    {:source-table (format "card__%s" ~'base-card-id)}}
                                                :result_metadata [{:name         "TAX"
                                                                   :display_name "Tax"
                                                                   :base_type    :type/Float}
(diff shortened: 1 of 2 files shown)


test/metabase/pulse/pulse_integration_test.clj
```diff
@@ -43,11 +47,41 @@
...
-                     Card {question-card-id :id} {:name          "Query based on model"
+                    Card {~question-card-id :id} {:name          "Query based on model"
                                                   :dataset_query {:type     :query
                                                                   :database (mt/id)
-                                                                  :query    {:source-table (format "card__%s" model-card-id)}}}
-                     Dashboard {dash-id :id} {:name "just dash"}
+                                                                  :query    {:source-table (format "card__%s" ~'model-card-id)}}}]
+       ~@body)))
+
+(defn- run-pulse-and-return-last-data-columns
```
adam-james-v, line 56: This is super handy. 
I can see adding a couple helpers here with a similar pattern, but with `return last-data-row`, and perhaps some other little things like that.

In the case of checking column formatting on exports, for example, you can check a couple formatting settings in one fell swoop by pulling a row and ensuring each column has the right format... 

Anyway, I'm trying to say I like this ns with the macro and the helper fn :)

adam-james-v (approved):
Looks good! I'm excited about this `pulse-integration-test` namespace.

PR #36361, opened 2023-12-04
Uploads: Make the display name for the _mb_row_id column "_mb_row_id"

### LINKED ISSUE #36354: Make the display name for CSV upload _mb_row_id columns "_mb_row_id"
[Slack discussion](https://metaboat.slack.com/archives/C04S696LRUM/p1701691243511839?thread_ts=1701357142.724109&cid=C04S696LRUM)

Metabase automatically adds a column called `_mb_row_id` to the table in the database when a CSV is uploaded. The column's display name in Metabase is title-cased by default on the initial sync, so the column's display name is "Mb Row ID". 

image

That doesn't allow users to download the CSV and upload again when we create a feature to replace or append to the CSV table.

We should change the display name to `_mb_row_id`.

calherries:
Resolves https://github.com/metabase/metabase/issues/36354

After syncing the newly created table and its fields, we change the display name to match the name.

### DIFF
--- a/src/metabase/api/card.clj
+++ b/src/metabase/api/card.clj
@@ -888,8 +888,13 @@
           stats             (upload/load-from-csv! driver db-id schema+table-name csv-file)
           ;; Sync immediately to create the Table and its Fields; the scan is settings-dependent and can be async
           table             (sync-tables/create-or-reactivate-table! database {:name table-name :schema (not-empty schema-name)})
-          _set_is_upload    (t2/update! Table (u/the-id table) {:is_upload true})
+          _set_is_upload    (t2/update! :model/Table (u/the-id table) {:is_upload true})
           _sync             (scan-and-sync-table! database table)
+          ;; Set the display_name of the auto-generated primary key column to the same as its name, so that if users
+          ;; download results from the table as a CSV and reupload, we'll recognize it as the same column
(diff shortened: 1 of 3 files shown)


src/metabase/api/card.clj
```diff
@@ -888,8 +888,13 @@
...
           table             (sync-tables/create-or-reactivate-table! database {:name table-name :schema (not-empty schema-name)})
-          _set_is_upload    (t2/update! Table (u/the-id table) {:is_upload true})
+          _set_is_upload    (t2/update! :model/Table (u/the-id table) {:is_upload true})
           _sync             (scan-and-sync-table! database table)
+          ;; Set the display_name of the auto-generated primary key column to the same as its name, so that if users
+          ;; download results from the table as a CSV and reupload, we'll recognize it as the same column
+          auto-pk-field     (t2/select-one :model/Field :table_id (u/the-id table)
+                                           :%lower.name upload/auto-pk-column-name)
+          _  (t2/update! :model/Field (u/the-id auto-pk-field) {:display_name (:name auto-pk-field)})
           card              (card/create-card!
```
qnkhuat, line 898:
```suggestion
          _                 (t2/update! :model/Field (u/the-id auto-pk-field) {:display_name (:name auto-pk-field)})
          card              (card/create-card!
```

qnkhuat (approved):
makes sense to me

PR #36375, opened 2023-12-04
Fix scrolling in filter picker

ranquild:
How to verify:
- Add at least one segment
- Modify `FilterColumnPicker` to have 1000 segments

```
const segmentItems = includeSegments
        ? Array(1000)
            .fill(0)
            .flatMap(() =>
              Lib.availableSegments(query, stageIndex).map(segment => ({
                ...Lib.displayInfo(query, stageIndex, segment),
                segment,
              })),
            )
        : [];
```

- Open the picker
- Ensure scrollbars work as expected

Screenshot 2023-12-05 at 13 41 27


kamilmielnik (approved):
Works well.

Can we add a test for it?
Maybe it does not deserve its own test suite, but we could just add `cy.log()` in an existing test and 1-2 assertions. It's easy to reproduce this with a normal items count just by having very small viewport height (300px).

frontend/src/metabase/querying/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
```diff
@@ -110,6 +110,9 @@ export function FilterColumnPicker({
       renderItemName={renderItemName}
       renderItemDescription={omitItemDescription}
       renderItemIcon={renderItemIcon}
+      // disable scrollbars inside the list
+      style={{ overflow: "visible" }}
```
uladzimirdev, line 114: why do you need `style`? Works in the same way for me with and without this line

ranquild: Otherwise it has `overflow: auto` inside, which is not expected

uladzimirdev (approved):

PR #36379, opened 2023-12-04
Update node and java versions for v48 release

iethree:
### Description

This doesn't matter at all now, but if our supported version change in v49, it will be necessary to make sure we get the right versions for v48.


alxnddr (approved):

nemanjaglumac (approved):
👌

uladzimirdev (approved):

PR #36380, opened 2023-12-04
fix backport comment command

alxnddr:
### Description

Fixes `@metabase-bot backport xxx` command which did not work for users who is hid their Metabase organization membership. The rest of the workflow is mostly the same.
I also updated github-app-token@v2.1.0 since the new version revokes generated tokens.

### How to verify

Check the result of the new verify-membership action [here](https://github.com/scarab-engineering/test-gha/actions/runs/7093127297/job/19305918408).

### Checklist

- [n/a] Tests have been added/updated to cover changes in this PR

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,19 +11,24 @@ jobs:
     if: contains(github.event.comment.body, '@metabase-bot backport')
     runs-on: ubuntu-22.04
     steps:
-      - uses: tibdex/github-app-token@v1.6.0
+      - uses: tibdex/github-app-token@v2.1.0
         id: generate-token
         with:
           app_id: ${{ secrets.METABASE_BOT_APP_ID }}
           private_key: ${{ secrets.METABASE_BOT_APP_PRIVATE_KEY }}
-      - id: is_organization_member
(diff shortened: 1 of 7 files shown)


perivamsi (approved):

.github/workflows/backport.yml
```diff
@@ -11,19 +11,24 @@ jobs:
...
+      - name: Verify Membership
+        id: verify-membership
+        uses: ./.github/actions/verify-membership
+        with:
+          github_token: ${{ steps.generate-token.outputs.token }}
+          organization_name: metabase
+          username: ${{ github.event.comment.user.login }}
       - run: |
-          result=${{ steps.is_organization_member.outputs.result }}
+          result=${{ steps.verify-membership.outputs.is_member }}
```
iethree, line 31: are we sure `is_member` is a boolean? I've seen some cases where booleans get cooerced to strings in github actions. Can we make sure we test the negative here?

iethree: or will bash just treat it like a string either way? my bash is so bad.

alxnddr: It is returned from a shell script so it will be a string. This value is compared to `false` string below so it should be working

Issue #18981, opened 2021-11-15
Unexpected attachment behavior when attaching CSV results to emailed Dashboard Subscription


andrewler:
**Describe the bug**
There are a total of two possible files to attach in the dashboard subscription email attachments list.
I select (checkbox) one file to attach in the Dashboard Subscription configuration panel.
When the email is sent out daily, I receive both CSV as attachments instead of the one I specifically selected

**To Reproduce**
1. Dashboard Subscriptions
2. Email it daily
3. Scroll down to 'attachments'
4. place a check mark in only the first attachement checkbox
5. Wait for the daily email to be automatically sent out

**Expected behavior**
I should receive the dashboard subscription email with the CSV attachment I specified.  No more, no less.

**Information about your Metabase Installation:**
{
  "browser-info": {
    "language": "en-US",
    "platform": "Win32",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.69 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.name": "OpenJDK Runtime Environment",
    "java.runtime.version": "1.8.0_292-8u292-b10-0ubuntu1~16.04.1-b10",
    "java.vendor": "Private Build",
    "java.vendor.url": "http://java.oracle.com/",
    "java.version": "1.8.0_292",
    "java.vm.name": "OpenJDK 64-Bit Server VM",
    "java.vm.version": "25.292-b10",
    "os.name": "Linux",
    "os.version": "4.4.0-210-generic",
    "user.language": "ko",
    "user.timezone": "Asia/Seoul"
  },
  "metabase-info": {
    "databases": [
      "mysql",
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
    "application-database-details": {
      "database": {
        "name": "H2",
        "version": "1.4.197 (2018-03-18)"
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
        "version": "1.4.197 (2018-03-18)"
      }
    },
    "run-mode": "prod",
    "version": {
      "tag": "v0.41.1",
      "date": "2021-10-21",
      "branch": "release-x.41.x",
      "hash": "76aa4a5"
    },
    "settings": {
      "report-timezone": null
    }
  }
}

**Severity**
Very low severity.  Just a small annoyance.

[...]

PR #36383, opened 2023-12-05
Migrate PulseCards include_csv = true when the card has :display table

### LINKED ISSUE #36441: Make dashboard subscription attachment behavior match the user's choice in-product
## Context

Issue: https://github.com/metabase/metabase/issues/18981
Slack discussion: [Thread](https://metaboat.slack.com/archives/C064QMXEV9N/p1701708292115839)

In dashboard subscriptions, we offer the ability to "Attach results" and select which questions you want to attach data for. However, due to historical reasons around pulses, we _always_ auto-attach data when there's a table viz type question with > 10 rows. 

This doesn't respect the user's choice in the subscription sidebar and is both surprising (why is this showing up?) and limiting (why can't I only send one file?) for creators.

## Tasks
...

adam-james-v:
Closes: #36441 and https://github.com/metabase/metabase/issues/18981

This migration is so that we can eliminate the behaviour of csv attachments being included in emails even when a user doesn't have them selected in the dashboard subscription pane. This is confusing behaviour and undesirable in many cases.

The likely better behaviour is that attachments for any card are only included if explicitly asked for by a user. But, in order to change the app to behave this way, we should first migrate things so that any user who relies on this old behaviour will still see the attachments they have come to expect.

This PR is still draft and some things must happen still:

 - [x] ~~the migration is incomplete! At the moment, this will include .csv attachments even for tables with n-rows <10. This migration would cause users to see more .csv attachments than before. So, the migration should only set include_csv = true when the row count of the results is greater than 10.~~ Note: discussed in Slack, this is an acceptable tradeoff
 - [x] the migration # is NOT claimed in the slack channel yet. This should be claimed as soon as the migration approach is reviewed to work and an appropriate issue for this change is linked.


markbastian (approved):
Looks good. 🚢 🇮🇹

markbastian (approved):
Looks good. 🚢 🇮🇹

PR #36393, opened 2023-12-05
Move filter components to area folder

ranquild:
Moves relevant components to the new `querying` top-level folder. It's expected that all MLv2-based query-related components would be there eventually.


frontend/src/metabase/visualizations/click-actions/modes/TimeseriesMode.ts
```diff
@@ -1,4 +1,4 @@
-import { TimeseriesChrome } from "metabase/common/components/TimeseriesChrome";
+import { TimeseriesChrome } from "metabase/querying/filters/components/TimeseriesChrome";
```
kamilmielnik, line 1:
```suggestion
import { TimeseriesChrome } from "metabase/querying/components/TimeseriesChrome";
```

kamilmielnik (approved):

PR #36396, opened 2023-12-05
Expressions in FilterPicker without legacy query

### LINKED ISSUE #35947: FE - Expressions in AggregationPicker without legacy query
Depends on #34830

### LINKED ISSUE #35948: FE - Expressions in FilterPicker without legacy query

kamilmielnik:
Closes #35948

Depends on #36031 (which closes #35947)

### Description

Drops `legacyFilter` and `onSelectLegacy` props from `FilterPicker`.

### How to verify

Add/edit filters with custom expressions in GUI editor.

image


frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
```diff
@@ -425,34 +424,87 @@ describe("FilterPicker", () => {
...
 
     it("should update a filter with a custom expression", async () => {
-      const { onSelect, onSelectLegacy } = setup(createQueryWithCustomFilter());
+      const { query, getNextFilter, onSelect } = setup(
+        createQueryWithCustomFilter(),
+      );
 
-      await editExpressionAndSubmit("{selectall}{backspace}[[Total] > 100");
+      await editExpressionAndSubmit("{selectall}{backspace}[[Total] > 100", {
+        delay: 50,
```
kamilmielnik, line 479: I haven't figured out why but this test will fail without the `delay` - input state won't be properly updated otherwise (the `{selectall}{backspace}` won't have an effect).

frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
```diff
@@ -425,34 +424,87 @@ describe("FilterPicker", () => {
...
-        ]),
+        expect(onSelect).toHaveBeenCalledWith(expect.anything()),
       );
-      expect(onSelect).not.toHaveBeenCalled();
+
+      const filter = getNextFilter();
+      const { args, operator, options } = Lib.expressionParts(query, 0, filter);
+      const [left, right] = args;
+      const leftMetadata = checkNotNull(left) as Lib.ColumnMetadata;
+      const rightMetadata = checkNotNull(right) as Lib.ColumnMetadata;
```
kamilmielnik, line 440: We could introduce type guards to avoid these casts, WDYT?

frontend/src/metabase/common/components/AggregationPicker/AggregationPicker.tsx
```diff
@@ -220,7 +220,12 @@ export function AggregationPicker({
         withName
         startRule="aggregation"
         header={<ExpressionWidgetHeader onBack={closeExpressionEditor} />}
-        onChangeClause={handleClauseChange}
+        onChangeClause={(name, clause) => {
+          handleClauseChange(
+            name,
+            clause as Lib.AggregationClause | Lib.ExpressionClause,
```
kamilmielnik, line 226: It's a pity that `AggregationClause` does not _extend_ `ExpressionClause` - then we would not need a cast

frontend/src/metabase-lib/types.ts
```diff
@@ -130,9 +134,12 @@ export type ColumnDisplayInfo = {
...
 
-  isCalculated: boolean;
-  isFromJoin: boolean;
-  isImplicitlyJoinable: boolean;
+  semanticType: string | null;
+  effectiveType: string;
+
+  isCalculated?: boolean;
+  isFromJoin?: boolean;
+  isImplicitlyJoinable?: boolean;
```
kamilmielnik, line 142: I updated these based on what `Lib.displayInfo` returned

frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
ranquild: You're not supposed to be calling these functions in a test, it's an MBQL implementation detail. `expressionParts` shouldn't be exported from the library

ranquild: Calling something simple, like `displayInfo` is enough

frontend/src/metabase-lib/types.ts
```diff
@@ -130,9 +134,12 @@ export type ColumnDisplayInfo = {
   displayName: string;
   longDisplayName: string;
 
-  isCalculated: boolean;
-  isFromJoin: boolean;
-  isImplicitlyJoinable: boolean;
+  semanticType: string | null;
+  effectiveType: string;
```
ranquild, line 138: AFAIK we shouldn't be using these types. It's very error prone unless you understand what you doing. As I said in another comment, we shouldn't be inspecting this.

ranquild (commented):
Comments inline

frontend/src/metabase/common/components/AggregationPicker/AggregationPicker.tsx
kamilmielnik: Done: 105e9a2e274f3f529e44c5794ee120a662f286e + 73cccadfe5be077c94a1cacdd7ffd73d7a10ce80 + e470aff03565da3acc43e87b8a80a84b2d78b20 + 2676fcb6c3f04452c6a8688e9edbea80f6ef0de

But we still need casts because:
- `FilterClause` should never be used with `withExpressionName`
- `AggregationClause` will not work with `Lib.filter()`

frontend/src/metabase-lib/types.ts
kamilmielnik: 60b6021cb111a02a67040231afcae72bc2f8107

frontend/src/metabase/common/components/FilterPicker/FilterPicker.unit.spec.tsx
kamilmielnik: a25f2e006cfbcb6e1e9d0ce336d399a0d7e27811

frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
```diff
@@ -77,26 +74,25 @@ export function FilterPicker({
...
-        const baseFilter =
-          legacyFilter || new LegacyFilter([], null, legacyQuery);
-        const nextFilter = baseFilter.set(expression);
-        onSelectLegacy(nextFilter);
-        onClose?.();
-      }
+  const handleClauseChange = useCallback(
+    (_name: string, clause: Lib.CustomExpressionClause) => {
+      onSelect(
+        clause as Exclude<Lib.CustomExpressionClause, Lib.AggregationClause>,
```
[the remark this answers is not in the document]
ranquild: +1 types aren't correct

ranquild (commented):
Much better but types are still not correct. There should be no casts

frontend/src/metabase/common/components/AggregationPicker/AggregationPicker.unit.spec.tsx
```diff
@@ -183,7 +183,7 @@ function setup({
   );
 
   function getRecentClause() {
-    expect(onSelect).toHaveBeenCalled();
+    expect(onSelect).toHaveBeenCalledWith(expect.anything());
```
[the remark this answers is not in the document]
kamilmielnik: Checks presence of the argument

frontend/src/metabase/common/components/FilterPicker/FilterPicker.tsx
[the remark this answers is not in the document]
kamilmielnik: I managed to avoid the casting by making `ExpressionWidget` generic: b9a82b65a37c9327dbbfff0937da5b718e7d3f65

I don't think it's possible to avoid the casts [otherwise](https://github.com/metabase/metabase/pull/36396#discussion_r1415496134).

ranquild (approved):

uladzimirdev (approved):

PR #36402, opened 2023-12-05
Add Metabase version argument to Dockerfile

dragonsahead:
When you want to build with a specific version, you need to pass it on the build args

thanks @iethree

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 FROM node:18-bullseye as builder
 
 ARG MB_EDITION=oss
-ARG MB_VERSION
+ARG VERSION
 
 WORKDIR /home/node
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,6 +5,7 @@
 FROM node:18-bullseye as builder
 
 ARG MB_EDITION=oss
+ARG MB_VERSION
```
nemanjaglumac, line 8: Can we please use a custom name to avoid the confusion with [the official `MB_` prefixed variables](https://www.metabase.com/docs/latest/configuring-metabase/environment-variables)?

paoliniluis: done

iethree (approved):

PR #36404, opened 2023-12-05
Move upload-csv! to metabase.upload

calherries:
This PR is a refactoring of the interface to the upload code. It moves most of the logic from `metabase.api.card/upload-csv!` to `metabase.upload/upload-csv!`. Reading settings values is kept in the api namespace to minimise the amount of global state `metabase.upload/upload-csv!` depends on.

The purpose of this refactoring is to allow adding new interfaces for the uploads model such as appending or replacing data, which can reuse some of the implementation of the `upload-csv!` function.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -7,10 +7,18 @@
    [flatland.ordered.map :as ordered-map]
    [flatland.ordered.set :as ordered-set]
    [java-time.api :as t]
+   [metabase.analytics.snowplow :as snowplow]
+   [metabase.api.common :as api]
    [metabase.driver :as driver]
+   [metabase.driver.util :as driver.u]
    [metabase.mbql.util :as mbql.u]
    [metabase.models :refer [Database]]
+   [metabase.models.card :as card]
+   [metabase.models.humanization :as humanization]
    [metabase.public-settings :as public-settings]
+   [metabase.sync :as sync]
+   [metabase.sync.sync-metadata.fields :as sync-fields]
+   [metabase.sync.sync-metadata.tables :as sync-tables]
    [metabase.upload.parsing :as upload-parsing]
    [metabase.util :as u]
    [metabase.util.i18n :refer [tru]]
@@ -265,11 +273,7 @@
 ;;;; |  Parsing values  |
 ;;;; +------------------+
 
-;;;; +------------------+
-;;;; | Public Functions |
-;;;; +------------------+
-
-(defn unique-table-name
+(defn- unique-table-name
   "Append the current datetime to the given name to create a unique table name. The resulting name will be short enough for the given driver (truncating the supplised `table-name` if necessary)."
   [driver table-name]
   (let [time-format                 "_yyyyMMddHHmmss"
@@ -295,7 +299,7 @@
   [driver col->upload-type]
   (update-vals col->upload-type (partial driver/upload-type->database-type driver)))
 
-(defn detect-schema
+(defn- detect-schema
   "Consumes a CSV file that *must* have headers as the first row.
 
   Returns a map with three keys:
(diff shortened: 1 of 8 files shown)


src/metabase/upload.clj
```diff
@@ -340,3 +348,130 @@
...
+  needed for QP permissions checks).
+  Returns the newly created model. May throw validation, permimissions, or DB errors.
+
+  Args:
+  - `collection-id`: the ID of the collection to create the model in.
+  - `filename`: the name of the file being uploaded.
+  - `file`: the file being uploaded.
+  - `database`: the database to upload to.
+  - `schema-name`: the name of the schema to create the table in (optional).
+  - `table-prefix`: the prefix to use for the table name (optional)."
```
qnkhuat, line 421: could you make this take a `user` argument too? with that we can make this function free of dynamic vars

src/metabase/upload.clj
```diff
@@ -340,3 +348,130 @@
...
+
+;;; +----------------------------------------------------------------------------------------------------------------+
+;;; |                                                 public interface                                               |
+;;; +----------------------------------------------------------------------------------------------------------------+
+
+(defn- scan-and-sync-table!
+  [database table]
+  (sync-fields/sync-fields-for-table! database table)
+  (future
+    (sync/sync-table! table)))
```
qnkhuat, line 360: I find this sync code a bit strange:
1. `sync-table!` will trigger `sync-fields-for-table!` too, so why do we need to do it separately?
2. why `sync-table!` is in a future but `sync-fields-for-table!` is not?

src/metabase/upload.clj
```diff
@@ -340,3 +348,130 @@
...
+                                stats))
+        card)
+      (catch Throwable e
+        (let [fail-stats (with-open [reader (bom/bom-reader file)]
+                           (let [rows (csv/read-csv reader)]
+                             {:size-mb     (/ (.length file) 1048576.0)
+                              :num-columns (count (first rows))
+                              :num-rows    (count (rest rows))}))]
+          (snowplow/track-event! ::snowplow/csv-upload-failed api/*current-user-id* fail-stats))
+
```
qnkhuat, line 476:
```suggestion
```

qnkhuat (commented):
I love the idea, this is exactly what I thought when I review one of your recent csv PR.

had some comments nevertheless

src/metabase/upload.clj
calherries: Unfortunately the user dynamic vars need to be set because they're used for permissions checks in the QP when we execute the queries to insert the data. I find this a bit weird and unexpected, but I don't want to make any more changes in this PR.

src/metabase/upload.clj
calherries: Indeed it's not perfect, and we're doing redundant work. But when we wrote this code I assume we didn't want to spend time expanding the sync interface for this use case. It feels a little risky to call individual functions inside the sync namespaces that are not tested individually and are implementation details of more "public" sync functions such as `sync/sync-table!`.

As for 2, we need the `sync-fields-for-table!` to run synchronously, but we don't want to wait for the table to be analyzed before continuing with the rest of the steps in `upload-csv!`.

src/metabase/upload.clj
calherries: I've left a brief explanation in the docstring https://github.com/metabase/metabase/blob/37a425c52587ca0187562ae7201fd898985c6850/src/metabase/upload.clj#L413-L414

src/metabase/upload.clj
```diff
@@ -340,3 +348,130 @@
...
+  - detects the schema of the CSV file
+  - inserts the data into a new table with a unique name, along with an extra auto-generated primary key column
+  - syncs and scans the table
+  - creates a model which wraps the table
+
+  Requires that current-user dynamic vars in [[metabase.api.common]] are bound as if by API middleware (this is
+  needed for QP permissions checks).
+  Returns the newly created model. May throw validation, permimissions, or DB errors.
+
+  Args:
```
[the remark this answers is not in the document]
calherries: Added. 37a425c52587ca0187562ae7201fd898985c6850
This also caught that `database` should have been `db-id`.

qnkhuat (approved):
Cool!

PR #36405, opened 2023-12-05
Convert expression parsing related files to typescript

uladzimirdev:
closes https://github.com/metabase/metabase/issues/36406

- Accept object in diagnose
- Fix merge conflicts
- Add types to files related to parseDimension

### Note
typecheck failure is unrelated


frontend/src/metabase-lib/types.ts
```diff
@@ -171,14 +172,6 @@ export type MetricDisplayInfo = {
   selected?: boolean;
 };
 
-export type SegmentDisplayInfo = {
```
uladzimirdev, line 174: merge conflict

frontend/src/metabase-lib/expressions/index.js
```diff
@@ -58,6 +58,12 @@ function isReservedWord(word) {
 
 // METRICS
 
+/**
+ *
+ * @param {string} metricName
+ * @param {object} param1
```
kamilmielnik: `options` sounds better than `param1`

frontend/src/metabase-lib/expressions/diagnostics.ts
```diff
@@ -151,8 +163,17 @@ function prattCompiler(source, startRule, legacyQuery, name) {
...
-    return err;
+    return err as ErrorWithMessage;
   }
 
   return null;
 }
+
+function isBooleanExpression(
+  expr: unknown,
+): expr is [string, ...Expr[]] & { node: Node } {
```
kamilmielnik, line 174: I tried `1 + 1` expression and there is no `node` attribute, but this type guard says there is.

frontend/src/metabase-lib/expressions/types.ts
```diff
@@ -30,3 +30,5 @@ export type MBQLClauseFunctionConfig = {
   requiresFeature?: string;
 };
 export type MBQLClauseMap = Record<string, MBQLClauseFunctionConfig>;
+
+export type ErrorWithMessage = { message: string; pos?: number; len?: number };
```
kamilmielnik, line 34: `CompileError` says `pos` and `len` can be `null`.

Are these attributes ever used?

We can avoid at least 1 cast with this:

```suggestion
export type ErrorWithMessage = Pick<Error, 'message'>;
```

frontend/src/metabase-lib/expressions/diagnostics.ts
```diff
@@ -151,8 +163,17 @@ function prattCompiler(source, startRule, legacyQuery, name) {
     }
   } catch (err) {
     console.warn("compile error", err);
-    return err;
+    return err as ErrorWithMessage;
```
kamilmielnik, line 166: We can avoid these casts by introducing type guards + having a default error.

Ccomponents use this default error: 
```
{ message: t`Invalid expression` }
```

kamilmielnik (commented):
There are 2 TS compile errors :thinking:

uladzimirdev:
> There are 2 TS compile errors 🤔

thanks for checking! 
image

frontend/src/metabase-lib/expressions/diagnostics.ts
uladzimirdev: IDK, from what I know, error in catch blocks is untyped because of many reasons, so you have to manually cast it

frontend/src/metabase-lib/expressions/types.ts
uladzimirdev: hmm what about `pos` and `len`? actually I copied this type

frontend/src/metabase-lib/expressions/diagnostics.ts
kamilmielnik: > error in catch blocks is untyped because of many reasons

That's precisely why they should not be casted.

```suggestion
    if (isErrorWithMessage(err)) {
      return err;
    }

    return { message: t`Invalid expression` };
```

frontend/src/metabase-lib/expressions/types.ts
kamilmielnik: Right, they're still used in `ExpressionEditorTextfield`.

But they can still be `null`:

```suggestion
export type ErrorWithMessage = { message: string; pos?: number | null; len?: number | null };
```

This allows to remove cast from line 115 in `diagnostics.ts`.

frontend/src/metabase-lib/expressions/diagnostics.ts
uladzimirdev: made node optional

frontend/src/metabase-lib/expressions/types.ts
uladzimirdev: added `number | null`

frontend/src/metabase-lib/expressions/diagnostics.ts
uladzimirdev: added type guard, thanks

frontend/src/metabase-lib/expressions/diagnostics.ts
```diff
@@ -151,8 +167,26 @@ function prattCompiler(source, startRule, legacyQuery, name) {
...
+  expr: unknown,
+): expr is [string, ...Expr[]] & { node?: Node } {
+  return (
+    Array.isArray(expr) &&
+    (LOGICAL_OPS.includes(expr[0]) || COMPARISON_OPS.includes(expr[0]))
+  );
+}
+
+function isErrorWithMessage(err: unknown): err is ErrorWithMessage {
+  return err instanceof Error && typeof err.message === "string";
```
kamilmielnik, line 191: We should either make the guard less strict or make `ErrorWithMessage` extend `Error`.
But probably less strict, because `prattCompiler` now returns something that is not an instance of Error, i.e.
```
{ message: t`Invalid expression` }
``` 

```suggestion
  return typeof err === 'object' && !err && typeof err.message === "string";
```

frontend/src/metabase-lib/expressions/index.ts
```diff
@@ -52,44 +59,47 @@ export function formatIdentifier(name, { quotes = EDITOR_QUOTES } = {}) {
...
+  metricName: string,
+  { legacyQuery }: { legacyQuery: StructuredQuery },
+) {
+  return checkNotNull(legacyQuery.table()).metrics?.find(
+    metric => metric.name.toLowerCase() === metricName.toLowerCase(),
+  );
 }
 
-export function formatMetricName(metric, options) {
+export function formatMetricName(metric: Metric, options: object) {
```
kamilmielnik, line 77: Can you please add typing for `options` here and in `formatSegmentName` + `formatDimensionName`?

frontend/src/metabase/query_builder/components/filters/FilterPopover/FilterPopover.tsx
```diff
@@ -165,10 +165,8 @@ export function FilterPopover({
   };
 
   if (editingFilter) {
-    const filterMBQL = filter?.raw();
-    const expression = isExpression(filterMBQL)
-      ? (filterMBQL as Expression)
-      : undefined;
+    const filterMBQL = filter?.raw() as Expression;
```
kamilmielnik, line 168:
```suggestion
    const filterMBQL = checkNotNull(filter).raw();
```

frontend/src/metabase-lib/expressions/index.ts
```diff
@@ -166,15 +182,15 @@ const STRING_UNESCAPE = {
...
+export function quoteString(string: string, quote: string) {
   if (quote === DOUBLE_QUOTE || quote === SINGLE_QUOTE) {
     let str = "";
     for (let i = 0; i < string.length; ++i) {
       const ch = string[i];
       if (ch === quote && string[i - 1] !== BACKSLASH) {
         str += BACKSLASH + ch;
       } else {
-        const sub = STRING_ESCAPE[ch];
+        const sub = STRING_ESCAPE[ch as keyof typeof STRING_ESCAPE];
```
kamilmielnik, line 193: We can avoid this cast if we add explicit type for `STRING_ESCAPE`, like this:

```tsx
const STRING_ESCAPE: Record<string, string> = {
```

frontend/src/metabase-lib/expressions/index.ts
uladzimirdev: those options are not defined, I changed them to `Record<string, any>`, but I don't like it better

frontend/src/metabase-lib/expressions/diagnostics.ts
```diff
@@ -151,8 +165,30 @@ function prattCompiler(source, startRule, legacyQuery, name) {
...
+    Array.isArray(expr) &&
+    (LOGICAL_OPS.includes(expr[0]) || COMPARISON_OPS.includes(expr[0]))
+  );
+}
+
+function isErrorWithMessage(err: unknown): err is ErrorWithMessage {
+  return (
+    typeof err === "object" &&
+    err != null &&
+    typeof (err as any).message === "string"
```
kamilmielnik, line 192:
```suggestion
    "message" in err &&
    typeof err.message === "string"
```

kamilmielnik (commented):
LGTM :+1: 

~I'll leave the approval to Alex.~

kamilmielnik (approved):

Issue #16519, opened 2021-06-09
Things to do after pulses are removed


kulyk:
As we're going to remove pulses in favor of dashboard subscriptions, it would make sense to keep a list of things we should do after the pulses are gone.

**How to use (proposal)**

We can leave a TODO in the code and reference a specific item in the list below. Example:

```
// We don't keep last edit info for pulses
// TODO Remove ternary when Pulses are gone (metabase#16519-1)
```

If that makes sense, it's important to only add new items to the end of the list and never move anything, so references stay correct

**Todos**

* [ ] remove [`last-edit-info` null check](https://github.com/metabase/metabase/blob/f4ab182d4cddc07c56efe4551aa492c25a04a4ae/frontend/src/metabase/collections/components/BaseTableItem.jsx#L66) (exists for pulses only)

PR #36407, opened 2023-12-05
FE: Things to do after pulses are removed

qnkhuat:
Closes https://github.com/metabase/metabase/issues/16519

Tried to grep for related TODO but couldn't find any, if you know any please lmk


kulyk (approved):
❤️

qnkhuat:
@kulyk I've reverted the change as we don't have lasteditinfo for collections either. This PR will remove the TODO comment only.

PR #36413, opened 2023-12-05
Add :timezone key to database metadata

### LINKED ISSUE #36662: `qp.timezone/database-timezone-id` returns nil for `::db-from-store`
While debugging past UTC midnight failures [this slack thread](https://metaboat.slack.com/archives/C5XHN8GLW/p1701787186845979?thread_ts=1701772468.810829&cid=C5XHN8GLW) is about, I've found out that `qp.timezone/database-timezone-id` returns nil for `::db-from-store`. According to the mentioned thread, this behavior is not desired.

It could be fixed adding a `:timezone` to pipeline building
```
t2.pipeline/build [#_query-type     :toucan.query-type/select.*
                                         #_model          :metadata/database
                                         #_resolved-query clojure.lang.IPersistentMap]
``` 
in `metabase.lib.metadata.jvm`.

This however breaks some tests for snowflake and oracle drivers. I'll look into that while working on the issue #20988 which I'm about to solve.

lbrdnk:
Fixes #36662

The PR is now review-ready. Description was updated.

---

~~Change is still WIP~~. I do however require ~~intermediate~~ review from:
- _qpd team_ to validate my approach ~~before going further~~ and 
- _ci team_ because ~~what's left~~ what was left to implement ~~is~~ was concerned with e2e tests and also to validate my approach ~~(not the code yet, at the time of writing)~~. You can possibly skip to the _e2e_ section near the end.

~~I'm not yet able to handle e2e failures efficiently, still learning that and I think we should be on a same page with those changes prior to me going further trying to tackle those.~~

Further on why and how I came to solving this can be found [on slack](https://metaboat.slack.com/archives/C5XHN8GLW/p1701803498776619?thread_ts=1701772468.810829&cid=C5XHN8GLW).

---

### Description

This PR adds `:timezone` key to the database metadata. This is essential for `qp.timezone/database-timezone-id` to work correctly. That's then essential for `qp.timezone/now` which forms a basis for Snowflake and Redshfit caching that is in development.

---

_Problem_. This change caused lot of test failures. Those are mostly caused by the fact that `qp.timezone/results-timezone-id` now returning actual database timezone id instead of system timezone.

_Aside on [`qp.timezone/results-timezone-id`](https://github.com/metabase/metabase/blob/a64885b9e51f5948fb4f2984c34fc214b3844749/src/metabase/query_processor/timezone.clj#L77)._ This function returns 1. report timezone if available and set or 2. database timezone or 3. system timezone. Prior to this change return of [`qp.timezone/database-timezone-id`](https://github.com/metabase/metabase/blob/a64885b9e51f5948fb4f2984c34fc214b3844749/src/metabase/query_processor/timezone.clj#L52) was `nil` for `::db-from-store`. As the database timezone was in aforementioned circumstances `nil`, `results-timezone-id` resorted to returning system timezone. That is by default UTC. Modifying that would require modifying JVM profile.

_Back to the problem_. Most of the driver tests handle date/time/stamp values as UTC. With this change in place however, actual database timezone is used in various places instead and values are handled differently.

_Solution_. I believe correct solution is to adjust database timezones we are using in tests to be UTC so our UTC test expectations are met.

_Snowflake_. For a Snowflake, there are multiple changes:
1. Prior to this PR, `db-default-timezone` returned actual _default_ timezone for Snowflake. For Redshift (`:postgres` implementation), it returned timezone that is result of the actual `show timezone;` -- so timezone that is in place for the connection if we do not override it (by eg. setting the report timezone). This PR adjusts Snowflake to do the same -- return actual timezone for a connection if we don't override it. Correct value of that is essential for `qp.timezone/now` that is used as a basis for Redshift and Snowflake caching.
2. Set the default timezone for tests to UTC. Snowflake uses `America/Los_Angeles` by default.

_Notion of default timezone?_ I believe our notion of default timezone should be like _the timezone that is used for connection without us overriding that_ (by setting report timezone). Because that's what is imo relevant for our relative date/time/stamp server side calculations.

_Oracle_. Oracle's default timezone is `+00:00`. Oracle failures were resolved handling UTC `unprepare/unprepare-value [:oracle ZonedDateTime]` the same way as it is handled in `unprepare/unprepare-value [:oracle OffsetDateTime]`, the one method above in code. So this was a different problem.

~~_e2e_. There are still few e2e test failures that should be resolved. I believe those are caused by underlying mysql timezone inconsistencies. Namely that it uses whatever timezone (based on [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html) -- I'm unable to anchor that, but see the Time Zone Variables section at the beginning) system uses and takes into account TZ environment variable. Based on that I should be able to override this behavior setting TZ variable for container in [prepare-e2e-containers action](https://github.com/metabase/metabase/blob/72893a8f1bfbf421a6668d8fe1d66f26f9127d3d/.github/actions/e2e-prepare-containers/action.yml#L76). I was not effective with that, hence I'm still investigating why.~~

_e2es update_. Most of the failures were resolved by changing how h2 determines database timezone. Previously we were using UTC offset computed from now value. Hence synced timezone was `-08:00`. That caused all date/time values eg. in generated filters to use -8 offset.  Cypress runner however starts backend with jvm timezone set to `US/Pacific` (that would be the results timezone, while `database-timezone-id` was returning nil) which respects DST, so some date/time values should have -7 offset, and tests expectations were set minding that. Mysql and Postgres failures were resolved by adjusting containers TZ variables and minor test modification.


modules/drivers/oracle/src/metabase/driver/oracle.clj
```diff
@@ -627,7 +627,8 @@
 
 (defmethod unprepare/unprepare-value [:oracle ZonedDateTime]
   [_ t]
-  (format "timestamp '%s'" (t/format "yyyy-MM-dd HH:mm:ss.SSS VV" t)))
+  (format "timestamp '%s'" (-> (t/format "yyyy-MM-dd HH:mm:ss.SSS VV" t)
+                               (str/replace #"Z$" "UTC"))))
```
lbrdnk, line 631: This is inspired by `unprepare/unprepare-value [:oracle OffsetDateTime]`, one method above in code.

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -517,10 +517,10 @@
   (sql-jdbc.execute/do-with-connection-with-options
    driver database nil
    (fn [^java.sql.Connection conn]
-     (with-open [stmt (.prepareStatement conn "show parameters like 'TIMEZONE';")
+     (with-open [stmt (.prepareStatement conn "show parameters like 'TIMEZONE' in user;")
```
lbrdnk, line 520: Rationale for changes to this method is further described in the PR description, Snowflake part.

Here are the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/parameters#timezone) on timezone setting. So we have account, user and session timezone, where latter has a preference. User timezone is persistent across sessions. Session timezone is not (obvious, I know...). Hence I consider the user timezone (if not set, inherited from the account) as _the timezone that is used for connection without us overriding that (by setting report timezone)_ from the PR description.

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -517,10 +517,10 @@
...
                  rset (.executeQuery stmt)]
        (when (.next rset)
-         (.getString rset "default"))))))
```
lbrdnk, line 523: `value` gives us actual timezone that user is using (that can change in between syncs). `default` is the default for every deployment and does not have to correspond to what's in actual use.

modules/drivers/snowflake/test/metabase/test/data/snowflake.clj
```diff
@@ -144,12 +144,26 @@
         (delete-old-datasets!)
         (reset! deleted-old-datasets? true)))))
 
+(defn- set-current-user-timezone!
```
lbrdnk, line 147: Effect of this was and is, that user of the instance tests are running on has their timezone set permanently to UTC. Even after running from this branch. I was checking for some negative impact this could have after running it but to this point I've found none. (No CI drivers failures related to this I mean.)

lbrdnk (commented):
There is still work to be done. To others, please add your remarks, so I can continue or shift the direction this is going.

snoe:
This approach seems good to me, I think snowflake and oracle likely haven't been exercised as much and thus what you found with default vs user vs session makes sense.

.github/actions/e2e-prepare-containers/action.yml
```diff
@@ -66,14 +66,14 @@ runs:
 
         if ${{ inputs.postgres }}; then
           echo -e "${Y}Starting postgres container..." &&
-          docker run -d -p 5404:5432 metabase/qa-databases:postgres-sample-12 &&
+          docker run -d -p 5404:5432 -e TZ='US/Pacific' metabase/qa-databases:postgres-sample-12 &&
```
uladzimirdev, line 69: what if we pass TZ as a global env variable instead? @metabase/core-ci

lbrdnk: I've tried to find a way how to do that reading [actions metadata syntax reference](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions), but I was unable to. According to that document, `runs`-global `env` may be specified only in `docker` action and we are using `composite`. Composite then can specify `env` per `runs.steps` entry, but I believe that is not desired. Could you please elaborate on how to achieve that? Thank you.

lbrdnk (commented):
I believe code is now review ready, with all tests passing.

src/metabase/driver/h2.clj
```diff
@@ -312,15 +312,10 @@
...
 (defmethod driver/db-default-timezone :h2
-  [driver database]
-  (sql-jdbc.execute/do-with-connection-with-options
-   driver database nil
-   (fn [^java.sql.Connection conn]
-     (with-open [stmt (.prepareStatement conn "select current_timestamp();")
-                 rset (.executeQuery stmt)]
-       (when (.next rset)
-         (when-let [zoned-date-time (.getObject rset 1 java.time.ZonedDateTime)]
-           (t/zone-id zoned-date-time)))))))
```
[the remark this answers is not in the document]
snoe: this seems ok as h2 does not allow extracting of the time zone region.

```
select TIMESTAMP WITH TIME ZONE '2005-12-31 23:59:59.123456789 Europe/London';
TIMESTAMP WITH TIME ZONE '2005-12-31 16:59:59.123456789-07'
2005-12-31 16:59:59.123456789-07
```

snoe (approved):
LGTM

metamben (changes requested):
Just a minor issue, otherwise LGTM.

modules/drivers/oracle/src/metabase/driver/oracle.clj
lbrdnk: Thank you.

modules/drivers/oracle/src/metabase/driver/oracle.clj
```diff
@@ -627,7 +627,8 @@
 
 (defmethod unprepare/unprepare-value [:oracle ZonedDateTime]
   [_ t]
-  (format "timestamp '%s'" (t/format "yyyy-MM-dd HH:mm:ss.SSS VV" t)))
+  (format "timestamp '%s'" (-> (t/format "yyyy-MM-dd HH:mm:ss.SSS VV" t)
+                               (str/replace #" Z$" " UTC"))))
```
metamben, line 631: `unprepare/unprepare-value [:oracle OffsetDateTime]` has the same problem.

lbrdnk: @metamben, I thought about that case while submitting your suggestion. My reasoning was as follows. Formatting `OffsetDateTime` for _some reason_ does not convert Z to 00:00. Hence the original workaround. But we expect other offsets to be as eg. `+13:00`, so it is fine to use just `#"Z$"` expression there.

`NZ` example:
```
(-> (t/format "yyyy-MM-dd HH:mm:ss.SSS ZZZZZ"
                (java.time.OffsetDateTime/now (java.time.ZoneId/of "NZ")))
      (str/replace #"Z$" "UTC"))
  ;; => "2024-01-17 07:57:25.306 +13:00"
```
I believe the current implementation of `unprepare/unprepare-value [:oracle OffsetDateTime]` is correct. Is there an error in my reasoning? Thanks.

metamben: Your reasoning is correct, but I would still keep that symmetrical. It would still work, wouldn't be so expensive that it hurts, and we would get a clearer error on the off chance that another special case ending in `Z` gets introduced.

But it's not essential for sure, so I'm totally fine with your leaving it as is.

metamben (approved):

.github/actions/e2e-prepare-containers/action.yml
uladzimirdev: you're right. I think it's fine to pass env vars right away, I checked docs and probably it will work if we pass env at the step level + pass this env to the docker, but it's a matter of taste and I'd not change anything in your version

👍

modules/drivers/oracle/src/metabase/driver/oracle.clj
lbrdnk: I did as suggested. Methods' implementations are now symmetric.

Issue #36327, opened 2023-12-01
Ensure consistent percentage formatting across FE and pulses


markbastian:
Dashboard rendering of percent columns and the html export produced when emailing a pulse do not use the same number of significant digits. In the example below, the dashboard shows 2 digits after the decimal point and the pulse html rounds to the nearest integer.

To reproduce:
* Create a model and edit the user metadata on a column with values in the range of [0, 100] as `:type/Percentage`.
* Create a question based on that model.
* Add both items to a dashboard.
* Email the dashboard.
* View the email body.

You should see the values represented with different precision.

The following animation demonstrates the problem behavior.

![formatting](https://github.com/metabase/metabase/assets/8381441/2c1a4fae-3793-4ac9-9d93-c88dd960e3e8)

The unit test `metabase.pulse.pulse-integration-test/result-metadata-preservation-in-html-static-viz-test` is a good basis to set up a unit test for this behavior.

AC:
* A unit test that computes both the in-app rendered value and the and the exported HTML value.
* These values should be the same.
* If the in-app values are determined on the FE, perhaps manually set the expected values to what is on the FE.

markbastian:
Fixed by https://github.com/metabase/metabase/pull/36431

PR #36431, opened 2023-12-05
Updating default `number-formatter` for percents

markbastian:
The number formatter used in some static viz rendering (e.g. tables) in `metabase.pulse.render.common/number-formatter` had logic to render only integer-valued percents for values between 1 and 100 and render two significant digits otherwise.

The FE default formatter uses 2 significant digits consistently in `frontend/src/metabase/lib/formatting/numbers.tsx`:

```js
const PRECISION_NUMBER_FORMATTER = d3.format(".2f");
```

This change modifies `number-formatter` to consistenly use 2 significant digits for percentage types.

Note that this does NOT apply to fields with custom formatting. Presently, custom column formatting does not appear to be applied for HTML tables at all for BE rendering while it is applied on the FE. This is a separate issue. This PR just brings the baseline behaviors based on typing into alignment.

### DIFF
--- a/src/metabase/pulse/render/common.clj
+++ b/src/metabase/pulse/render/common.clj
@@ -107,15 +107,12 @@
     (fn [value]
       (if (number? value)
         (let [scaled-value (* value (or scale 1))
-              percent-scaled-value (* 100 scaled-value)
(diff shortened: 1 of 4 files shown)


adam-james-v (approved):
Looks good!

markbastian:
Fixes #36327

PR #36440, opened 2023-12-05
Return `:pk` drills for non-PK values in rows with multiple PKs

### LINKED ISSUE #35618: [MLv2] [BE] [Bug] `drill-thru/pk` drill should be returned for query that has multiple PKs on non-PK columns click
In current implementation, pk drill (View details) is available for non-pk columns as well when the **query has no aggregations**.
In the mbqlv2 version it is available only for pk columns itself.

Repro test https://github.com/metabase/metabase/blob/daf45827f2ef15d8826ae02ea5cb68a98de96f2a/frontend/src/metabase-lib/tests/drills-pk.unit.spec.ts#L58

### LINKED ISSUE #36426: "View details" (PK) drill thru on non-PK value for a row with multiple PKs only adds a filter for one of the PKs, nondeterminiate
If you have a row with multiple PKs and click a cell for a non-PK value and click "view details", it will only add a filter for one of the PKs (which one is not determinate). See [this Slack thread](https://metaboat.slack.com/archives/C04CYTEL9N2/p1701740553972049) for more information.

This behavior doesn't make sense; we should add one filter for each PK.

camsaul:
If you have a Table with multiple PKs and click a non-PK column value cell, it should return a `:pk` filter drill. This drill should add one filter for each non-NULL PK value.

See discussion in https://metaboat.slack.com/archives/C04CYTEL9N2/p1701740553972049

As part of this I overhauled the schema for `:pk` drills, see discussion in https://metaboat.slack.com/archives/C04CYTEL9N2/p1701803047600169

I also refactored the `object-details` drill namespace a bit and moved the code for returning `:pk`, `:zoom`, and `:fk-details` drills into their respective namespaces -- it always seemed weird to have the tests there but then have the thing they're testing live somewhere else. Code is a little easier to follow now.

Fixes #35618
Fixes #36426

### DIFF
--- a/src/metabase/lib/drill_thru/object_details.cljc
+++ b/src/metabase/lib/drill_thru/object_details.cljc
@@ -1,51 +1,11 @@
 (ns metabase.lib.drill-thru.object-details
   (:require
-   [medley.core :as m]
-   [metabase.lib.aggregation :as lib.aggregation]
-   [metabase.lib.drill-thru.common :as lib.drill-thru.common]
-   [metabase.lib.metadata.calculation :as lib.metadata.calculation]
+   [metabase.lib.drill-thru.fk-details :as lib.drill-thru.fk-details]
+   [metabase.lib.drill-thru.pk :as lib.drill-thru.pk]
+   [metabase.lib.drill-thru.zoom :as lib.drill-thru.zoom]
    [metabase.lib.schema :as lib.schema]
    [metabase.lib.schema.drill-thru :as lib.schema.drill-thru]
-   [metabase.lib.types.isa :as lib.types.isa]
    [metabase.util.malli :as mu]))
-
-(defn- object-detail-drill-for [query stage-number {:keys [column row value] :as context} many-pks?]
-  (let [base {:lib/type  :metabase.lib.drill-thru/drill-thru
-              :type      :drill-thru/pk
-              :column    column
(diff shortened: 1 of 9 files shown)


src/metabase/lib/drill_thru/object_details.cljc
```diff
@@ -55,10 +15,11 @@
...
-    (object-detail-drill-for query stage-number context
-                             (> (count (lib.metadata.calculation/primary-keys query)) 1))))
+  [query        :- ::lib.schema/query
+   stage-number :- :int
+   context      :- ::lib.schema.drill-thru/context]
+  (some (fn [f]
+          (f query stage-number context))
+        [lib.drill-thru.fk-details/fk-details-drill
+         lib.drill-thru.pk/pk-drill
+         lib.drill-thru.zoom/zoom-drill]))
```
bshepherdson, line 25: These are in a different order than the conditions were checked previously - I assume that's on purpose? A comment about the order would be a good idea here.

bshepherdson (approved):

src/metabase/lib/drill_thru/object_details.cljc
camsaul: yeah I did initially change the order because it mattered but after reworking a few things it's not important anymore

PR #36444, opened 2023-12-05
i18n() syntax errors caught while translating

joebordes:
minor syntax and semantic errors detected while translating to ES

### DIFF
--- a/src/metabase/driver/sql/query_processor.clj
+++ b/src/metabase/driver/sql/query_processor.clj
@@ -757,7 +757,7 @@
 (defmethod ->honeysql [:sql :-]
   [driver [_ & [first-arg & other-args :as args]]]
   (cond (interval? first-arg)
-        (throw (ex-info (tru "Interval as first argrument to subtraction is not allowed.")
+        (throw (ex-info (tru "An interval as the first argument to subtraction is not allowed.")
                         {:type qp.error-type/invalid-query
                          :args args}))
         (and (some interval? other-args)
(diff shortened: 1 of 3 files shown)


joebordes:
Label: no-backport

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -445,7 +445,7 @@
   (when (< (try (.. ^PreparedStatement stmt getParameterMetaData getParameterCount)
                 (catch Throwable _ (count params)))
            (count params))
-    (throw (ex-info (tru "It looks like we got more parameters than we can handle, remember that parameters cannot be used in comments or as identifiers.")
```
darksciencebase, line 448: I believe the original message is correct here; it's not about _having_ more parameters, it's about having _gotten_ more parameters.

src/metabase/driver/sql/query_processor.clj
```diff
@@ -757,7 +757,7 @@
 (defmethod ->honeysql [:sql :-]
   [driver [_ & [first-arg & other-args :as args]]]
   (cond (interval? first-arg)
-        (throw (ex-info (tru "Interval as first argrument to subtraction is not allowed.")
```
darksciencebase, line 760: @jeff-bruemmer do we by any chance refer to intervals as proper nouns in cases like this?

darksciencebase:
@nemanjaglumac doesn't look related to i18n actually, we're adjusting the error messages here (and those just happen to be translated)

src/metabase/driver/sql/query_processor.clj
jeff-bruemmer: It's a common noun.

PR #36448, opened 2023-12-06
Misc `zoom-in-geographic` drill fixes

### LINKED ISSUE #36247: [MLv2] [BE] [Bug] `zoom-in.geographic` isn't returned in most cases

camsaul:
* Geo zoom drill should look at drill context `:dimensions`
* Geo zoom drill should add `:>=` and `:<` filters rather than `:between`, because `:between` is inclusive in MBQL (as in SQL)

Fixes #36247

FE unit tests in `drills-integration` are passing after spending an hour or so making numerous fixes (see https://metaboat.slack.com/archives/C04CYTEL9N2/p1701826072570649). Actual backend fixes took maybe 10 minutes


src/metabase/lib/convert.cljc
```diff
@@ -259,13 +259,17 @@
...
 
 (defmethod ->pMBQL :aggregation
-  [[tag value opts]]
-  (lib.options/ensure-uuid [tag opts (get-or-throw! *legacy-index->pMBQL-uuid* value)]))
+  [[tag aggregation-index opts, :as clause]]
+  (lib.options/ensure-uuid
+   [tag opts (or (get *legacy-index->pMBQL-uuid* aggregation-index)
+                 (throw (ex-info (str "Error converting :aggregation reference: no aggregation at index "
+                                      aggregation-index)
+                                 {:clause clause})))]))
```
camsaul, line 272: Unrelated but I changed the error here because "Unable to find 0" was a really confusing error when there was an orphaned aggregation reference.

bshepherdson (approved):

PR #36451, opened 2023-12-06
Add dev fn to render dashboards with static-viz renderers

adam-james-v:
image

Preview a Dashboard's exports by using the endpoint `/api/pulse/preview_dashboard/:id`

Run `(dev.render-png/render-dashboard-to-html-and-open 1)` to render the dashboard with id 1 to a handy html file which will open in your default browser.

This file contains renders for each dashcard in the dashboard 3 ways:
 - as a png, like you would see in Slack or in an email body (if it's a chart)
 - as html/svg, like you'd see in the email body (for tables).. might also be helpful to inspect the svg output from the graaljs interpreter
 - 10 row table representing what the .csv attachment for the dashcard would look like

Todo:
 - [x] avoid with-redefs to pass the linter
 - [x] add a preview endpoint that can be used so that you don't need to run a Clojure repl to get the same result.
 - [x] demo screenshots and/or recording
 - [ ] make CSV preview work with the endpoint too (not just in dev ns)


dev/src/dev/render_png.clj
```diff
@@ -85,29 +90,123 @@
...
-
-  (-> [["As" "Bs" "Cs" "Ds" "Es"]
-       ["aa" "bb" "cc" "dd" "ee"]
-       ["aaa" "bbb" "ccc" "ddd" "eee"]]
-      (render.tu/make-viz-data :table {:reordered-columns   {:order [2 3 1 0 4]}
-                                       :custom-column-names {:names ["-A-" "-B-" "-C-" "-D-"]}
-                                       :hidden-columns      {:hide [0 2]}})
-      :viz-tree
-      open-hiccup-as-html))
+(def ^:private execute-dashboard #'pulse/execute-dashboard)
```
markbastian, line 93: I usually just call private functions as regular aliased functions -- `(#'pulse/execute-dashboard ...)` inline. Were you just defining this here for convenience?

dev/src/dev/render_png.clj
```diff
@@ -85,29 +90,123 @@
...
+
+(defn render-dashboard-to-html-and-open
+  "Given a dashboard ID, renders all of the dashcards to an html file and opens it."
+  [dashboard-id]
+  (let [html-str (render-dashboard-to-html dashboard-id)
+        tmp-file (File/createTempFile "card-html" ".html")]
+    (with-open [w (io/writer tmp-file)]
+      (.write w ^String html-str))
+    (.deleteOnExit tmp-file)
+    (open tmp-file)))
```
markbastian, line 209: I love all the new dev functions here.

src/metabase/api/pulse.clj
```diff
@@ -275,6 +276,18 @@
                           render/*include-buttons* true]
                   (render/render-pulse-card-for-display (metabase.pulse/defaulted-timezone card) card result))]])}))
```
markbastian, line 278: My only big question here is if we want to enable this publicly, because then it's a supported API. I'd much rather expose it only if the dev profile is active. I don't see why a customer would want this, but it's exceptionally valuable for development. WDYT?

src/metabase/pulse/preview.clj
```diff
@@ -0,0 +1,163 @@
...
+   :border-radius "8px"
+   :margin        "10px"
+   :padding       "10px"})
+
+(def ^:private table-style
+  (style/style
+   {:border          "1px solid black"
+    :border-collapse "collapse"}))
+
+(def ^:private csv-row-limit 10)
```
markbastian, line 85: Might want to make this a query param and pass it in that way.

src/metabase/pulse/preview.clj
```diff
@@ -0,0 +1,163 @@
...
+  (condp
+      #(<= 0 (.indexOf ^String %2 ^String %1))
+      (.toLowerCase (System/getProperty "os.name"))
+      "win" :win
+      "mac" :mac
+      "nix" :unix
+      "nux" :unix
+      nil))
+
+;; taken from https://github.com/aysylu/loom/blob/master/src/loom/io.clj
```
markbastian, line 32: This fn and ns also feels very dev-y. Is there a way to move this and the defendpoint into our dev folder and conditionally activate it that way?

markbastian (changes requested):
I like it all except I think maybe we want to relocate it so it only shows up in dev. Can we load that defendpoint when you run `(dev/start!)`?

src/metabase/pulse/preview.clj
adam-james-v: removed from the preview ns

markbastian (approved):

PR #36453, opened 2023-12-06
🍒 Manual backported "Fix click behavior link to dashboards/questions not using client router"

WiNloSt:
A manual backport of https://github.com/metabase/metabase/pull/35975

### DIFF
--- a/e2e/support/commands/api/dashboard.js
+++ b/e2e/support/commands/api/dashboard.js
@@ -1,17 +1,23 @@
 Cypress.Commands.add(
   "createDashboard",
-  ({
-    name = "Test Dashboard",
-    enable_embedding,
-    embedding_params,
-    auto_apply_filters,
-    ...dashboardDetails
-  } = {}) => {
+  (
+    {
+      name = "Test Dashboard",
+      enable_embedding,
+      embedding_params,
+      auto_apply_filters,
+      ...dashboardDetails
+    } = {},
+    { wrapId = false, idAlias = "dashboardId" } = {},
+  ) => {
     cy.log(`Create a dashboard: ${name}`);
 
     // For all the possible keys, refer to `src/metabase/api/dashboard.clj`
     cy.request("POST", "/api/dashboard", { name, ...dashboardDetails }).then(
       ({ body }) => {
+        if (wrapId) {
+          cy.wrap(body.id).as(idAlias);
+        }
         if (enable_embedding != null || auto_apply_filters != null) {
           cy.request("PUT", `/api/dashboard/${body.id}`, {
             enable_embedding,
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -1,18 +1,47 @@
 import {
   addOrUpdateDashboardCard,
+  dashboardHeader,
+  editDashboard,
+  getDashboardCard,
+  modal,
+  popover,
   restore,
+  saveDashboard,
   resetTestTable,
   resyncDatabase,
-  visitDashboard,
-  editDashboard,
   showDashboardCardActions,
   sidebar,
+  visitDashboard,
+  visitEmbeddedPage,
 } from "e2e/support/helpers";
 import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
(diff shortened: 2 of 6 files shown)


e2e/support/commands/api/dashboard.js
```diff
@@ -1,17 +1,23 @@
...
+    { wrapId = false, idAlias = "dashboardId" } = {},
+  ) => {
     cy.log(`Create a dashboard: ${name}`);
 
     // For all the possible keys, refer to `src/metabase/api/dashboard.clj`
     cy.request("POST", "/api/dashboard", { name, ...dashboardDetails }).then(
       ({ body }) => {
+        if (wrapId) {
+          cy.wrap(body.id).as(idAlias);
+        }
```
WiNloSt, line 20: Added a change from `master` that's needed in the E2E tests

https://github.com/metabase/metabase/blob/18742132b608d15e3298d8a5dae2377a3f574a66/e2e/support/commands/api/dashboard.js#L18-L20

frontend/src/metabase/visualizations/lib/action.unit.spec.ts
WiNloSt, line 1: There wasn't this file on the 47 branch, so I copy the whole file from `master`

frontend/src/metabase/visualizations/types.ts
WiNloSt, line 1: What I did in this file is basically to export `UrlClickAction`. You can see the type in master at

https://github.com/metabase/metabase/blob/18742132b608d15e3298d8a5dae2377a3f574a66/frontend/src/metabase/visualizations/types/click-actions.ts#L79

And follow its dependencies. I also removed most of the `export` keywords from its dependencies, because they aren't used in the 47 branch.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
WiNloSt, line 1: There are 3 things I basically did to this file.
1. Copied the whole "allows setting dashboard without filters as custom destination and changing it back to default click behavior" test, because I modified that test on `master`, but there isn't even that test in 47 branch
2. Copied the whole "allows setting saved question as custom destination and changing it back to default click behavior" test. These 2 tests will assert that we're redirecting to custom questions and dashboards with router not page reload.
3. Copied the whole "allows opening custom URL destination that is not a Metabase instance URL using link (metabase#33379)" test. This test was added on `master` so the change should match.

Then I added all the dependencies and try to put them in the same order as in `master`

frontend/src/metabase/visualizations/lib/action.unit.spec.ts
WiNloSt, line 1: There wasn't this file on the 47 branch, so I copy the whole file from `master`

frontend/src/metabase/visualizations/types.ts
WiNloSt, line 1: What I did in this file is basically to export `UrlClickAction`. You can see the type in master at

https://github.com/metabase/metabase/blob/18742132b608d15e3298d8a5dae2377a3f574a66/frontend/src/metabase/visualizations/types/click-actions.ts#L79

And follow its dependencies. I also removed most of the `export` keywords from its dependencies, because they aren't used in the 47 branch.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
WiNloSt, line 1: There are 3 things I basically did to this file.
1. Copied the whole `allows setting dashboard without filters as custom destination and changing it back to default click behavior` test, because I modified that test on `master`, but there isn't even that test in 47 branch
2. Copied the whole `allows setting saved question as custom destination and changing it back to default click behavior` test. These 2 tests will assert that we're redirecting to custom questions and dashboards with router not page reload.
3. Copied the whole `allows opening custom URL destination that is not a Metabase instance URL using link (metabase#33379)` test. This test was added on `master` so the change should match.

Then I added all the dependencies and try to put them in the same order as in `master`

frontend/src/metabase/visualizations/types.ts
WiNloSt: there has been too much changes there there are too many files, and I don't think copying the whole folder is a good idea because we'll leave too much code unused.

npretto: Yeah that's fair, I'll approve

npretto (approved):

Issue #36452, opened 2023-12-06
[Epic]  Sync column index information


qnkhuat:
**Links**
- product doc: https://www.notion.so/metabase/Ingest-metadata-about-indexes-index-distribution-keys-etc-6c72d1f631c942e29eef292873e57d1e
- eng doc: https://www.notion.so/metabase/Sync-db-indexes-ae13269005934e99837c77fe355ec9ab?pvs=4
- feature branch: `branch-name` _this should be the feature branch where this work will be done in. PRs will be delivered against this branch_
- issue links: https://github.com/metabase/metabase/issues/28382#top
- database tiers list: https://www.notion.so/metabase/Database-Tiers-for-Feature-Development-fabba4d2bbfb4c729029f5e77d585f61

**Implementation Plan**

***Milestone 1*** Sync indexes for Postgres, H2, Mysql
- [x] Migration to add `metabase_field.is_indexed`
- [x] Sync step to sync index information

PR: https://github.com/metabase/metabase/pull/36462

***Milestone 2*** Sync for mongo
- [x] Mongo

PR: #36624 

***Milestone 3*** Sync for tier 4 databases that support indexing (optional)
(None of the drivers in tier 3 supports indexing)
- [x] SQL Server
- [x] Sqlite

PR: https://github.com/metabase/metabase/pull/36665

For reference, this is a matrix of which driver supports indexing / partitioning / clustering

Engine | Storage type | Indexing | Partitioning | Clustering
-- | -- | -- | -- | --
Postgres | Row | ✅ | ✅ | ✅
MySQL | Row | ✅ | ✅ | ❌
BigQuery | Column | ❌ | ✅ | ✅
Presto | Row | ❌ | ✅ | ❌
Redshift | Column | ❌ | ✅ | ✅
Snowflake | Column | ❌ | ❌ | ✅
Mongo | Document | ✅ | ❌ | ❌
Vertica | Column | ❌ | ✅ | ✅
SparkSQL | Column | ❌ | ✅ | ❌
Athena | Column | ❌ | ✅ | ❌
Druid | Column | ❌ | ✅ | ✅
Oracle | Row | ✅ | ✅ | ✅
SQLServer | Row | ✅ | ✅ | ✅
SQLite | Row | ✅ | ❌ | ❌

PR #36462, opened 2023-12-06
Sync index info

qnkhuat:
Milestone 1 of https://github.com/metabase/metabase/issues/36452. 

This adds a new column `metabase_field.database_indexed`, it's set to true during sync if there is a single index on the column, or it's the 1st column of a composite index.

For DBs that do not support indexing, `metabase_field.database_indexed` should be nil.


src/metabase/driver/sql_jdbc/sync/describe_table.clj
```diff
@@ -274,6 +274,36 @@
...
+                                              (:name table)
+                                              ;; when true, return only indices for unique values when
+                                              ;; false, return indices regardless of whether unique or not
+                                              false
+                                              ;; when true, result is allowed to reflect approximate or out of data
+                                              ;; values. when false, results are requested to be accurate
+                                              false)]
+
+       (-> (group-by :index_name (into []
+                                       (filter #(and (= (:type %) DatabaseMetaData/tableIndexOther) ;; we only care about other user defined index types
```
calherries, line 298: Why exclude `tableIndexHashed` and `tableIndexClustered`? These are still types of indices that we might care about. 

See https://learn.microsoft.com/en-us/sql/relational-databases/indexes/clustered-and-nonclustered-indexes-described?view=sql-server-ver16 for clustered indices on SQL Server.

And https://www.postgresql.org/docs/current/indexes-types.html#INDEXES-TYPES-HASH for hashed indices on Postgres.

src/metabase/driver.clj
```diff
@@ -532,7 +538,10 @@
     :connection-impersonation-requires-role
 
     ;; Does the driver require specifying a collection (table) for native queries? (mongo)
-    :native-requires-specified-collection})
+    :native-requires-specified-collection
+
+    ;; Does the driver support column(s) indexing
```
calherries, line 543: I wonder if this should read: 
`Does the driver support column(s) support storing index info`

Implementing the multimethods to sync index info should be separate from whether indices are supported on the database or not.

Also wondering whether the feature should be called `:index-info`

calherries (changes requested):

src/metabase/driver/postgres.clj
```diff
@@ -59,7 +59,8 @@
                               :now                      true
                               :persist-models           true
                               :schemas                  true
-                              :connection-impersonation true}]
+                              :connection-impersonation true
+                              :indexing                 true}]
```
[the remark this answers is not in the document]
qnkhuat: no I didn't, redshift does not support indexing.
fixed in https://github.com/metabase/metabase/pull/36462/commits/9b9292dc3f732f3536e7c16ee256a9057c7744c3

src/metabase/driver/sql_jdbc/sync/describe_table.clj
qnkhuat: right! I thought if you create an index of type hash, it's still a user index type but turns out it's not.

for clustered index on postgres, it requires an index to work on so I thought since we'll sync the original index already, we can ignore it. Turns out if you cluster a table with an index, that index will have type "clustered".

anw I synced all of them in https://github.com/metabase/metabase/pull/36462/commits/9b9292dc3f732f3536e7c16ee256a9057c7744c3 and added tests

src/metabase/driver.clj
qnkhuat: yea index-info sounds better to me, renamed in https://github.com/metabase/metabase/pull/36462/commits/07847b53438d0b22d3f058ace91904b64953dbc7

test/metabase/test/data/sql/ddl.clj
```diff
@@ -52,9 +52,11 @@
             (sql.tx/create-table-sql driver dbdef tabledef)))
     ;; Add the SQL for adding FK constraints
     (doseq [{:keys [field-definitions], :as tabledef} table-definitions
-            {:keys [fk], :as fielddef}                field-definitions]
+            {:keys [fk indexed?], :as fielddef}                field-definitions]
       (when fk
-        (add! (sql.tx/add-fk-sql driver dbdef tabledef fielddef))))
+        (add! (sql.tx/add-fk-sql driver dbdef tabledef fielddef)))
+      (when indexed?
+        (add! (sql.tx/create-index-sql driver (:table-name tabledef) [(:field-name fielddef)]))))
```
calherries, line 59: Adding this to the FK `doseq` block is inconsistent with the way column comments are added on lines 64-68 below, which has its own separate `doseq`. I don't care either way, but as long as its consistent it'll be easier to read and understand

calherries (commented):
LGTM, just a couple of things to tidy up

calherries (approved):
LGTM, just a couple of things to tidy up

PR #36474, opened 2023-12-06
Cypress TypeScript PoC

nemanjaglumac:
### Status
PENDING REVIEW 👀 

## What does this PR accomplish?
This PR is a PoC for introducing TypeScript to our Cypress codebase.

In this first phase our goal is to improve intellisense support, and to increase the chance of a helper being found. Type safety is of lesser concern at the moment and can be considered as a non-goal at this stage. With that in mind, we will start converting the existing E2E helpers to TS first.

## Challenges
We have a fair number of Cypress custom commands. They are used by many of the "regular" E2E helpers. Trying to convert such helper to TS yields an error because TS doesn't see our custom command as part of Cypress types. We have to register them.

> **Note**
> Dealing with custom commands should've been a second phase of this project but it seems impossible to separate helpers from custom commands.

## Potential Solutions
1. Introduce a single type definition file and keep all type definitions there
    a. This file can live in the current dir structure (e.g. `e2e/support/commands.d.ts`) or
    b. It can live in a separate folder, or even as a separate repo (akin to the [types for the Cypress testing library](https://www.npmjs.com/package/@types/testing-library__cypress))
2. Keep type definition for a custom command next to the code (in the custom command file itself)

**This PR uses the second approach**

#### Global scope augmentation

Converting the custom command to TypeScript out of the box yields this warning:
```
Augmentations for the global scope can only be directly nested in external modules or ambient module declarations.ts(2669)
```
image

The solution is to either (1) import a typed function `foo` to a custom command file and then use custom command api to add it to the `cy.` chain like so:
```ts
import { foo } from "./foo";

Cypress.Commands.add("foo", foo);
```
or 

(2) to have an empty export in the custom command file like so
```ts
export {};
```
The first approach adds unnecessary complexity and comes with its own quirks.
This PR uses the second approach instead.

#### Modules vs Namespaces
Now we're left with this eslint warning
```
ES2015 module syntax is preferred over custom TypeScript modules and namespaces.
```
image

The solution was to [tweak the e2e eslint config file and to turn this rule off](https://github.com/metabase/metabase/pull/36474/files#diff-edf96cd5b7b54f680325adcf88783f257dd14cc6c65119e04cc158df6b1699f6R7). Since this is scoped only to E2E files, it shouldn't be a controversial decision.

With this we're finally warning and error-free. No more obstacles in converting both helpers and custom commands to TS.
image

###  JSDoc?
JSDoc + TypeScript = super powers 💥 

JSDoc adds a nice layer of documentation with useful links and examples. The bottom line is: we use TS for type annotation and JSDoc for user (developer) documentation.


e2e/.eslintrc
iethree, line 1: @kulyk note: re trying to unify eslint configs. this might need to stay separate

e2e/support/helpers/e2e-notebook-helpers.ts
```diff
@@ -0,0 +1,201 @@
...
+ * to assert on the waited query response.
+ *
+ * @example
+ * visualize();
+ *
+ * visualize(response => {
+ *   expect(response.body.error).to.not.exist;
+ * });
+ */
+export function visualize(callback?: (response?: object) => void) {
```
kamilmielnik, line 42: Ok, so it's already useful to type the responses.

Eventually, when we start to convert `services.js` to TS, we will also need these types there. We should start adding them in the codebase and e2e tests should just import these types from there.

e2e/support/helpers/e2e-notebook-helpers.ts
```diff
@@ -0,0 +1,201 @@
...
+  metric:
+    | "Count of rows"
+    | "Sum of ..."
+    | "Average of ..."
+    | "Number of distinct values of ..."
+    | "Cumulative sum of ..."
+    | "Cumulative count of rows"
+    | "Standard deviation of ..."
+    | "Minimum of ..."
+    | "Maximum of ...";
```
kamilmielnik, line 75: 1. These aren't all [basic aggregation operators](https://github.com/metabase/metabase/blob/6265ac4fffe4889efa8175cb75fc78d10affaa1a/frontend/src/metabase-lib/operators/constants.js#L347-L355).
2. I don't think we should be adding types based on locale-dependent labels.

```suggestion
  metric: string;
```

e2e/support/helpers/e2e-notebook-helpers.ts
```diff
@@ -0,0 +1,201 @@
...
+      cy.findByText(table).click();
+    }
+    if (field) {
+      cy.findByText(field).click();
+    }
+  });
+}
+
+/**
+ * Group by.
```
kamilmielnik, line 99: It's called a [breakout](https://github.com/metabase/metabase/wiki/Metabase-Glossary).

Same for `removeSummaryGroupingField`.

kamilmielnik (commented):
Looks good :+1: Can't wait to start using it!

We need to solve the problem of typing API responses. If we want to do it in a separate PR, then this one should at least use `unknown` instead of `any` and `object` which conceal the problem.

> it seems impossible to separate helpers from custom commands.

What do you mean by that?

e2e/support/helpers/e2e-notebook-helpers.ts
nemanjaglumac: This is a good example of why types don't necessarily need to be shared between FE and E2E. From my point of view, this helper doesn't care about all aggregation operators at all. On the contrary. The function of these enumerated types is to inform you (consumer) of which possible values you can use when abstracting the user flow with this helper. And those are the strings that you can see in the UI.

The second argument is much stronger, imo. But then again, 99.999% of all our E2E tests are using English. There are maybe one or two tests that are testing different locales specifically.

nemanjaglumac:
>> it seems impossible to separate helpers from custom commands.

> What do you mean by that?

Sorry for the lack of clarity.
I meant that it's impossible to type only helpers in the first pass, and then do custom commands in the second one, because a lot of helpers rely on custom commands. So we need to type everything in the same pass.

e2e/support/helpers/e2e-notebook-helpers.ts
ranquild: I'd go with a `string` like @kamilmielnik suggests.

nemanjaglumac: Done https://github.com/metabase/metabase/pull/36474/commits/a65b52eb0d387400381538d6d2d06559a8e1a017

WiNloSt (commented):
I don't have major concerns about this PR. Looks overall very good to me. But I'll let others approve since they have more feedback.

Great work 🎉

e2e/support/helpers/e2e-notebook-helpers.ts
nemanjaglumac: After a call with @kamilmielnik , the deal was to use this
https://github.com/metabase/metabase/pull/36474/commits/b13f8dd2c2fa9db06d71a395277905db14e37265

kamilmielnik (approved):

PR #36478, opened 2023-12-06
[Help link] set error to null when typing

npretto:
Epic: https://github.com/metabase/metabase/issues/35915

### Description

From [this discussion](https://metaboat.slack.com/archives/C063Q3F1HPF/p1701797891625189?thread_ts=1701769982.351929&cid=C063Q3F1HPF) we decided to hide the error while a user is typing to have a better UX.

Scenario is:
1 user types "httpsssss://example.org" -> nothing happens
2 he blurs to save it -> error appears
3 he tries to edit it -> error goes away on the first keystroke
4 he blurs it -> if still invalid, it errors

See demo in loom here: https://www.loom.com/share/b477455ac6f746a2886e6f3957ea9086


WiNloSt (approved):
Tested. This looks good to me.

PR #36479, opened 2023-12-06
add target branch name to backporting branch for less conflicts

sansolovyov:
I've got into conflicts using `@metabase-bot backport ...` command twice in a row, since it used the same branch name for both of backports (to 47 and 48). This fixes the issue.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -66,7 +66,7 @@ jobs:
               ref: `heads/${targetBranch}`,
             });
 
-            const backportBranch = `backport-${originalPullRequest.head.ref}`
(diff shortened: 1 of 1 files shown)


nemanjaglumac (approved):
This solution solves the problem you faced.
It will create an even longer branch name, but we can deal with it separately. Not that anyone cares about the branch name, anyway :)

Just as an example
```
backport-a6cba791f27354854be2d63696290e32d6d9b754-release-x.47.x`
```

piranha:
Nah, it'll be `backport-backport-branch-name-release-x.48.x` if I backport this PR. :)

Issue #14504, opened 2021-01-23
CSV exports of timestamps at midnight (00:00:00) is truncated to date


flamber:
**Describe the bug**
CSV (and JSON) downloads of timestamp columns at midnight (00:00:00) is truncated to date. XLSX are correct (since 0.36.8).

**To Reproduce**
1. Native question > Sample Dataset:
```sql
SELECT CAST(TIMESTAMP '2020-01-01 00:00:00' AS TIMESTAMP(0)) AS "DATETIME", CAST(TIMESTAMP '2020-01-01 00:00:00' AS DATE) AS "DATE"
UNION ALL 
SELECT CAST(TIMESTAMP '2020-01-02 00:00:22' AS TIMESTAMP(0)), CAST(TIMESTAMP '2020-01-02 00:00:22' AS DATE)
UNION ALL 
SELECT CAST(TIMESTAMP '2020-01-03 03:03:03.333' AS TIMESTAMP(0)), CAST(TIMESTAMP '2020-01-03 03:03:03.333' AS DATE)
```
2. Export as CSV returns the following:
```
DATETIME,DATE
2020-01-01,2020-01-01
2020-01-02T00:00:22,2020-01-02
2020-01-03T03:03:03,2020-01-03
```

**Expected behavior**
```
DATETIME,DATE
2020-01-01T00:00:00,2020-01-01
2020-01-02T00:00:22,2020-01-02
2020-01-03T03:03:03,2020-01-03
```

**Information about your Metabase Installation:**
Tested 0.34.3 thru 0.37.7 - regression since 0.36.8, which fixed a lot of export issues in PR #13447

dpsutton:
I'm gonna disagree with the `Difficulty:Easy` label here. This behavior is intentional. The code in `metabase.query-processor.streaming.common/format-value` explicitly does this for `LocalDateTime`:

```clojure
(extend-protocol FormatValue
  ...
  LocalDateTime
  (format-value [t]
    (if (= (t/local-time t) (t/local-time 0))
      (format-value (t/local-date t))
      (u.date/format t)))
   ...
)
```

We've been using the presence of time at midnight to indicate the time is irrelevant and an artifact of using the wrong type (LocalDateTime) when we are dealing with only Date information. Blindly removing this check will just reintroduce the bug that this is fixing.

We do have the cols metadata at one stage of writing results. We could get the index of any `type/Date` columns and add some wrapper or type coercion to ensure that they get formatted correctly.

```clojure
[...]

Issue #17422, opened 2021-08-11
CSV/JSON export formatting


noahmoss:
Number, currency and datetime format settings should be respected in CSV and JSON exports.

An open question is what subset of format settings would be expected to be applied. Datetime format settings should probably always be applied. Number/currency formatting is a bit more ambiguous, and may depend on the use case.

For example: Metabase always displays thousands separators for numbers on the frontend. JSON doesn't support this without changing the data type from a number to a string. CSV also doesn't support this unless the values are quoted.

This issue replaces #10879 since XLSX formatting will be supported in v41 of Metabase.

:arrow_down: Please click the :+1: reaction instead of leaving a `+1` or `update?` comment

WhyNotHugo:
While date format seems relevant for CSVs, I think including thing like currency formatting or thousand separators can be problematic. CSV is often processed by automation or different kind of tools that would choke on something like "$1.000"  instead of "1000".

I think it's smart to have started with XSLX, which is what humans would usually work on.

gang-qiu-social-chorus:
i'm experiencing the issue of the exported CSV headers having the database column names, when we want to have the customized formatted name. Is there a hacky way to implement this for our personal use while the issue is in review for the core product?

flamber:
@gang-qiu-social-chorus You would have to use Custom Columns in GUI question or SQL with aliasing.
Please use the forum for questions and troubleshooting: https://discourse.metabase.com/

colombian-cannonball:
Hello, is this issue still being worked on? Not having the same column names that appear in a question or dashboard in the downloaded CSV renders reports useless when stakeholders want to export them.

devoto13:
This feature will also be helpful to ensure consistent number formatting in the exported file. Today I have faced an interesting behaviour where in CSV export most of the numbers were in the normal style, but a couple used scientific notation (e.g. `1.584235248E7`) which was not expected by the file consumer.

amureki:
Greetings!

It appears that v49 update has introduced a change that has impacted our user experience somewhat negatively.

Here is an example:
We have an org default in Metabase:
- separator style set to "100 000,00"
- dates style is "31 Jan, 2018"

This is the preferred way for us in all the reports we have in Metabase.
However, we rely heavily on automations that utilize JSON or CSV exports from these reports.
And now, since recent version, the JSON do contain the following:

```
[
  {
    "Created Date": "13 Mar, 2024",
    "Count": "188 857"
  },
...
]
```

[...]

PR #36484, opened 2023-12-06
Apply column formatting to CSV exports

markbastian:
This PR applies the same formatting logic to CSV exports as it does to pulse bodies (e.g. HTML).

Formerly, there were two related formatting paths:
- `metabase.query-processor.streaming.common/format-value`, which is from a protocol that takes an object and returns a string. It is what was used to export csv data. It applies no actual formatting, only converts objects to strings.
- `metabase.pulse.render.body/get-format`, which builds a formatter from an optional time zone, column metadata, and visualization settings. This formatter takes a string and formats it. It was only used for rendering inline artifacts, such as embedded HTML in emails.

The first function is insufficient to format row data as it is unaware of the column metadata and viz settings. We need to use that data everywhere data is exported in a uniform way.

The solution is to lift `get-format` from `metabase.pulse.render.body` to a more common location (`metabase.pulse.render.common` in this PR step, but needs to be moved out of the pulse code to be a more shared concern) and use it everywhere artifacts are generated.

For csv export, this was achieved as follows in `metabase.query-processor.streaming.csv`:

```clojure
(defmethod qp.si/streaming-results-writer :csv
  [_ ^OutputStream os]
  (let [writer     (BufferedWriter. (OutputStreamWriter. os StandardCharsets/UTF_8))
        formatters (atom {})]
    (reify qp.si/StreamingResultsWriter
      (begin! [_ {{:keys [ordered-cols results_timezone]} :data} viz-settings]
        (swap! formatters (constantly (zipmap
                                        ordered-cols
                                        (map (fn [col]
                                               (p.common/get-format results_timezone col viz-settings))
                                             ordered-cols))))
        (csv/write-csv writer [(map (some-fn :display_name :name) ordered-cols)])
        (.flush writer))

      (write-row! [_ row _row-num cols {:keys [output-order]}]
        (let [[ordered-row
               ordered-cols] (if output-order
                               (let [row-v  (into [] row)
                                     cols-v (into [] cols)]
                                 [(for [i output-order] (row-v i))
                                  (for [i output-order] (cols-v i))])
                               [row cols])]
          (csv/write-csv writer [(map (fn [c r]
                                        (let [formatter (@formatters c)]
                                          (formatter (common/format-value r))))
                                      ordered-cols ordered-row)])
          (.flush writer)))

      (finish! [_ _]
        ;; TODO -- not sure we need to flush both
        (.flush writer)
        (.flush os)
        (.close writer)))))
```

The formatters for each column are built in the `begin!` function and then looked up in each `write-row!`.  The existing `format-value` is used to produce a string then passed into our looked up column formatter.

Note that the new unit tests simulate a pulse and grab the provided temp files as attachments and analyzes those for correctness. This should work in a CI environment so long as the test process has permission to both write attachments to the temp directory and read those attachments back out. Also note that these tests can be slow (but not terribly so).

Primary changes:
- `metabase.email.messages` - fix spelling
- `metabase.pulse.render.body` - move `get-format` out of this ns
- `metabase.pulse.render.common` - move `get-format` into this ns
- `metabase.query-processor.streaming.csv` - new logic to apply pulse renderer formatting to csv exports
- `metabase.pulse.pulse-integration-test` - adding unit tests

One TODO before we complete #36484 is a follow-on PR to move the `get-format` logic out of a pulse ns into something more general. Ultimately, it would be nice if this was a common capability used by both BE and FE. That will be a follow-on since it will be noisy and it would be nice to be able to see the changes independently.

The follow-on refactor PR is #36490.

### DIFF
--- a/e2e/test/scenarios/sharing/downloads/reproductions/10803-timestamp-formatting.cy.spec.js
+++ b/e2e/test/scenarios/sharing/downloads/reproductions/10803-timestamp-formatting.cy.spec.js
@@ -51,7 +51,7 @@ describe("issue 10803", () => {
 
       // Excel and CSV will have different formats
       if (fileType === "csv") {
-        expect(sheet["A2"].v).to.eq("2026-06-03");
+        expect(sheet["A2"].v).to.eq("2026-06-03T00:00:00");
         expect(sheet["B2"].v).to.eq("2026-06-03T23:41:23");
       } else if (fileType === "xlsx") {
         // We tell the xlsx library to read raw and not parse dates
--- a/src/metabase/email/messages.clj
+++ b/src/metabase/email/messages.clj
@@ -389,14 +389,14 @@
       (no "less than %d rows in results" body/rows-limit))))
 
 (defn- stream-api-results-to-export-format
-  "For legacy compatability. Takes QP results in the normal `:api` response format and streams them to a different
+  "For legacy compatibility. Takes QP results in the normal `:api` response format and streams them to a different
   format.
 
-  TODO -- this function is provided mainly because rewriting all of the Pulse/Alert code to stream results directly
+  TODO -- this function is provided mainly because rewriting all the Pulse/Alert code to stream results directly
   was a lot of work. I intend to rework that code so we can stream directly to the correct export format(s) at some
   point in the future; for now, this function is a stopgap.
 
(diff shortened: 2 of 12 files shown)


adam-james-v (approved):
Looks good!

paoliniluis:
CSV part of https://github.com/metabase/metabase/issues/17422, will probably fix https://github.com/metabase/metabase/issues/14504

PR #36485, opened 2023-12-06
CSV appends: Merge 1, Milestone 0: add endpoint to append to table (part one)

calherries:
feature branch: `appends/milestone-0-endpoint` ([PR](https://github.com/metabase/metabase/pull/36640))

This PR is the first of many that will implement "Merge 1, Milestone 0" in the [eng doc](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df). I want to get this approved before this PR becomes too large to review well.

A copy of the current version of the eng doc is given below, with check-boxes ticked where behaviour is implemented.

The main things missing Merge 1, Milestone 0 are:
- special handling of `_mb_row_id`
- handling failure during insertion of rows
- permissions errors

I have also left some `TODO` comments where implementations are not complete yet.

**Eng spec:**

- Create a new endpoint **`POST /api/table/:id/append-csv`**
- Appends the rows in an uploaded CSV file to a table that was created with a CSV upload (the table must have `is_upload=true`)
- If the schema matches, inserts the CSV rows to the table. Undo all changes if there are any failures
- It must work for all drivers that support the “uploads” feature: MySQL, H2, and Postgres

**Schemas:**

- request: needs a `formData` body with `formData.append("file", file)` where `file` is a `File` (same as `POST /api/card/from-csv`). The table ID identifying the table to append to is in the path: **`/api/table/:id/append-csv`**
- response:
    - either an error:
        
        ```jsx
        { 
        	status: <404, 403, 422, or 500>,
          body: {
        		message: "something went wrong",
          }
        }
        ```
        
    - or success:
        
        ```jsx
        { 
        	status: 200,
          body: {
            row_count: <number of rows uploaded>
          }
        }
        ```
        

- [x]  Test: Happy path with a CSV file that contains all the upload column types
- [x]  Test: Happy path with a CSV file that contains only the header, and no rows

**Settings:**

- A similar configuration is required as uploading a CSV (`POST /api/card/from-csv`), except the schema and upload prefix can have changed since the uploaded table was created.
- [x]  Test: if uploads are not enabled, append fails
- [x]  Test: if upload database and schema aren’t set, append succeeds

**Special behaviour around _mb_row_id:**

- If a CSV file with `_mb_row_id` is uploaded, we will ignore the `_mb_row_id` column in the CSV
- If a CSV file without `_mb_row_id` is uploaded, we will create the column anyway
- This column is assumed to be auto-incrementing. We use the max value of this to roll back changes when failures occur during insertion
- [x]  Test: If a CSV file with `_mb_row_id` is uploaded to a table doesn’t have it, we will ignore the `_mb_row_id` column in the CSV but create the column anyway
- [x]  Test: If a CSV file without `_mb_row_id` is uploaded and the table doesn’t have it either, we will create the column anyway
- [x]  Test: If a CSV file without `_mb_row_id` is uploaded and the table does have it, we can upload just fine (it’s not required)

**Failures:**

- Each type of failure will be thrown as `(ex-info (tru <message>) {:type :upload.error-type/....}` and converted to the correct format for sending over the wire in the API endpoint handler. There will only be one type of error thrown at a time.
- These are the types of failures to consider:
    1. CSV file is empty (status: 422):
        - Message: Column mismatch error message (see below)
        - [x]  Test: if the CSV file is empty, return the correct message and status
    2. Table not found (status: 404)
        - Message: “Not found.”
        - Can be implemented by calling api/write-check on the table
        - [x]  Test: if a database doesn’t exist then the correct message and status is returned
    3. Table does not have `is_upload=true` i.e. the table wasn’t created from an upload (status: 422):
        - Message: “The table must be an uploaded table”
        - [x]  Test: if table is not is_upload=true then the correct message and status is returned
    4. Uploads must be enabled via the setting `uploads-enabled` (status: 422)
        - Message: “Uploads must be enalbed.”
        - [x]  Test: if uploads-enabled ≠ true then the correct message and status is returned
    5. Permissions errors (status: 403):
        - Message: “You don’t have permissions to do that.”
        - Same permissions checks as uploading a CSV (`POST /api/card/from-csv`) and unrestricted data permissions on the table
        - [x]  Test: if user has block permissions: native query editing is required to upload CSV
        - [x]  Test: if user doesn’t have data permissions then upload fails
        - [x]  Test: if user is sandboxed then upload fails
        - [x]  Test: if the user does have unrestricted data permissions but no native query editing permissions then upload fails
    6. Columns mismatch (status: 422):
        - Message: “The columns of the uploaded CSV file don’t match the columns of the table. Extra columns: …. Missing columns: …”
        - Details:
            - All columns in the table need to be present in the CSV file, with exact string comparison on the normalized column name, unless they are autogenerated by the database (such as the autogenerated `id` column we make for upload tables)
            - The order of the columns doesn’t matter
            - There must be no duplicate column names in the CSV file
            - There must be no extra columns in the CSV file that are not present in the table.
            - If these conditions aren’t satisfied, throw a columns mismatch error (see next section)
        - [x]  Test: If there are extra columns AND/XOR missing columns, return an appropriate error
        - [x]  Test: If the columns in the append CSV are in a different order to the originally uploaded table, the upload should succeed
        - [x]  Test: If the column names are different when unnormalized but are the same normalized, the upload should succeed
        - [x]  Test: If there are duplicate columns, return an appropriate error


src/metabase/upload.clj
```diff
@@ -483,3 +489,133 @@
...
+    - `parse-rows`, a function that takes a lazy-seq of rows, and returns a lazy-seq of rows. It parses the
+      string values in each row and removes columns that have the same normalized name as the generated columns.
+    - `table-col-names`, a seq of the column names in the table, in the same order as the columns in the CSV file.
+
+  Throws an exception if:
+    - the CSV file contains duplicate column names
+    - the schema of the CSV file does not match the schema of the table"
+  [table header]
+  ;; Assumes table-cols are unique when normalized
+  (let [table-cols           (t2/select :model/Field :table_id (:id table))
```
qnkhuat, line 521:
```suggestion
  (let [table-fields           (t2/select :model/Field :table_id (:id table))
```

I rarely see we use the word "col" for Field here, I think we should stick with `fields` for 2 reasons:
- cols should be reserved for column in csv file
- table-cols can also be understand as "table collection"

src/metabase/upload.clj
```diff
@@ -483,3 +489,133 @@
...
+                                              [(normalize-column-name (:name col)) col]))
+                                       table-cols)
+                                 (dissoc "_mb_row_id"))
+        normalized-header+header (into []
+                                       (map (fn [col]
+                                              [(normalize-column-name col) col]))
+                                       header)
+        normed first
+        normalized-header (map normed normalized-header+header)
+        normalized-table-cols (map normed table-cols-by-normed)
```
qnkhuat, line 533:
```suggestion
        normalized-header-name (map first normalized-header+header)
        normalized-table-cols-name (keys table-cols-by-normed)
```

why did you alias `normed as first`?

src/metabase/upload.clj
```diff
@@ -483,3 +489,133 @@
...
+        ;; check for duplicates
+        _ (when (some #(< 1 %) (vals (frequencies normalized-header)))
+            (throw (ex-info (tru "The CSV file contains duplicate column names.")
+                            {:status-code 422})))
+        [extra missing _both] (data/diff (set normalized-header)
+                                         (set normalized-table-cols))]
+
+    (if (and (nil? extra) (nil? missing))
+      {:parse-rows (fn [rows]
+                     (let [rows' rows ;; TODO: remove auto-pk columns here
```
qnkhuat, line 543: why do we need this extra binding?

src/metabase/upload.clj
```diff
@@ -483,3 +489,133 @@
...
+        normalized-header (map normed normalized-header+header)
+        normalized-table-cols (map normed table-cols-by-normed)
+        ;; check for duplicates
+        _ (when (some #(< 1 %) (vals (frequencies normalized-header)))
+            (throw (ex-info (tru "The CSV file contains duplicate column names.")
+                            {:status-code 422})))
+        [extra missing _both] (data/diff (set normalized-header)
+                                         (set normalized-table-cols))]
+
+    (if (and (nil? extra) (nil? missing))
```
qnkhuat, line 541:
```suggestion
    (if (= (set normalized-header) (set normalized-table-cols))
```

I think this is sufficient, or am I missing anything here?

test/metabase/upload_test.clj
```diff
@@ -1063,3 +1064,203 @@
...
+    (f)
+    (catch Exception e
+      {:message (ex-message e) :data (ex-data e)})))
+
+(defmacro catch-ex-info
+  [& body]
+  `(catch-ex-info* (fn [] ~@body)))
+
+(deftest can-append-test
+  (mt/test-driver :h2
```
qnkhuat, line 1116:
```suggestion
  (mt/test-driver   (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
```

test/metabase/upload_test.clj
```diff
@@ -1063,3 +1064,203 @@
...
+           is-upload          true}}]
+  (try
+    (mt/with-temporary-setting-values [uploads-enabled uploads-enabled]
+      (mt/with-current-user user-id
+        (mt/dataset (basic-db-definition (mt/random-name))
+          (let [tables (t2/select :model/Table :db_id (mt/id))
+                table (first tables)]
+            (assert (= (count tables) 1))
+            (t2/update! :model/Table (:id table) {:is_upload is-upload})
+            (@#'upload/append-csv! {:table-id (or table-id (:id table))
```
qnkhuat, line 1100:
```suggestion
            (upload/append-csv! {:table-id (or table-id (:id table))
```

test/metabase/api/table_test.clj
```diff
@@ -870,3 +871,30 @@
                       (map u/the-id))))))
       (finally (mt/user-http-request :crowberto :put 200 (format "table/%s" (mt/id :venues))
                                      {:field_order original-field-order})))))
+
+;;; +----------------------------------------------------------------------------------------------------------------+
+;;; |                                          POST /api/table/:id/append-csv                                        |
+;;; +----------------------------------------------------------------------------------------------------------------+
+
```
qnkhuat, line 878: we should have a test to check the file is deleted

src/metabase/upload.clj
```diff
@@ -483,3 +489,133 @@
...
+          {:row-count (count parsed-rows)})
+        (catch Throwable e
+          ;; TODO: insert rows failure handling
+          (throw (ex-info (ex-message e) {:status-code 400})))))))
+
+;;; +----------------------------------------------------------------------------------------------------------------+
+;;; |                                                 public interface                                               |
+;;; +----------------------------------------------------------------------------------------------------------------+
+
+(defn- can-append-error
```
qnkhuat, line 574: looks like a lot of these are the same as `can-upload-error`. could we reuse or generalize this helper?

src/metabase/upload.clj
```diff
@@ -483,3 +489,133 @@
...
+                               table-col-names
+                               parsed-rows)
+          {:row-count (count parsed-rows)})
+        (catch Throwable e
+          ;; TODO: insert rows failure handling
+          (throw (ex-info (ex-message e) {:status-code 400})))))))
+
+;;; +----------------------------------------------------------------------------------------------------------------+
+;;; |                                                 public interface                                               |
+;;; +----------------------------------------------------------------------------------------------------------------+
```
qnkhuat, line 572: this doesn't align with other blocks, for one this contains more than "public" interface and also the template is different.

qnkhuat (changes requested):
Is checking the type mismatch between the uploaded CSV and the table part of this PR?

nevertheless I think the flow is good, I still have a handful of suggestions

test/metabase/upload_test.clj
```diff
@@ -1063,3 +1064,203 @@
...
                             first :exists)))))))))
+
+;;; +----------------------------------------------------------------------------------------------------------------+
+;;; |                                           append-csv!                                                          |
+;;; +----------------------------------------------------------------------------------------------------------------+
+
+(defn basic-db-definition
+  "This creates a table with an auto-incrementing integer ID column, and a name column."
+  [database-name]
+  (tx/map->DatabaseDefinition
```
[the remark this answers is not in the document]
calherries: Nice, I like that it validates the field definitions too.

src/metabase/upload.clj
calherries: We don't now, but see the TODO, I'm going to update rows here. This is just a placeholder line for now

src/metabase/upload.clj
calherries: These are equivalent to me, but anyway I'm setting this up for more specific error messages in future PRs. We'll need to throw an error and display the extra columns and missing columns to the user if there are any. I think that's explicitly captured in `(and (nil? extra) (nil? missing))`.

src/metabase/upload.clj
calherries: Good question, it served as documentation that the first element in both `normalized-header+header` and `table-cols-by-normed` is the "normalized" version. But I think that should be obvious enough given the names, so I've accepted this suggestion

calherries:
I've gone through your suggestions and implemented most of them. Thanks

> checking the type mismatch between the uploaded CSV and the table part of this PR?

Indeed that's not part of this PR. Type mismatches (surfacing as parsing and insertion errors) will be done as part of "handling failure during insertion of rows".

qnkhuat (approved):
Cool, LGTM now

PR #36490, opened 2023-12-06
Refactoring formatting code for common utility

### LINKED ISSUE #36320: Ensure results metadata is applied to csv files in pulse exports
The `result-metadata-preservation-in-html-static-viz-test` test in `metabase.pulse.pulse-integration-test` from https://github.com/metabase/metabase/pull/36245 provides a good example of cases where results metadata is not applied to CSVs.

The task is to modify the codebase to apply formatting from result metadata to CSV files.

AC:
* Modify or duplicate the `metabase.pulse.pulse-integration-test/result-metadata-preservation-in-html-static-viz-test` to capture the csv attachments
* Add test assertions that:
  * The exported base question's csv does not reformat the `Tax Rate` column
  * The exported model's csv formats the `Tax Rate` column as a percentage
  * The exported question based on model's csv formats the `Tax Rate` column as a percentage

markbastian:
This PR refactors `metabase.pulse.render.common` to `metabase.formatter` as this is code we want applied to several locations, not just in pulses. It also updates references to these nses and the consistent alias.

A key observation of this formatting code, and reason for the refactor, is that it is "metabase-aware" in that it takes into account metadata columns and visualization settings when building formatters rather than just being a simple generic date or number formatter. This is a common code path that should be used any time we are rendering static assets and could potentially be used as common FE code with future development.

Moves:
- `metabase.pulse.render.common` to `metabase.formatter`
- `metabase.pulse.render.datetime` to `metabase.formatter.datetime`
- `metabase.pulse.render.common-test` to `metabase.formatter-test`
- `metabase.pulse.render.datetime-test` to `metabase.formatter.datetime-test`

With this change and #36484 we will close #36320.

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -298,6 +298,7 @@
     metabase.driver.sql.parameters.substitution                     sql.params.substitution
     metabase.email-test                                             et
     metabase.email.messages                                         messages
+    metabase.formatter                                              formatter
     metabase.http-client                                            client
     metabase.lib.util                                               lib.util
     metabase.mbql.normalize                                         mbql.normalize
@@ -336,7 +337,6 @@
     metabase.pulse.markdown                                         markdown
     metabase.pulse.render                                           render
     metabase.pulse.render.body                                      body
-    metabase.formatter                                              formatter
     metabase.pulse.render.style                                     style
     metabase.pulse.test-util                                        pulse.test-util
     metabase.pulse.parameters                                       pulse-params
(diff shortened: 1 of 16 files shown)


.clj-kondo/config.edn
```diff
@@ -336,7 +336,7 @@
     metabase.pulse.markdown                                         markdown
     metabase.pulse.render                                           render
     metabase.pulse.render.body                                      body
-    metabase.pulse.render.common                                    common
+    metabase.formatter                                              formatter
```
adam-james-v, line 339: Should this be sorted alphabetically?

markbastian: Good call. `kondo` should lint itself.

adam-james-v (approved):

PR #36499, opened 2023-12-06
Fix fields clauses with exprs against breakouts

### LINKED ISSUE #35067: [MLv2] - Removing field from multistage query with implicit breakout causes error
Currently in master, if you create a query that looks like this:
image

You are unable to remove the count or Double count fields, as it will result in an error. It looks as though the field ref that is generated for the implicitly joined breakout field is incorrect, as it doesn't understand what `PEOPLE__via__USER_ID__SOURCE` is.

image
image

snoe:
Fixes #35067

When you have a two stage query, and the second stage references a breakout in the first stage, adding a fields clause to the last stage, caused the qp to get confused.

The issue is that source-metadata was returning field-refs for breakouts (and aggregations) as seen from stage 0 instead of as seen from stage 1. It was using id references instead of names.

And then in stage 1, it was using those id references to compare against the `:fields` and not able to match `[:field <id>]` to `[:field <name>]`.

Updated the expected-cols middleware to produce `[:field <name>]` refs. And made sure that `:lib/desired-column-alias` was maintained through the annotation.


src/metabase/query_processor/middleware/annotate.clj
```diff
@@ -392,16 +369,38 @@
            :source :fields)))
 
 (mu/defn ^:private cols-for-ags-and-breakouts
```
metamben, line 371: I'm a bit confused by this name. This seems to be doing more than returning the columns for aggregations and breakouts. It does care about columns coming from fields when setting the source, but the call to `cols-for-fields` below is still there.

snoe:
This is not the approach to take, too many things rely on `source-metadata` returning id field-refs that making this change feels way way too fragile.

PR #32624, opened 2023-07-25
Global search: adding more filters

qnkhuat:
Milestone 2 of https://github.com/metabase/metabase/issues/27982

Sub PRs: 
- https://github.com/metabase/metabase/pull/32469
- https://github.com/metabase/metabase/pull/32658
- https://github.com/metabase/metabase/pull/32662
- https://github.com/metabase/metabase/pull/33146
- https://github.com/metabase/metabase/pull/32836
- https://github.com/metabase/metabase/pull/33525
- https://github.com/metabase/metabase/pull/34317

### BE

- `created_by` : this should be user id
- `verified` : `true` for verified only, `nil` to returns everything. `false` is invalid state.
- `created_at` : is a string like `past4hours` or `2023-04-01~2024-04-01`. For the full list check this [test](https://github.com/metabase/metabase/blob/master/test/metabase/driver/common/parameters/dates_test.clj#L130)

PR #36503, opened 2023-12-06
Use stretch variant on switch

iethree:
small nitpick tweak to #32624 

### Description

We don't need a styled component wrapper to get the flex: justify-between styling for the switch component, there is a `stretch` variant for that. Looks the same, less code. I think the variant may have been added after this component actually 🤔 .  

In general, if we're wrapping Mantine components to do simple things, we should consider making a variant, since they'll probably be needed elsewhere.

Screen Shot 2023-12-06 at 3 31 45 PM


oisincoveney (approved):
Nice catch, thanks for the change 😄

PR #36504, opened 2023-12-06
Bump dependencies based on vulnerabilities

snoe:


modules/drivers/googleanalytics/deps.edn
```diff
@@ -7,4 +7,5 @@
   com.google.oauth-client/google-oauth-client        {:mvn/version "1.34.1"}
   ;; for some reason, Google stopped depending on google-http-client-jackson2 from google-api-client somewhere between
   ;; 1.30.7 and 1.32.1, so we must explicitly bring it in because the google driver uses it directly
-  com.google.http-client/google-http-client-jackson2 {:mvn/version "1.43.3"}}}
+  com.google.http-client/google-http-client-jackson2 {:mvn/version "1.43.3"}
+  commons-codec/commons-codec                        {:mvn/version "1.16.0"}}}
```
metamben, line 11: It's not obvious why this is needed now.

modules/drivers/mongo/deps.edn
```diff
@@ -2,4 +2,6 @@
  ["src" "resources"]
 
  :deps
- {com.novemberain/monger {:mvn/version "3.6.0"}}}
+ {com.google.guava/guava {:mvn/version "32.1.3-jre"}
```
metamben, line 5: I see `com.google.guava/guava                    {:mvn/version "32.1.2-jre"}` in `metabase/deps.edn` on `master`. Shouldn't that be updated?

metamben: Doesn't the mongo driver inherit the root dependency? I would prefer not having to maintain the same thing at multiple places.

metamben (approved):
LGTM, but it would be even better if could have the guava dependency just once.

Issue #36253, opened 2023-11-30
[MLv2] [Testing Plan] Drill Thru


ranquild:
### automatic-insights

```[tasklist]
- [x] works for cell (dot) query (huh?)
- [x] works for pivot cell
- [x] works for legend item
- [x] not returned for query with no breakouts
- [x] checks `enable-xrays` Setting
- [x] removes filters
- [x] adds a filter for breakout `dimensions` using the same logic as in `underlying-records`
```

### column-filter

```[tasklist]
- [x] works for column header
- [x] checks column is not `:type/Structured`
```

### distribution

```[tasklist]
- [x] works for column header
- [x] not returned for query with aggregations or breakouts
- [x] checks column not `type/PK`, `type/SerializedJSON`, `type/Description`, `type/Comment`
- [x] Remove all aggregation, breakout, orderBy, limit clauses
- [x] Aggregate by "count" operator
- [x] Breakout by the selected column. If the column is a date column, add "Month" temporal unit. If the column is a numeric column, apply the default binning strategy. Otherwise use the column as it is.
```

### fk-details

```[tasklist]
- [x] works for Cell
- [x] checks Column is `type/FK`
- [x] checks `value` is not `null`
- [x] Finds the target table for the selected FK. Creates a new query that targets the FK table and add a `=` filter using the corresponding PK column and the selected value
- [x] Preserve filters from the original query that use `=` operator and are based on **other** FK columns targeting the same FK table as the selected FK. This can only happen when the target table's PK consists of multiple columns.
```

### fk-filter

```[tasklist]
- [x] works for Cell
- [x] checks Selected column is `type/FK`
- [x] Returns `columnName` and `tableName` for the FK column. On the FE we strip `ID` suffix and turn `Product ID` into `Product's` and pluralize the table name.
- [x] Add a `=` filter for the selected column and value. Make sure to append the query stage when needed.
```

### pivot

```[tasklist]
- [x] works for Cell
- [x] checks Query has at least 1 aggregation
- [x] checks Column from the aggregation clause was selected (???)
- [x] query shape is allowed: 1 aggregation and no breakouts - `type/Date`, `type/Address`, and `type/Category` (only which are not also `type/Address`)
- [x] query shape is allowed: At least 1 aggregation and exactly 1 breakout based on `Address` column - `Date`, `Category`
- [x] query shape is allowed: At least 1 aggregation and 1-2 breakouts based on `Category` columns - `Date`, `Category`
[...]

PR #36511, opened 2023-12-07
Copy drills dox from #36253

camsaul:
@ranquild did a good job of describing each of the different drill types in #36253, might as well include that information as docstrings in the code. Some of this info is FE-specific (not relevant to MLv2 code), but if none of this documentation ends up in the FE (😬) it would be good to have it somewhere accessible. Some is duplicated with what I already wrote, but let's just get this in and clean it up later


src/metabase/lib/drill_thru/fk_filter.cljc
```diff
@@ -1,4 +1,23 @@
 (ns metabase.lib.drill-thru.fk-filter
+  "Adds a simple `=` filter for the selected FK column. Enables option like`View this Product's Reviews`.
```
metamben, line 2:
```suggestion
  "Adds a simple `=` filter for the selected FK column. Enables option like `View this Product's Reviews`.
```

metamben (approved):
👍

PR #36519, opened 2023-12-07
[#35961 M2] Add setup modal

oisincoveney:
Epic #35961

### Description

Refactors the setup modal to match the designs found [here](https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=129-2053&mode=design&t=RDZFNZONOoxtV3T9-4).

Changed functionality:

* Static Embed:
	* When embedding is disabled, the button is disabled
	* When embedding is enabled and the resource is not published, we show "Set this up" on the button
	* When embedding is enabled and the resource is published, we show "Edit settings" on the button
* Public embed
	* If public sharing is disabled, the button is disabled and we have a link to the public sharing settings
	* If public sharing is enabled and a link doesn't exist, we should show "Get an embed link"
	* If public sharing is enabled and there is a link, we should show the link and a copy button immediately with the ability for admins to remove the link.

### How to verify

Follow the different states of functionality listed above and verify that you can still access the static embedding and public links when needed.

### Demo

_coming soon_

### Checklist

- [X] Tests have been updated in #36634 to prevent this PR from being polluted with a bunch of extra `cy.spec` files


frontend/src/metabase/public/components/widgets/SharingPane/icons/PublicEmbedIcon/PublicEmbedIcon.styled.tsx
```diff
@@ -0,0 +1,3 @@
+import styled from "@emotion/styled";
+
+export const PublicEmbedIconWrapper = styled.svg``;
```
oisincoveney, line 3: If there's a better way to ensure that the hover state works for both of these icons beyond what I've done, I would appreciate the feedback here. Not sure what else I can do

frontend/src/metabase/public/components/widgets/SharingPane/SharingPaneButton/SharingPaneButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const SharingPaneButton = ({
+  illustration,
+  children,
+  header,
+  description,
+  disabled,
+}: SharingOptionProps) => {
+  return (
+    <SharingPaneButtonContent disabled={disabled} withBorder>
+      <Center h="22.5rem" p="8rem">
```
oisincoveney, line 25: These values are from the design, maybe we can make them more responsive since the modal overflows on smaller screens

WiNloSt:
One thing I found a bit jarring is that when the public embed is created. The whole copy shifts a bit upward so the overall content is centered in the modal. That's the case in the design. This is minor, but I think it would feel a lot more polished if we keep the public embed aligned to the static embed copy.

#### design (content doesn't shift)
Screenshot 2023-12-13 at 12 31 26 PM

#### This PR before creating public embed link
Screenshot 2023-12-13 at 12 32 06 PM

#### This PR after creating public embed link
Screenshot 2023-12-13 at 12 31 43 PM

WiNloSt:
Clicking on the static embed header brings you back to the previous embed modal, but that does nothing on this PR.
I think we should either
1. Add the same functionality back or
2. Remove the hover effect on the header to reduce confusion because users would have thought it's clickable.

1. Screenshot 2023-12-13 at 1 04 45 PM
1. Screenshot 2023-12-13 at 1 04 50 PM

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.unit.spec.tsx
```diff
@@ -0,0 +1,194 @@
...
+          ),
+        ).toBeInTheDocument();
+
+        expect(screen.getByTestId("public-link-text")).toBeInTheDocument();
+        expect(screen.getByTestId("public-link-text")).toHaveTextContent(
+          /<iframe src="mock-uuid".*<\/iframe>/i,
+        );
+
+        expect(screen.getByTestId("copy-button")).toBeInTheDocument();
+        expect(screen.getByText("Remove public URL")).toBeInTheDocument();
```
WiNloSt, line 156: should we also test that `"Affects both embed URL and public link for this dashboard"` is rendered when hover over the remove button?

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.unit.spec.tsx
```diff
@@ -0,0 +1,194 @@
...
+      });
+
+      it("should call `onDisablePublicLink` when `Remove public URL` is clicked", () => {
```
WiNloSt, line 159: Should it be?
```suggestion
      it("should call `onDeletePublicLink` when `Remove public URL` is clicked", () => {
```

frontend/src/metabase/public/components/widgets/SharingPane/icons/StaticEmbedIcon/StaticEmbedIcon.styled.tsx
```diff
@@ -0,0 +1,3 @@
+import styled from "@emotion/styled";
+
+export const StaticEmbedIconWrapper = styled.svg``;
```
WiNloSt, line 3: You don't need a new wrapper in this case. Just inline `<svg />` instead.

frontend/src/metabase/public/components/widgets/SharingPane/SharingPaneButton/SharingPaneButton.styled.tsx
```diff
@@ -0,0 +1,82 @@
...
+        .innerFill {
+          stroke: ${theme.colors.text[2]};
+        }
+      }
+    `}
+
+  ${({ disabled, theme }) =>
+    !disabled &&
+    css`
+      &:hover {
```
WiNloSt, line 47: Maybe you could check with our product or designer, but I think the button should only be blue when hovering on the button itself, not on the card. I find it feels a little bit weird when the button turned blue even when I'm not hovering on it.

frontend/src/metabase/public/components/widgets/SharingPane/icons/PublicEmbedIcon/PublicEmbedIcon.styled.tsx
WiNloSt: You can inline `svg` where you use `PublicEmbedIconWrapper` and make `PublicEmbedIcon` accepts `className` then pass it to the `svg` tag.

WiNloSt (commented):
The overall shape looks fine, but I'd like some feedback addressed before merging.

frontend/src/metabase/public/components/widgets/EmbedModalContent.jsx
```diff
@@ -111,69 +111,63 @@ class EmbedModalContent extends Component {
...
-            />
-          </div>
-        ) : null}
+    return embedType == null ? (
+      <SharingPane
+        {...this.props}
+        onChangeEmbedType={embedType => setEmbedType(embedType)}
+      />
+    ) : embedType === "application" ? (
+      <div className="flex flex-full" style={{ height: "100%" }}>
```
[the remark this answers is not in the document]
oisincoveney: Ah yeah this was just copied from the previous thing - I'd like to adjust this for milestone 5 since I will definitely need to touch this

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const createPublicLink = () => {
+    MetabaseAnalytics.trackStructEvent(
+      "Sharing Modal",
+      "Public Link Enabled",
+      resourceType,
+    );
+    onCreatePublicLink();
+  };
+
+  const disablePublicLink = () => {
```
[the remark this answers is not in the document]
oisincoveney: Yeah, these should be merged from the previous PR

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.unit.spec.tsx
oisincoveney: yep

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.unit.spec.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const { history } = renderWithProviders(
+    <Route
+      path="*"
+      component={() => (
+        <SharingPane
+          resource={TEST_RESOURCE}
+          resourceType="dashboard"
+          onCreatePublicLink={onCreatePublicLink}
+          onDisablePublicLink={onDisablePublicLink}
+          extensions={[]}
```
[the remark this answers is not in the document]
oisincoveney: Turns out we don't need this now

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.unit.spec.tsx
oisincoveney: added

frontend/src/metabase/public/components/widgets/SharingPane/icons/PublicEmbedIcon/PublicEmbedIcon.styled.tsx
oisincoveney: Sweet, let me know if I did what you were thinking

vbenedetti:
> One thing I found a bit jarring is that when the public embed is created. The whole copy shifts a bit upward so the overall content is centered in the modal. That's the case in the design. This is minor, but I think it would feel a lot more polished if we keep the public embed aligned to the static embed copy.
> 
> #### design (content doesn't shift)
> Screenshot 2023-12-13 at 12 31 26 PM
> 
> #### This PR before creating public embed link
> Screenshot 2023-12-13 at 12 32 06 PM
> 
> #### This PR after creating public embed link
> Screenshot 2023-12-13 at 12 31 43 PM

This is a design bug somewhere. The elements should not shift up.

frontend/src/metabase/public/components/widgets/SharingPane/SharingPaneButton/SharingPaneButton.styled.tsx
WiNloSt: @oisincoveney I just looked at [the prototype](https://www.figma.com/proto/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=129-2053&t=3Tb9SQejXymI8wCi-0&scaling=min-zoom&page-id=129%3A1872&starting-point-node-id=129%3A2050&show-proto-sidebar=1), when hovering over the whole card the cursor would change to a click icon, that hints the design meant the entire card would be clickable, not just the button. If that's the case, then it makes sense for the button background to change when hovering over the card.

frontend/src/metabase/public/components/widgets/SharingPane/SharingPaneButton/SharingPaneButton.styled.tsx
```diff
@@ -0,0 +1,63 @@
...
+import type { PaperProps, ButtonProps } from "metabase/ui";
+import { Button, Paper, Title } from "metabase/ui";
+import { StaticEmbedIcon } from "../icons";
+
+export const SharingPaneButtonTitle = styled(Title)``;
+export const SharingPaneActionButton = styled(Button)<
+  ButtonProps & HTMLAttributes<HTMLButtonElement>
+>``;
+
+export const StaticEmbedIconWrapper = styled(StaticEmbedIcon)``;
```
WiNloSt, line 13: I've prototyped dealing with empty styled tags in this commit 35f9d79b464f9847e5c4dcc09cb1eba1b3d03000.

Basically, I inverse the CSS selector by putting the container CSS selector in the child component instead of the other way around.

The pros in my opinion would be we group related styles in a separate place, so there's no one big file that contains many individual children components with overridden styles.

Please take a look and if you think the approach looks good you can apply that to other empty-styled tags.

frontend/src/metabase/public/components/widgets/SharingPane/icons/PublicEmbedIcon/PublicEmbedIcon.styled.tsx
WiNloSt, line 1: I moved the style and hover style to the child component, so we don't have to have emtpy styled tags.

frontend/src/metabase/public/components/widgets/SharingPane/SharingPane.tsx
```diff
@@ -0,0 +1,124 @@
...
+          {resource.enable_embedding ? t`Edit settings` : t`Set this up`}
+        </SharingPaneActionButton>
+      </SharingPaneButton>
+
+      <SharingPaneButton
+        header={t`Public embed`}
+        disabled={!isPublicSharingEnabled}
+        description={
+          isPublicSharingEnabled ? (
+            hasPublicLink ? (
```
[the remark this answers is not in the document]
oisincoveney: Yeah, that's fair, I'll clean that up

frontend/src/metabase/public/components/widgets/SharingPane/icons/StaticEmbedIcon/StaticEmbedIcon.tsx
WiNloSt, line 1: Please take a look at [this comment](https://github.com/metabase/metabase/pull/36519#discussion_r1427919587).

WiNloSt (commented):
Almost ready to merge, I'd clean up the empty tags first. Let me know what you think about that.

frontend/src/metabase/public/components/widgets/SharingPane/SharingPaneButton/SharingPaneButton.styled.tsx
oisincoveney: Ah ok I see what you did. My only reservation about doing this is that I like having all of the hover states for a container to be in one place - for example, with the previous approach, I could go to `SharingPaneButtonContent` and see that the button would change colors along with everything else. But I'll take this over the previous approach still

WiNloSt: Fair enough, I couldn't find a way to do that without leaving empty tags. If you could do that it's fine. But I don't think having empty styled tags is a good idea. Looks like we're just hacking it together.

oisincoveney: I agree, I hated having them there as well which is why at the end of the day I think your approach is still better, but it is what it is

frontend/src/metabase/public/components/widgets/SharingPane/icons/StaticEmbedIcon/StaticEmbedIcon.tsx
```diff
@@ -0,0 +1,45 @@
+import { StaticEmbedIconRoot } from "metabase/public/components/widgets/SharingPane/icons/StaticEmbedIcon/StaticEmbedIcon.styled";
+
+export const StaticEmbedIcon = ({ className }: { className?: string }) => (
```
WiNloSt, line 3: You can remove `className` here (and line 5), since we're now styling `StaticEmbedIconRoot` directly.

WiNloSt (approved):
Great work! Looks good to me. Although I haven't tested the last commit, since I need to nuke my database again, but I assume you already make sure the style would work the same.

PR #36521, opened 2023-12-07
Add embedding types to Dashboards and Cards

oisincoveney:
For https://github.com/metabase/metabase/issues/35961

This just adds `enable_embedding` for the card and dashboard types, which is reflected in the API response that we get when we are accessing a card or dashboard already. For example:

_**Card**_
image

_**Dashboard**_

image

So this is just updating the types to match what we're already getting


npretto (approved):
Maybe worth adding a jsdoc to it since the name is not immediately clear, but lgtm

PR #36527, opened 2023-12-07
[#35961 M2] Split Link Panel from Popover

oisincoveney:
Epic #35961

### Description

This splits the copy/link display/remove button into its own component so I can use it in the embed modal for the iframe link. In terms of tests and behavior, absolutely nothing should change.


frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
```diff
@@ -52,70 +46,24 @@ export const PublicLinkPopover = ({
...
-          )}
+          <PublicLinkCopyPanel
+            loading={loading}
+            url={url}
+            onRemoveLink={isAdmin ? onRemoveLink : undefined}
+            extensions={extensions}
+            selectedExtension={selectedExtension}
+            onChangeExtension={setSelectedExtension}
+            removeButtonLabel={t`Remove public link`}
+            removeTooltipLabel={t`Affects both public link and embed URL for this dashboard`}
```
WiNloSt, line 65: Do you think these labels would be dynamic? Or put it another way, would `PublicLinkCopyPanel` be used somewhere else that will have different labels? Otherwise, I think hard coding them in `PublicLinkCopyPanel` for simplicity would make more sense. What do you think?

WiNloSt (approved):

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -83,7 +83,7 @@ const { PRODUCTS_ID } = SAMPLE_DATABASE;
             cy.findByTestId("public-link-popover-content").within(() => {
               cy.findByText("Public link").should("be.visible");
               cy.findByTestId("public-link-text").should("be.visible");
-              cy.findByText("Remove this public link").should("be.visible");
+              cy.findByText("Remove public link").should("be.visible");
```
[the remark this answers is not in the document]
oisincoveney: Yeah this is the [updated design](https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=438-3139&mode=design&t=BzOdawBOu6HJd1Kc-4) when we redid the tooltip

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
```diff
@@ -52,70 +46,24 @@ export const PublicLinkPopover = ({
   };
 
   return (
-    <Popover opened={isOpen} onClose={onClose}>
+    <Popover opened={isOpen} onClose={onClose} position="bottom-end">
```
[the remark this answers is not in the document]
oisincoveney: Hm.. that shouldn't be there. I think that was a merge conflict issue. The popover should be centered over the icon

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkPopover.tsx
oisincoveney: I guess the only issue here is that we don't have the resource available to tell which text we should provide - when we were just passing the dashboard/question it would have been a lot easier, but with just passing the URL that becomes trickier. We could check the substring for "question" or "dashboard" but I think that will just make things more complicated than they need to be

WiNloSt: yeah good point, I missed that. because currently they look like hard-coded strings.

PR #36455, opened 2023-12-06
Convert ExpressionStep to MLv2

kamilmielnik:
Closes #35982

Depends on #36533 (blocked by #36507)

### Description

One of e2e tests was failing and was skipped out because we need product input for #36574.

### How to verify

Add/edit/remove custom columns in GUI editor.

image

PR #36533, opened 2023-12-07
Port #36507 to master

metamben:
#36507


test/metabase/lib/util_test.cljc
```diff
@@ -338,11 +338,11 @@
         (lib/with-temporal-bucket :day-of-week))
     :type/Temporal))
 
-(deftest ^:parallel named-expression-clause-do-not-wrap-values-test
+(deftest ^:parallel top-level-expression-clause-do-not-wrap-values-test
   (testing "named-expression-clause should not wrap a :value clause in another :value clause"
```
kamilmielnik, line 342:
```suggestion
  (testing "top-level-expression-clause should not wrap a :value clause in another :value clause"
```

src/metabase/lib/convert.cljc
```diff
@@ -175,7 +175,7 @@
                           (mapv (fn [[k v]]
                                   (-> v
                                       ->pMBQL
-                                      (lib.util/named-expression-clause k))))
+                                      (lib.util/top-level-expression-clause k))))
```
kamilmielnik, line 178: Unfortunately this PR breaks `Lib.withExpressionName` - it has no effect on custom columns.
It works in `master`.
It causes [FE unit tests to fail](https://github.com/metabase/metabase/actions/runs/7140501215/job/19445925676?pr=36455#step:6:1701) in #36455.

kamilmielnik (changes requested):

kamilmielnik:
I merged `master` in and fixed conflicts in `src/metabase/lib/util.cljc`

kamilmielnik (approved):

PR #36535, opened 2023-12-07
understandable error when export target directory is not writable

### LINKED ISSUE #36468: Export fails with a confusing error message when the target directory is not writable
### Describe the bug

When attempting to export to a directory where the process doesn't have write permissions, the error shown is not helpful. 
We should improve it, especially since non-writable volumes are a frequent issue in Docker environments.

### To Reproduce

1. `mkdir /tmp/nowritabledirectory`
2. `chmod 555 /tmp/nowritabledirectory`
3. `java -jar metabase.jar export /tmp/nowritabledirectory`
4. (!) The export will fail with an error the does not clearly indicate the problem:
```
Command failed with exception: /tmp/nonwritabledirectory/databases/Sample Database/schemas/PUBLIC/tables/PRODUCTS/fields/CATEGORY.yaml (No such file or directory)
```

### Expected behavior

It should catch and display an error about a failed directory creation instead.

### Logs

```
[...]
2023-12-06 14:06:14,761 INFO models.serialization :: Extracting Field 71
2023-12-06 14:06:14,763 INFO v2.storage :: Storing Database Sample Database > Schema PUBLIC > Table PRODUCTS > Field CATEGORY
java.io.FileNotFoundException: /tmp/nonwritabledirectory/databases/Sample Database/schemas/PUBLIC/tables/PRODUCTS/fields/CATEGORY.yaml (No such file or directory)
        at java.base/java.io.FileOutputStream.open0(Native Method)
        at java.base/java.io.FileOutputStream.open(FileOutputStream.java:295)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:236)
[...]
        at metabase_enterprise.serialization.dump$spit_yaml.invokeStatic(dump.clj:31)
        at metabase_enterprise.serialization.dump$spit_yaml.invoke(dump.clj:27)
        at metabase_enterprise.serialization.v2.storage$store_entity_BANG_.invokeStatic(storage.clj:28)
        at metabase_enterprise.serialization.v2.storage$store_entity_BANG_.invoke(storage.clj:26)
        at metabase_enterprise.serialization.v2.storage$store_BANG_.invokeStatic(storage.clj:45)
        at metabase_enterprise.serialization.v2.storage$store_BANG_.invoke(storage.clj:37)
        at metabase_enterprise.serialization.cmd$v2_dump$fn__117051.invoke(cmd.clj:214)
        at metabase_enterprise.serialization.cmd$v2_dump.invokeStatic(cmd.clj:210)
        at metabase_enterprise.serialization.cmd$v2_dump.invoke(cmd.clj:203)
[...]
```

```JSON
v1.47.9
```

### Severity

P3 - we run into this only once so far

### Additional context

_No response_

sansolovyov:
Resolves #36468

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
@@ -1,6 +1,7 @@
 (ns metabase-enterprise.serialization.cmd
   (:refer-clojure :exclude [load])
   (:require
+   [clojure.java.io :as io]
    [metabase-enterprise.serialization.dump :as dump]
    [metabase-enterprise.serialization.load :as load]
    [metabase-enterprise.serialization.v2.entity-ids :as v2.entity-ids]
@@ -206,6 +207,10 @@
   (mdb/setup-db!)
   (check-premium-token!)
   (t2/select User) ;; TODO -- why??? [editor's note: this comment originally from Cam]
+  (let [f (io/file path)]
+    (.mkdirs f)
+    (when-not (.canWrite f)
+      (throw (ex-info (format "Destination path is not writeable: %s" path) {:filename path}))))
   (serdes/with-cache
     (-> (cond-> opts
           (seq collection-ids) (assoc :targets (v2.extract/make-targets-of-type "Collection" collection-ids)))
--- a/enterprise/backend/src/metabase_enterprise/serialization/dump.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/dump.clj
@@ -18,6 +18,7 @@
    [metabase.models.setting :as setting]
    [metabase.models.table :refer [Table]]
    [metabase.models.user :refer [User]]
+   [metabase.util.files :as u.files]
    [metabase.util.i18n :as i18n :refer [trs]]
    [metabase.util.log :as log]
    [metabase.util.yaml :as yaml]
@@ -61,16 +62,11 @@
 
   Writes (even nested) yaml keys in a deterministic fashion."
   [filename obj]
-  (cond
-    (io/make-parents filename)
+  (if (u.files/make-writable-parents! filename)
     (spit filename (yaml/generate-string (serialization-deep-sort obj)
                                          {:dumper-options {:flow-style :block :split-lines false}}))
 
-    (not (.canWrite (io/file filename)))
-    (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename}))
-
-    :else
-    (throw (ex-info (format "Unknown error writing to file: %s" filename) {:filename filename}))))
+    (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename}))))
 
 (defn- as-file?
   [instance]
--- a/enterprise/backend/src/metabase_enterprise/serialization/dump.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/dump.clj
@@ -61,9 +61,16 @@
 
   Writes (even nested) yaml keys in a deterministic fashion."
   [filename obj]
-  (io/make-parents filename)
-  (spit filename (yaml/generate-string (serialization-deep-sort obj)
-                                       {:dumper-options {:flow-style :block :split-lines false}})))
+  (cond
+    (io/make-parents filename)
+    (spit filename (yaml/generate-string (serialization-deep-sort obj)
+                                         {:dumper-options {:flow-style :block :split-lines false}}))
+
+    (not (.canWrite (io/file filename)))
+    (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename}))
+
+    :else
+    (throw (ex-info (format "Unknown error writing to file: %s" filename) {:filename filename}))))
 
 (defn- as-file?
   [instance]
--- a/enterprise/backend/src/metabase_enterprise/serialization/dump.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/dump.clj
@@ -18,7 +18,6 @@
    [metabase.models.setting :as setting]
    [metabase.models.table :refer [Table]]
    [metabase.models.user :refer [User]]
-   [metabase.util.files :as u.files]
    [metabase.util.i18n :as i18n :refer [trs]]
    [metabase.util.log :as log]
    [metabase.util.yaml :as yaml]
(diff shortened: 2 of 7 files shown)


levex (approved):

calherries (approved):

enterprise/backend/test/metabase_enterprise/serialization/v2/storage_test.clj
```diff
@@ -205,3 +205,13 @@
                                    (descend coll [:settings])
                                    (descend coll)))))]
           (storage/store! export "/non-existent"))))))
+
```
calherries, line 208:
```suggestion
```

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -61,9 +61,16 @@
...
-  (io/make-parents filename)
-  (spit filename (yaml/generate-string (serialization-deep-sort obj)
-                                       {:dumper-options {:flow-style :block :split-lines false}})))
+  (cond
+    (io/make-parents filename)
+    (spit filename (yaml/generate-string (serialization-deep-sort obj)
+                                         {:dumper-options {:flow-style :block :split-lines false}}))
+
+    (not (.canWrite (io/file filename)))
+    (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename}))
```
qnkhuat, line 70: do we want to translate this?

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -61,9 +61,16 @@
 
   Writes (even nested) yaml keys in a deterministic fashion."
   [filename obj]
-  (io/make-parents filename)
-  (spit filename (yaml/generate-string (serialization-deep-sort obj)
-                                       {:dumper-options {:flow-style :block :split-lines false}})))
+  (cond
+    (io/make-parents filename)
```
calherries, line 65: Ca you explain why this is a test expression? I would expect `(not (.canWrite (io/file filename)))` to come first, and otherwise we run these two lines

calherries (changes requested):
Looks like you have a real test failure. 
https://github.com/metabase/metabase/actions/runs/7131583530/job/19420426010?pr=36535#step:7:76

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: well `make-parents` promised to return `true` if it were able to create dirs and `false` otherwise. Turns out it returns `false` even if the directory exists already. 😁

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: You mean so it's easier to consume for users?... Hmmm, maybe this is a good case, yeah.

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
calherries: Where do you see that promise documented?

This is the docstring for `make-parents`:
```
  "Given the same arg(s) as for file, creates all parent directories of
   the file they represent."
```

I'd rather not rely on undocumented behaviour if we can avoid it

piranha:
```public boolean mkdirs()
Creates the directory named by this abstract pathname, including any necessary but nonexistent parent directories. Note that if this operation fails it may have succeeded in creating some of the necessary parent directories.
Returns:
true if and only if the directory was created, along with all necessary parent directories; false otherwise```

File: src/metabase/util/files.clj [hunk 4]
```diff
@@ -78,6 +78,13 @@
   [^Path path]
   (iterator-seq (.iterator (Files/list path))))
 
+(defn make-writable-parents!
```
calherries, line 81: The implementation of this function has a strange edge case. `(make-writable-parents! ".") returns `nil`. And even though this can never happen in the serialization code, it's very strange undocumented behaviour that could cause bugs in the future.

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -61,9 +62,11 @@
 
   Writes (even nested) yaml keys in a deterministic fashion."
   [filename obj]
-  (io/make-parents filename)
-  (spit filename (yaml/generate-string (serialization-deep-sort obj)
-                                       {:dumper-options {:flow-style :block :split-lines false}})))
+  (if (u.files/make-writable-parents! filename)
```
calherries, line 65: Rather than make a new untested util function with at least one weird edge case, I think it would be safer to just inline the check in this function body, like this
```
  (when (not (.canWrite (.getParentFile (io/file filename))))
    (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename})))
```

calherries (changes requested):

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
piranha: sure, why not

enterprise/backend/test/metabase_enterprise/serialization/test_util.clj
```diff
@@ -102,7 +103,7 @@
   `(~'&do-with-dest-db (fn [] ~@body)))
 
 (defn random-dump-dir [prefix]
-  (str (System/getProperty "java.io.tmpdir") "/" prefix (mt/random-name)))
+  (str (u.files/get-path (System/getProperty "java.io.tmpdir") prefix (mt/random-name))))
```
piranha, line 106: Simple `str` generated double slashes on my system.

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
```diff
@@ -206,6 +207,10 @@
   (mdb/setup-db!)
   (check-premium-token!)
   (t2/select User) ;; TODO -- why??? [editor's note: this comment originally from Cam]
+  (let [f (io/file path)]
+    (.mkdirs f)
+    (when-not (.canWrite f)
+      (throw (ex-info (format "Destination path is not writeable: %s" path) {:filename path}))))
```
piranha, line 213: This is just so that user doesn't need to wait until the extraction process started.

enterprise/backend/src/metabase_enterprise/serialization/dump.clj
```diff
@@ -62,11 +61,14 @@
...
-  (if (u.files/make-writable-parents! filename)
+  (io/make-parents filename)
+  (try
     (spit filename (yaml/generate-string (serialization-deep-sort obj)
                                          {:dumper-options {:flow-style :block :split-lines false}}))
-
-    (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename}))))
+    (catch Exception e
+      (if-not (.canWrite (.getParentFile (io/file filename)))
+        (throw (ex-info (format "Destination path is not writeable: %s" filename) {:filename filename}))
```
calherries, line 70: This code path is untested and in practice seems really unlikely to be hit. The folder they're extracting to would have to be writable but the folders inside it like `collections` would have to be not writeable. Do you have a reason to keep this behaviour or can we remove it? Otherwise it should be tested

calherries: On second thought, I guess it's possible and it's nice to have this otherwise it would be coupled to v2-dump!

calherries (approved):
Looks good. Thanks for putting up with my pedantic comments

piranha:
@metadata-bot backport release-x.47.x

piranha:

piranha:

PR #36543, opened 2023-12-07
adding perm-graph filtering on db or group id

### LINKED ISSUE #36647: Faster permission graph endpoints

escherize:
resolves #36647.

This PR adds 2 endpoints that will make our permission graph api a lot snappier when graphs get large.

- GET `api/permissions/graph/db/:db-id`
- GET `api/permissions/graph/group/:group-id`

These return the same graph structure, but filtered such that they return only {db,group}-ids respectively. 

Since this is a slice of the entire graph, we query for, post-process, and transmit to the client, much less data.


src/metabase/models/permissions.clj
```diff
@@ -861,6 +861,15 @@
        post-process-graph
        add-impersonations-to-permissions-graph))
 
+(defn ->v1-paths
+  "keep v1 paths, implicitly remove v2"
+  [pbpi]
```
escherize, line 866: todo: change name

test/metabase/api/permissions_test.clj
```diff
@@ -156,6 +156,24 @@
...
       (mt/user-http-request :rasta :get 403 "permissions/graph"))))
 
+(deftest fetch-perms-graph-by-group-id-test
+  (testing "GET /api/permissions/graph"
+    (testing "make sure we can fetch the perms graph from the API"
+      (t2.with-temp/with-temp [PermissionsGroup {group-id :id :as group}    {}
+                               Database         db                          {}]
+        (perms/grant-permissions! group (perms/data-perms-path db))
+        (let [graph (mt/user-http-request :crowberto :get 200 (format "permissions/graph/group/%s" group-id))]
+          (is (= #{group-id} (set (keys graph)))))))))
```
noahmoss, line 166: Maybe slightly better to assert on the shape of the entire result graph, to make sure that the queries to fetch the perms are working as intended? Same for the other tests.

noahmoss (approved):

test/metabase/api/permissions_test.clj
escherize: Good point: We have the schemas laying around anyway!

test/metabase/models/permissions_group_test.clj
```diff
@@ -105,3 +105,17 @@
...
+
+(deftest data-graph-for-group-check-all-groups-test
+  (doseq [group-id (t2/select-fn-set :id :model/PermissionsGroup)]
+    (testing (str "testing data perms graph for group graph with group-id: [" group-id "].")
+      (is (= #{group-id} (set (keys (perms/data-graph-for-group group-id))))))))
+
+(deftest data-graph-for-db-check-all-dbs-test
+  (doseq [db-id (t2/select-fn-set :id :model/Database)]
+    (testing (str "testing data perms graph for db graph with db-id: [" db-id "].")
+      (= #{db-id}
```
escherize, line 117: missing `is`

PR #35929, opened 2023-11-17
Upload test results to S3

iethree:
Closes https://github.com/metabase/metabase/issues/35682

### Description

Uploads junit test results to an s3 bucket for all tests we run. This will allow us to detect failures, flakes, and run time on a per-test basis for our entire test suite, which we cannot currently do.

The four groups of tests we have are
- frontend unit tests
- backend unit tests
- driver tests
- e2e tests

All of which now output junit test files which get zipped within each job and uploaded to s3 with this path structure:
- format: `s3://{bucket}/{date}/{runId}/{runAttempt}/{jobName}.zip`
- example: `s3://my-test-bucket/2023-11-30/7049572755/1/be-tests-bigquery-cloud-sdk-ee.zip`

## New Environment Variables Needed

✅  Added by Nemanja on 11/28

```sh
# secret (secrets)
AWS_TEST_RESULTS_ACCESS_KEY_ID
AWS_TEST_RESULTS_SECRET_ACCESS_KEY

# not secret (vars)
AWS_S3_TEST_RESULTS_BUCKET
```

PR #36553, opened 2023-12-07
Perf testing via cypress-audit [POC]

kdoh:
Looking into using [Cypress Audit](https://mfrachet.github.io/cypress-audit/) to start collecting some info about performance of some of our key flows, starting with dashboards. 

I have the basic round trip running and it's producing a stats report, but weirdly when I try and specify any thresholds in the test itself, it seems to always timeout, at least locally.

I think in terms of next steps, I'd want to
- [ ] Rebase once Ryan's [original stats collection PR](https://github.com/metabase/metabase/pull/35929) hits `master`
- [ ] Make sure tests don't fail if the thresholds aren't met, yet. We just want to establish a baseline at this point.
- [ ] Check with @iethree and @maxzheng that the report that's generated would be consumable by the work they're doing on CI stats
- [ ] Check with @nemanjaglumac about things I'm not thinking of that this might impact.

### DIFF
--- a/e2e/support/commands.js
+++ b/e2e/support/commands.js
@@ -34,3 +34,5 @@ import "./commands/overwrites/log";
 import "./commands/percy/createPercySnapshot";
 
 require("./commands/downloads/deleteDownloadsFolder").addCustomCommand();
+
+import "@cypress-audit/lighthouse/commands";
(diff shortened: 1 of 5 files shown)


kdoh:
Going to close this since based on a few internal conversations, I think we'll want to rethink this approach a little bit to run on a specific runner we have more control over.

Issue #36211, opened 2023-11-29
[MLv2] [BE] Support changing expressions in the query w/o removing parts depending on the original


metamben:
Currently, they way expressions (both simple and aggregates) can be renamed by calling `replace-clause`. The general behavior of `replace-clause` is to remove any parts of the query that depend on the clause being removed. This is a (reasonably) simple way to guarantee that MLv2 queries are valid at all times.

MLv1 supports replacing clauses without removing other parts of the query, but doesn't guarantee that queries so constructed are valid.

This issue is about enabling the renaming of expressions without removing dependent parts.
- This can be done by changing `replace-clause`, but this seems complicated, because replacing the whole expression opens up many ways to break the query. Also, if we do this, it makes sense to support this for other parts of the query too, not just for expressions.
  - *We decided to go this way.* The FE would use `replace-clause` and the BE makes a best effort attempt to keep as much of the query as it can.
    1.  Initially we will only support this in the trivial case when the name and the type of the expression don't change.
    2. Then we want to add support for renaming the expression when the type does not change.
    3. Finally, we want to support changing both the name and the type if it can be done without breaking the query.
- Alternatively, a special function can be implemented that allows only changing the name of expressions. This is easier than changing `replace-clause`, but isn't quite trivial either because changing the name may result in clashes on subsequent query stages.
- The third option is to leave it as it is. This means that renaming expressions would keep removing dependent query parts even if the renaming was harmless.
```[tasklist]
### Tasks
- [ ] #36558
- [ ] #37204
- [ ] #37409
```

PR #36558, opened 2023-12-07
Support changing the expression when the type and the name don't change

metamben:
Part of #36211.

In the code call making these trivial changes tweaking.

This PR introduces `(assignable? x y)` which is true if values of type `x` can be used in contexts where values of type `y` are expected. Without this, even adding an integer to a float would prevent keeping the dependent parts. Also, `(most-specific-common-ancestor :type/Integer :type/Float)` will be `:type/Float`. (`ancestor` might not be the best word for such semantics.)


src/metabase/lib/remove_replace.cljc
```diff
@@ -191,6 +193,48 @@
      (remove-join query stage-number target-clause)
      (remove-replace* query stage-number target-clause :remove nil))))
 
+(defn- expression-role
+  [an-expression]
+  (if (-> an-expression lib.options/options :lib/expression-name)
+    :custom-column
+    :custom-aggregation))
+
+(let [expression-validator (mc/validator :metabase.lib.schema.expression/expression)]
```
bshepherdson, line 202: `def`s not at the top level can make CLJS sad. Is there any reason not to make this a `(def ^:private expression-validator ...)` and let `tweak?` be at the top level?

src/metabase/types.cljc
```diff
@@ -345,14 +345,32 @@
...
+     :cljs (set! assignable-hierarchy (derive assignable-hierarchy x y))))
+
+(declare-assignable :type/Integer :type/Decimal)
+
+(defn assignable?
+  "Is a value of type `x` assignable to a variable of type `y`?"
+  [x y]
+  (or (isa? assignable-hierarchy x y)
+      (boolean (some #(assignable? x %) (descendants y)))
+      (boolean (some #(assignable? % y) (parents x)))))
```
bshepherdson, line 364: Should `descendants` and `parents` here also be linked to the `assignable-hierarchy`? If we're deliberately using the larger `metabase.types` hierarchy, a comment saying so would be worthwhile.

bshepherdson (approved):

src/metabase/lib/remove_replace.cljc
metamben: No serious reason, just that I didn't intend this validator to leak the function even. I'll change it. What problems does this cause in CLJS?

PR #36578, opened 2023-12-08
Workaround for an H2 bug

johnswanson:
There's a fun bug in H2: https://github.com/h2database/h2database/issues/3942

To reproduce:

- create a table, then
- create a view based on the table, then
- modify the original table, then
- generate a snapshot

The generated snapshot has the `CREATE VIEW` *before* the `CREATE TABLE`. This results in a view that can't be queried successfully until it is recompiled. Our workaround is to recompile ALL views immediately after we restore the app DB from a snapshot.

### DIFF
--- a/src/metabase/api/testing.clj
+++ b/src/metabase/api/testing.clj
@@ -64,7 +64,25 @@
     (doseq [sql-args [["SET LOCK_TIMEOUT 180000"]
                       ["DROP ALL OBJECTS"]
                       ["RUNSCRIPT FROM ?" snapshot-path]]]
-      (jdbc/execute! {:connection conn} sql-args)))
+      (jdbc/execute! {:connection conn} sql-args))
+
+    ;; We've found a delightful bug in H2 where if you:
+    ;; - create a table, then
+    ;; - create a view based on the table, then
+    ;; - modify the original table, then
+    ;; - generate a snapshot
+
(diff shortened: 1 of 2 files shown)


src/metabase/api/testing.clj
```diff
@@ -64,7 +64,25 @@
...
+    ;; the generated snapshot has the `CREATE VIEW` *before* the `CREATE TABLE`. This results in a view that can't be
+    ;; queried successfully until it is recompiled. Our workaround is to recompile ALL views immediately after we
+    ;; restore the app DB from a snapshot. Bug report is here: https://github.com/h2database/h2database/issues/3942
+    (when (not config/is-prod?)
+      (doseq [table-name
+              (->> (jdbc/query {:connection conn} ["SELECT table_name FROM information_schema.views WHERE table_schema=?" "PUBLIC"])
+                   (map :table_name))]
+        ;; parameterization doesn't work with view names. If someone maliciously named a table, this is bad. On the
+        ;; other hand, this is not running in prod and you already had to have enough access to maliciously name the
+        ;; table, so this is probably safe enough.
```
noahmoss, line 84: If you're worried about this, we do have a hard-coded list of audit view names in `metabase-enterprise.audit-app.permissions` that you could use instead.

But then again, if someone starts adding new views for a different use case in the future they might hit this issue again. So I might actually lean towards your current approach, but up to you.

noahmoss (approved):

PR #36595, opened 2023-12-08
Create an API key with `POST /api/api-key`

### LINKED ISSUE #36347: Write a `POST /api/api-key` endpoint to create an API key

johnswanson:
This creates a new endpoint, `POST /api/api-key` that takes two arguments, a `group_id` and a `name`, and creates an
API key (along with a special API key user).

Fixes #36347

### DIFF
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -1,0 +1,57 @@
+(ns metabase.api.api-key
+  "/api/api-key endpoints for CRUD management of API Keys"
+  (:require
+   [compojure.core :refer [POST GET]]
+   [crypto.random :as crypto-random]
+   [metabase.api.common :as api]
+   [metabase.models.api-key :as api-key]
+   [metabase.models.permissions-group :as perms-group]
+   [metabase.models.user :as user]
+   [metabase.util :as u]
+   [metabase.util.malli.schema :as ms]
+   [toucan2.core :as t2]))
+
+(defn- key-with-unique-prefix []
+  (u/auto-retry 5
+   (let [api-key (api-key/generate-key)
+         prefix (api-key/prefix api-key)]
+     ;; we could make this more efficient by generating 5 API keys up front and doing one select to remove any
+     ;; duplicates. But a duplicate should be rare enough to just do multiple queries for now.
+     (if-not (t2/exists? :model/ApiKey :key_prefix prefix)
+       api-key
+       (throw (ex-info "could not generate key with unique prefix" {}))))))
+
+(api/defendpoint POST "/"
+  "Create a new API key (and an associated `User`) with the provided name and group ID."
+  [:as {{:keys [group_id name] :as _body} :body}]
+  {group_id ms/PositiveInt
+   name     ms/NonBlankString}
+  (api/check-superuser)
+  (api/checkp (not (t2/exists? :model/ApiKey :name name))
+    "name" "An API key with this name already exists.")
+  (let [api-key (key-with-unique-prefix)
+        email   (format "api-key-user-%s@api-key.invalid" name)]
+    (t2/with-transaction [_conn]
+      (let [user (first (t2/insert-returning-instances! :model/User
+                                                        {:email    email
+                                                         :password (crypto-random/base64 16)
+                                                         :type     :api-key}))]
+        (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])
+        (u/prog1 (-> (t2/insert-returning-instances! :model/ApiKey
+                                                     {:user_id    (u/the-id user)
+                                                      :name       name
+                                                      :key        api-key
+                                                      :created_by api/*current-user-id*})
+                     (select-keys [:created_at :updated_at :id])
+                     (assoc :name name
+                            :group_id group_id
+                            :unmasked_key api-key
+                            :masked_key (api-key/mask api-key))))))))
+
+(api/defendpoint GET "/count"
+  "Get the count of API keys in the DB"
+  [:as _body]
+  (api/check-superuser)
+  (t2/count :model/ApiKey))
+
+(api/define-routes)
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -8,6 +8,7 @@
    [metabase.models.permissions-group :as perms-group]
    [metabase.models.user :as user]
    [metabase.util :as u]
+   [metabase.util.i18n :refer [tru]]
    [metabase.util.malli.schema :as ms]
    [toucan2.core :as t2]))
 
@@ -19,7 +20,7 @@
      ;; duplicates. But a duplicate should be rare enough to just do multiple queries for now.
      (if-not (t2/exists? :model/ApiKey :key_prefix prefix)
        api-key
-       (throw (ex-info "could not generate key with unique prefix" {}))))))
+       (throw (ex-info (tru "could not generate key with unique prefix") {}))))))
 
 (api/defendpoint POST "/"
   "Create a new API key (and an associated `User`) with the provided name and group ID."
@@ -37,16 +38,16 @@
                                                          :password (crypto-random/base64 16)
                                                          :type     :api-key}))]
         (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])
-        (u/prog1 (-> (t2/insert-returning-instances! :model/ApiKey
-                                                     {:user_id    (u/the-id user)
-                                                      :name       name
-                                                      :key        api-key
-                                                      :created_by api/*current-user-id*})
-                     (select-keys [:created_at :updated_at :id])
-                     (assoc :name name
-                            :group_id group_id
-                            :unmasked_key api-key
-                            :masked_key (api-key/mask api-key))))))))
+        (-> (t2/insert-returning-instances! :model/ApiKey
+                                            {:user_id      (u/the-id user)
+                                             :name         name
+                                             :unhashed_key api-key
+                                             :created_by   api/*current-user-id*})
+            (select-keys [:created_at :updated_at :id])
+            (assoc :name name
+                   :group_id group_id
+                   :unmasked_key api-key
+                   :masked_key (api-key/mask api-key)))))))
 
 (api/defendpoint GET "/count"
   "Get the count of API keys in the DB"
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -2,7 +2,6 @@
   "/api/api-key endpoints for CRUD management of API Keys"
   (:require
    [compojure.core :refer [POST GET]]
-   [crypto.random :as crypto-random]
    [metabase.api.common :as api]
    [metabase.models.api-key :as api-key]
    [metabase.models.permissions-group :as perms-group]
@@ -31,19 +30,17 @@
   (api/checkp (not (t2/exists? :model/ApiKey :name name))
     "name" "An API key with this name already exists.")
   (let [api-key (key-with-unique-prefix)
-        email   (format "api-key-user-%s@api-key.invalid" name)]
+        email   (format "api-key-user-%s@api-key.invalid" (u/slugify name))]
     (t2/with-transaction [_conn]
-      (let [user (first (t2/insert-returning-instances! :model/User
-                                                        {:email      email
-                                                         :first_name name
-                                                         :password   (crypto-random/base64 16)
-                                                         :type       :api-key}))]
-        (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])
-        (-> (t2/insert-returning-instances! :model/ApiKey
-                                            {:user_id      (u/the-id user)
-                                             :name         name
-                                             :unhashed_key api-key
-                                             :created_by   api/*current-user-id*})
+      (let [user (user/insert-new-user! {:email      email
+                                         :first_name name
+                                         :type       :api-key})]
+        (user/set-permissions-groups! user [(perms-group/all-users) group_id])
+        (-> (t2/insert-returning-instance! :model/ApiKey
+                                           {:user_id      (u/the-id user)
+                                            :name         name
+                                            :unhashed_key api-key
+                                            :created_by   api/*current-user-id*})
             (select-keys [:created_at :updated_at :id])
             (assoc :name name
                    :group_id group_id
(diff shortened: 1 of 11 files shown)


src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,75 @@
...
+   [metabase.util :as u]
+   [metabase.util.honey-sql-2 :as h2x]
+   [metabase.util.malli.schema :as ms]
+   [toucan2.core :as t2]))
+
+
+(def ^:private user-first-name "API Key:")
+
+(defn- full-name [n]
+  (str user-first-name " " n))
```
noahmoss, line 19: I don't really see why the comparison on full name is needed? The only thing we need to enforce is that API key names are unique.

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,75 @@
...
+        email   (format "api-key-user-%s@api-key.invalid" name)]
+    (api/check-superuser)
+    (api/checkp (not (t2/exists? :model/User :%lower.email email))
+      "name" "An API key with this name already exists.")
+    (t2/with-transaction [_conn]
+      (let [user (first (t2/insert-returning-instances! :model/User
+                                                        {:email      email
+                                                         :password   (crypto-random/base64 16)
+                                                         :first_name "API Key:"
+                                                         :last_name  name
```
noahmoss, line 53: I just realized something that may require a change of approach here: `API Key` should almost certainly be translated, but it won't be if we're storing it as a raw string in the DB. 

Could you change it to be a i18n'ed string stored in memory that is hydrated on the `User` model when an internal user is fetched? 

Another schema adjustment that might make sense is to store the API key name in the `api_key` table and hydrate the `last_name` field when internal users are fetched, rather than storing it directly in the `core_user` table. This would solve the potential race condition below, right? Unless I'm misunderstanding something there.

i.e. an `api-key` user would just be a shell and all data related to the API key would be stored in the `api_key` table, instead of split across that and the `core_user` table

src/metabase/models/api_key.clj
```diff
@@ -0,0 +1,52 @@
...
+(defn generate-key
+  "Generates a new API key - a random base64 string prefixed with `mb_`"
+  []
+  (str "mb_" (crypto-random/base64 bytes-key-length)))
+
+(t2/define-before-insert :model/ApiKey
+  [{:keys [key key_prefix] :as api-key}]
+  (assert (= key_prefix (prefix key)))
+  (merge api-key
+         (hashed-key api-key)))
```
noahmoss, line 44: nit: I usually assume a function name describes what it returns, so I thought that `hashed-key` returned the key directly when skimming through this. Not a big deal either way but my bias would be for it to take and return strings, then do something like 
```suggestion
  (update api-key :key hashed-key))
```

noahmoss (changes requested):
Looks pretty good overall, but I noticed an i18n issue which I think requires us to reevaluate some of the schema. Let me know if you have any questions about that (and sorry I didn't catch that earlier). Otherwise I just left some nits, but I'll do another pass whenever you're ready for a re-review

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,75 @@
...
+   [metabase.api.common :as api]
+   [metabase.models.api-key :as api-key]
+   [metabase.models.permissions-group :as perms-group]
+   [metabase.models.user :as user]
+   [metabase.util :as u]
+   [metabase.util.honey-sql-2 :as h2x]
+   [metabase.util.malli.schema :as ms]
+   [toucan2.core :as t2]))
+
+
```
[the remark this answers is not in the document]
johnswanson: Fixed!

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,75 @@
...
+(def ^:private user-first-name "API Key:")
+
+(defn- full-name [n]
```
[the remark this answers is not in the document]
johnswanson: Removed the whole function, but good call :)

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,75 @@
...
+  (str user-first-name " " n))
+
+(defn- count-api-key-users-with-name [n]
+  (:count
+   (t2/query-one {:select [[[:count :core_user.id] :count]]
+                  :from :core_user
+                  :where [:= (h2x/concat :first_name (h2x/literal " ") :last_name) (full-name n)]})))
```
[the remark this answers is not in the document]
johnswanson: Nice, I removed the function here but used this in the `GET /api-key/count` endpoint!

src/metabase/api/api_key.clj
johnswanson: I wanted to avoid the case where someone creates an API key with the same last name as a user and is blocked from doing so. But moot now since the refactor!

src/metabase/api/api_key.clj
johnswanson: I like this and implemented it! Thanks for the idea.

src/metabase/models/api_key.clj
johnswanson: I like this better too, good call!

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,57 @@
...
+
+(defn- key-with-unique-prefix []
+  (u/auto-retry 5
+   (let [api-key (api-key/generate-key)
+         prefix (api-key/prefix api-key)]
+     ;; we could make this more efficient by generating 5 API keys up front and doing one select to remove any
+     ;; duplicates. But a duplicate should be rare enough to just do multiple queries for now.
+     (if-not (t2/exists? :model/ApiKey :key_prefix prefix)
+       api-key
+       (throw (ex-info "could not generate key with unique prefix" {}))))))
```
noahmoss, line 22: If this message has a chance of being surfaced to end users it should be translated. (Granted, this one is extremely unlikely to be hit, but still...)

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,57 @@
...
+    "name" "An API key with this name already exists.")
+  (let [api-key (key-with-unique-prefix)
+        email   (format "api-key-user-%s@api-key.invalid" name)]
+    (t2/with-transaction [_conn]
+      (let [user (first (t2/insert-returning-instances! :model/User
+                                                        {:email    email
+                                                         :password (crypto-random/base64 16)
+                                                         :type     :api-key}))]
+        (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])
+        (u/prog1 (-> (t2/insert-returning-instances! :model/ApiKey
```
noahmoss, line 40: nit: don't think `u/prog1` is needed anymore

src/metabase/models/api_key.clj
```diff
@@ -0,0 +1,50 @@
...
+  [{:keys [key] :as api-key}]
+  (-> api-key
+      add-prefix
+      (update :key u.password/hash-bcrypt)))
+
+(t2/define-before-update :model/ApiKey
+  [{:keys [key] :as api-key}]
+  (-> api-key
+      add-prefix
+      (update :key u.password/hash-bcrypt)))
```
noahmoss, line 50: If the `key` stored in the DB is already a hashed key, why do we need to do that again on updates? Wouldn't that hash it again and it would become corrupted?

(I'm assuming that if the key name is updated or something, the FE would include the hashed value as the `:key` since it should never have access to the unhashed value after it's generated)

src/metabase/models/user.clj
```diff
@@ -188,13 +188,23 @@
            common-name)
       (assoc :common_name common-name))))
 
+(defn- maybe-get-name-from-api-key
+  "If the user is an API Key user, replace any existing name with the name from the API Key table."
+  [{:keys [type] :as user}]
+  (cond-> user
+    (= :api-key type)
+    (assoc :first_name (i18n/tru "API Key:")
```
noahmoss, line 196: nit: our convention is usually to refer translation macros. `tru` is already available in this namespace without the `i18n` alias
```suggestion
    (assoc :first_name (tru "API Key:")
```

src/metabase/models/user.clj
```diff
@@ -188,13 +188,23 @@
...
+           :last_name (:name (t2/select-one [:model/ApiKey :name] :user_id (:id user))))))
```
noahmoss, line 197: nit: marginally more concise (I think this is the syntax)
```suggestion
           :last_name (t2/select-one-fn :name :model/ApiKey :user_id (:id user)))))
```

noahmoss: also maybe worth adding some error handling here in case the `api-key` user doesn't have an associated `ApiKey` model? shouldn't happen in practice but I don't think we enforce that as a DB invariant

src/metabase/models/api_key.clj
johnswanson: I'm actually planning for the `regenerate-key` endpoint here. We'll want to update and set a new `key` on an existing API key, so I was imagining we would do something like `(t2/update! :model/ApiKey {:id ...} {:key (generate-a-new-key)})`.

I was not imagining that someone would do `(t2/update! :model/ApiKey {:id ...} an-api-key-selected-from-the-db)`. That seems undesirable to me, since you'll risk race conditions by possibly overwriting values that were set by other threads - much better to only explicitly update the particular fields you're interested in updating.

But if this is a concern, one option might be to stop overloading `:key` to mean "hashed key" *and* the unhashed key. You could only write to the `unhashed_key` field, and only read from the `hashed_key` field. Might be better anyway - WDYT?

noahmoss: Oh that's fair. But yeah it's a somewhat common pattern in the code base to read a model, change a single field, and then write it back, even if you're not planning on doing that directly here. I like the idea of not overloading the `:key` field to make it less error-prone

src/metabase/models/user.clj
johnswanson: I tried to do this and it ended up breaking API key creation, because we need to:
- insert a new `:api-key`-typed user, grabbing the ID
- insert a new ApiKey with that `user_id`
Adding error-handling here causes this to break, because `t2/insert-returning-pk!` or `t2/insert-returning-instance!` both run after-select hooks.

Maybe we could set the `:last_name` to `(tru "[Missing]")` or something along those lines? Not ideal, but I'm not sure of a better option...

noahmoss: Oh I see. I guess a workaround could be to avoid those functions when inserting the user and just use `t2/insert!`, and then `t2/query` to fetch the ID. But I don't think this error handling is _vital_, it's just nice to have.

cc @camsaul if there's a better t2 pattern to use in this sort of case; I don't really know the internals.

src/metabase/models/api_key.clj
```diff
@@ -0,0 +1,57 @@
...
+  (derive :hook/timestamped?))
+
+(defn prefix
+  "Given an API key, returns the standardized prefix for that API key."
+  [key]
+  (apply str (take prefix-length key)))
+
+(defn- add-prefix [{:keys [unhashed_key] :as api-key}]
+  (cond-> api-key
+    key (assoc :key_prefix (prefix unhashed_key))))
```
noahmoss, line 26: I'm confused by this `cond->`. `key` is the condition, but it's a function so it'll always be truthy. Can you just use a simple `assoc` here without the `cond->`?

src/metabase/models/user.clj
noahmoss: Still need to be updated?

test/metabase/api/user_test.clj
```diff
@@ -66,9 +66,10 @@
         (let [result (->> ((mt/user-http-request :crowberto :get 200 "user") :data)
                           (filter mt/test-user?))]
           ;; since this is an admin, all keys are available on each user
-          (is (= (set (concat
-                       user/admin-or-self-visible-columns
-                       [:common_name :group_ids :personal_collection_id]))
+          (is (= (disj (set (concat
+                             user/admin-or-self-visible-columns
+                             [:common_name :group_ids :personal_collection_id]))
+                       :type)
```
noahmoss, line 72: Would it be better to just exclude `:type` from `default-user-columns` if we're not typically going to return it when fetching users? removing `:type` manually here feels a little error-prone

noahmoss (approved):
Looks good overall, just a couple more nits

src/metabase/api/api_key.clj
```diff
@@ -0,0 +1,58 @@
...
+  (api/check-superuser)
+  (api/checkp (not (t2/exists? :model/ApiKey :name name))
+    "name" "An API key with this name already exists.")
+  (let [api-key (key-with-unique-prefix)
+        email   (format "api-key-user-%s@api-key.invalid" name)]
+    (t2/with-transaction [_conn]
+      (let [user (first (t2/insert-returning-instances! :model/User
+                                                        {:email    email
+                                                         :password (crypto-random/base64 16)
+                                                         :type     :api-key}))]
```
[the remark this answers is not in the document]
johnswanson: I saw that and was tempted, but the function is private and I wasn't sure whether it was worth making it public for this. WDYT?

src/metabase/models/api_key.clj
johnswanson: Oops, this was a mistake from a rename. The condition should have been `unhashed_key` - we only want to attach a `key_prefix` if someone is trying to set the `unhashed_key`.

test/metabase/api/user_test.clj
johnswanson: We needed to have it in `default-user-columns` so that the after-select hook would work - but now that we don't need to do any special massaging of the user's name, we can remove this.

src/metabase/api/api_key.clj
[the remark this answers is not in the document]
noahmoss: Oh you're right. I think that one would be fine to make public though.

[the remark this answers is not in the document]
noahmoss: Lots of private functions, especially in model namespaces, are just private by default since they're not used outside of the namespace. But unless it's a namespace with a clearly marked public API, lots of functions are fine to make public if it's useful. This case definitely seems pretty benign to me.

test/metabase/api/api_key_test.clj
```diff
@@ -0,0 +1,104 @@
...
+  (:require
+   [clojure.test :refer [deftest testing is]]
+   [metabase.models.api-key :as api-key]
+   [metabase.models.permissions-group :as perms-group]
+   [metabase.test :as mt]
+   [toucan2.tools.with-temp :as t2.with-temp]))
+
+(deftest api-key-creation-test
+  (t2.with-temp/with-temp
+      [:model/PermissionsGroup {group-id :id} {:name "Cool Friends"}]
```
noahmoss, line 12: nit: think this should be 2 spaces, not 4? same with the instances below.

noahmoss (approved):

PR #36600, opened 2023-12-08
docs - csv upload id name

jeff-bruemmer:
Just a mention of the name Metabase uses when adding an ID to an uploaded CSV.

See https://github.com/metabase/metabase/pull/36361.

@calherries Anything else special to mention here?


docs/databases/uploads.md
```diff
@@ -47,7 +47,7 @@ When people upload a CSV to a collection, Metabase will:
 
 When you upload data, Metabase will try to detect whether the uploaded CSV file includes a column with an incrementing ID. The check is pretty basic: it'll check for a column named `id` or `pk` that has integers for each row in the uploaded CSV. Each integer in this column should be unique, so if your CSV includes duplicate ID integers, the column won't work as [an entity key](../data-modeling/field-types.md#list-of-metabase-field-types).
 
-If Metabase doesn't detect a primary key, it will create an ID column for you, including this ID column as the first (left-most) column of the uploaded CSV table.
+If Metabase doesn't detect a primary key, it will create an ID column for you, including this ID column as the first (left-most) column of the uploaded CSV table: `_mb_row_id`.
```
calherries, line 50: Since [this PR](https://github.com/metabase/metabase/pull/36249) we now unconditionally create the column, and it is always the primary key.
```suggestion
Metabase will create a primary key column called `_mb_row_id` as the first (left-most) column of the uploaded CSV table.
```

calherries: I'm not sure there's a clearer way to say "automatically generated integer primary key column". But an example is  in postgres, the column is created with `bigserial primary key` as its constraint.

sarina-dd (approved):

docs/databases/uploads.md
```diff
@@ -45,7 +45,7 @@ When people upload a CSV to a collection, Metabase will:
...
+When you upload a CSV, Metabase will create an a unique primary key column, called `_mb_row_id`, as the first (left-most) column of the uploaded CSV table. This `_mb_row_id` column will contain automatically generated integers. Even if the uploaded CSV already includes a column with a primary key, Metabase will treat that column like any other column (that is, Metabase will ignore the fact that the column contains a primary key and instead use the `_mb_row_id` column it created as the table's primary key). Metabase will also ignore any columns in the upload that have a name that will be in the database with the same name as the auto-generated primary key column (e.g., `_MB row-ID` in the CSV will be `_mb_row_id` or `_MB_ROW_ID` in the database).
```
calherries, line 48:
```suggestion
When you upload a CSV, Metabase will create an a unique primary key column, called `_mb_row_id`, as the first (left-most) column of the uploaded CSV table. This `_mb_row_id` column will contain automatically generated integers. Metabase will also ignore any columns in the upload that have a name that will be in the database with the same name as the auto-generated primary key column (e.g., `_MB row-ID` in the CSV will be `_mb_row_id` or `_MB_ROW_ID` in the database).
```

I'd suggest removing this sentence because CSVs themselves can't have primary keys. They're just text. Maybe they came from a database where one of the columns were a primary key, but I think users are smart enough to know that we can't detect that because the CSV file doesn't contain any of that information.

PR #36609, opened 2023-12-08
Adding card change cleanup to `update-dashboard-card!`

markbastian:
When changing the card belonging to a dashcard some attributes of the dashcard will no longer make sense -- series, parameter_mappings, and visualization_settings. This PR will compare the card ids in `update-dashboard-card!` and remove those entries in the update if the card ids are not the same.


markbastian:
Per @kulyk, this was solved by going in a different direction.

PR #36613, opened 2023-12-08
Underlying records drills: preserve the temporal bucket on a breakout column in the previous stage

### LINKED ISSUE #36582: [MLv2] [BE] [Bug] Post aggregation filters

camsaul:
If the previous stage had a breakout on a bucketed column, we should preserve that bucket when generating the `=` filter, e.g. generate

```
month(column) = March 2023
```

instead of

```
column = March 2023
```

Fixes #36582

I'm not 100% the way I did this is the cleanest way of fixing this -- the `lib.underlying` stuff is only used here, and I could have just changed that code to preserve the `:metabase.lib.field/temporal-bucket` key (instead of renaming it), but I wasn't sure about the implications of that. So this seemed a little safer


metamben (approved):
LGTM

Issue #36615, opened 2023-12-09
[MLv2] [FE] [Bug] FE not using MLv2 to apply column-filter drills


camsaul:
You can repro this in the `drills-integration` branch by running [the tests](https://github.com/metabase/metabase/blob/3946997cb3b2eeb07b8ec111e16e3c2322e0b762/e2e/test/scenarios/models/reproductions/22715-remapped-values-override-column-identifier.cy.spec.js#L50) for #22715:

while MLv2 is returning a `column-filter` drill here for `availableDrillThrus()`, the FE MLv1 is adding an invalid `time-interval` filter, which causes all sorts of warnings:

![image](https://github.com/metabase/metabase/assets/1455846/5d478568-f5e2-4d0b-8af7-44a5ba0bb301)

and, despite the warning messages saying it's removing that clause,  running the query with that clause anyway (error response):

![image](https://github.com/metabase/metabase/assets/1455846/6f660e4f-c65d-46c1-9751-783b97292457)

I added logging, the MLv2 `drillThru()` function is never getting called at all here. Also, none of the MLv2 drills generate `time-interval` filters, it looks like this is getting added by `StructuredQuery.setDefaultQuery()`.

Expected behavior: 

- FE code should not be manipulating the query itself when applying a drill thru; it should use MLv2 to transform the query

ranquild:
@camsaul console log statements with "removing" invalid filter clauses are irrelevant here and harmless. We would only be able to remove them fully when the filter modal and steps branches are merged, but the most important part is that they are not modifying the query sent to the backend in any way.

"Filter by this column" drill is powered by MLv2 as other drills. When we get `column-filter` drill, we show this option in the menu and show the filter picker (which is powered by MLv2 as well). When the filter is selected, we call `filter` function to modify the query. When I debug this I see that the query is already invalid after `filter` and `toLegacyQuery` calls:

<img width="1350" alt="Screenshot 2023-12-11 at 14 30 25" src="https://github.com/metabase/metabase/assets/8542534/7cbdc219-53fc-47b3-9932-ed546496c3f0">

ranquild:
Should we close this issue since the FE uses MLv2 to modify the query?

PR #36616, opened 2023-12-09
Generate sane queries for native query models with renamed columns

### LINKED ISSUE #36583: [MLv2] [BE] [Bug] Filters with native models and metadata overrides
branch `drills-integration`
test [when done through the column header action (metabase#22715-1)](https://github.com/metabase/metabase/blob/3946997cb3b2eeb07b8ec111e16e3c2322e0b762/e2e/test/scenarios/models/reproductions/22715-remapped-values-override-column-identifier.cy.spec.js#L50)
run https://www.deploysentinel.com/ci/runs/65734b6e266c1b276846f6c7

Unrelated to drills in particular. `filter` call generates an invalid query for this specific case when metadata is overwritten.

camsaul:
Disable the "force broken Card ID refs" hack when the source Card is a model, because models have the ability to rename columns but preserve the original column ID in their metadata, which means things can result in busted queries.

Fixes #36583

(I'm 95% sure this will fix it, but I can't verify for sure until #36615 is fixed)


metamben (approved):

ranquild (approved):
Can confirm that this change fixes the e2e test

PR #36624, opened 2023-12-11
Sync index info: mongo

qnkhuat:
Milestone 2 of https://github.com/metabase/metabase/issues/36452.
this should wrap up all drivers in tiers 1 and 2 that support indexing.

Supported indexes:
- single index: ✅
- compound index:  ✅
- multikey index:  ✅
- geospatial index:  ✅
- text index:  ✅
- hashed index  ✅
- wildcard index: ❌, it's not field specific index

Reference: all index types: https://www.mongodb.com/docs/manual/core/indexes/index-types/


modules/drivers/mongo/test/metabase/driver/mongo_test.clj
```diff
@@ -220,6 +220,89 @@
...
+                   [[{:small "http://example.com/small.jpg" :large "http://example.com/large.jpg"}]]])
+
+      (try
+       (testing "singly index"
+         (is (true? (t2/select-one-fn :database_indexed :model/Field (mt/id :singly-index :indexed))))
+         (is (false? (t2/select-one-fn :database_indexed :model/Field (mt/id :singly-index :not-indexed)))))
+
+       (testing "compount index"
+         (mongo.util/with-mongo-connection [conn (mt/db)]
+           (mcoll/create-index conn "compound-index" {"first" 1 "second" 1}))
```
calherries, line 245: To be explicit that this test relies on the ordering of fields in the index map, I'd suggest using `array-map` here as you do in the next test

modules/drivers/mongo/src/metabase/driver/mongo.clj
```diff
@@ -204,6 +205,24 @@
...
+(defmethod driver/describe-table-indexes :mongo
+  [_ database table]
+  (with-mongo-connection [^com.mongodb.DB conn database]
+    (->> (mcoll/indexes-on conn (:name table))
+         (map :key)
+         ;; when the key is not a persistent array map, then the ordered of the key can't be determined
+         ;; so we ignore those.
+         ;; Practically this will be only the case for when an index has more than 8 keys because a map
+         ;; with less than 8 keys are PersistentArrayMap, it's a PersistentHashedMap if it has more than 8 keys.
+         (filter #(instance? clojure.lang.PersistentArrayMap %))
```
calherries, line 217: It would be great to remove this edge case. Instead of calling `mcoll/indexes-on`, perhaps you could implement your own function that gets the indexes in a data structure that preserves ordering regardless of the size of the compound index?

calherries (commented):
What happens when a collection has other MongoDB index types, such as hashed or text indexes?

I see you've only explicitly tested the first three types of index from this list in their docs:
image

qnkhuat:
@calherries yea we support all except "Wildcard indexes" since it's not a field-specific index. I added the tests for all other index and updated the description.

modules/drivers/mongo/src/metabase/driver/mongo.clj
qnkhuat: I thought about that but it's a very low-level stuff in mongo, basically, we need to redefine a protocol for [ConvertToDBObject](https://github.com/michaelklishin/monger/blob/84170f7c51973e3d789736cdeb3cd8ad013071c3/src/clojure/monger/conversion.clj#L52).

We shouldn't do this as it'll affect the performance when reading large objects.

Also, I think a compound key with > 8 keys is quite rare.

calherries: I don't think you need to redefine the monger protocol, you can just use the Java API directly in this one specific use case. Like start with `(.getIndexInfo (.getCollection db (name coll)))`.

> Also, I think a compound key with > 8 keys is quite rare.

Yes, it's probably not worth fixing it for customers, but I think it's worth fixing just because this edge case complicates the code and I have to read this huge comment to understand what's it's doing.

qnkhuat: ha, I didn't realize you can access the dbobject like a normal object :). Good to know!
Implemented here https://github.com/metabase/metabase/pull/36624/files/31b0a616a7861fd1aceec4a35584bfc4cc527918..f683341abfd69eddd31d4ed1f72f74709770a8f9

Thanks Cal!

src/metabase/sync/sync_metadata/indexes.clj
```diff
@@ -13,27 +13,47 @@
    :added-indexes 0
    :removed-indexes 0})
 
+(defn- nested-field-names->field-id
```
calherries, line 16: There's a column on `metabase_field` called `nfc_path` that should have the information you need to implement this function without resorting to recursion. I think this should be a function inside the `field` namespace though.

src/metabase/driver.clj
```diff
@@ -317,7 +317,8 @@
   :hierarchy #'hierarchy)
 
 (defmulti describe-table-indexes
-  "Returns a set of field names that are either indexed or are the first columns in a composite index."
+  "Returns a set of field names that are either indexed or are the first columns in a composite index.
+  Field name can be a string, nested field need to be represented as a sequence of strings."
```
calherries, line 321: Thinking ahead to being able to supporting composite indices, I don't love that nested fields are represented as a sequence of strings. It's going to make change harder and more error prone.

Perhaps this function should return a set of maps? e.g.
`{:type :normal-column-index, :field-name "..."}`
`{:type :nested-column-index, :nfc-path ["..." "..."]}`

calherries (changes requested):

modules/drivers/mongo/test/metabase/driver/mongo_test.clj
```diff
@@ -220,6 +220,101 @@
...
+                  ["compound-index-big"
+                   [{:field-name "a" :base-type :type/Text}]
+                   [[1]]]
+                  ["multi-key-index"
+                   [{:field-name "a" :base-type :type/Text}]
+                   [[1]]]
+                  ["advanced-index"
+                   [{:field-name "hashed-field" :indexed? false :base-type :type/Text}
+                    {:field-name "text-field" :indexed? false :base-type :type/Text}
+                    {:field-name "geospaitial-field" :indexed? false :base-type :type/Text}]
```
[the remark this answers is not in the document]
calherries: Other than that, these tests are great

test/metabase/driver/sql_jdbc/sync/describe_table_test.clj
```diff
@@ -469,7 +469,7 @@
                         (t2/select-one Table :db_id (mt/id) :name "json_without_pk"))))))))))))
 
 (deftest describe-table-indexes-test
-  (mt/test-drivers (mt/normal-drivers-with-feature :index-info)
+  (mt/test-drivers (disj (mt/normal-drivers-with-feature :index-info) :mongo)
```
calherries, line 472: This test should cover the sql-jdbc driver implementations only.
```suggestion
  (mt/normal-drivers-with-feature :index-info :sql-jdbc)
```

src/metabase/sync/sync_metadata/indexes.clj
calherries: Oh sorry, that's for JSON fields in SQL databases. I'm surprised something similar doesn't exist for mongoDB, oh well

test/metabase/sync/sync_metadata/indexes_test.clj
```diff
@@ -21,7 +21,7 @@
       (is (false? (t2/select-one-fn :database_indexed :model/Field (mt/id :table :not-indexed)))))))
 
 (deftest sync-composite-indexed-columns-test
-  (mt/test-drivers (mt/normal-drivers-with-feature :index-info)
+  (mt/test-drivers (disj (mt/normal-drivers-with-feature :index-info) :mongo)
```
calherries, line 24: As implemented this test can only work for `:sql-jdbc` drivers.
```suggestion
  (mt/test-drivers (mt/normal-drivers-with-feature :index-info :sql-jdbc)
```
Does it make sense to move this to `metabase.driver.sql-jdbc-test`?

modules/drivers/mongo/test/metabase/driver/mongo_test.clj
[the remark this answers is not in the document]
qnkhuat: yep, creating an index does not validate at all.

For normal indexes, the doc says you need to specify a field name and the sort order where 1 is asc and -1 dec.
But you can even input 5 or -2 sort order.

test/metabase/sync/sync_metadata/indexes_test.clj
qnkhuat: there is no feature called "sql-jdbc", we need to intersect the feature drivers with `(mt/sql-jdbc-drivers)`

qnkhuat: > Does it make sense to move this to metabase.driver.sql-jdbc-test?
I think it's fine here as we treat tests for mongo is a kind of special case where most mongo tests are defined in its namespace.

src/metabase/driver.clj
qnkhuat: yea makes sense.

One thing I'd change is to use a map of 2 keys `[:type :value]`.

The type is enough to differentiate the field name / path is represented, we don't need to use a different name for it.

Implement it here https://github.com/metabase/metabase/pull/36624/commits/20055e1504e1b24dc830ac3bb4ee1e0b823cd7f4

src/metabase/sync/sync_metadata/indexes.clj
qnkhuat: moved in https://github.com/metabase/metabase/pull/36624/commits/f96c83c58fd4e479576d6d22b41e6a3ac01273f9

test/metabase/sync/sync_metadata/indexes_test.clj
calherries: Yeah fair enough. I guess what I'm trying to get at is the implementation of this test uses `sql-jdbc.conn/db->pooled-connection-spec` and `sql.tx/create-index-sql`, so it is specific to `sql-jdbc` drivers. If bigquery had indexes this would be more of a problem, but I guess it's okay. Ideally there would be a multimethod `tx/create-index!` that had a default implementation for SQL drivers that called `sql.tx/create-index-sql`. This is how it works for `create-db!` anyway. https://github.com/metabase/metabase/blob/772e8eec90b82159aaa406615014e8d571937d5d/test/metabase/test/data/sql_jdbc/load_data.clj#L198

Anyway, it's not related to this PR and probably not worth the ceremony

calherries (approved):
Looks great now, thanks

PR #36641, opened 2023-12-11
Fix uploading CSV columns with boolean and integer values

### LINKED ISSUE #36629: CSV uploads: Can't upload CSV with a column that has boolean and integer values
[(slack thread where this was raised)](https://metaboat.slack.com/archives/C05MPF0TM3L/p1702074978417889)

### Describe the bug

If you try to upload a CSV with a column that has both integer and boolean values (where there is an integer value that is not 0 or 1), the upload fails with an error:

image

### To Reproduce

1. Enable uploads to the sample database
2. Upload a CSV file to a collection with the contents:
```
some_column
2
true
```
3. Observe the upload fails with an error message

### Expected behavior

the upload should succeed, and the column should be of type text

### Logs

_No response_

```JSON
master
```

### Severity

blocking some uploads

### Additional context

_No response_

calherries:
Fixes #36629 by making columns with boolean and integer values to be of the text type.

Now a CSV upload with a column like this will succeed:
```
some_column_name
2
true
false
no
yes
```

The column will have the base-type `:type/Text`.

### Implementation details

Fixing this bug required some pretty fundamental changes to the assumptions made by our schema detection code.

Read this change to the documentation first
 https://github.com/metabase/metabase/pull/36641/files#diff-af3188166dbf14ec61d196f0e71de90a50f7e2c9de2c152d7cd2e20c6af4051bR43-R70

**Before:**
1. all types of values correspond to a parse function, via `upload-type->parser`
2. types form a tree coercion hierarchy
```
;;              text
;;               |
;;               |
;;          varchar-255┐
;;              / \    │
;;             /   \   └──────────┬
;;            /     \             │
;;         float   datetime  offset-datetime
;;           │       │
;;           │       │
;;          int     date
;;         /   \
;;        /     \
;;    boolean auto-incrementing-int-pk
```

**Now:**
1. there is now a distinction between the types a column can take (column types), and the types of a single value (value types). Only valid column types correspond to a parse function via `upload-type->parser`. For example, `::boolean-or-int` is a valid value type, but not a valid column type.
2. types form a directed acyclic graph coercion hierarchy

```
;;              text
;;               |
;;               |
;;          varchar-255┐
;;        /     / \    │
;;       /     /   \   └──────────┬
;;      /     /     \             │
;;  boolean float   datetime  offset-datetime
;;     |     │       │
;;     │     │       │
;;     |    int     date
;;     |   /   \
;;     |  /     \
;;     | /       \
;;     |/         \
;; boolean-or-int  auto-incrementing-int-pk
```

Because our coercion hierarchy was limited to a tree structure, we had to make `boolean` as a subtype of `int` to allow "0" and "1" to be able to be coerced to both booleans and integers. But that's incorrect, because not all `boolean` values should be coerced to an `int`. Only `"0"` and `"1"` should be able to be coerced to `int`, not other values like "true" or "no". That lead to us incorrectly inferring the type of a column with integer values mixed with boolean values that are not "0" or "1" to be of type integer.

The solution is to make the coercion hierarchy a graph structure, where types can be coerced to more than one parent. Then "0" and "1" have a special type `boolean-or-int` that can be coerced as `boolean` if they are the only values in that column, and coerced as `int` if they are combined with other `int` values that aren't "0" or "1".

Example: If you have a CSV with one column:
```
some_column_name
1
2
true
```

The type detected for each value would have previously been: 
```
value | value type before | value type now
──────────────────────────────────────────
1     | boolean           | boolean-or-int
2     | int               | int
true  | boolean           | boolean
```

Before, the column would be of type `int`, because `int` is an ancestor of `boolean` in the coercion hierarchy. Now, the column would be of type `varchar-255`, because the lowest common ancestor of `boolean`, `boolean-or-int`, and `int`, is `varchar-255`.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -40,45 +40,72 @@
 ;;;; | Schema detection |
 ;;;; +------------------+
 
+;; Upload types form a DAG (directed acyclic graph) where each type can be coerced into any
+;; of its ancestors types. We parse each value in the CSV file to the most-specific possible
+;; type for each column. The most-specific possible type for a column is the lowest common
+;; ancestor of the types for each value in the column.
+;;
 ;;              text
 ;;               |
 ;;               |
 ;;          varchar-255┐
-;;              / \    │
-;;             /   \   └──────────┬
-;;            /     \             │
-;;         float   datetime  offset-datetime
-;;           │       │
-;;           │       │
-;;          int     date
-;;         /   \
-;;        /     \
-;;    boolean auto-incrementing-int-pk
+;;        /     / \    │
+;;       /     /   \   └──────────┬
+;;      /     /     \             │
+;;  boolean float   datetime  offset-datetime
+;;     |     │       │
+;;     │     │       │
+;;     |    int     date
+;;     |   /   \
+;;     |  /     \
+;;     | /       \
+;;     |/         \
+;; boolean-or-int  auto-incrementing-int-pk
+;;
+;; `boolean-or-int` is a special type with two parents, where we parse it as a boolean if the whole
+;; column's values are of that type. additionally a column cannot have a boolean-or-int type, but
+;; a value can. if there is a column with a boolean-or-int value and an integer value, the column will be int
+;; if there is a column with a boolean-or-int value and a boolean value, the column will be boolean
+;; if there is a column with only boolean-or-int values, the column will be parsed as if it were boolean
 ;;
 ;; </code></pre>
 
-(def ^:private type->parent
+(def ^:private type+parent-pairs
   ;; listed in depth-first order
-  {::text                     nil
-   ::varchar-255              ::text
-   ::float                    ::varchar-255
-   ::int                      ::float
-   ::auto-incrementing-int-pk ::int
-   ::boolean                  ::int
-   ::datetime                 ::varchar-255
-   ::date                     ::datetime
-   ::offset-datetime          ::varchar-255})
+  '([::boolean-or-int ::boolean]
+    [::boolean-or-int ::int]
+    [::auto-incrementing-int-pk ::int]
+    [::int ::float]
+    [::date ::datetime]
+    [::boolean ::varchar-255]
+    [::offset-datetime ::varchar-255]
+    [::datetime ::varchar-255]
+    [::float ::varchar-255]
+    [::varchar-255 ::text]))
+
+(def ^:private type->parents
+  (reduce
+   (fn [m [type parent]]
+     (update m type conj parent))
+   {}
+   type+parent-pairs))
 
 (def ^:private types
-  (keys type->parent))
+  "All types, including the root type, ::text"
+  (conj (keys type->parents) ::text))
 
-(def ^:private type->ancestors
+(defn- bfs-ancestors [type]
+  (loop [visit   (list type)
+         visited (ordered-set/ordered-set)]
+    (if (empty? visit)
+      visited
+      (let [parents (mapcat type->parents visit)]
+        (recur parents (into visited parents))))))
+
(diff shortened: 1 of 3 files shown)


test/metabase/upload_test.clj
```diff
@@ -171,42 +176,47 @@
            [" 2022-01-01T01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]
            [" 2022-01-01t01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]
            [" 2022-01-01 01:00:00.00Z "       (t/offset-date-time "2022-01-01T01:00+00:00") offset-dt-type]]]
-    (let [settings {:number-separators (or seps ".,")}
-          type     (#'upload/value->type string-value settings)
-          parser   (upload-parsing/upload-type->parser type settings)]
+    (let [settings   {:number-separators (or seps ".,")}
+          value-type (#'upload/value->type string-value settings)
+          ;; get the type of the column, if it were filled with only that value
+          col-type   (first (#'upload/column-types-from-rows settings 1 [[string-value]]))
```
calherries, line 182: I made this change to the test because types of values don't correspond to a single parser anymore. Only column types correspond to a parser now. So to get a column type from a value, you need to create a fake row with a column that contains only that value, and then get the column type.

test/metabase/upload_test.clj
```diff
@@ -258,16 +268,26 @@
...
+    (testing "Boolean and integers together"
+      (is (=? (with-ai-id {:vchar       vchar-type
+                           :bool        bool-type
+                           :bool_or_int bool-type})
+              (@#'upload/detect-schema
+               (csv-file-with ["vchar,bool,bool-or-int"
+                               " true,true,          1"
+                               "    1,   1,          0"
+                               "    2,   0,          0"
+                               "   no,  no,          1"])))))
```
piranha, line 290: Would be nice to see another column like 1/0/1/2 to check it correctly converges to int.

calherries: Good idea, added this to both the new tests ba2caccb11807b0d531a8e32977135df934d0814

src/metabase/upload.clj
```diff
@@ -232,14 +274,26 @@
...
+  "Returns a sequence of types, given the unparsed rows in the CSV file"
+  [settings column-count rows]
+  (->> rows
+       (map #(row->value-types % settings))
+       (reduce coalesce-types (repeat column-count nil))
+       (map (fn [type]
+              ;; if there's no values in the column, assume it's a string
+              (if (nil? type)
+                ::text
+                (column-type type))))))
```
piranha, line 295: Ehhh IDK feels like putting `nil ::text` as a case in `column-type` would be nice to get cleaner code (just `(map column-type)` here at the price of obscuring the logic somewhat.

sansolovyov (approved):
Nice, clear and sound!

src/metabase/upload.clj
calherries: Maybe, can you explain why? The idea is to keep the the logic for handling the edge case of an empty column separate from the logic in `column-type`, which handles when there is at least one value in a column.

If I put the logic in `column-type`, I'd have to explain in `column-type`'s docstring that a `nil` value for `type` corresponds to an empty column. That seems to couple `column-type`'s implementation to the implementation of `coalesce-types` a bit, and I think that doing it this way avoids that coupling.

piranha: I looked at it from the point where if data type is unknown (`nil`), then column type should be `text`. But, as I've said, "nicer code at the price of obscuring the logic". :)

calherries:

PR #36646, opened 2023-12-11
Audit Log should also allow statistics for Public/Embedding - including downloads

### LINKED ISSUE #28633: Audit Log should also allow statistics for Public/Embedding - including downloads
Is your feature request related to a problem? Please describe.
Audit Log currently doesn't show any information about Public//Embedding activity (not sure if this is more of a bug than feature)

Describe the solution you'd like
Audit Log should also allow statistics for Public/Embedding - including downloads

Describe alternatives you've considered
Nginx log and some tool to analyze those logs

Related https://github.com/metabase/metabase/issues/11909 and https://github.com/metabase/metabase/issues/18758

NOTE, also adding comments from @salsakran 
A user asked for the ability to track usage of embeds, especially

- the clicks on 'download data' underneath graphs
- the clicks on additional pages on graphs (ie when there are multiple pages to view below the tables)

In general, having an "embeds" section of the audit app might be useful.

qwef:
Closes https://github.com/metabase/metabase/issues/28633

### Description

Adds dashboard events for public sharing and embedding. 

### Demo

image


test/metabase/api/embed_test.clj
```diff
@@ -670,6 +684,25 @@
...
+(deftest embed-download-query-execution-test
+  (testing "Tests that embedding download context shows up in the query execution table when downloading cards."
+    ;; Clear out the query execution log so that test doesn't read stale state
+    (t2/delete! :model/QueryExecution)
+    (mt/with-ensure-with-temp-no-transaction!
+      (with-embedding-enabled-and-new-secret-key
+        (with-temp-dashcard [dashcard {:dash {:enable_embedding true}
+                                       :card {:dataset_query (mt/mbql-query venues)}}]
+          (client/client :get 200 (str (dashcard-url dashcard) "/csv"))
+          ;; the query execution is saved async, so we need to sleep a bit
```
qwef, line 696: copied from https://github.com/metabase/metabase/blob/jh-track-embed/test/metabase/actions/execution_test.clj#L19

luizarakaki:
This should be backported, right? Otherwise it will land only on 49?

test/metabase/api/embed_test.clj
noahmoss: Can you use `metabase.query-processor.middleware.process-userland-query-test/with-query-execution` here instead? It redefs `save-query-execution*` so that it stores the data in a promise, and then derefs it with a timeout instead of waiting a constant 200 ms each time. Bonus points if you can switch the actions test to use it as well.

Sleeping for hardcoded intervals in tests is generally an antipattern. A single case isn't very noticeable but if it proliferates it can add a lot of waiting time to the test suite.

noahmoss (changes requested):
Generally looks good. Requesting changes for switching the tests to use a helper macro instead of `Thread/sleep`

qwef:
Confused on using `metabase.query-processor.middleware.process-userland-query-test/with-query-execution ` here. Doesn't that function require a query as input? How would I be able to pass in the client GET?

noahmoss:
@qwef Oh, hmm. Looks like that macro just requires the query for computing its hash, so that it can compare it to the hash in the query execution map and ensure that it's the correct one. I think you could still pass in the query from the dashcard you're executing into that macro, and then call the endpoint in the body of the macro, and it should work as expected (as long as the query hashes match up).

That's just from reading the code though, I haven't tried it. If you're still hitting issues let me know and I can help dig into it more (or we could pair on it together if you want)

src/metabase/events/schema.clj
```diff
@@ -10,10 +10,14 @@
                       [:map {:closed true}
                        [:user-id pos-int?]
                        [:object [:fn #(t2/instance-of? :model/Dashboard %)]]])
+      view-only      (mc/schema
+                      [:map {:closed true}
```
noahmoss, line 14: nit: it was a little unclear to me what `view-only` means—the only difference is that `user-id` is optional, right?

I think it would actually be fine to just change the default-schema to take an optional user ID, but up to you

noahmoss (approved):
Nice job getting the test macro to work! I suppose it's a bit annoying to need to construct the query with all of the extra fields that get hashed like `:middleware`, etc... I wonder if there's a way to streamline that while keeping the core behavior of the macro the same. But don't need to block this PR on that kind of change. Just left one last nit, lgtm otherwise

PR #36657, opened 2023-12-11
WIP 35503 fix card filter connection

### LINKED ISSUE #35503: Connect a card to a filter in a dashboard doesn't update data

oleggromov:
Fixes https://github.com/metabase/metabase/issues/35503

WIP

### TODO:
- [x] find a correct approach
- [ ] confirm important use cases aren't broken by the fix
- [x] finish tests

### DIFF
--- a/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
+++ b/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
@@ -42,10 +42,6 @@ class DashboardInner extends Component {
     isSharing: false,
   };
 
-  constructor(props) {
-    super(props);
-  }
-
   static getDerivedStateFromProps({ parameters }, { parametersListLength }) {
     const visibleParameters = getVisibleParameters(parameters);
     return visibleParameters.length !== parametersListLength
@@ -54,14 +50,14 @@ class DashboardInner extends Component {
   }
 
   async componentDidMount() {
-    await this.loadDashboard(this.props.dashboardId);
+    await this.loadDashboard();
   }
 
   async componentDidUpdate(prevProps) {
     updateParametersWidgetStickiness(this);
 
     if (prevProps.dashboardId !== this.props.dashboardId) {
-      await this.loadDashboard(this.props.dashboardId);
+      await this.loadDashboard();
       return;
     }
 
@@ -73,6 +69,7 @@ class DashboardInner extends Component {
 
     if (
       !_.isEqual(prevProps.parameterValues, this.props.parameterValues) ||
+      !_.isEqual(prevProps.parameters, this.props.parameters) ||
       (!prevProps.dashboard && this.props.dashboard)
     ) {
       this.props.fetchDashboardCardData({ reload: false, clearCache: true });
@@ -83,51 +80,39 @@ class DashboardInner extends Component {
     this.props.cancelFetchDashboardCardData();
   }
 
-  async loadDashboard(dashboardId) {
-    const {
-      editingOnLoad,
-      addCardOnLoad,
-      addCardToDashboard,
-      fetchDashboard,
-      initialize,
-      loadDashboardParams,
-      location,
-      setErrorPage,
-      isNavigatingBackToDashboard,
-    } = this.props;
+  async loadDashboard() {
+    const p = this.props;
+    p.initialize({ clearCache: !p.isNavigatingBackToDashboard });
+    p.loadDashboardParams();
 
-    initialize({ clearCache: !isNavigatingBackToDashboard });
-
-    loadDashboardParams();
-
-    const result = await fetchDashboard({
-      dashId: dashboardId,
-      queryParams: location.query,
+    const result = await p.fetchDashboard({
+      dashId: p.dashboardId,
+      queryParams: p.location.query,
       options: {
-        clearCache: !isNavigatingBackToDashboard,
-        preserveParameters: isNavigatingBackToDashboard,
+        clearCache: !p.isNavigatingBackToDashboard,
+        preserveParameters: p.isNavigatingBackToDashboard,
       },
     });
 
     if (result.error) {
-      setErrorPage(result.payload);
+      p.setErrorPage(result.payload);
       return;
     }
 
     try {
-      if (editingOnLoad) {
+      if (p.editingOnLoad) {
         this.setEditing(this.props.dashboard);
       }
-      if (addCardOnLoad != null) {
-        addCardToDashboard({
-          dashId: dashboardId,
-          cardId: addCardOnLoad,
-          tabId: this.props.dashboard.tabs[0]?.id ?? null,
+      if (p.addCardOnLoad != null) {
+        p.addCardToDashboard({
+          dashId: p.dashboardId,
+          cardId: p.addCardOnLoad,
+          tabId: p.dashboard?.tabs[0]?.id ?? null,
         });
       }
     } catch (error) {
       if (error.status === 404) {
-        setErrorPage({ ...error, context: "dashboard" });
+        p.setErrorPage({ ...error, context: "dashboard" });
       } else {
         console.error(error);
         this.setState({ error });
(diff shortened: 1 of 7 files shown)


frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -83,51 +83,39 @@ class DashboardInner extends Component {
...
-
-    initialize({ clearCache: !isNavigatingBackToDashboard });
-
-    loadDashboardParams();
-
-    const result = await fetchDashboard({
-      dashId: dashboardId,
-      queryParams: location.query,
+  async loadDashboard() {
+    const p = this.props;
```
ranquild, line 87: We don't use this convention anywhere. We either destructure props or use them as it is. I'd recommend destructuring

ranquild (changes requested):
The fix is incorrect - it causes additional requests when connecting the dashboard parameter to the dashcard, like was in the original PR.

This PR
Screenshot 2023-12-12 at 12 54 25

master
Screenshot 2023-12-12 at 12 55 54

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
oleggromov: Reverted back.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -73,6 +69,11 @@ class DashboardInner extends Component {
 
     if (
       !_.isEqual(prevProps.parameterValues, this.props.parameterValues) ||
+      (!this.props.isEditing &&
```
oleggromov, line 72: I have updated the conditional so that we now check for `isEdited`. Do you think it's enough?

kulyk: What can be changing dashcards outside of edit mode?

kulyk: Another comment along the way: IMO, this condition block became too complex. We should extract parts of the expression into constants with human-readable names. I'm thinking of something like:

```ts
const didParameterValuesChange = !_.isEqual(prevProps.parameterValues, this.props.parameterValues);
const didDashCardsChange = !_.isEqual(
  prevProps.dashboard?.dashcards,
  this.props.dashboard?.dashcards,
))
const didLoadDashboard = !prevProps.dashboard && this.props.dashboard;

if (
  didLoadDashboard ||
  didParameterValuesChange ||
  (!isEditing && didDashCardsChange)
) { ... }
```

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
kulyk, line 1: So great to see we now have a `Dashboard` test suite, thanks for adding it 👍

oleggromov:
All feedback has been taken into account. 
This is how I will proceed: https://github.com/metabase/metabase/pull/36657#discussion_r1423987526

oleggromov:
It is a bit more complicated than I thought. 

1. It's not enough to compare each dashcard's `parameter_mappings`, since they change during the editing phase only
2. The previous conditional was triggering because of the `isDirty` changing `true => undefined`, when the card is saved. At this point `isEditing` of the dashboard was already false, which allowed us to correctly issue dashboard fetch action.

However, this is almost no different than reacting to any change in the `dashcards` map.

The bottom line is, this component isn't the right place to perform the check. What we need instead is to move this logic away to the "business logic layer" (which we don't have but somewhere between redux actions and entity loaders will be better) - but this is way beyond the scope of a bug fix.

So I'll try to find a combination of props that allows us to re-fetch things correctly for now.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -71,20 +67,43 @@ class DashboardInner extends Component {
...
       this.props.fetchDashboardCardData({ reload: false, clearCache: true });
     }
   }
 
+  _getDashcardsIsDirty(dashcards = []) {
+    return dashcards.reduce((acc, dashcard) => {
+      acc[dashcard.id] = dashcard.isDirty ?? false;
+      return acc;
+    }, {});
+  }
```
oleggromov, line 98: This isn't really a method but I haven't seen any function dangling around in component files, so I am putting it here. And I don't think adhoc functions like this one deserve to be called "utils" as they most likely won't be used twice.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+  selectedTabId: number | null; // when there's no tabs, is null
+  parameters: Parameter[]; // when empty, is an empty array
+  skipLoader?: boolean;
+  isEditing: boolean;
+};
+
+function dashboardWithCards(first: { isDirty?: boolean } = {}) {
+  return createMockDashboard({
+    dashcards: [
+      createMockDashboardCard({ id: 1, isDirty: first.isDirty } as any),
```
oleggromov, line 21: I didn't dare to change the return type to `StoreDashcard` to get `isDirty`. So meet `as any`.

frontend/src/metabase/dashboard/selectors/selectors-typed.ts
```diff
@@ -18,9 +18,11 @@ export function getSelectedTabId(state: State) {
   return state.dashboard.selectedTabId;
 }
 
-export const getDashcards = (state: State) => state.dashboard.dashcards;
+export function getDashcards(state: State) {
+  return state.dashboard.dashcards;
+}
 
-export const getDashCardById = (state: State, dashcardId: DashCardId) => {
+export function getDashCardById(state: State, dashcardId: DashCardId) {
```
oleggromov, line 25: I touched this file when I thought I'd need selectors but decided to leave these changes for consistency anyways.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
```diff
@@ -71,20 +67,43 @@ class DashboardInner extends Component {
...
+      this.props.parameterValues,
+    );
+    // This is a way of detecting card parameter mapping changes. Parameter
+    // mappings change happen independently of isEditing because of re-renders,
+    // so we use this heuristic to avoid additional state. The fetching logic
+    // should be moved outside.
+    const isDirtyChanged = !_.isEqual(
+      this._getDashcardsIsDirty(prevProps.dashboard?.dashcards),
+      this._getDashcardsIsDirty(this.props.dashboard?.dashcards),
+    );
```
kulyk, line 81: At first glance, this doesn't look so different from `!_.isEqual(prevDashcard, dashcard)`, does it?
We wanted to avoid that kind of comparison, so minor changes to a dashcard (like something purely visual) won't trigger query reruns. IIRC, `isDirty: true` should be set on a dashcard after _every_ edit, so this still feels like a more optional `_.isEqual(prevDashcard, dashcard)`. Does that make sense or did I miss something?

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+  createMockDashboard,
+} from "metabase-types/api/mocks";
+import type { Dashboard, Parameter } from "metabase-types/api";
+import { Dashboard as DashboardComponent } from "./Dashboard";
+
+type SetupOpts = {
+  dashboardId: number | null;
+  dashboard: Dashboard | null;
+  selectedTabId: number | null; // when there's no tabs, is null
+  parameters: Parameter[]; // when empty, is an empty array
```
kulyk, line 13: It feels like the code here is pretty self-explanatory and we probably don't need these comments

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+  return createMockDashboard({
+    dashcards: [
+      createMockDashboardCard({ id: 1, isDirty: first.isDirty } as any),
+      createMockDashboardCard({ id: 2 }),
+    ],
+  });
+}
+
+async function setup(overrides: Partial<SetupOpts> = {}) {
+  const mockDashboard = createMockDashboard({ id: 10 }); // value is irrelevant
```
kulyk, line 28: Should we just call `createMockDashboard()` without any args if we don't care about the values?
We can use `mockDashboard.id` below if we need it

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+        isNightMode={false}
+        isSharing={false}
+        isEditingParameter={false}
+        isNavbarOpen={false}
+        isHeaderVisible={false}
+        isAdditionalInfoVisible={false}
+        isNavigatingBackToDashboard={false}
+        editingOnLoad={false}
+        addCardOnLoad={null}
+        isAutoApplyFilters={false} // TODO true or false?
```
kulyk, line 68: Think we should use `dashboard?.auto_apply_filters` like [here](https://github.com/metabase/metabase/blob/fea5d8cb4dd3e8b8a25bc15d5b15fca36ff8ff04/frontend/src/metabase/dashboard/selectors/selectors.js#L157)

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+      rerender(<TestComponent {...{ ...opts, ...overrides }} />),
+    mockLoadDashboardParams,
+    mockFetchDashboard,
+    mockFetchDashboardCardData,
+    mockFetchDashboardCardMetadata,
+  };
+}
+
+describe("Dashboard data fetching", () => {
+  afterEach(() => jest.clearAllMocks());
```
kulyk, line 126: Doesn't Jest do that automatically?

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
```diff
@@ -0,0 +1,188 @@
...
+    mocks.rerender({
+      dashboardId: null,
+      dashboard: createMockDashboard({ id: 20 }),
+    });
+    expect(mocks.mockFetchDashboardCardData).toHaveBeenCalledTimes(1);
+  });
+
+  it("should fetch card data when isDirty changes and not editing", async () => {
+    const mocks = await setup({ dashboard: dashboardWithCards() });
+    jest.clearAllMocks();
```
kulyk, line 172: A few tests are using `setup()` + `jest.clearAllMocks()` pattern. Would it make sense to run `jest.clearAllMocks()` at the end of `setup`?

kulyk (commented):
Left a few suggestions, guess the main comment is about how we compare dashcards

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
oleggromov: It actually does, great catch!

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
oleggromov: Done!

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.unit.spec.tsx
oleggromov: I don't think so. IMO tests must not be too abstract, and this is quite an important detail of each test case to be hidden somewhere.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
ranquild: @oleggromov should we compare only `parameter_mappings`? Something like this:
```
getParameterMappings(dashboard) {
    return dashboard.dashcards.map(dashcard => dashcard.parameter_mappings);
}

....
if (!_.isEqual(
  getParameterMappings(this.props.dashboard)),
  getParameterMappings(prevProps.dashboard))
) {
  this.props.fetchDashboardCardData({ reload: false, clearCache: true });
}
```

oleggromov: https://github.com/metabase/metabase/pull/36657#issuecomment-1852265577

ranquild: @kulyk wdyt? Based on the original issue I think it's what we want to handle

ranquild (commented):
Comments inline

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx
oleggromov: We did explore it with @ranquild and I hope we've found a better fix within the `UPDATE_DASHBOARD_AND_CARDS` action. I'll make a new version of the fix and submit it for review soon.
Thank you.

oleggromov:
We might have found a better way, I'd try to implement, test it and submit for review.
@kulyk @ranquild thank you.

PR #36665, opened 2023-12-12
Sync index info: sqlserver and sqlite

qnkhuat:
Milestone 3 of https://github.com/metabase/metabase/issues/36452

These 2 drivers are JDBC drivers so it already works, this PR is mostly about making the tests work properly

After this we should be done syncing index info for all the drivers that support it


modules/drivers/sqlite/test/metabase/test/data/sqlite.clj
```diff
@@ -44,6 +44,13 @@
     (when (#{:count :cum-count} ag-type)
       {:base_type :type/Integer}))))
 
+(defmethod sql.tx/create-table-sql :sqlite
+  [driver dbdef table-def]
+  ;; Sqlite doesn't automatically create an index for pk, so we need to manually create one
```
calherries, line 49: Can you explain _why_ we need to manually create one?

qnkhuat: the tests in `metabase.driver.sql-jdbc.sync.describe-table-test/describe-table-indexes-advanced-index-type-test` assumes all id columns are indexed.

this is to comply with that.

src/metabase/driver/sql_jdbc/sync/describe_table.clj
```diff
@@ -301,7 +301,8 @@
                          ;; we only sync columns that are either singlely indexed or is the first key in a composite index
                          (first (map :column_name (sort-by :ordinal_position idx-values)))))
           vals
-          set)))))
+          set
+          (disj nil))))))
```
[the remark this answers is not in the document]
qnkhuat: this is a safeguard for cases where the jdbc sometimes returns an index that has nil in column name.
I found this while working with sqlserver. will add an explanation

modules/drivers/sqlite/test/metabase/test/data/sqlite.clj
calherries: If I'm reading the documentation correctly SQLite ([here](https://www.sqlite.org/lang_createtable.html#:~:text=3.6.%20UNIQUE%20constraints) and [here](https://www.sqlite.org/lang_createindex.html#:~:text=DROP%20INDEX%20command.-,1.1.%20Unique%20Indexes,-If%20the%20UNIQUE)), a single primary key column is indexed. Same with any column with a `UNIQUE` constraint.

If `.getIndexInfo` does not return them, then `describe-table-indexes` should have a special implementation that returns them anyway.

calherries:
Looks like there's a few sqlite tests failing, I'm not sure why though because they seem unrelated.

modules/drivers/sqlite/test/metabase/test/data/sqlite.clj
qnkhuat: can you refer to the part where it says the PK is indexed?

either way I did a check and it doesn't create an index by default.

```
sqlite> CREATE TABLE user ( id INTEGER PRIMARY KEY, name TEXT);


sqlite> CREATE INDEX idx_user_name ON user (name);CREATE INDEX idx_user_name ON user (name);
sqlite> PRAGMA index_list(user);
0|idx_user_name|0|c|0


sqlite> CREATE INDEX idx_user_id ON user (id);
sqlite> PRAGMA index_list(user);
0|idx_user_id|0|c|0
1|idx_user_name|0|c|0
sqlite>
```

calherries: > can you refer to the part where it says the PK is indexed?

[here](https://www.sqlite.org/lang_createtable.html#:~:text=3.6.%20UNIQUE%20constraints)
```
In most cases, UNIQUE and PRIMARY KEY constraints are implemented by creating a unique index in the database. (The exceptions are [INTEGER PRIMARY KEY](https://www.sqlite.org/lang_createtable.html#rowid) and PRIMARY KEYs on [WITHOUT ROWID](https://www.sqlite.org/withoutrowid.html) tables.) Hence, the following schemas are logically equivalent:

CREATE TABLE t1(a, b UNIQUE);

CREATE TABLE t1(a, b PRIMARY KEY);

CREATE TABLE t1(a, b);
CREATE UNIQUE INDEX t1b ON t1(b);
```

and in the documentation for [unique index](https://www.sqlite.org/lang_createindex.html#:~:text=DROP%20INDEX%20command.-,1.1.%20Unique%20Indexes,-If%20the%20UNIQUE) there's nothing saying a "unique index isn't a real index" or anything like that.

> either way I did a check and it doesn't create an index by default.

That's because it's an `INTEGER PRIMARY KEY`
https://stackoverflow.com/questions/55036387/why-would-pragma-schema-index-listtable-name-not-show-primary-keys

calherries: `index_list` does return an index for the primary key if it's `TEXT PRIMARY KEY`.

```
sqlite> CREATE TABLE user ( id TEXT PRIMARY KEY, name TEXT);
sqlite> PRAGMA index_list(user);
0|sqlite_autoindex_user_1|1|pk|0
```

PRIMARY KEY and unique constraints are included in the documentation for `index_list` [here](https://www.sqlite.org/pragma.html#:~:text=PRAGMA%20schema.index_list(table%2Dname)%3B).

calherries: I think the correct behaviour here is to make a special case for sqlite integer primary keys.

And we should also mark columns that are `UNIQUE` as indexed too. I think you might just need to add a test for that though, because your implementation might already handle it.

modules/drivers/sqlite/src/metabase/driver/sqlite.clj
```diff
@@ -77,6 +79,19 @@
...
+(defmethod driver/describe-table-indexes :sqlite
+  [driver database table]
+  (let [pk (first (sql-jdbc.execute/do-with-connection-with-options
+                   driver database nil
+                   (fn [conn]
+                     (sql-jdbc.describe-table/get-table-pks :sqlite conn (:name database) table))))]
+    ;; In sqlite a PK will implicitly have a UNIQUE INDEX, but if the PK is integer the getIndexInfo method from
+    ;; jdbc doesn't return it as indexed. so we need to manually get mark the pk as indexed here
+    (cond-> ((get-method driver/describe-table-indexes :sql-jdbc) driver database table)
+      (string? pk)
```
calherries, line 91: Why `string?`? What other type could `pk` be?
```suggestion
      (string? pk)
```

src/metabase/driver/sql_jdbc/sync/describe_table.clj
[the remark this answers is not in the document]
calherries: Ah, I see this is dependent on line 301's `when-let `.

`(-> group-by update-vals vals set disj)` can be replaced by `(group-by vals keep set)` and I think this would be easier to understand.

modules/drivers/sqlite/src/metabase/driver/sqlite.clj
qnkhuat: nil

calherries: Then ideally you should use `some?` instead. When I read this, I was questioning why such a specific check was made, and wondering whether the `pk` could be something other than a string. It doesn't help that `get-table-pks`'s docstring doesn't mention the return type.

calherries (commented):
Looks good

calherries (approved):

src/metabase/driver/sql_jdbc/sync/describe_table.clj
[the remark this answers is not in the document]
qnkhuat: nice. did that in https://github.com/metabase/metabase/pull/36665/commits/0213cac7bd033eb6df64b2f3fb06ca145258c6fc

Issue #35551, opened 2023-11-09
Support arbitrary date comparison


alxnddr:
**Problem:** Trend charts can only compare to the previous period - previous minute, previous hour, previous day, etc.. It is not able to skip a period or specify an arbitrary number of periods in the past from which to compare to the current metric value 
  - e.g. Cannot specify 6 hours ago, 7 days ago, 6 months ago, 4 quarters ago, X years ago

<br/>

**Value:** Instead of limiting users to selecting only the previous period, there will now be an option to select an arbitrary number of previous periods ago to compare against. So instead of comparing to last month, now we can compare to 3 months ago or 6 months ago, opening up a range of useful comparison options.

---
**Desired Behavior**/**Acceptance Criteria**

Solution Behaviors
- [ ] Create a data tab in the visualization settings sidebar for the settings:
  - [ ] Primary Number Drop-down (hide if data has only 1 number column)
  - [ ] Comparison Drop-down with Options:
    1. "Previous [period]" (default)
    2. "X [period]s ago"
    3. "Compare to previous" (special case)
       - When the time-series has non-regular periods (28 days, 2 months, etc.), default to comparing to the previous non-null value and add the full data label
- [ ] Add data settings option for comparison types
  - `scalar.comparisons`
    - `{ type: "previousPeriod" }`
    - `{ type: "periodsAgo", value: number }` (number is a positive integer)
    - `{ type: "previousValue" }`

<br />

Validation Tests
- **Unit**
  - ComputeTrend (`compute.unit.spec.js`):
    - [ ] `computeChange`: does it return the correct percentage according to the inputs?
    - [ ] `computeTrend`
      - [ ] Does it behave correctly according to the `scalar.comparisons` setting?
        - [ ] `comparison: previousValue`
        - [ ] `comparison: previousPeriod`
        - [ ] `comparison: periodsAgo`
      - [ ] Does it behave correctly according to the `scalar.field` setting?
  - Utils (`utils.unit.spec.ts`)
    - [ ] Do the `scalar.comparisons` visualization setting functions behave correctly
      - [ ] `getDefaultOption` => `getDefaultComparison`
      - [ ] `options` => `getComparisonOptions`
      - [ ] `isValid` => `isComparisonValid`

- **Integration** 
  - SmartScalar (`SmartScalar.unit.spec.js`)
    - [ ] Does it correctly display the current value and date?
    - [ ] Does it correctly display the comparison percent change, period, and comparison value?
[...]

PR #36666, opened 2023-12-12
Trend Charts: Support arbitrary date comparison

JesseSDevaney:
See issue below for problem statement, value statement, desired behavior & acceptance criteria, and the parent epic
 - Closes: https://github.com/metabase/metabase/issues/35551


JesseSDevaney:
The type errors that are displayed are definitely wrong. My IDE does not show any type errors and correctly displays the function signatures. 

I do not know what to do about besides es-lint ignoring them. Any ideas?

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -1,206 +1,475 @@
...
+    comparisonDescStr,
+    comparisonValue: prevValue,
   };
 }
 
-function computePreviousPeriodComparison({
+function getRowOfPeriodsAgo({
+  prevDate,
+  dateUnit,
+  dateUnitsAgo,
```
JesseSDevaney, line 324: not needed

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -261,15 +275,29 @@ Object.assign(SmartScalar, {
...
+      title: t`Primary number`,
       fieldFilter: isNumeric,
-      getHidden: ([
-        {
-          data: { cols },
-        },
-      ]) => cols.filter(isNumeric).length < 2,
     }),
+    "scalar.comparisons": {
+      section: "Data",
```
JesseSDevaney, line 283:
```suggestion
      section: t`Data`,
```

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
```diff
@@ -1,15 +1,194 @@
...
+      cy.get("input").click().type("3.90293{enter}");
+      cy.get("input").should("have.value", 2);
+
+      // should allow valid input
+      cy.get("input").click().type("3{enter}");
+    });
+    cy.findByTestId("scalar-previous-value").within(() => {
+      cy.findByText("vs. Jan:");
+      cy.findByText("52,249.59");
+    });
```
JesseSDevaney, line 98: @kulyk You'll want to check the new comparison in the E2E test here

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.tsx
```diff
@@ -0,0 +1,225 @@
...
+            size="xs"
+            w="3.5rem"
+            type="number"
+            required
+          ></NumberInputStyled>
+        </Group>
+      </Box>
+    </MenuItemStyled>
+  );
+}
```
JesseSDevaney, line 225: @kulyk You'll want to add the new static number comparison widget here and add it as a MenuItemComponent to the new option you create in [COMPARISON_SELECTOR_OPTIONS in utils.ts](https://github.com/metabase/metabase/pull/36666/files#diff-d33d87b8e2bc0dd1795e40605ac729d1dd938418fc0bd513dd437893bcb91859R92)

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -1,206 +1,475 @@
...
+      display: {
+        percentChange: percentChangeStr,
+        comparisonValue: comparisonValueStr,
+      },
+      percentChange,
+    },
+  };
+}
+
+function computeComparison({ currentMetricData, series, settings }) {
```
JesseSDevaney, line 84: @kulyk You'll want to add the new comparison option in this function. The settings will contain the comparison `const { type, label, value } = settings["scalar.comparisons"];`. For the static/custom value comparison, it is mainly checking that value is a number and if so, returning the comparison data as `{ comparisonDescStr: label, comparisonValue: value }`

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -1,325 +1,1651 @@
...
+            const insights = [{ unit: dateUnit, col: "Count" }];
+
+            expect(() =>
+              computeTrend(series({ rows, cols }), insights, settings),
+            ).toThrow();
+          });
+        });
+      });
+
+      describe(`comparisonType: ${COMPARISON_TYPES.PERIODS_AGO}`, () => {
```
JesseSDevaney, line 508: @kulyk You'll want to create a new describe section (after this one) for the new comparison type to validate the behaviors.

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
 export const getChangeWidth = (width: number): number => {
   return Math.max(width - ICON_SIZE - ICON_MARGIN_RIGHT - 2 * SPACING, 0);
 };
+
+export const COMPARISON_TYPES = {
```
JesseSDevaney, line 83: @kulyk You'll want to add the new comparison type here `CUSTOM_VALUE: "customValue"`

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+  PREVIOUS_VALUE: "previousValue",
+  PREVIOUS_PERIOD: "previousPeriod",
+  PERIODS_AGO: "periodsAgo",
+} as const;
+
+export type ComparisonTypes =
+  typeof COMPARISON_TYPES[keyof typeof COMPARISON_TYPES];
+
+export const COMPARISON_SELECTOR_OPTIONS = {
```
JesseSDevaney, line 92: @kulyk You'll want to add the new selector option here with the various information it needs 

```javascript
CUSTOM_VALUE: {
  type: COMPARISON_TYPES.CUSTOM_VALUE,
  name: t`...`,
  MenuItemComponent: CustomValueSelectorWidget
}
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+type PreviousPeriodMenuOption = {
+  type: typeof COMPARISON_TYPES.PREVIOUS_PERIOD;
+  name: string;
+};
+type PeriodsAgoMenuOption = {
+  type: typeof COMPARISON_TYPES.PERIODS_AGO;
+  name: string;
+  MenuItemComponent: typeof PeriodsAgoInputWidget;
+  maxValue: number;
+};
```
JesseSDevaney, line 121: @kulyk You'll want to create a new menu option type for custom value to declare what properties it will need for the selector

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+export type SelectedComparisonPeriodsAgo = {
+  type: typeof COMPARISON_TYPES.PERIODS_AGO;
+  value: number;
+};
+type SelectedComparisonPreviousPeriod = {
+  type: typeof COMPARISON_TYPES.PREVIOUS_PERIOD;
+};
+type SelectedComparisonCompareToPrevious = {
+  type: typeof COMPARISON_TYPES.PREVIOUS_VALUE;
+};
```
JesseSDevaney, line 136: @kulyk You'll want to add a new selected comparison type to declare what properties the visualization setting will have when saved and when used in the comparison menu selector.

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+    });
+  }
+
+  return createComparisonMenuOption({
+    type: COMPARISON_TYPES.PREVIOUS_PERIOD,
+    dateUnit,
+  });
+}
+
+export function getComparisonOptions(
```
JesseSDevaney, line 168: @kulyk You'll want to add the new custom value option always.

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+      type: typeof COMPARISON_TYPES.PREVIOUS_PERIOD;
+      dateUnit: RelativeDatetimeUnit;
+    }
+  | {
+      type: typeof COMPARISON_TYPES.PERIODS_AGO;
+      maxValue: number;
+      dateUnit: RelativeDatetimeUnit;
+    };
+
+function createComparisonMenuOption(
```
JesseSDevaney, line 291: @kulyk You can use this function to abstract the creation of the custom value menu option for use in the `getComparisonOptions` and `getDefaultOption` functions

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+  const earliestNonEmptyDate = earliestNonEmptyRow[dimensionIndex] as string;
+
+  if (latestNonEmptyDate === null || earliestNonEmptyDate === null) {
+    return null;
+  }
+
+  return dayjs(latestNonEmptyDate).diff(earliestNonEmptyDate, dateUnit);
+}
+
+type GetComparisonMenuOptionParameters =
```
JesseSDevaney, line 277: @kulyk The newly added type here could be something like

```typescript
{ 
  type: typeof COMPARISON_TYPES.CUSTOM_VALUE;
}
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.unit.spec.ts
```diff
@@ -21,6 +32,182 @@ const getAutoPrecisionOptions = (width: number) => {
...
+        );
+
+        expect(defaultComparison).toEqual({
+          type: COMPARISON_TYPES.PREVIOUS_PERIOD,
+          name: "Previous month",
+        });
+      });
+    });
+
+    describe("getComparisonOptions", () => {
```
JesseSDevaney, line 90: @kulyk You'll want to update tests here to include the new custom value option

JesseSDevaney (commented):
@kulyk I left some comments to aid in the understanding and development of the next milestone. Reach out if you have any questions!

frontend/src/metabase-types/api/dataset.ts
```diff
@@ -43,14 +43,22 @@ export interface ResultsMetadata {
...
   rows: RowValues[];
   cols: DatasetColumn[];
+  insights?: Insight[];
+  results_metadata: ResultsMetadata;
   rows_truncated: number;
   requested_timezone?: string;
   results_timezone?: string;
   download_perms?: DownloadPermission;
-  results_metadata: ResultsMetadata;
+  native_form?: any;
```
kulyk, line 61: Can we use a real type here? Not sure about `params` though

```ts
native_form: { query: string }
```
CleanShot 2023-12-13 at 17 19 44@2x

frontend/src/metabase-types/api/dataset.ts
```diff
@@ -43,14 +43,22 @@ export interface ResultsMetadata {
   columns: DatasetColumn[];
 }
 
+export type Insight = {
+  col: string;
+  unit: DatetimeUnit;
+  [key: string]: any;
+};
```
kulyk, line 50: WDYT about either dropping the `any` part or extending the type to match reality? Example:

CleanShot 2023-12-13 at 17 20 23@2x

frontend/src/metabase-types/api/dataset.ts
```diff
@@ -59,9 +67,9 @@ export type JsonQuery = DatasetQuery & {
 
 export interface Dataset {
   data: DatasetData;
-  database_id: DatabaseId;
+  database_id?: DatabaseId;
   row_count: number;
-  running_time: number;
+  running_time?: number;
```
kulyk, line 72: Are these really optional?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.unit.spec.js
```diff
@@ -5,13 +5,33 @@ import { DateTimeColumn, NumberColumn } from "__support__/visualizations";
 import Visualization from "metabase/visualizations/components/Visualization";
 import { getSettingsWidgetsForSeries } from "metabase/visualizations/lib/settings/visualization";
 import registerVisualizations from "metabase/visualizations/register";
+import { COMPARISON_TYPES } from "./utils";
+
+console.error = jest.fn();
```
kulyk, line 10: We probably don't want to commit this

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.unit.spec.js
```diff
@@ -21,222 +41,317 @@ const series = ({ rows, insights, field }) => {
...
+
+      expect(screen.getByText("100")).toBeInTheDocument();
       expect(screen.getByText("Nov 2019")).toBeInTheDocument();
-      const lastChange = screen.getByText("10%");
-      expect(lastChange).toBeInTheDocument();
-      userEvent.hover(lastChange);
+      expect(screen.getByRole("img")).toHaveAttribute(
+        "aria-label",
+        "arrow_up icon",
+      );
```
kulyk, line 104: WDYT about this?

```ts
import { getIcon } from "__support__/ui"

expect(getIcon("arrow_up")).toBeInTheDocument();
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.styled.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Menu, NumberInput } from "metabase/ui";
+
+export const MenuTargetStyled = styled(Menu.Target)`
+  width: 100%;
+
+  .emotion-Button-label {
+    width: 100%;
+    height: 100%;
+  }
+`;
```
kulyk, line 12: Can we avoid using a CSS class here directly? If Mantine changes the class name (and it most likely will be different in v7 since they're dropping emotion), the style will break

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.styled.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const MenuItemStyled = styled(Menu.Item)<{
+  onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
+  isSelected?: boolean;
+  py?: string;
+}>`
```
kulyk, line 18: Can we use `MenuItemProps` here instead?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.styled.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const NumberInputStyled = styled(NumberInput)`
+  .emotion-Input-wrapper {
+    margin-top: 0;
+  }
+
+  .emotion-Input-input {
+    text-align: center;
+  }
+`;
```
kulyk, line 34: Again, could we avoid using Mantine CSS class names directly here?

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+) {
+  const [
+    {
+      data: { insights },
+    },
+  ] = series;
+
+  const dateUnit = insights?.find(
+    insight => insight.col === settings["scalar.field"],
+  )?.unit as RelativeDatetimeUnit | undefined;
```
kulyk, line 154: `insight.unit` has a [`DatetimeUnit` type](https://github.com/metabase/metabase/blob/b4408e033fae30ea08e53037d73ad1dcac45f4dd/frontend/src/metabase-types/api/query.ts#L50) which is way broader than [`RelativeDatetimeUnit`](https://github.com/metabase/metabase/blob/b4408e033fae30ea08e53037d73ad1dcac45f4dd/frontend/src/metabase-types/api/query.ts#L111). IMO type casting here can be dangerous and lead to errors, we should avoid this

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+
+    const { suffix, MenuItemComponent } =
+      COMPARISON_SELECTOR_OPTIONS.PERIODS_AGO;
+
+    return {
+      type,
+      name: `${dateUnit}s ${suffix}`,
+      maxValue,
+      MenuItemComponent,
+    };
```
kulyk, line 318: Is `name` a user-facing string? Should we handle i18n here?

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
```diff
@@ -63,3 +79,244 @@ export const getValueHeight = (height: number): number => {
...
+
+  const latestNonEmptyRow = rows.findLast(row => !isEmpty(row[dimensionIndex]));
+  const earliestNonEmptyRow = rows.find(row => !isEmpty(row[dimensionIndex]));
+
+  if (latestNonEmptyRow === undefined || earliestNonEmptyRow === undefined) {
+    return null;
+  }
+
+  const latestNonEmptyDate = latestNonEmptyRow[dimensionIndex] as string;
+  const earliestNonEmptyDate = earliestNonEmptyRow[dimensionIndex] as string;
```
kulyk, line 268: Should we handle other `RowValue` types?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.tsx
```diff
@@ -0,0 +1,225 @@
...
+            value={inputValue}
+            onChange={(value: number) => setInputValue(value)}
+            onMouseDown={handleChildMouseDownAndUp}
+            onMouseUp={handleChildMouseDownAndUp}
+            onKeyPress={handleEnter}
+            size="xs"
+            w="3.5rem"
+            type="number"
+            required
+          ></NumberInputStyled>
```
kulyk, line 220: I _think_ it should be enough to set `onClick={event => event.stopPropagation()}` to avoid conflicts with menu item clicks?

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
```diff
@@ -1,15 +1,194 @@
...
+    cy.findByTestId("chartsettings-sidebar").findByText("Count").click();
+    popover().within(() => {
+      // should only have two options
+      cy.get('[role="option"]').should("have.length", 2);
+
+      // selected should be highlighted
+      cy.get('[aria-label="Count"]').should(
+        "have.css",
+        "background-color",
+        Color(colors.brand).string(),
```
alxnddr, line 44: E2E specs are not the best option to check styles, I'd suggest to avoid adding such assertions. Instead we could query an element with `aria-selected="true"` and verify it is "Count"

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
```diff
@@ -1,15 +1,194 @@
...
+      );
+
+      // should not be highlighted b/c not selected
+      cy.get('[aria-label="Sum of Total"]').should(
+        "have.css",
+        "background-color",
+        Color().alpha(0).string(),
+      );
```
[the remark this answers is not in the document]
alxnddr: Agree with Anton, querying element by `aria-selected` and verifying it is the "Count" menu item would be sufficient

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -1,206 +1,475 @@
 // eslint-disable-next-line no-restricted-imports -- deprecated usage
```
alxnddr, line 1: Would it be hard to convert it to TS?

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -1,206 +1,475 @@
...
+  dateUnit,
+  dateUnitsAgo,
+  dimensionColIndex,
+  nextValueRowIndex,
   rows,
-  metricIndex,
-  dimensionIndex,
-  nextValue,
+}) {
+  const targetDate = moment.parseZone(prevDate);
```
alxnddr, line 329: We are moving to `dayjs` so it would be nice to use it here instead of moment. Probably, should be easy to replace

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -1,325 +1,1651 @@
...
-      expect(computeChange(-3, -5)).toBe((-5 - -3) / Math.abs(-3));
-      expect(computeChange(-12, -3)).toBe((-3 - -12) / Math.abs(-12));
+
+    describe("comparisonVal < 0", () => {
+      it("should evaluate: - → 0 = 100%", () => {
+        expect(computeChange(-1, 0)).toBe(1);
+        expect(computeChange(-10, 0)).toBe(1);
+      });
+      it("should evaluate: - → - =  (currVal - comparisonVal) / Math.abs(comparisonVal)", () => {
+        expect(computeChange(-3, -5)).toBe((-5 - -3) / Math.abs(-3));
```
alxnddr, line 34: I think when you assert it with an expression like `toBe((-5 - -3) / Math.abs(-3))` the test has become less readable. It should be fine to assert with an expected result value in this case here and below

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -1,325 +1,1651 @@
...
-      expect(computeChange(10, 0)).toBe(-1);
-    });
-    it("should evaluate: + → + = (nextVal - prevVal) / Math.abs(prevVal)", () => {
-      expect(computeChange(3, 5)).toBe((5 - 3) / Math.abs(3));
-      expect(computeChange(12, 3)).toBe((3 - 12) / Math.abs(12));
-    });
-    it("should evaluate: + → - = (nextVal - prevVal) / Math.abs(prevVal)", () => {
-      expect(computeChange(3, -5)).toBe((-5 - 3) / Math.abs(3));
-      expect(computeChange(12, -3)).toBe((-3 - 12) / Math.abs(12));
+    describe("comparisonVal = 0", () => {
```
alxnddr, line 14: Can we utilize `test.each`?

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -1,325 +1,1651 @@
...
-          "should correctly fallback to day unit if backend doesn’t return valid unit",
-        rows: [
-          ["2018-09-01", 100],
-          ["2019-11-01", 300],
-        ],
-        dateUnit: null,
-        expected: "300; Nov 1, 2019; ↑ 200% vs. Sep 1, 2018: 100",
-      },
-    ];
+    const series = ({ rows, cols }) => [{ data: { rows, cols } }];
```
alxnddr, line 60: Can we use `createMockSeries` and other test helpers? It will make it easy converting this to typescript

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -1,325 +1,1651 @@
...
+        "scalar.field": "Count",
+        "scalar.comparisons": { type: comparisonType },
+      };
 
-    it.each(testCases)("$description", ({ rows, expected, dateUnit }) => {
-      const insights = [
-        { unit: dateUnit, col: "Count" },
-        { unit: dateUnit, col: "Sum" },
+      const cols = [
+        DateTimeColumn({ name: "Month" }),
```
alxnddr, line 72: We have typed helper `createMockColumn` so maybe it worth adding createMockDateTimeColumn, createMockNumberColumn and use it here

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.tsx
kulyk, line 1: Could we group `SmartScalarSettingsWidget` files into their own directory?

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
kulyk, line 1: AFAICT, this file contains constants, types, and utils for both `SmartScalar` and `SmartScalarSettingsWidget` components. I wonder if it'd make sense to break it down, so things don't mix up. Maybe even break it down into files like `constants.ts`, `types.ts` and `utils.ts` — I think we used that pattern in other areas of the app

frontend/src/metabase-types/api/dataset.ts
JesseSDevaney: I am going to drop it and leave it as an exercise for someone who ends up working on code that will use those types, since they will have better knowledge of what those types look like and how they are used.

frontend/src/metabase-types/api/dataset.ts
JesseSDevaney: I cannot find a place where `native_form` is even being used on the front-end to check in what cases `params` is used

frontend/src/metabase-types/api/dataset.ts
JesseSDevaney: Thank you for pointing this out. I cannot remember why I did this. Reverting it.

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.tsx
JesseSDevaney: It is enough to do that, but those functions have a little more frustration prevention baked into them.

Demo with `onClick={event => event.stopPropagation()}` set (no extra functionality)

https://github.com/metabase/metabase/assets/22608765/f7a55069-672a-4ee9-835a-acc90996cb1c

It does not select all text when a user clicks on it (to make inputting a new number easier). It causes accidental clicks in some scenarios as well: 
1. mouse down on button, but mouse up on input -> still counts as a click
2. mouse down on input, drags across input to select all, but accidentally mouses up on button -> counts as a click

With the extra logic, it will:
1. Select all the text present when a user clicks on it to allow for easier typing.
3. Prevent accidental clicking if a user tries to highlight all the text but accidentally mouses up outside the input on the button
4. Prevents accidental clicking if the user mouses down on the button but mouses up on the input (meaning the user wanted to alter the value not click the button)

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
JesseSDevaney: I plan to do this in a separate PR

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
JesseSDevaney: I attempted this, but it seems that in the use case I have, moment and dayjs have significantly different behaviors. Dayjs has weird behaviors that do not seem to have been fixed yet (https://github.com/iamkun/dayjs/issues/651, https://github.com/iamkun/dayjs/pull/2060).

**Not Preserving UTC Offsets**

image
image

**Changing UTC Offsets Based on User Time-zone**

_Timezone: UTC -05:00_
image
image

_Timezone: UTC +08:00_
image
image

---

@alxnddr WDYT?

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
JesseSDevaney: Thank you for that catch! I actually meant to type `insight.unit` as `RelativeDatetimeUnit`, not `DatetimeUnit`. So, the type casting is not even necessary... :face_in_clouds:

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
JesseSDevaney: Because we already checked that the column is a date column and checked that the rows we have are non-empty, we [know that the values are going to be strings](https://github.com/metabase/metabase/blob/trend-chart-improvements/milestone-2-support-arbitrary-date-comparison/frontend/src/metabase-lib/types/utils/isa.js#L105).

frontend/src/metabase/visualizations/visualizations/SmartScalar/utils.ts
JesseSDevaney: Yes, it is. I updated them for i18n

JesseSDevaney:
Thanks for all the review comments, @alxnddr and @kulyk! :heart: 

Just pushed up the changes - ready for re-review

JesseSDevaney:
Would either of you know why this type error is occurring?

image

1. It is not present in my IDE
2. It knows it is an array (`RowValues[]` above)
3. It knows what type of array values it holds (`RowValues` above)

**IDE**

Screenshot from 2023-12-19 20-09-32
Screenshot from 2023-12-19 20-09-41

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
```diff
@@ -1,15 +1,190 @@
...
+      },
+      { visitQuestion: true },
+    );
+
+    // scalar.switch_positive_negative setting
+    cy.icon("arrow_down").should(
+      "have.css",
+      "color",
+      Color(colors.error).string(),
+    );
```
kulyk, line 118: My gut tells me we shouldn't be asserting the colors, but I can't come up with a better alternative
@alxnddr WDYT?

frontend/src/metabase-types/api/visualization-settings.ts
```diff
@@ -0,0 +1,16 @@
...
+  type: "previousPeriod";
+};
+type SelectedComparisonCompareToPrevious = {
+  type: "previousValue";
+};
+
+export type SelectedComparison =
+  | SelectedComparisonCompareToPrevious
+  | SelectedComparisonPreviousPeriod
+  | SelectedComparisonPeriodsAgo;
```
kulyk, line 16: nit: I think the word "selected" makes way more sense in the UI context (when used with some input component). Should these names be a bit more generic? Maybe just `ComparisonPreviousPeriod` or `SmartScalarComparisonPreviousPeriod`

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/MenuItem.styled.tsx
```diff
@@ -0,0 +1,17 @@
+import styled from "@emotion/styled";
+import type { MouseEvent } from "react";
+import { Menu } from "metabase/ui";
+
+type MenuItemProps = {
+  onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
+  isSelected?: boolean;
+  py?: string;
+};
```
kulyk, line 9: WDYT about using the `aria-selected` prop instead of `isSelected`?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
```diff
@@ -0,0 +1,105 @@
...
+  selectedValue: SelectedComparison;
+  setOpen: Dispatch<SetStateAction<boolean>>;
+};
+
+function createMenuOption({
+  onChange,
+  optionArgs,
+  setOpen,
+  selectedValue,
+}: CreateMenuOptionProps) {
```
kulyk, line 74: nit: I think `renderMenuOption` would be a more expected name for something retuning JSX

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
```diff
@@ -0,0 +1,105 @@
...
+      <Menu.Target>
+        <ButtonStyled
+          data-testid={"comparisons-widget-button"}
+          disabled={isDisabled}
+        >
+          <Group spacing="sm">
+            {selectedDisplayName}
+            <Icon name="chevrondown" size="14" />
+          </Group>
+        </ButtonStyled>
```
kulyk, line 48: Is there a reason we're not using `Button's` `rightIcon` prop?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.styled.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const ButtonStyled = styled(Button)<ButtonProps>`
+  padding-right: 0;
+  padding-left: 1rem;
+  width: 100%;
+
+  span {
+    width: 100%;
+    height: 100%;
+  }
+`;
```
kulyk, line 17: Have you tried Mantine's `getStylesRef` helper here?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoOptionComponent.tsx
```diff
@@ -0,0 +1,158 @@
...
+  isSelected: boolean;
+  maxValue: number;
+  name: string;
+  onChange: (setting: { type: string; value?: number }) => void;
+  selectedValue?: SelectedComparisonPeriodsAgo;
+  setOpen: Dispatch<SetStateAction<boolean>>;
+  type: string;
+};
+
+export function PeriodsAgoOptionComponent({
```
kulyk, line 23: nit: I think it's clear it's a component and we don't need the word "Component" here and in the filename

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoOptionComponent.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const handleChildMouseDownAndUp = useCallback(
+    (e: MouseEvent<HTMLInputElement>) => {
+      e.stopPropagation();
+      e.preventDefault();
+      inputRef.current?.select();
+
+      mouseDownInChildRef.current = true;
+    },
+    [],
+  );
```
kulyk, line 61: @JesseSDevaney I've noticed Mantine Menu has two [props](https://v6.mantine.dev/core/menu/?t=props) that might help us avoid these hacks:

* `Menu` has a `closeOnItemClick` prop
* `MenuItem` has a `closeMenuOnClick` prop

Can we look into using one of them? FWIW, I've used `MenuItem's` `closeMenuOnClick` in #36951

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoOptionComponent.styled.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from "@emotion/styled";
+import { NumberInput } from "metabase/ui";
+
+export const NumberInputStyled = styled(NumberInput)`
+  input {
+    text-align: center;
+    padding: 0.25rem;
+  }
+`;
```
kulyk, line 9: Can't we do that with Mantine props? 

```tsx
<NumberInput p="0.25rem" ta="center" /> 
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -1,206 +1,501 @@
...
+// compute the percent change between two values (comparisonVal → currVal)
+// percentChange = (currVal - comparisonVal) / Math.abs(comparisonVal)
+export function computeChange(comparisonVal, currVal) {
+  if (comparisonVal === 0) {
+    // percentChange = currVal / 0
+    return currVal === 0 ? 0 : currVal > 0 ? Infinity : -Infinity;
+  }
+
+  return (currVal - comparisonVal) / Math.abs(comparisonVal);
+}
```
kulyk, line 434: The comments mostly duplicate the code, I think we can remove them

kulyk (approved):
@JesseSDevaney, thanks a million for addressing all the feedback so diligently 🙏

Left a few more nit-picky comments, but overall it looks good! I think I'd really want to find another way to prevent the "periods ago" input from closing the `Menu` though. It might be worth merging this as is and addressing other feedback in followup PRs, so we can also work on other milestones in parallel. @JesseSDevaney @alxnddr WDYT?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/MenuItem.styled.tsx
kulyk: Also, I was thinking about using Mantine's `MenuItemProps` type here, so we don't need to declare every prop we want to use inside a styled component. Here's an [example](https://github.com/metabase/metabase/blob/a28d750889e141c02c2862c30066a24453e9730c/frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.styled.tsx#L15) from my branch

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
alxnddr: I think it is fine to check styles in such cases because it is a reflection of the code logic. If possible, would be nice to test in isolation — in jest specs

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/MenuItem.styled.tsx
JesseSDevaney: Thanks for these callouts!

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
JesseSDevaney: Because how the label and `rightIcon` get separated, I do not know how to get the styles right using the `rightIcon` attribute

image

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.styled.tsx
JesseSDevaney: Just did. It works :tada:

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoOptionComponent.styled.tsx
JesseSDevaney: Unfortunately this does not work, since the input is actually wrapped by multiple divs.

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
JesseSDevaney: For now, I switched to using `rightIcon` and setting `justify-content: center` with inline styles, because that is the only pattern that has support across Mantine v6 -> v7. Alternatively we can target the child div for now and make that have `justify-content: center`. 

We can discuss this in a follow up PR.

PR #36670, opened 2023-12-12
Finish removing Honey SQL 1 

### LINKED ISSUE #28156: Migrate `:bigquery-cloud-sdk` driver to Honey SQL 2

### LINKED ISSUE #28161: Migrate `:presto` and `:presto-jdbc` drivers to Honey SQL 2
`:presto` should probably also use `:inline` instead of `unprepare` -- see #28169

### LINKED ISSUE #28165: Migrate `:sqlserver` driver to Honey SQL 2
We shouldn't do this until 49, so we still have at least one driver using Honey SQL 1 so we can test that it still works.

### LINKED ISSUE #28396: Remove Honey SQL 1 as a dependency

### LINKED ISSUE #28684: [Epic] Honey SQL 2 Migration, Continued [47-49]
Splitting off the 47+ work from #28009

```[tasklist]
### In 47 or 48:
- [ ] #28156
- [ ] #28161
- [ ] #28169
- [ ] Make Honey SQL 2 the default `honey-sql-version`. Driver authors can still use Honey SQL 1, but will have to implement this method and return `1` instead of `2` (instead of vice versa). Update driver changelog to mention this
```

Once the 46 stuff is done we can probably move this into a separate issue
```[tasklist]
### In 49:
- [ ] #28165
- [ ] https://github.com/metabase/metabase/issues/28396
```

camsaul:
This PR removes Honey SQL 1 as a dependency, completing the deprecation process we started in 46. It removes the Honey SQL 1 utils namespace and the Honey SQL 1/2 cross-compatible utils namespace, and removes all the conditional code in the SQL QP that handled either version. It converts our last three drivers, SQL Server, BigQuery, and Presto, to Honey SQL 2.

Fixing all the tests is probably going to take a few rounds (especially BigQuery), but this is ready for review now. Changes should be pretty straightforward so despite the PR size this should be an easy review hopefully.

Resolves #28156
Resolves #28161
Resolves #28165
Resolves #28396
Resolves #28684


metamben (approved):
LGTM. It's a lot but it's quite straightforward. 👍

Issue #36668, opened 2023-12-12
[Epic] Bigquery: Sync required partition column


qnkhuat:
**Links**
- product doc: https://www.notion.so/metabase/Support-scanning-on-tables-with-enforced-partitions-in-BigQuery-0b7bce07314c4f20a3ea3dedf3f723ba?pvs=4
- eng doc: https://www.notion.so/metabase/Bigquery-Sync-required-partition-key-c2943ffae97245b98dbb89fe1adf7ad4?pvs=4
- feature branch: `branch-name` _this should be the feature branch where this work will be done in. PRs will be delivered against this branch_
- issue links: https://github.com/metabase/metabase/issues/25752

**Implementation Plan**

```[tasklist]
### Milestone 1:  Sync which table requires a partition filter
- [x] Add new column `metabase_table.database_require_fitler`
- [x] Sync tables that require a filter and set `database_require_filter` to true
```
PR:  https://github.com/metabase/metabase/pull/36677

```[tasklist]
### Milestone 2: Sync partition keys
- [x] Add new column `metabase_field.database_partitioned`
- [x] Sync partition fields and set their `database_partitioned` to true
```

PR: https://github.com/metabase/metabase/pull/36892

```[tasklist]
### Milestone 3: Fix sync for tables that require a partition field
- [x] Fix sync field values
- [x] Fix fingerprint
```

PR: https://github.com/metabase/metabase/pull/37102

```[tasklist]
### ~Milestone 4: Automatically adds the required filter in notebook editor~ (out of scope)
- [x] BE part of https://github.com/metabase/metabase/issues/36983
- [ ] https://github.com/metabase/metabase/issues/37185
```

- [Ref](https://cloud.google.com/bigquery/docs/materialized-views-create#partitioned_materialized_views)

qnkhuat:
milestone 4 is out of the scope of this project, it'll be handled as part of another project in the future.

see https://metaboat.slack.com/archives/C04DN5VRQM6/p1707140273126219?thread_ts=1705615352.637879&cid=C04DN5VRQM6

carstenkadmos:
is there a ticket for milestone 4? still has big impact on our project.

qnkhuat:
That's out of the scope of this project.

You can upvote this [issue](https://github.com/metabase/metabase/issues/37185) and also add a comment describing your use case. It'd help us prioritize!

As a workaround, I recommend creating a model that includes a filter by default for these tables.

carstenkadmos:
[...]

PR #35973, opened 2023-11-21
Move `sample-dataset` into `test-data` for backend tests

kienstra:
Closes https://github.com/metabase/metabase/issues/29573

### Description
In the backend tests, moves `sample-dataset` into `test-data`, and updates tests

|  | Task | What I did | 
|---|---|---|
 | ✅ | Move definitions from `sample-dataset.edn` into `test-data.edn` | Moved [sample-dataset.edn](https://github.com/kienstra/metabase/blob/373f4c4ff27966fa913feb1dfea89b75a0c27425/test/metabase/test/data/dataset_definitions/sample-dataset.edn) verbatim into the end of [test-data.edn](https://github.com/kienstra/metabase/blob/2e2fd689f162e3656bd5aa6c6548fd761823da9e/test/metabase/test/data/dataset_definitions/test-data.edn) |  
 | ✅ |  Remove all the usages of `(mt/dataset sample-dataset ...)` in tests | Removed all of these, and only replaced them with`(mt/dataset test-data` in 3 places where they [were](https://github.com/kienstra/metabase/blame/921169e35ea6d9a82959eae5114d2ec7c0949492/modules/drivers/athena/test/metabase/test/data/athena.clj#L93) [needed](https://github.com/kienstra/metabase/blob/296f1dee8ed617d6c3f258643f5e9f4bed7a656b/test/metabase/api/action_test.clj#L197) | 
 | ✅ | Cycle the `v3_` key for BigQuery (if we're still prefixing it that way; we may change it in the near future) or any other Cloud databases that cache test dataset data loading. Athena will have to be done manually from a REPL.  |  Cycled the [BigQuery](https://github.com/kienstra/metabase/blame/296f1dee8ed617d6c3f258643f5e9f4bed7a656b/modules/drivers/bigquery-cloud-sdk/test/metabase/test/data/bigquery_cloud_sdk.clj#L57) and [Snowflake](https://github.com/kienstra/metabase/blame/296f1dee8ed617d6c3f258643f5e9f4bed7a656b/modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj#L48) keys from `v3_` to `v4_`. Didn't update Athena.  |  
[...]

PR #36677, opened 2023-12-12
Bigquery sync table require filter

qnkhuat:
First milestone of https://github.com/metabase/metabase/issues/36668

Known limitation: this won't work for materialized views, see https://github.com/metabase/metabase/pull/36677/files#diff-370f1464c7c10bd0d0bd84544e6f80102152617fcd99fea6e9bd4dbaff4903caR131-R135 for why

To read more about partitioned tables, see https://cloud.google.com/bigquery/docs/partitioned-tables


src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -189,32 +209,32 @@
   ([database :- i/DatabaseInstance db-metadata]
    ;; determine what's changed between what info we have and what's in the DB
    (let [db-tables               (table-set db-metadata)
-         our-metadata            (our-metadata database)
-         strip-desc              (fn [metadata]
-                                   (set (map #(dissoc % :description) metadata)))
+         name+schema->db-table   (m/index-by (juxt :name :schema) db-tables)
```
qnkhuat, line 212: I took a stab at refactoring this sync step a bit.
Details in the following comments

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -189,32 +209,32 @@
...
      (when (some? (:version db-metadata))
        (sync-util/with-error-handling (format "Error creating/reactivating tables for %s"
                                               (sync-util/name-for-logging database))
          (update-database-metadata! database db-metadata)))
      ;; create new tables as needed or mark them as active again
      (when (seq new-tables)
-       (sync-util/with-error-handling (format "Error creating/reactivating tables for %s"
-                                              (sync-util/name-for-logging database))
-         (create-or-reactivate-tables! database new-tables)))
+       (let [new-tables-info (set (map #(get name+schema->db-table [(:name %) (:schema %)]) new-tables))]
```
qnkhuat, line 226: previously, when creating a table, we didn't include a table description; we relied on the update step to do that. that's necessary, so I've changed it so that the description is included in the creation step.

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -189,32 +209,32 @@
...
      ;; mark old tables as inactive
      (when (seq old-tables)
        (sync-util/with-error-handling (format "Error retiring tables for %s" (sync-util/name-for-logging database))
          (retire-tables! database old-tables)))
 
-     ;; update description for changed tables
-     (when (seq changed-tables)
-       (sync-util/with-error-handling (format "Error updating table description for %s" (sync-util/name-for-logging database))
-         (update-table-description! database changed-tables)))
+     (sync-util/with-error-handling (format "Error updating table metadata for %s" (sync-util/name-for-logging database))
```
qnkhuat, line 235: previously, only table description could be updated during sync step, but I've changed it so that it could extend to other properties too.

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -140,27 +142,47 @@
...
+   metabase-table :- (ms/InstanceOf :model/Table)]
+  (log/infof "Updating table metadata for %s" (sync-util/name-for-logging metabase-table))
+  (let [to-update-keys [:description :database_require_filter]
+        old-table      (select-keys metabase-table to-update-keys)
+        new-table      (select-keys (merge
+                                     (zipmap to-update-keys (repeat nil))
+                                     table-metadata)
+                                    to-update-keys)
+        [_ changes _]  (data/diff old-table new-table)
+        changes        (cond-> changes
```
qnkhuat, line 164: this is a specical case, check test `metabase.sync.sync-metadata.comments-test/sync-existing-table-comment-test` and `metabase.sync.sync-metadata.comments-test/dont-overwrite-table-custom-description-test` to know how it's expected to work.

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -189,32 +209,32 @@
   ([database :- i/DatabaseInstance db-metadata]
    ;; determine what's changed between what info we have and what's in the DB
    (let [db-tables               (table-set db-metadata)
-         our-metadata            (our-metadata database)
-         strip-desc              (fn [metadata]
-                                   (set (map #(dissoc % :description) metadata)))
+         name+schema->db-table   (m/index-by (juxt :name :schema) db-tables)
+         our-metadata            (db->our-metadata database)
+         keep-table-id-info      (fn [metadata]
```
qnkhuat, line 214: This change somehow changes the order of sample dataset tables added on startup. 

that's why you'll see some tests affected by this, one of which is I have to update the table ides in `e2e/support/cypress_data.js`

test/metabase/sync/sync_metadata/comments_test.clj
```diff
@@ -114,7 +117,14 @@
       (mt/dataset (basic-table "table_with_comment_after_sync" nil)
         ;; modify the source DB to add the comment and resync
         (driver/notify-database-updated driver/*driver* (mt/db))
-        (tx/create-db! driver/*driver* (basic-table "table_with_comment_after_sync" "added comment"))
+        (tx/create-db! driver/*driver* (basic-table "table_with_comment_after_sync" nil))
```
qnkhuat, line 120: This test relied on the fact that descriptions are not included when inserting tables, the correct behavior is to start with a table without a description, then sync, then add a description, then sync. see https://github.com/metabase/metabase/pull/36677/files#r1427842665 for context.

test/metabase/automagic_dashboards/core_test.clj
```diff
@@ -185,8 +185,8 @@
   (mt/with-test-user :rasta
     (automagic-dashboards.test/with-dashboard-cleanup
       (doseq [[table cardinality] (map vector
-                                       (t2/select Table :db_id (mt/id) {:order-by [[:id :asc]]})
-                                       [15 11 7 5])]
+                                       (t2/select Table :db_id (mt/id) {:order-by [[:name :asc]]})
```
qnkhuat, line 188: There are 2 things happening here:
- this is a consequence of test tables ids being changed ([context](https://github.com/metabase/metabase/pull/36677/files#r1427855029)).
- after https://github.com/metabase/metabase/pull/35973, the test-data database now has 8 tables. so that's why you see my changes also include 4 more items.

e2e/support/cypress_data.js
```diff
@@ -25,15 +25,15 @@ export const SAMPLE_DB_SCHEMA_ID = "1:PUBLIC";
 // Use only for e2e helpers and custom commands. Never in e2e tests directly!
```
qnkhuat, line 25: see https://github.com/metabase/metabase/pull/36677/files#r1427855029 for why I have this change

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -140,27 +142,47 @@
...
+        old-table      (select-keys metabase-table to-update-keys)
+        new-table      (select-keys (merge
+                                     (zipmap to-update-keys (repeat nil))
+                                     table-metadata)
+                                    to-update-keys)
+        [_ changes _]  (data/diff old-table new-table)
+        changes        (cond-> changes
+                         ;; we only update the description if and only if the initial state is nil
+                         ;; because don't want to override the user edited description if it exists
+                         (some? (:description old-table))
```
calherries, line 167: Before, the check was `(str/blank? ...)` not `(some? ...)`. Why did you change it?

calherries (changes requested):

src/metabase/sync/sync_metadata/tables.clj
qnkhuat: it's a function so I think the name should contains a verb

how about `keep-schema+name-set`?

src/metabase/sync/sync_metadata/tables.clj
qnkhuat: we can't fetch the value from the diff because the diff only contains `schema + name`, whereas `create-or-reactivate-tables` need to get the full info from `db-tables`.

src/metabase/sync/sync_metadata/tables.clj
qnkhuat: I think that was a mistake. If a table does not have a description, then the jdbc will return `NULL`.

whereas if a table has a description and it's empty, then it might be a case where the user manually removes the description in Metabase.

In this case, we don't want to override that on update.

test/metabase/sync/sync_metadata/comments_test.clj
```diff
@@ -114,7 +117,14 @@
...
+        ;; create the comment
+        (jdbc/execute! (sql-jdbc.conn/db->pooled-connection-spec (mt/db))
+                       [(sql.tx/standalone-table-comment-sql
+                         driver/*driver*
+                         {:name "table_with_comment_after_sync_db"}
+                         {:table-name "table_with_comment_after_sync"
+                          :table-comment "added comment"})])
```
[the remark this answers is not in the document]
qnkhuat: addresed in cb4955805e

test/metabase/sync/sync_metadata/tables_test.clj
```diff
@@ -39,3 +43,20 @@
...
+(deftest sync-table-update-info-of-new-table-added-during-sync-test
+  (testing "during sync, if a table is reactived, make sure we do updates the table info if needed"
+    (mt/dataset (mt/dataset-definition "sync-retired-table"
+                  ["user" [{:field-name "name" :base-type :type/Text}] [["Ngoc"]]])
+      (t2/update! :model/Table (mt/id :user) {:active false})
+      ;; table description is changed
+      (jdbc/execute! (sql-jdbc.conn/db->pooled-connection-spec (mt/db))
+                     [(sql.tx/standalone-table-comment-sql
+                       (:engine (mt/db))
+                       {:name "sync-retired-table"}
```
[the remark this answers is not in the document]
qnkhuat: cb4955805e

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -189,32 +209,32 @@
...
-                                              (sync-util/name-for-logging database))
-         (create-or-reactivate-tables! database new-tables)))
+       (let [new-tables-info (set (map #(get name+schema->db-table [(:name %) (:schema %)]) new-tables))]
+         (sync-util/with-error-handling (format "Error creating/reactivating tables for %s"
+                                                (sync-util/name-for-logging database))
+           (create-or-reactivate-tables! database new-tables-info))))
      ;; mark old tables as inactive
      (when (seq old-tables)
        (sync-util/with-error-handling (format "Error retiring tables for %s" (sync-util/name-for-logging database))
          (retire-tables! database old-tables)))
```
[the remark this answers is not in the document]
qnkhuat: yea you're right. this sync code is quite old, so I think a lot of things need to be improved here.

anw I applied your suggestion here cb4955805e

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -140,27 +142,47 @@
...
+    (when (seq changes)
+      (t2/update! :model/Table (:id metabase-table)  changes))))
+
+(mu/defn ^:private update-tables-metadata-if-needed!
+  [table-metadatas :- [:set i/DatabaseMetadataTable]
+   metabase-tables :- [:set (ms/InstanceOf :model/Table)]]
+  (let [name+schema->table-metadata (m/index-by (juxt :name :schema) table-metadatas)
+        name+schema->metabase-table (m/index-by (juxt :name :schema) metabase-tables)]
+    (doseq [[name+schema metabase-table] name+schema->metabase-table]
+      (when-let [table-metadata (name+schema->table-metadata name+schema)]
```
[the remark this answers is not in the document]
qnkhuat: cb4955805e

e2e/test/scenarios/admin/datamodel/hide_tables.cy.spec.js
```diff
@@ -16,15 +16,15 @@ describe("scenarios > admin > datamodel > hidden tables (metabase#9759)", () =>
 
   it("hidden table should not show up in various places in UI", () => {
     // Visit the main page, we shouldn't be able to see the table
-    cy.visit(`/browse/${PRODUCTS_ID}`);
+    cy.visit(`/browse/${SAMPLE_DB_ID}`);
```
qnkhuat, line 19: this was a mistake and it accidentally works because `producs_id` and `sample_db_id` had the same id (=1)

src/metabase/sync/sync_metadata/tables.clj
calherries: Yes, that's why I passed the diff output through `(->> <diff> (map schema+name->table))`

Sorry for the poor formatting I was taking shortcuts to write this in the github UI.

Here's the correct formatting:
```
[new-tables old-tables] (->> (data/diff (set (map schema+name db-tables)) (set (map schema+name our-metadata)))
                             (map schema+name->table))
```
Of course you could do this in two steps. My main point is that `schema+name` can be extracted out into a function and used to do the diff and getting the full info.

src/metabase/sync/sync_metadata/tables.clj
calherries: Function names don't need to contain verbs, but sure `keep-schema+name-set` is better too.

src/metabase/sync/sync_metadata/tables.clj
calherries: Makes sense!

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
```diff
@@ -240,6 +249,60 @@
                    {:source-table (mt/id view-name)
                     :order-by     [[:asc (mt/id view-name :id)]]}))))))))
 
+(deftest sync-table-with-required-filter-test
```
calherries, line 252: I think you should add a test for materialized views too, based on this line 
https://github.com/metabase/metabase/pull/36677/files#diff-370f1464c7c10bd0d0bd84544e6f80102152617fcd99fea6e9bd4dbaff4903caR130

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -136,31 +140,52 @@
...
+  (log/infof "Updating table metadata for %s" (sync-util/name-for-logging metabase-table))
+  (let [to-update-keys [:description :database_require_filter]
+        old-table      (select-keys metabase-table to-update-keys)
+        new-table      (select-keys (merge
+                                     (zipmap to-update-keys (repeat nil))
+                                     table-metadata)
+                                    to-update-keys)
+        [_ changes _]  (data/diff old-table new-table)
+        changes        (cond-> changes
+                         ;; we only update the description if and only if the initial state is nil
```
calherries, line 169: redundant "only"
```suggestion
                         ;; we only update the description if the initial state is nil
```

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
qnkhuat: right so turns out we currently can't figure out whether a materialized view have required a filter or not. See https://github.com/metabase/metabase/pull/36677/files#diff-370f1464c7c10bd0d0bd84544e6f80102152617fcd99fea6e9bd4dbaff4903caR131-R135 for comments.

I'll add this as a limitation in the PR

src/metabase/sync/sync_metadata/tables.clj
qnkhuat: did that here https://github.com/metabase/metabase/pull/36677/commits/ce46551884f37acbd888e40e00d1d377d738df2b#diff-8a4b5a6ee44180ba51e170fe36de9d5f3f889cd95b1ee56d19c72b6aac130ae5R215-R222

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
calherries: Makes sense, thanks for investigating

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -123,6 +115,34 @@
...
+(defmethod driver/describe-database :bigquery-cloud-sdk
+  [_ database]
+  (let [tables (list-tables (:details database))]
+    {:tables (set (for [^Table table tables
+                        :let  [^TableId         table-id   (.getTableId table)
+                               ^String          dataset-id (.getDataset table-id)
+                               ^TableDefinition tabledef   (.getDefinition table)
+                               table-name                  (str (.getTable table-id))]]
+                    {:schema                  dataset-id
+                     :name                    #pp table-name
```
calherries, line 127: You left a #pp here

calherries (approved):
Solid work

calherries (changes requested):
Actually sorry, I think you're missing a test case for a table that is initially created without a required filter, then synced, then is modified to have a required filter, then synced again. Similar to `sync-existing-table-comment-test`

e2e/test/scenarios/admin/datamodel/hide_tables.cy.spec.js
nemanjaglumac: 🤦‍♂️

qnkhuat:
@calherries 

>  I think you're missing a test case for a table that is initially created without a required filter, then synced, then is modified to have a required filter, then synced again. Similar to sync-existing-table-comment-test

This is not possible in bigquery because whether a table requires partition or not is a property it's created with. afaik You can't change it after.

I could add sth like `sync-table-update-info-of-new-table-added-during-sync-test` tho.

calherries:
Based on the docs ([here](https://cloud.google.com/bigquery/docs/samples/bigquery-update-table-require-partition-filter#bigquery_update_table_require_partition_filter-java) and [here](https://www.javadoc.io/doc/com.google.apis/google-api-services-bigquery/v2-rev20200719-1.30.10/com/google/api/services/bigquery/model/Table.html#setRequirePartitionFilter-java.lang.Boolean-)), it looks like you can update the property after the table is created.

calherries:
Also using SQL ([source](https://stackoverflow.com/questions/59636355/bigquery-how-to-set-require-partition-filter-by-tablereference)):
```
ALTER TABLE IF EXISTS mydataset.newtable
SET OPTIONS(
    require_partition_filter = true
)
```

qnkhuat:
Good find! I was looking at the list of options for `ALTER TABLE SET OPTIONS` command([docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#column_set_options_list)) and didn't find this option so I thought it's not possible.

anw I added the test here 8452b0b2a2

calherries (approved):
Great, thanks!

PR #36693, opened 2023-12-12
Trim space and normalize empty string in admin settings

### LINKED ISSUE #25482: Whitelabel landing page string allows whitespace which leads to "page not found"
**Describe the bug**
It is currently possible to enter a simple whitespace for a custom landing page string, which leads to "Page not found". The problem is that this is hard to notice/fix.

**Logs**
Please include javascript console and server logs around the time this bug occurred. For information about how to get these, consult our [bug troubleshooting guide](https://metabase.com/docs/latest/troubleshooting-guide/bugs.html)

**To Reproduce**
1. Run EE version
2. Go to `/admin/settings/whitelabel`
3. For the "landing page" setting, enter a whitespace or `/   ` folowed by a few whitespaces, then click outside the input field to blur it
4. Exit admin and see that the page cannot be found with the message "We're a little lost...
The page you asked for couldn't be found."

**Expected behavior**
We should trim the whitespace because there's probably no scenario in which a URL would have it and be valid.

local dev EE, `master`, 2c7178eac84f8edf6d7c9ffaa45675aa131d3a93, H2, Sample Database

**Severity**
P3 but it can get hard to figure out what's going on.

WiNloSt:
Closes https://github.com/metabase/metabase/issues/25482

### Description

Unfortunately we don't have an API to normalize the input, maybe it's the time to introduce one, but that has more wide affect to all admin widgets, so I don't want to do that there unless we've discussed in with other FE engineers.

### How to verify

Follow the repro step from #25482

### Demo

https://www.loom.com/share/9cd422c864154dd0b96c4a81e7f0b907

### Checklist


enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
```diff
@@ -67,7 +68,7 @@ if (hasPremiumFeature("whitelabel")) {
         {
           key: "landing-page",
           display_name: t`Landing Page`,
-          type: "string",
+          widget: NormalizedTextSettingInput,
```
WiNloSt, line 71: I'm not sure if it's a good idea to have some kind of APIs that we could normalize each input. We do have `onBeforeChanged`
https://github.com/metabase/metabase/blob/dc56106904ba82a4575d1437d85e906b74f8fcfa/frontend/src/metabase/admin/settings/app/components/SettingsEditor/SettingsEditor.jsx#L92-L97
and `onChanged`, but that doesn't do what I wanted

https://github.com/metabase/metabase/blob/dc56106904ba82a4575d1437d85e906b74f8fcfa/frontend/src/metabase/admin/settings/app/components/SettingsEditor/SettingsEditor.jsx#L105-L110

I want a kind of API that compute the value before passing it to `onChange`. Maybe we could call it `normalize` or something, but I don't want to do that right now before discussing with other FE enginners.

frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
```diff
@@ -10,7 +10,7 @@ const getValue = (value: string, type: string) => {
   return value;
 };
 
-interface SettingInputProps {
+export interface SettingInputProps {
```
WiNloSt, line 13: used in `frontend/src/metabase/admin/settings/components/widgets/NormalizedTextSettingInput/NormalizedTextSettingInput.tsx`

npretto:
> Maybe we could call it normalize or something, but I don't want to do that right now before discussing with other FE enginners.

I agree with that idea of making a normalize function, this seems something that should be a prop/config and not a new component.
I also understand that before changing the settings behaviour we should discuss it with at least the webapp team.

I have a suggestion that is a bit in between, a normalize function that can be passed via `props`, this doesn't require a new component so it's less invasive. If other the rest of the FE team/webapp team likes it we could make it "first level" so that we don't have to pass it via the `props`  field.

What do you think? It should be easy to keep your tests by passing the prop accordingly

```diff
diff --git a/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js b/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
index 9a512cfdfb..3b208e97c0 100644
--- a/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
+++ b/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
@@ -69,6 +69,9 @@ if (hasPremiumFeature("whitelabel")) {
           display_name: t`Landing Page`,
           type: "string",
           placeholder: "/",
+          props: {
+            normalize: value => value.trim(), // will be a function exported from somehwere
+          },
         },
         {
           key: "loading-message",
diff --git a/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx b/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
index bec211dd03..0c725a5152 100644
--- a/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
+++ b/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
@@ -23,6 +23,7 @@ interface SettingInputProps {
   errorMessage?: string;
   id?: string;
   type?: string;
+  normalize?: (value: string | number | null) => string | number | null;
 }
 
 const SettingInput = ({
@@ -33,10 +34,11 @@ const SettingInput = ({
   fireOnChange,
   id,
   type = "text",
+  normalize = value => value,
 }: SettingInputProps) => {
   const changeHandler = (e: { target: HTMLInputElement }) => {
     const value = getValue(e.target.value, type);
-    onChange(value);
+    onChange(normalize(value));
   };
 
   return (
```

WiNloSt:
> > Maybe we could call it normalize or something, but I don't want to do that right now before discussing with other FE enginners.
> 
> I agree with that idea of making a normalize function, this seems something that should be a prop/config and not a new component. I also understand that before changing the settings behaviour we should discuss it with at least the webapp team.
> 
> I have a suggestion that is a bit in between, a normalize function that can be passed via `props`, this doesn't require a new component so it's less invasive. If other the rest of the FE team/webapp team likes it we could make it "first level" so that we don't have to pass it via the `props` field.
> 
> What do you think? It should be easy to keep your tests by passing the prop accordingly
> 
> ```diff
> diff --git a/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js b/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
> index 9a512cfdfb..3b208e97c0 100644
> --- a/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
> +++ b/enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
> @@ -69,6 +69,9 @@ if (hasPremiumFeature("whitelabel")) {
>            display_name: t`Landing Page`,
>            type: "string",
>            placeholder: "/",
> +          props: {
> +            normalize: value => value.trim(), // will be a function exported from somehwere
> +          },
>          },
>          {
>            key: "loading-message",
> diff --git a/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx b/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
> index bec211dd03..0c725a5152 100644
> --- a/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
> +++ b/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx
> @@ -23,6 +23,7 @@ interface SettingInputProps {
>    errorMessage?: string;
>    id?: string;
>    type?: string;
> +  normalize?: (value: string | number | null) => string | number | null;
>  }
>  
>  const SettingInput = ({
> @@ -33,10 +34,11 @@ const SettingInput = ({
>    fireOnChange,
>    id,
>    type = "text",
> +  normalize = value => value,
>  }: SettingInputProps) => {
>    const changeHandler = (e: { target: HTMLInputElement }) => {
>      const value = getValue(e.target.value, type);
> -    onChange(value);
> +    onChange(normalize(value));
>    };
>  
>    return (
> ```

I like the idea! I'll do that. Thanks!

WiNloSt:
@npretto I've tried your suggestion in https://github.com/metabase/metabase/pull/36693/commits/209cec45c7bf0691a7748c0ef8ca32131047a5bb, but it seems, there's a problem when typing space that the input box won't be updated, I'm not totally sure why this wasn't a problem with the preveius commit. I'll take a look at that tomorrow.

npretto (approved):
Left a small comment but look good 👍

frontend/src/metabase/admin/settings/components/widgets/SettingInput/SettingInput.tsx
```diff
@@ -45,6 +54,7 @@ const SettingInput = ({
         SettingsInput: type !== "password",
         SettingsPassword: type === "password",
       })}
+      normalize={normalizeWithProps}
```
WiNloSt, line 57: The reason we need to also pass `normalize` to the inner component is because it's possible for `SettingInputBlurChange` to have different value than this component. e.g. if we have `/` and we start adding more spaces `SettingInputBlurChange` would have a state of `/              ` while the normalize function would strip it down to `/` again, this means the`value` prop passed to it would be the same, and it won't know when it should copy the prop to its `internalState`.
https://github.com/metabase/metabase/blob/6b1fab0f3b2a31e84743d99e47a314f9d0463eb2/frontend/src/metabase/components/InputBlurChange/InputBlurChange.tsx#L31

Passing `normalize` ensures we always normalize the `internalState` as well so that it would stay in sync with `setting.value`

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
```diff
@@ -69,6 +69,16 @@ if (hasPremiumFeature("whitelabel")) {
           display_name: t`Landing Page`,
           type: "string",
           placeholder: "/",
+          props: {
+            normalize(value, { type }) {
+              if (type === "text" && typeof value === "string") {
```
npretto, line 74:
```suggestion
              if (typeof value === "string") {
```
is it the same type defined a few lines above? If so we already know it's text in this case, and I think we should be fine with trimming all string values

frontend/src/metabase/admin/settings/components/widgets/SettingInput/SettingInput.tsx
```diff
@@ -10,33 +11,41 @@ const getValue = (value: string, type: string) => {
...
   onChange,
   autoFocus,
   errorMessage,
   fireOnChange,
   id,
   type = "text",
+  normalize = value => value,
 }: SettingInputProps) => {
+  const normalizeWithProps = useCallback(
+    (value: Value) => normalize(value, { type }),
```
npretto, line 43: if we remove the `type` check in the other comment, this could be removed

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
WiNloSt: The reason is that there are more than just text inputs e.g. number input, but all of them will have [`value` as a string](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#value). So I detect that from a `type` which tells what kind of input this will be instead.

image

And the value could also be `null`
https://github.com/metabase/metabase/blob/e21e425b011b134328cd26e794981bcfa18f568c/frontend/src/metabase/admin/settings/components/widgets/SettingInput.tsx#L7

so we need to check the type of the value as well before attempting to `trim()` it.

frontend/src/metabase/admin/settings/components/widgets/SettingInput/SettingInput.tsx
WiNloSt: Replied here https://github.com/metabase/metabase/pull/36693#discussion_r1429862552. 

I don't think we can remove it.

Let me know what you think.

enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
npretto: I was referring to removing `type === "text"`, since we're passing this function directly to that specific widget, which is using a type="string" (at the setting level) it will use `type = "text"` as default (at the widget level).
That logic would make sense if we passed that function to all the settings by default,

WiNloSt: Ahh yeah, good point. This could all be simplified. Thanks for clarifying 🙏

WiNloSt: Fixed in https://github.com/metabase/metabase/pull/36693/commits/8a26f5420498ad4a8992ebcc21bb6465e3fe0b01

frontend/src/metabase/admin/settings/components/widgets/SettingInput/SettingInput.tsx
WiNloSt: Fixed in https://github.com/metabase/metabase/pull/36693/commits/8a26f5420498ad4a8992ebcc21bb6465e3fe0b01

frontend/src/metabase/admin/settings/components/widgets/SettingInput/SettingInput.tsx
```diff
@@ -10,33 +10,39 @@ const getValue = (value: string, type: string) => {
...
   autoFocus?: boolean;
   fireOnChange?: boolean;
   errorMessage?: string;
   id?: string;
   type?: string;
+  normalize?: (value: Value) => Value;
 }
 
-const SettingInput = ({
+const identity = (value: Value) => value;
```
WiNloSt, line 31: Previously I was inlining this in the component prop, but that will create a new function every time, so it was incorrect before. This change ensures that every render, the `normalize` would be the same across renders if caller doesn't pass the `normalize` prop

PR #36694, opened 2023-12-12
test(whats-new-notification): add test case for white labeling scenario

npretto:
### Description

While reviewing [Fix whitelabeled application name replaces Metabase in admin settings](https://github.com/metabase/metabase/pull/36032#top) I realized this test case was missing so I added it.

### DIFF
--- a/frontend/src/metabase/nav/components/WhatsNewNotification/WhatsNewNotification.unit.spec.tsx
+++ b/frontend/src/metabase/nav/components/WhatsNewNotification/WhatsNewNotification.unit.spec.tsx
@@ -53,22 +53,24 @@ const setup = ({
   setupPropertiesEndpoints(createMockSettings());
   setupSettingsEndpoints([createMockSettingDefinition()]);
 
+  const mockState = createMockState({
+    settings: mockSettings({
+      version: versionMock,
+      "version-info": createMockVersionInfo({ latest, older }),
+      "last-acknowledged-version": lastAcknowledged,
+      "application-name": isWhiteLabeling ? "My App" : "Metabase",
+      "token-features": createMockTokenFeatures({
+        whitelabel: isWhiteLabeling,
+      }),
+    }),
+  });
+
   if (isWhiteLabeling) {
     setupEnterprisePlugins();
   }
 
   return renderWithProviders(<WhatsNewNotification></WhatsNewNotification>, {
-    storeInitialState: createMockState({
-      settings: mockSettings({
-        version: versionMock,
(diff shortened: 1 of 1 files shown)


frontend/src/metabase/nav/components/WhatsNewNotification/WhatsNewNotification.unit.spec.tsx
```diff
@@ -47,16 +49,25 @@ const setup = ({
   const versionMock = createMockVersion({ tag: currentVersion });
 
   const [latest, ...older] = versions;
-  const mockSettings = createMockSettingsState({
+  const mockedSettings = mockSettings({
```
WiNloSt, line 52: I'd suggest you inline the value at line 70 instead because the use of **mockedSomething** is very rare in our codebase (try a text search with `mocked` and you'll see).

frontend/src/metabase/nav/components/WhatsNewNotification/WhatsNewNotification.unit.spec.tsx
```diff
@@ -47,16 +49,25 @@ const setup = ({
...
     version: versionMock,
     "version-info": createMockVersionInfo({ latest, older }),
     "last-acknowledged-version": lastAcknowledged,
+    "application-name": isWhiteLabeling ? "My App" : "Metabase",
+    "token-features": createMockTokenFeatures({
+      whitelabel: true,
```
WiNloSt, line 58: since you also conditionally set the `application-name`, you should do the same here.

```suggestion
      whitelabel: isWhiteLabeling,
```

WiNloSt (approved):
I have minor comments that would be great to address before merging but otherwise, this looks good to me.

frontend/src/metabase/nav/components/WhatsNewNotification/WhatsNewNotification.unit.spec.tsx
npretto: Ah, apparently plugins need to be setup *after* `mockSettings` has been called: https://github.com/metabase/metabase/blob/82cf4594ca8516f9d72aada296c25306f4b684a8/frontend/test/__support__/enterprise.js#L2-L3 (the test does indeed fail otherwise)
Seems I won't be able to inline it, I can take out the whole state though and just use the usual mock prefix that we use

WiNloSt: yeah that works. we also do `state = createMockState(...)` in many places

Issue #34112, opened 2023-09-26
[MLv2] Filters


ranquild:
Epic https://github.com/metabase/metabase/issues/28689
Design doc https://www.notion.so/metabase/Filters-API-e411d9d4f267454ba970a76b3f9ecd08?pvs=4#b778d2dcddb14ac6b5082eaca5f69642

### Prior work

```[tasklist]
### Tasks
- [x] #28692
- [x] https://github.com/metabase/metabase/issues/29867
- [x] #28703
- [x] https://github.com/metabase/metabase/issues/31676
- [x] https://github.com/metabase/metabase/issues/31677
- [x] https://github.com/metabase/metabase/issues/32058
- [x] https://github.com/metabase/metabase/issues/32060
- [x] https://github.com/metabase/metabase/issues/32159
- [ ] https://github.com/metabase/metabase/issues/32160
- [ ] https://github.com/metabase/metabase/issues/32409
- [ ] https://github.com/metabase/metabase/issues/33022
- [ ] https://github.com/metabase/metabase/issues/33023
```

### Milestone 1 - FilterPopover and notebook editor

Create MLv2-based `FilterPopover` and use it in the notebook editor.

Integration branch [34112-filters-milestone1](https://github.com/metabase/metabase/pull/34191)

```[tasklist]
- [ ] https://github.com/metabase/metabase/issues/34144
- [ ] https://github.com/metabase/metabase/issues/34190
- [ ] https://github.com/metabase/metabase/issues/34130
- [ ] https://github.com/metabase/metabase/issues/34269
- [ ] https://github.com/metabase/metabase/issues/34131
- [ ] https://github.com/metabase/metabase/issues/34134
- [ ] https://github.com/metabase/metabase/issues/34135
- [ ] https://github.com/metabase/metabase/issues/34245
- [ ] https://github.com/metabase/metabase/issues/34224
- [ ] https://github.com/metabase/metabase/issues/34274
- [ ] https://github.com/metabase/metabase/issues/34275
- [ ] https://github.com/metabase/metabase/issues/34378
- [ ] https://github.com/metabase/metabase/issues/34379
- [ ] https://github.com/metabase/metabase/issues/34511
- [ ] https://github.com/metabase/metabase/issues/34495
- [ ] https://github.com/metabase/metabase/issues/34496
- [ ] https://github.com/metabase/metabase/issues/34669
- [ ] https://github.com/metabase/metabase/issues/34332
[...]

PR #36696, opened 2023-12-12
[MLv2] [FE] Filter modal segments

ranquild:
Epic https://github.com/metabase/metabase/issues/34112

Changes:
- Adds segments to the filter modal. They are available in the first group (tab).
- Make sure they can be filtered by using the top-level search in the modal. 
- Converts multi-value pickers to `MultiSelect` from `mantine`. It's not styled yet, this is TBD separately. Also handling field values TBD in a separate PR.

How to verify:
- You can add a segment as a filter and remove it from the query. Query should be modified and work correctly.

Screenshot 2023-12-12 at 16 23 00


frontend/src/metabase/querying/components/FilterModal/CoordinateFilterEditor/CoordinateFilterEditor.tsx
```diff
@@ -85,23 +84,26 @@ export function CoordinateFilterEditor({
...
   onChange,
 }: NumberValueInputProps) {
   if (hasMultipleValues) {
     return (
-      <FilterValuePicker value={values} column={column} onChange={onChange} />
+      <FilterValuePicker
+        values={values.map(value => String(value))}
+        placeholder={t`Enter a number`}
+        getCreateLabel={query => (isFinite(Number(query)) ? query : null)}
+        onChange={values => onChange(values.map(value => Number(value)))}
```
kamilmielnik, line 105: `Number("") === 0`, it's possible to enter whitespace and it will be converted to 0.
Previously spaces were not converted to zeros.

I think it'd be better to use `parseInt` and filter out `NaN`s (in `onChange`).

frontend/src/metabase/ui/components/inputs/index.ts
```diff
@@ -4,6 +4,7 @@ export * from "./DateInput";
 export * from "./DatePicker";
 export * from "./FileInput";
 export * from "./Input";
+export * from "./MultiSelect";
```
kamilmielnik, line 7: Using the same value twice results in the following React warning:

```
Warning: Encountered two children with the same key, `9`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

image

Previously it was not possible to enter the same value twice.

frontend/src/metabase/querying/components/FilterModal/utils/filters.ts
```diff
@@ -63,6 +78,37 @@ export function removeFilters(query: Lib.Query) {
...
+  return segmentItems.reduce((query, { segment, stageIndex }) => {
+    return Lib.filter(query, stageIndex, segment);
+  }, query);
+}
+
+export function removeSegmentFilters(
+  query: Lib.Query,
+  segmentItems: SegmentItem[],
+) {
+  const filterGroups = segmentItems.flatMap(
```
kamilmielnik, line 94:
```suggestion
  const filterGroups = segmentItems.map(
```

frontend/src/metabase/querying/components/FilterModal/utils/search.ts
```diff
@@ -11,19 +11,29 @@ export function searchGroupItems(
   searchText: string,
 ): GroupItem[] {
   const searchValue = searchText.toLowerCase();
+  const searchValueHasSegments = t`segments`.includes(searchValue);
```
kamilmielnik, line 14: It took me a while to figure out why the value isn't ```searchValue.includes(t`segments`)```, but it was the variable name that was confusing.

```suggestion
  const isSearchingForSegments = t`segments`.includes(searchValue);
```

frontend/src/metabase/querying/components/FilterModal/CoordinateFilterEditor/CoordinateFilterEditor.tsx
```diff
@@ -85,23 +84,26 @@ export function CoordinateFilterEditor({
...
-  column,
   values,
   valueCount,
   hasMultipleValues,
   onChange,
 }: NumberValueInputProps) {
   if (hasMultipleValues) {
     return (
-      <FilterValuePicker value={values} column={column} onChange={onChange} />
+      <FilterValuePicker
```
kamilmielnik, line 101: Coordinates used to be formatted like this:

image

Now they're not.
Is this going to be handled in a separate PR?

frontend/src/metabase/querying/components/FilterModal/SegmentFilterEditor/SegmentFilterEditor.tsx
```diff
@@ -0,0 +1,55 @@
+import { useMemo } from "react";
```
kamilmielnik, line 1: I removed a segment and went back to the query builder - the removed segment was still available in the filter modal (until I refreshed a page).

In `master` this does not happen.

frontend/src/metabase/ui/components/inputs/index.ts
ranquild: https://github.com/metabase/metabase/issues/36748

frontend/src/metabase/querying/components/FilterModal/CoordinateFilterEditor/CoordinateFilterEditor.tsx
ranquild: https://github.com/metabase/metabase/issues/36749

frontend/src/metabase/querying/components/FilterModal/utils/filters.ts
ranquild: Fixed

frontend/src/metabase/querying/components/FilterModal/CoordinateFilterEditor/CoordinateFilterEditor.tsx
ranquild: Fixed

ranquild:
@kamilmielnik I've addressed simple things and create separate issues for bigger issues. Please take a look

frontend/src/metabase/querying/components/FilterModal/CoordinateFilterEditor/CoordinateFilterEditor.tsx
```diff
@@ -85,23 +84,26 @@ export function CoordinateFilterEditor({
...
+        values={values.map(value => String(value))}
+        placeholder={t`Enter a number`}
+        getCreateLabel={query => (isFinite(parseFloat(query)) ? query : null)}
+        onChange={values => onChange(values.map(value => Number(value)))}
```
kamilmielnik, line 105:
```suggestion
        onChange={values => onChange(values.map(value => parseFloat(value)).filter(value => isFinite(value)))}
```

Same comment for 3 other places

kamilmielnik (approved):

frontend/src/metabase/querying/components/FilterModal/CoordinateFilterEditor/CoordinateFilterEditor.tsx
ranquild: I'll address that in the next PR where the component is changed

PR #36701, opened 2023-12-12
Clear error message when unable to decrypt encrypted json field

### LINKED ISSUE #31553: [Serdes v2] Error parsing JSON when exporting data model
### Describe the bug

Then exporting the data model, some exports fail with JSON parsing errors.

### To Reproduce

1. Get the Audit AppDB creds on [1pass](https://start.1password.com/open/i?a=AEO4BXOSHNFCJLK26XA7AD522E&h=my.1password.com&i=7iueth2ylmucfjkea34wei5nja&v=jc42hdh6wgip5i74fuxzhtqe2i)
2. Serialize this instance `java -jar metabase.jar export audit_instance`
3. See errors on console

### Expected behavior

It should export all fields fine

### Logs

```
2023-06-12 18:14:12,054 INFO models.serialization :: Extracting Database 34
2023-06-12 18:14:12,148 ERROR models.interface :: Error parsing JSON
com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'LqgfJr6zyrxMuI': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (StringReader); line: 1, column: 15]
```

and

```
2023-06-12 18:11:15,073 INFO models.serialization :: Extracting Field 1,388
2023-06-12 18:11:15,195 ERROR models.interface :: Error parsing JSON
com.fasterxml.jackson.core.JsonParseException: Unexpected character ('m' (code 109)): Expected space separating root-level values
 at [Source: (StringReader); line: 1, column: 3]
```

```JSON
Running on `master`
```

### Severity

P2

### Additional context

_No response_

sansolovyov:
I've changed `maybe-decrypt` signature to remove this fiddling with argument, since it's never passed a secret key in real code (not in tests).

This logic will not help with other encrypted fields, but they at least won't die with a non-understandable message about being unable to parse JSON. Even though this message is not reporting which model/field is encrypted, it at least guides you to a thought that you shouldn't serialize encrypted DB with no key.

Resolves #31553

### DIFF
--- a/src/metabase/models/interface.clj
+++ b/src/metabase/models/interface.clj
@@ -17,7 +17,7 @@
    [metabase.util :as u]
    [metabase.util.cron :as u.cron]
    [metabase.util.encryption :as encryption]
-   [metabase.util.i18n :refer [trs tru]]
+   [metabase.util.i18n :refer [tru]]
    [metabase.util.log :as log]
    [metabase.util.malli.registry :as mr]
    [methodical.core :as methodical]
@@ -140,7 +140,7 @@
     (try
       (json/parse-string s keywordize-keys?)
       (catch Throwable e
-        (log/error e (str (trs "Error parsing JSON")))
+        (log/error e "Error parsing JSON")
         s))
     s))
 
@@ -225,9 +225,17 @@
   "Serialize encrypted json."
   (comp encryption/maybe-encrypt json-in))
 
-(def encrypted-json-out
+(defn encrypted-json-out
   "Deserialize encrypted json."
-  (comp json-out-with-keywordization encryption/maybe-decrypt))
+  [v]
+  (try
+    (-> v encryption/maybe-decrypt (json/parse-string true))
+    (catch Throwable e
+      (if (or (encryption/possibly-encrypted-string? v)
+              (encryption/possibly-encrypted-bytes? v))
+        (log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
+        (log/error e "Error parsing JSON"))  ; same message as in `json-out`
+      v)))
 
 ;; cache the decryption/JSON parsing because it's somewhat slow (~500µs vs ~100µs on a *fast* computer)
 ;; cache the decrypted JSON for one hour
--- a/src/metabase/models/interface.clj
+++ b/src/metabase/models/interface.clj
@@ -228,14 +228,15 @@
 (defn encrypted-json-out
   "Deserialize encrypted json."
   [v]
-  (try
-    (-> v encryption/maybe-decrypt (json/parse-string true))
-    (catch Throwable e
-      (if (or (encryption/possibly-encrypted-string? v)
-              (encryption/possibly-encrypted-bytes? v))
-        (log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
-        (log/error e "Error parsing JSON"))  ; same message as in `json-out`
-      v)))
+  (let [decrypted (encryption/maybe-decrypt v)]
+    (try
+      (json/parse-string decrypted true)
+      (catch Throwable e
+        (if (or (encryption/possibly-encrypted-string? decrypted)
+                (encryption/possibly-encrypted-bytes? decrypted))
+          (log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
+          (log/error e "Error parsing JSON"))  ; same message as in `json-out`
+        v))))
 
 ;; cache the decryption/JSON parsing because it's somewhat slow (~500µs vs ~100µs on a *fast* computer)
 ;; cache the decrypted JSON for one hour
--- a/src/metabase/util/encryption.clj
+++ b/src/metabase/util/encryption.clj
@@ -147,22 +147,22 @@
   "If `MB_ENCRYPTION_SECRET_KEY` is set and `v` is encrypted, decrypt `v`; otherwise return `s` as-is. Attempts to check
   whether `v` is an encrypted String, in which case the decrypted String is returned, or whether `v` is encrypted bytes,
   in which case the decrypted bytes are returned."
-  {:arglists '([secret-key? s & {:keys [log-errors?], :or {log-errors? true}}])}
+  {:arglists '([secret-key? s & [{:keys [log-errors?], :or {log-errors? true}}]])}
   [& args]
-  (let [[secret-key & more]        (if (and (bytes? (first args)) (string? (second args))) ;TODO: fix hackiness
+  ;; secret-key as an argument so that tests can pass it directly without using `with-redefs` to run in parallel
+  (let [[secret-key v options]     (if (and (bytes? (first args)) (string? (second args)))
                                      args
                                      (cons default-secret-key args))
-        [v & options]              more
         {:keys [log-errors?]
-         :or   {log-errors? true}} (apply hash-map options)
+         :or   {log-errors? true}} options
         log-error-fn (fn [kind ^Throwable e]
                        (when log-errors?
                          (log/warn (trs "Cannot decrypt encrypted {0}. Have you changed or forgot to set MB_ENCRYPTION_SECRET_KEY?"
-                                        kind)
+                                     kind)
                                    (.getMessage e)
                                    (u/pprint-to-str (u/filtered-stacktrace e)))))]
 
-    (cond (not (some? secret-key))
+    (cond (nil? secret-key)
           v
 
           (possibly-encrypted-string? v)
--- a/src/metabase/util/encryption.clj
+++ b/src/metabase/util/encryption.clj
@@ -154,9 +154,9 @@
                              args
                              (cons default-secret-key args))
         log-error-fn (fn [kind ^Throwable e]
-                       (log/errorf e
-                                   "Cannot decrypt encrypted %s. Have you changed or forgot to set MB_ENCRYPTION_SECRET_KEY?"
-                                   kind))]
+                       (log/warnf e
+                                  "Cannot decrypt encrypted %s. Have you changed or forgot to set MB_ENCRYPTION_SECRET_KEY?"
+                                  kind))]
 
     (cond (nil? secret-key)
           v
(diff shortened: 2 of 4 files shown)


luizarakaki (approved):

src/metabase/models/interface.clj
```diff
@@ -140,7 +140,7 @@
     (try
       (json/parse-string s keywordize-keys?)
       (catch Throwable e
-        (log/error e (str (trs "Error parsing JSON")))
```
qnkhuat, line 143: what's a weird line of code lol

src/metabase/util/encryption.clj
```diff
@@ -147,22 +147,22 @@
...
         log-error-fn (fn [kind ^Throwable e]
                        (when log-errors?
                          (log/warn (trs "Cannot decrypt encrypted {0}. Have you changed or forgot to set MB_ENCRYPTION_SECRET_KEY?"
-                                        kind)
+                                     kind)
                                    (.getMessage e)
                                    (u/pprint-to-str (u/filtered-stacktrace e)))))]
 
-    (cond (not (some? secret-key))
+    (cond (nil? secret-key)
```
piranha, line 165: @qnkhuat that one impressed me too :)

src/metabase/models/interface.clj
```diff
@@ -225,9 +225,17 @@
...
+(defn encrypted-json-out
   "Deserialize encrypted json."
-  (comp json-out-with-keywordization encryption/maybe-decrypt))
+  [v]
+  (try
+    (-> v encryption/maybe-decrypt (json/parse-string true))
+    (catch Throwable e
+      (if (or (encryption/possibly-encrypted-string? v)
+              (encryption/possibly-encrypted-bytes? v))
+        (log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
```
calherries, line 236: The caught error could be thrown from `parse-string`, not `maybe-decrypt`. In that case the error message would be wrong. Here's a test that passes and demonstrates the bug:
```clojure
(is (=? [[:error JsonParseException "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?"]]
        (tu.log/with-log-messages-for-level :error
          (mi/encrypted-json-out (encryption/maybe-encrypt "{\"a\": 1")))))
```

test/metabase/models/interface_test.clj
```diff
@@ -130,3 +133,11 @@
                (migrate {:pie.show_legend          legend
                          :pie.show_legend_perecent percent
                          :pie.show_data_labels     labels})))))))
+
+(deftest encrypted-data-with-no-secret-test
+  (is (= {:a 1}
+         (mi/encrypted-json-out "{\"a\": 1}")))
```
calherries, line 139: I can't tell what the behaviour is you're trying to test is in both these tests.

Without `testing` descriptions, we have to do mind-reading when reading tests to figure out what exactly is being covered, or extra work to guess it from the code.

calherries (changes requested):

src/metabase/models/interface.clj
piranha: It does not pass, because it logs `Unexpected end-of-input: expected close marker ....`. Also, the thrown error is always coming from `parse-string` and not `maybe-decrypt`, since latter one doesn't throw. It's works like "no key? okay, you get your string back", and changing that leads to too many changes elsewhere.

So this seems like a least invasive approach right now.

test/metabase/models/interface_test.clj
piranha: Oh, right, I went through too many ideas with this change and this test remained in the worst condition. Adding description now...

src/metabase/models/interface.clj
calherries: The test passes. Take note of the `=?`, maybe you are using `=`

`Unexpected end-of-input: expected close marker ....` is the exception message, which is logged as well, which is different from the logged message, which is being tested here.

I think you might be confused because the log includes the logged message, and the message of the `JsonParseException`. e.g. this is what is logged when I run the test:
```
2023-12-20 16:08:37,418 ERROR models.interface :: Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?
com.fasterxml.jackson.core.io.JsonEOFException: Unexpected end-of-input: expected close marker for Object (start marker at [Source: (StringReader); line: 1, column: 1])
 at [Source: (StringReader); line: 1, column: 8]
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportInvalidEOF(ParserMinimalBase.java:697)
        at com.fasterxml.jackson.core.base.ParserBase._handleEOF(ParserBase.java:512)
        at com.fasterxml.jackson.core.base.ParserBase._eofAsNextChar(ParserBase.java:529)
```

test/metabase/models/interface_test.clj
calherries: Looks good now, thanks

src/metabase/models/interface.clj
calherries: My setup is the same, and I see that too. Again, the error message is different from the logged message. In that screenshot you can only see the error message, but the logged message is off the screen (after the stack trace)

src/metabase/util/encryption.clj
```diff
@@ -147,22 +147,22 @@
   "If `MB_ENCRYPTION_SECRET_KEY` is set and `v` is encrypted, decrypt `v`; otherwise return `s` as-is. Attempts to check
   whether `v` is an encrypted String, in which case the decrypted String is returned, or whether `v` is encrypted bytes,
   in which case the decrypted bytes are returned."
-  {:arglists '([secret-key? s & {:keys [log-errors?], :or {log-errors? true}}])}
+  {:arglists '([secret-key? s & [{:keys [log-errors?], :or {log-errors? true}}]])}
```
calherries, line 150: Why was this changed? It doesn't look like `log-errors?` is used by callers anywhere so we should be able to remove this argument anyway.

src/metabase/util/encryption.clj
```diff
@@ -147,22 +147,22 @@
...
+  (let [[secret-key v options]     (if (and (bytes? (first args)) (string? (second args)))
                                      args
                                      (cons default-secret-key args))
-        [v & options]              more
         {:keys [log-errors?]
-         :or   {log-errors? true}} (apply hash-map options)
+         :or   {log-errors? true}} options
         log-error-fn (fn [kind ^Throwable e]
                        (when log-errors?
                          (log/warn (trs "Cannot decrypt encrypted {0}. Have you changed or forgot to set MB_ENCRYPTION_SECRET_KEY?"
```
calherries, line 160: Isn't it sufficient to change this from `log/warn` to `log/error`?

I tried this and the test passes
```
diff --git a/src/metabase/models/interface.clj b/src/metabase/models/interface.clj
index 891c2e4362..a644be3977 100644
--- a/src/metabase/models/interface.clj
+++ b/src/metabase/models/interface.clj
@@ -234,7 +234,7 @@
       (catch Throwable e
         (if (or (encryption/possibly-encrypted-string? decrypted)
                 (encryption/possibly-encrypted-bytes? decrypted))
-          (log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
+          nil #_(log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
           (log/error e "Error parsing JSON"))  ; same message as in `json-out`
         v))))
 
diff --git a/test/metabase/models/interface_test.clj b/test/metabase/models/interface_test.clj
index 5e2b59b22e..55275f2c5b 100644
--- a/test/metabase/models/interface_test.clj
+++ b/test/metabase/models/interface_test.clj
@@ -145,8 +145,8 @@
              (mi/encrypted-json-out
               (encryption/encrypt (encryption/secret-key->hash "qwe") "{\"a\": 1}"))))))
   (testing "Logs an error message when incoming data looks encrypted"
-    (is (=? [[:error JsonParseException "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?"]]
-            (tu.log/with-log-messages-for-level :error
+    (is (=? [[:warn nil #"^Cannot decrypt encrypted String\. Have you changed or forgot to set MB_ENCRYPTION_SECRET_KEY\? Message seems corrupt or manipulated\.[\s\S]*"]]
+            (tu.log/with-log-messages-for-level :warn
               (mi/encrypted-json-out
                (encryption/encrypt (encryption/secret-key->hash "qwe") "{\"a\": 1}"))))))
```

src/metabase/util/encryption.clj
piranha: Hmmm, good question, I'm going to remove that.

src/metabase/util/encryption.clj
piranha: Not really, it gets into `(nil? secret-key)` branch and never does anything. The problem here is not that key is wrong, it's that it's not present.

src/metabase/models/interface.clj
```diff
@@ -225,9 +225,18 @@
...
   "Deserialize encrypted json."
-  (comp json-out-with-keywordization encryption/maybe-decrypt))
+  [v]
+  (let [decrypted (encryption/maybe-decrypt v)]
+    (try
+      (json/parse-string decrypted true)
+      (catch Throwable e
+        (if (or (encryption/possibly-encrypted-string? decrypted)
+                (encryption/possibly-encrypted-bytes? decrypted))
+          (log/error e "Could not decrypt encrypted field! Have you forgot to set MB_ENCRYPTION_SECRET_KEY?")
```
calherries, line 237: The problem is now that this message will effectively be logged twice when the log level is set to `warn`.

I feel like the correct thing to do here is to skip logging errors in the maybe-decript step, and instead only log errors here. So that means reverting this change you just made (sorry)
https://github.com/metabase/metabase/pull/36701/files#diff-af1ac78232eccb809814746d4ffc7edfc00b17d340caf897b62992384f79acdaR150

piranha: Do you think it's a problem? :-)

calherries: I can't think of any downsides other than it's a little strange to have two checks that are almost the same, and logging slightly different messages. I guess we can let this slide for practicality sake. But on the other hand it's easily avoidable

piranha: Not really easily, except for checking for secret key in `models.interface` which I really don't want to do, or stopping using `maybe-decrypt` in favor of `decrypt` dying. I've tried that, but I feel like it changes behavior too much.

calherries (approved):

PR #36704, opened 2023-12-12
Retry failed Cypress tests only once

nemanjaglumac:
This PR lowers the number of Cypress internal retries to only one attempt.


iethree (approved):
🔥

Issue #35614, opened 2023-11-10
[Epic] Allow appending more data to CSV uploads


luizarakaki:
**Links**
- [product doc](https://www.notion.so/metabase/Allow-appends-on-existing-CSV-upload-models-f1f13864632d4cfd83522fce762890eb#58380830c6ca45df923e4cacc4af1964)
- [eng doc](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df)

```[tasklist]
### Milestone 0: Append from Model Page
- [ ] https://github.com/metabase/metabase/pull/36640
- [ ] https://github.com/metabase/metabase/pull/36705
- [ ] https://github.com/metabase/metabase/issues/36776
- [ ] https://github.com/metabase/metabase/pull/36950
- [ ] https://github.com/metabase/metabase/pull/36990
- [ ] https://github.com/metabase/metabase/pull/36995
- [ ] https://github.com/metabase/metabase/pull/37052
- [ ] https://github.com/metabase/metabase/pull/38274
- [ ] https://github.com/metabase/metabase/pull/37220
```

```[tasklist]
### Milestone 1: Append from Collection Page
- [ ] https://github.com/metabase/metabase/pull/38322
- [ ] https://github.com/metabase/metabase/issues/38435
```

calherries:
Moved outstanding follow-up tasks to another epic https://github.com/metabase/metabase/issues/38853

PR #36705, opened 2023-12-12
CSV appends: Handle _mb_row_id columns

calherries:

This PR implements special behaviour for `_mb_row_id` columns for CSV appends.

It corresponds to the following section of the eng doc ([block link](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df?pvs=4#e6549ef0587747039b7e9fc8a0e09ea3)):

### Special behaviour around _mb_row_id:

- If a CSV file with `_mb_row_id` is uploaded, we will ignore the `_mb_row_id` column in the CSV
- If a CSV file without `_mb_row_id` is uploaded, we will create the column anyway
- This column is assumed to be auto-incrementing. We use the max value of this to roll back changes when failures occur during insertion
- [x]  Test: If a CSV file with `_mb_row_id` is uploaded to a table doesn’t have it, we will ignore the `_mb_row_id` column in the CSV but create the column anyway
- [x]  Test: If a CSV file without `_mb_row_id` is uploaded and the table doesn’t have it either, we will create the column anyway
- [x]  Test: If a CSV file without `_mb_row_id` is uploaded and the table does have it, we can upload just fine (it’s not required)

### DIFF
--- a/src/metabase/driver/sql_jdbc.clj
+++ b/src/metabase/driver/sql_jdbc.clj
@@ -142,14 +142,12 @@
   [driver db-id table-name column-names values]
   (let [table-name (keyword table-name)
         columns    (map keyword column-names)
-        ;; There's nothing magic about 100, but it felt good in testing. There could well be a better number.
-        chunks     (partition-all (or driver/*insert-chunk-rows* 100) values)
         sqls       (map #(sql/format {:insert-into table-name
                                       :columns     columns
                                       :values      %}
                                      :quoted true
                                      :dialect (sql.qp/quote-style driver))
-                        chunks)]
+                        (partition-all 100 values))]
     ;; We need to partition the insert into multiple statements for both performance and correctness.
     ;;
     ;; On Postgres with a large file, 100 (3.76m) was significantly faster than 50 (4.03m) and 25 (4.27m). 1,000 was a
--- a/src/metabase/driver/sql_jdbc.clj
+++ b/src/metabase/driver/sql_jdbc.clj
@@ -142,12 +142,14 @@
   [driver db-id table-name column-names values]
   (let [table-name (keyword table-name)
         columns    (map keyword column-names)
+        ;; There's nothing magic about 100, but it felt good in testing. There could well be a better number.
+        chunks     (partition-all (or driver/*insert-chunk-rows* 100) values)
         sqls       (map #(sql/format {:insert-into table-name
                                       :columns     columns
                                       :values      %}
                                      :quoted true
                                      :dialect (sql.qp/quote-style driver))
-                        (partition-all 100 values))]
+                        chunks)]
     ;; We need to partition the insert into multiple statements for both performance and correctness.
     ;;
     ;; On Postgres with a large file, 100 (3.76m) was significantly faster than 50 (4.03m) and 25 (4.27m). 1,000 was a
--- a/src/metabase/driver/sql_jdbc.clj
+++ b/src/metabase/driver/sql_jdbc.clj
@@ -158,6 +158,17 @@
     (doseq [sql sqls]
       (qp.writeback/execute-write-sql! db-id sql))))
 
+(defmethod driver/add-columns! :sql-jdbc
+  [driver db-id table-name col->type]
+  (let [table-name (keyword table-name)
+        sql        (first (sql/format {:alter-table (keyword table-name)
+                                       :add-column (map (fn [[name type-spec]]
+                                                          (vec (cons name type-spec)))
+                                                        col->type)}
+                                      :quoted true
+                                      :dialect (sql.qp/quote-style driver)))]
+    (qp.writeback/execute-write-sql! db-id sql)))
+
 (defmethod driver/syncable-schemas :sql-jdbc
   [driver database]
   (sql-jdbc.execute/do-with-connection-with-options
--- a/src/metabase/driver/sql_jdbc.clj
+++ b/src/metabase/driver/sql_jdbc.clj
@@ -160,13 +160,12 @@
 
 (defmethod driver/add-columns! :sql-jdbc
   [driver db-id table-name col->type]
-  (let [table-name (keyword table-name)
-        sql        (first (sql/format {:alter-table (keyword table-name)
-                                       :add-column (map (fn [[name type-spec]]
-                                                          (vec (cons name type-spec)))
-                                                        col->type)}
-                                      :quoted true
-                                      :dialect (sql.qp/quote-style driver)))]
+  (let [sql (first (sql/format {:alter-table (keyword table-name)
+                                :add-column (map (fn [[name type-spec]]
+                                                   (vec (cons name type-spec)))
+                                                 col->type)}
+                               :quoted true
+                               :dialect (sql.qp/quote-style driver)))]
     (qp.writeback/execute-write-sql! db-id sql)))
 
 (defmethod driver/syncable-schemas :sql-jdbc
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -3,18 +3,14 @@
    [clj-bom.core :as bom]
    [clojure.data :as data]
    [clojure.data.csv :as csv]
-   [clojure.java.jdbc :as jdbc]
    [clojure.string :as str]
    [flatland.ordered.map :as ordered-map]
    [flatland.ordered.set :as ordered-set]
-   [honey.sql :as sql]
    [java-time.api :as t]
    [medley.core :as m]
    [metabase.analytics.snowplow :as snowplow]
    [metabase.api.common :as api]
    [metabase.driver :as driver]
-   [metabase.driver.sql-jdbc.connection :as sql-jdbc.conn]
-   [metabase.driver.sql.query-processor :as sql.qp]
    [metabase.driver.sync :as driver.s]
    [metabase.driver.util :as driver.u]
    [metabase.mbql.util :as mbql.u]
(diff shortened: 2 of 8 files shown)


src/metabase/upload.clj
```diff
@@ -216,6 +217,12 @@
   "The lower-case name of the auto-incrementing PK column. The actual name in the database could be in upper-case."
   "_mb_row_id")
 
+(defn- get-auto-pk-column [table-id]
+  (some (fn [field]
+          (when (= (normalize-column-name (:name field)) auto-pk-column-name)
+            field))
+        (t2/select :model/Field :table_id table-id)))
```
qnkhuat, line 224:
```suggestion
        (t2/select :model/Field :table_id table-id :active true)))
```

src/metabase/upload.clj
```diff
@@ -533,59 +539,75 @@
...
+          auto-pk-type (driver/upload-type->database-type driver ::auto-incrementing-int-pk)
+          create-auto-pk? (not table-has-auto-pk?)]
       (try
         (let [parsed-rows (parse-rows rows)]
-          (driver/insert-into! (driver.u/database->driver database)
+          (when create-auto-pk?
+            (driver/add-columns! driver
+                                 (:id database)
+                                 (table-identifier table)
+                                 {:_mb_row_id auto-pk-type}))
```
qnkhuat, line 601:
```suggestion
                                 {(keyword auto-pk-column-name) auto-pk-type}))
```

src/metabase/upload.clj
```diff
@@ -533,59 +539,75 @@
...
   [database table file]
   (with-open [reader (bom/bom-reader file)]
     (let [[header & rows] (csv/read-csv reader)
-          {:keys [parse-rows table-col-names]} (check-schema table header)]
+          {:keys [parse-rows table-col-names table-has-auto-pk?]} (check-schema table header)
+          driver (driver.u/database->driver database)
+          auto-pk-type (driver/upload-type->database-type driver ::auto-incrementing-int-pk)
+          create-auto-pk? (not table-has-auto-pk?)]
       (try
         (let [parsed-rows (parse-rows rows)]
```
qnkhuat, line 596: just realized that we are realizing the whole "rows" here. 
I think it'd be nice if `driver/insert-into!` take a lazy seq and returns the number of items inserted. this way we don't have to have the whole seq realized before hand.

This is not a blocker a but a potential TODO

src/metabase/upload.clj
```diff
@@ -533,59 +539,75 @@
...
+                                 (table-identifier table)
+                                 {:_mb_row_id auto-pk-type}))
+          (driver/insert-into! driver
                                (:id database)
                                (table-identifier table)
                                table-col-names
                                parsed-rows)
+          (scan-and-sync-table! database table)
+          (when create-auto-pk?
+            (let [auto-pk-field (get-auto-pk-column (:id table))]
```
qnkhuat, line 609: didn't we have a comment somewhere explaining why we need to do this? could you keep it here?

src/metabase/upload.clj
qnkhuat, line 601:
```suggestion
                                 {:_mb_row_id (driver/upload-type->database-type driver ::auto-incrementing-int-pk)}))
```

It's only used once and is used conditionally, IMO it'll reduce the number of binding I need to keep track while reading this function.

src/metabase/upload.clj
qnkhuat: also, wdyt if we handle this in `scan-and-sync-table!`?

src/metabase/upload.clj
```diff
@@ -533,59 +539,75 @@
...
-                                         (set normalized-field-names))]
-
+                                         ;; ignore the auto-pk field, it will be filled automatically
+                                         (disj (set normalized-field-names) auto-pk-column-name))]
+    ;; check for duplicates
+    (when (some #(< 1 %) (vals (frequencies normalized-header)))
+      (throw (ex-info (tru "The CSV file contains duplicate column names.")
+                      {:status-code 422})))
     (if (and (empty? extra) (empty? missing))
       {:parse-rows (fn [rows]
```
qnkhuat, line 575: This is duplicated with what we have in `detect-schema`.

it makes me think about whether we should stop having these functions returning a lexical scoped function like this.

WDYT about having a function
`(defn parse-rows [col->types rows excluded-columns-idx])`

this way we can decouple schema checking and parsing the rows.

qnkhuat (changes requested):
Have a couple of suggestions.

Also you mentioned
> This column is assumed to be auto-incrementing. We use the max value of this to roll back changes when failures occur during insertion

I don't see this being handled anywhere. Is it part of this PR?

src/metabase/upload.clj
calherries: You have a point, ideally these shouldn't be coupled and duplicated logic can be separated. But let's think about this. `check-schema` is doing a lot here, and its logic can be split.

Schema checking and parsing the rows are coupled in the sense that: 
(1) both the header and rows need to have their auto-pk columns removed.
(2) the schema checking uses the types of fields to know what parser to apply to each column

Creating `(defn parse-rows [col->types rows excluded-columns-idx])` might help reduce some coupling because the coupling from (1) is more explicitly in the form of the `excluded-columns-idx` argument. But they're still coupled by the shared decision to remove the auto-pk columns. To decouple them fully, I could remove the auto-pk column from both the `header` argument of `schema-check` and the `rows` argument of `parse-rows`.

```
(defn auto-pk-col-indices [header])
(defn schema-check [table-fields header-without-auto-pk-columns])
(defn parse-rows [col-upload-types rows-without-auto-pk-columns])
```

Then have them called like so in `append-csv!*`:

```clojure
          excluded-columns-indices (auto-pk-col-indices header)
          header' (remove-indices excluded-columns-indices header)
          table-fields (t2/select :model/Field :table_id (:id table) :active true)
          table-has-auto-pk? (contains? fields-by-normed-name auto-pk-column-name)
          fields-by-normed-name (m/index-by (comp normalize-column-name :name) fields)
          col-upload-types (map (comp base-type->upload-type :base_type fields-by-normed-name normalize-column-name) header)
          _ (check-schema table-fields header) ;; all this does is raise errors
          ... then later
          parsed-rows (parse-rows col-upload-types rows)
```

Now (1) and (2) are both outside `check-schema`.

src/metabase/upload.clj
calherries: Leaving a comment is a good idea, but I don't think we should put this in `scan-and-sync-table!` because they seem logically separate, they happen in different contexts for different reasons. I actually want to move `scan-and-sync-table!` to a sync namespace, as we discussed on the last PR. And there is a difference between the two cases it's used: when we create the CSV table the auto-pk column is updated unconditionally, and here it's updated only on the condition that it wasn't present before.

calherries:
> We use the max value of this to roll back changes when failures occur during insertion

No, that will be implemented with type mismatches exception handling.

src/metabase/upload.clj
qnkhuat: awesome, I think this is much better :)

calherries:
@qnkhuat I've implemented those tidy ups but MySQL tests are still faillling because of this 
[Uploaded CSVs with datetime fields create TIMESTAMP columns instead of DATETIME](https://github.com/metabase/metabase/issues/36761#top)

I will fix that test before this PR merges to the feature branch, either by fixing the bug or by temporarily excluding the datetime column from the test. Probably the latter

src/metabase/upload.clj
```diff
@@ -533,60 +525,54 @@
...
                       {:status-code 422})))))
 
 (defn- append-csv!*
   [database table file]
   (with-open [reader (bom/bom-reader file)]
-    (let [[header & rows] (csv/read-csv reader)
-          {:keys [parse-rows table-col-names]} (check-schema table header)]
+    (let [[header & rows]       (rows-without-auto-pk-columns (csv/read-csv reader))
+          driver                (driver.u/database->driver database)
+          fields-by-normed-name (m/index-by (comp normalize-column-name :name)
```
qnkhuat, line 553:
```suggestion
          normed-name->field (m/index-by (comp normalize-column-name :name)
```

I think this is what we often use when naming maps

src/metabase/upload.clj
```diff
@@ -340,29 +293,69 @@
...
+                  (when (pred item)
+                    idx))
+                coll))
+
+(defn- auto-pk-column-indices
+  "Returns the indices of columns that have the same normalized name as [[auto-pk-column-name]]"
+  [header]
+  (set (indices-where #(= auto-pk-column-name (normalize-column-name %)) header)))
+
+(defn- rows-without-auto-pk-columns
```
qnkhuat, line 328:
```suggestion
(defn- csv-without-auto-pk-columns
```

This deals with both header and rows, so calling it `rows` is misleading

src/metabase/driver/sql_jdbc.clj
```diff
@@ -158,6 +158,17 @@
     (doseq [sql sqls]
       (qp.writeback/execute-write-sql! db-id sql))))
 
+(defmethod driver/add-columns! :sql-jdbc
+  [driver db-id table-name col->type]
+  (let [table-name (keyword table-name)
```
qnkhuat, line 163:
```suggestion
  (let [
```

you're `keyword`-ing it twice

src/metabase/upload.clj
```diff
@@ -216,49 +216,25 @@
...
   "_mb_row_id")
 
-(defn- remove-indices
-  "Removes the elements at the given indices from the collection. Indices is a set."
-  [indices coll]
-  (keep-indexed (fn [idx item]
-                  (when-not (contains? indices idx)
-                    item))
-                coll))
+(defn- get-auto-pk-column [table-id]
```
qnkhuat, line 219:
```suggestion
(defn- table-id->auto-pk-field-id [table-id]
```
I think this is more descriptive

src/metabase/upload.clj
```diff
@@ -340,29 +293,69 @@
...
 
+(defn- parse-rows
+  "Returns a lazy seq of parsed rows, given a sequence of upload types for each column.
+  Replaces empty strings with nil."
+  [col-upload-types rows]
+  (let [settings (upload-parsing/get-settings)
+        parsers  (map #(upload-parsing/upload-type->parser % settings) col-upload-types)]
+    (for [row rows]
+      (for [[value parser] (map-with-nils vector row parsers)]
+        (when (not (str/blank? value))
```
qnkhuat, line 304:
```suggestion
        (when-not (str/blank? value)
```

test/metabase/api/table_test.clj
```diff
@@ -902,13 +901,14 @@
...
-      (mt/dataset (upload-test/basic-db-definition)
-        (let [filename (mt/random-name)
-              file (upload-test/csv-file-with ["id,name" "2,Luke Skywalker" "3,Darth Vader"] filename)
-              table (t2/select-one :model/Table :db_id (mt/id))]
-          (is (.exists file) "File should exist before append-csv!")
-          (mt/with-current-user (mt/user->id :crowberto)
-            (@#'api.table/append-csv! {:id   (:id table)
-                                       :file file}))
-          (is (not (.exists file)) "File should be deleted after append-csv!"))))))
+    (mt/test-driver :h2
```
qnkhuat, line 904:
```suggestion
    (mt/test-driver (mt/normal-drivers-with-feature :uploads)
```

test/metabase/upload_test.clj
```diff
@@ -1160,111 +1188,178 @@
...
-                     (upload/append-csv! {:file     file
-                                          :table-id (:id table)})))
+                     (append-csv! {:file     file
+                                   :table-id (:id table)})))
               (testing "Check the data was not uploaded into the table"
                 (is (= [[1 "Obi-Wan Kenobi"]]
                        (rows-for-table table))))
               (io/delete-file file))))))))
 
+(deftest append-mb-row-id-1-test
```
qnkhuat, line 1277:
```suggestion
(deftest append-mb-row-id-1-test
```

I'm not a fan of naming tests with numbers, I think we can come up with a short and meaningful name here. it helps reader scan quickly what has been covered.

qnkhuat (changes requested):
The new structure is awesome!

I have some suggestions but those are not blocking.

qnkhuat (approved):
Wrong button ✅

src/metabase/upload.clj
calherries: `csv` is misleading too, because it's kinda independent of the format because the CSV has already been parsed into rows and columns. How about `without-auto-pk-columns`? Then anything that has columns (like the header and rows) are acceptable and there's no confusion

calherries: I renamed it to `without-auto-pk-columns` and used `header-and-rows` in the arglist

```
(defn- without-auto-pk-columns
  [header-and-rows]
  (let [header (first header-and-rows)
        auto-pk-indices (auto-pk-column-indices header)]
    (cond->> header-and-rows
      auto-pk-indices
      (map (partial remove-indices auto-pk-indices)))))
```

calherries:
Suggestions implemented here dc9ddb8e969fb8ad478306e3ac17cff3ed426f17

Thanks for the great review, as always

PR #36715, opened 2023-12-12
🚧 Fix `createMockCard` returns a query without a source table

kulyk:
🚧 Work in progress — awaiting CI feedback

Apparently, `createMockCard` returns technically invalid card objects by default. It builds a `dataset_query` with `createMockStructuredQuery`, which returns an empty object by default. As a result, if we're testing some code that calls query methods, almost everything will be broken because the query doesn't have a table

Fixed by setting a default `source-table` to 1, corresponding to the sample Products table in our preset.

### DIFF
--- a/frontend/src/metabase-types/api/mocks/query.ts
+++ b/frontend/src/metabase-types/api/mocks/query.ts
@@ -5,9 +5,15 @@ import type {
   StructuredQuery,
 } from "metabase-types/api";
 
+// Corresponds to the Sample Database and sample Products table from the built-in preset:
+// https://github.com/metabase/metabase/blob/master/frontend/src/metabase-types/api/mocks/presets/sample_database.ts
+const DEFAULT_DB_ID = 1;
+const DEFAULT_TABLE_ID = 1;
+
 export const createMockStructuredQuery = (
   opts?: Partial<StructuredQuery>,
 ): StructuredQuery => ({
+  "source-table": DEFAULT_TABLE_ID,
   ...opts,
 });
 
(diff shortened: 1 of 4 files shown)


frontend/src/metabase-types/api/mocks/query.ts
```diff
@@ -5,9 +5,15 @@ import type {
   StructuredQuery,
 } from "metabase-types/api";
 
+// Corresponds to the Sample Database and sample Products table from the built-in preset:
+// https://github.com/metabase/metabase/blob/master/frontend/src/metabase-types/api/mocks/presets/sample_database.ts
+const DEFAULT_DB_ID = 1;
```
ranquild, line 10: We use `1` for all databases and tables created by these functions. I wouldn't imply here that it has something to do with out sample db

ranquild (approved):

frontend/src/metabase-types/api/mocks/query.ts
kulyk: Yeah, I think you're right. Updated 👍

frontend/src/metabase-types/api/mocks/query.ts
```diff
@@ -8,6 +8,7 @@ import type {
 export const createMockStructuredQuery = (
   opts?: Partial<StructuredQuery>,
 ): StructuredQuery => ({
+  "source-table": 1,
```
kamilmielnik, line 11:
```suggestion
  "source-table": PRODUCTS_ID,
```

plus

```tsx
import { PRODUCTS_ID } from "metabase-types/api/mocks/presets";
```

kulyk: I think we want presets to depend on mock object helpers, but not vice-versa

PR #36718, opened 2023-12-12
Support segment and metric metadata -> legacy reference conversion

### LINKED ISSUE #36699: BE: MLv2 Add support for segment and metric in legacyFieldRef
`legacyFieldRef` should accept segment and metric together with prev behaviour (column).

before MLv2 segments and metrics didn't support options, so the expected result should be a structure like this:

`['segment', id]`

metamben:
Fixes #36699.


uladzimirdev (approved):
LGTM

bshepherdson (approved):

PR #36724, opened 2023-12-12
add `skip-graph` query param to PUT api/permissions/graph

### LINKED ISSUE #36651: add skip-graph query param to `PUT api/permissions/graph/`

escherize:
Resolves #36651 

The `skip-graph` query parameter has no effect if it is left out of the query params, or included in the query params and set to false.

If PUT `api/permissions/graph/&skip-graph=true`, then we do not return the entire permissions graph (which is a potentially expensive operation).

### DIFF
--- a/src/metabase/api/permissions.clj
+++ b/src/metabase/api/permissions.clj
@@ -5,7 +5,6 @@
    [compojure.core :refer [DELETE GET POST PUT]]
    [honey.sql.helpers :as sql.helpers]
    [malli.core :as mc]
-   [malli.error :as me]
    [malli.transform :as mtx]
    [metabase.api.common :as api]
    [metabase.api.common.validation :as validation]
@@ -87,11 +86,10 @@
   `:sandboxes` feature flag is not present.
 
   If the skip-graph query param is truthy, then the graph will not be returned."
-  [:as {body :body params :params}]
+  [:as {body :body
+        {skip-graph :skip-graph} :params}]
   {body :map
-   params [:and
-           [:map-of :keyword :any]
-           [:map [:skip-graph [:maybe :boolean]]]]}
+   skip-graph [:maybe :boolean]}
   (api/check-superuser)
   (let [graph (mc/decode api.permission-graph/DataPermissionsGraph
                          body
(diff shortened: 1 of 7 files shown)


src/metabase/api/permissions.clj
```diff
@@ -83,23 +83,23 @@
...
   this permissions graph update. Since data sandboxing is an Enterprise Edition-only feature, a 402 (Payment Required)
   response will be returned if this key is present and the server is not running the Enterprise Edition, and/or the
-  `:sandboxes` feature flag is not present."
-  [:as {body :body}]
-  {body :map}
+  `:sandboxes` feature flag is not present.
+
+  If the skip-graph query param is truthy, then the graph will not be returned."
+  [:as {body :body
+        {skip-graph :skip-graph} :params}]
```
escherize, line 90: It's _possible_ to write a malli schema for the params directly here, but it would have to be optional (since the query param is optional), so this seemed a bit simpler.

test/metabase/api/common/internal_test.clj
```diff
@@ -89,6 +90,19 @@
   (let [{:keys [state po-box archipelago]} body]
     (str/join " | " [state po-box archipelago])))
 
+(api/defendpoint GET "/with-query-params/"
```
escherize, line 93: Here's a test that shows how to use query params (sort of a miss last time around).

src/metabase/api/permissions.clj
```diff
@@ -109,7 +109,7 @@
             impersonations         (when impersonation-updates
                                      (insert-impersonations! impersonation-updates))]
         (merge
-         (perms/data-perms-graph)
+         (if skip-graph {} (perms/data-perms-graph))
```
noahmoss, line 112: nit: could also structure it like
```
(merge {}
  (if-not skip-graph (perms/data-perms-graph))
  ...)
  ```

File: src/metabase/api/permissions.clj [hunk 4]
```diff
@@ -83,23 +83,23 @@
...
-      (let [explained (mc/explain api.permission-graph/DataPermissionsGraph body)]
-        (throw (ex-info (tru "Cannot parse permissions graph because it is invalid: {0}"
-                             (pr-str explained))
-                        {:status-code 400
-                         :error (str (me/humanize explained)
-                                     "\n"
-                                     (pr-str explained))}))))
+      (let [explained (mu/explain api.permission-graph/DataPermissionsGraph graph)]
+        (throw (ex-info (tru "Cannot parse permissions graph because it is invalid: {0}" (pr-str explained))
+                        {:status-code 400 :error (pr-str explained)}))))
```
noahmoss, line 102: maybe we don't need `explained` in both the error string and the info map

test/metabase/api/common/internal_test.clj
```diff
@@ -99,16 +113,30 @@
...
   (fn [req] (try
               (handler req)
               (catch Exception e (mw.exceptions/api-exception-response e)))))
 
+(deftest defendpoint-query-params-test
+  (let [^Server server (jetty/run-jetty (json-mw
+                                         (exception-mw
+                                          (wrap-params #'routes))) {:port 0 :join? false})
+        port   (.. server getURI getPort)
+        get! (fn [route]
```
noahmoss, line 131: nit: alignment

noahmoss (approved):
Do you want it to still return the revision number even if it skips the content of the graphs? I don't think it's doing that now, but the epic is saying that it should. If you do, it would be worth having a test for that too.

Other than that, just nits. I'm less familiar with the `defendpoint` infra stuff but the test looks reasonable to me.

src/metabase/api/permissions.clj
escherize: oh yeah, or `when-not` even.

noahmoss: oh yeah that's what I meant to use

escherize:
I noticed that these responses are kind of inconsistent with the versions they're replacing, so I am realigning them here.

I've made it so they all return `{groups: {...} revision: <number>}`.

PR #36732, opened 2023-12-13
Fix CollectionID type

nemanjaglumac:
We don't have a special collection id type called "personal", AFAIC.

Ran into this issue when trying to obtain a collection id using `Urls.extractCollectionId(props.params.slug)`.
image

I think the `personal` type was just a typo.
One way to confirm this is in the UI. Log in as an admin, and then visit these pages:
|route|result|
|-|-|
|`/collection/1`| works|
|`/collection/root`|works|
|`/collection/personal`|fails|
|`/collection/users`|works|

Visiting `/personal` will give you this error
image

Visiting `/users` will list all users personal collections for an admin.

### DIFF
--- a/frontend/src/metabase-types/api/collection.ts
+++ b/frontend/src/metabase-types/api/collection.ts
@@ -5,7 +5,7 @@ import type { DatabaseId } from "./database";
 
 export type RegularCollectionId = number;
 
-export type CollectionId = RegularCollectionId | "root" | "users";
+export type CollectionId = RegularCollectionId | "root" | "personal" | "users";
 
 export type CollectionContentModel = "card" | "dataset";
 
--- a/frontend/src/metabase-types/api/collection.ts
+++ b/frontend/src/metabase-types/api/collection.ts
@@ -5,7 +5,7 @@ import type { DatabaseId } from "./database";
 
 export type RegularCollectionId = number;
 
-export type CollectionId = RegularCollectionId | "root" | "personal";
+export type CollectionId = RegularCollectionId | "root" | "users";
 
 export type CollectionContentModel = "card" | "dataset";
 


frontend/src/metabase-types/api/collection.ts
```diff
@@ -5,7 +5,7 @@ import type { DatabaseId } from "./database";
 
 export type RegularCollectionId = number;
 
-export type CollectionId = RegularCollectionId | "root" | "personal";
```
WiNloSt, line 8: The reason `personal` is here is because we're using it in FE. This ID won't be from BE.

https://github.com/metabase/metabase/blob/df746e7015b0d41444ff53fd5c6653d0c870497c/frontend/src/metabase/entities/collections/constants.ts#L24

Screenshot 2023-12-13 at 2 40 01 PM

I think we should still leave it here.

nemanjaglumac: The comment above it says:
`fake collection for admins that contains all other user's collections`

We already have a string id for that - it's called `users` and it's working already with a router. Is there a reason why we are not using that string?

At the very least, if we insist on using `personal` instead of `users` for ItemPicker, I should go and expand the type because TS will scream at you in certain cases. `users` is a valid collection id that's missing from this type.

WiNloSt: I meant that we should not remove `personal`, but I'm fine with adding `users`. Hope that's clear.

nemanjaglumac: That is clear.
I was just wondering what's the technical reason for choosing `personal` instead of the existing `users`? Maybe @ranquild remembers?

WiNloSt: Ahh I got it now. I'm sorry that I missed your point earlier. I think we should be able to replace `personal` with `users`. But I don't know the details. I just know we have `personal` because I've worked on an area that uses `personal` as a collection name.

nemanjaglumac: Thank you.

Yeah, I don't mind updating all those references if @ranquild doesn't object to using `users` instead of `personal`. IMO, that approach is more clear and it removes an implicit the implicit abstraction.

OTOH, if we decide to leave `personal` along side `users` we ought to document it using JSDoc directly on the type. But I'm hoping to replace it completely with the existing type to avoid all this unnecessary hacking.

ranquild: I don't think that "users" is a collection ID unless it's supported on the backend. That fact we use this string for routing doesn't mean it's a collection id

nemanjaglumac: @ranquild sorry, but that doesn't answer the question, not does it give any pointers as to how to proceed.

Even if both `personal` and `users` are just FE string IDs that backend doesn't recognize, my question still stands. Why use `users` in router and then invent `personal` for other other UI-related cases?

I'm trying to understand the reasoning behind that decision because it's not documented anywhere. If there were some technical factors that affected that decision, that's helpful to know. I cannot arrive to the good solution without knowing these details.

To re-iterate:
Was there a technical reason we chose `personal` as ID identifier?
    a. If not, do you feel strongly about replacing it with `users` that we already use for routing?
    b. If yes, can you please briefly explain it, and I'll just add another string to the type, without removing anything.
    
Thank you

ranquild: Whatever string we use for routing doesn't mean we should patch our API types. I'd not do this change.

nemanjaglumac:
I can close this for now, but will argue that this is still a bug. In fact, adding collection TYPES to the CollectionID type makes no sense.

So rather than arguing whether we should use "personal" or "users", we should rethink the CollectionID type altogether.

PR #36760, opened 2023-12-13
Fix dashcard-filter connection edge cases

oleggromov:
This is a better than https://github.com/metabase/metabase/pull/36657 way of fixing https://github.com/metabase/metabase/issues/35503

## Description
Instead of changing `Dashboards`'s `componendDidUpdate` where some data fetch actions are issued, a better place is the `UPDATE_DASHBOARD_AND_CARDS` action. At the end of it, we now issue the `FETCH_DASHBOARD_DATA` action, which finds out which dashcards need update and does actual requests.

### Before:
https://github.com/metabase/metabase/assets/2196347/d3c72301-ebe2-4b64-84a0-28e17b01428b

### After:
https://github.com/metabase/metabase/assets/2196347/dfc7400a-fbc9-4ad6-940b-7b1ee620ef89

### DIFF
--- a/e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
@@ -19,6 +19,7 @@ import {
   modal,
   dashboardParametersContainer,
   openQuestionActions,
+  spyRequestFinished,
 } from "e2e/support/helpers";
 import {
   ORDERS_DASHBOARD_ID,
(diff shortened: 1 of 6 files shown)


oleggromov:
As discussed with @NevRA, this won't be backported to 48.

ranquild (approved):

frontend/src/metabase/dashboard/actions/save.js
```diff
@@ -131,13 +131,24 @@ export const updateDashboardAndCards = createThunkAction(
...
+      // There might have been changes to dashboard card-filter wiring,
+      // which require re-fetching card data (issue #35503). We expect
+      // the fetchDashboardCardData to decide which cards to fetch.
+      dispatch(
+        fetchDashboardCardData({
+          reload: false,
+          clearCache: false,
+          // source: "SAVE",
+        }),
+      );
```
kulyk, line 151: Should we remove this `source: "SAVE"` line?

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -151,6 +151,8 @@ export function DashCardCardParameterMapper({
           buttonText: formatSelected(selectedMappingOption),
           buttonIcon: (
             <CloseIconButton
+              role="button"
+              aria-label="Disconnect"
```
kulyk, line 155: Please translate the aria label

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
kulyk, line 1: Many of these tests are pretty compact, is there a chance we can group at least a few of them together?
We have a ton of e2e tests and setup + teardown phase for small specs can add up to minutes on CI

kulyk (approved):
Pardon, I meant to approve this. Looks good! Please translate the aria label before merging. And it'd be cool if we could group some e2e tests

frontend/src/metabase/dashboard/actions/save.js
oleggromov: Yeah, sorry - it's a debug leftover, will do.

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
oleggromov: Good idea - I'll have a look later, in a separate PR.

PR #36765, opened 2023-12-13
Fix embed error message to make sense for cards

qwef:
### Description

Previously was saying dashboard id fails when for cards, now we change that by having differing behavior if the card endpoint is called.

### DIFF
--- a/src/metabase/api/embed.clj
+++ b/src/metabase/api/embed.clj
@@ -332,8 +332,6 @@
 (defn- check-embedding-enabled-for-object
   "Check that embedding is enabled, that `object` exists, and embedding for `object` is enabled."
   ([entity id]
-   (api/check (pos-int? id)
-              [400 (tru "Dashboard id should be a positive integer.")])
    (check-embedding-enabled-for-object (t2/select-one [entity :enable_embedding] :id id)))
 
   ([object]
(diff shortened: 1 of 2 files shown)


src/metabase/api/embed.clj
```diff
@@ -332,8 +332,6 @@
 (defn- check-embedding-enabled-for-object
   "Check that embedding is enabled, that `object` exists, and embedding for `object` is enabled."
   ([entity id]
-   (api/check (pos-int? id)
-              [400 (tru "Dashboard id should be a positive integer.")])
```
dpsutton, line 336: I think you can just do it in this function here.

```clojure
[400 (case entity
       :model/Card (tru "Card id should be should be a positive integer")
       :model/Dashboard (tru "Dashboard id should be a positive integer")
       (tru "{0} id should be a positive integer" (name entity)))]
```

qwef: Thought it made less sense to do a if statement here as we already already know what the entity is in the upstream functions, but don't have strong feelings either way.

escherize (changes requested):
Does this handle Dashboards (and other models too)? It would be nice to have at tests for some of those.

qwef:
Currently only card and dashboard models gets called, the dashboard test was already added in the previous PR: https://github.com/metabase/metabase/pull/36491/files

escherize (approved):

PR #36777, opened 2023-12-13
Remove redundant `is` call from test.check defspec

escherize:
When writing a `defspec` we don't need to use `testing` or `is`, just the return value is good enough.

Also use a few less calls, I saw these as "Long Tests" in some CI logging output.

### DIFF
--- a/test/metabase/models/audit_log_test.clj
+++ b/test/metabase/models/audit_log_test.clj
@@ -15,28 +15,33 @@
 
 (def topic-generator (gen/fmap (fn [k] (keyword "event" (name k))) (mg/generator :keyword)))
 
-(defspec topic-gets-unqualified 5000
+(defspec topic-gets-unqualified 2000
   (prop/for-all [ns-kw topic-generator]
-    (is (= (keyword (name ns-kw)) (:unqualified-topic (audit-log/construct-event ns-kw {} nil))))))
+    (= (keyword (name ns-kw))
+       (:unqualified-topic (audit-log/construct-event ns-kw {} nil)))))
 
-(defspec params-take-precedence 1000
+(defspec user-id-params-take-precedence 1000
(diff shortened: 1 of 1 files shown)


test/metabase/models/audit_log_test.clj
```diff
@@ -15,28 +15,33 @@
...
+(defspec user-id-params-take-precedence 1000
+  (prop/for-all [topic topic-generator
+                 event-params (mg/generator ::audit-log/event-params)
+                 user-id (mg/generator [:maybe pos-int?])]
+    (let [constructed-event (audit-log/construct-event topic event-params user-id)]
+      (= (cond
+           (:user-id event-params) (:user-id event-params)
+           user-id user-id
+           :else nil)
+         (:user-id constructed-event)))))
```
escherize, line 32: I split these out, since IMO it makes more sense to check each one independently.

test/metabase/models/audit_log_test.clj
```diff
@@ -15,28 +15,33 @@
...
+(defspec topic-gets-unqualified 2000
   (prop/for-all [ns-kw topic-generator]
-    (is (= (keyword (name ns-kw)) (:unqualified-topic (audit-log/construct-event ns-kw {} nil))))))
+    (= (keyword (name ns-kw))
+       (:unqualified-topic (audit-log/construct-event ns-kw {} nil)))))
 
-(defspec params-take-precedence 1000
+(defspec user-id-params-take-precedence 1000
+  (prop/for-all [topic topic-generator
+                 event-params (mg/generator ::audit-log/event-params)
```
escherize, line 25: Yeah, we get this (semi-realistic) generator for free, thanks to malli.

markbastian (approved):
Looks good. I like splitting the tests out. Easier to debug.

PR #36778, opened 2023-12-13
Add sourceTableOrCardId

### LINKED ISSUE #36672: [MLv2] [BE] Add `sourceTableOrCardId(query)`
Should return `source-table` table id (number) or card id (`card__:id` string)

metamben:
Fixes #36672.

### DIFF
--- a/frontend/src/metabase-lib/query.ts
+++ b/frontend/src/metabase-lib/query.ts
@@ -73,6 +73,6 @@ export function replaceClause(
   return ML.replace_clause(query, stageIndex, targetClause, newClause);
 }
 
-export function sourceTableOrCardId(query: Query): number | string | null {
+export function sourceTableOrCardId(query: Query): TableId | null {
   return ML.source_table_or_card_id(query);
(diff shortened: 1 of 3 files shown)


ranquild (approved):

frontend/src/metabase-lib/query.ts
```diff
@@ -72,3 +72,7 @@ export function replaceClause(
 ): Query {
   return ML.replace_clause(query, stageIndex, targetClause, newClause);
 }
+
+export function sourceTableOrCardId(query: Query): number | string | null {
```
camsaul, line 76: There's actually a `TableId` typescript type defined, you can get it by importing `metabase-types/api`. It's effectively `number | string` so you could do `TableId | null` here maybe. Not sure if that's worth it or not but it does maybe make it easier to track down the places we're using `card__<id>` "table" IDs, hopefully one day we can get rid of that icky hack entirely.

camsaul (approved):

PR #18821, opened 2021-11-03
Adopt new migration numbering scheme

camsaul:
New migration numbering scheme. Context:

We want to backport the migrations to add indexes I did in https://github.com/metabase/metabase/pull/18798 and https://github.com/metabase/metabase/pull/18799 to 0.41.2 since they solve performance issues. The slightly weird thing is their numbers are going to be off: `316` was the last 0.41.0 migration and `317` was the first 42.0 migration, but these new ones will end up being `381` and `382` respectively.

The migration ID is mostly used by Liquibase to determine what order to run unrun migrations in. So it actually won't matter if we run `316`, then `381`, then `382`, then hang out for a bit and then run `317`, `318`, and so forth.

In this particular case `381` and `382` _would_ work fine regardless of whether they run after `316` or after `380`.
I'm not sure we'll be so lucky going forward. Hence the immediate adoption of the new migration ID format for all migrations in x.42.0+. The new format is:

```
vMM.mm-NNN
```
where

```
M = major version
m = minor version
N = migration number relative to that major+minor version
```
e.g. the first migration added to 0.42.0 should be numbered `v42.00-000` and the second migration should be numbered
`v42.00-001`. The first migration for 0.42.1 should be numbered `v42.01-000`, and so forth.

At the time of this decision there were only two 0.42.0 migrations already merged in; there are however a lot of outstanding ones. This PR changes the ones that are already merged in to use the new format.

PR #36787, opened 2023-12-14
RFC: new migration id format vMM.TIMESTAMP

qnkhuat:
This proposes a new format for migration id: `vMM.TIMESTAMP`

where 
- MM: is the major version
- TIMESTAMP: is the current UTC timestamp of format "yyyy-MM-dd'T'HH:mm:ss"

E.g: `v49.2023-04-01T10:30:00`

I've also created a helper function in the dev namespace `(dev/migration-timestamp)` that'll return the current timestamp with the right format for you, so next time you need a new migration all you need to do is call this helper function to get the timestamp and prefix it with a major version.

### Why?

For a long time, we used the format `vMM.mm-NNN` and each time we needed a migration ID, we had to go to #migrations to claim one, as our team is growing fast, this way of collaboration can lead to lots of issues ( as we currently had one: [slack](https://metaboat.slack.com/archives/C010L1Z4F9S/p1702510946081809) ) and also slow us down :) .

With this change, getting a migration ID is effortless, and we can archive the #migrations channel once and for all!

This RFC was originally proposed by Tim ([thread](https://metaboat.slack.com/archives/CKZEMT1MJ/p1670512862341689))

### Format discussion
Feel free to provide feedback if these points do not resonate with you

1. The original RFC proposed the `yyyyMMddHHmmss` for the timestamp, but I found `20230410103000` is much harder to parse compared to `v49.2023-04-01T10:30:00`, so I went with the ISO8601-like format.
2. I've also removed the "patch version" from the format as I don't see a good reason to have it. In fact, we have never used it before. Should we tho? (context: It was proposed in https://github.com/metabase/metabase/pull/18821)


resources/migrations/001_update_migrations.yaml
```diff
@@ -4325,6 +4325,15 @@ databaseChangeLog:
                     nullable: true
                   remarks: 'If the database supports indexing, this column indicate whether or not a field is indexed, or is the 1st column in a composite index'
 
+  - changeSet:
```
qnkhuat, line 4328: will remove this before merging

qnkhuat:
Ignore the failing CIs for now, it's because we changed the format and some of the tests rely on the old format.

dev/src/dev.clj
```diff
@@ -97,6 +98,11 @@
   (mbc/init!)
   (reset! initialized? true))
 
+(defn migration-timetsamp
```
calherries, line 101:
```suggestion
(defn migration-timestamp
```

dev/src/dev.clj
```diff
@@ -97,6 +98,11 @@
...
+  "Returns a UTC timestamp in format `YYYYMMDDHHMMSS` that you can used to postfix for migration ID."
```
calherries, line 102:
```suggestion
  "Returns a UTC timestamp in format `yyyy-MM-dd'T'HH:mm:ss` that you can used to postfix for migration ID."
```

resources/migrations/001_update_migrations.yaml
```diff
@@ -4341,24 +4350,18 @@ databaseChangeLog:
...
 #
-#   e.g. the first migration added to 0.42.0 should be numbered v42.00-000 and the second migration should be numbered
-#   v42.00-001. The first migration for 0.42.1 should be numbered v42.01-000, and so forth.
+#    M         = major version
+#    TIMESTAMP = the current timestamp with format `yyyy-MM-dd'T'HH:mm:ss`
+#                To get this timestamp, evalute this in your REPL: `(dev/migration-timestamp)`
 #
-#   This numbering scheme was adopted beginning with version 0.42.0 so that we could go back and add migrations to patch
-#   releases without the ID sequence getting wildly out of order. See PR #18821 for more information.
+#    E.g: You're adding a new migration for version 49, And it's 10:30:00AM on April 1, 2023,
```
calherries, line 4363:
```suggestion
#    E.g: You're adding a new migration for version 49, And it's 10:30:00AM on April 1, 2023 (UTC),
```

resources/migrations/001_update_migrations.yaml
```diff
@@ -4341,24 +4350,18 @@ databaseChangeLog:
...
-#    M = major version
-#    m = minor version
-#    N = migration number relative to that major+minor version
+#    Where
 #
-#   e.g. the first migration added to 0.42.0 should be numbered v42.00-000 and the second migration should be numbered
-#   v42.00-001. The first migration for 0.42.1 should be numbered v42.01-000, and so forth.
+#    M         = major version
+#    TIMESTAMP = the current timestamp with format `yyyy-MM-dd'T'HH:mm:ss`
+#                To get this timestamp, evalute this in your REPL: `(dev/migration-timestamp)`
```
calherries, line 4361:
```suggestion
#                To get this timestamp, evaluate this in your REPL: `(dev/migration-timestamp)`
```

calherries (approved):
Outside of these typos this approach LGTM. We discussed it during the backend guild sync and no issues with this implementation were introduced.

qnkhuat:
@camsaul the lint already enforces the timestamp in ID to be in order, I added some tests for that.

test/metabase/db/schema_migrations_test/impl.clj
```diff
@@ -95,6 +99,10 @@
                 1000.0)))
         (when (re-matches #"^\d+$" id)
           (Integer/parseUnsignedInt id))
+        (when-let [[_ & components ] (re-matches timestamp-migration-id-re id)]
+          (let [[major-version year month date hour minute second] (map #(Integer/parseUnsignedInt %) components)
+                unix-timestamp (-> (t/zoned-date-time year month date hour minute second 0 (t/zone-id "UTC")) .toInstant .getEpochSecond)]
```
camsaul, line 104: I think an easier way to parse this would be to just put everything after `v49.` in one capture group and then use `u.date/parse` to parse it to a `LocalDateTime`, then you can use `with-time-zone-same-instant` to put it in UTC

camsaul (approved):
Cool, we've been talking about this forever, let's get it in and finally be done with the `#migrations` channel once and for all

test/metabase/db/schema_migrations_test/impl.clj
qnkhuat: awesome, did that in https://github.com/metabase/metabase/pull/36787/commits/11738c0b8f85428a1803681df9ea0876da8c1715

PR #36790, opened 2023-12-14
CSV uploads: Use DATETIME instead of TIMESTAMP for datetime columns with MySQL

### LINKED ISSUE #36761: Uploaded CSVs with datetime fields create TIMESTAMP columns instead of DATETIME in MySQL
### Describe the bug

If you upload a CSV file with a column of datetime strings like `2020-01-01T01:01:01` that don't contain any time zone information, the resulting column that Metabase creates is a `TIMESTAMP` column, when it should be `DATETIME`.

In MySQL, `DATETIME` types have a time and date part, but isn't time zone aware like `TIMESTAMP` columns.

An uploaded CSV file column that contains a string without any time zone information, like `2023-01-01T00:00:00` should become a column in the database that has no time zone information, i.e. `DATETIME` and not `TIMESTAMP`.

### To Reproduce

1. Create a CSV file with a datetime column, e.g.
```
datetime_column
2020-01-01T01:01:01
```
2. Set the uploads database to a MySQL database
3. Upload the CSV to a collection using Metabase's uploads feature
4. Click on the model, and view the SQL to get the table name. 
5. Create a native query with the text `SHOW FIELDS FROM <tablename>` and observe the `datetime_column` has a `timestamp` type, but it should be `datetime`

### Expected behavior

The column should be `DATETIME`

### Logs

_No response_

```JSON
master, but this has been happening since uploads were introduced (47)
```

### Severity

medium

### Additional context

_No response_

calherries:
Fixes #36761

The fix was really simple, just replace `timestamp` with `datetime` and remove the special cases for MySQL in tests.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -282,7 +282,7 @@
         (when (not (str/blank? value))
           (parser value))))))
 
-(mu/defn ^:private column-types-from-rows :- [:sequential (into [:enum] column-types)]
+(mu/defn column-types-from-rows :- [:sequential (into [:enum] column-types)]
   "Returns a sequence of types, given the unparsed rows in the CSV file"
   [settings column-count rows]
   (->> rows
--- a/src/metabase/driver/mysql.clj
+++ b/src/metabase/driver/mysql.clj
@@ -649,7 +649,7 @@
     ::upload/float                    [:double]
     ::upload/boolean                  [:boolean]
     ::upload/date                     [:date]
-    ::upload/datetime                 [:timestamp]
+    ::upload/datetime                 [:datetime]
     ::upload/offset-datetime          [:timestamp]))
 
 (defmethod driver/table-name-length-limit :mysql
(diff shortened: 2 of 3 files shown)


sansolovyov (approved):

src/metabase/upload.clj
```diff
@@ -282,7 +282,7 @@
         (when (not (str/blank? value))
           (parser value))))))
 
-(mu/defn ^:private column-types-from-rows :- [:sequential (into [:enum] column-types)]
```
qnkhuat, line 285: why do you make it public? if it's for test purposes I suggest we should keep it private and use `#'` to access it.

qnkhuat (approved):
LGTM, has a suggestion to keep the column private.

src/metabase/upload.clj
calherries: Yeah, I would like to keep it private but I couldn't commit unless I made this public because I got an error from the linter saying was it was undeclared in the `upload_test` namespace. No idea why it's coming up now

PR #36796, opened 2023-12-14
Filter modal sorting

ranquild:
Epic https://github.com/metabase/metabase/issues/34112
Existing product doc https://www.notion.so/metabase/Bulk-filtering-modal-d8dca207ddcf43428f3814da2839d914?pvs=4#6fbb2c13629a4a498168b628b1401cd0

## Existing implementation

https://github.com/metabase/metabase/blob/1efa6c6f000cb4bd38065563176c35ea0027c865/frontend/src/metabase/query_builder/components/filters/modals/BulkFilterList/utils.ts#L6

Let's figure out how it worked before. Previously `getSortValue` used this map for `semantic_type` checks, then `base_type`. It compared types directly without taking the type hierarchy into account. 

In MLv2 base and semantic types cannot be accessed directly and there are helper functions like `isDate(column)` which check for the type including all subtypes. Let's go through the list and see how it was re-implemented. 

Up-to-date type hierarchy https://github.com/metabase/metabase/blob/c35a6fe966f56d3e0bb430b0d4d240d1e9107693/src/metabase/types.cljc#L71

### CreationTemporal

```
"type/CreationTemporal": 10,
"type/CreationTimestamp": 10,
"type/CreationDate": 10,
"type/CreationTime": 11,
```

These are subtypes of `CreationTemporal`. Only `CreationTimestamp`, `CreationDate`, `CreationTime` can be selected in UI. To achieve the same effect, we need to check for `CreationTimestamp` or `CreationDate` first, and `CreationTime` afterwards. They do not derive from each other.

### Date

````
"type/Date": 12,
````

`Lib.isDate` equivalent 

### Boolean

```
"type/Boolean": 20,
```

`Lib.isBoolean`

### Category

```
"type/Category": 30,
```

There is a caveat here is that some semantic subtypes of this type had lower priority:
```
"type/City": 51,
"type/State": 52,
"type/Country": 54,
```

`Lib.isCategory(column) && !Lib.isAddress(column)` would exclude `City`, `State`, `Country` because these are the only category types which are also addresses.

Another caveat here is that there are also `type/Category` subtypes which are not listed here:
```
type/Name
type/Title
type/Product
type/Company
type/Subscription
type/Source
```

Previously entires for their semantic types were not found and it there was a fallback to
```
"type/Text": 70,
```

In the current implementation I check only for `type/Name` and `type/Title`, assuming that these 2 types could be seen as "short text" types according to the product doc, and the fact that other category types were not included is probably an oversight. 

### Currency

```
"type/Currency": 40,
"type/Price": 40,
"type/Discount": 40,
"type/GrossMargin": 40,
"type/Cost": 40,
```

`type/Currency` and its subtypes. `Lib.isCurrency` should work correctly.

### Location

```
"type/Location": 50,
"type/Address": 50,
"type/City": 51,
"type/State": 52,
"type/ZipCode": 53,
"type/Country": 54,
```

Caveat is that latitude and longitude columns had lower priority:
```
"type/Latitude": 210,
"type/Longitude": 211,
```

`type/Location` cannot be selected directly. It has 2 subtypes - `type/Address` and `type/Coordinate`, which are also not selectable. `type/City`, `type/State`, `type/ZipCode`, `type/Country` are the only descendant types of `type/Address`. `type/Latitude` and `type/Longitude` are the only subtypes of `type/Coordinate`. To sum it up,  we can use:
```
Lib.isCity,
Lib.isState,
Lib.isZipCode,
Lib.isCountry,
```
instead of the list above and it would result in the exactly the same priorities.

### Number

```
"type/Number": 60,
"type/Float": 60,
"type/BigInteger": 60,
"type/Integer": 60,
```

Caveat:
```
"type/Latitude": 210,
"type/Longitude": 211,
```

We can use `Lib.isNumber(column) && !Lib.isCoordinate(column)` to exclude latitude and longitude.

### Text

```
"type/Text": 70,
```

`Lib.isString`

### PK

```
"type/PK": 90,
```

`Lib.isPrimaryKey`

### Latitude & longitude

```
"type/Latitude": 210,
"type/Longitude": 211,
```

`Lib.isLatitude` and `Lib.isLongitude`. Cannot be collapsed to `Lib.isCoordinate` because it would not preserve the order.

### LongText

```
"type/LongText": 220, // not a "real" metabase type, but having it as part of this list makes it easier to sort
```

`Lib.isComment` or `Lib.isDescription`

### FK

```
"type/FK": 230,
```

`Lib.isForeignKey`

### JSON

```
"type/JSON": 240,
```

I don't check for it specifically because it's the last type, so everything else would be at the bottom.

### Orders

No difference.

branch:

Screenshot 2023-12-14 at 16 15 54

master:

Screenshot 2023-12-14 at 16 16 08

### People

Apart from some columns being hidden in metadata, the only difference is that `type/Source` is treated as a "categorical" type in the branch and as a plain text column in `master`. That's expected and mentioned in the "Category" section. To re-iterate, previously some `type/Category` descendant types were not taken into account.

branch:

Screenshot 2023-12-14 at 16 17 33

master:

Screenshot 2023-12-14 at 16 17 39

### Product

The only difference is that "Vendor" field (which is of type `type/Company`) is treated as a categorical type, which was previously ignored.

branch:

Screenshot 2023-12-14 at 16 21 56

master:

Screenshot 2023-12-14 at 16 22 01

### Accounts

2 differences:
- `type/Category` subtypes as above
- boolean fields in `master` were not recognized correctly (because they have `type/Category` `semantic_type`); but the product doc states that boolean fields should go before category fields, so I assume it was a bug.

branch:

Screenshot 2023-12-14 at 16 24 04

master:

Screenshot 2023-12-14 at 16 24 22

### CODE REVIEW

File: frontend/src/metabase/querying/components/FilterModal/BooleanFilterEditor/BooleanFilterEditor.tsx [hunk 1]
```diff
@@ -72,16 +73,16 @@ export function BooleanFilterEditor({
         <Group spacing="md">
           <Checkbox
             label={t`True`}
-            checked={values[0] ?? false}
-            indeterminate={values[0] == null}
+            checked={values.length > 0 ? values[0] : false}
```
ranquild, line 76: Irrelevant bugfix

test/metabase/lib/types/isa_test.cljc
```diff
@@ -148,40 +148,44 @@
               (isa? x :Relation/*) {:semantic-type x}
               :else                {:effective-type x}))]
     (doseq [{:keys [pred positive negative]}
-            [{:pred #'lib.types.isa/temporal?,          :positive :type/DateTime,    :negative :type/City}
```
ranquild, line 151: Mostly whitespace changes with new test cases

bshepherdson (approved):

Issue #36481, opened 2023-12-06
[Epic]Embedding settings cleanup and surface interactive embedding Quickstart 


npretto:
**Links**
- [product doc](https://www.notion.so/metabase/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart-7e4ed0b3a47947ac9235e53998db1280)
- [Technical doc](https://www.notion.so/metabase/Technical-plan-48425a0168cf48728bb97d08e3ee207d)
- [Testing plan](https://www.notion.so/metabase/Test-plan-062f54a005cf4f0fa949a8850d828396)

## Tasks

```[tasklist]
### Milestone "analytics" `embed-settings-cleanup-analytics`
- [ ] https://github.com/metabase/metabase/issues/36754
- [ ] https://github.com/metabase/metabase/pull/36962
```

[Embedding settings cleanup - Milestone 1](https://github.com/metabase/metabase/pull/37043)
```[tasklist]
### Milestone 1 - General embedding settings page `embed-settings-cleanup-milestone1`
- [ ] https://github.com/metabase/metabase/pull/36802
- [ ] https://github.com/metabase/metabase/pull/36857
- [ ] https://github.com/metabase/metabase/pull/36819
- [ ] https://github.com/metabase/metabase/pull/36938
- [ ] https://github.com/metabase/metabase/pull/36944
- [ ] https://github.com/metabase/metabase/pull/36949
- [ ] https://github.com/metabase/metabase/pull/37044
- [x] Review by product & design
- [x] Verify if everything in the testing plan has been tested
```

[Embedding settings cleanup - Milestone 2](https://github.com/metabase/metabase/pull/37430)
```[tasklist]
### Milestone 2 - Polishment embed-settings-cleanup-milestone2
- [ ] https://github.com/metabase/metabase/pull/37053
- [ ] https://github.com/metabase/metabase/pull/37403
- [ ] https://github.com/metabase/metabase/pull/37523
```

PR #36802, opened 2023-12-14
feat: single view for embedding settings regardless of setting status

npretto:
### Description

Design: https://www.figma.com/file/lwyEC7Mx2jfh2jmUDzv3ZS/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart?type=design&node-id=1030-4141&mode=design&t=Mxhy74O4PYk37k7O-4

First task of ms1 of [[Epic]Embedding settings cleanup and surface interactive embedding Quickstart ](https://github.com/metabase/metabase/issues/36481)

This only addresses the top setting and makes the page render the same regardless of the setting being on or off.
Following the links to the other pages when the setting is off is still broken, it will be addressed in a later PR


frontend/src/metabase/admin/settings/components/widgets/EmbeddingSwitchWidget/EmbeddingSwitchWidget.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <Text lh={1.5}>
+      {t`Embed dashboards, questions, or the entire Metabase app into your application. Integrate with your server code to create a secure environment, limited to specific users or organizations.`}
+    </Text>
+    <Switch
+      labelPosition="left"
+      checked={setting.value ?? false}
+      label={<strong>{t`Embedding Enabled`}</strong>}
+      onChange={e => {
+        onChange(e.currentTarget.checked);
+        MetabaseAnalytics.trackStructEvent(
```
npretto, line 24: Previously this was only called when going from disabled -> enabled.
I'll ask product and change accordingly

frontend/src/metabase/admin/settings/components/widgets/EmbeddingSwitchWidget/EmbeddingSwitchWidget.tsx
```diff
@@ -0,0 +1,34 @@
...
+  setting,
+  onChange,
+}: EmbeddingSwitchWidgetProps) => (
+  <Stack spacing={"md"} className="text-measure">
+    <Text lh={1.5}>
+      {t`Embed dashboards, questions, or the entire Metabase app into your application. Integrate with your server code to create a secure environment, limited to specific users or organizations.`}
+    </Text>
+    <Switch
+      labelPosition="left"
+      checked={setting.value ?? false}
```
WiNloSt, line 20: It's more natural to cast this to boolean rather than specify a fallback value especially when the absence of the value means false.

```suggestion
      checked={Boolean(setting.value)}
```

frontend/src/metabase/admin/settings/components/widgets/EmbeddingSwitchWidget/EmbeddingSwitchWidget.tsx
```diff
@@ -0,0 +1,34 @@
...
+      label={<strong>{t`Embedding Enabled`}</strong>}
+      onChange={e => {
+        const newValue = e.currentTarget.checked;
```
WiNloSt, line 23: TIL `currentTarget`. This is new to me. But is there any reason why you couldn't use `e.target.checked` in this case? `e.target` is used more across our codebase.

WiNloSt (approved):
Looks good with minor comments.

frontend/src/metabase/admin/settings/components/widgets/EmbeddingSwitchWidget/EmbeddingSwitchWidget.tsx
npretto: Mmm I think I tried to switch it from target to currentTarget when I encountered an issue to see if it fixed it and forgot to change it back.
I'll change it back for consistency, thanks for pointing that out

PR #36803, opened 2023-12-14
Fixing `:time-enabled` is `nil` time formatting

### LINKED ISSUE #36559: Consistent Date and Time Formatting for `metabase.formatter.datetime`
The function `format-temporal-str` in `metabase.formatter.datetime` doesn't do a good job of handling the zoo of date and time types we might want to export in static artifacts. This was evident when applying the code to CSV exports in #36320.

The underlying issue in the above case is that our tabular exporters convert Java objects (ZonedDateTime, OffsetDateTime, OffsetTime, LocalTime, LocalDateTime, and LocalDate) to strings and then `format-temporal-str` converts those strings to formatted strings. This function has access to the semantic and effective types of the data as well as the visualization settings for the data but does not currently leverage them to their full potential.

The task is to create consistent, desirable formatting for our various time types. All of these can be found with `(descendants :type/Temporal)`, but the primary categories are:

```clojure
(descendants :type/DateTime)
;=>
#{:type/Instant
  :type/JoinTimestamp
  :type/DateTimeWithZoneID
  :type/UpdatedTimestamp
  :type/DateTimeWithZoneOffset
  :type/CreationTimestamp
  :type/DateTimeWithTZ
  :type/CancelationTimestamp
  :type/DeletionTimestamp
  :type/DateTimeWithLocalTZ}
(descendants :type/Date)
;=>
#{:type/UpdatedDate
  :type/CancelationDate
  :type/CancelationTime
  :type/DeletionDate
  :type/Birthdate
  :type/JoinDate
  :type/CreationDate}
(descendants :type/Time)
;=>
#{:type/TimeWithTZ
  :type/TimeWithLocalTZ
  :type/UpdatedTime
  :type/DeletionTime
  :type/JoinTime
  :type/CreationTime
  :type/TimeWithZoneOffset}
```

Note that the Metabase type hierarchy is hierarchical so not every type needs its own special case statement.

Also consider the design of this code. Do we want to stick with a single function with a case statement or use a multimethod or similar dispatch for future extension?

Related [Slack thread](https://metaboat.slack.com/archives/C064QMXEV9N/p1701961844090409).

AC:
* Create a card with a query that captures all our desired date and time formats as columns.
  * This could be more than one card or query, but the foundational requirement is that all desired formats are captured.
* Write a unit test or tests that captures the output of formatting each of these times and verifies the expected formatting.
  * This could be a data literal compared to a parsed pulse attachment in `metabase.pulse.pulse-integration-test`.
  * Note that the source of truth for this is the UI, so we should seek to format each time type as it is presented in Metabase itself.
* Demonstrate to product live or via Loom.

markbastian:
For date time and time strings, when the viz settings contains the `:time-enabled` key with a `nil` value this means we should not show the time portion of the item.

This means for date times, we just show the date and for times we don't show anything.

Closes #36559

### DIFF
--- a/src/metabase/formatter/datetime.clj
+++ b/src/metabase/formatter/datetime.clj
@@ -65,29 +65,21 @@
         (all-cols-settings {::mb.viz/column-name (or field-id-or-name column-name)}))))
 
 (defn- determine-time-format
-  "Given viz-settings with a time-style and possible time-enabled (precision) entry, create the format string."
-  [{:keys [time-enabled time-style] :or {time-enabled "minutes" time-style "h:mm A"}}]
-  (let [base-time-format (case time-enabled
-                           "minutes" "mm"
-
-                           "seconds" "mm:ss"
-
-                           "milliseconds" "mm:ss.SSS"
-
-                           ;; {::mb.viz/time-enabled nil} indicates that time is explicitly disabled, rather than
-                           ;; defaulting to "minutes"
-                           nil nil)]
-    (if base-time-format
-      (case time-style
-        "HH:mm" (format "HH:%s" base-time-format)
-
-        ;; Deprecated time style which should be already converted to HH:mm when viz settings are
-        ;; normalized, but we'll handle it here too just in case. (#18112)
-        "k:mm" (str "h" base-time-format)
-
-        ("h:mm A" "h:mm a") (format "h:%s a" base-time-format)
-
-        time-style)
+  "Given viz-settings with a time-style and possible time-enabled (precision) entry, create the format string.
+  Note that if the `:time-enabled` key is present but the value is nil, we explicitly do not show the time."
+  [{:keys [time-style] :or {time-style "h:mm A"} :as viz-settings}]
+  ;; NOTE - If :time-enabled is present but nil it will return nil
+  (when-some [base-time-format (case (get viz-settings :time-enabled "minutes")
+                               "minutes" "mm"
+                               "seconds" "mm:ss"
+                               "milliseconds" "mm:ss.SSS"
+                               nil nil)]
+    (case time-style
+      "HH:mm" (format "HH:%s" base-time-format)
+      ;; Deprecated time style which should be already converted to HH:mm when viz settings are
+      ;; normalized, but we'll handle it here too just in case. (#18112)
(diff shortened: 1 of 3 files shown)


src/metabase/formatter/datetime.clj
```diff
@@ -65,29 +65,21 @@
...
+  ;; NOTE - If :time-enabled is present but nil it will return nil
+  (when-some [base-time-format (case (get viz-settings :time-enabled "minutes")
+                               "minutes" "mm"
+                               "seconds" "mm:ss"
+                               "milliseconds" "mm:ss.SSS"
+                               nil nil)]
+    (case time-style
+      "HH:mm" (format "HH:%s" base-time-format)
+      ;; Deprecated time style which should be already converted to HH:mm when viz settings are
+      ;; normalized, but we'll handle it here too just in case. (#18112)
```
adam-james-v, line 80: Is there some way we can ensure this is deprecated appropriately and eliminate this part?

adam-james-v (approved):
I think in general this sort of thing could come up again with viz-settings. Keys that exist but have nil values. I think that a more robust solution is to disallow such cases in general. Like, the frontend must set a key to 'false' if some setting is meant to be disabled.

I suppose over time if we can arrive at a spec'd viz-settings these sorts of things will pop out more.

Regardless of that, I think this is a good thing to merge in now, especially because it does solve the problem.

markbastian:
> I think in general this sort of thing could come up again with viz-settings. Keys that exist but have nil values. I think that a more robust solution is to disallow such cases in general. Like, the frontend must set a key to 'false' if some setting is meant to be disabled.
> 
> I suppose over time if we can arrive at a spec'd viz-settings these sorts of things will pop out more.
> 
> Regardless of that, I think this is a good thing to merge in now, especially because it does solve the problem.

Yeah, this is the _explicit_ design, which I think is not the greatest, but it is the current state. I think we should, at some point, work towards using `false` or some other value and not rely on `nil` values. It's just confusing.

src/metabase/formatter/datetime.clj
markbastian: Good idea. Probably something to follow up on. This was lifted from some code @noahmoss wrote in the xlsx ns. We should all discuss.

PR #36809, opened 2023-12-14
Filter modal search values

ranquild:
This is an attempt to handle `has_more_values: true` in field values. Now even with `list` type we do searching with initial set of options.

### DIFF
--- a/frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
+++ b/frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
@@ -38,14 +38,14 @@ function FilterValuePicker({
   shouldCreate,
   onChange,
 }: FilterValuePickerOwnProps) {
-  const { fieldId, searchFieldId, hasFieldValues } = useMemo(
+  const fieldInfo = useMemo(
     () => Lib.fieldValuesInfo(query, stageIndex, column),
     [query, stageIndex, column],
   );
 
-  const { data: fieldValues = [], isLoading } = useFieldValuesQuery({
-    id: fieldId ?? undefined,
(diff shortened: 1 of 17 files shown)


frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
```diff
@@ -51,25 +56,27 @@ function FilterValuePicker({
...
+  if (canSearchFieldValues(fieldInfo)) {
     const columnInfo = Lib.displayInfo(query, stageIndex, column);
 
     return (
       <SearchValuePicker
-        fieldId={fieldId}
-        value={value}
+        fieldId={checkNotNull(fieldInfo.fieldId)}
+        searchFieldId={checkNotNull(fieldInfo.searchFieldId)}
+        fieldValues={fieldData?.values ?? []}
```
ranquild, line 78: It can be non-empty when `has_field_values` is `list` but `has_more_values` is `true`. Otherwise we don't load the field values and the whole object can be `undefined`.

frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
```diff
@@ -78,9 +85,9 @@ function FilterValuePicker({
   }
 
   return (
-    <SelectValuePicker
-      data={data}
-      value={value}
+    <StaticValuePicker
+      fieldValues={fieldData?.values ?? []}
```
ranquild, line 89: We can come here with field values if `has_more_values` is `true` but the field is not searchable, e.g. there is no `searchFieldId`. In this case there is nothing we can do but to display the list of values we have and allow to create new items.

PR #36811, opened 2023-12-14
Clean up the pulse page

kulyk:
A few visual fixes for the pulse page:

* removed broken card visualizations
* removed floating callouts (absolutely positioned yellow blocks text)

### Demo

| Before | After  |
|--------|-------|
|             |   after        |

### DIFF
--- a/e2e/test/scenarios/sharing/pulse.cy.spec.js
+++ b/e2e/test/scenarios/sharing/pulse.cy.spec.js
@@ -35,9 +35,6 @@ describe("scenarios > pulse", { tags: "@external" }, () => {
 
     popover().findByText("Orders, Count").click();
 
-    // pulse card preview
-    cy.get("@pulseData").contains("18,760");
-
     cy.findByRole("heading", { name: "Where should this data go?" })
       .parent()
       .within(() => {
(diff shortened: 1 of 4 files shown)


alxnddr:
The layout looks so much better now!

I noticed that newly added cards have empty titles but it worked previously:
https://github.com/metabase/metabase/assets/14301985/855c5b71-1a7f-4e9a-a606-b30b86112bb5

luizarakaki (approved):

luizarakaki:
> The layout looks so much better now!
> 
> I noticed that newly added cards have empty titles but it worked previously: https://github.com/metabase/metabase/assets/14301985/855c5b71-1a7f-4e9a-a606-b30b86112bb5

If it is simple to fix, it would be good. But we are not expecting people adding new cards to existing pulses, just consulting what is there so they can transition to subscriptions

kulyk:
Nice catch @alxnddr, thanks! Should be fixed now

https://github.com/metabase/metabase/assets/17258145/a1118291-3c21-4adc-bff6-52857888d05e

alxnddr (approved):
🚀

PR #36258, opened 2023-12-01
API Keys: add AuthCard and listing page

shaunlebron:
Closes https://github.com/metabase/metabase/issues/36805

Adding the entry point for managing API Keys at <http://localhost:3000/admin/settings/authentication>

## Summary

1. Added `ApiKey` type to “metabase-types/api/admin”
1. Added `ApiKeysApi` entry to “metabase/services”
1. Added `AuthCard` for API Keys
1. Added `ManageApiKeys` page
    * Added CRUD modals for creating, editing, and deleting keys.
    * Added secondary modals for regenerating a key and copying an unmasked key.
1. Edited `GroupListing` page to show number of API keys belonging to each group
1. Edited `GroupDetails` page to show each API key belonging the group

I also created some reusable components:
1. Added `FormGroupWidget` component for selecting a single Group.

PR #36813, opened 2023-12-14
Change e2e spec pattern to allow TypeScript extension

shaunlebron:
Continuing https://github.com/metabase/metabase/pull/36474

Allow `*.cy.spec.ts` files to be picked up by our Cypress test runner.  I’m adding the first typescript spec here: https://github.com/metabase/metabase/pull/36258

### DIFF
--- a/e2e/runner/cypress-runner-run-tests.js
+++ b/e2e/runner/cypress-runner-run-tests.js
@@ -12,7 +12,7 @@ const isFolder = !!folder;
 const isOpenMode = args["--open"];
 
 const getSourceFolder = folder => {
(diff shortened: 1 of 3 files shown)


nemanjaglumac (approved):
Nice, works for me!
I didn't think we'd have actual specs in TS so soon, but I don't see a reason why not to.

Thank you

kulyk (approved):

nemanjaglumac:
@shaunlebron I changed the label to "backport" because we'll need all the TS stuff in the release 48 branch as well.
I've manually backported the original PoC PR here.
https://github.com/metabase/metabase/pull/36849

Once that's merged, I think we can safely backport this as well (although, merging this first would still most likely not result in any conflicts anyway).

PR #36814, opened 2023-12-14
Fix dashboard filter panel spacing and border

kulyk:
Fixes two things about the dashboard filters panel:

* adds a vertical space between the dashboard grid and the dashboard filters bottom border
* makes the dashboard filters bottom border only show up on scroll

I also cleaned up some code in the area, I think reviewing changes by commit would be more convenient:

* [clean up sticky filters container code](https://github.com/metabase/metabase/commit/5e2c1340a0af38750babbdcd32eefe4bfb9903ab) — we used to have way more code to manage filter panel's "stickiness" before it was mostly migrated to CSS. The PR removes most of the code in favor of a [single function](https://github.com/metabase/metabase/blob/5e2c1340a0af38750babbdcd32eefe4bfb9903ab/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx#L398)
* [Hide parameters container border until scrolling](https://github.com/metabase/metabase/commit/d5f0e03c3a4dfe8e227e1961a596d9157a2cb060) — I suppose we lost this bit of logic when migrating it to CSS
* [Add margin between grid and parameter list bottom border](https://github.com/metabase/metabase/commit/9d06c14dbe59e75b303b70ff15c088b12563ddab) — adds the margin ✨ 

### Demo

| Before | After  |
|--------|-------|
| before            | after          |


kdoh (changes requested):
Looking good. Couple things I noticed.

Things are also very tight when there aren't filters. I think if we're fixing the filter panel we also need to fix this same problem here when you don't have filters.
Screenshot 2023-12-14 at 4 54 02 PM

It'd be nice to even out the top and bottom padding here.
Screenshot 2023-12-14 at 4 55 58 PM

kulyk:
Appreciate a thorough review, @kdoh 🙏 Please take another look:

<details>
<summary>With filters</summary>

filters

</details>

<details>
<summary>With tabs</summary>

tabs

</details>

<details>
<summary>No tabs or filters</summary>

none

</details>

kdoh (approved):
LGTM outcome wise. Didn't do a thorough code review though.

ranquild (approved):

PR #36815, opened 2023-12-14
docs - date formats for CSV uploads

jeff-bruemmer:
Adds notes about date formats for CSV uploads.

### DIFF
--- a/docs/databases/uploads.md
+++ b/docs/databases/uploads.md
@@ -69,6 +69,40 @@ If you have a file larger than 200 MB, the workaround here is to:
 2. Upload those files one by one. Metabase will create a new model for each sheet.
 3. Consolidate that data by creating a new question or model that joins the data from those constituent models created by each upload.
 
+## Date formats
+
+For now, Metabase only recognizes dates and datetimes from strings in uploaded CSVs with the following formats:
+
+### Dates
+
+Represents the year, month, and day without time information.
+
+**Format**: `yyyy-MM-dd`
+
+**Example**: `2023-01-01`
+
+### Datetimes
+
+Represents the year, month, day, hour, minute, second, and millisecond with optional fractional seconds.
+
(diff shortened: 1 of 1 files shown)


docs/databases/uploads.md
```diff
@@ -69,6 +69,40 @@ If you have a file larger than 200 MB, the workaround here is to:
...
+
+Represents the year, month, and day without time information.
+
+**Format**: `yyyy-MM-dd`
+
+**Example**: `2023-01-01`
+
+### Datetimes
+
+Represents the year, month, day, hour, minute, second, and millisecond with optional fractional seconds.
```
calherries, line 86: Minute and second are also optional.

e.g. `2023-01-01 00` is valid

Also millisecond is a type of fractional second.

`2023-01-01 00:00:00.0` is valid, as is `2023-01-01 00:00:00.0000000`, although that precision will likely be rounded by the database.

docs/databases/uploads.md
```diff
@@ -69,6 +69,40 @@ If you have a file larger than 200 MB, the workaround here is to:
...
+**Examples**: 
+
+- `2023-01-01 00:00:00.000`
+- `2023-01-01T00:00:00.000`
+
+### Datetimes with offsets
+
+Represents the datetime with an offset from Coordinated Universal Time (UTC).
+
+**Format**: `yyyy-MM-ddTHH:mm:ss.SSS±HH:mm:ss`. The "T" separator could also be a space (" ").
```
calherries, line 99: I think it's worth pointing out that the minutes and seconds in the offset are optional. Most time zones only have hour components, so the most common format will probably be `yyyy-MM-ddTHH:mm:ss±HH`.

Also `Z` is valid. 

The full spec is here
https://github.com/metabase/metabase/blob/dc9ddb8e969fb8ad478306e3ac17cff3ed426f17/src/metabase/upload/parsing.clj#L68-L78

calherries (changes requested):

calherries (approved):
Looks great, thank you

PR #36826, opened 2023-12-14
CSV appends: Handle type mismatch errors

calherries:

This PR implements handling type mismatches between the table and the uploaded CSV with the correct error handling.

It corresponds to the following section of the eng doc ([block link](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df?pvs=4#2e0b218dfb154e17a4cdbff821c4cd43)):

**Insertion failures**

If there are any failures during insertion, we need undo all changes that have been made.

- In some cases undoing changes is harder, because we might insert data in chunks for performance reasons, which means we can’t just rollback a transaction. We have two types of tables to deal with: one with an automatically generated ID column, and one without.

- The chosen solution: Add a column `_mb*_*row_id` if not already present, and use the MAX of this column for rollbacks
available for all DBs.

- [x]  Test: If there is a PK column in the table and the upload fails during insertion, the inserted rows are deleted
- [x]  Test: If there is not a PK column in the table and the upload fails during insertion, the inserted rows are deleted


src/metabase/upload.clj
```diff
@@ -557,25 +561,38 @@
...
-            (t2/update! :model/Field (:id auto-pk-field) {:display_name (:name auto-pk-field)})))
-        {:row-count (count parsed-rows)}
-        (catch Throwable e
-          ;; TODO: insert rows failure handling
-          (throw (ex-info (ex-message e) {:status-code 400})))))))
+                             {(keyword auto-pk-column-name) (driver/upload-type->database-type driver ::auto-incrementing-int-pk)}))
+      ;; If the insert fails, we need to delete the rows we just inserted. So we:
+      ;; 1. get the max PK value before inserting
+      ;; 2. insert the rows
+      ;; 3. if insertion fails, delete the rows with PKs greater than the max PK value
```
calherries, line 572: I'm hoping I can replace all this with database transactions later, but I'm still trying to figure out why transactions aren't rolling back correctly with MySQL and H2, and this works fine for now.

src/metabase/driver.clj
```diff
@@ -876,6 +876,10 @@
 ;;; |                                                    Upload                                                      |
 ;;; +----------------------------------------------------------------------------------------------------------------+
 
+(def ^:dynamic *insert-chunk-rows*
+  "The number of rows to insert at a time when uploading data to a database. This can be bound for testing purposes."
+  nil)
```
qnkhuat, line 881: why don't you set this to 100 by default?

src/metabase/upload.clj
```diff
@@ -557,25 +561,38 @@
...
+          (catch Throwable e
+            ;; 3.
+            (jdbc/execute! (sql-jdbc.conn/db->pooled-connection-spec database)
+                           (sql/format {:delete-from (keyword (table-identifier table))
+                                        :where       [:> (keyword auto-pk-column-name) max-pk]}
+                                       {:quoted  true
+                                        :dialect (sql.qp/quote-style driver)}))
+            (throw (ex-info (ex-message e) {:status-code 422})))
+          (finally
+            (scan-and-sync-table! database table)
```
qnkhuat, line 591: can you add a comment here on why we want these steps to be triggered even when insert fails?

src/metabase/upload/parsing.clj
```diff
@@ -35,7 +35,11 @@
   Supported formats:
     - yyyy-MM-dd"
   [s]
-  (t/local-date s))
```
qnkhuat, line 38: are you aiming for completeness here? if so we need to update `parse-datetime` too.

How about introducing a macro?

qnkhuat: and btw, this function can be made private. I suspect a bunch of others are too.

src/metabase/upload/parsing.clj
```diff
@@ -113,11 +117,14 @@
...
          (parse-plain-number number-separators))
-    (catch Throwable e
-      (throw (ex-info
-              (tru "{0} is not a recognizable number" s)
-              {}
-              e)))))
+    (catch Exception _
+      (throw (IllegalArgumentException. (tru "{0} is not a recognizable number" s))))))
+
+(defn parse-as-int
```
qnkhuat, line 123: it seems weird to me to convert the value to str in this function. 

we either should change the same for it to makes more sense, or we should convert it to str when we use it.

test/metabase/upload_test.clj
```diff
@@ -92,30 +104,30 @@
            ["0,0"        0              float-type ",."]
            ["0,0"        0              float-type ", "]
            ["0.0"        0              float-type ".’"]
-           ["$2"         2              int-type]
```
qnkhuat, line 95: why do we have this change?

test/metabase/upload_test.clj
```diff
@@ -1302,6 +1314,33 @@
...
+(deftest append-no-mb-row-id-failure-test
+  (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
+    (with-uploads-allowed
+      (testing "If the table doesn't have _mb_row_id and a failure occurs, we should create a _mb_row_id column anyway"
+        (mt/with-empty-db
+          (let [table (create-upload-table! {:col->upload-type (ordered-map/ordered-map
+                                                                :bool_column ::upload/boolean)
+                                             :rows [[true]]})
+                csv-rows ["bool_column" "not a bool"]
+                file  (csv-file-with csv-rows (mt/random-name))]
```
qnkhuat, line 1326: the indents for these bindings are not aligned

test/metabase/upload_test.clj
```diff
@@ -1364,3 +1403,183 @@
...
+            (binding [driver/*insert-chunk-rows* 10]
+              (let [table (create-upload-table! {:col->upload-type (ordered-map/ordered-map
+                                                                    :_mb_row_id ::upload/auto-incrementing-int-pk
+                                                                    :id         ::upload/int
+                                                                    :name       ::upload/varchar-255)
+                                                 :rows             [[10 "Obi-Wan Kenobi"]]})
+                    csv-rows `["id,name" ~@(repeat 50 "20,Darth Vadar") "not an int,Luke Skywalker"]
+                    file  (csv-file-with csv-rows (mt/random-name))]
+                (testing "Check integers"
+                  (is (= {:message "not an int is not a recognizable number"
```
qnkhuat, line 1424: should we quote the value in the error message?

test/metabase/upload_test.clj
```diff
@@ -1302,6 +1314,33 @@
...
+  (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
+    (with-uploads-allowed
+      (testing "If the table doesn't have _mb_row_id and a failure occurs, we should create a _mb_row_id column anyway"
+        (mt/with-empty-db
+          (let [table (create-upload-table! {:col->upload-type (ordered-map/ordered-map
+                                                                :bool_column ::upload/boolean)
+                                             :rows [[true]]})
+                csv-rows ["bool_column" "not a bool"]
+                file  (csv-file-with csv-rows (mt/random-name))]
+            (is (thrown? Exception
```
qnkhuat, line 1327: I'd add a test like this
```clojure
(is (false? (t2/exists? :model/Field :table_id (:id table) :name upload/auto-pk-column-name)))
```
for documentation purposes

src/metabase/driver.clj
calherries: What's the benefit of that? I don't want it to look like there should be a common default value for all drivers. It's just a coincidence that it's 100 for all drivers at the moment.

src/metabase/upload/parsing.clj
calherries: > if so we need to update parse-datetime too.

Unlike `parse-date`, `parse-datetime` is not used in an `upload-type->parser` function, except to help implement `parse-as-datetime`. So it doesn't need the same exception handling.

> How about introducing a macro?

This is subjective but I don't think wrapping a body in a try/catch like this crosses my threshold for "so much boilerplate I want a macro to hide it"

> this function can be made private

Good point, I'll fix this and others

test/metabase/upload_test.clj
calherries: this was because of the bigints problem, but nvm, I've changed it back

src/metabase/upload/parsing.clj
calherries: Fixed with bc5b7d44cb08d8d82aeed81a967d82ecdba6e22c

calherries:
@qnkhuat I've implemented most of your suggestions, can you take another look?

qnkhuat (approved):
Cool, LGTM now

PR #36839, opened 2023-12-15
Add field values search info

### LINKED ISSUE #36698: [MLv2] [BE] Add field values info
We need to handle field values in a special way in the filter modal. The first step is to get relevant field properties. Proposed signature:

```
function fieldValuesInfo(
  query: Query, 
  stageIndex: number, 
  column: ColumnMetadata
): FieldValuesInfo;

type FieldValuesInfo = {
  // null means that the underlying field was not found
  fieldId: number | null; 
  // a note for it below
  searchFieldId: number | null;
  // corresponds to has_field_values property, or "none" if the field is not found
  hasFieldValues: "list" | "search" | "none" 
};
```

The FE will call `/api/field/:fieldId/values` to fetch the list of values if `hasFieldValues` is `list`, and display them appropriately. 

We would also call `/api/field/:fieldId/search/:searchFieldId` to search for field values. `searchFieldId` should handle field remapping like the existing function does with `disablePKRemapping` = `true`: https://github.com/metabase/metabase/blob/0624d8d0933f577cc70c03948f4b57f73fe13ada/frontend/src/metabase-lib/metadata/Field.ts#L480

The FE will not attempt to call the search endpoint if `searchFieldId` is `null`.

camsaul:
Add `fieldValuesSearchInfo()` which is used to power some FE search widget in filters.

As part of this I needed to update the JS metadata provider to parse `:dimensions` and `:values` to `:lib/external-remapping` and :lib/internal-remapping` like the JVM metadata provider already does for corresponding QP stuff. I also moved a lot of the schemas and code that powers some of the hydrated keys that the REST API gives the FE into MLv2 so it can calculate this stuff directly instead of relying on whatever wack version might be in the results metadata. 

This PR also includes a lot of unrelated improvements, I couldn't help myself.

Resolves #36698

### DIFF
--- a/frontend/src/metabase-lib/fields.ts
+++ b/frontend/src/metabase-lib/fields.ts
@@ -77,7 +77,7 @@ export function legacyRef(
 /**
  * Info about FieldValues/remapping for the purposes of powering search widgets in filter modals.
  */
-function fieldValuesSearchInfo(
+export function fieldValuesSearchInfo(
   metadataProviderable: MetadataProvider | Query,
   column: ColumnMetadata,
 ): FieldValuesSearchInfo {
--- a/frontend/src/metabase-lib/fields.ts
+++ b/frontend/src/metabase-lib/fields.ts
@@ -3,6 +3,7 @@ import type { FieldReference } from "metabase-types/api";
 import type {
   Clause,
   ColumnMetadata,
+  MetadataProvider,
   MetricMetadata,
   Query,
   SegmentMetadata,
@@ -72,3 +73,22 @@ export function legacyRef(
 ): FieldReference {
   return ML.legacy_ref(column);
 }
+
+/**
+ * Info about FieldValues/remapping for the purposes of powering search widgets in filter modals.
+ */
+function fieldValuesSearchInfo(
+  metadataProviderable: MetadataProvider | Query,
+  column: ColumnMetadata,
+): FieldValuesSearchInfo {
+  return ML.field_values_search_info(metadataProviderable, column);
+}
+
+type FieldValuesSearchInfo = {
+  // null means that the underlying field was not found
+  fieldId: number | null;
+  // a note for it below
+  searchFieldId: number | null;
+  // corresponds to has_field_values property, or "none" if the field is not found
+  hasFieldValues: "list" | "search" | "none";
+};
--- a/src/metabase/lib/metadata/jvm.clj
+++ b/src/metabase/lib/metadata/jvm.clj
@@ -21,22 +21,24 @@
    [toucan2.pipeline :as t2.pipeline]
    [toucan2.query :as t2.query]))
 
+(set! *warn-on-reflection* true)
+
 (defn- qualified-key? [k]
   (or (qualified-keyword? k)
       (str/includes? k ".")))
 
-(def ^:private memoized-kebab-key
+(def ^:private ^{:arglists '([k])} memoized-kebab-key
   "Calculating the kebab-case version of a key every time is pretty slow (even with the LRU
-  caching [[u/->kebab-case-en]] has), since the keys here are static and finite we can just memoize them forever with
-  regular memoization and get a nice performance boost."
-  ;; actually, just using a simple atom like this is 10x faster than [[memoize]], and thus 100x faster
-  ;; than [[metabase.util.memoize/lru]]... see https://metaboat.slack.com/archives/C04CYTEL9N2/p1702610853943109
-  (let [cache (atom (hash-map))]
+  caching [[u/->kebab-case-en]] has), since the keys here are static and finite we can just memoize them forever and
+  get a nice performance boost."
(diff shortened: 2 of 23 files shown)


frontend/src/metabase-lib/fields.ts
```diff
@@ -72,3 +73,22 @@ export function legacyRef(
 ): FieldReference {
   return ML.legacy_ref(column);
 }
+
+/**
+ * Info about FieldValues/remapping for the purposes of powering search widgets in filter modals.
+ */
+function fieldValuesSearchInfo(
```
camsaul, line 80: I decided to call this `fieldValuesSearchInfo` and not just `fieldValuesInfo`, because these three little bits of info are hardly the only info you might have about field values or remapping.

src/metabase/lib/metadata/jvm.clj
```diff
@@ -25,12 +25,25 @@
...
+  caching [[u/->kebab-case-en]] has), since the keys here are static and finite we can just memoize them forever with
+  regular memoization and get a nice performance boost."
+  ;; actually, just using a simple atom like this is 10x faster than [[memoize]], and thus 100x faster
+  ;; than [[metabase.util.memoize/lru]]... see https://metaboat.slack.com/archives/C04CYTEL9N2/p1702610853943109
+  (let [cache (atom (hash-map))]
+    (fn [k]
+      (or (get @cache k)
+          (let [k' (u/->kebab-case-en k)]
+            (swap! cache assoc k k')
+            k')))))
```
camsaul, line 39: I spent a lot of time wringing my hands trying to decide if `instance->metadata` below was going to be too slow to use in hydration code vs just manually building a map with the three keys I needed, eventually I decided that speeding up `u/->kebab-case-en` in this way gave us a significant performance boost and could let me use `instance->metadata` without worrying about the performance too much. See https://metaboat.slack.com/archives/C04CYTEL9N2/p1702610853943109 for more info.

src/metabase/query_processor/store.clj
```diff
@@ -241,14 +241,14 @@
 (def ^:private ^{:deprecated "0.48.0"} LegacyFieldMetadata
   [:map
    [:name          ms/NonBlankString]
-   [:table_id      ::lib.schema.common/positive-int]
+   [:table_id      ::lib.schema.id/table]
```
camsaul, line 244: preferrable to be more specific, more than once I've found myself unconfused by what an `:id` key was supposed to be an ID for by looking at the schema. One day this stuff will come in handy when we do generative nonsense

test/metabase/lib/metadata/jvm_test.clj
```diff
@@ -157,7 +157,7 @@
                 (map #(lib/display-info agg-query %)
                      (lib.metadata.calculation/returned-columns agg-query))))))))
 
-(deftest ^:synchronized internal-remap-metadata-test
+(deftest ^:synchronized external-remap-metadata-test
```
camsaul, line 160: these test names were backwards

test/metabase/lib/test_util/metadata_providers/remap.cljc
```diff
@@ -3,17 +3,18 @@
...
    [metabase.lib.schema.id :as lib.schema.id]
+   [metabase.lib.schema.metadata :as lib.schema.metadata]
    [metabase.lib.test-util.metadata-providers.mock
     :as lib.tu.metadata-providers.mock]
    [metabase.lib.util :as lib.util]
    [metabase.util :as u]
    [metabase.util.malli :as mu]))
 
-(mu/defn ^:private external-remapped-column :- lib.metadata/ColumnMetadata
+(mu/defn ^:private external-remapped-column :- ::lib.schema.metadata/column
```
camsaul, line 13: preferrable to use the new `lib.schema` keywords, we don't need the old ones like `lib.metadata/ColumnMetadata` so we can hopefully get rid of them soon

src/metabase/util/snake_hating_map.clj
```diff
@@ -71,4 +71,4 @@
        (vary-meta assoc :metabase.driver/metadata-type :metabase.driver/metadata-type.mlv2)
        ->SnakeHatingMap))
   ([k v & more]
-   (snake-hating-map (into {} (cons [k v] (partition-all 2 more))))))
+   (snake-hating-map (into {k v} (partition-all 2) more))))
```
camsaul, line 74: Bug. I added a test for this.

ranquild (approved):

src/metabase/lib/metadata/jvm.clj
metamben: Have you considered using a `ConcurrentHashMap` with `computeIfAbsent` instead of the atom?

src/metabase/util/snake_hating_map.clj
metamben: Wow, this is a tricky fix. (The new line is what I find natural, but it's difficult to see how this fixes what bug.)

metamben (approved):

PR #36851, opened 2023-12-15
Serialization API

### LINKED ISSUE #36843: [Epic] Serialization API
- [product doc](https://www.notion.so/metabase/Add-a-Serialization-API-endpoints-export-and-import-d2e12ec97c3f4fc1ab1f005dea536ac2)
- [tech design doc](https://www.notion.so/metabase/Serialization-API-4d4dd2e3f3e3486d8db4b22247dbf4f7)

sansolovyov:
Adds two new methods `/api/ee/serialization/export` and `/api/ee/serialization/import`. All the files created during serialization are then cleaned up with a help of `StreamableResponseBody` - we have an ability to execute something after response is written.

I had to improve test http client somewhat to handle multipart uploads and that led to a few changes in unrelated tests. Also logging required extracting some functionality from tests to proper codebase.

Resolves #36843


deps.edn
```diff
@@ -234,6 +234,8 @@
     methodical/methodical        {:mvn/version "0.15.1"}                    ; drop-in replacements for Clojure multimethods and adds several advanced features
     org.clj-commons/hickory      {:mvn/version "0.7.3"}                     ; Parse HTML into Clojure data structures
     org.clojure/algo.generic     {:mvn/version "0.1.3"}
+    peridot/peridot              {:git/url "https://github.com/piranha/peridot.git"
+                                  :sha "999d0a02425c906c35bace749654dc095ecf3e6a"} ; mocking Ring requests; update to upstream when/if it merges changes
```
piranha, line 238: I'm thinking if I should just use Java lib directly? That would add about 20 lines of code in total, but one dependency less.

dev/src/dev.clj
```diff
@@ -301,3 +305,32 @@
...
+        form# (pr-str x)
+        pref# (str ";; " path# "\n"
+                   ";; " form# "\n")]
+    #_{:clj-kondo/ignore [:discouraged-var]}
+    `(let [~x# ~x]
+       (println (str ~pref# (puget/cprint-str ~x#) "\n"))
+       ~x#)))
+
+;; make spy available everywhere
+(intern 'clojure.core (with-meta 'spy {:macro true}) @#'spy)
```
piranha, line 336: This made debugging http-client issues much more bearable. :)

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,242 @@
...
+;;; Logging
+
+(defn- find-logger-layout
+  "Find any logger with a specified layout"
+  [^LoggerConfig logger]
+  (when logger
+    (or (first (keep #(.getLayout ^AbstractAppender (val %)) (.getAppenders logger)))
+        (recur (.getParent logger)))))
+
+(defn- make-logger ^AutoCloseable [ns ^File f]
```
piranha, line 121: I've been thinking about moving this to `metabase.logger` (`metabase.util.log` is both clj/cljs, so it's not applicable there), but maybe it's better done when it's needed somewhere else.

OTOH maybe this can be re-used in tests...

deps.edn
piranha: OTOH, it's tested there and that's nice, so no, let it be dependency and I'll try to get my changes upstream.

dev/src/dev.clj
```diff
@@ -89,10 +91,10 @@
   reset-changes!
   changes])
 
-(def initialized?
+(def initialized? "initialized?"
```
qnkhuat, line 94: we normally place the doc in a new line

piranha: Yeah... but it looked so lonely. 😁 It's just clj-kondo triggered on me since I've touched dev.clj.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,242 @@
...
+    (io/copy f baos)
+    (.toByteArray baos)))
+
+(defn- on-response! [req data]
+  (reify
+    ;; Real HTTP requests and mt/user-real-request go here
+    ring.protocols/StreamableResponseBody
+    (write-body-to-stream [_ response out]
+      (ring.protocols/write-body-to-stream data response out)
+      (run-callbacks! req))
```
qnkhuat, line 53: I believe `run-callbacks!` will keep the connection until it's completed right?
so:
- how about sending it to a future? 
- do we really need to use an atom here? how about we make `serialize&pack` and `unpack&import` returns a list of callback functions, then we use that as an input for `on-response!`? I don't love the idea of `register-cb!` because it can be called anywhere in the call stack; it makes debugging hard, and it's hard to know what could happen.

dev/src/dev.clj
qnkhuat: What's cool about spy? Have you tried [hashp](https://github.com/weavejester/hashp)? the best thing about hashp for me is that it's implemented as a data reader, so I can use it anywhere without changing the existing forms.

piranha: Yeah I did use hashp before spy, but it has two deficiencies to me: you have to require it (that is fairly easy to fix of course), and you can't use it in pipelines like `(-> 1 inc inc #p inc)`. There is not even a macro you can use in situation like that.

And interning this guy does two things: brings it everywhere, and makes it available to everyone. What it lacks is a line number, I should copy that from hashp for sure.

What do you think? :)

piranha: Orrr maybe I should try to improve hashp so that it works in this conditions and we just bring it in dev env for everyone.

qnkhuat: ah right, pipeline would be one of the reasons. I'm open for new tools tho. I was just curious about the use case of spy over hashp.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: They will, yeah. I like the idea of passing callbacks directly instead of having an atom, this makes it simpler. I'm not sure about the future though, since removal shouldn't take too much time, and waiting in tests... `with-redefs` won't work since it's macro, so how do I not introduce a flake then?

qnkhuat: You can use a dynamic var to make it run synchronously in test but async in prod

piranha: Okay, let's do it like that. Removal takes about 80ms per 1k files at my laptop, so I guess for a slower disk it's going to be even more noticeable.

enterprise/backend/test/metabase_enterprise/serialization/api_test.clj
```diff
@@ -0,0 +1,146 @@
...
+        (is (= known-files
+               (set (.list (io/file api.serialization/parent-dir)))))))))
+
+(deftest export-import-test
+  (testing "Serialization API e2e"
+    (let [known-files (set (.list (io/file api.serialization/parent-dir)))]
+      (premium-features-test/with-premium-features #{:serialization}
+        (binding [api.serialization/*in-tests* true]
+          (testing "POST /api/ee/serialization/export"
+            (mt/with-temp! [Collection coll  {}
```
qnkhuat, line 108: could you explain why we need to use `with-temp!` instead of `with-temp` here?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+     :log-file (when (.exists log-file)
+                 log-file)
+     :callback (fn []
+                 (when (.exists path)
+                   (run! io/delete-file (reverse (file-seq path))))
+                 (when (.exists dst)
+                   (io/delete-file dst)))}))
+
+(defn- unpack&import [^File file & [size]]
+  (let [dst      (io/file parent-dir (apply str (repeatedly 20 #(char (+ 65 (rand-int 26))))))
```
qnkhuat, line 169: > (apply str (repeatedly 20 #(char (+ 65 (rand-int 26))

we have `(mt/random-name)` that does exactly this, maybe we should extract it to `metabase.util` so we can reuse it?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+;;; HTTP API
+
+(api/defendpoint POST "/export"
+  "Serialize and retrieve Metabase data."
+  [:as {{:strs [all-collections collection settings data-model field-values database-secrets]
+         :or   {settings        true
+                data-model      true
+                all-collections true}}
+        :query-params}]
+  {collection       [:maybe [:or ms/PositiveInt [:sequential ms/PositiveInt]]]
```
qnkhuat, line 198:
```suggestion
  {collection_ids       [:maybe [:or ms/PositiveInt [:sequential ms/PositiveInt]]]
```

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+   all-collections  [:maybe ms/BooleanValue]
```
qnkhuat, line 199:
```suggestion
   all_collections  [:maybe ms/BooleanValue]
```

we use underscore pattern across APIs, I think we should be consistent here.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+      (let [res (io/input-stream (if (instance? File data)
+                                   (ba-copy data)
+                                   data))]
+        (callback)
+        res))))
+
+;;; Storage
+
+(def parent-dir "Dir for storing serialization API export-in-progress and archives."
+  (let [f (io/file (System/getProperty "java.io.tmpdir") "serdesv2-api")]
```
qnkhuat, line 62: `(System/getProperty "java.io.tmpdir")`

does this give a unique temp folder for each java process?

if not, we need to make the name of this folder randomized, otherwise this folder will be shared if we run multiple metabase instances on a single machine.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: also, why don't we use the same parameter names that we use in the cmd?
 `include-database-secrets` is a better name than `database-secrets` IMO.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: looks like it's platform dependent, https://stackoverflow.com/questions/16352326/when-does-system-getpropertyjava-io-tmpdir-return-c-temp

I think we should make the folder name randomized just to be sure.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+    (.mkdirs f)
+    (.deleteOnExit f)
+    (.getPath f)))
+
+(defn- compress-tgz [^File src ^File dst]
```
qnkhuat, line 67: wdyt if we move `compress-tgz` and `uncompress-tgz` to its own namespace?

and we need to add a test for it, the simplest I can think of is a folder after compress-tgz then uncompress-tgz should be identical.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+                        (GzipCompressorInputStream.)
+                        (TarArchiveInputStream.))]
+      (doseq [^TarArchiveEntry e (entries tar)]
+        (when-not @dir-name
+          (reset! dir-name (.getName e)))
+        (let [f (io/file dst (.getName e))]
+          (if (.isFile e)
+            (io/copy tar f)
+            (.mkdirs f)))))
+    @dir-name))
```
qnkhuat, line 105: sorry but I'm a bit nitty with the use of atom.

this is a version that doesn't use atom, it also a bit faster because we dont' need to deref `dir-name` in the loop.


```clojure
(defn- uncompress-tgz [^File archive ^File dst]
  (with-open [tar (-> (io/input-stream archive)
                      (GzipCompressorInputStream.)
                      (TarArchiveInputStream.))]
    (let [tar-entries (entries tar)]
      (doseq [^TarArchiveEntry e tar-entries]
        (let [f (io/file dst (.getName e))]
          (if (.isFile e)
            (io/copy tar f)
            (.mkdirs f))))
      (when-some [f (first tar-entries)]
        (.getName f)))))
```

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+    (io/copy f baos)
+    (.toByteArray baos)))
+
+(defn- on-response! [data callback]
+  (reify
+    ;; Real HTTP requests and mt/user-real-request go here
+    ring.protocols/StreamableResponseBody
+    (write-body-to-stream [_ response out]
+      (ring.protocols/write-body-to-stream data response out)
+      (if *in-tests*
```
qnkhuat, line 46:
```suggestion
      (if *in-tests*
```

binding this in tests shouldn't work because `mt/user-real-request` runs in its own thread.

it should only works with `mt/user-http-request`.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+          (.removeLogger config (.getName logger))
+          (.stop appender)
+          ;; this method is only present in AbstractConfiguration
+          (.removeAppender config (.getName appender))
+          (.updateLoggers (logger/context)))))))
+
+;;; Logic
+
+(defn- serialize&pack ^File [opts]
+  ;; TODO: how to get time in local timezone?
```
qnkhuat, line 143: > TODO: how to get time in local timezone?

`(t/local-date-time)` should do the trick?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,237 @@
...
+  (let [id       (format "%s-%s"
+                         (u/slugify (public-settings/site-name))
+                         (u.date/format "YYYY-MM-dd_HH-mm" (t/zoned-date-time)))
+        path     (io/file parent-dir id)
+        dst      (io/file (str (.getPath path) ".tar.gz"))
+        log-file (io/file path "export.log")]
+    (with-open [_logger (make-logger 'metabase-enterprise.serialization log-file)]
+      (try                              ; try/catch inside logging to log errors
+        (-> (extract/extract opts)
```
qnkhuat, line 152: do we need `serdes/with-cache` here?

enterprise/backend/test/metabase_enterprise/serialization/api_test.clj
```diff
@@ -0,0 +1,146 @@
...
+                             Dashboard     _     {:collection_id (:id coll)}
+                             Card          _     {:collection_id (:id coll)}]
+                (testing "API respects parameters"
+                  (let [f (mt/user-http-request :crowberto :post 200 "ee/serialization/export" {}
+                                                :all-collections false :data-model false :settings true)]
+                    (is (= #{:log :dir :settings}
+                           (with-open [tar (open-tar f)]
+                             (->> (entries tar)
+                                  (map (fn [^TarArchiveEntry e] (file-type (.getName e))))
+                                  set))))))
```
qnkhuat, line 69: looks like this could be a helper fn.

enterprise/backend/test/metabase_enterprise/serialization/api_test.clj
qnkhuat, line 1: we need permission tests

test/metabase/http_client.clj
```diff
@@ -153,18 +169,23 @@
...
-                     (if (map? credentials)
-                       (authenticate credentials)
-                       credentials))}
-    :cookie-policy :standard
-    :content-type :json}
-   (when (seq http-body)
-     {:body (json/generate-string http-body)})))
+  [credentials http-body request-options]
+  (let [content-type (or (get-in request-options [:headers "content-type"]) "application/json")]
+    (merge-with #(if (map? %1)
```
qnkhuat, line 174: IMO `merge` combined with a series of `(when ...)` is easier to read because I don't need to understand this lambda function.

test/metabase/http_client.clj
```diff
@@ -277,27 +298,24 @@
...
-         request-options)
-        (m/update-existing :body (fn [http-body]
-                                   (java.io.ByteArrayInputStream.
-                                    (.getBytes (if (string? http-body)
-                                                 ^String http-body
-                                                 (json/generate-string http-body)))))))))
+        url              (cond->> (first (str/split url #"\?")) ;; strip out the query param parts if any
+                           (not= (first url) \/)
+                           (str "/"))
+        content-type     (or (get-in request-options [:headers "content-type"]) "application/json")]
```
qnkhuat, line 304:
```suggestion
        content-type     (get-in request-options [:headers "content-type"] "application/json")]
```

super nitty :p

deps.edn
qnkhuat: maybe you should link to a PR that you fixed the bug you need, or comment on why we're using your fork.

qnkhuat (changes requested):
I don't have major concerns about a bunch of small suggestions.

Notebally:
- we should make sure the `parent-dir` is unique per metabase process
- API parameter name should match the cmd
- we need more tests for permission

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: I've been thinking about that back and forth... but yeah, why not 😁

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: it's going to look like `collection_ids=1&collection_ids=2`. To be in line with how query params are generated usually we could go with `collection-id=1&collection-id=2` then, but I don't see how that is an improvement.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: Oh, do we? That's kind of against the usual patterns of URLs... I thought we're doing underscores in JSON since it's easier (no quotes around) to use as keys in JS, but for query parameters to have underscore... it's like having underscores in command line options.

As for `include-database-secrets` — I think it's named like that because it's an argument-less option, you don't say `--include-database-secrets=true`, as you do in URL.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: I don't think this will work since tar as a stream is consumed already, so you'll have to seek back/restart this, and derefing will be much faster than that. Even so, with i/o taking place here performance of derefing is going to be negligible and it's much simpler than seeking back consumed streams.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: Hahaha, right! Well, it doesn't make sense for `user-http-request` since tests need no use of `future`. I wonder if passing bindings to a thread inside `user-real-request` will break anything?.. I really don't want to introduce a potential flake here. :)

Do you know why it's in a separate thread, is there a reason for this?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: Nope, still formats result in UTC. :\

test/metabase/http_client.clj
piranha: Ehmm but it's not the same? I'd have just `merge-with merge`, but it breaks if two maps have the same non-map top-level key, like `(merge-with merge {:a 1} {:a 2})`. So this lambda is saying "merge-with merge if vals are maps, else just merge". Let me see if extract this will make it clearer.

enterprise/backend/test/metabase_enterprise/serialization/api_test.clj
piranha: I want to see real requests are passing, since it's possible to return something that will make `user-http-request` pass — it gets raw body you returned and transforms it itself, without invoking all the http server machinery - and does not work in reality. Like, curl saves broken data. So I want `user-real-request` here.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: if I understand correctly, `tar-entries` is a lazy-seq and it'll be cached the second time you read from it. so it should work. 

in fact I tried it and it works for me.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: > Well, it doesn't make sense for user-http-request since tests need no use of future

I thought you wanted to make sure the call back is called synchronously? 

> Do you know why it's in a separate thread, is there a reason for this?

I'm not sure I understand the question. anw we're making a real HTTP request with `user-real-request`, so the thread that handle the request is different from the thread we make the request.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: ah to be clear, I think we should keep the same name, but change from dash to underscore.

so `include-database-secrets` will be `include_database_secrets`

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: Well, there is no point in having `(future)` inside `make-input-stream`, since this _only_ used for tests, not for the real service. So right now, if `user-real-request` goes through the network (I somehow imagined it still went through some inner call) this potentially introduces a flake with future in the `write-body-to-stream`. Not sure if it's worth it then...

piranha:
> API parameter name should match the cmd

IIRC we agreed in the meeting that cmd params are not exactly nice and so we're going to do a better api this time. :)

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: yea, I think we can remove `future` for `make-input-stream`, but we definitely should use `future` in `write-body-to-stream`

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,160 @@
...
+        (catch Exception e
+          (log/error e "Error during serialization"))))
+    {:log-file log-file
+     :callback #(when (.exists dst)
+                  (run! io/delete-file (reverse (file-seq dst))))}))
+
+(comment
+  ;; dump on disk and see that it works
+  (serialize&pack {:no-data-model true :no-settings true :targets [["Collection" 31409]]})
+  (unpack&import (io/file "metabase_test.tar.gz")))
```
qnkhuat, line 110: do you mean to keep this? it's kind of random IMO

qnkhuat:
> IIRC we agreed in the meeting that cmd params are not exactly nice and so we're going to do a better api this time. :)

do we have a plan to revamp the cmd params in the near future? if so when?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: > but I don't see how that is an improvement

Why not tho? IMO, the more specific, the better here since it's a public API.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: Eh, it kind of is, but I used it so many times that it felt useful to keep it in. :) But yeah, after a pause looks too random. 😁

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,160 @@
...
+
+;;; HTTP API
+
+(api/defendpoint POST "/export"
```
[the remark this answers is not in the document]
piranha: Ahh that's how we document our parameters! Just in a docstring, got it.

piranha:
> Currently, we have settings, field_values are both boolean, but how they are applied is the opposite in which settings=true means we exclude settings from the export, but when field_values=true means we include the field values in the export.

How so? It's actually other way around, HTTP API has a set of boolean parameters, which work in the same way: `true` means `include`, `false` means exclude. While CLI has `--no-settings` and `--include-field-values`, which include boolean in their name. This makes some sense for a command-line interface, since boolean options do not get any parameters, but less so for an HTTP API, where "naked" parameters (with no value passed to them) are much less common.

src/metabase/logger.clj
```diff
@@ -47,14 +52,88 @@
...
+       (.setTarget out)))))
+
+(defn for-ns
+  "Create separate logger for a given namespace to fork out some logs"
+  ^AutoCloseable [ns out & [{:keys [additive]
+                             :or   {additive true}}]]
+  (let [config        (configuration)
+        parent-logger (effective-ns-logger ns)
+        appender      (make-appender out (logger-name ns) (find-logger-layout parent-logger))
+        logger        (LoggerConfig. (logger-name ns) Level/INFO additive)]
```
qnkhuat, line 125: nit: look like the level can be parameterized

test/metabase/logger_test.clj
```diff
@@ -56,4 +57,21 @@
...
+                   (logger/messages))))))
+
+(deftest fork-logs-test
+  (testing "logger/for-ns works properly"
+    (let [f (io/file (System/getProperty "java.io.tmpdir") (mt/random-name))]
+      (try
+        (with-open [_ (logger/for-ns 'metabase.logger-test f {:additive false})]
+          (log/info "just a test"))
+        (is (=? #".*just a test\n"
+                (slurp f)))
```
qnkhuat, line 69: can you add tests to make sure after we close the outputstream, the logger doesn't keep appending to it?

qnkhuat:
> true means include, false means exclude
ah I forgot you flipped the boolean. 
It makes sense now :). I'm 👍 with the argument names.

I think the main logic

enterprise/backend/src/metabase_enterprise/serialization/api.clj
qnkhuat: bump

piranha: ah right, I need to remove `*in-tests*`. Now the only problem is that I'm not sure how to control that `future` is executed in time and this test does not become a flake. Any thoughts on this?..

qnkhuat (approved):
LGTM now, thank you for addressing all of my suggestions!

luizarakaki (approved):
Very nice work!

PR #36858, opened 2023-12-15
Update `with-report-timezone-id` and related

### LINKED ISSUE #36852: Update`with-report-timezone-id` to restore pools afterwards
While working on the issue #21518, after merging PR #35995 UTC night CI failures started to occur. More context on that can be found in [this slack thread](https://metaboat.slack.com/archives/C5XHN8GLW/p1701772468810829). Source of the problem was that after forcing a report timezone through `mt/with-report-timezone-id`, pool with timezone set was not disacrded, but reused in other tests. We should ensure that pool is reset after done executing body of `mt/with-report-timezone`.

lbrdnk:
Fixes  #36852

_Description_

This PR does following:
1. Adds `driver.tu` namespace containing vars for _general driver_ testing. That is currently only database pool restoration code.
2. Removes `notify-database-updated` _function_. Originally was meant as general implementation for jdbc _method_ of a same name. It currently however calls only `invalidate-pool-for-db!` which was added later. I've decided to remove former as the name of the second is more descriptive.
3. Swaps specific usages of `set-pool!` call that were used with same intent as `invalidate-pool-for-db!`.
4. ~~Renames `notify-database-updated!` as per Cam's comment.~~
5. And modifies `with-report-timezone-id` and `with-system-timezone-id` so pool is restored after thunk execution is completed. Linked issue is about former. Latter I've added (with comment) preemptively.
6. ~~Removed chunk that tries to handle snowflake failures that were happening for _probably_ the same reason as those in linked issue.~~
6. Add TODO instead of removing that snippet. Let's be conservative and remove it in future, when I can properly verify we don't need that in place.

_Testing_

No new tests were added. This PR adds simple testing macro and function. Please let me know in comments if I should add also tests for those.


src/metabase/driver.clj
```diff
@@ -708,18 +708,15 @@
   query)
 
 ;; TODO - we should just have some sort of `core.async` channel to handle DB update notifications instead
-;;
-;; TODO -- shouldn't this be called `notify-database-updated!`, since the expectation is that it is done for side
```
lbrdnk, line 712: I've updated the `notify-database-updated!` because of this comment. If it is not desired please let me know and I'll revert.

src/metabase/driver/sql_jdbc/connection.clj
```diff
@@ -215,13 +215,6 @@
   [database]
   (set-pool! (u/the-id database) nil nil))
 
-(defn notify-database-updated
```
lbrdnk, line 218: As mentioned in PR description, this (or rather the function above) adds level of indirection that we are not using currently. I've decided to leave in `invalidate-pool-for-db!` because of more descriptive name.

test/metabase/driver/postgres_test.clj
```diff
@@ -704,7 +704,7 @@
   (t2.with-temp/with-temp [Database database {:engine :postgres, :details (enums-test-db-details)}]
     (sync-metadata/sync-db-metadata! database)
     (f database)
-    (#'sql-jdbc.conn/set-pool! (u/id database) nil nil)))
```
lbrdnk, line 707: This line is the actual implementation of `invalidate-pool-for-db!` called now in `:sql-jdbc` implementation of method on following line.

test/metabase/query_processor/middleware/parameters/mbql_test.clj
```diff
@@ -94,10 +94,6 @@
 (deftest ^:parallel date-ranges-e2e-test
   (mt/test-drivers (params-test-drivers)
     (testing "check that date ranges work correctly"
-      ;; Prevent an issue with Snowflake were a previous connection's report-timezone setting can affect this test's
-      ;; results
-      (when (= :snowflake driver/*driver*)
-        (driver/notify-database-updated driver/*driver* (mt/id)))
```
lbrdnk, line 100: I'd like to find out if this is resolved by this PR. If not it will be reverted.

test/metabase/query_processor/test_util.clj
```diff
@@ -580,14 +581,10 @@
...
   [timezone-id thunk]
   {:pre [((some-fn nil? string?) timezone-id)]}
-  ;; This will fail if the app DB isn't initialized yet. That's fine — there's no DBs to notify if the app DB isn't
-  ;; set up.
-  (try
-    (#'driver/notify-all-databases-updated)
-    (catch Throwable _))
-  (binding [qp.timezone/*report-timezone-id-override* (or timezone-id ::nil)]
-    (testing (format "\nreport timezone id = %s" (pr-str timezone-id))
-      (thunk))))
```
lbrdnk, line 590: Now we are cleaning up connections in `do-with-report-timezone-id`. I believe that is desired.

test/metabase/test/util/timezone.clj
```diff
@@ -19,19 +18,17 @@
...
-        (try
-          (TimeZone/setDefault new-timezone)
-          (System/setProperty "user.timezone" timezone-id)
-          (t/testing (format "JVM timezone set to %s" timezone-id)
-            (thunk))
-          (finally
-            (TimeZone/setDefault original-timezone)
-            (System/setProperty "user.timezone" original-system-property)))))))
+        ;; Pre-emptively __wrapping__ try block in [[driver.tu/wrap-notify-all-databases-updated]]. I'm not sure that
+        ;; is currently necessary and I'm doing that to avoid surprises as with issue #36852 in future.
```
lbrdnk, line 22: Now connections are cleaned up also after `do-with-system-timezone-id`. Comment tries to explain why.

test/metabase/query_processor/middleware/parameters/mbql_test.clj
lbrdnk: Leaving just TODO in for now.

src/metabase/driver.clj
snoe: I would rather see this in a seperate pr, this would also need to update the driver changelog/docs/release notes for third party drivers. I'm not 100% sure of this process.

metamben: We don't rename driver methods like that. I think the standard procedure is deprecating the undesired thing (the old name in this case) but keeping it around for a few versions (not fewer than three, I guess). And yes, the change has to be documented in `docs/developers-guide/driver-changelog.md`.

test/metabase/driver/test_util.clj
```diff
@@ -0,0 +1,18 @@
...
+
+(defn notify-all-databases-updated []
+  ;; It makes sense to notify databases only if app db is initialized.
+  (when (initialize/initialized? :db)
+    (initialize/initialize-if-needed! :plugins)
+    (#'driver/notify-all-databases-updated)))
+
+(defmacro wrap-notify-all-databases-updated
+  [& body]
+  `(do (notify-all-databases-updated)
```
metamben, line 14: Is this really needed?

metamben (changes requested):

test/metabase/driver/test_util.clj
lbrdnk: If _this_ refers to `do` block, I could theoretically put `notify-all-databases-updated` in the `try` block. But then, if it throws, it would be called again in `finally`. And that's something I'd like to avoid. Please let me know if there is some error in my reasoning. Thanks.

src/metabase/driver.clj
lbrdnk: Thanks for explanation. Change was removed.

lbrdnk:
@snoe  and @metamben, thank you for the review. I believe I've addressed your comments.

test/metabase/driver/test_util.clj
metamben: This refers to the `(notify-all-databases-updated)` call. Are there two such calls because we don't trust tests clean up after themselves properly?

metamben (approved):

test/metabase/driver/test_util.clj
lbrdnk: I believe `mt/with-report-timezone-id` should do the clean up when done. See the change to `do-with-report-timezone-id`.

PR #36860, opened 2023-12-15
Update hardcoded plugins to use MB_PLUGINS_DIR

escherize:
We had one last spot hardcoding `"plugins/"`. That's been replaced with a call to `plugins-dir` which reads the path from `MB_PLUGINS_DIR` (and falls back to `"plugins/"`).

To verify:

Build with `bin/build.sh` then launch the app with:

```
MB_PLUGINS_DIR=~/path/to/custom/plugins java -jar metabase.jar
```

And see the logs indicate that:

```
Loading Analytics Content Complete (264) entities loaded.
```

Leaving out MB_PLUGINS_DIR works too.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/audit_db_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/audit_db_test.clj
@@ -68,23 +68,20 @@
         (is (= ::audit-db/no-op (audit-db/ensure-audit-db-installed!)))
         (t2/update! Database :is_audit true {:engine "h2"})))))
 
-(deftest audit-db-instance-analytics-content-is-copied-properly
-  (fs/delete-tree "plugins/instance_analytics")
-  (is (not (contains? (set (map str (fs/list-dir "plugins")))
-                      "plugins/instance_analytics")))
-
-  (#'audit-db/ia-content->plugins)
-  (is (= #{"plugins/instance_analytics/collections"
-           "plugins/instance_analytics/databases"}
-         (set (map str (fs/list-dir "plugins/instance_analytics"))))))
-
-(deftest audit-db-instance-analytics-content-is-copied-from-mb-plugins-dir-test
+(deftest instance-analytics-content-is-copied-to-mb-plugins-dir-test
   (mt/with-temp-env-var-value [mb-plugins-dir "card_catalogue_dir"]
     (let [plugins-dir (plugins/plugins-dir)]
       (#'audit-db/ia-content->plugins)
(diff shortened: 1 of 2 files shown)


enterprise/backend/test/metabase_enterprise/audit_db_test.clj
```diff
@@ -93,3 +99,7 @@
            #"Cannot sync Database: It is the audit db."
            (#'task.sync-databases/sync-and-analyze-database! "job-context"))))
     (is (= 0 (count (get-audit-db-trigger-keys))) "no sync occured even when called directly for audit db.")))
+
+(deftest plugins-path-is-plugins-str-test
+  (is (= (fs/path (plugins/plugins-dir))
+         (fs/path (str (plugins/plugins-dir))))))
```
noahmoss, line 105: what's the purpose of this test?

noahmoss (approved):
Stamping because the fix is trivial and we want to get it out. However would these tests definitely catch the issue in CI? This was on the code path that's only run in a JAR. Might be worth double checking that.

enterprise/backend/test/metabase_enterprise/audit_db_test.clj
escherize: To show that the types used will continue to work. Previously we passed a string , and now we are passing a sun.nio.fs.UnixPath.

escherize: Having another look at this, I am going to pass a string instead.

enterprise/backend/test/metabase_enterprise/audit_db_test.clj
```diff
@@ -32,7 +33,7 @@
     (testing "Audit DB content is not installed when it is not found"
       (t2/delete! :model/Database :is_audit true)
       (with-redefs [audit-db/analytics-dir-resource nil]
-        (is (= nil audit-db/analytics-dir-resource))
+        (is (= nil @#'audit-db/analytics-dir-resource))
```
johnswanson, line 36: Nit and it's already this way in the existing code, but `(nil? ...)` is very slightly more concise.

johnswanson (approved):

enterprise/backend/test/metabase_enterprise/audit_db_test.clj
noahmoss: Agree. Going to merge and put up a separate PR for this so that I don't need to wait for CI to run again before starting the backport

PR #36863, opened 2023-12-15
Apply formatters to JSON streaming results

### LINKED ISSUE #36554: Apply Consistent Formatting to JSON Artifacts
With #36320 we are applying consistent formatting to CSV artifacts (downloads and email attachments) as we do with HTML formatting. This _should_ also align better with FE rendering but that pipeline is not yet unified.

The product of this task is to apply the same `metabase/formatter` code to JSON artifacts as CSV artifacts.

This was done by updating `metabase.query-processor.streaming.csv` to compose the old static viz formatting code with the object to string code that was already present. This PR will likely do the same thing or something similar in `metabase.query-processor.streaming.json`.

AC:
* Add tests to check that formatting is applied to JSON attachments in `metabase.pulse.pulse-integration-test`.
  * `apply-formatting-in-csv-dashboard-test` is a good example to start with
  * Look for opportunities to refactor if they present themselves
* Demonstrate to product with loom recording properly formatted downloads and attachments

markbastian:
As is done in `metabase.query-processor.streaming.csv`, we now use column formatting in `metabase.query-processor.streaming.json`. This was done by creating a formatter for each column in the `:json` implementation of `qp.si/streaming-results-writer` (in the `begin!` function). These formatters were then applied on each row write.

One important design decision that had to be made here was the handling of JSON numeric types. In Metabase, we do several default formats of numbers and allow custom formatting that would alter the basic presentation of a json number. Examples:
- Integers default to showing a comma separator for large numbers
- Decimals round to 2 significant digits
- Percent, scientific, and currency formats may be applied

The only time a raw JSON number will look identical to what is shown in metabase is if formatting is removed for integers and just the right amount of digits are shown for floats.

Since the goal of exports is to preserve the data as shown in Metabase, the ony way to guarantee this outside of a few corner cases is to export all values as strings. Attempting to parse ints and floats after formatting may also cause unexpected behavior, as `3.14159` may be shown as `3.14` in Metabase and exported as this truncated number (if we parsed it). A user most likely would want to see exactly what they see in Metabase (so default to strings) or the original, unmodified value).

This PR also updates the CSV export code to use a `volatile!` rather than an `atom` for consistency.

Fixes #36554

### DIFF
--- a/src/metabase/query_processor/streaming/csv.clj
+++ b/src/metabase/query_processor/streaming/csv.clj
@@ -26,13 +26,12 @@
 (defmethod qp.si/streaming-results-writer :csv
   [_ ^OutputStream os]
   (let [writer             (BufferedWriter. (OutputStreamWriter. os StandardCharsets/UTF_8))
-        ordered-formatters (atom {})]
+        ordered-formatters (volatile! {})]
     (reify qp.si/StreamingResultsWriter
       (begin! [_ {{:keys [ordered-cols results_timezone]} :data} viz-settings]
-        (swap! ordered-formatters (constantly
-                                    (mapv (fn [col]
+        (vreset! ordered-formatters (mapv (fn [col]
                                             (formatter/create-formatter results_timezone col viz-settings))
-                                          ordered-cols)))
+                                          ordered-cols))
         (csv/write-csv writer [(map (some-fn :display_name :name) ordered-cols)])
         (.flush writer))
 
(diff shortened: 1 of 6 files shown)


adam-james-v (approved):
Looks good. Will be interesting to see what we decide in the future about strings/inferring stuff with JSON exports, but I think this is an important step for formatting consistency, nice!

PR #36866, opened 2023-12-15
When updating an alert, don't also update the question

### LINKED ISSUE #36395: Update alert shouldn't call update card API
### Describe the bug

When updating an alert, we call `PUT /api/card/:id` before calling `PUT /api/alert/:id` even though the card didn't change

### To Reproduce

1. Setup email
2. Create a new card
3. Setup alert and save

### Expected behavior

We probably don't need the update card API call

### Logs

_No response_

```JSON
master
```

### Severity

low, but it slows down the application and burn uncessary CPU cycles

### Additional context

_No response_

rafpaf:
Fixes #36395

### Description

On master, when you edit an alert, an additional `PUT` request is to send to update the card. This is not needed, because you can't edit an alert without first saving the question. On this branch, this request is not done.

### How to verify

On master:

1. Set up a new card and a new alert
2. Edit the alert
3. Note in the DevTools Network tab the PUT request to /api/card/:id

On this branch, do the same thing and note that there is no additional request.

### Demos

* [On master, updating the alert triggers a `PUT` request to /api/card:id](https://www.loom.com/share/4090bee88bf242abad12ff5b8b94b4f4?sid=98ef6bec-0d7a-48be-a155-b3f7ec794c62)
* [On this branch, updating the alert does not trigger a `PUT` request to /api/card/:id](https://www.loom.com/share/0558f085f70e40508e24642c152e3094?sid=a48ffa8a-5698-4765-81fc-5f311e8aeb98)

### Checklist

<s>Since this PR removes unneeded functionality, I haven't written a test for this. Of course, reviewers, please let me know if you think a test would make sense here.</s> I've now written a test, per @npfitz's request.

### DIFF
--- a/e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
+++ b/e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
@@ -66,7 +66,9 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
 
     saveAlert();
 
-    cy.findByText("Your alert is all set up.").should("be.visible");
+    cy.findByTestId("toast-undo").within(() => {
+      cy.findByText("Your alert is all set up.").should("be.visible");
+    });
 
     cy.wait("@savedAlert").then(({ response: { body } }) => {
       expect(body.channels[0].channel_type).to.eq("email");
@@ -74,13 +76,19 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
     });
   });
 
-  it.only("should enable alert to be updated (without updating question) (metabase#36866)", () => {
+  it("should enable alert to be updated (without updating question) (metabase#36866)", () => {
     openTable({
       table: PEOPLE_ID,
     });
 
     saveAlert();
-    cy.findByText("Your alert is all set up.").should("be.visible");
+
+    cy.log("Check that /api/card has been called once");
+    cy.get("@saveCard.all").should("have.length", 1);
+
+    cy.findByTestId("toast-undo").within(() => {
+      cy.findByText("Your alert is all set up.").should("be.visible");
+    });
 
     clickAlertBell();
 
@@ -89,23 +97,26 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
       cy.findByText("Edit").click();
     });
 
-    cy.findByText("How often should we check for results?")
-      .parent()
-      .parent()
-      .within(() => {
-        const buttons = cy.findAllByTestId("select-button");
-        buttons.should("have.length", 2);
-        buttons.eq(0).click();
-      });
+    cy.log("Change the frequency of the alert to weekly");
+
+    cy.get(".Modal--full").within(() => {
+      cy.findByText("How often should we check for results?")
+        .parent()
+        .parent()
+        .within(() => {
+          const buttons = cy.findAllByTestId("select-button");
+          buttons.should("have.length", 2);
+          buttons.eq(0).click();
+        });
+    });
+
     const weekly = cy.findByRole("option", { name: "Weekly" });
     weekly.should("have.attr", "aria-selected", "false");
     weekly.click();
     cy.findByRole("button", { name: "Save changes" }).click();
 
-    // FIXME: This doesn't work. Maybe look for "Your alert was set up" in the UI? Does that appear on master when we update the alert?
-    cy.wait("@saveCard").then(interception => {
-      assert.isUndefined(interception, "Route was not called");
-    });
+    cy.log("Check that /api/card has still only been called once");
+    cy.get("@saveCard.all").should("have.length", 1);
   });
 });
 
--- a/e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
+++ b/e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
@@ -99,7 +99,7 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
 
     cy.log("Change the frequency of the alert to weekly");
 
-    cy.findByRole("dialog", { name: "Edit your alert" })
+    cy.findByTestId("alert-edit")
       .findByText("How often should we check for results?")
       .parent()
       .parent()
--- a/e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
+++ b/e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
@@ -64,19 +64,9 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
       table: PEOPLE_ID,
     });
 
-    cy.icon("bell").click();
+    saveAlert();
 
-    cy.findByRole("dialog").within(() => {
-      cy.findByLabelText("Name").type(" alert");
-      cy.findByRole("button", { name: "Save" }).click();
-    });
-
-    cy.wait("@saveCard");
-
-    cy.findByTestId("alert-education-screen").within(() => {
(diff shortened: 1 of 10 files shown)


npfitz (dismissed):
The change looks good, but do you think we can add some e2e test coverage for this? We may be able to add a test in `email-alert.cy.spec.js` without much hassle.

perivamsi:
dont think this needs to be backported?

rafpaf:
> dont think this needs to be backported?

@perivamsi: I previously labeled this bugfix with `backport`, based on the [advice here](https://www.notion.so/metabase/Metabase-Branching-Strategy-6eb577d5f61142aa960a626d6bbdfeb3#33a1370301b048bda2a127584ce4b694), but following your guidance I've now labeled this `no-backport`.

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
```diff
@@ -64,25 +64,60 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
...
+    clickAlertBell();
+
+    cy.findByTestId("popover").within(() => {
+      cy.findByText("You set up an alert").should("be.visible");
+      cy.findByText("Edit").click();
+    });
+
+    cy.log("Change the frequency of the alert to weekly");
+
+    cy.get(".Modal--full").within(() => {
```
rafpaf, line 102: Another approach would be to add some attributes to [this full-page modal](https://github.com/metabase/metabase/blob/0634baa6200f17847d817ca38af0baa0e5ff4065/frontend/src/metabase/components/Modal/FullPageModal.tsx#L89), such as `role="dialog"` and `aria-labelledby="Edit your alert"` – or rather, the modal could have a `name` as a prop, and then we'd do `aria-labelledby={name}` – and then we could do `cy.findByRole('dialog', {name: "Edit your alert"})`, which is more user-centric. I'd be glad to do that, if reviewers prefer. Since that would involve adding a prop to a  For now, I've chosen a simpler option. I see the codebase already uses `cy.get('.Modal--full')` a few times, like [here](https://github.com/metabase/metabase/blob/0634baa6200f17847d817ca38af0baa0e5ff4065/e2e/test/scenarios/dashboard/dashboard.cy.spec.js#L877).

Related Slack thread: https://metaboat.slack.com/archives/C505ZNNH4/p1702557148449489?thread_ts=1702556878.526329&cid=C505ZNNH4

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
```diff
@@ -64,25 +64,60 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
...
+      cy.findByText("How often should we check for results?")
+        .parent()
+        .parent()
+        .within(() => {
+          const buttons = cy.findAllByTestId("select-button");
+          buttons.should("have.length", 2);
+          buttons.eq(0).click();
```
rafpaf, line 109: Perhaps there's an opportunity for refactoring the alert modal, since I think `data-testid` is typically a unique identifier.

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
iethree: good questions! 
- cy.get() is strongly discouraged. it exists in older tests, we shouldn't copy that pattern.
- accessible roles and other attributes that play nicely with react testing library are highly favored, but there are some existing helpers and tests that rely on a single "dialog" role being present, so if you're in a situation with multiple dialogs it could get messy.
- the easiest and safest option from a testing standpoint is using `data-testid`, it makes tests nice and easy to read because you can make it very specific, it's always safe to add, it's most likely to be stable, but it doesn't add any nice accessibility feature to the application.


TLDR be very bold:
200w

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
iethree: We don't usually enforce uniqueness. There are a lot of cases where we have lists of items with the same test id.

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
```diff
@@ -64,25 +64,60 @@ describe("scenarios > alert > email_alert", { tags: "@external" }, () => {
...
+        });
+    });
+
+    const weekly = cy.findByRole("option", { name: "Weekly" });
+    weekly.should("have.attr", "aria-selected", "false");
+    weekly.click();
```
iethree, line 115: cypress should let you chain all of these together

```ts
cy.findByRole("option", { name: "Weekly" })
  .should("have.attr", "aria-selected", "false")
  .click();
```

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
```diff
@@ -95,3 +130,25 @@ function openAlertForQuestion(id) {
 function toggleChannel(channel) {
   cy.findByText(channel).parent().find("input").click();
 }
+
+function clickAlertBell() {
+  cy.findByTestId("view-footer").within(() => {
+    cy.icon("bell").click();
+  });
```
iethree, line 137: if you just have a single assertion, you don't need .within

```suggestion
  cy.findByTestId("view-footer")
  	.icon("bell")
  	.click();
```

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
```diff
@@ -95,3 +130,25 @@ function openAlertForQuestion(id) {
...
+}
+
+function saveAlert() {
+  clickAlertBell();
+
+  cy.findByRole("dialog").within(() => {
+    cy.findByLabelText("Name").type(" alert");
+    cy.findByRole("button", { name: "Save" }).click();
```
iethree, line 145: we have a handy little shortcut for buttons:

https://github.com/metabase/metabase/blob/e21bde8b5d1daba523ed666e06b3ba2956e9690f/e2e/support/commands/ui/button.ts#L32

```suggestion
    cy.button( "Save" ).click();
```

iethree (approved):
Looks good! 🏆 

Some suggestions for improving the test.

e2e/test/scenarios/sharing/alert/email-alert.cy.spec.js
rafpaf: > accessible roles and other attributes that play nicely with react testing library are highly favored, but there are some existing helpers and tests that rely on a single "dialog" role being present, so if you're in a situation with multiple dialogs it could get messy.

I went ahead and tidied up some of the messiness around testing modals. The tidying happens in this new [refactoring PR](https://github.com/metabase/metabase/pull/37028). This "When updating an alert" PR now uses that refactoring PR as its base.

frontend/src/metabase/query_builder/components/AlertModals.jsx
```diff
@@ -286,16 +286,9 @@ class UpdateAlertModalContentInner extends Component {
...
-      question,
-      apiUpdateQuestion,
-      updateAlert,
-      updateUrl,
-      onAlertUpdated,
-    } = this.props;
+    const { question, updateAlert, updateUrl, onAlertUpdated } = this.props;
     const { modifiedAlert } = this.state;
 
-    await apiUpdateQuestion();
```
rafpaf, line 298: This is the core behavioral change in this PR: when we update an alert, we don't also update the question.

It might once have made sense to automatically update the question when the alert is updated, but no longer, I think, because it is now impossible to update an alert if the corresponding question has not yet been saved.

iethree (approved):
💖

PR #36868, opened 2023-12-15
Do not cache recently viewed items (#35379)

### LINKED ISSUE #35379: Recently viewed results are cached but should be fetched every time

rafpaf:
Closes #35379 

### Description

The problem is that recently viewed items are cached in the frontend, but this cache is never invalidated, so the user sees stale data.

### How to verify

Here's an illustration of the behavior, first on `master`, then on this feature branch. I've throttled the connection to "Slow 3G" to make it clearer when data is loading from the network:

https://www.loom.com/share/a686aa2791f0439c946aa0bf94f8d141?sid=f87c9a0d-59be-444c-afcd-dcf7d394eb39

As you can see, on `master`, when I visit a new page, the list of recently viewed items doesn't change. When I switch to this branch (`do-not-cache-recents`), the list changes after I visit a new page.

Since I made my internet deliberately slow, you can also see one downside of this approach, which is that the user must wait for the backend to provide accurate data. @iethree and I had a [discussion about this](https://metaboat.slack.com/archives/C0642RZ0E14/p1702647540969459) on Slack. As he suggested, it might eventually be nice to use the following pattern: the user sees a cached, possibly out-of-date list of recently viewed items right away; this list is then gracefully replaced with an up-to-date list once that information arrives over the network.

### Checklist

### DIFF
--- a/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
+++ b/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
@@ -53,6 +53,18 @@ describe("search > recently viewed", () => {
 
     cy.url().should("match", /\/question\/\d+-orders$/);
   });
+
+  it("shows up-to-date list of recently viewed items after another page is visited", () => {
+    openPeopleTable();
+    cy.findByTextEnsureVisible("Address");
+
+    cy.findByPlaceholderText("Search…").click();
+    // cy.findByTestId("loading-spinner").should("not.exist");
+
+    assertRecentlyViewedItem(0, "People", "Table");
+    assertRecentlyViewedItem(1, "Orders in a dashboard", "Dashboard");
+    assertRecentlyViewedItem(2, "Orders", "Question");
+  });
 });
 
 describeEE("search > recently viewed > enterprise features", () => {
--- a/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
+++ b/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
@@ -54,7 +54,7 @@ describe("search > recently viewed", () => {
     cy.url().should("match", /\/question\/\d+-orders$/);
   });
 
-  it("shows up-to-date list of recently viewed items after another page is visited", () => {
+  it("shows up-to-date list of recently viewed items after another page is visited (metabase#36868)", () => {
     openPeopleTable();
     cy.findByTextEnsureVisible("Address");
 
--- a/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
+++ b/e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
@@ -71,12 +71,8 @@ describe("search > recently viewed", () => {
 
     cy.findByPlaceholderText("Search…").click();
     cy.wait("@recent");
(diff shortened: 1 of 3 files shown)


e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
```diff
@@ -53,6 +53,20 @@ describe("search > recently viewed", () => {
...
+
+    cy.findByPlaceholderText("Search…").click();
+    cy.findByTestId("loading-spinner").should("not.exist");
+
+    assertRecentlyViewedItem(0, "People", "Table");
+    assertRecentlyViewedItem(1, "Orders in a dashboard", "Dashboard");
+    assertRecentlyViewedItem(2, "Orders", "Question");
+
+    cy.findAllByTestId("recently-viewed-item-title").should("have.length", 3);
+  });
```
rafpaf, line 69: @iethree The test we began writing in e2e/test/scenarios/onboarding/search/search.cy.spec.js wasn't working for some reason, but this similar test works

rafpaf:
This test fails in this branch but it also consistently fails on `master`: https://cloud.cypress.io/projects/ywjy9z/runs/473/overview/a5051abb-0f0a-466f-8860-c3668b0d5b3a?reviewViewBy=FAILED&utm_source=github&utm_medium=failed&utm_campaign=view%20test

e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
iethree: 🤔 I suspect that this test doesn't quite test what we want, and I think it would pass on master without your change - we don't do anything to assert that the recents endpoint gets called again and re-fetches updated data. 

I also think this test will be most useful on the home page, which has a recents endpoint call by default, which lets us assert that the search dropdown calls it again.

iethree (changes requested):
Code change looks good, but let's try to shore up the e2e test to meaningfully test the new behavior here.

e2e/test/scenarios/onboarding/search/recently-viewed.cy.spec.js
rafpaf: Thank you for that kind explanation 🙏🙏🙏

The test now uses a click (on line 70) rather than `cy.visit`, so I think it now properly tests the right behavior?

The test begins by visiting the home page (on line 34) - let me know if this isn't what you were referring to, though. Also, the test now fails on `master` and works on this branch.

Let me know if I'm missing anything!

iethree (approved):

PR #36870, opened 2023-12-15
Routine dependency bumps

noahmoss:
Routine dependency bumps for the start of the 49 cycle.

A few dependencies are excluded:
* Jetty, `info.sunng/ring-jetty9-adapter`, and `io.github.resilience4j/resilience4j-retry` are not bumped because the newer versions do not support Java 11
* H2 is not bumped because of another format upgrade between 2.1 and 2.2 ([Slack thread with more details](https://metaboat.slack.com/archives/C97M2U8M6/p1701904521480389))
* Liquibase is not bumped because it causes mysterious failures in (mostly) migration tests. Will investigate this separately. https://github.com/metabase/metabase/issues/36877

There also seems to be issues with the [new version of the Redshift driver](https://metaboat.slack.com/archives/C04DN5VRQM6/p1702506133795809) but I'm still bumping it here, [per Luis's comment](https://metaboat.slack.com/archives/CKZEMT1MJ/p1702665259619859?thread_ts=1677871576.302359&cid=CKZEMT1MJ)

### DIFF
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -20,7 +20,6 @@
    (liquibase Contexts LabelExpression Liquibase RuntimeEnvironment Scope Scope$Attr Scope$ScopedRunner)
    (liquibase.change.custom CustomChangeWrapper)
    (liquibase.changelog ChangeLogIterator ChangeSet ChangeSet$ExecType)
-   (liquibase.changelog.filter ChangeSetFilter)
    (liquibase.changelog.visitor AbstractChangeExecListener ChangeExecListener UpdateVisitor)
    (liquibase.database Database DatabaseFactory)
    (liquibase.database.jvm JdbcConnection)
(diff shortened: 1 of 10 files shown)


src/metabase/db/liquibase.clj
```diff
@@ -254,7 +255,7 @@
      :or {change-set-filters []}}]
    (let [change-log     (.getDatabaseChangeLog liquibase)
          database       (.getDatabase liquibase)
-         log-iterator   (ChangeLogIterator. change-log (into-array liquibase.changelog.filter.ChangeSetFilter change-set-filters))
+         log-iterator   (ChangeLogIterator. change-log (into-array ChangeSetFilter change-set-filters))
```
noahmoss, line 258: Made this change while trying to debug the Liquibase issues (unsuccessfully, so far) but didn't see a reason to revert it

metamben (approved):
👍

PR #36874, opened 2023-12-15
show error and disable save button when create metric fails

### LINKED ISSUE #36422: Show errors in the UI when the BE returns exception on metrics v1 creation
Our metrics v1 feature is limited. It generates a single query, regarless of metric composition or combination. If each metric has different filter sets, they will conflict and lead to correctness issues.
([#30574](https://github.com/metabase/metabase/issues/30574), [#25455](https://github.com/metabase/metabase/issues/25455), [#13437](https://github.com/metabase/metabase/issues/13437)

[Slack context](https://metaboat.slack.com/archives/C04DN5VRQM6/p1701699273567639)

We are now preventing composed/combined metrics creation, but the exception is not displayed in the UI.

If the BE returns an exception:
- [ ] turn the Save button red (failed state)
- [ ] show the error message somewhere

shaunlebron:
Fixes #36422

## Summary

The 400 error returned by `POST /api/metric` was hidden in the UI because MetricApp was catching the submission error inside `onSubmit` without rethrowing it.

After fixing that, wrapping the MetricForm in a FormProvider was the simplest way to handle the error button state and error message.

## How to review

Turn off whitespace, and look at e2e spec.


e2e/test/scenarios/admin/datamodel/reproductions/36422-show-create-metric-error.cy.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    selectTable("Orders");
+    cy.get('[name="name"]').type("x");
+    cy.get('[name="description"]').type("x");
+
+    cy.intercept("POST", "/api/metric", req => req.reply(400));
+    cy.button("Save changes").click();
+    cy.button("Failed")
+      .parent()
+      .parent()
+      .should("contain", "An error occurred");
```
npfitz, line 21: Same here. Ideally there is a role we can use to grab the status

npfitz (approved):
Aside from the test selectors, it looks good to me!

e2e/test/scenarios/admin/datamodel/reproductions/36422-show-create-metric-error.cy.spec.js
npfitz: Ran the test locally. I'd consider adding a role to the status text, or just adding a test id to the form footer on the page

rafpaf (approved):

PR #36875, opened 2023-12-15
add-implicit-clauses marks refs to ignore coercion

### LINKED ISSUE #33861: Timestamp column be converted twice in 2-layer questions query
### Describe the bug

I have a mysql table spam , and create a model on spam , then when create a Question based on this model.
One bigint column named create_timestamp of spam be translated twice from bigint to datetime by origin table and question layers .

Spam metadata config :
create_timestamp stored millisecond timestamp value of bigint type , configured to auto change timestamp to datetime in metabse table metadata.

in model A query : do nothing

in question based on model A : use create_timestamp in 'group by ' conditions , the raw sql is error , and create_timestamp by day is empty string

```
SELECT
  DATE(FROM_UNIXTIME(`source`.`create_timestamp` / 1000)) AS `create_timestamp`,  //second transform  in question layer
  COUNT(*) AS `count`
FROM
  (
    SELECT
      `source`.`user_id` AS `user_id`,
      `source`.`type` AS `type`,
      `source`.`origin_content` AS `origin_content`,
      `source`.`origin_type` AS `origin_type`,
      `source`.`status` AS `status`,
      `source`.`create_timestamp` AS `create_timestamp`
    FROM
      (
        SELECT
          `spam`.`user_id` AS `user_id`,
          `spam`.`type` AS `type`,
          `spam`.`origin_content` AS `origin_content`,
          `spam`.`origin_type` AS `origin_type`,
          `spam`.`status` AS `status`,
          FROM_UNIXTIME(`spam`.`create_timestamp` / 1000) AS `create_timestamp`  //first transform 
        FROM
          `spam`
       
WHERE
          (
            FROM_UNIXTIME(`spam`.`create_timestamp` / 1000) >= DATE(NOW(6))
          )
         
   AND (
            FROM_UNIXTIME(`spam`.`create_timestamp` / 1000) < DATE(DATE_ADD(NOW(6), INTERVAL 1 day))  
          )
      ) AS `source`
  ) AS `source`
GROUP BY
  DATE(FROM_UNIXTIME(`source`.`create_timestamp` / 1000))
ORDER BY
  DATE(FROM_UNIXTIME(`source`.`create_timestamp` / 1000)) ASC

### To Reproduce

1. origin table `spam` has bigint column, stored millisecond timestamp values 
2. create model (or question), the column changed to datetime correctly
3. create new question QB based on 2nd step model ( or question)
4. the datetime column changed from bigint to datetime again, got empty result

### Expected behavior

_No response_

### Logs

_No response_

```JSON
metabase V0.47.0
```

### Severity

 blocking my usage of Metabase entirely

### Additional context

Can I setting the metadata of the new datetime column in Model/Question?  So that, new questions based on this one can deal with this column like datetime type , not the bigint type in origin source table ?

snoe:
Fixes #33861

Renames `::nest-query/outer-select` to `qp/ignore-coercion` since it was
already applied outside nest-query and currently only controls that aspect.


lbrdnk (approved):
LGTM. Only change I'd recommend is adding white space as per my only comment.

I like the decision to rename to `:qp/ignore-coercion`. As the key is now used in more namespaces, there is no need for further requires because of that.

I was suspicious about other than `:sql` drivers handling of this situation. Locally modifying the new test to run with `:mongo` convinced me we are fine here. (Using no `mt/test-driver(s)` yields test running on h2, even when we are testing other drivers, right?)

Then, other `:sql` based drivers use parent method `->honeysql` for field dispatch, so those should be also fine.

PR #36889, opened 2023-12-18
Kondo: ignore all libraries config

qnkhuat:
I don't think we need to track these configs, it should be automatically copied when we start Kondo.
If not, you can run `./bin/kondo.sh` and it'll take care of that for you.

It frustrates me a little because everytime we use a new library, my local git will show an untracked folder of their config :)


calherries (approved):
Thanks, this has been a papercut for me for a while

PR #36892, opened 2023-12-18
Bigquery sync partitioned fields

qnkhuat:
Milestone 2 of https://github.com/metabase/metabase/issues/36668

This will sync whether a field is a partition field.

There is a special case for ingestion time partition because it'll be partitioned on a pseudo-column [_PARTITIONTIME or _PARTITIONDATE](https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table). So to support the milestone 3, where we need to make our sync adds a filter for tables that require a filter, I've changed the code so that these 2 pseudo-columns are added manually.


modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
```diff
@@ -269,21 +270,18 @@
                                 PARTITION BY
                                 transaction_date
                                 OPTIONS (
-                                partition_expiration_days = 3,
```
calherries, line 272: Can you explain why these were added before and now removed?

qnkhuat: I added it because it's from the documentation code, but they're irrelevant to the tests. Don't want to give people the wrong idea about what's being tested.

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
```diff
@@ -294,15 +292,65 @@
...
+                         (map (fn [{:keys [table_name field_name database_partitioned]}]
+                                [[table_name field_name] database_partitioned]))
+                         (into {})))))
+
            (finally
             (doseq [table-name (keys table-name->is-filter-required?)]
               (drop-table-if-exists! table-name)))))))))
 
+(deftest sync-pseudocolumn-for-ingestion-time-partitioned-table-test
+  (testing "for ingestion time partitioned tables, we should syncs the pseudocolumn _PARTITIONTIME"
```
calherries, line 328:
```suggestion
  (testing "for ingestion time partitioned tables, we should sync the pseudocolumn _PARTITIONTIME"
```

resources/migrations/001_update_migrations.yaml
```diff
@@ -4365,6 +4365,21 @@ databaseChangeLog:
...
+      changes:
+        - addColumn:
+            tableName: metabase_field
+            columns:
+              - column:
+                  name: database_partitioned
+                  type: boolean
+                  constraints:
+                    nullable: true
+                  remarks: 'Whether the field is a partitioned field'
```
calherries, line 4381: Technically fields aren't partitioned, tables are
```suggestion
                  remarks: 'Whether the table is partitioned by this field'
```

src/metabase/sync/interface.clj
```diff
@@ -44,6 +44,8 @@
    [:nfc-path                   {:optional true} [:any]]
    [:custom                     {:optional true} :map]
    [:database-is-auto-increment {:optional true} :boolean]
+   ;; nullable for databases that don't support field partition
+   [:database-partitioned       {:optional true} [:maybe :boolean]]
```
calherries, line 48: Why did you decide to make this nullable rather than just optional?

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -175,11 +175,30 @@
...
+     :fields (cond-> fields
+               ;; if table has time partition but no field is specified as partitioned
+               ;; meaning this table is partitioned by ingestion time
+               ;; so we manually sync this virtual column
+               (and (some? (.getTimePartitioning tabledef))
+                    (nil? partitioned-field-name))
+               (conj
+                {:name                 "_PARTITIONTIME"
+                 :database-type        "TIMESTAMP"
+                 :base-type            :type/DateTimeWithLocalTZ
```
calherries, line 199: Are you sure it shouldn't be `:type/DateTimeWithTZ`? Take a look at `bigquery.tx/base-type->bigquery-type`. Either way I think `driver/describe-table`'s implementation should avoid being coupled to the relationship between `:database-type` and `:base-type` for bigquery.

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -175,11 +175,30 @@
...
+        fields                 (set
+                                (map
+                                 #(assoc % :database-partitioned (= (:name %) partitioned-field-name))
+                                 (table-schema->metabase-field-info (. tabledef getSchema))))]
+    {:schema dataset-id
+     :name   table-name
+     :fields (cond-> fields
+               ;; if table has time partition but no field is specified as partitioned
+               ;; meaning this table is partitioned by ingestion time
+               ;; so we manually sync this virtual column
```
calherries, line 193: It's called a pseudo-column in [the docs](https://cloud.google.com/bigquery/docs/querying-partitioned-tables#better_performance_with_pseudo-columns), maybe we should be consistent
```suggestion
               ;; so we manually sync this pseudo-column
```

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -175,11 +175,30 @@
...
+               (and (some? (.getTimePartitioning tabledef))
+                    (nil? partitioned-field-name))
+               (conj
+                {:name                 "_PARTITIONTIME"
+                 :database-type        "TIMESTAMP"
+                 :base-type            :type/DateTimeWithLocalTZ
+                 :database-position    (count fields)
+                 :database-partitioned true}))}))
```
calherries, line 201: Based on [the docs](https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table), it looks like you also need to add a pseudo-column for `_PARTITIONDATE`.

calherries (changes requested):

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
calherries: Makes sense, just checking

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
```diff
@@ -294,15 +292,65 @@
...
+                                      :name [:in (keys table-name->is-filter-required?)]))))
+
+           (testing "partitioned fields are correctly identified"
+             (is (= {["not_partitioned"                 "transaction_id"]   false
+                     ["partition_by_range_not_required" "customer_id"]      true
+                     ["partition_by_range_not_required" "date1"]            false
+                     ["partition_by_range"              "customer_id"]      true
+                     ["partition_by_range"              "date1"]            false
+                     ["partition_by_ingestion_time"     "transaction_id"]   false
+                     ["partition_by_ingestion_time"     "_PARTITIONTIME"]   true
```
calherries, line 309: Do you need to add a test for `_PARTITIONDATE` too? ([docs](https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table))

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
qnkhuat: > Are you sure it shouldn't be :type/DateTimeWithTZ

from the types.cljc doc
```clojure
(derive :type/TimeWithLocalTZ :type/TimeWithTZ)    ; a column that is timezone-aware, but normalized to UTC or another offset at rest.
```

And also, this column has the type timestamp, and we are mapping timestamp to `:type/DateTimeWithLocalTZ`, so I'm pretty sure.

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
qnkhuat: `_PARTITIONDATE` is `_PARTITIONTIME` truncated to DATE. so I think we can sync only `_PARTITIONTIME`, for cases where we need `_PARTITIONDATE`, we can just truncate the time column.

> If the partition granularity is daily, the table also contains a pseudo-column named _PARTITIONDATE. The value is equal to _PARTITIONTIME truncated to a DATE value.

([ref](https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table))

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
qnkhuat: I think we don't need to sync `_PARTITIONDATE` column, see https://github.com/metabase/metabase/pull/36892/files#r1432189416

src/metabase/sync/interface.clj
qnkhuat: no strong reason, indeed we can make it tighter and remove the maybe.

4c19a94e14

qnkhuat: This schema not only being used for metadata from jdbc but also for field getting from `metabase_field`, see https://github.com/metabase/metabase/blob/4cf4e789cdaf47696db35f8630cb941a5edd17d1/src/metabase/sync/sync_metadata/fields/common.clj#L17

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
calherries: Makes sense, it just seemed a little strange to include one and not the other. What's the reason against doing so?

qnkhuat: >  What's the reason against doing so?

It's more about syncing one of the would serve our use cases. And since `_PARTITIONTIME` can be truncated to `DATE` so any use case that you need with `_PARTITIONDATE` can be handled by `_PARTITIONTIME`.

> One reason to include it is it is possible to use _PARTITIONDATE in native SQL, but it won't show in the data reference pane.

that might be a good reason to include it I think, this way, if you have the model, you can be sure you can map to the underlined column.

@luizarakaki do you think it makes sense to sync both of these pseudo columns?

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
calherries: Oh you're right, I should have been looking at `metabase.driver.bigquery-cloud-sdk/bigquery-type->base-type`, my bad. I think that's a problem with `bigquery.tx/base-type->bigquery-type` then.

Compare:
```clojure
;; metabase.test.data.bigquery-cloud-sdk
(defn base-type->bigquery-type [base-type]
  (let [types {:type/BigInteger     :INTEGER
               :type/Boolean        :BOOLEAN
               :type/Date           :DATE
               :type/DateTime       :DATETIME
               :type/DateTimeWithTZ :TIMESTAMP
               :type/Decimal        :NUMERIC
               :type/Dictionary     :RECORD
               :type/Float          :FLOAT
               :type/Integer        :INTEGER
               :type/Text           :STRING
               :type/Time           :TIME}]
    (or (get types base-type)
        (some base-type->bigquery-type (parents base-type)))))
```
and
```clojure
;; metabase.driver.bigquery-cloud-sdk
(defn- bigquery-type->base-type
  "Returns the base type for the given BigQuery field's `field-mode` and `field-type`. In BQ, an ARRAY of INTEGER has
  \"REPEATED\" as the mode, and \"INTEGER\" as the type name.

  If/when we are able to represent complex types more precisely, we may want to capture that information separately.
  For now, though, we will check if the `field-mode` is \"REPEATED\" and return our :type/Array for that case, then
  proceed to check the `field-type` otherwise."
  [field-mode field-type]
  (if (= Field$Mode/REPEATED field-mode)
    :type/Array
    (case field-type
      "BOOLEAN"    :type/Boolean
      "FLOAT"      :type/Float
      "INTEGER"    :type/Integer
      "RECORD"     :type/Dictionary ; RECORD -> field has a nested schema
      "STRING"     :type/Text
      "DATE"       :type/Date
      "DATETIME"   :type/DateTime
      "TIMESTAMP"  :type/DateTimeWithLocalTZ
      "TIME"       :type/Time
      "NUMERIC"    :type/Decimal
      "BIGNUMERIC" :type/Decimal
      :type/*)))
```

Anyway, my point about coupling still stands. You should use `base-type->bigquery-type` instead of hard-coding `:type/DateTimeWithLocalTZ`, like is done in `table-schema->metabase-field-info`.

qnkhuat: yes I did that in https://github.com/metabase/metabase/pull/36892/commits/4c19a94e14213e4b972416f02aed15455480e8ea

resources/migrations/001_update_migrations.yaml
```diff
@@ -4381,6 +4381,21 @@ databaseChangeLog:
...
+      comment: Add metabase_field.database_partitioned
+      changes:
+        - addColumn:
+            tableName: metabase_field
+            columns:
+              - column:
+                  name: database_partitioned
+                  type: boolean
+                  constraints:
+                    nullable: true
```
calherries, line 4396: Why did you decide to make this field nullable, rather than adding a default false value like `database_is_auto_increment`?

I prefer non-nullable constraints unless there's a reason not to, because it eliminate the possibility of accidentally handling `nil` values differently from `false` values.

src/metabase/sync/interface.clj
calherries: Sorry I realised the schema was the wrong place to put this comment. I've restated the question [here](https://github.com/metabase/metabase/pull/36892/files#diff-e6ec4fc8084d80ae2ebedde98c98f358ec7ac526de5c339ebcb454cbda62b221R4396)

resources/migrations/001_update_migrations.yaml
qnkhuat: not all databases support partition, so for fields from those databases, it should be `null.` it's in the same vein as `database_indexed`.

calherries: I don't understand your argument, but I guess there's a possible potential future where we want to handle these two cases differently:
- fields of databases that support partitions
- fields of databases that don't support partitions

So it might be helpful to differentiate them at the field level. I'm not sure either way so I'm okay with this

calherries (approved):

calherries (approved):
LGTM, I'm approving conditional on `_PARTITIONDATE` not being required

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
luizarakaki: The only reason to sync `_PARTITIONDATE` is for SQL queries.

If this doesn't impact our ability to enforce filters on `_PARTITIONTIME` and doesn't have other side-effects, I think it is a good idea to have a complete data reference. If this brings problems, I wouldn't add it.

My main concern is: if the table requires filtering on partition, having both columns would bring problems? If not, I think it is useful to add `_PARTITIONDATE` too.

qnkhuat: The only downside I can think of is this will break the assumption that a table has only one partition field.

qnkhuat: One downside here is that it'll break the assumption that a table has only one partitioned field.
In milestone 3 where I'll use these filter fields for sync field values and fingerprint of a partitioned table, this assumption will come in handy.

That's said if we don't have a strong reason to add this, I'd say let's not add it for now. We can always add it later if necessary.

@calherries thoughts?

calherries: > this assumption will come in handy.

Can you explain this more?

I had a brief look at [milestone 3](https://github.com/metabase/metabase/pull/37029) and couldn't find a reason why you can't have two fields with `:database-partitioned true`

qnkhuat: in this [part](https://github.com/metabase/metabase/pull/37029/files#diff-a3009f2547aa68da91de932bd775af0ce4df453500aff03c0fce30e1b6647171R42-R45) where I select the field I used as filtering field, it'd be nice to know for sure which field will be selected.

Tbh, it's not strictly needed, we can add a clause to that query to make sure either `PARTITIONTIME` or `PARTITIONDATE` is returned.

qnkhuat: alright I've synced them now

calherries (approved):

Issue #36891, opened 2023-12-18
BE flake: 404 Not found for bigquery dataset: metabase.driver.sql-jdbc.sync.describe-database-test/database-schema-filtering-test


calherries:
Flaky test: metabase.driver.sql-jdbc.sync.describe-database-test/database-schema-filtering-test

This failed twice for bigquery [here](https://github.com/metabase/metabase/actions/runs/7245617627/job/19735979059?pr=36826#step:3:1693) and [here](https://github.com/metabase/metabase/actions/runs/7171786155/job/19527544581#step:3:561)

```
                                       clojure.lang.ExceptionInfo: Invalid output: ["invalid type"]
                    error: {:schema
                            [:maybe [:sequential [:map [:start-time [:fn #:error{:message "Instance of a java.time.temporal.Temporal"} #object[clojure.core$partial$fn__5908 0x2d177426 "clojure.core$partial$fn__5908@2d177426"]]] [:end-time [:fn #:error{:message "Instance of a java.time.temporal.Temporal"} #object[clojure.core$partial$fn__5908 0x29ff22ec "clojure.core$partial$fn__5908@29ff22ec"]]] [:name :string] [:steps [:maybe [:sequential [:ref :metabase.sync.util/StepNameWithMetadata]]]]]]],
                            :value #error {
                            :cause "404 Not Found\nGET https://www.googleapis.com/bigquery/v2/projects/metabase-bigquery-driver/datasets/v4_KIYFJFAAYDOFNHSVZIXZ__transient_1702889390088/tables?prettyPrint=false\n{\n  \"code\": 404,\n  \"errors\": [\n    {\n      \"domain\": \"global\",\n      \"message\": \"Not found: Dataset metabase-bigquery-driver:v4_KIYFJFAAYDOFNHSVZIXZ__transient_1702889390088\",\n      \"reason\": \"notFound\"\n    }\n  ],\n  \"message\": \"Not found: Dataset metabase-bigquery-driver:v4_KIYFJFAAYDOFNHSVZIXZ__transient_1702889390088\",\n  \"status\": \"NOT_FOUND\"\n}"
                            :via
                            [{:type com.google.cloud.bigquery.BigQueryException
                              :message "Not found: Dataset metabase-bigquery-driver:v4_KIYFJFAAYDOFNHSVZIXZ__transient_1702889390088"
[...]

PR #36900, opened 2023-12-18
Fix flaking database-schema-filtering-test for bigquery

calherries:
Fixes BE flake https://github.com/metabase/metabase/issues/36891

My theory is the flake happens because the test takes so long to run, that there's a race condition and some datasets are changed while it's running. Running it locally it takes so long I didn't want to wait for it to finish, so I don't know exactly how long it takes.

There are two changes to tests:
1. I removed bigquery from the tested drivers in `metabase.driver.bigquery-cloud-sdk-test/database-schema-filtering-test`. The implementation of this test takes forever to run and was flaking out for bigquery, because the number of test datasets on our test DB is huge (181 by last count), and we were creating and syncing tables for almost every table in every test dataset.
2. I fixed the existing equivalent test for bigquery `metabase.driver.bigquery-cloud-sdk-test/dataset-filtering-test`, which wasn't actually testing anything before due to a bug in its implementation.

### DIFF
--- a/modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
+++ b/modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
@@ -63,23 +63,29 @@
 ;;; |                                                      Sync                                                      |
 ;;; +----------------------------------------------------------------------------------------------------------------+
 
+(defn- list-datasets
+  "Fetch all datasets given database `details`, applying dataset filters if specified."
+  [{:keys [project-id dataset-filters-type dataset-filters-patterns] :as details}]
+  (let [client (database-details->client details)
+        project-id (or project-id (bigquery.common/database-details->credential-project-id details))
+        datasets (.listDatasets client project-id (u/varargs BigQuery$DatasetListOption))
+        inclusion-patterns (when (= "inclusion" dataset-filters-type) dataset-filters-patterns)
+        exclusion-patterns (when (= "exclusion" dataset-filters-type) dataset-filters-patterns)]
+    (for [^Dataset dataset (.iterateAll datasets)
+          :let [^DatasetId dataset-id (.. dataset getDatasetId)]
+          :when (driver.s/include-schema? inclusion-patterns
+                                          exclusion-patterns
+                                          (.getDataset dataset-id))]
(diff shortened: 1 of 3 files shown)


modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -63,23 +63,29 @@
 ;;; |                                                      Sync                                                      |
 ;;; +----------------------------------------------------------------------------------------------------------------+
 
+(defn- list-datasets
```
calherries, line 66: The changes to this file is a refactoring, so I could use `list-datasets` in the test

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
```diff
@@ -409,30 +409,51 @@
...
-      (testing " with an exclusion filter"
-        (sync-and-assert-filtered-tables {:name    "BigQuery Test DB with dataset exclusion filters"
-                                          :engine  :bigquery-cloud-sdk
-                                          :details (-> (mt/db)
-                                                       :details
-                                                       (assoc :dataset-filters-type "exclusion"
-                                                              :dataset-filters-patterns "v*"))}
-          (fn [{dataset-id :schema}]
-            (is (not= \v (first dataset-id)))))))))
+      (mt/db) ;; force the creation of one test dataset
```
qnkhuat, line 420: why do we need the test dataset to be created?

calherries: Because to run the test in isolation, the test needs at least two datasets to exist in the bigquery database. One for testing inclusion and another for exclusion. For our bigquery test database that runs in CI there are hundreds, so this is a bit pedantic. But if you were to run this test on a fresh new bigquery test database this test would fail unless you created the two datasets beforehand.

piranha: Feels like this explanation deserves to be a comment in the code. :)

sansolovyov (approved):
LGTM

PR #36905, opened 2023-12-18
Exclude api key users from endpoints and billing calcuations

### LINKED ISSUE #36914: Exclude API Key users from the `/api/user` endpoint

johnswanson:
Pretty straightforward:

To exclude API Key users from billing calculations just set them to
`is_active: false`.

To exclude API Key users from the User API endpoints, use the new `type`
field on the User model to add conditions to each of the endpoints to
ensure that we're only listing, selecting, or modifying "personal"
users (as opposed to "internal" or "api-key").

Theoretically this means the `(check-not-internal-user)` calls are no
longer necessary, but I'm keeping them around to keep the nicer error
messages in that case.

Closes #36914

### DIFF
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -34,8 +34,7 @@
     (t2/with-transaction [_conn]
       (let [user (user/insert-new-user! {:email      email
                                          :first_name name
-                                         :type       :api-key
-                                         :is_active  false})]
+                                         :type       :api-key})]
         (user/set-permissions-groups! user [(perms-group/all-users) group_id])
         (-> (t2/insert-returning-instance! :model/ApiKey
                                            {:user_id      (u/the-id user)
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -34,7 +34,8 @@
     (t2/with-transaction [_conn]
       (let [user (user/insert-new-user! {:email      email
                                          :first_name name
-                                         :type       :api-key})]
+                                         :type       :api-key
+                                         :is_active  false})]
         (user/set-permissions-groups! user [(perms-group/all-users) group_id])
         (-> (t2/insert-returning-instance! :model/ApiKey
                                            {:user_id      (u/the-id user)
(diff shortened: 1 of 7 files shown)


src/metabase/api/api_key.clj
```diff
@@ -34,7 +34,8 @@
     (t2/with-transaction [_conn]
       (let [user (user/insert-new-user! {:email      email
                                          :first_name name
-                                         :type       :api-key})]
+                                         :type       :api-key
+                                         :is_active  false})]
```
noahmoss, line 38: I wonder if this could have unintended side effects? It seems like API users should be considered active as long as the associated API key itself exists and can be used to issue requests. Maybe this doesn't break anything at the moment, but if someone builds a feature in the future and includes an `is-active` check for some reason, it would break for API users and they might not notice.

My bias would be to just adjust the `active-users-count` logic in `metabase.public-settings.premium-features` to explicitly exclude API keys. Wdyt?

I also realized that our anonymous stats reporting in `metabase.analytics.stats` includes more granular user count metrics. I wonder if we should update this to report the count of API keys separately. Probably a question for Arakaki.

src/metabase/api/user.clj
```diff
@@ -64,7 +64,7 @@
            [400 (tru "Not able to modify the internal user")]))
 
 (defn- fetch-user [& query-criteria]
-  (apply t2/select-one (vec (cons User user/admin-or-self-visible-columns)) query-criteria))
+  (apply t2/select-one (vec (cons User user/admin-or-self-visible-columns)) :type "personal" query-criteria))
```
noahmoss, line 67: Generally we use keywords in queries for these sorts of enum-like columns. Toucan2 will automatically convert them to strings because you've already defined a keyword transformation for this field on the User model.
```suggestion
  (apply t2/select-one (vec (cons User user/admin-or-self-visible-columns)) :type :personal query-criteria))
```

noahmoss: Same applies to the instances below

noahmoss (changes requested):

noahmoss (approved):

src/metabase/api/api_key.clj
johnswanson: This makes sense to me, I changed `active-user-count`. For now, I just excluded API-key users from the stats in `metabase.analytics.stats` - if Arakaki wants us to we can add API Keys to the counts down the road, but I figured it's better to keep the status quo for now.

PR #36907, opened 2023-12-18
Fix Replay before:browser:launch ordering

jasonLaster:
### Description

The `before:browser:launch` event defined [here](https://github.com/metabase/metabase/blob/ed7295b5d6cdcf26bb81844c9559b3a1eb155b04/e2e/support/config.js#L48-L64) is not being called because the Replay plugin listens to the same event later on and Cypress only supports one handler per event ([issue](https://github.com/cypress-io/cypress/issues/22428))

This PR moves the Replay plugin setup to the top on and wrap's Cypress's native `on` so that it can handle multiple events. 

We're hopeful that this should fix 8 or so tests ([link](https://www.notion.so/replayio/Metabase-test-review-12-14-e0309735d02a428cbd6e0e845ab15f2d)). 

### How to verify

The success rate should not be affected.

### DIFF
--- a/e2e/support/config.js
+++ b/e2e/support/config.js
@@ -37,12 +37,12 @@ const defaultConfig = {
      ********************************************************************/
 
 
-     if (runWithReplay) {
+    if (runWithReplay) {
+      on = replay.wrapOn(on);
       replay.default(on, config, {
         upload: true,
         apiKey: process.env.REPLAY_API_KEY,
       });
-      on = replay.wrapOn(on);
     }
 
     on(
--- a/e2e/support/config.js
+++ b/e2e/support/config.js
@@ -36,6 +36,15 @@ const defaultConfig = {
      **                        PREPROCESSOR                            **
      ********************************************************************/
 
+
+     if (runWithReplay) {
+      replay.default(on, config, {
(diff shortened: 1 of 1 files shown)


nemanjaglumac (approved):
Approving now because the change looks good, but am [running the full CI](https://github.com/metabase/metabase/actions/runs/7254483333) on this from within the org first.

If all tests come back green, we can merge this.

e2e/support/config.js
```diff
@@ -36,6 +36,15 @@ const defaultConfig = {
      **                        PREPROCESSOR                            **
      ********************************************************************/
 
+
+     if (runWithReplay) {
+      replay.default(on, config, {
+        upload: true,
+        apiKey: process.env.REPLAY_API_KEY,
+      });
+      on = replay.wrapOn(on);
```
ryanjduffy, line 45: This should be before the replay plug-in

nemanjaglumac: Thanks for catching it @ryanjduffy 
Fixed in https://github.com/metabase/metabase/pull/36907/commits/48eed4ae8299216806b8c513e7cc30c2206b325b

PR #36910, opened 2023-12-18
Fix non-showing loading dashcard when switching tabs

### LINKED ISSUE #33767: We're not signaling when a card is loading if you change the filters on a dashboard and switch tabs
### Describe the bug

When you have a loaded dashboard and apply a filter, if you have a card in another tab that is connected to the same filter, the  card in another tab being filtered doesn't signal loading and, depending on the time, it may result in confusing results. 

### To Reproduce

1. Build a beautiful dashboard full of reports and complex questions and 2 tabs
2. Add filters to the dashboard and connect cards to it, in both tabs. I used the Orders table for almost every report to connect to filters, and in the video, I put the same card on both tabs to visually see the bug. 
3. Apply a filter in tab 1. When it finishes loading, change to tab 2. As filters are currently global (not tab-specific), they will apply to the cards in the new tab as well. 
4. See that there is no loading signal

For a more visual impact, in the following video I added the same cards that should be changed to the two tabs. The card in Tab 1 took almost 10 seconds without a loading signal. It's all Sample Database information.

https://github.com/metabase/metabase/assets/132273646/27581390-22c2-42b9-be8f-7af72c5b8b7a

### Expected behavior

The loading signal on cards affected by the filter should be there

### Logs

_No response_

```JSON
v1.47.0
```

### Severity

It can show misleading information.

oleggromov:
Fixes https://github.com/metabase/metabase/issues/33767

## Description

**This is a dirty minimalistic fix.**

Ideally `card` and `dashcard` data and loading states should be held entirely in the store: updated by actions and retrieved by selectors. We currently have `dashboard.loadingDashCards.loadingIds`, but it's not widely used. Our current components instead calculate `isLoading` at various levels ad-hoc, each time using slightly different code.

I am afraid of breaking something - for instance, I already broke Table rendering because the `loading` method in `Visualization` couldn't rely on props when bound to `this`, which may have stale, inappropriate for some lifecycle methods props. At least the bug fix part should be minimal - so I am adding an `OR` to the current conditionals. 

All the rest remains the same except a few tiny readability changes and changes to e2e tests, which were not waiting for dashboards to load and, apparently, checking the stale state (before the re-render when the new data is received).

### Before 
https://github.com/metabase/metabase/assets/2196347/4180acc8-83f2-4074-8228-87f95a50814b 

### After
https://github.com/metabase/metabase/assets/2196347/bb160e6a-999c-4899-8549-118950a78f1a

## Test plan
1. create a dashboard with 2 tabs, multiple cards on each
2. create 1 filter and connect it to all the cards
3. switch between the tabs once to allow initial requests to resolve
4. throttle network in devtools to slow down the requests
5. change the filter value being on one tab
6. without waiting for request to resolve, switch to another tab
7. see there's a loading spinner on the card

### DIFF
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -489,7 +489,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       cy.button("Done").should("be.enabled").click();
       saveDashboard({ waitMs: 250 });
 
-      clickLineChartPoint({ waitMs: 250 });
+      clickLineChartPoint();
       cy.get("@targetDashboardId").then(targetDashboardId => {
         cy.location().should(({ pathname, search }) => {
           expect(pathname).to.equal(`/dashboard/${targetDashboardId}`);
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -947,7 +947,13 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addTimeParameter();
       cy.get("aside").button("Done").click();
 
+      cy.intercept("POST", "/api/dashboard/*/dashcard/*/card/*/query").as(
+        "dashcardQuery",
+      );
+
       saveDashboard();
+      // We need to wait so that the query has enough time to get resolved
+      cy.wait("@dashcardQuery");
 
       clickLineChartPoint();
       cy.findAllByTestId("field-set")
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -215,7 +215,8 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       cy.get("aside").findByText("No available targets").should("exist");
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      // We need to wait so that the query has enough time to get resolved
+      saveDashboard({ waitMs: 500 });
 
       cy.intercept(
         "GET",
@@ -266,7 +267,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addTextParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       clickLineChartPoint();
       cy.findAllByTestId("field-set")
@@ -369,7 +370,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addTextParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       clickLineChartPoint();
       cy.findAllByTestId("field-set")
@@ -443,7 +444,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
         .should("not.exist");
       cy.button("Done").should("be.enabled").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       clickLineChartPoint();
       cy.get("@targetDashboardId").then(targetDashboardId => {
@@ -488,7 +489,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       cy.button("Done").should("be.enabled").click();
       saveDashboard();
 
-      clickLineChartPoint();
+      clickLineChartPoint({ waitMs: 500 });
       cy.get("@targetDashboardId").then(targetDashboardId => {
         cy.location().should(({ pathname, search }) => {
           expect(pathname).to.equal(`/dashboard/${targetDashboardId}`);
@@ -603,7 +604,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addSavedQuestionDestination();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       cy.intercept(
         "GET",
@@ -647,7 +648,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addSavedQuestionCreatedAtParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       clickLineChartPoint();
       cy.findByTestId("qb-filters-panel").should(
@@ -686,7 +687,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addSavedQuestionQuantityParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       clickLineChartPoint();
       cy.wait("@dataset");
@@ -756,7 +757,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       });
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 500 });
 
       onNextAnchorClick(anchor => {
         expect(anchor).to.have.attr("href", URL);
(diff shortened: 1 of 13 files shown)


frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
```diff
@@ -550,6 +542,17 @@ class Visualization extends PureComponent {
   }
 }
 
+function checkIsLoading(series, props) {
+  if (props?.isDashcardDataLoading || _.isEmpty(series)) {
```
oleggromov, line 546: `isEmpty` is the same as previously was but reads easier.

oleggromov:
This is currently blocked by a failing e2e test, which might be related to https://github.com/metabase/metabase/issues/36984

oleggromov:
I merged Anton's https://github.com/metabase/metabase/pull/36987 which should fix the failing e2e test.

frontend/src/metabase/dashboard/components/grid/GridLayout.jsx
```diff
@@ -44,20 +44,9 @@ export function GridLayout({
...
-    [layouts, currentBreakpoint],
-  );
-
-  const cols = useMemo(
-    () => columnsMap[currentBreakpoint],
-    [columnsMap, currentBreakpoint],
-  );
+  const margin = marginMap[currentBreakpoint];
+  const layout = layouts[currentBreakpoint];
+  const cols = columnsMap[currentBreakpoint];
```
oleggromov, line 49: These were really unnecessary as all these are literally just hashmaps with 2 keys each.

frontend/src/metabase/dashboard/components/DashCard/DashCard.tsx
```diff
@@ -160,8 +162,10 @@ function DashCardInner({
   }, [cards, dashcard.id, dashcardData, slowCards]);
 
   const isLoading = useMemo(
-    () => isDashcardLoading(dashcard, dashcardData),
-    [dashcard, dashcardData],
+    () =>
+      isDashcardLoading(dashcard, dashcardData) ||
```
ranquild, line 166: Would it make sense to update `isDashcardLoading` to include `loadingDashcardIds` check? Otherwise the function is somewhat invalid because it doesn't do what it promises without an extra check outside.

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
```diff
@@ -56,6 +56,7 @@ const defaultProps = {
   isEditing: false,
   isSettings: false,
   isQueryBuilder: false,
+  isDashcardDataLoading: false,
```
ranquild, line 59: I'd rather add a more generic `isLoading` prop

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
```diff
@@ -550,6 +542,17 @@ class Visualization extends PureComponent {
   }
 }
 
+function checkIsLoading(series, props) {
```
ranquild, line 545: maybe call it `isLoadingOrEmpty` to avoid overloading `isLoading` term

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -202,7 +202,8 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       cy.get("aside").findByText("No available targets").should("exist");
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      // We need to wait so that the query has enough time to get resolved
+      saveDashboard({ waitMs: 250 });
```
ranquild, line 206: Do you need to wait for some query to run instead? Maybe it's possible to add a boolean flag that makes the helper wait for some queries?

ranquild (commented):
Left some questions

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
oleggromov: Yes, but you don't know which queries exactly. And unfortunately, cypress won't allow you to wait for all queries at once.

So if you do `cy.intercept("GET", "/api/dashboard/*").as("dashboard")` and then `cy.wait('@dashboard')`, it will "resolve" with the first successful query without waiting for the remaining ones. 

This will be address in a separate task, as [mentioned in slack](https://metaboat.slack.com/archives/C5XHN8GLW/p1703278591376449), as it is clearly outside of the scope of this bug fix.

ranquild: Could you explain why you had to increase the timeout?

frontend/src/metabase/dashboard/components/DashCard/DashCard.tsx
oleggromov: Updated.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
oleggromov: I don't have a definitive answer to this but my understanding is the following:

1. `saveDashboard` helper was throwing requests without ever waiting for them
2. the `wait(1)` inside was fixing some rendering issues, basically allowing the browser to calculate/render everything (see the comment inside the helper)
3. tests, however, were always operating on a potentially stale state/data, without waiting for **anything** after `saveDashboard` call. All of them

Number 3 is wrong. I am not sure wether I am missing something or not, but it appears that the tests were working either by accident (1 ms was enough), or stale state was matching the expectations, or actually were flaky.

oleggromov: Actually, this particular in this particular test we fix additional _re-renders_ that are happening because of the `isLoading` updates. The tests that were failing were _hovering_ and trying to click the chart before it finishes rendering.

ranquild: I've tried data-fetching related fixes locally and they didn't work out. I agree with @oleggromov that increasing timeout seems to the only feasible thing here.

ranquild (approved):
`isDashcardDataLoading` looks odd and maybe just `isLoading` should be better, but I see pros and cons in this

oleggromov:
All percy changes are minor, mostly slight changes in rendering (scale of charts, paddings in tables or some other small non-determinisms). @alxnddr please confirm this is ok though:

Screenshot 2023-12-27 at 15 33 42

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
alxnddr: Agree with Alex P

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
alxnddr: I agree

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
alxnddr, line 545: This code has already been far from being ideal, but I think this change complicates it even further. Previously, missing data the sign the card was loading and now we are introducing another flag for that without replacing the old behavior.

The original issue can be fixed by removing stale data for all cards even on other tabs when filters change. We do not need stale data at all, the original issue is about wrongly showing it. At the same time, it would not add an additional flag `isLoading` and instead rely on the existing behavior.

@oleggromov how does this sound to you?

alxnddr (commented):
I'd try to explore a different route here and just remove stale data on filter update

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
oleggromov: This sounds good although I suspect it may break even more tests. I'll try now

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
nemanjaglumac: `visitDashboard` helper waits for all the dashcard requests, so I guess that approach can be used for `saveDashboard` as well.

oleggromov:
This was an incorrect way of fixing the issue. Fixed by https://github.com/metabase/metabase/pull/37134 instead

PR #36919, opened 2023-12-18
Add shallow and location param to /api/collection/tree

### LINKED ISSUE #36442: /collection/tree endpoint takes too long
The GET /collection/tree API request takes about 5 seconds in this screenshot and it returns a lot of data.

This API is called on every page load - optimizing this (both in terms of speed and the amount of data) will help with app performance.

Grafana perf dashboard: https://grafana.monitoring.metabase.com/d/WJb2BRSIk/backend-api?orgId=1&var-method=GET&var-service=metabase&var-target=%2Fapi%2Fcollection&from=now-30d&to=now

Screenshot 2023-12-05 at 4 04 25 PM
Screenshot 2023-12-05 at 4 04 54 PM

qwef:
Closes https://github.com/metabase/metabase/issues/36442

### Description

Adds `shallow` and `location` to `/api/collection/tree` endpoint. With shallow enabled, we no longer return `here`, `below`, and the actual collections in `children`. Instead we populate `children` with a boolean indicating if collections exist within this collection.

### Examples

The following `/api/collection/tree`:

```
[
  {
    "authority_level": null,
    "description": null,
    "archived": false,
    "children": [
      {
        "authority_level": null,
        "description": null,
        "archived": false,
        "children": [
          {
            "authority_level": null,
            "description": null,
            "archived": true,
            "children": [],
            "slug": "aa",
            "name": "aa",
            "personal_owner_id": null,
            "type": null,
            "id": 6,
            "entity_id": "nzYooDpLSz19ODb0Lc7Sx",
            "location": "/4/5/",
            "namespace": null,
            "created_at": "2023-12-18T07:15:10.308482Z"
          }
        ],
        "slug": "a",
        "name": "a",
        "personal_owner_id": null,
        "type": null,
        "id": 5,
        "entity_id": "dMIQectcoWQaf63-k-nmt",
        "location": "/4/",
        "namespace": null,
        "created_at": "2023-12-18T07:15:03.898405Z"
      }
    ],
    "slug": "abc",
    "name": "abc",
    "personal_owner_id": null,
    "type": null,
    "id": 4,
    "entity_id": "tqnHkZU2A2XNToWn88nnk",
    "location": "/",
    "namespace": null,
    "created_at": "2023-12-18T07:14:58.967969Z"
  },
  {
    "authority_level": null,
    "description": "Your instance data. To customize these questions and dashboards, you can duplicate them and save them in the custom reports collection.",
    "archived": false,
    "children": [
      {
        "authority_level": null,
        "description": null,
        "archived": false,
        "children": [],
        "slug": "custom_reports",
        "name": "Custom reports",
        "personal_owner_id": null,
        "type": null,
        "id": 2,
        "entity_id": "okNLSZKdSxaoG58JSQY54",
        "location": "/1/",
        "namespace": "analytics",
        "created_at": "2023-06-08T14:15:39.856869Z"
      }
    ],
    "slug": "metabase_analytics",
    "name": "Metabase analytics",
    "personal_owner_id": null,
    "type": "instance-analytics",
    "here": [
      "card",
      "dataset"
    ],
    "id": 1,
    "entity_id": "vG58R8k-QddHWA7_47umn",
    "location": "/",
    "namespace": "analytics",
    "created_at": "2023-06-08T14:12:32.4457Z"
  }
]
```

If we call `/api/collection/tree?shallow=true&location=/`:

```
[
  {
    "authority_level": null,
    "description": null,
    "archived": false,
    "children": true,
    "slug": "abc",
    "name": "abc",
    "personal_owner_id": null,
    "type": null,
    "id": 4,
    "entity_id": "tqnHkZU2A2XNToWn88nnk",
    "location": "/",
    "namespace": null,
    "created_at": "2023-12-18T07:14:58.967969Z"
  },
  {
    "authority_level": null,
    "description": "Your instance data. To customize these questions and dashboards, you can duplicate them and save them in the custom reports collection.",
    "archived": false,
    "children": true,
    "slug": "metabase_analytics",
    "name": "Metabase analytics",
    "personal_owner_id": null,
    "type": "instance-analytics",
    "id": 1,
    "entity_id": "vG58R8k-QddHWA7_47umn",
    "location": "/",
    "namespace": "analytics",
    "created_at": "2023-06-08T14:12:32.4457Z"
  }
]
```

Then using the collection `id` of the clicked collection, we can now call `/api/collection/tree?shallow=true&location=/4/`:

```
[
  {
    "authority_level": null,
    "description": null,
    "archived": false,
    "children": true,
    "slug": "a",
    "name": "a",
    "personal_owner_id": null,
    "type": null,
    "id": 5,
    "entity_id": "dMIQectcoWQaf63-k-nmt",
    "location": "/4/",
    "namespace": null,
    "created_at": "2023-12-18T07:15:03.898405Z"
  }
]
```

### Where this should be changed

We can update the sidebar and collection permission to use this new API. 

The model picker will still need to use `/api/collection/tree`. The reasoning behind this decision:
- User behavior will probably be more exploratory here (opening more than one collection), so makes sense to load everything before page load. 
- Need to populate `here` and `below`, which causes an extra query to `mdb.query/reducible-query`. From the flamegraph, this is around half the time taken. 

### Misc. Details

We currently pass in `tree=true` as a parameter to `/api/collection/tree` which isn’t needed/used. This can be changed in the FE.

I don't include `exclude-archived`, `exclude-other-user-collections`, or `namespace`. It wouldn't be hard to add these, but didn't want to prematurely add these fields if we don't need them.

### Possible Followup

Add a column in the collections DB with number of children and have the Collections db be sorted on location, allowing for faster lookups by avoiding a full table scan. Unsure if theres any instances with enough collections for this to be worth it though.


noahmoss (changes requested):
In general I really like the approach of only fetching a single collection's children at a time and using a boolean to indicate whether there are more children to fetch. Totally makes sense, it's essentially a user-directed tree traversal.

My feeling, though, is that this should be a separate API rather than being tacked on `/api/collection/tree`. It feels very natural for this to be `/api/collection/:id/children`, or something along those lines, rather than these options being appended as query params to the `/tree` endpoint.

Why?
* `tree` in the API path is misleading if you're not actually fetching a tree
* This endpoint is always scoped to the children of a single collection, so it feels natural to pass the collection ID as part of the path similar to `/api/collection/:id`. `root` would just have to be handled as a separate endpoint, similar to what we have with `/api/collection/root`.
* It reduces the complexity of each individual endpoint and makes them easier to reason about

WDYT? It's possible I'm missing details here so I'm happy to discuss further if there's a good reason to keep the current structure

test/metabase/api/collection_test.clj
```diff
@@ -234,6 +234,45 @@
                                    %)
                                 response)))))))))
 
+(deftest collection-tree-shallow-test
+  (testing "GET /api/collection/tree?shallow=true"
+    (with-collection-hierarchy [a b c d e f g]
+    (let [personal-collection (collection/user->personal-collection (mt/user->id :rasta))
```
noahmoss, line 240: Nit: indentation

test/metabase/api/collection_test.clj
```diff
@@ -234,6 +234,45 @@
...
+                      (filter (fn [coll] (contains? ids (:id coll))))
+                      (map #(select-keys % [:name :children])))))
+          (testing "Make sure each Collection comes back with the expected keys"
+            (is (partial= {:description       nil
+                           :archived          false
+                           :entity_id         (:entity_id personal-collection)
+                           :slug              "rasta_toucan_s_personal_collection"
+                           :name              "Rasta Toucan's Personal Collection"
+                           :personal_owner_id (mt/user->id :rasta)
+                           :id                (:id (collection/user->personal-collection (mt/user->id :rasta)))
```
noahmoss, line 258: nit
```suggestion
                           :id                (u/the-id personal-collection)
```

src/metabase/api/collection.clj
```diff
@@ -123,33 +142,37 @@
...
-                            :id
-                            (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]})
+               (t2/select Collection
+                          {:where [:and
+                                   (when exclude-archived?
+                                     [:= :archived false])
+                                   (perms/audit-namespace-clause :namespace namespace)
+                                   (collection/visible-collection-ids->honeysql-filter-clause
+                                    :id
+                                    (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]})
```
noahmoss, line 161: This is still reading all collections into memory, right? And then subsequently filtering it based on the API keys? That seems like it could still be a bottleneck if an instance has a lot of collections—the FE would be calling this endpoint many times, and each time the backend would be reading _every_ collection. That's a lot more DB load than the status quo, where we only read them all once per page load.

Could we instead change the query to only select the collections it needs to return? Could probably just use a `LIKE` operator on the `location` field to select it by prefix, right?

src/metabase/api/collection.clj
```diff
@@ -102,6 +102,32 @@
...
+  [colls]
+  (->> colls
+       (map collection/personal-collection-with-ui-details)
+       (collection/collections->tree nil)
+       (map (fn [coll] (update coll :children #(boolean (seq %)))))))
+
+(defn- location-from-collection-id-clause
+  [collection-id]
+  (if collection-id
+    [:like :location (tru "%/{0}/%" collection-id)]
```
noahmoss, line 128: I'm afraid this still has a similar issue of reading _all_ the children of the desired collection from the DB, even though only the direct children are returned.

It's a little opaque, but I think you can do what you need with something like 
```suggestion
    [:and
     [:like :location (format "%%/%d/%%" collection-id)]
     [:not [:like :location (format "%%/%d/%%/%%/%%" collection-id)]]]
```

Essentially, you're selecting all collections which have `collection-id` in their parent path, but not ones where it is succeeded by two or more slashes. So you're limiting the results to only children and sub-children of `collection-id`.

There's probably other ways to do it too with subqueries or maybe regexes (although that might require different queries for each app DB) but this was the simplest thing I could think of. I also changed `tru` to `format` in the suggestion code — don't use `tru` for strings that don't actually need to be translated, or else they'll end up in our i18n bundle. A plain `str` could work here too if the doubled-up `%%` are too annoying to read (that's how they're escaped in format strings).

noahmoss: It would also be great if you could add a unit test for the query itself, to ensure it's only reading the necessary collections _prior_ to all the post-processing of the tree. Might need to pull the query out as a helper function, but that seems like a good thing.

src/metabase/api/collection.clj
```diff
@@ -102,6 +102,32 @@
...
+    [:not [:like :location "/%/%/"]]))
```
noahmoss, line 129: This is just for selecting children and sub-children of the root, right? Could you add a comment to make that clearer, or a docstring to the function?

noahmoss (changes requested):
Almost there, but there's still an issue with reading more collections than necessary.

src/metabase/api/collection.clj
```diff
@@ -123,33 +149,35 @@
...
+   namespace                      [:maybe ms/NonBlankString]
+   shallow                        [:maybe :boolean]
+   collection-id                  [:maybe ms/PositiveInt]}
+    (let [colls (cond->>
+                 (t2/select Collection
+                            {:where [:and
+                                     (when exclude-archived
+                                       [:= :archived false])
+                                     (when shallow
+                                       (location-from-collection-id-clause collection-id))
```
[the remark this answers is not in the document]
qwef: Yep, that's right

src/metabase/api/collection.clj
qwef: Yep, will do

src/metabase/api/collection.clj
qwef: Yea that makes sense, thanks for the info on `tru`!

src/metabase/api/collection.clj
```diff
@@ -123,33 +149,35 @@
...
-                            (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]})
-                exclude-other-user-collections?
-                (remove-other-users-personal-collections api/*current-user-id*))
-        colls-with-details (map collection/personal-collection-with-ui-details colls)]
-    (collection/collections->tree coll-type-ids colls-with-details)))
-
+   namespace                      [:maybe ms/NonBlankString]
+   shallow                        [:maybe :boolean]
+   collection-id                  [:maybe ms/PositiveInt]}
+    (let [colls (cond->>
```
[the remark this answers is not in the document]
qwef: Ugh yea I think a calva update is messing some stuff up with this

src/metabase/api/collection.clj
```diff
@@ -102,6 +102,54 @@
...
+  [collection-id]
+  (if collection-id
+    [:and
+     [:like :location (str "%/" collection-id "/%")]
+     [:not [:like :location (str "%/" collection-id "/%/%/%")]]]
+    [:not [:like :location "/%/%/"]]))
+
+(defn- select-collections
+  "Select collections based off certain parameters. If shallow is true, we select only within the
+   tree and its children. This is to pass less data from SQL to metabase."
```
noahmoss, line 138: Just a wording suggestion for clarity
```suggestion
  "Select collections based off certain parameters. If `shallow` is true, we select only the requested collection (or
  the root, if `collection-id` is `nil`) and its immediate children, to avoid reading the entire collection tree when it
  is not necessary."
```

test/metabase/api/collection_test.clj
```diff
@@ -225,15 +225,86 @@
...
-                          (some #(when (= (:id %) (:id (collection/user->personal-collection (mt/user->id :rasta))))
-                                   %)
+                          (some #(when (= (:id %) (u/the-id personal-collection)) %)
                                 response)))))))))
 
+(deftest collection-tree-shallow-test
+  (testing "GET /api/collection/tree?shallow=true"
+    (with-collection-hierarchy [a b c d e f g]
+      (let [personal-collection (collection/user->personal-collection (mt/user->id :rasta))
+            ids      (set (map :id (cons personal-collection [a b c d e f g])))]
```
noahmoss, line 240: nit: alignment

test/metabase/api/collection_test.clj
```diff
@@ -225,15 +225,86 @@
...
+            (is (= [{:name     "B"
+                     :children false}
+                    {:name     "C"
+                     :children true}]
+                   (->> response
+                        (filter (fn [coll] (contains? ids (:id coll))))
+                        (map #(select-keys % [:name :children])))))))))))
+
+(deftest select-collections-shallow-test
+  (testing "Selecing collections based off collection-id equaling nil works."
```
noahmoss, line 275: typo
```suggestion
  (testing "Selecting collections based off collection-id equaling nil works."
```

test/metabase/api/collection_test.clj
```diff
@@ -225,15 +225,86 @@
...
+    (with-collection-hierarchy [a b c d e f g]
+      (let [personal-collection (collection/user->personal-collection (mt/user->id :crowberto))
+            ids      (set (map :id (cons personal-collection [a b c d e f g])))]
+        (mt/with-test-user :crowberto
+          (testing "Make sure we get the expected collections when collection-id is nil"
+            (let [collections (#'api.collection/select-collections nil nil nil true nil)]
```
noahmoss, line 281: I think it would make more sense to pass the first two arguments as `false` since they're expected to be booleans.

Another option is to use `partial` here, since the first three args are constant for all the call sites in this test:
```
(let [select-collections (partial #'api.collection/select-collections false false nil)]
  ... 
  (let [collections (select-collections true nil)]
   ...))
```

That's a stylistic choice though, I don't have a strong opinion either way.

test/metabase/api/collection_test.clj
```diff
@@ -914,7 +984,6 @@
       (is (partial= lucky-personal-subcollection-item
                     (api-get-lucky-personal-collection-with-subcollection :crowberto))))))
 
-
```
noahmoss, line 917: nit: We often do have two newlines _before_ a new section of code. Not sure if this is a real convention or not though tbh

noahmoss (approved):

escherize (approved):

PR #36924, opened 2023-12-18
prevent milestone from closing before release notes are generated

iethree:
### Description

#36816 introduced a race condition which would allow a milestone to be closed before release notes were generated, and this in fact happened.

Screen Shot 2023-12-18 at 4 08 22 PM

17 seconds later, the release notes found no open milestone for v48.0

Screen Shot 2023-12-18 at 4 08 42 PM

This change makes the close-milestone step require both publishing version info and release notes. This change somewhat diminishes the value of skipping those steps (you'll have to manually manage the milestone) but that should be a pretty rare occurrence.

## Testing

Unfortunately, I can't test this in my fork since it has some weird permission problem with pushing tags 😬

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -614,7 +614,7 @@ jobs:
   manage-milestones:
     permissions: write-all
     runs-on: ubuntu-22.04
-    needs: push-tags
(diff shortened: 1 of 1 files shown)


nemanjaglumac (approved):
🚀

PR #36927, opened 2023-12-19
Update FE to use faster /api/collection endpoint

### LINKED ISSUE #36921: Update FE to use new /api/collection/list endpoint
FE part of https://github.com/metabase/metabase/issues/36442

To speed up loading on most pages, we should use the new `/api/collection/list` endpoint, pending here: https://github.com/metabase/metabase/issues/36442

- [ ] Update sidebar to use endpoint
- [ ] Update collection permission page to use endpoint
- [ ] Remove `tree=true` param from `/api/collection/tree` request
- [ ] Check if we need `exclude-archived`, `exclude-other-user-collections`, or `namespace` attributes (omitted in this new endpoint but can be added in if needed)

shaunlebron:
Closes https://github.com/metabase/metabase/issues/36921

### DIFF
--- a/frontend/src/metabase/entities/collections/collections.ts
+++ b/frontend/src/metabase/entities/collections/collections.ts
@@ -33,8 +33,8 @@ const Collections = createEntity({
   displayNameMany: t`collections`,
 
   api: {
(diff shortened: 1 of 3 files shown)


iethree:
saving this for later

PR #36928, opened 2023-12-19
Tiny style tweak in audit-db-installation-test

noahmoss:
Addressing final comment from https://github.com/metabase/metabase/pull/36860 which I didn't put up earlier so I could merge the actual fix in faster, just in case


johnswanson (approved):
Thank you for doing this! I feel bad because it's so miniscule 😂

PR #36938, opened 2023-12-19
remove oss version of the interactive embedding settings page

npretto:
[EPIC](https://github.com/metabase/metabase/issues/36481)
[Product doc](https://www.notion.so/metabase/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart-7e4ed0b3a47947ac9235e53998db1280)
[Figma](https://www.figma.com/file/lwyEC7Mx2jfh2jmUDzv3ZS/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart?type=design&node-id=1030-4083&mode=design&t=k1iOMrA4s7QWuyjC-4)

### Description

This screen doesn't exist anymore with the new design, all the info is on the main embedding settings page


frontend/src/metabase/admin/settings/selectors.js
```diff
@@ -540,35 +539,12 @@ export const ADMIN_SETTINGS_SECTIONS = {
...
-        getHidden: (_, derivedSettings) =>
-          !derivedSettings["enable-embedding"] || PLUGIN_EMBEDDING.isEnabled(),
-      },
       {
         key: "-redirect-widget",
         widget: () => (
           <RedirectWidget to="/admin/settings/embedding-in-other-applications" />
         ),
-        getHidden: (_, derivedSettings) => derivedSettings["enable-embedding"],
+        getHidden: () => PLUGIN_EMBEDDING.isEnabled(),
```
npretto, line 547: The plugin returns if the feature is in the features of the token, in that case we don't have to redirect otherwise EE users won't be able to access the page either

WiNloSt: I think this is incorrect. This change means that the settings page is now accessible even if embedding is off. We should still redirect users back to `/admin/settings/embedding-in-other-applications` if embedding is off (as if the setting page is inaccessible). I tested that reverting to the previous change works. Is there any reason you chose to do this in the first place?

image

WiNloSt (changes requested):

frontend/src/metabase/admin/settings/selectors.js
npretto: That was on EE right?
On OSS that page doesn't exist anymore and I tried to redirect them to the main settings page, but you're correct that I should handle the EE case when the setting is off too.
If I'm correct this should be the new logic
`should redirect = plugin is disabled || setting is off`
-> 
`getHidden = !(plugin is disabled || setting is off)`

I'll add tests (unit should be ok for this) and fix the logic, thanks for pointing that out

WiNloSt: Not sure if that or the version without negation would read better, but it's your call.

`getHidden: embedding plugin is enabled && embedding is on`

enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppSameSiteCookieDescription/EmbeddingAppSameSiteCookieDescription.tsx
```diff
@@ -45,7 +45,7 @@ function AuthorizedOriginsNote() {
       <SameSiteAlert variant="warning" hasBorder>
         <Center>
           <Text>{jt`You should probably change this setting to ${(
-            <Text span fw="bold">
+            <Text key="inner" span fw="bold">
```
npretto, line 48: just a random key to make react shut up
image

WiNloSt (approved):
Tested locally. The redirect logic now works as described.

PR #36941, opened 2023-12-19
Add `PopoverBackButton` to Metabase UI

kulyk:
Extracts a commonly used component to `metabase/ui/buttons/PopoverBackButton` and replaces a couple of it's recent occurrences (in new `DatePicker` and `FilterPicker` components)

### How to verify

1. `yarn storybook`
2. Go to `https://localhost:6006/?path=/story/buttons-popoverbackbutton`
3. Click the button on the canvas, click "Nested item" to see the "Back" button

### Demo

Back button demo


frontend/src/metabase/ui/components/buttons/PopoverBackButton/PopoverBackButton.tsx
kulyk, line 1: * `padding: 0` by default
* `aria-label` is set to "Back" by default
* notice `PopoverBackButtonProps` extends `BoxProps` instead of `ButtonProps` as we don't want to expose props like `variant`, `leftIcon`, `loading`, etc.

ranquild (approved):
Code LGTM!

kdoh (approved):
Looks good to me 👍

PR #36944, opened 2023-12-19
merge lists of embedded questions and dashboards to have new heading structure

npretto:
[EPIC](https://github.com/metabase/metabase/issues/36481)
[Product doc](https://www.notion.so/metabase/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart-7e4ed0b3a47947ac9235e53998db1280)
[Figma](https://www.figma.com/file/lwyEC7Mx2jfh2jmUDzv3ZS/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart?type=design&node-id=1030-4083&mode=design&t=k1iOMrA4s7QWuyjC-4)

### Description
Implements this new heading structure:

image

#### Before 
image

It merges the the two "settings" into one to use "MANAGE EMBEDS" as title of the setting, and two Text components for the smaller headings


e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js
```diff
@@ -90,13 +90,11 @@ describe("scenarios > embedding > smoke tests", { tags: "@OSS" }, () => {
...
 
-        cy.findByTestId("-embedded-dashboards-setting").within(() => {
-          cy.findByText(/Embedded dashboards/i);
-          cy.findByText("No dashboards have been embedded yet.");
-        });
-
-        cy.findByTestId("-embedded-questions-setting")
+        cy.findByTestId("-embedded-resources-setting")
           .within(() => {
+            cy.findByText(/Embedded dashboards/i);
```
WiNloSt, line 95: I know you just moved the assertion inside, but can you as well convert from a regex to string? We prefer string when finding texts because it's easier to understand and makes it less prone to error, because regex will also match partial string, and if we really need to match partial strings that's when we would use regexs.

frontend/src/metabase/admin/settings/components/widgets/PublicLinksListing/PublicLinksListing.jsx
```diff
@@ -174,24 +176,32 @@ export const PublicLinksActionListing = connect(mapStateToProps)(
...
-      noLinksMessage={t`No questions have been embedded yet.`}
-    />
-  </div>
+export const EmbeddedResources = () => (
+  <Stack spacing="md" className="flex-full">
+    <div>
+      <Text mb="sm">{t`Embedded Dashboards`}</Text>
+      <div className="bordered rounded full" style={{ maxWidth: 820 }}>
+        <PublicLinksListing
+          data-testId="-embedded-dashboards-setting"
```
WiNloSt, line 185: We don't seem to use `-embedded-dashboards-setting` or `-embedded-questions-setting` anymore. Can you remove these `data-testId`s that also means we can revert the change at line 68 too?

npretto: They're used _dynamically_ here: https://github.com/metabase/metabase/blob/9be81f36ade0249fa280e73c78b6c51153a45b5b/e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js#L212-L214

I tried to make it use findByText.closest etc but it was becoming dirtier than leaving the original test-ids

WiNloSt: Well this is very hard to find. I'm fine leaving `data-testId` as is, but could you help fix `[embedding-smoketests.cy.spec.js](https://github.com/metabase/metabase/blob/9be81f36ade0249fa280e73c78b6c51153a45b5b/e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js#L212-L214)` so it doesn't use the ID dynamically? We should never create any ID dynamically as it's very hard to find and replace, and makes migrating and refactoring close to impossible.

npretto: > Well this is very hard to find.
💯, I spent quite some time yesterday because of that, I should have pointed that out with a comment on the PR.

I updated the tests to use an inline object to get the full test id so that's cmd+f friendly

e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js
```diff
@@ -204,7 +202,10 @@ describe("scenarios > embedding > smoke tests", { tags: "@OSS" }, () => {
           cy.visit(standalonePath);
           cy.wait("@currentlyEmbeddedObject");
 
-          const sectionTestId = new RegExp(`-embedded-${object}s-setting`);
+          const sectionTestId = {
```
WiNloSt, line 205: 🙏

WiNloSt (approved):
Looks good to me

PR #36950, opened 2023-12-19
CSV appends: permissions checks

calherries:

This PR implements the permissions checks required for appending CSV data to an existing upload table via the endpoint.

It corresponds to the following section of the eng doc ([block link](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df?pvs=4#473fd039820a4ab1bf1b4da51b97cbef)):

### Permissions errors (status: 403):
- Message: “You don’t have permissions to do that.”
- Same permissions checks as uploading a CSV (`POST /api/card/from-csv`) and unrestricted data permissions on the table
- [x]  Test: if user has block permissions: native query editing is required to upload CSV
- [x]  Test: if user doesn’t have unrestricted data permissions then upload fails
- [x]  Test: if the user does have unrestricted data permissions but no native query editing permissions then upload succeeds
- [x]  Test: if user is sandboxed then upload fails

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -606,7 +606,7 @@
         (ex-info (tru "The table must be an uploaded table.")
                  {:status-code 422})
 
-        (not (mi/can-write? table))
+        (not (mi/can-read? table))
         (ex-info (tru "You don''t have permissions to do that.")
                  {:status-code 403}))))
 
--- a/test/metabase/upload_test.clj
+++ b/test/metabase/upload_test.clj
@@ -1139,8 +1139,7 @@
   (try
     (mt/with-temporary-setting-values [uploads-enabled uploads-enabled]
       (mt/with-current-user user-id
-        (let [table-id (or table-id
-                         (:id (create-upload-table!)))]
+        (let [table-id (or table-id (:id (create-upload-table!)))]
           (t2/update! :model/Table table-id {:is_upload is-upload})
           (append-csv! {:table-id table-id
                         :file     file}))))
--- a/test/metabase/upload_test.clj
+++ b/test/metabase/upload_test.clj
@@ -1104,7 +1104,7 @@
   Returns the table.
 
   Defaults to a table with an auto-incrementing integer ID column, and a name column."
-  [& {:keys [table-name col->upload-type rows]
+  [& {:keys [schema-name table-name col->upload-type rows]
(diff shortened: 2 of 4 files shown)


test/metabase/upload_test.clj
```diff
@@ -1104,42 +1104,45 @@
...
+        _ (driver/insert-into! driver db-id schema+table-name insert-col-names rows)
+        table (sync-tables/create-or-reactivate-table! (mt/db) {:name table-name :schema schema-name})]
     (t2/update! :model/Table (:id table) {:is_upload true})
     (binding [upload/*sync-synchronously?* true]
       (#'upload/scan-and-sync-table! (mt/db) table))
     table))
 
 (defn append-csv-with-defaults!
-  "Upload a small CSV file to the given collection ID. Default args can be overridden"
+  "Upload a small CSV file to a newly created default table, or an existing table if `table-id` is provided. Default args can be overridden"
```
calherries, line 1129: This docstring was just wrong before

src/metabase/upload.clj
```diff
@@ -606,7 +606,7 @@
         (ex-info (tru "The table must be an uploaded table.")
                  {:status-code 422})
 
-        (not (mi/can-write? table))
```
calherries, line 609: I had this wrong before. You can read more about table permissions here:
https://github.com/metabase/metabase/blob/1bc18575ed8e453c42024a077d58ac96ca4cf0e5/src/metabase/models/table.clj#L70-L79

qnkhuat: from the doc 
> unrestricted data permissions on the table

could you explain how `can-write` resonates with that?

calherries: It doesn't. That's why I changed it from `can-write?` to `can-read?`.

And from the comments on table permissions I linked above:
[1bc1857/src/metabase/models/table.clj#L70-L79](https://github.com/metabase/metabase/blob/1bc18575ed8e453c42024a077d58ac96ca4cf0e5/src/metabase/models/table.clj#L70-L79)
  ;; To read (e.g., fetch metadata) a Table you must have either self-service data permissions for the Table, or write
  ;; permissions for the Table (detailed below). `can-read?` checks the former, while `can-write?` checks the latter;

qnkhuat (approved):
nice tests!

PR #36951, opened 2023-12-19
Trend Charts — Add static number comparison

### LINKED ISSUE #35552: Support a static number comparison
- Support a static number comparison

kulyk:
Closes #35552

Adds "static number" comparison type for trend charts

### How to verify

2. Summarize Count by Created At
3. Set visualization to "trend"
4. Visualization settings > Data tab > Comparisons > Custom Value
5. Try different values for "Label" and "Value" and ensure the chart calculates and displays the change as expected

### Demo

https://github.com/metabase/metabase/assets/17258145/1c41e00d-05a8-44be-8f4f-4f3a746f8914

### Checklist

### DIFF
--- a/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
@@ -92,6 +92,31 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
       cy.findByText("vs. Jan:");
       cy.findByText("52,249.59");
     });
+
+    // static number
+    cy.findByTestId("chartsettings-sidebar").findByText("3 months ago").click();
+    menu().within(() => {
+      cy.findByText("Custom value…").click();
+
+      // Test the back button
+      cy.findByLabelText("Back").click();
+      cy.findByText("Custom value…").click();
+
+      cy.findByLabelText("Label").type("My Goal");
+      cy.findByLabelText("Value").type("{selectall}42000");
+      cy.button("Done").click();
+    });
+    cy.findByTestId("scalar-previous-value").within(() => {
+      cy.findByText("vs. my goal:").should("exist");
+      cy.findByText("42,000").should("exist"); // goal
+      cy.findByText("26.76%").should("exist"); // down percentage
+    });
+    cy.findByTestId("chartsettings-sidebar").findByText("My Goal").click();
+    menu().within(() => {
+      cy.findByLabelText("Back").should("exist");
+      cy.findByLabelText("Label").should("have.value", "My Goal");
+      cy.findByLabelText("Value").should("have.value", "42000");
+    });
   });
 
   it("should allow display settings to be changed and display should reflect changes", () => {
--- a/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
@@ -183,32 +183,6 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
     cy.findByTestId("scalar-container").findByText("Woah: 68’800.0000% ! cool");
   });
 
-  it("should have data settings disabled if only one option to choose from", () => {
-    // create native question with irregular time periods
-    cy.createNativeQuestion(
-      {
-        name: "13710",
-        native: {
-          query:
-            "SELECT '2026-03-01'::date as date, 22 as \"Value\"\nUNION ALL\nSELECT '2026-04-01'::date, 44\nUNION ALL\nSELECT '2026-06-04'::date, 41",
-        },
-        display: "smartscalar",
-      },
-      { visitQuestion: true },
-    );
-
-    cy.findByTestId("viz-settings-button").click();
-    cy.findByTestId("chartsettings-sidebar").findByText("Data").click();
-
-    cy.findByTestId("chartsettings-sidebar").within(() => {
-      // only one primary number option
-      cy.findByTestId("select-button").should("be.disabled");
-
-      // only one comparison option
-      cy.findByTestId("comparisons-widget-button").should("be.disabled");
-    });
-  });
-
   it("should work regardless of column order (metabase#13710)", () => {
     cy.createQuestion(
       {
--- a/frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
+++ b/frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
@@ -2,10 +2,7 @@ import { useCallback, useState } from "react";
 import _ from "underscore";
 import { Icon } from "metabase/core/components/Icon";
 import { Button, Menu, Stack, Text } from "metabase/ui";
-import type {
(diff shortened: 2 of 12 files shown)


frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalarSettingsWidgets.tsx
```diff
@@ -217,9 +266,76 @@ export function PeriodsAgoInputWidget({
...
+      <Button
+        aria-label={t`Back`}
+        onClick={onBack}
+        variant="subtle"
+        leftIcon={<Icon name="chevronleft" />}
+        c="text.2"
+        fz="1rem"
+        lh="1.25rem"
+        p={0}
+      >{t`Custom value`}</Button>
```
kulyk, line 319: To be replaced with #36941

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
```diff
@@ -16,60 +19,109 @@ type SmartScalarComparisonWidgetProps = {
...
 
+type Tab = "staticNumber" | null;
+
 export function SmartScalarComparisonWidget({
   onChange,
   options,
   value: selectedValue,
 }: SmartScalarComparisonWidgetProps) {
   const [open, setOpen] = useState(false);
+  const [tab, setTab] = useState<Tab>(getTabForValue(selectedValue));
```
JesseSDevaney, line 30: Are we sure that we want to open the menu straight to the `staticNumber` tab? It may be confusing for users to click on the comparison and then just see a back button (which in my mind, would make me assume I am just closing the menu). 

Can we ask product & design what they think is best?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/StaticNumberForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+    onChange({
+      type: COMPARISON_TYPES.STATIC_NUMBER,
+      label,
+      value,
+    });
+  };
+
+  return (
+    <Box component="form" onSubmit={handleSubmit}>
+      <Flex direction="column" align="flex-start" gap="lg">
```
JesseSDevaney, line 51: I think this is a good use case for our Stack Mantine component

```suggestion
      <Stack align="flex-start" spacing="lg">
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/StaticNumberForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <PopoverBackButton
+          onClick={onBack}
+        >{t`Custom value`}</PopoverBackButton>
+        <Flex gap="sm">
```
JesseSDevaney, line 55: I think this a good use case for our Group Mantime component

```suggestion
        <Group spacing="sm">
```

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/StaticNumberForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+  NumberInput,
+  PopoverBackButton,
+  TextInput,
+} from "metabase/ui";
+import type { SelectedComparisonStaticNumber } from "metabase-types/api";
+import { COMPARISON_TYPES } from "../constants";
+import { DoneButton } from "./SmartScalarSettingsWidgets.styled";
+
+interface StaticNumberFormProps {
+  value: Partial<SelectedComparisonStaticNumber>;
```
JesseSDevaney, line 16:
```suggestion
  value?: SelectedComparisonStaticNumber;
```

We cannot say that the `selectedValue` that we are passing to this component is actually the `SelectedComparisonStaticNumber`. If we want to, we can optionally pass the selectedValue if we know it is of type `staticNumber` just like we do in the `PeriodsAgoOptionComponent`.

```
selectedValue={selectedValue.type === type ? selectedValue : undefined}
```

Then we will know for sure we have `undefined | SelectedComparisonStaticNumber`

JesseSDevaney (commented):
I love this! Thanks for working on this, Anton! :heart: 

I left some comments that can be addressed async.

But there are a couple things that might be better discussed synchronously:
1. If we want the `Custom value...` tab to open on initial click (pushing past the other menu options)
2. How we want to handle the merge conflicts from our work on adding in the edited state for submitting values even when clicking outside the menu

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/StaticNumberForm.tsx
kulyk: It seems to be designed for laying things horizontally, at least that's what its [docs page](https://v6.mantine.dev/core/group/?t=props) says:

> Compose elements and components in a horizontal flex container

kulyk:
@JesseSDevaney, thanks for the review! 🙏 Please take another look. I [reworked how menu options are assembled](https://github.com/metabase/metabase/pull/36951/commits/aa0651c3a51132d5dc229ad355b6dbea9b5f89c1), so the static number item isn't skipped when there's no date unit, etc. And also [fixed that `isComparisonValid` wasn't aware of the static number type](https://github.com/metabase/metabase/pull/36951/commits/07aa2d7459adfa42830cf14837ec4509909d3e70). Other changes seem more compact

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
```diff
@@ -158,32 +183,6 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
     cy.findByTestId("scalar-container").findByText("Woah: 68’800.0000% ! cool");
   });
 
-  it("should have data settings disabled if only one option to choose from", () => {
```
kulyk, line 161: With "previous value" and "static number" I don't think we can end up with a single option anymore, can't we?

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/StaticNumberForm.tsx
kulyk: I got confused about the layout here; we actually want to have input on the same row. Replaced `Flex` with `Group` here 👍

JesseSDevaney (approved):

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
```diff
@@ -35,90 +44,120 @@ export function SmartScalarComparisonWidget({
...
+          <StaticNumberForm
+            value={
+              selectedValue.type === COMPARISON_TYPES.STATIC_NUMBER
+                ? selectedValue
+                : undefined
+            }
+            onChange={nextValue => {
+              onChange(nextValue);
+              setOpen(false);
+            }}
```
JesseSDevaney, line 89:
```suggestion
            onChange={nextValue => handleEditedValueChange(nextValue, true)}
```

PR #36952, opened 2023-12-19
Stress test EE specs by default, and add mb_edition input

shaunlebron:
We need a way to stress test EE specs


iethree (approved):
this looks good 🥳 , I'd set it not required and default to `ee` to make the happy path easier.

nemanjaglumac (approved):
What Ryan said.
Please adjust the workflow, but otherwise it looks good. Thanks for the fix.

PR #36958, opened 2023-12-19
Add types to custom cypress commands for auth

shaunlebron:
Continuing https://github.com/metabase/metabase/pull/36813

Motivated by https://github.com/metabase/metabase/pull/36258#discussion_r1431879509

## Summary

Add types for custom commands like `cy.signInAsAdmin()` so we don’t have to use `ts-ignore`.

I only included the authentication commands for now, but it’s a start.

[Relevant Cypress docs here](https://docs.cypress.io/guides/tooling/typescript-support#Types-for-Custom-Commands)

## How to verify

No e2e specs in TS on master yet, but the fix is verified by [this line in another PR](https://github.com/metabase/metabase/pull/36258/files#diff-0edea2253acd2b512aade59ec3edb5019e5f6928cccd6c5d25bd0ceffbb5faf2R83), since the type-check passed and is rebased on this branch.


iethree:
maybe add in the changes that allow you to run .ts specs as well here?

shaunlebron:
> maybe add in the changes that allow you to run .ts specs as well here?

That was already merged to master here: https://github.com/metabase/metabase/pull/36813

e2e/support/index.ts
```diff
@@ -0,0 +1,15 @@
+declare global {
```
WiNloSt, line 1: Could you add these types in the same file we register these functions? This ensure we don't have to have a single file that contains every custom command type, because over time, it'll become unmanageable.

You can see the example here
https://github.com/metabase/metabase/blob/9bca2f2f646320bb775f54bac50935d898c1bfd3/e2e/support/commands/ui/button.ts#L12-L15

shaunlebron: Yes, good catch!

shaunlebron: e0e00b05b15866ba0e29ef0f93043f53959aafa2

e2e/support/commands/user/authentication.ts
```diff
@@ -1,5 +1,18 @@
 import { USERS } from "e2e/support/cypress_data";
 
+declare global {
+  namespace Cypress {
+    interface Chainable {
+      signIn: (user?: keyof typeof USERS) => void;
```
WiNloSt, line 6: For correctness, the return type would be `Cypress.Chainable<Cypress.Response<any>>` because we could chain it. There are some usage that we chain from `cy.signIn*` so technically the result isn't void.

https://github.com/metabase/metabase/blob/780e63246405e6ec3cbdd015a2742f6924311d06/e2e/test/scenarios/onboarding/notifications.cy.spec.js#L63

WiNloSt (approved):
Please see my comment. Also, I wish there were more documents in JS Doc e.g. https://github.com/metabase/metabase/blob/780e63246405e6ec3cbdd015a2742f6924311d06/e2e/support/commands/ui/button.ts#L4-L15


But this is already an improvement, so I'll approve.

PR #36962, opened 2023-12-20
BE analytics events for embedding

noahmoss:
This PR adds tracking of analytics data related to embedding from the backend as specified in [this product doc](https://www.notion.so/Embedding-settings-cleanup-and-surface-interactive-embedding-Quickstart-7e4ed0b3a47947ac9235e53998db1280).

* A couple new keys are added to anonymous usage tracking
* A new Snowplow namespace and two new events are added, sent whenever embedding is enabled or disabled.

Because the Snowplow calls needed to be added to the setter of `enable-embedding` in `metabase.public-settings`, and the Snowplow namespace needs access to a couple of other unrelated settings, there was a circular dependency that was tricky to untangle. The best solution I could find was to create a new `metabase.embed.settings` namespace for the embed settings, but maybe there's a different potential refactor here I'm overlooking.

### DIFF
--- a/src/metabase/embed/settings.clj
+++ b/src/metabase/embed/settings.clj
@@ -1,0 +1,33 @@
+(ns metabase.embed.settings
+  "Settings related to embedding Metabase in other applications."
+  (:require
+   [metabase.analytics.snowplow :as snowplow]
+   [metabase.api.common :as api]
+   [metabase.models.setting :as setting :refer [defsetting]]
+   [metabase.util.i18n
+    :as i18n
+    :refer [deferred-tru]]
+   [toucan2.core :as t2]))
+
+(defsetting embedding-app-origin
+  (deferred-tru "Allow this origin to embed the full {0} application"
+                ((resolve 'metabase.public-settings/application-name-for-setting-descriptions)))
+  :feature    :embedding
+  :visibility :public
+  :audit      :getter)
+
+(defsetting enable-embedding
+  (deferred-tru "Allow admins to securely embed questions and dashboards within other applications?")
+  :type       :boolean
+  :default    false
+  :visibility :authenticated
+  :audit      :getter
+  :setter     (fn [new-value]
+                (when (not= new-value (setting/get-value-of-type :boolean :enable-embedding))
+                  (setting/set-value-of-type! :boolean :enable-embedding new-value)
+                  (let [snowplow-payload {:embedding-app-origin-set   (boolean (embedding-app-origin))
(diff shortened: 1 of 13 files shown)


src/metabase/embed/settings.clj
```diff
@@ -0,0 +1,32 @@
...
+  (deferred-tru "Allow admins to securely embed questions and dashboards within other applications?")
+  :type       :boolean
+  :default    false
+  :visibility :authenticated
+  :audit      :getter
+  :setter     (fn [new-value]
+                (when (not= new-value (setting/get-value-of-type :boolean :enable-embedding))
+                  (setting/set-value-of-type! :boolean :enable-embedding new-value)
+                  (let [snowplow-payload {:embedding-app-origin-set   (boolean (embedding-app-origin))
+                                          :number-embedded-questions  (t2/count :model/Dashboard :enable_embedding true)
```
npretto, line 28: Are the parameters here swapped? :number-embedded-questions seems to count the dashboards not the questions

noahmoss: 🤦 yeah, good catch, thanks. fixed

src/metabase/embed/settings.clj
```diff
@@ -0,0 +1,32 @@
...
+  (deferred-tru "Allow admins to securely embed questions and dashboards within other applications?")
+  :type       :boolean
+  :default    false
+  :visibility :authenticated
+  :audit      :getter
+  :setter     (fn [new-value]
+                (when (not= new-value (setting/get-value-of-type :boolean :enable-embedding))
+                  (setting/set-value-of-type! :boolean :enable-embedding new-value)
+                  (let [snowplow-payload {:embedding-app-origin-set   (boolean (embedding-app-origin))
+                                          :number-embedded-questions  (t2/count :model/Card :enable_embedding true)
```
johnswanson, line 28: Out of curiosity, will this number be meaningful? Are people likely to have any cards or dashboards for which embedding is enabled at the time they enable embedding for everything?

johnswanson (approved):
Looks good to me - had one non-blocking question.

src/metabase/embed/settings.clj
npretto: (I'm guessing) it can help us understand if they are re-enabling it after they've used it. Initially we only had the event for the disable event, we added this one later and we used the same schema

albertoperdomo: I don't have significant assumptions to share at this time. I think we'd like to know how they use the toggle.

Do they flip is mostly once during setup? Or also later after they have embedding running? This is where this data could help.

PR #36964, opened 2023-12-20
Bump liquibase

noahmoss:
Bump to liquibase 4.25.1.

The main issue was the fact that Boolean columns on MySQL/MariaDB were now being created as `tinyint(4)` instead of `bit(1)` as was previously the case (see [this PR](https://github.com/liquibase/liquibase/pull/5130)). `tinyint(4)` is not recognized by the MySQL JDBC driver as a Boolean type, so these fields were being returned as integers (0 or 1) when queried by Metabase instead of Booleans as expected. See [this Slack thread](https://metaboat.slack.com/archives/CKZEMT1MJ/p1703098896963729) for more discussion.

Our workaround is to create a Liquibase property that aliases `${boolean.type}` to `bit(1)` on MySQL, as we've previously done for some other types like `${text.type}` being an alias for `longtext`. This also required migrations to change all existing Boolean columns that would be created as `boolean` on a new installation (i.e. anything in `001_update_migrations` but not in the SQL initialization file) to use `${boolean.type}` instead.

### DIFF
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -4742,28 +4742,6 @@ databaseChangeLog:
         - dbms:
             type: mysql,mariadb
 
-  - changeSet:
-      id: v49.00-032
-      author: noahmoss
-      comment: >-
-        Added 0.49.0 - modify type of DATABASECHANGELOGLOCK.LOCKED
-        from boolean to ${boolean.type} on mysql,mariadb
-      dbms: mysql,mariadb
-      changes:
-        - modifyDataType:
-            tableName: DATABASECHANGELOGLOCK
-            columnName: LOCKED
-            newDataType: ${boolean.type}
-      rollback:
-        - modifyDataType:
-            tableName: DATABASECHANGELOGLOCK
-            columnName: LOCKED
-            newDataType: boolean
-      preConditions:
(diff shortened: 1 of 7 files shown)


metamben (approved):
👍

resources/migrations/001_update_migrations.yaml
```diff
@@ -4381,6 +4390,380 @@ databaseChangeLog:
...
+            newDataType: ${boolean.type}
+      rollback:
+        - modifyDataType:
+            tableName: DATABASECHANGELOGLOCK
+            columnName: LOCKED
+            newDataType: boolean
+      preConditions:
+        - onFail: MARK_RAN
+        - dbms:
+            type: mysql,mariadb
```
metamben, line 4765: Why is this migration needed? Do we care about this type? `docs/installation-and-operation/migrating-from-h2.md` instructs users to use `BOOLEAN`.

metamben (approved):
LGTM. I'm a bit confused about the `DATABASECHANGELOGLOCK` migration.

noahmoss:
Tests were still failing, because apparently the `modifyDataType` Liquibase change type, _on MySQL only_, will remove `NOT NULL` constraints and default values. So I've had to add two more migrations for each column to add these back if necessary.

PR #36970, opened 2023-12-20
Remove inconsistency with readOnly and isEditable

ranquild:
Now `isEditable() === !readOnly()` in all cases. Previously there was an edge case with sandboxing where `isEditable()` was `false` and `readOnly` was `false`, which is incorrect. `isEditable` is more strict.

### DIFF
--- a/e2e/test/scenarios/visualizations-tabular/pivot_tables.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/pivot_tables.cy.spec.js
@@ -26,8 +26,6 @@ const {
   PEOPLE,
   REVIEWS,
   REVIEWS_ID,
-  ANALYTIC_EVENTS,
-  ANALYTIC_EVENTS_ID,
 } = SAMPLE_DATABASE;
 
 const QUESTION_NAME = "Cypress Pivot Table";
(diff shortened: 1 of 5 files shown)


e2e/test/scenarios/visualizations-tabular/pivot_tables.cy.spec.js
```diff
@@ -26,8 +26,6 @@ const {
   PEOPLE,
   REVIEWS,
   REVIEWS_ID,
-  ANALYTIC_EVENTS,
```
ranquild, line 29: New sample DB tables shouldn't be used in e2e tests

frontend/src/metabase-lib/queries/NativeQuery.ts
```diff
@@ -154,15 +154,15 @@ export default class NativeQuery extends AtomicQuery {
   /**
    * Returns true if the database metadata (or lack thererof indicates the user can modify and run this query
    */
-  readOnly() {
-    const database = this.database();
-    return !database || database.native_permissions !== "write";
+  readOnly(): boolean {
```
oleggromov, line 157: I am guessing the comment should be updated too?

ranquild: Fixed

kulyk: fwiw, i don't think we really need that comment anymore, it's clear it's the opposite of `isEditable` from the method itself

kulyk (approved):

PR #36744, opened 2023-12-13
Add "Replace question" action to dashboard cards

kulyk:
Closes #36497, closes #36586

Adds a "Replace" button to dashboard cards with questions or models. It makes it possible to swap the dashboard card's query while keeping the dashboard layout intact. It's important to call out that we would reset properties of a dash card that no longer makes sense with a new query: parameter mappings, click behavior, visualization settings, and added series.

### How to verify

1. Open any dashboard that has a few cards
2. Enter the editing mode
3. Hover a dashboard card using a question or a model
4. Find and click the "Replace" action in the top-right floating panel
5. Select a new question or a model from the modal picker
6. Ensure the dashboard card's content is replaced, and the query is rerun
7. Ensure parameters auto-wiring worked (your dashboard should have another card using the same table and mapped to some parameters)
8. Ensure you can undo the revert
9. Ensure changes are persisted after clicking "Save"

### Demo

<details>
<summary>Basic</summary>

https://github.com/metabase/metabase/assets/17258145/15363bc3-7a37-49f4-90b6-b3af2096c842

</details>

<details>
<summary>Filters auto-wiring</summary>

https://github.com/metabase/metabase/assets/17258145/38d75e0f-2185-470a-926f-bca7ab74c498

</details>

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #36987, opened 2023-12-20
Fix dashcard query sometimes errors after replacing its card

### LINKED ISSUE #36984: Error when replacing a dashboard card
### Describe the bug

When replacing a dashboard card, saving it leads to a failed (500-ed) request to `/api/card/:id/query` as well as error message displayed on the card.

### To Reproduce

1. Create a dashboard with 2 time filters
2. Add 2 cards (and prepare to have one spare) 
3. Edit the dashboard
4. Replace the card with another
5. Save it 

https://github.com/metabase/metabase/assets/2196347/443c312e-ae0a-4058-bc80-50fc7b32fd02

Observe the issue

### Expected behavior

_No response_

### Logs

_No response_

That's on the commit `02fa2c9850a01fb3ca20e8fe33218cb7bc1590ef`

### Severity

P2

### Additional context

_No response_

kulyk:
Fixes #36984

Dashcard queries could fail after saving dashboard changes if their card was replaced while editing (a new feature added in #36744). This happened after we started re-running some queries after clicking "Save" in dashboard editing mode (#36760):

1. We're replacing the dashcard's question and its `card_id` changes
2. We click "Save" and trigger a query rerun
3. 🐞 `fetchCardData` receives [an old dashcard state](https://github.com/metabase/metabase/blob/7ca324a5119408c554e53caf4c5af572f429b023/frontend/src/metabase/dashboard/actions/data-fetching.js#L424) when deciding which query endpoint to use (`/api/card/:id/query` vs. dashcard query endpoint). As a result, it considers a dashcard "dirty" and picks the wrong endpoint, leading to an error

That happened because the dashboard editing state was reset [after](https://github.com/metabase/metabase/blob/7ca324a5119408c554e53caf4c5af572f429b023/frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.jsx#L223) `updateDashboardAndCards` ([here](https://github.com/metabase/metabase/blob/7ca324a5119408c554e53caf4c5af572f429b023/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.jsx#L142)).

The ideal fix would require refactoring top-level dashboard components and moving more logic into redux actions. IMO this is too big for now, so the PR just adds the editing state reset right into the `updateDashboardAndCards` action. We'd be calling `setEditingDashboard(false)` twice because of that, but IMO that's a good enough compromise for now.

To verify: follow the repro steps from #36984


oleggromov (approved):
I can confirm it has fixed the issue https://github.com/metabase/metabase/issues/36984 and I agree with the suggested (and postponed) refactoring approach.

However, since I might be missing something important, you may want to wait for someone else's approvals too. Thanks!

alxnddr (approved):

PR #36992, opened 2023-12-20
Trend charts — Compact formatting viz setting

### LINKED ISSUE #35554: Support value formatting
- Support "compact" formatting setting

kulyk:
Closes #35554

Adds a new boolean viz setting for trend charts that makes it always use compact number format (1M vs 1,000,000 or 400K vs. 400,000)

### To Verify

1. New > Question > Raw Data > Orders
2. Click "Pick the metric you want to see" > Custom Expression > Use `Count * 10000` as a formula
3. Click "Pick a column to group by" > Created At
4. Visualize, change the visualization to "Trend"
5. Open visualization settings sidebar > Display tab
6. Toggle the "Compact number" option
7. Ensure the value changed to something like "3.4M"

### Demo

https://github.com/metabase/metabase/assets/17258145/948e7500-bc25-4a8d-93e6-00c16026f2d1

### DIFF
--- a/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
@@ -201,16 +201,19 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
     cy.findByTestId("chartsettings-sidebar").findByText("Display").click();
 
     cy.findByTestId("scalar-container").findByText("3,440,000");
+    cy.findByTestId("scalar-previous-value").findByText("5,270,000");
 
     cy.findByTestId("chartsettings-sidebar")
       .findByLabelText("Compact number")
       .click();
     cy.findByTestId("scalar-container").findByText("3.4M");
+    cy.findByTestId("scalar-previous-value").findByText("5.3M");
 
     cy.findByTestId("chartsettings-sidebar")
       .findByLabelText("Compact number")
       .click();
     cy.findByTestId("scalar-container").findByText("3,440,000");
+    cy.findByTestId("scalar-previous-value").findByText("5,270,000");
   });
 
   it("should work regardless of column order (metabase#13710)", () => {
--- a/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
@@ -5,6 +5,18 @@ import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
 
 const { ORDERS, ORDERS_ID } = SAMPLE_DATABASE;
 
+const BIG_NUMBER_AGGREGATION = [
+  "aggregation-options",
+  ["*", ["count"], 10000],
+  { name: "Mega Count", "display-name": "Mega Count" },
+];
+
+const AGGREGATIONS = [
+  ["count"],
+  ["sum", ["field", ORDERS.TOTAL, null]],
+  BIG_NUMBER_AGGREGATION,
+];
+
 describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
   beforeEach(() => {
     restore();
@@ -17,7 +29,7 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
         name: "13710",
         query: {
           "source-table": ORDERS_ID,
-          aggregation: [["count"], ["sum", ["field", ORDERS.TOTAL, null]]],
+          aggregation: AGGREGATIONS,
           breakout: [
             ["field", ORDERS.CREATED_AT, { "temporal-unit": "month" }],
           ],
@@ -34,8 +46,7 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
     cy.findByTestId("scalar-container").findByText("344");
     cy.findByTestId("chartsettings-sidebar").findByText("Count").click();
     popover().within(() => {
-      // should only have two options
-      cy.findAllByRole("option").should("have.length", 2);
+      cy.findAllByRole("option").should("have.length", AGGREGATIONS.length);
 
       // selected should be highlighted
       cy.findByLabelText("Count").should("have.attr", "aria-selected", "true");
@@ -48,7 +59,6 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
       );
       cy.findByText("Sum of Total").click();
     });
-    cy.findByTestId("scalar-container").findByText("30,759.47");
 
     // comparisons
     // default should be previous period (since we have a dateUnit)
@@ -125,7 +135,7 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
         name: "13710",
         query: {
           "source-table": ORDERS_ID,
-          aggregation: [["count"], ["sum", ["field", ORDERS.TOTAL, null]]],
+          aggregation: AGGREGATIONS,
           breakout: [
             ["field", ORDERS.CREATED_AT, { "temporal-unit": "month" }],
           ],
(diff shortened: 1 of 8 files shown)


frontend/src/metabase/visualizations/lib/scalar_utils.ts
kulyk, line 1: See [`b068f33`](https://github.com/metabase/metabase/commit/b068f331210989f2847a805a14c65c897f5ea365) for pure refactoring changes and [`1f2b5a8`](https://github.com/metabase/metabase/commit/1f2b5a83199a4f99a6331127041cedb9fc6b6e90) for functional change

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -96,11 +96,10 @@ export function SmartScalar({
...
 
-  const { displayValue, fullScalarValue } = compactifyValue(
-    value,
-    width,
-    formatOptions,
-  );
+  const { displayValue, fullScalarValue } = compactifyValue(value, width, {
+    ...formatOptions,
+    compact: settings["scalar.compact_primary_number"],
+  });
```
kulyk, line 102: Is it fine to set `compact` here? Should that maybe happen somewhere around [here](https://github.com/metabase/metabase/blob/fe2b937807065f056d917f38021f8ff3082b7323/frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js#L23) in `computeTrend`? 🤔

JesseSDevaney:
There was one change that did not get updated with the merge conflicts that causes a bug. We need to update the `staticNumber` forms `onChange` function so that it updates the edited state as well. i.e. update the onchange function to use `handleEditedValueChange` as well.

I left a comment on the merged PR, but we can address it here

image

https://github.com/metabase/metabase/pull/36951#discussion_r1435323589

kulyk:
> There was one change that did not get updated with the merge conflicts that causes a bug. We need to update the `staticNumber` forms `onChange` function so that it updates the edited state as well. i.e. update the onchange function to use `handleEditedValueChange` as well.

Oops, thanks @JesseSDevaney! Fixed in the last commit, please take a look 🙏

frontend/src/metabase/visualizations/lib/scalar_utils.unit.spec.ts
```diff
@@ -88,5 +88,18 @@ describe("scalar utils", () => {
...
+      const width = 350;
+
+      const { displayValue, fullScalarValue } = compactifyValue(value, width, {
+        ...formatOptions,
+        compact: true,
+      }) as { displayValue: string; fullScalarValue: string };
+
+      expect(displayValue).toBe("10.0B");
+      expect(fullScalarValue).toBe("10,010,010,010");
+    });
```
JesseSDevaney, line 103: I think this test case should be moved to `compute.unit.spec.js` since we are moving the compact setting to the `computeTrend` function - since we want that behavior to be replicated for free on static-viz.

Maybe we put this test in a new section: "format options"?

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
```diff
@@ -181,6 +191,26 @@ describe("scenarios > visualizations > trend chart (SmartScalar)", () => {
...
+
+    cy.findByTestId("chartsettings-sidebar")
+      .findByLabelText("Compact number")
+      .click();
+    cy.findByTestId("scalar-container").findByText("3.4M");
+
+    cy.findByTestId("chartsettings-sidebar")
+      .findByLabelText("Compact number")
+      .click();
+    cy.findByTestId("scalar-container").findByText("3,440,000");
```
JesseSDevaney, line 213: With the compact setting being moved to `computeTrend`, can we add a check to validate that the comparison value is compacted as well?

frontend/src/metabase/visualizations/lib/scalar_utils.ts
```diff
@@ -15,21 +22,19 @@ export function compactifyValue(
...
+  const displayValue =
+    formatOptions.compact || checkShouldCompact(fullScalarValue, width)
+      ? formatValue(value, {
+          ...formatOptions,
+          compact: true,
+        })
+      : fullScalarValue;
 
   return { displayValue, fullScalarValue };
 }
```
JesseSDevaney, line 40: See the reply to [this comment](https://github.com/metabase/metabase/pull/36992/files#r1434440644) for the suggested changes here as well. It has to do with moving the compact visualization setting to `computeTrend` and outside of `SmartScalar`

JesseSDevaney (changes requested):
This is looking good! :heart:
Thank you, Anton :pray: 

There a couple things we should address before merging though. Mainly moving the `compact` setting out of `SmartScalar` and into `computeTrend` (and the related changes this requires), so that we get this functionality for static-viz as well.

frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
kulyk: @JesseSDevaney thanks! Your suggestion makes a ton of sense; I think I went the same path, but without changing `compactifyValue.` I think it should be enough as long as `compactifyValue` respects the `formatOptions.compact` setting. I would avoid optimizing it for our use case because it is [also shared with a "number" visualization](https://github.com/metabase/metabase/blob/913c0b04554a9e8b9e3c112e10c6c3c3274e8f33/frontend/src/metabase/visualizations/visualizations/Scalar/Scalar.jsx#L189)

frontend/src/metabase/visualizations/lib/scalar_utils.unit.spec.ts
kulyk: Good call, added tests to `compute.unit.spec.js` in [`c6ec018`](https://github.com/metabase/metabase/pull/36992/commits/c6ec0187d5727e92b0dd8cc929d4498f5c6aeb40). How about we keep that `scalar_utils` test? Let's ensure `compactifyValue` respects the `compact` option

e2e/test/scenarios/visualizations-tabular/smartscalar-trend.cy.spec.js
kulyk: Good call, thanks! Added in [`9238b66`](https://github.com/metabase/metabase/pull/36992/commits/9238b669078b13148db3d07088c8ba4a70670e7b)

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -1952,6 +1952,55 @@ describe("SmartScalar > compute", () => {
...
+      it("should have `compact: true` with `scalar.compact_primary_number` viz setting", () => {
+        const { formatOptions } = computeTrend(
+          series,
+          insights,
+          createVizSettings({
+            "scalar.compact_primary_number": true,
+          }),
+        );
+        expect(formatOptions.compact).toBe(true);
+      });
```
JesseSDevaney, line 2002: Can we add a check for the display values returned by `computeTrend`, since these display values will also be used by the static-viz and we want to insure this behaviors works across any future changes?

**Data Setup**
```javascript
const series = [
        createMockSingleSeries(
          {},
          {
            data: {
              rows: [
                [105000, "2023-10-01T00:00:00"],
                [210000, "2024-10-01T00:00:00"],
              ],
              cols: [
                createMockNumberColumn({ name: "Count" }),
                createMockDateTimeColumn({ name: "Month" }),
              ],
            },
          },
        ),
      ];
```

**Compact Setting is Off**
```javascript
const { comparison, display, formatOptions } = computeTrend(
          series,
          insights,
          createVizSettings(),
        );

const { display: comparisonDisplay } = comparison;

expect(formatOptions.compact).toBeFalsy();
expect(display.value).toBe("210,000")
expect(comparisonDisplay).toBe("105,000")
```

**Compact Setting is On**
```javascript
const { comparison, display, formatOptions } = computeTrend(
          series,
          insights,
          createVizSettings({
            "scalar.compact_primary_number": true,
          }),
        );

const { display: comparisonDisplay } = comparison;

expect(formatOptions.compact).toBe(true);
expect(display.value).toBe("210k")
expect(comparisonDisplay).toBe("105k")
```

JesseSDevaney (approved):
Nice work, Anton! :heart: 

I would just like one small test update before merging. See comment below.

PR #36995, opened 2023-12-20
CSV appends: Use a transaction for inserting data

calherries:

This PR mostly preserves observable behaviour. It replaces the logic I created for handling insertion errors to fake transactions with real transactions. 

Originally I wanted to use `_mb_row_id` for deleting inserted rows if there was a failure, because I thought the reason we weren't using transactions before was to make insertions faster. Turns out I was wrong, using a transaction is actually slightly faster because a transaction doesn't add much overhead but each commit does.

I say "This PR mostly preserves observable behaviour" because it makes one change: if a failure occurs on inserting data and the table didn't have a `_mb_row_id` column before, no `_mb_row_id` column is created. That behaviour can be seen in `append-no-mb-row-id-failure-test`.

### DIFF
--- a/src/metabase/driver/mysql.clj
+++ b/src/metabase/driver/mysql.clj
@@ -758,7 +758,12 @@
           (with-open [^java.io.Writer writer (jio/writer file-path)]
             (doseq [value (interpose \newline tsvs)]
               (.write writer (str value))))
-          (sql-jdbc.execute/do-with-connection-with-options driver db-id nil (fn [conn] (jdbc/execute! conn sql))))
+          (sql-jdbc.execute/do-with-connection-with-options
+           driver
+           db-id
+           nil
+           (fn [conn]
+             (jdbc/execute! {:connection conn} sql))))
         (finally
           (.delete temp-file))))))
 
--- a/src/metabase/driver/mysql.clj
+++ b/src/metabase/driver/mysql.clj
@@ -745,7 +745,7 @@
   ;; `local_infile` must be turned on per
   ;; https://dev.mysql.com/doc/refman/8.0/en/load-data.html#load-data-local
   (if (not= (get-global-variable db-id "local_infile") "ON")
-      ;; If it isn't turned on, fall back to the generic "INSERT INTO ..." way
+    ;; If it isn't turned on, fall back to the generic "INSERT INTO ..." way
     ((get-method driver/insert-into! :sql-jdbc) driver db-id table-name column-names values)
     (jdbc/with-db-transaction [conn (sql-jdbc.conn/db->pooled-connection-spec db-id)]
       (let [temp-file (File/createTempFile table-name ".tsv")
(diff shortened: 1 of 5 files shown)


qnkhuat:
What's your thought on memory usage?

calherries:
Are you referring to the memory footprint on the JVM, or the database? 

Memory usage of the JVM should be independent of whether transactions are used or not, because the objects can be garbage collected after they are inserted.

As for the database, MySQL and Postgres are smart enough to write large files of data to disk, so memory isn't an issue there either. Otherwise large transactions would be impossible to execute.

qnkhuat:
I was thinking about database memory usage.

> MySQL and Postgres are smart enough to write large files of data to disk

Cool, that's good to know, in that case we don't need to worry about blowing the dataware house up.

src/metabase/driver/mysql.clj
```diff
@@ -748,20 +747,21 @@
...
+            file-path (.getAbsolutePath temp-file)]
+        (try
+          (let [tsvs (map (partial row->tsv driver (count column-names)) values)
+                sql  (sql/format {::load   [file-path (keyword table-name)]
+                                  :columns (map keyword column-names)}
+                                 :quoted  true
+                                 :dialect (sql.qp/quote-style driver))]
+            (with-open [^java.io.Writer writer (jio/writer file-path)]
+              (doseq [value (interpose \newline tsvs)]
+                (.write writer (str value))))
```
qnkhuat, line 761: this should be done outside of transaction

src/metabase/driver/sql_jdbc.clj
```diff
@@ -140,25 +140,26 @@
...
+          ;; One imagines that `(long (/ 65535 (count columns)))` might be best, but I don't trust the 65K limit to apply
+          ;; across all drivers. With that in mind, 100 seems like a safe compromise.
+          ;; There's nothing magic about 100, but it felt good in testing. There could well be a better number.
+          chunks     (partition-all (or driver/*insert-chunk-rows* 100) values)
+          sqls       (map #(sql/format {:insert-into table-name
+                                        :columns     columns
+                                        :values      %}
+                                       :quoted true
+                                       :dialect (sql.qp/quote-style driver))
+                          chunks)]
```
qnkhuat, line 160: let's move this preparation outside of the transaction.

qnkhuat (changes requested):
Some preparation steps should be done outside of transaction, other than that this looks good.

src/metabase/upload.clj
```diff
@@ -561,40 +557,20 @@
           _                  (check-schema (dissoc normed-name->field auto-pk-column-name) header)
           col-upload-types   (map (comp base-type->upload-type :base_type normed-name->field) normed-header)
           parsed-rows        (parse-rows col-upload-types rows)]
+      (try
+        (driver/insert-into! driver (:id database) (table-identifier table) normed-header parsed-rows)
+        (catch Throwable e
+
```
qnkhuat, line 563:
```suggestion
```

qnkhuat (approved):
LGTM

PR #36996, opened 2023-12-20
Add internal setting for table render row limit

### LINKED ISSUE #32165: Ability to customize table row limit in subscriptions
When a table has 11 rows and you can't see the full table in an email, and you have to open the csv to see the remaining row, that can be a bit annoying. Requested by a customer as well.

Ability to change the number of rows an email can render from the subscription setting. There may be an impossibility that I am not aware of.
There should be a limit that Metabase has to impose though, like 20 or something, but more than 10 can be very useful.

NA

Requested by a customer that sends a lot of emails, that at the same time has customers that send a lot of emails and encounter with this situation. 

**Additional context**
NA

adam-james-v:
Fixes: #32165

Adds the ability to specify an environment variable that controls the row-limit for rendered tables in alerts and subsriptions.

If you launch Metabase with the environment variable `MB_ATTACHMENT_TABLE_ROW_LIMIT` set to some value, table cards will render with that limit.

eg. 

```
MB_ATTACHMENT_TABLE_ROW_LIMIT=20 java -jar metabase.jar
```

Allows tables to render with that limit:

image


markbastian (approved):
Please check with @cdeweyx that we want to retain the default of 10 and let the customer override to 20 or if we want to default to 20 and make everyone else override to 10. Otherwise, 🚢 🇮🇹.

ojjj13:
Greetings, this feature seems not working in my local instance version 0.48.2, and the default is still 10 rows in my end, anyone else has the same issue?

adam-james-v:
@ojjj13 I'm sorry to hear it's not working for you. I think for any of us to best help it works best if you can [file an issue](https://github.com/metabase/metabase/issues/new?assignees=&labels=.Needs+Triage%2CType%3ABug&projects=&template=bug_report.yml)

In the issue if you give a quick explanation for what you've tried, we'll know what to look for as we try to solve the problem. If you'd like, you can reference this PR in the issue.

I think opening an issue will also allow others who may be experiencing the same problem see that it's already known. Others could add helpful context too, of course.

Thanks for asking, I hope we can get this solved for you nicely :)

abhishek-superk:
@ojjj13 Yes, I'm unable to get this to work as well. 
Tried `MB_ATTACHMENT_TABLE_ROW_LIMIT=20 java -jar metabase.jar` and `java -jar -DMB_ATTACHMENT_TABLE_ROW_LIMIT=20 metabase.jar`

PR #37001, opened 2023-12-20
Api key update, regenerate, and list

### LINKED ISSUE #36998: Write a `PUT /api/api-key/:id` endpoint to update API key names/groups

johnswanson:
While implementing this, I realized that I made a mistake when
implementing the Create endpoint - I returned a `group_id` instead of
the `group_name` that frontend had requested to get back from the List
endpoint. I think for consistency's sake it's best to return
`group_name` from both endpoints.

To achieve this, I added a hydration method to the ApiKey. It adds the
name of a single group to each ApiKey, preferring group names that
aren't "All Users".

Beyond that, the 3 new endpoints are pretty straightforward.

The update and regenerate endpoints are audit logged, and I added audit logging to the Create endpoint as well.

Resolves #36998, #36999, #37000

### DIFF
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -1,7 +1,7 @@
 (ns metabase.api.api-key
   "/api/api-key endpoints for CRUD management of API Keys"
   (:require
-   [compojure.core :refer [POST GET]]
+   [compojure.core :refer [POST GET PUT]]
    [metabase.api.common :as api]
    [metabase.models.api-key :as api-key]
    [metabase.models.permissions-group :as perms-group]
@@ -41,9 +41,9 @@
                                             :name         name
                                             :unhashed_key api-key
                                             :created_by   api/*current-user-id*})
-            (select-keys [:created_at :updated_at :id])
+            (t2/hydrate :group_name)
+            (select-keys [:created_at :updated_at :id :group_name])
             (assoc :name name
-                   :group_id group_id
                    :unmasked_key api-key
                    :masked_key (api-key/mask api-key)))))))
 
@@ -52,5 +52,45 @@
   [:as _body]
   (api/check-superuser)
   (t2/count :model/ApiKey))
+
+(api/defendpoint PUT "/:id"
+  "Update an API key by changing either its group or its name"
+  [id :as {{:keys [group_id name] :as _body} :body}]
+  {id       ms/PositiveInt
+   group_id [:maybe ms/PositiveInt]
+   name     [:maybe ms/NonBlankString]}
+  (api/check-superuser)
+  (let [api-key-before (t2/select-one :model/ApiKey :id id)]
+    (when name
+      (t2/with-transaction [_conn]
+        ;; A bit of a pain to keep these in sync, but oh well.
+        (t2/update! :model/User (:user_id api-key-before) {:first_name name})
+        (t2/update! :model/ApiKey id {:name name})))
+    (when group_id
+      (let [user (-> api-key-before (t2/hydrate :user) :user)]
+        (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])))
+    (-> (t2/select-one :model/ApiKey :id id)
+        (t2/hydrate :group_name)
+        (select-keys [:created_at :updated_at :id :name :masked_key :group_name]))))
+
+(api/defendpoint PUT "/:id/regenerate"
+  "Regenerate an API Key"
+  [id]
+  {id ms/PositiveInt}
+  (api/check-superuser)
+  (let [unhashed-key (key-with-unique-prefix)
+        [id] (t2/update-returning-pks! :model/ApiKey :id id {:unhashed_key unhashed-key})]
+    (-> (t2/select-one :model/ApiKey id)
+        (t2/hydrate :group_name)
+        (select-keys [:created_at :updated_at :id :name :group_name])
+        (assoc :unmasked_key unhashed-key
+               :masked_key (api-key/mask unhashed-key)))))
+
+(api/defendpoint GET "/"
+  "Get a list of API keys. Non-paginated."
+  []
+  (api/check-superuser)
+  (let [api-keys (t2/hydrate (t2/select :model/ApiKey) :group_name)]
+    (map #(select-keys % [:created_at :updated_at :id :name :group_name]) api-keys)))
 
 (api/define-routes)
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -3,6 +3,7 @@
   (:require
    [compojure.core :refer [POST GET PUT]]
    [metabase.api.common :as api]
+   [metabase.events :as events]
    [metabase.models.api-key :as api-key]
    [metabase.models.permissions-group :as perms-group]
    [metabase.models.user :as user]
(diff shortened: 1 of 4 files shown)


src/metabase/api/api_key.clj
```diff
@@ -29,28 +30,83 @@
...
+                     :masked_key (api-key/mask unhashed-key))))))))
 
 (api/defendpoint GET "/count"
   "Get the count of API keys in the DB"
   [:as _body]
   (api/check-superuser)
   (t2/count :model/ApiKey))
 
+(api/defendpoint PUT "/:id"
+  "Update an API key by changing either its group or its name"
```
noahmoss, line 61: Are these mutually exclusive, or could a single call update _both_ the group and its name? I'm assuming its the latter, in which case this docstring could be a little more accurate 
```suggestion
  "Update an API key by changing its group and/or its name"
```

src/metabase/api/api_key.clj
```diff
@@ -29,28 +30,83 @@
...
+  [id :as {{:keys [group_id name] :as _body} :body}]
+  {id       ms/PositiveInt
+   group_id [:maybe ms/PositiveInt]
+   name     [:maybe ms/NonBlankString]}
+  (api/check-superuser)
+  (let [api-key-before (-> (t2/select-one :model/ApiKey :id id)
+                           ;; hydrate the group_name for audit logging
+                           (t2/hydrate :group_name))]
+    (when name
+      (t2/with-transaction [_conn]
```
noahmoss, line 71: If `name` and `group_id` are changed in the same call, then maybe the same transaction should be used for both updates

src/metabase/api/api_key.clj
```diff
@@ -29,28 +30,83 @@
...
+        ;; A bit of a pain to keep these in sync, but oh well.
+        (t2/update! :model/User (:user_id api-key-before) {:first_name name})
+        (t2/update! :model/ApiKey id {:name name})))
```
noahmoss, line 74: I thought we were only going to store `name` on `ApiKey` and not on `User`? Or did that approach change after the product clarifications?

My bias would be to keep things normalized and only store it in one place, and it doesn't matter as much which place it is. But it's not a huge deal if you think this approach works OK.

src/metabase/api/api_key.clj
```diff
@@ -29,28 +30,83 @@
...
+        unhashed-key (key-with-unique-prefix)
+        _ (t2/update! :model/ApiKey :id id {:unhashed_key unhashed-key})
+        api-key (-> (t2/select-one :model/ApiKey id)
+                    (t2/hydrate :group_name))]
+    (events/publish-event! :event/api-key-regenerate
+                           {:object api-key
+                            :previous-object api-key-before
+                            :user-id api/*current-user-id*})
+    (-> (t2/select-one :model/ApiKey id)
+        (t2/hydrate :group_name)
```
noahmoss, line 100: This seems like it's doing more DB calls than it needs to. Why do you need to re-select the ApiKey and hydrate it again after publishing the `api-key-regenerate` event?

noahmoss: You probably don't even need to re-select it at all tbh. Just do this
* Read the API key
* Update the `:unhashed_key` on the object and write it back
* Use the updated object to generate the API response without needing to read it again

It would be nice if there was a `t2/update-returning-instances` but looks like that [hasn't been implemented yet](https://github.com/camsaul/toucan2/blob/c5346586c85f8dcd25573696d3e2f976b6ac0550/src/toucan2/update.clj#L65)

src/metabase/models/api_key.clj
```diff
@@ -12,6 +16,31 @@
...
+    (let [api-key-id->permissions-groups
+          (group-by :api-key-id
+                    (t2/query {:select [[:pg.name :group-name]
+                                        [:pg.id :group-id]
+                                        [:api_key.id :api-key-id]]
+                               :from [[:permissions_group :pg]]
+                               :join [[:permissions_group_membership :pgm]
+                                      [:= :pgm.group_id :pg.id]
+                                      :api_key [:= :api_key.user_id :pgm.user_id]]
+                               :where [:in :api_key.id (map u/the-id (t2/select :model/ApiKey))]}))
```
noahmoss, line 34: shouldn't this query be restricted to the IDs of `api-keys` passed in as an argument, instead of all API keys?

noahmoss (changes requested):
Looks pretty good but I have a couple questions

src/metabase/models/api_key.clj
johnswanson: Argh, yes, thank you - I was sending this to CIDER and forgot to change it back. Nice catch.

src/metabase/api/api_key.clj
johnswanson: I'd prefer to store them in a single location, too. The pain point with that is:
- if we store them only in the ApiKey, we don't have a reliable way to populate the name. If someone does `(t2/select-one [:model/User :first_name :last_name :email] :id 1)`, it will have a `:common_name` that's incorrect, because they didn't also select the `:type` column.
- if we store them only in the User, we don't have a database-level invariant that they're unique. Not the end of the world, but it is nice to have.
To me, the downsides of either option outweigh the benefits of not having to keep things synced, but I'm not super tied to this approach.

src/metabase/api/api_key.clj
johnswanson: Nice, thanks for noticing this.

src/metabase/api/api_key.clj
```diff
@@ -29,28 +30,84 @@
...
+        (t2/update! :model/User (:user_id api-key-before) {:first_name name})
+        (t2/update! :model/ApiKey id {:name name}))
+      (when group_id
+        (let [user (-> api-key-before (t2/hydrate :user) :user)]
+          (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])))
+      (let [updated-api-key (-> (t2/select-one :model/ApiKey :id id)
+                                (t2/hydrate :group_name))]
+        (events/publish-event! :event/api-key-update {:object updated-api-key
+                                                      :previous-object api-key-before
+                                                      :user-id api/*current-user-id*})
```
noahmoss, line 82: Sorry about all the code churn here, but I don't think we want to be publishing an event while still in the transaction, since there could be a number of downstream DB calls. It's _probably_ OK but I think it would be safest to limit the transaction to just the calls that need to happen atomically...if the audit log DB write fails for some reason it doesn't need to block the main operation

src/metabase/api/api_key.clj
```diff
@@ -29,28 +30,84 @@
...
+                             (t2/hydrate :group_name))]
+      (when name
+        ;; A bit of a pain to keep these in sync, but oh well.
+        (t2/update! :model/User (:user_id api-key-before) {:first_name name})
+        (t2/update! :model/ApiKey id {:name name}))
+      (when group_id
+        (let [user (-> api-key-before (t2/hydrate :user) :user)]
+          (user/set-permissions-groups! user [(perms-group/all-users) {:id group_id}])))
+      (let [updated-api-key (-> (t2/select-one :model/ApiKey :id id)
+                                (t2/hydrate :group_name))]
```
noahmoss, line 79: nit: Is it possible to create `updated-api-key` here by just updating the relevant fields on `api-key-before` instead of hitting the DB again? Similar to what you're doing below with the regenerate endpoint.

noahmoss (approved):

src/metabase/api/api_key.clj
johnswanson: I should maybe revisit this because I think it's possibly a sign that the design could use improving, but the problem is the hydration method for `group_name`. We hydrate the group name earlier (to get the previous name for audit logging) and now need to do it again - but calling `t2/hydrate!` a second time won't do anything if we've already hydrated before.

I'd definitely be happy to pull something out of the hydration function to allow us to go from a `group_id` to a `group_name` more easily, and reuse that here. Would you be ok with me doing that in a future PR?

noahmoss: Gotcha, yeah I was just curious if that was possible or not. No need for it to block the PR

PR #37005, opened 2023-12-20
Redshift support for CSV uploads

calherries:
Adds redshift support for CSV uploads / appends


qnkhuat:
Can you rebase? this PR's diff includes the changes from the append Epic

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -18,6 +18,7 @@
    [metabase.public-settings :as public-settings]
    [metabase.query-processor.store :as qp.store]
    [metabase.query-processor.util :as qp.util]
+   [metabase.upload :as-alias upload]
```
qnkhuat, line 21:
```suggestion
   [metabase.upload :as upload]
```
TIL about `:as-alias` but I don't think we want the "create the ns if not found" feature it offers

qnkhuat (approved):
lmk if you have a strong reason to use `:as-alias` over `:as`.

other than that this looks good.

modules/drivers/redshift/src/metabase/driver/redshift.clj
calherries: There's not really a good reason for this, other than that the namespace is only imported for aliasing keywords. It's not necessary so I've removed it.

crisptrutski: Also just learned about this feature. I guess it's aimed namespaces that bring in the [whole jungle](https://www.johndcook.com/blog/2011/07/19/you-wanted-banana/), or have side-effects on being loaded? A bit nervous about how this could mess with REPL based workflows, so agree with the decision against using it.

crisptrutski:
Not familiar with this area, but everything here makes sense and looks good 👍

PR #37007, opened 2023-12-20
Add SegmentedToggle component from mantine

deniskaber:
Needed for https://github.com/metabase/metabase/issues/35961

### Description

Implements the [Mantine SegmentedControl component](https://v6.mantine.dev/core/segmented-control/) in our metabase/ui component library.

### How to verify

1. `yarn storybook`
2. http://localhost:6006/?path=/story/inputs-segmentedcontrol--default-story , http://localhost:6006/?path=/story/inputs-segmentedcontrol--full-width
3. Confirm it looks and behaves like in provided designs - [here](https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=135-19223&mode=dev) and [here](https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?type=design&node-id=324-18627&mode=dev) 

### Demo

Screenshot 2023-12-21 at 01 48 45

### Checklist


WiNloSt:
I assume this is a part of milestone 5? I already put it there, so we track all the work that we've done.

frontend/src/metabase/ui/components/inputs/SegmentedControl/SegmentedControl.styled.tsx
```diff
@@ -0,0 +1,54 @@
...
+      },
+      styles: (
+        theme,
+        { fullWidth }: SegmentedControlStylesParams,
+        { size = "md" },
+      ) => {
+        return {
+          label: {
+            padding: fullWidth
+              ? `${getSize({
```
WiNloSt, line 40: I see no point in using [`getSize`](https://github.com/mantinedev/mantine/blob/v6/src/mantine-styles/src/theme/utils/get-size/get-size.ts), if all sizes have the same dimension (see `OPTION_PADDING`, `OPTION_PADDING_FULL_WIDTH_V`, `OPTION_PADDING_FULL_WIDTH_H`).

Would it make more sense to just inline `rem(8)` or `rem(16)` instead?

frontend/src/metabase/ui/components/inputs/SegmentedControl/SegmentedControl.styled.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const OPTION_PADDING_FULL_WIDTH_V = {
+  xs: rem(8),
+  sm: rem(8),
+  md: rem(8),
+};
+
+const OPTION_PADDING_FULL_WIDTH_H = {
+  xs: rem(16),
```
WiNloSt, line 20: I'm not entirely sure we need a different padding in full width, mostly when used in full width the label should be centered and have more padding than specified here anyway. [I know they're different in the design](https://www.figma.com/file/6cHI5jaztfjDYvEXEhy9uE/Setup-flow-for-static%2Fpublic-embedding-%26-embedding-feature-discoverability?node-id=324%3A18628&mode=dev) (horizontal padding 8px vs 16px), but it might not be that apparent in actual application. Can you confirm with @vbenedetti if that's intended?

If we could use the same padding, it will simplify this override function a lot.

WiNloSt: fair enough.

WiNloSt (approved):
Looks good to me.

mazameli:
Commenting after the merge, but just wanted to say I did manage to get a look at this in Storybook and seemed fine enough to me.

PR #37006, opened 2023-12-20
[Epic #35961] Milestone 4 - Add Legalese Step

oisincoveney:
Milestone 4 of Epic #35961

### Description

Adds the Legalese step for OSS and EE instances with no token

### How to verify

_In an OSS instances and EE instances without a valid token_
* Go to a dashboard/question and access the Embed modal (click the share icon and click the "Embed" option in the menu)
* Click into `Static Embed`
* You should see a screen containing the licensing terms for static embedding. Clicking `Agree and continue` should take you to the static embedding configuration screen.
* Refresh the page and try the same flow. You should immediately see the options for static embedding, as the user has already accepted the terms

_In an EE instance with a valid token_
* Go to a dashboard/question and access the Embed modal (click the share icon and click the "Embed" option in the menu)
* Click into `Static Embed`
* You should immediately see the options for static embedding. For EE instances, there is no legalese step

### Checklist

- [ ] Tests have been added/updated to cover changes in this PR

PR #37008, opened 2023-12-20
[#35961 M4] Add legalese step without any checks

oisincoveney:
Epic #35961 - Milestone 4

### Description

Adds the `Legalese` step to the static embedding modal flow for OSS users. Right now, there are no checks or actions done to indicate whether or not the user has actually accepted the terms - clicking "Agree" simply moves the user to the static embedding settings. This extra behavior (which will most likely require some BE work) will be sorted in #37327 .

### How to verify

* Ensure embedding is enabled. 
* Go to a question/dashboard and click the share dropdown menu, and click `Embed`.
* Click `Static embed`.
* There should be a legalese step. You should see a link to the embedding terms, and clicking `Agree` should take you to the static embedding settings for that question/dashboard

### Checklist

### DIFF
--- a/frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
+++ b/frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
@@ -43,12 +43,13 @@ export const DashboardSharingEmbeddingModal = (
 
   return (
     <EmbedModal isOpen={isOpen} onClose={onClose}>
-      {({ embedType, setEmbedType }) => (
+      {({ embedType, goToNextStep, goToPreviousStep }) => (
         <EmbedModalContent
           {...props}
           isLinkEnabled={isLinkEnabled ?? true}
           embedType={embedType}
-          setEmbedType={setEmbedType}
+          goToNextStep={goToNextStep}
+          goToPreviousStep={goToPreviousStep}
           className={className}
           resource={dashboard}
           resourceParameters={parameters}
--- a/frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
+++ b/frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
@@ -2,7 +2,7 @@ import type { Dashboard } from "metabase-types/api";
 import type { EmbedOptions } from "metabase-types/store";
 import { useDispatch, useSelector } from "metabase/lib/redux";
 import * as Urls from "metabase/lib/urls";
-import { EmbedModal } from "metabase/public/components/widgets/EmbedModal";
+import { EmbedModal } from "metabase/public/components/EmbedModal";
 import EmbedModalContent from "metabase/public/components/widgets/EmbedModalContent";
 import { getParameters } from "metabase/dashboard/selectors";
 
@@ -43,13 +43,13 @@ export const DashboardSharingEmbeddingModal = (
 
   return (
     <EmbedModal isOpen={isOpen} onClose={onClose}>
-      {({ embedType, goToNextStep, goToPreviousStep }) => (
+      {({ embedType, goToNextStep, goBackToEmbedModal }) => (
         <EmbedModalContent
           {...props}
           isLinkEnabled={isLinkEnabled ?? true}
           embedType={embedType}
           goToNextStep={goToNextStep}
-          goToPreviousStep={goToPreviousStep}
+          goBackToEmbedModal={goBackToEmbedModal}
           className={className}
           resource={dashboard}
           resourceParameters={parameters}
--- a/frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
+++ b/frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
@@ -1,4 +1,5 @@
 import { useState } from "react";
+import { t } from "ttag";
 import { useSelector } from "metabase/lib/redux";
 import { getApplicationName } from "metabase/selectors/whitelabel";
 import { Icon } from "metabase/core/components/Icon";
@@ -7,23 +8,18 @@ import type { WindowModalProps } from "metabase/components/Modal/WindowModal";
 import { Box, Text } from "metabase/ui";
 import { EmbedTitleContainer } from "./EmbedModal.styled";
 
-type EmbedModalStep = "application" | "legalese" | null;
+export type EmbedModalStep = "application" | "legalese" | null;
 
 const EmbedTitle = ({
-  embedStep,
   onClick,
+  label,
 }: {
-  embedStep: EmbedModalStep;
+  label: string;
   onClick?: () => void;
 }) => {
-  const applicationName = useSelector(getApplicationName);
-
-  const label =
-    embedStep === null ? `Embed ${applicationName}` : `Static embedding`;
-
   return (
-    <EmbedTitleContainer isActive={embedStep !== null} onClick={onClick}>
-      {embedStep !== null && <Icon name="chevronleft" />}
+    <EmbedTitleContainer onClick={onClick}>
+      {onClick && <Icon name="chevronleft" />}
       <Text fz="xl" c="text.1">
         {label}
       </Text>
@@ -42,14 +38,15 @@ export const EmbedModal = ({
   children: ({
     embedType,
     goToNextStep,
-    goToPreviousStep,
+    goBackToEmbedModal,
   }: {
     embedType: EmbedModalStep;
     goToNextStep: () => void;
-    goToPreviousStep: () => void;
+    goBackToEmbedModal: () => void;
   }) => JSX.Element;
 } & WindowModalProps) => {
   const [embedType, setEmbedType] = useState<EmbedModalStep>(null);
+  const applicationName = useSelector(getApplicationName);
 
   const goToNextStep = () => {
     if (embedType === null) {
@@ -61,7 +58,7 @@ export const EmbedModal = ({
     }
   };
 
-  const goToPreviousStep = () => {
+  const goBackToEmbedModal = () => {
     setEmbedType(null);
   };
 
@@ -70,20 +67,29 @@ export const EmbedModal = ({
     setEmbedType(null);
   };
 
-  const isFullScreen = embedType === "application";
+  const modalHeaderProps =
+    embedType === null
+      ? {
+          label: t`Embed ${applicationName}`,
+          onClick: undefined,
+        }
+      : {
+          label: t`Static embedding`,
+          onClick: goBackToEmbedModal,
+        };
 
   return (
     <Modal
       isOpen={isOpen}
       onClose={onEmbedClose}
-      title={<EmbedTitle embedStep={embedType} onClick={goToPreviousStep} />}
-      fit={!isFullScreen}
-      full={isFullScreen}
+      title={<EmbedTitle {...modalHeaderProps} />}
+      fit={embedType !== "application"}
+      wide={true}
       formModal={false}
       {...modalProps}
     >
       <Box bg="bg.0" h="100%">
-        {children({ embedType, goToNextStep, goToPreviousStep })}
+        {children({ embedType, goToNextStep, goBackToEmbedModal })}
       </Box>
     </Modal>
   );
--- a/frontend/src/metabase/public/components/EmbedModal/EmbedModal.unit.spec.tsx
+++ b/frontend/src/metabase/public/components/EmbedModal/EmbedModal.unit.spec.tsx
@@ -1,0 +1,103 @@
+import userEvent from "@testing-library/user-event";
+import { Route } from "react-router";
+import { mockSettings } from "__support__/settings";
+import { renderWithProviders, screen, within } from "__support__/ui";
+import { createMockState } from "metabase-types/store/mocks";
+import { Button, Group, Text } from "metabase/ui";
+import { EmbedModal } from "./EmbedModal";
+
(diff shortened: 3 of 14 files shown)


oisincoveney:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/37008?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #37006** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37006?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #37008** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37008?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37008?utm_source=stack-comment).

frontend/src/metabase/public/components/widgets/EmbedModal.tsx
```diff
@@ -35,14 +41,29 @@ export const EmbedModal = ({
...
     embedType: EmbedModalStep;
-    setEmbedType: (type: EmbedModalStep) => void;
+    goToNextStep: () => void;
+    goToPreviousStep: () => void;
   }) => JSX.Element;
 } & WindowModalProps) => {
   const [embedType, setEmbedType] = useState<EmbedModalStep>(null);
-  const applicationName = useSelector(getApplicationName);
+
+  const goToNextStep = () => {
```
oisincoveney, line 54: This function will be expanded to check if the user has already accepted the legalese in the future. Right now it's just cycling through the states

frontend/src/metabase/public/components/widgets/EmbedModalContent.jsx
```diff
@@ -111,9 +112,15 @@ class EmbedModalContent extends Component {
       embeddingParams,
     );
 
-    return embedType == null ? (
-      <SharingPane {...this.props} onChangeEmbedType={setEmbedType} />
-    ) : embedType === "application" ? (
+    if (embedType == null) {
```
oisincoveney, line 115: The ternary is getting a little much here, so split it out into `if` statements

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.tsx
```diff
@@ -0,0 +1,43 @@
...
+}: {
+  goToNextStep: () => void;
+}) => (
+  <Center bg="white" px="18rem" pt="6.25rem" pb="11.75rem">
+    <Stack align="center" spacing="3rem">
+      <Title order={2} fz="1.25rem">{t`First, some legalese`}</Title>
+
+      <LegaleseStepDetailsContainer p="lg" w="40rem">
+        <Text fw={700}>
+          {t`By clicking “Agree and continue” you're agreeing to`}{" "}
```
oisincoveney, line 17: Will splitting this sentence mess up the translation? I only ask because splitting the string here, between `…to` and `our embedding license` might introduce some issues when translating to languages that depend on prepositions for cases. I might be misunderstanding how the translation works here but I've been getting uneasy about splitting strings up like this so just want to make sure.

The issue exists in Estonian but I'll put a Russian example since @deniskaber can more easily verify it better than I can. I also know it's google translate which might be very wrong, but I think it still can get the general idea across

image

image

(i.e. `нашей лицензией` vs `наша лицензия`)

frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
```diff
@@ -35,14 +41,29 @@ export const EmbedModal = ({
   onClose: () => void;
   children: ({
     embedType,
-    setEmbedType,
+    goToNextStep,
+    goToPreviousStep,
   }: {
     embedType: EmbedModalStep;
-    setEmbedType: (type: EmbedModalStep) => void;
+    goToNextStep: () => void;
```
oisincoveney, line 48: I'd much rather have these `goTo` functions as explicitly `goToNext` and `goToPrevious` since we're going to need to add some logic to the `goToNext` function, while `goToPrevious` will always return us to the landing. I didn't want to reuse `setEmbedType` since someone else might look at this code and have no idea what `setEmbedType("application")` or `setEmbedType("legalese")` means at first glance

frontend/src/metabase/public/components/EmbedModal/EmbedModal.unit.spec.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { createMockState } from "metabase-types/store/mocks";
+import { Button, Group, Text } from "metabase/ui";
+import { EmbedModal } from "./EmbedModal";
+
+// We need to mock the Modal here because there's an issue with the RoutelessFullPageModal
+// component, which can't find the routing.locationBeforeTransitions prop when it's rendered with this test.
+// We'll just take the props that the Modal is using and render them so we know what the component is
+// working with, and then we can refactor this test later using the Mantine Modal when all inner
+// components are refactored to use Mantine.
+jest.mock("metabase/components/Modal", () => {
```
oisincoveney, line 14: let me know if there's a better way to handle this, I'm not exactly a fan

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.styled.tsx
```diff
@@ -0,0 +1,10 @@
+import { css } from "@emotion/react";
+import styled from "@emotion/styled";
+import { Stack } from "metabase/ui";
+
+export const LegaleseStepDetailsContainer = styled(Stack)`
+  ${({ theme }) => css`
+    border: 1px solid ${theme.colors.border[0]};
+    border-radius: ${theme.radius.md};
+  `}
```
npretto, line 9:
```suggestion
  border: 1px solid ${({ theme }) => theme.colors.border[0]};
  border-radius: ${({ theme }) => theme.radius.md};
```

I haven't found other places where we do `${({ theme }) => css`

(If you accept this, we'll need to remove the css import too)

oisincoveney: Hmm, I'm not sure I like this change. What's the benefit over the original? I like my solution since it keeps the number of arrow functions to a minimum, but if you have a better reason I'd be happy to change it

npretto: It's mostly for consistency with how we usually do it, I checked emotion's docs and I haven't find any example of using `css` inside a styled like that.
The docs mostly use tagged literals as in my suggestions with inline interpolation, or use the object syntax like this:
```tsx
const Container = styled.div(props => ({
  display: 'flex',
  flexDirection: props.column && 'column'
}))
```

Anyway, this is a nit/personal opinion so feel free to ignore it

oisincoveney: Makes sense. I have a bias towards consistency over what makes me feel good so I'll change it 🙂

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.tsx
npretto: Aren't our translations manual/hand-written anyway?
As far as I understand it, the translator will just have 2 keys/"original sentences" to translate but they can be translated as we want.

I looked if using `jt` helps with that, I believe I found the issue where jt originated from: https://github.com/ttag-org/ttag/issues/3#issuecomment-285283686 (funny that the usecase case is also translating something with a link in it :) )
If what they wrote it's correct it seems like even with jt the full context will be missed (as it creates two "keys").

oisincoveney: Ok then let's leave it as it is, and let the translators handle it

frontend/src/metabase/public/components/EmbedModal/EmbedModal.unit.spec.tsx
WiNloSt: I'd not set `full: true` because in [milestone 5 ](https://github.com/metabase/metabase/pull/36181), the modal would not be full anyway. This prevents `Modal` from rendering `RoutelessFullPageModal` which causes the error you mentioned.

This might be worth the compromise, but let's discuss with Product too.

You need to play around with the modal size a bit. In this example I pass `wide: true` and `fit: embedType !== "application"` to `Modal` component in `EmbedModal`

image

Otherwise, if we decide to keep this mock, let's add a comment to ensure we won't forget to revisit this after milestone 5 since we won't render `RoutelessFullPageModal` in embed modal anymore.

frontend/src/metabase/public/components/widgets/EmbedModalContent.jsx
WiNloSt: This looks a lot easier to understand 👍

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.styled.tsx
WiNloSt: Both look fine IMO. But if you have more than a handful of properties that need `theme`, I'd be biased toward @oisincoveney 's solution more because we don't have to repeat `theme` every time.

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.tsx
WiNloSt: The difference would be what would be presented to the translators.

In this case, there would be 2 strings
1. "By clicking “Agree and continue” you're agreeing to"
2. "our embedding license."

if you use `jt` e.g. 
```js
jt`By click... ${(<ExternalLink ...>{t`our embedding license.`}</ExternalLink>)}`
```

the translators would see
1. "By clicking “Agree and continue” you're agreeing to {0}"
2. "our embedding license."

which might give them a bit more context that there is some more texts after the first sentence.

tl;dr I like the use of `jt` in this case.

WiNloSt (approved):
Looks good to me, but I'd wait for @deniskaber 's review before merging since his milestone 5 PR has also modified many files here.

deniskaber (approved):

frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
oisincoveney: Yeah I think that's a good call. I'll change it to `goBackToEmbedModal` or something like that

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.tsx
oisincoveney: Ok I think the `jt` will work just fine here - it's weird that it still gives two keys (I mean, I understand why, but still) but if we can give the translators as much context then let's give it to them

frontend/src/metabase/public/components/widgets/LegaleseStep/LegaleseStep.styled.tsx
oisincoveney: Ok, I'll leave it as is for now then

frontend/src/metabase/public/components/EmbedModal/EmbedModal.unit.spec.tsx
oisincoveney: Yeah, I was thinking about that as well. My thinking was that I should try and reduce the amount of user-facing changes in this milestone since I know Denis is going to move things around, so I wanted to keep things relatively predictable for him

oisincoveney: I like this huge modal style for the time-being so we can reduce the code size, since Milestone 5 is coming very soon 😄

frontend/src/metabase/dashboard/containers/DashboardSharingEmbeddingModal/DashboardSharingEmbeddingModal.tsx
```diff
@@ -2,7 +2,7 @@ import type { Dashboard } from "metabase-types/api";
 import type { EmbedOptions } from "metabase-types/store";
 import { useDispatch, useSelector } from "metabase/lib/redux";
 import * as Urls from "metabase/lib/urls";
-import { EmbedModal } from "metabase/public/components/widgets/EmbedModal";
+import { EmbedModal } from "metabase/public/components/EmbedModal/EmbedModal";
```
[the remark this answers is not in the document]
oisincoveney: fixed 😄

frontend/src/metabase/public/components/EmbedModal/EmbedModal.unit.spec.tsx
oisincoveney: Nice! So I used your suggestion and I could immediately remove all of the mock code and it still looks good 😄

frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
```diff
@@ -1,29 +1,35 @@
...
-  <a className="flex align-center" onClick={onClick}>
-    <EmbedTitleLabel>{t`Sharing`}</EmbedTitleLabel>
-    {type && <Icon name="chevronright" className="mx1 text-medium" />}
-    {type}
-  </a>
-);
+}) => {
+  const applicationName = useSelector(getApplicationName);
+
+  const label =
```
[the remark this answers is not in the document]
oisincoveney: Done!

PR #37011, opened 2023-12-20
Add remove remark functionality to Athena

### LINKED ISSUE #29615: Query metadata within comments leads to cache misses in AWS Athena
```sql
-- Metabase:: userID: 5 queryType: native queryHash: ce6b9b0a85cf2f65a347f1fe10de417d9eeca4bf05f1562a3c9f0437e795e84f
```
I'm experiencing a similar issue as described in issue #9275, but with AWS Athena. It would be greatly appreciated if an option to disable the userID and/or queryHash could be provided

lbrdnk:
Closes #29615

### Description

Approach from #13923 was ported to Athena driver.

### How to verify

Run newly added test.

### Checklist


frontend/src/metabase/databases/constants.tsx
```diff
@@ -64,7 +64,7 @@ export const FIELD_OVERRIDES: Record<string, EngineFieldOverride> = {
   },
   "include-user-id-and-hash": {
     title: t`Include User ID and query hash in queries`,
-    description: t`This can be useful for auditing and debugging, but prevents BigQuery from caching results and may increase your costs.`,
+    description: t`This can be useful for auditing and debugging, but prevents database from caching results and may increase your costs.`,
```
metamben, line 67:
```suggestion
    description: t`This can be useful for auditing and debugging, but prevents  databases from caching results and may increase your costs.`,
```

modules/drivers/athena/test/metabase/driver/athena_test.clj
```diff
@@ -156,3 +159,51 @@
...
+(defn- query->native [query]
+  (let [check-sql-fn (fn [_ _ sql & _] (throw (ex-info "DONE" {:sql sql})))]
+    (with-redefs [sql-jdbc.execute/statement-or-prepared-statement check-sql-fn]
+      (try (qp/process-query query)
+           (catch Throwable t (-> t ex-data :sql))))))
+
+;; TODO: If we are extending remove remark functionality to other drivers, we should consider making this
+;;       test parallel. That would require making `sql-jdbc.execute/statement-or-prepared-statement` dynamic so it
+;;       can be temporarily modified using `binding` instead of `with-redefs` in [[query->native]]. Also it should be
+;;       generalized, taking into account existing solutions for removing remark in other drivers, eg. bigquery.
```
metamben, line 172: Drivers can now control what meta-info and how gets injected, so it can show up at the end of the query instead of at beginning.

metamben (approved):
LGTM

modules/drivers/athena/test/metabase/driver/athena_test.clj
lbrdnk: Thanks for the notice. As the PR got green light, I believe that is not the case for modifying what's been done. When there are issues requiring us to enable this functionality for other drivers, I believe it would be good idea to move code like this to the method you've mentioned, also for bigquery, and consider some more general testing approach.

metamben: I suggest extending the comment with this info. It's OK to do that in another PR if you're worried that it would prevent or significantly delay this one if added here.

lbrdnk: Generalization attempt would imo require more than just moving the logic. I think it would be good idea to (1) make this option available to all appropriate drivers (2) move remark generating call from `execute-reducible-query` function to that method, or maybe even closer to compilation, if that would make sense, (3) testing for all drivers and probably more steps. I'm not yet aware of ramifications of those potential changes, hence I'd like to do this one first and later when we decide to extend, I'd address that. I've updated the comment as you've suggested.

PR #37014, opened 2023-12-20
Add `updated_by` to API Keys

johnswanson:
This required a bit of modification of the DB.

First, I had named the column meant to hold the ID of the user who last updated an API key as `updated_by`. But it seemed like `updated_by` was the natural field to hold the user name that FE wanted, and I didn't want to clobber an existing field with `hydrate`.

So, I renamed the existing column to `updated_by_id` and used `updated_by` to hold the hydrated user that updated the ApiKey.

Second, for consistency, I renamed `created_by` to `creator_id`.

The logic for presenting an ApiKey was getting a bit complicated and spread out over many different endpoints, so I pulled it out into its own function that's called from almost every endpoint defined in the `metabase.api.api-key` namespace. Note that it calls `(t2/hydrate api-key :updated_by :group_name)` - but I verified that this will not result in additional DB calls if the model was already hydrated.

### DIFF
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -4391,6 +4391,52 @@ databaseChangeLog:
                   remarks: 'If true, the table requires a filter to be able to query it'
 
   - changeSet:
+      id: v49.00-013
+      author: johnswanson
+      comment: Add `api_key.updated_by_id`
+      rollback: # will be removed with the table
+      changes:
+        - addColumn:
+            tableName: api_key
+            columns:
+              - column:
+                  remarks: The ID of the user that last updated this API key
+                  name: updated_by_id
+                  type: integer
+                  constraints:
+                    # no default and nullable: false for a new column? yikes! But this should probably be ok, because
+                    # we don't yet have any UI for creating a new API key.
+                    nullable: false
+                    referencedTableName: core_user
+                    referencedColumnNames: id
+                    foreignKeyName: fk_api_key_updated_by_id_user_id
+
+  - changeSet:
+      id: v49.00-014
+      author: johnswanson
+      comment: Add an index on `api_key.updated_by_id`
+      rollback: # not necessary, will be removed with the table
+      changes:
+        - createIndex:
+            tableName: api_key
+            indexName: idx_api_key_updated_by_id
+            columns:
+              column:
+                name: updated_by_id
+
+  - changeSet:
+      id: v49.00-015
+      author: johnswanson
+      comment: Rename `created_by` to `creator_id`
+      rollback: # not necessary, will be removed with the table
+      changes:
+        - renameColumn:
+            tableName: api_key
+            columnDataType: integer
+            oldColumnName: created_by
+            newColumnName: creator_id
+
+  - changeSet:
       id: v49.00-016
       author: noahmoss
       comment: >-
--- a/src/metabase/models/api_key.clj
+++ b/src/metabase/models/api_key.clj
@@ -1,5 +1,6 @@
 (ns metabase.models.api-key
   (:require [crypto.random :as crypto-random]
+            [metabase.api.common :as api]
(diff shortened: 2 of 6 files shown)


johnswanson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/37014?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `api-key-crud`
  * **PR #37014** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37014?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #37051** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37014?utm_source=stack-comment).

resources/migrations/001_update_migrations.yaml
```diff
@@ -4390,6 +4390,52 @@ databaseChangeLog:
...
+            tableName: api_key
+            indexName: idx_api_key_updated_by_id
+            columns:
+              column:
+                name: updated_by_id
+
+  - changeSet:
+      id: v49.00-015
+      author: johnswanson
+      comment: Rename `created_by` to `created_by_id`
```
noahmoss, line 4430: We actually have `creator_id` as the column name in some tables, like `report_dashboard`. Maybe use that instead for consistency? 

You could also call the other one `updater_id` but it sounds a little less natural

src/metabase/models/api_key.clj
```diff
@@ -75,17 +76,26 @@
     unhashed_key (assoc :key (u.password/hash-bcrypt unhashed_key))
     true (dissoc :unhashed_key)))
 
+(defn- add-updated-by-id [api-key]
+  (update api-key :updated_by_id #(or % api/*current-user-id*)))
+
+(defn- add-created-by-id [api-key]
+  (update api-key :created_by_id #(or % api/*current-user-id*)))
```
noahmoss, line 83: Wouldn't this _not_ modify `:updated_by_id` once it's set for the first time? The anonymous fn takes the old value as an arg, so it'll always be used preferentially, right?

noahmoss: Also — I think generally we're trying to move away from using the dynamic vars in `metabase.api.common` in non-API namespaces to make things more explicit and REPL-friendly. (Granted, we have a lot of it at the moment, but there's been some recent discussions about curbing it, because it's annoying to constantly have to bind `*current-user-id*` during development)

I won't block the PR over this, but I think it would be slightly better to keep these in the API namespace like they were before. Or add a functions in the model namespace to create/update API keys which take a user ID as an argument, similar to `metabase.models.card/create-card!`

noahmoss (approved):
Stamping to unblock, but have a couple comments that I'm confident you can address

resources/migrations/001_update_migrations.yaml
johnswanson: I thought about this before and was torn between:
- `created_by_id` is consistent with `updated_by_id` in the same table, but inconsistent with other tables
- `creator_id` is consistent with other tables, but then not consistent with `updated_by_id`
But I think you're right and I'll change this - the inconsistency between `creator_id` and `updated_by_id` is pretty minimal. And if we wanted to hydrate the `creator` of the ApiKey, we could do so already (since it's already set up), while we'd need another line of code to tell t2 how to automagically hydrate the `created_by`.

src/metabase/models/api_key.clj
johnswanson: Good catch - and keeping `api` out of the models does seem more hygienic, I'll do that.

Issue #36185, opened 2023-11-28
Duplicated columns in a source query cannot be distinguished


bshepherdson:
_This is an umbrella issue tracking a deep problem with several diverse symptoms._

The root problem is that duplicated columns in a source-query cannot be properly distinguished in later stages of the query.

## Duplicated columns

I should perhaps say _nearly_ duplicated columns. Here's some ways to create very similar but distinct columns:
- A "self join" of the source table, or joining another table twice.
- The same column both explicitly and implicitly joined.
- The same column implicitly joined twice through different FKs (this might be one case that works?)
- Same column with different binning granularity (eg. month of year vs. month, day vs. month, etc.)
- Some things that depend on the precomputed `field_ref`s in the query result metadata (eg. viz settings) are broken because those `field_ref`s are the same between what should be two distinct columns.

## Backend issues

The query processor conflates these separate columns because their `:id`s are the same. The `data.cols` and `result_metadata` returned with query responses has several confused details due to this bug.

- #25113 is actually pretty much a dupe of this issue.
- #11829 discusses the problem of having no properly unique, permanent name for parts of a query like columns.
    - #17195 is related - we need a better "key" for columns.

## Frontend issues
In some places the FE tries to index columns by ID, which is not sound. The precomputed `field_ref`s in query results are sometimes wrong, and FE code that trusts them (viz settings, some model metadata, maybe filters) is broken.

## Fallout

I believe that all of these bugs stem from this issue:
- #35290 
- #27521 
- #27735 
- #17769
- #33562 (possibly fixed, or only appears if the results render before proper metadata is fetched) 
- #33563 (same as above)
- #28093 (maybe fixed by MLv2 overhaul of column picker; needs repro) 
- #23449
- #12930 
- #19893 (another near dupe of this bug)

And further I suspect that these bugs are at least made more difficult to fix by the same issue:
- #33972
- #19133
- #33565
- #25014 

## Work needed

- I have a draft PR #36085 to fix the most egregious issue with the query result metadata.
    - There's still some issues with the metadata, eg. `field_ref`s being wrong, but those have a nasty blast radius so I'm proceeding carefully.
- @kamilmielnik has a PR which fixes a case where the FE was keying columns by ID and getting confused.
- There are probably some more issues in the details on the FE, and maybe BE as well.
- We probably want to revisit #11829 and really come up with a sound scheme for "keys" on columns that are both unique and resilient even as the query changes.

### Blocking?

[...]

PR #37016, opened 2023-12-20
Renamed columns are displayed correctly in downloads (#18572)

### LINKED ISSUE #18572: Renamed columns in CSV/JSON exports
Renamed columns currently only work for XLSX exports, not for CSV or JSON exports.

There's a bit of a product question here of what to include for each format. There's an old comment [here](https://github.com/metabase/metabase/blob/3617ebd92fba8744ffe66b9612844d7838081c95/src/metabase/query_processor/streaming/json.clj#L28-L29) that posits that JSON output will be likely parsed programmatically, so we should actually keep the original DB column names in the output. Another question is whether we should include currency symbols in column names. These questions are similar to the questions in #17422 about what formatting we should apply.

#18494 should be unskipped once this is fixed for CSVs.

markbastian:
When custom viz settings or metadata is applied to questions and models, the downloaded artifact labels do not always mirror what is displayed on the tabular view in Metabase. This happens for several reasons:

- The `:visualization_settings` were not applied to CSV and JSON downloads
- `update-card-viz-settings` in `metabase.query-processor.middleware.visualization-settings` did not correctly merge keys, resulting in dropped column names in Excel file
- The logic for reconciling column metadata in a result set and the keys in `:visualization_settings` of a query result is not straightforward

This PR is an attempt to move this consistency in the right direction without fixing all the challenges of field matching as discussed in [this thread](https://metaboat.slack.com/archives/CKZEMT1MJ/p1703091539541279) and [this issue](Duplicated columns in a source query cannot be distinguished #36185) as this sounds like a significant effort.

The following changes were made:

1. Update `metabase.query-processor.middleware.visualization-settings/update-card-viz-settings` to merge field settings, if present, into `column-viz-settings` without introducing new keys, especially ambiguous keys as follows:

```clojure
(defn- update-card-viz-settings
  "For each field, fetch its settings from the QP store, convert the settings into the normalized form
  for visualization settings, and then merge in the card-level column settings."
  [column-viz-settings field-ids]
  ;; Retrieve field-level settings
  (let [field-id->settings (reduce
                             (fn [m field-id]
                               (let [field-settings      (:settings (lib.metadata/field (qp.store/metadata-provider) field-id))
                                     norm-field-settings (normalize-field-settings field-id field-settings)]
                                 (assoc m field-id norm-field-settings)))
                             {}
                             field-ids)]
    ;; For each column viz setting, if there is a match on the field settings, merge it in,
    ;; with the column viz settings being the default in the event of conflicts.
    (reduce-kv
      (fn [coll {field-id ::mb.viz/field-id :as k} column-viz-setting]
        (assoc coll k (merge (get field-id->settings field-id {}) column-viz-setting)))
      {}
      column-viz-settings)))
```

The primary difference is rather than merging in new keys as `{::mb.viz/field-id field-id} -> field-settings` this code will merge settings found by field id with any `column-viz-setting` containing that field id as part of its key. The merge prefers `column-viz-settings`.

This should be an improvement over what already exists.

2. Move `column-titles` from `metabase.query-processor.streaming.xlsx` to `metabase.query-processor.streaming.common` and use this common function for CSV and JSON export names.

Note that the lifted `column-titles` has some erroneous logic:

- It relies on column `:name` or `:id` as unique identifiers -- they are not always unique.
- It expects the `col-settings` key format to be exactly `{::mb.viz/field-id id}` or `{::mb.viz/column-name name}` -- Sometimes these keys have additional keys. One change made to this function is normalizing `col-settings` to conform to the expected format by removing extra keys. While this isn't a perfect solution, it conforms with the intent of what already exists and is an improvement in the majority of cases.

The _right_ thing to do is _probably_ to use `metabase.lib.equality/find-matching-column` to match the column and settings keys, _but_ the `col-settings` keys are in a weird format that isn't really conducive to doing this. I'd rather we make an incremental step forward and try doing this in another effort.

This second change should not make anything that wasn't already problematic worse, while making the general case better.

Fixes #18572

### DIFF
--- a/src/metabase/query_processor/middleware/visualization_settings.clj
+++ b/src/metabase/query_processor/middleware/visualization_settings.clj
@@ -18,20 +18,29 @@
   for visualization settings, and then merge in the card-level column settings."
   [column-viz-settings field-ids]
   ;; Retrieve field-level settings
-  (let [field-id->settings (reduce
-                             (fn [m field-id]
-                               (let [field-settings      (:settings (lib.metadata/field (qp.store/metadata-provider) field-id))
-                                     norm-field-settings (normalize-field-settings field-id field-settings)]
-                                 (assoc m field-id norm-field-settings)))
-                             {}
-                             field-ids)]
-    ;; For each column viz setting, if there is a match on the field settings, merge it in,
-    ;; with the column viz settings being the default in the event of conflicts.
-    (reduce-kv
-      (fn [coll {field-id ::mb.viz/field-id :as k} column-viz-setting]
-        (assoc coll k (merge (get field-id->settings field-id {}) column-viz-setting)))
-      {}
-      column-viz-settings)))
+  (let [field-id->settings      (reduce
+                                  (fn [m field-id]
+                                    (let [field-settings      (:settings (lib.metadata/field (qp.store/metadata-provider) field-id))
+                                          norm-field-settings (normalize-field-settings field-id field-settings)]
+                                      (cond-> m
+                                        (seq norm-field-settings)
+                                        (assoc field-id norm-field-settings))))
+                                  {}
+                                  field-ids)
+        ;; For each column viz setting, if there is a match on the field settings, merge it in,
+        ;; with the column viz settings being the default in the event of conflicts.
+        merged-settings         (reduce-kv
+                                  (fn [coll {field-id ::mb.viz/field-id :as k} column-viz-setting]
+                                    (assoc coll k (merge (get field-id->settings field-id {}) column-viz-setting)))
+                                  {}
+                                  column-viz-settings)
+        ;; The field-ids that are in the merged settings
+        viz-field-ids           (set (map ::mb.viz/field-id (keys merged-settings)))
+        ;; Keep any field settings that aren't in the merged settings and have settings
+        distinct-field-settings (update-keys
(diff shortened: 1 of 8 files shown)


bshepherdson (approved):

bshepherdson (approved):

src/metabase/query_processor/middleware/visualization_settings.clj
```diff
@@ -17,12 +17,30 @@
...
+                                    (assoc coll k (merge (get field-id->settings field-id {}) column-viz-setting)))
+                                  {}
+                                  column-viz-settings)
+        ;; The field-ids that are in the merged settings
+        viz-field-ids           (set (map ::mb.viz/field-id (keys merged-settings)))
+        ;; Keep any field settings that aren't in the merged settings and have settings
+        distinct-field-settings (update-keys
+                                  (remove (comp viz-field-ids first) field-id->settings)
+                                  (fn [k] {::mb.viz/field-id k}))]
+    (merge merged-settings distinct-field-settings)))
```
adam-james-v, line 43: Thanks for the comments in here, I think that's helpful! I know I'll forget the specifics in a few weeks anyway, so having them as a guide is nice :)

adam-james-v (approved):

PR #37022, opened 2023-12-20
Fix handling signals in cleanup

oleggromov:
When we catch an exist signal from the OS, we shouldn't try to `process.exit(signal)` because it is incorrect and leads to an error, as well as a possibly incomplete shutdown.

Screenshot 2023-12-20 at 23 07 37

### DIFF
--- a/e2e/runner/run_cypress_tests.js
+++ b/e2e/runner/run_cypress_tests.js
@@ -31,7 +31,11 @@ const cleanup = async (exitCode = 0) => {
     await CypressBackend.stop(server);
   }
 
-  process.exit(exitCode);
+  // We might get a signal code instead, which is a string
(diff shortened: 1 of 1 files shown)


e2e/runner/run_cypress_tests.js
```diff
@@ -31,7 +31,11 @@ const cleanup = async (exitCode = 0) => {
     await CypressBackend.stop(server);
   }
 
-  process.exit(exitCode);
+  // We might get a signal code instead, which is a string
+  // and doesn't require process.exit call
+  if (typeof exitCode === "number") {
+    process.exit(exitCode);
+  }
```
oleggromov, line 38: It comes from `process.on` at the end of this file.

metamben (approved):

PR #37028, opened 2023-12-21
Test some modals more semantically

rafpaf:
This sets the stage for PR https://github.com/metabase/metabase/pull/36866, which includes a semantic test of an AlertModal, and which fixes issue https://github.com/metabase/metabase/issues/36395

### DIFF
--- a/frontend/src/metabase/components/Modal/WindowModal.tsx
+++ b/frontend/src/metabase/components/Modal/WindowModal.tsx
@@ -73,10 +73,7 @@ export class WindowModal extends Component<WindowModalProps> {
         closeOnClickOutside={this.props.closeOnClickOutside}
       >
         <FocusTrap active={this.props.trapFocus}>
-          <div
-            className={cx(className, "relative bg-white rounded")}
-            role="dialog"
-          >
+          <div className={cx(className, "relative bg-white rounded")}>
             {getModalContent({
               ...this.props,
               fullPageModal: false,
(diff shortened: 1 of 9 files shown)


frontend/src/metabase/components/Modal/WindowModal.tsx
```diff
@@ -73,10 +73,7 @@ export class WindowModal extends Component<WindowModalProps> {
         closeOnClickOutside={this.props.closeOnClickOutside}
       >
         <FocusTrap active={this.props.trapFocus}>
-          <div
-            className={cx(className, "relative bg-white rounded")}
-            role="dialog"
-          >
+          <div className={cx(className, "relative bg-white rounded")}>
```
rafpaf, line 76: `role=dialog` is now set just on the ModalContent component, not any of its parents.

frontend/src/metabase/components/ModalContent/ModalContent.tsx
```diff
@@ -82,6 +82,9 @@ export default class ModalContent extends Component<ModalContentProps> {
           // add bottom padding if this is a standard "form modal" with no footer
           { pb4: formModal && !footer },
         )}
+        role="dialog"
+        aria-label={title}
+        aria-modal="true"
```
rafpaf, line 87: Tests that look for a specific modal look for the `ModalContent` component, which has these attributes so it can be found semantically in a unit test.

Why not put these attributes on one of the ancestors of the `ModalContent` component? The reason is that these ancestors don't receive a `title` prop.

rafpaf:
This refactoring PR is `no-backport` because [the PR it supports is `no-backport`](https://github.com/metabase/metabase/pull/36866#issuecomment-1861792194).

rafpaf:
This is still a work in progress. I need to fix the e2e tests.

PR #37043, opened 2023-12-21
Embedding settings cleanup - Milestone 1

npretto:
Milestone 1 of [[Epic]Embedding settings cleanup and surface interactive embedding Quickstart ](https://github.com/metabase/metabase/issues/36481)

Links will be updated in MS2 with utm tags.
MS2 also cointain a fix for the button/link color when in disabled mode and a a change in the label for the embedding setting.


albertoperdomo (approved):
As I mentioned, let's add UTM tracking to the [offer multi-tenant, self-service analytics](https://www.metabase.com/blog/why-full-app-embedding) link in milestone 2.

e2e/test/scenarios/embedding/embedding-smoketests.cy.spec.js
```diff
@@ -15,17 +15,9 @@ import {
 const embeddingPage = "/admin/settings/embedding-in-other-applications";
 const standalonePath =
   "/admin/settings/embedding-in-other-applications/standalone";
-const licenseUrl = "https://metabase.com/license/embedding";
-const upgradeUrl = "https://www.metabase.com/upgrade";
-const learnEmbeddingUrl =
-  "https://www.metabase.com/learn/embedding/embedding-charts-and-dashboards.html";
-
+const pricingUrl = "https://www.metabase.com/pricing";
```
npretto, line 18: This will actually change again to the upgrade url, it's in milestone 2 and we're waiting for the correct utm tags to use

deniskaber (approved):

frontend/src/metabase/admin/settings/selectors.js
```diff
@@ -540,24 +508,16 @@ export const ADMIN_SETTINGS_SECTIONS = {
...
-        getHidden: (_, derivedSettings) => !derivedSettings["enable-embedding"],
-      },
-      {
-        key: "-embedded-questions",
-        display_name: t`Embedded Questions`,
-        widget: EmbeddedQuestionListing,
+        key: "-embedded-resources",
+        display_name: t`Manage embeds`,
+
+        widget: EmbeddedResources,
```
WiNloSt, line 516: We should start thinking about a lighter way to cache API calls. This component calls APIs directly, so there's no cache, and that makes loading a bit jarring.
https://github.com/metabase/metabase/blob/37a620f214eefbdd50c6352d5284f1cc3bc49542/frontend/src/metabase/admin/settings/components/widgets/PublicLinksListing/PublicLinksListing.jsx#L184-L190

If you go back and fourth to the same settings page, it will still show the loading indicators every time.
Screenshot 2023-12-25 at 2 21 25 PM

enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppOriginDescription/EmbeddingAppOriginDescription.tsx
```diff
@@ -1,31 +1,18 @@
 import { jt, t } from "ttag";
 import ExternalLink from "metabase/core/components/ExternalLink";
-import MetabaseSettings from "metabase/lib/settings";
 
 export const EmbeddingAppOriginDescription = () => {
   return (
-    <div>
-      {jt`With this Pro/Enterprise feature you can embed the full Metabase app. Enable your users to drill-through to charts, browse collections, and use the graphical query builder. ${(
+    <div className="my4">
+      <strong className="block text-dark mb1">{t`Authorized origins`}</strong>
```
WiNloSt, line 7: Why do we have to manually code the settings title and not use `display_name` in the setting instead? Using `strong` makes the settings headers look different than other settings in admin settings which seems undesirable.

#### using `display_name` + some spacing adjustments
image

#### This PR
image

frontend/src/metabase/admin/settings/components/widgets/EmbeddingSwitchWidget/EmbeddingSwitchWidget.tsx
```diff
@@ -0,0 +1,24 @@
...
+  onChange,
+}: EmbeddingSwitchWidgetProps) => (
+  <Stack spacing={"md"} className="text-measure">
+    <Text lh={1.5}>
+      {t`Embed dashboards, questions, or the entire Metabase app into your application. Integrate with your server code to create a secure environment, limited to specific users or organizations.`}
+    </Text>
+    <Switch
+      labelPosition="left"
+      checked={Boolean(setting.value)}
+      label={<strong>{t`Embedding Enabled`}</strong>}
```
WiNloSt, line 20: It seems the copy is now updated to `Enabled Embedding`

enterprise/frontend/src/metabase-enterprise/embedding/components/EmbeddingAppOriginDescription/EmbeddingAppOriginDescription.tsx
WiNloSt: Fixed in https://github.com/metabase/metabase/pull/37043/commits/41bee969d22ec1a2bde8d26173f8e9dfe6939628

frontend/src/metabase/admin/settings/components/widgets/EmbeddingSwitchWidget/EmbeddingSwitchWidget.tsx
WiNloSt: fixed in https://github.com/metabase/metabase/pull/37043/commits/16a3c3512a4f27e9bdc73d3abeb2e95eeb60c657

PR #37020, opened 2023-12-20
WIP 36683 notebook-spec flake

oleggromov:
Will fix https://github.com/metabase/metabase/issues/36683

## Sources
- The flaky test reports are coming from [cypress cloud](https://cloud.cypress.io/projects/ywjy9z/analytics/flaky-tests/030a3d78-358a-57de-856b-b60f78359128-db96f242-5abe-940d-38bc-2511e8e7986d?branches=%5B%5D&browsers=%5B%5D&chartRangeMostCommonErrors=%5B%5D&chartRangeSlowestTests=%5B%5D&chartRangeTopFailures=%5B%5D&committers=%5B%5D&cypressVersions=%5B%5D&flaky=%5B%5D&operatingSystems=%5B%5D&runGroups=%5B%5D&specFiles=%5B%7B%22label%22%3A%22e2e%2Ftest%2Fscenarios%2Fquestion%2Fnotebook.cy.spec.js%22%2C%22suggested%22%3Afalse%2C%22value%22%3A%22030a3d78-358a-57de-856b-b60f78359128%22%7D%5D&status=%5B%7B%22label%22%3A%22Passed%22%2C%22value%22%3A%22PASSED%22%7D%2C%7B%22label%22%3A%22Failed%22%2C%22value%22%3A%22FAILED%22%7D%5D&tags=%5B%5D&timeInterval=WEEK&timeRange=%7B%22startDate%22%3A%222023-12-14%22%2C%22endDate%22%3A%222023-12-21%22%7D&viewBy=TEST_CASE) 
- this [E2E Stress Test Flake Fix](https://github.com/metabase/metabase/actions/runs/7290033665/job/19865927994) run
- `skip`ped tests in the suite

## TODO

- [x]   [should process the updated expression when pressing Enter](https://github.com/metabase/metabase/pull/37045)
- [x] [should work on custom column with `case`](https://github.com/metabase/metabase/pull/37049)
- [ ] ~~should allow to pick a saved question when there are models~~ - [no repro](https://github.com/metabase/metabase/actions/runs/7289876708/job/19865438509)
- [ ] should not show "median" aggregation option for databases that do not support "percentile-aggregations" driver feature
- [ ] should show the real number of rows instead of HARD_ROW_LIMIT when loading (metabase#17397)
- [ ] should prompt to join with a model if the question is based on a model
- [ ] should support custom columns
- [x] https://github.com/metabase/metabase/issues/19454

Issue #36683, opened 2023-12-12
[E2E] Fix the flake and remove the @flaky tag on question/notebook.cy.spec.js


perivamsi:

nemanjaglumac:
No more `@flaky` tags in notebook.cy.spec.js.

PR #37049, opened 2023-12-21
Fix flaky e2e notebook / should work on custom column with `case`

oleggromov:
Fixes part of the `notebook.spec.cy` flakiness, part of https://github.com/metabase/metabase/pull/37020 and https://github.com/metabase/metabase/issues/36683

## Problem
The default viewport size for cypress is `1000x660px`. This wasn't enough for a wide table that this test is about to fit into screen _sometimes_, perhaps because of some inherent non-determinism in text rendering or something like that.

1000x660 | 1440x1000
---|---
Screenshot 2023-12-22 at 13 15 52 | Screenshot 2023-12-22 at 13 15 27

## Solution
Increase the viewport size and add a comment so it is clear that it's important. May not keep the test afloat forever, but is good enough.

## Proof that it's passing
- Isolated run: https://github.com/metabase/metabase/actions/runs/7300207727/job/19894415958
- The entire `arithmetic (metabase#13175)` - https://github.com/metabase/metabase/actions/runs/7300516481/job/19895310689

### DIFF
--- a/e2e/test/scenarios/question/notebook.cy.spec.js
+++ b/e2e/test/scenarios/question/notebook.cy.spec.js
@@ -316,11 +316,12 @@ describe("scenarios > question > notebook", { tags: "@slow" }, () => {
       getNotebookStep("expression").contains("Example").should("exist");
 
       visualize(() => {
-        cy.findByTestId("TableInteractive-root").within(() => {
-          cy.contains("Example");
-          cy.contains("Big");
-          cy.contains("Small");
-        });
+        // eslint-disable-next-line no-unscoped-text-selectors
+        cy.contains("Example");
+        // eslint-disable-next-line no-unscoped-text-selectors
+        cy.contains("Big");
+        // eslint-disable-next-line no-unscoped-text-selectors
+        cy.contains("Small");
       });
     });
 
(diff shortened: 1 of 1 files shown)


uladzimirdev (approved):

e2e/test/scenarios/question/notebook.cy.spec.js
```diff
@@ -297,6 +297,9 @@ describe("scenarios > question > notebook", { tags: "@slow" }, () => {
 
   describe("arithmetic (metabase#13175)", () => {
     beforeEach(() => {
+      // This is required because TableInteractive won't render columns
+      // that don't fit into the viewport
+      cy.viewport(1400, 1000);
```
nemanjaglumac, line 302: @oleggromov 

For the future reference, please use Cypress' inline test config for this.
```
  describe(
    "arithmetic (metabase#13175)",
    {
      viewportWidth: 1400,
      viewportHeight: 1000,
    },
    () => {
      beforeEach(() => {
```

That way Cypress already starts the test with that viewport size, so there's no readjusting when the test is already underway. For this particular example it shouldn't make any difference because we're resizing even before we visit any page so effectively it's the same. But there might be some cases where doing this in the middle of the test introduces more trouble.

oleggromov: Thanks, noted. I am not doing it in the middle of a suite though.

nemanjaglumac: I know. That's why I'm saying it doesn't make a difference in this particular example.

Sometimes when we have nested describe blocks, people unknowingly run into this trap. Some of the previous blocks might have a step that's already using UI.

Again, not the case here.

PR #37051, opened 2023-12-21
Add a DELETE endpoint for API Keys

johnswanson:
Don't actually delete them from the database (to keep the invariant that a user with an `:api-key` type will always have an associated ApiKey). Just mark them as `:archived`, and:

- make them invisible to the count endpoint

- make them invisible to the list endpoint

- make them not work for authentication

### DIFF
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -1,7 +1,7 @@
 (ns metabase.api.api-key
   "/api/api-key endpoints for CRUD management of API Keys"
   (:require
-   [compojure.core :refer [POST GET PUT]]
+   [compojure.core :refer [POST GET PUT DELETE]]
    [metabase.api.common :as api]
    [metabase.events :as events]
    [metabase.models.api-key :as api-key]
@@ -85,7 +85,8 @@
   (api/check-superuser)
   (let [api-key-before (-> (t2/select-one :model/ApiKey :id id)
                            ;; hydrate the group_name for audit logging
-                           (t2/hydrate :group_name))]
+                           (t2/hydrate :group_name)
+                           (api/check-404))]
     (t2/with-transaction [_conn]
       (when group_id
         (let [user (-> api-key-before (t2/hydrate :user) :user)]
@@ -107,7 +108,8 @@
   {id ms/PositiveInt}
   (api/check-superuser)
   (let [api-key-before (-> (t2/select-one :model/ApiKey id)
-                           (t2/hydrate :group_name))
+                           (t2/hydrate :group_name)
+                           (api/check-404))
         unhashed-key (key-with-unique-prefix)
         api-key-after (assoc api-key-before
                              :unhashed_key unhashed-key
@@ -128,5 +130,21 @@
   (api/check-superuser)
   (let [api-keys (t2/hydrate (t2/select :model/ApiKey) :group_name :updated_by)]
     (map present-api-key api-keys)))
+
+(api/defendpoint DELETE "/:id"
+  "Delete an ApiKey"
+  [id]
+  {id ms/PositiveInt}
+  (api/check-superuser)
+  (let [api-key (-> (t2/select-one :model/ApiKey id)
+                    (t2/hydrate :group_name)
+                    (api/check-404))]
+    (t2/with-transaction [_tx]
+      (t2/delete! :model/ApiKey id)
+      (t2/update! :model/User (:user_id api-key) {:is_active false}))
+    (events/publish-event! :event/api-key-delete
+                           {:object api-key
+                            :user-id api/*current-user-id*})
+    api/generic-204-no-content))
 
 (api/define-routes)
--- a/src/metabase/events/audit_log.clj
+++ b/src/metabase/events/audit_log.clj
@@ -204,6 +204,7 @@
 (derive :event/api-key-create ::api-key-event)
 (derive :event/api-key-update ::api-key-event)
 (derive :event/api-key-regenerate ::api-key-event)
+(derive :event/api-key-delete ::api-key-event)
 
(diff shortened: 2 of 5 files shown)


johnswanson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/37051?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `api-key-crud`
  * **PR #37014** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37014?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #37051** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37051?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37051?utm_source=stack-comment).

noahmoss:
What's the advantage of archiving API keys instead of deleting the API Key _and_ the associated user?

Normal users can only be archived, not deleted, but api-key users don't necessarily have to have the same constraint since they're more of an internal mechanism than actual users.

And there's the slight risk that a bug in the future could allow deactivated API keys to still be used for auth. I know you've added checks to prevent that but it's still a potential risk if we're not allowing API keys to be fully deleted.

Don't have a strong opinion necessarily, but just want to make sure we have a good justification for the approach here.

johnswanson:
> What's the advantage of archiving API keys instead of deleting the API Key _and_ the associated user?
> 
> Normal users can only be archived, not deleted, but api-key users don't necessarily have to have the same constraint since they're more of an internal mechanism than actual users.

Deleting a user would cause other issues - for example, if an API key user creates a dashboard, then deleting that user would break the `creator_id` of that dashboard (I think there's a foreign key constraint to prevent that).

> And there's the slight risk that a bug in the future could allow deactivated API keys to still be used for auth. I know you've added checks to prevent that but it's still a potential risk if we're not allowing API keys to be fully deleted.

This is a good point to consider. What would you think of regenerating the API key as part of "deletion"? That way any existing key becomes unusable even if there's a bug that allows auth with a "deleted" key.

noahmoss:
Gotcha. Yeah I'd be in favor of either:
* Overwriting the hashed key in the API Key with a dummy value or just set it to `nil` (maybe `nil` would be clearer)
* Don't necessarily require that API Key users have an associated API Key, and delete the API Key but leave the user as archived

johnswanson:
> Gotcha. Yeah I'd be in favor of either:
> 
> * Overwriting the hashed key in the API Key with a dummy value or just set it to `nil` (maybe `nil` would be clearer)
> * Don't necessarily require that API Key users have an associated API Key, and delete the API Key but leave the user as archived

One additional point I want to consider is auditability. Say

- someone uses an API key to do something malicious
- later, for unrelated reasons, that ApiKey is deleted
- later, the malicious activity is noticed

We'd have an audit log that showed that User 123, with type `:api_key`, performed some malicious activity. But we wouldn't have any way of figuring out which ApiKey did it.

I think this makes me lean towards the first option - overwriting the ApiKey with a `nil` hashed key. That way, if you have a user ID that performed a malicious action, you can always get the associated ApiKey.

Does that sound reasonable? Another option would be to add `user_id` to the audited fields for ApiKeys, so the deletion audit log event contains the user ID. But this feels a bit less obvious.

src/metabase/api/api_key.clj
```diff
@@ -129,4 +129,19 @@
...
     (map present-api-key api-keys)))
 
+(api/defendpoint DELETE "/:id"
+  "Delete an ApiKey"
+  [id]
+  {id ms/PositiveInt}
+  (api/check-superuser)
+  (t2/with-transaction [_tx]
+    (let [api-key (-> (t2/select-one :model/ApiKey id)
+                      (t2/hydrate :group_name))]
```
noahmoss, line 139: Not a huge deal, but might be worth wrapping this in `check-404`. Even though it's a deletion endpoint, we should probably return a 404 if the caller is trying to delete something that already doesn't exist. 

The User update call below would also error in this case anyway, so probably better to catch that at the start and error.

src/metabase/api/api_key.clj
```diff
@@ -129,4 +129,19 @@
...
+      (t2/delete! :model/ApiKey id)
+      (t2/update! :model/User (:user_id api-key) {:is_active false})
+      (events/publish-event! :event/api-key-delete
+                             {:object api-key
+                              :user-id api/*current-user-id*})
```
noahmoss, line 144: Here's another case where I don't think the `publish-event!` call should be in the transaction, since we don't know what downstream DB calls are generated

src/metabase/api/api_key.clj
```diff
@@ -129,4 +129,19 @@
...
+      (present-api-key api-key))))
```
noahmoss, line 145: Is there a specific reason we need to return the API key after it's been deleted? If not, you could just return a 204 (`api/generic-204-no-content`)

noahmoss (approved):

src/metabase/api/api_key.clj
johnswanson: Good call. I'm also going to add this check to the `PUT` endpoints.

src/metabase/api/api_key.clj
johnswanson: Oops, thanks for noticing this - sorry about that!

src/metabase/api/api_key.clj
johnswanson: Good point, done!

PR #37052, opened 2023-12-21
CSV appends: Add based_on_upload to GET /api/card endpoint

calherries:
This PR implements the BE part of milestone 1, merge 2 for the CSV appends project.

[epic](https://github.com/metabase/metabase/issues/35614)
[eng doc](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df?pvs=4#1803a6c780af4c41bc0616d563a89100)

From the eng doc ([block link](https://www.notion.so/metabase/Eng-Allow-appending-more-data-to-CSV-uploads-4c40e06fa64744c699588b8d8576e5df?pvs=4#94b5b586e0ec4b168c21eb8bfdb9e685)):

### **Deriving the ability for a user to upload to a table or not**

To display the “Upload data to this model” button, the FE needs to know that the user is looking at a model backed by an uploaded table, and that they can upload to it.

Three things need to be true:

1. The model must have a GUI question which selects from a single table (no joins)
2. The user must have permission to append to the table (same as append endpoint permissions)
3. The table must be created with a CSV upload (is_upload=true)

This logic will be implemented on the BE, where the /api/card requests will respond with a property **`based_on_upload=<table-id>`** if the above criteria are met. `based_on_upload` does not need to be derived for questions, only saved models

**BE eng spec:**

- GET, PUT, and POST `/api/card/:id` endpoints should return the card with a new property `based_on_upload=<table-id>`
- [x]  Test: card must be a model
- [x]  Test: user must have permission to append to the table (same as append endpoint permissions)
- [x]  Test: table must be created with a CSV upload (is_upload=true)
- [x]  Test: card must be a GUI question which selects from a single table


src/metabase/driver/sql_jdbc.clj
```diff
@@ -171,6 +171,16 @@
                                :dialect (sql.qp/quote-style driver)))]
     (qp.writeback/execute-write-sql! db-id sql)))
 
+(defmethod driver/add-columns! :sql-jdbc
```
qnkhuat, line 174: Why is this needed now?

test/metabase/api/card_test.clj
```diff
@@ -3014,3 +3015,56 @@
...
+                                                                                             :source-table table-id
+                                                                                             :condition    [:= 1 2] ;; field-ids don't matter
+                                                                                             :alias        "SomeAlias"}]}}}]
+                (is (=? {:based_on_upload nil} (mt/user-http-request :crowberto :get 200 (str "card/" card-id'))))))
+            (testing "\nIf the table is not based on uploads, based_on_upload should be nil"
+              (t2/update! :model/Table table-id {:is_upload false})
+              (is (=? {:based_on_upload nil} (mt/user-http-request :crowberto :get 200 (str "card/" card-id))))
+              (t2/update! :model/Table table-id {:is_upload true}))
+            (testing "\nIf uploads are disabled, based_on_upload should be nil"
+              (mt/with-temp-copy-of-db
```
qnkhuat, line 3055: technically you don't need this because `mt/with-temp` will rollback all changes when exists the transaction.

qnkhuat (approved):
Solid stuff

src/metabase/driver/sql_jdbc.clj
calherries: It looks like this got copied from above, probably by git while merging. I've removed it 2fe2875b48c0a197c30eb275f9b9e9533e60d431

PR #37058, opened 2023-12-21
36797 partial permission graph

### LINKED ISSUE #36797: FE support partial permissions graph
- Update the reducers and pages for permissions to build out the permissions graph as we go
- Show a modal when permissions are out of date
- Will need to have a way to determine if something is loading

npfitz:
Closes https://github.com/metabase/metabase/issues/36797

### Description
This PR leverages 2 new endpoints to work with the permission graph: `/api/permissions/graph/group/:groupId` and `/api/permissions/graph/database/:databaseId`. These endpoints can be called to get specific parts of the permission graph much quicker than getting the whole graph at once from `/api/permissions/graph`, and are useful because the pages that manage data permissions only display 1 group or 1 database at a time.

When navigating to `/admin/permissions/data/group` (or `database`), we will programmatically load the permissions for the All Users and Admin groups, as they're used for some helpful warnings when making edits. From there, as you navigate between groups or databases, we will make fresh API calls and update our internal state with new pieces of the permission graph. Editing and saving the permission graph should work as it always has, but if the revision number changes while you are making edits then a modal will appear and force you to refresh the page. While this is intrusive, it's similar to how trying to save an graph with the wrong revision number simply presents you with an error and forces you to start over. Aside from this new modal and a few new loading states, the end user experience should be largely the same

### How to verify
Go to Admin -> Permissions and Update Group or database permissions as you normally would. Permission override warnings should still work (if the All Users group has access to something, then limiting a specific group to that same thing is pointless), and saving should still work. In another tab if you make a change and save, then return to your original tab, you should see a modal if you continue to navigate around and make edits.

### De
chrome_eO1IVcf6YL

### Checklist

### DIFF
--- a/e2e/test/scenarios/permissions/admin-permissions.cy.spec.js
+++ b/e2e/test/scenarios/permissions/admin-permissions.cy.spec.js
@@ -491,6 +491,30 @@ describe("scenarios > admin > permissions", { tags: "@OSS" }, () => {
           ["Reviews", "Unrestricted", "Yes"],
         ]);
       });
+
+      it("should show a modal when a revision changes while an admin is editing", () => {
+        cy.intercept("/api/permissions/graph/group/1").as("graph");
+        cy.visit("/admin/permissions");
+
+        selectSidebarItem("collection");
+
+        modifyPermission(
+          "Sample Database",
+          DATA_ACCESS_PERMISSION_INDEX,
+          "Unrestricted",
+        );
+
+        cy.get("@graph").then(data => {
+          cy.request("PUT", "/api/permissions/graph", {
+            groups: {},
+            revision: data.response.body.revision,
+          }).then(() => {
+            selectSidebarItem("data");
+
+            modal().findByText("Someone just changed permissions");
+          });
+        });
+      });
     });
 
     context("database focused view", () => {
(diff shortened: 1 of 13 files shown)


e2e/test/scenarios/permissions/admin-permissions.cy.spec.js
```diff
@@ -597,6 +621,30 @@ describe("scenarios > admin > permissions", { tags: "@OSS" }, () => {
           ["readonly", "Unrestricted", "Yes"],
         ]);
       });
+      it("should show a modal when a revision changes while an admin is editing", () => {
```
npfitz, line 624: I originally planned on having this test be navigating between DBs. That can still happen, but this demonstrates that we still check revisions when navigating to DB pages

frontend/src/metabase/admin/permissions/components/PermissionsPageLayout/PermissionsPageLayout.tsx
```diff
@@ -10,6 +10,13 @@ import { LeaveConfirmationModal } from "metabase/components/LeaveConfirmationMod
 import Modal from "metabase/components/Modal";
 import ModalContent from "metabase/components/ModalContent";
 
+import {
+  Modal as NewModal,
+  Text,
+  Button as NewButton,
```
npfitz, line 16: Should probably follow this up by removing the depreciated `<Modal>` and `<Button>` components

frontend/src/metabase/admin/permissions/components/PermissionsPageLayout/PermissionsPageLayout.tsx
```diff
@@ -132,6 +140,24 @@ function PermissionsPageLayout({
           {helpContent}
         </PermissionPageSidebar>
       )}
+      <NewModal
+        title="Someone just changed permissions"
+        opened={showRefreshModal}
+        size="lg"
+        padding="2.5rem"
+        withCloseButton={false}
+        onClose={() => true}
```
npfitz, line 149: onClose is a mandatory prop, despite the fact that we don't want this modal to be closable haha

frontend/src/metabase/admin/permissions/pages/DataPermissionsPage/DataPermissionsPage.tsx
```diff
@@ -48,10 +53,23 @@ function DataPermissionsPage({
 
   const loadPermissions = () => dispatch(loadDataPermissions());
```
npfitz, line 54: This probably needs to be updated so that we re-load from state, rather than make a new API call

luizarakaki (approved):
LGTM

iethree (approved):
looks good!

might be nice to follow up by slicing up the loading state to show the loader only on the table to make it less jumpy/flashy?

flash

PR #37062, opened 2023-12-21
Trend Charts: Address Milestone 2 Feedback

### LINKED ISSUE #37026: Milestone 2 Follow-up Tweaks
- [ ] Address feedback from Milestone 2 PR (was merged to unblock next milestone work)

---
**Demo of Periods Ago Input Updates**

https://github.com/metabase/metabase/assets/22608765/2c124743-1dbb-4afe-81cb-b79c89dde06c

JesseSDevaney:
- Closes: https://github.com/metabase/metabase/issues/37026


frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
```diff
@@ -35,67 +55,88 @@ export function SmartScalarComparisonWidget({
...
+      onClose={handleClose}
+      position="bottom-start"
+      shadow="sm"
+      closeOnItemClick={false}
+    >
       <Menu.Target>
-        <ButtonStyled
+        <Button
           data-testid={"comparisons-widget-button"}
+          styles={{ inner: { justifyContent: "space-between" } }}
```
JesseSDevaney, line 69: Starting point for https://github.com/metabase/metabase/pull/36666#discussion_r1432638800. There does not seem to be an easy to way to get the correct styles using `rightIcon` that will also be compatible with Mantine v7.

WDYT we should do to get the correct styles here but still use `rightIcon`. Note: `getStylesRef` was not working everywhere and does not seem to be a supported way of adjusting styles for component

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoOptionComponent.tsx
```diff
@@ -1,151 +1,95 @@
...
+import type { COMPARISON_TYPES } from "metabase/visualizations/visualizations/SmartScalar/constants";
 import { NumberInputStyled } from "./PeriodsAgoOptionComponent.styled";
 import { MenuItemStyled } from "./MenuItem.styled";
+import type { HandleEditedValueChangeType } from "./SmartScalarSettingsWidgets";
 
-type PeriodsAgoOptionComponentProps = {
-  isSelected: boolean;
+type PeriodsAgoMenuOptionProps = {
+  "aria-selected": boolean;
+  editedValue?: SmartScalarComparisonPeriodsAgo;
```
JesseSDevaney: In response to figuring out a more elegant solution for handling `periodsAgo` value submission on menu closing, this was introduced. It might help to review just the file itself since most of the diff is removing the hacks.

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
```diff
@@ -1,27 +1,47 @@
...
+        setOpen(false);
+      }
+    },
+    [onChange, setEditedValue, setOpen],
+  );
+
+  const handleClose = useCallback(() => {
+    if (_.isEqual(selectedValue, editedValue)) {
+      return;
+    }
```
JesseSDevaney, line 41: Let me know if discussing these changes would be better done in a huddle

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoMenuOption.tsx
```diff
@@ -0,0 +1,102 @@
...
+  editedValue?: SmartScalarComparisonPeriodsAgo;
+  maxValue: number;
+  name: string;
+  handleChange: HandleEditedValueChangeType;
+  type: typeof COMPARISON_TYPES.PERIODS_AGO;
+};
+
+const MIN_VALUE = 2;
+
+export function PeriodsAgoMenuOption({
```
JesseSDevaney, line 21: Updated component name according to https://github.com/metabase/metabase/pull/36666#discussion_r1432645994

JesseSDevaney (commented):
Hey Anton! Thanks for your feedback on the Milestone 2 PR. Here are the updates according to the suggested changes.

Also added in here is the `periodsAgo` input updates (and how we handle menu updates) according to our discussion. You can reference the demo video in this issue if you need further context to what I mean.
- https://github.com/metabase/metabase/issues/37026

kulyk (changes requested):
Looks great! I think we should address two things before merging:

* rework the `PeriodsAgoMenuOption` so we don't need to focus/blur the input programmatically
* remove the `@ts-ignore` comment

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoMenuOption.tsx
```diff
@@ -0,0 +1,102 @@
...
+    },
+    [onChange, type, value],
+  );
+  const handleInputClick = useCallback(
+    (e: MouseEvent<HTMLInputElement>) => {
+      e.stopPropagation();
+      selectInput();
+    },
+    [selectInput],
+  );
```
kulyk, line 80: nit: i think the input will be focused anyway and `selectInput` could be redundant

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/SmartScalarSettingsWidgets.tsx
kulyk: What kind of problem did you have with `getStylesRef`? I think it worked for me, but the tricky part was to use `.` in front of `getStylesRef()` call to get a valid CSS selector:

```tsx
const labelClassName = getStylesRef("label");

const MyStyledComponent = styled(MantineComponent)`
  ${labelClassName} {
    // doesn't work
  }

  .${labelClassName} {
    // works
  }
`
```

kulyk (approved):
🚀

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/PeriodsAgoMenuOption.tsx
JesseSDevaney: It will be focused, but it will not have all the inner text selected. I think it is helpful to have a fresh typing canvas instead of having to backspace if the number of periods ago you want does not start with a 2.

**Demo** (first half: without select, 2nd half: with select)

https://github.com/metabase/metabase/assets/22608765/76b81bac-2234-4e88-a5da-d0ac4ecd6128

JesseSDevaney:
Response to https://github.com/metabase/metabase/pull/37062#discussion_r1435249166. 

All `getStylesRef("label")` is doing here, is adding the supplied string `"label"` to the end of another string `"__ref"` such that `getStylesRef("label") === "__ref-label"`.

image
image

The issue is that the element I am trying to target is the div which has no existing `__ref-inner` to target.
image

PR #37077, opened 2023-12-22
add "application-colors" in createMockSettings

npretto:
### Description

This setting was not mocked ~nor present in the settings api type~ (it is, in the EnterpriseSettings type).
It was causing a console.error in all jest test that use enterprise plugins
```
console.error
    TypeError: Cannot convert undefined or null to object
        at Function.entries (<anonymous>)
        at entries (/Users/npretto/metabase/metabase/enterprise/frontend/src/metabase-enterprise/whitelabel/lib/whitelabel.js:6:48)
```

### How to reproduce
Run any jest test that uses the EE plugins, for example `yarn jest frontend/src/metabase/selectors/whitelabel/tests/premium.unit.spec.ts`
It should not print that error anymore

⚠️ **I'll be OOO for the next 2 weeks so feel free to take over this PR** 🏝️

### DIFF
--- a/frontend/src/metabase-types/api/mocks/settings.ts
+++ b/frontend/src/metabase-types/api/mocks/settings.ts
@@ -133,6 +133,7 @@ export const createMockSettingDefinition = (
 export const createMockSettings = (opts?: Partial<Settings>): Settings => ({
   "admin-email": "admin@metabase.test",
   "anon-tracking-enabled": false,
+  "application-colors": {},
   "application-font": "Lato",
   "application-font-files": [],
   "application-name": "Metabase",
--- a/frontend/src/metabase-types/api/mocks/settings.ts
+++ b/frontend/src/metabase-types/api/mocks/settings.ts
@@ -1,3 +1,4 @@
+import type { EnterpriseSettings } from "metabase-enterprise/settings/types";
 import type {
   Engine,
   EngineField,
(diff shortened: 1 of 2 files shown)


frontend/src/metabase-types/api/mocks/settings.ts
```diff
@@ -130,9 +131,12 @@ export const createMockSettingDefinition = (
   ...opts,
 });
 
-export const createMockSettings = (opts?: Partial<Settings>): Settings => ({
+export const createMockSettings = (
+  opts?: Partial<Settings | EnterpriseSettings>,
+): EnterpriseSettings => ({
```
npretto, line 136: EnterpriseSettings is Settings + some optional other types, so ti shouldn't create issues
https://github.com/metabase/metabase/blob/2550abbbc1d11755a6a0cf35ecdfb396c48a0c29/enterprise/frontend/src/metabase-enterprise/settings/types.ts#L12-L19

frontend/src/metabase-types/api/mocks/settings.ts
```diff
@@ -130,9 +131,12 @@ export const createMockSettingDefinition = (
   ...opts,
 });
 
-export const createMockSettings = (opts?: Partial<Settings>): Settings => ({
+export const createMockSettings = (
+  opts?: Partial<Settings | EnterpriseSettings>,
```
npretto, line 135: I copied this from https://github.com/metabase/metabase/blob/085c37800c59d93d9a92abfe334f8f5b212c5022/frontend/test/__support__/settings.ts#L8-L9
But considering the other comment, maybe we can just do 
```suggestion
  opts?: Partial<EnterpriseSettings>,
```

deniskaber (approved):

Issue #36800, opened 2023-12-14
Support more date formats for CSV uploads


calherries:
Users want to upload dates in formats like MM/DD/YYYY and Aug 12 2023.

Currently we support [these](https://github.com/metabase/metabase/blob/266b00c715d107d1e68d3c53006841e7c84bb0f4/src/metabase/upload/parsing.clj#L32-L86) formats

tl;dr
```
-- Dates
2023-01-01

-- Datetimes
2023-01-01 00:00:00.000
2023-01-01T00:00:00.000
where minutes, seconds, and fractional seconds are optional

-- Datetimes with offsets
2023-01-01 00:00:00+00:00:00
2023-01-01T00:00:00+00:00:00
where minutes, seconds, and fractional seconds are optional, also in the offset
```

PR #37078, opened 2023-12-22
CSV uploads: support more date formats

calherries:
Pursuant towards https://github.com/metabase/metabase/issues/36800

Extends the date parser for CSV uploads to accept more formats.

This PR doesn't include formats that are ambiguous, like MM/dd/YYYY vs dd/MM/YYYY. That will be implemented separately because they take considerably more effort. The approach is being discussed [here](https://metaboat.slack.com/archives/C04S696LRUM/p1703255693104709). 

**Before:**
```
yyyy-MM-dd
```
**After:**
```
yyyy-MM-dd
MMM dd yyyy
MMM dd, yyyy
dd MMM yyyy
dd MMM, yyyy
MMMM d yyyy
MMMM d, yyyy
d MMMM yyyy
d MMMM, yyyy
```

See tests for examples


src/metabase/upload/parsing.clj
```diff
@@ -29,19 +31,56 @@
...
-    - yyyy-MM-dd"
+    - yyyy-MM-dd
+    - MMM dd yyyy
+    - MMM dd, yyyy
+    - dd MMM yyyy
+    - dd MMM, yyyy
+    - MMMM d yyyy
+    - MMMM d, yyyy
+    - d MMMM yyyy
+    - d MMMM, yyyy"
```
calherries, line 75: I decided to keep this docstring using `yyyy` instead of `uuuu` because `uuuu` is less well known and only used because we want to have strict parsing. It feels more of an implementation detail

qnkhuat (approved):
lovely

src/metabase/upload/parsing.clj
crisptrutski: I have somehow never come across the `u` family of patterns before.

The following advice (from [this SO](https://stackoverflow.com/questions/41177442/uuuu-versus-yyyy-in-datetimeformatter-formatting-pattern-codes-in-java)) makes some sense:

> If you use y you should also use G. Since G is rarely used, the correct year symbol is u, not y, otherwise a non-positive year will show incorrectly.

That said, I think this is one of those "saved by the mission critical legacy code out there" situations - in practice G can only get bumped in a world where all existing software has become obsolete 🙂

crisptrutski: Oh nevermind, I see you're only using `y` for the docstring - a nice solution! ➕

crisptrutski (approved):

PR #37091, opened 2023-12-22
docs - upload privileges

jeff-bruemmer:
Adds section on upload privileges.

Are these the correct privileges to grant? Anything else we should add?


Tony-metabase (approved):

cdeweyx (approved):
Looks good 👍

PR #37095, opened 2023-12-22
Add CopyTextInput component

shaunlebron:
Screen Shot 2024-01-02 at 2 31 26 PM

PR #37097, opened 2023-12-22
API Keys: Add auth card and manage page

shaunlebron:
Group page:
Screen Shot 2024-01-02 at 3 07 52 PM

Create:
Screen Shot 2024-01-02 at 3 07 22 PM

Read:
Screen Shot 2024-01-02 at 3 07 16 PM

Empty state:
Screen Shot 2024-01-02 at 4 35 41 PM

Update:
Screen Shot 2024-01-02 at 3 09 09 PM

Delete:
Screen Shot 2024-01-02 at 3 07 32 PM

### DIFF
--- a/e2e/snapshot-creators/default.cy.snap.js
+++ b/e2e/snapshot-creators/default.cy.snap.js
@@ -307,6 +307,10 @@ function getDefaultInstanceData() {
     instanceData.databases = databases;
   });
 
+  cy.request("/api/permissions/group").then(({ body: groups }) => {
+    instanceData.groups = groups;
+  });
+
   cy.request("/api/collection").then(({ body: collections }) => {
     instanceData.collections = collections;
 
--- a/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
+++ b/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
@@ -140,7 +140,7 @@ describe("scenarios > admin > settings > API keys", () => {
       .clear()
       .type("Different key name");
 
-    cy.findByLabelText(/Select a group/).click();
+    cy.findByLabelText(/group/).click();
     cy.findByRole("listbox").findByText("collection").click();
 
     cy.button("Save").click();
@@ -247,7 +247,7 @@ describe("scenarios > admin > settings > API keys", () => {
             "Edited Dashboard Name",
           ).then(() => {
             cy.signInAsAdmin();
-            visitDashboard(ORDERS_QUESTION_ID);
+            visitDashboard(ORDERS_DASHBOARD_ID);
             cy.findByTestId("dashboard-header").findByText(
               "Edited Dashboard Name",
             );
--- a/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
+++ b/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
@@ -188,7 +188,6 @@ describe("scenarios > admin > settings > API keys", () => {
     it("should allow creating questions and dashboards with an API key", () => {
       createApiKey("Test API Key One", ADMINISTRATORS_GROUP_ID).then(
         ({ body }) => {
-          console.log(body);
           const apiKey = body.unmasked_key;
           createQuestionForApiKey(apiKey).then(({ body }) => {
             const questionId = body.id;
--- a/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
+++ b/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
@@ -1,16 +1,9 @@
-import {
-  popover,
-  restore,
-  visitDashboard,
-  visitQuestion,
-} from "e2e/support/helpers";
-import type { ApiKey } from "metabase-types/api";
+import { restore, visitDashboard, visitQuestion } from "e2e/support/helpers";
 
 import {
   ALL_USERS_GROUP_ID,
   READONLY_GROUP_ID,
   NOSQL_GROUP_ID,
-  COLLECTION_GROUP_ID,
   ADMINISTRATORS_GROUP_ID,
   ORDERS_QUESTION_ID,
   ORDERS_DASHBOARD_ID,
@@ -21,9 +14,6 @@ import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
 const { PRODUCTS_ID } = SAMPLE_DATABASE;
 
 describe("scenarios > admin > settings > API keys", () => {
-  // TODO: replace intercepts below with actual requests to test backend
-  const mockRows: ApiKey[] = [];
-
   beforeEach(() => {
     cy.intercept("GET", "/api/api-key/count").as("getKeyCount");
     cy.intercept("GET", "/api/api-key").as("getKeys");
@@ -192,41 +182,6 @@ describe("scenarios > admin > settings > API keys", () => {
 
     cy.button("Done").click();
     cy.findByTestId("api-keys-table").findByText(/mb_/);
-  });
-
-  it.skip("should warn before deleting a group with API keys", () => {
-    createApiKey("Personal API Key", COLLECTION_GROUP_ID);
-
-    cy.visit("/admin/people/groups");
-    cy.wait("@getKeys");
-    cy.get(".ContentTable")
-      .contains("collection")
-      .closest("tr")
-      .should("contain", "(Includes 1 API Key")
-      .icon("ellipsis")
-      .click();
-    popover().findByText("Remove Group").click();
-    cy.get(".Modal")
-      .findByText(/move the API Keys/)
-      .click();
-    cy.url().should("match", /\/admin\/settings\/authentication\/api-keys$/);
-  });
-
-  it.skip("should show API keys when viewing Group details", () => {
-    cy.intercept("GET", "/api/api-key", req => req.reply(200, mockRows)).as(
-      "getKeys",
-    );
-    cy.visit("/admin/people/groups/3");
-    cy.wait("@getKeys");
-    cy.get(".ContentTable")
-      .findByText("Personal API Key")
-      .closest("tr")
-      .icon("link")
-      .as("apiKeysLink")
-      .realHover();
-    cy.findByRole("tooltip").should("contain", "Manage API keys");
-    cy.get("@apiKeysLink").click();
-    cy.url().should("match", /\/admin\/settings\/authentication\/api-keys$/);
   });
 
   describe("api key actions", () => {
--- a/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
+++ b/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
@@ -1,4 +1,9 @@
-import { restore, visitDashboard, visitQuestion } from "e2e/support/helpers";
+import {
+  restore,
+  visitDashboard,
+  visitQuestion,
+  createApiKey,
+} from "e2e/support/helpers";
 
 import {
   ALL_USERS_GROUP_ID,
@@ -263,13 +268,6 @@ describe("scenarios > admin > settings > API keys", () => {
     });
   });
 });
-
-const createApiKey = (name: string, group_id: number) => {
-  return cy.request("POST", "/api/api-key", {
-    name,
-    group_id,
-  });
-};
 
 const createQuestionForApiKey = (apiKey: string) => {
   cy.signOut();
--- a/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
+++ b/e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
@@ -1,274 +1,192 @@
-import type { CyHttpMessages } from "cypress/types/net-stubbing";
-
 import { popover, restore } from "e2e/support/helpers";
 import type { ApiKey } from "metabase-types/api";
 
-const MOCK_ROWS: ApiKey[] = [
-  {
-    name: "Development API Key",
-    id: 1,
-    group: {
-      id: 1,
-      name: "All Users",
-    },
-    creator_id: 1,
-    masked_key: "asdfasdfa",
-    created_at: "2010 Aug 10",
-    updated_at: "2010 Aug 10",
-    updated_by: {
-      id: 10,
-      common_name: "John Doe",
-    },
-  },
-  {
-    name: "Production API Key",
-    id: 2,
-    group: {
-      id: 2,
-      name: "Administrators",
-    },
-    creator_id: 1,
-    masked_key: "asdfasdfa",
-    created_at: "2010 Aug 10",
-    updated_at: "2010 Aug 10",
-    updated_by: {
-      id: 11,
-      common_name: "Jane Doe",
-    },
-  },
-  {
-    name: "Personal API Key",
-    id: 3,
-    group: {
-      id: 3,
-      name: "collection",
-    },
-    creator_id: 1,
-    masked_key: "asdfasdfa",
-    created_at: "2010 Aug 10",
-    updated_at: "2010 Aug 10",
-    updated_by: {
-      id: 12,
-      common_name: "Jane Doe",
-    },
-  },
-];
-
-const MOCK_GROUPS = [
-  // NOTE: this is only here temporarily (remove when backend is working)
-  {
-    id: 2,
-    name: "Administrators",
-    member_count: 1,
-  },
-  {
-    id: 1,
-    name: "All Users",
-    member_count: 9,
-  },
-  {
-    id: 3,
-    name: "collection",
-    member_count: 4,
-  },
-  {
-    id: 4,
-    name: "data",
-    member_count: 2,
-  },
-  {
-    id: 6,
-    name: "nosql",
-    member_count: 1,
-  },
-  {
-    id: 5,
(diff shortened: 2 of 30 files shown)


shaunlebron:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/37097?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #37095** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #37096** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #37097** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37097?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
        * **PR #37098** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37097?utm_source=stack-comment).

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Stack spacing="md">
+              <FormTextInput
+                name="name"
+                label={t`Key name`}
+                size="sm"
+                required
+              />
+              <FormGroupWidget
+                name="group_id"
+                label={t`Select a group to inherit its permissions`}
```
jeff-bruemmer, line 61: Ryan flagged this phrasing as potentially confusing (I don't think we should use 'inheritance' here, as this is more about membership).

Maybe (as Ryan suggested):

> Which group should this key belong to?

Or we could elaborate:

```suggestion
                label={t`Which group should this key belong to? The key will have the same permissions granted to that group.`}
```

e2e/snapshot-creators/default.cy.snap.js
```diff
@@ -307,6 +307,10 @@ function getDefaultInstanceData() {
     instanceData.databases = databases;
   });
 
+  cy.request("/api/permissions/group").then(({ body: groups }) => {
+    instanceData.groups = groups;
+  });
```
iethree, line 312: we need to dump these group IDs to safely use them in our e2e tests.

frontend/src/metabase/plugins/builtin/auth/apiKeys.js
iethree, line 1: i rather dislike using plugins for OSS features, but this is the established pattern, so we won't rock the boat too much here.

rafpaf:
image

@iethree I figured I should double check with someone, is "API key deleted can't be recovered" right, or should it be something like: "The deleted API key can't be recovered"? Or maybe: "Deleted API keys can't be recovered. You'll have to create a new key." Who's the right person to check with?

e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    cy.button("Done").click();
+    cy.findByTestId("api-keys-table").findByText(/mb_/);
+  });
+
+  describe("api key actions", () => {
+    it("should allow creating questions and dashboards with an API key", () => {
+      createApiKey("Test API Key One", ADMINISTRATORS_GROUP_ID).then(
+        ({ body }) => {
+          console.log(body);
```
rafpaf, line 191: Is `console.log` used here for debugging, and if so, is it still needed?

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+              <FormGroupWidget
+                name="group_id"
+                label={t`Which group should this key belong to? The key will have the same permissions granted to that group.`}
+                size="sm"
+                required
+              />
+              <Text
+                my="sm"
+                size="sm"
+              >{t`We don’t version the Metabase API. We rarely change API endpoints, and almost never remove them, but if you write code that relies on the API, there’s a chance you might have to update your code in the future.`}</Text>
```
rafpaf, line 68: An almost literally nit-sized nitpick: There's a smart single quote here but normal single quotes in other strings. Does it matter?

e2e/test/scenarios/admin-2/api-keys.cy.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+    cy.button("Regenerate API Key").click();
+    cy.button("Regenerate").click();
+    cy.wait("@regenerateKey");
+    cy.findByLabelText("The API key").should("include.value", "mb_");
+
+    cy.wait("@getKeys").then(({ response }) => {
+      const { created_at, updated_at } = response?.body[0];
+      // after regeneration, created_at and updated_at should be different
+      // they're too close to check via UI though
+      expect(created_at).to.not.equal(updated_at);
```
rafpaf, line 180: I'm just thinking out loud, and this might be well known but I'll mention it anyway: Cypress's `equal` method checks for strict equality, so the following test passes:

`expect(new Date("2024-01-05")).to.not.equal(new Date("2024-01-05"));`

The silliness of this is pretty vivid if you look at Cypress's output:
image

Since `created_at` and `updated_at` are strings, the `not.equal` pattern works perfectly well on line 180 above, but I think the pattern should perhaps be viewed with a bit of skepticism.

This is an alternative, but it's probably overkill:

`expect(new Date(created_at).getTime()).to.not.equal(new Date(updated_at).getTime());`

Even this has an amusing drawback: if both variables are `undefined`, the assertion will pass! `new Date(undefined).getTime()` is `NaN` and, bless JS, `NaN !== NaN`. To avoid this one could add `expect(new Date(created_at).getTime()).should('be.gt', 0)`

iethree: Good to know, but I still feel pretty good about this check since we're not dealing with objects at all in api responses.

iethree:
> @iethree I figured I should double check with someone, is "API key deleted can't be recovered" right, or should it be something like: "The deleted API key can't be recovered"? Or maybe: "Deleted API keys can't be recovered. You'll have to create a new key." Who's the right person to check with?

Good catch. I actually updated this and most of the other strings after @jeff-bruemmer 's review, but failed to update the screenshots. It currently reads: "You won't be able to recover a deleted API key. You'll have to create a new key."

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
iethree: I think we use a mix throughout the app. no harm in making these consistent though. 👍

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+              <FormGroupWidget
+                name="group_id"
+                label={t`Which group should this key belong to? The key will have the same permissions granted to that group.`}
+                size="sm"
+                required
+              />
+              <Text
+                my="sm"
+                size="sm"
+              >{t`We don't version the Metabase API. We rarely change API endpoints, and almost never remove them, but if you write code that relies on the API, there's a chance you might have to update your code in the future.`}</Text>
```
rafpaf, line 68: I'm curious, will beginners find the phrase "version the Metabase API" clear enough?

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+      >
+        <FormProvider initialValues={{}} onSubmit={handleSubmit}>
+          <Form data-testid="create-api-key-modal">
+            <Stack spacing="md">
+              <FormTextInput
+                name="name"
+                label={t`Key name`}
+                size="sm"
+                required
+              />
```
rafpaf, line 58: I'm not sure this is important, but it seemed worth mentioning, just in case. If you enter nothing in this field, the form shows the error "value must be a non-blank string." But if you enter a very, very long string, you get an error message with SQL in it, like this:

image

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+              <FormGroupWidget
+                name="group_id"
+                label={t`Which group should this key belong to? The key will have the same permissions granted to that group.`}
+                size="sm"
+                required
+              />
```
rafpaf, line 64: If I leave this field blank, the 'Create' button becomes disabled and looks like this:

image

A few seconds later, the text changes back to 'Create' but the button is still disabled:

image

Is this expected behavior?

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
rafpaf: Here's another slightly unexpected bit of behavior in this modal.

If I re-use an API key name, I get this normal-looking error message:

image

However, when I blur the Key name text field, the phrase 'An error occurred' pops up:

image

rafpaf: The following issue seems like it might be a bit more serious.

If I create a key named "a" and another key named "a " (with a space), that works. But if I then delete both keys and try to create a key named "a", I get this error:

image

I think I've now put my database into a state where it is impossible to make a key named "a".

rafpaf: Another small issue relating to key names. Since non-alphanumeric characters are stripped from the key name ([here](https://github.com/metabase/metabase/blob/4d5b621879173c187d28bb58b234197d31dfbe7e/src/metabase/api/api_key.clj#L55)) to make an email address for internal use, if you make a key named `$a` and try to make a second key named `@a`, you get an error:

image

If this is a genuine problem, maybe we can append a timestamp before the '@' in the email address, in the backend?

rafpaf (approved):
This looks great! I found some little behaviors in the API Keys interface that seemed worth bringing up, but I don't know if they are blockers, so I'll leave it up to @iethree to say whether they are.

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
npfitz: I'd expect anyone generating API keys would have a development background and understand what the message is trying to convey.

frontend/src/metabase/admin/settings/components/ApiKeys/EditApiKeyModal.tsx
```diff
@@ -0,0 +1,184 @@
...
+                <FormGroupWidget
+                  name="group_id"
+                  label={t`Which group should this key belong to? The key will have the same permissions granted to that group.`}
+                  size="sm"
+                />
+                <FormTextInput
+                  name="masked_key"
+                  label={t`API Key`}
+                  size="sm"
+                  styles={{ input: { fontFamily: "Monaco, monospace" } }}
```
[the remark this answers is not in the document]
npfitz: That would probably work, but I think the correct way to do this with mantine is to add a monospace variant to our `Text.styled.tsx` component that applies the font family. I see `fontFamilyMonospace` is defined in our theme, but I don't see anywhere that it's actually used.

I believe in general a lot of the stuff in the `styled-components` folder is probably deprecated. @ranquild could confirm.

frontend/src/metabase/admin/settings/components/ApiKeys/ManageApiKeys.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  useEffect(() => {
+    refreshList();
+  }, [refreshList]);
+
+  const isShowingEmptyTable = !loading && !error && apiKeys?.length === 0;
+
+  return (
+    <>
+      {modal === "create" ? (
```
npfitz, line 121: I think we generally use redux state or routes to handle stuff like this, but but it's not something I'd get too hung up over

npfitz (approved):
It all looks pretty good to me. Left a few comments, mostly on things @rafpaf already mentioned, but I gave the branch a spin and it works. Super cool to see api keys in action too

frontend/src/metabase/admin/settings/components/ApiKeys/ManageApiKeys.tsx
rafpaf: nonblocking suggestion: To avoid the nested ternaries, some of this logic could be extracted into a `renderModal` function.

frontend/src/metabase/admin/settings/components/ApiKeys/ManageApiKeys.tsx
```diff
@@ -0,0 +1,169 @@
...
+    [],
+  );
+
+  const handleClose = () => setModal(null);
+
+  useEffect(() => {
+    refreshList();
+  }, [refreshList]);
+
+  const isShowingEmptyTable = !loading && !error && apiKeys?.length === 0;
```
rafpaf, line 117: I'm not yet certain why this variable is called `isShowingEmptyTable`. Why "showing"? It makes me wonder if the table can be hidden. Maybe `tableIsEmpty` or `isTableEmpty` could work well here.

frontend/src/metabase/admin/settings/components/ApiKeys/EditApiKeyModal.tsx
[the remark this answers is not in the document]
iethree:
```
import { getThemeOverrides } from "metabase/ui/theme";
const { fontFamilyMonospace } = getThemeOverrides();
```

does the trick, not sure if it makes sense to add a variant for the inputs, this is a pretty niche case.

frontend/src/metabase/admin/settings/components/ApiKeys/ManageApiKeys.unit.spec.tsx
```diff
@@ -0,0 +1,183 @@
...
+        id: 10,
+      },
+    },
+  ]);
+  renderWithProviders(<ManageApiKeys />);
+  await waitFor(async () => {
+    expect(
+      await fetchMock.calls("path:/api/api-key", { method: "GET" }),
+    ).toHaveLength(1);
+  });
```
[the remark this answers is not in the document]
iethree: there's a lint rule for this, maybe we should enable it 🤔

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
iethree: I think there might be a backend issue that isn't deleting users when API keys are deleted? @noahmoss (John is out this week) do you know if that is working as intended?

noahmoss: If I recall correctly, it was a conscious decision not to delete the users, since they could be the owners of existing content and we'd still want their names to display correctly on the UI.

It sounds like there are some bugs here though. We need to make sure that emails are unique, even when API Key names are the same as previously-deleted ones. 

Could you file an issue (or issues) and assign them to John for when he's back?

frontend/src/metabase/admin/settings/components/ApiKeys/EditApiKeyModal.tsx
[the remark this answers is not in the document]
npfitz: I'm happy with the middle ground 👍

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
iethree: I added `Yup` Schema to these forms and added initialValues, which cleans up most of this.

frontend/src/metabase/admin/settings/components/ApiKeys/ManageApiKeys.tsx
iethree: Not a fan of using global state to handle these little local modals. Extracted them into a `ApiKeysModals` component to encapsulate the modal logic instead of having nested ternaries.

frontend/src/metabase/admin/settings/components/ApiKeys/CreateApiKeyModal.tsx
iethree: I'm not too worried about the length errors (though I did add a max length to the input). what is scary is what having a 250 char key name does to the table formatting 😱

will constrain it.

PR #37102, opened 2023-12-25
Bigquery fix sync for tables that require a filter

qnkhuat:
Milestone 3 of https://github.com/metabase/metabase/issues/36668

Previously, when syncing partitioned tables that require a partition filter, we failed to fingerprint and scan field values because the query we used in those sync steps did not include a filter on the partitioned field.

This PR is about fixing those sync steps by fixing the queries to include the required filter.


src/metabase/sync/analyze/fingerprint.clj
qnkhuat, line 1: nothing changed here, only indentations

src/metabase/db/metadata_queries.clj
```diff
@@ -27,12 +27,52 @@
...
+                                           {:order-by [[:name :asc]]})
+                            (throw (ex-info (format "No partitioned field found for table: %d" table-id)
+                                            {:table_id table-id})))
+        filter-form     (partition-field->filter-form partition-field)]
+    (update query :filter (fn [existing-filter]
+                            (if (some? existing-filter)
+                              [:and existing-filter filter-form]
+                              filter-form)))))
+
+(defn- field-mbql-query
```
qnkhuat, line 58: this function is used in `field-distinct-values`, which is used to get a list of values for field values during the sync process.

src/metabase/db/metadata_queries.clj
```diff
@@ -126,7 +166,11 @@
                                                  [:expression expression-name]
                                                  [:field (u/the-id field) nil])))
                           :limit        limit}
-                   order-by (assoc :order-by order-by))
+                   order-by
+                   (assoc :order-by order-by)
+
+                   (:database_require_filter table)
```
qnkhuat, line 172: table-sample-rows for bigquery will use this query when sync a time ingestion table

modules/drivers/bigquery-cloud-sdk/src/metabase/driver/bigquery_cloud_sdk.clj
```diff
@@ -273,17 +284,24 @@
...
+               ;; to make sure we don't break existing instances.
+               TableDefinition$Type/EXTERNAL TableDefinition$Type/SNAPSHOT}
+             (.. bq-table getDefinition getType))
+            ;; if the table is partitioned by ingestion time, using .list or .listTableData won't return values for this
+            ;; the _PARTITIONTIME field, so we need to fall back to using sql
+            (ingestion-time-partitioned-table? (:id table)))
+      (do (log/debugf "%s.%s is a view or a table partitioned by ingestion time, so we cannot use the list API; falling back to regular query"
                       dataset-id table-name)
           ((get-method driver/table-rows-sample :sql-jdbc) driver table fields rff opts))
       (sample-table bq-table fields rff))))
```
qnkhuat, line 307: weirdly enough, the `.list` operator of big query can query the table without a filter even though the table requires a partition filter, so we don't need to do anything for range and date partitioned.

But for the time ingestion partition, we want the pseudo columns to be fingerprinted too, so we need to fallback to using an SQL query.

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
```diff
@@ -309,37 +310,83 @@
             (doall (map drop-table-if-exists! (keys table-name->is-filter-required?)))
             nil)))))))
 
-(deftest sync-pseudocolumn-for-ingestion-time-partitioned-table-test
-  (testing "for ingestion time partitioned tables, we should sync the pseudocolumn _PARTITIONTIME AND _PARTITIONDATE"
-    (mt/test-driver :bigquery-cloud-sdk
+(deftest full-sync-partitioned-table-test
```
qnkhuat, line 313: I incorporated the existing test `sync-pseudocolumn-for-ingestion-time-partitioned-table-test` into this one as it could reuse some of the setup.

Creating a table in bigquery is a slow process so I don't want to do that too many times.

src/metabase/db/metadata_queries.clj
```diff
@@ -27,12 +27,52 @@
...
+                              filter-form)))))
+
+(defn- field-mbql-query
+  [table mbql-query]
+  (cond-> mbql-query
+    true
+    (assoc :source-table (:id table))
+
+    ;; Some table requires a filter to be able to query the data
+    ;; Currently this only applied to Parittioned table in bigquery where the partition field
```
calherries, line 65:
```suggestion
    ;; Currently this only applied to Partitioned table in bigquery where the partition field
```

src/metabase/db/metadata_queries.clj
```diff
@@ -27,12 +27,52 @@
...
+                                           :active true
+                                           ;; prefer _PARTITIONDATE over _PARTITIONTIME for ingestion time query
+                                           {:order-by [[:name :asc]]})
+                            (throw (ex-info (format "No partitioned field found for table: %d" table-id)
+                                            {:table_id table-id})))
+        filter-form     (partition-field->filter-form partition-field)]
+    (update query :filter (fn [existing-filter]
+                            (if (some? existing-filter)
+                              [:and existing-filter filter-form]
+                              filter-form)))))
```
calherries, line 56: This code modifies a query, which should technically be done with MLv2, rather than being coupled to the query structure directly. It's probably okay to ship this because we already do it everywhere, and I'm not sure if there's a precedent for modifying queries using MLv2 on the BE yet. But would be huge if you could find the appropriate function in `metabase.lib.core` and use it here.

src/metabase/db/metadata_queries.clj
```diff
@@ -27,12 +27,52 @@
...
       :rows))
 
+(defn- partition-field->filter-form
+  "Given a partition field, returns the default value can be used to query."
+  [field]
+  (let [field-form [:field (:id field) nil]]
+    (condp #(isa? %2 %1) (:base_type field)
+      :type/Number   [:> field-form -9223372036854775808]
+      :type/Date     [:> field-form "0001-01-01"]
+      :type/DateTime [:> field-form "0001-01-01 00:00:00"])))
```
calherries, line 37: `:type/DateTime` is a superset of local datetimes and datetimes with offsets or zone IDs (`:type/DateTimeWithTZ`). This implementation might work for bigquery but is not guaranteed to work if there was a partition field with a `:type/DateTimeWithZoneOffset` base type in the future.

For bigquery the `_PARTITIONTIME` column will have a `:type/DateTimeWithLocalTZ` type, so this should work:
```suggestion
      :type/DateTimeWithLocalTZ [:> field-form "0001-01-01 00:00:00"])))
```

src/metabase/db/metadata_queries.clj
```diff
@@ -27,12 +27,52 @@
...
+    (update query :filter (fn [existing-filter]
+                            (if (some? existing-filter)
+                              [:and existing-filter filter-form]
+                              filter-form)))))
+
+(defn- field-mbql-query
+  [table mbql-query]
+  (cond-> mbql-query
+    true
+    (assoc :source-table (:id table))
```
calherries, line 62: This should also technically use a MLv2 function to modify the MBQL query

src/metabase/db/metadata_queries.clj
calherries: I've asked about this here https://metaboat.slack.com/archives/CKZEMT1MJ/p1704972904847889

modules/drivers/bigquery-cloud-sdk/test/metabase/driver/bigquery_cloud_sdk_test.clj
calherries: Very nice test.

test/metabase/db/metadata_queries_test.clj
```diff
@@ -74,6 +74,21 @@
             (let [query (#'metadata-queries/table-rows-sample-query table fields {:truncation-size 4})]
               (is (empty? (get-in query [:query :expressions]))))))))))
 
+(deftest mbql-on-table-requires-filter-will-include-the-filter-test
+  (mt/with-temp
+    [:model/Database db     {}
+     :model/Table    table  {:database_require_filter true :db_id (:id db)}
+     :model/Field    field1 {:name "name" :table_id (:id table)}
```
calherries, line 81: I think you should add the type here explicitly since you rely on it on line 85.
```suggestion
     :model/Field    field1 {:name "name" :table_id (:id table) :base_type :type/Integer}
```

calherries (approved):
Looks good! I'm approving on the condition that this gets resolved https://github.com/metabase/metabase/pull/37102/files#r1448701784

src/metabase/db/metadata_queries.clj
camsaul: Yeah it would definitely be good to use MLv2 for query manipulation like this going forward, for adding another filter you can just use `(lib/filter <query> <new filter>)` and it actually takes care this sort of stuff for you automatically. The benefit of using MLv2 everywhere is that we get to dogfood it a bit more, people get familiar with it, and we can have one big set of util functions to use everywhere instead of having to reinvent the wheel a bunch of times. Oh you also you don't need to know MBQL internals to use it.

src/metabase/db/metadata_queries.clj
qnkhuat: @calherries clause for datetime is needed because we can partition on a datetime field. Added this test as an [example](https://github.com/metabase/metabase/pull/37102/commits/bbf83d66f180fabe6c83449effd92ba39ac750ae)

calherries: Ah I stand corrected, in postgres and redshift it is fine to compare a timestamp with `:type/DateTimeWithZoneOffset` with a string literal that doesn't have any time zone information. This could be the case for other RDMBSs as well. And if it's not, a test will catch it. I'm okay with how it's implemented now then

PR #37107, opened 2023-12-25
Migrate dashboard with parameters -> query with filters transformation to MLv2

ranquild:
Modifies `_convertParametersToMbql` to use MLv2 under the hood. The function is used to create MBQL query filters for parameters. It is used to generate a query with filters when clicking on a question with connected parameters in a dashboard.

To make the change simple, uses `expressionClauseForLegacyExpression` and leaves raw mbql manipulation where it was. Dashboard parameters & their widgets heavily rely on (mostly invalid) mbql as their view model, and this is something that cannot be easily changed without migrating dashboard parameter widgets. Example of how it is used below. This is not important for this PR but just for the context. In this example the legacy date picker is used which is also MBQL-based:  https://github.com/metabase/metabase/blob/c8afac24548a9adf346ffff61ebc7d2cf7509184/frontend/src/metabase/components/DateAllOptionsWidget/DateAllOptionsWidget.tsx#L31

It should be the last place in the product that creates invalid filter clauses in **simple mode** (we still have notebook steps relying on it). With it gone, we can remove MLv1 filter validation when serializing the query into the URL.

### DIFF
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -63,7 +63,13 @@ const QUESTION_LINE_CHART = {
   display: "line",
   query: {
     aggregation: [["count"]],
-    breakout: [["field", ORDERS.CREATED_AT, { "temporal-unit": "month" }]],
+    breakout: [
+      [
+        "field",
+        ORDERS.CREATED_AT,
+        { "base-type": "type/DateTime", "temporal-unit": "month" },
+      ],
+    ],
     "source-table": ORDERS_ID,
     limit: 5,
   },
@@ -105,14 +111,21 @@ const DASHBOARD_FILTER_TIME = createMockActionParameter({
 
 const QUERY_FILTER_CREATED_AT = [
   "between",
-  ["field", ORDERS.CREATED_AT, null],
+  [
+    "field",
+    ORDERS.CREATED_AT,
+    {
+      "base-type": "type/DateTime",
+      "temporal-unit": "month",
+    },
+  ],
   "2022-08-01",
   "2022-08-31",
 ];
 
 const QUERY_FILTER_QUANTITY = [
   "=",
-  ["field", ORDERS.QUANTITY, null],
+  ["field", ORDERS.QUANTITY, { "base-type": "type/Integer" }],
   POINT_COUNT,
 ];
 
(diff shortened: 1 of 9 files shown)


frontend/test/metabase-lib/lib/queries/StructuredQuery-clean.unit.spec.js
```diff
@@ -145,22 +140,6 @@ describe("StructuredQuery", () => {
         const q = ordersTable.query().nest();
         expect(q.clean().query()).toEqual({ "source-table": ORDERS_ID });
       });
-
-      it("should remove clauses dependent on removed clauses in the parent", () => {
```
uladzimirdev, line 149: why did we remove those tests, did they become invalid?

uladzimirdev (commented):
Minor questions

frontend/src/metabase-lib/parameters/utils/mbql.js
```diff
@@ -177,21 +177,40 @@ function isFieldFilterParameterConveratableToMBQL(parameter) {
...
+    [parameter.target[1]],
+  );
+  if (columnIndex < 0) {
+    return null;
+  }
+
+  const column = columns[columnIndex];
+  const fieldRef = Lib.legacyRef(column);
 
   if (isDateParameter(parameter)) {
```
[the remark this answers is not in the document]
ranquild: Here it checks for the parameter type, not column type. Currently MLv2 doesn't know anything about parameters

frontend/test/metabase-lib/lib/queries/StructuredQuery-clean.unit.spec.js
ranquild: MLv2 disallows creating queries with invalid clauses, so this case became impossible

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -63,7 +63,13 @@ const QUESTION_LINE_CHART = {
   display: "line",
   query: {
     aggregation: [["count"]],
-    breakout: [["field", ORDERS.CREATED_AT, { "temporal-unit": "month" }]],
+    breakout: [
```
ranquild, line 66: This change already has impact - all field references in queries created from dashboards are valid now.

uladzimirdev (approved):

PR #37108, opened 2023-12-25
Add umbrella isQueryEditable function to ease migration from MLv1 query

ranquild:
We'd like to remove legacy `StructuredQuery` usage across the app. It's often used to check whether the user can edit and run the query. To make migration easier, let's consolidate all `query().isEditable()` calls under one function. Similar to how we call `question.isStructured()` or `question.isNative()`.

I've made it `isQueryEditable` and not `isEditable` because technically read/write permissions of the question itself are based on collection permissions; so let's make it explicit it's about queries.

### DIFF
--- a/frontend/src/metabase/query_builder/selectors.js
+++ b/frontend/src/metabase/query_builder/selectors.js
@@ -350,7 +350,7 @@ export const getQuestion = createSelector(
 );
 
 function isQuestionEditable(question) {
-  return question.isQueryEditable();
+  return question ? question.isQueryEditable() : false;
 }
 
 function areLegacyQueriesEqual(queryA, queryB, tableMetadata) {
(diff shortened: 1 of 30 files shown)


frontend/src/metabase/query_builder/selectors.js
```diff
@@ -350,7 +350,7 @@ export const getQuestion = createSelector(
 );
 
 function isQuestionEditable(question) {
-  return !question?.query().readOnly();
+  return question ? question.isQueryEditable() : false;
```
uladzimirdev, line 353:
```suggestion
  return question?.isQueryEditable() ?? false;
```

uladzimirdev (approved):
Cool 👍 Left one minore comment

PR #37110, opened 2023-12-26
Drop then recreate a column should refingerprint and analyze the field

### LINKED ISSUE #37047: Dropping and recreating a table doesn't update the effective type causing unwanted behaviour in the GUI
### Describe the bug

The issue spans different areas related to the sync most specifically the `database_type` and `effective_type` found in the table `metabase_field`

image

What happens is that when a table is dropped and re-created with a column that changed type the effective_type is not updated causing issues with the GUI editor during summation and aggregation. 

### To Reproduce

You need a Database for these replication steps. I am using PostgreSQL so connect This DB with metabase

1) Create a table:

```
CREATE TABLE test2
(
    id             serial PRIMARY KEY,
    this_is_a_text VARCHAR(50),
    this_is_a_time TIMESTAMP
);
```

2) Go to Admin -> Databases -> Sync Database Schema Now
3) Go to Question -> test2 -> You will notice that Metabase picks up the correct type of the field:

image

4) In fact if you connect to your appDB -> Go to table `metabase_field` and look for these two fields you will see:

image

5) Drop the above table:

```
DROP TABLE test2;
```

6) Re-Create another table which have the same column names but different types and insert some data:

```
CREATE TABLE test2
(
    id      serial PRIMARY KEY,
    this_is_a_text float8,
    this_is_a_time float8
);
```
```
INSERT INTO public.test2 (id, this_is_a_text, this_is_a_time)
VALUES (1::integer, 1::integer, 1::integer)
```

7) Go to Admin -> Databases -> Sync Database Schema Now
8) Go to Question -> test2 -> You will notice that Metabase picks up the old type for the fields and still thinks the fields are text and date:

image

In fact in this view you cannot even have the option to sum

image

9) On the other hand if you click visualize -> click on column -> you are given the option to Sum/Avg and it works when you do:

image

10) Go back to your appDB -> Go to table `metabase_field` and look for these two fields you will see only database_type changed: 

image

### Expected behavior

The expected bahavior is that the GUI editor and the column Modal are consistent. I should be getting the option to sum in both occasions ... I believe the issue is that `effective_type` should get updated as well

### Logs

None that are relevant

```JSON
1.47.9 and 1.48.1
```

### Severity

This is a serious problem because you are forced to go to the Table Metadata tab and change the field type if you have a table with a lot of changed columns this becomes a problem

### Additional context

_No response_

qnkhuat:
Closes https://github.com/metabase/metabase/issues/37047

When a field is dropped and then re-created with the same name but of a different database type, we need to make sure we reset its effective type and analyze + fingerprint it.

This doesn't handle cases where a field is re-created with the same database type. To do that, we need to change the way we ID field; currently, we consider the name to be unique per table; maybe we need to use something similar to the object ID from Postgres instead.


test/metabase/sync/sync_metadata/fields/sync_metadata_test.clj
```diff
@@ -236,3 +240,61 @@
...
+                   {:id             1
+                    :base-type      :type/Integer
+                    :effective-type :type/Integer}))))
+
+    (testing "and sync will re-fingerprint and analyze this field"
+      (mt/dataset (mt/dataset-definition "base-type-change"
+                    ["table"
+                     [{:field-name "field"
+                       :base-type  :type/Text}]
+                     [["ngoc@metabase.com"]]])
```
calherries, line 270: This dataset and later changes to the test db will remain after the test finishes, which means this test doesn't run in isolation. You should wrap this test in `mt/with-empty-db` to be completely safe

src/metabase/sync/sync_metadata/fields/sync_metadata.clj
```diff
@@ -79,11 +79,17 @@
...
                      old-base-type
                      new-base-type)
-           {:base_type new-base-type})
+           {:base_type           new-base-type
+            :effective_type      new-base-type
+            :coercion_strategy   nil
+            ;; reset fingerprint version so this field will get re-fingerprinted and analyzed
+            :fingerprint_version 0
+            :fingerprint         nil
+            :semantic_type       nil})
```
calherries, line 92: It's not clear why `:semantic_type` must be nil here. It took me a while to realise it was because of this:
https://github.com/metabase/metabase/blob/224b904a0ca5892134d83eda715a519eb27aa159/src/metabase/sync/analyze/fingerprint/fingerprinters.clj#L261

I think you should either document this here, or better, create a function that "resets" fingerprints and analysis that lives in an appropritate namespace under `metabase.sync.analyze`

calherries (approved):
LGTM, just left a couple of small suggestions

test/metabase/sync/sync_metadata/fields/sync_metadata_test.clj
qnkhuat: it was wrapped with a try-finally to delete the db afterward, so there should no new db after the test.

calherries: Sorry, my comment was ambiguous. I'm talking about changes to the actual test database that is created with `mt/dataset`, not the database data in the application db. The finally block only cleans up the database in the application db.

PR #37112, opened 2023-12-26
`with-premium-features` to `metabase.test`

qnkhuat:
Just a small clean up, move `premium-features-test/with-premium-features` to `mt/with-premium-features`


snoe (approved):
Great change.

PR #33771, opened 2023-09-06
Thread-safe `with-temporary-setting-values` and parallelize more QP tests

camsaul:
Resolves #33616

- Add new `with-temporary-setting-values` implementation that is thread-safe (binds a new dynamic var instead of updating the values globally) which speeds up things considerably regardless of whether tests are `^:parallel` or not since we don't need to hit the app DB and don't need to serialize and deserialize values 
- Remove restriction that `with-temporary-setting-values` is not allowed in `^:parallel` tests
- Add new `mt/test-helpers-set-global-values!` macro that toggles behavior for things like `with-temporary-setting-values`. When used inside a `test-helpers-set-global-values!` form, helpers like `with-temporary-setting-values` will set things globally (e.g. via `with-redefs`) rather than thread-locally (e.g. via `binding`)
- Mark another ~100 QP tests `^:parallel`. Lots of other tests can probably be made parallel as well, especially once #33738 lands, but I tried to resist the urge to mark a million tests parallel when that's not really the goal of this PR 
- Fix a few small bugs with Settings
  - `:double` Settings no longer ignore their value if you give them an integer value
  - if a value in the app DB/env var/etc. is invalid (the wrong type), Settings will correctly fall back to the default value instead of returning `nil`
- When JDBC connection pools are invalided, we now soft-reset the connection pool (leave existing connections open, but close them once they are returned) rather than completely destroying the pool; this improves parallelizability and speeds up the test suite a bit

```
% time clj -X:dev:test:ee:ee-dev:test/qp
...
Ran 946 tests in 69.168 seconds
....
Ran 783 tests in parallel, 163 single-threaded.
Finding and running tests took 1.4 mins.
All tests passed.

232.20s user 8.63s system 232% cpu 1:43.73 total
```

This is another ~100 or so `^:parallel` tests relative to #33429 

Compare this to 2-3 weeks ago, the QP test suite is 80% faster (almost a minute shaved off):

```
Ran 817 tests in 124.594 seconds
...
Ran 165 tests in parallel, 652 single-threaded.
Finding and running tests took 2.3 mins.
All tests passed.

229.65s user 7.42s system 156% cpu 2:31.37 total
```

PR #37115, opened 2023-12-27
Thread-safe `with-temporary-setting-values`

qnkhuat:
Most of the code here are from https://github.com/metabase/metabase/pull/33771, so all of the credit should be given to @camsaul 

Epic: https://github.com/metabase/metabase/issues/34362


qnkhuat:
Cam will continue this in his PR

PR #37124, opened 2023-12-27
Upgrade chromatic and sync storybook packages

uladzimirdev:
### Description

Upgrade chromatic and sync storybook packages to avoid duplications and to bypass CI issues like this https://github.com/metabase/metabase/actions/runs/7336443587/job/19975769824?pr=37013

Chromatic became broken after https://github.com/metabase/metabase/pull/37007 because of using different versions of storybook packages (both 6.5.15 and 6.5.16 were in packages.json)

### How to verify

- make sure chromatic deployment works
- `yarn build-storybook` should work
- `yarn storybook` should work

### Checklist


package.json
```diff
@@ -137,13 +137,12 @@
...
     "@storybook/addon-a11y": "^6.5.16",
-    "@storybook/addon-actions": "6.5.15",
-    "@storybook/addon-essentials": "6.5.15",
-    "@storybook/addon-links": "6.5.15",
+    "@storybook/addon-actions": "6.5.16",
+    "@storybook/addon-essentials": "6.5.16",
+    "@storybook/addon-links": "6.5.16",
     "@storybook/blocks": "^7.0.18",
-    "@storybook/builder-webpack5": "6.5.15",
-    "@storybook/client-api": "6.5.15",
```
uladzimirdev, line 145: client-api is removed as it's not used directly and it's a depedency of another package

uladzimirdev, line 145: `client-api` shouldn't be a direct dependency as it was used only to import `useArgs` and this function is a part of `addons` package

ranquild (approved):

PR #37126, opened 2023-12-27
RFC: kondo lint to enforce naming convention for thread not safe defn/defmacro in tests

qnkhuat:
This adds a lint to force `defn / defmacro` to have a name that ends with `!` if it contains not thread safe calls like `mt/with-temp!`.

Screenshot 2023-12-27 at 16 43 37

This is the first step to pursue this [RFC](https://metaboat.slack.com/archives/CKZEMT1MJ/p1692119906558039) from Cam. The goal here is to make identifying "parallelable" tests easier. To achieve that, we should adopt a convention name for any function / macro that cannot be used in parallel tests to have the name end with `!`.

Once we have this, it'll be easy to write a kondo to suggest whether or not a deftest can be paralleled.

This lint will only be used in test namespaces.

One more thing, the reason I enforce this lint on `defn` is because most of the time, we write macros in this pattern

```
(defn do-magic-happens-here
  [thunk]
  (thunk))

(defmacro magic-happens-here
  [& body]
  `(do-magic-happens-here (fn [] ~@body)))
```

so, to make sure we cover all macros, we should make all functions recognizable as safe / not safe.


.clj-kondo/config.edn
```diff
@@ -595,11 +595,11 @@
    metabase.related-test/with-world                                             macros.metabase.related-test/with-world
    metabase.shared.util.namespaces/import-fn                                    macros.metabase.shared.util.namespaces/import-fn
    metabase.test.data.users/with-group-for-user                                 macros.metabase.test.data.users/with-group-for-user
-   metabase.test.util/with-temp-env-var-value                                   macros.metabase.test.util/with-temp-env-var-value
```
qnkhuat, line 598: I changed this one macro as an example

camsaul (approved):
Big fan of this.

The main thing that stopped me from going any further with my previous PR was just that so many things needed to be fixed -- maybe we can have the linter ignore some of the test helpers that we know are missing `!` for the time being, and we can fix them in follow-on PRs. 

We should also make sure this respects `#_:clj-kondo/ignore` forms, because there are some legitimate cases where you might want to use `with-redefs` in something without marking it as `!`, for example functions that respect `test-helpers-set-global-values!`. Normally they would be thread-safe, but then they'd have a thread-unsafe behavior if used inside `test-helpers-set-global-values!`. That's ok tho because `test-helpers-set-global-values!` itself ends in a `!`

qnkhuat:
@camsaul yes, it does work if with `:clj-kondo/ignore`

PR #37133, opened 2023-12-27
Add delayResponse e2e helper

oleggromov:
This is handy when you need to set a delay until a request resolves.

The helper is used in https://github.com/metabase/metabase/pull/36910

### DIFF
--- a/e2e/support/helpers/e2e-request-helpers.js
+++ b/e2e/support/helpers/e2e-request-helpers.js
@@ -28,3 +28,18 @@ export function spyRequestFinished(name = "requestFinishedSpy") {
     spy,
   };
 }
+
+/**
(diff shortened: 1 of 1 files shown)


oleggromov:
All percy changes are minor, mostly slight changes in rendering (scale of charts, paddings in tables or some other small non-determinisms). @alxnddr please confirm this is ok though:

Screenshot 2023-12-27 at 15 33 42

oleggromov:
As suggested by @nemanjaglumac, switched over to using `res.setDelay`, which is built into Cypress. Not exactly the same (since we're now delaying responses instead of requests) but it doesn't matter and of course, works for me.

oleggromov:
After a discussion with Nemanja I think it shouldn't become a "global" helper yet. Will keep it within the test file/PR where it is used.

PR #37134, opened 2023-12-27
Fix card loading state in different tabs when parameters change

### LINKED ISSUE #33767: We're not signaling when a card is loading if you change the filters on a dashboard and switch tabs

oleggromov:
Fixes https://github.com/metabase/metabase/issues/33767

## Description
Unlike https://github.com/metabase/metabase/pull/36910, fixes the issue by reacting to change in _parameters_ in `fetchCardData`, which is the correct way to fix it.

### Before
https://github.com/metabase/metabase/assets/2196347/4180acc8-83f2-4074-8228-87f95a50814b 

### After

https://github.com/metabase/metabase/assets/2196347/1680c2b2-f39b-430e-9c9d-8a3dc55500ba

### DIFF
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -215,7 +215,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       cy.get("aside").findByText("No available targets").should("exist");
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
 
       cy.intercept(
         "GET",
@@ -266,7 +266,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addTextParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
 
       clickLineChartPoint();
       cy.findAllByTestId("field-set")
@@ -309,7 +309,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addTimeParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
 
       clickLineChartPoint();
       cy.findAllByTestId("field-set")
@@ -369,7 +369,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       addTextParameter();
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
 
       clickLineChartPoint();
       cy.findAllByTestId("field-set")
@@ -443,7 +443,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
         .should("not.exist");
       cy.button("Done").should("be.enabled").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
 
       clickLineChartPoint();
       cy.get("@targetDashboardId").then(targetDashboardId => {
@@ -486,7 +486,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
         .findByLabelText("Select a dashboard tab")
         .should("not.exist");
       cy.button("Done").should("be.enabled").click();
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
 
       clickLineChartPoint();
       cy.get("@targetDashboardId").then(targetDashboardId => {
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/dashboard/actions/data-fetching.js
```diff
@@ -345,7 +345,16 @@ export const fetchCardData = createThunkAction(
 
       cancelFetchCardData(card.id, dashcard.id);
 
-      if (clearCache) {
+      // When dashcard parameters change, we need to clean previous (stale)
+      // state so that the loader spinner shows as expected (#33767)
+      const hasParametersChanged =
+        !lastResult ||
```
oleggromov, line 351: @alxnddr do you think it should be this or `lastResult && equals(...)`?

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -410,3 +467,19 @@ describeWithSnowplow("scenarios > dashboard > tabs", () => {
...
+
+/**
+ * When you need to postpone a response (to check for loading spinners or alike),
+ * use this:
+ *
+ * `cy.intercept('POST', path, delayResponse(1000)).as('delayed')`
+ *
+ * `cy.wait('@delayed')` - you'll have 1000 ms until this resolves
+ */
+function delayResponse(delayMs) {
```
oleggromov, line 479: As discussed with @nemanjaglumac, we don't extract this as a global helper until needed by some other test.

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -351,6 +354,60 @@ describe("scenarios > dashboard > tabs", () => {
...
+    });
+
+    // Auto-connection happens here
+    selectDashboardFilter(getDashboardCard(0), "Created At");
+    saveDashboard();
+
+    cy.intercept(
+      "POST",
+      "/api/dashboard/*/dashcard/*/card/*/query",
+      delayResponse(500),
```
oleggromov, line 387: This is needed so that the test is deterministic and won't flake because of too fast responses.

e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
```diff
@@ -215,7 +215,7 @@ describe("scenarios > dashboard > dashboard cards > click behavior", () => {
       cy.get("aside").findByText("No available targets").should("exist");
       cy.get("aside").button("Done").click();
 
-      saveDashboard();
+      saveDashboard({ waitMs: 250 });
```
oleggromov, line 218: This is needed for [the same reason](https://github.com/metabase/metabase/pull/36910/files#r1437018691) it was needed before.

frontend/src/metabase/dashboard/actions/data-fetching.js
alxnddr: I think the equals part should definitely be `!equals(...)` since there should be difference. Regarding lastResult, I think we do not need to clear anything if there is no last result. But these are guesses based on logic which may not always work 😬

frontend/src/metabase/dashboard/actions/data-fetching.js
```diff
@@ -345,7 +345,16 @@ export const fetchCardData = createThunkAction(
 
       cancelFetchCardData(card.id, dashcard.id);
 
-      if (clearCache) {
+      // When dashcard parameters change, we need to clean previous (stale)
+      // state so that the loader spinner shows as expected (#33767)
```
alxnddr, line 349: 👍 for the explanation comment

alxnddr (approved):
This approach looks more reasonable to me, thanks for addressing my feedback! The timeouts in tests look not great, but I guess there is no simple solution to avoid them, right? cc @nemanjaglumac

oleggromov:
@alxnddr I haven't found a better way - and [neither did](https://github.com/metabase/metabase/pull/36910/files#r1437036987) @ranquild

ranquild (approved):

PR #37136, opened 2023-12-27
Filter modal hooks tests

ranquild:
Parametrized unit tests for shared hooks between `FilterPicker` and `FilterModal`.

### DIFF
--- a/frontend/src/metabase/querying/hooks/use-time-filter/use-time-filter.unit.spec.ts
+++ b/frontend/src/metabase/querying/hooks/use-time-filter/use-time-filter.unit.spec.ts
@@ -115,7 +115,6 @@ describe("useTimeFilter", () => {
 
       const { operator, values, getFilterClause } = result.current;
       const newFilter = checkNotNull(getFilterClause(operator, values));
-
       expect(
         Lib.displayInfo(defaultQuery, stageIndex, newFilter),
       ).toMatchObject({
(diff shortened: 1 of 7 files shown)


frontend/src/metabase/querying/hooks/use-time-filter/use-time-filter.unit.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+      const newOperator = ">";
+      setOperator(newOperator);
+      setValues(getDefaultValues(newOperator, values));
+    });
+
+    act(() => {
+      const { operator, values, getFilterClause } = result.current;
+      const newFilter = checkNotNull(getFilterClause(operator, values));
+
+      expect(
```
uladzimirdev, line 282: afaik `expect` should go outside `act` as there is no any side effect in it

frontend/src/metabase/querying/hooks/use-time-filter/use-time-filter.unit.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+          stageIndex,
+          column,
+        }),
+      );
+
+      act(() => {
+        const { getFilterClause } = result.current;
+        const newFilter = checkNotNull(getFilterClause(operator, values));
+
+        expect(
```
uladzimirdev, line 209: move expect outside `act`

frontend/src/metabase/querying/hooks/use-time-filter/use-time-filter.unit.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+      act(() => {
+        const { setValues } = result.current;
+        setValues(values);
+      });
+
+      act(() => {
+        const { operator, values, getFilterClause } = result.current;
+        const newFilter = checkNotNull(getFilterClause(operator, values));
+
+        expect(Lib.displayInfo(query, stageIndex, newFilter)).toMatchObject({
```
uladzimirdev, line 181: move `expect` outside act

uladzimirdev (approved):
LGTM, I'd only move `expect` from `act` everywhere

frontend/src/metabase/querying/hooks/use-time-filter/use-time-filter.unit.spec.ts
ranquild: Fixed

PR #37137, opened 2023-12-27
Deprecate TokenField

ranquild:


frontend/src/metabase/components/TokenField/TokenField.tsx
```diff
@@ -701,8 +701,13 @@ DefaultTokenFieldLayout.propTypes = {
   isFocused: PropTypes.bool,
 };
 
-// eslint-disable-next-line import/no-default-export -- deprecated usage
-export default Object.assign(TokenField, {
+/**
+ * @deprecated use MultiSelect or Autocomplete from metabase/ui
+ */
+const TokenField = Object.assign(_TokenField, {
```
uladzimirdev, line 707: I'd use named exports instead, so you woudln't need to use `_TokenField` weird naming

ranquild: I used it for compatibility so I won't need to update places that use this component. I would still need to use `_TokenField` or something similar because of `Object.assign` call

uladzimirdev: I meant to use `export const Token = Object.assign(...)`

I got it, there are 10 files to update. No need to refactor this atm 👍

uladzimirdev (approved):

PR #37141, opened 2023-12-28
Try again to fix failing perms tests by making testing macro more robust

noahmoss:
* Adds a new `metabase.permissions.test-util` namespace
* Adds a new `with-restored-perms!` macro which ensures that permissions are reset after the test body runs
* Updates `with-all-user-data-perms` to use `with-restored-perms!` under the hood, and renames it to `with-all-user-data-perms!`

This is my attempt to fix the state issues with permission tests in a more robust way. Generally, the `permissions` and `sandboxes` tables aren't too large, so it shouldn't be a ton of overhead to just fully reset these tables after a test runs—and certainly doesn't compare with the time lost to debugging state issues between perms tests. (I actually saw this test namespace run one second _faster_ with these changes, in a very unscientific test.)

I'm planning on adding more utils to the `metabase.permissions.test-util` namespace to make perms tests easier to write. But want to keep this PR focused on the change that will (hopefully) fix the tests failing on master.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
@@ -7,13 +7,15 @@
    [metabase.api.database :as api.database]
    [metabase.driver :as driver]
    [metabase.driver.sql-jdbc.connection :as sql-jdbc.conn]
-   [metabase.models :refer [Dashboard DashboardCard Database Field FieldValues
-                            Permissions Table]]
+   [metabase.models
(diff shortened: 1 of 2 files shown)


noahmoss:
* `master`
  * **PR #37141** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37141?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37141?utm_source=stack-comment).

dpsutton (approved):

PR #37152, opened 2023-12-28
Fix tabs ee test

oleggromov:
Fixes the test so it works on any day.

### DIFF
--- a/e2e/test/scenarios/dashboard/tabs.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/tabs.cy.spec.js
@@ -396,7 +396,7 @@ describe("scenarios > dashboard > tabs", () => {
     getDashboardCard(0).within(() => {
       cy.findByTestId("loading-spinner").should("exist");
       cy.wait("@saveCard");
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -396,16 +396,15 @@ describe("scenarios > dashboard > tabs", () => {
...
     });
 
     // Loader in the 1st tab
     goToTab("Tab 1");
     getDashboardCard(0).within(() => {
       cy.findByTestId("loading-spinner").should("exist");
       cy.wait("@saveCard");
-      cy.findByText("58.4") // this is total from the first row
-        .should("exist");
+      cy.findAllByTestId("table-row").should("exist");
```
oleggromov, line 407: Instead of checking for a particular value, which was no good anyways, we check for the existence of table rows.

nemanjaglumac (approved):
Ideally we should find something more concrete for the assertion, but for this particular case this seems enough.

I cannot imagine a scenario in which we render empty table.

PR #37155, opened 2023-12-28
Add code diff logic and switch to preview mode

deniskaber:
Part of a milestone 5 from https://github.com/metabase/metabase/issues/35961

### Description

This adds new logic needed for Static Embed settings modal - 

1. Highlight changed code on Parameters tab on parameters change.
2. Preserve selected Preview mode (Code / Preview) and code sample language across tabs.

Also:
* Added proper code highlight for all code examples (Ruby, Python, so on)

### Demo
https://www.loom.com/share/a60ddf0532b540feb433b0d810a589de

https://www.loom.com/share/1a7473a5cb84468e97ef74cee452fbb3

### Checklist


WiNloSt (commented):
Partial review part 1/x

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -14,87 +28,218 @@ const TABS = {
...
-  onChangeEmbeddingParameters,
-  onChangeParameterValue,
-  onChangePane,
+  onUpdateEnableEmbedding,
+  onUpdateEmbeddingParams,
 }: StaticEmbedSetupPaneProps): JSX.Element => {
+  const [activePane, setActivePane] = useState<ActivePreviewPane>("code");
+
+  const siteUrl = useSelector(state => getSetting(state, "site-url"));
+  const secretKey = checkNotNull(
```
WiNloSt, line 52: I saw that these 2 props are being passed through a bunch of components, what about we declare these states at the leaf component where they're used? That should probably help with the props drilling. What do yuo think?

WiNloSt:
I don't think I've tested this in the past, but choosing a different code language doesn't persist through different tabs. If you choose another language in any tab e.g. Ruby then go to a different tab, the language setting is lost. This is critical to be fixed.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/EmbedCodePane.tsx
```diff
@@ -0,0 +1,175 @@
...
+        title={t`Then insert this code snippet in your HTML template or single page app.`}
+        selectedOptionName={selectedClientCodeOptionName}
+        languageOptions={clientCodeOptions.map(({ name }) => name)}
+        source={selectedClientCodeOption.source}
+        textHighlightMode={selectedClientCodeOption.mode}
+        onChangeOption={setSelectedClientCodeOptionName}
+      />
+
+      {withExamplesLink && (
+        <div className="text-centered mb2 mt2">
```
WiNloSt, line 162: Is it possible to use Mantine components instead of using global classes? It seems we're trying to avoid using global class as much as possible as [we've planned to migrate it.](https://www.notion.so/metabase/83-Prepare-global-css-for-migration-to-css-modules-4ab29b8589b545e0960bf032d24a11a9)

WiNloSt (commented):
Looks great so far, I think there are 1 major issue and 1 minor issue we should fix before merging this PR.
Major: The language selection doesn't persist through different tabs
Minor: The copy titles don't match the design. They are all updated to follow the pattern of `[verb + ing] noun` e.g. **Previewing locked parameters** instead of **Preview locked parameters**

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -1,100 +1,301 @@
...
+
+  const handleSave = async () => {
+    try {
+      if (!resource.enable_embedding) {
+        await onUpdateEnableEmbedding(true);
+      }
+      await onUpdateEmbeddingParams(embeddingParams);
+    } catch (e) {
+      console.error(e);
+      throw e;
```
npretto, line 124: Why are we catching the error to then throw it again as is?

frontend/src/metabase/public/lib/code-templates.ts
```diff
@@ -0,0 +1,179 @@
...
+  ${ruby.getParametersSource(params)}
+  :exp => Time.now.to_i + (60 * 10) # 10 minute expiration
+}
+token = JWT.encode payload, METABASE_SECRET_KEY
+
+iframe_url = METABASE_SITE_URL + "/embed/${resourceType}/" + token +
+  ${ruby.getIframeQuerySource(displayOptions)}`,
+};
+
+export const closure = {
```
npretto, line 121: Should this be `clojure`?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.unit.spec.tsx
```diff
@@ -0,0 +1,839 @@
...
+  value,
+  highlightedText,
+}: {
+  value: string;
+  highlightedText?: string[];
+}) => (
+  <>
+    <div data-testid="text-editor-mock">{value}</div>
+    <div data-testid="text-editor-mock-highlighted-code">
+      {highlightedText?.map(text => text)}
```
npretto, line 26: nit: isn't this the same as just `{highlightedText?}`

albertoperdomo (approved):

vbenedetti (approved):

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/tests/common.unit.spec.tsx
```diff
@@ -0,0 +1,711 @@
...
+          name: "documentation",
+        });
+        expect(link).toBeVisible();
+        expect(link).toHaveAttribute(
+          "href",
+          "https://www.metabase.com/docs/latest/embedding/static-embedding.html?utm_source=oss&utm_media=static-embed-settings-overview",
+        );
+
+        if (resourceType === "dashboard") {
+          // eslint-disable-next-line jest/no-conditional-expect
```
npretto, line 93: I'm not really a fan of going against this rule, but the file is probably already to big as it is so let's see what others have to say

frontend/src/metabase/public/lib/embed.ts
```diff
@@ -1,16 +1,21 @@
...
-  secretKey,
-  previewEmbeddingParams,
+  resourceType: EmbedResourceType,
+  resourceId: EmbedResource["id"],
+  params: EmbeddingParameters = {},
+  secretKey: string,
+  previewEmbeddingParams: EmbeddingParameters,
 ) {
-  const unsignedToken = {
+  const unsignedToken: { [key: string]: any } = {
```
oisincoveney, line 18: Let's call this 
```
Record<string, any`
```
for consistency with other types that use object types

frontend/src/metabase/public/lib/embed.ts
```diff
@@ -45,18 +50,7 @@ export function getSignedPreviewUrl(
...
-  resourceId,
-  options,
-) {
-  return `${siteUrl}/public/${resourceType}/${resourceId}${optionsToHashParams(
-    options,
-  )}`;
-}
-
-export function optionsToHashParams(options = {}) {
+export function optionsToHashParams(options: { [key: string]: any } = {}) {
```
oisincoveney, line 53: same here

frontend/src/metabase/public/lib/types.ts
```diff
@@ -36,3 +30,29 @@ export type EmbeddingDisplayOptions = {
...
+  displayOptions: EmbeddingDisplayOptions;
+};
+
+export type ClientCodeSampleConfig = {
+  name: string;
+  source: string;
+  mode: string;
+};
+
+export type ServerCodeSampleConfig = {
```
oisincoveney, line 49: Can we be more specific than `string` here for some of these?

frontend/src/metabase/public/lib/code-templates.ts
```diff
@@ -0,0 +1,179 @@
+import type {
+  CodeSampleParameters,
+  EmbeddingDisplayOptions,
+  EmbeddingParameters,
+} from "./types";
+import { optionsToHashParams } from "./embed";
+
+export const node = {
```
oisincoveney, line 8: Let's move the `node` variable next to `python` on line 40

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/tests/common.unit.spec.tsx
oisincoveney: I agree. I think if we have to have conditionals within an `each` clause in a test, then we probably should split it out into somewhere else

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -1,100 +1,322 @@
...
+          <Tabs.Tab value={TABS.Overview}>{t`Overview`}</Tabs.Tab>
+          <Tabs.Tab value={TABS.Parameters}>{t`Parameters`}</Tabs.Tab>
+          <Tabs.Tab value={TABS.Appearance}>{t`Appearance`}</Tabs.Tab>
+        </Tabs.List>
+        <Tabs.Panel value={TABS.Overview}>
+          <OverviewSettings
+            resourceType={resourceType}
+            selectedServerCodeOption={selectedServerCodeOption}
+            serverEmbedCodeSlot={
+              <ServerEmbedCodePane
```
oisincoveney, line 153: Can we put this code into a variable/function, something like this?

```

const getEmbedCodePane = (variant: "appearance" | "overview" | "parameters") => (
	<ServerEmbedCodePane
                className="flex-full w-full"
                variant={variant}
                initialPreviewParameters={initialPreviewParameters}
                resource={resource}
                resourceType={resourceType}
                siteUrl={siteUrl}
                secretKey={secretKey}
                params={previewParametersBySlug}
                displayOptions={displayOptions}
                serverCodeOptions={serverCodeOptions}
                selectedServerCodeOptionName={selectedServerCodeOptionName}
                setSelectedServerCodeOptionName={
                  setSelectedServerCodeOptionName
                }
              />
)

```

I ran the `serverEmbedCodeSlot` props into a diff checker and `variant` is the only thing I saw was different

oisincoveney: or use `TABS.Apperance` etc etc as the parameter value

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.styled.tsx
```diff
@@ -17,12 +17,16 @@ const SettingsAsideBlock = styled.div`
   padding: 2rem;
   border-right: 1px solid ${color("border")};
   background-color: ${color("white")};
+
+  height: 45.125rem;
```
oisincoveney, line 21: Where does this value come from? It's pretty random to me - even if we convert it to pixel values we get `722px` which seems like a really random number

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ServerEmbedCodePane.tsx
```diff
@@ -0,0 +1,186 @@
...
+      siteUrl,
+      secretKey,
+      resourceType,
+      resource,
+      displayOptions,
+    });
+
+  if (variant === "overview") {
+    return (
+      <CodeSample
```
oisincoveney, line 75: Can we put the `CodeSample/title` values into a function and only call `CodeSample` once? Something like

```
const getTitleText = (variant) => {
	... put the three different titles here ...
}


// more things here

<CodeSample
        dataTestId="embed-backend"
        className={className}
        title={getTitleText()}
        selectedOptionName={selectedServerCodeOptionName}
        languageOptions={serverCodeOptions.map(({ name }) => name)}
        source={selectedServerCodeOption.source}
        textHighlightMode={selectedServerCodeOption.mode}
        highlightedText={highlightedText}
        onChangeOption={setSelectedServerCodeOptionName}
      />
``` 

since all of these calls seem the same

frontend/src/metabase/public/components/EmbedModal/EmbedModal.tsx
```diff
@@ -70,29 +50,20 @@ export const EmbedModal = ({ children, isOpen, onClose }: EmbedModalProps) => {
...
-      title={<EmbedTitle {...modalHeaderProps} />}
+      title={isEmbeddingSetupStage ? t`Embed ${applicationName}` : undefined}
       fit
       formModal={false}
     >
-      <Divider />
-      <Box bg="bg.0" h="100%">
-        {children({ embedType, goToNextStep, goBackToEmbedModal })}
-      </Box>
+      {!isEmbeddingSetupStage && (
```
oisincoveney, line 61: Can we keep the `EmbedTitle` here please? We've been back and forth on changing this between the milestones but it doesn't make sense to have the title either in the Modal or in the `EmbedModalHeader` considering the only thing that's changing is the string.

frontend/src/metabase/components/TextEditor/TextEditor.jsx
```diff
@@ -55,6 +62,13 @@ export default class TextEditor extends Component {
     this._editor.setReadOnly(this.props.readOnly);
     element.classList[this.props.readOnly ? "add" : "remove"]("read-only");
 
+    // highlightedText
```
oisincoveney, line 65: let's remove this lol

frontend/src/metabase/components/ModalContent/ModalHeader.tsx
oisincoveney, line 1: These changes really should be in separate smaller PRs. It'll make merges much easier.

frontend/src/metabase/components/ModalContent/ModalContent.styled.ts
```diff
@@ -41,3 +50,31 @@ export const ModalContentActionIcon = styled(Icon)`
...
   }
 `;
+
+export const ModalHeaderBackIcon = styled(ModalContentActionIcon)`
+  flex-shrink: 0;
+
+  margin: -0.5rem 0 -0.5rem -0.5rem;
+
+  :hover {
+    color: ${color("brand")};
```
oisincoveney, line 60: Mantine colors

frontend/src/metabase/public/lib/code.ts
oisincoveney, line 1: Since these are conversions that are independent of goal that we're trying to accomplish in this PR, we really should be putting these in separate smaller PRs.

package.json
oisincoveney, line 1: If we could split out the package upgrade + the changes accompanied with the package upgrade into a separate PR it would be amazing

oisincoveney (changes requested):
Let's reduce some of the copied code and a few other things. Again, let's try and have some smaller PRs please, especially related to the things that are independent of the PR's goals

oisincoveney (changes requested):
Let's reduce some of the copied code and a few other things. Again, let's try and have some smaller PRs please, especially related to the things that are independent of the PR's goals

oisincoveney (changes requested):
Let's reduce some of the copied code and a few other things. Again, let's try and have some smaller PRs please, especially related to the things that are independent of the PR's goals

frontend/src/metabase/public/lib/code.ts
```diff
@@ -0,0 +1,86 @@
...
+): ServerCodeSampleConfig[] => [
+  {
+    name: "Node.js",
+    source: node.getServerSource(codeSampleParameters),
+    parametersSource: node.getParametersSource(codeSampleParameters.params),
+    getIframeQuerySource: node.getIframeQuerySource(
+      codeSampleParameters.displayOptions,
+    ),
+    mode: "ace/mode/javascript",
+    embedOption: "Pug / Jade",
```
WiNloSt, line 51: Are all these options for the given server language finalized? I kind of remember the previous behavior, we only change the template language to ERB for Ruby, but not others? I could be wrong though. Could you double-check this?

WiNloSt (commented):
Haven't reviewed the test fully, but I've left some feedback that could be actionable.

frontend/src/metabase/public/lib/code-templates.ts
oisincoveney: I'm dumb, I mixed up the code within the string with the actual code. Sorry

oisincoveney (approved):
Let's get this in but let's have another person approve too. I think some of the little comments should be resolved as well

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/AppearanceSettings.tsx
```diff
@@ -81,15 +86,19 @@ export const AppearanceSettings = ({
           <StaticEmbedSetupPaneSettingsContentSection
             title={t`Customizing your embed’s appearance`}
           >
-            <Text>{jt`These cosmetic options requiring changing the server code. You can play around with and preview the options here, and check out the ${(
-              <ExternalLink
-                key="doc"
-                href={docsUrl}
-              >{t`documentation`}</ExternalLink>
-            )} for more.`}</Text>
+            <Text>{jt`These cosmetic options requiring changing the server code. You can play around with and preview the options here, and check out the ${
```
WiNloSt, line 89: I think the copy has already been updated. I couldn't find the exact doc how you're supposed to work with that. But the gist is for the copy that contains links to the documentation to be as isolated as possible, so we could hide the entire sentence. You can see that from the screenshot below that we've split `"Check out the documentation for more." into a separate sentence, so we could hide the entire sentence when we're not supposed to show the doc link as opposed to just remove the link and leave all the copy the same.

image

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/OverviewSettings.tsx
```diff
@@ -73,7 +68,14 @@ export const OverviewSettings = ({
               <br />
             </>
           )}
-          <Text>{jt`Check out the ${staticEmbedDocsLink} for more.`}</Text>
+          {showMetabaseLinks && (
```
WiNloSt, line 71: This link is correctly hidden, because we hide the entire sentence.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/OverviewSettings.tsx
```diff
@@ -86,18 +88,20 @@ export const OverviewSettings = ({
...
-                <ExternalLink
-                  key="examples"
-                  href="https://github.com/metabase/embedding-reference-apps"
-                >
-                  {t`examples on GitHub`}
-                </ExternalLink>
-              )}`}</h4>
-            </Center>
-          </Box>
+          {showMetabaseLinks && (
```
WiNloSt, line 91: I'm not actually sure about hiding the example links as in the project of hiding Metabase doc links, we only talk about the `/learn/*` and `/docs/*` links.

I asked a question [in Slack here.](https://metaboat.slack.com/archives/C063Q3F1HPF/p1706080479826559)

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.styled.tsx
WiNloSt: nit: I kind of get the same question. Like do `45rem` or `46rem` still work? Because those numbers look a lot better. I'd prefer 45rem since it also rounds to 720px for aesthetic.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/tests/premium.unit.spec.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+      it("should support hiding metabase links feature", () => {
+        setup({
+          props: {
+            resourceType,
+          },
+          activeTab: "Overview",
+          hasEnterprisePlugins: true,
+          tokenFeatures: createMockTokenFeatures({ whitelabel: true }),
+          showMetabaseLinks: false,
```
WiNloSt, line 47: Can you add the case for `showMetabaseLinks: false` in the other 2 files `common.*` and `enterprise.*` to ensure the links are still visible in those cases?

I know the tests look a bit verbose, but to be fair, we don't have a ton of places that link to our doc. IIRC when I was working on the project there were only 16 locations I need to hide.

WiNloSt (approved):
Approved, but we should fix the links and add a few more cases for `show-metabase-links: false` for `common.*`and `enterprise.*` to ensure those links are still showing.

PR #37159, opened 2023-12-28
Wrap `can-append-test` with mt/with-empty-db to fix seemingly unrelated download perms test

noahmoss:
This should fix the last failures in permission-related tests in Cloverage, which I had previously tried to fix in https://github.com/metabase/metabase/pull/37093 and https://github.com/metabase/metabase/pull/37141

Basically the issue was that `metabase.upload-test/can-append-test` was creating new tables in the test database to test CSV uploads. Most tests that did this were already wrapped in `mt/with-empty-db` so that the test database would be reset to its normal state afterwards, but this one didn't.

This caused a download perms test to fail because native query download perms for a DB are essentially derived from the download perms for all of its tables. If _any_ table has no download perms, then there are no download perms for native queries.

So whenever `can-append-test` was run _before_ the download perm tests, extra tables would be created in the test database, which would then cause native download perms for the sample database to be computed to `:none` when they should have been `:limited`, because the extra tables did not have download perms.

Apparently this order of test runs was only encountered when running Cloverage, not during normal test suites, so it only showed up on `master`. 🫠 

I've fixed it by adding `mt/with-empty-db` to `can-append-test`, like all the others in that test namespace. I've also:
* Switched `with-download-perms` to use `perms.test-util/with-restored-perms` from my earlier PR
* Did a couple formatting fixes

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/query_processor/middleware/permissions_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/query_processor/middleware/permissions_test.clj
@@ -7,8 +7,8 @@
    [metabase-enterprise.advanced-permissions.query-processor.middleware.permissions
     :as ee.qp.perms]
    [metabase.api.dataset :as api.dataset]
-   [metabase.models.permissions :as perms]
    [metabase.models.permissions-group :as perms-group]
+   [metabase.permissions.test-util :as perms.test-util]
    [metabase.public-settings.premium-features-test
     :as premium-features-test]
    [metabase.query-processor.reducible :as qp.reducible]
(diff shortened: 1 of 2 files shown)


noahmoss:
* `master`
  * **PR #37159** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37159?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37159?utm_source=stack-comment).

dpsutton (approved):

PR #37160, opened 2023-12-29
Unify type of timestamp columns

### LINKED ISSUE #33577: Unify type of timestamp columns
timestamp columns like `created_at` and `updated_at` should have type `timestamp with time zone`.
Currently it's not the case between tables, and between dbms. We should unify them

Below are the list of columns that have type `timestamp without time zone` in each db engine.

### h2
```
sql> select table_name, column_name, column_default, is_nullable from information_schema.columns where data_type = 'TIMESTAMP' order by table_name asc;
TABLE_NAME                           | COLUMN_NAME   | COLUMN_DEFAULT | IS_NULLABLE
ACTIVITY                             | TIMESTAMP     | null           | NO
APPLICATION_PERMISSIONS_REVISION     | CREATED_AT    | null           | NO
COLLECTION_PERMISSION_GRAPH_REVISION | CREATED_AT    | null           | NO
CORE_SESSION                         | CREATED_AT    | null           | NO
CORE_USER                            | DATE_JOINED   | null           | NO
CORE_USER                            | LAST_LOGIN    | null           | YES
CORE_USER                            | UPDATED_AT    | null           | YES
DATABASECHANGELOG                    | DATEEXECUTED  | null           | NO
DATABASECHANGELOGLOCK                | LOCKGRANTED   | null           | YES
DEPENDENCY                           | CREATED_AT    | null           | NO
DIMENSION                            | CREATED_AT    | null           | NO
DIMENSION                            | UPDATED_AT    | null           | NO
METABASE_DATABASE                    | CREATED_AT    | LOCALTIMESTAMP | NO
METABASE_DATABASE                    | UPDATED_AT    | LOCALTIMESTAMP | NO
METABASE_FIELD                       | CREATED_AT    | null           | NO
METABASE_FIELD                       | UPDATED_AT    | null           | NO
METABASE_FIELD                       | LAST_ANALYZED | null           | YES
METABASE_FIELDVALUES                 | CREATED_AT    | null           | NO
METABASE_TABLE                       | CREATED_AT    | null           | NO
METABASE_TABLE                       | UPDATED_AT    | null           | NO
METRIC                               | CREATED_AT    | null           | NO
METRIC                               | UPDATED_AT    | null           | NO
PERMISSIONS_REVISION                 | CREATED_AT    | null           | NO
PULSE                                | CREATED_AT    | null           | NO
PULSE                                | UPDATED_AT    | null           | NO
PULSE_CHANNEL                        | CREATED_AT    | null           | NO
PULSE_CHANNEL                        | UPDATED_AT    | null           | NO
RECENT_VIEWS                         | TIMESTAMP     | null           | NO
REPORT_CARD                          | CREATED_AT    | null           | NO
REPORT_CARDFAVORITE                  | CREATED_AT    | null           | NO
REPORT_CARDFAVORITE                  | UPDATED_AT    | null           | NO
REPORT_DASHBOARD                     | CREATED_AT    | null           | NO
REPORT_DASHBOARD                     | UPDATED_AT    | null           | NO
REPORT_DASHBOARDCARD                 | CREATED_AT    | LOCALTIMESTAMP | NO
REPORT_DASHBOARDCARD                 | UPDATED_AT    | LOCALTIMESTAMP | NO
SEGMENT                              | CREATED_AT    | null           | NO
SEGMENT                              | UPDATED_AT    | null           | NO
V_ALERTS                             | CREATED_AT    | null           | YES
V_ALERTS                             | UPDATED_AT    | null           | YES
V_DASHBOARDCARD                      | CREATED_AT    | null           | YES
V_DASHBOARDCARD                      | UPDATED_AT    | null           | YES
V_DATABASES                          | CREATED_AT    | null           | YES
V_DATABASES                          | UPDATED_AT    | null           | YES
V_FIELDS                             | CREATED_AT    | null           | YES
V_FIELDS                             | UPDATED_AT    | null           | YES
V_SUBSCRIPTIONS                      | CREATED_AT    | null           | YES
V_SUBSCRIPTIONS                      | UPDATED_AT    | null           | YES
V_TABLES                             | CREATED_AT    | null           | YES
V_TABLES                             | UPDATED_AT    | null           | YES
V_USERS                              | DATE_JOINED   | null           | YES
V_USERS                              | LAST_LOGIN    | null           | YES
V_USERS                              | UPDATED_AT    | null           | YES
(52 rows, 27 ms)
```

### MySQL

```
mysql> select table_name, column_name, column_default, is_nullable  from information_schema.columns where data_type = 'datetime' and TABLE_SCHEMA = 'metabase';
+--------------------------------------+---------------+----------------------------+-------------+
| TABLE_NAME                           | COLUMN_NAME   | COLUMN_DEFAULT             | IS_NULLABLE |
+--------------------------------------+---------------+----------------------------+-------------+
| activity                             | timestamp     | NULL                       | NO          |
| application_permissions_revision     | created_at    | NULL                       | NO          |
| collection_permission_graph_revision | created_at    | NULL                       | NO          |
| core_session                         | created_at    | NULL                       | NO          |
| core_user                            | date_joined   | NULL                       | NO          |
| core_user                            | last_login    | NULL                       | YES         |
| core_user                            | updated_at    | NULL                       | YES         |
| databasechangelog                    | DATEEXECUTED  | NULL                       | NO          |
| databasechangeloglock                | LOCKGRANTED   | NULL                       | YES         |
| dependency                           | created_at    | NULL                       | NO          |
| dimension                            | created_at    | NULL                       | NO          |
| dimension                            | updated_at    | NULL                       | NO          |
| metabase_field                       | created_at    | NULL                       | NO          |
| metabase_field                       | last_analyzed | NULL                       | YES         |
| metabase_field                       | updated_at    | NULL                       | NO          |
| metabase_fieldvalues                 | created_at    | NULL                       | NO          |
| metabase_table                       | created_at    | NULL                       | NO          |
| metabase_table                       | updated_at    | NULL                       | NO          |
| metric                               | created_at    | NULL                       | NO          |
| metric                               | updated_at    | NULL                       | NO          |
| permissions_revision                 | created_at    | NULL                       | NO          |
| pulse                                | created_at    | NULL                       | NO          |
| pulse                                | updated_at    | NULL                       | NO          |
| pulse_channel                        | created_at    | NULL                       | NO          |
| pulse_channel                        | updated_at    | NULL                       | NO          |
| recent_views                         | timestamp     | NULL                       | NO          |
| report_card                          | created_at    | NULL                       | NO          |
| report_cardfavorite                  | created_at    | NULL                       | NO          |
| report_cardfavorite                  | updated_at    | NULL                       | NO          |
| report_dashboard                     | created_at    | NULL                       | NO          |
| report_dashboard                     | updated_at    | NULL                       | NO          |
| segment                              | created_at    | NULL                       | NO          |
| segment                              | updated_at    | NULL                       | NO          |
| v_alerts                             | created_at    | NULL                       | NO          |
| v_alerts                             | updated_at    | NULL                       | NO          |
| v_content                            | created_at    | 0000-00-00 00:00:00.000000 | NO          |
| v_content                            | updated_at    | NULL                       | YES         |
| v_fields                             | created_at    | NULL                       | NO          |
| v_fields                             | updated_at    | NULL                       | NO          |
| v_subscriptions                      | created_at    | NULL                       | NO          |
| v_subscriptions                      | updated_at    | NULL                       | NO          |
| v_tables                             | created_at    | NULL                       | NO          |
| v_tables                             | updated_at    | NULL                       | NO          |
| v_users                              | date_joined   | NULL                       | NO          |
| v_users                              | last_login    | NULL                       | YES         |
| v_users                              | updated_at    | NULL                       | YES         |
+--------------------------------------+---------------+----------------------------+-------------+
46 rows in set (0.01 sec)
```

### Postgres
```
metabase=# SELECT TABLE_NAME, COLUMN_NAME, COLUMN_DEFAULT, IS_NULLABLE FROM INFORMATION_SCHEMA.COLUMNS where data_type like 'timestamp without time zone' and table_catalog = 'metabase' and table_schema = 'public' order by table_name;
              table_name              | column_name  | column_default | is_nullable
--------------------------------------+--------------+----------------+-------------
 application_permissions_revision     | created_at   |                | NO
 collection_permission_graph_revision | created_at   |                | NO
 core_user                            | updated_at   |                | YES
 databasechangelog                    | dateexecuted |                | NO
 databasechangeloglock                | lockgranted  |                | YES
 dimension                            | updated_at   |                | NO
 dimension                            | created_at   |                | NO
 permissions_revision                 | created_at   |                | NO
 recent_views                         | timestamp    |                | NO
 v_users                              | updated_at   |                | YES
(10 rows)
(10 rows)
```

qnkhuat:
Closes https://github.com/metabase/metabase/issues/33577

Changing all the columns that currently use type `'timestamp without time zone` to timestamp with time zone`, for list of affected columns, see the issue.

The reason I used custom migration to do this instead of relying on the liquibase changeset is that `preConditions` in liquibase do not apply on rollback, and the columns that need to be updated for each DBMS are not the same. So, there is no way to write a proper rollback to revert type change for each DBMS.


calherries (changes requested):
The change does not break behaviour as long as the timestamp values are calculated as local timestamps for the database time zone, e.g. with `now()`, and not local timestamps for the reporting or JVM time zone. As far as I can tell that is the case for each column, so this change is not breaking.

Since new columns keep being added with inconsistent types, I think we should add a linter rule to prevent this happening again. See the [`no-bare-boolean-types?`](https://github.com/metabase/metabase/pull/36964/files#diff-39d8a2bb95c75e8d3d7adee09ce4968592b41fc16b24bf9cbb62e4a98f212103R99-R110) function that was recently added by Noah for an example.

Some tests for MariaDB are failing though, can you check why? Otherwise this looks good.

metamben (approved):

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -61,9 +59,10 @@
 
       ;; a modifyDataType change; see if it changes a column to target-type
       (:modifyDataType x)
-      (if (= target-types (str/lower-case (or (get-in x [:modifyDataType :newDataType]) "")))
-        (do (swap! found-cols conj x)
-            x)
+      (if (contains? target-types (str/lower-case (or (get-in x [:modifyDataType :newDataType]) "")))
+        (do
+         (swap! found-cols conj x)
+         x)
```
crisptrutski, line 65: small nit: If I'm reading this right, I think you can reorganize this expression a bit to avoid repeating the result value.

```clojure
(do 
  (when (contains? target-types (str/lower-case (or (get-in x [:modifyDataType :newDataType]) "")))
    (swap! found-cols conj x))
  x
)
```

I'm also not sure if you're intentionally guarding against falsey`newDataType` values, but perhaps the `or` could be replaced by passing a default value to `get-in`?

crisptrutski: Actually, it looks like *every* case in the cond is returning x? Perhaps the `do` can move around that, so the cond *only* does side-effects.

crisptrutski: I there are two places where this `target-types` check is done, so it could be nice to break out a helper function to encapsulate the lowercase and empty string logic?

```clojure
(defn- target-type? [target-types raw-type]
  (contains? target-types (str/lower-case (or raw-type ""))))
```

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -45,8 +44,7 @@
   TODO: add and conform to a spec instead?"
   [target-types found-cols x]
   {:pre [(set? target-types) (instance? clojure.lang.Atom found-cols)]}
-  (if
-    (map? x)
+  (if (map? x)
```
piranha, line 47: I think `(if-not (map? x) x ...)` will be a little bit easier on eyes

bin/lint-migrations-file/src/lint_migrations_file.clj
piranha: > Actually, it looks like every case in the cond is returning x? Perhaps the do can move around that, so the cond only does side-effects.

Oh, maybe even like that:

```
(when (map? x)
  (cond ...))
x
```

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -77,15 +76,11 @@
...
-    (doseq [{{id :id} :changeSet, :as change-set} change-log
-            :when                                 (and id
-                                                       (string? id))]
-      [id change-set])
     (doseq [{{id :id} :changeSet :as change-set} change-log
             :when                                (and id
                                                       (string? id)
                                                       (str/starts-with? id "v"))]
-      (walk/postwalk walk-fn change-set))
+      (doall (map walk-fn (get-in change-set [:changeSet :changes]))))
```
piranha, line 83: `run!`?

qnkhuat: duh! d441a1ebc2

bin/lint-migrations-file/src/lint_migrations_file.clj
qnkhuat: thanks for the recommendations, I incorporated them in d441a1ebc2

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -44,30 +44,25 @@
...
+             cols   (filter (fn [col-def]
+                              (match-target-types? (get-in col-def [:column :type] "")))
+                            (get-in x [op :columns]))]
+         (doseq [col cols]
+           (swap! found-cols conj col)))
+
+
+       ;; a modifyDataType change; see if it changes a column to target-type
+       (:modifyDataType x)
+       (when (match-target-types? (str/lower-case (get-in x [:modifyDataType :newDataType] "")))
```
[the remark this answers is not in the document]
qnkhuat: it's better :) 4e64a4b200

qnkhuat:
@calherries mariadb was failing because by default a required timestamp column will have extra `on update current_timestamp` (tho I can only see this on CI, my local mariadb does not have this weird behavior :) ).

That'll be a bug for `created_at` columns because we don't want it to update. See c701e0afd2ea90b86df08e227fdb1261ff223140 for the test that demonstrates it. The fix is in https://github.com/metabase/metabase/pull/37160/commits/bf1dc0e38c419371d140718681aaaae62ac97ab1

src/metabase/db/custom_migrations.clj
```diff
@@ -889,3 +900,127 @@
...
+      ;; so we need to drop the view before changing the type, then re-create it again
+      (when (= [db-type table column]
+               [:postgres :core_user :updated_at])
+        (t2/query [(format "DROP VIEW IF EXISTS v_users;")]))
+      (t2/query [(alter-table-column-type-sql db-type (name table) (name column) target-type nullable?)])
+      (when (= [db-type table column]
+               [:postgres :core_user :updated_at])
+        (t2/query [(-> (io/resource "migrations/instance_analytics_views/users/v1/postgres-users.sql")
+                       slurp
+                       (str/replace #"\n" " "))])))))
```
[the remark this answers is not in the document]
qnkhuat: it doesn't , but I thought it does :D, removed in https://github.com/metabase/metabase/pull/37160/commits/df4b3e50b26e2b329839545412004adaf5b8a00c

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -45,30 +44,25 @@
...
+      (cond
+       ;; a createTable or addColumn change; see if it adds a target-type col
+       (or (:createTable x) (:addColumn x))
+       (let [op     (cond (:createTable x) :createTable (:addColumn x) :addColumn)
+             cols   (filter (fn [col-def]
+                              (match-target-types? (get-in col-def [:column :type] "")))
+                            (get-in x [op :columns]))]
+         (doseq [col cols]
+           (swap! found-cols conj col)))
+
```
calherries, line 59:
```suggestion
```

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -77,15 +71,11 @@
   [change-log]
   (let [problem-cols (atom [])
         walk-fn      (partial assert-no-types-in-change-set #{"blob" "text"} problem-cols)]
```
calherries, line 73: Removing the postwalk without also removing this state is leaving behind half-refactored code. If you're going to refactor `assert-no-types-in-change-set` you can probably go all the way and remove the need for this `problem-cols` atom. The use of `assert-no-types-in-change-set` can be simplified to a `filter` on changeset maps. It's a pure operation and doesn't need external state.

calherries (changes requested):

calherries:
@qnkhuat The change you made makes sense to me as explained, but I still see some errors in CI

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -35,75 +35,63 @@
...
   "Ensures that no \"text\" or \"blob\" type columns are added in changesets with id later than 320 (i.e. version
   0.42.0).  From that point on, \"${text.type}\" should be used instead, so that MySQL can handle it correctly (by using
   `LONGTEXT`).  And similarly, from an earlier point, \"${blob.type}\" should be used instead of \"blob\"."
   [change-log]
-  (let [problem-cols (atom [])
-        walk-fn      (partial assert-no-types-in-change-set #{"blob" "text"} problem-cols)]
-    (doseq [{{id :id} :changeSet :as change-set} change-log
-            :when                                (and id
-                                                      (string? id)
-                                                      (str/starts-with? id "v"))]
```
calherries, line 87: Can you explain why you removed the `(str/starts-with? id "v")` check?

bin/lint-migrations-file/src/lint_migrations_file.clj
```diff
@@ -36,70 +35,63 @@
...
+     (and (:modifyDataType change)
+          (match-target-types? (get-in change [:modifyDataType :newDataType] "")))
+     true)))
+
+(defn- check-change-set-use-types?
+  "Return true if `change-set` doesn't contain usage of any type in `types`."
+  [target-types change-set]
+  (some #(check-change-use-types? target-types %) (get-in change-set [:changeSet :changes])))
+
+(defn- check-change-log-use-types?
```
calherries, line 62: IMO we've lost some readability here because this function doesn't have a docstring. `assert-no-types-in-change-set` had a pretty good explanation of what was happening.

This function is used in a few places and its implementation is complex enough to warrant a docstring. Something like: 
```
"Iterates over the change-set maps in the change-log and returns a list of problematic changes. A change is problematic if it adds/changes any columns to be of a type in `target-types `."
```

bin/lint-migrations-file/src/lint_migrations_file.clj
qnkhuat: all the migrations in `001_update_migrations.yaml` now start with `v`; it was needed when we still had the integer ID in the same migration file.

bin/lint-migrations-file/src/lint_migrations_file.clj
calherries: Also, I noticed that everywhere this function is used, it is wrapped in a `not`. Perhaps the `not` can be added to the body of this function and it could be renamed to `assert-no-types-in-change-log`, similar to before.

calherries (approved):
Looks good now, just have some suggestions but feel free to disagree/ignore

PR #37176, opened 2023-12-29
Trend charts — multiple comparisons

### LINKED ISSUE #35555: Enable multiple comparisons
- Support selecting multiple comparisons, ideally by reusing the existing reordering widgets for funnel steps

kulyk:
Resolves #35555

Allows adding up to three comparisons for a single trend chart. Also, I had to add an `id` to every comparison object because `dndkit` needs stable IDs for sorting. Could be easier to review in order:

* [`fb89155`](https://github.com/metabase/metabase/pull/37176/commits/fb89155ef46581733502560d98933968bbfdfcca) — turns the `scalar.comparisons` setting into an array
* [`b84847e`](https://github.com/metabase/metabase/pull/37176/commits/b84847ed7b8ce920a5c1deb429cfdcf8d92c5b35) — adds an `id` property to comparison objects
* everything else

### Demo

https://github.com/metabase/metabase/assets/17258145/8ca1923d-3cce-42d9-b520-3f7a7817c711

### DIFF
--- a/frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
+++ b/frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
@@ -81,7 +81,7 @@ export function SmartScalar({
   if (trend == null) {
     return null;
   }
-  const { value, clicked, comparison, display, formatOptions } = trend;
+  const { value, clicked, comparisons, display, formatOptions } = trend;
 
   const innerHeight = isDashboard ? height - DASHCARD_HEADER_HEIGHT : height;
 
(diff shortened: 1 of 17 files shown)


frontend/src/metabase/visualizations/visualizations/SmartScalar/SmartScalar.jsx
```diff
@@ -284,12 +289,20 @@ Object.assign(SmartScalar, {
       section: t`Data`,
       title: t`Comparisons`,
       widget: SmartScalarComparisonWidget,
-      isValid: (series, vizSettings) => isComparisonValid(series, vizSettings),
-      getDefault: (series, vizSettings) =>
-        getDefaultComparison(series, vizSettings),
+      isValid: (series, vizSettings) => {
+        const comparisons = vizSettings["scalar.comparisons"];
+        return comparisons.every(comparison =>
+          isComparisonValid(comparison, series, vizSettings),
```
alxnddr, line 295: I would iterate inside `isComparisonValid` function and rename it. The same is about `getDefaultComparison` which can return an array right away. The reason is that we will reuse it as is in the static viz land so we won't duplicate the iteration part.

Also `getDefaultComparison` can be improved to return all valid comparisons if for instance there are 3 but only 1 is invalid

alxnddr (approved):
Looks and works great! I left one minor comment

kulyk:
@alxnddr good point! I'll open another PR for these changes (want the feature branch to be ready for a design review ASAP, we could figure out smaller todos in flight)

PR #37191, opened 2024-01-02
New permissions table schema and backend API

noahmoss:
Initial PR for my proposal to change the `permissions` table schema and move away from permission paths, as detailed [here](https://www.notion.so/metabase/Design-Doc-New-Permissions-Schema-and-API-2a5ff641e7e848e99f8b49bb412ca83f).

This PR:
- Adds a migration to create a `permissions_v2` table with a new schema
- Creates a `metabase.models.permissions-v2` namespace and bootstraps it with a `set-permission!`, `data-permissions-graph`, and `permission-for-user` functions, as well as a number of related utilities and tests.

### DIFF
--- a/src/metabase/models/permissions_v2.clj
+++ b/src/metabase/models/permissions_v2.clj
@@ -1,0 +1,231 @@
+(ns metabase.models.permissions-v2
+  "Model namespace for the V2 permissions system. This is not based on permission paths (as in V1), but rather explicit
+  permission types and associated values for every object (table/collection) or capability on the system."
+  (:require
+   [malli.core :as mc]
+   [metabase.models.interface :as mi]
+   [metabase.util :as u]
+   [metabase.util.i18n :refer [tru]]
+   [methodical.core :as methodical]
+   [toucan2.core :as t2]))
+
+(methodical/defmethod t2/table-name :model/PermissionsV2 [_model] :permissions_v2)
+
+(doto :model/PermissionsV2
+  (derive :metabase/model))
+
+(t2/deftransforms :model/PermissionsV2
+  {:type mi/transform-keyword
+   :value mi/transform-keyword})
+
+
+;;; ---------------------------------------- Permission definitions ---------------------------------------------------
+
+;; IMPORTANT: If you add a new permission type, `:values` must be ordered from *most* permissive to *least* permissive.
+;;
+;;  - When fetching a user's permissions, the default behavior is to return the *most* permissive value from any group the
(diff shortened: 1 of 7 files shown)


noahmoss:
* `master`
  * **PR #37191** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37191?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37191?utm_source=stack-comment).

src/metabase/models/permissions_v2.clj
```diff
@@ -0,0 +1,231 @@
...
+  (when (and (get-in Permissions [perm-type :model])
+             (not object-id))
+    (throw (ex-info (tru "Permission type {0} requires an object ID" perm-type)
+                    {perm-type (Permissions perm-type)}))))
+
+(def perm-types-by-model
+  "A map from model identifiers to a list of permission types that apply to that model."
+  (reduce-kv
+   (fn [m perm-type {:keys [model]}]
+     (when model (update m model (fnil conj #{}) perm-type)))
```
dpsutton, line 92: this should probably be `(if model (update ...) m)` otherwise you risk throwing away your built up state

noahmoss: Haha yeah I actually had fixed that but it got reverted when I undid other changes. Good catch

escherize: nit/wondering out loud: is this `(merge-with (fnil conj #{}) Permissions)` ?

escherize: nope! but this reducing function would be equivalent:

```
(fn [acc perm-type {:keys [model]}]
     (cond-> acc
       model (update model (fnil conj #{}) perm-type)))
```

src/metabase/models/permissions_v2.clj
```diff
@@ -0,0 +1,258 @@
...
+  [perm-type]
+  (-> Permissions perm-type :values last))
+
+(defn- most-permissive-value
+  "The *most* permissive value for a given perm type. This is the default value for superusers."
+  [perm-type]
+  (-> Permissions perm-type :values first))
+
+(def ^:private model-by-perm-type
+  "A map from permiiion types directly to model identifiers (or `nil`)."
```
escherize, line 83:
```suggestion
  "A map from permission types directly to model identifiers (or `nil`)."
```

escherize (approved):
LGTM

PR #37061, opened 2023-12-21
Return `group` instead of just `group_name` for ApiKeys

johnswanson:
Instead of just returning the `group_name`, the frontend also needs the ID.

PR #37021, opened 2023-12-20
add CopyTextInput component

shaunlebron:
Separated from this PR: https://github.com/metabase/metabase/pull/36258#discussion_r1432060057

TODO: make a component or two use this

PR #37205, opened 2024-01-02
Use simple `Secret`s for ApiKeys

johnswanson:
I decided to POC the idea I had here:
https://metaboat.slack.com/archives/CKZEMT1MJ/p1703790833895029

This wraps the generated ApiKeys in a special Secret type. If the object
is accidentally printed, logged, etc. it will appear as "<< REDACTED
SECRET >>" rather than the actual value.

When we're actually *using* the Secret, we call `expose` on it to turn
it into a String.

### DIFF
--- a/src/metabase/api/api_key.clj
+++ b/src/metabase/api/api_key.clj
@@ -10,7 +10,13 @@
    [metabase.util :as u]
    [metabase.util.i18n :refer [tru]]
    [metabase.util.malli.schema :as ms]
+   [metabase.util.secret :as u.secret]
    [toucan2.core :as t2]))
+
+(defn- maybe-expose-key [api-key]
+  (if (contains? api-key :unmasked_key)
+    (update api-key :unmasked_key u.secret/expose)
+    api-key))
 
 (defn- present-api-key
   "Takes an ApiKey and hydrates/selects keys as necessary to put it into a standard form for responses"
@@ -25,12 +31,13 @@
                     :unmasked_key
                     :name
                     :masked_key])
+      (maybe-expose-key)
       (update :updated_by #(select-keys % [:common_name :id]))))
 
 (defn- key-with-unique-prefix []
   (u/auto-retry 5
    (let [api-key (api-key/generate-key)
-         prefix (api-key/prefix api-key)]
+         prefix (api-key/prefix (u.secret/expose api-key))]
      ;; we could make this more efficient by generating 5 API keys up front and doing one select to remove any
      ;; duplicates. But a duplicate should be rare enough to just do multiple queries for now.
      (if-not (t2/exists? :model/ApiKey :key_prefix prefix)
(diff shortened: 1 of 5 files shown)


johnswanson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/37205?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #37014** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37014?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #37051** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #37061** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37061?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #37085** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37085?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #37205** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37205?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
          * **PR #37021** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37021?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #36258** <a href="https://app.graphite.dev/github/pr/metabase/metabase/36258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37205?utm_source=stack-comment).

src/metabase/api/api_key.clj
```diff
@@ -124,6 +131,7 @@
                             :unmasked_key unhashed-key
                             :masked_key (api-key/mask unhashed-key)))))
 
+
```
noahmoss, line 134: nit: unintended newline?

src/metabase/util/secret.clj
```diff
@@ -0,0 +1,27 @@
...
+
+;; Define a protocol for secrets to make things harder to accidentally expose.
+(defprotocol ISecret
+  (expose [this] "Expose the secret"))
+
+(defrecord Secret [value-fn]
+  ISecret
+  (expose [_this] (value-fn))
+  Object
+  (toString [_this] "<< REDACTED SECRET >>"))
```
noahmoss, line 14: I wonder if this should be translated, in case it pops up on the UI somewhere? That's not going to happen right now, of course, but if this is a utility namespace it's difficult to forsee future use cases.

Probably an edge case though, so nbd either way.

noahmoss (approved):
I like it

src/metabase/util/secret.clj
johnswanson: I think translation makes sense... but do we use `trs` or `tru`? It's actually kind of hard to say where these strings are more likely to end up, since the intent is that they won't end up anywhere... 🤔

I think if I had to guess I'd say it's slightly more likely for them to end up in server logs (or other telemetry data) than in the UI (it seems like it would take some intentional effort to put one in the UI, whereas you could just accidentally send a map as context on a log). So I'll use `trs` unless you think it makes sense to use `tru`.

johnswanson: I'm merging this with `trs` for now, we can change to `tru` in a followup PR if desired.

PR #37206, opened 2024-01-02
[MLv2] Normalize expression-clause

### LINKED ISSUE #37100: [MLv2] [BE] [Bug] fieldValuesSearchInfo issues with native queries
2 E2E tests were skipped in https://github.com/metabase/metabase/pull/37072 because of issues in `fieldValuesSearchInfo`:

- joins.cy.spec.js - `should join a native question`
- native.cy.spec.js - `shouldn't remove rows containing NULL when using "Is not" or "Does not contain" filter (metabase#13332)`

`fieldValuesSearchInfo` in these tests return `fieldId` and `searchFieldId` as strings, but it should return only real field ids (numbers). If there is no field id it should return `null` `fieldId` and `searchFieldId`. 

In other words, for native queries `fieldValuesSearchInfo` should return `null` for `fieldId` and `searchFieldId` unless it's a native model with column metadata overrides where a DB field is set; in that case the same logic as for MBQL queries should work.

Please unskip the tests when the issue is fixed.

### LINKED ISSUE #37175: [MLv2] [BE] [Bug] displayName for relative date time filters doesn't work correctly if query is not converted to legacy query first
`expressionClause` function doesn't convert strings to keywords when needed and this leads to invalid mbql queries being created. It works correctly in production only because we convert the query to "legacy query" first and then retrieve it back, and it fixes the string <-> keyword mismatch.

In the following example I create a date filter with `Lib.expressionClause(["time-interval", column, "current", "day"])` and later `displayInfo` for this clause doesn't work correctly because other functions expect clojure keywords. The condition is evaluated incorrectly here - see that `n` and `unit` are strings and the function expects keywords:

Screenshot 2023-12-29 at 18 26 21

snoe:
Fixes #37175 

This ports over some of `mbql.normalize` for particular expressions that take keywords

### DIFF
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -472,7 +472,11 @@
 (defn ^:export expression-clause
   "Returns a standalone clause for an `operator`, `options`, and arguments."
   [an-operator args options]
-  (lib.core/expression-clause (keyword an-operator) args (js->clj options :keywordize-keys true)))
+  (-> (lib.core/expression-clause
+        (keyword an-operator)
+        args
+        (js->clj options :keywordize-keys true))
+      (lib.core/normalize)))
 
 (defn ^:export expression-parts
   "Returns the parts (operator, args, and optionally, options) of `expression-clause`."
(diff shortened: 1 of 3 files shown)


metamben (approved):
LGTM

PR #37207, opened 2024-01-02
Update enterprise readme to 2024

perivamsi:


noahmoss (approved):
Hah, I missed this in my previous PR because I grepped for "2023". Nice catch.

PR #37208, opened 2024-01-02
Consistent formatting of all columns in Excel exports

### LINKED ISSUE #14393: Timestamp is always present in dates when exporting to Excel in the embedded application.
**Describe the bug**
Dates are displayed without timestamp in visualization for native queries, but when exported they include timestamp after I embed Metabase to our application.

**To Reproduce**
1. Create native query with a date and use convert function to have only dd-mm-yyyy
2. Save question (it worked fine if you download at this point) - saw that u have resolved the known issue #12964
3. Add Question in Dashboard
4. Embed the Dashboard to your application website
5. Download the result as Excel from website
6. Open Excel  and see that timestamp is still present

**Expected behavior**
I expect the Excel export to look just like what I am seeing in the table visualization, without the timestamp.

**Screenshots**
image

```json
  "browser-info": {
    "language": "en-US",
    "platform": "Win32",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.141 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.7+10",
    "java.version": "11.0.7",
    "java.vm.version": "11.0.7+10",
    "os.name": "Linux",
    "os.version": "4.15.0-1103-azure",
    "user.language": "en",
    "user.timezone": "GMT"
  },
  "metabase-info": {
    "databases": [
      "sqlserver"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "11.6"
      },
      "jdbc-driver": {
        "version": "42.2.8"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2021-01-05",
      "tag": "v0.37.5",
      "branch": "release-x.37.x",
      "hash": "be537ee"
    },
    "settings": {
      "report-timezone": "Europe/Brussels"
    }
  }
}
```

**Severity**
Our customers need this data for their analysis. So its really important to fix this.

### LINKED ISSUE #17753: Exports should include formatting everywhere
I really don't understand how our code functions, but I am wasting so much time on having to validate every single way we can export, which clearly seems like it should be unified to avoid regressions (we have had many of those).

Currently exports can have very different formatting in these cases:
- In-app, unsaved question
- In-app, saved question
- Public Sharing
- Signed Embedding
- Pulse/Dashboard Subscription
- *...other?*

Create one function that actually generates the export with correct formatting, which is then used by all the different endpoints.

Critical. I have wasted too much time testing exports, and we have had too many regressions (example #17752).

Related #16234

markbastian:
This PR has a set of changes that should provide consistent formatting across exports of Excel, CSV, and JSON files, including formatting applied globally, as metadata, and as visualization-specific settings.

Changes include:
- Consolidation of settings merge logic into `metabase.query-processor.streaming.common` as `viz-settings-for-col`. This is now used for `xlsx`, `csv`, and `json` exports.
  - The `xlsx` still needs to take these settings and produce cell formatters where the `csv` and `json` branches need to produce string formatters. Either way, this is a good incremental step towards a pipeline of all the settings -> unified settings per column -> format -> product of format (cell style or formatter).
  - This also provides a more functional style in which we reach for global settings far less and rely on what is passed in via the unified settings (which get global settings from the `metabase.query-processor.middleware.visualization-settings` qp middleware).
- Modify `metabase.query-processor.streaming.xlsx` to use the same strategy as csv and json exports of computing styles for visible columns only up front then reusing these styles
- Removing trailing zeroes for percentages in CSV and JSON files prior to formatting to conform to the FE. While this gets us close to consistent behavior, we should maybe revisit the idea of default stripping of trailing zeroes.
- Adds consistent seed values of `nil` to the `volatile!` fields used in stream processing for column formatting.
  - Note that this ns still has some potential to be streamlined to remove code that is close to code in the csv/json path but this PR is meant to be a start, not to completely refactor the whole ns.
- Unit tests
  - All existing tests pass. Note that `metabase.query-processor.streaming.xlsx-test/format-string` has been modified to compute a unified settings map to better follow what happens in practice with the qp middleware.
  - Additional smaller and large-scale tests were added to `metabase.api.card-test` to test end to end xlsx exports.

Fixes #14393
Closes #17753

### DIFF
--- a/src/metabase/formatter.clj
+++ b/src/metabase/formatter.clj
@@ -112,8 +112,10 @@
                              (not grouping) (str/replace #"," ""))]
     (fn [value]
       (if (number? value)
-        (let [scaled-value (* value (or scale 1))
-              decimals-in-value (digits-after-decimal (if percent? (* 100 scaled-value) scaled-value))
+        (let [scaled-value      (cond-> (* value (or scale 1))
+                                  percent?
+                                  (* 100))
+              decimals-in-value (digits-after-decimal scaled-value)
               decimal-digits (cond
                                decimals decimals ;; if user ever specifies # of decimals, use that
                                integral? 0
@@ -127,22 +129,25 @@
                                            decimals-in-value))))
               fmt-str (cond-> base
                         (not (zero? decimal-digits)) (str "." (apply str (repeat decimal-digits "0")))
-                        scientific? (str "E0")
-                        percent?    (str "%"))
+                        scientific? (str "E0"))
               fmtr (doto (DecimalFormat. fmt-str symbols) (.setRoundingMode RoundingMode/HALF_UP))]
           (map->NumericWrapper
            {:num-value value
-            :num-str   (str (when prefix prefix)
-                            (when (and currency? (or (nil? currency-style)
-                                                     (= currency-style "symbol")))
-                              (get-in currency/currency [(keyword (or currency "USD")) :symbol]))
-                            (when (and currency? (= currency-style "code"))
-                              (str (get-in currency/currency [(keyword (or currency "USD")) :code]) \space))
-                            (cond-> (.format fmtr scaled-value)
-                              (not decimals) (strip-trailing-zeroes decimal))
-                            (when (and currency? (= currency-style "name"))
-                              (str \space (get-in currency/currency [(keyword (or currency "USD")) :name_plural])))
-                            (when suffix suffix))}))
+            :num-str   (let [inline-currency? (and currency?
+                                                   (false? (::mb.viz/currency-in-header column-settings)))]
+                         (str (when prefix prefix)
+                              (when (and inline-currency? (or (nil? currency-style)
+                                                       (= currency-style "symbol")))
+                                (get-in currency/currency [(keyword (or currency "USD")) :symbol]))
+                              (when (and inline-currency? (= currency-style "code"))
+                                (str (get-in currency/currency [(keyword (or currency "USD")) :code]) \space))
+                              (cond-> (.format fmtr scaled-value)
+                                (and (not currency?) (not decimals))
+                                (strip-trailing-zeroes decimal)
+                                percent?    (str "%"))
+                              (when (and inline-currency? (= currency-style "name"))
+                                (str \space (get-in currency/currency [(keyword (or currency "USD")) :name_plural])))
+                              (when suffix suffix)))}))
         value))))
 
 (s/defn format-number :- NumericWrapper
--- a/src/metabase/query_processor/streaming/common.clj
+++ b/src/metabase/query_processor/streaming/common.clj
@@ -149,7 +149,7 @@
   (:type/Number global-column-settings {}))
 
 (defmethod global-type-settings :type/Currency [_ {::mb.viz/keys [global-column-settings] :as _viz-settings}]
-  (:type/Currency global-column-settings {}))
+  (:type/Currency global-column-settings {::mb.viz/number-style "currency"}))
 
 (defmethod global-type-settings :default [_ _viz-settings]
   {})
--- a/src/metabase/query_processor/streaming/common.clj
+++ b/src/metabase/query_processor/streaming/common.clj
@@ -1,6 +1,7 @@
 (ns metabase.query-processor.streaming.common
   "Shared util fns for various export (download) streaming formats."
   (:require
+   [clojure.string :as str]
    [java-time.api :as t]
    [metabase.public-settings :as public-settings]
    [metabase.query-processor.store :as qp.store]
@@ -69,7 +70,7 @@
   "Merge format settings defined in the localization preferences into the format settings
   for a single column."
   [format-settings global-settings-key]
-  (let [global-settings (global-settings-key (public-settings/custom-formatting))
+  (let [global-settings (get (public-settings/custom-formatting) global-settings-key)
         normalized      (mb.viz/db->norm-column-settings-entries global-settings)]
     (merge normalized format-settings)))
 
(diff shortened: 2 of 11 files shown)


src/metabase/formatter.clj
```diff
@@ -112,8 +112,10 @@
                              (not grouping) (str/replace #"," ""))]
     (fn [value]
       (if (number? value)
-        (let [scaled-value (* value (or scale 1))
-              decimals-in-value (digits-after-decimal (if percent? (* 100 scaled-value) scaled-value))
+        (let [scaled-value      (cond-> (* value (or scale 1))
+                                  percent?
+                                  (* 100))
```
adam-james-v, line 117: Yea, this is a nice small change that makes the code more sensible :)

src/metabase/formatter.clj
```diff
@@ -127,22 +129,25 @@
...
+                         (str (when prefix prefix)
+                              (when (and inline-currency? (or (nil? currency-style)
+                                                       (= currency-style "symbol")))
+                                (get-in currency/currency [(keyword (or currency "USD")) :symbol]))
+                              (when (and inline-currency? (= currency-style "code"))
+                                (str (get-in currency/currency [(keyword (or currency "USD")) :code]) \space))
+                              (cond-> (.format fmtr scaled-value)
+                                (and (not currency?) (not decimals))
+                                (strip-trailing-zeroes decimal)
+                                percent?    (str "%"))
```
adam-james-v, line 147: I can't quite follow why this `(str "5")` was moved from the `fmt-str` in the let binding? 
Whatever the logic is for moving it, should it also be applied to the `scientific?` pred that is still in `fmt-str`?

src/metabase/formatter.clj
markbastian: It also gives consistency with the FE by trimming trailing zeros for percentages. I'm still not sure I like that behavior, but it is consistent.

src/metabase/formatter.clj
markbastian: This was done so that trailing zeros could be trimmed from the string. This is to be consistent with the FE.

src/metabase/query_processor/streaming/common.clj
```diff
@@ -69,7 +70,7 @@
   "Merge format settings defined in the localization preferences into the format settings
   for a single column."
   [format-settings global-settings-key]
-  (let [global-settings (global-settings-key (public-settings/custom-formatting))
+  (let [global-settings (get (public-settings/custom-formatting) global-settings-key)
```
adam-james-v, line 73: This is just curiosity: why is `(get {} :key)` preferred here over `(:key {})`? Just style pref? Something more consequential?

src/metabase/formatter.clj
adam-james-v: Ah, of course! Thanks

src/metabase/query_processor/streaming/common.clj
```diff
@@ -112,3 +113,64 @@
       (if (and is-currency? (::mb.viz/currency-in-header merged-settings true))
         (str column-title " (" (currency-identifier merged-settings) ")")
         column-title))))
+
+(defn normalize-keys
```
adam-james-v, line 117: We have `metabase.util/normalize-map` and a few various key normalization functions in the util namespace. Should we consider using a util fn instead? 

I'm happy with this solution too, it's small and straightforward. Just wanted to mention I saw that util fn in case you weren't aware.

src/metabase/query_processor/streaming/common.clj
markbastian: Normally I just use the key as a function, but since the key is dereferenced value I felt that this gives some level of self-documentation -- it's obvious this is a lookup and not a ns-level function. I don't have a strong opinion on reverting this line, though.

src/metabase/query_processor/streaming/common.clj
markbastian: I'm actually thinking maybe we should drop this entirely and force `metabase.formatter.datetime` to use fully-qualified keys.

src/metabase/query_processor/streaming/common.clj
adam-james-v: No, I think keep it. I was really just curious what the thought was. Thanks for the explanation.

src/metabase/query_processor/streaming/common.clj
adam-james-v: That's not a bad idea, actually.

src/metabase/query_processor/streaming/common.clj
```diff
@@ -112,3 +113,64 @@
...
+
+(defn viz-settings-for-col
+  "Get the unified viz settings for a column based on the column's metadata (if any) and user settings (⚙)."
+  [{column-name :name metadata-column-settings :settings :keys [field_ref] :as col} viz-settings]
+  (let [[_ field-id-or-name] field_ref
+        all-cols-settings (-> viz-settings
+                              ::mb.viz/column-settings
+                              ;; update the keys so that they will have only the :field-id or :column-name
+                              ;; and not have any metadata. Since we don't know the metadata, we can never
+                              ;; match a key with metadata, even if we do have the correct name or id
```
adam-james-v, line 165: This part is debatable right? Like, in a Slack thread I thought it was discussed that column name or field-id is not technically sufficient to disambiguate columns in all cases. 

This assumption (dropping metadata) has been the behaviour already I'm pretty sure, so this PR shouldn't try to fix it, but this is something we should revisit later maybe? Please disregard this if I'm misremembering.

test/metabase/api/card_test.clj
```diff
@@ -1934,6 +1935,172 @@
...
+          ;; - "3D PCT" is a standard percentage with a customization of 3 significant digits
+          (testing "All formatting is applied correctly in a complex situation."
+            (is (= [["No Formatting TS" "DATE-ONLY TS" "TS W/FORMATTING" "Date" "Time" "Plain Currency ($)" "Col $" "USD Col" "DOL Col" "EXPO" "Scaled PCT" "3D PCT"]
+                    ["Jan 1, 2023, 12:34 PM" "2023-1-1" "1-1-2023, 12:34:56.000" "Jan 1, 2023" "12:34 PM" "1,234.05" "[$$]1,234.05" "[$USD] 2345.05" "4,321.05 US dollars" "7180.64E+0" "0.05%" "19.200%"]
+                    ["Jan 1, 2023, 12:34 PM" "2023-1-1" "1-1-2023, 12:34:56.000" "Jan 1, 2023" "12:34 PM" "2,345.30" "[$$]2,345.30" "[$USD] 3456.30" "2,931.30 US dollars" "1.71806E+4" "8.02%" "0.000%"]
+                    ["Jan 1, 2023, 12:34 PM" "2023-1-1" "1-1-2023, 12:34:56.000" "Jan 1, 2023" "12:34 PM" "3,456.00" "[$$]3,456.00" "[$USD] 2300.00" "2,250.00 US dollars" "12.7181E+4" "95.40%" "11.580%"]]
+                   (parse-xlsx-results-to-strings
+                     (mt/user-http-request :rasta :post 200 (format "card/%d/query/xlsx" (u/the-id card))))))))))))
```
adam-james-v, line 2100: I'm glad there's a test for all of the formatting. If something breaks, will the test failure be readable? I suppose the diff will show fairly quickly which formatter is not working as expected?

The alternative I can think of is tons of tiny tests, which I'm not sure what's better or worse?

adam-james-v (approved):
Any comments/questions I left are more curiosity or minor considerations, none of which I think should block this PR, so I'm approving it.

GeorgeVragalev:
when is this going to be deployed and available?
as from what I can see now the csv download from metabase UI for questions doesn't match the table columns that were made to have custom names

PR #37211, opened 2024-01-02
Implement brush filter helper functions

### LINKED ISSUE #37103: [MLv2] [BE] Add brush filter helper functions
In visualizations there are several drag-and-drop based drill thrus. Existing implementation:
- date filter https://github.com/metabase/metabase/blob/98bcd7fc3102bd7c07e8b68878c3738f3cb8727b/frontend/src/metabase-lib/queries/utils/actions.js#L75-L132
- numeric filter https://github.com/metabase/metabase/blob/98bcd7fc3102bd7c07e8b68878c3738f3cb8727b/frontend/src/metabase-lib/queries/utils/actions.js#L151-L154
- geo filter https://github.com/metabase/metabase/blob/98bcd7fc3102bd7c07e8b68878c3738f3cb8727b/frontend/src/metabase-lib/queries/utils/actions.js#L134-L149

Notes:
- They add a new filter or update an existing one for the same column
- Brush drill-thru for date columns also modifies the temporal unit
- In visualizations we only have `DatasetColumn` (coming from `data.cols`). It should be converted internally.

Proposed signature:

```
type BrushDrillThruRange = {
  start: RowValue,
  end: RowValue,
};

type BrushDrillThruBounds = {
  north: RowValue,
  west: RowValue,
  east: RowValue,
  south: RowValue,
};

// numeric or date filter
brushDrillThru(
  query: Query,
  stageIndex: number,
  column: DatasetColumn,
  range: BrushDrillThruRange,
): Query

// geo filter
brushDrillThru(
  query: Query,
  stageIndex: number,
  latitudeColumn: DatasetColumn,
  longitudeColumn: DatasetColumn,
  bounds: BrushDrillThruBounds,
): Query
```

camsaul:
Adds a handful of helper functions for powering the paintbrush-style zoom in filter stuff in the FE

Resolves #37103


src/metabase/lib/filter/update.cljc
```diff
@@ -0,0 +1,179 @@
...
+    numeric-column :- ::lib.schema.metadata/column
+    start          :- number?
+    end            :- number?]
+   (let [[start end] (sort [start end])]
+     (-> query
+         (remove-existing-filters-against-column stage-number numeric-column)
+         (lib.filter/filter stage-number (lib.filter/between numeric-column start end))))))
+
+(def ^:private temporal-filter-min-num-rows
+  "Minimum number of rows an updated query should return before; if it will return less than this, switch to
```
metamben, line 68:
```suggestion
  "Minimum number of rows an updated query should return; if it will return less than this, switch to
```
I don't understand what "before" is referring to.

I was a bit confused by the part about the "rows" returned by the query. It suggests that we actually run the/some query to see how many rows it returns. Reading further I understood this is about number of distinct values that the breakout can have given its unit.

src/metabase/lib/filter/update.cljc
```diff
@@ -0,0 +1,179 @@
...
+  that will."
+  [unit  :- ::lib.schema.temporal-bucketing/unit.date-time.truncate
+   start :- ::lib.schema.literal/temporal
+   end   :- ::lib.schema.literal/temporal]
+  (or
+   (let [num-points (shared.ut/unit-diff unit start end)]
+     (when (< num-points temporal-filter-min-num-rows)
+       (when-let [next-largest-unit (unit->next-largest-unit unit)]
+         (temporal-filter-find-best-breakout-unit next-largest-unit start end))))
+   unit))
```
metamben, line 89: I `loop` would be easier to read, I guess.

src/metabase/lib/filter/update.cljc
```diff
@@ -0,0 +1,179 @@
...
+  other breakouts against that column."
+  [query        :- ::lib.schema/query
+   stage-number :- :int
+   column       :- ::lib.schema.metadata/column
+   new-unit     :- ::lib.schema.temporal-bucketing/unit.date-time.truncate]
+  (transduce
+   identity
+   (fn
+     ([]
+      {:query query, :has-seen-column? false})
```
metamben, line 102: Wouldn't it be clearer to use the 4 arity version of `transduce` instead? Or just use `reduce` and return `:query` from its result?

src/metabase/lib/filter/update.cljc
```diff
@@ -0,0 +1,179 @@
...
+   (let [query (remove-existing-filters-against-column query stage-number temporal-column)
+         unit  (lib.temporal-bucket/raw-temporal-bucket temporal-column)]
+     (if-not unit
+       ;; Temporal column is not bucketed: we don't need to update any temporal units here. Add/update a `:between`
+       ;; filter.
+       (lib.filter/filter query stage-number (lib.filter/between temporal-column start end))
+       ;; temporal-column IS bucketed: need to update the breakout(s) against this column.
+       (let [;; clamp range to unit to ensure we select exactly what's represented by the dots/bars. E.g. if I draw my
+             ;; filter from at `2024-01-02` to `2024-03-05` and the unit is `:month`, we should only show the months
+             ;; between those two values, i.e. only `2024-02` and `2024-03`.
```
metamben, line 142: Does this mean that first value cannot be included such a selection?

metamben (approved):

frontend/src/metabase-lib/filter.ts
```diff
@@ -868,3 +868,64 @@ function deserializeExcludeDatePart(
...
+  stageIndex: number,
+  temporalColumn: ColumnMetadata,
+  start: string,
+  end: string,
+): Query {
+  return ML.update_temporal_filter(
+    query,
+    stageIndex,
+    temporalColumn,
+    longitudeColumn,
```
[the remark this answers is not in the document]
camsaul: good catch, fixed

src/metabase/lib/filter/update.cljc
camsaul: Fixed and added a lot more comments clarifying what I meant. I changed "rows" to "points" here since it's a better reflection

src/metabase/lib/filter/update.cljc
camsaul: fixed

src/metabase/lib/filter/update.cljc
camsaul: fixed

PR #37213, opened 2024-01-03
Modify type metabase_field.database_required to bit(1) in mysql

qnkhuat:
We change all of the boolean field in MySQL to `bit(1)` in https://github.com/metabase/metabase/pull/36964, but this is a missing case.

Also added a test.


noahmoss (approved):
nice, thanks!

PR #37218, opened 2024-01-03
Filter modal date tests

ranquild:
Related bug https://github.com/metabase/metabase/issues/37219


frontend/src/metabase/querying/components/FilterModal/DateFilterEditor/DateFilterEditor.unit.spec.tsx
```diff
@@ -0,0 +1,215 @@
...
+        column,
+      });
+
+      userEvent.click(screen.getByLabelText("More options"));
+      userEvent.click(await screen.findByText("Exclude…"));
+      userEvent.click(screen.getByText("Hours of the day…"));
+      userEvent.click(screen.getByText("5 PM"));
+      userEvent.click(screen.getByText("Add filter"));
+
+      expect(getNextFilterName()).toBe("Created At excludes the hour of 5 PM");
```
uladzimirdev, line 177: the issue claims `screen.getByText("Created At excludes the hour of 5 PM") should be a different string.`

ranquild: yep. The idea is that this test should be updated when the issue is fixed.

uladzimirdev (approved):

PR #37229, opened 2024-01-03
[MLv2] Removing a final JOIN condition removes the whole join

### LINKED ISSUE #36690: [MLv2] [BE] removeClause throws "Cannot remove the final join condition" error
Given the following query ([screenshot](https://github.com/metabase/metabase/assets/6830683/469a5495-47ad-478c-9ccb-5394e23b78ef)), `removeClause` throws `Cannot remove the final join condition` error when trying to remove the breakout.

```tsx
Lib.removeClause(query, 0, Lib.breakouts(query, 0)[0])
```

It is expected that `removeClause` drops the entire join clause because it becomes invalid.

### How to reproduce 

1. `git checkout 36516-notebook-steps`
2. Run any of the following e2e tests:
    - `joins.cy.spec.js` (`should handle joins on different stages`)
    - `reproductions/17767-cannot-join-on-aggregation-with-implicit-joins.cy.spec.js`
    - `reproductions/18818-crash-when-joining-on-custom-column.cy.spec.js`

### Additional context

See [Slack](https://metaboat.slack.com/archives/C04CYTEL9N2/p1702378656823469)

bshepherdson:
This still throws an exception on trying to directly remove the final
JOIN condition. But if the condition is being removed because a field it
depends on is being removed (eg. a custom column, a breakout, a
previous stage aggregation) then the entire JOIN is also removed, in the
usual cascading fashion.

Fixes #36690.

### DIFF
--- a/dev/src/dev/fe_helpers.cljs
+++ b/dev/src/dev/fe_helpers.cljs
@@ -1,15 +1,30 @@
 (ns dev.fe-helpers)
 
-(defn redux-state []
+(defn redux-state
+  "Returns the root Redux state, the JS object holding the complete state of the app.
+
+  This is hacky - it reaches deep into the internals of Redux, and may break in the future. That seems acceptable for a
+  dev time helper."
+  []
   (let [root  (js/document.querySelector "#root")
         store (.. root -_reactRootContainer -_internalRoot -current -child -memoizedProps -store)]
     (.getState store)))
 
-(defn current-card []
+(defn current-card
+  "Retrieves the current query's card from the Redux state.
+
+  Undefined behavior if there is not currently a single question loaded in the UI."
+  []
   (.. (redux-state) -qb -card))
 
-(defn current-legacy-query-js []
+(defn current-legacy-query-js
+  "Gets the legacy query for the currently loaded question."
+  []
   (.-dataset_query (current-card)))
 
-(defn current-query []
+(defn current-query
+  "Gets the MLv2 query for the currently loaded question.
+
+  Hack: This relies on a dev-mode-only global property that's set whenever a Question object is converted to MLv2."
+  []
   (.-__MLv2_query js/window))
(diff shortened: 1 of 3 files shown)


bshepherdson:
* `master`
  * **PR #37229** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37229?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37229?utm_source=stack-comment).

dev/src/dev/fe_helpers.cljs
```diff
@@ -13,3 +13,5 @@
 
 (defn current-query []
   (.-__MLv2_query js/window))
+
+
```
metamben, line 17: Is this intended?

src/metabase/lib/remove_replace.cljc
```diff
@@ -118,12 +119,24 @@
...
                                                       (set/subset? (set target-opts) (set %))))
                                        target-ref-id] [location clause]))))))
-                   (stage-paths query stage-number))]
-    (reduce
-     (fn [query [location target-clause]]
-       (remove-replace-location query stage-number unmodified-query-for-stage location target-clause #(lib.util/remove-clause %1 %2 %3 stage-number)))
-     query
-     to-remove)))
+                   (stage-paths query stage-number))
+        dead-joins (atom [])]
```
metamben, line 123: `volatile!` would be enough.

metamben (changes requested):
Don't we need a test checking the new behavior?

dev/src/dev/fe_helpers.cljs
bshepherdson: Ah, no it's not. I added a new helper that didn't pan out and didn't clean it up all the way.

metamben (approved):
👍

PR #37233, opened 2024-01-03
Rename `Question.query()` to `Question.legacyQuery()`

ranquild:
The motivation for this change is to make it easier to spot legacy query usages.

If this changes goes well, we can migrate `_getMLv2Query` to `query`.


uladzimirdev (approved):
image
LGTM 🤞

PR #37236, opened 2024-01-03
Make e2e tests less flaky

uladzimirdev:
### Description

Since drag&drop is not stable, filter results can vary between runs. So the idea to simplify assertion was discussed ([thread](https://metaboat.slack.com/archives/C5XHN8GLW/p1704299003934219))

nagivating back is flaky because of the number of printed rows, (1-7 vs 1-8), so I made it less dependent on the viewport size

### How to verify

Tests should pass

### Checklist

### DIFF
--- a/e2e/test/scenarios/visualizations-charts/line_chart.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-charts/line_chart.cy.spec.js
@@ -597,6 +597,12 @@ describe("scenarios > visualizations > line chart", () => {
       cy.wait("@dataset");
 
       cy.findByTestId("filter-pill").should("contain.text", "Created At is");
+
+      const X_AXIS_VALUE = "June 2022";
+      cy.get(".CardVisualization").within(() => {
+        cy.get(".x-axis-label").should("have.text", "Created At");
+        cy.findByText(X_AXIS_VALUE);
+      });
     },
   );
 
@@ -629,6 +635,11 @@ describe("scenarios > visualizations > line chart", () => {
       "contain.text",
       "Quantity is between",
     );
+    const X_AXIS_VALUE = 8;
+    cy.get(".CardVisualization").within(() => {
+      cy.get(".x-axis-label").should("have.text", "Quantity");
+      cy.findByText(X_AXIS_VALUE);
+    });
   });
 });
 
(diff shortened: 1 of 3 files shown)


e2e/test/scenarios/visualizations-charts/line_chart.cy.spec.js
```diff
@@ -596,12 +596,7 @@ describe("scenarios > visualizations > line chart", () => {
 
       cy.wait("@dataset");
 
-      cy.findByTestId("filter-pill").should(
-        "have.text",
-        "Created At is Apr 1, 12:00 AM – Sep 1, 2022, 12:00 AM",
-      );
-
-      cy.get(".Visualization .dot").should("have.length", 6);
+      cy.findByTestId("filter-pill").should("contain.text", "Created At is");
```
alxnddr, line 599: It would be nice to still assert that the date filter is correct because if drill provides invalid values I guess there will be `Created At is` and the test will pass

alxnddr: Maybe querying chart axis ticks would be enough to ensure it is rendered

uladzimirdev: It's not a test for a drill, it's a test for a brush filter 😄 anyhow, we need this test only to be sure that upcoming work about MLv2 migration of the related filters will still work.
Unfortunately, cypress doesn't provide deterministic results even if we pass correct coordinates for drag start and drag end - results vary between runs even locally

alxnddr: In ECharts scope we had to reimplement brush drilling so it would be handy there as well. I think if you add findByText(*expected x-axis tick value*) it would cover the scenario slightly better because invalid filters would produce wrong query and won't render anything

ranquild: Agree, maybe we could check for some value in between the range in the chart that won't be flaky

ranquild (approved):

e2e/test/scenarios/visualizations-charts/line_chart.cy.spec.js
uladzimirdev: https://github.com/metabase/metabase/pull/37236/commits/20a8711f76784502c4664afbfc729f15f387bee8

alxnddr (approved):

PR #37237, opened 2024-01-03
Guard against version duplication (#33803)

### LINKED ISSUE #33803: Version info could have duplicated versions
### Describe the bug

The older version from `http://static.metabase.com/version-info-ee.json` could contain the latest version
image

This could happen for cases where we release a version twice, but that' totally possible so the code to update version info should prevent this.

### To Reproduce

^

### Expected behavior

_No response_

### Logs

_No response_

```JSON
master
```

### Severity

P3? since it does not really introduce any bug

### Additional context

_No response_

crisptrutski:
Closes https://github.com/metabase/metabase/issues/33803

### Description

This change adds some defensive code to avoid introducing duplicate versions into the "version-info" files during a release.

It also adds test coverage to show that not only are duplicates now prevented, existing duplication will be remediated.

### How to verify

Run the build tests:

```sh
clj -X:dev:drivers:build:build-dev:build-test
```

### DIFF
--- a/bin/build/src/release/version_info.clj
+++ b/bin/build/src/release/version_info.clj
@@ -47,19 +47,25 @@
    ;; TODO -- these need to be curated a bit before publishing...
    :highlights (mapv :title (github/milestone-issues))})
 
+(defn- update-version-info [{:keys [latest] :as info} new-version-info]
+  (if (= (:version new-version-info) (:version latest))
+    ;; if the latest version has not changed, just return the current info
+    info
+    (-> info
+        ;; move the current `:latest` to the beginning of `:older`
+        (update :older (fn [older] (distinct (cons latest older))))
+        (assoc :latest new-version-info))))
+
 (defn- generate-version-info! []
   (let [filename (version-info-filename)
         tmpname  (tmp-version-info-filename)]
     (u/step (format "Generate %s" filename)
       (u/step (format "Delete and create %s" tmpname)
         (u/delete-file-if-exists! tmpname)
-        (let [{:keys [latest], :as info} (current-version-info)]
-          (spit tmpname (-> info
-                            ;; move the current `:latest` to the beginning of `:older`
-                            (update :older (fn [older]
-                                             (distinct (cons latest older))))
-                            (assoc :latest (info-for-new-version))
-                            json/generate-string)))))))
+          (let [current-info     (current-version-info)
+                new-version-info (info-for-new-version)
+                new-info         (update-version-info current-info new-version-info)]
+            (spit tmpname (json/generate-string new-info)))))))
 
 (defn- upload-version-info! []
   (u/step "Upload version info"
--- a/bin/build/test/release/version_info_test.clj
+++ b/bin/build/test/release/version_info_test.clj
@@ -17,34 +17,43 @@
    :patch      patch
    :highlights highlights})
 
+;; TODO -- can result in a flakey test if the date changes during this test run
 (def ^:private today (.format (DateTimeFormatter/ofPattern "yyyy-MM-dd") (LocalDate/now)))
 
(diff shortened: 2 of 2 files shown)


bin/build/src/release/version_info.clj
```diff
@@ -47,19 +47,25 @@
...
-        (let [{:keys [latest], :as info} (current-version-info)]
-          (spit tmpname (-> info
-                            ;; move the current `:latest` to the beginning of `:older`
-                            (update :older (fn [older]
-                                             (distinct (cons latest older))))
-                            (assoc :latest (info-for-new-version))
-                            json/generate-string)))))))
+          (let [current-info (current-version-info)
+                new-version-info (info-for-new-version)
+                new-info (update-version-info current-info new-version-info)]
```
qnkhuat, line 67:
```suggestion
          (let [current-info     (current-version-info)
                new-version-info (info-for-new-version)
                new-info         (update-version-info current-info new-version-info)]
```

nit: we try to align let binding and map this way, it's much easier to read IMHO.

bin/build/test/release/version_info_test.clj
```diff
@@ -17,34 +17,43 @@
...
+     :older  (for [v (rest versions)]
+               (apply make-version-map (conj v enterprise?)))}))
+
+(defn- duplicate-version [versions]
+  (into [(first versions)] versions))
+
+(def ^:private previous-version-cases
+  {"with previous versions" (rest test-versions)
+   "with duplicate previous versions" (duplicate-version (rest test-versions))
+   "including the current version" test-versions})
```
qnkhuat, line 42:
```suggestion
  {"with previous versions"           (rest test-versions)
   "with duplicate previous versions" (duplicate-version (rest test-versions))
   "including the current version"    test-versions})
```

nit: we try to align the map so that the values are aligned

If you're on emacs I think there are a plugin that do this automatically.

qnkhuat (approved):
LGTM, just has a couple of nits.

crisptrutski:
Totally agree with the style tips - had naively hoped that Calva would pick up the repo conventions. Will manually fix for now.

bin/build/test/release/version_info_test.clj
```diff
@@ -17,34 +17,43 @@
...
-                              ["38.3" "2021-04-01" true ["Fix 1" "Fix 2"]]
-                              ["38.2" "2021-03-21" true ["Older Fix 1" "Older Fix 2"]]
-                              ["38.1" "2021-03-14" true ["Much Older Fix 1" "Much Older Fix 2"]]
-                              ["38.0" "2021-02-15" true ["Super Old Fix 1" "Super Old Fix 2"]]])
+(def ^:private test-versions
+  [["39.0" today false ["Super New Fix 1" "Super New Fix 2"]]
+   ["38.3" "2021-04-01" true ["Fix 1" "Fix 2"]]
+   ["38.2" "2021-03-21" true ["Older Fix 1" "Older Fix 2"]]
+   ["38.1" "2021-03-14" true ["Much Older Fix 1" "Much Older Fix 2"]]
+   ["38.0" "2021-02-15" true ["Super Old Fix 1" "Super Old Fix 2"]]])
```
calherries, line 28: This is fairly subjective they were here before, but I would have expect these private defs to be inside `build-info-test`. Same with `previous-version-cases`. The main benefit is readability - these defs are very specific to the test's implementation and should be grouped with the assertions in a `deftest` block.

calherries (approved):
Nice! I left one cosmetic suggestion

bin/build/test/release/version_info_test.clj
crisptrutski: That makes sense, although one concern I have is that the deftest expression will balloon somewhat! 

I'm still adjusting to reading Lisp again, and in particular the Metabase flavour, but will take a look at this restructuring and see how it feels to me.

PR #37238, opened 2024-01-03
[MLv2] Fix field-values-search-info returning string ids

### LINKED ISSUE #37100: [MLv2] [BE] [Bug] fieldValuesSearchInfo issues with native queries

snoe:
Fixes #37100

Unrelated to this, the joins test is still not passing due to the search text.
image

### DIFF
--- a/src/metabase/lib/field.cljc
+++ b/src/metabase/lib/field.cljc
@@ -784,8 +784,11 @@
   widget in the QB filter modals."
   [metadata-providerable :- ::lib.schema.metadata/metadata-providerable
    column                :- ::lib.schema.metadata/column]
-  {:field-id         (:id column)
-   :search-field-id  (:id (search-field metadata-providerable column))
-   :has-field-values (if column
-                       (infer-has-field-values column)
-                       :none)})
+  (when column
+    (let [column-field-id (:id column)
+          search-field-id (:id (search-field metadata-providerable column))]
+      {:field-id (when (int? column-field-id) column-field-id)
+       :search-field-id (when (int? search-field-id) search-field-id)
+       :has-field-values (if column
+                           (infer-has-field-values column)
+                           :none)})))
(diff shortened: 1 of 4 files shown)


ranquild (approved):
@snoe I've fixed the join test

src/metabase/lib/field.cljc
```diff
@@ -784,8 +784,11 @@
...
-  {:field-id         (:id column)
-   :search-field-id  (:id (search-field metadata-providerable column))
-   :has-field-values (if column
-                       (infer-has-field-values column)
-                       :none)})
+  (when column
+    (let [column-field-id (:id column)
+          search-field-id (:id (search-field metadata-providerable column))]
+      {:field-id (when (int? column-field-id) column-field-id)
+       :search-field-id (when (int? search-field-id) search-field-id)
```
metamben, line 791: Shouldn't the `::field-values-search-info` schema reflect the fact that only integers are allowed?

metamben (approved):

src/metabase/lib/field.cljc
snoe: It does already.
`::lib.schema.id/field` -> `::common/positive-int` -> `pos-int?`

PR #37240, opened 2024-01-03
Remove other user collections defentively by adding other SQL query

### LINKED ISSUE #35652: Collection tree returns other's personal collection
### Describe the bug

The API `GET /api/colleciton/tree` return other's nested personal collection at top level.

### To Reproduce

Not really sure how to reproduce this on a fresh instance, but here is how to reproduce it on stats

1. I think you need to be a non-admin user on stats
2. create a new question
3. save it to your personal collection
4. When prompted to add the question to a dashboard, click yes
5. Now you should see the “Hosting Accounts” collection at the top-level

This "Hosting Accounts" is a nested personal collection so it should:
1. not be shown for non-admins
2. should be nested, not at the top level

Context: slack [thread](https://metaboat.slack.com/archives/C0641E4PB9B/p1699849571252549)

_No response_

```JSON
master
```

### Severity

P2?

### Additional context

_No response_

qwef:
Closes https://github.com/metabase/metabase/issues/35652

### Description

Fixes a bug on stats where a users collections is showing up on the `api/collection/tree` when it is not supposed to (due to some users collections not showing up in the permiessoned select).

This introduces another SQL query, which may make the /tree request slower (but hopefully correct)


src/metabase/api/collection.clj
```diff
@@ -46,19 +46,56 @@
...
+                         [:= :archived false])
+                       (when shallow
+                         (location-from-collection-id-clause collection-id))
+                       (when exclude-other-user-collections
+                         (exclude-other-user-collections-clause api/*current-user-id*))
+                       (perms/audit-namespace-clause :namespace namespace)
+                       (collection/visible-collection-ids->honeysql-filter-clause
+                        :id
+                        (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]
+               ;; Order NULL collection types first so that audit collections are last
```
qwef, line 95: To consolidate other SQL statements, I added the order-by found in the "/" API request to both. This shouldn't functionally change the "/tree" request, and I don't think it would introduce much latency.

noahmoss:
Looks like this also has the changes from `jh-collection-list` now that the base branch was squash-merged. Could you rebase?

(This is one of those use cases where Graphite can handle this automatically)

qwef:
> Looks like this also has the changes from `jh-collection-list` now that the base branch was squash-merged. Could you rebase?
> 
> (This is one of those use cases where Graphite can handle this automatically)

Ah yea I see, I merged master I think it should be good now.

src/metabase/api/collection.clj
```diff
@@ -46,19 +46,52 @@
...
-        personal?    (fn [{^String location :location id :id}]
-                       (or (personal-ids id)
-                           (prefixes (re-find #"^/\d+/" location))))]
-    (if (seq prefixes)
-      (remove personal? collections)
-      collections)))
+  (let [personal-ids        (t2/select-fn-set :id :model/Collection
+                                              {:where
+                                               [:and [:!= :personal_owner_id nil] [:!= :personal_owner_id user-id]]})
+        personal-descendant (comp personal-ids
```
noahmoss, line 65: nit: it's a little more readable when predicate functions end in `?`, I think ([style guide](https://guide.clojure.style/#naming-predicates))
```suggestion
        personal-descendant? (comp personal-ids
```

src/metabase/api/collection.clj
```diff
@@ -46,19 +46,52 @@
...
+   then restrict to the children and the grandchildren of the root collection. If collection-id is an an integer,
+   then restrict to that collection's parents and children."
+  [collection-id]
+  (if collection-id
+    [:and
+     [:like :location (str "%/" collection-id "/%")]
+     [:not [:like :location (str "%/" collection-id "/%/%/%")]]]
+    [:not [:like :location "/%/%/"]]))
+
 (defn- remove-other-users-personal-collections
```
noahmoss, line 60: nit
```suggestion
(defn- remove-other-users-personal-subcollections
```

noahmoss (approved):

PR #37241, opened 2024-01-03
Finally introduce the `mt/test-helpers-set-global-values!` pattern

camsaul:
#33771 sort of went off the rails when I ran into problems with changes to `report-timezone` triggering Snowflake connection pool resets, I went down a big rabbit hole there trying to get that to work in a thread-safe manner, but utilimately it was too much for one PR.

This PR adds just the new `mt/test-helpers-set-global-values!` pattern from that PR, which we concluded was a good idea in Slack discussions and elsewhere; I'll save actually making `with-temporary-setting-values` and other stuff parallel in follow-on PRs.

As a proof-of-concept application of this new pattern I updated `with-temp` to respect `test-helpers-set-global-values!`, and removed the thread-unsafe `with-temp!` and `with-ensure-with-temp-no-transaction!` and replaced them with usages of `test-helpers-set-global-values!`


qnkhuat (approved):
Thanks!

PR #37246, opened 2024-01-03
[MLv2] Support exclude date filter for filter args display name

### LINKED ISSUE #37219: [MLv2] [BE] [Bug] `filter-args-display-name` doesn't handle exclude date filters
Currently filter like "exclude -> hours -> 5 PM" is formatted as "Created At excludes the hour of 5 PM" while it should not include the column name. 

Repro test https://github.com/metabase/metabase/blob/48a882cabd4af9de5efaa77224b2514b01e3967b/frontend/src/metabase/querying/components/FilterModal/DateFilterEditor/DateFilterEditor.unit.spec.tsx#L177 

`screen.getByText("Created At excludes the hour of 5 PM")` should be a different string. Please update the test when fixing the bug.

snoe:
Fixes #37219 

It seems that this filter is largely unsupported by the FE time picker.

### DIFF
--- a/frontend/src/metabase/querying/components/FilterModal/DateFilterEditor/DateFilterEditor.unit.spec.tsx
+++ b/frontend/src/metabase/querying/components/FilterModal/DateFilterEditor/DateFilterEditor.unit.spec.tsx
@@ -194,9 +194,7 @@ describe("DateFilterEditor", () => {
         column,
         filter,
       });
-      expect(
-        screen.getByText("Created At excludes the hour of 5 PM"),
-      ).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


metamben (commented):
I find the fact that `:=` and `:!=` get displayed the same way pretty confusing. Can't we change that?

ranquild:
Agree with @metamben. Maybe it would be better to have `Excludes the hour of 5 PM` or something like that

ranquild (approved):
The PR does fix the initial issue, however the string itself can be improved.

metamben (approved):
Other than the displayed string, it LGTM.

PR #37248, opened 2024-01-04
Spin date util functions into separate lib

camsaul:
We've talked about doing this for years and years, time to finally do it I guess. Spun the date util stuff that almost never gets touched into a separate library https://github.com/metabase/second-date (actually I started this project like 4 years ago, so today mostly involved updating it and getting GH Actions and automatic deploy set up)

Removes ~1500 lines from the core codebase

This library would be hella useful to other people, so we can write some more dox and then announce it.

We have a bunch of Cljc date util+ stuff in the `metabase.shared.util.time` that we can probably move over to `second-date` in the future as well, that would be neat. But since we're actively iterating on it maybe we can do that once things settle down


.clj-kondo/config.edn
```diff
@@ -402,12 +399,16 @@
     ring.util.response                                              response
     ring.adapter.jetty9.servlet                                     servlet
     saml20-clj.core                                                 saml
+    second-date.core                                                u.date
```
camsaul, line 402: Didn't change the aliases yet to avoid too much churn

.clj-kondo/hooks/second_date/core.clj
```diff
@@ -0,0 +1,41 @@
+(ns hooks.second-date.core
+  (:require [clj-kondo.hooks-api :as hooks]))
+
+(defn- check-for-first-day-of-week-option [node]
```
camsaul, line 4: Since the new lib doesn't know about the `start-of-week` setting a few functions need you to explicitly pass it in, Kondo hook checks to make sure you do. Good to decouple this stuff anyway

modules/drivers/googleanalytics/src/metabase/driver/googleanalytics/query_processor.clj
```diff
@@ -30,10 +31,13 @@
     (:name (lib.metadata/field (qp.store/metadata-provider) id-or-name))
     id-or-name))
 
+(defn- truncate [t unit]
+  (u.date/truncate t unit {:first-day-of-week (public-settings/start-of-week)}))
+
```
bshepherdson, line 36: Duplicating this in a few places in the codebase doesn't seem ideal to me.

I think it's great to move the date helpers library out, but this feels like it still belongs in a `metabase.util.date` library. (Maybe bring that back when tweaking the aliases?)

bshepherdson (approved):

modules/drivers/googleanalytics/src/metabase/driver/googleanalytics/query_processor.clj
camsaul: haha I did want to do that but changing all the aliases was my main reason not to. I think it's worth it tho because this code is just too noisy

PR #37252, opened 2024-01-04
Migrate `databaseId`, `database` to use `databaseID` from MLv2

### LINKED ISSUE #37169: [MLv2] [FE] Migrate `databaseId`, `database` to use `databaseID` from MLv2

kamilmielnik:
Closes #37169

## Description

- In 2 places `metadata` and `query: Lib.Query` were not available, so I used `legacyQuery.question()`'s methods which use MLv2 under the hood
    - `databaseSupportsPivotTables` in `PivotTable/utils.ts` (`visualization.checkRenderable()` signature would need to change)
    - deprecated `FilterPopover`
- `StructuredQuery` & `NativeQuery` still use `databaseId` and `database` but the methods have been marked as deprecated and were prefixed with `_`


frontend/src/metabase-lib/Question.ts
```diff
@@ -1046,13 +1048,12 @@ class Question {
   }
 
   query(metadata = this._metadata): Query {
+    const databaseId = this.legacyQuery()?._databaseId() ?? null;
```
kamilmielnik, line 1051: This is the only `_database()`/`_databaseId()` usage left outside of old `Query`/`AtomicQuery`/`NativeQuery`/`StructuredQuery`.

ranquild: Maybe `const databaseId = this.datasetQuery().database`?

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -313,14 +313,19 @@ class View extends Component {
       return <LoadingAndErrorWrapper className="full-height" loading />;
     }
 
-    const isStructured = query instanceof StructuredQuery;
+    const isStructured = legacyQuery instanceof StructuredQuery;
```
ranquild, line 316: `question.isStructured()`

ranquild (approved):

frontend/src/metabase/query_builder/components/view/View.jsx
kamilmielnik: a3a3fc869c85c56e2f92bcff8e6e07dd4ef36efc

PR #37254, opened 2024-01-04
Drop the test which was mistakenly backported

uladzimirdev:
revert of https://github.com/metabase/metabase/pull/37174, it shouldn't have been backported as other filter related PRs weren't backported


lbrdnk (approved):
LGTM

PR #37258, opened 2024-01-04
Cleanup flaky tests

uladzimirdev:
### Description

As were mentioned in the thread, we don't need to specify viewport as the same default value is used

[thread](https://metaboat.slack.com/archives/C5XHN8GLW/p1704358805584909?thread_ts=1704358605.236789&cid=C5XHN8GLW)

config https://github.com/metabase/metabase/blob/master/e2e/support/config.js#L130:L131

### How to verify

- e2e tests should pass
- stress tests should pass

### Checklist


uladzimirdev:
Flaky tests do not pass stress tests

https://github.com/metabase/metabase/actions/runs/7409867159
https://github.com/metabase/metabase/actions/runs/7409850606

PR #37260, opened 2024-01-04
Move `FilterPanel` to `metabase/querying`

ranquild:
Cleanup QB and move `FilterPanel` to `metabase/querying`.


frontend/src/metabase/ui/components/overlays/Popover/Popover.styled.tsx
```diff
@@ -7,6 +7,7 @@ export const getPopoverOverrides = (): MantineThemeOverride["components"] => ({
       shadow: "md",
       withinPortal: true,
       middlewares: { shift: true, flip: true, size: true },
+      transitionProps: { duration: 0 },
```
ranquild, line 10: I'm pretty sure we want popover transitions to be consistent and not tweak it every time we use it.

frontend/src/metabase/query_builder/components/view/QuestionFilters/QuestionFilters.tsx
```diff
@@ -78,110 +58,25 @@ export function FilterHeaderToggle({
...
-  const filters = [...previousStageFilters, ...lastStageFilters];
-
-  const handleQueryChange = (nextQuery: Lib.Query) => {
-    const nextQuestion = question.setDatasetQuery(Lib.toLegacyQuery(nextQuery));
-    onQueryChange(nextQuestion.query() as LegacyQuery);
+  const handleChange = (query: Lib.Query) => {
+    updateQuestion(question._setMLv2Query(query), { run: true });
   };
 
-  if (filters.length === 0 || !expanded) {
```
ranquild, line 104: `filters.length. === 0` is handled in `shouldRender`

frontend/src/metabase/querying/components/FilterPanel/FilterPanelButton/FilterPanelButton.styled.tsx
```diff
@@ -0,0 +1,26 @@
...
+type FilterButtonProps = ButtonHTMLAttributes<HTMLButtonElement> &
+  ButtonProps & {
+    isExpanded: boolean;
+  };
+
+export const FilterButton = styled(Button)<FilterButtonProps>`
+  color: ${({ isExpanded }) => (isExpanded ? color("white") : color("filter"))};
+  background-color: ${({ isExpanded }) =>
+    isExpanded ? alpha("filter", 0.8) : alpha("filter", 0.2)};
+  transition: border 300ms linear, background 300ms linear;
```
kamilmielnik, line 16: I see this code has just been moved but what about color?

I'd just go with `all` and optimize by listing all necessary properties only if there are performance issues (but I don't think a button will suffer from those).

frontend/src/metabase/querying/components/FilterPanel/FilterPanelPopover/FilterPanelPopover.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <Popover
+      opened={isOpened}
+      position="bottom-start"
+      transitionProps={{ duration: 0 }}
+      onChange={setIsOpened}
+    >
+      <Popover.Target>
+        <FilterPill
+          onClick={() => setIsOpened(!isOpened)}
```
kamilmielnik, line 46: Technically this would have been more appropriate:

```suggestion
          onClick={() => setIsOpened(isOpened => !isOpened)}
```

frontend/src/metabase/ui/components/overlays/Popover/Popover.styled.tsx
kamilmielnik: I agree.

Why did you decide to revert this change?

kamilmielnik (commented):
Looks good :+1: 

Some tests need to be updated so I will approve after reviewing those changes.

ranquild:
@kamilmielnik the change with transitionProps broke tooltips, so I think we need to specify `transitionProps` for all popover-based components to make it work correctly. I thought that it looks a bit harder than I wanted it to be for now

frontend/src/metabase/querying/components/FilterPanel/FilterPanelButton/FilterPanelButton.styled.tsx
ranquild: It seems that we don't want `color` to be transitioned. At least it looks noticeably different if we do so. That's why I copied it as it was.

kamilmielnik (approved):

PR #37261, opened 2024-01-04
Add MB_ATTACHMENT_TABLE_ROW_LIMIT to docs

ignacio-mb:
With the default set to 20

### DIFF
--- a/docs/configuring-metabase/environment-variables.md
+++ b/docs/configuring-metabase/environment-variables.md
@@ -185,6 +185,14 @@ Since: v35.0
 
 Maximum number of async Jetty threads. If not set, then [MB_JETTY_MAXTHREADS](#mb_jetty_maxthreads) will be used, otherwise it will use the default.
 
+### `MB_ATTACHMENT_TABLE_ROW_LIMIT`
+
+Type: integer<br>
+Default: `20`<br>
+Since: v48.0
+
+Adds the ability to control the row limit for rendered tables in alerts and subscriptions.
+
 ### `MB_AUDIT_MAX_RETENTION_DAYS`
 
 Only available on Metabase [Pro](https://www.metabase.com/product/pro) and [Enterprise](https://www.metabase.com/product/enterprise) plans.<br>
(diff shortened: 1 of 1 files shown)


docs/configuring-metabase/environment-variables.md
```diff
@@ -185,6 +185,14 @@ Since: v35.0
 
 Maximum number of async Jetty threads. If not set, then [MB_JETTY_MAXTHREADS](#mb_jetty_maxthreads) will be used, otherwise it will use the default.
 
+### `MB_ATTACHMENT_TABLE_ROW_LIMIT`
+
+Type: integer<br>
+Default: `20`<br>
+Since: v48.0
```
jeff-bruemmer, line 192: > Since: v48.0

Has this env var been in since v48.0? 

In any case, we've only documented version numbers on some but not all env vars, and I don't think it's necessary to include this info as our docs are already versioned.

docs/configuring-metabase/environment-variables.md
```diff
@@ -185,6 +185,14 @@ Since: v35.0
...
+
+Adds the ability to control the row limit for rendered tables in alerts and subscriptions.
```
jeff-bruemmer, line 194:
```suggestion
Limits the number of rows Metabase will include in tables sent as attachments with dashboard subscriptions and alerts.
```
Is this more accurate?

jeff-bruemmer (changes requested):
- Is there an upper limit?
- What happens if the value is set to zero?

docs/configuring-metabase/environment-variables.md
ignacio-mb: Sorry, 48.2.

Ok! Saw it for some of them and added it because of that.

ignacio-mb:
@jeff-bruemmer I checked for 0, and it skips it and uses the default (20).
The max limit is [100](https://github.com/metabase/metabase/pull/36996/files#:~:text=%5B%22is%20limited%20to%20a%20max.%20of%20100%20rows.%22%20200%20100%5D%5D%5D)

ignacio-mb:
It's necessary to add the details of what happens with 100 and with 0 @jeff-bruemmer ?

jeff-bruemmer (approved):

PR #37273, opened 2024-01-04
Localize text for filter apply button in dashboard filters

### LINKED ISSUE #34408: "Apply" button has no translation
### Describe the bug

The new "Apply" (filter) button is not translated with the current language (french for us)

### To Reproduce

1. Go on a dashbaord
2. Edit Auto Apply filter behavior
3. Update a filter value
4. The Apply button appears but always in English

### Expected behavior

Translate it or allow label customization (EE) ? 

### Logs

_No response_

```JSON
v1.47.3
```

### Severity

low

### Additional context

I put it as a bug because I suppose that such UI component should be always translated on your process so I think that it is only a oversight. 

The option to customize the label should be a feature (I could create a dedicated issue if you agree on the idea)

Thanks

sloansparger:
Fixes #34408

### Description

Localizes text for the filter apply button in dashboard filters

### Demo
See button on bottom right of screenshots

Debug mode before:
before

Debug mode after:
after

### DIFF
--- a/frontend/src/metabase/parameters/components/FilterApplyButton/FilterApplyButton.tsx
+++ b/frontend/src/metabase/parameters/components/FilterApplyButton/FilterApplyButton.tsx
@@ -1,4 +1,5 @@
 import { useCallback } from "react";
+import { t } from "ttag";
 import { useDispatch, useSelector } from "metabase/lib/redux";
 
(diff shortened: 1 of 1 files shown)


iethree (approved):
✨

PR #37279, opened 2024-01-04
Fixing case of invalid graph.dimension

### LINKED ISSUE #37266: Slack integration failure on dashboard subscription
In [this slack thread](https://metaboat.slack.com/archives/C010L1Z4F9S/p1703762278437659) @nemanjaglumac reported that Slack rendering of [this query](https://stats.metabase.com/question/3706-core-pr-review-activity-by-week-for-the-past-3-months) produces an error.

@markbastian was able to reproduce this with the following steps:
* Add the above query to a dashboard
* Duplicate the query and add that card to the dashboard
* Send the dashboard to slack (e.g. `@mark`) immediately

Expected: Both cards render identically in slack message
Actual: First card renders with an error, second (a copy) renders just fine

This clip demonstrates the above steps:
slack

markbastian:
When trying to render a card with viz settings containing a nonexistent dimension, an error would result in `mult-x-axis-rowfn` causing a failure to render.

This fix simply replaces `map` with `keep` in `metabase.util.ui-logic/mult-x-axis-rowfn` and `metabase.util.ui-logic/mult-y-axis-rowfn` so that `nil` values are dropped if a bad dimension or metric name is provided.

The following easily shows the broken case (now fixed). Previously, an error was rendered. Now you get the chart.

```clojure
;; REPRO
(comment
  (ns tickets.37266-slack-error
    (:require
     [clojure.test :refer :all]
     [dev.render-png :as render-png]
     [metabase.models :refer [Card]]
     [metabase.pulse.render.body :as body]
     [metabase.query-processor :as qp]
     [metabase.test :as mt]
     [toucan2.core :as t2]))

  (let [busted? true]
    (mt/dataset test-data
      ;; Note that this uses orders since we have that as test data.
      (mt/with-temp [Card {base-card-id :id}
                     {:dataset_query          {:database (mt/id)
                                               :type     :query
                                                          :aggregation  [[:count]]
                                                          :filter       [:between [:field (mt/id :orders :created_at) nil] "2019-05-16" "2019-08-16"]
                                                          :breakout     [:field (mt/id :orders :created_at) {:temporal-unit :week}]}}
                      :display                :line
                      :visualization_settings {:table.pivot_column      "state"
                                               ;; This is also intentionally broken
                                               :graph.metrics           (if busted?
                                                                          ["frooby"]
                                                                          ["count"])
                                               :graph.show_trendline    false
                                               :graph.x_axis.title_text "Week"
                                               :graph.y_axis.title_text "PR Review Activity"
                                               ;; Using a missing dimension name breaks the results
                                               :graph.dimensions        (if busted?
                                                                          ["_sdc_extracted_at"]
                                                                          ["created_at"])
                                               :graph.show_values       false
                                               :table.cell_column       "count"}}]
        (render-png/render-card-to-png base-card-id)))))
```

Fixes #37266

### DIFF
--- a/test/metabase/pulse/render/body_test.clj
+++ b/test/metabase/pulse/render/body_test.clj
@@ -5,8 +5,12 @@
    [clojure.walk :as walk]
    [hiccup.core :refer [html]]
    [metabase.formatter :as formatter]
+   [metabase.models :refer [Card]]
    [metabase.pulse.render.body :as body]
-   [metabase.pulse.render.test-util :as render.tu]))
+   [metabase.pulse.render.test-util :as render.tu]
+   [metabase.query-processor :as qp]
+   [metabase.test :as mt]
+   [toucan2.core :as t2]))
 
 (use-fixtures :each
   (fn warn-possible-rebuild
@@ -38,7 +42,7 @@
     :semantic_type   nil
     :visibility_type :normal}])
 
-(def ^:private test-data
+(def ^:private example-test-data
   [[1 34.0996 "2014-04-01T08:30:00.0000" "Stout Burgers & Beers"]
    [2 34.0406 "2014-12-05T15:15:00.0000" "The Apple Pan"]
    [3 34.0474 "2014-08-01T12:45:00.0000" "The Gorbals"]])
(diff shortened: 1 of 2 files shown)


test/metabase/pulse/render/body_test.clj
```diff
@@ -38,7 +42,7 @@
     :semantic_type   nil
     :visibility_type :normal}])
 
-(def ^:private test-data
+(def ^:private example-test-data
```
markbastian, line 45: Renamed to prevent name collision with our sample datasets.

test/metabase/pulse/render/body_test.clj
```diff
@@ -730,3 +734,28 @@
...
+                                               ;; This test verifies that it now works.
+                                               :graph.dimensions ["_sdc_extracted_at"]}}]
+        (let [{:keys [dataset_query result_metadata dataset] :as card} (t2/select-one :model/Card :id card-id)
+              query-results (qp/process-query
+                              (cond-> dataset_query
+                                dataset
+                                (assoc-in [:info :metadata/dataset-metadata] result_metadata)))
+              {:keys [content]} (body/render :line :inline "UTC" card nil (:data query-results))]
+          (testing "Content is generated (rather than an exception being thrown)"
+            (is (= :div (first content)))))))))
```
adam-james-v, line 761: Does this test actually fail without the fix? 
We still render a png if the query fails, so if this just asserting that a PNG exists, it might not be catching what we're after.

adam-james-v (approved):
Looks good :)

test/metabase/pulse/render/body_test.clj
markbastian: It will error if the test fails:
image

PR #37283, opened 2024-01-04
Drop the *empty* and unused security holder dependency

andriibeee:
### Description

https://www.npmjs.com/package/yarn.lock is empty and holded by npm for security reasons. I assume it was added by mistake.


ranquild (approved):

nemanjaglumac (approved):
@andriibeee thank you so much for your contribution!

PR #37284, opened 2024-01-04
Avoid failing call to /api/model-index

### LINKED ISSUE #31671: Skip fetching a model index if there is no model
Original Slack convo: https://metaboat.slack.com/archives/C505ZNNH4/p1686918534218189

We should not call `/api/model-index` when the model editor is open and there is no model id.

## Steps
- New -> Model
- Open devtools -> network tab
- There is a failed request to `/api/model-index`

rafpaf:
Closes #31671

### Description

On master, when creating a new model index, a call is made to `/api/model-index` that returns error code 400. The reason is that there is no model id available, so the call is made without a `model_id` parameter. But the backend expects that parameter. On this branch, if there is no model id, the call is not made.

### How to verify

On master, go to **New** > **Model**, click **Use the notebook editor**, and note in the DevTools Network tab that the frontend calls `/api/model-index`, which returns error 400.

On the story branch, go through the same steps, and note in the DevTools Network tab that the frontend does not call `/api/model-index`.

### Demo

[Demo video (33 seconds, no audio)](https://www.loom.com/share/7df6c28eecb54b2dba26023fc9bde5d9?sid=299c7d74-26e9-46a8-9b3e-f7712ca46254)

### Checklist

The tests generate some warnings:
* `Warning: Failed prop type: The prop modelIndexes is marked as required in DatasetEditor, but its value is undefined` (in `DatasetEditor`)
* `Warning: Function components cannot be given refs` (in `Resizable`)
* `Warning: Can't perform a React state update on an unmounted component` (in `UnconnectedDataSelector`)

I've left these warnings in for now, but I'd be glad to try to remove them if reviewers wish.

### DIFF
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -32,7 +32,6 @@ import type {
   Parameter as ParameterObject,
   ParameterValues,
   ParameterId,
-  UnsavedCard,
   VisualizationSettings,
 } from "metabase-types/api";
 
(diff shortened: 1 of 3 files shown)


iethree (approved):
🥳

Issue #37070, opened 2023-12-22
[MLv2] [FE] `LeafletMap` - brush filter


ranquild:
~Blocked by https://github.com/metabase/metabase/issues/37103~

Blocked by https://github.com/metabase/metabase/issues/37304

Migrate this part to MLv2 https://github.com/metabase/metabase/blob/1f7e261ec4702b36aa1f815440396187c4b7e931/frontend/src/metabase/visualizations/components/LeafletMap.jsx#L161

PR #37295, opened 2024-01-05
[Leaflet] Replace brush filter with MLv2 alternative

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/37070

Blocked by https://github.com/metabase/metabase/issues/37304

### Description

Replaces old filter with MLv2 alternative

### How to verify

- rely on e2e, they must pass, especially `maps`

### Checklist


e2e/test/scenarios/visualizations-charts/maps.cy.spec.js
```diff
@@ -257,6 +257,7 @@ describe("scenarios > visualizations > maps", () => {
 
     cy.wait("@dataset");
 
+    cy.get(".CardVisualization").should("exist");
```
uladzimirdev, line 260: verifies that something is actually rendered

frontend/src/metabase/visualizations/components/LeafletMap.jsx
```diff
@@ -158,14 +158,24 @@ export default class LeafletMap extends Component {
 
     const question = new Question(card);
```
kamilmielnik, line 159: `metadata` could come in handy here as well.

kamilmielnik (approved):
Looks good :+1:

PR #34516, opened 2023-10-10
Invalid params reporting

sansolovyov:
It's two commits, but both really small (second one is literally just a documentation change).

So the problem is that `(:body request)` is coming in with keys being keywords, so `allowed-params` should contain keywords to meaningfully compare with the body. Right now it just complains about everything in the logs.

PR #37298, opened 2024-01-05
Add environ based suggestion for db config

crisptrutski:
### Description

It's nice for local config to be scoped to a specific project, but still work with non CLI created processes, e.g. via jack-in. 

Added instructions for the approach that I am using - an environ specific dot file.

Describe the overall approach and the problem being solved.

### How to verify

Follow these instructions, and verify that the associated fields are set in `metabase.db.env/env`, and that the H2 warning is no longer printed.

### DIFF
--- a/docs/developers-guide/devenv.md
+++ b/docs/developers-guide/devenv.md
@@ -180,7 +180,19 @@ You could also pass a full conection string in as the `mb.db.connection.uri`:
 "-Dmb.db.connection.uri=postgres://<user>:<password>@localhost:5432/<dbname>"
 ```
 
+Another option is to create a `.lein-env` file within your project directory:
 
+```
+{:mb-db-type   "postgres"
+ :mb-db-host   "localhost"
+ :mb-db-user   "<username>"
+ :mb-db-dbname "<dbname>"
+ :mb-db-pass   ""}
+```
+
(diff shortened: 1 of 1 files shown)


piranha:
I want to add there was [a little discussion](https://github.com/metabase/metabase/pull/34516#pullrequestreview-1667082508) about it (uncollapse a first review comment to see it).

calherries:
@qnkhuat's argument is the downside is essentially this

> I feel like we should guide ppl toward using a mechanism we explicitly support.

We can get around this downside by explicitly not supporting this mechanism in production. I would just add a note in these docs that says something along the lines of "Only use this for development. This is not supported for production use."

crisptrutski:
> I want to add there was [a little discussion](https://github.com/metabase/metabase/pull/34516#pullrequestreview-1667082508) about it (uncollapse a first review comment to see it).

Ah thanks for pointing that out. I had a brief chat with Ngoc and it sounds like he's happy if I reword things a bit to emphasize the 12 factor approaches versus an environ specific approach. I think quite a few of us are using this approach though, so I'll lay out some pros and cons.

crisptrutski:
@qnkhuat let me know if you're happy with my tweak

calherries (approved):
Looks good to go

crisptrutski:
@qnkhuat I'll leave it for you to merge if you're happy 🌻

PR #37313, opened 2024-01-05
Validate json settings during init (#36182)

### LINKED ISSUE #36182: Invalid json for env variable settings breaks site
### Describe the bug

We have a few `:json` type `defsettings` like `metabase.public-settings/custom-formatting`

An example of a well-formatted value for this is

```clojure
public-settings=> (custom-formatting)
{:type/Temporal {:time_style "HH:mm", :date_abbreviate true}}
```

We allow setting this by environmental variable like other settings. But if the json is invalid the entire application breaks.

### To Reproduce

I'm using an 0.47.5 jar:

in a terminal:

```shell
# the json is missing a closing `}`
MB_JETTY_PORT=3006 MB_CUSTOM_FORMATTING="{\"type/Temporal\":{\"time_style\":\"HH:mm\",\"date_abbreviate\":true}" java -jar 0.47.5.jar
```

From a repl to this jar we see:

```clojure
❯ netcat localhost 6006
user=> (doto 'metabase.public-settings require in-ns)
metabase.public-settings
metabase.public-settings=> (custom-formatting)
Execution error (JsonEOFException) at com.fasterxml.jackson.core.base.ParserMinimalBase/_reportInvalidEOF (ParserMinimalBase.java:696).
Unexpected end-of-input: expected close marker for Object (start marker at [Source: (StringReader); line: 1, column: 1])
 at [Source: (StringReader); line: 1, column: 63]
metabase.public-settings=>
```

And catastrophically, the application itself renders like so:

image

### Expected behavior

_No response_

### Logs

Interestingly enough, there are no logs generated in the terminal. The only stacktrace is rendered in the browser.

<details>
<summary>Stack trace in browser</summary>

{"via":[{"type":"clojure.lang.ExceptionInfo","message":"Error parsing Setting :custom-formatting: Unexpected end-of-input: expected close marker for Object (start marker at [Source: (StringReader); line: 1, column: 1])\n at [Source: (StringReader); line: 1, column: 63]","data":{"setting":"custom-formatting"},"at":["metabase.models.setting$get_raw_value$parse__46770","invoke","setting.clj",554]},{"type":"com.fasterxml.jackson.core.io.JsonEOFException","message":"Unexpected end-of-input: expected close marker for Object (start marker at [Source: (StringReader); line: 1, column: 1])\n at [Source: (StringReader); line: 1, column: 63]","at":["com.fasterxml.jackson.core.base.ParserMinimalBase","_reportInvalidEOF","ParserMinimalBase.java",696]}],"trace":[["com.fasterxml.jackson.core.base.ParserMinimalBase","_reportInvalidEOF","ParserMinimalBase.java",696],["com.fasterxml.jackson.core.base.ParserBase","_handleEOF","ParserBase.java",506],["com.fasterxml.jackson.core.base.ParserBase","_eofAsNextChar","ParserBase.java",523],["com.fasterxml.jackson.core.json.ReaderBasedJsonParser","_skipWSOrEnd","ReaderBasedJsonParser.java",2520],["com.fasterxml.jackson.core.json.ReaderBasedJsonParser","nextToken","ReaderBasedJsonParser.java",698],["cheshire.parse$parse_STAR_","invokeStatic","parse.clj",65],["cheshire.parse$parse_STAR_","invoke","parse.clj",63],["cheshire.parse$parse","invokeStatic","parse.clj",100],["cheshire.parse$parse","invoke","parse.clj",88],["cheshire.core$parse_string","invokeStatic","core.clj",208],["cheshire.core$parse_string","invoke","core.clj",194],["cheshire.core$parse_string","invokeStatic","core.clj",205],["cheshire.core$parse_string","invoke","core.clj",194],["metabase.models.setting$fn__46828$fn__46829","invoke","setting.clj",615],["metabase.models.setting$get_raw_value$parse__46770","invoke","setting.clj",551],["metabase.models.setting$get_raw_value","invokeStatic","setting.clj",558],["metabase.models.setting$get_raw_value","invoke","setting.clj",517],["metabase.models.setting$fn__46828","invokeStatic","setting.clj",615],["metabase.models.setting$fn__46828","invoke","setting.clj",613],["clojure.lang.MultiFn","invoke","MultiFn.java",234],["clojure.core$partial$fn__5908","invoke","core.clj",2641],["clojure.core$partial$fn__5908","invoke","core.clj",2640],["metabase.models.setting$get","invokeStatic","setting.clj",635],["metabase.models.setting$get","invoke","setting.clj",624],["metabase.models.setting$user_readable_values_map$fn__47016","invoke","setting.clj",1267],["clojure.core$map$fn__5931$fn__5932","invoke","core.clj",2759],["clojure.core$filter$fn__5958$fn__5959","invoke","core.clj",2823],["clojure.core.protocols$iter_reduce","invokeStatic","protocols.clj",49],["clojure.core.protocols$fn__8230","invokeStatic","protocols.clj",75],["clojure.core.protocols$fn__8230","invoke","protocols.clj",75],["clojure.core.protocols$fn__8178$G__8173__8191","invoke","protocols.clj",13],["clojure.core$transduce","invokeStatic","core.clj",6947],["clojure.core$into","invokeStatic","core.clj",6962],["clojure.core$into","invoke","core.clj",6950],["metabase.models.setting$user_readable_values_map","invokeStatic","setting.clj",1261],["metabase.models.setting$user_readable_values_map","invoke","setting.clj",1248],["metabase.server.routes.index$load_entrypoint_template","invokeStatic","index.clj",81],["metabase.server.routes.index$load_entrypoint_template","invoke","index.clj",78],["metabase.server.routes.index$entrypoint","invokeStatic","index.clj",106],["metabase.server.routes.index$entrypoint","invoke","index.clj",101],["clojure.core$partial$fn__5910","invoke","core.clj",2650],["compojure.response$fn__39631","invokeStatic","response.clj",70],["compojure.response$fn__39631","invoke","response.clj",67],["compojure.response$fn__39589$G__39584__39600","invoke","response.clj",17],["compojure.response$send","invokeStatic","response.clj",27],["compojure.response$send","invoke","response.clj",22],["compojure.core$wrap_response$fn__44405","invoke","core.clj",160],["compojure.core$wrap_route_middleware$fn__44389","invoke","core.clj",132],["compojure.core$wrap_route_info$fn__44394","invoke","core.clj",139],["compojure.core$wrap_route_matches$fn__44398","invoke","core.clj",151],["compojure.core$routes$fn__44417$f__44418","invoke","core.clj",198],["compojure.core$routes$fn__44417$f__44418$respond_SINGLEQUOTE___44419","invoke","core.clj",197],["metabase.server.routes$fn__112215$fn__112217","invoke","routes.clj",47],["compojure.core$routes$fn__44417$f__44418","invoke","core.clj",198],["compojure.core$routes$fn__44417","invoke","core.clj",200],["metabase.server.middleware.exceptions$catch_uncaught_exceptions$fn__108657","invoke","exceptions.clj",107],["metabase.server.middleware.exceptions$catch_api_exceptions$fn__108654","invoke","exceptions.clj",95],["metabase.server.middleware.log$log_api_call$fn__112502","invoke","log.clj",205],["metabase.server.middleware.browser_cookie$ensure_browser_id_cookie$fn__114646","invoke","browser_cookie.clj",42],["metabase.server.middleware.security$add_security_headers$fn__90803","invoke","security.clj",153],["metabase.server.middleware.json$wrap_json_body$fn__113202","invoke","json.clj",69],["metabase.server.middleware.offset_paging$handle_paging$fn__90827","invoke","offset_paging.clj",45],["metabase.server.middleware.json$wrap_streamed_json_response$fn__113220","invoke","json.clj",103],["ring.middleware.keyword_params$wrap_keyword_params$fn__114913","invoke","keyword_params.clj",55],["ring.middleware.params$wrap_params$fn__114932","invoke","params.clj",77],["metabase.server.middleware.misc$maybe_set_site_url$fn__57251","invoke","misc.clj",61],["metabase.server.middleware.session$reset_session_timeout$fn__71750","invoke","session.clj",457],["metabase.server.middleware.session$bind_current_user$fn__71717$fn__71718","invoke","session.clj",352],["metabase.server.middleware.session$do_with_current_user","invokeStatic","session.clj",331],["metabase.server.middleware.session$do_with_current_user","invoke","session.clj",315],["metabase.server.middleware.session$bind_current_user$fn__71717","invoke","session.clj",351],["metabase.server.middleware.session$wrap_current_user_info$fn__71698","invoke","session.clj",285],["metabase.server.middleware.session$wrap_session_id$fn__71681","invoke","session.clj",217],["metabase.server.middleware.auth$wrap_api_key$fn__100301","invoke","auth.clj",30],["ring.middleware.cookies$wrap_cookies$fn__114833","invoke","cookies.clj",216],["metabase.server.middleware.misc$add_content_type$fn__57233","invoke","misc.clj",29],["metabase.server.middleware.misc$disable_streaming_buffering$fn__57259","invoke","misc.clj",78],["ring.middleware.gzip$wrap_gzip$fn__114875","invoke","gzip.clj",86],["metabase.server.middleware.misc$bind_request$fn__57262","invoke","misc.clj",95],["metabase.server.middleware.ssl$redirect_to_https_middleware$fn__114662","invoke","ssl.clj",41],["metabase.server$async_proxy_handler$fn__79487","invoke","server.clj",77],["metabase.server.proxy$org.eclipse.jetty.server.handler.AbstractHandler$ff19274a","handle",null,-1],["org.eclipse.jetty.server.handler.StatisticsHandler","handle","StatisticsHandler.java",173],["org.eclipse.jetty.server.handler.HandlerWrapper","handle","HandlerWrapper.java",122],["org.eclipse.jetty.server.Server","handle","Server.java",563],["org.eclipse.jetty.server.HttpChannel","lambda$handle$0","HttpChannel.java",505],["org.eclipse.jetty.server.HttpChannel","dispatch","HttpChannel.java",762],["org.eclipse.jetty.server.HttpChannel","handle","HttpChannel.java",497],["org.eclipse.jetty.server.HttpConnection","onFillable","HttpConnection.java",282],["org.eclipse.jetty.io.AbstractConnection$ReadCallback","succeeded","AbstractConnection.java",314],["org.eclipse.jetty.io.FillInterest","fillable","FillInterest.java",100],["org.eclipse.jetty.io.SelectableChannelEndPoint$1","run","SelectableChannelEndPoint.java",53],["org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy","runTask","AdaptiveExecutionStrategy.java",416],["org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy","consumeTask","AdaptiveExecutionStrategy.java",385],["org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy","tryProduce","AdaptiveExecutionStrategy.java",272],["org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy","lambda$new$0","AdaptiveExecutionStrategy.java",140],["org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread","run","ReservedThreadExecutor.java",411],["org.eclipse.jetty.util.thread.QueuedThreadPool","runJob","QueuedThreadPool.java",936],["org.eclipse.jetty.util.thread.QueuedThreadPool$Runner","run","QueuedThreadPool.java",1080],["java.lang.Thread","run","Thread.java",833]],"cause":"Unexpected end-of-input: expected close marker for Object (start marker at [Source: (StringReader); line: 1, column: 1])\n at [Source: (StringReader); line: 1, column: 63]","message":"Error parsing Setting :custom-formatting: Unexpected end-of-input: expected close marker for Object (start marker at [Source: (StringReader); line: 1, column: 1])\n at [Source: (StringReader); line: 1, column: 63]","setting":"custom-formatting"}

</details>

```JSON
it's 0.47.5 jar
```

### Severity

since i'm the first reporter it seems not hit in the wild, but pretty serious bug

### Additional context

_No response_

crisptrutski:
### Description

Closes #36182 

This change adds a step during initialization to sanity check all JSON settings.

If any issues are found, the process will indicate which settings are to blame and then terminate.

### How to verify

First configure a setting which expects JSON to have an invalid value, for example use the following `.lein-env`:

```clojure
{:mb-custom-formatting "{\"type/Temporal\":{\"time_style\":\"HH:mm\",\"date_abbreviate\":true}"}
```

*This configuration is missing a closing curly brace.*

Then, attempt to start the server:

```bash
clojure -M:run
```

The process should terminate during startup, and print the following exception:

```text
2024-01-05 18:26:42,568 ERROR metabase.core :: Metabase Initialization FAILED
clojure.lang.ExceptionInfo: Invalid JSON configuration for settings: (:custom-formatting) 
  {:registered-settings (:active-users-count :admin-email :aggregated-query-row-limit ...
```

### Potential ideas for improvement

- Expose non-sensitive hints to assist operator fix the config (e.g. "expected closing brace at char 63")
- Avoid repeated fetching and parsing of the configuration values, e.g. cache and reuse these results 
- Make this mechanism extensible for other validation, e.g. can all raw values be read, is integer really positive, etc

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -97,6 +97,7 @@
    [toucan2.core :as t2])
   (:import
    (clojure.lang Keyword Symbol)
+   (com.fasterxml.jackson.core JsonParseException)
    (java.io StringWriter)
    (java.time.temporal Temporal)))
 
@@ -1342,3 +1343,22 @@
            (map (fn [[setting-name]]
                   [setting-name (get setting-name)])))
      @registered-settings)))
+
+(defn- has-invalid-json? [{:keys [name type]}]
+  (when (= :json type)
+    (try
+      ;; parse, but ignore the result
+      (json/parse-string (get-value-of-type :string name))
+      false
+      (catch JsonParseException _ true)
+      ;; something else went wrong, and it could be masking an issue with the json value
+      ;; we don't want to return false positives, but indicate an issue by returning a different falsey value
+      (catch Exception _ nil))))
+
+(defn validate-json-settings!
+  "Throws an exception if there are any settings with invalid JSON configuration.
+   Note that this will only validate settings whose defsetting forms have already been evaluated."
+  []
+  (when-let [invalid-settings (seq (map key (filter (comp has-invalid-json? val) @registered-settings)))]
+    (throw (ex-info (tru "Invalid JSON configuration for settings: {0}" invalid-settings)
+                    {:registered-settings (sort (keys @registered-settings ))}))))
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1344,21 +1344,41 @@
                   [setting-name (get setting-name)])))
      @registered-settings)))
 
-(defn- has-invalid-json? [{:keys [name type]}]
-  (when (= :json type)
-    (try
-      ;; parse, but ignore the result
-      (json/parse-string (get-value-of-type :string name))
-      false
-      (catch JsonParseException _ true)
-      ;; something else went wrong, and it could be masking an issue with the json value
-      ;; we don't want to return false positives, but indicate an issue by returning a different falsey value
-      (catch Exception _ nil))))
+(defn- set-via-env? [setting]
+  (contains? env/env (setting-env-map-name setting)))
 
-(defn validate-json-settings!
-  "Throws an exception if there are any settings with invalid JSON configuration.
-   Note that this will only validate settings whose defsetting forms have already been evaluated."
+(defmulti ^:private validate-setting-format :type)
+
+;; Treat setting as valid unless a validator has been defined for its type
+(defmethod validate-setting-format :default [_] nil)
+
+(defmethod validate-setting-format :json [{:keys [name]}]
+  (try
+    ;; attempt to parse the string to trigger any related exception
+    (doall (json/parse-string (get-value-of-type :string name)))
+    ;; there was no issue parsing the setting, so return a falsey value
+    nil
+    (catch JsonParseException e e)
+    ;; be conservative about what is returned, to avoid accidentally exposing secrets in the value, so do
+    (catch Exception e (ex-info "Unable to validate setting" {:exception-type (type e)}))))
+
+(defn- validate-setting [setting]
+  (when-let [error (validate-setting-format setting)]
+    (assoc (select-keys setting [:name :type])
+      :parse-error error
+      :env-var? (set-via-env? setting))))
+
+(defn validate-settings-formatting!
+  "Check whether there are any issues with the format of application settings, e.g. an invalid JSON string.
+
+   Note that this will only check settings whose [[defsetting]] forms have already been evaluated."
   []
-  (when-let [invalid-settings (seq (map key (filter (comp has-invalid-json? val) @registered-settings)))]
-    (throw (ex-info (trs "Invalid JSON configuration for settings: {0}" invalid-settings)
-                    {:registered-settings (sort (keys @registered-settings ))}))))
+  (doseq [invalid-setting (keep validate-setting (vals @registered-settings))]
+    (if (:env-var? invalid-setting)
+      (throw (ex-info (trs "Invalid {0} configuration for setting: {1}"
+                           #_:clj-kondo/ignore (str/upper-case (name (:type invalid-setting)))
+                           (name (:name invalid-setting)))
+                      (dissoc invalid-setting :parse-error)
+                      (:cause (:parse-error invalid-setting))))
+      (log/warn (:cause (:parse-error invalid-setting))
+                (format "Unable to parse setting %s" (:name invalid-setting))))))
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -647,7 +647,7 @@
 
 (defmethod get-value-of-type :json
   [_setting-type setting-definition-or-name]
-  (get-raw-value setting-definition-or-name coll? #(json/parse-string % true)))
+  (get-raw-value setting-definition-or-name coll? #(json/parse-string-strict % true)))
 
 (defmethod get-value-of-type :csv
   [_setting-type setting-definition-or-name]
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1383,16 +1383,17 @@
   "Test whether the value configured for a given setting can be parsed as the expected type.
    Returns an map containing the exception if an issue is encountered, or nil if the value passes validation."
   [setting]
-  (try
-    (get-value-of-type (:type setting) setting)
-    nil
-    (catch clojure.lang.ExceptionInfo e
-      (let [parse-error (or (ex-cause e) e)
-            parse-error (redact-sensitive-tokens parse-error setting)
-            env-var? (set-via-env-var? setting)]
-        (assoc (select-keys setting [:name :type])
-          :parse-error parse-error
-          :env-var? env-var?)))))
+  (when (= :json (:type setting))
+    (try
+      (get-value-of-type (:type setting) setting)
+      nil
+      (catch clojure.lang.ExceptionInfo e
+        (let [parse-error (or (ex-cause e) e)
+              parse-error (redact-sensitive-tokens parse-error setting)
+              env-var? (set-via-env-var? setting)]
+          (assoc (select-keys setting [:name :type])
+            :parse-error parse-error
+            :env-var? env-var?))))))
 
 (defn validate-settings-formatting!
   "Check whether there are any issues with the format of application settings, e.g. an invalid JSON string.
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -96,10 +96,11 @@
    [schema.core :as s]
    [toucan2.core :as t2])
   (:import
-   (clojure.lang Keyword Symbol)
-   (com.fasterxml.jackson.core JsonParseException)
-   (java.io StringWriter)
-   (java.time.temporal Temporal)))
+    (clojure.lang Keyword Symbol)
+    (com.fasterxml.jackson.core JsonParseException)
+    (com.fasterxml.jackson.core.io JsonEOFException)
+    (java.io StringWriter)
+    (java.time.temporal Temporal)))
 
 ;;; this namespace is required for side effects since it has the JSON encoder definitions for `java.time` classes and
 ;;; other things we need for `:json` settings
@@ -531,6 +532,13 @@
                       (core/get cache (setting-name setting-definition-or-name))))))]
         (not-empty v)))))
 
+(defn realize
+  "Parsing a setting may result in a lazy value. Use this to ensure we finish parsing."
+  [value]
+  (when (coll? value)
+    (dorun (map realize value)))
+  value)
+
 (defn default-value
   "Get the `:default` value of `setting-definition-or-name` if one was specified."
   [setting-definition-or-name]
@@ -571,7 +579,7 @@
   ([setting-definition-or-name pred parse-fn]
    (let [parse     (fn [v]
                      (try
-                       (parse-fn v)
+                       (realize (parse-fn v))
                        (catch Throwable e
                          (let [{setting-name :name} (resolve-setting setting-definition-or-name)]
                            (throw (ex-info (tru "Error parsing Setting {0}: {1}" setting-name (ex-message e))
@@ -1347,26 +1355,44 @@
                   [setting-name (get setting-name)])))
      @registered-settings)))
 
-(defmulti ^:private validate-setting-format :type)
+(defn- redact-parse-ex
+  "Substitute an opaque exception to ensure no sensitive information in the raw value is exposed"
+  [ex]
+  (ex-info (trs "Error of type {0} thrown while parsing a setting" (type ex))
+           {:ex-type (type ex)}))
 
-;; Treat setting as valid unless a validator has been defined for its type
-(defmethod validate-setting-format :default [_] nil)
+(defn- may-contain-raw-token? [ex setting]
+  (case (:type setting)
+    :json
+    (cond
+      (instance? JsonEOFException ex) false
+      (instance? JsonParseException ex) true
+      :else (do (log/warn ex "Unexpected exception while parsing JSON")
+                ;; err on the side of caution
+                true))
 
-(defmethod validate-setting-format :json [{:keys [name]}]
+    ;; TODO: handle the remaining formats explicitly
+    true))
+
+(defn- redact-sensitive-tokens [ex raw-value]
+  (if (may-contain-raw-token? ex raw-value)
+    (redact-parse-ex ex)
+    ex))
+
+(defn- validate-setting
+  "Test whether the value configured for a given setting can be parsed as the expected type.
+   Returns an map containing the exception if an issue is encountered, or nil if the value passes validation."
+  [setting]
   (try
-    ;; attempt to parse the string to trigger any related exception
-    (doall (json/parse-string (get-value-of-type :string name)))
-    ;; there was no issue parsing the setting, so return a falsey value
+    (get-value-of-type (:type setting) setting)
     nil
-    (catch JsonParseException e e)
-    ;; be conservative about what is returned, to avoid accidentally exposing secrets in the value, so do
-    (catch Exception e (ex-info "Unable to validate setting" {:exception-type (type e)}))))
-
-(defn- validate-setting [setting]
-  (when-let [error (validate-setting-format setting)]
-    (assoc (select-keys setting [:name :type])
-      :parse-error error
-      :env-var? (set-via-env-var? setting))))
+    (catch java.lang.Exception e
+      (let [parse-error (ex-cause e)
+            parse-error (redact-sensitive-tokens parse-error setting)
+            env-var? (set-via-env-var? setting)]
+        (assoc (select-keys setting [:name :type])
+          :parse-error parse-error
+          :env-var? env-var?)))))
 
 (defn validate-settings-formatting!
   "Check whether there are any issues with the format of application settings, e.g. an invalid JSON string.
@@ -1379,6 +1405,6 @@
                            #_:clj-kondo/ignore (str/upper-case (name (:type invalid-setting)))
                            (name (:name invalid-setting)))
                       (dissoc invalid-setting :parse-error)
-                      (:cause (:parse-error invalid-setting))))
-      (log/warn (:cause (:parse-error invalid-setting))
+                      (:parse-error invalid-setting)))
+      (log/warn (:parse-error invalid-setting)
                 (format "Unable to parse setting %s" (:name invalid-setting))))))
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1386,8 +1386,8 @@
   (try
     (get-value-of-type (:type setting) setting)
     nil
-    (catch java.lang.Exception e
-      (let [parse-error (ex-cause e)
+    (catch clojure.lang.ExceptionInfo e
+      (let [parse-error (or (ex-cause e) e)
             parse-error (redact-sensitive-tokens parse-error setting)
             env-var? (set-via-env-var? setting)]
         (assoc (select-keys setting [:name :type])
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1235,18 +1235,21 @@
       :else
       parsed-value)))
 
+(defn- set-via-env-var? [setting]
+  (some? (env-var-value setting)))
+
 (defn- user-facing-info
   [{:keys [default description], k :name, :as setting} & {:as options}]
-  (let [set-via-env-var? (boolean (env-var-value setting))]
+  (let [from-env? (set-via-env-var? setting)]
     {:key            k
      :value          (try
                        (m/mapply user-facing-value setting options)
                        (catch Throwable e
                          (log/error e (trs "Error fetching value of Setting"))))
-     :is_env_setting set-via-env-var?
+     :is_env_setting from-env?
      :env_name       (env-var-name setting)
      :description    (str (description))
-     :default        (if set-via-env-var?
+     :default        (if from-env?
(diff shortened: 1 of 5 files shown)


crisptrutski:
Not sure if this is related to my changes, or perhaps even just some thing flakey, but there was one failing test for the MySQL driver in CI:

```
 FAIL in metabase.db.schema-migrations-test/audit-v2-views-test (schema_migrations_test.clj:569)
Migrations v48.00-029 - v48.00-040 Migrations v48.00-029 thru end 
:mysql
 Views should be removed when downgrading
expected: (thrown?
           clojure.lang.ExceptionInfo
           (t2/query (str "SELECT 1 FROM " view-name)))
  actual: nil
```

calherries:
Nice start.

I think the main issues with this solution are:
(1) it's needlessly too specific to JSON, as it can be applied to all env vars 
(2) it's too general in that the validation is applied to all JSON settings and not just the ones read from environment variables.

on 1:

Dan wrote in [this comment](https://github.com/metabase/metabase/issues/36182#issuecomment-1850212704) on the option that was eventually selected by @luizarakaki:

> check on startup that all (just json?) env settings are valid according to their type and fail to start up with helpful error messages

The logic that leads one to apply this validation to JSON env settings should also apply to all env settings that require a certain format/type.

on 2:

For settings set via the admin panel, the "fail on startup" approach is not suitable. The context to fix those issues is in the admin panel where the setting has been set (after startup).

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1343,22 @@
...
+      ;; something else went wrong, and it could be masking an issue with the json value
+      ;; we don't want to return false positives, but indicate an issue by returning a different falsey value
+      (catch Exception _ nil))))
+
+(defn validate-json-settings!
+  "Throws an exception if there are any settings with invalid JSON configuration.
+   Note that this will only validate settings whose defsetting forms have already been evaluated."
+  []
+  (when-let [invalid-settings (seq (map key (filter (comp has-invalid-json? val) @registered-settings)))]
+    (throw (ex-info (tru "Invalid JSON configuration for settings: {0}" invalid-settings)
```
calherries, line 1363: `tru` means the string will be translated into the language of "the user" meaning an API user, which is only used when the message is passed to the frontend for display. See [`metabase.util.i18n/translate-user-locale`](https://github.com/metabase/metabase/blob/cd7a8c71821dae559e83fc92ba3173e7b8d6fa94/src/metabase/util/i18n.clj#L78) to understand what's going on. We use `trs` for this kind of message that is meant to be translated into the language of the metabase instance, see [this](https://github.com/metabase/metabase/blob/cd7a8c71821dae559e83fc92ba3173e7b8d6fa94/enterprise/backend/src/metabase_enterprise/advanced_config/file.clj#L211) example.
```suggestion
    (throw (ex-info (trs "Invalid JSON configuration for settings: {0}" invalid-settings)
```

calherries: As for this error message, I think we should display the exception message from the `JsonParseException`, because it gives useful information about what is invalid about the JSON string. And I'd say it's okay to raise one exception at a time to simplify the display and the code. The operator is likely to only have one setting in the incorrect format at a time anyway.

calherries:
1. Where this validation should be injected. 🔬 

LGTM

2. Where this logic should be defined. 🗺️ 

LGTM 

3. Good docstrings and error messages. ✍️ 

Left a comment 

4. What type of exception to throw, what context to put on it. 🦠 

We don't have a standard way to throw exceptions. Just throwing an exception with an appropriate error message is enough. I don't think context is required here.

5. Whether we should be explicitly handling this exception or printing it differently. 👁️ 

Throwing is sufficient to kill the process here, it LGTM

6. Code style and formatting - I don't think I've even set up Calva properly yet 🙈 

Code style is fine

8. How best to test this - e.g. do we need unit tests AND an integration test? 🔍 

Good question. Testing `settings/validate-json-settings!` with `mt/with-temp-env-var-value` should be sufficiently safe for me. If we were being super strict we should also test the validation happens at startup, but I think that's more effort than it's worth for something like this for us. This is a pretty subjective decision so others might disagree. But this is something like a P3 and I think it's unlikely we'll remove this call without realising it in a PR review.

10. Whether we should backport this. 🕥 

Yes, this is a bug-fix so it's fine to backport.

crisptrutski:
Thanks for the useful feedback.

> (1) it's needlessly too specific to JSON, as it can be applied to all env vars

I was trying to stick to the original intention of the issue. Are you happy that I make this approach at least extensible to the other types, and deal with them in a second PR? Also, should we update the ticket title and description, or create a follow-up issue?

> (2) for settings set via the admin panel, the "fail on startup" approach is not suitable.

Is it not possible that some of these settings might cause a runtime error before the user is able to edit it via the admin panel? My thought was that a start up error was still preferable to leaking a stack trace to end users, but I see your point and appreciate that this is perhaps an unlikely edge case.

How do you feel about still validating these settings, but only printing a warning, and falling back to the default value? Happy to leave this out of scope, just curious.

calherries:
> Are you happy that I make this approach at least extensible to the other types, and deal with them in a second PR? Also, should we update the ticket title and description, or create a follow-up issue?

Ah you're right, you can keep this PR focused on JSON invalidation but with an eye to extend to other types, including a TODO comment. Then create a second issue to show we're doing more work than originally intended and close them separately.

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1343,42 @@
            (map (fn [[setting-name]]
                   [setting-name (get setting-name)])))
      @registered-settings)))
+
+(defn- set-via-env? [setting]
+  (contains? env/env (setting-env-map-name setting)))
```
crisptrutski, line 1348: This will (unfortunately?) pick up `.lein-env` overrides as well. 

Unfortunately `System/getenv` can't get updated by `with-temp-env-var-value`, so we have to check the environ var directly. We *could* check against `(env/read-env-file ".lein-env")`, but that seems to over complicate things.

Since the file is "not officially supported" I think this is fine?

calherries: I think the implementation of this function is coupled to how we get environment variables and when we choose to use them, when you could use a higher level function `env-var-value` that hides these details. This also contains a very subtle bug, because there are settings that are not "site-wide", so even if they are in the environment we would not use them. `env-var-value` contains the logic for excluding non-site-wide settings, so using it will fix the bug.
```suggestion
  (some? (set-via-env-var? setting)
```

There is actually already an instance where a function like this is defined:
https://github.com/metabase/metabase/blob/b34970a2c27cb78ab09996475fab1d0a56b66a02/src/metabase/models/setting.clj#L1240
Side-note: I think this contains another bug: `boolean` should be changed to `some?`

You should also change this function to use your new `set-via-env?` function so we don't have duplicated logic.

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1343,42 @@
...
+;; Treat setting as valid unless a validator has been defined for its type
+(defmethod validate-setting-format :default [_] nil)
+
+(defmethod validate-setting-format :json [{:keys [name]}]
+  (try
+    ;; attempt to parse the string to trigger any related exception
+    (doall (json/parse-string (get-value-of-type :string name)))
+    ;; there was no issue parsing the setting, so return a falsey value
+    nil
+    (catch JsonParseException e e)
```
calherries, line 1361: There's no test for this case.

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1343,42 @@
            (map (fn [[setting-name]]
                   [setting-name (get setting-name)])))
      @registered-settings)))
+
+(defn- set-via-env? [setting]
+  (contains? env/env (setting-env-map-name setting)))
+
+(defmulti ^:private validate-setting-format :type)
```
calherries, line 1350: Even though this is private, I'd suggest adding a docstring so we know what contract each implementation needs to fulfill. Multimethods can easily become inconsistent otherwise. At the very least, you should document "Returns nil if the value passes validation"

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1343,42 @@
...
+
+   Note that this will only check settings whose [[defsetting]] forms have already been evaluated."
+  []
+  (doseq [invalid-setting (keep validate-setting (vals @registered-settings))]
+    (if (:env-var? invalid-setting)
+      (throw (ex-info (trs "Invalid {0} configuration for setting: {1}"
+                           #_:clj-kondo/ignore (str/upper-case (name (:type invalid-setting)))
+                           (name (:name invalid-setting)))
+                      (dissoc invalid-setting :parse-error)
+                      (:cause (:parse-error invalid-setting))))
```
calherries, line 1382: I see you're getting the `:cause` from the `:parse-error` value in both cases. What do you think of simplifying things and returning the thing we want displayed to the user in the `validate-setting-format` function?

src/metabase/models/setting.clj
crisptrutski: Agreed, I was just putting off writing it until I added the next type. Writing is hard, but fine I'll do it now 😄

src/metabase/models/setting.clj
crisptrutski: This is the case that's throwing in the current test, at least when I ran it in my REPL a few tweaks ago. Testing both cases is a fair point, but I'm not sure how best to trigger a non-JSON exception. If you have any ideas, let me know.

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1343,42 @@
...
+
+(defmethod validate-setting-format :json [{:keys [name]}]
+  (try
+    ;; attempt to parse the string to trigger any related exception
+    (doall (json/parse-string (get-value-of-type :string name)))
+    ;; there was no issue parsing the setting, so return a falsey value
+    nil
+    (catch JsonParseException e e)
+    ;; be conservative about what is returned, to avoid accidentally exposing secrets in the value, so do
+    (catch Exception e (ex-info "Unable to validate setting" {:exception-type (type e)}))))
```
calherries, line 1363: There's no test for this case. I'm actually wondering whether it's worth keeping around. Can you justify it?

src/metabase/models/setting.clj
crisptrutski: I went this route as it will keep things DRY as we add validation for more types, and also kept the type signature of the multimethod simpler. The latter is great as it simplifies the docstring I need to write 😆

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1354,57 @@
...
+  (if (may-contain-raw-token? ex raw-value)
+    (redact-parse-ex ex)
+    ex))
+
+(defn- validate-setting
+  "Test whether the value configured for a given setting can be parsed as the expected type.
+   Returns an map containing the exception if an issue is encountered, or nil if the value passes validation."
+  [setting]
+  (try
+    (get-value-of-type (:type setting) setting)
```
calherries, line 1387: Nice, using `get-value-of-type` here avoids coupling this function to `get-value-of-type`'s implementation.

src/metabase/models/setting.clj
```diff
@@ -1342,3 +1354,57 @@
...
+    nil
+    (catch java.lang.Exception e
+      (let [parse-error (ex-cause e)
```
calherries, line 1390: Not every exception is guaranteed to have a cause. So `parse-error` can be `nil`. I think we'll need to handle each type of exception separately, sometimes we'll want to display the message, sometimes the cause. Using the cause would be fine if we're just talking about JSON parsing exceptions, but I don't think this generalizes to all types of setting parsing exceptions.

I noticed there are also no tests for other types of settings. What about limiting this check to settings of type `:json` for now? If so, the rest of the code looks good.

src/metabase/models/setting.clj
```diff
@@ -530,6 +532,13 @@
                       (core/get cache (setting-name setting-definition-or-name))))))]
         (not-empty v)))))
 
+(defn realize
+  "Parsing a setting may result in a lazy value. Use this to ensure we finish parsing."
```
calherries, line 536: This surprised me. When can parsing a setting result in a lazy value?

crisptrutski: Yeah, it surprised me too. 

A json list, which I would expect to be parsed as a vector (matching JS semantics), returns not only a lazy list, that list is also parsed lazily. 

This seems absurd, given that you could close the stream that was parsed.

crisptrutski: Just to double check my sanity, I took the call to this function out again and confirmed that we stop throwing the exception in two of the test cases. 🤯

calherries: Nice find. It feels like that should be fixed inside the implementation of the `parse-fn`. `realize` shouldn't have to deal with this. You could update the default implementation of `get-value-of-type` `:json` to use the `json/parse-string-strict` to avoid lazily parsing

calherries: On the other hand, I'm a bit torn. Someone could easily write a getter for a setting that accidentally parses lazily without realising. maybe we should leave this.

crisptrutski: Note to self: read doc strings more often 😆

src/metabase/models/setting.clj
calherries: Ah you're right, there will always be a cause. It's not clear from the code these things are coupled like that. I think 2 helps decouple them a little, as for 1, I'm not too sure what it achieves. Maybe there's a better way to be explicit about coupling, but I don't think it's so important if the fallback is there and we have tests. There should still be tests for other types of settings though.

src/metabase/models/setting.clj
calherries: Yes!

test/metabase/models/setting_test.clj
```diff
@@ -1190,3 +1191,60 @@
...
+    (try
+      (setting/validate-settings-formatting!)
+      (throw (java.lang.RuntimeException. "This code should never be reached."))
+      (catch java.lang.Exception e e))))
+
+(defn- assert-parser-exception! [ex cause-message]
+  (is (= "Invalid JSON configuration for setting: test-json-setting" (ex-message ex)))
+  (is (= cause-message (ex-message (ex-cause ex)))))
+
+(deftest invalid-json-setting-test
```
calherries, line 1224: Last thing: since you've gone to the trouble of implementing it, we should test for other types of settings that can also have invalid values, e.g. csv, integer, and boolean

crisptrutski: Discussed offline, and leaving for a further PR

calherries (approved):
Looks good now!

PR #37330, opened 2024-01-05
Append to Model from CSV

iethree:
Part of https://github.com/metabase/metabase/issues/35614

### Description

Adds toolbar menu item  for adding data to a model.

Screen Shot 2024-01-09 at 4 50 55 PM

Updates Status Notification to reflect that we're uploading to a table and not a collection.

Screen Shot 2024-01-08 at 8 33 44 AM

Shows nice errors if your schema doesn't match

Screen Shot 2024-01-05 at 4 47 39 PM

Screen Shot 2024-01-05 at 4 47 52 PM

### How to verify

- enable CSV uploads in your metabase instance
- create a model from a CSV by uploading a csv to a collection
- open your model
- open the ... context menu and click "upload data to this model"
- select a CSV with the same schema (usually, you can upload the same CSV a second time)
- see that it adds the data
- try uploading a CSV with a different schema
- see that you get an error message explaining how the schema differs.

### Checklist


frontend/src/metabase/query_builder/components/QuestionActions.tsx
```diff
@@ -70,6 +73,23 @@ const QuestionActions = ({
   const isMetabotEnabled = useSelector(state =>
     getSetting(state, "is-metabot-enabled"),
   );
+
+  const canUpload = useSelector(state => {
+    const uploadsEnabled = getSetting(state, "uploads-enabled");
```
npfitz, line 78: Do we want to move this to `query_builder/selectors`? Doesn't really change much, but generally selector logic isn't so directly in components

frontend/src/metabase/query_builder/components/QuestionActions.tsx
```diff
@@ -215,6 +246,24 @@ const QuestionActions = ({
...
+  const handleFileUpload = (event: ChangeEvent<HTMLInputElement>) => {
+    const file = event.target.files?.[0];
+    if (file && question._card.based_on_upload) {
+      uploadFile({
+        file,
+        tableId: question._card.based_on_upload,
+      })(dispatch);
+
+      // reset the file input so that subsequent uploads of the same file trigger the change handler
+      const fileInput = document.getElementById(
```
npfitz, line 258: I think we would generally use a `ref` for this rather than `document.getElementById`. Will that work here?

frontend/src/metabase/status/components/FileUploadStatus/FileUploadStatus.tsx
```diff
@@ -49,19 +70,27 @@ const FileUploadStatusContent = ({
...
-  const { data: collection, isLoading } = useCollectionQuery({
-    id: collectionId,
+  const { value: entityInfo, loading } = useAsync(async () => {
+    if (tableId) {
+      const table = await MetabaseApi.table_get({ tableId: tableId });
+      return table;
+    }
+
+    if (collectionId) {
+      const collection = await CollectionsApi.get({ id: collectionId });
```
npfitz, line 80: I don't love sidestepping entities here, unless that's something we're doing across the application. Could we use the enabled prop for useCollectionQuery and useTableQuery instead?

npfitz (approved):
Couple suggestions, but they're all relatively minor. Approving to unblock, but if you don't mind please take a look at some of the suggestions below

frontend/src/metabase/status/components/FileUploadStatus/FileUploadStatus.tsx
iethree: 💡 thanks! I did not know about the `enabled` prop!

frontend/src/metabase/redux/uploads.ts
```diff
@@ -44,64 +41,83 @@ export const hasActiveUploads = (state: State) =>
...
+        clear();
+        return;
+      }
+
+      try {
+        const formData = new FormData();
+        formData.append("file", file);
+        formData.append("collection_id", String(collectionId));
+
+        const response = await (tableId
```
rafpaf, line 86: `tableId` can't be `0` or `''`, right?

I ask about `''` because I see that a `TableId` can be a number or a string.

frontend/src/metabase/status/components/FileUploadStatusLarge/FileUploadStatusLarge.tsx
```diff
@@ -91,16 +91,19 @@ const getName = (upload: FileUpload) => {
...
 
   if (isDone) {
-    return t`Data added to ${collection.name}`;
+    return t`Data added to ${uploadDestination.name}`;
   } else if (isError) {
-    return t`Error uploading your File`;
+    return t`Error uploading your file`;
   } else {
-    return t`Uploading data to ${collection.name} …`;
+    return t`Uploading data to ${uploadDestination.name} …`;
```
rafpaf, line 106: The space stood out to me. Why not "Uploading data to TableName..."?

rafpaf (approved):
Looks good! I had some minor questions.

frontend/src/metabase/status/components/FileUploadStatusLarge/FileUploadStatusLarge.tsx
iethree: sometimes tableId can be like `card_1234` if we're using another question as a data source 🥴

frontend/src/metabase/redux/uploads.ts
iethree: sometimes 	`tableId` can be a reference to another question when it's a query that queries another question like `card__1234`

npfitz (approved):
I think 2 of my earlier suggestions are still applicable, but otherwise it looks good to me

PR #37339, opened 2024-01-08
Sync should log something before it fetches metadata

qnkhuat:
We should log out ` STARTING: Sync metadata for redshift Database 5` before we call `(fetch-metadata/db-metadata database)` because that could be slow on big DBs


calherries (approved):
LGTM

PR #37340, opened 2024-01-08
Migrate topLevelFilters check to MLv2

ranquild:


frontend/src/metabase/query_builder/components/view/QuestionFilters/QuestionFilters.tsx
```diff
@@ -71,7 +70,7 @@ const shouldRender = ({
   queryBuilderMode === "view" &&
   question.isStructured() &&
   question.isQueryEditable() &&
-  (question.legacyQuery() as LegacyQuery).topLevelFilters().length > 0 &&
+  FilterPanel.shouldRender({ query: question.query() }) &&
```
kamilmielnik, line 73: It's not clear to me why this one condition is defined in `FilterPanel.shouldRender` and not in here.

So far this is the only `FilterPanel` usage.
I don't see a need for `FilterPanel.shouldRender` to exist.

----

Perhaps `getFilterItems` + `FilterItem` should get less-generic names and be moved to some place that encourages reusing them.

kamilmielnik (approved):

frontend/src/metabase/query_builder/components/view/QuestionFilters/QuestionFilters.tsx
ranquild: Removed `shouldRender` from `FilterPanel`. But I don't have a good understanding about whether we should drop this pattern from QB yet.

PR #37346, opened 2024-01-08
Use MLv2 to compute columns in parameter mapping

### LINKED ISSUE #37341: Use a MLv2 function to compute the list of available columns in parameter mapping

ranquild:
Closes https://github.com/metabase/metabase/issues/37341

### DIFF
--- a/frontend/src/metabase-lib/parameters/utils/filters.js
+++ b/frontend/src/metabase-lib/parameters/utils/filters.js
@@ -1,3 +1,4 @@
+import * as Lib from "metabase-lib";
 import TemplateTagVariable from "metabase-lib/variables/TemplateTagVariable";
 import {
   getParameterType,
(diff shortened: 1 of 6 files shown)


frontend/src/metabase-lib/parameters/utils/filters.js
```diff
@@ -25,6 +26,26 @@ export function fieldFilterForParameter(parameter) {
...
+    case "date":
+      return column => Lib.isDate(column);
+    case "id":
+      return column => Lib.isPrimaryKey(column) || Lib.isForeignKey(column);
+    case "category":
+      return column => Lib.isCategory(column);
+    case "location":
+      return column => Lib.isLocation(column);
+    case "number":
+      return column => Lib.isNumber(column) && !Lib.isLocation(column);
```
kamilmielnik, line 41: Do we need to check for `Lib.isLocation` if it's going to be checked before `case "number"` ?

If we do, then number could be a PK or FK as well.
Same goes for `case "string"`.

frontend/src/metabase/parameters/utils/mapping-options.js
```diff
@@ -1,30 +1,38 @@
...
+
+  return columns.map(column => {
+    const columnInfo = Lib.displayInfo(query, stageIndex, column);
+
+    return {
+      sectionName: getColumnGroupName(groupInfo),
+      name: columnInfo.displayName,
+      icon: getColumnIcon(column),
+      target: buildColumnTarget(column),
+      isForeign: columnInfo.isFromJoin || columnInfo.isImplicitlyJoinable,
```
kamilmielnik, line 33: It's a good opportunity to add JSDoc to `isImplicitlyJoinable` saying that it's `true` when column has a FK.

kamilmielnik (approved):

frontend/src/metabase-lib/parameters/utils/filters.js
ranquild: Not sure. Current checks will do the same as the old ones.

kamilmielnik: I'm pretty sure it does not matter in the body of this function.

>  Current checks will do the same as the old ones.

Previously `isCoordinate` was used for numbers, now `isLocation` is used.

ranquild: Yep, it's different but would result in the same output because of type hierarchy. E.g. Location can be Address or Coordinate, and only Coordinate are numbers.

ranquild: I did this because it would easier to follow switch logic where we previously check for `Location`

Issue #37232, opened 2024-01-03
Multi series questions won't get imported when using serialization


paoliniluis:
### Describe the bug

When using serialization, you get only 1 question when having a multi-series card on a dashboard

### To Reproduce

Create 2 questions
Add those 2 one on top of another
export
import

See that in the imported instance you get only the first question and there's no multi-series

### Expected behavior

We should get a multi-series card on import

### Logs

No logs, no exceptions, nothing

### Information about your Metabase installation

```JSON
v47.x
```

### Severity

P1

calherries:
@metabase-bot backport release-x.47.x

PR #37372, opened 2024-01-08
Serdes: tidy model lists

calherries:
WIP

This [recent issue](https://github.com/metabase/metabase/issues/37232) about not serializing `DashboardCardSeries` could have been prevented if we had checked the list of `excluded-models` to only exclude models that should actually be excluded from serialization.

This PR does a few things:
- adds comments to the `inlined-models` list explaining which model the model is inlined under
- adds comments to the `excluded-models` list explaining why each model is excluded
- changes `metabase-enterprise.models.entity-id-test/entities-not-exported` to use these lists instead of maintaining its own list


camsaul:
Is this WIP still?

calherries:
Yes

crisptrutski:
Really like all these changes from a code maintenance angle, but I'm not clear how they would have prevented the bug. Was the issue that the we added these models to the wrong list initially, or that our assumptions changed and we didn't shunt them over? Maybe I'm jumping the gun based on further changes you're going to make 😄

PR #27985, opened 2023-01-31
Rework sandboxing QP middleware to use `sandboxes` table as source of truth

noahmoss:
Closes [#27299](https://github.com/metabase/metabase/issues/27299)

This PR reworks the `row-level-restrictions` middleware to use the `sandboxes` table directly (formerly the `group_table_access_policy` table) to determine what sandboxes should be applied to a given query. It also updates the `segmented-user?` helper function accordingly.

There's some tricky logic related to figuring out whether any _other_ groups supersede a sandbox for the current user. I've done my best to explain these in the relevant doc strings.

Since this should be a no-op, I couldn't think of any new tests that would be necessary (but happy to write some if anyone has ideas). The main validation of this change though is that existing sandboxing tests continue to pass.

PR #37377, opened 2024-01-08
Incorporate sandboxes into perms graph based on the `sandboxes` table

noahmoss:
This PR adds a `defendpoint` function which updates the permissions graph to indicate that tables are sandboxed based explicitly on the entries in the `sandboxes` table, rather than relying on the paths in `permissions`.

The `sandboxes` table was made the source of truth for sandboxing enforcement in https://github.com/metabase/metabase/pull/27985 but the graph was still based entirely on the `permissions` table. This should be a no-op, because they should be consistent, but if we eventually move away from the current permission path format we'd need to make this change anyway.  

See the similar function `add-impersonations-to-permissions-graph` for comparison. Sandboxes and Impersonations are very similar: they both describe modifications to queries which happen _after_ our normal permission checks, so it makes sense for their sources of truth to be different tables than the normal `permissions` table.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/sandbox/api/permissions_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/sandbox/api/permissions_test.clj
@@ -44,12 +44,6 @@
                 :expected-perms   (fn []
                                     {:schemas {:PUBLIC {(mt/id :users) "all"}}})}
 
-               "when revoking segmented query permissions for Table"
-               {:updated-db-perms (fn []
-                                    {:native :none, :schemas {:PUBLIC {(mt/id :venues) {:read :all}}}})
-                :expected-perms   (fn []
-                                    {:schemas {:PUBLIC {(mt/id :venues) {:read "all"}}}})}
-
                "when changing permissions for DB to unrestricted access"
                {:updated-db-perms (constantly {:native :none, :schemas :all})
                 :expected-perms   (constantly {:schemas "all"})}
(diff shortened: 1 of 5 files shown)


noahmoss:
* `master`
  * **PR #37377** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37377?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37377?utm_source=stack-comment).

enterprise/backend/test/metabase_enterprise/sandbox/api/permissions_test.clj
```diff
@@ -44,12 +44,6 @@
                 :expected-perms   (fn []
                                     {:schemas {:PUBLIC {(mt/id :users) "all"}}})}
 
-               "when revoking segmented query permissions for Table"
-               {:updated-db-perms (fn []
-                                    {:native :none, :schemas {:PUBLIC {(mt/id :venues) {:read :all}}}})
-                :expected-perms   (fn []
-                                    {:schemas {:PUBLIC {(mt/id :venues) {:read "all"}}}})}
```
noahmoss, line 51: Removed this test case because this isn't how sandboxed perms work anymore, at least not via the UI. "Segmented query perms" are just normal data access perms + an entry in the `sandboxes` table. 

There still are `/query/`, `/read/`, and `/query/segmented/` paths hanging around in app DBs, but we're not doing any enforcement based off of them. And they'll be removed whenever we migrate to a new schema.

escherize (approved):
It'll be nice to get away from having multiple ways to store (& ignore) the sandboxed perms. 👍

PR #37381, opened 2024-01-08
Fix collection/tree giving 500 error when there is only one user

qwef:
### Description

`t2/select-fn-set` returns nil when there is the set is empty. This means when we pass collections through `personal-descendant?`,  then we get hit by a NPE.

See: https://metaboat.slack.com/archives/C010L1Z4F9S/p1704702599756839

### DIFF
--- a/src/metabase/api/collection.clj
+++ b/src/metabase/api/collection.clj
@@ -59,16 +59,16 @@
 
 (defn- remove-other-users-personal-subcollections
   [user-id collections]
-  (let [personal-ids         (t2/select-fn-set :id :model/Collection
-                                               {:where
-                                                [:and [:!= :personal_owner_id nil] [:!= :personal_owner_id user-id]]})
-        personal-descendant? (comp personal-ids
-                                   first
-                                   collection/location-path->ids
-                                   :location)]
-    (if (nil? personal-ids)
-      collections
-      (remove personal-descendant? collections))))
+  (let [personal-ids         (set (t2/select-fn-set :id :model/Collection
+                                                    {:where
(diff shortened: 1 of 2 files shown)


src/metabase/api/collection.clj
```diff
@@ -59,14 +59,16 @@
 
 (defn- remove-other-users-personal-subcollections
   [user-id collections]
-  (let [personal-ids        (t2/select-fn-set :id :model/Collection
-                                              {:where
-                                               [:and [:!= :personal_owner_id nil] [:!= :personal_owner_id user-id]]})
+  (let [personal-ids         (t2/select-fn-set :id :model/Collection
+                                               {:where
+                                                [:and [:!= :personal_owner_id nil] [:!= :personal_owner_id user-id]]})
         personal-descendant? (comp personal-ids
```
noahmoss, line 65: Another option could just be to call `set` on `personal-ids` to coerce `nil` into an empty set.

TBH — I'm not a huge fan of using `comp` to build up a function like this when it has more than 2-3 steps, since I think it's a little less readable than just using an anonymous function and maybe a `let` binding to describe what's happening. Something like
```
(fn [collection]
  (let [first-parent-collection-id ...]
    ((set personal-ids) first-parent-collection-id)))
```

Won't block the PR on this since it's a style thing but conciseness != readability

noahmoss (approved):

src/metabase/api/collection.clj
noahmoss: Or even just a `->` macro within the anonymous function so at least you can read the steps top down instead of bottom up 🤷‍♂️ but again, a style thing

escherize (approved):

PR #37401, opened 2024-01-08
Remove legacyQuery from DashCardMenu

ranquild:
`isQueryEditable` handles missing query (e.g. when you don't have data permissions) under the hood.


nemanjaglumac (approved):
Nice catch.

kamilmielnik (approved):

PR #29233, opened 2023-03-15
Properly return parameters for card in public and embedding endpoints

qnkhuat:
Modify public and embedding endpoints to return parameters on cards.

The reason we didn't return this before is that before 46, all parameter information is available in `card.dataset_query.template_tags`, so FE could infer from it without needing to get `card.parameters`.

in 46 we introduced custom values for parameters, and there is information such as `values_source_type` that is only stored in `card.parameters`, not `template-tags`. so we need to hydrate it so FE can behave properly.

## Screenshots

Setup:
- a native query `select * from products where category = {{category}}` with custom source for `category` is the `products.category` column.
- enable public/embed dashboard

Before
notice the `category` parameter is a input box, not a dropdown
Screenshot 2023-03-17 at 14 12 22

After:
now it's a dropdown.
Screenshot 2023-03-17 at 14 11 49

- - -
This change is [Reviewable](https://reviewable.io/reviews/metabase/metabase/29233)

PR #37405, opened 2024-01-08
Exclude snippet and card template tags from parameters returned from BE endpoints

### LINKED ISSUE #37270: Public native questions with snippets don't work
### Describe the bug

A native question that contains a snippet always returns errors when made public.

### To Reproduce

1. Enable public sharing in Admin settings
2. New -> SQL query
3. Create a native query snippet `CATEGORY = 'Gadget'` with name `IsGadget`
4. Set query to `SELECT * FROM ORDERS WHERE {{snippet:IsGadget}}`
5. Save the question
6. Share -> public question -> open the link
7. See the error

Screenshot 2024-01-04 at 15 53 17
Screenshot 2024-01-04 at 15 53 25

### Expected behavior

I should see question results

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.16+8-LTS",
    "java.version": "11.0.16",
    "java.vm.version": "11.0.16+8-LTS",
    "os.name": "Mac OS X",
    "os.version": "12.1",
    "user.language": "en",
    "user.timezone": "Asia/Jerusalem"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2024-01-04",
      "src_hash": "f566ffd0966ca096ad05e3d0a080a917e9932288",
      "tag": "v0.48.1-SNAPSHOT",
      "hash": "3d45805"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

P1

### Additional context

_No response_

calherries:
Fixes #37270

### Explanation of the bug

When snippet or card template tags are used in a native SQL question, the incorrect list of parameters is returned by the endpoint GET `/api/public/card`. There should be no parameters for snippet or card template tags.

example screenshot:
Screenshot 2024-01-05 at 15 46 57

The problem is related to the use of `metabase.models.card/template-tag-parameters`. It was introduced in this PR: [Properly return parameters for card in public and embedding endpoints](https://github.com/metabase/metabase/pull/29233#top).

The BE should mirror the FE and use the logic in `getTemplateTagsForParameters` [here](https://github.com/metabase/metabase/blob/67b3b1454eb85ec33e834c99a9107308307c78e4/frontend/src/metabase-lib/parameters/utils/template-tags.ts#L69) which excludes snippets from the result of `getTemplateTagParametersFromCard`.

### Summary of changes

- FE: 
  - refactoring to match the BE halfway. Previously the logic was split between two functions but now both the BE and FE have it in one
  - added e2e tests for (a) snippets in public questions and (b) card template tags in public questions
- BE: 
  - updated to match the logic on the FE, excluding template tags of type "snippet" and "card"

### What's not included 

`template-tag-parameters` contains duplicated logic with `metabase.query-processor.card/card-template-tag-parameters`. These should probably be merged into a MLv2 function so that logic is not duplicated in different parts of the FE and BE.

### DIFF
--- a/e2e/test/scenarios/sharing/public-question.cy.spec.js
+++ b/e2e/test/scenarios/sharing/public-question.cy.spec.js
@@ -2,10 +2,13 @@ import {
   restore,
   filterWidget,
   visitQuestion,
+  saveQuestion,
   downloadAndAssert,
   assertSheetRowsCount,
   openNewPublicLinkDropdown,
   createPublicQuestionLink,
+  modal,
+  openNativeEditor,
 } from "e2e/support/helpers";
 import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
 
@@ -56,14 +59,10 @@ describe("scenarios > public > question", () => {
     cy.signInAsAdmin();
 
     cy.request("PUT", "/api/setting/enable-public-sharing", { value: true });
-
-    cy.createNativeQuestion(questionData).then(({ body: { id } }) => {
-      cy.wrap(id).as("questionId");
-    });
   });
 
   it("adds filters to url as get params and renders the results correctly (metabase#7120, metabase#17033, metabase#21993)", () => {
-    cy.get("@questionId").then(id => {
+    cy.createNativeQuestion(questionData).then(({ body: { id } }) => {
       visitQuestion(id);
 
       // Make sure metadata fully loaded before we continue
@@ -97,24 +96,21 @@ describe("scenarios > public > question", () => {
   });
 
   it("should only allow non-admin users to see a public link if one has already been created", () => {
-    cy.get("@questionId").then(id => {
+    cy.createNativeQuestion(questionData).then(({ body: { id } }) => {
       createPublicQuestionLink(id);
       cy.signOut();
-    });
-
-    cy.signInAsNormalUser().then(() => {
-      cy.get("@questionId").then(id => {
+      cy.signInAsNormalUser().then(() => {
         visitQuestion(id);
-      });
 
-      cy.icon("share").click();
+        cy.icon("share").click();
 
-      cy.findByTestId("public-link-popover-content").within(() => {
-        cy.findByText("Public link").should("be.visible");
-        cy.findByTestId("public-link-input").then($input =>
-          expect($input.val()).to.match(PUBLIC_QUESTION_REGEX),
-        );
-        cy.findByText("Remove public URL").should("not.exist");
+        cy.findByTestId("public-link-popover-content").within(() => {
+          cy.findByText("Public link").should("be.visible");
+          cy.findByTestId("public-link-input").then($input =>
+            expect($input.val()).to.match(PUBLIC_QUESTION_REGEX),
+          );
+          cy.findByText("Remove public URL").should("not.exist");
+        });
       });
     });
   });
@@ -122,7 +118,7 @@ describe("scenarios > public > question", () => {
   Object.entries(USERS).map(([userType, setUser]) =>
     describe(`${userType}`, () => {
       it(`should be able to view public questions`, () => {
-        cy.get("@questionId").then(id => {
+        cy.createNativeQuestion(questionData).then(({ body: { id } }) => {
           cy.request("POST", `/api/card/${id}/public_link`).then(
             ({ body: { uuid } }) => {
               setUser();
@@ -140,6 +136,64 @@ describe("scenarios > public > question", () => {
       });
     }),
   );
+
+  it("should be able to view public questions with snippets", () => {
+    cy.visit("/");
+
+    openNativeEditor();
+
+    // Create a snippet
+    cy.icon("snippet").click();
+    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
+    cy.findByText("Create a snippet").click();
+
+    modal().within(() => {
+      cy.findByLabelText("Enter some SQL here so you can reuse it later").type(
+        `'test'`,
+      );
+      cy.findByLabelText("Give your snippet a name").type("string 'test'");
+      cy.findByText("Save").click();
+    });
+
+    cy.get("@editor").type(`{moveToStart}select `);
+
+    saveQuestion("test question", { wrapId: true });
+    cy.get("@questionId").then(id => {
+      cy.request("POST", `/api/card/${id}/public_link`).then(
+        ({ body: { uuid } }) => {
+          cy.visit(`/public/question/${uuid}`);
+          cy.get(".cellData").contains("test");
+        },
+      );
+    });
+  });
+
+  it("should be able to view public questions with card template tags", () => {
+    cy.createNativeQuestion({
+      name: "Nested Question",
+      native: {
+        query: "SELECT * FROM PEOPLE",
+      },
+    }).then(({ body: { id } }) => {
+      cy.visit("/");
+
+      openNativeEditor();
+
+      cy.get("@editor")
+        .type("select * from {{#")
+        .type(`{leftarrow}{leftarrow}${id}`);
+
+      saveQuestion("test question", { wrapId: true });
+      cy.get("@questionId").then(id => {
+        cy.request("POST", `/api/card/${id}/public_link`).then(
+          ({ body: { uuid } }) => {
+            cy.visit(`/public/question/${uuid}`);
+            cy.get(".cellData").contains("Winner");
+          },
+        );
+      });
+    });
+  });
 });
 
 const visitPublicURL = () => {
--- a/e2e/test/scenarios/sharing/public-question.cy.spec.js
+++ b/e2e/test/scenarios/sharing/public-question.cy.spec.js
@@ -144,8 +144,7 @@ describe("scenarios > public > question", () => {
 
     // Create a snippet
     cy.icon("snippet").click();
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("Create a snippet").click();
+    cy.findByTestId("sidebar-content").findByText("Create a snippet").click();
 
     modal().within(() => {
       cy.findByLabelText("Enter some SQL here so you can reuse it later").type(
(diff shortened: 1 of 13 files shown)


calherries:
@metabase/querying-components This fix includes a refactoring of `frontend/src/metabase-lib/parameters/utils/template-tags.ts` to match the BE code, but I realised that these functions on the FE and BE should probably be replaced by MLv2 functions. Can you take a look and confirm whether continuing to use these functions makes sense, or should I reach for MLv2?

test/metabase/models/card_test.clj
```diff
@@ -327,6 +328,35 @@
        (is (= expected
               (t2/select-one-fn :visualization_settings :model/Card :id (u/the-id card))))))))
 
+(deftest template-tag-parameters-test
+  (testing "Card with a Field filter parameter"
+    (t2.with-temp/with-temp [:model/Card card {:dataset_query (qp.card-test/field-filter-query)}]
```
qnkhuat, line 333:
```suggestion
    (mt/with-temp [:model/Card card {:dataset_query (qp.card-test/field-filter-query)}]
```'

we should use `mt/with-temp` moving forward. It makes future refactoring easier in case we want to change how with-temp works internally.

File: frontend/src/metabase-lib/parameters/utils/template-tags.unit.spec.js [hunk 2]
```diff
@@ -161,6 +168,38 @@ describe("parameters/utils/cards", () => {
...
+          "card-id": 123,
+        }),
+        createMockTemplateTag({
+          type: "snippet",
+          id: "3",
+          "snippet-id": 1,
+          "snippet-name": "C",
+        }),
+      ];
+      expect(getTemplateTagParameters(tags)).toEqual([
```
qnkhuat, line 192: should we write this test using negation? I think it's easier to understand what we want to test here if we write this test as  "template tag does not contains snippet" instead of using equal here.

qnkhuat (approved):
BE lgtm

frontend/src/metabase-lib/parameters/utils/template-tags.ts
```diff
@@ -60,25 +60,21 @@ export function getTemplateTagParameters(
     .filter(
       tag =>
         tag.type != null &&
+        tag.type !== "card" &&
```
kamilmielnik, line 63: I think the comment for this function needs to be updated (lines 51-52), at least the path where the clojure function is defined. Is the rest still relevant?

kamilmielnik (commented):
Can you please add an e2e test for it?

test/metabase/models/card_test.clj
calherries: Done a1e7cb5ad37a4a22a2fc2027de2498ee72cbe4f5

frontend/src/metabase-lib/parameters/utils/template-tags.unit.spec.js
calherries: Writing this test with negation is less safe I think. There are many changes I could make to code that would break the functionality but keep the test passing if it were written with negation. The same is not true with the way this test is currently written.

For example, if I wrote this test with negation in clojure, it might be:
```
(is (empty (set/intersection #{:snippet :card) (map :type (get-template-tag-parameters tags))
```
If the way `:type` is represented on a `template-tag` changes, e.g. `{:type :snippet}` changes to `{:type "snippet"}` then this test would still pass even though `get-template-tag-parameters` can be completely broken.

As for understanding, I think the test description is enough to understand pretty quickly what is going on.

frontend/src/metabase-lib/parameters/utils/template-tags.ts
calherries: Good catch, I've updated the comment. The rest is still relevant of course

calherries:
@kamilmielnik Can you explain why this warrants an e2e test? Maybe a BE test on the endpoint would suffice?

kamilmielnik:
> To me the BE endpoint is simply wrong, this has nothing to do with the FE.
> Maybe a BE test on the endpoint would suffice?

@calherries Fair enough, a BE-only reproduction test should suffice :+1:

ranquild (commented):
It would be great to add an E2E test for the original issue

calherries:
@ranquild why?

ranquild:
@calherries because if it broke like that, it means we don't have a single e2e test for public questions with snippets

calherries:
@qnkhuat can you give this a re-review? I've added some BE tests

qnkhuat (approved):
still good 👍

calherries:
@metabase/querying-components I've added e2e tests for using snippets and card template tags in public questions. Can you review?

e2e/test/scenarios/sharing/public-question.cy.spec.js
```diff
@@ -140,6 +136,64 @@ describe("scenarios > public > question", () => {
       });
     }),
   );
+
+  it("should be able to view public questions with snippets", () => {
+    cy.visit("/");
+
+    openNativeEditor();
```
kamilmielnik, line 143: `openNativeEditor` already does `cy.visit("/")` under the hood

```suggestion
    openNativeEditor();
```

e2e/test/scenarios/sharing/public-question.cy.spec.js
```diff
@@ -140,6 +136,64 @@ describe("scenarios > public > question", () => {
...
+        },
+      );
+    });
+  });
+
+  it("should be able to view public questions with card template tags", () => {
+    cy.createNativeQuestion({
+      name: "Nested Question",
+      native: {
+        query: "SELECT * FROM PEOPLE",
```
kamilmielnik, line 174: Our e2e tests will run a bit faster (which pays out in the long run) if we limit the amount of data exchanged in API requests:

```suggestion
        query: "SELECT * FROM PEOPLE LIMIT 5",
```

kamilmielnik (approved):
LGTM :+1:

e2e/test/scenarios/sharing/public-question.cy.spec.js
```diff
@@ -140,6 +136,62 @@ describe("scenarios > public > question", () => {
...
+      );
+      cy.findByLabelText("Give your snippet a name").type("string 'test'");
+      cy.findByText("Save").click();
+    });
+
+    cy.get("@editor").type(`{moveToStart}select `);
+
+    saveQuestion("test question", { wrapId: true });
+    cy.get("@questionId").then(id => {
+      cy.request("POST", `/api/card/${id}/public_link`).then(
```
WiNloSt, line 159: we could use `createPublicQuestionLink` which is already imported in this file. This applies to another test too.

e2e/test/scenarios/sharing/public-question.cy.spec.js
```diff
@@ -140,6 +136,62 @@ describe("scenarios > public > question", () => {
...
+        ({ body: { uuid } }) => {
+          cy.visit(`/public/question/${uuid}`);
```
WiNloSt, line 161: I'd sign out and call `cy.signInAsNormalUser()` like other tests to ensure the admin session doesn't have any effect on the public question. Well, it shouldn't anyway, but I'd still do that for consistency with other tests.

frontend/src/metabase-lib/parameters/utils/template-tags.ts
```diff
@@ -60,25 +60,21 @@ export function getTemplateTagParameters(
...
     .map(tag => getTemplateTagParameter(tag, parametersById[tag.id]));
 }
 
-export function getTemplateTagsForParameters(card: Card) {
-  const templateTags: TemplateTag[] =
-    card &&
+export function getTemplateTags(card: Card): TemplateTag[] {
+  return card &&
     card.dataset_query &&
     card.dataset_query.type === "native" &&
```
uladzimirdev, line 74: nit. `card` existence is checked above, dataset_query always should persist on card, but we can stay on the safe side and use optional chaining
```suggestion
  return card.dataset_query?.type === "native" &&
```

uladzimirdev (approved):
LGTM

frontend/src/metabase-lib/parameters/utils/template-tags.ts
uladzimirdev: not only this (I don't trust types I didn't write 😄 ) but also at `getParametersFromCard` there is an explicit check about card 

```
 if (!card) {
    return [];
  }
```

calherries: Right, but `getTemplateTagParametersFromCard` is called by three functions other than `getParametersFromCard`, so that's not sufficient.

calherries: Yeah I think I'll use `card?.dataset_query?.type === "native"` and leave it at that. I'm not confident in the change to remove the runtime check

calherries: Updated in [0d03b3d](https://github.com/metabase/metabase/pull/37405/commits/0d03b3d156cb3369e0d3f53a7f1221e00bc1e7c8)

e2e/test/scenarios/sharing/public-question.cy.spec.js
calherries: Done [0d03b3d](https://github.com/metabase/metabase/pull/37405/commits/0d03b3d156cb3369e0d3f53a7f1221e00bc1e7c8)

e2e/test/scenarios/sharing/public-question.cy.spec.js
calherries: Done [0d03b3d](https://github.com/metabase/metabase/pull/37405/commits/0d03b3d156cb3369e0d3f53a7f1221e00bc1e7c8)

PR #37409, opened 2024-01-08
Support replacing expressions changing types and names

### LINKED ISSUE #36211: [MLv2] [BE] Support changing expressions in the query w/o removing parts depending on the original

metamben:
Fixes #36211.

The idea is that we optimistically replace the target expression with the new one and if there are schema errors, we remove the offending top-level clauses. If the error cannot be fixed this way, we fall back to the old way, i.e. remove all top-level clauses depending on the target expression.

I haven't addressed the problem of updating the automatically generated join aliases, which means that queries can end up with quite meaningless join aliases.

This PR also resolves an old TODO item: it enforces that the various arg types of comparison operators have to agree, e.g. it prevents nonsense like `[:< {} <date> <integer>]`.

### DIFF
--- a/src/metabase/lib/remove_replace.cljc
+++ b/src/metabase/lib/remove_replace.cljc
@@ -14,7 +14,6 @@
    [metabase.lib.ref :as lib.ref]
    [metabase.lib.util :as lib.util]
    [metabase.mbql.util.match :as mbql.match]
-   [metabase.types :as types]
    [metabase.util :as u]
    [metabase.util.malli :as mu]))
 
@@ -236,9 +235,10 @@
 
 (defn- local-replace
   [stage target replacement]
-  (if (lib.util/expression-name target)
-    (local-replace-expression stage target replacement)
-    (walk/postwalk #(if (= % target) replacement %) stage)))
+  (->> (if (lib.util/expression-name target)
+         (local-replace-expression stage target replacement)
+         (walk/postwalk #(if (= % target) replacement %) stage))
+       (walk/postwalk #(if (= % (lib.options/uuid target)) (lib.options/uuid replacement) %))))
 
 (defn- returned-columns-at-stage
   [query stage-number]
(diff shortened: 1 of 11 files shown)


src/metabase/lib/remove_replace.cljc
```diff
@@ -304,6 +309,58 @@
...
+        (if (seq error-paths)
+          (recur (reduce (fn [q path]
+                           (try
+                             (remove-clause q (second path) (get-in q path))
+                             (catch #?(:clj Exception :cljs js/Error) e
+                               (let [{:keys [error join]} (ex-data e)]
+                                 (if (= error :metabase.lib.util/cannot-remove-final-join-condition)
+                                   ;; remove the dangling join
+                                   (remove-join q (second path) join)
+                                   (throw e))))))
```
metamben, line 355: This is the third place we play this trick. I'm not sure it's really worth preventing the cascading removal.

src/metabase/lib/schema/filter.cljc
```diff
@@ -20,21 +36,21 @@
...
-  (mbql-clause/define-tuple-mbql-clause op :- :type/Boolean
+  (mbql-clause/define-mbql-clause-with-schema-fn (tuple-clause-of-comparables-schema #{[0 1]})
+    op :- :type/Boolean
     #_x [:ref ::expression/orderable]
     #_y [:ref ::expression/orderable]))
 
-(mbql-clause/define-tuple-mbql-clause :between :- :type/Boolean
-  ;; TODO -- we should probably enforce additional constraints that the various arg types have to agree, e.g. it makes
-  ;; no sense to say something like `[:between {} <date> <[:ref ::expression/string]> <integer>]`
-  ;;
```
metamben, line 30: This is the TODO fixed in addition to the main issue.

ranquild (approved):

ranquild:
I did a very brief testing and at least basic cases work correctly 🎉

camsaul (approved):

PR #37038, opened 2023-12-21
Trend chart improvements

kulyk:
Resolves #33411

Various improvements to trend charts:

- #35184
- #36666 
- #36951 
- #36992 
- #37087 
- #37176 
- #37418

PR #34820, opened 2023-10-18
create unified BE path for isomorphic visualizations

alxnddr:
### Description

First static visualizations that used javascript for rendering were incomplete duplicates which tried to reproduce the logic of the dynamic visualizations. The logic was spread between Clojure and Javascript. This approach has multiple downsides because all viz features have to be implemented twice and the static chart versions were far behind.

With the new [technical vision](https://www.notion.so/metabase/Viz-Revamp-Research-and-Planning-4198b2726ac64f1d89ad1de6187532d2) for visualizations we will be using the same code both for dynamic and static charts. This PR creates a single BE rendering path for isomorphic visualizations and removes code of the BE code of line/area/bar/combo static charts (FE will be removed further).

### How to verify

1. Create a dashboard with line/area/bar/combo charts and two combined scalars on a single card.
2. Send a test subscription
3. Ensure the placeholders are being rendered

### Checklist

- [ ] Tests have been added/updated to cover changes in this PR

PR #37418, opened 2024-01-08
Static viz path for the updated trend chart

alxnddr:
Related to https://github.com/metabase/metabase/pull/37038

### Description

Creates a static visualization rendering path with a data contract similar to the application's data visualization contract to ensure maximum reuse of settings computation and actual viz rendering code.

In the scope of ECharts migration we did a similar change [here](https://github.com/metabase/metabase/pull/34820/commits/3f1ad8d6f1d9b5298a25f5533ab684ae835b7743#diff-98aa068e0239e92cdc383ef4a069b7563a7aed3e842509f83c24b9561b3260af) which we will rebase on top of the current PR change once it is merged. In the referred commit the rendering method is called `isomorphic` which was not the most precise name and `javascript_visualization` makes the code a bit more obvious.

### How to verify

1. Create a Trend chart
2. Send a test dashboard subscription with it
3. Ensure it renders the HTML placeholder

### Demo

Screenshot 2024-01-08 at 10 49 12 PM

### Checklist


alxnddr:
@markbastian can you please help with BE CI checks?

frontend/src/metabase/static-viz/components/StaticChart/StaticChart.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const Placeholder = ({ text }: { text: string }) => {
+  return <div style={{ width: "300px", height: "200px" }}>{text}</div>;
+};
+
+export interface StaticChartProps {
+  rawSeries: RawSeries;
+  dashcardSettings: VisualizationSettings;
+  renderingContext: RenderingContext;
+}
```
kulyk, line 11: I belive every static chart component would need that type, right? Would it make sense to pull it up? I was thinking of `metabase/static-viz/types` or maybe `metabase/visualization/types`?

kulyk: Also, it seems like we're missing a proper type for `RenderingContext`?

alxnddr: Good catch, thanks! Fixed in the latest commit

kulyk (approved):
FE LGTM 👍

markbastian (approved):
I fixed a breaking BE issue I found and did a small amount of cleanup so returned `:style` is a keyword. 🚢 🇮🇹

PR #37421, opened 2024-01-08
Multiple Cypress Reporters

### LINKED ISSUE #37195: Set up Multiple Cypress Reporters
**Context**

https://github.com/metabase/metabase/pull/35929 added the ability for us to upload junit test reports to AWS, it changed the default test reporter from mochaAwesome to junit in cypress.

One side effect of this is that the junit console output is pretty darn ugly and noisy, another downside is that we no longer get the pretty mochaAwesome test result reporter results.

Several people have complained about this, and it now seems worthwhile to [set up multiple reporters](https://docs.cypress.io/guides/tooling/reporters#Multiple-reporters) for cypress to get pretty console output as well as generate junit xml files for upload.

markbastian:
This PR modifies our Cypress reporter to support multiple reporters, currently pretty mochawesome reports and junit xml files. This was enabled using the [cypress-multi-reporters](https://www.npmjs.com/package/cypress-multi-reporters) reporter. Additional reporters are added via yarn packages:
- [mocha-junit-reporter](https://www.npmjs.com/package/mocha-junit-reporter)
- [mochawesome](https://www.npmjs.com/package/mochawesome) which was already installed

One of the biggest challenges here is simply getting the config keys right, primarily `mochaJunitReporterReporterOptions` as opposed to the more sane `mochaJunitReporterOptions`, but I guess that the legacy of Java (☕). Also be aware that `mochawesomeReporterOptions` is NOT the same as `mochaAwesomeReporterOptions` but `mochaJunitReporterReporterOptions` is not `mochJunitReporterReporterOptions` for obvious reasons.

For future reference, note that multiple `mochawesome` reporters are available. Read more [here](https://glebbahmutov.com/blog/the-awesome-battle/) in the event that one of them is mochawesomer than the other and we want to level up.

Closes #37195


package.json
```diff
@@ -240,7 +241,6 @@
     "jest": "^29.5.0",
     "jest-canvas-mock": "^2.4.0",
     "jest-environment-jsdom": "^29.5.0",
-    "jest-junit": "^16.0.0",
```
iethree, line 243: I think this will break junit reports for frontend unit tests 🤔

markbastian: IDK exactly what that does, but I was thinking since we were switching to a different reporter I wasn't sure if `jest-junit` was still needed. Easy enough to put the `junit-jest` changes back in.

nemanjaglumac (approved):
I tried it locally with one intentionally failing test. Works like a charm.

Also compared the changes from this file's history, and everything looks good.
Thanks for doing this.

PR #37428, opened 2024-01-09
Apply selective malli mu.fn enforcement for certain namespaces to improve performance

### LINKED ISSUE #37144: CPU consumption is very high after upgrading to v0.48.1
### Type: 
Metabase (Running on Docker)
### Version: 
v0.48.1
### Bug Description:
After upgrading from v0.47.8 to v0.48.1, the CPU consumption of server becomes higher, at the same time, some cards and dashboards  load  much more slowly than ever before. It never happened before.
For example: 
image

The main process that consumes CPU:

image

image

image

Btw, I also use metabase-clickhouse-driver (v1.3.1) plugin for my Clickhouse db.

### Expected behavior

Hope to improve the loading speed of dashboard.

### Logs

_No response_

```JSON
- Metabase hosting env: Docker
- Metabase version: v0.48.1
- Metabase internal database: PostgresSQL
- The OS that Metabase running on: CentOS 7
- My Database: PostgreSQL, MySQL, StarRocks
```

### Severity

The card and dashboard load very very slowly.

### Additional context

_No response_

escherize:
resolves #37144 

This PR is a step that I think will solve the perf issue. My approach has been to modify how schema validation functions behave, particularly in our production environment. Here's a summary of what I've done:

- Selective Schema Validation: I've augmented the macro to selectively disable the emission of schema validation code for certain namespaces from `mu.fn/fn` and `mu/defn`. This was done to reduce cpu load, especially validations that might not be critical in production

- Different Behavior in Production: The changes I made ensure that these schema validations are skipped when the system is in a production environment, as indicated by config/is-prod?. The idea is to enhance performance where it matters most to our users.

- Maintaining a List of Namespaces: I've put together a list of namespaces for which these changes should apply, but it is incomplete. This targeted approach allows us to mostly maintain the rigor of validations in prod while avoiding bottlenecks.

- Balancing Risk and Performance: Any changes like this carry some risk, so I've tried to be cautious, limiting the scope of these changes to reduce the possibility of unintended effects on other parts of our system.

Our primary goal has been to address the concerns raised about high CPU usage and slow loading times in the latest Metabase version. By not running any schema validation in production, I'm hopeful that we'll see improvements in these areas.

---

Another thing this does is skip printing a message we were suspicious of causing some heavy cpu usage (in prod only).


src/metabase/util/malli/fn.clj
```diff
@@ -295,8 +295,40 @@
...
+;; ------------------------------ Skipping Namespace Enforcement in prod ------------------------------
+
+;; FIXME: is there a cleaner way to setup `namespaces-toskip`?
+;; I was thinking we could `swap!` it from the namespaces themselves with:
+;; `(swap namespaces-toskip conj *ns*)`
+(require ['metabase.lib.metadata.calculation])
+
+(def namespaces-toskip
+  "Used to track namespaces to not enforce malli schemas on with `mu.fn/fn`."
+  (atom #{(find-ns 'metabase.lib.metadata.calculation)}))
```
escherize, line 322: Open to suggestions about a clean way to do this, as well as which namespaces we should toskip.

piranha: Maybe a metadata on namespace directly? Like `^:malli/skip`, for example?

test/metabase/util/malli/fn_skip_test.clj
```diff
@@ -0,0 +1,54 @@
...
+        (is (true? config/is-prod?) "Testing that `config/is-prod?` is on.")
+        (is (= 3 (f)) "f should be 3 since we didn't check the `:string` fn-output schema.")
+        (finally
+          (reset! mu.fn/namespaces-toskip original-nts))))))
+
+(comment
+
+  (clojure.test/run-tests)
+
+  )
```
calherries, line 54:
```suggestion
```

src/metabase/util/malli/fn.clj
bshepherdson: Yeah, I think decorating the namespaces makes sense. On the other hand we probably want this for all of `metabase.lib.*` and that's a lot of files to remember to do that in. Though I guess we could use a linter to require either `^:malli/skip` or `^:malli/enforce` within `metabase.lib.*`.

src/metabase/util/malli/fn.clj
```diff
@@ -344,7 +378,10 @@
   problem for another day. The passed function name comes back from [[mc/parse]] as `:name` if we want to attempt to
   fix this later."
   [& fn-tail]
-  (let [error-context (if (symbol? (first fn-tail))
-                        {:fn-name (list 'quote (first fn-tail))}
-                        {})]
-    (instrumented-fn-form error-context (parse-fn-tail fn-tail))))
+  (let [parsed (parse-fn-tail fn-tail)]
+    `(if (#'*skip-ns-decision-fn*)
```
bshepherdson, line 382: If I'm reading this right, this is compiling the `if` into the code where the function is being constructed. That doesn't seem ideal... but there doesn't seem to be a way to check at macro expansion time.

I suppose it's not so bad if it's ultimately a top-level thing like a `mu/defn`.

src/metabase/util/malli/fn.clj
dpsutton: I don't think I would operate on namespaces as objects, just on the symbol or string of the name. Have the set of namespaces be symbols or regexes rather than `clojure.lang.Namespace` objects.

src/metabase/util/malli/fn.clj
escherize: Yeah, I didn't start with this approach, but there doesn't seem to be a way to check at macro-expansion time. 

Pointing out the obvious, but the `if` is evaluated at `mu.fn/fn` compile-time, so any extra time to run it will occur during fn creation, and not fn execution.

src/metabase/util/malli/fn.clj
```diff
@@ -295,8 +295,38 @@
...
+  This is used to emit fn which ignore checks."
+  [parsed]
+  (deparameterized-fn-form parsed))
+
+;; ------------------------------ Skipping Namespace Enforcement in prod ------------------------------
+
+;; FIXME: is there a cleaner way to setup `namespaces-toskip`?
+;; I was thinking we could `swap!` it from the namespaces themselves with:
+;; `(swap namespaces-toskip conj *ns*)`
+(def namespaces-toskip
```
dpsutton, line 318: something like this might be better:

```clojure
(defn instrument-ns?
  "Used to track namespaces to not enforce malli schemas on with `mu.fn/fn`."
  [namespace]
  (let [lib-and-middleware [#"^metabase\.lib\..*"
                            #"^metabase\.query-processor\.middleware\..*"]
        matches?    (fn [namespace regexes]
                      (let [n (-> namespace ns-name str)]
                        (some #(re-matches % n) regexes)))
        ad-hoc      '#{metabase.pulse.render} ;; just to demonstrate
        m           (meta namespace)]
    (or (:instrument/always m)
        (true? (:instrument/never m))
        (not (matches? namespace lib-and-middleware))
        (not (contains? ad-hoc (ns-name namespace))))))
```

src/metabase/util/malli/fn.clj
dpsutton: I think we have everything we need at macro-expansion time right? The only input is the namespace which is known at compile time. I think that condition can be outside of emitted code.

escherize: I couldn't figure out how to get it working with re-binding or alter-var-rooting or swap!ing the state required to test it without emitting the if. Maybe I missed something there?

src/metabase/util/malli/fn.clj
```diff
@@ -312,16 +312,29 @@
...
+        m                  (meta namespace)]
+    (cond (:instrument/always m)                  true
+          (:instrument/never m)                   false
+          (matches? namespace lib-and-middleware) false
+          (contains? ad-hoc (ns-name namespace))  false
+          :else                                   true)))
+
+(def ^:dynamic *skip-ns-decision-fn*
+  (core/fn skip-ns-decision-fn [namespace]
+    (and #_config/is-prod? config/is-dev? ;;nocommit
```
dpsutton, line 333: we need to remove the nocommit and switch from `is-dev?` to `is-prod?`

src/metabase/util/malli/fn.clj
```diff
@@ -312,16 +312,29 @@
...
+         (let [should? (instrument-ns? namespace)]
+           (when (not should?)
+             (println "skipping instrumenting var in " (ns-name namespace)))
```
dpsutton, line 336: println needs to go

src/metabase/util/malli/fn.clj
escherize: I got the emission (and altering behavior in tests) working properly, with help from @dpsutton and @noahmoss 😎.

src/metabase/util/malli/fn.clj
escherize: done

src/metabase/util/malli/fn.clj
escherize: done

src/metabase/util/malli/defn.clj
```diff
@@ -71,10 +71,17 @@
...
+    (if skip?
+      `(def ~(vary-meta fn-name merge attr-map)
+         ~docstring
+         ~(macros/case
+            :clj  (let [error-context {:fn-name (list 'quote (symbol (name (ns-name *ns*)) (name fn-name)))}]
+                    (mu.fn/instrumented-fn-form error-context parsed))
+            :cljs (mu.fn/deparameterized-fn-form parsed)))
+      `(def ~(vary-meta fn-name merge attr-map)
+         ~docstring
+         ~(macros/case
```
escherize, line 85: this can just be `(mu.fn/deparameterized-fn-form parsed)`.

escherize: done

dpsutton (approved):

noahmoss (approved):
Macro changes make sense to me now that I'm more up to speed on the implementation. My one question is should the docstring for `mu/defn` be updated to indicate that there's a way to exclude certain namespaces/ns patterns from validation? Otherwise it might be a little hard to discover unless you dig into the macro code

PR #37431, opened 2024-01-09
[wip] defendpoint2

sansolovyov:
This is a rough implementation of my idea for [stricter typing for HTTP APIs](https://www.notion.so/Strict-typing-for-HTTP-API-Tech-plan-b46fa72d73f3474888e210bd8c173f10).

What it lacks: nice error reporting, async implementation, documentation generation (ideally docs to be generated parameter definition, not only types, but descriptions as well), clj-kondo linter.

What it has: schema definition in a shape of request, defaults/optional keys (obviously, from Malli), stripping extra keys during development.


dev/src/dev.clj
```diff
@@ -307,3 +317,15 @@
      (mt/with-driver (:engine db#)
        (mt/with-db db#
          ~@body))))
+
+(defn set-log-level!
+  "Set logging level for a namespace you're in right now."
+  ([level]
+   (set-log-level! (ns-name *ns*) level))
+  ([ns level]
+   (mt/set-ns-log-level! ns level)))
```
camsaul, line 326: `mt/set-ns-log-level!` already has a 1-arity that does it for `*ns*`, this is basically doing the same thing that function does, if you really want it here just import it with Potemkin. We don't need 2 functions that do the exact same thing

dev/src/dev.clj
```diff
@@ -307,3 +317,15 @@
...
+
+(defmacro p
+  "#p, but to use in pipelines like `(-> 1 inc dev/p inc)`"
+  [form]
+  (hashp/p* form))
```
camsaul, line 331: Can you write a little more about this? I have no idea what `#p` is so I don't really know what this does at all

camsaul: Does it print something?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,168 @@
...
+    (.mkdirs f)
+    (.deleteOnExit f)
+    (.getPath f)))
+
+;;; Request callbacks
+
+(defn- ba-copy [f]
+  (let [baos (ByteArrayOutputStream.)]
+    (io/copy f baos)
+    (.toByteArray baos)))
```
camsaul, line 41:
```suggestion
  (with-open [baos (ByteArrayOutputStream.)]
    (io/copy f baos)
    (.toByteArray baos)))
```

camsaul (commented):
In terms of evaluating the `defendpoint2` proposal, it would be a lot easier to see it if you had a smaller PR that just changed a few existing endpoints; it's a little hard to review this PR because it really seems like at least three separate PRs in one. I recommend splitting this out into a few separate PRs to make it easier to review

1. one PR adding a those new endpoints relating to SerDes stuff
2. Another PR adding the `defendpoint2` stuff, maybe converting just a few endpoints over so we can see the before and after
3. Another PR with all of your test tooling changes like all of the stuff in `metabase.logger`. It would be good to walk us thru the changes there too, I'm sure you're fixing something but I don't what it is so it's hard to review changes when we don't know what they're for

src/metabase/logger.clj
```diff
@@ -47,14 +52,89 @@
...
+        (.updateLoggers (context))))))
+
+;;; Custom loggers
+
+(defn logger-name
+  "Get string name from symbol or ns"
+  ^String [a-namespace]
+  (if (instance? clojure.lang.Namespace a-namespace)
+    (name (ns-name a-namespace))
+    (name a-namespace)))
```
camsaul, line 83: Maybe this actually like 3 PRs because this seems unrelated to either the `defendpoint2` stuff or the new SerDes API endpoints. I'm a big +1 on test tooling improvements but I would split significant changes out into their own PR so they can be reviewed separately. It's always good to explain the reasoning for the changes too

camsaul (changes requested):
So, I am 100% in favor of adding typing info to our request responses in principle, but I'd like to see a concrete before and after example before signing off on this change. Can you split out the unrelated stuff in this PR into different PRs, and maybe update a few existing endpoints so we can see what the before and after looks like? Once we have a small, focused PR with that stuff then please re-request a review, but I also think we should bring this up for a general discussion with the rest of the backend team since we're all going to have to live with it. 

Specifically I'd love to get @escherize's input on this too, he worked on the Malli version of `defendpoint` -- the current parameters typing stuff that we have.

camsaul:
If we're going to overhaul the syntax for `defendpoint`, maybe we can finally fix some of the annoying things about it, for example the weird `:as` syntax that no one can remember how to use correctly -- see `POST /api/collection` for example.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -0,0 +1,168 @@
...
+     :callback #(when (.exists dst)
+                  (run! io/delete-file (reverse (file-seq dst))))}))
+
+;;; HTTP API
+
+(defendpoint2 POST "/export"
+  "Serialize and retrieve Metabase instance.
+
+  Parameters:
+  - `dirname`: str, name of directory and archive file (default: `<instance-name>-<YYYY-MM-dd_HH-mm>`)
```
escherize, line 111: It seems like docstrings can be generated, which would keep it up to date and complete.

 Have you considered generating this by adding attributes to the malli schemas in the map below? Something like:

```
dirname [:maybe {:mb/doc "name of directory and archive file" 
                 :default "<instance-name>-<YYYY-MM-dd_HH-mm>"} 
         :string]
```

deps.edn
```diff
@@ -226,6 +226,7 @@
     cloverage/cloverage          {:mvn/version "1.2.4"}
     com.gfredericks/test.chuck   {:mvn/version "0.2.14"}                    ; generating strings from regexes (useful with malli)
     djblue/portal                {:mvn/version "0.51.0"}                    ; ui for inspecting values
+    hashp/hashp                  {:mvn/version "0.2.2"}                     ; debugging/spying utility
```
escherize, line 229: should this be a dev dependency?

src/metabase/util/compress.clj
```diff
@@ -0,0 +1,54 @@
+(ns metabase.util.compress
```
escherize, line 1: Why's this needed for defendpoint2? is it gzipping somehow? I would've thought that would be handled by a lower layer of the network stack

src/metabase/api/defendpoint2.clj
```diff
@@ -0,0 +1,185 @@
...
+                   :let  [v (dissoc v :as)]]
+               (case k
+                 :responses [k (update-vals v (fn [v] `(make-schema ~(update-keys v keyword))))]
+                 :as        nil
+                 [k `(make-schema ~(update-keys v keyword))])))))
+
+(defn- make-bindings [req-sym query-spec]
+  (letfn [(walker [m prefix]
+            (mapcat (fn [[k v]]
+                      (concat
```
escherize, line 55: feels like a `cond`

escherize:
I didn't read Cam's review until now, but I had the same feeling that it's a lot more than just the defendpoint2 bit of work. I'd like to see another 2 or 3 endpoints translated over, I can see how it works with query-params (nice), but a before and after for one endpoint with 2+ path params, and one with a non-trivial body data would go a long way.

I'm also curious about how this will work with our current doc generation (search `dox` in the codebase, there's a command that works with defendpoint), as well as creating useful pram-explaining docstrings.

> the weird `:as` syntax that no one can remember how to use correctly

seconded: please fix that `:as` syntax!

piranha:
@camsaul @escherize updates:

- I cleaned up the PR — it contained another branch previous time
- added doc generation, but can't get rid of eval if we want this doc to be generated at compile time. Do we actually want this, having a full docstring on a function? Or maybe let's just generate OpenAPI spec when it's requested and get rid of eval this way?
- improved error generation, now it has a spell-check (but it fails on `data-model` vs `data_model`, that's something to fix in Malli)

What's not done yet: checks for responses (I think it should only be done for dev), OpenAPI generation. 

What's unchanged: schema definition is still this map-like syntax. I think it's nice because it's used not only as a schema, but also as a parameter definition. If you strongly feel like writing `[:map ....]` is much more preferable, I'm open to trying that out, maybe it's just me being not used to `[:map ...]` everywhere. But this syntax right now doesn't seem to be too hard (I mean it's two really simple functions to support it), and it's much closer to how parameter destructuring usually looks.

src/metabase/api/defendpoint2.clj
```diff
@@ -0,0 +1,235 @@
...
+(defn- -entry [[k v]]
+  (let [optional (and (vector? v)
+                      (= :maybe (first v)))]
+    (cond-> [k] optional (conj {:optional true}) :always (conj (make-schema v)))))
+
+(defn make-schema
+  "Compile map-based syntax to Malli schema."
+  [x]
+  (mc/schema (if (map? x)
+               (into [:map {:closed config/is-dev?}] (map -entry x))
```
qnkhuat, line 32: I love this, but I'd be way better if we can make this log/warn missing keys during development, that way we have signal to update the schema.

piranha: Well.. it's using `log/warn` right now. :) I think just failing requests in development is a much stronger signal. 😁

camsaul:
Sorry just saw this got closed out. It would still be nice to get some version of this in at some point

piranha:
@camsaul well some part of it (OpenAPI) docs got in, and I want to have a "response" schema (haven't got to doing that), but overall changing `defendpoint` that much...

If you're thinking about dropping compojure's syntax in favor of something else I'm happy to talk! I haven't felt much excitement about syntax here, and I did not want to push it too much. :)

Issue #32978, opened 2023-08-07
[Epic] Know whether a table is writeable for actions


cdeweyx:
[Product doc](https://www.notion.so/metabase/Find-a-way-to-know-whether-a-DB-has-a-writeable-connection-or-not-090dd8a6bf244fc8bbaa042579a40228)

## Implementation Plan

```[tasklist]
### Milestone 1: Implement for Postgres
- [ ] https://github.com/metabase/metabase/pull/33415
```

```[tasklist]
### Milestone 2: Implement for MySQL
- [ ] https://github.com/metabase/metabase/pull/33598
```

```[tasklist]
### Milestone 3: Create driver feature
- [ ] https://github.com/metabase/metabase/pull/34128#top
```

## Not in scope

- We only store privileges for the DB connection's user. If connection impersonation is enabled for the database, we won't be able to return privileges for the given user. Metabase users using connection impersonation will get runtime errors if they try to run actions on the database.
- Column-level privileges are not considered. Only privileges down to table-level are.

## Testing plan

- Per driver, test a function that returns the set of privileges that a given user_role has from the database.
  - should work for current user
  - should work for a specific role
- Per driver, test that the data is returned by the set of privileges for a given role
  - should work for all privilege types: select/update/insert/delete
  - should work for current user
  - should work for a specific role

calherries:
@cdeweyx can you check the epic milestones look okay to you?

cdeweyx:
> @cdeweyx can you check the epic milestones look okay to you? Note I changed the title of the epic, because as we discussed on the call the solution will allow us to know whether a specific table is writable, not the DB.

Looks good to me, thanks for the heads up 👍

PR #37439, opened 2024-01-09
Remove N+1 in sync

### LINKED ISSUE #37250: Sync takes a massive amount of time to start
### Describe the bug

I created a redshift cluster with 89 schemas and 10 tables on each. In a local environment the sync hasn't started 1 hour after I added the DB and also from adding the DB to the:
```
2024-01-04 04:45:00,008 INFO task.sync-databases :: Starting sync task for Database 3.
```
... 30 minutes passed

### To Reproduce

1) set up Metabase
2) add a redshift I created in my personal account, details here
https://metaboat.slack.com/archives/CKZEMT1MJ/p1704342090053319
3) wait... probably forever?

### Expected behavior

The sync should start as soon as you add the DB

### Logs
check the times
```
2024-01-04 04:16:03,201 INFO driver.impl :: Initializing driver :redshift...
2024-01-04 04:16:03,206 INFO plugins.classloader :: Added URL file:/app/plugins/redshift.metabase-driver.jar to classpath
2024-01-04 04:16:03,218 INFO driver.impl :: Registered abstract driver :metabase.driver.sql-jdbc.execute.legacy-impl/use-legacy-classes-for-read-and-set  
2024-01-04 04:16:03,233 INFO driver.impl :: Registered driver :redshift (parents: [:postgres :metabase.driver.sql-jdbc.execute.legacy-impl/use-legacy-classes-for-read-and-set]) 
2024-01-04 04:16:03,249 INFO metabase.util :: Load lazy loading driver :redshift took 38.2 ms
2024-01-04 04:16:03,252 INFO driver.impl :: Initializing driver :metabase.driver.sql-jdbc.execute.legacy-impl/use-legacy-classes-for-read-and-set...
2024-01-04 04:16:05,761 INFO task.sync-databases :: Scheduling sync/analyze for database 3: trigger: metabase.task.sync-and-analyze.trigger.3
2024-01-04 04:16:05,772 INFO task.sync-databases :: Scheduling field-values for database 3: trigger: metabase.task.update-field-values.trigger.3
2024-01-04 04:16:05,778 DEBUG middleware.log :: POST /api/database 200 2.6 s (17 DB calls) App DB connections: 1/10 Jetty threads: 4/50 (8 idle, 0 queued) (80 total active threads) Queries in flight: 0 (0 queued)
2024-01-04 04:16:05,782 INFO sync.util :: STARTING: Sync redshift Database 3 'rs'
```

```
2024-01-04 04:45:00,008 INFO task.sync-databases :: Starting sync task for Database 3.
```

```JSON
v47.x
```

### Severity

P1

### Additional context

UPDATE: 8 hours have passed and the sync hasn't started

UPDATE 2: started 14:46, so it took 10:30 to start
```
2024-01-04 14:46:01,537 INFO sync.util :: STARTING: Sync metadata for redshift Database 3 'rs'
2024-01-04 14:46:01,547 INFO sync.util :: STARTING: step 'sync-dbms-version' for redshift Database 3 'rs'
2024-01-04 14:46:01,740 INFO sync.util :: FINISHED: step 'sync-dbms-version' for redshift Database 3 'rs' (193.8 ms)
2024-01-04 14:46:01,744 INFO sync.util :: STARTING: step 'sync-timezone' for redshift Database 3 'rs'
2024-01-04 14:46:02,287 INFO sync.util :: FINISHED: step 'sync-timezone' for redshift Database 3 'rs' (543.0 ms)
2024-01-04 14:46:02,290 INFO sync.util :: STARTING: step 'sync-tables' for redshift Database 3 'rs'
```

UPDATE 3: found 86 of the 93 schemas that are in the Database, so all good in terms of discoverability

qnkhuat:
Closes https://github.com/metabase/metabase/issues/37250
During sync metadata process, we need to check whether or not we have select permissions against a table.
Currently we do that by making a `select true from table limit 1` on the table. This is inefficient if we have a lot of tables.

In this PR, I've changed so that we rely on the information from `driver/current-user-table-privileges` to do the check.
This method was added recently in https://github.com/metabase/metabase/issues/32978. 

Currently, it only supports Postgres,redshift, and MySQL, so only those drivers will benefit from this. We can later add support for more drivers by implementing that method.

Technical note: I've looked into using the [getTablePrivileges](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getTablePrivileges-java.lang.String-java.lang.String-java.lang.String-) from JDBC, but I decided to not using them because it does not return usage grant. So if someone has select privilege but not usage privilege, we still think they can select, but in reality, they can't.


Tony-metabase:
This can be merged but me and @qnkhuat identified a new issue which might be the problem of the sync tables. It seems that the `sync database schema now` has 2 steps and if step 1 is still running the task skips to step 2, fyi:

image

I opened an issue for reference https://github.com/metabase/metabase/issues/37695

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -353,6 +353,27 @@
                  (testing "normally, ::fake-schema should be filtered out (because it does not exist)"
                    (is (not (contains? (schemas) fake-schema-name)))))))))))))
 
+(deftest sync-materialized-views-test
```
qnkhuat, line 356: added this as I noticed we didnt' have one

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -107,44 +106,77 @@
                           :schema      (.getString rset "TABLE_SCHEM")
                           :description (when-let [remarks (.getString rset "REMARKS")]
                                          (when-not (str/blank? remarks)
-                                           remarks))}))))))
+                                           remarks))
+                          :type        (.getString rset "TABLE_TYPE")}))))))
```
qnkhuat, line 110: `current-user-table-privileges` does not return privileges for external tables on redshift, and foreign table on postgres, so we need to use the select method on them, so to check select privilege for these tables we need to use the old method of trying a `select` query. In order to do that we need to return the type so we know what's the type of table we're dealing with.

see line 134.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -164,15 +196,11 @@
...
-                  [inclusion-patterns exclusion-patterns] (driver.s/db-details->schema-filter-patterns
-                                                           prop-nm
-                                                           database)]
-              (into #{} (sql-jdbc.sync.interface/active-tables driver conn inclusion-patterns exclusion-patterns)))
-            (default-active-tbl-fn))
-          (default-active-tbl-fn)))))})
+      (let [schema-filter-prop (driver.u/find-schema-filters-prop driver)
+            database           (db-or-id-or-spec->database db-or-id-or-spec)]
+        (if (some? schema-filter-prop)
+          (let [prop-nm              (:name schema-filter-prop)
```
qnkhuat, line 202: I've removed the branch were database is nil. doesn't make any sense to call `describe-database` on a database does not exist.

test/metabase/driver/postgres_test.clj
qnkhuat, line 1: we already have tests for syncing view and foreign table in this namespace.

src/metabase/driver/postgres.clj
```diff
@@ -847,27 +847,35 @@
...
+            "   pg_catalog.has_table_privilege(current_user, '\"' || t.schemaname || '\"' || '.' || '\"' || t.objectname || '\"',  'UPDATE') as update,"
+            "   pg_catalog.has_table_privilege(current_user, '\"' || t.schemaname || '\"' || '.' || '\"' || t.objectname || '\"',  'SELECT') as select,"
+            "   pg_catalog.has_table_privilege(current_user, '\"' || t.schemaname || '\"' || '.' || '\"' || t.objectname || '\"',  'INSERT') as insert,"
+            "   pg_catalog.has_table_privilege(current_user, '\"' || t.schemaname || '\"' || '.' || '\"' || t.objectname || '\"',  'DELETE') as delete"
+            " from ("
+            "   select schemaname, tablename as objectname from pg_catalog.pg_tables"
+            "   union"
+            "   select schemaname, viewname as objectname from pg_views"
+            "   union"
+            "   select schemaname, matviewname as objectname from pg_matviews"
```
qnkhuat, line 870: updated to have it returns privileges for views and materialized views

docs/developers-guide/driver-changelog.md
```diff
@@ -17,6 +17,10 @@ title: Driver interface changelog
   `metabase.driver.sql.query-processor/honey-sql-version` is now deprecated and no longer called. All drivers are
   assumed to use Honey SQL 2.
 
+- The method `metabase.driver.sql-jdbc.sync.interface/active-tables` that we added in 47 has been updated to take
```
calherries, line 20: `require` would be better to stress that this is not an optional arg
```suggestion
- The method `metabase.driver.sql-jdbc.sync.interface/active-tables` that we added in 47 has been updated to require
```

test/metabase/driver/postgres_test.clj
```diff
@@ -1028,7 +1028,7 @@
         (t2.with-temp/with-temp [Database database {:engine :postgres, :details test-user-details}]
           ;; make sure that sync still succeeds even tho some tables are not SELECTable.
           (binding [sync-util/*log-exceptions-and-continue?* false]
-            (is (some? (sync/sync-database! database))))
+            (is (some? (sync/sync-database! database {:scan :schema}))))
```
calherries, line 1031: Why?

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -84,14 +83,14 @@
...
     (try
-      (execute-select-probe-query driver conn sql-args)
-      (log/trace "SELECT privileges confirmed")
-      true
-      (catch Throwable e
-        (log/trace e "Assuming no SELECT privileges: caught exception")
-        (when-not (.getAutoCommit conn)
-          (.rollback conn))
-        false))))
+     (execute-select-probe-query driver conn sql-args)
```
calherries, line 86: Indentation was fine before

calherries (approved):
Looks good, all my comments are nits really. This is mergeable even if you want to dismiss some of them

docs/developers-guide/driver-changelog.md
```diff
@@ -17,6 +17,10 @@ title: Driver interface changelog
...
+  an additional argument: `database`.
+  The new function arglist is `[driver database conneciton schema-inclusion-filters schema-exclusion-filters]`.
```
metamben, line 22:
```suggestion
  The new function arglist is `[driver database connection schema-inclusion-filters schema-exclusion-filters]`.
```

src/metabase/sync/sync_metadata/sync_table_privileges.clj
```diff
@@ -14,7 +14,9 @@
    This is a cache of the data returned from `driver/table-privileges`, but it's stored in the database for performance."
   [database :- (mi/InstanceOf :model/Database)]
   (let [driver (driver.u/database->driver database)]
-    (when (driver/database-supports? driver :table-privileges database)
+    (when (and (not= :redshift driver)
+               ;; redshift does support table-privileges, but we don't want to sync it now
```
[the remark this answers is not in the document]
qnkhuat: sync last discussion you said that the primary reason we sync that is to enhance actions. and since redshift doesn't support actions, I don't think we have a good reason to have sync for it.

test/metabase/driver/postgres_test.clj
qnkhuat: it's just faster for test that doesn't benefit from doing the analyze step.

src/metabase/sync/sync_metadata/sync_table_privileges.clj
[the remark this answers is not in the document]
calherries: > since redshift doesn't support actions, I don't think we have a good reason to have sync for it.

Yes exactly, I'm suggesting to leave that reason in the comment. In general when we make design decisions like this the rationale should be either easily recoverable from the code or left as a comment

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -126,17 +126,19 @@
...
-      (fn [{schema :schema table :name ttype :type}]
-        ;; driver/current-user-table-privileges does not return privielges for external table on redshift, and foreign
-        ;; table on postgres, so we need to use the select method on them
-        (if (#{[:redshift "EXTERNAL TABLE"] [:postgres "FOREIGN TABLE"]}
-             [driver ttype])
-          (sql-jdbc.sync.interface/have-select-privilege? driver conn schema table)
-          (contains? schema+table-with-select-privileges [schema table]))))
-    (fn [{schema :schema table :name}]
-      (sql-jdbc.sync.interface/have-select-privilege? driver conn schema table))))
+   (let [schema+table-with-select-privileges (schema+table-with-select-privileges driver database)]
```
calherries, line 132: Something happened to the indentation here

src/metabase/driver/sql_jdbc/sync/describe_database.clj
qnkhuat: I think it was designed so that it can takes a spec as argument. I don't see anywhere that's used in production, only in tests.

but to be safe I will revert this change and follow up on this later. don't want another PR to have unnesscary change then turns out it's nessceary like #37620 :)

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -107,44 +106,77 @@
...
+(defn- have-select-privilege-fn
+  "Returns a function that take a map with 3 keys [:schema, :name, :type], return true if we can do a select query on the table.
+
+  This function shouldn't be called a `map` or anything alike, instead use it as a cache function like so:
+
+    (let [have-select-privilege-fn* (have-select-privilege-fn driver database conn)
+          tables                   ...]
+      (filter have-select-privilege-fn* tables))"
+  [driver database conn]
+  (if (driver/database-supports? driver :table-privileges database)
```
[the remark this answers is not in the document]
qnkhuat: added one above.

src/metabase/driver/mysql.clj
```diff
@@ -876,11 +876,12 @@
   ;; for the mysql database), so we can't query the full privileges of the current user.
   (when-not (mariadb? database)
     (let [conn-spec   (sql-jdbc.conn/db->pooled-connection-spec database)
+          db-name     (get-in database [:details :dbname])
           table-names (->> (jdbc/query conn-spec "SHOW TABLES" {:as-arrays? true})
                            (drop 1)
                            (map first))]
       (for [[table-name privileges] (table-names->privileges (privilege-grants-for-user conn-spec "CURRENT_USER()")
-                                                             (:name database)
+                                                             db-name
```
qnkhuat, line 884: @calherries, This was a bug that we didn't catch in our BE tests.
we should use the db from details, not the display name.  I only catch it because the e2e tests use a different display name on its test database.

see my change in test/metabase/driver/mysql_test.clj to make sure we covered this in tests.

src/metabase/sync/sync_metadata/sync_table_privileges.clj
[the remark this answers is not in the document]
qnkhuat: added a comment

PR #37443, opened 2024-01-09
Use colors by names (as with color utility) instead of indexes in mantine

ranquild:
Allows you to use `<Text color="text-dark" />` instead of `<Text color="text.2" />`; the same with `<Text color="error" />` instead of `<Text color="error.0" />` etc.


iethree (approved):
💖

PR #37450, opened 2024-01-09
Remove getQuery selector

ranquild:


frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -251,12 +252,13 @@ class View extends Component {
...
   renderMain = ({ leftSidebar, rightSidebar }) => {
-    const { query, mode, parameters, isLiveResizable, setParameterValue } =
+    const { question, mode, parameters, isLiveResizable, setParameterValue } =
       this.props;
 
+    const legacyQuery = question.legacyQuery();
     const queryMode = mode && mode.queryMode();
-    const isNative = query instanceof NativeQuery;
-    const validationError = _.first(query.validate?.());
+    const isNative = legacyQuery instanceof NativeQuery;
```
uladzimirdev, line 260: is it a general way of checking which type we should use for legacyQuery? I saw in different PR we distinguished native query by checking `question.isStructured()`

uladzimirdev (approved):

frontend/src/metabase/query_builder/components/view/View.jsx
ranquild: we should do it via `isNative` and `isStructured` functions, but I just left it as it was

PR #35235, opened 2023-10-31
Remove GET /api/dashboard endpoint

calherries:
This PR removes the `GET /api/dashboard` endpoint (aka `DashboardApi.list` on the frontend), because it is unused by our frontend. That conclusion is based on searching for `Dashboards.loadList`, `Dashboards.ListLoader`, or `Dashboards.actions.fetchList`, and getting no matches.

### Release notes mention

To list all the dashboards using the API, third parties can still use `GET /api/collection` to get the list of collections, and for each collection use `GET /api/collection/:id/items?models=dashboard`, and finally `GET /api/dashboard/:id` to get a specific dashboard. There an example of this in [e2e/snapshot-creators/default.cy.snap.js](https://github.com/metabase/metabase/pull/35235/files#diff-3bba924aa231cd18371d6e6aa058ac423f2c238b134aaeb21ba7f35052c1db97).

Related: https://github.com/metabase/metabase/pull/35237

PR #37458, opened 2024-01-09
Restore GET /api/dashboard endpoint

### LINKED ISSUE #37455: Restore GET /api/dashboard endpoint
The GET /api/dashboard endpoint was removed in 48 ([PR](https://github.com/metabase/metabase/pull/35235)) and it the change didn't make it to the release notes. 

We should restore it for the next release of 48, with a deprecation notice to remove in the next release.

[Slack context](https://metaboat.slack.com/archives/C010L1Z4F9S/p1704806235151159).

calherries:
Closes https://github.com/metabase/metabase/issues/37455

Restores the GET /api/dashboard endpoint that was removed in 48 (https://github.com/metabase/metabase/pull/35235) with no modifications.

No guarantees will be made to keep it up-to-date with the officially supported alternative, which is to use GET /api/collection to get the list of collections, and for each collection use GET `/api/collection/:id/items?models=dashboard`, and finally GET `/api/dashboard/:id` to get a specific dashboard.

I've added the following warning for users of GET /api/dashboard to its docstring:

"This endpoint is currently unused by the Metabase frontend and may be out of date with the rest of the application. It only exists for backwards compatibility and may be removed in the future."

### DIFF
--- a/src/metabase/api/dashboard.clj
+++ b/src/metabase/api/dashboard.clj
@@ -59,10 +59,13 @@
     (filter mi/can-read? <>)))
 
 (api/defendpoint GET "/"
-  "Get `Dashboards`. With filter option `f` (default `all`), restrict results as follows:
+  "This endpoint is currently unused by the Metabase frontend and may be out of date with the rest of the application.
+  It only exists for backwards compatibility and may be removed in the future.
+
+  Get `Dashboards`. With filter option `f` (default `all`), restrict results as follows:
   *  `all`      - Return all Dashboards.
   *  `mine`     - Return Dashboards created by the current user.
-  *  `archived` - Return Dashboards that have been archived. (By default, these are *excluded*.)"
+  *  `archived` - Return Dashboards that have been archived. (By default, these are *excluded*.)""
   [f]
   {f [:maybe [:enum "all" "mine" "archived"]]}
   (let [dashboards (dashboards-list f)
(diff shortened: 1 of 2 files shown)


src/metabase/api/dashboard.clj
```diff
@@ -49,6 +49,34 @@
...
+(defn- dashboards-list [filter-option]
+  (as-> (t2/select :model/Dashboard {:where    [:and (case (or (keyword filter-option) :all)
+                                                      (:all :archived)  true
+                                                      :mine [:= :creator_id api/*current-user-id*])
+                                                [:= :archived (= (keyword filter-option) :archived)]]
+                                     :order-by [:%lower.name]}) <>
+    (t2/hydrate <> :creator)
+    (filter mi/can-read? <>)))
+
+(api/defendpoint GET "/"
```
markbastian, line 61: Do we have any way to mark deprecation? E.g. `^:deprecated`. Just wondering. No action if the answer is no.

markbastian (approved):
LGTM

src/metabase/api/dashboard.clj
calherries: Good point, I think we've used `^:deprecated` with functions before so I can do the same thing here but add it to the endpoint method. Maybe we can use this data at some point. I've added it 953c0b8b3b738355688e010b38e75506cfc4e36e

PR #37479, opened 2024-01-09
Remove delete button from archive actions list

### LINKED ISSUE #18372: Delete button appears on archived collection
**Describe the bug**
A collection can't be deleted, but when we select an archived collection, the Delete button appears but it does nothing

**Logs**
No logs at all

**To Reproduce**
1) create a collection
2) archive it
3) go to the archiving section
4) select a collection with the icon
5) click on delete button

**Expected behavior**
The delete button should not appear

**Screenshots**
image

Very related to #2783

rafpaf:
Closes https://github.com/metabase/metabase/issues/18372

This is paused since we need to do some design thinking about this


rafpaf:
Postponed because this work will be done as part of a larger overhaul of the archive section

PR #37495, opened 2024-01-10
collections in 'analytics' namespace should not be extracted

### LINKED ISSUE #37453: [Serialization] Skip Metabase Analytics and Custom reports collections when exporting content
### Describe the bug

The Metabase analytics collection is a Pro/EE only feature and is automatically added on all instances. We shouldn't serialiaze it.

### To Reproduce

1. export any Pro/EE instance
2. see exported `Metabase Analytics` and `Custom reports` collection

### Expected behavior

Skip these collections

### Logs

_No response_

```JSON
Master
```

### Severity

P1 - exposes paid content

### Additional context

One way to find the `Metabase analytics` collection is by its type=`instance-analytics`, namespace=`analytics` or entity_id (stable across all instances = `vG58R8k-QddHWA7_47umn`)

To find `Custom reports`: namespace=`analytics` and entity_id=`okNLSZKdSxaoG58JSQY54` (note that custom reports has type=null)

sansolovyov:
Resolves #37453

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
@@ -48,7 +48,9 @@
   [user-id]
   (let [roots (t2/select Collection {:where [:and [:= :location "/"]
                                                   [:or [:= :personal_owner_id nil]
-                                                       [:= :personal_owner_id user-id]]]})]
+                                                       [:= :personal_owner_id user-id]]
+                                                  [:or [:= :namespace nil]
+                                                       [:!= :namespace "analytics"]]]})]
     ;; start with the special "nil" root collection ID
     (-> #{nil}
         (into (map :id) roots)
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
@@ -44,7 +44,9 @@
 
 (defn- collection-set-for-user
   "Returns a set of collection IDs to export for the provided user, if any.
-   If user-id is nil, do not include any personally-owned collections."
+   If user-id is nil, do not include any personally-owned collections.
+
+  Does not export ee-only analytics collections."
   [user-id]
   (let [roots (t2/select Collection {:where [:and [:= :location "/"]
                                                   [:or [:= :personal_owner_id nil]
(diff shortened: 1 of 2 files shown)


enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
```diff
@@ -1711,3 +1712,11 @@
                (by-model "Collection" ser)))
         (is (= #{ncard-eid}
                (by-model "Card" ser)))))))
+
+(deftest skip-analytics-collections-test
+  (testing "Collections in 'analytics' namespace should not be extracted, see #37453"
+    (mt/with-empty-h2-app-db
+      (mbc/ensure-audit-db-installed!)
+      (let [ser (extract/extract {:no-settings   true
+                                  :no-data-model true})]
```
qnkhuat, line 1721: I'd have this
```clojure
(testing "sanity check that the audit collection exists"
  (is (true? (t2/exists? :model/Collection :is_audit true))))
```

for document purposes

enterprise/backend/src/metabase_enterprise/serialization/v2/extract.clj
```diff
@@ -48,7 +48,9 @@
   [user-id]
   (let [roots (t2/select Collection {:where [:and [:= :location "/"]
                                                   [:or [:= :personal_owner_id nil]
-                                                       [:= :personal_owner_id user-id]]]})]
+                                                       [:= :personal_owner_id user-id]]
+                                                  [:or [:= :namespace nil]
+                                                       [:!= :namespace "analytics"]]]})]
```
qnkhuat, line 53: I don't think `[:= :namespace nil]` is necessary, right?

qnkhuat: and could you please update the doc of this function too?

piranha: It is, since `!=` will skip all NULL entries. So it's either that, or using `is distinct from`.

piranha: > and could you please update the doc of this function too?

I can, no probs, but it feels still relevant to me as it is.

qnkhuat: huh, so TIL then!

qnkhuat (approved):
LGTM

PR #37499, opened 2024-01-10
Migrate updateQuestion to MLv2

### LINKED ISSUE #37392: [MLv2] [FE] Migrate updateQuestion

ranquild:
Closes https://github.com/metabase/metabase/issues/37392

Simply the QB logic by removing `hasNewColumns` check. Now we decide that we need to-rerun the query in place where we change the query itself and not try to figure it out later.


frontend/src/metabase/query_builder/actions/core/updateQuestion.ts
```diff
@@ -29,15 +29,23 @@ import { onCloseQuestionInfo, setQueryBuilderMode } from "../ui";
 
 import { getQuestionWithDefaultVisualizationSettings } from "./utils";
 
-function hasNewColumns(question: Question, queryResult: Dataset) {
-  // NOTE: this assume column names will change
+function hasNewColumns(question: Question, queryResult: Dataset | null) {
+  if (!queryResult) {
+    return false;
+  }
+  if (question.isNative()) {
```
uladzimirdev, line 36: this change is not clear to me

uladzimirdev (approved):

frontend/src/metabase/query_builder/actions/core/updateQuestion.ts
kamilmielnik: It's due to `query instanceof StructuredQuery` part.

frontend/src/metabase/query_builder/actions/core/updateQuestion.ts
```diff
@@ -29,15 +29,22 @@ import { onCloseQuestionInfo, setQueryBuilderMode } from "../ui";
...
-  const query = question.legacyQuery({ useStructuredQuery: true });
-  const previousColumns =
-    (queryResult && queryResult.data.cols.map(col => col.name)) || [];
-  const nextColumns =
-    query instanceof StructuredQuery ? query.columnNames() : [];
-  return _.difference(nextColumns, previousColumns).length > 0;
+  const query = question.query();
+  const stageIndex = -1;
+  const queryColumns = Lib.returnedColumns(query, stageIndex);
+  return queryColumns.length > queryResultColumns.length;
```
kamilmielnik, line 47: I think this is not enough. Previously diff of arrays of names of columns was checked.
So it does not handle a case where 1 column is removed and another is added.

uladzimirdev: nice catch btw

ranquild: I'm experimenting to see if I can avoid the issue with the wrong column names in 
`returnedColumns`. But it seems that not. Reverted this change. The PR would have to stay here until the referenced MLv2 bug is fixed

e2e/test/scenarios/visualizations-charts/reproductions/21452-xhr-on-every-char-for-rename.cy.spec.js
```diff
@@ -36,8 +36,6 @@ describe("issue 21452", () => {
     cy.findByDisplayValue("Cumulative sum of Quantity").clear().type("Foo");
     // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
     cy.findByText("Display type").click();
-    // Blur will result in another POST request which is expected
-    cy.wait("@dataset");
```
ranquild, line 40: New streamline logic works even before than before - there is no duplicate request when changing viz settings only.

ranquild:
@kamilmielnik could you take a look one more time, please? The PR has significantly changed.

kamilmielnik (approved):
Nice :+1:

PR #37500, opened 2024-01-10
Make it easier to track Structured legacy query usages

ranquild:
Currently it's hard to track `StructuredQuery` usage via `legacyQuery` because it could be `NativeQuery` and we don't migrate it to MLv2 now. The idea is to require `useStructuredQuery` flag in `legacyQuery` call:
```
legacyQuery({ useStructuredQuery: true })
```
otherwise it would throw an exception. If this approach gets approved I'll add `useStructuredQuery` to all places where `legacyQuery` is called, and we can migrate them one by one.

### DIFF
--- a/enterprise/frontend/src/metabase-enterprise/audit_app/components/QueryViewer/QueryViewer.tsx
+++ b/enterprise/frontend/src/metabase-enterprise/audit_app/components/QueryViewer/QueryViewer.tsx
@@ -28,14 +28,11 @@ export function QueryViewer({ datasetQuery }: { datasetQuery: DatasetQuery }) {
   }, [card, dispatch]);
 
   const question = new Question(card, metadata);
-
-  const query = question.legacyQuery({ useStructuredQuery: true });
-
   if (question.isNative()) {
     return (
       <NativeQueryEditor
         question={question}
-        query={query}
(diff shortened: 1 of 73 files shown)


nemanjaglumac (commented):
This approach makes a ton of sense.

I'm looking forward to seeing explicit usages of Structured query.
Thanks for doing this @ranquild

enterprise/frontend/src/metabase-enterprise/audit_app/components/QueryViewer/QueryViewer.tsx
```diff
@@ -45,7 +42,8 @@ export function QueryViewer({ datasetQuery }: { datasetQuery: DatasetQuery }) {
     );
   }
 
-  if (query.tables()) {
+  const database = question.database();
+  if (database && database.tables) {
```
ranquild, line 46: This is exactly the same check as in `query.tables()` in `StructuredQuery`. Not 100% check what this check does but I want to leave it as it is.

uladzimirdev:
Looks nice, probably some further renaming is required, but approach LGTM

ranquild:
@uladzimirdev I've fixed failed tests

uladzimirdev (approved):

PR #37503, opened 2024-01-10
fix serialization when there is a slash in click_behavior id

### LINKED ISSUE #37555: Serialization fails when line chart has click behavior linked to a field with base-type
### Describe the bug

When you do what subject says, `click_behavior` key looks like this: `["field",7650,{"base-type":"type/Integer"}]`. This slash in `type/Integer` breaks serialization.

### To Reproduce

Add a line chart to a dashboard (i am using the Sample Database -> Invoices -> Count Group by created at month).
Then link it to another question, in my case a pivot.

### Expected behavior

_No response_

### Logs

_No response_

```JSON
Metabase 1.47.10
```

### Severity

Preventing serialization

### Additional context

_No response_

sansolovyov:
An escalation from @Tony-metabase - when a click_behavior id is something like `[:dimension [:field "something" {:base-type "type/Text"}]]`, going through keyword and back is a little bit tricky. Would be nice to stop keywordizing `click_behavior` keys, but that's a much harder problem to solve.

Resolves #37555

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
@@ -1,5 +1,6 @@
 (ns ^:mb/once metabase-enterprise.serialization.v2.extract-test
   (:require
+   [cheshire.core :as json]
    [clojure.set :as set]
    [clojure.string :as str]
    [clojure.test :refer :all]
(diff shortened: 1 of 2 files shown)


src/metabase/models/serialization.clj
```diff
@@ -1223,7 +1223,9 @@
   ported by [[export-viz-click-behavior-mapping]]."
   [mappings]
   (into {} (for [[kw-key mapping] mappings
-                 :let [k (name kw-key)]]
+                 :let [k (if (namespace kw-key)
+                           (str (namespace kw-key) "/" (name kw-key))
+                           (name kw-key))]]
```
qnkhuat, line 1228:
```suggestion
                 :let [k (u/qualified-name kw-key)]]
```
we have a helper for this

qnkhuat:
do we have an issue for this escalation?

piranha:
> do we have an issue for this escalation?

nope, just Slack discussions

src/metabase/models/serialization.clj
```diff
@@ -1223,7 +1223,7 @@
   ported by [[export-viz-click-behavior-mapping]]."
   [mappings]
   (into {} (for [[kw-key mapping] mappings
-                 :let [k (name kw-key)]]
+                 :let [k (u/qualified-name kw-key)]]
```
calherries, line 1226: Reading this usage of `u/qualified-name` without the test is pretty confusing - why would a mapping keyword have a namespace? To save devs time in the future puzzling over this, I would leave a comment

enterprise/backend/test/metabase_enterprise/serialization/v2/extract_test.clj
```diff
@@ -1496,14 +1497,23 @@
...
-                                                                 {:click_behavior {:type "link"
-                                                                                   :linkType "question"
-                                                                                   :targetId c3-2-id
-                                                                                   :parameterMappings {}}}}
+                                                                 (let [dimension  [:dimension [:field "something" {:base-type "type/Text"}]]
+                                                                       mapping-id (json/generate-string dimension)]
+                                                                   {:click_behavior {:type     "link"
+                                                                                     :linkType "question"
+                                                                                     :targetId c3-2-id
+                                                                                     :parameterMapping
```
calherries, line 1509: I noticed this was wrong before, probably because the docstrings of `import-viz-click-behavior-mappings` and `export-viz-click-behavior-mappings` have it wrong too. Can you clean those up with this fix?

calherries (approved):
LGTM

src/metabase/models/serialization.clj
piranha: I don't see a need: you (should) know a mapping keyword shouldn't been a keyword in the first place, it's just how it's processed after being selected from db. In an ideal world we'd either have different data layout for `click_behavior` or not convert it's keys to a keywords.

And `u/qualified-name` just says "give me all content".

darksciencebase:
> @piranha can you create an issue for this before it lands, we need issues for all bugs. The titles of closed issues that have a milestone corresponding to the release get added to the automatically

PRs with milestones work for release notes just as fine. no need to create an issue.

calherries:
@darksciencebase Oh cool, thanks for that clarification. I recall Luiggi asking for issues once, but that might be a thing of the past.

iethree:

PR #37505, opened 2024-01-10
Redshift support for CSV uploads (2)

calherries:
This PR fixes a bug with https://github.com/metabase/metabase/pull/37005 where redshift's support for uploads is overwritten from true to false accidentally.

I must not have noticed it when I wrote the original PR because of my REPL state.

It also makes three changes to the code:
1. Updates tests to get around the fact that for redshift `mt/with-empty-db` does not actually create an empty database with redshift, and uses the a session schema. All the test tables stick around so I've isolated them by using random names.
2. A change in `src/metabase/upload.clj`, so that redshift and future drivers to support uploads do not automatically create an `_mb_row_id` column in the table when appends are made. This is necessary because redshift does not allow you to add a PK column after the table is created. The original goal of adding `_mb_row_id` columns during appends was to clean up any uploaded tables that were created before 48. But redshift and future drivers that will support uploads don't need to do this, because all uploaded tables will have an `_mb_row_id` from when they were created.
3. Breaking changes to `driver/create-table!` and `driver/upload-type->database-type` driver multimethods. Instead of `driver/upload-type->database-type` returning a vector that includes the `:primary-key` information, which then gets passed to `driver/create-table!`, `primary-key` is now an additional arg in `create-table!`. That's probably how it should have been anyway, and I think it's cleaner to separate primary key definition from column type. But the change is necessary because redshift does not allow syntax like `CREATE TABLE accounts (user_id SERIAL PRIMARY KEY);`. Instead redshift requires the `PRIMARY KEY` constraint to be after the list of columns: `CREATE TABLE accounts (user_id BIGINT IDENTITY(1,1), PRIMARY KEY(user_id));`


sansolovyov (dismissed):
Hard to say something 😁

calherries:
tests are failing, seems like this isn't going to be so simple. temp test tables in redshift aren't being dropped as expected, probably because `mt/with-empty-db` behaves differently. I'll fix tomorrow

src/metabase/upload.clj
```diff
@@ -612,7 +620,12 @@
           normed-name->field (m/index-by (comp normalize-column-name :name)
                                          (t2/select :model/Field :table_id (:id table) :active true))
           normed-header      (map normalize-column-name header)
-          create-auto-pk?    (not (contains? normed-name->field auto-pk-column-name))
+          ;; Create auto-pk columns for drivers that supported uploads before auto-pk columns
+          ;; were introduced by metabase#36249. Otherwise we can assume that the table was created
+          ;; with an auto-pk column, or it has been deliberately removed.
+          create-auto-pk?    (and
+                              (contains? #{:mysql :postgres :h2} (:engine database))
```
qnkhuat, line 627: I think redshift is a special case here, I prefer writing this as negation : `(not #{:redshift} (:engine database))`.

Also, maybe it's better if we turn this into a `should-create-auto-pk?` helper.

src/metabase/driver/sql_jdbc.clj
```diff
@@ -117,19 +117,20 @@
   [_driver _semantic_type expr]
   (h2x/->timestamp expr))
 
-(defn- create-table-sql
-  [driver table-name col->type]
+(defn- create-table!-sql
```
qnkhuat, line 120: is it better to define `defmethod driver/create-table!` for `:redshift` rather than making a breaking change to `driver/create-table!` method?

now looking at this I also think maybe we should have `create-table-sql` as a defmulti.

src/metabase/upload.clj
calherries: It is possible for 3rd party drivers to have implemented CSV uploads already, and indeed it would be a nice-to-have to have them create an auto PK here. Creating a `should-create-auto-pk?` driver multimethod would solve the problem, but I considered it not worth it. It's unlikely that a 3rd party driver author cares enough about creating an auto PK column to implement the method (we're only doing this as a nice-to-have to anticipate future actions features).

src/metabase/driver/sql_jdbc.clj
calherries: IMO the `driver/create-table!` method was designed incorrectly, including constraint information like primary keys in the `col->type` argument. One downside of that is its previous form it doesn't support multiple column primary keys. I'd prefer to fix that design here.

As for a `create-table-sql` multimethod, that's a good idea. It would separate the execution logic (which should be the same for all SQL drivers) from the SQL format, which might differ.

calherries: I'd like to add the `create-table-sql` multimethod in another PR though, and batch it with more changes. If we do this for `create-table-sql` we should also do it for the other DDL multimethods like `driver/add-columns!` and `driver/drop-table!`, so the driver interface is consistent.

test/metabase/upload_test.clj
```diff
@@ -404,14 +413,65 @@
...
       (mt/with-empty-db
-        (let [file (csv-file-with ["id" "2" "3"])]
-          (testing "Can upload two files with the same name"
-            (is (some? (@#'upload/load-from-csv! driver/*driver* (mt/id) (format "table_name_%s" driver/*driver*) file)))
-            (is (some? (@#'upload/load-from-csv! driver/*driver* (mt/id) (format "table_name_2_%s" driver/*driver*) file)))))))))
+        (mt/with-current-user (mt/user->id :crowberto)
+          (let [csv-file-name (str (mt/random-name) ".csv")]
+            (let [model-1 (upload-example-csv! :csv-file-name csv-file-name)
+                  ;; sleep for a second to make sure the table name is different
+                  _ (Thread/sleep 1000)
```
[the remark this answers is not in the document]
crisptrutski: Update: ahh, I see it's in `unique-table-name`. I suppose the ship has sailed on the naming scheme being a human readable time, or at least changing that is out of scope here, but I think there's a good pattern to fix this function (it claims to be "unique"). Besides shaving off test seconds and warding off future flakes, the clock it's using its using can jump back in practice so it really not good to depend on.

A common pattern where any precision is fine is to use `System.nanoTime()`, which is guaranteed to be monotonic. If we don't want to change the format we could create our own helper that keeps track of the last time it returned, and returns the max of the latest time and that last time + 1 second.

test/metabase/upload_test.clj
```diff
@@ -539,387 +598,372 @@
...
+                        alternating (map even? (range (count bool-column)))]
+                    (is (= alternating bool-column))))))))))))
 
 (deftest load-from-csv-length-test
   (testing "Upload a CSV file with large names and numbers"
     (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
       (let [length-limit (driver/table-name-length-limit driver/*driver*)
-            long-name    (apply str (repeat 33 "abcdefgh")) ; 33×8 = 264. Max is H2 at 256
-            short-name   (subs long-name 0 (- length-limit (count "_yyyyMMddHHmmss")))]
+            long-name    (apply str (shuffle (apply concat (repeat 33 (seq "abcdefgh"))))) ; 33×8 = 264. Max is H2 at 256
```
crisptrutski, line 640: I'm curious why we need this shuffle here, is it to work around flakes with `unique-table-name` again?

test/metabase/upload_test.clj
```diff
@@ -539,387 +598,372 @@
...
-         driver/*driver*
-         (mt/id)
-         "upload_test"
-         (csv-file-with [",ship name,"
-                         "1,Serenity,Malcolm Reynolds"
-                         "2,Millennium Falcon, Han Solo"]))
-        (testing "Table and Fields exist after sync"
-          (sync/sync-database! (mt/db))
-          (let [table (t2/select-one Table :db_id (mt/id))]
-            (is (=? {:name #"(?i)upload_test"} table))
```
crisptrutski, line 620: We did we remove this assertion instead of updating it? I guess it was being repeated a lot, but is it still tested at least once somewhere?

test/metabase/upload_test.clj
```diff
@@ -539,387 +598,372 @@
...
+            (mt/with-empty-db
+              (@#'upload/load-from-csv!
+               driver/*driver*
+               (mt/id)
+               table-name
+               (csv-file-with ["_mb row id,ship,captain,_MB_ROW_ID"
+                               "100,Serenity,Malcolm Reynolds,200"
+                               "3,Millennium Falcon, Han Solo,4"]))
+              (testing "Table and Fields exist after sync"
+                (let [table (sync-upload-test-table! :database (mt/db) :table-name table-name)]
```
crisptrutski, line 830: Aside - pattern repeats a lot, and there was a lot of noise to update it. Seems like it would be worth DRYing up to simple macro. Perhaps something like:

```clojure
(with-csv-table [table]
  ["my,csv,contents"
   ...]
  ...)
```

test/metabase/upload_test.clj
```diff
@@ -97,6 +96,16 @@
   [& body]
   `(do-with-mysql-local-infile-off (fn [] ~@body)))
 
+(defn sync-upload-test-table!
```
crisptrutski, line 99: Awesome idea 🤟 

I'm curious as to how it's faster, and whether there are any gotchas related to that. Perhaps the docstring could metnion what's being skipped?

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -376,7 +376,7 @@
     ::upload/varchar-255              [[:varchar 255]]
     ::upload/text                     [:text]
     ::upload/int                      [:bigint]
-    ::upload/auto-incrementing-int-pk [:bigint [:identity 0 1]]
+    ::upload/auto-incrementing-int-pk [:bigint [:identity 1 1]]
```
crisptrutski, line 379: Why did this need to change? It would be nice if a comment explained these numeric params.

src/metabase/upload.clj
```diff
@@ -625,7 +638,7 @@
         (driver/add-columns! driver
                              (:id database)
                              (table-identifier table)
-                             {(keyword auto-pk-column-name) (driver/upload-type->database-type driver ::auto-incrementing-int-pk)}))
+                             {auto-pk-column-keyword (conj (driver/upload-type->database-type driver ::auto-incrementing-int-pk) :primary-key)}))
```
crisptrutski, line 641: So the `add-columns` method is keeping `:primary-key` as part of the database type? I'm a bit disturbed by this inconsistency between the different driver public methods.

While I understand the need for an extra parameter to support a compound key, e.g. to give their order, which affects performance, I don't see the harm in marking the corresponding types with this keyword as well. 

Are you planning to change this interface as well in another PR? I guess that's fine as I don't see anything but `upload.clj` using it.

If we are going to have slightly different formats and semantics for column types between `create-table!` and `add-columns!`, it feels like quite a smell to do it as an implementation detail of this method, and extracting a first class "uptrade-type" method would lesson the evil.

src/metabase/upload.clj
```diff
@@ -612,7 +620,12 @@
           normed-name->field (m/index-by (comp normalize-column-name :name)
                                          (t2/select :model/Field :table_id (:id table) :active true))
           normed-header      (map normalize-column-name header)
-          create-auto-pk?    (not (contains? normed-name->field auto-pk-column-name))
+          ;; Create auto-pk columns for drivers that supported uploads before auto-pk columns
+          ;; were introduced by metabase#36249. Otherwise we can assume that the table was created
+          ;; with an auto-pk column, or it has been deliberately removed.
+          create-auto-pk?    (and
```
crisptrutski, line 626: This is a very high level part of the spec, so I don't think it's appropriate for it to be squirreled away inside a large mutating method. I think extracting a top level function like `(append-pk-on-upload? database-egine)` is a good idea.

src/metabase/driver.clj
```diff
@@ -901,8 +901,9 @@
   :hierarchy #'hierarchy)
 
 (defmulti create-table!
-  "Create a table named `table-name`. If the table already exists it will throw an error."
-  {:added "0.47.0", :arglists '([driver db-id table-name col->type])}
+  "Create a table named `table-name`. If the table already exists it will throw an error.
+  `primary-key` is optional and is a vector of column names that will be used as the primary key."
+  {:added "0.47.0", :arglists '([driver database-id table-name column->type primary-key])}
```
crisptrutski, line 906: Is it not worth future proofing things somewhat and making the last parameter a metadata map or kwargs? I could imagine indexes for example becoming important if we make imports more scalable.

src/metabase/driver/sql_jdbc.clj
```diff
@@ -117,19 +117,20 @@
...
+  [driver table-name col->type primary-key]
   (first (sql/format {:create-table (keyword table-name)
-                      :with-columns (map (fn [[name type-spec]]
-                                           (vec (cons name type-spec)))
-                                         col->type)}
+                      :with-columns (cond-> (mapv (fn [[name type-spec]]
```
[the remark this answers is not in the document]
calherries: I think you might be counting parentheses incorrectly. The `cond->` definitely envelops the `primary-key (conj ...)` forms on line 126

src/metabase/driver/sql_jdbc.clj
qnkhuat: yea bug what do you think about changing `col->type` to something like `col-definitions` tho?  like what we have in FieldDefinition for defdataset.

I'm a bit resistant about adding `primary-key` as a special case here; what if, in the future, we want to create a column to be indexed?

crisptrutski: +1 to that renaming, the meaning of that variable wasn't immediately clear to me

src/metabase/driver/sql_jdbc.clj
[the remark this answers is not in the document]
crisptrutski: Ah, quite right. That makes more sense.

src/metabase/upload.clj
```diff
@@ -612,7 +620,12 @@
...
+                              (contains? #{:mysql :postgres :h2} (:engine database))
+                              (not (contains? normed-name->field auto-pk-column-name)))
```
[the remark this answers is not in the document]
crisptrutski: If it's already an established idiom, I'm happy. 

Thanks for the reference, I see that code even made non-local assumptions that the value was a concrete map, and in this case at least the variable is defined locally and it's even less ambiguous.

I've been meaning to read Elements of Clojure, so thanks for the excuse. 

> The naming is ambiguous to me

Sorry I feel like I'm missing something here - if it's ambiguous, I would to only assume the lowest common denominator, that it is an `IFn`. I guess a reasonable heuristic to follow that if it's an argument or local binding then it's "data" and hence a map, and global references can be looked up quickly to check.

src/metabase/upload.clj
crisptrutski: We tend to write pretty thorough docstrings, so I guess putting it there is a good idea. My main concern is that it's easy to lose track of every detail in large comments, and they can fall out of sync with what the code really does.

I'm not sure I see how the problem is the same - if we define a top level function then it shows up in the code outline as a first class concept, so more discoverable and atomic. On the tin instead of inside it. The same principle as extracting constants versus commenting on magic variables, even if they're only referenced once.

If you separated the business logic and implementation details into separate namespaces it would stand out even more prominently. In the general case it's also more testable, or usable by tests to determine what the expected side effects here.

If docstrings are king then it's also easier to keep that function's docstring uptodate as it'd be right next to the map literal.

I'm not going to push hard on this now, but it's got me thinking about how we structure things. I'm sure my thinking will evolve as I get more familiar with the codebase.

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: By this

> The naming is ambiguous to me, x->y is either a map or a function

I mean x->y is either an associative map (implementing `clojure.lang.Associative`) or a function that does not. And I mean you cannot tell from the name alone.

test/metabase/upload_test.clj
```diff
@@ -404,14 +413,65 @@
...
+  "Upload a small CSV file to the given collection ID. `grant-permission?` controls whether the
+  current user is granted data permissions to the database."
+  [& {:keys [table-prefix collection-id grant-permission? uploads-enabled user-id db-id sync-synchronously? csv-file-name]
+      :or {collection-id       nil ;; root collection
+           grant-permission?   true
+           uploads-enabled     true
+           user-id             (mt/user->id :rasta)
+           db-id               (mt/id)
+           sync-synchronously? true
+           ;; Make the file-name unique so the table names don't collide
```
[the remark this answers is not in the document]
calherries: The collision happens at the schema-level. Because names are created according to the second, if there are two tests that happen to run at the same instant, the names will not be unique.

test/metabase/upload_test.clj
```diff
@@ -404,14 +413,65 @@
...
+           csv-file-name       (str "example csv file " (random-uuid) ".csv")}
```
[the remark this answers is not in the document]
calherries: Done f497eb59a7dd61ac4227a1ff2edd2ab761882615

src/metabase/driver/sql_jdbc.clj
calherries: > another idea: what do you think about changing col->type to something like col-definitions? like what we have in FieldDefinition for defdataset.

I like that idea. I believe `col->type` is left over from when we added primary keys, and we didn't change the name back then.

> I'm a bit resistant about adding primary-key as a special case here; what if, in the future, we want to create a column to be indexed?

Yes, I will change primary-key to be in a kwargs map

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: Gotcha on the ambiguity, that's what I thought you meant. 

My original point was that if something can be A or B, and A is a kind of B, then I would think you should code against the weaker assumption that it is some kind of B. In this case A is `Associative` and B is `IFn`. We can code to that weaker assumption by using `(nil? (a->b k))` instead of `(not (contains? a->b k))` - it's even shorter, although slightly less semantically clear.

I'm happy to move on though, will just bear in mind there's always burden on the reader to be careful with making the stronger "it's data" assumption.

test/metabase/upload_test.clj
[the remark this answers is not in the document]
crisptrutski: I realize my confusion - the file name *on disk* will be unique without the uuid, because this local variable here is really just the prefix. The table name just depends on the prefix, not on the filename.

I think this could be slightly clearer if you changed "file-name" to "prefix", but not a big deal.

src/metabase/driver/sql_jdbc.clj
calherries: Renamed col->type here 95c2d3a82295fe9ba2760e53ec10d4757ca9aa91
Changed primary-key to be in a kwargs map here b7d90136dd84ba2952df8b8d9cb9f06a2087c8ab

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: Discovering points of divergence is how I learn so I'm happy to keep discussing this until we understand each other.

In this case you are correct in the sense that the name does imply that it is A or B, and if we only had that information on x->y I would agree, we should code against the weaker assumption. But we don't only have the name, we have the ability to look up the actual interface of x->y, from which we know that it is a map.

I think you are actually arguing that it is better to have names that disambiguate types and to that I agree that is a benefit of the `b-by-a` convention. OTOH this naming convention has the benefits I listed above, and I think its downsides are limited if used as a local variable name.

test/metabase/upload_test.clj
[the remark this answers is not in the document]
calherries: ok, changed file-name to prefix to avoid confusion in the future 83a10b480c7de7e03209da84c5205e0ef32c13ce

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: on this

> We can code to that weaker assumption by using `(nil? (a->b k))` instead of `(not (contains? a->b k))`

I'm not sure I would prefer that because if `a->b` was a function, the name alone doesn't imply that it returns `nil` for an arg that isn't in its domain. So it doesn't really fit the weaker assumption implied by `a->b` and actually puts quite strong assumptions of the implementation of the hypothetical function `a->b`. I think I prefer `(contains? a->b k)` because at least it's making the assumption clear that `a->b` implements Associative and those semantics are well understood. And a function could return nil for many reasons.

test/metabase/upload_test.clj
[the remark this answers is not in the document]
calherries: Yep, great call. Something for another PR.

test/metabase/upload_test.clj
```diff
@@ -404,14 +413,65 @@
...
+          (let [csv-file-name (str (mt/random-name) ".csv")]
+            (let [model-1 (upload-example-csv! :csv-file-name csv-file-name)
+                  ;; sleep for a second to make sure the table name is different
+                  _ (Thread/sleep 1000)
+                  model-2 (upload-example-csv! :csv-file-name csv-file-name)]
+              (testing "tables are different between the two uploads"
+                (is (some? (:table_id model-1)))
+                (is (some? (:table_id model-2)))
+                (is (not= (:table_id model-1)
+                          (:table_id model-2)))))))))))
```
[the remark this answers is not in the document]
calherries: I get the non-nil value in the failure summary with Calva:

```
(deftest x
  (let [not-nil ":)"]
    (is (nil? not-nil))))
```

```
; FAIL in metabase.models-test/x (models_test.clj:22):
; expected:
(nil? not-nil)

; actual:
(not (nil? :)))
```

> This type of assertion has come up more than once now, so a test util like all-distinct handling any arity could be nice.

Maybe, I've seen it a couple of times recently too. Feel free to create one the next time you see it, I'm not going to bother here

src/metabase/upload.clj
calherries: Fair point, the difference does feel a little strange. I've changed `add-column!` to be consistent cb71757bcaa331829ce68c8cb1095fee8f758b8a

test/metabase/upload_test.clj
calherries: It's to work around the fact that `mt/with-empty-db` doesn't clean up after itself with redshift. Instead there is one database for the entire test run, and a schema is created for the session during initialization. See [this](https://github.com/metabase/metabase/blob/99e3e107ebb8b2c5265682f6c311fd9e82237d7c/modules/drivers/redshift/test/metabase/test/data/redshift.clj#L132-L134). The table remains in the test database and the unused tables are cleaned up in regular intervals. I could probably fix this by cleaning up the table explicitly in each test run, but opted for this workaround.

test/metabase/upload_test.clj
calherries: Yeah, I've wanted to do this for a while. I'm definitely not making that change in this PR though, it's already big enough

src/metabase/upload.clj
calherries: I decided to create this as a multimethod called `driver/create-auto-pk-with-append-csv?` in 6a6a2e00bab0caf80144d8b796fd0c75b9ada2a9. The thing that tipped me over the edge was this conversation with Chris https://github.com/metabase/metabase/pull/37505#discussion_r1473562804

calherries:
@crisptrutski @qnkhuat I think I've addressed all your comments now, please take another look.

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: It took me we a while, but I think the penny dropped at how we were missing each other.

I believe you're saying that we can apply duck typing. We know that this variable is associative *because* there is code that calls `contains?` on it.

My concern was about checking that it's safe to add that call in the first place, and the long term safety of the assumption - for example a far away definition could get refactored from a map to a function. I guess it's on the reviewer to check and the tests to give confidence, and that's fine.

A long thread but valuable for to remember that kind "duck" thinking - seeing every method call as type information for the variable they're being passed. It's been a very long time since I worked in a dynamic language and this was a useful insight.

src/metabase/upload.clj
crisptrutski: I'm genuinely fascinated about how differently we think about code structure and documentation, look forward to discussing it more as we write more code together. Happy to take guidance on our existing conventions as a start.

Moving the logic to the multimethod certainly ticks my boxes on pushing the business logic out of the implementation details, but I have two passing comments.

1. I thought you gave a really good reason not to make this extensible at first - only this subset of drivers will have created non-pk tables in the past, so it's a moot point for the others. It seems unfortunate to bloat the public interface in this case, even if in practice new drivers will all use the default implementation.

2. I'm not that comfortable with having `csv` in the method name. This aspect of the behavior has nothing to do with the source file format, and we may support addition ones in future, e.g. parquet.

crisptrutski:
> @crisptrutski @qnkhuat I think I've addressed all your comments now, please take another look.

Based on the discussions, I'm happy with how everything was resolved. I'll try make some time later today to go read through the code carefully again but don't block on me if someone else ships it first.

src/metabase/upload.clj
calherries: On 1, I'm a little torn but I think the more "correct" thing to do is follow Ngoc's suggestion to add the multimethod, since it's possible (though unlikely) 3rd party drivers could have implemented uploads in 47. It's a little unfortunate the public interface has increased but it's a small cost to pay as you said, new drivers will use the default implementation. Since I was extracting the function anyway, it seemed worth doing right.

On 2 I agree. I'll make the change.

docs/developers-guide/driver-changelog.md
```diff
@@ -6,8 +6,17 @@ title: Driver interface changelog
 
 ## Metabase 0.49.0
 
-- The multimethod `metabase.driver/add-columns!` has been added. This method is used to add a column to a table.
-  Currently it only needs to be implemented if the database supports the `:uploads` feature.
+- The multimethod `metabase.driver/create-table!` can take an additional optional map with an optional key `primary-key`
```
qnkhuat, line 9:
```suggestion
- The multimethod `metabase.driver/create-table!` can take an additional optional map with an optional key `primary-key`.
```

src/metabase/driver/sql_jdbc.clj
qnkhuat: sorry I wasn't clear before; I was thinking about structuring `column-definitions` like this.

```
[{:column-name "name", :column-type :text}
 {:column-name "id", :column-type: int, :primary-key? true}]
```

This way it'd be easier for us to add new property(e.g, indexed?, default-value ....) in the future.
It borrows the same idea that we use for `FieldDefinition` in tests: https://github.com/metabase/metabase/blob/master/test/metabase/test/data/interface.clj#L45.

wdyt about this approach?

src/metabase/driver.clj
```diff
@@ -948,11 +952,21 @@
 
   - [:bigint]
   - [[:varchar 255]]
-  - [:generated-always :as :identity :primary-key]"
+  - [:generated-always :as :identity]"
   {:changelog-test/ignore true, :added "0.47.0", :arglists '([driver upload-type])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
+(defmulti create-auto-pk-with-append-csv?
```
qnkhuat, line 960: this looks like something we should use driver feature for?

calherries: What's the benefit you see? I felt like a multimethod is better because it (a) is more hidden away away it doesn't complicate the feature list, and (b) it doesn't require the database instance to be in scope in the caller, unlike database-supports?

src/metabase/driver/sql_jdbc.clj
calherries: Ah right. I think that's a good idea but unnecessary as part of this change. The primary key should be separate from individual column items because in theory you should be able to create a primary key from an ordered list of columns

qnkhuat: okay, that makes sense

qnkhuat (approved):
cool, LGTM now

modules/drivers/redshift/src/metabase/driver/redshift.clj
crisptrutski: Are you still planning to add this comment?

src/metabase/driver/sql_jdbc.clj
```diff
@@ -162,11 +164,15 @@
...
-  [driver db-id table-name col->type]
-  (let [sql (first (sql/format {:alter-table (keyword table-name)
-                                :add-column (map (fn [[name type-spec]]
-                                                   (vec (cons name type-spec)))
-                                                 col->type)}
+  [driver db-id table-name column-definitions & {:keys [primary-key]}]
+  (mu/validate-throw [:maybe [:cat :keyword]] primary-key) ; we only support adding a single primary key column for now
+  (let [primary-key-column (first primary-key)
+        sql (first (sql/format {:alter-table (keyword table-name)
+                                :add-column (cond-> (map (fn [[name type-and-constraints]]
```
crisptrutski, line 171: Afraid my eyes might betray me again, but it doesn't look like this outer `cond->` has any transformation clauses.

test/metabase/upload_test.clj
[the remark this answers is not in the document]
crisptrutski: I think you can remove the uuid here now as well. Also my OCD would prefer you didn't have whitespace in a filename 😆

test/metabase/upload_test.clj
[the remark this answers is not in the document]
crisptrutski: I think you can remove this sleep as well.

test/metabase/upload_test.clj
crisptrutski: A short comment to the effect would be nice

```clojure
;; Ensure the name is unique, as `with-empty-db` unfortunately leaks tables when using redshift
```

This whole "Solve for N in Nx8" mechanism is also a bit overly complex IMHO, how about something like ` (->> (random-uuid) str cycle (take 254) (apply str))`?

crisptrutski (commented):
There's still more stuff that could be cleaned up or tightened I think, but otherwise looking good 😎

modules/drivers/redshift/src/metabase/driver/redshift.clj
calherries: done d85d308f0371d9d532c7be75ab2e31573a1b3e05

src/metabase/driver/sql_jdbc.clj
calherries: Good eyes! 71cc678749b10ef0216f4321ed8a35727f22c2be

test/metabase/upload_test.clj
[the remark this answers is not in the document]
calherries: I can't remove the random UUID because of the redshift issue. I'll clean this up after https://github.com/metabase/metabase/pull/38453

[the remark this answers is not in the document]
calherries: Actually now that I think of it, the chance of a collision doesn't change after https://github.com/metabase/metabase/pull/38453. If test table names are created at the same second in two concurrent test runs, the names will still collide.

test/metabase/upload_test.clj
calherries: Actually it will still be necessary after https://github.com/metabase/metabase/pull/38453 so I've cleaned it up here 854ab1f51131ee601d62fac0fb1d3cc2dcd0e8eb

src/metabase/upload.clj
```diff
@@ -678,7 +690,9 @@
 ;;; +--------------------------------------------------
 
 (mu/defn append-csv!
-  "Main entry point for appending to uploaded tables with a CSV file."
+  "Main entry point for appending to uploaded tables with a CSV file.
+  This will create an auto-incrementing primary key (auto-pk) column in the table for drivers that supported uploads
+  before auto-pk columns were introduced by metabase#36249."
```
crisptrutski, line 695: Why not put a clickable link for the issue / PR?

crisptrutski (approved):

src/metabase/upload.clj
calherries: A clickable link is longer so we generally just link to issues/PRs like this. I think those on emacs have set up an automation to open these as links quickly.

PR #37518, opened 2024-01-10
Add TypeScript wrappers around native CLJS functions

### LINKED ISSUE #37504: [MLv2] Add TypeScript wrappers around `native` CLJS functions
For the reference, native queries MLv2 epic is here: https://github.com/metabase/metabase/issues/31003

We already have a few TS wrappers in https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/native.ts

The goal for this task is to add the remaining ones. They are defined here
https://github.com/metabase/metabase/blob/master/src/metabase/lib/core.cljc#L241:L253

> [!Important]
> We're also referencing one of these functions directly in https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/queries/NativeQuery.ts#L434 which shouldn't happen.

```[tasklist]
### Add wrappers for the following CLJS functions
- [x] [native-query](https://github.com/metabase/metabase/pull/37518/commits/18da9248a147aa3d6b1a7b9599362987b25fd926)
- [x] [raw-native-query](https://github.com/metabase/metabase/pull/37518/commits/80648a020dc0db6242a711a75fc2e13b3d7a0826)
- [x] [with-native-query](https://github.com/metabase/metabase/pull/37518/commits/26d50ad40f18f246a0d4a57725481ab617cfae82)
- [x] [with-template-tags](https://github.com/metabase/metabase/pull/37518/commits/7db99c4184a93186b16925f6ca4ae5c1edb2aeaa)
- [x] [template-tags](https://github.com/metabase/metabase/pull/37518/commits/d44989cc961925496cff0bcd2268b787a6112e00)
- [x] [extract-template-tags](https://github.com/metabase/metabase/pull/37518/commits/273edd3bdd4cf789ba10937a7955422ffc4be6cc)
- [x] [engine](https://github.com/metabase/metabase/pull/37518/commits/2e2a46e4696484a463a5e26ab7644412ac07e28b)
- [x] [with-different-database](https://github.com/metabase/metabase/pull/37518/commits/544441cb7d8f9da99490702ee893399a34034df4)
- [x] [has-write-permission](https://github.com/metabase/metabase/pull/37518/commits/fcd703f1df888fc7ce5940c40f2407cfa89509ba)
```

nemanjaglumac:
This PR adds the remaining TS wrappers for the native CLJS functions that are registered in `core.cljc`
For the full list, please see #37504 

It also [replaces the direct usage of a ML function with a new wrapper](https://github.com/metabase/metabase/pull/37518/commits/ab624f6517f7416391d55381e78abfd9592eaa0b).

### Hint (from something I learned today)
To compare the types, check the assigned Malli schemas in https://github.com/metabase/metabase/blob/master/src/metabase/lib/native.cljc

Resolves #37504.

---
### Outstanding questions
- Should I add some thing layer of unit tests around these wrappers before merging this PR? cc @metabase/querying-components

### DIFF
--- a/frontend/src/metabase-lib/native.ts
+++ b/frontend/src/metabase-lib/native.ts
@@ -28,6 +28,13 @@ export function templateTags(query: Query): string {
   return ML.template_tags(query);
 }
 
+export function extractTemplateTags(
+  queryText: string,
+  existingTags?: string,
+): string | null {
+  return ML.extract_template_tags(queryText, existingTags);
+}
+
 /**
  * Returns the extra keys that are required for this database's native queries, for example `:collection` name is
  *  needed for MongoDB queries.
--- a/frontend/src/metabase-lib/native.ts
+++ b/frontend/src/metabase-lib/native.ts
@@ -31,7 +31,7 @@ export function templateTags(query: Query): TemplateTags {
 export function extractTemplateTags(
   queryText: string,
   existingTags?: TemplateTags,
-): TemplateTags | Record<string, never> {
+): TemplateTags {
   return ML.extract_template_tags(queryText, existingTags);
 }
 
--- a/frontend/src/metabase-lib/native.ts
+++ b/frontend/src/metabase-lib/native.ts
@@ -47,6 +47,10 @@ export function withDifferentDatabase(
   return ML.with_different_database(query, databaseId, metadata);
 }
 
+export function engine(query: Query): string {
+  return ML.engine(query);
+}
+
 /**
  * Returns the extra keys that are required for this database's native queries, for example `:collection` name is
  *  needed for MongoDB queries.
--- a/frontend/src/metabase-lib/native.ts
+++ b/frontend/src/metabase-lib/native.ts
@@ -12,6 +12,10 @@ export function nativeQuery(
   return ML.native_query(databaseId, metadata, innerQuery);
 }
 
+export function rawNativeQuery(query: Query): string {
+  return ML.raw_native_query(query);
+}
+
 /**
  * Returns the extra keys that are required for this database's native queries, for example `:collection` name is
  *  needed for MongoDB queries.
--- a/frontend/src/metabase-lib/native.ts
+++ b/frontend/src/metabase-lib/native.ts
@@ -39,6 +39,14 @@ export function hasWritePermission(query: Query): boolean {
   return ML.has_write_permission(query);
 }
 
+export function withDifferentDatabase(
+  query: Query,
+  databaseId: DatabaseId,
+  metadata: MetadataProvider,
+): unknown {
+  return ML.with_different_database(query, databaseId, metadata);
+}
+
 /**
  * Returns the extra keys that are required for this database's native queries, for example `:collection` name is
  *  needed for MongoDB queries.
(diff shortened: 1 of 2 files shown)


frontend/src/metabase-lib/native.ts
```diff
@@ -4,6 +4,53 @@ import type { DatabaseId } from "metabase-types/api";
...
+  return ML.with_template_tags(query, tags);
+}
+
+export function templateTags(query: Query): string {
+  return ML.template_tags(query);
+}
+
+export function extractTemplateTags(
+  queryText: string,
+  existingTags?: string,
```
ranquild, line 33: existingTags - `TemplateTag` (array or an object with them?)

ranquild: Checked - it should be `Record<string, TemplateTag>`

frontend/src/metabase-lib/native.ts
```diff
@@ -4,6 +4,53 @@ import type { DatabaseId } from "metabase-types/api";
...
+
+export function withNativeQuery(query: Query, innerQuery: string): Query {
+  return ML.with_native_query(query, innerQuery);
+}
+
+export function withTemplateTags(query: Query, tags: string[]): Query {
+  return ML.with_template_tags(query, tags);
+}
+
+export function templateTags(query: Query): string {
```
ranquild, line 27: Perhaps `TemplateTag[]` or `Record<string, TemplateTag>`? Check what the function actually returns

frontend/src/metabase-lib/native.ts
```diff
@@ -4,6 +4,53 @@ import type { DatabaseId } from "metabase-types/api";
...
+
+export function rawNativeQuery(query: Query): string {
+  return ML.raw_native_query(query);
+}
+
+export function withNativeQuery(query: Query, innerQuery: string): Query {
+  return ML.with_native_query(query, innerQuery);
+}
+
+export function withTemplateTags(query: Query, tags: string[]): Query {
```
ranquild, line 23: Also not string

frontend/src/metabase-lib/native.ts
nemanjaglumac: We have the equivalent type for this already 🎉 
https://github.com/metabase/metabase/blob/master/frontend/src/metabase-types/api/dataset.ts#L145

frontend/src/metabase-lib/native.ts
```diff
@@ -1,9 +1,56 @@
...
+}
+
+export function templateTags(query: Query): TemplateTags {
+  return ML.template_tags(query);
+}
+
+export function extractTemplateTags(
+  queryText: string,
+  existingTags?: TemplateTags,
+): TemplateTags | Record<string, never> {
```
nemanjaglumac, line 34: The docstring says:
> Invalid patterns are simply ignored, so something like `{{&foo!}}` is just disregarded.

So I initially (erroneously) thought this falls back to `null`. But playing around in REPL confirmed that this function returns just an empty object (map in Clojure) when it encounters an invalid pattern.

ranquild: I think in that case I'd leave only `TemplateTags` as a type because an empty object satisfies it.

ranquild (approved):
Left 1 comment about the type

frontend/src/metabase-lib/native.ts
nemanjaglumac: Works for me.
Fixed in https://github.com/metabase/metabase/pull/37518/commits/9ab6688f07a972c3d3ca3f51863bfc10a21ec96d

Thank you for the thorough review.

PR #37519, opened 2024-01-10
Add validation for remaining setting types

crisptrutski:
### Description

This extends https://github.com/metabase/metabase/pull/37313 to cover the rest of the settings formats, other than `:string`.

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1361,26 +1361,36 @@
   (ex-info (trs "Error of type {0} thrown while parsing a setting" (type ex))
            {:ex-type (type ex)}))
 
-(defn- may-contain-raw-token? [ex setting]
-  (case (:type setting)
-    :json
-    (cond
-      (instance? JsonEOFException ex) false
-      (instance? JsonParseException ex) true
-      :else (do (log/warn ex "Unexpected exception while parsing JSON")
-                ;; err on the side of caution
-                true))
+(defmulti may-contain-raw-token?
+  "Indicate whether we must redact an exception to avoid leaking sensitive env vars"
+  (fn [_ex setting] (:type setting)))
 
-    :boolean false
-    :csv (not (instance? java.io.EOFException ex))
-    :double true
-    :keyword true
-    :integer true
-    :positive-integer true
-    :timestamp true
+;; fallback to redaction if we have not defined behaviour for a given format
+(defmethod may-contain-raw-token? :default [_ _] false)
 
-    ;; fallback to redaction if we have not defined behaviour for a given format
-    true))
+(defmethod may-contain-raw-token? :boolean [_ _] false)
+
+;; Non EOF exceptions will mention the next character
+(defmethod may-contain-raw-token? :csv [ex _] (not (instance? java.io.EOFException ex)))
+
+;; Number parsing exceptions will quote the entire input
+(defmethod may-contain-raw-token? :double [_ _] true)
+(defmethod may-contain-raw-token? :integer [_ _] true)
+(defmethod may-contain-raw-token? :positive-integer [_ _] true)
+
+;; Date parsing may quote the entire input, or a particular sub-portion, e.g. a misspelled month name
+(defmethod may-contain-raw-token? :timestamp [_ _] true)
+
+;; Keyword parsing can never fail, but let's be paranoid
+(defmethod may-contain-raw-token? :keyword [_ _] true)
+
+(defmethod may-contain-raw-token? :json [ex _]
+  (cond
+    (instance? JsonEOFException ex) false
+    (instance? JsonParseException ex) true
+    :else (do (log/warn ex "Unexpected exception while parsing JSON")
+              ;; err on the side of caution
+              true)))
 
 (defn- redact-sensitive-tokens [ex raw-value]
   (if (may-contain-raw-token? ex raw-value)
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1379,8 +1379,8 @@
     :positive-integer true
     :timestamp true
 
-    (do (log/warn "Redaction criteria not defined for type and obtain an exception of type" (:type setting) (type ex))
-        true)))
+    ;; fallback to redaction if we have not defined behaviour for a given format
+    true))
 
 (defn- redact-sensitive-tokens [ex raw-value]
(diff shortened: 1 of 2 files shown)


test/metabase/models/setting_test.clj
```diff
@@ -1204,45 +1204,189 @@
...
+    nil
+    (catch java.lang.Exception e e)) )
+
+(defn- get-json-parse-exception [raw-value]
+  (mt/with-temp-env-var-value [mb-test-json-setting raw-value]
+    (get-formatting-exception)))
+
+(defn- get-csv-parse-exception [raw-value]
+  (mt/with-temp-env-var-value [mb-test-csv-setting raw-value]
+    (get-formatting-exception)))
```
crisptrutski, line 1231: I could double down on drying things up with macros, but I thought it was nice for my IDE to know about these symbols. I could go in the other direction and reduce the tests all the way and make the tests completely macro driven, e.g.

```clojure 
(assert-valid :json "eg1" "eg2" ...)

(assert-invalid :json 
  "blag" #"Exception message pattern"
  ...)
  ```

File: test/metabase/models/setting_test.clj [hunk 2]
```diff
@@ -1204,45 +1204,189 @@
...
+(defmacro define-setting-for-type [format]
+  `(defsetting ~(symbol (str "test-" (name format) "-setting"))
+     "Setting to test validation of this format - this only shows up in dev"
+     :type ~(keyword (name format))))
+
+(defmacro define-settings-for-types [& formats]
+  (list 'dorun (cons 'list (for [f formats] `(define-setting-for-type ~f)))))
+
+;; Create stub settings for the remaining types that weren't used by any previous tests
+(define-settings-for-types :double :keyword :integer :positive-integer :timestamp)
```
crisptrutski, line 1217: I couldn't think of a quick way to catch if a new type is added without updating this field. I could open an issue for doing that? This rabbit hole keeps going 🙂

test/metabase/models/setting_test.clj
calherries: You could DRY up these so you can write `(get-parse-exception mb-test-csv-setting)` in the test, and that would be tidier to me. I think the limit would be removing the assertions from the tests. It might be fewer characters but it doesn't add much readability IMO.

src/metabase/models/setting.clj
```diff
@@ -1371,8 +1371,16 @@
...
-    true))
+    :boolean false
+    :csv (not (instance? java.io.EOFException ex))
+    :double true
+    :keyword true
+    :integer true
+    :positive-integer true
+    :timestamp true
+
+    (do (log/warn "Redaction criteria not defined for type and obtain an exception of type" (:type setting) (type ex))
```
calherries, line 1382: I think we can skip this log. logs should make troubleshooting easier, not for leaving TODOs for us developers

test/metabase/models/setting_test.clj
```diff
@@ -1204,45 +1204,189 @@
...
-          ex (get-parse-exception "{\"a\": \"123abc\", \"b\": 2")]
+          ex (get-json-parse-exception "{\"a\": \"123abc\", \"b\": 2")]
       (is (not (str/includes? (pr-str ex) password)))
       (assert-parser-exception!
-        ex
+        :json ex
         (str "Unexpected end-of-input: expected close marker for Object (start marker at [Source: REDACTED"
              " (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1])\n"
              " at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 23]")))))
+
```
calherries, line 1297:
```suggestion
```

src/metabase/models/setting.clj
crisptrutski: My desire was for it to prompt anyone writing a new settings format to "dot the i", but thinking about it more I see how it could easily not be triggered on the dev branch, fall through into a release, and be seen by a user.

Do you think it's fine to remove this without adding any other "coverage protection" mechanism?

test/metabase/models/setting_test.clj
crisptrutski: Nice suggestion, I think I can riff off that.

test/metabase/models/setting_test.clj
crisptrutski: Even with Refined Github, I can't see what's changed in this diff. Oh well, accepting on faith 😄

test/metabase/models/setting_test.clj
calherries: As we discussed I think this hits the limit of where we want to go for effort vs reward with regards to future-proofing. If a new setting type is created but we forget to handle custom validation logic.

Anyway, I think there's a simpler pattern than the `define-settings-for-types` macro.

```
(doseq [type setting-types]
  (define-setting-for-type type))
```

 Thankfully clojure is an imperative language too :)

calherries: It might be a good exercise to think about how to enforce this constraint if we wanted though. I think a check on executing a defsetting macro would do it, similar to how you're doing it in [this PR](https://github.com/metabase/metabase/pull/37624/files#diff-1fb7cd6dc14f96149b2b775f598268d8053ee68dca276c40001b41ca4965cff6), but the implementation could be done in a few ways.

crisptrutski: Found a great way to enumerate them: `(set (keys (methods get-value-of-type)))`

crisptrutski: > Anyway, I think there's a simpler pattern than the define-settings-for-types macro.

I don't think that doseq works, because the macro will only see the symbol `type, or am I missing something?

crisptrutski: I decided there wasn't much point in making this a one liner, and actually it's quite nice to use the definitions as "headers" for each set of tests. 

I added a test at the end to validate that we haven't missed any sections (i.e. we've defined all the header vars). 

I also turned the method which we removed the logging from into a multimethod so that we could also validate its completeness.

test/metabase/models/setting_test.clj
calherries: I just suggested to remove an extra new line.

calherries (approved):
LGTM. Nicely done

PR #37430, opened 2024-01-09
Embedding settings cleanup - Milestone 2

npretto:
This is the second and last milestone of [[Epic]Embedding settings cleanup and surface interactive embedding Quickstart ](https://github.com/metabase/metabase/issues/36481)

It contains code from the following pull requests:
- [fix color button when it's disabled](https://github.com/metabase/metabase/pull/37053)
- [new utm tags for embed settings](https://github.com/metabase/metabase/pull/37403)
- [Use new docs url for quickstart link](https://github.com/metabase/metabase/pull/37523)

Closes https://github.com/metabase/metabase/issues/36481

PR #37523, opened 2024-01-10
Use new docs url for quickstart link

npretto:
Part of [[Epic]Embedding settings cleanup and surface interactive embedding Quickstart ](https://github.com/metabase/metabase/issues/36481)

### Description

This PR updates the url as per

https://github.com/metabase/metabase/pull/37430#discussion_r1447576262
> The Quickstart has since moved.
> There is a redirect in place, so in theory there is nothing wrong, but I prefer we link to the correct URL now.

I'm using the util `getUpgradeUrl` to create the new url, the util appends `.html` to the end of the link.

I'm 60/40 if the tests should use a specific version or should just use the default version and end up with `/latest/` urls.
The util is already tested [here](https://github.com/metabase/metabase/blob/ec30700ac7821dfcd02a75008a50d87a8a377ce2/frontend/test/metabase/lib/settings.unit.spec.js#L13), and I've seen other tests that don't specify a version and just test for `/latest/` urls.
I have a slight preference for the actual versions used in test, but I split the changes in two commits so we can revert the tests to check for `/latest` if you prefer that

### Checklist

### DIFF
--- a/frontend/src/metabase/admin/settings/app/components/SettingsEditor/tests/embedding/common.embedding.unit.spec.tsx
+++ b/frontend/src/metabase/admin/settings/app/components/SettingsEditor/tests/embedding/common.embedding.unit.spec.tsx
@@ -72,12 +72,15 @@ describe("[OSS] embedding settings", () => {
 
       it("should link to quickstart for interactive embedding", async () => {
         await setupEmbedding({
-          settingValues: { "enable-embedding": false },
+          settingValues: {
+            "enable-embedding": false,
+            version: { tag: "v0.49.3" },
+          },
         });
         expect(getQuickStartLink()).toBeInTheDocument();
         expect(getQuickStartLink()).toHaveProperty(
           "href",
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/admin/settings/app/components/SettingsEditor/tests/embedding/common.embedding.unit.spec.tsx
```diff
@@ -72,12 +72,15 @@ describe("[OSS] embedding settings", () => {
 
       it("should link to quickstart for interactive embedding", async () => {
         await setupEmbedding({
-          settingValues: { "enable-embedding": false },
+          settingValues: {
+            "enable-embedding": false,
+            version: { tag: "v0.49.3" },
```
npretto, line 77: I used realistic values here, so v0.xx here, and v1.xx in the premium/ee tests

oisincoveney (approved):

PR #37079, opened 2023-12-22
[MLv2] Resolve a string `:fk-field-id` to a numeric ID if possible

bshepherdson:
Progress towards #37067, might not fix it entirely. There may be other
bugs stemming from the string `:fk-field-id`, but I can't reproduce it
locally.

Issue #7335, opened 2018-04-12
[Enhancement] Data Model: Attribute as Type "Foreign Key" Join on multiple Columns


nsiicm0:
This is my first ever issue reported on metabase, so bear with me.

I love the way how one can easily specify a foreign table to be joined using the foreign key during the data modeling process.

Currently it is only possible to point to one primary key in a foreign table using the foreign key. Which makes total sense in most of the cases.
For illustration purposes I will try to visualize this behavior.

```
-----------------                                       -------------------
| tblProducts   |                                       | tblParams       |
| - Key         |                                       | - PREFIX        |
| - Name        |                                       | - Key           |
| - Status      | n --------------------------------- 1 | - Name          |
-----------------                                       -------------------
```
In this illustration I have the relation of a product table which is has a 1:n reference on a params table, which basically contains the prolonged name of the product status. This way I can store the product status as the foreignkey of the params table inside my product table. 
This is in accordance to the 3rd normal form. Metabase correctly allows me to do that during the data modelling process.

Now comes my enhancement proposal:
As you might have guessed, my table is called "params". For whatever reason, the developer created this table in mind to house various attributes (not just our status). Let's assume the params table contains also product groups. To distinguish the keys inside the params table, the developer also introduced a field called PREFIX. A sample params table could look like this:

```
--------------------------
| PREFIX | Key | Name    |
--------------------------
| STATUS | 1   | Active  |
| STATUS | 2   | Retired |
| STATUS | 3   | Dead    |
| GROUP  | 1   | Chair   |
| GROUP  | 2   | Bench   |
--------------------------
```
My table relation would look like this:
```
-----------------                                       -------------------
| tblProducts   |                                       | tblParams       |
| - Key         |                                       | - PREFIX        |
| - Name        |                                       | - Key           |
| - Status      | n --------------------------------- 1 | - Name          |
| - Group       | n --------´                           -------------------
-----------------                                       
```

Now here is my problem. If I would go ahead and model my table products in Metabase and substitute my attributes Status and Group using the field type foreign key during the model process I will run into my problem because I cannot specify my PREFIX on which it should join.
Currently the SQL statement would look like this:
```
-- Query 1
SELECT
    p.Key,
    p.Name AS Name,
    status.Name AS Status,
    group.Name AS Group
FROM
    tblProducts
    LEFT OUTER JOIN tblParams status ON p.Status = status.Key
    LEFT OUTER JOIN tblParams group ON p.Group = group.Key
```
[...]

PR #37536, opened 2024-01-10
[MLv2] Only allow implicit join on a column if its real ID is known

### LINKED ISSUE #37067: `displayInfo` returns unexpected result for a column group
**Context**

To repro the bug:
1. [Go to this question](https://stats.metabase.com/model/15002-active-metabase-user-stats)
2. Click on the filter button

```
return groups.map((group, groupIndex) => {
      const groupInfo = Lib.displayInfo(query, stageIndex, group);
      ....
```

groupInfo is incorrect for a [question from stats](https://stats.metabase.com/model/15002-active-metabase-user-stats)

FE expects the next interface:

```
name: string;
displayName: string;
isSourceTable: boolean;
isFromJoin: boolean;
isImplicitlyJoinable: boolean;
fkReferenceName?: string;
```

And the returned value is different:

image

bshepherdson:
This is how legacy worked but MLv2 changed this behavior. Even if the
metadata for a native column includes `:semantic-type :type/FK` and
`:fk-target-field-id`, it can't be used for a JOIN unless it's a real
field in the database.

This undoes some hacks from #37079 that were misguided. Fixes #37067.

### DIFF
--- a/src/metabase/lib/column_group.cljc
+++ b/src/metabase/lib/column_group.cljc
@@ -1,12 +1,10 @@
 (ns metabase.lib.column-group
   (:require
    [metabase.lib.card :as lib.card]
-   [metabase.lib.equality :as lib.equality]
    [metabase.lib.join :as lib.join]
    [metabase.lib.join.util :as lib.join.util]
    [metabase.lib.metadata :as lib.metadata]
    [metabase.lib.metadata.calculation :as lib.metadata.calculation]
-   [metabase.lib.options :as lib.options]
    [metabase.lib.schema.common :as lib.schema.common]
    [metabase.lib.schema.id :as lib.schema.id]
    [metabase.lib.util :as lib.util]
@@ -47,8 +45,7 @@
          (>= (count (keys (select-keys m [:join-alias :table-id :card-id]))) 1))]]]
     [:group-type/join.implicit
      [:map
-      ;; TODO: Strings are allowed here to work around a QP bug; see #37067.
-      [:fk-field-id [:or [:ref ::lib.schema.id/field] ::lib.schema.common/non-blank-string]]]]]])
+      [:fk-field-id [:ref ::lib.schema.id/field]]]]]])
 
 (defmethod lib.metadata.calculation/metadata-method :metadata/column-group
   [_query _stage-number column-group]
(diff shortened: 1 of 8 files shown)


bshepherdson:
* `master`
  * **PR #37536** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37536?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37536?utm_source=stack-comment).

src/metabase/lib/metadata/calculation.cljc
```diff
@@ -545,9 +545,9 @@
 
 (defn implicitly-joinable-columns
   "Columns that are implicitly joinable from some other columns in `column-metadatas`. To be joinable, the column has to
-  have appropriate FK metadata, i.e. have an `:fk-target-field-id` pointing to another Field. (I think we only include
-  this information for Databases that support FKs and joins, so I don't think we need to do an additional DB feature
-  check here.)
+  have (1) appropriate FK metadata, i.e. have an `:fk-target-field-id` pointing to another Field, and (2) have a numeric
+  `:id`, i.e. be a real database column that can be used in a JOIN condition. (I think we only include this information
+  for Databases that support FKs and joins, so I don't think we need to do an additional DB feature check here.)
```
snoe, line 550: do we implicitly join compound keys?

src/metabase/lib/column_group.cljc
```diff
@@ -97,18 +94,7 @@
 (defmethod display-info-for-group-method :group-type/join.implicit
   [query stage-number {:keys [fk-field-id], :as _column-group}]
   (merge
-   (when-let [field (if (string? fk-field-id)
-                      ;; TODO: This is probably working around a QP bug. See #37067 - the `data.cols` from the API has
```
snoe, line 101: Do we have a qp test that would produce a stringy `fk-field-id`? I'm worried that this will crop again.

snoe (approved):
A couple questions, but otherwise looks good.

src/metabase/lib/metadata/calculation.cljc
bshepherdson: I'm not sure what you mean?

src/metabase/lib/column_group.cljc
bshepherdson: See the earlier PR #37079 - this logic was added speculatively trying to fix a stats-only issue. It didn't work, and this PR is the result. The string IDs were a symptom of the bug being fixed by this PR, so this logic can be removed now that the bug is properly fixed. There are tests for the real requirement here - FKs on SQL models.

src/metabase/lib/metadata/calculation.cljc
snoe: for some reason I thought we had something like this already. https://github.com/metabase/metabase/issues/7335
Obviously, irrelevant to this issue at this point.

src/metabase/lib/column_group.cljc
snoe: Understood

PR #37539, opened 2024-01-10
Optimize canonicalization of MBQL clauses

### LINKED ISSUE #37245: Mongo GUI question is way slower than it's native twin
### Describe the bug

Something that we identified while debugging it on a customer instance (and we couldn't repro locally): a GUI query is way slower (5x-10x) than it's native counterpart, but there's no delay on the query processor (the /native endpoint is fast)

### To Reproduce

We can't reproduce this locally, but we have all the traces here https://drive.google.com/drive/folders/1cNGbnepp_6YXTYO4YlJrqYHDHeCoJOGG?usp=sharing

### Expected behavior

GUI query should be only a few milliseconds slower than it's native counterpart

### Logs

NA

```JSON
v47.x
```

### Severity

P1

### Additional context

I have a strong suspicion this is MLv2 stuff

metamben:
Fixes #37245.

I created a query (see below) similar to that in the issue, and the bottleneck was the canonicalization of MBQL clauses in the compiled native query. The optimization in the PR makes difference between the MBQL and the native queries acceptable.

Unfortunately, this repro query is not quite as slow as in the reported issue (I measured 1.5--3 seconds, the reported issue is about ~ 6 seconds), so it's possible that this optimization is not enough.

image

### How to verify

Measure the time needed for getting the canonical form of some sample query fragments with and without the changes in this PR:

```clojure
(let [fast
      {:native
       {:query
        [{"$group"
          {"_id"
           {"created_at" {:$dateTrunc {:date "$created_at", :unit "week", :timezone "UTC", :startOfWeek "sunday"}},
            "rating" "$rating"},
           "count" {"$addToSet" "$reviewer"}}}]}}
      slow
      {:native
       {:query
        [{"$group"
          {"_id"
           {"created_at"
            {:$let
             {:vars
              {:parts
               {:$dateToParts
                {:timezone "UTC",
                 :date
                 {"$subtract"
                  ["$created_at"
                   {"$multiply"
                    [{"$subtract"
                      [{:$let
                        {:vars {:day_of_week {"$dayOfWeek" {:date "$created_at", :timezone "UTC"}}},
                         :in {"$cond" {:if {"$eq" [:$$day_of_week 0]}, :then 7, :else :$$day_of_week}}}}
                       1]}
                     86400000]}]}}}},
              :in {:$dateFromParts {:timezone "UTC", :year "$$parts.year", :month "$$parts.month", :day "$$parts.day"}}}},
            "rating" "$rating"},
           "count" {"$addToSet" "$reviewer"}}}]}}]
  (doseq [q [slow fast]]
    (time (canonicalize-mbql-clauses q))))
```

The first (`fast`) fragment is generated for the query in the picture if Mongo 5 or newer is used, the second (`slow`) is generated for Mongo versions earlier than 5. The reported issue was in the context of Mongo 4, as far as we know.


src/metabase/mbql/normalize.cljc
```diff
@@ -639,25 +639,30 @@
...
-                           e))))))
-   mbql-query))
+  [x]
+  (cond
+    (map? x)
+    (m/map-vals canonicalize-mbql-clauses x)
+
+    (not (mbql-clause? x))
+    (if (sequential? x)
+      (into (empty x) (map canonicalize-mbql-clauses) x)
```
metamben, line 649: This reverses lists and lazy sequences, I have to change this, if those can appear here.

src/metabase/mbql/normalize.cljc
```diff
@@ -639,25 +641,40 @@
...
+            (canonicalize-mbql-clause form)
+            (catch #?(:clj Throwable :cljs js/Error) e
+              (log/error (i18n/tru "Invalid clause:") form)
+              (throw (ex-info (i18n/tru "Invalid MBQL clause: {0}" (ex-message e))
+                              {:clause form}
+                              e))))]
+      (if (seq top-canonical)
+        (into (conj (empty top-canonical) (first top-canonical))
+              (map canonicalize-mbql-clauses)
+              (rest top-canonical))
```
metamben, line 664: This is assuming that the canonical MBQL clause is not a list, a lazy sequence or a map. If that's not true, this will have to get more complicated.

src/metabase/mbql/normalize.cljc
```diff
@@ -639,25 +639,40 @@
...
-                           {:clause x}
-                           e))))))
-   mbql-query))
+  [form]
+  (cond
+    (record? form)
+    (reduce-kv (fn [r k x] (assoc r k (canonicalize-mbql-clauses x))) form form)
+
+    (map? form)
+    (m/map-vals canonicalize-mbql-clauses form)
```
camsaul, line 648: Not really asking you to change this since it's existing code, but just wanted to point out that instead of using `m/map-vals` we should probably just use `update-vals` instead. `m/map-vals` isn't really needed now that there's an official Clojure core function that does the same thing. I have a PR to enforce this with a Kondo linter https://github.com/metabase/metabase/pull/32423 that I'm going to hopefully get green at some point

src/metabase/mbql/normalize.cljc
```diff
@@ -774,7 +785,7 @@
       query           (update :query canonicalize-inner-mbql-query)
       parameters      (update :parameters (partial mapv canonicalize-mbql-clauses))
       native          (update :native canonicalize-native-query)
-      true            canonicalize-mbql-clauses)
+      true            canonicalize-mbql-clauses-excluding-native)
```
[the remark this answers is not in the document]
camsaul: I think it should be ok, I'm struggling to think of a counterexample. Native template tags `:dimension` has some weird pseudo field references there but I don't think we need to do anything in the "canonicalize" step.

src/metabase/mbql/normalize.cljc
```diff
@@ -639,25 +639,40 @@
...
-         (catch #?(:clj Throwable :cljs js/Error) e
-           (log/error (i18n/tru "Invalid clause:") x)
-           (throw (ex-info (i18n/tru "Invalid MBQL clause: {0}" (ex-message e))
-                           {:clause x}
-                           e))))))
-   mbql-query))
+  [form]
+  (cond
+    (record? form)
+    (reduce-kv (fn [r k x] (assoc r k (canonicalize-mbql-clauses x))) form form)
```
camsaul, line 645: It's probably worth calling out that the reason you have different behavior here for records is that `m/map-vals` and `update-vals` return plain maps rather than preserving the record type. It took me like 5 minutes to work out the reason you were doing this

src/metabase/mbql/normalize.cljc
```diff
@@ -639,25 +639,40 @@
...
+        (into (conj (empty top-canonical) (first top-canonical))
+              (map canonicalize-mbql-clauses)
+              (rest top-canonical))
+        top-canonical))
+
+    (list? form)
+    (apply list (map canonicalize-mbql-clauses form))
+
+    (seq? form)
+    (doall (map canonicalize-mbql-clauses form))
```
camsaul, line 669: I think for both lists and `seq?`s  it's probably fine to do `mapv`. With `doall` or `apply list`, you're realizing everything in memory at once anyway, might as well do things the easy way if that's the case

camsaul (approved):
Generally I think this looks good but adding a comment about why you have a separate `record?` case would be nice so people don't get confused by it like I did, also I think `mapv` is perfectly fine for dealing with sequences as opposed to `apply list`/`doall`

src/metabase/mbql/normalize.cljc
metamben: Yes, I was considering this change. I'll do it in a separate PR.

src/metabase/mbql/normalize.cljc
metamben: I'll add a comment in the followup PR.

src/metabase/mbql/normalize.cljc
metamben: I did this purely for compatibility with the previous version that wouldn't change the types of these collections, but I gladly simplify these things since we are making things canonical, so changes are to be expected anyway.

PR #37540, opened 2024-01-10
Improve admin logs page loading and error experience

### LINKED ISSUE #24621: Admin > Troubleshooting > Logs page shows spinner if no logs
This bug kinda requires another bug to surface, but still hints at error handling and robustness of the page could be improved.

The `Admin > Troubleshooting > Logs` page continuously fires requests to `api/util/logs`. If there is a problem with our logger and the response is `[]` each time, the spinner continues to show rather than some empty collection. Makes us think if the api returns a 500 for some reason the page would not have a good error message.

Low priority since it only surfaces when other issues compound with it.

#### UI when logs return empty
image

#### UI when api throws an error
image

sloansparger:
Fixes https://github.com/metabase/metabase/issues/24621

### Description

This PR introduces an error and no results states to the admin logs page. Previously the loading state would appear when one of these two scenarios occurred.

This PR also migrates the component to Typescript (I `git mv`ed the files but GitHub UI doesn't appear to be diffing it as such). I also remove the broken auto-scroll feature, fetch the logs 1 whole second earlier, and fix calling setState after unmount.

### How to verify

As mentioned in [the related issue](https://github.com/metabase/metabase/issues/24621), these states would only be a product of something being broken. You'll have to purposely break code to enter these states.

- go to admin settings
- go to troubleshooting (in navbar)
- go to logs (in sidebar)
- for verifying error state, replace the code block in `src/metabase/api/util.clj` on line 22 to return a 500 like so:
```
(api/defendpoint GET "/logs"
  "Logs."
  []
  (throw (ex-info "Some error message." {}))
  (validation/check-has-application-permission :monitoring)
  (logger/messages))
```

- for verifying no results state, it's easiest to either mock the endpoint or to hard code an empty `[]` result value in `Log.tsx`'s `fetchLogs` method

### Demo
Error view:
Screenshot 2024-01-10 at 1 12 14 PM

No results view:
Screenshot 2024-01-10 at 1 14 01 PM

### Checklist


iethree (approved):
🥳

PR #37545, opened 2024-01-10
Check backport script workflow

alxnddr:
### Description

Since recently even for backports with conflicts we started creating PRs with commits that include the `./backport.sh` script that cherry-picks the right commit and removes itself. This makes it easier to see if there are still not merged backports exist by searching for PRs with `was-backported` label. However, sometimes the `./backport.sh` script can be accidentally left so this PR introduces a workflow that won't let merge PR in that case.

[Slack convo](https://metaboat.slack.com/archives/C010L1Z4F9S/p1704888265083849)

### DIFF
--- a/.github/workflows/check-backport-script.yml
+++ b/.github/workflows/check-backport-script.yml
@@ -1,0 +1,20 @@
+name: Prevent backport.sh from being merged
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  check-backport-script:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


uladzimirdev:
@alxnddr can't this check be a part of the original logic of backporting? `auto-backport` workflow. Two checks for a single task looks a bit overcomplicated

alxnddr:
I agree that this workflow is too granular but `auto-backport` is the workflow with different triggers (pr closed, label added) and it creates the backport PR so I don't see how it is suitable for the logic in this new workflow

uladzimirdev (approved):

PR #37551, opened 2024-01-11
Backend validation for landing-page to disallow external URLs

### LINKED ISSUE #32585: Custom landing page with an external URL will lock everyone out of MB
### Describe the bug

If we customize homepage with an external URL, MB will fail to load and display a blank page.

The only way to reset that value is to update the setting using SQL manually.

### To Reproduce

1. Enable enterprise
2. Go to Admin => Appearance
3. Set the custom homepage setting with `https://google.com`
4. open MB in a new tab

### Expected behavior

should not allow entering external URLs

### Logs

_No response_

```JSON
master
```

### Severity
locking everyone out of MB is quite serious

### Additional context

_No response_

qwef:
Closes https://github.com/metabase/metabase/issues/32585

### Description

Add a setter to `landing-page` to validate URLs that come in. If the `landing-page` is a URL, then check against `site-url`. Note that I do a prefix check for this, so the same port is required.  

Related (FE version): https://github.com/metabase/metabase/pull/37322

### DIFF
--- a/src/metabase/public_settings.clj
+++ b/src/metabase/public_settings.clj
@@ -1,5 +1,6 @@
 (ns metabase.public-settings
   (:require
+   [clojure.java.io :as io]
    [clojure.string :as str]
    [java-time.api :as t]
    [metabase.api.common :as api]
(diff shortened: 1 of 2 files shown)


src/metabase/public_settings.clj
```diff
@@ -257,12 +258,29 @@
...
   (deferred-tru "Default page to show people when they log in.")
   :visibility :public
   :type       :string
   :default    ""
-  :audit      :getter)
+  :audit      :getter
+  :setter     (fn [new-landing-page]
+                (when new-landing-page
+                  ;; If the landing page is a valid URL or mailto, sms, or file, then check with if site-url has the same origin.
+                  (when (and (or (re-matches #"^(mailto|sms|file):(.*)" new-landing-page) (u/url? new-landing-page))
```
noahmoss, line 280: is there a reason now to throw for any non-HTTP/HTTPS URLs?

noahmoss (approved):

src/metabase/public_settings.clj
qwef: hm good point, I think `io/as-url` won't work without the protocol.

Issue #37227, opened 2024-01-03
[MLv2] [FE] Migrate ViewHeader to MLv2


ranquild:
### ViewHeader

**Place 1**
https://github.com/metabase/metabase/blob/b14003c6de4e3b23ea7b0bde71579d92340115c5/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx#L96

In MLv2 it could be:
```
const query = question._getMLv2Query();
const previousQuery = usePrevious(query);

useEffect(() => {
  // existing code

  const filtersCount = Lib.filters(query, -1).length;
  const previousFiltersCount = Lib.filters(previousQuery, -1).length;

  // existing code
});
```

**Place 2**
https://github.com/metabase/metabase/blob/b14003c6de4e3b23ea7b0bde71579d92340115c5/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx#L110

In MLv2 it could be:
```
Lib.aggregations(query, -1).length > 0
```

For non-mbql queries `aggregations` would be an empty list.

`-1` means the last stage of the query.

**Place 3**
https://github.com/metabase/metabase/blob/b14003c6de4e3b23ea7b0bde71579d92340115c5/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx#L419

MLv2:
```
const isNewQuery = Lib.sourceTableOrCardId(query) == null;
```

```[tasklist]
- [ ] https://github.com/metabase/metabase/pull/37375
- [ ] Migrate `ViewTitleHeaderRightSide` component
```

nemanjaglumac:
Migrated steps 1 and 2 from the original issue description.
The only outstanding reference to the MLv1 in this file is (step 3):
https://github.com/metabase/metabase/blob/b14003c6de4e3b23ea7b0bde71579d92340115c5/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx#L419

but we cannot fully migrate it until we start supporting native queries with MLv2.

cc @NevRA

PR #37566, opened 2024-01-11
Remove redundant (and wrong) `hasData` check from ViewHeader component

### LINKED ISSUE #37516: [MLv2] [FE] ViewHeader - hasData check
https://github.com/metabase/metabase/blob/33d27d788fbb0c3a6798b1b364f67afa0bd04781/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx#L420

Can be done similar to https://github.com/metabase/metabase/issues/37515 but with `sourceTableOrCardId` for MBQL queries

nemanjaglumac:
This PR resolves #37516 by ~~migrating the `hasData` check to MLv2 for cases when the query is the instance of a StructuredQuery~~ removing the redundant check altogether.

~~We might soon be able to do this for the native query check as well, but for now it has to use the old version of the lib.
Once we migrate the second (native) part to MLv2, we can mark #37227 as done.~~

As it turned out that `hasData()` is not even needed in this file, it removes the need to handle structured and native queries separately.

This PR also marks `hasData()` methods on StructuredQuery and NativeQuery as deprecated.

### DIFF
--- a/e2e/test/scenarios/question/reproductions/18978-18977-nested-question-nodata-user.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/18978-18977-nested-question-nodata-user.cy.spec.js
@@ -4,7 +4,6 @@ import {
   appBar,
   popover,
   openNavigationSidebar,
-  leftSidebar,
   visitQuestion,
   POPOVER_ELEMENT,
 } from "e2e/support/helpers";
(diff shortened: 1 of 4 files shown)


ranquild (approved):

nemanjaglumac:
Failing E2E test seems like a legit failure because the Save button is missing.
https://app.replay.io/recording/e2etestscenariospermissionsreproductions22447-illogical-ui-elements-for-nodatacyspecjs--20a89267-7dda-4df1-8b11-bfef7f380133

The check that renders Save button relies on `(isNewQuery || canEditQuery)`.
It seems that the MLv2 replacement is not 1:1 the same as MLv1. I'll have to investigate further @ranquild

ranquild:
@nemanjaglumac let me know if you need any help with debugging

Issue #37506, opened 2024-01-10
[MLv2] [FE] Remove setDefaultQuery


ranquild:
See https://metaboat.slack.com/archives/C04CYTEL9N2/p1704811032663039

`setDefaultQuery` was only used to set default Google Analytics clauses. Since we don't want to support this anymore, `setDefaultQuery` becomes a no-op. It should be removed completely.

PR #37567, opened 2024-01-11
Remove `isEmpty` method in `StructuredQuery`

### LINKED ISSUE #37515: [MLv2] [FE] DatasetEditor - isEmpty check
https://github.com/metabase/metabase/blob/d1df0c6c68b5ff050f3c9338645e97ddc4da2fe6/frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx#L408

Could be something similar to this:
```
const isEmpty = question.isStructured() 
  ? Lib.databaseID(question.query()) == null
  : question.legacyQuery().isEmpty()
```

nemanjaglumac:
This PR resolves #37515 by migrating the `isEmpty` check for models (datasets) in `DatasetEditor` component.

After that migration, it's now possible to completely remove `isEmpty()` method from the StructuredQuery class.
This PR achieves that.

### DIFF
--- a/frontend/src/metabase-lib/queries/StructuredQuery.ts
+++ b/frontend/src/metabase-lib/queries/StructuredQuery.ts
@@ -129,13 +129,6 @@ class StructuredQuery extends AtomicQuery {
   /* Query superclass methods */
 
   /**
-   * @returns true if this is new query that hasn't been modified yet.
-   */
-  isEmpty() {
-    return !this._databaseId();
-  }
-
-  /**
    * @returns true if this query is in a state where it can be run.
    */
   canRun() {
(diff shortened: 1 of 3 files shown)


frontend/src/metabase-lib/queries/StructuredQuery.ts
```diff
@@ -278,11 +271,14 @@ class StructuredQuery extends AtomicQuery {
   setDefaultQuery(): StructuredQuery {
     const table = this.table();
 
+    const query = this.getMLv2Query();
+    const isEmpty = Lib.databaseID(query) == null;
```
nemanjaglumac, line 275: This is just an intermediary step until I remove `setDefaultQuery` in a separate PR.

See: https://github.com/metabase/metabase/issues/37506

ranquild (approved):

PR #37571, opened 2024-01-11
Pass email and siteURL to metabase.com/help/connect

gusaiani:
### To verify

1. Use a fresh install of Metabase, or remove all db connections.
2. Use an enterprise token.
3. Hover on "Get help connecting"

image

You should see an URL with get params like http://www.metabase.com/help/connect?email=gus@metabase.com&site_url=http://localhost:3000

If you click on it, metabase.com will (for now) redirect the page to Calendly. This will be updated in a separate PR.

### DIFF
--- a/frontend/src/metabase/nav/components/DatabasePromptBanner/DatabasePromptBanner.tsx
+++ b/frontend/src/metabase/nav/components/DatabasePromptBanner/DatabasePromptBanner.tsx
@@ -1,7 +1,8 @@
 import { t } from "ttag";
 import type { Location } from "history";
 
-import MetabaseSettings from "metabase/lib/settings";
+import { getSetting } from "metabase/selectors/settings";
+import { useSelector } from "metabase/lib/redux";
 
 import Link from "metabase/core/components/Link/Link";
 import { trackDatabasePromptBannerClicked } from "metabase/nav/analytics";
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/nav/components/DatabasePromptBanner/DatabasePromptBanner.tsx
```diff
@@ -27,12 +29,16 @@ export function DatabasePromptBanner({ location }: DatabasePromptBannerProps) {
     "/admin/databases/create",
   );
 
+  const adminEmail = MetabaseSettings.get("admin-email");
```
npretto, line 32: When possible we use `useSelector(state => getSetting(state, key))` as `MetabaseSettings.get` is deprecated.

Other than this, LGTM

frontend/src/metabase/nav/components/DatabasePromptBanner/DatabasePromptBanner.tsx
```diff
@@ -27,12 +29,16 @@ export function DatabasePromptBanner({ location }: DatabasePromptBannerProps) {
...
+  const siteURL = MetabaseSettings.get("site-url");
+  const helpURL = `https://metabase.com/help/connect?email=${adminEmail}&site_url=${siteURL}`;
```
ranquild, line 34: You would also need url encode them. You could do it like this:
```
const helpUrl = new URL("https://metabase.com/help/connect")
helpUrl.searchParams.set("email", adminEmail);
helpUrl.searchParams.set("site_url", siteUrl);

...

href={helpUrl.href}
```

ranquild (approved):
Approving to not block, but please address comments both about `useSelector` and encoding.

PR #37574, opened 2024-01-11
Remove aggregate, breakout, filter methods from Question

### LINKED ISSUE #37509: [MLv2] [FE] Remove aggregate, breakout, filter methods from Question
There should be no `Question`-level functions like this. Migrate places that use them to use the legacy StructuredQuery and remove these functions from the `Question` class.

Before:
```
const newQuestion = question.aggregate(["count"]);
```

After:
```
const newQuestion = question
  .legacyQuery({ useStructuredQuery: true })
  .aggregate(["count"])
  .question();
;
```

kamilmielnik:
Closes #37509


frontend/src/metabase-lib/Question.ts
```diff
@@ -194,7 +189,13 @@ class Question {
       console.warn("Unknown query type: " + datasetQuery?.type);
   });
 
-  legacyQuery({ useStructuredQuery }: { useStructuredQuery?: boolean } = {}) {
+  legacyQuery<UseStructuredQuery extends boolean>({
```
kamilmielnik, line 192: This is generic to help with the following error in `filter` call in `ObjectDetailView`:

```
Property 'filter' does not exist on type 'AtomicQuery | StructuredQuery'.
```

ranquild (approved):

Issue #37528, opened 2024-01-10
[MLv2] [FE] [Bug] Cache expressionableColumns results


ranquild:
Context https://metaboat.slack.com/archives/C04CYTEL9N2/p1704908756231269

We now call `expressionableColumns` on every keystroke (for suggestions) and for every token in the parsed query. Computing the list of columns is a slow operation and `expressionableColumns` should be called only once for the same query and results should be cached.

Same for available metrics and segments.

uladzimirdev:
Both FE and BE fixes are merged, closing

PR #37584, opened 2024-01-11
Improve performance of ExpressionEditor

uladzimirdev:
Related to https://github.com/metabase/metabase/issues/37528

### Description

Memoization will be added on BE side, but on FE we called MLv2 methods too often (3 times for every key down), this PR added debouncing and doesn't allow triggering MLv2 methods too often

### How to verify

Go to notebook editor and add expression - slow down CPU via dev tools

### Demo

After

https://github.com/metabase/metabase/assets/125459446/440a77bb-94cb-4d52-95ca-380ed968b095

Before

https://github.com/metabase/metabase/assets/125459446/e0d8f379-e42b-44f3-85bf-f885c4dbe51a

### Checklist

### DIFF
--- a/e2e/test/scenarios/question/reproductions/17963-mongo-filter-expression-compare-two-fields.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/17963-mongo-filter-expression-compare-two-fields.cy.spec.js
@@ -1,5 +1,11 @@
 import { WRITABLE_DB_ID } from "e2e/support/cypress_data";
-import { restore, popover, visualize, openTable } from "e2e/support/helpers";
+import {
+  restore,
+  popover,
+  visualize,
+  openTable,
+  getNotebookStep,
+} from "e2e/support/helpers";
 
 describe("issue 17963", { tags: "@mongo" }, () => {
   beforeEach(() => {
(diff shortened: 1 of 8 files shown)


e2e/test/scenarios/question/reproductions/17963-mongo-filter-expression-compare-two-fields.cy.spec.js
```diff
@@ -18,28 +24,21 @@ describe("issue 17963", { tags: "@mongo" }, () => {
...
     cy.findByRole("button", { name: "Filter" }).click();
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("Add filters to narrow your answer").click();
 
     popover().contains("Custom Expression").click();
 
     typeAndSelect([
       { string: "dis", field: "Discount" },
-      { string: "> qu", field: "Quantity" },
+      { string: "> quan", field: "Quantity" },
```
uladzimirdev, line 32: `qu` gives 2 results, interesting, how did it work before

frontend/src/metabase-lib/expression.ts
```diff
@@ -36,7 +36,7 @@ export function expressions(
 
 export function expressionableColumns(
   query: Query,
-  stageIndex: number,
+  stageIndex?: number,
```
uladzimirdev, line 39: BE method claims stageIndex is optional

frontend/src/metabase/query_builder/components/expressions/ExpressionEditorTextfield/ExpressionEditorTextfield.tsx
```diff
@@ -462,9 +462,9 @@ class ExpressionEditorTextfield extends React.Component<
     if (this.props.onBlankChange) {
       this.props.onBlankChange(source.length === 0);
     }
-  }
+  };
 
-  handleCursorChange(selection: Ace.Selection) {
+  handleCursorChange = _.debounce((selection: Ace.Selection) => {
```
uladzimirdev, line 467: probably the most important, change, based on my investigations, `ExpressionEditorTextfield` component is rendered 3 times per keypress, most likely because of separate updates of the state. I spent some time trying to fix the reason, but didn't find how to fix it. 

Anyways, having a small delay between typing and asking for columns from MLv2 methods reduces number of requests from 3 to 1

frontend/src/metabase/query_builder/components/expressions/ExpressionEditorSuggestions.jsx
```diff
@@ -71,6 +73,10 @@ export default class ExpressionEditorSuggestions extends Component {
     this.props.onSuggestionMouseDown && this.props.onSuggestionMouseDown(index);
   }
 
+  onMouseDownCapture = _.memoize(i => {
```
kamilmielnik, line 76: Nit: It's not a `onMouseDownCapture` handler, but a function returning these handlers, so I'd call it `createOnMouseDownCapture` or `getOnMuseDownCapture`.

kamilmielnik (approved):
With 6x CPU slowdown this PR definitely brings a noticeable performance improvement.

However, the changes in `ExpressionEditorSuggestions` do not seem to make any difference.

frontend/src/metabase/query_builder/components/expressions/ExpressionEditorSuggestions.jsx
uladzimirdev: I agree, it's not really necessary for performance, but overall code after refactoring looks a bit better to me

PR #37587, opened 2024-01-11
Remove internal query check

### LINKED ISSUE #37514: [MLv2] [FE] DashCardMenu - internal query check
https://github.com/metabase/metabase/blob/e11df0150b1c244d05c6dd9b60943290328cad9a/frontend/src/metabase/dashboard/components/DashCard/DashCardMenu/DashCardMenu.tsx#L153

Can we even get an internal query in dashboards now? If yes, add another `Question` method, e.g. `isInternal` or whatever similar to `isNative` or `isStructured`. But I'd try removing this check first, I believe it's irrelevant now

ranquild:
Closes https://github.com/metabase/metabase/issues/37514


uladzimirdev (commented):
Issues claims we need to add a method to Question. Should we still do it?

ranquild:
@uladzimirdev if we don't need the method then don't

kamilmielnik (approved):

PR #37588, opened 2024-01-11
docs - account for case when ordering endpoint links

jeff-bruemmer:
Small edit to account for case when sorting endpoint links.

### DIFF
--- a/src/metabase/cmd/endpoint_dox.clj
+++ b/src/metabase/cmd/endpoint_dox.clj
@@ -190,14 +190,16 @@
   [endpoints]
   (str/join "\n" (map (fn [[ep ep-data]] (build-endpoint-link ep ep-data)) endpoints)))
 
-(defn map-endpoints
+(defn- map-endpoints
   "Creates a sorted map of API endpoints. Currently includes some
   endpoints for paid features."
   []
   (->> (collect-endpoints)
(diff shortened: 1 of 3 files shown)


src/metabase/cmd/endpoint_dox.clj
```diff
@@ -197,7 +197,9 @@
   (->> (collect-endpoints)
        (map process-endpoint)
        (group-by :ns-name)
-       (into (sorted-map))))
+       (into (sorted-map-by (fn [a b] (compare
+                                       (str/lower-case a)
+                                       (str/lower-case b)))))))
```
camsaul, line 202: `str/lower-case` lower cases things based on the system locale, and if you're running this on a computer set to Turkish it tends to do really wacky stuff, it lower cases upper-case `I` to a non-English letter that tends to break stuff. So we have a Kondo linter rule telling people not to use `str/lower-case`, use our util function `u/lower-case-en` instead, which lowercases with English rules.

I doubt it's a big problem for generating API dox, but it's probably going to cause Kondo linter errors so you should just fix it here

```suggestion
                                       (u/lower-case-en a)
                                       (u/lower-case-en b)))))))
```

camsaul (approved):
LGTM

PR #37076, opened 2023-12-22
Add show metabase links toggle admin setting

WiNloSt:
> [!Note]
> I'll add self-annotation to help with reviewing this PR tomorrow (Fri, Jan 12)

Epic: https://github.com/metabase/metabase/issues/36966
Fixes #37376

### Description

Add an admin setting and hide metabase links according to [this spec](https://www.notion.so/metabase/Tech-Remove-Metabase-links-outside-of-admin-settings-f88ff0fa8f574d6393c6fe5a15911b0a?pvs=4#e597d19982c149b499f37a266c5d8306)

#### Run EE version
1. Observe every place [listed here](https://www.notion.so/metabase/Tech-Remove-Metabase-links-outside-of-admin-settings-f88ff0fa8f574d6393c6fe5a15911b0a?pvs=4#e597d19982c149b499f37a266c5d8306) that the doc links are still presented
2. Go to admin settings > appearance and toggle off **Show links and references**
3. Observe every place [listed here](https://www.notion.so/metabase/Tech-Remove-Metabase-links-outside-of-admin-settings-f88ff0fa8f574d6393c6fe5a15911b0a?pvs=4#e597d19982c149b499f37a266c5d8306) again that the doc links should be gone.

#### Run OSS version after you've toggled off **Show links and references** when running EE
1. Observe every place [listed here](https://www.notion.so/metabase/Tech-Remove-Metabase-links-outside-of-admin-settings-f88ff0fa8f574d6393c6fe5a15911b0a?pvs=4#e597d19982c149b499f37a266c5d8306) that the doc links are still presented
2. You shouldn't see the appearance settings anyway in admin settings.

### Demo 
Screenshot 2023-12-22 at 10 25 52 PM
Screenshot 2023-12-22 at 10 25 55 PM

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #37596, opened 2024-01-11
add eslint rule to avoid typos in paid feature test files

npretto:
## Description
We're adding more tests to paid features, which involves having 3 almost equal files (see [notion](https://www.notion.so/metabase/Testing-paid-features-49e03e1b2a3b41ffb4ebdc7616a93d90?pvs=4)).

The easiest way is writing the first file and then copy pasting, and it's easy to forget to change the descriptions of the tests (it happened recently in a pr).

Is this taking eslint too far or would you say it's okay?
We can always remove the rules without consequences if they end up creating more issues than they're solving.

## How to verify:
The ci should fail in this PR as there's one of those typos already on master

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -197,6 +197,34 @@
       "rules": {
         "jest/valid-title": ["error", { "ignoreTypeOfDescribeName": true }]
       }
-    }
+    },
+    {
+      "files": [
+        "premium.*.spec.ts"
+      ],
+      "rules": {
+        "no-restricted-syntax": [
+          "error",
+          {
+            "selector": "Literal[value=/EE without token/]",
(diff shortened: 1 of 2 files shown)


WiNloSt:
I think it's too early to say that we standardize the test names `EE with token` and `EE without token`. This is a pattern I found when I was looking for examples to create these test suites.

Most of the incorrect [names you mentioned](https://github.com/metabase/metabase/pull/37076#discussion_r1449233286) are from that PR.

I'd prefer we fix this manually first unless we agree this is what we want.

npretto:
Yeah I agree, I gave it a go because it was easy and quick but it could also give us a false confidence, gonna close it.

PR #37601, opened 2024-01-11
run frontend ci tasks when eslint config changes

npretto:
### Description

Currently we don't run the linting job when a pr only contains changes in `.eslintrc`.
To fix it I added the file name under `frontend_sources` even though it's not technically a source file.
Would it be better if we added it to `frontend_all`?

### How to verify

This pr https://github.com/metabase/metabase/pull/37596 contains only a `.eslintrc` change + the changes from this pr.

You can check under the [commits] tab(https://github.com/metabase/metabase/pull/37596/commits) that the job was skipped on the first commit, that didn't have the files-path change, and started (and correctly failed) on the second one)

image

### DIFF
--- a/.github/file-paths.yaml
+++ b/.github/file-paths.yaml
@@ -23,6 +23,7 @@ frontend_sources: &frontend_sources
   - "**/tsconfig*.json"
   - "package.json"
   - "babel.config.json"
+  - ".eslintrc"
   - "postcss.config.js"
   - "webpack.config.js"
   - "webpack.static-viz.config.js"


.github/file-paths.yaml
```diff
@@ -23,6 +23,7 @@ frontend_sources: &frontend_sources
   - "**/tsconfig*.json"
   - "package.json"
   - "babel.config.json"
+  - ".eslintrc"
```
npretto, line 26:
```suggestion
```

alxnddr (approved):

PR #37602, opened 2024-01-11
Trend charts — Static viz

### LINKED ISSUE #35556: Add new trend chart functionality to the static viz
- [ ] FE: create static trend chart component that reuse all the formatting logic from the dynamic visualization
     - ⚠️ The static viz is scheduled as the last milestone but while working on all previous ones, organize the display logic in a way it can be imported in the static viz environment, by not importing formatting, text measuring, files that rely on `window` object or use `setTimeout`, etc. In ECharts project we inject text measuring, static `formatValue` function, `fontFamily`, color getter.
     - To reduce the code duplication, we will render HTML by Javascript from the static viz bundle. It should return raw HTML with inline styles that use a supported by [CSSBox](https://github.com/radkovo/CSSBox) subset of CSS so that it can be converted by the BE to PNGs. 
  - [ ] BE: similarly to https://github.com/metabase/metabase/pull/34820 create isomorphic-html-chart rendering path that expects HTML, then embeds it in emails or renders as PNGs for Slack

kulyk:
Resovles #35556, based on #37418

- adds static viz support for trend charts
- adds static trend charts to Storybook (`Static-Viz/SmartScalar`)
- enables PNG exports for trend charts

### Demo

CleanShot 2024-01-12 at 19 07 44@2x


frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -1,16 +1,20 @@
...
 import { formatChange } from "metabase/visualizations/visualizations/SmartScalar/utils";
 import { isEmpty } from "metabase/lib/validate";
 import { isDate } from "metabase-lib/types/utils/isa";
 
-export function computeTrend(series, insights, settings) {
+export function computeTrend(
+  series,
+  insights,
+  settings,
+  { formatValue, color },
```
alxnddr, line 16: Really nit, but we use `getColor` everywhere when we inject the function. It makes it more obvious that it is not our imported `color` function and with verb prefix it looks like a bit more precise name

alxnddr (commented):
Looks great! I found a small inconsistency related to defaults but overall works as expected

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
kulyk: Fixed in [`441c424`](https://github.com/metabase/metabase/pull/37602/commits/441c4244deb93ce005de0981dc13bd607e201493)

frontend/src/metabase/static-viz/components/SmartScalar/settings.ts
```diff
@@ -0,0 +1,25 @@
+import type { RawSeries, VisualizationSettings } from "metabase-types/api";
+import { getCommonStaticVizSettings } from "metabase/static-viz/lib/settings";
+import type { ComputedVisualizationSettings } from "metabase/visualizations/types";
+import { getDefaultColumn } from "metabase/visualizations/lib/settings/utils";
+import { VIZ_SETTINGS_DEFAULTS } from "metabase/visualizations/visualizations/SmartScalar/constants";
+import { getDefaultComparison } from "metabase/visualizations/visualizations/SmartScalar/utils";
+
+export const computeSmartScalarSettings = (
```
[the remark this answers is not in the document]
kulyk: Nice catch! Should be fixed by [`ae770e7`](https://github.com/metabase/metabase/pull/37602/commits/ae770e73c6fbfc15f29e4c9afe02f366ef83cd43) — it didn't filter out non-numeric columns when selecting a default value for `scalar.field`

alxnddr (approved):
🥇

PR #37606, opened 2024-01-11
Handle unknown column-settings keys instead of passing a `nil` key.

### LINKED ISSUE #27941: Pivot table with the filter on a date column breaks dashboard sending
**Describe the bug**
Create a query, which summarize values by date and some other metric (e.g. sum of hours by developer, by date).
Use the `Pivot Table` visualization.
Dates go into columns, other metric into rows (not sure if it is important).
Add question to some dashboard and try to send it by email - should work.

Now edit the question and in column settings for the date field enable sorting.
Try to send the dashboard again - it will fail.
Check server logs - there will be a `NullPointerException` exception.

**Screenshots**
Question setup: 
Screenshot 2023-01-28 195442
Screenshot 2023-01-28 200432

**Logs**
[server.log](https://github.com/metabase/metabase/files/10527790/server.log)

You can get this information by going to Admin -> Troubleshooting.

- Your browser and the version: Chrome 09.0.5414.120 (Official Build) (64-bit)
- Your operating system: Windows 11
- Your databases: Postgres
- Metabase version: 0.45.2.1
- Metabase hosting environment: Elastic Beanstalk
- Metabase internal database: H2 (probably)

adam-james-v:
Fixes: #27941 

This PR: 
 - handles unknown column-settings keys by namespacing them with 'unknown'. 
 - adds 2 tests to ensure round-tripping of the viz-settings normalization still works.
 - adds a render/body test to check that table renders work even with these unknown column-settings keys.

The issue  called out pivot table static-viz failing when a sort was added to the column-viz.

This led to a null pointer exception in a formatter when the table render was attempted. Following the chain a bit, I noticed that there's a map of 'db' keys -> 'norm' keys defined in `metabase.shared.models.visualization-settings`

```clojure
(def ^:private db->norm-column-settings-keys
  {:column_title       ::column-title
   :date_style         ::date-style
   :date_separator     ::date-separator
   :date_abbreviate    ::date-abbreviate
   :time_enabled       ::time-enabled
   :time_style         ::time-style
   :number_style       ::number-style
   :currency           ::currency
   :currency_style     ::currency-style
   :currency_in_header ::currency-in-header
   :number_separators  ::number-separators
   :decimals           ::decimals
   :scale              ::scale
   :prefix             ::prefix
   :suffix             ::suffix
   :view_as            ::view-as
   :link_text          ::link-text
   :link_url           ::link-url
   :show_mini_bar      ::show-mini-bar})
```

Which doesn't contain the key `:pivot_table.column_sort_order`, which is the key added by the frontend when you sort the column via the viz settings.

This would cause the transform to leave a `nil` key in its place which ultimately led to the NPE further down the line.

I've made the change that any unknown key gets namespaced with `unknown`. So, `:pivot_table.column_sort_order` -> `:unknown/pivot_table.column_sort_order`.

This prevents the NPE.

Below I've pasted my namespace exploring this issue:

```clojure
(ns tickets.27941
  "## Pivot table with the filter on a date column breaks dashboard sending

  ### URL: https://github.com/metabase/metabase/issues/27941

  See the comment block below for details."
  (:require
   [clojure.string :as str]
   [clojure.test :refer :all]
   [metabase.email :as email]
   [metabase.pulse]
   [metabase.pulse.render.test-util :as render.tu]
   [metabase.test :as mt]
   [toucan2.core :as t2]))

;; Manual repro following steps in the issue: can repro!
;; See the null pointer exception in the repl.
;; Stacktrace pointes to render body table impl.

;; Let's grab the card's settings and have a look.

(-> (t2/select-one :model/Card 952)
    (select-keys [:display :dataset_query :visualization_settings]))

{:display :pivot,
 :dataset_query
 {:database 1,
  :type :query,
  :query
  {:source-table 8,
   :aggregation [[:sum [:field 66 {:base-type :type/Integer}]]],
   :breakout [[:field 69 {:base-type :type/DateTime, :temporal-unit :week}] [:field 70 {:base-type :type/Text}]],
   :filter [:between [:field 71 {:base-type :type/Integer}] 0 10]}},
 :visualization_settings
 {:pivot_table.column_split
  {:rows [[:field 70 {:base-type :type/Text}]],
   :columns [[:field 69 {:base-type :type/DateTime, :temporal-unit :week}]],
   :values [[:aggregation 0]]},
  :column_settings
  {"[\"ref\",[\"field\",69,{\"base-type\":\"type/DateTime\"}]]" {:pivot_table.column_sort_order "ascending"}}}}

;; The critical area is :column_settings in the :viz-settings, I think.

;; Here's the same question before any sort is applied:

(-> (t2/select-one :model/Card 953)
    (select-keys [:display :dataset_query :visualization_settings]))

{:display :pivot,
 :dataset_query
 {:database 1,
  :type :query,
  :query
  {:source-table 8,
   :aggregation [[:sum [:field 66 {:base-type :type/Integer}]]],
   :breakout [[:field 69 {:base-type :type/DateTime, :temporal-unit :week}] [:field 70 {:base-type :type/Text}]],
   :filter [:between [:field 71 {:base-type :type/Integer}] 0 10]}},
 :visualization_settings
 {:pivot_table.column_split
  {:rows [[:field 70 {:base-type :type/Text}]],
   :columns [[:field 69 {:base-type :type/DateTime, :temporal-unit :week}]],
   :values [[:aggregation 0]]}}}

;; Now lets build a fn that we can use to run the query in tests.

(defn- run-pulse-and-return-attached-image-data
  "Simulate sending the pulse email, get the inline pngs."
  [pulse]
  (mt/with-fake-inbox
      (mt/with-test-user nil
    (->>
       (get-in @mt/inbox ["rasta@metabase.com" 0 :body])
       #_(keep
        (fn [{:keys [type content-type content]}]
          (when (and
                 (= :inline type)
                 (= "image/png" content-type))
            content))))))

(defn breaks
  []
  (mt/dataset test-data
    (let [q   {:database (mt/id)
               :type     :query
               :query
               {:source-table (mt/id :reviews)
                :aggregation  [[:sum [:field (mt/id :reviews :rating) {:base-type :type/Integer}]]],
                :breakout     [[:field (mt/id :reviews :created_at) {:base-type :type/DateTime, :temporal-unit :week}]
                               [:field (mt/id :reviews :reviewer) {:base-type :type/Text}]],
                :filter       [:between [:field (mt/id :reviews :product_id) {:base-type :type/Integer}] 0 10]}}
          viz {:pivot_table.column_split
               {:rows    [[:field (mt/id :reviews :reviewer) {:base-type :type/Text}]],
                :columns [[:field (mt/id :reviews :created_at) {:base-type :type/DateTime, :temporal-unit :week}]],
                :values  [[:aggregation 0]]}
               :column_settings
               {(format "[\"ref\",[\"field\",%s,{\"base-type\":\"type/DateTime\"}]]" (mt/id :reviews :created_at))
                {:pivot_table.column_sort_order "ascending"}}}]
      (mt/dataset test-data
        (mt/with-temp [Card                 {card-id :id} {:display                :pivot
                                                           :dataset_query          q
                                                           :visualization_settings viz}
                       Dashboard {dash-id :id} {:name "just dash"}
                       DashboardCard {dash-card-id :id} {:dashboard_id dash-id
                                                         :card_id      card-id}
                       Pulse                {pulse-id :id :as p} {:name         "Test Pulse"
                                                                  :dashboard_id dash-id}
                       PulseCard             _             {:pulse_id          pulse-id
                                                            :card_id           card-id
                                                            :dashboard_card_id dash-card-id}
                                                            :pulse_id     pulse-id
                                                            :enabled      true}
          (-> (run-pulse-and-return-attached-image-data p)
              first))))))

;; The NPE is at `metabase.query-processor.streaming.common/normalize-keys$fn` (common.clj:121)


;; The `normalize-keys` fn:

(defn normalize-keys
  "Update map keys to remove namespaces from keywords and convert from snake to kebab case."
  [m]
  (update-keys m (fn [k] (-> k name (str/replace #"_" "-") keyword))))

;; Is seeing a map like this:

(def bad-map
  {:metabase.shared.models.visualization-settings/time-style "h:mm A"
   nil "ascending"})

(def good-map
  {:metabase.shared.models.visualization-settings/time-style "h:mm A"})


;; That 'nil' key is no good.
#_(normalize-keys bad-map) ;; fails
(normalize-keys good-map) ;; {:time-style "h:mm A"}


;; Ok, so that's the NPE location, but why does that map happen in the first place?
;; Let's explore from the other side by looking at how a :pivot display is meant to be rendered.

;; :display is :pivot but in our static viz code we don't have :pivot.
;; `metabase.pulse.render/detect-pulse-chart-type` returns :table for our :display :pivot cards.

;; So we can go to the render/body impl. for :table.
;; Nothing off there, but we do end up using formatters, in this case the datetime formatter.
;; That's in `metabase.formatter.datetime`, where we see viz-settings-for-col
;; viz-settings-for-col has viz-settings come in:

{:pivot_table.column_split
 {:rows [[:field 1808 {:base-type :type/Text}]],
  :columns [[:field 1804 {:base-type :type/DateTime, :temporal-unit :week}]],
  :values [[:aggregation 0]]}
 :metabase.shared.models.visualization-settings/column-settings
 {#:metabase.shared.models.visualization-settings{:field-id 1804, :field-metadata {"base-type" "type/DateTime"}} {nil "ascending"}},
 :metabase.shared.models.visualization-settings/global-column-settings #:type{:Temporal #:metabase.shared.models.visualization-settings{:time-style "h:mm A"}}}

;; So something's happening in the normalization that happens before this stage.

;; `metabase.shared.models.visualization-settings/db->norm-column-ref` Has this input value:

["ref" ["field" 1804 {"base-type" "type/DateTime"}]]

;; That actually works fine.
;; Ok, so the function `db->norm-column-settings` is responsible for normalizing the column settings.
;; In it there's a reduce-kv that uses `parse-db-column-ref` for the keys, which are field refs).
;; That's working fine.

;; Ther's also `db->norm-column-settings-entries` which is meant to handle the transform of the settings for that particular column key. So, the input to this fn is a map of stuff like `{:pivot.column_sort_order "ascending" ...}`

;; THIS is where things fail! The key normalization is just a map, and there's no :pivot.column_sort_order at all, so we get `nil`.

;; I think the correct fix here is to adjust this fn to return unknown keys either unchanged or namespaced as :unknown/the-key.

;; I should check too that round-tripping still works.
;; So, let's start with a viz-settings map:

(def viz-b
  {:pivot_table.column_split
   {:rows    [[:field 70 {:base-type :type/Text}]],
    :columns [[:field 69 {:base-type :type/DateTime, :temporal-unit :week}]],
    :values  [[:aggregation 0]]},
   :table.cell_column "sum",
   :column_settings
   {"[\"ref\",[\"field\",69,{\"base-type\":\"type/DateTime\"}]]"
    {:pivot_table.column_sort_order "ascending"}}})

(deftest viz-norm-round-trip-test
  (testing "The viz-settings db->norm and norm->db fns round trip properly with unknown column settings keys."
    (let [start {:column_settings
                 {"[\"ref\",[\"field\",69,{\"base-type\":\"type/DateTime\"}]]"
                  {:pivot_table.column_sort_order "ascending"}}}]
      (is (= start (-> start mb.viz/db->norm mb.viz/norm->db))))))

(deftest unknown-column-settings-test
  (testing "Unknown `:column_settings` keys don't break static-viz rendering with a Null Pointer Exception (#27941)."
    (mt/dataset test-data
      (let [q   {:database (mt/id)
                 :type     :query
                 :query
                 {:source-table (mt/id :reviews)
                  :aggregation  [[:sum [:field (mt/id :reviews :rating) {:base-type :type/Integer}]]],
                  :breakout     [[:field (mt/id :reviews :created_at) {:base-type :type/DateTime, :temporal-unit :week}]
                                 [:field (mt/id :reviews :reviewer) {:base-type :type/Text}]],
                  :filter       [:between [:field (mt/id :reviews :product_id) {:base-type :type/Integer}] 0 10]}}
            viz {:pivot_table.column_split
                 {:rows    [[:field (mt/id :reviews :reviewer) {:base-type :type/Text}]],
                  :columns [[:field (mt/id :reviews :created_at) {:base-type :type/DateTime, :temporal-unit :week}]],
                  :values  [[:aggregation 0]]}
                 :column_settings
                 {(format "[\"ref\",[\"field\",%s,{\"base-type\":\"type/DateTime\"}]]" (mt/id :reviews :created_at))
                  {:pivot_table.column_sort_order "ascending"}}}]
        (mt/dataset test-data
          (mt/with-temp [Card                 {card-id :id} {:display                :pivot
                                                             :dataset_query          q
                                                             :visualization_settings viz}]
            (testing "the render succeeds with unknown column settings keys"
              (is (seq (render.tu/render-card-as-hickory card-id))))))))))
```

### DIFF
--- a/src/metabase/query_processor/streaming/common.clj
+++ b/src/metabase/query_processor/streaming/common.clj
@@ -118,7 +118,7 @@
 (defn normalize-keys
   "Update map keys to remove namespaces from keywords and convert from snake to kebab case."
   [m]
-  (update-keys m (fn [k] (-> k name (str/replace #"_" "-") keyword))))
+  (update-keys m (fn [k] (some-> k name (str/replace #"_" "-") keyword))))
 
 (def col-type
   "The dispatch function logic for format format-timestring.
--- a/src/metabase/shared/models/visualization_settings.cljc
+++ b/src/metabase/shared/models/visualization_settings.cljc
@@ -553,7 +553,8 @@
     :time_style
     (assoc m ::time-style (db->norm-time-style v))
 
-    (assoc m (db->norm-column-settings-keys k) v)))
+    (assoc m (or (db->norm-column-settings-keys k)
+                 (keyword "unknown" (name k))) v)))
 
 (defn db->norm-column-settings-entries
   "Converts the DB form of a map of :column_settings entries to its normalized form."
@@ -626,7 +627,8 @@
   [m k v]
   (case k
     ::click-behavior (assoc m :click_behavior (norm->db-click-behavior v))
-    (assoc m (norm->db-column-settings-keys k) v)))
+    (assoc m (or (norm->db-column-settings-keys k)
+                 (keyword (name k))) v)))
 
 (defn norm->db-column-ref
   "Creates the DB form of a column ref (i.e. the key in the column settings map) for the given normalized args. Either
--- a/src/metabase/shared/models/visualization_settings.cljc
+++ b/src/metabase/shared/models/visualization_settings.cljc
@@ -553,8 +553,7 @@
     :time_style
     (assoc m ::time-style (db->norm-time-style v))
 
-    (assoc m (or (db->norm-column-settings-keys k)
-                 (keyword "unknown" (name k))) v)))
+    (assoc m (db->norm-column-settings-keys k) v)))
 
 (defn db->norm-column-settings-entries
   "Converts the DB form of a map of :column_settings entries to its normalized form."
@@ -627,8 +626,7 @@
   [m k v]
   (case k
     ::click-behavior (assoc m :click_behavior (norm->db-click-behavior v))
-    (assoc m (or (norm->db-column-settings-keys k)
-                 (keyword (name k))) v)))
+    (assoc m (norm->db-column-settings-keys k) v)))
 
 (defn norm->db-column-ref
   "Creates the DB form of a column ref (i.e. the key in the column settings map) for the given normalized args. Either
--- a/test/metabase/pulse/render/body_test.clj
+++ b/test/metabase/pulse/render/body_test.clj
@@ -799,3 +799,28 @@
               {:keys [content]} (body/render :line :inline "UTC" card nil (:data query-results))]
           (testing "Content is generated (rather than an exception being thrown)"
             (is (= :div (first content)))))))))
+
+(deftest unknown-column-settings-test
+  (testing "Unknown `:column_settings` keys don't break static-viz rendering with a Null Pointer Exception (#27941)."
+    (mt/dataset test-data
+      (let [q   {:database (mt/id)
+                 :type     :query
+                 :query
+                 {:source-table (mt/id :reviews)
+                  :aggregation  [[:sum [:field (mt/id :reviews :rating) {:base-type :type/Integer}]]],
+                  :breakout     [[:field (mt/id :reviews :created_at) {:base-type :type/DateTime, :temporal-unit :week}]
+                                 [:field (mt/id :reviews :reviewer) {:base-type :type/Text}]],
+                  :filter       [:between [:field (mt/id :reviews :product_id) {:base-type :type/Integer}] 0 10]}}
+            viz {:pivot_table.column_split
+                 {:rows    [[:field (mt/id :reviews :reviewer) {:base-type :type/Text}]],
+                  :columns [[:field (mt/id :reviews :created_at) {:base-type :type/DateTime, :temporal-unit :week}]],
+                  :values  [[:aggregation 0]]}
+                 :column_settings
+                 {(format "[\"ref\",[\"field\",%s,{\"base-type\":\"type/DateTime\"}]]" (mt/id :reviews :created_at))
+                  {:pivot_table.column_sort_order "ascending"}}}]
+        (mt/dataset test-data
+          (mt/with-temp [Card                 {card-id :id} {:display                :pivot
+                                                             :dataset_query          q
+                                                             :visualization_settings viz}]
+            (testing "the render succeeds with unknown column settings keys"
+              (is (seq (render.tu/render-card-as-hickory card-id))))))))))
--- a/test/metabase/shared/models/visualization_settings_test.cljc
+++ b/test/metabase/shared/models/visualization_settings_test.cljc
@@ -215,20 +215,3 @@
       ;; make sure all keywords have the right namespace in normalized form
       (t/is (empty? (filter #(some? (namespace %)) (all-keywords to-db))))
       (t/is (= db-form to-db)))))
-
-(t/deftest unknown-column-settings-keys-test
-  (t/testing "Unknown column-settings keys are namespaced with `:unknown`."
-    (let [db-form {:column_settings
-                   {"[\"ref\",[\"field\",69,{\"base-type\":\"type/DateTime\"}]]"
-                    {:pivot_table.column_sort_order "ascending"
-                     :some_weird_key                "yea, it's weird."}}}
-          norm-form (mb.viz/db->norm db-form)]
-      (t/is (= #{"metabase.shared.models.visualization-settings" "unknown"}
-               (set (map namespace (all-keywords norm-form))))))))
-
-(t/deftest unknown-column-settings-keys-roundtrip-test
-  (t/testing "The viz-settings db->norm and norm->db fns roundtrip properly with unknown column settings keys."
-    (let [db-form {:column_settings
-                   {"[\"ref\",[\"field\",69,{\"base-type\":\"type/DateTime\"}]]"
-                    {:pivot_table.column_sort_order "ascending"}}}]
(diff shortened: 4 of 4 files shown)


markbastian:
Love the explanation ns. Makes it so much easier to follow!

I'm curious, what happens if you change normalize-keys from:
```clojure
(defn normalize-keys
  "Update map keys to remove namespaces from keywords and convert from snake to kebab case."
  [m]
  (update-keys m (fn [k] (-> k name (str/replace #"_" "-") keyword))))
```
to
```clojure
(defn normalize-keys
  "Update map keys to remove namespaces from keywords and convert from snake to kebab case."
  [m]
  (update-keys m (fn [k] (some-> k name (str/replace #"_" "-") keyword))))
```

If that works I think it would an even smaller change that accomplishes the same thing.

markbastian (approved):
Love the explanation and repro ns! 🚢 🇮🇹

PR #37608, opened 2024-01-11
Add personal-only to collection endpoint

qwef:
### Description

Adds `personal-only` to GET `/api/collection` endpoint.


src/metabase/api/collection.clj
```diff
@@ -103,27 +103,36 @@
   `?archived=true`.
 
   By default, admin users will see all collections. To hide other user's collections pass in
-  `?exclude-other-user-collections=true`."
-  [archived exclude-other-user-collections namespace]
+  `?exclude-other-user-collections=true`.
+
+  If personal-only is true, then return only peronsal collections."
```
camsaul, line 108: This stuff gets markdown formatted in the API docs and elsewhere

```suggestion
  If `personal-only` is `true`, then return only personal collections."
```

src/metabase/api/collection.clj
```diff
@@ -103,27 +103,36 @@
...
-    ;; remove the :metabase.models.collection.root/is-root? tag since FE doesn't need it
-    ;; and for personal collections we translate the name to user's locale
-    (for [collection collections]
-      (-> collection
-          (dissoc ::collection.root/is-root?)
-          collection/personal-collection-with-ui-details))))
+   namespace                      [:maybe ms/NonBlankString]
+   personal-only                  [:maybe ms/BooleanValue]}
+  (if personal-only
+    (t2/select :model/Collection {:where [:and
```
camsaul, line 115: So, if you set `personal-only=true`, then `archived` and the other stuff get ignored? That seems not obvious... wouldn't it be better just to update `select-collections` to take an extra `personal-only` parameter? Then you don't need to introduce a whole separate branch here

src/metabase/api/collection.clj
```diff
@@ -103,27 +103,36 @@
...
+                                          [:!= :personal_owner_id nil]
+                                          (collection/visible-collection-ids->honeysql-filter-clause
+                                           :id
+                                           (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]})
+    (as->
+     (select-collections archived exclude-other-user-collections namespace false nil) collections
```
camsaul, line 121: `select-collections` has a lot of non-obvious positional parameters, I think at this point if you add another parameter we should refactor it to take keyword parameters. Five positional parameters is enough

test/metabase/api/collection_test.clj
```diff
@@ -127,6 +127,32 @@
...
+        (perms/grant-collection-read-permissions! (perms-group/all-users) collection-1)
+        (is (= (->> (t2/select :model/Collection {:where [:!= :personal_owner_id nil]})
+                    (map :name)
+                    (into #{}))
+               (->> (mt/user-http-request :crowberto :get 200 "collection" :personal-only true)
+                    (filter (fn [{collection-name :name}]
+                              (or (#{"Our analytics" "Collection 1" "Collection 2"} collection-name)
+                                  (str/includes? collection-name "Personal Collection"))))
+                    (map :name)
+                    (into #{}))))))))
```
camsaul, line 154: I think you should just split these into two separate tests. Any time you are thinking about putting a newline in a test, ask yourself if it wouldn't be better written as a separate `deftest` form... since these two don't share any state at all I think the answer here is yes.

We have been bad about writing really long tests in the past, but moving forward we're making more and more test parallelizable, so if you make this two separate tests we'll be able to run them both at the same time at some point in the near future

src/metabase/api/collection.clj
camsaul: Also if it actually was intentional that setting this parameter means all other parameters get ignored, you should definitely call that out in the endpoint docstring -- it's not obvious

camsaul (changes requested):
I think it would be a lot simpler if you just refactored `select-collections` to take an additional `personal-only` parameter. It's not sustainable or readable to introduce a new branch every time we have a new parameter... also setting `personal-only` here means other parameters get ignored which is not really obvious behavior.

qwef:
> I think it would be a lot simpler if you just refactored `select-collections` to take an additional `personal-only` parameter. It's not sustainable or readable to introduce a new branch every time we have a new parameter... also setting `personal-only` here means other parameters get ignored which is not really obvious behavior.

This all makes sense, thanks for the review!

test/metabase/api/collection_test.clj
qwef: Yea that makes a lot of sense.

src/metabase/api/collection.clj
```diff
@@ -75,25 +75,35 @@
...
-                       (collection/visible-collection-ids->honeysql-filter-clause
-                        :id
-                        (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]
-               ;; Order NULL collection types first so that audit collections are last
-               :order-by [[[[:case [:= :type nil] 0 :else 1]] :asc]
-                          [:%lower.name :asc]]})
-    exclude-other-user-collections (remove-other-users-personal-subcollections api/*current-user-id*)))
+  For `archived`, we can either include all values (when `archived` is `nil`),
+  or to include when `archived` is `true` or `false`.
+
```
escherize, line 80: I think the meaning here changed. (It's funny how looking at the code is so much easier to understand sometimes)
```suggestion
For archived, we can either include everthing (when archived is `nil`), only archived (when `archived` is true), or only non-archived (when `archived` is false).
```

src/metabase/api/collection.clj
```diff
@@ -75,25 +75,35 @@
...
+  To select only personal collections, pass in `personal-only` as `true`. This is disregard other parameters."
```
escherize, line 81: Could you explain exactly what it will get?

src/metabase/api/collection.clj
```diff
@@ -103,15 +113,22 @@
   `?archived=true`.
 
   By default, admin users will see all collections. To hide other user's collections pass in
-  `?exclude-other-user-collections=true`."
-  [archived exclude-other-user-collections namespace]
+  `?exclude-other-user-collections=true`.
+
+  If personal-only is true, then return only peronsal collections."
```
escherize, line 118:
```suggestion
  If personal-only is `true`, then return only personal collections, and all other query-params are ignored."
```

src/metabase/api/collection.clj
```diff
@@ -75,25 +75,35 @@
...
+
+  To select only personal collections, pass in `personal-only` as `true`. This is disregard other parameters."
+  [{:keys [archived exclude-other-user-collections namespace shallow collection-id personal-only]}]
+  (if personal-only
+    (t2/select :model/Collection
+               {:where [:and
+                        [:!= :personal_owner_id nil]
+                        (collection/visible-collection-ids->honeysql-filter-clause
+                         :id
+                         (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]})
```
escherize, line 89: I won't explicitly ask you to change it, but we've been trying to pass dynamic vars into functions more, in other words that would mean passing in the permissions set as an arg to `select-collections`. 

I don't see why we should ignore the other args when `personal-only`, but there's probably a good reason.

escherize (changes requested):
You mentioned this was ready for review, but I think there are some issues to fix up still, e.g. some docstrings should be reworded. Up to you if you decide to pass `api/*current-user-perm-set*`

src/metabase/api/collection.clj
qwef: Currently they don't make sense in terms of business logic (i.e. you can't archive a personal collection), so I think this made more sense than confusing API callers.

qwef: Thinking about it a bit more I think it makes the logic much cleaner that maybe we allow it even if some of the combinations make no sense. I'll change it

escherize (approved):
LGTM

src/metabase/api/collection.clj
```diff
@@ -75,25 +75,31 @@
...
                          (location-from-collection-id-clause collection-id))
+                       (when personal-only
+                         [:!= :personal_owner_id nil])
                        (when exclude-other-user-collections
                          [:or [:= :personal_owner_id nil] [:= :personal_owner_id api/*current-user-id*]])
                        (perms/audit-namespace-clause :namespace namespace)
                        (collection/visible-collection-ids->honeysql-filter-clause
                         :id
-                        (collection/permissions-set->visible-collection-ids @api/*current-user-permissions-set*))]
+                        (collection/permissions-set->visible-collection-ids permissions-set))]
```
camsaul, line 98: I'm worried that someone might not realize that you're expected to pass in `permissions-set` here, and forget to do it, and have something bad happen... I think having a default behavior keeps someone from doing something dangerous and also makes this a little nicer to use from the REPL and some tests. Then you don't need to make some of the changes to existing usages

```suggestion
                        (collection/permissions-set->visible-collection-ids (or permissions-set @api/*current-user-permissions-set*)))]
```

camsaul (approved):
Looks good now

iethree:
this makes me soooo happy

PR #37438, opened 2024-01-09
Migrate click behavior column computation to MLv2

ranquild:
Closes https://github.com/metabase/metabase/issues/37342

Blocked by https://github.com/metabase/metabase/issues/37501 and depends on https://github.com/metabase/metabase/pull/37648

PR #37609, opened 2024-01-11
[MLv2] Add `compatible-type?` wrapper that handles JS `DatasetColumn`s

### LINKED ISSUE #37501: [MLv2] [BE] Add a way to compare DatasetColumn type with ColumnMetadata type
**Context**

In the initiative of migrating click behavior to MLv2, we stumbled upon filtering columns, which used to be filtered and compared by `base_type`. With MLv2 we don't have access to `base_type`, but we still need to compare types of the columns and make sure they are related or not (using `isa` method).

Unfortunately we need to compare old (DatasetColulm, json like) columns with new one (originally from clojure).

@ranquild suggested a helper to compare `base_type` of ColumnMetadata columns

```
(defn ^:export compatible-type?
  "Do columns have compatible base types?"
  [column1 column2]
  (or
    (and (string? column1) (string? column2))
    (and (number? column1) (number? column2))
    (and (temporal? column1) (number? column2))
    (clojure.core/isa? (:base-type column1) (:base-type column2))))
```

but we need to extend it with `DatasetColumns`, so we would like to compare DatasetColumn with ColumnMetadata like it's done in DrillThrus.

Related FE change https://github.com/metabase/metabase/pull/37438

bshepherdson:
This wraps the logic of `metabase.lib.types.isa/compatible-type?` with
code to convert JS `DatasetColumn` objects into CLJS form.

Fixes #37501. (Unblocks #37438.)

### DIFF
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -24,6 +24,7 @@
    [metabase.lib.metadata.protocols :as lib.metadata.protocols]
    [metabase.lib.order-by :as lib.order-by]
    [metabase.lib.stage :as lib.stage]
+   [metabase.lib.types.isa :as lib.types.isa]
    [metabase.lib.util :as lib.util]
    [metabase.mbql.js :as mbql.js]
    [metabase.mbql.normalize :as mbql.normalize]
@@ -1178,3 +1179,11 @@
   "Add or update a filter against `temporal-column`. Modify the temporal unit for any breakouts."
   [a-query temporal-column stage-number start end]
   (lib.core/update-temporal-filter a-query temporal-column stage-number start end))
+
+(defn ^:export assignable?
+  "Given two CLJS `:metadata/columns` returns true if the first column's type `[[lib.types.isa/isa?]]` subtype of the
+  second column's type.
+
+  That is, that a value from `src-column` is assignable to `dst-column`."
+  [src-column dst-column]
+  (lib.types.isa/assignable? src-column dst-column))
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -1190,10 +1190,7 @@
   [a-query temporal-column stage-number start end]
   (lib.core/update-temporal-filter a-query temporal-column stage-number start end))
 
-(defn ^:export assignable?
-  "Given two CLJS `:metadata/columns` returns true if the first column's type `[[lib.types.isa/isa?]]` subtype of the
-  second column's type.
-
-  That is, that a value from `src-column` is assignable to `dst-column`."
+(defn ^:export valid-filter-for?
+  "Given two CLJS `:metadata/columns` returns true if `src-column` is a valid source to use for filtering `dst-column`."
   [src-column dst-column]
-  (lib.types.isa/assignable? src-column dst-column))
+  (lib.types.isa/valid-filter-for? src-column dst-column))
--- a/src/metabase/lib/types/isa.cljc
+++ b/src/metabase/lib/types/isa.cljc
@@ -308,9 +308,8 @@
     (or (clojure.core/isa? column-type :type/Text)
         (clojure.core/isa? column-type :type/TextLike))))
 
-(defn ^:export assignable?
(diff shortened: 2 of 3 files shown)


bshepherdson:
* `master`
  * **PR #37609** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37609?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37609?utm_source=stack-comment).

camsaul (approved):
I'm wondering if it wouldn't just be easier to expose a `legacy-metadata->mlv2-metadata` function since this is like the 5th place I've seen a need for it this past week. We already have code that does that in `metabase.lib.card` and probably other places as well, maybe just generalizing that and exposing it is the way to go. 

Maybe we call it something like `->mlv2-metadata` and have it accept either MLv2 metadata, in which case it's just `identity`, or legacy metadata, in which case it does the conversion. Idk 

Anyways this looks good for now, but if we need to do this stuff again it's probably time to write that helper

src/metabase/lib/types/isa.cljc
```diff
@@ -307,3 +307,16 @@
   (let [column-type (or effective-type base-type)]
     (or (clojure.core/isa? column-type :type/Text)
         (clojure.core/isa? column-type :type/TextLike))))
+
+(defn ^:export compatible-type?
+  "Do columns have compatible base types?
+
+  \"Compatible\" here means that they belong to the same family (strings, numbers, temporal) or `column1`'s type `isa?`
+  subtype of `column2`'s type."
```
metamben, line 315: > `column1`'s type `isa?` subtype of `column2`'s type
is asymmetrical, although `compatible-types?` sounds like a symmetrical predicate.

uladzimirdev:
https://github.com/metabase/metabase/pull/37648 adds a mechanism to get a column based on DatasetColumn, so we can compare exactly columns without any fallback to DatasetColumn

src/metabase/lib/js.cljs
```diff
@@ -1178,3 +1179,11 @@
   "Add or update a filter against `temporal-column`. Modify the temporal unit for any breakouts."
   [a-query temporal-column stage-number start end]
   (lib.core/update-temporal-filter a-query temporal-column stage-number start end))
+
+(defn ^:export assignable?
```
metamben, line 1183: I'm afraid this can lead to confusion given that we already have `metabase.types/assignable?`.

src/metabase/lib/types/isa.cljc
```diff
@@ -307,3 +307,18 @@
...
+  "Given two CLJS `:metadata/columns` returns true if the first column's type `[[lib.types.isa/isa?]]` subtype of the
+  second column's type. That is, that a value from `src-column` is assignable to `dst-column`.
+
+  That's the case if both are from the same family (strings, numbers, temporal) or if the `src-column` [[isa?]] subtype
+  of `dst-column`."
+  [src-column dst-column]
+  (or
+    (and (string? src-column)   (string? dst-column))
+    (and (number? src-column)   (number? dst-column))
+    (and (temporal? src-column) (temporal? dst-column))
```
metamben, line 321: I find this a very strange notion of assignability. Assigning a time to a date (or vice versa) seems very counterintuitive.

metamben (commented):
I don't want to block this because I guess this is so strange because of obscure historical reasons, but I would prefer more specific names and/or comments explaining why these seemingly wrong things are in fact OK.

uladzimirdev (approved):

src/metabase/lib/types/isa.cljc
bshepherdson: I've adjusted the name to be more specific `valid-filter-for?`.

src/metabase/lib/js.cljs
bshepherdson: I changed the name.

metamben (approved):

PR #37616, opened 2024-01-11
Register keywords defined by `mr/def` for LSP

camsaul:
In Emacs `M-x lsp-find-declaration` which is the totally convenient keybinding `<super>-l g g` for me jumps to the malli schema definition after adding this hook


metamben (approved):
Awesome! 🎉

PR #37617, opened 2024-01-11
Add snowplow events for embedding setup flow

oisincoveney:
Epic #35961 - Add snowplow events

### Description

Takes the events mentioned in the [Setup Flow doc](https://www.notion.so/metabase/Setup-flow-for-static-public-embedding-embedding-feature-discoverability-3f7362e71e63440883bfebae48ab54a0?pvs=4#e834f5744b1c421c8aecdd839961ba87). The events included are:

- `static_embed_discarded`
_Discarding changes on after changing parameters in static embed modal_
- `static_embed_published`
_Publishing a static embed modal_
- `static_embed_unpublished`
_Unpublishing a static embed modal_
- `static_embed_code_copied`
_Fired when copying the static embed code_
- `public_link_copied`
_Fired when copying the public link from the question/dashboard embed menu popover_
- `public_embed_code_copied`
_Fired when copying the iframe code in the embed modal_
- `public_link_removed`
_Fired when removing the public link from the question/dashboard embed menu popover_

### How to verify

All tests should pass. To manually test the snowplow events, refer to the list above and the actions that trigger them.

### Checklist


oisincoveney:
* `master`
  * **PR #37617** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37617?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37617?utm_source=stack-comment).

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -254,14 +259,166 @@ describe("embed modal display", () => {
...
+
+          cy.findByTestId("embed-modal-content-status-bar").within(() => {
+            cy.findByText("Publish changes").click();
+          });
+
+          expectGoodSnowplowEvent({
+            event: "static_embed_published",
+            artifact: resource,
+            new_embed: true,
+            // params: {
```
oisincoveney, line 384: I'm having issues getting the `expectGoodSnowplowEvent` function to match with the nested properties. Anyone have any ideas?

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -284,11 +441,24 @@ function createPublicResourceLink(resource, id) {
...
   }
 }
+
+function enableEmbeddingForResource({ resource, id, isDirty = false }) {
+  const endpoint = resource === "question" ? "card" : "dashboard";
+  cy.request("PUT", `/api/${endpoint}/${id}`, {
+    enable_embedding: true,
+    embedding_params: isDirty
+      ? {
+          test: "locked",
```
oisincoveney, line 459: We don't need a valid parameter here, since we check whether to display `Publish`, `Discard`, and `Unpublish` based on the difference between the `embedding_params` that the Question/Dashboard already has, and the new `embedding_params` that we're generating in the static embed modal

frontend/src/metabase/components/CopyButton/CopyButton.jsx
```diff
@@ -16,6 +16,7 @@ export default class CopyWidget extends Component {
   onCopy = () => {
     this.setState({ copied: true });
     setTimeout(() => this.setState({ copied: false }), 2000);
+    this.props.onCopy();
```
oisincoveney, line 19: Added a new prop of `onCopy` so we can pass in a function (generally to send a snowplow event)

frontend/src/metabase/dashboard/actions/sharing.ts
```diff
@@ -29,7 +29,11 @@ export const UPDATE_ENABLE_EMBEDDING =
 export const updateEnableEmbedding = createAction(
   UPDATE_ENABLE_EMBEDDING,
   ({ id }: DashboardIdPayload, enable_embedding: boolean) =>
-    DashboardApi.update({ id, enable_embedding }),
+    DashboardApi.update({
```
oisincoveney, line 32: I'm sure there's a way to do this on the BE so we can make sure the DB type and the timezone-aware string work normally. We could probably even send the `static_embed_published` and `static_embed_unpublished`  snowplow event from the BE but I'll need some guidance on that

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -20,6 +25,15 @@ import type {
 import { SelectEmbedTypePane } from "../SelectEmbedTypePane";
 import { EmbedModalContentStatusBar } from "./EmbedModalContentStatusBar";
 
+const countEmbeddingParameterOptions = (embeddingParams: EmbeddingParameters) =>
```
oisincoveney, line 28: We want to count the number of parameters for each embed state for the snowplow event i.e. disabled has 1, locked has 0, etc

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/AppearanceSettings.tsx
```diff
@@ -232,6 +232,7 @@ export const AppearanceSettings = ({
               secretKey={secretKey}
               params={params}
               displayOptions={displayOptions}
+              embedModalLocation={EMBED_MODAL_TABS.Appearance}
```
oisincoveney, line 235: I needed to add the location since each tab has its own instance of the same `CodeSample` component with the static embedding code, so the snowplow event can track that

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/CodeSample.jsx
```diff
@@ -37,7 +37,7 @@ export default class CodeSample extends Component {
     const selected = _.findWhere(options, { name });
     const source = selected && selected.source();
     return (
-      <div className={className}>
+      <div className={className} data-testid={`${dataTestId}-code-sample`}>
```
oisincoveney, line 40: The `data-testid` that's passed into `CodeSample` is only used as a prefix for the other test IDs within this component, so putting this ID at the top level matches the pattern here already. It might make more sense to keep this as just `data-testid={dataTestId}`, I'm open to suggestions

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/EmbedCodePane.jsx
```diff
@@ -7,13 +7,26 @@ import {
...
 import "ace/mode-clojure";
 import "ace/mode-javascript";
 import "ace/mode-ruby";
 import "ace/mode-html";
 import "ace/mode-jsx";
+import { trackStaticEmbedCodeCopied } from "metabase/public/lib/analytics";
+import { EMBED_MODAL_TABS } from "metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane";
+import CodeSample from "./CodeSample";
+
+const getEmbedModalTabLocation = tab => {
```
oisincoveney, line 19: Uses the locations that I mentioned earlier to get the right property enum for the snowplow event. Maybe we can throw this into `trackStaticEmbedCodeCopied` or as a helper function within that snowplow tracking file

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -8,7 +8,7 @@ import { AppearanceSettings } from "./AppearanceSettings";
 import type { OverviewSettingsProps } from "./OverviewSettings";
 import { OverviewSettings } from "./OverviewSettings";
 
-const TABS = {
+export const EMBED_MODAL_TABS = {
```
oisincoveney, line 11: Changed the name to something more descriptive since I need to use them in other files

frontend/src/metabase/query_builder/actions/sharing.ts
```diff
@@ -32,7 +32,11 @@ export const UPDATE_ENABLE_EMBEDDING = "metabase/card/UPDATE_ENABLE_EMBEDDING";
 export const updateEnableEmbedding = createAction(
   UPDATE_ENABLE_EMBEDDING,
   ({ id }: CardIdPayload, enable_embedding: boolean) =>
-    CardApi.update({ id, enable_embedding }),
+    CardApi.update({
+      id,
+      enable_embedding,
+      first_published_at: enable_embedding ? new Date().toISOString() : null,
```
oisincoveney, line 38: Same as before, where we might be able to do this on the BE

resources/migrations/001_update_migrations.yaml
```diff
@@ -5152,6 +5152,38 @@ databaseChangeLog:
         - dbms:
             type: mysql,mariadb
 
+  - changeSet:
```
oisincoveney, line 5155: Will need to check with BE if it makes more sense to just combine these migrations or if they need downgrades or different migrations for different DBs

resources/migrations/001_update_migrations.yaml
```diff
@@ -5152,6 +5152,38 @@ databaseChangeLog:
...
+        author: oisincoveney
+        comment: >-
+          Add report_dashboard.first_published_at
+        changes:
+          - addColumn:
+              tableName: report_dashboard
+              columns:
+                - column:
+                    name: first_published_at
+                    type: ${text.type}
```
oisincoveney, line 5182: I want this to be `timestamp.type` but I couldn't get the strings to budge. It might make more sense to handle this completely on the BE

resources/migrations/001_update_migrations.yaml
```diff
@@ -5152,6 +5152,38 @@ databaseChangeLog:
...
+      author: oisincoveney
+      comment: >-
+        Add report_card.first_published_at
+      changes:
+        - addColumn:
+            tableName: report_card
+            columns:
+              - column:
+                  name: first_published_at
+                  type: ${text.type}
```
calherries, line 5166: This will create a tz-aware column
```suggestion
                  type: ${timestamp_type}
```

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
npretto: `expectGoodSnowplowEvent` uses `underscore.isMatch` https://github.com/metabase/metabase/blob/eb44d76cc3643db309add31baa750aee6afded54/e2e/support/helpers/e2e-snowplow-helpers.js#L32
That in turns uses `==` https://underscorejs.org/docs/modules/isMatch.html to compare the values, which doesn't work for objects

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -20,6 +25,15 @@ import type {
...
+  Object.values(embeddingParams).reduce(
+    (acc, value) => {
+      acc[value] += 1;
+      return acc;
+    },
+    { disabled: 0, locked: 0, editable: 0 } as Record<string, number>,
```
npretto, line 34: I'm not sure if I'm right about it, but if EmbeddingParams can only have "disabled"|"locked"|"editable" as value, then we can change the type to 
```ts
export type EmbeddingParameters = {
  [key: string]: "disabled" | "locked" | "editable";
};
```
And remove the 'as Record...` here?
image

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/EmbedCodePane.jsx
npretto: if we ask product to rename `code_params` to `code_parameters` this could become `code_${tab}`, WDYT?

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
npretto: I'll look into it 👀

npretto: It seems that underscore doesn't have any way of doing a deep/recursive isMatch, the only "deep" function they have is `isEqual` but that fails is the first parameter has extra properties and that's not what we're doing at the moment.
Not sure if we should roll our own, it should be something like (this doesn't work for arrays but should work for the use cases we need at the moment)
```ts
const isDeepMatch = (obj, attrs) => {
  for (const [key, value] of Object.entries(attrs)) {
    if (!(key in obj)) {
      return false;
    }
    if (typeof value == "object") {
      if (!isDeepMatch(obj[key], value)) {
        return false;
      }
    } else {
      if (obj[key] !== value) {
        return false;
      }
    }
  }
  return true;
};
```

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
npretto: Also we could use [`_.countBy`](https://underscorejs.org/#countBy)

frontend/src/metabase/components/CopyButton/CopyButton.jsx
oisincoveney: https://github.com/metabase/metabase/pull/37933

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/AppearanceSettings.tsx
oisincoveney: Ok, sounds good. Maybe we need to get that PR in before this one.

frontend/src/metabase/public/lib/code-templates.ts
```diff
@@ -1,16 +1,38 @@
 import type {
   CodeSampleParameters,
   EmbeddingDisplayOptions,
-  EmbeddingParameters,
+  EmbeddingParametersValues,
 } from "./types";
 import { optionsToHashParams } from "./embed";
 
+function getIframeQuerySource(displayOptions: EmbeddingDisplayOptions) {
```
WiNloSt, line 8: Extract this into a function because every language basically use the same function. So I could just add some custom logic here.

frontend/src/metabase/public/lib/code-templates.ts
```diff
@@ -1,16 +1,38 @@
...
+  return JSON.stringify(
+    optionsToHashParams(
+      removeDefaultValueParameters(displayOptions, {
+        theme: "light",
+        hide_download_button: false,
+      }),
```
WiNloSt, line 14: I think this is better. So we keep the value as we want to store in React state e.g. `hide_download_button: false` but when using it we convert it to the value that's appropriate to each representation in this case we don't want to show `theme` and `hide_down_button` in URL parameters when their values are default values.

frontend/src/metabase/public/lib/code.ts
```diff
@@ -16,21 +16,25 @@ import {
 export const getEmbedClientCodeExampleOptions =
   (): ClientCodeSampleConfig[] => [
     {
+      id: "pug",
```
WiNloSt, line 19: this is the new value that'll be used in Select and also sent to analytics instead of using `name`

frontend/src/metabase/public/lib/types.ts
```diff
@@ -15,34 +15,41 @@ export type EmbedResourceParameter = {
   type: string;
 };
 
-export type EmbeddingParameters = Record<string, string>;
```
WiNloSt, line 18: We were misusing this type for both embedding parameters ("disabled" | "enabled" | "locked") and parameter values e.g. "Doohikey"

frontend/src/metabase/ui/components/overlays/Popover/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export type { PopoverBaseProps, PopoverProps } from "@mantine/core";
+export { getPopoverOverrides } from "./Popover.styled";
+
+const MantinePopoverDropdown = Popover.Dropdown;
+
+const PopoverDropdown = Object.assign(function PopoverDropdown(
+  props: PopoverDropdownProps,
+) {
+  return <MantinePopoverDropdown {...props} data-popover="mantine-popover" />;
```
WiNloSt, line 12: This little `data-popover` is a way for me to target Mantine popover in `mantinePopover()` E2E util.

snowplow/iglu-client-embedded/schemas/com.metabase/embed_flow/jsonschema/1-0-0
```diff
@@ -0,0 +1,198 @@
...
+      "type": [
+        "string",
+        "null"
+      ],
+      "enum": [
+        "html",
+        "csv",
+        "xlsx",
+        "json",
+        null
```
WiNloSt, line 159: I want to point out the difference between `"null"` in `type` and `null` in `enum`. As far as I understand. `"null"` in `type` tells us that we don't need to send this property all the time. But `null` in `enum` means that if you want to send the value in that property as `null`, you gotta have that in the enum as well.

I was a little confused at first, but hope this clears things up.

src/metabase/models/card.clj
```diff
@@ -181,7 +182,7 @@
 ;;; --------------------------------------------------- Revisions ----------------------------------------------------
 
 (def ^:private excluded-columns-for-card-revision
-  [:id :created_at :updated_at :entity_id :creator_id :public_uuid :made_public_by_id :metabase_version
+  [:id :created_at :updated_at :entity_id :creator_id :public_uuid :made_public_by_id :metabase_version :initially_published_at
```
WiNloSt, line 185: This prevents making a revision when `initially_published_at` changes.

src/metabase/models/dashboard.clj
```diff
@@ -227,7 +229,7 @@
    ;;   lower-numbered positions appearing before higher numbered ones.
    ;; TODO: querying on stats we don't have any dashboard that has a position, maybe we could just drop it?
    :public_uuid :made_public_by_id
-   :position])
+   :position :initially_published_at])
```
WiNloSt, line 232: This prevents making a revision when `initially_published_at` changes.

test/metabase/util/embed_test.clj
```diff
@@ -54,3 +58,26 @@
...
+      (testing "should populate `initially_published_at` when a Card's enable_embedding is changed to true"
+        (t2.with-temp/with-temp [model card {:enable_embedding false}]
+          (is (= nil (:initially_published_at card)))
+          (t2/update! model (u/the-id card) {:enable_embedding true})
+          (is (not= nil (:initially_published_at (t2/select-one model :id (u/the-id card)))))))
+      (testing "should keep `initially_published_at` value when a Card's enable_embedding is changed to false"
+        (t2.with-temp/with-temp [model card {:enable_embedding true :initially_published_at now}]
+          (is (not= nil (:initially_published_at card)))
+          (t2/update! model (u/the-id card) {:enable_embedding false})
+          (is (= (t/offset-date-time now) (:initially_published_at (t2/select-one model :id (u/the-id card)))))))
```
WiNloSt, line 74: I was having a trouble comparing time from the model and time created with `#t "2023"`. It turns out they're different types. I found this works, but I'm not sure what's the ideal way of doing this.

WiNloSt:
@calherries I saw you were the original reviewer of this PR, now that the PR is updated. Could you please take another look 🙏

frontend/src/metabase/public/lib/types.ts
oleggromov: My 2c here: it's really unclear how `EmbeddingParametersOptions` different from `EmbeddingParametersSettings`. Naming is hard, but maybe there's a better way? 

WDYT if we keep the name `EmbeddingParameters` for what is coming in the `embedding_params` key and rename `EmbeddingParametersOptions` into something like `EmbeddingParameterVisibility`?

WiNloSt: I like `EmbeddingParameterVisibility`.

And `EmbeddingParameters` could also work since it's from `embedding_params` like you said.

I agree with what you've suggested here.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ServerEmbedCodePane.tsx
```diff
@@ -45,15 +48,19 @@ export const ServerEmbedCodePane = ({
...
 
   className,
 }: EmbedCodePaneProps): JSX.Element | null => {
   const selectedServerCodeOption = serverCodeOptions.find(
-    ({ name }) => name === selectedServerCodeOptionName,
+    ({ id }) => id === selectedServerCodeOptionId,
   );
 
+  const canWhitelabel = useSelector(getCanWhitelabel);
+  const shouldShowDownloadData = canWhitelabel && resourceType === "question";
```
npretto, line 62: @WiNloSt what's the connection between downloading data and whitelabeling? is it a proxy to check if we have a valid license?

WiNloSt: Yep, it's the original condition when we render the download option on embedding.

So the condition was actually if users have valid license then show this and also it needs to be questions.

npretto: interesting, so we're using a unrelated feature toggle to detect if we're on pro/ee because we don't have a specific feature toggle for that?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -356,13 +431,24 @@ function getHasSettingsChanges({
...
     if (embeddingParams[key] !== "disabled") {
       result[key] = embeddingParams[key];
     }
 
     return result;
-  }, {} as EmbeddingParameters);
+  }, {} as EmbeddingParametersSettings);
+}
+
+function convertResourceParametersToEmbeddingParams(
```
npretto, line 445: @oleggromov is this the same logic you implementend in your PR?
It seems to convert the missing parameters to "disabled"

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ServerEmbedCodePane.tsx
WiNloSt: That was introduce will before we revamped the token features. Maybe we have more suitable selectors for that now. But I need to check.

npretto (approved):
Did a quick read and didn't see any issue in the code, I didn't check if the events send what they should though and I'd be happy if someone else approved too as it's a big-ish PR

e2e/support/helpers/e2e-snowplow-helpers.js
```diff
@@ -29,11 +29,58 @@ export const expectGoodSnowplowEvent = (eventData, count = 1) => {
     "micro/good",
     ({ body }) =>
       body.filter(snowplowEvent =>
-        _.isMatch(snowplowEvent?.event?.unstruct_event?.data?.data, eventData),
```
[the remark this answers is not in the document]
oisincoveney: Nice! we could definitely use this

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -280,11 +801,24 @@ function createPublicResourceLink(resource, id) {
...
 }
+
+function enableEmbeddingForResource({ resource, id }) {
+  const endpoint = resource === "question" ? "card" : "dashboard";
+  cy.request("PUT", `/api/${endpoint}/${id}`, {
+    enable_embedding: true,
+  });
+}
+
+function closeTo(value, offset) {
```
oisincoveney, line 820: this is really smart. nicely done 😄

frontend/src/metabase/dashboard/actions/sharing.ts
oisincoveney: yeah i think for now this is ok. From what I understand we have a few better utils for handling things like this on the FE but we can talk with some BEs in the future to make this nicer. It works for now

test/metabase/util/embed_test.clj
oisincoveney: @metabase/backend-developers could anybody chime in here? I think it would be great to get some input here on this

oisincoveney (commented):
This all looks good, if we can get some input from BE for the `embed_test` then it would be great but it still works so I'm good

resources/migrations/001_update_migrations.yaml
```diff
@@ -5317,6 +5317,38 @@ databaseChangeLog:
         - dbms:
             type: mysql,mariadb
 
+  - changeSet:
+      id: v49.2024-02-02T11:27:49
+      author: oisincoveney
+      comment: >-
+        Add report_card.initially_published_at
```
calherries, line 5324: any downsides to renaming this `published_at`? I'm just a boy with limited keystrokes and 120 characters of column width

src/metabase/util/embed.clj
```diff
@@ -108,6 +109,16 @@
...
 
+(defn maybe-populate-initially-published-at
+  "Populate `initially_published_at` if embedding is set to true"
+  [{:keys [enable_embedding initially_published_at] :as card}]
+  (cond
+    (and (true? enable_embedding) (nil? initially_published_at))
+    (assoc card :initially_published_at (now))
+
+    :else
+    card))
```
calherries, line 120: Nice use of destructuring!

1. It's idiomatic to use `cond->` when we're optionally modifying something. That's what it's for
2. `card` should be `card-or-dashboard` to reflect its use with both
```suggestion
  [{:keys [enable_embedding initially_published_at] :as card-or-dashboard}]
  (cond-> card-or-dashboard
    (and (true? enable_embedding) (nil? initially_published_at))
    (assoc :initially_published_at (now))))
```

resources/migrations/001_update_migrations.yaml
calherries: Ah okay, I didn't know that. `first_published_at` then?

oisincoveney: That was actually the name in the first iteration 😅 I'm not opposed to changing it but I think it's decent for now since we want to get this in

src/metabase/util/embed.clj
oisincoveney: Good calls! I didn't know `cond->` existed when I was writing it (of course lol) but that's pretty nice. I'll keep it for the future

test/metabase/util/embed_test.clj
```diff
@@ -54,3 +58,26 @@
...
+      (testing "should keep `initially_published_at` value when `enable_embedding` is already set to true"
+        (t2.with-temp/with-temp [model card {:enable_embedding true :initially_published_at now}]
+          (t2/update! model (u/the-id card) {:enable_embedding true})
+          (is (= (t/offset-date-time now) (:initially_published_at (t2/select-one model :id (u/the-id card)))))))
+      (testing "should keep `initially_published_at` value when `enable_embedding` is already set to false"
+        (t2.with-temp/with-temp [model card {:enable_embedding false}]
+          (is (= nil (:initially_published_at card)))
+          (t2/update! model (u/the-id card) {:enable_embedding false})
+          (is (= nil (:initially_published_at (t2/select-one model :id (u/the-id card))))))))))
```
calherries, line 83: You should use a mock clock to check it is indeed getting the current time
```
(deftest maybe-populate-initially-published-at-test
  (let [previously (t/offset-date-time #t "2022-09-01T12:34:56Z")
        now        (t/offset-date-time #t "2023-01-01T01:01:01Z")]
    (mt/with-clock (t/zoned-date-time now)
      (doseq [model [:model/Card :model/Dashboard]]
        (testing "should populate `initially_published_at` when a Card's enable_embedding is changed to true"
          (t2.with-temp/with-temp [model card {:enable_embedding false}]
            (is (= nil (:initially_published_at card)))
            (t2/update! model (u/the-id card) {:enable_embedding true})
            (is (= now (:initially_published_at (t2/select-one model :id (u/the-id card)))))))
        (testing "should keep `initially_published_at` value when a Card's enable_embedding is changed to false"
          (t2.with-temp/with-temp [model card {:enable_embedding true :initially_published_at previously}]
            (is (= previously (:initially_published_at card)))
            (t2/update! model (u/the-id card) {:enable_embedding false})
            (is (= previously (:initially_published_at (t2/select-one model :id (u/the-id card)))))))
        (testing "should keep `initially_published_at` value when `enable_embedding` is already set to true"
          (t2.with-temp/with-temp [model card {:enable_embedding true :initially_published_at previously}]
            (t2/update! model (u/the-id card) {:enable_embedding true})
            (is (= previously (:initially_published_at (t2/select-one model :id (u/the-id card)))))))
        (testing "should keep `initially_published_at` value when `enable_embedding` is already set to false"
          (t2.with-temp/with-temp [model card {:enable_embedding false}]
            (is (= nil (:initially_published_at card)))
            (t2/update! model (u/the-id card) {:enable_embedding false})
            (is (= nil (:initially_published_at (t2/select-one model :id (u/the-id card)))))))))))
```

test/metabase/util/embed_test.clj
calherries: Yeah those strings are a little weird for OffsetDateTimes with zero offset. You just have to know that `#t "2023-01-01T01:01:01Z"` is a ZonedDateTime in the UTC time zone. To get an OffsetDateTime with zero offset (so UTC), you have to convert it `(t/offset-date-time #t "2023-01-01T01:01:01Z")`.

src/metabase/util/embed.clj
calherries: Oh, that `now` function returns a weird form: 
```
(now)
[:metabase.util.honey-sql-2/typed :%now {:database-type "timestamptz"}]
```

Instead of `(now)` you should use `(t/offset-date-time)`, which returns the current system time with an offset.
`t` is the alias for `java-time.api`, which you can require with `[java-time.api :as t]`.

test/metabase/util/embed_test.clj
calherries: This test looks good though!

src/metabase/util/embed.clj
```diff
@@ -108,6 +109,13 @@
   (or (get-in unsigned-token keyseq)
       (throw (ex-info (tru "Token is missing value for keypath {0}" keyseq) {:status-code 400}))))
 
+(defn maybe-populate-initially-published-at
+  "Populate `initially_published_at` if embedding is set to true"
+  [{:keys [enable_embedding initially_published_at] :as card-or-dashboard}]
+  (cond-> card-or-dashboard
+    (and (true? enable_embedding) (nil? initially_published_at))
+    (assoc :initially_published_at (now))))
```
calherries, line 117: Don't use `(now)` here, use `:%now`.
```suggestion
    (assoc :initially_published_at :%now)))
```

calherries (approved):
LGTM

src/metabase/models/dashboard.clj
```diff
@@ -97,7 +98,8 @@
   (u/prog1 dashboard
```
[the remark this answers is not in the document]
WiNloSt: Just saw this didn't make it.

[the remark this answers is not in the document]
oisincoveney: Huh. don't know how I missed that. I'll put that in a cheeky quick PR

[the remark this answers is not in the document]
WiNloSt: nevermind I was clicking directly on the file name of this comment which links me to a snapshot of the code. I just checked the actual code again and it's all good.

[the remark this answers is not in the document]
WiNloSt: I took that back we call `maybe-populate-something-i-do-not-actually-remember-the-name` twice

image

PR #36121, opened 2023-11-27
Sync dynamic tables for snowflake

qnkhuat:
Closes #35194

This bug has 2 problems:

1. 
Currently, if you have a snowflake database containing dynamic tables, sync will fail for those dynamic tables with an exception.

```net.snowflake.client.jdbc.SnowflakeSQLException: SQL compilation error: show [constraint] command doesn't support this domain.```

This is because we're trying to find pks and fks for those dynamic tables, but the snowflake jdbc is throwing an error because apparently, they're not supported for dynamic tables.

2. 
We used the [getColumns method
](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-) to get a list of columns for each table, but currently there is a bug in snowflake jdbc where it returns an empty list for dynamic tables. I've filed  [bug](https://github.com/snowflakedb/snowflake-jdbc/issues/1574) on their repo, for now we'll rely on the fallback method that call a select query then use it to figure out column details

PR #37620, opened 2024-01-12
Fix snowflake sometimes does not find tables

### LINKED ISSUE #38135: Metabase won't see schemas other than public or the default namespace of the user in snowflake
### Describe the bug

I can't reproduce this by any means, but there are a few syptoms:
1) you moved from 47 to 48 and schemas that are not public OR the schema is not the default schema of the user will dissapear
2) if you connect again the same database, everything will be synced, no problems
3) users use a role, which have limited permissions to the database (usage on the db/schema, select on the table/s)

### To Reproduce

No way to reproduce

### Expected behavior

P1

### Logs

NA

```JSON
v48
```

### Severity

P1

### Additional context

_No response_

qnkhuat:
Fixes #38135

Fix https://discourse.metabase.com/t/snowflake-broken-in-0-48-1/63849

We changed to use getTables JDBC method to sync snowflake in [a recent PR](https://github.com/metabase/metabase/pull/36121/files?diff=split&w=0#diff-5472d9d00fa628881a6887a87e2be860f2d679f85aeada5d9f2c86173646800aR419-R426)
because I thought using JDBC will be more reliable and it'll be consistent across the drivers.

But it turns out it's not; Snowflake driver has a history of being buggy like this, and it's biting us again. In fact, we have this comment from 3 years ago, https://github.com/metabase/metabase/blob/ea22d293576e7c404152727d16a010df8facd2c5/modules/drivers/snowflake/src/metabase/driver/snowflake.clj#L222, but unfortunately, it was removed by accident I think.

I added back a comment with the word "IMPORTANT" to it, hopefully it won't get deleted this time.

I can't reproduce the bug and so is Luiggi.


calherries:
In theory this should fix the bug, but doesn't this imply we should eliminate usage of `getTables` everywhere it is used with snowflake e.g. [here](https://github.com/metabase/metabase/blob/dab994c424ee4678b3208e893362bb581c6d4c37/modules/drivers/snowflake/src/metabase/driver/snowflake.clj#L509)? This PR only changes one place where getTables is used.

```
           ;; IMPORTANT: using the JDBC `.getTables` method is buggy -- it works sometimes but other times randomly
           ;; we have tried to use it but people reported that some schemas disappear on their instances.
           ;; See https://discourse.metabase.com/t/snowflake-broken-in-0-48-1/63849
```

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -407,26 +406,56 @@
                                    :from   [[(qp.store/with-metadata-provider (u/the-id database)
                                                (sql.qp/->honeysql driver table))]]}))
 
+(defn- show-command-sql
+  "Execute the SHOW command and returns a reducible result."
```
qnkhuat, line 410:
```suggestion
```

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -493,9 +523,11 @@
...
      (catch SnowflakeSQLException e
        ;; dynamic tables doesn't support kks so it's fine to suppress the exception
-       (if (= "DYNAMIC_TABLE"
-              (-> (.getTables metadata db-name-or-nil schema-name table-name nil)
-                  jdbc/result-set-seq first :table_type))
+       (if (some?
+            (first (jdbc/query {:connection conn}
+                               (show-command-sql
+                                "DYNAMIC TABLES"
+                                :database db-name-or-nil :schema schema-name :like table-name))))
```
calherries, line 530: Based on the implemenation of `show-command-sql`, providing `:database` and `:schema` should be mutually exclusive with each other. If a `:schema` is provided, then `:database` should not be provided and vice versa. If both are provided it would compile to `SHOW DYNAMIC TABLES IN DATABASE <database> SCHEMA <schema>` which is invalid. I think that should be in the docstring or preconditions of `show-command-sql` too, to prevent misuse.

qnkhuat: I've updatd to make it works for both cases

calherries (approved):
Looks good

qnkhuat:
On second thought, let's see if this actually solve people's problem in https://discourse.metabase.com/t/snowflake-broken-in-0-48-1/63849

@paoliniluis would you bind sending them the latest build?

paoliniluis:
sure! thanks @qnkhuat!

paoliniluis:
@qnkhuat when I test with the release branch I get schemas and tables as "show objects..." returns fine and then the "select true as ..." query also returns fine.

When I test this branch, the "show objects..." returns fine but the "select true as..." queries fail with "Cannot perform SELECT. This session does not have a current database. Call 'USE DATABASE', or use a qualified name."

same user, same permissions. The user and password is saved on our password manager as metabasesuccess

dragonsahead (changes requested):
Just tested and does not return active tables

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -406,28 +405,62 @@
   (sql-jdbc/query driver database {:select [:*]
                                    :from   [[(qp.store/with-metadata-provider (u/the-id database)
                                                (sql.qp/->honeysql driver table))]]}))
+(defn- show-command-sql
+  ;; IMPORTANT: prefer using this command over using the JDBC `.getTables` method because the jdbc is buggy
+  ;; it works sometimes but other times randomly.
```
camsaul, line 410:
```suggestion
  ;; it works sometimes but other times randomly does not.
```

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -406,28 +405,62 @@
...
+  ;; we have tried to use it but people reported that some schemas disappear on their instances.
+  ;; See https://discourse.metabase.com/t/snowflake-broken-in-0-48-1/63849
+  [object-type & {:keys [like database schema]}]
+  ;; in case the name is escaped for schema using escape-name-for-metadata, we undo that just to be sure
```
camsaul, line 414: Not really super important but I really don't think you need to unescape stuff since the only place that uses it is the default implementation of `describe-database`, and since you're doing a custom implementation you shouldn't need to worry about it

camsaul: Either way it's always better to not escape stuff in the first place than to escape it and unescape it, less moving parts and less ways to accidentally do something wrong

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -406,28 +405,62 @@
   (sql-jdbc/query driver database {:select [:*]
                                    :from   [[(qp.store/with-metadata-provider (u/the-id database)
                                                (sql.qp/->honeysql driver table))]]}))
+(defn- show-command-sql
+  ;; IMPORTANT: prefer using this command over using the JDBC `.getTables` method because the jdbc is buggy
+  ;; it works sometimes but other times randomly.
+  ;; we have tried to use it but people reported that some schemas disappear on their instances.
+  ;; See https://discourse.metabase.com/t/snowflake-broken-in-0-48-1/63849
+  [object-type & {:keys [like database schema]}]
```
camsaul, line 413: Making this an `mu/defn` and giving `object-type` and `:enum` schema would make it a lot clearer for people using this what kinds of things it supports. A docstring with an example usage would be helpful too

modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj
```diff
@@ -420,3 +420,15 @@
             result-comment (second (re-find #"-- (\{.*\})" result-query))
             result-map (json/read-str result-comment)]
         (is (= expected-map result-map))))))
+
+(deftest show-command-sql-test
```
camsaul, line 424:
```suggestion
(deftest ^:parallel show-command-sql-test
```

camsaul (changes requested):
Overall this looks really good to me, but I think we need to be super careful around escaping stuff and keep in mind that Snowflake identifiers are allowed to contain wacky characters https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers

I would be more comfortable if we just made sure we never called `escape-name-for-metadata` in the first place (which should be easy enough since the one place that uses it is in the default implementation of `describe-database` that we're replacing, and then made sure we're properly escaping quotes in that `show-command-sql` function.

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -406,28 +405,62 @@
...
+  ;; in case the name is escaped for schema using escape-name-for-metadata, we undo that just to be sure
+  (let [quote-and-unescape #(quote-name (str/replace % "\\_" "_"))]
+    (str/join
+     " "
+     (cond-> [(format "SHOW %s" object-type)]
+       (some? like)
+       (conj (format "LIKE '%s'" like))
```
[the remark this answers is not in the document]
camsaul: Oh actually the dox do say you can create a Table with double quotes in its name 

> To use the double quote character inside a quoted identifier, use two quotes. For example:
>
> ```sql
> create table "quote""andunquote""" ...
> ```
>
> creates a table named:
>
> ```
> quote"andunquote"
> ```

paoliniluis:
@camsaul what do you think about sending schema.table instead of database.schema.table when we do the simple-select-probe-query on sync?

camsaul:
@paoliniluis does it do something different?

qnkhuat:
Closing in favor of https://github.com/metabase/metabase/pull/38160/files

PR #37624, opened 2024-01-12
Introduce mechanism for explicitly marking settings for export in their definition

### LINKED ISSUE #36240: Create a test to prevent serialization bugs with new settings
This bug https://github.com/metabase/metabase/issues/36232 happened because we created new settings, but they weren't added to the list of settings to be serialized, when they should have been. We should have a test that forces engineers introducing new settings to choose whether or not to add it to the exported list of settings.

crisptrutski:
Closes #36240 

### Description

This change starts the deprecation of `setting/exported-settings` in favor of a more explicit `:export?` key in the definition of the setting. The rationale is that it's too easy to forget to add new settings to this set.

The idea is to require an explicit true or false value for this key in the macro, in a subsequent PR. That step is deferred to avoid adding a large amount of noise to this PR.

~~I'm not thrilled about using metadata for filtering, but I wasn't thrilled about the other options I thought of either - looking to discuss offline with @calherries as there's a lot to say on the matter.~~ We selected another approach.

### How to verify

~~I'm not sure whether there is sufficient test coverage for this, it's probably worth adding more.~~ While it would be nice to have an e2e test, the new unit tests should give us enough confidence to merge.

### Checklist

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -323,7 +323,11 @@
    ;; should be used for most non-sensitive settings, and will log the value returned by its getter, which may be a
    ;; the default getter or a custom one.
    ;; (default: `:no-value`)
-   :audit       (s/maybe (s/enum :never :no-value :raw-value :getter))})
+   :audit       (s/maybe (s/enum :never :no-value :raw-value :getter))
+
+   ;; TODO: make this required and deprecate setting/exported-setting
+   (s/optional-key :export?)     s/Bool ; should this setting be serialized?
+   })
 
 (defonce ^{:doc "Map of loaded defsettings"}
   registered-settings
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -168,7 +168,7 @@
   [:key])
 
 (def ^:private exported-settings
-  '#{application-colors
+  '#{#_application-colors
      application-favicon-url
      application-font
      application-font-files
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -212,10 +212,12 @@
      uploads-database-id
      uploads-schema-name})
 
+(declare export?)
+
 (defmethod serdes/extract-all "Setting" [_model _opts]
-  (for [{:keys [key value] :as s} (admin-writable-site-wide-settings
+  (for [{:keys [key value]} (admin-writable-site-wide-settings
                              :getter (partial get-value-of-type :string))
-        :when (:export? (meta s))]
+        :when (export? key)]
     {:serdes/meta [{:model "Setting" :id (name key)}]
      :key key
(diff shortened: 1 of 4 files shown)


src/metabase/models/setting.clj
```diff
@@ -168,7 +168,7 @@
   [:key])
 
 (def ^:private exported-settings
-  '#{application-colors
+  '#{#_application-colors
```
crisptrutski, line 171: Decided to temporarily leave this commented here, so we still have a complete list in the code. This will serve as a useful reference for adding further tests, after which time the entire val will be removed.

src/metabase/models/setting.clj
```diff
@@ -1281,6 +1287,15 @@
                (when api/*is-superuser?*
                  [:admin]))))
 
+(defn- filtered-user-facing-settings
```
crisptrutski, line 1290: Not thrilled with this name

calherries: Just making sure - what's the purpose of this change? Is this just a refactoring?

crisptrutski: Yeah it's incidental, and happy to back it out. Made the change earlier when I'd anticipate changing `admin-writable-site-wide-settings` in some intrusive way, but ended up not going that route.

 I still think it's nice to DRY it up so just left it in.

piranha: I'm thinking that maybe it should be `make-user-facing-settings` or something in that vein? It's not only filtering them.

src/metabase/models/setting.clj
```diff
@@ -323,7 +325,11 @@
    ;; should be used for most non-sensitive settings, and will log the value returned by its getter, which may be a
    ;; the default getter or a custom one.
    ;; (default: `:no-value`)
-   :audit       (s/maybe (s/enum :never :no-value :raw-value :getter))})
+   :audit       (s/maybe (s/enum :never :no-value :raw-value :getter))
+
+   ;; TODO: make this required and deprecate setting/exported-setting
```
piranha, line 330: Maybe that should happen in #37640?

calherries: Chris wrote in the description of that PR that he'll make it required in a third PR. 

> The addition of explicit :export? false annotations, and making the key a required one, will be done in a subsequent PR.

piranha: Ah, right, description has merged in my head and I imagined the change coming in as two PRs in total. :)

sansolovyov (approved):
Aside from a nitpick — LGTM 😁

src/metabase/models/setting.clj
crisptrutski: Good point, I'll iterate on the name to reflect the transformation

PR #37636, opened 2024-01-12
Use UTF8 encoding for dashboard preview page

kulyk:
Adds `<meta charset="utf-8" />` to the dashboard subscription preview page

### DIFF
--- a/src/metabase/api/pulse.clj
+++ b/src/metabase/api/pulse.clj
@@ -291,8 +291,7 @@
    :body    (preview/style-tag-from-inline-styles
              (html5
                [:head
-                [:meta {:nonce "%NONCE%" ;; this will be str/replaced by 'style-tag-nonce-middleware
-                        :charset "utf-8"}]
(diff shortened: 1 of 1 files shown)


src/metabase/api/pulse.clj
```diff
@@ -291,6 +291,8 @@
    :body    (preview/style-tag-from-inline-styles
              (html5
                [:head
+                [:meta {:nonce "%NONCE%" ;; this will be str/replaced by 'style-tag-nonce-middleware
+                        :charset "utf-8"}]
```
adam-james-v, line 295: This is nearly correct! You don't actually need that :nonce in this case. The nonce is needed for styles to pass our CSP, but setting the meta here doesn't need anything like that.

```suggestion
                [:meta {:charset "utf-8"}]
```

adam-james-v (approved):
See my suggestion for a one line change. 
Otherwise this is good. I've tried the change locally too and it's working.

PR #37640, opened 2024-01-12
Use explicit definition for all exported settings, remove global var

crisptrutski:
### Description

This cleans up a bit after https://github.com/metabase/metabase/pull/37624, getting rid of the deprecated `exported-settings` var.

The addition of explicit `:export? false` annotations, and making the key a required one, will be done in a subsequent PR.

### How to verify

There's a temporary test asserting that we have not diverged from the previous definition. That test is to give the reviewer confidence, and we can delete it before merging.

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1245,8 +1245,8 @@
                (when api/*is-superuser?*
                  [:admin]))))
 
-(defn- filtered-user-facing-settings
-  "Docstrings are hard"
+(defn- user-facing-settings-matching
+  "Returns the user facing view of the registered settings satisfying the given predicate"
   [pred options]
   (into
     []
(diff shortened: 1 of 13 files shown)


calherries:
@crisptrutski I think you meant to switch this from draft to open

calherries (approved):
Looks solid. Since you're creating a series of PRs like this, I would add a list to either a comment or the description in the [original issue](https://github.com/metabase/metabase/issues/36240), so someone can easily follow all the steps that are being done to close the issue.

crisptrutski:
@piranha going to wait on your tick here as well

src/metabase/models/setting.clj
```diff
@@ -1293,8 +1245,8 @@
                (when api/*is-superuser?*
                  [:admin]))))
 
-(defn- filtered-user-facing-settings
-  "Docstrings are hard"
+(defn- user-facing-settings-matching
+  "Returns the user facing view of the registered settings satisfying the given predicate"
```
crisptrutski, line 1249: @piranha here's the rename i went with based on your nit in the previous PR

sansolovyov (approved):
I just had nothing to add 😁

src/metabase/models/setting.clj
piranha: Nice!

PR #37646, opened 2024-01-12
[Manual backport] Optimize canonicalization of MBQL clauses (#37539)

### LINKED ISSUE #37245: Mongo GUI question is way slower than it's native twin

metamben:
Fixes #37245.

### DIFF
--- a/shared/src/metabase/mbql/normalize.cljc
+++ b/shared/src/metabase/mbql/normalize.cljc
@@ -634,25 +634,45 @@
 
 (defn- canonicalize-mbql-clauses
   "Walk an `mbql-query` an canonicalize non-top-level clauses like `:fk->`."
-  [mbql-query]
-  (walk/prewalk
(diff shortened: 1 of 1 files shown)


lbrdnk (approved):
Diffs match.

PR #37653, opened 2024-01-12
Fix admin landing page e2e test flakes

sloansparger:
### Description

This PR should solve the e2e flakiness introduced in  #37322.

I reduce the tests to only test a single input and make sure to wait for the full settings refetch query before continuing the test. Unit tests should cover a greater breadth of inputs, so reducing the test size should be fine. I believe there was a race condition due to rapidly setting inputs & not waiting for the settings request to return a result. Waiting for both requests should solve.

### How to verify

- e2e tests should pass consistently in CI

### Checklist

### DIFF
--- a/e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
+++ b/e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
@@ -286,53 +286,35 @@ describeEE("formatting > whitelabel", () => {
   describe("Landing Page", () => {
     beforeEach(() => {
       cy.intercept("PUT", "/api/setting/landing-page").as("putLandingPage");
+      cy.signInAsAdmin();
+      cy.visit("/admin/settings/whitelabel");
     });
 
     it("should allow users to provide internal urls", () => {
-      cy.signInAsAdmin();
+      cy.findByTestId("landing-page").click().clear().type("/test-1").blur();
+      cy.wait("@putLandingPage");
 
-      cy.visit("/admin/settings/whitelabel");
-      cy.location("origin").then(origin => {
-        // test root / path
-        testValidLandingPageInput({ input: "/", expected: origin + "/" });
-
-        // test relative url
-        testValidLandingPageInput({
-          input: "/test-1",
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
```diff
@@ -286,53 +286,36 @@ describeEE("formatting > whitelabel", () => {
...
+    });
 
-        // test no input
-        testValidLandingPageInput({ input: "", expected: origin + "/" });
+    it("should allow users to provide internal urls", () => {
+      cy.findByTestId("landing-page").click().clear().type("/test-1").blur();
+      cy.wait(["@putLandingPage", "@getSettings"]);
 
-        // test absolute url matching the page's origin
-        testValidLandingPageInput({
```
npfitz, line 309: It looks like we've removed a few tests here for setting the landing page to the root path, no path, and absolute urls. If the cause of the flakes was due to testing these all back to back in the same test, can we split them into their own tests for better reliability while still maintaining coverage?

npfitz (changes requested):
Just the one comment. Otherwise, looks good!

e2e/test/scenarios/admin-2/whitelabel.cy.spec.js
sloansparger: I initially added this e2e test to test various inputs, but after folks thought of some other potential edge cases I added more thorough [unit testing here](https://github.com/metabase/metabase/blob/34ebffa9864319fdcf6cc5224a75fbd92c7629e7/enterprise/frontend/src/metabase-enterprise/whitelabel/components/LandingPageWidget/utils.unit.spec.ts). I think it makes sense for this e2e test to focus primarily on does the feature work as intended instead of validating various types of inputs.

npfitz: Gotcha. I think that's fair

npfitz (approved):

PR #37654, opened 2024-01-12
Fix isEditable check

ranquild:
Context https://metaboat.slack.com/archives/C0645JP1W81/p1705076059720359

Calling `rootTable` in MLv1 would break our changes to `databaseId` method because it would try to create a query from a card that doesn't have `dataset_query`. Cannot repro locally.


NevRA (approved):
LGTM, to unblock stats

PR #37670, opened 2024-01-12
Introduce same-column-lists?

### LINKED ISSUE #37517: [MLv2] [BE] [Bug] `returnedColumns` does not handle columns with the same name
`name` of the columns returned by `returnedColumns` does not match QP results. It's important for us that names of the columns generated by MLv2 match QP.

In this example `nextColumnNames` are generated by MLv2 and `previousColumnNames` are coming from query results (QP):

Screenshot 2024-01-10 at 18 24 34

Duplicate column names can be achieved e.g. by adding a self join to the query.

metamben:
Fixes #37517.

This function makes it possible for the FE to decide if the query has to be re-run.
For more context, see the Slack [discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1705008072684969?thread_ts=1705002980.394099&cid=C04CYTEL9N2).

### DIFF
--- a/dev/src/dev/fe_helpers.cljs
+++ b/dev/src/dev/fe_helpers.cljs
@@ -28,3 +28,9 @@
   Hack: This relies on a dev-mode-only global property that's set whenever a Question object is converted to MLv2."
   []
   (.-__MLv2_query js/window))
+
+(defn current-query-results
+  "Gets the current query results from the reduce store.
+  This is normally a JS array or result_metadata objects."
+  []
+  (.. (redux-state) -qb -queryResults))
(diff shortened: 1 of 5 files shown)


ranquild:
@metamben Maybe I found a way to streamline FE checks and make them work without check function. I'll know in several hours for sure.

ranquild:
I managed to find a way to simply QB so this function is no longer needed. Sorry for the confusion - for some reason I didn't understand the way to make it before.

metamben:
Closing because [it's not needed](https://metaboat.slack.com/archives/C04CYTEL9N2/p1705304978982559).

Issue #36397, opened 2023-12-05
BE flake: metabase.driver-test/check-can-connect-before-sync-test for bigquery


calherries:
`metabase.driver-test/check-can-connect-before-sync-test` flakes for bigquery

https://github.com/metabase/metabase/actions/runs/7098863639/job/19321687227?pr=36394#step:3:523

calherries:
I took a stab at reproducing this [here](https://github.com/metabase/metabase/pull/36409) but couldn't. The error is a timed out query so I'm going to suggest leaving this for now. I'll keep it open in case someone sees the same error but I'll remove it from our queue.

calherries:
This [flaked again](https://github.com/metabase/metabase/actions/runs/7612505007/job/20730281921?pr=37996)

calherries:
Closing again because the flake was on a branch that didn't contain the commit from the PR

PR #37685, opened 2024-01-15
Change check-can-connect-before-sync-test to help debug flake

calherries:
This PR changes metabase.driver-test/check-can-connect-before-sync-test to help debugging its flakiness (issue: https://github.com/metabase/metabase/issues/36397).

An example run that fails: https://github.com/metabase/metabase/actions/runs/7488691603/job/20383722724#step:3:945. 
If it fails in the same way after this change, we'll see the logged message in the test output for sense check 1.


crisptrutski (approved):
LGTM

calherries:
Closing because the test already flaked

PR #37691, opened 2024-01-15
Remove unused code from QuestionDataSource

ranquild:
Related to https://github.com/metabase/metabase/issues/37396

The component is not used in ObjectDetail and we can remove code for object detail here.


uladzimirdev (approved):
🤞

PR #37692, opened 2024-01-15
Migrate QuestionDataSource to MLv2

### LINKED ISSUE #37396: [MLv2] [FE] Migrate QuestionDataSource
```[tasklist]
- [ ] https://github.com/metabase/metabase/pull/37691
- [ ] https://github.com/metabase/metabase/pull/37692
```

ranquild:
Closes https://github.com/metabase/metabase/issues/37396


frontend/src/metabase-lib/join.ts
```diff
@@ -243,7 +243,10 @@ export type PickerInfo = {
   isModel?: boolean;
 };
 
-export function pickerInfo(query: Query, metadata: Joinable): PickerInfo {
+export function pickerInfo(
+  query: Query,
+  metadata: Joinable,
+): PickerInfo | null {
```
ranquild, line 249: returns `null` when the joined table/card isn't available, e.g. due to sandboxing

kamilmielnik: > returns `null` when the joined table/card isn't available, e.g. due to sandboxing

This would make a very helpful JSDoc for this function.

frontend/src/metabase/query_builder/components/view/QuestionDataSource.jsx
```diff
@@ -172,16 +173,24 @@ function getDataSourceParts({ question }) {
...
-      ...query.joins().map(j => j.joinedTable()),
-    ].filter(Boolean);
+      ...Lib.joins(query, -1)
+        .map(join => Lib.pickerInfo(query, Lib.joinedThing(query, join)))
+        .map(pickerInfo => {
+          if (pickerInfo?.tableId != null) {
+            return metadata.table(pickerInfo.tableId);
+          } else if (pickerInfo?.cardId != null) {
+            return metadata.table(getQuestionVirtualTableId(pickerInfo.cardId));
+          }
```
kamilmielnik, line 191: It's easier to read functions when:
- there's no `else` after `return`
- either all or none paths have `return` statements

```suggestion
          }
          
          if (pickerInfo?.cardId != null) {
            return metadata.table(getQuestionVirtualTableId(pickerInfo.cardId));
          }
          
          return undefined;
```

kamilmielnik (approved):

frontend/src/metabase-lib/join.ts
kamilmielnik: dd101ac279c4990ab3eefa19591306c642ac92b7

frontend/src/metabase/query_builder/components/view/QuestionDataSource.jsx
kamilmielnik: 04de16fbb278fa9810f30dcccea93472296ceaaa

PR #37639, opened 2024-01-12
[MLv2] [FE] Migrate object detail

kamilmielnik:
Closes #37628
Depends on #37648

PR #37694, opened 2024-01-15
[MLv2] [FE] Migrate data reference

### LINKED ISSUE #37629: [MLv2] [FE] Migrate data reference
https://github.com/metabase/metabase/blob/8876f689af093fdd6f5d859bee5e834ffc3dac12/frontend/src/metabase/reference/utils.js#L60

We already have:
- `metadataProvider`
- `tableOrCardMetadata`
- `queryFromTableOrCardMetadata`
- `segmentMetadata`
- `metricMetadata`

And:
- `breakout(query, stageIndex, column)`
- `availableAggregationOperators` -> you can check for `shortName` to find `count`
- `aggregationClause`
- `aggregate`
- `filter`
- `withDefaultTemporalBucket`

To change the query. But we don't have a way to get `ColumnMetadata` by field id yet. We probably need to expose another function similar to `tableOrCardMetadata` for that.

kamilmielnik:
Closes #37629


frontend/src/metabase/reference/databases/TableDetail.jsx
```diff
@@ -39,6 +40,7 @@ const interestingQuestions = table => {
         dbId: table.db_id,
         tableId: table.id,
         getCount: true,
+        metadata,
```
kamilmielnik, line 43: All `getQuestionUrl` calls need `metadata` now

frontend/src/metabase-lib/aggregation.ts
```diff
@@ -35,6 +36,22 @@ export function aggregate(
   return ML.aggregate(query, stageIndex, clause);
 }
 
+export function aggregateByCount(query: Query): Query {
```
kamilmielnik, line 39: I decided to introduce this util in `metabase-lib` because it's already needed in 3 places

frontend/src/metabase-lib/metadata.ts
```diff
@@ -1,11 +1,7 @@
 import * as ML from "cljs/metabase.lib.js";
 import * as ML_MetadataCalculation from "cljs/metabase.lib.metadata.calculation";
-import type {
-  DatabaseId,
-  DatasetColumn,
-  Field,
-  TableId,
-} from "metabase-types/api";
+import type { DatabaseId, DatasetColumn, TableId } from "metabase-types/api";
+import type Field from "./metadata/Field";
```
kamilmielnik, line 4: Wrong `Field` type was being referenced here, I didn't realize it earlier because `Lib.fromLegacyColumn` was only used in JS files.

frontend/src/metabase/query_builder/actions/object-detail.js
```diff
@@ -108,17 +108,25 @@ export const loadObjectDetailFKReferences = createThunkAction(
       const queryResult = getFirstQueryResult(state);
 
       async function getFKCount(card, queryResult, fk) {
-        const fkQuery = Q_DEPRECATED.createQuery("query");
-
-        fkQuery.database = card.dataset_query.database;
-        fkQuery.query["source-table"] = fk.origin.table_id;
-        fkQuery.query.aggregation = ["count"];
-        fkQuery.query.filter = getFilterForFK(objectId, fk);
+        const metadata = getMetadata(getState());
```
kamilmielnik, line 111: This part should probably have been migrated in #37639, but I missed it earlier.

frontend/src/metabase-lib/binning.ts
```diff
@@ -18,22 +18,22 @@ export function isBinnable(
...
+): ColumnMetadata {
   return ML.with_binning(column, binningStrategy);
 }
 
 export function withDefaultBinning(
   query: Query,
   stageIndex: number,
   column: ColumnMetadata,
-) {
+): ColumnMetadata {
```
kamilmielnik, line 36: At a first glance these functions looked to me like they return `Query`, and I lost some time debugging this, so I defined the return types explicitly.

frontend/src/metabase/reference/utils.unit.spec.js
```diff
@@ -124,150 +148,142 @@ describe("Reference utils.js", () => {
...
         getCount: true,
+        metadata,
       });
 
       expect(question).toEqual(
         getNewQuestion({
-          database: 3,
-          table: 4,
-          aggregation: [["count"]],
+          aggregation: [["count", null]],
```
kamilmielnik, line 169: I think the `null` should not be there. Looks like a backend issue. I asked [a question in Slack](https://metaboat.slack.com/archives/C04CYTEL9N2/p1705408133622419).

frontend/src/metabase-lib/aggregation.ts
```diff
@@ -35,6 +36,22 @@ export function aggregate(
...
+  const countOperator = operators.find(operator => {
+    const info = displayInfo(query, stageIndex, operator);
+    return info.shortName === "count";
+  });
+
+  if (!countOperator) {
+    return query;
+  }
+
+  const aggregation = aggregationClause(countOperator);
```
uladzimirdev, line 51: idk do we need to use wrappers or MLv2 functions directly

frontend/src/metabase-lib/metadata.ts
uladzimirdev: other places use `import type Field from "metabase-lib/metadata/Field";`. I'd stick to such format of imports, but if you have any other opinion, please share

frontend/src/metabase/query_builder/actions/object-detail.js
```diff
@@ -107,18 +107,26 @@ export const loadObjectDetailFKReferences = createThunkAction(
...
-        const fkQuery = Q_DEPRECATED.createQuery("query");
-
-        fkQuery.database = card.dataset_query.database;
-        fkQuery.query["source-table"] = fk.origin.table_id;
-        fkQuery.query.aggregation = ["count"];
-        fkQuery.query.filter = getFilterForFK(objectId, fk);
+      async function getFKCount(card, fk) {
+        const metadata = getMetadata(getState());
+        const databaseId = new Question(card, metadata).databaseId();
+        const tableId = fk.origin.table_id;
```
uladzimirdev, line 113: I think I saw a comment from Alex re using `const tableId = question.tableId();`. IDK the difference, but using `fk.origin.table_id` looks a bit verbose. Or are those tableIds different?

frontend/src/metabase/query_builder/components/view/sidebars/SummarizeSidebar/SummarizeSidebar.tsx
```diff
@@ -163,17 +163,9 @@ function getQuery(query: Lib.Query, isDefaultAggregationRemoved: boolean) {
...
-    operator => {
-      const { shortName } = Lib.displayInfo(query, STAGE_INDEX, operator);
-      return shortName === "count";
-    },
-  );
-
-  if (operator && shouldAddDefaultAggregation) {
-    const clause = Lib.aggregationClause(operator);
-    return Lib.aggregate(query, STAGE_INDEX, clause);
+  if (!shouldAddDefaultAggregation) {
```
uladzimirdev, line 166: could you please explain why this change is introduced?

frontend/src/metabase/reference/databases/TableDetail.jsx
uladzimirdev: how did it work before? 🤯

uladzimirdev (approved):
🤞

frontend/src/metabase-lib/aggregation.ts
kamilmielnik: It's better to use wrappers because of the typing.
1 extra call on the stack should not be a problem :)

frontend/src/metabase/reference/databases/TableDetail.jsx
kamilmielnik: Previously the query was constructed as JSON with simple manipulations and hardcoded clauses (see diff in `frontend/src/metabase/reference/utils.js`). Metadata was only used for fields.

Now as everything needs to go through MLv2 query, metadata is always needed.

frontend/src/metabase-lib/metadata.ts
kamilmielnik: I tend to use relative imports where it's possible to do so without "going up" (`../`).
Here it's consistent with `Metadata` import in line below.

frontend/src/metabase/query_builder/actions/object-detail.js
kamilmielnik: They are different.

frontend/src/metabase/query_builder/components/view/sidebars/SummarizeSidebar/SummarizeSidebar.tsx
kamilmielnik: I reused the new `Lib.aggregateByCount`, the function should work the same as before.
I "inverted" the condition to have as little code in `if` block as possible.

frontend/src/metabase/reference/utils.unit.spec.js
kamilmielnik: [Reverted](https://github.com/metabase/metabase/pull/37694/commits/4509c48a34429f94317fd62224c379f966b780a9), [skipped the tests for now](https://github.com/metabase/metabase/pull/37694/commits/06fd8a707a1d39b21fb94e2f82e021c295bf8bbe) and filed a BE issue for it: #37782

PR #37697, opened 2024-01-15
Add linter to rachet down implicitly non-exported defsettings

crisptrutski:
### Description

This create a new clj-kondo linter to prevent adding new settings without explicitly specifying whether it should be included in the export. 

Refs #36240

### How to verify

Either add a new defsetting without an `:export?` value, or remove the pair from an existing setting, and run the linter.


.clj-kondo/hooks/metabase/models/setting.clj
```diff
@@ -50,15 +59,15 @@
...
-             form)))}))))
+      (:node
+        (defsetting
+          {:node
+           (hooks/parse-string
+             (with-out-str
+               ((resolve 'clojure.pprint/pprint)
+                form)))}))))
 
-  (defn- x []
```
crisptrutski, line 61: Adding the usual kondo comment (^1) doesn't seem to work inside a `(comment ...)` block.

1. #_{:clj-kondo/ignore [:unused-private-value]}

crisptrutski:
I'm not sure whether it's worth backporting this - for me it's unclear whether this will prevent or cause more merge conflicts for backports.

.clj-kondo/hooks/metabase/models/setting.clj
```diff
@@ -3,6 +3,207 @@
...
+                           (meta setting-name))
+                         (hooks/string-node "Docstring.")
+                         (hooks/vector-node [(hooks/token-node '_value-or-nil)]))
+                        hooks/list-node
+                        (with-meta (update (meta node) :clj-kondo/ignore #(hooks/vector-node (cons :clojure-lsp/unused-public-var (:children %))))))]
+
+    (when (nil? (second (drop-while (comp not #{[:k :export?]} first) options-list)))
+      (when-not (contains? ignored-implicit-export? (:value setting-name))
+        #_{:clj-kondo/ignore [:discouraged-var]}
+        (prn (:value setting-name))
```
crisptrutski, line 191: This is temporary, just useful to copy-and-paste things from the CI logs :-)

calherries:
@crisptrutski when you say you left excludes deliberately thin, do you mean `ignored-implicit-export?` deliberately excludes some settings that don't have an `:export?` key?

calherries:
And I think you can try backporting but if there's any merge conflicts just close it. It's not worth spending any time fixing merge conflicts because the chance this will prevent merge conflicts is very low

crisptrutski:
> @crisptrutski when you say you left excludes deliberately thin, do you mean `ignored-implicit-export?` deliberately excludes some settings that don't have an `:export?` key?

Oops, that comment is out of date. I've updated the list to be complete, probably even accidentally including some test settings, since CI did an exhaustive lint.

calherries:
I think I would have preferred a runtime check on `defsetting` (probably inside `register-setting!`) while adding `:export?` to the existing settings, but this is a fine approach and lets us get away with not having to update 150+ settings at once.

For clarity I would still add a default for `:export?` [here](https://github.com/metabase/metabase/blob/7673aed5bcec6cf6077cd4db13a5884c91111c88/src/metabase/models/setting.clj#L881-L901), make it a required key [here](https://github.com/metabase/metabase/blob/7673aed5bcec6cf6077cd4db13a5884c91111c88/src/metabase/models/setting.clj#L286), and add it to the docstring of the `defsetting` macro [here](https://github.com/metabase/metabase/blob/7673aed5bcec6cf6077cd4db13a5884c91111c88/src/metabase/models/setting.clj#L1019).

I think you still need to tidy up this temporary test too, left over from the previous PR https://github.com/metabase/metabase/blob/7673aed5bcec6cf6077cd4db13a5884c91111c88/src/metabase/models/setting.clj#L1019

Also, please link to the original issue in the PR description. Even though github makes bidirectional links, the description is where I look for the issue so I can navigate to other PRs in this series.

.clj-kondo/hooks/metabase/models/setting.clj
```diff
@@ -3,6 +3,205 @@
...
+                           (meta setting-name))
+                         (hooks/string-node "Docstring.")
+                         (hooks/vector-node [(hooks/token-node '_value-or-nil)]))
+                        hooks/list-node
+                        (with-meta (update (meta node) :clj-kondo/ignore #(hooks/vector-node (cons :clojure-lsp/unused-public-var (:children %))))))]
+
+    (when (nil? (second (drop-while (comp not #{[:k :export?]} first) options-list)))
+      (when-not (contains? ignored-implicit-export? (:value setting-name))
+        (hooks/reg-finding! (assoc (meta node)
+                                   :message "Setting definition must provide an explicit value for :export?"
```
calherries, line 191: If a new dev hasn't seen the `:export?` setting before, this might save them some confusion
```suggestion
                                   :message "Setting definition must provide an explicit value for :export? indicating whether the setting should be exported or not with serialization."
```

calherries (approved):

Issue #37017, opened 2023-12-20
Make metadata table list scrollable independently from the body


iethree:
> Just calling out some weirdness: when you have a bunch of tables and you want to change metadata, you have to scroll down to find the table, and when you select it, you have to scroll back up to view the fields.

![image](https://github.com/metabase/metabase/assets/116838/7a1d8110-91ef-4671-b9a6-30a023a55457)

[Slack Message](https://metaboat.slack.com/archives/C064EB1UE5P/p1703109964246029?thread_ts=1703109964.246029&cid=C064EB1UE5P)

iethree:
this is done!

PR #22150, opened 2022-04-27
Sign in form: "Remember me" checkbox should be left-aligned

ariya:
Fixes #16340.

### Before

image

### After

image

PR #37706, opened 2024-01-15
Improve admin page layout

### LINKED ISSUE #37529: Improve Admin Scrollbars
Our Scrollbars in our admin interface aren't great.

The biggest problem is that they don't scroll independently of page content, which makes them very difficult to use  when the page length and the menu length differs: which is often. This could be a fairly easy fix - just modify the design of the sidebar to be less overflowy and allow it to scroll independently from the page content pane.

Table Metadata with a lot of tables
tableList

Logs page
logspage

Table metadata with more columns than tables
tableMetadataSettings

related: 
https://github.com/metabase/metabase/issues/37502

sloansparger:
Closes https://github.com/metabase/metabase/issues/37529 https://github.com/metabase/metabase/issues/37017

### Description

This PR makes a handful of adjustments to how the admin pages are laid out, particularly the pages that contain a left side bar along with the content (more specifically this would be any of the pages under "Settings", "Table Metadata", "People", and "Troubleshooting"). Exhaustively, the changes inlucde:

1. (Main improvement) The left sidebar and the right content scroll separately from each other and maintain their own heights:

- Left sidebar content longer (video link)
https://github.com/metabase/metabase/assets/7104357/8fa7ec73-8292-4660-bb16-d6ec64764037

- Right main content longer (video link)
https://github.com/metabase/metabase/assets/7104357/1757439e-fd95-4ca0-9ebe-fdf5bee765fc

- Both sides overflow (video link)
https://github.com/metabase/metabase/assets/7104357/1424ffa6-ab5f-4555-8e92-27d4d36426d6

2. The second improvement is setting a fixed min-width on smaller devices for the main content section. While not ideal, it makes some of the settings pages useable on smaller devices for the first time:

https://github.com/metabase/metabase/assets/7104357/1292e334-cbff-468d-9814-d99d35e39bf3

3. The third improvement is making spacing and titles more consistent across all pages. This keeps the pages from jittering around as the user navigates through various settings pages.

- appearance settings
  - before 
Screenshot 2024-01-12 at 3 13 19 PM
  - after
settings-appearance-after

 - people settings
   - before 
people-before
  - after
people-after

### How to verify

- go to admin settings
- go to "Settings", "Table Metadata", "People", or "Troubleshooting" sections of the admin settings

### Checklist


iethree:
This is SO nice, and is a huge QoL improvement for longer pages.

It also looks like you've handled https://github.com/metabase/metabase/issues/37017 here 💪 

A few preliminary thoughts:
- definitely get @naatz-co's approval on this before merging
- can we move the right-side padding inside the scroll container? I think it looks weird to have a scroll bar not on the edge of the screen
Screen Shot 2024-01-16 at 10 45 55 AM
Screen Shot 2024-01-16 at 10 43 23 AM
- can we set a min-height on the left side nav bar? I think really small ones look weird. Maybe doesn't have to be full height, WDYT @naatz-co ?
Screen Shot 2024-01-16 at 10 45 55 AM

naatz-co:
"Can we move the right-side padding inside the scroll container? I think it looks weird to have a scroll bar not on the edge of the screen" agree with Ryan, it looks weird.

Overall, I don't have any comments! It looks so much nicer indeed!

naatz-co (approved):
"Can we move the right-side padding inside the scroll container? I think it looks weird to have a scroll bar not on the edge of the screen" agree with Ryan, it looks weird.

Overall, I don't have any comments! It looks so much nicer indeed!

"can we set a min-height on the left side nav bar? I think really small ones look weird. Maybe doesn't have to be full height, WDYT @naatz-co ?"

Yes! Something we could do that I think would help is follow the content height of the right side; Here's an example (screenshot). So then the left column height wouldn't look weird compared to the right side column and would bring visibility to the two options on the left side that seem to be hidden a bit with the current min-height.
1s

sloansparger:
Screenshot 2024-01-17 at 2 47 22 PM
Implemented a default of 50vh as it is probably not worth the effort it would take to make left / right sides match heights by default.

iethree (approved):
💖  This is the greatest thing to hit metabase since https://github.com/metabase/metabase/pull/22150

Issue #36368, opened 2023-12-04
[Epic] Browse models


luizarakaki:
**Links**
- [product doc on alignment for next steps](https://www.notion.so/metabase/Alignment-on-direction-for-Browse-Data-Models-v0-2-dbc94b4f013f4f59be309728f89bf143)
- [original product doc](https://www.notion.so/metabase/Bring-models-into-Browse-Data-and-make-it-easy-to-find-a-starting-point-a170b6815bc4405e98bff5032d870d87)
- branch: `models-in-browse-data` (#37707) and several downstream branches

```[tasklist]
- [x] Initial porting and tidying of files (https://github.com/metabase/metabase/pull/38107)
- [x] Semantic cleanup of e2e files (https://github.com/metabase/metabase/pull/38126)
- [x] Make two tabs in the browse data page (https://github.com/metabase/metabase/pull/37707)
- [x] Give the models tab an empty-list message ("No models here yet") (https://github.com/metabase/metabase/pull/37707)
- [x] Add one dummy model to the model tab, correctly styled (https://github.com/metabase/metabase/pull/37707)
- [x] Populate model tab with models from the backend (https://github.com/metabase/metabase/pull/37707)
- [x] Organize models by parent collection (https://github.com/metabase/metabase/pull/37707)
- [x] Add loading indicator to model list (https://github.com/metabase/metabase/pull/37707)
- [x] Add image to empty-list message (https://github.com/metabase/metabase/pull/37707)
- [x] Ensure that design is responsive (https://github.com/metabase/metabase/pull/37707)
- [x] Separate URLs per tab (#https://github.com/metabase/metabase/pull/37707)
- [x] Exclude models in personal collections
- [x] Ensure clicking the models goes to the right place (https://github.com/metabase/metabase/pull/37707)
- [x] Add tests
- [x] Use abbreviated relative time strings like 30min, 20h, 10d
- [x] Clicking on the model should take you to the model “explore” url
- [x] Make styles exactly right
- [x] Bring breadcrumbs inside Database tab
- [x] Add a white background on the Browse Data header, which will require changing the deep structure of the page (#37707)
- [x] Make the database box-shadow match the model box shadow
- [x] Remember last active tab (as described in product doc) (#38291)
- [x] Add verified badges to model cards
- [x] Fix model description overflow
- [x] Fix overwide collection header link
- [x] Make cards look more cards on collections (#38518)
- [x] Elevate Browse data link in the side nav (#38518)
[...]

PR #37707, opened 2024-01-15
Basic version of Models in Browse Data

rafpaf:
Part of https://github.com/metabase/metabase/issues/36368

### Description

This PR adds models to the Browse Data tab. It provides these sub-features:

* Make two tabs in the browse data page, Models and Databases
* Give the models tab an empty-list message ("No models here yet") 
* Populate the models tab with models from the backend 
* Organize models by their collection (the immediate parent in the collection hierarchy)
* Add loading indicator to model list
* Exclude models in personal collections

These sub-features will be addressed in follow-up PRs:
* Abbreviating the editor's last name
* Virtualizing the list of models
* Virtualizing the list of databases
* Opening the databases tab by default if the user cannot access any models
* Remembering the user's last active tab
* Internationalization
* Verified badges
* Dismissible banner explaining what models are

### How to verify

Click 'Browse Data' in the sidebar on the left. Click the two tabs.

### Checklist


frontend/src/metabase/browse/components/BrowseApp.jsx
```diff
@@ -2,5 +2,12 @@
 import { BrowseAppRoot } from "./BrowseApp.styled";
 
 export default function BrowseApp({ children }) {
-  return <BrowseAppRoot data-testid="browse-data">{children}</BrowseAppRoot>;
+  return (
+    <BrowseAppRoot
+      data-testid="browse-data"
+      style={{ display: "flex", height: '100%', }}
```
rafpaf, line 8: I'll eventually switch to using styled components

frontend/src/metabase/admin/permissions/components/FilterableTree/FilterableTree.tsx
rafpaf, line 1: Small changes needed because I ported `LastEditInfoLabel.jsx` to typescript

frontend/src/metabase/admin/permissions/components/FilterableTree/FilterableTree.tsx
```diff
@@ -23,7 +24,7 @@ interface FilterableTreeProps {
   itemGroups: ITreeNodeItem[][];
   emptyState?: {
     text: string;
-    icon: string;
+    icon: IconName;
```
rafpaf, line 27: Small change needed because I ported LastEditInfoLabel.jsx to typescript

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const time = <time dateTime={lastEdit.timestamp}>{formattedDate}</time>;
+  const tooltipLabel = jt`Last edited by ${lastEditorName}${(<br />)}${time}`;
+
+  return (
+    <Link
+      aria-labelledby={`${collectionHtmlId} ${headingId}`}
+      key={model.id}
+      style={style}
+      to={Urls.model(model as unknown as Partial<Card>)}
+      // FIXME: Not sure that 'Model Click' is right; this is modeled on the database grid which has 'Database Click'
```
rafpaf, line 100: I've pinged Kyle about this

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  const time = <time dateTime={lastEdit.timestamp}>{formattedDate}</time>;
+  const tooltipLabel = jt`Last edited by ${lastEditorName}${(<br />)}${time}`;
+
+  return (
+    <Link
+      aria-labelledby={`${collectionHtmlId} ${headingId}`}
+      key={model.id}
+      style={style}
+      to={Urls.model(model as unknown as Partial<Card>)}
```
rafpaf, line 99: Please correct me if I'm wrong, but I believe some fudging with the types like this is needed, because a model is represented as a `SearchResult` or `CollectionItem` by some portions of the codebase (since models are retrieved via `/api/search`), and as a `Card` in other portions.

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+      // FIXME: Not sure that 'Model Click' is right; this is modeled on the database grid which has 'Database Click'
+      data-metabase-event={`${ANALYTICS_CONTEXT};Model Click`}
```
rafpaf, line 101: I've pinged Kyle about this 'Model Click' question

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+    >
+      <ModelCard>
+        <h4 className="text-wrap" style={{ lineHeight: "16px" }}>
+          <MultilineEllipsified id={headingId}>
```
rafpaf, line 105: The `id` supports `aria-labelledby`, which supports semantic unit testing

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+        </Text>
+        <LastEditInfoLabel
+          item={modelWithHistory}
+          fullName={modelWithHistory["last-edit-info"].full_name}
+          className={"last-edit-info-label-button"}
+          tooltipProps={{
+            label: tooltipLabel,
+            withArrow: true,
+          }}
+          // TODO: Use first name and last initial
```
rafpaf, line 125: There is an open question about whether the "Connor D" format makes sense, so for now we are using the last editor's full name.

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+ * thirdly by model name, and
+ * lastly by model id. */
+const sortModels = (a: SearchResult, b: SearchResult) => {
+  const fallbackSortValue = Number.MAX_SAFE_INTEGER;
+
+  // Sort first on the name of the model's parent collection, case insensitive
+  const collectionNameA =
+    a.collection?.name?.toLowerCase() || fallbackSortValue;
+  const collectionNameB =
+    b.collection?.name?.toLowerCase() || fallbackSortValue;
```
rafpaf, line 149: I don't believe it's possible in practice for collections to lack names, but Typescript doesn't require them to have names, so we need to provide a fallback value in case the `model.collection` or `model.collection.name` is falsy.

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  return 0;
+};
+
+const CollectionHeader = ({
+  collection,
+  style,
+  id,
+}: {
+  collection?: Pick<Collection, "id" | "name"> | null;
```
rafpaf, line 199: `Urls.collection` expects a `Collection`, but only a `Partial<Collection>` is available as the `collection` property of a model found through `/api/search`, so we need this weird intermediate type.

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+    const collectionIdChanged =
+      models[i - 1]?.collection?.id !== model.collection?.id;
+
+    const firstModelInItsCollection =
+      i === 0 || collectionIdChanged || model.collection?.id === undefined;
+
+    const collectionHtmlId = model?.collection?.id
+      ? `collection-${model.collection?.id}`
+      : `item-${cells.length}`;
```
rafpaf, line 239: This `id` supports `aria` attributes which enable semantic testing

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,278 @@
...
+    cells.push(
+      <ModelCell
+        collectionHtmlId={collectionHtmlId}
+        key={`model-${model.id}`}
+        model={model}
+      />,
+    );
+  }
+  return cells;
+};
```
rafpaf, line 264: We could also create this list of `<ModelCells>` in a loop inside some jsx, but this approach makes it much easier to drop in virtualization in [the next PR](https://github.com/metabase/metabase/pull/37818)

frontend/src/metabase/common/hooks/use-search-list-query/use-search-list-query.ts
```diff
@@ -10,9 +10,12 @@ import type {
...
+export const useSearchListQuery = <
+  TItem = CollectionItem,
+  TResult = Omit<SearchResults, "data">,
+>(
   props: UseEntityListQueryProps<SearchListQuery> = {},
-): UseEntityListQueryResult<CollectionItem, Omit<SearchResults, "data">> => {
+): UseEntityListQueryResult<TItem, TResult> => {
   return useEntityListQuery(props, {
     fetchList: Search.actions.fetchList,
     getList: Search.selectors.getList,
```
rafpaf, line 21: Per my [conversation on Slack](https://metaboat.slack.com/archives/C505ZNNH4/p1705603270151099) with @kulyk, I've modified `useSearchListQuery` so that it optionally takes a type. For example, if you do `const { data } = useSearchListQuery<SearchResult>()` then `data` will have the type `SearchResult[]`. If you do `const { data } = useSearchListQuery()` (a pattern found elsewhere in the codebase), `data` will have the type `CollectionItem[]` by default.

frontend/src/metabase/components/EmptyState/EmptyState.jsx
rafpaf, line 1: ported to tsx

frontend/src/metabase/browse/containers/DatabaseBrowser.jsx
rafpaf, line 1: I replaced this with `BrowseData.tsx` which lets you browse not just databases, as before, but now models as well

frontend/src/metabase/components/LastEditInfoLabel/LastEditInfoLabel.jsx
rafpaf, line 1: ported to tsx

frontend/src/metabase/components/EmptyState/EmptyState.tsx
rafpaf, line 1: The lines here that are actually new in the PR are labeled below

frontend/src/metabase/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,107 @@
...
+  message?: React.ReactNode;
+  title?: React.ReactNode;
+  action?: React.ReactNode;
+  link?: string;
+  illustrationElement?: React.ReactNode;
+  onActionClick?: () => void;
+  className?: string;
+  icon?: IconName;
+  image?: string;
+};
```
rafpaf, line 50: I replaced `PropTypes` with typed props

frontend/src/metabase/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,107 @@
...
+}: EmptyStateProps) => (
+  <div className={className}>
+    <EmptyStateHeader>
+      {illustrationElement && (
+        <EmptyStateIllustration className="empty-state-illustration">
+          {illustrationElement}
+        </EmptyStateIllustration>
+      )}
+      <div>
+        {isValidIconName(icon) && <LegacyIcon icon={icon} {...rest} />}
```
rafpaf, line 72: I added a more rigorous check of the icon name, which typescript needs

It turns out that there is at least one icon with the name set to "all", which is not a valid icon name
https://github.com/metabase/metabase/blob/ea7f5b7e443cd4f1cd6a725999c82b53bef87022/frontend/src/metabase/reference/metrics/MetricQuestions.jsx#L37

frontend/src/metabase/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,107 @@
...
+        {image && <LegacyImage image={image} {...rest} />}
+      </div>
+      {title && (
+        <h2 role="status" aria-live="polite" className="text-medium">
+          {title}
+        </h2>
```
rafpaf, line 78: I added the `role` and `aria-live` attributes for a11y. `role` also facilitates semantic testing

frontend/src/metabase/components/LastEditInfoLabel/LastEditInfoLabel.tsx
rafpaf, line 1: I ported and refactored this file

frontend/src/metabase/query_builder/components/DataSelector/saved-question-picker/SavedQuestionList.tsx
```diff
@@ -29,7 +29,7 @@ const SavedQuestionList = ({
 }: SavedQuestionListProps): JSX.Element => {
   const emptyState = (
     <SavedQuestionListEmptyState>
-      <EmptyState message={t`Nothing here`} icon="all" />
+      <EmptyState message={t`Nothing here`} />
```
rafpaf, line 32: Now that EmptyState is a typescript file, it won't tolerate this invalid icon name

e2e/test/scenarios/admin/databases/default-sample-database.cy.spec.js
rafpaf, line 1: Though there are small changes to other files, the core files in this PR are:

`frontend/src/metabase/browse/containers/BrowseData.tsx`
`frontend/src/metabase/browse/containers/BrowseModels.tsx`
`frontend/src/metabase/browse/containers/BrowseData.unit.spec.tsx`
`frontend/src/metabase/browse/containers/BrowseData.styled.tsx`

rafpaf:
I've put this back into draft mode because there are some lingering styling issues

iethree (dismissed):
🏆 💎 🏆  This is _really_ good work. You incrementally improved a bunch of the codebase while building a solid new feature without going too far down rabbit holes. Really excellent.

Before merging let's:
- Make sure browsing schema and tables keep us in the same browseDataPage view
- Do another pass on the css: generally, we want to write as few styled components as possible, and keep them as small as possible. I think we can get rid of many of these with some mantine components and style props.
- Chat in more detail about ways to restructure BrowseModels to be simpler.
- Add an e2e test or two. Your unit tests are good, but let's make sure we can get all the way through to a model via the browser.

e2e/test/scenarios/onboarding/urls.cy.spec.js
```diff
@@ -26,6 +26,7 @@ describe("URLs", () => {
   describe("browse databases", () => {
     it(`should slugify database name when opening it from /browse"`, () => {
       cy.visit("/browse");
+      cy.findByRole("tab", { name: "Databases" }).click();
```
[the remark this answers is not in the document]
rafpaf: Totally. I will do this in a follow-up PR

frontend/src/metabase/browse/containers/BrowseData.styled.tsx
```diff
@@ -0,0 +1,159 @@
...
+  h2 {
+    margin-bottom: 0.5rem;
+  }
+  .empty-state-illustration {
+    margin-bottom: 0.5rem;
+  }
+  .empty-state-message {
+    max-width: 24rem;
+  }
+`;
```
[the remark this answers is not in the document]
rafpaf: I tried removing these rules:

```
  display: flex;
  flex: 1;
  flex-direction: column;
  align-items: center;
  justify-content: center;
  width: 100%;
  height: 100%;
```

I think the only one I can remove without changing how things look is `width: 100%`

frontend/src/metabase/browse/containers/BrowseData.styled.tsx
```diff
@@ -0,0 +1,159 @@
...
+  align-items: center;
+  padding-top: 1rem;
+  margin-right: 1rem;
+  a {
+    display: flex;
+    flex-flow: row nowrap;
+    &:hover h4 {
+      color: ${color("brand")};
+    }
+  }
```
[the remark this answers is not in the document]
rafpaf: Is there a way to do this by styling the link directly?

```
    &:hover h4 {
      color: ${color("brand")};
    }
```

frontend/src/metabase/browse/containers/BrowseData.tsx
```diff
@@ -0,0 +1,129 @@
...
+} from "./BrowseData.styled";
+import { BrowseModels } from "./BrowseModels";
+
+interface BrowseDataTab {
+  label: string;
+  component: JSX.Element;
+}
+
+export const BrowseDataPage = () => {
+  const [currentTabId, setTabId] = useState<string | null>("models");
```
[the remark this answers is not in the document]
rafpaf: I thought about doing something like this but it makes `onTabChange` complain here:

```
      <BrowseTabs value={currentTabId} onTabChange={setTabId}>
```

frontend/src/metabase/browse/containers/BrowseData.styled.tsx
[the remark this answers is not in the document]
rafpaf: Nevermind, I'm using a styled component now

frontend/src/metabase/browse/containers/BrowseModels.tsx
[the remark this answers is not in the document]
rafpaf: That code should have just been in the virtualization PR. I'll simplify here, as you suggest

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,282 @@
...
+          <MultilineEllipsified id={headingId}>
+            {model.name}
+          </MultilineEllipsified>
+        </h4>
+        <Text size="xs" style={{ height: "32px" }}>
+          <MultilineEllipsified
+            tooltipMaxWidth="100%"
+            className={model.description ? "" : "text-light"}
+          >
+            {model.description || "No description."}{" "}
```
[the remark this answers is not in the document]
rafpaf: This is in Figma fwiw

image

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,282 @@
...
+            label: tooltipLabel,
+            withArrow: true,
+          }}
+          // TODO: Use first name and last initial
+        >
+          {lastEditorName}
+          {lastEditorName && howLongAgo ? (
+            <LastEditedInfoSeparator>•</LastEditedInfoSeparator>
+          ) : null}
+          {howLongAgo}
```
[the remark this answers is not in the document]
rafpaf: I agree. I'll make it a separate component

frontend/src/metabase/components/LastEditInfoLabel/LastEditInfoLabel.tsx
```diff
@@ -0,0 +1,112 @@
...
+  tooltipProps?: TooltipProps;
+  children?: React.ReactNode;
+}) {
+  const lastEditInfo = item["last-edit-info"];
+
+  const editorId = lastEditInfo?.id;
+  const timestamp = lastEditInfo?.timestamp;
+  const timeLabel = timestamp ? getHowLongAgo(timestamp) : "";
+
+  fullName ||= formatEditorName(lastEditInfo) || null;
```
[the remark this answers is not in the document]
rafpaf: It looks like a duck to me

frontend/src/metabase/routes.jsx
```diff
@@ -217,7 +217,7 @@ export const getRoutes = store => {
           </Route>
 
           <Route path="browse" component={BrowseApp}>
-            <IndexRoute component={DatabaseBrowser} />
+            <IndexRoute component={BrowseDataPage} />
             <Route path=":slug" component={SchemaBrowser} />
             <Route path=":dbId/schema/:schemaName" component={TableBrowser} />
```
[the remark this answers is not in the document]
rafpaf: Definitely. I'm planning to do both of these steps - assigning each tab its own URL, and ensuring that the Browse data header appears on `SchemaBrowser` and `TableBrowser` - in a separate PR, since this one is already big.

frontend/src/metabase/browse/containers/BrowseModels.tsx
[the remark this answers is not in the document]
iethree: WDYT @kdoh is this a figma-lies situation? or a ryans-data-lies situation?

Screen Shot 2024-01-22 at 2 18 10 PM

frontend/src/metabase/routes.jsx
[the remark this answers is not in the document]
iethree: I think the URL thing is fine to do separately, but the database click removing the tabs feels like a bug that shouldn't go into master to me.

rafpaf:
* `master`
  * **PR #37707** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37707?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #37818** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #37923** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #38019** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38019?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37707?utm_source=stack-comment).

e2e/test/scenarios/onboarding/urls.cy.spec.js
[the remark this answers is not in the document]
rafpaf: #38020

frontend/src/metabase/browse/containers/BrowseData.styled.tsx
```diff
@@ -0,0 +1,159 @@
...
+export const ModelCard = styled(Card)`
+  padding: 1.5rem;
+  padding-bottom: 1rem;
+
+  &:hover {
+    color: ${color("brand")};
+  }
+  .Button:hover {
+    color: ${color("text-medium")} !important;
+  }
```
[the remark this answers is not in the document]
rafpaf: I'm planning to resolve this by not longer using LastEditInfoLabel; you're right that I was overloading that component.

frontend/src/metabase/browse/containers/BrowseData.styled.tsx
```diff
@@ -0,0 +1,159 @@
...
+  align-items: flex-start;
+
+  box-shadow: 0px 1px 4px 0px rgba(0, 0, 0, 0.06) !important;
+
+  & .last-edit-info-label-button div,
+  & .last-edit-info-label-button span {
+    white-space: wrap !important;
+    text-align: left;
+    font-weight: normal;
+  }
```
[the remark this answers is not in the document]
rafpaf: As you suggested, I'll switch to using a separate component where I can have more control

frontend/src/metabase/browse/containers/BrowseModels.tsx
```diff
@@ -0,0 +1,282 @@
...
+      </ModelCard>
+    </Link>
+  );
+};
+
+/** Sort models firstly by collection name,
+ * secondly by collection id,
+ * thirdly by model name, and
+ * lastly by model id. */
+const sortModels = (a: SearchResult, b: SearchResult) => {
```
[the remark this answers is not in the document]
rafpaf: I want to ask you more about 1 sometime.

Re 2 and 3, I've taken your suggestions, thank you!

frontend/src/metabase/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,107 @@
...
+      )}
+      {message && (
+        <Text role="status" color="medium" className="empty-state-message">
+          {message}
+        </Text>
+      )}
+    </EmptyStateHeader>
+    {/* TODO - we should make this children or some other more flexible way to
+      add actions
+      */}
```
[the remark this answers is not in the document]
rafpaf: I can't take credit for this comment - it was in the file I ported! 😁

e2e/test/scenarios/admin/databases/database-exceptions.cy.spec.js
```diff
@@ -91,7 +91,8 @@ describe("scenarios > admin > databases > exceptions", () => {
     cy.visit("/admin/databases");
     cy.wait("@failedGet");
 
-    cy.findByRole("heading", { name: "Something's gone wrong" });
+    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
+    cy.findByText("Something's gone wrong");
```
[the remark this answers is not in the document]
rafpaf: Good idea. I will do this in a separate PR that cleans up problems like this.

rafpaf:
@iethree 

> Before merging let's:
> 
> * Make sure browsing schema and tables keep us in the same browseDataPage view

I found it easier to do this in conjunction with assigning routes to each tab, so this logic is a work in progress, in #38020.

> * Do another pass on the css: generally, we want to write as few styled components as possible, and keep them as small as possible. I think we can get rid of many of these with some mantine components and style props.

Done!

> * Chat in more detail about ways to restructure BrowseModels to be simpler.

Looking forward to that.

> * Add an e2e test or two. Your unit tests are good, but let's make sure we can get all the way through to a model via the browser.

Added [here](https://github.com/metabase/metabase/blob/models-in-browse-data/e2e/test/scenarios/onboarding/home/browse.cy.spec.js)

iethree:
comment unrelated to a specific code change. in the database browser, the "our data" breadcrumb should probably say "databases" now, and clicking it should return you to `/browse/databases` not `browse`

frontend/src/metabase/browse/components/BrowseDatabases.tsx
```diff
@@ -0,0 +1,64 @@
...
+import type { useDatabaseListQuery } from "metabase/common/hooks";
+
+import NoResults from "assets/img/no_results.svg";
+import { DatabaseCard, DatabaseGridItem } from "./BrowseDatabases.styled";
+import { CenteredEmptyState } from "./BrowseApp.styled";
+
+export type BrowseTabId = "models" | "databases";
+
+export const isValidBrowseTab = (value: string | null): value is BrowseTabId =>
+  value === "models" || value === "databases";
```
iethree, line 21: this should be in a utils function since it's used in multiple files

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+export const groupModels = (
+  models: SearchResult[],
+  locale: string | undefined,
+) => {
+  const groupedModels = _.groupBy(models, model => model.collection.id);
+  let collections = models.map(model => model.collection);
+  collections = _.uniq(collections, collection => collection.id) || [];
+  collections.sort((a, b) => a.name?.localeCompare(b.name, locale));
+  return { groupedModels, collections };
```
iethree, line 38: `let` is often a red flag for me. I think this would be simpler if we just sorted the grouped models and returned them as an array, then we wouldn't need to return collections at all.

```suggestion
  const groupedModels = Object.values(_.groupBy(models, model => model.collection.id))
    .sort((a, b) => a[0].collection?.name?.localeCompare(b[0].collection?.name, locale));
  return groupedModels;
```

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -0,0 +1,192 @@
...
+};
+
+export const BrowseModels = ({
+  modelsResult,
+}: {
+  modelsResult: ReturnType<typeof useSearchListQuery<SearchResult>>;
+}) => {
+  const { data: models = emptyArray, error, isLoading } = modelsResult;
```
iethree, line 46: creating a new variable for an empty array shouldn't be necessary.

```suggestion
  const { data: models = [], error, isLoading } = modelsResult;
```

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const sortedModels = models.sort(sortModels);
+  const collection = models[0].collection;
+
+  /** This id is used by aria-labelledby */
+  const collectionHtmlId = `collection-${collection?.id ?? `index-${index}`}`;
+
+  return (
+    <>
+      <CollectionHeader
+        collection={models[0].collection}
```
iethree, line 105: I think you probably just want to use `collection` that you extracted already here

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -0,0 +1,192 @@
...
+      </Group>
+    ) : (
+      <>{children}</>
+    );
+  return (
+    <CollectionHeaderContainer id={id} role="heading" style={style}>
+      <MaybeLink>
+        <Group spacing=".33rem">
+          <Icon name="folder" color={"text-dark"} size={16} />
+          <Text>{collection?.name || "Untitled collection"}</Text>
```
iethree, line 187: Here, we need to handle "Our Analytics" with the `metabase/collections/utils` > `isRootCollection()`


Screen Shot 2024-01-25 at 1 44 45 PM
Screen Shot 2024-01-25 at 1 44 35 PM

frontend/src/metabase/browse/components/BrowseModels.unit.spec.tsx
iethree, line 1: let's add a test here to make sure that "Our analytics" renders when we get a collection id of `null`

frontend/src/metabase/browse/utils.ts
```diff
@@ -0,0 +1,47 @@
+import type { SearchResult } from "metabase-types/api";
+
+export type SortableModel = Partial<
+  Pick<SearchResult, "id" | "name" | "collection">
+>;
+
+/** Sort models firstly by collection name (case insensitive), secondly by collection id, thirdly by model name (case insensitive). */
+export const sortModels = (a: SortableModel, b: SortableModel) => {
```
iethree, line 8: since we're now grouping by collection, we don't need to sort by collection anymore, right?

frontend/src/metabase/routes.jsx
```diff
@@ -219,10 +218,35 @@ export const getRoutes = store => {
...
+            <Route
+              path="databases/:dbId/schema/:schemaName"
+              component={({ params }) => (
+                <BrowseApp tab="databases">
+                  <TableBrowser params={params} />
+                </BrowseApp>
+              )}
+            />
+
+            <Redirect from=":dbId-:slug" to="databases/:dbId-:slug" />
```
iethree, line 245: I think a comment here would be helpful to let future developers know that this is to support legacy paths in versions 48 and earlier.

iethree (changes requested):
Looking good. just a few more little things.

frontend/src/metabase/browse/components/BrowseModels.tsx
rafpaf: Fixed

Btw, n.b.: https://github.com/metabase/metabase/blob/models-in-browse-data/frontend/src/metabase/query_builder/components/view/QuestionDataSource.jsx#L130

frontend/src/metabase/browse/components/BrowseModels.tsx
rafpaf: Good catch. That usage of `emptyArray` mattered when there was a `useEffect` here that depended on `models`.

iethree (approved):
jonah-hill-excited

Issue #36524, opened 2023-12-07
[Epic] Let creators make dashboard and native query filters required


NevRA:
[Product doc](https://www.notion.so/metabase/Let-creators-make-dashboard-filters-required-7ea3d7caf63545d2b713d1509e4267ad)
[Testing plan](https://github.com/metabase/metabase/issues/37520)

Follow-ups: [Required parameters: follow-ups ](https://github.com/metabase/metabase/issues/38603)

## Issues to close
- https://github.com/metabase/metabase/issues/18097
- https://github.com/metabase/metabase/issues/5320

## Milestones

```[tasklist]
### 1. Native queries
- [ ] https://github.com/metabase/metabase/pull/37800
- [ ] https://github.com/metabase/metabase/pull/37709
- [ ] https://github.com/metabase/metabase/pull/37858
- [ ] https://github.com/metabase/metabase/pull/37696
- [ ] https://github.com/metabase/metabase/pull/38006
- [ ] https://github.com/metabase/metabase/pull/38232
```
---
The two below are in fact a single milestone but I've split them out in two groups.
```[tasklist]
### 2. Dashboards & Embeddings
- [ ] https://github.com/metabase/metabase/pull/38050
- [ ] https://github.com/metabase/metabase/pull/38177
- [ ] https://github.com/metabase/metabase/pull/38203
- [ ] https://github.com/metabase/metabase/pull/38327
- [ ] https://github.com/metabase/metabase/pull/38395
```

oleggromov:
Everything is merged.

JeroenNelen:
Could you add an 'first value alphabetically' option for the 'default value' that becomes required after making the filter required?
Because if we want to use this in a multi-tenant setup where the filter is used for tenant-specific data
e.g. a setup where our database contains multiple schools (tenants) so we have a filter on tenant_ID but we also have a filter on 'student' and we want the student filter to be required, however when making it required it demands a default value but this default value is related to the specific tenant/school so we would need a generic option like 'first one (alphabetically)' because school A might have a student named Anna and school B will not have that student, but it could use the first one being Brian in that case (as an example).

uladzimirdev:
@JeroenNelen this task is completed. If you have any suggestions or improvements in mind, please create a new detailed issue and we'll prioritize it

PR #37709, opened 2024-01-15
Required parameters: add to native queries

oleggromov:

## How it works

- template tags can have the required attribute, so I [added the same for parameters](https://github.com/metabase/metabase/pull/37800) (within a card)
- whenever a parameter is required, we require default value in the variable settings sidebar to save the question — this is done in `NativeQuery._validateTemplateTags` and supported in presentational components here and there
- when a parameter is required, we show parameter widgets with an asterisk and allow resetting to the default value 

### Behavior details

1. Toggling "required" doesn't change the default value anymore
2. When we enable required toggle and the default value is set and the parameter widget is empty, we update it
3. When required is already enabled but there's no default value, entering or picking it updates the parameter widget to have it
4. When the value is required, the "X" in the input only appears when you type — it allows you to clear the input
5. If you remove focus from an empty parameter value input, it would reset to the default value

## How to test

1. Create a native question `select * from people where email = {{ email }}`
2. Change parameter to be required
3. Add a default value
4. Check that it is used, that you can reset back to it, that you can't save without it


e2e/test/scenarios/native-filters/reproductions/11480.cy.spec.js
```diff
@@ -32,11 +32,10 @@ describe("issue 11480", () => {
...
-    SQLFilter.runQuery();
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.contains(
-      "You'll need to pick a value for 'X' before this query can run.",
-    );
+    // We cannot run the query unless the default value for a required parameter is set.
+    cy.findByTestId("native-query-editor-container")
+      .icon("play")
+      .parents("button")
+      .should("be.disabled");
```
oleggromov, line 39: @nemanjaglumac please have a look. In this PR, we're making default value mandatory for required parameters in native queries. This means, we can no longer attempt to run a query without a default value set — instead, I am checking if the run button is disabled.

e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
```diff
@@ -53,16 +53,6 @@ describe("scenarios > filters > sql filters > field filter", () => {
         cy.findByText("Small Marble Shoes");
         cy.findByText("Rustic Paper Wallet");
       });
-
-      cy.log("set the value through the filter widget");
```
oleggromov, line 57: @nemanjaglumac  Same here, this checked if a required parameter with unset default value could be set through the filter widget — but now it's impossible, and I am removing this check.

NB: I haven't added new e2e tests covering the new behavior yet.

e2e/test/scenarios/native-filters/reproductions/11480.cy.spec.js
nemanjaglumac: I'd maybe narrow this down to the native query editor sidebar.

Even better, we can try getting this element directly by its aria-label.

frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
```diff
@@ -92,6 +95,21 @@ class ParameterValueWidget extends Component {
       return null;
     }
 
+    const canReset =
```
kamilmielnik, line 98: This function would be easier to read with the props destructured.

kamilmielnik (commented):
Nice work :+1: 

Left a few small comments inline.
The blocking one is marked with :x: .

nemanjaglumac (commented):
This behavior seems to be wrong, but I have to check with you first.
So, when you clear the input value, the toggle automatically turns off the required filter check. Was it specced like this, or just an accident?

https://github.com/metabase/metabase/assets/31325167/14634454-4d4a-4a1e-993c-9f2c0e8d1d79

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -347,45 +350,51 @@ class TagEditorParamInner extends Component<Props> {
...
-              }
-              value={tag.default}
-              setValue={value => {
-                this.setParameterAttribute("default", value);
-                this.props.setParameterValue(tag.id, value);
-              }}
-              isEditing
-              commitImmediately
+          <ToggleContainer>
+            <Toggle
```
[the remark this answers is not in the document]
oleggromov: I actually started this milestone from trying to refactor it into Mantine first. It went well, transition from Toggle to Switch is easy among the other things — but the field mapping component is complex. It is definitely a separate undertaking, which I have planned but won't start until the core feature is done. 
Screenshot 2024-01-23 at 09 37 13


So I decided to proceed with just reordering existing components in this PR. Also I am not sure that using a single Mantine component makes anything any better.

oleggromov:
> when you clear the input value, the toggle automatically turns off the required filter check. Was it specced like this, or just an accident?

@nemanjaglumac it definitely looks like a leftover from the previous (incomplete) required parameter logic. I'll take a look and fix it most likely. Thanks for the catch!

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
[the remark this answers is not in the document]
kamilmielnik: Fair enough - I didn't notice it's already been there, I thought it's a new usage.

frontend/src/metabase/query_builder/actions/core/core.js
```diff
@@ -305,6 +306,22 @@ export const setParameterValue = createAction(
   },
 );
 
+export const SET_DEFAULT_PARAMETER_VALUE =
+  "metabase/qb/SET_DEFAULT_PARAMETER_VALUE";
+export const setDefaultParameterValue = createThunkAction(
```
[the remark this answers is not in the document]
oleggromov: This is a good observation, thank you.

frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
oleggromov: IMO, it's a matter of taste — I am not a big fan of unnecessary repetition. So won't change.

frontend/src/metabase/components/FieldSet.tsx
```diff
@@ -20,6 +29,7 @@ export default function FieldSet({
           className="h5 text-bold text-uppercase px1 text-nowrap text-medium"
         >
           {legend}
+          {required && <span>&nbsp;*</span>}
```
uladzimirdev, line 32: 😄 afaik span cannot be moved to another line anyway, so just ` ` could work here

uladzimirdev:
I see two e2e test groups still fail, also one fix wasn't pushed seems.
Otherwise LGTM

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -347,45 +350,51 @@ class TagEditorParamInner extends Component<Props> {
...
         <InputContainer lessBottomPadding>
-          <ContainerLabel>{t`Required?`}</ContainerLabel>
-          <Toggle
-            id="tag-editor-required"
-            value={tag.required}
-            onChange={value => this.setRequired(value)}
+          <ContainerLabel>
+            {t`Default filter widget value`}
+            {hasNoDefaultValue && tag.required && (
+              <ErrorSpan>{t`(required)`}</ErrorSpan>
```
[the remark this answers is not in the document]
oleggromov: Yeah, I fixed that — https://github.com/metabase/metabase/pull/37709/commits/f8ca9dc42d1549c286f1539db8ff0b1c510f9a6d#diff-1b10cd34fe6aed27d19e61d5ea323bee48ba7142da466004511a33af636c53b4R286

But probably during a rebase it was lost :( Fixed again now, thanks.

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -346,46 +349,52 @@ class TagEditorParamInner extends Component<Props> {
           </InputContainer>
         )}
 
-        <InputContainer lessBottomPadding>
-          <ContainerLabel>{t`Required?`}</ContainerLabel>
-          <Toggle
-            id="tag-editor-required"
-            value={tag.required}
-            onChange={value => this.setRequired(value)}
+        <div>
```
oleggromov, line 352: @nemanjaglumac now that it's wrapped in a `div`, not a label, the bug with automatic unchecking the "required" switch is fixed.

frontend/src/metabase/components/FieldSet.tsx
oleggromov: I am used to agressive formatting, which puts spaces in `{" "}` so they'er not lost with line wrapping — hence the `&nbsp`. We don't have this formatting rules yet (and hopefully won't have ever), but I don't like the version in curlies and just a plain space doesn't communicate the intention well enough in my opinion.

frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
oleggromov: Sorry guys, not gonna change that 😄

uladzimirdev (approved):

oleggromov:
@cdeweyx and @dhuniverse please have a look and let me know if you like everything.
The only missing thing is that the reload icon is pointing backwards, but this is the only icon we have in the current set — and I am not sure it makes any difference to use the one pointing in the other direction.

cdeweyx (approved):

nemanjaglumac:
I'm still trying out various edge cases, but the main difference seems to be that we show the default value input field even when the required condition is not checked.

Before this change you'd have to first make a filter as required, and only then would the default input field appear.

Was this change intentional?
@oleggromov @cdeweyx

nemanjaglumac:
This behavior is slightly confusing to me. I understand the reasoning behind it, though. But it's still worth mentioning.
If you start by (1) making field as required, (2) don't set the default value but (3) set the filter value, you still can't preview the query.
https://github.com/metabase/metabase/assets/31325167/7f88ec11-5743-415c-80e5-d39925b0dfec

nemanjaglumac (approved):
I was mostly checking how this works in the app, because you already received some good feedback on the code.

Things checked:
- [optional clauses](https://www.metabase.com/docs/latest/questions/native-editor/sql-parameters#sql-variable-types)
- multiple variables
- mongo
Apart from a few comments I left, everything seems to be working well.

oleggromov:
> I'm still trying out various edge cases, but the main difference seems to be that we show the default value input field even when the required condition is not checked.
> Was this change intentional?

Yes, it is intentional — otherwise it was confusing.

oleggromov:
> This behavior is slightly confusing to me. I understand the reasoning behind it, though. But it's still worth mentioning.
If you start by (1) making field as required, (2) don't set the default value but (3) set the filter value, you still can't preview the query.

This is an interesting observation, thank you. Do you suggest that absence of a default value should only prevent us from _saving_ the question whereas running it should depend on either the "preview" parameter value or the default?

nemanjaglumac:
> Do you suggest that absence of a default value should only prevent us from _saving_ the question whereas running it should depend on either the "preview" parameter value or the default?

Well, that was certainly my instinctive expectation. It wasn't even fully conscious at first. I was just confused and then tried to rationalize it. But this can be highly subjective.

Just my 2c that shouldn't affect "mergeability" of this PR.

oleggromov:
> Well, that was certainly my instinctive expectation. It wasn't even fully conscious at first. I was just confused and then tried to rationalize it. But this can be highly subjective.

Thank you, I agree with that. To keep things simple, I think I'll merge this PR when I can and make a follow-up PR to change this behavior.

nemanjaglumac:
> and make a follow-up PR to change this behavior

Tbh, not even sure it's worth it. I'd check with the product and design teams first.

frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
oleggromov: Actually this ended up changed. So you win!

PR #37724, opened 2024-01-15
Add back auto-following logs on Admin Logs page

sloansparger:
### Description

Previously a layout change caused a regression which broke a feature on the logs page with tailed the logs as new logs came in. This feature also stopped following the logs if they user scrolled.

This PR add this feature back, and does so in a more modern react fashion.

### How to verify

- go to admin settings
- go to troubleshooting
- go to logs
- you should see the page auto-scrolled to bottom
- validate auto-follow
  - in a new tab make various requests (view dashboards, edits questions, etc.)
  - see that logs are still scrolled to bottom
- validate auto-follow pause
 - scroll up from the bottom in the logs (this should stop the auto-scrolling feature keeping you exactly where you were)
 - in a new tab make various requests (view dashboards, edits questions, etc.)
 - see that logs are in the same location / logs didn't scroll

### Demo

https://github.com/metabase/metabase/assets/7104357/abfe9c26-ab29-45d4-ab12-b377e2c95f7f

### Checklist


iethree (approved):
🥳 Lovely 👏 

(for future reference, we really prefer functional components, so turning this into a functional component would have been a good move, but also no harm in keeping this rarely used component as a class component 👍 )

PR #37725, opened 2024-01-15
Allow non-admins to see unarchive/delete buttons in archive list

### LINKED ISSUE #26169: Discrepency in the unarchive UI between admins and non-admins
**Describe the bug**
As an admin, if you go to the list of archived content, you can hover over any single entry and see an "Unarchive this" button. You can also select content by clicking its icon and then use "Unarchive" and "Delete" buttons that appear at the bottom of the screen.

As a non-admin, the buttons don't appear on hover, but you can still unarchive things from the bottom of the screen. There's no reason the buttons on hover should be hidden if we're allowing the functionality regardless, so this is a weird discrepancy.

Master

**Severity**
Very minor

sloansparger:
Closes https://github.com/metabase/metabase/issues/26169

### Description

Removes a check that only allowed admins to see unarchive or delete buttons in archive list. These are actions they are allowed to take and could do in another places in the UI, so this limitation was arbitrary.

### How to verify

- log in as non-admin
- archive a question
- go to archive page
- hover over an element in the list

### Demo

https://github.com/metabase/metabase/assets/7104357/4bb5fe2c-63bd-432a-a8a8-8aadd9b8e2f9

### Checklist

### DIFF
--- a/frontend/src/metabase/archive/containers/ArchiveApp.tsx
+++ b/frontend/src/metabase/archive/containers/ArchiveApp.tsx
@@ -12,7 +12,6 @@ import { isSmallScreen, getMainElement } from "metabase/lib/dom";
 
 import { openNavbar } from "metabase/redux/app";
 import { getIsNavbarOpen } from "metabase/selectors/app";
-import { getUserIsAdmin } from "metabase/selectors/user";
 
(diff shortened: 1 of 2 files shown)


iethree (approved):
Nice! Can you add a unit test as well?

sloansparger:
> Nice! Can you add a unit test as well?

We talked offline and decided it didn't make sense to add a test here.

PR #37727, opened 2024-01-16
[MLv2] Make `query=` treat missing `:base-type` as matching

### LINKED ISSUE #37659: [MLv2] [BE] [Bug] Make `query=` ignore absense of `base-type` and `effective-type` in field references
Blocks https://github.com/metabase/metabase/issues/37508

If one of the queries was generated by the legacy query, it might not have `base-type` and `effective-type` in some field references. To make it work, we need to ignore absence of these properties. Example: `base-type` is not set in one clause but set in another and apart from that other properties are equal, then these clauses are equal. Same for `effective-type`.

This issue breaks certain QB logic; the most important one is allowing users without data permissions to view the query when coming from dashboards.

bshepherdson:
Make `query=` treat missing `:base-type` and `:effective-type` as matching
any `:base-type` or `:effective-type` on the other side.

A mismatch is still considered not-equal.

Fixes #37659.

### DIFF
--- a/src/metabase/lib/js.cljs
+++ b/src/metabase/lib/js.cljs
@@ -340,6 +340,41 @@
                                     ;; match up. Therefore de-dupe with `frequencies` rather than simply `set`.
                                     (assoc inner-query :fields (frequencies fields)))))))
 
+(defn- compare-legacy-field-refs
+  [[key1 id1 opts1]
+   [key2 id2 opts2]]
+  ;; A mismatch of `:base-type` or `:effective-type` when both x and y have values for it is a failure.
+  ;; If either ref does not have the `:base-type` or `:effective-type` set, that key is ignored.
+  (letfn [(clean-opts [o1 o2]
+            (not-empty
+              (cond-> o1
+                (not (:base-type o2))      (dissoc :base-type)
+                (not (:effective-type o2)) (dissoc :effective-type))))]
+    (= [key1 id1 (clean-opts opts1 opts2)]
+       [key2 id2 (clean-opts opts2 opts1)])))
+
+(defn- query=* [x y]
+  (cond
+    (and (vector? x)
+         (vector? y)
+         (= (first x) (first y) :field))
+    (compare-legacy-field-refs x y)
+
+    ;; Otherwise this is a duplicate of clojure.core/=.
+    (and (map? x) (map? y))
+    (and (= (set (keys x)) (set (keys y)))
+         (every? (fn [[k v]]
+                   (query=* v (get y k)))
+                 x))
+
+    (and (sequential? x) (sequential? y))
+    (and (= (count x) (count y))
+         (every? true? (map query=* x y)))
+
+    ;; Either mismatched map/sequence/nil/etc., or primitives like strings.
+    ;; Either way, = does the right thing.
+    :else (= x y)))
+
 (defn ^:export query=
   "Returns whether the provided queries should be considered equal.
 
@@ -354,7 +389,7 @@
   ([query1 query2 field-ids]
    (let [n1 (prep-query-for-equals query1 field-ids)
          n2 (prep-query-for-equals query2 field-ids)]
-     (= n1 n2))))
+     (query=* n1 n2))))
 
 (defn ^:export group-columns
   "Given a group of columns returned by a function like [[metabase.lib.js/orderable-columns]], group the columns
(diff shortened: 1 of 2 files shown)


bshepherdson:
* `master`
  * **PR #37727** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37727?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37727?utm_source=stack-comment).

src/metabase/lib/js.cljs
```diff
@@ -339,6 +339,41 @@
...
+    (and (= (set (keys x)) (set (keys y)))
+         (every? (fn [[k v]]
+                   (query=* v (get y k)))
+                 x))
+
+    (and (sequential? x) (sequential? y))
+    (and (= (count x) (count y))
+         (every? identity (map query=* x y)))
+
+    ;; Either mismatched map/sequence/nil/etc., or primitives like strings.
```
metamben, line 373: Even if we don't currently have field refs in sets, I would feel better if we recursed into sets too.

src/metabase/lib/js.cljs
```diff
@@ -339,6 +339,41 @@
...
+                (not (:base-type o2))      (dissoc :base-type)
+                (not (:effective-type o2)) (dissoc :effective-type))))]
+    (= [key1 id1 (clean-opts opts1 opts2)]
+       [key2 id2 (clean-opts opts2 opts1)])))
+
+(defn- query=* [x y]
+  (cond
+    (and (vector? x)
+         (vector? y)
+         (= (first x) (first y) :field))
```
metamben, line 359: What about checking that the options element is missing or a map? `compare-legacy-field-refs` blows up if it tries to `dissoc` from a number or string, for example.

src/metabase/lib/js.cljs
```diff
@@ -339,6 +339,41 @@
...
+    ;; Otherwise this is a duplicate of clojure.core/=.
+    (and (map? x) (map? y))
+    (and (= (set (keys x)) (set (keys y)))
+         (every? (fn [[k v]]
+                   (query=* v (get y k)))
+                 x))
+
+    (and (sequential? x) (sequential? y))
+    (and (= (count x) (count y))
+         (every? identity (map query=* x y)))
```
metamben, line 371: `query=*` should return a boolean and we could emphasize this by 
```suggestion
         (every? true? (map query=* x y)))
```

metamben (approved):
I have a few suggestions, but LGTM overall.

src/metabase/lib/js.cljs
bshepherdson: Garbage in, garbage out. Sending a query _that_ badly malformed may as well throw; returning true or false is worse at that point. It works correctly on maps and `nil`/missing.

src/metabase/lib/js.cljs
bshepherdson: It's a pain to compare sets in this way. Since these are legacy queries and we don't use sets in them anywhere so far as I know, I've just added code to throw if sets are encountered.

bshepherdson: Actually I tried to write a test for that and it doesn't really work. This is consuming legacy queries, and `js->clj` doesn't convert JS `Set`s into CLJS sets (it treats them as opaque JS objects).

src/metabase/lib/js.cljs
metamben: Fair enough, if you know those points. (If I see such a generic function I prefer not to blow up, as I usually don't know if the part that's causing the problem is essential at all.)

bshepherdson: Do you feel the same way now that it's a much more targeted `valid-filter-for?` rather than `compatible-type?` or `assignable?`?

src/metabase/lib/js.cljs
metamben: Maybe just a comment about the need to revisit this if we ever want to use it for non-legacy queries.

src/metabase/lib/js.cljs
metamben: Sorry, I'm lost. How is this related to `valid-filter-for?`?

PR #37736, opened 2024-01-16
HK: remove noPopover and unnecessary function from ParameterValueWidget

oleggromov:
This logic was a leftover from the time when `Widget` wasn't used. Now this component implements the Widget type selection logic and `getWidgetDefinition` is no longer needed.


uladzimirdev (approved):

kamilmielnik (approved):

uladzimirdev (approved):
:shipit:

Issue #36966, opened 2023-12-20
Help white-labeling admins remove links to metabase.com outside of admin settings


WiNloSt:
**Links**
- [product doc](https://www.notion.so/metabase/Help-white-labeling-admins-remove-any-references-to-Metabase-outside-of-admin-settings-f5ddbca732ff400ab858d9e173562b0b?d=fe1fe3f2da854da8aa4260b81a37d19e#1793a7b1b07b427684a46d9413ff0856)
- [Technical design doc](https://www.notion.so/metabase/Tech-Remove-Metabase-references-outside-of-admin-settings-f88ff0fa8f574d6393c6fe5a15911b0a?d=b6c059f19e7e44d58c440a5181a4fb47#c600388e76734c9d9c2af302004b4b3f)
- [Testing plan](https://www.notion.so/metabase/WIP-Testing-Plan-Remove-Metabase-references-outside-of-admin-settings-f3c254867cf44fb3952ab4c9e677e6cf)
- feature branch: `tbd` _this should be the feature branch where this work will be done in. PRs will be delivered against this branch_

```[tasklist]
### Milestone 1
- [x] List all places where we want to remove Metabase links outside admin settings https://github.com/metabase/metabase/commit/43e0574148/
- [ ] https://github.com/metabase/metabase/pull/37076
```

```[tasklist]
### Milestone 2
- [ ] https://github.com/metabase/metabase/pull/37738
- [x] Analytics (done in https://github.com/metabase/metabase/issues/37656)
- [ ] [optional] implement an abstraction that deals with removing Metabase links
```

WiNloSt:
I'll close the epic since Milestone 2 doesn't change any behavior in the application. It just changes the developer experience.

PR #37738, opened 2024-01-16
ESLint for detecting the use of Metabase links

WiNloSt:
Milestone 2 of #36966

### Description

Tested with `time yarn lint-eslint-pure` both `master` and this PR yielded similar results over a couple of tries ~130s on my M1 Macbook pro.

Using `TIME=all yarn lint-eslint-pure` will show that this rule only took ~30ms and it accounts for 0.0% of the total time linting

Rule                                                   | Time (ms) | Relative
:------------------------------------------------------|----------:|--------:
no-unconditional-metabase-links-render                 |    33.337 |     0.0%

This adds a rule to detect that when dealing with Metabase links we should either
1. conditionally render the links or
2. disable the rule with a reason

So we don't shoot ourselves in the foot in the future after this feature has been fully shipped. Linting is necessary because we will definitely forget about this feature in no time, and this will prevent the mishandling with Metabase links in the future.

#### Types of links that this rule detects
 1. `MetabaseSettings.learnUrl(string)`
 2. `MetabaseSettings.docsUrl(string)`
 3. `getDocsUrl` selector from "metabase/selectors/settings"
 4. `getLearnUrl` selector from "metabase/selectors/settings"
 5. inline string "metabase.com/docs/"
 6. inline string "metabase.com/learn/"

### Changes
1. The ESLint rule
2. The rule's test
3. Other files are the result of ESLint working, and those are places we don't want to conditionally render the link with the specified reason.

### How to verify

1. Go to any existing `.jsx` or `.tsx` file, or even create a new one if you may.
4. Try to add any type of Metabase links from above
5. You should see the ESLint error with a message telling you how to fix it.

### Demo

- image
- image
- image
- image
- image
- image
- image
- image

### Checklist


.eslintrc
```diff
@@ -158,6 +158,18 @@
     }
   },
   "overrides": [
+    {
+      "files": ["*.jsx", "*.tsx"],
+      "rules": {
+        "no-unconditional-metabase-links-render": "error"
```
WiNloSt, line 164: This rule should only test JSX components because we expect to check if React components render the Metabase links conditionally or not. We don't care about links created outside of React component files.

.eslintrc
```diff
@@ -158,6 +158,18 @@
...
+      }
+    },
+    {
+      "files": ["*.unit.spec.jsx", "*.unit.spec.tsx"],
```
WiNloSt, line 168: Some unit tests will contain the assertions for Metabase learn and docs links, so we need to disable it.

frontend/lint/tests/no-unconditional-metabase-links-render.unit.spec.js
WiNloSt, line 1: All cases are performed according to the ESLint rule spec
https://github.com/metabase/metabase/blob/f992c45dcae991d90704c249c7c3fec49d9ab0d7/frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js#L5-L15

frontend/lint/tests/no-unscoped-text-selectors.unit.spec.js
```diff
@@ -1,6 +1,5 @@
-/* eslint-disable import/no-commonjs */
-const RuleTester = require("eslint").RuleTester;
-const rule = require("./no-unscoped-text-selectors");
+import { RuleTester } from "eslint";
```
WiNloSt, line 1: Convert to module import so it's more consistent with other tests. The reason it wasn't working in the first place is when we tried to change this file and husky would complain.

You could try to run `eslint --rulesdir frontend/lint/eslint-rules --max-warnings 0 --fix` and it seems like we shouldn't include the test files inside a directory passed to `rulesdir`. The [eslint repo](https://github.com/eslint/eslint) also have their tests in a different directory than the actual code, so I follow that convention.

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * @typedef DefaultImport
+     * @property {true} isDefault
+     * @property {string} source
+     *
+     * @typedef NamedImport
+     * @property {string} named
+     * @property {string} source
+     *
+     * @param {object} node
```
WiNloSt, line 55: I wanted to use `ASTNode` like in the ESLint repo, but I wasn't sure how to import it. e.g. 

https://github.com/eslint/eslint/blob/57089cb5166acf8b8bdba8a8dbeb0a129f841478/lib/rules/newline-after-var.js#L75

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,223 @@
...
+            variable.defs[0].node.type === "ImportSpecifier" &&
+            variable.name === named,
+        );
+      }
+
+      return false;
+    }
+
+    return {
+      ImportDeclaration(node) {
```
WiNloSt, line 79: Traversing the import statements e.g. `import xxxx from yyyy`

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,223 @@
...
+        if (
+          checkIfModuleIsImported(node, {
+            named: "getShowMetabaseLinks",
+            source: "metabase/selectors/whitelabel",
+          })
+        ) {
+          isGetShowMetabaseLinksSelectorImported = true;
+        }
+      },
+      CallExpression(node) {
```
WiNloSt, line 113: Traverse any call expression e.g. `myFunction()`, the good part is that our selectors are expected to always be called. `useSelector(state => getDocsUrl(state, { page: xxx })`. If they were to be passed directly e.g. `useSelector(getDocsUrl)`, it likely won't be traversed (I didn't test).

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,223 @@
...
+          node?.callee?.object?.name === "MetabaseSettings" &&
+          ["learnUrl", "docsUrl"].includes(node?.callee?.property?.name)
+        ) {
+          context.report({
+            node,
+            message: ERROR_MESSAGE,
+          });
+        }
+      },
+      Literal(node) {
```
WiNloSt, line 154: Check literal strings

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,223 @@
...
+          LITERAL_METABASE_URL_REGEX.exec(node.value) &&
+          !isGetShowMetabaseLinksSelectorImported
+        ) {
+          context.report({
+            node,
+            message: ERROR_MESSAGE,
+          });
+        }
+      },
+      TemplateLiteral(node) {
```
WiNloSt, line 169: Check literal strings in template tags.

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,223 @@
...
+        const quasis = node.quasis;
```
WiNloSt, line 170: You can think of `quasis` as a parts of a template tag e.g.
```ts
const tag = `This is the first part ${1+1} this is the second part`
```

`quasis` would contain 2 elements
1. one with a value `This is the first part ` inside
2. and another with a value ` this is the second part` inside.

It seems the interpolations aren't included in `quasis`

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,214 @@
...
+          /eslint-disable-next-line\s+no-unconditional-metabase-links-render/;
+        const ALLOWED_ESLINT_DISABLE_LINE_REGEX =
+          /eslint-disable-next-line\s+no-unconditional-metabase-links-render -- \w+/;
+        comments.forEach(comment => {
+          if (ESLINT_DISABLE_BLOCK_REGEX.exec(comment.value)) {
+            const { start, end } = comment.loc;
+            context.report({
+              loc: {
+                start: { line: start.line - 1, column: start.column },
+                end: { line: end.line - 1, column: end.column },
```
WiNloSt, line 195: It seems that disabling using block comment will prevent the error on the actual comment line. That makes sense because the rule is disabled _starting from_ the actual comment line.

So the hack would be to display the error on the line before. This works even if the block comment is the first line of the file.

The line comment works because it only disables the rule in the next line as the comment suggested `eslint-disable-next-line`

.eslintrc
```diff
@@ -158,6 +158,18 @@
...
   },
   "overrides": [
+    {
+      "files": ["*.js", "*.jsx", "*.ts", "*.tsx"],
+      "rules": {
+        "no-unconditional-metabase-links-render": "error"
+      }
+    },
+    {
+      "files": ["*.unit.spec.*"],
```
WiNloSt, line 168: A lot of unit tests file contains literal metabase links in assertions, so I need to disable it.

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -0,0 +1,217 @@
...
+            node,
+            message: ERROR_MESSAGE,
+          });
+        }
+
+        // call `MetabaseSettings.learnUrl` or `MetabaseSettings.docsUrl`
+        if (
+          metabaseSettings?.references.some(
+            reference => reference.identifier === node?.callee?.object,
+          ) &&
```
WiNloSt, line 147: This is where we check that the default import from `metabase/lib/settings` is used to call `learnUrl()` or `docsUrl()`.

Previously I only checked if it's named `MetabaseSettings`, but the default import could be named anything, so it's better to check the reference instead. e.g. This is where we call it `Settings` instead of `MetabaseSettings`'
https://github.com/metabase/metabase/blob/e865ab9594ecd59fa4affb420fb3b9f25acaf4b8/frontend/src/metabase/admin/settings/slack/components/SlackStatus/SlackStatus.tsx#L45

enterprise/frontend/src/metabase-enterprise/advanced_permissions/components/ImpersonationModal/ImpersonationModalView.tsx
```diff
@@ -77,6 +77,7 @@ export const ImpersonationModalView = ({
         {t`When the person runs a query (including native queries), Metabase will impersonate the privileges of the database role you associate with the user attribute.`}{" "}
         <ExternalLink
           className="link"
+          // eslint-disable-next-line no-unconditional-metabase-links-render -- Admin settings
```
WiNloSt, line 80: This is the example where we want to add a reason why we should disable the rule.

frontend/lint/tests/no-unconditional-metabase-links-render.unit.spec.js
```diff
@@ -0,0 +1,185 @@
+import { RuleTester } from "eslint";
```
WiNloSt, line 1: All cases are performed according to the ESLint rule spec
https://github.com/metabase/metabase/blob/f992c45dcae991d90704c249c7c3fec49d9ab0d7/frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js#L5-L15

deniskaber (approved):

frontend/src/metabase/admin/databases/components/DatabaseEditApp/Sidebar/ModelCachingControl/ModelCachingControl.tsx
```diff
@@ -40,6 +40,7 @@ function FeatureDescription({ schemaName }: { schemaName: string }) {
   const docsLink = (
     <ExternalLink
       key="model-caching-link"
+      // eslint-disable-next-line no-unconditional-metabase-links-render -- Admin settings
```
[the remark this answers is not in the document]
WiNloSt: Good point. I'll disable the rule for anything under `/admin` too.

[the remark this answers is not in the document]
WiNloSt: Fixed in https://github.com/metabase/metabase/pull/37738/commits/338c0874d5d760561746c2ee00cd06c15e2a94ed and removed the commented from files under `/admin/` folder in https://github.com/metabase/metabase/pull/37738/commits/3741d993c8118bc6d61ef3c774d2482c0525cb46

PR #37737, opened 2024-01-16
backported "Serdes for DashboardCardSeries" to 47

calherries:
Backports https://github.com/metabase/metabase/pull/37271 to 47 

Closes https://github.com/metabase/metabase/issues/37232

PR #37747, opened 2024-01-16
Fix e2e test for serdes dashcardseries

calherries:
This PR fixes an e2e test that I added in https://github.com/metabase/metabase/pull/37271

The test currently passes only accidentally, relying on the order in which objects are created in the database based on serialization export and import implementations.

I discovered this when I tried to backport it to 47 on this PR https://github.com/metabase/metabase/pull/37737, where the test broke. I've already fixed it on that PR, so this only needs backporting to 48

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
@@ -642,9 +642,18 @@
            :model/Card {c3-id :id :as c3} {:name "Series Question B", :collection_id coll-id}
            :model/Dashboard {dash-id :id :as dash} {:name "Shared Dashboard", :collection_id coll-id}
            :model/DashboardCard {dc1-id :id} {:card_id c1-id, :dashboard_id dash-id}
-           :model/DashboardCard _dc2 {:card_id c2-id, :dashboard_id dash-id}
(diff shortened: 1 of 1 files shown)


piranha:
Hm... can you point out why it was dependent on the order, I fail to follow the logic :) I see how it breaks in the tests before, but I can't understand why it's better than it was 😁

enterprise/backend/test/metabase_enterprise/serialization/v2/e2e_test.clj
```diff
@@ -654,19 +663,25 @@
                 (is (serdes/with-cache (serdes.load/load-metabase! (ingest/ingest-yaml dump-dir)))
                     "successful"))
               (testing "Series are loaded correctly"
-                (let [new-dc1-id (t2/select-one-pk :model/DashboardCard :card_id (:id c1))
-                      new-dc2-id (t2/select-one-pk :model/DashboardCard :card_id (:id c2))
```
calherries, line 658: @piranha The problem was that these two lines use the `id` of `c1` and `c2`. But serialization doesn't guarantee stable ids.

piranha: Ahhh I see!

sansolovyov (approved):

PR #37749, opened 2024-01-16
Fix SQL server default to be empty again

### LINKED ISSUE #37655: 0.48.3 : SQL Server dynamic ports
### Describe the bug

We have a SQL Server using dynamic ports.
Since upgrade to 0.48.3, our sql server database connection fails. It tries to connect on port 1433 while the port parameter is not set (it should then use dynamic ports).
If I set the port on the actual dynamic port, connection is OK.

### To Reproduce

Set a sql server connection using dynamic ports (port field left empty)
Connection is refused.
In the logs, we can see that metabase tried to connect on 1433 port.

### Expected behavior

_No response_

### Logs

Jan 12 18:03:31 metabase metabase: "Échec de la connexion TCP/IP à l'hôte SQL-SERVER-01, port 1433. Erreur : « Connexion refusée (Connection refused). Vérifiez les propriétés de connexion. Assurez-vous qu'une instance de SQL Server est en cours d'exécution sur l'hôte et accepte les connexions TCP/IP au port. Vérifiez que les connexions TCP au port ne sont pas bloquées par un pare-feu. »."}

```JSON
CentOS 7 / metabase 0.48.3 / mysql 8.0.35
```

### Severity

normal

### Additional context

_No response_

qwef:
Closes https://github.com/metabase/metabase/issues/37655

### DIFF
--- a/modules/drivers/sqlserver/resources/metabase-plugin.yaml
+++ b/modules/drivers/sqlserver/resources/metabase-plugin.yaml
@@ -11,8 +11,7 @@ driver:
     - host
     - merge:
         - port
-        - default: 1433
-          description: Leave empty to use Dynamic Ports, or input specific port. Standard port is 1433.
+        - description: Leave empty to use Dynamic Ports, or input specific port. Standard port is 1433, which needs to be inputted.
     - merge:
         - dbname
         - name: db
--- a/modules/drivers/sqlserver/resources/metabase-plugin.yaml
+++ b/modules/drivers/sqlserver/resources/metabase-plugin.yaml
@@ -11,7 +11,7 @@ driver:
     - host
     - merge:
         - port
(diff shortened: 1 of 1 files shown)


modules/drivers/sqlserver/resources/metabase-plugin.yaml
```diff
@@ -11,8 +11,7 @@ driver:
     - host
     - merge:
         - port
-        - default: 1433
-          description: Leave empty to use Dynamic Ports, or input specific port. Standard port is 1433.
+        - description: Leave empty to use Dynamic Ports, or input specific port. Standard port is 1433, which needs to be inputted.
```
noahmoss, line 14: `inputted` sounds a little weird to me here. I think the previous wording was actually fairly clear.

noahmoss (approved):

PR #37703, opened 2024-01-15
Convert top-level dashboard components to TypeScript

kulyk:
Converts top-level dashboard components to TypeScript. That should make other dashboard components a bit more reliable as TypeScript would know which props and redux state exist in top-level dashboard components.

Would be easier to review commit-by-commit:

- [convert `DashboardApp` to TypeScript](https://github.com/metabase/metabase/commit/ea30674e9fc43ddd113f809fae707f903a92248c)
- [convert `Dashboard` to TypeScript](https://github.com/metabase/metabase/commit/a4449f6d867910b3c801645d27746cd5e009c5d3)
- [convert `DashboardGrid` to TypeScript](https://github.com/metabase/metabase/commit/3ce5cc3a323add349cfd25b5a71a470976cea773)
- [convert `DashboardHeader` to TypeScript](https://github.com/metabase/metabase/commit/c4254d42fe123851acc3f9710cd7ae637333c906)

**To verify:** CI should be green 🟢

PR #37757, opened 2024-01-16
Turn `Dashboard` into a functional component

kulyk:
Preparation #37711, based on https://github.com/metabase/metabase/pull/37703

Turns the `Dashboard` component into a functional one, no functional changes are expected.
Should make it more straightforward to filter and hide certain dashboard filters in scope of #37711

**To verify:** CI is green 🟢

### DIFF
--- a/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
+++ b/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
@@ -1,6 +1,6 @@
 import type { ReactNode } from "react";
 import { useCallback, useEffect, useMemo, useState } from "react";
-import { useMount, usePrevious, useUnmount } from "react-use";
+import { usePrevious, useUnmount } from "react-use";
 import type { Route } from "react-router";
 import _ from "underscore";
 import type { Location } from "history";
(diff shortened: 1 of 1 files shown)


frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
```diff
@@ -173,167 +174,230 @@ interface DashboardProps {
...
+    setSharing(!isSharing);
+  }, [isSharing, setSharing]);
+
+  const handleLoadDashboard = useCallback(
+    async (dashboardId: DashboardId) => {
+      initialize({ clearCache: !isNavigatingBackToDashboard });
+
+      loadDashboardParams();
+
+      const result: any = await fetchDashboard({
```
kamilmielnik, line 285: I think we should be able to type `payload` in `fetchDashboard`'s result (`{ dashboard: IDashboard }`).

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
```diff
@@ -173,167 +174,230 @@ interface DashboardProps {
...
+    setEditingParameter,
+    setErrorPage,
+    setParameterIndex,
+    setParameterValue,
+    setSharing,
+    toggleSidebar,
+  } = props;
+
+  const [isInitialized, setIsInitialized] = useState(false);
+  const [error, setError] = useState(null);
```
kamilmielnik, line 217:
```suggestion
  const [error, setError] = useState<unknown>(null);
```

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
```diff
@@ -173,167 +174,230 @@ interface DashboardProps {
...
+    [parameters],
+  );
+
+  const tabHasCards = useMemo(() => {
+    if (!Array.isArray(dashboard?.dashcards)) {
+      return false;
     }
+    const tabDashCards = dashboard.dashcards.filter(
+      dc => dc.dashboard_tab_id === selectedTabId,
+    );
```
kamilmielnik, line 236: Previously it also checked that `selectedTabId` is not `undefined`. And `dc.dashboard_tab_id` can be `undefined`.

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
```diff
@@ -173,167 +174,230 @@ interface DashboardProps {
...
+  useEffect(() => {
+    if (isInitialized && !_.isEqual(previousParameterValues, parameterValues)) {
+      fetchDashboardCardData({ reload: false, clearCache: true });
+    }
+  }, [
+    fetchDashboardCardData,
+    isInitialized,
+    parameterValues,
+    previousParameterValues,
+  ]);
```
kamilmielnik, line 394: :x: 

`componentDidUpdate` got split into 3 `useEffect` hooks.
Notice that previously there were `return` statements in `componentDidUpdate` - we need to migrate this logic as well. Otherwise we will get unnecessary network requests.

So: 2nd hook should not run if the 1st runs, and 3rd one should not run if 1st or 2nd runs.

I suggest to merge these 3 hooks into 1, which would look identical to the old `componentDidUpdate`.

We might be able to get rid of `useMount` hook as well by incorporating it here. This should help avoid unnecessary network requests.

kamilmielnik (changes requested):

kulyk:
Once again, thanks for a brilliant review @kamilmielnik 
I've addressed your comments, please take another look 🙏

kamilmielnik (approved):
Nice :fire:

PR #37759, opened 2024-01-16
[MLv2] Add matrix tests for filter functions

snoe:
Addresses the test matrix in the filters epic #34112

The idea is that  test-util.matrix can be used for producing a column of a known of various origins. If this proves useful it can be expanded.


bshepherdson (approved):
Approving this. I'm not thrilled with the ergonomics of the matrix util, but I want to experiment with reworking this test into property-based style.

PR #37760, opened 2024-01-16
[MLv2] Fix column-group display-name

### LINKED ISSUE #37369: [MLv2] [BE] Make displayInfo for column groups straightforward to use
Currently the FE has to use a utility function everywhere when displaying `ColumnGroup`:

```
export function getColumnGroupName(groupInfo: ColumnGroupDisplayInfo) {
  return groupInfo.fkReferenceName || groupInfo.displayName;
}
```

This is not straightforward. We never use `fkReferenceName` or its own, and we never use `displayName` for `ColumnGroup` outside of this function.  

We should improve the API - only `displayName` should be used without additional checks on the FE. MLv2 should update `displayName` for `ColumnGroup` to be the result of this function under the hood.

Without this check when using raw `displayName` now:
Screenshot 2024-01-09 at 14 07 32

With this check:
Screenshot 2024-01-09 at 14 08 10

snoe:
Fixes #37369

Column group display names need to take fk-reference-name into account and fk-reference-name no longer needs to be exposed through display-info.

Also add an implementation of display-name for column-group.

### DIFF
--- a/src/metabase/lib/column_group.cljc
+++ b/src/metabase/lib/column_group.cljc
@@ -101,7 +101,7 @@
        ;; This is very intentional: one table might have several FKs to one foreign table, each with different
        ;; meaning (eg. ORDERS.customer_id vs. ORDERS.supplier_id both linking to a PEOPLE table).
        ;; See #30109 for more details.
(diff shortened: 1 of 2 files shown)


metamben (approved):
👍

PR #37761, opened 2024-01-16
docs - mysql version note

jeff-bruemmer:
Adds notes about MySQL version.


docs/databases/connecting.md
```diff
@@ -26,8 +26,8 @@ The databases listed below have official drivers maintained by the Metabase team
 - [Amazon Athena](./connections/athena.md)
 - [BigQuery](./connections/bigquery.md) (Google Cloud Platform)
 - [Druid](./connections/druid.md)
-- [MongoDB (version 4.2 or higher)](./connections/mongodb.md)
-- [MySQL (version 5.7 or higher, as well as MariaDB version 10.2 or higher)](./connections/mysql.md)
+- [MongoDB (recommend version 4.2 or higher)](./connections/mongodb.md)
+- [MySQL (recommend version 8.0.33 or higher, as well as MariaDB version 10.4 or higher)](./connections/mysql.md)
```
paoliniluis, line 30: I would say simply 8+ here

metamben: `8.0.33` is the version we currently test against. I'm reluctant to recommend something we haven't tested at all.
 I assume we don't want to see bug reports for versions older than that, do we?

That said, `8.0.33` might change, because the version is not explicitly specified in the tests, which is unfortunate.

metamben (approved):

PR #37764, opened 2024-01-16
Exclude :card from the parameter types which require defaults to be considered fully parameterized 

### LINKED ISSUE #25022: Pinning a native question that references another question will fail to render
**Describe the bug**
When creating a native question that references another question (no matter if native or GUI), the question won't render and will show a "open this question and fill in its variables to see it". If you open the question, everything is fine.

**Logs**
No logs on the BE, as there's no error

**To Reproduce**
1) create a gui question (like new->question->orders->visualize->save it as orders). It should have ID 1
2) create a native question (`select * from {{#1}} as pepe`). Save it
3) Pin the second (native) question to the collection. See it fail when it shouldn't, doesn't have any params

**Expected behavior**
Only fail when there are questions that need params, or when there are questions based on questions that need params and don't have a default param

**Screenshots**
NA

- Your databases: Postgres
- Metabase version: 44.1
- Metabase internal database: Postgres

**Severity**
P2

**Additional context**
Reported by a customer

noahmoss:
Fixes https://github.com/metabase/metabase/issues/25022

* Exclude :card from the parameter types which require defaults to be considered fully parameterized to resolve https://github.com/metabase/metabase/issues/25022
* Fix typo across the code base: `parametrized` -> `parameterized`

**Product question:**

We don't want to attempt to render cards that need user input (parameters, required filters, etc). There's no hope of running a query that needs user choices.

In the past, the mechanism that determined if a query was parameterized considered depending on another query as the same as being parameterized. This isn't necessarily true, but checking cards can get expensive, especially in light of the fact that those queries can depend on other queries, which can depend on other queries, which .... So this just bailed at the first nested query and wouldn't run the question (which is the linked issue).

The change here will consider nested queries as _not_ parameterized. This is probably correct in most cases. But we don't have a mechanism to check that all of the nested queries are not paramaterized. This just assumes they aren't. Based on my gut feeling of how many people are using a parameterized query as a nested query (this doesn't work as there's no way to pass parameters down into the nested query) I think we can get away with this.

#### Why we think using a nested, parameterized query is very rare
This type of question will currently be broken no matter what, since there's no way to pass parameters down into the card that is referenced. So just configuring the nested query will show immediately that it's broken, and then it not rendering in the collection pinning is understandable.

So when it's pinned, we show `There was a problem displaying this chart` because the question itself is essentially broken due to product limitations. Is this OK?

image

### DIFF
--- a/frontend/src/metabase-types/api/collection.ts
+++ b/frontend/src/metabase-types/api/collection.ts
@@ -75,7 +75,7 @@ export interface CollectionItem {
   copy?: boolean;
   collection_position?: number | null;
   collection_preview?: boolean | null;
-  fully_parametrized?: boolean | null;
+  fully_parameterized?: boolean | null;
   collection?: Collection | null;
   display?: CardDisplayType;
   personal_owner_id?: UserId;
--- a/src/metabase/api/collection.clj
+++ b/src/metabase/api/collection.clj
@@ -93,7 +93,7 @@
                ;; Order NULL collection types first so that audit collections are last
                :order-by [[[[:case [:= :type nil] 0 :else 1]] :asc]
                           [:%lower.name :asc]]})
-    exclude-other-user-collections (remove-other-users-personal-subcollections api/*current-user-id*)))
+   exclude-other-user-collections (remove-other-users-personal-subcollections api/*current-user-id*)))
 
 (api/defendpoint GET "/"
   "Fetch a list of all Collections that the current user has read permissions for (`:can_write` is returned as an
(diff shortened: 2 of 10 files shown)


frontend/src/metabase-types/api/collection.ts
```diff
@@ -75,7 +75,7 @@ export interface CollectionItem {
   copy?: boolean;
   collection_position?: number | null;
   collection_preview?: boolean | null;
-  fully_parametrized?: boolean | null;
+  fully_parameterized?: boolean | null;
```
iethree, line 78: 😁 thanks for fixing the typo

src/metabase/api/collection.clj
```diff
@@ -412,7 +412,7 @@
                                   (comp (filter params/Param?)
                                         (map :k))
                                   (params.parse/parse text))]
-      (and (every? #(or (#{:dimension :snippet} (:type %))
+      (and (every? #(or (#{:dimension :snippet :card} (:type %))
```
noahmoss, line 415: This line is the substance of the change

johnswanson (approved):
Makes sense to me

Issue #37173, opened 2023-12-29
[MLv2] [Testing plan] Custom columns


ranquild:
`availableMetrics`:
- [x] returns a list of metrics for the source table
- [x] returns an empty list when there are no metrics for the source table
- [x] does not return metrics for other tables not used in the query
- [x] does not return metrics for tables used in joins
- [x] does not return metrics for stages other than 0
- [x] does not return metrics for queries based on saved questions, both mbql and native

`availableSegments`:
- [x] same as `availableMetrics`

`legacyRef`:
- [x] works for columns
- [x] works for segments -> `["segment", 1]`
- [x] works for metrics => `["metric", 1]`

`expressionClauseForLegacyExpression`:
- [x] works for simple expressions `1 + 2`
- [x] works for expressions with columns (with `legacyRef`)
- [x] works for expressions with segments (with `legacyRef`)
- [x] works for expressions with metrics (with `legacyRef`)
- [x] result expression can be used to create a filter with `filter` call
- [x] result expression can be used to create an aggregation with `aggregate` call
- [x] result expression can be used to create a custom column with `expression` call

`legacyExpressionForExpressionClause`:
- [x] works for simple expressions `1 + 2`
- [x] works for expressions with columns
- [x] works for expressions with segments
- [x] works for expressions with metrics
- [x] works for filters obtained from `filters` call
- [x] works for aggregations obtained from `aggregations` call
- [x] works for custom columns obtained from `expressions` call
- [x] works with `expressionClauseForLegacyExpression` - should result in the original expression

`withExpressionName`:
- [x] works for filters obtained from `filters` call
- [x] works for aggregations obtained from `aggregations` call
- [x] works for custom columns obtained from `expressions` call
- [x] works for expressions obtained from `expressionClauseForLegacyExpression` call
- [x] allows to add the name for an expression without one and `displayInfo` uses this name
- [x] allows to update the name for an expression and `displayInfo` uses this name

`expression`
- [x] adds a new expression obtained from `expressionClauseForLegacyExpression` call

`expressions`
- [x] returns expressions added with `expression` call

PR #37767, opened 2024-01-16
Add test for available-metrics and available-segments

metamben:
Part of #37173.

### DIFF
--- a/src/metabase/lib/metric.cljc
+++ b/src/metabase/lib/metric.cljc
@@ -97,18 +97,19 @@
    (available-metrics query -1))
   ([query :- ::lib.schema/query
     stage-number :- :int]
-   (when-let [source-table-id (lib.util/source-table-id query)]
-     (let [metrics (lib.metadata.protocols/metrics (lib.metadata/->metadata-provider query) source-table-id)
-           metric-aggregations (into {}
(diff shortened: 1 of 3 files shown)


snoe (approved):

src/metabase/lib/metric.cljc
```diff
@@ -97,18 +97,19 @@
...
-                                                  aggregation-pos (assoc :aggregation-position aggregation-pos))))
-                                            metrics))))))
+   (when (zero? (lib.util/canonical-stage-index query stage-number))
+     (when-let [source-table-id (lib.util/source-table-id query)]
+      (let [metrics (lib.metadata.protocols/metrics (lib.metadata/->metadata-provider query) source-table-id)
+            metric-aggregations (into {}
+                                      (keep-indexed (fn [index aggregation-clause]
+                                                      (when (lib.util/clause-of-type? aggregation-clause :metric)
+                                                        [(get aggregation-clause 2) index])))
+                                      (:aggregation (lib.util/query-stage query stage-number)))]
```
snoe, line 107: This could probably use the `lib/aggregations` fn

metamben: I'll sneak that change in in my next PR.

Issue #36400, opened 2023-12-05
Error calculating display info on filtering with a model


darksciencebase:
> This seems MBQL-related. In the repro I:
> 1. Went to [the model](https://stats.metabase.com/model/14583)
> 2. Filtered in chill mode (which works)
> 3. Opened notebook mode
> 4. Removed the filter
> 5. Opened the filter UI again
> 6. It broke

[Slack Message](https://metaboat.slack.com/archives/C05MPF0TM3L/p1701771677901339?thread_ts=1701771677.901339&cid=C05MPF0TM3L) (there's a loom of the repro in the thread)

```
n0.gi {message: "Error calculating display info for column-group: Cannot read properties of null (reading 'replace')", data: n0.h, cause: TypeError: Cannot read properties of null (reading 'replace')
    at a (https://stats.metabase.com/…, name: 'Error', description: undefined, …}
iJ @ react-dom.production.min.js:209
n.payload @ react-dom.production.min.js:227
oO @ react-dom.production.min.js:129
iI @ react-dom.production.min.js:179
l @ react-dom.production.min.js:263
aH @ react-dom.production.min.js:246
(anonymous) @ react-dom.production.min.js:246
aW @ react-dom.production.min.js:246
(anonymous) @ react-dom.production.min.js:123
t.unstable_runWithPriority @ scheduler.production.min.js:19
on @ react-dom.production.min.js:122
oi @ react-dom.production.min.js:123
oo @ react-dom.production.min.js:122
Y @ react-dom.production.min.js:287
tU @ react-dom.production.min.js:68
```

bshepherdson:
I'll take a look at this. It sounds like it might be the same cause as #37067, which has a fixed out for review #37536. Once that is in stats it'll be easy to try again to repro it.

bshepherdson:
Actually I can't reproduce this.

Looking more closely at the steps and outcome, I suspect that this was fixed by removing empty stages (#37311). With a SQL model, creating the filter adds a new stage, and then removing the filter empties the stage. Then I guess something broke when trying to open the Filters UI again.

Closing as no repro.

bshepherdson:
Ah, I can repro this. I'm looking into it.

bshepherdson:
The immediate error is coming from trying to regex-match against the `:display-name` of a field, but it's `nil` rather than a string. The `:display-name` is nil I think because we don't have the target field in the metadata yet.

[...]

PR #37771, opened 2024-01-16
[MLv2] Look up FK columns in `visible-columns`, not metadata

bshepherdson:
Models can override the `:id` and `:fk-target-field-id` of a column, and
can even lie and treat a quantity as a FK. Therefore we have to use the
query's view of the column and not the global field from the metadata.

This is a speculative fix for #36400, which I can't reproduce locally.

### DIFF
--- a/src/metabase/lib/column_group.cljc
+++ b/src/metabase/lib/column_group.cljc
@@ -1,5 +1,6 @@
 (ns metabase.lib.column-group
   (:require
+   [medley.core :as m]
    [metabase.lib.card :as lib.card]
    [metabase.lib.join :as lib.join]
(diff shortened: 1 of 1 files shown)


bshepherdson:
* `master`
  * **PR #37771** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37771?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37771?utm_source=stack-comment).

metamben (approved):

PR #37780, opened 2024-01-17
bump hawk to include the :times option

qnkhuat:
Bump hawk so we can stress test to run multiple times with the `:times` option.

using it like so
```bash
clj -X :dev:test :only test-vars :times 10
```
this will run `test-vars` 10 times. 
test-var can be a namespace, or a single test, or a list of tests.

You can also use this option on GA UI, see https://github.com/metabase/metabase/pull/35919 for how

The option was implemented in https://github.com/metabase/hawk/pull/16


crisptrutski (approved):
Confirmed the SHA matched the corresponding commit to hawk trunk

PR #37781, opened 2024-01-17
env var MB_JETTY_SSL_INSECURE to disable SNI check

### LINKED ISSUE #29660: Env var to disable SNI check in Jetty
When we switched to Jetty 11, now you can't test a single HTTPS Metabase deployment locally, as SNI checks are not allowed for localhost. If we can disable all SNI checks https://www.eclipse.org/jetty/documentation/jetty-11/operations-guide/index.html#og-protocols-ssl-sni via an env var, that would allow us to keep testing HTTPS on a localhost

Add an env var like MB_JETTY_INSECURE that allows us to use HTTPS with SNI checks disabled in localhost

Tried to do this via JAVA_OPTS but it didn't work

Wanted to do some tests locally

**Additional context**
NA

sansolovyov:
Resolves #29660

### DIFF
--- a/src/metabase/server.clj
+++ b/src/metabase/server.clj
@@ -22,13 +22,15 @@
 (defn- jetty-ssl-config []
   (m/filter-vals
    some?
-   {:ssl-port       (config/config-int :mb-jetty-ssl-port)
-    :keystore       (config/config-str :mb-jetty-ssl-keystore)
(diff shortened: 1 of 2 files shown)


crisptrutski (approved):

paoliniluis:
adding a comment here. This var was changed to https://github.com/metabase/metabase/pull/37791

Issue #37765, opened 2024-01-16
[MLv2][BE] `editable?` function doesn't take into account native write permissions


nemanjaglumac:
When working on https://github.com/metabase/metabase/pull/37744, I realized that `editable?` function doesn't take into account native write permissions.
https://github.com/metabase/metabase/blob/1a411da8a27094bfbf6acdfd23e2ed70c7d1037a/src/metabase/lib/metadata.cljc#L213

It currently returns `true` for native queries even if the database doesn't have `native_permissions` set to `write`.
https://github.com/metabase/metabase/blob/1a411da8a27094bfbf6acdfd23e2ed70c7d1037a/src/metabase/api/database.clj#L80

This check has already been implemented in https://github.com/metabase/metabase/blob/backport-1a411da8a27094bfbf6acdfd23e2ed70c7d1037a/src/metabase/lib/native.cljc#L227

I tried to solve the problem in [this commit](https://github.com/metabase/metabase/pull/37744/commits/1175300250f5767a82da95b323a3e6b532bb8e79) but ran into circular dependency problems. Thought I had a workaround, but tests are still failing.

Issue #37389, opened 2024-01-08
[MLv2] [FE] Use MLv2 `displayInfo` in `isStructured`, `isNative`, `isQueryEditable`


ranquild:
Previously blocked by https://github.com/metabase/metabase/issues/36973.
```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/37786
- [ ] https://github.com/metabase/metabase/pull/37862
- [ ] #37805
- [ ] https://github.com/metabase/metabase/pull/37886
- [ ] https://github.com/metabase/metabase/pull/37977
```

PR #37786, opened 2024-01-17
[MLv2] Update types to accommodate `QueryDisplayInfo`

nemanjaglumac:
The progress of https://github.com/metabase/metabase/issues/37389 is blocked by https://github.com/metabase/metabase/issues/37765.

The initial attempt was to both add new types and take care of `isEditable` query methods. In order to unblock further work on #37389, I decided to split the tasks into three. This PR is just the first on the three.

It should unblock the work on migrating `isNative` and `isStructured` methods.

### DIFF
--- a/frontend/src/metabase-lib/metadata.ts
+++ b/frontend/src/metabase-lib/metadata.ts
@@ -144,6 +144,18 @@ declare function DisplayInfoFn(
   stageIndex: number,
   segment: SegmentMetadata,
 ): SegmentDisplayInfo;
+/**
+ * Even though it seems weird to pass the same query two times,
+ * this function follows the same pattern as the other displayInfo functions.
+ * The first two parameters are always a query, and a stage.
+ * The third parameter is what you would like to have info about.
+ * It just only happens that the thing we're examining is (again) the query itself.
+ *
+ * @example
+ * const query = question.query();
+ * const { isEditable, isNative } = Lib.displayInfo(query, -1, query);
+ *
+ */
 declare function DisplayInfoFn(
   query: Query,
   stageIndex: number,
--- a/frontend/src/metabase-lib/metadata.ts
+++ b/frontend/src/metabase-lib/metadata.ts
@@ -157,8 +157,8 @@ declare function DisplayInfoFn(
  *
  */
 declare function DisplayInfoFn(
-  query: Query,
-  stageIndex: number,
+  _query: Query,
+  _stageIndex: number,
   queryInfo: Query,
 ): QueryDisplayInfo;
 
--- a/frontend/src/metabase-lib/metadata.ts
+++ b/frontend/src/metabase-lib/metadata.ts
@@ -42,6 +42,7 @@ import type {
   SegmentDisplayInfo,
   TableDisplayInfo,
   TableMetadata,
+  QueryDisplayInfo,
 } from "./types";
 
 export function metadataProvider(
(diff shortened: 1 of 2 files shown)


frontend/src/metabase-lib/metadata.ts
```diff
@@ -143,6 +144,23 @@ declare function DisplayInfoFn(
...
   segment: SegmentMetadata,
 ): SegmentDisplayInfo;
+/**
+ * Even though it seems weird to pass the same query two times,
+ * this function follows the same pattern as the other displayInfo functions.
+ * The first two parameters are always a query, and a stage.
+ * The third parameter is what you would like to have info about.
+ * It just only happens that the thing we're examining is (again) the query itself.
+ *
+ * @example
```
uladzimirdev, line 154: I'd remove example as it's TS and you know a signature already.

frontend/src/metabase-lib/metadata.ts
```diff
@@ -143,6 +144,23 @@ declare function DisplayInfoFn(
...
+ * const query = question.query();
+ * const { isEditable, isNative } = Lib.displayInfo(query, -1, query);
+ *
+ */
+declare function DisplayInfoFn(
+  query: Query,
```
uladzimirdev, line 160: I'd still name not required things as `_query` and `_stageIndex` to stress out they are optional

uladzimirdev (commented):
Please check comments inline

frontend/src/metabase-lib/metadata.ts
nemanjaglumac: It seems that I don't understand something here. How are the initial query and stageIndex optional?

`Lib.displayInfo(query)` doesn't give you anything useful
`Lib.displayInfo(query, -1, query)` returns `{ isNative, isEditable }`

image

uladzimirdev: I meant that you could have in JS/TS a function that accepts one parameter - query - but passes down 3 (query, -1, query)

anyhow, I'd not spend time on it, if @kamilmielnik is agreed, I'd go with a current variant, but first two params with underscore

kamilmielnik: "optional" is not an accurate word here. Their presence is necessary, but they can be anything, i.e.
`Lib.displayInfo(null, undefined, query)` will work.

kamilmielnik: > anyhow, I'd not spend time on it, if @kamilmielnik is agreed, I'd go with a current variant, but first two params with underscore

Yeah, even though it bothers me a lot, we can just proceed with the 3 arguments.

To make it even more obvious that these arguments don't matter, we can type them as `unknown`:
```
declare function DisplayInfoFn(
  _query: unknown,
  _stageIndex: unknown,
  query: Query,
): QueryDisplayInfo;
```

uladzimirdev: first parameter has to be a query seems

frontend/src/metabase-lib/metadata.ts
```diff
@@ -143,6 +144,23 @@ declare function DisplayInfoFn(
...
+ *
+ * @example
+ * const query = question.query();
+ * const { isEditable, isNative } = Lib.displayInfo(query, -1, query);
+ *
+ */
+declare function DisplayInfoFn(
+  _query: Query,
+  _stageIndex: number,
+  queryInfo: Query,
```
uladzimirdev, line 162: just query
```suggestion
  query: Query,
```

nemanjaglumac: 🤦

nemanjaglumac: Sorry about that.
Fixed it and removed JSDoc `@comment` in https://github.com/metabase/metabase/pull/37786/commits/265fb7d4826063158af1b680eafd4657ab546fb8

kamilmielnik (approved):

uladzimirdev (approved):
:shipit:

PR #37793, opened 2024-01-17
Add SNI check env var docs

dragonsahead:
Adding to the docs the env var about https://github.com/metabase/metabase/pull/37791

### DIFF
--- a/docs/configuring-metabase/environment-variables.md
+++ b/docs/configuring-metabase/environment-variables.md
@@ -1331,7 +1331,7 @@ Type: string<br>
 Default: `"true"`<br>
 Since: v48.4
 
-If this environment variable is present, it will disable the SNI checks in the Jetty web server. Normally you would leave this enabled unless you're terminating the TLS connection on Metabase itself. If this is the case and you're getting an error like `HTTP ERROR 400 Invalid SNI`, consider setting this environment variable to false or use another SSL certificate that matches exactly the domain name of the server.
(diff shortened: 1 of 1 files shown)


docs/configuring-metabase/environment-variables.md
```diff
@@ -1325,6 +1325,14 @@ Since: v42.0
 
 Set the system files channel used by Metabase to store images. This channel has to be public, and is not intended to be used by humans. The Slack App has to be invited into this channel.
 
+### `MB_JETTY_SKIP_SNI`
+
+Type: string<br>
+Default: `"true"`<br>
+Since: v48.4
+
+If this environment variable is present, it will disable the SNI checks in the Jetty web server. Normally you would leave this enabled unless you're terminating the TLS connection on Metabase itself. If this is the case and you're getting an error like `HTTP ERROR 400 Invalid SNI`, consider setting this environment variable to false or use another SSL certificate that matches exactly the domain name of the server.
```
jeff-bruemmer, line 1334:
```suggestion
Setting `MB_JETTY_SKIP_SNI=true` (the default setting) turns off the Server Name Indication (SNI) checks in the Jetty web server. Normally you would leave this enabled. If, however, you're terminating the Transport Layer Security (TLS) connection on Metabase itself, and you're getting an error like `HTTP ERROR 400 Invalid SNI`, consider either setting `MB_JETTY_SKIP_SNI=false`, or use another SSL certificate that exactly matches the domain name of the server.
```

jeff-bruemmer (approved):

jeff-bruemmer:
One thing to note though. We should probably have all boolean knobs defined such that they default to off.

This env var in particular is strange because if the env var is set to on, something is actually turned off (true = no check, false = check).

Might make more sense to make the variable `MB_JETTY_CHECK_SNI` and default that to false. If you want to turn on the check, you turn on the env var.

PR #37794, opened 2024-01-17
Add uniqueness constraint on the logical identity of field values

crisptrutski:
### Description

Refs https://github.com/metabase/metabase/issues/668

This adds a uniqueness constraint over the columns encapsulating the context for a given set of field values.

In practice the `hash_key` should always be a 32 bit integer, at least for rows created by recent versions of the code (and old rows should be pruned), but out of an abundance of caution and lack of necessity we have kept it as a string type, merely giving it fixed width to enable the creation of the index. In any case H2 does not allow a direct conversion from text to integer, and would need to be made a varchar first.

This PR has been created so far simply to run CI as a smoke test to find code paths that create redundant rows.

### How to verify

...

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/sandbox/models/params/field_values_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/sandbox/models/params/field_values_test.clj
@@ -1,6 +1,6 @@
 (ns metabase-enterprise.sandbox.models.params.field-values-test
   (:require
-   [clojure.test :refer :all]
+   [clojure.test :refer [deftest testing is]]
    [java-time :as t]
    [metabase-enterprise.sandbox.models.group-table-access-policy
     :refer [GroupTableAccessPolicy]]
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -5155,16 +5155,24 @@ databaseChangeLog:
   - changeSet:
       id: v49.2024-01-16T12:50:35
       author: crisptruski
-      comment: Add unique constraint on fieldvalues
+      comment: Prepare  fieldvalues
       changes:
         - modifyDataType:
             tableName: metabase_fieldvalues
             columnName: hash_key
+            # we only need 11 characters to fit any 32 bit number, but we leave space in case the hash scheme changes
             newDataType: varchar(32)
-#        - modifyDataType:
-#            tableName: metabase_fieldvalues
-#            columnName: hash_key
-#            newDataType: integer
+      rollback:
+        - modifyDataType:
+            tableName: metabase_fieldvalues
+            columnName: hash_key
+            newDataType: ${text.type}
+
+  - changeSet:
+      id: v49.2024-01-17T13:55:14
+      author: crisptruski
+      comment: Add unique constraint on fieldvalues
+      changes:
         - addUniqueConstraint:
             tableName: metabase_fieldvalues
             constraintName: unique_metabase_fieldvalues_field_id_type_hash_key
@@ -5173,10 +5181,6 @@ databaseChangeLog:
         - dropUniqueConstraint:
             tableName: metabase_fieldvalues
             constraintName: unique_metabase_fieldvalues_field_id_type_hash_key
-        - modifyDataType:
-            tableName: metabase_fieldvalues
-            columnName: hash_key
-            newDataType: ${text.type}
 
   # >>>>>>>>>> DO NOT ADD NEW MIGRATIONS BELOW THIS LINE! ADD THEM ABOVE <<<<<<<<<<
 
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -5170,6 +5170,31 @@ databaseChangeLog:
 
   - changeSet:
       id: v49.2024-01-17T13:55:14
+      author: crisptrutski
+      comment: Prepare  fieldvalues
+      changes:
+        - sql:
+            sql: UPDATE metabase_fieldvalues SET hash_key = '' WHERE hash_key IS NULL;
+      rollback:
+        - sql:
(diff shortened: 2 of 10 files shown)


crisptrutski:
It turns out this constraint does not work in practice for `:full` rows, as for all of H2, MySQL, and Postgres we have `NULL != NULL`, so semantic duplicates are not prevented.

We could work around this by using the empty string instead of null, and we could use a toucan hook to hide this detail from the application - going to give that a shot.

crisptrutski:
This approach isn't working as toucan skips the field translation for nil values :-/

crisptrutski:
Remaining test failures seem to be where the uniqueness constraint is preventing previous duplicates.

It would be really elegant to replace the various insert statements with upserts, but unfortunately [toucan does not yet support this operation.](https://github.com/camsaul/toucan2/issues/57)

Probably the best route will be to compose our own special case upsert just for this model, and be diligent to use it. 

I suspect most of the failures in CI are due to lack of hygeine in tests rather than application bugs now - after running the test suite there are a bunch of values left in the DB.

camsaul (changes requested):
#668 is only about preventing the app DB from having multiple `:full` (non-advanced) FieldValues entries for a single Field, I think putting the unique constraint on `hash_key` like that is not preventing someone from saving multiple `:full` entries with different values of `hash_key` (not that this happens in practice, but then again creating duplicate entries isn't supposed to happen in practice either)

Not sure but maybe if you could do a unique constraint like 

```sql
field_id, CASE WHEN type = 'full' THEN type ELSE NULL)
```

(relying on magic `NULL` here), it might solve the problems. Not sure if MySQL or H2 let you do this tho.

But really tho I think we just need to refactor things and move the "advanced" FieldValues into a different Table and then have the regular FieldValues table just hold regular FieldValues. As I mentioned here https://github.com/metabase/metabase/issues/668#issuecomment-1373335680 it's just too easy for people to accidentally forget to include `type = 'full'` in their queries, and it breaks a lot of stuff. 

Oh yeah also if you have a migration that adds a unique constraint, you need to have a migration that will clean up duplicate entries before adding the constraint, and you should put a migration test around it. See https://github.com/metabase/metabase/issues/668#issuecomment-318789467 for an example of what I mean. Otherwise Metabase can potentially fail on launch if someone had a duplicate entry. Smoke testing CI is not going to expose that sort of stuff

crisptrutski:
Thanks for jumping on this PR, I really appreciate the help! 

> #668 is only about preventing the app DB from having multiple `:full` (non-advanced) FieldValues entries for a single Field

I saw that, but my interpretation was simply that that they were the most important case. Won't advanced fields have the similar issues if we have multiple entries for the same hash? The saving grace here is that the problem goes away once the parameters change and we get a new hash. Maybe I'm missing some bigger difference between the two cases?

> I think putting the unique constraint on `hash_key` like that is not preventing someone from saving multiple `:full` entries with different values of `hash_key` (not that this happens in practice, but then again creating duplicate entries isn't supposed to happen in practice either)

In participle you're right, this leaves a hole, but I'm not convinced it matters that much:

1. We have toucan model hooks to throw in we give a hash for a :full row.
2. We could also pop an is null check on the selects for two factor peace of mind.

There are so many essential invariants we can't enforce, this seems like a minor one compared to FKs on polymorphic rows or JSON fields, etc. We chose a language that values simplicity over restrictive or overly complex type systems, I'm not sure this constraint should twist our data modelling too much either.

> Not sure but maybe if you could do a unique constraint like
> 
> ```sql
> field_id, CASE WHEN type = 'full' THEN type ELSE NULL)
> ```
> 
> (relying on magic `NULL` here), it might solve the problems. Not sure if MySQL or H2 let you do this tho.

I suspect support will be an issue, and it feels a bit overwrought.

> But really tho I think we just need to refactor things and move the "advanced" FieldValues into a different Table and then have the regular FieldValues table just hold regular FieldValues. As I mentioned here [#668 (comment)](https://github.com/metabase/metabase/issues/668#issuecomment-1373335680) it's just too easy for people to accidentally forget to include `type = 'full'` in their queries, and it breaks a lot of stuff.

My gut feeling is that these entities have more in common than they have differences, especially from a FE perspective. I worry a bit that forking them will increase the code and operational maintenance burden. Perhaps I'm missing something about how they differ fundamentally?

> Oh yeah also if you have a migration that adds a unique constraint, you need to have a migration that will clean up duplicate entries before adding the constraint, and you should put a migration test around it. See [#668 (comment)](https://github.com/metabase/metabase/issues/668#issuecomment-318789467) for an example of what I mean. Otherwise Metabase can potentially fail on launch if someone had a duplicate entry. 

I was just deferring writing the migration until the PR was closer to done, as I was lazy. 😏 

> Smoke testing CI is not going to expose that sort of stuff

Definitely! The reason I parked this issue is that I'm terrified about adding an index without fixing all the bugs that led to inserting duplicates. I'm more nervous about customers hitting show stopper runtime bugs that really mess up their db, than about an inability to upgrade.

For now what I was hoping to achieve with this branch was a basis to "fuzz test" (being quite metaphorical) with some guidance from someone more familiar with the app.

Regarding the terror of actual rollout, perhaps it would be reasonable to approach rollout as follows:

1. Add a scheduled task which deletes duplicates as they arise, to mitigate issues.
2. Add telemetry to the app that reports any duplicates back home. 
3. Make fixes that we think will prevent further duplicates.
4. ...
5. Deploy migration to add db constraint.

Perhaps there are data collection concerns around (2), but it would be really great to have complete confidence in the safety of (5) after a sufficiently long "baking" period. We could report how many rows are deleted by (1), or perhaps catch them even earlier by a hook on our insert statements. The more context we can get on the process that created the duplicate the better!

crisptrutski:
This has rotted. I still like the idea of adding this constraint even with a shared table, but it's not high priority. Will discuss further on the parent issue.

Issue #37434, opened 2024-01-09
BE flake: metabase.db.schema-migrations-test/audit-v2-views-test


calherries:
Failed test: https://github.com/metabase/metabase/actions/runs/7458347037/job/20292163749?pr=37405#step:4:2097

qnkhuat:
I did some investigation, and it looks like Liquibase is skipping the rollback for these view migrations from `v48.00-029 -> v48.00-040`.

Switched the log level for liquibase to debug, you'll see that liquibase overshoot the rollback and rolled back until `v47.00-045`, whereas it should stop at `v47.00-059` because it's the last migration of 47.

I haven't figured out why but looks like a liquibase bug.

here is how I tested it locally
```
DRIVERS=mysql MB_DB_CONNECTION_URI=jdbc:mysql://localhost:3306/metabase?user=root clj -X:dev:ee:ee-dev:drivers:drivers-dev:test :only metabase.db.schema-migrations-test/audit-v2-views-test :times 10
```

calherries:
Thanks @qnkhuat for investigating. I've created a PR to comment out this test because we're pretty safe without this test anyway https://github.com/metabase/metabase/pull/37796, and this flake isn't getting picked up

calherries:
Update: this PR only comments out the part that is flaking, and not the whole test https://github.com/metabase/metabase/pull/37883

calherries:
@qnkhuat I've created an issue to track the general problem of rollback flakes here https://github.com/metabase/metabase/issues/37884

qnkhuat:
it came back https://metaboat.slack.com/archives/C0641E4PB9B/p1710857861360739?thread_ts=1710770688.837979&cid=C0641E4PB9B

calherries:
I already commented out this test as part of https://github.com/metabase/metabase/pull/40291

PR #37796, opened 2024-01-17
Comment out flaky audit-v2-views-test

calherries:
Temporary fix for https://github.com/metabase/metabase/issues/37434

This PR comments out `metabase.db.schema-migrations-test/audit-v2-views-test`.

Commenting out the test is relatively safe because it is testing a SQL migration, and changes to migrations shouldn't get past PR review.

### DIFF
--- a/test/metabase/db/schema_migrations_test.clj
+++ b/test/metabase/db/schema_migrations_test.clj
@@ -540,6 +540,7 @@
         (testing "should drop the existing color column"
           (is (not (contains? (t2/select-one :model/Collection :id collection-id) :color))))))))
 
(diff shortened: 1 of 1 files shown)


sansolovyov (approved):

qnkhuat:
I don’t think it’s safe to comment that test out. It was purposefully added to detect cases where a future migration changed and somehow broke the view

calherries:
You're entirely correct, I moved too fast here. I'll uncomment the test but that reasoning doesn't apply to the rollback assertion that is flaky, I'll just comment that out instead.

PR #37800, opened 2024-01-17
Add required field to the template tag parameter

oleggromov:
Part of the milestone 1 (required parameters in native queries), epic https://github.com/metabase/metabase/issues/36524

As we need to make parameters in native queries required, this is the first step. I am making a separate PR so that it's easier to track which tests fail (quite a few do on the BE) — and either merge it separately, or at least merge it back to the original PR.

### DIFF
--- a/src/metabase/models/card.clj
+++ b/src/metabase/models/card.clj
@@ -271,7 +271,8 @@
                 [:variable  [:template-tag (:name tag)]])
      :name    (:display-name tag)
      :slug    (:name tag)
-     :default (:default tag)}))
+     :default (:default tag)
+     :required (:required tag)}))
 
 (defn- check-field-filter-fields-are-from-correct-database
   "Check that all native query Field filter parameters reference Fields belonging to the Database the query points
--- a/src/metabase/models/card.clj
+++ b/src/metabase/models/card.clj
@@ -261,18 +261,18 @@
         :when                         (and tag-type
                                            (or (contains? mbql.s/raw-value-template-tag-types tag-type)
                                                (and (= tag-type :dimension) widget-type (not= widget-type :none))))]
-    {:id      (:id tag)
-     :type    (or widget-type (cond (= tag-type :date)   :date/single
-                                    (= tag-type :string) :string/=
-                                    (= tag-type :number) :number/=
-                                    :else                :category))
-     :target  (if (= tag-type :dimension)
-                [:dimension [:template-tag (:name tag)]]
-                [:variable  [:template-tag (:name tag)]])
-     :name    (:display-name tag)
-     :slug    (:name tag)
-     :default (:default tag)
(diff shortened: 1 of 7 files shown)


src/metabase/models/card.clj
```diff
@@ -271,7 +271,8 @@
                 [:variable  [:template-tag (:name tag)]])
      :name    (:display-name tag)
      :slug    (:name tag)
-     :default (:default tag)}))
+     :default (:default tag)
+     :required (:required tag)}))
```
oleggromov, line 275: Many tests fail with `:required nil`, so should we add this key conditionally?

metamben: Yes, it seems it's optional, but if it's there, it should be a boolean.

metamben (approved):
~Should` there be some tests for this? E.g., are there no existing tests that could/should be adjusted?~ I see, this is just to track failures, right?

oleggromov:
> ~Should` there be some tests for this? E.g., are there no existing tests that could/should be adjusted?~ I see, this is just to track failures, right?

Actually I intend to merge this PR, or at least make sure all tests are green. 
My question is: should I go and update all failing tests to simply have the `:required` key in the map?

metamben:
> should I go and update all failing tests to simply have the `:required` key in the map?

👍 I guess that will fix the bulk of the cases. You can always ask for help if things turn out to be complicated.

src/metabase/models/card.clj
oleggromov: I updated it to be `(boolean (:required tag))`. Is this how you usually convert `nil` to `false`?

metamben: Yes, that's the way to convert truthy/falsey values to boolean. The question is if this is what we need here. The other option is not to set it if `tag` doesn't have it. I can't tell just by quick look which way is the correct one.

oleggromov: Given that no other key in this map is set conditionally, I am assuming it's better to leave it as is.

metamben (approved):
I assume you intend to normalize `:required` in `template-tag-parameters`, e.g., it's not a problem that we lose the information that this property was not defined in the card.

src/metabase/models/card.clj
metamben: But note that for example `:default` can be set to `nil`, so it wouldn't lose information.

oleggromov:
@metamben and I have discussed in DM that since from the user's point of view an unset `required` equals `false`, it's ok to leave it as is.

camsaul (approved):
This looks good to me, the only thing I would suggest is that you add `:required` as an optional parameter to the schema for parameters in `src/metabase/mbql/schema.cljc` (line 1643, or just grep for `[:slug` to find it)

👍

kamilmielnik (approved):

PR #37802, opened 2024-01-17
[WIP - open to let the e2e run] better error for expired token in dashcard request

### LINKED ISSUE #11728: Ambiguous embedded dashboard message on expired JWT
Upon JWT expiration for embedded dashboards, cards return the ambiguous message <i>There was a problem displaying this chart.</i> leaving users hanging on what could possibly be wrong.  The same cards embedded separately return the far more specific and preferable message <i>Token is expired (1579057397)</i> telling users precisely why the resource failed to load.

Reproduce by embedding both a dashboard and one of it's cards as an embedded question.  Then change the JWT expiration mechanism such as <i>->setExpiration(time() + 3600)</i> from the example 1hr (3600s) to 1s.  Refresh the embeds and make filter selections until the request is declined to see the difference in the messages returned.  

Showing a clear message in dashboard cards like <i>Token is expired</i> would disambiguate the issue considerably.

Metabase: v0.34.1 jar
Host: Localhost:3000
Databases: H2 internal default, Server SQL added
Browser: Firefox 52.9.0
OS: Microsoft Windows [Version 6.0.6002]
Java: 1.8.0_45

npretto:
Closes https://github.com/metabase/metabase/issues/11728

### Description

`/embed/*` endpoints don't return error codes or too much information about the error by choice (see [here](https://github.com/metabase/metabase/blob/0a6e5ad9e9419cd1a2d525108bd7136750309316/src/metabase/api/routes/common.clj#L13-L16))
Nonetheless the linked issue needed a better explanation to not leave the user confused.
The issue arises when you have a dashboard with filters in a static embed and the user changes the filter after the signed JWT has expired.

### How to verify

- Create a dashboard
- Add a filter and connect it to a question
- Do a static embed of the dashboard, set the jwt to expire at the time of the creation (there is a leeway of 60 seconds anyway, so you'll have time to open it)
- load the dashboard, it should load and work just fine
- wait for the minute to pass so that the jwt expires
- change the value of the filter, it will trigger a reload of the card that will run into the error
- it should display the new error prompting to reload the page

### Demo

image

### Checklist


src/metabase/util/embed.clj
```diff
@@ -98,7 +98,9 @@
                   {:leeway 60})
       ;; if `jwt/unsign` throws an Exception rethrow it in a format that's friendlier to our API
       (catch Throwable e
-        (throw (ex-info (.getMessage e) {:status-code 400}))))))
+        (if (str/starts-with? (ex-message e) "Token is expired")
+          (throw (ex-info "Token is expired" {:status-code 403}))
```
npretto, line 102: Since we'll act on this message on the FE I wanted to make it more explicit.
Initially I thought to add an error code but I stopped at this step for now

frontend/src/metabase-types/api/mocks/dashboard.ts
```diff
@@ -26,6 +26,7 @@ export const createMockDashboard = (opts?: Partial<Dashboard>): Dashboard => ({
   archived: false,
   public_uuid: null,
   enable_embedding: false,
+  tabs: [],
```
npretto, line 29: It seems that the BE always returns this to us, and the test would fail because the FE code checks the `tabs.length > 0`

src/metabase/util/embed.clj
camsaul: I think error codes really are the right way to do this here. "Token is expired" really should be i18n'ed, and having the FE hardcode a specific error message and do text matching on it is a recipe for bugs in the future... it's way too easy for someone to not know this specific error message is looked for by the FE and change it or i18n it and then accidentally break stuff.

The FE can look for the 403 status code, right? When you say error codes aren't supported on `embed` endpoints, keep in mind that we can always tweak the middleware or whatever we need to make them supported -- let's take the extra time to do a robust solution rather that makes it hard to shoot yourself in the foot in the future

camsaul: btw all this stuff you wrote here in the PR comment is really useful information, it would be 100x more useful tho if you added it to the code itself as a comment. People really need to know all of this context.

camsaul (changes requested):
- Let's have the FE check the 403 status code or an error key in the API response JSON body rather than an error message that someone can accidentally change in the future. If you can't make it work easily, ping me on Slack and we'll pair on it, we might need to tweak some of our Ring middleware
- That error message should probably be i18n'ed with `tru`
- This code really needs some comments explaining what's going on, especially if there's something like a string that it is critical that you do not change (altho of course it is much better not to rely on the magic string here in the first place). 
- Where are the tests?

src/metabase/driver/sql/parameters/substitute.clj
```diff
@@ -95,7 +95,7 @@
                                         e))))]
     (log/tracef "=>%s\n%s" sql (pr-str args))
     (when (seq missing)
-      (throw (ex-info (tru "Cannot run the query: missing required parameters: {0}" (set missing))
-               {:type    qp.error-type/missing-required-parameter
-                :missing missing})))
+      (throw (ex-info (tru "Something something, this seems to be returned as json")
```
npretto, line 98: I changed this to use the same error-type I need to see if that was the problem, but this does get returned as a json from the endpoint.

src/metabase/query_processor/error_type.clj
```diff
@@ -94,3 +99,6 @@
 (deferror db
   "Generic ancestor type for all unexpected errors returned or thrown by a data warehouse when running a query."
   :parent server)
+
+
+(show-in-embeds? embed-token-expired)
```
npretto, line 104: This returns true 🤷 (the entire commit will be reverted)

src/metabase/util/embed.clj
```diff
@@ -100,7 +101,10 @@
                   {:leeway 60})
       ;; if `jwt/unsign` throws an Exception rethrow it in a format that's friendlier to our API
       (catch Throwable e
-        (throw (ex-info (.getMessage e) {:status-code 400}))))))
+        (if (str/starts-with? (ex-message e) "Token is expired")
+           (throw (ex-info (tru "Token is expired. This is not returned as json with the error code")
```
npretto, line 105: This gets returned as string in the body, I need the json with the error code

src/metabase/util/embed.clj
npretto: Thanks Cam, I looked into it again, it seems that endpoint is supposed to return error codes in some situations but I wasn't able to do it. I [asked for help](https://metaboat.slack.com/archives/C0641E4PB9B/p1706273158446959) on slack and added more context there

src/metabase/util/embed.clj
crisptrutski: I have a hunch this could be due to whatever is catching the exception checking the :status-code, and assuming that this is an internal error we don't want to expose ex-data for if it's missing or a 500.

crisptrutski: Maybe try adding `:status-code 401`?

PR #37805, opened 2024-01-17
Migrate `isNative` to MLv2

nemanjaglumac:
This PR is just one of the tasks for #37389.
It migrates the old `isNative()` method from the `Question` prototype to MLv2. It ultimately removes it the last commit.

The old way to determine whether a query is native was to check is it an instance of a `NativeQuery` class.
MLv2 provides a direct way to achieve this:
```js
const { isNative } = Lib.queryDisplayInfo(query);
```

### DIFF
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -204,15 +204,12 @@ class Question {
   }
 
   isNative(): boolean {
-    return (
-      this.legacyQuery({ useStructuredQuery: true }) instanceof NativeQuery
-    );
+    const { isNative } = Lib.queryDisplayInfo(this.query());
+    return isNative;
   }
 
   isStructured(): boolean {
-    return (
-      this.legacyQuery({ useStructuredQuery: true }) instanceof StructuredQuery
-    );
+    return !this.isNative();
   }
 
   /**
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -184,7 +184,7 @@ class Question {
     }
 
     const isVirtualDashcard = !this._card.id;
-    // `dataset_query` is null for questions on a dashboard the user don't have access to
+    // The `dataset_query` is null for questions on a dashboard the user doesn't have access to
     !isVirtualDashcard &&
       console.warn("Unknown query type: " + datasetQuery?.type);
   });
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -206,16 +206,9 @@ class Question {
   /**
    * @deprecated use MLv2
    */
-  isNative(): boolean {
-    const { isNative } = Lib.queryDisplayInfo(this.query());
-    return isNative;
-  }
-
-  /**
-   * @deprecated use MLv2
-   */
   isStructured(): boolean {
-    return !this.isNative();
+    const { isNative } = Lib.queryDisplayInfo(this.query());
+    return !isNative;
   }
 
   /**
@@ -1115,9 +1108,10 @@ class Question {
    */
   canExploreResults() {
     const canNest = Boolean(this.database()?.hasFeature("nested-queries"));
+    const { isNative } = Lib.queryDisplayInfo(this.query());
 
     return (
-      this.isNative() &&
+      isNative &&
       this.isSaved() &&
       this.parameters().length === 0 &&
       canNest &&
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -203,11 +203,17 @@ class Question {
     return query;
   }
 
+  /**
+   * @deprecated use MLv2
+   */
   isNative(): boolean {
     const { isNative } = Lib.queryDisplayInfo(this.query());
     return isNative;
   }
 
+  /**
+   * @deprecated use MLv2
+   */
   isStructured(): boolean {
     return !this.isNative();
   }
--- a/frontend/src/metabase-lib/parameters/utils/click-behavior.ts
+++ b/frontend/src/metabase-lib/parameters/utils/click-behavior.ts
@@ -134,15 +134,11 @@ function notRelativeDateOrRange({ type }: Parameter) {
 export function getTargetsForQuestion(question: Question): Target[] {
   const { isNative } = Lib.queryDisplayInfo(question.query());
 
-  if (!isNative) {
-    return getTargetsForStructuredQuestion(question);
-  }
-
   if (isNative) {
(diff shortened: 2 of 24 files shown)


frontend/src/metabase-lib/metadata.ts
```diff
@@ -153,9 +153,17 @@ declare function DisplayInfoFn(
...
 ): QueryDisplayInfo;
 
+declare function DisplayInfoFn(query: Query): QueryDisplayInfo;
+
 // x can be any sort of opaque object, e.g. a clause or metadata map. Values returned depend on what you pass in, but it
 // should always have display_name... see :metabase.lib.metadata.calculation/display-info schema
-export const displayInfo: typeof DisplayInfoFn = ML.display_info;
+export const displayInfo: typeof DisplayInfoFn = (...args: unknown[]) => {
+  if (args.length === 1) {
+    return ML.display_info(undefined, undefined, ...args);
```
uladzimirdev, line 162: I think you'll need to pass query and stageIndex anyway

```
const query = args[0]
const stageIndex = -1;

return ML.display_info(query, stageIndex, query)
```

nemanjaglumac: That was the most surprising thing - you don't 😮 
image

nemanjaglumac: @snoe confirmed in Slack that this currently works "only because Malli validation is not turned on in cljs".

So this is a dead end attempt.
We should forget about another function overload for now. Either we create a named abstraction (e.g. `queryDisplayInfo`) or we wait to see if these named, simplified functions start appearing in backend.

cc @kamilmielnik @uladzimirdev

kamilmielnik: `queryDisplayInfo` :+1:

nemanjaglumac: Split this step into a separate PR:
https://github.com/metabase/metabase/pull/37862

kamilmielnik (commented):
Changes look good but some tests are failing now.

frontend/src/metabase-lib/Question.ts
```diff
@@ -203,16 +203,19 @@ class Question {
     return query;
   }
 
+  /**
+   * @deprecated use MLv2
```
uladzimirdev, line 207: an example of how to use MLv2 would be helpful

frontend/src/metabase/parameters/utils/dashboards.ts
```diff
@@ -139,12 +138,15 @@ function buildFieldFilterUiParameter(
...
-    const question = new Question(card, metadata);
 
     try {
       const field = getTargetFieldFromCard(target, card, metadata);
 
-      return { field, shouldResolveFkField: !question.isNative() };
+      return {
+        field,
+        // The `dataset_query` is null for questions on a dashboard the user doesn't have access to
+        shouldResolveFkField: Boolean(card.dataset_query?.type !== "native"),
```
uladzimirdev, line 148: no need for `Boolean`

nemanjaglumac: `shouldResolveFkField` expects a boolean value. If you pass `undefined` of `null` TS complains.

frontend/src/metabase-lib/parameters/utils/click-behavior.ts
```diff
@@ -132,11 +132,13 @@ function notRelativeDateOrRange({ type }: Parameter) {
...
 export function getTargetsForQuestion(question: Question): Target[] {
-  if (question.isStructured()) {
+  const { isNative } = Lib.queryDisplayInfo(question.query());
+
+  if (!isNative) {
     return getTargetsForStructuredQuestion(question);
   }
 
-  if (question.isNative()) {
+  if (isNative) {
```
kamilmielnik, line 141: We can get rid of the default `return []` now - this function should only have 1 `if` statement.

frontend/src/metabase/dashboard/utils.ts
```diff
@@ -97,7 +96,8 @@ export function isLinkDashCard(dashcard: DashboardCard) {
 }
 
 export function isNativeDashCard(dashcard: DashboardCard) {
-  return dashcard.card && new Question(dashcard.card).isNative();
+  // The `dataset_query` is null for questions on a dashboard the user doesn't have access to
```
kamilmielnik, line 99: Please update `UnsavedCard["dataset_query"]` type to account for the `null`.
This comment could live next to that property, not here.

frontend/src/metabase/lib/click-behavior.js
```diff
@@ -42,7 +41,7 @@ export function getClickBehaviorDescription(dashcard) {
 export function hasActionsMenu(dashcard) {
   // This seems to work, but it isn't the right logic.
   // The right thing to do would be to check for any drills. However, we'd need a "clicked" object for that.
-  return !new Question(dashcard.card).isNative();
+  return dashcard.card.dataset_query?.type !== "native";
```
kamilmielnik, line 44: `hasActionsMenu` should probably evaluate to `false` when `dataset_query` is `null`.
We can also inverse the check for shorter notation.

```suggestion  
  return dashcard.card.dataset_query?.type === "query";
```

frontend/src/metabase/parameters/utils/dashboards.ts
kamilmielnik: Uladzimir is right :+1: 

----

`shouldResolveFkField` should probably evaluate to `false` when `dataset_query` is `null`.
We can also inverse the check for shorter notation.

```suggestion
        shouldResolveFkField: card.dataset_query?.type === "query",
```

frontend/src/metabase/lib/click-behavior.js
nemanjaglumac: Yeah, this makes sense. I don't know what was I thinking. Sorry @uladzimirdev 

No idea why TS was complaining originally.

frontend/src/metabase/dashboard/utils.ts
nemanjaglumac: Do you mind if I do that in a separate PR?
Updating the type will break type check in many places. It will pollute this diff.

frontend/src/metabase-lib/parameters/utils/click-behavior.ts
nemanjaglumac: I wonder why was `return []` there even before this change?
Because of the `InternalQuery`?

kamilmielnik: Yeah, probably.

kamilmielnik (approved):
:+1:

PR #37813, opened 2024-01-17
Add `align-temporal-unit-with-param-type-and-value` multimethod

### LINKED ISSUE #216: No results on specific date for datetime column
### Describe the bug
No results on specific date for datetime column

### Steps to reproduce
1. Create a table with datetime column
2. Create a date time filter
3. Select `Current` -> `Day` or `Specific dates` -> `On`
4. Get `No results!`

### Expected behaviour
Get results on the specific date, similar to what the v1.2.5 driver does.
In v1.2.5, CAST is used to ensure that the datetime column is comparable with the input date, but newer versions "Removed superfluous CAST calls from generated queries that use Date* columns and/or intervals".

#### Environment
* metabase-clickhouse-driver version: 1.3.2 (1.2.5 works well)
* metabase-clickhouse-driver configuration: 
* Metabase version: 0.48.3
* OS:

#### ClickHouse server
* ClickHouse Server version: 22.8.5.29
* ClickHouse Server non-default settings, if any:
* `CREATE TABLE` statements for tables involved:
```
CREATE TABLE test.test
(
    `min_time` DateTime
)
ENGINE = ReplacingMergeTree
PARTITION BY toYYYYMMDD(min_time)
SETTINGS index_granularity = 8192;
``` 
* Sample data for all these tables, use [clickhouse-obfuscator](https://github.com/ClickHouse/ClickHouse/blob/master/programs/obfuscator/Obfuscator.cpp#L42-L80) if necessary

slvrtrn:
A follow-up on https://github.com/metabase/metabase/pull/34873. Required to resolve https://github.com/ClickHouse/metabase-clickhouse-driver/issues/216.

### Description

`align-temporal-unit-with-param-type` lacked access to `value`, which in certain cases made it impossible to make a decision on which temporal unit we should choose for the substitution. The newly introduced `align-temporal-unit-with-param-type-and-value` is intended to be a direct replacement.

### How to verify

The default behavior is unchanged and there are core tests to verify that. "Demo" tests were updated accordingly. As `align-temporal-unit-with-param-type` is marked as deprecated, the drivers implementing an override for it (probably ClickHouse only?) will get a warning, and there is a mention of the new method in the code, so it should be clear how to proceed.

### Checklist


src/metabase/driver/sql/parameters/substitution.clj
```diff
@@ -89,19 +89,34 @@
...
-  [_driver _field param-type]
+(defmethod align-temporal-unit-with-param-type-and-value :default
+  [_driver _field param-type _value]
   (when (params.dates/date-type? param-type)
     :day))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
+(defmethod align-temporal-unit-with-param-type :default
+  [driver _field param-type]
+  (align-temporal-unit-with-param-type-and-value driver param-type nil))
```
metamben, line 119: These seem to be wrong way around. If a driver implemented `align-temporal-unit-with-param-type` but not yet `align-temporal-unit-with-param-type-and-value`, then it wouldn't get called. For the time being, the default `align-temporal-unit-with-param-type-and-value` has to forward to `align-temporal-unit-with-param-type`.

slvrtrn: Fixed, thanks! :+1:

src/metabase/driver/sql/parameters/substitution.clj
```diff
@@ -90,18 +90,34 @@
...
+  driver/dispatch-on-initialized-driver
+  :hierarchy #'driver/hierarchy)
+
+#_{:clj-kondo/ignore [:deprecated-var]}
 (defmethod align-temporal-unit-with-param-type :default
   [_driver _field param-type]
   (when (params.dates/date-type? param-type)
     :day))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
```
metamben, line 117: This is superfluous, isn't it?

slvrtrn: Without:
image

metamben: Doesn't it work on the `align-temporal-unit-with-param-type` call?

slvrtrn: Ah, yeah, can be moved a few lines down if necessary.

metamben: Please do that, otherwise it's a bit confusing.

slvrtrn: Done.

docs/developers-guide/driver-changelog.md
```diff
@@ -21,6 +21,10 @@ title: Driver interface changelog
   an additional argument: `database`.
   The new function arglist is `[driver database connection schema-inclusion-filters schema-exclusion-filters]`.
 
+- The method `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type` is now deprecated.
+  Use `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type-and-value` instead,
+  which has the access to `value` and intended to provide more flexibility on choosing the right conversion unit.
```
metamben, line 26:
```suggestion
  which has access to `value` and therefore provides more flexibility for choosing the right conversion unit.
```

metamben (approved):

darksciencebase:
@metamben are we backporting this?

metamben:
> @metamben are we backporting this?

No, this is planned for 49.

PR #37817, opened 2024-01-17
Fix salt in pulse unsub hash

### LINKED ISSUE #37916: Dashboard unsubscription links are not secure and will break on upgrade
### Describe the bug

Dashboard subscription messages to non-Users includes links to unsubscribe. The links contain a hash, which is meant to be salted using a secret uuid unique to each site. 

There is a bug in the we calculate the hash, with two issues:

1. The salt will change whenever the site is upgrade, breaking previously sent links.
1. The salt will not vary between different servers, and depends only on the release version.

### To Reproduce

1. Add a dashboard subscription for someone who is not a User.
2. Trigger a subscription message.
3. Upgrade to a new release
4. Click the unsubscribe link from the message sent in (2)
5. 😢 

I demonstrate the issue in the following video:

https://www.loom.com/share/dfaca78662514791876b95c49555a270

### Expected behavior

Links from previous messages continue to be valid after server upgrades.

### Logs

_No response_

```text
- Metabase version: 0.47.0 and above
```

### Severity

This is just an annoyance, after the next pulse the user can unsubscribe using the latest link. No users have reported being affected. The security aspect is minor, at worst (and with a lot of difficulty) someone could unsubscribe other users.

### Additional context

_No response_

crisptrutski:
Closes https://github.com/metabase/metabase/issues/37916

### Description

While looking at the `site-uuid` related code, I noticed that this setting was never being read. It looks like we're accidentally using the mangled var name in the relevant payload instead.

I don't think we can merge this as is, as I believe it would break unsubscription links out in the wild. I'm not sure how best we would handle this - perhaps being lenient and accepting both hashes. 

### How to verify

Executing the previous code in the REPL:

```clojure
(json/generate-string 
  {:salt public-settings/site-uuid-for-unsubscribing-url
   :email email
   pulse-id pulse-id}))))
                           
 "{\"salt\":\"metabase.models.setting$setting_fn$setting_getter_STAR___62091@3b698184\",\"email\":\"email\",\"pulse-id\":\"pulse-id\"}"
```

### Checklist

- [x] Confirm this is something worth fixing
- [x] Confirm whether we want to backport
- [x] ~~Make a plan for backwards compatibility, if necessary~~ Not worth it.

### DIFF
--- a/test/metabase/api/session_test.clj
+++ b/test/metabase/api/session_test.clj
@@ -569,6 +569,11 @@
 
 ;;; ------------------------------------------- TESTS FOR UNSUBSCRIBING NONUSERS STUFF --------------------------------------------
 
+(deftest unsubscribe-hash-test
+  (with-redefs [public-settings/site-uuid-for-unsubscribing-url "08534993-94c6-4bac-a1ad-86c9668ee8f5"]
+    (is (= "691501b28502126a6f59636ce147d26afa9d8fe8e5d0bd41f0de609194c0d9ba0042276d9b23cf10b1c7a4b18b3f946707e511d3e95e08fbe2c456080f619e36"
+           (messages/generate-pulse-unsubscribe-hash :pulse-id "rasta@pasta.com")))))
+
 (deftest unsubscribe-test
   (reset-throttlers!)
   (testing "POST /pulse/unsubscribe"
--- a/test/metabase/api/session_test.clj
+++ b/test/metabase/api/session_test.clj
@@ -570,19 +570,19 @@
 ;;; ------------------------------------------- TESTS FOR UNSUBSCRIBING NONUSERS STUFF --------------------------------------------
 
 (deftest unsubscribe-hash-test
-  (let [email         "rasta@pasta.com"
-        pulse-id      12345678
-        expected-hash "37bc76b4a24279eb90a71c129a629fb8626ad0089f119d6d095bc5135377f2e2884ad80b037495f1962a283cf57cdbad031fd1f06a21d86a40bba7fe674802dd"]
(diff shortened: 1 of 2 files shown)


crisptrutski:
Not sure how long these unsubscribe links "live" in practice - if users only care about the one included in the latest message, I don't think we need to preserve backwards compatibility, at at the very least could afford to remove it in a later release.

crisptrutski:
On further thought, I suspect we break backwards compatibility quite common as the symbols at the end of the munged name will not be stable.

test/metabase/api/session_test.clj
```diff
@@ -569,6 +569,11 @@
 
 ;;; ------------------------------------------- TESTS FOR UNSUBSCRIBING NONUSERS STUFF --------------------------------------------
 
+(deftest unsubscribe-hash-test
+  (with-redefs [public-settings/site-uuid-for-unsubscribing-url (constantly "08534993-94c6-4bac-a1ad-86c9668ee8f5")]
+    (is (= "691501b28502126a6f59636ce147d26afa9d8fe8e5d0bd41f0de609194c0d9ba0042276d9b23cf10b1c7a4b18b3f946707e511d3e95e08fbe2c456080f619e36"
+           (messages/generate-pulse-unsubscribe-hash :pulse-id "rasta@pasta.com")))))
```
calherries, line 575: This tests that the hash is stable, but doesn't test the security-related property that the `generate-pulse-unsubscribe-hash` return value depends on `site-uuid-for-unsubscribing-url`. What do you think about adding a test for that?

crisptrutski: Added, along with rebasing on master so we can stub out read-only settings more idiomatically.

calherries (approved):
Extra nice

PR #37822, opened 2024-01-17
Remove unused clojure release code 🥳 

iethree:
closes https://github.com/metabase/metabase-private/issues/175

### Description

Our new release script has been working in CI for a while now and has been validated on both major and minor releases. We also have a way of releasing in CI and locally with the new code, so I think we can safely delete the old clojure code that we're not using anymore.

### How to verify

`./bin/build.sh` should still work


escherize (approved):
LGTM. Thanks for pulling me in to pair on this.

PR #37841, opened 2024-01-18
Support legacy metadata and raw JS Dates for brush filter helper functions

### LINKED ISSUE #37304: [MLv2] [BE] [Bug] Brush filter helpers do not work as expected
**Context**

Brush filter helpers needs some modifications to better align with the FE code.

### `updateTemporalFilter`

1. it would be nice to support timestamps for start and end instead of ISO date string
2. it should be possible to pass `DatasetColumn` as it's done for [`available_drill_thrus`](https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/drills.ts#L18)
3. `dataset_query.query.filter` has an invalid field ref
image

and seems it appears after applying the method
Screenshot 2024-01-05 at 19 48 55

`"41"` should be `41` - number

### `updateNumericFilter`

1. it should be possible to pass `DatasetColumn` as it's done for [`available_drill_thrus`](https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/drills.ts#L18)
2. `dataset_query.query.filter` has an invalid field ref

### `updateLatLonFilter` 

1. it should be possible to pass `DatasetColumn` as it's done for [`available_drill_thrus`](https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/drills.ts#L18)

camsaul:
Resolves #37304

- Support legacy metadata for the JS versions of brush filter helper functions
- Support raw JS `Date` for the temporal brush filter helper. Our Cljs support for working with raw `Date`s is not very good, all of our tooling is around either ISO-8601 strings or Moment.js; I wrote a little bit of helper code to support `Date`s in our `metabase.shared.util.time` namespace and added a function to format as ISO-8601 based on effective type or base type. There's definitely a lot more work here to be done around this tooling -- in the future I think we should probably let you use a raw `Date` anywhere you can use a temporal literal. In the JVM, we already support `java.time` classes as temporal literals
- Updated Shadow Cljs versions because they were giving me problems locally


.dir-locals.el
```diff
@@ -88,10 +88,6 @@
   (cider-default-cljs-repl . shadow-select)
   (cider-shadow-default-options . "app"))
 
- ("shared"
```
camsaul, line 91: This folder was removed a while back

src/metabase/shared/util/internal/time.cljs
```diff
@@ -382,7 +387,8 @@
     (let [[t value-type] (iso-8601->moment+type t)
           t              (truncate t unit)]
       (moment+type->iso-8601 [t value-type]))
-    (.startOf ^moment/Moment t (name unit))))
+    (let [t (->moment t)]
+      (.startOf ^moment/Moment t (name unit)))))
```
metamben, line 391: I think it's more natural to hint when the name is introduced, not when it's used.
```suggestion
    (let [^moment/Moment t (->moment t)]
      (.startOf t (name unit)))))
```

src/metabase/shared/util/internal/time.cljs
```diff
@@ -391,4 +397,21 @@
     (let [[t value-type] (iso-8601->moment+type t)
           t              (add t unit amount)]
       (moment+type->iso-8601 [t value-type]))
-    (.add ^moment/Moment t amount (name unit))))
+    (let [t (->moment t)]
+      (.add ^moment/Moment t amount (name unit)))))
```
metamben, line 401:
```suggestion
    (let [^moment/Moment t (->moment t)]
      (.add t amount (name unit)))))
```

metamben (approved):
LGTM

PR #37846, opened 2024-01-18
Add unique constraint on databasechangelog if it's not existed

### LINKED ISSUE #37552: Missing databasechangelog index for new instances
In https://github.com/metabase/metabase/pull/34400 we split the migration into 2 files, for new instances the migration should only go through everything in `migrations/001_update_migrations.yaml` in which it has a `v00.00-000` migration that contains a SQL file that used to contains all migration from before 45.

It was an oversight that the initialization sql doesn't include custom changes we made to databasechangelog table, because it's supposed to be managed by liquibase itself.

There is a "nice-to-have" index `idx_databasechangelog_id_author_filename` that we added in [migration id](https://github.com/metabase/metabase/blob/e66c0754cace99451ba1bdc85ef50335ca8dc584/resources/migrations/000_legacy_migrations.yaml#L5562C9-L5562C9) `95` a while ago. This index was not included in the new migration file so new instances won't have this index.

As described in the migration itself, it's nice to have this index and we should add it back. 

Couple of notes:
- we should add this index back and backport it to 48
- the migration need to check whether this index existed before in case old instances try to upgrade

qnkhuat:
Closes #37552

Preconditions for this migration:
- [x] skip if have duplicated rows
- [x] skip if the index existsed

### DIFF
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -4195,6 +4195,8 @@ databaseChangeLog:
                     nullable: true
                   remarks: 'Hash of normalized query, calculated in middleware.cache'
 
+  # this index was added in migration 95, but during our split migration work in #34400 we didn't include this index
+  # in the sql initialization, so we need to recreate one here for new instances running 48+
   - changeSet:
       id: v48.00-067
       author: qnkhuat
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -4204,14 +4204,24 @@ databaseChangeLog:
         # If we're dumping the migration as a SQL file or trying to force-migrate we can't check the preconditions
         # so just go ahead and skip the entire thing. This is a non-critical migration
         - onUpdateSQL: IGNORE
-        - and:
+        - or:
           - sqlCheck:
               expectedResult: 0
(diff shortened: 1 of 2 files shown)


piranha:
Do you think it's better to skip creating the constraint instead of removing duplicate rows?

qnkhuat:
It's a risky business I think, if they have duplicated rows, there is probably something wrong during the migration, and to correctly identify the row we need to delete, we need to be able to calculate the desired hash of the "correct" migrations and delete the wrong ones so that on startup they don't run into check sum error.

Also considering this shouldn't happen in real life so I think the number of instances have this problem should be very vey small.

piranha:
Yeah... but would be nice to have some way out of that. I wonder if we should warn them somehow?.. I.e. if there are duplicate migrations, should the instance owner be worried? Or in general it's "meh" so we can ignore that and just make life better for most of the people ignoring others? 😁

qnkhuat:
according to the original migration 95, it's a "nice-to-have" migration https://github.com/metabase/metabase/commit/d6ad914cf8c9a126118cd12fc5963a567388f585

I thought I referenced this in the migrations but turns out I didn't. added a comment :)

sansolovyov (approved):
Well, ok then, let's deal with issues as they occur :)

Issue #37801, opened 2024-01-17
site-uuid changes 


dpsutton:
### Describe the bug

The site-uuid can change.
This has downstream effects that break model caching. We use the site-uuid to form the name of the schema where cached models will go. Since this changes model caching will begin to fail since the new schema it expects to find will not be present.

### To Reproduce

There are two ways that I know of to reproduce this. Each reproduction is using a recent jar from master.

#### Restart a single instance
Start the instance

```shell
 MB_DB_CONNECTION_URI="postgres://[details]/single-race" MB_JETTY_PORT=3008 java "$(socket-repl 6008)" -jar metabase.jar
```

And from a repl:

```clojure
/tmp/single-restart via ☕ v17.0.1 on ☁️  metabase-query
❯ netcat localhost 6008
user=> (metabase.public-settings/site-uuid)
"e3b1bf65-89af-4657-961b-e1c1c0333bb9"
user=> %

# restart the metabase process at this point
/tmp/single-restart via ☕ v17.0.1 on ☁️  metabase-query took 28s
❯ netcat localhost 6008
user=> (metabase.public-settings/site-uuid)
"8e1f40a8-5f7a-40d6-8f20-64537c56ba5a"
user=>
```

And observe that the site-uuid is changed

#### An instance joins the cluster

Run one instance

```shell
MB_DB_CONNECTION_URI="postgres://[details]race" MB_JETTY_PORT=3006 java "$(socket-repl 6006)" -jar metabase.jar
```

and check it's site-uuid

```clojure
persisted-cache=> (public-settings/site-uuid)
"722dd9b2-347d-498a-9c2f-aa10ec006c10"
```

Start another instance on the same app-db

```shell
MB_DB_CONNECTION_URI="postgres://[details]/race" MB_JETTY_PORT=3007 java "$(socket-repl 6007)" -jar metabase.jar
```

Check it's site-uuid

```clojure
persisted-cache=> (public-settings/site-uuid)
"d421b9b9-c39a-49ee-a64d-7e4cb4ef14a2"
```

And now ask the original it's site-uuid

```clojure
;; before other instance joined the cluster
persisted-cache=> (public-settings/site-uuid)
"722dd9b2-347d-498a-9c2f-aa10ec006c10"
;; after other instance joined the cluster
persisted-cache=> (public-settings/site-uuid)
"d421b9b9-c39a-49ee-a64d-7e4cb4ef14a2"
```

### Expected behavior

site-uuid should never change

[...]

PR #37849, opened 2024-01-18
Improve safety of lazily initialized settings

crisptrutski:
Refs https://github.com/metabase/metabase/issues/37801

### Description

This change replaces the "calling raw setter within getter" pattern with an `:init` hook, to prevent some foot guns and race conditions.

### How to verify

Here is the issue without this change:

```shell
 MB_DB_CONNECTION_URI="postgres://[details]/single-race" MB_JETTY_PORT=3008 java "$(socket-repl 6008)" -jar metabase.jar
```

And from a repl:

```clojure
/tmp/single-restart via ☕ v17.0.1 on ☁️  metabase-query
❯ netcat localhost 6008
user=> (metabase.public-settings/site-uuid)
"e3b1bf65-89af-4657-961b-e1c1c0333bb9"
user=> %

# restart the metabase process at this point
/tmp/single-restart via ☕ v17.0.1 on ☁️  metabase-query took 28s
❯ netcat localhost 6008
user=> (metabase.public-settings/site-uuid)
"8e1f40a8-5f7a-40d6-8f20-64537c56ba5a"
user=>
```

Observe that the UUID has changed. With this change it will remain the same across both processes.

### Checklist

- [x] Prevent uuid being initialized during validation.
- [x] Unit tests
- ~~Add integration test showing persistence of uuid across app start-ups~~ (left for a later PR)


test/metabase/models/setting_test.clj
```diff
@@ -142,14 +149,40 @@
...
+  (#'setting/db-or-cache-value custom-init-setting)
+  )
+
+(deftest custom-init-test
+  ;; Some other pre-existing test is initializing this value, this is a stop gap until we fix its hygiene
+  (#'setting/set! custom-init-setting nil :bypass-read-only? true)
+
+  (let [get-via-db #(#'setting/db-or-cache-value :test-setting-custom-init)]
+    (testing "The value will be initialized and saved"
+      (mt/discard-setting-changes [test-setting-custom-init]
```
crisptrutski, line 174: It's quite tedious adding this, I wonder if we couldn't instead add an `:each` fixture here which surrounds each test with a roll-back on all the settings in the namespace. We'd need to break up some of the tests for this to work, but I think it has a certain elegance.

crisptrutski:
@calherries what do you think about backporting? just worried about the size and scope.

calherries:
Since it's a safety improvement, there's not a lot of value backporting this. I would abandon if there's a merge conflict or two

crisptrutski:
All the `bb-tests-{driver}` tests are failing with some variation of the following:

```
FAIL in metabase.api.search-test/search-db-call-count-test (search_test.clj:907)

with temporary :model/Card with attributes
{:creator_id 1,
 :database_id 1,
 :dataset_query {},
 :display :table,
 :name "card db 1 JGBVOBTNVFZCZRBUPEPU",
 :visualization_settings {},
 :created_at #t "2024-01-18T16:29:41.916921Z[UTC]",
 :updated_at #t "2024-01-18T16:29:41.916923Z[UTC]"

...
 
with temporary :model/Segment with attributes
{:creator_id 1,
 :definition {},
 :description "Lookin' for a blueberry",
 :name "segment 3 JGBVOBTNVFZCZRBUPEPU",
 :table_id 1196,
 :created_at #t "2024-01-18T16:29:42.014008Z[UTC]",
 :updated_at #t "2024-01-18T16:29:42.014009Z[UTC]"}


expected: 9
  actual: 56
```

crisptrutski:
I think the common thread is actually the bulk of the `ee` tests failing, e.g. this [non driver one](https://github.com/metabase/metabase/actions/runs/7574901829/job/20630342432?pr=37849).

src/metabase/models/setting.clj
```diff
@@ -526,7 +576,9 @@
          source-fns [user-local-value
                      database-local-value
                      env-var-value
-                     db-or-cache-value
+                     (if *disable-init*
+                       db-or-cache-value
+                       db-or-cache-or-init-value)
```
calherries, line 581: this would do the same thing, but doesn't require an extra `db-or-cache-or-init-value` function
```suggestion
                     db-or-cache-value
                     (when *disable-init*
                       init!)
```

crisptrutski: So much nicer! Classic inverse typo - `s/when/when-not` though 😄

src/metabase/models/setting.clj
```diff
@@ -527,9 +570,11 @@
                      database-local-value
                      env-var-value
                      db-or-cache-value
+                     (when-not *disable-init*
+                       init!)
                      default-value]]
```
calherries, line 575: Since it's invalid to provide both an initialized value and a default, maybe it makes sense for readability (and therefore safety) to have the code reflect they are mutually exclusive here. For example:
```suggestion
                     (cond 
                       (:default setting) default-value
                       (:init setting) init!))]]
```

src/metabase/models/setting.clj
```diff
@@ -467,28 +470,68 @@
 
 (def ^:private ^:dynamic *disable-cache* false)
 
+(def ^:private ^:dynamic *disable-init* false)
+
+(declare get)
+
```
calherries, line 476: We generally don't add newlines between `declare` statements.
```suggestion
```

src/metabase/models/setting.clj
```diff
@@ -467,28 +470,68 @@
...
+(defn- db-value [setting-definition-or-name]
+  (t2/select-one-fn :value Setting :key (setting-name setting-definition-or-name)))
+
+(defn- call-me-maybe [f]
+  (when f (f)))
+
+(defn- db-is-set-up? []
+  ;; this should never be hit. it is just overly cautious against a NPE here. But no way this cannot resolve
+  (call-me-maybe (requiring-resolve 'metabase.db/db-is-set-up?)))
```
calherries, line 497: Maybe you have reasons for changing this, but I found it clearer as it was written before, mostly because it doesn't use a newly created function. Obviously subjective but my taste is to inline new short functions like this because the overhead of reading `call-me-maybe` and wondering what it does is not worth the abstraction it provides.
```suggestion
(def ^:private db-is-set-up?
  (or (requiring-resolve 'metabase.db/db-is-set-up?)
      ;; this should never be hit. it is just overly cautious against a NPE here. But no way this
      ;; cannot resolve
      (constantly false)))
```

src/metabase/models/setting.clj
crisptrutski: Nice, this also makes the behavior safer until we add schema validation.

calherries: Oops, I left out the `*disable-init*` from the suggestion but you still need it. Something like
```clojure
                     (cond 
                       (:default setting) default-value
                       (and (:init setting) (not *disable-init*)) init!))
```

src/metabase/models/setting.clj
crisptrutski: Just a matter of preference. I like the one liner, but I'll revert if you think it's obtuse. Maybe metabase.utils just needs a better named version of this helper - or has one already!

crisptrutski: Oh, actually this wasn't just preference - I need to put this inside another function to avoid a circular dependency. This ns is a transitive dependency of metabase.db.

calherries: Ah understood. In that case here's another suggestion. Feel free to leave it though
```
(cond-> (requiring-resolve 'metabase.db/db-is-set-up?) apply)
```

crisptrutski: Somehow `apply` was the function I couldn't find - forgot it had a 1-arity 😆

test/metabase/models/setting_test.clj
```diff
@@ -142,14 +155,54 @@
...
+    (is (some? (setting/user-facing-value custom-init-setting)))))
 
 (deftest do-not-define-setter-function-for-setter-none-test
   (testing "Settings with `:setter` `:none` should not have a setter function defined"
     (testing "Sanity check: getter should be defined"
       (is (some? (resolve `test-setting-calculated-getter))))
     (is (not (resolve `test-setting-calculated-getter!)))))
 
+(defn- get-setting-via-db []
+  (#'setting/read-setting :test-setting-custom-init))
```
calherries, line 170: I'm struggling to find the benefit of defining this function. Again this is subjective but I've noticed you tend to create extra little functions and I prefer to inline them for readability when DRYing doesn't bring any benefits. Up to you again but I wouldn't be so scared of duplication here.

test/metabase/models/setting_test.clj
```diff
@@ -15,10 +17,14 @@
    [metabase.util :as u]
    [metabase.util.encryption-test :as encryption-test]
    [metabase.util.i18n :as i18n :refer [deferred-tru]]
+   [metabase.util.log :as log]
    [toucan2.core :as t2]))
 
 (use-fixtures :once (fixtures/initialize :db))
 
+;; TODO: it would be nice to detect and discard the changes to *all* settings defined in this namespace
+(use-fixtures :each (fn [thunk] (mt/discard-setting-changes [:test-setting-custom-init] (thunk))))
```
calherries, line 26: `use-fixtures` is used sparingly for a good reason: it's difficult to notice it's happening when reading the test in isolation. This fixture is only helpful for a few tests, so IMO it doesn't cross the threshold. It's okay and preferable to duplicate this many times.

test/metabase/models/setting_test.clj
```diff
@@ -142,14 +155,54 @@
...
 
+(defn- get-setting-via-db []
+  (#'setting/read-setting :test-setting-custom-init))
+
+;; TODO: I suspect we're seeing stale values in CI due to parallel tests or state persisting between runs
+;;  We should at least make this an error when running locally without parallelism.
+(defn- clear-setting-if-leak! []
+  (when-let [existing-value (some? (get-setting-via-db))]
+    (log/warn "Test environment corrupted, perhaps due to parallel tests or state kept between runs" existing-value)
+    (setting/set! :test-setting-custom-init nil :bypass-read-only? true)))
```
calherries, line 177: That's weird. I don't know the answer to this but thanks for leaving a TODO

test/metabase/models/setting_test.clj
```diff
@@ -142,14 +155,54 @@
...
+      (is (= val (get-setting-via-db))))))
+
+(deftest validate-without-initialization-test
+  (testing "Validation does not initialize the setting"
+    (clear-setting-if-leak!)
+    (setting/validate-settings-formatting!)
+    (is (= nil (get-setting-via-db)))))
+
+(deftest init-requires-db-test
+  (testing "We will fail instead of implicitly initializing a setting if the db is not ready"
```
calherries, line 194: To keep this test isolated, you could set `test-setting-custom-init`'s value to `nil` at the beginning. Other tests should be using `mt/discard-settings-changes` but better to be safe than sorry

calherries (approved):
Looks solid from a code standpoint, I just have a few tiny suggestions. It would be great if you could document the new behaviour so the next person can recover the intention behind it

src/metabase/models/setting.clj
crisptrutski: 1. The `cond->` needs a separate clause predicate, so I think this would end up like:

```clojure
(let [f (requiring-resolve 'metabase.db/db-is-set-up?)]
  (cond-> f f apply))
```

There is `mu/maybe?` which is almost a shorthand for removing the repetition, but it'll return true if `f` is nil.

2. It turns out that `apply` does *not* have a 1-arity 😢 - so this would be `(cond-> f f (apply nil))`, taking even more mental gymnastics to read IMHO.

Personally I find using a helper function clearer - would giving it a better name and sticking it in utils perhaps assuage you? Something like `mu/call-if-some`. Other name ideas are `apply-maybe` or `some-apply`.

For the time being I'm just going to inline to helper at the two call sites 😄

test/metabase/models/setting_test.clj
crisptrutski: In this particular case I think the helper method was created initially to encapsulate fiddling with a dynamic var, and is now a bit vestigial that the `read-setting` method was added. I'll just inline that.

On your general observation - I'm a believer in the the top-down / declarative / single-level-of-abstraction / small methods school of "clean" code. I take it that you subscribe to a more procedural / details clearly exposed style. The both have their merits, and based on familiarity one is either more comfortable drilling between methods or parsing dense code chunks. I agree that "coincidental repetition" should not be duplicated, but otherwise I err on the side of small composable atomic things. For me part of the appeal of FP is that large sections of the code can be self describing aliases essentially, that should never need to change, and might simply get deleted one day.

I'll assimilate the local style as I get more familiar with it - consistency often beats personal taste 😄

test/metabase/models/setting_test.clj
crisptrutski: As stated in my comment, I see this fixture as becoming a "hygiene for internal settings" block that applies to all the tests. With the current helpers that would have two big issues:

1. All the settings would need to be explicitly enumerated for the macro.
2. All the settings would be captured and re-set, even if they weren't touched.

Even with a single setting, repeating the block across 4 tests feels brittle to me, but I'll go with your suggestion for now. Gets rid of a TODO 😆

src/metabase/models/setting.clj
```diff
@@ -467,28 +470,67 @@
 
 (def ^:private ^:dynamic *disable-cache* false)
 
+(def ^:private ^:dynamic *disable-init* false)
```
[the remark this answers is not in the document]
crisptrutski: My preference is to make `get` build on `read-setting`, rather than the other way around. This way we can just get rid of the dynamic var, which I agree is a big smell. I've updated the docstring that you pointed out and going to leave it at that for now.

Issue #37296, opened 2024-01-05
BE flaky test metabase.sync.analyze.fingerprint-test/refingerprint-fields-for-db!-test


calherries:
metabase.sync.analyze.fingerprint-test/refingerprint-fields-for-db!-test

failed for sparksql [here](https://github.com/metabase/metabase/actions/runs/7409351630/job/20159456580?pr=37110#step:4:13183), and took 30+ minutes to run

tsmacdonald:
After some false attempts on the fingerprinting side, it seems clear to me that it's an issue with running out of connections (see, for example, [this run](https://github.com/metabase/metabase/actions/runs/7400215362/job/20133420825#step:14:13061), which has nothing to do with fingerprinting; there are a few more too).

So I've transferred it to @metabase/core-backend-qp ; see discussion on Slack [here](https://metaboat.slack.com/archives/C04DN5VRQM6/p1706026472498399)

PR #37853, opened 2024-01-18
Fix one flaky [SparkSQL?] fingerprinting test

tsmacdonald:
This started because of issue #37296, which it does not solve. But while investigating that I realized I could get `metabase.sync.analyze.fingerprint-test/fingerprint-test` to fail pretty frequently while running the ns from the command line. It seems to assume that `fingerprint-fields!` has been called, and usually it *had* been called due to dirty testing state, but on my computer the race was finishing the wrong way.

### DIFF
--- a/src/metabase/driver/ddl/interface.clj
+++ b/src/metabase/driver/ddl/interface.clj
@@ -15,7 +15,7 @@
 
 (mu/defn schema-name :- ::lib.schema.common/non-blank-string
   "Returns a schema name for persisting models. Needs the database to use the db id and the site-uuid to ensure that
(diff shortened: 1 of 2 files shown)


tsmacdonald:
un-drafting this so CI runs, but it's not ready for review: I'm just hoping for some debug output.

sansolovyov (approved):
It still can't be run in parallel even given `fingerprint-fields!` is called?

tsmacdonald:
> It still can't be run in parallel even given fingerprint-fields! is called?

The linter said no. `fingerprint-fields!` changes DB state, so I think that makes sense

PR #37854, opened 2024-01-18
Rework findColumnIndexForColumnSetting to MLv2

### LINKED ISSUE #37677: [MLv2] [FE] findColumnIndexForColumnSetting and findColumnSettingIndexForColumn
### findColumnIndexForColumnSetting

https://github.com/metabase/metabase/blob/588ce1a8d8c9ed33d04cf06c946ee7e66cd4f43e/frontend/src/metabase-lib/queries/utils/dataset.js#L26

Existing signature with types:

```
findColumnIndexForColumnSetting(
  columns: DatasetColumn[], 
  columnSetting: TableColumnOrderSetting
): number;
```

Existing implementation:

```
// NOTE: need to normalize field refs because they may be old style [fk->, 1, 2]
const fieldRef = normalizeFieldRef(columnSetting.fieldRef);
// first try to find by fieldRef
if (fieldRef != null) {
  const dimension = Dimension.parseMBQL(fieldRef);
  const index = _.findIndex(columns, col =>
    dimension.isSameBaseDimension(fieldRefForColumn(col)),
  );
  if (index >= 0) {
    return index;
  }
}

// if that fails, find by column name
return _.findIndex(columns, col => col.name === columnSetting.name);
```

In MLv2 it could be:

```
// first try to find by fieldRef
if (columnSetting.fieldRef != null ) {
  const [columnIndex] = Lib.findColumnIndexesFromLegacyRefs(
      query,
      -1,
      columns,
      [columnSetting.fieldRef],
  );
  if (columnIndex >= 0) {
    return columnIndex;
  };
}

// if that fails, find by column name
return _.findIndex(columns, column => column.name === columnSetting.name);
```

`query` is required to handle aggregation references correctly, so unfortunately it should be passed to places where the function is used.

### findColumnSettingIndexForColumn

https://github.com/metabase/metabase/blob/588ce1a8d8c9ed33d04cf06c946ee7e66cd4f43e/frontend/src/metabase-lib/queries/utils/dataset.js#L43

Existing signature with types:

```
findColumnSettingIndexForColumn(
  columnSettings: TableColumnOrderSetting[], 
  column: DatasetColumn,
): number;
```

Existing implementation:

```
const fieldRef = normalizeFieldRef(fieldRefForColumn(column));
if (fieldRef == null) {
  return columnSettings.findIndex(
    columnSetting => columnSetting.name === column.name,
  );
}

const index = columnSettings.findIndex(columnSetting =>
  _.isEqual(fieldRef, normalizeFieldRef(columnSetting.fieldRef)),
);

return index;
```

In MLv2 it could be:

```
// ignore settings without fieldRef but preserve indexes
const items = columnSettings.reduce((items, { fieldRef }, settingIndex) => {
  if (fieldRef != null) {
    items.push({ fieldRef, settingIndex });
  }
  return items;
}, items);

// first try to find by fieldRef
const itemIndexes = Lib.findColumnIndexesFromLegacyRefs(
  query,
  -1,
  [column],
  items.map(({ fieldRef }) => fieldRef),
);
const itemIndex = itemIndexes.find(index => index >= 0);
if (itemIndex != null) {
  return items[itemIndex].settingIndex;
};

// if that fails, find by column name
return _.findIndex(columnSettings, columnSetting => columnSetting.name === column.name);
```

### Note!

Generally `question` or its `query` is out of sync with query results in viz. If you need `query` to be used with dataset columns coming from the query response, you need to construct a `Question` based on `card` available in `series`, like here https://github.com/metabase/metabase/pull/37634/files#diff-4b16a6cc5fc357af60212f45a4c9074fa08211383842c3e120ae974c80a4f3c0R322. Otherwise in some edge cases, e.g. when the query is re-running you would use an incorrect `query`. It could be OK in some cases, but be aware that `fromLegacyColumn` throws if it cannot find the aggregation column that corresponds to the dataset column, so always construct a query from query results like in the PR above.

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/37677

This is an umbrella PR, other changes related to the problem will be merged here

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/38171
- [ ] https://github.com/metabase/metabase/pull/38170
- [ ] https://github.com/metabase/metabase/pull/38169
- [ ] https://github.com/metabase/metabase/pull/38172
- [ ] https://github.com/metabase/metabase/pull/38168
- [ ] https://github.com/metabase/metabase/issues/38222
```

### Description

Migrate `findColumnIndexForColumnSetting` to MLv2 with a fallback to the old method.
Throwing an error is added to find all places where `query` is not passed, temporary


frontend/src/metabase-lib/queries/utils/dataset.ts
```diff
@@ -31,27 +32,56 @@ export function normalizeFieldRef(fieldRef: FieldReference) {
 export function findColumnIndexForColumnSetting(
   columns: DatasetColumn[],
   columnSetting: TableColumnOrderSetting,
+  query?: Lib.Query,
 ) {
   const fieldRef = columnSetting.fieldRef;
   // NOTE: need to normalize field refs because they may be old style [fk->, 1, 2]
+  // TODO: use import { normalize } from "metabase-lib/queries/utils/normalize";
```
kamilmielnik, line 39: Is this going to be handled here or in a subtask?

frontend/src/metabase-lib/queries/utils/dataset.ts
```diff
@@ -31,27 +32,56 @@ export function normalizeFieldRef(fieldRef: FieldReference) {
...
-          dimension.isSameBaseDimension(fieldRefForColumn(col)),
-        )
-      : -1;
+    let columnIndex: number;
 
-    if (index >= 0) {
-      return index;
+    if (!query) {
+      // TODO: remove once migration is completed
+      throw new Error("query is required to find column index");
```
kamilmielnik, line 47: Is this entire block going to be removed or just the `throw`?
I'm trying to understand is there anything that prevents us from doing a complete migration.

frontend/src/metabase-lib/queries/utils/dataset.ts
uladzimirdev: yes, I'll add it to the cleanup task

frontend/src/metabase-lib/queries/utils/dataset.ts
uladzimirdev: as we figured out during the pairing session, we need to double check with public questions where dataset_query is not available, if it provides correct results or not. but the plan is to remove `if(!query)` check completely

frontend/src/metabase/visualizations/components/ObjectDetail/ObjectDetailWrapper.tsx
```diff
@@ -12,12 +14,14 @@ export function ObjectDetailWrapper({
...
   dashcard,
   isObjectDetail,
-  ...props
+  ...rest
 }: ObjectDetailProps) {
   const [currentObjectIndex, setCurrentObjectIndex] = useState(0);
 
   // only show modal if this object detail was triggered via an object detail zoom action
   const shouldShowModal = isObjectDetail;
+  const getFallbackQuestion = () =>
```
ranquild, line 23: Do you know where `question` is not passed?

ranquild (approved):
Minor comments

PR #37856, opened 2024-01-18
Rework findColumnSettingIndexForColumn to MLv2

uladzimirdev:
Part of  https://github.com/metabase/metabase/issues/37677

### Description

Rework to MLv2

### How to verify

Test should pass

### Demo

This is why I updated the test 
https://www.loom.com/share/7b0c28c3afdb4643b0ea56e570babc8c

### Checklist


frontend/src/metabase-lib/queries/utils/dataset.ts
```diff
@@ -53,21 +54,32 @@ export function findColumnIndexForColumnSetting(
...
-      _.isEqual(normalizedFieldRef, normalizeFieldRef(columnSetting.fieldRef)),
+  // ignore settings without fieldRef but preserve indexes
+  const items = columnSettings.reduce<
+    { fieldRef: FieldReference; settingIndex: number }[]
+  >((items, { fieldRef }, settingIndex) => {
+    if (fieldRef != null) {
+      items.push({ fieldRef, settingIndex });
+    }
+    return items;
+  }, []);
```
kamilmielnik, line 69: This is a bit cleaner, does not need explicit typing and helps avoid a cast later on when calling `Lib.findColumnIndexesFromLegacyRefs` (we can just use `items.map(({ fieldRef }) => fieldRef)`):

```suggestion
  const items = columnSettings.flatMap((item, settingIndex) => {
    const fieldRef = item.fieldRef ? normalizeFieldRef(item.fieldRef) : null;
    return fieldRef ? [{ fieldRef, settingIndex }] : [];
  });
```

frontend/src/metabase-lib/queries/utils/dataset.ts
```diff
@@ -53,21 +54,32 @@ export function findColumnIndexForColumnSetting(
...
+    [column],
+    items.map(({ fieldRef }) => normalizeFieldRef(fieldRef) as FieldReference),
   );
 
-  return index;
+  const itemIndex = itemIndexes.find(index => index >= 0);
+
+  if (itemIndex != null) {
+    return items[itemIndex].settingIndex;
+  }
```
kamilmielnik, line 84: :x: Not all paths return a value

```suggestion
  }

  return -1;
```

frontend/src/metabase-lib/queries/utils/dataset.ts
uladzimirdev: [6f121a1](https://github.com/metabase/metabase/pull/37856/commits/6f121a1bd4dab41efac91abca2c06b082578b505)

kamilmielnik (approved):
:+1:

PR #37858, opened 2024-01-18
Refactor WidgetStatusIcon

oleggromov:
Part of the https://github.com/metabase/metabase/pull/37709, and a bigger epic https://github.com/metabase/metabase/issues/36524, this PR refactors `WidgetStatusIcon` component so that it's easier to use. 

A separate PR is needed because my original changes of the component broke many tests and I want to ensure this preparation step doesn't break anything.


nemanjaglumac (approved):
Green tests make me happy.
This is now much easier to read.

Thanks

frontend/src/metabase/parameters/components/WidgetStatusIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+type Props = {
+  name: "close" | "enter_or_return" | "empty" | "chevrondown";
+  onClick?: () => void;
+};
+
+export function WidgetStatusIcon({ name, onClick }: Props) {
+  const classes = cx(
+    "flex-align-right flex-no-shrink",
+    name === "close" && "cursor-pointer",
+  );
```
kamilmielnik, line 14: nit: 

```suggestion
  const classes = cx("flex-align-right flex-no-shrink", {
    "cursor-pointer": name === "close"
  });
```

Also we always just inline the call in JSX.

frontend/src/metabase/parameters/components/WidgetStatusIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  const handleOnClick = useCallback(
```
kamilmielnik, line 16: The hook is not necessary

frontend/src/metabase/parameters/components/WidgetStatusIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+    e => {
```
kamilmielnik, line 17: I'm not sure why TS let this one go, but `e` is `any` here.

```suggestion
    event: MouseEvent<HTMLImageElement | SVGElement>
```

PR #37862, opened 2024-01-18
[MLv2] Introduce a `queryDisplayInfo` abstraction

nemanjaglumac:
This PR introduces a FE-only abstraction called `queryDisplayInfo`.
Rather than doing another function overload, we're introducing a named export.

The original CLJS function accepts three parameters. The first two are always the query and the stage index. The third one is what we're trying to get the info about. In this case that is the query.

It's weird and cumbersome to always have to type `Lib.displayInfo(query, -1, query)`. That's the main motivation behind this abstraction.

### Usage
This function can be used in code to determine whether or not a query is editable and whether it is a native query.
```js
const { isEditable, isNative } = Lib.queryDisplayInfo(query);
```

### DIFF
--- a/frontend/src/metabase-lib/metadata.ts
+++ b/frontend/src/metabase-lib/metadata.ts
@@ -202,3 +202,7 @@ export function fromLegacyColumn(
 ): ColumnMetadata {
   return ML.legacy_column__GT_metadata(query, stageIndex, columnOrField);
 }
+
+export function queryDisplayInfo(query: Query): QueryDisplayInfo {
+  return displayInfo(query, -1, query);
+}
--- a/frontend/src/metabase-lib/metadata.ts
+++ b/frontend/src/metabase-lib/metadata.ts
@@ -140,18 +140,6 @@ declare function DisplayInfoFn(
   stageIndex: number,
   segment: SegmentMetadata,
 ): SegmentDisplayInfo;
-/**
- * Even though it seems weird to pass the same query two times,
- * this function follows the same pattern as the other displayInfo functions.
- * The first two parameters are always a query, and a stage.
- * The third parameter is what you would like to have info about.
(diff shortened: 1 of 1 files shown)


frontend/src/metabase-lib/metadata.ts
```diff
@@ -202,3 +202,7 @@ export function fromLegacyColumn(
 ): ColumnMetadata {
   return ML.legacy_column__GT_metadata(query, stageIndex, columnOrField);
 }
+
+export function queryDisplayInfo(query: Query): QueryDisplayInfo {
+  return displayInfo(query, -1, query);
```
kamilmielnik, line 207: We should not have 2 functions that do the same thing.
I'd remove the `QueryDisplayInfo`-related overload from `displayInfo`, use `ML.display_info` here directly, and move the comment about the first two arguments here.

frontend/src/metabase-lib/metadata.ts
```diff
@@ -202,3 +190,16 @@ export function fromLegacyColumn(
...
+ * Even though it seems weird to pass the same query two times,
+ * this function follows the same pattern as the other displayInfo functions.
+ * The first two parameters are always a query, and a stage.
+ * The third parameter is what you would like to have info about.
+ * It just only happens that the thing we're examining is (again) the query itself.
+ *
+ * Rather than adding another overload, we're introducing a special-case, named abstraction.
+ */
+export function queryDisplayInfo(query: Query): QueryDisplayInfo {
+  return ML.display_info(query, -1, query);
```
kamilmielnik, line 204: This comment applies to `ML.display_info` call, not to our `queryDisplayInfo`, so I'd move it inside.

Also:
`this function follows the same pattern as the other displayInfo functions` -> `this overload follows the same pattern as the other display_info overloads.`

kamilmielnik (approved):

PR #37867, opened 2024-01-18
[MLv2] Properly match aggregations across versions of a query

### LINKED ISSUE #37851: [MLv2] [BE] `findMatchingColumn` does not find matching column
It causes `Adding a breakout should not affect graph.metrics` unit test to fail in #37693.
These 2 usages are giving wrong result: https://github.com/metabase/metabase/pull/37693/files#diff-fdff338c9e6c27877274c27c438055055f0f9395a0a37c3fd360e194f6d2e461R635-R644 for "Count" and "Sum of Price" columns (they're not found when they should).

### To reproduce 

1. `git checkout sync-settings-query`
2. run `Adding a breakout should not affect graph.metrics` unit test in `frontend/test/metabase/lib/dataset.unit.spec.js` suite

----

It's also possibly causing these e2e tests to fail:
- `e2e/test/scenarios/visualizations-charts/waterfall.cy.spec.js`
    - `should show error for multi-series questions (metabase#15152)`
    - `should not allow you to choose X-axis breakout`

bshepherdson:
Because aggregation refs are based on `:source-uuid` they are not
comparable between queries, which breaks some logic to find columns
added and removed between two versions of a query.

This adds a fallback based on the newly added `:lib/source-name` option on
aggregation refs.

Fixes #37851.

### DIFF
--- a/src/metabase/lib/aggregation.cljc
+++ b/src/metabase/lib/aggregation.cljc
@@ -23,8 +23,9 @@
 (mu/defn column-metadata->aggregation-ref :- :mbql.clause/aggregation
   "Given `:metadata/column` column metadata for an aggregation, construct an `:aggregation` reference."
   [metadata :- lib.metadata/ColumnMetadata]
-  (let [options {:lib/uuid       (str (random-uuid))
-                 :effective-type ((some-fn :effective-type :base-type) metadata)}
+  (let [options {:lib/uuid        (str (random-uuid))
+                 :effective-type  ((some-fn :effective-type :base-type) metadata)
+                 :lib/source-name (:name metadata)}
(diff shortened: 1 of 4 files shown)


bshepherdson:
* `master`
  * **PR #37867** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37867?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37867?utm_source=stack-comment).

metamben (approved):
LGTM

PR #37873, opened 2024-01-18
[MLv2] Add `aggregation-column` and `breakout-column`

### LINKED ISSUE #37120: [MLv2] [BE] Add `breakoutColumn` and `aggregationColumn` APIs to simply columns retrieval from these clauses
```
// currently going from AggregationClause -> ColumnMetadata is non-trivial,
// so let's make a better API for this task
aggregationColumn(
  query: Query,
  stageIndex: number,
  clause: AggregationClause
): ColumnMetadata | null;

// same
breakoutColumn(
  query: Query,
  stageIndex: number,
  clause: BreakoutClause,
): ColumnMetadata
```

bshepherdson:
These are helpers for going from an aggregation or breakout
clause to the relevant `:metadata/column` they are based on.

Eg. the column being `SUM`'d by an aggregation or bucketed by a breakout.

(Nil for top-level aggregations like `:count`.)

Fixes #37120.

### DIFF
--- a/src/metabase/lib/aggregation.cljc
+++ b/src/metabase/lib/aggregation.cljc
@@ -13,6 +13,7 @@
    [metabase.lib.schema :as lib.schema]
    [metabase.lib.schema.aggregation :as lib.schema.aggregation]
    [metabase.lib.schema.common :as lib.schema.common]
+   [metabase.lib.schema.metadata :as lib.schema.metadata]
    [metabase.lib.temporal-bucket :as lib.temporal-bucket]
    [metabase.lib.types.isa :as lib.types.isa]
    [metabase.lib.util :as lib.util]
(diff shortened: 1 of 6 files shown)


bshepherdson:
* `master`
  * **PR #37873** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37873?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/37873?utm_source=stack-comment).

metamben (approved):
LGTM. Does this have to work only when the column is one level deep (i.e., not when we sum the double of column)?

bshepherdson:
> Does this have to work only when the column is one level deep (i.e., not when we sum the double of column)?

I would argue that "column" is not "field" - if one is doubling a column `A` that's an expression `E`. `SUM(E)`'s column of interest is `E`, not `A`, which is what gets returned here.

PR #37883, opened 2024-01-18
Uncomment audit-v2-views-test except the part that flakes

calherries:
Relates to this flake: https://github.com/metabase/metabase/issues/37434

This PR uncomments the previously commented ([PR](https://github.com/metabase/metabase/pull/37796)) test, but leaves the flaky part commented.

This is in response to Ngoc's comment [here](https://github.com/metabase/metabase/pull/37796#issuecomment-1897666340):

> I don’t think it’s safe to comment that test out. It was purposefully added to detect cases where a future migration changed and somehow broke the view

He's right about it not being safe to comment the whole test out. But it's still safe to comment the flaky part out, which is the part that tests the rollback is successful. This is relatively safe to comment because it is unlikely for changes to the rollback to make it past code review, and there is no change that can prevent them from failing in the future.

### DIFF
--- a/test/metabase/db/schema_migrations_test.clj
+++ b/test/metabase/db/schema_migrations_test.clj
@@ -540,7 +540,6 @@
         (testing "should drop the existing color column"
           (is (not (contains? (t2/select-one :model/Collection :id collection-id) :color))))))))
 
-#_ ;; TODO: this test is flaky so it's commented out for now
 (deftest audit-v2-views-test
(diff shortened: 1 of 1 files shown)


noahmoss (approved):
Thanks Cal! Looks like there's already another rollback test commented out in `migrate-field-database-type-test` due to a MySQL flake. Maybe there's a more general issue with rollbacks on MySQL 🙁

PR #37886, opened 2024-01-18
[MLv2] Migrate `isEditable` method to MLv2

nemanjaglumac:
This PR migrates `isEditable` and `isQueryEditable` methods to MLv2.
It completely removes them from the `Question`, `StructuredQuery` and `NativeQuery` prototypes.

**Please note:**
A commit 6e260a9043afaff8da508d339b7e3717e45c901a was already reviewed separately in https://github.com/metabase/metabase/pull/38058

## Some of the lessons learned while working on this
- Any time we're referencing a question inside a selector, we have to assume it might be `undefined`
- The way to reflect the real-world "missing data permissions" scenario in FE unit tests is to make sure there's no `database` in the mocked `metadata` object

### DIFF
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -503,11 +503,6 @@ class Question {
     return (db && db.auto_run_queries) || false;
   }
 
-  isQueryEditable(): boolean {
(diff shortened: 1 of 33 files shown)


kamilmielnik (approved):
:+1:

PR #37894, opened 2024-01-18
Render Auth Cards when env var is set

### LINKED ISSUE #33264: SAML auth page is weird when set by env vars
### Describe the bug

image

We can do a better job by keeping the SAML box and adding the env var text inside.

### To Reproduce

1. Set SAML using env vars
2. Go to `/admin/auth`

### Expected behavior

_No response_

### Logs

_No response_

```JSON
- master
```

### Severity

cosmetic

### Additional context

_No response_

npfitz:
Closes https://github.com/metabase/metabase/issues/33264

### Description
Small change to render auth cards in a read only fashion when their enables setting is set with an environment variable. This adds the optional forceRenderWidget prop to the setting definition, and makes changes to the `<AuthCard />` component to see if `setting.is_env_var` has been set to true, and if so provides `<AuthCardBody>` with a message / link to display, rather than a button link.

### How to verify

1. In your local environment, set `$MB_JWT_ENABLED` to true and start the backend
2. Navigate to Admin -> Settings -> Auth. You should see a card that informs you that JWT is set using the environment variable, and a link that will provide more information

### Demo
image

### Checklist

### DIFF
--- a/enterprise/frontend/src/metabase-enterprise/auth/index.js
+++ b/enterprise/frontend/src/metabase-enterprise/auth/index.js
@@ -38,6 +38,7 @@ PLUGIN_ADMIN_SETTINGS_UPDATES.push(sections =>
         noHeader: true,
         widget: SamlAuthCard,
         getHidden: () => !hasPremiumFeature("sso_saml"),
+        forceRenderWidget: true,
       },
       {
         key: "jwt-enabled",
--- a/enterprise/frontend/src/metabase-enterprise/auth/index.js
+++ b/enterprise/frontend/src/metabase-enterprise/auth/index.js
@@ -46,6 +46,7 @@ PLUGIN_ADMIN_SETTINGS_UPDATES.push(sections =>
         noHeader: true,
         widget: JwtAuthCard,
         getHidden: () => !hasPremiumFeature("sso_jwt"),
+        forceRenderWidget: true,
       },
       ...apiKeySettings,
       {
--- a/frontend/src/metabase/admin/settings/auth/components/AuthCard/AuthCard.unit.spec.tsx
+++ b/frontend/src/metabase/admin/settings/auth/components/AuthCard/AuthCard.unit.spec.tsx
@@ -1,5 +1,6 @@
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
+import { createMockSettingDefinition } from "metabase-types/api/mocks";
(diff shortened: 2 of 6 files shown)


frontend/src/metabase/admin/settings/auth/components/AuthCard/AuthCard.unit.spec.tsx
```diff
@@ -54,13 +55,31 @@ describe("AuthCard", () => {
...
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("card-badge")).not.toBeInTheDocument();
+  });
 });
 
+const getSetting = (opts?: Partial<AuthSetting>): AuthSetting =>
+  createMockSettingDefinition({
+    value: false,
+    ...opts,
+  }) as AuthSetting;
```
npfitz, line 81: I know this is a sin, but I couldn't figure out how to tell TS that this essentially resolves to the same thing without typecasting (AuthSetting is just a Setting Definition with a boolean value, and if one is being passed into getSetting it needs to be boolean anyways).

iethree: I think obscene type casting in tests is perfectly fine

enterprise/frontend/src/metabase-enterprise/auth/index.js
```diff
@@ -38,13 +38,15 @@ PLUGIN_ADMIN_SETTINGS_UPDATES.push(sections =>
         noHeader: true,
         widget: SamlAuthCard,
         getHidden: () => !hasPremiumFeature("sso_saml"),
+        forceRenderWidget: true,
```
iethree, line 41: I'm not following what `forceRenderWidget` is accomplishing?

npfitz: Originally, if `setting.is_env_setting` was set to true, we should not render a widget, only some text with a message. I made a change in `SettingsSetting.jsx` to check this new prop. If it's set to true, we'll render the widget defined in the setting definition regardless, and it'll be the widgets responsibility to handle the env var case

iethree: for the record, I hate this, but can't think of a better way

iethree (approved):
🎖️

PR #37900, opened 2024-01-18
Return “contains” and “does-not-contain” for comment/description

### LINKED ISSUE #36856: “contains” and “does-not-contain” returned for the wrong fields
### Describe the bug

image

### To Reproduce

filter by a boolean value. [slack thread](https://metaboat.slack.com/archives/C02H619CJ8K/p1702653982459689)

### Expected behavior

AlexP says: > `contains` and `does-not-contain` should be only returned for string columns that have either comment or description semantic type

### Logs

_No response_

```JSON
master
```

### Severity

P3

### Additional context

_No response_

metamben:
Fixes #36856.

`src/metabase/lib/drill_thru/quick_filter.cljc` contains the following in the namespace docstring:
>   - For string columns which have `type/Comment` or `type/Description` semantic type, allow `contains` and
    `does-not-contain` operators.

Because of this, I've removed `:=` and `:!=` from the list of available operators for comments and descriptions.

### DIFF
--- a/src/metabase/lib/drill_thru/quick_filter.cljc
+++ b/src/metabase/lib/drill_thru/quick_filter.cljc
@@ -76,10 +76,10 @@
         {:name   label
          :filter (operator op field-ref value)})
 
-      (lib.types.isa/string? column)
-      (for [[op label] [[:=  "="]
-                        [:!= "≠"]
-                        [:contains "contains"]
+      (and (lib.types.isa/string? column)
+           (or (lib.types.isa/comment? column)
+               (lib.types.isa/description? column)))
+      (for [[op label] [[:contains "contains"]
                         [:does-not-contain "does-not-contain"]]]
         {:name   label
          :filter (operator op field-ref value)})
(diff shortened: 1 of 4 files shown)


src/metabase/lib/drill_thru/quick_filter.cljc
```diff
@@ -76,10 +76,10 @@
...
 
-      (lib.types.isa/string? column)
-      (for [[op label] [[:=  "="]
-                        [:!= "≠"]
-                        [:contains "contains"]
+      (and (lib.types.isa/string? column)
+           (or (lib.types.isa/comment? column)
+               (lib.types.isa/description? column)))
+      (for [[op label] [[:contains "contains"]
                         [:does-not-contain "does-not-contain"]]]
```
bshepherdson, line 83: I'm not certain this is correct.

I read the issue as saying that contains/does-not-contain are appearing where they should not. Those are reserved for comment/description fields. But this also removes `:=` and `:!=`  exact matches for such fields, which seems strange to me.

bshepherdson (approved):

src/metabase/lib/drill_thru/quick_filter.cljc
metamben: The "specification" is not super clear. Having read the namespace's docstring, I decided this was the intended behavior, because the `:=` and `:!=` seem to be mentioned explicitly for other data types. I'm assuming we expect comments and descriptions to be too long to test with `:=` and `:!=` equal.
@bshepherdson, have you checked the namespace doc? How do you interpret it and why?

PR #37913, opened 2024-01-19
Fix flaking bigquery dataset-filtering-test

### LINKED ISSUE #37557: BE flake: metabase.driver.bigquery-cloud-sdk-test/dataset-filtering-test
Failed test: https://github.com/metabase/metabase/actions/runs/7485734698/job/20374725011?pr=37495#step:3:957

calherries:
Hopefully fixes https://github.com/metabase/metabase/issues/37557

I haven't been able to reproduce the flake, but based on [this failed run](https://github.com/metabase/metabase/actions/runs/7485734698/job/20374725011?pr=37495#step:3:957) I think it's caused by new datasets being created on the database by other test concurrent runs. I've reworked the test so that it checks the same properties hold as before but it can't flake from this cause anymore.


sansolovyov (approved):
Good catch!

PR #37915, opened 2024-01-19
add ENV to log snowplow events

npretto:
### Description

This both makes testing things locally easier and also makes us more aware of the testing we do (and not do).
This would also help help PMs and PDs make sure we're tracking things as they expect.

Ideally we could do something similar on the BE

### Demo

image

### DIFF
--- a/frontend/src/metabase/lib/analytics.js
+++ b/frontend/src/metabase/lib/analytics.js
@@ -41,6 +41,22 @@ export const trackStructEvent = (category, action, label, value) => {
 };
 
 export const trackSchemaEvent = (schema, version, data) => {
+  // eslint-disable-next-line no-undef
+  if (process?.env?.NODE_ENV === "development") {
+    try {
+      const { event, ...other } = data;
+      // eslint-disable-next-line no-console
+      console.log(
+        `%c[SNOWPLOW EVENT]%c, ${event}`,
+        "background: #222; color: #bada55",
+        "color: ",
+        other,
+      );
+    } catch (e) {
+      console.error("error while printing snowplow event");
+    }
+  }
+
   if (!schema || !Settings.trackingEnabled()) {
     return;
   }
--- a/frontend/src/metabase/lib/analytics.js
+++ b/frontend/src/metabase/lib/analytics.js
@@ -1,4 +1,5 @@
 import * as Snowplow from "@snowplow/browser-tracker";
+import { shouldLogAnalytics } from "metabase/env";
 import Settings from "metabase/lib/settings";
 import { getUserId } from "metabase/selectors/user";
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/lib/analytics.js
```diff
@@ -41,6 +41,22 @@ export const trackStructEvent = (category, action, label, value) => {
 };
 
 export const trackSchemaEvent = (schema, version, data) => {
+  // eslint-disable-next-line no-undef
+  if (process?.env?.NODE_ENV === "development") {
```
kamilmielnik, line 45: Can we not do this by default, but require e.g. an env variable to be set up for it?
The console is already a pain to work with due to React prop-types errors.

npretto: Good point, wrapped it in a env var

frontend/src/metabase/lib/analytics.js
```diff
@@ -41,6 +41,22 @@ export const trackStructEvent = (category, action, label, value) => {
 };
 
 export const trackSchemaEvent = (schema, version, data) => {
+  // eslint-disable-next-line no-undef
+  if (process.env.MB_LOG_ANALYTICS === "true") {
+    try {
```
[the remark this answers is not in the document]
npretto: Good catch, I added it when it wasn't under a env flag and I wasn't sure what we passed as data -> if my destructuring was safe

webpack.config.js
```diff
@@ -238,6 +238,7 @@ const config = (module.exports = {
     new NodePolyfillPlugin(), // for crypto, among others
     new webpack.EnvironmentPlugin({
       WEBPACK_BUNDLE: "development",
+      MB_LOG_ANALYTICS: "false",
```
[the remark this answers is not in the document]
npretto: I don't think that's very scalable, we can't have a command for any combination of envs.
For example almost everyday I use both `yarn dev` and `yarn build:hot`, others may use `yarn dev-ee` (which pretty much just sets MB_VERSION etc

frontend/src/metabase/lib/analytics.js
```diff
@@ -41,6 +41,18 @@ export const trackStructEvent = (category, action, label, value) => {
 };
 
 export const trackSchemaEvent = (schema, version, data) => {
+  // eslint-disable-next-line no-undef
+  if (process.env.MB_LOG_ANALYTICS === "true") {
```
kamilmielnik, line 45: We should probably add this variable to `frontend/src/metabase/env.js`.

kamilmielnik (approved):

deniskaber (approved):

PR #37942, opened 2024-01-19
Formatting for time types

### LINKED ISSUE #11398: Time column ignore formatting and always shows 12h-style AM/PM
**Describe the bug**
No matter [what formatting level](https://www.metabase.com/docs/latest/administration-guide/19-formatting-settings.html) is set, `Time` columns always shows 12 hour style.
Also, it shows Date-style options, which it shouldn't, since they have no effect.

**To Reproduce**
1. Have a database with a column type `Time`
2. Try settings formatting at any level (Global, Data Model, Question), it always shows 12h-style

**Screenshots**
image

Metabase 0.33.6 and master on various setups and databases

**Additional context**
https://discourse.metabase.com/t/time-format-does-not-effect/8140
Related to #11394

markbastian:
Currently, the frontend does not do any special formatting for time of day types. These are always formatted as 12 hour AM/PM times. This PR adds a new function to `frontend/src/metabase/lib/formatting/time.ts`, `formatTimeWithOptions`, that applies user-defined formatting to the time. Previously, `formatTime` was called, which takes no formatting options, so none can be applied.

Fixes #11398

### DIFF
--- a/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
@@ -373,6 +373,57 @@ describe("scenarios > visualizations > table > conditional formatting", () => {
   });
 });
 
+describe("scenarios > visualizations > table > time formatting (#11398)", () => {
+  const singleTimeQuery = `
+      WITH t1 AS (SELECT TIMESTAMP '2023-01-01 18:34:00' AS time_value),
+           t2 AS (SELECT CAST(time_value AS TIME) AS creation_time
+                  FROM t1)
+      SELECT *
+      FROM t2;
+  `;
+
+  beforeEach(() => {
+    restore();
+    cy.signInAsNormalUser();
+  });
+
+  it("should work with boolean columns", { tags: ["@external"] }, () => {
+    cy.createNativeQuestion(
+      {
+        name: "11398",
+        native: {
+          query: singleTimeQuery,
+        },
+      },
+      { visitQuestion: true },
+    );
+
+    // Open the formatting menu
+    cy.icon("chevrondown").click();
+    cy.findByTestId("drill-through-section").within(() => {
+      cy.icon("gear").click();
+    });
+
+    cy.findByTestId("column-formatting-settings").within(() => {
+      // Set to hours, minutes, seconds, 24-hour clock
+      cy.findByText("HH:MM:SS").click();
+      cy.findByText("17:24 (24-hour clock)").click();
+    });
+
+    // And you should find the result
+    cy.findByRole("gridcell").findByText("18:34:00");
+
+    cy.findByTestId("column-formatting-settings").within(() => {
+      // Add millisecond display and change back to 12 hours
+      cy.findByText("HH:MM:SS.MS").click();
+      cy.findByText("5:24 PM (12-hour clock)").click();
+    });
+
+    // And you should find the result
+    cy.findByRole("gridcell").findByText("6:34:00.000 PM");
+  });
+});
+
 function headerCells() {
   return cy.findAllByTestId("header-cell");
 }
--- a/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
@@ -387,7 +387,7 @@ describe("scenarios > visualizations > table > time formatting (#11398)", () => 
     cy.signInAsNormalUser();
   });
 
-  it("should work with boolean columns", { tags: ["@external"] }, () => {
+  it("should work with time columns", { tags: ["@external"] }, () => {
     cy.createNativeQuestion(
       {
         name: "11398",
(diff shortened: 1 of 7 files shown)


frontend/src/metabase/lib/formatting/time.ts
```diff
@@ -67,3 +67,34 @@ export function formatTimeWithUnit(
...
+}
+
+export function formatTimeWithOptions(
+  time: Moment,
+  unit: DatetimeUnit,
+  options: TimeOnlyOptions = {},
+) {
+  const parsedTime = parseTime(time);
+
+  const timeStyle = options.time_style
```
EmmadUsmani, line 85: Nit: `options.time_style ?? DEFAULT_TIME_STYLE` is a bit more elegant

frontend/src/metabase/lib/formatting/time.ts
```diff
@@ -67,3 +67,34 @@ export function formatTimeWithUnit(
...
+    ? options.time_style
+    : DEFAULT_TIME_STYLE;
+
+  const timeEnabled = options.time_enabled
```
EmmadUsmani, line 89: Nit: I thought we had a lint rule against nested ternaries since they are hard to read but I guess not. An if-else would be easier to read.

```typescript
  let timeEnabled;
  if (options.time_enabled) {
    timeEnabled = options.time_enabled;
  } else if (hasHour(unit)) {
    timeEnabled = "minute";
  } else {
    timeEnabled = null;
  }
  ```

File: frontend/src/metabase/lib/formatting/time.ts [hunk 3]
```diff
@@ -67,3 +67,34 @@ export function formatTimeWithUnit(
...
+    ? options.time_enabled
+    : hasHour(unit)
+    ? "minutes"
+    : null;
+
+  const timeFormat = options.time_format
```
EmmadUsmani, line 95: Nit: same as before `??` works well here

frontend/src/metabase/lib/formatting/value.tsx
```diff
@@ -177,7 +177,7 @@ export function formatValueRaw(
   } else if (isEmail(column)) {
     return formatEmail(value as string, options);
   } else if (isTime(column)) {
-    return formatTime(value as Moment);
```
EmmadUsmani, line 180: We should delete the old `formatTime` entirely I don't think it's used anywhere else, except the static viz `format.ts`, which we probably want to replace with `formatTimeWithOptions` as well

Screenshot 2024-01-19 at 4 11 39 PM

frontend/src/metabase/lib/formatting/time.ts
```diff
@@ -67,3 +67,34 @@ export function formatTimeWithUnit(
...
+    ? options.time_format
+    : getTimeFormatFromStyle(timeStyle, unit, timeEnabled as any);
+
+  return parsedTime.format(timeFormat);
```
EmmadUsmani, line 99:
```suggestion
return parsedTime.isValid() ? parsedTime.format("LT") : String(time);
```

I think we should copy the old return logic from `formatTime`

EmmadUsmani (commented):
Overall looks good, but I think we can also use the new function in `frontend/src/metabase/static-viz/lib/format.ts` and delete the old `formatTime` function.

frontend/src/metabase/lib/formatting/time.ts
```diff
@@ -67,3 +61,35 @@ export function formatTimeWithUnit(
 
   return m.format(timeFormat);
 }
+
+interface TimeOnlyOptions {
```
alxnddr, line 65: Wdyt about moving this type to `frontend/src/metabase/lib/formatting/types.ts`?
I think it would be reasonable to call it `TimeOptions` and extend `OptionsType` with it to avoid duplication

frontend/src/metabase/lib/formatting/time.ts
```diff
@@ -67,3 +61,35 @@ export function formatTimeWithUnit(
...
+  local?: boolean;
+  time_enabled?: "minutes" | "milliseconds" | "seconds" | null;
+  time_format?: string;
+  time_style?: string;
+}
+
+export function formatTimeWithOptions(
```
alxnddr, line 72: nit: `WithOptions` suffix may be omited since we won't have a separate `formatTime` or `formatTimeWithoutOptions`. Other formatters like `formatUrl` or `formatEmail` accept options and do not have the suffix

e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
```diff
@@ -373,6 +373,58 @@ describe("scenarios > visualizations > table > conditional formatting", () => {
...
+      {
+        name: "11398",
+        native: {
+          query: singleTimeQuery,
+        },
+      },
+      { visitQuestion: true },
+    );
+
+    // Open the formatting menu
```
alxnddr, line 401: 👍 for clarifying comments

alxnddr (approved):
Looks great! I left a couple of suggestions, please have a look before merging

frontend/src/metabase/lib/formatting/time.ts
markbastian: Good call. Especially since this effectively replaced the old one.

frontend/src/metabase/lib/formatting/time.ts
markbastian: So, OptionsType extends from TimeOptions? It seems like other small interfaces have more "locality of behavior" (they live near their functions if only used once), but I don't have strong opinion on this.

markbastian: Done.

Issue #36370, opened 2023-12-04
[Epic] Make the nav sidebar more predictable and discoverable


luizarakaki:
**Links**
- product doc: [link](https://www.notion.so/metabase/Make-the-nav-sidebar-more-predictable-and-discoverable-b205ed974d314aa5a8f21172911a3324)
- eng doc: _link to technical design doc, if any_
- feature branch: `branch-name` _this should be the feature branch where this work will be done in. PRs will be delivered against this branch_
- issue links: _related issues if any_

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/issues/36963
- [x] Make navbar UI match the new designs (https://github.com/metabase/metabase/pull/37821)
- [x] Handle embedding logic
- [x] Implement new heuristics for navbar
- [x] Handle case where user has a custom home page / landing page configured
- [x] Add test cases
- [x] Handle mobile hover states
- [x] Update sidebar UI
```

sloansparger:
@luizarakaki I've just merged the last change if you want to check it out!

PR #37944, opened 2024-01-19
Navbar Improvements: Updated Heuristics

sloansparger:
### Description

Implements new logic for the navbar being open/closed for a page and auto-closing in certain conditions.

This is the core change to complete https://github.com/metabase/metabase/issues/36370

### Technical description

The main part of this change is having the navbar state subscribe to react-router-redux `@@router/LOCATION_CHANGE`  event to decide if we should keep the navbar open/closed on route changes. A special case is accommodated for home page redirects where we want to treat a url we would typically auto-close for as if it were the normal home page. This is accomplished by adding a flag to the location state on the replace event which is used to return the prev state / ignore the event.

### How to verify

- Log into metabase, should see navbar open
- Visit a dashboard, should see navbar auto-close
- Click the MB logo, should stay closed
- Go back a page
- Manually open the navbar
- Click the MB logo, should stay open
- Go to a collection, should stay open
- Go to a dashboard within a collection, should auto-close
- Using the breadcrumbs at the top of the dashboard, go back to the collection, should stay closed
- Manually open the navbar
- Create a new question or sql query, should auto close

### Demo

https://github.com/metabase/metabase/assets/7104357/fc7c30c2-abe5-4504-b150-1ba083e23d71
(demo hangs at the end due to a perf issue unrelated to this change)

### Checklist


e2e/test/scenarios/navbar.cy.spec.js
```diff
@@ -0,0 +1,112 @@
...
+      cy.visit("/");
+      cy.findByTestId("main-navbar-root").should("be.visible");
+    });
+
+    it("state should preserve when clicking the mb logo", () => {
+      cy.visit("/collection/root");
+      cy.findByTestId("main-navbar-root").should("be.visible");
+      cy.findByTestId("main-logo-link").click();
+      cy.findByTestId("main-navbar-root").should("be.visible");
+      visitDashboard(8);
```
sloansparger, line 26: anyone know where I can import constants for things like this. i make a few magical references in this file that i'd like to fix...

frontend/src/metabase/nav/containers/MainNavbar/MainNavbar.styled.tsx
```diff
@@ -140,7 +140,7 @@ export const LoadingTitle = styled.h2`
 `;
 
 export const PaddedSidebarLink = styled(SidebarLink)`
-  padding-left: ${space(2)};
+  padding-left: 12px;
```
iethree, line 143: nit: I think we prefer 0.75rem

e2e/support/helpers/e2e-ui-elements-helpers.js
```diff
@@ -34,7 +34,7 @@ export function leftSidebar() {
 }
 
 export function navigationSidebar() {
-  return cy.get("#root aside").first();
+  return cy.findByTestId("main-navbar-root");
```
iethree, line 37: ❤️

iethree (approved):
I love everything about this 🎉 


200w-1

(but I still found a few little nits that you could address)

iethree:
But really, I just love how you consolidated all the random bits of sidebar logic so cleanly. 💖

e2e/test/scenarios/navbar.cy.spec.js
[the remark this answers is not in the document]
sloansparger: I think I don't actually need this. Good callout.

frontend/src/metabase/redux/app.ts
```diff
@@ -12,6 +12,19 @@ import {
 
 import type { Dispatch } from "metabase-types/store";
 
+interface LocationChangeAction {
+  type: string; // "@@router/LOCATION_CHANGE"
```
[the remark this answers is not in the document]
sloansparger: IT YELLS AT ME 😢 that's what I wanted to do, but I think there's some prerequisite typing we'd need to do elsewhere in our redux setup to get it happy again.

luizarakaki:
Nice work! Everything working as specced

PR #37954, opened 2024-01-22
Fix race-condition that could revert (un)subscriptions

crisptrutski:
### Description

I spotted a potential race condition with this API, which can accidentally undo concurrent (un)subscriptions.

It's a simple fix.

### DIFF
--- a/src/metabase/api/session.clj
+++ b/src/metabase/api/session.clj
@@ -357,15 +357,16 @@
    email    :string
    hash     :string}
   (check-hash pulse-id email hash (request.u/ip-address request))
-  (api/let-404 [pulse-channel (t2/select-one PulseChannel :pulse_id pulse-id :channel_type "email")]
-    (let [emails (get-in pulse-channel [:details :emails])]
-      (if (some #{email} emails)
-        (t2/update! PulseChannel (:id pulse-channel) (update-in pulse-channel [:details :emails] #(remove #{email} %)))
-        (throw (ex-info (tru "Email for pulse-id doesn't exist.")
-                        {:type        type
-                         :status-code 400}))))
-    (events/publish-event! :event/subscription-unsubscribe {:object {:email email}})
-    {:status :success :title (:name (pulse/retrieve-notification pulse-id :archived false))}))
+  (t2/with-transaction [_conn]
+    (api/let-404 [pulse-channel (t2/select-one PulseChannel :pulse_id pulse-id :channel_type "email")]
+      (let [emails (get-in pulse-channel [:details :emails])]
+        (if (some #{email} emails)
+          (t2/update! PulseChannel (:id pulse-channel) (update-in pulse-channel [:details :emails] #(remove #{email} %)))
+          (throw (ex-info (tru "Email for pulse-id doesn't exist.")
+                          {:type        type
+                           :status-code 400}))))
+      (events/publish-event! :event/subscription-unsubscribe {:object {:email email}})
+      {:status :success :title (:name (pulse/retrieve-notification pulse-id :archived false))})))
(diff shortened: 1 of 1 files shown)


src/metabase/api/session.clj
```diff
@@ -375,9 +375,8 @@
    hash     :string}
   (check-hash pulse-id email hash (request.u/ip-address request))
   (api/let-404 [pulse-channel (t2/select-one PulseChannel :pulse_id pulse-id :channel_type "email")]
-    (let [emails       (get-in pulse-channel [:details :emails])
-          given-email? #(= % email)]
-      (if (some given-email? emails)
+    (let [emails       (get-in pulse-channel [:details :emails])]
```
calherries, line 378:
```suggestion
    (let [emails (get-in pulse-channel [:details :emails])]
```

src/metabase/api/session.clj
```diff
@@ -360,7 +360,7 @@
   (api/let-404 [pulse-channel (t2/select-one PulseChannel :pulse_id pulse-id :channel_type "email")]
     (let [emails (get-in pulse-channel [:details :emails])]
       (if (some #{email} emails)
-        (t2/update! PulseChannel (:id pulse-channel) (assoc-in pulse-channel [:details :emails] (remove #{email} emails)))
+        (t2/update! PulseChannel (:id pulse-channel) (update-in pulse-channel [:details :emails] #(remove #{email} %)))
```
calherries, line 363: I don't understand how this solves the race-condition. `pulse-channel` is immutable and so is `emails`, so `emails` before the cange is the same as the argument (`%` ) to the anon function after the change.

calherries (changes requested):

src/metabase/api/session.clj
piranha, line 363: I don't think this will change anything, since `emails` are the thing you got from `pulse-channel` just before this line of code and there is no updates going on. So the logic stays the same...

crisptrutski: The issue is that query value can be stale - the database could have been changed by a concurrent API call.

Have I missed something around transaction isolation - do API calls implicitly run inside a serializable transaction? I assume not for performance reasons.

crisptrutski: Ah nevermind, I see what you mean - I though that `pulse-channel` was a reference to the entity returned within the update transaction. I'll figure out how to do this properly with toucan.

crisptrutski: Ah yes I see I misread how the `update!` is being used, I thought it was acting like a `swap!`. Can you point me at the right way to make this atomic with toucan2? I'll poke around, still new to the library.

piranha: Something like `(t2/update! PulseChannel id :details ...some way to pass sql to the db...)`, I think. But there is a problem - `details` is a json field (well, it's `text` in my postgres), so the way to remove an email from json could be different between pgsql and mysql?..

crisptrutski: I was thinking that toucan would have something like `(t2/update! ... swap-fn)` that would just run a select and update within transaction. 

I think just manually wrapping both of these functions in `(t2-with-transaction ...)` is the idiomatic way to fix this?

piranha: Well, translating Clojure code into an action inside a database sounds like an ambitious goal, so I doubt toucan will ever go there. :)

Yeah, I guess transaction should fix the problem.

crisptrutski: Looking around at some other APIs it looks like this is indeed the idiom.

calherries (approved):

PR #37966, opened 2024-01-22
Fix wrong `target` prop type on `DashCardParameterMapper`

### LINKED ISSUE #37964: `DashCardCardParameterMapper` wrong prop type for `target`
`DashCardCardParameterMapper` has a wrong prop type for `target`.
According to the current code, the type expected should be `object`, but according to the reality, it's an `array`. There's also a huge warning in the console about this:
image

### To reproduce:
1. Either create a dashboard with a question connected to it or create a new one
2. Open such dashboard
3. Edit dashboard
4. Click on the "cog" icon in the dashboard filter
image
5. See the error in the console

nemanjaglumac:
This PR fixes wrong `target` prop type on `DashCardParameterMapper`.

To confirm, follow the repro steps from #37964 and make sure there is no warning in console any more.
You can also open the React dev tools to confirm that the `target` accepts an array, indeed.

When a field is connected
image

Where a filter isn't connected, or when there's no field to connect it to:
image

Fixes #37964

### DIFF
--- a/frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
+++ b/frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
@@ -37,7 +37,7 @@ const setup = options => {
       card={createMockCard()}
       dashcard={createMockDashboardCard()}
       editingParameter={{}}
-      target={null}
+      target={undefined}
       mappingOptions={[]}
       metadata={metadata}
       setParameterMapping={jest.fn()}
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
```diff
@@ -37,7 +37,7 @@ const setup = options => {
       card={createMockCard()}
       dashcard={createMockDashboardCard()}
       editingParameter={{}}
-      target={null}
+      target={undefined}
```
uladzimirdev, line 40: if you pass nothing, it's going to be undefined

uladzimirdev (approved):

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
nemanjaglumac: I sad this pattern in a few other places in our codebase so I thought maybe we had some internal convention to explicitly mark some props as undefined in tests?

For example
https://github.com/metabase/metabase/blob/master/frontend/src/metabase/query_builder/components/expressions/ExpressionWidget.unit.spec.tsx#L196

kamilmielnik: > @kamilmielnik do we have a pattern?

I don't think so.

PR #37967, opened 2024-01-22
Migration to add new report_card.type

### LINKED ISSUE #37367: BE - Migrate `dataset: true` to enum value
See https://github.com/metabase/metabase/issues/29752#issuecomment-1501976631

This is just about turning the boolean into an enum.
Adding a new "metric" value to that enum should be done in #37348.

Suggested attribute name: `type`
Suggested enum values: `"question"` & `"model"` (in [the future also `"metric"`](https://github.com/metabase/metabase/issues/37348))

### Notes (Added by Cam 2/13/24)

This was partially implemented in #38074. #38074 adds the new `type` column, but does not remove the `dataset` column or migrate things to use it in the application database. Let's finish that work and remove `dataset` entirely. We also need to update MLv2 code and other things that #38074 missed to look at `type` rather than `dataset`.

This may need to be done in sync with https://github.com/metabase/metabase/issues/38706

qnkhuat:
WIP
- [ ] :get card
- [ ] :get card/:id/related
- [ ] :get card/:id
- [ ] :get /card/:id/series
- [ ] :post card
- [ ] :put card/:id
- [ ] :post card/:id/copy
- [ ] :post card/:id/query
- [ ] :post card/pivot/:id/query
- [ ] :get activity/popular_items
- [ ] :get activity/recent_views
- [ ] :get bookmark
- [ ] :get dashboard/:id
- [ ] :put dashboard/:id/cards
- [ ] :get database/:id/card_autocomplete_suggestions
- [ ] :get database/:id/usage_info
- [ ] :get revision
- [ ] :get transform/:id/:shcema/:transform-name

Resolves #37367

### DIFF
--- a/resources/migrations/001_update_migrations.yaml
+++ b/resources/migrations/001_update_migrations.yaml
@@ -5228,7 +5228,7 @@ databaseChangeLog:
             sql: >-
               UPDATE report_card
               SET type = 'model'
-              WHERE datset is true
+              WHERE dataset is true
       rollback:
         - sql:
             sql: >-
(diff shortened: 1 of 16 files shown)


qnkhuat:
Close in favor of https://github.com/metabase/metabase/pull/38074

PR #37972, opened 2024-01-22
Fix dashboard parameter values in embed preview not working

### LINKED ISSUE #37914: Text filters shown as input box in embed instead as dropdown
### Describe the bug

Decided to open this issue after spending some time reproducing #31347 and noticing the behavior has changed a lot.
Essentially, you expect filters to look the same in the embed, in the embed preview and in the dashboard inside of Metabase.

Findings:
* filter set as dropdown in dashboard, no default value for filter in dashboard, editable in embed -> filter is rendered as input box in embed 
* filter set as dropdown in dashboard, default value set for filter in dashboard, locked in embed -> filter is rendered as input box in embed
* filter set as droddown in dashboard, default value set for filter in dashboard, editable in embed -> filter is rendered as dropdown in embed as it should, but it's rendered differently in the preview

The preview features and linking of the filters seems to work properly, unlike reported in #31347.

### To Reproduce

1. Go to New -> Question -> Products -> Save
2. Create New Dashboard -> Add 3 Filters and Add the above question then link them to column Category
image
3. Link the first filter to the second one -> Save
image
4. Confirm that the link is working by selecting Doohickey for Second Filter and Gadget,Gizmo for Third filter ... Notice that on metabase works as expected and First filter only shows Doohickey as an option
image
5. Go to Sharing -> Embed your application -> Make the filters editable -> Publish
6. The filters are rendered in the embed and preview as input boxes instead of as dropdown
image

7. Edit the dashboard, set the default value for text 1 as Doohickey -> Save

8. Go to Sharing -> Embed your application -> Preview

9. In the preview Text filter is still rendered as input box but text 1 is rendered as multi-select, in the embed it is a proper dropdown

Text:
image

Text 1 in preview inside of UI:
image

Text 1 opening iframe from preview URL  in incognito mode:
image

10. Copy the iframe URL from the preview using inspect and open it in incognito mode. You will see that text and text1 are now rendered as dropdown and also that the linking of the filters is working as expected. Text2 is still rendered as input box.

11. Set text1 to locked, set a preview value for it as "Doohickey", publish, copy the iframe URL from the preview using inspect, open it in incognito mode. The results are properly filtered (Doohickey only), but text filter shows again as input box even though there is a default value
  

### Expected behavior

The filters should look the same across dashboard in MB, embed preview and real life embed. The combination of setting a default value in the dashboard and setting the filter as editable triggers this consistent behavior but it should not be needed

### Logs

_No response_

```JSON
Metabase v1.48.1
```

### Severity

Might be blocking in certain use cases

### Additional context

_No response_

WiNloSt:
Closes https://github.com/metabase/metabase/issues/37914

### Description

When we preview embed, we have some conditional logics that make sure the preview works e.g. we want to always allow previewing the dashboard even without publishing it first.

This has a consequence on the dashboard values that makes it calls an unavailable API. It calls the dashboard value with UUID instead of the number ID.

### How to verify

Follow the reproduce steps from #37914, and make sure the dashboard preview and embedded dashboard parameters behaviors are the same.

### Demo

### Checklist


e2e/test/scenarios/embedding/reproductions/37914-embed-preview-parameter-values-not-working.cy.spec.js
```diff
@@ -0,0 +1,173 @@
...
+import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
+
+const { PRODUCTS, PRODUCTS_ID } = SAMPLE_DATABASE;
+
+const questionDetails = {
+  name: "Products",
+  query: { "source-table": PRODUCTS_ID },
+};
+
+const filter3 = {
```
WiNloSt, line 18: I declare `filter3` first because I want to use `filter2`'s id in `filter`, and I don't want to hardcode that, so I need to declare `filter2` before `filter1`, and I figure I should put `filter3` first for consistency.

frontend/src/metabase/dashboard/actions/data-fetching.js
```diff
@@ -276,7 +277,7 @@ export const fetchDashboard = createAsyncThunk(
       return {
         entities,
         dashboard: result,
-        dashboardId: dashId,
+        dashboardId: result.id,
```
WiNloSt, line 280: I've taken a look at all the dashboard types above.
- `embed`: Embedded dashboard
- `transient`: x-ray dashboard
- `inline`: I don't think this is used anymore. Asked [a question in Slack.](https://metaboat.slack.com/archives/C505ZNNH4/p1706106083493939)
- and a normal dashboard.

All have `result.id === dashId` except the embed dashboard since I change that to `IS_EMBED_PREVIEW ? result.id : dashId` in this PR. I don't want to repeat the same condition again, so I figure using `result.id` would make more sense.

frontend/src/metabase/dashboard/actions/data-fetching.js
```diff
@@ -196,7 +197,7 @@ export const fetchDashboard = createAsyncThunk(
         );
         result = {
           ...result,
-          id: dashId,
+          id: IS_EMBED_PREVIEW ? result.id : dashId,
```
WiNloSt, line 200: This is essentially the fix. Previously `dashId` in this case (`dashboardType === "embed"`) would be a UUID, and that makes calling the parameter values API with the dashboard as a UUID

But that means we'd be calling `/api/dashboard/:uuid` which is wrong, we're supposed to call `/api/dashboard/:dashboard-id` instead. The thing is that in embed preview we don't call `/api/embed/*` endpoints because it's possible for the resource to not have been embedded yet, calling `/api/embed/*` in that case will not return anything.

You can see that we have a logic that changes the dashboard and question parameter values endpoints to have `/api/embed` prefix for embedding, but we don't do that in embed preview
https://github.com/metabase/metabase/blob/e2f07d813d0c60ad7131d6c3abb9e4e838757a76/frontend/src/metabase/services.js#L531-L540

frontend/src/metabase/public/components/EmbedModal/EmbedModalContent/EmbedModalContent.tsx
```diff
@@ -118,14 +118,33 @@ export const EmbedModalContent = (
...
-  };
+  }, [embeddingParams, parameterValues, resourceParameters]);
 
-  const previewParametersBySlug = getPreviewParamsBySlug();
   const lockedParameters = getLockedPreviewParameters(
     resourceParameters,
     embeddingParams,
   );
 
+  const iframeUrl = useMemo(() => {
```
WiNloSt, line 128: This fixes the problem where the iframe is rendered twice.

https://www.loom.com/share/c082515a29724bcdb7d8a67535f21a1c

deniskaber (approved):

PR #18827, opened 2021-11-03
Refactor question ViewHeader component

kulyk:
This PR is the first step for updating the question page for datasets (feature epic: #18659). ViewHeader is displayed for any question (saved vs ad-hoc, native vs GUI, etc.), so it has lots of possible states. This PR adds unit tests for the component and refactors it to simplify subsequent changes.

* turned into a functional component
* some visual components are extracted as styled-components
* the component itself is broken down into left side (are different for saved vs ad-hoc question) and right side components
* reordered imports

### To Verify

1. For saved GUI, ad-hoc GUI, saved native and ad-hoc native questions
2. Open the question page
3. Make sure there are no changes in the header's look and behavior

We have some Percy coverage for questions already. I've also submitted a PR with table visual tests that also cover ad-hoc header view #18826

PR #37977, opened 2024-01-22
Migrate `isStructured` method references to MLv2

### LINKED ISSUE #37389: [MLv2] [FE] Use MLv2 `displayInfo` in `isStructured`, `isNative`, `isQueryEditable`

nemanjaglumac:
This PR migrates `isStructured()` references to MLv2.
It also removes the method altogether from the `Question` prototype.

Resolves  #37389

### DIFF
--- a/frontend/src/metabase-lib/Question.ts
+++ b/frontend/src/metabase-lib/Question.ts
@@ -204,14 +204,6 @@ class Question {
   }
 
   /**
-   * @deprecated use MLv2
-   */
-  isStructured(): boolean {
-    const { isNative } = Lib.queryDisplayInfo(this.query());
-    return !isNative;
-  }
-
-  /**
    * Returns a new Question object with an updated query.
    * The query is saved to the `dataset_query` field of the Card object.
    */
@@ -494,8 +486,9 @@ class Question {
 
     const hasSinglePk =
       table?.fields?.filter(field => field.isPK())?.length === 1;
+    const isStructured = !Lib.queryDisplayInfo(this.query()).isNative;
 
-    return this.isStructured() && !Lib.hasClauses(query, -1) && hasSinglePk;
+    return isStructured && !Lib.hasClauses(query, -1) && hasSinglePk;
   }
 
   canAutoRun(): boolean {
@@ -549,8 +542,9 @@ class Question {
    * of Question interface instead of Query interface makes it more convenient to also change the current visualization
    */
   usesMetric(metricId): boolean {
+    const isStructured = !Lib.queryDisplayInfo(this.query()).isNative;
     return (
-      this.isStructured() &&
+      isStructured &&
       _.any(
         QUERY.getAggregations(
           this.legacyQuery({ useStructuredQuery: true }).legacyQuery({
@@ -563,8 +557,9 @@ class Question {
   }
 
   usesSegment(segmentId): boolean {
+    const isStructured = !Lib.queryDisplayInfo(this.query()).isNative;
     return (
-      this.isStructured() &&
+      isStructured &&
       QUERY.getFilters(
         this.legacyQuery({ useStructuredQuery: true }).legacyQuery({
           useStructuredQuery: true,
@@ -1021,11 +1016,13 @@ class Question {
   }
 
   _convertParametersToMbql(): Question {
-    if (!this.isStructured()) {
+    const query = this.query();
+    const { isNative } = Lib.queryDisplayInfo(query);
+
+    if (isNative) {
       return this;
     }
 
-    const query = this.query();
     const stageIndex = -1;
     const filters = this.parameters()
       .map(parameter =>
--- a/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
+++ b/frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
@@ -93,12 +93,17 @@ export function ViewTitleHeader(props) {
   const previousQuery = usePrevious(query);
 
   useEffect(() => {
-    if (!question.isStructured() || !previousQuestion?.isStructured()) {
+    const { isNative } = Lib.queryDisplayInfo(query);
+    const isPreviousQuestionNative =
+      previousQuery && Lib.queryDisplayInfo(previousQuery).isNative;
+
+    if (isNative || isPreviousQuestionNative) {
       return;
     }
 
     const filtersCount = Lib.filters(query, -1).length;
-    const previousFiltersCount = Lib.filters(previousQuery, -1).length;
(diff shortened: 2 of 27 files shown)


frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
```diff
@@ -93,12 +93,17 @@ export function ViewTitleHeader(props) {
...
+      previousQuery && Lib.queryDisplayInfo(previousQuery).isNative;
+
+    if (isNative || isPreviousQuestionNative) {
       return;
     }
 
     const filtersCount = Lib.filters(query, -1).length;
-    const previousFiltersCount = Lib.filters(previousQuery, -1).length;
+    const previousFiltersCount =
+      previousQuery && Lib.filters(previousQuery, -1).length;
```
kamilmielnik, line 106: It's good when "count" is always a number:

```suggestion
    const previousFiltersCount =
      previousQuery ? Lib.filters(previousQuery, -1).length : 0;
```

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
```diff
@@ -93,12 +93,17 @@ export function ViewTitleHeader(props) {
   const previousQuery = usePrevious(query);
 
   useEffect(() => {
-    if (!question.isStructured() || !previousQuestion?.isStructured()) {
+    const { isNative } = Lib.queryDisplayInfo(query);
+    const isPreviousQuestionNative =
+      previousQuery && Lib.queryDisplayInfo(previousQuery).isNative;
+
+    if (isNative || isPreviousQuestionNative) {
```
kamilmielnik, line 100: Previously, when `previousQuestion` was `undefined`/`null`, `!previousQuestion?.isStructured()` evaluated to `true`. Now it evaluates to a falsy value. Not sure if it makes things better or worse, but to make it work exactly like before, we should do this:

```suggestion
    const isPreviousQuestionStructured =
      previousQuery && !Lib.queryDisplayInfo(previousQuery).isNative;

    if (isNative || !isPreviousQuestionStructured) {
```

frontend/src/metabase/services.js
```diff
@@ -59,7 +60,7 @@ export function maybeUsePivotEndpoint(api, card, metadata) {
   }
   if (
     question.display() !== "pivot" ||
-    !question.isStructured() ||
+    !isStructured(card) ||
```
kamilmielnik, line 63: Why aren't we using `question.query()` here?

frontend/src/metabase-lib/Question.ts
```diff
@@ -494,8 +486,9 @@ class Question {
 
     const hasSinglePk =
       table?.fields?.filter(field => field.isPK())?.length === 1;
+    const isStructured = !Lib.queryDisplayInfo(this.query()).isNative;
```
kamilmielnik, line 489: In some files we're introducing `isStructured` and in some files we're just plainly using `!isNative`. I think we should go with 1 approach:
- we can get rid of `isStructured` and always use `!isNative`
- we can fully embrace `isStructured` by adding it to `Lib.queryDisplayInfo` result

WDYT?

----

_this is a non-blocking comment_

kamilmielnik (commented):
Looks good :+1: 

Left some comments and a question inline.

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
nemanjaglumac: I tried that.
Related E2E tests started exploding, and I could confirm in the UI that things don't work as intended.

At that point, I couldn't even dig into WHY. This particular solution does work.

frontend/src/metabase/services.js
nemanjaglumac: Because this endpoint is used in Audit internal dashboards, and it explodes when you try to construct MLv2 query with the `internal` query.

nemanjaglumac: I can add a comment, if you think it's necessary?

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
kamilmielnik: Thanks for checking!

> At that point, I couldn't even dig into WHY

Probably because `1 > undefined` is `false`.

frontend/src/metabase/services.js
kamilmielnik: I don't think it's necessary. Thanks!

frontend/src/metabase-lib/Question.ts
uladzimirdev: I'm for the first: `!isNative`

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
nemanjaglumac: I found the original code before @kulyk rewrote it.
https://github.com/metabase/metabase/pull/18827/files#diff-7eae47ba62c31da65c448230eacac88be86470a204d450011acb631d7bb74144L80-L83

So I think we can safely assume that the intention was to make sure the both questions are structured.

nemanjaglumac: > Previously, when previousQuestion was undefined/null, !previousQuestion?.isStructured() evaluated to true. 

@kamilmielnik I'm 99% sure the older implementation (rewrite) made this assumption by mistake. Do you mind if I leave the code as is?

frontend/src/metabase-lib/Question.ts
nemanjaglumac: https://github.com/metabase/metabase/pull/37977/commits/b9e8074472138a7821eafe77d7d75d948a580017

nemanjaglumac: Addressed in this commit
fdc93b1809862f4a4a97d508f9282e7870e6b78f

nemanjaglumac: After some weird CI failures that I couldn't reproduce locally, I moved this step to a separate PR.
https://github.com/metabase/metabase/pull/38199

uladzimirdev (approved):

Issue #14683, opened 2021-02-05
Proposal: consolidate different REST API error patterns or at least come up with consistent rules


camsaul:
We have a few different patterns for returning errors from our REST API endpoints. Here's the ones I can think of off the top of my head:

###### 4xx status code with JSON response body with `:errors`  map of `field -> error`

```clj
{:errors {:card_id "card_id should be an Integer"}}
```

This is used by `defendpoint` schema-based validation, for example, and lots of other general precondition checks. The idea here is that we can show the error next to the appropriate part of a form where you input things, e.g. an error for a `Database` `:name` can be shown next to the appropriate text field rather than at the bottom of the screen. Or, if there are errors for multiple fields, we can show each error next to the appropriate field (this is more in theory, than in practice -- usually we throw an Exception and return an error right away as soon as we see one). I think the main issue with this pattern is that it doesn't make sense for errors that really aren't specific to a certain Field.

###### 4xx status code with JSON response body with string error `:message` 

```clj
{:message "That is not ok"}
```

This is used in a few places and by the `metabase.server.middleware.exceptions` for `clojure.lang.ExceptionInfo` exceptions that don't include anything in their `ex-data` (except for `:status-code`).

e.g.

```clj
(throw (ex-info "Oops!" {:status-code 400}))
```

gets caught by the middleware and we return a response like

```clj
{:message "Oops!"}
```

This pattern isn't used quite as often as the other pattern but I think is useful for errors that really don't have a corresponding Field.

###### 4xx or 500 status code, plain text response

e.g.

```clj
;; 404
"Not found."
```

Used mostly for perms errors or 404s.

###### QP Errors

The way query processor endpoints return errors is crazy. They return 200 (actually 202 with a `:_status` of 200, because they're async -- that's another subject entirely) and then you have to look at `:status` to see if it's an error.

```clj
;; normal response -- 200
{:status :completed, :data {...}}

;; error response -- also 200
{:status :failed, :error "Oops!"}
```
I don't really like returning 200s on errors, but that's actually how it worked since before I joined the team and I didn't change it when I ported MB to Clojure. In hindsight, I think I should have changed it.

Additionally, I've been working on adding `:error-type` to QP exceptions with a error type hierarchy. This is currently only used internally (to determine whether the error means we shouldn't retry a request), but this is a work in progress.

### Can we consolidate stuff?

[...]

Issue #11175, opened 2019-10-18
Standardize and enhance our error API responses and frontend components


tlrobinson:
1. We've got a couple different ways of returning errors from API endpoints. We should try to standardize these.
2. We (@kdoh in particular) have talked about having a standard "progressive disclosure" error component that shows the main error message immediately and additional information behind a click. We should also have a standard way to pass the main error message and the additional info in the API response.
3. We should distinguish between errors where we have some idea of what happened vs. uncaught exceptions
  - For "known" errors where we have a "curated" error message, we'd show the user the error message immediately, and if we have additional info put that behind a click
 - For uncaught exceptions we should show a generic "Something happened" error message, and additional information (Java error message, maybe type + stacktrace?) behind the click
4. In production we might not even want to include the Java error type and stacktrace (at least for non-admins) as it can be a (likely minor) security risk. We could include a request identifier (@flyingmachine suggested this) an admin can look up in the logs to find more info like a stack trace.

The current state of things:

1. Most API errors are something like this:

```json
{
  "message": "error message",
  "type": "class java.lang.Exception",
  "stacktrace": […],
}
```

(from here I presume https://github.com/metabase/metabase/blob/master/src/metabase/middleware/exceptions.clj#L70-L72)

2. For `dataset` endpoints we have the following:

```json
{
  "error": "error message"
  "error_type": "error-name"
  "stacktrace": [...]
}
```

3. For things like 404s we just return a plain text string, which isn't ideal for endpoints that normally return JSON. Our frontend API client handles that but returns it as a string instead of a `{ "message": ... }` object, so we end up with code like `response.data.message || response.data` to handle both cases, which is gross.

4. Form validation errors (this doesn't need to change IMHO):

```json
{
  "errors": {
    "field name": "field error"
  }
}
```

Here are the things we could/should include (all are optional):

1. top-level error message (currently `message`, or `error` in dataset)
2. field-level error message (currently `errors`)
3. error type/identifier for custom error pages (currently `type` is actually Java exception class which is not what i'm talking about. `error_type` in dataset is a Metabase-specific identifier)
4. additional information for progressive disclosure widget
5. documentation page identifier (passed to `MetabaseSettings.docsUrl()` to get a link)
6. stacktrace (only in dev and/or only for admins?)
7. request identifier (to correlate errors with server logs)

There are a bunch of different schools of thought on how JSON API responses should be structured but I tend to agree with this: https://stackoverflow.com/a/18953828

[...]

PR #37995, opened 2024-01-22
Change CSV append schema mismatch exception format

calherries:


src/metabase/api/table.clj
```diff
@@ -516,10 +516,14 @@
...
     (catch Throwable e
-      {:status (or (-> e ex-data :status-code)
-                   500)
-       :body   {:message (or (ex-message e)
-                             (tru "There was an error uploading the file"))}})
+      (if (::upload/schema-mismatch? (ex-data e))
+        {:status (-> e ex-data :status-code)
+         :body   (merge {:message (ex-message e)}
+                        (select-keys (ex-data e) [:extra-columns :missing-columns]))}
+        {:status (or (-> e ex-data :status-code)
```
crisptrutski, line 523: Since `ex-data` is called twice and `ex-message` is called twice, I wonder if it isn't worth pulling them both out in a `let`.

calherries: Yep. This isn't actually ready for review yet, I created this PR because I wanted to park it with a public PR page while I waited for input from Ryan on the FE. I generally add the team as reviewers when it's ready for review

src/metabase/upload.clj
```diff
@@ -590,19 +590,11 @@
...
-                            extra
-                            (tru "The CSV file contains extra columns that are not in the table: {0}."
-                                 (format-columns extra))
-                            missing
-                            (tru "The CSV file is missing columns that are in the table: {0}."
-                                 (format-columns missing)))]
-        (throw (ex-info error-message {:status-code 422}))))))
+      (throw (ex-info "There were some error uploading your CSV file."
+                      (-> {:status-code 422
+                           ::schema-mismatch? true}
```
crisptrutski, line 595: Is using exception metadata like this for control flow a common and ordained pattern in the API?

src/metabase/upload.clj
```diff
@@ -590,19 +590,11 @@
...
+                          (m/assoc-some :extra-columns extra)
+                          (m/assoc-some :missing-columns missing)))))))
```
crisptrutski, line 597: It looks like `assoc-some` can take multiple kv pairs, so these calls can be comined.

src/metabase/upload.clj
calherries: We don't have a common and ordained pattern. I'm open to suggestions for how to do this better.
See these issues for some discussion
https://github.com/metabase/metabase/issues/11175
https://github.com/metabase/metabase/issues/14683

iethree: oh haha, I just mentioned #14683 in slack 😂

calherries:
We're going in the direction of markdown now

PR #38005, opened 2024-01-22
update `can-read?` and `can-write?` for Field

johnswanson:
The previous behavior was to cache the permissions set for the Field's table. I split this in two: we cache the `db_id` for a given `Field`, but we can tackle caching the specific permissions for that database down the road, when we implement caching on data permissions checks in general.

To fix tests: `set-table-permissions!` previously threw an exception if passed an empty map of `table->permission-value`. I modified it to just skip processing in this case instead.

### DIFF
--- a/src/metabase/models/field.clj
+++ b/src/metabase/models/field.clj
@@ -4,15 +4,18 @@
    [clojure.set :as set]
    [clojure.string :as str]
    [medley.core :as m]
+   [metabase.api.common :as api]
    [metabase.db.connection :as mdb.connection]
    [metabase.lib.field :as lib.field]
    [metabase.lib.metadata.jvm :as lib.metadata.jvm]
+   [metabase.models.data-permissions :as data-perms]
    [metabase.models.dimension :refer [Dimension]]
    [metabase.models.field-values :as field-values :refer [FieldValues]]
    [metabase.models.humanization :as humanization]
    [metabase.models.interface :as mi]
-   [metabase.models.permissions :as perms]
    [metabase.models.serialization :as serdes]
+   [metabase.public-settings.premium-features
+    :as premium-features]
    [metabase.util :as u]
    [metabase.util.i18n :refer [trs tru]]
    [metabase.util.log :as log]
@@ -123,8 +126,6 @@
 
 (doto :model/Field
   (derive :metabase/model)
-  (derive ::mi/read-policy.partial-perms-for-perms-set)
-  (derive ::mi/write-policy.full-perms-for-perms-set)
   (derive :hook/timestamped?))
 
 (t2/define-before-insert :model/Field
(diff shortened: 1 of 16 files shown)


test/metabase/api/automagic_dashboards_test.clj
```diff
@@ -49,24 +49,19 @@
...
-         (let [api-endpoint (apply format (str "automagic-dashboards/" template) args)
-               resp         (mt/user-http-request :rasta :get 200 api-endpoint)
-               _            (dashcards-schema-check (:dashcards resp))
-               result       (validation-fn resp)]
-           (when (and result
-                      (try
-                        (testing "Endpoint should return 403 if user does not have permissions"
-                          (data-perms/set-database-permission! (perms-group/all-users) (mt/id) :perms/data-access :no-self-service)
-                          (perms/revoke-data-perms! (perms-group/all-users) (mt/id))
+       (let [api-endpoint (apply format (str "automagic-dashboards/" template) args)
```
escherize, line 52: Seems like `mt/with-full-data-perms-for-all-users!` got removed; was it un-needed or should we add an analogue to it?

src/metabase/models/field.clj
```diff
@@ -166,22 +161,38 @@
...
+
+(defmethod mi/can-write? :model/Field
+  ([instance]
+   (if (premium-features/enable-advanced-permissions?)
+     (= :yes (data-perms/table-permission-for-user
+              api/*current-user-id*
+              :perms/manage-table-metadata
+              (field->db-id instance)
+              (:table_id instance)))
+     (mi/superuser?)))
```
noahmoss, line 193: I think this check should be moved to EE code. Currently `perms/data-model-write-perms-path` tries to resolve `metabase-enterprise.advanced-permissions.models.permissions/data-model-write-perms-path` and calls into it when EE functionality is enabled; otherwise it requires admin perms to write fields.

Could you do the same here, maybe with a `defenterprise` macro? We should make sure that enforcement of EE-only permissions is done in EE namespaces so that it's covered by the correct license.

src/metabase/models/field.clj
```diff
@@ -147,14 +148,8 @@
...
 
-(defn- perms-objects-set*
-  [db-id schema table-id read-or-write]
-  #{(case read-or-write
-      :read  (perms/data-perms-path db-id schema table-id)
-      :write (perms/data-model-write-perms-path db-id schema table-id))})
-
-(def ^:private ^{:arglists '([table-id read-or-write])} cached-perms-object-set
-  "Cached lookup for the permissions set for a table with `table-id`. This is done so a single API call or other unit of
+(def ^:private ^{:arglists '([table-id])} cached-db-id
```
noahmoss, line 151: We can probably just use `metabase.models.table/table-id->database-id` for this

noahmoss (approved):

test/metabase/api/automagic_dashboards_test.clj
johnswanson: Good call, I'll add it back in, think I missed this in a rebase.

PR #38006, opened 2024-01-22
HK: remove useCallback in WidgetStatusIcon

oleggromov:
This addresses feedback for https://github.com/metabase/metabase/pull/37858 that was added after I merged it.

### DIFF
--- a/frontend/src/metabase/parameters/components/WidgetStatusIcon.tsx
+++ b/frontend/src/metabase/parameters/components/WidgetStatusIcon.tsx
@@ -1,4 +1,3 @@
-import { useCallback } from "react";
 import cx from "classnames";
 import { Icon } from "metabase/ui";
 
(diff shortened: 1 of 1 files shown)


frontend/src/metabase/parameters/components/WidgetStatusIcon.tsx
```diff
@@ -8,20 +7,16 @@ type Props = {
...
-  const handleOnClick = useCallback(
-    e => {
-      if (onClick) {
-        e.stopPropagation();
-        onClick();
-      }
-    },
-    [onClick],
-  );
+  const handleOnClick = (e: React.MouseEvent) => {
```
kamilmielnik, line 14: 1. We always destruct the React imports
2. `event` > `e`

```suggestion
  const handleOnClick = (event: MouseEvent) => {
```

kamilmielnik (approved):

Issue #37209, opened 2024-01-02
Metabase problem with `directConnection` connection setting for MongoDB


mhbahmani:
### Describe the bug

We recently updated our metabase from 0.45 to 0.48.1. The update was successful and the new version Metabase is up and running. However, we have some problems with MongoDB databases (Replicasets). In older versions of Metabase, we set one of the Mongo replica set members in the database creation page in the admin panel and all the Metabase queries ran on that specific Mongo instance. But after this update, we found out that Metabase is now connecting to the MongoDB instance and gives all the replica set instances, and run the queries on one of them. So there is no limitation on the Mongo node that Metabase would use for running its query. However, to separate Metabase queries load from our production queries, we want Metabase queries to be executed on that specific MongoDB instance.

To get this functionality in the 0.48.1 version of Metabase, we tried using [directConnection](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/connection/connection-options/) in the connection URI, but we're facing this warning log that shows this connection setting is not supported.
We tried using srv, setting database config via both connection string and filling out individual fields.

* The problem exists in both bare and docker installations.
* We tried installing a fresh Metabase instance and the problem still was there. So the update process should not be the root cause.

Our connection string:
`mongodb://username:*******@mongodb_instance_ip:27017/database_name?ssl=false&replicaSet=rs0&authSource=admin&directConnection=true`

Is that a bug in Metabase or we should get out the desired functionality in some other way?

### To Reproduce

1. Setup a Metabase instance with version 0.48.1
2. Define a MongoDB in Metabase databases and use of the replica set instances as the database host.
3. See that the queries are executing on all of the Mongodb nodes.

### Expected behavior

Queries to be executed on the specified node as the database host.

### Logs

`Jan  3 02:04:05 Metabase-Virtual-Machine-Hostname metabase[1965685]: 2024-01-03 02:04:05,180 WARN driver.uri :: Connection string contains unsupported option 'directconnection'.`

### Information about your Metabase installation

```JSON
- OS: Ubuntu 22.04
- Java Version:
`openjdk 11.0.21 2023-10-17
OpenJDK Runtime Environment (build 11.0.21+9-post-Ubuntu-0ubuntu122.04)
OpenJDK 64-Bit Server VM (build 11.0.21+9-post-Ubuntu-0ubuntu122.04, mixed mode, sharing)`
```

### Severity

[...]

PR #38017, opened 2024-01-22
Mongo java driver upgrade

### LINKED ISSUE #37713: Mongo sessions do not work with hidden member deployments
Mongo driver was updated recently to use sessions (PR #35680). As per [this](https://metaboat.slack.com/archives/C05MPF0TM3L/p1704217744503529) slack thread, sessions do not work correctly in deployments containing hidden members. According to [further discussion](https://metaboat.slack.com/archives/C04DN5VRQM6/p1704714526083359), problem is resolved in `mongo-java-driver` version `4.10`. As we are using `monger`, the underlying `mongo-java-driver` is of a version `3.12.11`. We should update `mongo-java-driver` we are using to newer version.

lbrdnk:
Closes #37713, #37209

### Description

With this PR merged, deployment using hidden members should work correctly. Problem was that old java driver, which was part of monger does not allow sessions to be used with hidden member deployments. Monger is now removed from dependencies.

Some namespaces were copied from monger however. Those should be either removed or further updated as part of follow-up work. Namespaces in question contains `TODO`s in docstrings.

### How to verify

Run the tests for `:mongo` driver. Hidden members deployment could be verified manually. Same holds for use of `directConnection` parameter

### Checklist


modules/drivers/mongo/test/metabase/driver/mongo/java_driver_wrapper_test.clj
lbrdnk, line 1: This file now contains tests from `mongo.util-test`.

modules/drivers/mongo/src/metabase/driver/mongo/util.clj
lbrdnk, line 1: Removed in favor of `java_driver_wrapper.clj`.

modules/drivers/mongo/src/metabase/driver/mongo/query_processor.clj
```diff
@@ -1370,7 +1371,10 @@
   "Parse a serialized native query. Like a normal JSON parse, but handles BSON/MongoDB extended JSON forms."
   [^String s]
   (try
-    (mapv (fn [^org.bson.BsonValue v] (-> v .asDocument com.mongodb.BasicDBObject.))
+    ;; TODO: Fixme! In following expression we were previously creating BasicDBObject's. As part of Monger removal
+    ;;       in favor of plain mongo-java-driver we now create Documents. I believe following conversion was and is
+    ;;       responsible for #38181.
+    (mapv (fn [^org.bson.BsonValue v] (-> v .asDocument org.bson.Document.))
```
lbrdnk, line 1377: I believe this is a source of #38181.

modules/drivers/mongo/src/metabase/driver/mongo/operators.clj
lbrdnk, line 1: This file is now only remnant of monger. It would make sense to remove it completely. It should be discussed first however. Specifically, if we are for the removal, do we want to use strings or keywords instead of symbols provided by this namespace. Based on the result, schemas in mongo.qp would have to be adjusted.

modules/drivers/mongo/src/metabase/driver/mongo/java_driver_wrapper.clj
```diff
@@ -0,0 +1,341 @@
...
+          (:conn-uri database)            database            ; connection URI has all the parameters
+          (:conn-uri (:details database)) (:details database)
+          :else
+          (throw (Exception. (str "with-mongo-connection failed: bad connection details:"
+                                  (:details database)))))]
+    (validate-db-details! db-details)
+    (update-ssl-db-details db-details)))
+
+;; TODO: ADD explanation from notion doc.
+(defn db-details->connection-string
```
lbrdnk, line 252: While WIP, more context on building connection string can be found in [this notion doc](https://www.notion.so/metabase/Mongo-without-monger-9f43f2c6a1e14be9bb49a82fd5d65eda?pvs=4). Its contents will be moved here, into docstring, later.

modules/drivers/mongo/src/metabase/driver/mongo/java_driver_wrapper.clj
```diff
@@ -0,0 +1,341 @@
...
+                           (fn [m ^String k]
+                             (assoc m (keyword k) (from-document (.get input k) true)))
+                           (fn [m ^String k]
+                             (assoc m k (from-document (.get input k) false))))
+                         (ordered-map/ordered-map)
+                         (.keySet input))))
+
+(defprotocol ConvertToDocument
+  (^org.bson.Document to-document [input] 
+    "Converts given piece of Clojure data to org.bson.Document usable by java driver."))
```
metamben, line 65: Are `String`s and `Boolean`s really `org.bson.Document`? That would be magic or worse.

modules/drivers/mongo/src/metabase/driver/mongo/conversion.clj
```diff
@@ -0,0 +1,139 @@
...
+                 (t/instant t))
+
+  org.bson.Document
+  (from-document [input {:keys [keywordize] :or {keywordize true} :as opts}]
+    (reduce (if keywordize
+              (fn [m ^String k]
+                (assoc m (keyword k) (from-document (.get input k) opts)))
+              (fn [m ^String k]
+                (assoc m k (from-document (.get input k) opts))))
+            (ordered-map/ordered-map)
```
metamben, line 57:
```suggestion
    (persistent! (reduce (if keywordize
                           (fn [m ^String k]
                             (assoc! m (keyword k) (from-document (.get input k) opts)))
                           (fn [m ^String k]
                             (assoc! m k (from-document (.get input k) opts))))
            (transient (ordered-map/ordered-map))
```

modules/drivers/mongo/src/metabase/driver/mongo/conversion.clj
```diff
@@ -0,0 +1,139 @@
...
+  (to-document [_input] nil)
+
+  String
+  (to-document [input] input)
+
+  Boolean
+  (to-document [input] input)
+
+  java.util.Date
+  (to-document [input] input)
```
metamben, line 74: These look quite superfluous given that they are `Object`s and they don't change the behavior.

modules/drivers/mongo/src/metabase/driver/mongo/database.clj
```diff
@@ -0,0 +1,62 @@
...
+  (:import
+   (com.mongodb ConnectionString)))
+
+(set! *warn-on-reflection* true)
+
+(defn- fqdn?
+  "A very simple way to check if a hostname is fully-qualified:
+   Check if there are two or more periods in the name."
+  [host]
+  (<= 2 (-> host frequencies (get \. 0))))
```
metamben, line 17: Where is this logic from?

modules/drivers/mongo/src/metabase/driver/mongo/database.clj
```diff
@@ -0,0 +1,62 @@
...
+          (integer? database)             (qp.store/with-metadata-provider database
+                                            (:details (lib.metadata.protocols/database (qp.store/metadata-provider))))
+          (string? database)              {:dbname database}
+          (:dbname (:details database))   (:details database) ; entire Database obj
+          (:dbname database)              database            ; connection details map only
+          (:conn-uri database)            database            ; connection URI has all the parameters
+          (:conn-uri (:details database)) (:details database)
+          :else
+          (throw (Exception. (str "with-mongo-connection failed: bad connection details:"
+                                  (:details database)))))]
```
metamben, line 50: Why not `ex-info`?

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -105,16 +107,16 @@
...
             (let [col-parts (str/split col-name #"\.")
                   val       (reduce
-                             (fn [^BasicDBObject object ^String part-name]
+                             (fn [^org.bson.Document object ^String part-name]
                                (when object
                                  (.get object part-name)))
                              row
                              col-parts)]
-              (m.conversion/from-db-object val :keywordize)))
+              (mongo.conversion/from-document val true)))
```
metamben, line 119: Just a naked boolean looks suspicious, shouldn't it be an options map?

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -146,8 +148,8 @@
    ^ClientSession session
    stages timeout-ms]
   (let [coll      (.getCollection db coll)
-        pipe      (m.util/into-array-list (m.conversion/to-db-object stages))
-        aggregate (.aggregate coll session pipe BasicDBObject)]
+        pipe      (java.util.ArrayList. ^java.util.Collection (mongo.conversion/to-document stages))
```
metamben, line 151: I suggest  importing these classes to make this line shorter.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
```diff
@@ -176,36 +178,19 @@
...
-
-(defn- kill-session!
-  [^MongoDatabase db
-   ^ClientSession session]
-  (let [session-id (.. session getServerSession getIdentifier)
-        kill-cmd (BasicDBObject. "killSessions" [session-id])]
-    (.runCommand db kill-cmd)))
-
 (defn execute-reducible-query
   "Process and run a native MongoDB query."
```
metamben, line 182: I suggest documenting when things expect `mongo.connection/*mongo-client*` to be set.

modules/drivers/mongo/src/metabase/driver/mongo/query_processor.clj
```diff
@@ -1370,7 +1370,10 @@
   "Parse a serialized native query. Like a normal JSON parse, but handles BSON/MongoDB extended JSON forms."
   [^String s]
   (try
-    (mapv (fn [^org.bson.BsonValue v] (-> v .asDocument com.mongodb.BasicDBObject.))
+    ;; TODO: Fixme! In following expression we were previously creating BasicDBObject's. As part of Monger removal
+    ;;       in favor of plain mongo-java-driver we now create Documents. I believe following conversion was and is
+    ;;       responsible for #38181.
```
metamben, line 1375: Can you make this more specific? Reading this, I'm not quite sure what is to be fixed (and how).

modules/drivers/mongo/src/metabase/driver/mongo/util.clj
```diff
@@ -1,262 +1,93 @@
...
+(defn do-find
+  "Perform find on collection."
+  [^MongoCollection coll
+   & {:keys [limit skip batch-size sort-criteria] :as opts}]
+  (->> (cond-> ^FindIterable (.find coll)
+         limit (.limit limit)
+         skip (.skip skip)
+         batch-size (.batchSize (int batch-size))
+         sort-criteria (.sort (mongo.conversion/to-document sort-criteria)))
+       (mapv #(mongo.conversion/from-document % (select-keys opts [:keywordize])))))
```
metamben, line 70: Wouldn't the following work?

```suggestion
       (mapv #(mongo.conversion/from-document % opts))))
```

modules/drivers/mongo/src/metabase/driver/mongo/util.clj
```diff
@@ -1,262 +1,93 @@
...
+
+(defn insert-one
+  "Insert document into mongo collection."
+  [^MongoCollection coll document-map]
+  (.insertOne coll (mongo.conversion/to-document document-map)))
+
+(defn start-session!
+  "Start session on client `c`."
+  ^ClientSession [^MongoClient c]
+  (.. c startSession))
```
metamben, line 85:
```suggestion
  (.startSession c))
```

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
```diff
@@ -0,0 +1,143 @@
...
+   :use-srv true})
+
+(deftest ^:parallel fqdn?-test
+  (testing "test hostname is fqdn"
+    (is (true? (#'mongo.db/fqdn? "db.mongo.com")))
+    (is (true? (#'mongo.db/fqdn? "replica-01.db.mongo.com")))
+    (is (false? (#'mongo.db/fqdn? "localhost")))
+    (is (false? (#'mongo.db/fqdn? "localhost.localdomain")))))
+
+;; TODO: remove this!
```
metamben, line 32: Why not just do it?

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
```diff
@@ -0,0 +1,143 @@
...
+
+(deftest ^:parallel srv-conn-str-test
+  (let [db-details {:user "test-user"
+                    :pass "test-pass"
+                    :host "test-host.place.com"
+                    :dbname "datadb"
+                    :authdb "authdb"
+                    :use-srv true}]
+    (testing "mongo+srv connection string used when :use-srv is thruthy"
+      (is (re-find #"\Qmongodb+srv://test-user:test-pass@test-host.place.com/datadb?authSource=authdb\E"
```
metamben, line 44: Isn't this what we usually write as `str/includes?`?

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
```diff
@@ -0,0 +1,143 @@
...
+                   (mongo.connection/db-details->connection-string db-details))))
+    (testing "Only fqdn may be used with mongo+srv"
+      (is (thrown-with-msg? Throwable
+                            #"\QUsing DNS SRV requires a FQDN for host\E"
```
metamben, line 48:
```suggestion
                            #"Using DNS SRV requires a FQDN for host"
```

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
```diff
@@ -0,0 +1,143 @@
...
+                            (-> db-details (assoc :host "localhost")
```
metamben, line 49:
```suggestion
                            (-> db-details
                                (assoc :host "localhost")
```

modules/drivers/mongo/test/metabase/driver/mongo/conversion_test.clj
```diff
@@ -0,0 +1,28 @@
+(ns metabase.driver.mongo.conversion-test
+  (:require
+   [clojure.test :refer :all]
+   [clojure.walk :as walk]
+   [flatland.ordered.map :as ordered-map]
+   [metabase.driver.mongo.conversion :as mongo.conversion]))
+
+(set! *warn-on-reflection* true)
+
+(deftest transformation-test
```
metamben, line 10:
```suggestion
(deftest ^:parallel transformation-test
```

modules/drivers/mongo/test/metabase/driver/mongo/database_test.clj
```diff
@@ -0,0 +1,19 @@
...
+   [metabase.driver.mongo.database :as mongo.db]))
+
+(deftest ^:parallel fqdn?-test
+  (testing "test hostname is fqdn"
+    (is (true? (#'mongo.db/fqdn? "db.mongo.com")))
+    (is (true? (#'mongo.db/fqdn? "replica-01.db.mongo.com")))
+    (is (false? (#'mongo.db/fqdn? "localhost")))
+    (is (false? (#'mongo.db/fqdn? "localhost.localdomain")))))
+
+(deftest db-name-test
```
metamben, line 13:
```suggestion
(deftest ^:parallel db-name-test
```

src/metabase/driver/util.clj
```diff
@@ -596,15 +596,20 @@
...
   (let [ssl-context (SSLContext/getInstance "TLS")]
     (.init ssl-context
            (when (and private-key own-cert) (key-managers private-key (str (random-uuid)) own-cert))
            (when trust-cert (trust-managers trust-cert))
            nil)
-    (.getSocketFactory ssl-context)))
+    ssl-context))
+
+(defn ssl-socket-factory
+  "Generates an `SocketFactory` with the custom certificates added"
```
metamben, line 610: Here, and in the previous docstring:
```suggestion
  "Generates a `SocketFactory` with the custom certificates added"
```

metamben (approved):

camsaul (approved):

modules/drivers/mongo/src/metabase/driver/mongo.clj
```diff
@@ -53,18 +45,18 @@
...
-       ;; 1. check db.dbStats command completes successfully
-       (= (float (:ok db-stats))
-          1.0)
-       ;; 2. check the database is actually on the server
-       ;; (this is required because (1) is true even if the database doesn't exist)
-       (contains? db-names (:db db-stats))))))
+  [_ db-details]
+  (mongo.connection/with-mongo-client [^MongoClient c db-details]
+    (let [db-names (mongo.util/list-database-names c)]
+      (mongo.connection/with-mongo-database [^MongoDatabase db db-details]
```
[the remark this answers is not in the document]
lbrdnk: `MongoDatabase` class does not provide getter for `MongoClient` it was created with. `with-mongo-database` is meant as utility to avoid `(let [db (mongo.util/database client ...)` in calls of `with-mongo-client`, because we are using database without the need for client in various places.

As we need both in this function, I've modified the code to use the `let` form.

modules/drivers/mongo/src/metabase/driver/mongo.clj
```diff
@@ -200,65 +192,45 @@
 
 (defmethod driver/describe-database :mongo
   [_ database]
-  (with-mongo-connection [^com.mongodb.DB conn database]
-    {:tables (set (for [collection (disj (mdb/get-collection-names conn) "system.indexes")]
+  (mongo.connection/with-mongo-database [^MongoDatabase db database]
+    ;; TODO: factor next line!
+    {:tables (set (for [collection (disj (into #{} (mongo.util/list-collection-names db)) "system.indexes")]
```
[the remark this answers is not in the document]
lbrdnk: I think `into` should rather be used when I do some transformation to the source collection. Thanks.

Here set operations are used just to remove one element from a set. Form in question was simplified, to use `(for [... :when...]` instead of set operations.

modules/drivers/mongo/src/metabase/driver/mongo/connection.clj
```diff
@@ -0,0 +1,120 @@
...
+     host
+     (when (and (not use-srv) (some? port)) (str ":" port))
+     "/"
+     dbname
+     "?authSource=" (if (empty? authdb) "admin" authdb)
+     "&appName=" config/mb-app-id-string
+     "&connectTimeoutMS=" (driver.u/db-connection-timeout-ms)
+     "&serverSelectionTimeoutMS=" (driver.u/db-connection-timeout-ms)
+     (when ssl "&ssl=true")
+     (when (seq additional-options) (str "&" additional-options)))))
```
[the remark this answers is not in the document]
lbrdnk: I've decided to create connection string first and then apply it to `MongoClientSettings$Builder` using its `.applyConnectionString` method in `mongo.connection/db-details->mongo-client-settings`. Then, if we are setting some options that are not exposed via connection string (currently only ssl context), those are set directly on builder.

I've resorted to this solution as I find it simpler than set everything directly through builders. With new driver version there are actually 6 different builder classes involved in creation of `MongoClientSettings` (`SslSettings$Builder` used for setting ssl context is just one of them).

[Docs for MongoClientSettings$Builder](http://mongodb.github.io/mongo-java-driver/4.11/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html)

modules/drivers/mongo/src/metabase/driver/mongo/conversion.clj
lbrdnk: Thank you.

modules/drivers/mongo/src/metabase/driver/mongo.clj
```diff
@@ -200,65 +192,45 @@
...
+                    :value (str/split % #"\.")}
+                   {:type  :normal-column-index
+                    :value %}))
+           set))))
+
+(defn- sample-documents [^MongoDatabase db table sort-direction]
+  (let [coll (mongo.util/collection db (:name table))]
+    (mongo.util/do-find coll {:limit metadata-queries/nested-field-sample-limit
+                             :skip 0
+                             :sort-criteria {:_id sort-direction}
```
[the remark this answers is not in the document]
lbrdnk: I've used map instead of ordered map here because there is only one key set.

I've updated the `do-find` function handle either ordered map or sequence of kv pairs as argument for `:sort-criteria`. Also the docstring was upadted accordingly.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
lbrdnk: Done.

modules/drivers/mongo/src/metabase/driver/mongo/execute.clj
lbrdnk: Done.

modules/drivers/mongo/src/metabase/driver/mongo/query_processor.clj
lbrdnk: Done.

modules/drivers/mongo/src/metabase/driver/mongo/util.clj
lbrdnk: I wanted to separate `from-document` params from those of `do-find` to keep `from-document`s input "clean". That would make sense if there were using more parameters.

modules/drivers/mongo/src/metabase/driver/mongo/util.clj
lbrdnk: Done.

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
lbrdnk: Thanks.

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
lbrdnk: `re-find` is definitely not necessary here. Thanks.

modules/drivers/mongo/test/metabase/driver/mongo/connection_test.clj
lbrdnk: Done.

PR #37870, opened 2024-01-18
Reapply "Add validation for remaining setting types (#37519)" (#37809)

crisptrutski:
### Description

This reverts the near aP0calytic change which adds validation of the app settings during start-up.

It should be safe now after the preceding refactors.

PR #37863, opened 2024-01-18
Allow :type to be a map of inherited options for defsetting

crisptrutski:
Closes: https://github.com/metabase/metabase/issues/38250

### Description

This is a pure refactor which remove the repetition of the required "uuid-nonce" fields. It recovers the encapsulation previously given by the `::uid-nonce` type, but in a more terse way that also covers the read-only and non-audited aspects.

PR #38027, opened 2024-01-23
Disable auditing when `site-uuid` nonces are initialized

### LINKED ISSUE #38009: Initializing MB from a config file is broken
### Describe the bug

If you start up a new instance of Metabase with `MB_CONFIG_FILE_PATH` set, and `MB_PREMIUM_EMBEDDING_TOKEN` set to a token that provides the `config-text-file` feature, it will nevertheless error with:
```
Execution error (ExceptionInfo) at metabase-enterprise.advanced-config.file/initialize! (file.clj:261).
Metabase config files require a Premium token with the :config-text-file feature.
`````

### To Reproduce

1. Set `MB_CONFIG_FILE_PATH` as described [here](https://www.metabase.com/docs/latest/configuring-metabase/config-file)
2. Set `MB_PREMIUM_EMBEDDING_TOKEN` to a token with full features
3. Start up Metabase with a fresh app DB

### Expected behavior

Should not error

### Logs

_No response_

```JSON
Current master
```

### Severity

unsure, feature might be totally busted

### Additional context

_No response_

crisptrutski:
Fixes https://github.com/metabase/metabase/issues/38009

### Description

With the [ntroduction of the [init mechanism](https://github.com/metabase/metabase/pull/37849/files) for app settings, we introduced a subtle change - the initialization of `site-uuid` settings is now audited.

Unfortunately this change broke the loading of the Enterprise feature flags when first populating the app database, due to a concurrency bug in the auditing code. This resulted in the app silently falling back to a "no feature flags" state. In the case where we are using a config file this issue is then exposure by an immediate crash during startup.

The details of the concurrency bug are quite detailed - see [this comment](https://github.com/metabase/metabase/issues/38009#issuecomment-1906910158) on the parent issue.

This PR simply reverts to the previous behavior - no auditing. 

### How to verify

1. Drop all tables in the app database.
2. Set `MB_CONFIG_FILE_PATH` to point to a config file. A placeholder is fine, it won't be loaded.
3. Set a `MB_PREMIUM_EMBEDDING_TOKEN`. It must have the right format, but won't be used.
4. Start the server.
5.  🌈 

Without this change you should get the following instead:

```
2024-01-23 06:02:29,724 ERROR metabase.core :: Metabase Initialization FAILED
clojure.lang.ExceptionInfo: Metabase config files require a Premium token with the 
  :config-text-file feature. {}
```

### DIFF
--- a/src/metabase/analytics/snowplow.clj
+++ b/src/metabase/analytics/snowplow.clj
@@ -76,6 +76,7 @@
   :visibility :public
   :type       :string
   :setter     :none
+  :audit      :never
   :init       setting/random-uuid-str
   :doc        false)
 
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -511,6 +511,13 @@
                 (db-value setting)
                 (core/get cache (setting-name setting-definition-or-name))))))))))
 
+(defn- underlying-setter
+  "In the case of a read-only setting, this gives raw access to write a new value. Useful for initialization."
+  [{:keys [setter] setting-type :type :as setting}]
+  (if (= :none setter)
+    (partial set-value-of-type! setting-type setting)
+    setter))
+
 (defonce ^:private ^ReentrantLock init-lock (ReentrantLock.))
 
 (defn- init! [setting-definition-or-name]
@@ -527,7 +534,11 @@
             (read-setting setting)
             (when init
               (when-let [init-value (init)]
-                (metabase.models.setting/set! setting init-value :bypass-read-only? true))))
+                ;; We use the underlying setter to avoid triggering the audit logs.
+                ;; This is currently needed to avoid a deadlock when the database is first initialized via a config file.
+                (let [setter (underlying-setter setting)]
+                  (setter init-value))
+                #_(metabase.models.setting/set! setting init-value :bypass-read-only? true))))
           (finally
             (.unlock init-lock)))))))
 
@@ -878,9 +889,9 @@
 
 (defn- set-with-audit-logging!
   "Calls the setting's setter with `new-value`, and then writes the change to the `audit_log` table if necessary."
-  [{:keys [getter audit setter] setting-type :type :as setting} new-value bypass-read-only?]
-  (let [setter (if (and bypass-read-only? (= :none setter))
-                 (partial set-value-of-type! setting-type setting)
+  [{:keys [getter audit setter] :as setting} new-value bypass-read-only?]
+  (let [setter (if bypass-read-only?
+                 (underlying-setter setting)
                  setter)]
     (if (should-audit? setting)
       (let [audit-value-fn #(condp = audit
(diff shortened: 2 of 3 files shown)


qnkhuat (commented):
are you sure we don't want to audit all the initialized values? It's a set so it feels like we should.

wdyt if we fix this by making `site-uuid-for-premium-features-token-checks` to never audit?
that should fix the deadlock issue.

qnkhuat:
And sorry that I didn't review the earlier PR that you introduced, `:init,` but reviewing this PR got me to read about them, and I think it'd really help if we have a bit more document. especially for the `:init` option for `defsetting` and `bypass-read-only?` option for the `set!` function.

crisptrutski:
> are you sure we don't want to audit all the initialized values? It's a set so it feels like we should.
> 
> wdyt if we fix this by making `site-uuid-for-premium-features-token-checks` to never audit? that should fix the deadlock issue.

Right now the UUIDs are the only settings using `:init`, and in the old implementation they use the same non-audited setters to init as in this PR. I wanted to be as conservative as possible and just go back to the old semantics for now, to unblock the next `0.48` release.

I'd like to get more input about whether we should enable or explicitly opt-out of audit for these uuids, and prefer not to pick a side in the codebase yet. We don't audit "default" so I think it's also OK if you think of init just as a "per site" default.

As far as the deadlock goes, there's an [independent fix](https://github.com/metabase/metabase/pull/38029) for that which allows us to keep auditing if we want, I'm hoping to merge both.

crisptrutski:
> And sorry that I didn't review the earlier PR that you introduced `:init,` but reviewing this PR got me to read about them, and I think it'd really help if we have a bit more document. especially for the `:init` option for `defsetting` and `bypass-read-only?` option for the `set!` function.

Thanks for pointing out those spots. I plan to do a small clean up PR which also tightens up the schema, so I can update those docstrings at the same time.

qnkhuat:
I think there is a subtle difference between init and default is that `init` will write to DB.
And I think audit_log is meant to track these writes, I'm not sure how important it is if it's the first write tho. Worth checking with @luizarakaki or @noahmoss

crisptrutski:
After discussing with @luizarakaki I've changed this PR to explicitly . Since the race condition has been fixed in [another PR](https://github.com/metabase/metabase/pull/38029), I have re-enabled auditing for any future usage of `:init`.

crisptrutski:
* `master`
  * **PR #38027** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38027?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #37863** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #37870** <a href="https://app.graphite.dev/github/pr/metabase/metabase/37870?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38027?utm_source=stack-comment).

qnkhuat (approved):
Cool, LGTM!

PR #38029, opened 2024-01-23
Fix potential deadlock when loading feature flags

### LINKED ISSUE #38009: Initializing MB from a config file is broken

crisptrutski:
Fixes https://github.com/metabase/metabase/issues/38009

### Description

This makes the feature flag loading code more robust by making the function re-entrant. To understand how this function could deadlock before, see the description in https://github.com/metabase/metabase/pull/38027.

It works by ensuring that the call to initialize the nonce comes from the same thread that initially locked on fetching the tokens.

There is a subtle behavior change here - was allow for a 10s timeout on each call individually now. Since it is unlikely that both endpoints will timeout consecutively, and the combined wait time is still fairly short, I think this is OK.

### How to verify

You can use the same steps as https://github.com/metabase/metabase/pull/38027, either of these PRs are sufficient to fix the bug.

### DIFF
--- a/src/metabase/public_settings/premium_features.clj
+++ b/src/metabase/public_settings/premium_features.clj
@@ -133,21 +133,23 @@
     (let [site-uuid (setting/get :site-uuid-for-premium-features-token-checks)]
       (try (fetch-token-and-parse-body token token-check-url site-uuid)
            (catch Exception e1
-             (log/error e1 (trs "Error fetching token status from {0}:" token-check-url))
-             ;; Try the fallback URL, which was the default URL prior to 45.2
-             (try (fetch-token-and-parse-body token store-url site-uuid)
-                  ;; if there was an error fetching the token from both the normal and fallback URLs, log the
-                  ;; first error and return a generic message about the token being invalid. This message
-                  ;; will get displayed in the Settings page in the admin panel so we do not want something
-                  ;; complicated
-                  (catch Exception e2
-                    (log/error e2 (trs "Error fetching token status from {0}:" store-url))
-                    (let [body (u/ignore-exceptions (some-> (ex-data e1) :body (json/parse-string keyword)))]
-                      (or
-                        body
-                        {:valid         false
-                         :status        (tru "Unable to validate token")
-                         :error-details (.getMessage e1)})))))))))
+             ;; Unwrap exception from inside the future
+             (let [e1 (ex-cause e1)]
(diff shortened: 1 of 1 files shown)


src/metabase/public_settings/premium_features.clj
```diff
@@ -117,32 +130,26 @@
...
-                                      :error-details (.getMessage e1)})))))))
-          result (deref fut fetch-token-status-timeout-ms ::timed-out)]
-      (if (= result ::timed-out)
-        (do
-          (future-cancel fut)
-          {:valid         false
-           :status        (tru "Unable to validate token")
-           :error-details (tru "Token validation timed out.")})
-        result))))
+    (let [site-uuid (setting/get :site-uuid-for-premium-features-token-checks)]
```
qnkhuat, line 133: I'd add a comment here on why we want this here, you could link it to the other PR's description you have.

qnkhuat (approved):
The fix works and it makes sense to me, so I'm approving.

It'd be nice if you can find a way to write a test for this, it'd be hard I think so it's up to you.

crisptrutski:
### Merge activity

* **Jan 23, 8:59 AM**: @crisptrutski started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38029).

noahmoss:
@crisptrutski Does this need to be backported?

crisptrutski:
> @crisptrutski Does this need to be backported?

Strictly speaking we'll be OK with just https://github.com/metabase/metabase/pull/38027 being backported, but I'll sleep better if this is backported as well. Will try run the github action again.

PR #37436, opened 2024-01-09
Migrate `sourceTableId`, `sourceTable` to use `sourceTableOrCardId` from MLv2

kamilmielnik:
Closes #37170

PR #38046, opened 2024-01-23
Optimize MLv2 usage in parameters auto-wiring code

### LINKED ISSUE #37875: Adding a new question to a busy dashboard can be really slow
### Describe the bug

Adding a new question to a dashboard with a few filters and other dashcards can take a long take (seconds or longer)

### To Reproduce

1. Go to browse data (`/browse`) and click the Sample Database
2. Hover the "Orders" table, click the x-ray button (yellow bolt icon) and save the dashboard
3. On that dashboard, enter the editing mode and try to add a new question about Orders
4. The UI freezes for a few seconds

### Expected behavior

The UI shouldn't freeze and a new dashcard should show up instantly

### Logs

_No response_

```JSON
master (commit: 162349c)
```

### Severity

P2? IMO it'd be a blocker for 49

### Additional context

Metabase 48 doesn't seem to be affected

kulyk:
Closes #37875 

Fixes that adding a new question on a busy dashboard was taking seconds (~8s for an x-rayed sample orders table)

The performance drop appeared in #37436 when we added a [MLv1-MLv2 query conversion](https://github.com/metabase/metabase/blob/fa22eb83440162cfa5c62b03fcdfe46b081412a9/frontend/src/metabase-lib/queries/utils/structured-query-table.ts#L23) to `StructuredQuery.prototype.table()`. Here's how we get to it:

1. We're adding a new question to a dashboard with [`addCardToDashboard` redux action](https://github.com/metabase/metabase/blob/0cfa393d0bd0c26761e99e3dd7b7993c1eba2ccb/frontend/src/metabase/dashboard/actions/cards.js#L37)
2. At the end of `addCardToDashboard`, [we dispatch `autoWireParametersToNewCard`](https://github.com/metabase/metabase/blob/0cfa393d0bd0c26761e99e3dd7b7993c1eba2ccb/frontend/src/metabase/dashboard/actions/auto-wire-parameters/actions.ts#L83C17-L83C44)
3. `autoWireParametersToNewCard` [is calling `compareMappingOptionTargets` in a nested loop](https://github.com/metabase/metabase/blob/0cfa393d0bd0c26761e99e3dd7b7993c1eba2ccb/frontend/src/metabase/dashboard/actions/auto-wire-parameters/actions.ts#L127C11-L127C38) (a number of columns suitable for a given dashboard filter * number of dashcards)
4. `compareMappingOptionTargets` [instantiates a `Question`](https://github.com/metabase/metabase/blob/0cfa393d0bd0c26761e99e3dd7b7993c1eba2ccb/frontend/src/metabase-lib/parameters/utils/targets.ts#L62) internally. It actually happens two times (for the "target dashcard question" and the compared dashcard question)

So if we had 30 mappable columns and 15 dashcards, we'd instantiate a `Question` around 30 * 15 * 2 = 900 times. Because all of them are different instances, we can't benefit from the MLv2 query cache on the question instance.

Fixed by:
- switching `compareMappingOptionTargets` to `getParameterTargetField` that accepts a question instead of a card + metadata
- updating `autoWireParametersToNewCard` so we instantiate the target question once vs. every iteration
- updating `autoWireParametersToNewCard` so we instantiate dashcard questions upfront vs. every iteration

### Demo

**Before**
perf-before

**After**
(varies between ~500ms — ~1.2s for me)
perf-after

### DIFF
--- a/frontend/src/metabase-lib/parameters/utils/cards.ts
+++ b/frontend/src/metabase-lib/parameters/utils/cards.ts
@@ -28,7 +28,7 @@ export function getCardUiParameters(
     const target: ParameterTarget | undefined = (
       parameter as ParameterWithTarget
     ).target;
-    const field = getParameterTargetField(target, metadata, question);
+    const field = getParameterTargetField(target, question);
     if (field) {
       return {
         ...parameter,
--- a/frontend/src/metabase-lib/parameters/utils/targets.ts
+++ b/frontend/src/metabase-lib/parameters/utils/targets.ts
@@ -23,13 +23,13 @@ export function getTemplateTagFromTarget(target: ParameterTarget) {
 
 export function getParameterTargetField(
   target: ParameterTarget,
-  metadata: Metadata,
   question: Question,
 ) {
   if (isDimensionTarget(target)) {
     const query = question.legacyQuery({ useStructuredQuery: true }) as
       | NativeQuery
       | StructuredQuery;
+    const metadata = question.metadata();
     const dimension = Dimension.parseMBQL(target[1], metadata, query);
 
     return dimension?.field();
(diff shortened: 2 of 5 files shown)


kamilmielnik (approved):
Great improvement :+1: :star:

uladzimirdev:
@kulyk thanks, nice catch! I'd like to understand perf problems in advance during MLv2 migration, so did I get it right, creating new Question is a very heavy task and we must avoid it if possible?

kulyk:
> I'd like to understand perf problems in advance during MLv2 migration, so did I get it right, creating new Question is a very heavy task and we must avoid it if possible?

@uladzimirdev, I'm not sure how heavy it is, but it's way heavier than instantiating a question in MLv2. We're constructing a metadata provider, and the `dataset_query` gets validated and converted to another MBQL format internally (pMBQL). Maybe something else is going on. So if it's called hundreds of times in a nested loop or something, it can easily impact performance

I think it's helpful to think about MLv2 calls as overall expensive. So it's often a good idea to compute some things ahead of time instead of doing repeated inline calls, etc.

PR #38053, opened 2024-01-23
Navbar Improvements: Embedding

sloansparger:
### Description

This changes prevents the logo button from being pressed when user has configured their embed with the `top_nav` option to `true` and the `side_nav` to `false`. This will prevent users from getting stuck on the home page with no way back.

This is the last change for the work to complete https://github.com/metabase/metabase/issues/36370

### How to verify

- create interactive embed with the `top_nav=true&side_nav=false` query params, make sure the page set is any page but the home page
- try to click the logo (nothing should happen)

### Demo

Screenshot 2024-01-23 at 11 29 05 AM
(you can see the link is disabled)

### Checklist


npfitz:
I'm not sure if this is really in scope of this change, but things get unfortunate with the mobile view if TopNav is false and Sidebar is true. 
image

npfitz (approved):

PR #38054, opened 2024-01-23
Make the MLv2 display-name include the path of nested fields

### LINKED ISSUE #35938: MongoDB nested fields not showing in v0.47.x
### Describe the bug

MongoDB nested fields were not showing up in the query summary section in **v0.47.x**. It was working fine previously in **v0.46.x**.

Both versions are connected to the same database.

Check the screenshots below to see the differences between the versions:

**v0.46.x**
image

**v0.47.x**
image

### To Reproduce

Connect Metabase to a MongoDB database with documents that have nested JSON fields.

### Expected behavior

_No response_

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "Win32",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.version": "11.0.21",
    "os.name": "Linux",
    "os.version": "5.10.0-23-amd64",
    "user.language": "en",
    "user.timezone": "UTC"
  },
  "metabase-info": {
    "databases": [
      "mongo"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "16.1"
      },
      "jdbc-driver": {
        "version": "42.5.4"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2023-11-17",
      "tag": "v0.47.8",
      "branch": "?",
      "hash": "639fa5e"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

It is preventing to use the new version of Metabase.

### Additional context

_No response_

metamben:
Fixes #35938.

According to the [discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1705936770497669?thread_ts=1705933928.867939&cid=C04CYTEL9N2) the idea of this PR is to make the default display-name include the whole path for nested fields.

~I change the display-name property of fields right when they are returned by the metadata-provider.~

Since display-names are constructed recursively, doing the conversion later leads to the nesting logic applied several times or very complicated and brittle code to prevent multiple nesting. I've made an attempt to implement this. The result is quite ugly, an ideas how to make it better or at least less brittle are welcom.

Changing the display-name generated by the QP would have blown up the system, I think.

Currently, the original display-name property gets lost. I can add a copy of that under a different name (e.g., simple-display-name, local-display-name, or something similar) if we deem it necessary.


camsaul (changes requested):
We already have prior art for solving a problem like this -- we have `metabase.lib.metadata.calculation/*display-name-style*` that gets passed to `display-name-method` -- it seems to me like that `nest-display-name` stuff you're doing should be done that way, in the implementation for `:metadata/column`, rather than here. 

I really don't like the idea of suddenly having two different versions of `field` now, where the only difference is which version of `display-name` they return. It's not a sustainable pattern if we want to do the same for a different key or support a third option, or if you need to pass this option in to a function that ends up calling `lib.metadata/field`...

It's also going to make this hard to use because people will have to remember or guess which version to use, and if you need to change the version you might have to end up changing the function calls in a bunch of other places that might end up breaking stuff. 

So with a dynamic variable at least you don't need to have multiple versions of the same function. The main reason I put that stuff in there was so the QP could maintain its own different behavior than what MLv2 returned. So you should be able to build on that and have MLv2 do one thing by without breaking QP code.

I still think the dynamic variable is a little inconvenient to use, and maybe we should just refactor things so the metadata includes all the different variations of `:display-name` you might want by default. But that's a bigger project

metamben:
> I still think the dynamic variable is a little inconvenient to use, and maybe we should just refactor things so the metadata includes all the different variations of `:display-name` you might want by default. But that's a bigger project



> We already have prior art for solving a problem like this -- we have `metabase.lib.metadata.calculation/*display-name-style*` that gets passed to `display-name-method` -- it seems to me like that `nest-display-name` stuff you're doing should be done that way, in the implementation for `:metadata/column`, rather than here.

I agree. TBH, I didn't really mean the `with-nested` stuff is as a real solution. I just added that last commit to see if the problem can be "solved" at all this way. I can change `field` to respect that dynamic var, and it would work (for now), but even that seems to me a stupid solution. How to do it? Add a binding for `metabase.lib.metadata.calculation/*display-name-style*` around (almost) each and every MLv2 call in `js.cljs`? Add them around the calls that I changed to `-with-nested`? At some call sites I'm not sure that it's going to be used by both FE and BE...

I think the real solution is a separate display-name version that is not shared with the QP. @camsaul Is the solution with the dynamic var worth it as a short term solution, do you think?

test/metabase/lib/field_test.cljc
```diff
@@ -684,7 +739,7 @@
                :semantic-type :type/Category,
                :is-from-join false,
                :long-display-name "Price: Auto binned",
-               :display-name "Price",
+               :display-name "Price: Auto binned",
```
metamben, line 742: I think this is the proper display name for a binned column, so I consider this is a bug fix.

src/metabase/lib/field.cljc
```diff
@@ -228,13 +259,18 @@
...
+                             (lib.util/format "%s: %s" display-name (lib.binning/binning-display-name binning field-metadata)))]
+    ;; temporal unit and binning formatting are only applied if they haven't been applied yet
     (cond
-      temporal-unit (lib.util/format "%s: %s" display-name (-> (name temporal-unit)
-                                                               (str/replace \- \space)
-                                                               u/capitalize-en))
-      binning       (lib.util/format "%s: %s" display-name (lib.binning/binning-display-name binning field-metadata))
-      :else         display-name)))
+      (and temporal-unit (not= display-name (temporal-format humanized-name))) (temporal-format display-name)
+      (and binning       (not= display-name (bin-format humanized-name)))      (bin-format display-name)
```
metamben, line 272: This is definitely fragile. Any better way to track if the temporal/binning formatting has already been applied?

metamben:
@camsaul, here's an attempt at a local change. (I don't like this either.)

camsaul (approved):
Changes look good to me, I think including both versions of the display name in the metadata is probably the best way forward in the future (hopefully we can get rid of that dynamic variable eventually). Let's get this in to keep things moving and we can think about how to make it more robust in the future

PR #38058, opened 2024-01-23
[MLv2] Rewrite `QuestionDataSoruce` FE unit test

nemanjaglumac:
This PR is a part of https://github.com/metabase/metabase/pull/37886, but I decided to split it in order to make the review easier. This commit alone is huge.

This test was failing for the following case:
> "shows nothing if a user doesn't have data permissions"

The reason behind the failure is the change in the logic that we use to determine whether a query is editable or not (using MLv2). Test was previously resetting the database ID to achieve this. It was a hacky solution.
```js
const databaseId = question.card().dataset_query.database;
question.card().dataset_query.database = null;
// ...test
question.card().dataset_query.database = databaseId;
```

The new approach reflects the real world scenario. It mocks the `metadata` object. If there's no `database` in it, a user doesn't have data permissions.

This was challenging to achieve using the existing test logic, primarily because all questions were initialized prior to the `setup()` call. There was no easy way to parameterize those question functions, and to conditionally remove `database` from the metadata.

After spending hours trying to accommodate the old code, I decided to alter it in the following way:
- Test cases now only contain the card data, and not the full question function reference
- `card` is passed to the `setup()` function
- A question is being constructed inside of the `setup()` function using the card prop
- `metadata` is being conditionally constructed inside of the `setup()` function

### DIFF
--- a/frontend/src/metabase/query_builder/components/view/QuestionDataSource.unit.spec.js
+++ b/frontend/src/metabase/query_builder/components/view/QuestionDataSource.unit.spec.js
@@ -175,28 +175,12 @@ function getMetadata() {
   });
 }
 
-function getQuestion(card) {
-  return new Question(card, getMetadata());
+function getQuestion(card, metadata) {
+  return new Question(card, metadata);
 }
 
 function getAdHocQuestion(overrides) {
   return getQuestion({ ...BASE_GUI_QUESTION, ...overrides });
-}
-
-function getNativeQuestion() {
-  return getQuestion(BASE_NATIVE_QUESTION);
-}
-
-function getSavedGUIQuestion(overrides) {
-  return getQuestion({ ...BASE_GUI_QUESTION, ...SAVED_QUESTION, ...overrides });
-}
-
-function getSavedNativeQuestion(overrides) {
-  return getQuestion({
-    ...BASE_NATIVE_QUESTION,
-    ...SAVED_QUESTION,
-    ...overrides,
-  });
 }
 
 function getAdHocOrdersQuestion() {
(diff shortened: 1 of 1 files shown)


frontend/src/metabase/query_builder/components/view/QuestionDataSource.unit.spec.js
```diff
@@ -282,7 +230,16 @@ class ErrorBoundary extends Component {
...
-function setup({ question, subHead = false, isObjectDetail = false } = {}) {
+function setup({
+  card,
+  subHead = false,
+  isObjectDetail = false,
+  hasPermissions = true,
+  isMultiSchemaDB = false,
+} = {}) {
+  const metadata = hasPermissions ? getMetadata() : {};
+  const question = card && new Question(card, metadata);
```
nemanjaglumac, line 241: This is the main change.
We're constructing both the `metadata` and the `question` inside of the setup. This gives us greater control.

frontend/src/metabase/query_builder/components/view/QuestionDataSource.unit.spec.js
```diff
@@ -305,100 +262,95 @@ jest.mock("metabase/core/components/Link", () => ({ to: href, ...props }) => (
...
+        ...BASE_GUI_QUESTION,
+        ...SAVED_QUESTION,
         dataset_query: QUERY_WITH_PRODUCTS_JOIN,
-      }),
+      },
       questionType: "saved GUI question joining a table",
     },
     AD_HOC_PRODUCTS_JOIN: {
-      question: getAdHocQuestion({ dataset_query: QUERY_WITH_PRODUCTS_JOIN }),
+      card: { ...BASE_GUI_QUESTION, dataset_query: QUERY_WITH_PRODUCTS_JOIN },
```
nemanjaglumac, line 281: This line shows how we avoid one unnecessary jump with this new change. What we really want to do here is construct the card data that we'll feed later to the question.

Before the change I found myself constantly switching context by jumping to different function calls, just to figure out what are they doing.

frontend/src/metabase/query_builder/components/view/QuestionDataSource.unit.spec.js
```diff
@@ -282,7 +230,16 @@ class ErrorBoundary extends Component {
 }
 const SOURCE_CARD = createMockCard({ id: SOURCE_QUESTION_ID });
 
-function setup({ question, subHead = false, isObjectDetail = false } = {}) {
+function setup({
+  card,
+  subHead = false,
+  isObjectDetail = false,
+  hasPermissions = true,
+  isMultiSchemaDB = false,
```
kamilmielnik, line 238: Unused

frontend/src/metabase/query_builder/components/view/QuestionDataSource.unit.spec.js
```diff
@@ -282,7 +230,16 @@ class ErrorBoundary extends Component {
...
+} = {}) {
+  const metadata = hasPermissions ? getMetadata() : {};
```
kamilmielnik, line 240: 1. Let's make sure `metadata` is a valid instance of `Metadata`

```suggestion
  const metadata = hasPermissions ? getMetadata() : createMockMetadata({});
```

2. I'd make `metadata` a parameter of `setup` (instead  `hasPermissions`) - this gives even greater control. But maybe YAGNI.

kamilmielnik (approved):
:+1:

Issue #38056, opened 2024-01-23
Sorting Dashboard Tabs has a weird animation on drop


npfitz:
### Describe the bug

Just a minor visual issue when dropping tabs in order

### To Reproduce

1. Go to a dashboard with tabs
2. Click edit and re-order the tabs
3. See how the tab animation looks a little weird.
![chrome_S6ODqrb4aa](https://github.com/metabase/metabase/assets/1328979/d7cd9d14-3fea-46e4-ae1e-19d2a9d7f5f4)

### Information about your Metabase installation

```JSON
Currently in stats
```

### Severity

minor

PR #38059, opened 2024-01-23
37925 parameters list dnd kit

### LINKED ISSUE #37925: Migrate Parameters List (dashboard and native queries) to DnD Kit

npfitz:
Closes https://github.com/metabase/metabase/issues/37925

### Description
This PR updates the `<ParametersList/>` component to use dnd-kit for it's sorting. It also no longer renders different components based on the isEditing prop. We always render a sortable list, but we disable sorting if isEditing is set to false.

This PR also introduces the `useSortableList` hook. This was largely added to store (but hide) a bit of state to help dragging and dropping. 2 things that dnd kit doesn't necessarily handle too well are sorting items that are different sizes, and having the state that determines the order of sortable items take too long to update after dropping. In this case, dispatching redux actions to update state seemed to take too long to update state (this is the core issue with #38056 ). 

This hook gets around those issues by keeping it's own state, which is the order of ids. While an item is being sorted we update the order of ids in state, therefore updating the order they are rendered which handles the different sizes issue. This also means everything is in the right order on drop, so animations complete as they normally would.

All in all, this change should not really be noticeable to the end user

### How to verify

1. Go to a dashboard and re-arrange some filter parameters, Save.
2. Create a new SQL question with some parameters. These are also re-arrangeable.

### Demo
chrome_5A0b7ixv8L

### Checklist

### DIFF
--- a/frontend/src/metabase/core/components/Sortable/Sortable.styled.tsx
+++ b/frontend/src/metabase/core/components/Sortable/Sortable.styled.tsx
@@ -3,7 +3,9 @@ import styled from "@emotion/styled";
 export const SortableDiv = styled.div<{
   transform: string | undefined;
   transition: string | undefined;
+  isDragging: boolean;
 }>`
   transform: ${props => props.transform};
   transition: ${props => props.transition};
+  ${({ isDragging }) => (isDragging ? "z-index: 10;" : "")}
 `;
--- a/frontend/src/metabase/core/components/Sortable/SortableList.tsx
+++ b/frontend/src/metabase/core/components/Sortable/SortableList.tsx
@@ -1,0 +1,114 @@
+import type {
+  DragOverEvent,
+  DragStartEvent,
+  Modifier,
+  SensorDescriptor,
+} from "@dnd-kit/core";
+import { DndContext, DragOverlay } from "@dnd-kit/core";
+import { SortableContext, arrayMove } from "@dnd-kit/sortable";
+import { useState, useMemo, useEffect } from "react";
+import _ from "underscore";
+
+import { isNotNull } from "metabase/lib/types";
+
+type ItemId = number | string;
+
+interface RenderItemProps<T> {
+  item: T;
+  id: ItemId;
+  isDragOverlay?: boolean;
+}
+interface useSortableListProps<T> {
+  items: T[];
+  getId: (item: T) => ItemId;
+  renderItem: ({ item, id, isDragOverlay }: RenderItemProps<T>) => JSX.Element;
+  onSortStart?: (event: DragStartEvent) => void;
+  onSortEnd?: ({ id, newIndex }: { id: ItemId; newIndex: number }) => void;
+  sensors?: SensorDescriptor<any>[];
+  modifiers?: Modifier[];
+}
+
+export const SortableList = <T,>({
+  items,
+  getId,
+  renderItem,
+  onSortStart,
+  onSortEnd,
+  sensors = [],
+  modifiers = [],
+}: useSortableListProps<T>) => {
+  const [itemIds, setItemIds] = useState<ItemId[]>([]);
+  const [indexedItems, setIndexedItems] = useState<Record<ItemId, T>>({});
+  const [activeItem, setActiveItem] = useState<T | null>(null);
+
+  useEffect(() => {
+    setItemIds(items.map(getId));
+    setIndexedItems(_.indexBy(items, getId));
+  }, [items, getId]);
+
+  const sortableElements = useMemo(
+    () =>
+      itemIds
+        .map(id => {
+          const item = indexedItems[id];
+          if (item) {
+            return renderItem({ item, id });
+          }
+        })
+        .filter(isNotNull),
+    [itemIds, renderItem, indexedItems],
+  );
+
+  const handleDragOver = ({ active, over }: DragOverEvent) => {
+    if (over && active.id !== over.id) {
+      setItemIds(ids => {
+        const oldIndex = ids.indexOf(active.id);
+        const newIndex = ids.indexOf(over.id);
+        return arrayMove(ids, oldIndex, newIndex);
+      });
+    }
+  };
+
+  const handleDragStart = (event: DragStartEvent) => {
+    document.body.classList.add("grabbing");
+
+    onSortStart?.(event);
+
+    const item = items.find(item => getId(item) === event.active.id);
+    if (item) {
+      setActiveItem(item);
+    }
+  };
+
+  const handleDragEnd = () => {
+    document.body.classList.remove("grabbing");
+    if (activeItem && onSortEnd) {
+      onSortEnd({
+        id: getId(activeItem),
+        newIndex: itemIds.findIndex(id => id === getId(activeItem)),
+      });
+      setActiveItem(null);
+    }
+  };
+
+  return (
+    <DndContext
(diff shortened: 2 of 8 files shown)


frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,95 @@
...
+  items,
+  getId,
+  renderItem,
+  onSortStart,
+  onSortEnd,
+  disableSort = false,
+  sensors = [],
+  modifiers = [],
+}: useSortableListProps<T>) => {
+  const [itemIds, setItemIds] = useState<any[]>([]);
```
kulyk, line 33: `string[]` ?

frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [activeItem, setActiveItem] = useState<T | null>(null);
+
+  useEffect(() => {
+    setItemIds(items.map(getId));
+  }, [items, getId]);
+
+  const sortableElements = useMemo(
+    () =>
+      itemIds.map(id => {
+        const item = items.find(item => getId(item) === id);
```
kulyk, line 43: Should we have `_.indexBy(item, getId)` to avoid these lookups?

frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,95 @@
...
+        if (item) {
+          return (
+            <Sortable id={id} key={`sortable-${id}`} disabled={disableSort}>
+              {renderItem(item)}
+            </Sortable>
+          );
+        }
+      }),
+    [itemIds, items, renderItem, disableSort, getId],
+  );
```
kulyk, line 53: Should we do something like `itemIds.filter(isNotNull)` to ensure there won't be any surprises if we couldn't find an element for a given ID?

frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { DndContext, DragOverlay } from "@dnd-kit/core";
+import { SortableContext, arrayMove } from "@dnd-kit/sortable";
+import { Sortable } from "./Sortable";
+
+interface useSortableListProps<T> {
+  items: T[];
+  getId: (item: T) => string;
+  renderItem: (item: T) => JSX.Element;
+  onSortStart: (event: DragStartEvent) => void;
+  onSortEnd: ({ id, newIndex }: { id: string; newIndex: number }) => void;
```
kulyk, line 17: Should these be optional?

frontend/src/metabase/parameters/components/ParametersList.jsx
```diff
@@ -49,69 +29,73 @@ function ParametersList({
...
 
-  const handleSortEnd = ({ oldIndex, newIndex }) => {
-    document.body.classList.remove("grabbing");
-    if (setParameterIndex) {
-      setParameterIndex(parameters[oldIndex].id, newIndex);
-    }
-  };
+  const handleSortStart = useCallback(() => {
+    document.body.classList.add("grabbing");
+  }, []);
```
kulyk, line 42: Would it make sense to move this to `Sortable` / `SortableList` ?

kulyk (commented):
Looks great! Left a few questions

frontend/src/metabase/core/components/Sortable/Sortable.styled.tsx
```diff
@@ -3,7 +3,9 @@ import styled from "@emotion/styled";
 export const SortableDiv = styled.div<{
   transform: string | undefined;
   transition: string | undefined;
+  isDragging: boolean;
 }>`
   transform: ${props => props.transform};
   transition: ${props => props.transition};
+  opacity: ${({ isDragging }) => (isDragging ? 0.5 : undefined)};
```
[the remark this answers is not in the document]
npfitz: Good point. I can change that

package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@dnd-kit/core": "^6.0.8",
     "@dnd-kit/modifiers": "^6.0.1",
-    "@dnd-kit/sortable": "^7.0.2",
+    "@dnd-kit/sortable": "^8.0.0",
```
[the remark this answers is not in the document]
npfitz: I don't think we're using anything specific. Just seemed like a good idea to keep things up to date

frontend/src/metabase/core/components/Sortable/SortableList.tsx
npfitz: Yeah, I've updated the ID type to be `string | number`. I should do the same here

frontend/src/metabase/core/components/Sortable/SortableList.tsx
npfitz: Probably a good idea!

frontend/src/metabase/core/components/Sortable/SortableList.tsx
npfitz: If it's more efficient then we can

frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,126 @@
...
+  DragStartEvent,
+  Modifier,
+  SensorDescriptor,
+} from "@dnd-kit/core";
+import { DndContext, DragOverlay } from "@dnd-kit/core";
+import { SortableContext, arrayMove } from "@dnd-kit/sortable";
+import { isNotNull } from "metabase/lib/types";
+import { Sortable, type SortableProps } from "./Sortable";
+
+type itemId = number | string;
```
kulyk, line 15: Types are usually TitleCase

frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,126 @@
...
+        return arrayMove(ids, oldIndex, newIndex);
+      });
+    }
+  };
+
+  const handleDragStart = (event: DragStartEvent) => {
+    document.body.classList.add("grabbing");
+    if (onSortStart) {
+      onSortStart(event);
+    }
```
kulyk, line 95: nit: `onSortStart?.()`

frontend/src/metabase/core/components/Sortable/SortableList.tsx
```diff
@@ -0,0 +1,126 @@
...
+  items,
+  getId,
+  renderItem,
+  onSortStart,
+  onSortEnd,
+  disableSort = false,
+  sensors = [],
+  modifiers = [],
+  sortableWrapper: SortableWrapper = Sortable,
+  wrapperProps,
```
kulyk, line 40: This feels a little bit off, what if we had to add `Sortable` as a part of `renderItem`? 

```tsx
import { SortableList } from "metabase/core/components/SortableList"

<SortableList
  items={items}
  renderItem={item => (
    <SortableList.Item key={item.id} id={item.id} draggingStyle={{ opacity: 0.5 }}>
      <MoreJSX />
    </SortableList.Item>
  )}
/>
```

// `SortableList.Item` is `Sortable` here

kulyk (approved):

PR #38062, opened 2024-01-23
Malli defn improvements: print the incorrect value in the error

escherize:
This PR makes `mu/fn` and `mu/defn` return the invalid value it got, like we do with invalid values coming through defendpoint.

# Old message
Screen Shot 2024-01-23 at 2 34 06 PM

# New message
Screen Shot 2024-01-23 at 2 33 47 PM


src/metabase/util/malli/fn.clj
```diff
@@ -150,7 +150,9 @@
 (defn- validate [error-context schema value error-type]
   (when *enforce*
     (when-let [error (mr/explain schema value)]
-      (let [humanized (me/humanize error)
+      (let [humanized (me/humanize error {:wrap (core/fn humanize-include-value
+                                                  [{:keys [value message]}]
+                                                  (str message ", got: " (pr-str value)))})
```
escherize, line 155: decided to go with `got` so that it's clear where errors are coming from. There's a similar thing for `defendpoint` that says received.

qnkhuat (approved):

camsaul (approved):
works for me

PR #27660, opened 2023-01-12
Migrate field refs in visualization_settings.column_settings from legacy format

calherries:
This PR is a prerequisite to make #27487 easier to solve, and a trial of SQL migrations for JSON.

It includes adding UDFs to the H2 database on creating the Liquibase connection.

The SQL migrations ended up getting pretty big, our biggest yet. Most of the machinery is in getting data in and out of the JSON structure, which amounts to a `(update visualization_settings :column_settings (fn [x] (map (...) x))` operation.

### What's in the migration?

The migration updates the field refs in visualization_settings.column_settings from legacy formats (e.g. `["field-id",<id>]` to `["field",<id>,null]`). I recommend reading [the test](https://github.com/metabase/metabase/pull/27660/files#diff-41093075df90623837726c733afc60b57899a66d6e9878b4250413b74f045f67R824). 

Example transformations:
```
(in EDN for clarity, but the transformations are still done with JSON strings in the DB)

[:field-id 12] 
-> [:field 12 null]

[:field-literal "name" :type/Text]
-> [:field "name" {:base-type :type/Text}]

[:fk-> [:field-id 12] [:field-id 11]] 
-> [:field 11 {:source-field 12}]

[:fk-> 12 11]                                   
-> [:field 11 {:source-field 12}]
```

Note normalizing other legacy field tags like `datetime-field` or `binning-strategy`, is not needed here. In the past these field tags were removed to get the unwrapped field ref (aka base dimension field ref) before storing them in the database.

### Why do this now?
#27487 requires another migration (probably even bigger) following this to the field refs in visualization_settings.column_settings, see https://github.com/metabase/metabase/pull/27487.
It is simpler to do that migration if the field refs are standardized in one format first.

### Is this the best approach?

Maybe not. I'm putting this out there so we can debate on whether it's worth the hassle to do these sorts of migrations in SQL, or find another approach. Alternatives include:
1. Clojure-land migrations with liquibase, detailed by Ngoc [here](https://www.notion.so/metabase/clojure-land-migration-2-0-5e19d4de3ecc4a569145f913f0ad0d36)
2. We could avoid storing this data nested in JSON blobs. These field refs could be stored as a top-level field of a table. In that world the migration wouldn't need any complicated machinery other than a case statement. The migrations wouldn't have to implement `update-in...map` transformations in SQL. So we could aim to extract column_settings out from the visualization_settings JSON blob and into a table like `column_settings` first. 

TODO: write tests for the UDFs. Looking for validation on the approach though.

[...]

PR #38074, opened 2024-01-24
Add `report_card.type`

### LINKED ISSUE #37367: BE - Migrate `dataset: true` to enum value

qnkhuat:
Closes https://github.com/metabase/metabase/issues/37367 by adding `report_card.type` column, it's a text type with 3 possible values: question, model, metric( for metric v2 ).

Previously, we only had 2 types of questions: normal questions and model, and we used a boolean `dataset` column to differentiate them. Now we want a new question type : "metric" so a boolean column is no longer sufficient.

This PR changes it so that we can use `report_card.type` and `report_card.dataset` interchangeably at the moment, the rationale behind not dropping "dataset" column now is that doing so is going to be a huge effort since `report_card` is used anywhere. This is the list of [API](https://github.com/metabase/metabase/issues/37367#issuecomment-1905426067) that returns `report_card.dataset`, so to drop `dataset` completely, we'll need to update both FE and BE for a lot of places.

To save us from that hassle so we can focus on building metricv2, we'll make the transition from using `report_card.dataset` to `report_card.type` gradually. Since most of the app doesn't need to know the difference between `question, dataset` and `metric`, they can continue use `dataset` to differentiate between `question` and `model`. For example: persistent, actions ... etc.

For where the different matters, we'll migrate those part of the code to use the new `type` column. E.g: adding "metricv2" to search API.

By doing this we will have the minimum number of changes to unblock metricv2 project.


crisptrutski (commented):
Ran out of time to finish reviewing this PR, but left the comments I have so far.

Overall - nice plan.

src/metabase/api/card.clj
```diff
@@ -768,7 +784,7 @@
     param-key :- ms/NonBlankString
     query     :- [:maybe ms/NonBlankString]]
    (let [param (get (m/index-by :id (or (seq (:parameters card))
-                                        ;; some older cards or cards in e2e just use the template tags on native queries
+                                        ;; some olderkcards or cards in e2e just use the template tags on native queries
```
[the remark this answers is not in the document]
qnkhuat: acfaaf350b

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
...
+(mu/defn ^:private is-type? :- :boolean
+  "Check if card is a model?"
+  [target-type :- CardTypes
+   {:keys [type] :as _card} :- [:map [:type CardTypes]]]
+  (= target-type type))
+
+(def ^{:arglists '([card])
+       :doc      "Returns true if `card` is a model."}
+  is-model?
+  (partial is-type? "model"))
```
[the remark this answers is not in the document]
qnkhuat: yea makes sense, did this in acfaaf350b

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
   ([_ pk]
    (mi/can-read? (t2/select-one :model/Card :id pk))))
 
+(def ^{:private true
+       :doc     "All acceptable card types.
+                Previously (< 49), we only have 2 types of card: question and model, this was differentiated using
+                the boolean `dataset`. Soon we'll have more card types (e.g: metric) and we can't no longer use a boolean
```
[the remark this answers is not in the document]
qnkhuat: thanks, typed this with a hungry stomach :p.

updated in acfaaf350b

src/metabase/models/card.clj
```diff
@@ -306,27 +337,55 @@
                            :query-database        query-db-id
                            :field-filter-database field-db-id})))))))
 
-(defn- assert-valid-model
+(defn- ensure-type-and-dataset-are-aligned
```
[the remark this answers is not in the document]
qnkhuat: `assert-*` is usually for functions that throw an error if the test fail.

this function makes change to the card so I don't think `assert` will give people the right idea,

I've changed the function name to `ensure-type-and-dataset-are-consistency` tho. acfaaf350b

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
   ([_ pk]
    (mi/can-read? (t2/select-one :model/Card :id pk))))
 
+(def ^{:private true
+       :doc     "All acceptable card types.
+                Previously (< 49), we only have 2 types of card: question and model, this was differentiated using
+                the boolean `dataset`. Soon we'll have more card types (e.g: metric) and we can't no longer use a boolean
+                column to differentiate between all types. So we've add a new column `type` for this purpose.
+
+                Migrating all the code to use `report_card.dataset` is quite an effort, so we decided we'll migrate it gradually.
```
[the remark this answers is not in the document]
qnkhuat: oh I meant "migrate to use `report_card.type`",  changed in acfaaf350b

src/metabase/api/card.clj
```diff
@@ -383,12 +383,23 @@
...
 
+(defn- check-card-type-and-dataset
+  [{:keys [dataset type]}]
+  ;; if dataset is true then type must be model
+  ;; if dataset is false, then type can't be model
+  (when (and (some? type) (some? dataset))
+    (api/check (if (true? dataset)
+                 (= "model" type)
+                 (not= "model" type))
+      [400 (deferred-tru "Dataset and type doesn''t match")])))
```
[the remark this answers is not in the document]
qnkhuat: yes acfaaf350b

src/metabase/db/custom_migrations.clj
```diff
@@ -1033,3 +1033,29 @@
 (define-reversible-migration UnifyTimeColumnsType
   (unify-time-column-type! :up)
   (unify-time-column-type! :down))
+
+(define-reversible-migration CardRevisionAddType
```
qnkhuat, line 1037: we need to do this to make sure all the card revisions have the same schema as the active card.

src/metabase/models/card.clj
```diff
@@ -306,27 +337,55 @@
...
+
+   ;; if only type presents, make sure dataset follows
+   (some? type)
+   (assoc card :dataset (= type "model"))
+
+   ;; if only dataset presents, make sure type follows
+   (some? dataset)
+   (assoc card :type (if dataset
+                       "model"
+                       "question"))))
```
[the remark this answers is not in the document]
qnkhuat: I agree this is not ideal, I have this because a lot of our tests depend on creating a card with only `:dataset true`.

I think it's safe to have this because of 2 reasons:
- this relies on the idea that where the difference between "question, model," and "metric" matters, we should update the code to use "type."
- this function is only used during card creation and update, both code paths will be updated to use `type` instead of `dataset` (as noted in the PR's description). In production, this code path should never be reached, we will only see this in tests.

And I added a warning in [d87b778](https://github.com/metabase/metabase/pull/38074/commits/d87b7788674ac3c3d11d12af96b7972cb18cbd06)

src/metabase/api/card.clj
```diff
@@ -458,23 +472,26 @@
    result_metadata        [:maybe qr/ResultsMetadata]
    cache_ttl              [:maybe ms/PositiveInt]
    collection_preview     [:maybe :boolean]}
-  (let [card-before-update (t2/hydrate (api/write-check Card id)
-                                       [:moderation_reviews :moderator_details])]
+  (let [card-before-update      (t2/hydrate (api/write-check Card id)
+                                            [:moderation_reviews :moderator_details])
+        maybe-switched-to-model (when (and (some? dataset) (some? type))
```
crisptrutski, line 477: Is it not better to keep backwards and forwards compatibility and support the presence of only one param?

```clojure
(if (some? type) (= :type model) dataset)
```

src/metabase/db/custom_migrations.clj
```diff
@@ -1033,3 +1033,28 @@
...
+  (let [migrate! (fn [revision]
+                   (let [object    (json/parse-string (:object revision) keyword)
+                         new-object (-> object
```
crisptrutski, line 1040: nit: why bother threading through a single form? seems to add noise.

src/metabase/db/custom_migrations.clj
```diff
@@ -1033,3 +1033,28 @@
...
+                     (t2/query {:update :revision
+                                :set    {:object (json/generate-string new-object)}
+                                :where  [:= :id (:id revision)]})))]
+    (run! migrate! (t2/reducible-query {:select [:*]
+                                        :from   [:revision]
+                                        :where  [:= :model "Card"]})))
+  (let [rollback! (fn [revision]
+                    (let [object    (json/parse-string (:object revision) keyword)
+                          new-object (-> object
+                                         (assoc :dataset (= (:type object) "model"))
```
crisptrutski, line 1053: Devils advocate: doesn't this branch keeps the `dataset` field maintained? Why do we need to repopulate in this rollback? I think we could save this for rolling back a migration that deleted the column

src/metabase/db/custom_migrations.clj
```diff
@@ -1033,3 +1033,28 @@
...
+  (let [migrate! (fn [revision]
+                   (let [object    (json/parse-string (:object revision) keyword)
+                         new-object (-> object
+                                        (assoc :type (if (:dataset object)
+                                                       "model"
+                                                       "question")))]
+                     (t2/query {:update :revision
+                                :set    {:object (json/generate-string new-object)}
+                                :where  [:= :id (:id revision)]})))]
+    (run! migrate! (t2/reducible-query {:select [:*]
```
crisptrutski, line 1047: Ouch these reducible migrations seem very expensive. Is it perhaps worth special casing the migration for databases that support JSON operations natively?

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
...
+(mu/defn ^:private is-type? :- :boolean
+  "Check if card is a model?"
+  [target-type :- CardTypes
+   {:keys [type] :as _card} :- [:map [:type CardTypes]]]
+  (= target-type type))
+
+(defn  is-question?
+  "Returns true if `card` is a question"
+  [card]
+  (is-type? "model" card))
```
crisptrutski, line 128: Typo: s/model/question

src/metabase/models/card.clj
```diff
@@ -148,15 +179,17 @@
...
+   :dataset])
 
 (defmethod revision/serialize-instance :model/Card
   ([instance]
    (revision/serialize-instance Card nil instance))
   ([_model _id instance]
    (cond-> (apply dissoc instance excluded-columns-for-card-revision)
      ;; datasets should preserve edits to metadata
-     (not (:dataset instance))
+     (and (:type instance) (not (is-model? instance)))
```
crisptrutski, line 192: Why not keep backwards compatibility here, like other places in the code? `(or (is-model? instance) (:dataset instance))`. This version feels strange - checking that :type is present (i.e. not assuming the migration) but then not handling :dataset seems unbalanced.

src/metabase/models/card.clj
```diff
@@ -306,27 +339,57 @@
...
-  [{:keys [dataset dataset_query]}]
-  (when dataset
-    (let [template-tag-types (->> (vals (get-in dataset_query [:native :template-tags]))
-                                  (map (comp keyword :type)))]
-      (when (some (complement #{:card :snippet}) template-tag-types)
-        (throw (ex-info (tru "A model made from a native SQL question cannot have a variable or field filter.")
-                        {:status-code 400}))))))
+  [{:keys [type dataset_query]}]
+  (case type
+    "model" (let [template-tag-types (->> (vals (get-in dataset_query [:native :template-tags]))
```
crisptrutski, line 375: nit: if we're going to use the threading macro here, i think it's worth threading through `vals` as well to keep each clause a bit smaller.

crisptrutski (commented):
Mostly stylistic feedback, but one showstopping typo. Looks good!

resources/migrations/001_update_migrations.yaml
```diff
@@ -5205,6 +5205,40 @@ databaseChangeLog:
...
+
+  - changeSet:
+      id: v49.2024-01-22T11:51:00
+      author: qnkhuat
+      comment: Backfill `report_card.type`
+      changes:
+        - sql:
+            sql: >-
+              UPDATE report_card
+              SET type = 'model'
```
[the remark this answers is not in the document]
qnkhuat: `question` is the default value so we don't need to backfill it

src/metabase/api/card.clj
```diff
@@ -458,23 +472,26 @@
...
+                                  (or (= type :model) dataset))]
     ;; Do various permissions checks
     (doseq [f [collection/check-allowed-to-change-collection
                check-allowed-to-modify-query
                check-allowed-to-change-embedding]]
       (f card-before-update card-updates))
+    (check-card-type-and-dataset card-updates)
```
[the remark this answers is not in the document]
qnkhuat: moved to the top. 
I put it there initially to make it closes with all other checks.

but considering this check will be removed in the future so I fail fast is probably the better option.

src/metabase/models/card.clj
qnkhuat: dataset is no longer serialized for revision, you can see in line 184 I added `:dataset` to the list of `excluded-columns-for-card-revision`

resources/migrations/001_update_migrations.yaml
```diff
@@ -5205,6 +5205,41 @@ databaseChangeLog:
         - dbms:
             type: mysql,mariadb
 
+  - changeSet:
+<<<<<<< HEAD
```
piranha, line 5209: Looks like merge leftover.

src/metabase/api/card.clj
```diff
@@ -470,11 +486,11 @@
                                                              :query             dataset_query
                                                              :metadata          result_metadata
                                                              :original-metadata (:result_metadata card-before-update)
-                                                             :dataset?          (if (some? dataset)
-                                                                                  dataset
-                                                                                  (:dataset card-before-update))})
+                                                             :dataset?          (if (some? maybe-turn-to-datset)
+                                                                                  maybe-turn-to-datset
+                                                                                  (card/is-model? card-before-update))})
```
calherries, line 491: This change contains a bug. If we are changing `type` from `"model"` to `"question"`, and `dataset=nil`, then `maybe-turn-to-dataset=nil`. That means the value of `dataset?` here will be `true` but it should be `false`.

src/metabase/db/custom_migrations.clj
calherries: This is a fair point. I think this will be the most expensive migration to date. On a large instance there can be 100,000s of these updates needed, and from memory this kind of update is prohibitively slow.

All our app DBs support JSON operations to implement this with SQL. I've tried updating JSON objects with SQL before ([PR](https://github.com/metabase/metabase/pull/27660/files)), the downside is it gets ugly very quickly for complex queries and it's a lot easier to make mistakes (but this is migration is much simpler than that PR).

I think you have a couple of options here:
1. migrate the cards in a `revision/revert-to-revision!` implementation for cards
2. make this migration fast enough using SQL in a way.

I lean towards 2 because any bugs with the migration are likely to be caught immediately, as opposed to 1.

resources/migrations/001_update_migrations.yaml
```diff
@@ -5205,6 +5205,40 @@ databaseChangeLog:
...
+      id: v49.2024-01-22T11:50:00
+      author: qnkhuat
+      comment: Add `report_card.type`
+      changes:
+        - addColumn:
+            tableName: report_card
+            columns:
+              - column:
+                  remarks: The type of card, could be 'question', 'model', 'metric'
+                  name: type
```
qwef, line 5218: I think you should add:
```
constraints:
  nullable: false
```

qnkhuat: by default the column will not be nullable. but yea let's add it for clarity

src/metabase/db/custom_migrations.clj
qnkhuat: This is not the 1st time we've done it, and it seems we're ok with it; I've seen no failure reported for this kind of migration.

And also h2 does not support this out of the box, we'll have to inject our own UDF implementation like you did in #27660

qnkhuat: optimized for postgres and mysql in 70521d704c

calherries: Awesome. I think H2 is fine to leave with the slow clojure version.

> This is not the 1st time we've done it

I meant this is the first time we will be migrating revisions for all cards. We have only migrated dashboard revisions from clojure before. Previous clojure migrations on card revisions filter out the majority of cards in the select statement.

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
...
+  [target-type :- CardTypes
+   {:keys [type] :as _card} :- [:map [:type CardTypes]]]
+  (= target-type type))
+
+(defn is-question?
+  "Returns true if `card` is a question"
+  [card]
+  (is-type? "question" card))
+
+(defn is-model?
```
calherries, line 130:
```suggestion
(defn model?
```

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
...
+  "Malli schema for acceptable card types."
+  (into [:enum] card-types))
+
+(mu/defn ^:private is-type? :- :boolean
+  "Check if card is a model?"
+  [target-type :- CardTypes
+   {:keys [type] :as _card} :- [:map [:type CardTypes]]]
+  (= target-type type))
+
+(defn is-question?
```
calherries, line 125: functions that return boolean values conventionally don't have a `is-` prefix. see `metabase.lib.types.isa`
```suggestion
(defn question?
```

src/metabase/models/card.clj
```diff
@@ -101,6 +102,36 @@
...
+                Migrating all the code to use `report_card.type` will be quite an effort, we decided that we'll migrate it gradually.
+                In the mean time we'll have both `type` and `dataset` columns."} card-types
+  #{"model" "question" #_:metric})
+
+(def CardTypes
+  "Malli schema for acceptable card types."
+  (into [:enum] card-types))
+
+(mu/defn ^:private is-type? :- :boolean
+  "Check if card is a model?"
```
calherries, line 120:
```suggestion
  "Returns true if card is of type `target-type`"
```

src/metabase/models/card.clj
```diff
@@ -306,27 +339,58 @@
...
+  - If only one key is present, we'll assoc the correct value for the other key."
+  [{:keys [type dataset] :as card}]
+  (cond
+   ;; if none of the 2 keys presents, do nothing
+   (and (nil? type) (nil? dataset))
+   card
+
+   ;; if both type and dataset presents, we prioritize type
+   (and (some? type) (some? dataset))
+   (assoc card :dataset (= type "model"))
```
calherries, line 357: Can you explain why type is prioritized here?

src/metabase/models/card.clj
```diff
@@ -306,27 +339,58 @@
...
+  - If both keys are present, we make sure `dataset` is true if `type` is `model` else false.
+    This will make a different when we have `metric` type since a boolean can't represent tri-state
+  - If only one key is present, we'll assoc the correct value for the other key."
+  [{:keys [type dataset] :as card}]
+  (cond
+   ;; if none of the 2 keys presents, do nothing
+   (and (nil? type) (nil? dataset))
+   card
+
+   ;; if both type and dataset presents, we prioritize type
```
calherries, line 355:
```suggestion
   ;; if both type and dataset is present, we prioritize type
```

src/metabase/models/card.clj
```diff
@@ -601,7 +665,13 @@ saved later when it is ready."
...
          card-data            (assoc (zipmap data-keys (map card-data data-keys))
                                      :creator_id (:id creator)
-                                     :dataset (boolean (:dataset card-data))
+                                     :dataset    (or (and (:type card-data) (is-model? card-data))
+                                                     (boolean (:dataset card-data)))
+                                     :type       (or (:type card-data)
+                                                     ;; most tests will create a model using :dataset true, so we'll
+                                                     ;; respect that
+                                                     (when (:dataset card-data) "model")
+                                                     "question")
```
calherries, line 674: Why don't we reuse `ensure-type-and-dataset-are-consistent` here?

src/metabase/models/revision/diff.clj
```diff
@@ -76,11 +76,9 @@
...
     (deferred-tru "modified the query")
 
-    [:dataset false true]
-    (deferred-tru "turned this into a model")
-
-    [:dataset true false]
-    (deferred-tru "changed this from a model to a saved question")
+    ;; report_card.type
+    [:type old new]
+    (deferred-tru "type changed from {0} to {1}" old new)
```
calherries, line 81: You can keep the old behaviour here by matching on the values e.g. `[:type "model" "question"]`

test/metabase/api/card_test.clj
```diff
@@ -977,6 +978,55 @@
                          [:trace          [:sequential :any]]]
                         (create-card! :rasta 403)))))))))
 
+(deftest create-card-with-type-and-dataset-test
+  (mt/with-model-cleanup [:model/Card]
+    (testing "type and dataset must matches"
```
calherries, line 983:
```suggestion
    (testing "type and dataset must match"
```

src/metabase/models/card.clj
qnkhuat: because we can only infer dataset from `type` but not the other way around.

src/metabase/models/card.clj
calherries: Why is this not `(not (is-model? instance))`?

src/metabase/models/revision/diff.clj
qnkhuat: good call https://github.com/metabase/metabase/pull/38074/commits/9c2d57c5947e77e1494fe654c95c0448a898f4ee

src/metabase/models/card.clj
qnkhuat: makes sense, changed.

src/metabase/api/card.clj
```diff
@@ -458,23 +472,26 @@
...
     (let [result-metadata-chan  (card/result-metadata-async {:original-query    (:dataset_query card-before-update)
                                                              :query             dataset_query
                                                              :metadata          result_metadata
                                                              :original-metadata (:result_metadata card-before-update)
-                                                             :dataset?          (if (some? dataset)
-                                                                                  dataset
-                                                                                  (:dataset card-before-update))})
+                                                             :dataset?          (if (some? maybe-switched-to-model)
+                                                                                  maybe-switched-to-model
+                                                                                  (card/is-model? card-before-update))})
```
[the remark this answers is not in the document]
qnkhuat: I've updated following this logic. Thanks!

src/metabase/models/card.clj
```diff
@@ -306,27 +341,58 @@
...
+(defn- ensure-type-and-dataset-are-consistent
+  "We're in the process of migrating from using `report_card.dataset` to `report_card.type`.
+  In the future we'll drop `dataset` and only use `type`. But for now we need to make sure that both keys are aligned
+  when dealing with cards.
+  - If both keys are present, we make sure `dataset` is true if `type` is `model` else false.
+    This will make a different when we have `metric` type since a boolean can't represent tri-state
+  - If only one key is present, we'll assoc the correct value for the other key."
+  [{:keys [type dataset] :as card}]
+  (cond
+   ;; if none of the 2 keys presents, do nothing
```
calherries, line 353: this is still incorrect, same for comments below
```suggestion
   ;; if none of the 2 keys is present, do nothing
```

src/metabase/models/card.clj
qnkhuat: do you mean to set these values or as a check?

`ensure-type-and-dataset-are-consistent` will eventually be called in the `before-insert` hook so it's a bit redundant.

src/metabase/models/card.clj
calherries: Got it. I was concerned they could disagree but I think you're assuming here that they already match because it passed the `check-card-type-and-dataset` check.

src/metabase/models/card.clj
calherries: To set the values. From what I can tell this code duplicates logic from there.

It's not redundant because it looks like there's currently a bug on line 679 (I can't highlight it in github) where `dataset` is used but it should be the `dataset` that comes from `ensure-type-and-dataset-are-consistent`.

```
         result-metadata-chan (result-metadata-async {:query    dataset_query
                                                      :metadata result_metadata
                                                      :dataset? dataset})
```

Also where are the tests for this? I only see API and revision tests, no tests on the create-card! function itself.

src/metabase/models/card.clj
```diff
@@ -306,27 +341,58 @@
...
-  [{:keys [dataset dataset_query]}]
-  (when dataset
-    (let [template-tag-types (->> (vals (get-in dataset_query [:native :template-tags]))
-                                  (map (comp keyword :type)))]
-      (when (some (complement #{:card :snippet}) template-tag-types)
-        (throw (ex-info (tru "A model made from a native SQL question cannot have a variable or field filter.")
-                        {:status-code 400}))))))
+  [{:keys [type dataset_query]}]
+  (case type
+    "model" (let [template-tag-types (->> (get-in dataset_query [:native :template-tags])
```
calherries, line 377: not sure why you changed this to use `case`. I think the original was more idiomatic
```suggestion
  (when (model? card) 
```

src/metabase/models/card.clj
```diff
@@ -306,27 +341,58 @@
...
+                (throw (ex-info (tru "A model made from a native SQL question cannot have a variable or field filter.")
+                                {:status-code 400}))))
+    nil))
 
 ;; TODO -- consider whether we should validate the Card query when you save/update it??
 (defn- pre-insert [card]
   (let [defaults {:parameters         []
                   :parameter_mappings []}
-        card     (merge defaults card)]
+        card     (ensure-type-and-dataset-are-consistent (merge defaults card))]
```
calherries, line 389: This should be in the `t2/define-before-insert` instead. Otherwise there could be other functions called in `t2/define-before-insert` that depend on `:dataset` or `:type` that do not receive the valid card returned by `ensure-type-and-dataset-are-consistent`.

Also should you put this in `t2/define-before-update` too? There are at least some tests that call `t2/update!` that update `:dataset` directly using toucan .

src/metabase/api/card.clj
```diff
@@ -458,8 +472,13 @@
...
    collection_preview     [:maybe :boolean]}
-  (let [card-before-update (t2/hydrate (api/write-check Card id)
-                                       [:moderation_reviews :moderator_details])]
+  (check-card-type-and-dataset card-updates)
+  (let [card-before-update     (t2/hydrate (api/write-check Card id)
+                                           [:moderation_reviews :moderator_details])
+        is-model-after-update? (cond
+                                (some? type)    (= "model" type)
+                                (some? dataset) dataset
+                                :else           (card/model? card-before-update))]
```
calherries, line 481: This also seems to duplicate `ensure-type-and-dataset-are-consistent` logic

calherries:
@qnkhuat what did you mean by this section of the description? 

> For FE: There are 3 breaking changes in:
how we create question
how we create a model
how we turn model <-> question
Previously, we did this by toggling dataset; we need to change this to use the type column instead.
Other than that, the rest of the app should work without changes.

Are these todos for a FE engineer to follow up on? As far as I can tell this PRs changes do not break any API endpoints in the technical sense, we only extend them with an extra key, `:type`.

src/metabase/models/card.clj
calherries: Actually, I'm concerned this can paper over bugs in ways that are difficult to understand. I can imagine some old function (call it `f`) that takes a card with `:dataset` and `:type`, then updates `:dataset` but not `:type`. When the card gets to `ensure-type-and-dataset-are-consistent` we will undo the change that `f` made.

To defend against this possibility I think we need to assert that `:dataset` and `:type` do not contradict each other here with a runtime exception.

src/metabase/models/card.clj
qnkhuat: ah, I changed it to case so we can add a check for metric later

src/metabase/models/card.clj
qnkhuat: I'm not following.

`ensure-type-and-dataset-are-consistent` is both triggered in `define-before-insert` and `define-before-update`, it's just indirectly.

calherries: Sorry you're right, it is in `define-before-update`. I was looking at the wrong branch in my editor.

My first point is that I'd prefer `t2/define-before-insert` to include the call to `ensure-type-and-dataset-are-consistent` directly at at the beginning of the thread macro, exactly as you have done in `t2/define-before-update`. The benefit is to future-proof against accidentally depending on `:dataset` or `:type` being valid in the `define-before-insert` body before `ensure-type-and-dataset-are-consistent` has been called. For example, if `populate-result-metadata` was changed to use on `:dataset` for whatever reason, it could introduce a bug where `:dataset` is out-of-sync with `:type`.

src/metabase/models/card.clj
```diff
@@ -474,7 +540,7 @@
   ;;
   ;; We have to convert this to a plain map rather than a Toucan 2 instance at this point to work around upstream bug
   ;; https://github.com/camsaul/toucan2/issues/145 .
-  (-> (into {:id (:id card)} (t2/changes card))
+  (-> (into {:id (:id card)} (ensure-type-and-dataset-are-consistent (t2/changes card)))
```
calherries, line 543: how about this? 
```suggestion
  (-> (into {:id (:id card)} (t2/changes card))
       ensure-type-and-dataset-are-consistent
```

src/metabase/api/card.clj
qnkhuat: not sure how this is a duplication, this is meant to whether the card is a model after the update.

src/metabase/models/card.clj
qnkhuat: right, I updated how we set the values here: 3f5fd80bb0


On testing, I don't think we have anything to test regarding the `dataset?` flag there since it's used to make sure we preserve the original metadata when updating the card. Since this is during card creation, there is no "original metadata"

src/metabase/models/card.clj
qnkhuat: I don't want to change this function to throw an exception; its primary job is to fill out `type` and `dataset` if one of them is missing.

instead I've added `assert-card-type-and-dataset` and guard it on both `before-insert` and `before-update`, I think that'll help us avoid the cases where `:dataset` and `:type` is inconsistent.

src/metabase/models/card.clj
qnkhuat: makes sense, moved this call to `before-insert`

qnkhuat:
> Are these todos for a FE engineer to follow up on? As far as I can tell this PRs changes do not break any API endpoints in the technical sense, we only extend them with a redundant key, :type.

yea they're TODOs for FE. Originally, I planned to make those breaking changes, but I realized there are too many tests that depend on using `dataset` to create a model, so I decided to keep that behavior. removed that "breaking changes" from the description.

src/metabase/models/card.clj
```diff
@@ -594,29 +676,37 @@ saved later when it is ready."
...
-         card-data            (assoc (zipmap data-keys (map card-data data-keys))
-                                     :creator_id (:id creator)
-                                     :dataset (boolean (:dataset card-data))
-                                     :parameters (or parameters [])
-                                     :parameter_mappings (or parameter_mappings []))
+         card-data            (-> (zipmap data-keys (map card-data data-keys))
+                                  (assoc
+                                   :creator_id (:id creator)
+                                   :dataset    (boolean dataset)
+                                   :type       (or type
```
qnkhuat, line 689: I had 2 choices here:
1. we set value of `type` based on `dataset` here if `type` does not exist
2. update all tests that currently use `:dataset true` to create a model

I went with 1 since I want to avoid fixing hundred of tests, and also I think it's safe cause if you want to create new card type(metric), you'll have to include `type` anyway

src/metabase/models/card.clj
```diff
@@ -594,29 +676,37 @@ saved later when it is ready."
...
+                                                   ;; most tests will create a model using :dataset true, so we'll
+                                                   ;; respect that
+                                                   (when dataset "model")
+                                                   "question")
```
calherries, line 693: This is redundant now that you've added the call to `ensure-type-and-dataset-are-consistent`. You can just add `:type` and `:dataset` to data-keys.
```suggestion
```
That's why I suggested to use `ensure-type-and-dataset-are-consistent` to avoid duplicating logic

src/metabase/models/card.clj
calherries: Ah cool thanks for explaining.

Continuing the conversation [here](https://github.com/metabase/metabase/pull/38074/files#r1475028148) because the referenced code is out of date

src/metabase/api/card.clj
calherries: This code contains the logic: dataset=true implies type=model. That logic is duplicated in `ensure-type-and-dataset-are-consistent`. Here's an example of how to dedupe it
```suggestion
        is-model-after-update? (if (and (nil? type) (nil? dataset))
                                 (card/model? card-before-update)
                                 (card/model? (card/ensure-type-and-dataset-are-consistent card-updates)))]
```

src/metabase/models/card.clj
qnkhuat: this is needed to be able to create a model with `:dataset true`, see https://github.com/metabase/metabase/pull/38074#discussion_r1473770401

calherries: `ensure-type-and-dataset-are-consistent` sets `type=model` if `dataset=true`. The logic here is redundant and duplicated. I'm talking about these lines: https://github.com/metabase/metabase/blob/c17398925852b29f2eadd19888d55e3c70c3dfa5/src/metabase/models/card.clj#L378-L385

calherries:
I want to be sure we're clear on what you mean by this phrase in the PR description: "we can use report_card.type and report_card.dataset interchangeably". It seems that based on `ensure-type-and-dataset-are-consistent`, `dataset` and `type` are not strictly interchangeable, because type takes precedence when both are present. For example:

| dataset | type | resulting dataset + type |
| --------| -----| ---------------|
| true | model | true model |
| false | model | true model (type precedence) |
| nil | model | true model |
| true | question | false question (type precedence) |
| false | question | false question |
| nil | question | false question |

setting `dataset` to `false` doesn't imply `type=question` when `type` is present
but setting `type` to `question` always implies `dataset=false`

similarly 

setting `dataset` to `true` doesn't imply `type=model` when `type` is present
but setting `type` to `model` always implies `dataset=true`

so `dataset` and `type` are not perfectly interchangeable.

The implication is if we have code on the FE that changes `dataset`'s value while `type` is present, and then makes a request to PUT, we're going to have a bug where dataset's value is actually not changed.

Am I missing something here?

qnkhuat:
Maybe I wasn't clear; my intention was that: as far as the rest of the codebase(think model persistent, search, executing query...) is concerned, they can use `dataset` and `type` interchangeably. This PR is meant to make sure any code path that can change those columns will make sure they're consistent.

As for metric v2 projects, they will have to start migrating to use `type`, while the rest of the codebase can use `dataset`.

Does that make sense?

calherries:
I think I understand your intention, but I don't think this PR achieves being able to use dataset and type interchangeably.

A code path that changes only `dataset` and not `type` while `type` is present in the changes map passes the changes map to either the PUT request, `update-card!`, or `t2/update` will change its behaviour with this PR.

Consider this demonstration:

```clojure
(let [; assume we have some card created already, with dataset=true
      card (t2/insert-returning-instances! :model/Card {; ... card attributes
                                                        :dataset true})]
  ; the following code will behave differently now.
  ; before the PR, this code sets dataset to false.
  ; after the PR, this code sets dataset to true (because the type is model)
  (let [card-updates (assoc card :dataset false)]
    (t2/update! :model/Card (:id card) card-updates)))
```

> As for metric v2 projects, they will have to start migrating to use type, while the rest of the codebase can use dataset.

So I don't think you can have it both ways as long as it's possible for both keys to be present in a changes map, because when both are present we won't know which key to prefer when they contradict by the time the changes get to `ensure-type-and-dataset-are-consistent`. To preserve the correct behaviour I think you need to make it impossible for them to contradict.

qnkhuat:
well we've updated the code path that changes card type to use `type` on FE so I think that's unlikely to happen. But I agree that we shouldn't leave room for foot gunning ourselves.

An easier way to do that is to forbid changing type using `dataset` by adding an assert in `before-update`. wdyt?

calherries:
Oh, I just found the FE PR [here](https://github.com/metabase/metabase/pull/38127). Please reference this in the PR description next time. 

> An easier way to do that is to forbid changing type using dataset by adding an assert in before-update. wdyt?

Is the intention to make `dataset` read-only except when `type` changes? I'm curious how you imagine that assert would be implemented. I think it's equivalent to my previous suggestion to add a runtime assertion to `ensure-type-and-dataset-are-consistent` to make sure `dataset` doesn't contradict `type`.

> well we've updated the code path that changes card type to use type on FE so I think that's unlikely to happen.

Where's that change? I looked at the FE code briefly but it's not immediately obvious

calherries:
Ah, I think I found the FE code that almost guarantees that `dataset` and `type` won't contradict. 
https://github.com/metabase/metabase/pull/38127/files#diff-fdff338c9e6c27877274c27c438055055f0f9395a0a37c3fd360e194f6d2e461R276-R285

The reason why I say "almost guarantees", is there can be other ways to update a card, e.g. with the `setCardAndRun` action. 
https://github.com/metabase/metabase/pull/38127/files#diff-334f3a72d48f91a43e7daa89781867db281345092d3394bd1307157ad51e285fR167

But anyway, if we're sure they can't contradict each other on the FE that's cool, but the BE should validate the assumption with a runtime check in the toucan hooks. And then you should be free to remove the code that handles contradictions in `ensure-type-and-dataset-are-consistent` because it's a dead code path.

qnkhuat:
> but the BE should validate the assumption with a runtime check in the toucan hooks
isn't that what we do by calling `assert-card-type-and-dataset` in the hooks? It's both called in `before-update` and `before-insert`

qnkhuat:
@calherries I've updated so that we throw an exception if `dataset` and `type` contradicts in `ensure-type-and-dataset-are-consistent`. The `assert-card-type-and-dataset` method is no longer needed so I've removed it.

src/metabase/models/card.clj
```diff
@@ -594,29 +670,32 @@ saved later when it is ready."
...
-                               :parameters :parameter_mappings :collection_id :collection_position :cache_ttl]
-         card-data            (assoc (zipmap data-keys (map card-data data-keys))
-                                     :creator_id (:id creator)
-                                     :dataset (boolean (:dataset card-data))
-                                     :parameters (or parameters [])
-                                     :parameter_mappings (or parameter_mappings []))
+                               :parameters :parameter_mappings :collection_id :collection_position :cache_ttl :type]
+         card-data            (-> (zipmap data-keys (map card-data data-keys))
+                                  (assoc
+                                   :dataset    (or (when (= "model"type)
```
calherries, line 680: This is a typo and a bug, you meant to write:
```suggestion
                                   :dataset    (or (when (= "model" type)
```
(aside: this should have been caught by tests)

If you make this change, it's still too easy to use `create-card!` incorrectly, in the same way as I pointed out with the API endpoint before. Here's a passing test that demonstrates what I consider risky behaviour:

```
(is (=?  {:type "model" :dataset true}
         (metabase.models.card/create-card! (merge (card-with-name-and-query (mt/random-name))
                                                   {:type "model" :dataset false})
                                            {:id (mt/user->id :crowberto)})))
```

It looks like `dataset` is being set to `false`, but under the covers it's being overridden to `true`. This should throw an error like the API endpoint. Since we need to calculate the value of `dataset?` after the update here, you can reuse `card/ensure-type-and-dataset-are-consistent` to do it so that the logic isn't duplicated.

calherries (changes requested):
The endpoint looks better now, just one bug to fix with `create-card!`

src/metabase/models/card.clj
```diff
@@ -594,29 +674,34 @@ saved later when it is ready."
...
-                                     :creator_id (:id creator)
-                                     :dataset (boolean (:dataset card-data))
-                                     :parameters (or parameters [])
-                                     :parameter_mappings (or parameter_mappings []))
+                               :parameters :parameter_mappings :collection_id :collection_position :cache_ttl :type]
+         ;; `zipmap` instead of `select-keys` because we want to get `nil` values for keys that aren't present. Required by
+         ;; `api/maybe-reconcile-collection-position!`
+         card-data            (-> (zipmap data-keys (map card-data data-keys))
+                                  (assoc
+                                   :dataset    (or (when (= "model" type)
```
qnkhuat, line 685: this `(or ...` logic is needed because we want to be able to create a model by using `type=model` or `dataset=true`

calherries: This code is redundant and duplicates logic with `ensure-type-and-dataset-are-consistent`, similar to [this](https://github.com/metabase/metabase/pull/38074#discussion_r1475073288) comment.

Since you're threading this map into `ensure-type-and-dataset-are-consistent` you can just remove it with no change in behaviour.

src/metabase/models/card.clj
calherries: Bumping this. Including the `(:type instance)` is unnecessary imo. It implies it's possible to have `instance` without a `:type` key, and implies that case should be handled differently than if `(:type instance)` was present but `(not (is-model? instance))` is false. But it shouldn't be possible to have `instance` without a `:type` key here because `revision/serialize-instance` should be called with `instance` being a selected toucan object from the database after insertion/update.

src/metabase/models/card.clj
qnkhuat: without the default value, i'll need to add `:dataset` to the list of `data-keys`, but in that case the `card-data` won't have both `dataset` and `type` key, which will eventually throw an error because there is no default value for those keys.

calherries: Ah, you want to preserve the default if `type` or `dataset` aren't provided in the args? In that case you can explicitly handle that with a conditional that doesn't repeat the logic inside `ensure-type-and-dataset-are-consistent`. 

e.g.
```clojure
(-> (zipmap data-keys (map card-data data-keys))
    (cond-> (and (nil? type) (nil? dataset))
      (assoc :type "question"))
    (assoc ....)
    card/ensure-type-and-dataset-are-consistent)
```

qnkhuat: okay I like your version better, updated in 598d469ba2

calherries (approved):
Looks good, apologies for delaying this with my paranoia, but I'm happy we've simplified this a lot more so the most important logic is in one place now. I'll try be more specific and timely with my comments next time because these improvements definitely took longer than I expected.

I have one tiny suggestion [here](https://github.com/metabase/metabase/pull/38074#discussion_r1478280750). Resolve that and it's good to go

src/metabase/models/card.clj
qnkhuat: ah yea, the `(:type instance)` is only needed in revision tests because there are tests where the object will only contain a key that is needed to generate a revision description.

for example: https://github.com/metabase/metabase/blob/metric-v2/add-report-card-type-shortcut/test/metabase/models/revision_test.clj#L341

I'll added a comment to the code.

qnkhuat:
@calherries Thank you for the great review!

PR #38075, opened 2024-01-24
Upgrade `jsrsasign` to `v11.0.0`

nemanjaglumac:
Resolves https://github.com/metabase/metabase/security/dependabot/130


WiNloSt (approved):
This package is only used for embed preview. It's still working fine after the upgrade.

PR #38085, opened 2024-01-24
Repro #36122 - Table Metadata having 2 foreign keys from the same table pointing to the same key cause implicit joins to only show 1

### LINKED ISSUE #36122: Table Metadata having 2 foreign keys from the same table pointing to the same key cause implicit joins to only show 1
### Describe the bug

When you have a table in which you got two columns setup as foreign key types and pointing to the same key cause the GUI editor to only show 1 of them as an implicit join and ignore completely the other

### To Reproduce

1. Go to Admin -> Table Metadata -> Sample DB -> Reviews -> Set the Rating as Foreign Key to Product ID (so it matched the Product ID one)

image

2. Go to New Question -> Reviews -> Click on Group By -> Notice only the Product is Visible and there is no Rating option

image

### Expected behavior

Both implicit joins are visible and selectable 

### Logs

None that are relevant

```JSON
47.8 and `master`
```

### Severity

It's not a blocking issue since a user can still manually perform the join from the join option of the GUI but it's annoying

### Additional context

_No response_

kamilmielnik:
Closes #36122

### DIFF
--- a/e2e/test/scenarios/admin/datamodel/metadata.cy.spec.js
+++ b/e2e/test/scenarios/admin/datamodel/metadata.cy.spec.js
@@ -146,8 +146,13 @@ describe("scenarios > admin > datamodel > metadata", () => {
       .click();
 
     popover().within(() => {
-      cy.findAllByText("Rating").should("have.length", 2);
+      cy.findAllByTestId("dimension-list-item")
+        .eq(3)
+        .should("have.text", "Rating");
       cy.get(".List-section-header").should("have.length", 3);
+      cy.get(".List-section-header").eq(0).should("have.text", "Review");
+      cy.get(".List-section-header").eq(1).should("have.text", "Product");
+      cy.get(".List-section-header").eq(2).should("have.text", "Rating");
     });
   });
 
--- a/e2e/test/scenarios/admin/datamodel/metadata.cy.spec.js
+++ b/e2e/test/scenarios/admin/datamodel/metadata.cy.spec.js
@@ -1,4 +1,5 @@
 import {
+  getNotebookStep,
   restore,
   openOrdersTable,
   openReviewsTable,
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/admin/datamodel/metadata.cy.spec.js
```diff
@@ -132,6 +133,24 @@ describe("scenarios > admin > datamodel > metadata", () => {
...
+
+    openReviewsTable({ mode: "notebook" });
+    summarize({ mode: "notebook" });
+    getNotebookStep("summarize")
+      .findByText("Pick a column to group by")
+      .click();
+
+    popover().within(() => {
+      cy.findAllByText("Rating").should("have.length", 2);
+      cy.get(".List-section-header").should("have.length", 3);
```
uladzimirdev, line 150: why so and not `findByText` with explicit values?

kamilmielnik: Added more assertions: 16a82e26534a56ce456327e62a9f18b3c0b6130a

uladzimirdev: I meant a more general like `cy.findByText("Review").should('exist')`, but this should work as well

uladzimirdev (approved):

uladzimirdev:
can we run a stress test for it?

e2e/test/scenarios/admin/datamodel/metadata.cy.spec.js
kamilmielnik: We're testing for "Rating" which should appear both as section header and as an item within one of the sections

kamilmielnik:
> can we run a stress test for it?

All green: https://github.com/metabase/metabase/actions/runs/7653758434

PR #38087, opened 2024-01-24
Bump snowflake driver to test

dragonsahead:
Bumping the driver to test if this has anything to do with the regression in schemas we're seeing


tsmacdonald (approved):

qnkhuat:
why do you need to get this merge to test?

paoliniluis:
Canceling as the fixes that Cam introduced "should" fix the issue

PR #38088, opened 2024-01-24
Allow :base to be a map of inherited options for defsetting

### LINKED ISSUE #38250: Improve Application Setting inheritance
# Description

The `defsetting` mechanism allows us to define various kinds of settings for the application, notably 

An important type of setting is the `uuid-nonce`, a class of random identifiers used to identify customer installations with respect to billing, enabling feature flags, etc. This kind of setting needs to generate a UUID the first time the application loads, save this value, and never allow it to be changed.

The original implementation for this type involved overriding the low level `get-value-of-type` multimethod to do the initialization. This hidden side-effect violated the principle of surprise, especially in a immutable-first language, and led to a serious bug when we introduced setting validation during start-up. It also meant that auditing mechanism was being side-stepped, something that was accidentally "fixed" when we introduced a first-class initializer property, which led to another show-stopper bug due to a concurrency bug when auditing enterprise settings.

An issue with using the new initializer property is that there's no mechanism to set this via the existing type mechanism. We needed to revert all the instance to `{:type :string}` and add the the new `:init` setting explicitly at each site. Another failing with the previous `::uuid-nonce` type was that it did not encapsulate being read-only, not being audited, and having an underlying string representation, meaning that we needed to be careful to repeat these properties for every usage of the type.

TLDR - the `::uid-nonce` "type" is now a collection of 4 properties that need to be repeated together across their 6 usages. This repetition brings the usual risk of missing a spot to update, and its inherent noise.

It would be nice to have a mechanism that allows the `uuid-nonce` type to be fully encapsulated. There are probably also existing patterns that could be cleaned up with this mechanism. Particularly if we decide to open source this utility, I believe we should have good support for richer types we will not have foreseen.

## Proposals

1. Add more multimethods
2. Introduce a type map registry
3. Overload the `:type` property
4. Introduce `:base` property
5. Allow a map argument to `defsetting`
6. Do nothing.

### Multimethod

We could introduce a new multimethod

- Unable to extend base types
  - Only namespaced keywords can be used with `derive` in the global hierarchy
  - Would need to either namespace the base types or introduce a custom hierarchy.
- Large code surface area
  - Definition of each type is spread across a number of methods
  - Needs a new multimethod for each behavior to cover
- Inconsistency
  - Top level definitions just use keys and values for all properties
  - Harder to extract types, and more chance for errors.
  - A `get-base-setting-properties` multimethod is inconsistent with the other methods
- Does not feel idiomatic ("just data")

### Type map registry

This was my initial idea. Instead of registering multimethods, we could introduce a map of type keys to maps of their encapsulated properties. 

For now just putting this map in a `def` would be fine, but we could introduce an atom and a `register!` method if we want it to be extensible. A downside of having mutable state would be its impact on the stability of the REPL environment when changing branches, and more difficulty in finding the code where things have been defined.

### Type property

Instead of providing a key to a property map, what if we just linked to a var holding the map? The idea here was to allow the `:type` property to be a symbol for a var defining the map. Based on feedback this was extended to allow map literals and even code that defines a map, although in practice I think both of these would be anti-patterns.

By  allows us to enforce, disallowing the inheritance of`:export?` (it should always be explicit), making read-only "final", etc. We can also support transitive references to other maps to get transitive inheritance. 

**This is my favored approach.**

Implementation: https://github.com/metabase/metabase/pull/37863/files

### Base property

Instead of overloading the `:type` property, we could instead introduce a new key. This has a few advantages:

- We do not overload the type of `:type`
- We can inherit properties that are orthogonal to `:type`, e.g. just things like visibility and auditing.
- We could allow `:type` to be inherited

In my opinion it is confusing to have "parallel" hierarchies for `:type` and `:base`, and I don't think the additional affordances have real world value - in fact I think they would be foot guns.

Implementation: https://github.com/metabase/metabase/pull/37863/files

### Map argument to `defsetting`

Clojure keyword argument syntax allows for a final map argument. We could just allow a map to be passed directly.

This would be the most powerful and flexible approach allowing a la carte composition, e.g. `merge` for mixins, `dissoc` for "mixouts", `update` for decorators.

The downsides of this approach in my opinion:

- If we support mixing explicit pairs with a map, the map will shadow the explicit keys. This could be suprising 
  - Supporting mixing also creates issues for linting and producing good compile time errors, as it would not yet be a map at compile time for the macro, so we cannot make use of the `& {:as}` destructuring.
  - In short, we should probably only support a map if it's the only parameter.
- There's a large syntax inconsistency between flat keyword definitions and a map manipulation expression.
  - All explicit keys move to a final assoc operation at the bottom of the definition.
  - All explicit keys are now also far more indented than
  - Top level properties like `visibility` and `export` are now further down
- By not being opinionated, we introduce danger and sacrifice ergonomics.

Technically speaking we could expose an "extend setting properties" method to use, which takes the explicit keys as the first argument, enforces limitations on what can be overridden. In principle I have no problem with supporting "map only" as an escape hatch even if we go with one of the other options as the typical idiom.

WIP implementation: https://github.com/metabase/metabase/pull/38090

### Do nothing

Perhaps this is all YAGNI

---

### Note on existing multimethods

If we introduce one of these map-based approaches, we could even replace the multimethods with keys like `parse` and `stringify`. In my opinion that would be a nice ergonomic and consistency improvement. The top level type keywords could instead be replaced by maps. This could even get rid of the type's type-overloading if we go with Option 3.

IMHO it's awkward that `:type` is used to define the default values of `getter` and `setter`, but both of the latter can be overridden as regular options.

----

### Issue links

- https://github.com/metabase/metabase/issues/37801
- https://github.com/metabase/metabase/issues/38009

crisptrutski:
Closes: https://github.com/metabase/metabase/issues/38250

### Description

This is an alternate to [overloading type as a base type](https://github.com/metabase/metabase/pull/37863).

I'm not sure which one I prefer. Using a different key means we don't have to inherit the type, and avoids overloading a field, but on the other hand I feel that conflating "type" with "inheritance" is more natural and familiar.

Another idea is to riff off [how kwarg functions can accept maps](https://clojure.org/news/2021/03/18/apis-serving-people-and-programs). That might take a bit more macro foo, so I tried this first.


src/metabase/models/setting.clj
```diff
@@ -618,9 +618,16 @@
       (throw (Exception.
               (tru "Invalid value for string: must be either \"true\" or \"false\" (case-insensitive)."))))))
 
-(defn ^{:doc "The string representation of a type 4 random uuid"} random-uuid-str []
+(defn- random-uuid-str []
   (str (random-uuid)))
 
+;; This var-type allows the bundling of a number of attributes together. In some sense it is defining a subtype / mixin.
+(def ^{:doc "A random uuid value that should never change again"} uuid-nonce-type
```
calherries, line 625:
```suggestion
(def ^{:doc "A random uuid value that should never change again"} uuid-nonce-base
```

src/metabase/models/setting.clj
```diff
@@ -618,9 +618,16 @@
       (throw (Exception.
               (tru "Invalid value for string: must be either \"true\" or \"false\" (case-insensitive)."))))))
 
-(defn ^{:doc "The string representation of a type 4 random uuid"} random-uuid-str []
+(defn- random-uuid-str []
   (str (random-uuid)))
 
+;; This var-type allows the bundling of a number of attributes together. In some sense it is defining a subtype / mixin.
```
calherries, line 624:
```suggestion
;; This base allows the bundling of a number of attributes together. In some sense it is defining a subtype / mixin.
```

src/metabase/models/setting.clj
```diff
@@ -1079,6 +1086,15 @@
...
        (not (ns-in-test? (:namespace setting-definition)))))
 
+(defn- inherit-base
+  "Recursively uses the values in `:base` as default values for `setting-options`"
+  [{:keys [base] :as setting-options}]
+  (let [without-base (dissoc setting-options :base)]
+    (if-not base
+      without-base
+      (do (assert (map? base))
+          (merge (inherit-base base) without-base)))))
```
calherries, line 1096: I don't think this should recurse. It's simpler without it and we have no need for it.
```suggestion
  (let [without-base (dissoc setting-options :base)]
    (merge base without-base))
```
Also I feel the assertion is unnecessary. `merge` with other types should fail

calherries (changes requested):
This is still my preferred approach. Remember to add `:base` to the `defsetting` docstring and write a test with a real `defsetting` to check that the option is merged in that considers the property might have a default, e.g. try setting two different values of `:cache?` with `:base` and check the override works

src/metabase/models/setting.clj
crisptrutski: > I don't think this should recurse. 

Transitivity is something I would expect from an inheritance mechanism. Since we've gone the route of keeping type as separate concern, and users can still merge their own maps before passing to the macro, I'll go ahead and remove this. Using merge does mean losing safeguards like making read-only "final" etc, but perhaps this is YAGNI.

crisptrutski:
@calherries I think I've addressed all your feedback. Don't forget to review https://github.com/metabase/metabase/pull/38128, as it's a dependency of this branch.

src/metabase/models/setting.clj
```diff
@@ -618,9 +618,16 @@
       (throw (Exception.
               (tru "Invalid value for string: must be either \"true\" or \"false\" (case-insensitive)."))))))
 
-(defn ^{:doc "The string representation of a type 4 random uuid"} random-uuid-str []
+(defn- random-uuid-str []
   (str (random-uuid)))
 
+;; This base allows the bundling of a number of attributes together. In some sense it is defining a subtype / mixin.
+(def ^{:doc "A random uuid value that should never change again"} uuid-nonce-base
```
qnkhuat, line 625: nit, you can put the docstring after the var name, no need for using meta tag

test/metabase/models/setting_test.clj
```diff
@@ -199,6 +199,30 @@
               #"Cannot initialize setting before the db is set up"
               (setting/get :test-setting-custom-init)))))))
 
+(def ^:private base-options
+  {:setter   :none
+   :type     :csv
+   :default "base-default"})
```
qnkhuat, line 205:
```suggestion
   :default  "base-default"})
```

test/metabase/models/setting_test.clj
```diff
@@ -199,6 +199,30 @@
...
+
+(defsetting test-base-setting
+  "Setting to test the `:feature` property of settings. This only shows up in dev."
+    :visibility :internal
+    :default "my,default"
+    :base base-options)
```
qnkhuat, line 211:
```suggestion
  :visibility :internal
  :default "my,default"
  :base base-options)
```

test/metabase/models/setting_test.clj
```diff
@@ -199,6 +199,30 @@
...
+
+(deftest ^:parallel defsetting-with-base-test
+  (testing "A setting which specifies some base options"
```
qnkhuat, line 214:
```suggestion
  (testing "A setting which specifies a base option"
```

test/metabase/models/setting_test.clj
```diff
@@ -199,6 +199,30 @@
...
+    (let [setting (setting/resolve-setting :test-base-setting)]
+      (testing "Uses non-overridden base options"
+        (is (= :csv (:type setting)))
+        (is (= :none (:setter setting)))
+        ;; Obfuscate test-base-setting! to avoid a false positive from `deftest-check-parallel`
```
qnkhuat, line 219: a better way to do this is to tell kondo to ignore it.

```
#_{:clj-kondo/ignore [linter-name]}
```

src/metabase/models/setting.clj
```diff
@@ -1079,6 +1086,14 @@
        (not= (:setter setting-definition) :none)
        (not (ns-in-test? (:namespace setting-definition)))))
 
+(defn- inherit-base
```
qnkhuat, line 1089:
```suggestion
(defn- maybe-inherit-base
```

src/metabase/models/setting.clj
```diff
@@ -1079,6 +1086,14 @@
...
+  "Incorporate the values in `:base` as default values for `setting-options`"
```
qnkhuat, line 1090:
```suggestion
  "Inherit the values in `:base` as default values for `setting-options` if exists"
```

I think we should be clearer about `base` is an optional thing

src/metabase/models/setting.clj
```diff
@@ -1208,7 +1229,7 @@
         setting-setter-fn-symbol (-> (symbol (str (name setting-symbol) \!))
                                      (with-meta (meta setting-symbol)))
         setting-definition-symbol (gensym "setting-")]
-    `(let [setting-options#          (merge ~options ~setting-metadata)
+    `(let [setting-options#          (merge (#'inherit-base ~options) ~setting-metadata)
```
qnkhuat, line 1232:
```suggestion
    `(let [setting-options#          (merge (inherit-base ~options) ~setting-metadata)
```

IMO, I'd rather make `inherit-base` public than using the var-quote like this.

qnkhuat (approved):
Cool, same with Cal this is the approach I like :).
I have a couple of nits tho.

src/metabase/models/setting.clj
crisptrutski: I'm slightly torn, but will make the change. I don't feel that this belongs in the public API, but I also don't like exposing a "var" call in generated code, both aesthetically amd as it has performance overhead, although its mute for a form that only executes during compilation.

src/metabase/models/setting.clj
crisptrutski: Ah that's nice! In older versions of Clojure this did not work.

test/metabase/models/setting_test.clj
crisptrutski: I prefer the original wording here, it is very intentional. The way I use the terms:

- Each key value pair is one option
- The base map is therefore a collection of options, or "base options"
- We are specifying "some" base options, as it is optional.

test/metabase/models/setting_test.clj
crisptrutski: I think this only works when the linter reports a finding, but in this case the hook throws an exception. I'll give it a try though.

src/metabase/models/setting.clj
crisptrutski: When I see a function starting with "maybe" I assume this means it returns something optional, but in this case it always returns a map. That said, the "inherit" verb makes it clear that this is a transformation, so I guess it's clear that it means it is "maybe" a no-op in this case. I don't think the extra verbiage adds an important distinction though, in the same way that `merge` handles nil but isn't called `merge-maybe`. Let me know if you feel strongly.

crisptrutski:
> Cool, same with Cal this is the approach I like :). I have a couple of nits tho.

I still haven't quite accepted it - a uuid is a type, not just a bag of options! 

Maybe next time I'll be more cautious about presenting options that aren't my favorite  - hoping future setting reopens this chestnut 😜

test/metabase/models/setting_test.clj
qnkhuat: what's the key of that error? I'll see if I could fix it, looks like we need to update the hook to respect kondo/ignore

test/metabase/models/setting_test.clj
qnkhuat: makes sense to me

test/metabase/models/setting_test.clj
crisptrutski: Ah I took another look and see that `error!` is just an alias for `reg-finding!`, which encapsulates the key `:metabase/validate-deftest`.

I'd thought Cursive would have shown me the key if there was one, and that `error!` wasn't sending one. Cool, I've pushed a version with the ignore. Unfortunately it seems that I have to ignore the entire test, which isn't ideal. I might actually prefer the symbol-builder approach if we need to extend this test in future.

crisptrutski: Score one for Cider! I see that shows the ignore key front and center 😄

src/metabase/models/setting.clj
calherries: I'm with Chris on this one, especially since this is now called `merge-base` not `inherit-base`. I find it intuitive that if a base doesn't exist this is a no-op, like `(merge x nil)`.

src/metabase/models/setting.clj
calherries: How about making `merge-base` a local variable?

calherries:
> a uuid is a type, not just a bag of options!

Ha, maybe I'll come to agree with you after a few more years of changes to `defsetting`. I find it hard to see the point yet

test/metabase/models/setting_test.clj
```diff
@@ -199,6 +199,31 @@
               #"Cannot initialize setting before the db is set up"
               (setting/get :test-setting-custom-init)))))))
 
+(def ^:private base-options
+  {:setter   :none
+   :type     :csv
+   :default  "base-default"})
+
+(defsetting test-base-setting
+  "Setting to test the `:feature` property of settings. This only shows up in dev."
```
calherries, line 208:
```suggestion
  "Setting to test the `:base` property of settings. This only shows up in dev."
```

test/metabase/models/setting_test.clj
```diff
@@ -199,6 +199,31 @@
...
+;; Avoid a false positive from `deftest-check-parallel` due to referencing the test-base-setting! setter.
+;; Even though we only resolve the (non-existent) setter, and don't call anything, the linter flags it.
+#_{:clj-kondo/ignore [:metabase/validate-deftest]}
+(deftest ^:parallel defsetting-with-base-test
+  (testing "A setting which specifies some base options"
+    (let [setting (setting/resolve-setting :test-base-setting)]
+      (testing "Uses non-overridden base options"
+        (is (= :csv (:type setting)))
+        (is (= :none (:setter setting)))
+        (is (nil? (resolve 'test-base-setting!)))
```
calherries, line 222: I think this test would be clearer and less brittle if we tested a single attribute with two settings, one with a top-level override and one without. Otherwise we're coupling the test to the behaviour of the setting properties more than we need to. e.g.

```
(def ^:private base-options
  {:default "base-default"})

(defsetting test-no-default-with-base-setting
  "Setting to test the `:base` property of settings. This only shows up in dev."
  :base base-options)

(defsetting test-default-with-base-setting
  "Setting to test the `:base` property of settings. This only shows up in dev."
  :visibility :internal
  :base base-options
  :default "top-level-default")

(deftest ^:parallel defsetting-with-base-test
  (testing "A setting which specifies some base options"
    (let [no-top-level-default-setting (setting/resolve-setting :test-no-default-with-base-setting)
          top-level-default-setting              (setting/resolve-setting :test-default-with-base-setting)]
      (testing "Uses base options when no top-level options are specified"
        (is (= "base-default" (:default no-top-level-default-setting)))
        (is (= "base-default" (no-top-level-default-setting))))
      (testing "Uses top-level options when specified"
        (is (= "top-level-default" (:default top-level-default-setting)))
        (is (= "top-level-default" (no-top-level-default-setting)))))))
```

calherries: Also we don't need to use the messy kondo ignore if we can avoid using the setter

src/metabase/models/setting.clj
crisptrutski: You know [how I feel about long functions](https://dev.to/kanani_nirav/the-five-lines-of-code-principle-why-less-is-more-in-programming-31j6) - you see I'm a recovering old school Ruby programmer. Having more code emitted in macro expansion also makes me uncomfortable about code bloat, but this is probably trauma from C++ and Rust where it can have a big impact on compile time and binary sizes. 

You may still have won me over that it's better than emitting code with `#'` though 😄 

I've made peace with just having it public though. Since merge is no longer recursive, exposing it gives users a way to do their own transitive extensions.

test/metabase/models/setting_test.clj
crisptrutski: I tried to make the test as "real world" as possible, by closely resembling the uuid-nonce case, but without depending on the actual prod definition. I do really like your suggestion of reusing the same base between those settings, I agree it adds clarity.

I very intentionally do want to test the setter though, as the code needed to be reworked to avoid emitting a setter that just throws exceptions. This was in the parent PR but I still want some test coverage.

crisptrutski: I broke out the setter to a separate test, as it doesn't need to test both settings, and it's nice to reduce the scope of the kondo ignore.

calherries (approved):

PR #38090, opened 2024-01-24
Allow trailing map expression in defsetting, like a kwarg function

crisptrutski:
Refs: https://github.com/metabase/metabase/issues/38250

### Description

My third and final variation :-)

This version allows you to use [regular clojure syntax](https://clojure.org/news/2021/03/18/apis-serving-people-and-programs).

~~I stopped short of allowing arbitrary Clojure forms like `(merge base-opts mixin-1 mixin-2)`, I feel that it's OK to require an intermediate `(def base-with-mixin-1-and-2 ...)` var rather than calling `eval` inside the macro.~~ These forms are now supported.

One problem I have is the Cursive is getting false positives for linting the settings that use this. I'll survive, it struggles with a lot of the codebase. Perhaps I have some misconfiguration.


test/metabase/models/setting_test.clj
```diff
@@ -199,30 +199,27 @@
...
+   :type     :csv
+   :default "base-default"})
+
+(defsetting test-base-setting
+  "Setting to test the `:feature` property of settings. This only shows up in dev."
+  :visibility :internal
+  :default    "my,default"
+  base-options)
+
+(deftest defsetting-with-base-test
```
camsaul, line 213: can this be marked `^:parallel`?

test/metabase/models/setting_test.clj
```diff
@@ -199,30 +199,27 @@
...
-      (let [expected {:type   :json
-                      :setter :none
-                      :init   str
-                      :bloop? false
-                      :blep?  'obviously}]
-        (testing "You can use explicit maps for the mixin behaviour"
-          (is (= expected (expand with-map))))
-        (testing "You can use var references for the mixing behaviour"
-          (is (= expected (expand with-var))))))))
+(def base-options
```
camsaul, line 202:
```suggestion
(def ^:private base-options
```

It's always nice to make stuff that's only used in this specific test namespace `^:private`, so we can change it with worrying about it affecting stuff in other test namespaces, and so Kondo can warn us about it being unused if we change the test that was using it so it doesn't become dead code

src/metabase/models/setting.clj
```diff
@@ -1206,13 +1208,13 @@
   (default: `:no-value` for most settings; `:never` for user- and database-local settings, settings with no setter,
   and `:sensitive` settings.)"
   {:style/indent 1}
-  [setting-symbol description & {:as options}]
+  [setting-symbol description & kwargs]
```
camsaul, line 1211: I'm wondering if it wouldn't be better to just make this an either-or thing, either you use kwargs like 

```clj
(defsetting my-setting
  "This is my setting"
  :x 1 :y 2)
  ```
  
  **or** you use a map of args like
  
 ```clj
(defsetting my-setting
  "This is my setting"
  {:x 1 :y 2})
  ```
  
Then you just have two different arities of this function and you don't need all the extra stuff like `expand-kwargs`

Comment (camsaul) [reply src/metabase/models/setting.clj:1211]: But if you do want to keep it this way you might want to consider adding a clojure spec `fdef` for the macro, when Clojure does macroexpansion it will check the args against the spec and throw an error if they don't match

File: src/metabase/models/setting.clj [hunk 4]
```diff
@@ -1081,20 +1081,22 @@
...
-      without-base
-      (merge (inject-base (deref-map base)) without-base))))
+    (map? expr)    expr
+    (symbol? expr) (u/prog1 @(resolve expr)
+                     (when-not (map? <>)
+                       (throw (ex-info "Invalid :base in setting configuration. Must reference a map."
+                                       {:symbol expr :value <>})))
+                     (assert (map? <>)))
+    :else          (throw (ex-info "Invalid :base in setting configuration. Must be a literal map or a reference."
+                                   {:expression expr}))))
```
camsaul, line 1093: So, you can't do something like 

```clj
(defsetting my-setting
  "My Setting"
  (merge map-1 map-2 {:a 100})
```

That seems user-unfriendly, I don't think you would expect it to work like that. I think we should rework the macro so it's not checking if its args are symbols and calling resolve on them... have the macro just serve as syntactic sugar and do the `def` and then have it do the important stuff in a function call in the macroexpansion

qnkhuat:
I'm not sure I like this since it changes the syntax of `defsetting`, I think it's clearer if we only stick with key value pair.

And also, what happens if we have a setting like this?

```clojure
(defsetting str-setting
 :visibility :internal)

(defsetting new-setting
 :visibility :public
 str-setting)
```

Logically I tink the value  of `:visibility` for `new-setting` should be `:public`, but from reading the code it's not apparent because the map has to be at the tail of the macro.

crisptrutski:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/38090?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #38128** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #38090** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38090?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38090?utm_source=stack-comment).

test/metabase/models/setting_test.clj
crisptrutski: I don't see any reason why not. Taking this as a hint add this annotation where possible for tests that I create or edit - sure! 😄

test/metabase/models/setting_test.clj
crisptrutski: Those are great reasons, thanks for pointing them out. I guess I've taken those assurances for granted in the languages I've been working in before, where tests are written in private modules or classes.

test/metabase/models/setting_test.clj
crisptrutski: Hmm, I got a false positive from `deftest-check-parallel` due to referencing the setter function, even though it was quoted. 

For now I've worked around it by building the symbol programmatically, but perhaps there's a better way? 

I'm not sure if we can just trust any symbolic references to unsafe functions, but perhaps we could support safety annotations. Either some metadata that this function reads, or using `hooks/reg-finding!` instead of `error!` inside the linter.

src/metabase/models/setting.clj
crisptrutski: The ergonomics I wanted relied on mixing the two, I think the readability is greatly decreased if you need to add an `assoc` onto the map, and I was not planning on supporting arbitrary forms so that wouldn't have worked.

I realized today that I had a misconception about how Clojure handles repeated keys in kwargs, and thought it was "first one wins" (i.e. most explicit), but that seems not to be the case. That means the revision you reviewed had the wrong semantics, and fixing it breaks the usage pattern I wanted as well.

Thanks for the tip on `fdef`, I haven't used clojure spec before so I'll check that out. 

With schema, malli, and clj-kondo also in the mix it seems like there are a lot of ways to generate better errors for bad macro invocations, how does one decide? I went with modifying the existing kondo linter, as catching things statically is better than at compile time - for example preventing the commit versus letting each CI test bomb out individually. 

I'd love to get general guidance on using these tools, as they're all new to me and I'm still learning their strength and weaknesses.

src/metabase/models/setting.clj
crisptrutski: I really like the way we structure our macros to delegate to runtime function, and hadn't looked into why we we're going that route in this case. It turned out there were a few reasons, but with some experimentation I was able to find my way around them - I've cut a [standalone PR](https://github.com/metabase/metabase/pull/38128) to do this, and rebased this one so it doesn't have to fuss around following references or leaving gotchas like this.

crisptrutski: In practice I don't really like the ergonomics of explicit `assoc` and `merge` calls with this macro, I feel that we can be more opinionated and have a bit of sugar for the pattern I found myself wanting to use - inheriting from single "base type". I thought that a trailing "base" map was an elegant and minimal solution, but now I realize should *override* the key value literals to match Clojure semantics I'm going to fall back to some of the other ideas I had for that mechanism. 

While your example would be nice for a "multiple mixin" scenario, I'm not sure we have many use cases for that, and even so I would suspect these would not be "once off" combinations, and we might as well def the recipe rather than inline to construction. I'm still very wet behind the ears and haven't used settings in anger much, so perhaps there are good cases I just haven't thought of.

crisptrutski:
> I'm not sure I like this since it changes the syntax of `defsetting`, I think it's clearer if we only stick with key value pair.
> 
> And also, what happens if we have a setting like this?
> 
> ```clojure
> (defsetting str-setting
>  :visibility :internal)
> 
> (defsetting new-setting
>  :visibility :public
>  str-setting)
> ```
> 
> Logically I tink the value of `:visibility` for `new-setting` should be `:public`, but from reading the code it's not apparent because the map has to be at the tail of the macro.

Your example wouldn't work because `str-setting` would resolve to the setter function. I guess we *could* have some magic to look up the settings definition from the functions metadata... why are you putting this idea into my head 😄 

In the case where your first definition was just a map, then for revision 1 of this PR the visibility would have been `:public`, but with revision 2 it's gone to `:internal`, which I agree is non-intuitive and non-useful.

crisptrutski:
After some discussion and thought, I agree that it's unwise to support mixing key values with a trailing map, and if we're going to support maps at all (perhaps a YAGNI power feature if we're also adding `:base`) then it should be its own arity.

crisptrutski:
Actually if the user just wants to pass a map, they case use `(defsetting my-setting "docstring" :base options)`, so closing.

PR #38095, opened 2024-01-24
Run selected text loses parameters

### LINKED ISSUE #16584: "Run selected text" in Native editor does not include filter widget value and can return incorrect results
**Describe the bug**
When using filter variables in Native editor and selecting text to use the "Run selected text", then filter variables are not included, which throws an error. But if the filter has a default value, then that is used instead, leading to incorrect results.

**To Reproduce**
1. Native query > Sample Dataset > set mydate to Date filter
```sql
select count(*) from PRODUCTS
where year(CREATED_AT) = year({{mydate}}::date) --empty space
```
2. Set any date (today) in the filter widget and run query - shows **0** results as expected
image
3. Now select the query text and run it - fails with `Cannot run the query: missing required parameters: #{"mydate"}`
image
4. Now set "Default filter widget value" to `2019-12-01` and select the query text and run it - returns **18** results, which wrong
image

**Expected behavior**
"Run selected text" should include the variables selected.

Tested 0.37.8 thru 0.39.3 and `master`

**Additional context**
The "Run selected text" is not particular useful via the Run-button and just causes more confusion than good - #15315
There's already a context menu - #13450 - so perhaps let the Run-button and CTRL+Enter always run the entire query.

romeovs:
Closes https://github.com/metabase/metabase/issues/16584

### Description

The parameter values for a question aren't properly passed through to the card when going though the question → card → question loop. This PR changes the `overrideWithCard` parameter in `runQuestionQuery` with an `overrideWithQuestion` parameter, so we don't have to make that lossy loop. It also changes the one other call location where that parameter is used to make use of questions there.

I would like some feedback on whether or not the approach of moving to questions vs staying in card-land is the right one.

### How to verify

1. `+ New` → `SQL Query` → `Sample Database`
2. Type 
   ```sql
   SELECT * FROM "ACCOUNTS" WHERE  COUNTRY = {{country}} AND ID = 71;
   ```
3. Add the value `NL` for the country variable
4. Run the query en validate that you are getting results (one row).
5. Now select the following text in the query editor
   ```sql
   SELECT * FROM "ACCOUNTS" WHERE  COUNTRY = {{country}}
   ```
6. Run the selected text and verify there is no error and you are seeing more than one row.

### Checklist

### DIFF
--- a/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
+++ b/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
@@ -7,7 +7,7 @@ describe("issue 16584", () => {
     cy.intercept("POST", "/api/dataset").as("dataset");
   });
 
-  it("should pass parameters when running with 'Run select text' (metabase#16584)", async () => {
+  it("should pass parameters when running with 'Run select text' (metabase#16584)", () => {
     const editor = openNativeEditor();
 
     editor.type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country ");
--- a/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
+++ b/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
@@ -1,0 +1,24 @@
+import { restore, openNativeEditor } from "e2e/support/helpers";
+
+describe("issue 16584", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsNormalUser();
+    cy.intercept("POST", "/api/dataset").as("run");
+  });
+
+  it("should pass parameters when running with 'Run select text' (metabase#16584)", async () => {
+    const editor = openNativeEditor();
+
+    editor.type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country ");
+    editor.type("{selectAll}");
+
+    cy.get("input[placeholder='Country']").type("NL");
+
+    cy.get("button[aria-label='Get Answer']").first().click();
+    cy.wait("@run").then(({ request }) => {
+      const { body } = request;
+      expect(body.parameters[0].value).to.equal("NL");
+    });
+  });
+});
--- a/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
+++ b/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
@@ -4,7 +4,7 @@ describe("issue 16584", () => {
   beforeEach(() => {
     restore();
     cy.signInAsNormalUser();
-    cy.intercept("POST", "/api/dataset").as("run");
+    cy.intercept("POST", "/api/dataset").as("dataset");
   });
 
   it("should pass parameters when running with 'Run select text' (metabase#16584)", async () => {
@@ -16,7 +16,7 @@ describe("issue 16584", () => {
     cy.get("input[placeholder='Country']").type("NL");
 
     cy.get("button[aria-label='Get Answer']").first().click();
-    cy.wait("@run").then(({ request }) => {
+    cy.wait("@dataset").then(({ request }) => {
       const { body } = request;
       expect(body.parameters[0].value).to.equal("NL");
     });
--- a/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
+++ b/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
@@ -8,8 +8,12 @@ describe("issue 16584", () => {
   });
 
   it("should pass parameters when running with 'Run select text' (metabase#16584)", () => {
+    // The bug described in is #16584 can be further simplified:
+    // - the issue persists even when selecting the *entire* query
+    // - the issue is unrelated to using a date filter, using a text filter works too
+    // - the issue is unrelated to whether or not the parameter is required or if default value is set
     openNativeEditor()
-      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }}", {
+      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} ", {
         parseSpecialCharSequences: false,
         delay: 0,
       })
--- a/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
+++ b/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
@@ -13,7 +13,7 @@ describe("issue 16584", () => {
     // - the issue is unrelated to using a date filter, using a text filter works too
     // - the issue is unrelated to whether or not the parameter is required or if default value is set
     openNativeEditor()
-      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} ", {
+      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} LIMIT 10", {
         parseSpecialCharSequences: false,
         delay: 0,
       })
--- a/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
+++ b/e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
@@ -26,8 +26,6 @@ describe("issue 16584", () => {
 
     runNativeQuery();
 
-    cy.get(".Visualization").within(() => {
-      cy.findByText("NL").should("exist");
-    });
+    cy.get(".Visualization").findByText("NL").should("exist");
   });
 });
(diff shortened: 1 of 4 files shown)


kamilmielnik (commented):
Fix looks good :+1:

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { restore, openNativeEditor } from "e2e/support/helpers";
+
+describe("issue 16584", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsNormalUser();
+    cy.intercept("POST", "/api/dataset").as("run");
+  });
+
+  it("should pass parameters when running with 'Run select text' (metabase#16584)", async () => {
```
nemanjaglumac, line 10:
```suggestion
  it("should pass parameters when running with 'Run select text' (metabase#16584)", () => {
```

Why async?

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    const editor = openNativeEditor();
+
+    editor.type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country ");
+    editor.type("{selectAll}");
```
nemanjaglumac, line 14: Please don't do this.
If you take a look at the helper `openNativeEditor()`, you'll see that it actually visits a home page and then, starts the new > native query UI flow all over again. Apart from it being wasteful, it will not even work in this case.

Btw, please try to avoid assigning selectors to a `const` and use them directly.

nemanjaglumac (changes requested):

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
romeovs: I've fixed the assigning to a `const` bit, but I'm unclear as to why I shouldn't be using `openNativeEditor`, as it is currently being using in at least 48 test suites? Is there an alternative I should be using instead?

As to "it will not even work in this case": the test reproduces the issue and shows that the fix fixes it, can you elaborate on why you think it would not work?

kamilmielnik: > Apart from it being wasteful

We need at least one `cy.visit` call somewhere. Am I missing something?

> it will not even work in this case.

Can you elaborate on what will not work?

> Btw, please try to avoid assigning selectors to a `const` and use them directly.

Why?

nemanjaglumac: > Why?

https://learn.cypress.io/cypress-fundamentals/understanding-the-asynchronous-nature-of-cypress

https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Variables

frontend/src/metabase/query_builder/actions/core/core.js
```diff
@@ -74,12 +75,17 @@ export const reloadCard = createThunkAction(RELOAD_CARD, () => {
       Questions.actions.fetch({ id: outdatedQuestion.id() }, { reload: true }),
     );
     const card = Questions.HACK_getObjectFromAction(action);
+    const question = new Question(
+      card,
+      getMetadata(getState()),
+      outdatedQuestion.parameters(),
+    );
```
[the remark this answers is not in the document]
romeovs: I was assuming this was covered in the [saved question tests suite](https://github.com/metabase/metabase/blob/master/e2e/test/scenarios/question/saved.cy.spec.js). I can now see that that test suite does not have tests with parameters set, so that a wrong assumption.

OTOH, the behavior around parameters and reverting questions (which is the only place where `RELOAD_CARD` get's triggered AFAICT), seems broken on `master` too: the query gets run without parameters even if a parameter is set.
I can try and fix that (kind of unrelated) bug in this PR too, or create a separate issue and PR for that.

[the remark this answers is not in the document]
kamilmielnik: Let's file a separate issue for it :+1:

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -8,11 +8,18 @@ describe("issue 16584", () => {
...
     openNativeEditor()
-      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country ")
+      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} ", {
+        parseSpecialCharSequences: false,
+        delay: 0,
+      })
       .type("{selectAll}");
 
-    cy.findByPlaceholderText("Country").type("NL");
+    cy.findByPlaceholderText("Country").type("NL", { delay: 0 });
```
nemanjaglumac, line 22: FWIW, this is perfectly fine to leave without a delay. It's short enough.
You don't have to change it again. Just saying.

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    restore();
+    cy.signInAsNormalUser();
+    cy.intercept("POST", "/api/dataset").as("dataset");
+  });
+
+  it("should pass parameters when running with 'Run select text' (metabase#16584)", () => {
+    // The bug described in is #16584 can be further simplified:
+    // - the issue persists even when selecting the *entire* query
+    // - the issue is unrelated to using a date filter, using a text filter works too
+    // - the issue is unrelated to whether or not the parameter is required or if default value is set
```
nemanjaglumac, line 14: 👌

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+      .type("{selectAll}");
+
+    cy.findByPlaceholderText("Country").type("NL", { delay: 0 });
+
+    runNativeQuery();
+
+    cy.wait("@dataset").then(({ request }) => {
+      const { body } = request;
+      expect(body.parameters[0].value).to.equal("NL");
+    });
```
nemanjaglumac, line 29: This is testing implementation details.
Let's please add an assertion that checks whether or not actual results rendered.

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    cy.intercept("POST", "/api/dataset").as("dataset");
+  });
+
+  it("should pass parameters when running with 'Run select text' (metabase#16584)", () => {
+    // The bug described in is #16584 can be further simplified:
+    // - the issue persists even when selecting the *entire* query
+    // - the issue is unrelated to using a date filter, using a text filter works too
+    // - the issue is unrelated to whether or not the parameter is required or if default value is set
+    openNativeEditor()
+      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} ", {
```
nemanjaglumac, line 16: Nit:
You don't have to select all results for this particular repro. Cypress is already heavy on the resources.
I'd maybe go with `select country from...` or `select count(*)`. One additional, and very compelling reason could be that some tables have so many columns that some of them will be rendered outside of a viewport.

And since we use virtualization in the table, you might end up having to scroll.
Not saying it's the case here. Just sharing some of the traps I've ran into before.

In such cases, it's perfectly fine to limit the number of columns.

e2e/test/scenarios/native/reproductions/16584-run-selected-text-does-not-preserve-parameter-values.cy.spec.ts
```diff
@@ -13,7 +13,7 @@ describe("issue 16584", () => {
     // - the issue is unrelated to using a date filter, using a text filter works too
     // - the issue is unrelated to whether or not the parameter is required or if default value is set
     openNativeEditor()
-      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} ", {
+      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} LIMIT 10", {
```
nemanjaglumac, line 16:
```suggestion
      .type("SELECT * FROM ACCOUNTS WHERE COUNTRY = {{ country }} LIMIT 10 ", {
```

nemanjaglumac (approved):

frontend/src/metabase/query_builder/actions/querying.js
```diff
@@ -90,22 +87,20 @@ export const runDirtyQuestionQuery = () => async (dispatch, getState) => {
 };
 
 /**
- * Queries the result for the currently active question or alternatively for the card provided in `overrideWithCard`.
+ * Queries the result for the currently active question or alternatively for the card question provided in `overrideWithQuestion`.
```
nemanjaglumac, line 90: Nit
```suggestion
 * Queries the result for the currently active question or alternatively for the question provided in `overrideWithQuestion`.
```

PR #38098, opened 2024-01-24
Convert `ExplicitSize` to TypeScript

kulyk:
Converts the `ExplicitSize` HOC to TypeScript for better type-safety. Done in three steps:

1. Removed `wrapped` option in [`aefa6dc`](https://github.com/metabase/metabase/commit/aefa6dc7ea623c49b174d2a6ea8362b796d10c65), so there's less code to convert
2. Removed `AutoExpanding` HOC in [`a948aa1`](https://github.com/metabase/metabase/commit/a948aa1a53cdb934dc44bcc999c38059ed822227) — it was used by only one component and it was easier to remove it than convert to TypeScript as well
3. Converted the HOC to TypeScript and fixed the errors in [`915e7b4`](https://github.com/metabase/metabase/commit/915e7b46258dea186c865830b02f1b225cfef422) — because we've been using a JS HOC in a bunch of important and complex components (mostly charts) we accumulated some type errors

**To verify:** CI is green, visual tests are passing


frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -153,14 +151,27 @@ export function FieldValuesWidgetInner({
       disablePKRemappingForSearch,
     }),
   );
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
```
uladzimirdev, line 154: nit: when you set a boolean default value, no need to specify type

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
```diff
@@ -432,9 +443,9 @@ export function FieldValuesWidgetInner({
       <div
         data-testid="field-values-widget"
         style={{
-          width: expand ? maxWidth : containerWidth,
-          minWidth: minWidth,
-          maxWidth: maxWidth,
+          width: (isExpanded ? maxWidth : containerWidth) ?? undefined,
```
uladzimirdev, line 446: does it hurt if you pass `null`? (most likely you're covering this edge case with `??`

frontend/src/metabase/components/__mocks__/ExplicitSize.tsx
```diff
@@ -0,0 +1,10 @@
+const ExplicitSize = () => (ComposedComponent: any) => {
+  const WrappedComponent = (props: any) => (
+    <ComposedComponent width={1000} height={1000} {...props} />
+  );
+
+  return WrappedComponent;
+};
+
+// eslint-disable-next-line import/no-default-export
```
uladzimirdev, line 9: nit: we can remove this eslint rule from `__mocks__`

uladzimirdev (approved):
Significant work!

my only concern here is using `width: width ?? undefined` kind of constructions. Can't we get rid of `null` overall?

frontend/src/metabase/components/FieldValuesWidget/FieldValuesWidget.tsx
kulyk: Heah, types are defined like `width?: number` so it's not happy with nulls 🥴

frontend/src/metabase/components/__mocks__/ExplicitSize.tsx
kulyk: It's a typescript file, so eslint will actually complaint :(

uladzimirdev: I meant we can adjust eslint config to exclude such checks from `__mocks__`

Issue #13537, opened 2020-10-20
Chained filter endpoints fail without data permissions


paulrosenzweig:
The chained filter endpoints 403 when the user doesn't have data permissions. Access to these endpoints should match access to the dashboard itself.

Repro steps:
1. Create a dashboard with a filter.
2. Create a user with collection perms but no data perms.
2. Accessing that dashboard as that user will show a 200 for `GET /api/dashboard/:id` but 403 on `GET /api/dashboard/:id/params/:id/values` when you click on the filter.

camsaul:
That's the same behavior from the old field values endpoints... you have to have read perms for the underlying Field to be able to fetch values from it. Otherwise we'd be letting people bypass locked down perms and open the door to bad security situations. Maybe that behavior should be revisited, but it's been existing/intended behavior for years now so I don't think we should rush out a sudden change to it before shipping 37

camsaul:
The old endpoint does the same thing:

https://github.com/metabase/metabase/blob/master/src/metabase/api/field.clj#L171-L175

paulrosenzweig:
I think that behavior made sense on the old endpoint since there was no "parent" collection permission to grant you access to the field data. Here, I think we should take access to the dashboard as access to fully use the filters.

But yeah, I'm fine with waiting until 0.37.x to ship this. Rushing out permissions changes is scary.

camsaul:
I'm still not convinced we want to ship this anyway -- it just doesn't make sense when things like sandboxing are in play. See https://github.com/metabase/metabase-enterprise/issues/411

Basically what this issue is asking for is to treat any Users with no Field read perms whatsoever as if they had full unsandboxed read perms for that Field. So if you had sandboxed read perms you'd actually end up seeing less possible values than if the admin configured you to have no access whatsoever. It's a security behavior that is really confusing and really dangerous IMO.

Security implications WRT sandboxing aside, this isn't a trivial change. The value of Dashboard `:parameters` was really a frontend-only thing prior to 0.37.0. The backend didn't didn't read/validate this column at all -- if you have write perms for a Dashboard you could change it, and if you had read perms you could see it, but it was really just a column where the FE could store whatever metadata it wanted related to parameters.

[...]

PR #38104, opened 2024-01-24
Field values cache needs to check permissions

snoe:
If a user does not have permissions to field values, we should not return the cached field values.

Rely on the query processor permissions code, to check the query that would be run if the cache was not available. This considers gtap and block perms.

### DIFF
--- a/src/metabase/models/params/chain_filter.clj
+++ b/src/metabase/models/params/chain_filter.clj
@@ -74,12 +74,12 @@
    [metabase.models :refer [Field FieldValues Table]]
    [metabase.models.field :as field]
    [metabase.models.field-values :as field-values]
-   [metabase.models.interface :as mi]
    [metabase.models.params :as params]
    [metabase.models.params.chain-filter.dedupe-joins :as dedupe]
    [metabase.models.params.field-values :as params.field-values]
    [metabase.models.table :as table]
    [metabase.query-processor :as qp]
+   [metabase.query-processor.middleware.permissions :as qp.perms]
    [metabase.types :as types]
    [metabase.util :as u]
    [metabase.util.i18n :refer [tru]]
@@ -516,8 +516,14 @@
   [field-id]
   (and
     field-id
-    (mi/can-read? Field field-id)
     (field-values/field-should-have-field-values? field-id)))
+
+(defn- check-field-value-query-permissions!
+  "Check query permissions against the chain-filter-mbql-query (private #196)"
+  [field-id constraints options]
+  (->> (chain-filter-mbql-query field-id constraints options)
+       qp/preprocess
+       qp.perms/check-query-permissions*))
 
 (defn- cached-field-values [field-id constraints {:keys [limit]}]
   ;; TODO: why don't we remap the human readable values here?
@@ -562,7 +568,9 @@
      (-> (unremapped-chain-filter field-id constraints options)
          (update :values add-human-readable-values v->human-readable))
 
-     (and (use-cached-field-values? field-id) (nil? @the-remapped-field-id))
+     (and (use-cached-field-values? field-id)
+          (nil? @the-remapped-field-id)
+          (check-field-value-query-permissions! field-id constraints options))
      (cached-field-values field-id constraints options)
 
      ;; This is Field->Field remapping e.g. `venue.category_id `-> `category.name `;
--- a/src/metabase/models/params/chain_filter.clj
+++ b/src/metabase/models/params/chain_filter.clj
@@ -74,6 +74,7 @@
    [metabase.models :refer [Field FieldValues Table]]
    [metabase.models.field :as field]
    [metabase.models.field-values :as field-values]
+   [metabase.models.interface :as mi]
    [metabase.models.params :as params]
    [metabase.models.params.chain-filter.dedupe-joins :as dedupe]
    [metabase.models.params.field-values :as params.field-values]
@@ -515,6 +516,7 @@
   [field-id]
   (and
     field-id
+    (mi/can-read? Field field-id)
     (field-values/field-should-have-field-values? field-id)))
 
 (defn- cached-field-values [field-id constraints {:keys [limit]}]
--- a/src/metabase/models/params/chain_filter.clj
+++ b/src/metabase/models/params/chain_filter.clj
@@ -518,13 +518,12 @@
     field-id
     (field-values/field-should-have-field-values? field-id)))
 
-(defn- check-field-value-query-permissions!
+(defn- check-field-value-query-permissions
   "Check query permissions against the chain-filter-mbql-query (private #196)"
   [field-id constraints options]
   (->> (chain-filter-mbql-query field-id constraints options)
-         qp/preprocess
-         qp.perms/check-query-permissions*)
-  true)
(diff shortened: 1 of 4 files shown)


snoe:

src/metabase/models/params/chain_filter.clj
```diff
@@ -517,6 +518,14 @@
     field-id
     (field-values/field-should-have-field-values? field-id)))
 
+(defn- check-field-value-query-permissions!
+  "Check query permissions against the chain-filter-mbql-query (private #196)"
+  [field-id constraints options]
+  (->> (chain-filter-mbql-query field-id constraints options)
+         qp/preprocess
+         qp.perms/check-query-permissions*)
+  true)
```
camsaul, line 527: I don't think this really deserves an `!` since it's not done for the sort of side effects that [aren't safe in an STM context](https://guide.clojure.style/#naming-unsafe-functions), i.e. if you call this function 10 times it doesn't make any difference vs calling it once. It's still a pure function even if we're only calling it so maybe it will throw an exception.

I think calling it `check-field-value-query-permissions` is fine, that's generally the pattern we follow for stuff like this e.g. `check-valid-search-field` in this same namespace

src/metabase/models/params/chain_filter.clj
```diff
@@ -560,7 +569,9 @@
      (-> (unremapped-chain-filter field-id constraints options)
          (update :values add-human-readable-values v->human-readable))
 
-     (and (use-cached-field-values? field-id) (nil? @the-remapped-field-id))
+     (and (use-cached-field-values? field-id)
+          (nil? @the-remapped-field-id)
+          (check-field-value-query-permissions! field-id constraints options))
      (cached-field-values field-id constraints options)
```
camsaul, line 575: I think the way this code is structured sort of suggests that `check-field-value-query-permissions` can return falsey in some situations, when in reality the expectation is that it's meant to throw an exception if the check failures, and the return value doesn't matter. I think changing the code to this is more in line where how we do this sort of thing elsewhere and is clearer

```suggestion
          (nil? @the-remapped-field-id))
     (do
       (check-field-value-query-permissions field-id constraints options)
       (cached-field-values field-id constraints options))
```

camsaul: if you do this, then you don't need to have `check-field-value-query-permissions` return `true` (which seems a little icky tbh), you can just have it return nothing

test/metabase/api/dashboard_test.clj
```diff
@@ -2829,7 +2829,19 @@
                   :has_more_values false}
                  (->> (chain-filter-values-url (:id dashboard) (:category-name param-keys))
                       (mt/user-http-request :rasta :get 200)
-                      (chain-filter-test/take-n-values 3)))))))))
+                      (chain-filter-test/take-n-values 3)))))))
+
+    (testing "block data perms should not allow access to field values"
```
camsaul, line 2834: This `deftest` is already way too big, I really think you should just make this its own `deftest`. Otherwise debugging this is going to be a nightmare since you'll have to wade thru like 10 other separate test cases

src/metabase/models/params/chain_filter.clj
```diff
@@ -517,6 +518,14 @@
     field-id
     (field-values/field-should-have-field-values? field-id)))
 
+(defn- check-field-value-query-permissions!
+  "Check query permissions against the chain-filter-mbql-query (private #196)"
+  [field-id constraints options]
+  (->> (chain-filter-mbql-query field-id constraints options)
+         qp/preprocess
+         qp.perms/check-query-permissions*)
```
camsaul, line 526: indentation

```suggestion
  (->> (chain-filter-mbql-query field-id constraints options)
       qp/preprocess
       qp.perms/check-query-permissions*)
```

camsaul:
is there a related GH issue?

test/metabase/api/dashboard_test.clj
```diff
@@ -4048,3 +4060,31 @@
...
+                                              {:schemas :block})
+              (is (=? {:values (comp #(contains? % ["African"]) set)}
+                      (mt/user-http-request :crowberto :get 200
+                                            (str "dashboard/" (:id dashboard) "/params/" (:category-name param-keys) "/values")))))
+            (testing "Don't return with block perms."
+              (perms/update-data-perms-graph! [(:id (perms-group/all-users)) (mt/id) :data]
+                                              {:schemas :block})
+              (is (= "You don't have permissions to do that."
+                     (mt/user-http-request :rasta :get 403
+                                           (str "dashboard/" (:id dashboard) "/params/" (:category-name param-keys) "/values")))))))))))
```
camsaul, line 4090: We need a test case for people with "no self-service" (no data perms whatsoever, but no block bit set either)... that should still work (see https://github.com/metabase/metabase/issues/13537)

test/metabase/api/dashboard_test.clj
```diff
@@ -4048,3 +4060,31 @@
             (testing "We have values and they all match the expression concatenation"
               (is (pos? (count values)))
               (is (every? (partial re-matches #"[^🦜]+🦜🦜🦜[^🦜]+") (map first values))))))))))
+
+(deftest param-values-permissions-test
+  (testing "Users without permissions should not see all options in a dashboard filter (#196)"
```
camsaul, line 4065:
```suggestion
  (testing "Users without permissions should not see all options in a dashboard filter (private#196)"
```

test/metabase/api/dashboard_test.clj
camsaul, line 2834:
```suggestion
    (testing "block data perms should not allow access to field values (private#196)"
```

test/metabase/api/dashboard_test.clj
```diff
@@ -4048,3 +4060,31 @@
...
+    (when config/ee-available?
+      (mt/with-premium-features #{:advanced-permissions}
+        (mt/with-temp-copy-of-db
+          (with-chain-filter-fixtures [{:keys [dashboard param-keys]}]
+            (testing "Return values with access"
+              (is (=? {:values (comp #(contains? % ["African"]) set)}
+                      (mt/user-http-request :rasta :get 200
+                                            (str "dashboard/" (:id dashboard) "/params/" (:category-name param-keys) "/values")))))
+            (testing "Return values with self-service (#26874)"
```
camsaul, line 4074:
```suggestion
            (testing "Return values with no self-service (#26874)"
```

that's what this is testing, right?

camsaul (commented):
I was going to approve this but since it's marked auto-merge I'm going to hold off on doing so, I left a few style suggestions and some notes about where I think we should add some more comments explaining what's going on. For perms code since it's super important it's good to make it extra explicit and add lots of notes, I just spent like half an hour freaking out thinking this was going to cause a regression of  #13537 and then manually playing around with it in the REPL to be sure it wouldn't, a few comments could have saved me and future people looking at it a lot of trouble.

camsaul (approved):
well this LGTM but I would still like to see `check-field-value-query-permissions` either have a comment mentioning `*param-values-query*` or explicitly bind it... let's not make the perms code here more confusing than it already is

PR #27391, opened 2022-12-22
Allowing users to hide logo on embedding

WiNloSt:
Issue: #27293

Allow users to disable/hide logo in embedding by the param `logo` (defaults to `true`)

#### How to test
1. Run Metabase with this branch as an enterprise user (port 3000)
1. Configure enterprise token
1. Enable JWT authentication and make sure you set the secret to `ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` https://github.com/metabase/sso-examples/blob/master/app-embed-example/server.js#L9-L10
1. Clone https://github.com/metabase/sso-examples
1. Go to `/app-embed-example/`
1. Run `yarn dev` (it will start port 3001 and 3002)
1. Go to http://localhost:3001/analytics and test if embedding shows breadcrumbs as desired.
1. You could try append different parameters to see it in action e.g.
    1. `?logo=false`
    2. `?logo=false&side_nav=true`
    3. `?logo=false&breadcrumbs=false`
    4. `?breadcrumbs=false` 
1. Test mobile view.

- - -
This change is [Reviewable](https://reviewable.io/reviews/metabase/metabase/27391)

Issue #27293, opened 2022-12-16
Offer a new full-app embedding url param for logo


cdeweyx:
## Context
We offer several several url parameters for customers in full-app embedding to configure how their Metabase instance looks in the embedded iFrame. We recently made the `top_nav` default visible and added `breadcrumbs` as an option in https://github.com/metabase/metabase/pull/26977. 

With this new behavior, there's a new twist of "I want breadcrumbs so that my users can navigate the instance, but I already show a logo in my app, so having the logo in the top nav as well is duplicative and looks bad." Example below.

Let's offer a new `logo` sub-param for `top_nav` that is default on. When it's disabled, breadcrumbs should be scooted over to be left-aligned. Same logic as before if all sub-params are disabled, the whole top-nav is disabled as it serves no purpose.

![Image](https://user-images.githubusercontent.com/16455495/208149955-8054ff1a-4b32-4f01-abf3-5b29cdbe4adc.png)

bunnymatt:
We got creative and changed the logo to be the sidebar icon :) 

![Screen Shot 2022-12-16 at 11 29 45 AM](https://user-images.githubusercontent.com/103591073/208165526-9f79d2cd-cd62-47fc-8cb3-2962bcc727fb.png)

brunobergher:
Here's the spec we agreed to on Slack.

**Param**
The parameter is called `logo`, default `true`.

**States**
If `logo: true` and `sidenav: true`, it looks like regular Metabase
If `logo: true` and `sidenav: false`, just don’t show the sidenav hover icon on the logo.
If `logo: false` and `sidenav: true`, we always show the sidenav icon, as in the comment above, but with the usual color.
If `logo: false` and `sidenav: false`, the breadcrumbs move all the way to the left.

PR #38126, opened 2024-01-25
More tidying prior to Models in Browse Data

rafpaf:
The PR makes some e2e tests and associated components more semantic


e2e/test/scenarios/embedding/embedding-full-app.cy.spec.js
```diff
@@ -135,15 +137,14 @@ describeEE("scenarios > embedding > full app", () => {
...
         qs: { side_nav: false, logo: false },
       });
-      // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-      cy.findByText("Our data").should("be.visible");
-      // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-      cy.findByText("Our analytics").should("not.exist");
+      cy.findByRole("heading", { name: /Our data/ }).should("be.visible");
+      cy.findByRole("treeitem", { name: /Our data/ }).should("not.exist");
+      cy.findByRole("treeitem", { name: "Our analytics" }).should("not.exist");
       appBar().should("not.exist");
```
rafpaf, line 148: I'm not totally sure I've understood what this test is about

e2e/test/scenarios/onboarding/auth/signin.cy.spec.js
```diff
@@ -30,40 +30,36 @@ describe("scenarios > auth > signin", () => {
...
     cy.findByLabelText("Email address").type(admin.email);
     cy.findByLabelText("Password").type("INVALID" + admin.password);
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("Sign in").click();
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.contains("did not match stored password");
+    cy.button("Sign in").click();
+    cy.findByRole("alert")
+      .filter(':contains("did not match stored password")')
+      .should("be.visible");
```
rafpaf, line 36: `cy.findByRole("alert", {name: ...` does not work; this longer formula is needed.

iethree (changes requested):
looks great, except we don't want to keep adding more chainers off of `cy.`

e2e/support/commands/ui/button.ts
```diff
@@ -34,4 +46,21 @@ Cypress.Commands.add(
...
+      name: headingName,
+      timeout,
+    };
+
+    return subject
+      ? cy.wrap(subject).findByRole("heading", config)
+      : cy.findByRole("heading", config);
+  },
+);
+
```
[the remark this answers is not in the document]
nemanjaglumac: Yup, what Ryan said.
If anything, we want to reduce the amount of custom commands.

`findByRole` is already a custom command - it comes from [Cypress Testing Library](https://testing-library.com/docs/cypress-testing-library/intro/).

Can we please use it directly as is?

e2e/test/scenarios/embedding/embedding-full-app.cy.spec.js
nemanjaglumac: IIRC, @WiNloSt originally wrote these tests.
I think the test is doing what it says on the tin - if we hide the side navigation **and** hide the logo, we want to make sure the top bar is not rendered.

There is a test below that explicitly makes sure nav bar is hidden via param.

nemanjaglumac (changes requested):

e2e/test/scenarios/embedding/embedding-full-app.cy.spec.js
rafpaf: Thank you for helping me understand. I've reverted that change

e2e/support/commands/ui/button.ts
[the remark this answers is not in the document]
rafpaf: Thanks both. I've removed that extra helper function.

iethree (approved):
🚀

e2e/test/scenarios/embedding/embedding-full-app.cy.spec.js
WiNloSt: @nemanjaglumac got it exactly right. That test was added during the implementation of hiding Metabase logo.

For context if you're interested:
- [PR](https://github.com/metabase/metabase/pull/27391)
- [Issue](https://github.com/metabase/metabase/issues/27293)

nemanjaglumac (approved):
Thanks for improving a11y and for addressing the reviews.

PR #38127, opened 2024-01-25
FE - Migrate `dataset: true` to enum value

### LINKED ISSUE #37366: FE - Migrate `dataset: true` to enum value
Depends on #37367

See [this comment](https://github.com/metabase/metabase/issues/37367#issuecomment-1905426067) for the list of affected endpoints.
~At minimum, this affects `Card`, `Bookmark` and `CardCompletionItem` types in FE.~

----

Related [Slack discussion](https://metaboat.slack.com/archives/C0641E4PB9B/p1705576281775139)

kamilmielnik:
Closes #37366

Depends on #38074

### Description

#38074 description says that the only breaking changes are in the PUT/POST `/api/card/:id`, that's why `dataset` is not ported everywhere yet. We will do this incrementally.

- Made the new `type` attribute "writable" in `entities/questions.js` - this is required to pass it to aforementioned endpoints
- All e2e tests are updated to use `type` instead of `dataset`
  - This was needed due to breaking changes announced in the BE PR (#38074) - many e2e tests call PUT `/api/card/:id`
- Introduced new `Question.prototype.type`
  - There already was a method by that name but there were only few usages, so I inlined them
- `Question.prototype.isDataset` has been deprecated in favor of `Question.prototype.type`
  - I didn't replace usages yet because it requires lots of changes (I started but went into a rabbit hole - that's why many commits are reverted) - we can do this incrementally, as needed
- Introduced `Question.prototype.setType`
- `Question.prototype.setDataset` has been deprecated in favor of `Question.prototype.setType`
  - renamed to `_setDataset` and made it `private`
  - replaced all usages outside of `Question`
- `Card["dataset"]` has been deprecated in favor of `Card["type"]`


e2e/snapshot-creators/default.cy.snap.js
```diff
@@ -214,7 +214,7 @@ describe("snapshots", () => {
     cy.createQuestion({
       name: "Orders Model",
       query: { "source-table": ORDERS_ID },
-      dataset: true,
+      type: "model",
```
nemanjaglumac, line 217: 😍

Can't believe this is happening finally!

nemanjaglumac (approved):
Thanks for the useful comments and explanations.
The diff looks pretty straightforward mostly.

The only unfortunate thing is the clash with types and the temporary workaround.

PR #38128, opened 2024-01-25
Evaluate all defsetting arguments at runtime

crisptrutski:
### Description

This refactor changes `defsetting` so that we only process its arguments at runtime, allowing us to use references and expressions to define things dynamically.

~~Note that this PR does *not* add support for a trailing kwargs map, that's left in https://github.com/metabase/metabase/pull/38090, which I plan to rebase on this branch.~~ Scrapped that feature anyway.

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -1206,20 +1206,22 @@
         ;; preserve metadata from the `setting-symbol` passed to `defsetting`.
         setting-getter-fn-symbol setting-symbol
         setting-setter-fn-symbol (-> (symbol (str (name setting-symbol) \!))
-                                     (with-meta (meta setting-symbol)))]
-    `(let [setting-options#   (merge ~options ~setting-metadata)
-           setting-definition# (register-setting! setting-options#)]
-       (when (and ~maybe-i18n-exception (#'requires-i18n? setting-definition#))
-         (throw ~maybe-i18n-exception))
-       (-> (def ~setting-getter-fn-symbol (setting-fn :getter setting-definition#))
-           (alter-meta! merge (setting-fn-metadata :getter setting-definition#)))
+                                     (with-meta (meta setting-symbol)))
+        setting-definition-symbol (gensym "setting-")]
+    `(let [setting-options#          (merge ~options ~setting-metadata)
+           ~setting-definition-symbol (register-setting! setting-options#)]
+       ~(when maybe-i18n-exception
+          `(when (#'requires-i18n? ~setting-definition-symbol)
+             (throw ~maybe-i18n-exception)))
+       (-> (def ~setting-getter-fn-symbol (setting-fn :getter ~setting-definition-symbol))
+           (alter-meta! merge (setting-fn-metadata :getter ~setting-definition-symbol)))
        ;; unfortunately we can't evaluate this condition at compile time, as the options might contain runtime forms.
-       (when (not= (:setter setting-definition#) :none)
+       (when (not= (:setter ~setting-definition-symbol) :none)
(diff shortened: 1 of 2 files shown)


crisptrutski:
* `master`
  * **PR #38128** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38128?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38128?utm_source=stack-comment).

src/metabase/models/setting.clj
```diff
@@ -1055,24 +1055,29 @@
...
-  (str/ends-with? (ns-name *ns*) "-test"))
+    `(ex-info (str "defsetting docstrings must be a *deferred* i18n form unless the Setting has"
+                   " `:visibility` `:internal`, `:setter` `:none`, or is defined in a test namespace."
+                   (format " Got: ^%s %s"
+                           (some-> ~description-form class (.getCanonicalName))
+                           (pr-str ~description-form)))
+              {:description-form ~description-form})))
+
+;; This exists as its own method so that we can stub it in tests
+(defn- in-test? [ns] (str/ends-with? ns "-test"))
```
qnkhuat, line 1074:
```suggestion
(defn- ns-in-test? [ns] (str/ends-with? ns "-test"))
```

qnkhuat: 2 things:
- I think we should prefix the name with `ns`, it was fine before if it does not take an argument
- we should call `ns-name` on `ns` because I usually think when we see `ns,` we think of it as a symbol, not str. So we should avoid that footgun.

test/metabase/models/setting_test.clj
```diff
@@ -622,6 +622,23 @@
              (settings-last-updated-value-in-db))))))
 
 
+;;; ---------------------------------------------- Runtime Setting Options ----------------------------------------------
+
+(def my-default 3)
+
+(defsetting test-dynamic-setting
```
qnkhuat, line 629: I don't think this test reflects your change since it passes without it.

crisptrutski: Ah, good catch! Forgot how subtle the compile time gotchas were. Kept these assertions as I think this behavior is important too, but added another assertion that shows we fixed a gotcha with settings that are used in the conditional logic.

src/metabase/models/setting.clj
crisptrutski: 1. It feels non-idiomatic to repeat the name of the arg in the function name for a predicate, but I don't feel strongly.
2. The ns here *is* the symbol, it turns out the `str/end-with?` implicitly handles converting to a string.

src/metabase/models/setting.clj
```diff
@@ -1055,24 +1055,29 @@
...
-  (str/ends-with? (ns-name *ns*) "-test"))
+    `(ex-info (str "defsetting docstrings must be a *deferred* i18n form unless the Setting has"
+                   " `:visibility` `:internal`, `:setter` `:none`, or is defined in a test namespace."
+                   (format " Got: ^%s %s"
+                           (some-> ~description-form class (.getCanonicalName))
+                           (pr-str ~description-form)))
+              {:description-form ~description-form})))
+
+;; This exists as its own method so that we can stub it in tests
+(defn- ns-in-test? [ns-name] (str/ends-with? ns-name "-test"))
```
crisptrutski, line 1074: Shower thought - instead of stubbing this we could cover its logic by switching namespaces within the test. Not sure which is the lesser evil 🙂

calherries (approved):
LGTM

PR #38129, opened 2024-01-25
Fix typo in a password reset token defendpoint docstring

nemanjaglumac:


src/metabase/api/session.clj
```diff
@@ -296,7 +296,7 @@
       (api/throw-invalid-param-exception :password (tru "Invalid reset token"))))
 
 (api/defendpoint GET "/password_reset_token_valid"
-  "Check is a password reset token is valid and isn't expired."
+  "Check is a password reset token valid and isn't expired."
```
paoliniluis, line 299: Shouldn't it be "Check if a password reset token is valid and isn't expired"

dragonsahead (changes requested):
Left 1 comment

src/metabase/api/session.clj
nemanjaglumac: I think both work.
But I do like your suggestion better.

dragonsahead (approved):
LGTM!

PR #38131, opened 2024-01-25
remove extra border from recipient picker in dash subs

kdoh:
Removes an extraneous border from the To: Input in dashboard subscriptions. While normally I'm all for two for the price of one, this seems a little excessive.

Before:
Screenshot 2024-01-25 at 8 29 30 AM

After:
Screenshot 2024-01-25 at 9 12 35 AM


oleggromov (approved):
LGTM 👍

PR #38134, opened 2024-01-25
Add :impersonated to the cond in a debug log to fix impersonation on stats

noahmoss:
Connection impersonation can't be set on stats because they have log level set to `debug`, which is hitting this `cond`, which doesn't have a case for `:impersonated` so it's throwing an error.


noahmoss:
* `master`
  * **PR #38134** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38134?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38134?utm_source=stack-comment).

escherize (approved):
Makes sense. We must  handle the `changes` in the `case` if the `if` can lead here.

PR #38139, opened 2024-01-25
Fixes settings that should default to `false` returning null

### LINKED ISSUE #38138: Settings with default of `false` now return null
### Describe the bug

After commit 528b9c59c2058e4e72a1ddd85e1c7ed5a982e690 settings that have a default of `false` now return null, causing issues on the FE was expecting false/true

### To Reproduce

Create a new instance, check that the value of the setting `enable-public-sharing` is null, instead of false

Slack thread: https://metaboat.slack.com/archives/CKZEMT1MJ/p1706196655996729

npretto:
Closes #38138

I believe the issue was that `(:default setting)` returned the default value, and in the case of `:default false` it returned false -> it didn't return the default value

# How to reproduce

- Nuke the db
- Start metabase
- complete setup
- see that the settings are correct, one example is `enable-public-sharing` that should return `false` instead of `null`

### DIFF
--- a/src/metabase/models/setting.clj
+++ b/src/metabase/models/setting.clj
@@ -569,7 +569,7 @@
                      env-var-value
                      db-or-cache-value
                      (cond
-                       (:default setting) default-value
+                       (some?(:default setting)) default-value
                        (:init setting)    (when-not *disable-init*
                                             init!))]]
      (loop [[f & more] source-fns]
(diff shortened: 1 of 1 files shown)


crisptrutski (approved):
Thanks for fixing - I'll look at adding a regression test.

PR #38151, opened 2024-01-25
Implement MLv2 dependent-metadata

### LINKED ISSUE #36974: [MLv2] [BE] Dependent metadata
Existing `dependentMetadata()` function returns the list of all entities which should be fetched according to the current question/query. It returns data like this:

```
[
  { id: 1, type: "field" }
  { id: 2, type: "table" }
  { id: "card__3", type: "table" },
  { id: 4, type: "schema" } // here id is actually the database id
]
```

We use the following endpoints to fetch data:
- `field` => `/api/field/:fieldId`
- `table` => `/api/table/:tableId/query_metadata`, which also handles `card__:id`
- `schema` => `/api/database/:dbId/schemas`

Currently `dependentMetadata` never returns databases assuming that they are already loaded. It seems to work only because the full database list is loaded for the navigation bar which seems like an oversight in the existing implementation.

Existing implementation:
- Question https://github.com/metabase/metabase/blob/ed7295b5d6cdcf26bb81844c9559b3a1eb155b04/frontend/src/metabase-lib/Question.ts#L859
- StructuredQuery https://github.com/metabase/metabase/blob/5414bde814cef11e2fc6ee1686f697c83fb4085a/frontend/src/metabase-lib/queries/StructuredQuery.ts#L1552
- NativeQuery https://github.com/metabase/metabase/blob/5414bde814cef11e2fc6ee1686f697c83fb4085a/frontend/src/metabase-lib/queries/NativeQuery.ts#L445

Every time the query changes we call `dependentMetadata` and make sure all items are loaded. Here is the existing code that does that:
- https://github.com/metabase/metabase/blob/1490fd9cc8d3b7a9ea97aabb49eccb9547dc073a/frontend/src/metabase/questions/actions.ts#L13
- https://github.com/metabase/metabase/blob/ad9742cd413494583b02b35a94fa6e5b49f0cb52/frontend/src/metabase/redux/metadata.js#L299

In the scope of this task is to expose a MLv2 function that returns the list of objects using similar format that the FE needs to load, including databases, using existing `dependentMetadata` for the reference, loading data for foreign key tables. Proposed signature:

```
dependentMetadata(query: Query): DependentItem[];

type DatabaseItem = { id: DatabaseId, type: "database" };
type SchemaItem = { id: DatabaseId, type: "schema" }; 
type TableItem = { id: TableId, type: "table" }; 
type FieldItem = { id: FieldId, type: "field"; };
type DependentItem = DatabaseItem | SchemaItem | TableItem | FieldItem;

```

metamben:
Fixes #36974.

See the Slack [discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1706193649444639) for details.

### DIFF
--- a/src/metabase/lib/fe_util.cljc
+++ b/src/metabase/lib/fe_util.cljc
@@ -3,13 +3,16 @@
    [metabase.lib.common :as lib.common]
    [metabase.lib.field :as lib.field]
    [metabase.lib.filter :as lib.filter]
+   [metabase.lib.metadata :as lib.metadata]
    [metabase.lib.metadata.calculation :as lib.metadata.calculation]
    [metabase.lib.options :as lib.options]
    [metabase.lib.schema :as lib.schema]
    [metabase.lib.schema.common :as lib.schema.common]
    [metabase.lib.schema.expression :as lib.schema.expression]
+   [metabase.lib.schema.id :as lib.schema.id]
    [metabase.lib.schema.temporal-bucketing :as lib.schema.temporal-bucketing]
    [metabase.lib.temporal-bucket :as lib.temporal-bucket]
+   [metabase.lib.types.isa :as lib.types.isa]
    [metabase.lib.util :as lib.util]
    [metabase.mbql.util :as mbql.u]
    [metabase.shared.util.i18n :as i18n]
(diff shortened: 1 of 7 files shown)


src/metabase/lib/fe_util.cljc
```diff
@@ -108,3 +111,53 @@
...
       (lib.metadata.calculation/display-name query stage-number filter-clause))))
+
+(defn- query-dependents
+  [metadata-providerable query-or-join]
+  (let [base-stage (first (:stages query-or-join))
+        database-id (:database query-or-join -1)]
+    (concat
+     (when (pos? database-id)
+       [{:type :database, :id database-id}
+        {:type :schema,   :id database-id}])
```
metamben, line 122: I always return `:database` and `:schema` in pairs, which seems a bit fishy. Either there should be situations when one of them is missing, or just one of them should be enough.

test/metabase/lib/test_util.cljc
```diff
@@ -94,22 +103,23 @@
...
+                                 :coercion_strategy  nil
+                                 :name               "USER_ID"
+                                 :settings           nil
+                                 :source             :breakout
+                                 :field_ref          [:field (meta/id :checkins :user-id) nil]
+                                 :effective_type     :type/Integer
+                                 :nfc_path           nil
+                                 :parent_id          nil
+                                 :id                 (meta/id :checkins :user-id)
+                                 :fk_target_field_id (meta/id :users :id)
```
metamben, line 118: This was missing, I'm not sure why.

kamilmielnik (commented):
FE changes look good :+1:

test/metabase/lib/test_util.cljc
snoe: I believe this was added due to a the drill thru bugs.
Should we cache an actual qp response so we don't have to keep it in sync?

src/metabase/lib/fe_util.cljc
snoe: I would rather drop `:schema` in the FE if it means the same as `:database` or figure out when these can be different. Going forward with this, feels like it will always be a source of confusion.

snoe (approved):
I think the `database` and `schema` keys are confusing, either should be fixed in FE here or as followup.

Issue #36367, opened 2023-12-04
[Epic] Show more Store account info in the app itself


luizarakaki:
- [Product doc](https://www.notion.so/metabase/Show-more-Store-account-info-in-the-app-itself-85d6c436695e467c89633f476945783b?pvs=4)
- [Harbormaster issue](https://github.com/metabase/harbormaster/issues/4591)
- [Backend design doc](https://www.notion.so/metabase/Show-store-account-info-in-the-app-itself-Design-Doc-0c4843f8a25844d4ad11ade0837680a6)
```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/issues/38195
- [ ] https://github.com/metabase/metabase/issues/38197
```

PR #38153, opened 2024-01-25
Add BE for routing billing status

### LINKED ISSUE #38197: add backend ee route to hit HM
Expose an endpoint in an enterprise route that will hit HM, authenticate with email, token, and possibly version(?) to proxy to the FE.

Needs to cache the response. Not exactly clear on how long but on the magnitude of hours seems right.

Handle:
- 200 response
- 500 response
- expired token
- unrecognized token

Decisions we can make in the future (or now):
- can we send the version the FE expects to HM and it can always speak that version?
- if not, if HM responds with a higher version should MB not return any info? Or return a version that the FE won't speak?

qwef:
Closes #38197 

For https://github.com/metabase/metabase/issues/36367

### Description

BE used for routing information from the client to harbormaster.


src/metabase/api/billing.clj
```diff
@@ -0,0 +1,41 @@
...
+                         e)))))
+   :ttl/threshold (u/hours->ms 5)))
+
+(api/defendpoint GET "/"
+  "Get billing information. This acts as a proxy between `metabase-billing-info-url` and the client,
+   using the embedding token and signed in user's email to fetch the billing information."
+  []
+  (let [token    (premium-features/premium-embedding-token)
+        email    (t2/select-one-fn :email :model/User :id api/*current-user-id*)
+        language (public-settings/site-locale)]
```
sloansparger, line 38: should this be the user's local?

dpsutton: good call. `(i18n/user-locale-string)` is probably what we want

enterprise/backend/src/metabase_enterprise/api/routes.clj
```diff
@@ -28,6 +29,9 @@
   ;; and follow the convention.
   (compojure/context
    "/ee" []
+   (compojure/context
+    "/billing" []
+    billing/routes)
```
qwef, line 34: Should I add a premium feature here? Unsure because it seems like anyone with a token should access this, but no other route is like this.

noahmoss: This seems fine

enterprise/backend/src/metabase_enterprise/billing/billing.clj
```diff
@@ -0,0 +1,42 @@
+(ns metabase-enterprise.billing.billing
+  "`/api/ee/billing/` endpoint(s)"
+  (:require
+   [cheshire.core :as json]
+   [clj-http.client :as http]
+   [clojure.core.memoize :as memoize]
+   [compojure.core :as compojure :refer [GET PUT]]
```
noahmoss, line 7:
```suggestion
   [compojure.core :as compojure :refer [GET]]
```

enterprise/backend/src/metabase_enterprise/billing/billing.clj
```diff
@@ -0,0 +1,42 @@
...
+   [metabase.api.common :as api]
+   [metabase.public-settings.premium-features :as premium-features]
+   [metabase.util :as u]
+   [metabase.util.i18n :as i18n]
+   [toucan2.core :as t2])
+  (:import    [com.fasterxml.jackson.core JsonParseException]))
```
noahmoss, line 13:
```suggestion
  (:import
   (com.fasterxml.jackson.core JsonParseException)))
```

noahmoss (approved):
lgtm

PR #38157, opened 2024-01-25
Remove useless warnings from defendpoint missing a malli schema

escherize:
in defendpoint, we knowingly call umd/describe on a possibly-nil value (`nil` is not a malli schema). When unable to create a schema, malli will print a warning about that, however that's ok in this instance, so don't warn about it.

This might be a good place to log a warning about having a missing schema for a var, but didn't want to make that choice here. for now let's just get rid of the unhelpful warning.

### DIFF
--- a/src/metabase/api/common/internal.clj
+++ b/src/metabase/api/common/internal.clj
@@ -78,7 +78,10 @@
 (defn- dox-for-schema
   "Generate the docstring for `schema` for use in auto-generated API documentation."
   [schema route-str]
-  (try (with-out-str (umd/describe schema))
+  (try
+    ;; we can ignore the warning printed by umd/describe when schema is `nil`.
+    (binding [*out* (new java.io.StringWriter)]
+      (umd/describe schema))
        (catch Exception _
          (ex-data
           (when (and schema config/is-dev?) ;; schema is nil for any var without a schema. That's ok!
--- a/src/metabase/api/common/internal.clj
+++ b/src/metabase/api/common/internal.clj
@@ -78,7 +78,7 @@
 (defn- dox-for-schema
   "Generate the docstring for `schema` for use in auto-generated API documentation."
   [schema route-str]
-  (try (umd/describe schema)
+  (try (with-out-str (umd/describe schema))
        (catch Exception _
          (ex-data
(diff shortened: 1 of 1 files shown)


qnkhuat:
how do I get that warning message?

unrelated but do you how to fix these errors?

```
-- Schema Creation Error --------------------------------- NO_SOURCE_FILE:201 --

Invalid Schema

  ms/PositiveInt

More information:

  https://cljdoc.org/d/metosin/malli/CURRENT  
```

You can see a lot of these errors for endpoints that have a schema for URL params.
try to evaluate `api/defendpoint GET "/:id"`

piranha:
@qnkhuat oh, I got those warnings too, but restart helped me. Is it reliable for you?

@escherize is it maybe better to just check for `nil` rather than chew all the errors it reports to stdout? I don't know if it does anything else though, but `(when schema (umd/describe schema))` seems simpler than binding `*out*`.

escherize:
Well, the reason I didn't check for `nil`: it's because the only thing that's changed is that there's an error logged when the call to `umd/describe` throws. This is a less invasive change than using when, all it does is ignore the printed output. It is the same as `u/with-out-str` except it ignores the string instead of returning it (like it used to).

sansolovyov (approved):
Agree, in case if it throws an error there is a catch clause waiting :)

escherize:
> how do I get that warning message?

@qnkhuat Those are printed when there is a missing schema for a defendpoint form. so if you were to remove a line from a defendpoint, e.g. `PUT api/user/:id`, and compile it, those errors would print, but they don't indicated anything. Now they don't.

```
(api/defendpoint PUT "/:id"
  [id :as {{:keys [email first_name last_name user_group_memberships
                   is_superuser is_group_manager login_attributes locale] :as body} :body}]
  {id                     ms/PositiveInt
   email                  [:maybe ms/Email]
   first_name             [:maybe ms/NonBlankString]
   last_name              [:maybe ms/NonBlankString]
   user_group_memberships [:maybe [:sequential user/UserGroupMembership]]
   is_superuser           [:maybe :boolean]
   is_group_manager       [:maybe :boolean]
   login_attributes       [:maybe user/LoginAttributes] ;; <- remove one of these lines
   locale                 [:maybe ms/ValidLocale]}
  ,,,
  )
```

PR #38158, opened 2024-01-25
Update permissions QP middleware to use data-permissions APIs

noahmoss:
Pursuant to https://github.com/metabase/metabase/issues/38038

This PR updates the QP permissions middleware to enforce data permissions based on the new `data_permissions` table and APIs. It also updates block permissions and sandboxing accordingly. 

This does _not_ update any code related to download perms, data model editing, or database management permissions. It also doesn't update connection impersonation use the new table. These will come in the next PR.

**Don't be afraid of the line count** — a _lot_ of lines come from wrapping various tests in `mt/with-full-data-perms-for-all-users!`. Depending on test order, perms weren't necessarily being reset to their defaults when running the full test suite, so a lot of tests were failing if they expected `:rasta` to have full query access to the test dataset. I've fixed these tests by wrapping them with this macro to ensure nothing is dependent on app DB state, but this adds a lot more changed lines to the PR. There are lots of other tests repaired in other small ways, but for the most part tests worked fairly well despite this change.


enterprise/backend/src/metabase_enterprise/sandbox/query_processor/middleware/row_level_restrictions.clj
```diff
@@ -333,7 +330,9 @@
       original-query
       (-> sandboxed-query
           (assoc ::original-metadata (expected-cols original-query))
-          (update-in [::qp.perms/perms :gtaps] (fn [perms] (into (set perms) (sandboxes->perms-set (vals table-id->gtap)))))))))
+          (update-in [::query-perms/perms :gtaps]
+                     (fn [required-perms] (merge required-perms
```
johnswanson, line 334: If I understand correctly, the parameter here should be something like `existing-perms` or `perms`? We're figuring out what additional perms we need for the sandboxed query to run, and adding them to the set of permissions the user already has - right?

johnswanson (approved):
LGTM!

PR #38160, opened 2024-01-25
Fix Snowflake sync not returning all schemas regression in 48+

### LINKED ISSUE #38135: Metabase won't see schemas other than public or the default namespace of the user in snowflake

camsaul:
Fixes #38135

The root issue was that if you had a default schema (which I think you can set in the Snowflake admin console somehow) or session schema in play (i.e., `USE SCHEMA x` or `&schema=x` in the additional JDBC options), the new Snowflake sync code in #36121 would not return Tables in any other schema, causing them to get marked inactive. This was not a problem in earlier versions of MB (47 and earlier) since it used different sync code that still worked regardless of default schema.

I think this is actually an upstream bug with the Snowflake JDBC driver... the new sync code called [`DatabaseMetaData.getTables()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/DatabaseMetaData.html#getTables(java.lang.String,java.lang.String,java.lang.String,java.lang.String%5B%5D)) with a `null` schema pattern, and the JDBC spec says

> `schemaPattern` - a schema name pattern; must match the schema name as it is stored in the database; `""` retrieves those without a schema; `null` means that the schema name should not be used to narrow the search

And the Snowflake JDBC driver is clearly narrowing the search anyway. Passing in a wildcard pattern like `"%"` seemed to resolve the issue. Maybe I need to open an upstream bug against the Snowflake JDBC driver for our friends at Snowflake now.

Ultimately this had nothing to do with the MLv2 metadata refactor, that was a big ol' red herring. See https://metaboat.slack.com/archives/C04DN5VRQM6/p1706156558940489 if you want to read the play by play of my investigation of this or see more details

### DIFF
--- a/modules/drivers/snowflake/src/metabase/driver/snowflake.clj
+++ b/modules/drivers/snowflake/src/metabase/driver/snowflake.clj
@@ -341,8 +341,9 @@
   accept either. Throw an Exception if neither key can be found."
   {:arglists '([database])}
   [{details :details}]
-  (or (:db details)
-      (:dbname details)
+  ;; ignore any blank keys
+  (or (m/find-first (every-pred string? (complement str/blank?))
+                    ((juxt :db :dbname) details))
       (throw (Exception. (tru "Invalid Snowflake connection details: missing DB name.")))))
 
 (defn- query-db-name []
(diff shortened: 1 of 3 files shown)


modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -341,8 +341,9 @@
   accept either. Throw an Exception if neither key can be found."
   {:arglists '([database])}
   [{details :details}]
-  (or (:db details)
-      (:dbname details)
+  ;; ignore any blank keys
+  (or (m/find-first (every-pred string? (complement str/blank?))
+                    ((juxt :db :dbname) details))
```
camsaul, line 346: Not really needed for the fix, but while investigating the bug I realized that this wasn't ignoring blank strings like it should (just to be extra safe), so I fixed that here and added a test.

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -418,16 +419,31 @@
...
-           {:tables (->> (sql-jdbc.describe-database/db-tables driver (.getMetaData conn) nil db-name)
-                         (filter (fn [{schema :schema table-name :name}]
-                                   (and (not (contains? excluded-schemas schema))
-                                        (driver.s/include-schema? inclusion-patterns
-                                                                  exclusion-patterns
-                                                                  schema)
-                                        (sql-jdbc.sync/have-select-privilege? driver conn schema table-name))))
-                         (map #(dissoc % :type))
-                         set)}))))))
+           {:tables (into #{}
```
camsaul, line 433: The only thing I actually changed here was changing the schema pattern from `nil` to `"%"`, but I also reworked this to transduce rather than do things with lazy seqs for the slight performance boost and because the old sync code did stuff that way

modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj
```diff
@@ -166,22 +190,51 @@
...
+                  (driver/describe-database :snowflake database))))))))
+
 (deftest describe-database-views-test
   (mt/test-driver :snowflake
     (testing "describe-database views"
       (let [details (mt/dbdef->connection-details :snowflake :db {:database-name "views_test"})
+            db-name (:db details)
             spec    (sql-jdbc.conn/connection-details->spec :snowflake details)]
         ;; create the snowflake DB
-        (jdbc/execute! spec ["DROP DATABASE IF EXISTS \"views_test\";"]
```
camsaul, line 175: Also unrelated, I realized this test is dropping and creating the database `views_test` every time it runs and not prefixing it for the current test session at all... so if two runs were to run this same exact test at the same time it would cause one of them to flake. Maybe this hasn't happened yet, but I fixed it either way to be safe and to set a good example for other people writing tests in the future

bshepherdson (approved):

PR #38168, opened 2024-01-26
Use mlv2 findColumnIndexForColumnSetting in Table

uladzimirdev:


frontend/src/metabase-lib/queries/utils/dataset.ts
```diff
@@ -44,7 +44,7 @@ export function findColumnIndexForColumnSetting(
 
     if (!query) {
       // TODO: remove once migration is completed
-      throw new Error("query is required to find column index");
+      // throw new Error("query is required to find column index");
```
uladzimirdev, line 47: disable throwing temporary, so tests could show an error if any

frontend/src/metabase/visualizations/visualizations/Table.jsx
```diff
@@ -352,7 +354,11 @@ export default class Table extends Component {
             columnSetting.enabled || this.props.isShowingDetailsOnlyColumns,
         )
         .map(columnSetting =>
-          findColumnIndexForColumnSetting(cols, columnSetting),
+          findColumnIndexForColumnSetting(
+            cols,
+            columnSetting,
+            // question.query(),
```
uladzimirdev, line 360: if it still fails, then the order matters

kamilmielnik (approved):

ranquild (approved):

PR #38172, opened 2024-01-26
Use mlv2 findColumnIndexForColumnSetting in viz settings

uladzimirdev:
Pass query from viz settings (no metadata available). 

In the future there might be a better at BE for viz settings


frontend/src/metabase/visualizations/lib/settings/column.js
```diff
@@ -523,12 +524,22 @@ export const buildTableColumnSettings = ({
     widget: ChartSettingTableColumns,
     getHidden: (series, vizSettings) => vizSettings["table.pivot"],
     isValid: ([{ card, data }]) => {
+      const question = new Question(card /* metadata */);
```
uladzimirdev, line 527: the most problematic question, whether we need metadata here or not

frontend/src/metabase/visualizations/lib/settings/column.js
```diff
@@ -523,12 +524,22 @@ export const buildTableColumnSettings = ({
...
       const columns = card.visualization_settings["table.columns"];
       const enabledColumns = columns.filter(column => column.enabled);
+
+      let query;
+
+      try {
```
kamilmielnik, line 533: Leftover debugging code :)

frontend/src/metabase/visualizations/lib/settings/column.js
kamilmielnik: I think we do. I checked BE code and there was this call chain:

`find-column-indexes-from-legacy-refs` -> `find-column-indexes-for-refs` -> `find-matching-column` -> `resolve-field-id` (uses metadata, will throw if field is not found by id) -> `saved-question-metadata` (uses metadata)

uladzimirdev: I checked those methods as well, but I couldn't say it will 100% throw

frontend/src/metabase/visualizations/lib/settings/column.js
uladzimirdev: nope, I got usage of this method in Audit pages with internal query, so it would throw after our latest changes
https://github.com/metabase/metabase/commit/4acd8deb670b1102099c4340e2f135619e39bd5a

kamilmielnik (approved):
I'm not sure if this is the way to go but [I agree it's ok to merge it to an integration branch](https://metaboat.slack.com/archives/C0645JP1W81/p1707480911262879?thread_ts=1707479570.156889&cid=C0645JP1W81) and go from there.

PR #38175, opened 2024-01-26
 Replace anaphoric `with-source-and-dest-dbs` with hygienic alternative

### LINKED ISSUE #36103: Replace do-with-source-and-dest-dbs macro
We could remove some macrology in test code and replace `do-with-source-and-dest-dbs` with an extended `with-empty-h2-app-db`, that optionally binds the database of the created database to a var whose symbol is provided by the caller. Then you could write test code that switches between dbs using `(mt/with-db source-db ...)` or something like it

Context: https://metaboat.slack.com/archives/C0641E4PB9B/p1700834887052379?thread_ts=1700811376.911959&cid=C0641E4PB9B

crisptrutski:
Closes https://github.com/metabase/metabase/issues/36103

### Description

This change introduces a new `with-dbs` macro which is hygienic and supports an arbitrary number of databases.

For now I am leaving the old `with-source-and-dest-dbs` macro, so we can deal with the noise of upgrading the tests in a later PR. I upgrade a single test so that the usage is clear.

### Checklist

- [x] Finalize caller form of the macro
	 - ~~Can we make it work with Cursive?~~ Aha, resolve as `fn` works
- [x] Configure linting for this macro
- [x] Replace all usages and remove the old macro

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -419,6 +419,7 @@
   honeysql.helpers/defhelper                                                           clj-kondo.lint-as/def-catch-all
   honeysql.util/defalias                                                               clojure.core/def
   metabase-enterprise.serialization.test-util/with-random-dump-dir                     clojure.core/let
+  metabase-enterprise.serialization.test-util/with-dbs                                 clojure.core/fn
   metabase.actions.test-util/with-actions                                              clojure.core/let
   metabase.api.common/let-404                                                          clojure.core/let
   metabase.api.search-test/do-test-users                                               clojure.core/let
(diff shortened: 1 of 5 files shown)


crisptrutski:
* `master`
  * **PR #38175** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38175?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38175?utm_source=stack-comment).

sansolovyov (commented):
I like the minimal implementation of `with-db`, but it kind of feels like I'd love to use single macro when there is just a single db to use. What do you think?

crisptrutski:
> I like the minimal implementation of `with-db`, but it kind of feels like I'd love to use single macro when there is just a single db to use. What do you think?

I'm all for testing sugar. Unfortunately `with-db` is taken, otherwise we could use that with a symbol instead of a binding as first argument. I guess it's confusing that these two functions are named so similarly anyway...

crisptrutski:
@piranha On second thought, there's no reason to use this helper unless you want to switch between multiple databases. You only need the bindings if you need them to switch the current app context. For a single database you're just use `mt/with-empty-h2-app-db` (although it could use a snappier name).

calherries (approved):
Very nice

PR #38178, opened 2024-01-26
Forbid using InternalQuery with MLv2

uladzimirdev:
### Description

Using `InternalQuery` with `question.query()` led to problems, as `InternalQuery` is not supported by MLv2. 
From now `question.query()` will throw for `InternalQuery`

### How to verify

Tests should pass

### DIFF
--- a/enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditTable.jsx
+++ b/enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditTable.jsx
@@ -51,7 +51,7 @@ function AuditTable({
 
   const handleOnLoad = results => {
     setLoadedCount(results[0].row_count);
-    onLoad(results);
+    onLoad?.(results);
   };
 
   const card = chain(table.card)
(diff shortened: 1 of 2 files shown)


enterprise/frontend/src/metabase-enterprise/audit_app/containers/AuditTable.jsx
```diff
@@ -51,7 +51,7 @@ function AuditTable({
 
   const handleOnLoad = results => {
     setLoadedCount(results[0].row_count);
-    onLoad(results);
+    onLoad?.(results);
```
uladzimirdev, line 54: onLoad could be undefined, probably, not related, but I saw it during testing

kamilmielnik (approved):

frontend/src/metabase-lib/Question.ts
```diff
@@ -1045,6 +1045,10 @@ class Question {
   }
 
   query(metadata = this._metadata): Query {
+    if (this._legacyQuery() instanceof InternalQuery) {
+      throw new Error("Internal query is not supported by MLv2");
+    }
```
nemanjaglumac, line 1050: 🎉

nemanjaglumac (approved):
Thank you

PR #38198, opened 2024-01-26
Add path filtering for cljs workflow

### LINKED ISSUE #36221: Introduce path filtering for `cljs.yml` workflow
`cljs.yml` workflow currently runs always - for **any** change. We should trigger those tests on related changes only.

Whether we decide to use `file-paths.yml` or a [native GitHub path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), we shouldn't forget to include the workflow file itself in the list of triggers.

## Why?
This workflow ran more than 31K times so far with an average run time around ~4 minutes.

lbrdnk:
Closes  #36221

This PR makes CI use path filtering for `.cljs` and `.cljc` tests. Existing entries in `.github/file-paths.yaml` are reused and merged into `cljs` path.

### DIFF
--- a/.github/file-paths.yaml
+++ b/.github/file-paths.yaml
@@ -135,3 +135,4 @@ cljs:
   - *shared_sources
   - *shared_specs
   - ".github/workflows/cljs.yml"
+  - "shadow-cljs.edn"
--- a/.github/file-paths.yaml
+++ b/.github/file-paths.yaml
@@ -130,3 +130,8 @@ visualizations:
   - "frontend/src/metabase/visualizations/**"
   - "frontend/src/metabase/static-viz/**"
   - "src/metabase/pulse/**"
+
+cljs:
+  - *shared_sources
+  - *shared_specs
+  - ".github/workflows/cljs.yml"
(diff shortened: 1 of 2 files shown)


bshepherdson (approved):

.github/file-paths.yaml
```diff
@@ -130,3 +130,8 @@ visualizations:
   - "frontend/src/metabase/visualizations/**"
   - "frontend/src/metabase/static-viz/**"
   - "src/metabase/pulse/**"
+
+cljs:
+  - *shared_sources
+  - *shared_specs
+  - ".github/workflows/cljs.yml"
```
nemanjaglumac, line 137: 👌

nemanjaglumac: @bshepherdson I think we should include https://github.com/metabase/metabase/blob/master/shadow-cljs.edn in here somewhere.

Sources? Or directly in `cljs`?
Once that's solved, this PR looks good to go 🚀

bshepherdson (approved):

.github/file-paths.yaml
nemanjaglumac: Yeah, my reasoning was that one might alter the output path in that file, for example. And this workflow wouldn't run, and wouldn't give you feedback that it's broken. At least not until the first next time it runs.

PR #38201, opened 2024-01-26
 refactor: prepare SettingsPage to have conditionally rendered steps

npretto:
### Description

The upcoming project [Meet embedders at the door during setup](https://www.notion.so/metabase/Meet-embedders-at-the-door-during-setup-d291f12e5ba244ba8cf803af18f63e86) will need to render conditionally some steps, this PR makes it easy to do so while retaining the numeric labels correct.

Changes:
- The steps that had the numeric label are now rendered by a mapping from an array that passes the step number to them
- <DatabaseHelp/> is a fixed position card, so it doesn't matter where it's rendered by react
image

### How to verify

Nothing should fail in CI.
I added a fake toggle to remove one step to test it locally, see the demo

### Demo

https://github.com/metabase/metabase/assets/1914270/4056a433-d5af-4ef1-9b1c-b29b76870071

### Checklist

-> no behaviour change


WiNloSt (approved):
Looks good to me.

PR #38203, opened 2024-01-26
Dashboard required params

oleggromov:

## Description
We allow making parameters required on the dashboards, where a required parameter from now on needs a default value. Otherwise, we prevent the dashboard from being saved and show an explanatory tooltip near the save button.

The parameter widget behavior is the same as in the https://github.com/metabase/metabase/pull/37709

## How to check

https://github.com/metabase/metabase/assets/2196347/97680e59-5986-4066-a828-e32de726db59

Basically follow the steps shown in the screencast:

1. Open or create a dashboard with some cards and create connected parameters
2. Try making a parameter required and observe that the dashboard cannot be saved without a default value
3. Set a value and save it
4. Repeat for other parameters of different types
5. Once your dashboard is saved, check that parameter widgets work as intended
    - there's never an `x` to clear a required parameter
    - when the value is different from the default, we show the "reset" icon and allow reverting back to default
    - when you try to unset the value of a required filter, we revert it back to default

## How it works
- the `required` field for the parameters was partially supported and finally added here: https://github.com/metabase/metabase/pull/37800
- there are 2 new actions: `metabase/dashboard/SET_PARAMETER_VALUE_TO_DEFAULT` and `metabase/dashboard/SET_PARAMETER_REQUIRED` which are self-explanatory
- `DashboardHeader` is updated to use `getMissingRequiredParameters` selector, which returns parameters with `required` but without a `default` value — and then disables the Save button and shows a tooltip near it so it's clear why we cannot save
- `ParameterSettings` is updated to include the "Required" switch, identical to https://github.com/metabase/metabase/pull/37709
- `ParameterValueWidget` was also update in the PR mentioned above so the only required change was to add the `enableParameterRequiredBehavior` prop in the `Dashboard`, which is propagated down the component tree

## Tests
Testing e2e three parameter types: strings, numbers and dates. The rest behave pretty much in the same way.

### DIFF
--- a/e2e/support/helpers/e2e-dashboard-helpers.js
+++ b/e2e/support/helpers/e2e-dashboard-helpers.js
@@ -19,6 +19,10 @@ export function getDashboardCard(index = 0) {
   return getDashboardCards().eq(index);
 }
 
+export function ensureDashboardCardHasText(text, index = 0) {
+  cy.get(".Card").eq(index).findByText(text);
+}
+
 function getDashboardApiUrl(dashId) {
   return `/api/dashboard/${dashId}`;
 }
@@ -134,6 +138,10 @@ export function setFilter(type, subType) {
   });
 }
 
+export function toggleRequiredParameter() {
+  cy.findByLabelText("Always require a value").click();
+}
+
 export function createEmptyTextBox() {
   cy.findByLabelText("Edit dashboard").click();
   cy.findByLabelText("Add a heading or text box").click();
@@ -238,6 +246,10 @@ export const dashboardHeader = () => {
 export const dashboardGrid = () => {
   return cy.findByTestId("dashboard-grid");
 };
+
+export function dashboardSaveButton() {
+  return cy.findByRole("button", { name: "Save" });
+}
 
 /**
  * @param {Object=} option
--- a/e2e/support/helpers/e2e-dashboard-helpers.js
+++ b/e2e/support/helpers/e2e-dashboard-helpers.js
@@ -20,7 +20,7 @@ export function getDashboardCard(index = 0) {
 }
 
 export function ensureDashboardCardHasText(text, index = 0) {
-  cy.get(".Card").eq(index).findByText(text);
+  cy.get(".Card").eq(index).should("contain", text);
 }
 
 function getDashboardApiUrl(dashId) {
@@ -248,7 +248,7 @@ export const dashboardGrid = () => {
 };
 
 export function dashboardSaveButton() {
-  return cy.findByRole("button", { name: "Save" });
+  return cy.findByTestId("edit-bar").findByRole("button", { name: "Save" });
 }
 
 /**
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -30,7 +30,7 @@ export function getDefaultValuePopulatedParameters(
 
 // Needed because parameter values might be arrays
 // in which case order of elements isn't guaranteed
-export function isParameterValuesIdentical(a, b) {
+export function areParameterValuesIdentical(a, b) {
   return _.isEqual(
     Array.isArray(a) ? a.slice().sort() : a,
     Array.isArray(b) ? b.slice().sort() : b,
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -28,6 +28,15 @@ export function getDefaultValuePopulatedParameters(
   });
 }
 
+// Needed because parameter values might be arrays
+// in which case order of elements isn't guaranteed
+export function isParameterValuesIdentical(a, b) {
+  return _.isEqual(
+    Array.isArray(a) ? a.slice().sort() : a,
+    Array.isArray(b) ? b.slice().sort() : b,
+  );
+}
+
 export function normalizeParameter(parameter) {
   return {
     id: parameter.id,
--- a/frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
+++ b/frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
@@ -30,6 +30,7 @@ import TippyPopover from "metabase/components/Popover/TippyPopover";
 import { getPulseFormInput } from "metabase/pulse/selectors";
 import { fetchPulseFormInput } from "metabase/pulse/actions";
 import {
+  getCanSaveDashboard,
   getIsBookmarked,
   getIsShowDashboardInfoSidebar,
 } from "metabase/dashboard/selectors";
@@ -147,6 +148,7 @@ interface StateProps {
   isNavBarOpen: boolean;
   isShowingDashboardInfoSidebar: boolean;
   isHomepageDashboard: boolean;
+  canSaveDashboard: boolean;
 }
 
 interface DispatchProps {
@@ -176,6 +178,7 @@ const mapStateToProps = (state: State, props: OwnProps): StateProps => {
     isHomepageDashboard:
       getSetting(state, "custom-homepage") &&
       getSetting(state, "custom-homepage-dashboard") === props.dashboard?.id,
+    canSaveDashboard: getCanSaveDashboard(state),
   };
 };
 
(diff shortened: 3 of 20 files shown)


frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -299,15 +309,21 @@ class DashboardHeaderContainer extends Component<DashboardHeaderProps> {
       >
         {t`Cancel`}
       </Button>,
-      <ActionButton
+      <Tooltip
```
oleggromov, line 312: I tried using the Mantine component here but a) it doesn't work out of the box (it's either hidden or positioning if off) and b) we still use `Tooltip`s in the component anyway so it doesn't make a big difference. Will update it later, as planned in the epic.

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -628,6 +644,24 @@ class DashboardHeaderContainer extends Component<DashboardHeaderProps> {
...
+  }
+
+  const names = missingRequiredParams
+    .map(param => `"${param.name}"`)
+    .join(", ");
+
+  return ngettext(
+    msgid`The ${names} parameter requires a default value but none was provided.`,
+    `The ${names} parameters require default values but none were provided.`,
+    missingRequiredParams.length,
```
kamilmielnik, line 661: nit: messages explicitly depend on `names`, and only implicitly on `missingRequiredParams`

```suggestion
    names.length,
```

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -320,6 +320,19 @@ export const getParameters = createSelector(
...
+export const getRequiredParameters = createSelector(
+  [getParameters],
+  parameters => parameters.filter(param => param.required),
+);
+
+export const getMissingRequiredParameters = createSelector(
+  [getRequiredParameters],
+  requiredParameters =>
+    requiredParameters.filter(
+      param => !param.default || param?.default.length === 0,
```
kamilmielnik, line 332: 1. Redundant `?`
2. Let's make sure it's an array if we're inspecting `length`

```suggestion
      param => !param.default || (Array.isArray(param.default) && param.default.length === 0),
```

kamilmielnik (commented):
Works well :+1: 

Looking forward to tests.

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
oleggromov: This is not right because `names` is a concatenated string of the shape "param1, param2". What's funny is that I actually did that myself previously and then fixed that as a bug, because of course, the length of the string is always > 0.

frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.tsx
```diff
@@ -123,6 +125,13 @@ export const ParameterSidebar = ({
     [otherParameters],
   );
 
+  const handleChangeRequired = useCallback(
```
uladzimirdev, line 128: afaik ParametersSidebar will re-render when you select another parameter, so parameterId will change. 

because of that I would drop useCallback as unnecessary until you found unnecessary re-renderings without it

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -320,6 +320,19 @@ export const getParameters = createSelector(
   },
 );
 
+export const getRequiredParameters = createSelector(
```
uladzimirdev, line 323: I don't fully understand the idea of having two selectors and using only one. 

Is it some kind of memoization/optimization? I double there will be more than 20 dashboard parameters ever

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -291,6 +295,12 @@ class DashboardHeaderContainer extends Component<DashboardHeaderProps> {
   }
 
   getEditingButtons() {
+    const { missingRequiredParameters } = this.props;
```
uladzimirdev, line 298: now you desctructure variables from props 😄

uladzimirdev (commented):
Comments inline, overall LGTM

uladzimirdev (approved):

frontend/src/metabase/dashboard/selectors.ts
oleggromov: My idea was to have a separate selector to get all required parameters when needed, but you're right, I'll remove it as it's not needed.

e2e/support/helpers/e2e-dashboard-helpers.js
```diff
@@ -19,6 +19,10 @@ export function getDashboardCard(index = 0) {
   return getDashboardCards().eq(index);
 }
 
+export function ensureDashboardCardHasText(text, index = 0) {
+  cy.get(".Card").eq(index).findByText(text);
```
nemanjaglumac, line 23:
```suggestion
  cy.get(".Card").eq(index).should("contain", text);
```

A bit more readable, imo.

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
oleggromov: Yeah, when it's too long, it becomes worse to read `props.missingRequiredParameters`.
What I'd do instead is make a reference to `const p = this.props` and use `p.myParameter` but I believe one-letter variable names aren't favorable by the team.

nemanjaglumac (changes requested):

frontend/src/metabase/parameters/components/ParameterSidebar/ParameterSidebar.tsx
oleggromov: Thanks, removed. I actually looked at the other callbacks and just did the same for the sake of consistency. But you're right, it's unnecessary.

e2e/support/helpers/e2e-dashboard-helpers.js
```diff
@@ -219,6 +227,10 @@ export const dashboardGrid = () => {
   return cy.findByTestId("dashboard-grid");
 };
 
+export function dashboardSaveButton() {
+  return cy.findByRole("button", { name: "Save" });
```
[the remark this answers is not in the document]
oleggromov: Thanks, I'll fix the scoping.

frontend/src/metabase-lib/parameters/utils/parameter-values.js
```diff
@@ -28,6 +28,15 @@ export function getDefaultValuePopulatedParameters(
   });
 }
 
+// Needed because parameter values might be arrays
+// in which case order of elements isn't guaranteed
+export function isParameterValuesIdentical(a, b) {
```
nemanjaglumac, line 33: Nit: wouldn't it be more correct to call this `areParameterValuesIdentical`?

oleggromov: Yeah, I agree. I am just too used to calling booleans with `is`.

oleggromov:
@nemanjaglumac and @uladzimirdev please have a look again, addressed all your feedback.

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -358,7 +358,7 @@ class TagEditorParamInner extends Component<Props> {
           <ContainerLabel>
             {t`Default filter widget value`}
             {hasNoDefaultValue && tag.required && (
-              <ErrorSpan>{t`(required)`}</ErrorSpan>
+              <ErrorSpan>({t`required`})</ErrorSpan>
```
nemanjaglumac, line 361: Just a note that all strings so far have been translated with `()`. This will now change. I agree your proposal makes much more sense. Just drawing attention to the change, and the fact this will have to be updated in all translations.

Example:
```
#: frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx:248
#: frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx:283
#: frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx:325
msgid "(required)"
msgstr "(benötigt)"
```

nemanjaglumac (approved):
Works as advertised :)

Thanks for addressing the comments.
🚀

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
oleggromov: This is actually funny because in one of the previous PRs this was added after an opposite suggestion. So I changed it. And then this change was lost with a rebase :)

PR #33682, opened 2023-09-01
Migrate `metabase.sync*` to Malli

camsaul:

PR #38206, opened 2024-01-26
Exclude metabase.sync* and metabase.upload* from malli instrumentation

### LINKED ISSUE #38072: Sync and scan pegs the CPU to 100%
### Describe the bug

We upgraded from: 
v0.47.9 to v0.48.2

Previously we were happily running on a single core ECS instance. However now it is really struggling and we have to upgrade our ECS instance sizes.

Do you have any idea what might have caused this? And whether the new load is expected moving forward?

image

Screenshot 2024-01-24 at 10 21 26 am

### To Reproduce

Upgraded from v0.47.9 to v0.48.2

### Expected behavior

We wouldnt have expected this large a change in CPU utilization

### Logs

_No response_

```JSON
- Metabase: 0.48.3
- Metabase Postgres: 14.10
- Metabase Environment: Self Hosted (AWS Fargate)
- Data Sources: Postgres 12 => 14
```

### Severity

Medium

### Additional context

_No response_

calherries:
Hopefully fixes https://github.com/metabase/metabase/issues/38072.

This PR excludes `metabase.sync*` and `metabase.upload*` from `mu/defn` malli instrumentation in production.

I couldn't reproduce [the issue](https://github.com/metabase/metabase/issues/38072), but I noticed the issue is similar to https://github.com/metabase/metabase/pull/33682, where we found a performance regression was mostly fixed in another part of the codebase by disabling malli instrumentation in production.

We migrated `metabase.sync*` namespaces were migrated to use malli in 48: https://github.com/metabase/metabase/pull/33682

`metabase.upload*` is unrelated. I'm just familiar enough with this part of the code to also exclude it with this PR with confidence.

### DIFF
--- a/src/metabase/util/malli/fn.clj
+++ b/src/metabase/util/malli/fn.clj
@@ -301,7 +301,9 @@
   "Used to track namespaces to not enforce malli schemas on with `mu.fn/fn`."
   [namespace]
   (let [lib-and-middleware [#"^metabase\.lib\..*"
-                            #"^metabase\.query-processor\.middleware\..*"]
+                            #"^metabase\.query-processor\.middleware\..*"
+                            #"^metabase\.sync.*"
+                            #"^metabase\.upload.*"]
         matches?           (core/fn [namespace regexes]
                              (let [n (-> namespace ns-name str)]
                                (some #(re-matches % n) regexes)))


camsaul:
Is there some great reason not to disable `mu/defn` checks globally in production? As long as we still do the `defendpoint` ones which aren't done with `mu/defn` anyway I think we'll be ok... right?

calherries:
I don't know of a reason, but one reason is we could be using `mu/defn` exceptions for control flow somewhere that isn't covered by tests. I've raised the question [here](https://metaboat.slack.com/archives/CKZEMT1MJ/p1706518267990679?thread_ts=1704454622.640319&cid=CKZEMT1MJ), and it might be better to take the risk and remove it globally. How that resolves shouldn't block this PR from landing though.

sansolovyov (approved):

PR #38207, opened 2024-01-26
Remove frame-ancestors entirely if iframes allowed

tasos-boxlabs:
In Chromium-based browsers, the frame-ancestors check will fail when the HTML is served from file://. This means Metabase embeds cannot be used in Electron. To fix this, the frame-ancestors directive should be removed entirely when allowing iframes, rather than the catch-all *. This behaviour has been observed in other projects: https://github.com/orgs/community/discussions/13406

> [!IMPORTANT]
> If you are merging into master, please add either a `backport` or a `no-backport` label to this PR. You will not be able to merge until you do this step. Refer to the section [Do I need to backport this PR?](https://www.notion.so/metabase/Metabase-Branching-Strategy-6eb577d5f61142aa960a626d6bbdfeb3?pvs=4#89f80d6f17714a0198aeb66c0efd1b71) in the Metabase Branching Strategy document for more details.

> **Warning**
>
> If that is your first contribution to Metabase, please sign the [Contributor License Agreement](https://docs.google.com/a/metabase.com/forms/d/1oV38o7b9ONFSwuzwmERRMi9SYrhYeOrkbmNaq9pOJ_E/viewform) (unless it's a tiny documentation change). Also, if you're attempting to fix a translation issue, please submit your changes to our [POEditor project](https://poeditor.com/join/project/ynjQmwSsGh) instead of opening a PR.

### Description

The approach involves omitting frame-ancestor entirely and only appending it if allow-iframe is false.

### How to verify

1. Embed a dashboard into a simple HTML file 
2. Open it in a Chromium-based browser, without using an HTTP server.
3. The dashboard should be visible.


tasos-boxlabs:
Just following up on this, if this enables us to embed Metabase into our Electron apps it would be a huge win and could open up an entirely new client pool for Metabase as a company.

Issue #36430, opened 2023-12-05
[Epic] Enable saving questions directly to dashboards and make this the default path


luizarakaki:
**Links**
- product doc: [link](https://www.notion.so/metabase/Enable-saving-questions-directly-to-dashboards-and-make-this-the-default-path-a84be6e0540241189b251401fb35ca91?pvs=4)
- eng doc: _link to technical design doc, if any_
- feature branch: `branch-name` _this should be the feature branch where this work will be done in. PRs will be delivered against this branch_
- issue links: _related issues if any_

```[tasklist]
### BE Tasks
- [ ] https://github.com/metabase/metabase/issues/38205
- [ ] When we move dashboards, we also move their dependent cards
- [ ] When we archive dashboards, we also archive their dependent cards
- [ ] When we duplicate dashboards, we also correctly duplicate their dependent cards (aka change dashboard_id)
- [ ] Dashboard questions appear in search, but it returns the dashboard, not the question
- [ ] Hide dependent cards in collection endpoints
- [ ] Prevent the dashboard dependent card to be as a model
- [ ] Prevent the dashboard dependent card to be as a sandbox
- [ ] Prevent the dashboard dependent card to be as the starting point for another query (should not come back in the "Saved Questions" data sources option)
- [ ] Prevent the dashboard dependent card inside a native query with the {{#\1234}} syntax
```

camsaul:
I think you're missing some things here, for example we need code preventing you from using one of these Dashboard-dependent Cards:
   - as a Model
   - as a sandbox
   - as the starting point for another query (should not come back in the "Saved Questions" data sources option)
   - inside a native query with the `{{#1234}}` syntax

I think this should also be its own distinct `type` of saved question once #38074 lands

luizarakaki:
Closed in favor of [#47326](https://github.com/metabase/metabase/issues/47326)

PR #38216, opened 2024-01-27
Add dashboard_id to report_card and add to type column

### LINKED ISSUE #38205: Add migration of dashboard_id to report_card

qwef:
For https://github.com/metabase/metabase/issues/36430

Closes https://github.com/metabase/metabase/issues/38205

### Description

Migration for adding `dashboard_id` to `report_card`. If `dashboard_id` is populated, then that question is dependent to that dashboard.


resources/migrations/001_update_migrations.yaml
```diff
@@ -5228,6 +5228,26 @@ databaseChangeLog:
...
+      changes:
+        - addColumn:
+            tableName: report_card
+            columns:
+              - column:
+                  name: dashboard_id
+                  type: int
+                  remarks: "`dashboard_id` will be populated when this card is dependent on a dashboard"
+                  constraints:
+                    nullable: true
```
camsaul, line 5245: This should definitely be an actual FK. Right now looks like you can set this to `-1000` and no one will complain

camsaul (changes requested):
Honestly I'm not super convinced adding a `dashboard_id` column to `report_card` this is the best way to model the problem of having the ability for Cards to be tied to only one Dashboard, but if we were to go down that route I think this PR is still missing a lot of stuff:

- `dashboard_id` is a plain integer and not an actual foreign key, which means you could give it a value like `-1000`; deleting a Dashboard would leave an orphaned Card
- We need permissions checks that you're allowed to edit the Dashboard in question when you set a Card's `dashboard_id`
- We need to have an API-level check that disallows creating DashboardCards that would put one of these Dashboard-Restricted Cards in a different Dashboard. Right now, this column is ignored and nothing is stopping me from adding one of these Cards to a different Dashboard
- We need tests for all of this stuff 

There's also a lot more stuff missing from https://github.com/metabase/metabase/issues/36430 but I'll add more notes there.

qwef:
> Honestly I'm not super convinced adding a `dashboard_id` column to `report_card` this is the best way to model the problem of having the ability for Cards to be tied to only one Dashboard, but if we were to go down that route I think this PR is still missing a lot of stuff:

It does seem like this way can be reuse a lot of the existing code. The other way we thought of was when the `dashboard_id` column is populated, then `collection_id` will be nil. It seemed like this way could cause unforeseen problems (especially since `collection_id` = nil means root collection), and I thought it would be much easier if the dependent cards moved with the dashboard whenever the dashboard moves. 

If there's a better way, would love to hear it! Not super tied to the idea we have here.

PR #38221, opened 2024-01-29
[MLv2] [FE] Migrate `setDefaultDisplay` to MLv2

### LINKED ISSUE #37172: [MLv2] [FE] Migrate `setDefaultDisplay` to MLv2
Related BE tasks:
https://github.com/metabase/metabase/issues/37120
https://github.com/metabase/metabase/issues/36977
https://github.com/metabase/metabase/issues/36978
https://github.com/metabase/metabase/issues/38253

kamilmielnik:
Closes #37172

Depends on #38358
~Blocked from merging due to BE bug: #38253 - it causes `e2e-tests-binning-ee` CI check to fail.~
~But it's ready for review. I don't plan to add any more FE changes.~

### Description

- I'm not entirely sure if `Lib.defaultDisplay` should be a part of `Lib`. Suggestions where else to put it will be appreciated.
- `Lib.defaultDisplay` is meant to use as few MLv2 calls as possible


frontend/src/metabase-lib/display.ts
```diff
@@ -0,0 +1,160 @@
...
+    });
+    if (areBreakoutsCategories) {
+      return { display: "bar" };
+    }
+  }
+
+  return { display: "table" };
+};
+
+const getBreakoutColumnInfos = (
```
nemanjaglumac, line 123: Nit: I don't think "information" has a plural. I understand what you wanted to convey, but it just reads wrong. I'd go with `getBreakoutColumnInfo` or `getBreakoutAndColumnInfo` if you want to be really specific.

frontend/src/metabase-lib/display.ts
```diff
@@ -0,0 +1,160 @@
...
+import type {
+  CardDisplayType,
+  VisualizationSettings,
+} from "metabase-types/api";
+import { TYPE } from "metabase-lib/types/constants";
+import { isa } from "metabase-lib/types/utils/isa";
+
+type DefaultDisplay = {
+  display: CardDisplayType;
+  settings?: Partial<VisualizationSettings>;
```
nemanjaglumac, line 11: Is `map` the only display type that has settings?

nemanjaglumac:
> I'm not entirely sure if Lib.defaultDisplay should be a part of Lib. Suggestions where else to put it will be appreciated.

It feels right at home where it is. There can be a bit of confusion because we do have other `displayFoo` functions but I wouldn't worry too much about it at this moment. We can take inventory and consolidate this as the corpus of available functions grow.

Just my 2c.

frontend/src/metabase-lib/display.ts
kamilmielnik: AFAIK every display type can have settings, even "table" (e.g. conditional formatting)

frontend/src/metabase-lib/display.ts
kamilmielnik: Renamed to `getBreakoutsWithColumnInfo`: 4d78f22d9a5cf7973b4c5b74fd01c8f508b30abb

frontend/src/metabase-lib/display.ts
```diff
@@ -0,0 +1,168 @@
...
+    const [{ breakout, columnInfo }] = getBreakoutsWithColumnInfo(
+      query,
+      stageIndex,
+      breakouts,
+    );
+
+    if (isDate(columnInfo)) {
+      const info = Lib.displayInfo(query, stageIndex, breakout);
+
+      if (info.isTemporalExtraction) {
```
nemanjaglumac, line 71:
```suggestion
      if (info.isTemporalExtraction()) {
```

?

kamilmielnik: It's a boolean attribute, not a function.

frontend/src/metabase-lib/display.ts
```diff
@@ -0,0 +1,168 @@
...
+
+const getBreakoutsWithColumnInfo = (
+  query: Lib.Query,
+  stageIndex: number,
+  breakouts: Lib.BreakoutClause[],
+) => {
+  return breakouts.map(breakout => {
+    const column = Lib.breakoutColumn(query, stageIndex, breakout);
+    const columnInfo = Lib.displayInfo(query, stageIndex, column);
+    return { breakout, columnInfo };
```
uladzimirdev, line 139:
```suggestion
    return { breakout, column };
```

based on discussion in [slack](https://metaboat.slack.com/archives/C0645JP1W81/p1706864885372559?thread_ts=1706862717.614449&cid=C0645JP1W81), I believe we don't need to call displayInfo, but return a column itself instead, then check type of the column with methods available in `metabase-lib/column_types`

uladzimirdev:
re top level file/folder, I'd explicitly specify it's for `viz`, so probably it could to be `metabase-lib/viz/display.ts`

frontend/src/metabase-lib/display.ts
kamilmielnik: 7df3d78d24efba26a1a6f658572dad59ae299173

uladzimirdev (approved):
:shipit:

nemanjaglumac (approved):
Much more readable 🚀

PR #38232, opened 2024-01-29
Native query editor: Add an explanatory tooltip for the save button when required template tags miss defaults

oleggromov:
Part of the epic https://github.com/metabase/metabase/issues/36524

## Description 
Following the discussion and conclusion [on Slack](https://metaboat.slack.com/archives/C064QMXEV9N/p1706542715006239?thread_ts=1706276000.737189&cid=C064QMXEV9N), I am adding an explanatory tooltip for the disabled Save button.

https://github.com/metabase/metabase/assets/2196347/8de89193-3ea0-4ba2-ada0-a9a5f7c4e1ac

## How to test

1. Create a native query
2. Add 1 or 2 variables
3. Make either of them or both required but omit the default value
4. Observe the disabled save button (was there before) and the tooltip (added)

## Test
Updated existing e2e tests related to the save button being disabled when there are required parameters missing default values.

### DIFF
--- a/e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
+++ b/e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
@@ -53,6 +53,12 @@ describe("scenarios > filters > sql filters > field filter", () => {
       SQLFilter.getRunQueryButton().should("be.disabled");
       SQLFilter.getSaveQueryButton().should("have.attr", "disabled");
 
+      SQLFilter.getSaveQueryButton().realHover();
+      // eslint-disable-next-line no-unscoped-text-selectors -- tooltips are rendered in the body
+      cy.findByText(
+        'The "Filter" variable requires a default value but none was provided.',
+      );
+
       setDefaultFieldValue(4);
 
       SQLFilter.getRunQueryButton().should("not.be.disabled");
--- a/e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
+++ b/e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
@@ -54,11 +54,9 @@ describe("scenarios > filters > sql filters > field filter", () => {
       SQLFilter.getSaveQueryButton().should("have.attr", "disabled");
 
       SQLFilter.getSaveQueryButton().realHover();
-      cy.get("body").within(() => {
-        cy.findByText(
-          'The "Filter" variable requires a default value but none was provided.',
-        );
-      });
+      cy.get("body").findByText(
+        'The "Filter" variable requires a default value but none was provided.',
+      );
 
       setDefaultFieldValue(4);
 
--- a/e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
+++ b/e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
@@ -54,10 +54,11 @@ describe("scenarios > filters > sql filters > field filter", () => {
       SQLFilter.getSaveQueryButton().should("have.attr", "disabled");
 
       SQLFilter.getSaveQueryButton().realHover();
-      // eslint-disable-next-line no-unscoped-text-selectors -- tooltips are rendered in the body
-      cy.findByText(
-        'The "Filter" variable requires a default value but none was provided.',
-      );
+      cy.get("body").within(() => {
+        cy.findByText(
+          'The "Filter" variable requires a default value but none was provided.',
+        );
+      });
 
       setDefaultFieldValue(4);
 
--- a/e2e/test/scenarios/permissions/reproductions/22447-illogical-UI-elements-for-nodata.cy.spec.js
+++ b/e2e/test/scenarios/permissions/reproductions/22447-illogical-UI-elements-for-nodata.cy.spec.js
@@ -40,7 +40,7 @@ describe("UI elements that make no sense for users without data permissions (met
 
     cy.get("@saveButton").realHover();
     // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("You don't have permissions to save this question.");
+    cy.findByText("You don't have permission to save this question.");
 
     cy.findByTestId("qb-header-action-panel").within(() => {
       cy.icon("refresh").should("not.exist");
--- a/e2e/test/scenarios/permissions/reproductions/22447-illogical-UI-elements-for-nodata.cy.spec.js
+++ b/e2e/test/scenarios/permissions/reproductions/22447-illogical-UI-elements-for-nodata.cy.spec.js
@@ -40,7 +40,7 @@ describe("UI elements that make no sense for users without data permissions (met
 
     cy.get("@saveButton").realHover();
     // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("You don't have permission to save this question.");
+    cy.findByText("You don't have permissions to save this question.");
 
     cy.findByTestId("qb-header-action-panel").within(() => {
       cy.icon("refresh").should("not.exist");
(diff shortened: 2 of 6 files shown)


frontend/src/metabase/query_builder/selectors.js
```diff
@@ -994,3 +994,18 @@ export const getIsEditingInDashboard = state => {
 export const getDashboard = state => {
   return getDashboardById(state, getDashboardId(state));
 };
+
+export const getTemplateTags = createSelector([getCard], card => {
+  return getIn(card, ["dataset_query", "native", "template-tags"]);
+});
```
nemanjaglumac, line 1000: We already have a perfect MLv2 function for this.
`Lib.templateTags(query)` will give you exactly what you need.

In that case, can we not add another selector and use the MLv2 directly in ViewHeader instead?

This is the example for the following SQL:
`select * from PEOPLE where {{state}}` where `state` is a field filter mapped to people.state. It is required with the default value set to `CA`.

```json
{
  "state": {
    "widget-type": "contains",
    "default": [
      "CA"
    ],
    "name": "state",
    "type": "dimension",
    "id": "4200afba-5f26-4d38-bbcd-de6279c7d55e",
    "dimension": [
      "field",
      {
        "base-type": "Text",
        "effective-type": "Text",
        "uuid": "58e74d8d-df71-4485-86f2-c5b616010bb0"
      },
      34
    ],
    "display-name": "State",
    "options": {
      "case-sensitive": false
    },
    "required": true
  }
}
```

oleggromov: Actually to use MLv2 is not a bad idea, will do. However, this logic must stay within the `selectors` (or any other than rendering) layer for sure because ideally, our components should be almost purely presentational. Without that, we will never escape this neo-PHP realm where every single thing must be tested with an e2e test or what we call "unit tests".

nemanjaglumac: WFM

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
```diff
@@ -532,3 +547,19 @@ function ViewTitleHeaderRightSide(props) {
...
+
+  const names = requiredTemplateTags
+    .map(t => `"${t["display-name"] ?? t.name}"`)
+    .join(", ");
+
+  if (requiredTemplateTags.length > 1) {
+    return t`The ${names} variables require default values but none were provided.`;
+  }
+
+  return `The ${names} variable requires a default value but none was provided.`;
```
kamilmielnik, line 564: Please use `ngettext`

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
```diff
@@ -532,3 +547,19 @@ function ViewTitleHeaderRightSide(props) {
...
 
 ViewTitleHeader.propTypes = viewTitleHeaderPropTypes;
+
+function getMissingRequiredTemplateTagsTooltip(requiredTemplateTags = []) {
+  if (!requiredTemplateTags.length) {
+    return "";
+  }
+
+  const names = requiredTemplateTags
+    .map(t => `"${t["display-name"] ?? t.name}"`)
```
kamilmielnik, line 557: It took me a moment to realize the `t` is not from `ttag`.
Please use meaningful names :pray:

kamilmielnik (commented):
Please add tests :pray:

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
```diff
@@ -445,6 +449,17 @@ function ViewTitleHeaderRightSide(props) {
...
+  const missingValueRequiredTTags = requiredTemplateTags.filter(
+    t => t.required && !t.default,
+  );
+
+  const isSaveDisabled = !question.canRun() || !isEditable;
+  const disabledSaveTooltip = !question.canRun()
+    ? getMissingRequiredTemplateTagsTooltip(missingValueRequiredTTags)
+    : !isEditable
+    ? t`You don't have permissions to save this question.`
+    : "";
```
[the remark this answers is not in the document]
oleggromov: This is actually correct. In case both of them are `true`, we don't show a tooltip.
But I agree about the helper, will do.

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
```diff
@@ -532,3 +543,40 @@ function ViewTitleHeaderRightSide(props) {
...
 
 ViewTitleHeader.propTypes = viewTitleHeaderPropTypes;
+
+function getDisabledSaveTooltip(
+  question,
+  isEditable,
+  requiredTemplateTags = [],
+) {
+  if (!isEditable) {
+    return t`You don't have permissions to save this question.`;
```
nemanjaglumac, line 553: I don't want to block the PR because of this, but please note that this string doesn't exist in the translation files. It will not be translated automatically and will require all OSS translation contributors to add another translation.

Was it really that important to change?
I'd double check with the product before shipping this.

e2e/test/scenarios/permissions/reproductions/22447-illogical-UI-elements-for-nodata.cy.spec.js
```diff
@@ -40,7 +40,7 @@ describe("UI elements that make no sense for users without data permissions (met
 
     cy.get("@saveButton").realHover();
     // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("You don't have permission to save this question.");
+    cy.findByText("You don't have permissions to save this question.");
```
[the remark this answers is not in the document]
nemanjaglumac: Please see https://github.com/metabase/metabase/pull/38232/files#r1470995798

frontend/src/metabase/query_builder/components/view/ViewHeader.jsx
oleggromov: I understand that this translation doesn't exist yet — similar with the other new or updated strings.
Was it necessary or not, I don't know, in my opinion it's better to fix anything small like this. I am checking with the product about this though, thanks.

oleggromov: Actually I was wrong about it and `you don't have permission to` is grammatically correct. Thank you for noticing this!

oleggromov:
Hey @nemanjaglumac so `Lib.templateTags` is not a drop-in replacement unfortunately as it seems to be dropping the template tag attributes such as `required` somewhere in the process. I'll leave the code as is and get back to it later.

nemanjaglumac (approved):
Looks good.
Thanks for updating tests

e2e/test/scenarios/native-filters/sql-field-filter.cy.spec.js
```diff
@@ -53,6 +53,13 @@ describe("scenarios > filters > sql filters > field filter", () => {
       SQLFilter.getRunQueryButton().should("be.disabled");
       SQLFilter.getSaveQueryButton().should("have.attr", "disabled");
 
+      SQLFilter.getSaveQueryButton().realHover();
+      cy.get("body").within(() => {
```
oleggromov, line 57: I am keeping it `body` for now because I think doing `[data-tippy-root]` is too specific for the current implementation and e2e tests should rather avoid depending on the _3rd party lib_ implementation details.

nemanjaglumac: Hint: `cy.get("body").findByText("foo");` works just fine.

kamilmielnik: @oleggromov `cy.findByRole("tooltip")` should be used

PR #38238, opened 2024-01-29
Prototype — Dashboard sections

kulyk:


kulyk:
Implemented in #38463

PR #38239, opened 2024-01-29
Use "filled" bookmark icon when the item is bookmarked

### LINKED ISSUE #34327: Bookmark icon should be correct size; and should be filled when an item is bookmarked
### Describe the bug

I think this is my own doing as part of the big icons revamp in 0.47, but we should use the filled version of our bookmark icon when an item is bookmarked, not the outlined version that you see here:

image

Additionally, the size of the bookmark icon is too small when viewed on a dashboard, saved question, or model — it should have a viewbox of 20px by 20px (which would result in an effective pixel width and height of approximately 12x17px when scaled up).

### To Reproduce

1. Go to a dashboard or saved question
2. Bookmark it
3. See how it is a sad, hollow shell of its once glorious self, yearning to be filled to counter the yawning existential emptiness it feels within. (And it's also a bit too small relative to the other icons.)

### Expected behavior

It should be a filled icon when bookmarked, and should be the slightly larger size on the dashboard/question/model page, whether bookmarked or not.

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.11+9",
    "java.version": "11.0.11",
    "java.vm.version": "11.0.11+9",
    "os.name": "Mac OS X",
    "os.version": "11.5",
    "user.language": "en",
    "user.timezone": "America/Los_Angeles"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2022-11-18",
      "src_hash": "09f4ec50c4b6096192e18c140068ce8a7e67434d",
      "tag": "v0.45.1-SNAPSHOT",
      "branch": "maz-make-saving-a-question-less-fear-inducing",
      "hash": "f694966"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

P3

### Additional context

[Figma file with proposed solution](https://www.figma.com/file/brlp9N73Dok4aLU7AeHUms/bookmark-fix?type=design&node-id=0-1&mode=design&t=L9x6f1GeN5zZGFlh-11)

image

nemanjaglumac:
This PR resolves #34327 by increasing the icon size, and by changing the icon type to filled once it is selected.

Take a look at the original issue > Additional context to get a sense for how this should look.

These are the screenshots from this branch:
image
image

image
image

### DIFF
--- a/frontend/src/metabase/core/components/BookmarkToggle/BookmarkToggle.tsx
+++ b/frontend/src/metabase/core/components/BookmarkToggle/BookmarkToggle.tsx
@@ -37,6 +37,8 @@ const BookmarkToggle = forwardRef(function BookmarkToggle(
     setIsAnimating(false);
   }, []);
 
+  const iconName = isBookmarked ? "bookmark_filled" : "bookmark";
+
   return (
     <Tooltip
       tooltip={isBookmarked ? t`Remove from bookmarks` : t`Bookmark`}
@@ -49,7 +51,7 @@ const BookmarkToggle = forwardRef(function BookmarkToggle(
         onClick={handleClick}
       >
         <BookmarkIcon
-          name="bookmark"
+          name={iconName}
           isBookmarked={isBookmarked}
           isAnimating={isAnimating}
           onAnimationEnd={handleAnimationEnd}
--- a/frontend/src/metabase/nav/containers/MainNavbar/MainNavbarContainer/BookmarkList/BookmarkList.tsx
+++ b/frontend/src/metabase/nav/containers/MainNavbar/MainNavbarContainer/BookmarkList/BookmarkList.tsx
@@ -84,6 +84,8 @@ const BookmarkItem = ({
     bookmark.type === "collection" &&
     !PLUGIN_COLLECTIONS.isRegularCollection(bookmark);
 
+  const iconName = isSelected ? "bookmark_filled" : "bookmark";
+
   return (
     <Sortable id={bookmark.id} key={bookmark.id}>
       <SidebarBookmarkItem
@@ -97,7 +99,7 @@ const BookmarkItem = ({
         right={
           <button onClick={onRemove}>
             <Tooltip tooltip={t`Remove bookmark`} placement="bottom">
-              <Icon name="bookmark" />
+              <Icon name={iconName} />
             </Tooltip>
           </button>
         }
--- a/e2e/test/scenarios/organization/bookmarks-collection.cy.spec.js
+++ b/e2e/test/scenarios/organization/bookmarks-collection.cy.spec.js
@@ -54,18 +54,21 @@ describe("scenarios > organization > bookmarks > collection", () => {
     // Rename bookmarked collection
     cy.findByTestId("collection-name-heading").click().type(" 2").blur();
 
-    navigationSidebar().within(() => {
-      cy.findAllByText("First collection 2").should("have.length", 2);
-    });
+    navigationSidebar()
+      .findAllByText("First collection 2")
+      .should("have.length", 2);
 
     // Remove bookmark
-    cy.findByTestId("collection-menu").within(() => {
-      cy.icon("bookmark").click();
-    });
+    cy.findByTestId("collection-menu").icon("bookmark_filled").click();
 
-    navigationSidebar().within(() => {
-      cy.findAllByText("First collection 2").should("have.length", 1);
-    });
+    navigationSidebar()
+      .findAllByText("First collection 2")
+      .should("have.length", 1);
+
+    cy.findByTestId("collection-menu")
+      .icon("bookmark_filled")
+      .should("not.exist");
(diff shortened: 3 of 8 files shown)


mazameli (changes requested):
Thank you for doing this!

Looks like the icon isn't quite centered within this hover state though: 

image

frontend/src/metabase/core/components/BookmarkToggle/BookmarkToggle.tsx
```diff
@@ -37,6 +37,10 @@ const BookmarkToggle = forwardRef(function BookmarkToggle(
     setIsAnimating(false);
   }, []);
 
+  const getIcon = () => {
+    return isBookmarked ? "bookmark_filled" : "bookmark";
+  };
```
WiNloSt, line 42: I'd say it's more practical just to make this a variable i.e.

```suggestion
  const icon = isBookmarked ? "bookmark_filled" : "bookmark";
```

This is because functions are supposed to be called by others with some of these reasons
1. The function is called at a different time than its instantiation.
2. You expect the caller to pass some arguments to the function.
3. You don't want the value to be evaluated when the function is being declared e.g. you could reference a variable that's being instantiated after the declaration of a function. Because the function will be called after the variable is instantiated, it would have value. This wouldn't work if you evaluated the variable right away.

There could be more reasons, but these are some reasons that are perfectly valid for putting things in a function. Other than that, if both functions and variables work, I'd prefer the variables first. Like in this case, you could think of it like this. Why do I need a function when I'll immediately call it immediately below its declaration?

frontend/src/metabase/nav/containers/MainNavbar/MainNavbarContainer/BookmarkList/BookmarkList.tsx
```diff
@@ -84,6 +84,8 @@ const BookmarkItem = ({
     bookmark.type === "collection" &&
     !PLUGIN_COLLECTIONS.isRegularCollection(bookmark);
 
+  const getIcon = () => (isSelected ? "bookmark_filled" : "bookmark");
```
WiNloSt, line 87: Same as the previous comment. Let's make this a variable. Also please note the different style of the functions. The previous one is using `return` while this one is using a shorthand return. Mostly we'd want things to be consistent. It's doesn't matter in this case if we would change it to a variable anyway, but that's something to keep in mind.

nemanjaglumac:
> Two other things I noticed:
> 
> 1. My figma mock incorrectly assumed we were using `text-dark` for the action icons, but it looks like in prod we're using `text-medium`, so the bookmark one which is using `text-dark` in this branch now looks darker than the others.
> 2. Something else that I didn't call out is that we're using 20px icons for the action buttons on dashboards, but aren't for some reason on questions. I think we should be, which would then make the actions on saved questions match the size of the bookmark action. Here's the current state of things, with dashboard actions on the bottom, where you can see how they're larger.
> 
> image
> cc @kdoh just so he's aware of these changes.

@maz, now that you mentioned it, I found even more differences.

- dashboard icons use `text-medium`, question icons use `text-dark`
- dashboard icons' size is 20, question icons' size is 16

Which color should we use?

And something I've also just noticed - bookmark icon is the only one that doesn't change its color to `color-brand` on hover. Would you like me to fix that as well?

mazameli:
> dashboard icons use text-medium, question icons use text-dark
dashboard icons' size is 20, question icons' size is 16
Which color should we use?

@kdoh and I spoke, and let's go with 16px, `text-dark` for both dashboards and saved questions.

>And something I've also just noticed - bookmark icon is the only one that doesn't change its color to color-brand on hover. Would you like me to fix that as well?

Yes please! (I think we weren't doing that before, because before the only indication of whether an item was bookmarked or not was if it was blue or not.)

nemanjaglumac:
@mazameli Thank you for the input. I have addressed all your comments. Things look more unified now. Please give this one final look.

I have found some more inconsistencies that are not specific to the bookmark icon, but that's a broader topic I'm happy to chat about. That falls outside the scope of this particular PR, imo.

@WiNloSt addressed your comments as well. Please check once again.

WiNloSt (approved):
Code-wise. Looks good to me 👍

mazameli (approved):
Thank you!

nemanjaglumac:

Issue #36358, opened 2023-12-04
Default dashboards to a fixed width to give creators a more predictable canvas


cdeweyx:
## Context
[Product doc](https://www.notion.so/Default-dashboards-to-a-max-width-to-give-creators-a-more-predictable-canvas-0a4f5b03fcba42f59752a6c41c453ebb)
Epic: https://github.com/metabase/metabase/issues/36010
Non-blocking milestones of this feature: https://github.com/metabase/metabase/issues/38446

---
## Tasks

**Dashboards Needing Changes**
1. Full-App
2. Static Embedding (`PublicDashboard`)
3. X-Rays (`AutomaticDashboardApp`)

---
#### Milestone 1
**BE**
- https://github.com/metabase/metabase/pull/38247
  - [x] (BE) Create new `width` setting on dashboard object with 2 options: `"full" | "fixed"`
  - [x] (BE) Create migration to set all existing dashboards to `width = "full"`

**FE**
- https://github.com/metabase/metabase/pull/38377
  - [x] (FE) Create `width` setting toggle for changing between `"full"` and  `"fixed"`. This toggle should only be exposed in **edit mode**.
    - `"full"` means `On`
    - **(default option)** `"fixed"` means `Off`
  - [x] (FE) Dashboard grid scaling
    - [x] When `width = "full"`, dashboard grid should auto-scale to the full-width available
    - [x] When `width = "fixed"`, dashboard grid should remain at the fixed width until it is the size of the viewport and then should responsively resize until mobile view
    - [x] Both options should condense to one column when the normal mobile breakpoint is reached

---
#### Milestone 2

- https://github.com/metabase/metabase/pull/38450
  - [x] (FE) Collaborate with design to spec out new dash-card minimum and default sizes (because the new fixed-width may lead to the current defaults becoming nonsensical).

---
#### Milestone 3
- https://github.com/metabase/metabase/pull/38585
  - [x] Add fixed width behavior to dashboard filters
    - [x] Make sure dashboard sidebars still stay fixed to the edge of the viewport
- https://github.com/metabase/metabase/pull/38574
  - [x] Flip the order of tabs and filters in public dashboards

JesseSDevaney:
The rest of the work for this feature is now being tracked in a non-blocking issue
- https://github.com/metabase/metabase/issues/38446

PR #38247, opened 2024-01-30
Fixed width dashboard migration

adam-james-v:
This PR addressed the first 2 items in #36358

> - [x] (BE) Create new width setting on dashboard object with 2 options: "full" | "fixed"
> - [x] (BE) Create migration to set all existing dashboards to width = "full"

The need was:
 - add a :width setting on dashboards
 - the allowed values here are 'fixed' or 'full'
 - for all new dashboards, the default setting shall be 'fixed' (the new default behaviour we want)
 - for all existing dashboards, the value should be set to 'full' (the existing behaviour)

3 migrations were used to achieve this:
 - the migration adding the column, whose default value is 'fixed'. This ensures all new dashboards will have the value correctly set
 - another migration to change all existing dashboard entries to have the 'full' value. This makes sense as it won't break existing behaviours for users
 - the final migration setting the width column to notNull, ensuring it's always set going forward. 

A test was added to the Dashboard api test namespace to make sure these properties are met after the migration.

I've also manually tested in the REPL:

```clojure
;; an existing dashboard:
(:width (t2/select-one :model/Dashboard))
;; full.

;; a newly created dashboard:
(:width (t2/select-one :model/Dashboard :id 535))
;; fixed.

```

The setting is exposed to change via `PUT /api/dashboard/:id`.


resources/migrations/001_update_migrations.yaml
```diff
@@ -5228,6 +5228,45 @@ databaseChangeLog:
...
+            columns:
+              - column:
+                  name: width
+                  value: 'full'
+      rollback: []
+
+  - changeSet:
+      id: v49.2024-01-29T19:56:40
+      author: adam-james
+      comment: Set all existing report_dashboard width values to full.
```
markbastian, line 5261: Please update the comment and remarks to be accurate.

markbastian (changes requested):
Update the comment and remarks in the last migration and then it will be g2g.

resources/migrations/001_update_migrations.yaml
adam-james-v: Good catch, thanks. I've updated them.

markbastian (approved):
Looks good.

JesseSDevaney (approved):

Issue #37599, opened 2024-01-11
Error upgrading metabase 0.47.11 to 0.48.2


nuvme-devops:
### Describe the bug

Im trying to upgrade my self hosted metabase (docker) from v0.47.11 to v0.48.2, but the following migration fails: `Migration failed for changeset migrations/001_update_migrations.yaml`
with the following error:
`Reason: liquibase.exception.DatabaseException: (conn=99) Duplicate key name 'idx_query_execution_action_id' [Failed SQL: (1061) CREATE INDEX `idx_query_execution_action_id` ON `metabase`.`query_execution`(`action_id`)]
`

### To Reproduce

1. Run a metabase container v0.47.11 using mysql 8
2. Change the version to v0.48.2 and try running again
3. See error

### Information about your Metabase installation

```JSON
Ubuntu 22.04
Mysql 8.2.0
Metabase v0.48.2
Running in a EC2 instance
```

### Severity

Blocking an upgrade

noahmoss:
Likely related to https://github.com/metabase/metabase/issues/37552

qnkhuat:
@nuvme-devops I've followed your reproduction steps and can't reproduce this. Can you reproduce it with an empty DB?

darksciencebase:
@nuvme-devops have you been able to upgrade?

paoliniluis:
The reason why this happens is the following:
1) have 2 metabase servers running
2) both at the same time want to upgrade the app db
3) one makes the change, while the other tries to do it and finds that it will do a duplicate, so crashes
4) there should be a first one still running and doing the changes, but if for some reason the container engine kills the one that's still running, you'll hit this error forever and you'll not be able to upgrade

qnkhuat:
Could you give us a dump of `select * from databasechangelog`?

and are you running multi-metabase instances?

nuvme-devops:
I'm not running multi-metabase, just a single container.

There is the dump

Output of SELECT * FROM DATABASECHANGELOG:
[databasechangelog.txt](https://github.com/metabase/metabase/files/14097480/databasechangelog.txt)

Dump file (had to change the file ext):
[databasechangelog.csv](https://github.com/metabase/metabase/files/14097348/databasechangelog.csv)

qnkhuat:
Thanks for your info. 

For your case, please try dropping the `idx_query_execution_action_id` index and retry the upgrade.
make sure you backup your DB before doing so!

darksciencebase:
[...]

Issue #37344, opened 2024-01-08
Upgrade to 48.2 failed due to a migration error


perivamsi:
### Describe the bug

Upgrading from 47.7 to 48.2 failed with the following error

```
2024-01-06T21:08:44.822152+00:00 app[web.3]: 2024-01-06 21:08:44,817 ERROR metabase.core :: Metabase Initialization FAILED
2024-01-06T21:08:44.822159+00:00 app[web.3]: java.sql.BatchUpdateException: Batch entry 0 UPDATE databasechangelog SET FILENAME = CASE WHEN ID = 'v00.00-000' THEN 'migrations/001_update_migrations.yaml' WHEN ID < 'v45.00-001' THEN 'migrations/000_legacy_migrations.yaml' ELSE 'migrations/001_update_migrations.yaml' END was aborted: ERROR: duplicate key value violates unique constraint "idx_databasechangelog_id_author_filename"
2024-01-06T21:08:44.822159+00:00 app[web.3]: Detail: Key (id, author, filename)=(v00.00-000, qnkhuat, migrations/001_update_migrations.yaml) already exists. Call getNextException to see other errors in the batch.
```

### To Reproduce

Upgrade from 47.7 to 48.2

### Logs

[crash_metabase.log](https://github.com/metabase/metabase/files/13859083/crash_metabase.log)
[sql_logs.log](https://github.com/metabase/metabase/files/13859084/sql_logs.log)

### Information about your Metabase installation

```JSON
N/A
```

### Severity

P1

bunnywrote:
I run into the same issue by updating to v1.48.2. Downgrading to v1.47.11 was not possible, because of corrupted database.
Had to restore database from backup :(

```
{"log":"Caused by: org.mariadb.jdbc.internal.util.exceptions.MariaDbSqlException: Table 'metabase.data_migrations' doesn't exist","logtag":"F","time":"2024-01-09T17:20:56.270537319Z"}
{"log":"Caused by: java.sql.SQLSyntaxErrorException: (conn=65141) Table 'metabase.data_migrations' doesn't exist","logtag":"F","time":"2024-01-09T17:20:56.27053284Z"}
```

qnkhuat:
@bunnywrote I don't think it relates to this issue; that seems like a separate one. Could you create an issue?
It'd be easier for us to debug if you include the full log and your system info.
Thanks

qnkhuat:
[...]

PR #38257, opened 2024-01-30
Speculative fix for migration on multi instances

qnkhuat:
Speculative fixes for #37599 #37344
I could not reproduce both of those issues, but it seems to me that they have the same problem: liquibase tries to run a migration twice.

While debugging this, I noticed that if you have 2 metabase instances trying to run migration simultaneously, one will acquire the lock and do the migration, while the other will wait for a while to see if it can acquire the lock.

The strange thing is when the 1st instance finished migrating, for some reason, the second instance will still attempt to update`(.update liquibase)`. Even though liquibase will skip all the migrations on the 2nd instance, we still shouldn't trigger this.

It's clearer if you look at the log of the 2 instances I've included below.
log for the 1st instance that will do the migration
[1.txt](https://github.com/metabase/metabase/files/14097318/1.txt)
second instance that wait til the lock is released.
[2.txt](https://github.com/metabase/metabase/files/14097319/2.txt)

Notice the log of the 2nd instance.

```
2024-01-30 17:53:12,179 INFO db.liquibase :: Checking if Database has unrun migrations...
2024-01-30 17:53:12,418 INFO db.liquibase :: Database has unrun migrations. Checking if the migration lock is cleared...
2024-01-30 17:53:12,422 INFO db.liquibase :: Database has migration lock. Waiting for lock to be cleared...
2024-01-30 17:53:14,452 INFO db.liquibase :: Running 248 migrations ...
2024-01-30 17:53:14,452 INFO db.liquibase :: Unrun migration v00.00-000
2024-01-30 17:53:14,453 INFO db.liquibase :: Unrun migration v45.00-001
....
2024-01-30 17:53:14,467 INFO db.liquibase :: Unrun migration v49.00-059

UPDATE SUMMARY
Run:                          0
Previously run:             248
Filtered out:                17
-------------------------------
Total change sets:          265


FILTERED CHANGE SETS SUMMARY
DBMS mismatch:               17

```
It waited for the 1st instance to finish, but after that, it still thought that there were un-run migrations and tried to do the update again. Even though liquibase does not re-run any migration (indicated by run :0 in update summary), but we shouldn't be triggering that on the 2nd instance.

This PR will ensure we don't attempt migrating if the other instance finished it.

Comment on liquibase: I have 2 theories:
- Liquibase has a bug in how it checks unrun migrations internally; if that's true, then it might be the root cause of the 2 bugs in that some of the migrations do get re-run.  might relate: 
  - https://github.com/liquibase/liquibase/issues/5516
  - https://github.com/liquibase/liquibase/issues/5395
  - https://github.com/liquibase/liquibase/issues/5254
  - I also noticed that checksum has changed for different liquibase versions, for example the checksum for migration id '1' is `8:7182ca8e82947c24fa827d31f78b19aa` on 42, but it's `8:70d9a1021171b879cc55058f4d094a9c` on 47
- Migration operation is not atomic: a migration could be run but it does not get recorded into databasechangelog

again, I really can't reproduce those bugs, so this is just a speculation.

### How to reproduce
1. create an empty db
2. start 2 Metabase instances at the same time

### DIFF
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -130,6 +130,7 @@
   ;; Custom migrations use toucan2, so we need to make sure it uses the same connection with liquibase
   (let [f* (fn [liquibase]
              ;; trigger liquibase to create databasechangelog tables if needed
+             ;; we need to do this until https://github.com/liquibase/liquibase/issues/5537 is fixed
              (.checkLiquibaseTables liquibase false (.getDatabaseChangeLog liquibase) nil nil)
              (f liquibase))]
     (binding [t2.conn/*current-connectable* conn-or-data-source]
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -129,7 +129,7 @@
    f                   :- fn?]
   ;; Custom migrations use toucan2, so we need to make sure it uses the same connection with liquibase
   (let [f* (fn [^Liquibase liquibase]
-             ;; trigger liquibase to create databasechangelog tables if needed
+             ;; trigger creation of liquibase's databasechangelog tables if needed, without updating any checksums
              ;; we need to do this until https://github.com/liquibase/liquibase/issues/5537 is fixed
              (.checkLiquibaseTables liquibase false (.getDatabaseChangeLog liquibase) nil nil)
              (f liquibase))]
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -128,7 +128,7 @@
   [conn-or-data-source :- [:or (ms/InstanceOfClass java.sql.Connection) (ms/InstanceOfClass javax.sql.DataSource)]
    f                   :- fn?]
   ;; Custom migrations use toucan2, so we need to make sure it uses the same connection with liquibase
-  (let [f* (fn [liquibase]
+  (let [f* (fn [^Liquibase liquibase]
              ;; trigger liquibase to create databasechangelog tables if needed
              ;; we need to do this until https://github.com/liquibase/liquibase/issues/5537 is fixed
              (.checkLiquibaseTables liquibase false (.getDatabaseChangeLog liquibase) nil nil)
@@ -235,7 +235,7 @@
          (let [^Contexts contexts nil
                start-time         (System/currentTimeMillis)]
            (log/info (trs "Running {0} migrations ..." unrun-migrations-count))
-           (doseq [change to-run-migrations]
+           (doseq [^ChangeSet change to-run-migrations]
              (log/tracef "To run migration %s" (.getId change)))
            (.update liquibase contexts)
            (log/info (trs "Migration complete in {0}" (u/format-milliseconds (- (System/currentTimeMillis) start-time)))))
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -128,16 +128,20 @@
   [conn-or-data-source :- [:or (ms/InstanceOfClass java.sql.Connection) (ms/InstanceOfClass javax.sql.DataSource)]
    f                   :- fn?]
   ;; Custom migrations use toucan2, so we need to make sure it uses the same connection with liquibase
-  (binding [t2.conn/*current-connectable* conn-or-data-source]
-    (if (instance? java.sql.Connection conn-or-data-source)
-      (f (->> conn-or-data-source liquibase-connection database (liquibase conn-or-data-source)))
-      ;; closing the `LiquibaseConnection`/`Database` closes the parent JDBC `Connection`, so only use it in combination
-      ;; with `with-open` *if* we are opening a new JDBC `Connection` from a JDBC spec. If we're passed in a `Connection`,
-      ;; it's safe to assume the caller is managing its lifecycle.
-      (with-open [conn           (.getConnection ^javax.sql.DataSource conn-or-data-source)
-                  liquibase-conn (liquibase-connection conn)
-                  database       (database liquibase-conn)]
-        (f (liquibase conn database))))))
+  (let [f* (fn [liquibase]
+             ;; trigger liquibase to create databasechangelog tables if needed
+             (.checkLiquibaseTables liquibase false (.getDatabaseChangeLog liquibase) nil nil)
+             (f liquibase))]
+    (binding [t2.conn/*current-connectable* conn-or-data-source]
+      (if (instance? java.sql.Connection conn-or-data-source)
+        (f* (->> conn-or-data-source liquibase-connection database (liquibase conn-or-data-source)))
+        ;; closing the `LiquibaseConnection`/`Database` closes the parent JDBC `Connection`, so only use it in combination
+        ;; with `with-open` *if* we are opening a new JDBC `Connection` from a JDBC spec. If we're passed in a `Connection`,
+        ;; it's safe to assume the caller is managing its lifecycle.
+        (with-open [conn           (.getConnection ^javax.sql.DataSource conn-or-data-source)
+                    liquibase-conn (liquibase-connection conn)
+                    database       (database liquibase-conn)]
+          (f* (liquibase conn database)))))))
 
 (defmacro with-liquibase
   "Execute body with an instance of a `Liquibase` bound to `liquibase-binding`.
@@ -222,8 +226,8 @@
     (do
      (log/info (trs "Database has unrun migrations. Checking if migraton lock is cleared..."))
      (wait-for-migration-lock-to-be-cleared liquibase)
-    ;; while we were waiting for the lock, it was possible that another instance finished the migration(s), so make
-    ;; sure something still needs to be done...
+     ;; while we were waiting for the lock, it was possible that another instance finished the migration(s), so make
+     ;; sure something still needs to be done...
      (let [to-run-migrations      (unrun-migrations data-source)
            unrun-migrations-count (count to-run-migrations)]
        (if (pos? unrun-migrations-count)
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -17,6 +17,7 @@
   (:import
    (java.io StringWriter)
    (java.util List Map)
+   (javax.sql DataSource)
    (liquibase Contexts LabelExpression Liquibase RuntimeEnvironment Scope Scope$Attr Scope$ScopedRunner)
    (liquibase.change.custom CustomChangeWrapper)
    (liquibase.changelog ChangeLogIterator ChangeSet ChangeSet$ExecType)
@@ -170,8 +171,9 @@
   using them.
 
   (I'm not 100% sure whether `Liquibase.update()` still acquires locks if the database is already up-to-date)"
-  [^Liquibase liquibase]
-  (.listUnrunChangeSets liquibase nil (LabelExpression.)))
+  [^DataSource data-source]
+  (with-liquibase [liquibase (.getConnection data-source)]
+     (.listUnrunChangeSets liquibase nil (LabelExpression.))))
 
 (defn- migration-lock-exists?
   "Is a migration lock in place for `liquibase`?"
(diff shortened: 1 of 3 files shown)


src/metabase/db/liquibase.clj
```diff
@@ -197,31 +199,39 @@
...
     ;; sure something still needs to be done...
-     (let [unrun-migrations-count (count (unrun-migrations liquibase))]
+     (let [to-run-migrations      (unrun-migrations data-source)
+           unrun-migrations-count (count to-run-migrations)]
        (if (pos? unrun-migrations-count)
          (let [^Contexts contexts nil
                start-time         (System/currentTimeMillis)]
-           (log/info (trs "Migration lock is cleared. Running {0} migrations ..." unrun-migrations-count))
+           (log/info (trs "Running {0} migrations ..." unrun-migrations-count))
+           (doseq [change to-run-migrations]
```
qnkhuat, line 233: I'll remove this before merging, leaving it here for debugging purposes

dragonsahead (approved):
Functionally speaking, it works

src/metabase/db/liquibase.clj
```diff
@@ -127,16 +128,20 @@
...
-      ;; closing the `LiquibaseConnection`/`Database` closes the parent JDBC `Connection`, so only use it in combination
-      ;; with `with-open` *if* we are opening a new JDBC `Connection` from a JDBC spec. If we're passed in a `Connection`,
-      ;; it's safe to assume the caller is managing its lifecycle.
-      (with-open [conn           (.getConnection ^javax.sql.DataSource conn-or-data-source)
-                  liquibase-conn (liquibase-connection conn)
-                  database       (database liquibase-conn)]
-        (f (liquibase conn database))))))
+  (let [f* (fn [liquibase]
+             ;; trigger liquibase to create databasechangelog tables if needed
+             (.checkLiquibaseTables liquibase false (.getDatabaseChangeLog liquibase) nil nil)
```
qnkhuat, line 133: this is a work around for this bug https://github.com/liquibase/liquibase/issues/5537

test/metabase/db/liquibase_test.clj
```diff
@@ -50,15 +40,6 @@
        (liquibase/with-liquibase [liquibase (->> (mt/dbdef->connection-details :mysql :db {:database-name "liquibase_test"})
                                                  (sql-jdbc.conn/connection-details->spec :mysql)
                                                  mdb.test-util/->ClojureJDBCSpecDataSource)]
-         (testing "Make sure the first line actually matches the shape we're testing against"
```
qnkhuat, line 53: we now create the `databasechangelog` table during `with-liquibase` call, so it's no longer the 1st migration.

Also we have the next test that checks charset already, so I don't think we really need this test

src/metabase/db/liquibase.clj
```diff
@@ -169,9 +174,10 @@
   skip creating and releasing migration locks, which is both slightly dangerous and a waste of time when we won't be
   using them.
 
-  (I'm not 100% sure whether `Liquibase.update()` still acquires locks if the database is already up-to-date)"
-  [^Liquibase liquibase]
-  (.listUnrunChangeSets liquibase nil (LabelExpression.)))
+  IMPORTANT: this function takes `data-source` but not `liquibase` because `.listUnrunChangeSets` is buggy. See #38257."
+  [^DataSource data-source]
+  (with-liquibase [liquibase (.getConnection data-source)]
```
qnkhuat, line 179: I've tried to use `connection` instead of `data-source`, but it also doesn't fix the issue :)

src/metabase/db/liquibase.clj
```diff
@@ -197,31 +203,39 @@
...
 (defn migrate-up-if-needed!
   "Run any unrun `liquibase` migrations, if needed."
-  [^Liquibase liquibase]
+  [^Liquibase liquibase ^DataSource data-source]
   (log/info (trs "Checking if Database has unrun migrations..."))
-  (if (seq (unrun-migrations liquibase))
+  (if (seq (unrun-migrations data-source))
     (do
-     (log/info (trs "Database has unrun migrations. Waiting for migration lock to be cleared..."))
+     (log/info (trs "Database has unrun migrations. Checking if migraton lock is cleared..."))
```
crisptrutski, line 228: Out of scope nit on language - taking about a lock being created or cleared is a quirk in my opinion, speaking to the liquibase internals. I'd rather say things like "checking if the lock is taken".

src/metabase/db/liquibase.clj
```diff
@@ -197,31 +203,39 @@
...
      (wait-for-migration-lock-to-be-cleared liquibase)
```
crisptrutski, line 229: A note on lock safety - which should probably block on taking the lock immediately after we see it has cleared. This might be a factor in the race conditions we're seeing.

src/metabase/db/liquibase.clj
crisptrutski: What about leaving it as a `trace` statement? Do we have a good mechanism for dynamically toggling log levels per namespace?

crisptrutski (commented):
Know that this is still very much WIP, but left a few comments anyway as I'm following along.

src/metabase/db/liquibase.clj
qnkhuat: > without updating any checksums

wdym by this?

qnkhuat: > I think it's even worth extracting to ensure-liquibase-tables-exist function to encapsulate the misleadingly named checkLiquibaseTables method and its noisy unused arguments.

I'm in favor of extracting functions for reusing purposes; if for readability, I think the comment already does its justice here.

I also think we should keep the right number of "helpers" functions, cause if someone come in and read this for the first time, it's easy to get loss between helpers function vs main function that capture the business logic.

src/metabase/db/liquibase.clj
```diff
@@ -127,16 +128,20 @@
...
-      ;; it's safe to assume the caller is managing its lifecycle.
-      (with-open [conn           (.getConnection ^javax.sql.DataSource conn-or-data-source)
-                  liquibase-conn (liquibase-connection conn)
-                  database       (database liquibase-conn)]
-        (f (liquibase conn database))))))
+  (let [f* (fn [liquibase]
+             ;; trigger liquibase to create databasechangelog tables if needed
+             (.checkLiquibaseTables liquibase false (.getDatabaseChangeLog liquibase) nil nil)
+             (f liquibase))]
+    (binding [t2.conn/*current-connectable* conn-or-data-source]
```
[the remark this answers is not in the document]
qnkhuat: to be fair, this function only has many lines; what it does is pretty basic. I'm not even sure if we really need 2 branches here, I think it was added for testing purposes.

on macro, I don't think we should put the logic there, our style of writing macro is that it eventually calls `do-something` do the job. the defmacro basically is wrapper of this function.

So I'd rather keep all the logic in this `do-`function rather than scatter it in both places.

Wdyt?

src/metabase/db/liquibase.clj
```diff
@@ -169,9 +174,10 @@
...
 
-  (I'm not 100% sure whether `Liquibase.update()` still acquires locks if the database is already up-to-date)"
-  [^Liquibase liquibase]
-  (.listUnrunChangeSets liquibase nil (LabelExpression.)))
+  IMPORTANT: this function takes `data-source` but not `liquibase` because `.listUnrunChangeSets` is buggy. See #38257."
+  [^DataSource data-source]
+  (with-liquibase [liquibase (.getConnection data-source)]
+    (.listUnrunChangeSets liquibase nil (LabelExpression.))))
 
 (defn- migration-lock-exists?
```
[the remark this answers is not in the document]
qnkhuat: yea, from what I see we use `!` strictly for functions that have side-effect, we'll enforce this even further in tests see https://github.com/metabase/metabase/pull/37126

I'm open to idea for convention of this kind tho.

src/metabase/db/liquibase.clj
qnkhuat: I don't follow what you mean, can you expand on that?

src/metabase/db/liquibase.clj
qnkhuat: yea trace is a good call.

> Do we have a good mechanism for dynamically toggling log levels per namespace?

you can use `metabase.test/set-ns-log-level!` in the REPL, or update `resources/log4j2.xml` if you use a jar

src/metabase/db/liquibase.clj
```diff
@@ -197,31 +203,39 @@
...
-       (LockException.
-        (str
-         (trs "Database has migration lock; cannot run migrations.")
-         " "
-         (trs "You can force-release these locks by running `java -jar metabase.jar migrate release-locks`.")))))))
+  (if (migration-lock-exists? liquibase)
+    (do
+     (log/info "Database has migration lock. Waiting for lock to be cleared...")
+     (Thread/sleep 2000)
+     (u/auto-retry 4
```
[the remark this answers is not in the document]
qnkhuat: went with volatile in 4cf508c09a

src/metabase/db/liquibase.clj
qnkhuat: makes sense! 4cf508c09a

qnkhuat:
@crisptrutski I've resolved all of your reviews, Could you take a look again?

src/metabase/db/liquibase.clj
crisptrutski: Typically in concurrent code you would use an atomic operation to block on taking the lock yourself, rather than just waiting for it to be released. With the current behavior there's nothing to stop another process or server taking the lock after this thread resumes.

crisptrutski: I'm just pointing this out for now as it's independent of your change, but I'll probably send you a PR to fix it. Even if it's not a problem in practice, it's good to follow best practices with concurrent code rather than relying on global reasoning that make not hold in future.

src/metabase/db/liquibase.clj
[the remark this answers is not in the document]
crisptrutski: A pure function will always return a value, so a function that can throw is impure. Whether you say that this is a side effect is I guess debatable, but it is a "dangerous" function as it can fail.

Since throwing and catching exceptions is [very expensive on the JVM](https://wiki.c2.com/?CheckDontCatch), typically considered bad form for control flow, I like to indicate that kind of effect in the name somehow. Perhaps the bang is a poor choice, because it can mean a lot of other things, most commonly writing to the database, especially if it's not within a transaction. 

Since this codebase frequently uses exceptions for control flow and consists largely of impure code, perhaps its premature to introduce a new convention.

I still think we should incorporate `wait` somehow in the name to mitigate surprise. How about `wait-if-migration-lock-held`? A bit clumsy still, but `wait-for-migration-lock` would imply the lock is also taken.

src/metabase/db/liquibase.clj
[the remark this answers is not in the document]
crisptrutski: Cool - happy to drop this for now. Still have misgivings on this code, and would like to come back an improve this at a later stage. Here are some thoughts.

> To be fair, this function only has many lines; what it does is pretty basic.

Precisely what it does is so basic, I think the code should be as simple as possible. Creating a thunk over a thunk and not executing from top to bottom is complexity that future code readers could be spared in my opinion.

> I'm not even sure if we really need 2 branches here, I think it was added for testing purposes.

1. Test code quality matters too. If anything the bar for being easy to follow is higher.
2. This code is not in a test namespace.
3. Future production code may make use of this distinction.

> On macro, I don't think we should put the logic there.

That's a fair point, I do really like the "always delegate to a function" convention (when it's possible). On the other hand, being able to insert the the side-effect within regular lexical code is much more straightforward than a thunk decorator, and ultimately yields code that's easier to follow in my opinion. 

The "always delegate" convention exists to serve readability, and should be followed so rigorously that it conflicts IMHO. I'll sleep on it in this case, still calibrating my sensibilities.

src/metabase/db/liquibase.clj
qnkhuat: ah! I look forward to that PR of your, thank you for pointing that out!

src/metabase/db/liquibase.clj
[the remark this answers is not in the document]
qnkhuat: which function we're talking about again? I don't think `migration-lock-exists?` blocks right? 

or are you talking about `wait-for-migration-lock-to-be-cleared` which already prefix with `wait`?

> https://wiki.c2.com/?CheckDontCatch

good to know!

src/metabase/db/liquibase.clj
crisptrutski: > Without updating any checksums

I'm referencing to the `updateExistingNullChecksums` parameter that you set to false, and which makes the last two arguments irrelevant and thus ni.

> I'm in favor of extracting functions for reusing purposes; if for readability, the comment already does its justice here.

In my opinion comments should be used sparingly to explain *why* things are done, and only seldomly for "what" and even less the "how". The latter kinds of comments are [typically](https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/) code smells suggesting code could be more readable. 

In this instance I think the "what" comment is a crutch, and the better approach is to introduce a good name in the form of a helper function.

> I also think we should keep the right number of "helpers" functions.

From a bunch of similar discussions with you and Cal, I'm starting to suspect we have different opinions on "the right number" and code readability. Here's a wall of text about my take on code structure. 

This is just meant food for thought, ultimately a consistent and cohesive approach trumps a holy crusade for the one true style  ⛪ 

1. A key claim of FP is that breaking complex things down to the composition of "simple parts" makes reasoning easier, and yields code that is more robust to changing requirements.
2. [Simple things don't mix concerns, they do "one thing", and have "on reason to change".](https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html)
3. [Simple things don't mix levels of abstraction](http://principles-wiki.net/principles:single_level_of_abstraction). Jumping between abstraction levels is mentally taxing.
4. Removing duplication is [not a goal in itself](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction), it just  

> If someone comes in and reads this for the first time, it's easy to get lost between the helpers function ...

I think this is a culturally relative opinion. To be a bit snarky: I'm the new guy here, and giving you "more objective" feedback on what would have make the code easier to follow. 😄 

To maybe add some heft to my cheekiness - I was the custodian for a very large, old and non-trivial codebase over a long period of time, with multiple large teams working on it, and plenty of engineer turnover. My opinions reflect what seemed to work in my experience to tame this complexity and made code more approachable and flexible to change. YMMV.

Personally, I find it much easier to jump between simple functions than to comprehend large procedures. 

At the end of the day Clojure runs *a lot* of layers down until your reach the turtles, and all this code in particular is calling down to a pretty convoluted Java package. Clarity comes from having the right abstractions, not from to surfacing every detail. Ultimately I want to minimize the time taken to drill down to the code of interest, and minimize the amount of code read on the way there, rather then reading things from the bottom up in big procedures.

> vs the main function that captures the business logic.

I think part of the problems is that we're often trying to pack the business logic into these "main" functions, versus a "function core" approach where logic is kept separate to imperative, orchestrating code.

src/metabase/db/liquibase.clj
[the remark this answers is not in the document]
crisptrutski: > which function we're talking about again?

Wow, such a long thread for me not noticing a fold between the name of one function and the implementation of another one 😆

crisptrutski (approved):
Thanks for the tweaks 😄

src/metabase/db/liquibase.clj
qnkhuat: There is a lot to chew on here, thank you for writing down your reasoning. I'll add those resources to my to-read list for my weekend mornings :).

for now, I'll update the comment as you suggested and leave the call as is.

PR #38261, opened 2024-01-30
Browse models: Add banner

rafpaf:
Adds a dismissible banner to the Browse Data > Models page.


frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -56,15 +65,36 @@ export const BrowseModels = ({
...
+            <BannerModelIcon>
+              <Icon name="model" />
+            </BannerModelIcon>
+            <Text size="md" lh="1rem" mr="1rem">
+              Models help curate data to make it easier to find answers to
+              questions all in one place.
+            </Text>
+            <BannerCloseButton
+              onClick={() => {
+                updateSetting({ key: "dismissed-model-banner", value: true });
```
npfitz: This action needs to be dispatched. You'll either need to use the redux connect api or preferably the `useDispatch` hook. By doing this you also won't need forceUpdate

npfitz: It's also generally good practice to throw things like these inside of useCallbacks. Doesn't matter much here, but it can in other situations

frontend/src/metabase/browse/components/BrowseModels.styled.tsx
```diff
@@ -82,3 +83,23 @@ export const CollectionHeaderGroup = styled(Group)`
   position: relative;
   top: 0.5rem;
 `;
+
+export const Banner = styled.div`
```
npfitz, line 87: I think a lot of these styled components can mostly be done with Mantine components instead. `Banner` could be a `<Flex>` with some style props added, and we could grow the text so that it pushes the close button to the right.

npfitz (changes requested):
A few small adjustments, but nothing major!

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -56,15 +65,40 @@ export const BrowseModels = ({
...
-        ))}
-      </GridContainer>
+      <>
+        {shouldShowBanner && (
+          <Banner role="complementary">
+            <BannerModelIcon>
+              <Icon name="model" />
+            </BannerModelIcon>
+            <Text size="md" lh="1rem" mr="1rem">
+              Models help curate data to make it easier to find answers to
```
kdoh, line 75:
```suggestion
              Models curate data to make it easier to find answers to
```

Don't know if we need "help" in here seeing it now live. Makes it a little cleaner I think.

Also need to make sure this is translatable.

kdoh (changes requested):
Tiny little copy tweak and note on translation, but overall feels good.

frontend/src/metabase/browse/components/BrowseModels.tsx
rafpaf: @jeff-bruemmer Could you take a quick look at this sentence?

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -56,15 +64,51 @@ export const BrowseModels = ({
...
+            radius="0.25rem"
+            role="complementary"
+          >
+            <Flex>
+              <BannerModelIcon>
+                <Icon name="model" />
+              </BannerModelIcon>
+              <Group grow>
+                <Text size="md" lh="1rem" mr="1rem">
+                  {t`Models help curate data to make it easier to find answers to questions all in one place.`}
```
jeff-bruemmer, line 84:
```suggestion
                  {t`Create models to clean up raw data, join tables with related info, and make the data easier to query.`}
```

@kdoh what do you think about this edit? "Curate" is doing a lot of heavy lifting, so I wanted to spell out what we mean. Given how abstract models are, the elaboration might be worth it.

Also: this edit subtly implies that 1) _you_ can create models and 2) that _other people you work with_ are creating these models for their team(s). The original construction could make it seem like models just happen (or come from somewhere else, like raw tables), or that Metabase is curating them, a la X-rays or something.

As an aside, I'm not a fan of UI, like banners, that make me do work I didn't seek out, even if that work is a single click, one time. Do we really need to communicate this info about Models in a banner? Is this flash of context worth the risk of annoying people? Could this text be relegated to a tooltip? Or a subtitle beneath the tab?

e2e/test/scenarios/onboarding/home/browse.cy.spec.js
```diff
@@ -40,4 +40,24 @@ describe("scenarios > browse data", () => {
     cy.findByRole("heading", { name: "Products" }).click();
     cy.findByRole("gridcell", { name: "Rustic Paper Wallet" });
   });
+  it("can see and dismiss the modal explainer banner", () => {
+    cy.visit("/");
+    cy.findByRole("listitem", { name: "Browse data" }).click();
+    const banner = cy
```
npfitz, line 46: nit: This could probably just be a selector with a click action, rather than building all the variables. But it's also not that important.

frontend/src/metabase/browse/components/BrowseModels.styled.tsx
```diff
@@ -8,6 +8,7 @@ import Card from "metabase/components/Card";
 import { Ellipsified } from "metabase/core/components/Ellipsified";
 import Link from "metabase/core/components/Link";
 import { Group } from "metabase/ui";
+import IconButtonWrapper from "metabase/components/IconButtonWrapper";
```
npfitz, line 11: Do we want to be using the IconButtonWrapper? I see we're wrapping the Model icon inside a button, but I don't think it's clickable is it?

npfitz (approved):
None of my comments are worthy of blocking RC, but if you think you have a minute to take a look I think it would be good. 

Nice work!

frontend/src/metabase/browse/components/BrowseModels.styled.tsx
rafpaf: Good catch, fixed

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -43,6 +48,20 @@ export const BrowseModels = ({
     model => !isInstanceAnalyticsCollection(model.collection),
   );
   const groupsOfModels = groupModels(modelsFiltered, localeCode);
+  const hasDismissedBanner = useSelector(getHasDismissedBrowseModelsBanner);
```
iethree, line 51: nbd, but you could probably just use `getSetting` here

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -56,15 +75,43 @@ export const BrowseModels = ({
...
+              <BannerCloseButton
+                onClick={() => {
+                  dismissBanner();
+                }}
+              >
+                <Icon name="close" />
+              </BannerCloseButton>
+            </Flex>
+          </Paper>
+        )}
```
iethree, line 103: this would be a lot cleaner if it was its own component, it could probably keep track of its own state too.

naatz-co (approved):

frontend/src/metabase/browse/components/BrowseModels.tsx
rafpaf: Its state has style implications for the grid, unfortunately, so I wouldn't be able to move the state into a smaller component

iethree: I think I successfully handled this with a bit of negative margin. take a look:

[`78cb0ba` (#38635)](https://github.com/metabase/metabase/pull/38635/commits/78cb0bada41c1f878a55e4373896c43e046912e8#r1484872753)

Issue #38233, opened 2024-01-29
[Epic] Meet embedders at the door - minimal edition


npretto:
**Links**
- [product doc](https://www.notion.so/metabase/Meet-embedders-at-the-door-during-setup-d291f12e5ba244ba8cf803af18f63e86)
- [eng doc](https://www.notion.so/metabase/Meet-embedders-at-the-door-v49-d60a1242b7754a1cb9a3bcdae36fafe9)
- [testing plan](https://www.notion.so/Testing-plan-for-v49-d0e68bb6c94f4ff7b37860cd5b7faa35)

This Epic is only for the reduced version that we're aiming to deliver in v49:
- Usage question during the setup
- Hiding the database setup step for people who are interested in embedding
- [Very minimal homepage](https://www.figma.com/file/UFay4YjWyUMyl23T1cXzeC/Meet-first-time-embedders-at-the-door-with-a-dashboard?type=design&node-id=1449-15934&mode=design&t=U9k8MkZGkGEEO4BF-4) with just a couple of links

***Milestone 1*** - [`meet-embedders-v1-ms1`](https://github.com/metabase/metabase/pull/38520)

```[tasklist]
### Milestone 1
- [ ] https://github.com/metabase/metabase/pull/38201
- [ ] https://github.com/metabase/metabase/pull/38267
- [ ] https://github.com/metabase/metabase/pull/38390
- [ ] https://github.com/metabase/metabase/pull/38402
- [ ] https://github.com/metabase/metabase/pull/38465
- [x] Design/Product check
- [ ] https://github.com/metabase/metabase/pull/38521
```

npretto:
I extracted some tasks to https://github.com/metabase/metabase/issues/38556 to make it mergeable today before RC1

PR #38267, opened 2024-01-30
Usage reason question

npretto:
Part of [[Epic] Meet embedders at the door - minimal edition](https://github.com/metabase/metabase/issues/38233)

[design](https://www.figma.com/file/UFay4YjWyUMyl23T1cXzeC/Meet-first-time-embedders-at-the-door-with-a-dashboard?type=design&node-id=1264-20907&mode=design&t=0Ldr0em6rN8wruC5-4) (the card padding is not the same, I just re-used the component from the page to be consistent with what we currently have in prod)

### Description

This only adds the question and makes e2e pass.
The logic to hide the 'Add your data' will be added in another task, same thing for the analytics.
I'll also add tests with another PR after I implement all the logic, the setup is not tested on jest so it will take some time to do the setup and I first wanna show product/design a working thing with the whole flow.

### How to verify

There should be a new step in the setup, the answer has no effect on anything, nothing should be broken

### Demo

image


frontend/src/metabase/setup/actions.ts
```diff
@@ -88,6 +88,13 @@ export const submitUser = createAsyncThunk(
   },
 );
 
+export const submitUsageReason = createAsyncThunk(
+  "metabase/setup/SUBMIT_USAGE_REASON",
+  (_: UsageReason) => {
+    // TODO: add tracking
```
npretto, line 94: tracking is a later task in the epic

frontend/src/metabase/setup/components/UsageQuestionStep/UsageQuestionStep.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <Radio
+            value="embedding"
+            label={t`Embedding analytics into my application`}
+          />
+          <Radio value="both" label={t`A bit of both`} />
+          <Radio value="not-sure" label={t`Not sure yet`} />
+        </Stack>
+      </Radio.Group>
+      <Divider my="xl" />
+      <Button primary onClick={handleSubmit}>
```
npretto, line 92: This is a deprecated button, but I used to be consistent with the rest of the page that uses it

frontend/src/metabase/setup/constants.ts
```diff
@@ -3,8 +3,9 @@ export const LOCALE_TIMEOUT = 300;
 export const WELCOME_STEP = 0;
 export const LANGUAGE_STEP = 1;
 export const USER_STEP = 2;
-export const DATABASE_STEP = 3;
-export const PREFERENCES_STEP = 5; // backward compatibility for analytics
+export const USAGE_STEP = 3;
+export const DATABASE_STEP = 4;
+export const PREFERENCES_STEP = 5;
```
npretto, line 8: This is a breaking change in the analytics for some questions/queries, I have an ongoing discussion on slack and I will address it with a later task

frontend/src/metabase/setup/components/UsageQuestionStep/UsageQuestionStep.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const handleStepSelect = () => {
+    dispatch(selectStep(USAGE_STEP));
+  };
+
+  const handleSubmit = () => {
+    dispatch(submitUsageReason(usageReason));
+  };
+
+  const handleCHange = (value: UsageReason) => {
```
WiNloSt, line 45: 🤫 
```suggestion
  const handleChange = (value: UsageReason) => {
```

frontend/src/metabase/setup/constants.ts
npretto: Yes sorry: https://metaboat.slack.com/archives/C01MC2CKW2G/p1706719539434809

frontend/src/metabase-types/api/setup.ts
```diff
@@ -12,3 +12,9 @@ export interface SetupCheckListTask {
   triggered: boolean;
   is_next_step: boolean;
 }
+
+export type UsageReason =
```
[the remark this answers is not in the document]
npretto: Technically in the future this will be sent to the BE, should it be there when that will happen?

frontend/src/metabase/setup/reducers.ts
```diff
@@ -49,6 +51,10 @@ export const reducer = createReducer(initialState, builder => {
   });
   builder.addCase(submitUser.pending, (state, { meta }) => {
     state.user = meta.arg;
+    state.step = USAGE_STEP;
```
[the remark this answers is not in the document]
npretto: Yeah I don't like how the navigation is done either.
I was thinking of doing some refactor when I'll conditionally hide the "Add your data" step in the next PR, as in that scenario the "next" step will somewhat be dynamic.
I think we could make a selector that calculates the next step and use standardize how we go to the next step throughout the setup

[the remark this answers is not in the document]
WiNloSt: I think putting all of the navigation inside the selector would be the way because seems we rely on some async actions' results.

frontend/src/metabase-types/api/setup.ts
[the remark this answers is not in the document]
WiNloSt: It's hard to say. Most types under `/api` are used for both the request and the response. In our case, this would only be for the request if I'm not mistaken. I'd say it's fine to stay here then. Because it'd be weird for the request type to stay somewhere else.

WiNloSt (approved):
Looks good to me 👍

PR #38271, opened 2024-01-30
Keeps preview across different tabs

WiNloSt:
Part of a milestone 6 of
Epic: https://github.com/metabase/metabase/issues/35961

### Description

Make sure embed preview doesn't reload when changing between tabs/preview toggle/appearance settings

Please note that when going to the overview tab, the preview will be unmounted, so going to the preview again after visiting the overview tab will cause the iframe to reload. This is a compromise for the current codebase. I need to put a lot more work on the overview tab to make that happen, so it didn't happen.

### How to verify
Tests both dashboards and questions the same way.
1. Open static embed modal for a dashboard/question
2. Go to Parameters tab.
3. Click preview
4. Go to Appearance tab. The preview shouldn't rerender.
5. Modify settings in the Appearance tab, the preview shouldn't rerender.

### Demo

https://www.loom.com/share/f700c55d53634f5b974bba30f4f55867

### Checklist


frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/PreviewPane.styled.tsx
```diff
@@ -1,9 +1,20 @@
...
+  isTransparent?: boolean;
+  hidden?: boolean;
+}>`
   width: 100%;
   min-height: 17.5rem;
 
+  ${({ hidden }) =>
+    hidden &&
+    css`
+      visibility: hidden;
```
WiNloSt, line 14: Can't use `display: none` because it will cause the iframe to be reloaded again. So I need the iframe to stay in the dom. I set `position: absolute` so it doesn't take up space inside the modal when switching to other tabs.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -107,19 +113,17 @@ export const StaticEmbedSetupPane = ({
     embeddingParams,
   });
 
-  const iframeUrl = useMemo(
+  const iframeUrlWithoutHash = useMemo(
```
WiNloSt, line 116: I need the URL to stay consistent when only the appearance option is changed. Otherwise, it will cause the iframe to rerender.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -174,48 +183,90 @@ export const StaticEmbedSetupPane = ({
...
-          <AppearanceSettings
-            activePane={activePane}
-            resourceType={resourceType}
-            iframeUrl={iframeUrl}
-            displayOptions={displayOptions}
-            serverEmbedCodeSlot={getServerEmbedCodePane(TABS.Appearance)}
-            onChangePane={setActivePane}
-            onChangeDisplayOptions={setDisplayOptions}
-          />
+        <Tabs.Panel value={activeTab}>
```
WiNloSt, line 199: This is a bit hacky, but essentially, I need the preview to stay in the same component, otherwise, it will unmount and remount when changing between tabs because we have 3 parents (`Tabs.Panel`) previously. Now it's reduced to just 1.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -174,48 +183,90 @@ export const StaticEmbedSetupPane = ({
...
+          {activeTab === TABS.Overview ? (
```
WiNloSt, line 200: Ternary isn't the best looking thing here, but I need every element to be in the same bracket, otherwise the iframe preview will rerender.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -174,48 +183,90 @@ export const StaticEmbedSetupPane = ({
...
+            <OverviewSettings
```
WiNloSt, line 201: I wish I could keep the preview state even when navigating to the Overview tab, but it's a bit hard to extract the content inside it here, so we only keep the embedding state when switching between Parameters and Appearance tabs.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -174,48 +183,90 @@ export const StaticEmbedSetupPane = ({
...
+                    setParameterValues(state => ({
+                      ...state,
+                      [id]: value,
+                    }))
+                  }
+                />
+              }
+              previewSlot={
+                <>
+                  <PreviewModeSelector
```
WiNloSt, line 226: This is moved from both `ParametersSettings` and `AppearanceSettings`, the structures are similar that's why React doesn't unmount the preview when changing between these 2 tabs.

frontend/src/metabase/public/lib/embed.ts
```diff
@@ -29,12 +28,11 @@ function getSignedToken(
   });
 }
 
-export function getSignedPreviewUrl(
+export function getSignedPreviewUrlWithoutHash(
```
WiNloSt, line 31: This is only used in `frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx` so it's safe to modify the return value shape. I removed the hash part because I want the URL part before the hash to stay the same when only the hash is changed.

With the previous implementation, even when the hash is changed, the whole URL will be changed too.

e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
```diff
@@ -423,6 +426,94 @@ describe("scenarios > embedding > dashboard parameters with defaults", () => {
...
+      cy.findByLabelText("Transparent").click({ force: true });
+      getIframeBody()
+        .findByTestId("embed-frame")
+        .should("have.class", "Theme--transparent");
+      cy.get("@embedPreviewSpy").should("have.callCount", 1);
+
+      cy.log("Assert dashboard title");
+      getIframeBody().findByText(dashboardDetails.name).should("exist");
+      // We're getting an input element which is 0x0 in size
+      cy.findByLabelText("Dashboard title").click({ force: true });
```
npretto, line 488: Instead of finding the element with that label, can we click the label?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
npretto: It took a bit to understand why this is working lol, even with the context of the pr and your comments, should we add a comment? Otherwise it could be very easy to think "why are they doing this??" when reading the code :D

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
npretto: Isn't this only used in `iframeUrl` which also uses the hash params?

npretto (approved):
Added some comments but looks good to me

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
WiNloSt: Yeah, good point, the comment should go into the code.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
WiNloSt: yeah, this variable is used down below `const iframeUrl = iframeUrlWithoutHash + optionsToHashParams(displayOptions);`

Basically, I need to base URL to stay consistent, so I separate the calculation of base URL and hash. So if only the hash changes, the base URL would stay the same.

e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
WiNloSt: That's a good point. Do you have an example where we click the labels? I looked up testing-library, and they could only get the input using various matchers e.g. `byPlaceholderText`, `byLabelText`, but to find labels I could use `findByText` which is less specific, or `findByRole` with a lot of hoops. I'd rather use `findByLabelText` with `force: true` because I think it's still more readable.

What do you think?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
WiNloSt: Added a comment in https://github.com/metabase/metabase/pull/38271/commits/2b8172c10d0c10cf15f6aa25665b869f508ac84d

e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js
```diff
@@ -423,6 +426,112 @@ describe("scenarios > embedding > dashboard parameters with defaults", () => {
...
+
+  it("should not rerender the static embed preview unnecessarily (metabase#38271)", () => {
+    const textFilter = {
+      id: "3",
+      name: "Text filter",
+      slug: "filter-text",
+      type: "string/contains",
+    };
+
+    const dashboardDetails = {
```
WiNloSt, line 444: @npretto I added this filter, since I found out today that having a parameter will cause the appearance page preview to rerender every time we change any settings

e2e/test/scenarios/embedding/reproductions/37914-embed-preview-parameter-values-not-working.cy.spec.js
```diff
@@ -52,7 +53,9 @@ const dashboardDetails = {
 
 describe("issue 37914", () => {
   beforeEach(() => {
-    cy.intercept("GET", "/api/embed/dashboard/**").as("getEmbed");
+    cy.intercept("GET", "/api/preview_embed/dashboard/**").as(
```
WiNloSt, line 56: The preview didn't seem to work in Cypress before. I've fixed that in `frontend/src/metabase/lib/dom.js`

e2e/test/scenarios/sharing/reproductions/26988-embedding-dynamic-settings.cy.spec.js
```diff
@@ -40,7 +42,7 @@ describeEE("issue 26988", () => {
       acceptTerms: false,
     });
 
-    cy.wait("@dashboard");
+    cy.wait("@previewDashboard");
```
WiNloSt, line 45: The preview didn't seem to work in Cypress before. I've fixed that in `frontend/src/metabase/lib/dom.js`

frontend/src/metabase/lib/dom.js
```diff
@@ -26,7 +26,7 @@ window.METABASE = true;
 // used for detecting if we're previewing an embed
 export const IFRAMED_IN_SELF = (function () {
   try {
-    return window.self !== window.top && window.top.METABASE;
+    return window.self !== window.parent && window.parent.METABASE;
```
WiNloSt, line 29: `IFRAMED_IN_SELF` was only used in a few places. And all of them relate to how we're rendering the embed preview.

This was failing in Cypress because there are 3 frames in Cypress
- Cypress
    - Metabase Iframe
        - Embed preview Iframe

But in our application, there are only 2 frames
- Metabase
    - Embed preview Iframe
 
For this particular code, when running in Cypress `window.parent` will get Cypress frame, so the function will return `false`, which was wrong.

Changing it to `window.parent`, makes sure it returns the immediate parent frame which will be Metabase frame in both cases.

WiNloSt:
@npretto Can you help review the last 3 commits https://github.com/metabase/metabase/pull/38271/files/fe2a43912ed59d37bfbe77ecad5d9433042dfb1b..88ce7d07921a3c1c4bab7be362aac29bdd62570d?

I've fixed one more scenario and modified the test to capture that.

npretto (approved):

albertoperdomo (approved):
LGTM

PR #38274, opened 2024-01-30
CSV appends: Do not allow numbers with fractional component to be uploaded to an integer column

calherries:
Previously for CSV appends if you uploaded a CSV with a number with a fractional component, like `1.1`, the number would be rounded down to the closest integer before being inserted into the DB. That's a correctness issue and we definitely don't want to do that. 

With this PR, we'll throw an error in this scenario with a basic error message like `'1.1' is not an integer`


qnkhuat (approved):
WFM

PR #38276, opened 2024-01-30
CSV uploads: Expand valid date formats to include one digit days

calherries:
Previously we did not parse date strings as dates in a CSV upload if the day has one digit, e.g. "Jan 1 2020". We only accepted date formats where the day was two digits, like "Jan 01 2020". Similarly for other date formats, e.g. "1 Jan 2020".

This PR expands the set of valid date formats to include formats with one or two digits, e.g. both "Jan 1 2020" and "Jan 01 2020" are parsed as dates.


test/metabase/upload_test.clj
```diff
@@ -169,17 +169,23 @@
            [" 2022-01-01 "                    #t "2022-01-01"             date-type]
            [" 2022-02-30 "                    " 2022-02-30 "              vchar-type]
            [" -2022-01-01 "                   #t "-2022-01-01"            date-type]
+           [" Jan 1 2018"                     #t "2018-01-01"             date-type]
```
piranha, line 172: How about adding a test for `Jan 01 2018`?

test/metabase/upload_test.clj
```diff
@@ -169,17 +169,23 @@
...
            [" Jan 30 2018"                    #t "2018-01-30"             date-type]
            [" Jan 30 -2018"                   #t "-2018-01-30"            date-type]
            [" Jan 30, 2018"                   #t "2018-01-30"             date-type]
            [" Feb 30, 2018"                   " Feb 30, 2018"             vchar-type]
+           [" 1 Jan 2018"                     #t "2018-01-01"             date-type]
            [" 30 Jan 2018"                    #t "2018-01-30"             date-type]
            [" 30 Jan, 2018"                   #t "2018-01-30"             date-type]
+           [" January 1 2018"                 #t "2018-01-01"             date-type]
```
qnkhuat, line 180: we need a test for case where the day has a leading zero, like `January 01 2018`

qnkhuat (approved):
The changes LGTM, but let's add a test case for day that have leading zero.

test/metabase/upload_test.clj
calherries: Good point, covered here [f4bc5df](https://github.com/metabase/metabase/pull/38276/commits/f4bc5df1e621c469cda12553e51d9c234998a5ec)

test/metabase/upload_test.clj
calherries: Good point, covered here [f4bc5df](https://github.com/metabase/metabase/pull/38276/commits/f4bc5df1e621c469cda12553e51d9c234998a5ec)

PR #38289, opened 2024-01-30
Apply performance tweaks during dashboard editing

uladzimirdev:
> [!IMPORTANT]
> If you are merging into master, please add either a `backport` or a `no-backport` label to this PR. You will not be able to merge until you do this step. Refer to the section [Do I need to backport this PR?](https://www.notion.so/metabase/Metabase-Branching-Strategy-6eb577d5f61142aa960a626d6bbdfeb3?pvs=4#89f80d6f17714a0198aeb66c0efd1b71) in the Metabase Branching Strategy document for more details.

> **Warning**
>
> If that is your first contribution to Metabase, please sign the [Contributor License Agreement](https://docs.google.com/a/metabase.com/forms/d/1oV38o7b9ONFSwuzwmERRMi9SYrhYeOrkbmNaq9pOJ_E/viewform) (unless it's a tiny documentation change). Also, if you're attempting to fix a translation issue, please submit your changes to our [POEditor project](https://poeditor.com/join/project/ynjQmwSsGh) instead of opening a PR.

Closes https://github.com/metabase/metabase/issues/[issue_number]

### Description

Describe the overall approach and the problem being solved.

### How to verify

1. New question -> Sample Dataset -> ...
2. ...

### Demo

### Checklist


frontend/src/metabase/parameters/utils/dashboards.ts
```diff
@@ -136,7 +138,11 @@ function buildFieldFilterUiParameter(
   const mappingsForParameter = mappings.filter(
     mapping => mapping.parameter_id === parameter.id,
   );
-  const mappedFields = mappingsForParameter.map(mapping => {
+  const uniqueMappingsForParameters = _.uniq(mappingsForParameter, mapping =>
+    JSON.stringify(mapping.target),
```
oleggromov, line 142: An array-bound function using serialization per item immediately raises a red flag. It might become painfully slow at some point, probably even now. 

I don't fully understand the mapping semantics but would picking a bunch of ids and hashing them be enough to determine uniqueness? Calculating unique keys, hashing them and comparing numbers must be faster than serializing entire objects and comparing resulting strings.

Let's say we take a parameter:
```
{
    "parameter_id": "-1558815535",
    "target": [
        "dimension",
        [
            "field",
            11,
            {
                "source-field": 21
            }
        ]
    ],
    "card_id": "G__140569",
    "dashcard_id": "G__140589",
    "card": {}
}
```

Then we pick the fields into an array in a deterministic order, `[-1558815535, "G__140569", "G__140589"]`, then take an `md5` or any other hash of it or a string made of it.

WDYT?

uladzimirdev: good point.

From what I understood, `mappingsForParameter` array is used for generating some objects based on the `target` field, that's why we had a lot of duplicated results, but further calculations are 100x slower that's why we want to reduce the size of this array.

I don't understand the logic fully as well, so I might be wrong here, but that idea was suggested by @kulyk and I believe him 😄

oleggromov: We found that my suggestion won't fix anything as the most intensive calculations are happening elsewhere.

kulyk: A bit of context on why we're doing this: dashcards have `parameter_mappings` which associate a dashcard with many dashboard filters. For example, let's say we have a dashboard that looks roughly like that:

```js
{
  name: "Dashboard",
  parameters: [{
     id: "param-1",
     name: "Date",
     type: "date/all-options"
  }],
  dashcards: [{
     id: 1,
     card_id: 1,
     parameter_mappings: [{
       // if we've added other series to this dashcard (`series` prop),
       // it's important to set `card_id`
       card_id: 1,

       parameter_id: "param-1",

       // Let's pretend it's a reference to a "Created At" field
       target: [ "dimension", [ "field", 10, null ] ]
     }]
  }]
}
``` 

A parameter mapping object here defines how the daschard data is connected to the dashboard filter. As a result, once we change the filter value, we'll add a filter by the `[ "field", 10, null ]` field (specified in `parameter_mappings[0].target`)

Now imagine a dashboard with several filters and a few dozen of dashcards. They often would have queries to same tables, and will have very similar parameter mappings (imagine we have a datetime filter, there'll be a large chunk of dashcards connecting the same "CREATED_AT" column or something). In that case, `buildFieldFilterUiParameter` will loop over dozens of identical parameter mappings, instantiate a bunch of nearly identical MLV2 queries, etc. It doesn't make any sense, so we can just de-dupe this list upfront 

Instantiaing a MLv1 question is cheap, so we didn't bother about optimizing that step. But once we started constructing MLv2 queries inside this loop, the slowness started to stack up

frontend/src/metabase/parameters/utils/dashboards.ts
```diff
@@ -157,24 +163,21 @@ function buildFieldFilterUiParameter(
...
+      },
+    )
+    .map(({ field, shouldResolveFkField }) => {
+      return shouldResolveFkField ? field.target ?? field : field;
+    });
 
   return {
     ...parameter,
-    fields: uniqueFields,
+    fields: _.uniq(fields, field => field.id),
```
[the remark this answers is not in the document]
uladzimirdev: nope, unit test caught it (and raised memory issue instead of failure as it tried to print an instance of Field 😿 )

frontend/src/metabase/parameters/utils/dashboards.ts
uladzimirdev: > We need to de-duplicate `mappingsForParameter`, so we only have objects with unique `target` values. IMO, we should ideally have a MLv2 function comparing these targets. There's also [`compareMappingOptionTargets` function](https://github.com/metabase/metabase/blob/4ac58f9ded42c405bc43e9ca098d98116fa43337/frontend/src/metabase-lib/parameters/utils/targets.ts#L71), but I'm not sure if it's correct (it's comparing fields by IDs).

I used MLv2 version to compare option targets and with currently available MLv2 methods it's even slower

frontend/src/metabase/dashboard/actions/auto-wire-parameters/utils.ts
```diff
@@ -56,21 +56,24 @@ export function getMatchingParameterOption(
...
-      compareMappingOptionTargets(
+    parameterMappingOptions.find((param: { target: ParameterTarget }) => {
+      return compareMappingOptionTargets(
         targetDimension,
         param.target,
         targetQuestion,
-        new Question(dashcardToCheck.card, metadata),
-      ),
-    ) ?? null
+        question,
```
uladzimirdev, line 74: caching Question here is a significant improvement

frontend/src/metabase/parameters/utils/dashboards.ts
kulyk: 😢

uladzimirdev: we have an area for improvement in MLv2 way, e.g. we can cache `visibleColumns`, but it doesn't looks like MLv2 way of working with data for me

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -87,13 +83,18 @@ export function DashCardCardParameterMapper({
...
   editingParameter,
   target,
-  mappingOptions,
   metadata,
   setParameterMapping,
   isMobile,
 }) {
   const [isDropdownVisible, setIsDropdownVisible] = useState(false);
 
+  const mappingOptions = useMemo(
```
uladzimirdev, line 92: @WiNloSt I see in git blame that you have worked on a bug in this area. Do you have in mind any potential problem which useMemo can cause (instead of `getParameterMappingOptions` selector)?

WiNloSt: I've tried to follow the code, so the difference is only that this version has different `card` and `dashcard`.

While previously both are from `getCard` and `getDashCard` respectively. Now they're from the props which come from `getVisibleCards`

https://github.com/metabase/metabase/blob/4ac58f9ded42c405bc43e9ca098d98116fa43337/frontend/src/metabase/dashboard/components/DashboardGrid.tsx#L326-L342

I haven't gone into details about how different these `card` and `dashcard` are. But maybe you should add some comments here, on why this is necessary.

uladzimirdev: nope, card is the same as before, it comes from`[dashcard.card].concat(dashcard.series || []`

uladzimirdev: I probably don't follow how the card appears in the component, but how is the selector different to direct prop passing?

https://github.com/metabase/metabase/blob/a8e8e9e68464775373060946b4f52af69cc4106e/frontend/src/metabase/dashboard/selectors.ts#L291-L292

WiNloSt: Well, you're right I haven't carefully looked at the selector implementations, so both `getCard` and `getDashCard` values are still coming from the props passed to the selector. That means all parameters are the same.

I thought initially, the reason for using `useMemo` was because you want to pass different props. I'm not sure why this is different from using the selector directly (as before the change) though, because the selector `getParameterMappingOptions` is declared using `createSelector` from `@reduxjs/toolkit` which should produce a memoized selector if I'm not mistaken.

uladzimirdev: that's the key 😄 we're working with selectors in terms of `dashboard`. so whenever mapping is changed, all connected components are re-triggered, and `state` in our case is a dashboard, which recalculates everything. Do you have a better idea how to make memoization working with selectors?

WiNloSt: I don't have any answers on top of my head. Optimizing rendering isn't my expert. I'd need to test this out first. But to answer your first question again. I don't think this will ever cause any problem for sure.

frontend/src/metabase/parameters/utils/dashboards.ts
uladzimirdev: @kulyk what do you think if we keep this `JSON.stringify` thing with a comment? I'm working on a PR to improve calculations

PR #38292, opened 2024-01-30
docs - serialization endpoints

jeff-bruemmer:
Adds API endpoints to serialization docs.

### DIFF
--- a/docs/installation-and-operation/serialization.md
+++ b/docs/installation-and-operation/serialization.md
@@ -313,114 +313,132 @@ For now, these endpoints are synchronous. If the serialization process takes too
 
 You can append optional parameters to tell Metabase what to include or exclude from the export. You can also combine parameters (excluding, of course, `all_collections` and selective collections).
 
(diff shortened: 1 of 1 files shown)


docs/installation-and-operation/serialization.md
```diff
@@ -296,8 +296,152 @@ If you've written scripts to automate serialization, you'll need to:
...
+http://localhost:3000/api/ee/serialization/export?all_collections=false
+```
+
+You can include multiple parameters, separated by `&`. For example, to exclude both the settings and the data model from the export:
+
+```html
+http://localhost:3000/api/ee/serialization/export?data_model=false&settings=false
+```
+
+### Example params
```
luizarakaki, line 328: For these params, I think it is worth mentioning the default values.

For the diretory name, the default diretory name is `metabase-yyyy-mm-dd_HH-mm`

luizarakaki (approved):

PR #38308, opened 2024-01-31
use different folders for oss and ee for webpack cache

npretto:
### Description

If we don't wanna run into the annoying memory leak issue with webpack, some of us use `FS_CACHE=true`.

Our doc says:
> When using FS_CACHE=true you may need to remove the node_modules/.cache directory to fix scenarios where the build may be improperly cached, and **you must run rm -rf node_modules/.cache in order for the build to work correctly when alternating between open source and enterprise builds of the codebase**.

This is easy to forget and we may run into weird and hard to debug issues, this PR should fix that by using different cache folders based on the MB_EDITION env var

### How to verify

Not sure if there is a easy to reproduce error that should happen without this when switching between oss and ee, but everything should work fine and we should see the two folders being created when running `MB_EDITION=oss yarn build:hot` and then `MB_EDITION=ee yarn build:hot`

### Demo

image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -12,6 +12,7 @@ const WebpackNotifierPlugin = require("webpack-notifier");
 const ReactRefreshPlugin = require("@pmmmwh/react-refresh-webpack-plugin");
 
 const fs = require("fs");
+const path = require("path");
 
 const ASSETS_PATH = __dirname + "/resources/frontend_client/app/assets";
 const FONTS_PATH = __dirname + "/resources/frontend_client/app/fonts";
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -181,6 +183,11 @@ const config = (module.exports = {
   cache: useFilesystemCache
     ? {
         type: "filesystem",
+        cacheDirectory: path.resolve(
+          __dirname,
+          "node_modules/.cache/",
+          edition === "oss" ? "webpack-oss" : "webpack-ee",
```
npretto, line 189: https://webpack.js.org/configuration/cache/#cachecachedirectory
default folder is `node_modules/.cache/webpack`

WiNloSt:
Requesting @uladzimirdev's review since he's a webpack expert ❤️

WiNloSt (approved):
I tested that, I don't need to bust the cache anymore when changing between different versions. This is huge!

uladzimirdev (approved):

PR #38309, opened 2024-01-31
[E2E] Overload the existing E2E `visitDashboard` helper to make it work with the alias

nemanjaglumac:
This PR overloads the existing `visitDashboard` E2E helper to make it accept either:
1. a direct dashboard id (number) or
2. a dashboard id alias (string)

It then replaces and updates related specs that were grabbing the dashboard id alias and then calling the `visitDashboard` function. The code can be simplified like this:

```diff
- cy.get("@emptyDashboard").then(id => {
-   visitDashboard(id);
- });
+ visitDashboard("@emptyDashboard");
```

### DIFF
--- a/e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
@@ -667,9 +667,7 @@ const openDashboard = (params = {}) => {
     "cardQuery",
   );
 
-  cy.get("@dashboardId").then(dashboardId => {
-    visitDashboard(dashboardId, { params });
-  });
+  visitDashboard("@dashboardId", { params });
 };
 
 const openSlowDashboard = (params = {}) => {
--- a/e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
@@ -675,12 +675,7 @@ const openSlowDashboard = (params = {}) => {
     "cardQuery",
   );
 
-  cy.get("@dashboardId").then(dashboardId => {
-    return cy.visit({
-      url: `/dashboard/${dashboardId}`,
-      qs: params,
-    });
-  });
+  visitDashboard("@dashboardId", { params });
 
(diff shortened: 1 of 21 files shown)


e2e/test/scenarios/dashboard-filters/dashboard-filters-auto-apply.cy.spec.js
```diff
@@ -675,7 +675,12 @@ const openSlowDashboard = (params = {}) => {
     "cardQuery",
   );
 
-  visitDashboard("@dashboardId", { params });
+  cy.get("@dashboardId").then(dashboardId => {
+    return cy.visit({
+      url: `/dashboard/${dashboardId}`,
+      qs: params,
+    });
+  });
```
nemanjaglumac, line 683: For some reason (that I couldn't figure out), every test that uses `cy.clock` in combination with the `visitDashboard` helper fails. It works only when `cy.visit()` is explicitly used.

If anyone has ideas why, I'm all ears.

nemanjaglumac: An example of a failed run:
https://github.com/metabase/metabase/actions/runs/7761351435/job/21214199066#step:11:783

And the recording of that failure:
https://app.replay.io/recording/b9a3ad12-f5ff-465b-98be-8a9813df8aa6

kamilmielnik (approved):
:+1:

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -199,9 +199,7 @@ describe("embed modal display", () => {
   describeEE("when the user has a paid instance", () => {
     it("should display a link to the interactive embedding settings", () => {
       setTokenFeatures("all");
-      cy.get("@dashboardId").then(id => {
-        visitResource("dashboard", id);
-      });
+      visitDashboard("@dashboardId");
```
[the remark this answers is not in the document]
nemanjaglumac: I think it still does in a lot of places
https://github.com/metabase/metabase/blob/d24e534322c9c1345afd1804ea3197085f90a56b/e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js#L36

Issue #37833, opened 2024-01-17
CollectionMover


iethree:
- Frontend
  - https://github.com/metabase/metabase/pull/40388	
  - https://github.com/metabase/metabase/pull/39869
- Backend:  
  - https://github.com/metabase/metabase/issues/40395

PR #38315, opened 2024-01-31
✨ 📁 New Collection Picker 📁 ✨  

### LINKED ISSUE #36411: CollectionPicker
working branch:  https://github.com/metabase/metabase/tree/entity-picker/entity-store

```[tasklist]
### Tasks
- [x] Style Mantine NavLink
- [x] Create Building Block UI item list component
- [x] loads path to selected collection
- [x] handles personal collections
- [x] handle snippet collections
- [x] option: hide personal collections
- [x] handle "all personal collections"
- [x] handles adding a new collection
- [x] use entity cache for collection item fetching
- [x] add search functionality
- [x] tabs support
- [x] handle snap-to-right scroll
- [x] make collectionPicker the top level component
- [x] handles picking collections where a user is missing permissions on parent collections
- [x] make UI components modular and exportable
- [x] need a new entry point button that doesn't look like a dropdown (needs design)
- [x] improve all personal collection fetching performance
- [x] handle read-only collections in search (need can_write hydrated)
- [x] handle verified item icons
- [x] create e2e test helpers for new entity picker
- [x] virtualize item list
- [x] update existing e2e tests that pick collections
- [x] unit tests
- [x] add new e2e tests to fully cover new picker + search
- [x] Make sure there are no regression from the changes in epic #34733 (hide personal items on shared dashboards)
- [x] entities copied/modified from metabase analytics should default to saving in “custom reports”
- [ ] itemlist: scroll pre-selected item into view
```

npfitz:
Closes https://github.com/metabase/metabase/issues/36411

## Description

A New Collection Picking Experience For Metabase!

### Collection picking now happens in horizontal panes

currently, this is implemented for saving new collection, questions, models, and dashboards. More to come soon.

collectioncreate

### Improved, easy to access save in a searched-for location

save in searched place

### Streamlined collection-creation in save dialogs

Improved new collection creation

## Technical Details
There are two main groups of components here
- **Generic Picker building blocks.** These could be used to pick almost anything in metabase
	- `EntityPickerModal`
	- 	EntityPickerSearch`
	- `NestedItemPicker`
	- `ItemList`
	- `EntityItemList`

- **Components designed specifically to pick items in the collection tree.** These components variously wrap, use, and extend the generic building block components to provide a consistent UI, while handling all the fun special casing involved in our collection system.
	- `CollectionPickerModal`
	- `CollectionPicker`
	- `CollectionItemPickerResolver`
	- `RootItemList`
	- `PersonalCollectionItemList`

## Tests
- [x] lots of unit tests added + updated
- [x] lots of e2e tests updated (all the major user flows were already tested, and simply had to be updated)

## Next Steps
Close follow-ups will be to implement this same picking experience in more places
- https://github.com/metabase/metabase/issues/37833
- https://github.com/metabase/metabase/issues/36376
- https://github.com/metabase/metabase/issues/39140


frontend/src/metabase/browse/components/TableBrowser/TableBrowser.styled.tsx
```diff
@@ -11,6 +11,7 @@ import {
 
 export const TableGrid = styled(Grid)`
   width: 100%;
+  flex: 1;
```
iethree, line 14: should we delete this? did we mean to touch the table browser?

npfitz: Yeah, I don't think we meant to do this

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export function EntityPickerModal({
+  title = t`Choose an item`,
+  onItemSelect,
+  onConfirm,
+  selectedItem,
+  onClose,
+  tabs,
+  options = defaultOptions,
```
iethree, line 57: I wonder if we should spread in default options to provided options. if we get a partial options object things might break.

npfitz: Yeah, I like that idea

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  showSearch: true,
+  hasConfirmButtons: true,
+  allowCreateNew: true,
+};
+
+export interface EntityPickerModalProps {
+  title: string;
+  selectedItem: CollectionPickerItem | null;
+  onConfirm: () => void;
+  onItemSelect: (item: CollectionPickerItem) => void;
```
iethree, line 41: I think we should make these generics

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  onClose: () => void;
+  tabs: EntityTab[];
+  options?: EntityPickerOptions;
+  searchResultFilter?: (results: SearchResult[]) => SearchResult[];
+  actions?: JSX.Element[];
```
iethree, line 46: `actionButtons` would be more descriptive here

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.unit.spec.tsx
```diff
@@ -0,0 +1,212 @@
...
+  //       name: "Collection 2",
+  //       model: "collection",
+  //     }),
+  //     createMockCollectionItem({
+  //       id: 4,
+  //       name: "Collection 3",
+  //       model: "collection",
+  //     }),
+  //   ],
+  // });
```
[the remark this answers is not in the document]
npfitz: Agreed haha

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
npfitz: Sure!

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/NewCollectionDialog.tsx
iethree, line 1: maybe this should move to `collectionPicker/` ?

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/NewCollectionDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+        parent_id: parentCollectionId === "root" ? null : parentCollectionId,
+      }),
+    );
+    onNewCollection({ ...newCollection, model: "collection" });
+    onClose();
+  };
+
+  return (
+    <Modal
+      title="Create a new collection"
```
iethree, line 46: should be localized

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/TabsView.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onItemSelect,
+  searchQuery,
+  searchResults,
+  selectedItem,
+}: {
+  tabs: EntityTab[];
+  onItemSelect: (item: CollectionPickerItem) => void;
+  searchQuery: string;
+  searchResults: SearchResult[] | null;
+  selectedItem: CollectionPickerItem | null;
```
iethree, line 21: we should be able to make these types generic 🤞

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
```diff
@@ -0,0 +1,116 @@
...
+  searchResults: SearchResultType[] | null;
+  onItemSelect: (item: WrappedResult) => void;
+  selectedItem: CollectionPickerItem | null;
+}) => {
+  const dispatch = useDispatch();
+
+  if (!searchResults) {
+    return <SearchLoadingSpinner />;
+  }
+
```
iethree, line 84: we need an empty state for zero search results

iethree: haha, you snuck it in, @npfitz

frontend/src/metabase/common/components/EntityPicker/components/ItemList/EntityItemList.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { data, isLoading } = useSearchListQuery<TItem>({ query });
+
+  return (
+    <ItemList
+      items={data}
+      isLoading={isLoading}
+      onClick={onClick}
+      selectedItem={selectedItem}
+      itemName={itemName}
+      isFolder={isFolder}
```
iethree, line 33: huh, look like this is throwing again

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
npfitz: Already done haha

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
```diff
@@ -0,0 +1,80 @@
...
+              data-testid={`item-picker-level-${index}`}
+            >
+              <ErrorBoundary>
+                <ListResolver
+                  query={query}
+                  selectedItem={selectedItem}
+                  options={options}
+                  onClick={(item: TItem) => handleClick(item)}
+                  itemName={itemName}
+                  isCurrentLevel={index === path.length - 2}
```
iethree, line 70: will this logic work for non-collection pickers? if you select a non-folder item, it'll be `path.length - 1` 

this can probably be an issue for another day

frontend/src/metabase/common/components/EntityPicker/components/ItemList/RootItemList.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const { data: personalCollection, isLoading: isLoadingPersonalCollecton } =
+    useCollectionQuery({
+      id: currentUser?.personal_collection_id,
+      enabled: !!currentUser?.personal_collection_id,
+    });
+
+  const {
+    data: rootCollection,
+    isLoading: isLoadingRootCollecton,
+    error: rootCollectionError,
```
iethree, line 61: do we need to disable this when the user doesn't have root collection access?

frontend/src/metabase/common/components/EntityPicker/components/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./EntityPickerModal";
+export * from "./EntityPickerSearch";
+export * from "./LoadingSpinner";
+export * from "./NestedItemPicker";
+export * from "./CollectionPicker";
```
iethree, line 5: I think we should only export
- `NestedItemPicker`
- `ItemPicker`
- `CollectionPicker`

I don't think the others are useful outside this directory

frontend/src/metabase/common/components/EntityPicker/index.ts
```diff
@@ -0,0 +1,4 @@
+// export * from "./components/NestedItemPicker/NestedItemPicker";
+// export * from "./components/EntityPickerModal";
```
iethree, line 2: can delete

frontend/src/metabase/home/components/HomePage/HomePage.tsx
```diff
@@ -46,6 +46,7 @@ export const HomePage = (): JSX.Element => {
 
   return (
     <HomeLayout hasMetabot={getHasMetabot(databases, models, isMetabotEnabled)}>
+      {/* <EntityPickerModal onItemSelect={console.log} tabs={["collection"]} /> */}
```
iethree, line 49: let's delete this 🤣

frontend/src/metabase/hooks/use-modal-open.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect, useState } from "react";
+
+export function useModalOpen() {
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setOpen(true);
+    });
+  }, []);
```
iethree, line 10: is this so we get a transition? if so we should add a comment to that effect.

frontend/src/metabase/nav/components/search/RecentsList/RecentsListContent.tsx
```diff
@@ -85,7 +85,7 @@ export const RecentsListContent = ({
                   >
                     {getItemName(item)}
                   </ResultTitle>
-                  <ModerationIcon
+                  <PLUGIN_MODERATION.ModerationStatusIcon
```
iethree, line 88: why did we make this change?

npfitz: For some reason, the plugin object was getting destructured before the ee stuff was finished overwriting things. I couldn't track down the root cause though :/

frontend/src/metabase/hooks/use-modal-open.ts
npfitz: Correct

mazameli (changes requested):
The bg color of these parent selected items should actually be `background-medium` (EDF2F5).

image

A few other misc things in this new loom: https://www.loom.com/share/b678008378e74e578946c1ea884c11dc
- Not always being scrolled to the right consistently 
- Seeing loading spinner too frequently on a near-empty instance
- If possible, Nick and I had talked about seeing if we could put the Metabase instance analytics pseudo collection at the bottom of the list.

iethree:
> * Seeing loading spinner too frequently on a near-empty instance

https://www.loom.com/share/c59730dc6b504239a892c6d47444d2b1

just pushed an improvement for loading behavior in 2 cases:
- things that load in under 100ms no longer show a spinner, your right, it feels faster even though it's not actually
- creating a new collection no longer causes a cascade of loading spinners 😁

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/TabsView.tsx
```diff
@@ -0,0 +1,92 @@
...
+  selectedItem,
+}: {
+  tabs: EntityTab[];
+  onItemSelect: (item: TItem) => void;
+  searchQuery: string;
+  searchResults: SearchResult[] | null;
+  selectedItem: TItem | null;
+}) => {
+  const hasSearchTab = !!searchQuery;
+  const defaultTab = hasSearchTab ? { model: "search" } : tabs[0];
```
sloansparger, line 24: index access like `[0]` is unsafe in typescript (it should union Type | undefined, but it always returns Type). updating the tabs type like my previous comment mention partially solves this. i prefer to use something like lodash's first method, but unfortunately underscore's typings don't handle arrays with at least one length correctly...

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
```diff
@@ -0,0 +1,134 @@
...
+  models,
+  searchFilter = defaultSearchFilter,
+}: {
+  searchQuery: string;
+  setSearchQuery: (query: string) => void;
+  setSearchResults: (results: SearchResultType[] | null) => void;
+  models: string[];
+  searchFilter?: (results: SearchResultType[]) => SearchResultType[];
+}) {
+  useDebouncedEffect(
```
sloansparger, line 36: if two requests go out and the first resolves after the second, you'd have stale data presented to the user. also seeing that no search sets to null, having no search result would always resolve faster than the network request.

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <Box p="lg" h="100%">
+      {searchResults.length > 0 ? (
+        <Stack h="100%">
+          <VariableSizeItemsVirtualizedList>
+            {searchResults?.map(item => (
+              <EntityPickerSearchResult
+                key={item.model + item.id}
+                result={Search.wrapEntity(item, dispatch)}
+                onClick={(item: WrappedResult) => {
+                  onItemSelect(item as unknown as TItem);
```
sloansparger, line 95: why do we have to typecast?

sloansparger (changes requested):
made it about half way, submitting some feedback for someone to work on

frontend/src/metabase/collections/components/CollectionEmptyState/CollectionEmptyState.tsx
```diff
@@ -38,9 +38,21 @@ const CollectionEmptyState = ({
...
 };
 
-const CollectionEmptyIcon = (): JSX.Element => {
+export const CollectionEmptyIcon = ({
+  height = 94,
+  width = 117,
+}: {
+  height?: number;
+  width?: number;
+}): JSX.Element => {
```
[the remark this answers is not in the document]
npfitz: Could also just revert the changes to this file... We ended up going a different direction haha

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPickerModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    >
+      {t`Create a new collection`}
+    </Button>,
+  ];
+
+  const tabs = [
+    {
+      displayName: t`Collections`,
+      model: "collection",
+      icon: "folder" as IconName,
```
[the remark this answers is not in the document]
npfitz: We may be able to get around it. TS complains that it's a string, but it's limited to specific values

frontend/src/metabase/common/components/EntityPicker/components/ItemList/PersonalCollectionItemList.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const getSortedTopLevelPersonalCollections = (
+  personalCollections?: Collection[],
+): CollectionPickerItem[] | null =>
+  personalCollections
+    ?.filter(isRootPersonalCollection)
+    .map((collection: Collection) => ({
+      ...collection,
+      model: "collection" as SearchModelType,
+    }))
```
rafpaf, line 60: I noodled a little and noticed this also works:

```suggestion
    .map((collection: Collection): CollectionPickerItem => ({
      ...collection,
      model: "collection",
    }))
```

frontend/src/metabase/common/components/EntityPicker/components/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const DelayedLoadingSpinner = ({
+  text,
+  delay = 100,
+}: {
+  text?: string;
+  delay?: number;
+}) => {
+  const [show, setShow] = useState(false);
+
+  useMount(() => {
```
rafpaf, line 23: On this very day, my knowledge increased.

frontend/src/metabase/common/components/EntityPicker/components/ItemList/RootItemList.tsx
```diff
@@ -0,0 +1,123 @@
...
+  } = useCollectionQuery({ id: "root" });
+
+  const { value: data, loading: isLoading } = useAsync(async () => {
```
sloansparger, line 64: i don't think there's any async work happening here

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/AutoScrollBox.tsx
```diff
@@ -0,0 +1,73 @@
...
+    // right.
+    if (
+      container.scrollLeft + container.clientWidth <=
+      container.scrollWidth * 0.99
+    ) {
+      container.scrollLeft += 25;
+    } else {
+      clearInterval(intervalId);
+    }
+  }, 10);
```
rafpaf, line 23: Would `element.scrollIntoView({ behavior: 'smooth', inline: 'start' });` be useful here, or is it not the right effect?

https://codesandbox.io/p/sandbox/horizontal-scrolling-qlrty9

frontend/src/metabase/collections/containers/FormCollectionPicker/FormCollectionPicker.tsx
```diff
@@ -108,62 +74,67 @@ function FormCollectionPicker({
...
+          data-testid="collection-picker-button"
+          id={id}
+          onClick={() => setIsPickerOpen(true)}
+          fullWidth
+          rightIcon={<Icon name="ellipsis" />}
+          styles={{
+            inner: {
+              justifyContent: "space-between",
+            },
+            root: { "&:active": { transform: "none" } },
```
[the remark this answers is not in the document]
rafpaf: I think the "x" button on the collection picker modal probably doesn't need the animation.

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
```diff
@@ -0,0 +1,87 @@
...
+  isFolder: TisFolder<TItem>;
+  listResolver: React.FC<
+    EntityItemListProps<TItem> & {
+      options: EntityPickerOptions;
+    }
+  >;
+}
+
+const generateKey = (query?: SearchListQuery) =>
+  JSON.stringify(query ?? "root").slice(0, 255);
```
sloansparger, line 33: !nit we should use a proper hashing function

frontend/src/metabase/containers/SaveQuestionModal.unit.spec.tsx
```diff
@@ -665,17 +668,64 @@ describe("SaveQuestionModal", () => {
...
+        name: "Child collection",
+        can_write: true,
+      }),
+    };
+    COLLECTION.CHILD.location = `/${COLLECTION.PARENT.id}/`;
+
+    beforeEach(async () => {
+      window.Element.prototype.getBoundingClientRect = jest
+        .fn()
+        .mockReturnValue({ height: 100, width: 200 });
```
rafpaf, line 701: Dang, this is cool.

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
```diff
@@ -0,0 +1,87 @@
...
+    onFolderSelect({ folder });
+  };
+
+  const handleClick = (item: TItem) => {
+    if (isFolder(item)) {
+      handleFolderSelect(item);
+    } else {
+      onItemSelect(item);
+    }
+  };
```
sloansparger, line 54:
```suggestion
  const handleClick = (item: TItem) => {
    if (isFolder(item)) {
      onFolderSelect({ folder: item });
    } else {
      onItemSelect(item);
    }
  };
```
!nit

frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.unit.spec.tsx
```diff
@@ -148,16 +169,63 @@ const setup = async ({
...
+  //   collection: ROOT_COLLECTION,
+  //   collectionItems: [
+  //     createMockCollectionItem({
+  //       ...COLLECTION,
+  //       type: undefined,
+  //       model: "collection",
+  //     }),
+  //     ...dashboards.map(createMockCollectionItem),
+  //   ],
+  // });
```
rafpaf, line 227: Delete?

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  return (
+    <AutoScrollBox
+      data-testid="nested-item-picker"
+      contentHash={generateKey(path[path.length - 1].query)}
```
sloansparger, line 59:
```suggestion
      contentHash={generateKey(_.last(path)?.query)}
```
index access is unsafe

src/metabase/api/collection.clj
```diff
@@ -537,16 +538,20 @@
...
-          (dissoc :collection_position :display :moderated_status :icon
-                  :collection_preview :dataset_query :table_id :query_type :is_upload)
-          update-personal-collection))))
+      (let [visible-collection-ids (collection/permissions-set->visible-collection-ids
+                                  @api/*current-user-permissions-set*)]
+        (for [row rows]
+          ;; Go through this rigamarole instead of hydration because we
+          ;; don't get models back from ulterior over-query
+          ;; Previous examination with logging to DB says that there's no N+1 query for this.
+          ;; However, this was only tested on H2 and Postgres
```
rafpaf, line 547: > this was only tested on H2 and Postgres

I might not understand what this means but I wanted to double check - is this a concern?

rafpaf: Someone other than me - someone with clojure knowledge - should review the clojure code, I think.

iethree:
> - Not always being scrolled to the right consistently
> - Seeing loading spinner too frequently on a near-empty instance
> - If possible, Nick and I had talked about seeing if we could put the Metabase instance analytics pseudo collection at the bottom of the list.

@mazameli #1 and #2 are handled. Sorting usually happens on the backend for these endpoints, so we'll ask @john-metabase to sort that for us.

frontend/src/metabase/components/VirtualizedList/VariableSizeItemsVirtualizedList.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const items = virtualizer.getVirtualItems();
+
+  useMount(() => {
+    if (scrollTo && scrollTo < children.length) {
+      // we need to wait for dynamic measurements to be taken before scrolling
+      const timeout = setTimeout(() => {
+        virtualizer.scrollToIndex(scrollTo, { align: "center" });
+      }, 1);
+
+      return () => timeout && clearTimeout(timeout);
```
sloansparger, line 39: requestAnimationFrame less hacky?

sloansparger (changes requested):

rafpaf (changes requested):
Please pardon the firehose...

What follows are some thoughts about the design, as well as various comments and reactions - mostly nonblocking suggestions. I'd love to sit down and think more about the macro stuff like data flow. @sloansparger is better at that than me!

## Search results

Searching for a blank string triggered a 400 and a spinner that never stopped. Would it work to trim the input?

image

## Responsive picker header

When I launched the picker, then shrank my browser horizontally, the picker then looked like this for me. There's some tiny stuff you could think about doing to make the header look a bit neater: maybe a larger margin to the right of "Select a collection", a smaller line height, and vertically centered alignment.

image

When I launched the new picker with a narrow viewport, it looked like this, which puzzled me a bit. I think this is probably by design but I figured I should point it out.

image

## Personal collections inside "All personal collections"

I clicked "All personal collections", then my personal collection, and the result was a bit surprising. I'm sure this is on purpose, but I wanted to note it, in case it merits further discussion.

## The "Create New" dialog

I think "collection" rather than "folder", and title case ("Create new") rather than sentence case, would be more conventional?

image

For what it's worth, the 'Create' button is disabled if there is no input, but it remains enabled if I add and then remove my input.

When the error message is long, the buttons look like this:

image

## Selecting a collection

Here 'Metabase analytics' is selected, not 'Our analytics', but nothing visually distinguishes them. Maybe an outline?

image

When I launch the picker and I see this, the 'Select' button is initially disabled. When I click 'qwer', nothing changes about 'qwer', but the button becomes enabled. That seems a bit strange to me.

image

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
rafpaf: I wondered about collisions here. If two queries that are meant to be distinct end up using the same key, could that cause problems?

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPickerModal.tsx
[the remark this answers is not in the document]
iethree: I was able to make it happy without the cast by typing `tabs` 👌

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPickerModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  title = t`Choose a collection`,
+  onChange,
+  onClose,
+  value,
+  options = defaultOptions,
+}: CollectionPickerModalProps) => {
+  const [selectedItem, setSelectedItem] = useState<CollectionPickerItem | null>(
+    null,
+  );
+  const [createDialogOpen, setCreateDialogOpen] = useState(false);
```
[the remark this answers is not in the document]
iethree: we can use `metabase/hooks/use-toggle` 👍

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/NewCollectionDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+  onClose,
+  parentCollectionId,
+  onNewCollection,
+}: NewCollectionDialogProps) => {
+  const dispatch = useDispatch();
+
+  const onCreateNewCollection = async ({ name }: { name: string }) => {
+    const {
+      payload: { collection: newCollection },
+    } = await dispatch(
```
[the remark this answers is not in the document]
iethree: 🪄 

Screen Shot 2024-02-27 at 9 24 30 AM

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
iethree: out of curiosity, does RTK-Query have some nice magic to handle this?

sloansparger: You can cancel previous requests by providing an abort controller, but unfortunately there's no auto-canceling / debouncing magic.

iethree: this was surprisingly hard in combination with the debounce 😂

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
iethree: how much time you got? 😂

there's a couple type issues that are really tricky here. I'll grab some time with you to see if you have any better ideas than we came up with. these couple casts were by far the least bad option we came up with

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
sloansparger: I'm not sure what you landed on, but a potential strategy is you could create a variable called `requestId` that is equal to `new Date().getTime()` on line 39. You would also want a ref that would store the latest requestId and then only setState if the latest request id ref === the locally scoped requestId.

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
sloansparger: Let me think about this... there are native hashing function, but I'm not 100% confident about browser support:
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

frontend/src/metabase/components/VirtualizedList/VariableSizeItemsVirtualizedList.tsx
iethree: seems to work!

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
iethree: looks like that would fail on http:// origins https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto

sloansparger: > I think collision is basically impossibly with our current API
If that's the case with 255 characters, let's just make it literally impossible and not slice the JSON output?

frontend/src/metabase/common/components/EntityPicker/components/ItemList/RootItemList.tsx
iethree: you're right! (there used to be, but we moved it out)

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/NestedItemPicker.tsx
sloansparger: Sorry, I should have been more judicious about _why_ this was needed. Can't we rely on object equality? It looks like this has is ultimately just being used by the auto-scrolling feature to detect a change. Couldn't we pass down `path[path.length - 1].query` instead of `generateKey(path[path.length - 1].query)`?

frontend/src/metabase-types/api/search.ts
```diff
@@ -96,4 +97,5 @@ export interface SearchListQuery {
   offset?: number;
   collection?: CollectionId;
   filter_items_in_personal_collection?: "only" | "exclude";
+  namespace?: "snippets";
```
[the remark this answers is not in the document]
iethree:
```ts
// snippets namespace is stupid
```

frontend/src/metabase-types/api/collection.ts
```diff
@@ -56,7 +56,8 @@ export interface Collection {
   personal_owner_id?: UserId;
   is_personal?: boolean;
 
-  location?: string;
+  location: string | null;
+  effective_location?: string; // location with only user-accessible collections
```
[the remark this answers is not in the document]
iethree: unfortunately, not sure you can say a tiny bit here:

1. The metabase permissions model is ... unique ... in that a user can have access to a non-hierarchical set of collections
2. So there might be a set of collections `a/b/c/d` and a given user might only have access to collections `b` and `d`, so the effective path for `d` would be `b/d` for them because they can't access `a` or `c` 🤯 .

frontend/src/metabase-types/api/collection.ts
```diff
@@ -108,6 +111,7 @@ export interface CollectionListQuery {
   archived?: boolean;
   "exclude-other-user-collections"?: boolean;
   "exclude-archived"?: boolean;
+  "personal-only"?: boolean;
```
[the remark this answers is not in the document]
iethree: my understanding: clojure likes `kebab-case`, `snake_case` plays nicely with both js and http, the current idiom in js-land is to `camelCase`in code, but not over http.

frontend/src/metabase-types/api/search.ts
```diff
@@ -55,7 +55,7 @@ export type CollectionEssentials = Pick<
 >;
 
 export interface SearchResult {
-  id: number;
+  id: number | CollectionId;
```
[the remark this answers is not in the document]
iethree: I guess we should really:
```ts
  id: CollectionId | CardId | DatabaseId | TableId | DashboardId;
```

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPicker.tsx
```diff
@@ -0,0 +1,207 @@
...
+import { PERSONAL_COLLECTIONS } from "metabase/entities/collections";
+import { useSelector } from "metabase/lib/redux";
+import { getUserPersonalCollectionId } from "metabase/selectors/user";
+import type { CollectionId } from "metabase-types/api";
+
+import type {
+  PickerState,
+  CollectionPickerItem,
+  TypeWithModel,
+  TisFolder,
```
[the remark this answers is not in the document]
iethree: that's why I like about it

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPicker.tsx
```diff
@@ -0,0 +1,207 @@
...
+  showRootCollection: true,
+};
+
+interface CollectionPickerProps {
+  onItemSelect: (item: CollectionPickerItem) => void;
+  initialValue?: Partial<CollectionPickerItem>;
+  options?: CollectionPickerOptions;
+}
+
+const isFolder: TisFolder<CollectionPickerItem> = <TItem extends TypeWithModel>(
```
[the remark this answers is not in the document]
iethree: in entity picker land "folder" is a broader concept than collection - with respect to actions, models are folders. Soon, dashboards will be folders too. Databases and schemas will also be folders.

frontend/src/metabase/common/components/EntityPicker/components/ItemList/ItemList.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { PickerColumn } from "./ItemList.styled";
+
+interface ItemListProps<TItem extends TypeWithModel> {
+  items?: TItem[];
+  isLoading?: boolean;
+  onClick: (val: TItem) => void;
+  selectedItem: TItem | null;
+  isFolder: (item: TItem) => boolean;
+  isCurrentLevel: boolean;
+  itemName?: string;
```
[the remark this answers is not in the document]
iethree: ah good catch, we were drilling this, but then got a design change that made it unnecessary 👍

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+    () => ({
+      ...defaultOptions,
+      ...options,
+    }),
+    [options],
+  );
+
+  const { open } = useModalOpen();
+
+  const hasTabs = tabs.length > 1 || searchQuery;
```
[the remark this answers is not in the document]
iethree: from a UI standpoint if there is only 1 tab, we hide the tab bar altogether so there are no tabs.

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/SinglePickerView.tsx
```diff
@@ -0,0 +1,18 @@
...
+import type { EntityTab } from "../../types";
+
+export const SinglePickerView = ({ tab }: { tab: EntityTab }) => {
+  return (
+    <Box
+      style={{
+        borderTop: `1px solid ${color("border")}`,
+        flexGrow: 1,
+        height: 0,
+      }}
```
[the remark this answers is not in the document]
iethree: yeah, this made me realize we don't even need this file anymore, it used to do more, but it can just be a styled component now.

frontend/src/metabase/common/components/EntityPicker/components/ItemList/ItemList.tsx
```diff
@@ -0,0 +1,95 @@
...
+}: ItemListProps<TItem>) => {
+  const activeItemIndex = useMemo(() => {
+    if (!items) {
+      return -1;
+    }
+
+    return items.findIndex(
+      item =>
+        item.id === selectedItem?.id && item.model === selectedItem?.model,
+    );
```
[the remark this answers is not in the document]
iethree: it is 😂 (`isSelectedItem`) maybe I should use it here 🤦‍♂️

frontend/src/metabase/common/components/EntityPicker/components/ItemList/PersonalCollectionItemList.tsx
```diff
@@ -0,0 +1,64 @@
...
+    .sort((a, b) => a?.name.localeCompare(b.name)) ?? null;
+
+const isRootPersonalCollection = (collection: Collection) =>
+  collection.is_personal && collection.location === "/";
```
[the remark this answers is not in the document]
iethree: unfortunately not, it won't work with the search API we're using.

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/AutoScrollBox.tsx
iethree: 🤔 I'll give it a try

src/metabase/api/collection.clj
iethree: oh yes, good call, Dan wrote the clojure!

iethree:
@rafpaf thanks for the detailed review. On the design comments specifically, a number of those items were ones Maz flagged last week and have since been fixed. As to the small screen experience, we have a separate milestone for implementing small-screen specific views, so we'll handle those then.

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/AutoScrollBox.tsx
npfitz: Probably better than the derpy "JS scroll" I put together here haha.

johnswanson (approved):
Backend looks good!

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+  ref: React.Ref<unknown>,
+) => {
+  const [path, setPath] = useState<PickerState<CollectionPickerItem>>(() =>
+    getStateFromIdPath({
+      idPath: ["root"],
+      namespace: options.namespace,
+    }),
+  );
+
+  const { data: currentCollection, isLoading: loadingCurrentCollection } =
```
sloansparger, line 55: Do we need to handle the error case?

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/TabsView.tsx
sloansparger: Comment still stands, but I won't let it block releasing this.

frontend/src/metabase/common/components/EntityPicker/components/ItemList/EntityItemList.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export const EntityItemList = <TItem extends TypeWithModel>({
+  query,
+  onClick,
+  selectedItem,
+  isFolder,
+  isCurrentLevel,
+}: EntityItemListProps<TItem>) => {
+  const { data, isLoading } = useSearchListQuery<TItem>({ query });
```
sloansparger, line 22: do we need to handle an error state here?

frontend/src/metabase/common/components/EntityPicker/components/ItemList/PersonalCollectionItemList.tsx
```diff
@@ -0,0 +1,63 @@
...
+  isCurrentLevel: boolean;
+}
+
+export const PersonalCollectionsItemList = <TItem extends TypeWithModel>({
+  onClick,
+  selectedItem,
+  isFolder,
+  isCurrentLevel,
+}: PersonalCollectionsItemListProps<TItem>) => {
+  const { data: collections, isLoading } = useCollectionListQuery({
```
sloansparger, line 27: same error state comment

frontend/src/metabase/common/components/EntityPicker/components/ItemList/RootItemList.tsx
```diff
@@ -0,0 +1,125 @@
...
+  onClick,
+  selectedItem,
+  options,
+  isFolder,
+  isCurrentLevel,
+}: RootItemListProps<CollectionPickerItem>) => {
+  const isAdmin = useSelector(getUserIsAdmin);
+  const currentUser = useSelector(getUser);
+
+  const { data: personalCollection, isLoading: isLoadingPersonalCollecton } =
```
sloansparger, line 50: error state?

frontend/src/metabase/containers/AddToDashSelectDashModal/AddToDashSelectDashModal.unit.spec.tsx
```diff
@@ -148,16 +169,51 @@ const setup = async ({
...
+  window.Element.prototype.getBoundingClientRect = jest
+    .fn()
+    .mockReturnValue({ height: 100, width: 200 });
   const dashboards = Array.from(
     new Set([dashboard, mostRecentlyViewedDashboard].filter(isNotNull)),
   );
 
   setupCollectionsEndpoints({ collections, rootCollection: ROOT_COLLECTION });
-  setupDashboardCollectionItemsEndpoint(dashboards);
+  // setupDashboardCollectionItemsEndpoint(dashboards);
```
sloansparger, line 180: delete?

sloansparger (approved):
approved with comment (they may be non-issues)

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
```diff
@@ -0,0 +1,134 @@
...
+export const EntityPickerSearchTab = ({
+  searchResults,
+  searchQuery,
+}: {
+  searchResults: SearchResultType[] | null;
+  searchQuery: string;
+}) => (
+  <Tabs.Tab key="search" value="search" icon={<Icon name="search" />}>
+    {searchResults
+      ? t`${searchResults.length} results for "${searchQuery}"`
```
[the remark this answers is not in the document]
rafpaf: I think the `c` suggestion is not really blocking, but pluralization (ideally with `ngettext`) is quasi-blocking in my possibly incorrect opinion. I will leave that up to you two.

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/AutoScrollBox.tsx
rafpaf: @iethree @npfitz What do you think? Though Nick's code is rock solid, I was thinking it might improve DX to use the browser's native powers.

rafpaf (commented):
I think you're very close!

frontend/src/metabase/common/components/EntityPicker/components/NestedItemPicker/AutoScrollBox.tsx
npfitz: I like the scroll into view idea better than what we have here. I'll try to give it a go later today

iethree: my misgiving is the grossness of having to assign a ref to the last child and pass it around. maybe it's not that gross 🤷

npfitz: The reason I like it is because on more than 1 occasion, the entity picker has decided to always scroll left for me (infinite loop of scrolling left). Happened mostly when I'm on battery and things are slowing down on my machine. Using scrollIntoView would likely remove that possibility from ever happening

frontend/src/metabase/common/components/EntityPicker/components/CollectionPicker/CollectionPickerModal.tsx
```diff
@@ -67,19 +72,19 @@ export const CollectionPickerModal = ({
   const modalActions = [
     <Button
       key="collection-on-the-go"
-      onClick={() => setCreateDialogOpen(true)}
+      onClick={openCreateDialog}
```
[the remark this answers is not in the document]
iethree: I don't think there's any danger in this if it's parameterless function, and when you're doing `() => fn()`, you're creating a new function with a new memory allocation, etc. Not a big deal, but no reason to do it without a reason.

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerSearch/EntityPickerSearch.tsx
```diff
@@ -0,0 +1,134 @@
...
+                  selectedItem?.model === item.model
+                }
+              />
+            ))}
+          </VariableSizeItemsVirtualizedList>
+        </Stack>
+      ) : (
+        <Flex direction="column" justify="center" h="100%">
+          <EmptyState
+            title={t`Didn't find anything`}
```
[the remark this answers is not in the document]
iethree: this is an existing string. it won't be newly translated

frontend/src/metabase/common/components/EntityPicker/components/ItemList/ItemList.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }, [items, selectedItem]);
+
+  if (isLoading && !items) {
+    return (
+      <Box miw={310} h="100%" aria-label={t`loading`}>
```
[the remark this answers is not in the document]
iethree: since this string exists already, it won't be newly translated, so the context won't do anything here.

frontend/src/metabase/common/components/EntityPicker/components/ItemList/RootItemList.tsx
iethree: this one is actually handled implicitly and silently in the `useMemo`

luizarakaki (approved):

rafpaf (approved):
In honour - and in honor, since this is a binational venture - of this PR, please imagine Seamus Heaney reading this poem aloud to you in ringing tones: https://www.poetryfoundation.org/poems/50981/blackberry-picking

iethree (approved):
Wow this is all amazing.

mazameli (approved):
🥳

PR #38321, opened 2024-01-31
Fix issue where reverted question does not use parameter values

### LINKED ISSUE #38176: Reverting a query that has parameters to a previous version does not use the parameters on the reloaded dataset
### Describe the bug

Reverting a question to a previous version reloads the dataset, but it does not pass the current variable value when it does so, resulting in an error where it shouldn't.

### To Reproduce

1. `New` → `SQL Query` → `Example Dataset`
2. Write a query:
   ```sql
   SELECT * FROM "ACCOUNTS" WHERE "COUNTRY" = {{ country }}
   ```
3. `Save` the question (ie, as `Foo`)
4. Edit the SQL to:
   ```sql
   SELECT * FROM "ACCOUNTS" WHERE "COUNTRY" = {{ country }} AND true
   ```
5. `Save` the question again and `Replace the original question 'Foo'`
6. Open saved question
7. Enter a value for the `Country` filter: `NL`
8. Run the question and make sure you're getting data
9. Click the `More info` button on the top right (`i`)
10. In the history sidebar, on the first version click the `Revert this version` button (`⟲`)
11. You'll see and error in the explorer:
    ```
    Cannot run the query: missing required parameters: #{"country"}
    ```
    Even though the `Country` field has a value (ie, we entered `NL` before)

### Expected behavior

It should use the value for the parameter

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:121.0) Gecko/20100101 Firefox/121.0",
    "vendor": ""
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "21.0.1",
    "java.vendor": "Homebrew",
    "java.version": "21.0.1",
    "java.vm.version": "21.0.1",
    "os.name": "Mac OS X",
    "os.version": "14.2.1",
    "user.language": "en",
    "user.timezone": "Europe/Amsterdam"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2024-01-26",
      "src_hash": "8353342e1825630812764c607a2dc66211bd437a",
      "tag": "v0.48.1-SNAPSHOT",
      "hash": "c6949ca"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

P3

### Additional context

A related issue is being fixed in #38095

romeovs:
Closes https://github.com/metabase/metabase/issues/38176

### Description

Fixes #38176. The shape of the parameter values passed to `Question` was wrong.

My fix uses `question._parameterValues`. I'm not sure how to pull the current parameter values from a question in the correct shape, so I'm open to suggestions.

### How to verify

1. `New` → `SQL Query` → `Example Dataset`
2. Write a query:
   ```sql
   SELECT * FROM "ACCOUNTS" WHERE "COUNTRY" = {{ country }}
   ```
3. `Save` the question (ie, as `Foo`)
4. Edit the SQL to:
   ```sql
   SELECT * FROM "ACCOUNTS" WHERE "COUNTRY" = {{ country }} AND true
   ```
5. `Save` the question again and `Replace the original question 'Foo'`
6. Open saved question
7. Enter a value for the `Country` filter: `NL`
8. Run the question and make sure you're getting data
9. Click the `More info` button on the top right (`i`)
10. In the history sidebar, on the first version click the `Revert this version` button (`⟲`)
11. You should not see and error in the explorer, but instead see the data.

### Checklist

### DIFF
--- a/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
@@ -28,7 +28,8 @@ describe("issue 38176", () => {
     );
 
     cy.findByPlaceholderText("Country").type("NL");
-    cy.get("button[aria-label='Get Answer']").first().click();
+
+    cy.findByTestId("query-builder-main").button("Get Answer").click();
 
     questionInfoButton().click();
     rightSidebar().within(() => {
--- a/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
@@ -45,8 +45,6 @@ describe("issue 38176", () => {
       cy.findByText(/reverted to an earlier version/i).should("be.visible");
     });
 
-    cy.findByTestId("TableInteractive-root").within(() => {
-      cy.findByText("COUNTRY").should("be.visible");
-    });
+    cy.findAllByRole("gridcell").should("contain", "NL");
   });
 });
--- a/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
@@ -27,7 +27,7 @@ describe("issue 38176", () => {
       { visitQuestion: true },
     );
 
-    cy.findByPlaceholderText("Country").type("NL", { delay: 0 });
+    cy.findByPlaceholderText("Country").type("NL");
     cy.get("button[aria-label='Get Answer']").first().click();
 
     questionInfoButton().click();
--- a/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
+++ b/e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
@@ -1,0 +1,52 @@
+import { restore, questionInfoButton, rightSidebar } from "e2e/support/helpers";
+
+describe("issue 38176", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsNormalUser();
+    cy.intercept("PUT", "/api/card/**").as("updateQuestion");
+  });
+
+  it("restoring a question to a previous version should preserve the variables", () => {
+    cy.createNativeQuestion(
+      {
+        name: "30165",
+        native: {
+          query:
+            'SELECT "COUNTRY" from "ACCOUNTS" WHERE country = {{ country }} LIMIT 5',
+          "template-tags": {
+            country: {
+              type: "text",
+              id: "dd06cd10-596b-41d0-9d6e-94e98ceaf989",
+              name: "country",
+              "display-name": "Country",
+            },
+          },
+        },
+      },
+      { visitQuestion: true },
+    );
+
+    cy.findByPlaceholderText("Country").type("NL", { delay: 0 });
+    cy.get("button[aria-label='Get Answer']").first().click();
+
+    questionInfoButton().click();
+    rightSidebar().within(() => {
+      cy.findByText("History");
+
+      cy.findByPlaceholderText("Add description")
+        .type("This is a question")
+        .blur();
+
+      cy.wait("@updateQuestion");
+      cy.findByText(/added a description/i);
+      cy.findByTestId("question-revert-button").click();
+
+      cy.findByText(/reverted to an earlier version/i).should("be.visible");
+    });
+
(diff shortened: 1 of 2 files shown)


e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
```diff
@@ -0,0 +1,52 @@
+import { restore, questionInfoButton, rightSidebar } from "e2e/support/helpers";
+
+describe("issue 38176", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsNormalUser();
+    cy.intercept("PUT", "/api/card/**").as("updateQuestion");
+  });
+
+  it("restoring a question to a previous version should preserve the variables", () => {
```
nemanjaglumac, line 10:
```suggestion
  it("restoring a question to a previous version should preserve the variables (metabase#38176)", () => {
```

e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
```diff
@@ -0,0 +1,52 @@
...
+              name: "country",
+              "display-name": "Country",
+            },
+          },
+        },
+      },
+      { visitQuestion: true },
+    );
+
+    cy.findByPlaceholderText("Country").type("NL", { delay: 0 });
```
nemanjaglumac, line 30: Again, no need to shorten the delay on really short strings. Use this rarely, only when string is really long.

e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      cy.wait("@updateQuestion");
+      cy.findByText(/added a description/i);
+      cy.findByTestId("question-revert-button").click();
+
+      cy.findByText(/reverted to an earlier version/i).should("be.visible");
+    });
+
+    cy.findByTestId("TableInteractive-root").within(() => {
+      cy.findByText("COUNTRY").should("be.visible");
+    });
```
nemanjaglumac, line 50:
```suggestion
    cy.findByTestId("TableInteractive-root").findByText("COUNTRY").should("be.visible");
```

Not wrong, but also not needed. `cy.within()` is helpful only when you need to do multiple assertions scoped to a single block. If you only need to find one element within a block, chaining it directly is perfectly fine.

e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  beforeEach(() => {
+    restore();
+    cy.signInAsNormalUser();
+    cy.intercept("PUT", "/api/card/**").as("updateQuestion");
+  });
+
+  it("restoring a question to a previous version should preserve the variables", () => {
+    cy.createNativeQuestion(
+      {
+        name: "30165",
```
nemanjaglumac, line 13:
```suggestion
        name: "38176",
```

e2e/test/scenarios/question/reproductions/38176-revert-to-previous-does-not-honor-parameters.cy.spec.js
nemanjaglumac: Actually, since there is only one header cell, you can use this:
```js
cy.findByTestId("header-cell").should("have.text", "COUNTRY");
```

Btw, it is possible that the table header renders before we render table results. In that case, this test wouldn't catch potential failure to render the actual results. To err on the side of caution, let's add another assertion for those results:
```js
cy.findAllByRole("gridcell").should("contain", "NL");
```

frontend/src/metabase/query_builder/actions/core/core.js
```diff
@@ -76,10 +76,19 @@ export const reloadCard = createThunkAction(RELOAD_CARD, () => {
       Questions.actions.fetch({ id: outdatedQuestion.id() }, { reload: true }),
     );
     const card = Questions.HACK_getObjectFromAction(action);
+
+    const parameterValues = outdatedQuestion.parameters().reduce(
```
uladzimirdev, line 80: ❤️

uladzimirdev: it would be nice to have a comment that we need to get the same format as `_parameters` have

uladzimirdev (approved):

nemanjaglumac (approved):

PR #38322, opened 2024-01-31
BE: add `based_on_upload` to `GET /api/collections/:collection/items` response

calherries:
[epic](https://github.com/metabase/metabase/issues/35614)

This PR implements the BE part of milestone 1 of [the epic](https://github.com/metabase/metabase/issues/35614). 

It adds a `:based_on_upload` key to the response of the `GET /api/collections/:id/items` and `GET /api/collections/root/items` endpoints.

The `:based_on_upload key` is only added for items that correspond to models, not questions.

The `:based_on_upload` is either:
1. the ID of the underlying table if the model can be uploaded to
2. missing if the model cannot be uploaded to

It reuses the same logic as https://github.com/metabase/metabase/pull/37052 to derive the `:based_on_upload` key.


src/metabase/upload.clj
```diff
@@ -678,7 +676,53 @@
...
+                           ;; they should never be null in production
+                           [:dataset_query [:maybe ms/Map]]
+                           [:query_type    [:maybe [:or :string :keyword]]]
+                           [:table_id      [:maybe ms/PositiveInt]]
+                           ;; is_upload can be provided for an optional optimization
+                           [:is_upload {:optional true} [:maybe :boolean]]]]]
+  (let [table-ids (->> models
+                       ;; as an optimization, we might already know that the table is not an upload
+                       ;; if is_upload=false. We can skip making more queries if so
+                       (remove #(false? (:is_upload %)))
```
qnkhuat, line 711: in which case we benefit from this optimization?

src/metabase/upload.clj
```diff
@@ -678,7 +676,53 @@
...
+  [table-ids]
+  (if (empty? table-ids)
+    #{}
+    (let [tables (t2/hydrate (t2/select :model/Table :id [:in table-ids]) :db)]
+      (set (keep (fn [t]
+                   (when (can-upload-to-table? (:db t) t)
+                     (:id t)))
+                 tables)))))
+
+(defn- no-joins?
```
qnkhuat, line 690:
```suggestion
(defn- query-no-joins?
```

IMO, if the object that a helper function operates on is not the main object of the namespace, we should prefix it.

For example, it doesn't give much context if you call `(upload/no-joins ...)` since `upload` doesn't have a join property by itself.

src/metabase/upload.clj
```diff
@@ -678,7 +676,53 @@
...
+  "Batch hydrates `:based_on_upload` for each item of `models`. Assumes each item of `model` represents a model."
+  [models :- [:sequential [:map
+                           ;; query_type and dataset_query can be null in tests, so we make them nullable here.
+                           ;; they should never be null in production
+                           [:dataset_query [:maybe ms/Map]]
+                           [:query_type    [:maybe [:or :string :keyword]]]
+                           [:table_id      [:maybe ms/PositiveInt]]
+                           ;; is_upload can be provided for an optional optimization
+                           [:is_upload {:optional true} [:maybe :boolean]]]]]
+  (let [table-ids (->> models
```
qnkhuat, line 708: nit: these lets binding are not aligned, and I noticed you no longer do that for some of your PRs. Have we stopped going that :( IMO, it aids readability by a good amount.

src/metabase/upload.clj
```diff
@@ -678,7 +676,53 @@
...
+                       set)
+        uploadable-table-ids (set (uploadable-table-ids table-ids))
+        based-on-upload (fn [model]
+                          (when (:dataset_query model) ; dataset_query is sometimes null in tests
+                            (let [query (lib/->pMBQL (:dataset_query model))]
+                              (when (and (= (name (:query_type model)) "query")
+                                         (contains? uploadable-table-ids (:table_id model))
+                                         (no-joins? query))
+                                (lib/source-table-id query)))))]
+    (map #(m/assoc-some % :based_on_upload (based-on-upload %))
```
qnkhuat, line 722: why `assoc-some` ?

src/metabase/upload.clj
```diff
@@ -678,7 +676,53 @@
...
+                           [:is_upload {:optional true} [:maybe :boolean]]]]]
+  (let [table-ids (->> models
+                       ;; as an optimization, we might already know that the table is not an upload
+                       ;; if is_upload=false. We can skip making more queries if so
+                       (remove #(false? (:is_upload %)))
+                       (keep :table_id)
+                       set)
+        uploadable-table-ids (set (uploadable-table-ids table-ids))
+        based-on-upload (fn [model]
+                          (when (:dataset_query model) ; dataset_query is sometimes null in tests
```
qnkhuat, line 716:
```suggestion
                          (when (:dataset_query model) ; dataset_query is sometimes null in tests
```
nit: it can be `when-let`

src/metabase/api/collection.clj
```diff
@@ -597,7 +609,8 @@
   [:id :name :description :entity_id :display [:collection_preview :boolean] :dataset_query
    :model :collection_position :authority_level [:personal_owner_id :integer]
    :last_edit_email :last_edit_first_name :last_edit_last_name :moderated_status :icon
-   [:last_edit_user :integer] [:last_edit_timestamp :timestamp] [:database_id :integer]])
+   [:last_edit_user :integer] [:last_edit_timestamp :timestamp] [:database_id :integer]
+   [:table_id :integer] [:is_upload :boolean] :query_type])
```
qnkhuat, line 613: I think it's a good habit to annotate what are these special columns are for
```suggestion
   ;; for determining whether a model is based on csv upload
   [:table_id :integer] [:is_upload :boolean] :query_type])
```

test/metabase/api/collection_test.clj
```diff
@@ -639,6 +640,17 @@
                 (:data (mt/user-http-request :crowberto :get 200
                                              (str "collection/" (u/the-id collection) "/items"))))))))))
 
+(deftest collection-items-based-on-upload-test
+  (testing "GET /api/collection/:id/items"
+    (testing "check that based_on_upload is returned for cards correctly"
+      (api.card-test/run-based-on-upload-test
+       (fn [card]
+         (->> (mt/user-http-request :crowberto :get 200 (str "collection/" (:collection_id card) "/items"))
```
qnkhuat, line 648: you either should add `models=dataset` to the query param`, or you need to filter by model below.

src/metabase/upload.clj
calherries: The code path from the collection items endpoint. I'll add this to the comment

src/metabase/upload.clj
calherries: I can't see much benefit of applying that rule for private helper functions like this one which are only used in one function. I would agree with your example though, no one should use uploads/no-joins?without renaming it appropriately or moving it to an appropriate namespace.

src/metabase/upload.clj
calherries: I still align let bindings by default but sometimes break it if I think it leads to significantly more new lines. I can align this one though and can tend more towards aligning things in the future

src/metabase/upload.clj
calherries: Just thought it was cleaner. Why not?

qnkhuat: I'd expect `based_on_upload`  to be a boolean key after calling `model-hydrate-based-on-upload.`

it'd be inconsistent if, for a list of models, some will have the key, and some are not

test/metabase/api/collection_test.clj
calherries: why do you suggest this? I don't want to do this because we need to test responses for cards as well as datasets

src/metabase/upload.clj
calherries: Agreed it feels a little surprising. Using `assoc` here would mean `based_on_upload` would be not present for cards, but `based_on_upload` is present for models, which feels equally inconsistent to me.

test/metabase/api/collection_test.clj
qnkhuat: to make sure the thing you filtered by id is a model, not sth that accidentally have the same `id`

calherries: true, updated in [8e21f21](https://github.com/metabase/metabase/pull/38322/commits/8e21f2140ad8a6b87258456d7001bf309aa1b0e7)

qnkhuat (approved):
Cool, LGTM

src/metabase/api/collection.clj
```diff
@@ -402,7 +406,15 @@
...
   [_ rows]
-  (post-process-collection-children :card rows))
+  (let [dataset_queries (map :dataset_query rows)]
+    (->> rows
+         ;; normalize dataset queries just for based_on_upload hydration
+         (map #(update % :dataset_query (comp mbql.normalize/normalize json/parse-string)))
+         upload/model-hydrate-based-on-upload
+         (map (fn [dataset_query row]
+                (assoc row :dataset_query dataset_query))
+              dataset_queries)
```
crisptrutski, line 416: This "normalize then undo" pattern could be made clearer IMHO. I see this has already shipped, so I'll send another PR for it.

src/metabase/upload.clj
```diff
@@ -678,7 +676,54 @@
 ;;; |  hydrate based_on_upload for FE
 ;;; +--------------------------------
 
-(mi/define-simple-hydration-method based-on-upload
+(defn uploadable-table-ids
+  "Returns the subset of table ids where the user can upload to the table."
+  [table-ids]
+  (if (empty? table-ids)
```
crisptrutski, line 682: Does toucan / honeysql have a fast "no query" return path for `[:in []]` queries? It just seems a bit sad to hand roll this optimization 😄

src/metabase/upload.clj
```diff
@@ -678,7 +676,54 @@
...
+    #{}
+    (let [tables (t2/hydrate (t2/select :model/Table :id [:in table-ids]) :db)]
```
crisptrutski, line 684: nit: I think this would have been clearer with a threading macro: 

```clojure
(->> (t2/select ..)
      t2/hydrate
      (keep ...)
      set)
```

src/metabase/upload.clj
crisptrutski: Agree on not needing a prefix for a private method (implicit in the arg name), but I do find this a strange name for a predicate. Having a `has-joins?` method that just gets negated at the call site feels a bit clearer to me.

src/metabase/upload.clj
```diff
@@ -678,7 +676,54 @@
...
+  (let [all-joins (mapcat (fn [stage]
+                            (lib/joins query stage))
+                          (range (lib/stage-count query)))]
+    (empty? all-joins)))
+
+(mu/defn model-hydrate-based-on-upload
+  "Batch hydrates `:based_on_upload` for each item of `models`. Assumes each item of `model` represents a model."
+  [models :- [:sequential [:map
+                           ;; query_type and dataset_query can be null in tests, so we make them nullable here.
+                           ;; they should never be null in production
```
crisptrutski, line 702: How sure are we *really* that it's never null in production?

I'd love to have a backlog item to backfill and make non-nullable, and update all the tests.

src/metabase/upload.clj
```diff
@@ -678,7 +676,54 @@
...
+                           [:dataset_query [:maybe ms/Map]]
+                           [:query_type    [:maybe [:or :string :keyword]]]
+                           [:table_id      [:maybe ms/PositiveInt]]
+                           ;; is_upload can be provided for an optional optimization
+                           [:is_upload {:optional true} [:maybe :boolean]]]]]
+  (let [table-ids            (->> models
+                                  ;; as an optimization when listing collection items (GET /api/collection/items),
+                                  ;; we might already know that the table is not an upload if is_upload=false. We
+                                  ;; can skip making more queries if so
+                                  (remove #(false? (:is_upload %)))
```
crisptrutski, line 712: Is this relevant to the discussion about `assoc` versus `assoc-with`, or would there still be call paths that never populate the key for models?

src/metabase/upload.clj
```diff
@@ -678,7 +676,54 @@
...
+                                  (keep :table_id)
+                                  set)
+        uploadable-table-ids (set (uploadable-table-ids table-ids))
+        based-on-upload      (fn [model]
+                               (when-let [dataset_query (:dataset_query model)] ; dataset_query is sometimes null in tests
+                                 (let [query (lib/->pMBQL dataset_query)]
+                                   (when (and (= (name (:query_type model)) "query")
```
crisptrutski, line 719: Should this not be "failing closed" for nil query_type? `(= (name (:query_type model "query")) "query")`

src/metabase/upload.clj
```diff
@@ -678,7 +676,54 @@
...
+                           ;; is_upload can be provided for an optional optimization
+                           [:is_upload {:optional true} [:maybe :boolean]]]]]
+  (let [table-ids            (->> models
+                                  ;; as an optimization when listing collection items (GET /api/collection/items),
+                                  ;; we might already know that the table is not an upload if is_upload=false. We
+                                  ;; can skip making more queries if so
+                                  (remove #(false? (:is_upload %)))
+                                  (keep :table_id)
+                                  set)
+        uploadable-table-ids (set (uploadable-table-ids table-ids))
```
crisptrutski, line 715: This method already returns a set.

src/metabase/upload.clj
calherries: they don't

PR #38325, opened 2024-01-31
[Dashboard]: Convert mappings to a map for better performance

uladzimirdev:
Implementing @oleggromov idea https://github.com/metabase/metabase/pull/38289#discussion_r1472653468

Access is O(1) because of using map

### DIFF
--- a/frontend/src/metabase/parameters/utils/dashboards.ts
+++ b/frontend/src/metabase/parameters/utils/dashboards.ts
@@ -90,9 +90,7 @@ export function isDashboardParameterWithoutMapping(
   return parameterExistsOnDashboard && !parameterHasMapping;
 }
 
-function getMappings(
-  dashcards: DashboardCard[],
-): [ExtendedMapping[], Map<string, Map<string, ExtendedMapping>>] {
+function getMappings(dashcards: DashboardCard[]): ExtendedMapping[] {
   const extendedParameterMappings = dashcards.flatMap(dashcard => {
     const { parameter_mappings, card, series } = dashcard;
     const cards = [card, ...(series || [])];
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/parameters/utils/dashboards.ts
```diff
@@ -109,17 +109,41 @@ function getMappings(dashcards: DashboardCard[]): ExtendedMapping[] {
...
+  for (const mapping of mappings) {
+    if (!mappingsByParameterId.has(mapping.parameter_id)) {
+      mappingsByParameterId.set(mapping.parameter_id, new Map());
+    }
+
+    const targetKey = Symbol.for(mapping.target.toString());
+    if (!mappingsByParameterId.get(mapping.parameter_id)?.has(targetKey)) {
+      mappingsByParameterId.get(mapping.parameter_id)?.set(targetKey, mapping);
+    }
+  }
```
oleggromov, line 129: Nitpick: it appears to me that this is unnecessarily verbose.
I would suggest using a plan object, avoid symbols and find a way of doing something in the spirit of `assocIn(mappingsByParameterId, [mapping.parameter_id, mapping.target.toString()], mapping)` within the loop.

It's a matter of taste though 😉

oleggromov (approved):
Left a tiny suggestion, but otherwise LGTM!

frontend/src/metabase/parameters/utils/dashboards.ts
kulyk: +1, I'd also prefer a plain object unless `Map` provides some significant performance win of something

frontend/src/metabase/parameters/utils/dashboards.ts
```diff
@@ -109,17 +109,41 @@ function getMappings(dashcards: DashboardCard[]): ExtendedMapping[] {
...
 ): UiParameter[] {
   const { parameters, dashcards } = dashboard;
-  const mappings = getMappings(dashcards as DashboardCard[]);
+  const mappings = getMappingsByParameterId(dashcards as DashboardCard[]);
   const uiParameters: UiParameter[] = (parameters || []).map(parameter => {
     if (isFieldFilterParameter(parameter)) {
-      return buildFieldFilterUiParameter(parameter, mappings, metadata);
+      return buildFieldFilterUiParameter(
+        parameter,
+        [...(mappings.get(parameter.id)?.values() ?? [])],
```
kulyk, line 144: nit: there are many things going on here, and it feels a bit hard to parse
might be worth unpacking it into several lines

kulyk (approved):
LGTM, just two code-style comments

uladzimirdev:
the main problem was resolved in a different way, this PR doesn't change a lot, so closing

PR #38327, opened 2024-01-31
HK: Switch to using a single action for updating static embedding enabled and params

oleggromov:
WIP

### DIFF
--- a/frontend/src/metabase/dashboard/actions/sharing.ts
+++ b/frontend/src/metabase/dashboard/actions/sharing.ts
@@ -1,15 +1,22 @@
 import type { Dashboard, DashboardId } from "metabase-types/api";
-import type { Dispatch, EmbedOptions } from "metabase-types/store";
-import { createAction } from "metabase/lib/redux";
+import type { Dispatch } from "metabase-types/store";
+import { createAction, createThunkAction } from "metabase/lib/redux";
 
 import { SIDEBAR_NAME } from "metabase/dashboard/constants";
 
 import { DashboardApi } from "metabase/services";
 
+import type { EmbeddingParameters } from "metabase/public/lib/types";
 import { setSidebar, closeSidebar } from "./ui";
 
 type DashboardIdPayload = {
   id: DashboardId;
+};
+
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/dashboard/actions/sharing.ts
```diff
@@ -24,20 +31,23 @@ export const setSharing = (isSharing: boolean) => (dispatch: Dispatch) => {
...
+export const PUBLISH_EMBEDDING = "metabase/dashboard/PUBLISH_EMBEDDING";
+export const publishEmbedding = createThunkAction(
+  PUBLISH_EMBEDDING,
+  ({ id, enable_embedding, embedding_params }: PublishEmbeddingPayload) =>
+    async () => {
+      const result = await DashboardApi.update({
+        id,
+        enable_embedding,
+        embedding_params,
+      });
```
oleggromov, line 43: I was thinking about it and couldn't come up with any reason to exclude `embedding_parameters` when un-publishing an embedding.

Even more so, if I understand the logic correctly, we can change parameters and then un-publish, in which case we wouldn't even persist the changes? 

Might be wrong here but the less conditionals, the less confusing it is I suppose.

frontend/src/metabase/dashboard/reducers.js
```diff
@@ -131,21 +130,15 @@ const dashboards = handleActions(
...
-        ),
+    [PUBLISH_EMBEDDING]: {
+      next: (state, { payload: { id, enable_embedding, embedding_params } }) =>
+        produce(state, draftState => {
+          const draftDashboard = draftState[id];
+          draftDashboard.enable_embedding = enable_embedding;
+          if (enable_embedding) {
+            draftDashboard.embedding_params = embedding_params;
+          }
+        }),
```
oleggromov, line 141: LMK if there's a cleaner way of doing two `assocIn`s, where one is conditional.

oleggromov:
Turns out, although this might be a useful optimisation, it is not really required for the https://github.com/metabase/metabase/issues/36524

PR #38328, opened 2024-01-31
Move `canExploreResults` out of the Question prototype

nemanjaglumac:
This PR moves `canExploreResults` method out of the Question prototype, and into the only component it was using it. To be more precise, the method now lives in a utilities file related to the component.

For the sake of simplicity and the smaller diff, this PR does NOT introduce a new folder and file structure. That will be a subject of a subsequent PR.

Hence, we have maybe unfortunate name for the util file `ViewHeaderUtils.js`.

All unit tests have been transferred in almost identical format. The only difference is that we're now dealing with a function that takes a question as a param, rather than a method on the question itself.


kamilmielnik (approved):
:+1: 

----

> For the sake of simplicity and the smaller diff, this PR does NOT introduce a new folder and file structure. That will be a subject of a subsequent PR.

For the future: IMO it's better to do all this in 1 PR - it's still going to be small, and we wouldn't have this awkward intermediate state with `ViewHeaderUtils.js` file.

frontend/src/metabase/query_builder/components/view/ViewHeaderUtils.js
```diff
@@ -0,0 +1,19 @@
+import * as Lib from "metabase-lib";
+
+/**
+ * We can only "explore results" (i.e. create new questions based on this one)
```
uladzimirdev, line 4: we use capitalized file naming only for components, utils usually are placed in `utils` file

nemanjaglumac: Thanks.
Will follow up soon.

PR #38333, opened 2024-01-31
Migrate download perm enforcement to data_permissions

noahmoss:
Updates download permission enforcement to read from the `data_permissions` table.

### DIFF
--- a/src/metabase/models/data_permissions.clj
+++ b/src/metabase/models/data_permissions.clj
@@ -438,7 +438,8 @@
                                           :perm_value value
                                           :db_id      db-id})
       (when (= [:perms/data-access :block] [perm-type value])
-        (set-database-permission! group-or-id db-or-id :perms/native-query-editing :no)))))
+        (set-database-permission! group-or-id db-or-id :perms/native-query-editing :no)
+        (set-database-permission! group-or-id db-or-id :perms/download-results :no)))))
 
 (mu/defn set-table-permissions!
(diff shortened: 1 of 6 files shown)


src/metabase/models/data_permissions.clj
```diff
@@ -237,6 +237,34 @@
...
+                      :join [[:permissions_group :pg] [:= :pg.id :pgm.group_id]
+                             [:data_permissions :p]   [:= :p.group_id :pg.id]]
+                      :where [:and
+                              [:= :pgm.user_id user-id]
+                              [:= :p.perm_type (u/qualified-name :perms/download-results)]
+                              [:= :p.db_id database-id]]})
+          value-by-group
+          (-> (group-by :group_id perm-values)
+              (update-vals (fn [perms]
+                             (let [values (set (map :value perms))]
```
johnswanson, line 261: I added a function above called `coalesce-less-restrictive`, which I think is doing the same thing you're doing here?

johnswanson (approved):

src/metabase/models/data_permissions.clj
noahmoss: Where did you add it? Don't see it on this branch

noahmoss: oh it was `coalesce-most-restrictive`. updated thanks!

johnswanson: Oops, sorry I missed your earlier message. 🤦

PR #38336, opened 2024-01-31
[QC] Remove `hasData` method from the `StructuredQuery` prototype

nemanjaglumac:
This PR is just a tiny cleanup step. The Boy Scouting rule applied.
Logically the code should behave the same as before, but we're removing a deprecated method from the StructuredQuery prototype.

### DIFF
--- a/frontend/src/metabase-lib/queries/StructuredQuery.ts
+++ b/frontend/src/metabase-lib/queries/StructuredQuery.ts
@@ -321,7 +321,7 @@ class StructuredQuery extends AtomicQuery {
   }
 
   isValid() {
-    if (!this.hasData()) {
+    if (!this.table()) {
       return false;
     }
 
@@ -386,13 +386,6 @@ class StructuredQuery extends AtomicQuery {
       console.warn("Error thrown while validating clause", clause, e);
       return false;
     }
-  }
-
-  /**
-   * @deprecated use MLv2
-   */
-  hasData() {
-    return !!this.table();
   }
 
   hasAnyClauses() {


frontend/src/metabase-lib/queries/StructuredQuery.ts
```diff
@@ -321,7 +321,7 @@ class StructuredQuery extends AtomicQuery {
   }
 
   isValid() {
-    if (!this.hasData()) {
+    if (!this.table()) {
```
kamilmielnik, line 324: 1. We're losing semantics (`hasData`) with this change.
2. Inlining a deprecated function into another one, makes the other function deprecated.

kamilmielnik (commented):
I don't think we're getting much value with this change, it can stay as is, and we can remove `StructuredQuery.prototype.hasData` when we remove of `StructuredQuery.prototype.isValid`.

BTW, maybe can we remove `StructuredQuery.prototype.isValid` already?

frontend/src/metabase-lib/queries/StructuredQuery.ts
nemanjaglumac: 1. we can solve this by binding the expression to the const `hasData`
2. `isValid()` should be deprecated, imo.

I'll see if there's a way to migrate `isValid()`, but the most problematic part in all this is `getStructuredQueryTable` helper.

kamilmielnik: > I'll see if there's a way to migrate `isValid()`, but the most problematic part in all this is `getStructuredQueryTable` helper.

I don't think it deserves migration. We should just remove it once it's not used anymore.

nemanjaglumac: @kamilmielnik https://github.com/metabase/metabase/pull/38522 🎉

nemanjaglumac:
Closing in favor of https://github.com/metabase/metabase/pull/38522

PR #38337, opened 2024-01-31
Implement filtering cards by segment or metric ID

### LINKED ISSUE #37688: [MLv2] [BE] Add `metric-id` and `segment-id` to `GET /api/card` to return only cards that use this segment/metric
Blocks https://github.com/metabase/metabase/issues/37511
Blocks https://github.com/metabase/metabase/issues/37510

We need to show questions that use this segment or metric in data reference. Currently the FE loads all questions by calling `GET /api/card` and filters them out on the FE. 

Screenshot 2024-01-15 at 13 05 09

With MLv2 we don't have a function to do that and it seems incorrect to load **all** cards in this case. To make it work in a proper way, let's add `metric-id` and `segment-id` parameters to `GET /api/card` that would return cards that are available to the current user that use this metric or segment. We would never pass both `metric-id` and `segment-id` at once, however it's possible that a card uses both metrics and segments at the same time.

metamben:
Fixes #37688.

Access is granted if the user can read the database of the segment/metric.


src/metabase/api/card.clj
```diff
@@ -104,6 +106,24 @@
        ;; now check if model-id really occurs as a card ID
        (filter (fn [card] (some #{model-id} (-> card :dataset_query query/collect-card-ids))))))
 
+(defn- cards-for-segment-or-metric
+  [model-type model-id]
+  (->> (t2/select :model/Card {:where [:like :dataset_query (str "%" (name model-type) "%" model-id "%")]})
+       ;; now check if the segment/metric with model-id really occurs in a filter/aggregation expression
+       (filter (fn [card]
+                 (let [query (-> card :dataset_query lib.convert/->pMBQL)]
```
metamben, line 114: This is relying on `dataset_query` having been normalized when written into the DB.

src/metabase/api/card.clj
```diff
@@ -113,21 +133,31 @@
...
                                                          (name f)))
       (case f
-        :database    (api/read-check Database model_id)
-        :table       (api/read-check Database (t2/select-one-fn :db_id Table, :id model_id))
-        :using_model (api/read-check Card model_id)))
+        :database      (api/read-check Database model_id)
+        :table         (api/read-check Database (t2/select-one-fn :db_id Table, :id model_id))
+        :using_model   (api/read-check Card model_id)
+        :using_metric  (api/read-check Database (db-id-via-table :metric model_id))
+        :using_segment (api/read-check Database (db-id-via-table :segment model_id))))
```
metamben, line 160: Following the pattern used for checking `:table`.

src/metabase/api/card.clj
bshepherdson: A single conversion shouldn't be bad, especially when compared with the overhead of DB and network calls.

src/metabase/lib/query.cljc
```diff
@@ -188,3 +188,38 @@
...
+
+(defn- occurs-in-stage-clause?
+  "Tests whether predicate `pred` is true for an element of clause `clause` of `query-or-join`.
+  The test is transitive over joins."
+  [query-or-join clause pred]
+  (boolean
+   (some (fn [stage]
+           (or (some pred (clause stage))
+               (some #(occurs-in-stage-clause? % clause pred) (:joins stage))))
+         (:stages query-or-join))))
```
bshepherdson, line 209: Not for this PR, but I don't think this is the first time we've wanted a `lib.walk` or `lib.match` library to capture these recursive explorations. We'll probably end up building something like that for the QP eventually.

bshepherdson (approved):

src/metabase/lib/query.cljc
metamben: Let's see a few more examples. I grew vary of walking and matching: I had to replace them by custom code because they were the bottlenecks. (I've seen normalization dominate the time needed for I/O.)

Issue #38122, opened 2024-01-25
Trend chart gives misleading output for some date combinations


Tony-metabase:
### Describe the bug

Trend chart gives mixed outputs when you run an SQL query and shows date's which aren't even there. 

### To Reproduce

1. Go to New -> SQL Question:

```
Select '2023-12-31'::date, 28826833
UNION
Select '2022-12-31'::date, 33872127
```

2. Change the Visualization to a Trend Chart:

<img width="1505" alt="image" src="https://github.com/metabase/metabase/assets/110378427/e4eda18c-2d0f-451d-bb3f-fa2e8fadc0ca">

3. Notice that previously in 1.47 it used to only show the year `2023` now it shows `2023-2024` which is confusing 

### Expected behavior

Shows 2023 since the comparison year is 2023

### Logs

None that are relevant

### Information about your Metabase installation

```JSON
1.48.3 and master
```

### Severity

Even though this is an aesthetic bug it broke the implementation of a number of customers that had  been already using the Trend Chart

JesseSDevaney:
Closed by:
- https://github.com/metabase/metabase/pull/38344
- https://github.com/metabase/metabase/pull/38383

PR #38344, opened 2024-02-01
Display full-dates for trend charts backed by native queries

JesseSDevaney:
### Problem

For native queries, the chosen display date, for an aggregated value across range a range of dates, is up to the creator. This is unlike the query builder which chooses the display date (across an aggregated range of dates) to be at the start of the range. Trend charts expects a date for a given `dateUnit` to be at the start of the range as well.
- Examples:
  - Aggregating by month
    - Query Builder: Start of the month (Jan. 1, Feb. 1, Mar. 1, ...)
    - Native Query: Whatever the creator decides (Jan 1 or Jan 31, Feb 1 or Feb 28, ...)
  - Aggregating by year
    - Query Builder: Start of the year (Jan 1 2022, Jan 1 2023, ...)
    - Native Query: Whatever the creator decides (Jan 1 or Dec 31 2022, Jan 1 or Dec 31 2023, ...)

---
### Desired Behavior/Acceptance Criteria
**Solution Behaviors**
- For native queries
   image
  - [ ] always display the full-date (disregard the `dateUnit` only for formatting)
  - [ ] keep the inferred `dateUnit` behavior to still allow `periods ago` comparisons
- [ ] All other behaviors should be unaltered

**Validation Tests**
- [ ] (`compute.unit.spec.js`) Native queries always display the full-date regardless of the inferred `dateUnit`

---
### How to Manually Verify the Solution
1. Follow the steps listed in the issue
   - https://github.com/metabase/metabase/issues/38122
2. See that the full-dates are displayed instead of the condensed `dateUnit` versions
3. See that there are still options for `previous period` and `periods ago` comparisons
4. Create a regular (non-native) query and see that trend charts displays the condensed versions of the dates

---
**Context**
- Will close after backport: https://github.com/metabase/metabase/issues/38122


frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -379,11 +383,24 @@ function computeComparisonPeriodsAgo({
...
+    dateUnitsAgo,
+    dimensionColIndex,
+    metricColIndex,
+    nextValueRowIndex,
+    rows,
+  });
+
+  const prevDate = !isEmpty(rowPeriodsAgo)
+    ? rowPeriodsAgo[dimensionColIndex]
+    : computedPrevDate;
```
JesseSDevaney, line 403: Since we no longer want the condensed versions of the dates and the dates for native queries may not align perfectly with the `dateUnit` interval, we should get the actual date and display that (so we are not displaying an incorrect date as in the image below).

**Prior to Change**
Pasted image 20240131183512
- See that the display date is `December 31, 2021`, but should be `October 11, 2021`


**After Change**
image

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -431,14 +439,14 @@ function getRowOfPeriodsAgo({
   // skip the latest element since that is our current value
   const searchIndexStart = nextValueRowIndex - 1;
   if (searchIndexStart < 0) {
-    return -1;
+    return undefined;
```
JesseSDevaney, line 442: This was inconsistent with the rest of the functions return types. It could be causing errors for some people who look at dates in the past that are close to the start of their dataset, since the consuming function thinks it will either have a value or be `undefined`. `-1` is neither of those.

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
```diff
@@ -431,14 +439,14 @@ function getRowOfPeriodsAgo({
...
   }
 
   // only look dateUnitsAgo elements (dates) into the past,
   // since looking any further would automatically result in a date before
   // X periods ago and any prior dates would be further beyond our desired
   // comparison date
-  const lastCandidateIndex = nextValueRowIndex - 1 - (dateUnitsAgo - 1);
+  const lastCandidateIndex = searchIndexStart - (dateUnitsAgo - 1);
```
JesseSDevaney, line 449: No logic change here. Just taking the `searchIndexStart` variable from up above, since it represents this value.

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -2109,6 +2119,68 @@ describe("SmartScalar > compute", () => {
         expect(comparisonDisplay.comparisonValue).toBe("105.0k");
       });
     });
+
+    describe("bug fixes", () => {
+      describe("should display full dates for native queries (issue #38122)", () => {
```
kulyk, line 2124: I don't think we're grouping tests added alongside bug fixes anywhere else, let's just keep the test?

frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.unit.spec.js
```diff
@@ -2109,6 +2119,68 @@ describe("SmartScalar > compute", () => {
...
+                  metricValue: 200,
+                }),
+              },
+            },
+          },
+        ];
+
+        it.each(testCases)(
+          "$description",
+          ({ rows, expected, periodsAgo, dateUnit }) => {
```
kulyk, line 2171: There's no reason to use `describe` + `it.each` for a single test case, can we make it a single `it` block?

kulyk (approved):
Code looks good! Left a suggestion around testing

PR #38353, opened 2024-02-01
Improve CSV upload filename uniqueness

crisptrutski:
### Description

Before this test we needed to work around collisions when generating tablenames uniquely.

This change makes it impossible to generate the same "unique" name twice within the same process, although all bets are off across multiple servers.

As a drive-by improvement I also fix the helper throwing an exception if given a blank string.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -304,15 +304,35 @@
 ;;;; |  Parsing values  |
 ;;;; +------------------+
 
+(import 'java.time.LocalDateTime)
+(import 'java.time.temporal.ChronoUnit)
+
+(def ^:private last-timestamp (atom nil))
+
+(defn monotonic-now
+  "Return an adjusted version of the current time, so that it is guaranteed to be monotonic at a 1 second granularity."
+  []
+  ^LocalDateTime
+  (swap! last-timestamp
+         (fn [^LocalDateTime prev-timestamp]
+           (let [now (LocalDateTime/now)]
+             (if (nil? prev-timestamp)
+               now
+               (let [next-second (.truncatedTo (.plus prev-timestamp 1 ChronoUnit/SECONDS) ChronoUnit/SECONDS)]
+                 (if (.isAfter now next-second)
+                   now
+                   next-second)))))))
+
 (defn- unique-table-name
-  "Append the current datetime to the given name to create a unique table name. The resulting name will be short enough for the given driver (truncating the supplised `table-name` if necessary)."
+  "Append the current datetime to the given name to create a unique table name. The resulting name will be short enough for the given driver (truncating the supplied `table-name` if necessary)."
   [driver table-name]
   (let [time-format                 "_yyyyMMddHHmmss"
-        acceptable-length           (min (count table-name)
-                                         (- (driver/table-name-length-limit driver) (count time-format)))
-        truncated-name-without-time (subs (u/slugify table-name) 0 acceptable-length)]
+        slugified-name               (or (u/slugify table-name) "")
+        max-length                  (- (driver/table-name-length-limit driver) (count time-format))
+        acceptable-length           (min (count slugified-name) max-length)
+        truncated-name-without-time (subs slugified-name 0 acceptable-length)]
     (str truncated-name-without-time
-         (t/format time-format (t/local-date-time)))))
+         (t/format time-format (monotonic-now)))))
 
 (def ^:private max-sample-rows "Maximum number of values to use for detecting a column's type" 1000)
 
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -304,24 +304,20 @@
 ;;;; |  Parsing values  |
 ;;;; +------------------+
 
-(import 'java.time.LocalDateTime)
-(import 'java.time.temporal.ChronoUnit)
+(def ^:private last-timestamp (atom (t/local-date-time)))
 
-(def ^:private last-timestamp (atom nil))
+(set! *warn-on-reflection* true)
 
-(defn monotonic-now
-  "Return an adjusted version of the current time, so that it is guaranteed to be monotonic at a 1 second granularity."
+(defn strictly-monotonic-now
+  "Return an adjusted version of the current time, that it is guaranteed to never repeat the last second."
   []
-  ^LocalDateTime
   (swap! last-timestamp
-         (fn [^LocalDateTime prev-timestamp]
-           (let [now (LocalDateTime/now)]
-             (if (nil? prev-timestamp)
-               now
-               (let [next-second (.truncatedTo (.plus prev-timestamp 1 ChronoUnit/SECONDS) ChronoUnit/SECONDS)]
-                 (if (.isAfter now next-second)
-                   now
-                   next-second)))))))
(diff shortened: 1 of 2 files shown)


src/metabase/upload.clj
```diff
@@ -304,15 +304,35 @@
 ;;;; |  Parsing values  |
 ;;;; +------------------+
 
+(import 'java.time.LocalDateTime)
+(import 'java.time.temporal.ChronoUnit)
```
crisptrutski, line 308: Oops, didn't expect my IDE to take a shortcut like this. I'll fix this up after the reviews are in.

src/metabase/upload.clj
```diff
@@ -304,15 +304,35 @@
...
+  "Return an adjusted version of the current time, so that it is guaranteed to be monotonic at a 1 second granularity."
+  []
+  ^LocalDateTime
+  (swap! last-timestamp
+         (fn [^LocalDateTime prev-timestamp]
+           (let [now (LocalDateTime/now)]
+             (if (nil? prev-timestamp)
+               now
+               (let [next-second (.truncatedTo (.plus prev-timestamp 1 ChronoUnit/SECONDS) ChronoUnit/SECONDS)]
+                 (if (.isAfter now next-second)
```
tsmacdonald, line 322: I would use `java-time` for all of this (e.g., `t/after?` instead of `.isAfter`, `t/plus` instead of `.plus`, etc.)

tsmacdonald: It's documentation is kind of a hassle, but if you `:require [java-time.api :as t]` and have good auto-completion you should be able to find what you're after.

tsmacdonald: If you initialize `last-timestamp` to `(t/local-date-time)` instead of `nil` then all of these `if`s can condense down to

```clojure
(swap! last-timestamp (fn [prev] (t/max (t/plus prev (t/seconds 1)) (t/local-date-time))))
```

(maybe add a newline or two to make it look nicer)

src/metabase/upload.clj
```diff
@@ -304,15 +304,35 @@
...
+
 (defn- unique-table-name
-  "Append the current datetime to the given name to create a unique table name. The resulting name will be short enough for the given driver (truncating the supplised `table-name` if necessary)."
+  "Append the current datetime to the given name to create a unique table name. The resulting name will be short enough for the given driver (truncating the supplied `table-name` if necessary)."
   [driver table-name]
   (let [time-format                 "_yyyyMMddHHmmss"
-        acceptable-length           (min (count table-name)
-                                         (- (driver/table-name-length-limit driver) (count time-format)))
-        truncated-name-without-time (subs (u/slugify table-name) 0 acceptable-length)]
+        slugified-name               (or (u/slugify table-name) "")
```
tsmacdonald, line 330: when is `table-name` nil?

src/metabase/upload.clj
```diff
@@ -304,15 +304,35 @@
...
+        max-length                  (- (driver/table-name-length-limit driver) (count time-format))
+        acceptable-length           (min (count slugified-name) max-length)
+        truncated-name-without-time (subs slugified-name 0 acceptable-length)]
     (str truncated-name-without-time
-         (t/format time-format (t/local-date-time)))))
+         (t/format time-format (monotonic-now)))))
```
tsmacdonald, line 335: I've got my pedant hat on now, but "monotonically increasing" really means "non-decreasing" and doesn't guarantee unique values. What about renaming the function to `unique-now` or something like that?

test/metabase/upload_test.clj
```diff
@@ -412,7 +412,11 @@
     (testing "File name is slugified"
       (is (=? #"my_file_name_\d+" (@#'upload/unique-table-name driver/*driver* "my file name"))))
     (testing "semicolons are removed"
-      (is (nil? (re-find #";" (@#'upload/unique-table-name driver/*driver* "some text; -- DROP TABLE.csv")))))))
+      (is (nil? (re-find #";" (@#'upload/unique-table-name driver/*driver* "some text; -- DROP TABLE.csv")))))
+    (testing "No collisions"
+      (let [n 50
+            names (repeatedly n (partial #'upload/unique-table-name driver/*driver* ""))]
+        (is (= 50 (count (distinct names))))))))
```
tsmacdonald, line 419: There should be a test for what would've been a timestamp collision. There are test helpers to help fake the clock if that's useful.

tsmacdonald (changes requested):
**Blocking:** use `java-time`

**Blocking:** add tests

**Recommended:** use my terser logic

**Just an idea, I'm not going to fight for it:** renaming `monotonic-now`

tsmacdonald:
Is there a bug report this is tied to?

tsmacdonald:
>  Chris just spotted this in a PR review and was nice enough to solve it. Context is here

Awesome :+1: :+1:

src/metabase/upload.clj
crisptrutski: > If you initialize last-timestamp to (t/local-date-time)

I initially considered and decided against this, to avoid possiblty rounding up the first time it's called. It isn't really possible and doesn't matter in the slightest though, so cool let's simplify.

> I would use java-time for all of this

I'm not allergic to interop unless a Clojure wrapper gives me cross runtime support or much higher abstracts. Not 100% convinced in this library -its largely at the same level of abstraction as the Java classes, but less performant, and as you say its docs and generated source aren't very clear. I needed to google how it expects time units for truncate.

Using `t/max` is very nice and that sells it for me. Also happy to go with the flow if we've standardized on this library in general.

src/metabase/upload.clj
crisptrutski: I thought about calling it `strictly-monotonic`, but it's fairly common in software engineering to just assume the adjective - see the linux monotonic clock or logical monotonicity (CALM). I'd rather be more precise than more vague, so I've just added the adjective.

test/metabase/upload_test.clj
crisptrutski: I'm not sure I follow - this test is totally sufficient in my opinion. We want to assert that the within tests it is not necessary to call `sleep` between calling any methods that will rely on the table name, and the repetition and direct low level call give us strong confidence.

Were you imagining testing a higher level workflow? In my mind that's unnecessary - the "proof" is the lack of test flakes or slow tests. 

I'm not keen on faking the clock, that feels quite invasive. This test already fails without the changes and I feel it is sufficient. Have I swayed you?

src/metabase/upload.clj
crisptrutski: Slugify returns nil for the empty string. We probably don't hit this case, but it bothered me to have a "partial" function when it's trivial to make it total.

src/metabase/upload.clj
tsmacdonald: that might be a good thing to bring up at the eng guild meeting—I'm less an advocate for java-time in particular and more for standardizing on One Way for handling time (of which java-time is the latest attempt)

test/metabase/upload_test.clj
tsmacdonald: Sorry, that's my bad

tsmacdonald (approved):
Nice one, thanks Chris!

PR #38357, opened 2024-02-01
FE - Temporarily mock metrics CRUD endpoints

### LINKED ISSUE #38262: FE - Temporarily mock metrics CRUD endpoints
BE team won't be able to start working on #37348 or any other tasks in this step until at least next week (05-09 Feb 2024) so we need to temporarily mock related endpoints to enable us to make progress on Metrics v2 project.

Please use `metrics-v2` as target branch.

kamilmielnik:
Closes #38262

### Description

This PR mocks FE endpoint handlers to pretend that any question which has a name starting with "Metric" (case-insensitive) is a metric.
In other words: you can prefix your question names with "Metric" or "metric" to pretend they're metrics and not questions.
Be sure to pass correct `type: "metric"` to update/create endpoints though.


frontend/src/metabase/entities/questions.js
```diff
@@ -26,6 +27,62 @@ const Questions = createEntity({
...
+   * Any question with `type: "metric"` will be passed to API as `type: "question"`.
+   * Same goes for any questions with a name starting with "Metric" (case-insensitive).
+   */
+  api: {
+    get: async payload => {
+      const get = GET("/api/card/:id");
+      const result = await get(payload);
+
+      if (result.name.toLowerCase().startsWith("metric")) {
+        return { ...result, type: "metric", dataset: false };
```
nemanjaglumac, line 42: `dataset: false` is still only a temporary measure during the migration period, right?

frontend/src/metabase/entities/questions.js
```diff
@@ -26,6 +27,62 @@ const Questions = createEntity({
...
+      });
+    },
+
+    create: async payload => {
+      const create = POST("/api/card");
+
+      if (payload.type === "metric") {
+        const tweakedPayload = { ...payload, type: "question", dataset: false };
+        const result = await create(tweakedPayload);
+        return { ...result, type: "metric" };
```
nemanjaglumac, line 67: Why do we have to first create a tweakedPayload with a type "question", only to override it with a type "metric" in the last (return) step?

nemanjaglumac (approved):
Left some questions inline, but the change looks straightforward.
There are still some E2E tests failing, though - probably not related

frontend/src/metabase/entities/questions.js
kamilmielnik: Hopefully. We will have both `dataset` and `type`, but we should not be using `dataset` anymore.

frontend/src/metabase/entities/questions.js
kamilmielnik: Let's say we want to create a metric.
We will try to call this `create` endpoint with some payload which will include `type: "metric"`.
Backend won't accept this, as it's not implemented yet.
So this override will actually make FE send `type: "question"` to BE, but when the result comes back, we still want FE to believe it's a "metric", so we have to update the `type`.

frontend/src/metabase/entities/questions.js
```diff
@@ -26,6 +27,62 @@ const Questions = createEntity({
...
+    update: async payload => {
+      const update = PUT("/api/card/:id");
+
+      if (payload.type === "metric") {
+        const tweakedPayload = { ...payload, type: "question", dataset: false };
+        const result = await update(tweakedPayload);
+        return { ...result, type: "metric" };
+      }
+
+      return await update(payload);
```
uladzimirdev, line 82: idk we do we use `return await` without try catch, but seems there shouldn't be any difference comparing with just returning a promise

uladzimirdev (approved):

kamilmielnik:
> There are still some E2E tests failing, though - probably not related

Good point. I disabled the overrides in tests so that we can have green CI: 97f257720be0c1ed81e595ac7c3500cc95936682

nemanjaglumac:
> > There are still some E2E tests failing, though - probably not related
> 
> Good point. I disabled the overrides in tests so that we can have green CI: [97f2577](https://github.com/metabase/metabase/commit/97f257720be0c1ed81e595ac7c3500cc95936682)

You might also need to use `isCypressActive`.

Issue #38225, opened 2024-01-29
Extreme slowness while editing parameters on a dashboard


darksciencebase:
### Describe the bug

Broken column mappings and very slow-moving interaction with the dashboard, possibly related. Please see more details and a loom recorded on stats in [this thread](https://metaboat.slack.com/archives/C05MPF0TM3L/p1706535171602269).

### To Reproduce

in the loom

### Expected behavior

We shouldn't be silently breaking mappings, and stuff should be snappier

### Information about your Metabase installation

```JSON
stats at 6c9d52e
```

### Severity

potentially P1 if we're producing incorrect data

uladzimirdev:
I'm checking perf issue of frontend part

bshepherdson:
There's apparently an issue with edited models and the field refs not lining up anymore.

We're making a request for eg. `/api/dashboard/1907/dashcard/13045/card/14066/query` (`Total ARR`, the top-left card). 

Card 14066 adds an aggregation and breakout on top of card 14023, a native model.
* 14023 has the `recognized_at` field with ID and `field_ref: ["field", 575819, {"temporal-unit": "default"}]`
* 14066 still has the ID, but `field_ref: ["field", 575819, {"base-type": "type/Date", "temporal-unit": "month"}]`
* the parameter from the dashboard is mapped to `target: ["dimension", ["field", 575819, {"temporal-unit": "default"}]]`, which matches the inner 14023 field but not the outer 14066 one.

So if we think the models changed lately, that's likely the issue causing `Unknown Field`, that the field refs for the parameter mapping don't match correctly.

darksciencebase:
cc @NevRA

uladzimirdev:
For clarification: 

This issue covers 2 different problems
1. As Braden highlighted, wrong parameter mapping is generated, which leads to `Unknown field` problems
2. Slowness of the editing dashboard

PR #38360, opened 2024-02-01
[Dashboard] Do not update dashboard settings on every keypress in Text

uladzimirdev:

### Description

We used to update dashboard viz settings on every keystroke, but keeping in mind slowness of re-rendering, we can avoid unnecessary re-renderings and update a dashboard only when you stop typing and focus out of the input.

### How to verify

1. Open a dashboard with many cards
2. go to edit mode, add Text
3. start typing something -> you'll be able to type as usual without any delay

### Demo

Same fix as for Heading
https://www.loom.com/share/ebe74cebe016440d837c1cc5f789b821

### Checklist

### DIFF
--- a/frontend/src/metabase/visualizations/visualizations/Text/Text.unit.spec.tsx
+++ b/frontend/src/metabase/visualizations/visualizations/Text/Text.unit.spec.tsx
@@ -153,6 +153,24 @@ describe("Text", () => {
         userEvent.click(screen.getByTestId("editing-dashboard-text-preview"));
         expect(screen.getByDisplayValue("text text text")).toBeInTheDocument();
       });
+
+      it("should call onUpdateVisualizationSettings on blur", () => {
+        const mockOnUpdateVisualizationSettings = jest.fn();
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/visualizations/visualizations/Text/Text.unit.spec.tsx
```diff
@@ -153,6 +153,24 @@ describe("Text", () => {
...
+        setup(options);
+
+        userEvent.click(screen.getByTestId("editing-dashboard-text-preview"));
+        userEvent.type(screen.getByRole("textbox"), "foo");
+        userEvent.tab();
+
+        expect(mockOnUpdateVisualizationSettings).toHaveBeenCalledWith({
+          text: "textfoo",
+        });
+      });
```
kulyk, line 173: If `onUpdateVisualizationSettings` can have major performance impact, it could be beneficial to assert its only called once here

kulyk (approved):

PR #38361, opened 2024-02-01
[Dashboard] Do not update dashboard settings on every keypress in Heading

uladzimirdev:

### Description

We used to update dashboard viz settings on every keystroke, but keeping in mind slowness of re-rendering, we can avoid unnecessary re-renderings and update a dashboard only when you stop typing and focus out of the input.

### How to verify

1. Open a dashboard with many cards
2. go to edit mode, add Heading
3. start typing something -> you'll be able to type as usual without any delay

### Demo

https://www.loom.com/share/ebe74cebe016440d837c1cc5f789b821

### Checklist

### DIFF
--- a/frontend/src/metabase/visualizations/visualizations/Heading/Heading.unit.spec.tsx
+++ b/frontend/src/metabase/visualizations/visualizations/Heading/Heading.unit.spec.tsx
@@ -206,6 +206,26 @@ describe("Text", () => {
           screen.getByDisplayValue("Variable: {{variable}}"),
         ).toBeInTheDocument();
       });
+
+      it("should call onUpdateVisualizationSettings on blur", () => {
+        const mockOnUpdateVisualizationSettings = jest.fn();
+        const options = {
+          settings: getSettingsWithText("text"),
+          isEditing: true,
+          onUpdateVisualizationSettings: mockOnUpdateVisualizationSettings,
+        };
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/visualizations/visualizations/Heading/Heading.unit.spec.tsx
```diff
@@ -206,6 +206,26 @@ describe("Text", () => {
...
+        userEvent.click(
+          screen.getByTestId("editing-dashboard-heading-preview"),
+        );
+        userEvent.type(screen.getByRole("textbox"), "foo");
+        userEvent.tab();
+
+        expect(mockOnUpdateVisualizationSettings).toHaveBeenCalledWith({
+          text: "textfoo",
+        });
+      });
```
kulyk, line 228: If `onUpdateVisualizationSettings` can have major performance impact, it could be beneficial to assert its only called once here

kulyk (approved):

frontend/src/metabase/visualizations/visualizations/Heading/Heading.unit.spec.tsx
uladzimirdev: f4c7e42778bcc7f5665f54cbabe34083f040b0d9

Issue #36359, opened 2023-12-04
Let dashboard filters only apply to select tabs


cdeweyx:
## Context
[Product doc](https://www.notion.so/metabase/Let-dashboard-filters-only-apply-to-select-tabs-8c8bbdcab61f49e494a988e01592bd21)
Epic: https://github.com/metabase/metabase/issues/36010

## Tasks
...

PR #38368, opened 2024-02-01
Hide unmapped filters for public/embedded dashboards

kulyk:
Followup on #37711 (issue #36359)

After recent changes, we're now hiding a dashboard filter when it's not mapped to at least one card on the current tab
This was meant to work the same for public and statically embedded dashboards, but I missed that this code isn't shared between public/embedded and in-app dashboards.

### To verify

1. Sign in as admin
2. Go to `/admin/settings/public-sharing` and enable public sharing
3. Create a dashboard with a couple of tabs, several filters, and dashboards. Ensure you have a filter mapped to at least one card per tab, a filter used by a single tab, and a filter not used by any card at all.
4. While the dashboard page, click the sharing icon (in the top right), generate a public link and open it
5. Click through tabs, ensure filter are hidden on tabs where they're not connected to any cards


e2e/test/scenarios/embedding/shared/embedding-dashboard.js
```diff
@@ -71,9 +71,30 @@ const sourceFilter = {
 
 const userFilter = { name: "User", slug: "user_id", id: "4", type: "id" };
 
-const parameters = [idFilter, nameFilter, sourceFilter, userFilter];
+const notUsedFilter = {
```
WiNloSt, line 74: How about this?
```suggestion
const unusedFilter = {
```

e2e/test/scenarios/sharing/public-dashboard.cy.spec.js
```diff
@@ -181,4 +204,22 @@ describe("scenarios > public > dashboard", () => {
...
   });
+
+  it("should only display filters mapped to cards on the selected tab", () => {
+    cy.get("@dashboardId").then(id => {
+      visitPublicDashboard(id);
+    });
+
+    dashboardParametersContainer().within(() => {
+      cy.findByText(textFilter.name).should("be.visible");
+      cy.findByText(numberFilter.name).should("not.exist");
```
WiNloSt, line 215: Should we name something like `unusedFilter` to emphasize it's not used? Like how we're asserting the name "Not Used Filter" (or "Unused Filter") in `e2e/test/scenarios/embedding/embedding-dashboard.cy.spec.js`

WiNloSt (commented):
Looks and works great, but could you check if we don't need to store the `selectedTabId` for public dashboard then that'd be great.

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
```diff
@@ -94,6 +95,7 @@ function EmbedFrame({
   parameters,
   parameterValues,
   draftParameterValues,
+  hiddenParameterSlugs = [],
```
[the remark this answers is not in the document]
kulyk: Good point 👍 Made `EmbedFrame's` prop a string instead

kulyk:
@WiNloSt man I miss your reviews 😍 Special thanks on the callout re `selectTabId`, totally forgot we still have access to Redux in public env 👍

cdeweyx (approved):
LGTM 👍

frontend/src/metabase/public/containers/PublicDashboard.jsx
```diff
@@ -127,6 +126,30 @@ class PublicDashboard extends Component {
...
+  getCurrentTabDashcards = () => {
+    const { dashboard, selectedTabId } = this.props;
+    if (!Array.isArray(dashboard?.dashcards)) {
+      return [];
+    }
+    if (!selectedTabId) {
+      return dashboard.dashcards;
+    }
+    return dashboard.dashcards.filter(
+      dc => dc.dashboard_tab_id === selectedTabId,
```
WiNloSt, line 138: nit: I like explicit names more than implicit names. Even though they would definitely be longer, I don't have to think twice what they actually represent.

I see there are a few of abbreviated names here
- dc
- pm
- p

WiNloSt (approved):
Other than the small nitpick names. Everything looks good! I haven't retested it again though, but the tests should cover that.

frontend/src/metabase/public/containers/PublicDashboard.jsx
kulyk: Fair, fixed 👍

PR #38375, opened 2024-02-01
[MLv2] Cache JS -> CLJS metadata conversion

bshepherdson:
This previous got rebuilt from scratch over and over, since different
queries being handed the same JS `Metadata` instance were returning
completely separate `MetadataProvider`s.

The FE replaces the `Metadata` instance as new metadata arrives from the
BE, so the `metabase.lib.cache/side-channel-cache` which is attached to
the input object is implicitly invalidated when the old `Metadata` is
thrown away.

### DIFF
--- a/src/metabase/lib/cache.cljc
+++ b/src/metabase/lib/cache.cljc
@@ -10,19 +10,20 @@
 
   If there is not already a key `subkey` in the map, calls `(f x)` and caches the value at `subkey`.
   If there is a value at `subkey`, it is returned directly."
-  [subkey x f]
-  (comment subkey) ; Avoids lint warning for half-unused `subkey`.
-  #?(:clj  (f x)
-     :cljs (if (or (object? x) (map? x))
-             (do
-               (when-not (.-__mbcache ^js x)
-                 (set! (.-__mbcache ^js x) (atom {})))
-               (if-let [cache (.-__mbcache ^js x)]
-                 (if-let [cached (get @cache subkey)]
-                   cached
-                   ;; Cache miss - generate the value and cache it.
-                   (let [value (f x)]
-                     (swap! cache assoc subkey value)
-                     value))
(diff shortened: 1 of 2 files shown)


bshepherdson:
* `master`
  * **PR #38375** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38375?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38375?utm_source=stack-comment).

src/metabase/lib/cache.cljc
```diff
@@ -10,19 +10,29 @@
...
+                (when-not (.-__mbcache ^js x)
+                  (set! (.-__mbcache ^js x) (atom {})))
+                (when-not js/window.__mbcache_stats
+                  (set! js/window.__mbcache_stats (js-obj)))
+                (if-let [cache (.-__mbcache ^js x)]
+                  (if-let [cached (get @cache subkey)]
+                    (do
+                      (if-let [^js stats (aget js/window.__mbcache_stats subkey)]
+                        (set! (.-hits stats) (inc (.-hits stats)))
+                        (aset js/window.__mbcache_stats subkey (js-obj "hits" 1 "misses" 0)))
```
metamben, line 28: Just curious, how is the following worse?

```suggestion
                        (aset js/window.__mbcache_stats subkey #js {:hits 1 :misses 0}))
```

metamben (approved):

src/metabase/lib/cache.cljc
bshepherdson: I agree it's better, actually. But this stats-tracking code got removed in any case; I don't really want to check that into master.

PR #38377, opened 2024-02-01
Milestone 1: Default Dashboard Body to a Fixed Width

JesseSDevaney:
### Problem

Responsively sized dashboards lead to consumers seeing a dashboard in a way the creator did not intend for it to be consumed. The creator may have built their dashboard for consumption on a large screen only to have to their consumers view it on a small screen, potentially losing a lot of the useful value the creator intended the dashboard to have. 

### Value
This PR introduces a new `width` setting for dashboards. It has two values:
 1. `"full" = responsive`: the current behavior
2. `"fixed" = 1048px`: the default behavior for new dashboards

Creators will now be able to toggle this setting, build their dashboard, and have their consumers see it exactly the way they intended it to be seen.

---
### Desired Behavior/Acceptance Criteria
**Solution Behaviors**
- (in edit mode) Expose new ellipsis button that opens a popover filled with new settings options (at this time, only 1)
  image
  image
  - [ ] Expose width setting toggle in popover for changing dashboard width between "full" and "fixed".
    - [ ] Toggled on = "full"
    - [ ] Toggled off = "fixed"
- Dashboard body:
  - [ ] When width = "full", dashboard grid should auto-scale to the full-width available
    image
  - [ ] When width = "fixed", dashboard grid should remain at the fixed width (1048px) until it is the size of the viewport and then should responsively downscale until reaching the mobile view
    image
    image
    image

  - [ ] Both options should condense to one column when the normal mobile breakpoint is reached
- [ ] Newly created dashboards should default to `width = "fixed"` 
- [ ] Previously existing dashboards should retain `width = "full"`

**Validation Tests**
- [ ] (`dashboard.cy.spec.js`) Validate new width setting works as intended
- [ ] (`public-dashboard.cy.spec.js`) Validate public dashboards respect the new "width" setting
- [ ] (`x-rays.cy.spec.js`) Validate X-ray dashboards are created with and respect `width = "fixed"`

---
### How to Manually Verify the Solution
**Full-App**
1. Checkout branch and start development server
2. Open an existing dashboard and see that it still has `width = "full"`
3. Create a new dashboard and see that it defaults to `width = "fixed"`
4. Toggle the `width` setting, save it, and refresh. See that the `width` setting persists after saving and refreshing.

**X-Rays**
1. In the collection sidebar, go to Browse Data > Databases > Sample Database > X-Ray on Table
  image
2. See that X-Ray dashboard body has fixed-width.
  image

**Embedding**
1. Open up a dashboard in the full-app and enable embedding
  image
  image
4. Follow this guide to setup the rest of the static embedding app
   - https://github.com/metabase/metabase-nodejs-static-embedding-sample
5. Change the width toggle in the full-app, refresh the static embedding app, and see that the dashboard width is `"fixed"` when set to `"fixed"` and `"full"` when set to `"full"`

---
**Context**
- Completes Milestone 1: https://github.com/metabase/metabase/issues/36358


frontend/src/metabase-types/api/dashboard.ts
```diff
@@ -38,6 +38,7 @@ export interface Dashboard {
   archived: boolean;
   public_uuid: string | null;
   embedding_params?: Record<string, string> | null;
+  width: "full" | "fixed";
```
JesseSDevaney, line 41: Add new dashboard attribute to dashboard type

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.styled.tsx
```diff
@@ -138,6 +138,18 @@ export const ParametersWidgetContainer = styled(FullWidthContainer)<{
...
+  isFixedWidth: boolean;
+}>`
+  margin: 0 auto;
+  ${({ isFixedWidth }) =>
+    isFixedWidth &&
+    css`
+      max-width: ${FIXED_WIDTH};
+    `}
+`;
+
```
JesseSDevaney, line 152: Create the fixed width container and use `isFixedWidth` variable to dynamically set the width

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
```diff
@@ -523,10 +524,14 @@ function DashboardInner(props: DashboardProps) {
...
-                {renderContent()}
-              </CardsContainer>
+              <FixedWidthContainer
+                data-testid="fixed-width-container"
+                isFixedWidth={dashboard?.width === "fixed"}
+              >
+                <CardsContainer id="Dashboard-Cards-Container">
+                  {renderContent()}
+                </CardsContainer>
+              </FixedWidthContainer>
```
JesseSDevaney, line 534: Wrap `FixedWidthContainer` around the dashboard cards body. Use `dashboard.width` attribute to dynamically set a fixed/responsive width.

frontend/src/metabase/dashboard/components/DashboardButtonTooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+          {label}
+        </Text>
+      }
+      withArrow
+      arrowSize={10}
+    >
+      {children}
+    </Tooltip>
+  );
+}
```
JesseSDevaney, line 30: Update dashboard header button tooltips to use the new Mantine `Tooltip` component

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -425,18 +432,30 @@ class DashboardHeaderContainer extends Component<DashboardHeaderProps> {
...
 
+      // Extra Buttons Menu
+      buttons.push(
+        <ExtraEditButtonsMenu
+          key="extra-options-button"
+          setDashboardAttribute={setDashboardAttribute}
+          dashboard={dashboard}
+        />,
+      );
+
```
JesseSDevaney, line 458: Add the new edit mode button

frontend/src/metabase/public/containers/PublicDashboard.jsx
```diff
@@ -166,14 +167,19 @@ class PublicDashboard extends Component {
...
+                >
+                  <DashboardGridConnected
+                    {...this.props}
+                    isPublic
+                    className="spread"
+                    mode={PublicMode}
+                    metadata={this.props.metadata}
+                    navigateToNewCardFromDashboard={() => {}}
+                  />
+                </FixedWidthContainer>
```
JesseSDevaney, line 182: Setup `FixedWidthContainer` on public dashboards

frontend/src/metabase/ui/theme.ts
```diff
@@ -38,7 +38,7 @@ export const getThemeOverrides = (): MantineThemeOverride => ({
   primaryColor: "brand",
   primaryShade: 0,
   shadows: {
-    sm: "0px 4px 4px 0px rgba(0, 0, 0, 0.08)",
+    sm: "0px 1px 4px 2px rgba(0, 0, 0, 0.08)",
```
JesseSDevaney, line 41: Adjust shadow styling to match the designs styling. The other shadow styling did not create a shadow on the top of the popover making it blend in with the dashboard header

src/metabase/api/public.clj
```diff
@@ -231,7 +231,7 @@
   [& conditions]
   {:pre [(even? (count conditions))]}
   (binding [params/*ignore-current-user-perms-and-return-all-field-values* true]
-    (-> (api/check-404 (apply t2/select-one [Dashboard :name :description :id :parameters :auto_apply_filters], :archived false, conditions))
+    (-> (api/check-404 (apply t2/select-one [Dashboard :name :description :id :parameters :auto_apply_filters :width], :archived false, conditions))
```
JesseSDevaney, line 234: Add new dashboard `width` attribute to public dashboard API response

test/metabase/api/embed_test.clj
```diff
@@ -155,7 +155,7 @@
 
 (def successful-dashboard-info
   {:auto_apply_filters true, :description nil, :parameters [], :dashcards [], :tabs [],
-   :param_values {}, :param_fields nil})
+   :param_values {}, :param_fields nil :width "fixed"})
```
adam-james-v, line 158: Adding the :width "fixed" default here should allow the tests to pass. Since we now send that key via the api, it's fair to add it as part of the expected successful dashboard info.

frontend/src/metabase/dashboard/components/DashboardButtonTooltip.tsx
kulyk: Nice! 😍 QQ: did you have to apply all of these styles to make the tooltips look exactly as before?

e2e/test/scenarios/sharing/public-dashboard.cy.spec.js
```diff
@@ -181,4 +181,32 @@ describe("scenarios > public > dashboard", () => {
...
+    );
+
+    // toggle full-width
+    cy.get("@dashboardId").then(id => {
+      cy.signInAsAdmin();
+      cy.request("PUT", `/api/dashboard/${id}`, {
+        width: "full",
+      });
+      visitPublicDashboard(id);
+    });
```
kulyk, line 204: Do we actully need `cy.signInAsAdmin()` here?

frontend/src/metabase/public/containers/PublicDashboard.jsx
```diff
@@ -166,14 +167,19 @@ class PublicDashboard extends Component {
...
+                  isFixedWidth={dashboard?.width === "fixed"}
+                >
+                  <DashboardGridConnected
+                    {...this.props}
+                    isPublic
+                    className="spread"
+                    mode={PublicMode}
+                    metadata={this.props.metadata}
+                    navigateToNewCardFromDashboard={() => {}}
+                  />
```
kulyk, line 181: @JesseSDevaney WDYT about making `FixedWidthContainer` a part of `DashboardGridConnected`?

We have three top-level dashboard components (`DashboardApp,` `PublicDashboard`, and `AutomaticDashboardApp`), and they all use `DashboardGridConnected`. I think it'd make sense to move the width logic there, so all three dashboard components can get this featyre for free

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.styled.tsx
```diff
@@ -138,6 +138,18 @@ export const ParametersWidgetContainer = styled(FullWidthContainer)<{
     `}
 `;
 
+export const FIXED_WIDTH = "1048px";
```
kulyk, line 141: Not sure if it's expected or not, but even though we're setting 1048px, the actual content is a bit narrower because of some horizontal margin. I think we'd want to remove it @kdoh @cdeweyx ? I want these 28px back! 😂 

CleanShot 2024-02-02 at 18 09 50@2x

kulyk (commented):
Love this 😍 Left couple of comments

frontend/src/metabase/dashboard/components/DashboardButtonTooltip.tsx
JesseSDevaney: I actually got this from somewhere else where someone converted the old tool-tip to use the new Mantine tooltip (https://github.com/metabase/metabase/blob/master/frontend/src/metabase/dashboard/components/DashboardEmbedHeaderButton/DashboardEmbedHeaderButton.tsx#L39-L51)

e2e/test/scenarios/sharing/public-dashboard.cy.spec.js
JesseSDevaney: It was failing the request to change the width, so I had to add that in there

frontend/src/metabase/public/containers/PublicDashboard.jsx
JesseSDevaney: Great call-out. I moved the `FixedWidthContainer` into `DashboardGridConnected`

frontend/src/metabase/dashboard/components/Dashboard/Dashboard.styled.tsx
JesseSDevaney: After I moved `FixedWidthContainer` into `DashboardGridConnected`, this is no longer an issue. It still has padding when it reaches the edges of the viewport, but that is where we want it.

src/metabase/automagic_dashboards/core.clj
```diff
@@ -702,7 +702,8 @@
                   (format "%s#show=all" url))
           :transient_filters query-filter
           :param_fields (filter-referenced-fields root query-filter)
-          :auto_apply_filters true))))
+          :auto_apply_filters true
+          :width "fixed"))))
```
adam-james-v, line 706: Looked into it, and that’s as good a place as any to add the key.
One difference here, just to keep a good mental model, is that automagic dashboards are ephemeral. We are not pulling anything from the AppDB, and (as far as I understand it), once you hit ‘save’, it’s just a normal dashboard. So, the key you added is part of generating this ‘ephemeral’ dashboard, not requesting an additional column from the db.

The embed change we made pulls already existing data from the AppDB.

This doesn’t affect your change in any way, I just wanted to explain the difference between the embedded dashboard API and the automagic Dashboards.
I’ll add this as a review on the PR too. :slightly_smiling_face:

kulyk (approved):
Code looks good! 🚢 🇮🇹

cdeweyx (approved):

alxnddr (approved):

PR #38378, opened 2024-02-01
Return card__<id> for non-model cards and FK targets for tables

### LINKED ISSUE #38312: [MLv2] [BE] `Lib.dependentMetadata` does not return source cards
#### Unit test

```tsx
import { createSampleDatabase } from 'metabase-types/api/mocks/presets';
import * as Lib from 'metabase-lib';
import { createMockMetadata } from '__support__/metadata';

const SAMPLE_DATABASE = createSampleDatabase();

const SAMPLE_METADATA = createMockMetadata({
  databases: [SAMPLE_DATABASE],
});

const query = Lib.fromLegacyQuery(1, SAMPLE_METADATA, {
  type: 'query',
  query: {
    'source-table': 'card__5',
  },
  database: 1,
});

expect(Lib.dependentMetadata(query)).toEqual([
  { type: 'database', id: 1 },
  { type: 'schema', id: 1 },
  { type: 'table', id: 'card__5' },
]);
```

#### e2e test

1. `git checkout 37390-dependent-metadata`
2.  Run e2e test `title-drill.cy.spec.js` -> `should lead you to a table question with filtered ID (metabase#17213)`

The dataset query is: 
```js
{
    "type": "query",
    "query": {
        "source-table": "card__5"
    },
    "database": 1
}
```

### Outcome

`Lib.dependentMetadata` will return:
```js
[
  { type: 'database', id: 1 },
  { type: 'schema', id: 1 },
]
```

While it should return:
```js
[
  { type: 'database', id: 1 },
  { type: 'schema', id: 1 },
  { type: 'table', id: 'card__5' },
]
```

metamben:
Fixes #38312.

Returning `card__<id>` table items was restricted to models and foreign key targets were restricted to cards. Both of these restrictions are lifted in this PR.

### DIFF
--- a/src/metabase/lib/fe_util.cljc
+++ b/src/metabase/lib/fe_util.cljc
@@ -112,6 +112,14 @@
       _
       (lib.metadata.calculation/display-name query stage-number filter-clause))))
 
+(defn- foreign-key-items
+  [columns]
+  (for [column columns
+        :let [fk-target (:fk-target-field-id column)]
+        :when (and (integer? fk-target)
+                   (lib.types.isa/foreign-key? column))]
+    {:type :field, :id fk-target}))
+
 (defn- query-dependents
   [metadata-providerable query-or-join]
   (let [base-stage (first (:stages query-or-join))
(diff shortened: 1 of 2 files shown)


kamilmielnik (commented):
I confirm it fixes #38312

bshepherdson (approved):

PR #38380, opened 2024-02-02
Sort drill should remove ALL existing order bys

### LINKED ISSUE #37633: [MLv2] [BE] [Bug] `sort` drill does not remove existing order by clauses
Context https://metaboat.slack.com/archives/C01LQQ2UW03/p1705010842530169

camsaul:
We had an error in the old logic, the old logic only removed existing order bys on the same column. According to #37633, we actually want to remove all existing order bys.

Fixes #37633


metamben (approved):
LGTM

PR #38386, opened 2024-02-02
Fix - Not able to switch between databases after a question is created

kamilmielnik:
Closes #38354

### Description

Issue has been fixed in `master` by #37302.
Added a repro for `master` branch in a separate PR: #38387

Issue #38354, opened 2024-02-01
Not able to switch between databases after a question is created


Tony-metabase:
### Describe the bug

After you create a question you are not able to switch between databases

### To Reproduce

Make sure you have 2 Databases Connected. In my case I have the App DB and Sample Database
1. Go to New Question -> Sample DB -> Orders -> Save
2. Go back to this question -> App DB (or any other DB apart form Sample DB) -> Core Users

<img width="1457" alt="image" src="https://github.com/metabase/metabase/assets/110378427/1c3cac4c-b3e8-4215-9739-8b385e6aeb39">

3. The question will break and if you do save this state and refresh you end up with the following, a data source pointing to a table in Sample Database which doesn't even exist there

<img width="1492" alt="image" src="https://github.com/metabase/metabase/assets/110378427/30052849-b84d-499d-8345-2f23528f4817">

### Expected behavior

The GUI is able to identify changes in the Datasource

### Logs
```
[57e0224d-423e-4700-bdb3-ec659484817e] 2024-02-01T10:57:27+01:00 ERROR metabase.query-processor.middleware.catch-exceptions Error processing query: Failed to fetch :metadata/table 65
{:database_id 1,
 :started_at #t "2024-02-01T09:57:27.002306Z[GMT]",
 :action_id nil,
 :error_type :invalid-query,
 :json_query
 {:database 1,
  :type "query",
  :query {:source-table 65, :limit 10},
  :parameters [],
  :middleware {:js-int-to-string? true, :add-default-userland-constraints? true}},
 :native nil,
 :status :failed,
 :class clojure.lang.ExceptionInfo,
 :stacktrace
 ["--> query_processor.store$missing_bulk_metadata_error.invokeStatic(store.clj:182)"
  "query_processor.store$missing_bulk_metadata_error.invoke(store.clj:181)"
  "query_processor.store$fn__49942$_AMPERSAND_f__49943$fn__49958.invoke(store.clj:221)"
  "query_processor.store$fn__49942$_AMPERSAND_f__49943.invoke(store.clj:219)"
  "query_processor.store$fn__49942$fn__49964.invoke(store.clj:189)"
  "query_processor.middleware.resolve_source_table$resolve_source_tables.invokeStatic(resolve_source_table.clj:39)"
  "query_processor.middleware.resolve_source_table$resolve_source_tables.invoke(resolve_source_table.clj:34)"
[...]

PR #37302, opened 2024-01-05
Migrate data notebook step to MLv2

ranquild:
Epic https://github.com/metabase/metabase/issues/36983
Closes https://github.com/metabase/metabase/issues/36738
Blocked by https://github.com/metabase/metabase/issues/37303

Also completely removes legacy query from Notebook.

How to verify:
- Creating new queries based on tables and saved questions should work

Please note this is completely broken until the blocker above is fixed.

PR #38387, opened 2024-02-02
Repro #38354 - Not able to switch between databases after a question is created

kamilmielnik:
Repro for #38354

### Description

Fixed in #38386 which targets `release-x.48.x` branch.
The issue has been fixed in `master` by #37302, that's why this PR only has the e2e test.


e2e/test/scenarios/question/reproductions/38354-changing-source-database.cy.spec.js
```diff
@@ -0,0 +1,44 @@
...
+const QUESTION_DETAILS = {
+  query: {
+    "source-table": ORDERS_ID,
+    limit: 5,
+  },
+};
+
+describe("issue 38354", () => {
+  beforeEach(() => {
+    restore("postgres-12");
```
uladzimirdev, line 19: curious, why postgres-12?

uladzimirdev (approved):

e2e/test/scenarios/question/reproductions/38354-changing-source-database.cy.spec.js
nemanjaglumac: Because we need another database, and the snapshot already exists.
Did you mean - why not another H2?

e2e/test/scenarios/question/reproductions/38354-changing-source-database.cy.spec.js
```diff
@@ -0,0 +1,44 @@
...
+import { ORDERS_ID } from "metabase-types/api/mocks/presets";
+
+const QUESTION_DETAILS = {
+  query: {
+    "source-table": ORDERS_ID,
+    limit: 5,
+  },
+};
+
+describe("issue 38354", () => {
```
nemanjaglumac, line 17:
```suggestion
describe("issue 38354", { tags: "@external" }, () => {
```

nemanjaglumac (changes requested):
We need to tag tests that use QA DBs as external.

Either do that, or add another H2.

e2e/test/scenarios/question/reproductions/38354-changing-source-database.cy.spec.js
kamilmielnik: b6ac02ce2b4d5bdce6312fa9f7378257a0ceb58f

nemanjaglumac (approved):

PR #38390, opened 2024-02-02
Hide "Add your data" step for people interested in embedding

npretto:
[[Epic] Meet embedders at the door - minimal edition](https://github.com/metabase/metabase/issues/38233)

### Description

This only adds the basic logic, I want to have something to show to product and design.
In a later PR I'll refactor the logic and create a selector that will return the enabled steps, so that the logic to show them and to decide which is "next" step will have a single source of truth.
Also more tests will be added later when the logic of this question will be connected to the new custom homepage

### How to verify

If during the setup you select you're interested in embedding, the DB connection step won't be shown.

### Demo

https://github.com/metabase/metabase/assets/1914270/5756ca94-ca84-4da0-b98c-40feaf6b19d4

### Checklist

### DIFF
--- a/frontend/src/metabase/setup/components/SettingsPage/SettingsPage.tsx
+++ b/frontend/src/metabase/setup/components/SettingsPage/SettingsPage.tsx
@@ -1,5 +1,6 @@
 import LogoIcon from "metabase/components/LogoIcon";
 import { isNotFalsy } from "metabase/lib/types";
+import { useSelector } from "metabase/lib/redux";
 import { CloudMigrationHelp } from "../CloudMigrationHelp";
 import { DatabaseStep } from "../DatabaseStep";
 import { CompletedStep } from "../CompletedStep";
@@ -12,11 +13,13 @@ import { UsageQuestionStep } from "../UsageQuestionStep";
 import { PageBody, PageHeader } from "./SettingsPage.styled";
 
 export const SettingsPage = (): JSX.Element => {
+  const usageReason = useSelector(s => s.setup.usageReason);
+
   const numberedSteps = [
     { component: LanguageStep, key: "language-step" },
     { component: UserStep, key: "user-step" },
     { component: UsageQuestionStep, key: "usage-question-step" },
-    {
+    usageReason !== "embedding" && {
       component: DatabaseStep,
       key: "database-step",
     },
--- a/frontend/src/metabase/setup/components/SettingsPage/SettingsPage.tsx
+++ b/frontend/src/metabase/setup/components/SettingsPage/SettingsPage.tsx
@@ -1,6 +1,7 @@
 import LogoIcon from "metabase/components/LogoIcon";
 import { isNotFalsy } from "metabase/lib/types";
 import { useSelector } from "metabase/lib/redux";
+import { getUsageReason } from "metabase/setup/selectors";
 import { CloudMigrationHelp } from "../CloudMigrationHelp";
 import { DatabaseStep } from "../DatabaseStep";
 import { CompletedStep } from "../CompletedStep";
@@ -13,7 +14,7 @@ import { UsageQuestionStep } from "../UsageQuestionStep";
 import { PageBody, PageHeader } from "./SettingsPage.styled";
 
 export const SettingsPage = (): JSX.Element => {
-  const usageReason = useSelector(s => s.setup.usageReason);
+  const usageReason = useSelector(getUsageReason);
 
   const numberedSteps = [
     { component: LanguageStep, key: "language-step" },
--- a/frontend/src/metabase/setup/reducers.ts
+++ b/frontend/src/metabase/setup/reducers.ts
@@ -54,8 +54,9 @@ export const reducer = createReducer(initialState, builder => {
     state.step = USAGE_STEP;
   });
   builder.addCase(submitUsageReason.pending, (state, { meta }) => {
-    state.usageReason = meta.arg;
-    state.step = DATABASE_STEP;
+    const usageReason = meta.arg;
+    state.usageReason = usageReason;
+    state.step = usageReason === "embedding" ? PREFERENCES_STEP : DATABASE_STEP;
   });
   builder.addCase(updateDatabaseEngine.pending, (state, { meta }) => {
     state.databaseEngine = meta.arg;
--- a/frontend/src/metabase/setup/reducers.ts
+++ b/frontend/src/metabase/setup/reducers.ts
@@ -56,6 +56,7 @@ export const reducer = createReducer(initialState, builder => {
   builder.addCase(submitUsageReason.pending, (state, { meta }) => {
     const usageReason = meta.arg;
     state.usageReason = usageReason;
+    // this logic will be refactored before we introduce more steps, to be less fragile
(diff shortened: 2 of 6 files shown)


frontend/src/metabase/setup/setup.unit.spec.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  // there is some async stuff going on with the locale loading
+  await screen.findByText("Let's get started");
+
+  return;
+};
+
+describe("setup", () => {
+  describe("default step order", () => {
+    // eslint-disable-next-line jest/expect-expect
```
npretto, line 35: it didn't find the expect, because it is "hidden" in the util `expectSectionToHaveLabel`

frontend/src/metabase/setup/setup.unit.spec.tsx
```diff
@@ -0,0 +1,156 @@
...
+    it.each([
+      ["What's your preferred language?", "1"],
```
npretto, line 37: this test is just a sanity check put in place to check that I won't break random things when I'll do the big refactor

frontend/src/metabase/setup/setup.unit.spec.tsx
```diff
@@ -0,0 +1,156 @@
...
+        skipWelcomeScreen();
+
+        expectSectionToHaveLabel(step, label);
+      },
+    );
+  });
+
+  describe("Usage question", () => {
+    const setupForUsageQuestion = async () => {
+      await setup();
```
npretto, line 55: The setup manually fills the previous steps, I know that mocking the redux state to directly load the correct step would be faster, but it would also give us less confidence and depend on some implementation details that will change with the refactor of the step navigation

frontend/src/metabase/setup/components/SettingsPage/SettingsPage.tsx
```diff
@@ -12,11 +13,13 @@ import { UsageQuestionStep } from "../UsageQuestionStep";
 import { PageBody, PageHeader } from "./SettingsPage.styled";
 
 export const SettingsPage = (): JSX.Element => {
+  const usageReason = useSelector(s => s.setup.usageReason);
+
   const numberedSteps = [
```
npretto, line 18: Those steps will be moved to a custom selector, that will be used also the the thunks or reducers to check which step is the "next" to go after a submit

deniskaber (approved):

frontend/src/metabase/setup/reducers.ts
```diff
@@ -54,8 +54,9 @@ export const reducer = createReducer(initialState, builder => {
     state.step = USAGE_STEP;
   });
   builder.addCase(submitUsageReason.pending, (state, { meta }) => {
-    state.usageReason = meta.arg;
-    state.step = DATABASE_STEP;
+    const usageReason = meta.arg;
+    state.usageReason = usageReason;
+    state.step = usageReason === "embedding" ? PREFERENCES_STEP : DATABASE_STEP;
```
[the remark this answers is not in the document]
WiNloSt: should you add some comments that this logic is expected to be removed later, so we don't forget to handle this afterward?

frontend/src/metabase/setup/selectors.ts
```diff
@@ -21,6 +21,10 @@ export const getUserEmail = (state: State): string | undefined => {
   return getUser(state)?.email;
 };
 
+export const getUsageReason = (state: State) => {
```
WiNloSt, line 24: I assume the usage reason will land somewhere else once we decide we're not going to store it in Redux? And by then, this selector along with its counterparts will be removed as well, is that right?

frontend/src/metabase/setup/setup.unit.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+import type { UsageReason } from "metabase-types/api";
+import {
+  createMockSettingsState,
+  createMockSetupState,
+  createMockState,
+} from "metabase-types/store/mocks";
+import { renderWithProviders, screen } from "__support__/ui";
+import { Setup } from "./components/Setup";
+
+const setup = async ({ step = 0 } = {}) => {
```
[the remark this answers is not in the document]
WiNloSt: I've sampled a bunch of unit test files, and we mostly declare this:
- as `function` as you mentioned
- but declared as the first thing in a test file.

I agree with you I like reading things that are more important first, and leaving the implementation details later, but that's not what we seem to be doing at this point. I'd err on the side of consistency in this case that is leaving it here.

frontend/src/metabase/setup/setup.unit.spec.tsx
WiNloSt: Could you define the custom assertion function, so it's more explicit? 
e.g. `/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectSectionToHaveLabel"] }] */`


Ref: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/expect-expect.md

frontend/src/metabase/setup/setup.unit.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+const skipLanguageStep = () => clickNextStep();
+
+const submitUserInfoStep = async ({
+  firstName = "John",
+  lastName = "Smith",
+  email = "john@example.org",
+  companyName = "Acme",
+  password = "Monkeyabc123",
+} = {}) => {
+  fireEvent.change(screen.getByLabelText("First name"), {
```
WiNloSt, line 153: We use `userEvent` for testing because it simulates the user behaviors while `fireEvent` just dispatch a DOM event.

frontend/src/metabase/setup/setup.unit.spec.tsx
```diff
@@ -0,0 +1,192 @@
...
+  fireEvent.change(screen.getByLabelText("Confirm your password"), {
+    target: { value: password },
+  });
+  await waitFor(() =>
+    expect(screen.getByRole("button", { name: "Next" })).toBeEnabled(),
+  );
+  clickNextStep();
+};
+
+const fillUsageReason = (usageReason: UsageReason) => {
```
WiNloSt, line 177: Technically, it's not a fill, but selecting a usage reason.

```suggestion
const selectUsageReason = (usageReason: UsageReason) => {
```

fill feels more like an input or a textbox where you need to type texts.

WiNloSt (commented):
This mostly looks good, with some stuff that could be polished a bit more before merging.

frontend/src/metabase/setup/selectors.ts
WiNloSt: Ah I was confused. Ignore it then.

frontend/src/metabase/setup/setup.unit.spec.tsx
npretto: Done, thanks c0b86baa7dcaa595d91aca09a423a86050d69cb6

frontend/src/metabase/setup/reducers.ts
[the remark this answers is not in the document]
npretto: Good point, I added a comment

WiNloSt (approved):
The latest change looks good to me.

PR #38395, opened 2024-02-02
Embedding + required params

oleggromov:

## Description
Part of the feature [Let creators make parameters required](https://github.com/metabase/metabase/issues/36524), I add required parameters to the 1) question and 2) dashboard static embeddings.

### How it works
1. When a parameter is required, it _must have_ a default value
2. Embedding modal (both questions and dashboards):
    - a required parameter cannot be disabled anymore ("disabled" option is disabled)
    - it's value in the embedding code and JWT is no longer `null` by default — instead, it's set to the default value
    - when a parameter is locked, the filter widget works in the special required mode, which won't allow removing the value
    - when a parameter is editable, nothing changes in the modal
    - in preview, filter widgets work in the special required mode too
3. Static embedding
    - when a required parameter is locked, takes its value from the JWT
    - when a required parameter is editable, renders the filter widget in the required mode
    - non-required parameters should work as before
4. Setting up dashboard / parameter values
    - when a dashboard is embedded and the parameter you're changing is disabled, we prevent it from being made required with a warning

## How to check
There are a few flows we want to check. Since the logic is similar for questions and dashboards, I'll describe them together.

### 1. Not embedded yet question & dashboard
1. Create a question or a dashboard
2. Add required parameter(s) with default values
3. Observe that they can be made required as previously
4. Now go to the embedding settings and see that you cannot set this parameter to "disabled"
5. Check that when you choose "locked", you see the default values in the embedding code (JWT source)
6. Publish embedding and check that "editable" filters use widget in the default mode

### 2. Already embedded question or dashboard
1. Create a question or dashboard
2. Add parameters without making them required
3. In the embedding settings, make one of them "disabled" (this is the default anyway)
4. Edit the dashboard and observe that you cannot make this parameter required anymore

## TODO

- [x] disable "Disabled" option for required parameters + indicate a parameter is required
- [x] update texts if needed
- [x] locked preview widget + locked value in the embedding code
- [x] update parameter widget in the preview UI
- [x] update parameter widget in the actual embed
- [ ] ~~validation and errors: disallow publishing an embedding with required parameters set to disabled, with null locked value (BE) + check for errors (FE) + optionally refactor 2 API calls (FE, to be confirmed)~~
- [x] validation: generated embedding code for locker required parameters should include the default value
- [ ] ~~?when we pass empty value for a required parameter through JWT, use default~~
- [x] embedded questions, including disabling "required" toggle when parameter is "disabled" in an embedded question
- [x] update existing tests
- [x] add new e2e tests

### DIFF
--- a/e2e/support/helpers/e2e-embedding-helpers.js
+++ b/e2e/support/helpers/e2e-embedding-helpers.js
@@ -1,5 +1,5 @@
 import { METABASE_SECRET_KEY } from "e2e/support/cypress_data";
-import { modal } from "e2e/support/helpers/e2e-ui-elements-helpers";
+import { modal, popover } from "e2e/support/helpers/e2e-ui-elements-helpers";
 
 /**
  * @typedef {object} QuestionResource
@@ -195,6 +195,47 @@ export function openStaticEmbeddingModal({
       cy.findByText(previewModeToKeyMap[previewMode]).click();
     }
   });
+}
+
+export function closeStaticEmbeddingModal() {
+  modal().icon("close").click();
+}
+
+/**
+ * Open Static Embedding setup modal
+ * @param apiPath - "card" or "dashboard"
+ * @param callback
+ */
+export function publishChanges(apiPath, callback) {
+  cy.intercept("PUT", `/api/${apiPath}/*`).as("publishChanges");
+
+  cy.button(/^(Publish|Publish changes)$/).click();
+
+  // TODO this could be simplified when we send one publish request instead of two
+  cy.wait(["@publishChanges", "@publishChanges"]).then(xhrs => {
+    // Unfortunately, the order of requests is not always the same.
+    // Therefore, we must first get the one that has the `embedding_params` and then assert on it.
+    const targetXhr = xhrs.find(({ request }) =>
+      Object.keys(request.body).includes("embedding_params"),
+    );
+    callback?.(targetXhr);
+  });
+}
+
+export function getParametersContainer() {
+  return cy.findByLabelText("Configuring parameters");
+}
+
+export function setEmbeddingParameter(name, value) {
+  getParametersContainer().findByText(name).siblings("a").click();
+  popover().contains(value).click();
+}
+
+export function assertEmbeddingParameter(name, value) {
+  getParametersContainer()
+    .findByText(name)
+    .siblings("a")
+    .should("have.text", value);
 }
 
 // @param {("card"|"dashboard")} resourceType - The type of resource we are sharing
--- a/e2e/support/helpers/e2e-embedding-helpers.js
+++ b/e2e/support/helpers/e2e-embedding-helpers.js
@@ -203,7 +203,7 @@ export function closeStaticEmbeddingModal() {
 
 /**
  * Open Static Embedding setup modal
- * @param apiPath - "card" or "dashboard"
+ * @param {"card" | "dashboard"} apiPath
  * @param callback
  */
 export function publishChanges(apiPath, callback) {
@@ -227,15 +227,12 @@ export function getParametersContainer() {
 }
 
 export function setEmbeddingParameter(name, value) {
-  getParametersContainer().findByText(name).siblings("a").click();
+  getParametersContainer().findByLabelText(name).click();
   popover().contains(value).click();
 }
 
 export function assertEmbeddingParameter(name, value) {
-  getParametersContainer()
-    .findByText(name)
-    .siblings("a")
-    .should("have.text", value);
+  getParametersContainer().findByLabelText(name).should("have.text", value);
 }
 
 // @param {("card"|"dashboard")} resourceType - The type of resource we are sharing
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -18,7 +18,7 @@ export function getParameterValue({
   values = {},
   defaultRequired = false,
 }) {
-  const value = values[parameter.id];
+  const value = values?.[parameter.id];
   const useDefault = defaultRequired && parameter.required;
   return value ?? (useDefault ? parameter.default : null);
 }
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -18,7 +18,7 @@ export function getParameterValue({
   values = {},
   defaultRequired = false,
 }) {
-  const value = values?.[parameter.id];
+  const value = values[parameter.id];
   const useDefault = defaultRequired && parameter.required;
   return value ?? (useDefault ? parameter.default : null);
 }
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -9,12 +9,39 @@ import {
 export const PULSE_PARAM_EMPTY = null;
 export const PULSE_PARAM_USE_DEFAULT = undefined;
 
-export function getValuePopulatedParameters(parameters, parameterValues) {
+/**
+ * In some cases, we need to use default parameter value in place of an absent one.
+ * Please use this function when dealing with the required parameters.
+ */
+export function getParameterValue({
+  parameter,
+  values = {},
+  defaultRequired = false,
+}) {
+  const value = values?.[parameter.id];
+  const useDefault = defaultRequired && parameter.required;
+  return value ?? (useDefault ? parameter.default : null);
+}
+
+/**
+ * In some cases, we need to use default parameter value in place of an absent one.
+ * Please use this function when dealing with the required parameters.
+ */
+export function getValuePopulatedParameters({
+  parameters,
+  values = {},
+  defaultRequired = false,
+}) {
   return parameters.map(parameter => ({
     ...parameter,
-    value: parameterValues?.[parameter.id] ?? null,
+    value: getParameterValue({
+      parameter,
+      values,
+      defaultRequired,
+    }),
   }));
 }
+
 export function getDefaultValuePopulatedParameters(
   parameters,
   parameterValues,
--- a/frontend/src/metabase/dashboard/selectors.ts
+++ b/frontend/src/metabase/dashboard/selectors.ts
@@ -26,9 +26,8 @@ import type {
   ClickBehaviorSidebarState,
   EditParameterSidebarState,
   State,
-  StoreDashboard,
 } from "metabase-types/store";
-import type { EmbeddingParameters } from "metabase/public/lib/types";
+import type { EmbeddingParameterVisibility } from "metabase/public/lib/types";
 import Question from "metabase-lib/Question";
 import { isQuestionDashCard } from "./utils";
 
@@ -385,25 +384,21 @@ export const getDashcardParameterMappingOptions = createCachedSelector(
   return props.card.id ?? props.dashcard.id;
 });
 
-export const getDashboardEmbeddingParams = createSelector(
-  [getDashboard],
-  (dashboard: StoreDashboard | undefined): EmbeddingParameters => {
-    if (!dashboard?.enable_embedding) {
-      return {};
-    }
+// Embeddings might be published without passing embedding_params to the server,
+// in which case it's an empty object. We should treat such situations with
+// caution, assuming that an absent parameter is "disabled".
+export function getEmbeddedParameterVisibility(
+  state: State,
+  slug: string,
+): EmbeddingParameterVisibility | null {
+  const dashboard = getDashboard(state);
+  if (!dashboard?.enable_embedding) {
+    return null;
+  }
 
-    return dashboard.embedding_params ?? {};
-  },
-);
-
-export const getEmbeddingDisabledRequiredParameters = createSelector(
-  [getParameters, getDashboardEmbeddingParams],
-  (parameters, embeddingParameters) => {
-    return parameters.filter(
-      param => param.required && embeddingParameters[param.slug] === "disabled",
-    );
-  },
-);
+  const embeddingParams = dashboard.embedding_params ?? {};
+  return embeddingParams[slug] ?? "disabled";
+}
 
 export const getIsHeaderVisible = createSelector(
   [getIsEmbedded, getEmbedOptions],
--- a/frontend/src/metabase/dashboard/selectors.ts
+++ b/frontend/src/metabase/dashboard/selectors.ts
@@ -26,7 +26,9 @@ import type {
   ClickBehaviorSidebarState,
   EditParameterSidebarState,
   State,
+  StoreDashboard,
 } from "metabase-types/store";
+import type { EmbeddingParameters } from "metabase/public/lib/types";
 import Question from "metabase-lib/Question";
 import { isQuestionDashCard } from "./utils";
 
@@ -382,6 +384,26 @@ export const getDashcardParameterMappingOptions = createCachedSelector(
 )((state, props) => {
   return props.card.id ?? props.dashcard.id;
 });
+
+export const getDashboardEmbeddingParams = createSelector(
+  [getDashboard],
+  (dashboard: StoreDashboard | undefined): EmbeddingParameters => {
+    if (!dashboard?.enable_embedding) {
+      return {};
+    }
+
+    return dashboard.embedding_params ?? {};
+  },
+);
+
+export const getEmbeddingDisabledRequiredParameters = createSelector(
+  [getParameters, getDashboardEmbeddingParams],
+  (parameters, embeddingParameters) => {
+    return parameters.filter(
+      param => param.required && embeddingParameters[param.slug] === "disabled",
+    );
+  },
+);
 
 export const getIsHeaderVisible = createSelector(
   [getIsEmbedded, getEmbedOptions],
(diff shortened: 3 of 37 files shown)


frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -258,7 +259,11 @@ function getPreviewParamsBySlug({
   return Object.fromEntries(
     lockedParameters.map(parameter => [
       parameter.slug,
-      parameterValues[parameter.id] ?? null,
+      getParameterValue(
+        parameter,
+        parameterValues,
+        /* defaultRequired */ true,
```
npretto, line 265: can we make this function accept named params?

npretto (dismissed):

WiNloSt:
@oleggromov  I tested the scenario we [discussed on Slack.](https://metaboat.slack.com/archives/C063Q3F1HPF/p1707262527962139?thread_ts=1707218138.865479&cid=C063Q3F1HPF)

From Conor
> - Making a parameter required, when it already exists and is disabled, described [here](https://metaboat.slack.com/archives/C063Q3F1HPF/p1707242598571549?thread_ts=1707218138.865479&cid=C063Q3F1HPF)
    - Agree this needs to be handled to merge. Pulling the default makes sense.


Basically if we published a dashboard without required parameters. And went to edit some parameters to be required, the embedding still isn't aware of the change. Until we republish the embedding that sets the parameter to "Editable".

I kind of expect the embedding to treat `disabled` setting as `editable` as now the parameter is required.

https://www.loom.com/share/44282f926df54b3f902901c6df5d225b

frontend/src/metabase-lib/parameters/utils/parameter-values.js
```diff
@@ -9,12 +9,25 @@ import {
 export const PULSE_PARAM_EMPTY = null;
 export const PULSE_PARAM_USE_DEFAULT = undefined;
 
-export function getValuePopulatedParameters(parameters, parameterValues) {
-  return parameters.map(parameter => ({
-    ...parameter,
-    value: parameterValues?.[parameter.id] ?? null,
+// defaultRequired must be used in features that support required parameters
+export function getParameterValue(parameter, values, defaultRequired = false) {
```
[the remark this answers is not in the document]
WiNloSt: Agree with @npretto, using an object and turning all these parameters into its keys would definitely help since most IDEs will autocomplete it.

frontend/src/metabase-lib/parameters/utils/parameter-values.js
```diff
@@ -9,12 +9,25 @@ import {
...
+  return (
+    values?.[parameter.id] ??
+    (defaultRequired && parameter.required ? parameter.default : null)
```
WiNloSt, line 16: I think having a few if blocks would read better than trying to cramp everything into a single expression. This took me some time to understand it.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
```diff
@@ -62,10 +62,13 @@ export const ParametersSettings = ({
       getValuePopulatedParameters(
         lockedParameters,
         parameterValues,
+        /* defaultRequired */ true,
       ) as EmbedResourceParameterWithValue[],
     [lockedParameters, parameterValues],
   );
 
+  const hasRequiredParameters = resourceParameters.some(p => p.required);
```
WiNloSt, line 70: If possible try to use the unabbreviated names to ensure readers don't have to think twice what each variable would represent. I understand it won't make much difference in this case, but I've seen some folks get used to doing this all the time and we could see a bunch of abbreviated names in the function which could be confusing.

frontend/src/metabase/public/containers/PublicQuestion/PublicQuestion.jsx
```diff
@@ -166,6 +174,15 @@ class PublicQuestionInner extends Component {
     }
   };
 
+  getCardUiParameters() {
```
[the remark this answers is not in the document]
WiNloSt: - I think we could name this `getParameters` to avoid confusion.
- I think this method is needed to encapsulate the need for both `metadata` prop and `card` state.
- Agree with another line expecting this function to return non-array. We could make this function return an empty array instead when there is no `card`

WiNloSt (commented):
Looks good to me apart from the static embed that has `disabled` parameters and later on, has the parameter set to required.

frontend/src/metabase/parameters/components/ParameterSettings/ParameterSettings.tsx
```diff
@@ -143,7 +153,20 @@ const ParameterSettings = ({
               htmlFor={`parameter-setting-required_${parameter.id}`}
             >
               {t`Always require a value`}
+              {isEmbeddedDisabled && (
+                <Tooltip
+                  width={200}
+                  multiline
+                  withArrow
+                  label={
+                    'This filter visiblity is set to "disabled" in the embedded dashboard, so you can\'t make it required'
```
oleggromov, line 162: The actual text and behavior is TBC with the product/designers.

frontend/src/metabase/public/lib/types.ts
```diff
@@ -13,9 +13,13 @@ export type EmbedResourceParameter = {
...
   slug: string;
   type: string;
+  required?: boolean;
+  default?: unknown;
 };
 
-export type EmbeddingParameters = Record<string, string>;
+export type EmbeddingParameterVisibility = "disabled" | "enabled" | "locked";
+
+export type EmbeddingParameters = Record<string, EmbeddingParameterVisibility>;
```
oleggromov, line 22: This is copied from @WiNloSt's PR https://github.com/metabase/metabase/pull/37617/files#diff-e2b89f462d7fdf5e21ca1aff5a1d91ee203cd086e70c5937246a2c19ec8621e0 and discussed with him.

frontend/src/metabase-lib/parameters/utils/parameter-values.js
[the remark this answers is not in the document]
oleggromov: Agreed and done.

frontend/src/metabase/public/containers/PublicQuestion/PublicQuestion.jsx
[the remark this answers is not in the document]
oleggromov: I am keeping the method, which is now renamed, and I have changed how we call it. Thanks!

oleggromov:
> Looks good to me apart from the static embed that has `disabled` parameters and later on, has the parameter set to required.

This has been addressed.

oleggromov:
@npretto please have a look, I have addressed your feedback.

frontend/src/metabase/parameters/components/ParameterSettings/ParameterSettings.tsx
```diff
@@ -132,8 +137,13 @@ const ParameterSettings = ({
           setValue={onChangeDefaultValue}
         />
 
-        <SettingRequiredContainer>
+        <SettingRequiredContainer
+          // This forces re-rendering when switching between parameters,
+          // so that toggles don't slide, which is confusing.
```
npretto, line 142: I see what you mean, but technically this doesn't just make it re-render, it already does that just because the parameter changes. What it does is it makes react throw away the old component a create a new one.
```suggestion
          // This forces the toggle to be a new instance when the parameter changes,
          // so that toggles don't slide, which is confusing.
```

npretto (commented):
Added a small comment.

Unfortunately we found an issue:
on brand new dashboards, if you don't touch the embedding parameters, the object is empty and the logic to disable te "Required" toggle doesn't kick in.

oleggromov:
> on brand new dashboards, if you don't touch the embedding parameters, the object is empty and the logic to disable te "Required" toggle doesn't kick in.

This is fixed by https://github.com/metabase/metabase/pull/38395/commits/b2cc2e89e37d06116b953ccdd818eb811733535e and https://github.com/metabase/metabase/pull/38395/commits/ad12a54e1d5911c3f9e5443f02bd3bcb372bc51a

Thank you!

frontend/src/metabase/parameters/components/RequiredParamToggle/RequiredParamToggle.tsx
```diff
@@ -0,0 +1,43 @@
...
+  disabledTooltip: ReactNode;
+}
+
+export function RequiredParamToggle(props: RequiredParamToggleProps) {
+  const { disabled, value, onChange, uniqueId, disabledTooltip } = props;
+  const id = `required_param_toggle_${uniqueId}`;
+
+  return (
+    <Flex gap="sm" mt="md">
+      <Toggle disabled={disabled} id={id} value={value} onChange={onChange} />
```
oleggromov, line 21: Right now, the `disabled` state in new `Switch` is broken. I tried to fix it but couldn't do it quickly as it requires some sorcery with the styles. So it stays the previous component for now.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -290,19 +291,24 @@ function getDefaultEmbeddingParams(
...
+  // (Editable) because this is the default for a required parameter.
+  // This is needed to save embedding_parameters when a user clicks
+  // "Publish" without changing parameter visibility.
+  return validRequiredParams.reduce((acc, param) => {
+    if (!acc[param.slug] || acc[param.slug] === "disabled") {
+      acc[param.slug] = "enabled";
+    }
+    return acc;
+  }, defaultParams);
 }
```
oleggromov, line 312: I previously used a different but incorrect approach, which set the default value to "editable" for required parameters only [in the `ParametersSettings` component](https://github.com/metabase/metabase/blob/7ee514a0546dc4f79476e379c094e8db89500402/frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx#L79). 

It was incorrect because that value was only used for rendering, and if a user published the embedding right away, without even changing any of the parameter visibility settings, the `"enabled"` won't be sent. This function fixes it.

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -383,6 +384,22 @@ export const getDashcardParameterMappingOptions = createCachedSelector(
   return props.card.id ?? props.dashcard.id;
 });
 
+// Embeddings might be published without passing embedding_params to the server,
+// in which case it's an empty object. We should treat such situations with
+// caution, assuming that an absent parameter is "disabled".
+export function getEmbeddedParameterVisibility(
```
WiNloSt, line 390: I'm not totally sure if we should have 2 selectors (for dashboard and question). Or have only 1 function that receives either an embedding resource or embedding parameters.

I think having only one function might be better because engineers could forget about another function when modifying one function. Both seem pretty coupled to me.

What do you think?

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/StaticEmbedSetupPane.tsx
```diff
@@ -322,7 +328,11 @@ function getPreviewParamsBySlug({
   return Object.fromEntries(
     lockedParameters.map(parameter => [
       parameter.slug,
-      parameterValues[parameter.id] ?? null,
+      getParameterValue({
+        parameter,
+        values: parameterValues,
+        defaultRequired: true,
+      }) ?? null,
```
WiNloSt, line 335: Doesn't `getParameterValue` already return `null`? Then we could drop the last part

```suggestion
      }),
```

WiNloSt (approved):
Overall, this looks good to me. I've left some comments.

nemanjaglumac:
This is a question aimed more at the product. Wouldn't it make more sense to not even show the "Disabled" option for the required parameters rather than disabling it?
image

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
```diff
@@ -95,9 +120,16 @@ export const ParametersSettings = ({
...
                   parameters={valuePopulatedLockedParameters}
-                  setValue={(value: string) => {
-                    onChangeParameterValue(parameter.id, value);
+                  setValue={(value: string) =>
+                    onChangeParameterValue(parameter.id, value)
+                  }
+                  setParameterValueToDefault={() => {
+                    onChangeParameterValue(
+                      parameter.id,
+                      parameter.default as any,
```
uladzimirdev, line 129: why any, can it be converted to string?

frontend/src/metabase/parameters/components/ParameterSettings/ParameterSettings.tsx
```diff
@@ -132,21 +135,33 @@ const ParameterSettings = ({
...
-            <SettingRequiredLabel
-              htmlFor={`parameter-setting-required_${parameter.id}`}
-            >
-              {t`Always require a value`}
-            </SettingRequiredLabel>
-            <p>{t`When enabled, people can change the value or reset it, but can't clear it entirely.`}</p>
-          </div>
-        </SettingRequiredContainer>
+        <RequiredParamToggle
+          // This forces the toggle to be a new instance when the parameter changes,
```
uladzimirdev, line 139: 👍

frontend/src/metabase-lib/parameters/utils/parameter-values.js
```diff
@@ -9,12 +9,39 @@ import {
...
+/**
+ * In some cases, we need to use default parameter value in place of an absent one.
+ * Please use this function when dealing with the required parameters.
+ */
+export function getParameterValue({
+  parameter,
+  values = {},
+  defaultRequired = false,
+}) {
+  const value = values?.[parameter.id];
```
uladzimirdev, line 21: do you expect `null` in place of values? otherwise there is a default value `{}` and `?.` can be removed

uladzimirdev (commented):
LGTM overall

e2e/support/helpers/e2e-embedding-helpers.js
```diff
@@ -197,6 +197,47 @@ export function openStaticEmbeddingModal({
...
+    callback?.(targetXhr);
+  });
+}
+
+export function getParametersContainer() {
+  return cy.findByLabelText("Configuring parameters");
+}
+
+export function setEmbeddingParameter(name, value) {
+  getParametersContainer().findByText(name).siblings("a").click();
```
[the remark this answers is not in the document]
oleggromov: It actually does, thank you!

frontend/src/metabase/dashboard/selectors.ts
oleggromov: You're right, it would be nice to avoid it. Unfortunately, we already have lots of duplication here and elsewhere for parameters and other parts of Query Builder and Dashboards.

So since this change is consistent with the rest, I'll keep it as is. The refactoring is coming but is outside the scope of this PR.

oleggromov:
@nemanjaglumac 
> This is a question aimed more at the product. Wouldn't it make more sense to not even show the "Disabled" option for the required parameters rather than disabling it?

We discussed it in one of the meetings/threads and agreed that it's better to keep it disabled. 
I can't find the link though.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
oleggromov: It can't, parameters are often arrays of values.

frontend/src/metabase-lib/parameters/utils/parameter-values.js
oleggromov: Yeah, you're right, it's some leftover.

e2e/support/helpers/e2e-embedding-helpers.js
```diff
@@ -197,6 +197,47 @@ export function openStaticEmbeddingModal({
...
+export function closeStaticEmbeddingModal() {
+  modal().icon("close").click();
+}
+
+/**
+ * Open Static Embedding setup modal
+ * @param apiPath - "card" or "dashboard"
+ * @param callback
+ */
+export function publishChanges(apiPath, callback) {
```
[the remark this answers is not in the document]
oleggromov: I honestly don't think it's such a big issue but feel free to rename it if stays confusing :)

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
uladzimirdev: okay, but can you come up with a better type than `any`?

frontend/src/metabase-lib/parameters/utils/parameter-values.js
uladzimirdev: and tests failed 😄

oleggromov: Yeah, this is why I had it in the first place — won't touch it now though.

nemanjaglumac (approved):
Others have already reviewed the code. I just want to confirm that this works as advertised. I was playing with it and tried to break it in many ways, but it's solid.

frontend/src/metabase/public/components/EmbedModal/StaticEmbedSetupPane/ParametersSettings.tsx
oleggromov: I definitely will — not that I am discarding your suggestion.
But if you don't mind, as a part of a separated effort to tidy up the parameters everywhere 🙏

PR #38396, opened 2024-02-02
Implicitly delete obsolete FieldValues when fetching the current ones

crisptrutski:
Relates to https://github.com/metabase/metabase/issues/668

### Description

Before this change it was non-deterministic which FieldValues would be used in the presence of stale records. This change ensures that we always use the values that were last inserted, and delete the older rows in the process. 

It also fixes a query that failed to specify the `:type`, and so was totally non-deterministic.

### Checklist

- [x] Improve test quality
- [x] ~~Implement a Snowplow event to track when customer databases encounter duplicates~~ Going to use idempotent operations for all inserts, which will make it safe to insert the constraint. Don't think events will be worth the faff.

### DIFF
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -493,7 +493,7 @@
 (defmethod serdes/load-find-local "FieldValues" [path]
   ;; Delegate to finding the parent Field, then look up its corresponding FieldValues.
   (let [field (serdes/load-find-local (pop path))]
-    (t2/select-one FieldValues :field_id (:id field))))
+    (t2/select-one FieldValues :field_id (:id field) :type :full)))
 
 (defmethod serdes/load-update! "FieldValues" [_ ingested local]
   ;; It's illegal to change the :type and :hash_key fields, and there's a pre-update check for this.
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -334,6 +334,25 @@
       (log/error e (trs "Error fetching field values"))
       nil)))
 
+(defn- report-and-fix-duplicates! [rows]
+  (if (<= (count rows) 1)
+    (first rows)
+    ;; todo - rather use a descending comparator
+    (let [[latest & duplicates] (reverse (sort-by :updated-at rows))]
+      ;; send telemetry to snowplow
+      (t2/delete! FieldValues :id [:in (map :id duplicates)])
+      latest)))
+
+(defn- get-field-values-for-field
+  [field-id type hash]
+  ;; todo - we could rather put this validation in a toucan select hook
+  (assert (= (nil? hash) (= type :full)) ":hash must be nil iff :type is :full")
+  (report-and-fix-duplicates!
+    (t2/select FieldValues :field_id field-id :type type :hash_key hash)))
+
+(defn- get-full-field-values-for-field [field-id]
+  (get-field-values-for-field field-id :full nil))
+
 (defn create-or-update-full-field-values!
   "Create or update the full FieldValues object for `field`. If the FieldValues object already exists, then update values for
    it; otherwise create a new FieldValues object with the newly fetched values. Returns whether the field values were
@@ -341,7 +360,7 @@
 
   Note that if the full FieldValues are create/updated/deleted, it'll delete all the Advanced FieldValues of the same `field`."
   [field & [human-readable-values]]
-  (let [field-values              (t2/select-one FieldValues :field_id (u/the-id field) :type :full)
+  (let [field-values              (get-full-field-values-for-field (u/the-id field))
         {unwrapped-values :values
          :keys [has_more_values]} (distinct-values field)
         ;; unwrapped-values are 1-tuples, so we need to unwrap their values for storage
@@ -411,20 +430,19 @@
   [{field-id :id field-values :values :as field} & [human-readable-values]]
   {:pre [(integer? field-id)]}
   (when (field-should-have-field-values? field)
-    (let [existing (or (not-empty field-values)
-                       (t2/select-one FieldValues :field_id field-id :type :full))]
+    (let [existing (or (not-empty field-values) (get-full-field-values-for-field field-id))]
       (if (or (not existing) (inactive? existing))
         (case (create-or-update-full-field-values! field human-readable-values)
           ::fv-deleted
           nil
 
           ::fv-created
-          (t2/select-one FieldValues :field_id field-id :type :full)
+          (get-full-field-values-for-field field-id)
 
           (do
             (when existing
               (t2/update! FieldValues (:id existing) {:last_used_at :%now}))
-            (t2/select-one FieldValues :field_id field-id :type :full)))
+            (get-full-field-values-for-field field-id)))
         (do
           (t2/update! FieldValues (:id existing) {:last_used_at :%now})
           existing)))))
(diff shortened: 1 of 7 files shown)


src/metabase/models/field_values.clj
```diff
@@ -334,14 +334,33 @@
...
+    ;; todo - rather use a descending comparator
+    (let [[latest & duplicates] (reverse (sort-by :updated-at rows))]
+      ;; send telemetry to snowplow
+      (t2/delete! FieldValues :id [:in (map :id duplicates)])
+      latest)))
+
+(defn- get-field-values-for-field
+  [field-id type hash]
+  ;; todo - we could rather put this validation in a toucan select hook
+  (assert (= (nil? hash) (= type :full)) ":hash must be nil iff :type is :full")
```
calherries, line 349:
```suggestion
  (assert (= (nil? hash) (= type :full)) ":hash_key must be nil iff :type is :full")
```

src/metabase/models/field_values.clj
```diff
@@ -334,14 +334,33 @@
...
+(defn- report-and-fix-duplicates! [rows]
+  (if (<= (count rows) 1)
+    (first rows)
+    ;; todo - rather use a descending comparator
+    (let [[latest & duplicates] (reverse (sort-by :updated-at rows))]
+      ;; send telemetry to snowplow
+      (t2/delete! FieldValues :id [:in (map :id duplicates)])
+      latest)))
+
+(defn- get-field-values-for-field
```
calherries, line 346: This function name should reflect that it only selects one field values instance. How about `select-one-field-values`?

calherries (commented):
solid start

src/metabase/models/field_values.clj
crisptrutski: How about `select-latest-field-values`?

crisptrutski: I'm a bit torn about this, as needing to pick from multiple rows is an anomaly and I wanted to hide that detail away.

... or maybe I should ilt in the other direction and mention the inner hack in the name, maybe something like `get-latest-field-values-and-delete-stale-rows!` 🤔

calherries: better!

qnkhuat:
Is it better to write a migration to delete duplicates then add a unique constraint on `type, hash`?

crisptrutski:
> Is it better to write a migration to delete duplicates then add a unique constraint on `type, hash`?

That's the end goal - see discussion on https://github.com/metabase/metabase/pull/37794

src/metabase/models/field_values.clj
```diff
@@ -493,7 +516,8 @@
 (defmethod serdes/load-find-local "FieldValues" [path]
   ;; Delegate to finding the parent Field, then look up its corresponding FieldValues.
   (let [field (serdes/load-find-local (pop path))]
-    (t2/select-one FieldValues :field_id (:id field))))
```
crisptrutski, line 496: Note: this was a subtle bug fix

crisptrutski:
* `master`
  * **PR #38396** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38396?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38396?utm_source=stack-comment).

test/metabase/models/field_values_test.clj
```diff
@@ -128,6 +128,24 @@
   (sync/sync-database! db)
   (find-values field-values-id))
 
+(deftest implicit-deduplication-test
+  (let [now (t/zoned-date-time)
+        then (t/plus now (t/millis 1))]
+    (mt/with-temp [Database    {database-id :id} {}
+                   Table       {table-id :id}    {:db_id database-id}
+                   Field       {field-id :id}     {:table_id table-id}
```
calherries, line 136:
```suggestion
                   Field       {field-id :id}    {:table_id table-id}
```

test/metabase/models/field_values_test.clj
```diff
@@ -128,6 +128,24 @@
   (sync/sync-database! db)
   (find-values field-values-id))
 
+(deftest implicit-deduplication-test
+  (let [now (t/zoned-date-time)
+        then (t/plus now (t/millis 1))]
+    (mt/with-temp [Database    {database-id :id} {}
```
calherries, line 134: nit: technically vars for model identifiers are deprecated and apparently meant to be removed one day. We should honor that (even if I don't believe it will happen soon)

e.g.
```suggestion
    (mt/with-temp [:model/Database    {database-id :id} {}
```

test/metabase/models/field_values_test.clj
```diff
@@ -128,6 +128,24 @@
...
+                   Table       {table-id :id}    {:db_id database-id}
+                   Field       {field-id :id}     {:table_id table-id}
+                   FieldValues _                 {:field_id field-id :type :full :values ["a" "b"] :human_readable_values ["A" "B"] :updated_at now}
+                   FieldValues _                 {:field_id field-id :type :full :values ["c" "d"] :human_readable_values ["C" "D"] :updated_at then}]
+
+      (testing "When we have two FieldValues rows in the database, "
+        (is (= 2 (count (t2/select FieldValues :field_id field-id :type :full :hash_key nil))))
+        (testing "we always return the most recently inserted row"
```
calherries, line 142:
```suggestion
        (testing "we always return the most recently updated row"
```

test/metabase/models/field_values_test.clj
```diff
@@ -128,6 +128,24 @@
...
+          (is (= ["C" "D"] (:human_readable_values (field-values/get-latest-full-field-values field-id))))
```
calherries, line 143: This test could pass by coincidence if the DB's query implementation selects the first inserted row. A stronger test would check that the last row by `updated_at` is returned regardless of insertion order. Perhaps this would be sufficient: insert three field values rows, where the middle one has the latest updated_at date.

src/metabase/models/field_values.clj
```diff
@@ -334,14 +334,38 @@
       (log/error e (trs "Error fetching field values"))
       nil)))
 
+(defn- delete-duplicates-and-return-latest!
+  "This is a workaround for the issue of stale FieldValues rows (metabase#668)
+  In order to mitigate the impact of duplicates, we return the most recently updated row, and deleted the older rows."
```
calherries, line 339:
```suggestion
  In order to mitigate the impact of duplicates, we return the most recently updated row, and delete the older rows."
```

calherries (approved):
Awesome. Just left some nits. This is strictly better than the current situation where there's unpredictable behaviour in the case of duplicate `FieldValues` instances. The downside of this approach is it's not particularly future-proof to stop the original problem happening again. We'll need the uniqueness constraint and a toucan hook to require `:type` in a select for that.

crisptrutski:
The plan is indeed to follow with a db constraint and hooks, I just feel it's worth fixing the pain sooner than later, and the new "higher level" mutation methods will encapsulate the retries for concurrent inserts.

test/metabase/models/field_values_test.clj
crisptrutski: Good call-out

PR #38404, opened 2024-02-02
[Metrics v2] Add "Metric" to the new entity menu

### LINKED ISSUE #37352: FE - Add "Metric" option to "+ New" dropdown menu (+ `/metric/notebook` route boilerplate)
Depends on:
- #37348

----

- [x] Add new option to the menu. It should match visibility settings of models & questions.
- [x] Add icon for the metric.
- [x] Add a new empty `/metric/notebook` route and redirect there on click.

nemanjaglumac:
This PR adds a new option/entry to the "New" entity menu. This time for a metric.
It also adds a route for `/metric/query` that currently defaults to the `QueryBuilder` component.

image

`/metric/query` starts at the data selection point
image

New Metric flow from the home page
https://www.loom.com/share/5e5a2d7f0eb44a089ec6c9f35326d645?sid=2502275b-f81d-45d2-a069-796593f3fba4

New Metric flow from the collection
https://www.loom.com/share/4884a373224a472194fdfc834ce91d90?sid=7802db85-9c6e-4ce7-bff8-c9e193da74a7

Resolves #37352

### DIFF
--- a/frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
+++ b/frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
@@ -109,6 +109,7 @@ const NewItemMenu = ({
         action: () => setModal("new-collection"),
       },
     );
+
     if (hasNativeWrite) {
       const collectionQuery = collectionId
         ? `?collectionId=${collectionId}`
@@ -127,6 +128,15 @@ const NewItemMenu = ({
         title: t`Action`,
         icon: "bolt",
         action: () => setModal("new-action"),
+      });
+    }
+
+    if (hasDataAccess) {
+      items.push({
+        title: t`Metric`,
+        icon: "metric",
+        link: "/metric/query",
+        onClose: onCloseNavbar,
       });
     }
 
--- a/frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
+++ b/frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
@@ -135,7 +135,11 @@ const NewItemMenu = ({
       items.push({
         title: t`Metric`,
         icon: "metric",
-        link: "/metric/notebook",
+        link: Urls.newQuestion({
+          mode: "notebook",
+          collectionId,
+          cardType: "metric",
+        }),
         onClose: onCloseNavbar,
       });
     }
--- a/frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
+++ b/frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
@@ -135,7 +135,7 @@ const NewItemMenu = ({
       items.push({
         title: t`Metric`,
         icon: "metric",
-        link: "/metric/query",
+        link: "/metric/notebook",
         onClose: onCloseNavbar,
       });
     }
--- a/frontend/src/metabase/lib/urls/questions.ts
+++ b/frontend/src/metabase/lib/urls/questions.ts
@@ -107,9 +107,10 @@ export function newQuestion({
   });
 
   const entity = question.isDataset() ? "model" : "question";
+  const type = question.type() ?? entity;
 
   if (mode) {
-    return url.replace(/^\/(question|model)/, `/${entity}\/${mode}`);
+    return url.replace(/^\/(question|model)/, `/${type}\/${mode}`);
   } else {
     return url;
   }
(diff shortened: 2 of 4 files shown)


frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
```diff
@@ -130,6 +131,15 @@ const NewItemMenu = ({
       });
     }
 
+    if (hasDataAccess) {
```
nemanjaglumac, line 134: Even though #37352 says
> Add new option to the menu. **It should match visibility settings of models & questions.**

I've used only data access as a prerequisite here. Models have an additional requirement for the native write permissions, but that's because you can choose to create a model either from a GUI or from a native query. 

Metrics v2 are not possible to create from the native queries. At least not for the forseeable future.

frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
```diff
@@ -130,6 +131,15 @@ const NewItemMenu = ({
...
+      items.push({
+        title: t`Metric`,
+        icon: "metric",
+        link: "/metric/query",
```
kamilmielnik, line 138: 1. My apologies for mis-specifying this in the issue description :pray: (updated now)
  We should follow the same pattern as for questions: when creating a new question user lands on `/question/notebook`.

```suggestion
        link: "/metric/notebook",
```

2. Notice how URL for new question is generated with `Urls.newQuestion`. We need something like this as well to preserve all similar behaviors, e.g.
    - `collectionId` parameter
    - putting all essential attributes in URL, so that when user refreshes the page, they're still there

```suggestion
        link: generateMetricsV2Url({ collectionId }),
```

_(please treat the above suggestion as pseudocode)_

nemanjaglumac: Thanks for the further clarification.
I think we can use the existing `Urls.newQuestion` function with minimal modifications. The link for the new metric is absolutely the same as for the new notebook question. The only difference is that the URL should show `/metric/notebook#ey...` instead of `/question/notebook#ey...`

We already have the URL modification inside this function, based on whether we're dealing with the model (dataset) or a question. Please see:
https://github.com/metabase/metabase/blob/37352-new-metric/frontend/src/metabase/lib/urls/questions.ts#L109-L115

We can go two ways about this:
1. temporary hack
2. using the `card.type` (requires a functioning endpoint)

**Temporary hacks**
We can start explicitly passing the `entity` as a new argument to this function? This approach would require updating code in a lot of places, but it might be cleaner and more explicit in the end.
It might even tie well with #38127

Another approach could be to "sneak in" a metric keyword in `options`, but that's a blatant hack. The idea is that `Question.create()` method will ignore this keyword, and we might use it in the function itself to alter the URL.

**Proper way**
Once we have a proper endpoint, we can use `card.type` for this.
That would be the most correct of achieving this.

```ts
type CardType = "question" | "model" | "metric";
```

kamilmielnik: For the record: we had a huddle and agreed to use `card.type` because it's already available (#38457).

frontend/src/metabase/lib/urls/questions.ts
```diff
@@ -107,9 +107,10 @@ export function newQuestion({
   });
 
   const entity = question.isDataset() ? "model" : "question";
+  const type = question.type() ?? entity;
 
   if (mode) {
-    return url.replace(/^\/(question|model)/, `/${entity}\/${mode}`);
+    return url.replace(/^\/(question|model)/, `/${type}\/${mode}`);
```
nemanjaglumac, line 113: I'm relying on the existing behavior, and the fact that the metric is always initialized in a notebook mode.
There's weird and convoluted function execution order involved here.

We could be dealing with this logic inside the underlying `question` function
https://github.com/metabase/metabase/blob/37352-new-metric/frontend/src/metabase/lib/urls/questions.ts#L27

So why aren't we?
With the presumption that there was a good reason for it, I'll solve the logic here. But I'll make sure to follow up, and to investigate this call stack in the following PR.

cc @kamilmielnik

frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
```diff
@@ -135,7 +135,11 @@ const NewItemMenu = ({
       items.push({
         title: t`Metric`,
         icon: "metric",
-        link: "/metric/notebook",
+        link: Urls.newQuestion({
+          mode: "notebook",
+          collectionId,
+          cardType: "metric",
+        }),
```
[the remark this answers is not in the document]
nemanjaglumac: @cdeweyx said not to worry about this right now. So this should be good to go.

kamilmielnik (approved):

PR #38413, opened 2024-02-02
Implement equality for MLv2 MetadataProviders 

camsaul:
I split this off from #36068 since it was unrelated and that PR is way too big anyway.

Two equivalent MLv2 queries aren't always `=` to each other because equivalent metadata providers aren't `=`, so I implemented `Object.equals` (Clojure) / `IEquiv.-equiv` (ClojureScript) methods for all of our metadata provider types. This required changing some of them from anonymous `reify`s to actual `deftypes`. Now two equivalent metadata providers are equal to one another. Nice

This is mostly for REPL convenience since MLv2 queries constructed identically aren't usually equal anyway since they'll have different UUIDs but it's still a good change IMO

This was already reviewed by @metamben as part of #36068 and I applied the PR feedback he made there as part of this PR

### DIFF
--- a/src/metabase/lib/metadata/cached_provider.cljc
+++ b/src/metabase/lib/metadata/cached_provider.cljc
@@ -9,6 +9,8 @@
    [metabase.util.malli :as mu]
    #?@(:clj ([pretty.core :as pretty]))))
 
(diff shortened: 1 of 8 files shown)


metamben (approved):
👍

PR #38421, opened 2024-02-05
Organize `ViewHeader` component and its related files

nemanjaglumac:
This PR is a follow up after #38328.

It does two things only:
- Moves all related files to a single folder inside `metabase/query_builder/components`
- Renames utils (and its unit tests) and converts them to TS


kamilmielnik (dismissed):

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -44,7 +44,7 @@ import {
   QueryBuilderViewRoot,
   StyledDebouncedFrame,
   StyledSyncedParametersList,
-} from "./View.styled";
+} from "./ViewHeader/View.styled";
```
kamilmielnik, line 47: Shouldn't `View.styled.tsx` stay in the same directory as `View.jsx` ?

nemanjaglumac: Yep, my bad. Fat fingers :)
Thanks for catching this.

frontend/src/metabase/query_builder/components/view/ViewHeader/ViewHeader.jsx
```diff
@@ -17,16 +17,16 @@ import SavedQuestionHeaderButton from "metabase/query_builder/components/SavedQu
...
+import QuestionActions from "../../QuestionActions";
+import { ExploreResultsLink } from "../ExploreResultsLink";
+import { FilterHeaderButton } from "../FilterHeaderButton";
+import { HeadBreadcrumbs } from "../HeaderBreadcrumbs";
+import QuestionDataSource from "../QuestionDataSource";
+import QuestionDescription from "../QuestionDescription";
+import { QuestionNotebookButton } from "../QuestionNotebookButton";
+import ConvertQueryButton from "../ConvertQueryButton";
+import { FilterHeaderToggle, FilterHeader } from "../QuestionFilters";
+import { QuestionSummarizeWidget } from "../QuestionSummaries";
```
kamilmielnik, line 29: Some of these components are only used in `ViewHeader.jsx` (same goes for imports in `ViewHeader.styled.tsx`.
We could benefit by moving them to the same directory.
WDYT?
We can do this later too.

kamilmielnik (approved):

frontend/src/metabase/query_builder/components/view/ViewHeader/ViewHeader.jsx
nemanjaglumac: I like this idea.
If I'm reading the internal RFC 42 about the folder structure correctly, these should go inside the nested components folder?

`frontend/src/metabase/query_builder/components/view/ViewHeader/components/...`

kamilmielnik: > If I'm reading the internal RFC 42 about the folder structure correctly, these should go inside the nested components folder?

I think the RFC does not solve this particular case, but I like your proposal :+1:

uladzimirdev: I'd not move files without a clear reason. we create `components` folder for top level entities like dashboard, as far as I see.

I don't have a strong opinion here though

nemanjaglumac: Sorry @uladzimirdev - I didn't see your comment before I pushed commits https://github.com/metabase/metabase/pull/38421/commits/114b2539f1a816ad62874af21fee7616e5bd2ff4 to https://github.com/metabase/metabase/pull/38421/commits/dd7ed3fd66d73f5fd2d7f2e25eb43198f063d5b6.

I'll add this proposal to the RFC.
It makes for **much** cleaner folder structure and it provides an easier overview, imo.

nemanjaglumac: Look how much cleaner this is
```
frontend/src/metabase/query_builder/components/view/ViewHeader
├── ViewHeader.jsx
├── ViewHeader.styled.tsx
├── ViewHeader.unit.spec.js
├── components
│   ├── ConvertQueryButton
│   ├── ExploreResultsLink.tsx
│   ├── FilterHeaderButton.tsx
│   ├── HeaderBreadcrumbs
│   ├── QuestionActions
│   ├── QuestionDataSource
│   ├── QuestionDescription
│   ├── QuestionFilters
│   ├── QuestionNotebookButton
│   └── QuestionSummarizeWidget.jsx
├── index.ts
├── utils.ts
└── utils.unit.spec.ts
```

frontend/src/metabase/query_builder/components/view/ViewHeader/ViewHeader.jsx
```diff
@@ -17,16 +17,17 @@ import SavedQuestionHeaderButton from "metabase/query_builder/components/SavedQu
...
+import { ConvertQueryButton } from "./components/ConvertQueryButton";
+import { HeadBreadcrumbs } from "./components/HeaderBreadcrumbs";
+import { FilterHeaderButton } from "./components/FilterHeaderButton";
+import { FilterHeaderToggle, FilterHeader } from "./components/QuestionFilters";
+import { ExploreResultsLink } from "./components/ExploreResultsLink";
+import { QuestionActions } from "./components/QuestionActions";
+import { QuestionNotebookButton } from "./components/QuestionNotebookButton";
+import { QuestionDataSource } from "./components/QuestionDataSource";
+import { QuestionDescription } from "./components/QuestionDescription";
+import { QuestionSummarizeWidget } from "./components/QuestionSummarizeWidget";
```
kamilmielnik, line 30: This could be a single `import` statement if we had `./components/index.ts` : >

PR #38425, opened 2024-02-05
Add questions caching layer to improve dashboard performance after migration to MLv2

uladzimirdev:
Solves dashboard performance problem described at https://github.com/metabase/metabase/issues/38225

## Details

We have two places where calculation of parameter mapping happens:

1.  `getDashboardUiParameters` - a function which calculates how to display dashboard filters and Dashcard mapping
2. `getParameterMappingOptions` - as a part of auto-wiring logic, it calculates mapping for Dashcard mapping displaying

Both functions require a lot of mlv2 calls, both are triggered frequently, per any change in the dashboard state (also in metadata)

Results I got from my laptop - MacBook Pro M2 max
The app is built in Prod mode, CPU is 4x slowdown

### Before

<details><summary>Screenshots</summary>
<p>
Screenshot 2024-02-07 at 12 32 29
Screenshot 2024-02-07 at 12 32 22
Screenshot 2024-02-07 at 12 32 13
Screenshot 2024-02-07 at 12 32 03

</p>
</details> 

### After

<details><summary>Screenshots</summary>
<p>

Screenshot 2024-02-07 at 12 26 38

Screenshot 2024-02-07 at 12 26 23

Screenshot 2024-02-07 at 12 23 36

Screenshot 2024-02-07 at 12 23 26

Screenshot 2024-02-07 at 12 33 20

</p>
</details>


frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -1,5 +1,6 @@
 import _ from "underscore";
 import { createSelector } from "@reduxjs/toolkit";
+import { createCachedSelector } from "re-reselect";
```
kulyk, line 3: We can't use `reselect` for this yet, right? Do we need to be on v5?

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -301,13 +303,44 @@ export const getParameterTarget = createSelector(
...
+      const dashcard = getDashCardById(state, dashcardId);
+
+      const question = getQuestionByDashcardMemo(state, { dashcard });
+
+      acc[dashcard.id] = question;
+
+      return acc;
+    },
+    {},
+  );
```
kulyk, line 326: I'm afraid a dashcard ID won't be enough here. Dashcards can include > 1 card (and each card can be mapped to totally different parameters), so we'd need a `card_id` as well

You can add more cards to a dashboard using the "Add series" flow ([see more in docs](https://www.metabase.com/docs/latest/dashboards/multiple-series#combining-two-saved-questions)). IIRC, we'd store them in `dashcard.series` array. Also, that's the reason why `parameter_mappings` have `card_id`

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -301,13 +303,44 @@ export const getParameterTarget = createSelector(
...
+
+  return questionsById;
+};
+
 export const getParameters = createSelector(
-  [getDashboardComplete, getMetadata],
-  (dashboard, metadata) => {
+  [getDashboardComplete, getMetadata, getQuestions],
+  (dashboard, metadata, questions) => {
```
kulyk, line 333: Not sure if it makes sense after all the performance fixes, but this selector doesn't really need to be "subscribed" to an entire dashboard object, we only care about `dashboard.parameters` and `parameter_mappings` in dashcards

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -321,12 +354,41 @@ export const getMissingRequiredParameters = createSelector(
...
+    let question;
+
+    if (dashcard.card_id) {
+      question = q;
+    } else {
+      // TODO: probably we can remove it as dashcards are text or heading or action?
+      question = new Question(card, metadata);
+    }
+
+    return _getParameterMappingOptions(question, parameter, card, dashcard);
```
kulyk, line 387: Yeah, we might want to use `isVirtualDashCard` here to skip heading, text, link cards, etc.
I can't recall if we're using this code for actions, though. It's definitely possible to map dashboard filter values to action parameters, just not sure which code is doing that

kulyk (commented):
Overall looks correct to me! I think the key thing to fix is [support for multi-series dashcards](https://github.com/metabase/metabase/pull/38425#discussion_r1478421767)

frontend/src/metabase/dashboard/selectors.ts
uladzimirdev: I can't say exactly, I just saw that re-reselect is a part of our package.json and found its usage very straightforward for our use case. We indeed use reselect v4 that comes with RTK and v5 has more features and configurations, but I didn't dig into it

frontend/src/metabase/dashboard/selectors.ts
uladzimirdev: okay, I got it now, we can have multiple cards and because of that the unique key should be dashcard and not card. 

And every dashcard has cards from series + cards from card property

uladzimirdev: @kulyk please check 6d82acc3033eb66d172ec18332fd3a2d66e98b9d

frontend/src/metabase/dashboard/selectors.ts
uladzimirdev: I'll add a separate task for it

frontend/src/metabase/dashboard/selectors.ts
uladzimirdev: as far as I understand, we don't need to create a question and pass question for virtual card and action card

frontend/src/metabase-lib/parameters/utils/targets.ts
```diff
@@ -62,8 +62,13 @@ export function getTargetFieldFromCard(
...
+  question: Question,
 ) {
-  const question = new Question(card, metadata);
+  if (!question) {
+    // Fallback
+    question = new Question(card, metadata);
+  }
+
   const field = getParameterTargetField(target, question);
   return field ?? null;
```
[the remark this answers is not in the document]
uladzimirdev: removed

kamilmielnik (commented):
Code looks good :+1: 
I'll leave it to Anton to approve as he's way more familiar with dashboards.

----

Can you please share measurements related to the performance improvement?

frontend/src/metabase/dashboard/actions/auto-wire-parameters/utils.ts
```diff
@@ -48,19 +48,25 @@ export function getMatchingParameterOption(
...
 ): {
   target: ParameterTarget;
 } | null {
   if (!targetDashcard) {
     return null;
   }
 
-  const sourceQuestion = new Question(sourceDashcard.card, metadata);
-  const targetQuestion = new Question(targetDashcard.card, metadata);
+  // TODO: verify if fallback can be dropped
```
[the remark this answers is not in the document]
uladzimirdev: not really a leftover, I'm not certainly sure it can be dropped

frontend/src/metabase/dashboard/selectors.ts
```diff
@@ -321,12 +356,26 @@ export const getMissingRequiredParameters = createSelector(
...
-);
+)((_state, props) => {
+  return props.card.id;
+});
+
+export const getDashcardParameterMappingOptions = createCachedSelector(
+  [getMetadata, getEditingParameter, getCard, getDashCard, getQuestionByCard],
+  (metadata, parameter, card, dashcard, question) => {
+    return _getParameterMappingOptions(question, parameter, card, dashcard);
+  },
```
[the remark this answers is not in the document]
uladzimirdev: @kulyk do you think we shouldn't subscribe to metadata changes in this case?

uladzimirdev:
@kamilmielnik added some screenshots

frontend/src/metabase/dashboard/actions/auto-wire-parameters/utils.ts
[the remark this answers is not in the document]
kulyk: Looks safe to me. Don't think auto-wiring can be triggered before we have the `questions` map. Also types should give us a bit more confidence?

frontend/src/metabase/dashboard/selectors.ts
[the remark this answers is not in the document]
kulyk: I don't think so — it's subscribed to questions, and a question needs metadata to be instantiated, so it should have it by that point

frontend/src/metabase/parameters/utils/dashboards.ts
```diff
@@ -133,15 +144,21 @@ function buildFieldFilterUiParameter(
...
   metadata: Metadata,
+  questions: Record<CardId, Question>,
 ): FieldFilterUiParameter {
   const mappingsForParameter = mappings.filter(
     mapping => mapping.parameter_id === parameter.id,
   );
-  const mappedFields = mappingsForParameter.map(mapping => {
+  const uniqueMappingsForParameters = _.uniq(mappingsForParameter, mapping =>
+    mapping.target.toString(),
+  );
```
kulyk, line 154: I'm worried this can be dangerous and unpredictable. For example, a mapping target like that:

```js
const field_ref = [ "field", 1, { 'base-type': 'type/Integer', 'source-field': 40 } ]
const target = [ "dimension", field_ref ]
```

will turn into:

```js
"dimension,field,1,[object Object]"
```

Even though we should most likely ignore this options object when comparing field refs, I'm worried this could be too implicit

IMO we could use something like `_.isEqual` to find unique mappings (probably not gonna work with `_.uniq` tho). Or ideally have a CLJS method to do that

kulyk (commented):
Looks and works great!
I'm worried the way we compute `uniqueMappingsForParameters` is a blocker tho

frontend/src/metabase/parameters/utils/dashboards.ts
uladzimirdev: Fixed a91692bf68a11923b3f35d8615c9af29bdbe09ee

Based on my testing, we care only about the first two elements of the target, but let's stay on the safe side and improve it later

kulyk (approved):
:shipit:

PR #38430, opened 2024-02-05
Create X-or-insert! methods and use them for FieldValues

crisptrutski:

### Description

**Version 1:** Use a database transaction around the updates. 

*Problem: our default transactions aren't serializable.*

**Version 2:** Use a serializable transaction. 

*Problem: our tests use nested transactions. For some databases their notion of serializable is also still too weak.*

**Version 3:** Give up on atomicity without a database constraint.

This change has been decreased in scope to just improve ergonomics, and improve constraint forward compatibility.

### DIFF
--- a/src/metabase/db/util.clj
+++ b/src/metabase/db/util.clj
@@ -81,14 +81,16 @@
            (.setTransactionIsolation ^Connection conn# Connection/TRANSACTION_SERIALIZABLE)
            (t2/with-transaction [~'_conn]
              ;; We need to try select the row again now that we're in the transaction to track the dependency.
-             (or ~select-expr
-                 ;; ... and then we can execute the (potentially expensive) mutating branch.
-                 ~insert-expr)))
+             (u/or-with some?
+                      ~select-expr
+                      ;; ... and then we can execute the (potentially expensive) mutating branch.
+                      ~insert-expr)))
          (catch ExceptionInfo e#
            ;; We cannot introspect the exception cause definitively, as it will be driver specific and typically opaque,
            ;; but we should find a result now if we try selecting again after a concurrent modification exception.
-           (or ~select-expr
-               (throw (ex-info "Unable to find element after attempting an idempotent-insert!"
-                               {:select-expr '~select-expr :insert-expr '~insert-expr}
-                               ;; Expose the exception - it might not be due to a concurrent modification.
-                               e#)))))))
+           (u/or-with some?
+             ~select-expr
+             (throw (ex-info "Unable to find element after attempting an idempotent-insert!"
+                             {:select-expr '~select-expr :insert-expr '~insert-expr}
+                             ;; Expose the exception - it might not be due to a concurrent modification.
+                             e#)))))))
--- a/src/metabase/db/util.clj
+++ b/src/metabase/db/util.clj
@@ -5,7 +5,10 @@
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
    [toucan2.core :as t2]
-   [toucan2.model :as t2.model]))
+   [toucan2.model :as t2.model])
+  (:import
+   (clojure.lang ExceptionInfo)
+   (java.sql Connection)))
 
 (defn toucan-model?
   "Check if `model` is a toucan model."
@@ -66,3 +69,26 @@
   ;;    [:in :semantic_type #{"type/URL" "type/ImageURL"}]
   ([expr type-keyword]
    [:in expr (type-keyword->descendants type-keyword)]))
+
+(defmacro idempotent-insert!
+  "Upsert a database record where the computation may be expensive etc. Blah blah toucan.
+   This code is agnostic as to whether there is an underlying db constriant that would prevent duplicates."
+  [select-expr insert-expr]
+  ;; First attempt the select without a serializable transaction, since those are expensive.
+  `(or ~select-expr
+       (try
+         (t2/with-connection [conn#]
+           (.setTransactionIsolation ^Connection conn# Connection/TRANSACTION_SERIALIZABLE)
+           (t2/with-transaction [~'_conn]
+             ;; We need to try select the row again now that we're in the transaction to track the dependency.
+             (or ~select-expr
+                 ;; ... and then we can execute the (potentially expensive) mutating branch.
+                 ~insert-expr)))
+         (catch ExceptionInfo e#
+           ;; We cannot introspect the exception cause definitively, as it will be driver specific and typically opaque,
+           ;; but we should find a result now if we try selecting again after a concurrent modification exception.
+           (or ~select-expr
+               (throw (ex-info "Unable to find element after attempting an idempotent-insert!"
+                               {:select-expr '~select-expr :insert-expr '~insert-expr}
+                               ;; Expose the exception - it might not be due to a concurrent modification.
+                               e#)))))))
--- a/src/metabase/db/util.clj
+++ b/src/metabase/db/util.clj
@@ -71,8 +71,15 @@
    [:in expr (type-keyword->descendants type-keyword)]))
 
 (defmacro idempotent-insert!
-  "Upsert a database record where the computation may be expensive etc. Blah blah toucan.
-   This code is agnostic as to whether there is an underlying db constriant that would prevent duplicates."
+  "Upsert some database state, typically a single row.
+
+   This is more general than an `UPSERT` or `INSERT .. ON CONFLICT`, in that it can be used for multiple entities,
+   e.g. where we also upsert some parent resources. It is also useful in the case where calculating the data to insert
+   is expensive, which could not be done if sending a single SQL command.
+
+   The usage is just like the naive pattern `(or select-expr insert-expr)`, just papering over a number of sharp edges.
+
+   The mechanism is agnostic as to whether there is an underlying db constraint to prevent duplicates."
   [select-expr insert-expr]
   ;; First attempt the select without a serializable transaction, since those are expensive.
   `(or ~select-expr
--- a/src/metabase/db/util.clj
+++ b/src/metabase/db/util.clj
@@ -71,11 +71,11 @@
    [:in expr (type-keyword->descendants type-keyword)]))
 
 (defmacro idempotent-insert!
-  "Upsert some database state, typically a single row.
+  "Create or update some database state, typically a single row, atomically.
 
    This is more general than an `UPSERT` or `INSERT .. ON CONFLICT`, in that it can be used for multiple entities,
-   e.g. where we also upsert some parent resources. It is also useful in the case where calculating the data to insert
-   is expensive, which could not be done if sending a single SQL command.
+   e.g. where we also upsert some parent resources. It is also useful if we want to avoid calculating the data to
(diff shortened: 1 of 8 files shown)


crisptrutski:
* `master`
  * **PR #38430** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38430?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38430?utm_source=stack-comment).

tsmacdonald (approved):
:boat:

qnkhuat (changes requested):
not sure what this fix, the issue of duplicating field values we talked about happens when there are 2 requests that try to create an advanced field value at the same time


```clojure
(use 'metabase.models.params.field-values)

(let [fetch-advanced-field-values-orig fetch-advanced-field-values
      field                            (t2/select-one :model/Field :base_type :type/Text :has_field_values :auto-list)
      constraints                      [{:field-id 1 :op :> :value 2}]
      hash-key                         (str (hash [(:id field) constraints]))]
  (t2/delete! :model/FieldValues :field_id (:id field) :type :linked-filter)
 (with-redefs [fetch-advanced-field-values (fn [& args]
                                             (Thread/sleep 1000)
                                             (apply fetch-advanced-field-values-orig args))]
   (let [f1 (future (create-advanced-field-values! :linked-filter field hash-key constraints))
         f2 (future (create-advanced-field-values! :linked-filter field hash-key constraints))]
     @f1 @f2
     (= 1 (t2/count :model/FieldValues :field_id (:id field) :type :linked-filter)))))
     
```

This is a snippet to reproduce that

qnkhuat:
and this is how to reproduce on UI https://metaboat.slack.com/archives/C064TCGEAV6/p1707124648877009?thread_ts=1707122813.385479&cid=C064TCGEAV6

crisptrutski:
@qnkhuat You're right that this change doesn't quite fix the issue, and the reason is a bit subtle.

I naively thought that transactions by default would be "serializable", the only common (I think) isolation level that prevents committing transactions with phantom reads. Unfortunately it seem the actual level is "read committed", which is not sufficient. On reflection it makes sense to have this more relaxed default, as serializable ones are quite expensive and also put the burden of handling concurrent modification exceptions on the user code.

I've got the fix working locally by changing the isolation level, but it's a bit more complex than just setting this property of the connection, and I'm going to wrap it up in a nice high level helper for idempotent inserts.

On this topic, [this makes me want to refresh myself on consistency models](https://aphyr.com/posts/313-strong-consistency-models).

crisptrutski:
I'm also going to need to revisit my previous PRs similar to this one to upgrade them 😅

src/metabase/db/util.clj
```diff
@@ -66,3 +69,26 @@
   ;;    [:in :semantic_type #{"type/URL" "type/ImageURL"}]
   ([expr type-keyword]
    [:in expr (type-keyword->descendants type-keyword)]))
+
+(defmacro idempotent-insert!
+  "Upsert a database record where the computation may be expensive etc. Blah blah toucan.
```
crisptrutski, line 74: Haha whoops, need to polish this docstring 😂

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,54 @@
+(ns metabase.db.util-test
+  (:require
+   [clojure.test :refer [deftest testing is]]
+   [metabase.db.util :as mdb.u]
+   [metabase.models.setting :refer [Setting]]
+   [toucan2.core :as t2]))
+
+(set! *warn-on-reflection* true)
+
+(deftest idempotent-insert!-test
```
calherries, line 10: There's a chance `Setting` could be updated beyond `key` and `value` and in that world it would be helpful to understand why setting was chosen and why we're using key and value like this. 

It wouldn't hurt to be explicit about the assumptions we're making here too in a comment with the following info:
- Setting has two columns: `key` and `value`, which are both strings
- `key` has a uniqueness constraint on it, and `value` doesn't

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,57 @@
...
+
+              ;; hit it
+              (dotimes [_ threads]
+                (swap! promises conj (future (thunk))))
+
+              (let [result-keys (mapv deref @promises)
+                    latest-key  (:key (t2/select-one Setting search-col search-value))]
+
+                (testing "Every call returns the same row"
+                  (is (= (repeat threads latest-key) result-keys)))
```
calherries, line 46: this reads better imo
```suggestion
                  (is (apply = latest-key result-keys)))
```

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,57 @@
...
+              other-col    (first (remove #{search-col} columns))]
+          (try
+            ;; ensure there is no database detritus
+            (t2/delete! Setting search-col search-value)
+
+            (let [threads  5
+                  promises (atom [])
+                  thunk    (fn []
+                             (mdb.u/idempotent-insert!
+                               (:key (t2/select-one Setting search-col search-value))
```
calherries, line 31: just to reflect that this needs to return the PK
```suggestion
                               (t2/select-one-pk Setting search-col search-value)
```

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,57 @@
...
+                                   (t2/insert-returning-pk! Setting
+                                                            search-col search-value
+                                                            other-col (str (random-uuid))))))]
+
+              ;; hit it
+              (dotimes [_ threads]
+                (swap! promises conj (future (thunk))))
+
+              (let [result-keys (mapv deref @promises)
+                    latest-key  (:key (t2/select-one Setting search-col search-value))]
```
calherries, line 43:
```suggestion
                    latest-key  (t2/select-one-pk Setting search-col search-value)]
```

calherries (approved):
LGTM

test/metabase/db/util_test.clj
crisptrutski: I didn't feel it necessary to call out that the fields are strings - would it have mattered if they were were numbers?

test/metabase/db/util_test.clj
crisptrutski: Both have their weaknesses - Cursive only shows a two-way diff so makes it hard to introspect these failures. I also like the clear structural separation between the expected value and the returned values. Perhaps a better option would be `(= [latest-key] (distinct result-keys))`, but I'm happy enough as-is.

test/metabase/db/util_test.clj
calherries: Fair enough, I listed all the assumptions the test makes but probably not worth calling out to understand how it works at a high level

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,60 @@
...
+              other-col    (first (remove #{search-col} columns))]
+          (try
+            ;; ensure there is no database detritus
+            (t2/delete! Setting search-col search-value)
+
+            (let [threads  5
+                  promises (atom [])
+                  thunk    (fn []
+                             (mdb.u/idempotent-insert!
+                               (t2/select-one-pk Setting search-col search-value)
```
calherries, line 34: Now that I think of it this test would be stronger and read clearer if you used `t2/select-one` and `t2/insert-returning-instance!`
e.g.
```suggestion
                               (t2/select-one Setting search-col search-value)
```
At the moment the tests could pass even if every thread deleted any existing row and (re)created one.

crisptrutski: I don't understand how you think we'd get a false pass - the atom stores the id that each thread created, so we'd fail the assertion that they all returned the same value. Am I missing something?

test/metabase/db/util_test.clj
crisptrutski: Explaining the rationale was a good call out - let me know if you're happy with how I phrased it.

test/metabase/db/util_test.clj
crisptrutski: I ended up making other revisions to this PR anyway, so switched to using distinct.

test/metabase/db/util_test.clj
calherries: Sorry, could have been clearer: the tests could pass only on the `search-col=key` iteration. 

Why? The PK for Setting is the `key` column. When `search-col=key` the `key` returned by `thunk` will be the same for each thread, the random-uuid of line 24. So regardless of whether a new row was inserted or a previously created row was selected, each thread will return the same key. That's not true if you return the whole row.

test/metabase/db/util_test.clj
calherries: Yes, looks good

test/metabase/db/util_test.clj
crisptrutski: Ah, got you. I think there's some toucan metadata in there that caused issues, but perhaps I can strip that.

crisptrutski:
For some reason this method is not working at all in CI. I think the issue is H2, as I tested with Postgres locally. It definitely was throwing the concurrent modification error and not passing "accidentally". I'm very confused since H2 is meant to take locks for serialization, versus Postgres' MVCC, so if anything it should be even *more* solid. 😞

src/metabase/db/util.clj
```diff
@@ -66,3 +69,35 @@
...
+  [select-expr insert-expr]
+  ;; First attempt the select without a serializable transaction, since those are expensive.
+  `(or ~select-expr
+       (try
+         (t2/with-connection [conn#]
+           (.setTransactionIsolation ^Connection conn# Connection/TRANSACTION_SERIALIZABLE)
+           (t2/with-transaction [~'_conn]
+             ;; We need to try select the row again now that we're in the transaction to track the dependency.
+             (u/or-with some?
+                      ~select-expr
```
qnkhuat, line 92: Do you eval `select-expr` twice on purpose?

src/metabase/db/util.clj
```diff
@@ -66,3 +69,35 @@
   ;;    [:in :semantic_type #{"type/URL" "type/ImageURL"}]
   ([expr type-keyword]
    [:in expr (type-keyword->descendants type-keyword)]))
+
+(defmacro idempotent-insert!
```
qnkhuat, line 73: I'd note that `select-expr` shouldn't produce side-affects as it'll be evaluated multiple times

src/metabase/models/params/field_values.clj
```diff
@@ -131,17 +132,20 @@
 
   ([fv-type field constraints]
    (let [hash-key (hash-key-for-advanced-field-values fv-type (:id field) constraints)
-         fv       (or (t2/select-one FieldValues :field_id (:id field)
```
qnkhuat, line 134: can you add a note on why we use `idempotent-insert!` here?

qnkhuat (approved):
Nice!

crisptrutski:
Oops, this branch had an auto-merge from before it was reworked.

crisptrutski:
Recreated here https://github.com/metabase/metabase/pull/38760

PR #38442, opened 2024-02-05
deserialization should happen inside a transaction

### LINKED ISSUE #20632: Option to rollback changes in failed serialization load
Currently if a load fails, the destination app database will be in a state where it has some of the changes that are in the source database, so in order to keep things in order between environments, the serialization process should rollback changes if there's an error

Rollback changes if there's an error in the serialization load (if user uses --on-error rollback)

Backup the DB before you do the load, if it fails, restore the backup

A real life-saver for the self-hosted ops team

**Additional context**
NA

sansolovyov:
Deserialization can leave a target database in a broken state if an error happens. Wrapping it into transaction will prevent this.

Resolves #20632.

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/v2/load.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/v2/load.clj
@@ -7,7 +7,8 @@
    [metabase-enterprise.serialization.v2.ingest :as serdes.ingest]
    [metabase.models.serialization :as serdes]
    [metabase.util.i18n :refer [trs]]
-   [metabase.util.log :as log]))
+   [metabase.util.log :as log]
+   [toucan2.core :as t2]))
 
 (declare load-one!)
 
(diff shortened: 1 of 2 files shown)


crisptrutski:
I've just got some non functional concerns / structural questions.

The first thing I noticed was a disparity in the level of abstraction where the transaction was started - one at the top of the stack in a handler, and the other a few calls deep into the call stack for command. Why have we not pushed it further down in the former or placed it higher up in the latter?

My preference is to push it as low down the stack as necessary - it's a low level detail and in generally we'll be looking back up the stack from low level database code or exceptions to check if we're in a transaction. 

In the lower level case here, we're putting it around code to (1) set a caching scope, (2) load the yaml file, and then (3) load it. As far as I can see the former two do not interact with the database, and we can our transaction block down into `metabase-enterprise.serialization.v2.load/load-database!`, and this would cover both cases.

Even if we need it slightly higher up the stack I'd rather stick it in a function that takes a filepath and ingestion options, and DRYs up those three steps.

qnkhuat:
agree with Chris, this should be done within serdes somehow

piranha:
I had some doubts about having `with-cache` outside `with-transaction` — but now that I'm thinking about it, it does not make any sense. So yeah, I'll push it down a little bit.

piranha:
@crisptrutski @qnkhuat re-review please :)

crisptrutski (approved):

PR #38443, opened 2024-02-05
Make Bookmarks Normal

### LINKED ISSUE #36273: [Audit V2] bookmarking Metabase analytics has a different icon
### Describe the bug

If you bookmark Metabase analytics, it will use the folder icon in the bookmark section, rather than the analytics icon.

image

### To Reproduce

Bookmark Metabase Analytics

### Expected behavior

Should probably be the same icon as below, or we should disable bookmarking for that collection

### Logs

_No response_

```JSON
current master
```

### Severity

purely cosmetic

### Additional context

_No response_

iethree:
### Description

Instead of having a special, confusing, data format for bookmarks, make them follow something similar to the search payload that will play nicer with other data in the app.

Fixes https://github.com/metabase/metabase/issues/36273
Also fixes bookmarked verified collections icon (no issue for that one)

Screen Shot 2024-02-05 at 9 56 34 AM

### Checklist

### DIFF
--- a/frontend/src/metabase-types/api/bookmark.ts
+++ b/frontend/src/metabase-types/api/bookmark.ts
@@ -1,14 +1,5 @@
-export type BookmarkType = "card" | "collection" | "dashboard";
+import type { SearchResult } from "./search";
 
-export interface Bookmark {
-  authority_level?: string;
-  card_id: string;
-  display?: string;
-  id: string;
-  item_id: number;
-  name: string;
-  type: BookmarkType;
-
-  // For questions and models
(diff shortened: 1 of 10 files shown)


iethree:
The failed tests here revealed a hiccup: by making bookmarks normal, we now have possible id collisions because our entity store always works off of the `id` property - so we can't bookmark both collection 37 and dashboard 37.

I need to dig into whether I can teach the entity system to use `bookmark_id` 🤔

PS. hooray for e2e tests that caught this!

PR #38447, opened 2024-02-05
Convert `cards` dashboard actions to TypeScript

kulyk:
Converts `metabase/dashboard/actions/cards` actions file to TypeScript, no functional changes are expected
**To verify:** CI should be green

### DIFF
--- a/frontend/src/metabase-types/api/dashboard.ts
+++ b/frontend/src/metabase-types/api/dashboard.ts
@@ -54,7 +54,7 @@ export type DashCardId = number;
 export type BaseDashboardCard = {
   id: DashCardId;
   dashboard_id: DashboardId;
-  dashboard_tab_id?: DashboardTabId;
+  dashboard_tab_id?: DashboardTabId | null;
   card_id: CardId | null;
   card: Card | VirtualCard;
   collection_authority_level?: CollectionAuthorityLevel;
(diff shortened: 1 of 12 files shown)


frontend/src/metabase/dashboard/actions/cards.js
kulyk, line 1: Only keeping `addActionToDashboard` in JavaScript for now. It's awkward, but we can't remember if `card` on `ActionDashboardCard` is a model its saved to or a virtual card with `display: "action"`. We can followup on that one later

frontend/src/metabase-types/api/dashboard.ts
```diff
@@ -54,7 +54,7 @@ export type DashCardId = number;
 export type BaseDashboardCard = {
   id: DashCardId;
   dashboard_id: DashboardId;
-  dashboard_tab_id?: DashboardTabId;
+  dashboard_tab_id?: DashboardTabId | null;
```
alxnddr, line 57: Can we narrow it down to `dashboard_tab_id: DashboardTabId | null;`?

alxnddr (approved):
Looks good! I left one suggestion, please have a look before merging

frontend/src/metabase-types/api/dashboard.ts
kulyk: Good point 👍 Fixed in [`2ba9048`](https://github.com/metabase/metabase/pull/38447/commits/2ba90489dc1b3d1617306b18fe12c1d9a6537703)

PR #38448, opened 2024-02-05
Adding nil-safe code in model-hydrate-based-on-upload

markbastian:
Commit 8201e5457fcb16922d2b040e13a98d8e626c6070 introduced failing tests (e.g. `clj -X:dev:ee:ee-dev:drivers:test :only metabase.api.card-test/update-card-with-type-and-dataset-test`).

They all seem to originate from NPEs in `model-hydrate-based-on-upload` on the first condition in the `and` clause in the `based-on-upload` anonymous function. This PR just makes that first condition null-safe with a `some->` threading macro.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -732,7 +732,7 @@
         based-on-upload      (fn [model]
                                (when-let [dataset_query (:dataset_query model)] ; dataset_query is sometimes null in tests
                                  (let [query (lib/->pMBQL dataset_query)]
-                                   (when (and (= (name (:query_type model)) "query")
+                                   (when (and (some-> model :query_type name (= "query"))
                                               (contains? uploadable-table-ids (:table_id model))
                                               (no-joins? query))
                                      (lib/source-table-id query)))))]


crisptrutski:
I think in general we treat a missing `:query_type` as "question", but this would be implicitly treating it the same as "native" I believe. Perhaps we should instead be providing a default value for that keyword look-up?

I'm curious why the `:query_type` is missing, and what this means for the quality of the rest of the model data. Is this just a test data issue or does it reflect something we could see in practice?

adam-james-v (approved):

markbastian:
> I think in general we treat a missing `:query_type` as "question", but this would be implicitly treating it the same as "native" I believe. Perhaps we should instead be providing a default value for that keyword look-up?
> 
> I'm curious why the `:query_type` is missing, and what this means for the quality of the rest of the model data. Is this just a test data issue or does it reflect something we could see in practice?

Just saw this. The PR was set to auto-merge and we got another approval, but I think the best thing to do is have @calherries review this when he gets back and make any tweaks needed in a separate PR. This will get master unbroken for the moment.

crisptrutski:
Sounds good - I was waiting for him to come and clear things up in any case 😄

calherries:
Just seeing this now, thanks for tracking down the fix @markbastian!

Indeed this looks like `query_type` is nullable, so this will need changing as Chris suggests. I'll have a proper look tomorrow

PR #38449, opened 2024-02-05
Database `can-[read?,write?]` use DataPermissions

johnswanson:
Both DB `can-read?`/`can-write?` and the logic for determining whether a
user can upload need to check permissions in a way we haven't needed to
previously.

The logic is different for each.

For *schema-level* permissions (checking whether we can upload to a
schema), we want the *minimum* permission for any table in the schema:

- select the set of group-id/permission pairs for all tables within the
schema

- within each group, choose the MOST RESTRICTIVE permission (so a group
with `:unrestricted` access to one table and `:no-self-service` access
to another has `:no-self-service` access to the containing schema)

- of those permissions, choose the LEAST RESTRICTIVE permission (so if
one group has `:unrestricted` access and the other has
`:no-self-service`, the user ends up with `:unrestricted` access).

For database reads, on the other hand, the user has read permission if
they need it for any permissions they have on tables within that
database. In this case, we need to check the *most permissive*
permission for the user in the database - so if the user has
`:no-self-service` for every table in the database except one with
`:unrestricted` access, they should have read-access to the database.

### DIFF
--- a/src/metabase/api/database.clj
+++ b/src/metabase/api/database.clj
@@ -19,6 +19,7 @@
    [metabase.mbql.util :as mbql.u]
    [metabase.models.card :refer [Card]]
    [metabase.models.collection :as collection :refer [Collection]]
+   [metabase.models.data-permissions :as data-perms]
    [metabase.models.database
     :as database
     :refer [Database protected-password]]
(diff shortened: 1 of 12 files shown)


src/metabase/api/database.clj
```diff
@@ -1078,16 +1081,22 @@
 
 
 ;;; ------------------------------------------ GET /api/database/:id/schemas -----------------------------------------
+(defenterprise current-user-can-read-schema?
```
noahmoss, line 1084: nit
```suggestion

(defenterprise current-user-can-read-schema?
```

noahmoss (approved):

PR #38453, opened 2024-02-05
Make upload tests faster with with-uploads-table macro

calherries:
Uploads tests are slow, especially for redshift. The reason is mostly due to using `mt/with-empty-db` to clean up temporary tables created by the tests. `mt/with-empty-db` relatively slow because it performs a database sync on the empty database.

This PR replaces many uses of `mt/with-empty-db` with a new macro, `with-uploads-table`. The new macro doesn't create a new database and instead simply cleans up after a temporary table in the app and test database.

### DIFF
--- a/test/metabase/upload_test.clj
+++ b/test/metabase/upload_test.clj
@@ -482,24 +482,25 @@
 (defmacro with-uploads-allowed [& body]
   `(do-with-uploads-allowed (fn [] ~@body)))
 
-(defn- do-with-upload-table! [t f]
-  (try (f t)
+(defn- do-with-upload-table! [table thunk]
+  (try (thunk table)
        (finally
-         (driver/drop-table! driver/*driver* 
+         (driver/drop-table! driver/*driver*
                              (mt/id)
-                             (#'upload/table-identifier t))))
+                             (#'upload/table-identifier table)))))
 
-(defmacro with-upload-table!
-  "Execute `body` with a table created by evaluating the `table-creator` expression. The table instance is bound to
-  `table-sym` in `body`. The table is cleaned up from both the test and app DB after the body executes.
+(defmacro ^:private with-upload-table!
+  "Execute `body` with a table created by evaluating the expression `create-table-expr`. `create-table-expr` must evaluate
+  to a toucan Table instance. The instance is bound to `table-sym` in `body`. The table is cleaned up from both the test
+  and app DB after the body executes.
 
     (with-upload-table [table (create-upload-table! ...)]
       ...)"
   {:style/indent 1}
-  [[table-sym table-creator] & body]
+  [[table-binding create-table-expr] & body]
   `(with-uploads-allowed
      (mt/with-model-cleanup [:model/Table]
-       (do-with-upload-table! ~table-creator (fn [~table-sym] ~@body)))))
+       (do-with-upload-table! ~create-table-expr (fn [~table-binding] ~@body)))))
 
 (deftest load-from-csv-table-name-test
   (testing "Can upload two files with the same name"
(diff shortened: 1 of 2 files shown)


test/metabase/upload_test.clj
```diff
@@ -472,21 +472,47 @@
...
+(defmacro with-uploads-allowed [& body]
+  `(do-with-uploads-allowed (fn [] ~@body)))
+
+(defn- do-with-upload-table! [t f]
+  (try (f t)
+       (finally
+         (let [table-ident (#'upload/table-identifier t)]
+           (driver/drop-table! driver/*driver* (mt/id) table-ident)))))
+
+(defmacro with-upload-table!
```
qnkhuat, line 491: I think we can safely make macros private

test/metabase/upload_test.clj
```diff
@@ -472,21 +472,47 @@
...
+  "Set uploads-enabled to true, and uses an admin user, run the thunk"
+  [thunk]
+  (mt/with-temporary-setting-values [uploads-enabled true]
+    (mt/with-current-user (mt/user->id :crowberto)
+      (thunk))))
+
+(defmacro with-uploads-allowed [& body]
+  `(do-with-uploads-allowed (fn [] ~@body)))
+
+(defn- do-with-upload-table! [t f]
```
qnkhuat, line 485:
```suggestion
(defn- do-with-upload-table! [table thunk]
```
nit: short var is hard to read IMO

test/metabase/upload_test.clj
```diff
@@ -472,21 +472,47 @@
...
+           (driver/drop-table! driver/*driver* (mt/id) table-ident)))))
+
+(defmacro with-upload-table!
+  "Execute `body` with a table created by evaluating the `table-creator` expression. The table instance is bound to
+  `table-sym` in `body`. The table is cleaned up from both the test and app DB after the body executes.
+
+    (with-upload-table [table (create-upload-table! ...)]
+      ...)"
+  {:style/indent 1}
+  [[table-sym table-creator] & body]
```
qnkhuat, line 498:
```suggestion
  [[table-binding create-table-fn] & body]
```

- table-binding: I think this is how we name these kind of pseudo binding in macro
- create-table-fn: when I first read `table-creator` I thought it's an user, not a function that create a table

qnkhuat: oh I just realized `table-creator` is not a function but a table, in that case I think we can call it `upload-table` since as far as the macro concerned it's evaluated as a table.

qnkhuat (approved):
have a couple of styling nits, but overall this looks good

crisptrutski (commented):
Like Ngoc I also just think a few superficial things can be improved. Nice idea!

test/metabase/upload_test.clj
calherries: Good point, `table-creator` is a it misleading. I went with `create-table-expr` to make it clear that the table needs to be created inside the expression and not before it, which is the main sharp edge when using this macro.

test/metabase/upload_test.clj
```diff
@@ -472,21 +472,47 @@
...
+  `(do-with-uploads-allowed (fn [] ~@body)))
+
+(defn- do-with-upload-table! [t f]
+  (try (f t)
+       (finally
+         (let [table-ident (#'upload/table-identifier t)]
+           (driver/drop-table! driver/*driver* (mt/id) table-ident)))))
+
+(defmacro with-upload-table!
+  "Execute `body` with a table created by evaluating the `table-creator` expression. The table instance is bound to
```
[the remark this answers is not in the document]
calherries: Good point, this requirement was missing from the docstring. Updated in  ed612a658c32be08a5e7264b3cf0ecafe3a13ce3

```
Execute `body` with a table created by evaluating the expression `create-table-expr`. `create-table-expr` must evaluate
  to a toucan Table instance. The instance is bound to `table-sym` in `body`. The table is cleaned up from both the test
  and app DB after the body executes.
```

test/metabase/upload_test.clj
calherries: Done ed612a658c32be08a5e7264b3cf0ecafe3a13ce3

Issue #38320, opened 2024-01-31
MB_ATTACHMENT_TABLE_ROW_LIMIT does not work for email subscriptions


paoliniluis:
### Describe the bug

The env var doesn't seem to do anything, even the default does not get applied

### To Reproduce

1) set up Metabase
2) build a dashboard with a table and a subscription, send it (via email now in subscriptions)
3) table will come with 10 rows (not the default which should be 20)
4) set MB_ATTACHMENT_TABLE_ROW_LIMIT=50, same, you get 10 rows
![image](https://github.com/metabase/metabase/assets/1711649/d21af18e-840d-4833-b94b-306207bca54b)

### Expected behavior

It should respect the rows

### Logs

NA

### Information about your Metabase installation

```JSON
v48.4
```

### Severity

P2

PR #38456, opened 2024-02-06
Add internal setting for table render row limit (#36996)

adam-james-v:
This is a backport of https://github.com/metabase/metabase/pull/36996

Which has been reported as not working in issue #38320 

It's not actually a bug but a mistake on my part. The milestone was placed on the closed issue for 48.2 and so the feature was listed as part of that release. However, I marked 'no-backport' on the original PR and thus forgot to actually produce a backport!

This PR backports the feature to the next 48 point release and should then close the issue once the release is cut.


markbastian (approved):
Looks good.

PR #38458, opened 2024-02-06
Fix ESlint rules

WiNloSt:
Extracted the work from #37617

This should be backported since it's a setup change, so we won't have conflicts later on.

### Description

This PRs fixed
1. Cases where we could still commit `console.log` inside `e2e` folder
2. ESLint complain that the type syntax is wrong, when it is actually correct.

### How to verify

No static check failures

### Checklist

- [ ] ~Tests have been added/updated to cover changes in this PR~ Just fixing lintings


e2e/.eslintrc
```diff
@@ -11,5 +11,13 @@
     "cypress/globals": true,
     "node": true
   },
-  "plugins": ["cypress"]
+  "plugins": ["cypress"],
+  "overrides": [
+    {
+      "files": ["*.cy.spec.*"],
+      "rules": {
+        "no-console": "error"
```
WiNloSt, line 19: This used to be disabled. I suspect the reason is we have some legit uses of `console.log` inside `e2e` folder, but in the actual tests file, we should never allow that.

frontend/src/metabase/components/tree/TreeNodeList.tsx
```diff
@@ -55,7 +55,6 @@ function BaseTreeNodeList({
             />
             {isExpanded && (
               <BaseTreeNodeList
-                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
WiNloSt, line 58: All these removed comments are removed rules from `@typescript-eslint/eslint-plugin` V6

frontend/src/metabase/dashboard/components/PublicLinkPopover/PublicLinkCopyPanel.styled.tsx
```diff
@@ -15,7 +14,7 @@ export const PublicLinkCopyButton = styled(CopyButton)`
 `;
 
 export const RemoveLinkAnchor = styled(Anchor)<
-  AnchorProps & HTMLAttributes<HTMLAnchorElement>
+  ComponentProps<typeof Anchor<ElementType>> & HTMLAttributes<HTMLAnchorElement>
```
WiNloSt, line 17: This is the reason `@typescript-eslint/eslint-plugin` needs to be upgraded, because it was always complaining about this syntax.

oisincoveney: Nice! this is a much better way to get `component` other than just specifying it in the styled props.

oisincoveney (approved):
This looks great, nice work! Once the tests pass let's 🛳️

PR #38468, opened 2024-02-06
Update "Add/update filter button" in filter widgets

oleggromov:
## Description
While working on the [Required filters](https://github.com/metabase/metabase/issues/36524), I realised that there are many issues with the "Update filter" button (which sometimes was called "Add filter").

1. There was no consistency in calling it one thing or another
2. As we added the `required` paramters which must have a value, a strange case became possible: you could uncheck everything in a dropdown and clicking "Update", which in fact would reset the value to default
3. You could click "Update" without changing anything, so there was no meaningful `disabled` state for the button

🙏 Please don't be too harsh about the code itself because it touches ParameterValueWidget and its descendants, which are a mess and have to be overhauled.

## How it works
For each type of parameter filter widget, the logic should be the same:

- initially, as there's no value you see "Add filter"
- once the value is set, usually in an internal component state, you see "Update filter"
- however, when you choose the same value, "Update filter" becomes disabled
- when there's a default value, and it is chosen/entered, the button becomes "Set to default"
- when a parameter is required and you clear the value, the button becomes "Set to default" too 

See examples below.

### String widget
https://github.com/metabase/metabase/assets/2196347/1653f8c6-2af5-4507-97c0-db99b905733d

### Date widget
https://github.com/metabase/metabase/assets/2196347/0c02875e-7340-4a63-93ee-476b8672ae54

### Required parameter dropdown
https://github.com/metabase/metabase/assets/2196347/33ddbdf3-d67d-40d4-baff-4a65eba32f3c

## Test updates
I went for explicitly setting button label that we expect (as opposed to selecting either of "Add", "Update", "Reset") because it conveys an additional meaning of wether the value in the filter changed before.


e2e/test/scenarios/dashboard-filters/reproductions/24235-exlude-all-date-options.cy.spec.js
```diff
@@ -43,11 +43,15 @@ describe("issue 24235", () => {
...
-
     popover().within(() => {
       cy.findByText("Exclude...").click();
       cy.findByText("Days of the week...").click();
       cy.findByText("Select none...").click();
+      cy.findByText("Add filter").click();
+    });
+
+    filterWidget().click();
+    popover().within(() => {
```
oleggromov, line 54: Not sure how this tests was supposed to work, but clicking "Select none" without actually applying the filter didn't do anything. So I split this into two actions to verify it properly.

e2e/test/scenarios/native-filters/sql-field-filter-string.cy.spec.js
```diff
@@ -60,13 +67,18 @@ describe("scenarios > filters > sql filters > field filter > String", () => {
...
+            })
+          : FieldFilter.addDefaultStringFilter(value, {
+              buttonLabel: "Update filter",
+            });
 
         SQLFilter.runQuery();
 
         cy.get(".Visualization").within(() => {
           cy.findByText(representativeResult);
+          cy.findByText("Toucan").should("not.exist");
```
oleggromov, line 81: This entire test suite was broken or, rather, not checking anything.

Prior to my updates, it:

1. typed a string (for example, "Synergistic Granite Chair") into the input
2. **did not** chose it from the list and clicked the "Add filter" button, which **did not do** anything
3. checked that there's a result in the table — but since it **wasn't even filtered** in the first place, each result was always there

frontend/src/metabase/components/DateAllOptionsWidget/DateAllOptionsWidget.tsx
```diff
@@ -27,39 +26,41 @@ export const DateAllOptionsWidget = ({
   onClose,
   disableOperatorSelection,
   value,
+  defaultValue,
+  initialValue,
+  required = false,
 }: DateAllOptionsWidgetProps) => {
   const [filter, setFilter] = useState(
-    value != null ? dateParameterValueToMBQL(value, noopRef) || [] : [],
+    initialValue != null
```
oleggromov, line 34: We need a separate `initialValue` to be used at first render because actual `value` is unset at this point, and it might break the button label selection logic.

cdeweyx (approved):

mazameli:
Nice

frontend/src/metabase/parameters/components/ParameterWidget/ParameterWidget.unit.spec.tsx
```diff
@@ -60,7 +60,11 @@ describe("ParameterWidget", () => {
     userEvent.type(textInput, "{backspace}".repeat(text.length));
     expect(screen.getByRole("button", { name: "Add filter" })).toBeDisabled();
 
-    userEvent.click(screen.getByRole("button", { name: "Add filter" }));
+    userEvent.click(
+      screen.getByRole("button", { name: "Add filter" }),
+      undefined,
+      { skipPointerEventsCheck: true },
```
uladzimirdev, line 66: could you please explain why it is required?

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -233,6 +233,26 @@ class TagEditorParamInner extends Component<Props> {
     const hasNoDefaultValue = !tag.default;
     const isEmbeddedDisabled = embeddedParameterVisibility === "disabled";
 
+    // We want to remove "default" and "required" so that it
+    // doesn't show up in the default value input update button
+    const defaultParameterInputValue = _.omit(
```
uladzimirdev, line 238: nit: I'd try to simplify this construction

frontend/src/metabase/parameters/components/widgets/StringInputWidget/StringInputWidget.tsx
```diff
@@ -55,25 +55,27 @@ function StringInputWidget({
           value={unsavedArrayValue}
           onChange={setUnsavedArrayValue}
           placeholder={placeholder}
-          options={OPTIONS}
+          options={[]}
```
uladzimirdev, line 58: passing new array can cause unnecessary re-renderings, but it shouldn't be a big problem

uladzimirdev (approved):

frontend/src/metabase/parameters/components/ParameterWidget/ParameterWidget.unit.spec.tsx
oleggromov: Now, as we're using the Mantine button, it implements `pointer-events: none` for the `disabled` button for some reason. This means the testing lib starts complaining that it couldn't click the button that has this property. 

So instead of changing the test, which I think is valid (it's okay to try to click a disabled button), I simply "allowed" it to perform a click.

frontend/src/metabase/parameters/components/widgets/StringInputWidget/StringInputWidget.tsx
oleggromov: That is a good catch actually! Thanks.

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
oleggromov: Agreed. To be fair, it was like that before, I simply extracted it into a variable and did an `omit`, which is the entire reason for all this.

frontend/src/metabase/parameters/components/ParameterValueWidget.jsx
```diff
@@ -35,15 +35,6 @@ import { areParameterValuesIdentical } from "metabase-lib/parameters/utils/param
 import ParameterFieldWidget from "./widgets/ParameterFieldWidget/ParameterFieldWidget";
 import S from "./ParameterValueWidget.css";
 
-const DATE_WIDGETS = {
```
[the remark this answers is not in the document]
oleggromov: Yeah, as I was working with the code, I jumped between them a million times — and I believe it's not really a "constant" but rather an in-place map, which shouldn't even exist in first place.

So I'll keep it as is now if you don't mind and will come back to all this mess later. These components, including the very idea of having 6 date components for different widget types, require a significant overhaul anyway.

Issue #38445, opened 2024-02-05
Add connection impersonation to Redshift


luizarakaki:
We support connection impersonation for Snowflake and Postgres. We can extend support to Redshift too.

It behaves slightly differently from Postgres as it uses `SET SESSION AUTHORIZATION` and passes usernames instead of roles.

[Product doc](https://www.notion.so/metabase/Add-impersonation-to-Redshift-5fd1d93fba3b4a9683241ad3729a9041?pvs=4)
```[tasklist]
### Tasks
- [x] [BE] Tweak driver to use `set session authorization`
- [x] [FE] Change copy: replace "role" with "user"
```

marcoquerque:
In general, what are your thoughts on requiring the redshift user in metabase be a super user to use this?

luizarakaki:
Redshift connection impersonation uses `SET SESSION AUTHORIZATION`.

To run this command, the connection must use superuser

marcoquerque:
And what is our general thoughts on the metabase redshift user being a superuser? are we concerned from a security perspective - can we confidently say every query will use this?

luizarakaki:
Queries from users in groups with impersonated connection, yes. But admins are never impersonated, so at least queries run by admin won't use this.

calherries:
Pretty sure this is closed by https://github.com/metabase/metabase/pull/38530, so I'm closing this to get it into the release notes. Please yell at me if not

PR #38475, opened 2024-02-06
Connection Impersonation for Redshift

noahmoss:
Backend for https://github.com/metabase/metabase/issues/38445

Adds support for connection impersonation for Redshift on the backend. Pretty similar to postgres, just with different syntax for setting the role (or user, in this case) on the connection.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/driver/impersonation_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/driver/impersonation_test.clj
@@ -109,10 +109,10 @@
                                "CREATE TABLE PUBLIC.table_with_access (x INTEGER NOT NULL);"
                                "CREATE TABLE PUBLIC.table_without_access (y INTEGER NOT NULL);"
                                "DROP USER IF EXISTS \"impersonation.user\";"
-                               "CREATE USER \"impersonation.user\";"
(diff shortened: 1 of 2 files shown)


noahmoss:
* `master`
  * **PR #38475** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38475?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38475?utm_source=stack-comment).

enterprise/backend/test/metabase_enterprise/advanced_permissions/driver/impersonation_test.clj
```diff
@@ -97,6 +97,42 @@
                          mt/process-query
                          mt/rows))))))))))
 
+(deftest conn-impersonation-test-redshift
```
noahmoss, line 100: These individual driver tests are in the generic `metabase-enterprise.advanced-permissions.driver.impersonation-test` because they need to be in an EE namespace to use the `with-impersonations` macro, and we don't have driver-specific EE test namespaces already.

If we add a lot more drivers we can rethink the organization but I'm sticking this one here too for the time being.

luizarakaki (approved):
reviewed with Noah

qwef (approved):
lgtm

PR #38476, opened 2024-02-06
[MLv2] Add description and fingerprint to `displayInfo` for columns

bshepherdson:
If present. Also adds these fields to the TS types.

### DIFF
--- a/frontend/src/metabase-lib/types.ts
+++ b/frontend/src/metabase-lib/types.ts
@@ -135,9 +135,10 @@ type TableInlineDisplayInfo = Pick<
 
 export type ColumnDisplayInfo = {
   name: string;
+  description?: string;
   displayName: string;
   longDisplayName: string;
-  semanticType: string;
+  semanticType: string | null;
   effectiveType: string;
 
   isCalculated: boolean;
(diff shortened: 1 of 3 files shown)


bshepherdson:
* `master`
  * **PR #38476** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38476?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38476?utm_source=stack-comment).

src/metabase/lib/field.cljc
```diff
@@ -307,6 +307,12 @@
    ;; because the query processor doesn't produce nested display-names.
    {:display-name (lib.metadata.calculation/display-name query stage-number field-metadata)
     :long-display-name (lib.metadata.calculation/display-name query stage-number field-metadata :long)}
+   ;; Include description and fingerprint if they're present on the column. Only proper fields or columns from a model
+   ;; have these, not aggregations or expressions.
+   (when-let [description (:description field-metadata)]
+     {:description description})
+   (when-let [fingerprint (:fingerprint field-metadata)]
+     {:fingerprint fingerprint})
```
snoe, line 315: It would be nice to have a test for this - will this map convert to js properly?

snoe (approved):

src/metabase/lib/field.cljc
bshepherdson: It does, but it certainly doesn't hurt to test it. Doubly so as a couple of the keys are weird, like `"nil%"`.

bshepherdson: :sheep: Actually I found a bug by adding this test - the `displayInfo` keys are all converted to `camelCase` so while the `nil%` goes through intact, multi-word keys like `:distinct-count` become `"distinctCount"`, which doesn't match the fingerprint type used elsewhere.

I've updated the TS types accordingly.

PTAL, there's a fair chunk of new BE and FE code.

frontend/src/metabase-lib/types.ts
```diff
@@ -135,6 +135,7 @@ type TableInlineDisplayInfo = Pick<
 
 export type ColumnDisplayInfo = {
   name: string;
+  description?: string;
   displayName: string;
   longDisplayName: string;
   semanticType: string;
```
romeovs, line 141: Only tangentially related, but I've found that `semanticType` can be `null`.

snoe (approved):

PR #38481, opened 2024-02-06
[MLv2] Migrate `usesMetric` using new question filter from the `GET /api/card` endpoint

### LINKED ISSUE #37510: [MLv2] [FE] Migrate `usesMetric`
Blocked by https://github.com/metabase/metabase/issues/37688

https://github.com/metabase/metabase/blob/ba9375171cd4b43a61f2eab082e2377348e2881b/frontend/src/metabase-lib/Question.ts#L561

Similar to https://github.com/metabase/metabase/issues/37511

nemanjaglumac:
Prerequisite for this PR was https://github.com/metabase/metabase/issues/37688. It's been merged to `master`.

### What does this PR do?
Migrates `usesMetric` method on the `Question` prototype using new question filter from the `GET /api/card endpoint`

It achieves this by relying on our entities framework, and by directly loading filtered questions using `Questions.loadList()`.
Both the selector `getMetricQuestions` and the Question method `usesMetric ` are not needed anymore so this PR removes them.

### How to test?
- Create at least one metric[^1]
- Create a few questions that use that metric and save them
- Navigate to `/reference/metrics/:metric-id/questions` and make sure that only questions that use this segment are displayed
- Additionally, all tests should still pass
image

Resolves #37510

[^1]: https://www.metabase.com/docs/latest/data-modeling/segments-and-metrics

### DIFF
--- a/frontend/src/metabase/reference/metrics/MetricQuestions.jsx
+++ b/frontend/src/metabase/reference/metrics/MetricQuestions.jsx
@@ -49,6 +49,12 @@ const mapDispatchToProps = {
   ...metadataActions,
 };
 
+const getDescription = question => {
+  const timestamp = moment(question.getCreatedAt()).fromNow();
+  const author = question.getCreator().common_name;
+  return t`Created ${timestamp} by ${author}`;
+};
+
 export const MetricQuestions = ({ style, table, metric, metadata }) => {
   const {
     data = [],
@@ -72,20 +78,14 @@ export const MetricQuestions = ({ style, table, metric, metadata }) => {
             <div className="wrapper wrapper--trim">
               <List>
                 {data.map(question => {
-                  const card = question.card();
-                  return (
-                    question.id() &&
-                    question.displayName() && (
-                      <ListItem
-                        key={question.id()}
-                        name={question.displayName()}
-                        description={t`Created ${moment(
-                          card.created_at,
-                        ).fromNow()} by ${card.creator.common_name}`}
-                        url={Urls.question(card)}
-                        icon={visualizations.get(question.display()).iconName}
-                      />
-                    )
+                  question.id() && question.displayName() && (
+                    <ListItem
+                      key={question.id()}
+                      name={question.displayName()}
+                      description={getDescription(question)}
+                      url={Urls.question(question.card())}
+                      icon={visualizations.get(question.display()).iconName}
+                    />
                   );
                 })}
               </List>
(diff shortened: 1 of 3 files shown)


kamilmielnik:
Both review comments I posted in #38500 apply here as well.

frontend/src/metabase/reference/metrics/MetricQuestions.jsx
```diff
@@ -49,75 +43,71 @@ const mapStateToProps = (state, props) => ({
...
-  render() {
-    const { entities, style, loadingError, loading, table, metric, metadata } =
-      this.props;
+export const MetricQuestions = ({ style, table, metric, metadata }) => {
+  const {
+    data = [],
+    isLoading,
+    error,
+  } = useQuestionListQuery({
+    query: { f: "using_metric", model_id: metric.id },
```
nemanjaglumac, line 64: @ranquild and @kamilmielnik `metric` wasn't marked as required in this file (unlike `segment` in SegmentQuestions). I can add a guard `metric?.id` or I can mark it as required.

There is no way to reach this component by actually clicking in the UI. You first need a metric in order to see questions that use it.

BUT

There is a scenario that the old code didn't handle. What happens if someone manually enters `/reference/metrics/:random-non-existent-id/questions`? The answer (from `master` is this)
image

Given that someone needs to intentionally enter the wrong id, and given that we didn't handle this even before these new changes, do you still want me to handle that case or continue ignoring it?

If "ignore it" is the answer - I'll just mark `metric` as required.
If "handle it" is the answer - which component or which design should I use for the empty, not found scenario?

ranquild: I would ignore this case in this PR

nemanjaglumac: Done https://github.com/metabase/metabase/pull/38481/commits/5e1ee05927b75baa4730321baee3f585e20ca2fa

Thank you

frontend/src/metabase/reference/metrics/MetricQuestions.jsx
```diff
@@ -49,75 +43,73 @@ const mapStateToProps = (state, props) => ({
...
+    enabled: true,
```
kamilmielnik, line 65:
```suggestion
```

kamilmielnik (approved):

frontend/src/metabase/reference/metrics/MetricQuestions.jsx
nemanjaglumac: Thank you
I'll remove it both here and in SegmentQuestions when I rebase and before I merge https://github.com/metabase/metabase/pull/38500

PR #38482, opened 2024-02-06
Don't instrument any namespaces for mu.fn/fn in production

### LINKED ISSUE #38072: Sync and scan pegs the CPU to 100%

calherries:
Could fix #38072

Removes all malli schema instrumentation of `mu.fn/fn` and `mu/defn` across all namespaces to avoid performance regressions. Previously, we were only skipping namespaces matching these patterns:

```
#"^metabase\.lib\..*"
#"^metabase\.query-processor\.middleware\..*"
#"^metabase\.sync.*"
#"^metabase\.upload.*"
```

Follows https://github.com/metabase/metabase/pull/38206 and https://github.com/metabase/metabase/pull/37428

[Slack thread](https://metaboat.slack.com/archives/CKZEMT1MJ/p1704454622640319)


src/metabase/util/malli/fn.clj
```diff
@@ -300,23 +300,12 @@
...
-        m                  (meta namespace)]
-    (cond (:instrument/always m)                  true
-          (:instrument/never m)                   false
-          (matches? namespace lib-and-middleware) false
-          (contains? ad-hoc (ns-name namespace))  false
-          :else                                   true)))
+  (let [m (meta namespace)]
+    (cond (:instrument/always m) true
+          (:instrument/never m)  false
+          :else                  true)))
```
bshepherdson, line 308: Doesn't this turn instrumentation _on_ in all cases? Maybe I'm missing some code that does that at a higher level.

crisptrutski: It indeed looks like the `:else` case is backward here - perhaps got tripped up that `*skip-ns-decision-fn*` just inverts it? Perhaps it's worth adding unit tests for this function.

calherries: Good eyes, this indeed turns it on. Fixed here daca208c781952ae99c4e7c8020997eab7916a1e I'll write a unit test for `*skip-ns-decision-fn*`.

src/metabase/util/malli/fn.clj
```diff
@@ -300,23 +300,12 @@
 ;; ------------------------------ Skipping Namespace Enforcement in prod ------------------------------
 
 (defn instrument-ns?
-  "Used to track namespaces to not enforce malli schemas on with `mu.fn/fn`."
+  "Returns true if mu.fn/fn malli schema validation should happen in production for a given namespace."
   [namespace]
-  (let [lib-and-middleware [#"^metabase\.lib\..*"
```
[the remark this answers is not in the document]
calherries: Do you mean in the PR description? I think a list of all non-matching namespaces will be quite a bit harder to understand than "everything except this list of matching namespaces".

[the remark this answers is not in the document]
calherries: "everything else apart from these patterns" is difficult to sumarize. But I've updated the description, hopefully that makes it more approachable.

src/metabase/util/malli/fn.clj
```diff
@@ -300,33 +300,11 @@
...
-        ;; empty but placeholder for any namespaces we want to never instrument (in prod)
-        ad-hoc             #{}
-        m                  (meta namespace)]
-    (cond (:instrument/always m)                  true
-          (:instrument/never m)                   false
-          (matches? namespace lib-and-middleware) false
-          (contains? ad-hoc (ns-name namespace))  false
-          :else                                   true)))
-
-(def ^:private ^:dynamic *skip-ns-decision-fn*
```
calherries, line 321: I decided to replace the `*skip-ns-decision-fn*`, with `instrument-ns?`. 

Some differences:
- it returns true instead of false and vice versa. I think it is more natural to have a "do-something?" function return true when something should be done than having a "skip-something?" function return true when to not do something
- it doesn't log when instrumentation is skipped.
- it replaces the logic of `*skip-ns-decision-fn*` and the old `instrument-ns?` in one function

test/metabase/util/malli/fn_test.clj
```diff
@@ -234,14 +236,30 @@
     (testing "sanity check-error-location that it works"
       (is (= :yes (works? "valid input"))))))
 
+(deftest instrument-ns-test
```
calherries, line 239: I added this unit test. Previously this logic wasn't tested

src/metabase/util/malli/fn.clj
```diff
@@ -300,33 +300,11 @@
...
-                            #"^metabase\.sync.*"
-                            #"^metabase\.upload.*"]
-        matches?           (core/fn [namespace regexes]
-                             (let [n (-> namespace ns-name str)]
-                               (some #(re-matches % n) regexes)))
-        ;; empty but placeholder for any namespaces we want to never instrument (in prod)
-        ad-hoc             #{}
-        m                  (meta namespace)]
-    (cond (:instrument/always m)                  true
-          (:instrument/never m)                   false
```
calherries, line 316: I removed the `:instrument/never` case. There are no namespaces with this metadata and it's meaningless now, since the default is to not instrument.

camsaul (approved):
This works for me, let's get this in for the big performance bump. I don't anticipate this breaking anything, but even if it does it should be an easy fix and will be worth it because of the performance  G A I N S

Btw I haven't really looked at that instrument-ns stuff too much, so ignore me if I'm missing something, but I think it would be more useful to have the "always instrument" metadata be put on the `mu/defn` form instead. That's what Schema did... it had `^:always-instrument` or something like that, and you could put it on an individual `mu/defn`. I think that extra flexibility is ultimately more useful... I could see us wanting to have one function use Malli for parameter validation without wanting to enable it for the entire namespace.

crisptrutski (approved):

src/metabase/util/malli/fn.clj
crisptrutski: So much better 😍

crisptrutski: The one case I can think of for `binding` over `with-redefs` is that we're moving to running more tests in parallel. We're stubbing out this method in multiple tests so they'd risk flaking. This gives me a cheeky idea for a test helper that takes the intrinsic locks on the vars before redefining them, perhaps something we could use to whittle away at more dynamic-for-tests-only things.

src/metabase/util/malli/fn.clj
```diff
@@ -300,33 +300,11 @@
...
-(def ^:private ^:dynamic *skip-ns-decision-fn*
-  "Returns true to skip the emission of malli schema validation code in mu.fn/fn and mu/defn."
-  (core/fn skip-ns-decision-fn
-    [namespace]
-    (and config/is-prod?
-         (let [instrument? (instrument-ns? namespace)]
-           (when-not instrument?
-             (log/info "skipping instrumentation of var in " (ns-name namespace)))
-           (not instrument?)))))
+  (or (true? (:instrument/always (meta namespace)))
```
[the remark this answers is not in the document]
calherries: I'm not sure what extensions we'll do with this key in the future, but nothing else is restricting `:instrument/always` type to a boolean so I figured I'd explicitly restrict it with a `true?` call.

src/metabase/util/malli/fn.clj
calherries: That's a good idea. We use `with-redefs` in a lot of places so such a test helper would be helpful. 

I found a blog post describing exactly that:
https://medium.com/helpshift-engineering/a-study-in-parallelising-tests-b5253817beae#:~:text=Although%20it%20is%20a%20very,is%20visible%20globally%20across%20threads

At the moment the tests in this PR are not particularly costly so I think it makes sense to implement the thread-safe `with-redefs` and switch tests that use it to run in parallel as a bulk PR.

src/metabase/util/malli/fn.clj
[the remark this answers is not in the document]
crisptrutski: I feel like I'm missing the problem this solves. It seems more likely to surprise and confuse someone if richer metadata (e.g. a reason string) is silently ignored.

src/metabase/util/malli/fn.clj
crisptrutski: Interesting approach, much more sophisticated than what I was thinking but nicer too. Adds a bit more overhead but don't think we'll feel it.

+1 on leaving parallelism for another sweep

PR #38487, opened 2024-02-06
serialization: snowplow events for export/import

### LINKED ISSUE #38359: Send a Snowplow event to track Serialization
**Context**
We need to know if people are using Serialization, how often they use and which options they use

- Create a new structured event to track Serialization imports and exports

[export link](https://www.notion.so/metabase/Serialization-export-9f6ccdb0791348d8b9b0d5aba1e53599)
[import link](https://www.notion.so/metabase/Serialization-import-da09e8e1957f4b60a838b5e3ee55b290?pvs=4)

sansolovyov:
Resolves #38359

It also removes `abort-on-error` option, which never was false: it was either supplied and options map contained `{:abort-on-error true}`, or was not and option map was `{}`, and so default value of `true` in `load-metabase!` took precedence. Agreed with @luizarakaki.

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/api.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/api.clj
@@ -1,12 +1,14 @@
 (ns metabase-enterprise.serialization.api
   (:require
    [clojure.java.io :as io]
+   [clojure.string :as str]
    [compojure.core :refer [POST]]
    [java-time.api :as t]
    [metabase-enterprise.serialization.v2.extract :as extract]
    [metabase-enterprise.serialization.v2.ingest :as v2.ingest]
    [metabase-enterprise.serialization.v2.load :as v2.load]
    [metabase-enterprise.serialization.v2.storage :as storage]
+   [metabase.analytics.snowplow :as snowplow]
    [metabase.api.common :as api]
    [metabase.api.routes.common :refer [+auth]]
    [metabase.logger :as logger]
@@ -20,7 +22,7 @@
    [metabase.util.random :as u.random]
    [ring.core.protocols :as ring.protocols])
   (:import
-   (java.io File ByteArrayOutputStream)))
+   (java.io ByteArrayOutputStream File)))
 
 (set! *warn-on-reflection* true)
 
@@ -59,47 +61,56 @@
 ;;; Logic
 
 (defn- serialize&pack ^File [{:keys [dirname] :as opts}]
-  (let [dirname (or dirname
-                    (format "%s-%s"
-                            (u/slugify (public-settings/site-name))
-                            (u.date/format "YYYY-MM-dd_HH-mm" (t/local-date-time))))
+  (let [dirname  (or dirname
+                     (format "%s-%s"
+                             (u/slugify (public-settings/site-name))
+                             (u.date/format "YYYY-MM-dd_HH-mm" (t/local-date-time))))
         path     (io/file parent-dir dirname)
         dst      (io/file (str (.getPath path) ".tar.gz"))
-        log-file (io/file path "export.log")]
-    (with-open [_logger (logger/for-ns 'metabase-enterprise.serialization log-file)]
-      (try                              ; try/catch inside logging to log errors
-        (serdes/with-cache
-          (-> (extract/extract opts)
-              (storage/store! path)))
-        ;; not removing storage immediately to save some time before response
-        (u.compress/tgz path dst)
-        (catch Exception e
-          (log/error e "Error during serialization"))))
-    {:archive  (when (.exists dst)
-                 dst)
-     :log-file (when (.exists log-file)
-                 log-file)
-     :callback (fn []
-                 (when (.exists path)
-                   (run! io/delete-file (reverse (file-seq path))))
-                 (when (.exists dst)
-                   (io/delete-file dst)))}))
+        log-file (io/file path "export.log")
+        err      (atom nil)
+        report   (with-open [_logger (logger/for-ns 'metabase-enterprise.serialization log-file)]
+                   (try                 ; try/catch inside logging to log errors
+                     (let [report (serdes/with-cache
+                                    (-> (extract/extract opts)
+                                        (storage/store! path)))]
+                       ;; not removing dumped yamls immediately to save some time before response
+                       (u.compress/tgz path dst)
+                       report)
+                     (catch Exception e
+                       (reset! err e)
+                       (log/error e "Error during serialization"))))]
+    {:archive       (when (.exists dst)
+                      dst)
+     :log-file      (when (.exists log-file)
+                      log-file)
+     :report        report
+     :error-message (some-> @err str)
+     :callback      (fn []
+                      (when (.exists path)
+                        (run! io/delete-file (reverse (file-seq path))))
+                      (when (.exists dst)
+                        (io/delete-file dst)))}))
 
 (defn- unpack&import [^File file & [size]]
   (let [dst      (io/file parent-dir (u.random/random-name))
-        log-file (io/file dst "import.log")]
-    (with-open [_logger (logger/for-ns 'metabase-enterprise.serialization log-file)]
-      (try                              ; try/catch inside logging to log errors
-        (log/infof "Serdes import, size %s" size)
-        (let [path (u.compress/untgz file dst)]
-          (serdes/with-cache
-            (-> (v2.ingest/ingest-yaml (.getPath (io/file dst path)))
-                (v2.load/load-metabase! {:abort-on-error true}))))
-        (catch Exception e
-          (log/error e "Error during serialization"))))
-    {:log-file log-file
-     :callback #(when (.exists dst)
-                  (run! io/delete-file (reverse (file-seq dst))))}))
+        log-file (io/file dst "import.log")
+        err      (atom nil)
+        report   (with-open [_logger (logger/for-ns 'metabase-enterprise.serialization log-file)]
+                   (try                 ; try/catch inside logging to log errors
+                     (log/infof "Serdes import, size %s" size)
+                     (let [path (u.compress/untgz file dst)]
+                       (serdes/with-cache
+                         (-> (v2.ingest/ingest-yaml (.getPath (io/file dst path)))
+                             (v2.load/load-metabase!))))
+                     (catch Exception e
+                       (reset! err (str e))
+                       (log/error e "Error during serialization"))))]
+    {:log-file      log-file
+     :error-message @err
+     :report        report
+     :callback      #(when (.exists dst)
+                       (run! io/delete-file (reverse (file-seq dst))))}))
 
 ;;; HTTP API
 
@@ -129,7 +140,8 @@
    field_values     [:maybe ms/BooleanValue]
    database_secrets [:maybe ms/BooleanValue]}
   (api/check-superuser)
-  (let [opts               {:targets                  (mapv #(vector "Collection" %)
+  (let [start              (System/currentTimeMillis)
+        opts               {:targets                  (mapv #(vector "Collection" %)
                                                             collection)
                             :no-collections           (and (empty? collection)
                                                            (not all_collections))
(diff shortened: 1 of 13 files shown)


enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -129,7 +140,8 @@
    field_values     [:maybe ms/BooleanValue]
    database_secrets [:maybe ms/BooleanValue]}
   (api/check-superuser)
-  (let [opts               {:targets                  (mapv #(vector "Collection" %)
+  (let [start              (System/currentTimeMillis)
```
crisptrutski, line 143: nitpick: This method uses the system clock and is not guaranteed monotonic, it's preferable to use `System/nanoTime` whenever measuring elapsed time, as it avoids the overhead of a system call and precisely measures elapsed time since the process started.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -140,7 +152,22 @@
...
         {:keys [archive
                 log-file
+                report
+                error-message
                 callback]} (serialize&pack opts)]
+    (snowplow/track-event! ::snowplow/serialization-export api/*current-user-id*
+                           {:source          "api"
+                            :duration        (long (/ (- (System/currentTimeMillis) start) 1000))
+                            :count           (count (:seen report))
+                            :collection      (str/join "," (map str collection))
```
crisptrutski, line 162: Nitpick: you do not need to map over `str`, it is implicit in the `join`.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -160,8 +187,22 @@
   [:as {raw-params :params}]
   (api/check-superuser)
   (try
-    (let [{:keys [log-file callback]} (unpack&import (get-in raw-params ["file" :tempfile])
-                                                     (get-in raw-params ["file" :size]))]
+    (let [start              (System/currentTimeMillis)
```
crisptrutski, line 190: Ditto rather use nanoTime.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -160,8 +187,22 @@
...
+          {:keys [log-file
+                  error-message
+                  report
+                  callback]} (unpack&import (get-in raw-params ["file" :tempfile])
+                                            (get-in raw-params ["file" :size]))
+          imported           (set (map (comp :model last) (:seen report)))]
+      (snowplow/track-event! ::snowplow/serialization-import api/*current-user-id*
+                             {:source        "api"
+                              :duration      (long (/ (- (System/currentTimeMillis) start) 1000))
```
crisptrutski, line 199: I find it helpful to include the units in fields like this, e.g. `duration-seconds`

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -160,8 +187,22 @@
...
   (try
-    (let [{:keys [log-file callback]} (unpack&import (get-in raw-params ["file" :tempfile])
-                                                     (get-in raw-params ["file" :size]))]
+    (let [start              (System/currentTimeMillis)
+          {:keys [log-file
+                  error-message
+                  report
+                  callback]} (unpack&import (get-in raw-params ["file" :tempfile])
+                                            (get-in raw-params ["file" :size]))
+          imported           (set (map (comp :model last) (:seen report)))]
```
crisptrutski, line 196: Since you print out this map, it would be nice to use a `sorted-set` to make the string more stable and easy for humans to scan.

I also think `into` with a transducer is more idiomatic for this kind of transformation.

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
```diff
@@ -105,9 +106,25 @@
...
+                   (catch Exception e
+                     (reset! err e)))
+        imported (set (map (comp :model last) (:seen report)))]
+    (snowplow/track-event! ::snowplow/serialization-import nil
+                           {:source        "cli"
+                            :duration      (long (/ (- (System/currentTimeMillis) start) 1000))
+                            :models        (str/join "," imported)
+                            :count         (if (contains? imported "Setting")
+                                             (inc (count (remove #(= "Setting" (:model (first %))) (:seen report))))
+                                             (count (:seen report)))
```
crisptrutski, line 123: nitpick: This calculation is fairly non-trivial and repeated, I would pull out a helper method, but I know I'm more aggressive than others 😀

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
```diff
@@ -105,9 +106,25 @@
...
+                   (v2-load-internal! path opts :token-check? true)
+                   (catch Exception e
+                     (reset! err e)))
+        imported (set (map (comp :model last) (:seen report)))]
+    (snowplow/track-event! ::snowplow/serialization-import nil
+                           {:source        "cli"
+                            :duration      (long (/ (- (System/currentTimeMillis) start) 1000))
+                            :models        (str/join "," imported)
+                            :count         (if (contains? imported "Setting")
+                                             (inc (count (remove #(= "Setting" (:model (first %))) (:seen report))))
```
crisptrutski, line 122: I'm curious about this "remove then count one extra" logic - is it possible that 0 or more than 1 are removed? A small comment could be helpful.

crisptrutski (commented):
I can't comment on whether these are the right events and payloads to publish, but overall looks good. Left some minor comments.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: eh, I don't think precision is necessary here - serialization is usually in tens or hundreds of seconds

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: ohh indeed! nice!

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
piranha: 😁 those two namespaces are not really connected and I did not want to press DRY too much here, honestly, analytics is maintained better when it's in-place in my opinion...

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
piranha: Oh, ok! The thing is that every Setting is counted as a separate instance and it felt like it could skew understanding of a size of serialization batch a bit. So this is me accounting for it instead of making everyone analyzing account for settings in their head.

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
crisptrutski: Lack of a good place to put the helper is fair enough. It nice thing about the helper is that it would have forced a good name for the operation, which still slightly confuses me 😆

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
crisptrutski: I'm still a bit confused though - in practice how is this count different from what you'd get if you just did a direct count?

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: done

enterprise/backend/src/metabase_enterprise/serialization/api.clj
```diff
@@ -59,47 +61,56 @@
...
+        err      (atom nil)
+        report   (with-open [_logger (logger/for-ns 'metabase-enterprise.serialization log-file)]
+                   (try                 ; try/catch inside logging to log errors
+                     (log/infof "Serdes import, size %s" size)
+                     (let [path (u.compress/untgz file dst)]
+                       (serdes/with-cache
+                         (-> (v2.ingest/ingest-yaml (.getPath (io/file dst path)))
+                             (v2.load/load-metabase!))))
+                     (catch Exception e
+                       (reset! err (str e))
```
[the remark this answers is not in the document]
piranha: done

enterprise/backend/src/metabase_enterprise/serialization/api.clj
piranha: done

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
piranha: Settings are about 70-80+ keys and they are cheap, so for me it skewed the picture, it went from 5 entities to 90 actually. :) So to keep this number at least somewhat relevant I decided not to count each individual one.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
crisptrutski: The issue is not precision, it is accuracy. 

I'm not blocking on this, but "in my book" it's best to strictly use nanoTime for timing things - it saves you thinking about whether you need to worry about negative numbers or making an inner loop slower.

crisptrutski: I found a good quote on StackOverflow to summarize:

> The purpose of nanoTime is to measure elapsed time, and the purpose of currentTimeMillis is to measure wall-clock time. You can't use the one for the other's purpose.

enterprise/backend/src/metabase_enterprise/serialization/api.clj
crisptrutski: I don't see the changes 😕 

Not blocking, but maybe commit / push if you have it locally.

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
```diff
@@ -105,9 +106,26 @@
...
-            [:abort-on-error {:optional true} [:maybe :boolean]]
             [:backfill? {:optional true} [:maybe :boolean]]]]
-  (v2-load-internal! path opts :token-check? true))
+  (let [start    (System/currentTimeMillis)
+        err      (atom nil)
+        report   (try
+                   (v2-load-internal! path opts :token-check? true)
+                   (catch Exception e
+                     (reset! err e)))
+        imported (into (sorted-set) (map (comp :model last)) (:seen report))]
```
crisptrutski, line 116: Ah! This is where you made the update we discussed elsewhere. It would be nice if updated the other spot too.

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
```diff
@@ -215,13 +233,34 @@
...
+                 (serdes/with-cache
+                   (-> (cond-> opts
+                         (seq collection-ids)
+                         (assoc :targets (v2.extract/make-targets-of-type "Collection" collection-ids)))
+                       v2.extract/extract
+                       (v2.storage/store! path)))
+                 (catch Exception e
+                   (reset! err e)))]
+    (snowplow/track-event! ::snowplow/serialization-export nil
+                           {:source          "api"
```
crisptrutski, line 248: Aside: it would be nice if `track-event!` took care of turning kebab-case into snake_case. Underscores hurt my eyes in Clojure 😄

crisptrutski:
I'm happy, but would feel better if someone more versed and opinionated with our analytics weighed in on the event format. Hopefully schema evolution is easy enough for the event system?

crisptrutski (approved):

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
luizarakaki, line 248: Shouldn't this be `cli`?

luizarakaki (changes requested):

luizarakaki (approved):
Approved, but confirm that it is not `cli` there

enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
piranha: woah, thanks, and it's in tests too...

PR #38490, opened 2024-02-06
Allow appends from collection list

### LINKED ISSUE #38435: FE: add UI for uploading from collections list
[figma](https://www.figma.com/file/tLUjTffE2vTCRbQeb71HJQ/Allow-appends-on-existing-CSV-upload-models?type=design&node-id=1005-58551&mode=design&t=WqF2DT117ImHZ3VV-0)

Screen Shot 2024-02-05 at 1 49 12 PM

Screen Shot 2024-02-05 at 1 49 21 PM

Screen Shot 2024-02-05 at 1 49 48 PM

```[tasklist]
### Tasks
- [x] Create Model Selection Modal
- [x] Handle Errors
- [x] Handle Upload to new model
```

iethree:
resolves https://github.com/metabase/metabase/issues/38435

## Description

- Allows users to append to models from collections
- If a collection has appendable models, you'll get a nice little modal asking what you want to do with a new upload
- otherwise, we just upload it into a new model

Screen Shot 2024-02-08 at 4 24 55 PM

### Checklist

### DIFF
--- a/frontend/src/metabase/collections/components/ModelUploadModal.tsx
+++ b/frontend/src/metabase/collections/components/ModelUploadModal.tsx
@@ -13,7 +13,10 @@ import {
 import { useSearchListQuery } from "metabase/common/hooks";
 import type { CollectionId, TableId, CardId } from "metabase-types/api";
 
-type UploadMode = "append" | "create";
+enum UploadMode {
+  append = "append",
+  create = "create",
+}
 
 type CollectionOrTableIdProps =
   | { collectionId: CollectionId; tableId?: never }
@@ -30,7 +33,7 @@ export function ModelUploadModal({
   onUpload: ({ collectionId, tableId }: CollectionOrTableIdProps) => void;
   collectionId: CollectionId;
 }) {
-  const [uploadMode, setUploadMode] = useState<UploadMode>("create");
+  const [uploadMode, setUploadMode] = useState<UploadMode>(UploadMode.create);
   const [tableId, setTableId] = useState<TableId | null>(null);
   const models = useSearchListQuery({
     query: {
@@ -71,7 +74,7 @@ export function ModelUploadModal({
     return null;
   }
 
-  const isFormValid = uploadMode === "create" || !!tableId;
+  const isFormValid = uploadMode === UploadMode.create || !!tableId;
 
   return (
     <Modal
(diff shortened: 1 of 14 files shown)


frontend/src/metabase/collections/components/ModelUploadModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Icon,
+} from "metabase/ui";
+import { useSearchListQuery } from "metabase/common/hooks";
+import type { CollectionId, TableId, CardId } from "metabase-types/api";
+
+type UploadMode = "append" | "create";
+
+type CollectionOrTableIdProps =
+  | { collectionId: CollectionId; tableId?: never }
+  | { collectionId?: never; tableId: TableId; modelId?: CardId };
```
sloansparger, line 20: # 🧙‍♂️

frontend/src/metabase/collections/components/ModelUploadModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Stack,
+  Select,
+  Radio,
+  Text,
+  Icon,
+} from "metabase/ui";
+import { useSearchListQuery } from "metabase/common/hooks";
+import type { CollectionId, TableId, CardId } from "metabase-types/api";
+
+type UploadMode = "append" | "create";
```
sloansparger, line 16: !nit (<- meaning don't feel like you have to make this change)

```suggestion
enum UploadMode {
  append = "append",
  create = "create",
}
```

Then you can reference the enum below to avoid having magic strings below.

frontend/src/metabase/collections/components/ModelUploadModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+          onChange={(val: UploadMode) => setUploadMode(val)}
+        >
+          <Radio label={t`Create a new model`} value="create" />
+          <Radio mt="md" label={t`Append to a model`} value="append" />
+        </Radio.Group>
+        {uploadMode === "append" && (
+          <Select
+            icon={<Icon name="model" />}
+            placeholder="Select a model"
+            value={String(tableId) ?? ""}
```
sloansparger, line 96: I saw when poking around on Raffi's browse models stuff that tableId comes back null for this search list query. _If this is happening here_ `String(tableId)` would return `"undefined"`.

frontend/src/metabase/collections/containers/CollectionContent.jsx
```diff
@@ -122,12 +146,9 @@ function CollectionContent({
...
   }, [bookmarks, collectionId]);
 
-  const onDrop = useCallback(
-    acceptedFiles => {
-      uploadFile({ file: acceptedFiles[0], collectionId });
-    },
-    [collectionId, uploadFile],
-  );
+  const onDrop = acceptedFiles => {
+    saveFile(acceptedFiles[0]);
```
sloansparger, line 150: is worth having a check if `acceptedFiles[0]` is not defined?

iethree: I'll double check, but I don't think the library will fire this handler unless there's a valid file

frontend/src/metabase/collections/components/ModelUploadModal.tsx
iethree: well TIL https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime

cdeweyx (approved):

npfitz (approved):
I think this all looks good. Do we want any e2e coverage on the feature?

iethree:
> I think this all looks good. Do we want any e2e coverage on the feature?

good call! I'll submit a follow-up PR for that

PR #38496, opened 2024-02-06
Lock :display when using 'zoom' in Dashboard

### LINKED ISSUE #38307: Zoom in from dashboard change visualization
### Describe the bug

When a question use the bar chart visualization, if you click on "Zoom in" directly from the question, the zoom is displayed as a bar chart.
But, when this question is on a dashboard, then "Zoom in" clicked from the dashbord displays the question as a line chart.

### To Reproduce

Create a question that sums a value by "date > day"
Set visualisation to bar chart
Save the question
Add the question to a dashbord
From the dashbord, on the question click on Zoom In to see the value by hour
--> the question is displayed as a line chart

### Expected behavior

The question should use its initial visualization after zoom in

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "fr",
    "platform": "Win32",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/115.0",
    "vendor": ""
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.version": "11.0.21",
    "os.name": "Linux",
    "os.version": "5.14.0-362.8.1.el9_3.x86_64",
    "user.language": "en",
    "user.timezone": "GMT"
  },
  "metabase-info": {
    "databases": [
      "postgres"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "15.5 (Debian 15.5-1.pgdg120+1)"
      },
      "jdbc-driver": {
        "version": "42.6.0"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2024-01-11",
      "tag": "v0.48.3",
      "hash": "80d8323"
    },
    "settings": {
      "report-timezone": "Europe/Berlin"
    }
  }
}
```

### Severity

Low

### Additional context

_No response_

adam-james-v:
Fixes: #38307

Changing the drill query to lock the Question's display solves the issue where the display from the card is not used when doing a zoom drill through action from a Dashboard. 

When doing a zoom from a Question, the display was already locked. So now it's just also locked in the dashboard case.

https://github.com/metabase/metabase/assets/21064735/56960ec6-9bc1-4072-9e5c-707d98f0653d


kulyk:
Looks good to me, I think this should ideally be approved by @metabase/querying-components though
Drills are very metabase-lib driven, and its mostly QC's area of knowledge

e2e/test/scenarios/dashboard-cards/dashboard-drill.cy.spec.js
```diff
@@ -645,6 +645,41 @@ describe("scenarios > dashboard > dashboard drill", () => {
     });
   });
 
+  it("should keep card's display when doing zoom drill-through from dashboard (metabase#38307)", () => {
+    cy.log("Create a question");
```
kulyk, line 649: nit: i think the test is readable without the logs too 👍

e2e/test/scenarios/dashboard-cards/dashboard-drill.cy.spec.js
```diff
@@ -645,6 +645,41 @@ describe("scenarios > dashboard > dashboard drill", () => {
...
+        breakout: [["field", REVIEWS.CREATED_AT, { "temporal-unit": "month" }]],
+      },
+      display: "bar",
+    }).then(({ body: { id: QUESTION_ID } }) => {
+      cy.createDashboard().then(({ body: { id: DASHBOARD_ID } }) => {
+        cy.log("Add question to the dashboard");
+        addOrUpdateDashboardCard({
+          card_id: QUESTION_ID,
+          dashboard_id: DASHBOARD_ID,
+        });
```
kulyk, line 665: We can use [`cy.createQuestionAndDashboard` command](https://github.com/metabase/metabase/blob/master/e2e/support/commands/api/composite/createQuestionAndDashboard.js) to do all that

kulyk (commented):
LGTM, left two ideas on how to improve the tests

kulyk (approved):
LGTM, left two ideas on how to improve the tests

e2e/test/scenarios/dashboard-cards/dashboard-drill.cy.spec.js
adam-james-v: Ah, thanks for pointing that out :)

e2e/test/scenarios/dashboard-cards/dashboard-drill.cy.spec.js
adam-james-v: Good point. Got rid of logs those logs.

frontend/src/metabase/querying/utils/drills/query-drill.ts
```diff
@@ -18,6 +18,12 @@ export function queryDrill(
     clicked?.dimensions,
   );
 
+  const isDashboard = clicked?.extraData?.dashboard != null;
+
+  if (isDashboard) {
+    question = question.lockDisplay();
+  }
```
kamilmielnik, line 25: Consider this:
1. Create a question [like this](https://github.com/metabase/metabase/assets/6830683/5ff06e5c-3f7c-4c91-a7b0-193fa7a6a832), visualize it as pie chart, save it and add it to a dashboard
2. Now in the dashboard, click on data in the pie chart, select "Break out by..." -> "Location" -> "State"

In `master`, this will result in a nice [US map](https://github.com/metabase/metabase/assets/6830683/c3829281-0ac2-41be-a67c-bc03d10c84db).
With this change, we are presented with a [pie chart](https://github.com/metabase/metabase/assets/6830683/1444420a-526c-4f8c-ab8c-f50b6ae1ac92) - definitely 

From this we can conclude that it's not desired to simply always lock the display.

----

The condition when to lock the display will have to be more sophisticated.

kamilmielnik (dismissed):

frontend/src/metabase/querying/utils/drills/query-drill.ts
kamilmielnik: > When doing a zoom from a Question, the display was already locked.

This may be incorrect, as I would expect the above scenario to yield a US map both in dashboard and a question view.

adam-james-v: Ah this is a great point thanks. 
I will reconsider my approach and have a closer look at my assumptions about locking display in questions too.

adam-james-v: @kamilmielnik I tried your example on Stats but was unable to get the map. Any time I did the zoom in it still brought me to another pie chart. Could you perhaps show a quick screen recording? Or some more screenshots showing exactly how you achieved that?

kamilmielnik: @adam-james-v this is from yesterday's `master`:

https://github.com/metabase/metabase/assets/6830683/2c227064-95f6-48bd-b6c1-8f896ac98e02

kamilmielnik: Nice! I like the simplicity of the solution.

I could not come up with any counter-examples to this scenario but it does not mean they're not there.
Maybe it'd be a good idea to get product team's approval before we merge this change.

I have withdrawn my rejection.
It looks like the test this PR adds is failing in CI.

PR #38500, opened 2024-02-07
[MLv2] Migrate `usesSegment` using new question filter from the `GET /api/card endpoint`

### LINKED ISSUE #37511: [MLv2] [FE] Migrate `usesSegment`
Blocked by https://github.com/metabase/metabase/issues/37688

https://github.com/metabase/metabase/blob/ba9375171cd4b43a61f2eab082e2377348e2881b/frontend/src/metabase-lib/Question.ts#L571

This is used in data reference to display the list of questions that use this metric:
Screenshot 2024-01-15 at 13 05 09

It loads all questions available to this user via `GET /api/card` and then filters them with `usesSegment` method. This is not great and the correct way to do it would be to add `metric-id` parameter to `GET /api/card` that would return cards (questions) that are available to the current user and use this metric.

nemanjaglumac:
Prerequisite for this PR was https://github.com/metabase/metabase/issues/37688. It's been merged to `master`.

### What does this PR do?
Migrates `usesSegment` method on the `Question` prototype using new question filter from the `GET /api/card endpoint`

It achieves this by relying on our entities framework, and by directly loading filtered questions using `Questions.loadList()`.
Both the selector `getSegmentQuestions` and the Question method `usesSegment ` are not needed anymore so this PR removes them.

### How to test?
- Create at least one segment[^1]
- Create a few questions that use that segment and save them
- Navigate to `/reference/segments/:segment-id/questions` and make sure that only questions that use this segment are displayed
- Additionally, all tests should still pass
image

Resolves #37511 

[^1]: https://www.metabase.com/docs/latest/data-modeling/segments-and-metrics

### DIFF
--- a/frontend/src/metabase/reference/segments/SegmentQuestions.jsx
+++ b/frontend/src/metabase/reference/segments/SegmentQuestions.jsx
@@ -1,7 +1,6 @@
 /* eslint "react/prop-types": "warn" */
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
-import _ from "underscore";
 // eslint-disable-next-line no-restricted-imports -- deprecated usage
 import moment from "moment-timezone";
 import { t } from "ttag";
--- a/frontend/src/metabase/reference/segments/SegmentQuestions.jsx
+++ b/frontend/src/metabase/reference/segments/SegmentQuestions.jsx
@@ -1,5 +1,4 @@
 /* eslint "react/prop-types": "warn" */
-import { Component } from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import _ from "underscore";
@@ -9,6 +8,7 @@ import { t } from "ttag";
 import visualizations from "metabase/visualizations";
 import * as Urls from "metabase/lib/urls";
 
+import { useQuestionListQuery } from "metabase/common/hooks";
 import S from "metabase/components/List/List.css";
 
 import List from "metabase/components/List";
@@ -16,20 +16,13 @@ import ListItem from "metabase/components/ListItem";
 import AdminAwareEmptyState from "metabase/components/AdminAwareEmptyState";
 import LoadingAndErrorWrapper from "metabase/components/LoadingAndErrorWrapper";
 
-import Questions from "metabase/entities/questions";
-
 import * as metadataActions from "metabase/redux/metadata";
 import { getMetadata } from "metabase/selectors/metadata";
 import ReferenceHeader from "../components/ReferenceHeader";
 
 import { getQuestionUrl } from "../utils";
 
-import {
-  getError,
-  getLoading,
-  getTableBySegment,
-  getSegment,
-} from "../selectors";
+import { getTableBySegment, getSegment } from "../selectors";
 
 const emptyStateData = (table, segment, metadata) => {
   return {
@@ -47,8 +40,6 @@ const emptyStateData = (table, segment, metadata) => {
 const mapStateToProps = (state, props) => ({
   segment: getSegment(state, props),
   table: getTableBySegment(state, props),
-  loading: getLoading(state, props),
-  loadingError: getError(state, props),
   metadata: getMetadata(state),
 });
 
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/reference/segments/SegmentQuestions.jsx
```diff
@@ -118,4 +125,14 @@ class SegmentQuestions extends Component {
   }
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(SegmentQuestions);
+export default _.compose(
+  Questions.loadList({
```
kamilmielnik, line 129: :x:

`entityListLoader` is deprecated, so I'd say `entity.loadList` is [deprecated as well](https://github.com/metabase/metabase/blob/a989649ffd3310b14b7fcd694f25b9e3fb02c14a/frontend/src/metabase/entities/containers/index.js#L27-L28).

We should use `useQuestionListQuery` here and update `CardFilterOption` type.

Bonus points for marking `entity.loadList` as deprecated.

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
```diff
@@ -83,22 +90,22 @@ class SegmentQuestions extends Component {
...
+            questions.length > 0 ? (
               <div className="wrapper wrapper--trim">
                 <List>
-                  {Object.values(entities).map(
-                    entity =>
-                      entity &&
-                      entity.id &&
-                      entity.name && (
+                  {questions.map(
+                    ({ _card: question }) =>
```
kamilmielnik, line 97: Card is not exactly a Question, the rename is not entirely correct.

kamilmielnik (changes requested):

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
nemanjaglumac: Two points:
- `useQuestionListQuery` most likely has a bug and doesn't work with filter `f`
- even if it did work, you can't use a hook with a class component, and converting this to the functional component is useless since this whole UI will soon be removed

I spoke to @ranquild yesterday and he confirmed to not bother with the conversion to the functional component.

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
nemanjaglumac: I agree.
But the file is called SegmentQuestions. If I start renaming, I'll have to go all the way.

ranquild: Getting access to the underlying _card seems not great. Could you use Question methods instead?

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
ranquild: If converting to a functional component is easy - let’s do it, otherwise that’s not that important here imho

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
nemanjaglumac: It was impossible only in two places.
1. `card.created_at` and
2. `card.creator.common_name`

I've used question methods for everything else.
https://github.com/metabase/metabase/pull/38500/commits/e79a1332fdeb37d939e39473d02dd61f451353ff

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
nemanjaglumac: Done!
And the `useQuestionListQuery` works! 🚀 

Thanks for the encouragement to go down this route.

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
```diff
@@ -46,76 +39,73 @@ const emptyStateData = (table, segment, metadata) => {
...
-            )
-          }
-        </LoadingAndErrorWrapper>
-      </div>
-    );
-  }
-}
+                  return (
+                    question.id() &&
+                    question.displayName() && (
```
nemanjaglumac, line 76: I removed the superfluous check for `question &&` because [we wouldn't even reach this part if the array was empty](https://github.com/metabase/metabase/pull/38500/files#diff-d11769a4e0e90d45028548d93dfa62c4103fc55604634235a6e20307140aa312R68).

ranquild (approved):
Could you expose created_at and creator as Question methods to avoid using card() directly?

frontend/src/metabase/reference/segments/SegmentQuestions.jsx
```diff
@@ -46,76 +37,68 @@ const emptyStateData = (table, segment, metadata) => {
...
-    metadata: PropTypes.object.isRequired,
-  };
+export const SegmentQuestions = ({ style, table, segment, metadata }) => {
+  const {
+    data = [],
+    isLoading,
+    error,
+  } = useQuestionListQuery({
+    query: { f: "using_segment", model_id: segment.id },
+    enabled: true,
```
kamilmielnik, line 54: `enabled` is `true` by default. I double-checked and it works if we skip it.

```suggestion
```

kamilmielnik (approved):

PR #38507, opened 2024-02-07
Various readability improvements around based_on_upload

crisptrutski:
### Description

A bunch of cosmetic feedback that missed the boat for https://github.com/metabase/metabase/pull/38322

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -709,13 +709,10 @@
 (defn uploadable-table-ids
   "Returns the subset of table ids where the user can upload to the table."
   [table-ids]
-  (if (empty? table-ids)
-    #{}
-    (let [tables (t2/hydrate (t2/select :model/Table :id [:in table-ids]) :db)]
-      (set (keep (fn [t]
-                   (when (can-upload-to-table? (:db t) t)
-                     (:id t)))
-                 tables)))))
+  (set (when (seq table-ids)
+         (->> (t2/hydrate (t2/select :model/Table :id [:in table-ids]) :db)
+              (filter #(can-upload-to-table? (:db %) %))
+              (map :id)))))
 
 (defn- no-joins?
   "Returns true if `query` has no joins in it, otherwise false."
@@ -742,16 +739,15 @@
                                   (remove #(false? (:is_upload %)))
                                   (keep :table_id)
                                   set)
-        uploadable-table-ids (set (uploadable-table-ids table-ids))
-        based-on-upload      (fn [model]
-                               (when-let [dataset_query (:dataset_query model)] ; dataset_query is sometimes null in tests
-                                 (let [query (lib/->pMBQL dataset_query)]
-                                   (when (and (some-> model :query_type name (= "query"))
-                                              (contains? uploadable-table-ids (:table_id model))
-                                              (no-joins? query))
-                                     (lib/source-table-id query)))))]
-    (map #(m/assoc-some % :based_on_upload (based-on-upload %))
-         models)))
+        query?               (fn [model] (= "query" (name (:query_type model "query"))))
+        uploadable?          (comp (uploadable-table-ids table-ids) :table_id)]
+    (for [model models]
+      (m/assoc-some
+        model
+        :based_on_upload
+        (when-let [query (some-> model :dataset_query lib/->pMBQL)]
+          (when (and (query? model) (uploadable? model) (no-joins? query))
+            (lib/source-table-id query)))))))
 
 (mi/define-batched-hydration-method based-on-upload
   :based_on_upload
@@ -763,5 +759,6 @@
     - uploads are enabled
   Otherwise based_on_upload is nil."
   [cards]
-  (let [models-by-id (m/index-by :id (model-hydrate-based-on-upload (filter :dataset cards)))]
-    (map #(or (models-by-id (:id %)) %) cards)))
+  (let [id->model   (m/index-by :id (model-hydrate-based-on-upload (filter :dataset cards)))
+        card->model (comp id->model :id)]
+    (map #(or (card->model %) %) cards)))
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -743,11 +743,11 @@
         has-uploadable-table? (comp (uploadable-table-ids table-ids) :table_id)]
     (for [model models]
       (m/assoc-some
-        model
-        :based_on_upload
-        (when-let [query (some-> model :dataset_query lib/->pMBQL)]
-          (when (and (mbql? model) (has-uploadable-table? model) (no-joins? query))
-            (lib/source-table-id query)))))))
+       model
+       :based_on_upload
+       (when-let [query (some-> model :dataset_query lib/->pMBQL not-empty)] ; dataset_query can be empty in tests
+         (when (and (mbql? model) (has-uploadable-table? model) (no-joins? query))
+           (lib/source-table-id query)))))))
 
 (mi/define-batched-hydration-method based-on-upload
   :based_on_upload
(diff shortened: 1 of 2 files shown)


src/metabase/upload.clj
```diff
@@ -763,5 +759,6 @@
     - uploads are enabled
   Otherwise based_on_upload is nil."
   [cards]
-  (let [models-by-id (m/index-by :id (model-hydrate-based-on-upload (filter :dataset cards)))]
```
crisptrutski, line 766: This naming threw me, given the long discussion we had against using this naming schema 😆

calherries: Ha, maybe this is a different scenario where `x-by-y` adds to readability. `models-by-id` at least implies that it is a map of a _model_ ID to a model (maybe it should rather be `model-by-id`), as opposed to card ID to a model. Because the name reminds me of how it was created: either `m/index-by` or `clojure.core/group-by`. But I think `id->model` is also fine.

The use of `card->model` here is a little surprising for me though. The english description of this function based on its behaviour would be "Returns a hydrated model if a card is a model". Yet to me the name `card->model` implies that there is one model per card. Not an opinion I hold strongly though

src/metabase/upload.clj
```diff
@@ -742,16 +739,15 @@
...
-                               (when-let [dataset_query (:dataset_query model)] ; dataset_query is sometimes null in tests
-                                 (let [query (lib/->pMBQL dataset_query)]
-                                   (when (and (some-> model :query_type name (= "query"))
-                                              (contains? uploadable-table-ids (:table_id model))
-                                              (no-joins? query))
-                                     (lib/source-table-id query)))))]
-    (map #(m/assoc-some % :based_on_upload (based-on-upload %))
-         models)))
+        query?               (fn [model] (= "query" (name (:query_type model "query"))))
+        uploadable?          (comp (uploadable-table-ids table-ids) :table_id)]
```
calherries, line 743: This function takes a model as an arg, yet doesn't determine whether a model is "uploadable" or not. A more appropriate name would be `has-uploadable-table-id?` but at that point you might as well inline it
```suggestion
        uploadable-table?          (comp (uploadable-table-ids table-ids) :table_id)]
```

src/metabase/upload.clj
```diff
@@ -742,16 +739,15 @@
...
-        based-on-upload      (fn [model]
-                               (when-let [dataset_query (:dataset_query model)] ; dataset_query is sometimes null in tests
-                                 (let [query (lib/->pMBQL dataset_query)]
-                                   (when (and (some-> model :query_type name (= "query"))
-                                              (contains? uploadable-table-ids (:table_id model))
-                                              (no-joins? query))
-                                     (lib/source-table-id query)))))]
-    (map #(m/assoc-some % :based_on_upload (based-on-upload %))
-         models)))
+        query?               (fn [model] (= "query" (name (:query_type model "query"))))
```
calherries, line 742: `query` is so overloaded that this name is a little confusing. We frequently refer to `query_type=query` questions as "MBQL questions" so I would change this to `mbql?` instead.
```suggestion
        mbql?               (fn [model] (= "native" (name (:query_type model))))
```
See 
https://github.com/metabase/metabase/blob/1ca43e71eddea3ac8f884a3216fbd7ca03ef4623/src/metabase/mbql/schema.cljc#L1804

src/metabase/upload.clj
crisptrutski: How about `uploadable-id?` I have a pretty strong aversion to mixing abstraction levels in the `(when (and x y z) ...)` cluster, a capital offense in my fashion guide.

src/metabase/upload.clj
crisptrutski: How about `card->maybe-model` then?

src/metabase/upload.clj
crisptrutski: Or `uploadable-table`?

calherries: `uploadable-id?` suffers from the same problem because the "id" in the name is ambiguous. `uploadable-table?` is pretty clear though.

For myself I don't have such a strong aversion to mixing abstration levels. This is all in the name of readability and introducing extra names comes at a cost because they can be misleading and will make your eyes jump around if you have to understand them.

src/metabase/upload.clj
calherries: What's the point?

calherries: Ah, this is about mixing abstraction levels inside an `or`. I repeat the point here: introducing extra names comes at a cost because they can be misleading and will make your eyes jump around if you have to understand them.

calherries: `card->maybe-model` isn't misleading, so I'm happy with that name

src/metabase/upload.clj
crisptrutski: Cool, I like that predicate name. I think you forgot a `not=` here after changing to `"native"`?

src/metabase/upload.clj
crisptrutski: Also happy if you want to inline it, I felt it helped a little bit with the density of the code but not attached.

src/metabase/upload.clj
calherries: Ah yeah sorry, the original implementation was fine. Updated

src/metabase/upload.clj
calherries: I like the rule. I don't mind that much with `card->maybe-model`. Apologies for splitting hairs again on naming, maybe why I like inlining things is because I'm so critical of names

crisptrutski: Don't miss naming day in the gym! In the mean time I'll be doing some "reading code that is more dense than I'm used to" reps.

src/metabase/upload.clj
crisptrutski: That `has-` hurts me - especially at it forces a re-indent on the let bindings, but agree that without it there's a bit more ambiguity though - `uploadable-table?` could make it sound like there is "exactly one" uploadable tables, when really its "at least one". We rely on the lack of joins to make it exact, and that call is more expensive so it makes sense to keep it later in the `and`.

src/metabase/upload.clj
calherries: oh you should be aware of another reason I don't like names: I have an aversion to newlines because I work without an external monitor and hate scrolling :D

src/metabase/upload.clj
calherries: Yeah. To understand what the table in `uploadable-table?` means you almost need to know that it's accessing the table from the `source_table` from the first stage of the query. It's unfortunately complicated

src/metabase/upload.clj
crisptrutski: Is it at least 16"? 😆

I feel so blind when I'm just on one small screen for pairing, for sure following references is less painful when they open in a split pane.

calherries:
The tests are failing for some reason I can't deduce from inspection

crisptrutski:
> The tests are failing for some reason I can't deduce from inspection

I also can't grok it. Hoping it's a transient CI issue, will investigate if it still isn't passing this afternoon.

src/metabase/upload.clj
calherries: Yes, 16. Before metabase it was even smaller.

calherries:
It doesn't seem transient CI related, it's failing on the exact functions you're modifying

calherries (approved):
LGTM, I fixed the "bug" causing the test failures. It was because the `dataset_query` can be `{}` in tests that don't test the query processor, so `no-joins?` ran into a NPE because MLv2 expects valid queries

Issue #37368, opened 2024-01-08
FE - Metric page (`/metric/:id` + `/metric/:id/notebook`)


kamilmielnik:
Depends on #37348 but we can start working on it once #38262 is done

----

- [x] Display the metric page.
~- [ ] The metric displays as a time series using the primary time dimension by default when opened from anywhere e.g. search, collections, dashboards~
- [x] When a user alters a metric in chill mode they “dirty” the metric and it turns into an ad-hoc question like models do today. Where the header is now greyed out and clicking returns to the Metric itself.

----

Bookmarking has a dedicated task: #37359

kamilmielnik:
Closed by #38554

PR #38512, opened 2024-02-07
New Metric page (`/metric/query`)

### LINKED ISSUE #37353: FE - New Metric page (`/metric/query`)
Depends on:
- #37348
- #37352

----

- [x] Allow creating and saving a new metric.
- [x] Do not show Metadata tab for now

Enforcing Metric clause restrictions has a dedicated task: #37355
The right-hand sidebar is for later too: #37971

kamilmielnik:
Closes #37353

----

https://github.com/metabase/metabase/assets/6830683/7c98d89e-4c26-4529-97e7-0482bed8e994

### How to verify

1. "+ New" > "Metric"
2. Choose any data source
3. Choose any query shape (enforcing correct Metric query shape will be done in #37355)
4. Visualize the query - URL should still be `/metric/*`
5. Save - you should be able to provide description & choose any parent Collection
    - :warning:  Make sure to include the word "metric" in the metric name so that our mocks kick in
6. You should be redirected to `/metric/:slug` page (this page does not exist yet, it's being added in #37368)


frontend/src/metabase/lib/urls/questions.ts
```diff
@@ -105,9 +107,7 @@ export function newQuestion({
     creationType,
     query: objectId ? { objectId } : undefined,
   });
-
-  const entity = question.isDataset() ? "model" : "question";
-  const type = question.type() ?? entity;
+  const type = question.type();
```
kamilmielnik, line 110: The fallback is no longer needed thanks to 98b8579223391532c87276f89acbf271012c8c5f

frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
```diff
@@ -136,9 +138,10 @@ const NewItemMenu = ({
         title: t`Metric`,
         icon: "metric",
         link: Urls.newQuestion({
-          mode: "notebook",
-          collectionId,
+          mode: "query",
+          creationType: "custom_question",
```
nemanjaglumac, line 142: Let's remove this until it's needed for analytics.

kamilmielnik: d07a15c1e79b1362757af80cb222064938014aa7

frontend/src/metabase/query_builder/actions/core/updateQuestion.ts
```diff
@@ -129,17 +131,19 @@ export const updateQuestion = (
...
 
-    // This scenario happens because the DatasetQueryEditor converts the dataset/model question into a normal question
+    // This scenario happens because the DatasetQueryEditor converts the model/metric question into a normal question
     // so that its query is shown properly in the notebook editor. Various child components of the notebook editor have access to
     // this `updateQuestion` action, so they end up triggering the action with the altered question.
     if (queryBuilderMode === "dataset" && !newQuestion.isDataset()) {
-      newQuestion = newQuestion.setType("model");
+      const routing = getRouting(getState());
+      const type = getCardTypeFromLocation(routing.locationBeforeTransitions);
+      newQuestion = newQuestion.setType(type);
```
kamilmielnik, line 146: This is an extension to the existing "hack".
Now `queryBuilderMode === "dataset"` is shared by both models and metrics.
URL is the only place where I could find information on what type of entity are we dealing with.

Alternative solutions I considered:
- remove the original hack - [this is the best option, but also the most difficult](https://metaboat.slack.com/archives/C505ZNNH4/p1707400291390129)
- add one more piece of information to `state.qb.uiControls` indicating what kind of entity did we start with - simple, but might have edge cases
- introduce a new `queryBuilderMode` - would require a lot of changes

kamilmielnik: I managed to get rid of this: c1d3db383613b4bdebfdbfcfb1620b144d3e1503

frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
```diff
@@ -136,9 +138,9 @@ const NewItemMenu = ({
         title: t`Metric`,
         icon: "metric",
         link: Urls.newQuestion({
-          mode: "notebook",
-          collectionId,
+          mode: "query",
```
nemanjaglumac, line 141: I don't understand why did we change this to `query`? If you take a look at the question flow, the URL says `/notebook` which seems to reflect the UI state correctly.

Now, compare it to the metric:
image

Unless there was a really good reason for this change, I think we should change it back to the "notebook".

kamilmielnik: > If you take a look at the question flow

We want to follow the model flow.

frontend/src/metabase/containers/SaveQuestionModal.tsx
```diff
@@ -170,27 +190,18 @@ export const SaveQuestionModal = ({
...
     isSavedQuestionChanged &&
     originalQuestion != null &&
     originalQuestion.canWrite();
 
-  const singleStepTitle = getSingleStepTitle(questionType, showSaveType);
-
+  const { multiStepTitle, singleStepTitle, nameInputPlaceholder } = getLabels(
+    question,
+    showSaveType,
+  );
```
nemanjaglumac, line 202: So much cleaner 👏

frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx
```diff
@@ -444,21 +444,26 @@ function DatasetEditor(props) {
...
+                { id: "query", name: t`Query`, icon: "notebook" },
+                {
+                  id: "metadata",
+                  name: t`Metadata`,
+                  icon: "label",
+                  disabled: !resultsMetadata,
+                },
+              ]}
+            />
+          )
```
nemanjaglumac, line 466: 👌

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -350,7 +350,10 @@ class View extends Component {
       );
     }
 
-    if (question.isDataset() && queryBuilderMode === "dataset") {
+    const isModelOrMetric =
+      question.type() === "model" || question.type() === "metric";
```
nemanjaglumac, line 354: Did you intentionally avoid using `question.type() !== "question"` for the sake of being more explicit?

kamilmielnik: Also because `question.type()` can be `undefined`

frontend/src/metabase/components/NewItemMenu/NewItemMenu.tsx
nemanjaglumac: Ok, I was about to correct myself because I've just opened a model and saw that it has a `/query` in the URL.

Admittedly, this is adding new layer of confusion but that has nothing to do with your changes here.
Ignore my comment.

frontend/src/metabase/query_builder/components/view/View.jsx
nemanjaglumac: I thought that we're falling back to "question" if it's undefined.
But yeah, until we have the type safety here, this makes sense.

Thank you

nemanjaglumac (approved):
Very clean.
Let's merge this. 🚢

kamilmielnik:
> Very clean. Let's merge this. 🚢

Thanks for review @nemanjaglumac :pray: 
Some unit tests are still failing, I need to sort it out first.
If changes turn out to be large I'll request a re-review.

frontend/src/metabase/query_builder/actions/core/updateQuestion.ts
```diff
@@ -129,17 +130,23 @@ export const updateQuestion = (
...
       // to start building a new ad-hoc question based on a dataset
-      if (newQuestion.isDataset()) {
+      if (newQuestion.type() === "model" || newQuestion.type() === "metric") {
         newQuestion = newQuestion.setType("question");
         dispatch(onCloseQuestionInfo());
       }
     }
 
-    // This scenario happens because the DatasetQueryEditor converts the dataset/model question into a normal question
+    // This scenario happens because the DatasetQueryEditor converts the model/metric question into a normal question
```
nemanjaglumac, line 139:
```suggestion
    // This scenario happens because the DatasetQueryEditor converts a model or a metric into a normal question
```

The previous change could be easily misinterpreted.

PR #38513, opened 2024-02-07
Fix the problem with unknown fields in Dashcard mapping

uladzimirdev:

### Description

Because of differences between MLv2 and MLv1, `/api/dashboard/:id` sends `parameter_mappings` without `base-type` attribute.

This PR implements naive approach to compare targets, proper fix must be done on BE IMO

image

### How to verify
- you need to have a dashboard with filters and questions with join-alias specified
- edit this dashboard, click on filter, map filter to the field with join alias
- save dashboard
- edit again, open mappings - you'll see `unknown field`

### Checklist


frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -98,9 +98,22 @@ export function DashCardCardParameterMapper({
...
+
+        // This trick is needed because MLv2 has `base-type` in target[1]
+        // but response from BE doesn't provide `base-type`, so isEqual fails
+        return (
+          normalizedTarget[0] === normalizedOptionTarget[0] &&
+          _.isEqual(
+            normalizedTarget[1].slice(0, -1),
+            normalizedOptionTarget[1].slice(0, -1),
+          )
+        );
```
kulyk, line 114: I think this is only suitable for targets of `ParameterDimensionTarget` type, but not [other kinds of target objects](https://github.com/metabase/metabase/blob/98015a2d5f0cae574114aee1d3ca6c5df4893083/frontend/src/metabase-types/api/parameters.ts#L72). And we should be adding this kind of logic to `metabase-lib/parameters` (maybe we should update [`compareMappingOptionTargets` util](https://github.com/metabase/metabase/blob/98015a2d5f0cae574114aee1d3ca6c5df4893083/frontend/src/metabase-lib/parameters/utils/targets.ts#L60C17-L60C44)?)

kulyk (changes requested):

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
uladzimirdev: @kulyk as far as I get, we work here with mappings, generated by the next functions:

https://github.com/metabase/metabase/blob/master/frontend/src/metabase/parameters/utils/mapping-options.js#L31 -> https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/parameters/utils/targets.ts#L49

(for natives it's https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/parameters/utils/targets.ts#L41)

so both will be "dimension" first and "legacyRef" second

uladzimirdev:
shoud be a better fix https://github.com/metabase/metabase/pull/38545

PR #38514, opened 2024-02-07
Auto focus search input in question picker modal

kulyk:
IIRC, Mantine keeps its `Modal` components mounted, so `autoFocus` on inputs doesn't really work. It's expected to use a `data-autofocus` attribute instead (see [docs](https://v6.mantine.dev/core/modal/#initial-focus))

### Demo

https://github.com/metabase/metabase/assets/17258145/48cb4bbe-9232-4f30-966c-109555c2969c


kdoh (approved):

uladzimirdev (approved):

Somtom (approved):
Nice. Thanks for tackling that fast!

PR #38518, opened 2024-02-07
Improve look of models in Browse data

rafpaf:
This PR improves the appearance of the Browse data page. Specifically, it makes the cards representing models in Browse data look more like the cards on Collections pages. It also brings the Browse data link to the top of the nav.

### on `master`

image

### on this branch

image

### DIFF
--- a/.github/workflows/uberjar.yml
+++ b/.github/workflows/uberjar.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - "master"
-      - "models-in-browse-data-polish"
     paths-ignore:
       # config files
       - ".**"
--- a/.github/workflows/uberjar.yml
+++ b/.github/workflows/uberjar.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - "master"
+      - "models-in-browse-data-polish"
     paths-ignore:
       # config files
       - ".**"
--- a/frontend/src/metabase/browse/components/BrowseModels.tsx
+++ b/frontend/src/metabase/browse/components/BrowseModels.tsx
@@ -148,7 +148,7 @@ const ModelCell = ({ model, collectionHtmlId }: ModelCellProps) => {
     >
       <ModelCard>
         <Box mb="auto">
-          <Icon name="model" size={20} className={color("brand")} />
+          <Icon name="model" size={20} color={color("brand")} />
         </Box>
         <Title lh="1rem" mb=".25rem" size="1rem">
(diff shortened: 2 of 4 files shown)


.github/workflows/uberjar.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - "master"
+      - "models-in-browse-data-polish"
```
rafpaf, line 7: we must remove this before merging

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -140,29 +139,24 @@ const ModelCell = ({ model, collectionHtmlId }: ModelCellProps) => {
...
   return (
     <Link
       aria-labelledby={`${collectionHtmlId} ${headingId}`}
       key={model.id}
       to={Urls.model(model as unknown as Partial<Card>)}
     >
       <ModelCard>
-        <Title order={4} className="text-wrap" lh="1rem" mb=".5rem">
+        <Box mb="auto">
+          <Icon name="model" size={20} className="text-brand" />
```
iethree, line 153: we're trying to avoid using classNames these days

```suggestion
          <Icon name="model" size={20} className={color("brand")} />
```

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -140,29 +139,24 @@ const ModelCell = ({ model, collectionHtmlId }: ModelCellProps) => {
...
+        </Box>
+        <Title className="text-wrap" lh="1rem" mb=".25rem" size="1rem">
```
iethree, line 155: let's not use a className, also I don't think it does anything

```suggestion
        <Title lh="1rem" mb=".25rem" size="1rem">
```

iethree (approved):
few little nits

iethree:
couple design suggestions, can be follow-up

Maybe we should align styling for models and databases?

 Screen Shot 2024-02-08 at 10 45 29 AM | Screen Shot 2024-02-08 at 10 45 24 AM

should tooltips go down here so as not to block the card you're looking at?

Screen Shot 2024-02-08 at 11 05 21 AM

I've always kind of wanted 1px of margin between sidebar items so you get some separation when you're hovering something next to a selected item. 

Screen Shot 2024-02-08 at 11 04 24 AM

kdoh (approved):
LGTM (reviewed local host, not the code specifically)

PR #38519, opened 2024-02-07
Use data fetching hooks for collections content

iethree:
### Description

- uses data fetching hooks instead of the redux `connect()` function with nested entity loader HoCs for collections list
- adds a bookmarks api hook
- splits up the CollectionContent and CollectionContentView components, and converts the former to typescript
- moves CollectionContent files into their own cozy little subfolder in the `collections/components/` directory

**Benefits**

- the code is more explicit about where the data is coming from and how loading or error states are triggered
- gets rid of a bunch of layers of nesting in the react component tree from the HoCs
- theoretically faster since api requests can happen in parallel rather than sequentially now

. | .
---|---
Before | Screen Shot 2024-02-06 at 5 12 00 PM
After | Screen Shot 2024-02-06 at 5 12 33 PM

_(unclear if this performance benefit actually holds up in the real world)_

**Drawbacks**

- the lack of entity-loader HoC magic makes this a bit more boilerplatey

### How to verify

Collections should work!

### Demo

### Checklist


frontend/src/metabase/collections/containers/CollectionContent.jsx
```diff
@@ -363,20 +343,75 @@ function CollectionContent({
...
+
+  if (!bookmarks || !databases || !collections || !collection) {
+    return <LoadingAndErrorWrapper loading />;
+  }
+
+  const error =
+    bookmarksError || databasesError || collectionsError || collectionError;
+  if (error) {
+    return <LoadingAndErrorWrapper error={error} />;
+  }
```
sloansparger, line 399: if you get an error but no data, won't we keep showing the loading state? might be worth destructing the loading state off of all the `useXQuery` methods above.

frontend/src/metabase/collections/containers/CollectionContent.jsx
```diff
@@ -363,20 +343,75 @@ function CollectionContent({
...
 
-export default _.compose(
-  Bookmark.loadList(),
-  Databases.loadList(),
-  Collection.loadList({
-    query: {
-      tree: true,
-      "exclude-other-user-collections": true,
-      "exclude-archived": true,
+export function CollectionContent({ collectionId }) {
```
sloansparger, line 346: !nit / maybe not worth doing

this is a huge file, you could split out the CollectionContentView to its own .jsx file and turn this one into .tsx.

sloansparger (approved):
approved with comment

frontend/src/metabase/collections/containers/CollectionContent.jsx
iethree: we do a good bit of reloading data in these components, so we don't want to be showing a loading spinner every time it goes back into loading state. I moved the error check above the loading check to make sure we hit that and don't end up with forever-loading states.

frontend/src/metabase/collections/containers/CollectionContent.jsx
iethree: good idea! I did this!

frontend/src/metabase/collections/components/CollectionContent/CollectionContent.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const isAdmin = useSelector(getUserIsAdmin);
+  const isNavbarOpen = useSelector(getIsNavbarOpen);
+
+  const dispatch = useDispatch();
+
+  const createBookmark = (id: BookmarkId, type: BookmarkType) =>
+    dispatch(Bookmark.actions.create({ id, type }));
+  const deleteBookmark = (id: BookmarkId, type: BookmarkType) =>
+    dispatch(Bookmark.actions.delete({ id, type }));
```
sloansparger, line 74: !nit - don't feel like you have to change anything if you don't want to, but these hooks feel very technical.
something like this feels less cluttery imo:
```ts
const [uploadDbId, uploadsEnabled] = useSettings(["uploads-database-id", "uploads-enabled"]);
const { canUpload } = useDatabase(uploadDbId);
const { isAdmin } = useCurrentUser();
const { isOpen } = useNav():
const { create, delete } = useBookmarks();
```

similarly, the data fetching at the top of this render block is a pattern i keep seeing used. we fetch like 3-4 entities and then combine the values like `const error = bookmarksError || databasesError || collectionsError || collectionError;`. Feels like there's an opportunity to make this one hook that does this for us behind the scenes.

frontend/src/metabase/collections/components/CollectionContent/CollectionContent.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const error =
+    bookmarksError || databasesError || collectionsError || collectionError;
+
+  if (error) {
+    return <LoadingAndErrorWrapper error={error} />;
+  }
+
+  if (!bookmarks || !databases || !collections || !collection) {
+    return <LoadingAndErrorWrapper loading />;
+  }
```
sloansparger, line 93: I'm guessing our types are too loose, but I think observing the actual loading state would be preferable. Guessing you're needing to guard values as being defined before the code below though...

frontend/src/metabase/collections/components/CollectionLanding/CollectionLanding.tsx
```diff
@@ -16,11 +17,14 @@ const CollectionLanding = ({
   children,
 }: CollectionLandingProps) => {
   const collectionId = extractCollectionId(slug);
-  const isRoot = collectionId === "root";
+
+  if (!isNotNull(collectionId)) {
```
sloansparger, line 21: Can we avoid the double negative?

frontend/src/metabase/common/hooks/use-bookmark-list-query/use-bookmark-list-query.unit.spec.tsx
```diff
@@ -0,0 +1,80 @@
+import {
```
sloansparger, line 1: would using `renderHook` make more sense than testing a custom component?
https://testing-library.com/docs/react-testing-library/api/#renderhook

sloansparger (approved):
Approved with comments! Great work!

frontend/src/metabase/collections/components/CollectionContent/CollectionContent.tsx
iethree: I like the way you're thinking here, but that's a different PR

PR #38521, opened 2024-02-07
analytics for meet-embedders

npretto:
[[Epic] Meet embedders at the door - minimal edition](https://github.com/metabase/metabase/issues/38233)

### Description

This modifies the `trackStepSeen` function to accept both the step name and the step number, as the two are not linked anymore as we have conditional steps.

[slack thread about the above change](https://metaboat.slack.com/archives/C01MC2CKW2G/p1706719539434809), I'll also need to change the notion page for the event

It also adds a `trackUsageReasonSelected` function + event.

If read commit by commit it should be self describing, i made one commit to just copy the old schema to make it easier to see the changes.

I *think* that the new schema should be a patch update, as the new schema still accepts old events, but let me know if I got it wrong

### DIFF
--- a/snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-2
+++ b/snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-2
@@ -1,0 +1,88 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Setup flow",
+  "self": {
+    "vendor": "com.metabase",
+    "name": "setup",
+    "format": "jsonschema",
+    "version": "1-0-1"
+  },
+  "type": "object",
+  "properties": {
+    "event": {
+      "description": "Event name",
+      "type": "string",
+      "enum": [
+        "step_seen",
+        "database_selected",
+        "add_data_later_clicked"
+      ],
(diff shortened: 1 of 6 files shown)


snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-2
```diff
@@ -0,0 +1,104 @@
...
+      ],
+      "maxLength": 1024
+    },
+    "step_number": {
+      "description": "Integer describing the order of the referenced step",
+      "type": [
+        "integer",
+        "null"
+      ],
+      "enum": [
```
oisincoveney, line 52: What if we made the maximum of 6 instead of an enum here?

oisincoveney (approved):
Looks good from a code standpoint. I'm happy as long as the tests pass

snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-2
npretto: Ah, didn't even see that lol. (I think it's going to fail at the last step as it is not)
I'm not sure why we even have that, could it be that it's for optimization purposes?
As we'd release this tomorrow, I'd follow the Chesterton's Fence rule and just add "6" to the enum, we can figure a proper solution when we'll add the license activation step

PR #38530, opened 2024-02-07
Connection Impersonation UI for Redshift

iethree:
frontend for https://github.com/metabase/metabase/issues/38445

### Description

Our impersonation for redshift uses "users" rather than "roles" this updates the copy to reflect that.

Postgres | Redshift
--- | ---
Screen Shot 2024-02-09 at 1 22 37 PM | Screen Shot 2024-02-09 at 1 22 24 PM

### How to verify

a) run the unit test 😁
b) connect a redshift db and see that it changes the copy

### Checklist


npfitz (approved):
I don't have a redshift instance to connect to and test this out, but everything looks like it should work

PR #38531, opened 2024-02-07
v2 audit views and migrations

### LINKED ISSUE #37244: Deal with ApiKey users in the analytics views
ApiKey users appear in the analytics views (which is good!) but because they appear as regular users, things are a bit odd.

Screenshot 2024-01-03 at 8 50 30 AM

We should think about how exactly we want ApiKey users to appear in the analytics database - for example, at the least, we probably don't want `email` to appear for ApiKey users, since it's a dummy email address.

Some possibilities: we could update the definition of the `v_users` view to:
- set `email` to `""` for users with type `api-key`?
- add the `type` column and add that column to the analytics model?
- add an `api_key_prefix` column by joining to the `api_key` table and add that to the analytics model?

I think all of these would be relatively straightforward.

### LINKED ISSUE #38524: Update auditv2 content for 49
**Context**
- [x] Join query into query_execution (modeling)
- [x] Replace external with anonymous group
- [x] Do something about API key users
- [x] Check if dashboard overview works well with embedding data

noahmoss:
Adds SQL migrations for new audit versions of audit views for Metabase 49, and liquibase changeSets to run them.


noahmoss:
* `master`
  * **PR #38531** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38531?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38531?utm_source=stack-comment).

luizarakaki (approved):

Issue #38502, opened 2024-02-07
Improve dashboard selector `getParameters` to avoid unnecessary heavy operations


uladzimirdev:
**Context**
Improve dashboard selector `getParameters` to avoid unnecessary heavy operations

[suggestion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1707156521548289?thread_ts=1706701672.407039&cid=C04CYTEL9N2)
```
why typing is so slow is because our getParameters[ selector](https://github.com/metabase/metabase/blob/ea05f6ff8dc09692761514423400ccef0cab5af8/frontend/src/metabase/dashboard/selectors.ts#L305) is using getDashboardComplete even tho it only cares about dashboard.parameters and dashcards’ parameter_mappings

We might not need debouncing heading/text change events if we could optimize this
Basically we only need to rerun getParameters when:
- metadata changes (can be frequent when initially loading a busy dashboard, mostly stale later)
- dashboard parameters (their values are decoupled, so dashboard.parameters should only change in edit mode if I’m not missing something)
- dashcard parameter mappings (again, should only be changing while in edit mode)
```

uladzimirdev:
tested it https://github.com/metabase/metabase/pull/38765 but didn't notice any significant change in perf

PR #38540, opened 2024-02-08
Fix CI failure on master due to selector changes

WiNloSt:
Related to #38502
Likely caused by #38425

### Description

`getParameters` now return a new reference despite having thet same value.

The fix is temporary by introducing another cache layer in embedding modal to deep compare the value of `parameters` instead of relying on the reference change alone.

### How to verify

Run tests in `e2e/test/scenarios/embedding/reproductions/37914-embed-preview-parameter-values-not-working.cy.spec.js` and it should pass.

### Checklist


uladzimirdev:
I think I came up to a proper solution https://github.com/metabase/metabase/pull/38548

WiNloSt:
This is superseded by https://github.com/metabase/metabase/pull/38548

Issue #38499, opened 2024-02-07
Newly added MySQL DB tables not showing up, but old ones do show up


jfang:
**Describe the bug**
After the upgrade to Metabase v0.48.4, I am not able to see any newly added MySQL DB tables even after I have clicked on the "Sync database schema now" and "Re-scan field values now" buttons. The old tables that were created before the upgrade is showing up. Updating to v0.48.5 does not seem to resolve this issue.

When I go to "+ New" and click on "SQL Query", I am able to pull up the data from my newly added table. However, when I go to "+New" and click on "Question", the newly added table does not show up in the "Pick your starting data" section. Only the old tables are there.

It seems like the **"Sync database schema now" functionality** is not working correctly.

**Logs**
Feb 06 20:34:372024-02-07 04:34:37,037 DEBUG middleware.log :: GET /api/database/2/schemas 200 34.6 ms (4 DB calls) App DB connections: 1/4 Jetty threads: 4/50 (3 idle, 0 queued) (48 total active threads) Queries in flight: 0 (0 queued)
Feb 06 20:34:372024-02-07 04:34:37,149 DEBUG middleware.log :: GET /api/database/2/schema/ 200 56.8 ms (6 DB calls) App DB connections: 2/4 Jetty threads: 4/50 (3 idle, 0 queued) (48 total active threads) Queries in flight: 0 (0 queued)
Feb 06 20:34:372024-02-07 04:34:37,191 DEBUG middleware.log :: GET /api/database 200 170.7 ms (8 DB calls) App DB connections: 1/4 Jetty threads: 3/50 (3 idle, 0 queued) (49 total active threads) Queries in flight: 0 (0 queued)
Feb 06 20:35:332024-02-07 04:35:33,570 DEBUG middleware.log :: GET /api/database/2 200 25.0 ms (3 DB calls) App DB connections: 1/4 Jetty threads: 3/50 (4 idle, 0 queued) (49 total active threads) Queries in flight: 0 (0 queued)
Feb 06 20:35:352024-02-07 04:35:35,524 INFO metabase.events :: Loading events namespace: metabase.events.audit-log 👂
Feb 06 20:35:352024-02-07 04:35:35,602 INFO metabase.events :: Loading events namespace: metabase.events.driver-notifications 👂
Feb 06 20:35:352024-02-07 04:35:35,632 INFO metabase.events :: Loading events namespace: metabase.events.last-login 👂
[...]

PR #38542, opened 2024-02-08
Ignore REVOKE commands when processing mysql grants

crisptrutski:
References https://github.com/metabase/metabase/issues/38499

### Description

A customer is experiencing a "hang" in the database sync process due to an exception being thrown while parsing the database grants.

Neither this area of the driver, nor any related Java dependencies, have changed since v0.48.3, so I cannot explain the customer was not seeing the same issue when connecting to the same database before.

This change adds a clause to explicitly ignore the relevant grant expression from the bug report, and also adds a catch-all case to "fail open" with a warning on any expressions we do not recognize in future.

### DIFF
--- a/src/metabase/driver/mysql.clj
+++ b/src/metabase/driver/mysql.clj
@@ -789,6 +789,9 @@
     :roles #{'example_role_1'@'%' 'example_role_2'@'%'}}"
   [grant]
   (condp re-find grant
+    ;; Typically these statements are used to revoke mutation rights, and for now we ignore them.
+    #"^REVOKE "
+    nil
     #"^GRANT PROXY ON "
     nil
     #"^GRANT (.+) ON FUNCTION "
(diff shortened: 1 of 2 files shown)


crisptrutski:
At first I naively thought that we didn't track mutation rights for users, but I see now that we do. It shouldn't be too much work to add basic support for revoking those permissions, but I suspect it's better to get this simpler, more permissive version out first to see if it's sufficient to unblock the customer. We are for example not currently processing column privileges either.

crisptrutski:
After discussion with @calherries, we'll rather wait on handling negative permissions until we at least process the revocation of SELECT rights.

calherries:
For reference, MySQL supports partial revokes with the partial_revokes system variable enabled. We didn't anticipate this in the `driver/current-user-table-privileges` implementation used in sync.
https://dev.mysql.com/doc/refman/8.0/en/partial-revokes.html

PR #38544, opened 2024-02-08
Partially migrate is Question.prototype.isDataset

kamilmielnik:
This PR migrates roughly half of remaining usages of deprecated `Question.prototype.isDataset` in favor of `Question.prototype.type` - only those usages, where I had a good confidence they won't break anything.

Some of remaining usages should be ported as part of the tasks planned for Metrics v2 project.
The rest we will have to manually go over at the end of the project.

### How to verify

Everything should work the same as before for questions and models.


frontend/src/metabase/query_builder/components/SavedQuestionIntroModal.tsx
kamilmielnik, line 1: This component was converted to TS and a functional component

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -396,6 +396,7 @@ class View extends Component {
         {isShowingNewbModal && (
           <SavedQuestionIntroModal
             question={question}
+            isShowingNewbModal={isShowingNewbModal}
```
kamilmielnik, line 399: The missing prop came out after converting `SavedQuestionIntroModal` to TS

frontend/src/metabase/query_builder/components/SavedQuestionIntroModal.tsx
```diff
@@ -0,0 +1,59 @@
+import { t } from "ttag";
```
kamilmielnik, line 1: This component was converted to TS and a functional component

frontend/src/metabase/questions/containers/ArchiveQuestionModal.jsx
```diff
@@ -14,6 +14,33 @@ const mapDispatchToProps = {
...
+      message: t`This model will be removed from any dashboards or pulses using it.`,
+    };
+  }
+
+  if (type === "metric") {
+    return {
+      title: t`Archive this metric?`,
+      message: t`This metric will be removed from any dashboards or pulses using it.`,
+    };
+  }
```
nemanjaglumac, line 39:
```suggestion
  const allowedTypes = ["question", "model", "metric"];
  
  if (allowedTypes.includes(type)) {
    return {
      title: t`Archive this ${type}?`,
      message: t`This ${type} will be removed from any dashboards or pulses using it.`,
    };
  }
```

WDYT about something like this?

frontend/src/metabase/dashboard/actions/navigation.js
```diff
@@ -80,15 +80,18 @@ export const navigateToNewCardFromDashboard = createThunkAction(
         previousQuestion.query(),
       );
 
-      const isDrillingFromNativeModel =
-        previousQuestion.isDataset() && isPreviousNative;
+      const isFromModelOrMetric =
+        previousQuestion.type() === "model" ||
+        previousQuestion.type() === "metric";
+      const isDrillingFromNativeModelOrMetric =
+        isFromModelOrMetric && isPreviousNative;
```
nemanjaglumac, line 87: I don't know the context here, but why do we have to include metrics? Wasn't this check concerned literally only about the native models before?

frontend/src/metabase/questions/containers/ArchiveQuestionModal.jsx
kamilmielnik: I intentionally avoided it. Dynamic strings come with some degree of magic.
With static strings it's easier to add translations and find translation keys using simple text search.

frontend/src/metabase/dashboard/actions/navigation.js
kamilmielnik: Some behaviors will stay only for models, some behaviors will apply to both metrics and models.

Since both metrics and models cannot be edited in chill mode (i.e. instead they get wrapped in ad-hoc questions; if someone wants to edit model or a metric they need to go to "edit query definition" instead) I figured drills is one of those places where the behavior for models and metrics will be the same.

kamilmielnik:
I fixed tests and reverted one breaking change (I will do it in another PR).

enterprise/frontend/src/metabase-enterprise/moderation/index.js
```diff
@@ -45,9 +45,7 @@ if (hasPremiumFeature("content_verification")) {
           {
             title: isVerified
               ? t`Remove verification`
-              : isDataset
-              ? t`Verify this model`
-              : t`Verify this question`,
+              : getVerifyQuestionTitle(model),
```
nemanjaglumac, line 48: 😍

nemanjaglumac (approved):
I didn't see anything that sticks out in the code, and the tests are all green 🚢

PR #38545, opened 2024-02-08
Rework parameter mapping comparing to MLv2

uladzimirdev:

it's a proper fix instead of a workaround from https://github.com/metabase/metabase/pull/38513

### Description

Target is a `legacyRef`, so we can try to find a column from question instead of comparing targets

### How to verify

- you need to go to dashboard parameter mapping, choose some value and verify it's defined as "Unknown field"

### Before
image

After 

https://github.com/metabase/metabase/assets/125459446/856038fe-193a-40bb-bfbf-44da1a5e421a

### Checklist


frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -113,6 +111,33 @@ export function DashCardCardParameterMapper({
...
 
+  let selectedMappingOption;
+  if (isVirtual || isAction || isNative) {
+    selectedMappingOption = target
+      ? mappingOptions.find(option =>
+          _.isEqual(normalize(option.target), normalize(target)),
+        )
+      : undefined;
+  } else {
+    const columns = Lib.visibleColumns(question.query(), -1);
```
uladzimirdev, line 122: do we need visible or filterable columns here? or doesn't matter? /cc @bshepherdson

bshepherdson: Short answer: yes, `visibleColumns` is the right thing here.

Long answer: `filterableColumns` returns a subset of the `visibleColumns`. Usually it's actually the same set, since every column is filterable in principle even if some cases are clumsy (eg. string `=` on a JSON column). The main reason to call `filterableColumns` is that it includes a bunch of extra information about what filter operators can be applied to each column. For the purposes of this matching, that extra work is wasted.

(The other `___able-columns` often return smaller subsets, since not every column is orderable, joinable, etc. They also include some extra information, eg. whether the column is already used in an `ORDER BY` on this query.)

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -113,6 +111,33 @@ export function DashCardCardParameterMapper({
...
+                question.query(),
+                stageIndex,
+                columns,
+                [normalize(target[1]), normalize(mappingOption.target[1])],
+              )
+            : [-1];
+
+          return index1 >= 0 && index1 === index2;
+        })
+      : undefined;
```
kulyk, line 138: There's a lot of stuff going on here, and it could be hard to read, IMO. We're checking if `target` is null-ish on line 127, but we already checked that on line 125. We're also doing a similar check for `isVirtual || isAction || isNative` block above; maybe we could do an early return and simplify both of these blocks?

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
```diff
@@ -63,77 +63,105 @@ describe("DashCardParameterMapper", () => {
...
+    it("should render an informative error state for link cards", () => {
+      const linkCard = createMockCard({ dataset_query: {}, display: "link" });
+      setup({
+        card: linkCard,
+        dashcard: createMockDashboardCard({
+          visualization_settings: {
+            virtual_card: linkCard,
+          },
+        }),
+      });
```
kulyk, line 88: We have a mock function for every kind of virtual dashcards, I think it'd better if we use them instead
For example, [here is `createMockLinkDashboardCard`](https://github.com/metabase/metabase/blob/19334990ce3af9381d197ea8de31314e852e57aa/frontend/src/metabase-types/api/mocks/dashboard.ts#L154)

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
```diff
@@ -63,77 +63,105 @@ describe("DashCardParameterMapper", () => {
...
-      screen.getByLabelText(/cannot connect variables to link cards/i),
-    ).toBeInTheDocument();
-  });
 
-  it("should render an informative parameter mapping state for text cards without variables", () => {
-    const textCard = createMockTextDashboardCard({ size_x: 3, size_y: 3 });
-    setup({
-      dashcard: textCard,
+    it("should render an informative parameter mapping state for text cards without variables", () => {
+      const textCard = createMockTextDashboardCard({ size_x: 3, size_y: 3 });
```
kulyk, line 96: Curios if `size_x` and `size_y` affect the tested behavior?

kulyk (approved):
Nice work! Left a couple of suggestions, please take a look before merging

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
```diff
@@ -113,6 +111,33 @@ export function DashCardCardParameterMapper({
...
+        )
+      : undefined;
+  } else {
+    const stageIndex = -1;
+    const columns = Lib.visibleColumns(question.query(), stageIndex);
+
+    selectedMappingOption = target
+      ? mappingOptions.find(mappingOption => {
+          const [index1, index2] = target
+            ? Lib.findColumnIndexesFromLegacyRefs(
```
ranquild, line 128: It seems we have O(n^2) here because for every `mappingOption` we call `findColumnIndexesFromLegacyRefs` that iterates on each `column`. Can it be improved? Maybe something like this:

```
const mappingColumnIndexes = Lib.findColumnIndexesFromLegacyRefs(
  query,
  stageIndex,
  columns,
  mappingOptions.map(({ target }) => target[1]), 
);

const [selectedColumnIndex] = Lib.findColumnIndexesFromLegacyRefs(
  query,
  stageIndex,
  columns,
  [target[1]],
);

if (selectedColumnIndex >= 0) {
  const mappingIndex = mappingColumnIndexes.indexOf(selectedColumnIndex);
  if (mappingIndex >= 0) {
    return mappingOptions[mappingIndex];
  }
}

return undefined;
```

ranquild (commented):
@uladzimirdev looks good, left some comments to improve performance and readability

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
uladzimirdev: only `size_y` matters

uladzimirdev: removed size_x [f1a9811](https://github.com/metabase/metabase/pull/38545/commits/f1a981120db6760400349b5138332785843a1568)

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.unit.spec.jsx
uladzimirdev: fixed [f1a9811](https://github.com/metabase/metabase/pull/38545/commits/f1a981120db6760400349b5138332785843a1568)

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
uladzimirdev: simplified function and moved it to utils [480e5e3](https://github.com/metabase/metabase/pull/38545/commits/480e5e3220108fac3861e2ae2f47a2319f9c9839)

frontend/src/metabase/dashboard/components/DashCard/utils.ts
```diff
@@ -16,3 +32,67 @@ export function shouldShowParameterMapper({
     !(display && VIZ_WITH_CUSTOM_MAPPING_UI.includes(display))
   );
 }
+
+type ParameterDimensionTarget = ["dimension", LocalFieldReference];
```
uladzimirdev, line 36: @ranquild @kulyk 

we already have the type ParameterDimensionTarget like

```
export type ParameterDimensionTarget = [
  "dimension",
  ["field", FieldId, ReferenceOptions | null] | ["template-tag", string],
];
```

am I right that we expected only the field here?

uladzimirdev: nvm, provided a fix here [92b16cb](https://github.com/metabase/metabase/pull/38545/commits/92b16cb5d5a682344b04e121b89b2c6a4a3616af)

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.jsx
uladzimirdev: fixed here [480e5e3](https://github.com/metabase/metabase/pull/38545/commits/480e5e3220108fac3861e2ae2f47a2319f9c9839)

kulyk (approved):
Looks good!

ranquild (approved):

frontend/src/metabase/dashboard/components/DashCard/utils.ts
```diff
@@ -16,3 +32,65 @@ export function shouldShowParameterMapper({
...
+  // target not found - no need to look further
+  if (columnByTargetIndex === -1) {
+    return;
+  }
+
+  const mappingColumnIndexes = Lib.findColumnIndexesFromLegacyRefs(
+    question.query(),
+    stageIndex,
+    columns,
+    mappingOptions.map(({ target }) => target[1]),
```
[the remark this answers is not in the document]
uladzimirdev: yes 😄

PR #38548, opened 2024-02-08
Avoid unnecessary re-renderings in because of getQuestions selector

uladzimirdev:
### Description

`getParameters` selector depends on `getQuestions`, which returns a new object for every update.

deep equality selector solves the problem with `getParameters` unnecessary re-calculations and re-renderings of components which use `getParameters` with `useSelector`.

[stress test run](https://github.com/metabase/metabase/actions/runs/7830252277)

### How to verify

tests should pass

### Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -19133,6 +19133,11 @@ reselect@^4.1.7:
   resolved "https://registry.yarnpkg.com/reselect/-/reselect-4.1.7.tgz#56480d9ff3d3188970ee2b76527bd94a95567a42"
   integrity sha512-Zu1xbUt3/OPwsXL46hvOOoQrap2azE7ZQbokq61BQfiXvhewsKDwhMeZjTX9sX0nvw1t/U5Audyn1I9P/m9z0A==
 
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -19133,6 +19133,11 @@ reselect@^4.1.7:
   resolved "https://registry.yarnpkg.com/reselect/-/reselect-4.1.7.tgz#56480d9ff3d3188970ee2b76527bd94a95567a42"
   integrity sha512-Zu1xbUt3/OPwsXL46hvOOoQrap2azE7ZQbokq61BQfiXvhewsKDwhMeZjTX9sX0nvw1t/U5Audyn1I9P/m9z0A==
 
+reselect@^5.1.0:
```
npretto, line 19136: reselect is already a dependency of rtk, do you think it would make sense to try and use the same version?

uladzimirdev: Version 4 doesn't export lru cache

kulyk (approved):

Issue #38509, opened 2024-02-07
All tables in mysql databases are marked as non-active and cannot browse them or use table metadata


jcorrea-tkww:
**Issue**
With the newest version v.0.48.4 the metadata sync that we have scheduled to run every day at midnight is marking all of the tables in mysql databases (specific engines only) as non-active causing users cannot explore (browse data) and cannot fetch table metadata in Admin section or use table fields for field filters, etc.

We have different mysql databases, some are not affected, like the Metabase database itself or other ones running 5.x or 8.x versions
However those that are affected are running "Server version: 8.0.26-16 Percona Server (GPL), Release 16, Revision 3d64165" 

We tried to trigger the table metadata fetch manually and the results are the same.
We tried to update the table metabase_table and set active = 1 to all of the tables listed.  After this manual update (SQL update) we can browse the tables, etc.  However, the next time the auto metadata fetch runs, all tables are again marked as non active.

**To Reproduce**
1. Go to Browse data in the frontend and select to browse a mysql database (running the affected engine) .  There is not any table to explore.
![image](https://github.com/metabase/metabase/assets/52322349/c163bf30-c76d-4ed1-b688-97390df7f6fc)

3. Go to Admin->Table Metadata and select one of the mysql affected databases.  Thre is not any schema on it
![image](https://github.com/metabase/metabase/assets/52322349/01ec7a9f-97db-4c74-8191-a1ef392cddbd)

5. Go to your mysql client and query the Metabase database and run the following query
`select
     md.id
    , md.name
    , md.`engine`
    , sum(active) active_tables
    , sum(1) total_tables
from
    metabase_table mt
join metabase_database md on 
    md.id = mt.db_id
where
    true
    and md.`engine` = 'mysql'
group by 
    1,2, 3`
7. See that some mysql are not affected but the affecgted ones 100% of the tables are marked as inactive
![image](https://github.com/metabase/metabase/assets/52322349/fc226951-c86a-4e80-99c3-7cf3e2978d9f)

**Expected behavior**
After a scheduled or manual metadata fetch, the active tables (and columns) should keep "active" so users with permissions can browse those tables
Admins can administer the metadata of those mysql databases as regular
Editors can use any table-field in their query builders, field filters, etc

**Severity**
this is quite severe as disables the feature of browse data, build questions, add field filters, administer metadata, etc

**Additional context**
it seems that the mysql version affected is 
Server version: 8.0.26-16 Percona Server (GPL), Release 16, Revision 3d64165

[...]

PR #38549, opened 2024-02-08
Partially revert 'Remove N+1 in sync'

crisptrutski:
Refences:
- https://github.com/metabase/metabase/issues/38509
- https://github.com/metabase/metabase/issues/38499

### Description

This PR reverts some components of https://github.com/metabase/metabase/pull/37439/

1. We rollback the breaking changes `metabase.driver.sql-jdbc.sync.interface/active-tables` which added a new argument. We will need to maintain backwards compatibility for 3 releases, so will need to introduce a new multi-method instead.
2. We remove the select privileges check from those methods, which required the breaking change, and which we believe is responsible for the two issues linked above.
3. We disable `:table-privileges`for Redshift for now, but leave the queries.

### DIFF
--- a/src/metabase/driver/sql_jdbc/sync/interface.clj
+++ b/src/metabase/driver/sql_jdbc/sync/interface.clj
@@ -9,9 +9,7 @@
 
   Two different implementations are provided in this namespace: `fast-active-tables` (the default), and
   `post-filtered-active-tables`. You should be fine using the default, but refer to the documentation for those
-  functions for more details on the differences.
-
-  `metabase` is an instance of `DatabaseMetaData`."
+  functions for more details on the differences."
   {:added "0.37.1"
    :arglists '([driver
                 ^java.sql.Connection connection
--- a/src/metabase/driver/sql_jdbc/sync/interface.clj
+++ b/src/metabase/driver/sql_jdbc/sync/interface.clj
@@ -9,10 +9,11 @@
 
   Two different implementations are provided in this namespace: `fast-active-tables` (the default), and
   `post-filtered-active-tables`. You should be fine using the default, but refer to the documentation for those
-  functions for more details on the differences."
+  functions for more details on the differences.
+
+  `metabase` is an instance of `DatabaseMetaData`."
   {:added "0.37.1"
    :arglists '([driver
-                database
                 ^java.sql.Connection connection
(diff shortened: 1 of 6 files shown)


src/metabase/sync/sync_metadata/sync_table_privileges.clj
```diff
@@ -15,8 +15,7 @@
   [database :- (mi/InstanceOf :model/Database)]
   (let [driver (driver.u/database->driver database)]
     (when (and (not= :redshift driver)
```
calherries, line 17: You can remove the redshift exclusion now and the comment below
```suggestion
    (when (driver/database-supports? driver :table-privileges database))
```

calherries, line 17: You can remove the redshift exclusion now and the comment below, so this is just
```clojure
    (when (driver/database-supports? driver :table-privileges database) ...)
```

src/metabase/driver/sql_jdbc/sync/interface.clj
```diff
@@ -9,10 +9,11 @@
 
   Two different implementations are provided in this namespace: `fast-active-tables` (the default), and
   `post-filtered-active-tables`. You should be fine using the default, but refer to the documentation for those
-  functions for more details on the differences."
+  functions for more details on the differences.
+
+  `metabase` is an instance of `DatabaseMetaData`."
```
calherries, line 14: I think "metabase" was meant to be "database", but it's gone now anyway
```suggestion
  functions for more details on the differences."
```

qnkhuat:
Are we sure we want to revert this for PG and redshift too? see my [comment](https://metaboat.slack.com/archives/C010L1Z4F9S/p1707393770504909?thread_ts=1707313038.155569&cid=C010L1Z4F9S)

src/metabase/sync/sync_metadata/sync_table_privileges.clj
crisptrutski: I intentionally updated the comment and left this so we didn't unintentionally enabled it again if we roll the other changes forward again.

crisptrutski:
We have decided against this approach, as it would revert much needed Redshift sync optimization, and we are not aware of any defects in its table-privileges implementation.

PR #38551, opened 2024-02-08
[QC] Remove unused methods from the `StructuredQuery` prototype

nemanjaglumac:
These methods were completely unused.

### DIFF
--- a/frontend/src/metabase-lib/queries/StructuredQuery.ts
+++ b/frontend/src/metabase-lib/queries/StructuredQuery.ts
@@ -126,14 +126,6 @@ class StructuredQuery extends AtomicQuery {
 
   /* Query superclass methods */
 
-  /**
-   * @returns true if we have metadata for the root source table loaded
-   */
-  hasMetadata(): boolean {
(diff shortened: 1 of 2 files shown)


kamilmielnik (approved):

frontend/src/metabase-lib/queries/StructuredQuery.ts
```diff
@@ -737,10 +707,6 @@ class StructuredQuery extends AtomicQuery {
     return this._updateQuery(Q.removeField, arguments);
   }
 
-  clearFields() {
```
kamilmielnik, line 740: We can remove it in `Join` too

uladzimirdev (approved):

frontend/src/metabase-lib/queries/StructuredQuery.ts
nemanjaglumac: Great catch.
I think I found another thread related to that in the `Join` class, so let me follow up in a separate PR.

PR #38553, opened 2024-02-08
Add ESLint rule for Metabase strings

WiNloSt:
This is a follow up from #34920, similar to the ESLint rule we introduce for the epic #36966

### Description

This ESLint rule ensures we won't forget to treat user-facing `Metabase` string properly. When users are whitelabeling, any user facing strings that are expected to be seen by non-admins should be replaced with the `application-name` setting instead.

### How to verify

When you reference `Metabase` string. This ESLint rule should start complaining. I exclude some files from this ESLint rules and hope that makes sense.

### Checklist


.eslintrc
```diff
@@ -161,13 +161,23 @@
...
       "files": ["*.js", "*.jsx", "*.ts", "*.tsx"],
       "rules": {
-        "no-unconditional-metabase-links-render": "error"
+        "no-unconditional-metabase-links-render": "error",
+        "no-literal-metabase-strings": "error"
       }
     },
     {
-      "files": ["*.unit.spec.*", "frontend/src/metabase/admin/**/*"],
+      "files": [
```
WiNloSt, line 169: I added more stuff here that should not be linted.

frontend/src/metabase/setup/components/PreferencesStep/PreferencesStep.tsx
```diff
@@ -73,7 +73,6 @@ export const PreferencesStep = (): JSX.Element => {
       <StepDescription>
         {t`In order to help us improve Metabase, we'd like to collect certain data about product usage.`}{" "}
         <ExternalLink
-          // eslint-disable-next-line no-unconditional-metabase-links-render -- Metabase setup
```
WiNloSt, line 76: This is now excluded in the `.eslintrc`

frontend/lint/tests/no-unconditional-metabase-links-render.unit.spec.js
```diff
@@ -131,6 +139,15 @@ const docsLink = useSelector(state =>
     code: `
 function MyComponent() {
   return <a href="https://metabase.com/docs/latest/troubleshooting-guide/bugs.html">Troubleshooting</a>;
+}`,
+    error:
+      /Metabase links must be rendered conditionally\.(.|\n)*Please import `getShowMetabaseLinks`(.|\n)*Or add comment to indicate the reason why this rule needs to be disabled/,
+  },
+  {
+    name: 'Detect "metabase.com/docs" in template strings',
```
WiNloSt, line 147: add missing tests.

frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js
```diff
@@ -172,7 +172,10 @@ module.exports = {
       TemplateLiteral(node) {
         const quasis = node.quasis;
         quasis.forEach(quasi => {
-          if (LITERAL_METABASE_URL_REGEX.exec(quasi.value.raw)) {
+          if (
+            LITERAL_METABASE_URL_REGEX.exec(quasi.value.raw) &&
+            !isGetShowMetabaseLinksSelectorImported
```
WiNloSt, line 177: Fixes this rule not working in template literals. I also add relevant tests.

frontend/lint/eslint-rules/no-literal-metabase-strings.js
WiNloSt, line 1: This rule is very similar to `frontend/lint/eslint-rules/no-unconditional-metabase-links-render.js`, but it's simpler.

We detect if we have `"Metabase"` in any strings, and show that we could fix the error by either importing `getApplicationName` or disabling the rule with a comment. Because there would be some places where we don't want to use `getApplicationName` e.g. in admin settings.

frontend/lint/tests/no-literal-metabase-strings.unit.spec.js
```diff
@@ -0,0 +1,149 @@
...
+  },
+  {
+    // "No Metabase string",
+    code: `
+  const label = "some string"`,
+  },
+  {
+    // "Detect in literal strings",
+    code: `
+  import { getApplicationName } from 'metabase/selectors/whitelabel';
```
npretto, line 31: Why should we consider this as valid?
It allows this:
```tsx
import { t } from "ttag";
import { useSelector } from "metabase/lib/redux";
import { getApplicationName } from "metabase/selectors/whitelabel";

export function SomeComponent() {
  const appName = useSelector(getApplicationName);
  return (
    <div>
      <h1>{"Hello Metabase - hardcoded"}</h1>
      <h1>Metabase</h1>
      <h1>{t`Metabase`}</h1>
      <Something label="Metabase" />
      <Something label={t`Metabase`} />
      <h2>{appName}</h2>
    </div>
  );
}

const Something = ({ label }: { label: string }) => {
  return <h2>{label}</h2>;
};
```
where we us the applicationName once but can also hardcode it in other places

WiNloSt: It's a compromise for the rule to not get too complicated. Surely we can try to track where an imported name is being used, but that would definitely require more time to implement.

This simplifies the implementation so that we trust if users import the selector they will use it correctly.

At least this should catch the majority of the problem where people just forget to import `getApplicationName` from the start. Actually that won't help when you're adding a new string in the same file that already has the selector imported though.

I didn't want to spend too much time on this so I think this is good enough.

npretto: I'm still not understanding why we check if that import is present, isn't the name returned by that selector supposed to replace the hardcoded "Metabase" string?
ie, if before it was
```tsx
const Component = () => {

  return <h1>t`Welcome to Metabase`<h1>
}
 ```
 
 it should be fixed as
 ```tsx
import { useSelector } from "metabase/lib/redux";
import { getApplicationName } from "metabase/selectors/whitelabel";

const Component = () => {
   const applicationName = useSelector(getApplicationName); 

  return <h1>t`Welcome to ${applicationName}`<h1>
}
 ```
which removes "Metabase" from the source code.

Shouldn't we be able to remove the import check entirely?

WiNloSt: Fixed in https://github.com/metabase/metabase/pull/38553/commits/9219375409a6a48697498ec1b8db7f852fd9cd07. This pretty much moves all the previous tests that are supposed to fail from `VALID_CASES` to `INVALID_CASES`

npretto (approved):
Tested and it works 👍 
Not an expert of eslint specific code but I know it is similar to the already approved PR for the links

PR #38547, opened 2024-02-08
Fix card query endpoint usage on dashboards

kulyk:
Fixes #38245 

Fixes an issue around dashcard data fetching with two main scenarios:

1. Dashboard with filters and cards > edit > duplicate tab > all cards mapped to a parameter error out
2. Dashboard with filters > edit > add card > map to filter > change tab > go back > new card errors out

[Switching to another tab refetches questions on the tab](https://github.com/metabase/metabase/blob/43c1895afa40833e97e7cea7113588449c869301/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx#L371). New dashboard cards don't have an ID yet, so [we're using a `POST /api/card/:id/query` for them](https://github.com/metabase/metabase/blob/c141463eadc9a33cdbdb90421268414a18ed15d9/frontend/src/metabase/dashboard/actions/data-fetching.js#L451). We're also passing the same [request body](https://github.com/metabase/metabase/blob/c141463eadc9a33cdbdb90421268414a18ed15d9/frontend/src/metabase/dashboard/actions/data-fetching.js#L460) we normally use for the daschard query endpoint (`POST /api/dashboard/:id/dashcard/:id/card/:id/query`), including `parameters`. But `POST /api/card/:id/query` endpoint is using `parameters` only for the native query editor, so here it's erroring out.

Fixed by excluding irrelevant parameters from `POST /api/card/:id/query` request body. As a side note, it doesn't make sense to reload these cards while in edit mode, but it's not a simple change with our data fetching mechanism for now

### To verify

1. Open a dashboard with a few filters and dashcards
2. Enter the editing mode and duplicate a dashboard tab
3. Ensure nothing is erroring out

### Demo

**Before**

https://github.com/metabase/metabase/assets/17258145/a9e742bf-2ffe-4b58-84ae-6b1338d8ca1c

**After**

https://github.com/metabase/metabase/assets/17258145/51b0d5b2-1476-43ce-9822-09f6eead4266

Issue #38245, opened 2024-01-29
Unsaved dashcards with mapped params error when switching tabs


EmmadUsmani:
### Describe the bug

If you add a dashcard to a dashboard, map it to a parameter, switch to another tab, then switch back to the tab with that dashcard, the card will error due to a bad HTTP request.

### To Reproduce

1. Create a new dashboard
2. Create two tabs
3. Add a filter
4. Add a dashcard to tab 1
5. Link the dashcard to the filter
6. Switch to tab 2
7. Go back to tab 1

### Expected behavior

The dashcard should not have an error

### Information about your Metabase installation

```JSON
stats commit `fc53ec6`
```

### Severity

Moderately severe; it goes away after saving the dashboard, but it's very annoying to have to do that often while creating a dashboard

### Additional context

It seems like the `/api/card/<id>/query` endpoint cannot handle the negative dashcard id we use for unsaved cards.

https://github.com/metabase/metabase/assets/37751258/f61684e2-a688-4106-8aed-2e0c2b5f9a46

Curl
```
[...]

PR #38559, opened 2024-02-08
Fix dashcard queries refetching

kulyk:
Follow-up on #38547 (issue #38245)

The change felt too risky for v49, so I'm extracting it into its own PR that we can merge later. It's doing two things:

**First:** turns off dashcard data and metadata refetching in dashboard editing mode
I don't think we have any good reasons to reload data in editing mode at all

**Second:** explicitly separates request data objects for dashcard vs. card query endpoints

Most of the time dashcards use the dashcard query endpoint to fetch data:
`POST /api/dashboard/:id/dashcard/:id/card/:id/query`

But when we're adding a new question on the dashboard, there's no ID to use yet, so we're using `POST /api/card/:id/query` to get some data for the preview

We used to provide the same request body to both endpoints (e.g. `dashboardId`, `dashcardId`, `dashboard_id` and `parameters`). But they're not used by the card query endpoint (there's even a [warning in BE code about it](https://github.com/metabase/metabase/blob/179a5d22dfcfdb752ada062f501e1b7f3e6dbe07/src/metabase/api/card.clj#L635))

### DIFF
--- a/frontend/src/metabase/dashboard/actions/data-fetching.js
+++ b/frontend/src/metabase/dashboard/actions/data-fetching.js
@@ -447,26 +447,29 @@ export const fetchCardData = createThunkAction(
           dashcardBeforeEditing &&
           dashcardBeforeEditing.card_id !== dashcard.card_id;
 
-        // new dashcards and new additional series cards aren't yet saved to the dashboard, so they need to be run using the card query endpoint
-        const endpoint =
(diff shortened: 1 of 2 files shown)


kulyk:
Closing as the main change is ported to #38547
It looks like we can't completely disable data fetching in the editing mode. If we haven't loaded tab queries yet, that would block them from loading

PR #38562, opened 2024-02-08
Revert breaking change to the `active-tables` driver method

crisptrutski:
### Description

This reorganizes the code around `:table-privileges`, letting us revert a change that broke backwards compatibility.

It feels like there's a sharp edge here for JDBC drivers that implement (or have already) implemented the `metabase.driver` level method. Perhaps the methods should have different names? 🤷

### DIFF
--- a/docs/developers-guide/driver-changelog.md
+++ b/docs/developers-guide/driver-changelog.md
@@ -6,6 +6,8 @@ title: Driver interface changelog
 
 ## Metabase 0.49.0
 
+- The multimethod `metabase.driver.sql-jdbc.sync.interface/current-user-table-privileges` has been added. JDBC-based drivers can implement this to improve the performance of the default SQL JDBC implementation of `metabase.driver/describe-database`. It needs to be implemented if the database supports the `:table-privileges` feature and the driver is JDBC-based.
+
 - The multimethod `metabase.driver/create-table!` can take an additional optional map with an optional key `primary-key`.
   `metabase.driver/upload-type->database-type` must also be changed, so that if
   `:metabase.upload/auto-incrementing-int-pk` is provided as the `upload-type` argument, the function should return a
@@ -26,16 +28,11 @@ title: Driver interface changelog
   `metabase.driver.sql.query-processor/honey-sql-version` is now deprecated and no longer called. All drivers are
   assumed to use Honey SQL 2.
 
-- The method `metabase.driver.sql-jdbc.sync.interface/active-tables` that we added in 47 has been updated to require
-  an additional argument: `database`.
-  The new function arglist is `[driver database connection schema-inclusion-filters schema-exclusion-filters]`.
-
 - The method `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type` is now deprecated.
   Use `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type-and-value` instead,
   which has access to `value` and therefore provides more flexibility for choosing the right conversion unit.
 
 ## Metabase 0.48.0
-
 - The MBQL schema in `metabase.mbql.schema` now uses [Malli](https://github.com/metosin/malli) instead of
   [Schema](https://github.com/plumatic/schema). If you were using this namespace in combination with Schema, you'll
   want to update your code to use Malli instead.
--- a/docs/developers-guide/driver-changelog.md
+++ b/docs/developers-guide/driver-changelog.md
@@ -6,7 +6,10 @@ title: Driver interface changelog
 
 ## Metabase 0.49.0
 
-- The multimethod `metabase.driver.sql-jdbc.sync.interface/current-user-table-privileges` has been added. JDBC-based drivers can implement this to improve the performance of the default SQL JDBC implementation of `metabase.driver/describe-database`. It needs to be implemented if the database supports the `:table-privileges` feature and the driver is JDBC-based.
+- The multimethod `metabase.driver.sql-jdbc.sync.interface/current-user-table-privileges` has been added.
+  JDBC-based drivers can implement this to improve the performance of the default SQL JDBC implementation of
+  `metabase.driver/describe-database`. It needs to be implemented if the database supports the `:table-privileges`
+  feature and the driver is JDBC-based.
 
 - The multimethod `metabase.driver/create-table!` can take an additional optional map with an optional key `primary-key`.
   `metabase.driver/upload-type->database-type` must also be changed, so that if
--- a/modules/drivers/redshift/src/metabase/driver/redshift.clj
+++ b/modules/drivers/redshift/src/metabase/driver/redshift.clj
@@ -12,8 +12,7 @@
    [metabase.driver.sql-jdbc.execute :as sql-jdbc.execute]
    [metabase.driver.sql-jdbc.execute.legacy-impl :as sql-jdbc.legacy]
    [metabase.driver.sql-jdbc.sync :as sql-jdbc.sync]
-   [metabase.driver.sql-jdbc.sync.describe-table
-    :as sql-jdbc.describe-table]
+   [metabase.driver.sql-jdbc.sync.describe-table :as sql-jdbc.describe-table]
    [metabase.driver.sql.query-processor :as sql.qp]
    [metabase.lib.metadata :as lib.metadata]
    [metabase.mbql.util :as mbql.u]
@@ -399,14 +398,13 @@
   [driver db-id table-name column-names values]
   ((get-method driver/insert-into! :sql-jdbc) driver db-id table-name column-names values))
 
-(defmethod driver/current-user-table-privileges :redshift
-  [_driver database]
-  (let [conn-spec (sql-jdbc.conn/db->pooled-connection-spec database)]
-    ;; KNOWN LIMITATION: this won't return privileges for external tables, calling has_table_privilege on an external table
-    ;; result in an operation not supported error
-    (->> (jdbc/query
-          conn-spec
-          (str/join
+(defmethod sql-jdbc.sync/current-user-table-privileges :redshift
+  [_driver conn-spec & {:as _options}]
+  ;; KNOWN LIMITATION: this won't return privileges for external tables, calling has_table_privilege on an external table
+  ;; result in an operation not supported error
+  (->> (jdbc/query
+         conn-spec
+         (str/join
            "\n"
            ["with table_privileges as ("
             " select"
@@ -428,7 +426,7 @@
             ")"
             "select t.*"
             "from table_privileges t"]))
-         (filter #(or (:select %) (:update %) (:delete %) (:update %))))))
+         (filter #(or (:select %) (:update %) (:delete %) (:update %)))))
 
 
 ;;; ----------------------------------------------- Connection Impersonation ------------------------------------------
--- a/src/metabase/driver/mysql.clj
+++ b/src/metabase/driver/mysql.clj
@@ -86,6 +86,11 @@
   [database]
   (-> database :dbms_version :flavor (= "MariaDB")))
 
+(defn mariadb-connection?
+  "Returns true if the database is MariaDB."
+  [driver conn]
+  (->> conn (sql-jdbc.sync/dbms-version driver) :flavor (= "MariaDB")))
+
 (defmethod driver/database-supports? [:mysql :table-privileges]
   [_driver _feat _db]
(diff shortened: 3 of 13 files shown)


src/metabase/driver/sql_jdbc/sync/interface.clj
```diff
@@ -103,3 +102,25 @@
...
+     - table           :- :string
+     - select          :- :boolean
+     - update          :- :boolean
+     - insert          :- :boolean
+     - delete          :- :boolean
+
+   Either:
+   (1) role is null, corresponding to the privileges of the DB connection's current user
+   (2) role is not null, corresponding to the privileges of the role"
+  {:added "0.38.0" :arglists '([driver ^java.sql.Connection connection])}
```
calherries, line 124:
```suggestion
  {:added "0.48.0" :arglists '([driver ^java.sql.Connection connection])}
```

docs/developers-guide/driver-changelog.md
```diff
@@ -26,16 +26,16 @@ title: Driver interface changelog
...
 - The method `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type` is now deprecated.
   Use `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type-and-value` instead,
   which has access to `value` and therefore provides more flexibility for choosing the right conversion unit.
 
 ## Metabase 0.48.0
 
+- The multimethod `metabase.driver.sql-jdbc.sync.interface/current-user-table-privileges` has been added. All JDBC-
+  based drivers should implement this instead of the corresponding `metabase.driver` multimethod if they want to
+  leverage the built-in JDBC implementation of `metabase.driver/describe-database`.
+
```
calherries, line 38: This doesn't need to be backported, so this will be in 49.

src/metabase/driver/mysql.clj
```diff
@@ -874,19 +880,16 @@
...
   ;; MariaDB doesn't allow users to query the privileges of roles a user might have (unless they have select privileges
   ;; for the mysql database), so we can't query the full privileges of the current user.
-  (when-not (mariadb? database)
-    (let [conn-spec   (sql-jdbc.conn/db->pooled-connection-spec database)
-          db-name     (or (get-in database [:details :db])
-                          ;; some tests are stil using dbname
-                          (get-in database [:details :dbname]))
-          table-names (->> (jdbc/query conn-spec "SHOW TABLES" {:as-arrays? true})
+  (when-not (mariadb-connection? driver conn)
+    (let [db-name     (ffirst (jdbc/query conn "SELECT DATABASE()"))
```
calherries, line 888: works for me

docs/developers-guide/driver-changelog.md
crisptrutski: My thinking was as follows:

We introduced the `metabase.driver` in 48, but after this change there's a difference in how SQL drivers should be implemented. I wanted to backport it so that the same driver implementation will work across both of the latest releases.

For now I'll just skip the backport, but let me know why you disagree with this reasoning. Is it because 3rd party SQL drivers are rare, this feature is still obscure, or because we don't need to worry about this kind of cross release compatibility for a given driver version?

src/metabase/driver/postgres.clj
```diff
@@ -847,38 +848,37 @@
...
-            "   select schemaname, matviewname as objectname from pg_catalog.pg_matviews"
-            " ) t"
-            " where t.schemaname !~ '^pg_'"
-            "   and t.schemaname <> 'information_schema'"
-            "   and pg_catalog.has_schema_privilege(current_user, t.schemaname, 'USAGE')"
-            ")"
-            "select t.*"
-            "from table_privileges t"]))
-         (filter #(or (:select %) (:update %) (:delete %) (:update %))))))
+(defmethod sql-jdbc.sync.interface/current-user-table-privileges :postgres
```
calherries, line 851: `sql-jdbc.sync.interface`, despite the name, is actually part of `metabase.driver.sql-jdbc.sync`, which at the moment is the real sync interface for sql-jdbc drivers. The multimethods from `sql-jdbc.sync.interface` get imported there with `import-vars`. The explanation for why this oddity happens is [here](https://metaboat.slack.com/archives/CKZEMT1MJ/p1700471405194629). At the moment I think it pays to be consistent and implement the imported version from `metabase.driver.sql-jdbc.sync`.

So you should require `[metabase.driver.sql-jdbc.sync :as sql-jdbc.sync]` in this namespace, and rename this to:

```suggestion
(defmethod sql-jdbc.sync/current-user-table-privileges :postgres
```

You should also move this defmethod to the appropriate place in the file, where other defmethods for metabase.driver.sql-jdbc.sync are.

docs/developers-guide/driver-changelog.md
```diff
@@ -26,16 +30,11 @@ title: Driver interface changelog
   `metabase.driver.sql.query-processor/honey-sql-version` is now deprecated and no longer called. All drivers are
   assumed to use Honey SQL 2.
 
-- The method `metabase.driver.sql-jdbc.sync.interface/active-tables` that we added in 47 has been updated to require
-  an additional argument: `database`.
-  The new function arglist is `[driver database connection schema-inclusion-filters schema-exclusion-filters]`.
-
```
crisptrutski, line 32: I'm not sure if deleting this is the right thing to do, or whether we should leave it on record that we vacillated on this change.

src/metabase/driver/mysql.clj
```diff
@@ -874,19 +880,16 @@
                     [table-name privileges])))
           table-names)))
 
-(defmethod driver/current-user-table-privileges :mysql
-  [_driver database]
+(defmethod sql-jdbc.sync.interface/current-user-table-privileges :mysql
```
calherries, line 883: See my similar comment for postgres
```suggestion
(defmethod sql-jdbc.sync/current-user-table-privileges :mysql
```

docs/developers-guide/driver-changelog.md
calherries: Deleting it is fine imo, nobody should be depending on unreleased changelogs because they're subject to change before the release.

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -393,33 +393,32 @@
...
-            "   select schemaname, viewname as objectname from pg_views"
-            " ) t"
-            " where t.schemaname !~ '^pg_'"
-            "   and t.schemaname <> 'information_schema'"
-            "   and pg_catalog.has_schema_privilege(current_user, t.schemaname, 'USAGE')"
-            ")"
-            "select t.*"
-            "from table_privileges t"]))
-         (filter #(or (:select %) (:update %) (:delete %) (:update %))))))
+(defmethod sql-jdbc.sync.interface/current-user-table-privileges :redshift
```
calherries, line 396: See my similar comment for postgres 
https://github.com/metabase/metabase/pull/38562/files#r1483957565

```suggestion
(defmethod sql-jdbc.sync/current-user-table-privileges :redshift
```

src/metabase/driver/postgres.clj
crisptrutski: Aha, something had felt off with so few direct references to the namespace.

docs/developers-guide/driver-changelog.md
```diff
@@ -6,6 +6,10 @@ title: Driver interface changelog
 
 ## Metabase 0.49.0
 
+- The multimethod `metabase.driver.sql-jdbc.sync.interface/current-user-table-privileges` has been added. All JDBC-
+  based drivers should implement this instead of the corresponding `metabase.driver` multimethod if they want to
+  leverage the built-in JDBC implementation of `metabase.driver/describe-database`.
```
calherries, line 11: we should stress that implementing it is optional

```suggestion
- The multimethod `metabase.driver.sql-jdbc.sync.interface/current-user-table-privileges` has been added. JDBC-based drivers can implement this to improve the performance of the default SQL JDBC implementation of `metabase.driver/describe-database`. It needs to be implemented if the database supports the `:table-privileges` feature and the driver is JDBC-based.
```

src/metabase/driver/sql_jdbc/sync/interface.clj
```diff
@@ -103,3 +102,25 @@
...
+     - table           :- :string
+     - select          :- :boolean
+     - update          :- :boolean
+     - insert          :- :boolean
+     - delete          :- :boolean
+
+   Either:
+   (1) role is null, corresponding to the privileges of the DB connection's current user
+   (2) role is not null, corresponding to the privileges of the role"
+  {:added "0.49.0" :arglists '([driver ^java.sql.Connection connection])}
```
calherries, line 124: Lets leave room for kwargs
```suggestion
  {:added "0.49.0" :arglists '([driver ^java.sql.Connection connection & args])}
```

src/metabase/driver/mysql.clj
```diff
@@ -874,19 +880,16 @@
...
+  [driver conn]
```
calherries: Let's leave these implementations open to extension so we can do things like only getting select privileges
```suggestion
  [driver conn & {:as _args}]
```

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -115,8 +115,8 @@
                           :type        (.getString rset "TABLE_TYPE")}))))))
 
 (defn- schema+table-with-select-privileges
-  [driver database]
-  (->> (driver/current-user-table-privileges driver database)
+  [driver conn]
+  (->> (sql-jdbc.sync.interface/current-user-table-privileges driver conn)
```
calherries, line 119: I'm wondering whether we should call this with kwargs, so implementations can choose whether or not to support filtering by specific privileges.
```suggestion
  (->> (sql-jdbc.sync.interface/current-user-table-privileges driver conn :privileges #{:select}})
```

docs/developers-guide/driver-changelog.md
calherries: > Is it because 3rd party SQL drivers are rare, this feature is still obscure, or because we don't need to worry about this kind of cross release compatibility for a given driver version?

All of the above. We backported it without changing the changelog because this feature is still obscure there's good reason to assume no-one implemented the table-privileges feature unless they were crazy. We probably shouldn't have but that ship has sailed. And we don't need to bother backporting it because I suspect there will be conflicts and there's just not much of a reason to bother. Plus I find it confusing to add things to the backlog for a minor release, because there shouldn't be any changes to it after a major release under normal circumstances.

src/metabase/driver/sql_jdbc/sync/interface.clj
crisptrutski: I is intentional not to use `& {:as options}` here, in case we decide to add optional positional arguments too, or is this just a general style preference? I see a few `:arglists` in the codebase that explicitly show it as kwargs and even mention some keys.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
crisptrutski: Inclined to YAGNI on this, even if we decide to make this extension we might change our mind about the structure of the kwargs to use.

calherries: Fair point, let's leave it

test/metabase/driver/mysql_test.clj
```diff
@@ -710,9 +710,7 @@
                                    (sql-jdbc.conn/with-connection-spec-for-testing-connection
                                      [spec [:mysql new-connection-details]]
                                      (with-redefs [sql-jdbc.conn/db->pooled-connection-spec (fn [_] spec)]
-                                       (driver/current-user-table-privileges driver/*driver*
-                                                                             (assoc (mt/db) :name "test table privileges db"
-                                                                                    :details new-connection-details))))))]
+                                       (sql-jdbc.sync/current-user-table-privileges driver/*driver* spec)))))]
```
calherries, line 713: wdyt about a test that checks that we can call implementations with kwargs correctly. This might suffice
```suggestion
                                       (sql-jdbc.sync/current-user-table-privileges driver/*driver* spec {})))))]
```

calherries (approved):
Awesome. Our 3rd party driver authors and their users will be happy. You have some CI tests failing with unresolved symbols though

test/metabase/driver/mysql_test.clj
crisptrutski: Good idea

calherries (approved):
changes look good

PR #38567, opened 2024-02-08
Redshift support for CSV uploads (take 3)

calherries:
https://github.com/metabase/metabase/pull/37505 landed 2 days ago and is causing havoc with our redshift tests in CI. That PR caused new tables to be created in the redshift cloud database every test run that never get cleaned up.

This PR restores redshift support for uploads and changes the upload tests so that all the tables created by the CSV upload tests are cleaned up immediately afterwards.

Pro-tip for reviewers on Github: use this link to see the diff excluding whitespace https://github.com/metabase/metabase/pull/38567/files?w=1

[Slack thread](https://metaboat.slack.com/archives/C5XHN8GLW/p1707329164099649)


test/metabase/upload_test.clj
```diff
@@ -506,16 +514,17 @@
 (deftest load-from-csv-table-name-test
   (testing "Can upload two files with the same name"
     (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
-      (mt/with-empty-db
```
calherries, line 509: The main culprit is `mt/with-empty-db`, which doesn't do anything for redshift but sync the test database.

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -698,7 +698,8 @@
   (perms/revoke-application-permissions! (perms-group/all-users) :setting))
 
 (deftest upload-csv-test
-  (mt/test-drivers (mt/normal-drivers-with-feature :uploads :schemas)
+  (mt/test-drivers (disj (mt/normal-drivers-with-feature :uploads :schemas)
+                         :redshift) ; Remove redshift for now because writing this test is complicated with session schemas
```
calherries, line 702: This test doesn't rely on any driver-specific behaviour

test/metabase/upload_test.clj
```diff
@@ -506,16 +514,17 @@
...
-          (let [csv-file-prefix (mt/random-name)
-                model-1 (upload-example-csv! :csv-file-prefix csv-file-prefix)
-                model-2 (upload-example-csv! :csv-file-prefix csv-file-prefix)]
-            (testing "tables are different between the two uploads"
-              (is (some? (:table_id model-1)))
-              (is (some? (:table_id model-2)))
-              (is (not= (:table_id model-1)
-                        (:table_id model-2))))))))))
+      (let [csv-file-prefix "some file prefix"]
+        (with-upload-table!
```
calherries, line 518: Wrapping the creation of each table in `with-upload-table!` cleans up after the created table, making `mt/with-empty-db` unnecessary

calherries:
looking into CI failures

test/metabase/driver/sql_jdbc_test.clj
```diff
@@ -239,26 +239,32 @@
...
-                                                             :details (-> (mt/db)
-                                                                          :details
-                                                                          (assoc filter-type-prop "exclusion"
-                                                                                 patterns-type-prop "public"))}]
-                (is (= #{fake-schema-name}
-                       (driver/syncable-schemas driver/*driver* db-filtered)))))))))))
+        (is (contains? (driver/syncable-schemas driver/*driver* (mt/db))
+                       "public"))
+        (is (contains? (driver/syncable-schemas driver/*driver* (mt/db))
+                       fake-schema-name)))
```
calherries, line 245: I had to change this test, because it could pick up the "not_public" schema being left around by the other tests on postgres, mysql, and H2.

bshepherdson (dismissed):

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -700,59 +700,70 @@
...
-                  (is (some? (upload-csv!)))
-                  (is (thrown-with-msg?
-                        clojure.lang.ExceptionInfo
-                        #"You don't have permissions to do that\."
-                        (upload-csv!)))))
-              (with-all-users-data-perms! {db-id {:data {:native :write, :schemas ["not_public"]}}}
-                (is (some? (upload-csv!)))))))))))
+      (mt/with-model-cleanup [:model/Table]
+        (let [table-name (mt/random-name)
+              schema-name (or (sql.tx/session-schema driver/*driver*) "public")
```
calherries, line 705: Here's an example of how the change works.

Before we used the "not_public" schema and created tables in there, which wouldn't get cleaned up for redshift.

Now we use the result of sql.tx/session-schema for redshift, which is the `unique-session-schema` defined [here](https://github.com/metabase/metabase/blob/7dd097fd7728e879b1fe562d47c4252350aa2d47/modules/drivers/redshift/test/metabase/test/data/redshift.clj#L61)

"public" is used for postgres, mysql, and H2. The tests don't clean up the schema afterwards, only the tables.

crisptrutski (dismissed):

calherries:
@metabase/core-backend-components I'm putting this back in review again since the last reviews were done in haste to fix CI. I'm looking for a serious review this time

test/metabase/upload_test.clj
```diff
@@ -1230,10 +1231,18 @@
...
-                        :file     file}))))
+        (mt/with-model-cleanup [:model/Table]
+          (let [new-table (when (nil? table-id)
+                            (create-upload-table!))
+                table-id (or table-id (:id new-table))]
+            (t2/update! :model/Table table-id {:is_upload is-upload})
+            (try (append-csv! {:table-id table-id
+                               :file     file})
+                 (finally
+                   (when new-table
```
crisptrutski, line 1242: Do not be fooled by the length of this comment, it's just a nit.

On an initial reading this was not clear to me why the clean up code only existed for one case, and started writing a comment here. The penny dropped before I posted, but I think it's a natural question mark for a reader, and short comment about how the cases are handled differently would be nice. Perhaps something like "mt/with-model-cleanup retracts inserted rows, but we must manually retract new tables". 

In general it would be nice to use the `with` macro pattern consistently, which would also group all the clean-up together. It would be nice to cover file deletion this way too.
  
This is very much gilding the lily, given the narrow scope of this helper, but since I already started writing a comment, I think it's worth sharing the guidance. This is yet another example of the "single level of abstraction" principle I'm so fond of - `finally` is a lower level of abstraction than `with`, so they don't belong in the same function.

If we want to avoid proliferating macros, the idiom in JVM languages without macros is the "try with resources" pattern, i.e. `with-open`, combined with wrappers for files, tables etc that implement`AutoCloseable`. We could probably use a protocol for values with `:close-fn` metadata, since I think that macro just uses duck typing. What's nice about that pattern is we'd get leverage through the whole code base with just a wrapper per type, and we can indent just one no matter how many resources we are cleaning up.

test/metabase/upload_test.clj
```diff
@@ -1230,10 +1231,18 @@
...
+                     (driver/drop-table! driver/*driver*
+                                         (mt/id)
+                                         (#'upload/table-identifier new-table)))))))))
     (finally
       (io/delete-file file))))
```
crisptrutski, line 1247: This is a preexisting smell, but it's a footgun that the method will delete the file even if it did not create it.

test/metabase/upload_test.clj
```diff
@@ -1022,89 +1026,87 @@
...
+          (is (nil? (:schema table))))))))
 
 (deftest create-csv-upload!-auto-pk-column-display-name-test
   (testing "The auto-generated column display_name should be the same as its name"
    (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
-     (mt/with-empty-db
-       (let [new-model (upload-example-csv!)
-             new-field (t2/select-one Field :table_id (:table_id new-model) :name "_mb_row_id")]
+     (with-upload-table!
+       [table (card->table (upload-example-csv!))]
```
crisptrutski, line 1098: nit: typically we keep bindings for `with` macros up on the first line.

test/metabase/upload_test.clj
crisptrutski: We should probably put a warning in the `mt/with-empty-db` docstring about this footgun for redshift, and give the guidance to use something table-level.

test/metabase/upload_test.clj
```diff
@@ -483,14 +480,20 @@
 (defmacro with-uploads-allowed [& body]
   `(do-with-uploads-allowed (fn [] ~@body)))
 
-(defn- do-with-upload-table! [table thunk]
+(defn do-with-upload-table! [table thunk]
   (try (thunk table)
        (finally
          (driver/drop-table! driver/*driver*
                              (mt/id)
```
crisptrutski, line 487: Can we make this helper more generic by getting the database id from the table?

test/metabase/upload_test.clj
```diff
@@ -1022,89 +1026,87 @@
...
+                                                                  :is_full_sync false})]
+      (try
+        (with-redefs [ ;; do away with the `future` invocation since we don't want race conditions in a test
+                      future-call (fn [thunk]
+                                    (swap! in-future? (constantly true))
+                                    (thunk))]
+          (testing "Happy path with schema, and without table-prefix"
+            ;; make sure the schema exists
+            (jdbc/execute! (sql-jdbc.conn/connection-details->spec driver/*driver* (:details db))
+                           (format "CREATE SCHEMA IF NOT EXISTS %s;" (sql.tx/qualify-and-quote driver/*driver* schema-name)))
```
crisptrutski, line 1044: This is a fairly dense expression, and we repeat it in this file, so I think a helper method is warranted.

crisptrutski (approved):
I really like how this encapsulated a lot of the finicky details and unified the two branches in `append-csv-data-perms-test`. I left a bunch of nits, but nothing blocking.

enterprise/backend/test/metabase_enterprise/upload_test.clj
```diff
@@ -10,23 +10,24 @@
     (met/with-gtaps-for-user :rasta {:gtaps {:venues {}}}
       (is (thrown-with-msg? Exception #"Uploads are not permitted for sandboxed users\."
             (upload-test/upload-example-csv! {:grant-permission? false
-                                              :schema-name       "not_public"
                                               :table-prefix      "uploaded_magic_"}))))))
 
 (deftest appends-disabled-for-sandboxed-user-test
-  (mt/test-drivers (mt/normal-drivers-with-feature :uploads)
+  ;; FIXME: Redshift is flaking on `mt/dataset` and I don't know why, so I'm excluding it temporarily
```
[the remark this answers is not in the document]
calherries: I discovered the `mt/dataset` wasn't even necessary in the first place.  Fixed by 742cacf37447e3d633b6e39476996d9b1730482 and f9061cbd71bb2dfc83f32e86adbe05c5cf48737a

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -700,88 +698,61 @@
...
+        (upload-test/with-upload-table!
+          [table (upload-test/create-upload-table!)]
+          (let [db-id       (mt/id)
                 upload-csv! (fn []
                               (upload-test/upload-example-csv! {:grant-permission? false
-                                                                :schema-name       "not_public"
+                                                                :schema-name       (:schema table)
                                                                 :table-prefix      "uploaded_magic_"}))]
-            (doseq [[schema-perms can-upload?] {:all            true
-                                                :none           false
```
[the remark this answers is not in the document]
calherries: Added the case back. This was a mistake, thanks for catching! 957243a4739b21c4c368714f1cbcc6bf434300a3

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -700,88 +698,61 @@
...
-                (if can-upload?
-                  (is (some? (upload-csv!)))
-                  (is (thrown-with-msg?
-                        clojure.lang.ExceptionInfo
-                        #"You don't have permissions to do that\."
-                        (upload-csv!)))))
-              (with-all-users-data-perms! {db-id {:data {:native :write, :schemas ["not_public"]}}}
+            (doseq [[schema-perms can-upload? description]
+                    [[:all            true  "Data permissions on schema should succeed"]
+                     [{(:id table) :all} false "Data permissions on table should fail"]]]
```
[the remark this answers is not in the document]
calherries: e6bb7bc3b206e31b45cdd0257e469b3fda8115ed

test/metabase/upload_test.clj
```diff
@@ -1249,36 +1258,35 @@
...
-      (testing "Even if the uploads database, schema and table prefix is not set, appends succeed"
-        (mt/with-temporary-setting-values [uploads-database-id nil
-                                           uploads-schema-name nil
-                                           uploads-table-prefix nil]
-          (is (some? (append-csv-with-defaults!)))))
-      (testing "Uploads must be enabled"
-        (is (= {:message "Uploads are not enabled."
+    (testing "Happy path"
+      (is (= {:row-count 2}
+             (append-csv-with-defaults!))))
```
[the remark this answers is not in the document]
calherries: de7d392b7385249619b69df35912bd04d9b91cef

test/metabase/upload_test.clj
calherries: fixed for this case. d2344a9967c09a2bdb6e6407434fef58ea7e598e

I'm going to leave other cases such as [here](https://github.com/metabase/metabase/blob/d2344a9967c09a2bdb6e6407434fef58ea7e598e/test/metabase/upload_test.clj#L1046) where following this convention would make the code more unreadable to me.

There are many exceptions to this e.g. `mt/with-temp` where we add a new line after the macro symbol to avoid creeping to the right.

test/metabase/upload_test.clj
calherries: 6e0cabc9ecce5cfc82255e82c484959eb8426e70

calherries: New docstring: 
"Sets the current dataset to a freshly created db that gets destroyed at the conclusion of `body`.
   Use this to test destructive actions that may modify the data.
   WARNING: this doesn't actually create and destroy a temporary database for cloud databases (like redshift) that
   reuse a single database for all tests."

test/metabase/driver/sql_jdbc_test.clj
```diff
@@ -239,26 +239,32 @@
...
-              (t2.with-temp/with-temp [Database db-filtered {:engine  driver
-                                                             :details (-> (mt/db)
-                                                                          :details
-                                                                          (assoc filter-type-prop "exclusion"
-                                                                                 patterns-type-prop "public"))}]
-                (is (= #{fake-schema-name}
-                       (driver/syncable-schemas driver/*driver* db-filtered)))))))))))
+        (is (contains? (driver/syncable-schemas driver/*driver* (mt/db))
+                       "public"))
+        (is (contains? (driver/syncable-schemas driver/*driver* (mt/db))
```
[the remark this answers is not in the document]
calherries: These are all good points, and I would have used your solution but I tried it and I think I found a solution I find more
readable, that addresses most of the problems you had originally.

I updated this in 7a4b151be246efae0aea780400be12aebcb5098e

The main benefit is there's fewer names introduced like `relevant-schemas` which as you mentioned weren't great at
conveying meaning beyond the test's specific implementation.

test/metabase/upload_test.clj
calherries: We don't repeat it in this file. Anyway I've split it into two expressions with a let now, so hopefully it's less dense to read c4e70a700c230f9d7b8779a0c96ba34b35df52e3

test/metabase/driver/sql_jdbc_test.clj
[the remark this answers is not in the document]
crisptrutski: Looks good. The set comparison is nice when you also want to catch that unintended things aren't present, but agree it's a bit awkward in this case.

test/metabase/upload_test.clj
crisptrutski: Yeah those exceptions make sense

test/metabase/upload_test.clj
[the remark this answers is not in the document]
calherries: de7d392b7385249619b69df35912bd04d9b91cef

test/metabase/upload_test.clj
crisptrutski: Definitely looks better. I didn't check whether both cases were in the same file. Avoiding cross test dependencies and keeping tests as self contained as possible makes.

crisptrutski:
Nice amendments, LGTM

test/metabase/upload_test.clj
calherries: yep 34f5d99a1b54935d8ee509ebb09e132cab2e053f

test/metabase/upload_test.clj
calherries: Thanks for the detailed explanation! 

The reason why the `with-upload-table!` macro isn't being used here is because the table needs to be conditionally created depending on whether `table-id` is supplied as an argument. I can't think of an elegant way to get around that right now, but you might have an idea.

I'm going to make a mental note to come back to this technique in the future anyway. I haven't seen any implementations of AutoCloseable at metabase before.

test/metabase/upload_test.clj
calherries: I still can't find it in other files, but anyway good point for previous versions of these tests, I should have done that.

calherries:
Removed the milestone because I disabled redshift support again in https://github.com/metabase/metabase/pull/38776

PR #38574, opened 2024-02-08
Milestone 3B: Fix dashboard tabs and filters order on a public dashboard page

kulyk:
Fixes that public dashboards used to display filters before tabs

`PublicDashboard` renders the `EmbedFrame` component. The `EmbedFrame` renders the filters panel internally, and we're rendering dashboard tabs and a dashboard itself inside the `EmbedFrame` body.

IMO, the right way to approach this is to get rid of the `EmbedFrame` — it can get too limiting in terms of page layout
It can also make changes more risky as it's shared between actions, dashboards, and questions.

As a short-term fix before 49, this PR is adding a `header` prop to `EmbedFrame` that'd let us render tabs between the dashboard's title and filters.

### Demo

<details>
<summary>Fixed width dashboard</summary>

**Before**

fixed-width-before

**After**

fixed-width-after

</details>

<details>
<summary>Full width dashboard</summary>

**Before**

full-width-before

**After**

full-width-after

</details>

---
**Context**
- Will complete second half of Milestone 3: https://github.com/metabase/metabase/issues/36358


cdeweyx (approved):

frontend/src/metabase/public/containers/PublicDashboard.jsx
```diff
@@ -178,6 +177,7 @@ class PublicDashboard extends Component {
         actionButtons={
           buttons.length > 0 && <div className="flex">{buttons}</div>
         }
+        header={<StyledDashboardTabs location={this.props.location} />}
```
JesseSDevaney, line 180: nit: We use a bunch different things to make the header. Would prefer a different name here, so we know exactly what we are receiving in this. 
```suggestion
        dashboardTabs={<StyledDashboardTabs location={this.props.location} />}
```

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
```diff
@@ -88,6 +92,7 @@ function mapStateToProps(state: State) {
 function EmbedFrame({
   className,
   children,
+  header,
```
JesseSDevaney, line 95: nit: We use a bunch different things to make the header. Would prefer a different name here, so we know exactly what we are receiving in this.
```suggestion
  dashboardTabs,
```

JesseSDevaney (approved):
LGTM! 

Just a couple naming nits (even though we want to clean this up in the future anyway)

frontend/src/metabase/public/containers/PublicDashboard.jsx
WiNloSt: Since this is a short-term and very specific fix. I agree we should make this clear what it's suppose to represent.

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
```diff
@@ -147,12 +152,15 @@ function EmbedFrame({
...
-              />
+              <TitleAndDescriptionContainer>
+                <TitleAndDescription
+                  title={finalName}
+                  description={description}
+                  className="my2"
+                />
+              </TitleAndDescriptionContainer>
             )}
+            {isValidElement(header) ? header : null}
```
WiNloSt, line 163: not sure if this is a bit overkill. Since we're controlling what's being passed `StyledDashboardTabs` in this case. Would defaulting the `header` or (`dashboardTabs` as per Jesse's suggestion) to `null` and render simply render this better?

e.g.

```suggestion
            {dashboardTabs}
```

And in the component props
```ts
...
children,
dashboardTabs = null,
name,
...
```

WiNloSt (approved):

WiNloSt:
I have a visual feedback difference between the actual dashboard and public/static embed when there's no parameter.

### App
#### with parameters
image
#### without parameters
image

### Embed
#### with parameters
image
#### without parameters
image


There is an invisible parameters that's taking space on embedding. Maybe you want to fix that?
image

PR #38586, opened 2024-02-08
Make table `can-read?`/`can-write?` use `DataPermissions`

johnswanson:


enterprise/backend/src/metabase_enterprise/sandbox/api/table.clj
```diff
@@ -28,14 +27,15 @@
                            [:= :pgm.user_id (u/the-id user-or-user-id)]]}))
 
 (mu/defn only-sandboxed-perms? :- :boolean
-  "Returns true if the user has only segemented and not full table permissions. If the user has full table permissions
-  we wouldn't want to apply this segment filtering."
+  "Returns true if the user has sandboxed permissions. If a sandbox policy exists, it overrides existing permission on
+  the table."
```
noahmoss, line 31: There's a small distinction here between a sandboxing policy existing, and sandboxing actually being enforced for the user. Sandboxing is only enforced if the user isn't in any _other_ group which grants full access to the table.

It looks like the original function was checking whether sandboxing is enforced, while the new one is just checking whether any sandboxing policy exists.

I'm not sure if this distinction matters much for this namespace but it would definitely matter for the usages in `metabase-enterprise.sandbox.models.params.field-values` since it computes a different hash key for cached field values based on whether sandboxing is enforced or not.

To match the previous behavior you'd need to use `metabase-enterprise.sandbox.api.util/enforced-sandboxes`, or at least do something similar here.

enterprise/backend/src/metabase_enterprise/sandbox/api/table.clj
```diff
@@ -63,21 +63,18 @@
...
     ;; metadata
     (if sandboxed-perms?
-      (binding [api/*current-user-permissions-set* (atom
-                                                    (set/union
-                                                     @api/*current-user-permissions-set*
-                                                     (mi/perms-objects-set table :read)))]
-        (thunk))
+      (maybe-filter-fields
+       table
+       (data-perms/with-additional-table-permission :perms/data-access (:db_id table) (u/the-id table) :unrestricted
```
noahmoss, line 76: Nice, this is super useful

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
```diff
@@ -789,17 +795,12 @@
...
-            (testing test-string
-              (with-all-users-data-perms! {db-id {:data perms}}
-                (if can-append?
-                  (is (some? (append-csv!)))
-                  (is (thrown-with-msg?
-                       clojure.lang.ExceptionInfo
-                       #"You don't have permissions to do that\."
-                       (append-csv!))))))))))))
+          (testing "With blocked perms it should fail"
+            (with-all-users-data-perms! {db-id {:data {:native :none,  :schemas :block}}}
```
noahmoss, line 799: nit
```suggestion
            (with-all-users-data-perms! {db-id {:data {:native :none, :schemas :block}}}
```

noahmoss (changes requested):
Looks great! Just think there's one small correctness point about the sandboxing change.

enterprise/backend/src/metabase_enterprise/sandbox/api/table.clj
johnswanson: Thanks for noticing this and calling it out. I think I mentioned this in slack, but I had been thinking that our move from "sandbox is its own permission" to "sandbox is unrestricted perm + a GTAP" meant that we didn't need to check for unrestricted perms anymore, but as you pointed out, that's wrong - we need to check for unrestricted perms from _other_ groups.

I ended up modifying the `sandbox.api.util` namespace to add an `enforced-sandboxes-for` function. It takes a user ID and returns the sandboxes that will be enforced for that user. Then this function checks to see whether those enforced sandboxes include one for this table.

enterprise/backend/test/metabase_enterprise/advanced_permissions/common_test.clj
johnswanson: fixed in a rebase, thanks!

enterprise/backend/src/metabase_enterprise/sandbox/api/table.clj
noahmoss: Cool, makes sense!

noahmoss (approved):

PR #38587, opened 2024-02-08
[MLv2] Prescribe the order for `metabase.lib.column-group/group-columns`

### LINKED ISSUE #38329: When adding a filter to a Model with implicit joined tables, we should first show fields from the Model and then the implicit joined tables alphabetically
### Describe the bug

image

When adding a filter to a Model like [Account](https://stats.metabase.com/model/4523-account/notebook), we are showing fields in some random order.

We should show:
1. The source Model fields. It should open the filtering component showing the source Model fields and collapsing implicit joined tables.
2. Explicit joined tables (or Models). These shiould be collapsed and right after the source Model (alphabetically or in the join order, what is easier)
3. Implicit joined Tables (and Models in the future): in alphabetical order

### To Reproduce

1. Go to [Accounts](https://stats.metabase.com/model/4523-account/notebook) model
2. Try to add a filter
3. See Instance fields as the first option

### Expected behavior

_No response_

### Logs

_No response_

```JSON
Master
```

### Severity

P2

### Additional context

_No response_

bshepherdson:
The desired order is: own columns, explicitly joined, implicitly joined.
Explicit joins are ordered among themselves by join alias. Implicit
joins are ordered by the join alias of the FK (own columns first), then
by the name of the FK column.

Previously this order has usually been at least close to the desired
order by accident, since it's close to the iteration order of
`visible-columns`. However, `group-columns` uses `group-by`, which
returns a map. If there are enough groups (more than 8 I think) the map
implementation will switch from a `[key1 value1 key2 value2 ...]` array
to a real map, and the iteration order is undefined.

This change enforces the desired order, including a test that `shuffle`s
the columns. (And fixing the expectations for a few other tests.)

Fixes #38329.

### DIFF
--- a/test/metabase/lib/column_group_test.cljc
+++ b/test/metabase/lib/column_group_test.cljc
@@ -349,10 +349,10 @@
                 {:name field-name
                  :lib/desired-column-alias (str "Orders__" field-name)
                  :lib/source :source/joins})}
-             (implicit user-cols    "PEOPLE__via__USER_ID__" "")
              (implicit product-cols "PRODUCTS__via__PRODUCT_ID__" "")
-             (implicit user-cols    "PEOPLE__via__USER_ID__" "_2")
-             (implicit product-cols "PRODUCTS__via__PRODUCT_ID__" "_2")]
+             (implicit user-cols    "PEOPLE__via__USER_ID__" "")
+             (implicit product-cols "PRODUCTS__via__PRODUCT_ID__" "_2")
+             (implicit user-cols    "PEOPLE__via__USER_ID__" "_2")]
             (lib/group-columns marked)))))
 
 (deftest ^:parallel self-joined-cards-duplicate-implicit-columns-test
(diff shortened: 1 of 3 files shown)


bshepherdson:
* `master`
  * **PR #38587** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38587?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38587?utm_source=stack-comment).

test/metabase/lib/column_group_test.cljc
```diff
@@ -413,3 +413,43 @@
...
+               ;; - "Reviews - Product.PRODUCT_ID"
+               {::lib.column-group/group-type :group-type/join.implicit
+                :fk-join-alias                nil
+                :fk-field-id                  (meta/id :orders :product-id)}
+               {::lib.column-group/group-type :group-type/join.implicit
+                :fk-join-alias                nil
+                :fk-field-id                  (meta/id :orders :user-id)}
+               {::lib.column-group/group-type :group-type/join.implicit
+                :fk-join-alias                "Orders"
+                :fk-field-id                  (meta/id :orders :product-id)}
```
crisptrutski, line 448: You missed this entry in the comment above

crisptrutski (approved):

test/metabase/lib/column_group_test.cljc
bshepherdson: Ah, it was included in one line, but that was left over from when the comment was phrased differently and grouped the joins rather than listing each one.

Fixed.

bshepherdson:
Adding `backport` now since 49 has been cut. This is a small fix to a pretty bad UX, so it should get into 49.

PR #38589, opened 2024-02-08
Remove dead code in frontend/src/metabase/admin

sloansparger:
### Description

Ran `npx ts-prune` and verified / removed dead code in the `frontend/src/metabase/admin` folder.
Some code is also exported that didn't needed to be, so some export statements are removed.


iethree (approved):
💖

PR #38593, opened 2024-02-09
Remove flaky timeout in DelayGroup test

romeovs:
This PR removes a flaky test case that is breaking a lot of builds in CI.

All that's being tested is the fact that the timeout was reached, and a greater than assertion does the job just fine.

[Flaky test run seems to confirm this fix works](
https://github.com/metabase/metabase/actions/runs/7842196150/job/21400027071).


uladzimirdev (approved):

uladzimirdev (changes requested):
Reject before the stress test results

uladzimirdev (approved):

PR #38600, opened 2024-02-09
Scope find-root-test-failure! to same ns by default

crisptrutski:
### Description
 
Thanks for this awesome utility!

I had a situation where I knew the offender(s) were in the same test suite, so I've added a scope option to narrow things down. I made it the default too, since it's much faster.


johnswanson (approved):
LGTM!

PR #38601, opened 2024-02-09
Fix database pollution in setting tests

crisptrutski:
### Description

I tracked down which tests were polluting the database, and there about 5 of them. In order to fix them I needed to turn off initialization around 15 spots where we bulk retrieved settings values. In the end I decided it was easier to just opt into initialization for the few tests that cover it, this also protects better against future-flakes.

### DIFF
--- a/test/metabase/models/setting_test.clj
+++ b/test/metabase/models/setting_test.clj
@@ -21,6 +21,8 @@
 
 (use-fixtures :once (fixtures/initialize :db))
 
+(use-fixtures :each #(binding [setting/*disable-init* true] (%)))
+
 ;; ## TEST SETTINGS DEFINITIONS
 
 (defsetting test-setting-1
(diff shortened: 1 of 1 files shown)


calherries (commented):
I take issue with this - an important property of settings is that `*disable-init*` var defaults to `false` unless explicitly bound. I don't see any tests here that would break if we flipped it from `false` to `true`.

crisptrutski:
> I take issue with this - an important property of settings is that `*disable-init*` var defaults to `false` unless explicitly bound. I don't see any tests here that would break if we flipped it from `false` to `true`.

That's a good point, and I'm not sure how to address it with this approach. Do you mind looking at the first commit on this branch, where I add opt-outs instead, and let me know if you're happy with that instead?

Otherwise I could try find a way for the init tests to skip the fixture instead of reversing it.

Yet another idea I had is having a separate dynamic var for whether we should persist settings to the db, or just populate the cache. Not that keen on adding more hacks though.

crisptrutski:
Hmm, the *skip-persistence* route would create a similar issue, although more subtle. 

It's critical for a uuid-nonce that the value is persisted, so we'd need at least one test that (by default) persists it, and checks that it's persisted. We don't actually have such a persistence test as it stands as far as I know, but if we did we'd still need some way to opt out of the fixture. Looking at clojure.test there doesn't seem to be any way to examine the test var inside the fixture function.

crisptrutski:
We decided to abandon this approach in favor of a fixture hook that tracks and reverts the corresponding database inserts.

PR #38604, opened 2024-02-09
Use server side generated timestamp for relative datetime computation on Redshift v2

### LINKED ISSUE #21518: Compute results of timestamp/current-date functions during query processing

A customer reported that as the GETDATE function is evaluated at query time, so it'll evade Redshift caching. This is documented on [Redshift performance tips](https://docs.aws.amazon.com/redshift/latest/dg/c_challenges_achieving_high_performance_queries.html).

Calculate as much as possible at the app level rather than at the DB level in Redshift

None, unless you want to use text as filters for dates

Very important for Redshift users

**Additional context**

Reported by one of our customers
We should fix this for all databases, so we can close #20988 too.

lbrdnk:
Closes #21518

### Description

This PR is re-instantiation of PR #35995. Server side timestamp generation (instead of computing values database side using `getdate` function) enables Redshift caching for specific queries. This PR makes `:relative-datetime` values with unit `:day` or greater generated server side. More details can be found in original PR.

UTC night failures that started to occur after merging first attempt are resolved now. Only modifications made to the original code are updated comments and removal of caching test, as caching is best effort and independent of Metabase.

### How to verify

Run the test suite with newly added tests.

### Checklist


modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -210,6 +212,38 @@
         y (h2x/->timestamp y)]
     (sql.qp/datetime-diff driver unit x y)))
 
+(def server-side-relative-datetime-units
```
metamben, line 215: Does it have to be public?

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -210,6 +212,38 @@
...
+  "Units which, when used with :relative-datetime clause, cause computation server side. Those are
+   [[metabase.util.date-2/add-units]] coarser or equal to day. For more details see
+   [[server-side-relative-datetime-honeysql-form]]."
+  #{:day :week :month :quarter :year})
+
+(defn- use-server-side-relative-datetime?
+  [unit]
+  (contains? server-side-relative-datetime-units unit))
```
metamben, line 223: This function doesn't seem particularly useful.

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -480,3 +484,96 @@
...
+;;;; Server side generated timestamps for :relative-datetime tests follow.
+
+(defn- run-native-query [sql & params]
+  (-> (mt/native-query {:query sql
+                        :params params})
+      qp/process-query))
+
+(defn- getdate-vs-ss-ts-test-thunk-generator
+  ([]
+   (getdate-vs-ss-ts-test-thunk-generator :week -1))
```
metamben, line 497: This is not a very intuitive default, I'm not sure it's worth defining it.

metamben (approved):
LGTM

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
lbrdnk: That's right. I've removed zero arity and added arguments to calls where it was used. Thanks.

modules/drivers/redshift/src/metabase/driver/redshift.clj
lbrdnk: I've decided to move definition of `server-side-relative-datetime-units` (from [your previous comment](https://github.com/metabase/metabase/pull/38604#discussion_r1484401169)) and its docstring into this function. This function is redefined in some tests to force the old behavior and compare the result with new one.

lbrdnk:

PR #38612, opened 2024-02-09
[Docs] Fix outdated UI references 

alexyarosh:
Fix outdated UI references in [Basic exploration](https://www.metabase.com/docs/latest/exploration-and-organization/exploration).

### DIFF
--- a/docs/exploration-and-organization/exploration.md
+++ b/docs/exploration-and-organization/exploration.md
@@ -82,12 +82,12 @@ Searches take into account items’ titles, descriptions, and other metadata —
 
 **Bookmarks** are a way to quickly get back to things you visit frequently (or have been working on recently). Bookmarked items show up in the main navigation sidebar above [collections][collections].
 
-To bookmark an item, look for the **ribbon** icon - it's usually in the upper right corner of the page for that item. You can bookmark:
+To bookmark an item, look for the **ribbon** icon in the upper right corner of the item's page. You can bookmark:
 
-- Questions 
-- Models 
-- Dashboards 
-- Collections 
+- Questions
+- Models
+- Dashboards
+- Collections
 
 When viewing a collection, you can also click on the `...` next to an item to add or remove a bookmark from the item.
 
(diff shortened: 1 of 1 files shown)


docs/exploration-and-organization/exploration.md
```diff
@@ -24,7 +24,7 @@ To learn more, see [Exploring data with Metabase's data browser](https://www.met
 
 ![A collection](./images/collection-detail.png)
 
-If your teammates are cool, they'll have pinned some important dashboards or questions within your collections; if so, those pinned items will show up as cards at the top of the collection.
+Your teammates might have pinned some important dashboards or questions within your collections; if so, those pinned items will show up as cards at the top of the collection.
```
alexyarosh, line 27: Rephrasing this because "If your teammates are cool" goes a bit too far into "shit-talking" territory imo

docs/exploration-and-organization/exploration.md
```diff
@@ -82,14 +82,14 @@ Searches take into account items’ titles, descriptions, and other metadata —
...
-- Models (ribbon's in the editing sidebar)
-- Dashboards (ribbon's in the **...** menu)
-- Collections (ribbon's in the upper right of the collection's page)
+- Questions
+- Models
+- Dashboards
+- Collections
 
-When viewing a collection, you can also click on the **...** next to an item to add or remove a bookmark from the item.
+When viewing a collection, you can also click on the `...` next to an item to add or remove a bookmark from the item.
```
jeff-bruemmer, line 92: Minor:

```suggestion
To add or remove a bookmark from an item, click on the three-dot menu (**...**) next to the item in the collection.
```
We've referred to this menu like so elsewhere, so might as well be consistent. Generally we bold UI elements people can click on.

And it's good to lead with what we're doing, then follow with how to do it, so people can stop reading if they don't care.

jeff-bruemmer (approved):
Thanks, @alexyarosh.

docs/exploration-and-organization/exploration.md
alexyarosh: Ah ok it was in `...` also elsewhere in the article. I liked it more in `...` because otherwise it looks like punctuation. But if that's consistent with the rest of the docs, then it's better to keep it bold.

Completely agree with leading with what we're doing!

dragonsahead (approved):

PR #38622, opened 2024-02-09
Analytics should load only when changed, via checksum

escherize:
Most of this was already merged and reviewed, we needed some more handling for dealing with files inside a jar (a historically tricky task). @johnswanson had the good idea to simply reuse our jar+file aware unzipping capabilities to unzip the IA content first, then continue to calculate the checksum from the file system.

The new pieces improve handling of hashing IA content when running from a jar, by copying IA content _out of the jar_ first, and then checking the checksums afterward.

We end up doing a few more fast operations (install internal metabase user, etc.) on every startup, but we skip the only slow one (actually loading IA content into the app through serialization) when the previously-computed checksum, and the newly-computed checksum are equal.


test/metabase/api/search_test.clj
```diff
@@ -3,7 +3,7 @@
    [clojure.set :as set]
    [clojure.string :as str]
    [clojure.test :refer :all]
-   [java-time :as t]
+   [java-time.api :as t]
```
johnswanson, line 6: Did this sneak in somehow?

johnswanson (approved):

test/metabase/api/search_test.clj
escherize: I think this is required since the java-time kondo alias got renamed.

escherize: I've gone ahead and make the changes elsewhere

PR #38624, opened 2024-02-09
source-field populated in aggregates with foreign refs

### LINKED ISSUE #38618: X-Ray failure with message "Cannot determine source table..."
To reproduce:

* X-Ray the Reviews table in the sample database 
* Note that "Product → Price over time" is broken
* Click the "Product → Price over time" to focus on the query to see the same behavior

See @albertoperdomo's related examples in [this Slack thread](https://metaboat.slack.com/archives/C01MS7DQKR6/p1707478852175869).

Visual repro:
xraybroke

When this happens, you will see a message like so in the Metabase logs:
```
 :error "Cannot determine the source table or query for Field clause [:field 59 {:base-type :type/Float}]",
```

Here's an example broken query:
```
(qp/process-query
  {:database 1,
   :query {:aggregation [["sum" ["field" 59 {:base-type "type/Float"}]] 
                                        ["avg" ["field" 59 {:base-type "type/Float"}]]],
           :breakout [["field" 69 {:base-type "type/DateTime", :temporal-unit "month"}]],
           :source-table 8},
   :type "query"})
2024-02-09 16:02:29,815 WARN middleware.fix-bad-references :: Bad :field clause [:field 59 {:base-type :type/Float}] for field "PRODUCTS.PRICE" at [:aggregation :sum]: clause should have a :join-alias. Unable to infer an appropriate join. Query may not work as expected.
2024-02-09 16:02:29,816 WARN middleware.fix-bad-references :: Bad :field clause [:field 59 {:base-type :type/Float}] for field "PRODUCTS.PRICE" at [:aggregation :avg]: clause should have a :join-alias. Unable to infer an appropriate join. Query may not work as expected.
Execution error (ExceptionInfo) at metabase.query-processor.util.add-alias-info/fn$&f (add_alias_info.clj:193).
Cannot determine the source table or query for Field clause [:field 59 {:base-type :type/Float}]
```

The issue seems to be that when the breakout is added to the query in the x-ray code, it does not add in a `:source-field` or other join information since field 59 (PRICE) is from the products table, but we're querying from table 8 (REVIEWS).

When manually creating the question, the right query is:
```
{:database 1,
 :type     "query",
 :query    {:source-table 8,
            :aggregation  [["sum" ["field" 59 {:base-type "type/Float", :source-field 71}]]
                           ["avg" ["field" 59 {:base-type "type/Float", :source-field 71}]]],
            :breakout     [["field" 69 {:base-type "type/DateTime", :temporal-unit "month"}]]}}
```
Field 71 is product id.

markbastian:
When aggregates are populated that reference other tables, either a `:source-field` must be provided in the field metadata or a `:join-alias` needs to be provided. The former is preferred.

Prior to this PR, the first situation in the below simplified case could happen:

```clojure
;;Simplified case
;; busted
(qp/process-query
  {:database 1,
   ;; 59 is "PRODUCTS" -> "PRICE" which is not in the "REVIEWS" table
   :query    {:aggregation  [["sum" ["field" 59 {:base-type "type/Float"}]]]
              ;; "REVIEWS"
              :source-table 8},
   :type     "query"})

; Works
(qp/process-query
  {:database 1,
   ;; The `:source-field` metadata fixes the above query
   :query    {:aggregation  [["sum" ["field" 59 {:base-type    "type/Float"
                                                 ;; "REVIEWS" -> "PRODUCT_ID"
                                                 :source-field 71}]]]
              :source-table 8},
   :type     "query"})
```

This PR updates `metabase.automagic-dashboards.interestin/ground-metric` to use the `->reference` function for fields when building the `decoder` for transforming field names to references, which properly adds sources for external fields.

Fixes #38618

### DIFF
--- a/src/metabase/automagic_dashboards/interesting.clj
+++ b/src/metabase/automagic_dashboards/interesting.clj
@@ -93,6 +93,88 @@
            (group-by :semantic_type))
       set)))
 
+(defmulti
+  ^{:doc      "Get a reference for a given model to be injected into a template
+          (either MBQL, native query, or string)."
+    :arglists '([template-type model])}
+  ->reference (fn [template-type model]
+                [template-type (mi/model model)]))
+
+(defn- optimal-datetime-resolution
+  [field]
+  (let [[earliest latest] (some->> field
+                                   :fingerprint
+                                   :type
+                                   :type/DateTime
+                                   ((juxt :earliest :latest))
+                                   (map u.date/parse))]
+    (if (and earliest latest)
+      ;; e.g. if 3 hours > [duration between earliest and latest] then use `:minute` resolution
+      (condp u.date/greater-than-period-duration? (u.date/period-duration earliest latest)
+        (t/hours 3) :minute
+        (t/days 7) :hour
+        (t/months 6) :day
+        (t/years 10) :month
+        :year)
+      :day)))
+
+(defmethod ->reference [:mbql Field]
+  [_ {:keys [fk_target_field_id id link aggregation name base_type] :as field}]
+  (let [reference (mbql.normalize/normalize
+                    (cond
+                      link [:field id {:source-field link}]
+                      fk_target_field_id [:field fk_target_field_id {:source-field id}]
+                      id [:field id nil]
+                      :else [:field name {:base-type base_type}]))]
+    (cond
+      (isa? base_type :type/Temporal)
+      (mbql.u/with-temporal-unit reference (keyword (or aggregation
+                                                        (optimal-datetime-resolution field))))
+
+      (and aggregation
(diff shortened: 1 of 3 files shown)


src/metabase/automagic_dashboards/interesting.clj
```diff
@@ -118,8 +200,7 @@
          (apply math.combo/cartesian-product)
          (map (partial zipmap named-dimensions))
          (map (fn [nm->field]
```
markbastian, line 202: This is the only substantial change. The `->reference` changes are just raising the code in the ns to be above this.

perivamsi (approved):

PR #38626, opened 2024-02-09
Force reload in `ensure-audit-db-installed!`

johnswanson:
Sometimes analytics bits get removed by other tests (e.g. deleting all cards, dashboards, users, etc.). When we run
`ensure-audit-db-installed!`, though, the checksum hasn't changed (since the underlying analytics files didn't change), so we skip reloading the data.

This sets `last-analytics-checksum` to `-1` (meaning: skip the checksum and just load the analytics data) when running
`ensure-audit-db-installed!` in the tests that we've seen flake.


escherize (approved):
LGTM

PR #38628, opened 2024-02-09
Make qp metadata errors more descriptive

### LINKED ISSUE #38381: QP error for Table from wrong DB is not as good as it used to be
Since we switched to using MLv2 metadata providers in the QP the error message for running a query with a Table from a different DB is now something like 

```
Failed to fetch :metadata/table 65
```

which is not super helpful for people outside the QP eng team. Let's replace the error message with something a little more helpful. I think the old error message was something like 

```
Failed to fetch Table 65: Table does not exist or belongs to a different Database
```

lbrdnk:
Closes #38381

### Description

This PR makes qp metadata error messages more descriptive. Eg. `Failed to fetch :metadata/table 65` becomes `Failed to fetch Table 65: Table does not exist or belongs to a different Database`. Code is written in a way, that if we come up with more explanatory messages also for `Metric`, `Segment` and `Card`, those could be added easily.

### How to verify

Run modified tests.

### Checklist


src/metabase/query_processor/store.clj
```diff
@@ -178,8 +178,23 @@
   [database-id-or-metadata-provider & body]
   `(do-with-metadata-provider ~database-id-or-metadata-provider (^:once fn* [] ~@body)))
 
+(defn- pretty-metadata-error-msg
```
metamben, line 181: This name suggests more than it can actually do. It can only report problems related to failed lookups.

src/metabase/query_processor/store.clj
```diff
@@ -178,8 +178,23 @@
...
+  "Pretty error message for eg. [[missing-bulk-metadata-error]]."
+  [metadata-type id]
+  (let [metadata-type-str (pr-str metadata-type)
+        id-str (pr-str id)]
+    (case metadata-type
+      :metadata/card    (tru "Failed to fetch Card {0}." id-str)
+      :metadata/column  (tru "Failed to fetch Field {0}: Field does not exist or belongs to different Database."
```
metamben, line 188: Switching from fields to columns might become a source of problems. Not all columns are fields.

metamben (approved):

camsaul (approved):

PR #38629, opened 2024-02-09
Don't lower dashboard/question title when there is no last edit info

### LINKED ISSUE #38566: Instance Analytics titles drop after page load
### Describe the bug

Dashboards (and questions) tend to hide last edited info behind the title, and will appear on hover. IA dashboards don't have any of that information though, which causes the title to start off in the right place, but then drop lower than it should. A quick hack could be to add a min height to the `<HeaderBadges>` component, but I think the better solution is to disable that animation when there is no content in Header Badges

IA Dashboard:
chrome_3t2DhEPaxr

Normal Dashboard: 
chrome_AQMU2Qdv3L

### To Reproduce

Go to any IA dashboard and observe the title

### Expected behavior

It should look right 

### Logs

_No response_

```JSON
Current Master
```

### Severity

Not overly severe, but it looks bad

### Additional context

_No response_

npfitz:
Closes https://github.com/metabase/metabase/issues/38566

### Description
When a Dashboards or Questions / Models doesn't have any "last edit info" to show (such as things in the Metabase Analytics folder), the Entity title will start off in the correct position, but 4 seconds later will animate down, putting it out of place. This PR cancels that transition if we detect that no sub headers will be rendered

### How to verify

1. Go to any model or dashboard in Metabase Analytics and load the page
2. Observe the title in the header. It should not move after 4 seconds has passed
3. Now navigate to any Dashboard/Question/Model in out analytics
4. When the page loads, you should briefly see last edited info, possibly data source info as well. after 4 seconds, the title should lower and the sub header should fade away

### Demo
chrome_ZUlIbnwauv

### Checklist
No tests were added here, as it's unclear what type of testing would be helpful

### DIFF
--- a/frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeaderView.tsx
+++ b/frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeaderView.tsx
@@ -91,10 +91,12 @@ export function DashboardHeaderComponent({
 
   useEffect(() => {
     const timerId = setTimeout(() => {
-      setShowSubHeader(false);
+      if (isLastEditInfoVisible) {
+        setShowSubHeader(false);
+      }
     }, 4000);
     return () => clearTimeout(timerId);
-  }, []);
+  }, [isLastEditInfoVisible]);
 
   return (
     <div>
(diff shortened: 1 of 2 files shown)


sloansparger (approved):
lgtm

PR #38645, opened 2024-02-12
Type routing in State

kamilmielnik:
### Description

Our `State` type does not have `routing` key, even though it's often there.
This PR adds it and updates related mocks.

### DIFF
--- a/frontend/src/metabase-types/store/state.ts
+++ b/frontend/src/metabase-types/store/state.ts
@@ -1,3 +1,5 @@
+import type { RouterState } from "react-router-redux";
+
 import type { User } from "metabase-types/api";
 import type { AdminState } from "./admin";
 import type { AppState } from "./app";
@@ -21,8 +23,9 @@ export interface State {
   embed: EmbedState;
   entities: EntitiesState;
   metabot: MetabotState;
-  qb: QueryBuilderState;
   parameters: ParametersState;
+  qb: QueryBuilderState;
+  routing: RouterState;
   settings: SettingsState;
   setup: SetupState;
   upload: FileUploadState;
--- a/frontend/src/metabase/selectors/routing.ts
+++ b/frontend/src/metabase/selectors/routing.ts
@@ -1,6 +1,6 @@
 import type { State } from "metabase-types/store";
 
-export const getRouting = (state: State) => state.routing;
+const getRouting = (state: State) => state.routing;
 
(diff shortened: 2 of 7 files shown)


frontend/test/__support__/ui.tsx
```diff
@@ -72,6 +73,7 @@ export function renderWithProviders(
 
   if (withRouter) {
     Object.assign(reducers, { routing: routerReducer });
+    Object.assign(initialState, { routing });
```
kamilmielnik, line 76: `initialState` cannot have `routing` key present when `routingReducer` is not used - `createStore` from redux will spam the console with warning about an extraneous key.

frontend/src/metabase-types/store/state.ts
```diff
@@ -21,8 +23,9 @@ export interface State {
   embed: EmbedState;
   entities: EntitiesState;
   metabot: MetabotState;
-  qb: QueryBuilderState;
   parameters: ParametersState;
+  qb: QueryBuilderState;
+  routing: RouterState;
```
kamilmielnik, line 28: This is the primary change, all other are a consequence of this line

uladzimirdev (commented):
I'm not a fan of mocking redux state, I'd rather mock location itself on jsdom level, but it can be tricky

frontend/src/metabase/selectors/routing.ts
```diff
@@ -0,0 +1,6 @@
+import type { State } from "metabase-types/store";
+
+export const getRouting = (state: State) => state.routing;
```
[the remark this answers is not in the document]
kamilmielnik: d12b245b9ec635da94e60c82e8d4a4770a8fa530

frontend/src/metabase/selectors/routing.ts
```diff
@@ -0,0 +1,6 @@
+import type { State } from "metabase-types/store";
+
+const getRouting = (state: State) => state.routing;
+
+export const getLocation = (state: State) =>
+  getRouting(state).locationBeforeTransitions;
```
uladzimirdev, line 6: maybe this one is even cleaner

```suggestion
export const getLocation = (state: State) =>
  state.routing.locationBeforeTransitions;
```

ranquild (approved):

kamilmielnik:

PR #38647, opened 2024-02-12
Fix custom destination to public dashboard not working on static embedding

### LINKED ISSUE #38640: Custom Destination to Public Dashboard URLs is broken on Static Embeds
### Describe the bug

When you setup a Custom Destination to a public URL it breaks on static embeds with a `Not Found` when clicking on it.

### To Reproduce

To Reproduce

1) Turn on public sharing in admin settings
2) Go to New -> Question -> Sample DB -> Orders -> Save as Question 1
3) Go to New -> Question -> Sample DB -> Products -> Save as Question 2
4) Go to New -> Dashboard -> Add Question 1 -> Save as Dashboard and make public

image

5)  Go to New -> Dashboard -> Add Question 2 -> Add Click Behaviour  -> Save as Dashboard 2
---------------------------------------------> Click on Address 
---------------------------------------------> Go to Custom Destination 
--------------------------------------------->  URL
---------------------------------------------> Paste Public URL for Dashboard 1 

image

7) Enable static Embedding for Dashboard 2 -> On the preview Panel Click on Address -> You will get a Not Found

image

8) If you get the embed URL and paste the JWT link on the browser you will notice this:

image

Click on the Address Column and notice that the URL changes to the public one but you get a Not Found

image

Refreshing the browser works.

NOTE 
Directly from Metabase ->  Go to Dashboard 2 -> Click on Address -> Notice it works

### Expected behavior

Public link to Dashbaord opens from Static Embedding

### Logs

None that are relevant, even the Network Tab shows nothing only the image request

image

```JSON
Fails on 1.48.5 and master
```

### Severity

This broke the functionality of a customer which used to work on 47 but upgrade to 48 broke this

### Additional context

Since both logs and frontend logs (network and console) show nothing I am not sure if its a frontend or backend issue

WiNloSt:
Closes https://github.com/metabase/metabase/issues/38640

### Description

The problem stems from a fix for #33379. The fix was to navigate to custom URL which is the same URL as the Metabase instance using client-side. This caused a side-effect because `/embed/*` and `/public/*` serve different JS bundles which don't contain the route for each other. And they're meant to be standalone URLs.

So to fix this issue, I've extend the logic from the fix for #33379 to check that even we're navigating to the same Metabase URL, but to `/public/*` or `/embed/*` then we should open this as normal links.

Also, I've tested Metabase on subpath e.g. if Metabase is deployed on `/metabase` instead of `/` then the fix is still working correctly.

### How to verify

Follow the repro steps from #38640

### Demo

https://www.loom.com/share/1c4868e232974690bdf7cd4519d31dfc

### Checklist

### DIFF
--- a/frontend/src/metabase/lib/dom.js
+++ b/frontend/src/metabase/lib/dom.js
@@ -231,6 +231,17 @@ function isMetabaseUrl(url) {
     return true;
   }
 
+  const pathNameWithoutSubPath = getPathnameWithoutSubPath(urlPath);
+  const isPublicLink = pathNameWithoutSubPath.startsWith("/public/");
+  const isEmbedding = pathNameWithoutSubPath.startsWith("/embed/");
+  /**
+   * (metabase#38640) We don't want to use client-side navigation for public links or embedding
+   * because public app, or embed app are built using separate routes.
+   **/
+  if (isPublicLink || isEmbedding) {
+    return false;
+  }
+
   return isSameOrSiteUrlOrigin(url) && urlPath.startsWith(getSitePath());
 }
 
--- a/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-cards/click-behavior.cy.spec.js
@@ -11,6 +11,7 @@ import {
   getLinkCardDetails,
   getTextCardDetails,
   modal,
+  openStaticEmbeddingModal,
   popover,
   restore,
   saveDashboard,
@@ -18,6 +19,7 @@ import {
   updateDashboardCards,
   visitDashboard,
   visitEmbeddedPage,
+  visitIframe,
 } from "e2e/support/helpers";
 
 import {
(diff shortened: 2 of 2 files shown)


frontend/src/metabase/lib/dom.js
```diff
@@ -231,6 +231,17 @@ function isMetabaseUrl(url) {
...
   }
 
+  const pathNameWithoutSubPath = getPathnameWithoutSubPath(urlPath);
+  const isPublicLink = pathNameWithoutSubPath.startsWith("/public/");
+  const isEmbedding = pathNameWithoutSubPath.startsWith("/embed/");
+  /**
+   * (metabase#38640) We don't want to use client-side navigation for public links or embedding
+   * because public app, or embed app are built using separate routes.
+   **/
+  if (isPublicLink || isEmbedding) {
```
WiNloSt, line 241: Please see the [PR description](https://github.com/metabase/metabase/pull/38647#issue-2129748766) for an explanation.

oisincoveney (approved):
Nice fix! Just need to make sure the tests pass

WiNloSt:

WiNloSt:

PR #38655, opened 2024-02-12
should make test pass in release when we have a specific version in t…

npretto:
### Description

Solves https://github.com/metabase/metabase/actions/runs/7870757998/job/21472700573#step:10:818

During releases, metabase actually have a version so getDocsUrl will return a link to the specific version, not to latest


calherries (approved):
Works for me

npretto:
Manually backported in https://github.com/metabase/metabase/pull/38656

PR #38661, opened 2024-02-12
Browse models: Add filters

rafpaf:
This PR adds one filter to the Browse models page: a toggle that turns on and off the option of showing only models that have been verified

image

image

This is the successor to an earlier PR: https://github.com/metabase/metabase/pull/38560


frontend/src/metabase/browse/components/BrowseApp.tsx
```diff
@@ -40,6 +42,7 @@ export const BrowseApp = ({
       filter_items_in_personal_collection: "exclude",
     },
   });
+  const collectionsResult = useCollectionListQuery();
```
rafpaf, line 45: Collections are now retrieved so that we can use the correct icon for Metabase
analytics, which is the "audit" icon

frontend/src/metabase/browse/components/BrowseApp.styled.tsx
```diff
@@ -76,3 +63,8 @@ export const CenteredEmptyState = styled(EmptyState)`
   justify-content: center;
   height: 100%;
 `;
+
+export const LearnAboutDataIcon = styled(Icon)`
+  min-width: 14px;
+  min-height: 14px;
+`;
```
rafpaf, line 70: Without this, the icon begins to shrink when the viewport is narrow. I don't know why. I saw the same problem, and used the same solution, for the icon in the ModelExplanationBanner.

frontend/src/metabase/browse/components/BrowseApp.tsx
```diff
@@ -48,6 +52,62 @@ export const BrowseApp = ({
...
+            ? filter.activeByDefault
+            : storedFilterStatus === "on";
+        return {
+          ...acc,
+          [filterName]: shouldFilterBeActive,
+        };
+      },
+      {},
+    );
+  }, [availableModelFilters]);
```
rafpaf, line 76: On second thought, `useCallback` is probably not needed here

frontend/src/metabase/browse/utils.ts
```diff
@@ -29,25 +31,92 @@ export const groupModels = (
...
+  }
+  if (!a.name && !b.name) {
+    return 0;
+  }
+  const nameA = a.name.toLowerCase();
+  const nameB = b.name.toLowerCase();
+  return nameA.localeCompare(nameB, localeCode);
+};
+
+export type ActualModelFilters = Record<string, boolean>;
```
rafpaf, line 122: @iethree I might be wrong, but I think we might not be able to use the type system to enforce that the keys of this Record match the keys of AvailableModelFilters, since those keys can't be known at compile time - they depend on whether the enterprise plugins are loaded

iethree: yeah, you're probably right 😞

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -43,10 +42,13 @@ export const BrowseModels = ({
...
-    model => !isInstanceAnalyticsCollection(model.collection),
-  );
-  const groupsOfModels = groupModels(modelsFiltered, localeCode);
+
+  useEffect(() => {
+    if (error || isLoading) {
+      return;
+    }
+    localStorage.setItem("defaultBrowseTab", "models");
+  }, [error, isLoading]);
```
iethree, line 51: bad rebase? this should be deleted.

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -147,6 +137,10 @@ const ModelCell = ({ model, collectionHtmlId }: ModelCellProps) => {
     model.last_editor_common_name ?? model.creator_common_name;
   const timestamp = model.last_edited_at ?? model.created_at ?? "";
 
+  const dispatch = useDispatch();
+  const wrappedModel = Search.wrapEntity(model, dispatch);
+  const icon = wrappedModel.getIcon();
```
iethree, line 142: I found a nicer pattern for this that doesn't require dispatch

```ts

  const entity = entityForObject(model);
  const icon =  entity?.objectSelectors?.getIcon?.(model);
```

frontend/src/metabase/ui/components/icons/Icon/icons/model_with_badge.svg
iethree, line 1: 💖

iethree (changes requested):
Looks good, other than a couple things that I think were rebase mistakes.

enterprise/frontend/src/metabase-enterprise/content_verification/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  return 0;
+};
+
+export const availableModelFilters: AvailableModelFilters = {
+  onlyShowVerifiedModels: {
+    predicate: model => model.moderated_status === "verified",
+    activeByDefault: true,
+  },
+};
```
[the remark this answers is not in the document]
rafpaf: cheers

enterprise/frontend/src/metabase-enterprise/content_verification/ModelFilterControls.tsx
```diff
@@ -0,0 +1,36 @@
...
+        handleModelFilterChange("onlyShowVerifiedModels", e.target.checked);
+      }}
+      ml="auto"
+      size="sm"
+      labelPosition="left"
+      styles={{
+        body: { alignItems: "center" },
+        labelWrapper: { justifyContent: "center", padding: 0 },
+        track: { marginTop: "-.5px" },
+      }}
```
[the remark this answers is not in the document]
rafpaf: Originally I used

```ts
<Switch label={<strong>{t`Only show verified models`}</strong>} /* more props... */ />
```

But this doesn't look good when the viewport is narrow:

image

Per @naatz-co's request, I adjusted the line-height and text-align, with:

```ts
<Switch
  label={
    <Text
      align="right"
      weight="bold"
      lh="1rem"
      pr=".25rem"
    >{t`Only show verified models`}</Text>
  }
  /* more props...* /
/>
```

But that makes the switch look wonky:

image

I had to use Mantine's `styles` prop to control the appearance of the Switch's children.

Now it looks even:

image

And it looks better, I think, when the viewport is narrow:

image

naatz-co (approved):
It's good! I only have a few comments:
- Change the spacing between the verified models text and toggle to 12px
- Add tooltips in the verified and non-verified model icon so users can hover on it and see if it's a verified model or not (as we are already doing for collections in browse data)
- Metabase analytics should be at the bottom of the page

rafpaf:
@naatz-co 

> * Change the spacing between the verified models text and toggle to 12px

Done

> * Add tooltips in the verified and non-verified model icon so users can hover on it and see if it's a verified model or not (as we are already doing for collections in browse data)

Done

> * Metabase analytics should be at the bottom of the page

This will be possible once we improve the backend. (This improvement will also give us the correct icon for the Metabase analytics collection.) I will leave this for a follow-up PR.

enterprise/frontend/src/metabase-enterprise/content_verification/ModelFilterControls.tsx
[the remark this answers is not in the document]
rafpaf: @iethree, Nathalia reminded me of some of the context here: we will replace the Switch with a dropdown when we have more than one filter.

frontend/src/metabase/browse/components/BrowseApp.tsx
```diff
@@ -48,6 +52,62 @@ export const BrowseApp = ({
     }
   }, [tab]);
 
+  const availableModelFilters =
+    PLUGIN_CONTENT_VERIFICATION.availableModelFilters;
```
iethree, line 56: this can be outside the function body, right, that way it won't need to be in react dependency arrays

frontend/src/metabase/browse/components/BrowseApp.tsx
```diff
@@ -48,6 +52,62 @@ export const BrowseApp = ({
...
+  useEffect(() => {
+    const filteredModels = _.reduce(
+      actualModelFilters,
+      (acc, shouldFilterBeActive, filterName) =>
+        shouldFilterBeActive
+          ? acc.filter(availableModelFilters[filterName].predicate)
+          : acc,
+      unfilteredModels,
+    );
+    setFilteredModels(filteredModels);
```
iethree, line 93: nit, this would be nice if this logic was extracted in a more testable way. I know the parts are unit tested, but this combination is a little complex and a unit test would be nice of a more encapsulated function

I think it also might be even nicer as a `useMemo`

```ts
const filteredModels = useMemo(() => 
    filterModels(unfilteredModels, actualFilters, availableModelFilters)
, [unfilteredModels, actualFilters]);
```

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -155,7 +141,7 @@ const ModelCell = ({ model, collectionHtmlId }: ModelCellProps) => {
     >
       <ModelCard>
         <Box mb="auto">
-          <Icon name="model" size={20} color={color("brand")} />
+          <Icon {...icon} size={20} className="text-brand" />
```
iethree, line 144: rebase regression? we should use color instead of className

iethree (approved):
Looks good!

PR #38669, opened 2024-02-12
refactor e2e of onboarding/setup

npretto:
Part of [Post RC1 refactors for Meet embedders](https://github.com/metabase/metabase/issues/38556)

### Description

- converts `findByTextEnsureVisible` to ts (e070d46c8d07baf8112008efc136c8f1b2ab010d)
- extracts some repetitive steps into helpers
- make the snowplow test arrive at the last step to make sure we're testing all step_seen events
- adds assertions on the snowplow events we send from the FE


e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
```diff
@@ -406,44 +367,129 @@ describeWithSnowplow("scenarios > setup", () => {
     cy.visit(`/setup`);
 
     // 3 - setup/step_seen "welcome"
```
npretto, line 369: I left the comments even though there is the expectGood.... because we also check how many events are dispatched, and the comments are numbered

e2e/support/commands/visibility/findByTextEnsureVisible.ts
```diff
@@ -0,0 +1,15 @@
...
+    interface Chainable {
+      findByTextEnsureVisible(
+        text: string,
+      ): Cypress.Chainable<JQuery<HTMLElement>>;
+    }
+  }
+}
+
+Cypress.Commands.add("findByTextEnsureVisible", text => {
+  cy.findByText(text).should("be.visible");
```
WiNloSt, line 12: TBH, I never knew (or already forgot) this util exists. I'd usually find the element and call `.should("be.visible")` manually. That way you could possibly use any of the methods to find your elements and don't have to be limited to just `findByText`.

I'm not saying we should remove this util since it's being used in many places. I just want to share my opinion.

e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
```diff
@@ -406,44 +367,129 @@ describeWithSnowplow("scenarios > setup", () => {
...
+    skipWelcomePage();
 
     expectGoodSnowplowEvents(1);
   });
 });
+
+const skipWelcomePage = () => {
+  cy.findByTestId("welcome-page").within(() => {
+    cy.findByText("Welcome to Metabase");
+    cy.findByTextEnsureVisible("Let's get started").click();
```
WiNloSt, line 451: If I remember correctly, we couldn't click invisible elements in Cypress anyway. Remember when sometimes you need to pass `force: true` when the element isn't visible on the screen e.g. when it's outside viewport, has 0x0 in size, or actually not be visible to user's eyes.

This is redundant in my opinion. `cy.findByText("blabla").click()` should do the same thing. Although it's nice to be explicit about the assertions, so I'm not suggesting you change anything. It's just that personally, I wouldn't assert an element visibility before clicking it.

e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
```diff
@@ -175,24 +168,14 @@ describe("scenarios > setup", () => {
...
+      selectPreferredLanguageAndContinue();
 
       // User
-      cy.findByText("What should we call you?");
-      cy.findByLabelText("Email").type(admin.email);
-      cy.findByLabelText("Company or team name").type("Epic Team");
-      cy.findByLabelText("Create a password").type(admin.password);
-      cy.findByLabelText("Confirm your password").type(admin.password);
-      cy.button("Next").click();
+      fillUserAndContinue({ firstName: null, lastName: null });
```
WiNloSt, line 178: I'd like to see we require an argument for this util so we know what'll be set instead of having to go into the implementation detail to see that the default value is from an admin.

I think this would be easier to understand

```suggestion
      fillUserAndContinue({...admin, firstName: null, lastName: null});
```

We could probably do this for line 51 and below as well. There is `USERS.normal` maybe that will also work, otherwise, we could just create a `user` object which contains the information from line 51 and below. So we could use the same util
```ts
      fillUserAndContinue(normal);
```

e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
```diff
@@ -406,44 +367,129 @@ describeWithSnowplow("scenarios > setup", () => {
...
   it("should ignore snowplow failures and work as normal", () => {
     blockSnowplow();
     cy.visit(`/setup`);
-    cy.findByTestId("welcome-page").within(() => {
-      cy.findByText("Welcome to Metabase");
-      cy.findByTextEnsureVisible("Let's get started").click();
-    });
+    skipWelcomePage();
 
     expectGoodSnowplowEvents(1);
```
WiNloSt, line 444: I know it's the current assertion, but why would we have 1 good Snowplow event when we already block it on line 440?

WiNloSt (approved):
I have minor feedback, but overall it's in a great shape.

e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
npretto: It did confuse me too but I forgot to leave a comment, it's because when we block snoplow, we block it at the cypress/browser level, and we have an event being dispatched from the BE. I'll add a small comment as this is not immediate to understand

WiNloSt: That makes sense. Thanks for the explanation.

e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
npretto: Made all arguments non required so it's easier to read. e17682641defb9c484ef0a3fa892116962d39983
I left the test at line ~50 as it is as that was testing weak and strong password validation.

Issue #37363, opened 2024-01-08
FE - Metrics in Search filters


kamilmielnik:
Depends on:
- #37349

----

- [ ] Allow filtering for Metrics as Content type

Issue #37361, opened 2024-01-08
FE - Metrics in Search


kamilmielnik:
Depends on:
- #37349
- #37352

----

- [ ] Make sure to display correct icon (should have been added in #37352)
- [ ] For ranking logic, Metrics should be right above “models” in our current weighting system.

PR #38675, opened 2024-02-12
Add metrics to search

ranquild:
Closes https://github.com/metabase/metabase/issues/37361
Closes https://github.com/metabase/metabase/issues/37362
Closes https://github.com/metabase/metabase/issues/37363

Screenshot 2024-02-12 at 17 53 00
Screenshot 2024-02-12 at 17 53 05

Doesn't really work without the BE changes. To make it appear, create a metric v1.

### DIFF
--- a/frontend/src/metabase-types/api/search.ts
+++ b/frontend/src/metabase-types/api/search.ts
@@ -9,14 +9,15 @@ export type EnabledSearchModelType =
   | "collection"
   | "dashboard"
   | "card"
+  | "metric"
+  | "dataset"
   | "database"
   | "table"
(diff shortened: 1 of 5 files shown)


frontend/src/metabase-types/api/search.ts
```diff
@@ -9,14 +9,15 @@ export type EnabledSearchModelType =
   | "collection"
   | "dashboard"
   | "card"
+  | "metric"
```
kamilmielnik, line 12: Clicking a metric in search results should open `/metric/:id`, now it opens `/question/:id`

PR #38678, opened 2024-02-12
Make Mantine tooltips consistent with legacy tooltips

kulyk:
Updates the `Tooltip` component from `metabase/ui` to match the appearance of legacy tooltip components (padding, background color, font size and weight, in/out transition)

### Demo

**Before**
CleanShot 2024-02-12 at 16 45 43@2x

**After**
CleanShot 2024-02-12 at 16 45 13@2x


iethree (approved):
code looks good, but I had thought we were intentionally trying to change the design?

kdoh (approved):
Looks good. Right now we want to keep the tooltips the same as they have been for a while (with pointer, slightly lighter color). The raw mantine style got approved to support some other work and I did a poor job following up to get things sync'd back up.

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -324,10 +323,10 @@ class DashboardHeaderContainer extends Component<DashboardHeaderProps> {
...
       </Button>,
-      <DashboardButtonTooltip
-        disabled={!isSaveDisabled}
-        label={disabledSaveTooltip}
+      <Tooltip
         key="save"
+        label={disabledSaveTooltip}
+        disabled={!isSaveDisabled}
       >
         <ActionButton
```
JesseSDevaney, line 331: I think we need to add a span around this action button, because the tooltip is not rendering without it.

**Without**
image

**With**
image

JesseSDevaney (approved):
I found one thing we should address before merging. I am still going to approve, because the change is small

PR #38695, opened 2024-02-12
Extend param substitution to handle datetimes

### LINKED ISSUE #33492: Change native SQL datetime range field filters to be inclusive of the end minute
Currently native SQL datetime range field filters have a meaningfully different translation in SQL compared to GUI question filters. We should change native SQL datetime range filters to work the same as GUI question filters.

This change will cause some native questions to return different results.

[Slack context behind the decision](https://metaboat.slack.com/archives/C01LQQ2UW03/p1692607179959739)

This only applies to date ranges with a time component like this:

image

And it only applies to SQL drivers.

An explanation of the problem with an example:

```
For a date picker with this input:
08/08/2023 12:30 PM - 08/10/2023 12:30 PM
 
In a GUI question, a filter with that date range is translated to this SQL:
<column> >= timestamp '2023-08-08 12:30:00.000' AND <column> < timestamp '2023-08-10 12:31:00.000'

In an Native question, a template tag with that date range is translated to this SQL:
<column> BETWEEN timestamp '2023-08-08 12:30:00.000' AND timestamp '2023-08-10 12:30:00.000'

The difference is subtle but important. e.g.

                             GUI   |   Native
2023-08-10 12:30:00.000   included |  included <-- same
2023-08-10 12:30:00.001   included |  excluded <-- different
2023-08-10 12:30:59.999   included |  excluded <-- different
2023-08-10 12:31:00.000   excluded |  excluded <-- same
```

### LINKED ISSUE #38037: Relative date filter with time works in Field Filters different for SQL variables than for GUI
### Describe the bug

GUI filters give correct results; the issue comes when filtering with variables in SQL to datetime fields. The results given in each query differ, even though the data, tables, filter configuration, etc are the same.

### To Reproduce (EDIT: Below are easier repro steps)

1. Create a table in a MySQL database with a datetime field. Example SQL below the steps
2. Create a GUI query that shows all the fields, and use the table view to filter the Datetime Field. Select "Specific dates" as the filter type, and select "Add a time". 
3. To filter, select "Between" January 19th 2024 and January 19th 2024, and select an inclusive time for the fields you want (for example, 12:30pm to 5pm). See that the filter works correctly.
4. Now create a SQL question `select * from mytable where {{my_date_filter}}`
5. In the variables menu, select Field Filter as variable type, map it to datetime_field, and select Date filter as the Filter widget type. 
6. Now, filter by the _same_ dates and times that you used in the GUI, Jan 19th, between 12:30 pm and 5pm (inclusive).
7. See that there are no results.

**SQL for step 1**
```
CREATE DATABASE IF NOT EXISTS example_db;
USE example_db;

CREATE TABLE test_table (
    id INT AUTO_INCREMENT PRIMARY KEY,
    datetime_field DATETIME
);

INSERT INTO test_table (datetime_field) VALUES 
('2024-01-19 15:55:00'),
('2024-01-19 16:00:00');
```

This is the SQL generated by the GUI query:
```
SELECT
  `test_table`.`id` AS `id`,
  `test_table`.`datetime_field` AS `datetime_field`
FROM
  `test_table`
WHERE
  (
    `test_table`.`datetime_field` >= convert_tz(
      '2024-01-19 12:30:00.000',
      'America/Argentina/Buenos_Aires',
      @@session.time_zone
    )
  )
 
   AND (
    `test_table`.`datetime_field` < convert_tz(
      '2024-01-19 17:01:00.000',
      'America/Argentina/Buenos_Aires',
      @@session.time_zone
    )
  )
LIMIT
  1048575
```

And this is the query preview for the SQL query:
```
select * from test_table where DATE(`test_table`.`datetime_field`) BETWEEN timestamp '2024-01-19 13:30:00.000'
   AND timestamp '2024-01-19 12:30:00.000'
```

### Expected behavior

Results should be the same?

### Logs

_No response_

```JSON
- Metabase 48.3
- MySQL 8.1.0
```

### Severity

P2

### Additional context

Related: https://github.com/metabase/metabase/issues/4284

snoe:
Fixes #38037
Fixes #33492

Prior to this only dates were supported as values for native timestamp field parameters. By properly aligning the temporal unit, we can ensure that native filters match gui filters.

### DIFF
--- a/test/metabase/driver/sql/parameters/substitution_test.clj
+++ b/test/metabase/driver/sql/parameters/substitution_test.clj
@@ -14,7 +14,7 @@
    [metabase.test :as mt]
    [metabase.util.honey-sql-2 :as h2x])
   (:import
-   (java.time LocalDateTime)))
+   (java.time LocalDate LocalDateTime)))
 
 (set! *warn-on-reflection* true)
 
@@ -83,8 +83,10 @@
   [_driver _field _param-type _value]
   nil)
 
-;; The original implementation will call this method despite the value being past30minutes. This is likely a bug.
-;; However, `metabase.query-processor-test.alternative-date-test/substitute-native-parameters-test` depends on it.
+(defmethod sql.qp/date [::temporal-unit-alignment-original :minute]
+  [_driver _unit expr]
+  (h2x/minute expr))
+
 (defmethod sql.qp/date [::temporal-unit-alignment-original :day]
   [_driver _unit expr]
   (h2x/day expr))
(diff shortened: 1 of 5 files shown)


test/metabase/driver/sql/parameters/substitution_test.clj
```diff
@@ -93,7 +93,7 @@
   (mt/with-clock #t "2018-07-01T12:30:00.000Z"
     (mt/with-metadata-provider meta/metadata-provider
       (let [field-filter (params/map->FieldFilter
-                          {:field (lib.metadata/field (qp.store/metadata-provider) (meta/id :checkins :date))
+                          {:field (lib.metadata/field (qp.store/metadata-provider) (meta/id :orders :created-at))
```
metamben, line 96: It's unexpected for me that the rest of the test doesn't need changing. Where does/should `"PUBLIC"."CHECKINS"` come from?

metamben (approved):
LGTM, just one test is strange. Is there a GH issue for the problem this PR is fixing? If there is, we should include it,

calherries:

calherries:
@snoe I'm backporting to 48 because it looks like this never hit the 48 branch.

PR #38704, opened 2024-02-13
Browse models: expandable/collapsible collections

rafpaf:
Needs tests

This PR allows the user to expand and collapse collections in Browse data. The user can show and hide the entire collection using the chevron icon to the left of the header. Independently, the user can show all of the models within the collection, or just a portion, using the 'Show all'/'Show less' button. These toggles are independent.

The user's preferences about which collection is shown are persisted across sessions using a database setting. (Note: there is a concern about race conditions here, since the API calls that update the setting could hypothetically be handled by the backend in a different order than they were triggered in. I have left this problem unresolved for now.)

expand-collapes-colls

- [x] has tests

### DIFF
--- a/frontend/src/metabase/browse/components/BrowseDatabases.tsx
+++ b/frontend/src/metabase/browse/components/BrowseDatabases.tsx
@@ -3,16 +3,13 @@ import { t } from "ttag";
 import NoResults from "assets/img/no_results.svg";
 import type { useDatabaseListQuery } from "metabase/common/hooks";
 import LoadingAndErrorWrapper from "metabase/components/LoadingAndErrorWrapper";
-import Link from "metabase/core/components/Link";
-import { color } from "metabase/lib/colors";
-import * as Urls from "metabase/lib/urls";
 import { Box, Icon, Title } from "metabase/ui";
 
 import { CenteredEmptyState } from "./BrowseApp.styled";
 import {
   DatabaseCard,
+  DatabaseCardLink,
   DatabaseGrid,
-  DatabaseGridItem,
 } from "./BrowseDatabases.styled";
 
 export const BrowseDatabases = ({
@@ -33,8 +30,8 @@ export const BrowseDatabases = ({
   return databases.length ? (
     <DatabaseGrid data-testid="database-browser">
       {databases.map(database => (
-        <DatabaseGridItem key={database.id}>
-          <Link to={Urls.browseDatabase(database)}>
+        <div key={database.id}>
+          <DatabaseCardLink to={Urls.browseDatabase(database)}>
             <DatabaseCard>
               <Icon
                 name="database"
@@ -42,12 +39,12 @@ export const BrowseDatabases = ({
                 className="mb3"
                 size={32}
               />
-              <Title order={2} size="1rem" lh="1rem">
+              <Title order={2} size="1rem" lh="1rem" color="inherit">
                 {database.name}
               </Title>
             </DatabaseCard>
-          </Link>
-        </DatabaseGridItem>
+          </DatabaseCardLink>
+        </div>
       ))}
     </DatabaseGrid>
   ) : (
--- a/frontend/src/metabase/browse/components/BrowseDatabases.tsx
+++ b/frontend/src/metabase/browse/components/BrowseDatabases.tsx
@@ -3,6 +3,8 @@ import { t } from "ttag";
 import NoResults from "assets/img/no_results.svg";
 import type { useDatabaseListQuery } from "metabase/common/hooks";
 import LoadingAndErrorWrapper from "metabase/components/LoadingAndErrorWrapper";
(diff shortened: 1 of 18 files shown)


frontend/src/metabase/browse/components/ModelGroup.tsx
```diff
@@ -1,54 +1,199 @@
...
+  const dispatch = useDispatch();
+
+  const newCollectionPrefs = {
+    ...initialCollectionPrefs,
+    [collection.id]: { areAllModelsShown, areSomeModelsShown },
+  };
+  const newCollectionPrefsStringified = JSON.stringify(newCollectionPrefs);
+
+  useEffect(() => {
+    // FIXME: Can this lead to race conditions? Perhaps a system with Promise.all would make sense?
```
rafpaf, line 115: ^ Important to discuss this

frontend/src/metabase/browse/components/BrowseDatabases.tsx
```diff
@@ -4,7 +4,6 @@ import { color } from "metabase/lib/colors";
 import * as Urls from "metabase/lib/urls";
```
rafpaf, line 4: This small styling change is not related to the theme of this PR. Happy to extract it out.

frontend/src/metabase/browse/components/ModelGroup.tsx
```diff
@@ -1,54 +1,199 @@
...
+  try {
+    const parsed = JSON.parse(json || "");
+    return isValidCollectionPrefs(parsed) ? parsed : null;
+  } catch (e) {
+    if (e instanceof SyntaxError) {
+      return null;
+    }
+    throw e;
+  }
+};
```
iethree, line 71: 1) I don't think settings like this should be server-side, they're simple UI interactions, not data we need to persist. we don't want to cause a bunch of network requests and db write/reads every time the user simply manipulates the UI.
2) We should lift the expand/collapse show/hide state up since it's all one object and only pass down necessary properties - we have potentially thousands of collections here and we don't need them all doing the exact same parsing when we can do it once at the parent-component level.
3) these helper functions are great, but also perfect candidates to be extracted to a `utils` file to keep the component easier to understand at a glance.

kdoh (approved):

frontend/src/metabase/browse/components/BrowseModels.tsx
```diff
@@ -34,20 +39,65 @@ export const BrowseModels = ({
...
+                }
+                showAll={!!collectionViewPreferences?.[collectionId]?.showAll}
+                toggleExpanded={() =>
+                  handleToggleCollectionExpand(collectionId)
+                }
+                toggleShowAll={() =>
+                  handleToggleCollectionShowAll(collectionId)
+                }
+                models={groupOfModels}
+                key={`modelgroup-${groupOfModels[0].collection.id}`}
```
sloansparger, line 96: world's tiniest nit (only say this because I want us to ban index access at some point since it's not typesafe)
```suggestion
                key={`modelgroup-${collectionId}`}
```

frontend/src/metabase/browse/components/ModelGroup.tsx
```diff
@@ -1,34 +1,59 @@
...
+
+    return {
+      aboveFoldModels: sortedModels.slice(0, MAX_COLLAPSED_MODELS),
+      belowFoldModels: sortedModels.slice(MAX_COLLAPSED_MODELS),
+    };
+  }, [models, localeCode]);
+
+  const visibleModels = showAll
+    ? [...aboveFoldModels, ...belowFoldModels]
+    : aboveFoldModels;
```
sloansparger, line 56:
```suggestion
  const sortedModels = useMemo(() => {
     return [...models].sort((a, b) => sortModels(a, b, localeCode))
  }, [models, localeCode]);
  
  const visibleModels = useMemo(() => {
    return showAll
      ? sortedModel
      : sortedModels.slice(0, MAX_COLLAPSED_MODELS)
  }, [sortedModels, showAll]);
```

I think this would be preferable for two reason:
1. sort mutates the underlying array, so it's best to copy first
2. slice shallow copies, but then we reconstruct sortedModels by doing `[...aboveFoldModels, ...belowFoldModels]`. this way we can avoid the extra copies.

frontend/src/metabase/browse/utils.ts
```diff
@@ -136,3 +145,52 @@ export const filterModels = (
...
+
+type ModelVisibilityPrefs = {
+  expanded: boolean;
+  showAll: boolean;
+};
+
+const isRecordWithCollectionIdKeys = (
+  prefs: unknown,
+): prefs is Record<CollectionId, any> =>
+  typeof prefs === "object" &&
```
sloansparger, line 159: We should have some better type utils. This would incorrectly work with an array.

sloansparger (approved):
approved with one comment

PR #38707, opened 2024-02-13
Use dynamic redefs throughout tests

crisptrutski:
It's very convenient to stub out functions in tests, and we do this a lot. Unfortunately the way we were doing it was no thread-safe, which is a bother for tests that could otherwise be run in parallel.

This PR replaces all our test usage with our custom thread-safe version from https://github.com/metabase/metabase/pull/38620

Hopefully it is discoverable enough now and people internalize reaching for it. I'm going to be optimistic / lazy and skip creating a linter for this.

### DIFF
--- a/test/metabase/driver/sql/query_processor_test.clj
+++ b/test/metabase/driver/sql/query_processor_test.clj
@@ -243,7 +243,7 @@
 
 (deftest adjust-start-of-week-test
   (driver/with-driver :h2
-    (with-redefs [driver/db-start-of-week   (constantly :monday)
+    (mt/with-dynamic-redefs [driver/db-start-of-week   (constantly :monday)
                   setting/get-value-of-type (constantly :sunday)]
       (is (= [:dateadd
               (h2x/literal "day")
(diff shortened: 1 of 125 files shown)


crisptrutski:
Whitespace was a pain. I've probably messed up a few spots, but I got most of them. I think it's OK for boy scouts to fix up the occasional alignment as they come across it.

test/metabase/driver/sql_jdbc/connection_test.clj
```diff
@@ -48,7 +48,7 @@
             original-destroy   @#'sql-jdbc.conn/destroy-pool!
             connection-details {:db "mem:connection_test"}
             spec               (mdb.spec/spec :h2 connection-details)]
-        (with-redefs [sql-jdbc.conn/destroy-pool! (fn [id destroyed-spec]
+        (mt/with-dynamic-redefs [sql-jdbc.conn/destroy-pool! (fn [id destroyed-spec]
```
calherries, line 51: You broke formatting here

test/metabase/driver/sql/query_processor_test.clj
```diff
@@ -243,7 +243,7 @@
 
 (deftest adjust-start-of-week-test
   (driver/with-driver :h2
-    (with-redefs [driver/db-start-of-week   (constantly :monday)
+    (mt/with-dynamic-redefs [driver/db-start-of-week   (constantly :monday)
```
calherries, line 246: formatting

calherries (approved):
Nice, this will help with discoverability.

There is a lot of broken formatting and I gave up commenting on them so I don't blame you for missing them :D But you should clean these up before merging

calherries:
> I think it's OK for boy scouts to fix up the occasional alignment as they come across it.

I don't think that makes sense from an efficiency pov. It's much less effort to have everything updated in one go than a thousand times separately.

calherries:
It's a shame we don't have an auto-formatter for these things though

calherries:
@crisptrutski did you want to close this?

crisptrutski:
Yeah, given that the issues went deeper than just formatting, I've lost the energy for a big-bang refactor.

PR #38725, opened 2024-02-13
Add info icon to the column picker dropdown in the notebook editor

### LINKED ISSUE #38698: Add hover card on column picker
As part of #38394 we want to surface relevant metadata in places where it makes sense.

Adding the metadata hovercard on the column picker is a good first step.

Screenshot 2024-02-13 at 09 56 19

romeovs:
Closes https://github.com/metabase/metabase/issues/38698

### Description

- Refactor `FieldInfo`, `FieldInfoFingerprint` and `FieldInfoPopover` to also accept `{ query, stageIndex, column }` props (instead of the legacy `field`).
- Add `FieldInfoIcon` that displays an info icon that displays the field info popover on hover
- Add the icon to the field picker dropdown
- Update `TableInteractive` to use the MLv2 props of the `FieldInfoPopover`

This PR lays a lot of the ground work for the other items in the epic.

I'm merging this all to a feature branch so the design team can have a final look before we merge to `master`.

### How to verify

1. New question → Sample Dataset → Orders
2. Click the caret down next to the `Orders` button
3. Hover the items in the list
4. Verify that a popover appears with the correct description and type info

### Demo
Screenshot 2024-02-13 at 17 31 00

### Checklist


mazameli (approved):

ranquild (commented):
Can we improve how components are composed to avoid this:
`if ("field" in props) {`

Ideally we should have 2 top-level components, one accepting `query/stageIndex/column`, and another `field` for compatibility. And inner components should accept more granular data that wouldn't require prop checks and conditional statements.

E.g. `QueryColumnInfo`:

```
const columnInfo = Lib.displayInfo(query, stageIndex, column);

return (
  <ColumnInfoRoot>
    <ColumnDescription description={columnInfo.description} />
    <ColumnSemanticLabel description={columnInfo.description} />
    <ColumnFingerprint fingerprint={columnInfo.fingerprint} />
  </FieldInfoRoot>
);
```

`TableColumnInfo`:

```

return (
  <ColumnInfoRoot>
    <ColumnDescription description={field.description} />
    <ColumnSemanticLabel description={field.description} />
    <ColumnFingerprint fingerprint={field.fingerprint} />
  </FieldInfoRoot>
);
```
You can avoid type checks (`isDate` etc) and rely on fingerprint data. E.g. if `fingerprint.type["type/DateTime"]` is here it's a date column.

Another small thing is that it's better to use "column" because here an arbitrary query column can be passed. A "field" in Metabase is a real database field only.

romeovs:
@ranquild Ah, I was doing this pattern after discussions with members of the team, since I wasn't sure what pattern was preferable.

If making two version of the component, rather than overloading, what should their names be? Should they live in the same file?

ranquild:
> @ranquild Ah, I was doing this pattern after discussions with members of the team, since I wasn't sure what pattern was preferable.
> 
> If making two version of the component, rather than overloading, what should their names be? Should they live in the same file?

I think for now it's ok to have them in the same file. Hopefully we get rid of Field class-based version soon.

ranquild:
@romeovs good point. For category we would need to use a type check

romeovs:
@ranquild Ok, in that case I might not try to merge the `ColumnFingerPrint` components into a base component since the checks need to be different.

I did refactor the other components and I feel the code is much cleaner.

frontend/src/metabase/components/MetadataInfo/ColumnInfoIcon/ColumnInfoIcon.styled.tsx
```diff
@@ -0,0 +1,26 @@
+import styled from "@emotion/styled";
+import { css } from "@emotion/react";
+import { Icon } from "metabase/ui";
+
+export const PopoverHoverTarget = styled(Icon)<{ hasDescription: boolean }>`
+  padding: 0.7em 0.65em;
+  opacity: 0;
+
+  path {
```
ranquild, line 9: We don't target html elements like that. Can this be avoided? Could you set the opacity on while icon?

frontend/src/metabase/components/MetadataInfo/ColumnInfoIcon/ColumnInfoIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+  TableColumnInfoPopoverProps,
+} from "../ColumnInfoPopover";
+import {
+  QueryColumnInfoPopover,
+  TableColumnInfoPopover,
+} from "../ColumnInfoPopover";
+
+import { PopoverHoverTarget } from "./ColumnInfoIcon.styled";
+
+const LAST_STAGE = -1;
```
ranquild, line 15: Can we make this parameter required? Omitting `stageIndex` could lead to bugs

ranquild (commented):
Small issues left

frontend/src/metabase/common/components/FieldPicker/FieldPicker.styled.tsx
```diff
@@ -33,6 +34,18 @@ export const ColumnItem = styled.li`
       background: ${color("bg-medium")};
     }
   }
+
+  ${QueryColumnInfoIcon.HoverTarget} {
```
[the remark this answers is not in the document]
romeovs: The tricky part is that we want to hide the icon and then show it when a parent (eg. list item, dropdown item, ...) is being hovered.

We can't really know what that parent is going to be so we can't hard code it in the `HoverTarget`.

Setting up a hover handler or React context for this seems heavy-handed, though.

The only other think I can think of is to actually pass the parent selector as a prop to `ColumnInfoIcon`, eg.
```
<ColumnIconInfo {...props} parentSelector="li[role='option']" />
```

I'm open to other ideas, however.

frontend/src/metabase/components/MetadataInfo/ColumnInfoIcon/ColumnInfoIcon.styled.tsx
romeovs: This is related to the above issue. To keep the styles that need to be overridden simple I wanted to avoid doing opacity on the actual element. I can encapsulate the styles however and export them to make this less coupled.

frontend/src/metabase/common/components/FieldPicker/FieldPicker.styled.tsx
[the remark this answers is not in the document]
romeovs: One other way we _could_ do this is by using a `data-` prop to do the styling. Eg.

```
[data-popover-hover-target]:hover {
  opacity: 1;
}
```

That would require us setting the data- prop on the parent however, and I feel like it introduces the same amount of coupling, just less explicit.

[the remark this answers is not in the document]
romeovs: @ranquild Please have a look at the commits I've pushed (5ae9b59ffb6d10f64e1707b14d8e48356532ef6c and 1cdec12c6a2359662afa7b3cb9d85fbe75b83c8d) to see if you prefer any of these approaches.

ranquild (commented):
Looks good, last comment in slack https://metaboat.slack.com/archives/C0645JP1W81/p1708077733362619

frontend/src/metabase/common/components/FieldPicker/FieldPicker.styled.tsx
```diff
@@ -12,9 +14,17 @@ export const ItemIcon = styled(Icon)`
...
 
+export const QueryColumnInfoIcon = styled(_QueryColumnInfoIcon)`
+  color: ${alpha(darken(color("brand"), 0.6), 0.8)};
+  margin-left: auto;
+`;
+
 export const ColumnItem = styled.li`
   &:first-child {
     margin-top: 0.5em;
+    label {
```
ranquild, line 25: Please fix

frontend/src/metabase/components/MetadataInfo/ColumnInfoIcon/ColumnInfoIcon.styled.tsx
```diff
@@ -0,0 +1,17 @@
+import styled from "@emotion/styled";
+import { Icon } from "metabase/ui";
+
+export const HoverParent = styled.div``;
+
+export const PopoverHoverTarget = styled(Icon)<{ noDescription: boolean }>`
```
ranquild, line 6: Usually positive booleans are better `hasDescription`

ranquild (approved):

PR #38726, opened 2024-02-13
Distinguish different arities of aggregation-clause

### LINKED ISSUE #37782: [MLv2] [BE] Redundant `null` appears in serialized count aggregation
The count aggregation used to be serialized as `["count"]` but now it's `["count", null]`.
This is coming from MLv2's `toLegacyQuery` (`legacy_query`).

[Network payload screenshot](https://github.com/metabase/metabase/assets/6830683/3bd1589f-54f0-4956-8a9e-a90a19cc0853).

See [Slack discussion](https://metaboat.slack.com/archives/C04CYTEL9N2/p1705408133622419).

----

### To reproduce

Use this FE unit test ([test result screenshot](https://github.com/metabase/metabase/assets/6830683/380896cd-92fc-42d5-b5bc-c5e9975663be):

```tsx
import * as Lib from "metabase-lib";
import { createQuery } from "metabase-lib/test-helpers";

describe("#37782", () => {
  it("should serialize aggregation", () => {
    const query = createQuery();
    const stageIndex = -1;
    const operators = Lib.availableAggregationOperators(query, stageIndex);
    const countOperator = operators.find(operator => {
      const info = Lib.displayInfo(query, stageIndex, operator);
      return info.shortName === "count";
    });
    const aggregationclause = Lib.aggregationClause(countOperator);
    const finalQuery = Lib.aggregate(query, stageIndex, aggregationclause);
    const legacyQuery = Lib.toLegacyQuery(finalQuery);

    expect(legacyQuery.query.aggregation).toEqual(["count"]);
  });
});
```

### Additional context

:warning: 

Once this is fixed, **please unskip 4 test cases** in `frontend/src/metabase/reference/utils.unit.spec.js`.
See: https://github.com/metabase/metabase/pull/37694/commits/06fd8a707a1d39b21fb94e2f82e021c295bf8bbe

metamben:
Fixes #37782.

### DIFF
--- a/frontend/src/metabase-lib/aggregation.ts
+++ b/frontend/src/metabase-lib/aggregation.ts
@@ -63,9 +63,6 @@ export function aggregationClause(
   operator: AggregationOperator,
   column?: ColumnMetadata,
 ): AggregationClause {
-  if (typeof column === "undefined") {
-    return ML.aggregation_clause(operator);
-  }
   return ML.aggregation_clause(operator, column);
 }
 
--- a/frontend/src/metabase-lib/aggregation.ts
+++ b/frontend/src/metabase-lib/aggregation.ts
@@ -63,6 +63,9 @@ export function aggregationClause(
   operator: AggregationOperator,
   column?: ColumnMetadata,
 ): AggregationClause {
+  if (typeof column === "undefined") {
+    return ML.aggregation_clause(operator);
+  }
   return ML.aggregation_clause(operator, column);
 }
 
--- a/frontend/src/metabase/reference/utils.unit.spec.js
+++ b/frontend/src/metabase/reference/utils.unit.spec.js
@@ -156,9 +156,7 @@ describe("Reference utils.js", () => {
       expect(question).toEqual(getNewQuestion());
     });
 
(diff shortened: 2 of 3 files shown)


ranquild (commented):
@bshepherdson could you take a look? With drills you somehow managed to differentiate between `undefined` and `null` in CLJS

frontend/src/metabase-lib/aggregation.ts
```diff
@@ -63,6 +63,9 @@ export function aggregationClause(
   operator: AggregationOperator,
   column?: ColumnMetadata,
 ): AggregationClause {
+  if (typeof column === "undefined") {
```
bshepherdson, line 66: There's a function `cljs.core/undefined?` that distinguishes JS `js/undefined` from `null`, even though both are treated as `nil?` by normal CLJS. This logic could go inside `metabase.lib.js/aggregation-clause` instead.

metamben: It could, but is that preferable? And if it is, should the CLJS code have just a single arity (2) version?

bshepherdson (approved):

PR #38727, opened 2024-02-13
Fix search result display for items with large strings without word breaks

sloansparger:
### Description

before

You can see the (redacted) text above overflows the width of the container element. This is a product of using css grid's `auto` value with the `grid-template-columns` property in combination with long strings without word breaks. This typically appears when a user has text containing a long url in the description of the search result item.

Moving the description section inside the `ResultNameSection` puts it in a flex content which squishes the content in a way we would expect. The final result is more of what we're after

after

The downside to this approach is that we're not solving that any item who happens be to be large could break this layout since grid doesn't force it's children to shrink the same way something like flex does. It might be worth revisiting the layout of this component unless grid could be coerced to do something similar (I'm not familiar with grid at all really, so maybe it can). 

### Callout

Not sure if this is desired, but grid's unused columns on the right for items that do not have the xray or loading icon have some extra padding from the grid's gap. Just noting that in case this is another display bug worth fixing:

callout

### How to verify

- Modify a question/dashboard/model's description to contain a really long like
```
testingtestingtestingtestingtestingtestingtestingtestingt estingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtesting testingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtestingtesting
```
- Search for that entity in the search bar

### Checklist

### DIFF
--- a/e2e/test/scenarios/onboarding/search/search.cy.spec.js
+++ b/e2e/test/scenarios/onboarding/search/search.cy.spec.js
@@ -197,13 +197,13 @@ describe("scenarios > search", () => {
         description: `![alt](https://upload.wikimedia.org/wikipedia/commons/a/a2/Cat_outside.jpg)
 
         Lorem ipsum dolor sit amet.
-        
+
         ----
-        
+
         ## Heading 1
-        
+
         This is a [link](https://upload.wikimedia.org/wikipedia/commons/a/a2/Cat_outside.jpg).
-        
+
         Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. `,
       }).then(() => {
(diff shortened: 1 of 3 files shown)


oisincoveney:
@sloansparger I think this looks great. Are we sure that we can't do an E2E repro of searching for a question with a really long description, similar to your instructions?

sloansparger:
@oisincoveney done and done! let me know what you think

iethree (approved):
🎂

PR #38737, opened 2024-02-13
Use font fallbacks in the static-viz js

### LINKED ISSUE #38684: Trend Chart font is incorrect in certain email clients
### Describe the bug

in gmail on iOS, the trend chart card renders with an incorrect font, at least some of the time. It's especially noticeable because it is a Serif font, when the expected font is Lato (a Sans-serif font), or a more similar fallback (should be whatever Sans-serif font).

For some context, there is a Slack thread [here](https://metaboat.slack.com/archives/C064QMXEV9N/p1705688410924559) (Metabase internal Slack).

### Expected behavior

This should ideally render with the correct Lato font, but at least use a Sans-serif fallback.

### Logs

_No response_

```JSON
- gmail + iOS email client
- 49-RC1?
```

### Severity

P2

### Additional context

_No response_

adam-james-v:
Fixes: #38684 

The trend chart email was rendering incorrectly in gmail on iOS for some reason.

I think it happened because the Lato font was unavailable in the gmail+iOS context and this new :javascript_visualization (introduced with trend charts for the static-viz rendering pipeline) specified a font family of Lato with no fallbacks. It seems that CSS doesn't inherit font fallbacks from parent elements at all, so even though fallbacks were specified further up the tree, they weren't used.

This PR just adds the same fallbacks used elsewhere in static-viz render code.

Before:
iOS-wrong-font

After:
iOS-right-font

### DIFF
--- a/frontend/src/metabase/static-viz/index.js
+++ b/frontend/src/metabase/static-viz/index.js
@@ -20,7 +20,7 @@ export function RenderChart(rawSeries, dashcardSettings, colors) {
     getColor,
     formatValue: formatStaticValue,
     measureText: measureTextWidth,
-    fontFamily: "Lato",
+    fontFamily: "Lato, 'Helvetica Neue', Helvetica, Arial, sans-serif",
   };
 
   const props = {


markbastian (approved):
Ship it and let's see if it works.

alxnddr (approved):

PR #38740, opened 2024-02-13
Add caching for data permissions

johnswanson:
This caching is:

- thread-local. It's intended to be used just like
`api/*current-user-permission-set*` (although less directly): we use
middleware to wrap a single request/response handling cycle in
`with-relevant-permissions-for-user` to populate the cache, and then the
cache is used.

- scoped to a particular user. It will not be used if the
`api/*current-user-id*` isn't equal to the `user-id` in a particular
permissions query. The `user-id` is also used as an index when
retrieving permissions to prevent any possibility of accidentally using
cached permissions for a different user than intended.

- "indexed" by `[user-id, perm-type, db-id]` - the former for the above
reason, the latter two (perm-type and db-id) because essentially every
time we need to retrieve permissions we filter by these two columns.

To avoid having multiple codepaths for filtering permissions (e.g. one for cached vs. another for uncached), for now I'm retrieving all of a user's relevant permissions in *both* cases, then filtering them down when we calculate a given permission. Less efficient, yes, but this way we are always testing the same logic, where otherwise we might have subtle bugs where we filtered permissions differently in SQL vs from the cache. And I don't think we *ever* calculate permissions in production that *aren't* for the currently logged in user, so this shouldn't have any negative performance impact in prod.

(Edit: I actually decided that instead of select all of the user's relevant permissions, we could just select the ones related to the perm_type and db_id we're interested in.)

Closes [38040](https://github.com/metabase/metabase/issues/38040)

### DIFF
--- a/src/metabase/models/data_permissions.clj
+++ b/src/metabase/models/data_permissions.clj
@@ -89,29 +89,34 @@
 (defn relevant-permissions-for-user
   "Returns all relevant rows for permissions for the user"
   [user-id]
-  (group-by (juxt :perm_type :db_id)
+  (group-by (juxt :user_id :perm_type :db_id)
             (t2/select :model/DataPermissions
-                       {:select [:p.*]
+                       {:select [:p.* [:pgm.user_id :user_id]]
                         :from [[:permissions_group_membership :pgm]]
                         :join [[:permissions_group :pg] [:= :pg.id :pgm.group_id]
                                [:data_permissions :p] [:= :p.group_id :pg.id]]
-                        :where [:and
-                                [:= :pgm.user_id user-id]]})))
+                        :where [:= :pgm.user_id user-id]})))
 
 (def ^:dynamic *permissions-for-user*
   "Filled by `with-relevant-permissions-for-user` with the output of `(relevant-permissions-for-user [user-id])`. A map
-  with keys like `[perm_type db_id]`, because nearly always we want to get permissions for a particular permission
-  type and database id. Values are collections of rows of DataPermissions."
-  nil)
+  with keys like `[user_id perm_type db_id]`, the latter two because nearly always we want to get permissions for a
+  particular permission type and database id, `user_id` because we want to be VERY sure that we never accidentally use
+  the cache for the wrong user (e.g. if we were checking whether *another* user could perform some action for some
+  reason). Of course, we also won't use the cache if we're not checking for the current user - but better safe than
+  sorry. The values are collections of rows of DataPermissions."
+  (delay nil))
 
 (defmacro with-relevant-permissions-for-user
   "Populates the `*permissions-for-user*` dynamic var for use by the cache-aware functions in this namespace."
   [user-id & body]
-  `(binding [*permissions-for-user* (relevant-permissions-for-user ~user-id)]
+  `(binding [*permissions-for-user* (delay (relevant-permissions-for-user ~user-id))]
      ~@body))
 
-(defn- use-cached-values? [user-id]
-  (and (= user-id api/*current-user-id*) (seq *permissions-for-user*)))
+(defn- get-permissions [user-id perm-type db-id]
+  (if-let [cached-perms (and (= user-id api/*current-user-id*)
+                             (get @*permissions-for-user* [user-id perm-type db-id]))]
+    cached-perms
+    (get (relevant-permissions-for-user user-id) [user-id perm-type db-id])))
 
 ;;; ---------------------------------------- Fetching a user's permissions --------------------------------------------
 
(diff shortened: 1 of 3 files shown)


test/metabase/models/data_permissions_test.clj
```diff
@@ -151,6 +165,7 @@
                  :model/Database                   {database-id :id} {}
                  :model/Table                      {table-id-1 :id}  {:db_id database-id}
                  :model/Table                      {table-id-2 :id}  {:db_id database-id}]
+    ;; for cache
```
noahmoss, line 168: don't understand this comment

src/metabase/models/data_permissions.clj
```diff
@@ -277,17 +292,9 @@
...
-                        :from [[:permissions_group_membership :pgm]]
-                        :join [[:permissions_group :pg] [:= :pg.id :pgm.group_id]
-                               [:data_permissions :p]   [:= :p.group_id :pg.id]]
-                        :where [:and
-                                [:= :pgm.user_id user-id]
-                                [:= :p.perm_type (u/qualified-name perm-type)]
-                                [:= :p.db_id database-id]]})]
+    (let [perm-values (->> (get-permissions user-id perm-type database-id)
+                             (map :perm_value)
+                             (into #{}))]
```
noahmoss, line 297:
```suggestion
    (let [perm-values (->> (get-permissions user-id perm-type database-id)
                           (map :perm_value)
                           (into #{}))]
```

noahmoss (approved):
I like this approach a lot! Refactoring out the SQL query for fetching perms is a very nice benefit too. Just a couple nits.

I guess the only remaining component is to hook this up to the session middleware? Were you planning on doing that in a separate PR?

test/metabase/models/data_permissions_test.clj
```diff
@@ -138,7 +139,20 @@
         (is (= :no (data-perms/database-permission-for-user user-id :perms/native-query-editing database-id-2))))
 
       (testing "Admins always have the most permissive value, regardless of group membership"
-        (is (= :yes (data-perms/database-permission-for-user (mt/user->id :crowberto) :perms/native-query-editing database-id-2)))))))
+        (is (= :yes (data-perms/database-permission-for-user (mt/user->id :crowberto) :perms/native-query-editing database-id-2)))))
+    (testing "caching works as expected"
```
noahmoss, line 143:
```suggestion
        (is (= :yes (data-perms/database-permission-for-user (mt/user->id :crowberto) :perms/native-query-editing database-id-2)))))
        
    (testing "caching works as expected"
```

test/metabase/models/data_permissions_test.clj
```diff
@@ -162,7 +177,20 @@
         (is (= :block (data-perms/table-permission-for-user user-id :perms/data-access database-id table-id-2))))
 
       (testing "Admins always have the most permissive value, regardless of group membership"
-        (is (= :unrestricted (data-perms/table-permission-for-user (mt/user->id :crowberto) :perms/data-access database-id table-id-2)))))))
+        (is (= :unrestricted (data-perms/table-permission-for-user (mt/user->id :crowberto) :perms/data-access database-id table-id-2)))))
+      (mt/with-restored-data-perms-for-groups! [group-id-1 group-id-2]
```
noahmoss, line 181:
```suggestion
        (is (= :unrestricted (data-perms/table-permission-for-user (mt/user->id :crowberto) :perms/data-access database-id table-id-2)))))
        
     (mt/with-restored-data-perms-for-groups! [group-id-1 group-id-2]
```

noahmoss: (looks like this was at the same indentation as the previous `testing` block, but the parens weren't balanced, or something...not sure exactly but you might want to check the indentation here)

test/metabase/models/data_permissions_test.clj
johnswanson: Doh, sorry - I started writing a comment, then moved things around and forgot to delete it.

johnswanson:
> I guess the only remaining component is to hook this up to the session middleware? Were you planning on doing that in a separate PR?

That was my plan, but actually, I might as well do it here - I'll add that.

johnswanson:
@noahmoss I added the data-perms cache to the session middleware in the latest commit - quick and easy 🎉 

It also made me think about a potential flaw in my implementation of the cache - specifically, if we don't get _any_ rows back for a given user ID (as might happen for a `nil` user id) we'd always hit the DB, because `(group-by (juxt :user_id :perm_type :db_id) [])` will return `{}`, and `(get {} [user-id perm-type db-id])` will return `nil`.

I modified things slightly to instead have a nested map with user-id keys. Then, to check whether the cache exists, we can just do `(get {} user-id)`, which will return a truthy value (`{}`) if we've cached the relevant rows (i.e. no rows) already. Does that seem reasonable?

noahmoss:
@johnswanson Sounds good!

PR #38741, opened 2024-02-13
[MLv2] `remove-field` throws given aggregation, breakout, or expression

### LINKED ISSUE #34321: [MLv2] Removing an aggregation from a query using remove-field corrupts query
[remote-field](https://github.com/metabase/metabase/blob/87cddad248833e2995bd535ca1a8c4e64dbc7145/src/metabase/lib/field.cljc#L643) appears to handle columns with `:source/aggregations`, but calls `exclude-field` which will always assoc the remaning fields in the `:fields` key of the query, instead of removing the column from `:aggregations`. Screenshots below:

Column to be removed, and resulting query: 
image

Metabase Lib function used: 
image

Error Returned when calling `returned-columns` on the returned query:
image

bshepherdson:
`metabase.lib.field/remove-field` is intended for manipulating the
`:fields` list. That list controls which fields from the source,
previous stage, explicit join, or implicit join should be included in
the results.

All aggregations, breakouts, and custom expressions are always included
in the return columns for that stage, so they cannot be removed with the
`:fields` clause. The UI does not attempt to do this, but this check is
a guardrail for future changes. Trying to hide an aggregation from the
`:fields` clause can silently create a broken query.

Fixes #34321.

### DIFF
--- a/src/metabase/lib/field.cljc
+++ b/src/metabase/lib/field.cljc
@@ -633,6 +633,11 @@
 (defn- native-query-fields-edit-error []
   (i18n/tru "Fields cannot be adjusted on native queries. Either edit the native query, or save this question and edit the fields in a GUI question based on this one."))
 
+(defn- source-clauses-only-fields-edit-error []
+  (i18n/tru (str "Only source columns (those from a table, model, or saved question) can be adjusted on a query. "
+                 "Aggregations, breakouts and expressions are always returned, and must be removed from the query or "
+                 "hidden in the UI.")))
+
 (mu/defn add-field :- ::lib.schema/query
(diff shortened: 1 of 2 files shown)


bshepherdson:
* `master`
  * **PR #38741** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38741?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38741?utm_source=stack-comment).

metamben (approved):
👍

PR #38743, opened 2024-02-13
Remove all empty stages instead of just the last one

### LINKED ISSUE #37626: [MLv2] [BE] Change `dropStageIfEmpty` into `dropEmptyStages`
See [Slack discussion](https://metaboat.slack.com/archives/C0645JP1W81/p1705049044385729).
Complementary FE task: #37627

----

FE needs to be able to drop intermediate empty query stages, but `dropStageIfEmpty` works only with the last stage.

Up until now, FE incorrectly assumed that `dropStage` and `dropStageIfEmpty` accept `stageIndex` argument, but in BE these functions actually always work only with the last stage.

This `dropStageIfEmpty` function should be changed, so that it drops all intermediate empty stages as well.
Once this happens, this function should be also renamed, to e.g. `dropEmptyStages`.

metamben:
Fixes #37626.

A few points of interest:

- A JS function called `dropEmptyStages` has already been implemented, this PR replaces it's implementation to call the CLJS version.
- Calls to `dropStageIfEmpty` have been replaced by calls to `dropEmptyStages`, which is moderately fishy, but that's how I understand the goal.
- The name `dropEmptyStages` can be misleading: this function removes empty stages from the middle of the list of stages too.


frontend/src/metabase/querying/components/FilterModal/utils/filters.ts
```diff
@@ -15,8 +15,8 @@ export function appendStageIfAggregated(query: Lib.Query) {
     : query;
 }
 
-export function dropStageIfEmpty(query: Lib.Query) {
-  return Lib.dropStageIfEmpty(query, -1);
+export function dropEmptyStages(query: Lib.Query) {
```
kamilmielnik, line 18: We should inline this function as it no longer serves any purpose

frontend/src/metabase/querying/components/FilterPanel/utils.ts
```diff
@@ -11,6 +11,6 @@ export function getFilterItems(query: Lib.Query): FilterItem[] {
   });
 }
 
-export function dropStageIfEmpty(query: Lib.Query): Lib.Query {
-  return Lib.dropStageIfEmpty(query, -1);
+export function dropEmptyStages(query: Lib.Query): Lib.Query {
```
kamilmielnik, line 14: We should inline this function as it no longer serves any purpose

kamilmielnik (approved):
FE code looks good :+1: 

Left 2 minor comments.

ranquild (approved):

src/metabase/lib/stage.cljc
```diff
@@ -346,12 +346,22 @@
...
+  "Drops all empty stages in the pipeline."
   [query :- ::lib.schema/query]
-  (if-not (has-clauses? query -1)
-    (drop-stage query)
-    query))
+  (reduce (fn [query stage-number]
+            (cond-> query
+              (not (has-clauses? query stage-number)) (remove-stage stage-number)))
+          query
+          (range (lib.util/canonical-stage-index query -1) 0 -1)))
```
bshepherdson, line 364: I'm not sure about this structure. It seems simpler to have an inner function that removes the last stage, empty or not (or the specified stage I guess) and another that calls it in a `loop` until `has-clauses?` is true for the last stage.

bshepherdson (approved):
I think we can find an easier to understand implementation of `drop-empty-stages`.

Also I agree with Kamil about the two needless wrappers in the TS code.

frontend/src/metabase/querying/components/FilterModal/utils/filters.ts
metamben: done

frontend/src/metabase/querying/components/FilterPanel/utils.ts
metamben: done

bshepherdson (approved):

PR #38751, opened 2024-02-14
Fix update hook and add select hook

crisptrutski:

### Description

Fix holes in the mutation hooks and coerce queries to filter out bad data that's already gotten into the db.

I really wanted to protect against `select-one` queries that don't full specify the identity, but I couldn't find a way to make the hook dependent on the type of select.

### DIFF
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -111,7 +111,7 @@
     (when-not (contains? field-values-types type)
       (throw (ex-info (tru "Invalid field-values type.")
                       {:type        type
-                       :stauts-code 400})))
+                       :status-code 400})))
 
     (when (and (= type :full)
                hash_key)
@@ -142,7 +142,8 @@
   (u/prog1 (merge {:type :full}
                   field-values)
     (assert-valid-human-readable-values field-values)
-    (assert-valid-type-hash-combo field-values)
+    ;; Take into account that :type has a default value in the database
+    (assert-valid-type-hash-combo (update field-values :type #(or % :full)))
     ;; if inserting a new full fieldvalues, make sure all the advanced field-values of this field are deleted
     (when (= (:type <>) :full)
       (clear-advanced-field-values-for-field! field_id))))
@@ -158,8 +159,7 @@
                          :hash_key    hash_key
                          :status-code 400})))
       ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
-      (when (and (contains? field-values :values)
-                 (= :full (:type (t2/select-one FieldValues (:id field-values)))))
+      (when (and (contains? field-values :values) (= :full (:type field-values)))
         (clear-advanced-field-values-for-field! (:field_id field-values))))))
 
 (t2/define-before-select :model/FieldValues
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -124,15 +124,14 @@
                     {:type        type
                      :status-code 400}))))
 
-(defn- assert-no-identity-changes [field-values]
-  (let [changes (t2/changes field-values)]
-    (when (some #(contains? changes %) [:field_id :type :hash_key])
-      (throw (ex-info (tru "Can't update field_id, type, or hash_key for a FieldValues.")
-                      {:id          (:id field-values)
-                       :field_id    (:field_id changes)
-                       :type        (:type changes)
-                       :hash_key    (:hash_key changes)
-                       :status-code 400})))))
+(defn- assert-no-identity-changes [id changes]
+  (when (some #(contains? changes %) [:field_id :type :hash_key])
+    (throw (ex-info (tru "Can't update field_id, type, or hash_key for a FieldValues.")
+                    {:id          id
+                     :field_id    (:field_id changes)
+                     :type        (:type changes)
+                     :hash_key    (:hash_key changes)
+                     :status-code 400}))))
 
 (defn clear-advanced-field-values-for-field!
   "Remove all advanced FieldValues for a `field-or-id`."
@@ -156,12 +155,13 @@
 
 (t2/define-before-update :model/FieldValues
   [field-values]
-  (u/prog1 (update field-values :type #(keyword (or % :full)))
-    (assert-no-identity-changes <>)
-    (assert-valid-human-readable-values field-values)
-    ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
-    (when (and (contains? field-values :values) (= :full (:type <>)))
-      (clear-advanced-field-values-for-field! (:field_id field-values)))))
+  (let [changes (t2/changes field-values)]
+    (u/prog1 (update field-values :type #(keyword (or % :full)))
+      (assert-no-identity-changes (:id field-values) changes)
+      (assert-valid-human-readable-values field-values)
+      ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
+      (when (and (contains? changes :values) (= :full (:type <>)))
+        (clear-advanced-field-values-for-field! (:field_id field-values))))))
 
 (defn- assert-coherent-query [{:keys [type hash_key] :as field-values}]
   (cond
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -179,7 +179,7 @@
       (throw (ex-info "Invalid query - Advanced FieldValues must have a hash_key"
                       {:field-values field-values})))))
 
-(defn- add-defensive-type-mismatch-filter [{:keys [type] :as field-values}]
+(defn- add-mismatched-hash-filter [{:keys [type] :as field-values}]
   (cond
     (= :full type) (assoc field-values :hash_key nil)
     (some? type)   (update field-values :hash_key #(or % [:not= nil]))
@@ -188,8 +188,7 @@
 (t2/define-before-select :model/FieldValues
   [{:keys [kv-args] :as query}]
   (assert-coherent-query kv-args)
-  ;; Ensure that query filters out invalid rows
-  (update query :kv-args add-defensive-type-mismatch-filter))
+  (update query :kv-args add-mismatched-hash-filter))
 
 (t2/define-after-select :model/FieldValues
   [field-values]
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -176,7 +176,7 @@
                       {:field-values field-values})))
 
     (and (contains? field-values :hash_key) (nil? hash_key))
-    (throw (ex-info "Invalid query - Advanced FieldValues can only specify a non-nil hash_key"
+    (throw (ex-info "Invalid query - Advanced FieldValues can only specify a non-empty hash_key"
                     {:field-values field-values}))))
 
 (defn- add-mismatched-hash-filter [{:keys [type] :as field-values}]
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -180,9 +180,14 @@
                       {:field-values field-values})))))
 
 (t2/define-before-select :model/FieldValues
-  [query]
-  (u/prog1 query
-    (assert-coherent-query (:kv-args query))))
+  [{:keys [kv-args] :as query}]
+  (assert-coherent-query kv-args)
+  ;; Ensure that query filters out invalid rows
+  (let [type (:type kv-args)]
+    (cond
+      (= :full type) (assoc-in query [:kv-args :hash_key] nil)
+      (some? type)   (update-in query [:kv-args :hash_key] #(or % [:not= nil]))
+      :else          query)))
 
 (t2/define-after-select :model/FieldValues
   [field-values]
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -179,15 +179,17 @@
       (throw (ex-info "Invalid query - Advanced FieldValues must have a hash_key"
                       {:field-values field-values})))))
 
+(defn- add-defensive-type-mismatch-filter [{:keys [type] :as field-values}]
+  (cond
+    (= :full type) (assoc field-values :hash_key nil)
+    (some? type)   (update field-values :hash_key #(or % [:not= nil]))
+    :else          field-values))
+
 (t2/define-before-select :model/FieldValues
   [{:keys [kv-args] :as query}]
   (assert-coherent-query kv-args)
   ;; Ensure that query filters out invalid rows
-  (let [type (:type kv-args)]
-    (cond
-      (= :full type) (assoc-in query [:kv-args :hash_key] nil)
-      (some? type)   (update-in query [:kv-args :hash_key] #(or % [:not= nil]))
-      :else          query)))
+  (update query :kv-args add-defensive-type-mismatch-filter))
 
 (t2/define-after-select :model/FieldValues
   [field-values]
--- a/src/metabase/models/field_values.clj
+++ b/src/metabase/models/field_values.clj
@@ -105,39 +105,34 @@
                      :status-code           400}))))
 
 (defn- assert-valid-type-hash-combo
-  "If type is present, ensure that it is valid, and that a hash_key is provided iff this is an advanced field type."
+  "Ensure that type is present, valid, and that a hash_key is provided iff this is an advanced field type."
   [{:keys [type hash_key] :as _field-values}]
-  (let [type (keyword (or type :full))]
-    (when-not (contains? field-values-types type)
-      (throw (ex-info (tru "Invalid field-values type.")
-                      {:type        type
-                       :status-code 400})))
+  (when-not (contains? field-values-types type)
+    (throw (ex-info (tru "Invalid field-values type.")
+                    {:type        type
+                     :status-code 400})))
 
-    (when (and (= type :full)
-               hash_key)
-      (throw (ex-info (tru "Full FieldValues shouldn't have hash_key.")
-                      {:type        type
-                       :hash_key    hash_key
-                       :status-code 400})))
+  (when (and (= type :full)
+             hash_key)
+    (throw (ex-info (tru "Full FieldValues shouldn't have hash_key.")
+                    {:type        type
+                     :hash_key    hash_key
+                     :status-code 400})))
 
-    (when (and (advanced-field-values-types type)
-               (empty? hash_key))
-      (throw (ex-info (tru "Advanced FieldValues requires a hash_key.")
-                      {:type        type
-                       :status-code 400})))))
+  (when (and (advanced-field-values-types type)
+             (empty? hash_key))
+    (throw (ex-info (tru "Advanced FieldValues requires a hash_key.")
+                    {:type        type
+                     :status-code 400}))))
 
 (defn- assert-no-identity-changes [field-values]
-  (let [{:keys [field_id] :as changes} (t2/changes field-values)]
-    (when (some #(contains? changes %) [:type :hash_key])
-      (throw (ex-info (tru "Can't update type or hash_key for a FieldValues.")
+  (let [changes (t2/changes field-values)]
+    (when (some #(contains? changes %) [:field_id :type :hash_key])
+      (throw (ex-info (tru "Can't update field_id, type, or hash_key for a FieldValues.")
                       {:id          (:id field-values)
+                       :field_id    (:field_id changes)
                        :type        (:type changes)
                        :hash_key    (:hash_key changes)
-                       :status-code 400})))
-    (when (some? field_id)
-      (throw (ex-info (tru "Can't update field_id for a FieldValues.")
-                      {:id          (:id field-values)
-                       :field_id    field_id
                        :status-code 400})))))
 
 (defn clear-advanced-field-values-for-field!
@@ -153,20 +148,20 @@
 
 (t2/define-before-insert :model/FieldValues
   [{:keys [field_id] :as field-values}]
-  (u/prog1 (merge {:type :full} field-values)
+  (u/prog1 (update field-values :type #(keyword (or % :full)))
     (assert-valid-human-readable-values field-values)
     (assert-valid-type-hash-combo <>)
     ;; if inserting a new full fieldvalues, make sure all the advanced field-values of this field are deleted
-    (when (= :full (keyword (:type <>)))
+    (when (= :full (:type <>))
       (clear-advanced-field-values-for-field! field_id))))
 
 (t2/define-before-update :model/FieldValues
   [field-values]
-  (u/prog1 field-values
-    (assert-no-identity-changes field-values)
+  (u/prog1 (update field-values :type #(keyword (or % :full)))
+    (assert-no-identity-changes <>)
     (assert-valid-human-readable-values field-values)
     ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
-    (when (and (contains? field-values :values) (= :full (keyword (:type field-values))))
+    (when (and (contains? field-values :values) (= :full (:type <>)))
       (clear-advanced-field-values-for-field! (:field_id field-values)))))
 
 (defn- assert-coherent-query [{:keys [type hash_key] :or {type :full} :as field-values}]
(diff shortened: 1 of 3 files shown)


src/metabase/models/field_values.clj
```diff
@@ -138,28 +153,42 @@
...
 (t2/define-before-insert :model/FieldValues
   [{:keys [field_id] :as field-values}]
-  (u/prog1 (merge {:type :full}
-                  field-values)
+  (u/prog1 (merge {:type :full} field-values)
     (assert-valid-human-readable-values field-values)
-    (assert-valid-field-values-type field-values)
-    ;; if inserting a new full fieldvalues, make sure all the advanced field-values of this field is deleted
-    (when (= (:type <>) :full)
+    (assert-valid-type-hash-combo field-values)
```
calherries, line 158:
```suggestion
    (assert-valid-type-hash-combo <>)
```

src/metabase/models/field_values.clj
```diff
@@ -104,13 +104,14 @@
                     {:human-readable-values human-readable-values
                      :status-code           400}))))
 
-(defn- assert-valid-field-values-type
+(defn- assert-valid-type-hash-combo
+  "If type is present, ensure that it is valid, and that a hash_key is provided iff this is an advanced field type."
   [{:keys [type hash_key] :as _field-values}]
-  (when type
+  (let [type (keyword (or type :full))]
```
calherries, line 110: This logic is coupled to line 156. But you don't need this anymore following https://github.com/metabase/metabase/pull/38751/files#r1490613168
```suggestion
```

src/metabase/models/field_values.clj
```diff
@@ -125,6 +126,20 @@
...
+      (throw (ex-info (tru "Can't update type or hash_key for a FieldValues.")
+                      {:id          (:id field-values)
+                       :type        (:type changes)
+                       :hash_key    (:hash_key changes)
+                       :status-code 400})))
+    (when (some? field_id)
+      (throw (ex-info (tru "Can't update field_id for a FieldValues.")
+                      {:id          (:id field-values)
+                       :field_id    field_id
+                       :status-code 400})))))
```
calherries, line 141: this looks like it's almost the same thing in two ways, which seems arbitrary. why not combine them into one like this:
```suggestion
    (when (some #(contains? changes %) [:type :hash_key])
      (throw (ex-info (tru "Can't update type, hash_key, or field_id for a FieldValues.")
                      {:id          (:id field-values)
                       :type        (:type changes)
                       :hash_key    (:hash_key changes)
                       :field_id    (:field_id changes)
                       :status-code 400})))))
```

src/metabase/models/field_values.clj
```diff
@@ -138,28 +153,42 @@
...
-  (u/prog1 (merge {:type :full}
-                  field-values)
+  (u/prog1 (merge {:type :full} field-values)
     (assert-valid-human-readable-values field-values)
-    (assert-valid-field-values-type field-values)
-    ;; if inserting a new full fieldvalues, make sure all the advanced field-values of this field is deleted
-    (when (= (:type <>) :full)
+    (assert-valid-type-hash-combo field-values)
+    ;; if inserting a new full fieldvalues, make sure all the advanced field-values of this field are deleted
+    (when (= :full (keyword (:type <>)))
```
calherries, line 160: Was this a bug? should it have a test?

src/metabase/models/field_values.clj
```diff
@@ -138,28 +153,42 @@
...
-                         :status-code 400})))
-      ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
-      (when (and values
-                 (= (:type field-values) :full))
-        (clear-advanced-field-values-for-field! (:field_id field-values))))))
+  (u/prog1 field-values
+    (assert-no-identity-changes field-values)
+    (assert-valid-human-readable-values field-values)
+    ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
+    (when (and (contains? field-values :values) (= :full (keyword (:type field-values))))
```
calherries, line 169: Similar to https://github.com/metabase/metabase/pull/38751/files#r1490642171, does this have a test?

src/metabase/models/field_values.clj
```diff
@@ -138,28 +153,42 @@
...
+
+(defn- assert-coherent-query [{:keys [type hash_key] :or {type :full} :as field-values}]
+  (if (= :full (keyword type))
+    (when (some? hash_key)
+      (throw (ex-info "Invalid query - :full FieldValues cannot have a hash_key"
+                      {:field-values field-values})))
+    (when (and (contains? field-values :hash_key)
+               (nil? hash_key))
+      (throw (ex-info "Invalid query - Advanced FieldValues must have a hash_key"
+                      {:field-values field-values})))))
```
calherries, line 180: This logic is duplicated on https://github.com/metabase/metabase/pull/38751/files#diff-b28eaae7ad5ee9389359e276833936958b869f4e239e8b2c4adb76078562bed9R116-R127

perhaps `assert-valid-type-hash-combo` should be used in the `t2/define-before-select` as well?

src/metabase/models/field_values.clj
```diff
@@ -138,28 +153,42 @@
...
-                 (= (:type field-values) :full))
-        (clear-advanced-field-values-for-field! (:field_id field-values))))))
+  (u/prog1 field-values
+    (assert-no-identity-changes field-values)
+    (assert-valid-human-readable-values field-values)
+    ;; if we're updating the values of a Full FieldValues, delete all Advanced FieldValues of this field
+    (when (and (contains? field-values :values) (= :full (keyword (:type field-values))))
+      (clear-advanced-field-values-for-field! (:field_id field-values)))))
+
+(defn- assert-coherent-query [{:keys [type hash_key] :or {type :full} :as field-values}]
```
calherries, line 172: Why did you decide that `type` should default to `:full` here?

calherries (changes requested):
This catches a lot of ways to misuse fieldvalues with toucan, which is great. Unfortunately the one sharp edge it doesn't solve is the original point Cam had in [his comment](https://github.com/metabase/metabase/issues/668#issuecomment-1373335680).

> I think maybe it makes sense to move the cached field values into their own table. Otherwise I can't think of any real way to prevent people from shooting themselves in the foot by not knowing you're always supposed to filter by :type :full or whatever.

It's still possible to write `(t2/select-one :model/FieldValues :field_id 1)` and accidentally get an advanced FieldValues instance when you thought you were getting a `type=full` one all the time.

The important question is, will we ever want to select field values without specifying a  `:type` like `(t2/select :model/FieldValues :field_id 1)`? I can't imagine why we'd want to do that, unless we were doing it for auditing purposes or something unrelated to the value FieldValues provides, which is essentially caching. So it's fine to make `:type` required in all instances of selecting `FieldValues` I think, and eliminate this sharp edge.

src/metabase/models/field_values.clj
crisptrutski: Yip, added test

src/metabase/models/field_values.clj
crisptrutski: I initially went that route of reusing it, but eventually decided against the guard that Cam suggested and you referenced, will discuss further on your summary comment.

src/metabase/models/field_values.clj
crisptrutski: The field is non-nullable in the database, so we can't hit this edge case in practice or test it via `update!`. I added this before realising that the update hook gets a fully populated Toucan instance, not the map passed by the caller.

crisptrutski:
> This catches a lot of ways to misuse fieldvalues with toucan, which is great. Unfortunately the one sharp edge it doesn't solve is the original point Cam had .. The important question is, will we ever want to select field values without specifying a `:type` ... I can't imagine why we'd want to do that ... So it's fine to make `:type` required in all instances of selecting `FieldValues` I think, and eliminate this sharp edge.

I couldn't convince myself that we don't and never will have a use case for selecting *all* the field values, for example to warm an in-memory cache. I expecially didn't want to take on the risk of having missed an existing query either. I alluded to this in the PR description:

> I really wanted to protect against select-one queries that don't specify the full `(field_id, type, hash_key)` identity, but I couldn't find a way to make the hook dependent on the type of select.

I even though about only adding the check if field_id is present, but it really felt like it was overstepping the mark and being too opinionated. Logic that belongs in the application layer and not the model layer. Since in practice we access these values through "upsert" style patterns using methods one level up from toucan, I don't see a pressing need for it.

If feelings are especially strong here, we could do a careful audit to make sure it's currently safe, but I'd prefer to get these "slam dunk" validations in separately.

src/metabase/models/field_values.clj
crisptrutski: Oh, I remembered another reason I felt uncomfortable reusing it, the format of the exceptions. Having a translated error message and status code felt strange for an error that can be triggered outside of web requests. I really dislike how we're peppering presentation logic in the model layer - violating two layers of abstraction 😢

src/metabase/models/field_values.clj
crisptrutski: Is this a firm convention? For non-test code the approach I take is to put the primitive expression first since it's much easier to parse. With the only layout you need to count brackets.

calherries:
> Logic that belongs in the application layer and not the model layer.

You're right on this, but Cam's concern and mine is that there's not actually a strong split between these two things, or not strong enough to stop people using `t2/select` in application code instead of the upsert functions in the field values model namespace. Because of this, the toucan functions are effectively the interface that the application code uses to interact with models. We want that interface to be restricted so application code can't easily make mistakes. That's why Cam made that suggestion to split FieldValues into separate tables.

If we want to separate model logic from application logic, we need some way to make toucan functions on models "private", and not allow application code to call them. We don't have that capability yet, but I would like to think about it for the future. At the moment they're unfortunately intertwined.

And you are right in that there is some hypothetical use case in the future where we might need to select all types of field values, but it still feels like a no brainer to lock this down as long as we don't need that flexibility.

calherries:
I found another instance where `t2/select` was used with `FieldValues` without a ':type`. https://github.com/metabase/metabase/blob/4d552f131f509ad1aacaea379edb5c7fdc699601/src/metabase/models/field.clj#L193-L196

This function is only used in for legacy v1 serialization, I traced it back to [here](https://github.com/metabase/metabase/blob/4d552f131f509ad1aacaea379edb5c7fdc699601/src/metabase/cmd.clj#L193-L195) thankfully, which is deprecated. But it exemplifies the kind of bug we're hoping to avoid in the future.

src/metabase/models/field_values.clj
calherries: It's probably not a firm convention but I just find it slightly surprising otherwise. I hardly ever see non-test code that does it this way. More than two brackets is rare to see anyway, so maybe your point (which is valid) doesn't really matter that much.

crisptrutski: Just to make sure I understand the convention in your head, does this capture it correctly?

"When comparing an expression that depends on some variables, to a constant expression, put the context expression last"

src/metabase/models/field_values.clj
calherries: Oh right. In that case I think `assert-valid-type-hash-combo` can be replaced by `assert-coherent-query`. I don't see why we need `status-code` and `tru` at this level, so maybe it's a mistake to leave them here. If the app throws these errors during an API request it's definitely unexpected and not an issue with an API call.

src/metabase/models/field_values.clj
calherries: > the update hook gets a fully populated Toucan instance

Isn't this change introducing a bug then? `(contains? field-values :values)` will always be `true`, even if we're not updating the `values` field

src/metabase/models/field_values.clj
calherries: yeah

qnkhuat:
Food for thought: toucan2 has a nice feature in that the model is hierarchical. see https://github.com/camsaul/toucan2?tab=readme-ov-file#define-behavior-after-selecting-something

in this case you can define 2 new models
```
(derive :model/FullFieldValues :model/FieldValues)
(derive :model/AdvancedFieldValues :model/FieldValues)
```

Then you can define 2 new `before-select` to make sure each query to these new models has the proper `:type.`
this way we don't have to split the tables and reuse all the shared code for `FieldValues`

calherries:
Hmm, a hierarchy seems like it introduces more room for error, and actually makes changes harder. Like it would allow `t2/define-before-insert` overrides, requires creating clj-kondo rules to stop people using the parent model, and more validation to check that `:model/FullFieldValues` wasn't being used with columns it shouldn't. Another downside is it becomes a lot harder to operate on multiple types of field values from the one call site, which Chris rightly points out we might want to do in the future.

crisptrutski:
> You're right [about app and model layers separation] , but ... there's not actually a strong split between these two things, ... That's also why I see get-latest-field-values and get-latest-full-field-values as *temporary workarounds* until we get the constraints in. 

If we agree there should be clearer separation, why would these be temporary? I don't see any reason to remove them once we have database constraints, they're part of the making a clear abstraction layer (i.e. the field-values namespaces can use toucan, the rest of the app can use the namespace). The methods you mention in fact are themselves fairly internal, a better interface would also encapsulate lazy creation.

> If we want to separate model logic from application logic, we need some way to make toucan functions on models "private", and not allow application code to call them. We don't have that capability yet, but I would like to think about it for the future. At the moment they're unfortunately intertwined.

My first response to this was probably knee jerk and not speaking to your actual position, but I see this more as a cultural hurdle than a technical one. Clojure's design advocates judgment over guard rails - you can break any rule, anywhere, but a mix of socialized conventions and good ergonomics for "doing the right thing" is enough for most systems.

We've gone quite far down the road of using linters, and for all their limitations they're pretty nifty for enforcing subtle things. Having allow lists for which namespaces can use which models is perhaps a linter driven mechanism which has legs. A nice aspect is that one can easily and explicitly opt out of the guardrail when necessary, whereas there is no way to override logic put into a hook. 

I've never seen linters uses as extensively as we do before, it's be mostly a very positive experience, and I'm still getting a feel for good patterns, and their pros and cons. I don't think they fully replace judgment and shouldn't become too much of a crutch, and certainly not a requirement.

> That's why Cam made that suggestion to split FieldValues into separate tables.

This argument alone doesn't convince me, and I've explained why elsewhere. Thankfully for peace and harmony though, I do think splitting the tables is a great idea for a difference reason - the provenance of the human readable labels. For full instances they are the authoritative store of user provided data, and for advanced instances they are a mirrored copy that needs to be kept in sync. We could even use a foreign key constraint between the full and advanced keys to make sure we don't have purging bugs.

That said, I see splitting the table as a rather intrusive, risky, expensive, and low value tech debt item.

> And you are right in that there is some hypothetical use case in the future where we might need to select all types of field values, but it still feels like a no brainer to lock this down as long as we don't need that flexibility.

To rehash and summarize my take:

1. There's a qualitative difference between detecting "definitely broken" and "probably not what you intended" queries.
2. When selecting a single row, we could definitely be more opinionated. Toucan does not currently support that.
3. Auditing for more bugs, fixing them - great. I think that's out of scope for the parent issue, was not on my agenda.
4. We could add the extra validation if:
  - We are 100% confident that we haven't missed an accidental under defined queries
  - There truly were no valid use cases
  - We're bought in that putting logic in the "wrong" later is the lesser of two evils. (not my opinion, but can accept it)
 5. Adding an abstraction above the model is probably better long term, and probably not much more. Low risk.

I'm not keen on scope creep and unavoidable risks.

qnkhuat:
> Like it would allow t2/define-before-x overrides

you don't override methods, because toucan2 methods are chained.

```(t2/define-before-select :model/FullFieldValues
  [arg]
  (assoc arg :type :full))

(t2/define-before-select :model/FieldValues
  [arg]
  (println "HI")
  arg)

(t2/select-one :model/FullFieldValues)
;; this will trigger both `before-select`, the most specific is trigger first, meaning `FullFieldValues` is triggered first.
```

crisptrutski:
The hierarchy idea is interesting, in a way that's an iterative step towards splitting the tables.

We could have `FullFieldValues` and `AdvancedFieldValues` derive from `MixedFieldValues`, and allow mutation only on the leaf models. I quite like that, but don't know how Toucan handles them, for example can you "override" hooks or only add to them?

This has been a wider discussion than I anticipated, and there's a lot to think about. I worry that there are a lot of threads now between lots of issues and PRs, past and present. Perhaps we could pivot to a Notion doc?

calherries:
@qnkhuat you don't override methods, because toucan2 methods are chained.

cool! didn't know that. Still feels more complicated to me.

> you can still do (t2/select :model/FieldValues) if you want to

`:model/FieldValues` won't have the variant-specific hooks implemented (handling type=full and type!=full differently), so you'll have to implement the hooks again for `:model/FieldValues` again. That would work but still a downside

calherries:
@crisptrutski btw before I respond to everything in your comment, I agree that locking down the select clause can be merged separately. This PR is a cool improvement and I'd stamp it if you can resolve my unresolved comments

calherries:
I'm not going to reply to every sentence in your comment for concerns of our time, but I will say 

> linters

Yes 100% that would solve the concerns we're trying to solve here. As would runtime checks on the toucan hooks. I guess I'm arguing to at least pick one.

> When selecting a single row, we could definitely be more opinionated. Toucan does not currently support that.

I just showed you a bug where we were selecting more than one row, and still getting it wrong. Applying it in both cases would have solved it, so I'm not convinced we should only solve it for the single row case.

> Auditing for more bugs, fixing them - great. I think that's out of scope for the parent issue, was not on my agenda.

Sure. I brought it up to make my point that safety on `t2/select` is valuable.

> We're bought in that putting logic in the "wrong" later is the lesser of two evils. (not my opinion, but can accept it)

Maybe I'm missing something obvious, I'm not following this

> We are 100% confident that we haven't missed an accidental under defined queries

Can't argue with that. Maybe we should have lint rules for toucan select instead of runtime validation I'm suggesting here? This would bring us close to 100% confidence we weren't introducing any more bugs.

calherries:
> Perhaps we could pivot to a Notion doc?

Yeah I think we can close this PR and put the wider discussion in a Notion doc. There's some fundamental disagreements here that I'd like to resolve/keep around for reference

calherries:
Open comments for viz:
https://github.com/metabase/metabase/pull/38751#discussion_r1490811400
https://github.com/metabase/metabase/pull/38751#discussion_r1490817853

src/metabase/models/field_values.clj
crisptrutski: Ah yeah, thanks for spotting the regression. I think this could have hurt performance quite a bit because we keep track of the last accessed time on these records, and would have blown away all the derived fields quite often.

Will add a failing test before fixing.

src/metabase/models/field_values.clj
calherries: I don't see how `field/:id/values` can hit it, but fair point on not wanting to introduce behaviour changes. I still think there's a way to dedupe or couple them explicitly

crisptrutski: Just to make sure I'm not missing something silly - that API end-point can hit insert, which will hit `assert-valid-type-hash-combo`, and throw those status codes. I didn't see anything that would catch and translate them within the handler.

calherries: Ok, you're talking about this right? 

https://github.com/metabase/metabase/blob/4d552f131f509ad1aacaea379edb5c7fdc699601/src/metabase/api/field.clj#L323

`update-map` can only contain `:values` and `:human_readable_values`.

Then `create-field-values!` adds `type=full` and the `field_id`:
https://github.com/metabase/metabase/blob/4d552f131f509ad1aacaea379edb5c7fdc699601/src/metabase/api/field.clj#L304

`assert-valid-type-hash-combo` will never throw under these conditions.

crisptrutski: I was talking about the GET, which follows a slightly more tangled path and ends up somewhere different. In both cases the low level method hard codes the type and can't be passed a hash key. All I was say is that we reach the code, not that the current path is able to pass bad data. 

In practice I don't know if any APIs can trigger the exception, which I see was your point. My stance is that the scope is the contract, as finer details can be missed or change. I don't know anything about how our middleware or frontend handle errors yet, so I'm an extreme wimp in this area.

crisptrutski:
I saw a recent PR where Sanya bypassed hooks by using an simple keyword without the model namespace. I didn't realize it was that easy to opt out of the hooks without changing much code, I guess the only bother is losing the hydration. A big part of my reluctance was the idea that bypassing hooks would be very painful.

Creating another model also looks really promising, even if we avoid a hierarchy. What I really like about having separate models, is that we can go case-by-case migrating from laissez-faire to semantically precise.

I guess in short there looks like a good path to using a hook rather than a linter to catch bad usage. We could use linters to ensure that `select-one` uses a strict, full triplet asserting model. 

Linters seem quite addictive, so flexible in their scope, and infinite ways you could use them - but should you? I wish we'd actually made a session to talk about our learnings and best practices with clj-kondo at the offsite, we could also touch on spec and malli as well. The wit of the staircase.

src/metabase/models/field_values.clj
calherries: Right yeah I was talking about in practice - whether we would be changing any observable behaviour of the BE by removing them.

Anyway, what I'd suggest here:
1. leave a FIXME comment or something in `assert-valid-type-hash-combo` because we potentially found some code that could be deleted
2. I was torn about deduping `assert-valid-type-hash-combo` and `assert-coherent-query` for the reasons you said (why should select and insert be coupled like this), but I think they do contain some duplicated logic: `type` and `hash_key` and are coupled components of a unique identity for an advanced field values row. You shouldn't be able to select or insert part of the identity (e.g. only `hash_key`). So `assert-valid-type-hash-combo` is probably the correct name for such a function that checks the identity is whole, and it should be used in both the select and insert hook. I think it should be wrapped in a `(when type ...` too, as it was before.

crisptrutski: Looking closer, I can't de-duplicate the logic because it's not quite the same right now, although perhaps they will be if select validation is made more opinionated. The difference right now is that advanced field values query can omit the hash_key.

calherries: Right. What's the downside of making select validation require the `hash_key` if `type!=full`?

calherries: But you're right, I guess it's fair to leave it deduped. I'm not strongly convinced they're inherently coupled

calherries:
As per our call, I think it makes sense to follow this PR by splitting `:model/FieldValues` into two models: advanced and full.

Just this comment left to resolve https://github.com/metabase/metabase/pull/38751#discussion_r1490817853.

qnkhuat:
cool! Finally, we have a use case for derived models :)

calherries:
ah to be clear, I didn't suggest we should use derived models. Here we can use two models based on the same table. We discussed something like this:

```
(methodical/defmethod t2/table-name :model/FullFieldValues [_model] :metabase_fieldvalues)
(methodical/defmethod t2/table-name :model/AdvancedFieldValues [_model] :metabase_fieldvalues)
```

calherries (approved):
Sweet

calherries:
I have a few thoughts from reviewing this I hadn't considered before:
- maybe we should look into writing malli specs for toucan hook args
- for some models, it could make sense to lint away the use of toucan select/insert/update outside of model namespaces and so the model can expose more ergonomic and safer wrappers
- pointing two models to the same table could be a nice solution in some cases

calherries:
You have a failing test though

test/metabase/models/field_values_test.clj
```diff
@@ -307,11 +307,13 @@
       (doseq [[id update-map] [[sandbox-id {:field_id 1}]
                                [sandbox-id {:type :full}]
                                [sandbox-id {:type nil}]
+                               ;; this one should be ok, but toucan doesn't give the hook enough info to know better
```
calherries, line 310: Why should it be ok? `:type` is non-nullable in the database so this error is redundant, but it still seems like the correct behaviour to me

crisptrutski: The database has "full" as the default value, so this would be a no-op.

crisptrutski: Also the hook inserts an explicit "full" for redundancy / avoiding dependence on storage detail.

PR #38760, opened 2024-02-14
Create X-or-insert! methods and use them for FieldValues

crisptrutski:
### Description

This is a recreate version of https://github.com/metabase/metabase/pull/38430, since that merged unintentionally due to a forgotten auto-merge set a week before.

**Version 1:** Use a database transaction around the updates. 

*Problem: our default transactions aren't serializable.*

**Version 2:** Use a serializable transaction. 

*Problem: our tests use nested transactions. For some databases their notion of serializable is also still too weak.*

**Version 3:** Give up on atomicity without a database constraint.

This change has been decreased in scope to just improve ergonomics, and improve constraint forward compatibility.

### DIFF
--- a/src/metabase/api/field.clj
+++ b/src/metabase/api/field.clj
@@ -312,9 +312,9 @@
           update-map             {:values                (map first value-pairs)
                                   :human_readable_values (when human-readable-values?
                                                            (map second value-pairs))}
-          updated-pks            (mdb.u/update-or-insert! FieldValues {:field_id (u/the-id field), :type :full}
+          updated-pk             (mdb.u/update-or-insert! FieldValues {:field_id (u/the-id field), :type :full}
                                    (constantly update-map))]
-      (api/check-500 (pos? updated-pks))))
+      (api/check-500 (pos? updated-pk))))
   {:status :success})
 
 (api/defendpoint POST "/:id/rescan_values"
--- a/src/metabase/api/field.clj
+++ b/src/metabase/api/field.clj
@@ -312,10 +312,9 @@
           update-map             {:values                (map first value-pairs)
                                   :human_readable_values (when human-readable-values?
                                                            (map second value-pairs))}
-          updated-entity         (mdb.u/idempotent-upsert! FieldValues
-                                   {:field_id (u/the-id field), :type :full}
+          updated-pks            (mdb.u/update-or-insert! FieldValues {:field_id (u/the-id field), :type :full}
                                    (constantly update-map))]
-      (api/check-500 (pos? (:id updated-entity)))))
+      (api/check-500 (pos? updated-pks))))
   {:status :success})
 
 (api/defendpoint POST "/:id/rescan_values"
--- a/src/metabase/api/field.clj
+++ b/src/metabase/api/field.clj
@@ -4,6 +4,7 @@
    [compojure.core :refer [DELETE GET POST PUT]]
    [metabase.api.common :as api]
    [metabase.db.metadata-queries :as metadata-queries]
+   [metabase.db.util :as mdb.u]
    [metabase.lib.schema.metadata :as lib.schema.metadata]
    [metabase.models.dimension :refer [Dimension]]
    [metabase.models.field :as field :refer [Field]]
@@ -12,7 +13,7 @@
    [metabase.models.params.chain-filter :as chain-filter]
    [metabase.models.params.field-values :as params.field-values]
    [metabase.models.permissions :as perms]
-   [metabase.models.table :as table :refer [Table]]
+   [metabase.models.table :refer [Table]]
    [metabase.query-processor :as qp]
    [metabase.related :as related]
    [metabase.server.middleware.offset-paging :as mw.offset-paging]
@@ -314,13 +315,13 @@
       [400 (str "You can only update the human readable values of a mapped values of a Field whose value of "
                 "`has_field_values` is `list` or whose 'base_type' is 'type/Boolean'.")])
     (let [human-readable-values? (validate-human-readable-pairs value-pairs)
-          update-map  {:values                (map first value-pairs)
-                       :human_readable_values (when human-readable-values?
-                                                (map second value-pairs))}]
-      (t2/with-transaction [_conn]
-        (if-let [field-value-id (:id (field-values/get-latest-full-field-values id))]
-          (update-field-values! field-value-id update-map)
-          (create-field-values! field update-map)))))
+          update-map             {:values                (map first value-pairs)
+                                  :human_readable_values (when human-readable-values?
+                                                           (map second value-pairs))}]
+      (mdb.u/idempotent-insert!
+        (when-let [field-values-id (:id (field-values/get-latest-full-field-values id))]
+          (update-field-values! field-values-id update-map))
+        (create-field-values! field update-map))))
   {:status :success})
 
 (api/defendpoint POST "/:id/rescan_values"
--- a/src/metabase/api/field.clj
+++ b/src/metabase/api/field.clj
@@ -298,12 +298,6 @@
       [400 "If remapped values are specified, they must be specified for all field values"])
     has-human-readable-values?))
 
-(defn- update-field-values! [field-value-id update-map]
-  (api/check-500 (pos? (t2/update! FieldValues field-value-id update-map))))
-
-(defn- create-field-values! [field-or-id update-map]
-  (t2/insert! FieldValues (assoc update-map :type :full :field_id (u/the-id field-or-id))))
-
 (api/defendpoint POST "/:id/values"
   "Update the fields values and human-readable values for a `Field` whose semantic type is
   `category`/`city`/`state`/`country` or whose base type is `type/Boolean`. The human-readable values are optional."
@@ -317,11 +311,11 @@
     (let [human-readable-values? (validate-human-readable-pairs value-pairs)
           update-map             {:values                (map first value-pairs)
                                   :human_readable_values (when human-readable-values?
-                                                           (map second value-pairs))}]
-      (mdb.u/idempotent-insert!
-        (when-let [field-values-id (:id (field-values/get-latest-full-field-values id))]
-          (update-field-values! field-values-id update-map))
-        (create-field-values! field update-map))))
+                                                           (map second value-pairs))}
+          updated-entity         (mdb.u/idempotent-upsert! FieldValues
+                                   {:field_id (u/the-id field), :type :full}
+                                   (constantly update-map))]
+      (api/check-500 (pos? (:id updated-entity)))))
   {:status :success})
 
 (api/defendpoint POST "/:id/rescan_values"
--- a/src/metabase/db/util.clj
+++ b/src/metabase/db/util.clj
@@ -81,14 +81,16 @@
            (.setTransactionIsolation ^Connection conn# Connection/TRANSACTION_SERIALIZABLE)
            (t2/with-transaction [~'_conn]
              ;; We need to try select the row again now that we're in the transaction to track the dependency.
-             (or ~select-expr
-                 ;; ... and then we can execute the (potentially expensive) mutating branch.
-                 ~insert-expr)))
+             (u/or-with some?
+                      ~select-expr
+                      ;; ... and then we can execute the (potentially expensive) mutating branch.
+                      ~insert-expr)))
          (catch ExceptionInfo e#
            ;; We cannot introspect the exception cause definitively, as it will be driver specific and typically opaque,
            ;; but we should find a result now if we try selecting again after a concurrent modification exception.
-           (or ~select-expr
-               (throw (ex-info "Unable to find element after attempting an idempotent-insert!"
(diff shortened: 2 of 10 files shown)


src/metabase/api/field.clj
```diff
@@ -314,13 +309,12 @@
...
-      (t2/with-transaction [_conn]
-        (if-let [field-value-id (:id (field-values/get-latest-full-field-values id))]
-          (update-field-values! field-value-id update-map)
-          (create-field-values! field update-map)))))
+          update-map             {:values                (map first value-pairs)
+                                  :human_readable_values (when human-readable-values?
+                                                           (map second value-pairs))}
+          updated-pk             (mdb.u/update-or-insert! FieldValues {:field_id (u/the-id field), :type :full}
+                                   (constantly update-map))]
+      (api/check-500 (pos? updated-pk))))
```
crisptrutski, line 317: This check only covered `update` before, but there's no harm in spreading it to both scenarios, they return the same value.

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,134 @@
...
+   [metabase.db.util :as mdb.u]
+   [metabase.models.setting :refer [Setting]]
+   [metabase.util :as u]
+   [toucan2.core :as t2])
+  (:import
+   (java.util.concurrent CountDownLatch)))
+
+(set! *warn-on-reflection* true)
+
+(defn- repeat-concurrently [n f]
```
crisptrutski, line 14: I think this is generally useful, maybe it'll help simplify existing concurrent tests, get rid of Thread/sleeps, or encourage us to write me.

Will just leave it tucked away here for now.

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+  [& body]
+  `(try
+     ~@body
+     (catch ExceptionInfo e#
+       ;; The underlying exception thrown by the driver is database specific and opaque, so we treat any exception as a
+       ;; possible database conflict due to a concurrent insert. If we want to be more conservative, we would need
+       ;; a per-driver or driver agnostic way to test the exception.
+       ~@body)))
+
+(defn select-or-insert!
```
crisptrutski, line 84: After going through rounds and rounds of iteration, these mighty macros have reduced to simple functions 😢

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+(defmacro with-conflict-retry
+  "Retry a database mutation a single time if it fails due to concurrent insertions.
+   May retry for other reasons."
+  [& body]
+  `(try
+     ~@body
+     (catch ExceptionInfo e#
+       ;; The underlying exception thrown by the driver is database specific and opaque, so we treat any exception as a
+       ;; possible database conflict due to a concurrent insert. If we want to be more conservative, we would need
+       ;; a per-driver or driver agnostic way to test the exception.
```
crisptrutski, line 81: 🤡 😭

src/metabase/models/params/field_values.clj
```diff
@@ -100,26 +101,24 @@
...
-  The human_readable_values of Advanced FieldValues will be automatically fixed up based on the
-  list of values and human_readable_values of the full FieldValues of the same field."
+(defn prepare-advanced-field-values
+  "Fetch and construct the FieldValues for `field` with type `fv-type`. This does not do any insertion.
+   The human_readable_values of Advanced FieldValues will be automatically fixed up based on the
+   list of values and human_readable_values of the full FieldValues of the same field."
   [fv-type field hash-key constraints]
-  (when-let [{wrapped-values :values
-              :keys [has_more_values]} (fetch-advanced-field-values fv-type field constraints)]
+  (when-let [{wrapped-values :values :keys [has_more_values]}
```
crisptrutski, line 109: Aside: I found it very confusing that we'll only create advanced field values if we can fetch them. It turns out that the language here is very overloaded and this means "calculate them from the underlying fields". Perhaps it's worth renaming, but this whole side quest has gotten me exhausted enough already.

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+   To prevent this in a database agnostic way, during an existing non-serializable transaction, would be non-trivial."
+  {:style/indent 2}
+  [model select-map update-fn]
+  (with-conflict-retry
+   (let [select-kvs (mapcat identity select-map)
+         pks        (t2/primary-keys model)
+         _          (assert (= 1 (count pks)) "This helper does not currently support compound keys")
+         pk-key     (keyword (first pks))
+         validate   (fn [updated]
+                      (assert (= select-map (merge select-map (select-keys updated (keys select-map))))
```
calherries, line 128: This expression is complicated and reusable enough I feel like it deserves it's own name, and since it's duplicated it could be extracted and reduce code length. It could even be in `metabase.util`

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+   In the case where there is no underlying db constraint, concurrent calls may still result in duplicates.
+   To prevent this in a database agnostic way, during an existing non-serializable transaction, would be non-trivial."
+  {:style/indent 2}
+  [model select-map update-fn]
+  (with-conflict-retry
+   (let [select-kvs (mapcat identity select-map)
+         pks        (t2/primary-keys model)
+         _          (assert (= 1 (count pks)) "This helper does not currently support compound keys")
+         pk-key     (keyword (first pks))
+         validate   (fn [updated]
```
calherries, line 127: `validate` is a misleading name because its main operation is merging with a select map. The assertion is secondary imo, although that behaviour could be documented in the docstring.
```suggestion
         merge-with-select-map   (fn [updated]
```

test/metabase/db/util_test.clj
```diff
@@ -0,0 +1,134 @@
...
+                  (testing "we returned the same values that were inserted into the database"
+                    (is (= results (set (t2/select Setting search-col search-value)))))
+
+                  (testing "later calls just return an existing row as well"
+                    (is (contains? results (thunk)))
+                    (is (= results (set (t2/select Setting search-col search-value))))))))
+
+            ;; Since we couldn't use with-temp, we need to clean up manually.
+            (finally
+              (t2/delete! Setting search-col search-value))))))))
```
calherries, line 81: Did sanya get around to writing that `with-model-cleanup` helper? Could come in use here

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+   This is more general than using `UPSERT`, `MERGE` or `INSERT .. ON CONFLICT`, and it also allows one to avoid
+   calculating initial values that may be expensive, or require side effects.
+
+   In the case where there is an underlying db constraint to prevent duplicates, this method takes care of handling
+   rejection from the database due to a concurrent insert, and will retry a single time to pick up the existing row.
+   This may result in `insert-fn` being called a second time.
+
+   In the case where there is no underlying db constraint, concurrent calls may still result in duplicates.
+   To prevent this in a database agnostic way, during an existing non-serializable transaction, would be non-trivial."
+  {:style/indent 2}
```
calherries, line 120: Why? I don't remember seeing many functions with custom indents so this is surprising

src/metabase/db/util.clj
calherries: 😭

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+  [model select-map insert-fn]
+  (with-conflict-retry
+   (let [select-kvs (mapcat identity select-map)
+         validate   (fn [updated#]
```
calherries, line 100: this isn't shadowing anything
```suggestion
         validate   (fn [updated]
```

calherries (approved):
LGTM, I have some nits but behaviour looks good to merge

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+   In the case where there is an underlying db constraint to prevent duplicates, this method takes care of handling
+   rejection from the database due to a concurrent insert, and will retry a single time to pick up the existing row.
+   This may result in `insert-fn` being called a second time.
+
+   In the case where there is no underlying db constraint, concurrent calls may still result in duplicates.
+   To prevent this in a database agnostic way, during an existing non-serializable transaction, would be non-trivial."
+  {:style/indent 2}
+  [model select-map insert-fn]
+  (with-conflict-retry
+   (let [select-kvs (mapcat identity select-map)
```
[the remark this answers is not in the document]
crisptrutski: I have to manually set the indentation rules in Cursive, I haven't found a way for it to do this automatically from style metadata or def type yet. Cursive loses points every day. Thanks for spotting!

src/metabase/db/util.clj
```diff
@@ -66,3 +68,71 @@
...
+   calculating initial values that may be expensive, or require side effects.
+
+   In the case where there is an underlying db constraint to prevent duplicates, this method takes care of handling
+   rejection from the database due to a concurrent insert, and will retry a single time to pick up the existing row.
+   This may result in `insert-fn` being called a second time.
+
+   In the case where there is no underlying db constraint, concurrent calls may still result in duplicates.
+   To prevent this in a database agnostic way, during an existing non-serializable transaction, would be non-trivial."
+  {:style/indent 2}
+  [model select-map update-fn]
```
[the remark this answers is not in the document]
crisptrutski: Good point. The way I merge stuff is even more subtle, and the guard rails around it, is even more subtle

test/metabase/db/util_test.clj
crisptrutski: @piranha are you my knight with shining macro?

src/metabase/models/params/field_values.clj
crisptrutski: A further thought, scrolling past this again - I wonder whether we shouldn't cache the absence of values anyway. It might still be quite expensive to do the underlying query.

src/metabase/db/util.clj
crisptrutski: I struggled with this name, and eventually went with my bias for short names.

Re: primary / secondary, I wanted to highlight validation because `merge` doesn't sound like something that could throw an exception. I think you're right though that it should focus on the transformation, perhaps a comment about the capability to throw is enough.

src/metabase/db/util.clj
crisptrutski: I didn't think of a good name at the time, and that stopped me extracting it, but I agree it would really help to hide these details here. I'll try find inspiration for a util worthy name, how exciting 😁

src/metabase/db/util.clj
crisptrutski: Remnant from then this was a macro, thanks for spotting

crisptrutski:
> LGTM, I have some nits but behaviour looks good to merge

My radar is broken, because I thought this would be the more controversial change 😆 

Strongly agree with all the feedback so I'll fold it in

src/metabase/db/util.clj
crisptrutski: Another macro remnant

src/metabase/db/util.clj
calherries: `maps-conflict?`

crisptrutski: That's pretty good, clear and concise.

I can't think of many standard functions that have the type in their name, so that's one thing that puts me off.

How about making it slightly more general purpose - a function that returns the `conflicting-keys`.

src/metabase/db/util.clj
crisptrutski: Ok so I thought about this, and I think I might have found a naming loophole.

We need to remember to always call this function on the result of `update-fn`, so I'm inclined to fuse them together, and I think this may ironically make it easier to name.

```clojure
;; wrap update-fn to assert we preserve our identifiers, and make them explicit
(fn [existing]
  (let [updates (update-fn existing)]
    (assert (empty? (u/conflicting-keys select-map updates))
    (merge select-map updates))
```

I'm inclined to just shadow `update-fn`, or something obviously related like `safe-update-fn`.

src/metabase/db/util.clj
calherries: That limits its performance as `maps-conflict?` can short-circuit on the first conflict. but I'm bikeshedding, both are fine for this use case

src/metabase/db/util.clj
calherries: Nice. Yeah shadowing `update-fn` works, or naming it `update-fn'` even

test/metabase/db/util_test.clj
piranha: I could be, I just need to sort out that hooks madness!

src/metabase/db/util.clj
crisptrutski: I made it return a lazy sequence, so it still short-circuits.

test/metabase/db/util_test.clj
crisptrutski: I don't want to couple this test to that, as the fact that we're using that model is incidental, and we do not have a corresponding `defsetting` registration.

Issue #38650, opened 2024-02-12
[MLv2][Epic] Clean and remove obsolete FE MLv1 methods and utilities


nemanjaglumac:
This Epic serves the purpose of tracking assorted MLv2 cleanup efforts. So far, there is no specific structure to it, but all related cleanup tasks should be added to this epic.

```[tasklist]
### Cleanup
- [ ] https://github.com/metabase/metabase/pull/38570
- [ ] https://github.com/metabase/metabase/pull/38505
- [ ] https://github.com/metabase/metabase/pull/38551
- [ ] https://github.com/metabase/metabase/pull/38550
- [ ] https://github.com/metabase/metabase/pull/38525
- [ ] https://github.com/metabase/metabase/pull/38522
- [ ] https://github.com/metabase/metabase/pull/38508
- [ ] https://github.com/metabase/metabase/pull/38763
- [ ] https://github.com/metabase/metabase/pull/38774
- [ ] https://github.com/metabase/metabase/pull/38806
- [ ] https://github.com/metabase/metabase/pull/38820
- [ ] https://github.com/metabase/metabase/pull/38859
- [ ] Migrate MetabotQueryEditor to not use useStructuredQuery
- [ ] https://github.com/metabase/metabase/pull/38831
```

PR #38763, opened 2024-02-14
Remove MLv1 joins

ranquild:
`StructuredQuery` is used only for segments & metrics which don't have joins.


uladzimirdev (approved):

nemanjaglumac:
🎉 

Added to the main cleanup Epic list.
https://github.com/metabase/metabase/issues/38650

Issue #38446, opened 2024-02-05
Fixed width dashboards (non-blocking milestones)


JesseSDevaney:
## Context
[Product doc](https://www.notion.so/Default-dashboards-to-a-max-width-to-give-creators-a-more-predictable-canvas-0a4f5b03fcba42f59752a6c41c453ebb)
Epic: https://github.com/metabase/metabase/issues/36010
Blocking milestones of this feature: https://github.com/metabase/metabase/issues/36358

---
## Tasks

**Dashboards Needing Changes**
1. Full-App
2. Static Embedding (`PublicDashboard`)
3. X-Rays (`AutomaticDashboardApp`)

---
#### Milestone 4: Make Dashboard Header Fixed Width
- https://github.com/metabase/metabase/pull/38634
  - [x] Include the full dashboard header in the fixed width container (edit mode header is the obstacle)
    - [x] Save mode UI should still extend to the edge of the viewport

---
#### Milestone 5: Clean-up RC1 Commits
- https://github.com/metabase/metabase/pull/38678
  - [x] Clean-up new Mantine tool-tips replacing old tool-tips on dashboard header buttons
- [x] Clean-up the filters moved below tabs in public/embedded dashboards (pre-RC1 work)
  - [ ] Create an issue to track the problem of using EmbedFrame to handle multiple different embedded scenarios (dashboards, questions, actions, ...). Refactoring this to respect separation of concerns would make it easier  to understand, maintain, and develop on top of.
  - https://github.com/metabase/metabase/pull/38766
    - [x] Add sticky filters to public/embedded dashboards (related issue: #24726)
    - https://github.com/metabase/metabase/pull/38882
      - [x] Prevent parameters container from being rendered on public/embedded dashboards when no visible filters are present ([see comment](https://github.com/metabase/metabase/pull/38634#discussion_r1487418364))

---
#### Milestone 6: Setup Analytics
- https://github.com/metabase/metabase/pull/38871
- https://github.com/metabase/metabase/pull/38872
  - [x] Setup analytics to track this new settings usage
    - [Event Spec](https://www.notion.so/metabase/Dashboard-width-toggled-1b3b1e5693f44a44bf79b30fcb732a67?pvs=4)
    - [Analytics Doc](https://www.notion.so/5da1f874beda4153b4fccfa6c1e77caa?pvs=4)
  
---
#### :x: Milestone 7: Make dashboard body true fixed width w/ overflow scroll
_Left unimplemented: Value < Effort Required_
[...]

PR #38634, opened 2024-02-09
Milestone 4: Make Dashboard Header Fixed Width

JesseSDevaney:
### Problem
Making the dashboard body and filters fixed-width was a great start, because the body of the dashboard is the main experience that can change based on different screen sizes. But not including the dashboard header in the fixed width behavior, leads to a weird UI experience. Part of the dashboard controls are within a smaller area, while the other part - the header - is not.

---
#### Solution Behaviors

- Full-app
  - [ ] The dashboard header should now respect the fixed width setting just like the dashboard filters and dashboard grid 
- Public/Embed
  - [ ] The dashboard header should now respect the fixed width setting just like the dashboard filters and dashboard grid 
  - [ ] Public dashboards should look more consistent with full-app dashboards
    - [ ] Margin between dash-cards and filters should now be consistent
- X-Rays
  - [ ] The dashboard header should now respect the fixed width setting just like the dashboard filters and dashboard grid (This will always be toggled on for x-rays/automatic dashboards)

#### Validation Tests
- [ ] (`dashboard.cy.spec.js`) Validate full-app dashboard headers respect the width setting
- [ ] (`public-dashboard.cy.spec.js`) Validate public dashboard headers respect the width setting
- [ ] (`x-rays.cy.spec.js`) Validate X-ray dashboard headers respect the width setting

---
### Visual Context Changes

<details>
<summary>Full-App</summary>

**Before** (View Mode) 

image

**After** (View Mode) 

image

**Before** (Edit Mode) 

image

**After** (Edit Mode) 

image

</details>

<details>
<summary>Public/Embed</summary>

**Before**
image

**After**
image

</details>

<details>
<summary>X-Rays</summary>

**Before**
image

**After**
image

</details>

---
### How to Manually Verify the Solution
**Full-App**
1. Checkout branch and start development server
2. Open a dashboard with width setting set to "fixed"
  image
3. See that the dashboard header now respects the width setting (in view mode and edit mode)
4. Toggle the dashboard width setting to "full"
5. See that the behavior works as before (in view mode and edit mode)

**Public/Embedding**
1. Enable public sharing
  image
2. Open up a dashboard with width set to "fixed" and create a public link
  image
  image
  image
4. Paste the public dashboard link in a new tab
5. See that public/embed dashboard header is fixed width
6. In the full-app, change the width to "full"
7. Refresh the public dashboard view and see that the dashboard header respects width set to "full"

**X-Rays**
[...]

PR #38766, opened 2024-02-14
Make the filter bar sticky for public dashboards

### LINKED ISSUE #24726: Sticky filters on Embedded/Public dashboards
When creating larger dashboards with filters, then the user has to scroll up to set filters on Public or Embedded, because filters are not sticky.

For consistency sake, have sticky filters, since normal in-app (and FullApp embedding) has used sticky since v41 (#11187).
Or at least provide sticky filters as an options.

Create smaller dashboards, but that's not always an option.

**Additional context**
It's important to be aware of mobile to prevent problems like #20879 and #23983

kulyk:
Epic #38446, closes #24726

### To verify

1. Sign in as admin
2. Enable public sharing at `/admin/settings/public-sharing`
3. Open a dashboard that has several filters
4. Click the sharing icon at the top right, generate a public link and open it
5. Ensure the filter bar stays at the top of the viewport when scrolling

### Demo

https://github.com/metabase/metabase/assets/17258145/f1bb7428-e86f-4c35-a845-b23960ad00ee

### DIFF
--- a/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
+++ b/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
@@ -103,8 +103,10 @@ const footerVariantStyles = {
 
 export const ParametersWidgetContainer = styled(FullWidthContainer)<{
   isSticky: boolean;
-  hasVisibleParameters: boolean;
 }>`
+  padding-top: ${space(1)};
+  padding-bottom: ${space(1)};
+
   ${props =>
     props.isSticky &&
     css`
@@ -115,13 +117,6 @@ export const ParametersWidgetContainer = styled(FullWidthContainer)<{
       z-index: 3;
 
       background-color: ${color("white")};
-    `}
-
-  ${props =>
-    props.hasVisibleParameters &&
-    css`
-      padding-top: ${space(1)};
-      padding-bottom: ${space(1)};
     `}
 `;
 
--- a/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
+++ b/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
@@ -102,15 +102,20 @@ const footerVariantStyles = {
 };
 
 export const ParametersWidgetContainer = styled(FullWidthContainer)<{
+  isSticky: boolean;
   hasVisibleParameters: boolean;
 }>`
-  position: sticky;
-  top: 0;
-  left: 0;
-  width: 100%;
-  z-index: 3;
+  ${props =>
+    props.isSticky &&
+    css`
+      position: sticky;
+      top: 0;
+      left: 0;
+      width: 100%;
+      z-index: 3;
 
-  background-color: ${color("white")};
+      background-color: ${color("white")};
+    `}
 
   ${props =>
     props.hasVisibleParameters &&
--- a/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
+++ b/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
@@ -102,10 +102,17 @@ const footerVariantStyles = {
 };
 
 export const ParametersWidgetContainer = styled(FullWidthContainer)<{
+  hasScroll: boolean;
   isSticky: boolean;
 }>`
   padding-top: ${space(1)};
   padding-bottom: ${space(1)};
+
+  ${props =>
+    props.hasScroll &&
+    css`
+      border-bottom: 1px solid ${color("border")};
+    `}
 
   ${props =>
     props.isSticky &&
--- a/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
+++ b/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
@@ -36,13 +36,11 @@ export const Root = styled.div<{
     `}
 `;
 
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
```diff
@@ -103,9 +101,23 @@ const footerVariantStyles = {
...
+  z-index: 3;
+
+  background-color: ${color("white")};
+
+  ${props =>
+    props.hasVisibleParameters &&
+    css`
+      padding-top: ${space(1)};
+      padding-bottom: ${space(1)};
+    `}
```
kulyk, line 120: @WiNloSt this should address [your comment](https://github.com/metabase/metabase/pull/38634#discussion_r1487418364) on of the previous PRs

JesseSDevaney (changes requested):
It is feeling good to me. Thanks for picking this up and working on it Anton! :heart: 

I think there are a couple things we should address before merging though.

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
```diff
@@ -183,34 +187,37 @@ function EmbedFrame({
...
-                  />
-                  {dashboard && <FilterApplyButton />}
-                </ParametersFixedWidthContainer>
-              </ParametersWidgetContainer>
-            )}
           </Header>
         )}
+        {hasParameters && (
+          <ParametersWidgetContainer
+            hasVisibleParameters={hasVisibleParameters}
```
WiNloSt, line 194: since we already know we don't need to render the parameter should we not render the `ParametersWidgetContainer` instead of removing the padding via CSS?

We also do not render the parameter on the dashboard instead of using CSS magic.

https://github.com/metabase/metabase/blob/8ffcc27f275948a852e87d667144b29b1b2fe597/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx#L527-L543

kulyk:
@JesseSDevaney @WiNloSt thanks for thorough reviews! 🙏
Please take another look — all your comment should now be addressed

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
```diff
@@ -103,9 +101,30 @@ const footerVariantStyles = {
...
+  ${props =>
+    props.isSticky &&
+    css`
+      position: sticky;
+      top: 0;
+      left: 0;
+      width: 100%;
+      z-index: 3;
+
+      background-color: ${color("white")};
```
JesseSDevaney, line 126: Apologies for not catching this until now, but I just realized the embedded version has configurable color themes. We need to account for this in the parameters container, so that we change the color based on the theme

**Light** :heavy_check_mark: 
image

**Dark** :x: 
image

**Transparent** :x: 
image

I think we need to do something special for transparent though, since the sticky filters container will look bad if it is transparent. 

**Transparent Filters** :x:
image

JesseSDevaney (changes requested):
Thanks for addressing the feedback! It feels much better now.

Unfortunately I missed something in my first review that I only just now caught. :cry: 
Apologies for bringing it up late. Once we address this last piece, it is good to go for me.

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
WiNloSt: Good catch!

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
```diff
@@ -227,5 +258,15 @@ function EmbedFrame({
   );
 }
 
+function isParametersWidgetContainersSticky(parameterCount: number) {
```
WiNloSt, line 261: This looks the same as the main dashboard one so it's good.

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
```diff
@@ -183,34 +210,38 @@ function EmbedFrame({
...
-                      enableParameterRequiredBehavior
-                    }
-                  />
-                  {dashboard && <FilterApplyButton />}
-                </ParametersFixedWidthContainer>
-              </ParametersWidgetContainer>
-            )}
           </Header>
         )}
+        {hasVisibleParameters && (
```
WiNloSt, line 215: Now, the code is consistent with the main dashboard 👍

WiNloSt (commented):
Looks good on the parameter part. I'll let Jesse stamp this PR though.

frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.styled.tsx
kulyk: Good catch indeed! 🙏

kulyk: @JesseSDevaney please take another look. [`25fcde6946`](https://github.com/metabase/metabase/pull/38766/commits/25fcde6946bdf4898ab9e8f9282aab634bd4bbbd) should fix filter panel colors and [`46fcd66f0f`](https://github.com/metabase/metabase/pull/38766/commits/46fcd66f0f697cfb3d43cbcdccc4e8a00caee9d0) will disable the sticky behavior if a dashboard is embedded with a "transparent" them

JesseSDevaney (approved):
:rocket: 

Thanks for addressing those issues, Anton! :heart:

PR #38767, opened 2024-02-14
[WIP] Remove MLv1 nested queries

ranquild:
`StructuredQuery` is used only for segments & metrics which don't have nested queries.


ranquild:
Found out https://github.com/metabase/metabase/blob/80b5958943ed22329cbd9a1778fc03b1bb7306dc/frontend/src/metabase-lib/parameters/utils/targets.ts#L28 

dashboard still use MLv1

PR #38774, opened 2024-02-14
Remove some legacy query usages from QB

ranquild:


frontend/src/metabase/query_builder/components/notebook/NotebookSteps/NotebookSteps.tsx
```diff
@@ -22,7 +22,8 @@ interface NotebookStepsProps {
 }
 
 function getInitialOpenSteps(question: Question, readOnly: boolean): OpenSteps {
-  const isNew = !readOnly && !question.table();
+  const query = question.query();
+  const isNew = !readOnly && !Lib.sourceTableOrCardId(query);
```
kamilmielnik, line 26:
```suggestion
  const isNew = !readOnly && Lib.sourceTableOrCardId(query) === null;
```

_Just like in View.jsx in lines 204 and 324_

kamilmielnik (approved):
:+1:

PR #38779, opened 2024-02-14
Add server side generated relative datetime capabilities to Snowflake

### LINKED ISSUE #20988: Make effective use of Snowflake's query result caching
Dashboards take longer to load than necessary because query results from previous runs [can not be cached due to usage of `current_timestamp()` function](https://docs.snowflake.com/en/user-guide/querying-persisted-results.html#retrieval-optimization) for comparison even when the column compared to is a date (and `current_date()` could be used which allows for results caching).

When comparing against columns of `date` type use `current_date()` function. At least when using Snowflake dialect.

Writing queries by hand, but that limits the number of people that can self serve effectively.

It's what raises most questions from users, who expected dashboards will be snappy.

**Additional context**
Asking for a count of page views by day in last 30 days produces query like this
```sql
SELECT 
    date_trunc("day", CAST("PRODUCT_ANALYTICS"."PAGES_UNION"."DATE" AS timestamp)) AS "DATE", 
    count(*) AS "count"
FROM "DEV_SANDBOX"."PRODUCT_ANALYTICS"."PAGES_UNION"
WHERE ("PRODUCT_ANALYTICS"."PAGES_UNION"."DATE" >= date_trunc("day", CAST(dateadd(day, -30, CAST(current_timestamp() AS timestamp)) AS timestamp))
   AND "PRODUCT_ANALYTICS"."PAGES_UNION"."DATE" < date_trunc("day", CAST(current_timestamp() AS timestamp)))
GROUP BY date_trunc("day", CAST("PRODUCT_ANALYTICS"."PAGES_UNION"."DATE" AS timestamp))
ORDER BY date_trunc("day", CAST("PRODUCT_ANALYTICS"."PAGES_UNION"."DATE" AS timestamp)) DESC;
```
If I replace `current_timestamp()` with `current_date()` I get exact same results, except query returns in a matter of milliseconds next time because results are cached.

On a somewhat related note. I understand that compromises need to be taken when writing a generic query generation code, but there seems to be a lot of casting dates to timestamps only to truncate them back to dates. How come? 🤔 

I checked and metabase is aware that the field type is `date`
image

lbrdnk:
Closes #20988

### Description

This PR extends the approach used in #38604 to Snowflake.

### How to verify

Run modified tests also for Snowflake driver.

### Checklist

### DIFF
--- a/modules/drivers/snowflake/src/metabase/driver/snowflake.clj
+++ b/modules/drivers/snowflake/src/metabase/driver/snowflake.clj
@@ -32,6 +32,7 @@
    [metabase.query-processor.timezone :as qp.timezone]
    [metabase.query-processor.util :as qp.util]
    [metabase.query-processor.util.add-alias-info :as add]
+   [metabase.query-processor.util.relative-datetime :as qp.relative-datetime]
    [metabase.util :as u]
    [metabase.util.date-2 :as u.date]
    [metabase.util.honey-sql-2 :as h2x]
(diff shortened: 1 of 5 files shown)


modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -199,10 +200,6 @@
 (defmethod sql.qp/unix-timestamp->honeysql [:snowflake :milliseconds] [_ _ expr] [:to_timestamp expr 3])
 (defmethod sql.qp/unix-timestamp->honeysql [:snowflake :microseconds] [_ _ expr] [:to_timestamp expr 6])
 
-(defmethod sql.qp/current-datetime-honeysql-form :snowflake
```
lbrdnk, line 202: This was defined twice in the file. Another definition can be found on [this line](https://github.com/metabase/metabase/blob/3f1c95ad3f7c733f93629c1730a1707845478d52/modules/drivers/snowflake/src/metabase/driver/snowflake.clj#L547)

metamben (approved):
👍

PR #38787, opened 2024-02-14
Add Creator Sentiment Email task

### LINKED ISSUE #38739: measuring creator sentiment (BE)
https://www.notion.so/metabase/Measuring-Creator-Sentiment-f9b5e01e175540cf969c24b3159a29e8?d=d20237da699b481f8c63f16fe95996b5#aa173b3cce3a41cf9173492d2017c0b2
```[tasklist]
### BE Tasks
- [ ] monthly task to send emails
- [ ] query to find users (modding by month)
- [ ] setting to disable sending emails
- [ ] send the email to creators
```

qwef:
Closes https://github.com/metabase/metabase/issues/38739

### Demo

image

### DIFF
--- a/src/metabase/analytics/stats.clj
+++ b/src/metabase/analytics/stats.clj
@@ -16,23 +16,12 @@
    [metabase.integrations.google :as google]
    [metabase.integrations.slack :as slack]
    [metabase.models
-    :refer [Card
-            Collection
-            Dashboard
-            DashboardCard
-            Database
-            Field
-            Metric
-            PermissionsGroup
-            Pulse
-            PulseCard
-            PulseChannel
-            QueryCache
-            Segment
-            Table
-            User]]
+    :refer [Card Collection Dashboard DashboardCard Database Field Metric
+            PermissionsGroup Pulse PulseCard PulseChannel QueryCache Segment
+            Table User]]
    [metabase.models.humanization :as humanization]
    [metabase.public-settings :as public-settings]
+   [metabase.public-settings.premium-features :as premium-features]
    [metabase.util :as u]
    [metabase.util.honey-sql-2 :as h2x]
    [metabase.util.i18n :refer [trs]]
@@ -135,6 +124,16 @@
   "Returns true if the 'Chart Colors' have been customized"
   []
   (boolean (seq (apply dissoc (public-settings/application-colors) ui-colors))))
+
+(defn fetch-plan-info
+  "Figure out what plan this Metabase instance is on."
+  []
+  (cond
+    (and config/ee-available? (premium-features/is-hosted?) (premium-features/has-any-features?)) "pro-cloud/enterprise-cloud"
+    (and config/ee-available? (premium-features/is-hosted?) (not (premium-features/has-any-features?))) "starter"
+    (and config/ee-available? (not (premium-features/is-hosted?))) "pro-self-hosted/enterprise-self-hosted"
+    (not config/ee-available?) "oss"
+    :else "unknown"))
 
 (defn- instance-settings
   "Figure out global info about this instance"
--- a/src/metabase/analytics/stats.clj
+++ b/src/metabase/analytics/stats.clj
@@ -21,7 +21,6 @@
             Table User]]
    [metabase.models.humanization :as humanization]
    [metabase.public-settings :as public-settings]
-   [metabase.public-settings.premium-features :as premium-features]
    [metabase.util :as u]
    [metabase.util.honey-sql-2 :as h2x]
    [metabase.util.i18n :refer [trs]]
@@ -124,16 +123,6 @@
   "Returns true if the 'Chart Colors' have been customized"
   []
   (boolean (seq (apply dissoc (public-settings/application-colors) ui-colors))))
-
-(defn fetch-plan-info
-  "Figure out what plan this Metabase instance is on."
-  []
-  (cond
-    (and config/ee-available? (premium-features/is-hosted?) (premium-features/has-any-features?)) "pro-cloud/enterprise-cloud"
-    (and config/ee-available? (premium-features/is-hosted?) (not (premium-features/has-any-features?))) "starter"
-    (and config/ee-available? (not (premium-features/is-hosted?))) "pro-self-hosted/enterprise-self-hosted"
-    (not config/ee-available?) "oss"
-    :else "unknown"))
 
 (defn- instance-settings
   "Figure out global info about this instance"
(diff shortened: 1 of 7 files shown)


src/metabase/email/creator_sentiment_email.mustache
```diff
@@ -0,0 +1,36 @@
...
+    <p style="font-weight: 500; line-height: 1.375em; color: {{colorTextMedium}};">
+      Hi {{first-name}},
+    </p>
+
+    <p style="font-weight: 500; line-height: 1.375em; color: {{colorTextMedium}};">
+      We at Metabase are working hard to make the experience of people like you, who create lots of content (questions, dashboards, etc).
+    </p>
+
+    <p style="font-weight: 600; line-height: 1.375em; color: {{colorTextMedium}};">
+      We’d like to know how we’re doing. Can you spare 30 seconds to fill out this NPS survey?
```
brunobergher, line 16: Can you please add a line break?
> `We’d like to know how we’re doing.<br>Can you spare 30 seconds to fill out this NPS survey?`

src/metabase/email/creator_sentiment_email.mustache
```diff
@@ -0,0 +1,36 @@
+{{> metabase/email/_header }}
+  <div class="card-container">
+    <p style="font-weight: 500; font-size: 0.875em; line-height: 1.375em; color: {{colorTextMedium}};">
```
brunobergher, line 3: This paragraph shouldn't be shown if the instance has the `hosting` flag _enabled_.
Can't say if you included that in the clojure code, but in any case the `<p>` shouldn't be there, even if empty, in those cases.

src/metabase/task/creator_sentiment_emails.clj
```diff
@@ -0,0 +1,127 @@
...
+                  :from [[:report_dashboard :d]]
+                  :where [:= :d.archived false]})
+   :num_dashboards))
+
+(defn- fetch-num-cards []
+  (t2/query-one {:select [[[:count [:distinct [:case [:and [:= :rc.type "question"] [:= :rc.archived false]] :rc.id]]] :num_questions]
+                          [[:count [:distinct [:case [:and [:= :rc.type "model"] [:= :rc.archived false]] :rc.id]]] :num_models]]
+                 :from [[:report_card :rc]]}))
+
+(defn- fetch-plan-info []
```
brunobergher, line 74: This seems like something which would be useful elsewhere in the app, we run into stuff like this all the time.

brunobergher (commented):
Small things to tweak and a reuse suggestion,

src/metabase/email/creator_sentiment_email.mustache
```diff
@@ -0,0 +1,36 @@
...
+      <i>{{self-hosted}}</i>
+    </p>
+
+    <p style="font-weight: 500; line-height: 1.375em; color: {{colorTextMedium}};">
+      Hi {{first-name}},
+    </p>
+
+    <p style="font-weight: 500; line-height: 1.375em; color: {{colorTextMedium}};">
+      We at Metabase are working hard to make the experience of people like you, who create lots of content (questions, dashboards, etc).
```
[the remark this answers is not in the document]
qwef: Good catch

src/metabase/analytics/stats.clj
```diff
@@ -136,6 +125,16 @@
   []
   (boolean (seq (apply dissoc (public-settings/application-colors) ui-colors))))
 
+(defn fetch-plan-info
```
noahmoss, line 128: This namespace is specifically for our anonymous stats collection, so if this isn't also being used for that, I think it could be moved closer to the email code

src/metabase/email/creator_sentiment_email.mustache
```diff
@@ -0,0 +1,34 @@
+{{> metabase/email/_header }}
+  <div class="card-container" style="color: {{colorTextMedium}};">
+    {{{self-hosted}}}
```
noahmoss, line 3: I think it'd be a little cleaner to have the HTML in here for the self-hosted warning, and just show it conditionally based on a `self-hosted` boolean passed in via the context (rather than building the HTML in Clojure). You can do conditionals like `{{#self-hosted}} content {{/self-hosted}}` — see `pulse.mustache` for some examples

src/metabase/email/messages.clj
```diff
@@ -291,6 +291,36 @@
...
+                 json/generate-string
+                 .getBytes
+                 codecs/bytes->b64-str)
+        context (merge (common-context)
+                       {:emailType  "notification"
+                        :logoHeader true
+                        :first-name first_name}
+                       (if (public-settings/anon-tracking-enabled)
+                         {:link (str "https://metabase.com/feedback/creator?context=" blob)}
+                         {:link "https://metabase.com/feedback/creator"})
```
noahmoss, line 312: nit
```suggestion
                        :first-name first_name
                        :link (if (public-settings/anon-tracking-enabled)
                                (str "https://metabase.com/feedback/creator?context=" blob)
                                "https://metabase.com/feedback/creator")}
```

src/metabase/email/messages.clj
```diff
@@ -291,6 +291,36 @@
                :message      (stencil/render-file "metabase/email/follow_up_email" context)}]
     (email/send-message! email)))
 
+(defn send-creator-sentiment-email!
+  "Format and send an email to the system admin following up on the installation."
+  [{:keys [email created_at first_name num_dashboards num_questions num_models]} instance-data]
+  {:pre [(u/email? email)]}
+  (let [blob (-> {:instance instance-data
```
noahmoss, line 298: nit: `blob` only needs to be computed when anon tracking is enabled... maybe make it a function

src/metabase/task/creator_sentiment_emails.clj
```diff
@@ -0,0 +1,118 @@
...
+   (t2/query-one {:select [[:%count.* :num_users]]
+                  :from [[:core_user :u]]
+                  :where [:and
+                          [:= :u.is_active true]
+                          [:= :u.type "personal"]]})
+   :num_users))
+
+(defn- fetch-num-dashboards []
+  (->
+   (t2/query-one {:select [[:%count.* :num_dashboards]]
```
noahmoss, line 65: can you use `t2/count` for these simpler count queries?

src/metabase/task/creator_sentiment_emails.clj
```diff
@@ -0,0 +1,118 @@
...
+
+(setting/defsetting ^:private no-surveys
+  (deferred-tru "Enable or disable creator sentiment emails")
+  :type       :boolean
+  :default    false
+  :visibility :internal
+  :audit      :getter)
+
+(defn- fetch-creators
+  [has-whitelabelling?]
```
noahmoss, line 32: "Creator" isn't a formal concept on the BE so it would be useful to have a docstring/comment here briefly explaining what defines a creator

noahmoss (approved):
looks good, just nits & small comments

PR #38806, opened 2024-02-15
Remove MLv1 virtual tables

ranquild:

Remove `Table` generation from MLv1. We should use metadata from QP in all cases.


frontend/src/metabase/query_builder/components/dataref/QuestionPane/QuestionPane.tsx
```diff
@@ -106,6 +109,12 @@ export default _.compose(
   Questions.load({
     id: (_state: State, props: QuestionPaneProps) => props.question.id,
   }),
+  Tables.load({
```
ranquild, line 112: Load missing data to avoid creating the virtual table by MLv1

frontend/src/metabase-lib/Question.ts
```diff
@@ -720,13 +720,19 @@ class Question {
...
-    return query && typeof query.table === "function" ? query.table() : null;
+    const query = this.query();
+    const { isNative } = Lib.queryDisplayInfo(query);
+    if (isNative) {
+      return this.legacyQuery().table();
+    } else {
+      const tableId = Lib.sourceTableOrCardId(query);
+      const metadata = this.metadata();
+      return metadata.table(tableId);
+    }
```
kamilmielnik, line 735: :pray: 

```suggestion
    const { isNative } = Lib.queryDisplayInfo(query);
    
    if (isNative) {
      return this.legacyQuery().table();
    }

    const tableId = Lib.sourceTableOrCardId(query);
    const metadata = this.metadata();
    return metadata.table(tableId);
```

kamilmielnik (approved):
:100:

uladzimirdev (approved):

PR #38444, opened 2024-02-05
Clean up dashboard card types

kulyk:
A few changes around the `DashboardCard` type which should make it easier to write code handling different kinds of dashboard cards.

There're two main changes:

1. `DashboardCard` is now split into `QuestionDashboardCard` and `VirtualDashboardCard` type. The old code was basically combining the two, making some type checks awkward.
2. `DashboardCard` is now `ActionDashboardCard | QuestionDashboardCard | VirtualDashboardCard`, which should force us to narrow down the dashcards types while working with them (there are type guards)

The rest is fixing existing code to pass the type check

**To verify:** CI should be green

PR #38809, opened 2024-02-15
Translate dashboard/actions/parameters to typescript

oleggromov:
The PR does what is says 😅

### DIFF
--- a/frontend/src/metabase/actions/components/ActionViz/ActionParameterMapping.tsx
+++ b/frontend/src/metabase/actions/components/ActionViz/ActionParameterMapping.tsx
@@ -68,7 +68,7 @@ export const ActionParameterMappingForm = ({
         setParameterMapping(
           dashboardParameterId,
           dashcard.id,
-          undefined, // this is irrelevant for action parameters
+          -1, // this is irrelevant for action parameters
           target,
         ),
       );
--- a/frontend/src/metabase/dashboard/actions/parameters.ts
+++ b/frontend/src/metabase/dashboard/actions/parameters.ts
@@ -16,6 +16,21 @@ import { updateDashboard } from "metabase/dashboard/actions/save";
 import { autoWireDashcardsWithMatchingParameters } from "metabase/dashboard/actions/auto-wire-parameters/actions";
 import { getParameterMappings } from "metabase/dashboard/actions/auto-wire-parameters/utils";
 import { closeAutoWireParameterToast } from "metabase/dashboard/actions/auto-wire-parameters/toasts";
+import type { Dispatch, GetState } from "metabase-types/store";
+import type {
+  ActionDashboardCard,
+  CardId,
+  DashCardId,
+  Parameter,
+  ParameterId,
+  ParameterMappingOptions,
+  ParameterTarget,
+  QuestionDashboardCard,
+  ValuesQueryType,
+  ValuesSourceConfig,
+  ValuesSourceType,
+  WritebackAction,
+} from "metabase-types/api";
 import {
   isParameterValueEmpty,
   PULSE_PARAM_EMPTY,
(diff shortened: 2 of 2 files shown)


frontend/src/metabase/dashboard/actions/parameters.ts
```diff
@@ -115,30 +148,35 @@ export const SET_PARAMETER_MAPPING = "metabase/dashboard/SET_PARAMETER_MAPPING";
...
-      const dashcard = getDashCardById(getState(), dashcard_id);
+      const dashcard = getDashCardById(getState(), dashcardId);
 
       if (target !== null) {
         dispatch(
           autoWireDashcardsWithMatchingParameters(
-            parameter_id,
-            dashcard,
+            parameterId,
+            dashcard as QuestionDashboardCard,
```
oleggromov, line 166: @kulyk please have a look, maybe you have a better idea of how to deal with the type mismatch?

frontend/src/metabase/actions/components/ActionViz/ActionParameterMapping.tsx
```diff
@@ -68,7 +68,7 @@ export const ActionParameterMappingForm = ({
         setParameterMapping(
           dashboardParameterId,
           dashcard.id,
-          undefined, // this is irrelevant for action parameters
+          -1, // this is irrelevant for action parameters
```
oleggromov, line 71: I am not sure if that's entirely safe for the FE, but for the BE it [should be](https://metaboat.slack.com/archives/C05MPF0TM3L/p1708005245412669?thread_ts=1708004479.233609&cid=C05MPF0TM3L).

frontend/src/metabase/dashboard/actions/parameters.ts
kulyk: You might want to narrow down `dashcard's` type before doing the mapping. You can find a type-guard per each subtype of a daschard in `metabase/dashboard/utils` (`isQuestionDashCard`, `isActionDashCard` and `isVirtualDashCard`)

frontend/src/metabase/dashboard/actions/parameters.ts
```diff
@@ -64,36 +93,40 @@ function updateParameters(dispatch, getState, parametersUpdater) {
...
+    }
+  };
 
 export const ADD_PARAMETER = "metabase/dashboard/ADD_PARAMETER";
 export const addParameter = createThunkAction(
   ADD_PARAMETER,
-  parameterOption => (dispatch, getState) => {
-    let parameter;
+  (option: ParameterMappingOptions) => (dispatch, getState) => {
+    let newId: ParameterId;
```
ranquild, line 116: Maybe set to `undefined` here and in `dispatch` call `checkNotNull` or similar to make an explicit check?

ranquild (approved):

frontend/src/metabase/actions/components/ActionViz/ActionParameterMapping.tsx
oleggromov: Should be safe, as we only genereate [negative IDs for `dashcards`](https://github.com/metabase/metabase/blob/792a4ed7c627a2018de14f149f618dca53866833/frontend/src/metabase/dashboard/utils.ts#L329) and for new queries we don't add IDs as far as I can tell.

frontend/src/metabase/dashboard/actions/parameters.ts
oleggromov: This is actually a good idea, I have initialized it with `undefined`, which helped the compiler understand the intention without a warning.

I am using an explicit check instead of `checkNotNull` because the latter throws exceptions.

frontend/src/metabase/dashboard/actions/parameters.ts
```diff
@@ -115,16 +150,21 @@ export const SET_PARAMETER_MAPPING = "metabase/dashboard/SET_PARAMETER_MAPPING";
...
+    target: ParameterTarget | null,
+  ) => {
     return (dispatch, getState) => {
       dispatch(closeAutoWireParameterToast());
 
-      const dashcard = getDashCardById(getState(), dashcard_id);
+      const dashcard = getDashCardById(getState(), dashcardId);
 
-      if (target !== null) {
+      if (target !== null && isQuestionDashCard(dashcard)) {
```
oleggromov, line 164: A type guard makes sense here since auto-wiring is only working for questions.

frontend/src/metabase/dashboard/actions/parameters.ts
```diff
@@ -133,12 +173,13 @@ export const setParameterMapping = createThunkAction(
...
         setDashCardAttributes({
-          id: dashcard_id,
+          id: dashcardId,
           attributes: {
             parameter_mappings: getParameterMappings(
-              dashcard,
-              parameter_id,
-              card_id,
+              // TODO remove type casting when getParameterMappings is fixed
+              dashcard as QuestionDashboardCard,
```
oleggromov, line 180: However, not here, because I believe that [`getParameterMappings`](https://github.com/metabase/metabase/blob/hk-dashboard-actions-parameters-ts/frontend/src/metabase/dashboard/actions/auto-wire-parameters/utils.ts#L131-L165) currently _incorrectly_ expects only a `QuestionDashboardCard`.

@kulyk could you confirm that please? I am asking because you [changed this recently](https://github.com/metabase/metabase/pull/38444/files#diff-9d4407a615d5b08c4b7b4e07c6f83468444aa21d936c440b7e40b9d49559eab6R129).

TBH, I am not sure how to type it properly so I am leaving it outside of this PR.

PR #38810, opened 2024-02-15
More pervasive and atomic Liquibase locking 

### LINKED ISSUE #37553: Make executing `consolidate-liquibase-changesets!` safer
We have this function [consolidate-liquibase-changesets!](https://github.com/metabase/metabase/blob/e66c0754cace99451ba1bdc85ef50335ca8dc584/src/metabase/db/liquibase.clj#L307) that is executed during startup. What it does is updating `databasechangelog.filename` to the correct filename. It's needed because we sometimes move the destination of the migration file. The first time was when we moved from one migration per file to all migrations into 1 single file(`000_migrations.yaml`), and the 2nd time was when we split the old `000_migrations.yaml` into `000_legacy_migration.yaml` and `001_update_migrations.yaml` (https://github.com/metabase/metabase/pull/34400).

The problem is that executing this query in a multi-instance env can result in unnecessary updates. Ideally, there should be only one instance running these updates, and the other will have to wait.

I propose that we reuse liquidates locking mechanism to execute this query, and also, the query should not run the update blindly, use a select query to decide whether or not the update is needed.

crisptrutski:
Closes https://github.com/metabase/metabase/issues/37553

### Description

This change makes sure that we take locks before beginning mutation with Liquibase, in particular the consolidation of our changesets, and should hopefully fix some race conditions we have been seeing. 

~~The lock is re-entrant, so I have done this in a minimally invasive way. There is scope to coarsen the locking, and remove some redundancy. I'm on the fence about coarsening too much, as it's very easy to call into the internals.~~ I've coarsened the lock somewhat, and rely on a runtime assertion within the internals. There is scope to make it even more coarse, but I think this is good for now.

~~The linked ticket also calls for making the consolidation a no-op if the table is already up-to-date. I will do that in a separate PR.~~ I've ended up including this, because it tied in with taking the lock conditionally.

### How to verify

I haven't added any tests for these locks, as they would need to be very intrusive and mock out internals. 

Perhaps there's some prior art on a good way to do this? I wanted to first put this out for feedback before spending any time on tests, and to gauge how extensive we want to test it.

### DIFF
--- a/src/metabase/db/liquibase.clj
+++ b/src/metabase/db/liquibase.clj
@@ -219,34 +219,11 @@
       (log/warnf "Migration lock was cleared after %d retries." @retry-counter)
       (log/info "No migration lock found."))))
 
-(defn migrate-up-if-needed!
-  "Run any unrun `liquibase` migrations, if needed."
-  [^Liquibase liquibase ^DataSource data-source]
-  (log/info (trs "Checking if Database has unrun migrations..."))
-  (if (seq (unrun-migrations data-source))
-    (do
-     (log/info (trs "Database has unrun migrations. Checking if migraton lock is taken..."))
-     (wait-for-migration-lock-to-be-cleared liquibase)
-     ;; while we were waiting for the lock, it was possible that another instance finished the migration(s), so make
-     ;; sure something still needs to be done...
-     (let [to-run-migrations      (unrun-migrations data-source)
-           unrun-migrations-count (count to-run-migrations)]
-       (if (pos? unrun-migrations-count)
-         (let [^Contexts contexts nil
-               start-time         (System/currentTimeMillis)]
-           (log/info (trs "Running {0} migrations ..." unrun-migrations-count))
-           (doseq [^ChangeSet change to-run-migrations]
-             (log/tracef "To run migration %s" (.getId change)))
-           (.update liquibase contexts)
-           (log/info (trs "Migration complete in {0}" (u/format-milliseconds (- (System/currentTimeMillis) start-time)))))
-         (log/info
-          (trs "Migration lock cleared, but nothing to do here! Migrations were finished by another instance.")))))
-    (log/info (trs "No unrun migrations found."))))
-
 (defn run-in-scope-locked
   "Run function `f` in a scope on the Liquibase instance `liquibase`.
-  Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
-  Clojure). This function initializes the database and the resource accessor which are often required."
+   Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
+   Clojure). This function initializes the database and the resource accessor which are often required.
+   The underlying locks are re-entrant, so it is safe to nest these blocks."
   [^Liquibase liquibase f]
   (let [database (.getDatabase liquibase)
         ^LockService lock-service (.getLockService (LockServiceFactory/getInstance) database)
@@ -261,6 +238,40 @@
                        (finally
                          (.releaseLock lock-service))))))))
 
+(defmacro with-scope-locked
+  "Run `body` in a scope on the Liquibase instance `liquibase`.
+   Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
+   Clojure). This function initializes the database and the resource accessor which are often required.
+   The underlying locks are re-entrant, so it is safe to nest these blocks."
+  {:style/indent 1}
+  [liquibase & body]
+  `(run-in-scope-locked ~liquibase (fn [] ~@body)))
+
+(defn migrate-up-if-needed!
+  "Run any unrun `liquibase` migrations, if needed."
+  [^Liquibase liquibase ^DataSource data-source]
+  (log/info (trs "Checking if Database has unrun migrations..."))
+  (if (seq (unrun-migrations data-source))
+    (do
+     (log/info (trs "Database has unrun migrations. Checking if migration lock is taken..."))
+     (wait-for-migration-lock-to-be-cleared liquibase)
+     (with-scope-locked liquibase
+      ;; while we were waiting for the lock, it was possible that another instance finished the migration(s), so make
+      ;; sure something still needs to be done...
+      (let [to-run-migrations      (unrun-migrations data-source)
+            unrun-migrations-count (count to-run-migrations)]
+        (if (pos? unrun-migrations-count)
+          (let [^Contexts contexts nil
+                start-time         (System/currentTimeMillis)]
+            (log/info (trs "Running {0} migrations ..." unrun-migrations-count))
+            (doseq [^ChangeSet change to-run-migrations]
+              (log/tracef "To run migration %s" (.getId change)))
+            (.update liquibase contexts)
+            (log/info (trs "Migration complete in {0}" (u/format-milliseconds (- (System/currentTimeMillis) start-time)))))
+          (log/info
+           (trs "Migration lock cleared, but nothing to do here! Migrations were finished by another instance."))))))
+    (log/info (trs "No unrun migrations found."))))
+
 (defn update-with-change-log
   "Run update with the change log instances in `liquibase`."
   ([liquibase]
@@ -273,9 +284,8 @@
          log-iterator   (ChangeLogIterator. change-log ^"[Lliquibase.changelog.filter.ChangeSetFilter;" (into-array ChangeSetFilter change-set-filters))
          update-visitor (UpdateVisitor. database ^ChangeExecListener exec-listener)
          runtime-env    (RuntimeEnvironment. database (Contexts.) nil)]
-     (run-in-scope-locked
-      liquibase
-      #(.run ^ChangeLogIterator log-iterator update-visitor runtime-env)))))
+     (with-scope-locked liquibase
+       (.run ^ChangeLogIterator log-iterator update-visitor runtime-env)))))
 
 (mu/defn force-migrate-up-if-needed!
   "Force migrating up. This does three things differently from [[migrate-up-if-needed!]]:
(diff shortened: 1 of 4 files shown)


crisptrutski:
I've broken the force migration test, so clearly I've missed something important - hopefully not fundamental.

src/metabase/db/liquibase.clj
```diff
@@ -261,6 +238,40 @@
...
+  `(run-in-scope-locked ~liquibase (fn [] ~@body)))
+
+(defn migrate-up-if-needed!
+  "Run any unrun `liquibase` migrations, if needed."
+  [^Liquibase liquibase ^DataSource data-source]
+  (log/info (trs "Checking if Database has unrun migrations..."))
+  (if (seq (unrun-migrations data-source))
+    (do
+     (log/info (trs "Database has unrun migrations. Checking if migration lock is taken..."))
+     (wait-for-migration-lock-to-be-cleared liquibase)
```
qnkhuat, line 257: hmmm, this might no longer be needed. `with-scope-locked` will wait for locks anyway, what do you think if we bake this behavior into `with-scope-locked`?

src/metabase/db/setup.clj
```diff
@@ -74,7 +74,10 @@
     (log/info (trs "Setting up Liquibase..."))
     (liquibase/with-liquibase [liquibase conn]
       (try
-       (liquibase/consolidate-liquibase-changesets! conn)
+       (when (= :force direction)
```
qnkhuat, line 77: this is not needed here, since `force-migrate-up-if-needed!` will do the lock release.

test/metabase/db/schema_migrations_test/impl.clj
```diff
@@ -189,14 +189,13 @@
                                                 accept?)
                                     (ChangeSetFilterResult. accept? "decision according to range" (class this)))))]
           change-log-service (.getChangeLogService (ChangeLogHistoryServiceFactory/getInstance) database)]
-      (liquibase/run-in-scope-locked
+      (liquibase/with-scope-locked
        liquibase
```
qnkhuat, line 193:
```suggestion
      (liquibase/with-scope-locked liquibase
```

qnkhuat, line 193:
```suggestion
      (liquibase/with-scope-locked liquibase
```

qnkhuat (commented):
I'm a bit on the fence about nesting the `with-scope-locked` macro. it's too easy and temping to just use it wherever we want. I think a better mechanic is to assert if the lock exists.

for example `consolidate-liquibase-changesets!` does not have to acquire the lock, instead it only need to check whether it's executed under a locked scope.

WDYT?

src/metabase/db/liquibase.clj
crisptrutski: Yes I agree it feels awkward having both methods. Without some refactoring, it's best to leave the redundancy, as `wait-for-migration-lock-to-be-cleared` provides a better error message that mentions the `release-locks` command.

crisptrutski:
> I'm a bit on the fence about nesting the `with-scope-locked`. it's too easy and temping to just use it wherever we want. I think a better mechanic is to assert if locks exists.

I agree it feels "off", I would rather have a clearly defined boundary where the lock is taken, and internal methods which require a lock to be called. That said, I felt nervous about just removing the inner locking block as there were tests that call these internal methods, or alternatively you could say the boundary is porous and ill-defined. 

In a typed language typically the pattern I like is to pass around some kind of "proof" object that's brought into scope by the lock, to clearly indicate which methods assume a lock, but I balked at doing that refactoring here - it felt strange passing around an unused nil-able object, and I wasn't sure what value to use for it. 

Doing a runtime check on the lock service is an interesting idea, which would also catch the lock being accidentally released somehow. That gives me the idea of putting the two together - perhaps I can use the lock-service as the "marker object", and assert that it's holding the lock if I don't want it to be unused. I could make a protocol that hide to hide it's unsafe methods like `releaseLock` too.

Let me know what you think about this approach.

qnkhuat:
> Another idea would just be to add a function that checks lock status via the liquibase variable, and could called it locked-liquibase if we want a clearer indication in the signature.

yes, I was thinking we can create an `assert-liquibase-locked` that takes a `liquibase` and throws an error if it's not locked.

src/metabase/db/liquibase.clj
crisptrutski: I've consolidated the two methods. This fixes a subtle case where the lock could have been taken again between waiting and taking it ourselves, which would have resulted in a very long additional wait, and a totally different lock exception being thrown by Liquibase itself.

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
-           (doseq [^ChangeSet change to-run-migrations]
-             (log/tracef "To run migration %s" (.getId change)))
-           (.update liquibase contexts)
-           (log/info (trs "Migration complete in {0}" (u/format-milliseconds (- (System/currentTimeMillis) start-time)))))
-         (log/info
-          (trs "Migration lock cleared, but nothing to do here! Migrations were finished by another instance.")))))
-    (log/info (trs "No unrun migrations found."))))
+(defn- assert-locked [liquibase]
+  (when-not (.hasChangeLogLock (lock-service liquibase))
+    ;; TODO not sure this is the right exception class to use
```
qnkhuat, line 228: Most of the time, we will throw an ExceptionInfo, and I think it's fine here.

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
-  Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
-  Clojure). This function initializes the database and the resource accessor which are often required."
+   Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
+   Clojure). This function initializes the database and the resource accessor which are often required.
+   The underlying locks are re-entrant, so it is safe to nest these blocks."
   [^Liquibase liquibase f]
-  (let [database (.getDatabase liquibase)
-        ^LockService lock-service (.getLockService (LockServiceFactory/getInstance) database)
-        scope-objects {(.name Scope$Attr/database) database
+
```
qnkhuat, line 239:
```suggestion
```

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
+(def ^:private ^:dynamic *lock-depth* 0)
 
 (defn run-in-scope-locked
   "Run function `f` in a scope on the Liquibase instance `liquibase`.
-  Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
-  Clojure). This function initializes the database and the resource accessor which are often required."
+   Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
+   Clojure). This function initializes the database and the resource accessor which are often required.
+   The underlying locks are re-entrant, so it is safe to nest these blocks."
```
qnkhuat, line 237:
```suggestion
   The underlying locks are re-entrant, so it is safe to nest these blocks."
``
maybe don't encourage it? :p or you should at least let ppl know that they should use `assert-locked` if possible

File: src/metabase/db/liquibase.clj [hunk 7]
```diff
@@ -200,13 +200,16 @@
...
+  (.getLockService (LockServiceFactory/getInstance) (.getDatabase liquibase)))
+
+(defn- wait-for-migration-lock
   "Check and make sure the database isn't locked. If it is, sleep for 2 seconds and then retry several times. There's a
   chance the lock will end up clearing up so we can run migrations normally."
   [^Liquibase liquibase]
   (let [retry-counter (volatile! 0)]
     (u/auto-retry 5
-      (when (migration-lock-exists? liquibase)
+      (when-not (.acquireLock (lock-service liquibase))
```
qnkhuat, line 212: why not keep using `migration-lock-exists?`?

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
-           (log/info (trs "Running {0} migrations ..." unrun-migrations-count))
-           (doseq [^ChangeSet change to-run-migrations]
-             (log/tracef "To run migration %s" (.getId change)))
-           (.update liquibase contexts)
-           (log/info (trs "Migration complete in {0}" (u/format-milliseconds (- (System/currentTimeMillis) start-time)))))
-         (log/info
-          (trs "Migration lock cleared, but nothing to do here! Migrations were finished by another instance.")))))
-    (log/info (trs "No unrun migrations found."))))
+(defn- assert-locked [liquibase]
+  (when-not (.hasChangeLogLock (lock-service liquibase))
```
qnkhuat, line 227: is there any difference between this and how we check lock in `migration-lock-exists?`?

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
-  Clojure). This function initializes the database and the resource accessor which are often required."
+   Liquibase scopes are used to hold configuration and parameters (akin to binding dynamic variables in
+   Clojure). This function initializes the database and the resource accessor which are often required.
+   The underlying locks are re-entrant, so it is safe to nest these blocks."
   [^Liquibase liquibase f]
-  (let [database (.getDatabase liquibase)
-        ^LockService lock-service (.getLockService (LockServiceFactory/getInstance) database)
-        scope-objects {(.name Scope$Attr/database) database
+
+  (when (.hasChangeLogLock (lock-service liquibase))
```
qnkhuat, line 240:
```
(.hasChangeLogLock (lock-service liquibase))
```
this is repeated, looks like we should have a function for it. btw why don' you use `migration-lock-exists?`?

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
+    ;; In production it's better to just take the lock again, it is re-entrant.
+    ;; Fail in dev and CI in order to clean up the code flow.
+    (when-not config/is-prod?
+      (throw (LockException. "Attempted to take a Liquibase lock, but we already are holding it."))))
+
```
qnkhuat, line 245:
```suggestion
```

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
+  `(run-in-scope-locked ~liquibase (fn [] ~@body)))
+
+(defn migrate-up-if-needed!
+  "Run any unrun `liquibase` migrations, if needed."
+  [^Liquibase liquibase ^DataSource data-source]
+  (log/info (trs "Checking if Database has unrun migrations..."))
+  (if (seq (unrun-migrations data-source))
+    (do
+     (log/info (trs "Database has unrun migrations. Checking if migration lock is taken..."))
+     (with-scope-locked liquibase
```
qnkhuat, line 277: not the scope of this PR but I think we can acquire the lock at the top and remove all these double checking for unrun migrations. wdyt?

src/metabase/db/liquibase.clj
```diff
@@ -216,66 +219,92 @@
...
+
+  (let [database      (.getDatabase liquibase)
+        scope-objects {(.name Scope$Attr/database)         database
                        (.name Scope$Attr/resourceAccessor) (.getResourceAccessor liquibase)}]
     (Scope/child ^Map scope-objects
                  (reify Scope$ScopedRunner
                    (run [_]
-                     (.waitForLock lock-service)
+                     (wait-for-migration-lock liquibase)
+                     (assert-locked liquibase)
```
qnkhuat, line 253: why do you need to assert here?

src/metabase/db/liquibase.clj
qnkhuat: also, I think we need to document the locking behavior

src/metabase/db/liquibase.clj
crisptrutski: I wasn't sure what to put in the ex-data in that case, but yeah I think that type is less confusing. I also thought about an IllegalStateException, but ExceptionInfo is pretty ubiquitous for internal application errors.

src/metabase/db/liquibase.clj
crisptrutski: Yeah it's curious that the comment doesn't talk about how the lock itself works at all, will update.

src/metabase/db/liquibase.clj
crisptrutski: The semantics are different - this checks whether the given liquibase instance is holding the lock, versus it potentially being held by another thread or process.

src/metabase/db/liquibase.clj
crisptrutski: The semantics are different - this checks whether the given liquibase instance is holding the lock, versus it potentially being held by another thread or process.

src/metabase/db/liquibase.clj
crisptrutski: This method already encapsulates checking whether it is taken, and immediately returns false if it is. We don't want to call `waitForLock`, because that could trigger an additional pause, which by default is actually in minutes not seconds.

src/metabase/db/liquibase.clj
crisptrutski: Sure I will extract something, perhaps `holding-lock?`

src/metabase/db/liquibase.clj
crisptrutski: I agree. From the PR description:

> There is scope to make [the locking scope] even more coarse, but I think this is good for now.

src/metabase/db/liquibase.clj
crisptrutski: Oh this is left over from debugging - the `.clearCheckSums` call was clearing the lock so I put these statements all over the place to find where that was happening. Will remove.

src/metabase/db/liquibase.clj
crisptrutski: I felt the blank lines helped separate the guards from the rest of the method, but removed.

src/metabase/db/liquibase.clj
qnkhuat: yep, wfm

crisptrutski:
* `master`
  * **PR #38810** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38810?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38810?utm_source=stack-comment).

qnkhuat (approved):
LGTM!

PR #38813, opened 2024-02-15
docs - requiring a filter

jeff-bruemmer:
- Updates embedding docs to reflect new UI.
- Mentions the new feature where you can require a value for a filter.
- Does some other cleanup while I was in there.


docs/embedding/static-embedding-parameters.md
```diff
@@ -115,6 +118,8 @@ You can use locked parameters to display filtered data based on attributes captu
 
 Locked parameters will apply the selected filter values to your original dashboard or SQL question, but they won't be displayed as filter widgets on your embed. Locked parameters may also limit the values that are shown in your [editable filter widgets](#adding-a-filter-widget-to-a-static-embed).
 
+If you just want to require a value for the parameter, see [Always require a value](../dashboards/filters.md#requiring-a-filter).
```
albertoperdomo, line 121: Maybe we should mention that in this case they most likely want to make the filter editable and that it also works with Locked params as well?

albertoperdomo (approved):

docs/dashboards/filters.md
```diff
@@ -106,6 +106,15 @@ If you're trying to connect a filter to a card with a native/SQL questions, you'
 
 You can even wire up filters to text cards, but only if [the text card includes a variable](./introduction.md#including-variables-in-text-cards).
 
+## Requiring a filter
+
+You can tell Metabase to require a filter so that people _must_ filter the dashboard by a certain column on a card.
+
+You might want to require a filter value if:
```
cdeweyx, line 113: Worth noting that required filters need a default value.

cdeweyx (approved):

PR #38814, opened 2024-02-15
serialization: log actual model when couldn't generate entity id

### LINKED ISSUE #38804: Improve logging for Serialization Failures

Failures around serialisation are pretty unique and sometimes debugging them is hard. Currently the error log is not helpful when it comes to entity_id which fail to generate.

For example if a particular collection is missing (could have been deleted by mistake) you might get an error:

`ERROR v2.entity-ids :: Error updating entity ID: Assert failed: (some? entity)`

Which doesn't help in debugging 

Better error messages, pointing to at least the object that is failing to get serialized.

None

**Additional context**
Enterprise customers which handle their own Application DB have unique environments

sansolovyov:
Resolves #38804

### DIFF
--- a/src/metabase/models/collection.clj
+++ b/src/metabase/models/collection.clj
@@ -943,10 +943,12 @@
 (defn- parent-identity-hash [coll]
   (let [parent-id (-> coll
                       (t2/hydrate :parent_id)
-                      :parent_id)]
-    (if parent-id
-      (serdes/identity-hash (t2/select-one Collection :id parent-id))
-      "ROOT")))
+                      :parent_id)
+        parent    (when parent-id (t2/select-one Collection :id parent-id))]
+    (cond
+      (not parent-id) "ROOT"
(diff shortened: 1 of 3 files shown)


src/metabase/models/collection.clj
```diff
@@ -943,9 +943,12 @@
 (defn- parent-identity-hash [coll]
   (let [parent-id (-> coll
                       (t2/hydrate :parent_id)
-                      :parent_id)]
-    (if parent-id
-      (serdes/identity-hash (t2/select-one Collection :id parent-id))
+                      :parent_id)
+        parent (t2/select-one Collection :id parent-id)]
```
crisptrutski, line 947: It bothers me that we still perform a query when parent-id is nil 😞

crisptrutski (approved):

src/metabase/models/collection.clj
piranha: oh right, I'm weird 🙄

PR #38818, opened 2024-02-15
Replace MetabaseSettings.isEnterprise() with isEEBuild()

npretto:
### Description

We currently have a [`MetabaseSettings.isEnterprise()`](https://github.com/metabase/metabase/blob/3fc3fa79a24b73cfc5ba7c8c0be060c5211dc698/frontend/src/metabase/lib/settings.ts#L352) method, but most of that class is deprecated.

This replaces it with a `isEEBuild` function that returns the same value.

The name of the function is not `isEnterprise` to make sure we don't confuse it with the Enterprise **plan**

### DIFF
--- a/frontend/src/metabase/plugins/index.ts
+++ b/frontend/src/metabase/plugins/index.ts
@@ -139,6 +139,7 @@ export const PLUGIN_SELECTORS = {
   getApplicationName: (_state: State) => "Metabase",
   getShowMetabaseLinks: (_state: State) => true,
   getDashboardOverviewId: (_state: State) => undefined,
+  getIsEECode: (_state: State) => false,
 };
 
 export const PLUGIN_FORM_WIDGETS: Record<string, ComponentType<any>> = {};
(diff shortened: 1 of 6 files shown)


frontend/src/metabase/plugins/index.ts
```diff
@@ -350,3 +350,7 @@ export const PLUGIN_DASHBOARD_HEADER = {
 export const PLUGIN_QUERY_BUILDER_HEADER = {
   extraButtons: (_question: Question) => [],
 };
+
+export const PLUGIN_BUILD = {
```
oisincoveney, line 354: maybe we can make this name a little more precise? Something like `PLUGIN_IS_EE_BUILD` or something to help devs discover this a little more easily

npretto: Thanks, yeah I don't think we'll put other things in that plugin so it makes sense

oisincoveney (approved):
nice!

PR #38820, opened 2024-02-15
Remove legacy query from parameter targets

ranquild:

I tried maybe 20 different approaches to migrate this check from legacy query. The most important issue (for future reference) is that we have parameters with field targets in embedding where the query is not available (it's either not present at all or just `{ "type:": "query" }` without even `source-table`, so existing MLv2 functions cannot be used. 

I've came up with a very simple metadata-based retrieval that works for **parameters** that involves parsing the field ref in MLv1. This is based on the fact that parameters "care" only for real, database-based fields (either coming from the table or model metadata overrides), and not arbitrary columns.


frontend/src/metabase/selectors/metadata.ts
```diff
@@ -131,7 +131,9 @@ export const getMetadata: (
       Object.values(tables).map(t => [t.id, createTable(t, metadata)]),
     );
     metadata.fields = Object.fromEntries(
-      Object.values(fields).map(f => [f.uniqueId, createField(f, metadata)]),
+      Object.values(fields)
+        .filter(f => f.uniqueId != null)
```
ranquild, line 135: Remove broken field entries created for field values without any field info

frontend/src/metabase-lib/parameters/utils/cards.ts
```diff
@@ -37,6 +38,9 @@ export function getCardUiParameters(
       };
     }
 
-    return { ...parameter, hasVariableTemplateTagTarget: true };
+    return {
+      ...parameter,
+      hasVariableTemplateTagTarget: !isDimensionTarget(target),
```
ranquild, line 43: Now `getParameterTargetField` doesn't return broken/stub `Field` instances. Previously:
- broken `Field` instances were always returned by `ExpressionDimension`
- broken `Field` instances could be returned by `FieldDimension` in some cases when there was no match by field ref

We now don't create broken `Field` instances. With this change, if `field == null`, it no longer means that the parameter is mapped to a text variable (e.g. no field filter), so we need to add this check here.

frontend/src/metabase-lib/parameters/utils/targets.ts
```diff
@@ -41,16 +40,30 @@ export function getParameterTargetField(
...
+  if (isTemplateTagReference(fieldRef)) {
+    const dimension = TemplateTagDimension.parseMBQL(
+      fieldRef,
+      metadata,
+      question.legacyQuery() as NativeQuery,
+    );
     return dimension?.field();
   }
 
+  if (isConcreteFieldReference(fieldRef)) {
```
ranquild, line 61: We are still in the MLv1 land, not entirely MLv2 based. But we no longer parse the query. The check is always far more stable and predictable than before.

frontend/src/metabase/dashboard/selectors.unit.spec.js
```diff
@@ -55,20 +60,11 @@ const STATE = {
...
-    tables: {},
-    fields: {
-      1: { id: 1 },
-      2: { id: 2 },
-    },
-    metrics: {},
-    segments: {},
-    questions: {},
-  },
+  entities: createMockEntitiesState({
```
ranquild, line 63: Create real, not broken field instances

uladzimirdev (approved):
LGTM

PR #38823, opened 2024-02-15
Allow official collections inside personal collections

### LINKED ISSUE #30236: It's possible to create an official collection that lives in a personal collection
**Describe the bug**
It's possible to create an official collection that lives in a private collection, even though the UI makes you think that it's not possible.

**To Reproduce**
_EE edition, admin role_
1. Go to the root collection 
2. Click on `+ New` and then click "Collection"
3. Give it a name "Official Private Collection"
4. Mark it as official in the modal
image
5. Then change the parent collection from the root to your personal collection
6. Notice that the option to choose between Regular and Official collections disappeared from UI
image
7. Click "Create"
8. The collection was saved as an Official one
image

If we examine the payload and the response from the `POST /api/collection` you'll see that the `authority` is set to `official` in both cases
```js
// request payload
{
"authority_level": "official",
"color": "#509EE3",
"description": null,
"name": "Official Private Collection",
"parent_id": 1
}

// response
{
  "authority_level": "official",
  "description": null,
  "archived": false,
  "slug": "official_private_collection",
  "color": "#509EE3",
  "name": "Official Private Collection",
  "personal_owner_id": null,
  "id": 12,
  "entity_id": "tbp29OzkwD5t1NeHzAFUo",
  "location": "/1/",
  "namespace": null,
  "created_at": "2023-04-19T20:05:34.883359+02:00"
}
```

**Expected behavior**
According to the product, this shouldn't be possible. The UI seems to correctly suggest that you can't choose between "Regular" and the "Official" collections if you chose a Personal collection for its parent. But then we should respect that in the code and override the `authority` field.

local dev, master, https://github.com/metabase/metabase/commit/ae0d9c62dc2b33d22b791f09ba4f7f04ca14d666, H2, Sample Database

**Severity**
🤷‍♂️ 

**Additional context**
Keep in mind it is also possible to move the official collection from a root collection to a personal collection and it will still retain the "Official" status. Maybe we need to handle that case as well and just say that any collection living in a personal collection loses it's official status?

npfitz:
Closes https://github.com/metabase/metabase/issues/30236

### Description
PR to remove the restriction that official collections cannot exist inside of personal collections. The original issue revolved around forms keeping the `authority_level` prop after hiding the input, but after discussion with the product team it was decided that we should remove this restriction entierly. This change updates the form, as well as restores the "Make Official" buttons in the collection header (being an admin is still a restriction).

### How to verify

1. Go to New -> Collection and set your personal collection to be the parent
2. Note that the authority level picker is still present. Create the collection
3. Now in the collection header you should be able to toggle between normal and official collections

### Checklist

### DIFF
--- a/e2e/test/scenarios/organization/official-collections.cy.spec.js
+++ b/e2e/test/scenarios/organization/official-collections.cy.spec.js
@@ -107,26 +107,35 @@ describeEE("official collections", () => {
       });
     });
 
-    it("should not be able to manage collection authority level for personal collections and their children", () => {
+    it("should be able to manage collection authority level for personal collections and their children (metabase#30236)", () => {
       cy.visit("/collection/root");
 
       openCollection("Your personal collection");
       getCollectionActions().within(() => {
-        cy.icon("ellipsis").should("not.exist");
+        cy.icon("ellipsis").should("exist");
+        cy.icon("ellipsis").click();
       });
+
+      popover().findByText("Make collection official").should("exist");
 
       openNewCollectionItemFlowFor("collection");
       modal().within(() => {
-        assertNoCollectionTypeInput();
+        assertHasCollectionTypeInput();
         cy.findByLabelText("Name").type("Personal collection child");
         cy.button("Create").click();
       });
 
       openCollection("Personal collection child");
 
+      getCollectionActions().within(() => {
+        cy.icon("ellipsis").should("exist");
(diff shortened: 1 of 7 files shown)


iethree (approved):
Frontend looks great!

noahmoss (approved):
BE lgtm

Issue #38492, opened 2024-02-06
Getting fields in (Redshift) is slow


paoliniluis:
### Describe the bug
(leaving the old description on the Additional context section of the issue)

Before 48, we used to go table by table on each schema checking for permissions. This was fixed already (checked on my machine and it takes 10 min in v48, super fast to what it was before), but now the problem is on getting fields: for every field we will go one by one inserting those into the DB which takes a huge amount of time

### To Reproduce

1) just connect a redshift cluster with Metabase and see the sync logs, it takes a massive amount of time for lots of tables

### Logs

Above

### Information about your Metabase installation

```JSON
I believe it's been like this forever
```

### Severity

P1

### Additional context

OLD ISSUE here:
Seems that when we check for SELECT we do
```
2024-02-06 22:01:52,345 TRACE sync.describe-database :: Checking for SELECT privileges for "federated_sample1"."_metabase_metadata" with query ["SELECT TRUE AS \"_\" FROM \"federated_sample1\".\"_metabase_metadata\" WHERE 1 <> 1 LIMIT 0"]
2024-02-06 22:01:53,047 TRACE sync.describe-database :: [redshift] SELECT TRUE AS "_" FROM "federated_sample1"."_metabase_metadata" WHERE 1 <> 1 LIMIT 0
2024-02-06 22:01:53,965 TRACE sync.describe-database :: SELECT privileges confirmed
2024-02-06 22:01:53,969 TRACE sync.describe-database :: Checking for SELECT privileges for "federated_sample1"."accounts" with query ["SELECT TRUE AS \"_\" FROM \"federated_sample1\".\"accounts\" WHERE 1 <> 1 LIMIT 0"]
2024-02-06 22:01:54,683 TRACE sync.describe-database :: [redshift] SELECT TRUE AS "_" FROM "federated_sample1"."accounts" WHERE 1 <> 1 LIMIT 0
2024-02-06 22:01:55,472 TRACE sync.describe-database :: SELECT privileges confirmed
2024-02-06 22:01:55,476 TRACE sync.describe-database :: Checking for SELECT privileges for "federated_sample1"."analytic_events" with query ["SELECT TRUE AS \"_\" FROM \"federated_sample1\".\"analytic_events\" WHERE 1 <> 1 LIMIT 0"]
[...]

PR #24910, opened 2022-08-19
Fix for redshift late-binding view sync error

noahmoss:
Fixes #21215

Some specific field types in late-binding views cause the Redshift JDBC driver to throw an error on the `.getColumns` call when fetching field metadata (https://github.com/metabase/metabase/blob/nm-redshift-latebinding/src/metabase/driver/sql_jdbc/sync/describe_table.clj#L89)

Bumping the driver version seems to fix some of these cases, but not all. Specifically, a case statement that produces a mix of numeric and null values still results in an error.

To fix this, we can use the fallback method for determining field metadata in these cases, which uses a `SELECT *` query rather than JDBC methods. I've done this specifically for Redshift by converting `describe-table-fields` to a driver method with a custom Redshift implementation that has a try/catch around the default method. I've also added a test for this case with an example derived from the one in #21215.

PR #38828, opened 2024-02-15
Faster sync-fields

calherries:
This is part 2 towards resolving https://github.com/metabase/metabase/issues/38492
Part 1 is https://github.com/metabase/metabase/pull/38970

This PR adds an alternative implementation of the sync-fields step of sync where we sync the foreign key information of a database.

It adds a new driver feature: `describe-fields`. If a driver supports this feature, they need to implement the new driver method `metabase.driver/describe-fields`. If the driver doesn't support the feature, we'll continue to use `metabase.driver/describe-table`. The plan for driver authors to implement `describe-fields` in terms of `driver/describe-table`, which should be a really simple change for most drivers, assuming it doesn't need to be performant for large DBs.

The way this works is instead of querying the customer DB one table at a time using the `getColumns` JDBC method, we execute a single query that gets all the data at once with paginated results. We then reduce over those results, updating our App DB one foreign key at a time. This approach mirrors https://github.com/metabase/metabase/pull/38970

I've tested this with the "massive" redshift instance, and this is how long each sync step took:
- "sync-tables": 13.3 s
- "sync-fields": 1.0 mins (this is the step we're speeding up)
- "sync-fks": 1.3 s (there are no FKs to sync though)
- total time for all sync metadata steps: 5.0 minutes

Before this PR, `sync-fields` took hours (too long to bother measuring precisely anyway).


src/metabase/driver/sql_jdbc/sync/describe_table.clj
```diff
@@ -121,12 +121,25 @@
          (merge
            {:name                       column-name
             :database-type              (.getString rs "TYPE_NAME")
+            :table-schema               (.getString rs "TABLE_SCHEM")
```
luizarakaki, line 124: typo?

calherries: Nope, these are from the jdbc spec

calherries:
Note to self: remember to clean up this code https://github.com/metabase/metabase/pull/24910/files

docs/developers-guide/driver-changelog.md
```diff
@@ -4,6 +4,14 @@ title: Driver interface changelog
...
 # Driver Interface Changelog
 
+## Metabase 0.50.0
+
+- The multimethod `metabase.driver/describe-fields` has been added. The method needs to be
+  implemented if the database supports the `:fast-sync-fields` feature.
+
+- The multimethod `metabase.driver/describe-fks` has been added. The method needs to be implemented if the database
+  supports the `:foreign-keys` and `:fast-sync-fks` features.
+
```
crisptrutski, line 14: Do you mean for this PR to be stacked on #38970? It seems to include it.

calherries: yes

calherries:
Note to self: check database-position is not different

calherries:
depends on https://github.com/metabase/metabase/pull/40006

docs/developers-guide/driver-changelog.md
```diff
@@ -6,6 +6,12 @@ title: Driver interface changelog
 
 ## Metabase 0.50.0
```
qnkhuat, line 7: should be 49

docs/developers-guide/driver-changelog.md
```diff
@@ -6,6 +6,12 @@ title: Driver interface changelog
...
 
+- The multimethod `metabase.driver/describe-fields` has been added. The method needs to be
+  implemented if the database supports the `:fast-sync-fields` feature.
+
+- The multimethod `metabase.driver.sql-jdbc.sync.describe-table/describe-fields-sql` has been added. The method needs
+  to be implemented if you want to use the default JDBC implementation of `metabase.driver/describe-fields`.
```
qnkhuat, line 13: should be 49

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -359,18 +399,6 @@
...
-  [driver conn {schema :schema, table-name :name :as table} db-name-or-nil]
-  (let [parent-method (get-method sql-jdbc.describe-table/describe-table-fields :sql-jdbc)]
-    (try (parent-method driver conn table db-name-or-nil)
-         (catch Exception e
-           (log/error e (trs "Error fetching field metadata for table {0}" table-name))
-           ;; Use the fallback method (a SELECT * query) if the JDBC driver throws an exception (#21215)
-           (into
-            #{}
-            (sql-jdbc.describe-table/describe-table-fields-xf driver table)
-            (sql-jdbc.describe-table/fallback-fields-metadata-from-select-query driver conn db-name-or-nil schema table-name))))))
```
calherries, line 372: This fallback was added as part of https://github.com/metabase/metabase/pull/24910. 

The test for that issue, `redshift-lbv-sync-error-test`, passes without the fallback now, so I believe the root cause of the issue was with how we were calling `getColumns`.

test/metabase/driver/sql_jdbc/sync/describe_table_test.clj
```diff
@@ -126,8 +149,31 @@
            {:fk-column-name "VENUE_ID", :dest-table {:name "VENUES", :schema "PUBLIC"}, :dest-column-name "ID"}}
          (sql-jdbc.describe-table/describe-table-fks :h2 (mt/id) {:name "CHECKINS"}))))
 
+(deftest describe-fields-or-table-test
```
calherries, line 152: I didn't find any test coverage for `:database-position` for the default SQL-JDBC implementation so I added this test. It also sense checks a few other things, like sensible base types and names.

src/metabase/driver/postgres.clj
qnkhuat, line 1: why do we have these changes?

test/metabase/driver/sql_jdbc_test.clj
```diff
@@ -79,6 +80,19 @@
          #_{:clj-kondo/ignore [:deprecated-var]}
          (driver/describe-table-fks :h2 (mt/db) (t2/select-one Table :id (mt/id :venues))))))
 
+(deftest describe-fields-sync-with-composite-pks-test
```
calherries, line 83: I couldn't find any test coverage for composite PKs so I added this test.

src/metabase/sync/interface.clj
```diff
@@ -65,16 +66,15 @@
...
-   [:name                         ::lib.schema.common/non-blank-string]
-   [:schema                       [:maybe ::lib.schema.common/non-blank-string]]
-   [:fields                       [:set TableMetadataField]]
-   [:description {:optional true} [:maybe :string]]])
-
-(def TableMetadata
-  "Schema for the expected output of [[metabase.driver/describe-table]]."
-  [:ref ::TableMetadata])
+(mr/def ::FieldMetadataEntry
+  (-> (mr/schema ::TableMetadataField)
```
qnkhuat, line 70: why did you remove description? and any reason why you use these helpers instead of the normal syntax?

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -241,8 +241,8 @@
                view-nm))
           (let [table-id (t2/select-one-pk Table :db_id (u/the-id database), :name view-nm)]
               ;; and its columns' :base_type should have been identified correctly
-            (is (= [{:name "numeric_col",   :database_type "numeric(10,2)",         :base_type :type/Decimal}
-                    {:name "weird_varchar", :database_type "character varying(50)", :base_type :type/Text}]
+            (is (= [{:name "numeric_col",   :database_type "numeric",         :base_type :type/Decimal}
```
qnkhuat, line 244: any idea why we lose the precision here?

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -76,6 +77,40 @@
...
 
+(defmethod sql-jdbc.sync/describe-fields-sql :redshift
+  [driver & {:keys [schema-names table-names]}]
+  (sql/format {:select [[:c.column_name :name]
+                        [:c.data_type :database-type]
+                        [[:- :c.ordinal_position 1] :database-position]
+                        [:c.schema_name :table-schema]
+                        [:c.table_name :table-name]
+                        [nil :database-is-auto-increment] ; only needed for actions, which redshift doesn't support yet
+                        [nil :database-required]          ; only needed for actions, which redshift doesn't support yet
```
qnkhuat, line 88: probably don't even need to return it as these fields are optional

src/metabase/driver/postgres.clj
calherries: The subset of the  `getColumns` implementation we were using converts `timestamp with time zone` types to `timestamptz`. So previously fields with a `database-type` of `timestamptz` now have `timestamp with time zone` . Fortunately we have tests that caught this.

src/metabase/sync/interface.clj
calherries: Good question. We don't use the description anywhere. We get the description from `DatabaseMetadata` instead. https://github.com/metabase/metabase/pull/38828/files#diff-8eb3ce5146f34119083048007ec1c733c4c77e14b40be0aa84e583f1cc3ab6d1R10-R16

I removed the requirement to get the description from the interface as a way to show driver authors that they can  switch from `describe-table` to `describe-fields` without losing any information, since we only use the `fields` key from `describe-table`.

src/metabase/util/malli/schema.clj
```diff
@@ -368,3 +368,10 @@
   (mu/with-api-error-message
    [:re u/uuid-regex]
    (deferred-tru "value must be a valid UUID.")))
+
+(defn CollectionOf
+  "Helper for creating schemas to check whether something is an instance of a collection."
+  [item-schema]
+  [:fn
+   {:error/message (format "Collection of %s" item-schema)}
+   #(and (coll? %) (every? (partial mc/validate item-schema) %))])
```
calherries, line 377: This is actually dead code, but I think it's generally useful so I'm inclined to keep it around.

At the moment we have no way to type something that could be a set or a vector for example.

crisptrutski (commented):
Getting notified that diffs are out of date, flushing comments before I continue.

src/metabase/util/malli/schema.clj
crisptrutski: Surprised it's not a built-in!

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -87,9 +121,15 @@
 ;; custom Redshift type handling
 
 (def ^:private database-type->base-type
-  (sql-jdbc.sync/pattern-based-database-type->base-type
-   [[#"(?i)CHARACTER VARYING" :type/Text]       ; Redshift uses CHARACTER VARYING (N) as a synonym for VARCHAR(N)
-    [#"(?i)NUMERIC"           :type/Decimal]])) ; and also has a NUMERIC(P,S) type, which is the same as DECIMAL(P,S)
```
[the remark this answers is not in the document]
calherries: Thanks, this got deleted by accident 9edacd06c23a2af91bb32cc0d9d8577825a561eb

docs/developers-guide/driver-changelog.md
calherries: I've optimistically changed it to 49.1 in 137c1404a72ed9b0c3d1ad20e1a21778c73ab8db

.clj-kondo/config.edn
```diff
@@ -355,7 +355,7 @@
     metabase.sync.field-values                                      field-values
     metabase.sync.interface                                         i
     metabase.sync.schedules                                         sync.schedules
-    metabase.sync.sync-metadata.fields.fetch-metadata               fetch-metadata
+    metabase.sync.sync-metadata.fields.our-metadata                 our-metadata
```
qnkhuat, line 358:
```suggestion
    metabase.sync.sync-metadata.fields.our-metadata                 fields.our-metadata
```

without this prefix, it's hard to know whether it's fields metadata or table metadata or db metadata

src/metabase/sync/sync_metadata/fields.clj
```diff
@@ -39,29 +39,62 @@
...
+   (+ (sync-instances/sync-instances! table db-metadata (our-metadata/our-metadata table))
      ;; Now that tables are synced and fields created as needed make sure field properties are in sync.
      ;; Re-fetch our metadata because there might be somethings that have changed after calling
      ;; `sync-instances`
-     (sync-metadata/update-metadata! table db-metadata (fetch-metadata/our-metadata table))))
+      (sync-metadata/update-metadata! table db-metadata (our-metadata/our-metadata table))))
+
+(mu/defn sync-fields-for-db!
+  "Sync the Fields in the Metabase application database for a specific `table`."
+  [database :- i/DatabaseInstance]
```
qnkhuat, line 73: you can make this take `tables` and create a `[schema name] -> table` mapping that you can use inside the reducer to get the table.

Always good to reduce the number of DB calls

test/metabase/driver/sql_jdbc/sync/describe_table_test.clj
```diff
@@ -79,6 +92,16 @@
              :json-unfolding             false}}}
          (sql-jdbc.describe-table/describe-table :h2 (mt/id) {:name "VENUES"}))))
 
+(defn- describe-fields-for-table [db table]
+  (let [driver (driver.u/database->driver db)]
+    (sort-by :database-position
+             (if (driver/database-supports? driver :describe-fields db)
+               (vec (sql-jdbc.describe-table/describe-fields driver db
```
crisptrutski, line 99: nit: it seems wasteful to materialize a full vector just then build an array and sort it, for large very N we could optimize this further. Definitely not a bottleneck though.

src/metabase/sync/sync_metadata/fields.clj
```diff
@@ -39,29 +39,62 @@
...
+   [metabase.util.malli.schema :as ms]
+   [toucan2.core :as t2]))
 
 ;;; +----------------------------------------------------------------------------------------------------------------+
 ;;; |                                            PUTTING IT ALL TOGETHER                                             |
 ;;; +----------------------------------------------------------------------------------------------------------------+
 
-(mu/defn ^:private sync-and-update! :- ms/IntGreaterThanOrEqualToZero
-  "Sync Field instances (i.e., rows in the Field table in the Metabase application DB) for a Table, and update metadata
+ (mu/defn ^:private sync-and-update! :- ms/IntGreaterThanOrEqualToZero
```
crisptrutski, line 60: Accidental indentation?

crisptrutski (approved):
Nicely done 👌

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
calherries: The subset of the implementation of `getColumns` we used previouisly uses `pg_type.typbasetype`. See
https://github.com/aws/amazon-redshift-jdbc-driver/blob/master/src/main/java/com/amazon/redshift/jdbc/RedshiftDatabaseMetaData.java#L2365C24-L2365C35
But now we're using `svv_all_columns.data_type`, which stores the precision in a separate column. 
See https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_ALL_COLUMNS.html

What gives me confidence that these are roughly equivalent is that the JDBC driver implementation uses either, depending on the pattern of schemas that you select from. For example, if you select a schema pattern that corresponds to a "LOCAL_SCHEMA_QUERY", it will use `pg_type.typbasetype`, but if you specify a schema pattern that corresponds to a `NO_SCHEMA_UNIVERSAL_QUERY` it can use `svv_all_columns.data_type`.

docs/developers-guide/driver-changelog.md
```diff
@@ -4,6 +4,16 @@ title: Driver interface changelog
 
 # Driver Interface Changelog
 
+## Metabase 0.49.1
```
qnkhuat, line 7: Oh, didn't you tell me that we don't introduce driver methods in the patch releases?

src/metabase/sync/sync_metadata/fields.clj
crisptrutski: I was also bothered by the N+1 table queries, but ended up not commenting as for large N we'd rather not fetch all the tables into memory.

It seems unfortunate to me that there's a deep nest of methods to require a full table model, because in practice they're mostly using the id, name, and schema mostly. I suspect we could just add one or two more table fields to the educed rows and get rid of this query, but it would be brittle to use a partially populated instance, and refactoring all those methods to depend only on their required table fields would be pretty hairy and ugly.

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -87,9 +121,15 @@
...
+  (some-fn (sql-jdbc.sync/pattern-based-database-type->base-type
+            [[#"(?i)CHARACTER VARYING" :type/Text]       ; Redshift uses CHARACTER VARYING (N) as a synonym for VARCHAR(N)
+             [#"(?i)NUMERIC"           :type/Decimal]])
+           {:super       :type/*    ; (requested support in metabase#36642)
```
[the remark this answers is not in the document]
calherries: I fear that type of comment could get out of date here. If you have any questions about `type/*` it should be documented `metabase.types` namespace. I see that it isn't, but I don't think I could do it justice if I tried right now

I'm mostly listing these types to document these types are deliberately not being handled, as opposed to accidentally not being handled. Removing the log warnings is a nice side-effect though.

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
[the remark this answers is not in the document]
calherries: As far as I'm aware, we don't use the size / precision data from these strings anywhere.

src/metabase/driver.clj
```diff
@@ -567,7 +582,10 @@
     :index-info
 
     ;; Does the driver support a faster `sync-fks` step by fetching all FK metadata in a single collection?
-    :describe-fks})
+    :describe-fks
+
+    ;; Does the driver support a faster `sync-fields` step by fetching all FK metadata in a single collection?
```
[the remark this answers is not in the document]
calherries: nice idea, since the feature name is closely coupled to the multimethod I can't see this comment becoming out of date. done 0b1742f0c8289bd97850a477f5f6aa171ac1ae30

src/metabase/sync/fetch_metadata.clj
```diff
@@ -16,11 +16,22 @@
...
 
-(mu/defn table-metadata :- i/TableMetadata
+(mu/defn fields-metadata
+  "Effectively a wrapper for [[metabase.driver/describe-fields]] that also validates the output against the schema."
+  [database :- i/DatabaseInstance & {:as args}]
+  (cond->> (driver/describe-fields (driver.u/database->driver database) database args)
+    ;; Validate the output against the schema, except in prod.
+    ;; This is a workaround for the fact that [[mu/defn]] can't check reducible collections yet
+    (not config/is-prod?)
+    (eduction (map #(mu.fn/validate-output {} i/FieldMetadataEntry %)))))
```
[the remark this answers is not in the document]
calherries: Yeah I like the idea of coupling this to `metabase.util.malli.fn/fn`, and using `mu.fn/instrument-ns?` convenient. done e10924a1988450bcee4c9bf5a9c140c12edf7569

src/metabase/sync/sync_metadata/fields.clj
calherries: What Chris said, I'm fetching each table because I'm avoiding fetching all the tables into memory. The good thing is app DB queries should be quite cheap compared to customer DB queries, so this is still a huge win.

In the future we can avoid the N+1 by doing a refactoring like the one I did for `sync-fks`  as I did here https://github.com/metabase/metabase/pull/39679/files#diff-e09c063342055336a262f66623a6a1783a89871e4426c33c92f8f9b5f2bcaaec

But I don't think it's worth it. That refactoring took longer than I thought because of the quirks between different app DBs.

test/metabase/driver/sql_jdbc/sync/describe_table_test.clj
calherries: yeah, this is just a test helper for very small N. Not going to do this

crisptrutski: Somehow I missed this was just for a single table.

docs/developers-guide/driver-changelog.md
calherries: Yeah, maybe I did say that :) I can't remember how I justified it though, probably because we've never done it before. But putting it in the 0.49.0 section doesn't make much sense.

src/metabase/sync/sync_metadata/fields.clj
qnkhuat: okay, that makes sense; in that case, you need to "not" fetch all tables in metabase.sync.sync-metadata.fields line 117

.clj-kondo/config.edn
calherries: agreed. we don't use this namespace much so the extra characters doesn't have much downside 
40df070bd39e3ed15ece0bd36c50fa1c80daef2a

calherries:
@qnkhuat can you take another look? It would be nice to get this in 49.0

qnkhuat (approved):
Resolve [this](https://github.com/metabase/metabase/pull/38828/files#r1524547151) then it LGTM!

calherries:
@qnkhuat done

PR #38829, opened 2024-02-15
Flow :semantic-type through QP for aggregations

### LINKED ISSUE #38022: Currency Data Type Is Not Inferred in Summarized Visualizations
### Describe the bug

When summarizing currency fields, the currency type / formatting isn't carried through to the resulting summarized visualization. This works as expected in v47.

### To Reproduce

1. Browse a table that has a currency field type
2. Summarize the currency field and group it by another value to create a chart
3. Visualize the data and look at the axis that that references the currency data
4. Note that the axis doesn't have a currency symbol

### Expected behavior

When you summarize a a currency field the resulting izualization should infer the data type and automatically label the summarized data as currency in the resulting visualization.

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "Win32",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.version": "11.0.21",
    "os.name": "Linux",
    "os.version": "5.15.133.1-microsoft-standard-WSL2",
    "user.language": "en",
    "user.timezone": "GMT"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "mysql",
      "mongo",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "mysql",
      "database": {
        "name": "MySQL",
        "version": "8.2.0"
      },
      "jdbc-driver": {
        "name": "MariaDB Connector/J",
        "version": "2.7.10"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2024-01-11",
      "tag": "v0.48.3",
      "hash": "80d8323"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

When users summarize currency fields they must edit the resulting visualization to display the currency symbols for every single viz

### Additional context

- [Loom Recording](https://www.loom.com/share/801a6580a45549dcbbfea1a9da709c8b)
- Zendesk Ticket # 24224

adam-james-v:
Fixes: #38022

### Before
Before some changes to the QP, I think probably some mlv2 changes, particularly those affecting metadata calculation, we would return a `:semantic_type` on aggregations. Here's a screenshot of Metabase 47 (left) and master (49-RC1, but this existed in 48 as well I think) (right).

This example is showing the response from `/api/dataset` which runs queries for unsaved questions. Notice that in 47 the semantic type key does exist in the response but it does not exist in our latest code. This causes the sum to not be formatted correctly (there's no `$` on the sum on the right).

image

### After:
This method in `metabase.lib.aggregation`:

```clojure
(defmethod lib.metadata.calculation/metadata-method ::aggregation
  [query stage-number [_tag _opts first-arg :as clause]]
  (merge
   ;; flow the `:options` from the field we're aggregating. This is important, for some reason.
   ;; See [[metabase.query-processor-test.aggregation-test/field-settings-for-aggregate-fields-test]]
   (when first-arg
     ;; Here I also select :semantic-type from the field's metadata. 
     (select-keys (lib.metadata.calculation/metadata query stage-number first-arg) [:settings :semantic-type]))
   ((get-method lib.metadata.calculation/metadata-method :default) query stage-number clause)))
```

Now includes the :semantic-type from the field. This method does basically then proceed to use the :default metadata method, which I did consider changing, but I opted instead to change just the method for `::aggregation` because that's where the issue manifests. 

If I don't aggregate at all, the semantic_type of each column is returned as normal, which means that the metadata *is* being fetched/calculated correctly, it just was not being used completely on aggregations.

image


test/metabase/query_processor_test/aggregation_test.clj
```diff
@@ -284,6 +284,16 @@
                (or (-> results mt/cols first)
                    results)))))))
 
+(deftest semantic-type-for-aggregate-fields-test
```
markbastian, line 287: Might want to put a ticket number for https://github.com/metabase/metabase/issues/38022 in the test name or the testing string.

markbastian (approved):
Left one nonblocking comment, but looks good.

bshepherdson (commented):
I think there's some more test cases to exercise here.

The argument to an aggregation and the aggregation's own result don't necessarily have the same semantic type. On the other hand I see that `metadata-method ::aggregation` is what you're editing, and they all use that except `:count-where`, which overrides `:semantic-type :type/Quantity`.

You'll want to check all the aggregation operators - if they take an argument, make sure they either share the `:semantic-type` of the first argument (eg. `sum`, `avg`, `min`, `max`), or that there's a specific `metadata-method` (eg. `:count-where`).

I think this might already be broken for `distinct`, which should be a `:type/Quantity` rather than blank (now) or copied from the column (your PR).

Does this make sense?

bshepherdson:
Yes, that's all the aggregations.

I think individually these changes are good, but I think we should take advantage of multimethod dispatch to reduce duplication here, like is done elsewhere in this file. There's really only three cases: always a `:type/Quantity` (eg. `:count`), keep the inner `:semantic-type` (eg. `SUM`), and drop `:semantic-type` for the unit-less `:var`.

See the [docs](https://clojure.org/reference/multimethods) if you're not familiar with `derive` and hierarchies. There's a few existing ones in this namespace; I think we can reuse `::count-aggregation`.

```clojure
(lib.hierarchy/derive ::count-aggregation ::quantity-aggregation)
(lib.hierarchy/derive :distinct ::quantity-aggregation)

(doseq [tag [:percentile :var]]
  (lib.hierarchy/derive tag ::no-semantic-type))

(defmethod lib.metadata.calculation/metadata-method
  ::quantity-aggregation
  ; ...
 (assoc ... :semantic-type :type/Quantity))

;; The default preserves the semantic type.

;; But for ::no-semantic-type we should drop it.
(defmethod lib.metadata.calculation/metadata-method
  ::no-semantic-type
  ; ...
(dissoc ... :semantic-type))
```

bshepherdson (changes requested):
Sorry to send this back, but hopefully this is a modest edit to reduce duplication.

bshepherdson (approved):

bshepherdson (approved):
Good stuff!

PR #38582, opened 2024-02-08
Fix sidebar and home loading indefinitely when database data fails to load

sloansparger:
Closes https://github.com/metabase/metabase/issues/32793

### Description

https://github.com/metabase/metabase/assets/7104357/02a93eb4-e296-4e13-a2f7-e9655a23eadf

### How to verify

1. With `master` as your branch replace the returned object with an empty string
https://github.com/metabase/metabase/blob/master/src/metabase/api/database.clj#L301-L302
3. See the page infinitely loads
4. Checkout this branch while keeping the above error
5. See that the page works!

### Checklist

- [ ] Tests have been added/updated to cover changes in this PR

PR #38833, opened 2024-02-15
Revert "Fix sidebar and home loading indefinitely when database data fails to load"

sloansparger:
@iethree discovered an unintended consequence of #38582  [see comment](https://github.com/metabase/metabase/pull/38582#issuecomment-1947315244)
305221147-2e1faa2f-7921-4a18-970d-1aa25179d079
Reverting this PR for now until I can make get a better fix in..


iethree:
I don't think this needs a revert if you can fix it in a day or two

PR #38836, opened 2024-02-15
Add per viz-type card template sizes in x-rays

markbastian:
Previously, any card without a size (width or height) set in their yaml definition was set to a default of `default-card-width` or `default-card-width`, respectively. This PR adds a map of default and min dimensions per viz type (the `:display`) key and fills that in for each template. The template is merged into this result, keeping anything that already existed in the yaml file.

Prior to this PR, the default card sizes were found in `sizes.ts` and were inaccessible from the BE. This PR consolidates those changes into `metabase.shared.automagic-dashboards.constants`.

Note that @kulyk and @JesseSDevaney might want to consider further consolidation of dashboard related constants into this namespace.

### DIFF
--- a/src/metabase/automagic_dashboards/dashboard_templates.clj
+++ b/src/metabase/automagic_dashboards/dashboard_templates.clj
@@ -287,19 +287,20 @@
   [dashboard-template]
   (transduce (map (comp count ancestors)) + (:applies_to dashboard-template)))
 
+(defn- ensure-default-card-sizes
+  "Given a card definition from a template, fill in the card template with default width and height
+  values based on the template display type if those dimensions aren't already present."
+  [card-spec]
+  (update-vals
+    card-spec
+    (fn [{:keys [visualization] :as card-spec}]
+      (let [defaults (get-in magic.constants/card-size-defaults [(keyword visualization) :default])]
+        (into defaults card-spec)))))
+
 (defn- set-default-card-dimensions
   "Update the card template dimensions to align with the default FE dimensions."
   [dashboard-template]
-  (update dashboard-template :cards
-          (fn [cards]
-            (mapv
-              (fn [card-spec]
-                (update-vals
-                  card-spec
-                  (fn [{:keys [visualization] :as card-spec}]
-                    (let [defaults (get-in magic.constants/card-size-defaults [(keyword visualization) :default])]
-                      (into defaults card-spec)))))
-              cards))))
+  (update dashboard-template :cards #(mapv ensure-default-card-sizes %)))
 
 (defn- make-dashboard-template
   [entity-type {:keys [cards] :as r}]
--- a/src/metabase/automagic_dashboards/dashboard_templates.clj
+++ b/src/metabase/automagic_dashboards/dashboard_templates.clj
@@ -6,6 +6,7 @@
    [clojure.string :as str]
    [metabase.automagic-dashboards.populate :as populate]
    [metabase.query-processor.util :as qp.util]
+   [metabase.shared.automagic-dashboards.constants :as magic.constants]
    [metabase.util :as u]
    [metabase.util.files :as u.files]
    [metabase.util.i18n :as i18n :refer [deferred-trs LocalizedString]]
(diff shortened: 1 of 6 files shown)


src/metabase/automagic_dashboards/dashboard_templates.clj
```diff
@@ -286,6 +287,20 @@
...
+  [dashboard-template]
+  (update dashboard-template :cards
+          (fn [cards]
+            (mapv
+              (fn [card-spec]
+                (update-vals
+                  card-spec
+                  (fn [{:keys [visualization] :as card-spec}]
+                    (let [defaults (get-in magic.constants/card-size-defaults [(keyword visualization) :default])]
+                      (into defaults card-spec)))))
```
bshepherdson, line 301: This function would be a lot easier to read if this were a named helper. Then the outer function becomes `(update dashboard-template :cards #(mapv new-helper %))`.

src/metabase/shared/automagic_dashboards/constants.cljc
```diff
@@ -0,0 +1,34 @@
...
+   :map         {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :object      {:min {:width 4 :height 3} :default {:width 12 :height 9}}
+   :row         {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :heading     {:min {:width 1 :height 1} :default {:width GRID_WIDTH}}
+   :text        {:min {:width 1 :height 1} :default {:width 12 :height 3}}})
+
+(def ^:export card-size-defaults
+  "Default card sizes per visualization type"
+  #?(:clj  card-size-defaults*
+     :cljs (clj->js card-size-defaults*)))
```
bshepherdson, line 34: I'd do this same-data-different-format magic the other way around. Rename `card-size-defaults*` to remove the star - that's the symbol for use from **all** Clojure code. Then add
```clojure
#?(:cljs (def ^:export card-size-defaults-json (clj->js card-size-defaults)))
```
and import that from TS.

I'd certainly be surprised if I were working in some CLJC file, `:require`d some CLJC namespace, and received JS data! We use the approach I'm suggesting in MLv2: most of the files are all Clojure all the time and have very few `#?(:cljs ...)` conditionals, then `metabase.lib.js` wraps many of those functions with extra JS conversion for inputs and outputs.

markbastian: That make sense.

src/metabase/automagic_dashboards/dashboard_templates.clj
markbastian: Done

bshepherdson (approved):

kulyk (approved):

src/metabase/shared/automagic_dashboards/constants.cljc
JesseSDevaney, line 1: Nit: Can we change the namespace name to be `.dashboards.constants` and re-locate the file to `./shared/dashboards/constants.cljc` so that it is more obvious that these constants apply to all dashboards, not just X-Rays?

src/metabase/shared/automagic_dashboards/constants.cljc
```diff
@@ -0,0 +1,34 @@
...
+   :action      {:min {:width 1 :height 1} :default {:width 4 :height 1}}
+   :area        {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :pivot       {:min {:width 4 :height 3} :default {:width 12 :height 9}}
+   :funnel      {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :progress    {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :smartscalar {:min {:width 2 :height 2} :default {:width 6 :height 3}}
+   :map         {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :object      {:min {:width 4 :height 3} :default {:width 12 :height 9}}
+   :row         {:min {:width 4 :height 3} :default {:width 12 :height 6}}
+   :heading     {:min {:width 1 :height 1} :default {:width GRID_WIDTH}}
```
JesseSDevaney, line 29:
```suggestion
   :heading     {:min {:width 1 :height 1} :default {:width GRID_WIDTH :height 1}}
```

JesseSDevaney (changes requested):
99.99% LGTM

There is one thing we should address before merging, and a small nit.

JesseSDevaney (approved):

PR #38837, opened 2024-02-15
Appends follow-up UI fixes

iethree:
### Description

A couple little things came up during today's demo:

1) we should auto-select the only model to append if there's only 1
only-one

2) there was a bug where[ if you did certain things in the ui](https://www.loom.com/share/4f65e82e1d054aebbc58716d3a31702a) (like create a new question), we would re-trigger an error modal.

### Tests
- [x] tests have been added

### DIFF
--- a/frontend/src/metabase/collections/components/ModelUploadModal.tsx
+++ b/frontend/src/metabase/collections/components/ModelUploadModal.tsx
@@ -13,6 +13,8 @@ import {
 import { useSearchListQuery } from "metabase/common/hooks";
 import type { CollectionId, TableId, CardId } from "metabase-types/api";
 
+import { findLastEditedCollectionItem } from "./utils";
+
 enum UploadMode {
   append = "append",
   create = "create",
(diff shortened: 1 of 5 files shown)


mazameli:
> we should auto-select the only model to append if there's only 1

IMO, if there are _any_ to append to, we should just auto-select one of them, since even dumbly selecting the first one alphabetically has a higher probability of being the one the user intended (>0% chance of being right) than the probability that no selection is the correct one (exactly 0%, haha). If it's easy to be slightly smarter (like auto selecting the one with the most recent last_edited_at date or something), so much the better.

iethree:
> > we should auto-select the only model to append if there's only 1
> 
> IMO, if there are _any_ to append to, we should just auto-select one of them, since even dumbly selecting the first one alphabetically has a higher probability of being the one the user intended (>0% chance of being right) than the probability that no selection is the correct one (exactly 0%, haha). If it's easy to be slightly smarter (like auto selecting the one with the most recent last_edited_at date or something), so much the better.

ok, now this will auto-select the most recently update model (which will also be the only model if there's just one)

frontend/src/metabase/collections/components/ModelUploadModal.tsx
```diff
@@ -47,6 +49,18 @@ export function ModelUploadModal({
     [models.data],
   );
 
+  useEffect(
+    function setDefaultTableId() {
```
iethree, line 53: I named this function especially for @sloansparger

sloansparger (approved):
lgtm!

PR #38844, opened 2024-02-16
Remove redundant hydration of field values when getting param_values

qnkhuat:


src/metabase/models/params/field_values.clj
```diff
@@ -46,9 +46,7 @@
   [field-ids]
   (when (seq field-ids)
     (not-empty
-     (let [fields       (-> (t2/select :model/Field :id [:in (set field-ids)])
-                            (field/readable-fields-only)
-                            (t2/hydrate :values))
```
qnkhuat, line 51: the values is fetched on line 53, there is no need to hydrate it here.

crisptrutski: I'm not sure we should remove this, even though the behavior is still correct without it. The `get-or-create-full-field-values!` function will use the pre-hydrated :values if possible, rather than doing its own query, so essentially the existing code is doing 2 instead of N + 1 queries.

qnkhuat: oh I didn't realized `get-or-create-full-field-values` skip if `values` exists.

Issue #38699, opened 2024-02-13
[Epic] Appearance settings page polish


WiNloSt:
Target release: 0.50

**Links**
- [Product doc](https://www.notion.so/metabase/Appearance-Settings-page-polish-f9b02b47a0ae427ba8be810dfa924e0d)
- [Tech doc](https://www.notion.so/metabase/Tech-Appearance-Settings-page-polish-8ad4a201747d4e0e880bc090ba1b19a5)

```[tasklist]
# Milestone 1
- [ ] https://github.com/metabase/metabase/pull/38713#top
- [ ] https://github.com/metabase/metabase/pull/38850
```

PR #38850, opened 2024-02-16
Follow up on "Polish appearance settings"

WiNloSt:
Epic: https://github.com/metabase/metabase/issues/38699

[Discussion on Slack.](https://metaboat.slack.com/archives/C063Q3F1HPF/p1708014232222509?thread_ts=1707901949.108579&cid=C063Q3F1HPF)

### Description
These are the changes introduced in this PR:
1. The application colors and chart colors don't change after updating so it will force users to refresh the page before seeing the change taking effect. It would be more obvious than changing some colors and requiring users to still reload the page.
2. Update copy on Application colors, Custom dashboard, Custom Homepage

### How to verify

1. Go to Admin > Appearance > Color palette
3. Change any color and see that no element in the UI has its color changed yet.
4. Refresh the page
5. Elements should have new colors accordingly.

### Checklist


enterprise/frontend/src/metabase-enterprise/whitelabel/index.js
```diff
@@ -237,7 +237,6 @@ if (hasPremiumFeature("whitelabel")) {
 
   PLUGIN_APP_INIT_FUCTIONS.push(() => {
     updateColors();
-    MetabaseSettings.on("application-colors", updateColors);
```
WiNloSt, line 240: This is the part where we change the color without requiring reloading the page. Please see https://www.loom.com/share/c64729d5b60847388f1ba0bd2efcc4c2?sid=3f3a0ac0-78ba-443c-ac5d-cd8f5000a8fd at minute 4:20 for context.

frontend/src/metabase/selectors/settings.ts
```diff
@@ -1,28 +1,31 @@
...
   (state: State) => state.settings,
   settings => settings.loading,
 );
 
-export const getSetting = <T extends SettingKey>(
-  state: State,
+type GetSettings<S extends State> = S["settings"]["values"];
+type GetSettingKey<S extends State> = keyof GetSettings<S>;
+
+export const getSetting = <S extends State, T extends GetSettingKey<S>>(
```
WiNloSt, line 21: Makes this selector generic so that it will allow to use the enterprise setting keys which don't exist in OSS state.

It [was used here.](https://github.com/metabase/metabase/pull/38850/commits/1587411da685857ee586b3096a58148f6a0ce36c#diff-ce1e941a646242ac9341073cc9777328968aced31361ff7fc699502ab8352fabR35) But that change is already reverted. So I only keep the type improvement.

frontend/src/metabase/styled-components/containers/GlobalStyles/GlobalStyles.tsx
```diff
@@ -1,12 +1,38 @@
-import { connect } from "react-redux";
-import type { State } from "metabase-types/store";
+import { css, Global } from "@emotion/react";
```
WiNloSt, line 1: This combines the previous `GlobalStyles` View and this file. So we don't have them in separate places. We can now use a single file because we can use `useSelector`

deniskaber (approved):

albertoperdomo (approved):

vbenedetti (approved):
Looks good! bouncing is resolved and the toast was a great solution to the refresh page problem

PR #38858, opened 2024-02-16
adds /api/setup/token-check and add license-token to POST /api/setup

npretto:
### Description

Part of  [[Epic] Add license activation in the initial setup](https://github.com/metabase/metabase/issues/38867)

We're adding an _optional_ license activation step in the setup, only for people running EE code and without an already set license (so hosted people won't see it).

This PR adds a new endpoint `GET /api/setup/token-check` that checks if a token is valid.
This will be used to validate the input on the frontend. This endpoint won't work after the setup.

It also adds a field in `POST /api/setup` to accept the token and store it.

I tried to code it myself, but let me know if I'm too much off track and if this needs to be picked up by a BE team

### DIFF
--- a/src/metabase/api/setup.clj
+++ b/src/metabase/api/setup.clj
@@ -127,10 +127,10 @@
          {invited_first_name :first_name,
           invited_last_name  :last_name,
           invited_email      :email}                    :invite
-          license-token                                 :license-token
+          license_token                                 :license_token
          {:keys [allow_tracking site_name site_locale]} :prefs} :body, :as request}]
   {token              SetupToken
-   license-token      [:maybe ms/NonBlankString]
+   license_token      [:maybe ms/NonBlankString]
    site_name          ms/NonBlankString
    site_locale        [:maybe ms/ValidLocale]
    first_name         [:maybe ms/NonBlankString]
@@ -160,7 +160,7 @@
                                                        {:email email, :first_name first_name})
                   (setup-set-settings!
                    request
-                   {:email email, :site-name site_name, :site-locale site_locale, :allow-tracking? allow_tracking, :license-token license-token})
+                   {:email email, :site-name site_name, :site-locale site_locale, :allow-tracking? allow_tracking, :license-token license_token})
                   (assoc user-info :database db)))
               (catch Throwable e
                 ;; if the transaction fails, restore the Settings cache from the DB again so any changes made in this
--- a/src/metabase/api/setup.clj
+++ b/src/metabase/api/setup.clj
@@ -102,10 +102,12 @@
                                              (when schedules
                                                (sync.schedules/schedule-map->cron-strings schedules)))))))
 
-(defn- setup-set-settings! [_request {:keys [email site-name site-locale allow-tracking?]}]
+(defn- setup-set-settings! [_request {:keys [email site-name site-locale allow-tracking? license-token]}]
   ;; set a couple preferences
   (public-settings/site-name! site-name)
   (public-settings/admin-email! email)
+  (when (not (premium-features/premium-embedding-token))
+    (premium-features/premium-embedding-token! license-token))
   (when site-locale
     (public-settings/site-locale! site-locale))
   ;; default to `true` if allow_tracking isn't specified. The setting will set itself correctly whether a boolean or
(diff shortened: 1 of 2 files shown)


src/metabase/api/setup.clj
```diff
@@ -125,8 +127,10 @@
          {invited_first_name :first_name,
           invited_last_name  :last_name,
           invited_email      :email}                    :invite
+          license_token                                 :license_token
          {:keys [allow_tracking site_name site_locale]} :prefs} :body, :as request}]
   {token              SetupToken
+   license_token      [:maybe ms/NonBlankString]
```
npretto, line 133: I used `license_token` as we already use `token` for the setup token

test/metabase/api/setup_test.clj
```diff
@@ -615,3 +633,30 @@
...
+                                                             :features ["test" "fixture"]
+                                                             :trial    false})]
+          (is (= {:valid true}
+                 (client/client :get (str "setup/token-check?license_token=" random-fake-token)))))))
+    (testing "Check that returns {valid: false} for invalid token"
+      (mt/with-temporary-setting-values [has-user-setup false]
+        (with-redefs [premium-features/fetch-token-status (fn [_x] {:valid false})]
+          (is (= {:valid false}
+                 (client/client :get (str "setup/token-check?license_token=" random-fake-token))))))
+      (testing "Check that returns {valid: false} for invalid token"
```
WiNloSt, line 659: This test is nested inside the test on line 654. I guess it's not unintended?

WiNloSt (commented):
Overall from FE engineer's perspective this looks fine. But let's get a proper review from BE enginneers

src/metabase/api/setup.clj
```diff
@@ -156,7 +160,7 @@
                                                        {:email email, :first_name first_name})
                   (setup-set-settings!
                    request
-                   {:email email, :site-name site_name, :site-locale site_locale, :allow-tracking? allow_tracking})
+                   {:email email, :site-name site_name, :site-locale site_locale, :allow-tracking? allow_tracking, :license-token license_token})
```
npretto, line 163: I hope to have got the casing of the variables right 🙃

dpsutton:
What is the intended UX if the token is invalid or if the metabase instance has a network error? I think this would throw an error and reject user creation. Also, if there is a token set in the env we probably don't want to offer to save an ee token, nor should we attempt to set it.

dpsutton:
#### Invalid token shape
I've got a setup payload with an invalid token shape (too few characters (less than 64)) and I get this response:

```
❯ http post "localhost:3006/api/setup" -pb < post.json
{
    "cause": "Token format is invalid.",
    "data": {
        "error-details": "Token should be 64 hexadecimal characters.",
        "message": "Token format is invalid.",
        "status-code": 400,
        "toucan2/context-trace": [
            [
                "resolve connection",
                {
                    "toucan2.connection/connectable": "class metabase.db.connection.ApplicationDB"
                }
            ],
            [
                "resolve connection",
                {
                    "toucan2.connection/connectable": "default"
                }
            ],
            [
                "resolve connection",
                {
                    "toucan2.connection/connectable": null
                }
            ]
        ]
    },
    "error-details": "Token should be 64 hexadecimal characters.",
    "message": "Token format is invalid.",
    "trace": [
        [
            "metabase.public_settings.premium_features$fn__46575$fn__46576",
            "invoke",
...
```

#### Invalid token

and when i have an invalid token:

```
❯ http post "localhost:3006/api/setup" -pb < post.json
{
    "cause": "Token does not exist.",
    "data": {
        "error-details": null,
        "message": "Token does not exist.",
        "status-code": 400,
        "toucan2/context-trace": [
            [
                "resolve connection",
                {
                    "toucan2.connection/connectable": "class metabase.db.connection.ApplicationDB"
                }
            ],
            [
                "resolve connection",
                {
                    "toucan2.connection/connectable": "default"
                }
            ],
            [
                "resolve connection",
                {
                    "toucan2.connection/connectable": null
                }
            ]
        ]
    },
    "error-details": null,
    "message": "Token does not exist.",
    "trace": [
        [
            "metabase.public_settings.premium_features$fn__46575$fn__46576",
            "invoke",
            "premium_features.clj",
            231
        ],
```

Will the FE need help understanding that it's the token that is invalid and that's why the admin user creation was rejected?

npretto:
@dpsutton the FE should never (outside of bugs) do that request with an invalid token as we use the `token-check` endpoint to validate it first

The project was split a couple of times so I'll recap the behaviour:
- the new step will only be visible in EE builds if no feature is detected (so if somebody already passed a token via the env, or if the instance is hosted in the cloud it won't show up as it will have at least the `hosting` feature)
- when the FE asks for the token, it will validate it with the `token-check` endpoing, and only if it's valid it will store in redux
- the FE will then send that token to the POST setup endpoint


The only thing that comes to my mind is that the first request (the validation one) could pass and the one done in /setup could fail for network conditions, but if I understood the code the response is cached so it shouldn't big a problem

dpsutton (approved):

PR #38859, opened 2024-02-16
Migrate getPivotColumnSplit to Mlv2

ranquild:


frontend/src/metabase-lib/queries/utils/pivot.js
kamilmielnik, line 1: It's a great opportunity to refactor this file to TS

kamilmielnik (approved):

PR #38873, opened 2024-02-16
Add the ability to copy required checks from `master` to any existing branch

nemanjaglumac:


release/src/required-checks.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Octokit } from "@octokit/rest";
+
+const { GITHUB_TOKEN, GITHUB_OWNER, GITHUB_REPO } = process.env;
+
+const github = new Octokit({ auth: GITHUB_TOKEN });
+
+const config = {
+  owner: GITHUB_OWNER || "",
+  repo: GITHUB_REPO || "",
+};
```
nemanjaglumac, line 11: @iethree we said on Slack to extract `getGithubCredentials` to utils and to then use it in this file, but when I started doing that, I realized the diff will be unnecessarily large. API will throw without owner and the repo anyway, so I think this is just way simpler, and it makes TypeScript happy.

Unless you strongly object, I propose we leave it like this in the first iteration.

iethree (approved):
Works great! Sorry it took me so long to test this. I let my fork get really stale and I had to do some clean up there.

I pushed a commit with a few additions
- changed some variable names to make it a little clearer what was happening
- added some (colored! 🎨) console notifications to help with error/success states
- added a package.json script to make running this easier
- added a section to the readme to explain what it does.

let me know if you disagree with any of those choices!

One question: any ideas for how to remember/automate running this when we cut a new release branch?

nemanjaglumac:
> One question: any ideas for how to remember/automate running this when we cut a new release branch?

This can happen the moment we start automating the new release branch creation flow. If you imagine there is a function that creates a new branch and takes only one argument (branch name), we can use that same argument to copy the required checks from the master.

Ideally, we'd introduce a few more guards before that (e.g. to prevent accidentally deleting stuff)

PR #38874, opened 2024-02-16
Fixes filters against datetime binned expressions

### LINKED ISSUE #33528: Relative date filters don't work with custom date columns
### Describe the bug

When using a relative date filter with a custom date column the query always returns "No results" despite data being present for the selected range.

### To Reproduce

1. New -> Question -> Orders
2. Add custom column `= [Created At]`, with name DateColumn
3. Filter -> DateColumn -> Relative dates -> Current -> Year
4. Visualize

Actual result:
- No results found

Screenshot 2023-08-28 at 14 57 51

Expected result:
- Rows are returned for the current year. Compare with filtering by `Created At` directly.

Screenshot 2023-08-28 at 14 57 57

### Expected behavior

_No response_

### Logs

_No response_

```JSON
0.47.0
```

### Severity

P2

### Additional context

_No response_

snoe:
Fixes #33528

Turns out that `mbql/normalize` and `lib/normalize` behaved slightly different and `[:expression "abc" {...}]` refs would drop their opts in the former path. In order to properly query against binned datetimes it's important that expression ref does not lose its type or else the optimizer will not see that `time-interval` needs to convert to a `between` rather than an `=`.


src/metabase/mbql/normalize.cljc
```diff
@@ -86,12 +86,24 @@
   (cond-> [:aggregation aggregation-index]
     (some? option) (conj option)))
 
+(defn- normalize-ref-opts [opts]
+  (let [opts (normalize-tokens opts :ignore-path)]
+   (cond-> opts
+       (:base-type opts)     (update :base-type keyword)
+       (:temporal-unit opts) (update :temporal-unit keyword)
+       (:binning opts)       (update :binning (fn [binning]
```
bshepherdson, line 94: The RHS is equivalent to `(m/update-existing-in [:binning :strategy] keyword)`.

bshepherdson (approved):

PR #38875, opened 2024-02-16
Sidebar loading state regression fix

sloansparger:
### Description

Needs design approval.

I introduced a regression in #38582 due to misunderstanding loading state for entities could become true whenever that entity's data is being revalidated from somewhere else in the UI. This results in a subpar experience that looks like this (focus on sidebar reloading from csv uploads and bookmarking):

https://github.com/metabase/metabase/assets/7104357/e12a8f39-d59f-42aa-8ea6-5976662f95d3

As part of this PR, when an error occurs I no longer try to render what we can with what valid data we have, but instead just display an error if any error occurs. Ideally in the future we can have separate state for loading / revalidating for entities, but for now this produces a much better user experience:

https://github.com/metabase/metabase/assets/7104357/5a15400e-32f3-4de2-af65-a5599cdaced8

In the case of an errors (say some 500 HTTP error occurs when fetching databases), we will now render this:
failure-to-load-database-endpoint

### How to verify

1. Open browser devtools's network tab
2. Go to home page
3. Search for "database" in network requests
4. Right click the entry and select "Block request URL"
5. Refresh the page, you should see the error UI

### Checklist

### DIFF
--- a/e2e/test/scenarios/navigation/navbar.cy.spec.js
+++ b/e2e/test/scenarios/navigation/navbar.cy.spec.js
@@ -20,6 +20,12 @@ describe("scenarios > navigation > navbar", () => {
       navigationSidebar().should("be.visible");
     });
 
+    it("should display error ui when data fetching fails", () => {
+      cy.intercept("GET", "/api/database", req => req.reply(500));
+      cy.visit("/");
+      navigationSidebar().findByText(/An error occurred/);
+    });
+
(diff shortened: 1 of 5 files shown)


iethree (approved):
❇️

sloansparger:
Sizing of icon and text for the error state have been updated
Screenshot 2024-02-19 at 12 52 56 PM

sloansparger:
designed approved by @naatz-co

PR #38882, opened 2024-02-16
Use all hidden parameters to determine parameter container visibility

### LINKED ISSUE #13934: Embedding with `#titled=false` shows whitespace, when filters are hidden with `hide_parameters=...`
**Describe the bug**
When embedding/sharing with filters, and defining the title to be hidden, then it still shows a whitespace, when all filters are hidden with `hide_parameters`.

**To Reproduce**
1. Create dashboard and add a filter (eg Category), save dashboard, and enable embedding.
2. Now change border, title and parameter to hidden - the [EmbedFrame-header](https://github.com/metabase/metabase/blob/b16e923eca41ca7d75894779ca5849c9ff73e4c2/frontend/src/metabase/public/components/EmbedFrame.jsx#L90-L112) is still showing about 60px tall whitespace.
`/embed/dashboard/<token>#bordered=false&titled=false&hide_parameters=category`
image

**Expected behavior**
Hide the `EmbedFrame-header`, when title is hidden and there are none visible filters.

Tested 0.34.3 thru 0.37.2

**Additional context**
Related to #7778

JesseSDevaney:
Updates `visibleParameters` check to use all hidden parameters: both query string supplied `#hide_parameters=...` and `hiddenParameterSlugs` supplied.

- Closes: https://github.com/metabase/metabase/issues/13934

**Before**
image

**After**
image

### DIFF
--- a/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
+++ b/frontend/src/metabase/public/components/EmbedFrame/EmbedFrame.tsx
@@ -165,7 +165,7 @@ function EmbedFrame({
 
   const hasParameters = Array.isArray(parameters) && parameters.length > 0;
   const visibleParameters = hasParameters
-    ? getVisibleParameters(parameters, hiddenParameterSlugs)
+    ? getVisibleParameters(parameters, hideParameters)
     : [];
   const hasVisibleParameters = visibleParameters.length > 0;
 


WiNloSt (commented):
The change looks good, but I expect the change to also have relevant tests accompanying it.

I see we don't have even unit tests set up for `EmbedFrame` yet, but could you perhaps see if it's viable to add one?

I think the test for this change is something like testing that the parameter container isn't there in the DOM using `not.toBeInTheDocument()`

kulyk (approved):

Issue #37165, opened 2023-12-29
Bug when using custom columns to do filters for the past 1 day/month/year


Jerrylovescoding:
### Metabase Version: 
v0.48.1

### Describe Bug:
When I use custom datetime column to filter last 1 month, I get a wrong result. Viewing the sql, I find the converting sql is wrong.
![image](https://github.com/metabase/metabase/assets/25549512/d0c1537e-2086-46a1-b893-3a7b12583b02)

### To Reproduce

1. Go to query editor
![image](https://github.com/metabase/metabase/assets/25549512/946b8d02-b64e-49a2-a7ec-9739577faa3d)

2. Add a new custom datetime column
![image](https://github.com/metabase/metabase/assets/25549512/d1cf739d-88f6-43a2-946b-53c97370354a)

3. Use custom column as a filter
![image](https://github.com/metabase/metabase/assets/25549512/27e370d0-4bef-4211-bc90-1a39038c124d)

4. View the wrong sql
![image](https://github.com/metabase/metabase/assets/25549512/9083c2b5-4f94-47e2-a454-0c6587f58759)

### Information about your Metabase installation

```JSON
"metabase-info": {
    "databases": [
      "mysql",
      "postgres",
      "mongo",
      "clickhouse"
    ],
    "hosting-env": "docker",
    "application-database": "postgres",
    "application-database-details": {
      "database": {
        "name": "PostgreSQL",
        "version": "13.9"
      },
      "jdbc-driver": {
        "name": "PostgreSQL JDBC Driver",
        "version": "42.6.0"
      }
    }
 }
```

### Severity

We got a wrong result if we use query editor in that way.

darksciencebase:
I suspect this was fixed in https://github.com/metabase/metabase/pull/35523 which we had to revert because other things got broken. The [new approach to fixing it](https://github.com/metabase/metabase/pull/36068) is in review, I'll check if that one fixes this particular case

darksciencebase:
ok, it's not fixed there, we'll have to look into it.

lbrdnk:
Issue was fixed by PRs #38886 and #38874. It is not reproducible with those PRs merged.

PR #38886, opened 2024-02-16
Wrap literals around expression refs

### LINKED ISSUE #27185: Postgres "specific dates" filter errors with datetime custom expression functions
When using a Postgres database, applying a "specific dates" filter with a custom expression function results in an error:

`ERROR: operator does not exist: timestamp with time zone >= character varying Hint: No operator matches the given name and argument types. You might need to add explicit type casts.`

image

Steps to reproduce: 
1. Create a new question, selecting `Orders`
2. Create a custom expression with the formula `datetimeAdd([Created At], 1, "month")`
3. Create a filter on the custom expression column, selecting "specific dates"
image
4. Observe the error.

`ERROR: operator does not exist: timestamp with time zone >= character varying Hint: No operator matches the given name and argument types. You might need to add explicit type casts. Position: 1091`

Can replicate with `now`, `datetimeSubtract`, or `convertTimezone` by changing step 2.

- Your databases: Postgres
- Metabase version: 0.45.1

The problem might exist for other DBMS other than Postgres.

snoe:
Fixes #27185

Expression refs need to be handled like field refs as far as type hints in opts are concerned. Otherwise we don't pass the correct parameter types to the drivers.


snoe:
This requires https://github.com/metabase/metabase/pull/38874 to be merged first

test/metabase/query_processor/middleware/wrap_value_literals_test.clj
```diff
@@ -211,6 +211,13 @@
               {:source-query {:source-table $$checkins
                               :filter       [:> $date "2014-01-01"]}}))))))
 
+(deftest ^:parallel wrap-literals-around-expressions
```
lbrdnk, line 214: Should not test names end with `-test`?

test/metabase/query_processor/middleware/wrap_value_literals_test.clj
```diff
@@ -211,6 +211,13 @@
...
+  (testing "does wrapping literals work against expressions? (#27185)"
+    (wrap-value-literals
```
lbrdnk, line 216: I assume here we are testing that `qp.wrap-value-literals/wrap-value-literals` does not throw. If my assumption is correct, being explicit about that would make the code easier to reason about.

test/metabase/query_processor/middleware/wrap_value_literals_test.clj
```diff
@@ -211,6 +211,13 @@
...
+      (lib.tu.macros/mbql-query checkins
+        {:expressions {"foo" $date}
+         :filter       [:> [:expression "foo" {:base-type :type/DateTime}] "2014-01-01"]}))))
```
lbrdnk, line 219: NIT: whitespace.

lbrdnk (approved):
LGTM. I've tested the changes rebased on `fix-expression-datetime-binned-filters` branch. Apart from fixing the issue in the PR description, it also fixes the #37165. That's something to be confirmed when this PR and #38874 are in master.

metamben (approved):

test/metabase/query_processor/middleware/wrap_value_literals_test.clj
snoe: Thanks, dumb miss on my end, repl testing and forgot to make an assertion.

PR #38891, opened 2024-02-18
simulate time scale with category scale

alxnddr:
## Description

An attempt to simulate continuous scales with a discrete one

- [x] correct spacing between data points
- [x] handle tiny granularities that are way less than 1 pixel
- [x] click behavior should work
- [ ] render only desired ticks on x-axis
- [ ] brushing should work


alxnddr:
Closing in favor of https://github.com/metabase/metabase/pull/38911

PR #38893, opened 2024-02-19
Fallback to the old privilege check method if new method return `false`

qnkhuat:
Context: https://metaboat.slack.com/archives/C052ZBWRG3W/p1708022597711849
and https://metaboat.slack.com/archives/C052ZBWRG3W/p1708316467663349

In #37439, we introduced a fast way to check table privilege by using the result from the `driver/current-user-table-privileges` method.

But we see some users reported tables are disabled during sync, while I haven't found out the root cause of that, I think we should added a fallback to using the old privilege check if the new method return `false`. This way, we still get the performance boost, but it is safer when introducing this new method.

The only downside is it's hard to know why the new method gives false positive results. With this, we rely on users to see our warn log and open an issue, or maybe we'll collect some telemetry using a snowplow.

### DIFF
--- a/src/metabase/driver/sql_jdbc/sync/describe_database.clj
+++ b/src/metabase/driver/sql_jdbc/sync/describe_database.clj
@@ -12,6 +12,7 @@
    [metabase.lib.metadata :as lib.metadata]
    [metabase.models.interface :as mi]
    [metabase.query-processor.store :as qp.store]
+   [metabase.util :as u]
    [metabase.util.honey-sql-2 :as h2x]
    [metabase.util.log :as log]
    [metabase.util.malli :as mu]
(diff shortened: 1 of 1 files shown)


src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -135,13 +136,27 @@
...
+            [driver ttype])
+           (check-by-querying-fn)
+
+           ;; fast checking with table privileges is an experimental feature
+           ;; so in case it gives false result we'll double check using the old method to make
+           ;; sure we does not have false positives.
+           (not (contains? schema+table-with-select-privileges [schema table]))
+           (u/prog1 (check-by-querying-fn)
+             (when <>
+               (log/warn (str "Fast privilege check using :table-privileges gives false positive result. "
```
qnkhuat, line 154: I'll create an issue so users can just go there and comment about the table info.

calherries (commented):
We should have a test for the new code branch. As far as I know we don't have any existing ones

qnkhuat:
customer limit access to their tables on purpose, it's not because of our sync process ([thread](https://metaboat.slack.com/archives/C052ZBWRG3W/p1708522745138549?thread_ts=1708316467.663349&cid=C052ZBWRG3W)). closing

PR #38900, opened 2024-02-19
Disable immer auto freeze

uladzimirdev:
#### TL;DR

MLv2 modifies values which are passed as arguments and adds a new property for caching objects to speed up calculations. RTK uses immer under the hood and freezes state automatically, which leads to the error on cljs side.

#### Longer explanation

[PR](https://github.com/metabase/metabase/pull/38168) highlighted the problem of MLv2 which can't handle frozen objects which we get after RTK's state modifications. immer has auto-freeze option ([enabled by default](https://github.com/immerjs/immer/blob/f6736a4beef727c6e5b41c312ce1b202ad3afb23/src/core/immerClass.ts#L152-L159)) which also freezes the state to prevent implicit state modifications. 

[Even if it's a recommended approach](https://github.com/reduxjs/redux-toolkit/discussions/1189#discussioncomment-884343) which can save developers from shooting a foot, we follow the same approach across application and avoid implicit state modifications. 
You can't implicitly modify state inside reducers with RTK (because it uses immer under the hood and generates a new state even if you modify the old), but it doesn't save you from modifying state inside components by reference.

But I'd recommend disabling auto-freeze because freezing can result in such unexpected errors which really hard to debug when MLv2 accepts a frozen or partially frozen object.

As an example, here is what I got after passing a frozen object as a parameter
image

and I was able to reproduce it locally only in fast 3g simulation

### How to verify

tests should pass

### Checklist

### DIFF
--- a/frontend/src/metabase/store.js
+++ b/frontend/src/metabase/store.js
@@ -1,9 +1,18 @@
 import { configureStore, combineReducers } from "@reduxjs/toolkit";
+import { setAutoFreeze } from "immer";
 import { routerReducer as routing, routerMiddleware } from "react-router-redux";
 import promise from "redux-promise";
 import { PLUGIN_REDUX_MIDDLEWARES } from "metabase/plugins";
 
-export function getStore(reducers, history, intialState) {
+/**
+ * MLv2 modifies passed values and adds a new property for caching objects
+ * to speed up calculations. RTK uses immer under the hood and freezes
+ * state automatically, which leads to the error on cljs side.
+ * As a workaround, we disable immer's auto-freeze
+ */
+setAutoFreeze(false);
+
+export function getStore(reducers, history, initialState) {
   const reducer = combineReducers({
     ...reducers,
     routing,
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -12518,12 +12518,7 @@ image-size@^1.0.0:
   dependencies:
     queue "6.0.2"
 
-immer@^9.0.16:
```
uladzimirdev, line 12521: deduplication. RTK's immer was not the same as we get from `import {produce} from 'immer'`.

frontend/src/metabase/store.js
```diff
@@ -1,17 +1,26 @@
...
+export function getStore(reducers, history, initialState) {
   const reducer = combineReducers({
     ...reducers,
     routing,
   });
 
   return configureStore({
     reducer,
-    preloadedState: intialState,
+    preloadedState: initialState,
```
uladzimirdev, line 23:
```diff
+ initialState
- intialState
```

ranquild (approved):

uladzimirdev:
I'm not brave enough to merge this as it is, as a workaround please create a deep copy of the object where nested collections can be frozen, e.g. using [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)

PR #38903, opened 2024-02-19
Remove MLv1 fields, breakouts, expressions

ranquild:
These are not used in the product anymore


nemanjaglumac (approved):
I can't realistically verify all changes, but the CI is green and that should be good enough test that this all works.

Just out of curiosity, what was the first thread you started pulling that led to this whole thing unwrapping so beautifully?

PR #38906, opened 2024-02-19
Native query tag editor panel: use Mantine

oleggromov:
This is a partial refactoring of the variable editor panel in the native query builder.
Squared with **red components** are not changed. 

Screenshot 2024-02-19 at 22 17 19

The reason for that is that the **field mapping component** is fairly complex and is easier to be changed in a separate PR (coming soon), and the **parameter value widget** is a whole different story, which definitely needs to be address separately.


frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
```diff
@@ -23,79 +21,92 @@ export interface ValuesSourceSettingsProps {
...
     <>
-      <Radio
-        value={queryType}
-        options={radioOptions}
-        vertical
-        onChange={onChangeQueryType}
-      />
+      <Radio.Group value={queryType} onChange={onChangeQueryType}>
+        <Stack spacing="xs">
+          <RadioLine
```
kamilmielnik, line 41: What does the `Line` suffix mean? Did you want to communicate it's a [block](https://developer.mozilla.org/en-US/docs/Glossary/Block-level_content)?

How about communicating the purpose instead of including presentation details in the name?

```suggestion
          <QueryTypeRadio
```

frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
```diff
@@ -23,79 +21,92 @@ export interface ValuesSourceSettingsProps {
...
   onEditClick,
-  isEditDisabled = false,
-}: RadioLabelProps): JSX.Element => {
+}: {
+  selectedValue: ValuesQueryType;
+  ownValue: ValuesQueryType;
+  label: string;
+  disableEdit?: boolean;
+  hideEdit?: boolean;
+  onEditClick?: () => void;
```
kamilmielnik, line 93: 1. We don't need 2 props for this - lack of `onEditClick` could imply `hideEdit`
2. This component could have been much simpler (we could turn 3 props `disabledEdit` + `hideEdit` + `onEditClick` into 1: `children`) if it used composition instead, i.e.

Without editing capabilities:
```tsx
<RadioLine ownValue="none" selectedValue={queryType} label={t`Input box`} />
```

With editing capabilities:
```tsx
<RadioLine ownValue="search" selectedValue={queryType} label={t`Search box`}>
  {queryType === "search" && (
    <QueryTypeEditButton disabled={disableEdit} onClick={openModal} />
  )}
</RadioLine>
```

frontend/src/metabase/ui/components/inputs/TextInputBlurChange/index.tsx
kamilmielnik, line 1: This component should be defined in `TextInputBlurChange.tsx` and exported via `index.tsx`

frontend/src/metabase/ui/components/inputs/TextInputBlurChange/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  onBlurChange: (event: { target: HTMLInputElement }) => void;
+  normalize?: (value: Value) => Value;
+}
+
+/**
+ * A wrapper around TextInput to be used with onBlurChange prop.
+ *
+ * In case you don't need it, use TextInput directly.
+ */
+export function TextInputBlurChange({
```
kamilmielnik, line 22: There's 1 usage of `InputBlurChange` left, can we replace it with `TextInputBlurChange` now?

frontend/src/metabase/ui/components/inputs/TextInputBlurChange/index.tsx
```diff
@@ -0,0 +1,78 @@
+import { TextInput } from "@mantine/core";
```
kamilmielnik, line 1: `InputBlurChange` has a unit test suite - let's copy it for `TextInputBlurChange`, it should work the same.

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -227,148 +224,43 @@ class TagEditorParamInner extends Component<Props> {
     const hasSelectedDimensionField =
       isDimension && Array.isArray(tag.dimension);
     const hasWidgetOptions = widgetOptions?.length > 0;
```
kamilmielnik, line 226:
```suggestion
    const hasWidgetOptions = widgetOptions.length > 0;
```

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
```diff
@@ -1,15 +1,12 @@
-import { Component } from "react";
+import { Component, useMemo } from "react";
```
kamilmielnik, line 1: The labels for `Variable type` and `Filter widget type` inputs are at a different distance from the input than `Variable name`, `Field to map to` and `Filter widget label` inputs (notice the white area between green area and the input in the screenshot):

image

frontend/src/metabase/parameters/components/ValuesSourceSettings/ValuesSourceSettings.tsx
oleggromov: I agree it was a bit confusing. It meant "form line", but it's not the best name. Will rename.

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
oleggromov: This is a good catch, thanks!

frontend/src/metabase/query_builder/components/template_tags/TagEditorParam.tsx
oleggromov: Ok, agreed, but this code I haven't even touched 😅

frontend/src/metabase/ui/components/inputs/TextInputBlurChange/index.tsx
oleggromov: Done.

frontend/src/metabase/query_builder/components/template_tags/TagEditorParamParts/TagEditorParam.styled.tsx
```diff
@@ -17,12 +17,13 @@ export const TagName = styled.h3`
...
 interface ContainerLabelProps {
   paddingTop?: boolean;
+  smallPaddingBottom?: boolean;
 }
 
 export const ContainerLabel = styled.div<ContainerLabelProps>`
   display: block;
   font-weight: 700;
-  padding-bottom: 0.5rem;
+  padding-bottom: ${props => (props.smallPaddingBottom ? "0.25rem" : "0.5rem")};
```
kamilmielnik, line 26: We don't need `smallPaddingBottom` at all if we use `margin-bottom` instead of `padding-bottom`:

```suggestion
  margin-bottom: 0.5rem;
```

kamilmielnik (approved):
Looks good :+1: 

It'd be great if we could get rid of `any` in `TagEditorParam.prototype.setParameterAttribute` and `onChangeDefaultValue` prop in `DefaultRequiredValueControl`, but I guess we'd need to make `Parameter` `any`-less first, which would be a large task on its own. Hopefully some day :crossed_fingers:

frontend/src/metabase/ui/components/inputs/TextInputBlurChange/index.tsx
kamilmielnik: Got it, thanks for checking! :pray:

frontend/src/metabase/query_builder/components/template_tags/TagEditorParamParts/TagEditorParam.styled.tsx
oleggromov: Yeah, right! I'll address in the next PR to avoid waiting for the tests.

oleggromov: Addressed here: https://github.com/metabase/metabase/pull/38997/commits/79f11286dd88a3ad04378bc9a5e73365aa12f3cb

PR #38907, opened 2024-02-19
Filter picker search free input

### LINKED ISSUE #37906: "Is" filter in the header of the columns in a table context should be linked and allow free input
Let's imagine you're browsing a table in a massive DW with a few hundred million of rows. You know exactly the value of some columns you want to filter by, but not others.

If you first select in the column that you know of, typing the value with a filter of type "is" will make you wait, as Metabase wants you to wait for the value and then click on it.

After you made the first filter and now the values are way less, you pick other columns to filter by, and you want to try a few values, but again, Metabase will make you wait till the values come back from the DW for picking them. Pressing enter won't let you enter any value, as on filters of type "is" you can't do that

Filters of type "is" on search boxes should allow you to enter freely any value, and also when you select other filters in a table view, the filters should be linked to the actual values you're seeing.

Ergonomics

**Additional context**
Example:
1) go to browse data-> hosting insights in our internal metabase instance, search for embedding-demo on the instance-dns (you won't be able to select it till it comes back from the DB)
image
2) Click on add filter and wait till results come back
3) Now go to schema and search for "public"
image

see that you won't be able to find easily the schema you picked, as all schemas come back (this filter should come from the subset of data you have in the table)

ranquild:
Product doc https://www.notion.so/metabase/Free-up-text-input-performance-from-loading-values-in-filters-f70287f73d014794abb60653c6d28e39
Fixes https://github.com/metabase/metabase/issues/37906

Previously it wasn't possible to select a value not from the list for "searchable" columns. It becomes a problem when it takes a lot of time to load search results as they are **loaded from the customer DB directly**.  In this case the user is not able to create a filter at all. 

Now we allow entering any value, even without waiting for search results (but no duplicates). This is based on `creatable` property of the mantine's `MultiSelect`. I've decided to go with `New {input}` label but this is to be verified with Product.

How to verify:
- New -> Question -> People
- Filter by Email
- Trying adding filters

**Before**

With search results:
Screenshot 2024-02-19 at 14 46 04

No search results:
Screenshot 2024-02-19 at 14 46 10

**After**

With search results:
Screenshot 2024-02-19 at 14 39 28

No search results:
Screenshot 2024-02-19 at 14 39 09

### DIFF
--- a/frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
+++ b/frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
@@ -29,7 +29,7 @@ interface FilterValuePickerProps<T> {
 
 interface FilterValuePickerOwnProps extends FilterValuePickerProps<string> {
   placeholder: string;
-  shouldCreate: (query: string) => boolean;
+  canAddValue: (query: string) => boolean;
 }
 
 function FilterValuePicker({
@@ -40,7 +40,7 @@ function FilterValuePicker({
   placeholder,
   autoFocus = false,
   compact = false,
-  shouldCreate,
+  canAddValue,
   onChange,
   onFocus,
   onBlur,
@@ -86,7 +86,7 @@ function FilterValuePicker({
         fieldValues={fieldData?.values ?? []}
         selectedValues={selectedValues}
         placeholder={t`Search by ${columnInfo.displayName}`}
-        shouldCreate={shouldCreate}
+        canAddValue={canAddValue}
         autoFocus={autoFocus}
         onChange={onChange}
       />
@@ -97,7 +97,7 @@ function FilterValuePicker({
     <StaticValuePicker
       selectedValues={selectedValues}
       placeholder={placeholder}
-      shouldCreate={shouldCreate}
+      canAddValue={canAddValue}
       autoFocus={autoFocus}
       onChange={onChange}
       onFocus={onFocus}
@@ -111,7 +111,7 @@ export function StringFilterValuePicker({
   values,
   ...props
 }: FilterValuePickerProps<string>) {
-  const shouldCreate = (query: string) => {
+  const canAddValue = (query: string) => {
     return query.trim().length > 0 && !values.includes(query);
   };
 
@@ -121,7 +121,7 @@ export function StringFilterValuePicker({
       column={column}
       values={values}
       placeholder={isKeyColumn(column) ? t`Enter an ID` : t`Enter some text`}
-      shouldCreate={shouldCreate}
+      canAddValue={canAddValue}
     />
   );
 }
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.tsx
```diff
@@ -86,7 +86,7 @@ function FilterValuePicker({
         fieldValues={fieldData?.values ?? []}
         selectedValues={selectedValues}
         placeholder={t`Search by ${columnInfo.displayName}`}
-        nothingFound={t`No matching ${columnInfo.displayName} found.`}
```
ranquild, line 89: Now it's impossible to get this state

frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.unit.spec.tsx
```diff
@@ -368,6 +368,45 @@ describe("StringFilterValuePicker", () => {
...
+        searchValues: {
+          "a@b.com": createMockFieldValues({
+            field_id: PEOPLE.EMAIL,
+            values: [["testa@b.com"]],
+          }),
+        },
+      });
+
+      userEvent.type(screen.getByPlaceholderText("Search by Email"), "a@b.com");
+      userEvent.hover(screen.getByText("New a@b.com"));
```
ranquild, line 387: For some reason mantine doesn't work without hovering the item first :(

kamilmielnik: It'd be helpful if this comment lived in the code

frontend/src/metabase/querying/components/FilterValuePicker/SearchValuePicker/SearchValuePicker.tsx
```diff
@@ -63,12 +67,15 @@ export function SearchValuePicker({
       value={selectedValues}
       searchValue={searchValue}
       placeholder={placeholder}
-      nothingFound={isSearched ? nothingFound : null}
+      shouldCreate={shouldCreate}
+      getCreateLabel={searchValue => t`New ${searchValue}`}
```
kamilmielnik, line 71: cc: @cdeweyx @mazameli 

1. "New" suggests that something will be created. I'd skip the word and just use the value itself.
2. It's not ideal when this value is last in the options list because it can't be seen when there are other results:
image

kamilmielnik (approved):
LGTM :+1:

frontend/src/metabase/querying/components/FilterValuePicker/SearchValuePicker/SearchValuePicker.tsx
kamilmielnik: 3. Ideally the "new" value should not be displayed when an exact match is found:

image

cdeweyx (approved):
Looking good

Issue #38843, opened 2024-02-16
[dc.js migration] incorrect formatting on x-axis ticks using native query date columns


alxnddr:
Affects all cartesian charts:
<img width="1363" alt="Screenshot 2024-02-15 at 8 04 55 PM" src="https://github.com/metabase/metabase/assets/14301985/749daece-55cf-44cf-b2d5-a7d0b39aa7cd">

Issue #38841, opened 2024-02-16
[dc.js migration] time series charts with week bucketing show shifted x-axis ticks


alxnddr:
When data has weekly buckets x-axis tick start from the previous day of the first item:
<img width="1360" alt="Screenshot 2024-02-15 at 9 44 55 PM" src="https://github.com/metabase/metabase/assets/14301985/c624da7b-d5df-4d8a-a1a9-60df443cd235">

Even setting `xAxis.min: "dataMin"` does not fix the issue.

PR #38911, opened 2024-02-19
Improve time scale ticks behavior

alxnddr:
Closes 
- https://github.com/metabase/metabase/issues/38843
- https://github.com/metabase/metabase/issues/38841

## Description

The default ECharts tick selection behavior was confusing on weekly and quarterly data because the ticks did not match the actual data:
Screenshot 2024-02-16 at 12 41 51 AM (3)

This PR fixes these cases and overall improves time scale ticks behavior.

### Handle weeks and quarters
For these units we force ECharts to render ticks with smaller granularity and then in the formatter function we select desired ticks only.
Weeks: we select daily ticks that have the same day of week as the underlying data.
Quarters: we select monthly ticks that are start of quarters.

### Handle long durations
To render long durations relatively to the unit we use the largest possible unit for ticks that exists 3 times within the data which guarantees at least two ticks on the chart.
For example, given the dataset has weekly data but the range of the dataset is 4 years. If we render weekly ticks, they will have some random values that will match start of some weeks of the data. It will be harder to comprehend than just evenly spaced yearly ticks.

The example below demonstrates why it is needed. It has weekly data over a few years.

**Without switching to a larger unit for ticks** 
Screenshot 2024-02-22 at 7 03 46 PM

**After switching to a larger unit for ticks** 
Screenshot 2024-02-22 at 7 02 23 PM


cdeweyx (approved):

EmmadUsmani (changes requested):
Timeseries waterfall charts are no longer working in app (but they do in static viz), for example http://localhost:3000/question/133-waterfall-timeseries-uk-car-accidents

Screenshot 2024-02-23 at 8 57 40 AM

Screenshot 2024-02-23 at 8 58 34 AM

alxnddr:
@EmmadUsmani can you please share more details about the waterfall question that is not working? I don't use this database due to huge delays which makes it unusable for me locally

alxnddr:
@EmmadUsmani I recreated the same question and could not reproduce the problem. Can you please double check it exists?
Screenshot 2024-02-26 at 10 55 04 PM
Screenshot 2024-02-26 at 10 55 10 PM

EmmadUsmani:
In a lot of cases just the first tick is missing, which looks wrong from a user perspective. For example,

Screenshot 2024-02-28 at 9 17 24 AM

Screenshot 2024-02-28 at 9 17 35 AM

Screenshot 2024-02-28 at 9 19 04 AM

Also, I'm not really sure what's going on in this case. We went from displaying Feb 16th twice to Feb 27th twice. In either case it looks wrong, the same date shouldn't be displayed twice for different dimensions.

Screenshot 2024-02-28 at 9 18 31 AM

EmmadUsmani (changes requested):
Let's resolve the above issue regarding the first date, I think it's a big regression.

alxnddr:
@EmmadUsmani there is no actual regression in storybook or the app but Loki for some reason flakes and drop them, I'll investigate:
Screenshot 2024-02-28 at 2 53 43 PM
Screenshot 2024-02-28 at 2 54 01 PM

> Also, I'm not really sure what's going on in this case. We went from displaying Feb 16th twice to Feb 27th twice. Is Feb 27th the correct date?

I think this spec has relative date time x-axis but due to formatting issues it includes current date which is wrong. I'll check & fix, maybe in a separate PR since this is related to actual time scale x-axis

frontend/src/metabase/visualizations/lib/timeseries.js
```diff
@@ -40,29 +40,29 @@ export function dimensionIsExplicitTimeseries({ cols }, i) {
...
-  { interval: "week", count: 1, testFn: d => d.day() }, // (14) 1 week
-  { interval: "month", count: 1, testFn: d => d.date() }, // (15) 1 month
-  { interval: "month", count: 3, testFn: d => d.month() % 3 }, // (16) 3 months / 1 quarter
-  { interval: "year", count: 1, testFn: d => d.month() }, // (17) 1 year
-  { interval: "year", count: 5, testFn: d => d.year() % 5 }, // (18) 5 year
-  { interval: "year", count: 10, testFn: d => d.year() % 10 }, // (19) 10 year
-  { interval: "year", count: 50, testFn: d => d.year() % 50 }, // (20) 50 year
-  { interval: "year", count: 100, testFn: d => d.year() % 100 }, // (21) 100 year
+export const TIMESERIES_INTERVALS = [
+  { unit: "ms", count: 1, testFn: d => 0 }, //  (0) millisecond
```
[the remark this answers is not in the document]
EmmadUsmani: Let's move this to a `.ts` file in the `echarts` directory

frontend/src/metabase/visualizations/echarts/cartesian/model/axis.ts
```diff
@@ -484,6 +492,160 @@ export const getXAxisEChartsType = (
...
+
+  return isWidthinDataRange;
+};
+
+const findRangeBasedTicksInterval = (
+  range: DateRange,
+): TimeSeriesInterval | null => {
+  // For ticks we want to pick the largest interval that exist 3 times in the range.
+  // For example, if data has week granularity but the range is more than 3 months, we want to show monthly ticks.
+  const ticksIntervalThreshold = 3;
```
EmmadUsmani, line 519: can be a constant outside of the function like `const TICKS_INTERVAL_THRESHOLD = 3`

frontend/src/metabase/visualizations/echarts/cartesian/model/axis.ts
```diff
@@ -484,6 +492,160 @@ export const getXAxisEChartsType = (
...
+  const [minDate, maxDate] = range;
+
+  const isWidthinDataRange = (tickDateRaw: string) => {
+    const tickDate = dayjs(tickDateRaw);
+
+    if (minDate.isSame(maxDate)) {
+      return tickDate.isSame(minDate, "day");
+    }
+
+    const isBeforeRange = minDate && tickDate.isBefore(minDate);
```
EmmadUsmani, line 505: The `minDate &&` is unnecessary since the `DateRange` type is `[Dayjs, Dayjs]`, so it's guaranteed to be defined.

If the type is wrong and it can be null, we should update the type to reflect that, and add the null check earlier otherwise line 501 will error

frontend/src/metabase/visualizations/echarts/cartesian/option/series.ts
```diff
@@ -121,28 +144,32 @@ export const buildEChartsLabelOptions = (
 
 const buildEChartsBarSeries = (
   dataset: ChartDataset,
+  xAxisModel: XAxisModel,
   seriesModel: SeriesModel,
   settings: ComputedVisualizationSettings,
   yAxisIndex: number,
   barSeriesCount: number,
   hasMultipleSeries: boolean,
+  chartMeasurements: ChartMeasurements,
```
EmmadUsmani, line 153: This is unused

frontend/src/metabase/visualizations/echarts/cartesian/waterfall/model/axis.ts
```diff
@@ -28,53 +44,63 @@ export const getWaterfallXAxisModel = (
...
-      xAxisModel.timeSeriesInterval?.interval,
-    );
+    totalXValue = timeSeriesTotalXValue;
+    tickRenderPredicate = (tickValueRaw: string) => {
+      const tickValue = dayjs(tickValueRaw);
+      let tickUnit: CartesianChartDateTimeAbsoluteUnit | undefined;
 
+      // HACK: Due to ECharts default tick selection for weekly and quarterly data
+      // we render for each day and month respectively to select the desired ones only.
+      // This logic of unit selection should be in sync with `getTimeSeriesXAxisModel` in
```
EmmadUsmani, line 62: Can we extract it to a common function?

EmmadUsmani:
Screenshot 2024-02-28 at 10 43 38 AM


I'm seeing this bug on waterfall charts with the total disabled. It only happens in this case, not all waterfall charts without totals.

alxnddr:
@EmmadUsmani the issue with missing ticks is caused by timezones. Since this PR has grown a lot, I moved the timezone issue in a separate task https://github.com/metabase/metabase/issues/39322

alxnddr:
> I'm seeing this bug on waterfall charts with the total disabled. It only happens in this case, not on all waterfall charts without totals.

Fixed this, thanks

EmmadUsmani (approved):

PR #38927, opened 2024-02-19
Support JWT and other SSO methods at the same time

### LINKED ISSUE #18554: Support SAML and JWT logins at the same time
Currently when both SAML and JWT logins are enabled, then only a single "Sign in" button is shown on the login page, which is for the SAML IdP.
Even when sending a JWT payload directly to `/auth/sso`, then still redirects to SAML IdP.
image

Allow separate buttons. Or at least allow JWT payload to not redirect to SAML.
Example, "Google Sign-In" is shown as a separate button:
image

Depending on SAML IdP, integrating JWT into that, so everyone sign in via SAML.
This might not be an option for all, and might require a lot of work depending on the IdP.

### LINKED ISSUE #29492: [Epic] Support JWT and other SSO methods at the same time
Product doc: [Enable multiple SSO methods in the same instance](https://www.notion.so/metabase/Enable-multiple-SSO-methods-in-the-same-instance-7e3d1917b1104b1085ad8f87329fce63?pvs=4)

image

qwef:
Closes https://github.com/metabase/metabase/issues/18554
Closes https://github.com/metabase/metabase/issues/29492
Product doc: https://www.notion.so/metabase/Support-JWT-and-other-SSO-methods-at-the-same-time-7e3d1917b1104b1085ad8f87329fce63

### Description

Adds a if statement for if both JWT and SAML is enabled. 

Personally I think in the future we should split up SAML and JWT into separate endpoints and have two different buttons in the FE. 
- Other SSO methods are separate (Google, LDAP)
- Clears up the behavior of login instead of hiding the SSO method behind the query params
- Cleans up the BE code to have one layer less of indirection

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/sso/api/interface.clj
+++ b/enterprise/backend/src/metabase_enterprise/sso/api/interface.clj
@@ -3,11 +3,18 @@
    [metabase-enterprise.sso.integrations.sso-settings :as sso-settings]
    [metabase.util.i18n :refer [tru]]))
 
+(defn- jwt-saml-both-enabled
+  [req]
+  (if (contains? (:params req) :jwt)
+    :jwt
+    :saml))
+
 (defn- sso-backend
   "Function that powers the defmulti in figuring out which SSO backend to use. It might be that we need to have more
   complex logic around this, but now it's just a simple priority. If SAML is configured use that otherwise JWT"
-  [_]
+  [req]
   (cond
+    (and (sso-settings/saml-enabled) (sso-settings/jwt-enabled)) (jwt-saml-both-enabled req)
     (sso-settings/saml-enabled) :saml
     (sso-settings/jwt-enabled)  :jwt
     :else                       nil))
--- a/enterprise/backend/src/metabase_enterprise/sso/api/interface.clj
+++ b/enterprise/backend/src/metabase_enterprise/sso/api/interface.clj
@@ -3,7 +3,7 @@
    [metabase-enterprise.sso.integrations.sso-settings :as sso-settings]
    [metabase.util.i18n :refer [tru]]))
 
-(defn- jwt-saml-both-enabled
+(defn- select-sso-backend
   [req]
   (if (contains? (:params req) :jwt)
     :jwt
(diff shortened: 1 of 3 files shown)


enterprise/backend/src/metabase_enterprise/sso/api/interface.clj
```diff
@@ -3,11 +3,18 @@
    [metabase-enterprise.sso.integrations.sso-settings :as sso-settings]
    [metabase.util.i18n :refer [tru]]))
 
+(defn- jwt-saml-both-enabled
```
noahmoss, line 6: nit: this function name sounds to me like it returns a boolean; maybe just call it something like `select-sso-backend`?

you could keep `jwt-saml-both-enabled` and have it just abstract `(and (sso-settings/saml-enabled) (sso-settings/jwt-enabled))` maybe

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
```diff
@@ -120,6 +120,33 @@
...
+                   (get-in response [:headers "Location"]))))
+          (testing "login attributes"
+            (is (= {"extra" "keypairs", "are" "also present"}
+                   (t2/select-one-fn :login_attributes User :email "rasta@metabase.com"))))))
+
+      (testing "with SAML and JWT configured, a GET request without JWT params should redirect to SAML IdP"
+        (let [response (saml-test/client-full-response :get 302 "/auth/sso"
+                                                       {:request-options {:redirect-strategy :none}}
+                                                       :return_to default-redirect-uri)]
+          (is (not (saml-test/successful-login? response))))))))
```
noahmoss, line 148: Why wouldn't this be a successful SAML login if both JWT & SAML are set up? Is it just because of the `request-options` being wrong?

Could the test assert that a SAML login can succeed when both are enabled in the `:jwt` key is omitted?

enterprise/backend/src/metabase_enterprise/sso/api/interface.clj
qwef: Yea makes sense, I was thinking maybe sso didn't make sense since we are selecting only across SAML and JWT, but you're right the convention here is like sso -> saml,jwt.

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
qwef: It would redirect to the SAML login page in order to get the correct redirect params.

noahmoss (approved):

PR #38929, opened 2024-02-19
Use report timezone setting macro with cleanup

### LINKED ISSUE #38219: BE flake: metabase.query-processor-test.date-time-zone-functions-test/now-test
Failed run: https://github.com/metabase/metabase/actions/runs/7672181589/job/20912138477?pr=38206#step:3:1358

lbrdnk:
Closes #38219

### Description

Make use of `with-report-timezone-id` which does pre and post cleanup. I believe that test was flaking because session was reused with report timezone set. This change ensures that sessions are restarted after modifications to report timezone.


metamben (approved):
👍

Issue #37420, opened 2024-01-08
[dc.js migration] timeseries x-axis label is cut off


EmmadUsmani:
http://localhost:3000/question/85-line-aapl-stock

The last date label is cut off. This only happens when the width of the chart is too small, both on the dashboard and in the query builder.

![image](https://github.com/metabase/metabase/assets/37751258/99150edd-54e4-4b0d-9c01-f95aea821dee)

<img width="614" alt="Screenshot 2024-01-08 at 3 52 53 PM" src="https://github.com/metabase/metabase/assets/37751258/a302031a-3f77-4bf7-b0ac-f3e2fa974a67">

http://localhost:6006/?path=/story/static-viz-combochart--trend-combo

Same issue but here the first date label is cut off.

<img width="566" alt="Screenshot 2024-02-08 at 9 45 42 AM" src="https://github.com/metabase/metabase/assets/37751258/6291739f-f3f6-4950-b5c3-3efa271919b6">

EmmadUsmani:
@alxnddr This is probably caused by the `grid` option, as removing it fixes the cut off

https://github.com/metabase/metabase/assets/37751258/68c55aee-9182-44e9-94af-22cd8e928fbe

alxnddr:
> @alxnddr This is probably caused by the `grid` option, as removing it fixes the cut off

Default grid paddings are too huge so it looks ok in this specific case but it is still broken. We need to compute grid based on the ticks dimensions including adding some horizontal space for long x-axis ticks

EmmadUsmani:
> We need to compute grid based on the ticks dimensions including adding some horizontal space for long x-axis ticks

Yup I'm currently working on a solution that does this

alxnddr:
@EmmadUsmani the old `dc.js` implementation actually adds paddings and I think this route is simpler:
<img width="423" alt="Screenshot 2024-02-13 at 4 53 46 PM" src="https://github.com/metabase/metabase/assets/14301985/623f519f-d52f-468c-85fd-3c90de35f327">
The logic of side padding calculation should be like `sidePadding = Math.max(yTicksWidth + AxisName, xTickWidth / 2)`

Issue #37275, opened 2024-01-04
[dc.js migration] image is slightly cut off


EmmadUsmani:
http://localhost:3000/question/197-waterfall-data-labels-mixed

<img width="590" alt="Screenshot 2024-01-04 at 11 02 02 AM" src="https://github.com/metabase/metabase/assets/37751258/04f635b8-7119-4ae5-baa0-eec9cd81e31c">

http://localhost:6006/?path=/story/static-viz-combochart--trend-combo

<img width="566" alt="Screenshot 2024-02-08 at 9 45 42 AM" src="https://github.com/metabase/metabase/assets/37751258/6291739f-f3f6-4950-b5c3-3efa271919b6">

http://localhost:3000/question/236-scatter-multiseries-breakout-bubble-size-pokedex-def-vs-sp-def-breakout-by-type-size-by-hp

<img width="1904" alt="Screenshot 2024-02-09 at 10 31 49 AM" src="https://github.com/metabase/metabase/assets/37751258/b744b1ec-31c7-4c32-b14a-d699a46dbd37">

Related to https://github.com/metabase/metabase/issues/37420 and possibly has the same root cause

EmmadUsmani:
Right now the top data label being cut off is fixed, but the left y-axis label is still cut off. It only happens on the waterfall chart, so I'll pause work on this for now since Sasha is refactoring the waterfall chart.

EmmadUsmani:
Repoening since it's still just a little cut off in 

question/13529
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194580878449

Issue #38783, opened 2024-02-14
[dc.js migration] auto-selected `graph.x_axis.axis_enabled` setting does not match old implementation


EmmadUsmani:
Similar to https://github.com/metabase/metabase/issues/38782, this was also something that came up often in [#echarts-bug-bash](https://metaboat.slack.com/archives/C06G94JTWBS), below are some examples.

https://metaboat.slack.com/archives/C06G94JTWBS/p1706193475366799

https://metaboat.slack.com/archives/C06G94JTWBS/p1706193183854899

https://metaboat.slack.com/archives/C06G94JTWBS/p1706193082481509

Below is the logic that the old implementation used

```javascript
function beforeRenderComputeXAxisLabelType(chart) {
  // treat graph.x_axis.axis_enabled === true as "auto"
  if (chart.settings["graph.x_axis.axis_enabled"] === true) {
    const overlaps = checkXAxisLabelOverlap(chart);
    if (overlaps) {
      if (chart.isOrdinal()) {
        const spacing = computeXAxisSpacing(chart);
        if (spacing < X_LABEL_HIDE_THRESHOLD) {
          chart.settings["graph.x_axis.axis_enabled"] = false;
        } else if (spacing < X_LABEL_ROTATE_90_THRESHOLD) {
          if (checkLabelHeight(chart, 90)) {
            chart.settings["graph.x_axis.axis_enabled"] = "rotate-90";
          } else {
            chart.settings["graph.x_axis.axis_enabled"] = false;
          }
        } else {
          if (checkLabelHeight(chart, 45)) {
            chart.settings["graph.x_axis.axis_enabled"] = "rotate-45";
          } else {
            chart.settings["graph.x_axis.axis_enabled"] = false;
          }
        }
      } else {
        chart.settings["graph.x_axis.axis_enabled"] = false;
      }
    }
  }
}
```

To summarize, for numeric (linear, log, power) and timeseries x-axes, if labels will overlap then the axis will be hidden. For categorical (ordinal, historgarm) x-axes, if labels overlap it will try to use 90 degree or 45 degree rotated labels, or it will hide them if those overlap as well.

### Examples

question/7769
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194101957459

question/14061
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194432975859

question/97
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194969563009

PR #38935, opened 2024-02-19
fix cut off tick labels

EmmadUsmani:
Closes https://github.com/metabase/metabase/issues/37420
Closes https://github.com/metabase/metabase/issues/37275

### Description

Fixes the issue by adding extra padding for long dimension labels.

### Demo

250 label was cut off before

last date was cut off before
### Checklist


EmmadUsmani:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/38935?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `echarts`
  * **PR #38731** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38731?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #38935** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38935?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #38789** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38789?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38935?utm_source=stack-comment).

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
```diff
@@ -60,6 +60,37 @@ const getYAxisTicksWidth = (
...
+
+  if (!xAxisDisplay) {
+    return 0;
+  }
+
+  if (xAxisDisplay === "rotate-90") {
+    return CHART_STYLE.axisTicks.size;
+  }
+
+  const tickWidths = dataset.map(datum => {
```
alxnddr, line 79: I think it is more reasonable to check bordering ticks with since later in the calculation we assume they are at the ends of the axis

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
```diff
@@ -160,13 +217,22 @@ export const getChartPadding = (
...
   const yAxisNameTotalWidth =
     CHART_STYLE.axisName.size / 2 + CHART_STYLE.axisNameMargin;
-
-  if (chartModel.leftAxisModel?.label) {
-    padding.left += yAxisNameTotalWidth;
-  }
-  if (chartModel.rightAxisModel?.label) {
-    padding.right += yAxisNameTotalWidth;
-  }
+  const xAxisTickOverflow = tickDimensions.xTicksWidth / 2;
```
alxnddr, line 220: Based on the previous implementation, there still should be an allowed max limit so that the chart does not squash fully on very long ticks

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: Makes sense I can do that

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: That makes sense. Where in the code did you find the previous max limit?

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: Done

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: Done

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
```diff
@@ -161,12 +235,41 @@ export const getChartPadding = (
   const yAxisNameTotalWidth =
     CHART_STYLE.axisName.size / 2 + CHART_STYLE.axisNameMargin;
 
-  if (chartModel.leftAxisModel?.label) {
-    padding.left += yAxisNameTotalWidth;
-  }
-  if (chartModel.rightAxisModel?.label) {
-    padding.right += yAxisNameTotalWidth;
-  }
+  const currentBoundaryWidth =
```
alxnddr, line 238: For the boundary width calculation we still need to consider yAxisNameTotalWidth but you removed it above

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
```diff
@@ -161,12 +235,41 @@ export const getChartPadding = (
...
+    chartWidth -
+    padding.left -
+    padding.right -
+    ticksDimensions.yTicksWidthLeft -
+    ticksDimensions.yTicksWidthRight;
+  const dimensionWidth = currentBoundaryWidth / chartModel.dataset.length;
+  const firstTickPosition = dimensionWidth / 2;
+  const lastTickPosition = currentBoundaryWidth - dimensionWidth / 2;
```
alxnddr, line 246: Since we dont't really need ticks positions within the boundary I found this a bit confusing.
For category scales ticks are located at the center of each category segment so I guess it can be
```
const categoryTickOffset = dimensionWidth / 2
let firstTickOverflow = firstTickWidth / 2
let lastTickOverflow = lastTickWidth / 2

// on category scale the tick center is in the middle of the category segment
// while on time scale tick can center can be right on the edge
if (isOrdinal) {
  firstTickOverflow -= categoryTickOffset
  lastTickOverflow -= categoryTickOffset
}

// apply limiting of huge tick overflows
``

**Review by alxnddr — COMMENTED** [review]
When ticks are rotated, they don't take the space on the right:
Screenshot 2024-02-21 at 1 24 20 AM

I think allowing charts to shrinking that much to only 50% width because of ticks can be not ideal. In the old implementation we add smaller padding and let ticks overflow
Screenshot 2024-02-21 at 1 26 02 AM

I think something is not right because on this example there is an excessive padding on the left. The left tick is placed exactly at the edge so there should be only `CHART_STYLE.padding.x` picels between the edge and the tick which is 8 additional pixels, but there is 40:

Screenshot 2024-02-21 at 1 29 33 AM

Looks great on numeric scales 👍 
Screenshot 2024-02-21 at 1 34 06 AM

File: frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts [hunk 5]
```diff
@@ -136,14 +182,42 @@ export const getTicksDimensions = (
...
 
+const getExtraSidePadding = (
+  xAxisTickOverflow: number,
+  yAxisNameTotalWidth: number,
+  yAxisTicksWidth: number,
+  padding: number,
+  yAxisModel: YAxisModel | null,
+) => {
+  if (yAxisModel?.label == null) {
+    return xAxisTickOverflow > padding ? xAxisTickOverflow : 0;
```
alxnddr, line 207: Seems like this function can be:
```
const yAxisTotalWidth = yAxisTicksWidth + CHART_STYLE.axisTicksMarginY + (yAxisModel?.label != null ? yAxisNameTotalWidth : 0)
return Math.max(xAxisTickOverflow, yAxisTotalWidth)
```

alxnddr:
The axis name changet its position here:
Screenshot 2024-02-21 at 1 50 46 AM

EmmadUsmani:
@alxnddr 

> When ticks are rotated, they don't take the space on the right:

Makes sense I can fix this

> I think allowing charts to shrinking that much to only 50% width because of ticks can be not ideal. In the old implementation we add smaller padding and let ticks overflow

Can you share where in the old code this happens? I am not able to find anything related to it, so I don't know what the padding threshold should be. It may be a dc.js internal behavior. Regardless, it's not a big issue because this case won't come up often. In this case what the old implenation would do is automatically rotate or hide the labels, even when the viz setting is "show". For example:

https://github.com/metabase/metabase/assets/37751258/3467592b-6efc-4a46-8b79-17ab50cf4a29

This will be addressed by fixing https://github.com/metabase/metabase/issues/38783. 

> there is an excessive padding on the left

I'll look into this

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: Sure I can remove the position.

> // while on time scale tick can center can be right on the edge

Can you please share an example of this happening? On every timeseries chart I look at the label is centered over its datum

Screenshot 2024-02-23 at 10 18 21 AM

EmmadUsmani:
@alxnddr 

> The axis name changed its position here:

That's because I haven't rebased my PRs on top of the new waterfall chart yet. Once I do it will be fixed.

EmmadUsmani:
@alxnddr I ended up reducing the padding threshold, so now padding can only be up to 25% of the chart.

I still don't think this will come up often, more often than this the default setting will just do something better (hide or rotate labels).

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: Fixed

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
EmmadUsmani: Fixed

frontend/src/metabase/visualizations/echarts/cartesian/chart-measurements/index.ts
```diff
@@ -159,15 +225,46 @@ export const getChartPadding = (
...
-  if (chartModel.leftAxisModel?.label) {
-    padding.left += yAxisNameTotalWidth;
-  }
-  if (chartModel.rightAxisModel?.label) {
-    padding.right += yAxisNameTotalWidth;
+  const currentBoundaryWidth =
+    chartWidth -
+    padding.left -
+    padding.right -
+    yAxisNameTotalWidth -
```
alxnddr, line 234: We should consider `yAxisNameTotalWidth` based on the numer of axes and whether they have name displayed, that's why previously we explicitly checked for each axis label existence

alxnddr (approved):
Looks great! 🙌 
I left one last comment, please have a look before merging

alxnddr:
> Can you share where in the old code this happens?
I did not find it in our code, so it most likely happens inside dc.js and it looks like this:

Screenshot 2024-02-13 at 5 17 55 PM

EmmadUsmani:
### Merge activity

* **Feb 27, 3:57 PM EST**: @EmmadUsmani started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38935).

PR #30537, opened 2023-05-03
Escape schema and table names in Snowflake JDBC metadata calls

metamben:
Fixes #26054.

The Snowflake JDBC driver is buggy: schema and table name are interpreted as patterns in DatabaseMetaData.getPrimaryKeys and DatabaseMetaData.getImportedKeys calls.

This PR replaces the default JDBC functions with ones that escape the names.

The implementation is a tweaked copy of the functions in `metabase.driver.sql-jdbc.sync.describe-table`.
We could generalize those functions so that the snowflake driver can override just the pieces it has to, but
I decided that we shouldn't generalize just to be able to support buggy drivers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/metabase/metabase/30537)

PR #13340, opened 2020-10-01
Handle slashes in oracle schema/table names #13225

dpsutton:
table names like `table_slash/` are valid in our normal query
apparatus but are invalid in the calls to `.getColumns` and
`.getTables` when determining the shape of the database. In these
contexts it needs to be `table_slash//`

- also get fix TODO to reduce over the list of tables rather than
realizing entire result set and then reducing

###### Before submitting the PR, please make sure you do the following
- [ ] If you're attempting to fix a translation issue, please submit your changes to our [POEditor project](https://poeditor.com/join/project/ynjQmwSsGh) instead of opening a PR.
### Tests
-  [ ] Run the frontend and integration tests with  `yarn lint && yarn flow && yarn test`)
-  [x] If there are changes to the backend codebase, run the backend tests with `lein test && lein lint && ./bin/reflection-linter`

-  [x] Sign the [Contributor License Agreement](https://docs.google.com/a/metabase.com/forms/d/1oV38o7b9ONFSwuzwmERRMi9SYrhYeOrkbmNaq9pOJ_E/viewform)
(unless it's a tiny documentation change).

PR #38946, opened 2024-02-20
Postgres: Fix sync does not recognize schema with backslashes

### LINKED ISSUE #38000: Schemas with backslashes in Postgres don't get synced
### Describe the bug

Schemas with backslashes in Postgres don't get synced. Might affect other DBs too but I've only tested on Postgres.

### To Reproduce

1. Connect a Postgres database to Metabase
2. Create a new schema in the DB that contains an (escaped) backslash, e.g. `CREATE SCHEMA "schema_with\\backslash";`
3. Create a couple tables in the schema
4. Sync the database
5. The sync will succeed, but neither the schema nor any of its tables will have been synced

### Expected behavior

_No response_

### Logs

_No response_

```JSON
Current master
```

### Severity

P2?

### Additional context

_No response_

qnkhuat:
Closes #38000


calherries (changes requested):
This tests that `describe-table-fields` works but doesn't test that primary keys and foreign keys are correctly synced for postgres, which I suspect have the same problem because we use the jdbc metadata methods.

Seems like this is also an [issue with oracle](https://github.com/metabase/metabase/blob/2e800cc30be4fb8a17162a8240042dbb155ca249/modules/drivers/oracle/src/metabase/driver/oracle.clj#L558)'s primary keys.

I think it would be best to create a test for all drivers using `mt/with-temp-test-data` to create the table and columns, given this is at least the third time we've run into this. (see similar PRs for [oracle](https://github.com/metabase/metabase/pull/13340) and [snowflake](https://github.com/metabase/metabase/pull/30537/files)).

Unfortunately we don't have a general way to create and sync a test schema for all drivers because some cloud drivers use session schemas, and syncing new schemas in tests is a bit of a hack but could be done.

calherries:
Oh, it looks like you tried creating a test for all drivers [here](https://github.com/metabase/metabase/pull/38946/commits/2c0b084e4d17a7d9a2c9e5ff16a0690dd3e904f4#diff-6598413869c2c34f4fcc6b36e13f5296ab7a822f446240bb82289fc979fbe1bb), and then removed it. 

If you want to make this PR just about postgres that's cool. But I think it should be be followed up with a PR that adds a test for all drivers because I suspect there will be more cases.

test/metabase/driver/sql_jdbc/sync/describe_database_test.clj
```diff
@@ -208,3 +208,23 @@
                                   driver/*driver* conn schema (str table-name "_should_not_exist"))))
                      (is (true? (sql-jdbc.sync.interface/have-select-privilege?
                                  driver/*driver* conn schema table-name))))))))))))))
+
+(deftest sync-table-with-backslash-test
+  (mt/test-drivers #{:postgres} ;; TODO: fix and change this to test on (mt/sql-jdbc-drivers)
+    (testing "table with backslash in name, PKs, FKS are correctly synced"
+      (mt/with-temp-test-data [["human\\race"
+                                [{:field-name "humanraceid" :base-type :type/Integer :pk? true}
```
qnkhuat, line 216: the ability to define a custom PK is form #39038

test/metabase/driver/sql_jdbc/sync/describe_database_test.clj
```diff
@@ -208,3 +208,23 @@
...
+                                 {:field-name "race" :base-type :type/Text}]
+                                [[1 "homo sapiens"]]]
+                               ["citizen"
+                                [{:field-name "citizen\\id" :base-type :type/Integer :pk? true}
+                                 {:field-name "race\\id" :base-type :type/Integer :fk "human\\race"}]
+                                [[1 1]]]]
+        (sync/sync-database! (mt/db))
+        (let [tables            (t2/select :model/Table :db_id (:id (mt/db)))
+              field-name->field (t2/select-fn->fn :name identity :model/Field :table_id [:in (map :id tables)])]
+          (is (= #{"human\\race" "citizen"} (set (map :name tables))))
```
calherries, line 226: Should it be one backslash?
```suggestion
          (is (= #{"human\race" "citizen"} (set (map :name tables))))
```

qnkhuat: we need one for escaping the "\" :p

calherries: ah yeah doh

qnkhuat:
The failing test is from the upstream PR, I'll fix it there.

test/metabase/driver/sql_jdbc/sync/describe_database_test.clj
```diff
@@ -208,3 +208,23 @@
...
+    (testing "table with backslash in name, PKs, FKS are correctly synced"
+      (mt/with-temp-test-data [["human\\race"
+                                [{:field-name "humanraceid" :base-type :type/Integer :pk? true}
+                                 {:field-name "race" :base-type :type/Text}]
+                                [[1 "homo sapiens"]]]
+                               ["citizen"
+                                [{:field-name "citizen\\id" :base-type :type/Integer :pk? true}
+                                 {:field-name "race\\id" :base-type :type/Integer :fk "human\\race"}]
+                                [[1 1]]]]
+        (sync/sync-database! (mt/db))
```
calherries, line 223: This is a redundant sync since `mt/with-temp-test-data` will run a metadata sync
```suggestion
```
see https://github.com/metabase/metabase/blob/6c7fbaf818ac13fc326447907ab5c723d77f2b5d/test/metabase/test/data/impl/get_or_create.clj#L122

test/metabase/driver/postgres_test.clj
```diff
@@ -212,7 +212,20 @@
...
-                              :query    {:source-table (t2/select-one-pk Table :name "presents-and-gifts")}}))))))))))
+                              :query    {:source-table (t2/select-one-pk Table :name "presents-and-gifts")}}))))))))
+    (testing "Make sure that Schemas / Tables / Fields with backslashes in their names get escaped properly"
+      (mt/with-empty-db
+        (let [conn-spec (sql-jdbc.conn/db->pooled-connection-spec (mt/db))]
+          (doseq [stmt ["CREATE SCHEMA \"my\\schema\";"
+                        "CREATE TABLE \"my\\schema\".\"my\\table\" (\"my\\field\" INTEGER);"
+                        "INSERT INTO \"my\\schema\".\"my\\table\" (\"my\\field\") VALUES (42);"]]
+            (jdbc/execute! conn-spec stmt))
+          (sync/sync-database! (mt/db))
```
calherries, line 223: This doesn't need to be a full db sync
```suggestion
          (sync/sync-database! (mt/db) {:scan :schema})
```

test/metabase/driver/sql_jdbc/sync/describe_database_test.clj
```diff
@@ -208,3 +208,23 @@
                                   driver/*driver* conn schema (str table-name "_should_not_exist"))))
                      (is (true? (sql-jdbc.sync.interface/have-select-privilege?
                                  driver/*driver* conn schema table-name))))))))))))))
+
+(deftest sync-table-with-backslash-test
```
calherries, line 212: If we're going to do it for all drivers, ideally we would also test on forward-slashes too, since forward slashes were the problem in https://github.com/metabase/metabase/pull/13340

calherries (approved):
I apologize for asking for tests for primary key and foreign key sync. I now see that `getPrimaryKeys` and `getForeignKeys` are different from `getTables` and `getColumns`. The former take `table` and `schema` args which don't need escaping while the latter take `schemaPattern` and `tablePattern` args, which do need escaping.

I think `driver/escape-entity-name-for-metadata`'s name and docstring could be updated, because I mistakenly thought this would apply to all JDBC metadata calls.

You have some tests failing, but I'm approving this assuming they'll be fixed in the upstream PR.

PR #33883, opened 2023-09-13
RFC 70 - Automate sorting imports

kamilmielnik:
https://www.notion.so/metabase/70-Automate-sorting-imports-74133d1e110e4605b079d153baf57953

### Description

- Due to a huge diff the only way to review this is to go over [commits](https://github.com/metabase/metabase/pull/33883/commits) one by one
- There are 2 commits containing automatic fixes ([Run "yarn eslint-fix" 5 times](https://github.com/metabase/metabase/pull/33883/commits/a95906a9b3bf5af7da92ccbc1c2e446e7fcc2bba) & [Run "yarn prettier"](https://github.com/metabase/metabase/pull/33883/commits/004d139aef2dc500b9a6294adce8b526a094e350)), all other were manual or quasi-manual
    - I had to run `yarn eslint-fix` multiple times, as it was not able to fix all issues in a single run - with every subsequent run there were fewer and fewer rule violations
- `import/order` rule does not play well with comments or code in between import statements, so I had to do some adjustments before running `yarn eslint-fix`:
  - (75aa0de96bad2f9aeb0d6fdc714e12be44700979, 898d16ed8cc9061c453a20700b45815cb8578ec9) moved comments into the same line as imports
  - (d6b4e6d4395e6f4c40d64d57c5c8fd3d50de52a1) moved imports into a single block (where possible)
  - (1ae842d36e0bf967ace351fcd3984801c2e2d07a) moved or removed informative comments between import statements
  - (e1e072db4eb1df81c9fee182d4560f8df5a6c00d) removed imports-grouping comments (now superseded by `import/order` rule)
  - (071432d50668e907a3835b7e52b3ef53dbad4429) disabled `import/order` rule in files where apparently the order of imports matters
- After `yarn eslint-fix` there were still some files that eslint complained about, so I manually fixed them (b2e4a5420f037acbe3f5afef53ddd751b3aa5190)
- After all this there was 1 circular dependency issue that came up in unit tests - fixed in [this commit](https://github.com/metabase/metabase/pull/33883/commits/b31ec8d03a8e50e25b3ecc04cee37eb9739f9254)

This change should not be backported. I tried to work on this as soon as we cut the `release-x.49.x` branch.

### How to verify

- CI checks should pass (with exception of unrelated flaky tests mentioned below)
- App should work exactly the same
- Verify commits one by one
    - it's ok to skip the commits that merge `master` into this branch
[...]

PR #38948, opened 2024-02-20
Include e2e directory in prettier scripts

kamilmielnik:
Follow up on #33883

Resolves [Slack discussion](https://metaboat.slack.com/archives/C505ZNNH4/p1708419501886529?thread_ts=1708415055.711689&cid=C505ZNNH4)

### Description

Our `prettier` and `lint-prettier-pure` scripts did not include `e2e` dir: 96016df6d73ddd74d7e9ee8765bcdf0e078cc1b4
I updated them and then I just run `yarn prettier`: 9a6c36b523e08d7bd81dd406e895f1d242af6000


uladzimirdev (approved):

package.json
```diff
@@ -342,12 +342,12 @@
     "lint-eslint": "yarn build:cljs && yarn lint-eslint-pure",
     "lint-eslint-pure": "eslint --ext .js,.jsx,.ts,.tsx --rulesdir frontend/lint/eslint-rules --max-warnings 0 --report-unused-disable-directives enterprise/frontend/src frontend/src frontend/test e2e/test",
     "lint-prettier": "yarn && yarn lint-prettier-pure",
-    "lint-prettier-pure": "prettier '{enterprise/,}frontend/**/*.{js,jsx,ts,tsx,css}' --check",
+    "lint-prettier-pure": "prettier '{frontend,enterprise,e2e}/**/*.{js,jsx,ts,tsx,css}' --check",
```
WiNloSt, line 345: I'm not an expert in doing file patterns, but what I've tested is

```suggestion
    "lint-prettier-pure": "prettier '{{enterprise/,}frontend,e2e}/**/*.{js,jsx,ts,tsx,css}' --check",
```

This is supposed to keep the previous directory search
- `enterprise/frontend/*`
- `frontend/*`
- and add a new one `e2e/*`

The difference to the change in this PR is only `enterprise/*` vs `enterprise/frontend/*`

I think the impact would be minimal, but I'd like to keep just add `e2e` directory to the search if possible and not changing the other directories settings.

package.json
```diff
@@ -342,12 +342,12 @@
...
     "lint-yaml": "yamllint **/*.{yaml,yml} --ignore=node_modules/**/*.{yaml,yml}",
     "precommit": "lint-staged",
     "preinstall": "echo $npm_execpath | grep -q yarn || echo '\\033[0;33mSorry, npm is not supported. Please use Yarn (https://yarnpkg.com/).\\033[0m'",
     "prepare": "husky install",
-    "prettier": "prettier --write '{enterprise/,}frontend/**/*.{js,jsx,ts,tsx,css}'",
+    "prettier": "prettier --write '{frontend,enterprise,e2e}/**/*.{js,jsx,ts,tsx,css}'",
```
WiNloSt, line 350: The comment above also applies.

package.json
kamilmielnik: Notice that `lint-staged` scripts are working with the entire `enterprise` directory, not `enterprise/frontend`, so now it will be more in sync (still not perfectly though).

> I think the impact would be minimal, but I'd like to keep just add e2e directory to the search if possible and not changing the other directories settings.

Makes sense but only if we change `lint-staged` scripts too.

It's best to merge this PR ASAP to prevent other people from experiencing the problem.
I'll submit a follow-up PR that will align `scripts` and `lint-staged` scripts to use the same paths.

Sounds good?

WiNloSt: Sounds good. Glad to have this PR go into master ASAP

WiNloSt (approved):

npretto:
In the past I always did `prettier --write .` and let it format everything, if needed I setup a prettier ignore file (for example for the builds). For lint staged I usually just do `prettier --write --ignore-unknown` which ignores files it doesn't know how to format.

What do you think about it?

This ^ will also format json,md,yaml and all the other files types that prettier knows how to format, I like that, but it's a change not everybody could agree

kamilmielnik:
> In the past I always did `prettier --write .` and let it format everything, if needed I setup a prettier ignore file (for example for the builds). For lint staged I usually just do `prettier --write --ignore-unknown` which ignores files it doesn't know how to format.
> 
> What do you think about it?

@npretto I'm on board with this :+1:  

> This ^ will also format json,md,yaml and all the other files types that prettier knows how to format, I like that, but it's a change not everybody could agree

Not sure if RFC is needed, but it would not hurt asking during frontend guild or in Slack.

kamilmielnik:
> For lint staged I usually just do `prettier --write --ignore-unknown` which ignores files it doesn't know how to format.

@npretto Paths that we use for `lint-staged` are rather _optimized_, so this might result in increased running time of git hooks.

PR #38951, opened 2024-02-20
Align paths in Prettier & ESLint scripts

kamilmielnik:
Follow-up on https://github.com/metabase/metabase/pull/38948#discussion_r1495513696

### Description

- enables ESLint in entire `e2e` dir (previously `e2e/test` only)
- enables ESLint in entire `frontend` dir (previously `frontend/src` only)
- fixes ESLint-reported issues in aformentioned directories
- enables Prettier for `*.css` files in `e2e` dir (for consistency of the paths; no CSS files exist in `e2e` dir yet)
- aligns paths in Prettier & ESLint `scripts` & `lint-staged` scripts
- ensures test scenarios written in TS are checked (none exist yet, but we do support [TS in e2e tests now](https://github.com/metabase/metabase/pull/36474))

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,4 @@
 frontend/src/cljs
 frontend/src/cljs_release
+e2e/support/cypress_sample_database.js
+e2e/support/cypress_sample_instance_data.js
--- a/package.json
+++ b/package.json
@@ -388,7 +388,6 @@
       "prettier --write"
     ],
     "e2e/test/scenarios/*/{*.(js|ts),!(helpers|shared)/*.(js|ts)}": [
-      "eslint --rulesdir frontend/lint/eslint-rules --max-warnings 0 --fix",
       "node e2e/validate-e2e-test-files.js"
     ],
     "**/*.{clj,cljc,cljs,bb}": [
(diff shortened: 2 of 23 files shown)


.eslintignore
```diff
@@ -1,2 +1,4 @@
 frontend/src/cljs
 frontend/src/cljs_release
+e2e/support/cypress_sample_database.js
+e2e/support/cypress_sample_instance_data.js
```
kamilmielnik, line 4: Both of these files import an automatically generated JSON file which is not tracked in git.

In CI ESLint reports `import/no-unresolved` violation, so we need to `// eslint-disable import/no-unresolved` it.
But locally the JSON files exist, so ESLint reports `Unused eslint-disable directive (no problems were reported from 'import/no-unresolved')`.

To make ends meet, the simplest solution was to to just ignore these files.

WiNloSt: sounds reasonable

package.json
```diff
@@ -387,7 +387,7 @@
       "eslint --rulesdir frontend/lint/eslint-rules --max-warnings 0 --fix",
       "prettier --write"
     ],
-    "e2e/test/scenarios/*/{*.js,!(helpers|shared)/*.js}": [
+    "e2e/test/scenarios/*/{*.(js|ts),!(helpers|shared)/*.(js|ts)}": [
```
WiNloSt, line 390: One minor comment, not exactly about this change, but the line below.

Given we already run every staged .ts and .js file with `eslint --rulesdir frontend/lint/eslint-rules --max-warnings 0 --fix` via line 387.

Wouldn't it be doing the same thing twice on line 391? My hunch is that we could remove line 391 and leave only `node e2e/validate-e2e-test-files.js`

WiNloSt (approved):
Overall this is looking great.

package.json
kamilmielnik: Nice catch :heart:  45bafbfd30

Issue #38867, opened 2024-02-16
[Epic] Add license activation in the initial setup


npretto:
**Links**
- product doc: [Add license activation step in /setup](https://www.notion.so/metabase/Add-license-activation-step-in-setup-c186326d2efd4467a7e9d09138fca48c)
- eng doc: [Add license activation step in /setup](https://www.notion.so/metabase/Add-license-activation-step-in-setup-77d8f3e5c2484226b8a162098b6eea5b)
- feature branch: [`license-activation-step`](https://github.com/metabase/metabase/pull/39257)

***Milestone 1***

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/38858
- [ ] https://github.com/metabase/metabase/pull/38953
- [ ] https://github.com/metabase/metabase/pull/39151
- [ ] https://github.com/metabase/metabase/pull/39103
- [ ] https://github.com/metabase/metabase/pull/39292
- [ ] https://github.com/metabase/metabase/pull/39440
- [ ] https://github.com/metabase/metabase/pull/39443
- [ ] https://github.com/metabase/metabase/pull/39257
```

***Milestone 2 - polishing***

```[tasklist]
### Tasks
- [ ] https://github.com/metabase/metabase/pull/39067
- [ ] https://github.com/metabase/metabase/pull/39931
- [x] ask schema change for the events
```

PR #32432, opened 2023-07-17
Feature flags integration

ranquild:
Added a new feature flags, to make permissions more granular

PR #38953, opened 2024-02-20
Add the step in the FE and pass the token to the setup api call

npretto:
Part of [[Epic] Add license activation in the initial setup](https://github.com/metabase/metabase/issues/38867)

During development I encountered an issue, the getSteps selector that I introduced in the previous epic was being used in reducers. All good until it started needing access to global state, that the reducers don't have access to..
I did a little hacky solution to that in 0318a55b22b189af9e30d4a8116bacfb0365fcd2 with a initSetup function that copies the global piece of stat needed in the slice so that it's all good.
It's... a solution.. it works but meh

I tried to find other solutions and I opened two prs for them:

- [goToNextStep in all thunks (+42 −56)](https://github.com/metabase/metabase/pull/39103)
    - very simple, it works and is probably a good compromise
- [dispatchAndGoNextStep (+119 −108)](https://github.com/metabase/metabase/pull/39121)
    - I tried to be smarter, it backfired in tests failing because of async stuff 🤷 🤷 🤷
    - considering all the failures in the tests, I don't think it's worth going with this route)

Let me know which of the three solutions (initSetup + 2 in the prs) you want to go with 

 # other notable changes/commits in the PR
- ea045e93c1fb8932d3675d686065f57f7c203645 to see what I changed in the events schema
- 1d8d592072971a6a4119210323c055874c5c9aba useStep hook

# Things still missing that will come in future PRs
- analytics event
- handling of the error when the BE cannot connect to the token check server, I already started working on the BE for that [here](https://github.com/metabase/metabase/pull/39067)

### DIFF
--- a/e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
+++ b/e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
@@ -5,6 +5,7 @@ import {
   expectGoodSnowplowEvent,
   expectGoodSnowplowEvents,
   expectNoBadSnowplowEvents,
+  isEE,
   main,
   resetSnowplow,
   restore,
@@ -264,6 +265,11 @@ describe("scenarios > setup", () => {
       cy.button("Connect database").click();
     });
 
+    if (isEE) {
+      // license_token
+      cy.button("Next").click();
+    }
+
     // usage data
     cy.get("section").last().button("Finish").click();
 
@@ -369,18 +375,21 @@ describeWithSnowplow("scenarios > setup", () => {
   });
 
   it("should send snowplow events", () => {
-    // 1 - new_instance_created
-    // 2 - pageview
+    let goodEvents = 0;
+
+    goodEvents++; // 1 - new_instance_created
+    goodEvents++; // 2 - pageview
     cy.visit(`/setup`);
 
-    // 3 - setup/step_seen "welcome"
+    goodEvents++; // 3 - setup/step_seen "welcome"
     expectGoodSnowplowEvent({
       event: "step_seen",
       step_number: 0,
       step: "welcome",
     });
     skipWelcomePage();
-    // 4 - setup/step_seen  "language"
+
+    goodEvents++; // 4 - setup/step_seen  "language"
     expectGoodSnowplowEvent({
       event: "step_seen",
       step_number: 1,
@@ -388,12 +397,13 @@ describeWithSnowplow("scenarios > setup", () => {
     });
     selectPreferredLanguageAndContinue();
 
-    // 5 - setup/step_seen "user_info"
+    goodEvents++; // 5 - setup/step_seen "user_info"
     expectGoodSnowplowEvent({
       event: "step_seen",
       step_number: 2,
       step: "user_info",
     });
+
     cy.findByTestId("setup-forms").within(() => {
       fillUserAndContinue({
         ...admin,
@@ -401,7 +411,7 @@ describeWithSnowplow("scenarios > setup", () => {
       });
 
       cy.findByText("What will you use Metabase for?").should("exist");
-      // 6 - setup/step_seen "usage_question"
+      goodEvents++; // 6 - setup/step_seen "usage_question"
       expectGoodSnowplowEvent({
         event: "step_seen",
         step_number: 3,
@@ -409,41 +419,56 @@ describeWithSnowplow("scenarios > setup", () => {
       });
       cy.button("Next").click();
 
-      // 7 - setup/usage_reason_selected
+      goodEvents++; // 7 - setup/usage_reason_selected
       expectGoodSnowplowEvent({
         event: "usage_reason_selected",
         usage_reason: "self-service-analytics",
       });
-      // 8 - setup/step_seen "db_connection"
+
+      goodEvents++; // 8 - setup/step_seen "db_connection"
       expectGoodSnowplowEvent({
         event: "step_seen",
         step_number: 4,
         step: "db_connection",
       });
       cy.findByText("I'll add my data later").click();
-      // 9 - setup/add_data_later_clicked
+
+      goodEvents++; // 9/10 - setup/add_data_later_clicked
       expectGoodSnowplowEvent({
         event: "add_data_later_clicked",
       });
-      // 10 - setup/step_seen "data_usage"
+
+      // This step is only visile on EE builds
+      if (isEE) {
+        goodEvents++; // 10/11 - setup/step_seen "commercial_license"
+        expectGoodSnowplowEvent({
+          event: "step_seen",
+          step_number: 5,
+          step: "license_token",
+        });
+
+        cy.button("Next").click();
+      }
+
+      goodEvents++; // 10/11 - setup/step_seen "data_usage"
       expectGoodSnowplowEvent({
         event: "step_seen",
-        step_number: 5,
+        step_number: 5 + (isEE ? 1 : 0),
         step: "data_usage",
       });
 
       cy.findByRole("button", { name: "Finish" }).click();
-      // 11 - new_user_created (from BE)
+      goodEvents++; // 11/12 - new_user_created (from BE)
 
-      // 12- setup/step_seen "completed"
+      goodEvents++; // 12/13- setup/step_seen "completed"
       expectGoodSnowplowEvent({
         event: "step_seen",
-        step_number: 6,
+        step_number: 6 + (isEE ? 1 : 0),
         step: "completed",
       });
-    });
 
-    expectGoodSnowplowEvents(12);
+      expectGoodSnowplowEvents(goodEvents);
+    });
   });
 
   it("should ignore snowplow failures and work as normal", () => {
--- a/e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
+++ b/e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
@@ -456,7 +456,7 @@ describeWithSnowplow("scenarios > setup", () => {
       goodEvents++; // 10/11 - setup/step_seen "data_usage"
       expectGoodSnowplowEvent({
         event: "step_seen",
-        step_number: 5 + (isEE ? 1 : 0),
+        step_number: isEE ? 6 : 5,
         step: "data_usage",
       });
 
@@ -466,7 +466,7 @@ describeWithSnowplow("scenarios > setup", () => {
       goodEvents++; // 12/13- setup/step_seen "completed"
       expectGoodSnowplowEvent({
         event: "step_seen",
-        step_number: 6 + (isEE ? 1 : 0),
+        step_number: isEE ? 7 : 6,
         step: "completed",
       });
 
--- a/e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
+++ b/e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
@@ -127,6 +127,8 @@ describe("scenarios > setup", () => {
           // test database setup help card is hidden on the next step
           cy.findByText("Need help connecting?").should("not.be.visible");
 
+          skipLicenseStepOnEE();
+
           // ================
           // Data Preferences
           // ================
@@ -187,6 +189,8 @@ describe("scenarios > setup", () => {
       // Database
       cy.findByText("Add your data");
       cy.findByText("I'll add my data later").click();
+
+      skipLicenseStepOnEE();
 
       // Turns off anonymous data collection
       cy.findByLabelText(
@@ -265,10 +269,7 @@ describe("scenarios > setup", () => {
       cy.button("Connect database").click();
     });
 
-    if (isEE) {
-      // license_token
-      cy.button("Next").click();
-    }
+    skipLicenseStepOnEE();
 
     // usage data
     cy.get("section").last().button("Finish").click();
@@ -315,6 +316,8 @@ describe("scenarios > setup", () => {
 
       // Database
       cy.findByText("Add your data").should("not.exist");
+
+      skipLicenseStepOnEE();
 
       // Turns off anonymous data collection
       cy.findByLabelText(
@@ -528,4 +531,10 @@ const fillUserAndContinue = ({
     cy.findByLabelText("Confirm your password").type(password);
   }
   cy.button("Next").click();
+};
+
+const skipLicenseStepOnEE = () => {
+  if (isEE) {
+    cy.button("Next").click();
+  }
 };
--- a/frontend/src/metabase/selectors/settings.ts
+++ b/frontend/src/metabase/selectors/settings.ts
@@ -103,6 +103,9 @@ export const getUpgradeUrl = createSelector(
   },
 );
 
+/**
+ * ! The tokenStatus is only visible to admins
+ */
 export const getIsPaidPlan = createSelector(
   (state: State) => getSetting(state, "token-status"),
   (tokenStatus: TokenStatus | null) => {
--- a/frontend/src/metabase/setup/actions.ts
+++ b/frontend/src/metabase/setup/actions.ts
@@ -23,6 +23,7 @@ import {
   getIsTrackingAllowed,
   getLocale,
   getSetupToken,
+  getSteps,
   getUsageReason,
   getUser,
 } from "./selectors";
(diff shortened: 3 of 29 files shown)


frontend/src/metabase/setup/selectors.ts
```diff
@@ -77,19 +79,29 @@ export const getAvailableLocales = (state: State): LocaleData[] => {
...
 };
-
+``;
 export const getSteps = (state: State) => {
   const usageReason = getUsageReason(state);
   const activeStep = getStep(state);
 
+  // TODO: avoid using MetabaseSettings in a selector
+  // we can't do it right now as this selector is used in the reducer that doesn't have access to the whole state
+  const tokenFeatures = MetabaseSettings.get("token-features");
```
WiNloSt, line 89: It's uncommon for reducers to know about the whole state other than its own slice. The way this problem is solved is via sending necessary information via action payload.

You could do this by sending necessary data e.g. `isPaidPlan` in [the action](https://github.com/metabase/metabase/pull/38953/files#diff-f0f1344935a4c128194840b010588c50a1260aa9dae823a4b8c51dee7e130284R97). You can access `getState` in thunk action and that's how you could get `token-features` setting from the global Redux state. Then probably either send `isPaidPlan` or `shouldShowLicenseStep` depending on how you think it's best to encapsulate the information.

After that this reducer won't need the access to `MetabaseSettings` anymore.

frontend/src/metabase/selectors/settings.ts
```diff
@@ -103,6 +103,9 @@ export const getUpgradeUrl = createSelector(
   },
 );
 
+/**
+ * ! The tokenStatus is only visible to admins
```
npretto, line 107: @WiNloSt git says you made this, do you remember if there was a specific reason tokenStatus was used instead of token-features which is public?
It seems that most places that are using this are inside the admin settings so we should be fine, there's only one usage I found that's outside of it: https://github.com/metabase/metabase/blob/957efd97549833357a7cde139f4302c1964a74e7/frontend/src/metabase/admin/tasks/components/Help/Help.tsx#L95

WiNloSt: I originally used `token-features`
https://github.com/metabase/metabase/blob/31951b43ddc4c4e1f707d32357e738363af27223/frontend/src/metabase/selectors/settings.ts#L51-L55

but that got changed in https://github.com/metabase/metabase/pull/32432/files#diff-531efecfe0b6ca5c45e96de468780cabafe25a4379b1b29b2f2be6f6bede1e3aR104-R107

_screenshot in case GitHub doesn't go to the change_
image

Can you ask Alex P. why that's the case?

npretto: Thanks 🙏 , moved the discussion on slack here: https://metaboat.slack.com/archives/C505ZNNH4/p1708602083391909

frontend/src/metabase/setup/tests/setup.tsx
```diff
@@ -0,0 +1,122 @@
+import { waitFor, within } from "@testing-library/react";
```
npretto, line 1: This is pretty much all stuff extracted from the old test file, and adjusted to allow passing the enterprise plugin flag and the token features

e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts
```diff
@@ -369,81 +378,100 @@ describeWithSnowplow("scenarios > setup", () => {
...
+        });
+
+        cy.button("Next").click();
+      }
+
+      goodEvents++; // 10/11 - setup/step_seen "data_usage"
       expectGoodSnowplowEvent({
         event: "step_seen",
-        step_number: 5,
+        step_number: 5 + (isEE ? 1 : 0),
```
WiNloSt, line 459: I'd do `isEE ? 6 : 5`. It should be easier to understand since we don't need to compute.

frontend/src/metabase/setup/selectors.ts
```diff
@@ -81,15 +82,20 @@ export const getIsEmailConfigured = (state: State): boolean => {
 export const getSteps = (state: State) => {
   const usageReason = getUsageReason(state);
   const activeStep = getStep(state);
+  const isPaidPlan = state.setup.isPaidPlan;
```
WiNloSt, line 85: I wish we didn't have to save `isPaidPlan` in the redux store, since that's something that's not going to change ever (unless users login to another instance with a different token of course).

However, based on how `getSteps` is used and other selectors that depend on it. I think it might be justifiable to store it in this Redux slice so that we don't have to always pass `isPaidPlan` through the selector when calling in React components.

frontend/src/metabase/setup/actions.ts
```diff
@@ -94,6 +95,13 @@ export const submitUsageReason = createAsyncThunk(
   },
 );
 
+export const submitLicenseToken = createAsyncThunk(
+  "metabase/setup/SUBMIT_LICENSE_TOKEN",
+  (_token: string | null) => {
+    // TODO: add analytics
```
npretto, line 101: I'll do it in a separate PR

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+    "idle" | "loading" | "success" | "error"
+  >("idle");
+
+  const isInputCorrectLength = token.length === 64;
+
+  const submit = async () => {
+    setStatus("loading");
+    try {
+      const response = await SetupApi.validate_token({ token });
+      if (response.valid) {
```
npretto, line 23: In a [future PR](https://github.com/metabase/metabase/pull/39067) the response will have an error_code so that we can tell if it the token is invalid, or if they BE failed to validate it for other reasons

frontend/src/metabase/setup/tests/premium.unit.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const setupPremium = (opts?: SetupOpts) => {
+  return setup({
+    ...opts,
+    hasEnterprisePlugins: true,
+    tokenFeatures: createMockTokenFeatures({ hosting: true }),
+  });
+};
+
+describe("setup (EE, hosting feature)", () => {
```
npretto, line 21: when there's a valid token, the behaviour should be same as OSS, so this only checks that the license step is not shown

oisincoveney (commented):
Looks good, just a few changes that would be helpful I think

frontend/src/metabase/setup/components/DataUsageStep/DataUsageStep.tsx
```diff
@@ -46,7 +38,7 @@ export const DataUsageStep = ({
   };
 
   const handleStepSelect = () => {
```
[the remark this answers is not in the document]
WiNloSt: Wanna chime in that I'm fine with both solutions. They're equally the same to me.

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenForm.tsx
WiNloSt: I've [commented on that PR](https://github.com/metabase/metabase/pull/39067#discussion_r1503740720), and that is relevant to the change here.

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
```diff
@@ -0,0 +1,76 @@
...
+    dispatch(
+      addUndo({
+        message: t`Your license is activated`,
+      }),
+    );
+    dispatch(submitLicenseToken(token));
+  };
+
+  const handleNext = () => {
+    dispatch(submitLicenseToken(null));
```
WiNloSt, line 40: can you explain why we need to clear the token when going to the next step? Also, from the PR, it seems that this action is only for analytics (edited: and to determine if we want to go to the next step in reducer).

frontend/src/metabase/setup/reducers.ts
```diff
@@ -54,6 +60,12 @@ export const reducer = createReducer(initialState, builder => {
     state.usageReason = usageReason;
     state.step = getNextStep({ setup: state } as State);
   });
+  builder.addCase(submitLicenseToken.pending, (state, { meta }) => {
+    const token = meta.arg;
+    state.licenseToken = token;
+    state.step = getNextStep({ setup: state } as State);
```
WiNloSt, line 66: I was confused again on why we don't have to call any navigation function inside the Token step. I'd be great to make sure we don't do the same thing in different places.

Is it a plan to move all the navigation logic to the reducer as well or do we need to still have navigation logics in both React components and in reducer?

frontend/src/metabase/setup/actions.ts
npretto: https://github.com/metabase/metabase/pull/39151

snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-3
```diff
@@ -0,0 +1,106 @@
...
+      "type": "string",
+      "maxLength": 1024
+    },
+    "step": {
+      "description": "String describing the step that the set up step",
+      "type": [
+        "string",
+        "null"
+      ],
+      "enum": [
```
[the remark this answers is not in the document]
WiNloSt: https://github.com/metabase/metabase/blob/312869f0d9eeb17b61789d31346a18fd9f970b91/frontend/src/metabase/setup/types.ts#L1-L10

From the type, is doesn't have step name which is `null`, so I don't think we need to add `null` here.

snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-3
```diff
@@ -0,0 +1,106 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Setup flow",
+  "self": {
+    "vendor": "com.metabase",
+    "name": "setup",
+    "format": "jsonschema",
+    "version": "1-0-2"
```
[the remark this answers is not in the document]
npretto: Nice catch, I actually also found it yesterday while working on the analytics pr and fixed it there: https://github.com/metabase/metabase/pull/39151/files#diff-5e4a5b1c9358e8b592634b46508741486cc177caa9769cf75c6cac531e66082dR8

snowplow/iglu-client-embedded/schemas/com.metabase/setup/jsonschema/1-0-3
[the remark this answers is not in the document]
npretto: Actually reading [the docs here](https://json-schema.org/understanding-json-schema/reference/null) it may be that we don't even need the first null (in `type`) as we never pass it explicitly as null and it's not marked as required.
To keep the schema backward compatible I wouldn't remove it though

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
npretto: We're not clearing it, the token is local and sent to redux when we submit the step (with a valid token).
If you have a look at `frontend/src/metabase/setup/actions.ts` and the reducer, all but the `welcome` and `language` steps have a thunk (which is 90% only for analytics) and then the reducer changes the state with the answer for the step and updates the step to the following

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  if (!isStepActive) {
+    const title = isStepCompleted
```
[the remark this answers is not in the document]
npretto: Good idea, 17aec48ff7b7a186a3a5b87150f3d00bc0767a81

frontend/src/metabase/setup/components/LanguageStep/LanguageStep.tsx
```diff
@@ -6,12 +6,11 @@ import Button from "metabase/core/components/Button";
 import { useDispatch, useSelector } from "metabase/lib/redux";
 import type { Locale } from "metabase-types/store";
 
+import { useStep } from "../..//useStep";
```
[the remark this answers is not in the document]
npretto: 9af72f57a7dd57a79edc8251b9bcc14c664557ba 🙃

frontend/src/metabase/setup/components/DataUsageStep/DataUsageStep.tsx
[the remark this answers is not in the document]
npretto: Yeah it saves some lines :) c4b4fad9228fc2e0bae1a07e91069feaf8642f5f

frontend/src/metabase/setup/selectors.ts
npretto: Have you seen the two PRs linked in the description about this?
Especially [the first](https://github.com/metabase/metabase/pull/39103) solves this with not many changes

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
WiNloSt: I need to run the code to actually understand what's going on. I think the main point here is that the `Next` button should be renamed to `Skip` to indicate what it actually does. We're not validating the token then click the next button to go to the next step but we can either
1. validate the token and go to the next step when clicking `Activate` button OR
2. Skip activating the token in the first place.

frontend/src/metabase/setup/reducers.ts
WiNloSt: I was confused this with `handleStepSelect`, so all good here.

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
npretto: I remember I asked Vitorio about "Next" vs "Skip"  but it was before the "Activate"  button navigated to the next step, let me ask him again, you're right with the current implementation "Next" will only be used as "Skip"

frontend/src/metabase/setup/selectors.ts
WiNloSt: I didn't see the PR when I made that comment because it was created after my comment. But that one looks good to me.

WiNloSt (approved):
Almost good to go. Would be great to see these minor issues got resolved before merging this PR though.

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { LicenseTokenForm } from "./LicenseTokenForm";
+
+export const LicenseTokenStep = ({ stepLabel }: NumberedStepProps) => {
+  const { isStepActive, isStepCompleted, handleStepSelect } =
+    useStep("license_token");
+
+  const storeToken = useSelector(state => state.setup.licenseToken);
+
+  const isSetupCompleted = useSelector(getIsSetupCompleted);
```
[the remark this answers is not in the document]
npretto: Good catch, 5e3fc7d5defed3e8d8555ba94d62ffec03bf8b7e

frontend/src/metabase/setup/tests/enterprise.unit.spec.tsx
```diff
@@ -0,0 +1,145 @@
...
+      await setup();
+      skipWelcomeScreen();
+      skipLanguageStep();
+      await submitUserInfoStep();
+      selectUsageReason("embedding"); // to skip the db connection step
+      clickNextStep();
+
+      await screen.findByText(
+        "Unlock access to your paid features before starting",
+      );
```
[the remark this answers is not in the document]
npretto: Done in 74cec2d293462920aefcb235e3855c689e6103a7

frontend/src/metabase/setup/tests/enterprise.unit.spec.tsx
```diff
@@ -0,0 +1,145 @@
...
+        sampleToken,
+      );
+
+      screen.getByRole("button", { name: "Activate" }).click();
+
+      await waitFor(() => {
+        expect(
+          screen.getByRole("button", { name: "Activate" }),
+        ).not.toHaveProperty("data-loading", true);
+      });
```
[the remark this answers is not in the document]
npretto: Ah right, I still had to leave it for one test, as it was testing a specific prop that only changed after the loading was done.
Removed the others in 4e42c8818f86bfdcf4f019393bdfdfb55fc6cbad

npretto:
Merging this to the integration merge so I can start the rebase dance, @oisincoveney let me know if I missed anything from your comments

oisincoveney:
> Merging this to the integration merge so I can start the rebase dance, @oisincoveney let me know if I missed anything from your comments

You're all good. Thanks for reaching out

PR #38962, opened 2024-02-20
Prohibit creating new data permissions

johnswanson:
And stop trying to create them.

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/email/messages_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/email/messages_test.clj
@@ -14,7 +14,7 @@
     (mt/with-user-in-groups [group {:name "New Group"}
                              user [group]]
       (perms/grant-application-permissions! group :monitoring)
-      (testing "Users with monitoring but no `manage-databasee` permission do not recieve"
+      (testing "Users with monitoring but no `manage-database` permission do not recieve"
         (mt/with-premium-features #{:advanced-permissions}
           (is (= (set (#'messages/all-admin-recipients))
                  (set (#'messages/admin-or-ee-monitoring-details-emails db-id))))))
--- a/enterprise/backend/test/metabase_enterprise/email/messages_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/email/messages_test.clj
@@ -2,6 +2,7 @@
   (:require
    [clojure.test :refer :all]
    [metabase.email.messages :as messages]
+   [metabase.models.data-permissions :as data-perms]
    [metabase.models.permissions :as perms]
    [metabase.test :as mt]
    [metabase.test.fixtures :as fixtures]))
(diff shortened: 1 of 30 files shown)


enterprise/backend/test/metabase_enterprise/email/messages_test.clj
```diff
@@ -13,8 +14,12 @@
     (mt/with-user-in-groups [group {:name "New Group"}
                              user [group]]
       (perms/grant-application-permissions! group :monitoring)
-      (perms/grant-permissions! group (perms/feature-perms-path :details :yes db-id))
-      (testing "Include monitoring and details user when advanced-permissions are enabled"
+      (testing "Users with monitoring but no `manage-databasee` permission do not recieve"
```
noahmoss, line 17:
```suggestion
      (testing "Users with monitoring but no `manage-database` permission do not recieve"
```

enterprise/backend/test/metabase_enterprise/sandbox/api/card_test.clj
```diff
@@ -3,68 +3,26 @@
...
-   [metabase.models :refer [Card Collection Database PermissionsGroup
-                            PermissionsGroupMembership Table]]
-   [metabase.models.permissions :as perms]
-   [metabase.models.permissions-group :as perms-group]
    [metabase.query-processor :as qp]
    [metabase.test :as mt]
-   [metabase.util :as u]
    [toucan2.tools.with-temp :as t2.with-temp]))
 
-(deftest users-with-segmented-perms-test
```
noahmoss, line 15: This test seems like it'd still be relevant, no?

test/metabase/models/permissions_group_test.clj
```diff
@@ -76,19 +74,24 @@
 (mu/defn ^:private group-has-full-access?
   "Does a group have permissions for `object` and *all* of its children?"
   [group-id :- ms/PositiveInt
-   object   :- perms.u/PathSchema]
+   db-id    :- ms/PositiveInt]
   ;; e.g. WHERE (object || '%') LIKE '/db/1000/'
```
noahmoss, line 78: nit - this comment is also out of date now

noahmoss (approved):

enterprise/backend/test/metabase_enterprise/sandbox/api/card_test.clj
johnswanson: Good point, I was thinking that:
- this test was previously making sure that someone with segmented permissions could still save or update a query for a card. There's no GTAP involved, just the segmented perm.
- now, "segmented permissions" don't exist - the user has `:unrestricted` data permissions
- so the test would just be asserting that a user with `:unrestricted` data permissions can save or update the query for a card, not really testing the sandbox at all.

But it would probably make more sense to add a GTAP for the group and then make the same assertions - I'll do that.

PR #38965, opened 2024-02-20
Add eslint rule to handle quotes

uladzimirdev:
### Description

Applied [quotes](https://eslint.org/docs/latest/rules/quotes) eslint rule to e2e folder to have strings consistent.

will not be affected
- string templates with variables inside
- multiline strings with `

### How to verify

tests should pass

### DIFF
--- a/e2e/.eslintrc
+++ b/e2e/.eslintrc
@@ -5,7 +5,8 @@
     "no-color-literals": 0,
     "no-console": 0,
     "@typescript-eslint/no-namespace": "off",
-    "cypress/no-async-tests": "error"
+    "cypress/no-async-tests": "error",
+    "quotes": [2, "double", { "avoidEscape": true }]
   },
   "env": {
     "cypress/globals": true,
--- a/e2e/.eslintrc
+++ b/e2e/.eslintrc
@@ -6,7 +6,7 @@
     "no-console": 0,
     "@typescript-eslint/no-namespace": "off",
     "cypress/no-async-tests": "error",
-    "quotes": [2, "double", { "avoidEscape": true }]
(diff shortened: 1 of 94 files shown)


.eslintrc
```diff
@@ -30,6 +30,7 @@
     ],
     "curly": [1, "all"],
     "eqeqeq": [1, "smart"],
+    "quotes": [2, "double", { "avoidEscape": true }],
```
oleggromov, line 33: ❤️

oleggromov:
This is a great addition, however, you'll have a bad time merging it and cause conflicts (although tiny) in a lot of branches.
Maybe, for the sake of merging it faster, let's remove the actual changes across the codebase and only merge the rule?

oleggromov (approved):

e2e/.eslintrc
```diff
@@ -5,7 +5,8 @@
     "no-color-literals": 0,
     "no-console": 0,
     "@typescript-eslint/no-namespace": "off",
-    "cypress/no-async-tests": "error"
+    "cypress/no-async-tests": "error",
+    "quotes": [2, "double", { "avoidEscape": true }]
```
WiNloSt, line 9: 1. I personally don't like using numbers since I am always confused about which number refers to which configuration. Do you think using string `"error"` would be more concrete?
2. Why don't we do this for the whole project too? i.e. put this setting in `.eslintrc` in the root directory instead. Why do we want to do this specifically only for `e2e` directory?

WiNloSt (approved):

oleggromov (approved):
Go for it!

oisincoveney (approved):
this makes me so happy 😄

Issue #38863, opened 2024-02-16
Add hover card column metadata on join column picker


romeovs:
As part of https://github.com/metabase/metabase/issues/38394, we want to surface more information on columns in the join column picker.

romeovs:
Closed by #38967

PR #38967, opened 2024-02-20
Join column hovercard

romeovs:
Closes https://github.com/metabase/metabase/issues/38863

### Description

Add the column info icon to the join column picker.

### How to verify

1. New question → Sample Dataset → Orders
2. Join data → Products
3. Click the column picker and verify each item in the dropdown has an info icon when hovered
4. Hover the info icon and verify it shows the correct information

### Demo

Screenshot 2024-02-20 at 16 13 58

### Checklist

### DIFF
--- a/frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
+++ b/frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
@@ -12,6 +12,14 @@ export const TriggerIcon = styled(Icon)`
   margin-left: 0.5rem;
 `;
 
+export const ChevronDown = styled(Icon)`
+  flex: 0 0 auto;
+  width: 8px;
+  margin-left: 0.25em;
+  color: currentColor;
+  opacity: 0.75;
+`;
+
 export const TriggerButton = styled.button<{ hasDot?: boolean }>`
   display: flex;
   align-items: center;
@@ -26,6 +34,10 @@ export const TriggerButton = styled.button<{ hasDot?: boolean }>`
 
   cursor: pointer;
 
+  ${ChevronDown} {
+    color: currentColor;
+  }
+
   &:hover {
     color: ${color("white")};
   }
@@ -38,15 +50,6 @@ export const Dot = styled.div`
   background: currentColor;
   border-radius: 100%;
   opacity: 0.25;
-`;
-
-export const ChevronDown = styled(Icon)`
-  flex: 0 0 auto;
-  width: 8px;
-  margin-left: 0.25em;
-  color: inherit;
-  opacity: 0.75;
-  color: currentColor !important;
 `;
 
 export const SelectListItem = styled(BaseSelectList.Item)<{
--- a/frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
+++ b/frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
@@ -16,7 +16,6 @@ export const TriggerButton = styled.button<{ hasDot?: boolean }>`
   display: flex;
   align-items: center;
   min-width: 0;
-  max-width: 50%;
 
   color: ${alpha(color("white"), 0.5)};
   font-weight: 700;
@@ -45,7 +44,9 @@ export const ChevronDown = styled(Icon)`
   flex: 0 0 auto;
   width: 8px;
   margin-left: 0.25em;
+  color: inherit;
   opacity: 0.75;
+  color: currentColor !important;
 `;
 
 export const SelectListItem = styled(BaseSelectList.Item)<{
--- a/frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
+++ b/frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
@@ -1,3 +1,4 @@
+import type { ReactNode } from "react";
 import { useCallback, useMemo } from "react";
 
 import {
@@ -5,13 +6,17 @@ import {
   getColumnGroupName,
 } from "metabase/common/utils/column-groups";
 import { getColumnIcon } from "metabase/common/utils/columns";
+import {
+  QueryColumnInfoIcon,
+  HoverParent,
+} from "metabase/components/MetadataInfo/ColumnInfoIcon";
 import type { ColorName } from "metabase/lib/colors/types";
 import type { IconName } from "metabase/ui";
-import { Icon } from "metabase/ui";
+import { Icon, DelayGroup } from "metabase/ui";
 import * as Lib from "metabase-lib";
 
 import { BucketPickerPopover } from "./BucketPickerPopover";
-import { StyledAccordionList } from "./QueryColumnPicker.styled";
+import { StyledAccordionList, ItemName } from "./QueryColumnPicker.styled";
 
 const DEFAULT_MAX_HEIGHT = 610;
 
@@ -27,6 +32,7 @@ export interface QueryColumnPickerProps {
   hasBinning?: boolean;
   hasTemporalBucketing?: boolean;
   withDefaultBucketing?: boolean;
+  withInfoIcons?: boolean;
   maxHeight?: number;
   color?: ColorName;
   checkIsColumnSelected: (item: ColumnListItem) => boolean;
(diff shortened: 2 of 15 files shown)


frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.unit.spec.tsx
```diff
@@ -94,6 +94,16 @@ describe("QueryColumnPicker", () => {
...
   });
 
+  it("should render info icons", () => {
+    const { sampleColumnInfo } = setup();
+
+    expect(
+      within(
+        screen.getByLabelText(sampleColumnInfo.displayName),
+      ).getByLabelText("More info"),
+    ).toBeInTheDocument();
```
kamilmielnik, line 104: We should also test the behavior on hover.

frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
```diff
@@ -45,7 +44,9 @@ export const ChevronDown = styled(Icon)`
   flex: 0 0 auto;
   width: 8px;
   margin-left: 0.25em;
+  color: inherit;
   opacity: 0.75;
+  color: currentColor !important;
```
kamilmielnik, line 49: 1. `color` is specified twice
2. Previously the icon got highlighted on hover with text, now it does not.

frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
```diff
@@ -16,7 +16,6 @@ export const TriggerButton = styled.button<{ hasDot?: boolean }>`
   display: flex;
   align-items: center;
   min-width: 0;
-  max-width: 50%;
```
kamilmielnik, line 19: :x: Long names are not handled correctly:
- [now](https://github.com/metabase/metabase/assets/6830683/cce611e0-a019-446c-80c0-879a94778f80)
- [before](https://github.com/metabase/metabase/assets/6830683/313c4431-7d38-4d81-8540-e680b8a92618)

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
```diff
@@ -149,33 +166,54 @@ export function QueryColumnPicker({
       color,
       checkIsColumnSelected,
       handleSelect,
+      withInfoIcons,
     ],
   );
 
+  const renderItemExtra = useCallback(
+    item => (
```
kamilmielnik, line 174: Design comment:

Up until now **selected** and **hovered** items had the same styling, i.e.
- background was blue
- the temporal bucket picker has been shown

To match the above, the info icon should be always shown for the **selected** item. But it's shown only on hover.

cc: @mazameli

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
```diff
@@ -149,33 +166,54 @@ export function QueryColumnPicker({
...
+      <QueryColumnInfoIcon
+        query={query}
+        stageIndex={stageIndex}
+        column={item.column}
+        position="top-start"
```
kamilmielnik, line 179: The name `position` does not make much sense in the context of `QueryColumnInfoIcon`. It applies to its popover, so it'd be better to call it `popoverPosition`.

kamilmielnik, line 179: Design comment:

It's a bit annoying that this popover is shown on top of other icons that can trigger it. If I'm going up the list with my mouse cursor, the popover will prevent me from inspecting other info icons.

https://github.com/metabase/metabase/assets/6830683/443f087f-89c3-48bc-892f-bc4cc6325ff8

Can we just show the popover on the right, so it does not cover anything, [like this](https://github.com/metabase/metabase/assets/6830683/0c30a32e-0562-419b-a9c0-a9293c774520)?

cc: @mazameli

romeovs: Yeah the tricky thing is that, on small screens if the popover does not fit on the right of the icon, it will be flipped to the left and overlap on top of the item which is not what we want. 

I'll try to update the default positioning logic to avoid this.

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.unit.spec.tsx
romeovs: That's already tested in the `ColumnInfoIcon` unit tests, do we need to retest it here?

kamilmielnik: > That's already tested in the `ColumnInfoIcon` unit tests, do we need to retest it here?

Great! I think there's no need to retest it here :+1:

frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
romeovs: I've made some changes to the `AccordionList` component to accommodate this.
I done some spot-checks to make sure this didn't blow anything up elsewhere and I think it does not.

But the `AccordionList` component is used in quite a few places, so the possible blast radius from messing stuff up here is quite big. I'm happy to scope the changes to this component if that is considered to be more desirable.

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
mazameli: To the right is ideal, but above is second-best, given that I think it's a reasonable assumption that slightly more of the time if you're looking at the descriptions of multiple columns, you'd start from the top and work your way down.

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
```diff
@@ -1,17 +1,26 @@
+import type { ReactNode } from "react";
```
kamilmielnik, line 1: nit: 2 popovers are sometimes shown simultaneously

https://github.com/metabase/metabase/assets/6830683/ba2982b3-fc19-476d-be8a-0830d2272c2f

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
```diff
@@ -1,17 +1,26 @@
...
 import { useCallback, useMemo } from "react";
 
 import {
   getColumnGroupIcon,
   getColumnGroupName,
 } from "metabase/common/utils/column-groups";
 import { getColumnIcon } from "metabase/common/utils/columns";
+import {
```
kamilmielnik, line 9: nit (a11y): it's not possible to select and copy the text in the popover because it disappears when I press and hold left mouse button



https://github.com/metabase/metabase/assets/6830683/1106b24b-7f4d-4cc1-b580-e370e1fe499f

frontend/src/metabase/common/components/QueryColumnPicker/BucketPickerPopover/BaseBucketPickerPopover.styled.tsx
kamilmielnik: Definitely looks better :+1: 

I think we can improve layout in breakouts dropdown: the labels for binning seem to take more place than the column name, yet it's the column name that is more important IMO.

Screenshot using Polish translations:

image

kamilmielnik (approved):
Code looks good :+1: 

Left 2 minor comments (apologies if they should have been submitted in previous PRs).
It'd be great if my [previous design comment](https://github.com/metabase/metabase/pull/38967#discussion_r1498784864) was addressed.

frontend/src/metabase/common/components/QueryColumnPicker/QueryColumnPicker.tsx
romeovs: This applies to all popovers (not just the one added in this PR), I'll add it to the product doc and we can review it during the design review.

romeovs: As per @mazameli's [comments on Slack](https://metaboat.slack.com/archives/C0645JP1W81/p1709136662459169?thread_ts=1708701338.803019&cid=C0645JP1W81), the current implementation is what we want to go with

PR #38970, opened 2024-02-20
Faster sync-fks

calherries:
This is part 1 towards resolving https://github.com/metabase/metabase/issues/38492
Part 2 is https://github.com/metabase/metabase/pull/38828

This PR adds an alternative implementation of the `sync-fks` step of sync where we sync the foreign key information of a database.

It adds a new driver feature: `describe-fks`. If a driver supports this feature, they need to implement the new driver method `metabase.driver/describe-fks`. If the driver doesn't support the feature, we'll continue to use `metabase.driver/describe-table-fks`, which has been deprecated. The plan for driver authors to implement `describe-fks` in terms of `driver/describe-table-fks`, which should be a really simple change for most drivers, assuming it doesn't need to be performant for large DBs.

The way this works is instead of querying the customer DB one table at a time using the `getImportedKeys` JDBC method, we execute a single query that gets all the data at once with paginated results. We then reduce over those results (yay transducers), updating our App DB one foreign key at a time.

Only redshift is implemented right now but its implementation is 22 LOC, so more drivers can be added easily.

I tested with the redshift dev instance using a local postgres app DB with a schema with 10 foreign keys over 20 tables.
```clojure
(mt/with-driver :redshift
  (mt/with-temp-test-data
    (let [n-foreign-keys 10]
      (mapcat
       (fn [i]
         [[(str "continent_" i)
           [{:field-name "name", :base-type :type/Text}]
           [["ok"]]]
          [(str "country_" i)
           [{:field-name "name", :base-type :type/Text}
            {:field-name "continent_id", :base-type :type/Integer :fk (str "continent_" i)}]
           [["ok" 1]]]])
       (range n-foreign-keys)))))
```
I got these times from the sync logs:

Before: 11.5 seconds 🐌 
After: 867.7 ms 🚀 

I don't want to test any bigger DBs yet because they take a long time to sync their fields. The real perf tests will come with https://github.com/metabase/metabase/pull/38828

Low-level test coverage could probably be improved but there's so much that depends on sync I think it's pretty safe without adding more tests.


crisptrutski:
High level approach looks good, but will wait for the test to get fixed before taking a closer look.

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -714,6 +712,29 @@
...
+  [db [sql & params]]
+  (let [driver (:engine db)]
+    (reify clojure.lang.IReduceInit
+      (reduce [_ rf init]
+        (do-with-connection-with-options
+         driver
+         db
+         nil
+         (fn [^Connection conn]
+           (with-open [stmt          (statement-or-prepared-statement driver conn sql params nil)
```
calherries, line 726: Note this adds a `(.setFetchSize stmt (sql-jdbc-fetch-size))` which is how the pagination is implemented

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -49,6 +51,30 @@
   [& args]
   (apply (get-method driver/describe-table :sql-jdbc) args))
 
+(defmethod sql-jdbc.sync/describe-fks-sql :redshift
```
calherries, line 54: See the implementation of redshift's JDBC driver for reference
https://github.com/aws/amazon-redshift-jdbc-driver/blob/master/src/main/java/com/amazon/redshift/jdbc/RedshiftDatabaseMetaData.java#L3974

src/metabase/driver.clj
```diff
@@ -348,6 +348,21 @@
...
 
+(defmulti describe-fks
+  "Returns a reducible collection of maps, each containing information about foreign keys.
+  Takes keyword arguments to narrow down the results to a set of `schema-names` or `table-names`.
+
+  Results match [[metabase.sync.interface/FastFKMetadataEntry]].
+  Results are optionally filtered by `schema-names` and `table-names` provided.
+
+  Required for drivers that support `:fast-sync-fks`."
+  {:added "0.50.0" :arglists '([driver database & {:keys [schema-names table-names]}])}
```
calherries, line 359: `table-names` isn't used for redshift yet, but could be handy to have as an extension in the future. Then when we sync a single table we can reuse this

calherries, line 359: Because this takes `table-names` as an argument, we can use this instead of `describe-table-fks`. That means `describe-table-fks` can be deprecated in the future.

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -36,6 +37,7 @@
 
 (doseq [[feature supported?] {:test/jvm-timezone-setting false
                               :nested-field-columns      false
+                              :describe-fks             true
```
qnkhuat, line 40:
```suggestion
                              :describe-fks              true
```

docs/developers-guide/driver-changelog.md
```diff
@@ -4,6 +4,18 @@ title: Driver interface changelog
 
 # Driver Interface Changelog
 
+## Metabase 0.50.0
+
+- The multimethod `metabase.driver/describe-table-fks` has been deprecated in favor of `metabase.driver/describe-fks`.
+  `metabase.driver/describe-table-fks` will be removed in 0.53.0.
```
qnkhuat, line 10: > 0.53.0
are you sure? I'd put in future releases if we didn't have a plan in place for 53.

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -49,6 +51,31 @@
...
+                         :fk_column.attname   "fk-column-name"
+                         :pk_ns.nspname       "pk-table-schema"
+                         :pk_column.attname   "pk-column-name"
+                         :pk_table.relname    "pk-table-name"})
+               :from   [[:pg_constraint :c]]
+               :join   [[:pg_namespace :fk_ns]     [:= :c.connamespace :fk_ns.oid]
+                        [:pg_class     :fk_table]  [:= :c.conrelid :fk_table.oid]
+                        [:pg_attribute :fk_column] [:= :c.conrelid :fk_column.attrelid]
+                        [:pg_class     :pk_table]  [:= :c.confrelid :pk_table.oid]
+                        [:pg_namespace :pk_ns]     [:= :pk_table.relnamespace :pk_ns.oid]
```
qnkhuat, line 68:
```suggestion
                        [:pg_namespace :pk_ns]     [:= :pk_table.relnamespace :pk_ns.oid]
                        [:pg_class     :pk_table]  [:= :c.confrelid :pk_table.oid]
```

Make it easier to read since it's the same order with join for fk_*

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -49,6 +51,31 @@
...
 
+(defmethod sql-jdbc.sync/describe-fks-sql :redshift
+  [driver & {:keys [schema-names table-names]}]
+  (sql/format {:select (vec
+                        {:fk_ns.nspname       "fk-table-schema"
+                         :fk_table.relname    "fk-table-name"
+                         :fk_column.attname   "fk-column-name"
+                         :pk_ns.nspname       "pk-table-schema"
+                         :pk_column.attname   "pk-column-name"
+                         :pk_table.relname    "pk-table-name"})
```
qnkhuat, line 62:
```suggestion
                         :pk_table.relname    "pk-table-name"
                         :pk_column.attname   "pk-column-name"})
```

move table up to keep the order consistently with fk

src/metabase/sync/fetch_metadata.clj
```diff
@@ -20,13 +22,31 @@
...
+(mu/defn fk-metadata
+  "Effectively a wrapper for [[metabase.driver/describe-fks]] that also validates the output against the schema."
+  [database :- i/DatabaseInstance & {:as args}]
+  (cond->> (driver/describe-fks (driver.u/database->driver database) database args)
+    ;; Validate the output against the schema, except in prod.
+    ;; This is a workaround for the fact that [[mu/defn]] can't check reducible collections yet
+    (not config/is-prod?)
+    (eduction (map #(mu.fn/validate-output {} i/FKMetadataEntry %)))))
+
+(mu/defn table-fk-metadata :- [:maybe [:sequential i/FKMetadataEntry]]
```
qnkhuat, line 34: I find the need for this function unnecessary, we can use `fk-metadata` with a `table-names` arg for this use case right?

src/metabase/driver/sql_jdbc/sync/describe_table.clj
```diff
@@ -264,14 +264,25 @@
...
-     (fn [^Connection conn]
-       (describe-table-fks* driver conn table db-name-or-nil)))))
+  (sql-jdbc.execute/do-with-connection-with-options
+   driver
+   db-or-id-or-spec-or-conn
+   nil
+   (fn [^Connection conn]
+     (describe-table-fks* driver conn table db-name-or-nil))))
+
+(defmulti describe-fks-sql
```
qnkhuat, line 274: You should mention in this ns's docstraing that `describe-table-fks` is deprecated and we should implement `describe-fks-sql` instead.

src/metabase/sync/fetch_metadata.clj
qnkhuat: ah I just realized that initially `fk-metadata` is per-table and now you changed it to per-db.

But I think it's cleaner to makes `fk-metadata` deals with the fact that there is 2 ways to get table metadata.

I think you can make `fk-metadata` smarter and remove this function.

```clojure
(mu/defn fk-metadata
  "Effectively a wrapper for [[metabase.driver/describe-fks]] that also validates the output against the schema."
  [database :- i/DatabaseInstance & {:keys [table-names] :as args}]
    (cond->> (if (driver/database-supports? driver :describe-fks database)
               (driver/describe-fks (driver.u/database->driver database) database args)
               (flatten (for [table (or table-names (sync.u/db->sync-tables database))]
                          (vec (for [x (driver/describe-table-fks driver database table)]
                                 {:fk-table-name   (:name table)
                                  :fk-table-schema (:schema table)
                                  :fk-column-name  (:fk-column-name x)
                                  :pk-table-name   (:name (:dest-table x))
                                  :pk-table-schema (:schema (:dest-table x))
                                  :pk-column-name  (:dest-column-name x)})))))
      ;; Validate the output against the schema, except in prod.
      ;; This is a workaround for the fact that [[mu/defn]] can't check reducible collections yet
      (not config/is-prod?)
      (eduction (map #(mu.fn/validate-output {} i/FKMetadataEntry %))))))
```
For case where only fk-metadata for one table is needed, you need to pass in the name of the tables, otherwise it'll return all tables within that database.

This way, `sync-fks!` doesn't need to do the feature check at all.

modules/drivers/redshift/src/metabase/driver/redshift.clj
```diff
@@ -49,6 +51,31 @@
   [& args]
   (apply (get-method driver/describe-table :sql-jdbc) args))
 
+(defmethod sql-jdbc.sync/describe-fks-sql :redshift
```
qnkhuat, line 54: We need a test to check if this query filters table name and schema name correctly. Do we need to escape the names?

src/metabase/sync/sync_metadata/fks.clj
```diff
@@ -77,26 +78,41 @@
...
+                                                                :schema (:fk-table-schema metadata))
+                              (sync-util/field-name-for-logging :name (:pk-column-name metadata)))))))
+
+(mu/defn sync-fks-for-db!
+  "Sync the foreign keys for a `database`."
+  [database :- i/DatabaseInstance]
+  (sync-util/with-error-handling (format "Error syncing FKs for %s" (sync-util/name-for-logging database))
+    (let [schema-names (sync-util/db->sync-schemas database)
+          fk-metadata  (fetch-metadata/fk-metadata database :schema-names schema-names)]
+      (transduce (map (fn [x]
```
qnkhuat, line 103: we need to call `sync-util/set-initial-table-sync-complete` here too

src/metabase/sync/sync_metadata/fks.clj
```diff
@@ -106,27 +122,32 @@
...
+                           (update update-info :total-failed inc)
+                           (merge-with + update-info table-fk-info))))
+                     {:total-fks    0
+                      :updated-fks  0
+                      :total-failed 0}
+                     (sync-util/db->reducible-sync-tables database)))
+    ;; Mark the table as done with its initial sync once this step is done even if it failed, because only
+    ;; sync-aborting errors should be surfaced to the UI (see
+    ;; `:metabase.sync.util/exception-classes-not-to-retry`).
+    (sync-util/set-initial-table-sync-complete-for-db! database)))
```
qnkhuat, line 153: It's not related to this PR, but I'm surprised to see this is done in `sync-fks!` as it has nothing to do with fk. maybe it should be its own sync step.

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -714,6 +712,29 @@
...
+                       ^ResultSet rs (try
+                                       (let [max-rows 0] ; 0 means no limit
```
qnkhuat, line 728: > ; 0 means no limit

is this from jdbc or it's driver-specific?

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -714,6 +712,29 @@
               results-metadata {:cols (column-metadata driver rsmeta)}]
           (respond results-metadata (reducible-rows driver rs rsmeta qp.pipeline/*canceled-chan*))))))))
 
+(defn sql->reducible-rows
```
qnkhuat, line 715: how is this different from `jdbc/reducible-query`?

src/metabase/driver/sql_jdbc/sync/describe_table.clj
```diff
@@ -264,14 +264,25 @@
...
-    (describe-table-fks* driver db-or-id-or-spec-or-conn table db-name-or-nil)
-    (sql-jdbc.execute/do-with-connection-with-options
-     driver
-     db-or-id-or-spec-or-conn
-     nil
-     (fn [^Connection conn]
-       (describe-table-fks* driver conn table db-name-or-nil)))))
+  (sql-jdbc.execute/do-with-connection-with-options
+   driver
+   db-or-id-or-spec-or-conn
```
qnkhuat, line 269: `do-with-connection-with-options` does not take `conn` as an argument.

this will fail

```clojure
(let [db (toucan2.core/select-one :model/Database :engine :h2)]
  (sql-jdbc.execute/do-with-connection-with-options
   :h2
   db
   nil
   (fn [conn1]
     (sql-jdbc.execute/do-with-connection-with-options
      :h2
      conn1
      nil
      (fn [conn2]
        (jdbc/query {:connection conn2} "select 1;"))))))
```

But I doubt we use `defn describe-table-fks` with a connection anywhere. Still, we need to check to make sure we don't create regression here.

src/metabase/sync/util.clj
```diff
@@ -303,6 +303,17 @@
   (when (not= (:initial_sync_status table) "complete")
     (t2/update! :model/Table (u/the-id table) {:initial_sync_status "complete"})))
 
+(def ^:private sync-tables-kv-args
+  {:active          true
+   :visibility_type nil})
+
+(defn set-initial-table-sync-complete-for-db!
+  "Marks initial sync for all tables in `db` as complete so that it becomes usable in the UI, if not already
+  set"
```
qnkhuat, line 312:
```suggestion
  set."
```
docstring need to be a complete sentence

qnkhuat (changes requested):
Generally, I think this is good, but I still have some suggestions that need resolving.

Also CI is failing for redshift, not sure if it's related to this PR, looks like a flake tho.

docs/developers-guide/driver-changelog.md
calherries: Good question. Happy discuss whether or not to remove it, but if we decide to remove it I believe we should commit to a release, rather than drawing it out. Breaking changes are bad but it's better if we do them earlier rather than later.

Here's how we could maintain backwards compatibility:
- provide a default implementation for `describe-fks` in terms of `describe-table-fks`. The implementation will select tables from the app DB in a reducible collectoin, and reduce over them, using `describe-table-fks` instead to select the FK metadata from the DB.
- use the new default implementation instead of `describe-table-fks`, unless the driver explicitly supports the `describe-fks` feature

Here's why breaking change wouldn't be too much of a hassle for driver authors:
- they could simply copy+paste the default implementation to implement `describe-fks` instead of implementing `describe-table-fks`.

In all, the pros and cons of breaking backwards compatibility aren't all that large either way. The benefit is that the interface for drivers is smaller (one less driver method and feature), so new drivers are slightly easier to implement. The downside is that all existing drivers have to copy the default implementation of `describe-fks`.

We could defer this decision until I've written the above default implementation, and we can assess the pros vs cons.

So, how do you feel about each option, between:
1. maintain backwards compatibility forever, keeping both `describe-fks` and `describe-table-fks`, but one needs to be implemented
2. deprecate in 53
3. defer this decision until the default implementation is used instead. Just say it's deprecated and may be removed

src/metabase/sync/fetch_metadata.clj
calherries: Seems like a nice idea, the problem is `fk-metadata` needs to take `:schema-names`, and I don't want to complicate this function further by implementing support for that if we don't need it yet.

But this kind of refactor will be possible in the future if we implement `driver/describe-fks` in terms of `driver/describe-table-fks`, which I'm planning to do.

modules/drivers/redshift/src/metabase/driver/redshift.clj
calherries: Name escaping is handled by this being a honeysql query.

src/metabase/driver/sql_jdbc/execute.clj
calherries: JDBC

calherries:
@qnkhuat can you take another look? I've resolved all your suggestions with a 👍  in the last commit: [8172cbd](https://github.com/metabase/metabase/pull/38970/commits/8172cbddaaadc53574b393df1d9e505b0c4538d4)

Indeed redshift looks like a flake.

src/metabase/driver/sql_jdbc/execute.clj
calherries: I've added a docstring and renamed the function to make the purpose a little clearer, and also nod towards it being similar to `jdbc/reducible-query`
https://github.com/metabase/metabase/blob/2d5975e9ea8d3d1ee7af755be34d9340cc06fbc5/src/metabase/driver/sql_jdbc/execute.clj#L717-L720

calherries: I'm not sure about the name `simple-reducible-query`, but I wanted to distinguish it as the simpler version of `execute-reducible-query` in the same namespace.

qnkhuat: why not call it `reducible-query`? 

and I would mention the difference with `jdbc/reducible-query` in the docstring, I'm sure people will have the same question as I did.

qnkhuat (approved):
LGTM!

src/metabase/driver/sql_jdbc/execute.clj
calherries: > why not call it reducible-query?

I wanted to distinguish it as the simpler version of `execute-reducible-query` in the same namespace. That's not so obvious though. I will accept your suggestion :)

> and I would mention the difference with jdbc/reducible-query in the docstring, I'm sure people will have the same question as I did.

I have included this already, though not in such detail:

https://github.com/metabase/metabase/blob/2d5975e9ea8d3d1ee7af755be34d9340cc06fbc5/src/metabase/driver/sql_jdbc/execute.clj#L718-L719

modules/drivers/redshift/src/metabase/driver/redshift.clj
calherries: I initially tried this reordering but had to reverse the order again because this join order is needed. But I've made it consistent at least with fk now in [reorder joins](https://github.com/metabase/metabase/pull/38970/commits/4702a7fdd3fbef98f0dcbcc0ca42bef00269d0ad)

docs/developers-guide/driver-changelog.md
calherries: We currently don't use milestones for this, I wonder what the current process is. I've asked here https://metaboat.slack.com/archives/CKZEMT1MJ/p1710244644500109

PR #38972, opened 2024-02-20
Use batched hydration of fields for tables

noahmoss:
See investigation here: https://metaboat.slack.com/archives/C064EB1UE5P/p1708387795469919?thread_ts=1708358454.946419&cid=C064EB1UE5P

Essentially, we were issuing a separate DB call for every table when hydrating fields instead of doing batch hydration with a single DB call.

Not sure how to test this since the behavior should be the same. Maybe existing tests passing + checking that this solves the slow page on stats is enough validation.


noahmoss:
* **#38972** <a href="https://app.graphite.dev/github/pr/metabase/metabase/38972?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/38972?utm_source=stack-comment).

qnkhuat (approved):

PR #38976, opened 2024-02-20
Trend chart tweaks

kulyk:
Related to [#33411](https://github.com/metabase/metabase/issues/33411)

Implements several UX and visual tweaks to trend charts:

- when adding a new comparison, we'd automatically open the comparison picker ([Auto open the comparison picker](https://github.com/metabase/metabase/commit/f984e1412e494ef58fb3c4b2685d8a2aeb97f39a))
- when selecting the "Value from another column..." option, we'd preselect the value if there's just one available column ([`1b19d03`](https://github.com/metabase/metabase/commit/1b19d03234f9db63b06c76d46dc0f70cdf6ba2c1))
- swaps "Label" and "Value" inputs on the "Custom value" form for consistency with other forms ([`1d2fe23`](https://github.com/metabase/metabase/commit/1d2fe23c6577f1b6d75f8ef864419e110d6ac398))
- fix a custom value comparison's label case wasn't respected by the chart (a label set to something like "my GOAL" was displayed like "my goal") ([`9ceb4e5`](https://github.com/metabase/metabase/commit/9ceb4e542c6d51949f5d1a1d25478400cfeb307b))


cdeweyx (approved):

frontend/src/metabase/visualizations/visualizations/SmartScalar/SettingsComponents/AnotherColumnForm.tsx
```diff
@@ -96,3 +98,29 @@ export function AnotherColumnForm({
...
+  if (value?.column) {
+    return value.column;
+  }
+  if (columns.length === 1) {
+    return columns[0].name;
+  }
+  return "";
+}
+
+function getInitialLabel(
```
EmmadUsmani, line 115: Nit: it would be nice to combine these two functions into one to keep the logic together instead of decoupled, for example:

```typescript
  const { initialColumn, initialLabel } = getInitialColumnAndLabel(
    selectedValue,
    columns,
  );

  const [label, setLabel] = useState(initialColumn);
  const [column, setColumn] = useState(initialLabel);
  
  ...
  
  function getInitialColumnAndLabel(
  value: SmartScalarComparisonAnotherColumn | undefined,
  columns: DatasetColumn[],
) {
  if (value?.column) {
    return { initialColumn: value.column, initialLabel: value.label };
  }
  if (columns.length === 1) {
    return {
      initialColumn: columns[0].name,
      initialLabel: columns[0].display_name,
    };
  }
  return { initialColumn: "", initialLabel: "" };
}

  
  ```

EmmadUsmani (approved):
Looks good  👍

Issue #30218, opened 2023-04-19
Param that link to inactive fieldvalues will not create on demand


qnkhuat:
We have 2 types of FieldValues: full vs advanced(check out the docs in `metabase.models.field-values` for more details).

but the TLDR is that advanced field values is considered on-demand, and full fieldvalues are created during sync.

Before [Disable field value scanning for inactive fields](https://github.com/metabase/metabase/pull/26974), a field that is qualified to have FieldValues will be guaranteed to have a fieldvalues, but now it's no longer true cause a field could be inactive and we don't create fieldvalues for them.

So we should update the getter for fieldvalues to handle `get-or-create` instead.

Fix the following test should resolve the issue.

```clojure
(deftest dashboard-param-link-to-a-field-without-full-field-values
  (mt/with-temp* ['Dashboard     [{dashboard-id :id} {:name "Test Dashboard"}]
                  'Card          [{card-id :id}      {:name "Dashboard Test Card"}]
                  'DashboardCard [{_ :id}            {:dashboard_id       dashboard-id
                                                      :card_id            card-id
                                                      :parameter_mappings [{:card_id      card-id
                                                                            :parameter_id "foo"
                                                                            :target       [:dimension
                                                                                           [:field (mt/id :venues :name) nil]]}]}]]
    ;; delete the full fieldvalues, this could be because the field is inactive and we stopped syncing it
    (t2/delete! 'FieldValues :field_id (mt/id :venues :name) :type :full)
    (is (= {(mt/id :venues :name) {:values                ["20th Century Cafe"
                                                           "25°"
                                                           "33 Taps"]
                                   :field_id              (mt/id :venues :name)
                                   :human_readable_values []}
            (let [response (:param_values (mt/user-http-request :rasta :get 200 (str "dashboard/" dashboard-id)))]
                 (into {} (for [[field-id m] response]
                            [field-id (update m :values (partial take 3))])))}))))
```

PR #38980, opened 2024-02-20
Do not create param_values on Dashboard Load, only fetch if they exist

### LINKED ISSUE #38826: We should never run get-field-values on a dashboard load if the field does not have field values in the metabase_fieldvalues table
### Describe the bug

A customer with a massive database took down many dashboards as the metabase_fieldvalues was wiped by the time they did an upgrade, so Metabase started doing field-values on the DB on production time and several dashboards didn't load at all for the following reason: if a filter is a dropdown and the metabase_fieldvalues entry does not exist for that filter, we should simply return an empty array, not fire a query to the DB to get those

### To Reproduce

1) set up Metabase from scratch
2) create a question of the people table (just a raw table)
3) add a filter connected to the source field
4) go to the app db and wipe the field values from the metabase_fieldvalues for the source field (it's the one with Google, Facebook... fields)
5) load the dashboard, see Metabase running the `SELECT "public"."people"."source" AS "source" FROM "public"."people" GROUP BY "public"."people"."source" ORDER BY "public"."people"."source" ASC LIMIT 1000` query

this query can be extremely hard on some dwh and we're taking those dashboards (and probably the DWH) down due to this

### Expected behavior

If the field is not there just simply return an empty array, don't try to do that on a dashboard run

### Logs

NA

```JSON
v48.6, but probably been like that forever?
```

### Severity

P1

### Additional context

_No response_

adam-james-v:
Fixes: #38826

This change removes the *-or-create* behaviour of `field-ids->param-field-values`. This means that on Dashboard Load, if the param values already exist, they'll be grabbed. But, if they don't, we won't try to fetch or create them.

The changes introduced while fixing #30218, via #32299 seem to be the source of this and actually added 2 tests that assert behaviour we don't want so I've removed those 2 tests.

The test I've added does check that the `GET /api/dashboard/:id/params/:param_id/values` endpoint does indeed result in Field Value creation, so we're still able to properly load the values, just at time of Filter usage instead.

```clojure
(deftest param-values-not-fetched-on-load-test
  (testing "Param values are not needed on initial dashboard load and should not be fetched. #38826"
    (t2.with-temp/with-temp
        [:model/Dashboard {dashboard-id :id} {:name       "Test Dashboard"
                                              :parameters [{:name      "FOO"
                                                            :id        "foo"
                                                            :type      :string/=
                                                            :sectionId "string"}]}
         :model/Card {card-id :id} {:name "Dashboard Test Card"}
         :model/DashboardCard _ {:dashboard_id       dashboard-id
                                 :card_id            card-id
                                 :parameter_mappings [{:card_id      card-id
                                                       :parameter_id "foo"
                                                       :target       [:dimension
                                                                      [:field (mt/id :venues :name) nil]]}]}]
        (let [dashboard-load-a (mt/user-http-request :rasta :get 200 (str "dashboard/" dashboard-id))
              _                (t2/delete! :model/FieldValues :field_id (mt/id :venues :name) :type :full)
              dashboard-load-b (mt/user-http-request :rasta :get 200 (str "dashboard/" dashboard-id))
              param-values     (mt/user-http-request :rasta :get 200 (format "dashboard/%s/params/%s/values" dashboard-id "foo"))]
          (testing "The initial dashboard-load doesn't fetch parameter values (#38826)"
            (is (some? (:param_values dashboard-load-a)))
            (is (nil? (:param_values dashboard-load-b))))
          (testing "Request to values endpoint triggers computation of field values if missing."
            (is (= ["20th Century Cafe" "25°" "33 Taps"]
                   (take 3 (apply concat (:values param-values))))))))))
```

### DIFF
--- a/src/metabase/models/params.clj
+++ b/src/metabase/models/params.clj
@@ -88,7 +88,7 @@
    (into {}
          (map (comp (juxt :field_id identity)
                     #(select-keys % [:field_id :human_readable_values :values])
-                    field-values/get-or-create-full-field-values!))
+                    #(field-values/get-latest-full-field-values (:id %))))
          (t2/hydrate (t2/select :model/Field :id [:in (set param-field-ids)]) :values))))
 
 (defn- field-ids->param-field-values
@@ -186,7 +186,7 @@
 
 
 (mu/defn ^:private param-field-ids->fields
-  "Get the Fields (as a map of Field ID -> Field) that shoudl be returned for hydrated `:param_fields` for a Card or
+  "Get the Fields (as a map of Field ID -> Field) that shoudld be returned for hydrated `:param_fields` for a Card or
   Dashboard. These only contain the minimal amount of information necessary needed to power public or embedded
   parameter widgets."
   [field-ids :- [:maybe [:set ms/PositiveInt]]]
@@ -204,7 +204,7 @@
 #_{:clj-kondo/ignore [:unused-private-var]}
 (mi/define-simple-hydration-method ^:private hydrate-param-values
   :param_values
-  "Hydration method for `:param_fields`."
+  "Hydration method for `:param_values`."
   [instance]
   (param-values instance))
 
--- a/src/metabase/models/params.clj
+++ b/src/metabase/models/params.clj
@@ -186,7 +186,7 @@
 
 
 (mu/defn ^:private param-field-ids->fields
-  "Get the Fields (as a map of Field ID -> Field) that shoudld be returned for hydrated `:param_fields` for a Card or
+  "Get the Fields (as a map of Field ID -> Field) that should be returned for hydrated `:param_fields` for a Card or
   Dashboard. These only contain the minimal amount of information necessary needed to power public or embedded
   parameter widgets."
   [field-ids :- [:maybe [:set ms/PositiveInt]]]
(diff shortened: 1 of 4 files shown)


src/metabase/models/params.clj
```diff
@@ -186,7 +186,7 @@
 
 
 (mu/defn ^:private param-field-ids->fields
```
markbastian, line 188: Typo?

markbastian (approved):
See comment on type, but otherwise looks good.

src/metabase/models/params.clj
adam-james-v: 😆I didn't like the typo originally so I changed it up. Typo remix.
(Good catch, I'll fix that)

src/metabase/models/params.clj
```diff
@@ -186,7 +186,7 @@
...
-  "Get the Fields (as a map of Field ID -> Field) that shoudl be returned for hydrated `:param_fields` for a Card or
+  "Get the Fields (as a map of Field ID -> Field) that shoudld be returned for hydrated `:param_fields` for a Card or
```
qnkhuat, line 189:
```suggestion
  "Get the Fields (as a map of Field ID -> Field) that should be returned for hydrated `:param_fields` for a Card or
```

qnkhuat (approved):
The reason I created https://github.com/metabase/metabase/issues/30218 was to make sure if you have a parameter that maps to a field with remapped value, and if you set the default value for it, on dashboard we want to show in the widget the "remapped value". 

Without fetching from DB, and field values is wiped ( because of inactive) the dashboard will not show the "remapped value".

it's a subtle use case. 

But on further check, the remapped value now has been handled separately by calling `/api/field/12/remapping/9?value=..` so this PR makes sense.

It also makes me wonder whether we still need `param_values` :), my guess is we don't.

PR #38985, opened 2024-02-20
Kondo hook to check for blank lines in deftest

camsaul:
Doesn't work if there are comments so I need to figure out what to do about that.


camsaul:
this will have to be done with `cljfmt` since Kondo doesn't tell you info about comment forms

PR #38986, opened 2024-02-20
Persist last selected database in the native query editor for a user

### LINKED ISSUE #8780: Make new native queries default to using the last-selected database
Opening this on behalf of a user who [proposed this](https://twitter.com/dotpem/status/1055586084246056960). I think the behavior would simply be that when you click on new question -> native/sql, that it would default to whichever database you last selected in that database dropdown. This would remove a click if you tend to query the same database most of the time. 

### LINKED ISSUE #39035: [Testing Plan] Make new native queries default to using the last-selected database
For issue: #8780

## Dimensions
- User
    - Id (this should be a user-local setting) 
    - Permissions
        - With native write permissions
        - Without native write permissions
    - Session (should persist after logging out and logging in again)
- Database type
    - Requires an additional table to run native query
    - Runs native query without a table
- Where native query editor is instantiated from
    - SQL query
    - Model
    - Action
- Does database still exist?
    - Yes
    - No

~~- Is there a query, was query run and was it saved?~~
    - This bit is irrelevant because the action is fired upon selecting a database and has nothing to do with actually running a query

## Actions
- Starting new (SQL Question | Model | Action)
- Database gets deleted after it's persisted
- User logs out, logs in again
- User logs out, some other user logs in in the same browser
- ~~Editing previously saved (SQL Question | Model | Action)~~

### Unit tests
_Will implement only backend unit tests._
- Make sure the setting can be set
- Make sure the setting can be read
- Make sure the setting is user-local only

### Integration tests
_Not going to implement these. They will require insane effort and will be almost as close as an E2E test._

### E2E tests
(Prerequisites: _set up postgres and mongo as additional databases, have at least one additional user in All users group_)
- As admin, open new native question
    - Choose a database (let's say Sample Database)
    - Navigate away from that page
    - Open new native question again
    - Make sure previously selected database is pre-selected
    - Log out
    - Log in as normal user
    - Open new native question
    - Make sure you are offered a choice of a database selection
    - Choose a different database (Mongo, for example)
    - Navigate away
    - Start new native question again
    - Make sure Mongo is the pre-selected database, and make sure the table picker is rendered
    - Log out
    - Log in as admin again
    - Start new native question
    - Make sure Sample Database is still selected for you, and not Mongo

- As admin, open new native question
    - Choose a database (let's say Postgres)
    - Go to admin settings and delete Postgres
    - Start new native question again
    - Make sure nothing breaks, and the UI prompts you to select database (Sample or Mongo)

- Make sure no redundant requests are fired when selecting the same database again

- Make sure this setting does not apply to model actions
    - Persisting a database for a native model or a native question should not influence model actions, and vice versa

- Make sure users without sufficient permissions cannot choose or persist databases they cannot query against

nemanjaglumac:
## Info
|||
|-|-|
| 📝 | [Product Doc](https://www.notion.so/metabase/Default-to-the-last-selected-database-in-native-queries-3bf8f844a6e8407a998bbf11a04f106b) |
| 🧪 | [Testing Plan](https://github.com/metabase/metabase/issues/39035) |

## What does PR accomplish?
This PR resolves #8780 by introducing a user-level setting for remembering the last used database in the native query.
Closes #39035.

```[tasklist]
### Tasks
- [x] Add new user-level setting to the backend
- [x] Wire native query editor to remember the last used database
- [x] Narrow the scope to just native questions and native models
- [x] Write unit tests
- [x] Write E2E tests
```

### About the `last-used-database` setting
- It should be user-local only
- It should work for any authenticated user; frontend ensures that a user without sufficient native permissions is not even offered to select and persist the last used database
- The information about the last used database currently should not be included in the audit information

> [!Note]
> The backend currently does have any additional checks to make sure the previously used database is still present. Frontend simply falls back to no database selected and prompts the user to select a new one.

### DIFF
--- a/e2e/test/scenarios/native/native-database-source.cy.spec.js
+++ b/e2e/test/scenarios/native/native-database-source.cy.spec.js
@@ -1,3 +1,4 @@
+import { SAMPLE_DB_ID } from "e2e/support/cypress_data";
 import {
   restore,
   popover,
@@ -88,6 +89,22 @@ describe(
         "Sample Database",
       );
     });
+
+    it("should not update the setting when the same database is selected again", () => {
+      cy.request("PUT", "/api/setting/last-used-database-id", {
+        value: SAMPLE_DB_ID,
+      });
+
+      startNativeQuestion();
+      cy.findByTestId("selected-database")
+        .should("have.text", "Sample Database")
+        .click();
+
+      cy.log("Pick the same database again");
+      selectDatabase("Sample Database");
+      cy.get("@persistDatabase").should("be.null");
+    });
+
     describe("permissions", () => {
       it("users with 'No self-service' data permissions should be able to choose only the databases they can query against", () => {
         cy.signIn("nodata");
--- a/e2e/test/scenarios/native/native-database-source.cy.spec.js
+++ b/e2e/test/scenarios/native/native-database-source.cy.spec.js
@@ -8,6 +8,7 @@ import {
 
 const PG_DB_ID = 2;
 const mongoName = "QA Mongo";
+const postgresName = "QA Postgres12";
 const additionalPG = "New Database";
 
 const { DATA_GROUP } = USER_GROUPS;
@@ -26,7 +27,7 @@ describe(
     });
 
     it("smoketest: persisting last used database should work, and it should be user-specific setting", () => {
-      const adminPersistedDatabase = "QA Postgres12";
+      const adminPersistedDatabase = postgresName;
       const userPersistedDatabase = "Sample Database";
 
       startNativeQuestion();
@@ -76,10 +77,10 @@ describe(
       startNativeModel();
       assertNoDatabaseSelected();
 
-      selectDatabase("QA Postgres12");
+      selectDatabase(postgresName);
 
       startNativeQuestion();
-      assertSelectedDatabase("QA Postgres12").click();
+      assertSelectedDatabase(postgresName).click();
       selectDatabase("Sample Database");
 
       startNativeModel();
@@ -123,7 +124,7 @@ describe(
       cy.log(
         "Persisting a database for a native model should not affect actions",
       );
-      selectDatabase("QA Postgres12");
+      selectDatabase(postgresName);
       cy.wait("@persistDatabase");
 
       startNewAction();
@@ -137,7 +138,7 @@ describe(
         startNativeQuestion();
         cy.wait("@persistDatabase");
         cy.findByTestId("selected-database")
-          .should("have.text", "QA Postgres12")
+          .should("have.text", postgresName)
           .click();
 
         cy.get(POPOVER_ELEMENT).should("not.exist");
@@ -151,11 +152,11 @@ describe(
         startNativeQuestion();
 
         cy.findByTestId("selected-database")
-          .should("have.text", "QA Postgres12")
+          .should("have.text", postgresName)
           .click();
 
         popover()
-          .should("contain", "QA Postgres12")
+          .should("contain", postgresName)
           .and("contain", "New Database");
       });
     });
@@ -166,7 +167,7 @@ describe(
       startNativeQuestion();
       cy.wait("@persistDatabase");
       cy.findByTestId("selected-database")
-        .should("have.text", "QA Postgres12")
+        .should("have.text", postgresName)
         .click();
 
       cy.get(POPOVER_ELEMENT).should("not.exist");
@@ -189,7 +190,7 @@ describe(
       cy.signInAsNormalUser();
       startNativeQuestion();
       // Postgres will be automatically selected because it's the only dataabse this user can query
-      assertSelectedDatabase("QA Postgres12");
+      assertSelectedDatabase(postgresName);
     });
   },
 );
--- a/e2e/test/scenarios/native/native-database-source.cy.spec.js
+++ b/e2e/test/scenarios/native/native-database-source.cy.spec.js
@@ -10,13 +10,16 @@ describe(
     });
 
     it("smoketest: persisting last used database should work, and it should be user-specific setting", () => {
+      const adminPersistedDatabase = "QA Postgres12";
+      const userPersistedDatabase = "Sample Database";
+
       startNativeQuestion();
       assertNoDatabaseSelected();
 
-      selectDatabase("QA Postgres12");
+      selectDatabase(adminPersistedDatabase);
 
       startNativeQuestion();
-      assertSelectedDatabase("QA Postgres12");
+      assertSelectedDatabase(adminPersistedDatabase);
 
       cy.signOut();
       cy.signInAsNormalUser();
@@ -24,16 +27,16 @@ describe(
       startNativeQuestion();
       assertNoDatabaseSelected();
 
-      selectDatabase("Sample Database");
+      selectDatabase(userPersistedDatabase);
 
       startNativeQuestion();
-      assertSelectedDatabase("Sample Database");
+      assertSelectedDatabase(userPersistedDatabase);
 
       cy.signOut();
       cy.signInAsAdmin();
 
       startNativeQuestion();
-      assertSelectedDatabase("Sample Database");
+      assertSelectedDatabase(adminPersistedDatabase);
     });
 
     it("deleting previously persisted database should result in the new database selection prompt", () => {
--- a/e2e/test/scenarios/native/native-database-source.cy.spec.js
+++ b/e2e/test/scenarios/native/native-database-source.cy.spec.js
@@ -82,10 +82,41 @@ describe(
   },
 );
 
+describe("mongo as the default database", { tags: "@mongo" }, () => {
+  beforeEach(() => {
+    restore("mongo-5");
+    cy.signInAsAdmin();
+  });
+
+  const MONGO_DB_NAME = "QA Mongo";
+
+  it("should persist Mongo database, but not its selected table", () => {
+    startNativeQuestion();
+    assertNoDatabaseSelected();
+
+    selectDatabase(MONGO_DB_NAME);
+    cy.findByTestId("native-query-top-bar")
+      .findByText("Select a table")
+      .click();
+    popover().findByText("Reviews").click();
+    cy.findByTestId("native-query-top-bar").should(
+      "not.contain",
+      "Select a table",
+    );
+
+    startNativeQuestion();
+
+    assertSelectedDatabase(MONGO_DB_NAME);
+    cy.findByTestId("native-query-top-bar").should("contain", "Select a table");
+  });
+});
+
 function startNativeQuestion() {
   cy.visit("/");
   cy.findByTestId("app-bar").findByText("New").click();
-  popover().findByTextEnsureVisible("SQL query").click();
+  popover()
+    .findByTextEnsureVisible(/(SQL|Native) query/)
+    .click();
 }
 
 function startNativeModel() {
--- a/e2e/test/scenarios/native/native-database-source.cy.spec.js
+++ b/e2e/test/scenarios/native/native-database-source.cy.spec.js
@@ -6,6 +6,8 @@ import {
   POPOVER_ELEMENT,
 } from "e2e/support/helpers";
 
+const PG_DB_ID = 2;
+
 describe(
   "scenarios > question > native > database source",
   { tags: "@external" },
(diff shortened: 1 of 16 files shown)


frontend/src/metabase/query_builder/components/NativeQueryEditor/NativeQueryEditor.tsx
```diff
@@ -909,6 +921,14 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
     );
     return Questions.HACK_getObjectFromAction(action);
   },
+  onSetDatabaseId: (id: DatabaseId) => {
```
ranquild, line 924: Please remove `shouldPersistLastUsedDatabase`, add a callback to `NativeQueryEditor` like `onSetDatabaseId` and put the action in `View` and not here

ranquild (commented):
@nemanjaglumac we would also need to handle database ids after the database is removed. I think we should do in clojure in `last-used-database-id` getter and return `null` if the corresponding DB cannot be found (could be done in another PR)

frontend/src/metabase/query_builder/components/NativeQueryEditor/NativeQueryEditor.tsx
nemanjaglumac: Help me understand this better.
What are we trying to achieve by moving the dispatch up to the `View`?

The reason I chose this approach was because I wanted a fine-grained control of when exactly do we persist the database. NativeQueryEditor can be rendered by many different components.

How are we going to control when does native query editor persists the database ID, and when does it ignore it?

nemanjaglumac: To answer my own question - this should do it.
https://github.com/metabase/metabase/pull/38986/commits/4dd6aad70815b5780baf5a675b2f62ed98146ebc

I'd still like to learn why is one approach considered better than the other?

e2e/test/scenarios/native/native-database-source.cy.spec.js
```diff
@@ -0,0 +1,247 @@
+import { SAMPLE_DB_ID } from "e2e/support/cypress_data";
+import {
+  restore,
+  popover,
+  addPostgresDatabase,
+  POPOVER_ELEMENT,
+} from "e2e/support/helpers";
+
+const PG_DB_ID = 2;
```
kamilmielnik, line 9: Isn't this `WRITABLE_DB_ID` ?

e2e/test/scenarios/native/native-database-source.cy.spec.js
```diff
@@ -0,0 +1,247 @@
...
+
+    startNativeQuestion();
+
+    assertSelectedDatabase(mongoName);
+    cy.findByTestId("native-query-top-bar").should("contain", "Select a table");
+  });
+});
+
+function startNativeQuestion() {
+  cy.visit("/");
```
kamilmielnik, line 203: This is expensive. Can we use SPA navigation instead?

```suggestion
  cy.findByTestId("main-logo-link").click();
```

----

The same comment applies to other `cy.visit` calls in this test suite.

e2e/test/scenarios/native/native-database-source.cy.spec.js
```diff
@@ -0,0 +1,247 @@
...
+
+      selectDatabase(additionalPG);
+
+      cy.log("Delete previously persisted database.");
+      cy.get("@postgresID").then(databaseId => {
+        cy.request("DELETE", `/api/database/${databaseId}`);
+      });
+
+      startNativeQuestion();
+      assertNoDatabaseSelected();
```
kamilmielnik, line 70: I wonder if this test is going to pass if `startNativeQuestion` does not call `cy.visit`.
It should not, and we should ensure that the setting value is not stale (i.e. is refetched) when visiting native query editor.

e2e/test/scenarios/native/native-database-source.cy.spec.js
```diff
@@ -0,0 +1,247 @@
...
+      );
+      selectDatabase("QA Postgres12");
+      cy.wait("@persistDatabase");
+
+      startNewAction();
+      assertNoDatabaseSelected();
+    });
+
+    describe("permissions", () => {
+      it("users with 'No self-service' data permissions should be able to choose only the databases they can query against", () => {
```
kamilmielnik, line 132: I think this scenario (similar to when DB is removed) is also worth testing:

1. User has access to DB X
2. User creates a native query based on DB X
3. User is revoked permissions to DB X
4. User opens native query editor

Assertion: DB X is not listed and not selected

src/metabase/models/user.clj
```diff
@@ -439,6 +439,12 @@
   :user-local :only
   :type :string)
 
+(defsetting last-used-database-id
+  (deferred-tru "The last database a user has used for a native query.")
```
kamilmielnik, line 443: I suggest to be more specific with the name.

```suggestion
(defsetting last-used-native-query-database-id
  (deferred-tru "The last database a user has used for a native query.")
```

kamilmielnik (commented):
Beautifully written test suite <3

frontend/src/metabase/query_builder/components/DataSelector/TriggerComponents.tsx
```diff
@@ -80,7 +80,12 @@ export function FieldTrigger({
 
 export function DatabaseTrigger({ database }: { database: Database }) {
   return database ? (
-    <span className="text-wrap text-grey no-decoration">{database.name}</span>
+    <span
+      className="text-wrap text-grey no-decoration"
+      data-testid="selected-database"
```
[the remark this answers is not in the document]
nemanjaglumac: Yes, it is the same because they can never be rendered at the same time. It's either-or type of situation, but the meaning of that element is the same - it shows the selected database.

1. when it's the only option and thus rendered by default
2. when it's the chosen option out of many

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -397,4 +403,11 @@ class View extends Component {
...
 
-export default ExplicitSize({ refreshMode: "debounceLeading" })(View);
+const mapDispatchToProps = dispatch => ({
+  onSetDatabaseId: id => dispatch(rememberLastUsedDatabase(id)),
+});
+
+export default _.compose(
+  ExplicitSize({ refreshMode: "debounceLeading" }),
+  connect(null, mapDispatchToProps),
+)(View);
```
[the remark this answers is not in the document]
nemanjaglumac: It's funny cause [I used that exact same logic originally](https://github.com/metabase/metabase/pull/38986/commits/bffccafc66dd8b30707ce5f4724512e9d1e35bd8) (defined the action in the `NativeQueryEditor`) but @ranquild [asked me](https://github.com/metabase/metabase/pull/38986#discussion_r1499174779) to lift it to the `View`.

e2e/test/scenarios/native/native-database-source.cy.spec.js
nemanjaglumac: In theory yes.
But it can be misleading. We're using `WRITABLE_DB_ID` only in the context of actual writable scenarios, and we're not writing anything to this database.

I'm working on a better solution for all QA DB references, btw.

frontend/src/metabase-types/api/settings.ts
```diff
@@ -285,6 +285,7 @@ export interface Settings {
   "user-visibility": string | null;
   "last-acknowledged-version": string | null;
   "show-static-embed-terms": boolean | null;
+  "last-used-database-id"?: number | null;
```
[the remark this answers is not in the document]
nemanjaglumac: Great question @kamilmielnik !
This is something that I didn't know about, and it's not documented anywhere.

@ranquild explained it, and I plan to document it AND to update types to reflect this. But I thought it falls outside the scope of this PR.

tl;dr
`null` or `value` - user has access and it reads the actual value
`undefined` - user does not have permissions to fetch this

e2e/test/scenarios/native/native-database-source.cy.spec.js
nemanjaglumac: It is expensive, but it was the only reliable approach that gave me confidence this thing works.
I can explore some other options.

frontend/src/metabase-types/api/mocks/settings.ts
```diff
@@ -227,5 +227,6 @@ export const createMockSettings = (
   "uploads-schema-name": null,
   "user-visibility": null,
   "last-acknowledged-version": "v1",
+  "last-used-database-id": 1,
```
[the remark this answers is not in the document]
ranquild: I wouldn't do that. Technically these presents are not connected

frontend/src/metabase/query_builder/components/view/View.jsx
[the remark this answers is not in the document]
ranquild: We should `connect` early. The current QueryBuilder/View/ViewHeader structure is broken. This is closer to the correct way of doing it

e2e/test/scenarios/native/native-database-source.cy.spec.js
ranquild: +1

ranquild (commented):
Looks good. I'm worried that the case with deleting the last selected database doesn't work well. There are some comments from @kamilmielnik about that. Could you double check?

test/metabase/models/user_test.clj
```diff
@@ -194,12 +194,12 @@
...
-     (mt/with-temporary-setting-values [enable-password-login false]
-       (ldap.test/with-ldap-server
-         (invite-user-accept-and-check-inboxes! :invitor default-invitor , :accept-invite? false)
-         (is (seq (mt/regex-email-bodies #"/auth/login"))))))))
+  (testing "if sso enabled and password login is disabled, email should send a link to sso login"
+    (mt/with-premium-features #{:disable-password-login}
+      (mt/with-temporary-setting-values [enable-password-login false]
+        (ldap.test/with-ldap-server
+          (invite-user-accept-and-check-inboxes! :invitor default-invitor , :accept-invite? false)
+          (is (seq (mt/regex-email-bodies #"/auth/login"))))))))
```
oleggromov, line 202: My 2c about changes like that — I am not sure which linters/prettifiers everyone uses for clojure, but probably different. Let's avoid changes like that because they unnecessarily change the history.

nemanjaglumac: Oopsy.
I didn't even notice that because I have the whitespace turned off in GH UI.

Thanks

frontend/src/metabase-types/api/mocks/settings.ts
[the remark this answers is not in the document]
ranquild: E.g. leave 1

ranquild (approved):

e2e/test/scenarios/native/native-database-source.cy.spec.js
nemanjaglumac: Unfortunately, our caching logic creates all sorts of issues not just here, but in a lot of other places. Unless we do a hard refresh (which visiting a home page achieves in Cypress), most of this will not work.

This also applies to [the other comment](https://github.com/metabase/metabase/pull/38986#discussion_r1503801570) and [my answer](https://github.com/metabase/metabase/pull/38986#discussion_r1503871038) there.

e2e/test/scenarios/native/native-database-source.cy.spec.js
nemanjaglumac: Added a new test for this.
As with the other scenarios, unless there is a hard refresh user will still see the old database.

Spoke to AlexP about this and we concluded that this is a global problem we have - definitely not tied to just this PR.

calherries (commented):
What happens when the last used database is deleted? Do we need a test for that?

Is there a FE benefit of implementing `last-used-native-database-id` as a user-local setting, instead of a new column of the `core_user` table? (fetched with `GET /api/user/current` and updated with `PUT /api/user/:id`)?

From my perpective on the BE, the advantage of adding a new column in the `core_user` table over using a setting is that it can be a foreign key to the database ID. Then we can do things like automatically set the value to NULL if the database is deleted.

calherries (commented):
What happens when the last used database is deleted? Do we need a test for that?

Is there a FE benefit of implementing `last-used-native-database-id` as a user-local setting, instead of a new column of the `core_user` table? (fetched with `GET /api/user/current` and updated with `PUT /api/user/:id`)?

From my perpective on the BE, the advantage of adding a new column in the `core_user` table over using a setting is that it can be a foreign key to the database ID. Then we can do things like automatically set the value to NULL if the database is deleted.

calherries:
Answering the bonus question here:

> Would it be ok to define a custom :getter for this new defsetting value that would first check if the database still exists AND if the current user has necessary permissions to query it?

Yes. We could also do this in `GET /api/user/current`

nemanjaglumac:
> What happens when the last used database is deleted? Do we need a test for that?

@calherries 
We already have it [here](https://github.com/metabase/metabase/pull/38986/files#diff-7140ba4e9ce90fa912c83d5042756afd2c4ee3fb0369b0e078683027ca7b2742R59). Frontend will prompt the user to select a new database.

calherries (approved):
Looks good enough to merge for me like this anyway, it's more of a nit if you're handling this on the backend. Tests are tight too

PR #38995, opened 2024-02-21
Use `getColumnKey` for `table.columns` viz setting

ranquild:
Migrates `table.columns` viz setting to use `getColumnKey` for matching and not MLv2 `findColumnIndexesFromLegacyRefs`. The latter has been proven to be unstable and I spent more than a month total figuring out issues with this function. `getColumnKey` is used for almost all other viz settings already.

Before, `table.columns` had this format:
```
[{
  name: "TAX",
  fieldRef: ["field", 1, null],
  enabled: true
}]
```

Now we add an additional `key` property that is computed with `getColumnKey`:
```
[{
  name: "TAX",
  fieldRef: ["field", 1, null],
  key: '["ref",["field",1,null]]',
  enabled: true
}]
```

Migration:
- We re-create `key` based on `name` and `fieldRef` when it is empty. 
- We set `key` in `column.js`, so all existing table settings would get this property when modifying the setting for the first time after upgrade.
- We still support missing `key` when matching columns with column settings for cases where table settings are accessed outside of the settings framework, like in `Question.ts`.

Downgrade:
- Because we don't remove existing properties, downgrade should work as expected.

Query -> Viz settings sync changes:
- To make it work correctly in all cases, we cannot rely on MLv2 to generate column names or `fieldRef`. This PR changes how `table.columns` setting is synced for MBQL queries to make it work like it does for native queries. Now, when the query results are available, we sync the setting using metadata from the QP and not MLv2. QP-based metadata is used for all viz settings, and it's great that we can make the QP the single source of truth here.
- One note is that syncing `graph.metrics` is still based on MLv2, and this is something we should refactor as well in the near future.

How to verify:
- CI is green
- Try hiding/showing columns in table viz setings

What's next?
- Migrate `graph.metrics` to use QP results and not MLv2 for syncing
- I could force setting `key` for all updated questions so we might be able to get rid of legacy properties after several releases:
```
[{
  name: "TAX",
  key: '["ref",["field",1,null]]',
  enabled: true
}]
```


frontend/src/metabase/visualizations/lib/settings/column.js
```diff
@@ -5,23 +5,35 @@ import _ from "underscore";
...
-  numberFormatterForOptions,
   getCurrencySymbol,
   getDateFormatFromStyle,
+  numberFormatterForOptions,
 } from "metabase/lib/formatting";
 
 import { hasHour } from "metabase/lib/formatting/datetime-utils";
 
 import { currency } from "cljs/metabase.shared.util.currency";
+import MetabaseSettings from "metabase/lib/settings";
```
ranquild, line 18: eslint

frontend/src/metabase/visualizations/lib/settings/column.js
```diff
@@ -532,24 +530,16 @@ export const buildTableColumnSettings = ({
...
-        query = question.query();
-      } catch (e) {
-        console.warn(e);
-      }
-
-      return _.all(
-        enabledColumns,
-        columnSetting =>
-          findColumnIndexForColumnSetting(data.cols, columnSetting, query) >= 0,
+      const query = question.query();
```
ranquild, line 533: actual change

frontend/src/metabase/visualizations/components/settings/ChartSettingTableColumns/TableColumnPanel/utils.ts
```diff
@@ -9,78 +9,70 @@ import type {
...
+  const updatedIndexes = [...originalIndexes];
+  const updatedSettings = [...originalSettings];
+  columns.forEach((column, columnIndex) => {
+    const columnSettingIndex = originalIndexes[columnIndex];
+    if (columnSettingIndex < 0) {
+      updatedIndexes[columnIndex] = updatedSettings.length;
+      updatedSettings.push({
+        name: column.name,
+        fieldRef: column.field_ref,
+        enabled: false,
```
ranquild, line 36: Previously it was `true`, but that was a mistake in the previous PR

frontend/src/metabase/visualizations/lib/settings/column.js
```diff
@@ -530,27 +526,6 @@ export const buildTableColumnSettings = ({
     // title: t`Columns`,
     widget: ChartSettingTableColumns,
     getHidden: (series, vizSettings) => vizSettings["table.pivot"],
-    isValid: ([{ card, data }]) => {
```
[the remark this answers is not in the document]
uladzimirdev: if you know what you do 👍

uladzimirdev (approved):
🤞

alxnddr:
I found one little bug with CC but I'm not sure if it is related. Can you please have a look and tell?

https://github.com/metabase/metabase/assets/14301985/4a87945b-016b-41cf-8ecf-05b92945acbf

ranquild:
@alxnddr MLv2 bug in `removeField` function, unrelated to this PR and viz settings in general

alxnddr:
@ranquild got it, thanks for checking 👍 

Probably this is also not related: implicit joins get lost after changing query table columns selection
https://github.com/metabase/metabase/assets/14301985/9bb657f1-569e-4ca1-8ff2-cdac347828bc

ranquild:
@alxnddr yep, also MLv2. In the notebook editor we use `withFields` when the user changes the selection, and this overrides implicitly joinable fields atm. Also not viz related

frontend/src/metabase-lib/queries/utils/dataset.ts
```diff
@@ -1,93 +1,46 @@
...
- */
-export function fieldRefForColumn(
-  column: DatasetColumn,
-): FieldReference | null | undefined {
-  // NOTE: matching existing behavior of returning the unwrapped base dimension until we understand the implications of changing this
-  return (
-    column.field_ref &&
-    Dimension.parseMBQL(column.field_ref)?.baseDimension().mbql()
-  );
+export function getColumnSettingKey({
```
alxnddr, line 12: I assume the backend implementation for the table static viz should match this to have entirely synced behavior. If I'm right, can you please create an issue and provide an example where it did not work so that BE engineers can work on it? cc @markbastian @adam-james-v

alxnddr (approved):
FE change looks good to me, all cases I tried worked (except for a few mlv2 bugs)

PR #38999, opened 2024-02-21
Query execution time

### LINKED ISSUE #9408: Feature request: Query response time
### Feature request

It would be great if we can see the actual query response time on the UI, in questions or in SQL editor

Like "Query took 2.02 seconds to complete"

kamilmielnik:
Closes #9408

Design doc: https://www.notion.so/metabase/Show-query-response-time-in-native-queries-ca4c47e01349463198446589848c2714

Slack discussion: https://metaboat.slack.com/archives/C0645JP1W81/p1708516928531709

### Description

Shows query execution time for native queries in the footer of the query builder.
- If the time is less than 1 second, the result is formatted as milliseconds (e.g. `123 ms`)
- If the time is at least 1 second, the result is formatted as seconds with 1 decimal place (e.g. `3.9 s`)

### Demo

https://github.com/metabase/metabase/assets/6830683/72d518f5-c1c9-4a40-9eb6-f8c3241d1925


frontend/src/metabase/query_builder/components/view/ExecutionTime/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { t } from "ttag";
+
+export const formatDuration = (time: number): string => {
```
uladzimirdev, line 3: nit: we have other time utils in `frontend/src/metabase/lib/formatting/time.ts`

frontend/src/metabase/query_builder/components/view/ExecutionTime/ExecutionTime.tsx
```diff
@@ -0,0 +1,15 @@
+import { t } from "ttag";
+
+import { Tooltip } from "metabase/ui";
+
```
uladzimirdev, line 4: nit: I believe there shouldn't be an empty line between those groups as both they both point out to the relative files

frontend/src/metabase/query_builder/components/view/ExecutionTime/utils.unit.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+
+describe("formatDuration", () => {
+  it("formats duration correctly", () => {
+    expect(formatDuration(100)).toBe("100 ms");
+    expect(formatDuration(999)).toBe("999 ms");
+    expect(formatDuration(1000)).toBe("1.0 s");
+    expect(formatDuration(1425)).toBe("1.4 s");
+    expect(formatDuration(1475)).toBe("1.5 s");
+    expect(formatDuration(1500)).toBe("1.5 s");
+    expect(formatDuration(120000)).toBe("120.0 s");
```
uladzimirdev, line 11: `.0` looks weird

uladzimirdev (approved):
Minor comments

frontend/src/metabase/query_builder/components/view/ExecutionTime/ExecutionTime.tsx
kamilmielnik: If it starts with a dot it's _relative_, if it's absolute it's _internal_, so all is good here

frontend/src/metabase/query_builder/components/view/ExecutionTime/utils.ts
kamilmielnik: I know but none of them could be reused here and there already is a `duration` function - I didn't want to put another function like that in there, because it'd be hard to distinguish them.

So I decided it's best to keep it local to the component it's used in.

frontend/src/metabase/query_builder/components/view/ExecutionTime/utils.unit.spec.ts
kamilmielnik: On the other hand `2.1 s` and `2 s` look inconsistent and would cause small layout shift

frontend/src/metabase/query_builder/components/view/ExecutionTime/ExecutionTime.tsx
uladzimirdev: that's not true, you know, it's a relative path with webpack paths mapping

kamilmielnik: I'm not sure about terminology here but this is what `import/order` enforces with the current config

PR #39005, opened 2024-02-21
Fix Interactive embedding dashboard with multiple tabs and parameter header=false will not load

WiNloSt:
> [!note]
> This fix is required in order to work on https://github.com/metabase/metabase/issues/38429

Closes https://github.com/metabase/metabase/issues/39002

### Description

The problem came from the way we use `header=false`.

https://github.com/metabase/metabase/blob/ac3734876c162d8ebc909967ee8507309b4fbaf1/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx#L487-L492

Here we see that we don't render `DashboardHeader` when `!isHeaderVisible` and this property is read from the `header` parameter.

https://github.com/metabase/metabase/blob/02be47075120e8121a6817a22f66f1d6c1112447/frontend/src/metabase/dashboard/selectors.ts#L415-L418

Inside `DashboardHeader` it calls `useDashboardTabs`
https://github.com/metabase/metabase/blob/02be47075120e8121a6817a22f66f1d6c1112447/frontend/src/metabase/dashboard/components/DashboardTabs/DashboardTabs.tsx#L36

which will initiate the Redux state `dashboard.selectedTabId`, without this value, the dashboard card will never load.

https://github.com/metabase/metabase/blob/02be47075120e8121a6817a22f66f1d6c1112447/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx#L373-L377

### How to verify

- Follow the repro steps in https://github.com/metabase/metabase/issues/39002
- Or run `e2e/test/scenarios/embedding/interactive-embedding.cy.spec.js`'s `should hide the dashboard with multiple tabs header by a param (metabase#39002)` test.

#### Before
The dashboard card will never start loading

#### After
image

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

Issue #39002, opened 2024-02-21
Interactive embedding dashboard with multiple tabs and parameter `header=false` will not load


WiNloSt:
### Describe the bug

I found this bug while reproducing https://github.com/metabase/metabase/issues/38429

In interactive embedding, if you visit a dashboard that has multiple tabs **and** has the parameter `header=false`, the dashboard will not finish loading

### To Reproduce

0. Set up interactive embedding
1. Create a dashboard with multiple tabs
2. Visit the interactive embedding page where the iframe URL has this parameter `?header=false`
3. Visit the dashboard we created in 1)
4. See that the card will not stop spinning. In fact the card requests will never be initiated.

### Expected behavior

The dashboard loads normally

### Information about your Metabase installation

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.name": "OpenJDK Runtime Environment",
    "java.runtime.version": "11.0.14.1+1-LTS",
    "java.vendor": "Azul Systems, Inc.",
    "java.vendor.url": "http://www.azul.com/",
    "java.version": "11.0.14.1",
    "java.vm.name": "OpenJDK 64-Bit Server VM",
    "java.vm.version": "11.0.14.1+1-LTS",
    "os.name": "Mac OS X",
    "os.version": "14.2.1",
    "user.language": "en",
    "user.timezone": "Asia/Bangkok"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
    "application-database-details": {
      "database": {
        "name": "H2",
        "version": "2.1.214 (2022-06-13)"
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
        "version": "2.1.214 (2022-06-13)"
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2024-02-20",
      "src_hash": "e3b554bc3d8e4448572d8fdde9681523aca052d8",
      "tag": "v0.48.1-SNAPSHOT",
      "hash": "11823a5"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

[...]

PR #39007, opened 2024-02-21
Fix can't select dashboard tabs in interactive embedding with parameter `header=false`

### LINKED ISSUE #38429: No option to hide/show tabs in Interactive embedding as a parameter. header=false will hide all tabs and go to tab 1 by default
Setting `header=false` will hide tab selection and default to 1
In addition, I tried to put tab=x in the embedded URL and couldn't set the desired tab

Have header=false not hide tabs in a dashboard, and instead have a new parameter that hides/shows tabs

Enabling header, but you might want the ability to hide one and not the other.

Requested by a customer, internal ticket [24697](https://metabase.zendesk.com/agent/tickets/24697)

**Additional context**
N/A

WiNloSt:
> [!note]
> This PR supersedes #39005

Closes #38429, #39002

### Description

The [`header` documentation](https://www.metabase.com/docs/latest/embedding/interactive-embedding#header) indicates
> To hide a question or dashboard’s title, additional info, and action buttons

Dashboard tabs is a very crucial part of the dashboard and should, in my opinion, never been hidden.

This PR fixes the problem by only hiding the other parts in the dashboard header except the tabs

**Note** Please also see the #39005 explaining the fix for the dashboard card not loading which is also fixed in this PR with a different approach.

### How to verify
0. Set up interactive embedding
1. Create a dashboard with multiple tabs
2. Visit the interactive embedding page where the iframe URL has this parameter `?header=false`
3. Visit the dashboard we created in 1)
4. See that the tabs are visible and you can select each tab.

#### Before
image

#### After
image

### Checklist


frontend/src/metabase/dashboard/components/Dashboard/Dashboard.tsx
```diff
@@ -484,21 +483,24 @@ function DashboardInner(props: DashboardProps) {
...
-                {...props}
-                onEditingChange={handleSetEditing}
-                setDashboardAttribute={handleSetDashboardAttribute}
-                addParameter={addParameter}
-                parametersWidget={parametersWidget}
-                onSharingClick={handleToggleSharing}
-              />
-            </DashboardHeaderContainer>
-          )}
+          <DashboardHeaderContainer
```
WiNloSt, line 486: Pro tip: Use the hide whitespace option for EZ review.
image

frontend/src/metabase/dashboard/components/DashboardHeader/DashboardHeaderView.tsx
```diff
@@ -120,52 +124,54 @@ export function DashboardHeaderComponent({
...
-        <HeaderRow
-          className={cx("QueryBuilder-section", headerClassName)}
-          data-testid="dashboard-header"
-          ref={header}
-        >
-          <HeaderFixedWidthContainer
-            data-testid="fixed-width-dashboard-header"
-            isNavBarOpen={isNavBarOpen}
-            isFixedWidth={dashboard?.width === "fixed"}
+        {isDashboardHeaderVisible && (
```
WiNloSt, line 127: Only hide the part before the dashboard tabs, because the dashboard tabs should always be visible.

kulyk (approved):

albertoperdomo (approved):
LGTM

Related: I am getting the connect your database to get get the most from MB banner at the top in the embed, but I am also seeing that on master, so related but separate issue.

WiNloSt:
@albertoperdomo Feel free to open an issue for that too. I remember we haven't talked about hiding that banner specifically in embedding, we only talked about hiding that database prompt banner when users are whitelabeling. That feels to me we should do that for interactive embedding as well, since interactive embedding is also closely related to whitelabeling.

WiNloSt:

npretto (commented):
~Tested locally~
commented the wrong one 😅

PR #39010, opened 2024-02-21
[serdes-v1] can now dump/load entity_id

### LINKED ISSUE #38961: Add entity-id to Serdes v1 as an optional flag
**Context**
As we introduced entity-ids before releasing Serdes v2, some customers are facing problems to upgrade to Serdes v2.
[Thread](https://metaboat.slack.com/archives/C01LQQ2UW03/p1708378329446559)

The solutions is
1.v1 dump and load will copy the source's new entity_ids, but use v1's matching of source and target entities (which is presumably working okay for them and doesn't care about entity_ids)

now they have matching entity_ids so v2 works fine.

To make this happen, v1 must transfer entity-ids and overwrite the target entity-ids if they exist.
It should be added behind a flag --include-entity-id to both dump and load commands

This should be backported to 47 and 48.

sansolovyov:
Specifying `--include-entity-id` to a serdes v1 dump command will dump `entity_id` field into yamls. Load will use this field with no options though - but it'll check if it is present in data to use as one of markers that data has changed.

I decided not to add an option to load since right now it will use all data provided, just the markers that say "data has changed" were changed.

Resolves #38961

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
+++ b/enterprise/backend/src/metabase_enterprise/serialization/cmd.clj
@@ -5,6 +5,7 @@
    [clojure.string :as str]
    [metabase-enterprise.serialization.dump :as dump]
    [metabase-enterprise.serialization.load :as load]
+   [metabase-enterprise.serialization.serialize :as serialize]
    [metabase-enterprise.serialization.v2.entity-ids :as v2.entity-ids]
    [metabase-enterprise.serialization.v2.extract :as v2.extract]
    [metabase-enterprise.serialization.v2.ingest :as v2.ingest]
@@ -181,7 +182,7 @@
 
 (defn v1-dump!
   "Legacy Metabase app data dump"
-  [path {:keys [state user] :or {state :active} :as opts}]
+  [path {:keys [state user include-entity-id] :or {state :active} :as opts}]
   (log/info (trs "BEGIN DUMP to {0} via user {1}" path user))
   (mdb/setup-db!)
   (check-premium-token!)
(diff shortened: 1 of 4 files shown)


enterprise/backend/src/metabase_enterprise/serialization/upsert.clj
```diff
@@ -53,14 +53,16 @@
...
-       (select-keys entity)
-       (m/map-vals (fn [v]
-                     (if (coll? v)
-                       (json/encode v)
-                       v)))
-       (m/mapply t2/select-one model)))
+  (let [entity-id (if (:entity_id entity) [:entity_id] [])]
+    (->> (or (identity-condition model)
+             (throw (ex-info (trs "Model {0} does not support upsert" model) {:model model})))
+         (into entity-id)
```
bshepherdson, line 59: If I'm understanding this correctly, this will mean we need to clear the destination's `entity_id`s first. Otherwise a user with `entity_id` mismatches will get duplicates.

luizarakaki: We shouldn't check entity-id for identity-matching, right?
The purpose here is to overwrite entity-ids based on Serdes v1 identity-condition

piranha: Hmmm but this is not identity matching, this is "has data changed" call, if I understand correctly. I mean test passes right now and does not generate duplicates even though entity ids are different?

piranha: @bshepherdson I have thought about this a little bit more and I agree this change looks like it could lead to duplicates. IDK why it did not in my test. But now I've replaced it with just `--mode update` command option and it seems like this is what we need?

bshepherdson (approved):

piranha:

piranha:

PR #39011, opened 2024-02-21
Remove special describe-table-fks for redshift

calherries:
This PR removes an implementation for `driver/describe-table-fks` for redshift. 

We had a special case for redshift because apparently [`getImportedKeys`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getImportedKeys-java.lang.String-java.lang.String-java.lang.String-) didn't work 5 years ago when the driver was first implemented. 

It's not required anymore because `getImportedKeys` has since been implemented, at least it I could trace it back to 2.1.0.11 in [this commit](https://github.com/aws/amazon-redshift-jdbc-driver/commit/7a1efb19412831bd0b9d20842d172df988664b8f).

### DIFF
--- a/modules/drivers/redshift/src/metabase/driver/redshift.clj
+++ b/modules/drivers/redshift/src/metabase/driver/redshift.clj
@@ -49,38 +49,6 @@
   [& args]
   (apply (get-method driver/describe-table :sql-jdbc) args))
 
-;; The Postgres JDBC .getImportedKeys method doesn't work for Redshift, and we're not allowed to access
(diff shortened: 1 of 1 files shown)


crisptrutski (approved):
Min code min problems. Does this give us any measurable speedup?

calherries:
I didn't check the difference in speed tbh. I plan to overwrite the current implementation of `describe-table-fks` in [faster sync-fks PR](https://github.com/metabase/metabase/pull/38970) anyway, so this PR is mostly about making that PR smaller

Issue #38811, opened 2024-02-15
[Epic] Remove global plain CSS styles


deniskaber:
Relates to https://github.com/metabase/metabase/issues/33854

**Links**
- Eng doc: [Link](https://www.notion.so/metabase/83-Prepare-global-css-for-migration-to-css-modules-4ab29b8589b545e0960bf032d24a11a9)
- Problem description: [Link](https://www.notion.so/metabase/Embedding-SDK-8103306366be4f0786b489ad2324235c?pvs=4#83440af594a941ee8b14aaa1e1ac742b)

**Implementation Plan**

```[tasklist]
### Milestone 1 - Migrate global CSS to CSS modules
- [x] https://github.com/metabase/metabase/pull/39020 - Migrate global CSS to CSS modules - Add ":global" for all empty global styles; Remove ":local"
- [x] https://github.com/metabase/metabase/pull/39020 - Configure css-loader to have local scope by default
- [x] https://github.com/metabase/metabase/pull/39020 - Update css-loader dependency
- [x] https://github.com/metabase/metabase/pull/39020 - Find a way to isolate global third-party stylesheets - leaflet, ace-builds, tippy.js ...
- [x] https://github.com/metabase/metabase/pull/39220 - Remove unused classes
- [ ] https://github.com/metabase/metabase/pull/39239
```

```[tasklist]
### Milestone 2 - Get rid of global styles in Embedding SDK components
- [ ] https://github.com/metabase/metabase/issues/39123
- [ ] https://github.com/metabase/metabase/pull/39324
- [ ] https://github.com/metabase/metabase/pull/40103
- [ ] https://github.com/metabase/metabase/pull/40106
- [ ] https://github.com/metabase/metabase/pull/40007
- [ ] https://github.com/metabase/metabase/issues/40128
- [ ] https://github.com/metabase/metabase/issues/40127
- [ ] https://github.com/metabase/metabase/issues/40349
- [ ] https://github.com/metabase/metabase/issues/40722
- [ ] https://github.com/metabase/metabase/issues/40904
- [ ] https://github.com/metabase/metabase/issues/40496
- [ ] https://github.com/metabase/metabase/issues/40129
- [ ] https://github.com/metabase/metabase/issues/40910
- [ ] https://github.com/metabase/metabase/issues/41019
- [ ] https://github.com/metabase/metabase/issues/41077
- [ ] https://github.com/metabase/metabase/issues/40231
- [ ] https://github.com/metabase/metabase/issues/41256
[...]

PR #39020, opened 2024-02-21
38811 no global css 1

deniskaber:
Relates to https://github.com/metabase/metabase/issues/38811 , Milestone 1

### Description

This migrates global CSS files to css modules (but without classes isolation).
+ upgrade css-loader package

### How to verify
- Switch to master branch, build the app with `yarn build`
- Copy `app-main.css`, `styles.css`, `vendor.css` from `resources/frontend_client/app/dist`.
- Switch to this branch, build the app with `yarn build`
- Compare generated `app-main.css`, `styles.css`, `vendor.css` with ones from master branch.

Those should have same content.


webpack.config.js
```diff
@@ -42,9 +42,13 @@ const BABEL_CONFIG = {
 };
 
 const CSS_CONFIG = {
-  localIdentName: devMode
-    ? "[name]__[local]___[hash:base64:5]"
-    : "[hash:base64:5]",
+  modules: {
+    auto: filename => !filename.includes("node_modules"),
```
WiNloSt, line 46: Seems we're still relying on loading style from ACE editor

#### `master`
Screenshot 2024-02-23 at 4 10 03 PM


#### This PR
Screenshot 2024-02-23 at 4 03 35 PM

I checked that the rules for ace classes like `.ace_text-layer` is missing.

uladzimirdev:
My idea to this epic was to go step by step, I ended up with the similar changes in my attempts: hundreds of files and many broken or possibly broken places

webpack.config.js
```diff
@@ -228,19 +239,19 @@ const config = (module.exports = {
...
       chunksSortMode: "manual",
-      chunks: ["vendor", "styles", "app-main"],
+      chunks: ["vendor", "vendor-styles", "styles", "app-main"],
       template: __dirname + "/resources/frontend_client/index_template.html",
     }),
     new HtmlWebpackPlugin({
       filename: "../../public.html",
       chunksSortMode: "manual",
-      chunks: ["vendor", "styles", "app-public"],
+      chunks: ["vendor", '"vendor-styles", styles', "app-public"],
```
WiNloSt, line 248: Typo here

```suggestion
      chunks: ["vendor", "vendor-styles", "styles", "app-public"],
```

postcss.config.js
```diff
@@ -21,5 +21,6 @@ module.exports = {
         "focus-visible-pseudo-class": false,
       },
     },
+    "postcss-discard-comments": {},
```
WiNloSt, line 24: 👍

frontend/src/metabase/css/query_builder.module.css
```diff
@@ -496,22 +461,22 @@
...
 }
 
-.FilterInput:hover .FilterRemove-field {
+:global(.FilterInput:hover .FilterRemove-field) {
   opacity: 1;
 }
 
 /* need to do this ugliness to override the locally scoped font size from Scalar.css */
-.QueryBuilder .ScalarValue {
+:global(.QueryBuilder .ScalarValue) {
```
WiNloSt, line 475: You already removed this class `.QueryBuilder` from `frontend/src/metabase/query_builder/components/view/View.jsx`

So this one should be migrated.

#### This PR
image

#### `master`
image

frontend/src/metabase/admin/datamodel/components/GuiQueryEditor/GuiQueryEditor.jsx
```diff
@@ -212,7 +213,13 @@ export class GuiQueryEditor extends Component {
     } else {
       // TODO: move this into AggregationWidget?
       return (
-        <div className="Query-section Query-section-aggregation disabled">
+        <div
+          className={cx(
+            S.querySection,
+            S.disabled,
+            "Query-section-aggregation",
```
WiNloSt, line 220: This class is unused

```suggestion
```

frontend/src/metabase/admin/datamodel/components/GuiQueryEditor/GuiQueryEditor.jsx
```diff
@@ -224,9 +231,10 @@ export class GuiQueryEditor extends Component {
 
     return (
       <div
-        className={
-          "GuiBuilder-section GuiBuilder-data flex align-center arrow-right"
-        }
+        className={cx(
+          CS.arrow_right,
```
WiNloSt, line 235: There's not even this class exposed. Please see comments in `frontend/src/metabase/css/core/arrow.module.css`

frontend/src/metabase/css/core/arrow.module.css
```diff
@@ -1,37 +1,33 @@
 /* TODO: based on popover.css, combine them? */
 /* TODO: other arrow directions */
 
-.arrow-right {
+:global(.arrow-right) {
```
WiNloSt, line 4: there's no class exposed via css module. You referenced it like `CS.arrow_right`.

there are 2 things here.
 1. I think we should use camel case for CSS Module so this should be `.arrowRight`
 2. we need to add those classes in the CSS module (there are only global classes)

frontend/src/metabase/components/Calendar/Calendar.module.css
```diff
@@ -1,32 +1,32 @@
-.Calendar-weeks {
+:global .Calendar-weeks {
```
WiNloSt, line 1: I'm not super familiar with [scoping without parentheses](https://github.com/webpack-contrib/css-loader?tab=readme-ov-file#scope). But could it be possible to wrap these within parentheses for consistency?

frontend/src/metabase/components/LabelIcon/LabelIcon.module.css
```diff
@@ -1,16 +1,16 @@
...
-:local(.emojiIcon) {
+.emojiIcon {
   composes: text-centered from "style";
   font-size: 20px;
 }
 
-:local(.icon) {
-  composes: transition-color from "style";
+.icon {
+  transition: color 0.3s linear;
```
WiNloSt, line 14: you're inlining this property everywhere. Should you still keep that as a CSS module rule so you could compose it like other rules?

frontend/src/metabase/css/admin.module.css
```diff
@@ -7,60 +7,60 @@
 }
 
 /* utility to get a simple common hover state for admin items */
-.HoverItem:hover {
+:global(.HoverItem:hover) {
```
WiNloSt, line 10: Yeah just to reiterate, it'd be great if we could stick to one style either using parentheses or not if that's possible

WiNloSt (commented):
base on the number of files. This is 20% review. I'll continue later.

frontend/src/metabase/components/LabelIcon/LabelIcon.module.css
uladzimirdev: that's why I suggested separating removal of composes

WiNloSt (commented):
Based on the number of files this is about 45% review

uladzimirdev (commented):
Exceptional work! In this PR I like everything except removal/rewriting styles as it's really hard to track such changes.

Minor comments left

frontend/src/metabase/css/core/animation.module.css
```diff
@@ -1,5 +1,5 @@
 /* TODO - ideally this would be more reusable and not hardcode a value */
-@keyframes progress-bar {
+:global @keyframes progress-bar {
```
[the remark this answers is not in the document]
WiNloSt: @uladzimirdev If I remember correctly, I did check the output CSS file and it was correct. I think your syntax also works as I've seen that when searching too.

postcss.config.js
WiNloSt: Not required but could be a nice addition. I couldn't think of a reason to want to leave comments in the output CSS files.

uladzimirdev (approved):
go for it and fix problems if any

frontend/src/metabase/css/query_builder.module.css
```diff
@@ -69,174 +52,169 @@
...
   padding-left: 0 !important;
   padding-bottom: 0 !important;
 }
 
-.TooltipFilterList .Query-filterList {
+:global(.TooltipFilterList .Query-filterList) {
   flex-direction: column;
 }
 
-.TooltipFilterList .Query-section {
```
WiNloSt, line 81: I still found these 2 classes in use. Did you remove this rule or move it somewhere else?

https://github.com/metabase/metabase/blob/a1df251cd75d454e4f56d0b14993e59bd943af14/frontend/src/metabase/admin/datamodel/components/QueryDefinitionTooltip/QueryDefinitionTooltip.jsx#L30

https://github.com/metabase/metabase/blob/a1df251cd75d454e4f56d0b14993e59bd943af14/frontend/src/metabase/admin/datamodel/components/GuiQueryEditor/GuiQueryEditor.jsx#L129

https://github.com/metabase/metabase/blob/a1df251cd75d454e4f56d0b14993e59bd943af14/frontend/src/metabase/admin/datamodel/components/GuiQueryEditor/GuiQueryEditor.jsx#L215

frontend/src/metabase/models/components/NewModelOption.jsx
```diff
@@ -17,7 +19,7 @@ export default class NewModelOption extends Component {
     return (
       <Link
         {...props}
-        className="block no-decoration bg-white p4 align-center bordered rounded cursor-pointer transition-all full-height text-centered"
```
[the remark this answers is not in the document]
WiNloSt: Yeah this takes longer to review.

frontend/src/metabase/nav/components/AdminNavbar/AdminNavbar.tsx
```diff
@@ -36,7 +36,10 @@ export const AdminNavbar = ({
   const isPaidPlain = useSelector(getIsPaidPlan);
 
   return (
-    <AdminNavbarRoot className="Nav" aria-label={t`Navigation bar`}>
+    <AdminNavbarRoot
+      aria-label={t`Navigation bar`}
+      data-testid="admin-navbar-root"
```
WiNloSt, line 41: I've commented about `data-testid` somewhere. In short there are 2 points I want to be addressed here
1. Don't use `data-testid` for non-test implementation. Using `data` attribute is fine as long as it's not `data-testid` since people could mistake this as only for testing, which is not the truth here.
2. Would be nice to keep the logic the same by not separating between admin navbar and app navbar. My rationale would be to keep things as consistent as possible. This PRs is mainly refactoring so it shouldn't change the implementation and if we could use a single class we could also use a single data attribuse with the same value as well.

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -352,7 +352,10 @@ class View extends Component {
 
     return (
       <div className="full-height">
-        <QueryBuilderViewRoot className="QueryBuilder">
+        <QueryBuilderViewRoot
+          className="QueryBuilder"
+          data-testid="query-builder-root"
```
WiNloSt, line 357: I've commented about `data-testid` somewhere please take a look at those.

frontend/src/metabase/visualizations/components/ChartWithLegend.module.css
```diff
@@ -1,101 +1,101 @@
...
+.ChartWithLegend.horizontal.flexChart .Chart {
   flex-grow: 1;
   flex-shrink: 1;
   min-width: 66%;
 }
 
 /* DEBUG */
 /*
-:local .ChartWithLegend .Legend {
+.ChartWithLegend .Legend {
```
WiNloSt, line 92: These rules are commented out and they've been here for 2 years I think it's safe to remove them.

frontend/src/metabase/visualizations/components/LineAreaBarChart.module.css
```diff
@@ -0,0 +1,301 @@
...
+}
+
+:global(.LineAreaBarChart .dc-chart .voronoi-drill) {
+  cursor: pointer;
+}
+
+/* we put the brush behind everything so this isn't necessary
+/*:global(.LineAreaBarChart .dc-chart .brush) {
+  pointer-events: none;
+}*/
```
WiNloSt, line 218: This is commented out, so we should remove it.

WiNloSt (commented):
This is the final part of the review. This is mostly looking great. My main concerns are the use of `data-testid` in the non-test code. And how this PR differentiates between the admin and app navbars when we have been treating them the same with `.Nav` class.

uladzimirdev:
`ParameterValueWidget.css` problem

PR #39031, opened 2024-02-21
Admin logs layout fix & perf improvements

### LINKED ISSUE #39025: admin logs in the UI are hard to follow
### Describe the bug

the logs scroll now. But they also scroll when you found interesting bits you want to watch. this is an issue to make them better with no specific dictated goal in mind except "better".

### To Reproduce

view the logs of an instance that has some moderate traffic.

### Expected behavior

_No response_

### Logs

_No response_

```JSON
master
```

### Severity

p2

### Additional context

_No response_

sloansparger:
Close #39025

### Description

**Auto-follow bug**
There was a layout issue where if a single log was taller than the contain for the logs it would make the container grow taller than what would be visible to the user. This affected the calculation of if the users was scrolled to the bottom and would disable the auto-follow feature.

**User scroll position bug**
The logs page was not preserving the user's scroll position in some cases when logs would be updated. Upon inspecting our internal MB instance, it appears this was due to request times taking longer than our poll duration, causing state updates to happen out of order. This would remove logs from the list and thus change the relative position of other logs.

**Performance issues**
In our internal MB instance, payload sizes were coming in ~200kb-800kb gziped / ~2MB-5MB uncompressed. Given we combine logs from more than one machine, the internal list of logs  combined to ~8MB. Since we are polling for new updates, we were taking our 8MB list of logs and trying to merge / 2-5MB on top of it. Between our merging implementation and usage of map/filter in our Logs component's render method, we were making several copies of this in memory every second. This results in large freezes in the UI. Here you can see setState alone taking almost 3 second to update:
Screenshot 2024-02-22 at 4 59 36 PM

**The fix**

1. For the auto-follow bug - it was a matter of using flex on the container component for force the content to shrink to the expected size.
2. User scroll position - I have updated the UI to skip polling for new logs if a previous request is still in flight.
3. Performance issues
  a. skipping requests helps
  b. updated mergeLogs function to be more performant (3 sorts w/ 3 copies reduced to 1, _.uniq was slow so replaced with hand rolled dedup implementation)
  c. skip state merging / state updates if latest log in the most recent result is already present in our state
  d. moment date formatting on large logs sets was slow, moving to dayjs and caching the last result reduced date significantly

### How to verify

- go to admin logs page
- generate a large number of logs (for internal reviewers, you can borrow the response from stats)
- override your response body in dev tools with your large logs set ([how to do this in Chrome](https://developer.chrome.com/docs/devtools/overrides#set-up))
- your page should no longer freeze

### Checklist


frontend/src/metabase/admin/tasks/components/Logs/hooks.ts
```diff
@@ -0,0 +1,107 @@
...
+      console.warn("skipping logs request as a request is currently in flight");
+      return;
+    }
+
+    try {
+      isFetchingRef.current = true;
+      const newLogs: Log[] = await UtilApi.logs();
+      if (isMountedRef.current) {
+        setLoaded(true);
+        setError(null);
```
sloansparger, line 31: page will not be stuck on an error if we get a failed request

frontend/src/metabase/admin/tasks/components/Logs/hooks.ts
```diff
@@ -0,0 +1,107 @@
...
+        setLogs(logs => {
+          const newestLog = _.first(newLogs);
+          const hasFetchedNewLogs = !hasLog(logs, newestLog);
+          if (hasFetchedNewLogs) {
+            return mergeLogs([logs, newLogs.reverse()]);
+          }
+          return logs;
+        });
```
sloansparger, line 39: not updating the logs if we already have the latest produces really big gains when there's no updates

before optimizations + before skipping if already fetched (~180ms updates):
Screenshot 2024-02-21 at 2 46 38 PM

optimizations + before skipping if already fetched (~50ms updates):
Screenshot 2024-02-21 at 2 43 59 PM

optimizations + skipping if already fetched (<1ms updates):
Screenshot 2024-02-21 at 5 00 49 PM
Screenshot 2024-02-21 at 5 00 58 PM

frontend/src/metabase/admin/tasks/components/Logs/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+      return prevOutput;
+    }
+    prevInput = input;
+    prevOutput = fn(input);
+    return prevOutput;
+  }
+  return memoizedFn;
+}
+
+const formatTs = memoizeOne((ts: number) => dayjs(ts).format());
```
sloansparger: large amount of time was being spent formatting dates. since logs are sorted by time when we want to format them, memoizing just the last value was taking 25-80ms compute times down to ~10ms consistently.

paoliniluis:
Can you also order them inverted? I want the new lines to appear on the top so I don’t need to scroll. Also a button to copy all the log lines would be awesome

sloansparger:
@paoliniluis I've fixed the auto-following feature so that should solve your issue with needing to scroll.

If we reverse the order and prepend results we wouldn't have able to preserve the users scroll position as displaying new logs would change the position of all logs. Appending will just make the list longer but not change the relative position of older logs, allowing us to preserve the user's scroll position.

Let me follow up on the copy button. I think that's an obviously great addition, but since logs can be large I need to lazy compute a copyable list. It's not hard to do but our current copy button doesn't allow for this making it a bit more work.

frontend/src/metabase/admin/tasks/components/Logs/Logs.tsx
```diff
@@ -1,165 +1,68 @@
...
-
-function logEventKey(ev: Log) {
-  return `${ev.timestamp}, ${ev.process_uuid}, ${ev.fqns}, ${ev.msg}`;
+import { usePollingLogsQuery, useTailLogs } from "./hooks";
+import { filterLogs, formatLog, getAllProcessUUIDs } from "./utils";
+
+interface LogsProps {
+  // NOTE: fetching logs could come back from any machine if there's multiple machines backing a MB isntance
+  // make this frequent enough that you will most likely get every log from every machine in some reasonable
+  // amount of time
```
iethree, line 16: 😁 good comment. this is not obvious.

frontend/src/metabase/admin/tasks/components/Logs/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  function memoizedFn(input: T) {
+    if (prevInput === input) {
+      return prevOutput;
+    }
+    prevInput = input;
+    prevOutput = fn(input);
+    return prevOutput;
+  }
+  return memoizedFn;
+}
```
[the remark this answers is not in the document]
iethree: I think this will work, right? https://underscorejs.org/#memoize

iethree (approved):
👏  very nice

PR #39038, opened 2024-02-22
TableDefinition can define a custom PK

qnkhuat:
Currently if you define a dataset, all tables will be automatically created with an ID PK column.
This extends the tabledefinition so that you can specify a custom PK.

```clojure
(mt/defdataset  custom-pk
  ["user"
   [{:field-name "uuid" :base-type :type/UUID :pk? true}]
   [[#uuid "201014e3-81bc-4349-a64f-d5519571297b"]]]
  ["group"
   [{:field-name "user_uuid" :base-type :type/UUID :fk "user"}]
   [[#uuid "201014e3-81bc-4349-a64f-d5519571297b"]]])
```
now will create table with
```sql
CREATE TABLE "USER" ("UUID" VARCHAR, PRIMARY KEY ("UUID")) ;
CREATE TABLE "GROUP" ("USER_UUID" VARCHAR, "ID" BIGINT AUTO_INCREMENT, PRIMARY KEY ("ID")) ;
ALTER TABLE "GROUP" ADD CONSTRAINT "fk_group_user_uuid" FOREIGN KEY ("USER_UUID") REFERENCES "USER" ("UUID");
```

I also made `with-temp-test-data` able to take multiple tables, currently you can only create a dataset with a single table with it.


test/metabase/actions/test_util.clj
```diff
@@ -103,11 +103,18 @@
...
+                           [{:field-name \"name\" :base-type :type/Text}]
+                           [[\"Tesla Model S\"]]]
+                          [\"rating\"
+                           [{:field-name \"score\" :base-type :type/Integer}]
+                           [[5]]]]
+      ...)"
   {:style/indent :defn}
-  [dataset-definition & body]
-  `(do-with-dataset-definition (tx/dataset-definition ~(str (gensym)) ~dataset-definition) (fn [] ~@body)))
+  [table-definitions & body]
```
crisptrutski, line 116: It looks like all the non test usages have a single table definition, is it not worth keeping a simple version, rather than wrapping all the call sites in an extra vector?

Given that you've changed from dataset to table, perhaps you could have `with-temp-test-table`, which delegates to the more general `with-temp-test-tables`?

test/metabase/test/data/dataset_definition_test.clj
```diff
@@ -0,0 +1,32 @@
...
+                  ["user"
+                   [{:field-name "uuid" :base-type :type/UUID :pk? true}]
+                   [[#uuid "201014e3-81bc-4349-a64f-d5519571297b"]]]
+                  ["group"
+                   [{:field-name "user_uuid" :base-type :type/UUID :fk "user"}]
+                   [[#uuid "201014e3-81bc-4349-a64f-d5519571297b"]]])
+      (let [user-fields  (t2/select [:model/Field :name :semantic_type :fk_target_field_id] :table_id (mt/id :user))
+            group-fields (t2/select [:model/Field :name :semantic_type :fk_target_field_id] :table_id (mt/id :group))]
+        (testing "user.uuid is PK"
+          (is (= [{:name               (ddl.i/format-name driver/*driver* "uuid")
```
crisptrutski: nit I would break out a `format-name` helper function which closes over the driver in the let block above, as you call this 3 times.

test/metabase/test/data/sql.clj
```diff
@@ -272,9 +271,17 @@
...
-  [driver {:keys [database-name]} {:keys [table-name]} {dest-table-name :fk, field-name :field-name}]
-  (let [quot            #(sql.u/quote-name driver %1 (ddl.i/format-name driver %2))
-        dest-table-name (name dest-table-name)]
+  [driver {:keys [database-name] :as dbdef} {:keys [table-name]} {dest-table-name :fk, field-name :field-name}]
+  (let [quot               #(sql.u/quote-name driver %1 (ddl.i/format-name driver %2))
+        dest-table-name    (name dest-table-name)
+        dest-table-pk-name (->> dbdef
+                                :table-definitions
+                                (filter #(= (:table-name %) dest-table-name))
+                                first
```
crisptrutski, line 280: A `(get-table table-name table-definitions)` method would be nice to trim the density here.

test/metabase/test/data/sql/ddl.clj
```diff
@@ -39,31 +39,46 @@
   tx/dispatch-on-driver-with-test-extensions
   :hierarchy #'driver/hierarchy)
 
+(defn- add-pk-if-needed
+  [driver {:keys [field-definitions] :as tabledef}]
+  (if (not (some :pk? field-definitions))
+    (update tabledef :field-definitions #(cons (tx/map->FieldDefinition
```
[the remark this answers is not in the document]
qnkhuat: yea, sadly it is, some tests have "ID" as first column :(

test/metabase/actions/test_util.clj
qnkhuat: I actually need it for a downstream PR: https://github.com/metabase/metabase/pull/38946/files#diff-6598413869c2c34f4fcc6b36e13f5296ab7a822f446240bb82289fc979fbe1bbR215

> Given that you've changed from dataset to table, perhaps you could have with-temp-test-table, which delegates to the more general with-temp-test-tables?

well, I think the original version of this is already weird to start with, I think it should work with multiple in the first place since it called `test-data`, not `test-table`, and a `dataset` definition should have multiple tables.

Like what we have in `mt/with-actions-test-data` or `mt/dataset-definition`

test/metabase/test/data/sql.clj
qnkhuat: good point, created a `get-tabledef` here [98b3216](https://github.com/metabase/metabase/pull/39038/commits/98b3216404e50c36c56ccb53a61f6255b39245fe)

test/metabase/test/data/dataset_definition_test.clj
qnkhuat: did that [98b3216](https://github.com/metabase/metabase/pull/39038/commits/98b3216404e50c36c56ccb53a61f6255b39245fe)

test/metabase/test/data/sql/ddl.clj
```diff
@@ -39,31 +39,46 @@
...
-        (add! (sql.tx/add-fk-sql driver dbdef tabledef fielddef)))
-      ;; Add the SQL for creating index
-      (when indexed?
-        (add! (sql.tx/create-index-sql driver (:table-name tabledef) [(:field-name fielddef)])))
-      ;; Add the SQL for adding column comments
-      (when (:field-comment fielddef)
-        (add! (sql.tx/standalone-column-comment-sql driver dbdef tabledef fielddef))))
-    @statements))
+  (let [{:keys [table-definitions]
+         :as   dbdef}             (update dbdef :table-definitions #(map (fn [table-def] (add-pk-if-needed driver table-def)) %))
```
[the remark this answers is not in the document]
qnkhuat: oh I meant to remove the db-def and table-definitions at lower lines, lmk what you think about the new format: [98b3216](https://github.com/metabase/metabase/pull/39038/commits/98b3216404e50c36c56ccb53a61f6255b39245fe)

test/metabase/test/data/sql.clj
```diff
@@ -272,9 +271,17 @@
...
+                                :field-definitions
+                                (filter :pk?)
+                                first
+                                :field-name)]
```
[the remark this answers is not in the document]
qnkhuat: done [98b3216](https://github.com/metabase/metabase/pull/39038/commits/98b3216404e50c36c56ccb53a61f6255b39245fe)

test/metabase/test/data/sql/ddl.clj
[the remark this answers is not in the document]
crisptrutski: Oh well 😢 

Breaking out a constructor method for the field and using `partial` could help a bit with nesting.

```clojure
(defn default-pk-field [driver]
  {:field-name ...})

...

    (update tabledef :field-definitions (partial cons (default-pk-field driver)))
```

test/metabase/test/data/sql/ddl.clj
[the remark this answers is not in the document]
crisptrutski: Ah that makes sense, I didn't read closely enough to spot the duplication.

I think a little helper method `(add-pks-if-needed [driver table-definitions)` would be nice to help with the line density then.

crisptrutski (commented):
A bunch of the drivers have failing tests now, where they can't create the PK, e.g.

```
ERROR in metabase.test.data.dataset-definition-test/dataset-with-custom-pk-test (get_or_create.clj:159)

         clojure.lang.ExceptionInfo: Failed to create :oracle 'custom-pk' test database: Error loading data: INSERT FAILED: ORA-00904: "id": invalid identifier
```

or even create a table at all:

```
[CLI-agent-send-off-pool-17] FATAL metabase.test.data.athena - Athena database creation is disabled: not creating database "custom_pk". Tests will likely fail.
See metabase.test.data.athena/*allow-database-creation* for more info.

ERROR in metabase.test.data.dataset-definition-test/dataset-with-custom-pk-test (impl.clj:177)

java.lang.Exception: No Table "user" found for :athena Database 498 "custom-pk".
```

test/metabase/test/data/dataset_definition_test.clj
```diff
@@ -23,11 +23,12 @@
...
                  user-fields)))
-        (testing "user_custom_id is a FK non user.custom_id"
-          (is (= #{{:name               (format-name "user_custom_id")
-                    :fk_target_field_id (mt/id :user :custom_id)
-                    :semantic_type      :type/FK}
-                   {:name               (format-name "id")
-                    :fk_target_field_id nil
-                    :semantic_type      :type/PK}}
-                 (set group-fields))))))))
+        (when-not (#{:sqlite} driver/*driver*) ;; our implement does not support adding fk for custom dataset yet
```
crisptrutski, line 26: 😢

test/metabase/test/data/interface.clj
```diff
@@ -61,6 +61,7 @@
     ;; default is nullable
    [:not-null?         {:optional true} [:maybe :boolean]]
    [:unique?           {:optional true} [:maybe :boolean]]
+   [:pk?               {:optional true} [:maybe :boolean]]
```
calherries, line 64: I disagree with this design, primary keys should be a property of a table and not a property of a field. One practical limitation of this approach is that it cannot support ordered composite primary keys.

That's why for uploads, I changed the design so that the primary key is defined outside of the `column-definitions` of the `create-table!` args
See https://github.com/metabase/metabase/blob/3f351e4ec4b72060743f62277b980646efc58d4a/src/metabase/driver.clj#L919-L925

qnkhuat: yea, but I'm not sure about the practicality of that. Do you think we need to create composite PKs that often?

calherries: Not often, but when we do, we'll either need to make the design more complicated by allowing two places to define primary keys. It would be nicer if it were in one place. I'm just wary that we'll want to change this eventually (actually now, since I want composite PKs for sync tests).

On the other hand, switching from defining the PK at the field level to defining it at the table-level does not seem so difficult in the future, given the number of tests for which we define a custom PK will probably be small.

test/metabase/test/data/dataset_definition_test.clj
```diff
@@ -0,0 +1,43 @@
...
+  (mt/test-drivers (disj (mt/sql-jdbc-drivers)
+                         ;; presto doesn't create PK for test data :) not sure why
+                         :presto-jdbc
+                         ;; creating db for athena is expensive and require some extra steps,
+                         ;; so it's not worth testing against, see [[metabase.test.data.athena/*allow-database-creation*]]
+                         :athena
+                         ;; there is no PK in sparksql
+                         :sparksql
+                         ;; redshift has some problem with creating db in the same schema
+                         :redshift)
```
calherries, line 19: Can you explain why you excluded redshift?

modules/drivers/athena/test/metabase/test/data/athena.clj
```diff
@@ -128,11 +128,8 @@
 ;; Customize the create table table to include the S3 location
 ;; TODO: Specify a unique location each time
 (defmethod sql.tx/create-table-sql :athena
-  [driver {:keys [database-name]} {:keys [table-name], :as tabledef}]
-  (let [field-definitions (cons
-                           {:field-name "id", :base-type :type/Integer, :semantic-type :type/PK}
-                           (:field-definitions tabledef))
-        fields            (->> field-definitions
+  [driver {:keys [database-name]} {:keys [table-name field-definitions], :as _tabledef}]
+  (let [fields            (->> field-definitions
```
calherries, line 132:
```suggestion
  (let [fields (->> field-definitions
```

test/metabase/actions/test_util.clj
```diff
@@ -103,11 +103,18 @@
...
+  `(do-with-dataset-definition (tx/dataset-definition ~(str (gensym)) ~@table-definitions) (fn [] ~@body)))
```
calherries, line 117: fyi my version uses `apply` so the `table-definitions` can be calculated without needing to be a literal list

calherries (approved):
Works for me, let's get this in

qnkhuat:
backport because I'll need this in a bugfix [PR](https://github.com/metabase/metabase/pull/38946) which I'll backport as well

PR #39045, opened 2024-02-22
Mitigate BE flakes caused by redshift-lbv-sync-error-test and redshift-types-test

calherries:
This PR hopefully mitigates flakes on redshift in parallel tests.

I've seen a few flakes like [this](https://github.com/metabase/metabase/actions/runs/7991986113/job/21853226941?pr=39011#step:3:991) where the views / tables that are created in these tests are being picked up in other tests.

See this error output:

```
ERROR in metabase.query-processor-test.date-bucketing-test/filter-by-expression-time-interval-test (get_or_create.clj:159)

                clojure.lang.ExceptionInfo: Failed to create :redshift 'checkins_interval_86400' test database: Failed to sync test database "checkins_interval_86400": Error in sync step Sync redshift Database 458 ''checkins_interval_86400'': Error in sync step Sync metadata for redshift Database 458 ''checkins_interval_86400'': ERROR: relation "2024_02_22_f82431ee_73f5_4652_a3f6_02a8caad6c85_schema.late_binding_view" does not exist
```

I suspect the problem is a race condition with parallel tests. This PR shortens the time it takes for the test to run, hopefully decreasing the likelihood that another parallel test will collide with it. After I'm done with escalations I'll circle back and fix the race conditions.

### DIFF
--- a/modules/drivers/redshift/test/metabase/driver/redshift_test.clj
+++ b/modules/drivers/redshift/test/metabase/driver/redshift_test.clj
@@ -235,7 +235,7 @@
              qual-tbl-nm
              qual-view-nm)
             ;; sync the schema again to pick up the new view (and table, though we aren't checking that)
-            (sync/sync-database! database)
+            (sync/sync-database! database {:scan :schema})
             (is (contains?
                  (t2/select-fn-set :name Table :db_id (u/the-id database)) ; the new view should have been synced
                  view-nm))
@@ -269,7 +269,7 @@
                   "CASE WHEN shop_status = 'open' THEN 11387.133 END AS case_when_numeric_inc_nulls "
                   "FROM test_data) WITH NO SCHEMA BINDING;")
              qual-view-nm)
-            (sync/sync-database! database)
+            (sync/sync-database! database {:scan :schema})
             (is (contains?
                  (t2/select-fn-set :name Table :db_id (u/the-id database)) ; the new view should have been synced without errors
                  view-nm))


modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -217,71 +217,68 @@
...
-      (let [db-details   (tx/dbdef->connection-details :redshift nil nil)
-            tbl-nm       "redshift_specific_types"
+      (let [tbl-nm       "redshift_specific_types"
             qual-tbl-nm  (format "\"%s\".\"%s\"" (redshift.test/unique-session-schema) tbl-nm)
             view-nm      "late_binding_view"
-            qual-view-nm (format "\"%s\".\"%s\"" (redshift.test/unique-session-schema) view-nm)]
-        (t2.with-temp/with-temp [Database database {:engine :redshift, :details db-details}]
-          (try
+            qual-view-nm (format "\"%s\".\"%s\"" (redshift.test/unique-session-schema) view-nm)
+            database     (mt/db)]
```
calherries, line 226: We don't try to create a new database, we just use the `test-data` database. This is for speed (syncing an existing database is faster than syncing a new one) but also so we can use its connection pool.

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -217,71 +217,68 @@
...
+                "CREATE OR REPLACE VIEW %2$s AS SELECT * FROM %1$s WITH NO SCHEMA BINDING;")
+           qual-tbl-nm
+           qual-view-nm)
             ;; sync the schema again to pick up the new view (and table, though we aren't checking that)
-            (sync/sync-database! database)
-            (is (contains?
-                 (t2/select-fn-set :name Table :db_id (u/the-id database)) ; the new view should have been synced
-                 view-nm))
-            (let [table-id (t2/select-one-pk Table :db_id (u/the-id database), :name view-nm)]
+          (sync/sync-database! database {:scan :schema})
```
calherries, line 236: Unrelated to the flake, but doing a full sync is unnecessary work. We just need to sync the schema here

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -217,71 +217,68 @@
...
+        (try
+          (execute!
+           (str "CREATE OR REPLACE VIEW %1$s AS ("
+                "WITH test_data AS (SELECT 'open' AS shop_status UNION ALL SELECT 'closed' AS shop_status) "
+                "SELECT NULL as raw_null, "
+                "'hello' as raw_var, "
+                "CASE WHEN shop_status = 'open' THEN 11387.133 END AS case_when_numeric_inc_nulls "
+                "FROM test_data) WITH NO SCHEMA BINDING;")
+           qual-view-nm)
+          (sync/sync-database! database {:scan :schema})
```
calherries, line 269: Same as above https://github.com/metabase/metabase/pull/39045/files#r1498922726

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -206,7 +206,7 @@
 
 (defn- execute! [format-string & args]
   (let [sql  (apply format format-string args)
-        spec (sql-jdbc.conn/connection-details->spec :redshift @redshift.test/db-connection-details)]
+        spec (sql-jdbc.conn/db->pooled-connection-spec (mt/db))]
```
calherries, line 209: We now get a pooled connection spec on the test database

sansolovyov (approved):
👌

PR #39056, opened 2024-02-22
FE - Migrate `dataset: true` to enum value in Bookmarks

### LINKED ISSUE #38807: FE - Migrate `dataset: true` to enum value in Bookmarks
Depends on #37367

See https://github.com/metabase/metabase/blob/8b19867/frontend/src/metabase-types/api/bookmark.ts#L13

:warning: please remove `card.dataset` usage in `getIcon` in `frontend/src/metabase/entities/questions.js` as a part of this task (see https://github.com/metabase/metabase/pull/38714/commits/a1204874e8785926cf181d58ced9c9b61bfdaa62)

kamilmielnik:
Closes #38807

Depends on #38981

### Description

The [target PR](https://github.com/metabase/metabase/pull/38981) has some failing tests, that's why this PR does not pass CI 100%.
However, this PR should make `E2E Tests / e2e-tests-organization-ee` suite pass.

### DIFF
--- a/frontend/src/metabase/entities/bookmarks.js
+++ b/frontend/src/metabase/entities/bookmarks.js
@@ -48,12 +48,12 @@ const Bookmarks = createEntity({
   },
   reducer: (state = {}, { type, payload, error }) => {
     if (type === Questions.actionTypes.UPDATE && payload?.object) {
-      const { archived, dataset, id, name } = payload.object;
+      const { archived, card_type, id, name } = payload.object;
       const key = `card-${id}`;
       if (archived) {
         return dissoc(state, key);
       } else {
-        return updateIn(state, [key], item => ({ ...item, dataset, name }));
+        return updateIn(state, [key], item => ({ ...item, card_type, name }));
       }
     }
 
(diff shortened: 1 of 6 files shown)


frontend/src/metabase/entities/bookmarks.js
```diff
@@ -109,11 +109,24 @@ function getEntityFor(type) {
...
+    return bookmarkEntity.objectSelectors.getIcon({
+      ...bookmark,
+      /**
+       * Questions.objectSelectors.getIcon works with Card instances.
+       * In order to reuse it we need to map Bookmark["card_type"] to Card["type"]
+       * because Bookmark["type"] is something else.
+       */
+      type: bookmark.card_type,
+    });
+  }
```
kamilmielnik, line 123: I thought it would be simple to refactor but it's a bit tricky because of plugins (in questions and collections) and because the Bookmark does not fully fit into shape of Collection or Card.

For example: `getCollectionIcon` uses `collection.id` to determine the icon, but in bookmarks this information is in `bookmark.item_id` - so there is at least 1 bug there.

I'd rather do this simple change here to keep things working exactly like before, and handle the icon computation in a separate PR.

ranquild (approved):

PR #39062, opened 2024-02-22
Backport Extend param substitution to handle datetimes to 48 (#38695)

calherries:
Manually backports: https://github.com/metabase/metabase/pull/38695

Also cherry-picks https://github.com/metabase/metabase/pull/37813 which is a necessary prerequisite for the fix.

I manually tested this fixes https://github.com/metabase/metabase/issues/38037 using the repro steps here https://github.com/metabase/metabase/issues/38037#issuecomment-1931983021

### DIFF
--- a/docs/developers-guide/driver-changelog.md
+++ b/docs/developers-guide/driver-changelog.md
@@ -4,6 +4,12 @@ title: Driver interface changelog
 
 # Driver Interface Changelog
 
+## Metabase 0.48.7
+
+- The method `metabase.driver.sql.parameters.substitution/align-temporal-unit-with-param-type` is now deprecated.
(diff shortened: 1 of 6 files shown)


snoe (approved):
LGTM do we need to update `docs/developer-guide/driver-changelog.md` for `align-temporal-unit-with-param-type-and-value`?

calherries:
@snoe I wondered about that. I'm not sure what to do when backporting driver changes, because we only have a section for the 48.0 release. This is just a warning that we're deprecating the multimethod so I think it's fine?

calherries:
Looks like there are more failures

calherries:
Looks like the entry needs to be made in the changelog for the tests to pass. I'll add it then

PR #39063, opened 2024-02-22
Set `native_permissions` field correctly on DB

johnswanson:
This was my mistake - we're treating the response of `database-permission-for-user` as if it's the boolean "does this user have permission", but it's actually the keyword representing the permission level the user has.


noahmoss:
I wonder if we should rename that function to make it clearer?

noahmoss (approved):

PR #39067, opened 2024-02-22
add support for networking errors in the token validation endpoint

npretto:
Part of [[Epic] Add license activation in the initial setup](https://github.com/metabase/metabase/issues/38867)

Follow up to  [adds /api/setup/token-check and add license-token to POST /api/setup](https://github.com/metabase/metabase/pull/38858) to allow us to tell between invalid token and an error while trying to validate it.

This changes the backend to return `error_code: "unable_to_validate"` when there's a network error or it's impossible to validate the token for other reasons.

The FE will display a custom error ([design on figma](https://www.figma.com/file/UFay4YjWyUMyl23T1cXzeC/Meet-first-time-embedders-at-the-door-with-a-dashboard?type=design&node-id=1757-9917&mode=design&t=LRD239xZS9jD37p7-4))

Note: this is stacked on top of the [frontend PR](https://github.com/metabase/metabase/pull/38953) which is still waiting review

### DIFF
--- a/frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenForm.tsx
+++ b/frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenForm.tsx
@@ -17,7 +17,7 @@ export const LicenseTokenForm = ({
 }: LicenseTokenFormProps) => {
   const [token, setToken] = useState(initialValue);
   const [status, setStatus] = useState<
-    "idle" | "loading" | "success" | "error"
+    "idle" | "loading" | "success" | "invalid_token" | "unable_to_validate"
   >("idle");
 
   const isInputCorrectLength = token.length === 64;
@@ -29,11 +29,13 @@ export const LicenseTokenForm = ({
       if (response.valid) {
         setStatus("success");
         onValidSubmit(token);
+      } else if (response.error_code === "unable_to_validate") {
+        setStatus("unable_to_validate");
       } else {
-        setStatus("error");
+        setStatus("invalid_token");
       }
     } catch (e) {
-      setStatus("error");
+      setStatus("unable_to_validate");
     }
   };
 
@@ -50,10 +52,16 @@ export const LicenseTokenForm = ({
               oldState !== "loading" ? "idle" : "loading",
             );
           }}
-          error={status === "error"}
+          error={status === "invalid_token"}
         />
-        {status === "error" && (
+        {status === "invalid_token" && (
           <Text color="error">{t`This token doesn’t seem to be valid. Double-check it, then contact support if you think it should be working`}</Text>
+        )}
+        {status === "unable_to_validate" && (
+          <>
+            <Text color="error">{t`We couldn’t connect to our servers to activate the license. Please try again.`}</Text>
+            <Text color="error">{t`You can also set this up at a later time in settings.`}</Text>
+          </>
         )}
       </Box>
       <Flex gap="lg">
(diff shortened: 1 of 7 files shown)


test/metabase/public_settings/premium_features_test.clj
```diff
@@ -58,6 +58,7 @@
                                  ;; throw an ex-info here
                                  (throw (Exception. "network issues")))]
           (is (= {:valid         false
+                  :error_code    :unable_to_validate
```
WiNloSt, line 61: It feels a bit weird we need this particular `error_code` to make sure we can differentiate between `invalid_token` and `unable_to_validate` errors.

In FE, we don't even need to check `error_code` we just say anything that's not `unable_to_valiadate`, what if we say if the status is `invalid` then it's `invalid_token` and `unable_to_validate` otherwise? We wouldn't need the BE change would we?

npretto: You mean this status https://github.com/metabase/metabase/blob/46feb24a931642958c5ce86e84c7a95f33e3b5cb/src/metabase/public_settings/premium_features.clj#L131 ?

Isn't it sent only when the token is an invalid format (ie, not 64 chars)?

When the BE is not able to validate it, it still sends the status but as a translated string: https://github.com/metabase/metabase/blob/46feb24a931642958c5ce86e84c7a95f33e3b5cb/src/metabase/public_settings/premium_features.clj#L155-L156 so we can't really use it in an if.

WiNloSt: yeah we could say
```ts
// if valid === false
if (status === "invalid")
  errorState = "invalid_token"
else {
  errorState = "unabled_to_validate"
```

Couldn't we? Or did I missed something.

npretto: Status is "invalid" [only when the token is not formatted correctly](https://github.com/metabase/metabase/blob/46feb24a931642958c5ce86e84c7a95f33e3b5cb/src/metabase/public_settings/premium_features.clj#L129-L132), when the token is of the correct size, but doesn't exist we get a different output.

Here's some sample responses of `premium-features/fetch-token-status`:

invalid 64 chars token:
```json
{
    "valid": false,
    "status": "Token does not exist."
}
``` 

token with wrong format
```json
{
  "valid": false,
  "status": "invalid",
  "error-details": "Token should be 64 hexadecimal characters."
}
```

valid token
```json
{
    "valid": true,
    "status": "Token is valid.",
    "trial": true,
    "valid-thru": ...,
    "features": ...,
    "store-users": ...
}
```
network error (I turned off wifi to simulate)
```json
{
    "valid": false,
    "error_code": "unable_to_validate", // added with this pr
    "status": "Unable to validate token",
    "error-details": "token-check.metabase.com: nodename nor servname provided, or not known"
}
```

The issue is that `status` is sometimes a translated message, and sometimes "invalid", and when it's a message it's not probably user friendly enough for the initial setup

WiNloSt: Good point. I didn't see we have another case for `Token does not exist.`

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenForm.tsx
```diff
@@ -39,11 +41,17 @@ export const LicenseTokenForm = ({ onValidSubmit }: LicenseTokenFormProps) => {
           placeholder={t`Paste your token here`}
           value={token}
           onChange={e => setToken(e.target.value)}
-          error={status === "error"}
+          error={status === "invalid_token"}
```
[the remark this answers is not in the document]
npretto: Nope that was my mistake, fixed in ecafba0dc5fc77246422136691c2902f721a4edc

npretto:
Closing as this will be worked on from the webapp team and the endpoint used here will be removed after the api/auth split in the setup

PR #39070, opened 2024-02-22
Clean QueryBuilder props [PoC]

nemanjaglumac:
This PR is just a PoC, and a preview of what we could (should?) do with the QueryBuilder.
It currently fetches way more data than it needs, and stores it in props . This leads to prop drilling that's hard to track.

We can identify where some of these props are used and move the fetching logic closer to the place it's needed.
By doing this examination, we might also find some completely unused props, like it happened [here](https://github.com/metabase/metabase/pull/39070/commits/ac20ea224d784ba48aac66204afd161b1189721a).


kamilmielnik (commented):
I think this is the right way forward :+1: 

It'd be much easier to do if query builder was 100% TS.

It looks like we're at ~85%.

```
kamil@desktop:~/projects/metabase$ find frontend/src/metabase/query_builder/ | grep \.js | wc -l
86
kamil@desktop:~/projects/metabase$ find frontend/src/metabase/query_builder/ | grep \.ts | wc -l
567
```

ranquild (approved):

PR #31891, opened 2023-06-27
Allow native questions with filters with default values to be cleared

calherries:
Resolves https://github.com/metabase/metabase/issues/13961, and also makes updating the default value of a parameter in the native query editor also update the value of the parameter.

### Summary of changes:
**FE:** 
- In dashboards and the native query editor we now distinguish between:
  1. params that are in the URL query params but have an empty value, e.g. `"...?foo="` (meaning, empty value)
  2. params that don't appear in the URL query params at all (possibly getting a default value, if it has one)
- Previously both cases were treated the same in that the default value would be applied, if it had one. For example if there was a default value set for the parameter "foo", then navigating to `".../question/<card-slug>"` or `".../question/<card-slug>?foo="` would automatically update the URL to `".../question/<card-slug>?foo=<default>"` with the param/filter having the default value. After this change, only in case (ii) will the default value be applied. So in the previous example, navigating to `".../question/<card-slug>"` will update the URL to `".../question/<card-slug>?foo=<default>"` and apply the default, but navigating to `".../question/<card-slug>?foo="` would keep the URL and the parameter value would be empty.
- After opening a dashboard or a native query, every filter / parameter will appear in the URL. For example navigating to `".../question/<card-slug>"` where the native query has "foo" with a default and "bar" without a default, the URL will update to  `".../question/<card-slug>?foo=<default>&bar="`. The same logic applies to dashboards and their filters.
- When the default value of a parameter is updated in the native query editor, we now update the corresponding value in the filter, even if there is already one present. See this [slack context](https://metaboat.slack.com/archives/C057T1QTB3L/p1688486429475699) for the context behind the decision.

New e2e tests were added for:
- Dashboards
- Native questions
- Embedded dashboards
- Embedded native questions

**BE:** 
- Updated `param-value-for-raw-value-tag` (for non field-filter params) and `field-filter-value` (for field-filter params) in `src/metabase/driver/common/parameters/values.clj` to not use default values when the value of a parameter is nil. See the tests in `test/metabase/driver/common/parameters/values_test.clj`.

The BE tests cover:
- API tests:
  - GET /api/public/card/:uuid/query
  - GET /api/public/dashboard/:uuid/card/:card-id
  - GET /api/embed/dashboard/:token/dashcard/:dashcard-id/card/:card-id
  - GET /api/embed/card/:token/query
  - GET /api/preview_embed/card/:token/query
  - POST /api/dashboard/:dashboard-id/card/:card-id/query
- Query processor tests:
[...]

PR #39072, opened 2024-02-22
Fix applying default values for a pinned question

### LINKED ISSUE #37831: Default values are not respected for pinned questions in collections in v48
### Describe the bug

When you add default values to the filters (normal or field filters) on a native questions, those are not taken into account in v48

### To Reproduce

1) new question -> select * from people where {{something}}
2) something is a field filter mapped to the state field, it's of type "location is"
3) add a default value, like NY
4) select a state, run the question, it will run
5) now remove the state from the filter (not from the default value). Run the question, see it failing

### Expected behavior

It should run with the default filter

### Logs

NA

```JSON
v48.x
```

### Severity

P2

### Additional context

Pretty sure this worked in v47

uladzimirdev:
should fix #37831 in v48, but not in master

### Description

https://github.com/metabase/metabase/pull/31891 changed the way of preparing parameters for BE - we started adding `{value: null}` if value wasn't presented.

> The point of the change is so we can differentiate between value: null (which means the no value should be applied even if it has a default) and the absence of a value key (which means we should apply the default value)

That comment made me think there could be a problem of handling cases like this on BE.

This PR adds a check whether it's collection preview or not, uses old way of calculating parameters (with value: null) for collection preview and the way introduced in #31891 for other cases.

Code in master has changed comparing to v48, so we need to provide a slightly different solution there

### DIFF
--- a/e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
+++ b/e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
@@ -13,7 +13,7 @@ import {
   ORDERS_DASHBOARD_ID,
 } from "e2e/support/cypress_sample_instance_data";
 
-const { ORDERS, ORDERS_ID } = SAMPLE_DATABASE;
+const { ORDERS, ORDERS_ID, PEOPLE } = SAMPLE_DATABASE;
 
 const DASHBOARD_NAME = "Orders in a dashboard";
 const QUESTION_NAME = "Orders, Count";
@@ -66,7 +66,7 @@ const SQL_QUESTION_DETAILS_WITH_DEFAULT_VALUE = {
         id: "4b77cc1f-ea70-4ef6-84db-58432fce6928",
         "display-name": "date",
         default: "1999-02-26~2024-02-26",
-        dimension: ["field", 35 /* PEOPLE.BIRTH_DATE */, null],
+        dimension: ["field", PEOPLE.BIRTH_DATE, null],
         "widget-type": "date/range",
       },
     },
--- a/e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
+++ b/e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
@@ -66,7 +66,7 @@ const SQL_QUESTION_DETAILS_WITH_DEFAULT_VALUE = {
         id: "4b77cc1f-ea70-4ef6-84db-58432fce6928",
         "display-name": "date",
         default: "1999-02-26~2024-02-26",
-        dimension: ["field", 18 /* PEOPLE.BIRTH_DATE */, null],
+        dimension: ["field", 35 /* PEOPLE.BIRTH_DATE */, null],
         "widget-type": "date/range",
       },
     },
--- a/e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
+++ b/e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
@@ -55,6 +55,25 @@ const SQL_QUESTION_DETAILS = {
   },
 };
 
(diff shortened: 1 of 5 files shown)


calherries:
As discussed, this is a FE problem, the BE is behaving as designed. I've added a BE test [here](https://github.com/metabase/metabase/pull/39106) to document behaviour of the `POST /api/card/:id/query` endpoint in case this comes up again.

frontend/src/metabase-lib/Question.ts
```diff
@@ -1090,12 +1090,13 @@ class Question {
...
 
-  // TODO: Fix incorrect Flow signature
-  parameters(): ParameterObject[] {
+  parameters({ collectionPreview } = {}): ParameterObject[] {
     return getCardUiParameters(
       this.card(),
       this.metadata(),
       this._parameterValues,
+      undefined,
+      collectionPreview,
```
oleggromov, line 1099: Not a big fan of comments like this but this function has become ugly. Let's probably use named parameters?

frontend/src/metabase-lib/parameters/utils/parameter-values.js
```diff
@@ -9,7 +9,27 @@ import {
...
+      ? parameters.map(parameter => {
+          return parameter.id in parameterValues
+            ? {
+                ...parameter,
+                value: parameterValues[parameter.id],
+              }
+            : parameter;
+        })
+      : parameters;
+  }
```
oleggromov, line 31: What I don't understand, is why if we just sent `undefined` instead of `null` as it happens now in [line 35](https://github.com/metabase/metabase/pull/39072/files#diff-8b5188e613c51e0dd17810bc8859b25544494e8b319ecdb428038048fe4b0542R35) the default value isn't picked up by the BE?

@calherries I thought that sending `undefined` means "use default" but here we send empty parameter list instead.

frontend/src/metabase-lib/Question.ts
uladzimirdev: you did similar refactoring in master, but it's not backported to 48 (see the base branch)

uladzimirdev: v48 branch in "support state" atm, so I'll not do this refactoring

frontend/src/metabase-lib/parameters/utils/parameter-values.js
oleggromov: The discussion has moved here: https://github.com/metabase/metabase/pull/39106/files#r1502499549
Not related to or blocking this PR

oleggromov (approved):

e2e/test/scenarios/collections/collection-pinned-overview.cy.spec.js
```diff
@@ -55,6 +55,25 @@ const SQL_QUESTION_DETAILS = {
...
+  display: "scalar",
+  native: {
+    "template-tags": {
+      filter: {
+        type: "dimension",
+        name: "filter",
+        id: "4b77cc1f-ea70-4ef6-84db-58432fce6928",
+        "display-name": "date",
+        default: "1999-02-26~2024-02-26",
+        dimension: ["field", 35 /* PEOPLE.BIRTH_DATE */, null],
```
nemanjaglumac, line 69: Use `ORDERS.BIRTH_DATE` instead.

```js
import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";

const { PEOPLE } = SAMPLE_DATABASE;
```

nemanjaglumac (changes requested):
Please, never ever hard code field IDs in Cypress tests.
These are guaranteed to fail sooner or later.

We've had the mechanism to dynamically and programmatically reference those fields for years. All the data that you need lives in `e2e/support/cypress_sample_database`. It is dynamically generated each time you spin Cypress up.

nemanjaglumac (approved):

PR #35457, opened 2023-11-07
Prototype — Scrollable table dashboard cards

kulyk:
### Demo

https://github.com/metabase/metabase/assets/17258145/4cdfd3cf-08ef-43ae-a93c-fed895b3b37e

PR #39074, opened 2024-02-22
Easy List Virtualization

iethree:
### Description

Inspired by @kulyk's [prototype](https://github.com/metabase/metabase/pull/35457), I used [@tanstack/react-virtual](https://github.com/TanStack/virtual) to make a generic list virtualizer that is easier to use than our current approach, and much easier to use for lists with variable-height items.

This new list virtualizer
- simply virtualizes all of its children (which must be a list of react components) no matter what height they are.
- must have a fixed-height container
- doesn't require any special measuring logic, it's all handled automatically 🪄

As proof of concept, I implemented it in the archive app. While I was there I made a couple little design tweaks:
- Instead of setting the width of the content to 2/3, i gave the items a max-width of 40rem.
- Fixed the header and made the items scroll within their own container
- Removed the tiny bit of horizontal scroll on the page 🔥

Before | After
--- | ---
oldarchive | new archive page

PS, this page is getting a total rework soon, so this is mostly proof of the usability of the virtualization library, which has a few other potential use cases.

### Demo

### Checklist


frontend/src/metabase/components/VirtualizedList/VariableHeightVirtualizedList.unit.spec.tsx
```diff
@@ -0,0 +1,49 @@
+import { screen, render } from "@testing-library/react";
+
+import { VirtualizedList } from "./VariableHeightVirtualizedList";
+
+describe("VariableHeightVirtualizedList", () => {
+  beforeEach(() => {
+    // jsdom doesn't have getBoundingClientRect, so we need to mock it
```
uladzimirdev, line 7: it would be nice to have a cleanup as well

frontend/src/metabase/components/VirtualizedList/VariableHeightVirtualizedList.tsx
```diff
@@ -0,0 +1,68 @@
...
+    getScrollElement: () => parentRef.current,
+    estimateSize: () => estimatedItemSize,
+  });
+
+  const items = virtualizer.getVirtualItems();
+
+  return (
+    <div
+      ref={parentRef}
+      style={{
```
uladzimirdev, line 34: can we have styles defined as a css module and not inlined?

uladzimirdev (commented):
looks great

frontend/src/metabase/components/VirtualizedList/VariableHeightVirtualizedList.tsx
iethree: since the virtualizer has to set some of them programmatically, I thought for this file it would make sense to keep them all inline so you don't have to look elsewhere to figure out what's happening

frontend/src/metabase/components/VirtualizedList/VariableHeightVirtualizedList.unit.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+        // this will make all elements, the container and the children have the same size
+        // so it will render the first 2 children (one beyond visible)
+        // and not render the last 2 children
+        return { height: 100, width: 200 };
+      });
+  });
+
+  afterAll(() => {
+    jest.restoreAllMocks();
+  });
```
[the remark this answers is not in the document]
uladzimirdev: I don't think so, you're rewriting window property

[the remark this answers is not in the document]
npfitz: I think you may need to use `jest.spyOn()` if you want restoreAllMocks to reset the property @iethree 

https://jestjs.io/docs/jest-object#jestrestoreallmocks

npfitz (approved):
Aside from @uladzimirdev's comment on restoring the window prop in tests, this looks good 🔥

PR #39076, opened 2024-02-22
Show filters without values in fullscreen mode

### LINKED ISSUE #35038: Filters without selected values are hidden on the dashboard in full-screen mode
### Describe the bug

Filters without selected values are hidden on the dashboard in full-screen mode. Since they are hidden, I can't choose a value for them.

### To Reproduce

1. Сreate a dashboard
2. Add a query to the dashboard (for example, PEOPLE)
3. Add filters and link them as in the screenshots
4. Select a value for one filter, leave the second one empty
5. Switch to full-screen mode
6. Only filters with the selected value are shown. Is this a bug? I can't select a value for other filters because they are not shown.

ksnip_20231025-114801

ksnip_20231025-114812

ksnip_20231025-114828

ksnip_20231025-114852

### Expected behavior

All filters are shown

### Logs

_No response_

```JSON
The problem continues to exist in version 0.47.5
```

### Severity

To change an empty filter, you need to exit full-screen mode, select the filter value and go back to full-screen mode. Also, the filter cannot be reset.

### Additional context

_No response_

kulyk:
Fixes #35038

### To reproduce

1. Open a dashboard with filters
2. Set a value for a few filters and leave a few more empty
3. Enter fullscreen mode (click "More" at the top right and then "Fullscreen")
4. Ensure you can see all filters (with and without values)

### Demo

full-screen-new

### DIFF
--- a/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.styled.tsx
+++ b/frontend/src/metabase/dashboard/components/Dashboard/Dashboard.styled.tsx
@@ -90,12 +90,18 @@ export const CardsContainer = styled(FullWidthContainer)`
   margin-top: 8px;
 `;
 
+function getParametersWidgetBgColor(isNightMode: boolean) {
+  return isNightMode ? color("bg-black") : color("bg-light");
+}
+
 export const ParametersWidgetContainer = styled(FullWidthContainer)<{
   isSticky: boolean;
   hasScroll: boolean;
+  isNightMode: boolean;
 }>`
-  background-color: ${color("bg-light")};
-  border-bottom: 1px solid ${color("bg-light")};
+  background-color: ${props => getParametersWidgetBgColor(props.isNightMode)};
+  border-bottom: 1px solid
+    ${props => getParametersWidgetBgColor(props.isNightMode)};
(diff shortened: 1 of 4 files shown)


alxnddr (commented):
It would be nice to improve the dark mode a bit:

1) Support dark filter widgets
Screenshot 2024-02-22 at 8 51 58 PM

2) Really minor: add transition to the filters sections
<video src="https://github.com/metabase/metabase/assets/14301985/66e69e98-971e-4dde-8aa4-6d13e2a9c2dd" />

kulyk:
@alxnddr the colors transition should be fixed, please take a look:

https://github.com/metabase/metabase/assets/17258145/788481a9-e82e-48bb-9dbb-a07814d323f7

WRT filter popover colors, I think we're keeping them light because it'd be a pain to maintain without proper dark mode support. For example, the popovers are also light in embedded dashboards :(

CleanShot 2024-02-23 at 15 37 20@2x

EmmadUsmani (approved):
We can file a separate issue for the dark mode in the popover, I think it's worth doing to support it for both fullscreen and embedded dashboards, but it shouldn't block this PR.

PR #39089, opened 2024-02-22
Allow binning field refs from expressions

### LINKED ISSUE #26202: Custom Column in Models doesn't respect filtering when binning is applied
**Describe the bug**
When you create a model having a custom column then you proceed to create a question with this model having filters and binning by this custom column the binning strategy applied doesn't respect the filter and uses the full data within the custom column.

**To Reproduce**
1. Go to New -> Question -> Orders -> Custom Column ( Field Formula =[Total] * 1 and name = custom_tax) -> Save -> Turn into model 
2. Go to Metadata -> Custom_Column -> Set field_type to Quantity
4. Go to New > Question -> Model -> Filter Custom_Column > 20 and < 40 -> Group by custom_column (10 bins) and count -> Result only 1 bin

image

**Expected behavior**
Since for this bin there are more than 2000 values I would expect it to bin these values in 10 or at least some kind of binning.

Version 42 - 44.5 

**Additional context**

If you save the question and want to investigate further, you can get the question metadata by going to the URL /api/card/123, where 123 is the question ID. You can notice that in the binning part it's taking the whole column data and generating the binnig structure on that rather than the result of the filters (min_value should be 20 and max_value should be 40):

image

snoe:
Fixes #26202

Previously we only looked for binned field-ids in order to update binning strategies in qp. With this change we also look for field-name refs. This allows expressions to be binned in models.

### DIFF
--- a/test/metabase/query_processor/middleware/binning_test.clj
+++ b/test/metabase/query_processor/middleware/binning_test.clj
@@ -6,6 +6,7 @@
    [metabase.lib.card :as lib.card]
    [metabase.lib.convert :as lib.convert]
    [metabase.lib.core :as lib]
+   [metabase.lib.metadata :as lib.metadata]
    [metabase.lib.metadata.jvm :as lib.metadata.jvm]
    [metabase.lib.test-metadata :as meta]
    [metabase.lib.test-util :as lib.tu]
@@ -14,6 +15,7 @@
    [metabase.query-processor.middleware.binning :as binning]
    [metabase.query-processor.preprocess :as qp.preprocess]
    [metabase.query-processor.store :as qp.store]
+   [metabase.query-processor.test-util :as qp.test-util]
    [metabase.test :as mt]))
 
 (deftest ^:parallel filter->field-map-test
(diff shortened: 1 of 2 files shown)


test/metabase/query_processor/middleware/binning_test.clj
```diff
@@ -209,3 +216,35 @@
...
+  (testing "fields referencing source expressions can still properly update binning strategies (#26202)"
+    (mt/with-temp [:model/Card card (-> orders
+                                        (mt/mbql-query {:fields [$total [:expression "foo"]]
+                                                        :expressions {"foo" [:+ $total 0]}})
+                                        qp.test-util/card-with-source-metadata-for-query
+                                        (assoc-in [:result_metadata 1 :semantic_type] :type/Quantity))]
+      (let [mp (lib.metadata.jvm/application-database-metadata-provider (mt/id))
+            query (lib/query mp (lib.metadata/card mp (:id card)))
+            expr-col (m/find-first #(= (:name %) "foo") (lib/breakoutable-columns query))
+            _ (is (some? expr-col))
```
metamben, line 230: Could be moved in the body of the `let`.

metamben (approved):
LGTM

PR #39092, opened 2024-02-23
RTKQuery integration: Setup

sloansparger:
see: https://www.notion.so/metabase/Picking-a-real-data-fetching-library-d38077b06b9d4b09ba15db09ad608be0

This PR sets up RTKQuery in our application along with some utils for dealing with RTKQ's cache.


iethree (approved):
❤️  🎉  ❤️

PR #39094, opened 2024-02-23
Migrate `graph.metrics` sync to use QP metadata

ranquild:
Similar to https://github.com/metabase/metabase/pull/38995, migrate the whole viz settings sync process to rely on the QP metadata and not MLv2.

How to verify:
- Create a question with an aggregation and a breakout, e.g. orders -> Count by Created At (month)
- Visualize as a line chart
- Change some viz settings
- Add a new breakout
- A new series in the chart should appear

We already have a lot of tests for this, not adding a new test here.


frontend/src/metabase-lib/Question.ts
```diff
@@ -567,22 +560,17 @@ class Question {
...
     if (queryResults && !queryResults.error) {
       question = question._syncTableColumnSettings(queryResults);
-    }
 
-    if (previousQuestion) {
-      const previousQuery = previousQuestion.query();
-      const { isNative: isPreviousQuestionNative } =
-        Lib.queryDisplayInfo(previousQuery);
-
-      if (!isNative && !isPreviousQuestionNative) {
```
ranquild, line 584: We now check column source to ignore native queries so we don't depend on the query in any way.

uladzimirdev (approved):
👍

PR #39103, opened 2024-02-23
goToNextStep in all thunks

npretto:
This is a refactor on the redux issue described in the description of  [Add the step in the FE and pass the token to the setup api call](https://github.com/metabase/metabase/pull/38953)

It works by reverting the initSetup action and manually dispatching `goToNextStep` in all thunks that should advance the step. Thunks can access the global redux state so we don't have to worry about how to access the token-features anymore.

### DIFF
--- a/frontend/src/metabase-types/store/mocks/setup.ts
+++ b/frontend/src/metabase-types/store/mocks/setup.ts
@@ -41,7 +41,6 @@ export const createMockSubscribeInfo = (
 export const createMockSetupState = (
   opts?: Partial<SetupState>,
 ): SetupState => ({
-  isPaidPlan: false,
   step: "welcome",
   isLocaleLoaded: false,
   isTrackingAllowed: false,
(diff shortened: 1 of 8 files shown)


deniskaber (commented):
Code-wise this looks good. But I don't see any tests affected. Is this flow covered with any tests?

npretto:
> Code-wise this looks good. But I don't see any tests affected. Is this flow covered with any tests?

This is just a refactor to "undo" some hacky thing I did in https://github.com/metabase/metabase/pull/38953 , the tests were added there in the [frontend/src/metabase/setup/tests/](https://github.com/metabase/metabase/pull/38953/files#diff-a3f7418f0f4c9bac5c625b323c01fc2451d730a80272cad90f2ee844f5ab5a03L1) folder + `e2e/test/scenarios/onboarding/setup/setup.cy.spec.ts` for the e2e

deniskaber (approved):

PR #39106, opened 2024-02-23
Add test for POST /api/card/:id/query for parameters with defaults

calherries:
This PR adds a test to clarify the BE behaviour of `POST /api/card/:id/query` when there are defaults applied to parameters. This behaviour was implemented in https://github.com/metabase/metabase/pull/31891, but there were no tests specifically for this endpoint.

The test is mostly redundant because the same code path is used for other endpoints e.g. `POST /api/dashboard/:dashboard-id/card/:card-id/query`, which are tested and the code that does default parameter substitution is tested [here](https://github.com/metabase/metabase/pull/31891/files#diff-ce58525a897b47b32a26ef9067c537a410de2e1e006ef9f6b1ec5443d1dd5bc1). But since it's pretty easy for the FE to use this endpoint incorrectly (see the confusion caused by https://github.com/metabase/metabase/issues/37831), I thought it would be nice to add a test for this endpoint to document and lock down it's behaviour.

### DIFF
--- a/test/metabase/api/card_test.clj
+++ b/test/metabase/api/card_test.clj
@@ -224,10 +224,10 @@
                                                     :value  1}]}))))))
         (testing "the field filter should not apply if the parameter has a nil value"
           (is (= [[100]]
-                 (mt/rows (request {:parameters [{:id "_VENUE_ID_",
-                                                   :target ["dimension" ["template-tag" "venue_id"]],
-                                                   :type "id",
-                                                   :value nil}]})))))))))
+                 (mt/rows (request {:parameters [{:id     "_VENUE_ID_",
+                                                  :target ["dimension" ["template-tag" "venue_id"]],
+                                                  :type   "id",
+                                                  :value  nil}]})))))))))
 
 
 ;;; +----------------------------------------------------------------------------------------------------------------+
--- a/test/metabase/api/card_test.clj
+++ b/test/metabase/api/card_test.clj
@@ -189,6 +189,46 @@
                                 :target [:variable [:template-tag :category]]
                                 :value  2}]})))))))
 
+(deftest run-query-with-default-parameters-test
+  (testing "POST /api/card/:id/query with parameters with default values"
+    (mt/with-temp
+      [:model/Card card {:dataset_query
+                         {:database (mt/id)
+                          :type     :native
+                          :native   {:query         "SELECT count(*) AS Count FROM venues where {{venue_id}}"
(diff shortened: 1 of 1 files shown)


test/metabase/api/card_test.clj
```diff
@@ -189,6 +189,46 @@
...
+                                                                 :id           "_VENUE_ID_",
+                                                                 :name         "venue_id",
+                                                                 :required     false,
+                                                                 :default      1
+                                                                 :type         :dimension,
+                                                                 :widget-type  :id}}}}}]
+      (let [request (fn [body]
+                      (mt/user-http-request :rasta :post 202 (format "card/%d/query" (:id card)) body))]
+        (testing "the default can be overridden"
+          (is (= [[1]]
```
adam-james-v, line 210: Should this be `[[2]]` or am I misunderstanding the test?

test/metabase/api/card_test.clj
```diff
@@ -189,6 +189,46 @@
...
+                 (mt/rows (request {:parameters [{:id    "_VENUE_ID_"
+                                                  :name  "venue_id"
+                                                  :slug  "venue_id"
+                                                  :type  "number"
+                                                  :value 2}]})))))
+        (testing "the default should apply if no param value is provided"
+          (is (= [[1]]
+                 (mt/rows (request {:parameters []}))))
```
oleggromov, line 218: What if parameter is passed but its value is unset?

calherries: `value: null` and omitting the `value` key are equivalent.

oleggromov: Got it, thanks.
Don't we have a convention somewhere that passing `nil` as a value (`null` in json) means "we deliberately want the value to be empty" while making it unset (`undefined` or no property in json) means "take default"?

test/metabase/api/card_test.clj
calherries: Oleg's right but it wasn't a strong test at all. Thanks for pointing this out, I've updated it now.

test/metabase/api/card_test.clj
calherries: I'm pretty sure that was the behaviour before https://github.com/metabase/metabase/pull/31891, but now it's like this. Having a semantic difference between `value: null` and omitting the `value` key is asking for trouble, especially in Clojure where most functions will treat them the same.

calherries:
@adam-james-v can you take another look?

adam-james-v (approved):
Looks good :)

Issue #32822, opened 2023-08-02
Hidden behaviors with `metabase.models.interface/to-json` methods


qnkhuat:
This multimethod is used to control what happens when we render a toucan instance onto API resonse.

There are 2 problems with this:
1. Hidden behaviors
When you write an API
```
(defendpoint GET "/database/the_first_one"
 []
 (t2/select-one :model/Database 1))
 ```

The response got back from calling API does not contain `:details` keys, but the key is there when I call (t2/select-one :model/Database 1)) .

After debugging I know it’s because there is a `defmethod to-json :model/Database` that manually `dissoc :details`. 
But this is hidden and feels unexpected.

What’s worse is that if an exception is thrown in `metabase.models.interface/to-json` , The exception is ignored when calling via API!

=> try modify `(defmethod to-json :model/Database)` to throw an exception and calls the API, the API will returns 200 with empty response.

2. Very easy to work around
The to-json method dispatches on toucan2 instance type. Meaning if I do `(into {} (t2/select-one :model/Database 1))` before returning, I could easily work around this.

Proposed fix: Let’s do this manually by defining functions in API layers (edited)

**Comment by qnkhuat** [comment]
the reason why exception is ignored is properly that the streamed-json-response middleware calls to rui/piped-input-stream ([src](https://github.com/metabase/metabase/blob/master/src/metabase/server/middleware/json.clj#L79)).
And this function write the output in a seperate thread with try/finally wrapped.

```clojure
(defn piped-input-stream
  "Create an input stream from a function that takes an output stream as its
  argument. The function will be executed in a separate thread. The stream
  will be automatically closed after the function finishes.

  For example:

    (piped-input-stream
      (fn [ostream]
        (spit ostream \"Hello\")))"
  {:added "1.1"}
  [func]
  (let [input  (PipedInputStream.)
        output (PipedOutputStream.)]
    (.connect input output)
    (future
      (try
        (func output)
        (finally (.close output))))
    input))
```

qnkhuat:
a good context by @dpsutton https://metaboat.slack.com/archives/CKZEMT1MJ/p1690955485333629?thread_ts=1690952451.508079&cid=CKZEMT1MJ

crisptrutski:
Leaving some notes to self as I start figuring this out. 

Started taking a look at how the code is working today.

[...]

PR #39109, opened 2024-02-23
Avoid add or removing keys when redacting :model/Database

crisptrutski:
References https://github.com/metabase/metabase/issues/32822

### Description

This change updates the redaction logic for `:model/Database` so that we never outright remove a key (`:details`) or set the value to nil (`:settings`), but rather leave a sentinel value to indicate that we have redacted content. These strings are easily searchable to save time debugging where the redaction is happening.

In addition to never removing the keys, we now also avoid adding the keys back in, if for example they were never selected in the first place.

### DIFF
--- a/src/metabase/models/database.clj
+++ b/src/metabase/models/database.clj
@@ -303,12 +303,12 @@
   "**MetabasePass**")
 
 (def ^:const protected-db-details
-  "The string to replace details with when serializing Databases."
-  "**MetabaseDatabaseDetails**")
+  "The map to replace details with when serializing Databases."
+  {:metabase/redacted "db/details"})
 
 (def ^:const protected-db-settings
-  "The string to replace settings with when serializing Databases."
-  "**MetabaseDatabaseSettings**")
+  "The map to replace settings with when serializing Databases."
+  {:metabase/redacted "db/settings"})
 
 (defn sensitive-fields-for-db
   "Gets all sensitive fields that should be redacted in API responses for a given database. Delegates to
@@ -332,7 +332,7 @@
                                       (sensitive-fields-for-db db))))))
 
 (defn- redact-settings [settings]
-  (if-not (map? settings)
+  (if-not (or (nil? settings) (map? settings))
     protected-db-settings
     (m/filter-keys
      (fn [setting-name]
--- a/src/metabase/models/database.clj
+++ b/src/metabase/models/database.clj
@@ -357,7 +357,8 @@
   Also remove settings that the User doesn't have read perms for."
   [db json-generator]
   (next-method
-   (-> db redact-db-details (m/update-existing :settings redact-settings))
+   (-> (redact-db-details db)
+       (m/update-existing :settings redact-settings))
    json-generator))
 
 ;;; ------------------------------------------------ Serialization ----------------------------------------------------
--- a/src/metabase/models/database.clj
+++ b/src/metabase/models/database.clj
@@ -302,6 +302,14 @@
   "The string to replace passwords with when serializing Databases."
   "**MetabasePass**")
 
+(def ^:const protected-db-details
+  "The string to replace details with when serializing Databases."
+  "**MetabaseDatabaseDetails**")
+
+(def ^:const protected-db-settings
+  "The string to replace settings with when serializing Databases."
+  "**MetabaseDatabaseSettings**")
+
 (defn sensitive-fields-for-db
   "Gets all sensitive fields that should be redacted in API responses for a given database. Delegates to
   driver.u/sensitive-fields using the given database's driver (if valid), so refer to that for full details. If a valid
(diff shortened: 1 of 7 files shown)


src/metabase/util.cljc
```diff
@@ -532,22 +532,6 @@
        :cljs (js/Math.floor (/ (js/Math.log (abs x))
                                (js/Math.log 10))))))
 
-(defn update-if-exists
```
crisptrutski, line 535: The medley version of these functions are much more efficient, and we already use them more prevalently.

src/metabase/models/database.clj
```diff
@@ -302,6 +302,14 @@
   "The string to replace passwords with when serializing Databases."
   "**MetabasePass**")
 
+(def ^:const protected-db-details
+  "The map to replace details with when serializing Databases."
+  {:metabase/redacted "db/details"})
```
crisptrutski, line 307: Not so sure about these values. I initially used strings like the password above, but this violated API schemas. I was not inclined to change the schema type, as this could a break. 

At the end of the day, having second thoughts about overloading the "type" of these fields - even when it's still a map like this, it still changes semantics and could cause confusing errors for an API client.

crisptrutski: Perhaps using empty maps are the least of all evils - although they don't give you anything to search for when debugging.

qnkhuat:
did you change tho quote in FE intentionally? if so you should check with them to see if it fits their convention.

crisptrutski:
> did you change tho quote in FE intentionally? if so you should check with them to see if it fits their convention.

Do you mind pointing me at the change you're talking about?

qnkhuat:
oh it's gone. there was some changes from " -> ` on lots of FE files before.

test/metabase/api/database_test.clj
```diff
@@ -564,6 +565,8 @@
                   {:engine        "h2"
                    :name          "test-data"
                    :features      (map u/qualified-name (driver.u/features :h2 (mt/db)))
+                   :details       database/protected-db-details
+                   :settings      nil
```
calherries, line 569: Why is settings now `nil` here?

test/metabase/api/database_test.clj
```diff
@@ -134,9 +134,10 @@
     (testing "DB details visibility"
       (testing "Regular users should not see DB details"
         (is (= (-> (db-details)
-                   (dissoc :details :schedules))
-               (-> (mt/user-http-request :rasta :get 200 (format "database/%d" (mt/id)))
-                   (dissoc :schedules :can_upload)))))
+                  (dissoc :schedules)
+                  (assoc :details database/protected-db-details))
+              (-> (mt/user-http-request :rasta :get 200 (format "database/%d" (mt/id)))
+                  (dissoc :schedules :can_upload)))))
```
calherries, line 140:
```suggestion
                   (dissoc :schedules)
                   (assoc :details database/protected-db-details))
               (-> (mt/user-http-request :rasta :get 200 (format "database/%d" (mt/id)))
                   (dissoc :schedules :can_upload)))))
```

test/metabase/api/database_test.clj
```diff
@@ -778,8 +781,7 @@
     (testing "Test that we can get all the DBs (ordered by name, then driver)"
       (testing "Database details/settings *should not* come back for Rasta since she's not a superuser"
```
calherries, line 782: This test description is incorrect now. We can just remove this line I think.

calherries (approved):
Looks good to me, you have a failing test though.

It's fair to be nervous about making this breaking change, but I'm not too worried about it. After a brief search through the FE codebase I couldn't find any code that depended the properties of `details` or `settings` you're breaking here.

You should ideally update the types on the FE though, or get a FEer to help you out. This can be another PR. At present they're not very tight at all, and a FE engineer could mistakenly write some code that depends on the keys being present in `database.settings` or `database.details`.
https://github.com/metabase/metabase/blob/5c5651157728eeae104cb02bfcdbceef681958eb/frontend/src/metabase-types/api/database.ts#L10-L13

src/metabase/models/database.clj
```diff
@@ -314,6 +322,33 @@
...
+  (if (not (mi/can-write? db))
+    (u/assoc-existing db :details protected-db-details)
+    (m/update-existing db :details (fn [details]
+                                     (reduce
+                                      #(m/update-existing %1 %2 (constantly protected-password))
+                                      details
+                                      (sensitive-fields-for-db db))))))
+
+(defn- redact-settings [settings]
+  (if-not (or (nil? settings) (map? settings))
```
qnkhuat, line 335: when `settings` is not a map?

src/metabase/models/database.clj
```diff
@@ -314,6 +322,33 @@
             driver.u/default-sensitive-fields))
       driver.u/default-sensitive-fields))
 
+(defn- redact-db-details [db]
+  (if (not (mi/can-write? db))
```
qnkhuat, line 326:
```suggestion
  (if-not (mi/can-write? db)
```

src/metabase/models/database.clj
crisptrutski: This was just preserving existing code. I suspect if it shouldn't be, and we should probably fail this when querying the data.

crisptrutski:
Closing this for now, as I'm disinclined to pursue it. I've moved the conversation back to the parent issue for now, and we can come back and reopen this if we so choose after a high level reconsideration.

PR #39121, opened 2024-02-23
dispatchAndGoNextStep

npretto:
This uses the `goToNextStep` from the other solution but tries to hide it in the `dispatchAndGoNextStep` function.
I liked the idea of leaving the thunks as they were before.

Unfortunately all the tests started failing because of the await on the thunk... after a couple of hours I just gave up... (should have stopped earlier...)


npretto:
Closing this as it's not worth the broken tests 🤷

PR #39132, opened 2024-02-23
Add env var to hide built-in geojson

### LINKED ISSUE #39079: Hide the default maps
We should have a way to hide the default maps. The reason is that if you want a map of the world that is different from the default one, you'll end up with 2 versions of the world map, and you won't be able to hide the default one and leave the one you created.

Hide the default maps. This should be done by an admin

None, unless you do a custom build

A customer that has many customers around the world needs to hide default maps

**Additional context**
Related to https://github.com/metabase/metabase/issues/38919

noahmoss:
Adds an environment variable to hide the built-in geojson maps, per [product request](https://metaboat.slack.com/archives/C05MPF0TM3L/p1708636681879369). Closes https://github.com/metabase/metabase/issues/39079

### DIFF
--- a/src/metabase/api/geojson.clj
+++ b/src/metabase/api/geojson.clj
@@ -37,17 +37,29 @@
 
 (def ^:private CustomGeoJSONValidator (mc/validator CustomGeoJSON))
 
-(def ^:private builtin-geojson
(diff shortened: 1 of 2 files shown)


escherize (approved):
LGTM

PR #39145, opened 2024-02-26
Add unit tests for `useModelIndexesListQuery` hook

nemanjaglumac:
This PR adds missing unit tests for `useModelIndexesListQuery` hook that was introduced in #38664.

### DIFF
--- a/frontend/src/metabase/common/hooks/use-model-indexes-list-query/use-model-indexes-list-query.unit.spec.tsx
+++ b/frontend/src/metabase/common/hooks/use-model-indexes-list-query/use-model-indexes-list-query.unit.spec.tsx
@@ -10,7 +10,8 @@ import { createMockModelIndex } from "metabase-types/api/mocks";
 
 import { useModelIndexesListQuery } from "./use-model-indexes-list-query";
 
-const TEST_ITEM = createMockModelIndex();
+const model_id = 1;
+const TEST_ITEM = createMockModelIndex({ model_id });
 
 const TestComponent = () => {
   const {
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/common/hooks/use-model-indexes-list-query/use-model-indexes-list-query.unit.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    setup();
+    expect(screen.getByTestId("loading-spinner")).toBeInTheDocument();
+  });
+
+  it("should show data from the response", async () => {
+    setup();
+    await waitForLoaderToBeRemoved();
+    expect(screen.getByText(TEST_ITEM.state)).toBeInTheDocument();
+  });
+  it("should not have any metadata in the response", async () => {
```
kamilmielnik, line 56: nit:

```suggestion
  });

  it("should not have any metadata in the response", async () => {
```

frontend/src/metabase/common/hooks/use-model-indexes-list-query/use-model-indexes-list-query.unit.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const TEST_ITEM = createMockModelIndex();
+
+const TestComponent = () => {
+  const {
+    data = [],
+    metadata,
+    isLoading,
+    error,
+  } = useModelIndexesListQuery({ query: { model_id: 1 } });
```
kamilmielnik, line 21: `1` could be a constant and could be reused in `setup`.

kamilmielnik (approved):

frontend/src/metabase/common/hooks/use-model-indexes-list-query/use-model-indexes-list-query.unit.spec.tsx
nemanjaglumac: Good call!

PR #39154, opened 2024-02-26
[FIX: #39152] Public links refer to `null` when question is new

### LINKED ISSUE #39152: Public Links are null for new questions
### Describe the bug

When trying to use the embed button + public link with new questions, the public link popover presents the user with a blank input box. When the user copies the link, the link comes out to 
```
<INSTANCE URL>/api/card/null/public_link
```

We shouldn't allow users to share a public link until the question has been saved.

### To Reproduce

1. Create a question (`New` → `Question`)
2. Click on the `Share` button (on the bottom right of the question view)
3. Click `Create a public link`
4. See that the link box is empty and copying the link gives a `null` URL

### Expected behavior

We shouldn't allow users to share a public link until the question has been saved.

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "21.0.2",
    "java.vendor": "Homebrew",
    "java.version": "21.0.2",
    "java.vm.version": "21.0.2",
    "os.name": "Mac OS X",
    "os.version": "14.1.1",
    "user.language": "en",
    "user.timezone": "Europe/Rome"
  },
  "metabase-info": {
    "databases": [
      "postgres",
      "h2"
    ],
    "hosting-env": "unknown",
      "database": {
        "name": "PostgreSQL",
        "version": "13.11"
      },
      "jdbc-driver": {
        "version": "42.7.1"
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2024-01-29",
      "tag": "v1.48.1-SNAPSHOT",
      "hash": "1bb09a7"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

P2

### Additional context

_No response_

oisincoveney:
Closes #39152

### Description

When questions are new, they don't have an ID and aren't able to possess a public_uuid since they aren't saved. To solve this, we need to make sure that the user has saved the question before they try to create a public link for it.

### How to verify

1. Create a new question
2. See that, when trying to click the "Share" button in the bottom right corner, that a `First, save your question` modal pops up first.

### Demo

### Checklist


frontend/src/metabase/dashboard/components/EmbedMenu/AdminEmbedMenu.tsx
```diff
@@ -53,8 +56,22 @@ export const AdminEmbedMenu = ({
     );
   }
 
+  const onMenuToggle = (t: boolean) => {
+    if (t && resourceType === "question" && !resource.isSaved()) {
+      onModalOpen(MODAL_TYPES.SAVE_QUESTION_BEFORE_EMBED);
```
WiNloSt, line 61: I don't like this solution. It makes this component tightly coupled with how it should be used, ultimately, in `ViewFooter` or `DashboardEmbedAction` and this makes the `onModalOpen` API inconsistent which creates confusion.

If you look in `ViewFooter` (query builder)
https://github.com/metabase/metabase/blob/ace3337474ef1860e78d25b95bebd7622b3a22fd/frontend/src/metabase/query_builder/components/view/ViewFooter.jsx#L151

`onModalOpen` = (modalMode) => void

but in `DashboardEmbedAction`
https://github.com/metabase/metabase/blob/ace3337474ef1860e78d25b95bebd7622b3a22fd/frontend/src/metabase/dashboard/components/DashboardEmbedAction/DashboardEmbedAction.tsx#L22
`onModalOpen` = (void) => void

Also, `onMenuToggle` here _conditonally_ calls `onModalOpen` which makes it hard to understand when my `onModalOpen` would be called or not.

Could you simplify this part more? I'd expect the `onModalOpen` API to be more consistent, and don't have this kind of condition inside the button. The components that use `EmbedMenu` should be the one that tells which modal to open instead of doing that in this component

oisincoveney: Yep, I felt uncomfortable writing this solution as well. You make a good point about the consistency of the `onModalOpen` call, I'll see if I can make it consistent across the different components

frontend/src/metabase/query_builder/components/view/ViewFooter.jsx
```diff
@@ -142,19 +144,23 @@ const ViewFooter = ({
...
-              resourceType="question"
-              hasPublicLink={!!question.publicUUID()}
-              onModalOpen={() =>
-                question.isSaved()
-                  ? onOpenModal("embed")
-                  : onOpenModal("save-question-before-embed")
-              }
-            />
-          ),
+          type === "question" &&
```
oisincoveney, line 147: I moved the `question.isSaved()` logic up here. It almost mimics `DashboardEmbedAction` where it gatekeeps the actual menu until the question is saved.

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
```diff
@@ -817,6 +840,27 @@ function enableEmbeddingForResource({ resource, id }) {
   });
 }
 
+function assertValidPublicLink({ resource, shouldHaveRemoveLink = true }) {
```
oisincoveney, line 843: Let me know if there's an easier way to do this. I've added the regex because it will help us make sure that we're not getting any `null` values in tests

frontend/src/metabase/query_builder/components/view/ViewFooter.jsx
WiNloSt: I like this. It's just that now we're using `DashboardEmbedHeaderButton` inside `query_builder/` directory which is for questions.

If this isn't specifically for dashboard anymore we should rename it accordingly. Like to just `ResourceEmbedHeaderButton` or something.

oisincoveney: Will change right now. Good call

oisincoveney: @WiNloSt let me know what you think. I renamed `DashboardHeaderEmbedButton` to `ResourceEmbedButton` (in dashboards, the button is in the header, but in questions it's in the footer so "Header" is not the greatest name).

Also I moved the component into the `public/components` folder, together with the `EmbedModal` and `EmbedFrame` components which the button leads to in the user flow.

e2e/test/scenarios/sharing/public-sharing-embed-button-behavior.cy.spec.js
WiNloSt: Fair enough. Even though this looks a bit ugly, it gives me confident that the public link gotta be in a certain format, which is better than just asserting it's not null or empty.

WiNloSt: I'd suggest making `shouldHaveRemoveLink` required, this will force you to pass `shouldHaveRemoveLink` in every assertion and make the assertions clearer.

It's much more clear because you don't need to assume there would be default assertions somewhere else.

e.g. between
```ts
assertValidPublicLink({ resource: "question" });
```

VS
```ts
assertValidPublicLink({
  resource,
  shouldHaveRemoveLink: true,
});
```

It isn't that apparent here, but when you have a bunch of optional flags, when reading the test you gotta always think about how do I interpret this tests, what optional flags do I need to take into account, which could make reading the tests a lot harder.

WiNloSt (approved):

PR #39156, opened 2024-02-26
Fix applying default values for a pinned sql question

### LINKED ISSUE #37831: Default values are not respected for pinned questions in collections in v48

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/37831

### Description

For an SQL question, which is pinned to the collection, we need to run query applying the default parameters values.
To achieve this we pass an empty array as a value for `parameters` in `/api/card/:id/query` POST request

### How to verify
follow steps from e2e test or

- create a new sql question with parameter e.g. `select count(*) from people where {{date}}`
- add a filter mapping (field filter type) to the people -> birth date
- select a default value (e.g. `1999-02-26~2024-02-26`)
- save a question and put it to the root collection
- pin this question (it should appear on the top of collection) - make sure the filter is applied

### Demo

https://github.com/metabase/metabase/assets/125459446/1238b880-7571-4e0e-83f9-a7cf50bd4c39

### Checklist

### DIFF
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -34,6 +34,9 @@ export function getValuePopulatedParameters({
   defaultRequired = false,
   collectionPreview = false,
 }) {
+  // pinned native question can have default values on parameters, usually we
+  // get them from URL, which is not the case for collection preview. to force
+  // BE to apply default values to those filters, empty array is provided
   if (collectionPreview) {
     return [];
   }
--- a/frontend/src/metabase-lib/parameters/utils/parameter-values.js
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-values.js
@@ -32,7 +32,7 @@ export function getValuePopulatedParameters({
   parameters,
   values = {},
   defaultRequired = false,
-  collectionPreview,
+  collectionPreview = false,
 }) {
   if (collectionPreview) {
     return [];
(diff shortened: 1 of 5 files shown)


frontend/src/metabase-lib/parameters/utils/parameter-values.js
```diff
@@ -32,7 +32,12 @@ export function getValuePopulatedParameters({
   parameters,
   values = {},
   defaultRequired = false,
+  collectionPreview = false,
 }) {
+  if (collectionPreview) {
+    return [];
+  }
```
oleggromov, line 39: It makes me suspicious that we exclude all parameters, but as I am reasoning through:

1. no filters are set — no parameters have values
2. default values should be populated on the backend
3. required parameters/variables can't exist without a default so 2 applies here

In case the logic above is correct, all e2e tests must pass. Otherwise something _might_ fail.

uladzimirdev: do you think we should have some comment explaining this?

oleggromov: Yeah I guess it won't do any harm.

oleggromov (approved):

PR #39028, opened 2024-02-21
Add analytics events for browse data

iethree:
Part of https://github.com/metabase/metabase/issues/36368

Adds two Snowplow click tracking events to the Browse Data page, for clicks on models and tables, per @cdeweyx's [request](https://metaboat.slack.com/archives/C064EB1UE5P/p1705676820030029?thread_ts=1705674700.235379&cid=C064EB1UE5P)

- [x] add schema file
- [x] add to backend
- [x] add to frontend
- [x] add tests

PR #39159, opened 2024-02-26
Required filters: track when a filter is made required

### LINKED ISSUE #39066: Track required parameter toggling in analytics
https://www.notion.so/metabase/Dashboard-filter-required-9e3702101f3748f38e6ed50a0e4c0e74

oleggromov:
Adds the `dashboard_filter_required` event and updates schemas. Closes #39066 

## What exactly is tracked
Every time the toggle is flipped to required, we will log an event.
[Confirmed by Conor](https://metaboat.slack.com/archives/C0645JP1W81/p1708963461276029?thread_ts=1708961675.462709&cid=C0645JP1W81) on Slack. 

## Testing
Tracking works locally with a setup described [here](https://www.notion.so/metabase/Snowplow-integration-5da1f874beda4153b4fccfa6c1e77caa).

### DIFF
--- a/frontend/src/metabase/dashboard/actions/parameters.ts
+++ b/frontend/src/metabase/dashboard/actions/parameters.ts
@@ -34,7 +34,10 @@ import type {
 } from "metabase-types/api";
 import type { Dispatch, GetState } from "metabase-types/store";
 
-import { trackAutoApplyFiltersDisabled } from "../analytics";
+import {
+  trackAutoApplyFiltersDisabled,
+  trackFilterRequired,
+} from "../analytics";
 import {
   getAutoApplyFiltersToastId,
   getDashboard,
@@ -297,6 +300,13 @@ export const setParameterRequired = createThunkAction(
         ...parameter,
         required,
       }));
+    }
+
+    if (required) {
+      const dashboardId = getDashboardId(getState());
+      if (dashboardId) {
+        trackFilterRequired(dashboardId);
+      }
     }
   },
 );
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/dashboard/actions/parameters.ts
```diff
@@ -298,6 +301,13 @@ export const setParameterRequired = createThunkAction(
         required,
       }));
     }
+
+    if (required) {
```
uladzimirdev, line 305: shouldn't we track only when parameter becomes required from a not-required state?

oleggromov: Do you mean the case when `required` is `true` and `paramater.required` is also true? 
We don't emit this action like that but I could add an additional check if you think it makes sense.

uladzimirdev: I actually mean (if I understand the logic correctly) that we need to track when parameter wasn't required and then became required

uladzimirdev (approved):
I think we need to update some test (probably e2e like it's done [here](https://github.com/metabase/metabase/pull/39028/files#diff-f9fe4fbdf321fea7785d3395db658497cd99411824f5386a467299daec0c7939R31)) to be sure events are logged

frontend/src/metabase/dashboard/actions/parameters.ts
oleggromov: Yeah, I understood. In case we were emitting this action without checks (now because it's a "switch" you can't emit it when a parameter is already required), it would make sense. B broadly speaking, I don't think this is unnecessary. 

This event is used to track feature adoption, which Conor agreed doesn't have to be very precise. That's why we didn't track it at saving for instance and instead track just toggling it on.

PR #39183, opened 2024-02-26
useSetting hook

iethree:
### Description

Credit to @sloansparger  for the idea, we can save a lot of imports and boilerplate if we just create a `useSetting` hook instead of combining `useSelector` with `getSetting`;

**Before**
```ts
const embeddingSameSiteCookieSetting = useSelector(state =>
  getSetting(state, "session-cookie-samesite"),
);
const embeddingAuthorizedOrigins = useSelector(state =>
  getSetting(state, "embedding-app-origin"),
);
```

**After**
```ts
const embeddingSameSiteCookieSetting = useSetting("session-cookie-samesite");
const embeddingAuthorizedOrigins = useSetting("embedding-app-origin");
```

## Todo
- figure out how to programatically replace all these

### Checklist

### DIFF
--- a/frontend/src/metabase/common/hooks/use-setting/use-setting.ts
+++ b/frontend/src/metabase/common/hooks/use-setting/use-setting.ts
@@ -2,6 +2,8 @@ import { useSelector } from "metabase/lib/redux";
 import { getSetting } from "metabase/selectors/settings";
 import type { Settings } from "metabase-types/api";
 
-export const useSetting = (settingName: keyof Settings) => {
+export const useSetting = <SettingName extends keyof Settings>(
+  settingName: SettingName,
+) => {
   return useSelector(state => getSetting(state, settingName));
 };
--- a/frontend/src/metabase/common/hooks/use-setting/use-setting.ts
+++ b/frontend/src/metabase/common/hooks/use-setting/use-setting.ts
@@ -1,0 +1,7 @@
+import { useSelector } from "metabase/lib/redux";
+import { getSetting } from "metabase/selectors/settings";
+import type { Settings } from "metabase-types/api";
+
+export const useSetting = (settingName: keyof Settings) => {
+  return useSelector(state => getSetting(state, settingName));
+};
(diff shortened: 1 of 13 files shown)


frontend/src/metabase/common/hooks/use-setting/use-setting.ts
```diff
@@ -0,0 +1,7 @@
+import { useSelector } from "metabase/lib/redux";
+import { getSetting } from "metabase/selectors/settings";
+import type { Settings } from "metabase-types/api";
+
+export const useSetting = (settingName: keyof Settings) => {
+  return useSelector(state => getSetting(state, settingName));
+};
```
sloansparger, line 7:
```suggestion
export const useSetting = <SettingName extends keyof Settings>(settingName: SettingName) => {
  return useSelector(state => getSetting(state, settingName));
};
```
make generic

iethree:
> I'm wondering if this should return `[value, updateFn(WrappedInDispatch?)]`. It'd be handy when we need to update the settings, but I think we mostly read them so it may not be worth having to do `const [value] = useSettings` when we don't need to update them

@sloansparger just asked me the same question. I don't have a strong preference either way, but we rarely update settings in application code, it's mostly in settings, which isn't largely hook-based yet.

sloansparger:
@iethree I think it's fair to leave it as is for now. I do think @naatz-co was eventually wanting to have settings success/failure messages appear next to our save on blur settings inputs. The [val,set] api might be a more desirable change if/when we decide to that.

sloansparger (approved):

iethree:
for posterity: here's the codemod I used to update things

```js
const fs = require('fs');

const path = require('path');

function replacePatternInFile(filePath, oldPattern, newPattern) {
    const content = fs.readFileSync(filePath, 'utf8');
    const updatedContent = content.replace(oldPattern, newPattern);
    if (content !== updatedContent) {
      fs.writeFileSync(filePath, updatedContent);
      console.log(`✅ Updated file: ${filePath}`);
    }
}

function searchReplaceInDirectory(directory, oldPattern, newPattern) {
    fs.readdirSync(directory).forEach(file => {
        const filePath = path.join(directory, file);
        if (fs.lstatSync(filePath).isDirectory()) {
          searchReplaceInDirectory(filePath, oldPattern, newPattern);
        } else if (file.endsWith('.jsx') | file.endsWith('.tsx')) {
          replacePatternInFile(filePath, oldPattern, newPattern);
        }
    });
}

// Example usage:
const oldPattern = /useSelector\(state\s=>\sgetSetting\(\s*state\s*,*\s*("[\w\d_-]+"),*\s*\)\);/ig;
const newPattern = 'useSetting($1);';

searchReplaceInDirectory('./frontend/src/metabase/', oldPattern, newPattern);
```

Issue #37009, opened 2023-12-20
Updated native models may lose their result set metadata


zbodi74:
### Describe the bug

Upon updating a native model which is based on another model, the result set metadata may be lost.
When this occurs, the query editor won't show available columns for questions that are based on this model.

The workaround is to update the model, and making sure to preview the results before saving it.

### To Reproduce

1. Create native model `model1`:
```
      select 1 as "client_id", 1 as "value"
UNION select 1 as "client_id", 2 as "value"
UNION select 2 as "client_id", 3 as "value"
UNION select 3 as "client_id", 4 as "value"
```
2. Create and save a native model `model2`, which uses `model1` as its source:
```
select * from {{ #nnn-model1 }}
```

3. Update the query definition of model2, and make a minor, non-impactful change, for instance:
```
select cte.* from {{ #nnn-model1 }} as cte
```
, and save it, without previewing the results in the editor.

4. Now start a GUI question based on `model2`. 
5. (!)The columns of the model won't be available in the editor.
<img width="859" alt="image" src="https://github.com/metabase/metabase/assets/11080638/94e6a189-b349-487a-925f-2af09a0bcdef">

### Expected behavior

It should wirk

### Logs

Logs:
```
2023-12-20 18:47:06,445 INFO api.dataset :: Source query for this query is Card 77
2023-12-20 18:47:06,455 WARN middleware.add-implicit-clauses :: Warning: cannot determine fields for an explicit `source-query` unless you also include `source-metadata`.
Query: {:params [],
 :native
 "select cte.* from (select 1 as \"client_id\", 1 as \"value\"\nUNION select 1 as \"client_id\", 2 as \"value\"\nUNION select 2 as \"client_id\", 3 as \"value\"\nUNION select 3 as \"client_id\", 4 as \"value\") as cte"}
```

/api/card response for `model2`:
```
{
  "description": null,
  "archived": false,
  "collection_position": 1,
  "table_id": null,
  "result_metadata": null,
  "creator": {
[…]
  },
  "can_write": true,
  "database_id": 1,
  "enable_embedding": false,
  "collection_id": null,
  "query_type": "native",
  "name": "model2",
  "last_query_start": "2023-12-20T18:49:08.371778Z",
  "dashboard_count": 0,
  "persisted": false,
  "average_query_time": 21.0,
  "creator_id": 1,
  "moderation_reviews": [],
  "updated_at": "2023-12-20T18:47:06.210264Z",
  "made_public_by_id": null,
[...]

PR #39184, opened 2024-02-26
Keep result-metadata of native queries if possible

metamben:
Part of #37009.

Related FE PR: #39201

The problem this PR is fixing is that for native queries the before update hook of the card doesn't populate `result_metadata`, but sets it to `nil` instead (to make sure we have no stale metadata, which we consider worse than no metadata at all). When a native question is updated and the result metadata doesn't change, the `t2/changes` function omits `result_metadata` and the above behavior is triggered.

To solve the problem, when a question is updated via the API and we are able to determine the result metadata, we set a flag that forces the use of that metadata (which might mean that the field is left untouched, if there's no change).

This PR doesn't fully solve the problem, because we only set the result metadata if the query finishes in 1.5 s. In other cases, the behavior is unchanged, and the FE will get `nil` as metadata.

### DIFF
--- a/src/metabase/models/card.clj
+++ b/src/metabase/models/card.clj
@@ -547,16 +547,22 @@
     (parameter-card/upsert-or-delete-from-parameters! "card" (:id card) (:parameters card))))
 
 (t2/define-before-update :model/Card
-  [card]
+  [{:keys [verified-result-metadata?] :as card}]
   ;; remove all the unchanged keys from the map, except for `:id`, so the functions below can do the right thing since
   ;; they were written pre-Toucan 2 and don't know about [[t2/changes]]...
   ;;
   ;; We have to convert this to a plain map rather than a Toucan 2 instance at this point to work around upstream bug
   ;; https://github.com/camsaul/toucan2/issues/145 .
-  (-> (into {:id (:id card)} (t2/changes card))
+  (-> (into {:id (:id card)} (t2/changes (dissoc card :verified-result-metadata?)))
       ensure-type-and-dataset-are-consistent
       maybe-normalize-query
-      populate-result-metadata
+      ;; If we have fresh result_metadata, we don't have to populate it anew. When result_metadata doesn't
+      ;; change for a native query, populate-result-metadata removes it (set to nil) unless prevented by the
+      ;; verified-result-metadata? flag (see #37009).
+      (cond-> #_changes
+        (or (empty? (:result_metadata card))
+            (not verified-result-metadata?))
+        populate-result-metadata)
       pre-update
       populate-query-fields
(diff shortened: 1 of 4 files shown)


src/metabase/models/card.clj
```diff
@@ -857,12 +863,14 @@ saved later when it is ready."
...
+                  ;; Other values should only be modified if they're passed in as non-nil
                   (u/select-keys-when card-updates
-                                      :present #{:collection_id :collection_position :description :cache_ttl :dataset :type}
-                                      :non-nil #{:dataset_query :display :name :visualization_settings :archived :enable_embedding
-                                                 :parameters :parameter_mappings :embedding_params :result_metadata :collection_preview}))))
+                                      :present #{:collection_id :collection_position :description :cache_ttl :dataset
+                                                 :type}
+                                      :non-nil #{:dataset_query :display :name :visualization_settings :archived
+                                                 :enable_embedding :parameters :parameter_mappings :embedding_params
+                                                 :result_metadata :collection_preview :verified-result-metadata?}))))
```
metamben, line 873: Only this line contains interesting an change.

src/metabase/models/card.clj
```diff
@@ -547,16 +547,22 @@
...
 (t2/define-before-update :model/Card
-  [card]
+  [{:keys [verified-result-metadata?] :as card}]
   ;; remove all the unchanged keys from the map, except for `:id`, so the functions below can do the right thing since
   ;; they were written pre-Toucan 2 and don't know about [[t2/changes]]...
   ;;
   ;; We have to convert this to a plain map rather than a Toucan 2 instance at this point to work around upstream bug
   ;; https://github.com/camsaul/toucan2/issues/145 .
-  (-> (into {:id (:id card)} (t2/changes card))
+  (-> (into {:id (:id card)} (t2/changes (dissoc card :verified-result-metadata?)))
```
bshepherdson, line 556: Having to add and then remove this data seems a bit noisy. How about `update-meta` on the card instead of adding a new key to the card itself?

bshepherdson (approved):

kamilmielnik (commented):
Looks good from FE perspective :+1: 

I have a PR based on this one coming: #39201

src/metabase/models/card.clj
metamben: I'm a bit vary of relying on `meta` being preserved. Just naively putting metadata at the point where I now `assoc` `:verified-result-metadata?` is not enough, that metadata doesn't reach the before update hook (i.e., `card` doesn't have it).

metamben:
Since this is for an escalation, I'm merging this as is, we can revisit using meta to pass in the flag.

PR #39196, opened 2024-02-27
Optimize the CI milestone reminder

nemanjaglumac:
This PR narrows down the matching pattern for the milestone reminder to only the descriptions that explicitly contain one of the closing keywords followed by the issue (number).

To verify that the proposed solution works, please check:
- https://regexr.com/7sk94

Also see the list of the closing keywords as defined by GitHub. https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

### DIFF
--- a/.github/workflows/check-milestone.yml
+++ b/.github/workflows/check-milestone.yml
@@ -26,8 +26,8 @@ jobs:
 
             // the github API doesn't expose linked issues, so we have to parse the body ourselves
             function getLinkedIssueId(pull) {
-              // https://regexr.com/7mjnp
-              const match = pull.data.body.match(/(\/issues\/|#)(\d+)/);
+              // https://regexr.com/7sk94
+              const match = pull.data.body.match(/(close(s|d)?|fixe?(s|d)?|resolve(s|d)?) (#|https?:\/\/github.com\/.+\/issues\/)(\d+)/);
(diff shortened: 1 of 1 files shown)


.github/workflows/check-milestone.yml
```diff
@@ -26,8 +26,8 @@ jobs:
 
             // the github API doesn't expose linked issues, so we have to parse the body ourselves
             function getLinkedIssueId(pull) {
-              // https://regexr.com/7mjnp
-              const match = pull.data.body.match(/(\/issues\/|#)(\d+)/);
+              // https://regexr.com/7sk94
+              const match = pull.data.body.match(/(close(s|d)?|fixe?(s|d)?|resolve(s|d)?) (#|https?:\/\/github.com\/.+\/issues\/)(\d+)/);
```
npretto, line 30: I believe this would be case-sensitive if we don't add
```suggestion
              const match = pull.data.body.match(/(close(s|d)?|fixe?(s|d)?|resolve(s|d)?) (#|https?:\/\/github.com\/.+\/issues\/)(\d+)/i);
```

I often see it capitalized as `Closes #xxxx` when at the start of issues

iethree (approved):

PR #39200, opened 2024-02-27
Migrate from legacy popover in notebook components

ranquild:
Epic https://github.com/metabase/metabase/issues/39210
https://github.com/metabase/metabase/actions/runs/8068495149/job/22041427386?pr=39200

In order to make it easier to implement a new feature, let's remove the old popover from notebook components.

The change triggered a lot changes in tests because the new popover is async. Also in terms of a11y it's somewhat different, so E2E tests are also affected.


uladzimirdev (approved):
🤞

Issue #38603, opened 2024-02-09
Required parameters: follow-ups 


oleggromov:
Once the [[Epic] Let creators make dashboard and native query filters required](https://github.com/metabase/metabase/issues/36524) lands, we need to do a bunch of follow-ups.

## Milestone 3
- [x]  https://github.com/metabase/metabase/pull/39159
## Core updates
Please leave these to @oleggromov 

- [x] https://github.com/metabase/metabase/pull/38906
- [x] https://github.com/metabase/metabase/pull/38997
- [x] https://github.com/metabase/metabase/pull/39128

## UI/UX updates
These mostly depend on the core updates above, because it's impractical to update the currently convoluted components (in particular, `ParameterValueWidget`) _before_ we change them to use Mantine.

- [x] https://github.com/metabase/metabase/pull/38468
- [x] https://github.com/metabase/metabase/pull/39068
- [x] increase gap between the toggle and input by 4px - addressed in https://github.com/metabase/metabase/pull/38906
- [x] https://github.com/metabase/metabase/pull/39206

oleggromov:
This round of follow-ups is merged.
The next is https://github.com/metabase/metabase/issues/39211

PR #39206, opened 2024-02-27
Disable done button in Dashboard parameter editing for required param

oleggromov:
The last from the [follow-ups list](https://github.com/metabase/metabase/issues/38603), this one disables Done button in the dashboard parameter sidebar when a parameter is required but doesn't have a default value.

https://github.com/metabase/metabase/assets/2196347/53c2ef28-ab36-4be1-b9e3-2bb561b244a4

### DIFF
--- a/e2e/test/scenarios/dashboard-filters/dashboard-filters-date.cy.spec.js
+++ b/e2e/test/scenarios/dashboard-filters/dashboard-filters-date.cy.spec.js
@@ -17,6 +17,7 @@ import {
   ensureDashboardCardHasText,
   resetFilterWidgetToDefault,
   sidebar,
+  dashboardParametersDoneButton,
 } from "e2e/support/helpers";
 
 import * as DateFilter from "../native-filters/helpers/e2e-date-filter-helpers";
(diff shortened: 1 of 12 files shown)


e2e/test/scenarios/dashboard-filters/dashboard-filters-date.cy.spec.js
```diff
@@ -106,12 +107,19 @@ describe("scenarios > dashboard > filters > date", () => {
...
     dashboardSaveButton().realHover();
-
     cy.findByRole("tooltip").should(
       "contain.text",
       'The "Month and Year" parameter requires a default value but none was provided.',
     );
 
+    // Can't close sidebar without a default value
+    dashboardParametersDoneButton().should("be.disabled");
+    dashboardParametersDoneButton().realHover();
```
nemanjaglumac, line 117: This is the nittiest nit ever, but just FYI this also works:
```js
dashboardSaveButton().should("be.disabled").realHover();
```

frontend/src/metabase/dashboard/components/Sidebar.tsx
```diff
@@ -46,20 +47,25 @@ export function Sidebar({
...
           {onClose && (
-            <ButtonDeprecated
-              primary
-              small
-              className="ml-auto"
-              onClick={onClose}
-              disabled={closeIsDisabled}
-            >{t`Done`}</ButtonDeprecated>
+            <Tooltip label={closeTooltip} hidden={!closeTooltip}>
+              {/* without a div we will need hacks to make tooltip work */}
```
nemanjaglumac, line 59: Just out of curiosity, what kind of hacks? Does React Fragment not work here?

nemanjaglumac (approved):

e2e/test/scenarios/dashboard-filters/dashboard-filters-date.cy.spec.js
oleggromov: Thanks, I'll change that in one of the next commits/PRs so that I don't have to wait for the CI again.

frontend/src/metabase/dashboard/components/Sidebar.tsx
oleggromov: Mantine turns off pointer events on a disabled button so that the tooltip can't receive a mouseenter or something like that. Therefore we need a real DOM element that is not messing with the mouse events.

PR #39212, opened 2024-02-27
Trend Charts: Improve Error Messages

### LINKED ISSUE #39129: Throw better error messages in trend charts
Discussion: https://metaboat.slack.com/archives/C064QMXEV9N/p1708360957209439
PR: https://github.com/metabase/metabase/pull/39212

JesseSDevaney:
As part of the response to a customer's bug report (https://discourse.metabase.com/t/an-error-occurred-while-displaying-this-card-v0-49-rc1/78633/1), we are improving error messages, so that we can better establish what is happening if a trend chart is not working as expected.

Closes:
- https://github.com/metabase/metabase/issues/39129


frontend/src/metabase/visualizations/visualizations/SmartScalar/compute.js
kulyk, line 1: I wonder if these messages should to be translated? 🤔

kulyk (commented):
LGTM! Please double-check if we should translate these

kulyk (approved):

PR #39220, opened 2024-02-27
Remove unused global css

deniskaber:
Relates to https://github.com/metabase/metabase/issues/38811 , Milestone 1

### Description

This removes global CSS rules that are not used in the code.

### How to verify
* Mostly visual testing
* Also we should double-check that removed rules are not used in the codebase


frontend/src/metabase/css/dashboard.module.css
```diff
@@ -104,14 +70,6 @@
   transition: border 0.3s, background-color 0.3s;
 }
 
-:global(.Dash--editing .Card-title:first-of-type) {
```
WiNloSt, line 107: I found both `.Dash--editing` and `.Card-title` are still being used before this PR.

https://github.com/metabase/metabase/blob/afc28b95811276d042d5dc1a1eff05faafd9393e/frontend/src/metabase/dashboard/components/DashboardGrid.tsx#L589

https://github.com/metabase/metabase/blob/afc28b95811276d042d5dc1a1eff05faafd9393e/frontend/src/metabase/visualizations/components/LegendHeader.jsx#L61

https://github.com/metabase/metabase/blob/afc28b95811276d042d5dc1a1eff05faafd9393e/frontend/src/metabase/visualizations/visualizations/Scalar/Scalar.jsx#L221

Can you make sure the `margin-top: 0.5rem` is still applied? I haven't tried to test if this rule is even possible, but it would be nice to double check.

frontend/src/metabase/css/dashboard.module.css
```diff
@@ -205,10 +163,6 @@
   transition: all 0.15s linear;
 }
 
-:global(.Dash--editing .Card-title) {
```
WiNloSt, line 208: The previous comment applies. Please double check if this rule is still used or not.

frontend/src/metabase/dashboard/components/DashCard/DashCard.tsx
```diff
@@ -257,7 +257,7 @@ function DashCardInner({
     <ErrorBoundary>
       <DashCardRoot
         data-testid="dashcard"
-        className="Card rounded flex flex-column hover-parent hover--visibility"
+        className="Card relative rounded flex flex-column hover-parent hover--visibility"
```
WiNloSt, line 260: Why could we remove `relative` here? Is it because the component already have `position: relative` applied?

WiNloSt (approved):
Overall this looks great, please take a look at my comments and see if there is any fix needed.

PR #39223, opened 2024-02-27
Document `null` vs `undefined` setting type

nemanjaglumac:
Documents important distinction between `null` and `undefined` in the context of the application settings.
The comment in the diff should be self-explanatory.

### DIFF
--- a/frontend/src/metabase-types/api/settings.ts
+++ b/frontend/src/metabase-types/api/settings.ts
@@ -359,6 +359,35 @@ export type UserSettings = {
   "show-updated-permission-banner": boolean;
 };
 
+/**
+ * Important distinction between `null` and `undefined` settings values.
+ *  - `null` means that the setting actually has a value of `null`.
+ *  - `undefined` means that the setting is not available in a certain context.
+ *
+ * Further longer explanation:
+ *
+ * `null` is a backend way of deleting a value, which can backfire on frontend if we are not
+ * aware of this distinction!
+ *
+ * Do not use `undefined` when checking for a setting value! Use `null` instead.
+ * Use `undefined` only when checking does the setting exist in a certain context.
+ *
+ * Contexts / Scopes:
+ * Settings types are divided into contexts to make this more explicit:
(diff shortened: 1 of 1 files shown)


frontend/src/metabase-types/api/settings.ts
```diff
@@ -359,6 +359,35 @@ export type UserSettings = {
...
 };
 
+/**
+ * Important distinction between `null` and `undefined` settings values.
+ *  - `null` means that the setting actually has a value of `null`.
+ *  - `undefined` means that the setting is not available in a certain context.
+ *
+ * Further longer explanation:
+ *
+ * `null` is a backend way of deleting a value, which can backfire on frontend if we are not
```
npretto, line 369: can you clarify "deleting" in this context?
Deleting the setting (as in we don't need it anymore so we remove the defsetting) or "un-setting" it (as in the user removed clicked on "reset to default" or similar)?

npretto: I am confused by "deleting a value", in js deleting means removing it from the object, which means it won't show up in the object (and technically the value is now undefined):
image

Do you have an example of when the BE "deletes" a value and in which way it can backfire?

nemanjaglumac: I'll clarify this in the comment.

It helps if you look at types.
`"last-used-native-database-id"?: number | null;`

`?` - this is user level setting, it might not be available for unauthenticated users
`null` - default value, or "no value" (this is where the confusion might kick in because on FE we're used to think of this as `undefined` more often than not
`number` - the value has been set

If you want to delete/reset it, you'd change it to `nulll`.

Makes more sense now?

nemanjaglumac: Does this make it less confusing?
https://github.com/metabase/metabase/pull/39223/commits/cadf6e3098372073dbe48eee41e68d7ff58022b0

ranquild (approved):

Issue #39163, opened 2024-02-26
Create new UI for performance admin, with three possible cache invalidation strategies for databases


rafpaf:
```[tasklist]
### Milestone 1
- [x] Add Performance tab (https://github.com/metabase/metabase/pull/39234)
- [x] Add data caching settings tab (https://github.com/metabase/metabase/pull/39234)
- [x] Add three-panel configuration editor (https://github.com/metabase/metabase/pull/39234)
- [x] Add don't cache (with no connection to backend (https://github.com/metabase/metabase/pull/39234)
- [x] Add TTL (with no connection to backend) (https://github.com/metabase/metabase/pull/39234)
- [x] Add don't cache (with real connection to backend) https://github.com/metabase/metabase/pull/39234
- [x] Add TTL (with real connection to backend) https://github.com/metabase/metabase/pull/39234
- [x] Split out enterprise-specific code #39234
- [x] Add configuration option for cache strategy: duration #39234
- [x] Add error messages to form (#39234)
- [x] Add confirmation modal before resetting databases to default  (#39234)
- [x] Add confirmation modal before leaving page  (#39234)
- [x] Remove superseded items from cache settings page  (#39234)
- [x] Add basic unit tests  (#39234)
- [x] Ensure "reset all" button shows different loading states  (#39234)
- [x] Add background hover color to "reset all" button  (#39234)
- [x] Ensure "Save changes" button says "Save changes" after form is edited  (#39234)
- [x] Ensure "Reset all to default" button says "Reset all to default" when form is edited  (#39234)
- [x] Add 'example' link with tooltip  (#39234)
- [x] To say that the root strategy is "nocache, delete it  (#39234)
- [x] Use 'Reset all database caching policies to the default?' in the reset button confirmation modal  (#39234)
- [x] Split StrategyEditorForDatabases into smaller components and hooks  (#39234)
- [x] Split out enterprise code thoroughly
- [x] Tweak: reset button from 'Saved' to 'Save changes' when the user edits the form
```

**Milestone 2 is now in a separate issue: #40978*

```[tasklist]
### Open questions (note: these are not tasks)
- [x] It might be confusing to say 'resetting to duration' if the database strategies are also duration. Perhaps we should say 'resetting to duration (10 hours)' or just 'reset to the default strategy'
```

rafpaf:
Moved this list into the epic issue: https://github.com/metabase/metabase/issues/39153

PR #39234, opened 2024-02-27
Basic cache admin UI

rafpaf:
Adds:
* Performance tab 
* Data caching settings tab
* Two-panel strategy editor
* "Don't cache" strategy
* "TTL" strategy
* "Duration" strategy
* "Use default" option for databases
* OSS page
* Unit tests

[Loom](https://www.loom.com/share/b1b173a52be9401588ffb63f718e052a?sid=bed4bdf5-f967-4d93-af55-9b96b6b52730)


frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+export const StrategyEditorForDatabases = withRouter(
+  ({
+    canOverrideRootStrategy,
+    router,
+    route,
+  }: {
+    canOverrideRootStrategy: boolean;
+    router: InjectedRouter;
+    route?: Route;
+  }) => {
```
sloansparger, line 45: I think I'd like to see this component split up a bit personally. We have 204 lines of code before we get to the actual returned mark up, which is a sign to me that this is probably doing too much.

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+        );
+        if (values.type === "inherit") {
+          await CacheConfigApi.delete(baseConfig, { hasBody: true }).then(
+            () => {
+              setConfigs(otherConfigs);
+            },
+          );
+        } else {
+          const validFields = getFieldsForStrategyType(values.type);
+          const newStrategy = pick(values, validFields) as Strategy;
```
sloansparger, line 202: Why is this needed?

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+    const error = databasesResult.error || configsResult.error;
+    const loading = databasesResult.isLoading || configsResult.loading;
+
+    if (error || loading) {
+      return showLoadingSpinner ? (
+        <LoadingAndErrorWrapper error={error} loading={loading} />
+      ) : (
+        <></>
+      );
+    }
```
sloansparger, line 248: This should be it's own component. Even if you don't have time to make this a generic component for everyone to use, it feels quite unrelated to strategies being edited so moving it out would be nice.

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+        <Stack spacing="xl" lh="1.5rem" maw="32rem" mb="1.5rem">
+          <aside>{PLUGIN_CACHING.explanation}</aside>
+        </Stack>
+        {confirmationModal}
+        <Box
+          style={{
+            display: "grid",
+            gridTemplateColumns: "minmax(5rem, 30rem) minmax(5rem, auto)",
+            overflow: "hidden",
+          }}
```
sloansparger, line 261: I'm seeing a lot of using of the style prop in this PR. I think this is the flaw with mantine's style attribute props... wondering if these should be styled-components instead, but I'm personally unsure of the team's philosophy here.

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+        </Box>
+      </TabWrapper>
+    );
+  },
+);
+
+const getFieldsForStrategyType = (strategyType: StrategyType) => {
+  const strategy = Strategies[strategyType];
+  const validationSchemaDescription =
+    strategy.validateWith.describe() as SchemaObjectDescription;
```
sloansparger, line 324: shouldn't SchemaObjectDescription be the return type for this method?

frontend/src/metabase/admin/performance/components/StrategyForm.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const selectedStrategyType = values.type;
+
+  useEffect(() => {
+    setIsDirty(dirty);
+  }, [dirty, setIsDirty]);
+
+  useEffect(() => {
+    if (selectedStrategyType === "duration") {
+      setFieldValue("unit", "hours");
+    }
```
sloansparger, line 76: enums for all the things! avoid any hard coded strings. future you will appreciate it when you need to refactor this feature and you're trying to search for every instance of the word "unit", "duration", or "hours" in the codebase.

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+    const { show: askConfirmation, modalContent: confirmationModal } =
+      useConfirmation();
+
+    const askBeforeDiscardingChanges = useCallback(
+      (onConfirm: () => void) =>
+        askConfirmation({
+          title: t`Discard your changes?`,
+          message: t`Your changes haven’t been saved, so you’ll lose them if you navigate away.`,
+          confirmButtonText: "Discard",
```
sloansparger, line 117: should this be translated?

sloansparger (changes requested):
only doing a partial review for now, I'll try to get to the rest of it later today

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
rafpaf: Hah, I did already move it out, but I forgot

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
rafpaf: Right now that call to `describe` returns a `SchemaDescription`. Maybe this indicates I misconfigured something?

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
rafpaf: During development it's faster for me to use the style prop (though I wish there was a css-to-js converter for styles). My sense is, if it's only a few style properties, it doesn't need a separate styled component.

frontend/src/metabase/admin/performance/hooks/useDelayedLoadingSpinner.tsx
iethree, line 1: nice!

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,389 @@
...
+  router: InjectedRouter;
+  route?: Route;
+}) => {
+  const [
+    // The targetId is the id of the model that is currently being edited
+    targetId,
+    setTargetId,
+  ] = useState<number | null>(null);
+
+  const databasesResult = useDatabaseListQuery();
```
rafpaf, line 52: Porting this to RTK is a Milestone 2 task

frontend/src/metabase/ui/components/buttons/Button/Button.styled.tsx
```diff
@@ -32,6 +37,7 @@ export const getButtonOverrides = (): MantineThemeOverride["components"] => ({
               marginLeft: 0,
             },
           },
+          "&:active": animate ? "" : { transform: "none" },
```
sloansparger, line 40: it's looking like you're using this to disable animations? does the empty string break existing animations?

frontend/src/metabase/services.js
```diff
@@ -613,3 +613,9 @@ export const ApiKeysApi = {
   edit: PUT("/api/api-key/:id"),
   regenerate: PUT("/api/api-key/:id/regenerate"),
 };
+
+export const CacheConfigApi = {
+  list: GET("/api/cache"),
+  update: PUT("/api/cache"),
+  delete: DELETE("/api/cache"),
+};
```
sloansparger, line 621: Sorry, I think I already called this out but to be sure, can we have a follow up task to move this to RTKQuery?

frontend/src/metabase/plugins/index.ts
```diff
@@ -276,14 +280,16 @@ export const PLUGIN_MODERATION = {
 };
 
 export const PLUGIN_CACHING = {
+  cacheTTLFormField: null as any,
   dashboardCacheTTLFormField: null,
   questionCacheTTLFormField: null,
-  getQuestionsImplicitCacheTTL: (_question?: any) => null,
-  QuestionCacheSection: PluginPlaceholder,
-  DashboardCacheSection: PluginPlaceholder,
-  DatabaseCacheTimeField: PluginPlaceholder,
```
sloansparger, line 284: Did you forget to add a type for your plugin?

frontend/src/metabase/forms/components/FormSubmitButton/FormSubmitButton.tsx
```diff
@@ -8,11 +8,13 @@ import { Button } from "metabase/ui";
...
 import { useFormSubmitButton } from "../../hooks";
 
-export interface FormSubmitButtonProps extends Omit<ButtonProps, "children"> {
-  label?: string;
-  activeLabel?: string;
-  successLabel?: string;
-  failedLabel?: string;
+export interface FormSubmitButtonProps
+  extends Omit<ButtonProps, "children">,
+    Omit<HTMLAttributes<HTMLButtonElement>, "color"> {
```
sloansparger, line 13: I _think_ ButtonProps already extends the HTMLButton type, so you might be able to make this `Omit<ButtonProps, "children" | "color">`

frontend/src/metabase/services.js
rafpaf: Def, I've got a follow-up task in Milestone 2 for this

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,328 @@
...
+        lists.push(CacheConfigApi.list({ model: "database" }));
+      }
+      const [rootConfigsFromAPI, savedConfigsFromAPI] = await Promise.all(
+        lists,
+      );
+      const rootConfig = rootConfigsFromAPI?.data?.[0] ?? {
+        model: "root",
+        model_id: rootId,
+        strategy: { type: "nocache" },
+      };
```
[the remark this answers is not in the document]
rafpaf: Good point. I'm inclined to leave this unremedied but I've asked Sanya on Slack

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
rafpaf: Good question. I've added a comment to explain

frontend/src/metabase/admin/performance/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+  multiplier: positiveInteger.default(10),
+});
+
+export const durationStrategyValidationSchema = Yup.object({
+  type: Yup.string().equals(["duration"]),
+  duration: positiveInteger.default(24),
+  unit: Yup.string().matches(unitOfTimeRegex),
+});
+
+export const strategyValidationSchema = Yup.object().test(
```
sloansparger, line 35: Yup? More like yuk. I can't believe they don't have a `Yup.oneOf` method 🤮

frontend/src/metabase/admin/performance/types.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ModelPersistence = "modelPersistence",
+  CachingStats = "cachingStats",
+}
+export const isValidTabId = (tab: unknown): tab is TabId =>
+  typeof tab === "string" && Object.values(TabId).map(String).includes(tab);
+
+export type ObjectWithType = {
+  type: string;
+  [key: string]: string;
+};
```
sloansparger, line 131: Probably not what you want
Screenshot 2024-04-02 at 9 18 35 AM

frontend/src/metabase/admin/performance/types.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export enum TabId {
+  DataCachingSettings = "dataCachingSettings",
+  DashboardAndQuestionCaching = "dashboardAndQuestionCaching",
+  ModelPersistence = "modelPersistence",
+  CachingStats = "cachingStats",
+}
+export const isValidTabId = (tab: unknown): tab is TabId =>
+  typeof tab === "string" && Object.values(TabId).map(String).includes(tab);
```
sloansparger, line 126:
```suggestion
const validTabIds = new Set(Object.values(TabId).map(String))

export const isValidTabId = (tab: unknown): tab is TabId =>
  typeof tab === "string" && validTabIds.has(tab);
```

frontend/src/metabase/admin/performance/types.tsx
```diff
@@ -0,0 +1,146 @@
...
+    return null;
+  }
+  const type = Strategies[strategy.type];
+  return type.shortLabel ?? type.label;
+};
+
+export const isValidStrategyName = (
+  strategy: string,
+): strategy is StrategyType => {
+  return Object.keys(Strategies).includes(strategy);
```
sloansparger, line 58: prefer Set created outside of this function (avoid allocating a new array on every check) and use the O(1) `.has` method

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,379 @@
...
+
+  const showSpinner = useDelayedLoadingSpinner();
+  const error = databasesResult.error || configsResult.error;
+  const loading = databasesResult.isLoading || configsResult.loading;
+
+  if (error || loading) {
+    return showSpinner ? (
+      <LoadingAndErrorWrapper error={error} loading={loading} />
+    ) : (
+      <></>
```
sloansparger, line 272:
```suggestion
      null
```

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,379 @@
+import type { Location } from "history";
```
sloansparger, line 1: This file is really long. Breaking ~200 LoC is when I personal start thinking about code getting split out to another file, 250 is my limit unless there's an obviously good reason for it.

frontend/src/metabase/forms/components/FormProvider/FormProvider.tsx
```diff
@@ -17,15 +18,17 @@ export function FormProvider<T extends FormikValues, C = unknown>({
...
     initialValues,
     validationSchema,
     validationContext,
   });
+  const setStatus = (status: FormStatus) =>
+    setState(state => ({ ...state, status }));
 
   return (
-    <FormContext.Provider value={state}>
+    <FormContext.Provider value={{ ...state, setStatus }}>
```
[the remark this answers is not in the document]
sloansparger: Nope, value will be a new object on each render because you're spreading state and adding setStatus each time.

frontend/src/metabase/admin/performance/types.tsx
rafpaf: Good catch. This is actually a bit of dead code - oops.

frontend/src/metabase/admin/performance/types.tsx
rafpaf: Good idea, thank you!

frontend/src/metabase/admin/performance/hooks/useHasVerticalScrollbar.tsx
```diff
@@ -0,0 +1,14 @@
...
+  const [hasVerticalScrollbar, setHasVerticalScrollbar] = useState(false);
+
+  const el = ref.current;
+  useEffect(() => {
+    if (el) {
+      setHasVerticalScrollbar(el.scrollHeight > el.clientHeight);
+    }
+  }, [el?.scrollHeight, el?.clientHeight, el, setHasVerticalScrollbar]);
+  return { hasVerticalScrollbar, ref };
+};
```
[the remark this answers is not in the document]
rafpaf: Nice catch, thank you

frontend/src/metabase/admin/performance/components/StrategyForm.tsx
```diff
@@ -0,0 +1,266 @@
...
+  strategyType: StrategyType;
+} & Partial<FormTextInputProps>) => {
+  return (
+    <FormTextInput
+      type="number"
+      name={props.name ?? ""}
+      min={1}
+      styles={{
+        input: {
+          // This is like `text-align: right` but it's RTL-friendly
```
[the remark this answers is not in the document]
rafpaf: This is currently a Notion doc. https://www.notion.so/metabase/Logical-properties-in-CSS-8861ee38904e400ebf2f748e4ab4d325 Perhaps it'll become a markdown doc eventually

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
rafpaf: I've moved a bunch of logic into separate hooks and components, so the file is now down to 223 LOC, and is more readable. Big thanks to you for the suggestion.

frontend/src/metabase/plugins/index.ts
rafpaf: I tried to add types to these plugins in a separate branch, but there were some type inconsistencies in preexisting code that made this hard, so I decided to pull back.

frontend/src/metabase/admin/performance/components/ResetAllToDefaultButton.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const rootConfig = findWhere(configs, { model_id: rootId });
+
+  // TODO:: It might be confusing to say 'resetting to duration' if the
+  // database strategies are also duration. Perhaps we should say 'resetting to
+  // duration (10 hours)' or just 'reset to the default strategy'
+
+  const resetAllToDefault = useCallback(async () => {
+    const originalConfigs = [...configs];
+    setConfigs((configs: Config[]) =>
+      configs.filter(({ model }) => model !== "database"),
```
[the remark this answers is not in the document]
rafpaf: This is a string union type so I'm inclined to leave it - but you're welcome to press me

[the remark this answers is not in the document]
sloansparger: perhaps this is a crusade for another day... move alone 😉

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
rafpaf: Thanks for the nudge - I've now split it into more components and hooks

mazameli (changes requested):
Minor things:
1. I think we should Performance after Permissions and before Tools in the top nav. It doesn't belong quite so "soon" in the list IMO. CC @luizarakaki 
2. Luiz already mentioned this, but we want the TTL min query duration setting to be in seconds.
3. The "Cache time-to-live multiplier" setting [in the Figma has an "Example" bit of text](https://www.figma.com/file/x87hHuf4RabjMBuyV2Lq7c/caching-arc-low-fi-explorations?type=design&node-id=551-2015&mode=design&t=R3hWNsCaOpZRqzr0-11) that's meant to display a tooltip on hover.

rafpaf:
Thanks @mazameli. Replies below:

> 1. I think we should Performance after Permissions and before Tools in the top nav. It doesn't belong quite so "soon" in the list IMO. CC @luizarakaki 

Done

> 2. Luiz already mentioned this, but we want the TTL min query duration setting to be in seconds.

Fair point. Right now the FE mirrors the BE with respect to that field. I've scheduled the switch to seconds for [Milestone 2](https://github.com/metabase/metabase/issues/39163). Luiz and I talked about it in this [Slack thread](https://metaboat.slack.com/archives/C06KX7QECN4/p1712067660127329?thread_ts=1712063081.342889&cid=C06KX7QECN4)

> 3. The "Cache time-to-live multiplier" setting [in the Figma has an "Example" bit of text](https://www.figma.com/file/x87hHuf4RabjMBuyV2Lq7c/caching-arc-low-fi-explorations?type=design&node-id=551-2015&mode=design&t=R3hWNsCaOpZRqzr0-11) that's meant to display a tooltip on hover.

I've also scheduled this for [Milestone 2](https://github.com/metabase/metabase/issues/39163).

mazameli (approved):

frontend/src/metabase/admin/performance/hooks/useVerticallyOverflows.tsx
```diff
@@ -0,0 +1,13 @@
...
+  const [verticallyOverflows, setVerticallyOverflows] = useState(false);
+  const ref = useRef<HTMLElement>(null);
+  const el = ref.current;
+  useEffect(() => {
+    if (el) {
+      setVerticallyOverflows(el.scrollHeight > el.clientHeight);
+    }
+  }, [el?.scrollHeight, el?.clientHeight, el, setVerticallyOverflows]);
+  return { verticallyOverflows, ref };
+};
```
rafpaf, line 13: @sloansparger The variable should update when the scrollHeight or clientHeight changes - that's why it requires `useLayoutEffect`

frontend/src/metabase/admin/performance/components/StrategyFormLauncherPanel.tsx
```diff
@@ -0,0 +1,80 @@
...
+}: {
+  configs: Config[];
+  setConfigs: Dispatch<SetStateAction<Config[]>>;
+  targetId: number | null;
+  updateTargetId: UpdateTargetId;
+  databases: Database[];
+  isStrategyFormDirty: boolean;
+  shouldShowResetButton: boolean;
+}) => {
+  const databaseIds = databases.map(db => db.id);
```
sloansparger, line 34: memoize this value and maybe databases further up the tree

sloansparger (approved):
approved with last couple of comments

frontend/src/metabase/admin/performance/components/StrategyForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+      h="100%"
+      style={{
+        display: "flex",
+        flexDirection: "column",
+        justifyContent: "space-between",
+      }}
+    >
+      <Stack p="lg" spacing="xl" maw="27.5rem">
+        <StrategySelector targetId={targetId} />
+        {selectedStrategyType === "ttl" && (
```
iethree, line 104: This is fine for now, but when we have more than a few different strategies and some of them are in enterprise and some are not, I wonder if it might make sense to just have totally different form components for each strategy. That might also obviate the necessity for some of the form resets you're doing elsewhere: you don't need to swap around which fields are visible in the form if you just have different forms.

frontend/src/metabase/admin/performance/components/StrategyForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+              <PositiveNumberInput strategyType="ttl" name="multiplier" />
+            </Field>
+          </>
+        )}
+        {selectedStrategyType === "duration" && (
+          <>
+            <Field title={t`Cache results for this many hours`}>
+              <PositiveNumberInput strategyType="duration" name="duration" />
+            </Field>
+            <input type="hidden" name="unit" />
```
iethree, line 149: I'm not sure if this will work for you here, but if this field is in `initialValues` you should be able to work with it even without an invisible input.

frontend/src/metabase/admin/performance/components/StrategyForm.tsx
```diff
@@ -0,0 +1,293 @@
...
+  return (
+    <section>
+      <FormRadioGroup
+        label={
+          <Text lh="1rem">{t`When should cached query results be invalidated?`}</Text>
+        }
+        name="type"
+      >
+        <Stack mt="md" spacing="md">
+          {_.map(availableStrategies, (option, name) => {
```
iethree, line 217: nice, this will work well with enterprise-only strategies.

frontend/src/metabase/admin/performance/components/StrategyForm.tsx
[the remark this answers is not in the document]
iethree: I don't really like this comment here (are we going to comment this everywhere we use the correct properties?) I would support an educational lint rule though!

frontend/src/metabase/admin/performance/components/StrategyFormLauncher.tsx
iethree, line 1: I think this whole file needs to be in the enterprise folder

frontend/src/metabase/admin/performance/components/StrategyFormLauncherPanel.tsx
iethree, line 1: enterprise folder

frontend/src/metabase/admin/performance/hooks/useConfirm.tsx
iethree, line 1: duplicate of `useConfirmOnRouteLeave.tsx` right?

frontend/src/metabase/admin/performance/hooks/useVerticallyOverflows.tsx
iethree, line 1: I'm pretty sure we can do this without JS by changing the markup a bit

frontend/src/metabase/admin/performance/types.tsx
```diff
@@ -0,0 +1,152 @@
...
+  nocache: {
+    label: t`Don't cache results`,
+    validateWith: doNotCacheStrategyValidationSchema,
+    shortLabel: t`No caching`,
+  },
+  inherit: {
+    label: t`Use default`,
+    validateWith: inheritStrategyValidationSchema,
+  },
+};
```
iethree, line 42: this isn't a type, this should probably be in a `util` or `constants` file

frontend/src/metabase/admin/performance/types.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }
+  const type = Strategies[strategy.type];
+  return type.shortLabel ?? type.label;
+};
+
+const validStrategyNames = new Set(Object.keys(Strategies));
+
+export const isValidStrategyName = (
+  strategy: string,
+): strategy is StrategyType => validStrategyNames.has(strategy);
```
iethree, line 60: these look like utils

frontend/src/metabase/admin/performance/types.tsx
```diff
@@ -0,0 +1,152 @@
...
+};
+
+export const getFieldsForStrategyType = (strategyType: StrategyType) => {
+  const strategy = Strategies[strategyType];
+  const validationSchemaDescription =
+    strategy.validateWith.describe() as SchemaObjectDescription;
+  const fieldRecord = validationSchemaDescription.fields;
+  const fields = Object.keys(fieldRecord);
+  return fields;
+};
```
iethree, line 152: looks like a util

frontend/src/metabase/admin/performance/validation.ts
iethree, line 1: I really like how you put all the validation in its own file ⭐

frontend/src/metabase/admin/settings/selectors.js
```diff
@@ -563,39 +562,10 @@ export const ADMIN_SETTINGS_SECTIONS = {
...
-        allowValueCollection: true,
-      },
-      {
-        widget: SectionDivider,
-      },
       {
         key: "persisted-models-enabled",
         display_name: t`Models`,
-        description: jt`Enabling cache will create tables for your models in a dedicated schema and Metabase will refresh them on a schedule. Questions based on your models will query these tables. ${(
+        description: jt`Enabling caching will create tables for your models in a dedicated schema and Metabase will refresh them on a schedule. Questions based on your models will query these tables. ${(
```
iethree, line 568: 🏆

frontend/src/metabase/forms/components/FormProvider/FormProvider.tsx
```diff
@@ -17,15 +19,23 @@ export function FormProvider<T extends FormikValues, C = unknown>({
...
     validationContext,
   });
+  const value = useMemo(
+    () => ({
+      ...state,
+      setStatus: (status: FormStatus) =>
+        setState(state => ({ ...state, status })),
+    }),
+    [state, setState],
+  );
```
iethree, line 35: I don't hate this, but I wonder if you could avoid it by having different forms for different strategies?

iethree (dismissed):
👏  This is great, I love it.

A few suggestions and nitpicks, the only real concern is moving the enterprise code to the enterprise folder. The principle there isn't just that enterprise code is _disabled_ in the oss version, it has to actually not be in the oss bundle.

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.styled.tsx
```diff
@@ -0,0 +1,75 @@
...
+  &:first-child {
+    border-inline-start-width: 2px;
+    border-start-start-radius: 1rem;
+    border-end-start-radius: 1rem;
+  }
+  &:last-child {
+    border-inline-end-width: 2px;
+  }
+  ${props =>
+    !props.verticallyOverflows &&
```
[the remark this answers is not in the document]
rafpaf: To keep the ball rolling I've scheduled this for milestone 2

frontend/src/metabase/admin/performance/components/StrategyEditorForDatabases.tsx
```diff
@@ -0,0 +1,225 @@
...
+        setConfigs([...otherConfigs, newConfig]);
+      }
+    },
+    [configs, setConfigs, targetId],
+  );
+
+  const {
+    verticallyOverflows: formPanelVerticallyOverflows,
+    ref: formPanelRef,
+  } = useVerticallyOverflows();
```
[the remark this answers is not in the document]
rafpaf: I've scheduled this for Milestone 2 https://github.com/metabase/metabase/issues/39163

frontend/src/metabase/admin/performance/types.tsx
rafpaf: There are closely related and interdefined utils, constants, and types, so I've put most of them into a new `strategies.ts` file (containing strategy-related code). The Strategy and Config types now live in metabase-types/api, though.

frontend/src/metabase/forms/components/FormProvider/FormProvider.tsx
rafpaf: I've scheduled an exploration of this idea for Milestone 2. I haven't tried this approach yet because I think it would probably mean putting the strategy radio button outside the form (though I might be wrong).

rafpaf:
### Merge activity

* **Apr 3, 12:04 PM EDT**: @rafpaf merged this pull request with [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/39234).

PR #39235, opened 2024-02-28
No more Schema, only Malli (finish the migration)

camsaul:
Convert the last few Schema schemas to Malli and remove schema as a dependnency


escherize (approved):
Thank you for writing this!

A couple minor things, nothing blocking though.

Thanks for splitting up so many deftest forms as well.

crisptrutski:
This is epic 💪 😍

PR #39244, opened 2024-02-28
Fix missing bucket picker on selected item

romeovs:
Fixes an issue where the bucket popover was not visible when the summarize items is selected.

Screenshot 2024-02-28 at 12 54 56


kamilmielnik (approved):
:+1:

PR #39251, opened 2024-02-28
Do not fetch app settings when updating user-local setting

### LINKED ISSUE #39221: Setting or updating a `user-local` setting by a non-admin is followed by fetching all settings, which results in `403`
We have so called `user-local` settings. They can be limited to admins, setting managers or allowed to be used by any authenticated user. Read more about it [here](https://github.com/metabase/metabase/blob/master/src/metabase/models/setting.clj#L1098).

The problem arises when a non-admin user sets or updates the setting.
The `updateSetting` action (as defined [here](https://github.com/metabase/metabase/blob/master/frontend/src/metabase/admin/settings/settings.js#L53)) then fetches all settings which is something that's allowed only by admins, or superusers.
https://github.com/metabase/metabase/blob/7f0d671/src/metabase/api/setting.clj#L28-L29

## Current behavior
- Non-admin sets or updates the setting.
- Frontend tries to fetch all settings
- The endpoint returns `403`

## Expected behavior
- Non-admin sets or updates the setting.
- Frontend either fetches `/api/session/properties` or it skips the fetching logic altogether

```[tasklist]
### User-local settings not restricted to admins
- [x] last-used-native-database-id (#39251)
- [ ] dismissed-browse-models-banner (#39359)
- [ ] dismissed-custom-dashboard-toast (#39359)
```

### Related
#39223

nemanjaglumac:
### Description

As explained in the [original issue](https://github.com/metabase/metabase/issues/39221), we have two types of settings. App, or site, settings are accessible only by admins. OTOH, there are user-local settings that should be accessible to non-admins.

Before this PR
- Updating any setting (app or user) would result in fetching all settings

Now
- Updating admin settings results in fetching all settings (we're not changing this)
- Updating a user-local setting results in fetching only settings available to all users (we get these from `/api/session/properties` endpoint)

### How to verify
- Log in as non-admin user
- Start a new native query or a new native model
- Choose a database
- Make sure there is no `GET /api/setting` call in your browser's network tab (previously resulted in `403`)

or

Simply run a provided E2E reproduction.

### Demo
Before
image

After
image

### Checklist

Fixes #39221.

### DIFF
--- a/e2e/test/scenarios/permissions/reproductions/39221-fetch-settings-only-if-admin.cy.spec.js
+++ b/e2e/test/scenarios/permissions/reproductions/39221-fetch-settings-only-if-admin.cy.spec.js
@@ -15,21 +15,15 @@ describe("issue 39221", () => {
     });
   });
 
-  it("admins should fetch site settings", () => {
-    startNewNativeQuestion();
-    popover().findByText("Sample Database").click();
-    cy.wait("@sessionProperties");
+  ["admin", "normal"].forEach(user => {
+    it(`${user.toUpperCase()}: updating user-specific setting should not result in fetching all site settings (metabase#39221)`, () => {
+      user === "normal" ? cy.signInAsNormalUser() : null;
 
-    cy.get("@siteSettings").should("not.be.null");
-  });
+      startNewNativeQuestion();
+      popover().findByText("Sample Database").click();
+      cy.wait("@sessionProperties");
 
-  it("non-admins should not fetch site settings (metabase#39221)", () => {
-    cy.signInAsNormalUser();
-
-    startNewNativeQuestion();
-    popover().findByText("Sample Database").click();
-    cy.wait("@sessionProperties");
-
-    cy.get("@siteSettings").should("be.null");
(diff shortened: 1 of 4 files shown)


ranquild (changes requested):
-

ranquild (changes requested):
FWIW, the fix is incorrect and won't work for setting managers. Trying to emulate/guess the correct permission check would either fail or lead to unmaintable code. But that's not the point I want to make.

The point is that admin actions are intended to be used for admin features only. `/api/setting` is only called there and has additional metadata such as whether the setting is set by an env variable, whether is has the default value, what the actual user-supplied value is etc. This information shouldn't be available for regular cases.

What I think we should do instead is create a separate action that updates user-local settings and put it where we currently load user settings. There is a file called `redux/settings.ts`  where we call `/api/setting/properties`. We should add an action e.g. `updateUserSetting` or whatever and reload only `/api/setting/properties` there (with `refreshSiteSettings` call).

ranquild (approved):

PR #39252, opened 2024-02-28
Add hovercard to column picker in sandbox filter

### LINKED ISSUE #39240: Add hover card when picking a column to use for simple sandbox filtering
As part of #38394 we want to add the info icon to the sandbox filter picker.

Screenshot 2024-02-28 at 16 42 33

romeovs:
Closes https://github.com/metabase/metabase/issues/39240

### Description

Adds the info icon to the columns in the column picker in the sandbox permission editor.

### How to verify

1. Admin → Permissions → Databases → Sample Database → Orders
2. For "All Users" select "Sandboxed"
3. In the popover select "Limit access"
4. Select "Filter by a column in the table"
5. Click "Pick a column"
6. In the popover, verify that all items have an info icon on hover
7. Verify that the info shown in the popover is correct when hovering an icon

### Demo

Screenshot 2024-02-28 at 16 42 33

### Checklist

### DIFF
--- a/frontend/src/metabase/parameters/utils/mapping-options.ts
+++ b/frontend/src/metabase/parameters/utils/mapping-options.ts
@@ -36,6 +36,9 @@ export type StructuredQuerySectionOption = {
   icon: string;
   target: StructuredParameterDimensionTarget;
   isForeign: boolean;
+  query?: Lib.Query;
+  stageIndex?: number;
+  column?: Lib.ColumnMetadata;
 };
 
 function buildStructuredQuerySectionOptions(
@@ -55,6 +58,9 @@ function buildStructuredQuerySectionOptions(
       icon: getColumnIcon(column),
       target: buildColumnTarget(query, stageIndex, column),
       isForeign: columnInfo.isFromJoin || columnInfo.isImplicitlyJoinable,
+      query,
+      stageIndex,
+      column,
     };
   });
 }
--- a/frontend/src/metabase/parameters/utils/mapping-options.ts
+++ b/frontend/src/metabase/parameters/utils/mapping-options.ts
@@ -36,9 +36,9 @@ export type StructuredQuerySectionOption = {
   icon: string;
   target: StructuredParameterDimensionTarget;
   isForeign: boolean;
-  query?: Lib.Query;
-  stageIndex?: number;
-  column?: Lib.ColumnMetadata;
+  query: Lib.Query;
+  stageIndex: number;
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/parameters/utils/mapping-options.ts
```diff
@@ -55,6 +58,9 @@ function buildStructuredQuerySectionOptions(
       icon: getColumnIcon(column),
       target: buildColumnTarget(query, stageIndex, column),
       isForeign: columnInfo.isFromJoin || columnInfo.isImplicitlyJoinable,
+      query,
+      stageIndex,
```
kamilmielnik, line 62: It feels like these 2 values don't really fit into `StructuredQuerySectionOption`.
We have an example of this in [`FieldPicker.tsx`](https://github.com/metabase/metabase/blob/2739b5cf9bea0f96ce62a90f017f2daebba3b84b/frontend/src/metabase/common/components/FieldPicker/FieldPicker.tsx#L97-L100) - `item` has `column`, but `query` and `stageIndex` come from component props - not the `item`.

WDYT about making it the same here?

frontend/src/metabase/parameters/utils/mapping-options.ts
```diff
@@ -36,6 +36,9 @@ export type StructuredQuerySectionOption = {
   icon: string;
   target: StructuredParameterDimensionTarget;
   isForeign: boolean;
+  query?: Lib.Query;
+  stageIndex?: number;
+  column?: Lib.ColumnMetadata;
```
kamilmielnik, line 41: When can these be `undefined`? Because it feels they should not never be.

kamilmielnik (approved):
Looks good, works well :+1: 

Left some minor comments.

frontend/src/metabase/parameters/utils/mapping-options.ts
romeovs: That's what I tried first but there is some [custom logic](https://github.com/metabase/metabase/pull/39252/files/34ed51066b044881b1da43ebea99a7796dfd048d#diff-43ecf70f8f11873a26920b1b0a37628e34427cfaa5ed16c8fc520a832c1d65e9R162-R169) for handling models in the `getParameterMappingOptions` that I wanted to avoid duplicating.

kamilmielnik: Fair enough :+1:

Issue #37348, opened 2024-01-08
BE - Add "metric" card type


kamilmielnik:
Depends on:
- https://github.com/metabase/metabase/issues/38706

----

- [ ] Add a new value to enum from #37367.
- [ ] MLv2 schema and other Malli schemas should allow this
- [ ] Add to TypeScript signatures?
- [ ] Rest API should accept this when creating a new Card. Don't worry about additional constraints, we'll do that in #38733

crisptrutski:
Hi @kamilmielnik, let me know if you want to take over the task that blocks this. I'm brand new here and will be pretty slow on the task. I have my first pairing session with my [onboarding buddy](https://github.com/calherries) tomorrow as another option to speed things up.

kamilmielnik:
> Hi @kamilmielnik, let me know if you want to take over the task that blocks this. I'm brand new here and will be pretty slow on the task. I have my first pairing session with my [onboarding buddy](https://github.com/calherries) tomorrow as another option to speed things up.

@crisptrutski  Actually that task was linked by mistake. Updated. Thanks for staying alert!

camsaul:
This was partially done in #38981, we probably need to add some more stuff in and add some tests to make sure it works end to end

PR #39262, opened 2024-02-28
Enable saving metric type cards

metamben:
Part of #37348.

### DIFF
--- a/frontend/src/metabase/entities/questions.js
+++ b/frontend/src/metabase/entities/questions.js
@@ -6,8 +6,6 @@ import Collections, {
   getCollectionType,
   normalizedCollection,
 } from "metabase/entities/collections";
-import { isCypressActive, isProduction, isTest } from "metabase/env";
-import { GET, POST, PUT } from "metabase/lib/api";
 import { color } from "metabase/lib/colors";
 import { createEntity, undo } from "metabase/lib/entities";
 import * as Urls from "metabase/lib/urls";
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/entities/questions.js
```diff
@@ -40,55 +40,21 @@ const Questions = createEntity({
...
           get: async payload => {
             const get = GET("/api/card/:id");
-            const result = await get(payload);
-
-            if (result.name.toLowerCase().includes("metric")) {
-              return { ...result, type: "metric" };
-            }
-
-            return result;
+            return await get(payload);
```
kamilmielnik, line 43: This whole section should be removed, lines 31-95.

kamilmielnik: Done: f7651d414f8777c0ece2985c5b60ecfdf6331b76

ranquild (approved):

test/metabase/api/card_test.clj
```diff
@@ -1035,6 +1035,14 @@
                :type    "question"}
               (mt/user-http-request :crowberto :post 200 "card" (card-with-name-and-query (mt/random-name))))))))
 
+(deftest create-card-with-metric-type
+  (mt/with-model-cleanup [:model/Card]
+    (testing "can create a metric card"
+      (is (=? {:dataset false
+               :type    "metric"}
+              (mt/user-http-request :crowberto :post 200 "card" (assoc (card-with-name-and-query (mt/random-name))
```
snoe, line 1043: should we be checking that a metric card's query conforms to the rules around metrics (must have 1 aggregation)?

snoe (approved):
LGTM - I assume the query check can be addressed in the future.

PR #39263, opened 2024-02-28
Utilize diagnoseExpression to detect expression errors

### LINKED ISSUE #38944: Custom column disappears after creating with case(isempty([Some Column]), "true", "false")
### Describe the bug

Creating a custom column like:
- `case(isempty([Some Column]), "true", "false")`
- disappears after clicking Done.

### To Reproduce

1. Create a new question using the Sample Database > Accounts table
2. Add a custom column `case(isempty([Active Subscription] ),"true","false")` called "active subscription is empty"
3. Click Done
4. No custom column appears

### Expected behavior

Custom column should appear after creating it

### Logs

_No response_

```JSON
- `master` as of `f509213`
```

### Severity

P1: Major functionality not working

### Additional context

_No response_

### LINKED ISSUE #38954: [MLv2] fromLegacyQuery drops invalid expressions
**Context**

`fromLegacyQuery` drops valid expression and returns incorrect result

Screenshot 2024-02-20 at 14 58 30

### Steps to reproduce

```
Lib.fromLegacyQuery(
        databaseId,
        metadata,
        this.datasetQuery(),
      );
```

content of `this.datasetQuery()`

```
{
    "database": 1,
    "type": "query",
    "query": {
        "source-table": 6,
        "expressions": {
            "333": [
                "case",
                [
                    [
                        [
                            "is-empty",
                            [
                                "field",
                                5,
                                {
                                    "base-type": "type/Boolean"
                                }
                            ]
                        ],
                        "true"
                    ]
                ],
                {
                    "default": "false"
                }
            ]
        }
    }
}
```

and MLv2 query was 
image

- issue links: https://github.com/metabase/metabase/issues/38944

### LINKED ISSUE #39438: Not possible to save static text as a custom expression
### Describe the bug

I've spotted a weird behaviour on [stats.metabase.com](http://stats.metabase.com/), if I put any string like 'test' or "asd" as a custom column expression and click outside the field, the page breaks.
I'm not sure if this is supported, I wanted to create a column with a fixed text value.
But even if this is not supported, a custom expression syntax error should be shown, I suppose.
Bug of MBQLv2?
Captura de ecrã 2024-02-29, às 23 29 36
Captura de ecrã 2024-02-29, às 23 29 46

### To Reproduce

1. Open query builder for any dataset
2. Click on custom column button
3. Enter "test" as expression
4. Click outside the expression field
5. The page breaks

### Expected behavior

I believe the static text should be possible to save as a custom column value

### Logs

react-dom.production.min.js:209 Error: Unknown MBQL clause ["test"]
    at e (format.ts:74:9)
    at tf.UNSAFE_componentWillReceiveProps (ExpressionEditorTextfield.tsx:167:22)
    at oR (react-dom.production.min.js:135:172)
    at iF (react-dom.production.min.js:179:157)
    at l (react-dom.production.min.js:263:490)
    at aH (react-dom.production.min.js:246:265)
    at react-dom.production.min.js:246:194
    at aW (react-dom.production.min.js:246:199)
    at react-dom.production.min.js:123:115
    at t.unstable_runWithPriority (scheduler.production.min.js:19:467)

```JSON
stats.metabase.com
```

### Severity

Blocks some (rare) use cases

### Additional context

_No response_

uladzimirdev:
closes https://github.com/metabase/metabase/issues/38944
closes https://github.com/metabase/metabase/issues/38954
closes https://github.com/metabase/metabase/issues/39438

### Description

Invalid expressions are dropped by MLv2, before migration invalid expressions weren't dropped, but they couldn't be executed anyway, so this PR uses `diagnoseExpression` method from MLv2 to detect invalid expressions.

Currently, it's not possible to highlight the exact place with an error and get user-friendly error message from MLv2, so we use a generic message `Invalid expression`.

### How to verify

follow steps from the changed e2e test

or 

- Create a new question using the Sample Database > Orders table
- Add a custom column `case(isempty([Discount]), "true", "false")` called "discount is empty"
- Done button should be disabled and you should see a message "invalid expression"

### Demo

image

### Checklist

### DIFF
--- a/frontend/src/metabase-lib/expressions/diagnostics.ts
+++ b/frontend/src/metabase-lib/expressions/diagnostics.ts
@@ -47,7 +47,7 @@ export function diagnose({
   expressionPosition,
 }: {
   source: string;
-  startRule: "expression" | "aggregation" | "boolean";
+  startRule: Lib.ExpressionMode;
   query: Lib.Query;
   stageIndex: number;
   name?: string | null;
@@ -95,7 +95,7 @@ export function diagnose({
     return { message };
   }
 
-  // try to compile on FE
+  // make initial light diagnostics
   let mbqlOrError: Expr | ErrorWithMessage;
   try {
     mbqlOrError = prattCompiler({ source, startRule, name, query, stageIndex });
@@ -120,18 +120,11 @@ export function diagnose({
     return { message: t`Invalid expression` };
   }
 
-  // now diagnose on BE
-  const startRuleToExpressionModeMapping: Record<string, Lib.ExpressionMode> = {
-    boolean: "filter",
-  };
-
-  const expressionMode: Lib.ExpressionMode =
-    startRuleToExpressionModeMapping[startRule] ?? startRule;
-
+  // if first passed, we'll do a more thorough check
   const possibleError = Lib.diagnoseExpression(
     query,
     stageIndex,
-    expressionMode,
+    startRule,
     mbqlOrError,
     expressionPosition,
   );
--- a/frontend/src/metabase-lib/expressions/diagnostics.ts
+++ b/frontend/src/metabase-lib/expressions/diagnostics.ts
@@ -111,8 +111,6 @@ export function diagnose({
       );
     }
   } catch (err) {
-    console.warn("compile error", err);
-
     if (isErrorWithMessage(err)) {
       return err;
     }
--- a/frontend/src/metabase-lib/expressions/diagnostics.ts
+++ b/frontend/src/metabase-lib/expressions/diagnostics.ts
@@ -47,7 +47,7 @@ export function diagnose({
   expressionPosition,
 }: {
   source: string;
-  startRule: Lib.ExpressionMode;
+  startRule: "expression" | "aggregation" | "boolean";
   query: Lib.Query;
   stageIndex: number;
   name?: string | null;
@@ -95,7 +95,7 @@ export function diagnose({
     return { message };
   }
 
-  // make initial light diagnostics
+  // try to compile on FE
   let mbqlOrError: Expr | ErrorWithMessage;
   try {
     mbqlOrError = prattCompiler({ source, startRule, name, query, stageIndex });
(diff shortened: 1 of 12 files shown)


ranquild (approved):
Could we clean up `startRule` to match `ExpressionMode`? And remove console statements

frontend/src/metabase-lib/expressions/diagnostics.ts
```diff
@@ -88,19 +90,61 @@ export function diagnose({
...
+    console.warn("compile error", err);
+
     if (isErrorWithMessage(err)) {
       return err;
     }
 
     return { message: t`Invalid expression` };
   }
+
+  // now diagnose on BE
```
metamben, line 123: Would "BE" not confuse a newcomer? They might assume the browser talks to the BE, but nothing like that is happening. I think eventually we should correct our terminology and talk about ML instead of BE (or something similar).

frontend/src/metabase-lib/expressions/diagnostics.ts
```diff
@@ -88,19 +90,61 @@ export function diagnose({
...
+  const possibleError = Lib.diagnoseExpression(
+    query,
+    stageIndex,
+    expressionMode,
+    mbqlOrError,
+    expressionPosition,
+  );
+
+  if (possibleError) {
+    console.warn("diagnostic error", possibleError.message);
```
metamben, line 140: `possibleError.message` is a translated message, so I wonder if we can do something better than just throwing away.

metamben (approved):
The Clojure part looks good to me. :)

frontend/src/metabase-lib/expressions/diagnostics.ts
metamben: What about keeping the part before the colon? I think `Type error` might help a bit. And it's very different from the error you get if you try to create circular definitions.

uladzimirdev: I don't think `Type error` is good enough for the user

uladzimirdev:
> Could we clean up `startRule` to match `ExpressionMode`? And remove console statements

@ranquild I tried to use ExpressionMode for a startRule, but the change seems to be huge because of testing functions like `createQueryWithBooleanFilter` which I don't see how to rework atm, I'll create a separate PR to address it

PR #39266, opened 2024-02-28
RTKQuery Implementation: AuditInfo

iethree:
## Description

Replaces custom reducers and selectors for audit info api caching in favor of using RTK Query for fetching data from audit info endpoint.

This also establishes a pattern for defining certain api hooks within the enterprise directory at `metabase-enterprise/api`

Screen Shot 2024-03-15 at 11 16 03 AM


sloansparger (approved):
I think you need to rebase, but LGTM

PR #39269, opened 2024-02-28
Do not return value if the database id used for `last-used-native-database-id` does not exist (anymore)

nemanjaglumac:
This PR is a follow up after #38986.

### Description

It is quite possible that a user persists a database (id), and that this database gets deleted afterwards.
Even though the frontend handles this pretty well (which is confirmed by the existing E2E test), we wanted to follow up with a custom `:getter` for this particular setting that would return `nil` if the persisted database doesn't exist anymore.

This PR tries to achieve exactly that.
It updates and expands the previously written backend unit tests to reflect this new condition.

### How to verify
At the very minimum, all tests should pass.

You can do this manually:
1.  Send a `PUT /api/setting/last-used-native-database-id`. For the `value` use a database id value that doesn't exist (e.g. 999)
2.  Call `GET /api/session/properties` and make sure that `last-used-native-database-id` is `null`

### Checklist

### DIFF
--- a/test/metabase/models/user_test.clj
+++ b/test/metabase/models/user_test.clj
@@ -538,29 +538,32 @@
 (deftest last-used-native-database-id-can-be-read-and-set
   (testing "last-used-native-database-id can be read and set"
     (mt/with-test-user :rasta
-      (let [old-db-id (user/last-used-native-database-id)
-            new-db-id 42]
-        (try
-          (is (not= new-db-id old-db-id))
-          (user/last-used-native-database-id! new-db-id)
-          (is (= new-db-id (user/last-used-native-database-id)))
-          (finally
-            (user/last-used-native-database-id! old-db-id))))))
+      (let [initial-value  (user/last-used-native-database-id)
+            existing-db-id (:id (t2/select-one Database))
+            wrong-db-id    999]
+        (is (nil? initial-value))
+        (user/last-used-native-database-id! existing-db-id)
+        (is (= existing-db-id (user/last-used-native-database-id)))
+        (testing "returns nil if the database doesn't exist"
+          (user/last-used-native-database-id! wrong-db-id)
+          (is (nil? (user/last-used-native-database-id)))))))
 
   (testing "last-used-native-database-id should be a user-local setting"
     (is (=? {:user-local :only}
-            (setting/resolve-setting :last-acknowledged-version)))
-    (mt/with-test-user :rasta
-      (let [old-db-id (user/last-used-native-database-id)]
-        (user/last-used-native-database-id! 42)
-        (mt/with-test-user :crowberto
-          (let [old-db-id (user/last-used-native-database-id)]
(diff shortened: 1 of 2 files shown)


ranquild (approved):
Please wait for the BE approval

test/metabase/models/user_test.clj
```diff
@@ -538,29 +538,32 @@
...
-            new-db-id 42]
-        (try
-          (is (not= new-db-id old-db-id))
-          (user/last-used-native-database-id! new-db-id)
-          (is (= new-db-id (user/last-used-native-database-id)))
-          (finally
-            (user/last-used-native-database-id! old-db-id))))))
+      (let [initial-value  (user/last-used-native-database-id)
+            existing-db-id (:id (t2/select-one Database))
+            wrong-db-id    999]
```
metamben, line 543: This is a bit brittle. Can we set a negative value? (Normal DB IDs are positive.)

test/metabase/models/user_test.clj
```diff
@@ -538,29 +538,32 @@
...
+        (let [old-db-id (user/last-used-native-database-id)]
+          (user/last-used-native-database-id! id1)
+          (mt/with-test-user :crowberto
+            (let [old-db-id (user/last-used-native-database-id)]
+              (user/last-used-native-database-id! id2)
+              (is (= (user/last-used-native-database-id) id2))
+              (mt/with-test-user :rasta
+                (is (= (user/last-used-native-database-id) id1)))
+              (user/last-used-native-database-id! old-db-id)))
+          (user/last-used-native-database-id! old-db-id))))))
```
metamben, line 566: It looks as if we tried to return to the original settings. If that's indeed the intent, then it's better done in a `finally` form.
```suggestion
          (try
            (user/last-used-native-database-id! id1)
            (mt/with-test-user :crowberto
              (let [old-db-id (user/last-used-native-database-id)]
                (try
                  (user/last-used-native-database-id! id2)
                  (is (= (user/last-used-native-database-id) id2))
                  (mt/with-test-user :rasta
                    (is (= (user/last-used-native-database-id) id1)))
                  (finally
                    (user/last-used-native-database-id! old-db-id))))
             (finally
               (user/last-used-native-database-id! old-db-id)))))))
```

metamben (approved):
LGTM, just a couple of small things that might make the test more robust.

test/metabase/models/user_test.clj
nemanjaglumac: I ended up using `mt/discard-setting-value` to simplify this a bit.
https://github.com/metabase/metabase/pull/39269/commits/2d7328066b11570760b8941ea5b0831b3010763d

Please see [this Slack convo](https://metaboat.slack.com/archives/CKZEMT1MJ/p1709558933798589) for context.

metamben (approved):
LGTM. It seems it can be further simplified by using `with-temporary-setting-values`.

PR #39280, opened 2024-02-28
Rename :model/Metric to :model/LegacyMetric

metamben:


enterprise/backend/src/metabase_enterprise/serialization/v2/models.clj
```diff
@@ -51,7 +51,7 @@
    "HTTPAction"
    "ImplicitAction"
    "LoginHistory"
-   "MetricImportantField"
```
camsaul, line 54: We might actually be able to just drop MetricImportantField, I'm not sure this is even used anymore TBH. If it is we can probably sunset that immediately since Metrics v1 are going away anyway

src/metabase/models/metric.clj
```diff
@@ -30,30 +30,30 @@
 (def Metric
```
camsaul, line 30: I was thinking we should rename this and rename the namespaces as well e.g. `LegacyMetric` and `metabase.models.legacy-metric`.

Ideally we'd even rename the API endpoints, it wouldn't be too hard to update the FE code that calls them since they're all in one place IIRC. But maybe that's less important since Metrics v1 are going away soon anyway

camsaul (approved):
I think we should go ahead and merge this in now, but maybe follow it on with a couple more PRs

- remove MetricImportantFields and related API endpoints e.g. `PUT /api/metric/:id/important_fields` altogether, I am 99% sure these aren't even used anymore (grepped the codebase for `important_fields` and couldn't find any references to it in the FE codebase)
- rename `metabase.models.metric` to `metabase.models.legacy-metric` so we can free up that namespace for a potential Toucan 2 sub-model of Card or other Metrics v2 model code

metamben:
Another PR has been merged.

PR #39284, opened 2024-02-29
Persistable with-temp

### LINKED ISSUE #39282: A way to make `with-temp` to persist test data
It's useful for debugging to avoid cleaning data created by `with-temp`.

Let's devise a way to provide an escape hatch to do that easily.

Proposal: create a `with-temp-persisted` macro that does what `with-temp` does except it cleans up data afterward.

qnkhuat:
```clojure
(t2/exists? :model/Card (mt/with-temp!-persisted [:model/Card card {}]
                          (:id card)))
;; => true
```

Closes https://github.com/metabase/metabase/issues/39282

### DIFF
--- a/test/metabase/test/redefs.clj
+++ b/test/metabase/test/redefs.clj
@@ -62,7 +62,7 @@
         (next-method model attributes f)))
     (next-method model attributes f)))
 
-(defmacro with-temp-persisted!
+(defmacro with-temp!-persisted
   "Like [[mt/with-temp]], but all the created items are persisted after."
   [& args]
   `(binding [*with-temp-clean-up*           false
--- a/test/metabase/test/redefs.clj
+++ b/test/metabase/test/redefs.clj
@@ -2,16 +2,49 @@
   "Redefinitions of vars from 3rd-party namespaces to make sure they do extra stuff we want (like initialize things if
   needed when running)."
   (:require
+   [clojure.pprint :as pprint]
+   [clojure.test :as t]
    [mb.hawk.parallel]
    [metabase.plugins.classloader :as classloader]
    [metabase.test.util.thread-local :as tu.thread-local]
+   [metabase.util.log :as log]
    [methodical.core :as methodical]
    [toucan2.connection :as t2.connection]
-   [toucan2.tools.with-temp :as t2.with-temp]))
+   [toucan2.core :as t2]
+   [toucan2.tools.with-temp :as t2.with-temp]
+   [toucan2.util :as t2.u]))
 
 (def ^:dynamic ^:private *in-tx*
   "Used to detect whether we're in a nested [[with-temp]]. Default is false."
   false)
+
+(def ^:dynamic ^:private *with-temp-clean-up*
+  "Whether with-temp should clean up afterward, default is true."
+  true)
+
+(methodical/defmethod t2.with-temp/do-with-temp* :default
+  [model explicit-attributes f]
+  (assert (some? model) (format "%s model cannot be nil." `with-temp))
+  (when (some? explicit-attributes)
+    (assert (map? explicit-attributes) (format "attributes passed to %s must be a map." `with-temp)))
+  (let [defaults          (t2.with-temp/with-temp-defaults model)
+        merged-attributes (merge {} defaults explicit-attributes)]
+    (t2.u/try-with-error-context ["with temp" {::model               model
+                                               ::explicit-attributes explicit-attributes
+                                               ::default-attributes  defaults
+                                               ::merged-attributes   merged-attributes}]
+     (log/debugf "Create temporary %s with attributes %s" model merged-attributes)
+     (let [temp-object (t2/insert-returning-instance! model merged-attributes)]
+       (log/debugf "[with-temp] => %s" temp-object)
+       (try
+        (t/testing (format "\nwith temporary %s with attributes\n%s\n"
+                           (pr-str model)
+                           #_{:clj-kondo/ignore [:discouraged-var]}
+                           (with-out-str (pprint/pprint merged-attributes)))
+          (f temp-object))
+        (finally
+         (when *with-temp-clean-up*
+           (t2/delete! model :toucan/pk ((t2/select-pks-fn model) temp-object)))))))))
 
 (methodical/defmethod t2.with-temp/do-with-temp* :around :default
   "Initialize the DB before doing the other with-temp stuff.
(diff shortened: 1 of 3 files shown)


test/metabase/test/redefs.clj
```diff
@@ -2,17 +2,50 @@
...
 
 (def ^:dynamic ^:private *in-tx*
   "Used to detect whether we're in a nested [[with-temp]]. Default is false."
   false)
 
+(def ^:dynamic ^:private *with-temp-clean-up*
+  "Whether with-temp should clean up afterward, default is true."
+  true)
+
+(methodical/defmethod t2.with-temp/do-with-temp* :default
```
qnkhuat, line 25: faithfully copied from `toucan2`, the only difference is in the finally clause

test/metabase/test/redefs.clj
```diff
@@ -29,6 +62,13 @@
         (next-method model attributes f)))
     (next-method model attributes f)))
 
+(defmacro with-temp-persisted!
+  "Like [[mt/with-temp]], but all the created items are persisted after."
+  [& args]
+  `(binding [*with-temp-clean-up*           false
+             tu.thread-local/*thread-local* true]
```
calherries, line 69: Can you explain why `tu.thread-local/*thread-local*` is set to `true`?

tsmacdonald:
1. How do we feel about adding this directly to Toucan2, so that:
  a. Other projects can benefit
  b. We're not reusing code by copy-and-paste (and we get any future upstream updates automatically)
  c. It'd  make sense to document it alongside the `with-temp` documentation, which is in T2.

2.  I'm not in love with `with-temp-persisted!`; there's nothing temporary about them. Options I can think of: `with-persisted!` `with-scaffolding!` `with-models!` `with-test-models!`

test/metabase/test/redefs.clj
qnkhuat: It should be `false`! I need to set this so we don't run `with-temp` in a transaction.

I think the reason why test passed is because `{:rollback-only true}` from `t2/with-transaction` does not actually works :)

qnkhuat:
@tsmacdonald I'm not sure how useful is it for other because this is meant to be used during development only. What I imagine the workflow is:
1. you have a failing test
2. you change `with-temp!` to `with-temp-perisisted!` 
3. do the magic and fix the bug
4. revert `with-temp-persisted!` back to `with-temp!`

Do you think that's beneficial from a library level?

About naming: the reason why I chose to make it a prefix is because I just want to change between the two versions quickly :p because this is not meant to be committed. lmk if that resonates with you

tsmacdonald:
> Do you think that's beneficial from a library level?

Yeah, any other project that was using T2 would probably want the exact same workflow in their own tests.

> About naming: the reason why I chose to make it a prefix is because I just want to change between the two versions quickly :p because this is not meant to be committed. lmk if that resonates with you

That's totally a reasonable desire, but surely typing the extra word (`s/temp/models!` or whatever) is pretty quick? I guess another option is using metadata

```clojure
(with-temp ...)

(with-temp ^:persisted ...) ; useful for a quick experiment

(with-temp {:persisted true} ...) ;useful for toggling back and forth a lot
(with-temp {:persisted false} ...)
```

but that might be overengineering.

qnkhuat:
also another reason with post-fixing is that it's more discoverable. if you type `with-temp!` and `with-temp!-persisted` will show up in your autocomplete ;)

test/metabase/test/redefs.clj
crisptrutski: This duplication is unfortunate. It looks to me like we could avoid it if we merely decorated `f`:

```clojure
(fn [temp-object]
  (try 
    (f temp-object)
    (finally 
      (when *with-temp-clean-up*
        ...))))
```

This just assumes that nothing in the `t/testing` call will throw, which seems fair. Having the clean-up inside this testing scope also seems fine.

crisptrutski: Oh nevermind, I should have looked at the base implementation first 🙈

crisptrutski: This duplication feels off to me, I feel like it's the lesser of two evils to add the dynamic var upstream.

This debugging flow seems like something generally useful to toucan2 users in anycase.

test/metabase/test/redefs.clj
```diff
@@ -29,6 +62,13 @@
         (next-method model attributes f)))
     (next-method model attributes f)))
 
+(defmacro with-temp!-persisted
```
crisptrutski, line 65: Have you thought about breaking this up into orthogonal parts? i.e. a `with-cleanup-disabled` macro that just updates the binding. Two advantages I see:

1. More obvious that the behavior will cascade into nested `with-temp!` calls.
2. Avoid the awkward naming of "temporary but not really". A good enough name to go into toucan2 IMHO.

I guess one thing you'd lose is that "visibility in search" aspect, but I think that can be readily compensated for by mentioning this utility in the `with-temp!` docstring.

crisptrutski (commented):
This looks useful, and simple enough, but I'd seriously consider the interface change and moving it upstream to make nested usage less surprising, and be less brittle to upstream changes.

test/metabase/test/redefs.clj
tsmacdonald: >Another downside is that wrapping the form will trigger indentation and that's annoying, and could lead to whitespace changes to rollback after unwrapping.

Hopefully you just hit `paredit-raise` on the `with-temp!` form and it removes the `with-cleanup-disabled` and fixes the indentation all at once.

Or, if you're feeling saucy,

```
(with-cleanup-diabled (with-temp! ...
   ...))
```

crisptrutski: Yeah I'd probably do that one liner.

The issue I'm talking about is that when I surround a form in my editor, it triggers reformatting for the inner forms. This would then wipe out any manual alignment, e.g. that done for the multiple columns in a multi-line `with-temp!`...

piranha: I wrote a test which required 7 nested macros and I'm thinking if we should invent some kind of `application macro` which can then wrap body in list of somethings. Like `with-open`, maybe called `with-wrap`? Something like Python's `with`, which will call `__enter__` and `__exit__` on a provided object.

crisptrutski:
This is starting to rot - should we finish it off, or close for now?

qnkhuat:
I'll try to find some time to work on this next week :) I think it's useful.

PR #39285, opened 2024-02-29
Fixing E2E Cross-version Tests

### LINKED ISSUE #39229: Get e2e-cross-version.yml working
The `.github/workflows/e2e-cross-version.yml` action is consistently failing. We want to get this working again. See [this Slack thread](https://metaboat.slack.com/archives/C0669P4AF9N/p1708515050922759) for more context.

markbastian:
E2E cross version tests were broken for multiple reasons, primarily due to changes made in UI layout, cypress testing, and data changes. Overall, the basic workflows still work, but subtle click order differences and other small changes made the entire job consistently fail.

To fix this, this PR:
- Provides a function, `parseVersionString`, which computes version information provided a test version string. This information includes the raw string version, the edition (ee or oss), and the major, minor, and patch version of the version string.
- In the same file, `cross-version-helpers.js`, adds constants identifying versions in which certain breaking UI or Cypress changes occurred. 
- Finally, a set of exported functions are provided which take a version object and execute conditional logic based on the provided version so that behavior is consistent across Metabase versions.

The versions are computed in `cross-version-source-helpers.js` and `cross-version-target-helpers.js` and are used in js files as shown:

```js
import { visualize } from "e2e/support/helpers";
import {
  fillAreaUnderLineChart,
  newQuestion,
  saveQuestion,
} from "e2e/test/scenarios/cross-version/helpers/cross-version-helpers.js";
```

This then allows logic like the below, where the high level call is made and the right logic is dispatched based on the version information.

```js
it("should create questions", () => {
  cy.signInAsAdmin();

  cy.visit("/question/new");

  newQuestion(version);

  cy.findByText("Orders").click();
})
```

Other smaller changes were also made that created a more universal sequence of UI actions.

Fixes #39229


e2e/test/scenarios/cross-version/source/02-datamodel.cy.spec.js
```diff
@@ -21,7 +21,13 @@ it("should configure data model settings", () => {
   );
 
   cy.get(".AdminList").findByText("Orders").click();
-  cy.findByDisplayValue("Product ID").parent().find(".Icon-gear").click();
+
+  cy.findByDisplayValue("Product ID")
```
markbastian, line 25: This is a disgusting change required since newer versions of metabase did some extra <div> wrapping or something. I verified that it works on older versions.

.github/workflows/e2e-cross-version.yml
```diff
@@ -1,8 +1,8 @@
 name: E2E Cross-version Tests
 
-on:
-  schedule:
-    - cron: '0 9 * * 0'
+on: [push]
+#  schedule:
+#    - cron: '0 9 * * 0'
```
nemanjaglumac, line 5: We can make this on demand with `workflow_dispatch`.

nemanjaglumac: Just for more clarity, I mean this as for after this PR. It's perfectly valid to have it run on every push while you're still iterating on it.

e2e/test/scenarios/cross-version/source/03-questions.cy.spec.js
```diff
@@ -86,8 +110,14 @@ it("should create questions", () => {
   visualize();
   cy.get("circle");
 
-  cy.findByTestId("viz-settings-button").click();
-  cy.icon("area").click();
+  cy.findByTestId("viz-type-button").click();
+  cy.findByTestId("Area-button").click();
+
+  // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
```
iethree, line 116: the purpose of this rule is that we shouldn't  be writing unscoped text selectors. It makes the tests hard to read and potentially introduces flakes and ambiguities.

These tests might be a special case, where we can't add scoping elements for older versions though 😢

e2e/test/scenarios/cross-version/helpers/cross-version-helpers.js
```diff
@@ -0,0 +1,175 @@
...
+    version: versionString,
+    edition: edition,
+    majorVersion: segmentInts[1],
+    minorVersion: segmentInts[2],
+    patchVersion: segmentInts[3],
+  };
+}
+
+// Versions in which a breaking GUI change was introduced. For example, version
+// 32 added a "What will you use Metabase for?" stage in the initial setup.
```
iethree, line 31: 🥴  this is rough, but I can't immediately think of a better way to handle it.

iethree (approved):
👏  this is a great start!

.github/workflows/e2e-cross-version.yml
markbastian: It's back to the original version.

PR #39287, opened 2024-02-29
Split setup process into atomic stages

### LINKED ISSUE #32768: Split Setup process into two atomic stages
**Context**
The unauthenticated use of `/api/setup/validate/` was a bit of a problem recently. In theory the setup token is a credential, but given that recent history shows we might not always control it being leaked, we should add another layer of protection.

I think we should split the setup process into two api calls -
/api/setup/create_user, which creates the first admin user and the remainder into a separate api endpoint. `/api/setup/validate` can then require an admin account to access just like in the normal app.

Downside - we'd need to figure out what to do with the setup equivalent of an abandoned cart, where a user is created but a database never added.

sloansparger:
Closes #32768

### Description

Updates the setup process to call POST /api/setup after the language and user details step is completed then proceeds to make every step after completely atomic. This helps reduce the time the instance spends using the setup token as an auth strategy and tightens security. Consequently, we can no longer need to rely on some of the other setup endpoints as we already have endpoints for fully authenticated users (which the user becomes after they submit their details).

### Product callouts

- There's been some UX changes required to make this feature more modular / easier to ship. Since the main setup endpoint can only be called once, we can't allow users
to go back to the first two steps. 

- The other downside to this new approach is we have no "abandoned cart" functionality if the user doesn't complete the setup after the user details step. They will have provided as much information as needed to initialize their instance, so all further data will have to be collected from within the application.

### Summary of the list of updates make to FE/BE:

- make first two steps of setup flow an atomic step
- reduces setup options accepted by POST /api/setup
- removes POST /api/setup/validate endpoint to validate a db connection
- updates POST /api/database/validate endpoint to result with same response schema as POST /api/setup/validate did (preserves existing data as well)
- DB creation happens via POST /api/database
- FE relies on settings endpoints for things like allowing/disallowing tracking, saving license token, etc.
- removes POST /api/setup/token-check as we can rely on a 400 response to detect validity

### Demo

https://github.com/metabase/metabase/assets/7104357/5bbb1963-e7b4-4daa-80df-5928a7839e16

### How to verify
- Setup a new instance and test that setup flow is functionally identical to existing behaviors (except the aforementioned no going back)

### Checklist


sloansparger:
Remaining backend work
- remove unused fields in setup endpoint (you can see the used fields [here](https://github.com/metabase/metabase/pull/39287/files#diff-f0f1344935a4c128194840b010588c50a1260aa9dae823a4b8c51dee7e130284R104-R112))
- remove old setup db validation endpoint (`POST /api/setup/validate` - src/metabase/api/setup.clj:190)
- move `POST /api/setup/token-check` to something like `/api/token-check` and replace `setup/has-user-setup` clause with an admin check (src/metabase/api/setup.clj:355) - can we make sure that the response body returns the same validation data that `POST /api/setup` is doing now? 
- double check that you can't validate a database when unauthenticated (old endpoint was giving me 4XX status but still with a validation check body)

e2e/support/helpers/e2e-snowplow-helpers.js
```diff
@@ -91,7 +91,7 @@ export const expectGoodSnowplowEvents = count => {
 };
 
 export const expectNoBadSnowplowEvents = () => {
-  sendSnowplowRequest("micro/bad").its("body").should("have.length", 0);
+  sendSnowplowRequest("micro/bad").its("body").should("deep.equal", []);
```
sloansparger, line 94: make it easier to debug failures

frontend/src/metabase/selectors/app.ts
```diff
@@ -20,6 +20,7 @@ export interface RouterProps {
 }
 
 const PATHS_WITHOUT_NAVBAR = [
+  /^\/setup/,
```
sloansparger, line 23: navbar and sidebar would appear when the setup endpoint was called, but the user was not finished with the setup process

e2e/test/scenarios/native/reproductions/30680-native-editor-buttons.cy.spec.js
```diff
@@ -13,7 +13,7 @@ describe("issue 30680", () => {
       .findByText("Use a native query")
       .click();
 
-    cy.get(".ace_editor").type("select * from orders ");
+    cy.get(".ace_editor").type("select * from orders ", { delay: 100 });
```
sloansparger, line 16: unrelated to my changes but this was failing consistently on my branch for some reason 🤷‍♂️

e2e/test/scenarios/question/reproductions/17514-ui-overlay.cy.spec.js
```diff
@@ -149,10 +150,15 @@ describe("issue 17514", () => {
...
+
+      // wait until view results are done rendering
       cy.wait("@dataset");
+      cy.findByTestId("query-builder-main").within(() => {
+        cy.findByText("Doing science...").should("not.exist");
+      });
 
       // Cypress cannot click elements that are blocked by an overlay so this will immediately fail if the issue is not fixed
-      cy.findByTextEnsureVisible("Subtotal").click();
+      openColumnOptions("Subtotal");
```
sloansparger, line 161: also unrelated to my changes and also consistently failing for some reason 🤷‍♂️

frontend/src/metabase/setup/components/LicenseTokenStep/LicenseTokenStep.tsx
```diff
@@ -14,20 +14,22 @@ import type { NumberedStepProps } from "../types";
...
-    );
-    dispatch(submitLicenseToken(token));
+  const handleSubmit = async (token: string | null) => {
+    try {
+      await dispatch(submitLicenseToken(token)).unwrap();
+      dispatch(addUndo({ message: t`Your license is activated` }));
+    } catch (err) {
+      console.error(err);
+      throw new Error(
+        t`This token doesn’t seem to be valid. Double-check it, then contact support if you think it should be working.`,
```
sloansparger, line 30: There will be follow up work to make this submission provide better error messaging based on the work done by: https://github.com/metabase/metabase/pull/39067/

npfitz (approved):
Gave it a test, and on the surface I could hardly tell things have changed haha. Good work!

johnswanson (approved):
Backend changes look good to me!

iethree (approved):

frontend/src/metabase/setup/actions.ts
```diff
@@ -93,9 +92,28 @@ export const updateLocale = createAsyncThunk(
...
+        prefs: {
+          site_name: user.site_name,
+          site_locale: locale?.code,
+        },
+      });
+    } catch (error) {
+      return rejectWithValue(error);
+    }
+
+    MetabaseSettings.set("setup-token", null);
```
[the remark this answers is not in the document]
iethree: (seeing this was just moved from another place) 👍

frontend/src/metabase/setup/tests/enterprise.unit.spec.tsx
```diff
@@ -36,6 +36,10 @@ const setupEnterprise = (opts?: SetupOpts) => {
 const sampleToken = "a".repeat(64);
 
 describe("setup (EE, no token)", () => {
+  beforeEach(() => {
+    fetchMock.reset();
+  });
```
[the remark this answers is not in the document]
sloansparger: changing the order of the tests was causing a previously existing/passing test to fail. this was due to the new test mocking out fetch differently. ideally each test should be isolated, so i feel like we might should be doing this more places...

PR #39314, opened 2024-02-29
update e2e_specs to include ts files

npretto:
### Description

While testing something I noticed we don't consider ts files in the e2e_specs


nemanjaglumac (approved):
👏

PR #39321, opened 2024-02-29
switch back to ring-jetty-adapter - addresses security issue

sansolovyov:
Since jetty is still a security issue, and ring-jetty9-adapter actually pulls http2-server - which we don't use and don't need in our deps.


escherize (approved):
LGTM

piranha:

PR #39324, opened 2024-02-29
Global to local css migration - part of global styles

deniskaber:

### Description

This migrates global CSS rules to be locally scoped.

### How to verify

### DIFF
--- a/frontend/src/metabase/query_builder/components/QueryVisualization.jsx
+++ b/frontend/src/metabase/query_builder/components/QueryVisualization.jsx
@@ -5,6 +5,7 @@ import { useTimeout } from "react-use";
 import { t } from "ttag";
 
 import LoadingSpinner from "metabase/components/LoadingSpinner";
+import CS from "metabase/css/core/index.css";
 import { useSelector } from "metabase/lib/redux";
 import { getWhiteLabeledLoadingMessage } from "metabase/selectors/whitelabel";
 import { HARD_ROW_LIMIT } from "metabase-lib/v1/queries/utils";
@@ -31,17 +32,21 @@ export default function QueryVisualization(props) {
   const [warnings, setWarnings] = useState([]);
 
   return (
-    <div className={cx(className, "relative stacking-context full-height")}>
-      {isRunning ? <VisualizationRunningState className="spread z2" /> : null}
+    <div
+      className={cx(className, CS.relative, CS.stackingContext, CS.fullHeight)}
+    >
+      {isRunning ? (
+        <VisualizationRunningState className={cx(CS.spread, CS.z2)} />
+      ) : null}
       <VisualizationDirtyState
         {...props}
         hidden={!isResultDirty || isRunning || isNativeEditorOpen}
-        className="spread z2"
+        className={cx(CS.spread, CS.z2)}
       />
       {!isObjectDetail && (
         <Warnings
           warnings={warnings}
-          className="absolute top right mt2 mr2 z2"
+          className={cx(CS.absolute, CS.top, CS.right, CS.mt2, CS.mr2, CS.z2)}
           size={18}
         />
       )}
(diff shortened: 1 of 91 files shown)


frontend/src/metabase/query_builder/components/QueryVisualization.jsx
```diff
@@ -31,28 +32,32 @@ export default function QueryVisualization(props) {
...
         <Warnings
           warnings={warnings}
-          className="absolute top right mt2 mr2 z2"
+          className={cx(CS.absolute, CS.top, CS.right, CS.mt2, CS.mr2, CS.z2)}
           size={18}
         />
       )}
       <div
-        className={cx("spread Visualization z1", {
+        className={cx(CS.spread, "Visualization z1", {
```
WiNloSt, line 54: technically you could use `CS.z1` here too right?

frontend/src/metabase/query_builder/components/QueryVisualization.jsx
```diff
@@ -96,9 +101,7 @@ export function VisualizationRunningState({ className = "" }) {
       )}
     >
       <LoadingSpinner />
-      <h2 className="Loading-message text-brand text-uppercase my3">
-        {message}
-      </h2>
+      <h2 className="text-brand text-uppercase my3">{message}</h2>
```
WiNloSt, line 104: 2 questions.
1. Is `Loading-message` unused?
2. Should you convert the rest 3 global classes to CSS modules (CS.xxx)?

frontend/src/metabase/reference/components/EditHeader.jsx
```diff
@@ -15,7 +16,7 @@ const EditHeader = ({
   onSubmit,
   revisionMessageFormField,
 }) => (
-  <div className={cx("EditHeader wrapper py1 px3", S.editHeader)}>
+  <div className={cx("EditHeader", CS.wrapper, "py1 px3", S.editHeader)}>
```
WiNloSt, line 19: 2 things here.
1. `EditHeader` doesn't seem to be used anywhere. Could you double check and possibly remove it if that's the case?
2. Should you use `CS.py1` and `CS.px3` too?

frontend/src/metabase/visualizations/components/LegendItem.jsx
```diff
@@ -50,10 +51,13 @@ export default class LegendItem extends Component {
         className={cx(
           className,
           "LegendItem",
-          "no-decoration flex align-center fullscreen-normal-text fullscreen-night-text",
+          "fullscreen-normal-text fullscreen-night-text",
+          CS.noDecoration,
```
WiNloSt, line 55: super nit: I'd just preserve the order for ease of comparison and when we're tracking the history.

I'd put `CS.noDecoration` and the other 2 classes above `"fullscreen-normal-text fullscreen-night-text"`

frontend/src/metabase/visualizations/visualizations/Scalar/Scalar.jsx
```diff
@@ -218,7 +220,18 @@ export class Scalar extends Component {
 
     return (
       <ScalarWrapper>
-        <div className="Card-title absolute top right p1 px2">
+        <div
+          className={cx(
+            "Card-title",
+            CS.textDefault,
+            "text-smaller",
```
WiNloSt, line 227: where are these 2 classes from? I couldn't see in from the previous version.

WiNloSt (approved):
Overall this is great. I have only minor feedback.

frontend/src/metabase/visualizations/visualizations/Scalar/Scalar.jsx
WiNloSt: That makes sense. Thanks!

Issue #35126, opened 2023-10-26
Browser back button not working after applying click behavior filter


Hritik14:
### Describe the bug

After applying click behavior that updates a dashboard filter, the URL changes with clicks but browser back button does not go back to previous filter.

### To Reproduce

1. Create a SQL question with one variable
2. Create a dashboard and insert this question
3. Create a filter to filter by an ID in this question
4. Configure click behavior to update the filter when user clicks on one of the id columns
5. Click on above configured id columns
6. Hit browser back button

https://github.com/metabase/metabase/assets/7457065/6183819d-b472-4932-96cf-705f590b9187

### Expected behavior

As the URL keeps on updating, back button should push these states to browser history so that back button works correctly

### Information about your Metabase installation

```JSON
- Your browser and the version: Firefox 119.0 (64-bit)
- Your operating system: MacOS Ventura 13.4 (22F66)
- Your databases: Postgres
- Metabase version: v0.47.6
- Metabase hosting environment: jar file
- Metabase internal database: h2


{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/119.0",
    "vendor": ""
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.name": "OpenJDK Runtime Environment",
    "java.runtime.version": "11.0.20.1+0",
    "java.vendor": "Homebrew",
    "java.vendor.url": "https://github.com/Homebrew/homebrew-core/issues",
    "java.version": "11.0.20.1",
    "java.vm.name": "OpenJDK 64-Bit Server VM",
    "java.vm.version": "11.0.20.1+0",
    "os.name": "Mac OS X",
    "os.version": "13.4",
    "user.language": "en",
    "user.timezone": "Asia/Kolkata"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
    "application-database-details": {
      "database": {
        "name": "H2",
        "version": "2.1.214 (2022-06-13)"
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
        "version": "2.1.214 (2022-06-13)"
      }
    },
    "run-mode": "prod",
    "version": {
      "date": "2023-10-26",
      "tag": "v0.47.6",
      "branch": "?",
[...]

PR #24762, opened 2022-08-15
Fix click behavior for links

ranquild:
Fixes https://github.com/metabase/metabase/issues/24756
Reproduces https://github.com/metabase/metabase/pull/24762

How to test:
- Set `IFRAMED = true` here https://github.com/metabase/metabase/blob/4ac2ef97082c0a34c71aef7f8be550910ab828bc/frontend/src/metabase/lib/dom.js#L13
- Open Metabase with a full-app embedding param, e.g. `http://localhost:3000?top_nav=true`
- Open a dashboard with a card 
- Add a click behavior with a link that navigates to a place within the app to a card
- Click on the card
- Make sure the navigation happens and the top nav bar is preserved

PR #39340, opened 2024-02-29
Fix click behavior routing to a different dashboard with a non-default target tab

### LINKED ISSUE #39320: Click behavior causes race conditions when routing to a non-default tab on a different dashboard
### Describe the bug

Click behavior causes race conditions when routing to a non-default tab on a different dashboard. When clicking on the card with click behavior configured, user may be routed to:
1. the default tab on the target dashboard (incorrect)
2. the target tab on the target dashboard (correct)

It is non-deterministic.

### Reproduction

https://github.com/metabase/metabase/assets/22608765/c3fe6a4b-c846-4ca1-89ec-15044d8745f0

### Expected behavior

For the user to be sent to the target tab on the target dashboard

```
master
```

JesseSDevaney:
- Closes: https://github.com/metabase/metabase/issues/39320

---
### How to Manually Verify
1. Create two dashboards
  - (1) Add a card configured with click behavior routing to the second dashboard with the 2nd tab being the target tab
    image
  - (2) Create a two tabbed dashboard with a filter applied to a card on each tab
    image
    
2. Test click behavior on the first dashboard and verify that it routes you to the correct dashboard and correct tab
  image
  image
4. Test multiple times: try refreshing on the first dashboard and testing it again. Verify that it routes you to the correct tab on every attempt.


frontend/src/metabase/visualizations/lib/action.js
```diff
@@ -20,7 +19,6 @@ export function performAction(action, { dispatch, onChangeCardAndRun }) {
       open(url, {
         openInSameOrigin: location => {
           dispatch(push(location));
-          dispatch(setParameterValuesFromQueryParams(location.query));
         },
```
kulyk, line 22: Doesn't it conflict with #24762? (that PR added this line)
Not sure how its related to tabs

JesseSDevaney: Yes it does, but only when routing to questions. I did not notice this until looking at the failing tests and when working on another issue. Removing that line does fix the race condition of pushing a new dashboard + tab location when performing a click action. Somehow it is competing for setting the URL when this occurs. 

I am investigating alternative fixes because it solves one bug but causes another.

JesseSDevaney:
The two-way sync that would solve:
- https://github.com/metabase/metabase/issues/35126
- https://github.com/metabase/metabase/issues/39320
appears to be at odds with a lot of different behaviors (buggy behaviors).

Unsure if we want to continue fixing this in this moment due to the effort and complexity. I pushed up some changes to check to see how many tests were fixed. If there are still a lot failing, I will update the issue with my findings and come back around to this after tackling easier issues.

frontend/src/metabase/query_builder/utils.ts
JesseSDevaney, line 1: If this PR ends up being closed, we should still open up a separate PR for this fix in particular

PR #39421, opened 2024-02-29
Move refresh test helper into test namespace

crisptrutski:
### Description

Actions a trivial TODO - we only use this function from a single test, so move it to the test namespace.


tsmacdonald (approved):
:boat:

PR #39422, opened 2024-02-29
Update various TODO comments

crisptrutski:
### Description

- Remove a 2019 TODO with indecision about a date format - the ship has sailed.
- Remove an obsolete TODO around a schema that was fixed when moving to Malli.
- Link a TODO back to its corresponding issue - it's actually a bug, not just tech debt.


tsmacdonald (approved):
:boat:

PR #39423, opened 2024-02-29
Enforce new permissions

johnswanson:


e2e/snapshot-creators/default.cy.snap.js
```diff
@@ -99,26 +129,28 @@ describe("snapshots", () => {
...
 
-  function addUsersAndGroups() {
+  function addUsersAndGroups(isEE = false) {
+    const lowest_read_data_permission = isEE ? "blocked" : "unrestricted";
+
     // groups
     cy.request("POST", "/api/permissions/group", { name: "collection" }).then(
       ({ body }) => {
-        expect(body.id).to.eq(COLLECTION_GROUP); // 4
+        expect(body.id).to.eq(COLLECTION_GROUP); // 3
```
johnswanson, line 138: Just correcting these comments - e.g. `COLLECTION_GROUP` is set to `3` in `cypress_data.js`.

e2e/snapshot-creators/default.cy.snap.js
```diff
@@ -132,19 +164,40 @@ describe("snapshots", () => {
 
     cy.updatePermissionsGraph({
       [ALL_USERS_GROUP]: {
-        [SAMPLE_DB_ID]: { data: { schemas: "none", native: "none" } },
+        [SAMPLE_DB_ID]: {
+          // set the data permission so the UI doesn't warn us that "all users has higher permissions than X"
```
johnswanson, line 168: The UI is still based on data permissions (for now!) so I'm setting all those just as before.

e2e/test/scenarios/permissions/impersonated.cy.spec.js
```diff
@@ -29,7 +29,8 @@ describeEE("impersonated permission", () => {
       setTokenFeatures("all");
     });
 
-    it("can set impersonated permissions", () => {
+    // skipping for now as it's testing UI. We should update this when the UI sets split data perms
```
noahmoss, line 32: @sloansparger Just want to point out these spots to you. Maybe you can unskip them in your PR?

enterprise/backend/src/metabase_enterprise/sandbox/query_processor/middleware/row_level_restrictions.clj
```diff
@@ -257,11 +257,38 @@
     (qp.store/cached card-id
                      (query-perms/required-perms (:dataset-query (lib.metadata.protocols/card (qp.store/metadata-provider) card-id))
                                                  :throw-exceptions? true))
-    {:perms/data-access (zipmap (sandbox->table-ids sandbox) (repeat :unrestricted))}))
+    {:perms/view-data :unrestricted
+     :perms/create-queries (zipmap (sandbox->table-ids sandbox) (repeat :query-builder))}))
+
+(defn- merge-perms
```
noahmoss, line 263: Nice doc string, this makes sense to me. I guess previously the merge behavior didn't matter since `:data-access` would only be at the table level and `:native-query-editing` only at the DB-level for sandboxes. But now `:create-queries` could be either, depending on whether the sandbox involves a card?

src/metabase/models/data_permissions.clj
```diff
@@ -234,11 +244,37 @@
...
+  (at-least-as-permissive? perm-type
+                           (database-permission-for-group group-id perm-type database-id)
+                           perm-value))
+
 (mu/defn table-permission-for-group :- PermissionValue
   "Returns the effective permission value for a given *group*, permission type, and database ID, and table ID."
   [group-id perm-type database-id table-id]
   (when (not= :model/Table (model-by-perm-type perm-type))
-    (throw (ex-info (tru "Permission type {0} is a table-level permission." perm-type)
+    (throw (ex-info (tru "Permission type {0} is not a table-level permission." perm-type)
```
noahmoss, line 277: 👍 nice catch

src/metabase/models/data_permissions.clj
```diff
@@ -651,15 +682,21 @@
...
                                                                      [:= :db_id db-id]
                                                                      [:not [:in :id table-ids]]]})
                     other-new-perms (map (fn [table]
                                            {:perm_type   perm-type
                                             :group_id    group-id
-                                            :perm_value  (if (= :block existing-db-perm-value)
-                                                           :no-self-service
+                                            :perm_value  (case existing-db-perm-value
+                                                           ;; this will be removed when `:perms/data-access` goes away
+                                                           :block                    :no-self-service
```
noahmoss, line 693: can you remove this now?

src/metabase/models/query/permissions.clj
```diff
@@ -124,7 +124,7 @@
                                    query)}
                        e)]
         (if throw-exceptions? (throw e) (log/error e)))
-      {:perms/data-access {0 :unrestricted}}))) ; table 0 will never exist
+      {:perms/view-data {0 :unrestricted}}))) ; table 0 will never exist
```
noahmoss, line 127: we'll probably change `view-data` to be DB-level once we can drop `legacy-no-self-service`. so it might be better to use `:create-queries {0 :query-builder}` here since that'll stay table-level.

the main important thing is that this it's an impossible permission

noahmoss (approved):
👏 awesome

enterprise/backend/src/metabase_enterprise/audit_app/permissions.clj
```diff
@@ -64,11 +64,15 @@
...
+                                  :none  :no-self-service
+                                  :write (throw (ex-info (tru (str "Unable to make audit collections writable."))
+                                                         {:status-code 400})))
+              create-queries-value (case type
+                                     :read :query-builder
+                                     :none :no)
               view-tables         (t2/select :model/Table :db_id perms/audit-db-id :name [:in audit-db-view-names])]
           (doseq [table view-tables]
-            (data-perms/set-table-permission! group-id (u/the-id table) :perms/data-access permission-value))))))
+            (data-perms/set-table-permission! group-id table :perms/data-access data-access-value)
```
johnswanson, line 77: TODO: I just noticed I'm not setting `:perms/read-data` here, and I think I should be?

noahmoss: Hmm, actually, I don't think so. The previous behavior would only ever create `:unrestricted` or `:no-self-service` `:data-access` which are both mapped to `:unrestricted` `:view-data`.

i.e. we're only controlling audit DB access via query perms. nobody should have `:blocked` data perms, particularly since we can't guarantee that an EE instance with the audit feature flag also has advanced permissions enabled.

johnswanson: Oh right, thanks. I think I thought about this when I implemented it and then forgot about it when I was looking back over things.

Issue #39279, opened 2024-02-28
[Prototype] Implement server side PDF rendering for dashboard subscriptions using chromium


alxnddr:
**Context**

While having fully functional Metabase as single jar is great, however, it limits our ability to provide certain functionality such as PDF email subscriptions that would be relatively easy to implement if we could spin up Chromium as rendering engine. We could still have an easy enough setup by making this feature optional only for users who are willing to set up Chromium, and to make it simple we can ship a Docker image that includes both Metabase instance and Chromium.

This epic describes the work needed to create a proof of concept we could iterate on further.

**Plan** 

### Step 1: JAR <> Browser communication

The goal of this milestone is to add the ability by Metabase instance to connect to a locally running headless Chromium, open any URL in the local Chromium, and capture the page as PDF or PNG.

[Chrome DevTools protocol](https://chromedevtools.github.io/devtools-protocol/) seems like a viable option for that. Selenium WebDriver may also work but I'm not sure if it lets to use an already running Chromium instead of spinning up a new instance.

```[tasklist]
### Checklist
- [ ] BE: Metabase connects to local browser via `MB_CHROMIUM_URL`
- [ ] BE: Metabase can render its a login page in the local headless Chromium and retrieve a PDF
```

### Step 2: Render an actual dashboard as PDF and attach it to a subscription

At this point when we open Metabase in the headless browser it is not logged in and don't have the access to anything. I see two options how we can render an actual dashboard data for the PDF creation.

**a. Authenticate the browser tab**
We can authenticate a tab by setting cookies of a session of a fake user with the [same level](https://www.metabase.com/docs/latest/permissions/notifications) of permissions as the subscription creator. This will give the headless browser tab access to dashboard data but we the set up does not sound easy for the BE and we will explicitly have to ensure there is tabs cookies are sandboxed to avoid any data leakages. 

**b. Provide the right dashboard data to the browser**
With this approach, FE will have a dedicated `SubscriptionsDashboad` page which won't fetch any data but will expect it to be present for example in `window.dashboard` property before rendering dashboard. Compared to opening a page as an authenticated user, this will not require to deal with authentication and with the detection when cards data is loaded so we can capture a PDF.

For me the **option B** seems more straightforward and secure, however it will require just a bit more frontend to add the dashboard page that renders existing static data.

```[tasklist]
### Checklist
[...]

PR #39433, opened 2024-03-01
basic pdf dashboard subscriptions FE

alxnddr:
Epic: https://github.com/metabase/metabase/issues/39279

### Description

Adds a page to be used for the backend to generate PDF dashboard subscriptions.

To do so, using Chrome DevTools Protocol or any other tool:
- prepare a dashboard data in the following format https://github.com/metabase/metabase/pull/39433/files#diff-886ddd0262eff9162d1b97de078de831d853d514fc4acc041859769f2e040e3d and serialize it as JSON
- open `http://*instance_url*/subscriptions_dashboard`
- evaluate a script `window.render("*your_JSON*")` on the page
- capture the page as PDF


.clj-kondo/config.edn
```diff
@@ -44,6 +44,10 @@
     (clojure.core.logic/run)
     (clojure.test/is [query= =? malli=])
     (clojure.tools.macro/macrolet)
+    (etaoin.api/with-chrome)
```
camsaul, line 47: instead of adding a bunch of things like this in here we should write Kondo config hooks so it can properly lint these forms

PR #39436, opened 2024-03-01
Refactor: move fingerprinting and classifier code from sync directory

calherries:
This PR moves all the code in the metabase/sync/analyze folder that isn't coupled to sync into its own top-level folder.

Now the query processor and sync use the contents of the analyze directory.

This was necessary to create a sync module, otherwise we would have circular dependencies between the query processor and sync code.


sansolovyov (approved):
Looks good!

qnkhuat:
wdyt about modulizing this as we move? 

E.g:
- we can create `metabase.analyze.fingerprint.interface` and import things from its sub-namespace into it.
- then `metabase.sync.analyze.fingerprint` should only require `metabase.analyze.fingerprint.interface` for its usage

calherries:
> wdyt about modulizing this as we move?

Yep, I've started that. I just wanted to land this refactor before modularizing further, creating a module is more controversial and involves many more decisions (how to create the linter etc).

src/metabase/sync/analyze/classify.clj
```diff
@@ -70,34 +67,6 @@
...
-  `:sync.classify/original`."
-  [field       :- i/FieldInstance
-   fingerprint :- [:maybe i/Fingerprint]]
-  (reduce (fn [field classifier]
-            (or (sync-util/with-error-handling (format "Error running classifier on %s"
-                                                       (sync-util/name-for-logging field))
-                  (classifier field fingerprint))
-                field))
-          (vary-meta field assoc :sync.classify/original field)
-          classifiers))
```
calherries, line 98: This is now in `src/metabase/analyze/classifiers.clj`

src/metabase/analyze/schema.clj
```diff
@@ -0,0 +1,33 @@
...
+     (every? (fn [k]
+               (not (str/includes? k "-")))
+             (keys m)))])
+
+(mr/def ::Table
+  [:and
+   (ms/InstanceOf :model/Table)
+   ::no-kebab-case-keys])
+
+;; TODO: fix memory issues with `mu/defn` and `ms/InstanceOf` so we don't need to do this
```
calherries, line 21: I'll address this TODO in another PR.

src/metabase/analyze/fingerprint.clj
qnkhuat, line 1: `metabase.analyze.fingerprint.schema` ?

src/metabase/analyze/classifiers.clj
qnkhuat, line 1: what do you think about having this namespace closer to its sub ns?

something like `metabase.analyze.classifiers.core`.

This will make browsing the code base easier.

FWIW mlv2 have `metabase.lib.core`

src/metabase/analyze/fingerprint.clj
calherries: yeah that's tidier thanks

qnkhuat (approved):
Much better. Thanks!

PR #39440, opened 2024-03-01
hide error when the user changes the token

npretto:
[[Epic] Add license activation in the initial setup](https://github.com/metabase/metabase/issues/38867)

### Description

The error was persisted when the user changed the token, this addresses that.
[Slack thread where the feedback was given](https://metaboat.slack.com/archives/C063Q3F1HPF/p1709244578414609?thread_ts=1709224918.522439&cid=C063Q3F1HPF)

### How to verify

- Run ee code without a token as env
- type a non valid token
- press active
- you should see the error
- change the token
- the error should go away

### Checklist


WiNloSt (approved):
Looks good to me.

PR #39441, opened 2024-03-01
Optimize some parts of Notebook editor

uladzimirdev:
### Description

This PR addresses several perf problems, hovewer other problems seems cannot be addressed because of a nature of cljs and data immutability together with probably missing methods and some perf problems in cljs code.

### Problems we addressed
1. `NotebookSteps` component tracks in his state a list of opened steps - steps which user added compared to the saved state or all steps for a new question. There was added a premature optimization to close an opened step on query change, but we didn't track was the step opened and not and with MLv2 we get a new query instance in many cases (e.g. when some specific column is selected/unselected in the join step). So even if no steps were opened, we updated a state of `NotebookSteps` component and triggered re-rendering of all children components down by the tree.

🔧 Check if the step is opened, and do not update state if it's not

2. `TimeSeriesChrome` component is not shown in Notebook mode, but it performs calculations under the hood. 

🔧 Do not render `TimeSeriesChrome` component in Notebook mode

3. `ViewTitleHeaderRightSide` components triggers `question.canRun()` method, which is extremely expensive for some reason, twice for the same question.

🔧 trigger `question.canRun()` once and reuse the result

### What can't be updated, but it would help

- `JoinStep` is re-rendered because of results of `Lib.joinStrategy(join)` and `Lib.joinConditions(join)`. As query reference is new every time, joinStrategy and joinConditions return a new reference every time causing re-rendering of the component
- `JoinCondition` is re-rendered because of JoinStep and `Lib.joinConditionParts`, which returns new references for `lhsColumn` and `rhsColumn`
- `question.canRun` should be faster (maybe cached)

### How to verify

tests should pass

play with QB with multiple steps (especially several join steps)

### Demo

All tests were done on the same question and with CPU throttling emulation x6, performing one click on the dropdown

#### Before

image

image

#### After

image

image

### DIFF
--- a/frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
+++ b/frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
@@ -282,7 +282,7 @@ function JoinCondition({
     setOperator,
     setLHSColumn,
     setRHSColumn,
-  } = useJoinCondition(query, stageIndex, table, join, condition);
+  } = useJoinCondition(query, stageIndex, condition);
 
   const getLhsColumnGroup = () => {
     const lhsColumns = Lib.joinConditionLHSColumns(
--- a/frontend/src/metabase/query_builder/components/view/View.jsx
+++ b/frontend/src/metabase/query_builder/components/view/View.jsx
@@ -265,8 +265,14 @@ class View extends Component {
   };
 
   renderMain = ({ leftSidebar, rightSidebar }) => {
-    const { question, mode, parameters, isLiveResizable, setParameterValue } =
-      this.props;
+    const {
+      question,
+      mode,
+      parameters,
+      isLiveResizable,
+      setParameterValue,
+      queryBuilderMode,
+    } = this.props;
 
     const queryMode = mode && mode.queryMode();
     const { isNative } = Lib.queryDisplayInfo(question.query());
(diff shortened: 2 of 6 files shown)


frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
```diff
@@ -282,7 +282,7 @@ function JoinCondition({
     setOperator,
     setLHSColumn,
     setRHSColumn,
-  } = useJoinCondition(query, stageIndex, table, join, condition);
+  } = useJoinCondition(query, stageIndex, condition);
```
ranquild, line 285: There should be useMemo here

frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -295,7 +301,13 @@ class View extends Component {
             mode={queryMode}
           />
         </StyledDebouncedFrame>
-        <TimeseriesChrome {...this.props} className="flex-no-shrink" />
+        {queryBuilderMode === "view" && (
```
ranquild, line 304: Should we call `renderMain` only when it's "view"?

ranquild (approved):

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
uladzimirdev: could you please explain the idea?

frontend/src/metabase/query_builder/components/view/View.jsx
uladzimirdev: at least tests passed [10e689a](https://github.com/metabase/metabase/pull/39441/commits/10e689a2a1c20dc5fc294a9f57072e89b34a3bbd)

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.tsx
uladzimirdev: if I get the idea, I'll create a follow-up PR

PR #39452, opened 2024-03-01
docs - add refresh parameter

jeff-bruemmer:
Adds `refresh` to list of parameters for static embeds.


albertoperdomo (changes requested):
I think we should add it here, too: https://www.metabase.com/docs/latest/embedding/static-embedding#customizing-the-appearance-of-static-embeds

Also, I am wondering if we could not have a short section about it, similar to [this](https://www.metabase.com/docs/latest/embedding/static-embedding#resizing-dashboards-to-fit-their-content)

albertoperdomo (approved):

jeff-bruemmer:

Issue #39211, opened 2024-02-27
[Epic] Better UI/UX for picking default parameter values


oleggromov:
**Links**
Product doc: ❤️ [Better UI/UX for picking default parameter values](https://www.notion.so/metabase/Better-UI-UX-for-picking-default-parameter-values-6ce0b45af9604f2e90330b44c457311a)
Previous project: ⭐️ [Let creators make dashboard filters required](https://www.notion.so/metabase/Let-creators-make-dashboard-filters-required-7ea3d7caf63545d2b713d1509e4267ad) 

## Milestone 0
- [ ] **WIP** Identify and catalogue all value picker widget behaviors
- [ ] Figure out possible complications 
- [ ] Put together a project plan 

The next steps will be based upon these findings. 

## Milestone 1 — native query default value pickers
- [x] https://github.com/metabase/metabase/pull/39454 — plain values, dates
- [x] https://github.com/metabase/metabase/pull/39852 — categories with different sources to pick from
- [ ] Field parameters
- [ ] Error states, default toggle/blur behavior, other small additions
- [ ] Add a comprehensive set of tests covering:
  - [ ] new UI behaviors
  - [ ] parameter processing (lib)

## Milestone 2 — dashboard default value pickers
TBD

## Next milestones
TBD:
- native question filter components 
- dashboard filter componets
- embeddings
- remove the entire `ParameterValueWidget` with deps

## Issues to be addressed (incomplete list)

- when input is losing focus, restore [the default value](https://metaboat.slack.com/archives/C02H619CJ8K/p1707957402805569?thread_ts=1707417723.403539&cid=C02H619CJ8K)
  - In a TextWidget, press Enter - and it won't set default value back
-  improvements to the default value picking flow (points [A-C here](https://metaboat.slack.com/archives/C02H619CJ8K/p1707957402805569?thread_ts=1707417723.403539&cid=C02H619CJ8K)), and also [Conor's comment](https://metaboat.slack.com/archives/C02H619CJ8K/p1708040647469919?thread_ts=1707417723.403539&cid=C02H619CJ8K)
  - Same as above but might have additional details: ensure input error states are correct as in the [slack comment](https://metaboat.slack.com/archives/C0645JP1W81/p1706217318961009?thread_ts=1706207086.466069&cid=C0645JP1W81)

PR #39454, opened 2024-03-01
Native query default parameter values: use ParameterValuePicker for single strings, numbers, and dates

oleggromov:
Part of [[Epic] Better UI/UX for picking default parameter values](https://github.com/metabase/metabase/issues/39211)

## What's done
We start from updating Default value picker in Native queries.

This PR does the following:

- 🐞 fixes: when we reset widget type (think `date/relative` -> `date/single`), reset the default value too
- 🐞 fixes ugly year picker lacking gap with the quarter selector in Quarter and Year picker
- disables Filter widget type when it has a single value (ID or None)
- uses Mantine for a single number (when you selected Number) or string (String + Input box)
- uses Mantine for date picker popovers
- stops using moment for generating ISO-style dates 2024-03-06
- tidies up `template-tags` utils

### How is it done
Instead of changing the original `ParameterValueWidget`, I have built a `ParameterValuePicker` component which reuses the former internally. 

```
tag, parameter -> ParameterValuePicker -> can I render it? ---YES--> do it
                                                          \---NO---> defer to ParameterValueWidget
```

With this approach, we can:
- precisely control what is rendered by the new component (it defaults to `false`)
- selectively roll out the change by choosing which components use `ParameterValuePicker`
- in case something breaks, easily revert the changes
- once we're done with Native queries, reuse the component for Dashboard parameters

### Component design
I deliberately made the component sub-tree (until it reaches `Date*` widgets) stateless. This way we don't struggle with the mismatch between the internal component state and what's coming to it from the store.

This was previously controlled by `commitImmediately` but as of this PR, I don't see why we would need internal state, which makes internals more complicated.

## Why it is done
- we still need to transition to Mantine and match UIs 
- there are multiple bugs and inadequacies in the current default value picking flow

## Screencast
Everything except shown in the screencast remains unchanged.

https://github.com/metabase/metabase/assets/2196347/9e9a9ba4-25ea-4310-8a3e-a8ca74ce954b

## Tests
- reset default value when changing widget type: covered by `sql-field-filter-[date|number|string].cy.spec.js`
- disabled widget type select: covered by `sql-field-filter.cy.spec.js`,  `31606.cy.spec.js`, `TagEditorParam.unit.spec.tsx`

### DIFF
--- a/frontend/src/metabase/parameters/components/ParameterValuePicker/OwnDatePicker.tsx
+++ b/frontend/src/metabase/parameters/components/ParameterValuePicker/OwnDatePicker.tsx
@@ -8,7 +8,11 @@ import { DateQuarterYearWidget } from "metabase/components/DateQuarterYearWidget
 import { DateRelativeWidget } from "metabase/components/DateRelativeWidget";
 import { formatParameterValue } from "metabase/parameters/utils/formatting";
 import { Popover } from "metabase/ui";
-import type { Parameter, ParameterType } from "metabase-types/api";
+import type {
+  DateParameterType,
+  Parameter,
+  ParameterType,
+} from "metabase-types/api";
 
 import {
   TextInputIcon,
@@ -23,6 +27,7 @@ export function OwnDatePicker(props: {
   const { value, parameter, onValueChange } = props;
   const [isOpen, setIsOpen] = useState(false);
   const formatted = formatParameterValue(value, parameter);
+  const parameterType = parameter.type as DateParameterType; // TODO fix types in Parameter
 
   const openPopover = () => setIsOpen(true);
   const closePopover = () => setIsOpen(false);
@@ -48,10 +53,10 @@ export function OwnDatePicker(props: {
     : { style: { pointerEvents: "none" } };
 
   // TODO this should be removed as soon as we reconcile all dropdowns and make them use Mantine
-  const Z_INDEX = 2;
+  const zIndex = hasInnerPopovers(parameterType) ? 3 : undefined;
 
   return (
-    <Popover opened={isOpen} zIndex={Z_INDEX}>
+    <Popover opened={isOpen} zIndex={zIndex}>
       <Popover.Target>
         <TextInputTrirgger
           ref={setTriggerRef}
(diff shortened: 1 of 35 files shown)


frontend/src/metabase/parameters/components/ParameterValuePicker/OwnDatePicker.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const { value, parameter, onValueChange } = props;
+  const [isOpen, setIsOpen] = useState(false);
+  const formatted = formatParameterValue(value, parameter);
+  const parameterType = parameter.type as DateParameterType; // TODO fix types in Parameter
+
+  const openPopover = () => setIsOpen(true);
+  const closePopover = () => setIsOpen(false);
+
+  const [triggerRef, setTriggerRef] = useState<HTMLDivElement | null>(null);
+  const dropdownRef = useClickOutside(closePopover, null, [triggerRef]);
```
ranquild, line 36: It should work out of the box, you shouldn't need this. Try `<Popover ... onChange={setIsOpen}>`

frontend/src/metabase/parameters/components/ParameterValuePicker/OwnDatePicker.tsx
```diff
@@ -0,0 +1,153 @@
...
+    if (parameterType === "date/range") {
+      const now = getIsoDate();
+      return `${now}~${now}`;
+    }
+  }
+
+  return value;
+}
+
+function getIsoDate() {
```
ranquild, line 151: Please use `dayjs`. This is our "modern" alternative to moment

frontend/src/metabase/parameters/components/ParameterValuePicker/OwnDatePicker.tsx
```diff
@@ -0,0 +1,153 @@
...
+  Parameter,
+  ParameterType,
+} from "metabase-types/api";
+
+import {
+  TextInputIcon,
+  TextInputTrirgger,
+} from "./ParameterValuePicker.styled";
+
+export function OwnDatePicker(props: {
```
ranquild, line 22: nit: In some newer files (but not all) we put nested components in subfolders (take a look at `metabase/querying`). We had an RFC about this and it was accepted, but I don't think that all people follow this approach

ranquild (approved):
It would be great to do small cleanup but this is a good first step

oleggromov:
Thanks @ranquild, I'll address your suggestions in the next PR and merge this as is as we agreed.

PR #39456, opened 2024-03-01
Differentiate `UserSettings` type from the rest of the `Settings`

nemanjaglumac:
### Description

This PR introduces a new type called `UserSettings` to differentiate it from the rest of the `Settings`.
It also brings type safety to the `updateUserSetting` action.

### How to verify

Try passing a wrong key or a wrong value to the `updateUserSetting` action. TS should complain.

### Demo

Wrong key:
image

Correct key, wrong value
image

### Additional Info
This is a follow-up after #39251 and #39359

### DIFF
--- a/frontend/src/metabase-types/api/mocks/settings.ts
+++ b/frontend/src/metabase-types/api/mocks/settings.ts
@@ -8,7 +8,6 @@ import type {
   Settings,
   TokenFeatures,
   TokenStatus,
-  UserSettings,
   Version,
   VersionInfo,
   VersionInfoRecord,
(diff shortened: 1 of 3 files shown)


iethree (changes requested):
I would do this a bit differently, and create `InstanceSettings` and `UserSettings` and then make a union type of both: `type Settings = InstanceSettings & UserSettings;`

nemanjaglumac:
> I would do this a bit differently, and create `InstanceSettings` and `UserSettings` and then make a union type of both: `type Settings = InstanceSettings & UserSettings;`

That is the end goal of https://github.com/metabase/metabase/issues/39461. This is just a mid-step in that direction, but I can implement the union immediately. It doesn't hurt.

iethree (approved):
❤️

Issue #38778, opened 2024-02-14
"Contains" filter types on dashboards should allow users to pick values in the same way as an "is" filter


paoliniluis:
**Is your feature request related to a problem? Please describe.**
If we can give the contains filter the same behavior as the "is" filter (picklist from pre-set values, get values from a query, etc), we could solve a very specific use case that users been having for some time:

let's say that users have an array of values in a field (comma separated or a joined string altogether):
id, value
1, luis,tony
2, bruno,conor
3, cynthia,sameer

With the contains filter it's easy to set a value like "bruno", but maybe you don't know that the bruno value exists from the beginning (maybe in a dashboard), so if we can pre-set values in a contains type of filter we will solve this use case. Yes, it will allow us to give a workaround for people that want to filter on arrays

**Describe the solution you'd like**
Contains type of filter should allow the same features as the "is" filter (why not the rest of the types as well???)

**Describe alternatives you've considered**
SQL questions, which don't let you use all the magic

**How important is this feature to you?**
Requested by a lead

**Additional context**
NA

PR #39462, opened 2024-03-01
[WIP] Show field values for all string operators for filters in the QB

ranquild:
Related to https://github.com/metabase/metabase/issues/38778

For MBQL query filters based on string columns, show field values for all operators and not just `Is` and `Is not`.

Screenshot 2024-03-05 at 18 50 54


ranquild:
Closing - the UX is weird. We'd try to do it for dashboard parameters only

PR #39464, opened 2024-03-01
Convert `dashboard/actions/core` to TypeScript

kulyk:
Converts the "core" dashboard actions to TypeScript (`setDashboardEditing`, `setDashboardAttributes`, etc.)
Backporting to v49 to prevent merge conflicts later, should be a safe type-only change.

### DIFF
--- a/frontend/src/metabase/dashboard/actions/actions.ts
+++ b/frontend/src/metabase/dashboard/actions/actions.ts
@@ -8,8 +8,6 @@ import { ActionsApi, PublicApi } from "metabase/services";
 import type {
   ActionDashboardCard,
   ActionFormSubmitResult,
-  ActionParametersMapping,
-  CardId,
   Dashboard,
   ParametersForActionExecution,
   WritebackAction,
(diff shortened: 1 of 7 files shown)


frontend/src/metabase/dashboard/actions/actions.ts
```diff
@@ -21,16 +19,12 @@ import { getDashboardType } from "../utils";
...
-  card_id?: CardId | null;
-  action?: WritebackAction | null;
-  parameter_mappings?: ActionParametersMapping[] | null;
-  visualization_settings?: ActionDashboardCard["visualization_settings"];
-}
-
 export function updateButtonActionMapping(
   dashCardId: number,
-  attributes: DashboardAttributes,
+  attributes: Pick<
```
uladzimirdev, line 24: I'd not inline such type, previous version was more readable

uladzimirdev (approved):

frontend/src/metabase/dashboard/actions/actions.ts
kulyk: Updated [here](https://github.com/metabase/metabase/pull/39472/commits/1e7be5ddb9ee20c92e040ff9b73e094855ca15dc) in the next PR in a chain (#39472)

PR #39466, opened 2024-03-01
Additional X-Ray Documentation

markbastian:
This just expands the documentation in the primary x-ray namespace (`metabase.automagic-dashboards.core`). At this point, someone new to the namespace should be able to get into the x-ray code without difficulty.


adam-james-v (approved):
Looks good! Always nice to have more documentation

PR #39468, opened 2024-03-01
Define the `PublicSettings` type

nemanjaglumac:
This PR continues to break down the ambiguous `Settings` type into smaller, well-defined settings based on their nature. We're defining all Public settings here.

> [!Note]
> All these settings should be available for the unauthenticated user.

### How to verify?
1. Run a local Metabase instance.
2. Clear the network tab
3. Sign out
4. Check the `GET /api/session/properties` response

Every single key listed in the `PublicSettings` type should be there.
Keep in mind that not all possible Metabase public settings have been documented/defined on the Frontend in the first place so you shouldn't be surprised if you don't see them in this type.


frontend/src/metabase-types/api/settings.ts
```diff
@@ -285,13 +239,62 @@ interface InstanceSettings {
...
+  "enable-enhancements?": boolean;
+  "enable-password-login": boolean;
+  engines: Record<string, Engine>;
+  "ga-enabled": boolean;
+  "google-auth-client-id": string | null;
+  "google-auth-enabled": boolean;
+  "has-user-setup": boolean;
+  "hide-embed-branding?": boolean;
+  "help-link": HelpLinkSetting;
+  "help-link-custom-destination": string;
```
kamilmielnik, line 264: Alphabetical order: 
```suggestion
  "help-link": HelpLinkSetting;
  "help-link-custom-destination": string;
  "hide-embed-branding?": boolean;
```

frontend/src/metabase-types/api/settings.ts
```diff
@@ -285,13 +239,62 @@ interface InstanceSettings {
...
+  "ldap-enabled": boolean;
+  "loading-message": LoadingMessage;
+  "map-tile-server-url": string;
+  "other-sso-enabled?": boolean | null;
+  "password-complexity": PasswordComplexity;
+  "persisted-models-enabled": boolean;
+  "report-timezone-short": string;
+  "report-timezone-long": string;
+  "setup-token": string | null;
+  "session-cookies": boolean | null;
```
kamilmielnik, line 277: Alphabetical order:
```suggestion
  "session-cookies": boolean | null;
  "setup-token": string | null;
```

kamilmielnik (approved):

frontend/src/metabase-types/api/settings.ts
```diff
@@ -285,13 +239,62 @@ interface InstanceSettings {
...
+  "hide-embed-branding?": boolean;
+  "help-link": HelpLinkSetting;
+  "help-link-custom-destination": string;
+  "is-hosted?": boolean;
+  "is-metabot-enabled": boolean;
+  "ldap-configured?": boolean;
+  "ldap-enabled": boolean;
+  "loading-message": LoadingMessage;
+  "map-tile-server-url": string;
+  "other-sso-enabled?": boolean | null;
```
kamilmielnik, line 271: This attribute is not present in GET `/api/session/properties` response in my local instance.

nemanjaglumac: 🤷‍♂️ 

https://github.com/metabase/metabase/blob/be-user-local-settings/enterprise/backend/src/metabase_enterprise/sso/integrations/sso_settings.clj#L255

nemanjaglumac: This seems to be an Enterprise setting. Probably that's why.
TBH, I don't know how to treat enterprise-only settings that are also public. Should they stay here, or should they be moved to the Enterprise type?

nemanjaglumac: @kamilmielnik moved it to EnterpriseSettings for now.

kamilmielnik: > @kamilmielnik moved it to EnterpriseSettings for now.

Good idea :+1: 

> TBH, I don't know how to treat enterprise-only settings that are also public. Should they stay here, or should they be moved to the Enterprise type?

Or maybe we need `EnterprisePublicSettings`?

nemanjaglumac: I think we'll eventually end up having Settings as a union of AdminSettings | PublicSettings, etc.
And in the same way, we'll have EnterpriseSettings as a union of EnterpriseAdminSettings | EnterprisePublicSettings, etc.

nemanjaglumac: Unfortunately, had to move it back to unblock this PR.
https://github.com/metabase/metabase/pull/39468/commits/26d905a606c2f12c0f6265c8012d3fb953cbb512

The proper solution would be bigger than all changes in this PR so I'll make sure to follow up.
See [this Slack convo](https://metaboat.slack.com/archives/C0645JP1W81/p1709639951515289) for more context.

PR #39469, opened 2024-03-01
Move user-local settings to user model

nemanjaglumac:
This PR simply moves a two user-local settings to the `public_settings` to the `user` model.
The previous location where these settings have been defined is misleading, IMO.


calherries (approved):
Cool, it makes sense to have user-local only settings in the models namespace because these are basically like columns in the user table.

PR #39471, opened 2024-03-01
Add hostname support to snowflake

### LINKED ISSUE #22133: Snowflake alternative hostname implementation
Snowflake supports using an alternative hostname (e.g. not <ACCOUNT>.snowflakecomputing.com). Without setting it, in environments such as with privatelink or custom hostnames (such as proxied), it's not possible to configure Snowflake.

Adding an advanced settings field of "alternative hostname", which if present will override the existing hostname (<account>.snowflakecomputing.com). Note that the account itself needs to be sent with the "account" JDBC parameter.

For us this is specifically important, as we proxy traffic to Snowflake accounts.

vieux:
Closes https://github.com/metabase/metabase/issues/22133

### Description

Add an hostname support to snowflake. When set, it will be used instead of `<account_name>.snowflakecomputing.com`

### How to verify

1. Have a snowflake instance not accessible directly on *.snowflakecomputing.com
2. When adding the Snowflake connection, select `Use hostname` and set an hostname.
3. Validate metabase can connect.

### Demo
Screenshot 2024-03-14 at 12 56 23 AM

`then, when Use hostname is clicked`

Screenshot 2024-03-14 at 12 56 30 AM

### Checklist


luizarakaki:
@vieux Can you make this selection automatic? If you can come up with regex rules or something to use a single field for account/hostname, it will be easier to merge the PR.

Automatic selection is likely how we would implement internally.

luizarakaki:
If automatic selection is really not possible, we would likely implement a toggle close to the account/hostname field

vieux:
@luizarakaki I'm not sure how to have a 100% reliable automatic selection since "account" can use several formats (with and  without `.` for example, if you include the region)

I'll take a stab at implementing a toggle

vieux:
@luizarakaki @qnkhuat I updated the PR, please take another look, thanks!

vieux:
The UI is similar to the MongoDB driver with the connection uri.

luizarakaki:
Thanks @vieux, the UI looks good. I'll ask for a code review.

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -123,15 +123,21 @@
...
   (when (get "week_start" (sql-jdbc.common/additional-options->map additional-options :url))
     (log/warn (trs "You should not set WEEK_START in Snowflake connection options; this might lead to incorrect results. Set the Start of Week Setting instead.")))
   (let [upcase-not-nil (fn [s] (when s (u/upper-case-en s)))]
     ;; it appears to be the case that their JDBC driver ignores `db` -- see my bug report at
     ;; https://support.snowflake.net/s/question/0D50Z00008WTOMCSA5/
     (-> (merge {:classname                                  "net.snowflake.client.jdbc.SnowflakeDriver"
                 :subprotocol                                "snowflake"
-                :subname                                    (str "//" account ".snowflakecomputing.com/")
+                ;; see https://github.com/metabase/metabase/issues/22133
+                :subname                                    (let [base-url (if (and (not use-account-name) (string? host) (not (empty? host)))
```
metamben, line 135:
```suggestion
                :subname                                    (let [base-url (if (and (not use-account-name) (string? host) (not (str/blank? host)))
```

modules/drivers/snowflake/src/metabase/driver/snowflake.clj
```diff
@@ -123,15 +123,21 @@
...
+                                                                              (if-not (= (last host) \/)
+                                                                                (str host "/")
+                                                                                host)
```
metamben, line 138:
```suggestion
                                                                              (cond-> host
                                                                                (not= (last host) \/) (str "/"))
```

modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj
```diff
@@ -87,13 +87,28 @@
...
                                  (sql-jdbc.conn/connection-details->spec :snowflake details)))
         true "\"v3_sample-dataset\""
-        false "v3_sample-dataset"))))
+        false "v3_sample-dataset"))
+    (testing "Subname is replaced if hostname is provided (#22133)"
+      (are [use-account-name alternative-host expected-subname] (=? expected-subname
+                               (:subname (let [details (-> details
+                                                           (assoc :host alternative-host)
+                                                           (assoc :use-account-name use-account-name))]
+                                 (sql-jdbc.conn/connection-details->spec :snowflake details))))
```
metamben, line 103: Please fix the indentation.

modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj
```diff
@@ -87,13 +87,28 @@
...
+        false nil "//ls10467.us-east-2.aws.snowflakecomputing.com/"
+        false "" "//ls10467.us-east-2.aws.snowflakecomputing.com/"
```
metamben, line 105:
```suggestion
        false "" "//ls10467.us-east-2.aws.snowflakecomputing.com/"
        false "  " "//ls10467.us-east-2.aws.snowflakecomputing.com/"
```

modules/drivers/snowflake/resources/metabase-plugin.yaml
```diff
@@ -8,11 +8,20 @@ driver:
   lazy-load: true
   parent: sql-jdbc
   connection-properties:
+    - name: use-account-name
+      type: section
+      default: true
```
metamben, line 13: It would be more natural for me if we flipped the condition, i.e., having `use-host-name` with `default: false`. That would tell me that now one can opt-in to using the host name.

metamben (commented):
I can't really comment about the TS code, I'm assuming it's more or less a copy of `DatabaseConnectionSectionField`.

What's the argument for implementing opt-out instead of opt-in? Opt-in feels more natural for me and I think the code at line 135 of `modules/drivers/snowflake/src/metabase/driver/snowflake.clj` would be ever so slightly more readable.

It looks good to me otherwise, I have only some trivial remarks. 👍  I'll go ahead and run the tests internally.

metamben:
@vieux please see the test result of #40137. The inter errors are real. The issue the BE linter is complaining about has been pointed to already, the FE linter issue is "new".

modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj
vieux: @metamben the linter doesn't input any issue. I'm not too familiar with closure, if there a script to check/fix indentation on the project?

iethree:
frontend code looks good to me 👍

modules/drivers/snowflake/test/metabase/driver/snowflake_test.clj
metamben: Never mind then, we'll fix it.

metamben (approved):

PR #39476, opened 2024-03-01
Fix duplicate native column nesting

### LINKED ISSUE #39098: Duplicate column name error when exploring results of native query with join
### Describe the bug

https://github.com/metabase/metabase/assets/6830683/c35e8123-337c-4b86-be19-2742a76ad758

### To Reproduce

1. Create a new native question
2. Use sample database
3. Use this SQL: `select * from orders o join products p on o.product_id = p.id;`
4. Preview it - it works
5. Save it as a new question
6. Click "Explore results"

POST `/api/dataset` gives the following error:

```
Duplicate column name "ID"; SQL statement:
-- Metabase:: userID: 1 queryType: MBQL queryHash: 56c77cd504578108ac83f1b5fa3f7c8bd4cb0753a50657cb376e26608bb07ead
SELECT "source"."ID" AS "ID", "source"."USER_ID" AS "USER_ID", "source"."PRODUCT_ID" AS "PRODUCT_ID", "source"."SUBTOTAL" AS "SUBTOTAL", "source"."TAX" AS "TAX", "source"."TOTAL" AS "TOTAL", "source"."DISCOUNT" AS "DISCOUNT", "source"."CREATED_AT" AS "CREATED_AT", "source"."QUANTITY" AS "QUANTITY", "source"."ID_2" AS "ID_2", "source"."EAN" AS "EAN", "source"."TITLE" AS "TITLE", "source"."CATEGORY" AS "CATEGORY", "source"."VENDOR" AS "VENDOR", "source"."PRICE" AS "PRICE", "source"."RATING" AS "RATING", "source"."CREATED_AT_2" AS "CREATED_AT_2" FROM (select * from orders o join products p on o.product_id = p.id) AS "source" LIMIT 2000 [42121-214]
```

Payload sent to this endpoint:
```JSON
{
    "database": 1,
    "query": {
        "source-table": "card__3"
    },
    "type": "query",
    "parameters": []
}
```

master, b98bca4
Also reproducible in 0.47.0 & 0.48.0

### Severity

P2

snoe:
Fixes #39098

A native query that produces ambiguous columns `select id, id from orders` cannot be wrapped to reference those columns without disambiguation: `select source.id from (select id, id from orders) as source` fails. In SQL 92 support for table column aliases was added `select source.id from (select id, id from orders) as source (id, id_2)` but many drivers still do not support this including mariadb, sqlite, bigquery

However, `with` statements are widely supported and can rename columns as needed too. I was surprised that nesting with statements was so well supported but it appears that all of our drivers handle it well.


src/metabase/driver/sql/query_processor.clj
```diff
@@ -1409,23 +1411,44 @@
...
+        source-clause (cond
+                        persisted
+                        (sql-source-query persisted nil)
+
+                        native
+                        (sql-source-query native params)
+
+                        :else
+                        (apply-clauses driver {} source-query))
+        ;; TODO: Use MLv2 here to get source and desired-aliases
```
snoe, line 1429: Is there a better way to do this now?

src/metabase/driver/sql/query_processor.clj
```diff
@@ -1409,23 +1411,44 @@
...
+        alias-info (mapv (fn [{[_ desired-ref-name] :field_ref source-name :name}]
+                           [source-name desired-ref-name])
+                         source-metadata)
+        source-aliases (mapv first alias-info)
+        desired-aliases (mapv second alias-info)
+        duplicate-source-aliases? (not (apply distinct? source-aliases))
```
lbrdnk, line 1435: When I run the code with fresh app db, I get an sync error. Further examining what's going on (original exception is not preserved), I discovered the exception is thrown on this line:

```
clojure.lang.ExceptionInfo: Error compiling query: Wrong number of args (0) passed to: clojure.core/distinct? {:query {:database 1, :type :query, :query {:source-table 1, :expressions {"substring172365" [:substring [:field 17 nil] 1 1234], "substring172366" [:substring [:field 18 nil] 1 1234], "substring172367" [:substring [:field 19 nil] 1 1234]}, :fields [[:expression "substring172365"] [:expression "substring172366"] [:expression "substring172367"] [:field 22 {:temporal-unit :default}] [:field 21 nil]], :limit 10000}, :middleware {:format-rows? false, :skip-results-metadata? true}}, :type :driver, :e #error {
 :cause "Wrong number of args (0) passed to: clojure.core/distinct?"
 :via
 [{:type clojure.lang.ArityException
   :message "Wrong number of args (0) passed to: clojure.core/distinct?"
   :at [clojure.lang.AFn throwArity "AFn.java" 429]}]
 :trace
 [[clojure.lang.AFn throwArity "AFn.java" 429]
  [clojure.lang.RestFn invoke "RestFn.java" 399]
  [clojure.lang.AFn applyToHelper "AFn.java" 152]
  [clojure.lang.RestFn applyTo "RestFn.java" 132]
  [clojure.core$apply invokeStatic "core.clj" 667]
  [clojure.core$apply invoke "core.clj" 662]
  [metabase.driver.sql.query_processor$apply_source_query invokeStatic "query_processor.clj" 1435]
  [metabase.driver.sql.query_processor$apply_source_query invoke "query_processor.clj" 1413]
  [metabase.driver.sql.query_processor$apply_clauses invokeStatic "query_processor.clj" 1461]
  ...
```

Problem seems to be, that under specific circumstances we do `(apply distinct? [])` which results in call to `distinct?` with no arguments.

Ids from query to reproduce on (first line of the trace) follows:
```
metabase=# select id, name from metabase_database where id = 1;
 id |      name       
----+-----------------
  1 | Sample Database
(1 row)

metabase=# select id, name from metabase_table where id = 1;                                  
 id |      name       
----+-----------------
  1 | ANALYTIC_EVENTS
(1 row)    

metabase=# select id, name from metabase_field where id > 16 and id < 23;
 id |     name     
----+--------------
 17 | EVENT
 18 | PAGE_URL
 19 | BUTTON_LABEL
 20 | ID
 21 | ACCOUNT_ID
 22 | TIMESTAMP
(6 rows)
```

src/metabase/driver/sql/query_processor.clj
```diff
@@ -1409,23 +1411,44 @@
...
+        needs-columns? (and
+                         (seq desired-aliases)
+                         (> (count desired-aliases) 1)
+                         duplicate-source-aliases?
+                         (apply distinct? desired-aliases)
+                         (every? some? desired-aliases)
+                         (every? string? desired-aliases))]
```
lbrdnk, line 1442: Why do we need to do the `some?` check prior to `string?`?

src/metabase/driver/sql/query_processor.clj
```diff
@@ -1409,23 +1411,44 @@
...
-
-                   :else
-                   (apply-clauses driver {} source-query))
-                 (let [table-alias (->honeysql driver (h2x/identifier :table-alias source-query-alias))]
-                   [table-alias])]]))
+  "Handle a `:source-query` clause by adding a recursive `SELECT` or native query.
+   If the source query has ambiguous column names, use a `WITH` statement to rename the source columns.
+   At the time of this writing, all source queries are aliased as `source`."
+  [driver honeysql-form {{:keys [native params] persisted :persisted-info/native :as source-query} :source-query
+                         source-metadata :source-metadata :as _x}]
```
lbrdnk, line 1418: Why do we need to alias third arg as `_x`?

test/metabase/query_processor_test/native_test.clj
```diff
@@ -37,6 +38,34 @@
            (mt/native-query
             {:query "select name from users;"}))))))
 
+(deftest ^:parallel native-with-duplicate-column-names
+  (testing "Should be able to run native query referring a question referring a question (#25988)"
+    (mt/with-test-drivers (sql-jdbc.tu/sql-jdbc-drivers)
```
lbrdnk, line 43: Test is running against jdbc drivers, but the change affects `sql.qp`. I'm wondering if the change could not impact also bigquery (parent there is sql).

lbrdnk (changes requested):
I've left some remarks to be resolved first.

test/metabase/query_processor_test/native_test.clj
snoe: great catch

lbrdnk (approved):
LGTM

src/metabase/driver/sql/query_processor.clj
```diff
@@ -1409,23 +1411,43 @@
...
+        ;; TODO: Use MLv2 here to get source and desired-aliases
+        alias-info (mapv (fn [{[_ desired-ref-name] :field_ref source-name :name}]
+                           [source-name desired-ref-name])
+                         source-metadata)
+        source-aliases (mapv first alias-info)
+        desired-aliases (mapv second alias-info)
+        duplicate-source-aliases? (and (> (count source-aliases) 1)
+                                       (not (apply distinct? source-aliases)))
+        needs-columns? (and (seq desired-aliases)
+                            (> (count desired-aliases) 1)
```
metamben, line 1438:
```suggestion
        needs-columns? (and (> (count desired-aliases) 1)
```

metamben (approved):
LGTM

PR #39490, opened 2024-03-02
Implement an order preserving tag hierarchy

crisptrutski:
### Description

This takes the previous PR further and encapsulates things further within the hierarchy abstraction.

A small practical benefit is that one no longer has to manually manually define the "more specific" parent for every node with multiple parents.

Having all the hierarchy verbs behave consistently as one expects removes some footguns.

PR #39489, opened 2024-03-02
Use hierarchy instead of home-rolled DAG structure

crisptrutski:
### Description

Use the built-in hierarchy data structure rather than an order sensitive list. This sets us up to leverage verbs like `isa?` for upcoming feature work.

PR #39491, opened 2024-03-03
Use toposort to simplify and fix CSV column type operations

crisptrutski:
### Description

This PR replaces the breadth first search, which gave an "unintuitive"[^1] ordering, with a topological sorting.

Given this stronger foundation, we can then replace the hard coded parsing order with a graph traversal. 

This also implicitly corrects a benign defect where `::date` and `::datetime` were out of order in value type inference.

[^1]: When traversing the ancestors of `::boolean-or-int`, we would reach `::varchar-255` before `::float`.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -32,7 +32,7 @@
    [metabase.util.i18n :refer [tru]]
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
-   [metabase.util.ordered-hierarchy :as ordered-hierarchy :refer [derive make-hierarchy parents]]
+   [metabase.util.ordered-hierarchy :as ordered-hierarchy :refer [derive make-hierarchy]]
    [toucan2.core :as t2])
   (:import
    (java.io File)))
@@ -76,6 +76,9 @@
 ;; </code></pre>
 
 (def ^:private h
+  "This hierarchy defines a relationship between value types and their specializations.
+  We use an [[metabase.util.ordered-hierarchy]] for its topological sorting, which simplify writing efficient and
+  consistent implementations for of our type inference, parsing, and relaxation."
   (-> (make-hierarchy)
       (derive ::boolean-or-int ::boolean)
       (derive ::boolean-or-int ::int)
@@ -88,9 +91,9 @@
       (derive ::float ::varchar-255)
       (derive ::varchar-255 ::text)))
 
-(def ^:private concrete-type?
-  "Nodes which can correspond to a concrete database column type."
-  (complement #{::boolean-or-int}))
+(def ^:private abstract->concrete
+  "Not all value types do not correspond to concrete database column types."
+  {::boolean-or-int ::boolean})
 
 (def ^:private value-types
   "All type tags which values can be inferred as. An ordered set from most to least specialized."
@@ -98,15 +101,14 @@
 
 (def ^:private column-types
   "All type tags that correspond to concrete column types. An ordered set from most to least specialized."
-  (into #{} (filter concrete-type?) value-types))
+  (into #{} (remove abstract->concrete) value-types))
 
 (defn ^:private column-type
   "The most specific column type corresponding to the given value type."
   [value-type]
-  ;; If we know nothing about the value type, treat it as an arbitrary string.
-  (cond (nil? value-type) ::text
-        (concrete-type? value-type) value-type
-        :else (recur (first (parents h value-type)))))
+  (or (abstract->concrete value-type value-type)
+      ;; If we know nothing about the value type, treat it as an arbitrary string.
+      ::text))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; [[value->type]] helpers
@@ -200,21 +202,23 @@
   [type->check value]
   (when-not (str/blank? value)
     (let [trimmed (str/trim value)]
-      (first (filter #(when-let [f (type->check %)] (f trimmed)) value-types)))))
+      (->> (remove non-inferable-types value-types)
+           (filter #((type->check %) trimmed))
+           first))))
 
 (defn- row->value-types
   [type->check row]
   (map (partial value->type type->check) row))
 
 (defn- most-specific-common-ancestor
-  "Return the first \"ancestor\" of `base-tag` which is also an \"ancestor\" of `new-tag`.
+  "Return the first \"ancestor\" of `base-type` which is also an \"ancestor\" of `new-type`.
   We use a more relaxed definition of \"ancestor\" here than usual, which includes the tag itself."
-  [base-tag new-tag]
-  (when (or base-tag new-tag)
-    (or (ordered-hierarchy/first-common-ancestor h base-tag new-tag)
+  [base-type new-type]
+  (when (or base-type new-type)
+    (or (ordered-hierarchy/first-common-ancestor h base-type new-type)
         (throw (IllegalArgumentException. (tru "Could not find a common type for {0} and {1}"
-                                               base-tag
-                                               new-tag))))))
+                                               base-type
+                                               new-type))))))
 
 (defn- coalesce-types
   "Given two collections of type tags, find the most specific common ancestor for each pair.
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -92,7 +92,7 @@
       (derive ::varchar-255 ::text)))
 
 (def ^:private abstract->concrete
-  "Not all value types do not correspond to concrete database column types."
+  "Not all value types correspond to database types. For those that don't, this maps to their concrete ancestor."
   {::boolean-or-int ::boolean})
 
 (def ^:private value-types
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -92,14 +92,21 @@
   "Nodes which can correspond to a concrete database column type."
   (complement #{::boolean-or-int}))
 
+(def ^:private value-types
+  "All type tags which values can be inferred as. An ordered set from most to least specialized."
+  (ordered-hierarchy/sorted-tags h))
+
 (def ^:private column-types
-  "All column types"
-  (into #{} (filter concrete-type?) (ordered-hierarchy/tags h)))
+  "All type tags that correspond to concrete column types. An ordered set from most to least specialized."
+  (into #{} (filter concrete-type?) value-types))
 
 (defn ^:private column-type
-  "Get the most specific concrete type corresponding to the given tag."
-  [tag]
-  (if (concrete-type? tag) tag (recur (first (parents h tag)))))
+  "The most specific column type corresponding to the given value type."
+  [value-type]
+  ;; If we know nothing about the value type, treat it as an arbitrary string.
+  (cond (nil? value-type) ::text
+        (concrete-type? value-type) value-type
+        :else (recur (first (parents h value-type)))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; [[value->type]] helpers
@@ -238,11 +245,7 @@
   (->> rows
        (map #(row->value-types % settings))
        (reduce coalesce-types (repeat column-count nil))
-       (map (fn [tag]
-              ;; if there's no values in the column, assume it's a string
-              (if (nil? tag)
-                ::text
-                (column-type tag))))))
+       (map column-type)))
 
 (defn- detect-schema
   "Consumes the header and rows from a CSV file.
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -208,8 +208,8 @@
                                                new-tag))))))
 
 (defn- coalesce-types
-  "Given two collections of types tags, find the first-common-ancestor for each pair.
-  If one of the collections is longer, we resolve to its literal value for the remaining length."
+  "Given two collections of type tags, find the most specific common ancestor for each pair.
+  If one of the collections is longer, we return its existing tags for the remaining length."
   [types-a types-b]
   (u/map-all most-specific-common-ancestor types-a types-b))
 
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -167,42 +167,44 @@
 (defn- boolean-or-int-string? [s]
   (contains? #{"0" "1"} s))
 
+(defn- varchar-255? [s]
+  (<= (count s) 255))
+
 ;; end [[value->type]] helpers
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(def ^:private non-inferable-types
+  #{::auto-incrementing-int-pk})
+
+(defn- assert-all-types-checks-present
+  [type->check]
+  (assert (every? type->check (remove non-inferable-types value-types))
+          "Every value-type must be registered in settings->type->check")
+  type->check)
+
+(defn- settings->type->check [{:keys [number-separators] :as _settings}]
+  (assert-all-types-checks-present
+   {::boolean-or-int  boolean-or-int-string?
+    ::boolean         boolean-string?
+    ::offset-datetime offset-datetime-string?
+    ::date            date-string?
+    ::datetime        datetime-string?
+    ::int             (partial re-matches (int-regex number-separators))
+    ::float           (partial re-matches (float-regex number-separators))
+    ::varchar-255     varchar-255?
+    ::text            (constantly true)}))
+
 (defn- value->type
-  "The most-specific possible type for a given value. Possibilities are:
-
-    - `::boolean`
-    - `::int`
-    - `::float`
-    - `::varchar-255`
-    - `::date`
-    - `::datetime`
-    - `::offset-datetime`
-    - `::text` (the catch-all type)
-
-  NB: There are currently the following gotchas:
-    1. ints/floats are assumed to use the separators and decimal points corresponding to the locale defined in the
-       application settings
-    2. 0 and 1 are assumed to be booleans, not ints."
-  [value {:keys [number-separators] :as _settings}]
-  (let [trimmed (str/trim value)]
-    (cond
-      (str/blank? value)                                        nil
-      (boolean-or-int-string? trimmed)                          ::boolean-or-int
-      (boolean-string? trimmed)                                 ::boolean
-      (offset-datetime-string? trimmed)                         ::offset-datetime
-      (datetime-string? trimmed)                                ::datetime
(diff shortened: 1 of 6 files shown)


crisptrutski:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/39491?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#39491** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39491?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#39490** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#39489** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39489?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

calherries:
The code I had reviewed in [Use hierarchy instead of home-rolled DAG structure](https://github.com/metabase/metabase/pull/39489) is being rewritten by the code in this PR. This means both reviewing the code and responding to the reviews is going to take more of everyone's time. So if you're going to rewrite the code several times I'd prefer if all the changes were commits in the same PR, not separate PRs.

calherries:
> This also implicitly corrects a small defect where ::date and ::datetime were out of order in value type inference.

Can you reproduce this bug with a test?

src/metabase/upload.clj
```diff
@@ -179,75 +165,62 @@
...
-      (re-matches (int-regex number-separators) trimmed)        ::int
-      (re-matches (float-regex number-separators) trimmed)      ::float
-      (<= (count trimmed) 255)                                  ::varchar-255
-      :else                                                     ::text)))
+  "Determine the most specific type that is compatible with the given value.
+  Numbers are assumed to use separators corresponding to the locale defined in the application settings"
+  [type->check value]
+  (when-not (str/blank? value)
+    (let [trimmed (str/trim value)]
+      (first (filter #(when-let [f (type->check %)] (f trimmed)) value-types)))))
```
calherries, line 203: How can `f` be `nil`? I don't think it can, so I would rewrite this code to use tighter assumptions
```clojure
(first (filter #((type->check %) trimmed) value-types)))))
```

src/metabase/upload.clj
```diff
@@ -74,53 +75,38 @@
...
+  "All type tags that correspond to concrete column types. An ordered set from most to least specialized."
+  (into #{} (filter concrete-type?) value-types))
+
+(defn ^:private column-type
+  "The most specific column type corresponding to the given value type."
+  [value-type]
+  ;; If we know nothing about the value type, treat it as an arbitrary string.
+  (cond (nil? value-type) ::text
+        (concrete-type? value-type) value-type
+        :else (recur (first (parents h value-type)))))
```
calherries, line 109: Do you have a test for this recursion?

src/metabase/upload.clj
```diff
@@ -74,53 +75,38 @@
...
+      (derive ::date ::datetime)
+      (derive ::boolean ::varchar-255)
+      (derive ::offset-datetime ::varchar-255)
+      (derive ::datetime ::varchar-255)
+      (derive ::float ::varchar-255)
+      (derive ::varchar-255 ::text)))
+
+(def ^:private concrete-type?
+  "Nodes which can correspond to a concrete database column type."
+  (complement #{::boolean-or-int}))
```
calherries, line 93: I find it strange to use insertion order to which type to fall back to. Maybe you have other reasons for it. Anyway, this is what I had in mind:

```clojure
(defn- specialise-ambiguous-type
  "For tags that correspond to more than one database type, specialise the tag so it corresponds to a single database type"
  {::boolean-or-int :boolean})
```

and then
```clojure
(defn ^:private column-type
  "The most specific column type corresponding to the given value type."
  [value-type]
  ;; If we know nothing about the value type, treat it as an arbitrary string.
  (cond (nil? value-type) ::text
        (specialise-ambiguous-type value-type) (specialise-ambiguous-type value-type)
        :else value-type))))
```

src/metabase/upload.clj
```diff
@@ -74,53 +75,38 @@
...
-    type))
-
-(def ^:private type->parents
-  (reduce
-   (fn [m [type parent]]
-     (update m type conj parent))
-   {}
-   type+parent-pairs))
+(def ^:private h
+  (-> (make-hierarchy)
```
calherries, line 79: I don't see the need for an ordered-hierarchy here, but if you have a reason, maybe leaving a comment here can help future readers understand the design decision

src/metabase/util/ordered_hierarchy.clj
```diff
@@ -0,0 +1,131 @@
...
+
+(defn- first-common-tag
+  "Given two ordered sets corresponding to ancestor in a hierarchy, return the first ancestor of tag-a which is also an
+  ancestor of tag-b.
+  Returns nil if there is no such intersection."
+  [ancestors-a ancestors-b]
+  ;; I suspect this is not commutative, so this ordering is important - we care more about being close to tag-a.
+  (some ancestors-b ancestors-a))
+
+(defn first-common-ancestor
```
calherries, line 116: If you move a function into a util namespace, you should write a test for it. Otherwise it's too easy to drop test coverage. Also, I think `least-common-ancestor` is the common name for this, see https://en.wikipedia.org/wiki/Lowest_common_ancestor

calherries (approved):
Much better! Ignore my previous quibbles on the ancestor PRs. This looks good enough to merge but I have a few suggested improvements

src/metabase/upload.clj
calherries: Copied from the other PR: to me explicitly defining the parent type to use for types with multiple parents is preferable to making it depend on the insertion order. Why would you choose to couple these two things?

Another very tangible downside of this approach is you've had to implement an ordered hierarchy to achieve the same behaviour.

src/metabase/upload.clj
crisptrutski: It will be nil for any `non-inferable-types`, e.g. `::auto-incrementing-int-pk`

src/metabase/upload.clj
crisptrutski: It's well covered by the high level tests, but I don't mind adding a unit test.

calherries: Really? what test will break if there's no recursion i.e.

```
:else (first (parents h value-type))
```

currently there's only one non-concrete type, so I can't see how this will break any high level tests

src/metabase/upload.clj
calherries: right, yes. then I think I would prefer this implementation to document this logic explicitly
```clojure
(first (filter #((type->check %) trimmed) (remove non-inferable-types value-types))))))
```

src/metabase/upload.clj
```diff
@@ -179,75 +165,62 @@
...
 ;; end [[value->type]] helpers
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(def ^:private non-inferable-types
+  #{::auto-incrementing-int-pk})
+
+(defn- assert-all-types-checks-present
+  [type->check]
+  (assert (every? type->check (remove non-inferable-types value-types))
+          "Every value-type must be registered in settings->type->check")
```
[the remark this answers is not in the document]
crisptrutski: The dev ex felt a bit better to me - instant feedback. This is a classic invariant that you'd catch with a static type checker.

Perhaps not a factor here, but when a bunch of distant tests break due to some trivial invariant being violated it's waste of time to read through the failures to diagnose. I guess a malli schema could solve this, but I still think a "compile time" error is a better experience.

Can shunt to a test if it feels deeply wrong to you - I'm still finding my style in this ecosystem.

src/metabase/util/ordered_hierarchy.clj
crisptrutski: Missing test coverage - good point, will add. 

In general the tests feel pretty thin, but to some extent I'm waiting on seeing whether it gets used again before putting further effort into testing etc.

The rename came from finding "lowest" quite confusing in the context of value types, as I intuitively found myself thinking of the leaves as the "highest" (most specific) designations. Since this is now abstracted away from types the revert makes sense. TIL some standardized terminology.

src/metabase/upload.clj
```diff
@@ -179,75 +165,62 @@
...
-     (when (or s1 s2)
-       (cons (f (first s1) (first s2))
-             (map-with-nils f (rest s1) (rest s2)))))))
+  [type->check row]
+  (map (partial value->type type->check) row))
+
+(defn- most-specific-common-ancestor
+  "Return the first \"ancestor\" of `base-tag` which is also an \"ancestor\" of `new-tag`.
+  We use a more relaxed definition of \"ancestor\" here than usual, which includes the tag itself."
+  [base-tag new-tag]
```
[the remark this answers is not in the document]
crisptrutski: Agree. Remnant of some of the back and forth noise in the PR stack.

src/metabase/util/ordered_hierarchy.clj
crisptrutski: On reflection, I just realized that this function is not actually calculating a LCA - it returns the first member in the toposort walk, which I'm not sure will always be the same as the first member in a BFS.

Given this uncertainty, I'd rather keep the current name which is semantically more precise.

calherries: Good point, the lowest common ancestor is not necessarily unique for a DAG. The name is good

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: I guess this feels a little strange because it's proving some property of `settings->type->check`, not checking the pre/post-conditions of each function call. 

Run-time assertions can be costly, but in this case it's not that bad because you're only running this assertion once per upload, which is a slow operation anyway.

I'm inclined to discourage this style in general, but at the same time this specific case is fine and not worth bikeshedding over

src/metabase/upload.clj
crisptrutski: The reason I like this abstraction is that it ties together a bunch of nouns in a consistent and intuitive way - in particular for CSV uploads these are the orderings for `parents`, `ancestors`, and `value-types`.

1. We depend on the order of parents in order to resolve ambiguity of which concrete column type to use for which overloaded node.

2. We depend on the order of the ancestors in order to determine the minimal relaxation of the base type in order to accommodate some new value we've encountered during appending. This lack of symmetry is the crux of why we moved to a topological sort from a breadth first walk.

3. We depend on the order of all values in order to determine how to parse values.

The fact that *all* collections use a consistent order avoids the error prone (as born out by the existing errors[^1]) and repetitive hard coding of the graph structure in column type resolution and parsing. Things would get even more complicated in terms of defining preferences if we had more multiple-parent relationships, which could mean having multiple lowest common ancestors. I'm not sure how we'd want to piecewise manually resolve that.

This change also removes the assumption that only "abstract" nodes can have multiple parents - I'm not sure that's fundamentally true, so it's nice to be future proof.

To summarize in a different way - we have a nice ASCII representation of the graph, and with this change all our behavior falls out of that picture. I think that's much simpler to reason about and extend than hard coding properties of that graph.

[^1]: There are two that pop to mind, although I (perhaps unreliably) recall a third one. 1) We considered `::varchar-255` to be a more specific ancestor of `::boolean-or-int` than `::float`. 2) We would try parsing a value as a `datetime` before trying as a `date`. This meant that if we "fixed" `parse-local-datetime` to handle plain dates (which would be required to allow mixing them in a column, as the hierarchy implies) then we would incorrectly parse every plain date as a datetime.

crisptrutski: As for writing a motivating docstring - I'll put that on the backburner as I've already spent a bunch of time "off piste" from the remaining CSV append tasks.

src/metabase/upload.clj
crisptrutski: I actually started out with exactly this approach, plus a test that we hadn't forgotten any entries in this map. To give some weight to your preference, this is how Clojure does things too - you specify these preferences per multi-method.

I'm quite firmly convinced this is the way to go though, and hope these reasons can persuade you as well:

1. For the linearization we use in the parser to be well defined, we need to specify an ordering for every parent, not just which is first. Perhaps we'll never have more than two parents for value-types, but an abstract `ordered-hierarchy` needs to handle this.
2. These preferences must be consistent with the ordering we use in the parser, and that seems error prone. If they're not consistent then you could get a different table from uploading one file or uploading it in two parts.
3. Having this map separate to the hierarchy means that we need an assert or unit test to ensure we haven't forgotten any entries.
4. Having this map separate to the derivations means that we can't maintain the ordering of all the internal nouns as an invariant.

The key thing I like about using order dependence like this is it establishes a 1-1 mapping between the derivation code and the graph visualization in our comment. We can just use that picture to reason about how our our type inference and relaxation works, and we can similarly see quickly how the picture maps to the code.

They key difference between Clojure's "a la carte" approach and ours, is that there isn't much value add for our hierarchy to be open - it makes sense in our case to define the full graph in one expression.

crisptrutski: While I really like using a `hierarchy` inspired data structure to power this graph, you've got me thinking that `make-hierarchy` and `derive` might not be the ideal interface for us to define the graph here - it's quite low level.

Perhaps you'd be less put off by a function that takes the full graph blueprint as arguments? It could be an ordered list of edges, which is a pretty shallow cosmetic change, or perhaps just a map. It would be *fairly* easy to turn a `parents` map into the corresponding derive statements (or even bypass them). I think a `children` map would be a bit more natural, at least it would map better to the visual depiction.

Taken to the extreme, we could even discard `derive` completely, and generate "closed" hierarchies that can't be extended. Even if you needed this extension you could still extract the parents map, modify that as a regular Clojure map, then pass that back into a hierarchy constructor.

Personally I find the "ordered list of edges" approach quite natural, and the original code before using a hierarchy already had an order dependency, in fact a more onerous one (all children before parents).

src/metabase/upload.clj
crisptrutski: Sounds good. I don't like leaving an NPE but our assert's got us covered.

crisptrutski:
> > This also implicitly corrects a small defect where ::date and ::datetime were out of order in value type inference.
> 
> Can you reproduce this bug with a test?

I think I was mistaken about this being a bug in practice.

We mitigate the "mistake" in the ordering by having a different "attempt to parse" function (`parse-local-datetime`) and actual parser (`parse-as-datetime`). The latter respects the hierarchy - I missed that detail at first.

I still think it's nice not to rely on that kind subtle implementation detail though.

src/metabase/upload.clj
crisptrutski: I'm sure you know this, but explaining my reasoning - if a file has a column with only 0s or 1s (`::boolean-or-int`), we rely on this line (coerce to `::boolean`) to avoid a runtime error when we call `driver/upload-type->database-type` - in my comment I assumed we have at least one such.

Checking now, I see we only have a test that appends these values to an existing table. This gap in the testing is surprising to me. 

I'll write a unit test for this `column-type`.

calherries:
As discussed on zoom: 
- let's make the preference for non-concrete types explicit, not using parent order
- ordered-hierarchy is cool for understanding/deterministic behaviour, so lets leave that as is
- add a docstring to `upload/h` to saying the main reason we use ordered-hierarchy is to get topologically sorted ancestors

src/metabase/upload.clj
crisptrutski: We had an offline discussion, and here are the main points:

1. If we need to identify which types are abstract in any case. A map isn't much more code than a set, and provides nice documentation. We could document the same things in a unit test, but it's nice to have an explicit source of truth.
2. Using the ordering of the graph to define these projections was not adequately explained in the code, and is less obvious than an explicit map.
3. There is at least one important constraint we have between this projection and the graph structure. [^1]
4. There doesn't seem to be any constraint between the topological ordering and which parent we project an abstract type on to, for things to work out between importing, appending, etc with the current hierarchy.
5. I'm personally not convinced that we won't discover more important constraints as the hierarchy becomes more complex.
6. Testing the constraints is complex if the orderings are inconsistent, and guaranteed if they are.
7. Aesthetically its nice to normalize the shape of the graph, and read the implicit conversions right off its visualization.

Outcome:

1. We're bringing back the concrete map for now.
2. We'll test it satisfies the required constraints.

[^1]: There should be a route between every abstract type and its concrete type projection that consists only of abstract nodes.

crisptrutski: Thinking about abstract and concrete types at a higher level, I wonder whether we shouldn't late bind more on this projection. On a philosophical level, perhaps that are concrete for some drivers should be abstract for others, due to the differences in the granularity of their types. 

Today we would work around this by having driver methods generate the same definition for both nodes, but this feels hacky and error prone to me. It might actually be simpler for the driver not to implement that multimethod and fallback to nil, and for those nodes to be our implicit abstract nodes. We could continue walking the ancestors until we hit a type that the database does understand. This feels much simpler and robust to me, and would avoid us breaking existing drivers whenever we refine the hierarchy.

If we still wanted some drivers to simply break on some unsupported types rather than reading them as text, we could define that explicitly.

This is all very speculative around how more sophisticated the hierarchy could become in future.

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: I agree that it feels strange - the shape of the map is not even dynamic. It's pity that it exists only to close over the settings. I thought about just having a static map whose functions all took settings as well, but then we'd be rebuilding the regular expressions constantly.

I've been assuming that assertions are disabled in production, I've never worked on a production Java app that leaves them enabled, but I guess we don't control how people run the JAR.

I'll send a small PR with using Malli instead. It's more declarative and always off in production. The error message seems unhelpful, but I'll chat to a Malli guru about that.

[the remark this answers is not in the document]
crisptrutski: PR: [Use malli to ensure all column types are covered](https://github.com/metabase/metabase/pull/39612)

calherries (approved):
Changes look good, thanks

src/metabase/upload.clj
calherries: Yep. A concrete example of this is JSON columns. Not all our drivers support JSON/JSONB typed columns. One solution is to have the hierarchy and functions like `abstract->concrete` to depend on the driver.

Issue #38958, opened 2024-02-20
CSV upload fails on rows with float type among mostly integer values


calherries:
https://metaboat.slack.com/archives/C04S696LRUM/p1707242309280519?thread_ts=1707238777.488509&cid=C04S696LRUM

crisptrutski:
I think that the issue here is that we infer the schema based on a [subset](https://github.com/metabase/metabase/blob/ddfd9f636072034b9bff9681f3391ad17ee39b60/src/metabase/upload.clj#L340) of the rows, and are missing the sparse float values.

The solution for this will build on https://github.com/metabase/metabase/issues/37069

crisptrutski:
The issue was indeed the row sampling. 

With the optimized inference method there's no longer any need to do the sampling, and the reference file loads correctly.

The merge for the fix is just waiting behind the implicit int to float column promotion PR.

PR #39493, opened 2024-03-03
Convert column to float when appending floats to an integer column

### LINKED ISSUE #37069: CSV appends: integer columns should be convertable to float
Integer columns in an upload table should be converted to a floating-point type if the uploaded column in the CSV file contains a floating-point number.

- [x]  Test: if a column is an integer type, uploading a single float should convert the column to float

crisptrutski:
Closes https://github.com/metabase/metabase/issues/37069

### Description

This adds a type detection step during append which checks for columns that would need to be relaxed, and does so if we've configured that transition as OK.

### How to verify

1. Upload a CSV with an all integer column as a new table.
2. View the table and confirm it has integer type.
3. Append a CSV with the same column containing floats to the same table.
4. View the table and confirm it has a float type, and numbers were not truncated.

### Demo

[Watch it in action](https://www.loom.com/share/47b7d927c8d94e41bfd09b8d5d8a950c?sid=63f59a70-3008-4b46-b334-0008e4830831)

### DIFF
--- a/src/metabase/driver.clj
+++ b/src/metabase/driver.clj
@@ -924,7 +924,7 @@
 (defmulti alter-columns!
   "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
   Currently we do not currently support changing the the primary key."
-  {:added "0.??.0", :arglists '([driver db-id table-name column-definitions])}
+  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
--- a/src/metabase/driver.clj
+++ b/src/metabase/driver.clj
@@ -923,7 +923,7 @@
 
 (defmulti alter-columns!
   "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
-  Currently we do not currently support changing the the primary key."
+  Currently we do not currently support changing the the primary key, or take any guidance on how to coerce values."
   {:added "0.49.0", :arglists '([driver db-id table-name column-definitions])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
--- a/src/metabase/driver.clj
+++ b/src/metabase/driver.clj
@@ -921,6 +921,13 @@
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
+(defmulti alter-columns!
+  "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
+  Currently we do not currently support changing the the primary key."
+  {:added "0.??.0", :arglists '([driver db-id table-name column-definitions])}
+  dispatch-on-initialized-driver
+  :hierarchy #'hierarchy)
+
 (defmulti syncable-schemas
   "Returns the set of syncable schemas in the database (as strings)."
   {:added "0.47.0", :arglists '([driver database])}
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -267,16 +267,11 @@
    The value of `extant-columns` and `generated-columns` is an ordered map of normalized-column-name -> type for the
    given CSV file. Supported types include `::int`, `::datetime`, etc. A column that is completely blank is assumed to
    be of type `::text`."
-  [header rows]
-  (let [normalized-header (->> header
-                               (map normalize-column-name))
-        unique-header     (->> normalized-header
-                               mbql.u/uniquify-names
-                               (map keyword))
-        column-count      (count normalized-header)
-        settings          (upload-parsing/get-settings)
-        col-name+type-pairs (->> rows
-                                 (column-types-from-rows settings column-count)
+  [settings header rows]
+  (let [normalized-header   (map normalize-column-name header)
+        unique-header       (map keyword (mbql.u/uniquify-names normalized-header))
+        column-count        (count normalized-header)
+        col-name+type-pairs (->> (column-types-from-rows settings column-count rows)
                                  (map vector unique-header))]
     {:extant-columns    (ordered-map/ordered-map col-name+type-pairs)
      :generated-columns (ordered-map/ordered-map auto-pk-column-keyword ::auto-incrementing-int-pk)}))
@@ -335,14 +330,12 @@
 (defn- parse-rows
   "Returns a lazy seq of parsed rows, given a sequence of upload types for each column.
   Empty strings are parsed as nil."
-  ([col-upload-types rows]
-   (parse-rows (upload-parsing/get-settings) col-upload-types rows))
-  ([settings col-upload-types rows]
-   (let [parsers (map #(upload-parsing/upload-type->parser % settings) col-upload-types)]
-     (for [row rows]
-       (for [[value parser] (u/map-all vector row parsers)]
-         (when-not (str/blank? value)
-           (parser value)))))))
+  [settings col-upload-types rows]
+  (let [parsers (map #(upload-parsing/upload-type->parser % settings) col-upload-types)]
+    (for [row rows]
+      (for [[value parser] (u/map-all vector row parsers)]
+        (when-not (str/blank? value)
+          (parser value))))))
 
 (defn- remove-indices
   "Removes the elements at the given indices from the collection. Indices is a set."
@@ -379,12 +372,13 @@
   [driver db-id table-name ^File csv-file]
   (with-open [reader (bom/bom-reader csv-file)]
     (let [[header & rows]         (without-auto-pk-columns (csv/read-csv reader))
-          {:keys [extant-columns generated-columns]} (detect-schema header rows)
+          settings                (upload-parsing/get-settings)
+          {:keys [extant-columns generated-columns]} (detect-schema settings header rows)
           cols->upload-type       (merge generated-columns extant-columns)
           col-definitions         (column-definitions driver cols->upload-type)
           csv-col-names           (keys extant-columns)
           col-upload-types        (vals extant-columns)
-          parsed-rows             (vec (parse-rows col-upload-types rows))]
+          parsed-rows             (vec (parse-rows settings col-upload-types rows))]
       (driver/create-table! driver
                             db-id
                             table-name
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -665,7 +665,8 @@
           ;; for now we just plan for the worst and perform a fairly expensive operation to detect any type changes
           ;; we can come back and optimize this to an optimistic-with-fallback approach later.
           type->value->type  (partial relax-type (settings->type->check settings))
-          relaxed-types      (->> (sample-rows rows)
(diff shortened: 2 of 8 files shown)


crisptrutski:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/39493?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#39493** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39493?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#39491** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39491?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#39490** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#39489** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39489?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

crisptrutski:
There's a regression, where we're upgrading the type unnecessarily.

```
Uploading 2.0 into a column of type int should be coerced to 2
expected: [[1 2]]
  actual: [[1 2.0]]
    diff: - [[nil 2]]
          + [[nil 2.0]]
```

src/metabase/driver.clj
```diff
@@ -921,6 +921,13 @@
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
+(defmulti alter-columns!
+  "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
+  Currently we do not currently support changing the the primary key."
+  {:added "0.??.0", :arglists '([driver db-id table-name column-definitions])}
```
calherries, line 927: You should be able to backport this relatively easily
```suggestion
  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions])}
```

calherries: It needs an entry in the driver changelog too.

src/metabase/upload.clj
```diff
@@ -347,13 +357,14 @@
...
   Empty strings are parsed as nil."
-  [col-upload-types rows]
-  (let [settings (upload-parsing/get-settings)
-        parsers  (map #(upload-parsing/upload-type->parser % settings) col-upload-types)]
-    (for [row rows]
-      (for [[value parser] (u/map-all vector row parsers)]
-        (when-not (str/blank? value)
-          (parser value))))))
+  ([col-upload-types rows]
+   (parse-rows (upload-parsing/get-settings) col-upload-types rows))
```
calherries, line 361: Why the change?

src/metabase/upload.clj
```diff
@@ -619,8 +655,25 @@
...
+          new-column-types   (map #(if (matching-or-upgradable? %1 %2) %2 %1) old-column-types relaxed-types)
+          _                  (when (and (not= old-column-types new-column-types)
+                                        ;; if we cannot coerce all the columns, don't bother coercing any of them
+                                        ;; we will instead throw an error when we try to parse as the old type
+                                        (= relaxed-types new-column-types))
+                               (let [fields (map normed-name->field normed-header)]
+                                 (->> (changed-field->new-type fields old-column-types relaxed-types)
+                                      (alter-columns! driver database table))))
+          ;; this will fail if any of our required relaxations were rejected.
+          parsed-rows        (parse-rows new-column-types rows)]
```
calherries, line 676: Oh I think you meant to use `settings` here.
```suggestion
          parsed-rows        (parse-rows settings new-column-types rows)]
```

src/metabase/upload.clj
```diff
@@ -619,8 +655,25 @@
...
           _                  (check-schema (dissoc normed-name->field auto-pk-column-name) header)
-          col-upload-types   (map (comp base-type->upload-type :base_type normed-name->field) normed-header)
-          parsed-rows        (parse-rows col-upload-types rows)]
+          settings           (upload-parsing/get-settings)
+          old-column-types   (map (comp base-type->upload-type :base_type normed-name->field) normed-header)
+          ;; in the happy, and most common, case all the values will match the existing types
+          ;; for now we just plan for the worst and perform a fairly expensive operation to detect any type changes
+          ;; we can come back and optimize this to an optimistic-with-fallback approach later.
+          type->value->type  (partial relax-type (settings->type->check settings))
+          relaxed-types      (->> (sample-rows rows)
```
calherries, line 664: I think we should forget sampling rows here. Uploaded CSVs can only be 50 MB and parsing them should be lightning fast (in the future that is). What we want right now is correctness first, speed second.

src/metabase/upload.clj
crisptrutski: 1. By passing this in the 3-arity function becomes pure. I like purity 😄 
2. There's a bit of work to get the settings, so it's wasteful to fetch it multiple times.

An earlier version of this was passing it in somewhere, but I see that's gone. There's still one spot I meant to update and missed.

src/metabase/upload.clj
crisptrutski: Yip, thanks

src/metabase/upload.clj
crisptrutski: Should we remove it from uploads as well then, as a short-term fix for [missing the floats in a mostly int column](https://github.com/metabase/metabase/issues/38958)?

src/metabase/upload.clj
```diff
@@ -619,8 +655,25 @@
                               (driver/create-auto-pk-with-append-csv? driver)
                               (not (contains? normed-name->field auto-pk-column-name)))
           _                  (check-schema (dissoc normed-name->field auto-pk-column-name) header)
-          col-upload-types   (map (comp base-type->upload-type :base_type normed-name->field) normed-header)
-          parsed-rows        (parse-rows col-upload-types rows)]
+          settings           (upload-parsing/get-settings)
```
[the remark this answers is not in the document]
crisptrutski: The purity is the main thing for me.

src/metabase/upload.clj
```diff
@@ -209,6 +209,16 @@
...
 
+(defn- relax-type
+  "Given an existing column type, and a value to insert into it, relax the type until it can parse the value."
+  [type->check current-type value]
+  (cond (nil? value) current-type
+        (nil? current-type) (value->type type->check value)
+        :else (let [trimmed (str/trim value)]
+                (->> (cons current-type (ancestors h current-type))
+                     (filter #((type->check %) trimmed))
+                     first))))
```
calherries, line 220: Could we use `allowed-type-upgrades` to limit the type checks? Then relax-type can return only allowed upgrades.

src/metabase/upload.clj
```diff
@@ -619,8 +655,25 @@
...
+          settings           (upload-parsing/get-settings)
+          old-column-types   (map (comp base-type->upload-type :base_type normed-name->field) normed-header)
+          ;; in the happy, and most common, case all the values will match the existing types
+          ;; for now we just plan for the worst and perform a fairly expensive operation to detect any type changes
+          ;; we can come back and optimize this to an optimistic-with-fallback approach later.
+          type->value->type  (partial relax-type (settings->type->check settings))
+          relaxed-types      (->> (sample-rows rows)
+                                  (reduce #(u/map-all type->value->type %1 %2) old-column-types)
+                                  (map column-type))
+          new-column-types   (map #(if (matching-or-upgradable? %1 %2) %2 %1) old-column-types relaxed-types)
```
calherries, line 667: Intuitively, I feel like we should be able to skip this line. During the reduction we should be able to get a list of columns that have new column types by construction, rather than checking whether they are matching afterwards. Maybe there's a reason why you've separated these two things though.

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: Oh. Well in that case I just disagree that purity is worth it here, but I do see some benefits so if you feel strongly about it let's keep it

src/metabase/upload.clj
crisptrutski: The idea is to avoid projecting to concrete types until we've finished relaxing the type, to avoid a case like turning `::boolean-or-int` to `::boolean` before we see a later `::int`. Only once we turn each into a `column-type` can we check whether this each "upgrade" is allowed.

This is a bit theoretical, since our only abstract type is a leaf, but I'd rather have a bit more code and computation than make assumptions that could be painful to remove later.

src/metabase/upload.clj
crisptrutski: That's a nice idea, but I was thinking to leave this method agnostic of whether it's being used for append. That way we could also use it if we chunk up the original insertion, in which case we'd always allow the upgrade. I guess we could pass in `allow-upgrade?` as an argument though... 

Another subtle issue with this approach occurs if we had any non-leaf ambiguous nodes, see my other comment about wanting to wait until we've converted to `column-type`.  Perhaps we'll never have a use case for non-leaf non-column types, but I'd prefer not to make limiting assumptions in areas where there is bigger low hanging fruit for optimization. 

Simplicity is another thing, but in this case having separate lines where we relax and then check if it's allowed is arguably simpler in a Hickey-ian sense than a function that interleaves both.

I'll add this idea to a larger backlog optimization issue, since it can be made to work around the abstract node thing.

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: My stance is that you need a really good reason for making things impure. I might get worn down over time by how little existing code is pure.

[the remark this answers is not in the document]
calherries: +1 on criterium, we should add it to `:dev`. I sometimes go to the trouble of using it myself. Point noted on not taking microbenchmarks too seriously, although I do find them handy as ballpark estimates.

> My stance is that you need a really good reason for making things impure.

I guess I'm not so much of a purist haha. Pure code makes some code hard to read for little benefit.

calherries (approved):
Works for me!

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: All benchmarks are better than guesses, but it'll be really nice to a solid quite of real world benchmarks too.

Code being complex can be a good reason itself, unfortunately its so subjective.

When you just need environmental state, bundling that all up in the first argument is a really easy convention that composes pretty well. You can always partially apply the function to get back the impure version. 

This "context first arg" is the approach I typically take for injecting things like database connections, metrics emitters, etc, even in imperative languages like Java. I find it immensely useful to be able to trace where effects can happen - things get so complex otherwise!

crisptrutski:
> Works for me!

Thanks for the approval! Unfortunately this isn't merge-able yet as it has one failing test. It turns out that the innocuous looking test needed a pretty complex change to fix. I'm going to stack that fix on top of this PR as in many ways it's orthogonal to this one.

crisptrutski:
@calherries This should be good-to-go now, just need feedback on where I put the new multi-method.

Issue #38229, opened 2024-01-29
[Epic] Capture more metadata: table count, field usage, last viewed/used_at


luizarakaki:
Technical doc: https://www.notion.so/metabase/Accurate-fingerprint-729a720cdd4f41029a49d647fa9583b2?pvs=4
Related to https://www.notion.so/metabase/Capture-field-and-table-usage-metadata-990e99226d234d8b916478109162cd42?pvs=4

```[tasklist]
### Milestone 1: Sync table row count (Postgres only)
- [ ] https://github.com/metabase/metabase/pull/39670
- [ ] https://github.com/metabase/metabase/pull/39494
```

```[tasklist]
### Milestone 2: Count field usage
- [x] Add new `field_usage` table
- [ ] https://github.com/metabase/metabase/issues/40494
```

```[tasklist]
### Milestone 3: Capture last\_viewed\_at and last\_used\_at
- [ ] #41006 
- [ ] https://github.com/metabase/metabase/pull/41748
```

The use case for Milestone 3 is [Auto cleanup collections](https://www.notion.so/metabase/Auto-cleanup-collections-54b3b58e254a48e69b26a054c84fd944)
For this project, we need signal that an item (collection, question, model, dashboard) is irrelevant.
For dashboards, `last_viewed_at` is enough.
For questions and models, we need to know if it is used by other things (like nested questions, sandbox definition, custom field values, goals), so we are proposing a `last_used_at` property that registers everytime that item is used.

[New technical design doc](https://www.notion.so/metabase/Capturing-last_used_at-last_viewed_at-45f16ce82494477688733b05596743fb)

```[tasklist]
### Milestone 4: Capture view count for questions, models, and dashboards.
- [ ] https://github.com/metabase/metabase/pull/41840
- [ ] https://github.com/metabase/metabase/pull/41886
- [ ] https://github.com/metabase/metabase/pull/41882
```
The use case for Milestone 4 is TBD, but most likely this would be shown in collections to help users find relevant content or help rank search results. 
[Milestone 4 technical doc](https://www.notion.so/metabase/e12aa6ef472c4d6ab871f60d1947327b?v=fd15329da9594408807fc38c3a4edb7a&p=086dec987ce84759baca325399c43da5&pm=s)

qnkhuat:
Technical doc https://www.notion.so/metabase/Accurate-fingerprint-729a720cdd4f41029a49d647fa9583b2?pvs=4

PR #39494, opened 2024-03-03
Postgres: Sync table row count

qnkhuat:
Milestone 1 of : https://github.com/metabase/metabase/issues/38229, part of https://github.com/metabase/metabase/issues/39176

This epic is about introducing a new fingerprint process that will do a full table scan on the table `select count(distinct(name)) from table` for example.

In order to not blow up DW, we want to limit to only doing this on reasonably small tables (< 1M rows to start with) by syncing the `estimated-row-count` of a table.


resources/migrations/001_update_migrations.yaml
```diff
@@ -5598,6 +5598,18 @@ databaseChangeLog:
...
+  - changeSet:
+      id: v50.2024-03-02T17:41:00
+      author: qnkhuat
+      comment: Add metabase_table.properties
+      changes:
+        - addColumn:
+            tableName: metabase_table
+            columns:
+              - column:
+                  name: properties
```
crisptrutski, line 5610: I'm curious how much thought has been given to the name "properties" - being so generic, are we intentionally leaving it open for ad hoc future use cases? I personally like to error towards narrow and specific, as one can always add further columns. What do you think about calling it something like `statistics`?

src/metabase/driver.clj
```diff
@@ -971,3 +974,12 @@
   {:added "0.48.0", :arglists '([driver database & args])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
+
+(defmulti schema+table->row-count
+  "Returns a map of [schema, table name] -> the estimated row count of all tables in the current database.
```
crisptrutski, line 979: nitpick: "the estimated row count of all tables ..." sounds like a summation over every table, how about:

> Returns a map of [schema, table name] -> the estimated row count, for each table in the current database.

src/metabase/driver/postgres.clj
```diff
@@ -892,3 +894,20 @@
...
+   driver
+   database
+   nil
+   (fn [conn]
+     (try
+      (into {}
+            (map (fn [{:keys [schemaname relname n_live_tup]}]
+                   [[schemaname relname] n_live_tup]))
+            ;; https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ALL-TABLES-VIEW
+            (next.jdbc/plan conn ["SELECT schemaname, relname, n_live_tup FROM pg_stat_user_tables"]))
```
crisptrutski, line 910: What do you think about using some "as" aliases in the query to make the mapping function read a bit nicer?

e.g. `(fn [{:keys [schema table row_count]}] ...)`

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -206,15 +215,11 @@
...
-              (into #{} (sql-jdbc.sync.interface/active-tables driver conn inclusion-patterns exclusion-patterns)))
-            (default-active-tbl-fn))
-          (default-active-tbl-fn)))))})
+            [inclusion-patterns
+             exclusion-patterns]    (when (and has-schema-filter-prop? (some? database))
+                                      (driver.s/db-details->schema-filter-patterns
+                                       (:name schema-filter-prop)
+                                       database))]
+        (->> (sql-jdbc.sync.interface/active-tables driver conn inclusion-patterns exclusion-patterns)
+             (tables-with-properties driver database)
```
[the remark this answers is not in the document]
qnkhuat: I'm not fond of `join` as it makes me think of join in SQL, wdyt about `hydrate-table-properties`? `add` would be okay to IMO.

src/metabase/sync/interface.clj
```diff
@@ -12,7 +12,10 @@
    [:schema         [:maybe ::lib.schema.common/non-blank-string]]
    [:require-filter {:optional true} :boolean]
    ;; `:description` in this case should be a column/remark on the Table, if there is one.
-   [:description    {:optional true} [:maybe :string]]])
+   [:description    {:optional true} [:maybe :string]]
+   [:properties     {:optional true} [:maybe
+                                      [:map {:closed true}
```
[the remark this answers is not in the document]
qnkhuat: I think it's a bit of a taste, IMO sync interface should all be closed as it forces us to keep the schema up-to-date as it's easy to forget about it :)

resources/migrations/001_update_migrations.yaml
qnkhuat: I indeed wanted this to be generic, IMO optional fields like `show_in_getting_started` or `database_require_filter` could fall into this category. 

It actually started as Cam suggested me to put `database_require_filter` into a json column like this in the first place, and I think that's a good idea :)

src/metabase/sync/interface.clj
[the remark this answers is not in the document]
crisptrutski: In this case it might be annoying if we try to work on an old branch with a new database. I don't think that trumps your reasoning, I was just curious, and in any case also a conservative clam when it comes to typing sensibilities 😄

src/metabase/driver/sql_jdbc/sync/describe_database.clj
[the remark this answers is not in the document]
crisptrutski: I'm not super keen on `hydrate`, as we we use that term for toucan models. How about `calculate-table-properties`? Still fine with `add` as well, but thinking now it's nice to indicate that this function "does some work".

resources/migrations/001_update_migrations.yaml
crisptrutski: If we explicitly want this to be a wide assortment the name is fine I guess. I'm slightly allergic to generic names but don't have a counter suggestion.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -193,6 +193,15 @@
         :else
         nil))
 
+(defn- tables-with-properties [driver database tables]
+  (if (driver/database-supports? driver :sync-row-count database)
+    (let [schema+table->row-count (when (driver/database-supports? driver :sync-row-count database)
+                                    (driver/schema+table->row-count driver database))]
+      (eduction (map (fn [{:keys [schema name] :as table}]
```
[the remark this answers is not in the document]
qnkhuat: `tables` here is an eduction and I don't want to turn it into a seq yet.

src/metabase/driver.clj
qnkhuat: [81087df](https://github.com/metabase/metabase/pull/39494/commits/81087df06972202cf006c1f21214e18071bc7f24)

src/metabase/driver/sql_jdbc/sync/describe_database.clj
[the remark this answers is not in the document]
qnkhuat: I'm not sure about `calculate` since to me it means doing some arithmetic calculation, but here we just fetch the row count from DB.

That's said I went with `add-table-properties` here :)

crisptrutski:
I've been burned by these estimates being by off by orders of magnitude. They're typically used for query planning and my experience with them is biased to when I've needed to debug sudden performance issues, tracing back to illogical query plans, tracing back to bogus numbers. My experience was with MySQL, and quite some time ago, and I'm not sure how far the tech has come since then, but I have some lingering apprehension. 

What is the use case for these the counts, and how tolerant will that be of volatility and inaccuracy?

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -193,6 +193,14 @@
         :else
         nil))
 
+(defn- add-table-properties [driver database tables]
+  (if (driver/database-supports? driver :sync-row-count database)
+    (let [schema+table->row-count (driver/schema+table->row-count driver database)]
+      (eduction (map (fn [{:keys [schema name] :as table}]
+                       (assoc table :properties {:row-count (get schema+table->row-count [schema name])})))
```
crisptrutski, line 200: Can we rename this to `:estimated-row-count`? I'm worried about future features not taking these numbers with sufficient salt.

test/metabase/driver/postgres_test.clj
```diff
@@ -234,8 +234,23 @@
...
 
-(defn- default-table-result [table-name]
-  {:name table-name, :schema "public", :description nil})
+(defn- default-table-result
+  ([table-name]
+   (default-table-result table-name {}))
+  ([table-name opts]
+   (merge
+    {:name table-name :schema "public" :description nil
+     ;; row-count are estimated, so the value can't be know for sure "during" test without
```
crisptrutski, line 243: grammar nit: 

row-count *is* estimated, so the value can't be *known* for sure .. So for tests that *don't* ~~concern~~ require ...

qnkhuat: updated in https://github.com/metabase/metabase/pull/39494/commits/1ef0ec58be8a5ae875ec3c0a42cb13d219d6b94e. Thanks!

src/metabase/driver/sql_jdbc/sync/describe_database.clj
qnkhuat: yep, renamed in https://github.com/metabase/metabase/pull/39494/commits/1ef0ec58be8a5ae875ec3c0a42cb13d219d6b94e

qnkhuat:
@crisptrutski I've updated the PR description to better clarify the usage of `estimated-row-count`, I'd not say we require this row count to be highly accurate but if it's order of magnitude off then that's no good.

At least in PG, the auto vacuum process runs every minute (by default), so I'd expect it not to be "that" off https://www.postgresql.org/docs/current/routine-vacuuming.html#AUTOVACUUM

test/metabase/driver/postgres_test.clj
```diff
@@ -1303,3 +1320,22 @@
...
+                               [{:field-name "name"
+                                 :base-type  :type/Text}]
+                               [["Los Angeles"]
+                                ["Las Vegas"]]]
+        (sql-jdbc.execute/do-with-connection-with-options
+         :postgres
+         (mt/db)
+         nil
+         (fn [conn]
+           ;; row count are estimated so we VACUUM so the statistic table is updated before syncing
```
crisptrutski: I'm not sure how precise you expect the stats to be afterward a vacuum, but as far as I know it won't directly update them and the improvements happening will be indirect through updating things like the visibility map. I think you would want to call `ANALYZE`. Even then I believe it does sampling and there are no guarantees at scale.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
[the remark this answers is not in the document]
crisptrutski: Fair enough. For me touching a database is doing work 😄

crisptrutski (approved):

crisptrutski:
> @crisptrutski I've updated the PR description to better clarify the usage of `estimated-row-count`, I'd not say we require this row count to be highly accurate but if it's order of magnitude off then that's no good.

Large errors will be pretty uncommon. They'll mostly affect tables that are already pretty large, probably outside that limit.

resources/migrations/001_update_migrations.yaml
```diff
@@ -5598,6 +5598,18 @@ databaseChangeLog:
             sql: >-
               DROP VIEW IF EXISTS v_content;
 
+  - changeSet:
+      id: v50.2024-03-02T17:41:00
+      author: qnkhuat
+      comment: Add metabase_table.properties
```
calherries, line 5604: Remind me, what's the benefit of stuffing this data in a JSON blob over nullable columns? I've tried to make my own list here and I'm not sure it stacks up:

The benefits:
- you can separate fields that are generally required from fields that are specific to certain driver features?
- you don't have to update as many tests

The downsides:
- harder to migrate
- harder to inspect with SQL
- slower to filter/select on

I genuinely feel like I'm missing something though.

src/metabase/driver.clj
```diff
@@ -971,3 +974,12 @@
   {:added "0.48.0", :arglists '([driver database & args])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
+
+(defmulti schema+table->estimated-row-count
+  "Returns a map of [schema, table name] -> the estimated row count, for each table in the current database.
```
calherries, line 979: I know we hold all tables in memory in a few places in sync code, but I'm wondering whether we should aim for better and instead return a reducible collection here, so we can choose not to hold everything in memory in the future.

Based on [this question on hosting insights](https://stats.metabase.com/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyNiwidHlwZSI6InF1ZXJ5IiwicXVlcnkiOnsiYWdncmVnYXRpb24iOltbImNvdW50Il1dLCJicmVha291dCI6W1siZmllbGQiLCJpbnN0YW5jZV9pZCIseyJiYXNlLXR5cGUiOiJ0eXBlL1VVSUQifV0sWyJmaWVsZCIsImRhdGFiYXNlX2lkIix7ImJhc2UtdHlwZSI6InR5cGUvSW50ZWdlciJ9XV0sIm9yZGVyLWJ5IjpbWyJkZXNjIixbImFnZ3JlZ2F0aW9uIiwwXV1dLCJzb3VyY2UtdGFibGUiOiJjYXJkX18xMzQzNCJ9fSwiZGlzcGxheSI6InRhYmxlIiwicGFyYW1ldGVycyI6W10sInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9LCJvcmlnaW5hbF9jYXJkX2lkIjoxMzQzNH0=) we have some pretty large databases out there in the wild. 

I tested the memory footprint of a map similar to the one returned by `schema+table->estimated-row-count` with 1,000,000 tables, and it comes to over 200 MB. I'm pretty sure we'd prefer not to do that.

Here's my test code:
```
(require '[metabase.test :as mt])

(defn gc []
  (dotimes [_ 4] (System/gc)))

(defn used-memory []
  (let [runtime (Runtime/getRuntime)]
    (gc)
    (- (.totalMemory runtime) (.freeMemory runtime))))

(defn mbs [f]
  (let [before (used-memory)
        _ (f)
        after (used-memory)]
    (int (/ (- after before) (* 1024 1024)))))

(mbs (fn []
           (def x (into {} (take 1000000 (map vector (repeatedly (fn [] [(mt/random-name) (mt/random-name)])) (repeat 1000)))))))
=> 232 (MB)
```

calherries: To be clear I'm not advocating against holding this data in memory now, as we're currently doing that elsewhere so this is not a memory bottleneck. I'm just suggesting to have a path to not holding the data in memory without having to change the driver method implementations.

calherries (commented):
This is the second time in recent history where we're adding a new query to save table metadata during sync where we join its results with the results of another query in Clojure. The last time was when we introduced the current user table privileges check. Now it's row counts.

Every time we add a new query it costs time, memory, and extra clojure code to join the data in clojure-land and iterate/reduce over the results. We currently have these queries for table metadata:

1. Table name, schema, description and type
2. Current user select privileges by table
3. Current user create/update/delete privileges by table

This PR will add a fourth query for table row counts. I wonder, when do we bite the bullet and just write a single query for all table metadata per driver?

Switching to do all of this in one big query will never make sense on an individual feature level, but looking at the system as a whole, it will be a lot easier to change and optimize.

One concrete benefit of having one query is this feature would have been easier to implement. You would just add a join to the query. Instead, you've had to create a new driver multimethod which comes packaged with assumptions about memory usage that I think we'll want to break one day.

Another concrete benefit is we can solve memory worries once and for all, instead of having ad-hoc strategies to iterate/page over results per query.

The downside of this proposed approach is that we'd have to maintain the implementation of the logic that JDBC's `getTables` gives us. But my guess is that's not a big cost compared to the benefits. Take a look at the [postgres implementation](https://github.com/davecramer/pgjdbc/blame/2fa24b3a0939c193e79e1ba5f71cb260eae4b148/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L1272) of `getTables`: I can't see any changes in the last 5 years that would have required us to update an implementation that meets our needs.

Thoughts? Should we pause this project to refactor first?

camsaul:
We used to sync row count but took it out because it was extremely slow. Putting `select count(distinct(name)) from table` back in is probably going to make things extremely slow again.

resources/migrations/001_update_migrations.yaml
camsaul: Agree with Cal. I think this just belongs in a regular column. The old version was a regular column

src/metabase/driver.clj
camsaul: 200 MB is really not ok, we want people (at least theoretically) to be able to run Metabase on Heroku or small EC2 instances. We need to rework this to not hold everything in memory at once. Metabase shouldn't need a gig of RAM to sync a database

src/metabase/sync/interface.clj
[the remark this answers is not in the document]
camsaul: I prefer leaving maps open, that way drivers can be backwards compatible with older versions if we introduce a new key somewhere.

camsaul (dismissed):
I think Cal is right and the estimated row count should be its own column, its doesn't make sense to introduce another JSON blob for this.

Also please let's just stick with the estimated row count and not introduce any full table scans. Take that out of the epic. It's going to destroy sync performance (already bad) and make everyone mad at us. We took it out for a reason

qnkhuat:
@camsaul this will now be synced to `metabase_table.esimated_row_count`.

> Also please let's just stick with the estimated row count and not introduce any full table scans. Take that out of the epic. It's going to destroy sync performance (already bad) and make everyone mad at us. We took it out for a reason

we(me and Arakaki and Cal) had a chat about this and we agreed to put a hold on that.

resources/migrations/001_update_migrations.yaml
```diff
@@ -6055,6 +6055,19 @@ databaseChangeLog:
...
+      id: v50.2024-03-21T17:41:00
+      author: qnkhuat
+      comment: Add metabase_table.estimated_row_count
+      changes:
+        - addColumn:
+            tableName: metabase_table
+            columns:
+              - column:
+                  name: estimated_row_count
+                  type: integer
```
calherries, line 6068: This needs to be a biginteger type to match `pg_stat_user_tables.n_live_tup`. Otherwise it'll fail for tables over ~2 billion rows

test/metabase/driver/postgres_test.clj
```diff
@@ -264,9 +279,9 @@
                        ["DROP MATERIALIZED VIEW IF EXISTS test_mview;
                        CREATE MATERIALIZED VIEW test_mview AS
                        SELECT 'Toucans are the coolest type of bird.' AS true_facts;"])
-        (t2.with-temp/with-temp [Database database {:engine :postgres, :details (assoc details :dbname "materialized_views_test")}]
-          (is (= {:tables #{(default-table-result "test_mview")}}
-                 (driver/describe-database :postgres database))))))))
+        (mt/with-temp [:model/Database database {:engine :postgres, :details (assoc details :dbname "materialized_views_test")}]
+          (is (=? [(default-table-result "test_mview" {:estimated_row_count (mt/malli=? int?)})]
```
calherries, line 283: Just for my understanding, why do you not have to execute `VACUUM` here but you do in the [other test](https://github.com/metabase/metabase/pull/39494/files#diff-821f0be02cebbea5f2f5b86239bd9df5e5cb4f671634a761fa6da1c511487125R1429)?

calherries (dismissed):
Sweet, this looks pretty neat now

I left a few minor suggestions, but I'll approve once this is resolved https://github.com/metabase/metabase/pull/39494/files#r1534324338.

src/metabase/sync/interface.clj
```diff
@@ -9,11 +9,12 @@
 
 (mr/def ::DatabaseMetadataTable
   [:map
-   [:name           ::lib.schema.common/non-blank-string]
-   [:schema         [:maybe ::lib.schema.common/non-blank-string]]
-   [:require-filter {:optional true} :boolean]
+   [:name                                  ::lib.schema.common/non-blank-string]
+   [:schema                                [:maybe ::lib.schema.common/non-blank-string]]
+   [:esetimated-row-count {:optional true} [:maybe :int]]
```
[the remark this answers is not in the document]
qnkhuat: >  when the driver supports :sync-estimated-row-count. 

I meant to remove that feature since we don't have a partial reason for having it now. maybe we need to let third party drivers know about it but it doesn't have much benefit in syncing it now, do you it still worth mentioning in driver log?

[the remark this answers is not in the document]
qnkhuat: also made this schema a closed map to avoid situations like this.

resources/migrations/001_update_migrations.yaml
qnkhuat: good catch! updated [8e78370](https://github.com/metabase/metabase/pull/39494/commits/8e78370cb7edb0b9846a615e587bf9c4a9034635)

test/metabase/driver/postgres_test.clj
qnkhuat: for the `sync-estimated-row-count-test` test I want an accurate number, so executing VACUUM to trigger Postgres to get the exact number. but for these tests we don't need to test the exact number.

test/metabase/driver/postgres_test.clj
```diff
@@ -1396,3 +1415,18 @@
...
                (do-test :table-pattern "table" :tables ["table"])))))))))
+
+(deftest sync-estimated-row-count-test
+  (mt/test-driver :postgres
+    (testing "Can sync row count"
+      (mt/with-temp-test-data [["city"
+                                [{:field-name "name"
+                                  :base-type  :type/Text}]
+                                [["Los Angeles"]
+                                 ["Las Vegas"]]]]
```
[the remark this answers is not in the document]
qnkhuat: good point, use `test-data` and moved it to a more generic namespace here sync-estimated-row-count-test

resources/migrations/001_update_migrations.yaml
qnkhuat: it's now in its own column `metabase_table.estimated_row_count`

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -193,10 +193,10 @@
         (remove metabase-metadata/is-metabase-metadata-table?)
         (:tables db-metadata)))
 
-(mu/defn ^:private db->our-metadata :- [:set i/DatabaseMetadataTable]
+(mu/defn ^:private db->our-metadata :- [:set (ms/InstanceOf :model/Table)]
```
qnkhuat, line 196: the `i/DatabaseMetadataTable` schema is now a closed one, so it was failing because the table from our metadata includes an `:id`. rather than adding an `:id` to `i/DatabaseMetadataTable`, I think it makes more sense that our-metadata has its own schema here. This is how we do in sync fields as well.

src/metabase/sync/interface.clj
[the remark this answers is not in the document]
calherries: As I said on slack I guess we technically don't need a driver feature yet, but when we use the row count we should definitely add the feature to help with testing and optimizations. So up to you. There's not much of a downside if we introduce a driver feature, never use it, then remove it later, so I might just add one now.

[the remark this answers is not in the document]
calherries: > made this schema a closed map

Hmm maybe we want to be careful here, assuming we want to turn on malli schema validation in production again, making this map closed constitutes a breaking change. Driver implementations can currently return whatever keys they want in this map, and if we turn malli schema checking on this will break their sync process. Unlikely, but possible

test/metabase/driver/sql_jdbc_test.clj
```diff
@@ -282,3 +285,18 @@
               (let [syncable (driver/syncable-schemas driver/*driver* db-filtered)]
                 (is (not (contains? syncable "public")))
                 (is (not (contains? syncable fake-schema-name)))))))))))
+
+(deftest sync-estimated-row-count-test
```
calherries, line 289: The previous comment got lost because the code was removed, so here it is
https://github.com/metabase/metabase/pull/39494/files/5e15c627f41d4fa81f89680ab05cae675f745828#r1534344084

I mentioned this should be in a sync test namespace, but this is in a driver test namespace. If we're to take modularization seriously there shouldn't be any sync related code in this namespace.

Maybe a good place to put this test is `metabase.sync.sync-metadata.tables-test`?

It's unfortunate this test uses `(next.jdbc/execute! conn ["VACUUM;"])` which is postgres specific but I think we can worry about that when we get to having multiple drivers support this feature

test/metabase/driver/postgres_test.clj
calherries: got it, thanks

test/metabase/driver/postgres_test.clj
```diff
@@ -1396,3 +1415,18 @@
...
+        (fn [conn]
+          ;; row count is estimated so we VACUUM so the statistic table is updated before syncing
+          (next.jdbc/execute! conn ["VACUUM;"])
```
crisptrutski: I'm a bit worried about this flaking, as there's no guarantee this fixes the statistics. AFAIK `ANALYZE` will be more reliable.

src/metabase/api/notify.clj
```diff
@@ -58,7 +58,7 @@
       (let [driver (driver.u/database->driver database)
             {db-tables :tables} (driver/describe-database driver database)]
         (if-let [table (some (fn [table-in-db]
-                               (when (= (dissoc table-in-db :description)
+                               (when (= (select-keys table-in-db [:schema :name])
                                         {:schema schema_name :name table_name})
```
crisptrutski, line 62: It's slightly shorter and slightly more efficient to just check the individual fields.

```suggestion
                               (when (and (= schema_name (:schema table-in-db)))
                                          (= table_name (:name table-in-db)))
```

src/metabase/sync/interface.clj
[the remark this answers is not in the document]
qnkhuat: both fair points, I guess we can add the feature now, and also open up the schema again

test/metabase/driver/postgres_test.clj
```diff
@@ -251,8 +251,23 @@
...
-  {:name table-name, :schema "public", :description nil})
+(defn- default-table-result
+  ([table-name]
+   (default-table-result table-name {}))
+  ([table-name opts]
+   (merge
+    {:name table-name :schema "public" :description nil
+     ;; estimated-row-count is estimated, so the value can't be known for sure "during" test without
+     ;; VACUUM-ing. So for tests that don't concern the exact value of estimated-row-count, use schema instead
+     :estimated_row_count (mt/malli=? [:maybe :int])}
```
crisptrutski, line 262: nitpick: I prefer 1 key-value per line unless a map is a 1-liner.

test/metabase/driver/postgres_test.clj
```diff
@@ -251,8 +251,23 @@
...
+    opts)))
+
+(defn- describe-database->tables
```
crisptrutski, line 265: The way we typically use this naming schema I'd expect this to either return a map, or to take a "describe-database" as an argument. How about `describe-tables`?

test/metabase/driver/sql_jdbc_test.clj
crisptrutski: Perhaps vacuum (or analyze) could be put behind a test multimethod (something like `update-table-statistics!`)

crisptrutski (commented):
Looks great, just waiting on Cal's call out.

test/metabase/driver/postgres_test.clj
qnkhuat: > AFAIK ANALYZE will be more reliable.
interesting, how so?

on flaking: the test-data db is quite small so vacuuming shouldn't take a lot of time. and also no one modify it, so it should be pretty safe IOM.

test/metabase/driver/sql_jdbc_test.clj
qnkhuat: Moved to metabase.sync.sync-metadata.tables-test

on making vacuum a multimethod: absolutely, but I'll get to it once we add more driver supports :)

calherries (approved):
LGTM 

you've just got an extra bracket at the end of line 61
https://github.com/metabase/metabase/pull/39494/files#diff-d7645fcaa393ee95545bbf8d6f9fc58a52f375e1ffb1dc6c1a6f8d9ee833efd0R61-R62

src/metabase/api/notify.clj
```diff
@@ -58,8 +58,8 @@
       (let [driver (driver.u/database->driver database)
             {db-tables :tables} (driver/describe-database driver database)]
         (if-let [table (some (fn [table-in-db]
-                               (when (= (dissoc table-in-db :description)
-                                        {:schema schema_name :name table_name})
+                               (when (and (= schema_name (:schema table-in-db)))
```
qnkhuat, line 61:
```suggestion
                               (when (and (= schema_name (:schema table-in-db))
```

PR #39502, opened 2024-03-04
Repro #29094 - Notebook filter field allows non boolean custom expressions

### LINKED ISSUE #29094: Notebook filter field allows non boolean custom expressions
**Describe the bug**
Notebook filter field allows non boolean custom expressions

**To Reproduce**

1. Start a new question based on Orders table
2. Add a filter with Custom Expression `[Tax] * 22`
3. Click "Done"

**Expected behavior**
Non-boolean expression is not allowed

kamilmielnik:
Closes #29094

No backport because in v49 it behaves slightly different than in `master`, see [comment](https://github.com/metabase/metabase/issues/29094#issuecomment-1976261054).

Stress test: https://github.com/metabase/metabase/actions/runs/8140145441

### DIFF
--- a/e2e/test/scenarios/filters/reproductions/29094-non-boolean-custom-expressions.cy.spec.js
+++ b/e2e/test/scenarios/filters/reproductions/29094-non-boolean-custom-expressions.cy.spec.js
@@ -6,7 +6,7 @@ import {
   startNewQuestion,
 } from "e2e/support/helpers";
 
-describe("issue 29094", function () {
+describe("issue 29094", () => {
   beforeEach(() => {
     restore();
     cy.signInAsNormalUser();
--- a/e2e/test/scenarios/filters/reproductions/29094-non-boolean-custom-expressions.cy.spec.js
+++ b/e2e/test/scenarios/filters/reproductions/29094-non-boolean-custom-expressions.cy.spec.js
@@ -1,0 +1,35 @@
+import {
+  enterCustomColumnDetails,
+  getNotebookStep,
+  popover,
+  restore,
+  startNewQuestion,
+} from "e2e/support/helpers";
+
+describe("issue 29094", function () {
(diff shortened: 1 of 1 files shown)


nemanjaglumac (approved):
Can you please stress-test this before merging?

e2e/test/scenarios/filters/reproductions/29094-non-boolean-custom-expressions.cy.spec.js
```diff
@@ -0,0 +1,35 @@
+import {
+  enterCustomColumnDetails,
+  getNotebookStep,
+  popover,
+  restore,
+  startNewQuestion,
+} from "e2e/support/helpers";
+
+describe("issue 29094", function () {
```
nemanjaglumac, line 9:
```suggestion
describe("issue 29094", () => {
```

Please change this to an arrow function. @kamilmielnik

nemanjaglumac: It's just a convention we've been following for hundreds of other E2E tests. I mean - we can document it and even enforce the lint rule for it.

Using an arrow function prevents you doing silly things like assigning some value to a variable (`foo`) and then latter trying to access that using `this.foo`.

To sum it up:
There is nothing inherently wrong with using `function` instead of an arrow function, but the latter prevents some less-than-ideal hacks. The rest of the E2E code base is using arrow functions.

kamilmielnik: c7ee4698adb1b77b07776620d982a8149f127289

kamilmielnik: #39512

PR #39505, opened 2024-03-04
Optimize data step in the notebook editor

ranquild:
Changes:
- Columns are computed only when the field picker is opened
- `displayInfo` is used only once per column in `FieldPicker`
- `addField` and `removeField` are used instead of reselecting all fields
- Joins TBD https://github.com/metabase/metabase/pull/39509


kamilmielnik (approved):
:+1:

PR #39509, opened 2024-03-04
Optimize join step

ranquild:
Blocked by https://github.com/metabase/metabase/issues/39514

Changes:
- `addField` and `removeField` are used for updating join fields when the join clause is "complete"
- `joinableColumns` are called only when the picker is opened and not every time the query changes


uladzimirdev (approved):
🤞

PR #39512, opened 2024-03-04
Use arrow functions in e2e & unit tests definitions

kamilmielnik:
See https://github.com/metabase/metabase/pull/39502#discussion_r1511101634


nemanjaglumac (approved):
Thank you

PR #39519, opened 2024-03-04
[WIP]: Less ambiguous loading state

### LINKED ISSUE #39503: [Epic] Make “Doing Science” less ambiguous for long-running queries
- [product doc](https://www.notion.so/metabase/Make-Doing-Science-less-ambiguous-for-long-running-queries-b720f9d9ec144afcbda2edc463f2e596)
- feature branch: `39503-less-ambiguous-loading-state` 
- issues: #31135

**Implementation Plan**

```[tasklist]
### 1. Doing Science
- [ ] https://github.com/metabase/metabase/issues/39615
```

```[tasklist]
### 2. Dashboard cards
- [ ] https://github.com/metabase/metabase/issues/39620
```

romeovs:
Closes https://github.com/metabase/metabase/issues/39503
### Description

Allows the `getLoadingMessage` option to return a function which is parametrised on the elapsed query time.

### How to verify

1. Open a new query that loads very slowly
2. Verify that the loading message changes from `Doing Science...` to `Talking to the database...` after 4 seconds

### Demo

<div>
    <a href="https://www.loom.com/share/61345929d1574e398551fc2350a29fff">
      <p>Question · Metabase - 4 March 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/61345929d1574e398551fc2350a29fff">
      
    </a>
  </div>

### Checklist

### DIFF
--- a/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
@@ -305,6 +305,28 @@ describe("scenarios > visualizations > table", () => {
       expect(isScrollableHorizontally($popover[0])).to.be.false;
     });
   });
+
+  it(
+    "should show the slow loading text when the query is taking too long",
+    { tags: ["@slow"] },
+    () => {
+      openOrdersTable({ mode: "notebook" });
+
+      cy.intercept("POST", "/api/dataset", req => {
+        req.on("response", res => {
+          res.setDelay(6000);
+        });
+      });
+
+      cy.button("Visualize").click();
+
+      cy.findByTestId("query-builder-main").findByText("Doing science...");
+      cy.findByTestId("query-builder-main").findByText(
+        "Talking to the database...",
+        { timeout: 6000 },
+      );
+    },
+  );
 });
 
 describe("scenarios > visualizations > table > conditional formatting", () => {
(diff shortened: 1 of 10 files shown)


ranquild (commented):
I think this can be very expensive to re-render the viz component on every few ms. I think we could change `loadingMessage` to be like `(xxx: boolean) => string` and only re-render when the 4000ms timer times out. With something like this https://github.com/streamich/react-use/blob/master/docs/useTimeout.md

romeovs:
@ranquild the component now rerenders only once every `250ms` (and only when it's in the loading state), which should be quite cheap.

I'm using the [`useElapsed`](https://github.com/metabase/metabase/pull/39519/files#diff-41de5f50ca9a6e6f3da9296d1fe66b11cea97af2690bc8991c495d687e85311eR81-R96) hook to do so.

I had it at 1s before but I thought every 250ms is a nice balance between flexibility and performance.

ranquild:
@romeovs you don't need to rerender it every 250ms.

romeovs:
@ranquild Not for our purposes but we don't know ahead of time what users will set the `getLoadingMessage` function to.

If we want to hardcode the 4s limit I am fine with that but it seems less flexible towards whitelabeling and we'd be introducing a magic number.

Let me know if you just want me to hardcode the loading message timeouts instead of allowing them to be configured in the white label options.

ranquild:
@romeovs yeah currently there are no external users of `getLoadingMessage`, we're free to change it to anything else in the future; so I'd go with the most performant solution now

PR #39531, opened 2024-03-04
Fixes after css modules migration

deniskaber:
Relates to https://github.com/metabase/metabase/issues/38811

### Description

This fixes two issues reported after css modules migration from global to local scope:

image

image

### How to verify

1. Open a dashboard, navigate to a question by clicking on a card title. Back button should be round.

2. Try to create a new model. Cards should have background and shadow.

### DIFF
--- a/frontend/src/metabase/components/List/List.module.css
+++ b/frontend/src/metabase/components/List/List.module.css
@@ -9,11 +9,6 @@
   margin-right: auto;
 }
 
-.list-wrapper {
-  margin-left: auto;
-  margin-right: auto;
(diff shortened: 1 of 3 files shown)


mazameli (approved):
This indeed fixes the two things I pointed out (the circular back button, and the cards behind the two options in the "new model" screen).

Another item I noticed that could be related is that we're missing `cursor: pointer` on hover when hovering on the filter and summarization elements of the notebook editor:

image

uladzimirdev (approved):

PR #39545, opened 2024-03-04
[docs] Fix link in Configuration file docs

alexyarosh:
The link was leading back to the same article instead of article on environment variables


jeff-bruemmer (approved):
Thanks, @alexyarosh!

PR #39559, opened 2024-03-04
docs - update basic exploration

jeff-bruemmer:
Updates the basic exploration page to include the new models browser shipping in 49.

I did some cleanup while I was in there, but it's an ancient page when the docs had a very different shape, so there's a case for breaking the page into multiple pages (e.g., one page for search, one for how drill-through works and how you can set it up, etc). If we tackle that, we should do so in another pull request.


docs/exploration-and-organization/exploration.md
```diff
@@ -98,17 +84,51 @@ Some things to remember with bookmarks:
...
 - To reorder bookmarks, simply drag and drop them in the sidebar.
 
-## Verified items
+## Search
 
-{% include plans-blockquote.html feature="Verification" %}
+![Search results](./images/search-results.png)
 
-Verified questions and models are marked with a blue checkmark icon:
+Use the search bar to find dashboards, questions, collections, subscriptions, and alerts. You can select from the typeahead's dropdown results, or hit enter to view a search results page. You can also activate the search bar from anywhere by pressing the `/` key.
```
abdhabli, line 91: we can't search for Models in the search bar?
also do we have a list of shortcuts of using metabase, it would be nice (if we have a lot of them) to have a page about it

docs/exploration-and-organization/exploration.md
```diff
@@ -6,77 +6,63 @@ redirect_from:
 
 # Basic exploration
 
-## See what your teammates have made
+How to poke around your Metabase.
```
abdhabli, line 9: it would be nice to a small introduction what is the expectation of reading this page, it's full of really important information for users to move around metabase.

abdhabli: Actually this will be a good topic for a video :D

docs/exploration-and-organization/exploration.md
```diff
@@ -6,77 +6,63 @@ redirect_from:
...
+[Collections][collections] in Metabase are a lot like folders. They're where Metabase keeps all your team's dashboards and charts.
 
-If you click on a part of a chart, such as a bar in a bar chart, or a dot on a line chart, you'll see a the **Action menu**, with actions you can take to dive deeper into that result, branch off from it in a different direction, or see automatic insights to [X-ray](x-rays.md) the data.
+![A collection](./images/collection-detail.png)
 
-![Drill-through menu](images/automatic-insights.png)
+If your teammates are cool, they'll have [pinned](./collections.md#pinned-items) some important dashboards or questions within your collections; if so, those pinned items will show up as cards at the top of the collection.
 
-In this example of orders by product category per month, clicking on a dot on this line chart gives us the ability to:
+## Exploring models, questions, and dashboards
```
abdhabli, line 37: what's the difference between exploring models and browsing models? 
I couldn't see anything explaining models in this paragraph

docs/exploration-and-organization/exploration.md
```diff
@@ -98,17 +84,51 @@ Some things to remember with bookmarks:
 - Items that you bookmark will get a boost in your search results (but not the search results of other people).
 - To reorder bookmarks, simply drag and drop them in the sidebar.
 
-## Verified items
+## Search
```
abdhabli, line 87: I agree with you, we can have a separate page about "Search", also might be good idea for a small video.

abdhabli (approved):
Good to go I left a few questions to better understand and got inspired with more ideas for future videos we can create. Thank you Jeff!

docs/exploration-and-organization/exploration.md
jeff-bruemmer: This is an old page (pre-models). Its intent was to show people that you can poke around questions and dashboards as well. 

So: 

- Browsing: Selecting between models. 
- Exploring: drill-through, filtering, summarizing, etc.

PR #39576, opened 2024-03-04
Fail number parsing if it is only a substring

### LINKED ISSUE #38956: BE: coercion of 01/01/2012 to 1 (skippable for release)

crisptrutski:
Closes https://github.com/metabase/metabase/issues/38956

### Description

This fixes an issue where we would silently coerce date strings to integers when parsing them as such.

This change should fix this relaxed coercion for any non-number that incidentally starts with something numeric.

### How to verify

1. Upload a CSV with an integer column
2. Append a CSV with a date in the same column

Previously the append would have succeeded, by treating the first chunk of the date (e.g. day) as the value.

With this change the upload will fail due to type incompatibility.

### DIFF
--- a/src/metabase/upload/parsing.clj
+++ b/src/metabase/upload/parsing.clj
@@ -5,7 +5,7 @@
    [metabase.public-settings :as public-settings]
    [metabase.util.i18n :refer [tru]])
   (:import
-   (java.text NumberFormat)
+   (java.text NumberFormat ParsePosition)
    (java.time LocalDate)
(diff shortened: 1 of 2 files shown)


tsmacdonald (approved):
I'm sold. TIL about `ParsePosition`, but [it checks out](https://download.java.net/java/early_access/panama/docs/api/java.base/java/text/ParsePosition.html).

:boat:

calherries (approved):
Nice

PR #39612, opened 2024-03-05
Use malli to ensure all column types are covered

crisptrutski:
### Description

This is an alternate to using an assert. 

#### Pros
- Know for sure that the check is turned off in production.
- Code reads a bit better.
- Really comprehensive, standardised error message.

#### Cons
- Confuses Cursive (I seem to have to say this a lot)
- ~~Generates an unhelpful exception message without enough context~~ (Thanks @escherize for helping)


calherries (approved):
Nice!

PR #39613, opened 2024-03-05
Convert QP to MLv2 (Part 1)

### LINKED ISSUE #39825: [MLv2 × QP] Migrate pre-processing middleware `metabase.query-processor.middleware.normalize-query`

### LINKED ISSUE #39826: [MLv2 × QP] Migrate pre-processing middleware `metabase.query-processor.middleware.permissions`

### LINKED ISSUE #39827: [MLv2 × QP] Migrate pre-processing middleware `metabase.query-processor.middleware.constraints`

### LINKED ISSUE #39828: [MLv2 × QP] Migrate pre-processing middleware `metabase.query-processor.middleware.validate`

### LINKED ISSUE #39829: [MLv2 × QP] Migrate pre-processing middleware `metabase.query-processor.middleware.fetch-source-query`

camsaul:
Convert the first part of the QP preprocessing middleware to MLv2. This seemed like a good stopping point for PR 1. I will call out important changes in the comments

Resolves #39825
Resolves #39826 
Resolves #39827
Resolves #39828
Resolves #39829


enterprise/backend/src/metabase_enterprise/sandbox/query_processor/middleware/row_level_restrictions.clj
```diff
@@ -23,8 +23,9 @@
    [metabase.permissions.util :as perms.u]
    [metabase.public-settings.premium-features :refer [defenterprise]]
    [metabase.query-processor.error-type :as qp.error-type]
-   [metabase.query-processor.middleware.fetch-source-query
-    :as fetch-source-query]
+   #_{:clj-kondo/ignore [:deprecated-namespace]}
+   [metabase.query-processor.middleware.fetch-source-query-legacy
+    :as fetch-source-query-legacy]
```
camsaul, line 28: I ported the `fetch-source-query` middleware, however a few other pieces of middleware depended on functions there. Instead of having to port all of that stuff too before I'm completely ready, I just copied the old implementation to a new namespace called `fetch-source-query-legacy` for now. Once we port these next pieces of middleware we can remove the copy.

src/metabase/lib/card.cljc
```diff
@@ -116,19 +116,27 @@
 (def ^:private CardColumns
   [:maybe [:sequential {:min 1} CardColumnMetadata]])
 
-(mu/defn ^:private card-metadata-columns :- CardColumns
+(def ^:private ^:dynamic *card-metadata-columns-card-ids*
+  "Used to track the ID of Cards we're resolving columns for, to avoid inifinte recursion for Cards that have circular
+  references between one another."
+  #{})
```
camsaul, line 122: Improvement: MLv2 should not barf when fetching metadata for Cards that have circular references to one another

src/metabase/lib/equality.cljc
```diff
@@ -412,3 +412,21 @@
          (and (not (contains? matching nil))
               (= (count matching) (count columns))
               (every? #(= (count %) 1) (vals matching))))))
+
+(defn clauses-equal-ignoring-lib-uuids?
```
camsaul, line 416: New util function: decide whether two MBQL clauses are equal regardless of their `:lib/uuid`s. This is important for filter deduplication (see below)

src/metabase/lib/filter.cljc
```diff
@@ -293,8 +293,54 @@
...
+                                                         (do
+                                                           (vreset! did-some-flattening? true)
+                                                           args)
+                                                         [filter-clause])))
+                                             filter-clauses)]
+              (if @did-some-flattening?
+                (recur filter-clauses')
+                filter-clauses')))]
+    (cond-> stage
+      (seq (:filters stage)) (update :filters flatten-filters))))
```
camsaul, line 340: New filter helper functions to add filter(s) to a specific stage. We're going to need more stuff like this going forward since QP middleware has to recurse to stages inside joins inside stages. Helper functions to remove duplicate filter clauses and to flatten compound filters e.g. `[1 2 [:and 3 [:and 4 5]]]` becomes `[1 2 3 4 5]`. Filter clauses are actually supposed to always be distinct, this constraint is enforced by the legacy MBQL schema and we need to add it to the new schema as well still

src/metabase/lib/filter.cljc
```diff
@@ -293,8 +293,54 @@
...
+
 (mu/defn filter :- :metabase.lib.schema/query
-  "Sets `boolean-expression` as a filter on `query`."
+  "Sets `boolean-expression` as a filter on `query`. Ignores duplicate filters (ignoring :lib/uuid)."
```
camsaul, line 343: I went ahead and reworked `filter` so it uses the new low-level stuff and thus gets the duplicate filter removal for free

src/metabase/lib/metadata/composed_provider.cljc
```diff
@@ -22,6 +27,44 @@
...
+
+      :else
+      (let [newly-fetched     (if (satisfies? metadata.protocols/BulkMetadataProvider provider)
+                                (metadata.protocols/bulk-metadata provider metadata-type unfetched-ids)
+                                (fetch-bulk-metadata-with-non-bulk-provider provider metadata-type unfetched-ids))
+            newly-fetched-ids (into #{} (map :id) newly-fetched)
+            unfetched-ids     (set/difference unfetched-ids newly-fetched-ids)]
+        (recur more-providers
+               unfetched-ids
+               (into fetched newly-fetched))))))
```
camsaul, line 66: This stuff used to live in the QP store namespace in a special `qp.store/bulk-metadata` function; I'm working on phasing out the QP Store altogether (we can just use MLv2 queries as the metadata-providerable going forward), so I moved this here so the composed providers we use in tests and elsewhere can implement bulk provider functionality.

At some point we might just want to rework the protocols since probably everything should support the BulkProvider methods, and at that point there's no reason to have separate methods like `card`, `field`, `metric`, etc anymore... but that's a project for another day

src/metabase/lib/util.cljc
```diff
@@ -523,3 +523,23 @@
           ;; subvec holds onto references, so create a new vector
           (update :stages (comp #(into [] %) subvec) 0 (inc (canonical-stage-index query stage-number))))
       new-query)))
+
+(defn fresh-uuids
```
camsaul, line 527: Another new util. We need this when expanding legacy Metrics and Segments, because a query might have the same one more than once e.g. if you join the same source query twice.

src/metabase/lib/walk.cljc
```diff
@@ -0,0 +1,112 @@
+(ns metabase.lib.walk
+  "Tools for walking and transforming a query.")
```
camsaul, line 2: This new namespace is probably the best new thing in this PR. I've been working on a version of this for a while now, this is one I'm pretty happy with.

These are tools for walking and replacing things in a pMBQL query at certain points of interest, see `walk` below. Lots of our middleware currently rolls its own version of this sort of code, now we have one that can be used everywhere. Using `mbql.u/replace` is slow and buggy since it recurses into stuff like `:source-metadata` or `:lib/metadata` and gets its grubby paws on `:field_refs`; other times we've ran into bugs where we didn't recurse into joins inside previous stages (source queries) etc.

Now those bugs will hopefully be a thing of the past.

src/metabase/lib/walk.cljc
```diff
@@ -0,0 +1,112 @@
...
+   (fn [query stage-number]
+     (let [path-to-stage (conj (vec path-to-stages) stage-number)]
+       (walk-stage* query path-to-stage f)))
+   query
+   (range (count (get-in query path-to-stages [])))))
+
+(defn- walk-query* [query f]
+  (walk-stages* query [:stages] f))
+
+(defn walk
```
camsaul, line 62: This is the best new tool I added to this PR IMO

src/metabase/query_processor/middleware/expand_macros.clj
```diff
@@ -1,206 +1,193 @@
 (ns metabase.query-processor.middleware.expand-macros
-  "Middleware for expanding `:metric` and `:segment` 'macros' in *unexpanded* MBQL queries.
```
camsaul, line 2: I completely rewrote this middleware, with the MLv2 tooling and `lib.walk` the new code is a lot easier to follow IMO.

src/metabase/query_processor/middleware/fetch_source_query.clj
```diff
@@ -1,40 +1,16 @@
 (ns metabase.query-processor.middleware.fetch-source-query
```
camsaul, line 1: Also rewritten from scratch, it was doing a lot of stuff that wasn't needed. New version using MLv2 tooling is ~150 LoC vs old version which was ~300 LoC.

src/metabase/query_processor/middleware/fetch_source_query_legacy.clj
```diff
@@ -0,0 +1,263 @@
+(ns ^:deprecated metabase.query-processor.middleware.fetch-source-query-legacy
```
camsaul, line 1: This is just the old `fetch-source-query` middleware copied over for the time being since some of its functions are used by other middleware that I haven't ported yet. Pls don't critique this because I didn't write any of this stuff (not recently, at least)

src/metabase/query_processor/middleware/normalize_query.clj
```diff
@@ -1,47 +1,71 @@
 (ns metabase.query-processor.middleware.normalize-query
```
camsaul, line 1: We now convert to pMBQL automatically when processing a query, then we convert back to legacy a little while later.

src/metabase/query_processor/middleware/process_userland_query.clj
```diff
@@ -64,12 +64,12 @@
                                                              (catch Throwable e
                                                                (log/error e (trs "Error saving query execution info"))))))))
 
-(defn- save-successful-query-execution! [cache-details is_sandboxed? query-execution result-rows]
+(defn- save-successful-query-execution! [cache-details is-sandboxed? query-execution result-rows]
```
camsaul, line 67: We need a Kondo linter to keep people from using snake_case parameter names

(except maybe in `defendpoint` because that's how they come in)

src/metabase/query_processor/middleware/results_metadata.clj
```diff
@@ -26,15 +26,16 @@
...
   (try
     ;; At the very least we can skip the Extra DB call to update this Card's metadata results
     ;; if its DB doesn't support nested queries in the first place
     (when (and metadata
                driver/*driver*
                (driver/database-supports? driver/*driver* :nested-queries (lib.metadata/database (qp.store/metadata-provider)))
                card-id
-               (not source-card-id))
+               ;; don't want to update metadata when we use a Card as a source Card.
+               (not (:qp/source-card-id query)))
```
camsaul, line 38: I tried to change the unnamespaced mystery keys added by the `fetch-source-query` middleware to ones prefixed by `:qp`. This makes it a little more obvious that they came from the QP itself and should make it much easier to grep the codebase for them

src/metabase/query_processor/preprocess.clj
```diff
@@ -52,8 +53,11 @@
    #'qp.perms/remove-permissions-key
    #'qp.constraints/maybe-add-default-userland-constraints
    #'validate/validate-query
-   #'fetch-source-query/resolve-card-id-source-tables
+   #'fetch-source-query/resolve-source-cards
    #'expand-macros/expand-macros
+   ;; ↑↑↑ ALL MIDDLEWARE ABOVE THIS POINT WILL SEE MLV2 PMBQL QUERIES ↑↑↑
+   ;; ↓↓↓ ALL MIDDLEWARE BELOW THIS POINT WILL SEE LEGACY MBQL QUERIES ↓↓↓
+   #'qp.convert-to-legacy/convert-to-legacy
```
camsaul, line 60: Check it out everyone, this is how we can convert the QP to MLv2 a few chunks at a time. Each successive PR will just move these three lines farther and farther down the middleware stacc until we don't have any more pre-processing middleware to convert. Then we'll do the same with compilation and post-processing.

I suppose we could actually do multiple conversions back and forth if we want to convert different pieces of middleware independently.

src/metabase/query_processor/store.clj
```diff
@@ -208,14 +209,10 @@
...
-                                      :metadata/column  lib.metadata.protocols/field
-                                      :metadata/metric  lib.metadata.protocols/metric
-                                      :metadata/segment lib.metadata.protocols/segment
-                                      :metadata/table   lib.metadata.protocols/table)]
-                              (for [id ids-set]
-                                (f provider id)))))
+                            (lib.metadata.composed-provider/fetch-bulk-metadata-with-non-bulk-provider
+                             provider
+                             metadata-type
+                             ids-set)))
```
camsaul, line 215: Moved to somewhere more general. The QP Store is basically deprecated, and will be removed when the MLv2 QP project is complete. Don't need a QP Store when the query itself has a metadata provider attached

src/metabase/query_processor/util.clj
```diff
@@ -84,23 +85,41 @@
...
+(defn- remove-lib-uuids
+  "Two queries should be the same even if they have different :lib/uuids, because they might have both been converted
+  from the same legacy query."
+  [x]
+  (walk/postwalk
+   (fn [x]
+     (if (map? x)
+       (dissoc x :lib/uuid)
+       x))
+   x))
```
camsaul, line 97: Query hash is now done on pMBQL queries, I had to remove `:lib/uuid` to not break everything because one queries converted from legacy two times will get two different hashes otherwise.

test/metabase/query_processor/middleware/expand_macros_test.clj
```diff
@@ -11,15 +13,30 @@
...
 
 (defn- mbql-query [inner-query]
-  {:database 1, :type :query, :query (merge {:source-table 1}
-                                            inner-query)})
+  {:database (meta/id)
+   :type     :query
+   :query    (merge {:source-table 1}
+                    inner-query)})
+
+(defn- expand-macros
```
camsaul, line 21: I didn't want to rewrite all of these tests (yet), and I wanted to prove that the new simpler middleware is working the same as the old hairy middleware, so I put a bit of magic in to convert to pMBQL and back to legacy so the old tests could otherwise remain unchanged.

src/metabase/lib/query.cljc
```diff
@@ -84,21 +84,15 @@
...
-
 (mu/defn ^:private query-from-existing :- ::lib.schema/query
+  "Create a pMBQL query from either an existing pMBQL query (attaching metadata provider as needed), or from a legacy MBQL
+  query (converting it to pMBQL)."
   [metadata-providerable :- lib.metadata/MetadataProviderable
    query                 :- lib.util/LegacyOrPMBQLQuery]
-  (let [query (lib.convert/->pMBQL query)]
-    (query-with-stages metadata-providerable (:stages query))))
+  (merge
+   (dissoc query :query :native :type)
```
camsaul, line 93: we were accidentally throwing out important stuff like `:info` before

test/metabase/query_processor/middleware/process_userland_query_test.clj
```diff
@@ -75,7 +80,7 @@
                :cached                 false}
               (process-userland-query query))
           "Result should have query execution info")
-      (is (=? {:hash         "310be80813db561159a218a09fa84afb4bec04ae1f9e1a1d036c944dabcdbdb0"
+      (is (=? {:hash         "840eb7aa2a9935de63366bacbe9d97e978a859e93dc792a0334de60ed52f8e99"
```
camsaul, line 83: query hashes have changed a little bit, hopefully not significantly tho.

src/metabase/query_processor/middleware/fetch_source_query.clj
```diff
@@ -44,262 +20,132 @@
...
+  ;; dependencies if they occur
+  (assert (<= recursion-depth max-recursion-depth)
+          (format "Source Cards not fully resolved after %d iterations." max-recursion-depth))
+  (let [updated-query                              (resolve-source-cards-in-joins original-query dep-graph)
+        {updated-stages :stages, card-id :card-id} (resolve-source-cards-in-stages updated-query (:stages updated-query) dep-graph)
+        ;; `:qp/source-card-id` is used by [[metabase.query-processor.middleware.results-metadata/record-metadata!]] to
+        ;; decide whether to record metadata as well as by the [[add-dataset-info]] post-processing middleware, and
+        ;; by [[metabase.query-processor.middleware.permissions/check-query-permissions*]]
+        updated-query                              (cond-> updated-query
+                                                     updated-stages (assoc :stages updated-stages)
```
snoe, line 128: It would be nice if the new `walk` util had an ability to replace a single stage with multiple, like macro expansion `~@body`

snoe (approved):
Looks good, double check that unresolved-ids can't remain unresolved for ever.

src/metabase/lib/metadata/composed_provider.cljc
```diff
@@ -22,6 +27,44 @@
...
+
+      (not provider)
+      fetched
+
+      :else
+      (let [newly-fetched     (if (satisfies? metadata.protocols/BulkMetadataProvider provider)
+                                (metadata.protocols/bulk-metadata provider metadata-type unfetched-ids)
+                                (fetch-bulk-metadata-with-non-bulk-provider provider metadata-type unfetched-ids))
+            newly-fetched-ids (into #{} (map :id) newly-fetched)
+            unfetched-ids     (set/difference unfetched-ids newly-fetched-ids)]
```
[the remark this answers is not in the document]
camsaul: It only iterates throw all the different providers, if none of them return that Field then it just doesn't come back. The `(not provider)` base case above will catch that.

Right now the bulk provider methods don't check for active-ness either way

src/metabase/query_processor/middleware/convert_to_legacy.clj
```diff
@@ -0,0 +1,12 @@
+(ns metabase.query-processor.middleware.convert-to-legacy
+  (:require
+   [metabase.lib.convert :as lib.convert]))
+
+(defn convert-to-legacy
+  "Middleware that converts and MLv2 query back to a legacy query. This is temporary until we concert the entire QP to
+  use MLv2 everywhere."
+  [query]
+  (letfn [(->legacy-MBQL [query]
+            (lib.convert/->legacy-MBQL query))]
```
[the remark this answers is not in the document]
camsaul: good catch, I had it doing more stuff at first (logging and what not), I'll remove it now that it's doing nothing useful

src/metabase/query_processor/middleware/fetch_source_query.clj
camsaul: That's a good point, that would be useful. I'll see if I can add that

PR #39619, opened 2024-03-05
Split up "Doing science..." message

### LINKED ISSUE #39503: [Epic] Make “Doing Science” less ambiguous for long-running queries

romeovs:
Closes https://github.com/metabase/metabase/issues/39503
### Description

Allows the `getLoadingMessage` option to return a function which is parametrised on the elapsed query time.

### How to verify

1. Open a new query that loads very slowly
2. Verify that the loading message changes from `Doing Science...` to `Talking to the database...` after 4 seconds

### Demo
<div>
    <a href="https://www.loom.com/share/a05eea3db62c47b0b9e2ce010c8d33e6">
      <p>Question · Metabase - 6 March 2024 - Watch Video</p>
    </a>
</div>

### Checklist

### DIFF
--- a/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
@@ -325,6 +325,28 @@ describe("scenarios > visualizations > table", () => {
       expect(isScrollableHorizontally($popover[0])).to.be.false;
     });
   });
+
+  it(
+    "should show the slow loading text when the query is taking too long",
+    { tags: ["@slow"] },
+    () => {
+      openOrdersTable({ mode: "notebook" });
+
+      cy.intercept("POST", "/api/dataset", req => {
+        req.on("response", res => {
+          res.setDelay(6000);
+        });
+      });
+
+      cy.button("Visualize").click();
+
+      cy.findByTestId("query-builder-main").findByText("Doing science...");
+      cy.findByTestId("query-builder-main").findByText(
+        "Talking to the database...",
+        { timeout: 6000 },
+      );
+    },
+  );
 });
 
 describe("scenarios > visualizations > table > conditional formatting", () => {
--- a/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
+++ b/e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
@@ -326,27 +326,24 @@ describe("scenarios > visualizations > table", () => {
     });
   });
 
-  it(
-    "should show the slow loading text when the query is taking too long",
-    { tags: ["@slow"] },
-    () => {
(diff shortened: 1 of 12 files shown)


e2e/test/scenarios/visualizations-tabular/table.cy.spec.js
```diff
@@ -325,6 +325,28 @@ describe("scenarios > visualizations > table", () => {
       expect(isScrollableHorizontally($popover[0])).to.be.false;
     });
   });
+
+  it(
```
ranquild, line 329: there is `cy.clock` which can be used to advance time

frontend/src/metabase/query_builder/components/QueryVisualization.jsx
```diff
@@ -98,22 +83,30 @@ export const VisualizationEmptyState = ({ className }) => (
...
-    </h2>
-  </div>
-);
+export function VisualizationRunningState({ className = "", loadingMessage }) {
+  const [isSlow] = useTimeout(SLOW_MESSAGE_TIMEOUT);
+
+  // show the slower loading message only when the loadingMessage is
+  // not customised
+  const message =
+    loadingMessage === "Doing science..." && isSlow()
```
ranquild, line 92: Few things:
- The message should be translated
- This logic is very easy to break

Wdyt if we do the following:
- Change `getLoadingMessage()` to `getLoadingMessage(isSlow: boolean)`
- Apply the changes in the default plugin
- Apply the changes in the EE plugin. If the setting value (enum value) is still `doing-science`, only then change the message
- Remove `loadingMessage: getWhiteLabeledLoadingMessage(state)` from the QueryBuilder component. Where needed, use the selector directly, e.g. with `useSelector`.

ranquild (changes requested):
Comments inline

frontend/src/metabase/query_builder/components/QueryVisualization.jsx
romeovs: Ok, I've implemented you proposal, please have another look.

ranquild (approved):
LGTM but maybe after removing props drill it would be not hard to make it `getLoadingMessage(state, isSlow)` and not `(state)(isSlow)`. But not sure if it would be cleaner

PR #39633, opened 2024-03-05
Join performance issues

ranquild:
Changes:
- Optimize MLv2 calls. Defer as much as possible. 
- Remove effects and any async logic. 
- Split `JoinStep` into `Join` and `JoinDraft`. Draft version is used for incomplete joins (without the table / left-hand-side / right-hand-side column). Draft joins have different state, and all updates are done differently. **It should be much cleaner and easier to follow than before**. If in doubt, compare with `use-join`, `use-join-condition` and `JoinStep` from `master`. Because of the simplification I could do MLv2 calls in an optional way for each case - complete or draft join/condition.

I'd recommend reviewing from `JoinStep`. In files that are changed heavily the diff is irrelevant. 

How to verify:
- Add a question with a join
- Open the column picker for the joined table
- Start React profiler
- Tick or untick or a column
- `Notebook` should re-render only once

Currently there are 2 rerenders, however the notebook is re-rendered indeed once.
1st:
Screenshot 2024-03-06 at 09 09 25

2nd:
Screenshot 2024-03-06 at 09 08 13


frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinStep.unit.spec.tsx
```diff
@@ -420,7 +420,7 @@ describe("Notebook Editor > Join Step", () => {
 
     userEvent.click(notEqualsOperator);
     await waitFor(() =>
-      expect(screen.queryByTestId("select-list")).not.toBeVisible(),
+      expect(screen.queryByTestId("select-list")).not.toBeInTheDocument(),
```
ranquild, line 423: Because of new popover.

e2e/test/scenarios/joins/joins.cy.spec.js
```diff
@@ -292,11 +292,11 @@ describe("scenarios > question > joined questions", () => {
     // Test LHS column infers RHS column's temporal unit
 
     cy.findByLabelText("Left column").click();
-    popover().findByText("by month").click({ force: true });
-    popover().last().findByText("Week").click();
+    popover().findByText("Created At").click();
```
ranquild, line 295: I believe that was an actual bug. At least the test didn't do what the comment says "Test LHS column infers RHS column's temporal unit". Previously the temporal unit of the RHS column was always overwritten by the unit from the LHS column. Now the unit of the last selected column "wins".

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/Join/Join.tsx
```diff
@@ -0,0 +1,167 @@
...
+      </Box>
+      <JoinConditionCell color={color}>
+        {conditions.map((condition, index) => {
+          const testId = `join-condition-${index}`;
+          const isLast = index === conditions.length - 1;
+
+          return (
+            <Flex key={index} align="center" gap="sm" data-testid={testId}>
+              <JoinCondition
+                key={index}
```
kamilmielnik, line 124:
```suggestion
```

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinConditionColumnPicker/JoinConditionColumnPicker.tsx
```diff
@@ -1,143 +1,190 @@
...
-interface JoinConditionColumnPickerProps
-  extends Omit<QueryColumnPickerProps, "checkIsColumnSelected"> {
-  isNewCondition: boolean;
-  column?: Lib.ColumnMetadata;
-  table?: Lib.Joinable;
-  label?: string;
+interface JoinConditionColumnPickerProps {
+  query: Lib.Query;
+  stageIndex: number;
+  joinable: Lib.Join | Lib.Joinable;
```
kamilmielnik, line 17: We have `JoinOrJoinable` that could be used here and in `JoinColumnDropdownProps`, it just need to be exported.

```suggestion
  joinable: Lib.JoinOrJoinable;
```

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinConditionColumnPicker/JoinConditionColumnPicker.tsx
```diff
@@ -1,143 +1,190 @@
...
+    return Lib.groupColumns(columns);
+  }, [query, stageIndex, joinable, lhsColumn, rhsColumn, isLhsColumn]);
+
+  return (
+    <JoinColumnPicker
+      query={query}
+      columnGroups={columnGroups}
+      stageIndex={stageIndex}
+      hasTemporalBucketing
+      checkIsColumnSelected={checkColumnSelected}
```
kamilmielnik, line 180: nit:
```suggestion
      checkIsColumnSelected={checkIsColumnSelected}
```

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinConditionDraft/JoinConditionDraft.tsx
```diff
@@ -0,0 +1,131 @@
...
+          operator={operator}
+          isReadOnly={isReadOnly}
+          isConditionComplete={false}
+          onChange={setOperator}
+        />
+        <Box mr={!rhsColumn ? "xs" : undefined}>
+          <JoinConditionColumnPicker
+            query={query}
+            stageIndex={stageIndex}
+            joinable={rhsTable}
```
kamilmielnik, line 111: I don't fully understand this: `joinable` is the same (`rhsTable`) for both `JoinConditionColumnPicker` - is this correct?

kamilmielnik (approved):
Wow!

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinConditionDraft/JoinConditionDraft.tsx
ranquild: Fixed, great catch! It should give identical results but it was still incorrect

frontend/src/metabase/query_builder/components/notebook/steps/JoinStep/JoinConditionColumnPicker/JoinConditionColumnPicker.tsx
ranquild: Fixed

PR #39635, opened 2024-03-05
[chain filters] should not be using inactive fields for joins

### LINKED ISSUE #39618: Chain filters do not respect inactive fields
There is a bug in `src/metabase/models/params/chain_filter.clj`  in the definition for `database-fk-relationships*`:  `JOIN` or `WHERE` clause that do not have any conditions to get `metabase_field.active = true` and so it is returning fk relationships for fields that are inactive and breaking things downstream.

[thread](https://metaboat.slack.com/archives/C05MPF0TM3L/p1709188348023799)

sansolovyov:
Resolves #39618


test/metabase/models/params/chain_filter_test.clj
```diff
@@ -670,3 +671,25 @@
...
+        (try
+          (t2/update! :model/Field {:id %messages.receiver_id} {:active false})
+          (testing "check that it switches to sender once receiver is inactive"
+            (memoize/memo-clear! @#'chain-filter/database-fk-relationships)
+            (memoize/memo-clear! @#'chain-filter/find-joins)
+            (is (= [{:lhs {:table $$messages, :field %messages.sender_id}
+                     :rhs {:table $$users, :field %users.id}}]
+                   (#'chain-filter/find-joins (mt/id) $$messages $$users))))
+          (finally
+            (t2/update! :model/Field {:id %messages.receiver_id} {:active true})))))))
```
piranha, line 695: That doesn't look fantastic, but `(t2/with-transaction [... {:rollback-only true}] ...)` did not work for me to stop the changes unfortunately. Anyway, clearing memoization cache doesn't look like it could be done in parallel either.

crisptrutski (approved):
Makes sense to me. A pity about the manual reset, but doesn't seem worth rabbit holing for a small bug fix.

test/metabase/models/params/chain_filter_test.clj
```diff
@@ -670,3 +671,25 @@
...
             (is (= true
                    (:has_more_values (chain-filter venues.latitude {venues.price 4} :limit 1))))))))))
+
+(deftest chain-filter-inactive-test
+  (testing "Inactive fields are not used to generate joins"
+    (mt/dataset avian-singles
+      (mt/$ids nil
+        (testing "receiver_id is active and should be used for the join"
+          (memoize/memo-clear! @#'chain-filter/database-fk-relationships)
+          (memoize/memo-clear! @#'chain-filter/find-joins)
```
crisptrutski, line 681: A `without-memoization` macro that hides these details could be nice. 

Implementation wise it'd be nice if it would be nice for something like this to also avoids populating the cache for any internal calls too, then you'd only need to wrap this whole test once.

To simplify the macro I guess you could pass the list of functions, then it's fairly straightfoward.

```clojure
(without-memoization [#'chain-filter/database-fk-relationships
                      #'chain-filter/find-joins]
  ...)
```

```clojure
(defmacro without-memoization [fns & body]
   `(do (doseq [f fns] (memoize/memo-clear! f))
        (mt/with-dynamic-redefs [~@(map wrap-them-so-they-clear-afterwards fns)]
          ~@body))
```

piranha: Eh, it won't be doing `without-memoization`, so it'll be misleading somewhat. But it now makes me think if just `with-redefs` is going to be easier to test with. :)

crisptrutski: Am I missing something - I thought you were just using the clears to bypass any pre-existing cache?

Where possible try to use `mt/with-dynamic-redefs` instead, it's the threadsafe version 😄

test/metabase/models/params/chain_filter_test.clj
```diff
@@ -670,3 +670,25 @@
           (testing "`true` if the limit is less than the number of values the field has"
             (is (= true
                    (:has_more_values (chain-filter venues.latitude {venues.price 4} :limit 1))))))))))
+
+(deftest chain-filter-inactive-test
+  (testing "Inactive fields are not used to generate joins"
+    (mt/dataset avian-singles
+      (mt/$ids nil
```
crisptrutski, line 677: I see that you can skip putting `nil` here (it is implicit if the macro only gets a single form)

calherries (changes requested):
This PR seems to be fixing a symptom of the issue, not the root cause.

The fact that a field referenced by `fk_target_field_id` is inactive is real issue here. That state should be impossible. I thought I fixed that in [Remove a retired field's FK target references during sync](https://github.com/metabase/metabase/pull/35496). It's possible that there was some edge case that I missed. If so we'll need to do another migration like [this](https://github.com/metabase/metabase/pull/35531).

This fix is unfortunately too narrow because it fixes one symptom of the bug but it means there's probably other ways this becomes an issue out there. Having to check that a field is active every time we use `fk_target_field_id` is error prone and complicates the code everywhere we use this information.

src/metabase/db/connection.clj
```diff
@@ -143,14 +143,17 @@
 
 (def ^:private ^:dynamic *transaction-depth* 0)
 
-(defn- do-transaction [^java.sql.Connection connection f]
+(defn- do-transaction [^java.sql.Connection connection f {:keys [rollback-only]}]
```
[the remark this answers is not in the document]
piranha: It's how it's named in `next.jdbc`, I'd prefer to keep the name. But it seems like tests blew up, so I'm going to revert this.

test/metabase/models/params/chain_filter_test.clj
piranha: No way I'm doing this, I've been burned so much by just adding a second form and then not understanding what's going on 🤣 This is the worst arity dispatching situation ever 🤣

piranha:
@calherries hmm... I'm not understanding semantics completely, but is a field being inactive an error at all times?

test/metabase/models/params/chain_filter_test.clj
crisptrutski: Makes sense, and my condolences - we should probably give the table-nameless version of this function a different name to fix that foot gun.

calherries:
No, but a foreign key reference to an inactive field is definitely a bug, because the constraint doesn't exist in the database. This is a special case of the bug I raised yesterday https://github.com/metabase/metabase/issues/39687 (total coincidence though)

src/metabase/models/params/chain_filter.clj
```diff
@@ -193,14 +193,20 @@
   (let [rows (mdb.query/query {:select    [[:fk-field.id :f1]
```
calherries, line 193: Woah this selecting every FK relationship in the database? Surely this is latency + memory issue for large DBs.

calherries:
Opening this again because I feel we should just land this rather than talking about solutions more.

src/metabase/models/params/chain_filter.clj
```diff
@@ -193,14 +193,20 @@
...
+                                                                                :fk-table.active]
                                            [:metabase_database :database] [:= :fk-table.db_id :database.id]
-                                           [:metabase_field :pk-field]    [:= :fk-field.fk_target_field_id :pk-field.id]]
+                                           [:metabase_field :pk-field]    [:and [:= :fk-field.fk_target_field_id :pk-field.id]
+                                                                                :pk-field.active]]
                                :where     [:and
                                            [:= :database.id database-id]
-                                           [:not= :fk-field.fk_target_field_id nil]]})]
+                                           [:not= :fk-field.fk_target_field_id nil]
+                                           :fk-field.active]
```
calherries, line 207: I'm probably missing some context - why are you not checking that `pk-field` is active too?

calherries: Ah, you're assuming an active field's `fk_target_field_id` will always point to another active field. That's unfortunately not true because of this other bug [If a foreign key constraint is dropped, sync ignores it](https://github.com/metabase/metabase/issues/39687). 

If an active field `fk_target_field_id` targets another primary key field that is dropped or renamed, the primary key field will become inactive and `fk_target_field_id` will still point to it.

calherries: Okay you are checking for `pk-field.active`, just need to test it then

src/metabase/models/params/chain_filter.clj
piranha: 💯

calherries: Created a tech debt issue for this https://github.com/metabase/metabase/issues/39843

calherries (approved):
Looks good

PR #39636, opened 2024-03-05
include clj-memory-meter in :dev alias

sansolovyov:
This adds [clj-memory-meter](https://github.com/clojure-goes-fast/clj-memory-meter) to a regular dev startup; no requires or anything, but this way it'll always be on hand, and it's too useful at times not to include it.

Used like that:

```
user> (require '[metabase.test :as mt] '[clj-memory-meter.core :as mm])
user> (mm/measure (into {} (take 100000 (map vector (repeatedly #(mt/random-name)) (repeat 1000)))))
"8.9 MiB"
```

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -225,6 +225,8 @@
   {:extra-deps
    {com.clojure-goes-fast/clj-async-profiler
                                  {:mvn/version "1.1.1"}                     ; Enables local profiling and heat map generation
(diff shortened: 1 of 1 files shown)


calherries (approved):
sweet

piranha:
Nah, I don't want any milestones so it won't appear in release notes.

PR #39638, opened 2024-03-05
Log warnings when we redact database details

crisptrutski:
Refers https://github.com/metabase/metabase/issues/32822

### Description

Help developers discover that redaction is triggered during JSON encoding, and see where it is happening.

Updated: remove the "log at most once" latch, just made it debug.


qnkhuat (commented):
coolio

qnkhuat (approved):

PR #39650, opened 2024-03-05
Update expression-parts to satisfy filter picker

### LINKED ISSUE #12496: Drill through to question from dashboard doesn't set filter UI properly
**Describe the bug**
When clicking "View these items" from a dashboard, the date filter widget UI is not represented properly on the resulting question. While the results are accurate, the Filter value displayed in the UI is for a specific date as opposed to a range and the calendar widget does not properly display said range.

**Logs**
n/a

**To Reproduce**
1. Build a question with a time-based summarization, e.g. `Count by Created At: Week`
2. Click on a data point and "View these orders"
3. Note that the date filter says "Created at is <a specific date>"
4. Expect that the date filter says "Created at is between <date 1 and date 2>"

**Expected behavior**
Expect the Filter widget to represent the underlying query.

**Screenshots**
- Example dashboard
image

- Resulting filters
image

- Calendar dropdown
image

```json
{
  "browser-info": {
    "language": "en-US",
    "platform": "Linux x86_64",
    "userAgent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Cypress/3.8.2 Chrome/78.0.3904.130 Electron/7.1.7 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "ANSI_X3.4-1968",
    "java.runtime.version": "1.8.0_252-8u252-b09-1~18.04-b09",
    "java.vendor": "Private Build",
    "java.vendor.url": "http://java.oracle.com/",
    "java.version": "1.8.0_252",
    "java.vm.version": "25.252-b09",
    "os.name": "Linux",
    "os.version": "5.3.0-1017-aws",
    "user.language": "en",
    "user.timezone": "US/Pacific"
  },
  "metabase-info": {
    "databases": [
      "h2"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
        "version": "1.4.197 (2018-03-18)"
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
        "version": "1.4.197 (2018-03-18)"
      }
    },
    "run-mode": "prod",
    "version": {
      "src_hash": "77aa5e404c41882abb222122b272efa409323d08",
      "tag": "v0.35.2",
      "date": "2020-05-11",
      "branch": "master",
      "hash": "763fb50"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

**Severity**
Annoying, but the query is correct.

**Additional context**
n/a

lbrdnk:
Closes #12496

### Description

This PR is the backend implementation of "smaller fix" described in [this issue comment](https://github.com/metabase/metabase/issues/12496#issuecomment-1629317661).

`expression-parts` are used on FE to set filter picker for filter expression. Filter expressions that check equality of column with `:temporal-unit` set to value other than `:minute` or `:day` were handled incorrectly. Eg. `[:= {...} [:field {:temporal-unit :week} 111] "2024-05-12"]` would make FE use filter picker set to _on_ even though this equality effectively describes interval. Current filter picker implementation does not support that.

To overcome that, this PR expands the expressions that would require range to be rendered correctly in current filter picker to use `:between` instead of `:=`.

### DIFF
--- a/src/metabase/lib/fe_util.cljc
+++ b/src/metabase/lib/fe_util.cljc
@@ -13,9 +13,8 @@
    [metabase.lib.temporal-bucket :as lib.temporal-bucket]
    [metabase.lib.util :as lib.util]
    [metabase.mbql.util :as mbql.u]
-   [metabase.shared.formatting.internal.date-builder :as shared.formatting.internal.builder]
+   [metabase.shared.formatting.date :as fmt.date]
    [metabase.shared.util.i18n :as i18n]
-   [metabase.shared.util.internal.time-common :as shared.ut.common]
    [metabase.shared.util.time :as shared.ut]
    [metabase.util :as u]
    [metabase.util.malli :as mu]))
@@ -54,10 +53,10 @@
    This functionality is backend approach to \"smaller solution\"."
   [[_operator options column-arg dt-arg :as _expression-clause]]
   (let [temporal-unit (:temporal-unit (lib.options/options column-arg))
-        interval (shared.ut.common/to-range (shared.ut/coerce-to-timestamp dt-arg) {:unit temporal-unit :n 1})
-        format' (cond-> [:year "-" :month-dd "-" :day-of-month-dd]
-                 (contains? expandable-time-units temporal-unit) (into ["T" :hour-24-dd ":" :minute-dd]))
-        formatter (shared.formatting.internal.builder/->formatter format')]
+        interval (shared.ut/to-range (shared.ut/coerce-to-timestamp dt-arg) {:unit temporal-unit :n 1})
+        formatter (if (contains? expandable-time-units temporal-unit)
+                    fmt.date/datetime->iso-string
+                    fmt.date/date->iso-string)]
     (into [:between options column-arg] (map formatter) interval)))
 
 (defn- maybe-expand-temporal-expression
--- a/src/metabase/lib/fe_util.cljc
+++ b/src/metabase/lib/fe_util.cljc
@@ -13,7 +13,9 @@
    [metabase.lib.temporal-bucket :as lib.temporal-bucket]
    [metabase.lib.util :as lib.util]
    [metabase.mbql.util :as mbql.u]
+   [metabase.shared.formatting.internal.date-builder :as shared.formatting.internal.builder]
    [metabase.shared.util.i18n :as i18n]
+   [metabase.shared.util.internal.time-common :as shared.ut.common]
    [metabase.shared.util.time :as shared.ut]
    [metabase.util :as u]
    [metabase.util.malli :as mu]))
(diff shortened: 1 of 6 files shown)


src/metabase/lib/fe_util.cljc
```diff
@@ -13,7 +13,9 @@
    [metabase.lib.temporal-bucket :as lib.temporal-bucket]
    [metabase.lib.util :as lib.util]
    [metabase.mbql.util :as mbql.u]
+   [metabase.shared.formatting.internal.date-builder :as shared.formatting.internal.builder]
```
lbrdnk, line 16: Would it be ok to use `metabase.shared.formatting.internal.date-builder` in this namespace?

src/metabase/lib/fe_util.cljc
```diff
@@ -13,7 +13,9 @@
...
    [metabase.shared.util.i18n :as i18n]
+   [metabase.shared.util.internal.time-common :as shared.ut.common]
```
bshepherdson, line 18: Don't import this - it's internal for a reason, as I said above.

But also the only function used from here (`to-range`) is re-exported from `metabase.shared.util.time`. (See the `(shared.ns/import-fns ...)` macro at the top of that namespace.) So you can call the same function as `shared.ut/to-range`.

src/metabase/lib/fe_util.cljc
```diff
@@ -25,6 +27,45 @@
...
+   To mitigate that expressions are converted to `:between` form which is handled correctly by filter picker. For more
+   info on the issue see the comment [https://github.com/metabase/metabase/issues/12496#issuecomment-1629317661].
+   This functionality is backend approach to \"smaller solution\"."
+  [[_operator options column-arg dt-arg :as _expression-clause]]
+  (let [temporal-unit (:temporal-unit (lib.options/options column-arg))
+        interval (shared.ut.common/to-range (shared.ut/coerce-to-timestamp dt-arg) {:unit temporal-unit :n 1})
+        format' (cond-> [:year "-" :month-dd "-" :day-of-month-dd]
+                 (contains? expandable-time-units temporal-unit) (into ["T" :hour-24-dd ":" :minute-dd]))
+        formatter (shared.formatting.internal.builder/->formatter format')]
+    (into [:between options column-arg] (map formatter) interval)))
```
bshepherdson, line 61: This logic should be captured in two new functions of `metabase.shared.formatting.date`.

- `(date->iso-string dt)` coerces `dt` with `coerce-to-timestamp` and returns eg. `"2024-03-05"`
- `(datetime->iso-string dt)` coerces `dt` and returns a standard ISO8601 `"2024-03-05T15:21:33Z"`

Then the logic here becomes:
- Call `to-range` to get the endpoints
- Choose one of the above functions based on the unit.
- `(into [:between options column-arg] (map f) the-range)`

src/metabase/lib/fe_util.cljc
bshepherdson: No, the point of "internal" is to disallow this kind of thing.

See below about new functions in `metabase.shared.formatting.date` that give you the strings you need behind a *much* simpler interface. The `date-builder` stuff should remain an implementation detail.

bshepherdson (changes requested):

bshepherdson (approved):

PR #39652, opened 2024-03-05
Fix field-vales query to specify :full type

crisptrutski:

### Description

This looks like yet another case which could accidentally return only a subset of the field values (one of the advanced sets).


calherries (commented):
Is this hydration method used? It looks unused and untested, so I would remove it.

calherries (approved):
If you have a good reason to keep it, this LGTM so approving anyway

PR #39664, opened 2024-03-05
Fix hanging cache refresh jobs

### LINKED ISSUE #39138: Model Caching stuck in "pending" state not cleared out by next scheduled refresh
### Describe the bug

Models scheduled to cache may error for various reasons including but not limited to:
- [Caching process interrupted by SIGTERM] (https://github.com/metabase/metabase/issues/30360)
- data upon which model based changes (i.e. schema change on underlying data)

When this happens, caching status stuck in "pending" status and not fixed by next refresh

### To Reproduce

See https://metabase.zendesk.com/agent/tickets/25109

### Expected behavior

Expect errors in models to be handled allowing process to complete and "pending" status to be fixed at next refresh

### Logs

Requesting logs from affected Cloud customers.

```JSON
- Metabase v1.48.6
```

### Severity

P1: significant performance degradation with no indication or simple workaround

### Additional context

From customer experiencing:
- `all of this extra work when the model isn't cached puts a huge amount of additional load on our redshift DB in ways that caching is designed precisely to solve for. With caching we can control exactly when data is cached, so that it runs in sequences and often after hours. Because model caching fails transparently - nothing then stops the queries from running when people try to access the data, which then creates the unpredictable load on our DB. This has resulted in a huge amount of wasted time from everyone affected - operations, devops, and of course practitioners using the data in whatever way it's meant to be used.`

Related to:
- https://github.com/metabase/metabase/issues/30360
- https://github.com/metabase/metabase/issues/29689

johnswanson:
I started a much more complex fix with a job that tried to clean up stale `refreshing` PersistedInfos... then talked to Dan, who pointed out that because only one instance is running these jobs at any given time, we can just add `refreshing` to the list of refreshable states.

Will fix https://github.com/metabase/metabase/issues/39138

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/advanced_config/caching.clj
+++ b/enterprise/backend/src/metabase_enterprise/advanced_config/caching.clj
@@ -16,7 +16,7 @@
   "States of `persisted_info` records which can be refreshed."
   :feature :cache-granular-controls
   []
-  #{"creating" "persisted" "error"})
+  #{"refreshing" "creating" "persisted" "error"})
 
 (defenterprise prunable-states
   "States of `persisted_info` records which can be pruned."
--- a/src/metabase/models/persisted_info.clj
+++ b/src/metabase/models/persisted_info.clj
@@ -85,7 +85,7 @@
    refreshable with the feature flag disabled."
   metabase-enterprise.advanced-config.caching
   []
-  #{"creating" "persisted" "error" "off"})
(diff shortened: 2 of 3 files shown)


noahmoss (commented):
should there be a test for this?

johnswanson:
> should there be a test for this?

Good idea, just added one!

noahmoss (approved):

PR #39666, opened 2024-03-05
Use timestamp in attachment filenames

### LINKED ISSUE #39120: Add the timestamp to the filename in the subscription attachment
We should make all exports have the same naming convention: question_name_timestamp.format. Right now normal exports have this, while the attachments on the subscriptions do not 
 
Add the timestamp to the subscription attachments

None

Requested by a lead

**Additional context**
NA

adam-james-v:
Fixes: #39120

This modifies the subscriptin attachment filenames to match the same format used when doing a direct download from a dashboard or question.

That is:
 - card names are slugified (spaces and dashes turned to underscores, etc.)
 - a timestamp is appended to the slugified card name


test/metabase/pulse/pulse_integration_test.clj
```diff
@@ -166,6 +166,17 @@
...
+(defn- strip-timestamp
+  "Remove the timestamp portion of attachment filenames.
+  This is useful for creating stable filename keys in tests.
+  For example, see `run-pulse-and-return-attached-csv-data` below.
+
+  Eg. test_card_2024-03-05T22:30:24.077306Z.csv -> test_card.csv"
+  [fname]
+  (let [ext (last (str/split fname #"\."))
+        name-parts (butlast (str/split fname #"_"))]
+    (format "%s.%s" (str/join "_" name-parts) ext)))
```
markbastian, line 178: Will a date ever have an _ in it? Assuming no. Just want to make sure this is a good splitter. This is probably good, though.

markbastian (approved):
See one comment. Non-blocking. LGTM.

test/metabase/pulse/pulse_integration_test.clj
adam-james-v: Hmm... yea good point. I think it's fine though. I guess if some tests fail in the future this can be adjusted as needed.

Issue #39411, opened 2024-02-29
[dc.js migration] waterfall chart does not render series bars when there are many of them


EmmadUsmani:
| dc.js | echarts |
|--------|--------|
| <img width="1897" alt="Screenshot 2024-02-29 at 11 45 37 AM" src="https://github.com/metabase/metabase/assets/37751258/9b73c9e6-a4b4-4e72-bd12-0cd26fafca85"> | <img width="1890" alt="Screenshot 2024-02-29 at 11 45 22 AM" src="https://github.com/metabase/metabase/assets/37751258/21eb5574-5bc0-444a-b7dc-56d124d4b2d5"> |

alxnddr:
I assume this may be the limitation of candlestick series, maybe we will need to switch to custom series

PR #39669, opened 2024-03-06
fix waterfall series is missing on enormous datasets

alxnddr:
Closes https://github.com/metabase/metabase/issues/39411

### Description

Candlestick series hides elements when they become too small, so this PR replaces it with `custom` series for waterfall charts.

### How to verify

Create a waterfall chart with a small enough bar width like Orders Count by Created At: Day and ensure it shows the waterfall series.

### Demo

Screenshot 2024-03-05 at 9 54 48 PM

### Checklist


EmmadUsmani:
https://github.com/metabase/metabase/assets/37751258/a0ba4704-fcae-479a-8def-c97eab6ec80d

Not necessarily related to this PR but I noticed timescale has different spacing than the ordinal scale for the same number of bars

EmmadUsmani (approved):

alxnddr:
> Not necessarily related to this PR but I noticed timescale has different spacing than the ordinal scale for the same number of bars

The spacing of these axes types is inherently different and it was different in the previous implementation. I think this is not worth even creating a P3

PR #39670, opened 2024-03-06
Homebrew getTables for postgres

qnkhuat:
Implementing our own DatabaseMetadata.getTables in SQL, currently, it'll return exactly the same as the jdbc method. but in the future we can add more fields to it; one thing I'll add after this is row count for the deep fingerprint project https://github.com/metabase/metabase/issues/38229

Kudos to @calherries  for coming up with this idea!


src/metabase/driver/postgres.clj
```diff
@@ -880,6 +883,137 @@
           "from table_privileges t"]))
        (filter #(or (:select %) (:update %) (:delete %) (:update %)))))
 
+(def ^:private table-type-clauses
+  {"TABLE"              [:and
```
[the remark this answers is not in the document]
calherries: As someone who has also been doing a bit of SQL -> HoneySQL translation myself lately, I've come to agree. I imagined there might be some point in the future where we want to do more sophisticated string manipulation, and HoneySQL can help. But translating SQL to HoneySQL is mostly a waste of time in this type of scenario. Since this has been done already though there's no need to undo it. The `:inline` repetition doesn't bother me too much on these lines since it's all neatly aligned.
https://github.com/metabase/metabase/pull/39670/files#diff-122d8f1486c5b97fd092fb98e7419b767c12d378bc6e668c16b1041cd965f9cbR971-R1000

I do think `table-type-clauses` is unnecessary though, and can be deduped with the logic on lines 971-1000. 

The only types we care about are these: `"TABLE" "PARTITIONED TABLE" "VIEW" "FOREIGN TABLE" "MATERIALIZED VIEW" "EXTERNAL TABLE" "DYNAMIC_TABLE"`

I think we just need to excluding system tables with 
`[:<> :n.nspname [:inline "information_schema"]]` and `[(keyword "!~") :n.nspname [:inline "^pg_"]]`.

and then use CTEs to filter on `:type`, generated from this part of the case expression https://github.com/metabase/metabase/pull/39670/files#diff-122d8f1486c5b97fd092fb98e7419b767c12d378bc6e668c16b1041cd965f9cbR991-R1001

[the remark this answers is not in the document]
qnkhuat: yea I guess I was too focused on converting them. simplified and use raw SQL for types.

calherries (commented):
I left a few nit-like comments but the more I look at the code I start to wonder this: the call stack under `driver/describe-database` is getting seriously complicated for something conceptually so simple. You may think this is necessary to keep us from duplicating code but I think it has serious downsides. The complexity around sync is arguably the root cause of the performance problems and what makes it so hard to improve.

Here's my rough breakdown of the call-stack from `describe-database` to `get-tables`:

```
metabase.driver/describe-database has a default sql-jdbc implementation which:
  - optionally gets the database from an ID or spec (but we never use feature this AFAICT, we always pass in a db)
  - gets the inclusion and exclusion patterns from the db
  - calls sql-jdbc.sync/active-tables
sql-jdbc.sync/active-tables calls sql-jdbc.sync.describe-database/fast-active-tables (I think this is unnecessary though, it could be inlined)
sql-jdbc.sync/fast-active-tables calls
  - sql-jdbc.sync.interface/filtered-syncable-schemas
  - sql-jdbc.sync/have-select-privilege-fn
    (effectively filtering tables by schema and select privileges)
  - sql-jdbc.sync/db-tables
sql-jdbc.sync/db-tables calls sql-jdbc.sync.interface/get-tables
sql-jdbc.sync.interface/get-tables finally has a postgres implementation that calls get-table-sql
```

I think all this indirection adds more complication than is necessary. The only logic that is shared between drivers that use the default implementation is the part that gets inclusion and exclusion patterns from the database instance. The rest of the code feels like it's just connecting multimethods in trivial ways.

Cutting through all the noise of the multimethods, describe-database's default SQL-JDBC implementation basically does the following:
1. optionally gets a list of filtered-syncable-schemas (the method by which we do this depends on the driver)
2. gets a list of tables (the method depends on the driver, especially since we are moving away from JDBC)
3. filters tables by select privileges (the method now depends on the driver since we are moving away from select probes)

We need to merge 2 and 3 into one query so we don't need to hold all that information in memory. And we could even merge the driver-specific part of 1 with 2 and 3 into one query in the future. That would again remove some queries.

Each of these merges will require even more multimethods and driver features, making performance issues even harder to find without instrumentation.

To conclude, I'm not convinced adding another multimethod to the mix is doing us any favours over scrapping the whole hierarchy of default implementations and overrides and just implementing `driver/describe-database` for postgres from scratch. We already do this for the majority of other drivers. The default `driver/describe-database` is an overcomplicated mess already and it would be great if we could avoid it.

thoughts?

src/metabase/driver/sql_jdbc/sync.clj
```diff
@@ -24,6 +24,7 @@
   excluded-schemas
   fallback-metadata-query
   filtered-syncable-schemas
+  get-tables
```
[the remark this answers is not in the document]
qnkhuat: haha yea I did find this and the interface confusing, I wanted to do a bit of refactoring to remove one of them since it makes grepping code hard for these methods.

src/metabase/sync/sync_metadata/tables.clj
```diff
@@ -129,8 +129,8 @@
   "Create `new-tables` for database, or if they already exist, mark them as active."
   [database :- i/DatabaseInstance
    new-tables :- [:set i/DatabaseMetadataTable]]
-  (log/info "Found new tables:"
-            (for [table new-tables]
+  (doseq [table new-tables]
```
qnkhuat, line 132: I hate it when it finds 1k tables, and the log is just crazy.

qnkhuat:
yea, I did feel a bit itchy adding a new method like that; defining a multimethod `describe-database` for Postgres does look a lot cleaner! Please take another look.

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -717,27 +717,22 @@
...
-                                       (catch Throwable e
-                                         (throw (ex-info (tru "Error executing query: {0}" (ex-message e))
-                                                         {:driver driver
-                                                          :sql    (str/split-lines (driver/prettify-native-form driver sql))
-                                                          :params params}
-                                                         e))))]
-             ;; TODO - we should probably be using [[reducible-rows]] instead to convert to the correct types
-             (reduce rf init (jdbc/reducible-result-set rs {})))))))))
+  [driver conn [sql & params]]
+  (reify clojure.lang.IReduceInit
```
qnkhuat, line 721: taking a conn instead of db is more generic and helpful IMO.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -32,6 +32,23 @@
    (fn [^ResultSet rs]
      #(.getString rs "TABLE_SCHEM"))))
 
+(defn syncable-schemas-for-db
```
qnkhuat, line 35: I placed it here so we can further re-use it. I can't reuse it in the default imp of `describe-database` because `active-tables` does not take a `db`.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -32,6 +32,23 @@
...
+  "Returns a collection of syncable schemas for a db if exists."
+  [database]
+  (let [driver                 (:engine database)
+        schema-filter-prop     (driver.u/find-schema-filters-prop driver)
+        schema-filter-patterns (when (and (some? schema-filter-prop) (some? database))
+                                 (driver.s/db-details->schema-filter-patterns (:name schema-filter-prop) database))]
```
calherries, line 41: This looks like these lines duplicate the 1-arity version of `metabase.driver.sync/db-details->schema-filter-patterns`

calherries: Actually I have already created a multimethod for this entire function. 
`driver/syncable-schemas`
see https://github.com/metabase/metabase/blob/f0b764e092b1212f186880dd5284842a5c8a1529/src/metabase/driver/sql_jdbc.clj#L188-L197

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -92,27 +109,33 @@
          (.rollback conn))
        false))))
 
+(defn- jdbc-get-tables
```
calherries, line 112: I'm not sure why this function needs to exist separately from db-tables anymore. I think it's a relic of the past version where you wanted its signature to match `get-tables`?

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,94 @@
...
+    (seq tables)
+    (sql.helpers/where [:in :c.relname tables])
+
+    (seq types)
+    (sql.helpers/where (into [:or] (map #(get table-type-clauses %) types))))
+   {:dialect :ansi}))
+
+(defn- get-tables
+  ;; have it as its own method for ease of testing
+  [conn schemas tables types]
```
calherries, line 286: `tables` is used in this namespace to mean table instances, not table names. perhaps we should disambiguate them to be clear
```suggestion
  [conn schemas table-names types]
```

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,94 @@
...
+   (sql-jdbc.execute/do-with-connection-with-options
+    driver
+    database
+    nil
+    (fn [^Connection conn]
+      (if-let [syncable-schemas (seq (sql-jdbc.describe-database/syncable-schemas-for-db database))]
+        (let [have-select-privilege? (sql-jdbc.describe-database/have-select-privilege-fn :postgres conn)]
+          (into #{} (comp (filter have-select-privilege?) (map #(dissoc % :type)))
+                (get-tables conn syncable-schemas nil
+                            ["TABLE" "PARTITIONED TABLE" "VIEW" "FOREIGN TABLE" "MATERIALIZED VIEW"])))
```
calherries, line 301: I feel like the types are unneccessary argument to this function. This doesn't seem like it makes this code any more future-proof to me; I can't imagine why we would call `get-tables` with another set of types. Wdyt about removing the types parameter?

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,94 @@
...
+                            [:inline "i"] [:inline "SYSTEM TOAST INDEX"]
+                            :else nil]
+
+                           [:case-expr :c.relkind
+                            [:inline "r"] [:inline "TEMPORARY TABLE"]
+                            [:inline "p"] [:inline "TEMPORARY TABLE"]
+                            [:inline "i"] [:inline "TEMPORARY INDEX"]
+                            [:inline "s"] [:inline "TEMPORARY SEQUENCE"]
+                            [:inline "v"] [:inline "TEMPORARY VIEW"]
+                            :else nil]]
```
calherries, line 250: We don't care about all these table types, so we should just filter them out instead of making this case expression longer.

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,94 @@
...
+
+                          false ;; non system tables
+                          [:case-expr :c.relkind
+                           [:inline "r"] [:inline "TABLE"]
+                           [:inline "p"] [:inline "PARTITIONED TABLE"]
+                           [:inline "i"] [:inline "INDEX"]
+                           [:inline "P"] [:inline "PARTITIONED INDEX"]
+                           [:inline "S"] [:inline "SEQUENCE"]
```
calherries, line 258: Again, I think we can make this query much smaller if we only kept the code relevant to the types we actually use
```suggestion
```

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,94 @@
...
+   "FOREIGN TABLE"      [:raw "c.relkind = 'f'"]
+   "MATERIALIZED VIEW"  [:raw "c.relkind = 'm'"]})
+
+(defn- get-tables-sql
+  [schemas tables types]
+  ;; Should track this implementation https://github.com/davecramer/pgjdbc/blob/a714bfd/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L1272
+  (sql/format
+   (cond-> {:select    [[:n.nspname :schema]
+                        [:c.relname :name]
+                        [[:case-expr [:or [(keyword "~") :n.nspname "^pg_"] [:= :n.nspname "information_schema"]]
```
calherries, line 229: You could use this `or` clause should be used to filter out system tables, because we don't want their info anyway.

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -158,9 +181,10 @@
                                                                                      schema-inclusion-filters schema-exclusion-filters)
         have-select-privilege-fn? (have-select-privilege-fn driver conn)]
     (eduction (mapcat (fn [schema]
-                        (->> (db-tables driver metadata schema db-name-or-nil)
-                             (filter have-select-privilege-fn?)
-                             (map #(dissoc % :type))))) syncable-schemas)))
+                        (eduction
+                         (comp (filter have-select-privilege-fn?) (map #(dissoc % :type)))
```
calherries, line 185: nit: filter and map steps are usually on separate lines
```suggestion
                         (comp (filter have-select-privilege-fn?) 
                               (map #(dissoc % :type)))
```

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -203,18 +227,9 @@
...
-                  [inclusion-patterns exclusion-patterns] (driver.s/db-details->schema-filter-patterns
-                                                           prop-nm
-                                                           database)]
-              (into #{} (sql-jdbc.sync.interface/active-tables driver conn inclusion-patterns exclusion-patterns)))
-            (default-active-tbl-fn))
-          (default-active-tbl-fn)))))})
+      (let [schema-filter-prop   (driver.u/find-schema-filters-prop driver)
+            database             (db-or-id-or-spec->database db-or-id-or-spec)
+            [inclusion-patterns
+             exclusion-patterns] (when (and (some? schema-filter-prop) (some? database))
```
calherries, line 233: I'm not sure how database can be `nil`. Maybe you've traced this back up the call stack?
```suggestion
             exclusion-patterns] (when (and (some? schema-filter-prop))
```

test/metabase/driver/postgres_test.clj
```diff
@@ -1356,3 +1357,42 @@
     ;; Special characters
     (is (= "SET ROLE \"Role.123\";"   (driver.sql/set-role-statement :postgres "Role.123")))
     (is (= "SET ROLE \"$role\";"      (driver.sql/set-role-statement :postgres "$role")))))
+
+(deftest get-tables-parity-with-jdbc-test
```
calherries, line 1361: Nice test. Is this going to be removed afterwards, or do you think it's worth keeping around? It will make it harder to extend `get-tables` in the future to have more features, because this test will fail.

test/metabase/driver/postgres_test.clj
```diff
@@ -1356,3 +1357,42 @@
...
+                           "CREATE TABLE private.table (id INTEGER);"]]
+               (next.jdbc/execute! conn [stmt]))
+             (testing "without any filters"
+               (do-test))
+             (testing "filter by schema"
+               (do-test :schema-pattern "private" :schemas ["private"]))
+             (testing "filter by table name"
+               (do-test :table-pattern "table" :tables ["table"]))
+             (testing "filter by type"
+               (do-test :types ["TABLE" "PARTITIONED TABLE" "VIEW" "FOREIGN TABLE" "MATERIALIZED VIEW"])))))))))
```
calherries, line 1398: I don't think `get-tables` needs to have the type parameter, so we can remove this test.

calherries:
Much better! With this new change I think a lot more code can still be removed.

camsaul (dismissed):
- I think Cal is on the right track when he pointed out that the way `describe-database` normally works is a way more complicated than it needs to be, why don't you just implement `describe-database` instead so you have one method implementation instead of like 10
- Do not change public function signatures in places like `metabase.driver` without the usual three-cycle deprecation process and notice in the driver changelog. Otherwise we might break third-party drivers.
- Cal is also right about the performance being bad if we have a lot of tables that he mentioned in the other PR. If we're going to go down the route of doing `describe-database` with a custom implementation, I think we should figure out how to address the performance issues while we're at it. Either paging this or doing some sort of streaming makes sense. It might be appropriate to introduce a new driver method and new sync logic for the new and improved `paged-describe-database` or `reduce-database-tables` or whatever we're going to call it. 

Slightly off-topic but I was also thinking it might make more sense to store our big hairy SQL queries we use for sync and stuff in actual `.sql` files, something like `src/metabase/driver/postgres/describe-database.sql` would make sense, and then you can `slurp` it with `io/resource` whenever you need it. Editing and looking at big SQL queries in a SQL file is a lot nicer than doing it embedded in a string. 

Templatizing it is a little trickier but we could do something simple like `str/replace` with Mustache-style tags like `{{table-name}}` or maybe even look at using HugSQL to do it for us

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -717,27 +717,22 @@
   "Returns a reducible collection of rows as maps from `db` and a given SQL query. This is similar to [[jdbc/reducible-query]] but reuses the
   driver-specific configuration for the Connection and Statement/PreparedStatement. This is slightly different from [[execute-reducible-query]]
   in that it is not intended to be used as part of middleware. Keywordizes column names. "
-  [db [sql & params]]
```
camsaul, line 720: You can't change the signatures of public driver functions like this out of the blue. We need to give our third-party driver authors notice and go thru the usual 3-cycle deprecation process. I don't even see so much as a mention in the driver changelog

src/metabase/driver/sql_jdbc/sync/describe_database.clj
```diff
@@ -92,27 +109,33 @@
...
+             :description (when-let [remarks (.getString rset "REMARKS")]
+                            (when-not (str/blank? remarks)
+                              remarks))
+             :type        (.getString rset "TABLE_TYPE")}))))
+
 (defn db-tables
   "Fetch a JDBC Metadata ResultSet of tables in the DB, optionally limited to ones belonging to a given
   schema. Returns a reducible sequence of results."
-  [driver ^DatabaseMetaData metadata ^String schema-or-nil ^String db-name-or-nil]
+  [driver ^Connection conn ^String schema-or-nil ^String db-name-or-nil]
```
camsaul, line 130: This one too, this can break third-party drivers

src/metabase/driver/sql_jdbc/sync/describe_database.clj
qnkhuat: oh nice!

src/metabase/driver/sql_jdbc/sync/describe_database.clj
qnkhuat: hmm I thought I had a note for it some where, but I left it as its own function for testing purposes, see the test I have in `test/metabase/driver/postgres_test.clj`

calherries: Yeah what I'm not understanding is why test `jdbc-get-tables` when you can test `db-tables`? I thought you did this because originally `get-tables` had a `types` argument, but now it doesn't.

What we actually want to test is that `postgres/get-tables` returns the same thing as `db-tables`, because `fast-active-tables` uses `db-tables`. `jdbc-get-tables` feels more like an implementation detail. The fact that you've pasted the exact list of types that postgres needs to return [here](https://github.com/metabase/metabase/pull/39670/commits/b5b9b8a6f7cb305afe85b25e251889d441a6a838#diff-821f0be02cebbea5f2f5b86239bd9df5e5cb4f671634a761fa6da1c511487125R1378) effectively makes that test equivalent to testing `db-tables` anyway. But it would be clearer if you just used `db-tables` instead.

test/metabase/driver/postgres_test.clj
qnkhuat: I think it's nice to keep it. for future use cases, we are probably going to add more fields to table metadata; in those cases, we can dissoc the new keys for this test

test/metabase/driver/postgres_test.clj
qnkhuat: removed types in https://github.com/metabase/metabase/pull/39670/commits/b5b9b8a6f7cb305afe85b25e251889d441a6a838

src/metabase/driver/postgres.clj
qnkhuat: did that in https://github.com/metabase/metabase/pull/39670/commits/b5b9b8a6f7cb305afe85b25e251889d441a6a838

src/metabase/driver/postgres.clj
qnkhuat: removed in https://github.com/metabase/metabase/pull/39670/commits/b5b9b8a6f7cb305afe85b25e251889d441a6a838

qnkhuat:
@camsaul I've reverted the arity changes in driver public functions.
On making this more memory efficient: while that's a great idea, I think this can be shipped on its own. The get-tables implementation for Postgres already returns a reducible, so it's a good first start toward having a `reducible-describe-database`. I didn't want to make it into this PR because it involves changing how we sync tables. I've looked into it and it's not trivial to do that in a reducing way, maybe there is a way but I'll need to spend time on it.

on SQL templating, I'm not very fond of it for this use case because we want to able to conditionally construct where clauses here, and also, the query is quite small, and honeysql is a good fit for it IMO.

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,58 @@
...
+            :order-by  [:type :schema :name]}
+     (seq schemas)
+     (sql.helpers/where [:in :n.nspname schemas])
+
+     (seq table-names)
+     (sql.helpers/where [:in :c.relname table-names]))
+   {:dialect :ansi}))
+
+(defn- get-tables
+  ;; have it as its own method for ease of testing
```
calherries, line 250: Good idea leaving this comment here. Will help anyone thinking of "cleaning it up" in the future not to try it 👍

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,58 @@
...
+  {:tables
+   (sql-jdbc.execute/do-with-connection-with-options
+    driver
+    database
+    nil
+    (fn [^Connection conn]
+      (if-let [syncable-schemas (seq (driver/syncable-schemas driver database))]
+        (let [have-select-privilege? (sql-jdbc.describe-database/have-select-privilege-fn :postgres conn)]
+          (into #{} (comp (filter have-select-privilege?) (map #(dissoc % :type)))
+                (get-tables database syncable-schemas nil)))
```
calherries, line 265: I think this is a little denser than idiomatic styling, which would have the `filter` and `map` steps on separate lines, similar to a `->>` expression:
```suggestion
          (into #{} 
                (comp (filter have-select-privilege?)
                      (map #(dissoc % :type)))
                (get-tables database syncable-schemas nil)))
```

src/metabase/driver/sql_jdbc/sync/describe_database.clj
calherries: cool, that makes sense

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,58 @@
...
+
+(defn- get-tables
+  ;; have it as its own method for ease of testing
+  [database schemas tables]
+  (sql-jdbc.execute/reducible-query database (get-tables-sql schemas tables)))
+
+(defmethod driver/describe-database :postgres
+  [driver database]
+  {:tables
+   (sql-jdbc.execute/do-with-connection-with-options
```
calherries, line 257: As Cam suggested, in the future it would be great to have a new multimethod that returned a reducible of results so we don't have keep all the tables in memory. While we're writing this code, maybe we should make that change as easy as possible.

I don't think we need to create the new multimethod yet, but we can prepare to switch in the future. I'd suggest a simple refactor here, creating a new function that returns a reducible collection of table metadata, and have `driver/describe-database` use it.

e.g.

```clojure
(defmethod driver/describe-database :postgres
  [driver database]
  ;; TODO: convert describe-syncable-tables into its own multimethod
  {:tables (into #{} (describe-syncable-tables driver database))
```

Then when we come to create the new multimethod, we can simply replace `(defn describe-syncable-tables ..)` with `(defmulti describe-syncable-tables ...)` or whatever we want to call it

src/metabase/driver/sql_jdbc/execute.clj
```diff
@@ -728,18 +729,17 @@
...
                        ^ResultSet rs (try
-                                       (let [max-rows 0] ; 0 means no limit
-                                         (execute-statement-or-prepared-statement! driver stmt max-rows params sql))
-                                       (catch Throwable e
-                                         (throw (ex-info (tru "Error executing query: {0}" (ex-message e))
-                                                         {:driver driver
-                                                          :sql    (str/split-lines (driver/prettify-native-form driver sql))
-                                                          :params params}
-                                                         e))))]
+                                      (let [max-rows 0] ; 0 means no limit
```
calherries, line 732: whitespace was correct before

calherries (approved):
Nicely done. Left a couple of minor suggestions

src/metabase/driver/sql_jdbc/execute.clj
qnkhuat: [a84ee86](https://github.com/metabase/metabase/pull/39670/commits/a84ee8604d75e92f1a180d429114cbe61ec6ff1d)

src/metabase/driver/postgres.clj
qnkhuat: good idea 👍 [a84ee86](https://github.com/metabase/metabase/pull/39670/commits/a84ee8604d75e92f1a180d429114cbe61ec6ff1d)

src/metabase/driver/postgres.clj
qnkhuat: [a84ee86](https://github.com/metabase/metabase/pull/39670/commits/a84ee8604d75e92f1a180d429114cbe61ec6ff1d)

calherries:
@qnkhuat the postgres failures in CI are legit. I think you have to wrap the `sql-jdbc.execute/do-with-connection-with-options` call in a `(reify clojure.lang.IReduceInit ...)` like this https://github.com/metabase/metabase/blob/1a4b2dd84bd079ddc4e5d71265b848da805afde3/src/metabase/driver/sql_jdbc/execute.clj#L722-L724

src/metabase/driver/postgres.clj
```diff
@@ -209,6 +213,64 @@
...
+
+(defn- describe-syncable-tables
+  [{driver :engine :as database}]
+  (sql-jdbc.execute/do-with-connection-with-options
+   driver
+   database
+   nil
+   (fn [^Connection conn]
+     (when-let [syncable-schemas (seq (driver/syncable-schemas :postgres database))]
+       (let [have-select-privilege? (sql-jdbc.describe-database/have-select-privilege-fn :postgres conn)]
```
calherries, line 262:
```suggestion
     (when-let [syncable-schemas (seq (driver/syncable-schemas driver database))]
       (let [have-select-privilege? (sql-jdbc.describe-database/have-select-privilege-fn driver conn)]
```

calherries (changes requested):
I just realized redshift will use this implementation for postgres by default. That will need a different implementation entirely for `get-tables` because external tables are not returned by the postgres query. 
See https://github.com/aws/amazon-redshift-jdbc-driver/blob/master/src/main/java/com/amazon/redshift/jdbc/RedshiftDatabaseMetaData.java#L1748-L1780

Let's make redshift use the sql-jdbc driver implementation for now.

qnkhuat:
@calherries updated to make sure we don't prematurely close the connection on a reducible and also make sure redshift use jdbc implementation.

calherries (approved):
sweet

qnkhuat:
@camsaul could you please take another look.

PR #39671, opened 2024-03-06
Count Field usage

### LINKED ISSUE #40494: Record field usage during query execution
### What

We want to record how fields are used during query execution. The schema for field usage.

| Column | Type | remark | Example |
| --- | --- | --- | --- |
| field_id | int |  |  |
| query_execution_id | integer | FK to query_execution.id | 
| used_in | varchar(25) |	which part of the query this field is used in | breakout, aggregation, filter |
| created_at | timestamp with timezone |  |  |
| filter_op | varchar(25) | the operator of the filter | > |
| aggregation_function | varchar(25) | name of the aggregation function | :sum, :max |
| breakout_binning_strategy | varchar(25) | breakout binning strategy |  num-bins |
| breakout_binning_num_bins | integer | breakout num bin binning option |  20 |
| breakout_binning_bin_width| integer | breakout bin width binning option |  10 |
| breakout_temporal_unit | text | breakout temporal unit option | |

### Gathering field usage

We’ll rely on our friends over at the QP team and ask them to help us with an API that takes a query and returns field_usage. The schema for the output is:

```clojure
[:and
 [:map
  [:clause-type [:enum :expression :breakout :aggregation :filter]]]
 [:multi {:dispatch :clause-type}
  [:expression  [:map
                 [:type                 [:= :expression]]
                 [:field-id             pos-int?]]
  [:breakout    [:map
                 [:type                 [:= :breakout]]
                 [:field-id             pos-int?]
                 [:breakout-param       [:metabase.lib.schema.ref/field.options]]]]
  [:aggregation [:map
                 [:type                 [:= :aggregation]]
                 [:field-id             pos-int?]
                 [:aggregation-function [:enum :avg :count :cum-count :count-where :distinct :var
                                         :max :median :min :percentile :share :stddev :sum :cum-sum :sum-where]]]]
  [:fitler      [:map
                 [:type                 [:= :filter]]
                 [:field-id             pos-int?]
                 [:filter-args        [:sequential :any]
                 [:filter-op           [:enum := :!= :inside :between :< :> :<= :>= :is-null :not-null :is-empty :not-empty :contains :does-not-contain :starts-with :ends-with]]]]]]
```

### Saving field usage

We already have a process that save query_execution asynchronously in  https://github.com/metabase/metabase/blob/cd1d3c70675c4dc2591dadad3968accbd2531585/src/metabase/query_processor/middleware/process_userland_query.clj#L59

For our task, we’ll extend `save-query-execution!*` to save field_usage.

qnkhuat:
Closes https://github.com/metabase/metabase/issues/40494
Milestone 2 of https://github.com/metabase/metabase/issues/38229


src/metabase/query_processor/middleware/process_userland_query.clj
```diff
@@ -32,23 +37,85 @@
...
+  from the same legacy query."
+  [x]
+  (walk/postwalk
+   (fn [x]
+     (if (map? x)
+       (dissoc x :lib/uuid)
+       x))
+   x))
+
+(defn- pmbql->field-usages
```
qnkhuat, line 52: to QP people: do you think I should break this down and make finding aggregation, breakout, filters, expressions an mlv2 function? if yes, what'd be an useful API?

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,104 @@
...
+  (concat
+    (keep #(filter->field-usage query stage-number %) (lib/filters query stage-number))
+    (keep #(aggregation->field-usage query stage-number %) (lib/aggregations query stage-number))
+    (keep #(breakout->field-usage query stage-number %) (lib/breakouts query stage-number))
+    (flatten (keep expression->field-usage (lib/expressions query stage-number)))
+    (flatten (keep #(pquery->field-usages-method query stage-number %) (:joins x)))))
+
+(defmethod pquery->field-usages-method :mbql/join
+  [query _stage-number x]
+  (let [join-query (fetch-source-query/resolve-source-cards (assoc query :stages (get x :stages)))]
```
qnkhuat, line 95: what's the best way to do this upstream? ideally `pquery->field-usages` would receive a query that get all stages and joins resolved

src/metabase/query_processor/middleware/process_userland_query.clj
```diff
@@ -141,6 +154,12 @@
    :result_rows       0
    :start_time_millis (System/currentTimeMillis)})
 
+(defn- query->field-usages
+  [query]
+  (->> (normalize/normalize-preprocessing-middleware query)
```
qnkhuat, line 159: I used normalize instead of  `lib/query` because I realized that `query` now can also be a mlv2 query. Here is the PR that introduced it https://github.com/metabase/metabase/pull/39969

test/metabase/query_processor/middleware/process_userland_query_test.clj
```diff
@@ -60,32 +64,33 @@
                                                       (respond metadata rows))]
                       (qp.pipeline/*run* query rff))))]
     (binding [driver/*driver* :h2]
-      (qp query qp.reducible/default-rff))))
+      (qp.store/with-metadata-provider (mt/id)
+        (qp query qp.reducible/default-rff)))))
 
 (deftest success-test
-  (let [query {:type ::success-test}]
+  (let [query {:database 2, :type :query, :query {:source-table 26}}]
```
qnkhuat, line 71: the tests were failing because this is an invalid mbql schema and converting legacy query to pmbql query is throwing an error.

I don't think it's possible for an invalid query to ever reach this place, so instead of guarding the implementation with try catch, I think it's better to update this test query.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+    ;; treat the source query as a :mbql/query
+    (pquery->field-usages-method join-query nil join-query)))
+
+(mu/defn pquery->field-usages
```
qnkhuat, line 95: naming suggestions are welcomed! I called this `pquery` to distinguish between normal query and mlv2 query aka pipelined mbql.

crisptrutski: I feel like just `query` vs `legacy-query` would be good names for mlv2 and mlv1 respectively. Otherwise just `mlv2-query` would be so much clearer that it's worth the verbosity.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+                                   not-empty)})))
+
+(defn- expression->field-usage
+  [expresison-clause]
+  (when-let [field-ids (seq (lib.util.match/match expresison-clause [:field _ (id :guard int?)] id))]
+    (for [field-id field-ids]
+      {:field_id field-id
+       :used_in  :expression})))
+
+(defmulti ^:private pquery->field-usages-method
```
calherries, line 69: Does this need to be a multimethod? From what I can see, for every call site of `pquery->field-usages-method` we know the dispatch type from the context, so there doesn't seem to be a need for any polymorphism here.

e.g. on [line 87](https://github.com/metabase/metabase/pull/39671/files#diff-8f297ee24e62bb71f2aed630a98593bee3f2a3e353ae8db0ab2c0edf08a03eeaR87) we already know that `x` is going to be an `mbql/join`:
```clojure
    (flatten (keep #(pquery->field-usages-method query stage-number %) (:joins x)))))
```

The only thing that unites the three implementations is that it takes _something_ and returns a sequence of field usages. But this could be three separate functions, each with more appropriate names and argument lists.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+   :filter_args            mi/transform-json})
+
+(defn- filter->field-usage
+  [query stage filter-clause]
+  (let [filter-parts (lib/filter-parts query stage filter-clause)
+        field-id     (-> filter-parts :column :id)]
+    (when (int? field-id)
+      {:field_id    field-id
+       :used_in     :filter
+       :filter_op   (-> filter-parts :operator :short)
```
calherries, line 38: I guess reaching into `::lib.schema.filter/operator` is allowed?

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+       :filter_args (:args filter-parts)})))
```
calherries, line 39: The fact the `filter_args` are a JSON array and we don't have one arg per row will mean we need to parse the queries in SQL, which means one query for each dialect. I suppose that's an okay price to pay because it makes storing this data easier (and we don't even have a use case for it), but I just want to make sure we're okay paying it in the future.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+
+(defn- aggregation->field-usage
+  [query stage aggregation-clause]
+  (let [aggregation-column (lib/aggregation-column query stage aggregation-clause)
+        field-id           (:id aggregation-column)]
+    (when (int? field-id)
```
calherries, line 45: Is `field-id=nil` being covered in the tests? I might have missed it

src/metabase/lib/core.cljc
```diff
@@ -301,8 +303,12 @@
...
+  raw-temporal-bucket
   temporal-bucket
   with-temporal-bucket]
  [lib.util
   normalized-query-type
-  source-table-id])
+  source-table-id]
+ [lib.walk
+  walk
+  walk-stages])
```
calherries, line 314: did you mean to leave these here?

src/metabase/models/field_usage.clj
calherries: from the [MLv2 primer](https://www.notion.so/metabase/MBQL-Lib-v2-Primer-b89b9fba2d894f1db759554452bba354?pvs=4#be18da1d03e1420b9395bf610cee4704):

> “MLv2 query” is a slightly-incorrect shorthand for a pMBQL query, to distinguish it from the “legacy query” form.

We often use `sql` as a name for a SQL query, so maybe we should use `pmbql` as a name for a pMBQL query

src/metabase/query_processor/middleware/process_userland_query.clj
calherries: Can't `lib/query` also handle MLv2? 

Looking at it's implementation, it seems to be designed to handle it here
https://github.com/metabase/metabase/blob/679fd3506c95e1c3b8e1a5194f315ecf647546c7/src/metabase/lib/query.cljc#L128

resources/migrations/001_update_migrations.yaml
```diff
@@ -6090,6 +6090,100 @@ databaseChangeLog:
...
+      comment: Add field_usage table
+      changes:
+        - createTable:
+            tableName: field_usage
+            remarks: Used to store field usage during query execution
+            columns:
+              - column:
+                  name: id
+                  remarks: Unique ID
+                  type: int
```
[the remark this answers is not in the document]
qnkhuat: fair concern.

src/metabase/lib/core.cljc
qnkhuat: personally I find it's useful but yea I guess I don't need it for now. will remove

src/metabase/models/field_usage.clj
qnkhuat: as for future, I was hopeing we can use jsonb or sth like similar :)

src/metabase/models/field_usage.clj
qnkhuat: the downside is that calling this function will require a metaprovider to be bound.

cc @bshepherdson for ideas.

src/metabase/query_processor/middleware/process_userland_query.clj
qnkhuat: sadly know, because `lib/type` in the JSON query is a string, not a keyword, `normalize/normalize-preprocessing-middleware` will take care of converting.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+  [query stage breakout-clause]
+  (let [breakout-column (lib/breakout-column query stage breakout-clause)
+        field-id        (:id breakout-column)]
+    (when (int? field-id)
+      {:field_id               field-id
+       :used_in                :breakout
+       :breakout_temporal_unit (lib/raw-temporal-bucket breakout-clause)
+       :breakout_binning       (-> (lib/binning breakout-clause)
+                                   (select-keys [:strategy :bin-width :num-bins])
+                                   not-empty)})))
```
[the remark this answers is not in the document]
qnkhuat: yea that's a reasonable desire. but otoh this is going to be a big table and has too many nullable columns will waste a lot of space. not sure if it's worth it.

src/metabase/models/field_usage.clj
qnkhuat: I wrote a version that did not use multimethod at first. But I took Cam's advice and tried to follow how `returned-columns-method` is defined. 

It's a bitter sweet experience since it's a bit to grasp at first. But overall I find dealing with recursive using multimethod makes the implementation of each multimethods simpler.

src/metabase/models/field_usage.clj
qnkhuat: not sure, but we surely don't want to store the whole map. All the filter functions return this operator map :(.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,100 @@
...
+
+(defn- expression->field-usage
+  [expresison-clause]
+  (when-let [field-ids (seq (lib.util.match/match expresison-clause [:field _ (id :guard int?)] id))]
```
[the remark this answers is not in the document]
qnkhuat: added a `referenced-field-ids` into `lib.util`

src/metabase/models/field_usage.clj
qnkhuat: pmbql it is 3eaa54e01b

src/metabase/models/field_usage.clj
qnkhuat: added an aggregation over a "sum" column in 1dcf8c9a6ce44defc4a9aa95fa903d2ea80fdb1c.

not sure how extensive we want to test this but it looks safe to me.

src/metabase/models/field_usage.clj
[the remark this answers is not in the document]
calherries: If you're worried about space in the DB (and I'm not sure we are), breaking it into multiple columns should actually use less space, not more. It seems counterintuitive but the reason is because you need to have a wide column to store JSON data, which dominates any overhead of having the extra columns.

I was curious and tested this with postgres just to be sure:

```
testdb=# -- Create the first table
CREATE TABLE table1 (
  id SERIAL PRIMARY KEY,
  breakout_binning VARCHAR(256)
);

-- Insert a dummy row with example values into table1
INSERT INTO table1 (breakout_binning)
VALUES ('Example binning');

-- Insert 1000 rows with only the ID column having not-null values into table1
INSERT INTO table1 (breakout_binning)
SELECT NULL
FROM generate_series(1, 1000);
CREATE TABLE
INSERT 0 1
INSERT 0 1000

testdb=# select oid, relname 
from pg_class limit 1;
  oid   |              relname               
--------+------------------------------------
 635346 | test_20231201143858__mb_row_id_seq
(1 row)


testdb=# -- Create the second table
CREATE TABLE table2 (
  id SERIAL PRIMARY KEY,
  strategy VARCHAR(25),
  "bin-width" INTEGER,
  "num-bins" INTEGER
);

-- Insert a dummy row with example values into table2
INSERT INTO table2 (strategy, "bin-width", "num-bins")
VALUES ('Example strategy', 10, 5);

-- Insert 1000 rows with only the ID column having not-null values into table2
INSERT INTO table2 (strategy, "bin-width", "num-bins")
SELECT NULL, NULL, NULL
FROM generate_series(1, 1000);
CREATE TABLE
INSERT 0 1
INSERT 0 1000


testdb=# select pg_table_size(oid) from pg_class where relname = 'table1' limit 1;
 pg_table_size 
---------------
         73728
(1 row)


testdb=# select pg_table_size(oid) from pg_class where relname = 'table2' limit 1;
 pg_table_size 
---------------
         65536
```

table2 is smaller, even though it has more columns.

src/metabase/lib/util.cljc
```diff
@@ -560,3 +561,8 @@
   (when (map? query)
     (keyword (some #(get query %)
                    [:lib/type :type "lib/type" "type"]))))
+
+(mu/defn referenced-field-ids :- [:maybe [:sequential ::lib.schema.id/field]]
+  "Find all the integer field IDs in `coll`, Which can arbitrarily be anything that is part of MLv2 query schema."
```
calherries, line 566: typo
```suggestion
  "Find all the integer field IDs in `clause`, Which can arbitrarily be anything that is part of MLv2 query schema."
```

I'm not sure this is the best place to put this function. It's coupled to the internal structure of a `:mbql.clause/field` as defined here:
https://github.com/metabase/metabase/blob/cf29360f37a4f2cb76f1d61fc8f1642f2e1a8e61/src/metabase/lib/schema/ref.cljc#L51

So maybe it makes sense to put it there

src/metabase/models/field_usage.clj
calherries: yeah, for sure. Would be nice to get qpd eyes on this though.

src/metabase/models/field_usage.clj
calherries: To be clear space and performance is not the concern I have here, so jsonb vs text doesn't make a difference. 

It's that any queries on this data will need implementing in all three dialects of the app DB if we want them to be fast. And this might not be trivial for H2. 

e.g. for postgres, I can imagine a query like this:

```
SELECT value, COUNT(*) AS count
FROM field_usage_test
CROSS JOIN UNNEST(string_to_array(values, ',')) AS t(value)
WHERE field_id = 100 AND filter_op = '='
GROUP BY value
ORDER BY count DESC;
```

We'll need to do some serious work to get this query working in H2. A guy seems to have achieved it [here](https://github.com/h2database/h2database/issues/1815#issuecomment-1363015627) by creating an alias for a function implemented in Java, but that's a lot of code.

calherries: I wonder if we should just leave filter values out of this PR, and come back to it later when we actually need to use them. It's not clear how we're going to use them, and if we do use them, we'd probably want to have them stored differently than in a JSON array.

calherries:
I would recommend someone from @metabase/core-backend-qp have a look at the code that uses MLv2. We're reaching into a few maps to get data and I'm not confident that we're not breaking any module boundaries.

src/metabase/models/field_usage.clj
calherries: cheers, that works

calherries: I have a similar question for the other clause types though, not just aggregations. is there coverage for `filter`, `breakout`, and `expression`?

src/metabase/models/field_usage.clj
bshepherdson: `lib/filter-parts` and `lib/expression-parts` return those operator maps, which are intended to be an external AST for expressions and a public interface. So **yes**, feel free to look into those maps and pull out just the pieces you need, like in this code.

src/metabase/models/field_usage.clj
calherries: > overall I find dealing with recursive using multimethod makes the implementation of each multimethods simpler.

I needed to check this out for myself, because I thought I was missing something. I don't think I am.

Here's an approach without multimethods: 3 functions and 23 LOC.

```clojure
(declare pmbql->field-usages)

(defn- join->field-usages
  [query join]
  (let [join-query (fetch-source-query/resolve-source-cards (assoc query :stages (get join :stages)))]
    ;; treat the source query as a :mbql/query
    (pmbql->field-usages join-query)))

(defn- stage->field-usages
  [query stage-number]
  (concat
    (keep #(filter->field-usage query stage-number %) (lib/filters query stage-number))
    (keep #(aggregation->field-usage query stage-number %) (lib/aggregations query stage-number))
    (keep #(breakout->field-usage query stage-number %) (lib/breakouts query stage-number))
    (flatten (keep expression->field-usage (lib/expressions query stage-number)))
    (flatten (keep #(join->field-usages query %) (:joins x)))))

(mu/defn pmbql->field-usages
  "Given a pmbql query, returns field usages from filter, breakout, aggregation, expression of a query.
  Walk all stages and joins.
  Expects all the source cards were resolved"
  [query :- ::lib.schema/query]
  (mapcat #(stage->field-usages query %) (range (lib/stage-count query))))
```

vs the original: 4 functions (counting multimethod implementations) and 48 LOC

```clojure
(defmulti ^:private pmbql->field-usages-method
  "Impl for [[pmbql->field-usages]]."
  {:arglists '([query stage-number x])}
  (fn [_query _stage-number x] (lib.dispatch/dispatch-value x))
  :hierarchy lib.hierarchy/hierarchy)

(defmethod pmbql->field-usages-method :mbql/query
  [query _stage-number x]
  (flatten (for [stage (range (lib/stage-count x))]
                (pmbql->field-usages-method query stage (lib.util/query-stage query stage)))))

(defmethod pmbql->field-usages-method :metabase.lib.stage/stage
  [query stage-number x]
  (concat
    (keep #(filter->field-usage query stage-number %) (lib/filters query stage-number))
    (keep #(aggregation->field-usage query stage-number %) (lib/aggregations query stage-number))
    (keep #(breakout->field-usage query stage-number %) (lib/breakouts query stage-number))
    (flatten (keep expression->field-usage (lib/expressions query stage-number)))
    (flatten (keep #(pmbql->field-usages-method query stage-number %) (:joins x)))))

(defmethod pmbql->field-usages-method :mbql/join
  [query _stage-number x]
  (let [join-query (fetch-source-query/resolve-source-cards (assoc query :stages (get x :stages)))]
    ;; treat the source query as a :mbql/query
    (pmbql->field-usages-method join-query nil join-query)))

(mu/defn pmbql->field-usages
  "Given a pmbql query, returns field usages from filter, breakout, aggregation, expression of a query.
  Walk all stages and joins.
  Expects all the source cards were resolved"
  [pmbql :- ::lib.schema/query]
  (pmbql->field-usages-method pmbql nil pmbql))
```

src/metabase/lib/util.cljc
qnkhuat: those lib.schema.* ns are all about schema. I don't think anyone will expect to find a helper in there 🤔

src/metabase/models/field_usage.clj
qnkhuat: yea, model shouldn't require QP ns.

> perhaps this code wants to run on the fully preprocessed query

this sounds nice. is there a helper like this, or should I roll one myself?

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,99 @@
...
+                (pmbql->field-usages-method query stage (lib.util/query-stage query stage)))))
+
+(defmethod pmbql->field-usages-method :metabase.lib.stage/stage
+  [query stage-number x]
+  (concat
+    (keep #(filter->field-usage query stage-number %) (lib/filters query stage-number))
+    (keep #(aggregation->field-usage query stage-number %) (lib/aggregations query stage-number))
+    (keep #(breakout->field-usage query stage-number %) (lib/breakouts query stage-number))
+    (flatten (keep expression->field-usage (lib/expressions query stage-number)))
+    (flatten (keep #(pmbql->field-usages-method query stage-number %) (:joins x)))))
```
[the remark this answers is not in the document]
qnkhuat: good eye. updated in [ea4aff0](https://github.com/metabase/metabase/pull/39671/commits/ea4aff04d5d1a2d1c113db614a5b603841a0d45a)

src/metabase/models/field_usage.clj
qnkhuat: added some more dummy clause that will have string field-id in [ea4aff0](https://github.com/metabase/metabase/pull/39671/commits/ea4aff04d5d1a2d1c113db614a5b603841a0d45a)

src/metabase/models/field_usage.clj
qnkhuat: removed filter_args in [ea4aff0](https://github.com/metabase/metabase/pull/39671/commits/ea4aff04d5d1a2d1c113db614a5b603841a0d45a)

src/metabase/models/field_usage.clj
qnkhuat: ok I'll bite. `defn` all the way in [ea4aff0](https://github.com/metabase/metabase/pull/39671/commits/ea4aff04d5d1a2d1c113db614a5b603841a0d45a)

src/metabase/lib/util.cljc
calherries: true, that's unfortunate because that's where the code is coupled to. maybe if it were in `metabase.lib.field` it would be slightly more discoverable? maybe a good q for qpd

calherries (approved):
Great, looks good from my perspective. I'd still like someone from QPD to review the MLv2 usage in `src/metabase/models/field_usage.clj`, and the changes in `metabase.lib`, and `metabase.query_processor` though

src/metabase/lib/util.cljc
bshepherdson: I agree that `metabase.lib.schema.*` is the wrong place for helper functions.

This is about field IDs but that doesn't mean it belongs in `lib.field`... I'm not sure there's really a better place than `util`.

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,87 @@
...
+      (let [binning-option (lib/binning breakout-clause)]
+       {:field_id                   field-id
+        :used_in                    :breakout
+        :breakout_temporal_unit     (lib/raw-temporal-bucket breakout-clause)
+        :breakout_binning_strategy  (:strategy binning-option)
+        :breakout_binning_bin_width (:bin-width binning-option)
+        :breakout_binning_num_bins  (:num-bins binning-option)}))))
+
+(defn- expression->field-usage
+  [expresison-clause]
```
bshepherdson, line 59: typo: `expression-clause`

src/metabase/models/field_usage.clj
```diff
@@ -0,0 +1,87 @@
...
+(defn- aggregation->field-usage
+  [query stage aggregation-clause]
+  (let [aggregation-column (lib/aggregation-column query stage aggregation-clause)
+        field-id           (:id aggregation-column)]
+    (when (int? field-id)
+      {:field_id             field-id
+       :used_in              :aggregation
+       :aggregation_function (first aggregation-clause)})))
+
+(defn- breakout->field-usage
```
bshepherdson, line 45: Up to you whether you want to make this change, but we do a lot of this "different logic for each kind of clause" in `metabase.lib.*`. Probably it would be

```clojure
(defmulti field-usage-method lib.dispatch/dispatch-value)

(defmethod field-usage-method :lib/aggregation [query stage aggregation-clause]
  ...)

(defn field-usage [query stage clause]
  (field-usage-method query stage clause))
```
and the defmethod can work for stages, queries, etc.; they're all different dispatch values.

You can see examples in lots of places, but maybe `display-info-method` or `metabase.lib.convert` are good examples.

As I say, this is your code so it's up to your team. Just wanted to make sure you were aware of the prior art.

src/metabase/query_processor/middleware/process_userland_query.clj
bshepherdson: This is misusing the middleware. It's not a helper function intended to be called like this. The middleware runs in a defined order, see `metabase.query-processor.preprocess/middleware` for the big list.

In particular, you'll see that `normalize/normalize-preprocessing-middleware` is the first step, and `fetch-source-query/resolve-source-cards` is near the top. So you shouldn't have to call either of these - the query will be fully preprocessed already.

test/metabase/query_processor/middleware/process_userland_query_test.clj
```diff
@@ -168,3 +172,31 @@
...
+
+(deftest save-field-usage-test
+  (testing "execute a query will save field usages"
+    (qp.store/with-metadata-provider (mt/id)
+      (doseq [query-type [:mbql :mlv2]]
+        (let [;; randomize the binning so we can use it as the identifier for fieldusage down below
+              random-binning  (rand-int 13371337)
+              breakout-clause [:field (mt/id :products :price) {:binning {:num-bins  random-binning
+                                                                          :strategy  :num-bins
+                                                                          :bin-width random-binning}}]]
```
bshepherdson, line 184: If this query is really getting executed (which seems to be the case) 10M is a huge number of bins and bin width. `(rand-int 100)` will suffice here.

bshepherdson (changes requested):
A few miscellany, but the note about misusing the QP middleware is important. Please fix that.

src/metabase/query_processor/middleware/process_userland_query.clj
qnkhuat: If I understand correctly, these "round" middlewares are executed before the "preprocessing" middleware. So the query here is not full processed.

qnkhuat: Should I create a preprocessing middleware for this? not sure if it's the right way to introduce side-affects to QP tho.

src/metabase/models/field_usage.clj
qnkhuat: My initial implementation used defmulti, but we decided to go with pure functions because it's less code and we don't need the ability parse other types.

see [this](https://github.com/metabase/metabase/pull/39671#discussion_r1547548825) discussion for context

src/metabase/models/field_usage.clj
qnkhuat: fixed

test/metabase/query_processor/middleware/process_userland_query_test.clj
qnkhuat: I've updated the query so that it's not executed.
see https://github.com/metabase/metabase/pull/39671/commits/698fa2354051f611337751c345573bebc1d2a729

qnkhuat:
@bshepherdson I've updated so we don't process the query in process-userland-query middleware, instead we get the passed the processed query through metadata and then dissoc it before returning.

Please help reviews!

src/metabase/query_processor/middleware/process_userland_query.clj
```diff
@@ -172,13 +171,16 @@
...
       (let [query          (assoc-in query [:info :query-hash] (qp.util/query-hash query))
-            execution-info (query-execution-info query)
-            field-usages   (query->field-usages query)]
+            execution-info (query-execution-info query)]
         (letfn [(rff* [metadata]
-                  (let [result (rff metadata)]
+                  (let [preprocessed-query (:preprocessed_query metadata)
+                        ;; we only need the preprocessed query to find field usages, so make sure we don't return it
+                        result             (rff (dissoc metadata :preprocessed_query))
+                        field-usages       (field-usage/pmbql->field-usages
```
piranha, line 182: This seem to be happening synchronously during query execution, I wonder if there is any sense into passing it somewhere to execute async? Especially if we add parsing SQL in there.

qnkhuat: wdym by parsing SQL?

piranha: Well right now it get field usages from mbql queries, right? But not from SQL?

qnkhuat: yes, only mbql

bshepherdson (approved):

PR #39673, opened 2024-03-06
Fix mistake in API changelog: type can be "question" not "query"

calherries:
This PR fixes a mistake in the API changelog: type can be "question" not "query".


calherries:
Just to sense check, see Cam's PR:
https://github.com/metabase/metabase/pull/38981/files#diff-63be21df3f5f1d4567f91ecde84d2b1e08a726f45b44172e27f958c264f90ee4R25

crisptrutski (approved):

PR #39679, opened 2024-03-06
Update existing foreign keys during metadata sync

### LINKED ISSUE #39685: If a foreign key target changes, sync ignores the change
### Describe the bug

Currently if a foreign key relationship changes in the customer database, we don't pick it up on syncs after the initial sync. 

### To Reproduce

Repro test:
```clojure
(deftest update-fk-relationships-test
  (testing "Check that Foreign Key relationships can be updated"
    (let [; dataset tables need at least one field other than the ID column, so just add a dummy field
          name-field-def {:field-name "dummy", :base-type :type/Text}]
      (mt/with-temp-test-data
        [["continent_1" [name-field-def]
          []]
         ["continent_2" [name-field-def]
          []]
         ["country" [name-field-def {:field-name "continent_id", :base-type :type/Integer}]
          []]]
        (let [db            (mt/db)
              get-fk-target #(t2/select-one-fn :fk_target_field_id :model/Field (mt/id :country :continent_id))]
          ;; 1. add FK relationship in the database targeting continent_1
          (t2/query-one db "ALTER TABLE country ADD CONSTRAINT country_continent_id_fkey FOREIGN KEY (continent_id) REFERENCES continent_1(id);")
          (sync/sync-database! db {:scan :schema})
          (testing "initially country's continent_id is targeting continent_1"
            (is (= (mt/id :continent_1 :id)
                   (get-fk-target))))
          ;; 2. drop the FK relationship in the database with SQL
          (t2/query-one db "ALTER TABLE country DROP CONSTRAINT country_continent_id_fkey;")
          ;; 3. add back the FK relationship but targeting continent_2
          (t2/query-one db "ALTER TABLE country ADD CONSTRAINT country_continent_id_fkey FOREIGN KEY (continent_id) REFERENCES continent_2(id);")
          (sync/sync-database! db {:scan :schema})
          (testing "initially country's continent_id is targeting continent_2"
            (is (= (mt/id :continent_2 :id)
                   (get-fk-target)))))))))
```

This fails with 
```
FAIL in (update-fk-relationships-test) (NO_SOURCE_FILE:8395)
Check that Foreign Key relationships can be updated using dataset-definition dataset
 initially country's continent_id is targeting continent_2
expected: 7
  actual: 8
    diff: - 7
          + 8
```

### Expected behavior

The fk_target_field_id should change and the above test should pass

### Logs

_No response_

```JSON
master
```

### Severity

low, no issue has been logged yet

### Additional context

_No response_

calherries:
Fixes https://github.com/metabase/metabase/issues/39685

This is also a pre-requisite for [Faster sync-fks](https://github.com/metabase/metabase/pull/38970), because the approach I'm taking there will use more memory without this change.

It also reduces the memory footprint of the "sync-fks" step for databases with many tables. Previously we would allocate all of the tables of a database every time in a vector.

I tested the impact on speed using the below criterium benchmark, which simulates the `sync-fks!` step for a database with 100 foreign keys:
```clojure
(let [n-foreign-keys 100]
  (mt/with-temp-test-data
    (mapcat
     (fn [i]
       [[(str "continent_" i)
         [{:field-name "name", :base-type :type/Text}]
         []]
        [(str "country_" i)
         [{:field-name "name", :base-type :type/Text}
          {:field-name "continent_id", :base-type :type/Integer}]
         []]])
     (range n-foreign-keys))
    (metabase.sync/sync-database! (mt/db) {:scan :schema})
    (doseq [i (range n-foreign-keys)]
      (t2/query-one (mt/db) (str "ALTER TABLE country_" i " ADD CONSTRAINT contraint_" i " FOREIGN KEY (continent_id) REFERENCES continent_" i "(id);")))
    (t2/with-call-count [call-count]
      (criterium.core/quick-bench (metabase.sync.sync-metadata.fks/sync-fks! (mt/db)))
      (call-count))))
```

Using a local postgres app DB and an H2 test DB:

Before:
Evaluation count : 12 in 6 samples of 2 calls.
Execution time mean : 73.440179 ms
After: 
Evaluation count : 6 in 6 samples of 1 calls.
Execution time mean : 132.344456 ms

I'd need to dive into the reason behind the difference in performance, but it gives me enough confidence to move on to [Faster sync-fks](https://github.com/metabase/metabase/pull/38970), which is where the real performance improvement will come from and more than make up for it.

### DIFF
--- a/modules/drivers/mongo/test/metabase/driver/mongo_test.clj
+++ b/modules/drivers/mongo/test/metabase/driver/mongo_test.clj
@@ -379,16 +379,16 @@
     (with-redefs [metadata-queries/nested-field-sample-limit 2]
       (binding [tdm/*remove-nil?* true]
         (mt/with-temp-test-data
-          ["bird_species"
-           [{:field-name "name", :base-type :type/Text}
-            {:field-name "favorite_snack", :base-type :type/Text}
-            {:field-name "max_wingspan", :base-type :type/Integer}]
-           [["Sharp-shinned Hawk" nil 68]
-            ["Tropicbird" nil 112]
-            ["House Finch" nil nil]
-            ["Mourning Dove" nil nil]
-            ["Common Blackbird" "earthworms" nil]
-            ["Silvereye" "cherries" nil]]]
+          [["bird_species"
+            [{:field-name "name", :base-type :type/Text}
+             {:field-name "favorite_snack", :base-type :type/Text}
+             {:field-name "max_wingspan", :base-type :type/Integer}]
+            [["Sharp-shinned Hawk" nil 68]
+             ["Tropicbird" nil 112]
+             ["House Finch" nil nil]
+             ["Mourning Dove" nil nil]
+             ["Common Blackbird" "earthworms" nil]
+             ["Silvereye" "cherries" nil]]]]
           ;; do a full sync on the DB to get the correct semantic type info
           (sync/sync-database! (mt/db))
           (is (= #{{:name "_id", :database_type "java.lang.Long", :base_type :type/Integer, :semantic_type :type/PK}
--- a/src/metabase/sync/sync_metadata/fks.clj
+++ b/src/metabase/sync/sync_metadata/fks.clj
@@ -41,7 +41,9 @@
                                   :semantic_type      "type/FK"}
                             :where [:and
                                     [:= :f.id fk-field-id-query]
-                                    dest-field-id-query]})
+                                    [:or
+                                     [:= :f.fk_target_field_id nil]
+                                     [:not= :f.fk_target_field_id dest-field-id-query]]]})
       (when (= <> 1)
         (log/info (u/format-color 'cyan "Marking foreign key from %s %s -> %s %s"
                                   (sync-util/table-name-for-logging table)
--- a/src/metabase/sync/sync_metadata/fks.clj
+++ b/src/metabase/sync/sync_metadata/fks.clj
@@ -8,7 +8,8 @@
    [metabase.util :as u]
    [metabase.util.log :as log]
    [metabase.util.malli :as mu]
-   [toucan2.core :as t2]))
+   [toucan2.core :as t2]
+   [toucan2.realize :as t2.realize]))
 
 (mu/defn ^:private mark-fk!
   "Updates the `fk_target_field_id` of a Field. Returns 1 if the Field was successfully updated, 0 otherwise."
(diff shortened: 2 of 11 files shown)


src/metabase/sync/sync_metadata/fks.clj
```diff
@@ -93,4 +77,4 @@
           {:total-fks    0
            :updated-fks  0
            :total-failed 0}
-          (sync-util/db->sync-tables database)))
+          (sync-util/db->reducible-sync-tables database)))
```
calherries, line 81: This fixes an issue where we hold all tables in memory during this sync step. This is okay for most databases but could become problematic when there are 100,000s of tables.

modules/drivers/mongo/test/metabase/driver/mongo_test.clj
```diff
@@ -379,16 +379,16 @@
...
-           [{:field-name "name", :base-type :type/Text}
-            {:field-name "favorite_snack", :base-type :type/Text}
-            {:field-name "max_wingspan", :base-type :type/Integer}]
-           [["Sharp-shinned Hawk" nil 68]
-            ["Tropicbird" nil 112]
-            ["House Finch" nil nil]
-            ["Mourning Dove" nil nil]
-            ["Common Blackbird" "earthworms" nil]
-            ["Silvereye" "cherries" nil]]]
+          [["bird_species"
```
calherries, line 382: I changed `mt/with-temp-test-data` to take a list of table definitions, since it only supported one table definition before.

src/metabase/sync/sync_metadata/fks.clj
```diff
@@ -1,65 +1,101 @@
...
+           fk-column-name
+           pk-table-name
+           pk-table-schema
+           pk-column-name]}]
+  (let [field-id-query (fn [db-id table-schema table-name column-name]
+                         {:select [[[:min :f.id] :id]]
+                          ;; Cal 2024-03-04: We use `min` to limit this subquery to one result (limit 1 isn't allowed
+                          ;; in subqueries in MySQL) because it's possible for schema, table, or column names to be
+                          ;; non-unique when lower-cased for some DBs. We have been doing case-insensitive matching
+                          ;; since #39679 so this preserves behaviour to avoid possible regressions.
```
crisptrutski, line 30: Is this the right resource link? It points back to the current PR.

src/metabase/sync/sync_metadata/fks.clj
```diff
@@ -1,65 +1,101 @@
...
+                      [:and
+                       [:or
+                        [:= :f.fk_target_field_id nil]
+                        [:not= :f.fk_target_field_id :pk.id]]]]
+             :set    {:fk_target_field_id :pk.id
+                      :semantic_type      "type/FK"}}
+            :postgres
+            {:update [:metabase_field :f]
+             :from   [[fk-field-id-query :fk]]
+             :join   [[pk-field-id-query :pk] true]
```
crisptrutski, line 61: Does Postgres not support the constraint going here for the statement, or is this a quirk for performance?

crisptrutski (approved):
LGTM

PR #39683, opened 2024-03-06
[E2E] Overload the existing E2E visitQuestion helper to make it work with the alias

kamilmielnik:
Similar to #38309 (I even copied the PR description from there)

Depends on #39629

### Description

This PR overloads the existing `visitQuestion` E2E helper to make it accept either:
1. a direct question id (number) or
2. a question id alias (string)

It then replaces and updates related specs that were grabbing the question id alias and then calling the `visitQuestion` function. The code can be simplified like this:

```diff
- cy.get("@myQuestion").then(id => {
-   visitQuestion(id);
- });
+ visitQuestion("@myQuestion");
```


e2e/support/helpers/e2e-misc-helpers.js
```diff
@@ -237,14 +252,6 @@ function visitDashboardById(dashboard_id, config) {
   });
 }
 
-/**
- * Visit a dashboard by using its previously saved dashboard id alias.
- * @param {string} alias
- */
-function visitDashboardByAlias(alias, config) {
```
kamilmielnik, line 244: I decided to inline this function, because:
- it's 1 LOC
- it's used only in 1 place
- there would be a 1 function like that for: dashboard, question, model (in the future), metric (in the future), etc.

e2e/test/scenarios/native/native_subquery.cy.spec.js
```diff
@@ -265,15 +265,14 @@ describe("scenarios > question > native subquery", () => {
...
+      "have.text",
+      "Showing 41 rows",
+    );
 
-    cy.get("@toplevelQuestionId").then(toplevelQuestionId => {
-      visitQuestion(toplevelQuestionId);
-      cy.contains("41");
-    });
+    visitQuestion("@toplevelQuestionId");
+    cy.get("#main-data-grid .cellData").should("have.text", "41");
```
kamilmielnik, line 275: The selector can be prettier only if we use `contain` instead of `have`. Since `have` is a stronger assertion, I went with an uglier selector.

uladzimirdev (approved):

kamilmielnik:

PR #39686, opened 2024-03-06
Show models correctly in the data source picker

### LINKED ISSUE #39699: Cannot see Models as the data source in the data selector when opening a model
### Describe the bug

When you open a saved model or a question/model based on a saved model in a notebook mode, you currently cannot see the underlying source model in the data picker. This renders the data picker nearly unusable for models.

### To Reproduce

The simplest possible scenario:
1. Open any saved model
2. Switch to the notebook mode
3. Click on the model name in the "data" step (in our case it is the "Accounts Model")
image
4. It opens "Saved Questions" and it doesn't find the model there, so nothing is highlighted
image

### Expected behavior

Clicking on an entity based on a model should show Models data selector with the resulting model being highlighted.
image

### Logs

_No response_

local dev, master, 4a78412, H2, Sample Database

### Severity

P2

### Additional context

This behavior existed since forever - basically from the moment we introduced models.

nemanjaglumac:
Models were supported in the data selector (data picker) only when starting the "New > Question" or a "New > Model" flow. But once you save that model and want to see its source, the data picker didn't work anymore. This was due to this check (that is passed as a prop to many child components):
```jsx
isDatasets={ selectedDataBucketId === DATA_BUCKET.MODELS }
```

But when you are opening a saved entity, this data bucket doesn't exist, and the component doesn't know what to display, so it always defaulted to "Saved Questions". Luckily, there is already an information (`id`) about the saved entity we're opening. This PR introduces a fetching logic that uses said `id` to fetch a question, and it then checks its `type`. The type can either be a question or a model (for now).

Now our previous logic that determines whether we're showing models in the data picker becomes:
```js
const isDatasets =
  // Are we in the "New > ___" mode
  selectedDataBucketId === DATA_BUCKET.MODELS ||
  // or do we open an already saved entity?
  savedEntityType === "model";
```

### Tests
An E2E reproduction has been added for now, until I figure out how much can I cover with FE unit/integration tests.

Reproduces and fixes #39699

### DIFF
--- a/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
+++ b/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
@@ -330,7 +330,6 @@ export class UnconnectedDataSelector extends Component {
       fetchFields,
       fetchQuestion,
       selectedDataBucketId,
-      selectedQuestion,
       selectedTableId: sourceId,
     } = this.props;
 
@@ -345,10 +344,10 @@ export class UnconnectedDataSelector extends Component {
     if (sourceId) {
       await fetchFields(sourceId);
       if (this.isSavedEntitySelected()) {
-        fetchQuestion(sourceId);
+        await fetchQuestion(sourceId);
 
         this.showSavedEntityPicker({
-          entityType: selectedQuestion?.type(),
+          entityType: this.props.selectedQuestion?.type(),
         });
       }
     }
--- a/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
+++ b/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
@@ -20,6 +20,7 @@ import { getMetadata } from "metabase/selectors/metadata";
 import { getSetting } from "metabase/selectors/settings";
 import {
   isVirtualCardId,
+  getQuestionIdFromVirtualTableId,
   SAVED_QUESTIONS_VIRTUAL_DB_ID,
 } from "metabase-lib/metadata/utils/saved-questions";
 import { getSchemaName } from "metabase-lib/metadata/utils/schema";
@@ -124,6 +125,7 @@ export class UnconnectedDataSelector extends Component {
       selectedFieldId: props.selectedFieldId,
       searchText: "",
       isSavedEntityPickerShown: false,
+      savedEntityType: null,
     };
     const computedState = this._getComputedState(props, state);
     this.state = {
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
```diff
@@ -330,13 +332,16 @@ export class UnconnectedDataSelector extends Component {
...
-      this.showSavedEntityPicker();
+      this.showSavedEntityPicker({ entityType: "model" });
     }
 
     if (sourceId) {
       await this.props.fetchFields(sourceId);
       if (this.isSavedEntitySelected()) {
-        this.showSavedEntityPicker();
+        const id = getQuestionIdFromVirtualTableId(sourceId);
+        const question = this.props.metadata?.question(id);
```
nemanjaglumac, line 342: @kulyk suggested we don't get a question from the existing metadata in case it gets stale. He suggested using our entity loader for this. I haven't been able to find an elegant way to do this.

@ranquild does this solution seem acceptable, and/or would you also suggest fetching the question using the entity framework?

---
Before this change we'd always instantiate the "Saved question picker" (now renamed to the saved entity picker) for saved questions. Unless you were going through the "New > ..." flow, you could never see a model as a source.

With the introduction of new entities (models, metrics) this saved entity picker needs to differentiate between different entity types. So I needed to construct the question somehow to be able to use `question.type()`. Since I already have the question id, I thought the easiest way was to get it from the metadata.

ranquild: Using `metadata` here won't work. It might not be present in the metadata. 

Add to `mapDispatchToProps`:
```
fetchQuestion: selectedTableId => Qiestions.actions.fetch({ 
  id: getQuestionIdFromVirtualTableId(selectedTableId),
}),
```

then add to `mapStateToProps`
```
selectedQuestion: (state, { selectedTableId }) => Questions.selectors.getObject(state, { 
  entityId:  getQuestionIdFromVirtualTableId(selectedTableId),
}),
```

then here:
```
if (this.isSavedEntitySelected()) {
  await this.fetchQuestion(sourceId);
  this.showSavedEntityPicker({ entityType: this.selectedQuestion?.type() });
}
```

nemanjaglumac: Thank you.
Done in https://github.com/metabase/metabase/pull/39686/commits/513cb483b5de1507e20189f3d1a202c9456c38e2

frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
```diff
@@ -323,20 +326,30 @@ export class UnconnectedDataSelector extends Component {
...
 
     if (sourceId) {
-      await this.props.fetchFields(sourceId);
+      await fetchFields(sourceId);
       if (this.isSavedEntitySelected()) {
-        this.showSavedEntityPicker();
+        fetchQuestion(sourceId);
+
+        this.showSavedEntityPicker({
+          entityType: selectedQuestion?.type(),
```
ranquild, line 351: AFAIK this won't work if you don't use `this.props.selectedQuestion` here, otherwise you would reference the `undefined` value before the `fetchQuestion` call. 

Btw, `fetchQuestion` should be await'ed as well.

ranquild (approved):

PR #39688, opened 2024-03-06
Do not create questions for dashcards with permission errors

ranquild:
Spotted by https://github.com/metabase/metabase/pull/39682

We should never create an MLv2 query from `dataset_query: null`.

### DIFF
--- a/frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.tsx
+++ b/frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.tsx
@@ -150,17 +150,17 @@ export function DashCardCardParameterMapper({
     }
 
     // virtual or action dashcard
-    if (!question) {
+    if (!isQuestionDashCard(dashcard)) {
       return true;
     }
 
-    if (!card.dataset_query) {
+    if (!question || !card.dataset_query) {
       return false;
     }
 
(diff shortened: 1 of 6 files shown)


frontend/src/metabase/parameters/utils/mapping-options.ts
```diff
@@ -144,7 +144,11 @@ export function getParameterMappingOptions(
     return actionParams || [];
   }
 
-  if (!card.dataset_query || (dashcard && isVirtualDashCard(dashcard))) {
+  if (
+    !question ||
+    !card.dataset_query ||
```
kamilmielnik, line 149: We can use the new `isQuestionCard` here.

frontend/src/metabase/dashboard/components/DashCard/DashCardParameterMapper/DashCardCardParameterMapper.tsx
```diff
@@ -150,17 +150,17 @@ export function DashCardCardParameterMapper({
     }
 
     // virtual or action dashcard
-    if (!question) {
+    if (!isQuestionDashCard(dashcard)) {
       return true;
     }
 
-    if (!card.dataset_query) {
+    if (!question || !card.dataset_query) {
```
kamilmielnik, line 157: We can use the new `isQuestionCard` here.

kamilmielnik (approved):
Code looks good.
It'd be great to cover these changes with tests. WDYT?

PR #39703, opened 2024-03-06
Allow dropdown source for all string operators in dashboards

ranquild:
How to verify:
- Add string/contains parameter to the dashboard and connect it to a field
- Set "dropdown" options in the sidebar for this parameter
- Make sure you can select field values in the widget for this parameter
- Custom value source options should also work

Screenshot 2024-03-06 at 18 38 27
Screenshot 2024-03-06 at 18 38 33

### DIFF
--- a/frontend/src/metabase-lib/parameters/utils/parameter-source.unit.spec.ts
+++ b/frontend/src/metabase-lib/parameters/utils/parameter-source.unit.spec.ts
@@ -71,7 +71,7 @@ describe("getQueryType", () => {
     );
   });
 
-  it('should return "none" query type by default for legacy parameter types', () => {
+  it('should return "list" query type by default for legacy parameter types', () => {
     const parameter = createMockUiParameter({
       type: "category",
     });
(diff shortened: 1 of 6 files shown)


cdeweyx (approved):
LGTM

frontend/src/metabase-lib/parameters/utils/parameter-source.unit.spec.ts
```diff
@@ -30,6 +31,54 @@ const searchField = metadata.field(SEARCH_FIELD_ID) as Field;
...
+      (operator, queryType) => {
+        const parameter = createMockUiParameter({
+          type: `${type}/${operator}`,
+        });
+        expect(getQueryType(parameter)).toBe(queryType);
+      },
+    );
+  });
+
+  it('should return "none" query type by default for legacy parameter types', () => {
```
kamilmielnik, line 74: The name of the test does not match the assertion (`none` vs `list`).

kamilmielnik (commented):
Works well :+1: 
Left a comment inline.

frontend/src/metabase-lib/parameters/utils/parameter-source.unit.spec.ts
ranquild: Fixed

kamilmielnik (approved):

PR #39705, opened 2024-03-06
Chart Settings Table Formatting DnD kit

### LINKED ISSUE #39655: Chart Settings - `ChartSettingTableFormatting`

npfitz:
Closes https://github.com/metabase/metabase/issues/39655

### Description
PR to transition DnD logic in `<ChartSettingsTableFormatting />`  from `react-sortable-hoc` to `@dnd-kit`. Needed to introduce the idea of an ID to formatting rules, but this should not impact user experience. Added e2e test for adding/removing/reordering formatting rules

### How to verify

1. Go to a table visualization, add / remove / reorder any formatting rules
2. Same for pivot table visualization.

### Checklist

### DIFF
--- a/frontend/src/metabase/visualizations/components/settings/ChartSettingsTableFormatting.jsx
+++ b/frontend/src/metabase/visualizations/components/settings/ChartSettingsTableFormatting.jsx
@@ -91,7 +91,7 @@ const INPUT_CLASSNAME = "mt1 full";
 const getValueForDescription = rule =>
   ["is-null", "not-null"].includes(rule.operator) ? "" : ` ${rule.value}`;
 
-const ChartSettingsTableFormatting = props => {
+export const ChartSettingsTableFormatting = props => {
   const [editingRule, setEditingRule] = useState();
   const [editingRuleIsNew, setEditingRuleIsNew] = useState();
 
@@ -134,7 +134,8 @@ const ChartSettingsTableFormatting = props => {
           setEditingRule(index);
           setEditingRuleIsNew(false);
         }}
-        //This needs to be
+        // This needs to be an async function so that onChange will complete (and value will be updated)
+        // Before we set the state values for the next render
         onAdd={async () => {
           await onChange([
             {
@@ -551,5 +552,3 @@ const RuleEditor = ({
     </div>
   );
 };
-
-export default ChartSettingsTableFormatting;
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/visualizations/components/settings/ChartSettingsTableFormatting.jsx
```diff
@@ -92,106 +91,131 @@ const INPUT_CLASSNAME = "mt1 full";
...
+              ...DEFAULTS_BY_TYPE["single"],
+              // if there's a single column use that by default
+              columns: cols.length === 1 ? [cols[0].name] : [],
+              id: value.length,
+            },
+            ...value,
+          ]);
+          setEditingRuleIsNew(true);
+          setEditingRule(0);
+        }}
```
JesseSDevaney, line 150: Why does this function need to be async? It looks like you typed half of your thought in that comment :laughing:

frontend/src/metabase/visualizations/components/settings/ChartSettingsTableFormatting.jsx
```diff
@@ -92,106 +91,131 @@ const INPUT_CLASSNAME = "mt1 full";
 const getValueForDescription = rule =>
   ["is-null", "not-null"].includes(rule.operator) ? "" : ` ${rule.value}`;
 
-export default class ChartSettingsTableFormatting extends Component {
-  state = {
-    editingRule: null,
-    editingRuleIsNew: null,
-  };
+const ChartSettingsTableFormatting = props => {
```
JesseSDevaney, line 94: Can we update the exports and imports to use non-default imports?

```suggestion
export function ChartSettingsTableFormatting(props) {
```

JesseSDevaney (approved):
:gem: Worked well and LGTM

Just left a couple nit comments that would be nice to address before merging

frontend/src/metabase/visualizations/components/settings/ChartSettingsTableFormatting.jsx
npfitz: I'm easily distracted 🤦 

The logic here is kinda weird, but this needs to be async because otherwise value isn't updated by the time we do the next render. The end result is that we don't automatically pop open the column select for you if you click "Add a rule" and you already have an existing rule.

frontend/src/metabase/visualizations/components/settings/ChartSettingsTableFormatting.jsx
npfitz: 👍

Issue #38926, opened 2024-02-19
ParseSQL: Figure out which fields are used in which query


tsmacdonald:
Use the Field values from DB syncs

tsmacdonald:
Fixed by #39707

PR #39707, opened 2024-03-06
SQL Lineage: Track field usage, expose stale queries

tsmacdonald:
This links Cards (with native queries) to Fields, showing us which queries use which Fields.

QueryFields are automatically maintained as part of the Card lifecycle.

When a Field is deactivated, the QueryField becomes stale. Affected cards are available via HTTP request.


src/metabase/driver/mysql/ddl.clj
```diff
@@ -44,7 +44,7 @@
     (when result?
       ;; Can't use a prepared parameter with these statements
       (sql.ddl/execute! conn [(str "kill " pid)])
-      (throw (Exception. (trs "Killed mysql process id {0} due to timeout." pid))))))
```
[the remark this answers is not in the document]
crisptrutski: I hope you let them know that their contribution was valued in the end 🖖

.clj-kondo/config.edn
```diff
@@ -153,6 +153,7 @@
     honeysql.helpers                                                hh
     honeysql.types                                                  htypes
     java-time                                                       t
+    macaw.core                                                      mac
```
crisptrutski, line 156: Macaw! Ma-core!

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,55 @@
...
+  [name]
+  (-> name
+      (str/replace "\"" "")
+      u/lower-case-en))
+
+(defn- fields-for-query
+  [q db-id]
+  (let [{:keys [columns tables]} (-> q
+                                     mac/parsed-query
+                                     mac/query->components)]
```
crisptrutski, line 24: pico nit: I think this threading expression is easier to read if it's kept on one line.

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,55 @@
...
+    (t2/query {:select [[:f/id :field_id] [:f/name :column_name] [:t/name :table_name]]
+               :from [[(t2/table-name :model/Field) :f]]
+               :join [[(t2/table-name :model/Table) :t] [:= :table_id :t.id]]
+               :where [:and
+                       [:= :t.db_id db-id]
+                       [:in :%lower.t/name (map normalize-name tables)]
+                       [:in :%lower.f/name (map normalize-name columns)]]})))
```
crisptrutski, line 31: Does this handle the case where the columns where namespaced with a table alias in the query?

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,55 @@
...
+  (-> name
+      (str/replace "\"" "")
+      u/lower-case-en))
+
+(defn- fields-for-query
+  [q db-id]
+  (let [{:keys [columns tables]} (-> q
+                                     mac/parsed-query
+                                     mac/query->components)]
+    (t2/query {:select [[:f/id :field_id] [:f/name :column_name] [:t/name :table_name]]
```
crisptrutski, line 25: I was expecting us to do some kind of filtering here where we resolve ambiguous columns. Is that idea that we'll just track the union of all things that the query could possibly refer to?

It's not clear to me from this code whether we handle the non-ambiguous cases where columns are qualified - unit tests covering all these edges would be great!

Since there are no bad ideas in a brainstorm, in the extreme case we could shrink the query or use a cache of its results to reason about which fields it used. Sounds like this would break horrifically for complex queries, but you know what they say about 80% 😜

src/metabase/driver/mysql/ddl.clj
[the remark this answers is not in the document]
tsmacdonald: Absolutely!

src/metabase/models/card.clj
```diff
@@ -521,6 +522,10 @@
       pre-update
       populate-query-fields
       maybe-populate-initially-published-at
+      ;; TODO: this should go in after-update once camsaul/toucan2#145 is fixed
```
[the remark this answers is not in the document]
tsmacdonald: We are, but my hands are mostly tied until https://github.com/camsaul/toucan2/issues/145 is fixed (which I'm working on this week)

src/metabase/native_query_analyzer.clj
tsmacdonald: > Is that idea that we'll just track the union of all things that the query could possibly refer to?

For now. I wanted to get the basic plumbing of this PR out the door. The next step is making things smarter along the lines you described, but I wanted to separate the concerns of making new app DB tables and API endpoints and such from twiddling around with table inference.

resources/migrations/001_update_migrations.yaml
```diff
@@ -5823,6 +5823,60 @@ databaseChangeLog:
         - sql:
             sql: >-
               DROP VIEW IF EXISTS v_content;
+  - changeSet:
+      id: v50.2024-02-29T15:06:43
+      author: tsmacdonald
+      comment: Add the field_usage join table
+      changes:
```
tsmacdonald, line 5830: @qnkhuat this is our chance to bikeshed about the `FieldUsage` name

src/metabase/api/card.clj
```diff
@@ -105,7 +106,8 @@
                                                    [:like :c.dataset_query (format "%%#%s%%" model-id)]]]]
                         :where [:and [:= :m.id model-id] [:not :c.archived]]})
        ;; now check if model-id really occurs as a card ID
-       (filter (fn [card] (some #{model-id} (-> card :dataset_query query/collect-card-ids))))))
+       (filter (fn [card] (some #{model-id} (-> card :dataset_query query/collect-card-ids))))
+       (sort-by :name)))
```
tsmacdonald, line 110: Some local tests suggest that the `sort-by` is not necessary, but looking at the query I have no idea how that could be, so I added it for reassurance. If you can prove that it's redundant I'm happy to remove it, though (and add an explanatory comment for bozos like myself)

src/metabase/api/card.clj
```diff
@@ -125,6 +128,20 @@
...
+                                                            {:select [:fu.* [:f.name :new_column_name] [:t.name :new_table_name]]
+                                                             :from   [[(t2/table-name :model/FieldUsage) :fu]]
+                                                             :join   [[(t2/table-name :model/Field) :f] [:= :f.id :fu.field_id]
+                                                                      [(t2/table-name :model/Table) :t] [:= :t.id :f.table_id]]
+                                                             :where  [:= false :fu.is_current]}))
+        card-ids              (keys card-id->field-usages)]
+    (if (empty? card-ids)
+      []
+      (map #(assoc % :field_usages (get card-id->field-usages (:id %)))
+           (t2/select Card :id [:in card-ids] default-order)))))
```
tsmacdonald, line 143: I can be a clown about more complex hydration cases like this. Doing it without toucan's `hydrate` seems as efficient as it could be and like fairly clear code, but if there's a case for using `hydrate` I'm happy to hear it.

src/metabase/api/card.clj
```diff
@@ -125,6 +128,20 @@
...
   (cards-for-segment-or-metric :segment model-id))
 
+(defmethod cards-for-filter-option* :stale
+  [_]
+  (let [card-id->field-usages (group-by :card_id (t2/select :model/FieldUsage
+                                                            {:select [:fu.* [:f.name :new_column_name] [:t.name :new_table_name]]
+                                                             :from   [[(t2/table-name :model/FieldUsage) :fu]]
+                                                             :join   [[(t2/table-name :model/Field) :f] [:= :f.id :fu.field_id]
+                                                                      [(t2/table-name :model/Table) :t] [:= :t.id :f.table_id]]
+                                                             :where  [:= false :fu.is_current]}))
```
tsmacdonald, line 138: Queries like this are exactly in the gray area of "should this be in the model or the API?" for me. It's like the old Rails debates: fat model or fat controller?

crisptrutski:
> Still to do: identifying stale FieldUsages as part of sync.

Since this is marked for review I'm guessing this is for a later PR? Just checking so I don't review prematurely 😄

src/metabase/api/card.clj
crisptrutski: > Should this be in the model or the API? It's like the old Rails debates...

I was always more of a Domain Context Interaction guy 😜 - a third way, for the third estate.

src/metabase/models/card.clj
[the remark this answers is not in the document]
tsmacdonald: sorry sorry sorry, I mean 129 https://github.com/camsaul/toucan2/issues/129

src/metabase/api/card.clj
crisptrutski: This does feel like Model code though.

tsmacdonald:
> Since this is marked for review I'm guessing this is for a later PR? Just checking so I don't review prematurely 😄

@crisptrutski nope, that was a stale comment. Updated the description, sorry!

src/metabase/api/card.clj
```diff
@@ -125,6 +128,20 @@
...
+        card-ids              (keys card-id->field-usages)]
+    (if (empty? card-ids)
+      []
```
crisptrutski, line 141: I wish toucan2 would just optimize away these empty selects for us...

crisptrutski: Superficially, why not use nil-punning for the empty case - `(when (seq card-ides) (map ...))`

src/metabase/api/card.clj
```diff
@@ -125,6 +128,20 @@
...
+      (map #(assoc % :field_usages (get card-id->field-usages (:id %)))
```
crisptrutski, line 142: nitpick: I'd just lean on function application of the map, as its name encourages us to think of it as an opaque function.

src/metabase/models/card.clj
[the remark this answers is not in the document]
crisptrutski: Update the comment to link to that one then? I also think its worth explicitly describing the risk it exposes in the comment.

src/metabase/api/card.clj
crisptrutski: I'd love someone to take me through hydration works, and until then I don't have a strong opinion.

I would however hoist up the append function, big readability improvement IMHO.

I think it would even be nice to just make it a top-level function that gets curried with the map:

```clojure
(defn- assoc-field-usages [card-id->field-usages {:keys [id] :as card}]
  (assoc card :field_usages (card-id->field-usages id)))
```

I know this level of decomposition is not for everyone though.

src/metabase/api/card.clj
```diff
@@ -61,37 +63,36 @@
...
-                                                                      :user_id api/*current-user-id*)
-                                                                  :card)
-                    :when                                 (not archived)]
-                card)]
-    (sort-by :name cards)))
+  (as-> (t2/select [CardBookmark :card_id] :user_id api/*current-user-id*) <>
+      (t2/hydrate <> :card)
+      (map :card <>)
+      (remove :archived <>)
+      (sort-by :name <>)))
```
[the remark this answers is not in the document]
crisptrutski: I think using a let and then a uniform arrow is even easier on the eyes:

```clojure
(let [bookmarks (t2/select ...)]
  (->> (t2/hydrate bookmarks :card)
       (map :card)
       (remove :archived)
       (sort-by :name)))
```

src/metabase/api/card.clj
```diff
@@ -105,7 +106,8 @@
                                                    [:like :c.dataset_query (format "%%#%s%%" model-id)]]]]
                         :where [:and [:= :m.id model-id] [:not :c.archived]]})
        ;; now check if model-id really occurs as a card ID
-       (filter (fn [card] (some #{model-id} (-> card :dataset_query query/collect-card-ids))))))
+       (filter (fn [card] (some #{model-id} (-> card :dataset_query query/collect-card-ids))))
```
crisptrutski, line 109: Not your code, so don't need to action this.

nitpicks: 

- It bothers me that we're building the set each time inside the predicate (I think - maybe the Clojure compiler is smarter than I realize). Could hoist into a `model-id?` binding.
- I also think the threading here makes it harder to read - eyes need to start in the middle, scan right, then jump left again. One solution could be to go full-thread: 

```clojure
#(->> (:dataset_query %) query/collect-card-ids (some model-id?)))
```

src/metabase/api/card.clj
crisptrutski: Perhaps incidental sorting in the order the database is returning them?

Since you're otherwise preserving the order of the query, why not use an `:order-by` instead?

crisptrutski (commented):
Just skimmed through and replied to your comments first, going to read through the meat now.

src/metabase/api/card.clj
tsmacdonald: Yeah, I've done some more rigorous testing and there was a series of happy coincidences. `sort-by` definitely required; just transitioned it to `order-by` after some frustrating HoneySQL hiccups

tsmacdonald:
[Fixes #38926]

tsmacdonald:
Also [Fixes #39950]

crisptrutski (commented):
Sorry, nit city. Still need to zoom out and give high level review.

.clj-kondo/macros/metabase/models/field_usage_test.clj
```diff
@@ -0,0 +1,9 @@
+(ns macros.metabase.models.field-usage-test
+  (:require [macros.common]))
+
+(defmacro with-test-setup [& body]
+  `(let [~(macros.common/ignore-unused 'card-id)  1
+         ~(macros.common/ignore-unused 'table-id) 2
+         ~(macros.common/ignore-unused 'tax-id)   3
+         ~(macros.common/ignore-unused 'total-id) 4]
```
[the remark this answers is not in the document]
tsmacdonald: I don't want to be responsible for remembering the completely arbitrary order of four arguments—I'd end up transposing the tax-id and total-id by mistake or something.

There's quite a lot of prior art for anaphoric `with-`s in the test codebase

resources/migrations/001_update_migrations.yaml
```diff
@@ -5823,6 +5823,62 @@ databaseChangeLog:
...
+                  remarks: referenced field
+                  type: int
+                  constraints:
+                    nullable: false
+                    referencedTableName: metabase_field
+                    referencedColumnNames: id
+                    foreignKeyName: fk_field_usage_field_id
+                    deleteCascade: true
+              - column:
+                  remarks: Name of the used column, as it is in the query (i.e., the old name)
```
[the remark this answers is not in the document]
tsmacdonald: > When the name changes though, don't we create a new column anyway though? I don't understand how this could be "stale".

You're right, but I'm not filled with confidence about depending on that forever. If you think I'm being a coward I'm open to reconsidering though

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,55 @@
...
+   [clojure.string :as str]
+   [macaw.core :as mac]
+   [metabase.util :as u]
+   [toucan2.core :as t2]))
+
+(defn- normalize-name
+  ;; TODO: This is wildly naive and will be revisited once the rest of the plumbing is sorted out
+  [name]
+  (-> name
+      (str/replace "\"" "")
```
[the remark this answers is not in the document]
tsmacdonald: agree to disagree? :stuck_out_tongue:

resources/migrations/001_update_migrations.yaml
qnkhuat: how about field_ref? or field_reference?

resources/migrations/001_update_migrations.yaml
```diff
@@ -5824,6 +5824,89 @@ databaseChangeLog:
...
+                    nullable: false
+              - column:
+                  remarks: Name of the used table, as it is in the query
+                  name: table_name
+                  type: varchar(254)
+                  constraints:
+                    nullable: false
+              - column:
+                  remarks: Is this field usage still valid?
+                  name: is_currently_valid
```
qnkhuat, line 5878:
```suggestion
                  name: valid
```
nit: I think we usually just use an adjective for it, like report_card.archived or metabase_field.active

resources/migrations/001_update_migrations.yaml
```diff
@@ -5824,6 +5824,89 @@ databaseChangeLog:
...
+                  remarks: referenced field
+                  type: int
+                  constraints:
+                    nullable: false
+                    referencedTableName: metabase_field
+                    referencedColumnNames: id
+                    foreignKeyName: fk_field_usage_field_id
+                    deleteCascade: true
+              - column:
+                  remarks: Name of the used column, as it is in the query
```
qnkhuat, line 5865: is this the alias name?

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,83 @@
...
+
+  If you're invoking this from a test, be sure to turn on [[*parse-queries-in-test?*]]."
+  [{card-id :id, query :dataset_query :as card}]
+  (when (and (active?)
+             (= :native (:type query)))
+    ;; This feels inefficient at first glance, but the number of records should be quite small and doing some sort of
+    ;; upsert-or-delete would involve comparisons in Clojure-land that are more expensive than just "killing and
+    ;; filling" the records.
+    (t2/delete! :model/FieldUsage :card_id card-id)
+    (t2/insert! :model/FieldUsage (field-usages-for-card card))))
```
qnkhuat, line 83: should this operation be atomic?

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,83 @@
...
+      (str/replace "\"" "")
+      u/lower-case-en))
+
+(defn- fields-for-query
+  "Very naively selects Fields that could be used in the query. Improvements to this are planned for Q2 2024,
+  c.f. Milestone 3 of https://github.com/metabase/metabase/issues/36911"
+  [q db-id]
+  (try
+    (let [{:keys [columns tables]} (mac/query->components (mac/parsed-query q))]
+      (t2/query {:select [[:f/id :field_id] [:f/name :column_name] [:t/name :table_name]]
```
qnkhuat, line 42: Inconsistent use of `.` and `/` in this query. FWIW I prefer `.`

resources/migrations/001_update_migrations.yaml
qnkhuat: same with table_name, do we need it?

resources/migrations/001_update_migrations.yaml
qnkhuat: upon further reading, this is just a duplication of `metabase_field.active`, is there a reason why we need to keep track of it in `field_usage`? 

the only place I find it's used is to find stale queries, which you already joined with metabase_field

src/metabase/api/card.clj
```diff
@@ -125,6 +128,21 @@
   [_filter-option model-id]
   (cards-for-segment-or-metric :segment model-id))
 
+(defmethod cards-for-filter-option* :stale
+  [_]
+  (let [card-id->field-usages (group-by :card_id
+                                        (t2/select :model/FieldUsage
+                                                   {:select [:fu.* [:f.name :new_column_name] [:t.name :new_table_name]]
```
qnkhuat, line 135: 1. what's `new_column_name`, `new_table_name` is for?
2. We already joined with Field, Table here, so `metabase_table.name, metabase_field.name, metabase_field.active` can be selected from those tables, I don't see any reason we need to keep track of those in `field_usage`

qnkhuat (changes requested):
I have some questions about the schema of `field_usage`, it's not clear to me why we need to keep `column_name`, `table_name`, `is_currently_active` in field_usage, we can always join with field and table to get those.
Unless I'm missing something here

src/metabase/native_query_analyzer.clj
tsmacdonald: fixed. The `%lower` stuff needs to be that way, AFAIU

src/metabase/native_query_analyzer.clj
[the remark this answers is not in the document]
crisptrutski: Hopefully we can converge on a consistent style 😄 

(looks ominously over to the linter directory and plots against you)

.clj-kondo/macros/metabase/models/field_usage_test.clj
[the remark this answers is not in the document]
crisptrutski: Given that my onboarding tasks included getting rid of some anaphors, I assumed we were heading away from that style. I'm particularly biased against them as its impossible to teach Cursive to understand them. 

Re-ordering, it'd be fairly trivial to handle any ordering of any subset of known bindings. If you don't object to that approach I'll happily sneak in afterwards.

src/metabase/api/card.clj
```diff
@@ -103,13 +104,15 @@
                                                   [:or
                                                    [:like :c.dataset_query (format "%%card__%s%%" model-id)]
                                                    [:like :c.dataset_query (format "%%#%s%%" model-id)]]]]
-                        :where [:and [:= :m.id model-id] [:not :c.archived]]})
+                        :where [:and [:= :m.id model-id] [:not :c.archived]]
+                        :order-by [[[:lower :c.name] :asc]]})
```
qnkhuat, line 108: merge this with `order-by-name` ?

src/metabase/models/card.clj
```diff
@@ -521,6 +524,10 @@
       pre-update
       populate-query-fields
       maybe-populate-initially-published-at
+      ;; TODO: this should go in after-update once camsaul/toucan2#129 is fixed
+      ;; It's at the end for now so that all the before-update validations have a chance to run
+      ;; TODO the Second: No reason this couldn't be async, especially once it's in the after-update
```
qnkhuat, line 529: I don't think it should be async, though. Updating should be atomic; that's why before-update -> update -> after-update is executed in a transaction. Moving this out of the transaction will break that promise.

test/metabase/api/card_test.clj
```diff
@@ -299,6 +299,24 @@
...
+    (mt/with-temp [:model/Field        {active-field-id :id}    {:active true}
+                   :model/Field        {inactive-field-id :id}  {:active false}
+                   :model/Card         relevant-card            {:name "Card with stale query"}
+                   :model/Card         not-stale-card           {:name "Card whose columns are up to date"}
+                   :model/Card         irrelevant-card          {:name "Card with no QueryFields at all"}
+                   :model/QueryField   _                        {:card_id  (u/the-id relevant-card)
+                                                                 :field_id inactive-field-id}
+                   :model/QueryField   _                        {:card_id  (u/the-id not-stale-card)
+                                                                 :field_id active-field-id}]
+      (with-cards-in-readable-collection [relevant-card not-stale-card irrelevant-card]
```
qnkhuat, line 314: nit: I'd not bother with this macro at all, instead make the call with `:crowberto`

test/metabase/models/query_field_test.clj
```diff
@@ -0,0 +1,62 @@
...
+            (t2/delete! :model/QueryField :card_id card-id)))))))
+
+(defmacro ^:private with-test-setup
+  "Creates a new card that queries one column that exists (TOTAL) and one that does not (NOT_TAX). Anaphorically
+  provides `card-id`, `table-id`, `tax-id`, and `total-id`."
+  [& body]
+  `(do-with-test-setup (fn [{:keys [~'table-id ~'tax-id ~'total-id ~'card-id]}]
+                         ~@body)))
+
+(defn- trigger-parse!
```
qnkhuat, line 38: this is only used once, why not make the call inline?

test/metabase/models/query_field_test.clj
```diff
@@ -0,0 +1,62 @@
...
+(defn- query-fields-for-card
+  [card-id]
+  (t2/select-fn-set #(select-keys % query-field-keys) :model/QueryField
+                    :card_id card-id))
+
+(defn- do-with-test-setup [f]
+  (binding [query-analyzer/*parse-queries-in-test?* true]
+    (let [table-id (mt/id :orders)
+          tax-id   (t2/select-one-pk :model/Field :table_id table-id :name "TAX")
+          total-id (t2/select-one-pk :model/Field :table_id table-id :name "TOTAL")]
```
qnkhuat, line 20:
```suggestion
          tax-id   (mt/id :orders :tax)
          total-id (mt/id :orders :total)]
```

test/metabase/models/query_field_test.clj
```diff
@@ -0,0 +1,62 @@
...
+  [& body]
+  `(do-with-test-setup (fn [{:keys [~'table-id ~'tax-id ~'total-id ~'card-id]}]
+                         ~@body)))
+
+(defn- trigger-parse!
+  "Update the card to query two columns that do exist: TAX and TOTAL"
+  [card-id]
+  (t2/update! :model/Card card-id {:dataset_query (mt/native-query {:query "SELECT TAX, TOTAL FROM orders"})}))
+
+(deftest query-fields-created-by-queries-test
```
qnkhuat, line 43: I'd add 2 more tests:
- update the query to remove 1 QF
- update with a query that contains a field that does not exist in MB; this is to make sure we don't throw an exception while doing so.

src/metabase/native_query_analyzer.clj
```diff
@@ -0,0 +1,82 @@
...
+(defn- normalize-name
+  ;; TODO: This is wildly naive and will be revisited once the rest of the plumbing is sorted out
+  ;; c.f. Milestone 3 of the epic: https://github.com/metabase/metabase/issues/36911
+  [name]
+  (-> name
+      (str/replace "\"" "")
+      u/lower-case-en))
+
+(defn- field-ids-for-query
+  "Very naively selects the IDs of Fields that could be used in the query. Improvements to this are planned for Q2 2024,
```
qnkhuat, line 41: I was a bit surprised by how naive this is; I think that's fine for now, but we should make it clearer so people who want to use this are aware of this fact.

How about documenting it in this NS's docstring?

qnkhuat (approved):
Structurally LGTM, have some comments that I think you'll want to tackle, but they are non-blocking.

src/metabase/api/card.clj
tsmacdonald: Doesn't quite work since this uses `c.` as the table alias and `order-by-name` doesn't have that

test/metabase/api/card_test.clj
tsmacdonald: you're not wrong, but I was following the style of the surrounding tests

test/metabase/models/query_field_test.clj
tsmacdonald: When I first wrote it it was used in several places, but then with the `valid` column going away I killed a test. I think it adds a little clarity to the tests where it's used, and it will almost certainly be used more as the test suite grows

test/metabase/models/query_field_test.clj
tsmacdonald: neat!

PR #39708, opened 2024-03-06
Upgrade redux-auth-wrapper package

sloansparger:
### Description

Required dependency upgrade to unblock react v18 upgrade

### Checklist

Existing test coverage should be sufficient


npfitz (approved):
Looks good. I don't know that we have tests for all of the route guards (I don't see an e2e test for the admin guard), but we certainly cover some of them and they're all working as expected

PR #39706, opened 2024-03-06
Take upload settings as an argument for easier testing

crisptrutski:
### Description

This updates all the pure functions which use settings to take them as an argument.

Even if we're not similarly enamored with purity, perhaps we can agree on consistency? 😆

PR #39724, opened 2024-03-06
Graph gymnastics to handle type coercing and promotion

crisptrutski:
### Description

This follows on from the [int to float promotion](https://github.com/metabase/metabase/pull/39493) work, to keep float-to-int coercion working as well.

It turns out to be quite complex, but also served as a nice stress test of the `ordered-hierarchy`, and it was gratifying to see it handle it.

Two things that occur to me are:

1. Perhaps this complexity isn't worth it, and we should just let these `::float-or-int` values force a column promotion.
2. Getting this behavior might have been easier if we relaxed our DAG to a graph, instead of inserting two more abstract nodes to avoid a cycle. I'm not sure how those details would work out, but I have a hunch there's a way. Might just add a whole different kettle of complex lifeforms though.

### How to verify

See that the test suite is passing again - specifically this test:

```clojure
{:upload-type ::upload/int, :uncoerced "2.0", :coerced 2} ;; upload-test/append-type-coercion-test
```

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -605,7 +605,7 @@
 
 (def ^:private allowed-type-upgrades
   "A mapping of which types a column can be implicitly relaxed to, based on the content of appended values."
-  {::integer #{::float}})
+  {::int #{::float}})
 
 (defn- check-schema
   "Throws an exception if:
@@ -627,20 +627,23 @@
 (defn- matching-or-upgradable? [current-type relaxed-type]
   (or (= current-type relaxed-type)
       (when-let [f (allowed-type-upgrades current-type)]
-        (f relax-type))))
+        (f relaxed-type))))
 
-(defn- changed-field+new-types
+(defn- changed-field->new-type
   "Given some fields and old and new types, filter out fields with unchanged types, then pair with the new types."
   [fields old-types new-types]
   (let [new-if-changed #(when (not= %1 %2) %2)]
     (->> (map new-if-changed old-types new-types)
          (map vector fields)
-         (filter second))))
+         (filter second)
+         (into {}))))
 
-(defn- convert-columns! [_database _table field+new-types]
-  (doseq [[_field new-type] field+new-types]
-    (let [_column-type (column-type new-type)]
-      ::TODO)))
+(defn- alter-columns! [driver database table field->new-type]
+  (driver/alter-columns! driver (:id database) (table-identifier table)
+                         (m/map-kv (fn [field column-type]
+                                     [(keyword (:name field))
+                                      (driver/upload-type->database-type driver column-type)])
+                                   field->new-type)))
 
 (defn- append-csv!*
   [database table file]
@@ -669,8 +672,8 @@
                                         ;; we will instead throw an error when we try to parse as the old type
                                         (= relaxed-types new-column-types))
                                (let [fields (map normed-name->field normed-header)]
-                                 (->> (changed-field+new-types fields old-column-types relaxed-types)
-                                      (convert-columns! database table))))
+                                 (->> (changed-field->new-type fields old-column-types relaxed-types)
+                                      (alter-columns! driver database table))))
           ;; this will fail if any of our required relaxations were rejected.
           parsed-rows        (parse-rows new-column-types rows)]
       (try
@@ -682,7 +685,7 @@
         (driver/add-columns! driver
                              (:id database)
                              (table-identifier table)
-                             {auto-pk-column-keyword (conj (driver/upload-type->database-type driver ::auto-incrementing-int-pk))}
+                             {auto-pk-column-keyword (driver/upload-type->database-type driver ::auto-incrementing-int-pk)}
                              :primary-key [auto-pk-column-keyword]))
       (scan-and-sync-table! database table)
       (when create-auto-pk?
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -209,6 +209,16 @@
            (filter #((type->check %) trimmed))
            first))))
 
+(defn- relax-type
+  "Given an existing column type, and a value to insert into it, relax the type until it can parse the value."
+  [type->check current-type value]
+  (cond (nil? value) current-type
+        (nil? current-type) (value->type type->check value)
+        :else (let [trimmed (str/trim value)]
+                (->> (cons current-type (ancestors h current-type))
+                     (filter #((type->check %) trimmed))
+                     first))))
+
 (defn- row->value-types
   [type->check row]
   (map (partial value->type type->check) row))
@@ -614,13 +624,10 @@
       (let [error-message (extra-and-missing-error-markdown extra missing)]
         (throw (ex-info error-message {:status-code 422}))))))
 
-(defn- matching-or-upgradable? [old-types common-types]
-  (let [pairs (map vector old-types common-types)]
-    (every? (fn [[old common]]
-              (or (= old common)
-                  (when-let [f (allowed-type-upgrades old)]
-                    (f common))))
-            pairs)))
+(defn- matching-or-upgradable? [current-type relaxed-type]
+  (or (= current-type relaxed-type)
+      (when-let [f (allowed-type-upgrades current-type)]
+        (f relax-type))))
 
 (defn- changed-field+new-types
   "Given some fields and old and new types, filter out fields with unchanged types, then pair with the new types."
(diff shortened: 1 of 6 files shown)


crisptrutski:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/39724?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#39724** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#39493** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39493?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#39706](https://github.com/metabase/metabase/pull/39706) <a href="https://app.graphite.dev/github/pr/metabase/metabase/39706?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#39491** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39491?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

src/metabase/upload.clj
```diff
@@ -136,6 +164,15 @@
         ", " #"\d[\d \u00A0]*"
         ".’" #"\d[\d’]*"))))
 
+(defn- float-or-int-regex [number-separators]
+  (with-parens
+   (with-currency
+    (case number-separators
+      ("." ".,") #"\d[\d,]*\.0+"
+      ",." #"\d[\d.]*\,[0]+"
```
crisptrutski, line 172: I don't think we need the square braces around zero, I just cargo cult-ed them from the float regex.

src/metabase/upload.clj
```diff
@@ -136,6 +164,15 @@
...
+      ", " #"\d[\d \u00A0]*\,[0.]+"
+      ".’" #"\d[\d’]*\.[0.]+"))))
```
crisptrutski, line 174: I don't know why we allow additional decimal points, this might be duplicating a bug.

crisptrutski:
Tests are failing for Postgres - it appears to be generating an incorrect "ALTER COLUMN" statement.

src/metabase/upload.clj
```diff
@@ -45,33 +45,55 @@
...
 ;;
-;; `boolean-or-int` is a special type with two parents, where we parse it as a boolean if the whole
-;; column's values are of that type. additionally a column cannot have a boolean-or-int type, but
-;; a value can. if there is a column with a boolean-or-int value and an integer value, the column will be int
-;; if there is a column with a boolean-or-int value and a boolean value, the column will be boolean
-;; if there is a column with only boolean-or-int values, the column will be parsed as if it were boolean
+;;                 text
+;;                  |
+;;                  │
+;;              varchar-255────────┐
```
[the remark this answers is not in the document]
tsmacdonald: I was opening my mouth to say "this graph is becoming a maintenance nightmare", but the hairier it gets the gladder I am it exists…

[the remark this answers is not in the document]
tsmacdonald: I'm fairly sure it used to have `<pre><code>` tags around it so that it would render right in Marginalia-land. We should get those back. (Or if I'm misremembering, add them)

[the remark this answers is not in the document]
tsmacdonald: Oh, just saw the closing tags on L98. They should only surround the DAG, not the text, IMHO

src/metabase/upload.clj
tsmacdonald: Agreed, we should remove them

src/metabase/upload.clj
tsmacdonald: Was just about to ask that. L173 is also curious (can you really write Pi as `3,141.592[etc]`?)

test/metabase/upload_test.clj
```diff
@@ -111,37 +113,37 @@
...
-           ["2"          2              int-type]
-           ["-86"        -86            int-type]
-           ["9,986,000"  9986000        int-type]
-           ["9,986,000"  9986000        int-type "."]
-           ["9,986,000"  9986000        int-type ".,"]
-           ["9.986.000"  9986000        int-type ",."]
-           ["9’986’000"  9986000        int-type ".’"]
-           ["$0"         0              int-type]
-           ["-1"         -1             int-type]
+          [["0.0"        0              float-or-int-type"."]
```
tsmacdonald, line 116:
```suggestion
          [["0.0"        0              float-or-int-type "."]
```

test/metabase/upload_test.clj
```diff
@@ -155,7 +157,7 @@
            [".14"        ".14"          vchar-type ".,"] ;; TODO: this should be a float type
            ["0.14"       0.14           float-type ".,"]
            ["-9986.567"  -9986.567      float-type ".,"]
-           ["$2.0"       2              float-type ".,"]
+           ["$2.0"       2              float-or-int-type ".,"]
            ["$ 3.50"     3.50           float-type ".,"]
            ["-4300.23€"  -4300.23       float-type ".,"]
            ["£1,000.23"  1000.23        float-type]
```
tsmacdonald, line 163: If we can have `.` as a decimal separator there should be a test for it.

test/ordered_hierarchy_test.clj
```diff
@@ -24,14 +28,14 @@
     (is (nil? (parents h ::text)))
     (is (= [::text] (vec (parents h ::varchar-255))))
     (is (= [::float] (vec (parents h ::int))))
-    (is (= [::boolean ::int] (vec (parents h ::boolean-or-int))))))
+    (is (= [::boolean ::explicit-int] (vec (parents h ::boolean-or-int))))))
```
tsmacdonald, line 31: I get that ordered-hierarchy tests are orthogonal to CSV type-detection testing, but it doesn't seem right to shadow the type hierarchy. I would either:

1.  [preferred] come up with a new hierarchy just for testing. Polygons are a classic example (especially with the quadrilateral/rhombus/parallelogram/rectangle/square/trapezoid soup), or you could do animal species or musical instruments or modes of transportation or whatever your heart desires.

2. Recycle the hierarchy from CSV uploading.

tsmacdonald (commented):
Mostly like it, just a couple questions/comments

crisptrutski (commented):
After some discussion with @calherries we've decided to accept a small breaking change, i.e. update a single test case, rather than embrace this complexity yet.

It's not a one way day, I still think this is better behavior for the app, our approach could just benefit with more hammock time.

tsmacdonald:
@crisptrutski is this still ready for review? Happy to do so after you fix the merge conflicts

crisptrutski:
> @crisptrutski is this still ready for review? Happy to do so after you fix the merge conflicts

Feel free to review as is. I'm primarily waiting on @calherries to have hammock time here on whether we the underlying behavior is worth a more complex graph structure.

I've rebased the branch locally on https://github.com/metabase/metabase/pull/39793 which IMHO will make it harder to review since that's still up in the air as well, so decided to just leave this PR as is until the Hiccup PR is either merged or abandoned.

calherries:
another approach that would "work" is to special-case the `relax` function on uploads vs appends, so that

```
for uploads: concretize(relax(int, "1.0")) => float
for appends: concretize(relax(int, "1.0")) => int
```

This means either:
1. removing *float-or-int* from the type graph for uploads
2. having separate `type->check` functions for uploads vs appends. for uploads, "1.0" would be a float, and for appends "1.0" would be an int. But I think this suffers from the same type of problem as this PR proposes in that these are complications to try and fit a flawed model to the desired behaviour, instead of reworking the model.

calherries:
@crisptrutski you closing this?

crisptrutski:
> @crisptrutski you closing this?

Yip, I was just waiting for a chance to reply to some of your comments for posterity.

Re: the epicycle comment, I just wanted to remark on the irony that the gymnastics here were about *avoiding* cycles. That said, I'd had a hunch that somehow a restricted class of reversed edges might also work, and your idea of taking an extra argument to `concretize` proved to be just such a mechanism. Thanks!

We both were bothered by some of the complexity, but the bulk of these changes still made it through in the [alternate approach](https://github.com/metabase/metabase/pull/40275/files#diff-d72e1e62dbb726e6e563e8b2caec58e6cac2bbb9a650f10f0f42ecf2f2d00debL1793-R1794) in the end. From my perspective the major difference is putting a nice explicit, declarative layer on top, similar to how we introduce the explicit `abstract->concrete` map before. In the end I think we ended up with something even more complex, but with a better human interface - using maps instead of depending or edge orderings.

The only material part we left behind was using separate symbols for the value type classification (covering only integers without a decimal separator), and the concrete column type (accepting a superset which includes float literals that have a zero fractional component). As I laid out a bit in [this lengthy post](https://github.com/metabase/metabase/pull/40235#issuecomment-2004122233) ("If I had more time, I would have written a shorter letter."), I think that a precise relationship between the value type and column type objects that share overloaded symbols is a really good thing

So, essentially in many ways I still prefer this graph. It's still compatible with what we ended up implementing so far. 

Ignoring the actual names for a second, I thing having some symbol `int-value` (an abstract, i.e. non column-type) as distinct from `int-column` (a new kind of thing we haven't named, a non value-type) would be a good thing. A clean way to get past the awkward `*explicit-int*` naming would be to use separate symbols for all our types, e.g. `upload.value/int` and `upload.column/int`. This would allow us to keep the relationships both precise and intuitive relationships. It'd also remove ambiguity as to which type we mean by a given value being passed around.

Rather than being a "flawed model", I counter that maybe just a bit of an ivory tower. I can't think of a nice way to fork the symbols this way without adding a lot of code, but maybe it's something we'll come back to if the hacky overloading we've gone with ends up painting us into a corner. There might even be a really neat and concise way of expressing it.

crisptrutski:
Whoops, I also meant to comment on why we didn't go with any of the above suggestions of splitting based on on creating and appending contexts:

1. The semantics are column level, we want the "create" semantics when adding new columns to existing tables too.
2. There are more verbs coming, like "replace" and "merge", and they'll also have these "append" semantics.

Overloading the type-checking would also essentially add a third category (splitting value types into column-creating-value-types and column-promoting-value-types). Removing "float-or-type" would go even further and give these two categories a different hierarchical structure!

With a very slight semantic and structural tweak though that 1st suggestion worked a treat.

crisptrutski:
Whoops "and one more thing" 🍏 

I'm not sure how our current formulation would cope on making types "abstract" on a per-driver basis (i.e. does it support JSON). That might be the driver for another round of graphonimics.

PR #39728, opened 2024-03-06
Do not allow `undefined` when updating a user setting

nemanjaglumac:
As mentioned in [this comment](https://github.com/metabase/metabase/pull/39734#discussion_r1516072045), we define settings types as potentially undefined but we don't know if they are going to be used when setting/updating (dispatch) or when getting (fetch) a setting.

This PR removes the ability to erroneously update the setting with an `undefined` value. Backend would reject it anyway.
Inspiration for [this pattern](https://github.com/metabase/metabase/pull/39728/commits/712b1278838b59c7108c1883bb04037c071fe584): https://www.totaltypescript.com/tips/derive-a-union-type-from-an-object

Credits to @hypernurb for teaching me about the [mapping modifiers](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers).

### DIFF
--- a/frontend/src/metabase/redux/settings.ts
+++ b/frontend/src/metabase/redux/settings.ts
@@ -35,18 +35,13 @@ export const settings = createReducer(
   },
 );
 
-type RequiredUserSettings = {
-  [K in keyof UserSettings]-?: UserSettings[K];
-};
-
-type UserSettingsMap = {
-  [K in keyof RequiredUserSettings]: { key: K; value: RequiredUserSettings[K] };
-}[keyof RequiredUserSettings];
-
 export const UPDATE_USER_SETTING = "metabase/settings/UPDATE_USER_SETTING";
 export const updateUserSetting = createThunkAction(
   UPDATE_USER_SETTING,
-  function (setting: UserSettingsMap) {
+  function <K extends keyof UserSettings>(setting: {
+    key: K;
+    value: Exclude<UserSettings[K], undefined>;
+  }) {
     return async function (dispatch) {
       try {
         await SettingsApi.put(setting);
--- a/frontend/src/metabase/redux/settings.ts
+++ b/frontend/src/metabase/redux/settings.ts
@@ -35,13 +35,18 @@ export const settings = createReducer(
   },
 );
 
+type RequiredUserSettings = {
+  [K in keyof UserSettings]-?: UserSettings[K];
+};
+
+type UserSettingsMap = {
+  [K in keyof RequiredUserSettings]: { key: K; value: RequiredUserSettings[K] };
+}[keyof RequiredUserSettings];
+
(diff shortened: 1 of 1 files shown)


frontend/src/metabase/redux/settings.ts
```diff
@@ -35,13 +35,18 @@ export const settings = createReducer(
   },
 );
 
+type RequiredUserSettings = {
+  [K in keyof UserSettings]-?: UserSettings[K];
+};
+
+type UserSettingsMap = {
```
nemanjaglumac, line 42: image

frontend/src/metabase/redux/settings.ts
```diff
@@ -35,13 +35,18 @@ export const settings = createReducer(
   },
 );
 
+type RequiredUserSettings = {
+  [K in keyof UserSettings]-?: UserSettings[K];
```
nemanjaglumac, line 39: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers

frontend/src/metabase/redux/settings.ts
```diff
@@ -35,13 +35,18 @@ export const settings = createReducer(
...
+};
```
kamilmielnik, line 40: There is a utility type for that

```suggestion
type RequiredUserSettings = Required<UserSettings>;
```

frontend/src/metabase/redux/settings.ts
```diff
@@ -35,13 +35,18 @@ export const settings = createReducer(
...
+}[keyof RequiredUserSettings];
+
 export const UPDATE_USER_SETTING = "metabase/settings/UPDATE_USER_SETTING";
 export const updateUserSetting = createThunkAction(
   UPDATE_USER_SETTING,
-  function <K extends keyof UserSettings>(setting: {
-    key: K;
-    value: UserSettings[K];
-  }) {
+  function (setting: UserSettingsMap) {
```
kamilmielnik, line 49: We can do it without `RequiredUserSettings` and `UserSettingsMap`.
I think this is simpler:

```suggestion
  function <K extends keyof UserSettings>(setting: {
    key: K;
    value: Exclude<UserSettings[K], undefined>;
  }) {
```

frontend/src/metabase/redux/settings.ts
nemanjaglumac: TIL, thanks.

frontend/src/metabase/redux/settings.ts
nemanjaglumac: Thanks for the thorough review, as always @kamilmielnik !
Really appreciate it.

I've applied your suggested changes and the diff is indeed much smaller. 🎉 

As a side note:
I personally really like and even prefer the explicit intellisense preview that the previous solution provided. But this a very biased and a very personal preference.

iethree (approved):
❤️

Issue #39461, opened 2024-03-01
[Epic] Consolidate Settings (types)


nemanjaglumac:
This mini-epic aids efforts in consolidating somewhat messy `Settings` types. The bulk of this work is concentrated on the frontend. The epic was born as a result of multiple follow-ups after the work on #38986 where we introduced another user-local setting.

## Goal
Identify, separate and document different possible settings types. The source of truth is defined in a [`defsetting`](https://github.com/metabase/metabase/blob/master/src/metabase/models/setting.clj) macro.
```
    Visibility       | Who Can See It?              | Who Can Update It?
    ---------------- | ---------------------------- | --------------------
    :public          | The entire world             | Admins and Settings Managers
    :authenticated   | Logged-in Users              | Admins and Settings Managers
    :settings-manager| Admins and Settings Managers | Admins and Settings Managers
    :admin           | Admins                       | Admins
    :internal        | Nobody                       | No one (usually for env-var-only settings)
```

```[tasklist]
### User Settings
- [x] #39251
- [x] #39359 
- [ ] #39456
- [ ] https://github.com/metabase/metabase/pull/39469
- [ ] https://github.com/metabase/metabase/pull/39728
```

```[tasklist]
### Public Settings
- [ ] #39468
```

```[tasklist]
### Privileged Settings (Admin + Settings Manager)
- [ ] #39734
```

> [!Important]
> Do not define `internal` settings types on the FE.

PR #39734, opened 2024-03-06
Define PrivilegedSettings type

nemanjaglumac:
This PR further breaks down the ambiguous "Settings" type and identifies admin and setting manager related settings. This should help us identify the privileged settings more easily in the future, and make us use appropriate actions to dispatch these values.

A part of https://github.com/metabase/metabase/issues/39461.

### DIFF
--- a/frontend/src/metabase-types/api/settings.ts
+++ b/frontend/src/metabase-types/api/settings.ts
@@ -215,12 +215,12 @@ interface AdminSettings {
   "active-users-count"?: number;
   "deprecation-notice-version"?: string;
   "embedding-secret-key"?: string;
-  "query-caching-ttl-ratio": number;
   "query-caching-min-ttl": number;
+  "query-caching-ttl-ratio": number;
   "google-auth-auto-create-accounts-domain": string | null;
   "google-auth-configured": boolean;
-  "jwt-enabled"?: boolean;
   "jwt-configured"?: boolean;
+  "jwt-enabled"?: boolean;
   "premium-embedding-token": string | null;
   "saml-configured"?: boolean;
   "saml-enabled"?: boolean;
@@ -235,9 +235,9 @@ interface SettingsManagerSettings {
   "bcc-enabled?": boolean;
   "ee-openai-api-key"?: string;
   "openai-api-key": string | null;
-  "openai-organization": string | null;
-  "openai-model": string | null;
   "openai-available-models"?: OpenAiModel[];
+  "openai-model": string | null;
+  "openai-organization": string | null;
   "session-cookie-samesite": SessionCookieSameSite;
   "slack-app-token": string | null;
   "slack-files-channel": string | null;
--- a/frontend/src/metabase-types/api/settings.ts
+++ b/frontend/src/metabase-types/api/settings.ts
@@ -191,53 +191,61 @@ export interface OpenAiModel {
 export type HelpLinkSetting = "metabase" | "hidden" | "custom";
 
 interface InstanceSettings {
-  "active-users-count"?: number;
   "admin-email": string;
-  "bcc-enabled?": boolean;
-  "deprecation-notice-version"?: string;
-  "ee-openai-api-key"?: string;
-  "embedding-secret-key"?: string;
   "enable-embedding": boolean;
   "enable-nested-queries": boolean;
   "enable-query-caching"?: boolean;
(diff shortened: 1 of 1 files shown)


frontend/src/metabase-types/api/settings.ts
```diff
@@ -191,54 +191,62 @@ export interface OpenAiModel {
...
+  "google-auth-auto-create-accounts-domain": string | null;
+  "google-auth-configured": boolean;
+  "jwt-enabled"?: boolean;
+  "jwt-configured"?: boolean;
+  "premium-embedding-token": string | null;
+  "saml-configured"?: boolean;
+  "saml-enabled"?: boolean;
+  "show-database-syncing-modal": boolean;
+  "token-status": TokenStatus | null;
+  "version-info": VersionInfo | null;
```
nemanjaglumac, line 229: I don't think having a setting marked as potentially `undefined` is the best possible approach here. This interface is ambiguous in a sense that you don't know if it is going to be used in a setter or a getter.

If possible, we should strive to define types by looking only at the potential values for the particular setting. So when you use it in a setter, you don't risk allowing undefined.

And when used in a getter, the result of a fetch should be either Type or undefined.

Btw, this is not related to AdminSettings only.
It is a broader topic/problem that I think we need to solve.

frontend/src/metabase-types/api/settings.ts
```diff
@@ -191,54 +191,62 @@ export interface OpenAiModel {
...
   "uploads-table-prefix": string | null;
   "user-visibility": string | null;
+}
+
+interface AdminSettings {
+  "active-users-count"?: number;
+  "deprecation-notice-version"?: string;
+  "embedding-secret-key"?: string;
+  "query-caching-ttl-ratio": number;
+  "query-caching-min-ttl": number;
```
kamilmielnik, line 219: Alphabetical order:

```suggestion
  "query-caching-min-ttl": number;
  "query-caching-ttl-ratio": number;
```

frontend/src/metabase-types/api/settings.ts
```diff
@@ -191,54 +191,62 @@ export interface OpenAiModel {
...
+  "google-auth-auto-create-accounts-domain": string | null;
+  "google-auth-configured": boolean;
+  "jwt-enabled"?: boolean;
+  "jwt-configured"?: boolean;
```
kamilmielnik, line 223: Alphabetical order:

```suggestion
  "jwt-configured"?: boolean;
  "jwt-enabled"?: boolean;
```

frontend/src/metabase-types/api/settings.ts
```diff
@@ -191,54 +191,62 @@ export interface OpenAiModel {
...
   "show-static-embed-terms": boolean | null;
 }
 
+interface SettingsManagerSettings {
+  "bcc-enabled?": boolean;
+  "ee-openai-api-key"?: string;
+  "openai-api-key": string | null;
+  "openai-organization": string | null;
+  "openai-model": string | null;
+  "openai-available-models"?: OpenAiModel[];
```
kamilmielnik, line 240: Alphabetical order:

```suggestion
  "openai-api-key": string | null;
  "openai-available-models"?: OpenAiModel[];
  "openai-model": string | null;
  "openai-organization": string | null;
```

frontend/src/metabase-types/api/settings.ts
```diff
@@ -191,54 +191,62 @@ export interface OpenAiModel {
...
   "show-homepage-pin-message": boolean;
   "show-homepage-xrays": boolean;
   "site-uuid": string;
-  "slack-app-token": string | null;
-  "slack-files-channel": string | null;
-  "slack-token": string | null;
-  "slack-token-valid?": boolean;
   "subscription-allowed-domains": string | null;
-  "token-status": TokenStatus | null;
   "user-locale": string | null;
```
kamilmielnik, line 206: I'm guessing this belongs to `UserSettings`.

Not sure about `user-visibility`.

kamilmielnik (approved):
:+1:

frontend/src/metabase-types/api/settings.ts
nemanjaglumac: That's going to be the next PR.
We have a specific requirement for some of the settings called `:authenticated`. This one falls into that scope.

PR #39738, opened 2024-03-06
Adds property based test for `u/map-all`

escherize:
Adds a simple property-based test for a util function, as an example.

### DIFF
--- a/test/metabase/util_test.cljc
+++ b/test/metabase/util_test.cljc
@@ -1,14 +1,16 @@
 (ns ^:mb/once metabase.util-test
   "Tests for functions in `metabase.util`."
   (:require
+   #?@(:clj [[metabase.test :as mt]])
    [clojure.string :as str]
    [clojure.test :refer [are deftest is testing]]
    [clojure.test.check.clojure-test :refer [defspec]]
    [clojure.test.check.generators :as gen]
    [clojure.test.check.properties :as prop]
    [flatland.ordered.map :refer [ordered-map]]
-   [metabase.util :as u]
-   #?@(:clj [[metabase.test :as mt]])))
+   #_:clj-kondo/ignore
+   [malli.generator :as mg]
+   [metabase.util :as u]))
 
 (deftest ^:parallel add-period-test
   (is (= "This sentence needs a period."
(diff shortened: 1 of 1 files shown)


test/metabase/util_test.cljc
```diff
@@ -470,11 +472,20 @@
     (is (= [:c :e] (u/conflicting-keys {:a 1 :b 2 :c 3 :e nil}
                                        {:b 2 :c 4 :d 5 :e 6})))))
 
+(defn- join [& crumbs] (str/join ":" crumbs))
```
crisptrutski, line 475: As the idiot who wrote this original test, I probably should have just used `vector`:

```clojure
(deftest map-all-test
  (testing "map-all with 1 collection is just map"
    (is (= [[0] [1] [2] [3] [4]] (u/map-all vector (range 5)))))
  (testing "map-all works with 3 collections"
    (is (= [[0 0] [1 1] [2 2] [3 nil] [4 nil]] (u/map-all vector (range 5) (range 3)))))
  (testing "map-all works with higher arity"
    (is (= [[0 0 0] [1 1 1] [2 2 2] [3 nil 3] [nil nil 4]] (u/map-all vector (range 4) (range 3) (range 5))))))
```

crisptrutski: Since you don't use the the result of the function in any way in the property-based test, I'm not sure it's worth hoisting the function out even, use whatever n-arity built-in you want.

test/metabase/util_test.cljc
```diff
@@ -470,11 +472,20 @@
...
-      (is (= ["0:0:0" "1:1:1" "2:2:2" "3::3" "::4"] (u/map-all join (range 4) (range 3) (range 5)))))))
+  (testing "map-all with 1 collection is just map"
+    (is (= ["0" "1" "2" "3" "4"] (u/map-all join (range 5)))))
+  (testing "map-all works with 3 collections"
+    (is (= ["0:0" "1:1" "2:2" "3:" "4:"] (u/map-all join (range 5) (range 3)))))
+  (testing "map-all works with higher arity"
+    (is (= ["0:0:0" "1:1:1" "2:2:2" "3::3" "::4"] (u/map-all join (range 4) (range 3) (range 5))))))
+
+#?(:clj
+   (defspec map-all-returns-max-coll-input-size-test 1000
```
crisptrutski, line 486: Some other invariant that pop into my head:

- depends on f / consistent with map: `(map f ...)` == `(take n (map-all f ...))` where `n` = min count
- depends on every value in every collection: `(map-all (fnil + 0 0 0) ...)` == `(sum (map sum colls))`

Feels like together with your invariant this would give pretty good coverage.

.. there's also the option to just compare to a naive implementation, but slippery if it's complex enough to have its own bugs.

```clojure
(let [n   (apply max (map count colls))
      pad #(take n (concat % (repeat nil)))]
  (apply map f (map pad colls)))
```

test/metabase/util_test.cljc
```diff
@@ -470,11 +472,20 @@
...
+     (prop/for-all [xs (mg/generator [:sequential :int])
```
crisptrutski, line 487: How simple is it to extend this to `n pos?` collections?

crisptrutski (approved):

test/metabase/util_test.cljc
escherize: It is not hard, but it's a bit harder to read:

```clojure
(defspec map-all-returns-max-coll-input-size-test 100
  (prop/for-all [xss (mg/generator [:sequential {:min 1} ;; <-- map-all requires at least 1 collection
                                    [:sequential pos?]])]
    (let [expected (apply
                    #(apply (partial u/map-all vector) %&)
                    xss)]
      (= (count expected)
         (apply max (map count xss))))))
```

escherize: for the `pos?` bit, we could use :any, but that can sometimes create very large examples. So I've found that scalar values work best

test/metabase/util_test.cljc
escherize: That fnil example is a good idea, I'll write that one in too.

escherize: > naive implementation, but slippery if it's complex enough to have its own bugs.

Yeah, I have found that it can be tricky to specify complicated constraints without re-writing the code.

test/metabase/util_test.cljc
crisptrutski: Oh, by `pos?` I just meant that `{:min 1}` part. Sticking with `:int` is good.

PR #39746, opened 2024-03-07
Column info fixes

romeovs:
### Description

Adds some small fixes for the #39296 PR.

### How to verify

1. New Question → Sample Dataset
2. Hover the info icons on the tables and verify that the loading spinner only shows up when there is something to be loaded
3. Select People
4. Add an aggregation: Birth Date: Quarter of Year
5. Add another aggregation and hover the `Birth Date: Quarter of Year` item, the icon should not be overlapping the rest of the text

### Demo

Screenshot 2024-03-07 at 10 41 58

### DIFF
--- a/frontend/src/metabase/components/MetadataInfo/TableInfo/TableInfo.tsx
+++ b/frontend/src/metabase/components/MetadataInfo/TableInfo/TableInfo.tsx
@@ -64,7 +64,7 @@ function useDependentTableMetadata({
   fetchMetadata,
 }: Pick<AllProps, "tableId" | "table" | "fetchForeignKeys" | "fetchMetadata">) {
   const isMissingFields = !table?.numFields();
-  const isMissingFks = _.isEmpty(table?.fks);
+  const isMissingFks = table?.fks === undefined;
   const shouldFetchMetadata = isMissingFields || isMissingFks;
   const [hasFetchedMetadata, setHasFetchedMetadata] = useState(
     !shouldFetchMetadata,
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/components/MetadataInfo/TableInfo/TableInfo.tsx
```diff
@@ -113,7 +114,9 @@ export function TableInfo({
       <MetadataContainer>
         <Fade visible={!hasFetchedMetadata}>
           <AbsoluteContainer>
-            <LoadingSpinner size={24} />
+            <Delayed delay={400}>
+              <LoadingSpinner size={24} />
+            </Delayed>
```
kamilmielnik, line 119: The problem is that `hasFetchedMetadata` is always initially `false` because `isMissingFks` check is always `true` for tables without FKs.
If we could determine whether FKs need to be fetched for a given table, then we could fix this without additional components. I don't know if we have this information.
But the least we could do is to make sure that FKs are not refetched for a given table if they were already fetched once.

kamilmielnik (commented):
Flex fixes look good :+1: 
I think we should explore a different approach for the spinner issue.

frontend/src/metabase/components/MetadataInfo/TableInfo/TableInfo.tsx
romeovs: Yes that was my initial reaction too but I'm afraid we don't have that information (ie. the state for "we haven't fetched fk's yet" and "we have fetched fk's but there are none" is the same), so I don't think we have this information.

The request to fetch Fk's is cached higher up the chain AFAICT so there are no superfluous requests going out.

kamilmielnik (approved):
:+1:

PR #39762, opened 2024-03-07
Make tabs e2e test stable again

uladzimirdev:
### Description

Fixes test flakiness as we could have 2 requests to `/query` (first canceled, second is actual) depends on the network speed

### How to verify

run it locally 

[stress test](https://github.com/metabase/metabase/actions/runs/8191547958)

### DIFF
--- a/e2e/test/scenarios/dashboard/tabs.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/tabs.cy.spec.js
@@ -402,8 +402,9 @@ describe("scenarios > dashboard > tabs", () => {
     });
   });
 
-  it("should only fetch cards on the current tab", { tags: "@flaky" }, () => {
+  it("should only fetch cards on the current tab", () => {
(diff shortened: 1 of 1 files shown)


ranquild (approved):

e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -416,16 +417,22 @@ describe("scenarios > dashboard > tabs", () => {
...
+
+    cy.wait("@cardQuery");
+
     saveDashboard();
 
     cy.wait("@saveDashboardCards").then(({ response }) => {
       cy.wrap(response.body.dashcards[1].id).as("secondTabDashcardId");
     });
 
+    // it's possible to have two requests firing (but first one is canceled before running second)
```
alxnddr, line 429: I think our specs should not adapt to different environemnts, why not to wait until requests are resolved? https://github.com/metabase/metabase/pull/39783#top

alxnddr (changes requested):

PR #39766, opened 2024-03-07
Only return helpText for supported features

### LINKED ISSUE #35156: Let's not present usage pop-ups for unsupported functions in custom expression interface
Issue #34993 requires us to fix `percentile` support for mariadb, saying that user tried to use percentile in __the custom expression interface__ but translation to sql which they've got was not right. However we do not support `:percentile-expressions` for mariadb on the backend. Let alone, mariadb supports only window version on `percetile_disc` and currently query builder also does not support use of window functions (related issue is #9393).

I believe the fact, that they were seeing the usage pop-up after writing in `percentile(` made them think it is something we are able to correctly compute from query builder and that's false. 

1. Not showing that pop-up could make us less prone to get bug reports for bugs which are not bugs.
2. Same case for completion of arguments to functions that are not supported.

With regards to the completion of function name -- This is something we do not do for unsupported functions and that's right.

Attached screenshots contain (1) no completion when writing `percentile` and (2) pop-up  after entring in `percentile(` and (3) completion of arguments that should imo happen only for supported functions.

Screenshot 2023-10-27 at 16 02 40

Screenshot 2023-10-27 at 16 02 46

Screenshot 2023-10-27 at 16 02 56

---

And super ideal behavior would be to not even enable user to generate request for query containing use of a function that is not supported by a driver or make backend return eg. 418 when request like this is processed.

romeovs:
Closes https://github.com/metabase/metabase/issues/35156

### Description

Avoid showing help text for unsupported functions.

### How to verify

1. New question → Mariadb
2. Select a table
3. Add a custom expression filter
4. In the textbox type: `percentile(`
5. No help text should show

Doing the same for a Postgres database, should show the help text because it supports percentiles.

### Demo

Screenshot 2024-03-07 at 15 48 04

### DIFF
--- a/frontend/src/metabase-lib/expressions/suggest.ts
+++ b/frontend/src/metabase-lib/expressions/suggest.ts
@@ -82,7 +82,12 @@ export function suggest({
       if (name && database) {
         const helpText = getHelpText(name, database, reportTimezone);
         if (helpText) {
-          return { suggestions, helpText };
+          const clause = MBQL_CLAUSES[helpText?.name];
+          const isSupported =
+            !clause || database?.hasFeature(clause.requiresFeature);
+          if (isSupported) {
+            return { suggestions, helpText };
+          }
         }
       }
     }
--- a/frontend/src/metabase-lib/expressions/suggest.unit.spec.ts
+++ b/frontend/src/metabase-lib/expressions/suggest.unit.spec.ts
@@ -1,5 +1,6 @@
 import _ from "underscore";
 
+import { createMockMetadata } from "__support__/metadata";
 import * as Lib from "metabase-lib";
 import {
   SAMPLE_DATABASE,
@@ -8,6 +9,7 @@ import {
 } from "metabase-lib/test-helpers";
 import type { DatasetQuery, Join } from "metabase-types/api";
 import {
+  createSampleDatabase,
   ORDERS,
   ORDERS_ID,
   REVIEWS,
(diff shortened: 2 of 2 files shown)


frontend/src/metabase-lib/expressions/suggest.unit.spec.ts
```diff
@@ -278,6 +280,27 @@ describe("metabase/lib/expression/suggest", () => {
         });
       });
 
+      it("should not provide help text an unsupported function", () => {
```
kamilmielnik, line 283:
```suggestion
      it("should not provide help text for an unsupported function (metabase#39766)", () => {
```

frontend/src/metabase-lib/expressions/suggest.ts
```diff
@@ -82,7 +82,12 @@ export function suggest({
       if (name && database) {
         const helpText = getHelpText(name, database, reportTimezone);
         if (helpText) {
-          return { suggestions, helpText };
+          const clause = MBQL_CLAUSES[helpText?.name];
+          const isSupported =
+            !clause || database?.hasFeature(clause.requiresFeature);
```
kamilmielnik, line 87: Shouldn't this be the other way around?

```suggestion
            clause && database?.hasFeature(clause.requiresFeature);
```

romeovs: I'm being conservative here: we still want to return help text even when the clause is not defined in `MBQL_CLAUSES` but _is_ defined in [`HELPER_TEXT_STRINGS`](https://github.com/metabase/metabase/blob/master/frontend/src/metabase-lib/expressions/helper-text-strings.ts#L39).

So if there is no clause found (ie. `!clause`) we allow the help text to be returned.

kamilmielnik (approved):

Issue #39291, opened 2024-02-29
[Testing plan] Add Extract action to chill mode column headings


kamilmielnik:
Testing plan for #38964

### Drills (CLJS)
- **Date** column **header**: can extract **day of month / day of week / month / quarter / year**
- **Date** column **header**: cannot extract **hour of day**
- **Time** column **header**: can extract **hour of day**
- **Time** column **header**: cannot extract **day of month / day of week / month / quarter / year**
- **Timestamp** column **header**: can extract **hour of day / day of month / day of week / month / quarter / year**
- **Date** / **Time** / **Timestamp** column **cell**: cannot extract any columns
- Drills are offered in the following order: hour of day, day of month, day of week, month, quarter, year
- Drills are not available for pivot tables
- Drills are not available for non-date/time columns

### Extracted columns (e2e)

- Extracted column is added to the right of the source column
- Extracted column can be edited in the GUI editor
- Extracted column name uses instance language
- Adding the same extraction twice is possible, and names of columns are unique (e.g. `Year`, `Year (2)`)
- Hour of day
  - Column name should default to "Hour of day"
  - The values should be [formatted](https://www.figma.com/file/yl8Yy4iMSpaWzjuNoEjBZu/Add-Extract-action-to-chill-mode-column-headings?type=design&node-id=5916-4999&mode=design&t=SfMuJuDm2kKMGR9F-4) as hours according to ["Time style" setting](https://www.metabase.com/docs/latest/configuring-metabase/localization#dates-and-times) and instance language
- Day of month
  - Column name should default to "Day of month"
  - The values should be [integers](https://www.figma.com/file/yl8Yy4iMSpaWzjuNoEjBZu/Add-Extract-action-to-chill-mode-column-headings?type=design&node-id=5916-4999&mode=design&t=SfMuJuDm2kKMGR9F-4) (formatted according to instance language? think: Arabic, Thai, etc.)
- Day of week
  - Column name should default to "Day of week"
  - The values should be [formatted](https://www.figma.com/file/yl8Yy4iMSpaWzjuNoEjBZu/Add-Extract-action-to-chill-mode-column-headings?type=design&node-id=5916-4999&mode=design&t=SfMuJuDm2kKMGR9F-4) as names of days of the week according to ["Abbreviate names of days and months" setting](https://www.metabase.com/docs/latest/configuring-metabase/localization#dates-and-times) and instance language
- Month
  - Column name should default to "Month"
[...]

PR #39162, opened 2024-02-26
"Extract column" drill thru

ranquild:
Epic https://github.com/metabase/metabase/issues/38964

Screenshot 2024-02-29 at 16 59 48
Screenshot 2024-02-29 at 16 59 51

PR #39770, opened 2024-03-07
Extract column e2e

ranquild:
Epic https://github.com/metabase/metabase/pull/39162
Testing plan https://github.com/metabase/metabase/issues/39291

Checks:
- Each option type (unit)
- Breakout and aggregation columns
- Editing the generated expression in the notebook editor
- Duplicate column names
- Using localized values for temporal units

Doesn't check:
- Column is appended to the right - that's not the case now
- Different viz settings for each option - that's not the case now and I'm not sure if everything from the testing plan is correct about this


e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
...
+    cy.intercept("POST", "/api/dataset").as("dataset");
+    cy.intercept("POST", "/api/card/*/query").as("cardQuery");
+  });
+
+  describe("date columns", () => {
+    describe("should add a date expression for each option", () => {
+      DATE_CASES.forEach(({ option, value }) => {
+        it(option, () => {
+          openOrdersTable({ limit: 1 });
+          cy.wait("@dataset");
```
nemanjaglumac, line 71:
```suggestion
```

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
...
+      extractColumnAndCheck({
+        column: "Min of Created At: Default",
+        option: "Year",
+        value: "2,022",
+      });
+    });
+
+    it("should handle duplicate expression names", () => {
+      openOrdersTable({ limit: 1 });
+      cy.wait("@dataset");
```
nemanjaglumac, line 101:
```suggestion
```

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
...
+      extractColumnAndCheck({
+        column: "Created At",
+        option: "Hour of day",
+        newColumn: "Hour of day_2",
+      });
+    });
+
+    it("should be able to modify the expression in the notebook editor", () => {
+      openOrdersTable({ limit: 1 });
+      cy.wait("@dataset");
```
nemanjaglumac, line 116:
```suggestion
```

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
...
+      cy.findByRole("gridcell", { name: "2,027" }).should("be.visible");
+    });
+
+    it("should use current user locale for string expressions", () => {
+      cy.request("GET", "/api/user/current").then(({ body: user }) => {
+        cy.request("PUT", `/api/user/${user.id}`, { locale: "de" });
+      });
+      openOrdersTable({ limit: 1 });
+      cy.wait("@dataset");
```
nemanjaglumac, line 137:
```suggestion
```

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
...
+        option: "Year",
+        value: "2,025",
+      });
+      cy.wait("@dataset");
+      openNotebook();
+      getNotebookStep("expression").findByText("Year").click();
+      enterCustomColumnDetails({ formula: "+ 2" });
+      popover().button("Update").click();
+      visualize();
+      cy.wait("@dataset");
```
nemanjaglumac, line 128:
```suggestion
```

nemanjaglumac (commented):
`openOrdersTable` and `visualize` helpers are already waiting for the `@dataset`. There's no need to do it additionally in the test.

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
...
+      cy.findByRole("gridcell", { name: "2,027" }).should("be.visible");
+    });
+
+    it("should use current user locale for string expressions", () => {
```
kamilmielnik, line 132: I think we should also check "Quarter" translations, it's something that could easily get overlooked.
For example in Polish and Russian quarters should be formatted as Roman numerals.

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -0,0 +1,157 @@
+import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
```
kamilmielnik, line 1: It'd be great if new e2e tests were implemented in TS.

kamilmielnik (commented):
> Different viz settings for each option - that's not the case now and I'm not sure if everything from the testing plan is correct about this

Let's discuss this in [Slack](https://metaboat.slack.com/archives/C0645JP1W81/p1709882492209779)

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
ranquild: Apparently quarter names are always not localized neither on the FE or BE `(t/format "'Q'Q" t)`

ranquild:
@nemanjaglumac for the record, this is not how "intercept" in cypress works. The problem arises when you have multiple intercepts for the same url:
```
cy.intercept("POST", "/api/dataset").as("dataset");
openOrdersTable(); // calls intercept for the same URL but different alias
// ... some action
cy.wait("@dataset"); // <-- this call would wait for the /api/dataset request made by `openOrdersTable` and will return immediately
```

We can solve this issue by using `intercept` closer to the action, and using unique aliases.

kamilmielnik (approved):

PR #39778, opened 2024-03-07
Detect the use of unsupported features in the expression editor

### LINKED ISSUE #39773: Show better error message when function is not supported by the database
### Describe the bug

Currently when a function is not supported by the database, we don't show an error message.
In some cases we even show an error that is about calling the function incorrectly, even though it does not exist.

### To Reproduce

1. New Question → MariaDB instance
2. Select a table
3. Add an aggregation and select Custom Expression
4. Type `percentile(1, 0.9)`
5. Tab out of the expression field
6. No error is show
7. Now type `percentile(1)` in the expression field and tab out
8. An error is displayed: `Function Percentile expects 2 arguments`, even though this database does not support `percentile`

### Expected behavior

There should be an error when a function is called that is not supported

### Logs

_No response_

```JSON
{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:123.0) Gecko/20100101 Firefox/123.0",
    "vendor": ""
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "11.0.20.1+1",
    "java.version": "11.0.20.1",
    "java.vm.version": "11.0.20.1+1",
    "os.name": "Mac OS X",
    "os.version": "14.2.1",
    "user.language": "en",
    "user.timezone": "Europe/Amsterdam"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "mysql",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "date": "2024-03-07",
      "src_hash": "e35fc04830cd05dc48aa5d85ab07fc5c2eade141",
      "tag": "v0.48.1-SNAPSHOT",
      "hash": "a5261dc"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

P3

### Additional context

This is related to https://github.com/metabase/metabase/issues/35156

romeovs:
Closes https://github.com/metabase/metabase/issues/39773

### Description

This adds an extra check to the diagnostics to make sure a functions is supported by the database in the custom expression editor.

### How to verify

1. New question → Sample Dataset → Orders (or any database that does not support `percentile-aggregations`)
2. Click "Pick the metric you want to see" → Custom Expression
3. In the expression editor, type `percentile([Discount], 0.9)`
4. Tab out of the expression editor text field
5. An error should show up: `Unsupported function percentile`

### Demo

Screenshot 2024-03-07 at 16 49 50

### Checklist

### DIFF
--- a/e2e/test/scenarios/custom-column/cc-data-type.cy.spec.js
+++ b/e2e/test/scenarios/custom-column/cc-data-type.cy.spec.js
@@ -5,7 +5,7 @@ import {
   popover,
   enterCustomColumnDetails,
   filter,
-  openOrdersTable,
+  startNewQuestion,
   visualize,
   getNotebookStep,
 } from "e2e/support/helpers";
@@ -15,6 +15,8 @@ const { ORDERS_ID, PEOPLE_ID, PRODUCTS_ID } = SAMPLE_DATABASE;
 describe("scenarios > question > custom column > data type", () => {
   beforeEach(() => {
     restore();
+    restore("postgres-12");
+
     cy.signInAsAdmin();
   });
 
@@ -38,7 +40,12 @@ describe("scenarios > question > custom column > data type", () => {
   });
 
   it("should understand date functions", () => {
-    openOrdersTable({ mode: "notebook" });
+    startNewQuestion();
+    popover().within(() => {
+      cy.findByText("Raw Data").click();
+      cy.findByText("QA Postgres12").click();
+      cy.findByText("Orders").click();
+    });
 
     addCustomColumns([
       { name: "Year", formula: "year([Created At])" },
@@ -132,11 +139,11 @@ function addCustomColumns(columns) {
     if (index) {
       getNotebookStep("expression").icon("add").click();
     } else {
-      cy.findByText("Custom column").click();
+      cy.findByLabelText("Custom column").click();
     }
 
     enterCustomColumnDetails(column);
-    cy.button("Done").click();
+    cy.button("Done").click({ force: true });
   });
 }
 
--- a/frontend/src/metabase-lib/expressions/resolver.js
+++ b/frontend/src/metabase-lib/expressions/resolver.js
@@ -69,18 +69,19 @@ const isCompatible = (a, b) => {
 };
 
 /**
- *
- * @param {import("./pratt").Expr} expression
- * @param {string} type
- * @param {?(kind: string, name: string, node: import("./pratt").Node) => void} fn
- * @param {Database | null} database
+ * @param {{
+ *   expression: import("./pratt").Expr
+ *   type?: string
+ *   fn?: ?(kind: string, name: string, node: import("./pratt").Node) => void
+ *   database?: Database | null
+ * }} options
  */
-export function resolve(
+export function resolve({
   expression,
   type = "expression",
   fn = undefined,
   database = undefined,
-) {
+}) {
   if (Array.isArray(expression)) {
     const [op, ...operands] = expression;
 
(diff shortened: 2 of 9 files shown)


frontend/src/metabase-lib/expressions/resolver.js
```diff
@@ -152,6 +158,11 @@ export function resolve(expression, type = "expression", fn = undefined) {
     if (!clause) {
       throw new ResolverError(t`Unknown function ${op}`, expression.node);
     }
+
+    if (database && !database.hasFeature(clause)) {
```
kamilmielnik, line 162: `clause` is an object, `Database.prorotype.hasFeature` accepts a `string`.
Currently it's not possible to use any function :see_no_evil: 

```suggestion
    if (database && !database.hasFeature(clause.displayName)) {
```

frontend/src/metabase-lib/expressions/resolver.js
```diff
@@ -73,8 +73,14 @@ const isCompatible = (a, b) => {
...
  * @param {string} type
  * @param {?(kind: string, name: string, node: import("./pratt").Node) => void} fn
+ * @param {Database | null} database
  */
-export function resolve(expression, type = "expression", fn = undefined) {
+export function resolve(
+  expression,
+  type = "expression",
+  fn = undefined,
+  database,
```
kamilmielnik, line 82: nit: optional arguments should go last

Maybe in this case we'd be better off with a param object instead of N arguments.

kamilmielnik (changes requested):
There's 1 mistake and code has not been updated everywhere (see failing `type-check` job).

frontend/src/metabase-lib/expressions/resolver.js
romeovs: My brain must've been turned of when I wrote this. I've fixed this and added database to the recursive `resolve` calls too.

frontend/src/metabase-lib/expressions/resolver.js
```diff
@@ -69,12 +69,19 @@ const isCompatible = (a, b) => {
...
+ *   fn?: ?(kind: string, name: string, node: import("./pratt").Node) => void
+ *   database?: Database | null
+ * }} options
  */
-export function resolve(expression, type = "expression", fn = undefined) {
+export function resolve({
+  expression,
+  type = "expression",
+  fn = undefined,
+  database = undefined,
```
kamilmielnik, line 83: Should we make `database` required and pass it in all of the calls?
e2e tests for custom columns are failing, this might be related.

romeovs: Not sure what to do here: the tests are broken because they incorrectly assumed that the built in database supports the `convert-timezone` features, whereas it does not.

I think the only correct way to fix this is to remove that case from the test, but then we don't have a test for those functions anymore.

kamilmielnik (approved):
I'm not sure if it's ready for review but LGTM :+1:

e2e/test/scenarios/custom-column/cc-data-type.cy.spec.js
```diff
@@ -132,11 +139,11 @@ function addCustomColumns(columns) {
     if (index) {
       getNotebookStep("expression").icon("add").click();
     } else {
-      cy.findByText("Custom column").click();
+      cy.findByLabelText("Custom column").click();
     }
 
     enterCustomColumnDetails(column);
-    cy.button("Done").click();
+    cy.button("Done").click({ force: true });
```
kamilmielnik, line 146: It should just work. Why do we need `force: true`?

romeovs: Because the help text popover might still be covering the button.

I could add timeouts to wait for it to go away, but I think this is more robust.

kamilmielnik: It might be a good idea to [stress-test](https://github.com/metabase/metabase/actions/workflows/e2e-stress-test-flake-fix.yml) this change.

PR #39780, opened 2024-03-07
Disclosure triangle and Fullpage modal transitions

### LINKED ISSUE #39764: Disclosure Triangle

### LINKED ISSUE #39772: Fullpage Modal

npfitz:
Closes https://github.com/metabase/metabase/issues/39764
Closes https://github.com/metabase/metabase/issues/39772

### Description
Grouping a couple of these because they're both quite small. Adjusting the `<DisclosureTriangle />` to be powered by CSS animations, and the `<FullPageModal/>` to use mantine transitions. Both are simple adjustments that remove the need for the `react-motion` package. The only change that may require a look is how the opened state is now handled in FullPageModal. We need to mount the `Transition` component with the "out" styles before changing them if we want them to animate in. Because of how the modals are mounted and unmounted, this is handled by setting the state to false in the constructor, then setting it to true in `componentDidMount`

### How to verify
Disclosure Triangles can be found in the JWT, SAML, and LDAP auth forms.
The only instance of a full page modal being used that I could find was creating alerts for Questions (in the future, this could probably be migrated to a Mantine modal with fullPage set to true)

### Demo
Fullpage Modal:
chrome_DcgbwIJKRa

Disclosure Triangle:
chrome_RTTPFMwMQS

### Checklist

### DIFF
--- a/frontend/src/metabase/components/Modal/FullPageModal.tsx
+++ b/frontend/src/metabase/components/Modal/FullPageModal.tsx
@@ -99,11 +99,7 @@ export class FullPageModal extends Component<
                 handleDismissal={this.handleDismissal}
                 closeOnClickOutside={this.props.closeOnClickOutside}
               >
-                <div
-                  className="full-height relative scroll-y"
-                  // style={motionStyle}
-                  style={styles}
-                >
+                <div className="full-height relative scroll-y" style={styles}>
                   {getModalContent({
                     ...this.props,
                     fullPageModal: true,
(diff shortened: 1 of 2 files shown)


frontend/src/metabase/components/Modal/FullPageModal.tsx
```diff
@@ -99,7 +101,8 @@ export class FullPageModal extends Component<
               >
                 <div
                   className="full-height relative scroll-y"
-                  style={motionStyle}
+                  // style={motionStyle}
```
sloansparger, line 104: you can delete this line

sloansparger (approved):
approved, but delete that unused comment first plz

PR #39783, opened 2024-03-07
Fixes tabs cy spec flake

alxnddr:
### Description

We cancel cards requests when switching to other tabs if they haven't been loaded. The spec has been updated to wait until cards data is loaded (visible) before changing tabs.

### How to verify

To reproduce the flake on powerful machines put `cy.pause()` on line 411. Once it gets there, enable throttling Fast 3G and resume the spec. It should not flake

### DIFF
--- a/e2e/test/scenarios/dashboard/tabs.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/tabs.cy.spec.js
@@ -377,6 +377,7 @@ describe("scenarios > dashboard > tabs", () => {
 
   it("should only fetch cards on the current tab", () => {
     cy.intercept("PUT", "/api/dashboard/*").as("saveDashboardCards");
+    cy.intercept("POST", "/api/card/*/query").as("cardQuery");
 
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/dashboard/tabs.cy.spec.js
```diff
@@ -415,16 +415,21 @@ describe("scenarios > dashboard > tabs", () => {
     cy.get("@firstTabQuery").should("have.been.calledOnce");
     cy.get("@secondTabQuery").should("not.have.been.called");
 
+    getDashboardCard().findByText("User ID");
```
adam-james-v, line 418: maybe add a comment explaining why this is done? Just like you've described in the PR?

adam-james-v (approved):
Looks good, perhaps just add a comment in the test. Non-blocking though :)

alxnddr:
There is a race condition has to be fixed on the initial dashboard opening

PR #39786, opened 2024-03-07
Remove `user-locale` from settings

nemanjaglumac:
`user-locale` doesn't exist anymore in settings.
You can easily verify that by greping the code base for `defsetting user-locale` or simply by observing the `GET /api/session/properties` response.


rafpaf (approved):
Should we also change `src/metabase/util/i18n.clj`? I see references there to "user-locale".

nemanjaglumac:
> Should we also change src/metabase/util/i18n.clj? I see references there to "user-locale".

Thanks for the review, @rafpaf 
I believe we don't have to touch that file, but I will double check with @metabase/backend-developers and will follow-up if needed.

PR #39788, opened 2024-03-07
tweak default modal spacing

kdoh:
Sets an updated default of "lg" for modal spacing.

Note something is wrong with fonts in storybook right now, so ignore that part and just concentrate on the spacing for the purposes of this PR

Old default:
Screenshot 2024-03-07 at 12 27 57 PM

New default:
Screenshot 2024-03-07 at 12 27 31 PM

### DIFF
--- a/frontend/src/metabase/ui/components/overlays/Modal/Modal.styled.tsx
+++ b/frontend/src/metabase/ui/components/overlays/Modal/Modal.styled.tsx
@@ -29,6 +29,7 @@ export const getModalOverrides = (): MantineThemeOverride["components"] => ({
   ModalHeader: {
     defaultProps: {
       p: DEFAULT_MODAL_SPACING,
+      pb: "sm",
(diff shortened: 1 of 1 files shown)


iethree:
@kdoh can you check this out with some body content? The header padding goes all the way around, so at larger padding values puts a lot of space between the header and the body content. 

I don't have a dog in this fight, I just want to make sure it's "right" from a design standpoint so I can just use it and not have to tweak it every time.

kdoh:
@iethree. Yep good catch on the even-ness. I set it so the header accommodates for that. This looks right to my eye in terms of relation between everything. XL is overly spaced out I think. (again excuse the type being busted in storybook)

Screenshot 2024-03-07 at 12 41 01 PM

ranquild (approved):

iethree (approved):
❤️

PR #39793, opened 2024-03-07
Use hiccup syntax for value-type DAG

crisptrutski:
### Description

Introduce some sugar - it's good for you.

This change switches from low-level `derive` calls to a data-driven hiccup expression. While these data structures are not trees, if we avoid listing the children of every node whenever it is repeated, it seems like we get something quite readable.[^1]

It's arguably easier to scan the structure this way, objectively, especially when it comes to noting the ordering of the child edges.

A limitation of this representation is that we are no longer able to order the edges from each child to their parents.

[^1]: Another option would be to repeat the children exactly each time, but this could add a lot of bloat. If we're not repeating however, perhaps it's worth adding some indication that that this is a repeated node, and not a leaf.

### DIFF
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -33,7 +33,7 @@
    [metabase.util.i18n :refer [tru]]
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
-   [metabase.util.ordered-hierarchy :as ordered-hierarchy :refer [derive make-hierarchy]]
+   [metabase.util.ordered-hierarchy :as ordered-hierarchy :refer [make-hierarchy]]
    [toucan2.core :as t2])
   (:import
    (java.io File)))
@@ -80,17 +80,14 @@
   "This hierarchy defines a relationship between value types and their specializations.
   We use an [[metabase.util.ordered-hierarchy]] for its topological sorting, which simplify writing efficient and
   consistent implementations for of our type inference, parsing, and relaxation."
-  (-> (make-hierarchy)
-      (derive ::boolean-or-int ::boolean)
-      (derive ::boolean-or-int ::int)
-      (derive ::auto-incrementing-int-pk ::int)
-      (derive ::int ::float)
-      (derive ::date ::datetime)
-      (derive ::boolean ::varchar-255)
-      (derive ::offset-datetime ::varchar-255)
-      (derive ::datetime ::varchar-255)
-      (derive ::float ::varchar-255)
-      (derive ::varchar-255 ::text)))
+  (-> (make-hierarchy
+       [::text
+        [::varchar-255
+         [::boolean ::boolean-or-int]
+         [::float
+          [::int ::boolean-or-int ::auto-incrementing-int-pk]]
+         [::datetime ::date]
+         [::offset-datetime]]])))
 
 (def ^:private abstract->concrete
   "Not all value types correspond to database types. For those that don't, this maps to their concrete ancestor."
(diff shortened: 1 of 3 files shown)


crisptrutski:
* **#39793** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

src/metabase/upload.clj
```diff
@@ -79,17 +79,14 @@
...
-      (derive ::float ::varchar-255)
-      (derive ::varchar-255 ::text)))
+  (-> (make-hierarchy
+       [::text
+        [::varchar-255
+         [::boolean ::boolean-or-int]
+         [::float
+          [::int ::boolean-or-int ::auto-incrementing-int-pk]]
+         [::datetime ::date]
+         [::offset-datetime]]])))
```
piranha, line 89: that's readable!

piranha:
I mean, yeah, we're testing that our hierarchy works, but I would love to see tests for behavior of `make-hierarchy`. I was thinking something like this maybe:

```
(let [coll (atom [])]
  (mt/with-dynamic-redefs [derive #(swap! coll conj %&)]
    (make-hierarchy ...)))
```

then we'll see with our eyes what it rolls out into. If behavior ever gets more complex we'll get to see what's going on there with examples.

crisptrutski:
Thinking about things a bit more, we essentially capture the entire structure of the graph in (parents h), I could just assert that it has the right structure, coercing the parent nodes to vectors, and relying on the arbitrary order of keys in a map. I prefer that to the "test spy" approach, I am not a mockist 😁

test/util/ordered_hierarchy_test.clj
```diff
@@ -88,3 +84,38 @@
     (is (= ::boolean-or-int (ordered-hierarchy/first-common-ancestor h ::boolean-or-int ::boolean-or-int)))
     (is (= ::boolean (ordered-hierarchy/first-common-ancestor h ::boolean-or-int ::boolean)))
     (is (= ::varchar-255 (ordered-hierarchy/first-common-ancestor h ::boolean ::int)))))
+
+
```
crisptrutski, line 88: This epic quadruple line break signifies where I'm going break this namespace in half once all the other CSV related PRs have been merged ➗

test/util/ordered_hierarchy_test.clj
```diff
@@ -88,3 +84,38 @@
...
+            :square                [:rhombus :rectangle]
+            :parallelogram         [:quadrilateral]
+            :rectangle             [:parallelogram]
+            :scalene-triangle      [:triangle]
+            :isosceles-triangle    [:triangle]
+            :equilateral-triangle  [:isosceles-triangle :acute-triangle]
+            :acute-triangle        [:triangle]
+            :right-angled-triangle [:triangle]
+            :obtuse-triangle       [:triangle]}
+           (update-vals (:parents polygons) vec)))))
```
crisptrutski, line 121: Actually I realize now that the order of the `parent->child` values matters as well because of how we do the toposort.

The order of these the `child->parent`values actually only matters for the `abstract->concrete` mapping now I think, which actually seems like a bit of a smell.

I'm going to actually park this PR as a draft again until the dust settles on all the related changes.

crisptrutski:
The bulk of this has now been smuggled in through [this PR](https://github.com/metabase/metabase/pull/40275) already. It's a simplified version that only supports the top-down vector based DSL, and not the bottom-up map based one. We have removed any dependence on the order of child-to-parent edges in the application, so it's YAGNI for now.

The introduction of the polygon tests which decouple this "library" from the application will come in another PR.

PR #39794, opened 2024-03-07
Using CSS transition and  component for QB header

### LINKED ISSUE #39776: Question Header

npfitz:
Closes https://github.com/metabase/metabase/issues/39776

### Description
Small change to migrate QB header away from `react-motion`. Essentially before you have picked a data source, we show you one header, and when you've selected one we fade the current header out and a new header in. The transition component will actually take care of unmounting the `NewQuestionHeader` if it's been removed, or not mounting it at all if it was never needed.

### How to verify

1. New question -> Sample Dataset
2. Observe the header as you select a table. You should notice a transition

### Demo
chrome_5UnlMKR4tw

### Checklist
This is purely presentation. there are existing tests that cover whether or not we show "Pick your starting data"

### DIFF
--- a/frontend/src/metabase/query_builder/components/view/View.jsx
+++ b/frontend/src/metabase/query_builder/components/view/View.jsx
@@ -1,6 +1,5 @@
 /* eslint-disable react/prop-types */
 import { Component } from "react";
-import { Motion, spring } from "react-motion";
 import { connect } from "react-redux";
 import { t } from "ttag";
 import _ from "underscore";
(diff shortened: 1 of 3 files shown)


frontend/src/metabase/query_builder/components/view/View.jsx
```diff
@@ -204,22 +210,19 @@ class View extends Component {
...
-        )}
-      </Motion>
+      <QueryBuilderViewHeaderContainer>
+        <BorderedViewTitleHeader
+          {...this.props}
+          style={{
+            transition: "opacity 300ms linear",
+            opacity: isNewQuestion ? 0 : 1,
+          }}
+        />
```
sloansparger, line 220:
```suggestion
     <Transition keepMounted mounted={isNewQuestion} transition={fadeIn} duration={300}>
        <BorderedViewTitleHeader {...this.props} />
    </Transition>
```
Just curious, would something like this work?

sloansparger (approved):

frontend/src/metabase/query_builder/components/view/View.jsx
npfitz: I gave it a quick try. `<BorderedViewTitleHeader />` didn't like being put in the transition layout wise, though that's probably fixable. The transition bit worked just fine though

PR #39796, opened 2024-03-07
Fix janky db refreshing

### LINKED ISSUE #39732: Browse database page refreshes N seconds when a new db is added that makes the page unusable
### Describe the bug

On the Browse data -> Databases tab, after adding a new db, it would show a popup at the bottom right corner that syncs the tables. When that sync is taking place, it will refresh that tab every N seconds. If you have many dbs, trying to scroll down is not possible. 

### To Reproduce

1. Add new db with many tables that takes the db sync awhile to finish. Sync table popup shows at bottom right corner. 
2. Go to Browse data -> Databases and scroll down to the bottom / next page. There must be many dbs to require scrolling down. Even without scrolling, it blinks as refresh happens. 
3. Scrolling down doesn't work as refresh will take you to the top of the page

### Expected behavior

Not blink or refresh should not take user to the top of page

### Logs

n/a

```JSON
metabase @ master
```

### Severity

low since it only happens during sync and if you have many dbs 

### Additional context

_No response_

rafpaf:
Closes #39732 

### Description

The problem is that when a new database is added, the tables are handled in a background process which updates the data available through `/api/database`. The Browse databases page pulls in data from that endpoint. But when the endpoint has fresh data to provide, the page goes into a loading state for a split second, causing the page to instantly scroll back up to the top, which is annoying.

This PR fixes the problem by showing the loading spinner only when there are no databases already loaded into the page.

[Loom](https://www.loom.com/share/2afbc170b42a414e8b78d5fbbd6904a3?sid=6180d129-61ce-490e-a88e-0daf69eb7224)

### DIFF
--- a/frontend/src/metabase/browse/components/BrowseDatabases.tsx
+++ b/frontend/src/metabase/browse/components/BrowseDatabases.tsx
@@ -19,17 +19,17 @@ export const BrowseDatabases = ({
 }: {
   databasesResult: ReturnType<typeof useDatabaseListQuery>;
 }) => {
-  const { data: databases = [], error, isLoading } = databasesResult;
+  const { data: databases, error, isLoading } = databasesResult;
 
   if (error) {
     return <LoadingAndErrorWrapper error />;
   }
 
-  if (isLoading) {
+  if (!databases && isLoading) {
     return <LoadingAndErrorWrapper loading />;
   }
 
-  return databases.length ? (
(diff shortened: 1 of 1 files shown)


iethree (approved):
🥳

iethree:
let's backport this to 49, it's a simple bugfix for a feature in 49

PR #39797, opened 2024-03-07
Remove deprecated and unused `searchTypeaheadEnabled` method

nemanjaglumac:
This PR removes a deprecated and unused `searchTypeaheadEnabled` method from the `MetabaseSettings` prototype.

### How to verify?
- All tests should still pass.
- Search for `search-typeahead-enabled` in all files. You should never see `MetabaseSettings.get("search-typeahead-enabled")` or `MetabaseSettings.searchTypeaheadEnabled()`

This setting is fetched properly (using a hook) in https://github.com/metabase/metabase/blob/master/frontend/src/metabase/nav/components/search/SearchBar/SearchBar.tsx

### DIFF
--- a/frontend/src/metabase/lib/settings.ts
+++ b/frontend/src/metabase/lib/settings.ts
@@ -207,13 +207,6 @@ class MetabaseSettings {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


iethree (approved):
🏆

PR #39804, opened 2024-03-07
Debug QP 2.0

camsaul:
New QP debugger uses Portal and can be enabled with a dynamic var so you can use it anywhere now.

image

```clj
;; start Portal if you have not done so already. Open http://localhost:4000 in your browser
(dev.debug-qp/start-portal!)

;; run a query with debugging enabled
(binding [metabase.query-processor.debug/*debug* true]
  (metabase.query-processor/process-query query)
```

You can add debug output with a new `qp.debug/debug>` tapping macro. Currently I've added preprocessing middleware as well as some of the `annotate` post-processing stuff. We can add other stuff as well going forward.


noahmoss (approved):

dev/src/dev/debug_qp.clj
```diff
@@ -394,3 +402,11 @@
   ([driver sql]
    #_{:clj-kondo/ignore [:discouraged-var]}
    (println (driver/prettify-native-form driver sql))))
+
+(defn start-portal! []
+  (portal.api/start {:port 4000})
+  (add-tap #'portal.api/submit))
```
bshepherdson, line 408: Cypress uses this port for the Metabase BE; we should configure something else.

bshepherdson (approved):

PR #39813, opened 2024-03-07
Maintain user-level state in dashboards for filters

### LINKED ISSUE #39809: [Epic] Preserve Dashboard Filter States
**Links**
- [product doc](https://www.notion.so/metabase/Preserve-dashboard-filter-states-for-users-0f89ff0da0c24026a3692b0f9d704fed)
- feature branch: `preserve-dashboard-filter-states`
- [Testing plan](https://github.com/metabase/metabase/issues/39808)

**Implementation Plan**

***Milestone 1***
```[tasklist]
### Frontend
- [ ] https://github.com/metabase/metabase/pull/39511
- [ ] Replace localStorage usage with a BE implementation - https://github.com/metabase/metabase/pull/39813
- [ ] Verify implementation on embeds, public and regular dashboards with tests
```
```[tasklist]
### Backend
- [ ] https://github.com/metabase/metabase/issues/39810
```

uladzimirdev:
Closes https://github.com/metabase/metabase/issues/39809

### Description

1. If there is no default value and user used some value on the dashboard, we need to restore this value when user gets back to the dashboard, no matter how user opens a dashboard (with a direct URL, after a search)
2. values are saved on BE after every run of the `/query`
3. public dashboards does not provide `last_used_param_values` property on the dashboard (should it be added at BE?), so 
preserving functionality doesn't work there 

### How to verify

- Go to any dashboard with filters
- set value on the filter
- either remove URL search parameters (after `?`) or go to the root collection
- then find your dashboard and with filters and make sure that 
a. values are preserved
b. query for the mapped questions includes preserved parameter value into account

### Demo

1. main functionality

saves last usage of the filter, resets saved value if you reset a filter

https://github.com/metabase/metabase/assets/125459446/75a171f7-7e0e-423d-990a-15a9b8b2d892

2. no auto-apply filter

a. if you use `apply` button, values are saved and reused next time
b. if you do not click "apply", then dashboard is not affected

https://github.com/metabase/metabase/assets/125459446/fa739376-736f-4bc7-9eb0-4511d55caf9b

3. Static embedding is not affected
image

4. Interactive embedding - not tested


uladzimirdev:
Stress test for a flaky parameters test https://github.com/metabase/metabase/actions/runs/9097780561/job/25006416075

frontend/src/metabase-lib/v1/parameters/utils/parameter-values.js
```diff
@@ -33,6 +37,8 @@ export function getValuePopulatedParameters({
   values = {},
   defaultRequired = false,
   collectionPreview = false,
+  // should be a better name for it
+  localDashboardParameters = {},
```
ranquild, line 41: I think `localDashboardParameters` are not needed and are side affect of changing `getParameterValue`. Could you pass `null` for `lastUsedParameterValue` and remove logic to reseting the parameters? If not, what is the issue that we need to solve here?

ranquild (commented):
Inline comment/question. Unclear logic with reseting parameters. Could be okay but I don't understand atm.

I also don't see new tests for this feature.

e2e/test/scenarios/dashboard-filters/dashboard-filters-sql-required-simple-filter.cy.spec.js
```diff
@@ -113,8 +113,8 @@ describe("scenarios > dashboard > filters > SQL > simple filter > required ", ()
 
     saveDashboard();
 
-    // The URL query params should include the parameter with an empty value
-    cy.location("search").should("eq", "?text=");
+    // The URL query params should include the last used parameter value
+    cy.location("search").should("eq", "?text=Bar");
```
uladzimirdev, line 117: this PR changes default behaviour and keeps Bar as a parameter value, parameter value is synced with URL so, URL will contain Bar as well

e2e/test/scenarios/dashboard-filters/reproductions/22788-flter-cc-dropped-on-second-edit.cy.spec.js
```diff
@@ -90,7 +90,8 @@ describe("issue 22788", () => {
 
     saveDashboard();
 
-    addFilterAndAssert();
+    cy.findAllByText("Gizmo");
```
uladzimirdev, line 93: we care only about the resulted values

e2e/test/scenarios/dashboard-filters/parameters.cy.spec.js
```diff
@@ -246,14 +246,15 @@ describe("scenarios > dashboard > parameters", () => {
...
+      "There should only be one filter remaining and its value is preserved",
+    );
 
-    cy.log("There should only be one filter remaining and its value cleared");
     filterWidget().contains(new RegExp(`${endsWith.name}`, "i"));
-    cy.location("search").should("eq", `?${endsWith.slug}=`);
 
-    // eslint-disable-next-line no-unscoped-text-selectors -- deprecated usage
-    cy.findByText("37.65");
+    cy.location("search").should("eq", `?${endsWith.slug}=zmo`);
```
uladzimirdev, line 257: idea of this test was in checking the ability to remove parameter, after removal other filters should keep working. 

as we changed the default behaviour for filters state, URL is not updated and it doesn't make sense to verify what's rendered as a new query is not fired

e2e/test/scenarios/dashboard-filters/reproductions/24500-gracefully-deal-with-corrupted-filter.cy.spec.js
```diff
@@ -119,11 +119,16 @@ describe("issues 15279 and 24500", () => {
...
     // Check the list filter again
-    filterWidget().contains("List").click();
+    filterWidget().contains("List").parent().click();
     cy.wait("@values");
 
-    // Check that the search filter works
+    cy.log("Check that the search filter works");
+
+    // reset filter value
+    filterWidget().contains("Search").parent().icon("close").click();
```
uladzimirdev, line 128: when filterWidget contains some value, title `Search` still presents in DOM, but it's not a root element anymore

frontend/src/metabase-lib/v1/parameters/utils/parameter-values.js
uladzimirdev: there are several edge cases with the logic that's why I added this override for sure, but let me double check maybe it's not needed anymore

uladzimirdev: @ranquild CI is green so you were right, nice catch!

ranquild (approved):

PR #39205, opened 2024-02-27
Add UI to swap clauses in the notebook

ranquild:
Epic https://github.com/metabase/metabase/issues/39210
Demo https://www.loom.com/share/46322c56db044c4d8d8de5c890510e62

Allows to drag-n-drop custom columns, filters, aggregations, and breakouts. 

How to verify:
- Add a question with all of these clauses
- Try to reorder them via drag-n-drop

PR #39850, opened 2024-03-08
[MLv2] Introduce `swap-clauses` to reorder aggregations, etc.

bshepherdson:
This includes some generative testing, since that felt like
a great way to try all kinds of permutations of swapping clauses.

This also patches `metabase.types/assignable?` to memoize it.
Deeply nested `assignable?` calls made 100 tests of swapping
filters take 1600ms rather than the 50-60ms of the other clauses.

I think this is safe and reasonable on memory to cache forever,
but speak up if you're concerned. There's perhaps O(60) type
keywords, so `type X type -> Boolean` is a small space impact.

### DIFF
--- a/test/metabase/lib/swap_test.cljc
+++ b/test/metabase/lib/swap_test.cljc
@@ -1,0 +1,114 @@
+(ns metabase.lib.swap-test
+  (:require
+   [clojure.test :refer [deftest is testing]]
+   [clojure.test.check.clojure-test :refer [defspec]]
+   [clojure.test.check.generators :as gen]
+   [clojure.test.check.properties :as prop]
+   [medley.core :as m]
+   [metabase.lib.core :as lib]
+   [metabase.lib.options :as lib.options]
+   [metabase.lib.test-metadata :as meta]
+   [metabase.test.util.log :as tu.log]
+   #?@(:cljs ([metabase.test-runner.assert-exprs.approximately-equal]))))
+
+#?(:cljs (comment metabase.test-runner.assert-exprs.approximately-equal/keep-me))
+
+(defn- swapped-clauses-prop [clause-fn clause-key query]
+  (prop/for-all*
+    [(gen/elements (clause-fn query))
+     (gen/elements (clause-fn query))]
+    (fn [source-clause target-clause]
+      (or (= source-clause target-clause) ; Skip any case where we happened to draw the same one twice.
+          (let [swapped (lib/swap-clauses query -1 source-clause target-clause)]
+            (and ;; Correctly rearranged the clauses.
+                 (= (for [clause (clause-fn query)]
+                      (cond
+                        (= source-clause clause) target-clause
+                        (= target-clause clause) source-clause
+                        :else                    clause))
+                    (clause-fn swapped))
+                 ;; And didn't change anything else.
+                 (= (m/dissoc-in query   [:stages 0 clause-key])
+                    (m/dissoc-in swapped [:stages 0 clause-key]))))))))
+
+(defspec swap-clauses-on-aggregations-test-permutations
+  (swapped-clauses-prop
+    lib/aggregations :aggregation
+    (-> (lib/query meta/metadata-provider (meta/table-metadata :orders))
+        (lib/aggregate (lib/count))
+        (lib/aggregate (lib/sum (meta/field-metadata :orders :subtotal)))
+        (lib/aggregate (lib/avg (meta/field-metadata :orders :discount)))
+        (lib/aggregate (lib/max (meta/field-metadata :orders :tax)))
+        (lib/aggregate (lib/min (meta/field-metadata :orders :total))))))
+
+(defspec swap-clauses-on-breakouts-test-permutations 300 ; Extra tests to hit the similar cases all the time!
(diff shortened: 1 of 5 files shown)


bshepherdson:
* **#39850** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39850?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @bshepherdson and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

ranquild (approved):

test/metabase/lib/swap_test.cljc
```diff
@@ -0,0 +1,123 @@
...
+        (lib/aggregate (lib/max (meta/field-metadata :orders :tax)))
+        (lib/aggregate (lib/min (meta/field-metadata :orders :total))))))
+
+(defspec swap-clauses-on-breakouts-test-permutations
+  (swapped-clauses-prop
+    lib/breakouts :breakout
+    (as-> (lib/query meta/metadata-provider (meta/table-metadata :orders)) $q
+      (lib/breakout $q (meta/field-metadata :products :category))
+      (lib/breakout $q (meta/field-metadata :people :source))
+      (lib/breakout $q (lib/with-binning
```
snoe, line 50: Would be nice to test the same field with different/no levels of binning

test/metabase/lib/swap_test.cljc
```diff
@@ -0,0 +1,123 @@
...
+                                           (meta/field-metadata :people :id))]))
+        (lib/join (lib/join-clause (meta/table-metadata :products)
+                                   [(lib/= (meta/field-metadata :orders :product-id)
+                                           (meta/field-metadata :products :id))]))
+        (lib/join (lib/join-clause (meta/table-metadata :checkins)
+                                   [(lib/= (meta/field-metadata :checkins :user-id)
+                                           (meta/field-metadata :orders :user-id))]))
+        (lib/join (lib/join-clause (meta/table-metadata :venues)
+                                   [(lib/= (meta/field-metadata :people :latitude)
+                                           (meta/field-metadata :venues :latitude))])))))
```
snoe, line 89: Would swapping this with the join to `:people` break the query? I don't see anything that would prevent that.

test/metabase/lib/swap_test.cljc
```diff
@@ -0,0 +1,123 @@
...
+
+(defspec swap-clauses-on-expressions-test-permutations
+  (swapped-clauses-prop
+    lib/expressions :expressions
+    (-> (lib/query meta/metadata-provider (meta/table-metadata :orders))
+        (lib/expression "clean-discount" (lib/coalesce (meta/field-metadata :orders :discount) 0))
+        (lib/expression "discount fraction" (lib// (meta/field-metadata :orders :discount)
+                                                   (meta/field-metadata :orders :subtotal)))
+        (lib/expression "order month" (lib/get-month (meta/field-metadata :orders :created-at)))
+        (lib/expression "signup year" (lib/get-year  (meta/field-metadata :people :created-at))))))
```
snoe, line 72: What about expressions that depend on a previous expression?

snoe (approved):
I believe order can be dependent in expressions & joins. I'm not sure if the FE is requesting these swaps or not, if not then I would leave them out of swappable causes for the moment, if so then I would prevent illegal swaps.

test/metabase/lib/swap_test.cljc
bshepherdson: Expressions are not ordered (semantically) AFAIK. They will end up siblings in a `SELECT ... FROM` eventually, and those can be in any order.

Actually today I wrote a complicated expression, then split it up into two for ease of understanding. I edited the first/left one to reference the second and it was all good.

snoe: Thanks, I was surprised this work, but it does because the qp runs substitution on expression references so `[two] + 2`  turns into native `(1 + 1) + 2`

```
(let [mp (lib.metadata.jvm/application-database-metadata-provider (mt/id))
      query (-> (lib/query mp (lib.metadata/table mp (mt/id :products)))
                       (lib/expression "two" (lib/+ 1 1))
                       (as-> $q (lib/expression $q "four" (lib/+ (lib/expression-ref $q "two") 2))))
      query (update-in query [:stages 0 :expressions] reverse)]
 (qp/process-query query))
 ```

ranquild:
> I believe order can be dependent in expressions & joins. I'm not sure if the FE is requesting these swaps or not, if not then I would leave them out of swappable causes for the moment, if so then I would prevent illegal swaps.

Current the FE would not call `swapClauses` with joins.

ranquild:
FE changes approved here https://github.com/metabase/metabase/pull/39205

test/metabase/lib/swap_test.cljc
bshepherdson: Joins have been dropped.

test/metabase/lib/swap_test.cljc
bshepherdson: Done.

Issue #34523, opened 2023-10-10
Redundant GET /api/model-index fired when creating a model


kamilmielnik:
### Describe the bug

GET `/api/model-index` is redundantly called without any parameters and fails with HTTP 400 and following JSON:

```JSON
{
    "errors": {
        "model_id": "value must be an integer greater than zero."
    },
    "specific-errors": {
        "model_id": [
            "value must be an integer greater than zero., received: nil"
        ]
    }
}
```

### To Reproduce

1. Click "+ New"
2. Choose "Model"
3. Choose "Use the notebook editor" (or "Use a native query")

GET `/api/model-index` is called.

### Expected behavior

GET `/api/model-index` is not called.

### Information about your Metabase installation

`master`, v47

### Severity

P3

### Additional context

Transparent to the user, but it affects our tests setup because of #34277

Code that causes it: https://github.com/metabase/metabase/blob/bf764dd9cc11de44e5f3efacc3cc310e3ede8fa4/frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx#L532-L535

----

:warning: please remove [this workaround](https://github.com/metabase/metabase/blob/6cec76055335c80092f3697c3011c8c8461916c3/frontend/src/metabase/query_builder/containers/QueryBuilder.unit.spec.tsx#L249-L252) when fixing the issue :warning:

PR #39851, opened 2024-03-08
Fix - Description for calculated columns in model metadata editing page is incorrectly populated with description from other columns

### LINKED ISSUE #25884: Adding a description to a model column duplicates the description to other columns if they are null
**Describe the bug**
If a model contains columns without a description, adding a description to one column propagates the same description to the following.

**To Reproduce**
1. Go to https://stats.metabase.com/model/5371-view-logs/metadata
2. Edit metadata of the first field
3. It propagates to other fields (deleting the description also propagates the empty field).

**Expected behavior**
It should alter only the defined field description.

**Screenshots**
[Loom](https://www.loom.com/share/b1d6c0cf97e54118ad5e0fbcb5f4e4bb)

**Severity**
P2

### LINKED ISSUE #34349: Description for calculated columns in model metadata editing page is incorrectly populated with description from other columns
### Describe the bug

Description for calculated columns in model metadata editing page is incorrectly populated with description from other columns

### To Reproduce

1. New Question > Raw Data > Sample Dataset > Orders
2. Add calculated column "County" with the expression `substring("United States", 1, 20)`
3. Run query
4. Save Question
5. Convert to model
6. Edit model metadata
7. Click `Country` column. Description is from whatever column you clicked last instead of being empty

### Expected behavior

Description should be empty for the calculated column

### Logs

_No response_

```JSON
This is on stats on pre-48 (October 3rd, 2023)
```

### Severity

annoying. Correctness issue

### Additional context

https://www.loom.com/share/a2a4943784f347408e4f5b5f270719a5?sid=41e527ff-5883-4af4-bcf9-6d143fc5da32

kamilmielnik:
Fixes #34349
Fixes #25884

Stress test: https://github.com/metabase/metabase/actions/runs/8204344171

### DIFF
--- a/e2e/test/scenarios/models/reproductions/25884-34349-null-description-in-column-metadata.cy.spec.ts
+++ b/e2e/test/scenarios/models/reproductions/25884-34349-null-description-in-column-metadata.cy.spec.ts
@@ -3,13 +3,16 @@ import { openQuestionActions, popover, restore } from "e2e/support/helpers";
 
 const { ORDERS, ORDERS_ID } = SAMPLE_DATABASE;
 
-describe("issue 34349", () => {
+const ID_DESCRIPTION =
+  "This is a unique ID for the product. It is also called the “Invoice number” or “Confirmation number” in customer facing emails and screens.";
+
+describe("issues 25884 and 34349", () => {
   beforeEach(() => {
     restore();
     cy.signInAsAdmin();
   });
 
-  it("should show empty description input for columns without description in metadata (metabase#34349)", () => {
+  it("should show empty description input for columns without description in metadata (metabase#25884, metabase#34349)", () => {
     cy.createQuestion(
       {
         type: "model",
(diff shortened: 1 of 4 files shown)


frontend/src/metabase/query_builder/containers/test-utils.tsx
```diff
@@ -246,7 +246,6 @@ export const setup = async ({
     setupModelIndexEndpoints(card.id, []);
   }
 
-  // this workaround can be removed when metabase#34523 is fixed
```
kamilmielnik, line 249: _This is totally unrelated, just found during triage and I didn't want to have a separate PR for it._

#34523 was fixed, but this comment was inaccurate - we still need to mock `/model-index` endpoint which is used after a model is created.

e2e/test/scenarios/models/reproductions/25884-34349-null-description-in-column-metadata.cy.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      {
+        type: "model",
+        query: {
+          "source-table": ORDERS_ID,
+          expressions: {
+            Country: ["substring", "United States", 1, 20],
+          },
+          fields: [
+            ["field", ORDERS.ID, { "base-type": "type/BigInteger" }],
+            ["expression", "Country", { "base-type": "type/Text" }],
```
kamilmielnik, line 26: Using only 2 columns for this model so that we don't have to scroll the table horizontally to click the custom column header (which is the last column). Scrolling the table with `scrollTo("right")` didn't work very well, we'd have to call it twice or after a delay.

ranquild (approved):

PR #39853, opened 2024-03-08
Remove info icon on table

romeovs:
### Description

Incorporate @mazameli's feedback for the info icon on the table header

### How to verify

1. Browse data -> Orders
2. Make sure there is an info icon in the header
3. Save the question
4. Open the saved question and hover the header
5. Make sure the `Orders` link in the subheader does not have the info icon


ranquild (approved):
Should `hideIfEmpty` be the default behavior without an additional prop?

PR #39858, opened 2024-03-08
Fixes for LLM auto-description

kulyk:
1. Fixes the FE was using the wrong card summarize endpoint
2. Fixes "AI Features" section in admin settings was visible for tokens without the LLM feature
3. Fixes the AI button was visible only starting from `settings-manager` user role by introducing a "AI features" on/off public setting


frontend/src/metabase/admin/settings/selectors.js
```diff
@@ -691,9 +691,14 @@ export const ADMIN_SETTINGS_SECTIONS = {
   },
   llm: {
     name: t`AI Features`,
-    getHidden: () => !PLUGIN_IS_EE_BUILD.isEEBuild(),
+    getHidden: () => !PLUGIN_LLM_AUTODESCRIPTION.isEnabled(),
```
adam-james-v, line 694: I think this isn't the right check. We should still expose the AI Feature section in settings based on 

`PLUGIN_IS_EE_BUILD.isEEBuild()`

kulyk: `PLUGIN_LLM_AUTODESCRIPTION.isEnabled()` will be true only if there's an enterprise token with the autodescription feature (code [here](https://github.com/metabase/metabase/blob/cbdf9ae8be3db2df1093f67a9a86a35a3e41cf9b/enterprise/frontend/src/metabase-enterprise/llm_autodescription/index.ts#L7)). I think the `EE_BUILD` check might be redundant in this case. Or am I missing something?

adam-james-v: Well, when I started up this branch locally, I was admin, and had EE enabled, the API key was filled in, but `ee-ai-features-enabled` was `false`. 

The AI features section did not show up in this situation.

adam-james-v: Ok, yea this was a problem locally, please disregard the comment, you are correct :)

adam-james-v (approved):

Issue #39861, opened 2024-03-08
BE flake: postgres tests fail with `mt/with-temp-copy-of-db`


calherries:
Postgres tests are often flaking on `copy-db-fks!`. 

The tests that fail are not consistent between runs, but the common factor is that they use `mt/with-temp-copy-of-db`

example:
https://github.com/metabase/metabase/actions/runs/8192774915/job/22406695644#step:4:2096

My theory is the fields aren't found because they're inactive.

crisptrutski:
Definitely noticed the uptick in these flakes, thanks for jumping on it.

calherries:
Closed by https://github.com/metabase/metabase/pull/39926

calherries:
closing this because it's a dupe of https://github.com/metabase/metabase/issues/38052

PR #39862, opened 2024-03-08
Only copy active fields and tables with data/with-temp-copy-of-db

calherries:
Hopefully fixes this flake https://github.com/metabase/metabase/issues/39861

We're checking for fields and tables being `active=true` inconsistently in the implementation of `data/with-temp-copy-of-db`. This PR changes that so we're always checking for `active=true`, so we'll never attempt to copy inactive fields and tables.

### My theory for why this fixes the flake:
I think some fields are being made inactive in the app db, and when we try to copy over their fk relationships we can't find the field the fk relationships correspond to because the field is inactive.

### DIFF
--- a/test/metabase/test/data/impl.clj
+++ b/test/metabase/test/data/impl.clj
@@ -133,7 +133,8 @@
   [database-id :- ::lib.schema.id/database]
   (t2/select-fn->pk (juxt (constantly database-id) :name)
                     [:model/Table :id :name]
-                    :db_id database-id))
+                    :db_id  database-id
+                    :active true))
 
 (mu/defn ^:private build-field-lookup-map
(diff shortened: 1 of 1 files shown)


test/metabase/test/data/impl.clj
```diff
@@ -246,7 +246,7 @@
 
 (defn- copy-table-fields! [old-table-id new-table-id]
   (t2/insert! Field
-    (for [field (t2/select Field :table_id old-table-id {:order-by [[:id :asc]]} :active true)]
+    (for [field (t2/select Field :table_id old-table-id, :active true, {:order-by [[:id :asc]]})]
```
crisptrutski, line 249: Nice query readability improvements

crisptrutski (approved):

Issue #39558, opened 2024-03-04
[dc.js migration] missing `(empty)` label on x-axis


EmmadUsmani:
question/13431
https://metaboat.slack.com/archives/C06G94JTWBS/p1706193308481349

<img width="1485" alt="Screenshot 2024-03-04 at 12 11 40 PM" src="https://github.com/metabase/metabase/assets/37751258/f6264cc4-28b2-4e8a-86b4-a75a6a405376">

question/13429
https://metaboat.slack.com/archives/C06G94JTWBS/p1706196244195369

PR #39913, opened 2024-03-08
Fix null ordinal value formatting

alxnddr:
Closes https://github.com/metabase/metabase/issues/39558

### Description

Fixes null formatting on ordinal scales.

### How to verify

Create a question
```
select 1 x, 10 y
union all select null, 10
union all select 3, 10
```
Switch x-axis to Ordinal and ensure the null is formatted as `(empty)`

### Demo

Screenshot 2024-03-08 at 7 20 27 PM

### Checklist


frontend/src/metabase/visualizations/echarts/cartesian/model/dataset.ts
```diff
@@ -310,6 +311,14 @@ export const applyVisualizationSettingsDataTransformations = (
       condition: settings["graph.y_axis.scale"] === "pow",
       fn: getKeyBasedDatasetTransform(seriesDataKeys, applySquareRootScaling),
     },
+    {
+      condition: isCategoryAxis(xAxisModel),
+      fn: getKeyBasedDatasetTransform([X_AXIS_DATA_KEY], value => {
+        return isCategoryAxis(xAxisModel) && value === null
```
EmmadUsmani, line 317: I don't think it's possible for a row value to be `undefined`, but just in case maybe we should use `==` instead?

EmmadUsmani (approved):

PR #39930, opened 2024-03-11
Correctly show a selected table (in the data selector) which comes from a multi-schema database

### LINKED ISSUE #39807: Data source is not properly highlighted for simple saved questions based on a table in a multi-schema database
### Describe the bug

Data source is not properly highlighted for simple saved questions based on a table in a multi-schema database. This is a part of this general problem: https://github.com/metabase/metabase/issues/39542

Somewhat similar to https://github.com/metabase/metabase/issues/39699, but the source of the bug is different as it's highly specific to multi-schema databases.

### To Reproduce

_You'll need a multi-schema database to test this_
1. Create a GUI question based on a table from a multi-schema database and save it
2. Open that question
3. Switch to the notebook editor
4. Click on the "data pill" that should display the table name (in my case, it was the "Table Two" from the `schema_two`) to open the data picker
5. Notice that the data picker doesn't show the underlying table and it doesn't even highlight the database this table belongs to
image

### Expected behavior

The data picker should highlight the table this question is based on image

### Logs

_No response_

```JSON
local dev, H2, Sample Database, `master`, 180cf78
```

### Severity

P2/P3

### Additional context

Some additional product/design feedback and support might be needed.

nemanjaglumac:
Similar to #39699, we don't have the necessary information when opening the table from a multi-schema database. And similar to #39699, the information is there when we go through "New > Question/Model" flow. The `DataSelector` component seems to have been built with that one way flow in mind.

This PR is adding this missing piece of information about the schema to the state when we're opening a saved question.

Reproduces and fixes #39807.

## Testing
This PR also introduces a real [multi-schema database](https://github.com/metabase/metabase/pull/39930/commits/7063c0569126ebec31ac040bb6ac1270608857e8) in E2E tests. I had to work around some of the existing implicit contracts in the related helpers. Ideally, we will adjust those helpers to support the multi-schema context in a more natural way.

The biggest downside currently is that the information about the schema names and the table names stays unavailable to the outside world. This can easily be solved with a callback that takes the return value of a `resetTable` Cypress task as an argument. cc @iethree 

Solving that is outside the scope of this PR.

### DIFF
--- a/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
+++ b/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
@@ -623,6 +623,8 @@ export class UnconnectedDataSelector extends Component {
       [TABLE_STEP]: () => {
         if (this.state.selectedSchemaId != null) {
           return this.props.fetchSchemaTables(this.state.selectedSchemaId);
+        } else if (this.state.selectedSchema?.id != null) {
+          return this.props.fetchSchemaTables(this.state.selectedSchema?.id);
         }
       },
       [FIELD_STEP]: () => {
--- a/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
+++ b/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
@@ -864,9 +864,6 @@ export class UnconnectedDataSelector extends Component {
 
   isSavedEntitySelected = () => isVirtualCardId(this.props.selectedTableId);
 
-  isTableSelected = () =>
-    this.props.selectedTableId && !this.isSavedEntitySelected();
-
   handleSavedEntitySelect = async tableOrCardId => {
     await this.props.fetchFields(tableOrCardId);
     if (this.props.setSourceTableFn) {
--- a/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
+++ b/frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
@@ -259,6 +259,9 @@ export class UnconnectedDataSelector extends Component {
     }
     if (selectedTableId != null) {
       setSelectedTable(getTable(selectedTableId));
+      // We need the schema information when we open already saved question (vitual table id),
+      // or already selected table (the actual table id).
+      setSelectedSchema(selectedTable.schema);
     }
     if (selectedFieldId != null) {
       setSelectedField(getField(selectedFieldId));
@@ -458,9 +461,9 @@ export class UnconnectedDataSelector extends Component {
     } else if (
       // If we go from New > Question/Model flow, schemaId will be selected.
       // OTOH, if we're opening already saved question, or already chosen table,
-      // the schema id information is not preserved.
-      // In that case, we can rely on the selected table information.
-      (this.state.selectedSchemaId || this.isTableSelected()) &&
+      // the schema id information is not preserved and we have to access it directly
+      // through the schema object.
+      (this.state.selectedSchemaId || this.state.selectedSchema?.id) &&
       steps.includes(TABLE_STEP)
     ) {
       await this.switchToStep(TABLE_STEP);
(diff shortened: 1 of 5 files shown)


frontend/src/metabase/query_builder/components/DataSelector/DataSelector.jsx
```diff
@@ -455,7 +455,12 @@ export class UnconnectedDataSelector extends Component {
       await this.switchToStep(DATABASE_STEP);
     } else if (this.state.selectedTableId && steps.includes(FIELD_STEP)) {
       await this.switchToStep(FIELD_STEP);
-    } else if (this.state.selectedSchemaId && steps.includes(TABLE_STEP)) {
+    } else if (
+      // Schema id is explicitly set when going through the New > Question/Model flow,
+      // whereas we have to obtain it from the state when opening a saved question.
+      (this.state.selectedSchemaId || this.state.selectedSchema?.id) &&
```
nemanjaglumac, line 461: This line here is the actual fix.
Everything else is related to tests.

nemanjaglumac: @ranquild no matter what I do, there's always some part of the internal entities state that's not happy. This particular solution works perfectly for multi-schema databases. There is an E2E test that confirms that.

But there are some other E2E tests that fail because we're not fetching`/api/database/${SAMPLE_DB_ID}/schema/PUBLIC` anymore. This can also result in only the selected table being shown
image

If you have any other ideas how to approach this, please LMK.
Also, feel free to commit directly to the branch.

nemanjaglumac: As per our previous conversation, one of the attempts at fixing this problem was setting only the `selectedSchemaId` but in practice it was still often `undefined` so it didn't solve the multi-schema problem.
https://github.com/metabase/metabase/pull/39930/commits/d55b91596c0e94e6867d8a3900eeca0af5f9a91c#diff-b8d140ac6041c995aeefb03f0ac704c4ec7410ef7772858b93f4dec5a57fc7c6R265-R267

nemanjaglumac: > If you have any other ideas how to approach this, please LMK.
> Also, feel free to commit directly to the branch.

Btw, this is my attempt at solving the problem. LMK if the approach makes sense:
https://github.com/metabase/metabase/pull/39930/commits/83ce3c249a76502dbd4dc89a38f75c22a7030bcb

frontend/src/metabase/query_builder/components/DataSelector/DataSelectorTablePicker/DataSelectorTablePicker.tsx
```diff
@@ -180,11 +180,16 @@ const Header = ({
...
+        {selectedDatabase.name}
+      </HeaderDatabaseName>
     </HeaderClickable>
 
     {selectedSchema?.name && schemas.length > 1 && (
-      <HeaderSchemaName>- {selectedSchema.displayName()}</HeaderSchemaName>
+      <>
+        <HeaderSchemaDivider>-</HeaderSchemaDivider>
+        <span data-testid="source-schema">{selectedSchema.displayName()}</span>
+      </>
```
nemanjaglumac, line 192: I could've let this one go, but it really bothered me that we're faking the distance on the left side of the divider `-` with the margin, while using a regular space on the right.

Also saw a test that asserted the following string was in the DOM "- schemaName" which felt wrong.
This is not the ideal solution either and this whole component could be improved, but it does the job for now, IMO.

frontend/src/metabase/query_builder/components/DataSelector/DataSelectorTablePicker/DataSelectorTablePicker.tsx
```diff
@@ -180,11 +180,16 @@ const Header = ({
...
+        {selectedDatabase.name}
+      </HeaderDatabaseName>
     </HeaderClickable>
 
     {selectedSchema?.name && schemas.length > 1 && (
-      <HeaderSchemaName>- {selectedSchema.displayName()}</HeaderSchemaName>
+      <HeaderSchemaName>
+        {"- "}
+        <span data-testid="source-schema">{selectedSchema.displayName()}</span>
+      </HeaderSchemaName>
```
nemanjaglumac, line 192: I could've let this one go, but it really bothered me that we're asserting that the following string exists in the DOM "- schemaName". This just felt wrong.

This is not the ideal solution either and this whole component could be improved, but it does the job for now, IMO.
I'll follow up with another PR that fixes https://github.com/metabase/metabase/issues/39999

ranquild (approved):

PR #39941, opened 2024-03-11
Add token parameter to JWT SSO

oisincoveney:
Adds a `token` parameter to the JWT SSO endpoint.

### Why?

Two reasons, based on our work on the Embedding SDK.

1. Using tokens, we can monitor the expiration and renew the token ahead of its expiration. Right now, we're using cookies in interactive embedding, which can cause an embedded Metabase instance to redirect to a JWT identity provider which will break user flows and possibly lose a user's progress if they're editing or creating something new.

2. [Browsers are moving away from third-party cookies to preserve customer privacy](https://developers.google.com/privacy-sandbox/3pcd). This PR will help us refactor iframes and such in the future to make sure we're using token authentication rather than cookies, and hopefully help us ensure that things won't immediately break when this policy is put into place.

### How to verify

The tests should pass. We'll have some examples from the embedding side in the future. In the meantime, no behavior should change at all.

### Checklist

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
+++ b/enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
@@ -76,7 +76,7 @@
                                                      (group-names->ids group-names)
                                                      (all-mapped-group-ids))))))
 
-(defn- login-jwt-user
+(defn- get-jwt-data
   [jwt {{redirect :return_to} :params, :as request}]
   (let [redirect-url (or redirect (URLEncoder/encode "/"))]
     (sso-utils/check-sso-redirect redirect-url)
@@ -103,25 +103,29 @@
 (defn ^:private generate-response-token
   [session jwt-data]
   (validation/check-embedding-enabled)
-  (response/response {:id (:id session)
+  (response/response {:id  (:id session)
                       :exp (:exp jwt-data)
                       :iat (:iat jwt-data)}))
 
+(defn ^:private redirect-to-idp
+  [idp redirect]
+  (let [return-to-param (if (str/includes? idp "?") "&return_to=" "?return_to=")]
+    (response/redirect (str idp (when redirect
+                                  (str return-to-param redirect))))))
+
+(defn ^:private handle-jwt-authentication
+  [{:keys [session redirect-url jwt-data]} token request]
+  (if token
+    (generate-response-token session jwt-data)
+    (mw.session/set-session-cookies request (response/redirect redirect-url) session (t/zoned-date-time (t/zone-id "GMT")))))
+
 (defmethod sso.i/sso-get :jwt
-  [{{:keys [jwt redirect token]
-     :or   {token false}} :params
-    :as                                             request}]
+  [{{:keys [jwt redirect token] :or {token false}} :params, :as request}]
   (premium-features/assert-has-feature :sso-jwt (tru "JWT-based authentication"))
   (check-jwt-enabled)
   (if jwt
-    (let [{:keys [session redirect-url jwt-data]} (login-jwt-user jwt request)]
-      (if token
-        (generate-response-token session jwt-data)
-        (mw.session/set-session-cookies request (response/redirect redirect-url) session (t/zoned-date-time (t/zone-id "GMT")))))
-    (let [idp             (sso-settings/jwt-identity-provider-uri)
-          return-to-param (if (str/includes? idp "?") "&return_to=" "?return_to=")]
-      (response/redirect (str idp (when redirect
-                                    (str return-to-param redirect)))))))
+    (handle-jwt-authentication  (get-jwt-data jwt request) token request)
+    (redirect-to-idp (sso-settings/jwt-identity-provider-uri) redirect)))
 
 (defmethod sso.i/sso-post :jwt
   [_]
--- a/enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
+++ b/enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
@@ -8,6 +8,7 @@
    [metabase-enterprise.sso.integrations.sso-settings :as sso-settings]
    [metabase-enterprise.sso.integrations.sso-utils :as sso-utils]
    [metabase.api.common :as api]
+   [metabase.api.common.validation :as validation]
    [metabase.api.session :as api.session]
    [metabase.integrations.common :as integrations.common]
    [metabase.public-settings.premium-features :as premium-features]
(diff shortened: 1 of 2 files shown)


oisincoveney:
* **#39941** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39941?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @oisincoveney and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
```diff
@@ -334,3 +334,24 @@
...
+    (with-jwt-default-setup
+      (let [jwt-iat-time (buddy-util/now)
+            jwt-exp-time (+ (buddy-util/now) 3600)
+            result       (saml-test/client-full-response :get 200 "/auth/sso"
+                                                         :token true
+                                                         :jwt (jwt/sign {:email
+                                                                         "rasta@metabase.com"
+                                                                         :first_name "Rasta"
+                                                                         :last_name  "Toucan"
+                                                                         :extra      "keypairs"
```
tsmacdonald, line 349: did you want to test that extra keypairs come back too?

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
```diff
@@ -93,22 +94,34 @@
...
+    :as                                             request}]
   (premium-features/assert-has-feature :sso-jwt (tru "JWT-based authentication"))
   (check-jwt-enabled)
   (if jwt
-    (login-jwt-user jwt request)
-    (let [idp (sso-settings/jwt-identity-provider-uri)
+    (let [{:keys [session redirect-url jwt-data]} (login-jwt-user jwt request)]
+      (if token
+        (generate-response-token session jwt-data)
+        (mw.session/set-session-cookies request (response/redirect redirect-url) session (t/zoned-date-time (t/zone-id "GMT")))))
```
tsmacdonald, line 120: This is a lot of `if`s and `let`s; I'd extract the two branches of `if token` into helper fns.

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
tsmacdonald: Now that I think about it I guess it doesn't even matter if extra keypairs roundtrip; it's just about making sure they don't mess up the stuff we do care about

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
```diff
@@ -93,22 +94,34 @@
...
+      {:session session, :redirect-url redirect-url, :jwt-data jwt-data})))
 
 (defn- check-jwt-enabled []
   (api/check (sso-settings/jwt-enabled)
     [400 (tru "JWT SSO has not been enabled and/or configured")]))
 
+(defn ^:private generate-response-token
+  [session jwt-data]
+  (validation/check-embedding-enabled)
+  (response/response {:id (:id session)
```
tsmacdonald, line 106:
```suggestion
  (response/response {:id  (:id session)
```

tsmacdonald (commented):
This seems to make sense to me, but I'm going to do the lame thing and say that since it's so critical to get right and I'm not super familiar with this corner of the code I don't feel confident enough to :heavy_check_mark: it

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
oisincoveney: Exactly - we could definitely fire those keypairs into the sun but I just want to make sure there isn't any extra weirdness when they're introduced

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
```diff
@@ -75,7 +76,7 @@
                                                      (group-names->ids group-names)
                                                      (all-mapped-group-ids))))))
 
-(defn- login-jwt-user
+(defn- get-jwt-data
```
tsmacdonald, line 79: usually Clojure avoids the `get-` prefix. I was going to say it should be called `jwt-data`, but it's more than that (it returns a map with a key called `:jwt-data`, so…)

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
```diff
@@ -334,3 +335,25 @@
           clojure.lang.ExceptionInfo
           #"Sorry, but you'll need a test account to view this page. Please contact your administrator."
           (#'mt.jwt/fetch-or-create-user! "Test" "User" "test1234@metabase.com" nil)))))))
+
+(deftest get-jwt-token-test
```
tsmacdonald, line 339:
```suggestion
(deftest jwt-token-test
```

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
```diff
@@ -334,3 +335,25 @@
...
+  (testing "should return a session token when token=true"
```
tsmacdonald, line 340: worth testing the `token=false` case just in case?

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
```diff
@@ -93,22 +94,38 @@
...
+  [{{:keys [jwt redirect token] :or {token false}} :params, :as request}]
   (premium-features/assert-has-feature :sso-jwt (tru "JWT-based authentication"))
   (check-jwt-enabled)
   (if jwt
-    (login-jwt-user jwt request)
-    (let [idp (sso-settings/jwt-identity-provider-uri)
-          return-to-param (if (str/includes? idp "?") "&return_to=" "?return_to=")]
-      (response/redirect (str idp (when redirect
-                                   (str return-to-param redirect)))))))
+    (handle-jwt-authentication  (get-jwt-data jwt request) token request)
```
noahmoss, line 127: nit
```suggestion
    (handle-jwt-authentication (get-jwt-data jwt request) token request)
```

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
```diff
@@ -93,22 +94,38 @@
...
+
+(defn ^:private handle-jwt-authentication
+  [{:keys [session redirect-url jwt-data]} token request]
+  (if token
+    (generate-response-token session jwt-data)
+    (mw.session/set-session-cookies request (response/redirect redirect-url) session (t/zoned-date-time (t/zone-id "GMT")))))
+
 (defmethod sso.i/sso-get :jwt
-  [{{:keys [jwt redirect]} :params, :as request}]
+  [{{:keys [jwt redirect token] :or {token false}} :params, :as request}]
```
noahmoss, line 123: I'd probably let this be `nil` when not provided. It'll behave the same as `false` in conditionals, just a bit more idiomatic.
```suggestion
  [{{:keys [jwt redirect token]} :params, :as request}]
```

noahmoss (approved):
agree with Tim's comments, added a couple nits, otherwise lgtm

enterprise/backend/test/metabase_enterprise/sso/integrations/jwt_test.clj
oisincoveney: that's a good idea too

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
oisincoveney: I changed it to `session-data` which I think is a more accurate way to describe that function (at least more than just `jwt-data`)

enterprise/backend/src/metabase_enterprise/sso/integrations/jwt.clj
oisincoveney: In general, I should default to `nil` instead of `false` for these types of things?

tsmacdonald: `nil` means you don't have it, `false` means you have it but it's, uh, false. Not being provided in a map is a classic nil case.

oisincoveney: Ok, perfect. Thanks for the tip

PR #39944, opened 2024-03-11
Publish snowplow events on CSV append success / failure

### LINKED ISSUE #39867: Publish snowplow events when appending to uploads
See the corresponding event that we are already sending when uploads are created.

crisptrutski:
Closes https://github.com/metabase/metabase/issues/39867

### Description

This takes "heavy inspiration" from the initial upload events to give analytics coverage for appends as well - the only salient difference is that we don't refer back to the original model created when the table was first created.

### How to verify

... I'd actually appreciate a pointer here, I'm not sure how to test all the wiring. 

Could even have missed something, in fact.

### DIFF
--- a/snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-0
+++ b/snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-0
@@ -12,7 +12,7 @@
     "event": {
       "description": "Event name",
       "type": "string",
-      "enum": ["csv_upload_successful", "csv_upload_failed"],
+      "enum": ["csv_upload_successful", "csv_upload_failed", "csv_append_successful", "csv_append_failed"],
       "maxLength": 128
     },
     "model_id": {
--- a/snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-1
+++ b/snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-1
@@ -5,7 +5,7 @@
     "vendor": "com.metabase",
     "name": "csvupload",
     "format": "jsonschema",
-    "version": "1-0-0"
+    "version": "1-0-1"
   },
   "type": "object",
   "properties": {
--- a/snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-1
+++ b/snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-1
@@ -12,7 +12,7 @@
     "event": {
       "description": "Event name",
       "type": "string",
-      "enum": ["csv_upload_successful", "csv_upload_failed"],
+      "enum": ["csv_upload_successful", "csv_upload_failed", "csv_append_successful", "csv_append_failed"],
       "maxLength": 128
     },
     "model_id": {
--- a/src/metabase/analytics/snowplow.clj
+++ b/src/metabase/analytics/snowplow.clj
@@ -71,6 +71,8 @@
    ::action-executed                ::action
    ::csv-upload-successful          ::csvupload
    ::csv-upload-failed              ::csvupload
+   ::csv-append-successful          ::csvupload
+   ::csv-append-failed              ::csvupload
    ::metabot-feedback-received      ::metabot
    ::embedding-enabled              ::embed_share
    ::embedding-disabled             ::embed_share
--- a/src/metabase/analytics/snowplow.clj
+++ b/src/metabase/analytics/snowplow.clj
@@ -32,7 +32,7 @@
   {::account              "1-0-1"
    ::browse_data          "1-0-0"
    ::invite               "1-0-1"
-   ::csvupload            "1-0-0"
+   ::csvupload            "1-0-1"
    ::dashboard            "1-1-4"
    ::database             "1-0-1"
    ::instance             "1-1-2"
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -705,7 +705,7 @@
 
         {:row-count row-count}))
     (catch Throwable e
-      (snowplow/track-event! ::snowplow/csv-upload-failed api/*current-user-id* (fail-stats file))
+      (snowplow/track-event! ::snowplow/csv-append-failed api/*current-user-id* (fail-stats file))
       (throw e))))
 
 (defn- can-append-error
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -469,6 +469,16 @@
 ;;; |  public interface for creating CSV table
 ;;; +-----------------------------------------
 
+(defn- fail-stats
+  "If a given upload / append / replace fails, this function is used to create the failure event payload for snowplow.
+  It may involve redundantly reading the file, or even failing again if the file is unreadable."
+  [file]
+  (with-open [reader (bom/bom-reader file)]
+    (let [rows (csv/read-csv reader)]
+      {:size-mb     (file-size-mb file)
+       :num-columns (count (first rows))
+       :num-rows    (count (rest rows))})))
+
 (mu/defn create-csv-upload!
   "Main entry point for CSV uploading.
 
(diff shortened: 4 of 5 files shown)


snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-0
qnkhuat, line 1: I'm not sure if it's acceptable to modify the enum of a snowplow event, but generally, you can't change a version once it's published; you're supposed to create a new version for it.

so you either can confirm of you can modify enum of a version in #external-snowcat-cloud, or you can do it the normal way by creating a new schema

test/metabase/upload_test.clj
```diff
@@ -1511,6 +1510,41 @@
...
+  ;; Just test with h2 because snowplow should be independent of the driver
+  (mt/test-driver :h2
+    (snowplow-test/with-fake-snowplow-collector
+
+     (with-upload-table! [table (create-upload-table!)]
+       (let [csv-rows ["name" "Luke Skywalker"]
+             file     (csv-file-with csv-rows (mt/random-name))]
+         (append-csv! {:file file, :table-id (:id table)})
+
+         (is (=? {:data {"size_mb"         1.811981201171875E-5
```
qnkhuat, line 1523:
```suggestion
         (is (=? {:data {"size_mb"         (mt/malli=? pos-int?)
```
I'd avoid testing the exact size here

qnkhuat (changes requested):
I think you'll need to create a new schema for the added enum value and we need to assoc `model_id` to the snowplow event data

snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-0
crisptrutski: I guess I just assumed we could add values - will follow up and confirm. 

Is there a way I can test out these events in practice before merging?

test/metabase/upload_test.clj
crisptrutski: Fair enough, there could be some platform sensitivity. What do you think about matching the order of magnitude eat least?

src/metabase/upload.clj
```diff
@@ -636,67 +640,74 @@
...
-                                         :schema-name (:schema table)
-                                         :table-name  (:name table)
-                                         :stats       {:num-rows          row-count
-                                                       :num-columns       (count new-column-types)
-                                                       :generated-columns (if create-auto-pk? 1 0)
-                                                       :size-mb           (file-size-mb file)
-                                                       :upload-seconds    (since-ms timer)}}})
-
-      {:row-count row-count})))
+        (snowplow/track-event! ::snowplow/csv-append-successful api/*current-user-id* stats)
```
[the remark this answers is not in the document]
crisptrutski: We don't have a model in scope here - a model is created in `create-csv-upload!` when we first create the table, but the append action runs directly against the table. 

We could search for the first model created against this table, but that would be brittle as the model could have been archived, deleted, or edit in some semantic way.

snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-0
qnkhuat: Try this https://www.notion.so/metabase/Snowplow-integration-5da1f874beda4153b4fccfa6c1e77caa?pvs=4#6665b4c007d64d1d8e45c5d4c423ac29

crisptrutski:
* **#39951** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#39944** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

test/metabase/upload_test.clj
crisptrutski: I've decided to leave this using the exact size:

1. Our initial upload case uses an explicit number.
2. We don't really have any other tests of the size calculation helper.
3. It seems to be deterministic at least on unix platforms.
4. It's easy to update if the value changes.

src/metabase/analytics/snowplow.clj
```diff
@@ -32,7 +32,7 @@
   {::account              "1-0-1"
    ::browse_data          "1-0-0"
    ::invite               "1-0-1"
-   ::csvupload            "1-0-0"
+   ::csvupload            "1-0-1"
```
crisptrutski, line 35: According to the process doc, we wait until the commit is merged and only then engage our Snowcat Cloud partners to update it. 

So AFAIK this branch is g2g

snowplow/iglu-client-embedded/schemas/com.metabase/csvupload/jsonschema/1-0-1
```diff
@@ -0,0 +1,51 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "CSV upload events",
+  "self": {
+    "vendor": "com.metabase",
+    "name": "csvupload",
+    "format": "jsonschema",
+    "version": "1-0-0"
```
qnkhuat, line 8:
```suggestion
    "version": "1-0-1"
```

src/metabase/upload.clj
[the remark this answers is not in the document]
qnkhuat: hmm, I think the model_id is desirable. I'd see a use case for it in which we want to know how of  X uploads were created, how many of them are then appended.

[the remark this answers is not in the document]
crisptrutski: I'm not fond of guessing the ID, I don't think there's enough value. I'm not sure how we query the snowcat cloud database, but imagine we can join on the upload event there and it'd be more robust.

src/metabase/upload.clj
```diff
@@ -633,67 +637,74 @@
...
+        (scan-and-sync-table! database table)
+
+        (when create-auto-pk?
+          (let [auto-pk-field (table-id->auto-pk-column (:id table))]
+            (t2/update! :model/Field (:id auto-pk-field) {:display_name (:name auto-pk-field)})))
+
+        (events/publish-event! :event/upload-append
+                               {:user-id  (:id @api/*current-user*)
+                                :model-id (:id table)
+                                :model    :model/Table
```
calherries, line 696: If we want to track the model_id that was uploaded from, that is only available on the frontend.

crisptrutski: Feel like I'm missing something in your comment, so apologies if my response doesn't quite speak to it.

We use the underlying table itself as the anchor resource for the event, not the model - this is consistent with the event published on initial upload. For the initial upload the `model_id` for the model that we create is put inside the `:details` map instead.

Using the Model whose page we append from on the frontend is not something I thought of, slightly different but compatible semantics. It still leaves us without a natural id to include if the action was taken from the Collections screen though, so I don't see the harm in just excluding it from both for now.

calherries: resolved on zoom

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: @qnkhuat 

> I'd see a use case for it in which we want to know how of X uploads were created, how many of them are then get appended.

We can calculate that statistic without the model_id, because there is a one-to-one relationship between models created from uploads and tables created from uploads.

We would know how many uploads were created: the number of model IDs from successful initial upload events.
We would know how many uploads have been appended to: the number of distinct table IDs from successful append events.

test/metabase/upload_test.clj
```diff
@@ -1511,6 +1511,42 @@
...
+       (mt/with-dynamic-redefs [upload/load-from-csv! (fn [_ _ _ _] (throw (Exception.)))]
+         (let [csv-rows ["mispelled_name, unexpected_column" "Duke Cakewalker, r2dj"]
+               file     (csv-file-with csv-rows (mt/random-name))]
+           (try
+             (append-csv! {:file file, :table-id (:id table)})
+             (catch Throwable _)
+             (finally
+               (io/delete-file file))))
+
+         (is (= {:data {"size_mb"      5.245208740234375E-5
```
calherries, line 1544:
```suggestion
         (is (= {:data {"size_mb"     5.245208740234375E-5
```

test/metabase/upload_test.clj
```diff
@@ -1511,6 +1511,42 @@
                    (rows-for-table table))))
           (io/delete-file file))))))
 
+(deftest csv-append-snowplow-test
+  ;; Just test with h2 because snowplow should be independent of the driver
+  (mt/test-driver :h2
+    (snowplow-test/with-fake-snowplow-collector
+
+     (with-upload-table! [table (create-upload-table!)]
```
calherries, line 1519: Can you leave some `testing` blocks here? One for the success case, and one for the failure case

calherries (approved):
LGTM

test/metabase/upload_test.clj
crisptrutski: Good idea. I'll add for both creates and appends.

qnkhuat (approved):
okay I think it's okay to leave out model_id then.
LGTM now!

Issue #39946, opened 2024-03-11
Replace all usage of `mc/validate`, `mc/explain`, etc. in `src/` trees and enforce a ban with Kondo


bshepherdson:
These one-shot functions, eg. `(mc/validate schema value)`, compile the Malli schema into a validation function, apply it to `value`, and then throw it away. That's fine in a test or a cold code path, but there are several places where these are applied in production code in the FE, for example when converting from a legacy query.

Compiling the validator for `::lib.schema/query` takes 30x longer than validating a query, so this is extremely wasteful.

We should eliminate all uses of these functions in favour of either:
- Using `metabase.util.malli.registry/validate`/`validator`/`explain`/`explainer` which do their own caching of our schemas.
- `(def some-validator (mc/validator some-schema))` and reusing it `(some-validator value)`.

And then enforce this with Kondo lint rules so that more uses don't crop up. (There's already some rules for this but I don't think they're complete, and there are existing violations.)

PR #39947, opened 2024-03-11
[MLv2] Use cached Malli validators/explainers; make `:fn` schemas cacheable

bshepherdson:
This improves our Malli performance substantially! Many of our code paths were calling vanilla `malli.core/validate` or `malli.core/explain`; these redo the (possibly expensive) compilation of the schema into a validator or explainer for every call.

We should use the caching versions in `metabase.util.malli.registry` everywhere, and our runtime performance will be much improved.

However when I started using these cached versions, I found that the memory use was growing out of control. Eventually I tracked this down to `:fn` schemas. Functions are only comparable by pointer equality, so they make poor cache keys. `(fn ...)` or `comp` calls in a schema on a `mu/defn` function **get re-created for every call of the `mu/defn`'d function**! That's a big time sink recompiling the schemas if we're *not* caching, and a huge memory sink if we *are* caching!

This PR pulls every such schema I could find out into a `def` so it uses the same closure and is cacheable. I'd like to automate that in some part of the Malli pipeline, or maybe a linter rule, but I haven't found a good way to do it yet.

Part of #39946.

### DIFF
--- a/enterprise/backend/src/metabase_enterprise/sandbox/api/table.clj
+++ b/enterprise/backend/src/metabase_enterprise/sandbox/api/table.clj
@@ -7,14 +7,13 @@
    [metabase.lib.util.match :as lib.util.match]
    [metabase.models.card :refer [Card]]
    [metabase.models.data-permissions :as data-perms]
-   [metabase.models.interface :as mi]
    [metabase.models.table :as table :refer [Table]]
    [metabase.util :as u]
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]
    [toucan2.core :as t2]))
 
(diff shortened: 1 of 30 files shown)


bshepherdson:
* **#39947** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39947?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @bshepherdson and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

escherize (approved):
Can't speak to the logic, but using validator etc looks right. It would be nice to figure out how to auto-cache the malli noun-forms (validator, explainer, parser, etc.) for better devex, but that is wayy less important than making the app performant enough.

Issue #39868, opened 2024-03-08
Support column addition / reordering


crisptrutski:
If the user uploads a file with the same column names and compatible types as the existing table, it should Just Work.

Similarly, if the set of columns is a superset, we should backfill with nil and it should also Just Work.

See: [Notion link](https://www.notion.so/metabase/Allow-append-and-replace-to-existing-CSV-uploads-f1f13864632d4cfd83522fce762890eb?pvs=4#a2dab2eb8fff43edbe479a4a44c2ac69)

PR #39948, opened 2024-03-11
Add test for column re-ordering

crisptrutski:
Refs https://github.com/metabase/metabase/issues/39868

### Description

It turns out that we already supported re-ordering, but this added an explicit test to cover that.

It turns out that a bunch of tests were not actually being run, due to mixing up helpers for running with a single driver or a collection of them, so fixed that too.

PR #39951, opened 2024-03-11
Support appending new CSV fields

crisptrutski:
Closes https://github.com/metabase/metabase/issues/39868

### Description

With this change new CSV files can add new fields.

### DIFF
--- a/src/metabase/driver.clj
+++ b/src/metabase/driver.clj
@@ -967,7 +967,7 @@
 (defmulti alter-columns!
   "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
   Currently we do not currently support changing the the primary key, or take any guidance on how to coerce values."
-  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions & args])}
+  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
--- a/src/metabase/driver.clj
+++ b/src/metabase/driver.clj
@@ -967,7 +967,7 @@
 (defmulti alter-columns!
   "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
   Currently we do not currently support changing the the primary key, or take any guidance on how to coerce values."
-  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions])}
+  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions & args])}
   dispatch-on-initialized-driver
   :hierarchy #'hierarchy)
 
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -588,6 +588,9 @@
 (defn- extra-and-missing-error-markdown [extra missing]
   (when (seq missing)
     (->> [[(tru "The CSV file is missing columns that are in the table:") missing]
+          ;; Even though we allow new columns to be implicitly added by uploads, we mention then in the error messages
+          ;; for missing fields as a common case will be the misspelling of names. Seeing the actual and expected
+          ;; names together could help customers spot the root cause more easily.
           [(tru "There are new columns in the CSV file that are not in the table:") extra]]
          (keep (fn [[header columns]]
                  (when (seq columns)
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -628,7 +628,7 @@
         (f relaxed-type))))
 
 (defn- field-changes
-  "Given existing and newly inferred types for the given `fields`, calculate the added and updated fields."
+  "Given existing and newly inferred types for the given `field-names`, calculate which fields need to be added or updated, along with their new types."
   [field-names existing-types new-types]
   (reduce
    (fn [m [f e n]]
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -683,7 +683,7 @@
                                  (let [fields (map normed-name->field normed-header)]
                                    (->> (changed-field->new-type fields old-column-types detected-types)
                                         (alter-columns! driver database table))
-                                   (->> (added-field->new-type header old-column-types detected-types)
+                                   (->> (added-field->new-type normed-header old-column-types detected-types)
                                         (add-columns! driver database table))))
             ;; this will fail if any of our required relaxations were rejected.
             parsed-rows        (parse-rows settings new-column-types rows)
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -616,7 +616,7 @@
         (f relaxed-type))))
 
 (defn- field-changes
-  "Given existing and newly inferred types for the given `fields`, calculate the added and changed fields."
+  "Given existing and newly inferred types for the given `fields`, calculate the added and updated fields."
   [field-keys existing-types new-types]
   (reduce
    (fn [m [f e n]]
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -617,7 +617,7 @@
 
 (defn- field-changes
   "Given existing and newly inferred types for the given `fields`, calculate the added and updated fields."
-  [field-keys existing-types new-types]
+  [field-names existing-types new-types]
   (reduce
    (fn [m [f e n]]
      (cond
@@ -625,21 +625,25 @@
        (not= e n) (assoc-in m [:updated f] n)
        :else      m))
    {:added {}, :updated {}}
-   (map vector field-keys existing-types new-types)))
+   (map vector field-names existing-types new-types)))
 
 (defn- field->db-type [driver field->col-type]
-  (m/map-vals (partial driver/upload-type->database-type driver) field->col-type))
+  (m/map-kv
+   (fn [field-name col-type]
+     [(keyword field-name)
+      (driver/upload-type->database-type driver col-type)])
+   field->col-type))
 
-(defn- add-columns! [driver database table field-key->type & args]
-  (when (seq field-key->type)
+(defn- add-columns! [driver database table field->type & args]
+  (when (seq field->type)
     (apply driver/add-columns! driver (:id database) (table-identifier table)
-           (field->db-type driver field-key->type)
+           (field->db-type driver field->type)
            args)))
 
-(defn- alter-columns! [driver database table field-key->new-type & args]
-  (when (seq field-key->new-type)
+(defn- alter-columns! [driver database table field->new-type & args]
+  (when (seq field->new-type)
     (apply driver/alter-columns! driver (:id database) (table-identifier table)
-           (field->db-type driver field-key->new-type)
+           (field->db-type driver field->new-type)
             args)))
 
 (defn- append-csv!*
@@ -668,8 +672,7 @@
             ;; be parsed as its previous type - there is scope to improve these error messages in the future.
             modify-schema?     (and (not= old-types new-types) (= detected-types new-types))
             _                  (when modify-schema?
-                                 (let [field-keys (map keyword normed-header)
-                                       changes    (field-changes field-keys old-types new-types)]
+                                 (let [changes (field-changes normed-header old-types new-types)]
                                    (add-columns! driver database table (:added changes))
                                    (alter-columns! driver database table (:updated changes))))
             ;; this will fail if any of our required relaxations were rejected.
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -615,45 +615,32 @@
       (when-let [f (allowed-type-upgrades current-type)]
         (f relaxed-type))))
 
-(defn- changed-field->new-type
-  "Given some fields and old and new types, filter out fields with unchanged types, then pair with the new types."
-  [fields old-types new-types]
-  (let [new-if-changed #(when (and %1 (not= %1 %2)) %2)]
-    (->> (map new-if-changed old-types new-types)
-         (map vector fields)
-         (filter second)
-         (into {}))))
+(defn- field-changes
+  "Given the existing and newly inferred types for the given `fields`, return a map of the added and changed fields."
+  [field-keys existing-types new-types]
+  (reduce
+   (fn [m [f e n]]
+     (cond
+       (nil? e)   (assoc-in m [:added f] n)
+       (not= e n) (assoc-in m [:updated f] n)
+       :else      m))
+   {}
+   (map vector field-keys existing-types new-types)))
 
-(defn- add-columns! [driver database table field->type & args]
-  (when (seq field->type)
-    (apply driver/add-columns!
-           driver (:id database) (table-identifier table)
-           (m/map-kv (fn [field-or-name column-type]
-                       [(cond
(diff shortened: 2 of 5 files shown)


crisptrutski:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/metabase/metabase/39951?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#39951** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#39948** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#39944** <a href="https://app.graphite.dev/github/pr/metabase/metabase/39944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

src/metabase/upload.clj
```diff
@@ -573,28 +573,21 @@
...
+      :type/Boolean                ::boolean
+      :type/DateTimeWithTZ         ::offset-datetime
+      :type/DateTime               ::datetime
+      :type/Date                   ::date
+      :type/Text                   ::text)))
+
+(defn- missing-error-markdown [missing]
+  (when (seq missing)
+    (let [header (tru "The CSV file is missing columns that are in the table:")]
+      (str/join "\n" (cons header (map #(str "- " %) missing))))))
```
tsmacdonald, line 590: well that feels nicer (not having to juggle the extra message too)

src/metabase/upload.clj
```diff
@@ -608,35 +601,60 @@
...
+  (or (nil? current-type)
+      (= current-type relaxed-type)
       (when-let [f (allowed-type-upgrades current-type)]
         (f relaxed-type))))
 
 (defn- changed-field->new-type
   "Given some fields and old and new types, filter out fields with unchanged types, then pair with the new types."
   [fields old-types new-types]
-  (let [new-if-changed #(when (not= %1 %2) %2)]
+  (let [new-if-changed #(when (and %1 (not= %1 %2)) %2)]
```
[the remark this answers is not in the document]
tsmacdonald: that said maybe easier to parse if it's `(fn [old new] (when (and old ...)))`

test/metabase/upload_test.clj
```diff
@@ -1351,35 +1351,29 @@
...
-                 (trim-lines "The CSV file contains extra columns that are not in the table:
-                              - extra_2
-                              - extra_1
-
-                              The CSV file is missing columns that are in the table:
+                 (trim-lines "The CSV file is missing columns that are in the table:
                               - id
+                              - name")
+
+                 ["_mb_row_id,id, extra 2"]
```
tsmacdonald, line 1364: Should have a test where one of id or name is provided but not the other, maybe?

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: Yeah I think you're right that it's clearer with named args.

Given that you found the truth table helpful, maybe I'll hoist this to a private function and turn the table into a unit test.

test/metabase/upload_test.clj
```diff
@@ -1634,6 +1628,22 @@
...
+         ;; Reorder as well for good measure
+         (let [csv-rows ["game,name" "Witticisms,Fluke Skytalker"]
+               file     (csv-file-with csv-rows (mt/random-name))]
+           (testing "The new row is inserted with the values correctly reordered"
+             (is (= {:row-count 1} (append-csv! {:file file, :table-id (:id table)})))
+             (is (= [[1 "Obi-Wan Kenobi" nil]
+                     [2 "Fluke Skytalker" "Witticisms"]]
+                    (rows-for-table table))))
+           (io/delete-file file)))))))
+
```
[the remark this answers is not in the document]
crisptrutski: Type promotion is covered in `append-type-coercion-test`, in the following case:

```clojure
{:upload-type ::upload/int, :uncoerced "2.1", :coerced 2.1} ;; column is promoted to float
```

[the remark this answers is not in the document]
crisptrutski: I'd like to reorganize these tests a bit in future, and also add an explicit assertion on the final type.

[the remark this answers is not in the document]
crisptrutski: I plan to use the Replacement epic as impetus for cleaning up the tests a bit.

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: `new-if-changed` is coupled to line 625 because if the value is `nil` then it will be removed from the result.

I _think_ this would be slightly easier to understand if this were a reduction over `fields`, because then you could make this coupling more explicit.

src/metabase/upload.clj
```diff
@@ -603,100 +600,132 @@
...
 
+(defn- add-columns! [driver database table field->type & args]
+  (when (seq field->type)
+    (apply driver/add-columns!
+           driver (:id database) (table-identifier table)
+           (m/map-kv (fn [field-or-name column-type]
+                       [(cond
+                          (keyword? field-or-name) field-or-name
+                          (string? field-or-name) (keyword field-or-name)
+                          (map? field-or-name) (keyword (:name field-or-name)))
```
calherries, line 636: where does this complication come from?

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: I gave it a try with a reduction, and avoiding a helper method, and this is what I got:

```clojure
(reduce 
   (fn [m [field old new]]
     (if (and old new (not= old new))
       (assoc m field new)
       m))
   {}
   (map vector fields old-types new-types))
```

Overall I find the existing code easier to read personally, and the "coupling" seems pretty thin to me. The nil values still have semantic meaning, its incidental that we want a map that filters them out.

src/metabase/upload.clj
```diff
@@ -603,100 +600,132 @@
...
+            new-column-types   (map #(if (matching-or-upgradable? %1 %2) %2 %1) old-column-types relaxed-types)
+            _                  (when (and (not= old-column-types new-column-types)
+                                          ;; if we cannot coerce all the columns, don't bother coercing any of them
+                                          ;; we will instead throw an error when we try to parse as the old type
+                                          (= relaxed-types new-column-types))
+                                 (let [fields (map normed-name->field normed-header)]
+                                   (->> (changed-field->new-type fields old-column-types relaxed-types)
+                                        (alter-columns! driver database table))
+                                   (->> (added-field->new-type normed-header old-column-types relaxed-types)
+                                        (add-columns! driver database table))))
```
calherries, line 688: I read this a few times and figured that `changed-field->new-type` and `added-field->new-type` are mostly doing the same thing, are both complicated by the fact that they need to handle both new fields and existing fields. Splitting them out like the following should remove the need for them.

(I haven't tested this at all, and it most definitely doesn't work. Treat it like pseudocode.)

```clojure
(let [{new-fields      false
       existing-fields true}
      (group-by (comp some? :old-type)
                (map (fn [x y z] {:name x :old-type y, :new-type z})
                     normed-header
                     old-column-types
                     relaxed-types))]
  (alter-columns! driver database (into {}
                                        (for [{:keys [name old-type new-type]} existing-fields
                                              :when (not= old-type new-type)]
                                          [name new-type])))
  (add-columns! driver database (into {}
                                      (for [{:keys [name new-type]} new-fields]
                                        [name new-type]))))
```

calherries: Perhaps this whole section could be extracted into a helper function, since this function is getting quite large.

crisptrutski: Yeah, calculating both change sets together makes a lot of sense, it avoids the "update" logic needing to explicitly check that something isn't new, and lets us deal with the awkward triple-iteration once.

I've gone with a slightly different implementation in the next revision, but taken inspiration.

src/metabase/upload.clj
crisptrutski: This comes from my choice to use first-class field entities back when adding the `alter-columns!` method, and then wanting consistency when I added this proxy method. There wasn't much thought put into it, it just felt natural because the fields where in scope.

You're right, it'll be simpler to just require keywords, and there's very little code to change.

test/metabase/upload_test.clj
crisptrutski: I'm missing something - this precise case that you're commenting on does provide `id`, but not `name`. How is what you're suggesting different from that?

src/metabase/upload.clj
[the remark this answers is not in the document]
crisptrutski: I agree on changing the whole approach. 

Since I'd already put some time into thinking about this implementation, and talked through it with Tim in our 1-1, I thought I'd still share the solution we'd both ended up liking the most.

```clojure
(defn- changed-field->new-type [fields old-types new-types]
  (let [updated?        (fn [o n] (and o (not= o n)))
        type-if-updated (fn [o n] (when (updated? o n) n))]
    (u/zipkeep fields (map type-if-updated old-types new-types))))
```

I suspect it's not to your taste, but interested in how and why.

The new `u/zipkeep` here is like `zipmap`, except it skips key-value pairs where either the key or value is nil. 

Not so sure about the name, also thought about `zipmap-keep` and `zipmap-some`. Just an optimization for `(m/filter-vals some? (zipmap ...))` in this instance. Naming is hard, so luckily we're not adding it today.

src/metabase/upload.clj
crisptrutski: Resolving this orphaned thread, but feel free to pick it up on the new code.

src/metabase/upload.clj
```diff
@@ -607,35 +609,51 @@
...
-         (into {}))))
-
-(defn- alter-columns! [driver database table field->new-type]
-  (driver/alter-columns! driver (:id database) (table-identifier table)
-                         (m/map-kv (fn [field column-type]
-                                     [(keyword (:name field))
-                                      (driver/upload-type->database-type driver column-type)])
-                                   field->new-type)))
+(defn- field-changes
+  "Given existing and newly inferred types for the given `fields`, calculate the added and updated fields."
```
calherries, line 628: I added "with their new types" because an important part about this is that their types are returned as well as calculating which fields are to be added vs updated
```suggestion
  "Given existing and newly inferred types for the given `field-names`, calculate the added and updated field names with their new types."
```

src/metabase/upload.clj
```diff
@@ -570,28 +570,30 @@
...
-  (->> [[(tru "The CSV file contains extra columns that are not in the table:") extra]
-        [(tru "The CSV file is missing columns that are in the table:") missing]]
-       (keep (fn [[header columns]]
-               (when (seq columns)
-                 (str/join "\n" (cons header (map #(str "- " %) columns))))))
-       (str/join "\n\n")
-       (not-blank)))
+  (when (seq missing)
+    (->> [[(tru "The CSV file is missing columns that are in the table:") missing]
+          [(tru "There are new columns in the CSV file that are not in the table:") extra]]
```
calherries, line 591: how about leaving a comment explaining the reason behind this? I don't feel strongly about leaving this in, and someone might come across this and remove it without understanding why we kept it.

test/metabase/upload_test.clj
```diff
@@ -1349,40 +1349,53 @@
...
-
-                 [""]
+                {[""]
                  (trim-lines "The CSV file is missing columns that are in the table:
                               - id
                               - name")
 
+                 ;; Extra columns are fine, as long as none are missing.
+                 ["_mb_row_id,id,extra 1, extra 2,name"]
+                 nil
```
calherries, line 1361: Is this test necessary when we coverage in `append-new-column-test`?

src/metabase/driver.clj
```diff
@@ -967,7 +967,7 @@
 (defmulti alter-columns!
   "Alter columns given by `column-definitions` to a table named `table-name`. If the table doesn't exist it will throw an error.
   Currently we do not currently support changing the the primary key, or take any guidance on how to coerce values."
-  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions])}
+  {:added "0.49.0", :arglists '([driver db-id table-name column-definitions & args])}
```
calherries, line 970: Technically this is a breaking change to the interface of `alter-columns!`. If you're going to do this you need to make a driver changelog entry. Perhaps this is YAGNI

calherries (commented):
Mostly left nits. I'll approve when this is resolved https://github.com/metabase/metabase/pull/39951/files#r1526481802

src/metabase/upload.clj
```diff
@@ -607,35 +609,51 @@
...
+(defn- field-changes
+  "Given existing and newly inferred types for the given `fields`, calculate the added and updated fields."
+  [field-names existing-types new-types]
+  (reduce
+   (fn [m [f e n]]
+     (cond
+       (nil? e)   (assoc-in m [:added f] n)
+       (not= e n) (assoc-in m [:updated f] n)
+       :else      m))
+   {:added {}, :updated {}}
```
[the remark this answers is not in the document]
crisptrutski: Using active verbs feels strange to me, as if the values should be functions. My gut response would be to at least adjectivize them, e.g. :to-add

I think of the result being descriptive, a diff, rather than prescriptive of the specific actions we'll take, and find the current names more natural in that context.

test/metabase/upload_test.clj
crisptrutski: You're right that this overlap is awkward. Since this is the only test that covers inserting multiple columns, and that turned out to be an important edge case, I'm inclined to leave for now.

Let's use the Replacement Arc as an opportunity to factor the tests better in a bunch of ways.

src/metabase/driver.clj
crisptrutski: The asymmetry between the methods bothers me slightly, as does the prospect of a future breaking change - better to break thing before wide adoption.

Will YAGNI for now to get this merged, and hammock a bit on whether we'll ever actually need settings.

src/metabase/upload.clj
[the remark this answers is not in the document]
calherries: sure, won't die on this hill

test/metabase/upload_test.clj
calherries: yeah, maybe the test that explicitly has the title `append-new-column-test` should test inserting multiple columns? It seems more incidental that this tested multiple columns.

src/metabase/driver.clj
calherries: To be clear YAGNI for me means removing the extra args.

I don't mind either way, I am also bothered by the assymetry. But if you keep it, add it to the changelog.

crisptrutski: Aligned. Curious what else we could have meant by YAGNI 😄

test/metabase/upload_test.clj
crisptrutski: @tsmacdonald I'm going to go ahead and merge this PR. If I legitimately missed something here please let me know and I'll fix it in a follow-up.

PR #39955, opened 2024-03-11
[Metrics V2] query processor phase 1

snoe:


src/metabase/lib/schema.cljc
```diff
@@ -144,8 +144,8 @@
     {:error/message ":source-query is not allowed in pMBQL queries."}
     #(not (contains? % :source-query))]
    [:fn
-    {:error/message "A query cannot have both a :source-table and a :source-card."}
-    (complement (every-pred :source-table :source-card))]
+    {:error/message "A query can only have one of :source-table, :source-card, and :sources"}
```
bshepherdson, line 147: I'd make this wording even stronger, to match the condition below: `A query must have exactly one of :source-table, :source-card, or :sources`

src/metabase/lib/schema.cljc
```diff
@@ -161,18 +161,34 @@
...
+   [:ref ::stage.mbql.with-sources]])
 
 ;;; Schema for an MBQL stage that DOES NOT include `:source-table` -- an MBQL stage that is not the initial stage.
 (mr/def ::stage.mbql.without-source
   [:and
    [:ref ::stage.mbql]
    [:fn
-    {:error/message "Only the initial stage of a query can have a :source-table or :source-card."}
-    (complement (some-fn :source-table :source-card))]])
+    {:error/message "Only the initial stage of a query can have a :source-table, :source-metrics, or :source-card."}
```
bshepherdson, line 190: Unless I'm missing something, this is inconsistent with `:sources` elsewhere. I didn't think `:source-metrics` was a thing?

src/metabase/query_processor/middleware/metrics.clj
```diff
@@ -0,0 +1,51 @@
...
+              new-aggregations (->> (mbql.u/replace aggregation
+                                      [:metric {} source-metric] (first source-aggregations))
+                                    lib.util/fresh-uuids)]
+          (-> source-query
+              (lib.util/update-query-stage -1 dissoc :aggregation :breakout :order-by :fields)
+              (reduce-threaded lib/filter filters)
+              (reduce-threaded lib/breakout breakout)
+              (reduce-threaded lib/order-by order-by)
+              (reduce-threaded expression-with-name-from-source expressions)
+              (reduce-threaded lib/aggregate new-aggregations)
```
bshepherdson, line 49: Let's not have `reduce-threaded`. I think it's much easier to read `(as-> source-query $q #_... (reduce lib/filter $q filters))` for this.

test/metabase/query_processor/middleware/metrics_test.clj
```diff
@@ -0,0 +1,119 @@
...
+   [metabase.lib.metadata :as lib.metadata]
+   [metabase.lib.metadata.jvm :as lib.metadata.jvm]
+   [metabase.lib.options :as lib.options]
+   [metabase.lib.test-metadata :as meta]
+   [metabase.lib.test-util :as lib.tu]
+   [metabase.query-processor :as qp]
+   [metabase.query-processor.middleware.metrics :as metrics]
+   [metabase.test :as mt]))
+
+(def counter (atom 2000))
```
bshepherdson, line 16: Consider randomizing this to prevent the tests from accidentally relying on it.
```suggestion
(def counter (atom (rand-int 10000)))
```

test/metabase/query_processor/middleware/metrics_test.clj
```diff
@@ -0,0 +1,119 @@
...
+    (is (=?
+          {:stages [{:source-table (meta/id :products)
+                     :aggregation [[:avg {} [:field {} (meta/id :products :rating)]]
+                                   [:+ {} [:avg {} [:field {} (meta/id :products :rating)]] 1]]}]}
+          (metrics/expand query)))))
+
+(deftest expand-basic-test
+  (let [[source-metric mp] (mock-metric)
+        query (-> (lib/query mp source-metric)
+                  (add-metric-source source-metric))]
```
bshepherdson, line 60: This feels like a repeated block that's worth a helper function around `(mock-metric)`. Like:
```clojure
(defn- metric-query [& args]
  (let [[source-metric mp] (apply mock-metric args)]
    (-> (lib/query mp source-metric)
        (add-metric-source source-metric))))
```

bshepherdson: Oh, with maybe an optional function to go between the query and adding the metric aggregation, since I see that a bunch of things add filters etc. before it and the order might matter.

bshepherdson (approved):

test/metabase/query_processor/middleware/metrics_test.clj
snoe: `add-metric-source` should get wiped out once the `lib` functions exist to build these queries.

PR #39957, opened 2024-03-11
No view logging for oss

escherize:
Follow on fix for #35288

We mustn't record view logging in OSS, but we were anyway.


test/metabase/events/view_log_test.clj
```diff
@@ -30,6 +31,18 @@
             :context    "question"}
            (latest-view (u/id user) (u/id card)))))))
 
+;; we need to run this test in oss only:
+
+(deftest card-read-oss-no-view-logging-test
+  (if config/ee-available?
+    (is true "Only test in OSS")
```
noahmoss, line 38: Never seen a test assertion used as documentation like this. Might be simpler to just use a `when-not`?

noahmoss (approved):

PR #39958, opened 2024-03-11
Extract action column placement

ranquild:
To put the newly created column in the right place, we need to set `table.columns` viz setting. It requires the following props: `name`, `field_ref`. Later, when displaying the query results, vizualizations match these settings with columns from query results (coming from the QP). So, do we need to get the name and ref of the column and put it into the settings? Nope, this won't work because of the following issues:
- `name` and `field_ref` computed by CLJS and QP differ. When viz calls `getColumnKey`, it won't match the setting with the query column if the setting was produced with CLJS input. This is so far the biggest challenge in this project. So, to make it work, we need to somehow "remember" the current drill and update the viz settings when we receive new query results.
- viz settings are "computed", and only manually changed parts are set in the card object. We can get the up-to-date computed settings with `getVisualizationSettings` selector, or via `computedSettings` property available in viz and click actions. This does not affect the final solution but was a big issue when I was trying different approches.
- viz settings are kept in-sync with the query by inspecting query results and comparing them with the settings. If there is a new column in the results, for instance, for which there is no `table.columns` entry, such an entry would be automatically created. This makes new columns in the query automatically appear in query results. 
- viz settings sync mechanism relies not only the latest query results but also on previous query results. There was a bug in the code where such results before removed from the redux state, and the sync process didn't work as expected. Unfortunately we cannot stop removing such query results, and the alternative I came up with is manually passing them in case where we remove them from the store.
- by default the viz settings sync process doesn't change `table.columns` property if it has the default value. We have to adjust that to re-generate the viz settings for the columns to be able to manipulate them.

The final solution does the following:
- Passes `column` object from the `column-extract` drill to the viz sync process
- Compares query results with previous query results and finds out what the new columns are (if multiple)
- Finds the corresponding viz setting for the column selected by the drill
- Moves settings for the newly created columns just after the selected column

How to verify:
- Orders -> Created At -> Extract ... -> Year
- Make sure the new column is placed correctly

Screenshot 2024-03-12 at 17 31 26


frontend/src/metabase-lib/Question.ts
```diff
@@ -470,110 +465,24 @@ class Question {
     });
   }
 
-  private _syncGraphMetricSettings(
-    { data: { cols = [] } = {} }: Dataset,
-    { data: { cols: prevCols = [] } = {} }: Dataset,
+  syncColumnsAndSettings(
```
ranquild, line 468: Moved to `column_settings.ts`

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,181 @@
...
+    );
+
+    if (prevQueryResults && !prevQueryResults.error) {
+      newSettings = syncGraphMetricSettings(
+        newSettings,
+        queryResults,
+        prevQueryResults,
+      );
+
+      if (clicked) {
```
ranquild, line 31: New part

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,181 @@
...
+          clicked,
+        );
+      }
+    }
+  }
+
+  return newSettings;
+}
+
+function syncTableColumnSettings(
```
ranquild, line 45: The new bit here is `isDefaultUnwrapped`

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,181 @@
...
+    fieldRef: col.field_ref,
+    enabled: true,
+  }));
+  return {
+    ...settings,
+    "table.columns": [...existingColumnSettings, ...addedColumnSettings],
+  };
+}
+
+function syncTableColumnSettingsAfterClickAction(
```
ranquild, line 93: Unchanged

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,181 @@
...
+    ...settings,
+    "table.columns": [
+      ...existingColumnSettings.slice(0, columnSettingIndex + 1), // before the selected column
+      ...addedColumnSettings,
+      ...existingColumnSettings.slice(columnSettingIndex + 1), // after the selected column
+    ],
+  };
+}
+
+function syncGraphMetricSettings(
```
ranquild, line 134: Unchanged

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,188 @@
...
+    fieldRef: col.field_ref,
+    enabled: true,
+  }));
+  return {
+    ...settings,
+    "table.columns": [...existingColumnSettings, ...addedColumnSettings],
+  };
+}
+
+function moveNewTableColumnsAfterColumn(
```
ranquild, line 100: New

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,188 @@
...
+  { data: { cols } }: Dataset,
+  { data: { cols: prevCols } }: Dataset,
+  column: DatasetColumn,
+): VisualizationSettings {
+  const columnSettings = settings["table.columns"];
+  if (!column || !columnSettings) {
+    return settings;
+  }
+
+  const prevColumnNames = new Set(prevCols.map(col => col.name));
```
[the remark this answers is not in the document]
ranquild: Yes, column names are unique within the query results

frontend/src/metabase-lib/viz/column_settings.ts
```diff
@@ -0,0 +1,188 @@
...
+  prevQueryResults?: Dataset,
+  options?: SettingsSyncOptions,
+): VisualizationSettings {
+  let newSettings = settings;
+
+  if (queryResults && !queryResults.error) {
+    newSettings = syncTableColumnSettings(
+      newSettings,
+      queryResults,
+      options == null,
```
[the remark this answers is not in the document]
ranquild: Fixed - passed options directly

e2e/test/scenarios/visualizations-tabular/drillthroughs/column_extract_drill.cy.spec.js
```diff
@@ -77,6 +79,42 @@ describe("extract action", () => {
...
+          column: "Created At",
+          columnIndex,
+        });
+        checkColumnIndex({
+          column: "Year",
+          columnIndex: columnIndex + 1,
+        });
+      });
+
+      it("saved question without viz settings", () => {
```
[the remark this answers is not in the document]
ranquild: Added a test

kamilmielnik (approved):
That was a hard one :1st_place_medal:

PR #39969, opened 2024-03-12
E2E support for CRUD + running pMBQL queries

### LINKED ISSUE #39024: BE - Support MLv2 queries end-to-end (REST API and application database)
We need to support MLv2/pMBQL queries at the REST API and application database level in order to power Metrics v2. Metrics v2 will require some changes to MBQL, and we don't want to have to backport these changes to legacy MBQL as well.

@ranquild confirmed that the FE is ready to go WRT to sending MLv2 queries over the wire without converting to legacy (and vice versa), with the exception of QP results metadata, which we should leave in legacy format for the near-term future. See https://metaboat.slack.com/archives/C051AG38B2S/p1708542180586809 for more info. 

Right now the QP supports pMBQL queries at the `qp/process-query` level -- if it sees them, it will convert them to legacy, since the QP middleware is still written for legacy at the time of this writing. So nothing needs to change in QP internals as part of this issue.

We do however need to change the various REST API endpoints to make sure they accept MLv2 queries **in addition to** legacy queries:

- ad-hoc query endpoints like `POST /api/dataset`
- endpoints for running queries saved in the app DB like `POST /api/card/:id/query` and the DashCard version
- endpoints for saving Cards like `POST /api/card` 

These endpoints should leave the MLv2 queries as-is and let the QP do whatever it wants with it. **We should start saving MLv2 queries in the application database as MLv2**, which means some code will need to be updated to be cool with this (e.g. stuff that pulls Cards out of the app to get statistics or stuff like that about them). We can either have these sorts of places convert pMBQL to legacy as needed or convert from legacy to pMBQL as needed and update them to use MLv2 instead of digging into the query manually. The latter is obviously preferrable (our ultimate goal is to use MLv2 everywhere) but it is more work and ok to punt to a later date.

- The behavior for legacy queries should remain unchanged at this point in time. 
- Endpoints that return queries should return legacy queries as legacy and MLv2 queries as MLv2
- MLv2 queries should get saved as and come out of the application database as MLv2; legacy should get saved as and come out as legacy (for the near-term, at least)

The MLv2 code used by the FE should already be smart enough to do the right thing when given either a MLv2 or legacy query. We might need to tweak the MLv2 normalization code a bit tho so a JSON-serialized MLv2 code gets deserialized correctly for use in the FE (or when coming out of the app DB) -- not sure how much needs to be done there or not; it may be worth creating a separate subissue if it turns out to be a bigger project.

We should not update the Metrics v1 and Segments v2 endpoints or related code, since these are being deprecated we can leave them as legacy-only.

camsaul:
Support saving MLv2 queries in the application database, and running them from the ad-hoc `POST /api/dataset` REST API endpoint as well as the Card and Dashcard endpoints. I will follow this on with more PRs to support other things like pivot queries (#40141) and other things

Resolves #39024


metamben (approved):
I need this!

PR #39988, opened 2024-03-12
Remove DELETE endpoints for cards and dashboards

tsmacdonald:
Archive them instead, as we have been for 7 years


calherries (commented):
We use these endpoints. 

Go to `/archive`

image

You can delete cards and dashboards

image

tsmacdonald:
This is going to change, and the endpoints are almost certainly _not_ deprecated, but I'm delaying decisionmaking until https://metaboat.slack.com/archives/C02H619CJ8K/p1708528706182949 is sorted out

calherries:
@tsmacdonald we should probably do this at least anyway [Remove TODOs and warnings around hard deletes](https://github.com/metabase/metabase/pull/39994)

PR #39992, opened 2024-03-12
Move extracted column after the column for which the drill was applied for

ranquild:


frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -470,110 +465,19 @@ class Question {
...
 
-  private _syncGraphMetricSettings(
-    { data: { cols = [] } = {} }: Dataset,
-    { data: { cols: prevCols = [] } = {} }: Dataset,
+  syncColumnsAndSettings(
+    queryResults?: Dataset,
+    prevQueryResults?: Dataset,
+    clicked?: Lib.ClickObject,
   ) {
-    const graphMetrics = this.setting("graph.metrics");
```
ranquild, line 477: Moved to a separate file

PR #39994, opened 2024-03-12
Remove TODOs and warnings around hard deletes

calherries:
We have some warnings and TODOs in the code for hard-deleting cards and dashboards. 

These are super out-of-date. We've been using these endpoints for years for hard-deleting cards and dashboards from the /archive page. 

In the future we might change the /archive page to also soft-delete, but we'll always need to offer a way to hard delete things for legal requirements.


tsmacdonald (approved):
Yeah, shipit, true for now. Looks like it's going to change soon anyway, but…cross that bridge when we get there.

PR #39996, opened 2024-03-12
Repro #29943 - Wrong field highlighted when editing model metadata if custom column added to left of column

### LINKED ISSUE #29943: Wrong field highlighted when editing model metadata if custom column added to left of column
**Describe the bug**
When editing model metadata on a model with a custom column in front of the non-custom columns, clicking a column name highlights the wrong column (seems to be offset by the number of custom columns).

**Logs**
image

**To Reproduce**
Steps to reproduce the behavior (if you can reproduce the bug using the Sample Database, we will find the issue faster):
1. Create and save a question from Accounts table in Sample Database with all columns.
2. Turn the question into a model
3. Edit query definition, add custom column "Seating type" with `CASE([Seats] = 1, "Single", [Seats] = 2, "Double", "Other")`
4. Move new column to be the first column
5. Save model
6. Edit metadata, select any column name, notice that the column 1 to the left is highlighted instead of the column selected.

**Expected behavior**
- clicked column is expected to be highlighted

**Screenshots/videos**
- After clicking "Seats" at the top, "Source" is highlighted instead:
- Have tried with 2 custom columns, and the offset is 2 instead of 1 column off

image

- Your browser and the version: Chrome
- Your operating system: MacOS
- Your databases: H2
- Metabase version: master (as of 2023-04-10) and 45.3.1

**Severity**
- Confusing user experience

kamilmielnik:
Closes #29943

Stress test: https://github.com/metabase/metabase/actions/runs/8249658549

### DIFF
--- a/e2e/test/scenarios/models/reproductions/29943-custom-column-in-metadata.cy.spec.ts
+++ b/e2e/test/scenarios/models/reproductions/29943-custom-column-in-metadata.cy.spec.ts
@@ -1,0 +1,89 @@
+import { SAMPLE_DATABASE } from "e2e/support/cypress_sample_database";
+import { openQuestionActions, popover, restore } from "e2e/support/helpers";
+
+const { ORDERS, ORDERS_ID } = SAMPLE_DATABASE;
+
+describe("issue 29943", () => {
+  beforeEach(() => {
+    restore();
+    cy.signInAsAdmin();
+    cy.intercept("POST", "/api/dataset").as("dataset");
+  });
+
+  it("selects the right column when clicking a column header (metabase#29943)", () => {
+    cy.createQuestion(
+      {
+        type: "model",
+        query: {
+          "source-table": ORDERS_ID,
+          expressions: {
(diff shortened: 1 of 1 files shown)


e2e/test/scenarios/models/reproductions/29943-custom-column-in-metadata.cy.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function assertColumnSelected(columnIndex: number, name: string) {
+  getHeaderCell(columnIndex, name)
+    .find("div")
+    .should("have.css", "background-color")
+    .and("eq", "rgb(80, 158, 227)");
+
+  cy.findByLabelText("Display name").should("have.value", name);
+}
```
nemanjaglumac, line 82: It's unfortunate that we don't have a better way to determine whether or not a column is selected.

nemanjaglumac (approved):
This was very easy to read/follow.
🙌

PR #40006, opened 2024-03-12
TableDefinition can define a composite PK

calherries:
Follows on from https://github.com/metabase/metabase/pull/39038

Allows us to create composite PKs in test datasets by having multiple FieldDefinitions with `{:pk? true}`.

It might be more future-proof to allow defining a PK order separate from the order of the fields, but I don't need that right now and we can easily change it later.

### DIFF
--- a/test/metabase/test/data/sql.clj
+++ b/test/metabase/test/data/sql.clj
@@ -240,21 +240,23 @@
         inline-comment (inline-column-comment-sql driver field-comment)]
     (str/join " " (filter some? [field-name field-type not-null unique inline-comment]))))
 
-(defn fielddefs->pk-field-name
-  "Find the pk field name in fieldefs"
+(defn fielddefs->pk-field-names
+  "Find the pks field name in fieldefs"
   [fieldefs]
-  (->> fieldefs (filter :pk?) first :field-name))
+  (->> fieldefs (filter :pk?) (map :field-name)))
 
 (defmethod create-table-sql :sql/test-extensions
   [driver {:keys [database-name], :as _dbdef} {:keys [table-name field-definitions table-comment]}]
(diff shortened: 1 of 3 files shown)


qnkhuat (approved):
Lovely

test/metabase/test/data/sql.clj
```diff
@@ -285,11 +287,15 @@
...
-  (let [quot               #(sql.u/quote-name driver %1 (ddl.i/format-name driver %2))
-        dest-table-name    (name dest-table-name)
-        dest-table-pk-name (->> (get-tabledef dbdef dest-table-name)
-                                :field-definitions
-                                fielddefs->pk-field-name)]
+  (let [quot                #(sql.u/quote-name driver %1 (ddl.i/format-name driver %2))
+        dest-table-name     (name dest-table-name)
+        pk-names            (->> (get-tabledef dbdef dest-table-name)
+                                 :field-definitions
+                                 fielddefs->pk-field-names)
```
crisptrutski, line 294: nit: indentation is off here - it seems from `dest-table-pk-name` first being pluralized, and only later being de-prefixed to `pk-names`.

test/metabase/test/data/sql.clj
```diff
@@ -240,21 +240,23 @@
         inline-comment (inline-column-comment-sql driver field-comment)]
     (str/join " " (filter some? [field-name field-type not-null unique inline-comment]))))
 
-(defn fielddefs->pk-field-name
-  "Find the pk field name in fieldefs"
+(defn fielddefs->pk-field-names
+  "Find the pks field name in fieldefs"
```
crisptrutski, line 244: This comment is the only place you pluralize "pk", and it's inconsistent with the function name:

```suggestion
  "Find the pk field names in fieldefs"
```

crisptrutski (approved):

PR #40007, opened 2024-03-12
Global to local css migration - grabber

deniskaber:

### Description

This migrates global CSS rules from `grabber.module.css` to be locally scoped.

### How to verify


frontend/src/metabase/css/core/colors.module.css
```diff
@@ -100,6 +100,7 @@
 
 /* white  - move to bottom for specificity since its often used on hovers, etc */
 :global(.text-white),
+.text-white,
```
oisincoveney, line 103: do we still need to add this one?

oisincoveney (approved):
just the one question (plus tests passing obv), but looks good otherwise

PR #40008, opened 2024-03-12
Move search to toplevel in accordionlist

### LINKED ISSUE #41085: Make search surface columns when I’m working on joined tables
[Product doc](https://www.notion.so/metabase/Make-search-surface-columns-when-I-m-working-on-joined-tables-bc79b831564a4db2a646ab7a70b86580)

### Goal
Make searching on joined data in filters etc. more usable by surfacing column names as well.

- [ ] Update `AccordionList` to allow searching of all sections at once
- [ ] Keep supporting the previous usecases of AccordionList

romeovs:
Closes https://github.com/metabase/metabase/issues/41085

Moves the search box in the accordion list to the top level.

Screenshot 2024-03-12 at 17 51 17

Screenshot 2024-04-01 at 12 56 52

### DIFF
--- a/frontend/src/metabase/admin/datamodel/metadata/components/SemanticTypeAndTargetPicker/SemanticTypeAndTargetPicker.tsx
+++ b/frontend/src/metabase/admin/datamodel/metadata/components/SemanticTypeAndTargetPicker/SemanticTypeAndTargetPicker.tsx
@@ -111,6 +111,7 @@ const SemanticTypeAndTargetPicker = ({
         optionSectionFn={getTypeOptionSection}
         placeholder={t`Select a semantic type`}
         searchProp="name"
+        globalSearch
       />
       {showCurrencyTypeSelect && hasSeparator && <FieldSeparator />}
       {showCurrencyTypeSelect && (
--- a/frontend/src/metabase/core/components/AccordionList/AccordionList.jsx
+++ b/frontend/src/metabase/core/components/AccordionList/AccordionList.jsx
@@ -62,6 +62,7 @@ export default class AccordionList extends Component {
     sections: PropTypes.array.isRequired,
 
     initiallyOpenSection: PropTypes.number,
+    globalSearch: PropTypes.bool,
     openSection: PropTypes.number,
     onChange: PropTypes.func,
     onChangeSection: PropTypes.func,
@@ -104,6 +105,7 @@ export default class AccordionList extends Component {
   static defaultProps = {
     style: {},
     width: 300,
+    globalSearch: false,
     searchable: section => section.items && section.items.length > 10,
     searchProp: "name",
     searchCaseInsensitive: true,
@@ -390,15 +392,20 @@ export default class AccordionList extends Component {
     itemIsSelected,
     hideEmptySectionsInSearch,
     openSection,
+    _globalSearch,
+    searchText,
   ) => {
+    // if any section is searchable just enable a global search
+    let globalSearch = _globalSearch;
+
     const sectionIsExpanded = sectionIndex =>
-      alwaysExpanded || openSection === sectionIndex;
+      alwaysExpanded ||
+      openSection === sectionIndex ||
+      (globalSearch && searchText?.length > 0);
+
     const sectionIsSearchable = sectionIndex =>
       searchable &&
       (typeof searchable !== "function" || searchable(sections[sectionIndex]));
-
-    // if any section is searchable just enable a global search
-    let globalSearch = false;
 
     const rows = [];
     for (const [sectionIndex, section] of sections.entries()) {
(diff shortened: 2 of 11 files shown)


cdeweyx (approved):

AlessioLaiso (approved):

frontend/src/metabase/querying/components/FilterPicker/FilterColumnPicker/FilterColumnPicker.tsx
```diff
@@ -29,16 +29,20 @@ export interface FilterColumnPickerProps {
...
+  alwaysVisible?: boolean;
 };
 
 const CUSTOM_EXPRESSION_SECTION: Section = {
   key: "custom-expression",
+  type: "action",
   name: t`Custom Expression`,
   items: [],
   icon: "filter",
+  alwaysVisible: true,
```
kamilmielnik, line 45: I understand that currently `alwaysVisible` will be `true` only when `type` is `"action"`, so this information is equivalent.
Do we plan to allow for different permutations?
If not, WDYT about dropping `alwaysVisible` and assuming sections of type `"action"` are always visible?

frontend/src/metabase/core/components/AccordionList/AccordionList.jsx
```diff
@@ -62,6 +62,7 @@ export default class AccordionList extends Component {
     sections: PropTypes.array.isRequired,
 
     initiallyOpenSection: PropTypes.number,
+    globalSearch: PropTypes.bool,
```
kamilmielnik, line 65: In Admin -> Table metadata -> Semantic type input, when searching, section names will no longer be displayed. Is this expected?

See [`master`](https://github.com/metabase/metabase/assets/6830683/b8e51e46-28d0-4060-9de5-9bd48dda9d50) vs [this PR](https://github.com/metabase/metabase/assets/6830683/631eb015-2c21-4909-b5a4-79d3d9358f10)

kamilmielnik (commented):
Wow, this `AccordionList` component looks... adventurous.
Good job finding your way in!

We could use Mantine for the new section type, but given the state of this component I think it's ok to skip it.

I left some comments with possibly unexpected differences in behavior that I found - maybe they're expected, but it's better to double-check.
I believe 1 test has been updated incorrectly.

Can you please link relevant issue/epic/Slack link in the PR description?

frontend/src/metabase/core/components/AccordionList/AccordionList.jsx
romeovs: Good catch: that was a bug. It now works as expected:

Screenshot 2024-04-04 at 17 03 07


Screenshot 2024-04-04 at 17 04 01

frontend/src/metabase/core/components/AccordionList/AccordionList.jsx
```diff
@@ -480,6 +498,18 @@ export default class AccordionList extends Component {
     }
 
     if (globalSearch) {
+      const isSearching = searchText.length > 0;
+      const isEmpty = rows.filter(x => x.type === "item").length === 0;
```
kamilmielnik, line 502:
```suggestion
      const isEmpty = rows.filter(row => row.type === "item").length === 0;
```

frontend/src/metabase/admin/datamodel/metadata/components/SemanticTypeAndTargetPicker/SemanticTypeAndTargetPicker.tsx
```diff
@@ -111,6 +111,7 @@ const SemanticTypeAndTargetPicker = ({
         optionSectionFn={getTypeOptionSection}
         placeholder={t`Select a semantic type`}
         searchProp="name"
+        globalSearch
```
kamilmielnik, line 114: `globalSearch` reminds me of search in the app header or in the command palette.

How about `allSectionsSearch`?

kamilmielnik (approved):
Looks good, works well :+1: 
Approving although I'm not sure about the status of this project (is this PoC?).

@romeovs Can you please link relevant issue/epic/Slack/docs in the PR description?

frontend/src/metabase/core/components/AccordionList/AccordionList.jsx
```diff
@@ -104,6 +105,7 @@ export default class AccordionList extends Component {
   static defaultProps = {
     style: {},
     width: 300,
+    globalSearch: false,
```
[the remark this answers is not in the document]
romeovs: @AlessioLaiso what do you think?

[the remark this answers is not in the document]
AlessioLaiso: Romeo and I just caught up on this and yes, let's add dividers where needed, like here. This matches the new designs for custom expressions too, where the various expression shortcuts are separated by a divider from the "custom expression" option.

PR #40009, opened 2024-03-12
Fix flaky redshift tests due to missing late_binding_view

calherries:
I saw a flaky test where the `late_binding_view` table was missing during sync. I lost the link to an example test run unfortunately. 

I'm pretty sure there's no harm in keeping an extra table around, and this should fix the flake.

### DIFF
--- a/modules/drivers/redshift/test/metabase/driver/redshift_test.clj
+++ b/modules/drivers/redshift/test/metabase/driver/redshift_test.clj
@@ -227,31 +227,25 @@
             view-nm      "late_binding_view"
             qual-view-nm (format "\"%s\".\"%s\"" (redshift.test/unique-session-schema) view-nm)]
         (t2.with-temp/with-temp [Database database {:engine :redshift, :details db-details}]
-          (try
-            ;; create a table with a CHARACTER VARYING and a NUMERIC column, and a late bound view that selects from it
-            (execute!
-             (str "DROP TABLE IF EXISTS %1$s;%n"
-                  "CREATE TABLE %1$s(weird_varchar CHARACTER VARYING(50), numeric_col NUMERIC(10,2));%n"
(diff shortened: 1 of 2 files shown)


modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -227,31 +227,25 @@
...
-              (is (= [{:name "numeric_col",   :database_type "numeric(10,2)",         :base_type :type/Decimal}
-                      {:name "weird_varchar", :database_type "character varying(50)", :base_type :type/Text}]
-                     (map
-                      mt/derecordize
-                      (t2/select [Field :name :database_type :base_type] :table_id table-id {:order-by [:name]})))))
-            (finally
-              (execute! (str "DROP TABLE IF EXISTS %s;%n"
-                             "DROP VIEW IF EXISTS %s;")
-                        qual-tbl-nm
-                        qual-view-nm))))))))
```
calherries, line 254: No dropping `late_binding_view` anymore.

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -227,31 +227,25 @@
...
+            (is (= [{:name "numeric_col",   :database_type "numeric",         :base_type :type/Decimal}
+                    {:name "weird_varchar", :database_type "character varying", :base_type :type/Text}]
```
qnkhuat, line 245:
```suggestion
            (is (= [{:name "numeric_col",   :database_type "numeric(10,2)",         :base_type :type/Decimal}
                    {:name "weird_varchar", :database_type "character varying(50)", :base_type :type/Text}]
```

calherries: Thanks for committing this, I accidentally brought these over from another PR

Issue #39564, opened 2024-03-04
[dc.js migration] histogram x-axis labels placed differently


EmmadUsmani:
question/2772
https://metaboat.slack.com/archives/C06G94JTWBS/p1706193714570929

![image](https://github.com/metabase/metabase/assets/37751258/1396891e-f482-481b-9180-92a7085115eb)

question/14933
https://metaboat.slack.com/archives/C06G94JTWBS/p1706193940915449

question/5834

PR #40026, opened 2024-03-12
fix histogram label positioning

EmmadUsmani:
Closes https://github.com/metabase/metabase/issues/39564

### Description

To fix the inaccurate placement, there are two things we need to do:

1. Position the labels on the left of the bar, instead of the center

We do this by adding a `padding` option to `xAxis.axisLabel` in `option/axis.ts`

2. Add a label for the end of the last bin

3. Add padding to the sides of the bars (wasn't mentioned in the original issue, but it's also a regression from the old implementation)

We achieve both of these by adding two extra values to the chart dataset, one at the beginning where the x-axis value is less than the minimum dimension value by one bin interval, and one at the end where the x-axis value is greater than the maximum dimension value by one bin interval.

Note that the padding is larger than it was on the old implementation. Unfortunately there's no easy way to fix this, the padding has to be one dimension width, echarts does not allow customization of a single datum's width in a bar series.

### Checklist


EmmadUsmani:
* **#40026** <a href="https://app.graphite.dev/github/pr/metabase/metabase/40026?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `echarts`

<h2></h2>

 Join @EmmadUsmani and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

frontend/src/metabase/visualizations/echarts/cartesian/model/dataset.ts
```diff
@@ -577,3 +577,24 @@ export const getBubbleSizeDomain = (
...
+  dataset: ChartDataset,
+  xAxisModel: XAxisModel,
+) {
+  if (!isCategoryAxis(xAxisModel) || !xAxisModel.isHistogram) {
+    return dataset;
+  }
+
+  const interval = xAxisModel.histogramInterval ?? 1;
+
+  dataset.unshift({
```
alxnddr, line 591: This should be breaking tooltips and click behavior since now the `dataIndex` of the transformed dataset does not match the dataIndex of the orignial dataset which stores required values

frontend/src/metabase/visualizations/echarts/cartesian/model/index.ts
```diff
@@ -105,6 +106,8 @@ export const getCartesianChartModel = (
     renderingContext,
   );
 
+  dataset = getHistogramDataset(dataset, xAxisModel);
```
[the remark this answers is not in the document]
EmmadUsmani: Makes sense I can move it

frontend/src/metabase/visualizations/echarts/cartesian/model/dataset.ts
EmmadUsmani: Screenshot 2024-03-12 at 11 47 09 AM

Tooltips seem to work fine for me

EmmadUsmani: https://github.com/metabase/metabase/assets/37751258/a8c40821-def4-4e40-a723-3a86935b17b1

Click behavior as well

alxnddr: I missed that initially you changed the raw dataset instead of the transformed one, so now there is an issue with tooltips and clicks since the dataIndex points to a wrong datum:
https://github.com/metabase/metabase/assets/14301985/f1f940c5-4429-4d09-9209-d2daeed4d344
But it should be straightforward to fix

EmmadUsmani: Ah that makes sense I'll fix it, I should just be able to offset the index if it's a histogram

alxnddr: I think simply adding an offset won't work when there was `null` which was removed. But we can store the original datum index

EmmadUsmani: @alxnddr Should be fixed now

alxnddr (commented):
I think a simpler solution that involves adding an optional key `const ORIGINAL_INDEX_KEY = ...` to the transformed dataset would do the trick do that the correct data index for events would be `transformedDataset[dataIndex][ORIGINAL_INDEX_KEY] ?? dataIndex` which would not require changing other cartesian charts like the Waterfall one, but you can keep the existing solution. Now the waterfall chart needs to be fixed — events and clicks stopped working

EmmadUsmani:
@alxnddr Makes sense, I implemented this change on another PR that also needs it and made that the parent of this PR

https://github.com/metabase/metabase/pull/40154

alxnddr (approved):

EmmadUsmani:
### Merge activity

* **Mar 15, 8:43 PM EDT**: @EmmadUsmani started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/metabase/metabase/40026).

PR #40028, opened 2024-03-12
docs - updates api docs script to handle exception

jeff-bruemmer:
Updates API docs to handle advanced-config namespace. 

See https://github.com/metabase/metabase/pull/38864.


src/metabase/api/common/internal.clj
```diff
@@ -27,6 +27,27 @@
 ;;; |                                              DOCSTRING GENERATION                                              |
 ;;; +----------------------------------------------------------------------------------------------------------------+
 
+(defn handle-nonstandard-namespaces
```
jeff-bruemmer, line 30: Consolidated exceptions into a single function.

camsaul (approved):

PR #33415, opened 2023-08-22
Create a privilege table and update it on sync for postgres

calherries:
Epic: https://github.com/metabase/metabase/issues/32978

This PR implements storing data for postgres DBs on sync to know whether the DB connection's current user has access to select, update, insert, or delete rows for a given table.

It creates a new table called `table_privileges` that stores privileges by table for the connection's user and all other roles with privileges on the table.

The goal is to be able to query the set of privileges for a metabase user on a table. The `table_privileges` table can be joined against `metabase_table` to determine whether a user can select, update, insert, or delete rows from the table. The table is synced with the privileges in the customer's database on DB sync. It's important to note the privileges in the customer database and those in the application DB can get out of sync with one another. There's no way around this, but at least DB admins have the power to sync them using the existing database sync button.

`table_privileges` has the following columns:
- `table_id`: int, non-nullable, foreign key referencing `metabase_table(id)`.
- `role`: varchar(255): The role name in the database. This will be used in the future when DBs with connection impersonation on are supported. Currently all values of this column are NULL, which means that the privileges apply to the connection's current user
- `select`: boolean: If true, the role should be allowed to select from the table.
- `update`: boolean: If true, the role should be allowed to update records in the table.
- `insert`: boolean: If true, the role should be allowed to insert records into the table.
- `delete`: boolean: If true, the role should be allowed to delete records from the table.

**Connection impersonation plans**

To support DBs with connection impersonation ([product doc here](https://www.notion.so/metabase/New-sandbox-mode-connection-impersonation-49ecf43c3ffa4983883131c2af10580f)) we will add more rows to `table_privileges` for each role and table in the customer's DB. When the user has connection impersonation enabled for a particular DB, we need to select the subset of rows where `connection_impersonations.attribute = table_privileges.role`.

We can safely ignore connection impersonation at the moment because actions aren't likely to be used with connection impersonation enabled, as yet.

When we create any actions features that use this data, we can assume that no tables with connection impersonation enabled are editable, effectively banning users from running basic actions, even if they had insert/update/delete access.

PR #40034, opened 2024-03-12
Postgres: fix sync tables with partial select perm

### LINKED ISSUE #40338: postgresql tables that are visible through column grants aren't available for simple questions
### Describe the bug

As part of CSV uploads introduced in 0.47 via https://github.com/metabase/metabase/pull/33415 metabase began excluding postgresql tables that have only a subset of their columns selectable from the schema for simple questions (you can still query them via native sql). This appears to have been unintentional since scanning for permissions looks to have been to determine which tables are writable by csv upload and tables that aren't even in the list, like the known limitation for foreign tables, are still included in the database schema sync for building simple questions on. 

This is a serious security limitation as it's the safest way to remove access for particular columns from even metabase users constructing native sql, and it is the only way to achieve this with postgres:

https://www.postgresql.org/docs/current/sql-grant.html
"Granting the privilege at the table level and then revoking it for one column will not do what one might wish: the table-level grant is unaffected by a column-level operation."

### To Reproduce

Grant access to only some columns of a table that doesn't have a full table grant, e.g. GRANT SELECT (id) ON foo and when metabase syncs the database schema it will be missing for simple questions.  

### Expected behavior

Tables with only a subset of columns available should be usable in simple questions.  

### Logs

_No response_

```JSON
Metabase versions 0.47 and up talking to a postgres database
```

### Severity

blocking users of tables with limited permissions

### Additional context

A fix is available in https://github.com/metabase/metabase/pull/40034

ericcj:
Fixes https://github.com/metabase/metabase/issues/40338

### Description

As part of CSV uploads introduced in 0.47 via https://github.com/metabase/metabase/pull/33415 metabase began excluding postgresql tables that have only a subset of their columns selectable from the schema for simple questions (you can still query them via native sql).  This appears to have been unintentional since scanning for permissions looks to have been to determine which tables are writable by csv upload and tables that aren't even in the list, like the known limitation for foreign tables, are still included in the database schema sync for building simple questions on.  It'd be even nicer if columns without select permission were automatically hidden so simple questions would work by default on them but that's beyond the scope of this bugfix.  

This is a serious security limitation as it's the safest way to remove access for particular columns from even metabase users constructing native sql, and it is the only way to achieve this with postgres:

https://www.postgresql.org/docs/current/sql-grant.html
"Granting the privilege at the table level and then revoking it for one column will not do what one might wish: the table-level grant is unaffected by a column-level operation."

The has_any_column_privilege function has been around since postgresql 8.4 so doesn't seem to introduce any new compatibility restriction.  

I believe they could be dug out of the column_privileges information_schema or pg_attribute.attacl ala https://dba.stackexchange.com/a/285632/113307 but that seems more complex and this was a smaller change.

column grants don't appear to be supported by the same code used in the redshift driver  https://docs.aws.amazon.com/redshift/latest/dg/r_System_information_functions.html

### How to verify

1. Instead of granting whole table permission grant it on a set of columns like the test does, e.g. GRANT SELECT (id) ON foo TO bar
2. See that the table is available for simple questions.  

I also ran the query current-user-table-privileges uses with this modification and saw that our tables that are this way now appear in it:

```
       with table_privileges as (
        select
          NULL as role,
          t.schemaname as schema,
          t.objectname as table,
          pg_catalog.has_any_column_privilege(current_user, '"' || t.schemaname || '"' || '.' || '"' || t.objectname || '"',  'update') as update,
          pg_catalog.has_any_column_privilege(current_user, '"' || t.schemaname || '"' || '.' || '"' || t.objectname || '"',  'select') as select,
          pg_catalog.has_any_column_privilege(current_user, '"' || t.schemaname || '"' || '.' || '"' || t.objectname || '"',  'insert') as insert,
          pg_catalog.has_table_privilege(current_user, '"' || t.schemaname || '"' || '.' || '"' || t.objectname || '"',  'delete') as delete
        from (
          select schemaname, tablename as objectname from pg_catalog.pg_tables
          union
          select schemaname, viewname as objectname from pg_catalog.pg_views
          union
          select schemaname, matviewname as objectname from pg_catalog.pg_matviews
        ) t
        where t.schemaname !~ '^pg_'
          and t.schemaname <> 'information_schema'
          and pg_catalog.has_schema_privilege(current_user, t.schemaname, 'usage')
       )
       select t.*
       from table_privileges t
```

### Checklist


test/metabase/driver/postgres_test.clj
```diff
@@ -1292,7 +1292,7 @@
                                      "CREATE MATERIALIZED VIEW \"dotted.schema\".\"dotted.materialized_view\" AS SELECT 'hello world';"
                                      "DROP ROLE IF EXISTS privilege_rows_test_example_role;"
                                      "CREATE ROLE privilege_rows_test_example_role WITH LOGIN;"
-                                     "GRANT SELECT ON \"dotted.schema\".\"dotted.table\" TO privilege_rows_test_example_role;"
+                                     "GRANT SELECT (id) ON \"dotted.schema\".\"dotted.table\" TO privilege_rows_test_example_role;"
```
qnkhuat, line 1295: Instead of modifying this test, could you add a new table and a new column grant instead?
I'd also add some test for "update" and "insert" grant too.

qnkhuat (changes requested):
Hi Eric, Thank you for your contribution; while the fix works, I'd like to request adding more test cases. 

Detail in the comment, please check it out!

qnkhuat:
CI PR: https://github.com/metabase/metabase/pull/40351

qnkhuat:
@ericcj on testing. 
I'd only add 3 more cases:
- a table with select permission on an ID column
- same but with update permission
- same but with insert permission

I can write those tests; just lmk if you want me to.

On executing the test, since it's a driver test, you'll need to have set the env var `DRIVERS=postgres`. This is to make driver tests executed for Postgres.

```bash
DRIVERS=postgres clj -X:dev:drivers:drivers-dev:test :only metabase.driver.postgres-test/table-privileges-test
```

This will run only the `table-privileges-test` for Postgres.

It expects a local Postgres server running at localhost:5432.
Tips: you can use this [script](https://github.com/metabase/dev-scripts/blob/master/run-postgresql-13.sh) to get a postgres server running.

qnkhuat (approved):
CI is green and the fix works, I'll update the test after this is merged.

PR #40043, opened 2024-03-12
RTKQuery integration: API Keys

sloansparger:
see: https://www.notion.so/metabase/Picking-a-real-data-fetching-library-d38077b06b9d4b09ba15db09ad608be0

This PR ports our API Keys feature to use RTKQuery


frontend/src/metabase/api/api-keys.ts
```diff
@@ -0,0 +1,56 @@
...
+  CreateApiKeyResponse,
+  UpdateApiKeyInput,
+  UpdateApiKeyResponse,
+  RegenerateApiKeyResponse,
+} from "metabase-types/api/admin";
+
+import { Api } from "./api";
+import { providesList, API_KEY_TAG, API_KEY_LIST_TAG } from "./tags";
+
+const extendedApi = Api.injectEndpoints({
```
iethree, line 13: how about 

```suggestion
export const apiKeyApi = Api.injectEndpoints({
```

frontend/src/metabase/api/api-keys.ts
```diff
@@ -0,0 +1,56 @@
...
+});
+
+export const {
+  useListApiKeyQuery,
+  useCountApiKeyQuery,
+  useCreateApiKeyMutation,
+  useRegenerateApiKeyMutation,
+  useUpdateApiKeyMutation,
+  useDeleteApiKeyMutation,
+} = extendedApi;
```
iethree, line 56:
```suggestion
} = apiKeyApi;
```

frontend/src/metabase/api/tags.ts
```diff
@@ -14,5 +14,8 @@ export function providesList<
     : [listTag];
 }
 
-export const tagTypes = [];
+export const API_KEY_TAG = "ApiKey" as const;
+export const API_KEY_LIST_TAG = getListTag(API_KEY_TAG);
```
iethree, line 18: WDYT about defining these in the api-keys file instead?

sloansparger: This file was born out of a circular import issue 😢

sloansparger: I think it's ultimately safer to do this in the case we may have two api definitions try to invalidate some part of each others cache. Also since tags need to be provided to the createApi call, you can't get away from editing two files.

iethree (approved):
🚀

PR #40049, opened 2024-03-12
[docs] Explicit callouts about data download in embedding

alexyarosh:
Add an explicit callout to check the iframe URL for data download parameter, as we had multiple questions about this.

### DIFF
--- a/docs/embedding/static-embedding-parameters.md
+++ b/docs/embedding/static-embedding-parameters.md
@@ -165,11 +165,11 @@ You can preview appearance settings from your question or dashboard's [embedded 
 
 \* Available on [Pro](https://www.metabase.com/product/pro) and [Enterprise](https://www.metabase.com/product/enterprise) plans.
 
-### Download results
+### Allow people to download the results of an embedded question
 
 {% include plans-blockquote.html feature="Downloading results" %}
 
-You can enable or disable downloading results for embedded questions by setting the `hide_download_button` parameter in the embedding URL to `false` or `true`.
+By default, Metabase will include a **Download** button on embedded questions. You can remove the download button by setting `hide_download_button=true` in the embedding URL in the iframe's `src` attribute, see [customizing the appearance of static embeds](#customizing-the-appearance-of-static-embeds).
 
 If the download button is missing when you expected it to be available, check that the URL in the `src` attribute for your iframe has the parameter `hide_download_button=false`.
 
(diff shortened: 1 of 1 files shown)


dragonsahead (approved):

docs/embedding/static-embedding-parameters.md
```diff
@@ -165,6 +165,17 @@ You can preview appearance settings from your question or dashboard's [embedded
 
 \* Available on [Pro](https://www.metabase.com/product/pro) and [Enterprise](https://www.metabase.com/product/enterprise) plans.
 
+### Download results
```
jeff-bruemmer, line 168:
```suggestion
### Allow people to download the results of an embedded question
```

Slightly more specific, and frames it from the perspective of an embedder.

Otherwise maybe just:

> ### Download results of an embedded question.

docs/embedding/static-embedding-parameters.md
```diff
@@ -165,6 +165,17 @@ You can preview appearance settings from your question or dashboard's [embedded
...
+
+{% include plans-blockquote.html feature="Downloading results" %}
+
+You can enable or disable downloading results for embedded questions by setting the `hide_download_button` parameter in the embedding URL to `false` or `true`.
```
jeff-bruemmer, line 172: We should say that the default is to show the download button. 

```suggestion
By default, Metabase will include a **Download** button on embedded questions. You can remove the download button by setting `hide_download_button=true` in the embedding URL in the iframe's `src` attribute.
```
And can you link to the [appearance section](https://www.metabase.com/docs/latest/embedding/static-embedding#customizing-the-appearance-of-static-embeds)? 

image

docs/embedding/static-embedding-parameters.md
```diff
@@ -165,6 +165,17 @@ You can preview appearance settings from your question or dashboard's [embedded
...
+
+If the download button is missing when you expected it to be available, check that the URL in the `src` attribute for your iframe has the parameter `hide_download_button=false`.
```
jeff-bruemmer, line 174: Is this line necessary? Won't it also work if that parameter isn't in the URL?

jeff-bruemmer (approved):
Thanks, @alexyarosh!

docs/embedding/static-embedding-parameters.md
alexyarosh: This was the problem that several customers had: they said the download button was missing, and when we asked them to check the URL, turned out they had this parameter set to true.  I'm open to rephrasing, but i would like to keep the general message of "check the URL if the embedding not behaving as you expect" explicitly written

PR #40054, opened 2024-03-13
Make static field values select searchable

ranquild:
Small UX fix. Now in the filter modal you can actually "search the list" of static field values:

Screenshot 2024-03-13 at 09 13 43


kamilmielnik (approved):
:+1:

nemanjaglumac (approved):
Works like a charm.

PR #40055, opened 2024-03-13
Elevate selected filter field values

ranquild:
Product doc https://www.notion.so/metabase/Show-the-selected-filter-values-on-top-of-multi-select-list-6e3f2e1880dc40bc8c50170236ca0aaa

Elevate selected field values on initial render when rendering them as a list of checkboxes.

How to verify:
- New -> Question -> Products
- Add a filter Category -> Gizmo
- Click on the added filter
- See that the Gizmo is at the top of the list
- Try checking and unchecking options. Gizmo stays at the same position, other options also don't jump.

Screenshot 2024-03-13 at 13 37 01
Screenshot 2024-03-13 at 13 37 07


frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.unit.spec.tsx
```diff
@@ -242,6 +242,66 @@ describe("StringFilterValuePicker", () => {
...
+          ],
+        }),
+      });
+
+      rerender(
+        <StringFilterValuePicker
+          query={query}
+          stageIndex={stageIndex}
+          column={column}
+          values={["p", "c"]}
```
kamilmielnik, line 291: Let's also make sure that unchecking values doesn't move the items.
This can be a separate test case, or this one:

```suggestion
          values={["c"]}
```

plus `expect(checkboxes[0]).not.toBeChecked();`

frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.unit.spec.tsx
```diff
@@ -242,6 +242,66 @@ describe("StringFilterValuePicker", () => {
       expect(onChange).toHaveBeenCalledWith(["t", "p"]);
     });
 
+    it("should elevate selected field values on initial render", async () => {
```
ranquild, line 245: All tests here with custom remapping to handle the most advance case.

frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.unit.spec.tsx
```diff
@@ -57,7 +57,7 @@ async function setupStringPicker({
     setupFieldSearchValuesEndpoints(result.field_id, value, result.values);
   });
 
-  renderWithProviders(
+  const { rerender } = renderWithProviders(
```
kamilmielnik, line 60: Should this case be handled too?

https://github.com/metabase/metabase/assets/6830683/5aee86fa-e1a1-4baa-b8f2-ddb7445999a2

ranquild: Fixed, great catch!

frontend/src/metabase/querying/components/FilterValuePicker/FilterValuePicker.unit.spec.tsx
ranquild: Added a test

kamilmielnik (approved):

PR #40060, opened 2024-03-13
Fix redshift flake by not clean up tables on some tests

### LINKED ISSUE #40058: Proposal to make redshift testing more robust
Currently, each test run on Redshift is isolated to the schema level, meaning all tests will operate on a single schema.

During testing we create a lot of datasets (test-data, sample-dataset, Tupac-insights... etc.); this leads to a problem in that these databases, when synced, will include all the tables in the current schema. 

E.g.: if you need a test that operates on a custom dataset ( defined by `mt/defdataset`), the database for this custom dataset, when synced, will contain all the tables that were created from other `deftest` (tables from test data, sample-dataset ... etc.). This is not ideal; each database for a custom dataset should only contain tables of that custom dataset.

Some ideas to tackle this:
1. Increase the isolation level to DB: make each test run(aka CI job) operate a newly created database. Each custom dataset should be created with a new schema. 
  - One downside of this is that we can only create 100 DBs per cluster, so it might not be enough if we have too many parallel runs.
2. Modify the sync code during the test so that when it syncs, it only includes tables from the current dataset.

qnkhuat:
Found these flakes in : 
- https://github.com/metabase/metabase/actions/runs/8236592940/job/22523412407#step:3:1096
- https://github.com/metabase/metabase/actions/runs/8235733448/job/22522618837#step:3:952

This is a temporary fix, we have a proposal for a more robust fix: https://github.com/metabase/metabase/issues/40058

### DIFF
--- a/enterprise/backend/test/metabase_enterprise/advanced_permissions/driver/impersonation_test.clj
+++ b/enterprise/backend/test/metabase_enterprise/advanced_permissions/driver/impersonation_test.clj
@@ -130,11 +130,13 @@
                                           mt/process-query
                                           mt/rows)))))
             (finally
-              (doseq [statement [(format "DROP TABLE IF EXISTS %s.table_with_access;" schema)
-                                 (format "DROP TABLE IF EXISTS %s.table_without_access;" schema)
-                                 (format "DROP SCHEMA IF EXISTS %s;" schema)
-                                 (format "DROP USER IF EXISTS %s;" user)]]
-                (jdbc/execute! spec [statement])))))))))
+             ;; comment out since it's causing flake
+             ;; can uncomment after we tackle #40058
+             #_(doseq [statement [(format "DROP TABLE IF EXISTS %s.table_with_access;" schema)
+                                  (format "DROP TABLE IF EXISTS %s.table_without_access;" schema)
+                                  (format "DROP SCHEMA IF EXISTS %s;" schema)
+                                  (format "DROP USER IF EXISTS %s;" user)]]
+                 (jdbc/execute! spec [statement])))))))))
 
 (deftest conn-impersonation-test-snowflake
   (mt/test-driver :snowflake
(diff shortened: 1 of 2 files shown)


tsmacdonald:
I don't think I understand how this works. The error I'm seeing in the test run is

```
clojure.lang.ExceptionInfo: Failed to create :redshift 'VXFVIFQTUITWKYLCDFUQ' test database: Failed to sync test database "VXFVIFQTUITWKYLCDFUQ": Error in sync step Sync redshift Database 373 ''VXFVIFQTUITWKYLCDFUQ'': Error in sync step Sync metadata for redshift Database 373 ''VXFVIFQTUITWKYLCDFUQ'': ERROR: relation "2024_03_11_486b968a_fce2_4d72_a16d_70ece5397dc0_schema.mv_tbl__test_tp_materialized_view__0" does not exist
    database-name: "VXFVIFQTUITWKYLCDFUQ"
           driver: :redshift
```

How would the `DROP TABLE IF EXISTS` affect that? It seems strictly helpful to clean up unneeded tables if they exist, but obviously I'm missing something

tsmacdonald:
meta-comment: there are a ton of whitespace changes in this PR which I think are not right

calherries (approved):

calherries:
> How would the DROP TABLE IF EXISTS affect that? It seems strictly helpful to clean up unneeded tables if they exist, but obviously I'm missing something

This was discussed on [this thread](https://metaboat.slack.com/archives/C064TCGEAV6/p1710323277228199?thread_ts=1710315562.594649&cid=C064TCGEAV6). But to summarize,  my leading hypothesis for why these flakes happen is the table is dropped somewhere between the start of a sync (in another test) and the end. If that happens the table is picked up during the `sync-tables` step, and then in a sync step following that, the table is missing.

Normally missing tables wouldn't be an issue, we would just log the error and move on. But in tests, we're intentionally throwing exceptions to guarantee that the sync works, see this https://github.com/metabase/metabase/blob/f52b46692d967a44da6264d7e9e2d35311fd05ca/test/metabase/test/data/impl/get_or_create.clj#L121

tsmacdonald:
Cool, thanks for the explanation! Makes sense

PR #40079, opened 2024-03-13
Fix `@cypress/grep` types by upgrading the library to its latest version

nemanjaglumac:
Prior to this PR if you wanted to create a TS Cypress spec that was using `tags`, it would've failed. Adding its types to our E2E TS config didn't help either. We were one major version behind.

Once I updated the library locally, TS config propagated its types and the TS was happy.
As a good test, I converted the existing JS test to TS. It should pass in CI, and `type-check` should not complain.


kamilmielnik (approved):
:+1:

PR #40082, opened 2024-03-13
[FE] "Combine columns" drill thru

### LINKED ISSUE #39980: [FE] "Combine columns" drill thru
Depends on #39978.

kamilmielnik:
Closes #39980

### Description
Adds combine column drill through to the chill mode header columns.

### How to verify

- Open the `Accounts` table in chill mode
- Click the header of a string column (eg. `Email`, `First Name` or `Plan`)
- Click `+ Combine columns`
- Play around with the column editor
- Click `Done`
- Verify the new column is added as the last column

### Demo

See video attached to [this Slack message](https://metaboat.slack.com/archives/C0645JP1W81/p1710427151050319)

Screenshot 2024-04-19 at 14 52 22
Screenshot 2024-04-19 at 14 52 26
Screenshot 2024-04-19 at 14 52 34


frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/ColumnAndSeparatorRow/ColumnAndSeparatorRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      <Select
+        className={styles.column}
+        data={options}
+        label={showLabels ? t`Column` : undefined}
+        styles={{
+          wrapper: {
+            "&:not(:only-child)": {
+              marginTop: 0,
+            },
```
kamilmielnik, line 57: We have `marginTop` defined for most Mantine inputs by default.
I think it's because of the assumption that if the input is `:not(:only-child)` there will be a label above it.
This assumption does not hold for `Select` because it never is the only child due to `<input type="hidden"` being rendered before it (see [screenshot](https://github.com/metabase/metabase/assets/6830683/ba484c9a-99fe-400b-9823-d20984b195fd)).

It'd be better to assume that every label that is `:not(:only-child)` will have an input underneath, and add `marginBottom` to labels instead. Though it'd be ideal if the low-level building blocks never made such assumptions at all.

Anyway, fixing that is out of the scope of this PR, so I added this band-aid here.

frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/ColumnAndSeparatorRow/ColumnAndSeparatorRow.module.css
```diff
@@ -0,0 +1,11 @@
+.separator {
```
kamilmielnik, line 1: It's not clear to me which styling should go to CSS modules, and which should be overridden with `styles` prop in Mantine components. Suggestions are appreciated.

frontend/src/metabase-lib/metadata.ts
```diff
@@ -145,7 +145,29 @@ declare function DisplayInfoFn(
...
+    return {
+      type: "drill-thru/combine-columns",
+      defaultSeparator: "/",
+      column: createdAtColumn!,
+      availableColumns: columns.slice(0, 5),
+    };
+  }
+
+  return result;
+};
```
kamilmielnik, line 170: Mocks compensating for the lack of BE implementation.
Don't bother reviewing it - this file should be reverted before the PR is merged.

frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/CombineColumnsDrill/CombineColumnsDrill.tsx
```diff
@@ -0,0 +1,127 @@
...
+                <Button p={0} variant="subtle" onClick={handleEditSeparators}>
+                  {jt`Separated by ${formatSeparator(defaultSeparator)}`}
+                </Button>
+              </Box>
+            )}
+          </Stack>
+
+          <Flex align="center" gap="md" justify="space-between">
+            <Button
+              leftIcon={<Icon name="add" />}
```
kamilmielnik, line 111: Mantine does not correctly handle it in RTL languages, i.e. the icon will be on the left in RTL layout as well.

frontend/src/metabase-lib/expression.ts
```diff
@@ -99,3 +100,15 @@ export function diagnoseExpression(
...
+export function previewExpression(
+  _query: Query,
+  _stageIndex: number,
+  _expressionClause: ExpressionClause,
+  queryResults: RowValues[],
+): RowValue[] {
+  // TODO: remove this mock
+  return queryResults.map(
+    () => `https://www.example.com/${Math.floor(1000 * Math.random())}`,
+  );
```
kamilmielnik, line 113: Mocks compensating for the lack of BE implementation.

kamilmielnik:
Dismissing request for review for now due to uncertainty on how we want to shape the UI here.

frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/ColumnAndSeparatorRow/ColumnAndSeparatorRow.module.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.remove {
+  flex: 0 0 auto;
+  border-color: transparent !important;
+}
+
+.wrapper {
+  &:not(:only-child) {
+    margin-top: 0;
```
kamilmielnik, line 16: We have `marginTop` defined for most Mantine inputs by default.
I think it's because of the assumption that if the input is `:not(:only-child)` there will be a label above it.
This assumption does not hold for `Select` because it never is the only child due to `<input type="hidden"` being rendered before it (see [screenshot](https://github.com/metabase/metabase/assets/6830683/ba484c9a-99fe-400b-9823-d20984b195fd)).

It'd be better to assume that every label that is `:not(:only-child)` will have an input underneath, and add `marginBottom` to labels instead. Though it'd be ideal if the low-level building blocks never made such assumptions at all.

Anyway, fixing that is out of the scope of this PR, so I added this band-aid here.

frontend/src/metabase/querying/utils/drills/combine-columns-drill/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+  const defaultColumn = columns[0];
+  const nextUnusedOption = options.find(option => {
+    return columnsAndSeparators.every(({ column }) => column !== option.column);
+  });
+  const column = nextUnusedOption ? nextUnusedOption.column : defaultColumn;
+  return { column, separator };
+};
+
+export const formatSeparator = (separator: string) => {
+  if (separator) {
```
romeovs, line 55: This will cause `formatSeparator` to return `(empty)` in most cases. 

You probably want:
```suggestion
  if (!separator) {
```

frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/CombineColumnsDrill/CombineColumnsDrill.tsx
```diff
@@ -0,0 +1,185 @@
...
+              >
+                {t`Add column`}
+              </Button>
+            </Flex>
+          </Stack>
+
+          <Example example={example} />
+
+          <Flex align="center" gap="md" justify="end">
+            <Button type="submit" variant="filled">
```
romeovs, line 177: We should probably disable the button when the form is in an invalid state.

frontend/src/metabase/querying/utils/drills/combine-columns-drill/utils.ts
kamilmielnik: Fixed on Friday, forgot to push: 9bd5d51e3b6f0e4ed26dd5467a306da3984faa2f :)

frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/CombineColumnsDrill/CombineColumnsDrill.tsx
kamilmielnik: I don't think this form can be in an invalid state. Am I missing something?

romeovs: I guess it's a decision we can make, but when a column dropdown I think the empty the form is in an invalid state. The alternative is that we just ignore that column in the resulting expression.

kamilmielnik: In this branch column dropdown is never empty.
When adding a new "row", a column is automatically selected, see `getNextColumnAndSeparator` and how it's used.

frontend/src/metabase/querying/utils/drills/combine-columns-drill/components/ColumnAndSeparatorRow/ColumnAndSeparatorRow.module.css
romeovs: Fixed here: https://github.com/metabase/metabase/pull/41728

nemanjaglumac (commented):
I'm seeing some weirdness with the column position when switching between the visualization and the raw table views.
https://www.loom.com/share/a7cfa7b8b09d4314955ec03dfcfdd786?sid=071f3a6b-86ea-46e7-9cfa-82ffe5906cea

romeovs:
> I'm seeing some weirdness with the column position when switching between the visualization and the raw table views. https://www.loom.com/share/a7cfa7b8b09d4314955ec03dfcfdd786?sid=071f3a6b-86ea-46e7-9cfa-82ffe5906cea

Is that not was is expected?

The way I understand it, the table view is showing the "raw" table and thus showing the column at the end, whereas the visualization as a table is showing the "customized" table.

You can do the same with any column (even non-generated ones) in the table and it would yield the same issue.

nemanjaglumac (approved):

PR #40087, opened 2024-03-13
Disable saving models without up-to-date result_metadata

### LINKED ISSUE #37009: Updated native models may lose their result set metadata

ranquild:
Fixes https://github.com/metabase/metabase/issues/37009

Disables creating new and updating existing models without `result_metadata`. That means that if the query was changed, the user would have to run the query first and wait for results.

How to verify:
- New -> Modal -> Native
- SELECT * FROM PRODUCTS
- Try to save the model without running the query. The button should be disabled and there should be a tooltip.
- Run the query
- Now the save button should become enabled
- Save the model
- ... menu -> Edit query definition
- Try changing the model query. The button should become disabled
- Run the updated query. The button should become enabled

Screenshot 2024-03-15 at 16 23 04


frontend/src/metabase/entities/tables.js
```diff
@@ -82,13 +81,6 @@ const Tables = createEntity({
     // loads `query_metadata` for a single table
     fetchMetadata: compose(
       withAction(FETCH_METADATA),
-      withForceReload((state, { id }) => {
```
ranquild, line 85: This hack is no longer needed

luizarakaki (approved):
lgtm

frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx
```diff
@@ -464,20 +489,17 @@ function DatasetEditor(props) {
...
-            activeText={t`Saving…`}
-            failedText={t`Save failed`}
-            successText={t`Saved`}
-            className={cx(
-              ButtonsS.Button,
-              ButtonsS.ButtonPrimary,
-              ButtonsS.ButtonSmall,
-            )}
-          />,
+          saveButtonTooltipLabel ? (
```
kamilmielnik, line 492: Instead of conditionally rendering the `Tooltip` we could pass `disabled={!saveButtonTooltipLabel}` prop to it.
This way we could inline the `ActionButton` back here and remove `key` prop from it.

frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx
```diff
@@ -413,18 +418,21 @@ function DatasetEditor(props) {
...
-    return everyFieldHasDisplayName && isDirty;
-  }, [dataset, fields, isDirty]);
+  const canSaveChanges =
+    !isEmpty &&
+    isDirty &&
+    (!isNative || !isResultDirty) &&
+    fields.every(field => field.display_name);
+
+  const saveButtonTooltipLabel =
+    !isEmpty && isDirty && isNative && isResultDirty
```
kamilmielnik, line 433: Why only native models?

frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx
ranquild: Didn't about about `disabled`, thanks!

frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx
ranquild: Because for MBQL queries `result_metadata` can be computed by the BE without running the query

kamilmielnik (approved):

frontend/src/metabase/query_builder/components/DatasetEditor/DatasetEditor.jsx
```diff
@@ -464,20 +472,27 @@ function DatasetEditor(props) {
...
-              ButtonsS.ButtonPrimary,
-              ButtonsS.ButtonSmall,
-            )}
-          />,
+            refProp="innerRef"
+            label={saveButtonTooltipLabel}
+            disabled={!saveButtonTooltipLabel}
+          >
+            <ActionButton
+              key="save"
```
kamilmielnik, line 482: Redundant

PR #40101, opened 2024-03-13
Fix multi-schema data selector header styling

### LINKED ISSUE #39999: Inconsistent spacing around table and schema divider
Due to the way this component is structured, we have some inconsistent spacing in the data selector between the table name and the schema name.

Here it is zoomed in.
image

Notice the bigger gap on the left side of the `-`.
The text color is also different but this might have been intentional. I'll check with the design team.

nemanjaglumac:
- Makes spacing equal on both sides of the divider
- Changes the divider from the hyphen to the forward slash

### Before
image

### After
image

Fixes #39999


mazameli (approved):
Nit/question: in your screenshots the words "Multi-schema" and "Schema two" look slightly like different colors. Are my eyes deceiving me?

But as for the actual issue being addressed in this PR: I love your attention to detail, and I'm cool with the change here.

nemanjaglumac:
> Are my eyes deceiving me?

They are not.
Chrome vs Firefox screenshots :)

PR #40103, opened 2024-03-13
Global to local css migration - Buttons

deniskaber:

### Description

This migrates global CSS rules related to buttons to be locally scoped.

### How to verify


oisincoveney (approved):
I _think_ it should be good except this one button, which is a bit small but I'm not sure if this PR changed that (but it is a Button)

image

oisincoveney (approved):

Issue #39608, opened 2024-03-05
[dc.js migration] chart errors when number of insight objects does not match number of series


EmmadUsmani:
question/5331
https://metaboat.slack.com/archives/C06G94JTWBS/p1706198465633859

<img width="684" alt="Screenshot 2024-03-04 at 4 49 17 PM" src="https://github.com/metabase/metabase/assets/37751258/7e90240b-7494-44db-909f-c1167718b7bd">

question/9318
https://metaboat.slack.com/archives/C06G94JTWBS/p1706199292996149

PR #40104, opened 2024-03-13
fix insights error

EmmadUsmani:
Closes https://github.com/metabase/metabase/issues/39608

### Description

The error happens when there is a chart where not all columns are visualized, for example the chart has 4 columns, 1 is a dimension, 2 are metrics, but one is not rendered. In this case the insights array will still contain the object for that column, which causes an error to be thrown in `trend-line.ts` because we only have 2 `seriesModels`, but 3 insight objects.

The solution is to filter out insight objects that are not being rendered as a series.

### Checklist


EmmadUsmani:
* **#40104** <a href="https://app.graphite.dev/github/pr/metabase/metabase/40104?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `echarts`

<h2></h2>

 Join @EmmadUsmani and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

alxnddr (approved):
The fix looks good 👍 

I'm thinking that probably the proper way for trend lines is in the chart model. There can be `trendLinesDataset: ChartDataset` and `trendSeries: SeriesModel`. It would be better from the performance perspective as well since we would not recompute trend lines data when resizing

EmmadUsmani:
@alxnddr That makes sense. Can you file an issue for this and put it in the code review feedback portion of the regressions epic?

PR #40112, opened 2024-03-13
Remove global css module that was applying red borders

rafpaf:
This PR
* removes global.module.css, which was applying a red border to anything with class "undefined"
* adds a simple hook that checks the CSS for aberrations every 15s. Alerts are sent via `console.error`

Here's how the red border previously looked:
image

Screenshot from `/reference/segments`


deniskaber (changes requested):
Hey @rafpaf ! We should find why `undefined` is passed as classname to those components rather than removing this css rule. It was added to track such incorrect class names usages.

iethree:
🤔  This feels like a strange approach to me. I can see how this red box would be helpful in development, but it looks like we're shipping debug code in production here.

rafpaf:
I'm still concerned about this. I only know part of the story, so let me share that part - how things look from my point of view. The red line might make sense *if we will easily notice it*. On the screenshotted page, though, it appeared on items that are tucked away on the "Learn about our data" page. That is a somewhat high-visibility page (you can get to it by clicking two high-visibility links in a row). But it's not a page that we frequent a lot, and we don't have any visual tests that cover it. To my mind, this indicates that the red border strategy is quite risky. It risks creating an embarrassing visual error that we might not notice until we hear about it from users.

I am sure that I don't understand the whole context though. What am I missing?

rafpaf:
A middle way might be to switch this on just in development environments and in stats, if this is possible.

dpsutton:
@deniskaber would a good solution being coming up with a way to enable the red CSS in stats and local dev, but suppressing it in the regular course of business in production builds?

I really like stats to highlight these as it's heavily used. Highlighting these CSS cases in the regular production jar feels far worse than an undefined class to me.

Does this seem like a good way to track CSS issues while not drawing unwarranted attention to perhaps minor mistakes in customer builds?

rafpaf:
@deniskaber @dpsutton I've just pushed an approach that I think gets most of the value here cheaply: every 15s the app checks for problematic elements and `console.error`s if any are found. I think that's probably sufficient visibility. Wdyt?

deniskaber (commented):
Hey @rafpaf ! 
I don't think we need a hook to do this checks.

Let's just remove `global.module.css` and that's it.

rafpaf:
@deniskaber Done! The PR now just removes `global.module.css`

deniskaber (approved):
Nice! Thanks, @rafpaf !

PR #40113, opened 2024-03-13
[WIP] Collection perm graph reduce payload

escherize:


escherize:
This isn't being worked on currently, let's reopen when it is.

PR #40114, opened 2024-03-13
Funnel Chart Static-Viz should handle row data more robustly

### LINKED ISSUE #39743: funnel chart does not work in static viz
### Describe the bug

When adding this specific funnel chart question to a dashboard and sending a subscription email, there is an error in the subscription.

https://github.com/metabase/metabase/assets/37751258/28a95277-df22-46de-9239-fee79d925953

### To Reproduce

[Simple User Flow Funnel.csv](https://github.com/metabase/metabase/files/14517331/Simple.User.Flow.Funnel.csv)

Create the following question using the above csv data

Screenshot 2024-03-06 at 5 04 41 PM

Visualize as a bar chart, add to a dashboard, send a test subscription email

### Expected behavior

The chart should appear in the email instead of an error

### Logs

[backend] java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number (java.lang.String and java.lang.Number are in module java.base of loader 'bootstrap')
[backend]       at clojure.lang.Numbers.dec(Numbers.java:147)
[backend]       at metabase.pulse.render.body$eval92227$_AMPERSAND_f__92229$fn__92239.invoke(body.clj:1042)

```JSON
master commit 9413706bd2

{
  "browser-info": {
    "language": "en-US",
    "platform": "MacIntel",
    "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36",
    "vendor": "Google Inc."
  },
  "system-info": {
    "file.encoding": "UTF-8",
    "java.runtime.version": "17.0.6+10",
    "java.version": "17.0.6",
    "java.vm.version": "17.0.6+10",
    "os.name": "Mac OS X",
    "os.version": "13.6.2",
    "user.language": "en",
    "user.timezone": "America/Los_Angeles"
  },
  "metabase-info": {
    "databases": [
      "h2",
      "postgres"
    ],
    "hosting-env": "unknown",
    "application-database": "h2",
      "database": {
        "name": "H2",
      },
      "jdbc-driver": {
        "name": "H2 JDBC Driver",
      }
    },
    "run-mode": "dev",
    "version": {
      "tag": "vLOCAL_DEV",
      "hash": "06d1ba2ae111e66253209c01c244d6379acfc6dcb1911fa9ab6012cec9ce52e5"
    },
    "settings": {
      "report-timezone": null
    }
  }
}
```

### Severity

Moderately severe, viz type that should work in static viz does not

### Additional context

_No response_

adam-james-v:
Fixes: #39743 

The row data coming into the funnel chart render method can evidently have a few slightly different shapes, and may need to be 'coordinated' with the :funnel.rows key from viz-settings (names and order are specified there).

The specific bug is that we previously assumed the rows always had indices: `[idx val]`, but they can actually be strings too, so (dec idx) would fail.

This PR makes the method more robust to different row data shapes. For example:

`[[1 100] [2 200]]` should work as well as `[["A" 100] ["B" 200]]` and correctly handle the viz-settings, when possible. There may be cases where the :funnel.rows key has non-unique keys. In such a case, the row order can't be determined, so we hope that the query's results are returned in order. I suspect this case is rare in practice, so this feels like an acceptable solution for now.

I also noticed that the funnel.rows may have `:enabled false` and took the opportunity to filter those rows out as well.


test/metabase/pulse/render/body_test.clj
```diff
@@ -621,6 +622,50 @@
...
+   ["product_page" 5000]
+   ["purchase" 225]])
+
+(tx/defdataset funnel-data
+  [["stages"
+    [{:field-name "stage", :base-type :type/Text}
+     {:field-name "count", :base-type :type/Quantity}]
+    funnel-rows]])
+
+(deftest render-funnel-text-row-labels-test-2
```
tsmacdonald, line 638: needs a `^:synchronized`, no?

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
...
 
+(defn- all-unique?
+  [funnel-rows]
+  (let [ks (into #{} (map :key funnel-rows))]
+    (= (count ks) (count funnel-rows))))
+
+(defn- build-funnel-rows
+  [funnel-rows raw-rows]
+  (let [funnel-rows (map-indexed (fn [idx row] (assoc row :idx (inc idx))) funnel-rows)
+        xf-rows     (into {} (map (fn [row] (vec (take-last 2 row))) raw-rows))]
```
tsmacdonald, line 1041: this is probably a silly question, but under what circumstances would the row have more than two elements? I kinda thought in this context a row would always be an `[x, y]` pair?

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
...
              :src   (:image-src image-bundle)}]]}))
 
+(defn- all-unique?
+  [funnel-rows]
+  (let [ks (into #{} (map :key funnel-rows))]
+    (= (count ks) (count funnel-rows))))
+
+(defn- build-funnel-rows
+  [funnel-rows raw-rows]
+  (let [funnel-rows (map-indexed (fn [idx row] (assoc row :idx (inc idx))) funnel-rows)
```
tsmacdonald, line 1040: I usually don't like shadowing bindings (and would have called this `indexed-funnel-rows` myself) but I guess it's not a huge deal in this particular fn

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
...
 (mu/defmethod render :funnel :- formatter/RenderedPulseCard
   [_chart-type render-type _timezone-id card _dashcard {:keys [rows cols viz-settings] :as data}]
   (let [[x-axis-rowfn
          y-axis-rowfn] (formatter/graphing-column-row-fns card data)
         funnel-rows    (:funnel.rows viz-settings)
         raw-rows       (map (juxt x-axis-rowfn y-axis-rowfn)
                             (formatter/row-preprocess x-axis-rowfn y-axis-rowfn rows))
-        rows           (cond->> raw-rows
-                         funnel-rows (mapv (fn [[idx val]]
-                                             [(get-in funnel-rows [(dec idx) :key]) val])))
```
tsmacdonald, line 1042: For my own sanity: the bug is that `idx` could be `"cart"` or whatever and you can't call `dec` on that

tsmacdonald (approved):
Makes sense to me

src/metabase/pulse/render/body.clj
crisptrutski: I used to belong to the "never shadow" camp. I had my mind changed after working in Rust for a while though, where it's fairly idiomatic to use shadowing when transforming values. The language has a pretty strict linter, and the linter for shadowing is that it's only an error if the expression for the new binding doesn't use the old binding.

This is partly a stylistic thing, to make functional interfaces look more imperative (😫), but also practical as it moves the value (often off limits through the borrow checker) out of scope. When the conceptual entity is the same, it's nice to avoid a proliferation of similar or awkward names, and rules out the failure mode where you can pick the wrong one.

That said, `indexed-funnel-rows` is a very natural name, so I can go either way on this case. Mostly mentioning because we have a fair amount of code that uses this pattern of shadowing, and I don't think we need a holy war against it 😄

src/metabase/pulse/render/body.clj
crisptrutski: Minor note, but it's idiomatic when using `into` and `map` to move the brackets a bit and use the "transducing". This tiny adjustment avoids constructing a lazy sequence just to immediately iterate over and discard it.

```suggestion
        xf-rows     (into {} (map (fn [row] (vec (take-last 2 row)))) raw-rows)]
```

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
       [:img {:style (style/style {:display :block :width :100%})
              :src   (:image-src image-bundle)}]]}))
 
+(defn- all-unique?
+  [funnel-rows]
+  (let [ks (into #{} (map :key funnel-rows))]
```
crisptrutski, line 1035: Another classic case of a simple transducer switch.

```suggestion
  (let [ks (into #{} (map :key) funnel-rows)]
```

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
...
+    (= (count ks) (count funnel-rows))))
+
+(defn- build-funnel-rows
+  [funnel-rows raw-rows]
+  (let [funnel-rows (map-indexed (fn [idx row] (assoc row :idx (inc idx))) funnel-rows)
+        xf-rows     (into {} (map (fn [row] (vec (take-last 2 row))) raw-rows))]
+    (for [{idx     :idx
+           k       :key
+           enabled :enabled} funnel-rows]
+      (when enabled
```
crisptrutski, line 1045: The PR description doesn't mention this condition, and it wasn't present in the old code. Can you update the PR to mention and explain this change? 

Also, do you mean to leave `nil` values in this list when the row is not enabled? There's a very neat way to filter them out, which is to replace:

```clojure
(for [{..., enabled :enabled} funnel-rows]
  (when enabled
    ...))
```

with just

```clojure
(for [{..., enabled? :enabled} funnel-rows
      :when enabled?]
  ...)

File: test/metabase/pulse/render/body_test.clj (hunk 1 above)
Comment (crisptrutski) [reply test/metabase/pulse/render/body_test.clj:638]: This is a question for Tim, but I assume that `mt/dataset` is the non-threadsafe part? I see for h2 at least it'll clobber an existing table for the same dataset.

Since this test is the only one using that dataset, and thus the corresponding database, it looks like this can actually be `:parallel`? I guess that's brittle because it'd be easy for another test to use the same dataset, or at least a binding with the same name, in the same namespace (which determines the database name).

File: src/metabase/pulse/render/body.clj [hunk 7]
```diff
@@ -1030,16 +1030,31 @@
       [:img {:style (style/style {:display :block :width :100%})
              :src   (:image-src image-bundle)}]]}))
 
+(defn- all-unique?
+  [funnel-rows]
+  (let [ks (into #{} (map :key funnel-rows))]
+    (= (count ks) (count funnel-rows))))
+
+(defn- build-funnel-rows
```
crisptrutski, line 1038: A minor note here, but in Clojure the approach to putting verbs in function names is slightly different than in a lot of other languages.

Quoting from the [Elements of Clojure](https://leanpub.com/elementsofclojure/read_sample) on Naming Functions

> If a function only transforms data, we should avoid verbs wherever possible. A function that calculates an MD5 hash, defined in our payload namespace, should be called md5.

Essentially the recommendation is to reserve verbs for functions that do things like fetching, mutating, publishing, etc, and use "declarative" names otherwise.

Just sharing this because I'm trying to break the habit of sticking verbs on every function myself. In this case the verb is better than shadowing the binding of the same name though.

src/metabase/pulse/render/body.clj
crisptrutski: Following on from Tim's question, what are the semantic meaning of the key and the value in this map? I think that the name of the binding should reflect that. For example if the keys represent some kind of `foo` and the values some kind of `bar`, then I would name it `foo->bar`. 

The current name trips me up as `xf` is often short for an `xform`, a transducing function, and `rows` implies some kind of sequence rather than a mapping.

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
...
+  [funnel-rows raw-rows]
```
crisptrutski, line 1039: There are so many comments in this area, it's hard to comment on the function as a whole. This is a bit of a zoom out comment on the whole function.

I think breaking that readability could be improved by making it clearer that the transformed funnel-rows are only used for their indexing values. The best way to do this I think is to actually discard the other fields and stop calling them funnel-rows.

Also it seems strange to call the result "funnel-rows", as those come in as maps, and the output is a sequence of tuples whose shape is more akin to the `raw-rows`.

```clojure
(defn- reordered-rows [funnel-rows raw-rows]
  (let [idx+keys (map-indexed (fn [idx {k :key, enabled :enabled}]
                                (when enabled [idx k]))
                              funnel-rows)
        foo->bar (into {} (map (fn [row] (vec (take-last 2 row)))) raw-rows)]
    (for [[idx k] idx+keys :when (or idx k)]
      [k (or (get foo->bar k) (get foo->bar idx))])))
```

test/metabase/pulse/render/body_test.clj
crisptrutski, line 638: Naming nit, how about calling out the semantic distinction between the test (indexes versus keys), rather than having "test" and "test-2".

Otherwise, I think that test names should always end in `-test`, so any "suffix" should come before that.

crisptrutski (commented):
Hope the comments aren't too overwhelming! Mostly "by the way" tips I hope you find interesting, but aren't essential.

The key thing for me is improving clarity for future developers, and here are my chief call-outs:

1. The `build-funnel-rows` method is non-trivial and subtle, and took me a second to grok. I encourage you to try find a clearer name, or otherwise write a doc-string explaining it at a high level.
2. The "take 2" seems weird. Is that intended to filter out a primary key or something? If it's necessary I'd suggest adding a line comment to explain it for future maintainers.
3. Call out the new "enabled" filtering in the PR description, and either filter the nils or explain with a line comment why you're keeping them.

src/metabase/pulse/render/body.clj
```diff
@@ -1030,16 +1030,31 @@
...
+  (let [funnel-rows (map-indexed (fn [idx row] (assoc row :idx (inc idx))) funnel-rows)
+        xf-rows     (into {} (map (fn [row] (vec (take-last 2 row))) raw-rows))]
+    (for [{idx     :idx
+           k       :key
+           enabled :enabled} funnel-rows]
+      (when enabled
+        [k (or (get xf-rows k) (get xf-rows idx))]))))
```
crisptrutski, line 1046: This fallback feels a bit magical to me, especially since I don't imagine we'd want to support a mixture of keys and indexes in our raw rows.

Just food for thought - did you consider the approach where you pick a case at the start, based on whether the raw rows have any `string?` values?

PR #40133, opened 2024-03-14
Global to local css migration - form

deniskaber:

### Description

This migrates global CSS rules from `form.module.css` to be locally scoped.

### How to verify


frontend/src/metabase/components/Grabber/Grabber.jsx
```diff
@@ -1,10 +1,13 @@
...
 import cx from "classnames";
 
+import GrabberS from "metabase/css/components/grabber.module.css";
+import CS from "metabase/css/core/index.css";
+
 export default function Grabber({ className = "", style, ...props }) {
   return (
     <div
-      className={cx("Grabber cursor-grab", className)}
+      className={cx(GrabberS.Grabber, CS.cursorGrab, className)}
```
npretto, line 10: I think this is missing the test-id, which is making the e2e fail

oisincoveney (approved):
noice!

PR #40134, opened 2024-03-14
Convert `pulse.module.css` to CSS modules

oisincoveney:
### Description

Removes `pulse.module.css` from the global CSS and uses local styles to (hopefully) make the styling for the components more easily understood.

### How to verify

Go to the Question Alerts button (the bell button on the bottom right of the view) and create an alert. Ensure that there are no weird or obvious style failures when using the email and slack widgets.

### Checklist


frontend/src/metabase/query_builder/components/AlertModals/AlertModals.jsx
```diff
@@ -381,7 +383,12 @@ export class DeleteAlertSection extends Component {
     const { onDeleteAlert } = this.props;
 
     return (
-      <DangerZone className="DangerZone mt4 pt4 mb2 p3 rounded bordered relative">
+      <DangerZone
+        className={cx(
+          AlertModalsS.AlertModalsContainer,
```
WiNloSt, line 388: I'm not sure why do we need an extra class here as it's not here before.

oisincoveney: Looks like this component is only used here.

frontend/src/metabase/query_builder/components/AlertModals/AlertModals.module.css
```diff
@@ -0,0 +1,4 @@
+.AlertModalsBorder {
+  border-width: 2px;
+  border-color: var(--color-border);
+}
```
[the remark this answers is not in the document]
oisincoveney: I did a little look into the code and the HTML output and I don't see any components with `.input` anywhere within `.PulseEdit`

deniskaber (approved):

PR #40141, opened 2024-03-14
Improved pivot QP: use MLv2 and avoid recursive preprocessing (performance improvement)

camsaul:
Updates pivot QP to handle MLv2 queries (actually, it uses MLv2 now and converts legacy queries if it sees them). This simplified the code a lot. I added a lot of documentation and Malli schemas here while I was in the process of figuring out how this code worked, hopefully it will make it easier to work with in the future.

While I was working on this I realized that the recursive preprocessing and `:name`-based column matching to line up columns from various subqueries was unnecessary, expensive, and error-prone; I replaced it with something that should lead to a nice performance boost and probably fix a few bugs (I'll have to profile and look for matching bugs soon.)

Part of #39024


src/metabase/query_processor/pivot.clj
```diff
@@ -39,20 +53,34 @@
...
-(defn- breakout-combinations
+(mr/def ::breakout-combination
+  [:sequential ::index])
+
+(mr/def ::breakout-combinations
+  [:and
+   [:sequential ::breakout-combination]
+   [:fn
+    {:error/message "Distinct combinations"}
+    distinct?]])
```
snoe, line 72: I think this will always return true, because `distinct?` is dumb  (careful of `(apply distinct? [])` too).
https://clojuredocs.org/clojure.core/distinct_q

src/metabase/query_processor/pivot.clj
```diff
@@ -96,82 +124,116 @@
         ;; bottom right corner [_ _ _ _] => 1111 => Group #15
         [[]]))))))
 
-(defn- add-grouping-field
+(mu/defn ^:private keep-breakouts-at-indecies :- ::lib.schema/query
```
bshepherdson, line 127: typo: `indices`, throughout this function.

Or perhaps better, the also-accepted spelling of `indexes`.

src/metabase/query_processor/pivot.clj
```diff
@@ -96,82 +124,116 @@
...
+  [:sequential :any])
 
-(defn- row-mapping-fn
+(mu/defn ^:private row-mapping-fn :- [:=> [:cat ::row] ::row]
   "This function needs to be called for each row so that it can actually shape the row according to the
-  [[*column-mapping-fn*]] above."
-  [row]
+  `column-mapping-fn` we build at the beginning.
+
+  Row mapping function is a function that can reorder the row add `nil`s for columns that aren't present in a particular
```
bshepherdson, line 204: nit: "reorder the row and add `nil`s" I think?

src/metabase/query_processor/pivot.clj
```diff
@@ -1,35 +1,49 @@
 (ns metabase.query-processor.pivot
-  "Pivot table actions for the query processor"
+  "Pivot table query processor. Runs a bunch of subqueries separately and stitches results together. TODO -- this whole
+  thing is dumb, why don't we just generate a big ol' UNION query? -- Cam"
```
bshepherdson, line 3: I'm struggling to follow the below since I don't understand how we implement pivot queries.

A paragraph or two about how this works would go a long way to help out future readers.

src/metabase/query_processor/pivot.clj
```diff
@@ -305,6 +372,77 @@
...
+    ;;    {1 2, 3 0}
+    ;;
+    ;; next, let's use that map to make a vector of like
+    ;;
+    ;;    [nil 2 nil 0]
+    ;;
+    ;; e.g.
+    ;;
+    ;; * canonical column 0 has no corresponding column in the subquery
+    ;; * canonical column 2 corresponds to subquery column 1
```
bshepherdson, line 417: Beautiful!

bshepherdson (approved):

PR #40142, opened 2024-03-14
MLv2 normalization overhaul

camsaul:
MLv2 normalization was previously pretty incomplete, this PR overhauls it so it actually works 100% when things come out of the app DB or in thru the REST API; adds some QP preprocessing middleware to verify this as well.

The new normalization code uses Malli coercion using the `:decode/normalize` key attached to various MLv2 schemas.

I've also added `:encode/serialize` for preparing a MLv2 query for serialization to the app DB or over the REST API, this is currently mostly to help the new normalization tests avoid failing (shouldn't need to normalize keys that are added by QP entrypoints or middleware, like `:info`).

The MLv2 schema was also missing a lot of stuff like `:info`, `:parameters`, and `:template-tags`, so this adds them as well.

I'll call more stuff out in the comments


.clj-kondo/hooks/clojure/test.clj
```diff
@@ -28,7 +28,6 @@
      metabase.test.util.log/with-log-level
      metabase.test.util.log/with-log-messages-for-level
      metabase.test.util.misc/with-single-admin-user
-     metabase.test.util.timezone/with-system-timezone-id
```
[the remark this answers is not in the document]
camsaul: All these functions should be renamed to end in `!` at some point honestly.

src/metabase/actions.clj
```diff
@@ -147,12 +144,15 @@
...
   (check-actions-enabled-for-database! (api/check-404 (database-for-action action-or-id))))
 
-(defn perform-action!
+(mu/defn perform-action!
   "Perform an `action`. Invoke this function for performing actions, e.g. in API endpoints;
   implement [[perform-action!*]] to add support for a new driver/action combo. The shape of `arg-map` depends on the
   `action` being performed. [[action-arg-map-spec]] returns the specific spec used to validate `arg-map` for a given
   `action`."
-  [action arg-map]
+  [action
```
camsaul, line 152: As chronicled [here](https://metaboat.slack.com/archives/C04DN5VRQM6/p1710449728575199), the Actions stuff used to accept either keywords or strings for column names. That made normalization hard to test, since it needs to decide between one or the other. Keyword column names are icky anyway, since column names can be blank, or have spaces, or slashes, so I updated all the Actions code to use string names and only string names.

src/metabase/actions.clj
```diff
@@ -199,25 +199,6 @@
    :actions.args/common
    (s/keys :req-un [:actions.args.crud.row.common/query])))
 
-;;; the various `:row/*` Actions all treat their args map as an MBQL query.
-
-(defn- normalize-as-mbql-query
```
camsaul, line 204: I also removed the custom actions normalization code and rolled it in to the normal MBQL normalization code

src/metabase/api/common.clj
```diff
@@ -123,7 +123,7 @@
 
     (check test1 code1 message1
            test2 code2 message2)"
-  {:style/indent 1, :arglists '([condition [code message] & more] [condition code message & more])}
+  {:style/indent [:form], :arglists '([condition [code message] & more] [condition code message & more])}
```
camsaul, line 126: This is a regular function so should get regular function indentation instead of fancy `let`-style indentation, that's just the rules.

src/metabase/api/dashboard.clj
```diff
@@ -918,7 +919,7 @@
     (get-in card [:dataset_query :native :template-tags (u/qualified-name tag)])))
 
 (defn- param-type->op [type]
-  (if (get-in mbql.s/parameter-types [type :operator])
+  (if (get-in lib.schema.parameter/types [type :operator])
```
camsaul, line 922: Originally moved the schemas for parameters into MLv2 so we can add normalization logic as needed. Didn't end up doing this in the end, but previously the MLv2 schema didn't check `:parameters` at all, so this is still a net win

test/metabase/api/dashboard_test.clj
```diff
@@ -3998,7 +4000,7 @@
 ;;; +----------------------------------------------------------------------------------------------------------------+
 
 (deftest param-values-no-field-ids-test
-  (testing "Ensure param value lookup works for values where field ids are not provided, but field refs are."
+  (testing "Ensure param value looktup works for values where field ids are not provided, but field refs are."
```
metamben, line 4003: Typo.

metamben (approved):
Great! ❇️ And difficult to review. Having some sweeping changes like renaming the thread unsafe functions in separate PRs would have helped. (This PR is so big that GH gave up and didn't load even the small diffs.)

src/metabase/lib/normalize.cljc
```diff
@@ -1,133 +1,91 @@
...
+   :metadata/metric   ::lib.schema.metadata/metric})
+
+(defn- infer-schema [x]
+  (cond
+    (map? x)
+    (or (-> x lib-type lib-type->schema)
+        :map)
+
+    (and (vector? x)
+         ((some-fn simple-keyword? string?) (first x)))
```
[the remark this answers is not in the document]
camsaul: It's actually a super simple change to have the normalize code add options maps automatically and add `:lib/uuid`s automatically if we wanted. I'm not sure if we want to introduce that or not, but I thought for now leaving the door open to it wasn't a bad idea

PR #40143, opened 2024-03-14
Backport fix to check base-type in isa if effective-type is missing

### LINKED ISSUE #35640: Discrepancy between versions on what cards can be connected to others in multi-series charts
### Describe the bug

A customer reported that charts that were created on older versions couldn't be connected with charts on new versions. I did some testing and there are some things that we should check out

Repro:
(note: use numbers 1-8 for the chart names)
1) start v44.7 
2) create the following questions:
- (orders) sum of total and sum of subtotal per created_at-month
image
click visualize and save
- then change the visualization to bar chart and save with a new name
- create a model of the orders table
- then create the same 2 questions now based on the model

3) add these 4 visualizations to a dashboard

Now move to v47.7
4) create again the same 4 questions and add these to the same dashboard

Alright, so now we'll have: 
- 4 questions created in 44 and 4 questions created in 47 of which,
- 4 line charts and 4 bar charts, 2 of each created from a model

All these charts should be able to connect to each other, but no. In the new version, all charts can be connected to the new charts, but no: try on each chart and see which charts are allowed to be connected)
image

### To Reproduce

Above

### Expected behavior

if charts are compatible, they should be able to be added

### Logs

NA

```JSON
v47.x
```

### Severity

P2

### Additional context

_No response_

snoe:
Fixes #35640

Pulls snippet from a66db9e here https://github.com/metabase/metabase/blame/ed7e147b7cdace92b6f84e35523558927d1f5821/src/metabase/lib/types/isa.cljc#L24-L27

### DIFF
--- a/src/metabase/lib/types/isa.cljc
+++ b/src/metabase/lib/types/isa.cljc
@@ -20,7 +20,11 @@
   "Returns if `column` is of category `category`.
   The possible categories are the keys in [[metabase.lib.types.constants/type-hierarchies]]."
   [category column]
-  (let [type-definition (lib.types.constants/type-hierarchies category)]
+  (let [type-definition (lib.types.constants/type-hierarchies category)
+        column          (cond-> column
+                          (and (map? column)
+                               (not (:effective-type column)))
+                          (assoc :effective-type (:base-type column)))]
     (cond
       (nil? column) false
 


metamben (approved):

paoliniluis:
Is this in v49 already @snoe ?

snoe:
This backport was deemed unnecessary
https://metaboat.slack.com/archives/C052ZBWRG3W/p1710335533341969

PR #40144, opened 2024-03-14
Upgrade to React 17

### LINKED ISSUE #40178: Upgrade to React 17

sloansparger:
Closes #40178

### Description

This PR upgrades us to React 17. A direct upgrade to React 18 was causing too many test failures, so the goal is to start walk some of our dependencies forward to help surface a smaller amount of errors at a time and make the source of those issues clearer.

I've tried to highlighted in code comments the motivation for each of the changes that feel like they require more context.

### How to verify

- Play around with the app and make sure nothing unexpected happens.

### Checklist


frontend/src/metabase/admin/settings/components/widgets/GroupMappingsWidget/AddMappingRow/AddMappingRow.tsx
```diff
@@ -43,10 +43,7 @@ function AddMappingRow({
   return (
     <tr>
       <td colSpan={3} style={{ padding: 0 }}>
-        <form
-          className="m2 p1 bordered border-brand justify-between rounded relative flex align-center"
-          onSubmit={isMappingNameUnique ? handleSubmit : undefined}
-        >
+        <div className="m2 p1 bordered border-brand justify-between rounded relative flex align-center">
```
sloansparger, line 46: the nested form instead of another form was causing the inner form to not work

frontend/src/metabase/visualizations/shared/components/RowChart/RowChart.unit.spec.tsx
```diff
@@ -213,7 +213,7 @@ describe("RowChart", () => {
       userEvent.click(bars[0]);
 
       expect(onClick).toHaveBeenCalledWith(
-        expect.objectContaining({ currentTarget: expect.anything() }),
+        expect.objectContaining({ target: expect.anything() }),
```
sloansparger, line 216: event pooling was removed in react 17 which changes how currentTarget works

package.json
```diff
@@ -54,7 +54,7 @@
     "dayjs": "^1.10.4",
     "dc": "2.1.9",
     "diff": "^3.2.0",
-    "formik": "^2.2.9",
+    "formik": "^2.4.5",
```
sloansparger, line 57: upgrading formik fixes type errors caused by earlier version + react 17's types

frontend/src/metabase/components/InputBlurChange/InputBlurChange.tsx
```diff
@@ -58,7 +58,7 @@ const InputBlurChange = (props: InputBlurChangeProps) => {
     [normalize, onBlurChange, value],
   );
 
-  useUnmount(() => {
+  useUnmountLayout(() => {
```
sloansparger, line 61: React 17 now runs useEffect cleanup functions asynchronously, so inputRef was coming back as undefined. Leveraging useLayoutEffect's clean up function allows to access the input right before it's unmounted.

frontend/src/metabase/ui/components/overlays/Popover/Popover.styled.tsx
```diff
@@ -22,7 +22,8 @@ export const getPopoverOverrides = (): MantineThemeOverride["components"] => ({
       onMouseDownCapture: (event: SyntheticEvent) => {
         // prevent nested popovers from closing each other
         // see useClickOutside in @mantine/hooks for the reference
-        event.nativeEvent.stopImmediatePropagation();
+        const target = event.target as HTMLElement;
+        target.setAttribute("data-ignore-outside-clicks", "true");
```
sloansparger, line 26: The way event bubbling works in React 17 is now more correct. Previously listeners mounted on `document` would not be affected by `stopImmediatePropagation`, but now they are. Mantine provides this data-attribute for ignoring outside clicks, so we can leverage this for the same effect.

ranquild (approved):

iethree (approved):
🥳

npfitz (approved):
🔥

PR #40149, opened 2024-03-14
Diagnostic Modal Updates

iethree:
### Description

Here, we've got a couple updates to the diagnostic modal:

1. **Handle uncaught errors better.** Adding the download button to the uncaught error component was causing certain errors to overflow, and made some pages look really bad. We made this better in a few ways:
	1. went back to just having a small ⚠️ icon that should fit in basically any errored component
	2. added a small border to the error component to make it clear what has errored - this is especially helpful on dashboards.
	3. gave that ⚠️ component an updated tooltip and made it clickable
	4. clicking the ⚠️ now pops up an intermediate generic error explanation modal from which a user can choose to open the error diagnostic modal.
	5. as a side-effect of all this, we no longer trigger API calls to gather diagnostic information unless a user opens the diagnostic modal - this will prevent duplicate information-gathering on pages that have multiple errored components at once.
2. **Better frontend error capturing** I discovered when testing this on stats that the production bundle handles error reporting differently than the hot-reloading dev bundle. 
    1. The biggest issue I discovered was that any error caught by our error boundary were swallowed. By adding a console.error to the error boundary component, we can be sure that our frontend error buffer gets those errors as well.   
    2. I also added some custom error serialization handling to make sure that error objects get properly serialized when downloaded in our diagnostic json.

Screen Shot 2024-03-14 at 3 22 09 PM
diagnosticsmodals

### Checklist

### DIFF
--- a/frontend/src/metabase/components/ErrorPages/use-error-info.ts
+++ b/frontend/src/metabase/components/ErrorPages/use-error-info.ts
@@ -9,6 +9,18 @@ import { UtilApi, MetabaseApi, SessionApi } from "metabase/services";
 import type { ErrorPayload, ReportableEntityName } from "./types";
 import { getEntityDetails, hasQueryData } from "./utils";
 
+const maybeSerializeError = (key: string, value: any) => {
+  if (value?.constructor.name === "Error") {
+    return {
+      name: value.name,
+      message: value.message,
+      stack: value.stack,
+      cause: value.cause,
+    };
+  }
+  return value;
+};
+
 export const useErrorInfo = (
   { enabled }: { enabled?: boolean } = { enabled: true },
 ) => {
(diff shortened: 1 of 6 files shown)


dhuniverse (approved):
Looks good to me.

frontend/src/metabase/components/ErrorPages/use-error-info.ts
```diff
@@ -9,6 +9,18 @@ import { UtilApi, MetabaseApi, SessionApi } from "metabase/services";
 import type { ErrorPayload, ReportableEntityName } from "./types";
 import { getEntityDetails, hasQueryData } from "./utils";
 
+const maybeSerializeError = (key: string, value: any) => {
+  if (value?.constructor.name === "Error") {
```
sloansparger, line 13: I _think_ it might be preferable to do `value instanceof Error` to capture custom err values that have extended the Error type:
```ts
class CustomError extends Error {};
const err1 = new Error('err1');
const err2 = new CustomError('err2');
err1?.constructor.name === "Error" // true
err1 instanceof Error // true
err2?.constructor.name === "Error" // false
err2 instanceof Error // true
```

sloansparger (approved):
Approved with one comment

Issue #39600, opened 2024-03-05
[dc.js migration] missing tooltip for goal line label


EmmadUsmani:
<img width="151" alt="Screenshot 2024-03-04 at 4 23 19 PM" src="https://github.com/metabase/metabase/assets/37751258/01a18e95-63e7-4943-8c2c-fd041ecef020">

Doesn't appear in echarts

PR #40150, opened 2024-03-14
Fix missing tooltip for goal line label

JesseSDevaney:
- Closes: https://github.com/metabase/metabase/issues/39600

**Before**

image

**After**

image

---
**Things to Consider:**
- Hovering over line causes line to be bold
  - Before
   image
  - After
   image
- See code comments for additional things to consider

### DIFF
--- a/frontend/src/metabase/visualizations/echarts/cartesian/option/goal-line.ts
+++ b/frontend/src/metabase/visualizations/echarts/cartesian/option/goal-line.ts
@@ -42,7 +42,6 @@ export function getGoalLineSeriesOption(
     },
     // we hide the above line, it only exists to prevent the goal line from
     // rendering out of bounds
-    silent: true,
     markLine: {
       data: [{ name: "goal-line", yAxis: settings["graph.goal_value"] }],
       label: {
--- a/frontend/src/metabase/visualizations/visualizations/CartesianChart/use-chart-events.ts
+++ b/frontend/src/metabase/visualizations/visualizations/CartesianChart/use-chart-events.ts
@@ -10,6 +10,7 @@ import type {
 import {
   canBrush,
   getBrushData,
+  getGoalLineHoverData,
   getSeriesClickData,
   getSeriesHoverData,
   getTimelineEventsForEvent,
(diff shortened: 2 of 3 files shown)


frontend/src/metabase/visualizations/echarts/cartesian/option/goal-line.ts
```diff
@@ -42,7 +42,6 @@ export function getGoalLineSeriesOption(
     },
     // we hide the above line, it only exists to prevent the goal line from
     // rendering out of bounds
-    silent: true,
```
JesseSDevaney, line 45: Removed so that we could start receiving goal line events on hover

frontend/src/metabase/visualizations/visualizations/CartesianChart/use-chart-events.ts
```diff
@@ -86,6 +87,16 @@ export const useChartEvents = (
             return;
           }
 
+          if (
+            event.name === "goal-line" &&
+            event.event?.target?.style?.text === settings["graph.goal_label"]
```
JesseSDevaney, line 92: The line and label trigger hover similar hover events. Is there a way to make only the label cause hover events to be triggered besides manually checking the element that triggered the event?

For example, without this manual check all hovers over the line and label cause tooltips to be created.

**Without Manual Check**

https://github.com/metabase/metabase/assets/22608765/9bf3f1f6-e117-4774-a7af-2e56baff45fb

**With Manual Check**


https://github.com/metabase/metabase/assets/22608765/a00e9b65-eb45-40a1-80e9-c8d55b41bb9c

JesseSDevaney:
Also, it is expected to have so many failing CI checks with the ECharts branch right now?

frontend/src/metabase/visualizations/visualizations/CartesianChart/use-chart-events.ts
alxnddr, line 92: That make sense, I think you can change the check to `target.nodeName === "text"`

alxnddr:
> Also, is it expected to have so many failing CI checks with the ECharts branch right now?

This is okay, there is a separate epic for fixing specs. Now we have bugs that would not make CI green and also all the selectors have changed significantly. But we do expect Loki specs to be green

alxnddr (approved):
I left one minor comment, besides that looks great 👍

PR #40151, opened 2024-03-14
Handle double escape in entityPicker

### LINKED ISSUE #39966: `Esc` exits both the collection picker and the question save modal

iethree:
Closes https://github.com/metabase/metabase/issues/39966

### Description

Prevents hitting `escape` in the entitypicker from also closing any other modals that are open at the same time. (Since entitypicker is often used above another modal (like new collection or new dashboard modal)

_caption: note the modals close separately when the user presses `escape` twice_
DOUBLE ESCAPE WHAT DOES IT MEAN

### Checklist

### DIFF
--- a/e2e/test/scenarios/dashboard/dashboard.cy.spec.js
+++ b/e2e/test/scenarios/dashboard/dashboard.cy.spec.js
@@ -71,12 +71,20 @@ describe("scenarios > dashboard", () => {
       appBar().findByText("New").click();
       popover().findByText("Dashboard").should("be.visible").click();
 
+      cy.log(
+        "pressing escape should only close the entity picker modal, not the new dashboard modal",
+      );
+      modal().findByTestId("collection-picker-button").click();
+      entityPickerModal().findByText("Select a collection");
+      cy.realPress("Escape");
+      modal().findByText("New dashboard").should("be.visible");
+
       cy.log("Create a new dashboard");
       modal().within(() => {
         // Without waiting for this, the test was constantly flaking locally.
         cy.findByText("Our analytics");
 
-        cy.findByLabelText("Name").type(dashboardName);
+        cy.findByPlaceholderText(/name of your dashboard/i).type(dashboardName);
         cy.findByLabelText("Description").type(dashboardDescription, {
           delay: 0,
         });
--- a/frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
+++ b/frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
@@ -1,4 +1,5 @@
-import { useState, useMemo, useEffect } from "react";
+import { useWindowEvent } from "@mantine/hooks";
+import { useState, useMemo } from "react";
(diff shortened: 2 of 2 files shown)


frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
```diff
@@ -83,6 +83,23 @@ export function EntityPickerModal<TItem extends TypeWithModel>({
...
 
+  useEffect(() => {
+    const handleEscapeModal = (event: KeyboardEvent) => {
+      if (event.key === "Escape") {
+        event.stopPropagation();
+        onClose();
+      }
+    };
+
+    document.addEventListener("keydown", handleEscapeModal, true);
```
iethree, line 94: I do not love attaching this event listener to the document, or forcing it to listen on the capture phase, but I couldn't figure out any more constrained way to keep it from preventing propagation to other modals.

e2e/test/scenarios/dashboard/dashboard.cy.spec.js
```diff
@@ -71,12 +71,20 @@ describe("scenarios > dashboard", () => {
       appBar().findByText("New").click();
       popover().findByText("Dashboard").should("be.visible").click();
 
+      cy.log(
+        "pressing escape should only close the entity picker modal, not the new dashboard modal",
+      );
+      modal().findByTestId("collection-picker-button").click();
+      entityPickerModal().findByText("Select a collection");
+      cy.realPress("Escape");
+      modal().findByText("New dashboard").should("be.visible");
```
rafpaf, line 80: theory: if pressing escape triggers the close of the "New dashboard" modal, that might take a split second, but line 80 might run before that happens, so the test won't catch it

rafpaf: maybe `modal().findByText("New dashboard").should("be.visible").wait(500).should('be.visible')`

rafpaf: not sure if there's a way to avoid the dreaded `cy.wait`...

iethree: (although, some context, the current behavior is that the "new" modal closes before the entity picker, destroying the entity picker as a consequence, so I'm pretty confident this test will pass for the correct reason)

frontend/src/metabase/common/components/EntityPicker/components/EntityPickerModal/EntityPickerModal.tsx
rafpaf: If I understand right, Mantine's `onClose` prop handles the modal's closure logic. Is `onClose` called twice now, once by `handleEscapeModal` and once by Mantine?

rafpaf: @iethree I'll approve and let you decide whether this is a concern

rafpaf (approved):
gorgeous!

Issue #38682, opened 2024-02-12
[dc.js migration] areas stacking is broken in any data that includes nulls on any axis type


alxnddr:
All stacked area charts that contain null y-values are broken. We encountered this issue multiple times during our bug bash, here is an example question: https://stats.metabase.com/question/4678-earned-impressions-by-channel ([screenshot attached](https://metaboat.slack.com/archives/C05NQSWPRMK/p1707493380950789)).

After we patched echarts, we need to handle the case when for a given X-axis value 1+ series have non-null values while others with linear interpolation enabled have nulls.

### Examples

question/14309
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194251461129

question/14607
https://metaboat.slack.com/archives/C06G94JTWBS/p1706193732427339

question/13718
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194079748279

question/3835
https://metaboat.slack.com/archives/C06G94JTWBS/p1706194544443529

PR #40157, opened 2024-03-15
fix messed up stacked areas with various missing.values setting

alxnddr:
Follow up on https://github.com/metabase/metabase/issues/38682

### Description

Fixes stacked areas on datasets with multiple series that include null values. This PR replicates the area stacking behavior on master which is: **if at least one series has "line.missing" is not "none", then we replace all missing values for all series with zeros**

### How to verify

Create stacked areas with multiple series and null values, try different compinations of "Replace missing values" settings for series.

The question used in the demo:
Screenshot 2024-03-14 at 10 17 37 PM

#### All series has `"line.missing": "none"`
master
Screenshot 2024-03-14 at 9 47 20 PM

echarts
Screenshot 2024-03-14 at 9 47 12 PM

#### 3 series have `"line.missing": "none"` but one has `"line.missing": "interpolate"`
master
Screenshot 2024-03-14 at 9 46 29 PM

echarts
Screenshot 2024-03-14 at 9 46 16 PM

### Checklist


frontend/src/metabase/static-viz/components/ComboChart/stories-data/area-stack-interpolate-missing-values.json
```diff
@@ -1,369 +1,400 @@
 [
   {
     "card": {
-      "original_card_id": 62,
```
alxnddr, line 4: Replaced this card since it did not catch all cases with a better variant

EmmadUsmani (approved):

PR #40158, opened 2024-03-15
Rename `metabase.mbql` to `metabase.legacy-mbql`

camsaul:
Also, move `metabase.mbql.util.match/match` and friends to `metabase.lib.util.match`, and remove the aliases in `metabase.legacy-mbql.util`, since they made things tricky with Cljs. 

`match` was the only thing we wanted to keep there; everything else in `metabase.legacy-util.*` is effectively deprecated at this point now


snoe (approved):
LGTM but I wonder two things:

Thinking about modularization of `metabase.lib` having these `lib.util` requires outside of `lib.core` makes it difficult to keep track of what's internal/external to mlv2.

How specific is `match` / `replace` to mlv2, I THINK it's more of a `core.match++` so maybe it belongs in `metabase.match`?

camsaul:
I do think we need to clean up the exports of `metabase.lib.*` and put a lot more stuff into `metabase.lib.core`. I want to do that at some point, but I was trying to keep this PR from getting too big.

The `match`/`replace` stuff has a lot of logic in it specific to MBQL queries, for example it can has built-in matching on clause tag or tags. So it's not really a general `match++`

PR #40162, opened 2024-03-15
OpenAPI docs

sansolovyov:
Adds OpenAPI generation for `defendpoint` and an interface for it at `/api/docs/`.

Caveats:
- I don't understand why clj-kondo refuses to see those 4 endpoints by name (`GET_docs*` mentioned in `api.routes` and others I'm using for tests)
- There is no security scheme defined, so it says "Authentication: Not Required" everywhere (see screenshot)
- If we have POST/PUT requests having query parameters, we need some way of differentiating them
- Lots of fields are `required`, but `... or null` - that needs to improve

I'd really love to fix clj-kondo, but don't understand how. I think others are fine to fix later on when we get a bit of feedback on this.

image


enterprise/backend/src/metabase_enterprise/api/routes/common.clj
```diff
@@ -16,9 +16,11 @@
   if we do not."
   [feature feature-name handler]
   (assert (i18n/localized-string? feature-name), "`feature-name` must be i18ned")
-  (fn [request respond raise]
-    (premium-features/assert-has-feature feature feature-name)
-    (handler request respond raise)))
+  (with-meta
+   (fn [request respond raise]
```
escherize, line 20: I saw a weird thread on clojurians slack about functions not "technically" having support for metadata, and that it's "an implementation detail". I think this is fine though, I doubt they'll ever change it. [thread here](https://clojurians.slack.com/archives/C06E3HYPR/p1709908290330649)

src/metabase/api/automagic_dashboards.clj
```diff
@@ -272,22 +272,24 @@
 (api/defendpoint GET "/:entity/:entity-id-or-query/rule/:prefix/:dashboard-template"
   "Return an automagic dashboard for entity `entity` with id `id` using dashboard-template `dashboard-template`."
   [entity entity-id-or-query prefix dashboard-template show]
-  {entity Entity
-   show   Show
-   prefix Prefix
-   dashboard-template   DashboardTemplate}
+  {entity             Entity
+   entity-id-or-query string?
```
escherize, line 276:
```suggestion
   entity-id-or-query ms/NonBlankString
```

I think the rule should be: on the api layer prefer the ms/* vars (since they have error api messages on them), then if one exists a keyword spec `:string` and when there isn't a keyword for it, to use the function, like with `pos-int?`.

src/metabase/api/automagic_dashboards.clj
```diff
@@ -296,23 +298,25 @@
...
 
 (api/defendpoint GET "/:entity/:entity-id-or-query/compare/:comparison-entity/:comparison-entity-id-or-query"
   "Return an automagic comparison dashboard for entity `entity` with id `id` compared with entity
    `comparison-entity` with id `comparison-entity-id-or-query.`"
   [entity entity-id-or-query show comparison-entity comparison-entity-id-or-query]
-  {show              Show
-   entity            Entity
-   comparison-entity ComparisonEntity}
+  {show               Show
+   entity-id-or-query string?
```
escherize, line 317:
```suggestion
   entity-id-or-query ms/NonBlankString
```

src/metabase/api/automagic_dashboards.clj
```diff
@@ -324,28 +328,30 @@
   "Return an automagic comparison dashboard for entity `entity` with id `id` using dashboard-template `dashboard-template`;
    compared with entity `comparison-entity` with id `comparison-entity-id-or-query.`."
   [entity entity-id-or-query prefix dashboard-template show comparison-entity comparison-entity-id-or-query]
-  {entity            Entity
-   show              Show
-   prefix            Prefix
-   dashboard-template              DashboardTemplate
-   comparison-entity ComparisonEntity}
+  {entity             Entity
+   entity-id-or-query string?
```
escherize, line 332:
```suggestion
   entity-id-or-query ms/NonBlankString
```

src/metabase/api/automagic_dashboards.clj
```diff
@@ -324,28 +328,30 @@
...
   "Return an automagic comparison dashboard for cell in automagic dashboard for entity `entity`
    with id `id` defined by query `cell-query`; compared with entity `comparison-entity` with id
    `comparison-entity-id-or-query.`."
   [entity entity-id-or-query cell-query show comparison-entity comparison-entity-id-or-query]
-  {entity            Entity
-   show              Show
-   cell-query        Base64EncodedJSON
-   comparison-entity ComparisonEntity}
+  {entity             Entity
+   entity-id-or-query string?
```
escherize, line 351:
```suggestion
   entity-id-or-query ms/NonBlankString
```

src/metabase/api/automagic_dashboards.clj
```diff
@@ -358,17 +364,18 @@
...
   [entity entity-id-or-query cell-query prefix dashboard-template show comparison-entity comparison-entity-id-or-query]
-  {entity            Entity
-   show              Show
-   prefix            Prefix
-   dashboard-template              DashboardTemplate
-   cell-query        Base64EncodedJSON
-   comparison-entity ComparisonEntity}
+  {entity             Entity
+   entity-id-or-query string?
+   show               Show
```
escherize, line 369:
```suggestion
   entity-id-or-query ms/NonBlankString
```

escherize (commented):
It looks good, and I like the tests, but I'm not sure what the exact goal is, could you put in link to the issue so I can know exactly what this is solving?

I am also lacking some context for parsing out what a few of your pr's comments mean:

> - I don't understand why clj-kondo refuses to see those 4 endpoints by name

Which 4?

> - There is no security scheme defined, so it says "no authentication" everywhere

Do we need to define one? What's everywhere?

> - If we have POST/PUT requests having query parameters, we need some way of differentiating them

Is this going to be some followup work?

enterprise/backend/src/metabase_enterprise/api/routes/common.clj
piranha: Uhhh weird right?! Anyway, I think we don't have too many options here - operating vars in this place would _weird_.

src/metabase/api/automagic_dashboards.clj
piranha: Agree, I was just too eager to finally see something and that place was in the way. 😁 Fixed all of them.

piranha:
> Is this going to be some followup work?

Yeah, I think so. Making this perfect will take lots of time - meanwhile it's going to be useful as it is.

escherize (approved):
If we're good with follow up work, let's merge it and keep adding to it along the way: 👍

piranha:
Right... I still need to understand why `clj-kondo` is unhappy. 😁

PR #40169, opened 2024-03-15
Fix milestone name in release script

kamilmielnik:
See https://metaboat.slack.com/archives/C864UT5CZ/p1710495982900369

Yesterday we've had milestone `0.50.0` created instead of `0.50`.


nemanjaglumac (approved):
👏

PR #40181, opened 2024-03-15
Migrate composeDataset and composeThisQuery to MBQL lib

ranquild:
For metrics, creating a ad-hoc question is not as simple as `"source-table": "card__1"`. There is an aggregation clause based on the metric, default breakouts, filters etc. All of this should be handled internally by MBQL lib if we call the appropriate method under the hood.

We need to call `queryFromTableOrCardMetadata` to create an ad-hoc question based on this query. The problem I faced immediately is that we create ad-hoc questions before metadata for the current (saved) model is fetched. So we need to fix our metadata fetching logic to load data for the saved model upfront before even attempting to create an ad-hoc question for this model. This is what this PR does. 

How to verify:
- CI is green


frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -605,33 +597,6 @@ class Question {
...
-    const type = this.type();
-    const isModel = type === "model";
-    if (isModel && this.isSaved()) {
-      dependencies.push({
-        type: "table",
-        id: getQuestionVirtualTableId(this.id()),
-      });
-    }
-
-    this.getResultMetadata().forEach(field => {
```
ranquild, line 623: Moved to `fe_util`

frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -605,33 +597,6 @@ class Question {
...
 
-  dependentMetadata(): Lib.DependentItem[] {
-    const dependencies = [];
-
-    // we frequently treat dataset/model questions like they are already nested
-    // so we need to fetch the virtual card table representation of the Question
-    // so that we can properly access the table's fields in various scenarios
-    const type = this.type();
-    const isModel = type === "model";
-    if (isModel && this.isSaved()) {
```
ranquild, line 616: This is now in `question/actions`

frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -90,6 +89,10 @@ class Question {
    */
   _parameterValues: ParameterValues;
 
+  private __mlv2Query: Lib.Query | undefined;
```
ranquild, line 92: Declaring the fields here. They exists before this PR

frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -433,36 +436,25 @@ class Question {
    * of Question interface instead of Query interface makes it more convenient to also change the current visualization
    */
 
-  composeThisQuery(): Question | null | undefined {
```
ranquild, line 436: This is now called `composeQueryAdHoc`

src/metabase/lib/fe_util.cljc
```diff
@@ -151,6 +153,15 @@
       _
       (lib.metadata.calculation/display-name query stage-number filter-clause))))
 
+(defn- query-dependents-foreign-keys
+  [metadata-providerable columns]
+  (for [column columns
+        :let [fk-target-field-id (:fk-target-field-id column)]
+        :when (and fk-target-field-id (lib.types.isa/foreign-key? column))]
+    (if-let [fk-target-field (lib.metadata/field metadata-providerable fk-target-field-id)]
```
ranquild, line 161: For implicit joins we need the whole table. If the field is available, get `table-id` from it. Otherwise we need another FE-BE round-trip to fetch missing metadata.

frontend/src/metabase/dashboard/actions/metadata.js
```diff
@@ -19,18 +16,11 @@ export const loadMetadataForDashboard = dashCards => async dispatch => {
...
-  const dependentItems = questions.flatMap(question => [
-    ...question.dependentMetadata(),
-    ...Lib.dependentMetadata(question.query()),
-  ]);
-
-  return dispatch(loadMetadataForDependentItems(dependentItems));
+  return Promise.all(
+    cards
+      .filter(card => card.dataset_query) // exclude queries without perms
+      .map(card => dispatch(loadMetadataForCard(card))),
```
ranquild, line 22: Use `loadMetadataForCard` instead of sem-duplicated logic

frontend/src/metabase/query_builder/actions/core/updateQuestion.ts
```diff
@@ -214,20 +214,8 @@ export const updateQuestion = (
...
-      : [];
-    const nextDependencies = [
-      ...newQuestion.dependentMetadata(),
-      ...Lib.dependentMetadata(newQuestion.query()),
-    ];
     try {
-      if (!_.isEqual(currentDependencies, nextDependencies)) {
-        await dispatch(loadMetadataForCard(newQuestion.card()));
-      }
+      await dispatch(loadMetadataForCard(newQuestion.card()));
```
ranquild, line 218: Always try to load dependencies because new items can appear after loading

frontend/src/metabase/query_builder/actions/models.js
```diff
@@ -49,8 +49,7 @@ export const turnQuestionIntoDataset = () => async (dispatch, getState) => {
   const metadata = getMetadata(getState());
   const dataset = metadata.question(question.id());
 
-  const dependentItems = dataset.dependentMetadata();
-  await dispatch(loadMetadataForDependentItems(dependentItems));
+  await dispatch(loadMetadataForCard(dataset.card()));
```
ranquild, line 52: Use `loadMetadataForCard` instead of semi-duplicated logic

frontend/src/metabase/questions/actions.ts
```diff
@@ -11,15 +11,15 @@ export interface LoadMetadataOptions {
 
 export const loadMetadataForCard =
   (card: Card, options?: LoadMetadataOptions) =>
-  (dispatch: Dispatch, getState: GetState) => {
+  async (dispatch: Dispatch, getState: GetState) => {
     const question = new Question(card, getMetadata(getState()));
-    const queries =
-      question.type() === "question"
-        ? [question.query()]
-        : [question.query(), question.composeDataset().query()];
```
ranquild, line 19: We cannot create an ad-hoc question before fetching metadata for the underlying question; doing so would fail

frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -433,36 +436,25 @@ class Question {
...
-        },
-      };
-      return this.setCard(card);
+  composeQuestion(): Question {
+    if (!this.isSaved()) {
+      return this;
     }
-  }
 
-  composeDataset(): Question {
```
ranquild, line 452: This is now called `composeQuestion`

frontend/src/metabase-lib/v1/Question.ts
```diff
@@ -433,36 +436,25 @@ class Question {
...
-    const type = this.type();
+    const metadata = this.metadataProvider();
+    const tableId = getQuestionVirtualTableId(this.id());
+    const table = Lib.tableOrCardMetadata(metadata, tableId);
+    const query = Lib.queryFromTableOrCardMetadata(metadata, table);
+    return this.setQuery(query);
+  }
 
-    if (type === "question" || !this.isSaved()) {
+  composeQuestionAdhoc(): Question {
```
kamilmielnik, line 451: I fail to recognize the difference between `composeQuestionAdhoc` and `composeQuestion`.
Can't we just remove `composeQuestionAdhoc` and use `composeQuestion().composeQuestion()`?

Can you briefly describe it please?

kamilmielnik (approved):

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
```diff
@@ -200,12 +200,12 @@ class Visualization extends PureComponent {
     const question = new Question(card, metadata);
 
     // Datasets in QB should behave as raw tables opened in simple mode
-    // composeDataset replaces the dataset_query with a clean query using the dataset as a source table
+    // composeQuestion replaces the dataset_query with a clean query using the dataset as a source table
     // Ideally, this logic should happen somewhere else
```
nemanjaglumac, line 204: Does this still hold true, or we can remove the last line in the comment?

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
```diff
@@ -200,12 +200,12 @@ class Visualization extends PureComponent {
     const question = new Question(card, metadata);
 
     // Datasets in QB should behave as raw tables opened in simple mode
-    // composeDataset replaces the dataset_query with a clean query using the dataset as a source table
+    // composeQuestion replaces the dataset_query with a clean query using the dataset as a source table
```
nemanjaglumac, line 203: Nit: Can we replace `datasets` with `models` in the comment?

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
ranquild: Still true

frontend/src/metabase/questions/actions.ts
nemanjaglumac: This comment deserves to live in the code, imo.

frontend/src/metabase-lib/v1/Question.ts
ranquild: `composeQuestionAdhoc` creates an ad-hoc question based on this question. `composeQuestion` replaces the query in the current question "in-place" without making it an ad-hoc question.

frontend/src/metabase/visualizations/components/Visualization/Visualization.jsx
ranquild: I'll address that in another PR

PR #40184, opened 2024-03-15
Re-export Mantine `rem` util

nemanjaglumac:
This PR re-exports a very helpful Mantine util `rem`.
It also uses it in a about a dozen components just as a PoC.

### DIFF
--- a/frontend/src/metabase/ui/components/utils/index.ts
+++ b/frontend/src/metabase/ui/components/utils/index.ts
@@ -3,4 +3,5 @@ export * from "./DelayGroup";
 export * from "./Divider";
 export * from "./FocusTrap";
 export * from "./Paper";
+export * from "./Rem";
 export * from "./Transition";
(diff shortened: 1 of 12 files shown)


ranquild (commented):
Inline comment

frontend/src/metabase/ui/components/utils/index.ts
```diff
@@ -3,4 +3,5 @@ export * from "./DelayGroup";
 export * from "./Divider";
 export * from "./FocusTrap";
 export * from "./Paper";
+export * from "./Rem";
```
ranquild, line 6: That's not a util "component". I'd re-export `rem` from `metabase/ui/index.ts` directly.

nemanjaglumac: Tbh, I was confused that we have util components.
This belongs to utils, but we don't have it.

To avoid introducing more confusion, I'll export it directly from index.ts as you suggested.

ranquild (approved):

PR #40192, opened 2024-03-15
docs - update list of releases

jeff-bruemmer:
Updates list of releases.


alexyarosh (approved):
Thank you @jeff-bruemmer !

PR #40198, opened 2024-03-15
Embedding SDK - integration commit

deniskaber:
Relates to https://github.com/metabase/metabase/issues/40241

### Description

Embedding SDK code.

### Checklist


frontend/src/metabase/lib/api.js
```diff
@@ -58,6 +64,10 @@ export class Api extends EventEmitter {
       };
 
       return async (rawData, invocationOptions = {}) => {
+        if (this.onBeforeRequest) {
+          await this.onBeforeRequest();
+        }
```
WiNloSt, line 69: I haven't seen this before, what is this for?

frontend/src/metabase/redux/settings.ts
```diff
@@ -20,7 +20,7 @@ export const refreshSiteSettings = createAsyncThunk(
 );
 
 export const settings = createReducer(
-  { values: window.MetabaseBootstrap, loading: false },
+  { values: window.MetabaseBootstrap || {}, loading: false },
```
WiNloSt, line 23: Why is this needed? Have some types changed?

frontend/src/metabase/visualizations/components/ChoroplethMap.jsx
```diff
@@ -153,7 +153,7 @@ export default class ChoroplethMap extends Component {
           this.setState({
             geoJson: geoJson,
             geoJsonPath: geoJsonPath,
-            minimalBounds: computeMinimalBounds(geoJson.features),
+            minimalBounds: computeMinimalBounds(geoJson?.features ?? []),
```
WiNloSt, line 156: Why is this needed? Have some types changed?

tsconfig.sdk.json
```diff
@@ -0,0 +1,38 @@
...
+  "include": [
+    "frontend/src/**/*",
+    "enterprise/frontend/src/**/*",
+    "frontend/test/**/*",
+    "enterprise/frontend/test/**/*"
+  ],
+  "exclude": [
+    "node_modules",
+    // The following files will load `frontend/src/metabase/app` which will conflict with
+    // `frontend/src/metabase/App.tsx`, since now we're loading both JS and TS files.
```
WiNloSt, line 26: We probably should do something about this (not in this PR though). VS Code confuses `frontend/src/metabase/App.tsx` with `frontend/src/metabase/app.js` and that's frustrated when working with these files.

WiNloSt (commented):
I haven't reviewed the webpack config. I think Uladzimir work with that a lot, it could be great to have him review it as well.

tsconfig.sdk.json
npretto: maybe renaming `app.js` to `init.js` could help, as it seems it's just code to init the app

WiNloSt: yeah, that would do. We don't have `init.*` in `frontend/src/metabase/` currently.

enterprise/frontend/src/embedding-sdk/components/public/QueryVisualization.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const { questionId } = props;
+
+  const [{ loading, card, result }, setState] = useState<State>({
+    loading: false,
+    card: null,
+    result: null,
+  });
+
+  const loadCardData = async ({ questionId }: { questionId: number }) => {
+    setState(prevState => ({
```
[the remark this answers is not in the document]
npretto: I haven't found any use of it in our codebase, is the slightly more concise code worth the headache of understanding which one we're using every time we see a `setXXX`?
Using both `useState` and `useSetState` means we'd have to check which hook (that differ in 3 letters) we're using to understand if we're writing over the state, or just updating some fields.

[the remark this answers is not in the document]
WiNloSt: Fair enough.

tsconfig.sdk.json
WiNloSt: Yeah, I know that because I wrote that comment 😅 . It's just that this is something we should be aware of at least because I'm not particularly satisfied with the solution I went with either.

enterprise/frontend/src/embedding-sdk/components/public/QueryVisualization.tsx
```diff
@@ -0,0 +1,165 @@
...
+  useEffect(() => {
+    if (!isInitialized || !isLoggedIn) {
+      setState({
+        loading: false,
+        card: null,
+        result: null,
+        cardError: null,
+        resultError: null,
+      });
+    } else {
```
WiNloSt, line 93: Oh, I like the `else` part more compared to the previous solution. I was thinking something felt off, but I couldn't think of anything else at that time.

enterprise/frontend/src/embedding-sdk/hooks/private/use-init-data.ts
```diff
@@ -0,0 +1,55 @@
...
+      setIsLoggedIn(false);
+      return;
+    }
+
+    Promise.all([
+      dispatch(refreshCurrentUser()),
+      dispatch(reloadSettings()),
+    ]).then(() => {
+      setIsInitialized(true);
+      setIsLoggedIn(true);
```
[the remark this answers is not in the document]
WiNloSt: Alright, I found that we reject anything not within 2xx status, so this is good. I was worried errored requests are still going through `.then`

https://github.com/metabase/metabase/blob/dc279fde80ee68f2324244839a963e5e4a1812c4/frontend/src/metabase/lib/api.js#L202-L212

enterprise/frontend/src/embedding-sdk/components/private/AppInitializeController.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }, [dispatch, font]);
+
+  return (
+    <EmbeddingContext.Provider
+      value={{
+        isInitialized,
+        isLoggedIn,
+      }}
+    >
+      <SdkContentWrapper data-elementid="sdk-content-wrapper" font={font}>
```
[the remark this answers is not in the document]
WiNloSt: And what is `data-elementid` for?

WiNloSt (approved):
This is already a lot better, but I'd like more people to have a look at the webpack config.

webpack.embedding-sdk.config.js
```diff
@@ -0,0 +1,215 @@
...
+          type: "asset/inline",
+          resourceQuery: { not: [/component|source/] },
+        },
+        {
+          test: /\.css$/,
+          use: [
+            {
+              loader: "style-loader",
+            },
+            //   { loader: MiniCssExtractPlugin.loader },
```
uladzimirdev, line 102: I think I tried it in a main webpack config and it broke half of the styles

webpack.embedding-sdk.config.js
```diff
@@ -0,0 +1,215 @@
...
+      hints: false,
+    },
+
+    entry: "./index.ts",
+
+    output: {
+      path: BUILD_PATH + "/dist",
+      publicPath: "",
+      filename: "[name].bundle.js",
+      libraryTarget: "commonjs2",
```
uladzimirdev, line 65: https://webpack.js.org/configuration/output/#type-commonjs2

```
output: {
    library: {
      // note there's no `name` here
      type: 'commonjs2',
    },
  },
```

webpack.embedding-sdk.config.js
```diff
@@ -0,0 +1,215 @@
...
+        }
+      : undefined,
+
+    plugins: [
+      new NodePolyfillPlugin(), // for crypto, among others
+      // https://github.com/remarkjs/remark/discussions/903
+      new webpack.ProvidePlugin({
+        process: "process/browser.js",
+      }),
+      new ForkTsCheckerWebpackPlugin({
```
uladzimirdev, line 199: I don't see a reason to use it instead of normal tsc at CI, but if you have arguments :+1:

uladzimirdev (approved):
Added some minor comments re webpack config. It's very similar to our main webpack config.

Overall comments:

1. webpack config can be reused, we really override just a few lines, so no need to copy everything, especially it will become a nightmare in support
2. nit. It looks like we can keep `package.json.template` with empty dependencies if the idea is to use it only for publishing and remove `.template` in the pre-publish step. It _can_ potentially save us from unnecessary node_modules installation.

webpack.embedding-sdk.config.js
uladzimirdev: isn't tsc enough for it?

Issue #39590, opened 2024-03-04
[dc.js migration] normalized stacked tooltip missing `%` symbol


EmmadUsmani:
question/13700
https://metaboat.slack.com/archives/C06G94JTWBS/p1706195416551909

| dc.js | echarts |
|--------|--------|
| <img width="167" alt="Screenshot 2024-03-04 at 3 36 17 PM" src="https://github.com/metabase/metabase/assets/37751258/c5dbf6d9-459e-455f-900e-5976c5c3989e"> | <img width="178" alt="Screenshot 2024-03-04 at 3 36 21 PM" src="https://github.com/metabase/metabase/assets/37751258/0abed8e6-3fc5-44b1-b08d-f37b7ccb3330"> |

alxnddr:
Not a regression

PR #40201, opened 2024-03-15
Fix normalized, stacked bar tooltip names

JesseSDevaney:
- Closes: https://github.com/metabase/metabase/issues/39590

### Before

**Non-normalized Stacking**

image

**Normalized Stacking**

image

### After

**Non-normalized Stacking**

image

**Normalized Stacking**

image


alxnddr (commented):
@JesseSDevaney sorry for not noticing it before, to me it seems that this tooltip does not need "%" prefix. I started the discussion [here](https://metaboat.slack.com/archives/C05NQSWPRMK/p1710817189360489)

PR #40203, opened 2024-03-15
/api/collection/:id/items gets `here` and `below`

johnswanson:
This code could probably use a refactor, but:

During postprocessing of `collection` children retrieved by `/api/collection/:id/items`, annotate each of the children with `below` and `here` keys representing the presence of questions, models, or collections at this level of the hierarchy or below it.

### DIFF
--- a/src/metabase/api/collection.clj
+++ b/src/metabase/api/collection.clj
@@ -575,8 +575,6 @@
                          (conj accu parent-id)))
                      #{}))
 
-
-
         child-type->coll-id-set
         (merge child-type->coll-id-set {:collection collections-containing-collections
                                         :dashboard collections-containing-dashboards})
--- a/src/metabase/api/collection.clj
+++ b/src/metabase/api/collection.clj
@@ -546,6 +546,12 @@
   (let [visible-collection-ids (collection/permissions-set->visible-collection-ids
                                 @api/*current-user-permissions-set*)
 
+        descendant-collections (collection/descendants-flat parent-coll (collection/visible-collection-ids->honeysql-filter-clause
+                                                                         :id
+                                                                         visible-collection-ids))
+
+        descendant-collection-ids (map u/the-id descendant-collections)
+
         child-type->coll-id-set
         (reduce (fn [acc {collection-id :collection_id, card-type :type, :as _card}]
                   (update acc (if (= (keyword card-type) :model) :dataset :card) conj collection-id))
@@ -553,22 +559,22 @@
                  :card    #{}}
                 (mdb.query/reducible-query {:select-distinct [:collection_id :type]
                                             :from            [:report_card]
-                                            :where           [:= :archived false]}))
+                                            :where           [:and
+                                                              [:= :archived false]
+                                                              [:in :collection_id descendant-collection-ids]]}))
 
         collections-containing-dashboards
         (->> (t2/query {:select-distinct [:collection_id]
                         :from :report_dashboard
-                        :where [:= :archived false]})
+                        :where [:and
+                                [:= :archived false]
+                                [:in :collection_id descendant-collection-ids]]})
              (map :collection_id)
              (into #{}))
 
-        descendants (collection/descendants-flat parent-coll (collection/visible-collection-ids->honeysql-filter-clause
-                                                              :id
-                                                              visible-collection-ids))
-
         ;; the set of collections that contain collections (in terms of *effective* location)
         collections-containing-collections
-        (->> descendants
+        (->> descendant-collections
              (reduce (fn [accu {:keys [location] :as _coll}]
                        (let [effective-location (collection/effective-location-path location visible-collection-ids)
                              parent-id (collection/location-path->parent-id effective-location)]
@@ -576,15 +582,16 @@
                      #{}))
 
         child-type->coll-id-set
-        (merge child-type->coll-id-set {:collection collections-containing-collections
-                                        :dashboard collections-containing-dashboards})
+        (merge child-type->coll-id-set
+               {:collection collections-containing-collections
+                :dashboard collections-containing-dashboards})
 
         ;; why are we calling `annotate-collections` on all descendants, when we only need the collections in `colls`
         ;; to be annotated? Because `annotate-collections` works by looping through the collections it's passed and
         ;; using them to figure out the ancestors of a given collection. This could use a refactor - probably the
         ;; caller of `annotate-collections` could be generating both `child-type->parent-ids` and
         ;; `child-type->ancestor-ids`.
-        coll-id->annotated (m/index-by :id (collection/annotate-collections child-type->coll-id-set descendants))]
+        coll-id->annotated (m/index-by :id (collection/annotate-collections child-type->coll-id-set descendant-collections))]
     (for [coll colls]
       (merge coll (select-keys (coll-id->annotated (:id coll)) [:here :below])))))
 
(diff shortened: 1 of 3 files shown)


test/metabase/api/collection_test.clj
```diff
@@ -966,6 +966,37 @@
...
+            (is (nil? (:below (item)))))
+          (t2.with-temp/with-temp [:model/Collection _ {:location (format "/%d/%d/%d/" id1 id2 id3)}]
+            (testing "the item still has a collection in it"
+              (is (= ["collection"] (:here (item)))))
+            (testing "the item now has a collection below it"
+              (is (= ["collection"] (:below (item))))))
+          (t2.with-temp/with-temp [:model/Card _ {:name "card" :collection_id id2}
+                                   :model/Card _ {:name "dataset" :type :model :collection_id id2}]
+            (testing "when the item has a card/dataset, that's reflected in `here` too"
+              (is (= #{"collection" "card" "dataset"} (set (:here (item)))))
```
johnswanson, line 991: `card` and `dataset` are for consistency with `/tree` - but it occurs to me that we probably want to change this someday to `question` and `model`. Not sure if we should do that now, or if it will make things harder for the FE.

src/metabase/models/collection.clj
```diff
@@ -1221,29 +1227,38 @@
...
-                            collections)]
+  in the collection at that level.
+
+  The second argument is the list of collections to annotate.
+
+  The first argument to this function could use a bit of explanation: `child-type->parent-ids` is a map. Keys are
+  object types (e.g. `:collection`), values are sets of collection IDs that are the (direct) parents of one or more
+  objects of that type.
+  "
+  [child-type->parent-ids collections]
```
noahmoss, line 1240: I like `child-type->parent-ids` as a name. Maybe use this for `collections->tree` below as well for consistency?

src/metabase/models/collection.clj
```diff
@@ -1221,29 +1227,38 @@
...
+                (zipmap (keys child-type->parent-ids) (repeat #{}))
+                collections)]
     (map (fn [{:keys [id] :as collection}]
-           (let [types (cond-> #{}
-                         (contains? (:dataset parent-info) id)
-                         (conj :dataset)
-                         (contains? (:card parent-info) id)
-                         (conj :card))]
+           (let [below (apply set/union
+                              (for [[type coll-id-set] child-type->ancestor-ids]
```
noahmoss, line 1253: nit: maybe use a different binding than `type` to avoid shadowing `clojure.core/type`? a couple other place in the codebase use `tyype` (though I'm not sure how much I like that either 🤔 )

src/metabase/models/collection.clj
```diff
@@ -1221,29 +1227,38 @@
                                :allowed-namespaces   allowed-namespaces
                                :collection-namespace collection-namespace})))))))
 
-(defn- annotate-collections
+(defn annotate-collections
```
noahmoss, line 1230: What's the main change here? Is it just changing `:below` to reflect the entire subtree instead of just one level down? Maybe that can be made clearer in the doc string since the code is a bit dense

src/metabase/api/collection.clj
```diff
@@ -541,16 +542,65 @@
...
+                                            :where           [:= :archived false]}))
+
+        collections-containing-dashboards
+        (->> (t2/query {:select-distinct [:collection_id]
+                        :from :report_dashboard
+                        :where [:= :archived false]})
+             (map :collection_id)
+             (into #{}))
+
+        descendants (collection/descendants-flat parent-coll (collection/visible-collection-ids->honeysql-filter-clause
```
noahmoss, line 566: `descendants` is shadowing `clojure.core/descendants`, would be nice if we could avoid that

src/metabase/api/collection.clj
```diff
@@ -541,16 +542,65 @@
...
+        ;; the set of collections that contain collections (in terms of *effective* location)
+        collections-containing-collections
+        (->> descendants
+             (reduce (fn [accu {:keys [location] :as _coll}]
+                       (let [effective-location (collection/effective-location-path location visible-collection-ids)
+                             parent-id (collection/location-path->parent-id effective-location)]
+                         (conj accu parent-id)))
+                     #{}))
+
+
```
noahmoss, line 579: nit
```suggestion
```

src/metabase/api/collection.clj
```diff
@@ -541,16 +542,65 @@
...
+                                @api/*current-user-permissions-set*)
+
+        child-type->coll-id-set
+        (reduce (fn [acc {collection-id :collection_id, card-type :type, :as _card}]
+                  (update acc (if (= (keyword card-type) :model) :dataset :card) conj collection-id))
+                {:dataset #{}
+                 :card    #{}}
+                (mdb.query/reducible-query {:select-distinct [:collection_id :type]
+                                            :from            [:report_card]
+                                            :where           [:= :archived false]}))
```
noahmoss, line 557: I'm a little concerned about the performance here. This is fetching _all_ collections on the instance that contain cards, dashboards, or collections, just to annotate `:here` and `:below` on a subset of collections passed into this function?

That seems like it's doing a lot more work than necessary. Could there be issues if the instance has a huge number of collections? Can we scope these queries to just the collections that are relevant?

(If the answer is no, and I'm misunderstanding something, maybe a docstring/comment could help to explain why for other readers)

noahmoss (commented):
Looks like it functions well; I mainly have a question on performance

src/metabase/api/collection.clj
johnswanson: Hmm, I think we could actually - we could get the *collections* that are descendants of the parent first. Then add `[:in :collection_id (map :id descendants)]` to the conditions here (and for the similar query selecting `collection_id` from dashboards below).

Worth noting that we do something very similar for the `/tree` endpoint, so I think so far this hasn't been a problem for performance. But it made me uncomfortable too, I just hadn't thought of a good solution until now.

src/metabase/models/collection.clj
johnswanson: The difference is more about generality. The old code took a map of `{:dataset set-of-collections-that-contain-datasets :card set-of-collections-that-contain-cards}` and annotated by setting `:here` and `:below` to a subset of `#{:card :dataset}`. In other words it just annotated `:card` and `:dataset`.

The new code takes a map of `{[arbitrary-entity] set-of-collections-that-contain-that-entity-type}` and annotates by setting `:here` and `:below` accordingly. I used it here to add `:collection` and `:dashboard` to the possible set in `:here`/`:below`.

src/metabase/models/collection.clj
```diff
@@ -444,7 +444,27 @@
                             [:children [:set [:ref ::children]]]]]}}
    [:ref ::children]])
 
-(mu/defn ^:private descendants :- [:set Children]
+(mu/defn descendants-flat :- [:set CollectionWithLocationAndIDOrRoot]
+  "Return all descendant collections of a `collection`, including children, grandchildren, and so forth."
+  [collection :- CollectionWithLocationAndIDOrRoot, & additional-honeysql-where-clauses]
+  (or
+   (t2/select-fn-set identity [:model/Collection :name :id :location :description]
```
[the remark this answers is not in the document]
johnswanson: Oops, good point.

noahmoss (approved):

PR #40217, opened 2024-03-17
Convert simple hydration to batched hydration

### LINKED ISSUE #38993: Replace toucan2 simple hydrations with batched hydrations
A batched hydration works for both single entity or a seq of entities, so we should use batched hydration in most cases because it saves us from running into N+1 problems like [this](https://metaboat.slack.com/archives/C064EB1UE5P/p1708387795469919?thread_ts=1708358454.946419&cid=C064EB1UE5P).

From Cam
> The main use case of the simple hydration stuff was for adding calculated columns like :full_name to User (first_name + last_name), fetching additional stuff from the DB is obviously a no-no. We could easily have Toucan 2 warn or even error if you're hitting the app DB inside a simple hydration method if we don't want to eliminate them altogether.

E.g: you want to write a `:fields` hydration for `:table`
```clojure
(mi/define-batched-hydration-method with-fields
  :fields
  "Efficiently hydrate the Fields for a collection of `tables`."
  [tables]
  (let [table-id->fields (group-by :table_id (t2/select :model/Field
                                                        :table_id [:in (map :id tables)]
                                                        :active true))]
    (for [table tables]
      (assoc table :fields (get table-id->fields (:id table))))))

(t2/hydrate (t2/select-one :model/Table) :fields)
;; => (toucan2.instance/instance
;;     :model/Table
;;     {:name "v_users",
;;      :fields
;;      [(toucan2.instance/instance
;;        :model/Field
;;        {:name "user_id", ...})
;;       ...]})

(t2/hydrate (t2/select :model/Table {:limit 2}) :fields)
;; => [(toucan2.instance/instance
;;      :model/Table
;;      {:name "v_users",
;;       :fields
;;       [(toucan2.instance/instance
;;         :model/Field
;;         {:name "user_id", ...})
;;        ...]})
;;     (toucan2.instance/instance
;;      :model/Table
;;      {:name "v_log",
;;       :fields
;;       [(toucan2.instance/instance
;;         :model/Field
;;         {:name "user_id", ...})
;;        ...]})]

```

qnkhuat:
closes https://github.com/metabase/metabase/issues/38993

I didn't convert all because some hydration is easier to write as a simple hydrations.

I also have [Detect N+1 hydration and throw an exception](https://github.com/metabase/metabase/pull/40272) to make sure we don't have N+1 with the remaining simple hydrations.

### DIFF
--- a/src/metabase/models/card.clj
+++ b/src/metabase/models/card.clj
@@ -128,37 +128,36 @@
 
 (methodical/defmethod t2/batched-hydrate [:model/Card :dashboard_count]
   [_model k cards]
-  (mi/common-batched-hydration
-   k
-   cards
+  (mi/instances-with-hydrated-data
+   cards k
    #(->> (t2/query {:select    [[:%count.* :count] :card.id]
                     :from      [[:report_dashboardcard :dashcard]]
                     :left-join [[:report_card :card] [:= :dashcard.card_id :card.id]]
                     :where     [:in :card.id (map :id cards)]
                     :group-by  [:card.id]})
          (map (juxt :id :count))
-         (into (zipmap (map :id cards) (repeat 0))))
-   :id))
+         (into {}))
+   :id
+   {:default 0}))
 
 (methodical/defmethod t2/batched-hydrate [:model/Card :parameter_usage_count]
   [_model k cards]
-  (mi/common-batched-hydration
-   k
-   cards
+  (mi/instances-with-hydrated-data
+   cards k
    #(->> (t2/query {:select    [[:%count.* :count] :card.id]
                     :from      [[:parameter_card :pc]]
                     :left-join [[:report_card :card] [:= :pc.card_id :card.id]]
                     :where     [:in :card.id (map :id cards)]
                     :group-by  [:card.id]})
          (map (juxt :id :count))
-         (into (zipmap (map :id cards) (repeat 0))))
-   :id))
+         (into {}))
+   :id
+   {:default 0}))
 
 (methodical/defmethod t2/batched-hydrate [:model/Card :average_query_time]
   [_model k cards]
-  (mi/common-batched-hydration
-   k
-   cards
+  (mi/instances-with-hydrated-data
+   cards k
    #(->> (t2/query {:select [[:%avg.running_time :running_time] :card_id]
                     :from   [:query_execution]
                     :where  [:and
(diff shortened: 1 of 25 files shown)


src/metabase/models/card.clj
```diff
@@ -126,43 +126,63 @@
...
-      first vals first))
+(methodical/defmethod t2/batched-hydrate [:model/Card :dashboard_count]
+  [_model k cards]
+  (mi/instances-with-hydrated-data
+   cards k
+   #(->> (t2/query {:select    [[:%count.* :count] :card.id]
+                    :from      [[:report_dashboardcard :dashcard]]
+                    :left-join [[:report_card :card] [:= :dashcard.card_id :card.id]]
+                    :where     [:in :card.id (map :id cards)]
+                    :group-by  [:card.id]})
```
piranha, line 137: Why join when you can `group by report_dashboardcard.card_id`? Seems like this will lead to smaller and simpler query.

src/metabase/models/interface.clj
```diff
@@ -666,3 +667,26 @@
...
+      :tables
+      #(t2/select-fn->fn :db_id identity :model/Table)
+      :id)
+    ;; => [{:id 1 :tables [...tables-from-db-1]}
+           {:id 2 :tables [...tables-from-db-2]}]
+
+  - key->hydrated-items-fn: is a function that returns a map with key is `instance-key` and value is the hydrated data of that instance."
+  [instances                      :- [:sequential :any]
+   hydration-key                  :- :keyword
+   instance-key->hydrated-data-fn :- fn?
```
piranha, line 686: Hmmm, why function instead of just data?

src/metabase/models/interface.clj
```diff
@@ -666,3 +667,26 @@
...
+   instance-key                   :- :keyword
+   & [{:keys [default] :as _options}]]
+  (when (seq instances)
+    (let [key->hydrated-items (instance-key->hydrated-data-fn)]
+      (for [item instances]
```
piranha, line 691: Should this maybe return a vector? I mean no real reason for laziness here, `instances` is not going to be lazy.

src/metabase/models/database.clj
```diff
@@ -306,12 +306,26 @@
...
+  "Batch hydrate `Tables` for the given `Database`."
+  [_model k databases]
+  (mi/instances-with-hydrated-data
+   databases k
+   #(group-by :db_id
+              ;; TODO - do we want to include tables that should be `:hidden`?
+              (t2/select :model/Table
+                         :db_id  [:in (map :id databases)]
+                         :active true
+                         {:order-by [[:db_id :asc][:%lower.display_name :asc]]}))
```
piranha, line 326: this line could use a space or a newline

sansolovyov (approved):
Aside from a few nitpicks I think it's great!

src/metabase/models/interface.clj
qnkhuat: yea we can, but what are the benefits of doing so?

src/metabase/models/interface.clj
qnkhuat: yep, save some DB calls. or else we'll have to wrap a `(when (seq instances) (t2/select....))` all the time, which is kinda boilterplate IMO.

src/metabase/models/card.clj
qnkhuat: right, nice catch!

src/metabase/models/interface.clj
piranha: Mostly simpler stacktraces 😁

Issue #39246, opened 2024-02-28
[Epic] Let white-labeling customers customize illustrations


WiNloSt:
**Links**
- [Product doc](https://www.notion.so/metabase/Let-white-labeling-customers-customize-illustrations-24e120498d5f414aaf52ddd0de9e20fc)
- [Eng doc](https://www.notion.so/metabase/Tech-Let-white-labeling-customers-customize-illustrations-e8adba727d0e4297b2ed8679ab79a8d6)
- [Testing plan](https://www.notion.so/metabase/Testing-plan-Let-white-labeling-customers-customize-illustrations-3985c2369ca647a19f5f454e390a9bca)
- integration branch: `39246-milestone-1`

```[tasklist]
#### Milestone 1
- [ ] https://github.com/metabase/metabase/pull/39299
- [ ] https://github.com/metabase/metabase/pull/40487
- [ ] https://github.com/metabase/metabase/pull/40491
- [ ] https://github.com/metabase/metabase/pull/40139
- [ ] https://github.com/metabase/metabase/pull/40710
- [x] https://github.com/metabase/metabase/pull/40219 - integration branch
```

```[tasklist]
#### Milestone 2
- [x] ~Connect to the BE referencing files~ We decided [not to do this](https://metaboat.slack.com/archives/C063Q3F1HPF/p1709737723369799)
- [ ] https://github.com/metabase/metabase/pull/40681
- [ ] Ping Thomas once analytics is merged to master
```

PR #40139, opened 2024-03-14
4. Connect illustrations to new setting values

WiNloSt:
Epic: #39246

### Description

Connect the rest 3 illustration settings to the UI
1. Landing page (existing)
1. No question results
1. No search results

### How to verify
Change the 4 illustration settings to any values and visit
1. Login page
2. Landing page
3. Unsubscribe page `/unsubscribe?hash=hash&email=email&pulse-id=pulse-id`
4. Question without results
5. Questions without results as a card in a dashboard
6. `+ New` > Dashboard > Click the collection > you'll see the new collection picker
7. From the collection picker search anything that would return nothing
8. From the collection picker go to an empty collection
9. search in the app for anything that would return nothing

### Demo
Screenshot 2024-03-14 at 10 45 39 PM

### Checklist

- [x] Tests have been added/updated to cover changes in this PR

PR #40219, opened 2024-03-18
[M1] Metabase illustration settings

WiNloSt:
Milestone 1 branch of #39246
[Product doc](https://www.notion.so/metabase/Let-white-labeling-customers-customize-illustrations-24e120498d5f414aaf52ddd0de9e20fc)
[Figma design](https://www.figma.com/file/yOksAHQ5ih6ospZd6f3RVc/Let-white-labeling-customers-customize-illustrations?type=design&node-id=1486-122158&mode=design&t=sKXiTjXPk4lqncYe-0)

This integration PR contains 5 other PRs for milestone 1 which are all approved and merged to this branch.

There are some merge conflict fixes that are introduced directly to this branch, but they're mostly global CSS migration conflicts.

### How to verify
We add 4 new illustration settings here.
1. Login and unsubscribe pages
    - Test login page with different illustration options
    - Test unsubscribe page with different illustration options. You can access the unsubscribe page here http://localhost:3000/unsubscribe?hash=hash&email=email&pulse-id=pulse-id
1. Landing page (aka home page)
    - Test home page with different illustration options
1. When calculations return no results. This is for query results.
    - Question that returns no results
    - A card in a dashboard that returns no results.
1. When no objects can be found. This is for anything that's not query results e.g. any entity that couldn't be found
    - Search app
    - Collection picker -> Search
    - Collection picker -> select an empty collection


WiNloSt:
* **#40139** <a href="https://app.graphite.dev/github/pr/metabase/metabase/40139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#40219** <a href="https://app.graphite.dev/github/pr/metabase/metabase/40219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @WiNloSt and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

mazameli (changes requested):
- Custom Login and landing page images: this seems to be scaling the image even if I choose a really wide one, which is different behavior from how the lighthouse default illustration works. In general for these large images I don't think we want to be scaling images, since it'll make a small image look bad, and defeat the purpose of a large image that overflows the viewport. Example: this image of mine is much taller than this: 

image

- Nit: when changing the no-results images to a custom one or to no-image, I found I had to hard refresh to see that take effect. Must've been cached.

mazameli (approved):
Removing my change request because I misunderstood what @vbenedetti had intended here. I think the image behavior in the branch is neither right nor wrong — it could be exactly what our users want. It's just different from the default image's behavior.

WiNloSt:
> Nit: when changing the no-results images to a custom one or to no-image, I found I had to hard refresh to see that take effect. Must've been cached.

I couldn't reproduce this after updating the branch with the latest `master`.

I've tried both no calculation results and no object results and made sure I click exit admin, so the page won't reload and I could still see those image settings reflect on the app.

PR #40230, opened 2024-03-18
Fix mysql + liquibase flakes

### LINKED ISSUE #37884: Fix liquibase rollbacks causing flaky tests on MySQL
This test flake was caused by flakiness of Liquibase rollbacks in tests for MySQL: https://github.com/metabase/metabase/issues/37434

Ngoc's comment adds more details https://github.com/metabase/metabase/issues/37434#issuecomment-1894930584

There have been other tests where we had to comment out the rollback as well, e.g. `migrate-field-database-type-test`.

sansolovyov:
Resolves #37884


qnkhuat (approved):
it seems like we can't no longer reproduce this anymore, so I'm okay with uncommenting those tests and see how it goes on CI.

PR #40233, opened 2024-03-18
Revert "Revert updating FKs during metadata sync (#39679) (#39924)"

calherries:
My hunch was wrong that reverting https://github.com/metabase/metabase/pull/39679 would solve the flake, so this PR reverts the revert I made in https://github.com/metabase/metabase/pull/39924.

This is the flake fix https://github.com/metabase/metabase/pull/40291


src/metabase/sync/sync_metadata/fks.clj
```diff
@@ -51,7 +51,10 @@
                       ;; - fk_target_field_id is NULL and the new target is not NULL
                       ;; - fk_target_field_id is not NULL but the new target is different and not NULL
                       [pk-field-id-query :pk]
-                      [:= :f.fk_target_field_id nil]]
+                      [:and
```
crisptrutski, line 54: Is this :and required? Seems weird

crisptrutski (approved):

src/metabase/sync/sync_metadata/fks.clj
calherries: definitely not, thanks 9472594a89aadfaa1c58abb2dd276674bfc9192a

PR #40235, opened 2024-03-18
Upload CSVs: Rename boolean-or-int to boolean-int

calherries:
This PR comes from a moment of confusion where I forgot whether `boolean-or-int` includes true and false, and I had to check.

The reason for the confusion is the name: `boolean-or-int` could be interpreted to mean "the union of boolean and integer values". But actually the meaning is "the intersection of boolean and integer values".

I had the same confusion when I saw the `float-or-int` type in https://github.com/metabase/metabase/pull/39724.

This PR proposes a new name: `boolean-int`. 

It is more precise, unambiguous, and fits the intuition from English. In English, "true" is definitely not a "boolean integer", but "0" is.

### DIFF
--- a/src/metabase/sync/analyze/classifiers/name.clj
+++ b/src/metabase/sync/analyze/classifiers/name.clj
@@ -11,7 +11,7 @@
    [metabase.util.malli :as mu]
    [metabase.util.malli.schema :as ms]))
 
-(def ^:private bool-or-int-type #{:type/Boolean :type/Integer})
+(def ^:private bool-int-type #{:type/Boolean :type/Integer})
 (def ^:private float-type       #{:type/Float})
 (def ^:private int-type         #{:type/Integer})
 (def ^:private int-or-text-type #{:type/Integer :type/Text})
@@ -46,7 +46,7 @@
    [#"^.*_latitude$"               float-type       :type/Latitude]
    [#"^.*_type$"                   int-or-text-type :type/Category]
    [#"^.*_url$"                    text-type        :type/URL]
-   [#"^active$"                    bool-or-int-type :type/Category]
+   [#"^active$"                    bool-int-type    :type/Category]
    [#"^city$"                      text-type        :type/City]
    [#"^country"                    text-type        :type/Country]
    [#"_country$"                   text-type        :type/Country]
--- a/src/metabase/upload.clj
+++ b/src/metabase/upload.clj
@@ -66,13 +66,13 @@
 ;;     |  /     \
 ;;     | /       \
 ;;     |/         \
-;; boolean-or-int  auto-incrementing-int-pk
+;; boolean-int  auto-incrementing-int-pk
 ;;
-;; `boolean-or-int` is a special type with two parents, where we parse it as a boolean if the whole
-;; column's values are of that type. additionally a column cannot have a boolean-or-int type, but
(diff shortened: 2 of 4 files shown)


crisptrutski:
I've actually had the same brain fart, so I sympathize. I'm totally open to renaming.

While I agree that the naming is ambiguous, I disagree that it's imprecise. It has a crisp interpretation, which is intuitive when looking at it from the right perspective.

1. Every raw value must be assigned to some specific Value Type
2. Certain raw values may support parsing as any number multiple Column Types.
3. Eventually we must choose exactly one Column Type to parse the value as.

In this sense the "or" is the natural joining clause for the set in 2, corresponding to an "exclusive or" for selection that must be made in 3.

---

The sense you explain thinking of "true" as as a being the union ("inclusive or") from the dual perspective of sets is another way of thinking of things. In programming `or` always corresponds to "inclusive or", so it makes some sense to come from that angle, but I'd argue that it's less natural, intuitive and less general to do so.

To get into it I should quickly lay out how I think about the type correspondences.

(Update: it was not at all quick to lay this out in the end)

It can be natural to identify types with the set of their members. For example the Column Types can be identified with set of raw values that can parse to it, i.e. the set of values its parser won't throw an error for. 

*The fact that each column type can be **uniquely** identified by this set of raw values currently is nice, but there's no fundamental reason this holds for our domain - one can imagine distinct database column types which correspond to the exact same set of of  values, e.g. JSON and BSON.*

There's an interesting correspondence `C` we can create between most value types `V` and each concrete type `c`, where `R` is the set of all raw values. This correspondence

1. $`\left\{c(V)\right\} = \bigcap_{v \in V} C(v)`$
2. $`V = \bigcup_{v \in R} c(V) \in C(v)`$

Obviously the correspondence only exists where the RHS of 1 is a singleton, but a fascinating property right now is whenever that is true, the reflexive condition in 2 is always satisfied.

Almost every Value Type has this correspondence, and the vast majority of raw value have such a Value Type. We call all such Value Types "concrete", and all other types are "abstract". Currently we have only 1 such abstract type in master, the one you're renaming, and then the closely related `float-or-int` being discussed in a PR.

Phew, that was harder to explain than I expected. Just the way I've been thinking about things intuitively 🤷 

Both of these "abstract types" have members which correspond to the intersection of two "concrete types". While it was a bit formal to get here, I think this framing is where your intuition comes from - an `or` typically corresponds to a *union* rather an intersection, and so this set would instead include values like `true` and `2`. 

Given this outlook, you could imagine rather calling these types `boolean-and-int` or `float-and-int`. This is similarly precise, and from a certain angle more intuitive.

I hold out that it's natural to think of other abstract types that require other set operations, like the set `explicit-int` being the complement of `float-or-int` within `int`. I'll have more to say on that on the other PR thread 😄 

When in a certain frame of mind, I find this more intuitive, but when dealing with the code I'm typically dealing with the values, and argue it is less natural.

I don't like using this kind of set intuition for the following reasons:

1. Identifying column types with their raw or logical values could break down.
2. The neat 1-1 correspondence between value and column types could break down.
3. The interesting reflexive condition could break down.
4. We could find ourselves dealing with more sets that aren't just intersections.

---

Regarding the proposal to simply strip out the conjunction, I think that's the worst of all worlds. One needs to guess at the relationship between the nouns, and one needs to also assume that it's the only relation we'll every have. It could also be ambiguous with column types that themselves are compound.

PR #40236, opened 2024-03-18
[E2E] Refactor question commands to function helpers

kamilmielnik:
### Description

This PR marks the beginning of an effort to use function helpers instead of custom Cypress commands.
By migrating custom Cypress commands to function helpers we will be able to organize the helpers in any way we want, particularly in a way that will improve their discoverability (which affects the learning curve and convenience when writing tests).

----

While it's not clear how we would organize these helpers yet, we know that it'd be good to:
- make UI helpers (functions that "click" through the UI) clearly distinguishable from API helpers (functions that call API endpoints)
- have helpers related to particular entities live closely
- have helpers related to particular pages live closely

----

There's one more positive consequence (see line 234 in `notebook-data-source.cy.spec.ts`).
It's sometimes convenient to define objects (e.g. `questionDetails`) using types defined for arguments of custom commands/helper functions. When working with custom commands, getting references to those types is cumbersome, we can either:
- `const questionDetails: Parameters<typeof cy.createQuestion>[0];`
- import the type by referencing the file with custom command (`import { QuestionDetails } from "e2e/support/commands/question"`)
- or define the type for command argument somewhere else (no good place for it; type for function argument should live next to the function anyway)

But once we turn a custom command into a function helper, it feels natural to import the type along with the function (`import { createQuestion, QuestionDetails } from "e2e/support/helpers"`), similarly to how we do it with component props sometimes.

----

This PR:
- Refactors `cy.createQuestion`, `cy.createNativeQuestion` and `cy.archiveQuestion` to function helpers
- Deprecates `cy.createQuestion` and `cy.createNativeQuestion` commands
- Replaces `cy.createQuestion` usages in TS files with `createQuestion`
- Replaces the only `cy.archiveQuestion` usage with `archiveQuestion`
- Removes `cy.archiveQuestion`


e2e/support/helpers/e2e-question-helpers.ts
```diff
@@ -0,0 +1,226 @@
+import { SAMPLE_DB_ID } from "e2e/support/cypress_data";
```
kamilmielnik, line 1: Things in this file have been moved from `e2e/support/commands/api/question.ts`, implementations/interfaces have not been changed.

e2e/test/scenarios/question/notebook-data-source.cy.spec.ts
```diff
@@ -266,7 +269,7 @@ describe("scenarios > notebook > data source", () => {
       });
 
       cy.findByTestId("select-list")
-        .findByLabelText(modelDetails.name)
+        .findByLabelText(checkNotNull(modelDetails.name))
```
kamilmielnik, line 272: TS complained that `modelDetails.name` can be `undefined`

ranquild (approved):

nemanjaglumac (approved):
🚀

PR #40248, opened 2024-03-18
adding a new section "How x-ray works"

abdhabli:
Description:
This PR introduces a succinct "How X-rays Work" section to the Metabase documentation, responding to customer inquiries about understanding how X-rays work.

Project link: https://www.notion.so/metabase/project-Documenting-X-rays-and-Semantic-Types-02ff4b7b5e0d4603ab262cf4664844ae?pvs=4


alexyarosh (approved):
I think it's good to have a little more explanation than we have now. 

But I hope we'll keep expanding this doc in the future, because even with this change, it's kind of unsatisfactory: this section doesn't really explain how xrays work, it just says what the xrays _are_ and that you get those results by magic. It could be a good idea to explain the magic at some point.

abdhabli:
Thanks, Jeff and Alex!

Issue #40005, opened 2024-03-12
[Epic] Initial homepage experience for embedding admins


npretto:
**Links**
- product doc: [Initial homepage experience for embedding admins](https://www.notion.so/metabase/Initial-homepage-experience-for-embedding-admins-9431f50fdf3a4f63b291dc801349c335)
- eng doc: [Initial homepage experience for embedding admins](https://www.notion.so/metabase/Initial-homepage-experience-for-embedding-admins-was-Meet-embedders-at-the-door-full-version-63fa3a8e07634c9faf67d5bdd5a2df08)
- feature branch: `embed-homepage` / https://github.com/metabase/metabase/pull/40382

**Implementation Plan**

MS1: https://github.com/metabase/metabase/pull/40382

```[tasklist]
### MS 1 - Homepage without example dashboard -`embed-homepage`
- [ ] https://github.com/metabase/metabase/pull/40251
- [ ] https://github.com/metabase/metabase/pull/40455
- [ ] https://github.com/metabase/metabase/pull/40528
- [ ] https://github.com/metabase/metabase/pull/40587
- [x] double check unit tests and e2es (I'm actually writing them as I go)
- [ ] https://github.com/metabase/metabase/pull/40638
- [ ] https://github.com/metabase/metabase/pull/40813
```

MS2: https://github.com/metabase/metabase/pull/41180

```[tasklist]
### MS 2 - Add example dashboard reference
- [x] dependency on https://github.com/metabase/metabase/issues/40066
- [ ] https://github.com/metabase/metabase/pull/41138
```

```[tasklist]
### MS 3 - Dismiss with feedback - `embed-homepage-ms3`
- [x] dependency on https://github.com/metabase/harbormaster/issues/4783
- [ ] https://github.com/metabase/harbormaster/issues/4814
- [ ] https://github.com/metabase/metabase/pull/40746
- [ ] https://github.com/metabase/metabase/pull/40974
```

MS4: https://github.com/metabase/metabase/pull/41990
```[tasklist]
### MS 4 - Analytics + polishment - `embed-homepage-milestone4`
- [ ] https://github.com/metabase/metabase/pull/41575
- [ ] https://github.com/metabase/metabase/pull/41638
- [ ] https://github.com/metabase/metabase/pull/41644
- [ ] https://github.com/metabase/metabase/pull/41725
- [ ] https://github.com/metabase/metabase/pull/41784
- [ ] https://github.com/metabase/metabase/pull/42042
[...]

PR #40251, opened 2024-03-18
create ui for the full embedding homepage, without any logic

npretto:
Part of  [[Epic] Initial homepage experience for embedding admins](https://github.com/metabase/metabase/issues/40005)

### Description

This only adds the "dumb" UI, all the logic and tests will come in separate PRs

Note that I'm currently importing `List` from `mantine`, I'll open a PR to re-export it from `metabase/ui` (see [slack](https://metaboat.slack.com/archives/C057WD5L0JG/p1710772511289899)) and to also fix the numbers not showing up without the inline style.

### How to verify

- `yarn storybook`
- open http://localhost:6006/?path=/story/features-embedhomepage--default
- play around with the controls

### DIFF
--- a/frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.stories.tsx
+++ b/frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.stories.tsx
@@ -41,7 +41,7 @@ export const Default: Story = {
       "https://www.metabase.com/docs/latest/embedding/interactive-embedding-quick-start-guide.html",
     embeddingDocsUrl:
       "https://www.metabase.com/docs/latest/embedding/start.html",
-    customerFacingAnalyticsDocsUrl:
+    analyticsDocsUrl:
       "https://www.metabase.com/learn/customer-facing-analytics/",
   },
 };
--- a/frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.stories.tsx
+++ b/frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.stories.tsx
@@ -1,0 +1,47 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { EmbedHomepageView } from "./EmbedHomepageView";
+
+type Args = Omit<
+  React.ComponentProps<typeof EmbedHomepageView>,
+  "exampleDashboardId"
+> & {
+  hasExampleDashboard: boolean;
+};
+
+const meta: Meta<Args> = {
+  title: "FEATURES/EmbedHomepage",
+  component: EmbedHomepageView,
+  parameters: {
+    controls: {
+      exclude: "exampleDashboardId",
+    },
+  },
+};
+export default meta;
+
+type Story = StoryObj<Args>;
+
+export const Default: Story = {
+  render: args => {
+    return (
+      <EmbedHomepageView
+        {...args}
+        exampleDashboardId={args.hasExampleDashboard ? 1 : undefined}
+        key={args.plan}
+      />
+    );
+  },
+  args: {
+    embeddingAutoEnabled: true,
+    hasExampleDashboard: true,
+    licenseActiveAtSetup: true,
+    plan: "oss-starter",
+    interactiveEmbeddingQuickstartUrl:
+      "https://www.metabase.com/docs/latest/embedding/interactive-embedding-quick-start-guide.html",
+    embeddingDocsUrl:
+      "https://www.metabase.com/docs/latest/embedding/start.html",
+    customerFacingAnalyticsDocsUrl:
+      "https://www.metabase.com/learn/customer-facing-analytics/",
+  },
+};
--- a/frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.tsx
+++ b/frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.tsx
@@ -1,0 +1,42 @@
+import { useSelector } from "metabase/lib/redux";
+import { getDocsUrl } from "metabase/selectors/settings";
+
+import { EmbedHomepageView } from "./EmbedHomepageView";
+
+export const EmbedHomepage = () => {
+  const interactiveEmbeddingQuickStartUrl = useSelector(state =>
+    // eslint-disable-next-line no-unconditional-metabase-links-render -- only visible to admins
+    getDocsUrl(state, {
+      page: "embedding/interactive-embedding-quick-start-guide",
+    }),
+  );
+  const embeddingDocsUrl = useSelector(state =>
+    // eslint-disable-next-line no-unconditional-metabase-links-render -- only visible to admins
+    getDocsUrl(state, { page: "embedding/start" }),
+  );
+
+  const learnMoreInteractiveEmbedding = useSelector(state =>
+    // eslint-disable-next-line no-unconditional-metabase-links-render -- this is only visible to admins
+    getDocsUrl(state, { page: "embedding/interactive-embedding" }),
+  );
+
+  const learnMoreStaticEmbedding = useSelector(state =>
+    // eslint-disable-next-line no-unconditional-metabase-links-render -- this is only visible to admins
+    getDocsUrl(state, { page: "embedding/static-embedding" }),
+  );
+
+  return (
+    <EmbedHomepageView
+      exampleDashboardId={1}
+      embeddingAutoEnabled={false}
+      licenseActiveAtSetup={true}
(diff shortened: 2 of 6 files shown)


frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import type { Meta, StoryObj } from "@storybook/react";
```
npretto, line 1: This may or may not land in master ([slack](https://metaboat.slack.com/archives/C505ZNNH4/p1710781977025789)), but for now it's helpful to test all the variants of the UI.

frontend/src/metabase/home/components/EmbedHomepage/StaticTabContent.tsx
```diff
@@ -0,0 +1,65 @@
+// eslint-disable-next-line no-restricted-imports
+import { List } from "@mantine/core";
```
npretto, line 2: I'll move this to mantine/ui, see description of PR

frontend/src/metabase/home/components/EmbedHomepage/StaticTabContent.tsx
```diff
@@ -0,0 +1,65 @@
...
+        lh="1.5"
+      >{t`Use static embedding to present data that applies to all of your tenants at once without ad hoc query access to their data. Available in all plans.`}</Text>
+
+      <Text fw="bold" mb="md">{t`The TL;DR:`}</Text>
+
+      <List
+        type="ordered"
+        mb="lg"
+        size="sm"
+        // TODO: fix this in a better way
```
npretto, line 27: I'll also fix this (I hope?) when re-exporting List from metabase/ui

frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.stories.tsx
WiNloSt: I have no objection for having this in the repo if it helps with feature development.

frontend/src/metabase/home/components/EmbedHomepage/InteractiveTabContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <>
+      <Text
+        mb="md"
+        lh="1.5"
+      >{t`Use interactive embedding to offer multi-tenant, self-service analytics and dashboard creation in their own data sandbox. Pro and Enterprise plans only. `}</Text>
+
+      <Text fw="bold" mb="md">{t`The TL;DR:`}</Text>
+
+      <List
```
WiNloSt, line 28: What are the reasons/benefits for using `List` here? I mean I understand Mantine has it, but this doesn't look anything different from using simple `ol` and `li` tags. If we need to use icons then maybe I could understand we might need `List` but that doesn't seem to be the case here.

frontend/src/metabase/home/components/EmbedHomepage/InteractiveTabContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+import type { EmbedHomepageViewProps } from "./EmbedHomepageView";
+
+export const InteractiveTabContent = (props: EmbedHomepageViewProps) => {
+  const {
+    embeddingAutoEnabled,
+    interactiveEmbeddingQuickstartUrl,
+    exampleDashboardId,
+    licenseActiveAtSetup,
+    learnMoreInteractiveEmbedUrl,
+  } = props;
```
WiNloSt, line 18: If we don't need to spread props like in `frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepageView.tsx` what do you think about destructuring these props in the function arguments instead for consistency?

frontend/src/metabase/home/components/EmbedHomepage/InteractiveTabContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <Text fw="bold" mb="md">{t`The TL;DR:`}</Text>
+
+      <List
+        type="ordered"
+        mb="lg"
+        size="sm"
+        // TODO: fix this in a better way
+        style={{ listStyleType: "decimal" }}
+      >
+        {licenseActiveAtSetup === false && (
```
WiNloSt, line 35: I feel like booleans that could be undefined would be a bit awkward when the default value should be true. I'm not sure if that's the case, but the structure suggests so.

The reason sometimes we have `someBoolean === false` is to ensure `undefined` doesn't satisfy this clause. I feel like making the boolean that would have the default value as false would feel more natural. Since you can just do

```suggestion
        {!licenseActiveAtSetup && (
```

That being said, I guess from the name, this value should be `false` unless users activate their license, so I don't see the reason why we need to use `xxx === false`

WiNloSt (approved):
Looks good with some minor comments.

frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.tsx
```diff
@@ -0,0 +1,42 @@
...
+    // eslint-disable-next-line no-unconditional-metabase-links-render -- this is only visible to admins
+    getDocsUrl(state, { page: "embedding/static-embedding" }),
+  );
+
+  return (
+    <EmbedHomepageView
+      exampleDashboardId={1}
+      embeddingAutoEnabled={false}
+      licenseActiveAtSetup={true}
+      plan="oss-starter"
```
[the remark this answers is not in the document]
npretto: That prop will be calculated in this component the same way as the urls, so it's kind of a "private prop".
It should make more sense in the following PRs, I should have left a comment that the values were temporarily hardcoded though

frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.tsx
```diff
@@ -0,0 +1,42 @@
...
+      interactiveEmbeddingQuickstartUrl={interactiveEmbeddingQuickStartUrl}
+      embeddingDocsUrl={embeddingDocsUrl}
+      // eslint-disable-next-line no-unconditional-metabase-links-render -- only visible to admins
+      customerFacingAnalyticsDocsUrl="https://www.metabase.com/learn/customer-facing-analytics/"
```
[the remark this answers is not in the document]
WiNloSt: sounds good to me.

frontend/src/metabase/home/components/EmbedHomepage/InteractiveTabContent.tsx
npretto: Yeah itt felt overkill to me too but right now we have a [reset that removes the list type](https://github.com/metabase/metabase/blob/49586158e6c6d607f89d410c5d5af2aaea4fa6e2/frontend/src/metabase/css/core/base.module.css#L44-L49), so we'd have to make a `styled(ol)`/css class with
```css
  list-style-type: decimal;
  list-style-position: inside;
 ```
I did it for the minimal embedding card but it felt like it should have been a re-usable component.

It should also help with making sure spacing, font sizes etc are consistent

frontend/src/metabase/home/components/EmbedHomepage/InteractiveTabContent.tsx
WiNloSt: to elaborate on what I meant by the default value as false. That covers the case for `null` and `undefined` too. I should have said falsy value by default.

It would feel more natural if you pass `undefined`, `null`, or `false` and gets the same result. Because It gets awkward sometimes if you pass `undefined` and `false` and get different results.

npretto: Ah I see, yeah that makes sense, I'll make sure not to render this when the any of the flags are undefined/null as it means they did the setup before this was introduced

frontend/src/metabase/home/components/EmbedHomepage/EmbedHomepage.tsx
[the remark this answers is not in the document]
npretto: 25fb1b4ed3798d365bbe82968597a85c5f880b32 ✅

frontend/src/metabase/home/components/EmbedHomepage/InteractiveTabContent.tsx
npretto: 19d24864a269e9be5fcb51da58c772a21976d1ee ✅

PR #40253, opened 2024-03-18
Fix time-interval-expression-test flake

snoe:
Updates test to handle the problem encountered in #40238

> 
> time-interval-expression-test has started to fail consistently on master and the 49 branch.
> 
> The test is flawed and only worked by coincidence when the test was added in February.
> 
> It uses checkins:1-per-day, which has this docstring:
> "Dynamically generated dataset with 30 checkins spaced 24 hours apart, from 15 days ago to 14 days in the future."
> 
> And then goes on to test a query that has filters with (lib/time-interval ... :current :quarter).
> 
> The problem is the current quarter, which ends on March 31, is 13 days in the future.
> 
> I've commented out the test for now, and will leave fixing the test to the QPD team.

### DIFF
--- a/test/metabase/query_processor_test/date_bucketing_test.clj
+++ b/test/metabase/query_processor_test/date_bucketing_test.clj
@@ -1035,8 +1035,7 @@
                     {:aggregation [[:count]]
                      :filter      [:time-interval $timestamp :last :week]})))))))))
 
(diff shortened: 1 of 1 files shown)


metamben (approved):

WiNloSt:

PR #40255, opened 2024-03-18
Update athena.md

alinerguio:
Missing permissions in documentation (AWS Athena)


paoliniluis:
Those permissions are already included

Issue #40254, opened 2024-03-18
[Epic] Show generated SQL in a side bar automatically


nemanjaglumac:
**Links**
- [Product doc](https://www.notion.so/metabase/Show-generated-SQL-in-a-side-bar-automatically-b56dbea3ece942ac981ee0326e5867c8)
- [Eng doc](https://www.notion.so/metabase/Tech-Show-generated-SQL-in-a-side-bar-automatically-febee0259d2d40e294a06d2f4930122e?pvs=4)
- Feature branch:  check under each of the milestones
- [Figma](https://www.figma.com/file/TZseFDIOAhNU6ArKtFmxBx/Show-generated-SQL-in-a-side-bar-automatically?type=design&node-id=1-286&mode=design&t=ptjbABjX6rJIqwRt-0)
- Testing plan: https://github.com/metabase/metabase/issues/40579

# Implementation Plan

## Milestone 0
Good litmus test for the scope of this milestone is that all steps in it should easily be merged into `master` and even backported.
```[tasklist]
## Prep and necessary cleanup
- [ ] #40136
- [ ] https://github.com/metabase/metabase/pull/40268
- [ ] https://github.com/metabase/metabase/pull/40286
- [ ] https://github.com/metabase/metabase/pull/40316
```

## Milestone 1
For the proof of concept, our goal is to have a SQL preview that automatically updates with each new step being added/removed in the Notebook GUI editor. Code highlighting, line numbering, and the ability to resize the sidebar are not important for this phase.

Feature branch: https://github.com/metabase/metabase/tree/sql-sidebar-preview-poc
```[tasklist]
## Working Prototype
- [x] Show SQL preview in the sidebar and update synchronously with steps in the notebook GUI editor (#40279)
- [x] Hook the sidebar to the toggle query button and preserve the preference in the Redux store `uiState`  (#40327)
- [x] Convert to the native query (#40362)
- [x] Make minimal necessary CSS tweaks (#40463)
- [x] Adjust the sidebar to work with small screens (#40463)
- [x] Cover with E2E tests before merging (https://github.com/metabase/metabase/pull/40540)
```

```[tasklist]
## Extras
- [x] #40430
```

## Milestone 2
Tasks that should happen after we gather some initial feedback on PoC from the product, design, and from the users.
```[tasklist]
## Visual tweaks
- [ ] https://github.com/metabase/metabase/issues/40668
- [ ] https://github.com/metabase/metabase/issues/40697
```

```[tasklist]
[...]

PR #40268, opened 2024-03-18
SQL in a sidebar prep: `NativeQueryPreview` component

nemanjaglumac:
This is a step in the Milestone 0 of #40254
This PR is mostly just about renaming.

Before:
A component called `NativeQueryModal` was used in two other components (modals)
- `PreviewQueryModal` and
- `ConvertQueryModal`

After:
`NativeQueryModal` is renamed to `NativeQueryPreview` because it conveys what the component is doing more truthfully, and it gives us some flexibility in the future if we want to use this component in anything that is not a modal.

I've used the opportunity to do some cleanup while working on this. Two most notable changes:
1. named exports are now used where possible
2. Mantine is now used instead of styled components

### Testing
Everything should still work and look the same as before. The end user shouldn't notice any difference, and all tests should pass.


frontend/src/metabase/query_builder/components/view/NativeQueryPreview/NativeQueryPreview.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <Title>{error ? t`An error occurred in your query` : title}</Title>
+        <CloseButton onClose={onClose} />
+      </Header>
+      {error && <Divider />}
+      <Flex
+        direction="column"
+        justify={isLoading ? "center" : undefined}
+        align={isLoading ? "center" : undefined}
+        mih={0}
+        style={{ flex: "1 1 auto" }}
```
ranquild, line 93: there are flex properties available as props

ranquild (approved):

PR #40274, opened 2024-03-19
Fix dnd clause composition

ranquild:
Context https://metaboat.slack.com/archives/C0645JP1W81/p1710769889848229

Trying to select text in the expression popover for an existing clause won't start a drag-n-drop interaction.

### DIFF
--- a/e2e/test/scenarios/question/notebook.cy.spec.js
+++ b/e2e/test/scenarios/question/notebook.cy.spec.js
@@ -721,6 +721,26 @@ describe("scenarios > question > notebook", { tags: "@slow" }, () => {
         .should("have.text", name);
     }
 
+    function verifyPopoverDoesNotMoveElement({
+      type,
+      name,
+      index,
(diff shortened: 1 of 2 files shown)


kamilmielnik (changes requested):

kamilmielnik:
Please also add tests :pray:

ranquild:
@kamilmielnik how do you expect this to be tested?

kamilmielnik:
> @kamilmielnik how do you expect this to be tested?

An e2e test simulating text selection with `mousedown` + `mousemove` + `mouseup` in a way that would reorder the clauses (like in the loom video) and asserting that clauses have the same order afterwards.

ranquild:
@kamilmielnik 
before
Screenshot 2024-03-19 at 09 31 46

after
Screenshot 2024-03-19 at 09 32 06

kamilmielnik (approved):
:+1:

PR #40286, opened 2024-03-19
`createNativeDataset` API using RTK query

nemanjaglumac:
### What does this PR accomplish?

It adds a new way to `POST` data on the `/api/dataset/native` endpoint using RTK Query.


ranquild (approved):
I'd call the file just `dataset` because that's how the api file in clj is called

PR #40288, opened 2024-03-19
Fix: allow running native queries with required parameters and no defaults

### LINKED ISSUE #40250: Required variables broken in 0.49.0
### Describe the bug

Our devops upgraded our metabase last week to 0.49.0 and since then any report that has a required field (but no default) will not run!.  No matter what we set the value to in the input fields it *thinks* the field is not set and refuses to run the report.

We have to toggle the require property of the variable and then the report will run. 

### To Reproduce

1. Have a  SQL report with a required parameter variable
2. have that variable be required, but with no default value.
3. specify a value for the property
4. the "run" icon is NOT clickable (remains disabled)

### Expected behavior

To be able to run the report as in previous metabase versions.

### Logs

_No response_

```JSON
- browser - Firefox 123.0.1
- browser OS macOS 12
- DBs for queries (doesnt matter) redshfit & postgresql
- Metabase 0.49.0
- hosting AWS
- internal DB postgresql
```

### Severity

annoying and semi-blocking

### Additional context

_No response_

oleggromov:
Fixes https://github.com/metabase/metabase/issues/40250

Before, having a required parameter without a default value in a native query prevented both saving and running the query. However, the latter is incorrect — and was in fact a regression for one of the clients.

## How to check
1. create a native query `select * from people where state = {{ state }}`
2. set state to be required and don't set a default for it
3. make sure you cannot save it but can run it

Alternatively, you might add a `required` parameter to a saved native query's template tags to exactly reproduce the situation that the client has. But the behavior will be the same.

### DIFF
--- a/frontend/src/metabase-lib/v1/queries/NativeQuery.ts
+++ b/frontend/src/metabase-lib/v1/queries/NativeQuery.ts
@@ -119,10 +119,6 @@ export default class NativeQuery extends AtomicQuery {
     );
   }
 
-  canBeSaved(): boolean {
-    return this.canRun() && this._allRequiredTagsHaveDefaults();
-  }
-
   isEmpty() {
     return this._databaseId() == null || this.queryText().length === 0;
   }
@@ -309,18 +305,6 @@ export default class NativeQuery extends AtomicQuery {
   private _allTemplateTagsAreValid() {
     const tagErrors = this._validateTemplateTags();
     return tagErrors.length === 0;
-  }
-
-  private _allRequiredTagsHaveDefaults() {
-    const isEmpty = v => {
-      const value = Array.isArray(v) ? v[0] : v;
-      return value == null || value === "";
-    };
-
-    const requiredNoDefaults = this.templateTags().filter(
-      tag => tag.required && isEmpty(tag.default),
-    );
-    return requiredNoDefaults.length === 0;
   }
 
   setTemplateTag(name: string, tag: TemplateTag): NativeQuery {
--- a/frontend/src/metabase-lib/v1/queries/NativeQuery.ts
+++ b/frontend/src/metabase-lib/v1/queries/NativeQuery.ts
@@ -86,7 +86,6 @@ export function updateCardTemplateTagNames(
 
 // eslint-disable-next-line import/no-default-export -- deprecated usage
 export default class NativeQuery extends AtomicQuery {
-  // For Flow type completion
   _nativeDatasetQuery: NativeDatasetQuery;
 
   constructor(
@@ -118,6 +117,10 @@ export default class NativeQuery extends AtomicQuery {
         this.queryText().length > 0 &&
         this._allTemplateTagsAreValid(),
     );
+  }
+
+  canBeSaved(): boolean {
+    return this.canRun() && this._allRequiredTagsHaveDefaults();
   }
 
   isEmpty() {
@@ -295,11 +298,6 @@ export default class NativeQuery extends AtomicQuery {
         if (!dimension) {
           return new ValidationError(t`Invalid template tag: ${tag.name}`);
         }
-        if (tag.required && !tag.default) {
-          return new ValidationError(
-            t`Missing default value for a required template tag: ${tag.name}`,
-          );
-        }
 
         return dimension.validateTemplateTag();
       })
@@ -311,6 +309,18 @@ export default class NativeQuery extends AtomicQuery {
   private _allTemplateTagsAreValid() {
     const tagErrors = this._validateTemplateTags();
     return tagErrors.length === 0;
+  }
+
+  private _allRequiredTagsHaveDefaults() {
+    const isEmpty = v => {
+      const value = Array.isArray(v) ? v[0] : v;
+      return value == null || value === "";
+    };
+
+    const requiredNoDefaults = this.templateTags().filter(
+      tag => tag.required && isEmpty(tag.default),
+    );
+    return requiredNoDefaults.length === 0;
   }
 
   setTemplateTag(name: string, tag: TemplateTag): NativeQuery {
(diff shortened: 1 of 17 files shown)


frontend/src/metabase-lib/v1/queries/NativeQuery.ts
```diff
@@ -313,6 +311,18 @@ export default class NativeQuery extends AtomicQuery {
     return tagErrors.length === 0;
   }
 
+  private _allRequiredTagsHaveDefaults() {
+    const isEmpty = v => {
```
oleggromov, line 315: Is there an existing function to detect an empty value (empty being anything false-y in an array or standalone) I could use here?

ranquild (commented):
@oleggromov I think it makes sense to have this logic in CLJS as `can-run`. I've pushed some draft for the reference. Wdyt?

src/metabase/lib/query.cljc
```diff
@@ -64,6 +64,21 @@
...
+  (fn [query]
+    (:lib/type (lib.util/query-stage query 0))))
+
+(defmethod can-save-method :default
+  [_query]
+  true)
+
+(mu/defn can-save :- :boolean
+  "Returns whether the query can be saved."
+  [query :- ::lib.schema/query]
```
oleggromov, line 78: I don't understand this part exactly.
For a non-native query, will it _always_ return true? I am not sure this is how it has to be but this is how I implemented it in a JS version 😁

ranquild: for mbql queries it will be the same as `can-run`. Is it correct?

oleggromov: Given that it didn't even exist before, I don't think so. Perhaps for MBQL it should always be `true`, this was my assumption.

ranquild: Discussed - it should be `can-run` which checks whether the query is not empty in this case.

src/metabase/lib/native.cljc
```diff
@@ -260,3 +260,10 @@
   [query :- ::lib.schema/query]
   (assert-native-query! (lib.util/query-stage query 0))
   (:engine (lib.metadata/database query)))
+
+(defmethod lib.query/can-save-method :mbql.stage/native
+  [query]
+  (every? (fn [[_k {:keys [required, default]}]]
```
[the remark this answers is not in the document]
bshepherdson: I'm not sure I agree. I think it's easier to understand the logic with this phrasing. It reads pretty naturally: "every template tag must either not be required, or have a default."

src/metabase/lib/native.cljc
```diff
@@ -260,3 +260,10 @@
...
+            (or (not required)
+                (not-empty (if (vector? default) (first default) default))))
```
bshepherdson, line 268: It's not clear to me at a glance why this unwrapping is necessary. I saw it in the TS code too, but I don't know the tag structure that well.

Can you add a comment showing the different cases we need to check for? (Alternatively, split that logic out into `(has-default? template-tag)` and document it there.)

src/metabase/lib/native.cljc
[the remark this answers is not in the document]
bshepherdson: Also it's easier to read the inner function args without the `[k v]` pair. Since we're not using the key, change the function to expect just the map `(fn [{:keys [required default]}] ...)` and make the second arg to `every?` be `(vals (template-tags query))`.

(Note also that we typically don't use commas when destructuring.)

src/metabase/lib/query.cljc
bshepherdson: I see that you've discussed changing the default value; that makes sense to me.

For understanding, one way to think about how `defmulti` + `defmethod` work is that it resembles a `switch`/`case` with its code scattered across files. The function on the `defmulti` is the `switch (...)` argument, it returns the value we're switching on. Then you specify `case`s for particular values with `(defmethod lib.query/can-save-method :mbql.stage/native ...)`, as we do above in `metabase.lib.native`.

If none of the overrides matches, the `:default` case is used instead, just like the `default:` clause of a `switch` statement.

(There's more to multimethods, but that's a solid analogy for basic uses like this one.)

test/metabase/lib/native_test.cljc
```diff
@@ -314,6 +314,65 @@
...
 
+(deftest ^:parallel can-save-native-test
+  (is (lib/can-save (lib/native-query meta/metadata-provider "select 1")))
+  (is (lib/can-save (lib/native-query meta/metadata-provider "select * {{foo}}")))
+  (is (lib/can-save (lib/with-template-tags
+                      (lib/native-query meta/metadata-provider "select * {{foo}}")
+                      {"foo" {:type :text
+                              :id "1"
+                              :name "foo"
+                              :display-name "foo"}})))
```
bshepherdson, line 325: I think this would be easier to read with a helper function, since there's a low "signal-to-noise" ratio in these tests. It's tricky to spot the differences between them.

I'd define
```clojure
(defn- query-with-default [default]
  (lib/with-template-tags
   (lib/native-query meta/metadata-provider "select * {{foo}}")
   {"foo" {:type :text
           :id "1"
           :name "foo"
           :display-name "foo"
           :required true
           :default default}}))
```
and then the tests become
`(is (lib/can-save (query-with-default "A")))` and so on. Setting `:default nil` works for this first case, where we don't want to set the default.

(The first few with simpler queries and no template tag can stay as they are.)

bshepherdson (changes requested):
Coming along, but I think we can tidy it up a bit.

ranquild:
@bshepherdson addressed the comments. Please let me know if I forgot something

bshepherdson (approved):

ranquild (approved):

PR #40289, opened 2024-03-19
Convert `arrow.module.css` to CSS Modules

### LINKED ISSUE #40231: Migrate global css in `frontend/src/metabase/css/core/animation.module.css`, `frontend/src/metabase/css/core/arrow.module.css`

oisincoveney:
### Description

Converts `arrow.module.css` to CSS modules. There is only one place that this class is used, it might make more sense to put it into a local `.modules.css` file that's specific to the component. Let me know what you think.

### How to verify

Go to `Admin > Table Metadata > Segments` OR `Metrics`, then click `New Segment/Metric`. You should see under the `Data` element that there's a subtle arrow. That's the `arrowRight` class that has been converted to CSS modules. 

image

### Checklist


frontend/src/metabase/css/core/arrow.module.css
```diff
@@ -2,13 +2,13 @@
 
 /* TODO: other arrow directions */
 
-:global(.arrow-right) {
+.arrowRight {
```
WiNloSt, line 5: > There is only one place that this class is used, it might make more sense to put it into a local `.modules.css` file that's specific to the component. Let me know what you think.

Yeah, I think so, so we wouldn't expect to use this elsewhere. If this file is in `css/core` it feels as if it's meant to be used through out the app.

frontend/src/metabase/admin/datamodel/components/GuiQueryEditor/GuiQueryEditor.jsx
```diff
@@ -225,9 +225,12 @@ export class GuiQueryEditor extends Component {
...
     return (
       <div
-        className={
-          "GuiBuilder-section GuiBuilder-data flex align-center arrow-right"
-        }
+        className={cx(
+          CS.arrowRight,
+          CS.alignCenter,
+          CS.flex,
+          "GuiBuilder-section GuiBuilder-data",
```
WiNloSt, line 232: 2 things here.
1. Please try to preserve the CSS order, so it's easier to compare to previous versions.
2. Please also coordinate this with @deniskaber since his PR migrated these 2 classes

WiNloSt (approved):
I'll approve it, but it'd be nice to address some comments. Also make sure to work with Denis so you both are aware that you're touching the same file.

frontend/src/metabase/admin/datamodel/components/GuiQueryEditor/GuiQueryEditor.jsx
oisincoveney: Yep, of course. I think it will be ok in terms of migration since I'm only making these small changes, and preserving the order of the classes _should_ help make sure there isn't a merge conflict here

Issue #38052, opened 2024-01-23
Flaky postgres driver tests


noahmoss:

noahmoss:
```
ERROR in metabase.api.dashboard-test/dashboard-chain-filter-test (impl.clj:219)
Uncaught exception, not in assertion.

clojure.lang.ExceptionInfo: Couldn't find Field "PRODUCT_ID" for Table "REVIEWS".
                            Found:
                            {6326 "ID", 6327 "LONGITUDE", 6328 "PRICE", 6329 "CATEGORY_ID", 6330 "NAME", 6331 "LATITUDE"}
    all-fields: {6326 "ID", 6327 "LONGITUDE", 6328 "PRICE", 6329 "CATEGORY_ID", 6330 "NAME", 6331 "LATITUDE"}
    field-name: "PRODUCT_ID"
         table: "REVIEWS"
      table-id: 1059
metabase.test.data.impl/throw-unfound-field-errror            impl.clj:  219
             metabase.test.data.impl/the-field-id*            impl.clj:  229
                     metabase.test.data.impl/fn/&f            impl.clj:  236
                                               ...                          
                                clojure.core/apply            core.clj:  671
                     metabase.test.data.impl/fn/fn            impl.clj:  231
                                               ...                          
              metabase.test.data.impl/copy-db-fks!            impl.clj:  282
metabase.test.data.impl/copy-db-tables-and-fields!            impl.clj:  287
   metabase.test.data.impl/do-with-temp-copy-of-db            impl.clj:  326
                    metabase.api.dashboard-test/fn  dashboard_test.clj: 2825
                          clojure.test/test-var/fn            test.clj:  717
                             clojure.test/test-var            test.clj:  717
                             mb.hawk.core/run-test            core.clj:  136
                     eftest.runner/test-vars/fn/fn          runner.clj:  106
                      clojure.test/default-fixture            test.clj:  687
                        eftest.runner/test-vars/fn          runner.clj:   98
             eftest.runner/wrap-test-with-timer/fn          runner.clj:   38
                               clojure.core/map/fn            core.clj: 2770
                                               ...                          
                                 clojure.core/next            core.clj:   64
                                clojure.core/dorun            core.clj: 3143
                        eftest.runner/test-vars/fn          runner.clj:  115
              metabase.test.fixtures/initialize/fn        fixtures.clj:   12
               clojure.test/compose-fixtures/fn/fn            test.clj:  694
                      clojure.test/default-fixture            test.clj:  687
[...]

PR #37473, opened 2024-01-09
[Feature branch] Permissions schema + API change

noahmoss:
Feature branch for https://github.com/metabase/metabase/issues/26918

This should _only_ contain changes that have already been code reviewed and fully passed CI.

PR #40291, opened 2024-03-19
Fix flaking tests using `mt/with-temp-copy-of-db`

calherries:
(hopefully) solves https://github.com/metabase/metabase/issues/38052 (also the dupe https://github.com/metabase/metabase/issues/39861)

This fixes nondeterministic behaviour in `metabase.test.data.impl/copy-db-tables!` which can result in accidental mismatches between old and new copies of tables. When a Table instance is copied, we can accidentally associate fields from another table to that table, e.g. the new `CHECKINS` fields having a `table_id` that references the new `USERS` table instead of the new `CHECKINS` table. This explains why we can get puzzling errors like this:

```
ERROR in metabase-enterprise.sandbox.models.params.field-values-test/get-or-create-advanced-field-values!-test (impl.clj:219)

clojure.lang.ExceptionInfo: Couldn't find Field "USER_ID" for Table "CHECKINS".
                            Found:
                            {2345 "ID", 2346 "LAST_LOGIN", 2347 "PASSWORD", 2348 "NAME"} <---- these are columns from the USER table, not the CHECKINS table
```

`copy-db-tables!` has always had undeterministic behaviour, but a recent change probably made it turn into an observable problem.

I don't know exactly why it has come up now, but my best guess is Noah's https://github.com/metabase/metabase/issues/37473, which changed the way permissions checks are done for the model table and coincided with the start of the flakes.

### DIFF
--- a/test/metabase/test/data/impl.clj
+++ b/test/metabase/test/data/impl.clj
@@ -265,9 +265,9 @@
 
 (defn- copy-db-tables! [old-db-id new-db-id]
   (let [old-tables    (t2/select Table :db_id old-db-id, :active true, {:order-by [[:id :asc]]})
-        new-table-ids (t2/insert-returning-pks! Table
+        new-table-ids (sort (t2/insert-returning-pks! Table
                         (for [table old-tables]
-                          (-> table (dissoc :id) (assoc :db_id new-db-id))))]
+                          (-> table (dissoc :id) (assoc :db_id new-db-id)))))]
     (doseq [[old-table-id new-table-id] (zipmap (map :id old-tables) new-table-ids)]
       (copy-table-fields! old-table-id new-table-id))))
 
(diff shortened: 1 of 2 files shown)


test/metabase/test/data/impl.clj
```diff
@@ -265,9 +265,10 @@
 
 (defn- copy-db-tables! [old-db-id new-db-id]
   (let [old-tables    (t2/select Table :db_id old-db-id, :active true, {:order-by [[:id :asc]]})
-        new-table-ids (t2/insert-returning-pks! Table
-                        (for [table old-tables]
-                          (-> table (dissoc :id) (assoc :db_id new-db-id))))]
+        new-table-ids (sort ; insert-returning-pks! doesn't order by pk or by insertion order
```
crisptrutski, line 268: I would explicitly say that ordering by pk recovers the insertion order

crisptrutski (approved):

test/metabase/test/data/impl.clj
calherries: yep. went with `sorting by PK recovers the insertion order, because insert-returning-pks! doesn't guarantee this`

Issue #39046, opened 2024-02-22
BE flakes: redshift-lbv-sync-error-test and redshift-types-test


calherries:
These are two seemingly related tests flaking on master. Example runs:

`metabase.driver.redshift-test/redshift-types-test`: https://github.com/metabase/metabase/actions/runs/7979299717/job/21786414703#step:3:954

`metabase.driver.redshift-test/redshift-lbv-sync-error-test `:
`https://github.com/metabase/metabase/actions/runs/7978492812/job/21785111342#step:3:920`

Judging by the stack trace, the only info we have is that a redshift exception is being thrown inside `metabase.sync.sync-metadata/sync-db-metadata!`

WiNloSt:
fixed by https://github.com/metabase/metabase/pull/40310

PR #40310, opened 2024-03-19
Fake test dataset isolation for redshift

### LINKED ISSUE #40058: Proposal to make redshift testing more robust

calherries:
Resolves https://github.com/metabase/metabase/issues/40058 and https://github.com/metabase/metabase/issues/39046

This PR avoids duplicating Table instances between databases during test runs with a simple trick: we limit sync for a database to only sync tables that are prefixed with the DB name. We prefix table names with the DB name already to avoid name conflicts.

### Context

We currently use one redshift database for all test runs.  

We "isolate" test runs from each other using a schema, and that means all test tables are in the same schema. So every time we fake creating a new test "database", we sync all the tables in all test databases that were previously created. 

That leads to many duplicated Table instances: one for each fake test database.

That's a waste of resources and time, but it also leads to test failures and flakes when tests expect databases to have non-overlapping sets of tables.

### What's in it for me? 

Fewer redshift flakes, no more commenting out tests, and faster test runs.


dev/src/dev.clj
```diff
@@ -212,7 +212,7 @@
   first arg:
 
     (dev/query-jdbc-db
-     [:sqlserver 'test-data-with-time]
+     [:sqlserver 'time-test-data]
```
calherries, line 215: I renamed `test-data-with-timezones` -> `timezone-test-data` to keep tables under 30 characters.
so I renamed `test-data-with-time` -> `time-test-data` for consistency

modules/drivers/redshift/test/metabase/driver/redshift_test.clj
```diff
@@ -369,15 +369,15 @@
            (finally
             (execute! "DROP TABLE IF EXISTS %s CASCADE;" qual-tbl-nm))))))))
 
-(mt/defdataset numeric-unix-timestamps
+(mt/defdataset unix-timestamps
```
calherries, line 372: I renamed this to keep table names under 30 characters. explained later

modules/drivers/redshift/test/metabase/test/data/redshift.clj
```diff
@@ -208,3 +209,18 @@
...
+;; For test databases, only sync the tables that are qualified by the db name
+(defmethod driver/describe-database :redshift
+  [driver database]
+  (if *use-original-filtered-syncable-schemas-impl?*
+    (original-describe-database driver database)
+    (let [r (original-describe-database driver database)]
+      (update r :tables (fn [tables]
+                          (into #{}
+                                (filter #(tx/qualified-by-db-name? (:name database) (:name %)))
+                                tables))))))
```
calherries, line 226: transducing in hope that `tables` becomes a reducible one day that doesn't hold every table in memory

test/metabase/query_processor_test/date_bucketing_test.clj
```diff
@@ -900,7 +900,7 @@
   [^TimestampDatasetDef this]
   (let [interval-seconds (.intervalSeconds this)]
     (mt/dataset-definition
-     (str "checkins_interval_" interval-seconds)
+     (str "interval_" interval-seconds)
```
calherries, line 903: this seemed like a good enough name for the dataset, since there's only one table anyway: "checkins"

test/metabase/test/data/interface.clj
```diff
@@ -213,21 +213,33 @@
   Currently, this only affects `db-qualified-table-name`."
   nil)
 
+(defn- normalize-qualified-name [n]
+  (-> n u/lower-case-en (str/replace #"-" "_")))
+
+(defn- db-qualified-table-name-prefix [db-name]
+  (str (normalize-qualified-name (or *database-name-override* db-name))
+       \_))
```
calherries, line 221: I extracted this logic out so it could be reused in both `db-qualified-table-name` and `qualified-by-db-name?`, making the coupling explicit

modules/drivers/redshift/test/metabase/test/data/redshift.clj
```diff
@@ -192,3 +193,25 @@
    (sql-jdbc.conn/connection-details->spec driver @db-connection-details)
    {:write? true}
    delete-session-schema!))
+
+(def ^:dynamic *use-original-describe-database-impl?*
```
crisptrutski, line 197: Inverting this boolean to be explicit about why we'd override it would be clearer, e.g. something like `*override-describe-database-to-filter-by-db-name?*`.

test/metabase/test/data/interface.clj
```diff
@@ -213,21 +213,33 @@
...
-      (str \_ table-name)
-      u/lower-case-en
-      (str/replace #"-" "_")
-      (->>
-        (take-last 30)
-        (apply str))))
+  {:pre [(string? database-name)
+         (string? table-name)]
+   :post [(qualified-by-db-name? database-name %)
+          (<= (count %) 30)]}
```
crisptrutski, line 240: This appears to have broken the Presto tests:

```
 ERROR in metabase.query-processor-test.timezones-test/native-sql-params-filter-test (interface.clj:228)

java.lang.AssertionError: Assert failed: (<= (count %) 30)
                 metabase.test.data.interface/db-qualified-table-name       interface.clj:  228
```

crisptrutski (commented):
Ingenious! Just waiting to see a clean CI run as not all of the failures look like flakes.

One concern is whether we could ever have a test that triggers table creation in a way that doesn't apply this prefix. Perhaps unlikely, and we'd probably want to just fix that test then, but perhaps there's a way we can make this trick more obvious to save a developer tearing their hair out for a bit.

test/metabase/test/data/interface.clj
crisptrutski: Ah see you figured this out already 😄

modules/drivers/redshift/test/metabase/test/data/redshift.clj
calherries: https://github.com/metabase/metabase/blob/e6fe24f3aa6f151b50af5bfd4983a1863830edde/modules/drivers/redshift/test/metabase/test/data/redshift.clj#L197-L200

test/metabase/test/data/interface.clj
calherries: Yeah, got a bit confident with asking for a review before all tests passed

calherries:
> One concern is whether we could ever have a test that triggers table creation in a way that doesn't apply this prefix. Perhaps unlikely, and we'd probably want to just fix that test then, but perhaps there's a way we can make this trick more obvious to save a developer tearing their hair out for a bit.

Yes, that's a fair concern. That's unfortunately a similar problem with writing tests for redshift anyway - you need to understand that when you create a table, you create it in the unique schema for the test run.

I've added a docstring to the redshift test data namespace so hopefully that gives new developers some context. 
https://github.com/metabase/metabase/blob/097b8fe42803790bd18a92e1c44a0538bab01d4a/modules/drivers/redshift/test/metabase/test/data/redshift.clj#L2-L11

crisptrutski (approved):

